Issue #2790, opened 2023-10-10
Tracking issue: Discovery API


coszio:
To track the development of a new type of search: discovery. This issue will be updated when there are updates

## Tasks
- #2791 
- #2809
- #2820 
- #2834
- #2836 
- #2851 
- #2867
- #2861 
- qdrant/qdrant-client#368
- qdrant/landing_page#408

## Backlog
- [x] Discovery custom scorer (rank + sigmoided score)
- [x] Integrate discovery custom scorer into `QueryVector` enum
- [x] Context of pairs custom scorer (find best zone)
- [x] Allow custom entry point in hnsw
- [x] Discovery internals (2-step search)
    - [x] First step: find best entrypoint
    - [x] Second step: start searching with custom scorer from entrypoint
    - [x] Preprocessing logic for converting ids to vectors
- [x] Interface
    - [x] Use same vector fetching logic from recommendations
    - [x] Create single and batch endpoints 
- [x] Client
  - [x] Local mode
  - [x] rest and grpc clients
- [ ] Docs 
  - [x] qdrant documentation
  - [ ] article

PR #2867, opened 2023-10-23
Add `DiscoveryRequest` and preprocessing

coszio:
Needs #2851, tracked by #2790

- Created `DiscoveryRequest`
- Added `discovery.rs` with the logic for transforming a request with `RecommendExamples` to a `CoreSearchRequest`, so it includes all the batched vector fetching logic.
 
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/discovery.rs
+++ b/lib/collection/src/discovery.rs
@@ -1,0 +1,234 @@
+use futures::Future;
+use itertools::Itertools;
+use segment::data_types::vectors::{NamedQuery, NamedVector, DEFAULT_VECTOR_NAME};
+use segment::types::{Condition, Filter, HasIdCondition, ScoredPoint};
+use segment::vector_storage::query::context_query::{ContextPair, ContextQuery};
+use segment::vector_storage::query::discovery_query::DiscoveryQuery;
+use tokio::sync::RwLockReadGuard;
+
+use crate::collection::Collection;
+use crate::common::fetch_vectors::{convert_to_vectors, PointRef, ReferencedPoints};
+use crate::operations::consistency_params::ReadConsistency;
+use crate::operations::types::{
+    CollectionError, CollectionResult, CoreSearchRequest, CoreSearchRequestBatch, DiscoverRequest,
+    DiscoverRequestBatch, QueryEnum, RecommendExample, UsingVector,
+};
+
+pub async fn discover<'a, F, Fut>(
+    request: DiscoverRequest,
+    collection: &Collection,
+    collection_by_name: F,
+    read_consistency: Option<ReadConsistency>,
+) -> CollectionResult<Vec<ScoredPoint>>
+where
+    F: Fn(String) -> Fut,
+    Fut: Future<Output = Option<RwLockReadGuard<'a, Collection>>>,
+{
+    if request.limit == 0 {
+        return Ok(vec![]);
+    }
+    // `discover` is a special case of discover_batch with a single batch
+    let request_batch = DiscoverRequestBatch {
+        searches: vec![request],
+    };
+    let results = discover_batch(
+        request_batch,
+        collection,
+        collection_by_name,
+        read_consistency,
+    )
+    .await?;
+    Ok(results.into_iter().next().unwrap())
+}
+
+pub async fn discover_batch<'a, F, Fut>(
+    request_batch: DiscoverRequestBatch,
+    collection: &Collection,
+    collection_by_name: F,
+    read_consistency: Option<ReadConsistency>,
+) -> CollectionResult<Vec<Vec<ScoredPoint>>>
+where
+    F: Fn(String) -> Fut,
+    Fut: Future<Output = Option<RwLockReadGuard<'a, Collection>>>,
+{
+    // shortcuts batch if all requests with limit=0
+    if request_batch.searches.iter().all(|s| s.limit == 0) {
+        return Ok(vec![]);
+    }
+
+    // Validate context_pairs and/or target have value(s)
+    request_batch.searches.iter().try_for_each(|request| {
+        let no_pairs = request.context_pairs.is_none()
+            || request
+                .context_pairs
+                .as_ref()
+                .is_some_and(|pairs| pairs.is_empty());
+
+        let no_target = request.target.is_none();
+
+        if no_pairs && no_target {
+            return Err(CollectionError::bad_request(
+                "target and/or context_pairs must be specified".to_string(),
+            ));
+        }
+
+        Ok(())
+    })?;
+
+    // Pack all referenced ids for retrieving them once
+    let mut id_references = ReferencedPoints::new();
+    request_batch.searches.iter().for_each(|request| {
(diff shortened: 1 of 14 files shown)


lib/collection/src/operations/conversions.rs
```diff
@@ -796,6 +796,8 @@ impl From<QueryEnum> for api::grpc::qdrant::QueryEnum {
                     },
                 )),
             },
+            QueryEnum::Discover(_) => todo!("luis in other PR"),
+            QueryEnum::Context(_) => todo!("luis in other PR"),
```
timvisee, line 800: Do we guarantee not to call this in the current implementation?

lib/collection/src/operations/types.rs
```diff
@@ -560,6 +574,69 @@ pub struct RecommendGroupsRequest {
...
+/// When using both context and target, the score behaves a little different: The
+/// integer part of the score represents the "rank" with respect to the context, while the
+/// decimal part of the score relates to the distance to the target.
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Validate, Clone)]
+pub struct DiscoverRequest {
+    /// Look for vectors closest to this
+    pub target: Option<RecommendExample>,
+
+    /// Pairs of (positive, negative) examples to provide context to the search
+    pub context_pairs: Option<Vec<(RecommendExample, RecommendExample)>>,
```
timvisee, line 597: Shall we drop pairs that have the exact same A/B vector?

We can probably do that in the search implementation with `pairs.filter(A != B)`.

We can also add validation here to prevent it, but I guess that might break the discovery API unexpectedly for some users.

lib/collection/src/discovery.rs
```diff
@@ -0,0 +1,234 @@
...
+            if !referenced_vectors.contains_key(&PointRef {
+                collection_name: lookup_collection_name,
+                point_id,
+            }) {
+                return Err(CollectionError::PointNotFound {
+                    missed_point_id: point_id,
+                });
+            }
+            Ok(())
+        })?;
```
timvisee, line 117: A `.iter().find()` or `if .iter().any()` may be a bit nicer here. What do you think?

lib/collection/src/discovery.rs
```diff
@@ -0,0 +1,234 @@
...
+                query: DiscoveryQuery::new(target, pairs),
+                using: Some(lookup_vector_name),
+            }),
+
+            // Only pairs => Context
+            (None, pairs) => QueryEnum::Context(NamedQuery {
+                query: ContextQuery::new(pairs),
+                using: Some(lookup_vector_name),
+            }),
+        };
```
timvisee, line 170: It makes a bit more sense to me if it is ordered like this:

```suggestion
        let query: QueryEnum = match (target, context_pairs) {
            // Target and pairs => Discovery
            (Some(target), pairs) if !pairs.is_empty() => QueryEnum::Discover(NamedQuery {
                query: DiscoveryQuery::new(target, pairs),
                using: Some(lookup_vector_name),
            }),

            // Only target => Nearest
            (Some(target), _pairs) => QueryEnum::Nearest(
                NamedVector {
                    name: lookup_vector_name,
                    vector: target,
                }
                .into(),
            ),

            // Only pairs => Context
            (None, pairs) => QueryEnum::Context(NamedQuery {
                query: ContextQuery::new(pairs),
                using: Some(lookup_vector_name),
            }),
        };
```

- Discovery
- Nearest
- Context

lib/collection/src/discovery.rs
```diff
@@ -0,0 +1,234 @@
...
+            (Some(target), pairs) => QueryEnum::Discover(NamedQuery {
+                query: DiscoveryQuery::new(target, pairs),
+                using: Some(lookup_vector_name),
+            }),
+
+            // Only pairs => Context
+            (None, pairs) => QueryEnum::Context(NamedQuery {
+                query: ContextQuery::new(pairs),
+                using: Some(lookup_vector_name),
+            }),
```
timvisee, line 169: What if `pairs` is empty now?

lib/collection/src/discovery.rs
```diff
@@ -0,0 +1,234 @@
...
+}
+
+fn iterate_examples(request: &DiscoverRequest) -> impl Iterator<Item = &RecommendExample> {
+    request
+        .context_pairs
+        .iter()
+        .flat_map(|pairs| {
+            pairs
+                .iter()
+                .flat_map(|pair| std::iter::once(&pair.0).chain(std::iter::once(&pair.1)))
```
timvisee, line 218: Nitpick, I guess we can:

```rust
use std::iter;

iter::once(...);
```

timvisee (commented):
Some remarks, suggestions and nitpicks.

Please feel free to ignore some of them if they feel unnecessary. It's my first look through Discovery code :smile:.

lib/segment/src/data_types/vectors.rs
```diff
@@ -320,12 +320,12 @@ impl BatchVectorStruct {
 }
 
 #[derive(Debug, Clone)]
-pub struct NamedRecoQuery {
-    pub query: RecoQuery<VectorType>,
+pub struct NamedQuery<T> {
+    pub query: T,
```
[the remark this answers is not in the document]
coszio: We don't actually use that constraint, but I like that it should improve readability

lib/collection/src/operations/types.rs
```diff
@@ -560,6 +574,69 @@ pub struct RecommendGroupsRequest {
...
+///
+/// Using only a target is equivalent to regular search, so the score is the distance to the target.
+///
+/// When using both context and target, the score behaves a little different: The
+/// integer part of the score represents the "rank" with respect to the context, while the
+/// decimal part of the score relates to the distance to the target.
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Validate, Clone)]
+pub struct DiscoverRequest {
+    /// Look for vectors closest to this
+    pub target: Option<RecommendExample>,
```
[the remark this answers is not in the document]
coszio: Unfortunately we already released `RecommendExample` in v1.6, in hindsight I would remove the `Recommend` part, as it is just something that can be either a vector or an ID

lib/collection/src/operations/conversions.rs
coszio: Yes, we haven't exposed any apis at this point, so there would be no reason that this variant would be run

lib/collection/src/operations/types.rs
coszio: Good question, in the case of Discovery search, rank would be 0 if distance is the same against both pairs, so it should not have any adverse effect. For Context search, we add a very small margin so it may behave a little weird.

I think it is a good idea to cases in preprocessing 👍

lib/collection/src/discovery.rs
coszio: We already validated the case of `no_target && no_pairs` at the start of the function 😀

timvisee (approved):

PR #2869, opened 2023-10-24
Hnsw parallel build accuracy

generall:
This PR contains several fixes for the HNSW parallel index building process

related to: https://discord.com/channels/907569970500743200/1164130123868557392

> We have been bnechmarking recall@1 of qdrant HNSW index using 60K, 100K and 1M 1024-dimensional embeddings.
Within each benchmarking process we have been waiting for collection status to turn 'green' before running the search.
After comparing the results for different HNSW parameters (M, ef_construct, ef_search), we have discovered that the parameter 'M' works in a very unexpected way (the lower the better).
We have reproduced the same experiment for FAISS HNSW index and got the expected results for 'M' (the higher the better) with higher recall@1 values.
Shortly after, we have discovered that the problem with qdrant HNSW recall was caused by small number of unindexed vectors.
Despite that the collection status has turned 'green', we've had around 2k unindexed vectors from a total of 90k, which was the cause of drastic change in our metrics.
We have found that we can change this behaviour by switching the indexing_threshold_kb in IndexingOptimizer such that it waits for all vectors to be indexed.
In our case this change had a significant impact on our metrics, especially results for 'M' parameter and 10% higher recall@1 values (despite only ~3% change in the number of indexed vectors).
We understand that the default value of this threshold has been set to compromise index build time and searching recall, however the drop in recall metrics is significant 
and we want to ask if this behaviour is proper for the index and are there any good practices that you could recommend for getting high search recall@1 value while having reasonable index build time.
I am adding two charts to this message to give a better visualization of the problem. Both represent recall@1 values for hnsw index with ef_construct=64. 
The one with present m=8 is the one with ~3% unindexed vectors, the one without m=8 is the one with all vectors indexed.

What we discovered during investigation:

- Sometime during parallel build of HNSW it might create disconnected sub-graphs, which leads to poor search accuracy, even with very high EF value.
- The disconnection effect is observed almost always, but the size of the effect varies between runs
- Effect disappears if: single thread is used, or only one hnsw level is used

Hypothesis:

During the parallel indexing, it was possible, the under-constructed point was used as an entry point on N-1 level.
Which led to incorrect linking for some fraction of points. If those points happened to be closer to original entry point - the large chunk of the graph became disconnected.

What this PR does:

- Introduce test for HNSW graph connectivity in case of parallel building
- Introduce a new check o index building stage: allow to traverse only those points, which are completely constructed by other threads
- Introduce initial building of first `K=512` points with single threaded  builder to improve initial spread of points and ensure warm start for parallel building process

### DIFF
--- a/lib/segment/src/index/hnsw_index/entry_points.rs
+++ b/lib/segment/src/index/hnsw_index/entry_points.rs
@@ -37,11 +37,6 @@ impl EntryPoints {
             extra_entry_points: FixedLengthPriorityQueue::new(extra_entry_points),
         }
     }
-
-    pub fn iterator(&self) -> impl Iterator<Item = &EntryPoint> {
-        self.entry_points.iter()
-    }
-
     pub fn merge_from_other(&mut self, mut other: EntryPoints) {
         self.entry_points.append(&mut other.entry_points);
         // Do not merge `extra_entry_points` to prevent duplications
--- a/lib/segment/src/index/hnsw_index/entry_points.rs
+++ b/lib/segment/src/index/hnsw_index/entry_points.rs
@@ -38,6 +38,10 @@ impl EntryPoints {
         }
     }
 
+    pub fn iterator(&self) -> impl Iterator<Item = &EntryPoint> {
+        self.entry_points.iter()
+    }
+
     pub fn merge_from_other(&mut self, mut other: EntryPoints) {
         self.entry_points.append(&mut other.entry_points);
         // Do not merge `extra_entry_points` to prevent duplications
--- a/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
@@ -894,7 +894,7 @@ mod tests {
             GraphLayersBuilder::connect_new_point(&mut links, id, 0, level_m, scorer)
         }
         let mut result = Vec::new();
-        graph_layers_builder.links_map(0, 0, |link| result.push(link));
+        graph_layers_builder.links_layers[0][0].read().iter().for_each(|x| result.push(*x));
         assert_eq!(&result, &vec![1, 2, 3, 4, 5, 6]);
     }
 }
--- a/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
@@ -282,6 +282,7 @@ impl GraphLayersBuilder {
         result_list
     }
 
+    #[allow(dead_code)]
     fn search_existing_point_on_level(
         &self,
         level_entry: ScoredPointOffset,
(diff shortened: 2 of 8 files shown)


generall:
There is, however, slight performance reduction during indexing, which is to be expected:

```
Indexing time


old 50k

m: 16

26.225863178 - 21.458603293 = 4.76726
60.845421938 - 57.025841127 = 3.819580811

m: 32
55.255734155 - 51.444479446 = 3.811254709
12.748021045 - 8.767797220 = 3.980223825 

m: 64
28.977501937 - 25.491052888 = 3.486449049
23.317064256 - 20.005943829 = 3.311120427

new 50k

m: 16
56.869508763 - 52.745561647 = 4.123947116
53.037147406 - 48.920342120 = 4.116805286

m: 32
38.028887134 - 34.001564295 = 4.027322839
52.900207227 - 49.201557128 = 3.698650099


m: 64
20.122459039 - 16.364955756 = 3.757503283
55.910761077 - 52.180513751 = 3.730247326


-----

1M segment, old 
    48:16.761876464 -
    42:50.662556577 
    =
    5:26.099319887

1M segment, new
    38:09.985085651 -
    32:32.850728207
    =
    5:37.134357444
```

Or about 3%

lib/segment/src/index/hnsw_index/graph_layers_builder.rs
```diff
@@ -365,29 +376,51 @@ impl GraphLayersBuilder {
...
-                            Self::select_candidates_with_heuristic(nearest_points, level_m, scorer);
-                        self.links_layers[point_id as usize][curr_level]
-                            .write()
-                            .clone_from(&selected_nearest);
+                        let selected_nearest = {
+                            let mut existing_links =
+                                self.links_layers[point_id as usize][curr_level].write();
+                            {
+                                let ready_list = self.ready_list.read();
+                                for &existing_link in existing_links.iter() {
```
IvanPleshkov, line 404: This logic of existing links was moved from `search_existing_point_on_level`. You may copy this logic for non-heuristic case too

lib/segment/src/index/hnsw_index/entry_points.rs
```diff
@@ -38,6 +38,10 @@ impl EntryPoints {
         }
     }
 
+    pub fn iterator(&self) -> impl Iterator<Item = &EntryPoint> {
```
IvanPleshkov, line 41: This function is not needed, no any reference to it

IvanPleshkov (changes requested):
Looks good. Request change because of non-heuristic case: it seems `existing_links` check has to be added

lib/segment/src/index/hnsw_index/graph_layers_builder.rs
```diff
@@ -422,7 +426,7 @@ impl GraphLayersBuilder {
                             }
                         }
                     } else {
-                        for nearest_point in &nearest_points {
+                        for nearest_point in &search_context.nearest {
```
IvanPleshkov, line 429:
```suggestion
                        {
                            let mut existing_links =
                                    self.links_layers[point_id as usize][curr_level].write();
                            let ready_list = self.ready_list.read();
                            for &existing_link in existing_links.iter() {
                                if !visited_list.check(existing_link)
                                    && ready_list[existing_link as usize]
                                {
                                    search_context.process_candidate(ScoredPointOffset {
                                        idx: existing_link,
                                        score: points_scorer.score_point(existing_link),
                                    });
                                }
                            }
                        }
                        let mut nearest = search_context.nearest;
                        for nearest_point in &search_context.nearest {
```

IvanPleshkov (approved):

Issue #2432, opened 2023-08-14
Tracking issue: shard snapshot transfer with index


timvisee:
This issue tracks progress on the feature to use snapshots for shard transfers. The shard snapshot will include index and quantization data. This should make the transfer process faster and prevents rebuilding index or quantization data. This is a work in progress, the issue will be updated to reflect latest developments.

### Tasks

- [x] https://github.com/qdrant/qdrant/pull/2406
- [x] https://github.com/qdrant/qdrant/pull/2410
- [x] https://github.com/qdrant/qdrant/pull/2421
- [x] https://github.com/qdrant/qdrant/issues/2532
    - [x] https://github.com/qdrant/qdrant/pull/2571
    - [x] https://github.com/qdrant/qdrant/pull/2562
    - [x] https://github.com/qdrant/qdrant/pull/2563
    - [x] https://github.com/qdrant/qdrant/pull/2554
- [x] https://github.com/qdrant/qdrant/pull/2434
- [x] https://github.com/qdrant/qdrant/pull/2657
- [x] https://github.com/qdrant/qdrant/pull/2813
- [x] https://github.com/qdrant/qdrant/pull/2825
- [x] https://github.com/qdrant/qdrant/pull/2833
- [x] https://github.com/qdrant/qdrant/pull/2873
- [x] https://github.com/qdrant/qdrant/pull/2874
- [x] https://github.com/qdrant/qdrant/pull/2858
- [x] https://github.com/qdrant/qdrant/pull/2881
- [x] https://github.com/qdrant/qdrant/pull/2843
- [x] https://github.com/qdrant/qdrant/pull/2846
- [x] https://github.com/qdrant/qdrant/pull/2467
- [x] https://github.com/qdrant/qdrant/pull/2884
- [x] https://github.com/qdrant/qdrant/pull/2844
- [x] https://github.com/qdrant/qdrant/pull/2870
- [x] https://github.com/qdrant/qdrant/pull/2898
- [x] https://github.com/qdrant/qdrant/pull/2899
- [x] https://github.com/qdrant/qdrant/pull/2903
- [x] https://github.com/qdrant/qdrant/pull/2907
- [x] https://github.com/qdrant/qdrant/pull/2895
- [x] https://github.com/qdrant/qdrant/pull/2951
- [x] https://github.com/qdrant/qdrant/pull/2918
- [x] https://github.com/qdrant/qdrant/pull/2919
- [x] https://github.com/qdrant/qdrant/pull/2924
[...]

PR #2870, opened 2023-10-24
Make shard snapshot methods and API cancel-safe

ffuugoo:
Part of #2432.

We want to make shard snapshot API cancellable, but due to how Rust futures work, arbitrary cancelling some futures in the middle of their execution could lead to logic errors/inconcistent state.

This PR:
- documents if shard snapshot methods are cancel-safe or not
- adds `common/cancel` sub-crate with utilities to spawn safely cancellable futures
- refactor shard snapshot API so that it is safely cancellable
  - this is mostly done by spawning some top-level future and propagating cancellation token into it

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


src/actix/api/snapshot_api.rs
```diff
@@ -333,24 +333,34 @@ async fn upload_shard_snapshot(
...
+        let collection = cancel::future::on_token(cancel.clone(), future).await??;
 
+        // `recover_shard_snapshot_impl` is *not* cancel-safe!
         common::snapshots::recover_shard_snapshot_impl(
             &toc,
             &collection,
             shard,
             form.snapshot.file.path(),
             priority.unwrap_or_default(),
+            cancel,
```
generall, line 356: Let's make cancellation optional for the REST and user-facing API. So that timing out of the recovery/upload requests are still result in complete recovery in the background by default. (this should preserve backward compatibility)

The internal API might be cancellable by default

lib/common/cancel/src/blocking.rs
```diff
@@ -0,0 +1,35 @@
...
+    };
+
+    let guard = cancel.drop_guard();
+    let output = tokio::task::spawn_blocking(task).await?;
+    guard.disarm();
+
+    Ok(output)
+}
+
+pub async fn on_token<Out, Task>(cancel: CancellationToken, task: Task) -> Result<Out, Error>
```
generall, line 22: Suggestion: rename to `cancelable_spawn_blocking_with_token`

src/actix/api/snapshot_api.rs
ffuugoo: Done.

lib/common/cancel/src/blocking.rs
```diff
@@ -0,0 +1,35 @@
+use super::*;
+
+pub async fn on_drop<Out, Task>(task: Task) -> Result<Out, Error>
```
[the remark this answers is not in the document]
ffuugoo: Done.

lib/common/cancel/src/blocking.rs
ffuugoo: Done.

lib/common/cancel/src/blocking.rs
```diff
@@ -19,7 +19,7 @@ where
     Ok(output)
 }
 
-pub async fn on_token<Out, Task>(cancel: CancellationToken, task: Task) -> Result<Out, Error>
+pub async fn span_cancel_on_token<Out, Task>(cancel: CancellationToken, task: Task) -> Result<Out, Error>
```
timvisee:
```diff
-span
+spawn
```

ffuugoo: Fixed. 🤦‍♀️

coszio (approved):
Seems good from my limited knowledge of snapshot internals. Added TODO reminders and nit

lib/collection/src/shards/shard_holder.rs
```diff
@@ -574,7 +634,10 @@ impl ShardHolder {
         this_peer_id: PeerId,
         is_distributed: bool,
         temp_dir: &Path,
+        cancel: cancel::CancellationToken,
     ) -> CollectionResult<()> {
+        // This future is *not* cancel-safe!
```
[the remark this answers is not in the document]
timvisee: I agree, I would prefer that as well.

Maybe it's also enough to just note functions that are **not** cancel safe.

src/common/snapshots.rs
```diff
@@ -61,69 +69,101 @@ pub async fn recover_shard_snapshot(
...
+            let (snapshot_path, snapshot_temp_path) = match snapshot_location {
+                ShardSnapshotLocation::Url(url) => {
+                    if !matches!(url.scheme(), "http" | "https") {
+                        let description = format!(
+                            "Invalid snapshot URL {url}: URLs with {} scheme are not supported",
+                            url.scheme(),
+                        );
+
+                        return Err(StorageError::bad_input(description));
+                    }
```
timvisee, line 93: I thought that checks like this are handled inside the download function. I don't think we have to do it here as well. Please correct me if I'm wrong.

lib/collection/src/shards/shard_holder.rs
```diff
@@ -561,11 +592,40 @@ impl ShardHolder {
...
+        // We already moved the file to the desired destination, but `temp_file` will still try to
+        // delete the file when dropped, so we use `tempfile::NamedTempFile::keep` to consume it safely.
+        //
+        // `tempfile::NamedTempFile::keep` "needs to mark the file as non-temporary" on Windows,
+        // which is a fallible operation. But we already moved the file, so we expect it to fail
+        // and explicitly ignore the result.
+        //
+        // We may mark the wrong file as "non-temporary", if someone creates another file with the
+        // same name in between `move_file` and `tempfile::NamedTempFile::keep` calls, but this is
+        // better than *deleting* it, if we drop `temp_file` as-is
```
timvisee, line 622: I don't think we need all this extra text

```suggestion
        // Temp file is moved away and is persisted, don't delete it anymore
```

src/common/snapshots.rs
```diff
@@ -43,13 +47,17 @@ pub async fn delete_shard_snapshot(
     shard_id: ShardId,
     snapshot_name: String,
 ) -> Result<(), StorageError> {
+    // This future is cancel-safe
+    //
+    // TODO: Explain why?
```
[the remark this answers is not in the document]
timvisee: I assume because there's only a single 'essential' thing happening: remove_file. That either happens or not, and won't leave some broken state.

But, can we cancel the future in the middle of this file operation? And if so, would that leave some broken state?

src/common/snapshots.rs
ffuugoo: `download_file` allows `file://` URLs. Shard snapshot API does not.

timvisee: Check.

Do you think it would make sense to move all functionality for this into a single module, to prevent spreading it all over the code base?

Maybe we can define some specialized functions like: `download_snapshot`, `file_or_download_snapshot`, ... . Or would that only over complicate things?

I know he had this code before, but I'm only thinking about this now.

ffuugoo: Sure... but not in this PR :D

src/actix/api/snapshot_api.rs
```diff
@@ -333,24 +333,34 @@ async fn upload_shard_snapshot(
...
-        collection.assert_shard_exists(shard).await?;
+    let future = cancel::future::spawn_cancel_on_drop(move |cancel| async move {
+        let future = async {
+            let collection = toc.get_collection(&collection).await?;
+            collection.assert_shard_exists(shard).await?;
+
+            Result::<_, helpers::HttpError>::Ok(collection)
+        };
+
+        let collection = cancel::future::cancel_on_token(cancel.clone(), future).await??;
```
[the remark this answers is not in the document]
ffuugoo: Good point, but as far as I understand, `upload_shard_snapshot` is only called after file is fully uploaded (because `form` contains a ready-to-use `TempFile`). So if you cancel the request during file upload, request should be cancelled and partially uploaded file would be removed.

src/common/snapshots.rs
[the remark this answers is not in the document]
ffuugoo: Have you ever seen "partially deleted files"? (:

There are still only two options: either the file is deleted, or not.

lib/storage/src/content_manager/snapshots/recover.rs
```diff
@@ -198,7 +198,11 @@ async fn _do_recover_from_snapshot(
             );
 
             let recovered = collection
-                .recover_local_shard_from(&snapshot_shard_path, *shard_id)
+                .recover_local_shard_from(
+                    &snapshot_shard_path,
+                    *shard_id,
+                    cancel::CancellationToken::new(), // TODO!?
```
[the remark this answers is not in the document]
ffuugoo: I only updated "shard snapshot" API in this PR, and this particular call is inside one of "collection snapshot" APIs. I've had to add `cancel` parameter to the `recover_local_shard_from`, but I don't really use it in "collection snapshot" API yet, so in this case I just pass a cancellation token, that is never triggered, hence the `TODO`.

I'll add a note to this `TODO` now.

src/actix/api/snapshot_api.rs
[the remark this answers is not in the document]
timvisee: Brilliant!

lib/collection/src/shards/shard_holder.rs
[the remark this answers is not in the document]
ffuugoo: I think I'll add both notes, because otherwise it would be unclear which functions are cancel-safe, and which were not yet analyzed.

src/common/snapshots.rs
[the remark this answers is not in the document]
timvisee: I've been thinking too much about futures here :dizzy_face: :dizzy:

ffuugoo:
@generall @timvisee @coszio Ok, this should finally be done now (fixed all comments and failing test). Can you guys, please, give it a final review and approve?

src/actix/api/snapshot_api.rs
```diff
@@ -362,8 +362,8 @@ async fn upload_shard_snapshot(
 
         Result::<_, helpers::HttpError>::Ok(())
     })
-    .map_ok(|_| true)
-    .map_err(Into::into);
+    .map_err(Into::into)
+    .map(|res| res.and_then(|res| res));
```
timvisee, line 366: Too bad [`flatten`](https://doc.rust-lang.org/std/result/enum.Result.html#method.flatten) isn't stabilized yet :(

timvisee (approved):

src/actix/api/snapshot_api.rs
ffuugoo: We will, probably, get annoying `clippy` warning once it does. :D

lib/collection/src/collection/snapshots.rs
```diff
@@ -284,14 +290,24 @@ impl Collection {
...
     pub async fn restore_shard_snapshot(
         &self,
         shard_id: ShardId,
         snapshot_path: &Path,
         this_peer_id: PeerId,
         is_distributed: bool,
         temp_dir: &Path,
+        cancel: cancel::CancellationToken,
     ) -> CollectionResult<()> {
+        // TODO:
```
coszio, line 305: Is this TODO tracked elsewhere?

lib/collection/src/shards/replica_set/snapshots.rs
```diff
@@ -70,8 +70,16 @@ impl ShardReplicaSet {
         Ok(())
     }
 
-    /// Returns if local shard was recovered from path
-    pub async fn restore_local_replica_from(&self, replica_path: &Path) -> CollectionResult<bool> {
+    /// # Safety
+    ///
+    /// This function is *not* cancel-safe!
```
coszio, line 75: I am still a bit confused about the term *cancel-safe* while also adding a cancellation token option 😅. 

Not requesting changes for it, but I kinda wish the term was a Future being *drop-safe*, if only it wasn't an already established term

coszio (approved):

lib/collection/src/shards/replica_set/snapshots.rs
timvisee: Convention is to call it 'cancel safety', so I guess we better stick with it.

It's commonly documented like this, so I guess it's best to follow that exact pattern, sometimes with additional context:

```rust
/// # Cancel safety
///
```

This is on the Tokio codebase: 
image

timvisee: I'll do some Vim magic on it now to make them all follow the same pattern.

lib/collection/src/collection/snapshots.rs
timvisee: I've created an issue for it here so that we don't lose track of it: https://github.com/qdrant/qdrant/issues/2913

PR #2467, opened 2023-08-18
Shard snapshot transfer integration

timvisee:
Tracked in https://github.com/qdrant/qdrant/issues/2432

First end-to-end integration of the shard snapshot transfer feature. There's still more to do to finalize this feature, but that'll be done in different PRs as listed in the [tracking issue](https://github.com/qdrant/qdrant/issues/2432).

So the main purpose of this PR is to connect all the shard transferring bits together. This did require some changes in existing types to make this integration happen. Most of this has already been merged in separate PRs.

In short, this:
- **Adds `ShardTransferMethod`:**
  Allows users to configure to the existing (`StreamRecords`) method or the `Snapshot` method for transferring shards. It currently still defaults to `StreamRecords` if not set.
- **Implements logic to orchestrate shard transfer with snapshot:**
  When `ShardTransferMethod::Snapshot` is selected these changes orchestrates creating a shard snapshot, transferring it, recovering it, transferring all new operations to it and activating it.
- **Transforms the `QueueProxyShard` into a `ForwardProxyShard` during the transfer:**
  When the transfer starts, the local shard is wrapped in a `QueueProxyShard` to track all new operations to it. When the shard snapshot is recovered on the remote peer, all operations are sent to it and the queue proxy shard is transformed into a `ForwardProxyShard`. From then on, operations are simply forwarded during the rest of the transfer. This is done because unwrapping shards from a queue proxy shard can fail while unwrapping from the forward proxy shard cannot fail. This is important as failing here would set the cluster on fire. With this explicit transformation we can handle any queue proxy transfer errors during the shard transfer process.
- **Updates `QueueProxyShard` architecture to allow safe unwrapping of inner shards:**
  The queue proxy shard has been completely reworked to make the above happen.

Tests still have to be added. I'd like to do this in a separate PR, and might swap implementing tests with someone else.

To manually test this, you can do the following:

1. Start a cluster with at least two nodes
2. Create a collection with some data
3. [Move](https://qdrant.tech/documentation/guides/distributed_deployment/#cluster-scaling) a shard with `"method": "snapshot"`
4. Watch the transfer happen

This integration step turned out to be very difficult and required some wild modifications. Please review it with care.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

[...]

PR #2873, opened 2023-10-25
Add consensus interface to shard transfer, repurpose dispatcher

timvisee:
Tracked in https://github.com/qdrant/qdrant/issues/2432.

Merges into https://github.com/qdrant/qdrant/pull/2467.

Adds a consensus interface as trait to be passed into shard transfer. This gives the shard transfer logic access to consensus operations through specialized functions. This is required for shard snapshot transfer.

This PR simply defines the interface (trait) and brings it from our table of contents up to our shard transfer logic. For this I've repurposed `Dispatcher` as it gives access to the consensus proposal channel and the current consensus state.

The implementation is a bit hacky, but it seems to be the best we can do. I've attempted some big refactoring operations but these failed. The other option to achieve this is to merge `storage` and `collection` together, which I like even less.
I suggest to merge it like this. We can look into improving it later, when the shard snapshot transfer feature is complete.

Will merge into https://github.com/qdrant/qdrant/pull/2467 as it requires these changes for approval. The next PR will integrate this change.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/storage/src/content_manager/toc/mod.rs
+++ b/lib/storage/src/content_manager/toc/mod.rs
@@ -34,13 +34,13 @@ use tokio::sync::{Mutex, RwLock, RwLockReadGuard, Semaphore};
 use tonic::transport::Channel;
 use tonic::Status;
 
+use self::transfer::ShardTransferDispatcher;
 use crate::content_manager::alias_mapping::AliasPersistence;
 use crate::content_manager::collection_meta_ops::CreateCollectionOperation;
 use crate::content_manager::collections_ops::{Checker, Collections};
 use crate::content_manager::consensus::operation_sender::OperationSender;
 use crate::content_manager::errors::StorageError;
 use crate::content_manager::shard_distribution::ShardDistributionProposal;
-use crate::dispatcher::Dispatcher;
 use crate::types::{PeerAddressById, StorageConfig};
 use crate::ConsensusOperations;
 
@@ -74,7 +74,7 @@ pub struct TableOfContent {
     /// Effectively, this lock ensures that `create_collection` is called sequentially.
     collection_create_lock: Mutex<()>,
     /// Dispatcher for shard transfer to access consensus.
-    shard_transfer_dispatcher: parking_lot::Mutex<Option<Dispatcher>>,
+    shard_transfer_dispatcher: parking_lot::Mutex<Option<ShardTransferDispatcher>>,
 }
 
 impl TableOfContent {
@@ -630,7 +630,7 @@ impl TableOfContent {
     }
 
     /// Insert dispatcher into table of contents for shard transfer.
-    pub fn with_shard_transfer_dispatcher(&self, dispatcher: Dispatcher) {
+    pub fn with_shard_transfer_dispatcher(&self, dispatcher: ShardTransferDispatcher) {
         self.shard_transfer_dispatcher.lock().replace(dispatcher);
     }
 }
--- a/lib/storage/src/content_manager/toc/mod.rs
+++ b/lib/storage/src/content_manager/toc/mod.rs
@@ -34,13 +34,13 @@ use tokio::sync::{Mutex, RwLock, RwLockReadGuard, Semaphore};
 use tonic::transport::Channel;
 use tonic::Status;
 
-use self::transfer::ShardTransferConsensus;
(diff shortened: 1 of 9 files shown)


lib/storage/src/content_manager/toc/mod.rs
```diff
@@ -71,10 +73,13 @@ pub struct TableOfContent {
     /// A lock to prevent concurrent collection creation.
     /// Effectively, this lock ensures that `create_collection` is called sequentially.
     collection_create_lock: Mutex<()>,
+    /// Dispatcher for shard transfer to access consensus.
+    shard_transfer_dispatcher: parking_lot::Mutex<Option<Dispatcher>>,
```
timvisee, line 77: `Mutex<Option<_>>` because we can only assign this after ToC is constructed.

src/main.rs
```diff
@@ -268,6 +268,8 @@ fn main() -> anyhow::Result<()> {
 
         dispatcher = dispatcher.with_consensus(consensus_state.clone());
 
+        toc_arc.with_shard_transfer_dispatcher(dispatcher.clone());
```
timvisee, line 271: After ToC is constructed, we create the dispatcher and can insert it here.

lib/storage/src/content_manager/toc/transfer.rs
```diff
@@ -0,0 +1,15 @@
...
+impl ShardTransferConsensus for Dispatcher {
+    fn consensus_commit_term(&self) -> CollectionResult<(u64, u64)> {
+        self.consensus_state()
+            .map(|consensus| {
+                let state = consensus.hard_state();
+                (state.commit, state.term)
+            })
+            .ok_or_else(|| CollectionError::service_error("Consensus is not available"))
+    }
+}
```
timvisee, line 15: The trait will eventually get specialized functions for operations we need on consensus. They'll be implemented here where we have access to the consensus proposal channel and consensus state through `Dispatcher`.

lib/storage/src/content_manager/toc/mod.rs
ffuugoo: Btw, I just realized, if we store `Dispatcher` in the `TableOfContent`, we have an `Arc` cycle and `TableOfContent` would never be dropped. (But it might already be so, though.) Can this be an issue? 🤔

timvisee: Good catch. I think you're right.

timvisee: I'll try to see if I can integrate `Weak<T>` somewhere to prevent this.

ffuugoo (commented):
LGTM. The only _potential_ issue I see is the `Arc` cycle.

lib/storage/src/content_manager/toc/mod.rs
ffuugoo: Would it be possible to store `Weak<TableOfContent>` and `ConsensusStateRef` inside `TableOfContent` and construct `Dispatcher` on-demand?

timvisee: > Would it be possible to store `Weak<TableOfContent>` and `ConsensusStateRef` inside `TableOfContent` and construct `Dispatcher` on-demand? (Seems even more hacky, though.)

Both can only be constructed after constructing ToC. So, both would require some `Mutex<Optional<_>>` fields in order to access them.

By using this 'dispatcher' they are grouped together.

lib/storage/src/content_manager/toc/transfer.rs
```diff
@@ -0,0 +1,28 @@
...
+use collection::shards::transfer::ShardTransferConsensus;
+
+use super::TableOfContent;
+use crate::content_manager::consensus_manager::ConsensusStateRef;
+
+#[derive(Clone)]
+pub struct ShardTransferDispatcher {
+    _toc: Weak<TableOfContent>,
+    consensus_state: ConsensusStateRef,
+}
```
timvisee, line 12: @ffuugoo We now have a custom type, holding a `Weak<T>` reference to prevent an `Arc` cycle.

One good thing is that this also removes the option from the fields.

It is currently still called *dispatcher, but we can decide to name it differently.

ffuugoo (approved):
LGTM 👌

PR #2874, opened 2023-10-25
Synchronize consensus across all nodes for shard snapshot transfer

timvisee:
Tracked in https://github.com/qdrant/qdrant/issues/2432.

Merges into https://github.com/qdrant/qdrant/pull/2467.

Adds a call to the shard snapshot transfer logic to synchronize all peers with the consensus state of the current node. It is one of the required synchronization steps for shard snapshot transfer.

More specifically, this synchronization will happen right before the shard snapshot transfer is finalized. We do that to make sure all nodes have at least reached the `Partial` state for the shard. That way we are sure that all nodes will include the shard for operations. This logic awaits on all nodes explicitly and falls back to a 10 second timeout on failure. This timeout matches the consensus timeout.

It also:
- moves methods for awaiting a consensus commit from ToC into the channel service, seems to be a better place
- fixes error handling for awaiting a consensus commit on all peers, some errors were ignored before
- fixes wait conditions for consensus, see comment below
- waits for replica set to reach partial state for the remote peer
- defines some TODOs in the shard snapshot transfer function to clearly define what's left to do in #2432 to be approved.

I've manually tested this locally.

A future PR will implement setting the shard state from `PartialSnapshot` to `Partial`.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/transfer/shard_transfer.rs
+++ b/lib/collection/src/shards/transfer/shard_transfer.rs
@@ -288,6 +288,11 @@ async fn await_consensus_sync(
     channel_service: &ChannelService,
     this_peer_id: PeerId,
 ) {
+    let peer_count = channel_service.id_to_address.read().len().saturating_sub(1);
+    if peer_count == 0 {
+        return;
+    }
+
     let sync_consensus = async {
         let await_result = consensus
             .await_consensus_sync(this_peer_id, channel_service)
@@ -300,7 +305,7 @@ async fn await_consensus_sync(
     let timeout = sleep(defaults::CONSENSUS_META_OP_WAIT);
 
     log::trace!(
-        "Waiting on all peers to reach consensus before finalizing shard snapshot transfer..."
+        "Waiting on {peer_count} peer(s) to reach consensus before finalizing shard snapshot transfer..."
     );
     tokio::select! {
         Ok(_) = sync_consensus => {
--- a/lib/collection/src/shards/transfer/shard_transfer.rs
+++ b/lib/collection/src/shards/transfer/shard_transfer.rs
@@ -9,7 +9,6 @@ use std::time::Duration;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use tokio::time::sleep;
-use url::Url;
 
 use super::ShardTransferConsensus;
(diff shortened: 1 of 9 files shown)


lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -609,20 +608,20 @@ impl<C: CollectionContainer> ConsensusManager<C> {
...
             let is_ok = state.term == term && state.commit >= commit;
             if is_ok {
-                return true;
+                return Ok(());
             }
 
-            // Fail if on a different term
-            let is_fail = state.term != term;
+            // Fail if on a newer term
+            let is_fail = state.term > term;
```
timvisee, line 615: I changed the condition:

```diff
-let is_fail = state.term != term;
+let is_fail = state.term > term;
```

This is on a remote node, on which we wait for it to reach some consensus (commit + term).

Before, waiting on a consensus commit always failed right away if the current term is not exactly the same as got requested.

After this change, it won't fail right away anymore if the current term has not caught up yet.  This gives the node some time to reach the same term if a reelection happened.

@ffuugoo Maybe you have some good insight on this specifically.

lib/collection/src/shards/transfer/shard_transfer.rs
```diff
@@ -246,31 +248,85 @@ async fn transfer_snapshot(
...
     }
 
-    // We must keep partial state for 10 seconds to allow all nodes to catch up
-    // Future improvement: instead of waiting 10 seconds, confirm all nodes reached consensus
-    log::trace!("Shard snapshot transfer is waiting 10 seconds for consensus to catch up");
-    sleep(Duration::from_secs(10)).await;
+    // Synchronize, make sure all nodes have reached consensus before continuing and unproxying
+    // All other nodes must consider the shard state to be at least partial so that all nodes send
+    // operations to it.
+    await_consensus_sync(consensus, &channel_service, transfer_config.from).await;
```
timvisee, line 284: The essential integration.

- wait for shard to be in `Partial` state
- synchronize consensus across all nodes
- once all nodes reach consensus, complete transfer (and set shard to `Active` outside this function)

generall (approved):

Issue #2880, opened 2023-10-26
Error when gRPC search batch points does not specify a collection name for each search


russcam:
Qdrant version: v1.6.1, docker container running on WSL

When attempting to call the search batch points API over gRPC with a client generated from the protos, an internal error occurs when
each search points in the batch does not specify a collection name. The error occurs here with a [C# client](https://github.com/russcam/qdrant-dotnet-client), but from memory, the same issue also occurs with a [Java client](https://github.com/metaloom/qdrant-java-client).

## Current Behavior

An error occurs with the following panic backtrace

```
2023-10-26 22:35:21 2023-10-26T12:35:21.637303Z ERROR qdrant::startup: Panic backtrace: 
2023-10-26 22:35:21    0: qdrant::startup::setup_panic_hook::{{closure}}
2023-10-26 22:35:21    1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
2023-10-26 22:35:21              at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/alloc/src/boxed.rs:2007:9
2023-10-26 22:35:21    2: std::panicking::rust_panic_with_hook
2023-10-26 22:35:21              at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/std/src/panicking.rs:709:13
2023-10-26 22:35:21    3: std::panicking::begin_panic::{{closure}}
2023-10-26 22:35:21    4: std::sys_common::backtrace::__rust_end_short_backtrace
2023-10-26 22:35:21    5: std::panicking::begin_panic
2023-10-26 22:35:21    6: validator::types::ValidationErrors::merge
2023-10-26 22:35:21    7: <alloc::vec::Vec<V> as api::grpc::validate::ValidateExt>::validate
2023-10-26 22:35:21    8: <qdrant::tonic::api::points_api::PointsService as api::grpc::qdrant::points_server::Points>::search_batch::{{closure}}
2023-10-26 22:35:21    9: <<api::grpc::qdrant::points_server::PointsServer<T> as tower_service::Service<http::request::Request<B>>>::call::SearchBatchSvc<T> as tonic::server::service::UnaryService<api::grpc::qdrant::SearchBatchPoints>>::call::{{closure}}
2023-10-26 22:35:21   10: <api::grpc::qdrant::points_server::PointsServer<T> as tower_service::Service<http::request::Request<B>>>::call::{{closure}}
[...]

PR #2888, opened 2023-10-29
Fix gRPC validation panic on list with multiple failing items

generall:
Fixes https://github.com/qdrant/qdrant/issues/2880.

Our custom validation handling on gRPC, done through `ValidateExt`, can panic when validating lists. This happens when there is more than one item having validation errors.

Sadly, the `validator` [type](https://docs.rs/validator/0.16.1/validator/struct.ValidationErrors.html) holding field errors is limited. We cannot use the item index at runtime because field names must be static string slices. Nor can we merge multiple errors into the same field name.

To mitigate panics and satisfy the limitation this PR now returns on the first error it comes across when validating list types in gRPC.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/api/src/grpc/validate.rs
+++ b/lib/api/src/grpc/validate.rs
@@ -42,16 +42,11 @@ where
     V: Validate,
 {
     #[inline]
-    #[allow(clippy::manual_try_fold)] // `try_fold` can't be used because it shortcuts on Err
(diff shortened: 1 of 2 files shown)


timvisee (approved):
Remark: I've also implemented the fix.

PR #2895, opened 2023-10-30
Add custom TLS certificate support for remote snapshot downloads to snapshot recover APIs

ffuugoo:
Required for #2432.

During snapshot shard transfer, the __recipient__ of the transfer downloads a shard snapshot from __sender__ of the transfer using shard snapshot HTTP API (because there's no standard support for file downloads in gRPC).

It is possible to enable HTTPS for HTTP API, but required TLS certificates are not _yet_ propagated to the HTTP client that is used to download shard snapshot, so snapshot shard transfer fails if HTTPS is enabled.

This PR fixes this by initializing HTTP client with all required TLS magic.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -111,6 +111,32 @@ jobs:
         run: ./test_tls.sh
         working-directory: ./tests/tls
 
+  test-tls-snapshot-shard-transfer:
+
+    name: "TLS: Snapshot Shard Transfer"
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Docker build
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          tags: qdrant_tls
+          cache-from: |
+            type=gha,scope=${{ github.ref }}
+            type=gha,scope=${{ github.base_ref }}
+          cache-to: type=gha,mode=max,scope=${{ github.ref }}
+          load: true
+          build-args: |
+            PROFILE=ci
+      - name: Run integration tests
+        run: ./test_tls_snapshot_shard_transfer.sh
+        working-directory: ./tests/tls
+
   test-consistency:
 
     runs-on: ubuntu-latest
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -89,6 +89,8 @@ jobs:
 
   test-tls-compose:
 
+    name: "Basic TLS/HTTPS tests"
+
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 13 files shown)


generall (approved):

src/common/http_client.rs
timvisee, line 1: This feels over-engineered to me.

I think I'd prefer constructing the reqwest client on each usage with a single `fn reqwest_from_settings()` function and passing settings along.  All places were we'd do this are cold and already have other file operations. It would significantly reduce the code required for this.

What do you think?

timvisee (changes requested):

src/common/http_client.rs
ffuugoo: Hmmm... I don't know, tbh. Reading and parsing PEM files and initializing new client on each request feels wrong.

ffuugoo:
I hate writing integration tests for such features. It always takes so much time and energy to automate a rather trivial check... 🥲

src/common/http_client.rs
timvisee: @generall What is your opinion on this?

generall: If it is only about reading certificate, it should be fine to do it every time. Downloading snapshot is not that time sensitive and we get benefit of: simpler code + no thinking about certificate reload.

ffuugoo: Somehow missed @generall reply yesterday. I'll fix this now. 👌

ffuugoo: Done.

tests/tls/test_tls_snapshot_shard_transfer.sh
```diff
@@ -0,0 +1,121 @@
...
+	(( POINTS > 0 ))
+}
+
+function curl {
+	declare URL="$1"
+	declare ARGS=( "${@:2}" )
+
+	declare STATUS=0
+
+	command curl -sS --fail-with-body \
```
timvisee, line 104: Rather than bypassing our function `curl` I'd prefer to rename our function to prevent this ambiguity.

src/common/http_client.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+fn https_client_identity(cert: &Path, key: &Path) -> Result<reqwest::tls::Identity> {
+    let mut identity_pem =
+        fs::read(cert).map_err(|err| Error::failed_to_read(err, "certificate", cert))?;
+
+    let mut key_file = fs::File::open(key).map_err(|err| Error::failed_to_read(err, "key", key))?;
+
+    io::copy(&mut key_file, &mut identity_pem)
+        .map_err(|err| Error::failed_to_read(err, "key", key))?;
```
timvisee, line 78: Why do we replace the identity bytes (cert) with the key material here?

tests/tls/test_tls_snapshot_shard_transfer.sh
timvisee: On second opinion, I think it was fine. We only used this inside the `curl` function itself.

src/common/http_client.rs
ffuugoo: We don't replace, we _append_.

ffuugoo: `reqwest::tls::Identity` takes a single PEM that should contain both the certificate and the key, and seems like PEM files can be simply concatenated.

timvisee: 🤦‍♂️ Thanks for elaborating

timvisee (approved):

.github/workflows/integration-tests.yml
```diff
@@ -111,6 +111,32 @@ jobs:
         run: ./test_tls.sh
         working-directory: ./tests/tls
 
+  test-tls-snapshot-shard-transfer:
```
generall, line 114: in a long run, can we really afford building custom CI task for individual tests?

ffuugoo: Should be fixed now.

PR #2900, opened 2023-10-31
SparseVectorIndex implements VectorIndex

agourlay:
This PR adds an implementation of `VectorIndex` for the sparse vector index.

Handles:
- deleted points
- query filter
- search telemetry

Includes new unit tests to make sure everything is working.

### DIFF
--- a/lib/segment/src/fixtures/payload_fixtures.rs
+++ b/lib/segment/src/fixtures/payload_fixtures.rs
@@ -4,6 +4,7 @@ use itertools::Itertools;
 use rand::seq::SliceRandom;
 use rand::Rng;
 use serde_json::{json, Value};
+use sparse::common::sparse_vector::SparseVector;
 
 use crate::data_types::vectors::VectorElementType;
 use crate::types::{
@@ -115,6 +116,43 @@ pub fn random_bool_payload<R: Rng + ?Sized>(
 
 pub fn random_vector<R: Rng + ?Sized>(rnd_gen: &mut R, size: usize) -> Vec<VectorElementType> {
     (0..size).map(|_| rnd_gen.gen()).collect()
+}
+
+/// Generates a non empty sparse vector
+pub fn random_sparse_vector<R: Rng + ?Sized>(rnd_gen: &mut R, max_size: usize) -> SparseVector {
+    let size = rnd_gen.gen_range(1..max_size);
+    let mut tuples: Vec<(i32, f64)> = vec![];
+
+    for i in 1..=size {
+        let no_skip = rnd_gen.gen_bool(0.5);
+        if no_skip {
+            tuples.push((i as i32, rnd_gen.gen_range(0.0..100.0)));
+        }
+    }
+
+    // make sure we have at least one vector
+    if tuples.is_empty() {
+        tuples.push((
+            rnd_gen.gen_range(1..max_size) as i32,
+            rnd_gen.gen_range(0.0..100.0),
+        ));
+    }
+
+    SparseVector::from(tuples)
+}
+
+/// Generates a sparse vector with all dimensions filled
+pub fn random_full_sparse_vector<R: Rng + ?Sized>(
+    rnd_gen: &mut R,
+    max_size: usize,
+) -> SparseVector {
+    let mut tuples: Vec<(i32, f64)> = vec![];
+
+    for i in 1..=max_size {
+        tuples.push((i as i32, rnd_gen.gen_range(0.0..100.0)));
+    }
+
+    SparseVector::from(tuples)
 }
 
 pub fn random_uncommon_condition<R: Rng + ?Sized>(rnd_gen: &mut R) -> Condition {
--- a/lib/segment/src/fixtures/payload_fixtures.rs
+++ b/lib/segment/src/fixtures/payload_fixtures.rs
@@ -4,7 +4,6 @@ use itertools::Itertools;
 use rand::seq::SliceRandom;
 use rand::Rng;
 use serde_json::{json, Value};
-use sparse::common::sparse_vector::SparseVector;
 
 use crate::data_types::vectors::VectorElementType;
 use crate::types::{
@@ -116,43 +115,6 @@ pub fn random_bool_payload<R: Rng + ?Sized>(
 
 pub fn random_vector<R: Rng + ?Sized>(rnd_gen: &mut R, size: usize) -> Vec<VectorElementType> {
     (0..size).map(|_| rnd_gen.gen()).collect()
-}
-
-/// Generates a non empty sparse vector
-pub fn random_sparse_vector<R: Rng + ?Sized>(rnd_gen: &mut R, max_size: usize) -> SparseVector {
-    let size = rnd_gen.gen_range(1..max_size);
-    let mut tuples: Vec<(i32, f64)> = vec![];
-
-    for i in 1..=size {
-        let no_skip = rnd_gen.gen_bool(0.5);
-        if no_skip {
-            tuples.push((i as i32, rnd_gen.gen_range(0.0..100.0)));
-        }
-    }
-
-    // make sure we have at least one vector
-    if tuples.is_empty() {
-        tuples.push((
-            rnd_gen.gen_range(1..max_size) as i32,
-            rnd_gen.gen_range(0.0..100.0),
-        ));
-    }
-
-    SparseVector::from(tuples)
-}
-
-/// Generates a sparse vector with all dimensions filled
-pub fn random_full_sparse_vector<R: Rng + ?Sized>(
-    rnd_gen: &mut R,
-    max_size: usize,
-) -> SparseVector {
-    let mut tuples: Vec<(i32, f64)> = vec![];
-
-    for i in 1..=max_size {
-        tuples.push((i as i32, rnd_gen.gen_range(0.0..100.0)));
-    }
-
-    SparseVector::from(tuples)
 }
 
 pub fn random_uncommon_condition<R: Rng + ?Sized>(rnd_gen: &mut R) -> Condition {
--- a/lib/segment/src/index/sparse_index/sparse_vector_index.rs
+++ b/lib/segment/src/index/sparse_index/sparse_vector_index.rs
@@ -64,11 +64,11 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         for vector in vectors {
             check_process_stopped(is_stopped)?;
             // measure time according to filter
-            if with_filter {
-                let _timer = ScopeDurationMeasurer::new(&self.searches_telemetry.filtered_sparse);
+            let _timer = if with_filter {
+                ScopeDurationMeasurer::new(&self.searches_telemetry.filtered_sparse)
             } else {
-                let _timer = ScopeDurationMeasurer::new(&self.searches_telemetry.unfiltered_sparse);
-            }
+                ScopeDurationMeasurer::new(&self.searches_telemetry.unfiltered_sparse)
+            };
             let vector = match vector {
                 QueryVector::Nearest(vector) => vector,
                 QueryVector::Recommend(_) => {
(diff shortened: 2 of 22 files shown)


lib/segment/src/telemetry.rs
```diff
@@ -61,10 +65,30 @@ pub struct VectorIndexSearchesTelemetry {
...
     pub filtered_exact: OperationDurationStatistics,
 
+    #[serde(skip_serializing_if = "OperationDurationStatistics::is_empty")]
+    pub filtered_sparse: OperationDurationStatistics,
+
     #[serde(skip_serializing_if = "OperationDurationStatistics::is_empty")]
     pub unfiltered_exact: OperationDurationStatistics,
 }
 
+impl From<&SparseSearchesTelemetry> for VectorIndexSearchesTelemetry {
```
IvanPleshkov, line 75: I propose to move this convertion into `sparse_search_telemetry.rs` file

lib/sparse/src/index/inverted_index/mod.rs
```diff
@@ -15,4 +20,28 @@ pub trait InvertedIndex {
...
+    /// Create inverted index from ram index
+    fn from_ram_index<P: AsRef<Path>>(
+        ram_index: InvertedIndexRam,
+        path: P,
+    ) -> std::io::Result<Self>
+    where
+        Self: Sized;
+
+    /// Returns the maximum number of results that can be returned by the index for a given sparse vector
+    fn max_result_count(&self, query_vector: &SparseVector) -> usize {
```
IvanPleshkov, line 36: I propose to move this function out of this trait. Because this function is required for test only and not required in vector index

lib/sparse/src/index/inverted_index/inverted_index_ram.rs
```diff
@@ -25,7 +25,24 @@ impl InvertedIndex for InvertedIndexRam {
     }
 
     fn indexed_vector_count(&self) -> usize {
-        self.postings.len()
+        // TODO(sparse) not super efficient
```
IvanPleshkov, line 28: This calculation is not efficient. But we can resolve it.
In `hnsw.rs` we can read how was implemented indexed counting. It counts existing points only in `build` method and does not change the saved value. We can do the same, in `SparseVectorIndex` precalculate indexed count in `build` method and reuse it

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -26,28 +35,150 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
+        let borrowed_vector_storage = self.vector_storage.borrow();
+        let borrowed_id_tracker = self.id_tracker.borrow();
+        let points_count = borrowed_vector_storage.total_vector_count();
+        let mut ram_index = InvertedIndexRam::empty();
+        for id in 0..points_count as PointOffsetType {
+            // do not add deleted points to the index
+            if borrowed_id_tracker.is_deleted_point(id) {
+                continue;
+            }
+            let vector: &SparseVector = borrowed_vector_storage.get_vector(id).try_into()?;
```
IvanPleshkov, line 157: We have to check if `id` is deleted in `borrowed_vector_storage`

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -26,28 +35,150 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
+            }
+        }
+    }
+
+    fn build_index(&mut self, _stopped: &AtomicBool) -> OperationResult<()> {
+        let borrowed_vector_storage = self.vector_storage.borrow();
+        let borrowed_id_tracker = self.id_tracker.borrow();
+        let points_count = borrowed_vector_storage.total_vector_count();
+        let mut ram_index = InvertedIndexRam::empty();
+        for id in 0..points_count as PointOffsetType {
```
IvanPleshkov, line 152: I propose to use ready iterator with filter of deleted flags:
```
for id in  id_tracker.iter_ids_excluding(deleted_bitslice) {
```
where `deleted_bitslice` is deleted flags from vector_storage

IvanPleshkov (changes requested):

lib/segment/src/fixtures/payload_fixtures.rs
```diff
@@ -117,6 +118,43 @@ pub fn random_vector<R: Rng + ?Sized>(rnd_gen: &mut R, size: usize) -> Vec<Vecto
     (0..size).map(|_| rnd_gen.gen()).collect()
 }
 
+/// Generates a non empty sparse vector
+pub fn random_sparse_vector<R: Rng + ?Sized>(rnd_gen: &mut R, max_size: usize) -> SparseVector {
```
IvanPleshkov, line 122: `random_sparse_vector` is already presented in `dev` branch

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -26,28 +35,150 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         id_tracker: Arc<AtomicRefCell<IdTrackerSS>>,
         vector_storage: Arc<AtomicRefCell<VectorStorageEnum>>,
         payload_index: Arc<AtomicRefCell<StructPayloadIndex>>,
-        path: PathBuf,
+        path: &Path, // TODO(sparse) use path to load/save index
```
IvanPleshkov, line 38: I propose to rebuild index while creation so that we have a persistency:
```
    pub fn open(
        id_tracker: Arc<AtomicRefCell<IdTrackerSS>>,
        payload_index: Arc<AtomicRefCell<StructPayloadIndex>>,
        path: &Path, // TODO(sparse) use path to load/save index
    ) -> OperationResult<Self> {
        let searches_telemetry = SparseSearchesTelemetry::new();
        let path = path.to_path_buf();
        let s = Self {
            id_tracker,
            vector_storage,
            payload_index,
            path,
            inverted_index: Default::default(),
            searches_telemetry,
            indexed_count: 0,
        };
        s.build_index(false)?; // TODO(sparse) load from disk
        Ok(s)
    }
```

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -26,28 +35,150 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
-        search_context.search()
+        with_filter: bool,
+        condition: impl Fn(PointOffsetType) -> bool,
+    ) -> OperationResult<Vec<Vec<ScoredPointOffset>>> {
+        let mut result = vec![];
+
+        for vector in vectors {
+            check_process_stopped(is_stopped)?;
+            // measure time according to filter
+            if with_filter {
```
IvanPleshkov, line 67: `_timer` has drop with timer stopping. we have to move `_timer` to higher scope:
```
            let _timer = if with_filter {
                ScopeDurationMeasurer::new(&self.searches_telemetry.filtered_sparse)
            } else {
                ScopeDurationMeasurer::new(&self.searches_telemetry.unfiltered_sparse)
            };
```

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -26,28 +35,150 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
+                        description: "Cannot discovery sparse vectors".to_string(),
+                    })
+                }
+                QueryVector::Context(_) => {
+                    return Err(OperationError::ValidationError {
+                        description: "Cannot context query sparse vectors".to_string(),
+                    })
+                }
+            };
+            let sparse_vector: &SparseVector = vector.to_vec_ref().try_into()?;
```
IvanPleshkov, line 90: Do we have a `TryFrom` from `Vector` to `SparseVector`? We can convert shorter:
```
let sparse_vector: SparseVector = vector.try_into()?;
```

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -26,28 +35,150 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
+        _params: Option<&SearchParams>,
+        is_stopped: &AtomicBool,
+    ) -> OperationResult<Vec<Vec<ScoredPointOffset>>> {
+        let id_tracker = self.id_tracker.borrow();
+        let vector_storage = self.vector_storage.borrow();
+        let deleted_point_bitslice = id_tracker.deleted_point_bitslice();
+        let deleted_vectors = vector_storage.deleted_vector_bitslice();
+        // filter for deleted points
+        let not_deleted_condition = |idx: PointOffsetType| -> bool {
+            !deleted_point_bitslice
```
IvanPleshkov, line 120: This is a copypaste from raw scorer. We have to do some abstraction with this check of deleted vector. Or mark as `TODO(sparse)` =)

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -26,28 +35,150 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
+                })
+            }
+            None => {
+                // query sparse index directly
+                self.search_sparse(vectors, top, is_stopped, false, not_deleted_condition)
+            }
+        }
+    }
+
+    fn build_index(&mut self, _stopped: &AtomicBool) -> OperationResult<()> {
```
IvanPleshkov, line 147: We can use stop condition already and place stop check in `for` loop:
```
check_process_stopped(stopped)?;
```

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -26,28 +35,150 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
+        for id in 0..points_count as PointOffsetType {
+            // do not add deleted points to the index
+            if borrowed_id_tracker.is_deleted_point(id) {
+                continue;
+            }
+            let vector: &SparseVector = borrowed_vector_storage.get_vector(id).try_into()?;
+            ram_index.upsert(id, vector.to_owned());
+        }
+
+        self.inverted_index = TInvertedIndex::from_ram_index(ram_index, &self.path)?;
```
IvanPleshkov, line 161: There is OOM situation. For reindexing of large disk inverted index we have to load all structure into RAM and get OOM. We need to solve this problem as separate issue and I propose to place `TODO` at least

agourlay: any ideas how to actually tackle that?

lib/sparse/src/index/inverted_index/mod.rs
agourlay: You are right, it does not belong in this trait :+1: 
I decided to move in in the sparse index for simplicity https://github.com/qdrant/qdrant/pull/2900/commits/50757440162bd15b006e15c0c9592de4871154c9

lib/segment/src/fixtures/payload_fixtures.rs
agourlay: good observation :+1: 

decided to unify the fixtures into the sparse crate https://github.com/qdrant/qdrant/pull/2900/commits/3552f08324797a9d7cc71ebc934b3380106aafcb

lib/segment/src/index/sparse_index/sparse_vector_index.rs
agourlay: https://github.com/qdrant/qdrant/pull/2900/commits/79b5de9b0f4f049ef6dedea406a979b0033d0157

lib/segment/src/index/sparse_index/sparse_vector_index.rs
agourlay: I decided to both reuse the implementation and add a TODO.
This is a bit of a hack but will make sure we don't forget to update the implementation.

A wider refactoring is necessary to remove the need for a dedicated implementation for `SparseRawScorer`.
cc @coszio :)

lib/segment/src/index/sparse_index/sparse_vector_index.rs
agourlay: I went to extra mile and added open to the inverted index trait to load properly mmaped index
https://github.com/qdrant/qdrant/pull/2900/commits/ff69e3ad9969cb03defefc987ae71f9bcb9617aa

I will add in the future unit test for using the `SparseVectorIndex` is a mmaped inverted index.

agourlay:
@IvanPleshkov Thank you for the thorough review :+1: 

I believe I have addressed all your comments, please check it out again!

IvanPleshkov (approved):

PR #2902, opened 2023-10-31
Test snapshot transfer during updates

generall:
tracking: https://github.com/qdrant/qdrant/issues/2432

- Integration test for snapshot-based shard transfer
- Test for transfer cancellation

### DIFF
--- a/tests/consensus_tests/test_shard_snapshot_transfer.py
+++ b/tests/consensus_tests/test_shard_snapshot_transfer.py
@@ -1,0 +1,95 @@
+import multiprocessing
+import pathlib
+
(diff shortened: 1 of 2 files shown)


timvisee:
All cancellation logic has been merged in <https://github.com/qdrant/qdrant/pull/2926>.

In d5f0b15e098194fa41b2e11b64554cf95bda19c5, I've throttled point insertions slightly. Otherwise the queue proxy transfer couldn't keep up with insertions. In a real scenario we expect insertions to slow down eventually, after which the queue proxy can catch up. That doesn't happen in this test however, and we also have a hard timeout of 30 seconds.

timvisee (approved):

PR #2771, opened 2023-10-06
Refactor `with_channel_timeout`

coszio:
**Note**: This should only affect the inter-node channels.

- Removes the intermediate timeout implementation that was working separately from the request and channel timeouts.
- Now the channel timeout is set to a hard 1 hour (not configurable)
- All the individual requests now pass through an interceptor, which checks if they already have a timeout set, and if not, they set it to the `grpc_timeout_ms` from settings.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #2903, opened 2023-10-31
Use long timeout for shard snapshot recovery process

timvisee:
Depends on <https://github.com/qdrant/qdrant/pull/2771>.

Before, the shard snapshot transfer recovery process could only take 60 seconds because a global gRPC timeout enforced this.

This change allows the snapshot transferring and recovery process to take much longer. Now it allows to take up to 24 hours useful for very large snapshots with slow connections.

This number is arbitrary, though it would allow transferring 100GB over a 10Mbps link (taking ~22 hours).

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/transfer/shard_transfer.rs
+++ b/lib/collection/src/shards/transfer/shard_transfer.rs
@@ -6,6 +6,7 @@ use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 use std::time::Duration;
 
+use api::grpc::transport_channel_pool::MAX_GRPC_CHANNEL_TIMEOUT;
 use common::defaults;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


lib/collection/src/shards/transfer/shard_transfer.rs
```diff
@@ -316,6 +320,8 @@ async fn transfer_snapshot(
             shard_id,
             &shard_download_url,
             SnapshotPriority::ShardTransfer,
+            Some(SNAPSHOT_TRANSFER_RECOVER_TIMEOUT),
+            api::grpc::transport_channel_pool::DEFAULT_RETRIES,
```
timvisee, line 324: The timeout for the transfer/recovery process is now explicitly configured here as part of the logic driving the shard snapshot transfer.

ffuugoo (approved):

lib/collection/src/shards/transfer/shard_transfer.rs
coszio: Why do we need to set it explicitly? Will it be configurable in the future? else I think we could avoid parameter drilling if they just depend on constants anyways

timvisee: Good point. I may have pulled them up a bit too far here.

I'll downgrade the timeout into `recover_shard_snapshot_from_url`, as I don't think it makes sense to move it further down into `with_shard_snapshot_client`. I've also removed the retries parameter.

See 44cbd9142c4ff567ff44375bb36bbf04f8a81223.

generall (approved):

PR #2904, opened 2023-10-31
Fix recommend `best_score` formula

coszio:
Reported in Discord:
https://discord.com/channels/907569970500743200/1168571201828946022

In short, using only one negative example would yield the same results as using only the same point but as a positive example when dealing with euclidean distance.

Why was this?
Euclidean distances are negated during search, so that the logic for keeping best candidates always considers higher is better for scores. For positive examples, our formula is basically pass-through, since it just chooses the similarity score without extra modifications. However, when choosing a similarity from a negative example, the choice was to square it and negate it, e.g. `-(max_negative * max_negative)`.
With this report we discovered that this formula fails to invert the order of candidates when their similarities are negative. 

Now, after some discussion, the formula was decided to be:
```rust
if max_positive > max_negative {
     scaled_fast_sigmoid(max_positive)
} else {
     -scaled_fast_sigmoid(max_negative)
}
```
such that it fulfills these 3 properties:
1. Choosing two points from positives preserves their similarity order
2. Choosing two points from negatives inverts their similarity order
3. When choosing one point from positive and one point from negative, the former is preferred over the latter

## Changes

- Changes RecommendBestScore formula to the above formula
- Introduces checks for testing the mentioned properties

## Caveats

- Previously, similarities to positive examples were unchanged, so scores would be the same between using search vs recommend with `best_score` strategy. Now similarities are still taken with consideration of chosen distance metric, but scores will not display the distance. This is a **BREAKING CHANGE** if the applications used the positive scoring afterwards.
- For the same reason `score_threshold` would also need to be adjusted appropriately. For reference, ranges will always be (-1, 0) for negative scoring, and (0, 1) for positive scoring. 

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/local_shard_operations.rs
+++ b/lib/collection/src/shards/local_shard_operations.rs
@@ -78,7 +78,9 @@ impl LocalShard {
                             scored_point.score = distance.postprocess_score(scored_point.score);
                         }
                         // Don't post-process if we are dealing with custom scoring
-                        QueryEnum::RecommendBestScore(_) => {}
+                        QueryEnum::RecommendBestScore(_)
+                        | QueryEnum::Discover(_)
+                        | QueryEnum::Context(_) => {}
                     };
                     scored_point
                 });
--- a/lib/collection/src/shards/local_shard_operations.rs
+++ b/lib/collection/src/shards/local_shard_operations.rs
@@ -13,7 +13,7 @@ use crate::collection_manager::segments_searcher::SegmentsSearcher;
 use crate::common::stopping_guard::StoppingGuard;
 use crate::operations::types::{
     CollectionError, CollectionInfo, CollectionResult, CoreSearchRequestBatch, CountRequest,
-    CountResult, PointRequest, Record, SearchRequestBatch, UpdateResult, UpdateStatus,
+    CountResult, PointRequest, QueryEnum, Record, SearchRequestBatch, UpdateResult, UpdateStatus,
 };
 use crate::operations::CollectionUpdateOperations;
 use crate::optimizers_builder::DEFAULT_INDEXING_THRESHOLD_KB;
@@ -73,6 +73,12 @@ impl LocalShard {
                     .unwrap()
                     .distance;
                 let processed_res = vector_res.into_iter().map(|mut scored_point| {
+                    // Don't post-process if we are dealing with custom scoring
+                    // TODO(luis): add discovery variant here too
+                    if matches!(req.query, QueryEnum::RecommendBestScore(_)) {
+                        return scored_point;
+                    }
+
                     scored_point.score = distance.postprocess_score(scored_point.score);
                     scored_point
                 });
--- a/lib/collection/src/shards/local_shard_operations.rs
+++ b/lib/collection/src/shards/local_shard_operations.rs
@@ -73,13 +73,13 @@ impl LocalShard {
                     .unwrap()
                     .distance;
                 let processed_res = vector_res.into_iter().map(|mut scored_point| {
-                    // Don't post-process if we are dealing with custom scoring
-                    // TODO(luis): add discovery variant here too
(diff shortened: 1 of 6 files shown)


lib/collection/src/shards/local_shard_operations.rs
```diff
@@ -73,6 +73,12 @@ impl LocalShard {
                     .unwrap()
                     .distance;
                 let processed_res = vector_res.into_iter().map(|mut scored_point| {
+                    // Don't post-process if we are dealing with custom scoring
+                    // TODO(luis): add discovery variant here too
+                    if matches!(req.query, QueryEnum::RecommendBestScore(_)) {
+                        return scored_point;
+                    }
+
```
timvisee, line 81: What about doing an exhaustive match here, so that we don't require a TODO?

lib/segment/src/vector_storage/query/reco_query.rs
```diff
@@ -87,33 +88,101 @@ impl From<RecoQuery<VectorType>> for RecoQuery<Vector> {
...
+        #[test]
+        fn correct_positive_and_negative_order(p in -100f32..=100f32, n in -100f32..=100f32) {
+            let dummy_similarity = |x: &f32| *x as ScoreType;
+
+            let query_p = RecoQuery::new(vec![p], vec![]);
+            let query_n = RecoQuery::new(vec![], vec![n]);
+
+            let ordering = query_p.score_by(dummy_similarity).total_cmp(&query_n.score_by(dummy_similarity));
+
+            assert!(ordering == std::cmp::Ordering::Equal || ordering == std::cmp::Ordering::Greater);
```
timvisee, line 185: Or are you doing this on purpose to make it more explicit?

```suggestion
            assert_ne!(ordering, std::cmp::Ordering::Less);
```

lib/segment/src/vector_storage/query/reco_query.rs
```diff
@@ -87,33 +88,101 @@ impl From<RecoQuery<VectorType>> for RecoQuery<Vector> {
...
+
+            let query_a = RecoQuery::new(vec![], vec![a]);
+            let query_b = RecoQuery::new(vec![], vec![b]);
+
+            let negative_ordering = query_a.score_by(dummy_similarity).total_cmp(&query_b.score_by(dummy_similarity));
+
+            if positive_ordering == std::cmp::Ordering::Equal {
+                assert_eq!(positive_ordering, negative_ordering);
+            } else {
+                assert!(positive_ordering != negative_ordering)
```
timvisee, line 154:
```suggestion
                assert_ne!(positive_ordering, negative_ordering)
```

lib/segment/src/vector_storage/tests/custom_query_scorer_equivalency.rs
```diff
@@ -268,8 +272,8 @@ fn scoring_equivalency(
             // both calculations are done on raw vectors, so score should be exactly the same
             assert_eq!(
                 raw_scores, other_scores,
-                "Scorer results are not equal, attempt: {}",
-                _i
+                "Scorer results are not equal, attempt: {}, query: {:?}",
+                _i, query
```
timvisee, line 276: `_i` -> `i`

timvisee (approved):

timvisee (commented):
I've done a static analysis.

lib/segment/src/vector_storage/query/reco_query.rs
coszio: Yours looks more concise 👍

lib/collection/src/shards/local_shard_operations.rs
coszio: Much better

coszio:
@timvisee thanks for the comments, I have included them now. Want to do another pass?

kacperlukawski (approved):
I like the idea. I'll have to update the article with the new formula, though.

PR #2907, opened 2023-11-01
Simple cancellation in shard snapshot transfer

timvisee:

Adds simple cancellation checks between each shard snapshot transfer step. This allows stopping half way through the shard transfer.

For this I've used [`CancellationToken`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) rather than an atomic boolean. It is similar to <https://github.com/qdrant/qdrant/pull/2870>, which seems to be the way forward.

I've also refactored `StoppableAsyncTaskHandle` into `CancellableAsyncTaskHandle` to make use of `CancellationToken`.

This PR implements cancellation checking in just a basic way. The <ins>create snapshot</ins>, <ins>transfer/recover snapshot</ins> and <ins>transfer queue</ins> operations are currently not canceled in the middle by itself. Instead, each must run to completion before cancellation is checked again. This will be improved upon in a later PR, once <https://github.com/qdrant/qdrant/pull/2870> is merged.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/common/stoppable_task_async.rs
+++ b/lib/collection/src/common/stoppable_task_async.rs
@@ -1,15 +1,16 @@
 use std::future::Future;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, Weak};
+use std::sync::Arc;
 
 use parking_lot::Mutex;
 use tokio::task::JoinHandle;
+use tokio_util::sync::CancellationToken;
 
 pub struct StoppableAsyncTaskHandle<T: Clone> {
     pub join_handle: JoinHandle<T>,
     result_holder: Arc<Mutex<Option<T>>>,
     finished: Arc<AtomicBool>,
-    stopped: Weak<AtomicBool>,
+    stopped: CancellationToken,
 }
 
 impl<T: Clone> StoppableAsyncTaskHandle<T> {
@@ -18,9 +19,7 @@ impl<T: Clone> StoppableAsyncTaskHandle<T> {
     }
 
     pub fn ask_to_stop(&self) {
-        if let Some(v) = self.stopped.upgrade() {
-            v.store(true, Ordering::Relaxed);
-        }
+        self.stopped.cancel();
     }
 
     pub fn stop(self) -> JoinHandle<T> {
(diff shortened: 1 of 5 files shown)


lib/collection/src/common/stoppable_task_async.rs
```diff
@@ -33,38 +32,35 @@ impl<T: Clone> StoppableAsyncTaskHandle<T> {
...
+            async move {
+                let res = f(cancel).await;
+                let mut result_holder_w = result_holder.lock();
+                result_holder_w.replace(res.clone());
+
+                // We use `Release` ordering to ensure that `f` won't be moved after the `store`
+                // by the compiler
+                finished.store(true, Ordering::Release);
+                res
+            }
```
ffuugoo, line 59: TBH, I don't understand, why we bother with `finished` and `result_holder` at this point. `JoinHandle` already provides both of these.

- we can use [`JoinHandle::is_finished`] instead of `finished`
- and we can use [`JoinHandle.await`] instead of `result_holder`
  - it's trivial to do a `try_ready` check with a `Future`, if we don't want to do a full `.await`

[`JoinHandle::is_finished`]: https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#method.is_finished
[`JoinHandle.await`]: https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#impl-Future-for-JoinHandle%3CT%3E

ffuugoo (approved):
LGTM

I think, we can do deeper refactoring after we merge both #2870 and #2907 (this one)

lib/collection/src/common/stoppable_task_async.rs
timvisee: Thank you for diving into this. I'm not sure why this was implemented this way, maybe we have a good reason for it. `git blame` shows that Andrey knows more about this.

In this PR I simply left it the way we had but ported the atomic boolean to the cancellation token.

Let's dive into t his in another PR.

timvisee:
> I think, we can do deeper refactoring after we merge both #2870 and #2907 (this one)

:raised_hands:

lib/collection/src/common/stoppable_task_async.rs
timvisee: Quick note, I don't think our `finish` is the same as `is_finished`. Our finish will not be `true` if we abort the task, while `is_finished()` will be `true` eventually.

PR #2908, opened 2023-11-01
Document and test underfetching with sparse vectors

agourlay:
While testing the index integration in https://github.com/qdrant/qdrant/pull/2900 I realized that the sparse search can under fetch and that it is fine.

By under-fetch I mean not retrieving all possible points in the index for large values of `top`.

This happens because only the indices of the sparse query vectors are considered during the search.
Therefore, none of the points **without an intersection** of dimensions with the query vector will ever be retrieved.

This PR makes this behavior clearer by adding two unit tests for the underlying cursor manipulation infrastructure.

### DIFF
--- a/lib/sparse/src/index/search_context.rs
+++ b/lib/sparse/src/index/search_context.rs
@@ -196,9 +196,14 @@ impl<'a> SearchContext<'a> {
 
 #[cfg(test)]
 mod tests {
+    use rand::rngs::StdRng;
+    use rand::{Rng, SeedableRng};
+
     use super::*;
     use crate::index::inverted_index::inverted_index_mmap::InvertedIndexMmap;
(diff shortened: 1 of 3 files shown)


lib/sparse/src/index/search_context.rs
```diff
@@ -598,4 +603,107 @@ mod tests {
...
+            SparseVector {
+                indices: vec![1, 2, 3],
+                values: vec![1.0, 1.0, 1.0],
+            },
+            3,
+            &inverted_index_ram,
+            &is_stopped,
+        );
+
+        let mut all_next_min_observed = vec![];
```
IvanPleshkov, line 662: I propose to use `HashSet` instead of `Vec`. With set it's guaranteed that all ids are unique and check for `len()` means iterating over all points without duplications

IvanPleshkov (approved):

PR #2915, opened 2023-11-02
Raw scorer for sparse vectors

agourlay:
This PR introduces a `RawScorer` for sparse vector so that we can fallback to payload index scan in case of low filter cardinality estimation.

### DIFF
--- a/lib/segment/src/data_types/vectors.rs
+++ b/lib/segment/src/data_types/vectors.rs
@@ -81,6 +81,17 @@ impl From<Vector> for VectorType {
     }
 }
 
+impl TryFrom<Vector> for SparseVector {
+    type Error = OperationError;
+
+    fn try_from(value: Vector) -> Result<Self, Self::Error> {
+        match value {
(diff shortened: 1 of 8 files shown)


IvanPleshkov (approved):

coszio:
Hey, @agourlay @IvanPleshkov, I arrived a little too late to the review party 😞, but I think there is a lot of copy-pasting here that could be avoided by creating a new `QueryScorer` instead 💡. 

From what I see, it all boils down to the [`score` and `score_internal`](https://github.com/qdrant/qdrant/blob/8a05462c6da510649dd75d3b59bdb6faee9a611d/lib/segment/src/vector_storage/query_scorer/mod.rs#L9-L11) functions from the `QueryScorer` trait, which should be quite simple to implement for sparse vectors, see the [`MetricQueryScorer` implementation](https://github.com/qdrant/qdrant/blob/8a05462c6da510649dd75d3b59bdb6faee9a611d/lib/segment/src/vector_storage/query_scorer/metric_query_scorer.rs#L31-L39) as example. 

Did you consider that before?

PR #2918, opened 2023-11-02
Support cancellation in queue proxy to forward proxy method

timvisee:

Depends on <https://github.com/qdrant/qdrant/pull/2907>.

In the shard snapshot transfer logic we use `queue_proxy_into_forward_proxy()` method to safely transform one shard type into the other. This method could have a long runtime, for example, when a lot of updates have to be transferred to a remote.

This PR adds support for cancelling this process early through a `CancellationToken`. Not removing the queue proxy here on cancellation is fine because our shard transfer retry logic does this for us on (cancellation) failure.

This also defines cancellation safety in some function comments.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -1,8 +1,6 @@
 use std::path::Path;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
-#[cfg(debug_assertions)]
-use std::thread;
 use std::time::Duration;
 
 use async_trait::async_trait;
@@ -269,7 +267,7 @@ impl ShardOperation for QueueProxyShard {
 #[cfg(debug_assertions)]
 impl Drop for QueueProxyShard {
     fn drop(&mut self) {
-        if !self.is_finalized() && !thread::panicking() {
+        if !self.is_finalized() && !std::thread::panicking() {
             panic!("To drop a queue proxy shard, finalize() must be used");
         }
     }
@@ -376,10 +374,10 @@ impl Inner {
 
         // Transfer batch with retries and store last transferred ID
         let last_idx = batch.last().map(|(idx, _)| *idx);
-        for attempts in (0..BATCH_RETRIES).rev() {
+        for remaining_attempts in (0..BATCH_RETRIES).rev() {
             match transfer_operations_batch(&batch, &self.remote_shard).await {
                 Ok(()) => {}
-                Err(err) if attempts > 0 => {
+                Err(err) if remaining_attempts > 0 => {
                     log::error!(
                         "Failed to transfer batch of updates to peer {}, retrying: {err}",
                         self.remote_shard.peer_id,
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -133,7 +133,7 @@ impl QueueProxyShard {
     ///
     /// # Cancel safety
     ///
-    /// This method is *not* cancel safe.
+    /// This method is *not* cancel safe, it consumes `self`.
     ///
     /// If `cancel` is triggered - finalization may not actually complete in which case an error is
     /// returned. None, some or all operations may be transmitted to the remote.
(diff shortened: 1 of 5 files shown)


lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -120,11 +129,24 @@ impl QueueProxyShard {
...
-            Ok(_) => Ok(self.forget_updates_and_finalize().await),
-            Err(err) => Err((err, self)),
+        match cancel_on_token(cancel, self.transfer_all_missed_updates()).await {
+            Ok(Ok(_)) => Ok(self.forget_updates_and_finalize().await),
+            // Transmission error
+            Ok(Err(err)) => Err((err, self)),
+            // Cancellation error
+            Err(err) => Err((err.into(), self)),
         }
     }
```
ffuugoo, line 151: This doesn't seem right. What's the difference if we explicitly cancel at `transfer_all_missed_updates` (e.g., with the `cancel_on_token`) or simply drop the whole `finalize` future? Either way, `forget_updates_and_finalize` may not be called.

IMO, we don't need to propagate `cancel` here, but I'm not yet sure if we should treat this as cancel-safe or not.

We call `revert_queue_proxy_local` explicitly if shard transfer task fails (which would call `forget_updates_and_finalize`), though.

timvisee: The issue is that `forget_updates_and_finalize()` is not cancel safe. That means we cannot simply drop this function.

The usual fix for this would be to spawn `forget_updates_and_finalize()` in a task (possibly through `spawn_cancel_on_drop`). But this doesn't seem possible due to ownership/lifetime issues.

ffuugoo: Disregard this.

`finalize` and `forget_updates_and_finalize` are not cancel safe, cause they consume `QueueProxyShard`.

But:
- we *have to* call `forget_updates_and_finalize` even if `transfer_all_missed_updates` returns an error
  - because `finalize` consumes `QueueProxyShard`
  - so there won't be `QueueProxyShard`
  - so we won't call `set_max_ack_version(None)` on error
    - *probably*, I didn't check the whole call stack yet
    - but it seems reasonable to handle it here anyway
- if `forget_updates_and_finalize` is handled this way, we can use `cancel_on_token` with `transfer_all_missed_updates`

timvisee: On second thought, dropping `forget_updates_and_finalize()` may actually be safe. The only issue is that we're not sure if we actually end up setting the max acknowledged version.

_This stuff is hard..._

ffuugoo: Yeah, it's *mostly* safe, with the only caveat of `set_max_ack_version`.

My original point was:
- if we *do* care about `set_max_ack_version`, then we *can't* simply use `cancel_on_token(cancel, self.transfer_all_missed_updates())`
  - because if we trigger `cancel`, we will exit before calling `set_max_ack_version`
  - and `finalize` consumes `self`, so it won't be available to the caller (to retry/explicitly handle `set_max_ack_version`) after we return `Err(Cancelled)`
- and if we *don't* care about `set_max_ack_version`
  - (e.g., we assume that it is not critical, or it should be handled explicitly somewhere else)
  - then we can simply mark `finalize` as cancel safe, and don't bother with `cancel` token

ffuugoo: Ok, I'm silly. I noticed that we return `self` on error. I guess it's fine as-is then. 🤦‍♀️

timvisee: > we _have to_ call `forget_updates_and_finalize` even if `transfer_all_missed_updates` returns an error

Not needed. Ownership is given back on error or on cancellation.

> if we do care about set_max_ack_version, then we can't simply use cancel_on_token(cancel, self.transfer_all_missed_updates())

Not necessarily. If the finalize call is cancelled, finalization will be incomplete. The user gets ownership back and has to finalize again.

---

So I believe the current implementation is fine. Or is it?

timvisee: @ffuugoo I think we concluded it's fine like this, so I'll resolve this comment. Please reopen it something is still outstanding.

timvisee:
@ffuugoo Requested re-review to ask whether we can merge into `dev`.

ffuugoo (approved):
LGTM

PR #2919, opened 2023-11-02
Add `error_if_cancelled` to `CancellationToken` for easy early returns

timvisee:
Merges into <https://github.com/qdrant/qdrant/pull/2918>.

This adds `error_if_cancelled()` to the `CancellationToken` to easily return early with a cancellation error if triggered.

```rust
let cancel = CancellationToken::new();

cancel.error_if_cancelled()?;
```

@ffuugoo What do you think? Maybe you have a better idea of doing this, or maybe it conflicts with your current work on this branch.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/common/cancel/src/lib.rs
+++ b/lib/common/cancel/src/lib.rs
@@ -11,3 +11,19 @@ pub enum Error {
     #[error("task was cancelled")]
     Cancelled,
 }
+
+pub trait CancelError {
+    /// Return an error if cancelled.
+    #[must_use = "a returned error must be propagated down function calls"]
+    fn error_if_cancelled(&self) -> Result<(), Error>;
+}
+
+impl CancelError for CancellationToken {
+    fn error_if_cancelled(&self) -> Result<(), Error> {
+        if self.is_cancelled() {
+            Err(Error::Cancelled)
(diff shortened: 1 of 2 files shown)


lib/common/cancel/src/lib.rs
```diff
@@ -11,3 +11,19 @@ pub enum Error {
     #[error("task was cancelled")]
     Cancelled,
 }
+
+pub trait CancelError {
+    /// Return an error if cancelled.
+    #[must_use = "a returned error must be propagated down function calls"]
+    fn error_if_cancelled(&self) -> Result<(), Error>;
```
coszio, line 18: How about 
```suggestion
    fn try_ok(&self) -> Result<(), Error>;
```
?

timvisee: I like that better, thanks!

timvisee:
We likely won't end up using this because other changes remove the need for the `error_if_cancelled()` check in shard snapshot transfer.

timvisee:
We won't be needing this anymore with changes in <https://github.com/qdrant/qdrant/pull/2926> which are about to get merged.

PR #2925, opened 2023-11-03
Extra tests for Sparse Vector Index

agourlay:
This PR adds extra tests for the `SparseVectorIndex` to assert that it can work with an underlying memmap inverted index.

I will comment some of the changes made.

### DIFF
--- a/lib/segment/src/index/sparse_index/sparse_vector_index.rs
+++ b/lib/segment/src/index/sparse_index/sparse_vector_index.rs
@@ -1,4 +1,5 @@
 use std::collections::HashSet;
+use std::fs::create_dir_all;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
@@ -33,13 +34,16 @@ pub struct SparseVectorIndex<TInvertedIndex: InvertedIndex> {
 }
 
 impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
-    /// Create new sparse vector index
+    /// Open a sparse vector index at a given path
     pub fn open(
         id_tracker: Arc<AtomicRefCell<IdTrackerSS>>,
         vector_storage: Arc<AtomicRefCell<VectorStorageEnum>>,
         payload_index: Arc<AtomicRefCell<StructPayloadIndex>>,
         path: &Path,
     ) -> OperationResult<Self> {
+        // create directory if it does not exist
+        create_dir_all(path)?;
+
         let searches_telemetry = SparseSearchesTelemetry::new();
         let max_point_id = 0;
         let inverted_index = TInvertedIndex::open(path)?;
@@ -158,13 +162,14 @@ impl<TInvertedIndex: InvertedIndex> VectorIndex for SparseVectorIndex<TInvertedI
         let borrowed_id_tracker = self.id_tracker.borrow();
         let deleted_bitslice = borrowed_vector_storage.deleted_vector_bitslice();
         let mut ram_index = InvertedIndexRam::empty();
-        let available_point_count = borrowed_id_tracker.available_point_count();
+        let mut index_point_count: usize = 0;
         for id in borrowed_id_tracker.iter_ids_excluding(deleted_bitslice) {
             check_process_stopped(stopped)?;
             let vector: &SparseVector = borrowed_vector_storage.get_vector(id).try_into()?;
             ram_index.upsert(id, vector.to_owned());
+            index_point_count += 1;
         }
-        self.max_point_id = available_point_count as PointOffsetType - 1;
+        self.max_point_id = index_point_count.saturating_sub(1) as PointOffsetType;
         // TODO(sparse) this operation loads the entire index into memory which can cause OOM on large storage
         self.inverted_index = TInvertedIndex::from_ram_index(ram_index, &self.path)?;
         Ok(())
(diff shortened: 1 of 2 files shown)


lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -33,13 +34,16 @@ pub struct SparseVectorIndex<TInvertedIndex: InvertedIndex> {
...
-    /// Create new sparse vector index
+    /// Open a sparse vector index at a given path
     pub fn open(
         id_tracker: Arc<AtomicRefCell<IdTrackerSS>>,
         vector_storage: Arc<AtomicRefCell<VectorStorageEnum>>,
         payload_index: Arc<AtomicRefCell<StructPayloadIndex>>,
         path: &Path,
     ) -> OperationResult<Self> {
+        // create directory if it does not exist
+        create_dir_all(path)?;
```
agourlay, line 45: I think the `SparseVectorIndex` is responsible for creating the index folder.

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -158,13 +162,14 @@ impl<TInvertedIndex: InvertedIndex> VectorIndex for SparseVectorIndex<TInvertedI
         let borrowed_id_tracker = self.id_tracker.borrow();
         let deleted_bitslice = borrowed_vector_storage.deleted_vector_bitslice();
         let mut ram_index = InvertedIndexRam::empty();
-        let available_point_count = borrowed_id_tracker.available_point_count();
+        let mut index_point_count: usize = 0;
```
agourlay, line 165: switched to actually counting the points to avoid inconsistencies

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -158,13 +162,14 @@ impl<TInvertedIndex: InvertedIndex> VectorIndex for SparseVectorIndex<TInvertedI
...
         for id in borrowed_id_tracker.iter_ids_excluding(deleted_bitslice) {
             check_process_stopped(stopped)?;
             let vector: &SparseVector = borrowed_vector_storage.get_vector(id).try_into()?;
             ram_index.upsert(id, vector.to_owned());
+            index_point_count += 1;
         }
-        self.max_point_id = available_point_count as PointOffsetType - 1;
+        self.max_point_id = index_point_count.saturating_sub(1) as PointOffsetType;
```
agourlay, line 172: protect from overflow if zero point

lib/segment/tests/integration/sparse_vector_index_search_tests.rs
```diff
@@ -21,25 +23,21 @@ use segment::vector_storage::VectorStorage;
...
-
-/// Prepares a sparse vector index with random sparse vectors
-fn fixture_sparse_index<I: InvertedIndex, R: Rng + ?Sized>(
-    rnd: &mut R,
-    max_dim: usize,
-    stopped: &AtomicBool,
-) -> SparseVectorIndex<I> {
-    // test params
-    let num_vectors = 1000;
+const MAX_SPARSE_DIM: usize = 512;
```
agourlay, line 32: decreased this value to speed up the tests execution

lib/segment/tests/integration/sparse_vector_index_search_tests.rs
```diff
@@ -170,20 +180,59 @@ fn sparse_vector_index_ram_consistent_with_storage() {
...
+    let mut rnd = StdRng::seed_from_u64(42);
+
+    let sparse_vector_ram_index = fixture_sparse_index_ram(&mut rnd, MAX_SPARSE_DIM, &stopped);
+
+    // check consistency with underlying RAM inverted index
+    check_index_storage_consistency(&sparse_vector_ram_index);
+
+    let mmap_index_dir = Builder::new().prefix("mmap_index_dir").tempdir().unwrap();
+    // copy index mmap and save to disk
+    let _mmap_inverted_index = InvertedIndexMmap::convert_and_save(
```
agourlay, line 201: important moment here

agourlay:
@IvanPleshkov I also want to discuss the semantic difference for the `open` function between RAM and memmap.

With memmap,  we open an existing wellformed  index and can use it right away.

For RAM, we still have no persistence.
So we create an empty RAM  index that still needs to be filled up. Concretelys, the caller must call `build()` after `open()`.

Is this ok for now?

lib/segment/src/index/sparse_index/sparse_vector_index.rs
IvanPleshkov: Agreed

lib/segment/src/index/sparse_index/sparse_vector_index.rs
IvanPleshkov: agreed, we count points that are not deleted in id tracker and vector storage both

lib/segment/src/index/sparse_index/sparse_vector_index.rs
IvanPleshkov: Or we can keep count of points instead of max index. We use this field to return the count of indexed points. And increment this counter in update if `Self::update_vector` added new vector

lib/segment/tests/integration/sparse_vector_index_search_tests.rs
IvanPleshkov: makes sense

lib/segment/tests/integration/sparse_vector_index_search_tests.rs
IvanPleshkov: I guess we don't have to keep this index alive. I propose to remove `_mmap_inverted_index` variable. And next loading will be processed when this index is dropped

IvanPleshkov (approved):
Also we can add test for empty index if you find it useful

PR #2926, opened 2023-11-03
Improve abort/cancellation support for shard transfer

ffuugoo:
Tracked in #2432.

This PR continues the work in #2918 and makes most of the shard transfer methods and functions cancel safe.

There are a bunch of `TODO`s added by this PR, but I think they should be left as-is for now, and can be worked on in the following PRs.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -4,8 +4,6 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use async_trait::async_trait;
-use cancel::future::cancel_on_token;
-use cancel::CancellationToken;
 use segment::types::{
     ExtendedPointId, Filter, ScoredPoint, WithPayload, WithPayloadInterface, WithVector,
 };
@@ -17,8 +15,8 @@ use super::transfer::shard_transfer::MAX_RETRY_COUNT;
 use super::update_tracker::UpdateTracker;
 use crate::operations::point_ops::WriteOrdering;
 use crate::operations::types::{
-    CollectionError, CollectionInfo, CollectionResult, CoreSearchRequestBatch, CountRequest,
-    CountResult, PointRequest, Record, SearchRequestBatch, UpdateResult,
+    CollectionInfo, CollectionResult, CoreSearchRequestBatch, CountRequest, CountResult,
+    PointRequest, Record, SearchRequestBatch, UpdateResult,
 };
 use crate::operations::CollectionUpdateOperations;
 use crate::shards::local_shard::LocalShard;
@@ -53,9 +51,13 @@ pub struct QueueProxyShard {
 }
 
 impl QueueProxyShard {
-    pub async fn new(wrapped_shard: LocalShard, remote_shard: RemoteShard) -> Self {
+    pub fn new(
+        wrapped_shard: LocalShard,
+        remote_shard: RemoteShard,
+        max_ack_version: Arc<AtomicU64>,
+    ) -> Self {
         Self {
-            inner: Some(Inner::new(wrapped_shard, remote_shard).await),
+            inner: Some(Inner::new(wrapped_shard, remote_shard, max_ack_version)),
         }
     }
 
@@ -79,10 +81,11 @@ impl QueueProxyShard {
     ///
     /// This method is cancel safe.
     ///
-    /// If cancelled - none, some or all operations of that batch may be transmitted to the remote.
+    /// If cancelled - none, some or all operations may be transmitted to the remote.
     ///
-    /// The maximum acknowledged WAL version likely won't be updated. In the worst case this might
-    /// cause double sending operations. This should be fine as operations are idempotent.
+    /// The internal field keeping track of the last transfer and maximum acknowledged WAL version
+    /// likely won't be updated. In the worst case this might cause double sending operations.
+    /// This should be fine as operations are idempotent.
     pub async fn transfer_all_missed_updates(&self) -> CollectionResult<()> {
         self.inner
             .as_ref()
@@ -120,35 +123,6 @@ impl QueueProxyShard {
     #[cfg(debug_assertions)]
     fn is_finalized(&self) -> bool {
         self.inner.is_none()
-    }
-
-    /// Finalize: transfer all updates to remote and unwrap the wrapped shard
-    ///
-    /// This method helps with safely destructing the queue proxy shard, ensuring that remaining
-    /// queue updates are transferred to the remote shard, and only unwrapping the local shard
-    /// on success. It also releases the max acknowledged WAL version.
-    ///
-    /// Because we have ownership (`self`) we have exclusive access to the internals. It guarantees
-    /// that we will not process new operations on the shard while finalization is happening.
-    ///
-    /// # Cancel safety
-    ///
-    /// This method is *not* cancel safe, it consumes `self`.
-    ///
-    /// If `cancel` is triggered - finalization may not actually complete in which case an error is
-    /// returned. None, some or all operations may be transmitted to the remote.
-    pub async fn finalize(
-        self,
-        cancel: CancellationToken,
-    ) -> Result<(LocalShard, RemoteShard), (CollectionError, Self)> {
-        // Transfer all updates, do not unwrap on failure but return error with self
-        match cancel_on_token(cancel, self.transfer_all_missed_updates()).await {
-            Ok(Ok(_)) => Ok(self.forget_updates_and_finalize()),
-            // Transmission error
-            Ok(Err(err)) => Err((err, self)),
-            // Cancellation error
-            Err(err) => Err((err.into(), self)),
-        }
     }
 
     /// Forget all updates and finalize.
@@ -301,14 +275,13 @@ struct Inner {
 }
 
 impl Inner {
-    pub async fn new(wrapped_shard: LocalShard, remote_shard: RemoteShard) -> Self {
+    pub fn new(
+        wrapped_shard: LocalShard,
+        remote_shard: RemoteShard,
+        max_ack_version: Arc<AtomicU64>,
+    ) -> Self {
         let last_idx = wrapped_shard.wal.lock().last_index();
-        let max_ack_version = wrapped_shard
-            .update_handler
-            .lock()
-            .await
-            .max_ack_version
-            .clone();
+
         let shard = Self {
             wrapped_shard,
             remote_shard,
@@ -331,8 +304,9 @@ impl Inner {
     ///
     /// If cancelled - none, some or all operations of that batch may be transmitted to the remote.
     ///
-    /// The maximum acknowledged WAL version likely won't be updated. In the worst case this might
-    /// cause double sending operations. This should be fine as operations are idempotent.
+    /// The internal field keeping track of the last transfer and maximum acknowledged WAL version
+    /// likely won't be updated. In the worst case this might cause double sending operations.
+    /// This should be fine as operations are idempotent.
     pub async fn transfer_all_missed_updates(&self) -> CollectionResult<()> {
         while !self.transfer_wal_batch().await? {}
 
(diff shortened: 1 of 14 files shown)


lib/collection/src/shards/transfer/shard_transfer.rs
```diff
@@ -282,25 +283,23 @@ async fn transfer_snapshot(
...
 
-    error_if_cancelled(&cancel)?;
-
     // Create shard snapshot
     log::trace!("Creating snapshot of shard {shard_id} for shard snapshot transfer");
     let snapshot_description = shard_holder_read
         .create_shard_snapshot(snapshots_path, collection_name, shard_id, temp_dir)
         .await?;
 
+    // TODO: If future is cancelled until `get_shard_snapshot_path` resolves, shard snapshot may not be cleaned up...
```
timvisee, line 302: Maybe we do want some function that returns a `TempFile` to prevent this.

lib/collection/src/shards/transfer/shard_transfer.rs
```diff
@@ -773,67 +760,72 @@ where
...
+
+            result = cancel::future::cancel_on_token(cancel.clone(), future)
+                .await
+                .map_err(Into::into)
+                .and_then(|res| res);
+
+            match &result {
+                Ok(()) => break,
+
+                // TODO: Should we break early if shard transfer is cancelled, or should we handle it as regular error?
```
timvisee, line 802: We want to handle it as regular error. More specifically, we want to call `revert_queue_proxy_local()`. We probably call this later as well as part of aborting the snapshot, but doing it here as well is a good thing to be sure.

lib/collection/src/shards/transfer/shard_transfer.rs
```diff
@@ -773,67 +760,72 @@ where
...
-            // On the end of transfer, the new shard is active but most likely is under the optimization
-            // process. Requests to this node might be slow, but we rely on the assumption that
-            // there should be at least one other replica that is not under optimization.
-            on_finish.await;
-        } else {
-            on_error.await;
+        match &result {
+            Ok(()) => on_finish.await,
+
+            // TODO: Should we ignore `on_error` if shard transfer is cancelled, or should we handle it as regular error?
```
timvisee, line 823: I think this branch is fine; doing nothing on cancel.

lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -53,9 +51,13 @@ pub struct QueueProxyShard {
 }
 
 impl QueueProxyShard {
-    pub async fn new(wrapped_shard: LocalShard, remote_shard: RemoteShard) -> Self {
+    pub fn new(
+        wrapped_shard: LocalShard,
+        remote_shard: RemoteShard,
+        max_ack_version: Arc<AtomicU64>,
```
timvisee, line 57: I don't like that we expose the `max_ack_version` here. I get that it's to make the function non-async.

But I don't think we should expose this to the outside.

Can we find some middle ground here?

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -369,10 +373,16 @@ impl ShardReplicaSet {
...
     /// Uses a blocking thread internally.
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is cancel safe.
     async fn wait_for<F>(&self, check: F, timeout: Duration) -> CollectionResult<()>
     where
         F: Fn(&ReplicaSetState) -> bool + Send + 'static,
     {
+        // TODO: Propagate cancellation into `spawn_blocking` task!?
```
timvisee, line 384: Probably a good idea. It currently relies on the timeout to eventually drop the thread.

Moving the cancellation flag in and checking on each loop is probably already good enough.

lib/collection/src/shards/replica_set/shard_transfer.rs
```diff
@@ -205,57 +253,72 @@ impl ShardReplicaSet {
...
-        }
+            )));
+        };
+
+        log::trace!(
+            "Transferring indexes to shard {}",
+            proxy.remote_shard.peer_id,
+        );
+
+        // TODO: Is cancelling `ForwardProxyShard::transfer_indexes` safe for *receiver*?
```
timvisee, line 320: Yes.

lib/collection/src/shards/transfer/shard_transfer.rs
timvisee: What about changing this `match` to:

```rust
match &result {
    Ok(()) => break,

    Err(err) => {
        // Revert queue proxy if we still have any to clean up
        if let Some(shard) = shards_holder.read().await.get_shard(&transfer.shard_id) {
            shard.revert_queue_proxy_local().await;
        }

        if let CollectionError::Cancelled { .. } = err {
            break;
        }

        log::error!(
            "Failed to transfer shard {} -> {}: {err}",
            transfer.shard_id,
            transfer.to,
        );
    },
}
```

ffuugoo: Ooooh, that's a nice suggestion, I like it. 👌

timvisee: It does reorder logging after reverting the queue proxy, but I guess it's fine.

timvisee: Now that I think of it, we can probably remove the whole match and do an 'early return' if `Ok(())`` to clean this up even further.

lib/collection/src/shards/queue_proxy_shard.rs
ffuugoo: I don't like it either, but I don't see other way to do it.

And it's either make `QueueProxyShard::now` sync, or `ShardReplicaSet::queue_proxify_local` is *not* cancel safe and there are cancellation tokens everywhere. :/

timvisee:
Very nice work! Happy to approve once we resolve the above comments.

lib/collection/src/shards/replica_set/mod.rs
ffuugoo: It's just rather non-trivial to do (because we have a sync task blocked on a cond-var), so I left it as `TODO` for now.

lib/collection/src/shards/transfer/shard_transfer.rs
```diff
@@ -496,8 +495,6 @@ pub async fn handle_transferred_shard_proxy(
     to: PeerId,
     sync: bool,
 ) -> CollectionResult<bool> {
-    // TODO: Ensure cancel safety!
-
```
ffuugoo: I don't think this is resolved, cause there's `replica_set.remove_local().await` below that may not be cancel safe, and so `handle_transferred_shard_proxy` may not be cancel safe.

lib/collection/src/shards/queue_proxy_shard.rs
timvisee: Yeah this is tricky.

Does using explicit cancellation tokens in `queue_proxify_local` look that bad though? Because to be honest, I don't like all the restructuring we're required to do just for this any better.

lib/collection/src/shards/transfer/shard_transfer.rs
timvisee: Fair enough, but the same TODO is still there a few lines below.

timvisee: My eye now catches the `add_remote` line which may be troublesome as well. 

Alright, lets put it back. Sorry.

lib/collection/src/shards/queue_proxy_shard.rs
ffuugoo: Hmmm... Maybe you're right. We'd have to propagate cancellation token:

- from the shard transfer task to `transfer_shard`
- from `transfer_shard` to `transfer_snapshot`
- and refactor `transfer_snapshot` so that it would `cancel_on_token` everything before and after `queue_proxify_local` and propagate into it

Which might be reasonable.

lib/collection/src/shards/transfer/shard_transfer.rs
ffuugoo: > Fair enough, but the same TODO is still there a few lines below.

My bad. It's how I format these things. I put two of these `TODO`s, one at the top, and one at a particular thing that need work (if there's a specific item identified).

lib/collection/src/shards/queue_proxy_shard.rs
timvisee: I guess we should keep it as-is for now, since changing this again may be a lot of work. We can revisit this idea later. The current implementation seems logically correct to me. :+1:

lib/collection/src/shards/transfer/shard_transfer.rs
ffuugoo: Another refactoring of `create_shard_snapshot` (or adding a separate method) 🥲

lib/collection/src/shards/transfer/shard_transfer.rs
ffuugoo: I've restructured `spawn_transfer_task`, what do you think? 🤔

lib/collection/src/shards/transfer/shard_transfer.rs
ffuugoo: Removed.

lib/collection/src/shards/transfer/shard_transfer.rs
```diff
@@ -790,41 +786,37 @@ where
...
+            // If `transfer_shard` returned error or task was cancelled...
+            if matches!(result, Ok(Err(_)) | Err(_)) {
+                // Revert queue proxy if we still have any to prepare for the next attempt
+                if let Some(shard) = shards_holder.read().await.get_shard(&transfer.shard_id) {
+                    shard.revert_queue_proxy_local().await;
                 }
             }
+
+            // If `transfer_shard` returned success or task was cancelled...
+            if !matches!(result, Ok(Err(_))) {
```
ffuugoo, line 809: This may be more readable, though...

```suggestion
            if matches!(result, Ok(Ok(())) | Err(_)) {
```

timvisee: Yeah lets do this one. I was a bit confused by the negation.

lib/collection/src/shards/transfer/shard_transfer.rs
```diff
@@ -773,67 +759,64 @@ where
...
+                // Revert queue proxy if we still have any to prepare for the next attempt
+                if let Some(shard) = shards_holder.read().await.get_shard(&transfer.shard_id) {
+                    shard.revert_queue_proxy_local().await;
+                }
+            }
+
+            // If `transfer_shard` returned success or task was cancelled...
+            if matches!(result, Ok(Ok(())) | Err(_)) {
+                break;
             }
```
timvisee, line 811: What about this? We use this pattern in some other places as well.

```suggestion
            let is_cancelled = result.is_err();
            let is_ok = matches!(result, Ok(Ok(())));
            let is_err = matches!(result, Ok(Err(_)));

            // If `transfer_shard` returned error or task was cancelled...
            if is_cancelled || is_err {
                // Revert queue proxy if we still have any to prepare for the next attempt
                if let Some(shard) = shards_holder.read().await.get_shard(&transfer.shard_id) {
                    shard.revert_queue_proxy_local().await;
                }
            }

            // If `transfer_shard` returned success or task was cancelled...
            if is_ok || is_cancelled {
                break;
            }
```

ffuugoo: That's... much better, actually. I don't know why I didn't implement it like that in the first place. 🤦‍♀️

ffuugoo: Done.

timvisee: I also always get blind for improvements like this after refactoring some code a million times. :laughing: 

> Done.

Awesome!

timvisee:
Let's merge once :green_circle:

timvisee (approved):

PR #2927, opened 2023-11-03
Int indexing

generall:
Currently, if we insert new value in payload index with `add_point`, it only clears existing data if the new value is "parsable" (compatible with the index). 
But in reality, it can be any value, e.g. "null". That would mean that old value should still be removed from index even if the new one is not indexable.

Simple way to reproduce problem:

```
POST collections/benchmark/points/payload
{
  "points": [1,2,3],
  "payload": {
    "c": null
  }
}
```


coszio (approved):
Looks like a sane fix for me

IvanPleshkov (approved):
Make sense, great observation!

timvisee (approved):

PR #2935, opened 2023-11-06
Attempt to reduce build cache invalidation

timvisee:
Attempts to reduce build cache invalidation. I'm seeing a lot of seemingly unreasonable invalidations, even if there are no code changes, causing expensive rebuilds. This PR attempts to minimize this.

I've mostly come to this with a trial and error approach using `cargo build -vv`. I don't have hard proof for all changes but it seems to improve the situation on my system.

- Update troublesome dependencies:
  Some dependencies, such as `rustix` ([e.g.](https://github.com/bytecodealliance/rustix/issues/575)) and `procfs` seem to have a bad effect making the build cache dirty. This updates them, and dependents if needed.
- Remove generated gRPC `qdrant.rs` from repository:
  Removes 10k lines of code we have to keep up-to-date. It is replaced with [`include_proto!()`](https://docs.rs/tonic/latest/tonic/macro.include_proto.html), which is intended for this
- Remove hacky `append_to_file` for gRPC validation:
  To get our validation additions into the generated gRPC code, we manually appended an import statement to the end of the generated file during build time. It caused the generated file to be modified a second time. This is now entirely removed, possible due to the above macro.

After this PR I'm still seeing this every once in a while:

```
       Dirty api v0.11.1 (/home/timvisee/git/qdrant/lib/api): the dependency build_script_build was rebuilt (1699264815.924878183s, 924878183ns after last build at 1699264815.000000000s)
   Compiling api v0.11.1 (/home/timvisee/git/qdrant/lib/api)
```

Though, according to [this](https://stackoverflow.com/a/76552659/1000145) it's caused by a Cargo bug (<https://github.com/rust-lang/cargo/issues/12060>) which is yet to be resolved.

### All Submissions:


generall (approved):

timvisee:
Drafting this because removing `qdrant.rs` may actually be quite annoying for inspecting the generated gRPC Rust code. I'll have to think another time if we want to go through with this.

You can inspect the generated code with `cargo expand --manifest-path ./lib/api/Cargo.toml grpc::qdrant` but it's a pain to work with.

agourlay:
I like the direction of this PR a lot, it seems way less hacky :+1: 

> Drafting this because removing qdrant.rs may actually do more bad than good for some developers. 

Is it much more difficult to see the generated content with this PR?

timvisee:
> Is it much more difficult to see the generated content with this PR?

Yes, as in, it isn't as simple anymore as opening `qdrant.rs` in your favorite editor.

You can use `cargo expand --manifest-path ./lib/api/Cargo.toml grpc::qdrant`, but it dumps to stdout. It also over-expands macros making the file 50k lines rather than 10k.

JetBrains has an option to 'Show expanded macro' though I assume it would result in the same as above.

The generated source is available in the `target` directory, but it contains multiple (hashed) versions (at `target/debug/build/api-*/out/qdrant.rs`). So it's tricky to point to the current latest one. Not impossible, but just annoying.

coszio:
@timvisee a point in pro of this PR would be to not having to run `cargo build` after making changes to the proto, is this correct?

timvisee:
> @timvisee a point in pro of this PR would be to not having to run `cargo build` after making changes to the proto, is this correct?

No. You'd always have to run the compiler once to generate the new gRPC files, even for how it's changed in this PR.

timvisee:
The original issue does not appear for me anymore. So I'll close this now.

If we ever decide removing the generated `qdrant.rs` gRPC boiletplate from the repository is desirable, as suggested in this PR, we can revisit this.

PR #2942, opened 2023-11-07
Add backoff when syncing locally disabled peers state to consensus

ffuugoo:
This PR adds a "backoff" when synchronizing local state to consensus during `sync_local_state`.

This should not be necessary after changes in #3012/#3013/#3026 but still seems "nice to have".

Currently based on #3026, will be rebased on `dev` once it's merged.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -1,4 +1,5 @@
 mod execute_read_operation;
+mod locally_disabled_peers;
 mod read_ops;
 mod shard_transfer;
 mod snapshots;
@@ -78,7 +79,7 @@ pub struct ShardReplicaSet {
     /// List is checked on each consensus round and submitted to the consensus.
     /// If the state of the peer is changed in the consensus, it is removed from the list.
     /// Update and read operations are not performed on the peers marked as dead.
-    locally_disabled_peers: parking_lot::RwLock<HashSet<PeerId>>,
+    locally_disabled_peers: parking_lot::RwLock<locally_disabled_peers::Registry>,
     pub(crate) shard_path: PathBuf,
     pub(crate) shard_id: ShardId,
     notify_peer_failure_cb: ChangePeerState,
@@ -285,7 +286,7 @@ impl ShardReplicaSet {
             replica_set
                 .locally_disabled_peers
                 .write()
-                .insert(this_peer_id);
+                .disable_peer(this_peer_id);
         }
 
         replica_set
@@ -662,11 +663,18 @@ impl ShardReplicaSet {
 
     /// Check if the are any locally disabled peers
     /// And if so, report them to the consensus
-    pub async fn sync_local_state<F>(&self, get_shard_transfers: F) -> CollectionResult<()>
+    pub fn sync_local_state<F>(&self, get_shard_transfers: F) -> CollectionResult<()>
     where
         F: Fn(ShardId, PeerId) -> Vec<ShardTransfer>,
     {
-        for &failed_peer in self.locally_disabled_peers.read().iter() {
+        let peers_to_notify: Vec<_> = self
+            .locally_disabled_peers
+            .write()
+            .notify_elapsed()
+            .collect();
+
+        for failed_peer in peers_to_notify {
+            // TODO: Only `notify_peer_failure` if `failed_peer` is *not* the last `Active` peer? 🤔
             self.notify_peer_failure(failed_peer);
 
             for transfer in get_shard_transfers(self.shard_id, failed_peer) {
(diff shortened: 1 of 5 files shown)


lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -789,6 +795,39 @@ pub enum ReplicaState {
...
+    fn default() -> Self {
+        Self {
+            last_attempt: Instant::now(),
+            backoff: Duration::from_secs(0),
+        }
+    }
+}
+
+impl SyncState {
+    pub fn retry(&mut self) -> bool {
```
timvisee, line 814: Nitpick, but maybe we should renamed this to something like `should_retry_now()`.

timvisee (approved):
Looks fine by me.

Though, we should test this on a cluster first before merging it into the debug branch.

lib/collection/src/shards/replica_set/mod.rs
ffuugoo: Done.

ffuugoo:
@generall @timvisee I've created a copy of this PR targeting `master` branch at #2945.

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -77,7 +78,7 @@ pub struct ShardReplicaSet {
     /// List is checked on each consensus round and submitted to the consensus.
     /// If the state of the peer is changed in the consensus, it is removed from the list.
     /// Update and read operations are not performed on the peers marked as dead.
-    locally_disabled_peers: parking_lot::RwLock<HashSet<PeerId>>,
+    locally_disabled_peers: parking_lot::RwLock<HashMap<PeerId, parking_lot::Mutex<SyncState>>>,
```
generall, line 81: WDYT about changing the whole HashSet into custom structure? Otherwise looks like too much internals are exposed

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -789,6 +795,39 @@ pub enum ReplicaState {
     PartialSnapshot,
 }
 
+#[derive(Copy, Clone, Debug)]
+struct SyncState {
```
generall, line 799: let's move into separate file

lib/collection/src/shards/replica_set/mod.rs
ffuugoo: Done. Kinda. I've added a separate structure for backoff.

lib/collection/src/shards/replica_set/mod.rs
ffuugoo: Done.

lib/collection/src/shards/replica_set/backoff.rs
```diff
@@ -0,0 +1,74 @@
+use std::cmp;
+use std::collections::HashMap;
+use std::time::{Duration, Instant};
+
+use crate::shards::shard::PeerId;
+
+#[derive(Clone, Debug, Default)]
+pub struct Registry {
```
ffuugoo, line 8: Not sure about the name, suggestions welcome. Don't want to call it `NotifyPeerFailureBackoff`, though, because the type itself is _kinda_ generic.

E.g., we can add `abort_shard_transfer: backoff::Registry` to the `ShardReplicaSet` and use it to rate-limit `abort_shard_transfer` calls independent of `notify_peer_failure` calls.

timvisee: What about something like `RetryLimiter`?

I find `Registry` a bit took ambiguous for something that specifically throttles/limits retry calls for a list of peers.

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -767,9 +790,14 @@ impl ShardReplicaSet {
 
         locally_disabled.remove(&peer_id_to_remove);
 
+        self.notify_peer_failure_backoff
+            .write()
+            .reset(peer_id_to_remove);
+
         if active_peers.is_subset(&locally_disabled) {
             log::warn!("Resolving consensus/local state inconsistency");
```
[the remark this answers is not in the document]
ffuugoo: The idea is that if at some point we have *all* active peers "locally disabled" on a node, we treat is as a major consensus/local state desync and clear locally disabled peers list.

If all these nodes are truly dead, then following updates would fail, and we'll add them back to the locally disabled list. But if they're *not* dead (e.g., there was a network issue on this particular node, and it marked everything as dead), then we better clear the locally disabled list.

ffuugoo:
@generall @timvisee Merged `ShardReplicaSet::locally_disabled_peers` and `replica_set::backoff::Registry` into a single `locally_disabled_peers::Registry` structure. I think this is done now and ready for the review.

generall (approved):

PR #2944, opened 2023-11-07
Sparse vector search benchmarks

agourlay:
This PR adds the benchmark infrastructure for sparse vector search.

e.g
search-perf

This enables to optimize the underlying implementation (e.g https://github.com/qdrant/qdrant/pull/2937) and size properly the filter based search heuristic in the future.

To make this possible, some fixture files were refactored and extracted into the segments crate lib code,

### DIFF
--- a/lib/segment/benches/sparse_index_search.rs
+++ b/lib/segment/benches/sparse_index_search.rs
@@ -1,0 +1,62 @@
+#[cfg(not(target_os = "windows"))]
+mod prof;
+
+use std::sync::atomic::AtomicBool;
+
+use criterion::{criterion_group, criterion_main, Criterion};
+use rand::rngs::StdRng;
+use rand::SeedableRng;
+use segment::fixtures::sparse_fixtures::fixture_sparse_index_ram;
+use segment::index::VectorIndex;
+use sparse::common::sparse_vector_fixture::random_sparse_vector;
+
+const NUM_VECTORS: usize = 10000;
+const MAX_SPARSE_DIM: usize = 512;
+
+fn sparse_vector_index_search_benchmark(c: &mut Criterion) {
+    let mut group = c.benchmark_group("sparse-vector-search-group");
+
+    let stopped = AtomicBool::new(false);
+    let mut rnd = StdRng::seed_from_u64(42);
+    let sparse_vector_index =
+        fixture_sparse_index_ram(&mut rnd, NUM_VECTORS, MAX_SPARSE_DIM, &stopped);
+
+    let mut result_size = 0;
+    let mut query_count = 0;
+
+    group.bench_function("sparse-index-search", |b| {
+        b.iter(|| {
+            let sparse_vector = random_sparse_vector(&mut rnd, MAX_SPARSE_DIM);
+            let query_vector = sparse_vector.into();
+            result_size += sparse_vector_index
+                .search(&[&query_vector], None, 1, None, &stopped)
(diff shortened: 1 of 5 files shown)


lib/segment/tests/integration/sparse_vector_index_search_tests.rs
```diff
@@ -191,18 +116,20 @@ fn sparse_vector_index_consistent_with_storage() {
...
 
     let mmap_index_dir = Builder::new().prefix("mmap_index_dir").tempdir().unwrap();
     // copy index mmap and save to disk
-    let _mmap_inverted_index = InvertedIndexMmap::convert_and_save(
+    let mmap_inverted_index = InvertedIndexMmap::convert_and_save(
         &sparse_vector_ram_index.inverted_index,
         &mmap_index_dir,
     )
     .unwrap();
+    drop(mmap_inverted_index);
```
agourlay, line 132: suggestion from previous review :)

agourlay:
@IvanPleshkov I recommend treating this PR as merely setup to move files around.

The real work is taking place in https://github.com/qdrant/qdrant/pull/2947

lib/segment/benches/sparse_index_search.rs
```diff
@@ -0,0 +1,62 @@
...
+    let mut rnd = StdRng::seed_from_u64(42);
+    let sparse_vector_index =
+        fixture_sparse_index_ram(&mut rnd, NUM_VECTORS, MAX_SPARSE_DIM, &stopped);
+
+    let mut result_size = 0;
+    let mut query_count = 0;
+
+    group.bench_function("sparse-index-search", |b| {
+        b.iter(|| {
+            let sparse_vector = random_sparse_vector(&mut rnd, MAX_SPARSE_DIM);
```
IvanPleshkov, line 29: I'm not sure about the speed of random creation. When I worked on quantization I realized, that `rand` crate is not fast as a light. What do you think? May I check numbers with predefined request vectors?

lib/segment/benches/sparse_index_search.rs
```diff
@@ -0,0 +1,62 @@
...
+            result_size += sparse_vector_index
+                .search(&[&query_vector], None, 1, None, &stopped)
+                .unwrap()
+                .len();
+
+            query_count += 1;
+        })
+    });
+
+    eprintln!(
```
IvanPleshkov, line 40: Why do we need this counting? Criterion should measure mean search time in this case

lib/segment/src/fixtures/sparse_fixtures.rs
IvanPleshkov, line 1: I see this file is just moving of code

IvanPleshkov (approved):

lib/segment/benches/sparse_index_search.rs
agourlay: So far the `rnd` code does not show in the profile but I think it is a good idea to build a predefine set of vectors upfront to compare :+1: 

Will do in next PR!

lib/segment/benches/sparse_index_search.rs
agourlay: I will remove those :+1:

lib/segment/src/fixtures/sparse_fixtures.rs
agourlay: yes, moving for making it visible in bench.

PR #2947, opened 2023-11-08
Benchmark sparse filter plain search

agourlay:
This PR introduces benchmarks for various access pattern on the sparse index vector.

The idea is to gather data to find the sweet spot for a potential fallback to plain & payload search.

## Benchmarks

Run it yourself with:

`cargo bench -p segment --bench sparse_index_search`

The current parameters are:

- 50k points
- sparse vector query with 200 values
- top 10
- with and without filter
- with and without payload index

sparse-plain

- `inverted-index`: sparse search without filter
- `inverted-index-filtered-payload-index`: sparse search with filter on indexed payload key
- `inverted-index-filtered-payload-plain`: sparse search with filter on non-indexed payload key
- `payload-index`: plain search using filter on indexed payload key
- `plain-storage`: plain search using filter on non-indexed payload key

## Insights

While playing with the size of the sparse vector query, I noticed it has a large impact on the inverted index performance (smaller better).
  - less than 100 is fast
  - less than 300 is Ok. 

## Future Work

Wire the `search_plain` method behind a heuristic for fallback.

Something like:
- if small query vector then use sparse index
- if low cardinality use plain search

### DIFF
--- a/lib/segment/src/index/sparse_index/sparse_vector_index.rs
+++ b/lib/segment/src/index/sparse_index/sparse_vector_index.rs
@@ -21,7 +21,7 @@ use crate::index::{PayloadIndex, VectorIndex};
 use crate::telemetry::VectorIndexSearchesTelemetry;
 use crate::types::{Filter, SearchParams};
 use crate::vector_storage::sparse_raw_scorer::sparse_check_vector;
-use crate::vector_storage::{VectorStorage, VectorStorageEnum};
+use crate::vector_storage::{new_stoppable_raw_scorer, VectorStorage, VectorStorageEnum};
 
 pub struct SparseVectorIndex<TInvertedIndex: InvertedIndex> {
     pub id_tracker: Arc<AtomicRefCell<IdTrackerSS>>,
@@ -69,7 +69,7 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         with_filter: bool,
         condition: impl Fn(PointOffsetType) -> bool,
     ) -> OperationResult<Vec<Vec<ScoredPointOffset>>> {
-        let mut result = vec![];
+        let mut result = Vec::with_capacity(vectors.len());
 
         for vector in vectors {
             check_process_stopped(is_stopped)?;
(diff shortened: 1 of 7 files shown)


lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -120,6 +120,35 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         }
         unique_record_ids.len()
     }
+
+    /// Plain search not using the inverted index.
```
agourlay, line 124: new plain implementation not wired yet

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -120,6 +120,35 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
+        &self,
+        vectors: &[&QueryVector],
+        filter: &Filter,
+        top: usize,
+        is_stopped: &AtomicBool,
+    ) -> OperationResult<Vec<Vec<ScoredPointOffset>>> {
+        let mut results = Vec::with_capacity(vectors.len());
+        let id_tracker = self.id_tracker.borrow();
+        let payload_index = self.payload_index.borrow();
+        let vector_storage: &VectorStorageEnum = &self.vector_storage.borrow();
```
IvanPleshkov, line 135: I guess `&VectorStorageEnum` is not required here

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -120,6 +120,35 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         }
         unique_record_ids.len()
     }
+
+    /// Plain search not using the inverted index.
+    pub fn search_plain(
+        &self,
+        vectors: &[&QueryVector],
+        filter: &Filter,
```
IvanPleshkov, line 128: I agree with the filter. plain search is applicable only when filtering is enabled

lib/sparse/src/common/sparse_vector_fixture.rs
```diff
@@ -2,23 +2,33 @@ use rand::Rng;
 
 use crate::common::sparse_vector::SparseVector;
 
+const MIN_WEIGHT: f64 = -100.0;
```
IvanPleshkov, line 5: As a small nitpick I can propose to define a constant range instead of min-max values

IvanPleshkov (approved):

PR #2951, opened 2023-11-08
Transform `max_ack_version` from mutex into atomic

timvisee:

This transforms the `max_ack_version` parameter in the update handler from a shared mutex into a shared atomic.

It is required for safe shard snapshot transfer cancellation. For safety, it cannot rely on `async`/`await` which was used for locking, and a synchronous lock would block the runtime. An atomic parameter solves the issue.

As a bonus this removes some locking steps across our code base, and removes the `async` marker from some methods.

### All Submissions:

### New Feature Submissions:


ffuugoo (approved):
LGTM. This will makes shard transfer cancel safety _trivial_. 🎉

timvisee:
I'll merge once this is green and rebase/update the other PRs I had ready. Then we'll easily be able to get through with this feature.

PR #2958, opened 2023-11-09
Share filter results for plain search in HNSW

agourlay:
This PR improves the plain search fallback for HNSW.

The plain search is used for small cardinality and exact search, so it may handle a large amount of points.

I have noticed that we are re-evaluating the filter for each query vectors in the batch query.

This PR pre-computes the filtered points once to reuse it.


IvanPleshkov (approved):
Good catch! We have a todo to remove a collected list of filtered points. But currently it's a true low-hanging fruit, great!

timvisee (approved):

PR #2963, opened 2023-11-09
Add possible fallback to plain search for sparse vectors

agourlay:
This PR adds a simple fallback to filtered search based on the cardinality estimated.

The fallback is done using a simplified version of the heuristic done for HNSW.

While writing tests to assert the equivalence of the results between inverted index and plain search, I noticed two issues.

1 - The inverted index does not handle negative weights whereas plain search does.

Negative weights do not interact well with pruning, I will propose a fix in a different PR

2 - The plain search is returning more vector than the inverted index

It happens for the vectors which have no intersections with the query vector in dimension.
The sparse inverted index never returns those, because those posting lists are not explored.
Whereas the plain search returns those and scores them zero.

The proposed solution is to filter out explicitly plain filtered vectors without overlap in dimensions within the `SparseRawScorer.`

### DIFF
--- a/lib/segment/src/fixtures/sparse_fixtures.rs
+++ b/lib/segment/src/fixtures/sparse_fixtures.rs
@@ -12,6 +12,7 @@ use sparse::index::inverted_index::InvertedIndex;
 use crate::common::operation_error::OperationResult;
 use crate::common::rocksdb_wrapper::{open_db, DB_VECTOR_CF};
 use crate::fixtures::payload_context_fixture::FixtureIdTracker;
+use crate::index::sparse_index::sparse_index_config::SparseIndexConfig;
 use crate::index::sparse_index::sparse_vector_index::SparseVectorIndex;
 use crate::index::struct_payload_index::StructPayloadIndex;
 use crate::index::VectorIndex;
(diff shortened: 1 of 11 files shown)


lib/segment/src/fixtures/sparse_fixtures.rs
```diff
@@ -22,11 +23,15 @@ use crate::vector_storage::VectorStorage;
 
 /// Helper to open a test sparse vector index
 pub fn fixture_open_sparse_index<I: InvertedIndex>(
-    index_dir: &Path,
-    payload_dir: &Path,
-    storage_dir: &Path,
+    data_dir: &Path,
```
agourlay, line 26: refactor to have less arguments

lib/sparse/src/common/sparse_vector.rs
```diff
@@ -42,6 +42,21 @@ impl SparseVector {
         }
         score
     }
+
+    /// Check if this vector overlaps with another vector.
+    /// Warning: Expects both vectors to be sorted by indices.
```
IvanPleshkov, line 47: We can remove this warning because we validate this condition in `SparseVector` constructor

lib/segment/src/vector_storage/sparse_raw_scorer.rs
```diff
@@ -98,28 +109,16 @@ impl<'a, TVectorStorage: SparseVectorStorage> RawScorer for SparseRawScorer<'a,
...
             .take_while(|_| !self.is_stopped.load(Ordering::Relaxed))
             .filter(|point_id| self.check_vector(*point_id))
-            .map(|point_id| {
-                let vector = self.vector_storage.get_sparse(point_id);
-                ScoredPointOffset {
-                    idx: point_id,
-                    score: vector.score(&self.query),
-                }
-            });
+            .filter_map(|point_id| self.score_overlapping_point(point_id));
```
IvanPleshkov, line 112: I guess we have a bug here. If the search query vector does not overlap everything in dataset, the user will get an empty result instead of some `TOP` zero-scored vectors.
I didn't get an idea why do we need such filtering and overlap checks

IvanPleshkov (changes requested):

lib/segment/src/vector_storage/sparse_raw_scorer.rs
agourlay: We need such filter because the inverted index does not return points which have zero overlaps in dimension with the query.

The goal is to make index and plain search equivalent.

IvanPleshkov (approved):

PR #2965, opened 2023-11-09
Sparse vector validation

agourlay:
This PR adds an implementation of `Validate` for sparse vectors.

This will ensure we are not writing tests for malformed vectors.

Most constraints are not controversial except having the indices sorted which needs to be discussed later again.


lib/sparse/src/common/sparse_vector.rs
```diff
@@ -1,19 +1,31 @@
...
-    pub fn new(indices: Vec<DimId>, values: Vec<DimWeight>) -> SparseVector {
-        SparseVector { indices, values }
+    pub fn new(indices: Vec<DimId>, values: Vec<DimWeight>) -> Result<Self, ValidationErrors> {
+        let vector = SparseVector { indices, values };
+        vector.validate()?;
+        Ok(vector)
     }
 
     /// Score this vector against another vector using dot product.
+    /// Warning: Expects both vectors to be sorted by indices.
```
agourlay, line 28: important requirements for the current implementation

lib/sparse/src/common/sparse_vector.rs
```diff
@@ -32,14 +45,88 @@ impl SparseVector {
...
+        }
+        if self.indices.len() != self.indices.iter().unique().count() {
+            errors.add("indices", ValidationError::new("must be unique"));
+        }
+
+        if self.indices.windows(2).any(|w| w[0] > w[1]) {
+            errors.add("indices", ValidationError::new("must be sorted"));
+        }
+
+        if self.values.iter().any(|v| v.is_nan() || *v == 0.0) {
```
IvanPleshkov, line 86: Why value cannot be zero?

lib/sparse/src/common/sparse_vector.rs
```diff
@@ -32,14 +44,81 @@ impl SparseVector {
...
+        if self.values.is_empty() {
+            errors.add("values", ValidationError::new("must be non-empty"));
+        }
+        if self.indices.len() != self.values.len() {
+            errors.add(
+                "values",
+                ValidationError::new("must be the same length as indices"),
+            );
+        }
+        if self.indices.len() != self.indices.iter().unique().count() {
```
IvanPleshkov, line 77: I propose to remove this check because `unique` is not cheap and next check cover this one

IvanPleshkov (approved):

lib/sparse/src/common/sparse_vector.rs
agourlay: I decided to introduce instead a cheaper linear check and keep it separated from the sorting check.
This makes a better UX.
Also added a bunch of tests :innocent: 
https://github.com/qdrant/qdrant/pull/2965/commits/ca9f868fe9c216e814aa00f4fca5801f864b68ab

PR #2968, opened 2023-11-09
Fix results ordering for custom scored searches

coszio:
Following #2904 and the new discovery api, while implementing brute-force local mode I realized ordering was not quite right when using euclidean metric. It would show the same results as in python implementation, but in inverse order.

This fixes it by setting the order to depend on metric but also on `QueryEnum` variant of the search.


generall (approved):

generall:
> while implementing brute-force local mode I realized ordering was not quite right

Having a second implementation of the same thing is the best test case! ;)

PR #2969, opened 2023-11-09
check that there are no pending operations, before syncing local state

generall:
I haven't check that it have any effect on misfire of `sync_local_state`, but it looks safer to have this condition rather than not have.


ffuugoo (approved):
LGTM

PR #2970, opened 2023-11-10
Remove deprecated search methods

generall:
Last version introduced duplicated search methods for the purposes of the migration to a unified internal API for extendable modes of the search.
This PR removed duplication and completely switches search into the new version of internal API.

This change is required, as custom sharding requires a significant change in those search methods and duplication expected changes seemed unpractical.

Compatibility notes: this change makes migration from versions <=1.5 directly to >=1.7   (skipping 1.6.x) potentially error-prone. In particular: search requests between two nodes with too different versions in cluster mode will fail. This, however, doesn't violate the contract of supporting backward compatibility & migration compatibility for one minor version.

### DIFF
--- a/lib/collection/src/recommendations.rs
+++ b/lib/collection/src/recommendations.rs
@@ -16,8 +16,7 @@ use crate::common::fetch_vectors::{convert_to_vectors, PointRef, ReferencedPoint
 use crate::operations::consistency_params::ReadConsistency;
 use crate::operations::types::{
     CollectionError, CollectionResult, CoreSearchRequest, CoreSearchRequestBatch, QueryEnum,
-    RecommendRequest, RecommendRequestBatch, RecommendStrategy, SearchRequest, SearchRequestBatch,
-    UsingVector,
+    RecommendRequest, RecommendRequestBatch, RecommendStrategy, UsingVector,
 };
 
 fn avg_vectors<'a>(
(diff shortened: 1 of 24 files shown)


lib/collection/src/recommendations.rs
```diff
@@ -172,12 +171,9 @@ where
...
     // In the future we'll fix this by unify them into CoreSearchRequests and make a single batch
     for (strategy, run) in batch_by_strategy(&request_batch.searches) {
-        let mut searches = Vec::new();
         let mut core_searches = Vec::new();
-        match strategy {
-            RecommendStrategy::AverageVector => searches.reserve_exact(run.len()),
-            RecommendStrategy::BestScore => core_searches.reserve_exact(run.len()),
-        }
+
+        core_searches.reserve_exact(run.len());
```
timvisee, line 176:
```rust
let core_searches = Vec::with_capacity(run.len());
```

timvisee (approved):

src/tonic/api/points_internal_api.rs
```diff
@@ -181,55 +181,24 @@ impl PointsInternal for PointsInternalService {
...
-            shard_id,
-        } = request.into_inner();
-
-        let mut search_points =
-            search_points.ok_or_else(|| Status::invalid_argument("SearchPoints is missing"))?;
-
-        search_points.read_consistency = None; // *Have* to be `None`!
-
-        search(self.toc.as_ref(), search_points, shard_id).await
+        return Err(Status::unimplemented(
```
coszio, line 186: are we going to delete this message in a later version too?

coszio (approved):
Made more cleanup refactors, specially in `recommendations.rs`. I think it is good to go now 👍

PR #2010, opened 2023-06-01
Read-only API Keys

amrhassan:
Superseded by <https://github.com/qdrant/qdrant/pull/2979>.

Hi,

This is an attempt to support read-only API keys for auth besides the current read-write ones, as described in #1959.

I've tried to keep it as contained as possible in the two middlewares for Tonic and Actix as requested. 

The main issue I faced was being able to tell the Read and Write APIs apart. I went with a very pragmatic approach of matching the RPC names in the case of gRPC and HTTP paths + methods in the case of REST against sane patterns. I could reconsider other approaches as well if you think this is not good enough or if there's a simpler way to do it that I'm missing.

There's a bit of code duplication in handling the keys because of the two different implementations, I could try to factor that out if you think it won't over-complicate it.

Finally, because of how the Actix middleware API is, there seems to be no easy way to avoid constructing the middleware if it's not used, but that was already the case before I introduced this change. I might look into avoiding paying the cost of that outside the scope of this PR.

Remaining work:
- [x] REST Integration tests
- [x] gRPC Integration tests
- [x] Use pytests instead
- [x] Same CI integration test runner

/claim #1959

Issue #1959, opened 2023-05-24
Read-only API key


generall:
**Is your feature request related to a problem? Please describe.**
Sometimes it is useful to be able to expose Qdrant instance for read/search queries.
Some use-case could be: search for web-sites backend-less. 

**Describe the solution you'd like**
Introduce API Key which only allows read operations on Qdrant instance.

API key should work for both: gRPC and REST endpoints, should only allow read subset of APIs.

API to allow:

- search (+batch), scroll, get-point, recommend, group-by inside collections
- list collections
- get collection info

**Describe alternatives you've considered**
 - Reverse proxy stuff
- Make API list configurable API to allow (this is too much work)

**Additional context**

- Should be implemented as a middle-ware (similar to how current API key works)
- Disabled by default, middleware should be only enabled if API key specified
- Read-only API key should be specified in the config 
- PR Should include tests

BabaBert:
@generall Should the middleware's check for different request headers or use the same one? If so, should the read only resources be ambiguous in regard to the used key?

timvisee:
> Should the middleware's check for different request headers or use the same one?

We want to reuse the same API key header.

> If so, should the read only resources be ambiguous in regard to the used key?

I'm not entirely sure what you mean, but if API key usage is enabled:

- A user uses either a full API key, read-only API key or no API key in the request header
- With both API keys, all read-only operations as listed in the first comment should work as normal, otherwise return an authentication error
- With the full API key, all other operations should work as normal, otherwise return an authentication error

Does that clear things up?

amrhassan:
Hi. I've made an attempt [here](https://github.com/qdrant/qdrant/pull/2010) with some dubious design decisions. Happy to discuss and take it further if necessary.

timvisee:
> but I assume the read only key would then be disabled too.

Good question.

No I think, if only a read key is set, only read operations should be allowed. That could be useful if some instance has a static dataset.

BabaBert:
Quick question, what is the status of this issue? I postponed my attempt as I assumed @amrhassan had a working solution.

amrhassan:
@BabaBert Yes the PR has a working solution, pending merging.

ologbonowiwi:
/attempt #1959

hey @amrhassan, if you don't mind, I'll take up from what you did

ologbonowiwi:
Hey @ttejas123, a PR to it is already open. 

I'm only waiting the contributors review.

PR #2979, opened 2023-11-12
Read-Only API keys

ologbonowiwi:
Hey guys!

I'm opening this one to finish the work that @amrhassan started on #2010.

/claim #1959
/split @amrhassan


src/actix/api_key.rs
```diff
@@ -3,20 +3,36 @@ use std::future::{ready, Ready};
...
 use actix_web::http::header::Header;
+use actix_web::http::Method;
 use actix_web::{Error, HttpResponse};
 use actix_web_httpauth::headers::authorization::{Authorization, Bearer};
-use constant_time_eq::constant_time_eq;
 use futures_util::future::LocalBoxFuture;
 
+use crate::common::auth::AuthKeys;
+
+const READ_ONLY_POST_PATTERNS: [&str; 11] = [
```
generall, line 13: would need to find a to somehow force this list updated, otherwise we will, for sure, forget to update it next time we introduce another API

ologbonowiwi: Do you think enum it'd work?

generall: I am not sure I understand how enum could help in this case

ologbonowiwi: Not sure if it's a good idea; I'm just throwing out brainstorming results to see our possibilities here.

Every solution I can think of requires us to set some parameters or new things manually.

The best option I can think about would be to add in the PR template to confirm that if new APIs were added, we should add the pattern here (as well on RPC_PATHS, if required)

ologbonowiwi: One option that I thought about would be to make the APIs return their patterns, but this is sub-optimal since it'd increase complexity

generall (approved):

PR #2984, opened 2023-11-13
Benchmark sparse index build

agourlay:
This PR adds benchmark for building the sparse index vector.

sparse-build

The takeaway is that:
- building the inverted index via repeated upsert operations is very fast 
- the binary search is the most expensive operation (annotated in source code)
- converting to mmap is very fast
- found a bug in computing vector count for mmap which I will fix in a next PR


IvanPleshkov (approved):
Clean PR with agreat diagram. Sweet!

PR #2992, opened 2023-11-13
Fix vector count in sparse index

agourlay:
This PR fixes the indexed vector count for the sparse vector index.

The are two bugs:
- empty sparse RAM index would report a size of 1 instead of 0
- loading a non empty memmap index would report a size of zero

The logic for keeping track of the number of indexed points has been pushed lower into the `InvertedIndex` trait to enable different implementations.

### DIFF
--- a/lib/segment/src/fixtures/sparse_fixtures.rs
+++ b/lib/segment/src/fixtures/sparse_fixtures.rs
@@ -84,6 +84,9 @@ pub fn fixture_sparse_index_ram<R: Rng + ?Sized>(
         num_vectors
     );
 
+    // assert no points are indexed following open for RAM index
+    assert_eq!(sparse_vector_index.indexed_vector_count(), 0);
+
     // build index to refresh RAM index
     sparse_vector_index.build_index(stopped).unwrap();
     assert_eq!(sparse_vector_index.indexed_vector_count(), num_vectors);
--- a/lib/segment/src/index/sparse_index/sparse_vector_index.rs
+++ b/lib/segment/src/index/sparse_index/sparse_vector_index.rs
@@ -30,7 +30,6 @@ pub struct SparseVectorIndex<TInvertedIndex: InvertedIndex> {
     path: PathBuf,
     pub inverted_index: TInvertedIndex,
     searches_telemetry: SparseSearchesTelemetry,
-    max_point_id: PointOffsetType, // used to compute the number of indexed vectors
 }
 
 impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
@@ -45,7 +44,6 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         create_dir_all(path)?;
 
         let searches_telemetry = SparseSearchesTelemetry::new();
-        let max_point_id = 0;
         let inverted_index = TInvertedIndex::open(path)?;
         let path = path.to_path_buf();
         let index = Self {
@@ -55,7 +53,6 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
             path,
             inverted_index,
             searches_telemetry,
-            max_point_id,
         };
         Ok(index)
     }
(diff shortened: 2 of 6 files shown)


lib/segment/src/fixtures/sparse_fixtures.rs
```diff
@@ -84,6 +84,9 @@ pub fn fixture_sparse_index_ram<R: Rng + ?Sized>(
         num_vectors
     );
 
+    // assert no points are indexed following open for RAM index
+    assert_eq!(sparse_vector_index.indexed_vector_count(), 0);
```
agourlay, line 88: this is the first bug, the assertion failed before the test

lib/segment/tests/integration/sparse_vector_index_search_tests.rs
```diff
@@ -162,6 +162,11 @@ fn sparse_vector_index_consistent_with_storage() {
     )
     .unwrap();
 
+    assert_eq!(
```
agourlay, line 165: this is the second bug, the assertion would fail before the fix

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -214,15 +213,12 @@ impl<TInvertedIndex: InvertedIndex> VectorIndex for SparseVectorIndex<TInvertedI
     }
 
     fn indexed_vector_count(&self) -> usize {
-        // internal ids start at 0
-        self.max_point_id as usize + 1
+        self.inverted_index.indexed_vector_count()
```
agourlay, line 216: now delegates to the underlying inverted index

lib/sparse/src/index/inverted_index/mod.rs
```diff
@@ -32,4 +32,7 @@ pub trait InvertedIndex {
     ) -> std::io::Result<Self>
     where
         Self: Sized;
+
+    /// Number of indexed vectors
+    fn indexed_vector_count(&self) -> usize;
```
IvanPleshkov, line 37: I guess we can rename it into `vector_count`. we already know here that vectors are indexed

lib/sparse/src/index/inverted_index/inverted_index_ram.rs
```diff
@@ -68,16 +78,23 @@ impl InvertedIndexRam {
...
                     // initialize new posting for dimension
                     self.postings[dim_id] = PostingList::new_one(id, weight);
                 }
             }
         }
+        // update indexed vector count by guessing whether we already have seen this point id
+        // given that there are no holes in the internal ids and that we are not deleting from the index
+        // we can just use the id as a proxy the count
+        if self.indexed_vector_count() < id as usize {
+            self.indexed_vector_count += 1;
```
IvanPleshkov, line 92: with `self.indexed_vector_count() < id` condition we cannot increment counter, we have to take `max`:
```
self.indexed_vector_count = max(self.indexed_vector_count, id);
```
Because counter will brake if you upsert points in non-sorted order

lib/segment/src/fixtures/sparse_fixtures.rs
IvanPleshkov: I like that you covered this changes in tests

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -198,7 +195,9 @@ impl<TInvertedIndex: InvertedIndex> VectorIndex for SparseVectorIndex<TInvertedI
             ram_index.upsert(id, vector.to_owned());
             index_point_count += 1;
         }
-        self.max_point_id = index_point_count.saturating_sub(1) as PointOffsetType;
+        // the underlying upsert operation does not guarantee that the indexed vector count is correct
+        // so we set the indexed vector count to the number of points we have seen
+        ram_index.indexed_vector_count = index_point_count;
```
IvanPleshkov, line 200: TBH, I don't like this approach. `ram_index.upsert` updates already this count, why do we need set it again?

Issue #2807, opened 2023-10-12
Tracking issue: user defined (aka custom) sharding


generall:
- https://github.com/qdrant/qdrant/pull/2808 - Introduce a `shard_key` field into the related structures, keep it Optional and do not serialize if None
- https://github.com/qdrant/qdrant/pull/2810 - On the collection, allow to store explicit list of shards instead of current number of shards.
- https://github.com/qdrant/qdrant/pull/2810 - Optional mapping of `shard_key` to id (list of ids) of the `shard_id` - on the collection level. 
- https://github.com/qdrant/qdrant/pull/2811 - Make sure consensus is able to synchronize this mapping
- https://github.com/qdrant/qdrant/pull/2810 - On the collection level: store which sharding strategy is used. Do not change the interface first, just allow to select internally between default and custom sharding strategies.
- https://github.com/qdrant/qdrant/pull/2811 Implement API for creating/deleting shards by keys
-  https://github.com/qdrant/qdrant/pull/2810 Enable choice of custom sharding strategy on collection creation
- https://github.com/qdrant/qdrant/pull/2909 - routing for points update requests
- routing for search/read requests 
- https://github.com/qdrant/qdrant/pull/2938 Make sure new shards have proper index created
- https://github.com/qdrant/qdrant/pull/3004 Update all point-level API to support optional shard_key selection
- https://github.com/qdrant/qdrant/pull/3018 Validation for querying/inserting into non-existing shards
- https://github.com/qdrant/qdrant/pull/3023 Propagate shard_key into search response
- https://github.com/qdrant/qdrant/pull/3027 Await for consensus operation to complete, so new shards should be available after request is OK 
- Implement API to change `shard_key` (post v1.7.x)

MudityaQQ:
**Hello @generall,**

I've been diving deep into the issue and analyzing the progress you've made by introducing the shard_key to the shard info API and the shard holder. I took into account the feedback from @timvisee regarding the efficiency of retrieving the shard ID to key mapping and have come up with two potential solution approaches:

**Approach 1: Direct Mapping**
1. Direct Map Structure: Instead of iterating over data to retrieve the mapping between shard keys and IDs, we can maintain two direct maps. One for mapping shard keys to IDs and the other for the reverse. This should optimize the retrieval process.

2. Shard Management API: I suggest the development of a dedicated API for shard management which would handle all CRUD operations for shards and their mappings.

[...]

PR #3004, opened 2023-11-14
Huge refactoring to make read requests aware of shard key selector

generall:
Tracking: https://github.com/qdrant/qdrant/issues/2807

This PR introduces shard key selector into read APIs.

ToDo:

- fill the shard key field of the result


timvisee (approved):
We did a pair review in a call. All looks good. :+1:

PR #3011, opened 2023-11-14
Improve sparse index pruning

agourlay:
This PR improves the sparse index pruning algorithm on several aspects.

- clearer implementation that makes each case explicit
- fixed two bugs related strict equality and wrong usage of `max_next_score`
- improved performance by replacing the posting lists sort by pushing the longest to the front.

Future work:
- support negative weights
- trigger pruning only if it makes sense

### DIFF
--- a/lib/sparse/src/index/search_context.rs
+++ b/lib/sparse/src/index/search_context.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering::Relaxed;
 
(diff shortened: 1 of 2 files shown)


lib/sparse/src/index/search_context.rs
```diff
@@ -109,14 +106,20 @@ impl<'a> SearchContext<'a> {
     /// Returns the next min record id from all posting list iterators
     ///
     /// returns None if all posting list iterators are exhausted
-    fn next_min(to_inspect: &[IndexedPostingListIterator<'_>]) -> Option<u32> {
+    fn next_min_id(to_inspect: &[IndexedPostingListIterator<'_>]) -> Option<u32> {
```
agourlay, line 109: renamed to make it clearer that it is an id and not a score

lib/sparse/src/index/search_context.rs
```diff
@@ -625,6 +666,89 @@ mod tests {
...
+        assert_eq!(
+            search_context.postings_iterators[0]
+                .posting_list_iterator
+                .len_to_end(),
+            0
+        );
+    }
+
+    #[test]
+    #[ignore] // TODO(sparse) make this test pass
```
agourlay, line 728: will be enabled in next PR with fix

IvanPleshkov (approved):

PR #3012, opened 2023-11-14
Abort shard transfer during `sync_local_state`, if transfer-er node failed (#2975)

### LINKED ISSUE #2975: Cancel shard transfer if shard is locally disabled
There is a situation possible in consensus:

- we have 2 replicas
- there is a currently running transfer from A(Active) -> B(Partial)
- Node A is killed
- Update is coming and it is written to B, but not to A
- A is reported as locally disabled and this change is actively propagated to consensus, but always declined due to "can't disable last active".
- reports to consensus generate a lot of spam in consensus WAL

Some limitations:

- If only `Partial` node receives the update, we can't ignore it. If the error is transient, the consistency of replicas will be compromised.
- We can't force disable A, cause in this case we won't have any active replicas

If we, in addition to reporting disabled state, will try to cancel all outbound transfers from the reported-as-dead replica,
the following scenario should happen:

- Node B reporting to the consensus 2 operations: cancel transfer and disable replica A
- First operation should be accepted, as there are no reason to not abort transfer
- Second operation will be rejected, as we can't disable last Active
- While shard transfer is aborted - it will change state of B from Partial -> Dead, that will clean the `locally_disabled` state and prevent further consensus spam 

Also this solution doesn't look too hacky, as cancelling transfer from dead shard is nothing but reasonable.

ffuugoo:
`sync_local_state` is called on every consensus tick (which is 100 ms by default). Under certain conditions, this may cause a cluster node to spam `SetShardReplicaState(Dead)` consensus proposals (that would ultimately be rejected) every 100 ms.

This PR extends `sync_local_state` to also abort shard transfer (if one exists) along with `SetShardReplicaState(Dead)` notification, which (once propagated through consensus and applied on this node) would partially/temporarily disable `SetShardReplicaState(Dead)` spam. (And further improvements in #3013 and #3026 makes this a complete/permanent fix for `SetShardReplicaState(Dead)` spam.)

Resolves #2975.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -450,15 +450,18 @@ impl Collection {
     ) -> CollectionResult<()> {
         // Check for disabled replicas
         let shard_holder = self.shards_holder.read().await;
-        let get_shard_transfer = |key| shard_holder.get_transfer(&key);
+
+        let get_shard_transfers = |shard_id, from| {
+            shard_holder
+                .get_transfers(|transfer| transfer.shard_id == shard_id && transfer.from == from)
+        };
+
         for replica_set in shard_holder.all_shards() {
-            replica_set.sync_local_state(get_shard_transfer).await?;
+            replica_set.sync_local_state(get_shard_transfers).await?;
         }
 
         // Check for un-reported finished transfers
-        let outgoing_transfers = shard_holder
-            .get_outgoing_transfers(&self.this_peer_id)
-            .await;
+        let outgoing_transfers = shard_holder.get_outgoing_transfers(&self.this_peer_id);
         let tasks_lock = self.transfer_tasks.lock().await;
         for transfer in outgoing_transfers {
             match tasks_lock.get_task_result(&transfer.key()) {
@@ -521,7 +524,7 @@ impl Collection {
             }
 
             // Try to find dead replicas with no active transfers
-            let transfers = shard_holder.get_transfers(|_| true).await;
+            let transfers = shard_holder.get_transfers(|_| true);
 
             // Try to find a replica to transfer from
             for replica_id in replica_set.active_remote_shards().await {
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -452,11 +452,6 @@ impl Collection {
         let shard_holder = self.shards_holder.read().await;
         let get_shard_transfer = |key| shard_holder.get_transfer(&key);
         for replica_set in shard_holder.all_shards() {
-            // TODO:
-            //   Aborting shard transfer here doesn't *really* work, if we keep requesting new
-            //   transfer in `Collection::request_shard_transfer` (at the bottom).
-            //
-            //  Shard transfer should be initiated by the *sender*.
             replica_set.sync_local_state(get_shard_transfer).await?;
         }
 
@@ -547,12 +542,6 @@ impl Collection {
                     this_peer_id,
                     replica_id
                 );
-
-                // TODO:
-                //   Aborting shard transfer in `ShardReplicaSet::sync_local_state` (at the top)
-                //   doesn't *really* work, if we keep requesting new transfer here.
-                //
-                //  Shard transfer should be initiated by the *sender*.
                 self.request_shard_transfer(transfer);
                 break;
             }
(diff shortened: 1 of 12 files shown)


lib/collection/src/collection/mod.rs
```diff
@@ -442,8 +449,14 @@ impl Collection {
...
         let shard_holder = self.shards_holder.read().await;
+        let get_shard_transfer = |key| shard_holder.get_transfer(&key);
         for replica_set in shard_holder.all_shards() {
-            replica_set.sync_local_state().await?;
+            // TODO:
+            //   Aborting shard transfer here doesn't *really* work, if we keep requesting new
+            //   transfer in `Collection::request_shard_transfer` (at the bottom).
+            //
+            //  Shard transfer should be initiated by the *sender*.
+            replica_set.sync_local_state(get_shard_transfer).await?;
```
ffuugoo, line 459: - we abort shard transfer, if transfer-receiver detects that transfer-sender failed

ffuugoo (commented):
@generall we missed a minor detail. :/

(I mistyped @generall when sending comment and might have pinged some random person on GH. Sorry! 😅)

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -655,9 +662,28 @@ impl ShardReplicaSet {
...
-    pub async fn sync_local_state(&self) -> CollectionResult<()> {
-        for failed_peer in self.locally_disabled_peers.read().iter() {
-            self.notify_peer_failure(*failed_peer);
+    pub async fn sync_local_state<F>(&self, get_shard_transfer: F) -> CollectionResult<()>
+    where
+        F: Fn(ShardTransferKey) -> Option<ShardTransfer>,
+    {
+        for &failed_peer in self.locally_disabled_peers.read().iter() {
+            self.notify_peer_failure(failed_peer);
+
```
timvisee, line 671: Maybe it's worth to add a comment here:

```rust
// Abort any transfers from a failed peer to prevent spamming `SetShardReplicaState(Dead)`
```

or something similar.

timvisee (approved):

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -655,9 +662,28 @@ impl ShardReplicaSet {
...
+            let key = ShardTransferKey {
+                shard_id: self.shard_id,
+                from: failed_peer,
+                to: self.this_peer_id(),
```
generall, line 675: I am not sure it should be `this_peer_id`. Failure can be reported by the third node, which may not have any shards at all. I think we should cancel all transfers from `failed_peer` no matter destination.

ffuugoo: But in that case, we would be cancelling shard transfers of _other_ shards from `failed_peer`, right? Cause there can't be multiple shard transfers of `self.shard_id` from `failed_peer`. (Also, not sure if there can be two different shard transfers from `failed_peer` to `this_peer_id`. If so, then this is even worse.)

Seems safer to only cancel `self.shard_id`/`this_peer_id` transfer. Like, if `failed_peer` _is_ available, but under high load, and it only fail _sometimes_, cancelling (and then restarting) all shard transfers will probably make everything even worse. If we only cancel `self.shard_id`/`this_peer_id`, then there is a chance that at least some of the transfers may finish successfully.

generall: >  And if all shard transfers fail, then each shard would cancel its own transfer.

Local state doesn't exist on all peers, the situation possible, where there are 3 nodes: A, B, C

The transfer A->B happens and A shuts down. Replica only exist on A and B, C is empty.
But the update can still come to C, so local state will exist on C.
In this situation we still want to cancel transfer from A->B.
No other node won't be able to cancel transfer, cause no one else knows that the inconsistency exist.

timvisee: Another thing:
If we would cancel all transfers from a node, wouldn't it be possible for a badly behaving node to constantly break _all_ ongoing transfers while they would otherwise succeed.

Badly behaving node as in: flaky network connection or something similar.

I'd expect that only cancelling A->B transfers would be a safer bet then.

ffuugoo: Done.

generall (approved):

PR #3016, opened 2023-11-14
Fix sparse index build benchmark compilation

agourlay:
This PR fixes the issue of broken clippy on dev.

Two PRs conflicting with each other got merged without an intermediary rebase causing compilation issue.


lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -50,7 +50,7 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         // load config
         let config_path = SparseIndexConfig::get_config_path(path);
         let config = if config_path.exists() {
-            SparseIndexConfig::load(path)?
+            SparseIndexConfig::load(&config_path)?
```
agourlay, line 53: this one is actually a proper bug discovered while fixing the benchmark :)

IvanPleshkov (approved):

PR #3023, opened 2023-11-15
fill read response with shard_key

generall:
Tracking: https://github.com/qdrant/qdrant/issues/2807

If shard key is available, changes in this PR will propagate them into the response

image

### DIFF
--- a/lib/collection/src/shards/shard_holder.rs
+++ b/lib/collection/src/shards/shard_holder.rs
@@ -41,6 +41,9 @@ pub struct ShardHolder {
     pub(crate) shard_transfers: SaveOnDisk<HashSet<ShardTransfer>>,
     rings: HashMap<Option<ShardKey>, HashRing<ShardId>>,
     key_mapping: SaveOnDisk<ShardKeyMapping>,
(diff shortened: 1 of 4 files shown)


lib/collection/src/shards/shard_holder.rs
```diff
@@ -297,23 +308,26 @@ impl ShardHolder {
...
         match shard_selector {
             ShardSelectorInternal::Empty => {
                 debug_assert!(false, "Do not expect empty shard selector")
             }
             ShardSelectorInternal::All => {
-                res.extend(self.all_shards());
+                for (shard_id, shard) in self.shards.iter() {
+                    let shard_key = self.shard_id_to_key_mapping.get(shard_id);
+                    res.push((shard, shard_key));
+                }
```
timvisee, line 325: Do we do this so that we have lock-free access?

timvisee (approved):

lib/collection/src/shards/shard_holder.rs
generall: lock-free and clone-free

PR #3025, opened 2023-11-15
Splade benchmark example

agourlay:
This PR is not be merged but to show how to do a quick experimentation with SPLADE embeddings.

```
cargo bench -p segment --bench sparse_index_splade
   Compiling segment v0.6.0 (/home/agourlay/Workspace/qdrant/lib/segment)
    Finished bench [optimized] target(s) in 5.64s
     Running benches/sparse_index_splade.rs (target/release/deps/sparse_index_splade-9ae7079638a4c37e)
Data size: 264 mb
34880 vectors
30265 max dimensions
Loading data...
Data loaded in 7539 ms
query sparse vector size = 300
TOP = 100
sparse-vector-splade-group/inverted-index
                        time:   [29.237 ms 29.447 ms 29.695 ms]
                        change: [-1.6655% +0.0957% +1.8522%] (p = 0.92 > 0.05)
                        No change in performance detected.

```

~~One learning is that we may remove the constraint regarding having the indices sorted as it is annoying for the users of SPLADE.~~

Overview where the time is spent.

splade-bench

### DIFF
--- a/lib/segment/Cargo.toml
+++ b/lib/segment/Cargo.toml
@@ -121,3 +121,8 @@ harness = false
 [[bench]]
 name = "sparse_index_build"
 harness = false
+
+[[bench]]
+name = "sparse_index_splade"
+harness = false
+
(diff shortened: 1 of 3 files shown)


agourlay:
not meant to be merged

Issue #2976, opened 2023-11-10
Report error to user if no active shards were updated


generall:
**Is your feature request related to a problem? Please describe.**
Currenly, update operation sends updates to the following states:

```
   Some(ReplicaState::Active) => true,
   Some(ReplicaState::Partial) => true,
   Some(ReplicaState::Initializing) => true,
   Some(ReplicaState::Dead) => false,
   Some(ReplicaState::Listener) => true,
   Some(ReplicaState::PartialSnapshot) => false,
```

However if update operation is successful only on `Partial` shards, it is not enough for cluster to ensure persistence of the change.

**Describe the solution you'd like**

Update operation should fail if only partial shards were updated

PR #3026, opened 2023-11-15
Add a health check before starting shard transfer during `sync_local_state` (#2977)

### LINKED ISSUE #2977: Check node pings before requesting transfer
It might be possible, that cluster ends up in the following situation:

2 replicas
Node A: Alive, marked as DEAD
Node B: Dead, marked as ALIVE

In this situation, node A will constantly try to request shard transfer from B, but his transfer.
If, at the same time, we will continue push updates, it might create a stream of shard transfer creation and cancellations.

Before requesting shard transfer in the consensus, Node A should try to ping Node B to ensure it is alive.
If node B is not alive, it will allow Node A to either look for another possible sources of transfer, or, at least, do not spam doomed requests to the consensus.

ffuugoo:
This PR adds explicit "health check" before requesting shard transfer, to see if the other node is _available_. 

It is possible for the cluster to be in a state, when the last `Active` node is offline/unavailable, and without this check `Dead` node may request shard transfer (and switch into `Partial`) a bit too soon, which may then enable a bunch of other issues (e.g., #2975/#3012 and #2976/#3013).

Resolves #2977. Currently based on #3013, will be rebased on `dev` once it's merged.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -535,9 +535,25 @@ impl Collection {
                     sync: true,
                     method: None,
                 };
+
                 if check_transfer_conflicts_strict(&transfer, transfers.iter()).is_some() {
                     continue; // this transfer won't work
                 }
+
+                // TODO: Should we, maybe, throttle/backoff this requests a bit?
+                if let Err(err) = replica_set.health_check(replica_id).await {
+                    // TODO: This is rather verbose, not sure if we want to log this at all... :/
+                    log::trace!(
+                        "Replica {replica_id}/{}:{} is not available \
+                         to request shard transfer from: \
+                         {err}",
+                        self.id,
+                        replica_set.shard_id,
+                    );
+
+                    continue;
+                }
+
                 log::debug!(
                     "Recovering shard {}:{} on peer {} by requesting it from {}",
                     self.name(),
@@ -545,6 +561,7 @@ impl Collection {
                     this_peer_id,
                     replica_id
                 );
+
                 self.request_shard_transfer(transfer);
                 break;
             }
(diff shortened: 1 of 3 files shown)


lib/collection/src/collection/mod.rs
```diff
@@ -531,16 +531,33 @@ impl Collection {
...
                     method: None,
                 };
+
                 if check_transfer_conflicts_strict(&transfer, transfers.iter()).is_some() {
                     continue; // this transfer won't work
                 }
+
+                // TODO: Should we, maybe, throttle/backoff this requests a bit?
+                if let Err(err) = replica_set.health_check(replica_id).await {
+                    // TODO: This is rather verbose, not sure if we want to log this at all... :/
```
generall, line 541: Looks ok for a `trace` log

lib/collection/src/collection/mod.rs
```diff
@@ -531,16 +531,33 @@ impl Collection {
                     sync: true,
                     method: None,
                 };
+
                 if check_transfer_conflicts_strict(&transfer, transfers.iter()).is_some() {
                     continue; // this transfer won't work
                 }
+
+                // TODO: Should we, maybe, throttle/backoff this requests a bit?
```
generall, line 539: In current implementation it doesn't create any additional requests, in case of the offline peer. It just replaces request to consensus with this `health_check`. So I think it is ok to keep this part as-is in this PR

generall (approved):

lib/collection/src/collection/mod.rs
```diff
@@ -531,16 +531,33 @@ impl Collection {
...
+                if let Err(err) = replica_set.health_check(replica_id).await {
```
timvisee, line 540: Do we eventually want to improve this call with something that actually checks on the remote that the shard is ready and active?

Or do we solely want to depend on consensus to synchronize this state, and leave this call as simple as possible?

I assume the latter, right?

timvisee (approved):

PR #3030, opened 2023-11-16
Remove sorted constraint on sparse vectors

agourlay:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/data_types/named_vectors.rs
+++ b/lib/segment/src/data_types/named_vectors.rs
@@ -146,7 +146,10 @@ impl<'a> NamedVectors<'a> {
                     let preprocessed_vector = distance.preprocess_vector(v.to_vec());
                     *vector = CowValue::Dense(Cow::Owned(preprocessed_vector))
                 }
-                CowValue::Sparse(_) => {}
+                CowValue::Sparse(v) => {
+                    // sort by indices to enable faster dot product and overlap checks
+                    v.to_mut().sort_by_indices();
+                }
             }
         }
     }
--- a/lib/sparse/src/common/sparse_vector.rs
+++ b/lib/sparse/src/common/sparse_vector.rs
@@ -9,7 +9,8 @@ use crate::common::types::{DimId, DimWeight};
 /// - indices to be unique
 /// - indices and values to be the same length
 /// - indices and values to be non-empty
-/// - indices to be sorted TODO(sparse) this is not validated yet
+/// - values can be any f32
+/// - indices do not need to sorted (they are sorted on persistence)
 #[derive(Debug, PartialEq, Clone, Default, Serialize, Deserialize)]
 pub struct SparseVector {
     pub indices: Vec<DimId>,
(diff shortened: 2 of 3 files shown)


lib/segment/src/data_types/named_vectors.rs
```diff
@@ -146,7 +146,10 @@ impl<'a> NamedVectors<'a> {
                     let preprocessed_vector = distance.preprocess_vector(v.to_vec());
                     *vector = CowValue::Dense(Cow::Owned(preprocessed_vector))
                 }
-                CowValue::Sparse(_) => {}
+                CowValue::Sparse(v) => {
+                    // sort by indices to enable faster dot product and overlap checks
+                    v.to_mut().sort_by_indices();
+                }
```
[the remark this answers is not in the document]
NirantK: Never mind, I saw that this function handles that case!

lib/sparse/src/common/sparse_vector.rs
```diff
@@ -147,7 +177,7 @@ mod tests {
         assert!(different_length.is_err());
 
         let not_sorted = SparseVector::new(vec![1, 3, 2], vec![1.0, 2.0, 3.0]);
-        assert!(not_sorted.is_err());
+        assert!(not_sorted.is_ok());
```
NirantK, line 180: Do we want to hint the user that if their indices are pre-sorted, this indexing operation will be much faster?

generall: honestly, sorting 100 integers is a 100 nanosecod operation

NirantK: Aaah. At a few ns*1000 → total a few ms, this is perfectly fine. Thanks!

lib/sparse/src/common/sparse_vector.rs
```diff
@@ -9,7 +9,8 @@ use crate::common::types::{DimId, DimWeight};
 /// - indices to be unique
 /// - indices and values to be the same length
 /// - indices and values to be non-empty
-/// - indices to be sorted TODO(sparse) this is not validated yet
+/// - values can be any f32
```
IvanPleshkov, line 12: This comment line may be unnecessary. Even more, `NaN` is a valid `f32` value but we don't expect nan (and +- inf too)

lib/segment/src/data_types/named_vectors.rs
```diff
@@ -146,7 +146,10 @@ impl<'a> NamedVectors<'a> {
                     let preprocessed_vector = distance.preprocess_vector(v.to_vec());
                     *vector = CowValue::Dense(Cow::Owned(preprocessed_vector))
                 }
-                CowValue::Sparse(_) => {}
+                CowValue::Sparse(v) => {
```
IvanPleshkov, line 149: I find this place of sorting very untrivial. Can we do sort in `SparseVectors::new` before validation?

IvanPleshkov: And do sorting while serde deserialization

lib/sparse/src/common/sparse_vector.rs
agourlay: Sure I can remove it.

lib/segment/src/data_types/named_vectors.rs
agourlay: There is indeed a problem with the current implementation.
It only makes sure that the vector are sorted on insertion and updates.

I still need to add a sorting instruction for the query vectors when they reach the `SparseRawScorer` :hammer:

IvanPleshkov (approved):

Issue #3005, opened 2023-11-14
Flaky test_recover_from_snapshot_2


agourlay:
https://github.com/qdrant/qdrant/actions/runs/6859970392/job/18653056947?pr=2994#step:10:989

```
=========================== short test summary info ============================
FAILED tests/consensus_tests/test_snapshot_recovery.py::test_recover_from_snapshot_2 - AssertionError: assert {'id': 68, 'p...r': None, ...} == {'id': 68, 'p...r': None, ...}
  Omitting 4 identical items, use -vv to show
  Differing items:
  {'version': 1} != {'version': 2}
  Full diff:
    {
     'id': 68,
     'payload': {'city': 'London'},
     'score': 0.7719247,
     'vector': None,
  -  'version': 2,
  ?             ^
  +  'version': 1,
  ?             ^
    }
============= 1 failed, 33 passed, 1 skipped in 535.24s (0:08:55) ==============
Error: Process completed with exit code 1.
```

agourlay:
https://github.com/qdrant/qdrant/actions/runs/6860045949/job/18653239661?pr=2995#step:10:1202

timvisee:
Fixed in <https://github.com/qdrant/qdrant/pull/3035>

PR #3031, opened 2023-11-16
Fix flaky snapshot recovery test

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/3005>.

Our snapshot recovery test recently became flaky and started failing most of the time. This was due to search results having different point versions after a snapshot was recovered, while before the version was always the same.

I bisected <https://github.com/qdrant/qdrant/pull/2938> to have introduced this problem. More specifically in [3b5920b^...243da10](https://github.com/qdrant/qdrant/compare/3b5920b07b27412e45739d04af1b7d1394c69046%5E...243da10fc1c36aecef8fdf3cd669f6f5ff0d78bb).

In <https://github.com/qdrant/qdrant/pull/2938> we've restructured the payload schema definition to be part of both segments and the collection. For example, to create a new payload index, we now internally use two operations to update the schema on both levels. It causes the point version number to be one higher.

I've updated the test to support this; version numbers may now be off by one.

If we eventually remove payload schema definitions from the segment level and go back to a single operation, we can go back to doing exact version matches again.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/storage/src/content_manager/snapshots/recover.rs
+++ b/lib/storage/src/content_manager/snapshots/recover.rs
@@ -179,17 +179,17 @@ async fn _do_recover_from_snapshot(
     for (shard_id, shard_info) in &state.shards {
         let shards = latest_shard_paths(tmp_collection_dir.path(), *shard_id).await?;
 
-        let snapshot_shard_path = shards
-            .into_iter()
-            .filter_map(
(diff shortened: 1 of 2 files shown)


timvisee:
Fixed in <https://github.com/qdrant/qdrant/pull/3035> with a different approach which is better.

If we wait for both nodes to reach the same consensus the version numbers will be the same.

PR #3039, opened 2023-11-17
Default `max_values_per_point` to zero, add tests

timvisee:
I took a deeper look into the `max_values_per_point` field of index types and what the purpose of it is.

This PR:
- defaults `max_values_per_point` to `0` rather than `1`
- fixes potential divide by zero panic for the geo index if it contained no items
- adds some unit tests to ensure indexes work properly if it doesn't contain points
- shortcuts cardinality estimation if `max_values_per_point` is `0`

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/index/field_index/geo_index.rs
+++ b/lib/segment/src/index/field_index/geo_index.rs
@@ -472,6 +472,11 @@ impl GeoMapIndex {
     }
 
     pub fn match_cardinality(&self, values: &[GeoHash]) -> CardinalityEstimation {
+        let max_values_per_point = self.max_values_per_point();
+        if max_values_per_point == 0 {
+            return CardinalityEstimation::exact(0);
+        }
+
         let common_hash = common_hash_prefix(values);
 
         let total_points = self.get_points_of_hash(&common_hash);
--- a/lib/segment/src/index/field_index/geo_index.rs
+++ b/lib/segment/src/index/field_index/geo_index.rs
@@ -1307,4 +1307,81 @@ mod tests {
         assert_eq!(new_index.points_count(), 1);
         assert_eq!(new_index.points_values_count(), 2);
     }
+
+    #[test]
+    fn test_empty_index_cardinality() {
+        let polygon = GeoPolygon {
+            exterior: GeoLineString {
+                points: vec![
+                    GeoPoint {
+                        lon: 19.415558242000287,
+                        lat: 69.18533258102943,
+                    },
+                    GeoPoint {
+                        lon: 2.4664944437317615,
+                        lat: 61.852748225727254,
+                    },
+                    GeoPoint {
+                        lon: 2.713789718828849,
+                        lat: 51.80793869181895,
+                    },
+                    GeoPoint {
+                        lon: 19.415558242000287,
+                        lat: 69.18533258102943,
+                    },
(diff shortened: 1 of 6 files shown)


lib/segment/src/index/field_index/geo_index.rs
```diff
@@ -467,6 +472,11 @@ impl GeoMapIndex {
     }
 
     pub fn match_cardinality(&self, values: &[GeoHash]) -> CardinalityEstimation {
+        let max_values_per_point = self.max_values_per_point();
+        if max_values_per_point == 0 {
+            return CardinalityEstimation::exact(0);
+        }
```
timvisee, line 478: Not 100% sure if we can shortcut here. It would be awesome if someone else could confirm this.

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -155,14 +155,24 @@ impl<T: Encodable + Numericable> NumericIndex<T> {
...
             NumericIndex::Mutable(index) => index.max_values_per_point,
             NumericIndex::Immutable(index) => index.max_values_per_point,
         }
     }
 
     fn range_cardinality(&self, range: &Range) -> CardinalityEstimation {
+        let max_values_per_point = self.max_values_per_point();
+        if max_values_per_point == 0 {
+            return CardinalityEstimation::exact(0);
+        }
```
timvisee, line 174: Not 100% sure if we can shortcut here. It would be awesome if someone else could confirm this.

generall (approved):

lib/segment/src/index/field_index/geo_index.rs
```diff
@@ -481,7 +491,7 @@ impl GeoMapIndex {
 
         // Assume all selected points have `max_values_per_point` value hits.
         // Therefore number of points can't be less than `total_hits / max_values_per_point`
-        let min_hits_by_value_groups = sum / self.max_values_per_point();
+        let min_hits_by_value_groups = sum.checked_div(max_values_per_point).unwrap_or(0);
```
IvanPleshkov, line 494: you have already checked zero condition, do we need `checked_div` here? I find `unwrap_or(0)` confusing because i cannot say why `0`

lib/segment/src/index/field_index/geo_index.rs
IvanPleshkov: Looks reasonable. if there is no any value in index, cardinality is exact zero.

lib/segment/src/index/field_index/numeric_index/mod.rs
IvanPleshkov: looks good like in geo index

IvanPleshkov (approved):

lib/segment/src/index/field_index/geo_index.rs
timvisee: Smart! 

I added the `== 0` shortcut later so I forgot about this. I'll remove it since it is redundant.

timvisee: Done in 5cbc795f1840da4c933a0cf5554fc626c9e56197.

I'll merge when all is :green_circle: .

PR #3043, opened 2023-11-18
alter deactivation behavior for write_consistency_factor

generall:
If user specified `write_consistency_factor = N`, it would mean, that failure to apply operation to `N` replicas will result in API failure (e.g. http 500).

This PR changes logic of shard deactivation to prevent reporting shards as dead in case if the whole operation is considered failed.
This logic might be desirable, cause it should create overall less amount of expensive shards transfers. 
In case write operation failed, it would be users responsibility to retry it and resolve storage consistency by re-trying the update operation.

Practical examples:

- replication factor 2, write_consistency_factor = 2

In this case update operation will never cause dead shards, but update will fail if any replica fail. 
Might be useful to scale throughput and eliminate SPOF for reads in the installation where writes can be delayed and retried.

- replication factor 3,  write_consistency_factor = 2

In this case max number of dead shards is 1. Useful in cases, where we want to have minimal number of available read replicas, as the same time we assume that 1 machine can fail and we still can recover it without interrupting writes.

### DIFF
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -185,13 +185,23 @@ impl ShardReplicaSet {
 
         let total_results = all_res.len();
 
+        let write_consistency_factor = self
+            .collection_config
+            .read()
+            .await
+            .params
+            .write_consistency_factor
+            .get() as usize;
+
+        let minimal_success_count = write_consistency_factor.min(total_results);
+
         let (successes, failures): (Vec<_>, Vec<_>) = all_res.into_iter().partition_result();
 
         // Notify consensus about failures if:
         // 1. There is at least one success, otherwise it might be a problem of sending node
         // 2. ???
 
-        if !successes.is_empty() {
+        if successes.len() >= minimal_success_count {
             let wait_for_deactivation =
                 self.handle_failed_replicas(&failures, &self.replica_state.read());
             // report all failing peers to consensus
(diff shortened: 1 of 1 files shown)


lib/collection/src/shards/replica_set/update.rs
```diff
@@ -185,13 +185,23 @@ impl ShardReplicaSet {
...
+        let minimal_success_count = write_consistency_factor.min(total_results);
+
         let (successes, failures): (Vec<_>, Vec<_>) = all_res.into_iter().partition_result();
 
         // Notify consensus about failures if:
         // 1. There is at least one success, otherwise it might be a problem of sending node
         // 2. ???
 
-        if !successes.is_empty() {
+        if successes.len() >= minimal_success_count {
```
timvisee, line 204: Nitpick, but maybe we can assign this condition to a boolean variable, so we simplify our if statement conditions a bit.

```rust
let got_enough_successes = successes.len() >= minimal_success_count;

if got_enough_successes {}

if !failures.is_empty() && !got_enough_successes {}
```

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -228,20 +238,10 @@ impl ShardReplicaSet {
...
-                .params
-                .write_consistency_factor
-                .get() as usize;
-            let minimal_success_count = write_consistency_factor.min(total_results);
-            if successes.len() < minimal_success_count {
-                // completely failed - report error to user
-                let (_peer_id, err) = failures.into_iter().next().expect("failures is not empty");
-                return Err(err);
-            }
+        if !failures.is_empty() && successes.len() < minimal_success_count {
```
timvisee, line 241: Could we add a comment before here to explain why we test these two conditions exactly?

timvisee (approved):

PR #3048, opened 2023-11-20
Better heuristic for sparse vector pruning

agourlay:
This PR introduces a simple heuristic to drastically decrease the number of pruning operations during sparse vector search.

Instead of trying to prune on all search candidates, we only try to prune when the worst best score changes.
This is a good signal that we should try pruning again as the eviction threshold increased.

This increases performance significantly on our synthetic benchmark.

```
sparse_vector size = 196
sparse-vector-search-group/inverted-index
                        time:   [15.326 ms 15.374 ms 15.422 ms]
                        change: [-42.208% -41.984% -41.762%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/lib/segment/tests/integration/sparse_vector_index_search_tests.rs
+++ b/lib/segment/tests/integration/sparse_vector_index_search_tests.rs
@@ -22,10 +22,10 @@ use sparse::index::inverted_index::InvertedIndex;
 use tempfile::Builder;
 
 /// Max dimension of sparse vectors used in tests
-const MAX_SPARSE_DIM: usize = 2048;
+const MAX_SPARSE_DIM: usize = 4096;
 
 /// Number of vectors to index in tests
-const NUM_VECTORS: usize = 1000;
+const NUM_VECTORS: usize = 2000;
 
 /// Default full scan threshold in tests
 /// very low value to force usage of index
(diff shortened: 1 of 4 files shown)


lib/segment/tests/integration/sparse_vector_index_search_tests.rs
```diff
@@ -22,10 +22,10 @@ use sparse::index::inverted_index::InvertedIndex;
 use tempfile::Builder;
 
 /// Max dimension of sparse vectors used in tests
-const MAX_SPARSE_DIM: usize = 2048;
+const MAX_SPARSE_DIM: usize = 4096;
```
agourlay, line 25: Took the opportunity to increase the size of the test data set given the recent performance improvements.

lib/segment/tests/integration/sparse_vector_index_search_tests.rs
```diff
@@ -82,8 +82,14 @@ fn compare_sparse_vectors_search_with_without_filter(full_scan_threshold: usize)
             .iter()
             .zip(index_results_no_filter.iter())
         {
-            assert_eq!(filter_result.len(), top);
```
agourlay, line 85: this was was actually incorrect, there can be zero match

IvanPleshkov (approved):

PR #3057, opened 2023-11-20
Accept empty sparse vectors

agourlay:
This PR removes the validation that a sparse vector must be non-empty.

The rational is that it is overly restrictive given that:
- embeddings might generate such vectors
- users might want to only attach payload to a sparse vector

Instead of forcing more work onto the user, we decided to loosen up that constraint.

Empty sparse vectors cannot be indexed but can be found in storage through the scrolling API.

### DIFF
--- a/lib/sparse/src/index/inverted_index/inverted_index_ram.rs
+++ b/lib/sparse/src/index/inverted_index/inverted_index_ram.rs
@@ -69,13 +69,7 @@ impl InvertedIndexRam {
     }
 
     /// Upsert a vector into the inverted index.
-    ///
-    /// Empty vectors are ignored.
     pub fn upsert(&mut self, id: PointOffsetType, vector: SparseVector) {
-        // do not upsert empty vectors
-        if vector.is_empty() {
-            return;
-        }
         for (dim_id, weight) in vector.indices.into_iter().zip(vector.values.into_iter()) {
             let dim_id = dim_id as usize;
             match self.postings.get_mut(dim_id) {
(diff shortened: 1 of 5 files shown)


coszio (approved):

lib/sparse/src/index/inverted_index/inverted_index_ram.rs
```diff
@@ -69,7 +69,13 @@ impl InvertedIndexRam {
     }
 
     /// Upsert a vector into the inverted index.
+    ///
+    /// Empty vectors are ignored.
     pub fn upsert(&mut self, id: PointOffsetType, vector: SparseVector) {
+        // do not upsert empty vectors
+        if vector.is_empty() {
```
IvanPleshkov, line 76: I propose to move this check to `SparseVectorIndex::update_vector`. I expect that only one entity is responsible for skipping empty vectors. It can be `SparseVectorIndex` or `InvertedIndex`. Because `build` function has already a skipping, I propose to move this skip

IvanPleshkov (approved):
LGTM, just one small optional comment

lib/sparse/src/index/inverted_index/inverted_index_ram.rs
IvanPleshkov: You may also have two checks: here and in `SparseVectorIndex::update_vector` both

agourlay:
pushed a fix, waiting for Github to finally show the commit in the UI :hourglass_flowing_sand:

lib/sparse/src/index/inverted_index/inverted_index_ram.rs
agourlay: I like the idea of having a single responsible entity for skipping :+1: 
https://github.com/qdrant/qdrant/pull/3057/commits/d8281a738793e16e22e4c3c8bf33ff67a5b78da3

PR #3068, opened 2023-11-20
chore: bump to tonic 0.10.2 and prost to 0.12.2

coszio:
- https://github.com/hyperium/tonic/pull/1401 has been merged, so we should be able to bump tonic and prost versions now. 

- `from_32` was deprecated in favor of `TryFrom::try_from`, so that's the other source of changes here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -480,9 +480,9 @@ dependencies = [
  "env_logger",
  "log",
  "parking_lot",
- "prost 0.11.9",
+ "prost 0.12.2",
  "prost-build 0.11.9",
- "prost-types 0.11.9",
+ "prost-types 0.12.2",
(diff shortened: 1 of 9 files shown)


timvisee (approved):

agourlay:
The Raft crate is still not compatible with prost 0.12.0 AFAIK.
This means now our gRPC infra. and our consensus infra. are running on different prost minor version.

It looks fine but I guess we should be careful there.
WDYT?

timvisee:
Ouch, missed that bit. Thanks for the heads up.

In that case it might actually be better to revert and wait for `raft` to reach `prost 0.12.0` as well.

agourlay:
I would prefer to revert to keep things clearer :+1:

PR #3074, opened 2023-11-21
cargo +nightly fmt

coszio:
Something's changed in latest nightly rustfmt. All these changes are only from `cargo +nightly fmt` on `dev`.

An alternative solution would be to ditch the nightly formatting, but I'm not sure which is option is better.


timvisee:
It seems that our CI formatter/linter currently fails on this.

I think we should wait a little and see if this remains the case. It may very well be an accidental change in nightly that gets reverted tomorrow.

timvisee (commented):
It seems that our CI formatter/linter currently fails on this.

I think we should wait a little and see if this remains the case. It may very well be an accidental change in nightly that gets reverted tomorrow.

We may merge when CI starts failing without these changes.

coszio:
> It may very well be an accidental change in nightly that gets reverted tomorrow. 

Hopefully 🤞

ffuugoo:
https://github.com/rust-lang/rustfmt/issues/5964

PR #3084, opened 2023-11-23
Extend `/readyz` with shards readiness check (#3053)

### LINKED ISSUE #3053: Consensus-aware health check request
Currently, while doing rolling update, there is a possible situation when 
next instance is restarted before completing synchronization with previously restarted nodes.

This situation might cause downtime of the service for the period of node restart, which is problematic for some deployments.

Create a new health-check endpoint (or augment existing one), which would return 200 only in the current node doesn't 
have dead/partial shards.
It should also consider the status of consensus and do not return OK if consensus operations are delayed.

???

**Additional context**

We should decide if it is safe enough to rely on such a condition in case of cluster deployment.
There are 2 potentially problematic scenarios which we should be especially careful about:

- Condition might be never satisfied: that would cause the indefinite await of the cluster restart. So it might be a good idea to introduce fail-safe timeout
- False-positive 200. it might be caused by late synchronization with the consensus. In this case rolling update may trigger downtime on both search and update

ffuugoo:
Resolves #3053.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/api/src/grpc/proto/qdrant_internal_service.proto
+++ b/lib/api/src/grpc/proto/qdrant_internal_service.proto
@@ -4,10 +4,19 @@ package qdrant;
 option csharp_namespace = "Qdrant.Client.Grpc";
 
 service QdrantInternal {
+  rpc GetCommitIndex (GetCommitIndexRequest) returns (GetCommitIndexResponse) {}
+
   /*
   Wait until the target node reached the given commit ID.
    */
   rpc WaitOnConsensusCommit (WaitOnConsensusCommitRequest) returns (WaitOnConsensusCommitResponse) {}
+}
+
+message GetCommitIndexRequest {}
+
+message GetCommitIndexResponse {
+  int64 commit = 1; // Raft commit as u64
+  int64 term = 2; // Raft term as u64
 }
 
 message WaitOnConsensusCommitRequest {
--- a/lib/api/src/grpc/proto/qdrant_internal_service.proto
+++ b/lib/api/src/grpc/proto/qdrant_internal_service.proto
@@ -4,6 +4,9 @@ package qdrant;
 option csharp_namespace = "Qdrant.Client.Grpc";
 
 service QdrantInternal {
+  /*
+  Get current commit and term on the target node.
+   */
   rpc GetCommitIndex (GetCommitIndexRequest) returns (GetCommitIndexResponse) {}
 
   /*
--- a/lib/api/src/grpc/proto/qdrant_internal_service.proto
+++ b/lib/api/src/grpc/proto/qdrant_internal_service.proto
@@ -7,7 +7,7 @@ service QdrantInternal {
   /*
   Get current commit and term on the target node.
    */
-  rpc GetCommitIndex (GetCommitIndexRequest) returns (GetCommitIndexResponse) {}
+  rpc GetConsensusCommit (GetConsensusCommitRequest) returns (GetConsensusCommitResponse) {}
 
   /*
   Wait until the target node reached the given commit ID.
@@ -15,9 +15,9 @@ service QdrantInternal {
   rpc WaitOnConsensusCommit (WaitOnConsensusCommitRequest) returns (WaitOnConsensusCommitResponse) {}
 }
 
-message GetCommitIndexRequest {}
+message GetConsensusCommitRequest {}
 
-message GetCommitIndexResponse {
+message GetConsensusCommitResponse {
   int64 commit = 1; // Raft commit as u64
   int64 term = 2; // Raft term as u64
 }
--- a/lib/common/common/src/panic.rs
+++ b/lib/common/common/src/panic.rs
@@ -1,10 +1,12 @@
-use std::any;
+use std::any::Any;
 
-/// Downcast panic payload into a string
+pub type Payload = dyn Any + Send + 'static;
+
+/// Downcast panic payload into a string.
 ///
-/// Downcast `&'static str` and `String` panic payloads into a `&str`
-pub fn downcast_str(any: &dyn any::Any) -> Option<&str> {
-    if let Some(str) = any.downcast_ref::<&str>() {
+/// Downcast `&'static str` and `String` panic payloads into a `&str`.
+pub fn downcast_str<'a>(any: &'a Box<Payload>) -> Option<&'a str> {
+    if let Some(str) = any.downcast_ref::<&'static str>() {
         return Some(str);
     }
 
--- a/lib/common/common/src/panic.rs
+++ b/lib/common/common/src/panic.rs
@@ -4,7 +4,7 @@ use std::any;
 ///
 /// Downcast `&'static str` and `String` panic payloads into a `&str`
 pub fn downcast_str(any: &dyn any::Any) -> Option<&str> {
-    if let Some(str) = any.downcast_ref::<&'static str>() {
+    if let Some(str) = any.downcast_ref::<&str>() {
         return Some(str);
     }
 
--- a/lib/common/common/src/panic.rs
+++ b/lib/common/common/src/panic.rs
@@ -1,15 +1,16 @@
-use std::any::Any;
+use std::any;
 
-type PanicPayload = Box<dyn Any + Send + 'static>;
-
-/// Convert a panic payload into a string
+/// Downcast panic payload into a string
 ///
-/// This converts `String` and `&str` panic payloads into a string.
-/// Other payload types are formatted as is, and may be non descriptive.
-pub fn panic_payload_into_string(payload: PanicPayload) -> String {
-    payload
-        .downcast::<&str>()
-        .map(|msg| msg.to_string())
-        .or_else(|payload| payload.downcast::<String>().map(|msg| msg.to_string()))
-        .unwrap_or_else(|payload| format!("{payload:?}"))
+/// Downcast `&'static str` and `String` panic payloads into a `&str`
+pub fn downcast_str(any: &dyn any::Any) -> Option<&str> {
+    if let Some(str) = any.downcast_ref::<&'static str>() {
+        return Some(str);
+    }
+
+    if let Some(str) = any.downcast_ref::<String>() {
+        return Some(str);
+    }
+
+    None
 }
--- a/lib/common/common/src/panic.rs
+++ b/lib/common/common/src/panic.rs
@@ -5,7 +5,11 @@ pub type Payload = dyn Any + Send + 'static;
 /// Downcast panic payload into a string.
 ///
 /// Downcast `&'static str` and `String` panic payloads into a `&str`.
-pub fn downcast_str<'a>(any: &'a Box<Payload>) -> Option<&'a str> {
+#[allow(clippy::borrowed_box)]
+// We *have to* use `&Box<dyn Any>`, because `Box<dyn Any>` implements `Any` itself,
+// so `downcast_str(&boxed_any)` would *not* do an auto-deref to inner `dyn Any`,
+// but *coerce* `&Box<dyn Any>` itself to `&dyn Any` :(
+pub fn downcast_str(any: &Box<Payload>) -> Option<&str> {
     if let Some(str) = any.downcast_ref::<&'static str>() {
         return Some(str);
     }
--- a/lib/common/common/src/panic.rs
+++ b/lib/common/common/src/panic.rs
@@ -1,0 +1,15 @@
+use std::any::Any;
+
+type PanicPayload = Box<dyn Any + Send + 'static>;
+
+/// Convert a panic payload into a string
+///
+/// This converts `String` and `&str` panic payloads into a string.
+/// Other payload types are formatted as is, and may be non descriptive.
+pub fn panic_payload_into_string(payload: PanicPayload) -> String {
+    payload
+        .downcast::<&str>()
+        .map(|msg| msg.to_string())
+        .or_else(|payload| payload.downcast::<String>().map(|msg| msg.to_string()))
+        .unwrap_or_else(|payload| format!("{payload:?}"))
+}
--- a/src/actix/api/service_api.rs
+++ b/src/actix/api/service_api.rs
@@ -109,7 +109,7 @@ async fn livez() -> impl Responder {
 
 #[get("/readyz")]
 async fn readyz(ready: web::Data<Ready>) -> impl Responder {
-    let (status, body) = if ready.ready().await {
+    let (status, body) = if ready.check_ready().await {
         (StatusCode::OK, "all shards are ready")
     } else {
         (StatusCode::SERVICE_UNAVAILABLE, "some shards are not ready")
--- a/src/actix/api/service_api.rs
+++ b/src/actix/api/service_api.rs
@@ -1,4 +1,5 @@
 use actix_web::http::header::ContentType;
+use actix_web::http::StatusCode;
 use actix_web::rt::time::Instant;
 use actix_web::web::Query;
 use actix_web::{get, post, web, HttpResponse, Responder};
@@ -10,6 +11,7 @@ use storage::content_manager::toc::TableOfContent;
 use tokio::sync::Mutex;
 
 use crate::actix::helpers::process_response;
+use crate::common::health::Ready;
 use crate::common::helpers::LocksOption;
 use crate::common::metrics::MetricsData;
 use crate::common::stacktrace::get_stack_trace;
@@ -106,8 +108,16 @@ async fn livez() -> impl Responder {
 }
 
 #[get("/readyz")]
-async fn readyz() -> impl Responder {
-    kubernetes_healthz().await
+async fn readyz(ready: web::Data<Ready>) -> impl Responder {
+    let (status, body) = if ready.ready().await {
+        (StatusCode::OK, "all shards are ready")
+    } else {
+        (StatusCode::SERVICE_UNAVAILABLE, "some shards are not ready")
+    };
+
+    HttpResponse::build(status)
+        .content_type(ContentType::plaintext())
+        .body(body)
 }
 
 /// Basic Kubernetes healthz endpoint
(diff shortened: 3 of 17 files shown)


src/common/health.rs
```diff
@@ -0,0 +1,54 @@
...
+        Self {
+            toc,
+            consensus_sync_instant: Instant::now() + consensus_sync_timeout,
+        }
+    }
+
+    pub async fn ready(&self) -> bool {
+        if Instant::now() < self.consensus_sync_instant {
+            return false;
+        }
```
ffuugoo, line 30: So far, the only reasonable way to handle consensus synchronization on startup that I've come up with, is to return "not ready" for the first `n` seconds after startup.

src/common/health.rs
```diff
@@ -0,0 +1,104 @@
...
+            toc,
+            consensus_sync_instant: Instant::now() + consensus_sync_timeout,
+            initial_unhelathy_shards: Default::default(),
+        }
+    }
+
+    pub async fn ready(&self) -> bool {
+        if Instant::now() < self.consensus_sync_instant {
+            return false;
+        }
```
ffuugoo, line 37: The only reasonable way to handle consensus synchronization on startup (that I've come up with) is to _always_ return "not ready" for the first `n` seconds after startup (to give the node some time to catch up with consensus).

src/common/health.rs
```diff
@@ -0,0 +1,104 @@
...
+                        initial_unhealthy_shards.remove(&replica);
+                    } else if is_first_run {
+                        initial_unhealthy_shards.insert(replica);
+                    }
+                }
+
+                initial_unhealthy_shards.is_empty()
+            }
+        }
+    }
```
ffuugoo, line 91: I have a few questions regarding `/readyz` implementation.

### First question:

- Do we want `/readyz` to be a continuous check?
  - e.g., we return "non-ready" *any* time *any* shard becomes non-`Active`
- Or do we want `/readyz` to only check for "startup" readiness?
  - e.g., we collect all shards that are non-`Active` after startup and wait till each shard becomes `Active` *once*, and then we always return "ready", even if some shard becomes non-`Active` again

The first two commits () implement "continuous" check, and the last commit () implements "startup" check.

### Second question:

Is read-locking and iterating through `TableOfContent::collections` and `Collection::shards_holder` on each `/readyz` request is fine?

It is possible to make `TableOfContent`/`Collection`/`ShardHolder` *notify* `Ready` when shard changes state and so `Ready` won't have to lock anything... but the code would become, like, 3 times more complicated.

src/common/health.rs
```diff
@@ -0,0 +1,104 @@
...
+        Self::with_consensus_sync_timeout(toc, Duration::from_secs(10))
+    }
+
+    pub fn with_consensus_sync_timeout(
+        toc: Arc<TableOfContent>,
+        consensus_sync_timeout: Duration,
+    ) -> Self {
+        Self {
+            toc,
+            consensus_sync_instant: Instant::now() + consensus_sync_timeout,
```
timvisee, line 29: Should we maybe call this `consensus_warmup` or something along those lines.

src/common/health.rs
```diff
@@ -0,0 +1,104 @@
...
+
+pub struct Ready {
+    toc: Arc<TableOfContent>,
+    consensus_sync_instant: Instant,
+    initial_unhelathy_shards: RwLock<Option<HashSet<Replica>>>,
+}
+
+impl Ready {
+    pub fn new(toc: Arc<TableOfContent>) -> Self {
+        Self::with_consensus_sync_timeout(toc, Duration::from_secs(10))
```
timvisee, line 20: Instead of using `10` directly, maybe we want to use the `CONSENSUS_META_OP_WAIT` constant.

src/common/health.rs
timvisee: > Do we want /readyz to be a continuous check?

I think only for the initial non-active ones. I think its reasonable to assume a node will remain ready even if you do an arbitrary shard move to it, not checking just the initial ones would break this.

> Or do we want /readyz to only check for "startup" readiness?

I think yes.

Maybe Andrey has a different view on this though.

src/actix/api/service_api.rs
```diff
@@ -106,8 +108,16 @@ async fn livez() -> impl Responder {
...
+async fn readyz(ready: web::Data<Ready>) -> impl Responder {
+    let (status, body) = if ready.ready().await {
+        (StatusCode::OK, "all shards are ready")
+    } else {
+        (StatusCode::SERVICE_UNAVAILABLE, "some shards are not ready")
+    };
+
+    HttpResponse::build(status)
+        .content_type(ContentType::plaintext())
+        .body(body)
```
timvisee, line 120: I think we should format this differently, better matching https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health

We probably want something like:

```
[+]shards/initial-ready ok
healthz check passes
```

```
[-]shards/initial-ready fail
healthz check failed
```

though the fine details of it are unclear to me.

src/common/health.rs
ffuugoo: Nice one, thanks! I just didn't want to spend 15 minutes figuring out the name for the field. 😂

timvisee: I wasn't sure if my suggestion would be either useful or annoying. Happy to see it's the first. :smile:

src/actix/api/service_api.rs
ffuugoo: _As far as I understand_, that's how _Kubernetes_ will format its output based solely on health endpoints status code. We don't need to format response body (or provide any) in any way.

timvisee: Looks like you're right! Resources on it are hard to find.

In that case shouldn't we just return HTTP 200 with `ready`

src/common/health.rs
```diff
@@ -50,55 +81,65 @@ impl Ready {
...
+fn filter(
+    shards: impl IntoIterator<Item = (Shard, bool)>,
+    filter: bool,
+) -> impl Iterator<Item = Shard> {
+    shards.into_iter().filter_map(move |(shard, is_ready)| {
+        if is_ready == filter {
+            Some(shard)
+        } else {
+            None
+        }
```
timvisee, line 143: There's [`then_some`](https://doc.rust-lang.org/std/primitive.bool.html#method.then_some) for this:

```rust
(is_ready == filter).then_some(shard)
```

ffuugoo:
I think I've figured "caught up with consensus after restart" conditions. They are a bit different if the node was a "follower" or a "leader" before the restart.

**Follower** is "caught up with consensus" if:

- it *becomes leader*
  - *IIRC*, only the most up-to-date node can become a leader
- or it receives non-`RequestVote` *request* message with `term >= local_term`, `commit > 0` and `commit <= local_commit`
  - *any* node can send `RequestVote`
    - e.g., even the node that has been in crash loop/disconnected from network for the last 30 minutes
    - so, we can't use `RequestVote` for `commit` checks
  - other *requests* should only come from the leader
    - but sometimes a *previous* leader may not realize it is not a leader anymore
    - so `term >= local_term` check is *required*
  - `commit` may be undefined on some requests (e.g., set to `0`)
    - so `commit > 0` check is *required*

**Leader** is "caught up with consensus" if:
- it is downgraded to a follower node and then fulfill any of the two conditions for a follower node
- it is a leader *and* it receives any *response* message from *half* of cluster nodes
  - when node starts an election,
    - it *downgrades to a follower*
    - increments `local_term`
    - and sends `RequestVote` to other nodes
  - when node receives a request with `term > local_term`
    - it *downgrades to a follower*
    - and updates `local_term = term`
  - when node receives a request with `term < local_term`
    - it ignores it
    - and *does not send a response*
  - so, if *current* node starts an election
    - *it will downgrade to a follower*
    - and condition won't hold
  - if *other* node starts an election and sends `RequestVote` to *current* node
    - *current node will downgrade to a follower*
    - and condition won't hold
  - if *other* node starts an election and sends `RequestVote` to enough *other* nodes
    - these nodes *won't send a response* to the *current* node
    - and (eventually) *current node will downgrade to a follower*
    - and condition won't hold
  - if current node receives enough responses
    - current node + half of cluster nodes is a *majority*
    - even if election happens a *moment* after fulfilling this condition
    - it means current node *was* a leader a *moment* ago
    - and nothing can be committed to consensus without a leader
    - so current node is "caught up with consensus"
    - and condition holds

Now I just have to figure out how to implement these checks. 😵⚰

@generall

ffuugoo:
@generall Proposed an alternative, somewhat simpler solution: after startup, during `sync_local_state`, propose a `Nop` operation (with unique token value) to consensus, and wait until this operation is committed and "applied" on the node. Once the operation is applied, treat it as "caught up with consensus" condition, and start probing shards' status.

timvisee:
> @generall proposed an alternative, somewhat simpler solution: after startup, during `sync_local_state`, propose a `Nop` operation (with unique token value) to consensus, and wait until this operation is committed and "applied" on the node. Once the operation is applied, treat it as "caught up with consensus" condition, and start probing shards' status.

I like this idea!

Maybe we can add an operation for it called `Marker(usize)` (where `usize` is our token).

src/common/health.rs
```diff
@@ -0,0 +1,328 @@
...
+            }
+
+            // TODO: Limit parallelism? (E.g., `StreamExt::buffer_unordered`)
+            let requests: FuturesUnordered<_> = peer_address_by_id
+                .values()
+                .map(|uri| get_commit_index(&transport_channel_pool, uri))
+                .collect();
+
+            // TODO: Handle single-node cluster!
+            let required_commit_indices_count = cmp::min(peer_address_by_id.len() / 2, 1);
```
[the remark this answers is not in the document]
generall: I would skip this check completely, is there are no nodes or just one node in the cluster

lib/api/src/grpc/proto/qdrant_internal_service.proto
```diff
@@ -4,12 +4,24 @@ package qdrant;
 option csharp_namespace = "Qdrant.Client.Grpc";
 
 service QdrantInternal {
+  /*
+  Get current commit and term on the target node.
+   */
+  rpc GetCommitIndex (GetCommitIndexRequest) returns (GetCommitIndexResponse) {}
```
timvisee, line 10: Shall we rename this to `GetConsensusCommit` to be consistent with the preexisting things in this file?

src/common/health.rs
```diff
@@ -0,0 +1,359 @@
...
+            self.wait_ready().await
+        }
+    }
+
+    pub fn is_ready(&self) -> bool {
+        self.is_ready.load(atomic::Ordering::Relaxed)
+    }
+
+    async fn wait_ready(&self) -> bool {
+        time::timeout(Duration::from_millis(500), self.notify_ready.notified())
```
timvisee, line 65: This can data race. Even if ready now, we aren't guaranteed to receive a notification here when this is called from `check_ready()`.

An example scenario would be that we become ready right after the `is_ready()` check, after which `notify_task()` returns early and `wait_ready()` doesn't get notified.

It wouldn't be a disaster, but nice to fix anyway.

src/common/health.rs
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    async fn check_ready(&mut self) -> bool {
+        for _ in 0..10 {
+            match self.check_ready.recv().await {
+                Ok(()) => return true,
+                Err(broadcast::error::RecvError::Closed) => return false,
+                Err(broadcast::error::RecvError::Lagged(_)) => (),
+            }
+        }
```
timvisee, line 285: Is 10 just an arbitrary value? Wouldn't a single retry not be enough with `Lagged(_)`? I'm unsure.

src/common/health.rs
ffuugoo: Oh, you're right.

It *won't* data race, if we use `Notify::notify_one` (instead of `Notify::notify_waiters`) in `Task::exec` to prevent data race. Didn't read the docs for `Notify::notify_waiters` carefully enough...

lib/api/src/grpc/proto/qdrant_internal_service.proto
ffuugoo: Sure.

src/common/health.rs
ffuugoo: I only found `Notify` channel today, and I think it would be easier to use `CancellationToken` + `Notify` than a single `broadcast` channel.

timvisee: What about a CondVar to prevent this situation completely? Not sure how well that works in async context though.

generall: Last time we had problems with `CondVar` in async, so I would rather use tokio here

src/common/health.rs
ffuugoo: Still may be possible for subscriber to miss both initial `self.is_ready()` and `self.notify_ready.notified()`, but it will resolve correctly after 500ms wait.

ffuugoo: Or what @generall said, it's just a bit less convenient to use.

src/common/health.rs
```diff
@@ -14,260 +14,209 @@ use collection::shards::CollectionId;
...
+            log::error!("health::Task panicked") // TODO!?
         }
+    }
 
-        let mut commit_to_await = u64::MAX;
+    async fn exec_catch_unwind(&mut self) -> thread::Result<()> {
+        panic::AssertUnwindSafe(self.exec_cancel())
+            .catch_unwind()
+            .await
+    }
```
ffuugoo, line 109: IMO, this is simpler than the previous `spawn_task()` implementation. The only way it can fail, is if `log::error!(...)` panics.

src/main.rs
```diff
@@ -286,6 +286,14 @@ fn main() -> anyhow::Result<()> {
 
         // Runs raft consensus in a separate thread.
         // Create a pipe `message_sender` to communicate with the consensus
+        let health_checker = Arc::new(common::health::HealthChecker::new(
+            toc_arc.clone(),
+            consensus_state.clone(),
+            runtime_handle.clone(),
+        ));
+
+        runtime_handle.block_on(health_checker.check_ready());
```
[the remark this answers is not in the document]
ffuugoo: Not an issue after my last update.

src/common/health.rs
ffuugoo: Replaced `broadcast` with `Notify` (and `CancellationToken`).

src/common/health.rs
ffuugoo: Fixed as @generall suggested.

src/common/health.rs
ffuugoo: **UPD:** if `downcast_str` or `log::error!(...)` panics

lib/api/src/grpc/proto/qdrant_internal_service.proto
ffuugoo: Done.

generall (approved):
Tested locally, works as I would expect.

Before enabling in default helm-chart, we should test on actual k8s:

- create 3 machines cluster
- start uploading data into collection with replication_factor=2
- make rolling restart
- check that data upload never interrupted

src/common/health.rs
```diff
@@ -0,0 +1,337 @@
...
+            }
+        }
+
+        unhealthy_shards
+    }
+
+    fn set_ready(&self) {
+        self.is_ready.store(true, atomic::Ordering::Relaxed);
+        self.is_ready_signal.notify_one();
+    }
```
timvisee, line 284: I guess with our current implementation we want to use `notify_waiters` here, because you're unsure how many waiters there are.

What do you think?

src/common/health.rs
```diff
@@ -0,0 +1,337 @@
...
+    )
+}
+
+type GetConsensusCommitResult = Result<
+    tonic::Response<GetConsensusCommitResponse>,
+    transport_channel_pool::RequestError<tonic::Status>,
+>;
+
+fn is_shard_ready(state: &ReplicaState) -> bool {
+    matches!(state, ReplicaState::Active | ReplicaState::Listener)
```
timvisee, line 310: Should we prefer an exhaustive match here, or move the function into `ReplicaState`?

src/common/health.rs
timvisee: No problem! Thanks for picking it up.

src/common/health.rs
timvisee: I recently added a function similar to this, maybe we can add another one for this check below it: https://github.com/qdrant/qdrant/blob/ef9a11c200f773765ea2a5424226a761e5fc6db2/lib/collection/src/shards/replica_set/mod.rs#L882-L893

lib/common/common/src/panic.rs
```diff
@@ -0,0 +1,15 @@
...
+///
+/// This converts `String` and `&str` panic payloads into a string.
+/// Other payload types are formatted as is, and may be non descriptive.
+pub fn panic_payload_into_string(payload: PanicPayload) -> String {
+    payload
+        .downcast::<&str>()
+        .map(|msg| msg.to_string())
+        .or_else(|payload| payload.downcast::<String>().map(|msg| msg.to_string()))
+        .unwrap_or_else(|payload| format!("{payload:?}"))
+}
```
ffuugoo, line 15: I'd use my `downcast_str`:
- does not require `Send` and `'static`
- does not consume `payload`
- does not allocate
- also, IMO, it is easier to read the way I structured it :D

timvisee: Sure! I don't really mind since we only call this on panic.

I selected the owned version because you cannot `format!("{payload:?}")` into a `&str`.

timvisee: Does it? Wasn't aware of that.

In that case sorry for destroying your work :clown_face:

src/common/health.rs
ffuugoo: Fixed.

src/common/health.rs
ffuugoo: Fixed.

lib/common/common/src/panic.rs
ffuugoo: Fixed.

PR #3086, opened 2023-11-23
spare one extra worker by replacing spawn with block_on

generall:
When migrating from local -> distributed,
qdrant should explicitly fill consensus WAL with missing operations, so they could be replicated to other collection.

It appears, that in current implementation the migration requires one extra parallel thread of execution.

If we try to run migration of low CPU machine (<=2 CPU), the allocated thread pool doesn't have enough processes to execute the operation, which results in stuck consensus:

```
"raft_info": {
      "term": 0,
      "commit": 0,
      "pending_operations": 0,
      "leader": null,
      "role": null,
      "is_voter": true
    }
```

This PR changes `spawn` into `block_on` for migration process, which should prevent over-usage of threads.

An alternative could be to allocate +1 thread minimum by default or use different runtime.
But the current option looks the simplest.

Workaround for previous versions:
Run qdrant with `QDRANT_NUM_CPUS=3`

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -327,7 +327,7 @@ fn main() -> anyhow::Result<()> {
         };
 
         if !collections_to_recover_in_consensus.is_empty() {
-            runtime_handle.spawn(handle_existing_collections(
+            runtime_handle.block_on(handle_existing_collections(
                 toc_arc.clone(),
                 consensus_state.clone(),
                 dispatcher_arc.clone(),


timvisee (approved):

ffuugoo (approved):
Smells like "blocking call in async context"....

generall:
> Smells like "blocking call in async context"....

main is not async and we have other `block_on` calls just nearby

ffuugoo:
> main is not async and we have other `block_on` calls just nearby

I meant the future that is passed into `spawn` is doing a blocking call that blocks the only available async-runtime thread and prevents _other_ async task (_that the same future depends on_) from progress, deadlocking consensus. Or, maybe, it's the _other_ task doing a blocking call. But I'm 90% sure there's a blocking call somewhere. Tokio has a single-thread runtime mode, so the number of threads should not be an issue for well-behaved futures.

PR #3092, opened 2023-11-27
Mmap InvertedIndex uses Sequential madvice

agourlay:
This PR changes the madvice for the mmap inverted index to be sequential instead of random.

The rational is that posting lists are traversed sequentially during search so it should match the access pattern.

I was not able to validate a performance improvement on the included benchmark because:
- everything fits in memory nicely.
- the query vector does not change.

I still think it is a good change based on the design of the search algorithm.

### DIFF
--- a/lib/sparse/src/index/inverted_index/inverted_index_mmap.rs
+++ b/lib/sparse/src/index/inverted_index/inverted_index_mmap.rs
@@ -111,7 +111,8 @@ impl InvertedIndexMmap {
         create_and_ensure_length(file_path.as_ref(), file_length)?;
 
         let mut mmap = open_write_mmap(file_path.as_ref())?;
-        madvise::madvise(&mmap, madvise::get_global())?;
+        // posting lists are traversed sequentially
+        madvise::madvise(&mmap, madvise::Advice::Sequential)?;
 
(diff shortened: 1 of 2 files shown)


timvisee:
> I was not able to validate a performance improvement on the included benchmark because:
> 
>     * everything fits in memory nicely.
> 
>     * the query vector does not change.
> 
> 
> I still think it is a good change based on the design of the search algorithm.

Would it be possible to test this with randomized data, larger than your available memory?

Madvise can give very unexpected results, even if some advice seems very logical.

lib/sparse/src/index/inverted_index/inverted_index_mmap.rs
```diff
@@ -144,7 +145,8 @@ impl InvertedIndexMmap {
         // read index data into mmap
         let file_path = Self::index_file_path(path.as_ref());
         let mmap = open_read_mmap(file_path.as_ref())?;
-        madvise::madvise(&mmap, madvise::get_global())?;
+        // posting lists are traversed sequentially
+        madvise::madvise(&mmap, madvise::Advice::Sequential)?;
```
generall, line 149: I would set `Normal`

PR #3109, opened 2023-11-29
extend shard key tests

generall:


tests/consensus_tests/test_custom_sharding.py
```diff
@@ -112,4 +126,94 @@ def test_shard_consistency(tmp_path: pathlib.Path):
...
+            "limit": 10,
+            "with_payload": True,
+        }
+    )
+
+    assert_http_ok(r)
+    result = r.json()["result"]
+    assert len(result) == 8
+    for point in result:
+        assert point["shard_key"] in ["cats", "dogs"]
```
timvisee, line 164: Shall we also assert that we have results for at least both, and not just one of them?

generall: `assert len(result) == 8` should cover that

timvisee (approved):

PR #2802, opened 2023-10-12
Sparse index segment and collection config

IvanPleshkov:
### All Submissions:

* [ ] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [ ] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [ ] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #3111, opened 2023-11-29
sparse vector propagate validation

IvanPleshkov:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/operations/point_ops.rs
+++ b/lib/collection/src/operations/point_ops.rs
@@ -33,13 +33,14 @@ pub enum WriteOrdering {
     Strong,
 }
 
-#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, Validate)]
 #[serde(rename_all = "snake_case")]
 pub struct PointStruct {
     /// Point id
     pub id: PointIdType,
     /// Vectors
     #[serde(alias = "vectors")]
+    #[validate]
     pub vector: VectorStruct,
     /// Payload values (optional)
     pub payload: Option<Payload>,
@@ -149,6 +150,7 @@ pub struct PointSyncOperation {
 
 #[derive(Debug, Deserialize, Serialize, Clone, Validate, JsonSchema)]
 pub struct PointsBatch {
+    #[validate]
     pub batch: Batch,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub shard_key: Option<ShardKeySelector>,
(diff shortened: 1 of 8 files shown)


lib/collection/src/tests/sparse_vectors_validation_tests.rs
```diff
@@ -0,0 +1,197 @@
...
+        payload: None,
+    }
+}
+
+fn wrong_recommend_example() -> RecommendExample {
+    RecommendExample::Sparse(wrong_sparse_vector())
+}
+
+fn check_validation_error(result: Result<(), ::validator::ValidationErrors>) {
+    println!("{:?}", result);
```
agourlay, line 43: can be remove now :)

agourlay:
It would be great to have integration tests on the API side.
Do you plan doing it in the API PR after rebasing?

lib/collection/src/tests/sparse_vectors_validation_tests.rs
```diff
@@ -0,0 +1,197 @@
...
+        vector: VectorStruct::Multi(vector_data),
+        payload: None,
+    }
+}
+
+fn wrong_recommend_example() -> RecommendExample {
+    RecommendExample::Sparse(wrong_sparse_vector())
+}
+
+fn check_validation_error(result: Result<(), ::validator::ValidationErrors>) {
```
agourlay, line 42: maybe the input could something that impl `Validate` so that you could to the `.validate()` call here only once

agourlay (approved):
Thank you for tracking those call sites :+1: 

LGTM I have only minor comments.

IvanPleshkov:
> It would be great to have integration tests on the API side. Do you plan doing it in the API PR after rebasing?

I don't know. Maybe integration tests are better than unit tests that are presented here. I guess you are right and I ask for skip it here. Because we don't have integration tests at all we can plan error cases included validation

lib/collection/src/tests/sparse_vectors_validation_tests.rs
IvanPleshkov: Thanks for this comment. I had an annoying integration test failure with a non-matched string of the error message. I spend one hour of fixing and forgot about removing debug stuff. Thanks for help!

lib/collection/src/tests/sparse_vectors_validation_tests.rs
IvanPleshkov: It's a good idea, like it! Fixed

agourlay:
Let's merge it as it is then and iterate with the integration tests in https://github.com/qdrant/qdrant/pull/2802

PR #3115, opened 2023-11-30
Dynamic CPU saturation integration

timvisee:
In <https://github.com/qdrant/qdrant/pull/3364> we made the groundwork for dynamically saturating CPU usage while optimizing.

This PR integrates the feature with additionally required public API/configuration changes. Instead of having fixed optimization limits in the configuration, everything now defaults to being dynamic (automatic).

Configuration changes are mostly described in review comments below. I've also added a `cpu_budget` property to configure what number of CPUs to saturate for optimization tasks.

The idea is to merge this as part of Qdrant 1.8.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -41,8 +41,15 @@ storage:
     max_search_threads: 0
 
     # Max number of threads (jobs) for running optimizations across all collections, each thread runs one job.
+    # If 0 - have no limit and choose dynamically to saturate CPU.
     # Note: each optimization job will also use `max_indexing_threads` threads by itself for index building.
-    max_optimization_threads: 1
+    max_optimization_threads: 0
+
+    # CPU budget, how many CPUs (threads) to allocate for an optimization job.
+    # If 0 - auto selection, keep 1 or more CPUs unallocated depending on CPU size
+    # If negative - subtract this number of CPUs from the available CPUs.
+    # If positive - use this exact number of CPUs.
+    optimizer_cpu_budget: 0
 
     # Prevent DDoS of too many concurrent updates in distributed mode.
     # One external update usually triggers multiple internal updates, which breaks internal
@@ -109,8 +116,9 @@ storage:
 
     # Max number of threads (jobs) for running optimizations per shard.
     # Note: each optimization job will also use `max_indexing_threads` threads by itself for index building.
+    # If null - have no limit and choose dynamically to saturate CPU.
     # If 0 - no optimization threads, optimizations will be disabled.
-    max_optimization_threads: 1
+    max_optimization_threads: null
 
   # Default parameters of HNSW Index. Could be overridden for each collection or named vector individually
   hnsw_index:
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -5526,7 +5526,7 @@
             "nullable": true
           },
           "max_indexing_threads": {
-            "description": "Number of parallel threads used for background index building. If 0 - auto selection.",
+            "description": "Number of parallel threads used for background index building. If 0 - automatically select from 8 to 16. Best to keep between 8 and 16 to prevent likelihood of building broken/inefficient HNSW graphs. On small CPUs, less threads are used.",
             "type": "integer",
             "format": "uint",
             "minimum": 0,
@@ -5726,7 +5726,7 @@
             "minimum": 0
           },
           "max_indexing_threads": {
-            "description": "Number of parallel threads used for background index building. If 0 - auto selection.",
+            "description": "Number of parallel threads used for background index building. If 0 - automatically select from 8 to 16. Best to keep between 8 and 16 to prevent likelihood of slow building or broken/inefficient HNSW graphs. On small CPUs, less threads are used.",
             "default": 0,
             "type": "integer",
             "format": "uint",
@@ -5752,7 +5752,6 @@
           "default_segment_number",
           "deleted_threshold",
           "flush_interval_sec",
-          "max_optimization_threads",
           "vacuum_min_vector_number"
         ],
         "properties": {
@@ -5806,10 +5805,12 @@
             "minimum": 0
           },
           "max_optimization_threads": {
-            "description": "Maximum available threads for optimization workers",
+            "description": "Max number of threads (jobs) for running optimizations per shard. Note: each optimization job will also use `max_indexing_threads` threads by itself for index building. If null - have no limit and choose dynamically to saturate CPU. If 0 - no optimization threads, optimizations will be disabled.",
+            "default": null,
             "type": "integer",
             "format": "uint",
-            "minimum": 0
+            "minimum": 0,
+            "nullable": true
           }
         }
       },
@@ -7484,7 +7485,7 @@
             "nullable": true
           },
           "max_optimization_threads": {
-            "description": "Maximum available threads for optimization workers",
+            "description": "Max number of threads (jobs) for running optimizations per shard. Note: each optimization job will also use `max_indexing_threads` threads by itself for index building. If null - have no limit and choose dynamically to saturate CPU. If 0 - no optimization threads, optimizations will be disabled.",
             "type": "integer",
             "format": "uint",
             "minimum": 0,
--- a/lib/collection/src/optimizers_builder.rs
+++ b/lib/collection/src/optimizers_builder.rs
@@ -70,8 +70,12 @@ pub struct OptimizersConfig {
     pub indexing_threshold: Option<usize>,
     /// Minimum interval between forced flushes.
     pub flush_interval_sec: u64,
-    /// Maximum available threads for optimization workers
-    pub max_optimization_threads: usize,
+    /// Max number of threads (jobs) for running optimizations per shard.
+    /// Note: each optimization job will also use `max_indexing_threads` threads by itself for index building.
+    /// If null - have no limit and choose dynamically to saturate CPU.
+    /// If 0 - no optimization threads, optimizations will be disabled.
+    #[serde(default)]
+    pub max_optimization_threads: Option<usize>,
 }
 
 impl OptimizersConfig {
@@ -85,7 +89,7 @@ impl OptimizersConfig {
             memmap_threshold: None,
             indexing_threshold: Some(100_000),
             flush_interval_sec: 60,
-            max_optimization_threads: 0,
+            max_optimization_threads: Some(0),
         }
     }
 
--- a/lib/common/common/src/cpu.rs
+++ b/lib/common/common/src/cpu.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -7,6 +8,8 @@ use thiserror::Error;
 use thread_priority::{set_current_thread_priority, ThreadPriority, ThreadPriorityValue};
 use tokio::sync::{OwnedSemaphorePermit, Semaphore, TryAcquireError};
 use tokio::time;
+
+use crate::defaults::default_cpu_budget_unallocated;
 
 /// Try to read number of CPUs from environment variable `QDRANT_NUM_CPUS`.
 /// If it is not set, use `num_cpus::get()`.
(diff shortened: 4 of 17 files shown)


lib/common/common/src/defaults.rs
```diff
@@ -1,4 +1,47 @@
...
+
+/// Default value of CPU budget parameter.
+///
+/// Dynamic based on CPU size.
+///
+/// On low CPU systems, we want to reserve the minimal amount of CPUs for other tasks to allow
+/// efficient optimization. On high CPU systems we want to reserve more CPUs.
+#[inline(always)]
+pub fn default_cpu_budget_param(num_cpu: usize) -> NonZeroIsize {
+    let cpu_budget = if num_cpu <= 32 {
```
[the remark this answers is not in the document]
timvisee: `0..=2` CPUs now default to 0, meaning for these 0 CPUs are kept unallocated.

lib/common/common/src/cpu.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// Block until we have any CPU budget available.
+    ///
+    /// Uses an exponential backoff strategy to avoid busy waiting.
+    pub fn block_until_budget(&self) {
+        if self.has_budget() {
+            return;
+        }
+
+        // Wait for CPU budget to be available with exponential backoff
+        // TODO: find better way, don't busy wait
```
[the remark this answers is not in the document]
timvisee: Agreed. The question remains whether it's worth the extra complexity though.

lib/collection/src/optimizers_builder.rs
```diff
@@ -71,8 +70,12 @@ pub struct OptimizersConfig {
     pub indexing_threshold: Option<usize>,
     /// Minimum interval between forced flushes.
     pub flush_interval_sec: u64,
-    /// Maximum available threads for optimization workers
-    pub max_optimization_threads: usize,
+    /// Max number of threads (jobs) for running optimizations per shard.
+    /// Note: each optimization job will also use `max_indexing_threads` threads by itself for index building.
+    /// If null - have no limit and choose dynamically to saturate CPU.
+    /// If 0 - no optimization threads, optimizations will be disabled.
```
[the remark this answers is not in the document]
timvisee: Do you have a different suggestion? We need to support both options.

I'd argue that using `0` for `auto` in other places is wrong, and that we should (have) use(d) `null` there instead.

lib/storage/src/content_manager/toc/mod.rs
```diff
@@ -59,6 +59,9 @@ pub struct TableOfContent {
     search_runtime: Runtime,
     update_runtime: Runtime,
     general_runtime: Runtime,
+    /// Global CPU budget in number of cores for all optimization tasks.
+    /// Assigns CPU permits to tasks to limit overall resource utilization.
+    cpu_budget: CpuBudget,
```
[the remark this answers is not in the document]
timvisee: Good suggestion. Thanks!

config/config.yaml
```diff
@@ -39,10 +39,17 @@ storage:
...
+    # Max number of threads (jobs) for running optimizations across all collections, each thread runs one job.
+    # If 0 - have no limit and choose dynamically to saturate CPU.
+    # Note: each optimization job will also use `max_indexing_threads` threads by itself for index building.
+    max_optimization_threads: 0
+
+    # CPU budget, how many CPUs (threads) to allocate for an optimization job.
+    # If 0 - auto selection, keep 1 or more CPUs unallocated depending on CPU size
+    # If negative - subtract this number of CPUs from the available CPUs.
+    # If positive - use this exact number of CPUs.
+    cpu_budget: 0
```
timvisee, line 52: Should we rename this one to `optimizer_cpu_budget` as well? I don't want to make names more difficult if not necessary.

cc @coszio

lib/collection/src/optimizers_builder.rs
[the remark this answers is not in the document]
coszio: One solution can be an enum so that users can set number or "auto" string. I am mostly worried about the case of deserialization where it's hard to capture the semantic of `{ "field": null }` vs `{ }`, which has been the problem in other places

config/config.yaml
coszio: I agree with you, looks too verbose for being user-facing. Talking about _necessity_, though, the only thing capturing the semantic of this being applied only to optimizers is the comment above it. Not sure either

timvisee: > (...) the only thing capturing the semantic of this being applied only to optimizers is the comment above it. Not sure either

A CPU budget doesn't currently make sense for anything else. But the average user probably doesn't have this Insight.

generall: I would vote for `optimizer_cpu_budget`

lib/common/common/src/cpu.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub fn new(cpu_budget: usize) -> Self {
+        Self {
+            semaphore: Arc::new(Semaphore::new(cpu_budget)),
+        }
+    }
+
+    /// Try to acquire CPU permit for optimization task from global CPU budget.
+    pub fn try_acquire(&self, desired_cpus: usize) -> Option<CpuPermit> {
+        // Determine what number of CPUs to acquire based on available budget
+        let num_cpus = self.semaphore.available_permits().min(desired_cpus) as u32;
```
generall, line 68: I don't think we should allow creation of permits with less then `desired_cpus` number of cores (except for the cases, where total number of CPUs are less than `desired_cpus`).


Imagine the situation, where we have queue of 2 tasks with `desired_cpus = 7`  and 8 CPUs available. 

That will generate permit for 7 and 1 CPUs respectively. Assuming linear CPU/task efficiency, the total time of the optimization will be defined by the slowest task - the one which was executed by 1 CPU. Let's say it is X seconds.

The other task will be completed in X/7 seconds, which will left 7 CPUs un-utilized for X * 6/7 seconds.


A scenario where we schedule tasks only if `desired_cpus` is available would take X * 2/7 seconds

generall: But I would agree, if we make the limit to be `self.semaphore.available_permits() > (desired_cpus) / 2 + 1`

lib/collection/src/update_handler.rs
```diff
@@ -221,24 +230,55 @@ impl UpdateHandler {
...
+                }
+
                 let nonoptimal_segment_ids =
                     optimizer.check_condition(segments.clone(), &scheduled_segment_ids);
                 if nonoptimal_segment_ids.is_empty() {
                     break;
                 }
 
+                // Determine how many CPUs we prefer for optimization task, acquire permit for it
+                let max_indexing_threads = optimizer.hnsw_config().max_indexing_threads;
```
generall, line 258: What if `max_indexing_threads > optimizer_cpu_budget`?

docs/redoc/master/openapi.json
```diff
@@ -5649,10 +5648,12 @@
...
           },
           "max_optimization_threads": {
-            "description": "Maximum available threads for optimization workers",
+            "description": "Max number of threads (jobs) for running optimizations per shard. Note: each optimization job will also use `max_indexing_threads` threads by itself for index building. If null - have no limit and choose dynamically to saturate CPU. If 0 - no optimization threads, optimizations will be disabled.",
+            "default": null,
             "type": "integer",
             "format": "uint",
-            "minimum": 0
+            "minimum": 0,
+            "nullable": true
```
generall, line 5656: this is not backward compatible with client=(

generall: Would it be possible to avoid API changes, so that we could fit it into patch release?

lib/collection/src/update_handler.rs
generall: I think this could become important to explicitly handle once we have better logic in `try_acquire`

timvisee: This is fine in the latest implementation. The desired number of CPUs is internally limited to the total CPU budget.

For example:
- total CPU budget: 8
- desired: usize::MAX
- permit aquired: 4..=8

lib/common/common/src/cpu.rs
timvisee: The new implementation uses the _'at least half'_ approach: `semaphore.available_permits() > ceil(desired_cpus / 2)`.

It also limits the desired CPU to the total CPU budget to prevent having abnormal 'minimum required' numbers.

timvisee:
We want to merge the internals of this feature first, so that we can release it as part of a patch release. We can integrate the public API changes for the next minor release to make this feature fully functional.

Please see <https://github.com/qdrant/qdrant/pull/3364> for the internals. I'll draft this PR until it is merged.

timvisee:
## Indexing benchmark

I've done some benchmarking to measure the indexing performance of this PR. I used a machine with 64 CPU cores, using the default configuration. For each test I insert a million points with indexing disabled, then I enable it again and measure indexing time. _Old_ refers to the current latest `dev` branch, while _new_ refers to this PR.

I don't have fancy graphs, but I think the list below is good enough to show off the gains.

### Results
**2 CPU cores:**
- `bfb -n 100000`:
  - Old: 40s
  - New: 40s

**8 CPU cores:**
- `bfb -n 1000000`:
  - Old: 163s
  - New: 186s (14% slower)
    - This is slower because the new approach uses 7 cores rather than 8. We do this intentionally to spare a core for handling other stuff, such as ongoing searches. Users can configure this with `storage.performance.cpu_budget`. We can also consider to change this default [here](https://github.com/qdrant/qdrant/blob/c480b3ea1172789f5f57f30af86a43d14917b5b8/lib/common/common/src/defaults.rs#L6-L23).

**16 CPU cores:**
- `bfb -n 1000000`:
  - Old: 123s
  - New: 92s (34% faster)
- `bfb -n 1000000 --segments 4`:
  - Old: 126s
  - New: 113s (12% faster)
- `bfb -n 1000000 --segments 8`:
  - Old: 123s
  - New: 92s (34% faster)

**32 CPU cores:**
- `bfb -n 1000000`:
  - Old: 121s
  - New: 54s (124% faster)
- `bfb -n 1000000 --segments 4`:
  - Old: 127s
  - New: 96s (32% faster)
- `bfb -n 1000000 --segments 8`:
  - Old: 121s
  - New: 54s (124% faster)
- `bfb -n 1000000 --segments 16`:
  - Old: 113s
  - New: 53s (113% faster)

**64 CPU cores:**
- `bfb -n 1000000`:
  - Old: 122s
  - New: 47s (160% faster)
- `bfb -n 1000000 --segments 4`:
  - Old: 129s
  - New: 74s (74% faster)
- `bfb -n 1000000 --segments 8`:
  - Old: 119s
  - New: 48s (148% faster)
- `bfb -n 1000000 --segments 16`:
  - Old: 120s
  - New: 43s (179% faster)

When using 64 cores, here we see CPU saturation in `htop` without this PR:
Screenshot from 2024-01-29 16-10-32

And here it is with this PR:
Screenshot from 2024-01-29 16-13-03

config/config.yaml
```diff
@@ -41,8 +41,15 @@ storage:
     max_search_threads: 0
 
     # Max number of threads (jobs) for running optimizations across all collections, each thread runs one job.
+    # If 0 - have no limit and choose dynamically to saturate CPU.
     # Note: each optimization job will also use `max_indexing_threads` threads by itself for index building.
-    max_optimization_threads: 1
+    max_optimization_threads: 0
```
timvisee, line 46: `performance.max_optimization_threads: 0` now means there is no limit and the CPU should be saturated dynamically. This is the new default.

Even though we have a new `cpu_budget` parameter now, this still works. We can deprecate, but I guess it's fine to keep it as-is.

config/config.yaml
```diff
@@ -109,8 +116,9 @@ storage:
 
     # Max number of threads (jobs) for running optimizations per shard.
     # Note: each optimization job will also use `max_indexing_threads` threads by itself for index building.
+    # If null - have no limit and choose dynamically to saturate CPU.
     # If 0 - no optimization threads, optimizations will be disabled.
-    max_optimization_threads: 1
+    max_optimization_threads: null
```
timvisee, line 121: `optimizers.max_optimization_threads` is per shard, not per collection. Now supports `null` to have no default and dynamically saturate the CPU, this is the new default.

generall (approved):

lib/common/common/src/cpu.rs
```diff
@@ -32,23 +35,19 @@ pub fn get_num_cpus() -> usize {
...
-    //     Ordering::Greater => cpu_budget_param as usize,
-    // }
+pub fn get_cpu_budget(cpu_budget_param: isize) -> usize {
+    match cpu_budget_param.cmp(&0) {
+        // If less than zero, subtract from available CPUs
+        Ordering::Less => get_num_cpus()
+            .saturating_sub(-cpu_budget_param as usize)
+            .max(1),
+        // If zero, use automatic selection
+        Ordering::Equal => get_num_cpus()
```
agourlay, line 45: I think you could call `get_num_cpus` only once here

timvisee: Good catch. Thanks!

Patched in fd0157cbfe2e1fe2296e9d12149bec8e03f299ef.

agourlay (approved):

PR #3117, opened 2023-11-30
Fix for init_from to honor collection creation timeout

agourlay:
The collection creation timeout is not applied to the `init_from` data transfer as it is being executed on an offline task.

Found via Python congruence tests.

### DIFF
--- a/lib/storage/src/content_manager/toc/create_collection.rs
+++ b/lib/storage/src/content_manager/toc/create_collection.rs
@@ -201,7 +201,7 @@ impl TableOfContent {
 
         if let Some(init_from) = init_from {
             self.run_data_initialization(init_from.collection, collection_name.to_string())
-                .await;
+                .await?;
         }
 
         Ok(true)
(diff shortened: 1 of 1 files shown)


lib/storage/src/content_manager/toc/create_collection.rs
```diff
@@ -293,5 +293,7 @@ impl TableOfContent {
                 Err(err) => log::error!("Initialization failed: {}", err),
             }
         });
+        handle.await?;
```
generall, line 296: Now the collection create API will await for the data copying, isn't it?

generall: I don't think this is good decision. With large collection it will block for too long, which is especially critical in distributed env, where you only have a few ticks to complete operation or reelection will be triggered.

The full in-background job was a conscious design decision

agourlay:
Thank you for answering the question :+1: 

This is a fine outcome, I will change the existing congruence tests to not assume the data copy of `init_from` is done synchronously.

PR #3119, opened 2023-12-01
Promote collection create/update/delete debug messages to info

timvisee:
We have debug log messages for creating, updating and deleting collections. This promotes them from debug to info.

I think these are important markers, and may function as some form of audit log to discover "hey this collection was indeed deleted". I've seen a few users now where it was unclear to them if a collection had been deleted through Qdrant.

The only concern is that they may introduce some noise if a user is creating/deleting a lot of collections. But we generally recommend against this anyway: https://qdrant.tech/documentation/guides/multiple-partitions/

### All Submissions:

### New Feature Submissions:


agourlay (approved):
Seems reasonable to me :+1:

PR #3120, opened 2023-12-01
Fix gRPC conversion for named sparse vectors

agourlay:
Fix plus test to enable once the API lands on dev.

### DIFF
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -528,18 +528,7 @@ impl From<NamedVectors> for HashMap<String, segment::data_types::vectors::Vector
         vectors
             .vectors
             .into_iter()
-            .map(|(name, vector)| {
-                let vec = match vector.indices {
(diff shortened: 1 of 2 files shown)


lib/api/src/grpc/conversions.rs
```diff
@@ -528,7 +528,18 @@ impl From<NamedVectors> for HashMap<String, segment::data_types::vectors::Vector
         vectors
             .vectors
             .into_iter()
-            .map(|(name, vector)| (name, vector.data.into()))
+            .map(|(name, vector)| {
```
IvanPleshkov, line 531: Looks strange. The same convertion you may find from grpc vector into rest vector (`convertions.rs`, line 450). Try to understand why `into()` didn't work here

IvanPleshkov: @agourlay it seems change:
```
- .map(|(name, vector)| (name, vector.data.into()))
+ .map(|(name, vector)| (name, vector.into()))
```
Is enought

IvanPleshkov (approved):

PR #3123, opened 2023-12-01
Sparse vector grpc configs

IvanPleshkov:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/conversions.rs
+++ b/lib/storage/src/content_manager/conversions.rs
@@ -1,4 +1,5 @@
 use collection::operations::conversions::sharding_method_from_proto;
+use collection::operations::types::SparseVectorsConfig;
 use tonic::Status;
 
 use crate::content_manager::collection_meta_ops::{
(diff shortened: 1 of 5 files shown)


agourlay:
I need a `basic_grpc_api_test.sh` :)

lib/storage/src/content_manager/conversions.rs
```diff
@@ -30,9 +31,11 @@ impl TryFrom<api::grpc::qdrant::CreateCollection> for CollectionMetaOperations {
             CreateCollection {
                 vectors: match value.vectors_config.and_then(|config| config.config) {
                     Some(vector_config) => vector_config.try_into()?,
-                    None => return Err(Status::invalid_argument("vectors config is required")),
```
agourlay, line 33: where is the validation taking place assuring that `vectors` OR `sparse_vectors` are defined for gRPC?

IvanPleshkov: As TODO comment, it's an untrivial moment. We have to validate 2 separate fields with 10 another ones with regular validation - manual implementation would be large =) Moreover, after discussion with @generall, collection without any vectors are valid. added one more todo comment here

IvanPleshkov:
> I need a `basic_grpc_sparse_api_test.sh` :)

done

tests/basic_sparse_grpc_test.sh
```diff
@@ -0,0 +1,159 @@
...
+$docker_grpcurl -d '{}' $QDRANT_HOST qdrant.Collections/List
+
+$docker_grpcurl -d '{
+  "collection_name": "test_sparse_collection",
+  "wait": true,
+  "ordering": null,
+  "points": [
+    {
+      "id": { "num": 1 },
+      "vectors": {
```
agourlay, line 35: can you please explain me what are those 3 layers on `vectors` respectively?

lib/storage/src/content_manager/conversions.rs
agourlay: sorry I am a bit a confused now.
if collection without any vectors are valid, do still need a TODO?

tests/basic_sparse_grpc_test.sh
generall: surprisingly enough, that's not new. In old version for named vector you would get same. Price we should pay for backward compatibility

generall (approved):
I am not sure is this changes are covering all places, but I think it is good to merge

PR #2797, opened 2023-10-11
Quantization storage as separate entity

IvanPleshkov:
Currently, quantized data is a part of `VectorStorage`. However, vector storage does not use any information from quantized data and just provides access to quantized data to the higher-level logic. It makes `VectorStorage` trait overcomplicated. And if sparse vectors are integrated as a new variant of vector storage enum, it will make an integration difficulties.

So in this PR I propose to separate vector storage and quantized data as separate entities. It makes code simpler to read and support. Nice bonus, in this PR more lines were deleted than added =)

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your core changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #3133, opened 2023-12-03
build HNSW with quantized data

generall:
This is a hot-fix for an error introduced in https://github.com/qdrant/qdrant/pull/2797

Proper solution should require full refactoring of https://github.com/qdrant/qdrant/blob/4b51fa0fefb61746eebc8e1a42fceff406b118fe/lib/segment/src/segment_constructor/segment_builder.rs#L184
so it won't rely on monkey-patches of the segment objects

### DIFF
--- a/lib/segment/src/index/plain_payload_index.rs
+++ b/lib/segment/src/index/plain_payload_index.rs
@@ -305,12 +305,9 @@ impl VectorIndex for PlainIndex {
 
     fn set_quantized_vectors(
         &mut self,
-        quantized_vectors: Option<Arc<AtomicRefCell<QuantizedVectors>>>,
+        _quantized_vectors: Option<Arc<AtomicRefCell<QuantizedVectors>>>,
     ) {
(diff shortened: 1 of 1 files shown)


lib/segment/src/index/plain_payload_index.rs
```diff
@@ -301,6 +302,16 @@ impl VectorIndex for PlainIndex {
     fn update_vector(&mut self, _id: PointOffsetType, _vector: VectorRef) -> OperationResult<()> {
         Ok(())
     }
+
+    fn set_quantized_vectors(
```
IvanPleshkov, line 306: If this function is called,  `quantized_vectors` is always `Some`. It's possible to configure qdrant in some way to trigger this function (because we can use plain index for non-appendable segments and configure quantization simultaneously). I find it a correct behavior and propose to do nothing. I can add support of quantization for plain index

IvanPleshkov (approved):

PR #3142, opened 2023-12-04
Estimate sparse vector dimension for optimizer

agourlay:
The indexer optimizer decides if a segment should be rebuilt based on the vector sizes in bytes computed from the vector dimension.

The current implementation for sparse vectors returns zero which prevents the optimizer from running.

This PR proposes a simple scheme to compute an estimation based on the average size of sparse vectors.

## Topics for further discussion:

- a segment rebuild all its indices regardless of the named vector triggering the rebuild 
  - a.k.a rebuild HNSW because of sparse changes
- the current size estimations in the optimizer are not counting number of of actual vectors per named vectors
  - mix setup dense & sparse will not trigger rebuild with high precision

### DIFF
--- a/lib/segment/src/vector_storage/simple_sparse_vector_storage.rs
+++ b/lib/segment/src/vector_storage/simple_sparse_vector_storage.rs
@@ -127,6 +127,11 @@ impl SimpleSparseVectorStorage {
         Ok(())
     }
 
+    /// Estimate average vector size based on total number of non-zero elements in all vectors.
+    ///
+    /// This is needed because the optimizer relies on the vector dimension * size_of_f32 * point_count to
+    /// trigger reindexing and on_disk data move.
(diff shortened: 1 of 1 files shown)


timvisee:
I like that you use the average dimension to reuse `vector_dim()`.

> - a segment rebuild all its indices regardless of the named vector triggering the rebuild
>   - a.k.a rebuild HNSW because of sparse changes
> - the current size estimations in the optimizer are not counting number of of actual vectors per named vectors
>   - mix setup dense & sparse will not trigger rebuild with high precision

I think these concerns are fine for now.

We should make sure though that we don't get some sort of infinite indexation loop. This can probably be covered by converting segment types into `Special` if that's not already done.

IvanPleshkov (approved):

PR #3143, opened 2023-12-04
Fix optimizer change sparse index on disk

agourlay:
This PR fixes the optimizer to apply the `on_disk` parameter to a sparse vector index once the memap threshold is reached.

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -652,6 +652,7 @@ impl SegmentConfig {
         }
     }
 
+    /// Check if any vector storages are indexed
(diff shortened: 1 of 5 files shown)


lib/segment/src/types.rs
```diff
@@ -677,7 +679,10 @@ impl SegmentConfig {
         self.vector_data
             .values()
             .any(|config| config.storage_type.is_on_disk())
-            || !self.sparse_vector_data.is_empty()
+            || self
```
IvanPleshkov, line 682: Nice catch! Forgot about it

IvanPleshkov (approved):

PR #3156, opened 2023-12-04
Better shard transfer init

generall:
Prevents

```
2023-12-04T21:07:20.019429Z ERROR qdrant::tonic::logging: gRPC /qdrant.PointsInternal/Sync unexpectedly failed with Internal error "Service internal error: No target shard 1 found for update" 0.253019    
```

on shard reocvery.

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -327,12 +327,7 @@ impl Collection {
                         |state| {
                             state
                                 .get_peer_state(&this_peer_id)
-                                .map_or(false, |peer_state| {
-                                    matches!(
-                                        peer_state,
-                                        ReplicaState::Partial | ReplicaState::PartialSnapshot
-                                    )
-                                })
+                                .map_or(false, |peer_state| peer_state.is_partial_like())
                         },
                         defaults::CONSENSUS_META_OP_WAIT,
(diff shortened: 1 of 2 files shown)


lib/collection/src/collection/shard_transfer.rs
```diff
@@ -307,14 +307,35 @@ impl Collection {
...
+                shard_transfer_registered
+                    && replica_set.wait_for_state_condition_sync(
+                        |state| {
+                            state
+                                .get_peer_state(&this_peer_id)
+                                .map_or(false, |peer_state| {
+                                    matches!(
+                                        peer_state,
+                                        ReplicaState::Partial | ReplicaState::PartialSnapshot
+                                    )
```
timvisee, line 334: Maybe we can add some sort of `is_partial` check to cover all partial states, also ones that we might add in the future.

timvisee (approved):

lib/collection/src/collection/shard_transfer.rs
timvisee: I'll quickly add this myself :+1:

timvisee: Implemented in 0b1972d07dc05ce24fd4a44f309db1d44c0fb16f.

PR #3161, opened 2023-12-05
Plain search by ids in sparse vector inverted index

agourlay:
This PR adds a new mechanism for plain search on the sparse inverted index for a set of ids.

Supersedes  https://github.com/qdrant/qdrant/pull/3113


IvanPleshkov (approved):
Rockstar's PR!

PR #3166, opened 2023-12-05
Fix sparse mmap snapshot

agourlay:
Snapshots for mmap sparse index could not be restored because a mandatory configuration files was not captured.

I validated the fix manually.


IvanPleshkov (approved):

timvisee:
I've just confirmed that this works.

Awesome!

PR #3160, opened 2023-12-05
Optimizer detects on_disk update sparse vector index

agourlay:
This PR fixes the optimizer to detect when the sparse vector index `on_disk` field changed.

TODO: currently testing on BFB, opening for CI run and early reviews

PR #3167, opened 2023-12-05
Fix infinite optimization loop with `on_disk` and `mmap_threshold` fighting

timvisee:
It is possible to create an infinite optimization loop when using the `on_disk` and `mmap_threshold` properties together. They'll start fighting which each other and the collection will remain yellow forever.

Setting `on_disk: false` and `mmap_threshold: 1000` will trigger this. The indexing optimizer will put vectors on disk because the threshold is reached. The config mismatch optimizer will try to remove vectors from disk again because `on_disk: false` is configured.

Before this PR, you can create this situation with:

```rust
bfb -n 20000 --dim 2048 --on-disk-vectors false --mmap-threshold 10000 --indexing-threshold 0
```

This PR fixes this issue by preferring `on_disk` over `mmap_threshold`. More specifically, it:
- doesn't put vectors on disk anymore because of the threshold if `on_disk` was explicitly set
- prevents the indexing optimizer triggering due to this threshold if `on_disk` is explicitly set for vectors

I see this is hotfix to get it in for the release. I'd prefer to re-architecture our optimization logic a bit in the near future to prevent situations like this.

I've manually tested this extensively.

I discovered this issue when working on <https://github.com/qdrant/qdrant/pull/3160#pullrequestreview-1765853907>.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
@@ -125,21 +125,34 @@ impl IndexingOptimizer {
                 let are_all_vectors_indexed = segment_config.are_all_vectors_indexed();
                 let is_any_on_disk = segment_config.is_any_on_disk();
 
-                let big_for_mmap = vector_size
-                    >= self
-                        .thresholds_config
-                        .memmap_threshold
-                        .saturating_mul(BYTES_IN_KB);
                 let big_for_index = vector_size
                     >= self
                         .thresholds_config
                         .indexing_threshold
                         .saturating_mul(BYTES_IN_KB);
+                let big_for_mmap = vector_size
+                    >= self
+                        .thresholds_config
+                        .memmap_threshold
+                        .saturating_mul(BYTES_IN_KB);
 
-                let require_indexing = (big_for_mmap && !is_any_on_disk)
-                    || (big_for_index && !are_all_vectors_indexed);
+                // Check whether all vectors have `on_disk` specified, either true or false,
+                // because then we don't want to override it with the mmap threshold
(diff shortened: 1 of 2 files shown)


agourlay:
> They'll start fighting which each other and the collection will remain yellow forever.

Is it possible to create simple unit test for this case?

timvisee:
> > They'll start fighting which each other and the collection will remain yellow forever.
> 
> Is it possible to create simple unit test for this case?

Not simple, but yes, it's something we should have!

lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
```diff
@@ -121,25 +121,71 @@ impl IndexingOptimizer {
...
+                            on_disk_config && !is_on_disk
+                        } else {
+                            is_big_for_mmap && !is_on_disk
+                        };
+
+                        if optimize_for_index || optimize_for_mmap {
+                            require_optimization = true;
+                        }
+                    }
+                }
```
agourlay, line 155: If `require_optimization` is `true`, we don't need to look at the sparse vectors because they will be also handled by the optimizer.

WDYT about an early exit there

timvisee: Sure! Noticed this, but we implemented this quite hastily.

I'll make the change anyway :+1:

timvisee: Implemented in <9c7852438e7f494a3a53ecbf1a824d2b4457e88e>. We can improve it further after the release.

agourlay (approved):

generall (approved):

PR #3171, opened 2023-12-06
Fix bench for mmap sparse search

agourlay:
The latest changes for opening the sparse inverted index broke the mmap benchmark.

```
Benchmarking sparse-vector-search-group/mmap-inverted-index: Warming up for 3.0000 sthread 'main' panicked at lib/segment/benches/sparse_index_search.rs:89:13:
  left: 0
 right: 10
```

This because the index was built manually without generating the underlying sparse configuration file.

### DIFF
--- a/lib/segment/benches/sparse_index_search.rs
+++ b/lib/segment/benches/sparse_index_search.rs
@@ -64,22 +64,21 @@ fn sparse_vector_index_search_benchmark(c: &mut Criterion) {
 
     // mmap inverted index
     let mmap_index_dir = Builder::new().prefix("mmap_index_dir").tempdir().unwrap();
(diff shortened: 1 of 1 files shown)


lib/segment/benches/sparse_index_search.rs
```diff
@@ -124,23 +123,6 @@ fn sparse_vector_index_search_benchmark(c: &mut Criterion) {
         })
     });
 
-    group.bench_function("plain-storage", |b| {
```
agourlay, line 127: this seems like a duplication from the benchmark just above it

IvanPleshkov (approved):
looks better, calling `build_index` is a more honest way to initialize data

PR #3186, opened 2023-12-09
Add plain_search_limit search parameter

keremnc:
We observe a ~20% drop in recall when using `indexed_only: true` in our searches in a collection with `default_segment_number: 5`.
I believe what is happening is:
1. An indexed segment gets selected for another round of optimization
2. A proxy segment is created where the original indexed segment is `wrapped` and a new temporary plain segment is `write`
3. The `write` segment accrues new vectors
4. The `is_search_optimized` function is eventually `false` for the entire ProxySegment because `num_vectors` of `VectorDataInfo` is summed across both `write` and `wrapped` segments but `num_indexed_vectors` only includes the value from the optimized `wrapped` segment.

In the above case, really only the plain `write` segment of the proxy segment should be excluded from searches, not the optimized `wrapped` segment. 

This PR is a proposal to introduce a new `plain_search_limit` search parameter which is applied at a lower level. It:
1. Sets a limit on the number of vectors that will be scanned when performing a plain index search.
2. Pushes the logic down into the vector index, allowing ProxySegment to propagate reads to its underlying segments properly.
3. Decouples index creation-time `indexing_threshold` and `full_scan_threshold` parameters with search-time parameters to allow for better control and fine-tuning

I did not make any changes to `indexed_only` logic. 

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/grpc/docs.md
+++ b/docs/grpc/docs.md
@@ -3008,6 +3008,7 @@ For example, if `oversampling` is 2.4 and `limit` is 100, then 240 vectors will 
 | exact | [bool](#bool) | optional | Search without approximation. If set to true, search may run long but with exact results. |
 | quantization | [QuantizationSearchParams](#qdrant-QuantizationSearchParams) | optional | If set to true, search will ignore quantized vector data |
 | indexed_only | [bool](#bool) | optional | If enabled, the engine will only perform search among indexed or small segments. Using this option prevents slow searches in case of delayed index, but does not guarantee that all uploaded vectors will be included in search results |
+| plain_search_limit | [uint64](#uint64) | optional | If set, the engine will perform truncated searches up to this limit when scanning plain segments. Using this option allows an upper bound to be set on the cost of searching plain segments exhaustively when indexing throughput cannot keep up with ingest throughput, but does not guarantee that all uploaded vectors will be included in search results |
 
 
 
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -6672,6 +6672,13 @@
             "description": "If enabled, the engine will only perform search among indexed or small segments. Using this option prevents slow searches in case of delayed index, but does not guarantee that all uploaded vectors will be included in search results",
             "default": false,
             "type": "boolean"
+          },
+          "plain_search_limit": {
+            "description": "If set, the engine will perform truncated searches up to this limit when scanning plain segments. Using this option allows an upper bound to be set on the cost of searching plain segments exhaustively when indexing throughput cannot keep up with ingest throughput, but does not guarantee that all uploaded vectors will be included in search results",
+            "type": "integer",
+            "format": "uint",
+            "minimum": 0,
+            "nullable": true
           }
         }
       },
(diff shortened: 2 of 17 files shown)


timvisee (commented):
First of all, thank you for your contribution!

Before doing a full review, I do have a question. What is the reason for adding the `plain_search_limit` parameter, and not keeping this internal-only?

Can't we achieve something similar by restructuring the internals slightly, where we will search the `wrapped` segment (assuming it is indexed) but not `write`?

keremnc:
@timvisee, I think the fact that `indexed_only` uses a combination of `indexing_threshold` and `full_scan_threshold` to determine which "plain" (in quotes because the logic is not solid, as mentioned in PR description) segments to search is too rigid. 
The problems with performing searches over plain segments is clear:
- the asymptotic complexity is O(n) compared to O(log n) for HNSW
- maximum size (and thus search latency) of a plain index cannot be bounded like it can for HNSW graphs
- bursts of writes and optimizer selection/scheduling can lead to unpredictably-sized plain segments

This means scanning over plain indexes is probably the component of the search which has the most unpredictable latency/perf characteristics and thus allowing for an upper bound on computational cost here is a reasonable approach to improve stability.

In practice, one might say "I am OK with doing up to N iterations through a plain segment on each search, because on my hardware, this bounds per-segment search latency at X us". 
It should not also imply that `indexing_threshold` should also be set at the same value - I may be OK with 30k iterations through the plain segment but I don't want to wait for it to grow to 30k vectors before being eligible for optimization, I would simply like to relax my requirements for the optimizer completing in time in order to keep recall high.

It's also different from `full_scan_threshold` - this is the cutoff below which it's actually more performant to perform plain instead of HNSW search, not an upper bound after which the operator deems the search to expensive to perform.

So I did not touch `indexed_only` because I did not want to break any existing logic, but it could be seen as a special case of `plain_search_limit: 0` if migrated to the logic in my proposal.

generall:
Hi @keremnc, thanks for the suggestion!

I would agree that behavior of proxy segment is not optimal in this situation. The wrapped data should always stay available for fast search, even if the `write` segment is big.

But I don't think I like the solution with additional parameter. It is quite confusing and hard to understand what exactly it is doing even for me.

I would prefer a solution where `ProxySegment` internally decides, based on the same logic, should it search through both, none or only wrapped segments.

keremnc:
Hey @generall,
> based on the same logic

which logic are you referring to here?

generall:
> > based on the same logic
> 
> which logic are you referring to here?

the logic currently used to determine if the whole segment should be searched or not

keremnc:
I think there are two main drawbacks to the current logic, aside from the issue with ProxySegment not performing the recursive computation correctly:
1. The threshold used to determine if a plain segment is searchable with `indexed_only` is `max(indexing_threshold, full_scan_threshold)`. I don't think either of these values (and especially not the combination of both) can accurately represent how much brute-force search a user can tolerate, while also being used to configure when/how the indexes should be built.
2. As plain segments grow, they will eventually cause a drop-off in recall. For example, if `indexing_threshold` is set to 10KiB, and a plain segment is 9KiB, it is still searchable. However, once it breaches 10KiB and is eligible for optimization, the whole segment becomes unavailable to search with `indexed_only`. Allowing the user's tolerance for plain searches to manifest as "partial" searches, as in my proposal, can help mitigate this recall drop-off.

Introducing a separate configuration `plain_search_limit` addresses (1.) and pushing the logic down into the scorer addresses (2.).

I am not completely attached to this solution though. Our goal is simply to maximize recall while placing a strict upper-bound on latency, even in the face of unpredictable and spiky write traffic. Would you be more receptive to implementing `plain_search_limit` as a collection param, and tweaking the logic of `indexed_only` to match (1.)?

PR #3191, opened 2023-12-10
introducing two new payload schema type: integer_lookup & integer_range

dzhao:
introduce two new index types:
integer_lookup: it only creates a hash index
integer_range: it only creates a btree index
This is useful for two use cases where both payload indexing and hnsw
indexing will be reduced:
1. for example: posts with a payload user_id. In this case, we just need
   to lookup a user_id, hence no need for a btree index
2. for example: posts with a payload creation_ts. In this case, we just
   need to lookup posts based on a range filter, hence no need for a
   hash index

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


timvisee (commented):
Great idea, these make a lot of sense. Thanks for implementing it!

I do wonder whether we want to take this route though, of defining a new index type for each variant. Instead we may be able to use the same `"int"` but with some `{"lookup": true, "range": true}` properties to achieve the same, even though we don't have infrastructure for this at this time.

Question, if I may: have you tested these new types in practice? And if possible, could you share some results?

dzhao:
will test the new types soon. 
I would slightly prefer flatten the type instead of adding attributes. "lookup", "range" won't be applicable to geo or text based index.

dzhao:
tested this. especially integer_lookup would speed up hnsw building quite significantly since it no longer needs to build subgraphs for btree ranges. Let me know if you are ok with merging. I am open to have better configuration options

timvisee:
> tested this. especially integer_lookup would speed up hnsw building quite significantly since it no longer needs to build subgraphs for btree ranges. Let me know if you are ok with merging. I am open to have better configuration options

Cool! Thanks for letting us know.

In the mean time I've implemented a parameterized variant in <https://github.com/qdrant/qdrant/pull/3380>. It would be fantastic if you could let us know if you have any comments on it.

Everything should be have the same, instead for the API interface. I've also cherry-picked your test into it. I still have to confirm it's actually compatible with our API definitions and clients, so that question remains for a bit.

timvisee:
Closing this since <https://github.com/qdrant/qdrant/pull/3380> is merged. Thanks a lot for your initial efforts on this @dzhao!

PR #3192, opened 2023-12-10
Rename `VectorType` to `DenseVector`

IvanPleshkov:
`VectorType` is an obsolete name. After sparse vector release it's a dense vector.

Also was renamed:
`NamedVectorStruct::Named` -> `NamedVectorStruct::Dense`
`RecommendExample::Vector` -> `RecommendExample::Dense`

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -6136,14 +6136,14 @@
             }
           },
           {
-            "$ref": "#/components/schemas/NamedVector"
+            "$ref": "#/components/schemas/NamedDenseVector"
           },
           {
             "$ref": "#/components/schemas/NamedSparseVector"
           }
         ]
       },
-      "NamedVector": {
+      "NamedDenseVector": {
         "description": "Vector data with name",
         "type": "object",
         "required": [
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -6136,14 +6136,14 @@
             }
           },
           {
-            "$ref": "#/components/schemas/NamedDenseVector"
+            "$ref": "#/components/schemas/NamedVector"
           },
           {
             "$ref": "#/components/schemas/NamedSparseVector"
           }
         ]
       },
-      "NamedDenseVector": {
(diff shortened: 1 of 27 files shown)


docs/redoc/master/openapi.json
```diff
@@ -6136,14 +6136,14 @@
...
           {
-            "$ref": "#/components/schemas/NamedVector"
+            "$ref": "#/components/schemas/NamedDenseVector"
           },
           {
             "$ref": "#/components/schemas/NamedSparseVector"
           }
         ]
       },
-      "NamedVector": {
```
agourlay, line 6146: This looks like a breaking change for the clients generation.
This will change the types in the users code.
WDYT?

timvisee: >  This will change the types in the users code.

Correct. I'm afraid we can't do that.

We could make the change internally and do a serde [rename](https://serde.rs/container-attrs.html), though that will make the two inconsistent.

generall: +, don't change this

generall: But it is not about serde, it is about type names in clients. If we generate clients with this renaming, old code on the client side won't work.

IvanPleshkov: Okay, will revert this type. Type rename is unsafe. What about enum variant renaming?

timvisee: > But it is not about serde, it is about type names in clients. If we generate clients with this renaming, old code on the client side won't work.

What I meant was that we can consider to rename these types internally, and utilize serde rename to keep our API specification consistent.

> What about enum variant renaming?

I'm afraid it's the same story. If it results in a change in our API specifications, it's breaking.

IvanPleshkov: Reverted type name.
It seems enum variants remanings do not affect opeanpi generated models

agourlay (approved):

timvisee (approved):

Issue #3193, opened 2023-12-11
Unable to upgrade cluster from 1.6.0 to 1.7.0


mautini:
When upgrading a Qdrant cluster from 1.6.0 to 1.7.1, the new pods are not able to start correctly while the old ones are still running and so the upgrade cannot be done without downtime.

## Current Behavior
When performing

## Steps to Reproduce
1. Deploy a Qdrant cluster using the Qdrant official helm chart in 1.6.0
2. Bump the Qdrant version in the helm chart, or use the new helm chart version
3. Apply the changes in k8s

## Expected Behavior
New node can start without issue and the rolling update could finish successfully, without downtime.

## Context (Environment)
- Qdrant versions: v1.6.0 to v1.7.0
- 2 nodes running in a k8s cluster. The Qdrant cluster is deployed using official Qdrant Helm Chart and ArgoCD

## Detailed Description

I have a Qdrant cluster (2 nodes) running in my k8s cluster. I tried to upgrade it to version 1.7.0 by performing a rolling upgrade of the Statefulset but the new pods are not able to start correctly and so the upgrade cannot finish. This means I cannot perform the upgrade without downtime. 

Here is the error logs I have for the new pod (in 1.7.0): 
```
2023-12-11T10:10:15.860912Z ERROR qdrant::common::health: GetCommitIndex request failed: Error in closure supplied to transport channel pool: status: Unimplemented, message: "", details: [], metadata: MetadataMap { headers: {"date": "Mon, 11 Dec 2023 10:10:15 GMT", "content-type": "application/grpc", "content-length": "0"} }
2023-12-11T10:10:15.861237Z  WARN qdrant::common::health: Not enough cluster nodes responded to GetConsensusCommit request: required 1,                 responded 0 out of 2
```

timvisee:
Thank you for reporting this! We've just confirmed this issue. We'll release a patch version ASAP to resolve this.

A temporary workaround might be to disable `/readyz` checks for now. If using `helm` you could:
- Disable the readiness probe for the duration of the upgrade:
  ```bash
  helm upgrade qdrant qdrant/qdrant --version 0.7.0 --set replicaCount=3 --set readinessProbe.enabled=false
  ```
- If the update is already in progress and stuck, delete the stuck Pod to restart it with the new config, e.g:
  ```bash
  kubectl delete pod qdrant-2
  ```
- Once all Pods upgraded, activate the probe again:
  ```bash
[...]

PR #3196, opened 2023-12-11
Fix `/readyz` when upgrading cluster from 1.6 to 1.7

ffuugoo:
Fixes <https://github.com/qdrant/qdrant/issues/3193>.

When doing rolling upgrade from Qdrant v1.6.* to v1.7.*, the `/readyz` check would fail _indefinitely_, practically preventing graceful rolling upgrades in Kubernetes. 🤦‍♀️

The issue is that v1.6.* (and earlier) versions simply don't have gRPC APIs required for the `/readyz` check in v1.7.*, and "unimplemented" errors aren't handled properly.

This PR implements "unimplemented" errors handling during `/readyz`, allowing upgrade to v1.7.*.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/common/health.rs
+++ b/src/common/health.rs
@@ -7,7 +7,7 @@ use std::{cmp, panic, thread};
 
 use api::grpc::qdrant::qdrant_internal_client::QdrantInternalClient;
 use api::grpc::qdrant::{GetConsensusCommitRequest, GetConsensusCommitResponse};
-use api::grpc::transport_channel_pool::{self, TransportChannelPool};
+use api::grpc::transport_channel_pool::{self, RequestError, TransportChannelPool};
 use collection::shards::shard::ShardId;
 use collection::shards::CollectionId;
 use common::defaults;
@@ -181,6 +181,8 @@ impl Task {
 
         let peer_address_by_id = self.consensus_state.peer_address_by_id();
 
+        let mut unimplemented_errors_count = 0;
+
         let mut requests = peer_address_by_id
             .iter()
             .filter_map(|(&peer_id, uri)| {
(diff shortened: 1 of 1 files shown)


ffuugoo:
I'm testing this manually right now, will mark PR as ready if/once it works properly.

src/common/health.rs
```diff
@@ -181,6 +181,8 @@ impl Task {
 
         let peer_address_by_id = self.consensus_state.peer_address_by_id();
 
+        let mut unimplemented_errors_count = 0;
```
timvisee, line 184: Maybe we can put a comment before here, something like:

```suggestion
        // Endpoint we check is not available before Qdrant 1.7.0, allow checks on nodes with earlier versions to fail by counting them here
        let mut unimplemented_errors_count = 0;
```

timvisee (commented):
Nice fix! I didn't test it myself yet, but it seems simple and effective.

src/common/health.rs
```diff
@@ -181,6 +181,9 @@ impl Task {
 
         let peer_address_by_id = self.consensus_state.peer_address_by_id();
 
+        // Endpoint we check is not available before Qdrant 1.7.0, allow checks on nodes with earlier versions to fail by counting them here
```
agourlay, line 184: Should we have a clear TODO to remove this trick in Qdrant 1.8.0?

timvisee: We can do, but it doesn't hurt to keep it here for a bit longer.

timvisee:
Rather than doing a simple subtraction on the required nodes when one returns not-implemented, I think doing this a little different will make it more reliable.

With the current approach, if you have 5 nodes and the first two return not-implemented, it shortcuts immediately assuming everything is okay. But if the other two nodes are up-to-date and return a wildly different cluster status, you have a problem.

Instead, if a node returns a not-implemented failure, I think we should decrease the list of nodes to consider and adjust all counters accordingly. In that case you make sure you query enough nodes running 1.7 to be confident everything is sound.

There is a downside, when updating the first node to 1.17, it will have to go to all nodes asking their status. This improves once more nodes reach 1.17 though, and completely disappears for further versions. But I think this is fine to do to ensure `/readyz` is more reliable.

I've partially discussed this with Roman already, but wanted to mention it publicly here as well.

generall (approved):

timvisee:
I'll publish and release the patch version in about 10 minutes :+1:

PR #3215, opened 2023-12-12
Fix `/readyz` if nodes are unavailable

### LINKED ISSUE #3193: Unable to upgrade cluster from 1.6.0 to 1.7.0

ffuugoo:
This PR removes strict check that node received `cluster_size / 2` successful `GetConsensusCommit` responses during `/readyz` check to allow `/readyz` check to pass, when not enough nodes are available.

Fixes #3193.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/actix/api/service_api.rs
+++ b/src/actix/api/service_api.rs
@@ -110,9 +110,9 @@ async fn livez() -> impl Responder {
 }
 
 #[get("/readyz")]
-async fn readyz(ready: web::Data<Option<Arc<health::HealthChecker>>>) -> impl Responder {
-    let is_ready = match ready.as_ref() {
-        Some(ready) => ready.check_ready().await,
+async fn readyz(health_checker: web::Data<Option<Arc<health::HealthChecker>>>) -> impl Responder {
+    let is_ready = match health_checker.as_ref() {
+        Some(health_checker) => health_checker.check_ready().await,
         None => true,
     };
 
--- a/src/actix/mod.rs
+++ b/src/actix/mod.rs
@@ -49,7 +49,7 @@ pub async fn index() -> impl Responder {
 pub fn init(
     dispatcher: Arc<Dispatcher>,
     telemetry_collector: Arc<tokio::sync::Mutex<TelemetryCollector>>,
-    ready: Option<Arc<health::HealthChecker>>,
+    health_checker: Option<Arc<health::HealthChecker>>,
     settings: Settings,
 ) -> io::Result<()> {
     actix_web::rt::System::new().block_on(async {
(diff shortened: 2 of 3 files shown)


generall (approved):

timvisee (approved):
I tried to test the scenarios I could come up with manually in a very simple setup, and did not find any issues.

I don't really like the error messages that come up in 1.16 when doing this, but that's not really something we can resolve.

timvisee:
@ffuugoo Please feel free to merge right away once ready (unless there's other significant changes). Then we can start releasing :+1:

ffuugoo:
I think I figured (and workaround) another possible corner-case:
- imagine cluster of 5 nodes (and all nodes are in the cluster *already*)
- we restart all nodes in sequence
- first node starts, tries to query `cluster_commit_index`
  - because other nodes are unavailable yet, it just skips the check and returns ready
- second node starts, treis to query `cluster_commit_index`
  - only first node responds, so `cluster_commit_index == first_node_commit_index`
- if `second_node_commit_index < first_node_commit_index`, then second node would wait till `second_node_commit_index >= first_node_commit_index`
- but 2 nodes in cluster of 5 nodes can't elect a leader, so second node may *never* catch up with the first one
- and third node would *never* be started, because second node is not ready

PR #3225, opened 2023-12-13
Validate sparse indices in grpc search

IvanPleshkov:
Add custom validation step in grpc while conversion. We cannot use default field-based validation because sparse requests require checking 2 separate fields. I find the custom implementation of struct validation bad because It affects another validated field.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/api/build.rs
+++ b/lib/api/build.rs
@@ -164,7 +164,6 @@ fn configure_validation(builder: Builder) -> Builder {
             ("CreateFieldIndexCollection.field_name", "length(min = 1)"),
             ("DeleteFieldIndexCollection.collection_name", "length(min = 1, max = 255)"),
             ("DeleteFieldIndexCollection.field_name", "length(min = 1)"),
-            // TODO(sparse) validate sparse vector for `SearchPoints`
             ("SearchPoints.collection_name", "length(min = 1, max = 255)"),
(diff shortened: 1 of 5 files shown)


agourlay:
Can you please add a quick test for the new validation in `basic_sparse_grpc_test.sh` to make sure that it works :)

agourlay (changes requested):
needs a test :)

agourlay (approved):

lib/api/build.rs
timvisee, line 1: Can we also implement actual validation here, instead of doing validation on conversion?

I know it's somewhat annoying to do here, but it's where we do it for everything else as well.

See some examples here: https://github.com/qdrant/qdrant/blob/33094335dd06040d325946e66afd079afcb98b32/lib/api/src/grpc/validate.rs

IvanPleshkov: We have one:
https://github.com/qdrant/qdrant/blob/33094335dd06040d325946e66afd079afcb98b32/lib/api/src/grpc/validate.rs#L195
If you find an additional check on vector conversion wrong, mark please PR as `request changes`

timvisee: You're totally right. Thanks!

timvisee (approved):

PR #3231, opened 2023-12-14
Faster sparse vectors plain search

agourlay:
This PR optimizes the plain search through the inverted index for sparse vectors.

```
sparse-vector-search-group/inverted-index-filtered-plain
                        time:   [134.17 ms 134.44 ms 134.70 ms]
                        change: [-34.599% -34.462% -34.317%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
...
sparse-vector-search-group/plain-filtered-payload-index
                        time:   [128.26 ms 128.41 ms 128.56 ms]
                        change: [-35.495% -35.337% -35.185%] (p = 0.00 < 0.05)
                        Performance has improved.


```

The insight is that by sorting the input ids, we can actually fully leverage the posting list iterator by performing binary search on a shrinking space.

### DIFF
--- a/lib/sparse/src/index/search_context.rs
+++ b/lib/sparse/src/index/search_context.rs
@@ -57,6 +57,10 @@ impl<'a> SearchContext<'a> {
 
     /// Plain search against the given ids without any pruning
     pub fn plain_search(&mut self, ids: &[PointOffsetType]) -> Vec<ScoredPointOffset> {
+        // sort ids to fully leverage posting list iterator traversal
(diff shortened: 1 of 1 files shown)


IvanPleshkov (approved):

agourlay:
It seems the Python congruence tests do not like this change.
I need to investigate this before merging.

IvanPleshkov (approved):

PR #3235, opened 2023-12-16
Custom thread priorities on Linux

timvisee:
Set custom priorities for consensus and HNSW build threads on Linux. This should not affect regular operation but will improve reliability in extreme cases. For example, this can help prevent consensus lagging behind if the machine if very busy building HNSW.

For HNSW build threads the priority is lowered, since these usually fully saturate CPU cores lowering the priority attempts to make affecting ongoing updates/searches less likely. The consensus threads the priority is raised, to allow handling consensus updates as soon as possible. Raising the thread priority likely fails because it requires special permissions on most systems, so it soft-fails with a debug message.

More specifically, on Linux this tries to set:
- consensus threads: nice=-10
- HNSW build threads: nice=10
- all other threads: nice=0 (unchanged)

This is only implemented for Linux and not other Unix-like platforms. POSIX threads share priority/nice values across all process threads. Linux breaks this behavior and allows setting per-thread priority/nice values[^1][^2]. I'm not sure about Windows.

I've done a basic `bfb` benchmark to see the effect of this.

Search timings during HNSW building without thread priorities:

```bash
$ bfb --skip-create --skip-upload --skip-wait-index --search --p9=4 -n5000
 [00:01:36] ████████████████████ [51.7576/s] 5000/5000 (eta:0s)
--- Server timings ---
Min search time: 0.030139231
Avg search time: 0.037740009691600004
p95 search time: 0.042813196
p99 search time: 0.043735879
p999 search time: 0.045643002
p9999 search time: 0.050678771
Max search time: 0.050678771
```

Search timings during HNSW building with thread priorities:

```bash
$ bfb --skip-create --skip-upload --skip-wait-index --search --p9=4 -n5000
 [00:01:31] ████████████████████ [54.4541/s] 5000/5000 (eta:0s)
--- Server timings ---
Min search time: 0.031290399
Avg search time: 0.03583264708799998
p95 search time: 0.042004593
p99 search time: 0.042975104
p999 search time: 0.043672693
p9999 search time: 0.043811394
Max search time: 0.043811394
```

[^1]: https://linux.die.net/man/7/pthreads
[^2]: https://linux.die.net/man/2/setpriority

### All Submissions:

### New Feature Submissions:

4. [x] Have you successfully ran tests with your changes locally?

### DIFF
--- a/lib/common/common/src/cpu.rs
+++ b/lib/common/common/src/cpu.rs
@@ -1,0 +1,71 @@
+use thiserror::Error;
+use thread_priority::{
+    get_current_thread_priority, set_current_thread_priority, ThreadPriority, ThreadPriorityValue,
+};
+
+#[derive(Error, Debug)]
+pub enum ThreadPriorityError {
+    #[error("Failed to get thread priority: {0:?}")]
+    GetThreadPriority(thread_priority::Error),
+    #[error("Failed to set thread priority: {0:?}")]
+    SetThreadPriority(thread_priority::Error),
+    #[error("Got unexpected thread priority type, cannot change: {0:?}")]
+    UnexpectedThreadPriorityType(ThreadPriority),
+    #[error("Thread priority is left unchanged to keep it in bounds")]
+    UnchangedNice,
+    #[error("Failed to parse niceness value: {0}")]
+    ParseNice(&'static str),
+}
+
+/// Make current thread lower priority (renice=+1).
+///
+/// Only has an effect on Unix platforms, ignored on other platforms.
+pub fn current_thread_lower_priority() -> Result<(), ThreadPriorityError> {
+    current_thread_renice(1)
+}
+
+/// Make current thread high priority (renice=-10).
+///
+/// Only has an effect on Unix platforms, ignored on other platforms.
+pub fn current_thread_high_priority() -> Result<(), ThreadPriorityError> {
+    current_thread_renice(-10)
+}
+
+/// Update thread priority and niceness.
+///
+/// Only has an effect on Unix platforms, ignored on other platforms.
+fn current_thread_renice(relative_nice: i8) -> Result<(), ThreadPriorityError> {
+    #[cfg(not(unix))]
+    {
+        return Ok(());
+    }
+
+    #[cfg(unix)]
+    {
+        // Get thread priority value
+        let current =
+            match get_current_thread_priority().map_err(ThreadPriorityError::GetThreadPriority)? {
+                ThreadPriority::Crossplatform(current) => current,
+                thread_priority => {
+                    return Err(ThreadPriorityError::UnexpectedThreadPriorityType(
+                        thread_priority,
+                    ))
+                }
+            };
+
+        // Calculate new niceness, but stay within bounds
+        let old_nice: u8 = current.into();
+        let new_nice = (old_nice as i8).saturating_add(relative_nice).clamp(
+            ThreadPriorityValue::MIN as i8,
+            ThreadPriorityValue::MAX as i8,
+        ) as u8;
+        if old_nice == new_nice {
+            return Err(ThreadPriorityError::UnchangedNice);
+        }
+
+        let new_priority = ThreadPriority::Crossplatform(
+            ThreadPriorityValue::try_from(new_nice).map_err(ThreadPriorityError::ParseNice)?,
+        );
+        set_current_thread_priority(new_priority).map_err(ThreadPriorityError::SetThreadPriority)
+    }
+}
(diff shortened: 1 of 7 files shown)


lib/common/common/src/cpu.rs
```diff
@@ -0,0 +1,41 @@
...
+    // 25% corresponds to a nice value of 10
+    set_linux_thread_priority(25)
+}
+
+/// On Linux, make current thread high priority (nice: -10).
+#[cfg(target_os = "linux")]
+pub fn linux_high_thread_priority() -> Result<(), ThreadPriorityError> {
+    // 75% corresponds to a nice value of 10
+    set_linux_thread_priority(75)
+}
```
ffuugoo, line 27: IIRC, negative values require root.

ffuugoo: https://manned.org/renice.1#head6

> Users other than the superuser may only alter the priority of processes they own. Furthermore, an unprivileged user can only increase the "nice value" (i.e., choose a lower priority) and such changes are irreversible unless (since Linux 2.6.12) the user has a suitable "nice" resource limit (see ulimit(1p) and getrlimit(2)).
>
> The superuser may alter the priority of any process and set the priority to any value in the range -20 to 19. Useful priorities are: 19 (the affected processes will run only when nothing else in the system wants to), 0 (the "base" scheduling priority), anything negative (to make things go very fast).

ffuugoo:
> I'm not sure about Windows.

AFAIK, Windows have "relative" thread priorities. E.g., the whole process is allocated a fixed amount of CPU time, and then it rationed to all process threads proportionally to their priorities or something like that. But I have no idea how to set/change it on Windows.

ffuugoo (commented):
LGTM otherwise

ffuugoo (approved):

lib/common/common/src/cpu.rs
timvisee: Correct. I mentioned this elsewhere and do a soft-fail. But I guess it would be nice to mention it here as well.

ffuugoo: Didn't read the PR description, went straight to the code.... 🙈

timvisee: I've added an extra warning in bb8387e586c8429238f6d86575bed875652259f1. I assume that is enough.

timvisee:
> > I'm not sure about Windows.
> 
> AFAIK, Windows have "relative" thread priorities. E.g., the whole process is allocated a fixed amount of CPU time, and then it rationed to all process threads proportionally to their priorities or something like that. But I have no idea how to set/change it on Windows.

For now I'll just not bother messing with priorities on Windows. I assume that hosting Qdrant in production on Windows is very uncommon.

lib/common/common/src/cpu.rs
```diff
@@ -0,0 +1,47 @@
...
+pub fn linux_low_thread_priority() -> Result<(), ThreadPriorityError> {
+    // 25% corresponds to a nice value of 10
+    set_linux_thread_priority(25)
+}
+
+/// On Linux, make current thread high priority (nice: -10).
+///
+/// # Warning
+///
+/// This is very likely to fail because decreasing the nice value requires special privileges. It
```
agourlay, line 26: Did you check if it works with Docker?

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -661,6 +664,27 @@ impl<TGraphLinks: GraphLinks> VectorIndex for HNSWIndex<TGraphLinks> {
         let pool = rayon::ThreadPoolBuilder::new()
             .thread_name(|idx| format!("hnsw-build-{idx}"))
             .num_threads(max_rayon_threads(self.config.max_indexing_threads))
+            .spawn_handler(|thread| {
+                let mut b = thread::Builder::new();
+                if let Some(name) = thread.name() {
```
agourlay, line 669: Did you check if the thread name shows up in htop?

timvisee: Yes: 
image

I've sourced the above code from [here](https://docs.rs/rayon/1.8.0/rayon/struct.ThreadPoolBuilder.html#examples-1), but have added the call for setting the thread priority.

lib/common/common/src/cpu.rs
timvisee: I did not, but I expect the same behavior. I'll give that a spin now.

timvisee: Yes, the behavior is the same. So that means things work fine :+1:

agourlay (approved):

generall (approved):

PR #3261, opened 2023-12-20
Fix potential data race when confirming consensus operation

timvisee:
I found a potential data race in our logic for waiting on consensus operations.

We used a [broadcast](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html) channel to notify waiters with consensus operations results once they resolve. The problem is that these waiters are not notified if it resolves before subscribing.

If we resubmit an operation we subscribed to the existing broadcast channel. It is easily possible for it to be resolved already, in which case we never get notified.

This PR changes the logic to use a [watch](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) instead. It allows reusing the result value even if we await (subscribe) to it later. In other words, when subscribing we check if we already resolved to a value and start waiting if we have not resolved yet. Because the watch channel needs to hold some value at all times, it uses an `Option`.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/storage/src/content_manager/consensus_manager.rs
+++ b/lib/storage/src/content_manager/consensus_manager.rs
@@ -35,7 +35,7 @@ use crate::types::{
     PeerInfo, RaftInfo,
 };
 
-type ConsensusChannelType = Result<bool, StorageError>;
+type ConsensusResult = Result<bool, StorageError>;
 
 pub mod prelude {
     use crate::content_manager::toc::TableOfContent;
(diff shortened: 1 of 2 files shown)


generall:
>  If we resubmit an operation we subscribed to the existing broadcast channel. It is easily possible for it to be resolved already, in which case we never get notified.


Could you please elaborate on this scenario?
As I can see it, there is `on_consensus_op_apply`, which is locked when we try to insert/resolve operation. 
If `sender` exists within the `on_consensus_op_apply` it is never resolved, because we always remove sender from `on_consensus_op_apply` before resolution

timvisee:
> > If we resubmit an operation we subscribed to the existing broadcast channel. It is easily possible for it to be resolved already, in which case we never get notified.
> 
> Could you please elaborate on this scenario? As I can see it, there is `on_consensus_op_apply`, which is locked when we try to insert/resolve operation. If `sender` exists within the `on_consensus_op_apply` it is never resolved, because we always remove sender from `on_consensus_op_apply` before resolution

}iy're

timvisee:
> > If we resubmit an operation we subscribed to the existing broadcast channel. It is easily possible for it to be resolved already, in which case we never get notified.
> 
> Could you please elaborate on this scenario? As I can see it, there is `on_consensus_op_apply`, which is locked when we try to insert/resolve operation. If `sender` exists within the `on_consensus_op_apply` it is never resolved, because we always remove sender from `on_consensus_op_apply` before resolution

You're right, that was an oversight while working on this. Thanks for the insight!

From an architectural point of view I like `watch` better because it ensure we can never have this issue of subscribing after sending. Sadly it adds quite a bunch of boilerplate, so I'll drop this PR for now.

We can build a wrapper around it to remove the boilerplate, but we'd only end up using it in this single place, so that's probably not feasible.

PR #3270, opened 2023-12-22
Fix clippy 1.75

agourlay:
Rust 1.75 is landing on [December 28th](https://blog.rust-lang.org/inside-rust/2023/12/21/1.75.0-prerelease.html) so now is a good time to make sure our CI won't fail on that day.

This PR fixes new clippy lints issues and a bunch of minor reports from RustRover.

The main source of noise is the [get_first lint](https://rust-lang.github.io/rust-clippy/master/index.html#/get_first).

### DIFF
--- a/lib/segment/src/common/utils.rs
+++ b/lib/segment/src/common/utils.rs
@@ -7,6 +7,8 @@ use crate::data_types::vectors::Vector;
 use crate::index::field_index::FieldIndex;
 use crate::types::PayloadKeyType;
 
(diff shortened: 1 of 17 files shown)


lib/segment/src/common/utils.rs
```diff
@@ -779,5 +781,3 @@ mod tests {
         );
     }
 }
-
-pub type IndexesMap = HashMap<PayloadKeyType, Vec<FieldIndex>>;
```
agourlay, line 783: types cannot be placed under tests modules anymore

generall (approved):

PR #3296, opened 2023-12-28
Filed index point-to-values as separate struct

IvanPleshkov:
`ImmutableMapIndex` and `ImmutableNumericIndex` have a similar pattern for mapping from point id to values. This pattern is a flattened `Vec<Vec<_>>` structure.

Because geo index will use the same pattern too, I find it important to separate this pattern to avoid logic copying.

`ImmutablePointToValues` is a flattened representation of `Vec<Vec<_>>` with deletion.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/immutable_point_to_values.rs
+++ b/lib/segment/src/index/field_index/immutable_point_to_values.rs
@@ -1,0 +1,56 @@
+use std::ops::Range;
+
+use common::types::PointOffsetType;
+
+// Flatten points-to-values map
+#[derive(Debug, Clone, Default)]
+pub struct ImmutablePointToValues<N: Default> {
+    point_to_values: Vec<Range<u32>>,
+    point_to_values_container: Vec<N>,
+}
+
+impl<N: Default> ImmutablePointToValues<N> {
+    pub fn new(src: Vec<Vec<N>>) -> Self {
+        let mut point_to_values: Vec<_> = Default::default();
+        let mut point_to_values_container: Vec<_> = Default::default();
+        for values in src {
+            let values = values.into_iter().collect::<Vec<_>>();
+            let container_len = point_to_values_container.len() as u32;
+            let range = container_len..container_len + values.len() as u32;
+            point_to_values.push(range.clone());
+            point_to_values_container.extend(values);
+        }
+        Self {
+            point_to_values,
+            point_to_values_container,
+        }
+    }
+
+    pub fn get_values(&self, idx: PointOffsetType) -> Option<&[N]> {
+        let range = self.point_to_values.get(idx as usize)?.clone();
+        let range = range.start as usize..range.end as usize;
+        Some(&self.point_to_values_container[range])
+    }
+
+    pub fn remove_point(&mut self, idx: PointOffsetType) -> usize {
+        if self.point_to_values.len() <= idx as usize {
+            return 0;
+        }
+
+        // Point removing has to remove `idx` from both maps: points-to-values and values-to-points.
+        // The first one is easy: we just remove the entry from the map.
+        // The second one is more complicated: we have to remove all mentions of `idx` in values-to-points map.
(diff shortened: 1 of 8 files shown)


lib/segment/src/index/field_index/immutable_point_to_values.rs
```diff
@@ -0,0 +1,101 @@
...
+    // We can expect that we will never have more than 4 billion values per segment
+    point_to_values: Vec<Range<u32>>,
+    // flattened values
+    point_to_values_container: Vec<N>,
+}
+
+impl<N: Default> ImmutablePointToValues<N> {
+    pub fn new(src: Vec<Vec<N>>) -> Self {
+        let mut point_to_values: Vec<_> = Default::default();
+        let mut point_to_values_container: Vec<_> = Default::default();
```
timvisee, line 22:
```suggestion
        let mut point_to_values: Vec<_> = Vec::with_capacity(src.len());
        let mut point_to_values_container: Vec<_> = Vec::with_capacity(src.len());
```

The second capacity isn't accurate of course, bit it's cheap and would reduce reallocations by `log2(src.len())`.

We could count the total number of points or estimate based on an average, but that extra work probably isn't worth it.

lib/segment/src/index/field_index/immutable_point_to_values.rs
```diff
@@ -0,0 +1,101 @@
...
+        for values in src {
+            let values = values.into_iter().collect::<Vec<_>>();
+            let container_len = point_to_values_container.len() as u32;
+            let range = container_len..container_len + values.len() as u32;
+            point_to_values.push(range.clone());
+            point_to_values_container.extend(values);
```
timvisee, line 28:
```suggestion
            let container_len = point_to_values_container.len() as u32;
            let range = container_len..container_len + values.len() as u32;
            point_to_values.push(range.clone());
            point_to_values_container.extend(values);
```

Haven't tested this, but this would prevent an intermediate `Vec`. [`extend`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend) can take an `IntoIterator`.

lib/segment/src/index/field_index/immutable_point_to_values.rs
```diff
@@ -0,0 +1,101 @@
...
+        self.point_to_values[idx as usize] = Default::default();
+
+        let mut result = Vec::with_capacity(removed_values_range.len());
+        for value_index in removed_values_range {
+            // deleted values still use RAM, but it's not a problem because optimizers will actually reduce RAM usage
+            let value = std::mem::take(&mut self.point_to_values_container[value_index as usize]);
+            result.push(value);
+        }
+
+        result
```
timvisee, line 57: I guess leaving the original values is fine, since we simply don't use them anymore.

In that case it's likely more efficient to just do the following to return the values:

```rust
self.point_to_values_container[removed_values_range].to_vec()
```

(or `.into_iter().collect()` if it's an iterator)

lib/segment/src/index/field_index/immutable_point_to_values.rs
IvanPleshkov: Fixed

lib/segment/src/index/field_index/immutable_point_to_values.rs
IvanPleshkov: You are right! Thanks for this finding, this intermediate `Vec` is not needed. Fixed

lib/segment/src/index/field_index/immutable_point_to_values.rs
IvanPleshkov: In the case of the map index, this `N` type is a string and I want to clear the string here

timvisee: Smart!

timvisee (approved):
Perfect! :smile:

PR #3297, opened 2023-12-29
Immutable geo index

IvanPleshkov:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/geo_index/immutable_geo_index.rs
+++ b/lib/segment/src/index/field_index/geo_index/immutable_geo_index.rs
@@ -1,12 +1,16 @@
-use std::{collections::{BTreeMap, HashSet}, sync::Arc, cmp::max};
+use std::cmp::max;
+use std::collections::{BTreeMap, HashSet};
+use std::sync::Arc;
 
 use common::types::PointOffsetType;
 use parking_lot::RwLock;
 use rocksdb::DB;
 
-use crate::{index::field_index::geo_hash::{GeoHash, encode_max_precision}, types::GeoPoint, common::{rocksdb_wrapper::DatabaseColumnWrapper, operation_error::{OperationResult, OperationError}}};
-
 use super::GeoMapIndex;
+use crate::common::operation_error::{OperationError, OperationResult};
+use crate::common::rocksdb_wrapper::DatabaseColumnWrapper;
+use crate::index::field_index::geo_hash::{encode_max_precision, GeoHash};
+use crate::types::GeoPoint;
 
 pub struct ImmutableGeoMapIndex {
     /*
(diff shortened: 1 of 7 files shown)


lib/segment/src/index/field_index/geo_index/mutable_geo_index.rs
```diff
@@ -0,0 +1,310 @@
...
+        values: &[GeoPoint],
+    ) -> OperationResult<()> {
+        if values.is_empty() {
+            return Ok(());
+        }
+
+        if self.point_to_values.len() <= idx as usize {
+            // That's a smart reallocation
+            self.point_to_values.resize_with(idx as usize + 1, Vec::new);
+        }
```
timvisee, line 185: If we, as user, insert points one by one. Do we call this branch every time? In that case, this would actually be quite inefficient. It would then be better to grow by a factor of 2 to prevent constant expensive reallocations.

lib/segment/src/index/field_index/geo_index/immutable_geo_index.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    pub fn remove_point(&mut self, idx: PointOffsetType) -> OperationResult<()> {
+        let removed_geo_points = self.point_to_values.remove_point(idx);
+        if removed_geo_points.is_empty() {
+            return Ok(());
+        }
+
+        self.points_count -= 1;
+        self.points_values_count -= removed_geo_points.len();
```
timvisee, line 141: We probably want to subtract counts here before returning early.

lib/segment/src/index/field_index/geo_index/mod.rs
```diff
@@ -528,26 +230,47 @@ impl GeoMapIndex {
...
+        match self {
+            GeoMapIndex::Mutable(index) => Box::new(
+                values
+                    .into_iter()
+                    .flat_map(|top_geo_hash| {
+                        index
+                            .get_stored_sub_regions(&top_geo_hash)
+                            .flat_map(|(_geohash, points)| points.iter().copied())
+                    })
+                    .unique(),
```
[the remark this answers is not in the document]
IvanPleshkov: Agreed, I don't like `unique` here too. And I strongly believe that we can implement custom iterator that avoids `unique`. But I find it out of scope here. Created issue to keep it in mind: https://github.com/qdrant/qdrant/issues/3503

lib/segment/src/index/field_index/geo_index/immutable_geo_index.rs
IvanPleshkov: In geo index, points count is a count of points with non-empty values. Not sure that it's the correct behavior but I kept this logic as is for backward compatibility

lib/segment/src/index/field_index/geo_index/mod.rs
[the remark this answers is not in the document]
timvisee: Thank you for opening an issue for it. Resolving this now.

lib/segment/src/index/field_index/geo_index/mutable_geo_index.rs
timvisee: You're totally right! Thank you for confirming.

timvisee (approved):

PR #3314, opened 2024-01-04
Remove one usage of `async-trait`

timvisee:
Since Rust 1.75 we can remove this usage of [`async-trait`](https://docs.rs/async-trait/latest/async_trait/).

We still have some other usages left but we cannot remove those due to other constraints and due to code generation including them.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4841,7 +4841,6 @@ version = "0.2.0"
 dependencies = [
  "anyhow",
  "api",
- "async-trait",
  "atomicwrites",
  "cancel",
  "chrono",
--- a/lib/storage/Cargo.toml
+++ b/lib/storage/Cargo.toml
@@ -26,7 +26,6 @@ serde = { version = "~1.0", features = ["derive"] }
 serde_json = "~1.0"
 schemars = { version = "0.8.16", features = ["uuid1", "preserve_order", "chrono"] }
 itertools = "0.12"
-async-trait = "0.1.75"
 log = "0.4"
 tonic = { version = "0.9.2", features = ["gzip", "tls"] }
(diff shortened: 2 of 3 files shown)


ffuugoo:
I think we have to update Rust toolchain version in the Dockerfile, then.

timvisee:
> I think we have to update Rust toolchain version in the Dockerfile, then.

Already done in <https://github.com/qdrant/qdrant/pull/3306>. But thanks for the heads up! :smile:

ffuugoo (approved):
The day have come, folks! Rejoice! 🎉

agourlay:
The change looks good to, I am just a tiny concerned about our `Minimum Supported Rust Version`.

Because It is important to note that this makes our MSRV equal to 1.75.
What was it before?

Not sure if there are folks building from source stuck on older versions or if we want to actually accommodate anyone else.
I just want to make sure we make this decision explicit :memo:

timvisee:
> Because it is important to note that this makes our MSRV equal to 1.75.
> What was it before?

I believe our MSRV was 1.74 before this change :smile: 

> Not sure if there are folks building from source stuck on older versions or if we want to actually accommodate anyone else.
> I just want to make sure we make this decision explicit 📝

I don't think we should care too much, since no-one[^1] is using this as library. The general suggestion is to always update to the latest stable Rust. I'd vote for something different if we had a lot of users using our product as library.

In any case, it's definitely good to be explicit. Let's discuss this in our daily.

[^1]: not counting `qdrant-lib`

agourlay (approved):
Thanks for the detailed answer :+1:

Issue #3254, opened 2023-12-19
Flaky test_collection_creation_after_dropping test


generall:
https://github.com/qdrant/qdrant/actions/runs/7264432083/job/19791956029?pr=3251
https://github.com/qdrant/qdrant/actions/runs/7346890187/job/20002395601?pr=3296

timvisee:
Duplicate of <https://github.com/qdrant/qdrant/issues/2616>

Issue #2616, opened 2023-09-08
Flaky consensus test test_collection_creation_after_dropping


agourlay:
https://github.com/qdrant/qdrant/actions/runs/6109334007/job/16580136885

```
=========================== short test summary info ============================
FAILED tests/consensus_tests/test_collection_creation_after_dropping.py::test_collection_creation_after_dropping - Exception: HTTP request to http://127.0.0.1:48537/collections/test_collection/points?wait=false&ordering=weak failed with status code 404 after 0.00576s with response body:
{'status': {'error': "Not found: Collection `test_collection` doesn't exist!"}, 'time': 9.0301e-05}
============= 1 failed, 28 passed, 1 skipped in 215.55s (0:03:35) ==============
Error: Process completed with exit code 1.
```

timvisee:
- https://github.com/qdrant/qdrant/actions/runs/7264432083/job/19791956029?pr=3251
- https://github.com/qdrant/qdrant/actions/runs/7346890187/job/20002395601?pr=3296

timvisee:
This is now fixed, merged and released.

PR #3318, opened 2024-01-04
Synchronize nodes after collection creation

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/2616>, <https://github.com/qdrant/qdrant/issues/3254>.

After creating a collection in distributed mode, synchronize all nodes. This ensures the collection is ready on all nodes for follow-up operations such as upsertions.

If we don't do this, upserting into the collection on another node can fail as it may still be setting things up.

When synchronization fails a warning is reported and the collection creation call returns as if it was successful. This can happen if a node is unreachable during that time. This is better than failing, because this will eventually resolve itself and firing the same call another time will result in a 'already exists' error.

We don't have to do this for collection deletion, as it is unlikely that we expect follow-up operations on a collection right after deleting it as it should be gone. If we do, they'll fail eventually.

This change also fixes the above linked flaky test. I've extensively tested this locally with all scenarios I can think of to confirm this. Running the test 1000 times locally didn't trigger the error anymore.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/storage/src/dispatcher.rs
+++ b/lib/storage/src/dispatcher.rs
@@ -126,7 +126,7 @@ impl Dispatcher {
                         AliasOperations::DeleteAlias(_) => false,
                     })
                 }
-                // Do not sync nodes for other operations
+                // No need to sync nodes for other operations
                 CollectionMetaOperations::UpdateCollection(_)
                 | CollectionMetaOperations::DeleteCollection(_)
                 | CollectionMetaOperations::TransferShard(_, _)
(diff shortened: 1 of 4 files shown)


timvisee:
@generall This is ready now. 

It removes double-waiting for synchronizing nodes on shard key creation and now centrally waits when:
- creating a collection
- creating a shard key
- creating/updating collection aliases

agourlay (approved):

generall (approved):

PR #3329, opened 2024-01-05
Fix consensus operations not proposed if already listening to result

timvisee:
We have two important consensus methods:
1. `propose_consensus_op_with_await()`: propose new operation and listen for operation result
2. `await_for_multiple_operations()`: listen for operation result (without proposing)

With our current implementation, just listening <sup>(2)</sup> for an operation result prevents it from being proposed <sup>(1)</sup> after. That is because creating the notification channel for just listening makes the proposal method assume that the operation has already been proposed.

In code, if we are already listening for an operation result, we cannot propose the operation here because we subscribe on an _existing sender_: https://github.com/qdrant/qdrant/blob/c6a351c82d57eb786b75dac785de7b085c6f80ff/lib/storage/src/content_manager/consensus_manager.rs#L667-L681

This PR fixes this problem. We track the 'proposed' status in a boolean. If we didn't propose yet while trying to do so, we propose and flip the boolean.

I did not find immediate problems caused by this in our current code base, though I'm unsure about edge cases. It appears that we sometimes do listen without proposing first, but for these all seems to work fine according to our tests. I did hit this bug with some testing code though, causing waiting forever. I decided to make this patch in an attempt to prevent issues with it in the future, these are painful to debug.

An alternative is to add a strong warning `await_for_multiple_operations()` mentioning that awaiting can break further proposals.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/storage/src/content_manager/consensus_manager.rs
+++ b/lib/storage/src/content_manager/consensus_manager.rs
@@ -79,8 +79,7 @@ pub struct ConsensusManager<C: CollectionContainer> {
     /// Fires a signal if some specific operation is applied to the state machine.
     /// Signal is changed on change proposal and triggered if the change was applied by consensus on this peer.
     /// Also sends the result of the operation.
-    on_consensus_op_apply:
-        Mutex<HashMap<ConsensusOperations, broadcast::Sender<Result<bool, StorageError>>>>,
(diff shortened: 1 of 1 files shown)


generall:
Changes of this PR will be redundant as soon as we get rid of `Initializing` state. So I propose to just continue with deprecation

Issue #3336, opened 2024-01-05
Create AppImage for Qdrant


timvisee:
**Is your feature request related to a problem? Please describe.**
As part of Qdrant releases, we currently ship some binaries as [artifacts](https://github.com/qdrant/qdrant/releases/tag/v1.7.3). On Linux, these binaries are not compatible because each host has different dynamically linked library versions.

**Describe the solution you'd like**
We'd like to have an AppImage build of Qdrant as well to solve this issue. This should make it possible to run Qdrant on any system with a single binary.

Steps:
- Package Qdrant as AppImage
- Include web UI files (if possible) (see: <https://github.com/qdrant/qdrant/blob/c6a351c82d57eb786b75dac785de7b085c6f80ff/tools/sync-web-ui.sh>)
- Add release artifact to release builds

**Describe alternatives you've considered**
An alternative could be MUSL builds, we have a separate issue for this and likely want to have both:
- https://github.com/qdrant/qdrant/issues/3334

**Additional context**
Documentation on packaging: https://docs.appimage.org/packaging-guide/index.html

urbit-pilled:
This seems doable for me, I'll start working on this issue.
/attempt #3336

yangboz:
and  please add  CJK languages supported like chroma,  thanks.

timvisee:
> and please add CJK languages supported like chromadb, thanks.

At this time you can compile yourself with the `multiling-chinese`, `multiling-japanese` and or the `multiling-korean` flags to your desire.

yangboz:
> > and please add CJK languages supported like chromadb, thanks.
> 
> At this time you can compile yourself with the `multiling-chinese`, `multiling-japanese` and or the `multiling-korean` flags to your desire.

and how , any detailed tutorials that I can follow? thanks.

yangboz:
> > > and please add CJK languages supported like chromadb, thanks.
> > 
> > 
> > At this time you can compile yourself with the `multiling-chinese`, `multiling-japanese` and or the `multiling-korean` flags to your desire.
> 
> and how , any detailed tutorials that I can follow? thanks.

still wondering the dockerize way with our strong desire of CJK languages  support . thanks a lot!

PR #3342, opened 2024-01-06
feat: add musl release 

Dhanus3133:
Added MUSL build release targets for
- x86_64-unknown-linux-musl
- aarch64-unknown-linux-musl

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

You can check on the [Fork action](https://github.com/Dhanus3133/qdrant/actions/runs/7429125259) for successful builds

/claim #3334

PR #3343, opened 2024-01-06
build: AppImage release

Dhanus3133:
AppImage release build in CI

Fork Action - [Runner](https://github.com/Dhanus3133/qdrant/actions/runs/7430277956/job/20219775451)
Release - [Release](https://github.com/Dhanus3133/qdrant/releases/tag/appimage-final)

/claim #3336

### All Submissions:

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -43,6 +43,20 @@ jobs:
           bin: qdrant
           target: ${{ matrix.target }}
           token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build Debian Package
+        if: matrix.target == 'x86_64-unknown-linux-musl'
+        run: |
+          cargo install cargo-deb
+          cargo deb --no-strip --target ${{ matrix.target }}
+      - name: Upload Debian package
+        if: matrix.target == 'x86_64-unknown-linux-musl'
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: target/${{ matrix.target }}/debian/*.deb
+          tag: ${{ github.ref }}
+          overwrite: true
+          file_glob: true
 
   build-mac-binaries:
     strategy:
@@ -151,6 +165,28 @@ jobs:
               --executable qdrant \
               -d qdrant.desktop \
               -i qdrant.svg \
+
+          rm AppDir/AppRun
+
+          cat << EOF > "AppDir/AppRun"
+          #!/bin/bash
+          APPDIR="\$(dirname "\$(readlink -f "\$0")")"
+          if [ ! -d "static" ]; then
+              STATIC_DIR="\$APPDIR/usr/share/static"
+              mkdir -p static
+              cp -r "\$STATIC_DIR/"* "static/"
+          fi
+          exec "\$APPDIR/usr/bin/qdrant" "\$@"
+          EOF
+
+          wget -O dist-qdrant.zip "$(curl --silent "https://api.github.com/repos/qdrant/qdrant-web-ui/releases/latest" | jq -r '.assets[] | select(.name=="dist-qdrant.zip") | .browser_download_url')"          
+          unzip -o dist-qdrant.zip -d static
+          mv -n static/dist/* static/
+          rm -rf static/dist
+          mv static AppDir/usr/share
+
+          ./linuxdeploy-x86_64.AppImage \
+              --appdir AppDir \
               --output appimage
 
       - name: Upload AppImage
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -17,12 +17,12 @@ jobs:
     steps:
       - name: Install dependencies
         run: |
-            sudo apt-get update
-            sudo apt-get install -y gcc-multilib clang cmake protobuf-compiler
-            rustup default stable
-            rustup update
-            rustup show
-            cargo -Vv
+          sudo apt-get update
+          sudo apt-get install -y gcc-multilib clang cmake protobuf-compiler
+          rustup default stable
+          rustup update
+          rustup show
+          cargo -Vv
       - uses: actions/checkout@v4
       - name: Build and publish
         uses: taiki-e/upload-rust-binary-action@v1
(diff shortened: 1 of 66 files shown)


.github/workflows/release-artifacts.yml
```diff
@@ -81,3 +81,31 @@ jobs:
...
+          sudo apt-get install -y protobuf-compiler libfuse2
+          sudo wget https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-$(uname -m).AppImage -O /usr/local/bin/appimagetool
+          sudo chmod +x /usr/local/bin/appimagetool
+          cargo install cargo-appimage
+          cargo appimage
+      - name: Upload AppImage
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: "target/appimage/qdrant.AppImage"
```
generall, line 108: Is it corrent, that this AppImage will only work on x86_64 machines? If so, let's call it  `qdrant-x86_64.AppImage`

generall (changes requested):
When I try to run `qdrant.AppImage` on my machine, I get the following error:


```
Error: Service internal error: Read-only file system (os error 30)
```

image

It looks like it tries to write changes inside internal "filesystem".
I would expect it should rather try to act like binary and write into the current directory

.github/workflows/release-artifacts.yml
Dhanus3133: Yes, I guess. Let me just change it.

Dhanus3133:
@generall Could you check this again! [Release](https://github.com/Dhanus3133/qdrant/releases/tag/new-appimage3) | [Action](https://github.com/Dhanus3133/qdrant/actions/runs/7438904439)

image

generall (approved):
It works on my machine now. @timvisee could you please check if we need anything else here or it is fine

urbit-pilled:
Good job on getting the appimage to work @Dhanus3133,  but I think will this be similar to the previously compiled binaries and doesn't fix the dynamically linked library issue.

You could fix this by wrapping the musl binary you added in #3342 in an appimage, or finding a different way to create a statically linked binary.

generall:
@urbit-pilled it looks you are right.
I tried on another machine and it is what I got:
image

Also noticed that web-ui is not included into AppImage

generall (changes requested):
Refer to the last comment

Dhanus3133:
@generall Kindly try the new build [Release](https://github.com/Dhanus3133/qdrant/releases/tag/final-image2) | [Action](https://github.com/Dhanus3133/qdrant/actions/runs/7517543811/). Also, web UI is included.

generall:
The current AppImage is still somehow less portable than `musl`.

That is what happens when I try to run inside `ubuntu:latest` container:

image

Dhanus3133:
> The current AppImage is still somehow less portable than `musl`.
> 
> That is what happens when I try to run inside `ubuntu:latest` container:
> 
> image

You can use this Dockerfile, worked well for me.
```Dockerfile
FROM ubuntu:latest
RUN apt-get update && apt-get install -y wget
RUN wget https://github.com/Dhanus3133/qdrant/releases/download/final-image2/qdrant-x86_64.AppImage
RUN chmod +x qdrant-x86_64.AppImage
CMD APPIMAGE_EXTRACT_AND_RUN=1 ./qdrant-x86_64.AppImage
```

Regarding the usage of MUSL, I tried the [MUSL Build Release](https://github.com/Dhanus3133/qdrant/releases/tag/musl-image) cases errors when executed but the build was successful. Also noticed still app images are not compatible with statically linked runtimes.

generall:
> APPIMAGE_EXTRACT_AND_RUN=1

Can this be avoided?

Dhanus3133:
> > APPIMAGE_EXTRACT_AND_RUN=1
> 
> Can this be avoided?

Kindly refer this https://github.com/AppImage/AppImageKit/wiki/FUSE#docker

timvisee (changes requested):
When running the AppImage, it creates a `static` folder on your host to serve the web UI. I think this should be avoided, and the `static` folder should remain within the AppImage.

```bash
$ chmod a+x qdrant-x86_64.AppImage 
$ ./qdrant-x86_64.AppImage
$ ls
snapshot/
static/
storage/
```

Dhanus3133:
> When running the AppImage, it creates a `static` folder on your host to serve the web UI. I think this should be avoided, and the `static` folder should remain within the AppImage.
> 
> ```shell
> $ chmod a+x qdrant-x86_64.AppImage 
> $ ./qdrant-x86_64.AppImage
> $ ls
> snapshot/
> static/
> storage/
> ```

Try this [Release](https://github.com/Dhanus3133/qdrant/releases/tag/staticdir2) made some changes.

src/actix/mod.rs
```diff
@@ -64,11 +64,13 @@ pub fn init(
         let http_client = web::Data::new(HttpClient::from_settings(&settings)?);
         let health_checker = web::Data::new(health_checker);
         let auth_keys = AuthKeys::try_create(&settings.service);
-        let static_folder = settings
-            .service
-            .static_content_dir
-            .clone()
-            .unwrap_or(DEFAULT_STATIC_DIR.to_string());
+        let static_folder = env::var("STATIC_DIR").unwrap_or_else(|_| {
```
generall, line 67: If you want to overwrite value from settings in Qdrant, you can use the following pattern:

`QDRANT__SERVICE__STATIC_CONTENT_DIR=...` instead of introducing a new env variable that doesn't follow the pattern. More info here: https://qdrant.tech/documentation/guides/configuration/#environment-variables

timvisee: The only objection I have against this, is that this prevents the user from customizing the directory. But I guess that's an implementation detail we can resolve later.

timvisee: Sweet! Thank you for picking this up so quickly.

timvisee (approved):

.github/workflows/release-artifacts.yml
```diff
@@ -108,3 +112,88 @@ jobs:
...
+          Comment=Advanced and High-Performant Vector Search Technology
+          TryExec=qdrant
+          Exec=qdrant %F
+          Terminal=true
+          Type=Application
+          Keywords=Database;
+          Icon=qdrant
+          Categories=Utility;
+          StartupNotify=false
+          EOF
```
timvisee, line 161: On second thought, can we remove all EOF entries here into separate files to clean up this build script?

Maybe we can move this into `pkg/appimage/qdrant.desktop` for example.

Dhanus3133: Sure!

Dhanus3133: Cleaned up the code, have a look once again!

timvisee (approved):
Fantastic work. Thank you!

generall (approved):

PR #3346, opened 2024-01-08
feat: Expose git commit id in the health check endpoint

KShivendu:
### All Submissions:

This will help with identifying commit ID in our CI Grafana dashboard.

```
➜  ~ curl localhost:6333
{"title":"qdrant - vector search engine","version":"0.11.1","commit":"64ec1149956f298c54b9f11e765949a0e141565e"}%                                  

➜  ~ grpcurl -plaintext localhost:6334 qdrant.Qdrant/HealthCheck
{
  "title": "qdrant - vector search engine",
  "version": "0.11.1",
  "commit": "64ec1149956f298c54b9f11e765949a0e141565e"
}
```

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/api/src/grpc/models.rs
+++ b/lib/api/src/grpc/models.rs
@@ -4,7 +4,7 @@ use schemars::JsonSchema;
 use serde;
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, JsonSchema)]
 pub struct VersionInfo {
     pub title: String,
     pub version: String,
--- a/lib/api/src/grpc/models.rs
+++ b/lib/api/src/grpc/models.rs
@@ -8,6 +8,7 @@ use serde::{Deserialize, Serialize};
 pub struct VersionInfo {
     pub title: String,
     pub version: String,
+    pub commit_id: String,
 }
 
 impl Default for VersionInfo {
@@ -15,6 +16,7 @@ impl Default for VersionInfo {
         VersionInfo {
             title: "qdrant - vector search engine".to_string(),
             version: env!("CARGO_PKG_VERSION").to_string(),
+            commit_id: env!("GIT_COMMIT_ID").to_string(),
         }
     }
 }
--- a/lib/api/src/grpc/models.rs
+++ b/lib/api/src/grpc/models.rs
@@ -8,7 +8,8 @@ use serde::{Deserialize, Serialize};
 pub struct VersionInfo {
     pub title: String,
     pub version: String,
-    pub commit_id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub commit_id: Option<String>,
 }
 
 impl Default for VersionInfo {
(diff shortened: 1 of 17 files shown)


KShivendu:
@generall please review

lib/api/src/grpc/models.rs
```diff
@@ -4,17 +4,20 @@ use schemars::JsonSchema;
 use serde;
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, JsonSchema)]
 pub struct VersionInfo {
     pub title: String,
     pub version: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub commit_id: Option<String>,
```
timvisee, line 12: Nitpick: I believe this is commonly called `commit`, would that be better?

timvisee (approved):
Awesome work! :rocket: 

As a minor addition to this I quickly made it optional. This prevents `git` and a repository from being a requirement to build Qdrant. That may be useful for packaging in some cases, for example.

lib/api/src/grpc/models.rs
KShivendu: I was thinking the same. Will do. Thanks!

timvisee:
It appears that my `optional` changes broke compilation somehow, but I'm a bit clueless on why this suddenly popped up. :sweat_smile:

KShivendu:
So as per [probuf.dev](https://protobuf.dev/programming-guides/field_presence/#how-to-enable-explicit-presence-in-proto3), proto3 doesn't allow that by default. 

I tried this and it fixed the error:
```rust
tonic_build::configure()
        .configure_validation()
        .file_descriptor_set_path(build_out_dir.join("qdrant_descriptor.bin"))
        .protoc_arg("--experimental_allow_proto3_optional") // --> this
        .out_dir("src/grpc/")
        .compile("...")
....
```

However, I'm sceptical of doing this because it's "experimental" and it would probably force us to use the same flag in all of our SDKs. I think it's better to just use an empty string if git isn't present while building. What do you think? 

cc @generall

KShivendu:
Ohh looks like the newer versions (v3.15.0) of `protoc` [supports this](https://protobuf.dev/programming-guides/field_presence/#protoc-invocation) without the experimental flag. 

I tried [v3.15.0](https://github.com/protocolbuffers/protobuf/releases/download/v3.15.0/protoc-3.15.0-linux-x86_64.zip) locally and it worked without the flag. 

This means the [setup-protoc](https://github.com/arduino/setup-protoc) GH action that we use doesn't pick up the latest versions of protoc by default.

timvisee:
The weird thing is that we use `optional` in a lot of other `.proto` files which does work fine, so this isn't anything new.

openapi/tests/openapi_integration/test_ping_api.py
```diff
@@ -1,14 +1,14 @@
 from .helpers.helpers import request_with_validation
 
-def test_health_api():
+def test_ping_api():
```
timvisee, line 3: Nitpick: maybe just `test_root` or `test_root_api`.

KShivendu: Good one. I was confused what to name it. I'll update the OpenAPI docs as well.

PR #3347, opened 2024-01-08
Optimize inverted index search by tracking min record ids

agourlay:
This PR optimizes the sparse vector search.

One short-coming of the current implementation is that it recomputes continuously the min record ids across posting lists to know how to advance the cursors.

This PR proposes to track the min records ids across all posting lists in a min-heap to avoid performing extra work.

It yields a 20% performance improvement.

```
sparse-vector-search-group/inverted-index-search
                        time:   [23.966 ms 24.069 ms 24.180 ms]
                        change: [-21.418% -20.990% -20.558%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/lib/sparse/src/index/search_context.rs
+++ b/lib/sparse/src/index/search_context.rs
@@ -23,8 +23,8 @@ pub struct SearchContext<'a> {
     top: usize,
     is_stopped: &'a AtomicBool,
     result_queue: FixedLengthPriorityQueue<ScoredPointOffset>, // keep the largest elements and peek smallest
+    min_record_ids: FixedLengthPriorityQueue<PointOffsetType>, // min record ids across all posting lists
     use_pruning: bool,
-    min_record_id: Option<u32>, // min record id across all posting lists
 }
 
 impl<'a> SearchContext<'a> {
@@ -52,15 +52,21 @@ impl<'a> SearchContext<'a> {
         // The max contribution per posting list that we calculate is not made to compute the max value of two negative numbers.
         // This is a limitation of the current pruning implementation.
         let use_pruning = query.values.iter().all(|v| *v >= 0.0);
-        let min_record_id = Self::next_min_id(&postings_iterators);
+        // find min record id across all posting lists
+        let mut min_record_ids = FixedLengthPriorityQueue::new(query.indices.len());
+        for posting_iterator in postings_iterators.iter() {
+            if let Some(element) = posting_iterator.posting_list_iterator.peek() {
+                min_record_ids.push(element.record_id);
+            }
+        }
         SearchContext {
             postings_iterators,
             query,
             top,
             is_stopped,
             result_queue,
+            min_record_ids,
             use_pruning,
-            min_record_id,
         }
     }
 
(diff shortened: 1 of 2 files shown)


timvisee (approved):

lib/sparse/src/index/search_context.rs
```diff
@@ -130,24 +139,31 @@ impl<'a> SearchContext<'a> {
     /// b,  21, 34, 60, 200
     /// b,  30, 34, 60, 230
     fn advance(&mut self) -> Option<ScoredPointOffset> {
-        let min_record_id = Self::next_min_id(&self.postings_iterators)?;
```
IvanPleshkov, line 133: Because there is only one usage of `next_min_id` left, WDYT about `next_min_id` signature refactoring?

lib/sparse/src/index/search_context.rs
```diff
@@ -51,12 +52,20 @@ impl<'a> SearchContext<'a> {
         // The max contribution per posting list that we calculate is not made to compute the max value of two negative numbers.
         // This is a limitation of the current pruning implementation.
         let use_pruning = query.values.iter().all(|v| *v >= 0.0);
+        // find min record id across all posting lists
+        let mut min_record_ids = FixedLengthPriorityQueue::new(query.indices.len());
```
IvanPleshkov, line 56: I'm not sure right now about safe capacity value. It seems we can use just `usize::MAX`, isn't it?

lib/sparse/src/index/search_context.rs
```diff
@@ -130,24 +139,31 @@ impl<'a> SearchContext<'a> {
...
-                if element.record_id == min_record_id {
+                if element.record_id == current_min_record_id {
                     score += element.weight * posting_iterator.query_weight;
                     // advance posting list iterator to next element
                     posting_iterator.posting_list_iterator.advance();
+                    // pop min record id
+                    self.min_record_ids.pop();
+                    // look-ahead to find next min record id
+                    if let Some(next_element) = posting_iterator.posting_list_iterator.peek() {
+                        self.min_record_ids.push(next_element.record_id);
```
IvanPleshkov, line 158: I cannot understand why we add record id here, we have already filled `self.min_record_ids` in `SearchContext::new` with all query indices

lib/sparse/src/index/search_context.rs
agourlay: What do you in mind exactly?
FYI the next PR is removing this feature completely :broom:

lib/sparse/src/index/search_context.rs
agourlay: We can't put a very large number as capacity because the `BinaryHeap` actually pre-allocates it.

Also, by construction there will be at most `query.indices.len()` element due a careful dance of peek/pop/push.

Do you have a specific concern in mind?

lib/sparse/src/index/search_context.rs
agourlay: `min_record_ids` tracks the min record point ids of the posting list, not the query indices.

As we `pop` an element for a the posting list, we need to the new `next` min record id.
Each non empty posting list must have an entry in the heap.

Does it help?

lib/sparse/src/index/search_context.rs
IvanPleshkov: Great! 🧹🧹🧹

lib/sparse/src/index/search_context.rs
IvanPleshkov: I forgot about preallocation, nevermind

lib/sparse/src/index/search_context.rs
IvanPleshkov: Yep

IvanPleshkov (approved):

PR #3359, opened 2024-01-08
Faster inverted index with minheap traversal

agourlay:
This PR builds on top of https://github.com/qdrant/qdrant/pull/3347 to go even faster.

The key insight is that it is unnecessary to constantly probe all posting list for finding the ones that should be explored.

By leveraging the previously introduced min-heap to drive the search, we can access directly the right positing lists.

### Performance

This brings a a large performance improvement to the criterion benchmarks.
Those are using 50k indexed sparse vectors with around 300 dimensions with a max dimension of 30k.

```
sparse-vector-search-group/mmap-inverted-index-search
                        time:   [8.7720 ms 8.7792 ms 8.7891 ms]
                        change: [-72.942% -72.337% -71.780%] (p = 0.00 < 0.05)
                        Performance has improved.

sparse-vector-search-group/inverted-index-search
                        time:   [10.112 ms 10.142 ms 10.174 ms]
                        change: [-66.641% -66.489% -66.342%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### Notes

1. Given that the posting list cannot be re-ordered anymore, the logic to decide which posting to prune has been changed to target any postings which has the lowest unique min record id. But the pruning logic as not changed.

2. The score seems to not be completely equivalent due to the change of order for float multiplication.

3. I have verified that the python client congruence tests are still green with this change.

### DIFF
--- a/lib/segment/tests/integration/sparse_vector_index_search_tests.rs
+++ b/lib/segment/tests/integration/sparse_vector_index_search_tests.rs
@@ -111,7 +111,16 @@ fn compare_sparse_vectors_search_with_without_filter(full_scan_threshold: usize)
                 .filter(|s| s.score != 0.0)
                 .zip(no_filter_result.iter().filter(|s| s.score != 0.0))
             {
-                assert_eq!(filter_result, no_filter_result);
+                assert_eq!(filter_result.idx, no_filter_result.idx);
+                // the scores between search and plain search can be slightly different due to different order in which points are processed
+                let score_error = (filter_result.score - no_filter_result.score).abs();
+                assert!(
+                    score_error < 0.01,
+                    "query = {:#?}, filter_result = {:#?} no_filter_result = {:#?}",
+                    query,
+                    filter_result,
+                    no_filter_result,
+                );
             }
         }
     }
--- a/lib/sparse/src/index/posting_list.rs
+++ b/lib/sparse/src/index/posting_list.rs
@@ -196,6 +196,11 @@ impl<'a> PostingListIterator<'a> {
         self.elements.get(self.current_index)
     }
 
+    /// Returns the next element *without* bounds checking.
+    pub fn get_unchecked(&self) -> &PostingElement {
+        unsafe { self.elements.get_unchecked(self.current_index) }
+    }
+
     /// Returns the number of elements from the current position to the end of the list.
     pub fn len_to_end(&self) -> usize {
         self.elements.len() - self.current_index
--- a/lib/sparse/src/index/search_context.rs
+++ b/lib/sparse/src/index/search_context.rs
@@ -17,13 +17,44 @@ pub struct IndexedPostingListIterator<'a> {
     query_weight: DimWeight,
 }
 
+/// Tracks the min record of posting list based on their position.
+///
+/// Sorted by min record id to allow for efficient min record id lookup.
+#[derive(Debug, PartialEq)]
+struct PostingTracker {
+    min_record_id: PointOffsetType,
+    index: usize, // the index of the posting list in `SearchContext::postings_iterators`
+}
+
+impl PostingTracker {
+    fn new(min_record_id: PointOffsetType, index: usize) -> PostingTracker {
+        PostingTracker {
+            min_record_id,
+            index,
+        }
+    }
+}
+impl Eq for PostingTracker {}
+
+impl Ord for PostingTracker {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.min_record_id.cmp(&other.min_record_id)
+    }
+}
+
+impl PartialOrd for PostingTracker {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
 pub struct SearchContext<'a> {
     postings_iterators: Vec<IndexedPostingListIterator<'a>>,
     query: SparseVector,
     top: usize,
     is_stopped: &'a AtomicBool,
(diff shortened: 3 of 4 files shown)


agourlay:
A consensus [test](https://github.com/qdrant/qdrant/actions/runs/7458986467/job/20296989912?pr=3359) is failing for sparse vector search. 

```
  -  [{'id': 4, 'payload': None, 'score': 1.3620001, 'vector': None, 'version': 0},
  ?                                            ----
  +  [{'id': 4, 'payload': None, 'score': 1.362, 'vector': None, 'version': 0},
  ```

My guess is that the minheap does not break ties deterministically which creates slight difference in float precision due to a different multiplication ordering.

**Comment by timvisee** [comment]
> A consensus [test](https://github.com/qdrant/qdrant/actions/runs/7458986467/job/20296989912?pr=3359) is failing for sparse vector search.
> 
> ```
>   -  [{'id': 4, 'payload': None, 'score': 1.3620001, 'vector': None, 'version': 0},
>   ?                                            ----
>   +  [{'id': 4, 'payload': None, 'score': 1.362, 'vector': None, 'version': 0},
> ```
> 
> My guess is that the minheap does not break ties deterministically which creates slight difference in float precision due to a different multiplication ordering.

Potentially stupid question, but are you sure you're not mutating any other items in the min heap? Asking because that's unsupported.

**Comment by agourlay** [comment]
@timvisee I fixed the test with https://github.com/qdrant/qdrant/pull/3359/commits/1f6714027faa6c666141518dbb6229cba4262109

I assume the internal ids can be different in replicas leading to a different order of float computation.

File: lib/sparse/src/index/search_context.rs [hunk 1]
```diff
@@ -139,25 +171,30 @@ impl<'a> SearchContext<'a> {
...
     fn advance(&mut self) -> Option<ScoredPointOffset> {
-        // Get current min record id from all posting list iterators
-        let current_min_record_id = *self.min_record_ids.peek()?;
+        // get current min tracker from all posting list iterators
+        let min_tracker = self.min_record_ids.peek()?;
+        // current record_id to score
+        let current_min_record_id = min_tracker.min_record_id;
         let mut score = 0.0;
-
-        // Iterate to advance matching posting iterators
```
agourlay, line 146: This is the important change explaining the gain.
Previously we were iterating over all posting iterator heads to find the min record id.
Gets slower as the query grows.

lib/sparse/src/index/posting_list.rs
```diff
@@ -196,6 +196,11 @@ impl<'a> PostingListIterator<'a> {
         self.elements.get(self.current_index)
     }
 
+    /// Returns the next element *without* bounds checking.
+    pub fn get_unchecked(&self) -> &PostingElement {
+        unsafe { self.elements.get_unchecked(self.current_index) }
+    }
```
timvisee, line 202: Do we guarantee this is always safe?

I think we should either:
- if always safe, rename to `get()`
- if not always safe, color this function to be `unsafe` as well

Also, does this give us any measurable difference?

lib/segment/tests/integration/sparse_vector_index_search_tests.rs
```diff
@@ -111,7 +111,16 @@ fn compare_sparse_vectors_search_with_without_filter(full_scan_threshold: usize)
                 .filter(|s| s.score != 0.0)
                 .zip(no_filter_result.iter().filter(|s| s.score != 0.0))
             {
-                assert_eq!(filter_result, no_filter_result);
+                assert_eq!(filter_result.idx, no_filter_result.idx);
+                // the scores between search and plain search can be slightly different due to different order in which points are processed
+                let score_error = (filter_result.score - no_filter_result.score).abs();
+                assert!(
+                    score_error < 0.01,
```
timvisee, line 118: Can we use [`EPSILON`](https://doc.rust-lang.org/std/primitive.f32.html#associatedconstant.EPSILON) here, or is it too small?

lib/sparse/src/index/search_context.rs
```diff
@@ -17,13 +17,44 @@ pub struct IndexedPostingListIterator<'a> {
...
+
+impl PostingTracker {
+    fn new(min_record_id: PointOffsetType, index: usize) -> PostingTracker {
+        PostingTracker {
+            min_record_id,
+            index,
+        }
+    }
+}
+impl Eq for PostingTracker {}
```
timvisee, line 37: Can we `#[derive(Eq)]` instead?

lib/sparse/src/index/search_context.rs
```diff
@@ -267,60 +256,71 @@ impl<'a> SearchContext<'a> {
...
+    ///
+    /// Returns true if the a posting list was pruned.
+    pub fn prune_lowest_posting_list(&mut self, min_score: f32) -> bool {
+        let trackers_len = self.min_record_ids.len();
+        // no trackers left
+        if trackers_len == 0 {
+            return false;
+        }
+        // one tracker left, let's try to prune fully the corresponding posting
+        if trackers_len == 1 {
```
timvisee, line 270: Though perfectly fine, maybe we can do this  to prevent if-elsing:

```rust
match self.min_record_ids.len() {
    0 => return false,
    1 => {
        // ...
    }
    2.. => {
        // ...
    }
}
```

agourlay:
Converted back into a draft as I need to re-evaluate this approach using a realistic dataset.
The data distribution has a strong impact on the performance.

agourlay:
closing as the approach does not seem to shine with highly skewed data

PR #3364, opened 2024-01-09
Dynamic CPU saturation internals

timvisee:
Internals for <https://github.com/qdrant/qdrant/pull/3115>.

This is an implementation of the above PR, but with all public API changes removed. That way we're left with all internals we can merge it in the next patch release. We'll bring back the public API changes in a different PR when doing the next minor release, to make this feature fully functional.

For implementation details and further discussion, please resort to <https://github.com/qdrant/qdrant/pull/3115>.
With these changes the optimizer and CPU usage semantics should remain the same.

The following command shows that there are no changes in the public API, except for some improvement comments in `config.yaml`:

```bash
git fetch
git diff origin/dev origin/dynamic-optimizer-cpus-internal -- docs/redoc/master/openapi.json docs/grpc/docs.md config/config.yaml lib/api/src/grpc/qdrant.rs
```

To eventually introduce the public API changes, we can revert 9f91aee4ada86bd916a1dc5aa7149b1e1248eb9c.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/collection/src/collection/mod.rs
```diff
@@ -112,6 +115,7 @@ impl Collection {
                 channel_service.clone(),
                 update_runtime.clone().unwrap_or_else(Handle::current),
                 search_runtime.clone().unwrap_or_else(Handle::current),
+                optimizer_cpu_budget.clone(),
```
coszio, line 118: Does this mean that all nodes will have same cpu budget for optimizations? I am thinking maybe we should let each node have its own config

timvisee: I think you meant 'that all collections will have the same CPU budget for optimizations'. 

CPU optimizations are on node/instance level. So a node has a global CPU budget internally, which is shared across all collections.

We must share this budget this to prevent two collections consuming more CPU together than is available on the host.

_Once the public API and configuration changes are exposed as well:_ each node has the global budget configured in their own `config.yaml`. Concurrency across collections _can_ be controlled with the optimizer parameters at collection creation, such as `optimizers_config.max_optimization_threads` ([API](https://qdrant.github.io/qdrant/redoc/index.html#tag/collections/operation/get_collection)).

Do you think this is sufficient?

coszio: Ah, right. I got it now. It's same budget but per instance, not per collection across instances. It makes much more sense this way

timvisee: Yes, that's another good way of saying it! :smile:

lib/collection/src/update_handler.rs
```diff
@@ -401,12 +458,33 @@ impl UpdateHandler {
...
+                    log::trace!(
+                        "Blocking optimization check, waiting for CPU budget to be available"
+                    );
+                    let desired_cpus = max_rayon_threads(max_indexing_threads);
+                    optimizer_cpu_budget.block_until_budget(desired_cpus);
+                    log::trace!("Continue with optimizations, new CPU budget available");
+
+                    // Determine optimization handle limit based on max handles we allow
+                    // Skip if we reached limit, an ongoing optimization that finishes will trigger this loop again
+                    let limit = max_handles.saturating_sub(optimization_handles.lock().await.len());
```
coszio, line 472: Can you briefly explain how this `limit` interacts with the cpu budget?

coszio (commented):
Few more comments. 
Pushed a small commit to fix a nit that wasn't commentable in the PR :)

config/config.yaml
```diff
@@ -95,27 +95,34 @@ storage:
...
+
     # Minimal size (in KiloBytes) of vectors for additional payload-based indexing.
     # If payload chunk is smaller than `full_scan_threshold_kb` additional indexing won't be used -
     # in this case full-scan search should be preferred by query planner and additional indexing is not required.
     # Note: 1Kb = 1 vector of size 256
     full_scan_threshold_kb: 10000
-    # Number of parallel threads used for background index building. If 0 - auto selection.
+
+    # Number of parallel threads used for background index building.
+    # If 0 - automatically select from 8 to 16.
```
generall, line 118: From 8 to 16 applies only to high CPU machines. I propose to hide implementation details from here

lib/collection/src/shards/transfer/mod.rs
```diff
@@ -79,6 +80,9 @@ pub enum ShardTransferMethod {
 /// Interface to consensus for shard transfer operations.
 #[async_trait]
 pub trait ShardTransferConsensus: Send + Sync {
+    /// Get current CPU budget.
+    fn cpu_budget(&self) -> CollectionResult<CpuBudget>;
```
generall, line 84: how is `cpu_budget` a part of `ShardTransferConsensus`?

src/settings.rs
```diff
@@ -281,14 +280,13 @@ impl Settings {
...
     let max_workers = settings.service.max_workers;
-
-    if max_workers == Some(0) {
-        let num_cpu = get_num_cpus();
-        std::cmp::max(1, num_cpu - 1)
-    } else if max_workers.is_none() {
-        settings.storage.performance.max_search_threads
-    } else {
-        max_workers.unwrap()
+    match max_workers {
```
xzfc, line 283: Nit style: you could as well inline `max_workers` variable.

lib/collection/src/update_handler.rs
timvisee: It is unrelated to CPU budget, it is a concrete limit of optimizations per shard that are allowed at a given time. It can be configured by the user. We have to keep this for backwards compatibility, otherwise I'd probably remove this extra limit. It no defaults to no limit at all (to then limit optimizations based on CPU budget).

I've added this comment:

```rust
// Not related to the CPU budget, but a different limit for the maximum number
// of concurrent concrete optimizations per shard as configured by the user in
// the Qdrant configuration.
```

This limit can be set with: `storage.optimizers.max_optimization_threads`

lib/segment/src/index/hnsw_index/mod.rs
```diff
@@ -12,15 +14,10 @@ mod search_context;
 #[cfg(test)]
 mod tests;
 
-// In case if we are dealing with high-CPU system, creating more than
-// this amount of threads will most likely not improve performance
-// But we still allow to override this value by setting `max_indexing_threads` to non-zero value
-const MAX_AUTO_RAYON_THREADS: usize = 16;
-
 pub fn max_rayon_threads(max_indexing_threads: usize) -> usize {
```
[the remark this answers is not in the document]
timvisee: I've renamed it to `num_rayon_threads` and added this documentation:

```rust
/// Number of threads to use with rayon for HNSW index building.
///
/// Uses [`thread_count_for_hnsw`] heuristic but accepts a `max_indexing_threads` parameter to
/// allow configuring this.
```

It is in the `hnsw` module, so I guess that by itself adds the extra needed context.

config/config.yaml
timvisee: Good idea, removed the numbers.

```diff
-# If 0 - automatically select from 8 to 16.
+# If 0 - automatically select.
```

src/settings.rs
timvisee: Done. Thanks!

lib/collection/src/shards/transfer/mod.rs
timvisee: Where we access it [here](https://github.com/qdrant/qdrant/blob/4122b7e49cae7916db27f47b47faa9616c4ff213/lib/collection/src/collection/shard_transfer.rs#L95), we can simply do `self.optimizer_cpu_budget`. I've updated it and removed CPU budget from the trait. Thanks for this insight @generall!

lib/collection/src/update_handler.rs
coszio: Do we plan to deprecate this then, or it will still play along indefinitely?

coszio (approved):
Looks good to me now, though it would be great to be able to get some performance numbers with and without these changes once the api changes are introduced

lib/collection/src/update_handler.rs
timvisee: > Do we plan to deprecate this then

Didn't think about a schedule for deprecation yet. I guess it doesn't hurt to keep for a while.

timvisee:
> though it would be great to be able to get some performance numbers with and without these changes once the api changes are introduced

Good idea. I did see a significant improvement on high CPU systems with some segment/shard configurations, but I never expressed this in numbers. On low CPU systems (<16) it should behave the same.

agourlay:
> Good idea. I did see a significant improvement on high CPU systems with some segment/shard configurations, but I never expressed this in numbers (yet). On low CPU systems (<16) it should behave the same.

The code _seems_ fine but they are a lot of moving pieces.
I think we need a clear performance validation for landing such a large PR.
Would be great to publish some of those numbers to know that we are moving in the right direction :pray:

timvisee:
> > Good idea. I did see a significant improvement on high CPU systems with some segment/shard configurations, but I never expressed this in numbers (yet). On low CPU systems (<16) it should behave the same.
> 
> The code _seems_ fine but they are a lot of moving pieces. I think we need a clear performance validation for landing such a large PR. Would be great to publish some of those numbers to know that we are moving in the right direction 🙏

Good point. I'll have to do that later though.

I did test on 24 cores extensively. I've also tested on 1..24 cores quite a few times to validate all changes, ensuring low-core machines didn't break.

I've also tested on 256 cores, which had a huge effect, indexing about 15 times faster. @KShivendu can probably confirm this. Sadly I don't think we have this machine available at this time.

Sorry for this whole message of nothing, but during all that testing I didn't write things down. I'll try to document this properly later.

generall:
Running from the commit `142599f0`, looks like it causes dedlock on simple operations like get collection info.


```
docker run --rm -it --network=host  qdrant/bfb:dev ./bfb -n 30000  --indexing-threshold 1000


^Y [00:00:00] ░░░░░░░░░░░░░░░░░ [1,372,017.2535/s] 100/30000 (eta:0

thread 'main' panicked at src/main.rs:486:10:
called `Result::unwrap()` on an `Err` value: status: Cancelled, message: "Timeout expired", details: [], metadata: MetadataMap { headers: {} }

Caused by:
    0: transport error
    1: Timeout expired
```

timvisee:
That's weird. In all my testing I've never seen this myself.

Thanks for providing a base case to test.

generall (changes requested):
There is a deadlock problem on upsert

timvisee:
I found out what is causing this. When upserting we send update operations over an update channel. This update channel is not consuming these in time because it is waiting for CPU budget to be available. When the channel is full, upserting gets stuck because it cannot send more updates over  the channel.

I'll try to fix this on Monday.

timvisee:
> There is a deadlock problem on upsert

This problem has now been resolved.

We have a channel over which we send optimizer signals. Updates send a message over this channel to trigger optimizer. Our optimizer worker was waiting for CPU budget to be available, which prevented this channel from being drained in time, filling it up. New upsertions would be blocked on sending operations to this channel.

The fix in 61c5dc423b41218140e6d59e4d9e4d3508f468e2 prevents blocking this worker. Instead we change back to what we had before and spawn an asynchronous task that triggers the optimizers one more time once CPU budget becomes available again.

cc @generall

timvisee:
I've done some basic benchmarking [here](https://github.com/qdrant/qdrant/pull/3115#issuecomment-1915099907).

I guess this is good enough to consider this as notable improvement. Please let me know if you have any other cases I should test.

cc @agourlay

---

## Indexing benchmark

I've done some benchmarking to measure the indexing performance of this PR. I used a machine with 64 CPU cores, using the **default configuration**. For each test I insert a million points with indexing disabled, then I enable it again and measure indexing time. _Old_ refers to the current latest `dev` branch, while _new_ refers to this PR.

I don't have fancy graphs, but I think the list below is good enough to show off the gains.

### Results
**2 CPU cores:**
- `bfb -n 100000`:
  - Old: 40s
  - New: 40s

**8 CPU cores:**
- `bfb -n 1000000`:
  - Old: 163s
  - New: 186s (14% slower)
    - This is slower because the new approach uses 7 cores rather than 8. We do this intentionally to spare a core for handling other stuff, such as ongoing searches. Users can configure this with `storage.performance.cpu_budget`. We can also consider to change this default [here](https://github.com/qdrant/qdrant/blob/c480b3ea1172789f5f57f30af86a43d14917b5b8/lib/common/common/src/defaults.rs#L6-L23).

**16 CPU cores:**
- `bfb -n 1000000`:
  - Old: 123s
  - New: 92s (34% faster)
- `bfb -n 1000000 --segments 4`:
  - Old: 126s
  - New: 113s (12% faster)
- `bfb -n 1000000 --segments 8`:
  - Old: 123s
  - New: 92s (34% faster)

**32 CPU cores:**
- `bfb -n 1000000`:
  - Old: 121s
  - New: 54s (124% faster)
- `bfb -n 1000000 --segments 4`:
  - Old: 127s
  - New: 96s (32% faster)
- `bfb -n 1000000 --segments 8`:
  - Old: 121s
  - New: 54s (124% faster)
- `bfb -n 1000000 --segments 16`:
  - Old: 113s
  - New: 53s (113% faster)

**64 CPU cores:**
- `bfb -n 1000000`:
  - Old: 122s
  - New: 47s (160% faster)
- `bfb -n 1000000 --segments 4`:
  - Old: 129s
  - New: 74s (74% faster)
- `bfb -n 1000000 --segments 8`:
  - Old: 119s
  - New: 48s (148% faster)
- `bfb -n 1000000 --segments 16`:
  - Old: 120s
  - New: 43s (179% faster)

When using 64 cores, here we see CPU saturation in `htop` without this PR:
Screenshot from 2024-01-29 16-10-32

And here it is with this PR:
Screenshot from 2024-01-29 16-13-03

lib/collection/src/operations/conversions.rs
```diff
@@ -469,9 +469,8 @@ impl From<api::grpc::qdrant::OptimizersConfigDiff> for OptimizersConfig {
             memmap_threshold: optimizer_config.memmap_threshold.map(|x| x as usize),
             indexing_threshold: optimizer_config.indexing_threshold.map(|x| x as usize),
             flush_interval_sec: optimizer_config.flush_interval_sec.unwrap_or_default(),
-            max_optimization_threads: optimizer_config
-                .max_optimization_threads
-                .unwrap_or_default() as usize,
+            max_optimization_threads: optimizer_config.max_optimization_threads.unwrap_or(1)
```
agourlay, line 472: It feels like this new default to `1` is a new logic, I am unsure if the gRPC conversion should be responsible for it.
Why do we need this now?

config/config.yaml
```diff
@@ -108,27 +107,34 @@ storage:
...
     # Minimal size (in KiloBytes) of vectors for additional payload-based indexing.
     # If payload chunk is smaller than `full_scan_threshold_kb` additional indexing won't be used -
     # in this case full-scan search should be preferred by query planner and additional indexing is not required.
     # Note: 1Kb = 1 vector of size 256
     full_scan_threshold_kb: 10000
-    # Number of parallel threads used for background index building. If 0 - auto selection.
+
+    # Number of parallel threads used for background index building.
+    # If 0 - automatically select.
+    # Best to keep between 8 and 16 to prevent likelihood of building broken/inefficient HNSW graphs.
```
agourlay, line 131: this comment is a bit scary, we would need an official documentation to back it up

lib/common/common/src/cpu.rs
```diff
@@ -1,7 +1,196 @@
...
+    pub async fn wait_until_budget(&self, desired_cpus: usize) {
+        let min_required = self.min_permits(desired_cpus);
+        if self.has_budget_exact(min_required) {
+            return;
+        }
+
+        // Wait for CPU budget to be available with exponential backoff
+        // TODO: find better way, don't busy wait
+        let mut delay = Duration::from_micros(100);
+        while !self.has_budget_exact(min_required) {
```
agourlay, line 139: Can we have an upper bound on the wait time here instead of waiting forever?

lib/common/common/src/cpu.rs
```diff
@@ -1,7 +1,196 @@
...
+    ///
+    /// Waits for at least the minimum number of permits based on the given desired CPUs. For
+    /// example, if `desired_cpus` is 8, this will wait for at least 4 to be available. See
+    /// [`Self::min_permits`].
+    ///
+    /// - `1` to wait for any CPU budget to be available.
+    /// - `0` will always return immediately.
+    ///
+    /// Uses an exponential backoff strategy up to 10 seconds to avoid busy waiting.
+    pub async fn wait_until_budget(&self, desired_cpus: usize) {
```
agourlay, line 130: This function waits until the budge is available but does not guarantee that the caller will be able to use it?
I am wondering what is happening when we have several tasks waiting on the same budge.

lib/common/common/src/cpu.rs
```diff
@@ -1,7 +1,196 @@
...
+/// CPU permit, used to limit number of concurrent CPU-intensive operations
+///
+/// This permit represents the number of CPUs allocated for an operation, so that the operation can
+/// respect other parallel workloads. When dropped or `release()`-ed, the CPUs are given back for
+/// other tasks to acquire.
+///
+/// These CPU permits are used to better balance and saturate resource utilization.
+pub struct CpuPermit {
+    /// Number of CPUs acquired in this permit.
+    pub num_cpus: u32,
```
agourlay, line 161: Maybe we can use a `u16` instead to represent the number of cpu :)

generall (approved):

lib/common/common/src/cpu.rs
generall: have you seen those nvidia monsters? 
image

config/config.yaml
generall: Not related to the PR, and also I think was fixed already.

lib/collection/src/operations/conversions.rs
generall: those changes are only needed for the possibility of doing patch release. It will be reverted in the next PR

lib/common/common/src/cpu.rs
generall: yeah, I guess I would rather name it "notify on budget change" in this case.

lib/collection/src/update_handler.rs
```diff
@@ -401,12 +461,46 @@ impl UpdateHandler {
...
+                    // Continue if we have enough CPU budget available to start an optimization
+                    // Otherwise skip now and start a task to trigger the optimizer again once CPU
+                    // budget becomes available
+                    let desired_cpus = num_rayon_threads(max_indexing_threads);
+                    if !optimizer_cpu_budget.has_budget(desired_cpus) {
+                        let trigger_active = cpu_available_trigger
+                            .as_ref()
+                            .map_or(false, |t| !t.is_finished());
+                        if !trigger_active {
+                            cpu_available_trigger.replace(trigger_optimizers_on_cpu_budget(
```
generall, line 474: I would say, this function is not exactly required: if we don't have CPU budget to run single optimization, that would mean that at least one another optimization is already running. When optimization finishes - it generates exactly same signal as `trigger_optimizers_on_cpu_budget`. 

On the other hand, having this function doesn't hurt either, cause then we can avoid relying on external logic.
And even if it will trigger one extra signal, it won't create any more computations.

lib/collection/src/operations/conversions.rs
timvisee: That's correct indeed.

lib/common/common/src/cpu.rs
timvisee: Is there a good reason for setting an upper bound in this case? I don't really see a way for this getting stuck forever.

lib/common/common/src/cpu.rs
timvisee: > This function waits until the budge is available but does not guarantee that the caller will be able to use it?
I am wondering what is happening when we have several tasks waiting on the same budge.

You're right. We cannot do this in an atomic way.

The idea is that we don't really care whether we can eventually claim it, because we'd just try it again later. It is more like a notification: hey, you can give it another try.

I like the name suggestion, I'll make that change now.

lib/common/common/src/cpu.rs
timvisee: I used a `u32` because we pass this type to the semaphore when claiming permits: https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.acquire_many

Do you think that's a good enough reason?

agourlay: Let's keep `u32` :+1:

config/config.yaml
timvisee: > Best to keep between 8 and 16 to prevent likelihood of building broken/inefficient HNSW graphs.

Do you mean this specifically?

It is still in there, because a user _could_ still set something much higher. This actually happened with a user in Discord last Friday or Monday.

lib/common/common/src/cpu.rs
agourlay: >  I don't really see a way for this getting stuck forever. It it would get stuck, it wouldn't be a huge problem.

Is it because the number of waiting tasks is at most equal to the number of optimizers running?

lib/collection/src/update_handler.rs
timvisee: > I would say, this function is not exactly required: if we don't have CPU budget to run single optimization, that would mean that at least one another optimization is already running. When optimization finishes - it generates exactly same signal as trigger_optimizers_on_cpu_budget.

I thought this too, at first, but this is not the case.

It assumes that we only have one shard in which we do optimizations, with one instance of CPU budget.

This becomes a problem if you have multiple shards, because these share CPU budget (as you only have one SoC). Imagine that you have two shards. Shard A claimed all CPU budget, shard B will not start optimization because there is no CPU budget left. If optimizations on shard A finish, we don't trigger shard B. B will then only be triggered by external factors, such as upsertions.

That's the only reason for having this logic here. I'd love to remove it though, but I don't see a better way of doing things.

lib/common/common/src/cpu.rs
timvisee: The number of waiting tasks never exceeds 1 per shard. That is because we only spawn a 'wait task' once per shard if we run out of CPU budget.

Waiting this way is fair, meaning that there would be no waiter having an unfair advantage, always going first.
Though, a shard with a lot of ongoing updates may claim CPU budget before our wait function notices, but I don't see that as a problem.

CPU budget is eventually always returned once optimizations complete because it's handled in on-drop.

Now, if an optimization would get stuck somehow CPU budget is not returned. We could timeout in this wait function, but it wouldn't make much sense. If we'd do that, we still wouldn't be able to claim CPU budget afterwards.

Maybe it's important to add this: every time an upsertion happens, we try to claim CPU budget again. We're never stuck at this wait function, it's only spawned once on the side to trigger optimizations one more time. That is so we ensure we also optimize if we don't have upsertions.

Here it is in code. If we do any upsertion, we always go over [this](https://github.com/qdrant/qdrant/blob/08930fefa940729064f649c1f7ec0bef53daad66/lib/collection/src/update_handler.rs#L469) if-statement. If we have budget available at that time, we continue and try to optimize. If we don't, we make sure we spawn a separate task [here](https://github.com/qdrant/qdrant/blob/08930fefa940729064f649c1f7ec0bef53daad66/lib/collection/src/update_handler.rs#L473C25-L479C26) (once) with the above wait function, it triggers this loop again when CPU budget is available. While that task is ongoing, all new upsertions will try to pass this if-statement again.

Sorry for the wall of text, but I hope it clarifies :smile: 

Based on that, do you still think this can be an issue?

agourlay: Thank you very much for the wall of text, it starts clearing up the fog :face_in_clouds: 

I think my only remaining concerns is the number of notification tasks.
In case of a very heavy upsertion stream, are we going to spawn a lot of notification tasks if an HNSW indexing is currently running?

timvisee: > In case of a very heavy upsertion stream, are we going to spawn a lot of notification tasks if an HNSW indexing is currently running?

No, a maximum of 1 per shard. And there is no concurrency here, since each shard has a single optimization worker.

So if you'd have 1000 shards, then you could reach 1000 notification tasks. But that would be an unreasonable number anyway.

agourlay (approved):
Approving because:

- there are good performance numbers reported
- I tested it manually
- I _think_ I understand the state machine

timvisee:
:pray: thank you all for going over this! Merging now. :tada:

PR #3367, opened 2024-01-09
Make update API cancel safe

ffuugoo:
This PR makes update API (both REST and gRPC) cancel safe, so that if client drops the connection (e.g., due to timeout) it won't cause Qdrant to cancel update request in progress in a way that might break some invariant locally or introduce an inconsistency between nodes.

__TODO:__
- [x] Fix tests
  - e.g., propagate `cancel::CancellationToken` where needed
- [x] Evaluate and resolve `TODO`s
  - e.g., there are a bunch of places where some `update*` method is called that requires `cancel::CancellationToken`, but the caller of `update*` does not have cancellation token propagated into it yet

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/toc/point_ops.rs
+++ b/lib/storage/src/content_manager/toc/point_ops.rs
@@ -9,7 +9,8 @@ use collection::operations::shard_selector_internal::ShardSelectorInternal;
 use collection::operations::types::*;
 use collection::operations::CollectionUpdateOperations;
 use collection::{discovery, recommendations};
-use futures::future::try_join_all;
+use futures::stream::FuturesUnordered;
+use futures::TryStreamExt as _;
 use segment::types::{ScoredPoint, ShardKey};
 
 use super::TableOfContent;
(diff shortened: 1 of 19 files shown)


generall:
Would it be possible to simplify this PR by not implementing the cancellation token?
My assumption is that we have a critical section, which should be always complete, no matter timeout of the request.
But what is the purpose of the cancellation token for us? 
Why can't we just not await for the rest of the code outside of the critical section?

ffuugoo:
> Would it be possible to simplify this PR by not implementing the cancellation token?

Sure. (But, IMO, using the cancellation token is _correct_ way to do it.)

> My assumption is that we have a critical section, which should be always complete, no matter timeout of the request. But what is the purpose of the cancellation token for us? Why can't we just not await for the rest of the code outside of the critical section?

When client drops the connection, the top-level future (e.g., `upsert_points` async function in `update_api.rs` or `PointsService::upsert` method in `points_api.rs`) is stopped being polled and dropped.

E.g., if the future call stack was deep inside `ShardReplicaSet::update` doing requests to remote shards, it may be cancelled before all of them complete (and so introduce an inconsistency, cause some shards would have the change and some would not).

To fix this, we have to `tokio::spawn` a `ShardReplicaSet::update` _or its caller-future_, so that even if top-level future is dropped, `ShardReplicaSet::update` will still run to completion.

It's impossible to spawn `ShardReplicaSet::update` directly, cause it reference a lot of stuff with `&` references, so we have to just spawn `TableOfContent::update`.

__So, the cancellation token is used, so that we can still cancel the request _if it did not reach `ShardReplicaSet::update` yet_.__ (You can see that all locks are wrapped with `cancel::future::cancel_on_token`.)

IMO, this is the right way to do it. E.g.:
- we have a rate limiter for client update request
- If Qdrant is flooded with requests, and they are "stuck" on the rate limiter, and we __don't__ use cancellation token
- then even if client times outs and drops the request, we will still keep update "queued"
  - (which would also be impossible to track, cause it would just be some future spawned on the runtime)
- and if the client retries the request, it would get "stuck" on the rate limiter again, adding _more_ useless requests to the "queue"
- and __with__ cancellation token we can cancel all requests that are waiting on rate limiter/some lock if the client disconnects

ffuugoo:
> Would it be possible to simplify this PR by not implementing the cancellation token?

Reorganized commits a bit. Now 8add532 just `tokio::spawn`s everything, and further commits add cancellation tokens. (But I still think cancellation token is the proper way to implement it.)

@generall

generall (approved):

generall (approved):
LGTM

timvisee (approved):

lib/storage/src/content_manager/toc/point_ops.rs
```diff
@@ -246,29 +247,39 @@ impl TableOfContent {
...
         wait: bool,
         ordering: WriteOrdering,
     ) -> Result<UpdateResult, StorageError> {
-        if shard_keys.is_empty() {
-            return Err(StorageError::bad_input("Empty shard keys selection"));
-        }
+        // `Collection::update_from_client` is cancel safe, so this method is cancel safe.
 
-        let updates: Vec<_> = shard_keys
+        // TODO: Is this cancel safe!? I.e., is it safe to cancel *multiple* `update_from_client` requests?
```
timvisee, line 262: Do we still need to resolve this?

I'd assume it is **not** cancel safe, because it may apply an operation partially.

timvisee: It is cancel safe, I've described this in the function documentation in a new commit.

PR #3374, opened 2024-01-10
Fix potential context search edge case

coszio:
Context search combines the "loss"es of multiple positive-negative pairs, right now the formula is 

$L(x)=min(s(a,p)-s(a,n), 0)$
where $s(v_1, v_2)$: similarity between two vectors, $a$: vector we are evaluating, $p$: positive vector, $n$: negative vector

This function creates a __linear__ loss once a point $a$ is closer to the negative vector. However, it may be problematic, as pointed out by @generall, when a pair of examples is at the periphery of the embedding space, and the negative vector points towards the rest of the points, relative to the positive vector.

This makes the combination with other pairs hard, because the loss of single pair can eclipse the loss of other pairs. There is no lower limit to how much loss is assigned to a pair.

By plugging in our current loss function into a sigmoid(-ish) function, we get a non-linear loss which is always in the range of (-1, 0], for instance: $sig(L(x))$

Graphical modeling on a 2-D scale of the loss profile: https://www.desmos.com/calculator/lbxycyh2hs
imagen

Black line: sample anchor point. Red continuous line: new loss. Purple line: current loss

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/vector_storage/query/context_query.rs
+++ b/lib/segment/src/vector_storage/query/context_query.rs
@@ -130,7 +130,6 @@ mod test {
         *x as ScoreType
     }
 
-
     /// Possible similarities
     fn sim() -> impl Strategy<Value = f32> {
         (-100.0..=100.0).prop_map(|x| x as f32)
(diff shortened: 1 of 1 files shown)


generall:
I will test it on demo dataset

timvisee (approved):
Implementation looks good, but someone else has to test this against a real input :+1:

generall:
On some real-life examples I can't see neither clear improvement, nor clear degradation. Overall I think it makes to merge, as from the theoretical point of view the change makes sense

generall (approved):

PR #3375, opened 2024-01-10
Issues API: internal crate

coszio:
Tracked by #3471.

Creates a new internal crate for submitting and solving issues related to possible improvements of qdrant usage.

This works in a logger-like fashion, in which you have a singleton for using it from anywhere in the code. All issues are compared only using their `code`.

Sample usage (from the module documentation):
```rust
 use issues;
 use issues::{Issue, Solution};

 pub(crate) struct DummyIssue {
   pub code: String,
 }

 impl DummyIssue {
     pub fn new(code: impl Into<String>) -> Self {
         Self { code: code.into() }
     }
 }

 impl Issue for DummyIssue {
     fn code(&self) -> String {
         self.code.clone()
     }

     fn description(&self) -> String {
         "".to_string()
     }

     fn solution(&self) -> Solution {
         Solution::None
     }
 }

 // Submit an issue
 issues::submit(DummyIssue::new("issue1"));
 assert!(!issues::submit(DummyIssue::new("issue1"))); // can't submit the same issue twice

 // Solve an issue
 issues::solve("issue1"); // returns true if the issue was solved
 assert!(issues::submit(DummyIssue::new("issue1"))); // Now we can submit it again
```

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/common/issues/Cargo.toml
+++ b/lib/common/issues/Cargo.toml
@@ -2,7 +2,7 @@
 name = "issues"
 version = "0.0.0"
 authors = [
-    "Andrey Vasnetsov <vasnetsov93@gmail.com>",
+    "Andrey Vasnetsov <andrey@vasnetsov.com>",
     "Qdrant Team <info@qdrant.tech>",
 ]
 license = "Apache-2.0"
--- a/lib/common/issues/Cargo.toml
+++ b/lib/common/issues/Cargo.toml
@@ -11,8 +11,8 @@ publish = false
 
 [dependencies]
 http = "1.0.0"
-serde = "1.0.194"
 http-serde = "2.0.0"
-serde_json = "1.0.111"
 log = "0.4.20"
+serde = "1.0.194"
+serde_json = "1.0.111"
 serial_test = { version = "3.0.0", default-features = false }
--- a/lib/common/issues/Cargo.toml
+++ b/lib/common/issues/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "issues"
+version = "0.0.0"
+authors = [
+    "Andrey Vasnetsov <vasnetsov93@gmail.com>",
+    "Qdrant Team <info@qdrant.tech>",
+]
+license = "Apache-2.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+http = "1.0.0"
+serde = "1.0.194"
+http-serde = "2.0.0"
+serde_json = "1.0.111"
--- a/lib/common/issues/Cargo.toml
+++ b/lib/common/issues/Cargo.toml
@@ -13,6 +13,7 @@ publish = false
 http = "1.0.0"
 http-serde = "2.0.0"
 log = "0.4.20"
+schemars = "0.8.16"
 serde = "1.0.194"
 serde_json = "1.0.111"
 serial_test = { version = "3.0.0", default-features = false }
--- a/lib/common/issues/Cargo.toml
+++ b/lib/common/issues/Cargo.toml
@@ -10,11 +10,11 @@ edition = "2021"
 publish = false
 
 [dependencies]
-chrono = "0.4.33"
+chrono = { version = "0.4.33", features = ["serde"] }
 http = "1.0.0"
 http-serde = "2.0.0"
 log = "0.4.20"
-schemars = "0.8.16"
+schemars = { version = "0.8.16", features = ["chrono"] }
 serde = "1.0.194"
 serde_json = "1.0.111"
 serial_test = { version = "3.0.0", default-features = false }
--- a/lib/common/issues/Cargo.toml
+++ b/lib/common/issues/Cargo.toml
@@ -10,6 +10,7 @@ edition = "2021"
 publish = false
 
 [dependencies]
+chrono = "0.4.33"
 http = "1.0.0"
 http-serde = "2.0.0"
 log = "0.4.20"
--- a/lib/common/issues/Cargo.toml
+++ b/lib/common/issues/Cargo.toml
@@ -11,6 +11,7 @@ publish = false
 
 [dependencies]
 chrono = { version = "0.4.33", features = ["serde"] }
+dashmap = "5.5.3"
 http = "1.0.0"
 http-serde = "2.0.0"
 log = "0.4.20"
--- a/lib/common/issues/Cargo.toml
+++ b/lib/common/issues/Cargo.toml
@@ -15,7 +15,7 @@ dashmap = "5.5.3"
 http = "1.0.0"
 http-serde = "2.0.0"
 log = "0.4.20"
-schemars = { version = "0.8.16", features = ["chrono"] }
-serde = "1.0.194"
-serde_json = "1.0.111"
+schemars = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
 serial_test = { version = "3.0.0", default-features = false }
--- a/lib/common/issues/Cargo.toml
+++ b/lib/common/issues/Cargo.toml
@@ -14,3 +14,5 @@ http = "1.0.0"
 serde = "1.0.194"
 http-serde = "2.0.0"
 serde_json = "1.0.111"
+log = "0.4.20"
+serial_test = { version = "3.0.0", default-features = false }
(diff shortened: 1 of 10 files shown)


lib/common/issues/src/dashboard.rs
```diff
@@ -0,0 +1,175 @@
+use std::collections::HashSet;
+use std::sync::{Mutex, MutexGuard, OnceLock};
+
+use serde::{Serialize, Serializer};
+
+use crate::issue::{DummyIssue, Issue};
+
+#[derive(Default, Serialize)]
+struct Dashboard {
+    pub issues: HashSet<Box<dyn Issue>>,
```
agourlay, line 10: It seems the Hash & Eq is determined solely by the `code` field.
We could use a `Hashmap<code, Issue>` instead to keep a generic definition of hash and equality on the `Issue` type.

coszio: Great point 😄

lib/common/issues/src/problems/unindexed_field.rs
```diff
@@ -0,0 +1,50 @@
...
+            "field_name": self.field_name,
+            "field_schema": self.field_type,
+        });
+
+        Ok(Solution {
+            message: format!(
+                "Create an index on field '{}' of type '{}' in collection '{}'. Check the documentation for more details: https://qdrant.tech/documentation/concepts/indexing/#payload-index",
+                self.field_name, self.field_type, self.collection
+            ),
+            action: Some(Action {
```
agourlay, line 42: are we going to run solution automatically?

lib/common/issues/src/issue.rs
```diff
@@ -0,0 +1,66 @@
...
+use serde::ser::SerializeMap;
+use serde::{Serialize, Serializer};
+
+use crate::solution::Solution;
+
+/// An issue that can be hashed by its code
+pub trait Issue: Sync + Send {
+    fn code(&self) -> String;
+    fn description(&self) -> String;
+    fn solution(&self) -> Result<Solution, String>;
```
agourlay, line 12: It looks like the `solution` field uses `Result` as a generic either type, without any bias on the side.
It is a bit confusing.
Maybe we need an Enum instead?

lib/common/issues/src/problems/unindexed_field.rs
coszio: No, only provide it as an option to the user. It will be specially useful for cloud to make one-click fixes

lib/common/issues/src/issue.rs
timvisee: Agreed, that would be better.

```rust
enum Solvable {
    Yes(Solution),
    No(String),
}
```

or something like that.

lib/common/issues/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "issues"
+version = "0.0.0"
+authors = [
+    "Andrey Vasnetsov <vasnetsov93@gmail.com>",
```
generall, line 5:
```suggestion
    "Andrey Vasnetsov <andrey@vasnetsov.com>",
```

lib/common/issues/src/dashboard.rs
generall: agreed, making custom Hashing looks hacky

coszio: Turned into HashMap. It does make it clearer

lib/common/issues/src/issue.rs
coszio: I like making it an enum instead. The `Err` variant was for when a solution couldn't be constructed for some reason, like when parsing a url. But in that case I thing the better thing to do is log and return `Solution::None`

coszio (commented):
This is now ready for another review round :)
cc @agourlay @generall

lib/common/issues/src/dashboard.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// Deactivates an issue by its code, returning true if the issue was active before
+    fn remove_issue<S: AsRef<str>>(&mut self, code: S) -> bool {
+        self.issues.remove(code.as_ref()).is_some()
+    }
+}
+
+fn dashboard<'a>() -> MutexGuard<'a, Dashboard> {
+    static DASHBOARD: OnceLock<Mutex<Dashboard>> = OnceLock::new();
+    match DASHBOARD
+        .get_or_init(|| Mutex::new(Dashboard::default()))
```
agourlay, line 29: It seems like a `RwLock` could be a fit as a few operations only need read access.
I'd recommend using directly the variant from `parking_lot`.

lib/common/issues/src/dashboard.rs
```diff
@@ -0,0 +1,177 @@
...
+/// Clears all issues from the dashboard
+pub fn clear() {
+    dashboard().issues.clear();
+}
+
+/// Solves (clears) all issues with the given prefix in their code
+pub fn prefix_solve(prefix: &str) {
+    dashboard()
+        .issues
+        .retain(|code, _issue| !code.starts_with(prefix));
```
agourlay, line 77: I am not sure I am a fan of the String matching here.
We would need to have a consistent view of all existing code when adding a new one.

lib/common/issues/src/dashboard.rs
coszio: I hadn't considered this, and now I want to change the access pattern for insertion, for example, because the same issue is possible to be generated many times. So that we can check for existence only with a read before acquiring write access

agourlay: I you want to avoid juggling with locks you could try to use a concurrent hashmap like https://github.com/xacrimon/dashmap

lib/common/issues/src/dashboard.rs
coszio: I'm not sure I understand what you mean here. Adding a new issue can only be done via `insert` operation. This operation **is** expected to be relatively heavy because of scanning the whole hashmap.

Anyway, I only added this as a utility function without a concrete case of when to use it. It should be fine if we remove it.

agourlay: Sorry for not being clear :)

What I meant is that the idea of doing issue grouping based on String prefix is bothering me a bit.

It means the String values are important because adding a new value without checking the existing ones could create problems where an issue is deleted by mistake

Is it clearer?

coszio: To avoid any premature implementations, I'll just remove these utility functions until there is an actual need for them.

lib/common/issues/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.0.0"
+authors = [
+    "Andrey Vasnetsov <andrey@vasnetsov.com>",
+    "Qdrant Team <info@qdrant.tech>",
+]
+license = "Apache-2.0"
+edition = "2021"
+publish = false
+
+[dependencies]
```
agourlay, line 12: A bunch of those deps version are now defined in the workspace setting https://github.com/qdrant/qdrant/pull/3490/files

Can you use it there as well?

lib/common/issues/src/dashboard.rs
```diff
@@ -0,0 +1,121 @@
...
+            return false;
+        }
+        let issue = IssueRecord::from(issue);
+        self.issues.insert(code, issue).is_none()
+    }
+
+    /// Deactivates an issue by its code, returning true if the issue was active before
+    fn remove_issue<S: AsRef<str>>(&self, code: S) -> bool {
+        if self.issues.contains_key(code.as_ref()) {
+            return self.issues.remove(code.as_ref()).is_some();
```
agourlay, line 26: I think you could just do the `remove` operation.
It is most likely optimized will prevent any kind of race condition on our side.

coszio: Just wanted to make sure it does not request write access without checking with a read access. But it makes sense that it is optimized for these situations. I'll check the source code.

coszio: After checking the source code for both insert and remove, I found that there is no read before write optimization, see https://docs.rs/dashmap/latest/src/dashmap/lib.rs.html#939

So I'll keep this since it is very probable that issues will want to be re-added or re-removed many times

agourlay (approved):
I think this reached an acceptable state where I do not want to block you anymore.

Ultimately we need to see how it integrates in the code base to decide if the approach works.

PR #3380, opened 2024-01-11
Parameterize integer index, allow lookup or range exclusively

timvisee:
Alternative for <https://github.com/qdrant/qdrant/pull/3191>.

This parameterizes the integer index, to allow indexing for just lookup xor range. This allows fine tuning in very specific cases.

<https://github.com/qdrant/qdrant/pull/3191> explains why this is useful, but this PR hosts a different implementation for it.

This PR supports setting `lookup`/`range` flags on the integer index, and does prevent creating a bunch of new index types. The idea is that this is more extensible in case we ever want to add more flags. It matches the same layout as we use [here](https://qdrant.tech/documentation/concepts/indexing/#full-text-index) for a parameterized text index.

For example, two REST requests to just enable `lookup` or `range:

```
PUT /collections/{collection}/index
{
    "field_name": "test",
    "field_schema": {
      "type": "integer",
      "lookup": true,
      "range": false
    }
}

PUT /collections/{collection}/index
{
    "field_name": "test",
    "field_schema": {
      "type": "integer",
      "lookup": false,
      "range": true
    }
}
```

The regular unparameterized integer index still works as well:

```
PUT /collections/{collection}/index
{
    "field_name": "test",
    "field_schema": "integer"
}
```

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


generall:
Aesthetically, it appeals better.

Also, it we would ever need some other parameters for integer index, it would be easier to add it in this option. So I vote for this.

timvisee:
I've updated this implementation to use the less-nice internals, with an explicit `type` field, which we had in text parameterization before. Sadly we need this to generate proper OpenAPI schemas.

See my comment on this in the description above:

> Internally, this uses the same approach as our existing parameterized text index. It is ugly, but required to generate properly named OpenAPI schemas. In the branch [`int-index-params-clean`](https://github.com/qdrant/qdrant/tree/int-index-params-clean) I introduce the significantly nicer internals in 95f42509341d1e637b9b33486e757320fd84f3df, but indeed this breaks OpenAPI schema generation. So we'll probably have to go through with this implementation instead.

dzhao:
this parameter setup adds more flexibility. lgtm

PR #3381, opened 2024-01-12
add optional checksum to snapshot recovery request

### LINKED ISSUE #3371: Snapshot recovery should validate checksum if available
In <https://github.com/qdrant/qdrant/pull/2840> we've added checksums for snapshot files. The implementation is somewhat limited however, and requires further integration to make full use of it.

Currently, a malformed snapshot file can be recovered even if the integrity check with an attached checksum would fail. This is problematic.

When a snapshot is being recovered, it should verify the attached checksum if any is provided. If this fails, recovery should be rejected.

Qdrant has a few REST and gRPC endpoints to trigger snapshot recovery. There are two flavors, recovering from an URI and by uploading a snapshot file directly. Both should get support for checksum integrity checks. These endpoints are described here: https://qdrant.tech/documentation/concepts/snapshots/#recover-via-api

A user should be able to (optionally) provide a checksum along with the request, so that Qdrant can verify its integrity. If the integrity check fails, we should clean up and return an error. If no checksum is provided we skip the integrity check and continue recovering like we do now.

A recovery request with checksum might look something like this:

```
PUT /collections/{collection_name}/snapshots/recover
{
  "location": "http://qdrant-node-1:6333/collections/{collection_name}/snapshots/snapshot-2022-10-10.shapshot",
  "checksum": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
}
```

So, in short:
- in all REST and gRPC snapshot recovery endpoints, add optional `checksum` field
- if a checksum is provided, Qdrant must verify the snapshot file integrity
- if integrity check with checksum failed, delete snapshot file and return an error

**Additional context**
There's other work to be done to properly integrate checksums, but that will be handled in different issues/PRs.

xzfc:
Add `checksum` option to `RecoverShardSnapshotRequest` (GRPC), `/collections/{collection_name}/snapshots/recover`, `/collections/{collection}/shards/{shard}/snapshots/recover`, and (REST).
Closes #3371.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -9738,6 +9738,12 @@
                 "nullable": true
               }
             ]
+          },
+          "checksum": {
+            "description": "Optional SHA256 checksum to verify snapshot integrity before recovery.",
+            "default": null,
+            "type": "string",
+            "nullable": true
           }
         }
       },
@@ -10409,6 +10415,12 @@
                 "nullable": true
               }
             ]
+          },
+          "checksum": {
+            "description": "Optional SHA256 checksum to verify snapshot integrity before recovery.",
+            "default": null,
+            "type": "string",
+            "nullable": true
           }
         }
       },
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -1630,6 +1630,15 @@
             "schema": {
               "$ref": "#/components/schemas/SnapshotPriority"
             }
+          },
+          {
+            "name": "checksum",
+            "in": "query",
+            "description": "Optional SHA256 checksum to verify snapshot integrity before recovery.",
+            "required": false,
+            "schema": {
+              "type": "string"
+            }
           }
         ],
         "requestBody": {
@@ -1741,6 +1750,15 @@
             "required": false,
             "schema": {
               "type": "boolean"
+            }
+          },
+          {
+            "name": "checksum",
+            "in": "query",
+            "description": "Optional SHA256 checksum to verify snapshot integrity before recovery.",
+            "required": false,
+            "schema": {
+              "type": "string"
             }
           }
         ],
@@ -2486,6 +2504,15 @@
             "schema": {
               "$ref": "#/components/schemas/SnapshotPriority"
             }
+          },
+          {
+            "name": "checksum",
+            "in": "query",
+            "description": "Optional SHA256 checksum to verify snapshot integrity before recovery.",
+            "required": false,
+            "schema": {
+              "type": "string"
+            }
           }
         ],
         "requestBody": {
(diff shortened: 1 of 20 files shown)


src/actix/api/snapshot_api.rs
timvisee, line 1: We should also add a `checksum` parameter to `/collections/{name}/snapshots/upload`. It likely needs to be a query parameter in the `SnapshotUploadingParam` struct. We only have this extra endpoint in REST, not in gRPC.

timvisee (commented):
Thank you for your effort on this! :smile: 

I've left some review comments below.

src/actix/api/snapshot_api.rs
xzfc: Added for checksum checking for `/collections/{name}/snapshots/upload` and `/collections/{collection}/shards/{shard}/snapshots/upload`.

xzfc:
Updated according to comments, please take a look.

docs/redoc/master/openapi.json
timvisee, line 1: Could you add the `checksum` parameter for the upload endpoint(s) in `openapi/openapi-snapshots.ytt.yaml` and `openapi/openapi-shard-snapshots.ytt.yaml`? Then these will be properly reflected in this schema as well once you regenerate them.

Sadly we've to define these parameters manually at this time.

timvisee (changes requested):
Thanks a lot for your improvements!

I think there are two more things left to do:

1. update the OpenAPI specification, see the review comment below.
2. add a basic integration test.

Regarding the test, I imagine the simplest one to be:
- have a collection with some points
- create a snapshot, download the snapshot file
- delete the collection
- try to restore the file with a wrong hash (one different character?), assert error
- try to restore the file with correct hash, assert success
- preferably an integration test function per snapshot type (collection and shard snapshots)

You can probably add this test [here](https://github.com/qdrant/qdrant/tree/master/openapi/tests/openapi_integration), and reuse some components.

xzfc:
@timvisee Done.

Initially, I did not add tests because there were no existing pytests involving downloading/uploading files.
(there is `tests/snapshots/snapshots-recovery.sh`, but it seems it is not used on CI and is not very convenient to maintain IMO)
Now, I come up with a pytest fixture to serve files from the temporary directory over HTTP.

timvisee:
Thanks! I skimmed through it just now, and everything looks sound. I'll try to do a more in-depth review later today.

I don't think starting a web server to serve the files is strictly necessary for this. You _can_ also provide an `file:///some/dir` URI to grab a snapshot file from disk. But still, it doesn't hurt to keep this in as long as it doesn't give us any flakiness. The good thing is that this now also tests downloading snapshots from an URL, which we did not cover anywhere else yet. So, I suggest to keep it like this :+1:

timvisee (approved):
Perfect! I like that you also validate the hash in Python, and that you set the HTTP server up to be a fixture.

Before we can merge, would you mind to rebase on `dev` resolving the conflicts?

xzfc:
Rebased.

timvisee (approved):
Thanks once again! :raised_hands:

Issue #3376, opened 2024-01-11
qdrant does not update multiple named vectors on the same point


milosgajdos:
I might be missing something here with regard to how to update named vectors on the same point but when using the [Go client](https://github.com/qdrant/go-client), it would appear that only one of them gets updated. What gives?

## Steps to Reproduce

1. Create a collection with 2 named vectors
2. Upsert a point
3. Update named vectors on the same point
4. List the named vectors on the upserted point

Here is the Go code that can reproduce the issue:
<details><summary>Go code</summary>

```Go
package main

import (
	"context"
	"log"

	"github.com/google/uuid"
	pb "github.com/qdrant/go-client/qdrant"
	"google.golang.org/grpc"
	"google.golang.org/grpc/credentials/insecure"
)

var (
	defaultDistance = pb.Distance_Dot
)

func main() {
	conn, err := grpc.DialContext(context.Background(),
		"localhost:6334",
		grpc.WithTransportCredentials(insecure.NewCredentials()))
	if err != nil {
		log.Fatalf("Failed to connect: %v", err)
	}
	defer conn.Close()

	col := pb.NewCollectionsClient(conn)

	uid := uuid.New().String()
	ctx := context.Background()

	_, err = col.Create(ctx, &pb.CreateCollection{
		CollectionName: uid,
		VectorsConfig: &pb.VectorsConfig{
			Config: &pb.VectorsConfig_ParamsMap{
				ParamsMap: &pb.VectorParamsMap{
					Map: map[string]*pb.VectorParams{
						"": {
							Size:     4,
							Distance: defaultDistance,
						},
						"2D": {
							Size:     2,
							Distance: defaultDistance,
						},
						"3D": {
							Size:     3,
							Distance: defaultDistance,
						},
					},
				},
			},
		},
	})
	if err != nil {
		log.Fatalf("collection create failed: %v", err)
	}

	log.Printf("created collection: %s", uid)

	pts := pb.NewPointsClient(conn)

	pointUID := uuid.NewString()
	upsertPoints := []*pb.PointStruct{
		{
			Id: &pb.PointId{
				PointIdOptions: &pb.PointId_Uuid{
					Uuid: pointUID,
				},
			},
			Vectors: &pb.Vectors{
				VectorsOptions: &pb.Vectors_Vector{
					Vector: &pb.Vector{
						Data: []float32{1.0, 2.0, 3.0, 4.0},
					},
				},
			},
		},
	}

	waitUpsert := true
[...]

PR #3386, opened 2024-01-12
Fix multiple vector updates on same point in batch not working

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/3376>.

This PR fixes an issue where multiple vector (not point) updates on the same point don't work if they are in the same batch. Instead, only the last operation for each point was applied.

For example, the following would drop the first change, and only apply the second change, because they are for the same point ID. Updating `b` and leaving `a` unchanged:
```
PUT /collections/{collection}}/points/vectors
{
    "points": [
        {
            "id": 1,
            "vector": {
                "a": [0.8, 0.1, 0.3]
            }
        },
        {
            "id": 1,
            "vector": {
                "b": [0.2, 0.4, 0.3]
            }
        }
    ]
}
```

To fix this, all vector updates are now merged into a single operation per point. This merged from first to last, to ensure the last item in the batch has the highest priority.

I've also added a bunch of tests ensuring correctness, also covering the exact case in <https://github.com/qdrant/qdrant/issues/3376>.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/segments_updater.rs
+++ b/lib/collection/src/collection_manager/segments_updater.rs
@@ -49,8 +49,16 @@ pub(crate) fn update_vectors(
     op_num: SeqNumberType,
     points: &[PointVectors],
 ) -> CollectionResult<usize> {
-    let points_map: HashMap<PointIdType, &PointVectors> =
-        points.iter().map(|p| (p.id, p)).collect();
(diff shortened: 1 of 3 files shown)


generall (approved):

agourlay (approved):
Thank you very much for the extensive testing :trophy:

PR #3387, opened 2024-01-12
Disable large limit optimization for plain indices

coszio:
Interesting side-effect of Context Search made this shine: many (or all) results can have score of `0.0`, which triggers search re-runs. 

The main reason of this optimization is that using large search `ef` values is expensive, but currently it is also being used for non-indexed segments (where there is no `ef` value). This causes that even for small limits, there is sampling happening.

This fix disables the sampling optimization for when a plain index is being used.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -388,7 +388,11 @@ fn sampling_limit(
     let poisson_sampling =
         find_search_sampling_over_point_distribution(limit as f64, segment_probability)
             .unwrap_or(limit);
-    let effective = effective_limit(limit, ef_limit.unwrap_or(0), poisson_sampling);
+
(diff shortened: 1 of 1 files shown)


generall (approved):

timvisee (approved):

lib/collection/src/collection_manager/segments_searcher.rs
```diff
@@ -388,7 +388,11 @@ fn sampling_limit(
     let poisson_sampling =
         find_search_sampling_over_point_distribution(limit as f64, segment_probability)
             .unwrap_or(limit);
-    let effective = effective_limit(limit, ef_limit.unwrap_or(0), poisson_sampling);
+
+    // if no ef_limit was found, it is a plain index => sampling optimization is not needed.
```
agourlay, line 392: could it be a sparse vector search request as well?

PR #3393, opened 2024-01-15
Optimize sparse vector for real life distribution

agourlay:
This PR optimizes the sparse vector search for real life data.

# Context

The work done in https://github.com/qdrant/qdrant/pull/3347 & https://github.com/qdrant/qdrant/pull/3359 has yielded great gains on synthetic random data.

However, a more end to end test with a real SPLADE embeddings for MSMARCO revealed that the performance actually worsened compared to 1.7.3.

Using a min-heap to track the minimum record ids works well when posting lists have more or less the same size but **not** when the distribution is highly skewed. The cost of repeatedly pushing new values onto the heap starts to dominate.

Here is an example of real distribution where most posting lists are very short except a few which contains a large amount of data.

neurIPS_posting_len_example
 
# Optimization

In order to fix this issue it is necessary to rollback the min-heap approach to restore the performance of 1.7.3 and then try to get more performance on top.

This PR proposes a new approach to avoid both tracking the min record id explicitly and constantly computing it.

By relying on the fact that internal ids have no gaps and that the posting lists are sorted, we can optimistically increment the id manually and fallback to a search if it does not work.

This approach works well when the ids have no holes:
- with a low amount of segment to maximize contiguous ids
- with a low amount of deleted points

In the worst case, it performs an additional scan to find the minimum id like in 1.7.3

# Results

Using the [new sparse vector test bed](https://github.com/qdrant/sparse-vectors-benchmark) on the 100k dataset,

```
DEV:
Latency distribution:
50p: 61.06 millis
95p: 103.23 millis
99p: 123.85 millis
999p: 157.92 millis
max: 182.39 millis

v1.7.3
Latency distribution:
50p: 60.8 millis
95p: 104.22 millis
99p: 126.7 millis
999p: 151.25 millis
max: 179.12 millis

PR:
Latency distribution:
50p: 54.86 millis
95p: 85.31 millis
99p: 100.23 millis
999p: 115.24 millis
max: 126.31 millis
```

- dev

neurIPS_bench_small_dev_dim_count

- 1.7.3

neurIPS_bench_small_1 7 3_dim_count

- PR

neurIPS_bench_small_NEW_dim_count

### DIFF
--- a/lib/sparse/src/index/search_context.rs
+++ b/lib/sparse/src/index/search_context.rs
@@ -1,4 +1,4 @@
-use std::cmp::{max, Ordering};
+use std::cmp::Ordering;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering::Relaxed;
 
@@ -23,7 +23,7 @@ pub struct SearchContext<'a> {
     top: usize,
     is_stopped: &'a AtomicBool,
     result_queue: FixedLengthPriorityQueue<ScoredPointOffset>, // keep the largest elements and peek smallest
-    min_record_ids: FixedLengthPriorityQueue<PointOffsetType>, // min record ids across all posting lists
+    min_record_id: Option<PointOffsetType>, // min record ids across all posting lists
     use_pruning: bool,
 }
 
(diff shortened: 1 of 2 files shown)


lib/sparse/src/index/search_context.rs
```diff
@@ -140,27 +135,32 @@ impl<'a> SearchContext<'a> {
...
         }
 
+        // update min record id for next iteration
+        if found {
+            // assume the next min record id is the current one + 1
+            self.min_record_id = self.min_record_id.map(|min_id| min_id + 1);
+        } else {
+            // no match found, compute next min record id from all posting list iterators
+            self.min_record_id = Self::next_min_id(&self.postings_iterators);
+            return self.advance();
```
agourlay, line 161: the recursion is bounded as the nested call will always return right away because it is using the proper min record id

timvisee (approved):
:raised_hands: :chart_with_upwards_trend:

Issue #3320, opened 2024-01-04
Implement datetime support for the payload index


generall:
**Is your feature request related to a problem? Please describe.**

Currently, if users wants to search for specific data / time range, it is required to convert dates into unix timestamps (or similar) and work with dates as integers. It might inconvenient and annoying to do.

**Describe the solution you'd like**

Create a special type of index, which could automatically parse dates and create appropriate index for them (it could be same integer index internally).
User should be able to filter requests using date as well (in the range queries).

For this to work properly, we would need to extend Range data structure with ability to specify dates.

In rest interface it should look like this:

```json
{
  "must": [
    {
      "key": "my-date",
      "range": {
          "lte": "2024-12-01"
       }
    }
  ]
}
```

**Describe alternatives you've considered**

Create a dedicated filtering condition for dates, but that might be confusing.

**Additional context**

We need to be extra careful with backward compatibility in gRPC interface. We can't simply change Range structure into enum. Most likely in gRPC the best option would be to create another type "RangeDates" and implement proper conversions on server side.

Additionally, it might make sense to do same in REST (but reuse the same key `range`), so that it would be impossible to use 
dates and numbers in a single range statement.

--- 

Note for contributors: Please consider this as tracking issue. If you think that it would be beneficial to split the task into multiple smaller PRs, please you are welcome to do so. Bounty will be rewarded for each PR independently

xzfc:
Couldn't this use case be covered by providing a general-purpose index for strings instead? The lexicographical order of timestamps in a format like `2024-01-14 16:34:23` is the same as the chronological order, so range comparisons could be used for date strings as well.
Besides timestamps, this could be used for postal codes and so on.

timvisee:
> Couldn't this use case be covered by providing a general-purpose index for strings instead? The lexicographical order of timestamps in a format like `2024-01-14 16:34:23` is the same as the chronological order, so range comparisons could be used for date strings as well.

Probably, or possibly even better, a timestamp in an integer index.

But a dedicated index would allow for further optimization (and parameterization) if possible and/or required.

timvisee:
Implemented in <https://github.com/qdrant/qdrant/pull/3395>.

There are some follow-up issues, but these are handles separately:
- https://github.com/qdrant/qdrant/issues/3529
- https://github.com/qdrant/qdrant/issues/3531

PR #3395, opened 2024-01-15
Datetime payload index

xzfc:
This PR adds datetime payload index. /claim #3320

Open for discussion/feedback:
- Internal representation. Current format: milliseconds in `i64`. Possible alternatives: nanoseconds in `u64` or seconds in `f64`.
- Timestamp representation. Current format: RFC3339, e.g., `2019-01-01T00:00:00.000Z`.
- Currently `datetime_range` is a separate key. Need to investigate if this could be merged with the `range` key in a straightforward way.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -5770,7 +5770,8 @@
           "float",
           "geo",
           "text",
-          "bool"
+          "bool",
+          "datetime"
         ]
       },
       "PayloadSchemaParams": {
@@ -6345,7 +6346,18 @@
             "description": "Check if points value lies in a given range",
             "anyOf": [
               {
-                "$ref": "#/components/schemas/Range"
+                "$ref": "#/components/schemas/Range_for_double"
+              },
+              {
+                "nullable": true
+              }
+            ]
+          },
+          "datetime_range": {
+            "description": "Check if datetime is within a given range",
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/Range_for_DateTime"
               },
               {
                 "nullable": true
@@ -6494,7 +6506,7 @@
           }
         }
       },
-      "Range": {
+      "Range_for_double": {
         "description": "Range filter request",
         "type": "object",
         "properties": {
@@ -6520,6 +6532,36 @@
             "description": "point.key <= range.lte",
             "type": "number",
             "format": "double",
+            "nullable": true
+          }
+        }
+      },
+      "Range_for_DateTime": {
+        "description": "Range filter request",
+        "type": "object",
+        "properties": {
+          "lt": {
+            "description": "point.key < range.lt",
+            "type": "string",
+            "format": "date-time",
+            "nullable": true
+          },
+          "gt": {
+            "description": "point.key > range.gt",
+            "type": "string",
+            "format": "date-time",
+            "nullable": true
+          },
+          "gte": {
+            "description": "point.key >= range.gte",
+            "type": "string",
+            "format": "date-time",
+            "nullable": true
+          },
+          "lte": {
+            "description": "point.key <= range.lte",
+            "type": "string",
+            "format": "date-time",
             "nullable": true
           }
         }
--- a/lib/api/build.rs
+++ b/lib/api/build.rs
@@ -241,6 +241,10 @@ fn configure_validation(builder: Builder) -> Builder {
             ("DiscoveryQuery.context", ""),
             ("ContextQuery.context", ""),
         ], &[])
+        .field_attribute("DatetimeRange.lt", "#[serde(skip)]")
+        .field_attribute("DatetimeRange.gt", "#[serde(skip)]")
+        .field_attribute("DatetimeRange.lte", "#[serde(skip)]")
+        .field_attribute("DatetimeRange.gte", "#[serde(skip)]")
         .type_attribute(".", "#[derive(serde::Serialize)]")
         // Service: points_internal_service.proto
         .validates(&[
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -1,15 +1,17 @@
 use std::collections::{HashMap, HashSet};
 use std::time::Instant;
 
-use chrono::{NaiveDateTime, Timelike};
+use chrono::{NaiveDateTime, TimeZone as _, Timelike};
 use segment::data_types::integer_index::IntegerIndexType;
 use segment::data_types::text_index::TextIndexType;
 use segment::data_types::vectors::VectorElementType;
-use segment::types::default_quantization_ignore_value;
+use segment::types::{default_quantization_ignore_value, DateTimePayloadType, FloatPayloadType};
 use tonic::Status;
 use uuid::Uuid;
 
-use super::qdrant::{BinaryQuantization, CompressionRatio, GeoLineString, GroupId, SparseIndices};
+use super::qdrant::{
+    BinaryQuantization, CompressionRatio, DatetimeRange, GeoLineString, GroupId, SparseIndices,
+};
 use crate::grpc::models::{CollectionsResponse, VersionInfo};
 use crate::grpc::qdrant::condition::ConditionOneOf;
 use crate::grpc::qdrant::payload_index_params::IndexParams;
@@ -222,6 +224,7 @@ impl From<segment::types::PayloadIndexInfo> for PayloadSchemaInfo {
                 segment::types::PayloadSchemaType::Geo => PayloadSchemaType::Geo,
                 segment::types::PayloadSchemaType::Text => PayloadSchemaType::Text,
                 segment::types::PayloadSchemaType::Bool => PayloadSchemaType::Bool,
+                segment::types::PayloadSchemaType::Datetime => PayloadSchemaType::Datetime,
             }
             .into(),
             params: schema.params.map(|params| match params {
@@ -312,6 +315,7 @@ impl TryFrom<PayloadSchemaInfo> for segment::types::PayloadIndexInfo {
                 PayloadSchemaType::Geo => segment::types::PayloadSchemaType::Geo,
                 PayloadSchemaType::Text => segment::types::PayloadSchemaType::Text,
                 PayloadSchemaType::Bool => segment::types::PayloadSchemaType::Bool,
+                PayloadSchemaType::Datetime => segment::types::PayloadSchemaType::Datetime,
                 PayloadSchemaType::UnknownType => {
                     return Err(Status::invalid_argument(
                         "Malformed payload schema".to_string(),
@@ -942,6 +946,7 @@ impl TryFrom<FieldCondition> for segment::types::FieldCondition {
             geo_radius,
             values_count,
             geo_polygon,
+            datetime_range,
         } = value;
 
         let geo_bounding_box =
@@ -952,6 +957,7 @@ impl TryFrom<FieldCondition> for segment::types::FieldCondition {
             key,
             r#match: r#match.map_or_else(|| Ok(None), |m| m.try_into().map(Some))?,
             range: range.map(Into::into),
+            datetime_range: datetime_range.map(Into::into),
             geo_bounding_box,
             geo_radius,
             geo_polygon,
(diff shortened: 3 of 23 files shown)


lib/api/build.rs
```diff
@@ -241,6 +241,10 @@ fn configure_validation(builder: Builder) -> Builder {
             ("DiscoveryQuery.context", ""),
             ("ContextQuery.context", ""),
         ], &[])
+        .field_attribute("DatetimeRange.lt", "#[serde(skip)]")
+        .field_attribute("DatetimeRange.gt", "#[serde(skip)]")
+        .field_attribute("DatetimeRange.lte", "#[serde(skip)]")
+        .field_attribute("DatetimeRange.gte", "#[serde(skip)]")
```
coszio, line 247: Is it not possible to make a serializer for `prost_types::Timestamp`? 

We should also add validation for these types, to make sure they fit within 0001-01-01T00:00:00Z to 9999-12-31T23:59:59.999999999Z, as established by the [`prost_types::Timestamp` documentation](https://docs.rs/prost-types/latest/prost_types/struct.Timestamp.html#fields)

lib/segment/src/index/field_index/field_index_base.rs
```diff
@@ -116,6 +117,7 @@ pub trait ValueIndexer<T> {
 #[allow(clippy::enum_variant_names)]
 pub enum FieldIndex {
     IntIndex(NumericIndex<IntPayloadType>),
+    DatetimeIndex(NumericIndex<IntPayloadType>),
```
coszio, line 120: Is this enum variant really needed? I'm thinking maybe we can redirect to `IntIndex` for datetime payloads without this.

openapi/tests/openapi_integration/test_payload_indexing.py
```diff
@@ -134,6 +134,42 @@ def test_boolean_index():
...
+        path_params={'collection_name': collection_name},
+        query_params={'wait': 'true'},
+        body={
+            "field_name": datetime_key,
+            "field_schema": "datetime"
+        }
+    )
+    assert response.ok
+
+    response = request_with_validation(
```
coszio, line 156: can we add at least one test with different timezones?

coszio (changes requested):
Great work here @xzfc! 
My only concern is the at the api layer. We should keep the `Range` type at this level to not break backward compatibility in the clients. A newtype instead of generic for datetime range may fix this.

Also, I would love the tests to include some non-UTC timezone examples as a sanity check.

timvisee:
Some comments on your open questions, but I haven't gone through your code yet:

> Internal representation. Current format: milliseconds in `i64`. Possible alternatives: nanoseconds in `u64` or seconds in `f64`.

I think this is a good choice. This way we have a constant resolution of 1 millisecond, while we'd loose precision with f64.

I do wonder whether we'd somehow want to make this timezone aware as well. Maybe we shouldn't care for the index and always use UTC.

> Timestamp representation. Current format: RFC3339, e.g., `2019-01-01T00:00:00.000Z`.

Perfect. Following a solid standard.

lib/api/build.rs
timvisee: We likely have a more limited range of supported times due to us fitting milliseconds in i64.

xzfc:
@timvisee
> Though, what's the maximum value we can represent here if milliseconds and i64?

I've checked the supported datetime ranges for chrono library and i64 values.

| unit | range in years |
|-| -: |
| seconds | 584,942,417,355 |
| milliseconds | 584,942,417 |
| microseconds |584,942 |
| nanoseconds | 584 |

- Microseconds in `i64` is enough to cover the whole range of dates supported by `chrono` (~524288 years). So if we stick to `chrono`, no reason to use milliseconds instead of microseconds.
- Nanoseconds in `i64` would give us a range of about 584 years, from 1677 to 2262.
- Chrono doesn't support sub-nanosecond accuracy.
- Seconds in `f64` would give us sub-microsecond precision for dates in the current year, and sub-millisecond precision for dates in the max year supported by chrono (262143). <details><summary>calculations</summary><pre><code>$ python3 -c 'import math,time;t=time.time();print(f"{(math.nextafter(t, t+1)-t) * 1e6:.10f} µs")'
0.2384185791 µs
$ python3 -c 'import math,time;t=8210298412799.0;print(f"{(math.nextafter(t, t+1)-t) * 1e6:.10f} µs")'
976.5625000000 µs</code></pre> </details>

I'll switch to microseconds for now if you don't have objections.

lib/api/build.rs
xzfc: It's possible through field annotations, and I'll do it. I've just followed the example of `SnapshotDescription.creation_time` which also uses `#[serde(skip)]`.

lib/segment/src/index/field_index/field_index_base.rs
xzfc: It's used here to select the appropriate parser. (for `DatetimeIndex` we should parse strings, and for `IntIndex` we should get index).
https://github.com/qdrant/qdrant/blob/c5bc3aef6619bb95b5cd1f61ca20dbd2e345008a/lib/segment/src/index/field_index/field_index_base.rs#L270-L275

I see three ways to handle this:
1. Add a new enum variant `FieldIndex::DatetimeIndex`. (initial implementation of this PR)
2. Use numeric values and strings interchangeably. I think it's a too loose way.
3. Select parser using both `FieldIndex` and `PayloadFieldSchema`. I've updated the PR to pass `PayloadFieldSchema` for this. (commits `Introduce IndexMapItem` and `Drop FieldIndex::DatetimeIndex` )

timvisee:
@xzfc Thank you so much for diving into this.

> I'll switch to microseconds for now if you don't have objections.

No objections, we can extend/change this later if desired.

docs/redoc/master/openapi.json
```diff
@@ -6494,7 +6506,7 @@
           }
         }
       },
-      "Range": {
+      "Range_for_double": {
```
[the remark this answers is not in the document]
xzfc: Rolled back.

lib/api/build.rs
xzfc: Added validation and serializer using `prost_wkt_types` crate.

openapi/tests/openapi_integration/test_payload_indexing.py
xzfc: Done.

lib/segment/src/index/field_index/field_index_base.rs
coszio: I liked solution #1 better. In this case it makes total sense for me to add the variant. Sorry for missing it :D

lib/segment/src/types.rs
```diff
@@ -1322,21 +1324,35 @@ impl From<Vec<IntPayloadType>> for MatchExcept {
 }
 
 /// Range filter request
-#[derive(Debug, Deserialize, Serialize, JsonSchema, Default, Clone, PartialEq)]
+#[macro_rules_attribute::macro_rules_derive(crate::common::macros::schemars_rename_generics)]
+#[derive_args(<FloatPayloadType> => "Range", <DateTimePayloadType> => "DatetimeRange")]
```
coszio, line 1328: 🧙

coszio (approved):
Thanks for improving this @xzfc, looks much better now.

I made a few improvements to push this to the endline:
- restored the enum variant approach
- other minor stuff

I'm not sure how much I like approving macros, which are hard to maintain, but we can refactor that later.

openapi/tests/openapi_integration/test_payload_indexing.py
```diff
@@ -134,6 +134,44 @@ def test_boolean_index():
...
+            api="/collections/{collection_name}/points/scroll",
+            method="POST",
+            path_params={"collection_name": collection_name},
+            body={
+                "with_vector": False,
+                "filter": {"must": [{"key": datetime_key, "datetime_range": range_}]},
+            },
+        )
+        assert response.ok
+        assert all(p["id"] in expected for p in response.json()["result"]["points"])
```
xzfc, line 173: This assertion wouldn't fail even if `/scroll` returned no points.

coszio: That's true, it should be the other way around

lib/api/Cargo.toml
```diff
@@ -14,7 +14,7 @@ tracing = ["dep:tracing", "segment/tracing"]
 [dependencies]
 tonic = { version = "0.9.2", features = ["gzip", "tls"] }
 prost = "0.11.9"
-prost-types = "0.11.9"
+prost-wkt-types = "0.4.2"
```
timvisee, line 17: Do we strictly require this? Would it be possible to implement it in a different way somehow?

We already have some problems with `prost` being slow to update, causing a bunch of other problems. I imagine this to only take longer when using a dependency like this.

lib/api/src/grpc/conversions.rs
```diff
@@ -1210,13 +1237,32 @@ impl From<HnswConfigDiff> for segment::types::HnswConfig {
...
+) -> chrono::DateTime<chrono::Utc> {
+    chrono::Utc.from_utc_datetime(
+        &chrono::NaiveDateTime::from_timestamp_opt(
+            date_time.seconds,
+            date_time.nanos.try_into().unwrap_or(0),
+        )
+        .unwrap_or_else(|| {
+            log::warn!("Failed to decode {date_time}, fallback to UNIX_EPOCH");
+            NaiveDateTime::UNIX_EPOCH
+        }),
```
timvisee, line 1262: Instead of _silently_ failing/falling back here, I think it would be better to return and propagate an error.

lib/segment/src/index/field_index/index_selector.rs
```diff
@@ -49,6 +49,11 @@ pub fn index_selector(
                 field,
             ))],
             PayloadSchemaType::Bool => vec![FieldIndex::BinaryIndex(BinaryIndex::new(db, field))],
+            PayloadSchemaType::Datetime => {
+                vec![FieldIndex::DatetimeIndex(
+                    NumericIndex::<IntPayloadType>::new(db, field, is_appendable),
+                )]
```
timvisee, line 55: What about another index here based on `FieldIndex::IntMapIndex(MapIndex::new(...))`?

I think we need it to properly support direct lookups. Or do we not support direct lookups on dates anyway?

timvisee (changes requested):

lib/api/Cargo.toml
xzfc: I did switch to `prost_wkt_types::Timestamp` because of validator and serde support, as requested in https://github.com/qdrant/qdrant/pull/3395#discussion_r1452756856.

It's not possible to solve it using annotations over `prost_types::Timestamp` due to [this issue](https://github.com/Keats/validator/issues/285). However, according to a maintainer, it will be fixed in next versions.

---

> We already have some problems with prost being slow to update

Could you elaborate, please? BTW, I did use old version of `prost-wkt-types` because it's compatible with an old version of `prost` that we use.

timvisee: Thanks for elaborating and linking to the sources.

I'd be fine to continue with it then to not make this more difficult. I guess we can switch back to regular `prost` once the necessary bits are in place.

> Could you elaborate, please? BTW, I did use old version of `prost-wkt-types` because it's compatible with an old version of `prost` that we use.

Yes. On second look, it's not actually prost being the problem here, but other dependencies around it that we use. Still, I'd prefer to minimize our dependency hell around prost. 
For example, see the comments here: https://github.com/qdrant/qdrant/pull/2589

But, lets leave it like this for now :+1:

lib/segment/src/index/field_index/index_selector.rs
xzfc: Only the range query for dates is supported. I think, `MapIndex` shouldn't be used here?

timvisee: > Only the range query for dates is supported

Then yes, thank you for confirming!

lib/api/src/grpc/conversions.rs
timvisee: For me this seems like the only remaining thing. After which I'm very happy to approve! :heavy_check_mark:

coszio: Good catch! Just pushed this propagation

timvisee (approved):
@xzfc Thanks a lot for your effort on implementing this! :heavy_check_mark:

PR #3403, opened 2024-01-15
Use Absolute Qdrant Directory in Debian Package

### LINKED ISSUE #3370: Debian package should use absolute Qdrant directory
In <https://github.com/qdrant/qdrant/pull/3344>, we've introduced a Debian package for Qdrant. It installs `qdrant` on your system.

Starting `qdrant` will use the current working directory for configuration, storage and snapshots. This can be problematic.

Instead, Qdrant should use a absolute path that, just like similar tools. That way we ensure that Qdrant files are always in the same place, no matter from what working directory we start it from.

We should store everything in `/var/lib/qdrant` and `/etc/qdrant` by default, so that we get:
- `/var/lib/qdrant/storage`
- `/var/lib/qdrant/snapshots`
- `/var/lib/qdrant/static` (for web UI [files](https://github.com/qdrant/qdrant/blob/master/tools/sync-web-ui.sh))
- `/etc/qdrant/config.yaml`

Other suggestions are welcome as well.

**Additional context**
Mimicking defaults other similar tools have is likely a good idea here.

Snehil-Shah:
@timvisee I am a bit confused but shouldn't just changing the paths in config.yaml change the paths to absolute directory? What exactly am I overlooking here?

[Release](https://github.com/Snehil-Shah/qdrant/releases/tag/abs)

closes #3370

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -23,13 +23,17 @@ jobs:
     steps:
       - name: Install dependencies
         run: |
-            sudo apt-get update
-            sudo apt-get install -y gcc-multilib clang cmake protobuf-compiler
-            rustup default stable
-            rustup update
-            rustup show
-            cargo -Vv
+          sudo apt-get update
+          sudo apt-get install -y gcc-multilib clang cmake protobuf-compiler
+          rustup default stable
+          rustup update
+          rustup show
+          cargo -Vv
       - uses: actions/checkout@v4
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v2
+        with:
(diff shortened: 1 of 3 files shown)


timvisee:
@Snehil-Shah The main project should be kept as-is, these paths should only change for the Debian package.

You're right about changing the paths in the configuration, but this should only happen as part of building the Debian package.

To put it differently: we want to keep these files unchanged. When we build the Debian package, it should include these configurations but changed with absolute paths.

Does that resolve the confusion?

I elaborated on this here as well: https://github.com/qdrant/qdrant/issues/3370#issuecomment-1893339235

Snehil-Shah:
Understood, thank you so much!

PR #3411, opened 2024-01-17
fix: Update main dev image tag in ghcr

KShivendu:
This should make it easier to pull the latest dev branch container from Github Container registry in our CI benchmarks

### All Submissions:

### DIFF
--- a/.github/workflows/dev-docker-image-build.yml
+++ b/.github/workflows/dev-docker-image-build.yml
@@ -22,7 +22,7 @@ jobs:
             docker login https://ghcr.io -u qdrant --password ${{ secrets.GITHUB_TOKEN }}
 
             # Build regular image for Github Registry
-            GITHUB_TAG="-t ghcr.io/qdrant/qdrant/qdrant:dev-${{ github.sha }}"
+            GITHUB_TAG="-t ghcr.io/qdrant/qdrant/qdrant:dev-${{ github.sha }} -t ghcr.io/qdrant/qdrant/qdrant:dev"
 
             # Pull, retag and push to GitHub packages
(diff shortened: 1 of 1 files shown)


timvisee (approved):
Nice!

Does this always work, or do we have to enable packages explicitly?
Also, do we need to clean up old images due to size constraints?

agourlay:
Why is this better?

KShivendu:
> Does this always work, or do we have to enable packages explicitly?

Yes. it works. See list of containers created so far https://github.com/qdrant/qdrant/pkgs/container/qdrant%2Fqdrant

> Also, do we need to clean up old images due to size constraints?

We have a new workflow that [deletes ghcr images after a month](https://github.com/qdrant/qdrant/blob/dev/.github/workflows/dev-docker-image-prune.yml). However, it has not been picked up by Github actions so far because it needs to be merged in master branch first.  

> Why is this better?

Context: We are using ghcr to preserve dev containers. This will save us time whenever we want to reproduce some issues with a version. 
Answer: This `dev` tagged container on ghcr will have commit id info included. And it's easier to pull the latest version like this when our continuous benchmarks cron is triggered.

timvisee:
> Context: We are using ghcr to preserve dev containers. This will save us time whenever we want to reproduce some issues with a version. It is also used in our CI benchmarks
> Answer: It's easier to pull the latest version like this when our continuous benchmarks cron is triggered. You just do `docker run ghcr.io/qdrant/qdrant/dev` without having to find the latest container with `dev-*` pattern.

I think the question is, why not push to Docker Hub on which we also have all our other images?

KShivendu:
> I think the question is, why not push to Docker Hub on which we also have all our other images?

Mainly because it will quickly bloat up the bloat our DockerHub profile. We have lots of dev branch merges. Plus, we will have to write custom code for container retention on Dockerhub because by default it deletes containers only after 6 months of inactivity.

agourlay (approved):

PR #3413, opened 2024-01-17
Optimize sparse search by deleting empty postings

agourlay:
This PR optimizes the sparse vector search by removing empty posting lists.

This is based on the insight that real life data sets are [very skewed](https://github.com/qdrant/qdrant/pull/3393) with a large amount of very small posting lists.

By tracking the exact moment when to remove those, we can shorten the length of the loops over `self.postings_iterators` which are very hot.

## Perf. numbers

Using the [sparse-vector-benchmarks](https://github.com/qdrant/sparse-vectors-benchmark) with `python main.py --skip-creation false --dataset small --graph-y-limit 150`

```
DEV
Latency distribution:
50p: 54.86 millis
95p: 85.31 millis
99p: 100.23 millis
999p: 115.24 millis
max: 126.31 millis

PR
Latency distribution:
50p: 47.02 millis
95p: 73.8 millis
99p: 88.48 millis
999p: 105.0 millis
max: 117.82 millis
```

DEV             |  PR
  |

### DIFF
--- a/lib/sparse/src/index/search_context.rs
+++ b/lib/sparse/src/index/search_context.rs
@@ -24,6 +24,7 @@ pub struct SearchContext<'a> {
     is_stopped: &'a AtomicBool,
     result_queue: FixedLengthPriorityQueue<ScoredPointOffset>, // keep the largest elements and peek smallest
     min_record_id: Option<PointOffsetType>, // min record ids across all posting lists
+    contains_empty_posting: bool, // whether the posting list iterators contain exhausted posting lists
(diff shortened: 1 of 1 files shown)


agourlay:
> Is there a good reason for doing this lazily with this boolean flag, versus doing it immediately when a list is emptied?

Good question :+1: 

I tried performing the delete as soon as an empty list is detected bit it hurts performance.

Using the flag also enables to amortize the cost of removal by batching several deletions at once.

Issue #3323, opened 2024-01-04
Simplify REST API nesting


generall:
**Is your feature request related to a problem? Please describe.**

Current structure of filter request might sometimes be too verbose:

```
POST collections/my-col/points/search
{
	"filter": {
		"must": [
			{
                               "key": "test",
				"match": {
					"value": 123
				}
			}
		]
	}
}
```

**Describe the solution you'd like**

Allow it to be like this

```
POST collections/my-col/points/search
{
	"filter": {
		"must": {
                        "key": "test",
			"match": {
				"value": 123
			}
		}
	}
}
```
if there is only one condition.

**Describe alternatives you've considered**

Go even further and 

```
POST collections/my-col/points/search
{
	"filter": {
            "key": "test",
            "match": {
                "value": 123
            }
	}
}
```

**Additional context**

This changes should only be applied to REST API, gRPC should not change.
Changes should be backward compatible, meaning that old clients should still be able to work with new interface.

Other suggestions of how to simplify API schema are welcome.

---

Note for contributors: Please consider this as tracking issue. If you think that it would be beneficial to split the task into multiple smaller PRs, please you are welcome to do so. Bounty will be rewarded for each PR independently

darsto:
@HodaSalim you still working on this? Your post was deleted but you haven't cancelled

darsto:
I implemented the parsing of single JSON objects as single-element arrays for the search Filter, but there's some interesting serde_json funtionality that breaks existing tests. If we decide to do this, we inevitably break some backward compatibility.

Consider this code:
```rust
let query1 = r#"
[
    {
        "key": "docId"
    }
]
"#;
let cond = serde_json::from_str::<IsEmptyCondition>(query1);
println!("{cond:#?}");
```

The JSON describes an array with one element - it could be an empty FieldCondition (without any fields, so it would fail the validation). But the above code parses the whole thing successfully into a single IsEmptyCondition:
```
Ok(
    IsEmptyCondition {
        is_empty: PayloadField {
            key: "docId",
        },
    },
)
```

...even though there wasn't any "is_empty" in the input.

IsEmptyCondition doesn't have #[serde(deny_unknown_fields)], but adding it doesn't change anything.

If I deserialize the same JSON to serde_json::Value, I get:

```
Ok(
    Array [
        Object {
            "key": String("docId"),
        },
    ],
)
```

as expected.
[...]

PR #3414, opened 2024-01-17
Simplify REST API nesting (#3323)

amiremohamadi:
/claim #3323

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -6455,20 +6455,15 @@
             "default": null,
             "anyOf": [
               {
-                "anyOf": [
-                  {
-                    "$ref": "#/components/schemas/Condition"
-                  },
-                  {
-                    "nullable": true
-                  }
-                ]
+                "$ref": "#/components/schemas/Condition"
               },
               {
                 "type": "array",
                 "items": {
                   "$ref": "#/components/schemas/Condition"
-                },
+                }
+              },
+              {
                 "nullable": true
               }
             ]
@@ -6489,20 +6484,15 @@
             "default": null,
             "anyOf": [
               {
-                "anyOf": [
-                  {
-                    "$ref": "#/components/schemas/Condition"
-                  },
-                  {
-                    "nullable": true
-                  }
-                ]
+                "$ref": "#/components/schemas/Condition"
               },
               {
                 "type": "array",
                 "items": {
                   "$ref": "#/components/schemas/Condition"
-                },
+                }
+              },
+              {
                 "nullable": true
               }
             ]
@@ -6512,20 +6502,15 @@
             "default": null,
             "anyOf": [
               {
-                "anyOf": [
-                  {
-                    "$ref": "#/components/schemas/Condition"
-                  },
-                  {
-                    "nullable": true
-                  }
-                ]
+                "$ref": "#/components/schemas/Condition"
               },
               {
                 "type": "array",
                 "items": {
                   "$ref": "#/components/schemas/Condition"
-                },
+                }
+              },
+              {
                 "nullable": true
               }
             ]
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -6452,11 +6452,26 @@
         "properties": {
           "should": {
             "description": "At least one of those conditions should match",
-            "type": "array",
-            "items": {
-              "$ref": "#/components/schemas/Condition"
-            },
-            "nullable": true
+            "default": null,
+            "anyOf": [
+              {
+                "anyOf": [
+                  {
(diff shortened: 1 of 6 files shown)


amiremohamadi:
there are two things that need to be considered:

1. for documentations, models are generated based on the code and internal structures. and since we used a `serde deserializer` for our purpose, the structure remained the same.

2. as @darsto [mentioned](https://github.com/qdrant/qdrant/issues/3323#issuecomment-1880077721) serde has some problems with parsing untagged enums and writing a simple deserializer wouldn't work perfectly (it breaks some of existing tests). with the help of `serde-untagged` (which provides helper `Visitors` for untagged enums) the problem is solved.
but if we don't want to add any dependencies, maybe with some efforts we will be able to implement our serde `Visitor` instead of using `serde-untagged`.

darsto:
Putting all the implementation problems aside, the main concern is that JSON arrays could be already parsed as objects. This seems to be a standard JSON thing (at least every serde-json app works this way) so there's a chance someone uses this format.
- and we're breaking it now

amiremohamadi:
@darsto i think we already solved it by using a different serde `Visitor`. can u provide some examples on how it can break backward compatibility?

darsto:
I had to think a little but it seems you're right! :) I can't come up with anything that's broken now

lib/segment/src/types.rs
```diff
@@ -2009,6 +2012,25 @@ impl Filter {
...
 
+fn deserialize_filter_item<'de, D>(deserializer: D) -> Result<Option<Vec<Condition>>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    use serde_untagged::UntaggedEnumVisitor;
+
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum Helper {
```
darsto, line 2023: this seems unused

docs/redoc/master/openapi.json
```diff
@@ -6187,6 +6187,7 @@
         "properties": {
           "should": {
             "description": "At least one of those conditions should match",
+            "default": null,
             "type": "array",
```
coszio, line 6191: The changes in REST API should be reflected here as well

lib/segment/src/types.rs
```diff
@@ -1954,12 +1954,15 @@ pub struct WithPayload {
...
     pub should: Option<Vec<Condition>>,
     /// All conditions must match
+    #[serde(default, deserialize_with = "deserialize_filter_item")]
     #[validate]
     pub must: Option<Vec<Condition>>,
     /// All conditions must NOT match
+    #[serde(default, deserialize_with = "deserialize_filter_item")]
     #[validate]
     pub must_not: Option<Vec<Condition>>,
 }
```
coszio, line 1968: This works effectively, but it's hard to generate an openapi schema from this

coszio (changes requested):
I like that it is only a matter of a few lines to achieve the deserialization like this, but we need to reflect it also in the OpenApi spec to propagate it to auto-generated clients. 

One alternative is using the untagged enum approach.

lib/segment/src/types.rs
amiremohamadi: @coszio thanks for the review. although using untagged enums might be a clean way of fixing this, it requires a lot of changes in different parts of the code. so how about using `#[schemars(schema_with = "...")]` in order to generate the desired openapi schema for this?

coszio: Sure, let's see how that looks. In my opinion it looks like this parallel implementation might be harder to maintain. 

On another note, how about making the function generic and renaming to `deserialize_one_or_many`

amiremohamadi:
conflicts are resolved. may I get a review on this?
@coszio 
@generall

generall (changes requested):
I am not quite sure about the current version either. 
It creates a strange-looking OpenAPI "schema" 

image

Which is quite likely to be a problem for generated clients

amiremohamadi:
> Which is quite likely to be a problem for generated clients

@generall  thanks for your review.

i think we used a similar schema for `SparseVectorParams` as well. however it's possible to generate sth like this, by some minor changes:



what's your opinion?

amiremohamadi:
ping @generall @coszio

coszio:
@amiremohamadi Unfortunately adding this `OneOrMany_for_Condition` will also be problematic with client generation. However, with a few tweaks your implementation can be fixed for a simpler schema.

I'll push them here, I hope you don't mind. I've also rebased this branch on top of current dev.

coszio:
New look: 
image

generall (approved):

PR #3420, opened 2024-01-18
Non-blocking snapshots

timvisee:
Non-blocking collection and segment snapshots, by utilizing segment proxies.

While it already significantly decreases the likeliness of blocking significantly, it may not be fully gone yet.

I've added a few segment holder functions to make this happen in a controlled manner:
- `proxy_all_segments_and_apply`: proxy all segments, run a closure on them, and unproxy all
  - `proxy_all_segments`: proxy all segments
  - `unproxy_all_segments`: unproxy all segments and propagate changes

Snapshotting now simply calls `proxy_all_segments_and_apply` to snapshot each segment individually in a non-blocking manner.

The consensus tests asserts data consistency when upsertions are happening while snapshots are made. It helps detecting unsoundness with proxying and propagating changes because this is error prone.

I cannot think of a good test for asserting upsertions are not blocked. For that we need a huge collection/cluster, which is not suitable for CI.

### Problems
- [x]  Some hick-ups in implementation with concurrent `bfb` inserts, possibly due to saturated disk
   - I could not reproduce this 2 months later
- [x] Non-blocking snapshots and optimizers conflict on segment holder (exclusive locking attempt in [`non-blocking-snapshots-locked`](https://github.com/qdrant/qdrant/tree/non-blocking-snapshots-locked))
  - Fixed by using a single shared lock we upgrade/downgrade, during the whole snapshot process no other write locks can be taken on the segment holder.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -173,6 +173,7 @@ impl ProxySegment {
         let write_segment = self.write_segment.get();
         let mut write_segment = write_segment.write();
 
+        // TODO: define proper op_num here!
         let op_num = 0;
 
         // Propagate deleted points
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -172,7 +172,7 @@ impl ProxySegment {
     pub(super) fn propagate_to_wrapped(&self) -> OperationResult<()> {
         let wrapped_segment = self.wrapped_segment.get();
         let mut wrapped_segment = wrapped_segment.write();
-        let op_num = wrapped_segment.version() + 1;
+        let op_num = wrapped_segment.version();
 
         // Propagate deleted points
         {
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -179,29 +179,29 @@ impl ProxySegment {
 
         // Propagate deleted points
         {
-            let mut deleted_points = self.deleted_points.write();
+            let deleted_points = self.deleted_points.upgradable_read();
             for point_id in deleted_points.iter() {
                 write_segment.delete_point(op_num, *point_id)?;
             }
-            deleted_points.clear();
+            RwLockUpgradableReadGuard::upgrade(deleted_points).clear();
         }
 
         // Propagate deleted indexes
         {
-            let mut deleted_indexes = self.deleted_indexes.write();
+            let deleted_indexes = self.deleted_indexes.upgradable_read();
             for key in deleted_indexes.iter() {
                 write_segment.delete_field_index(op_num, key)?;
             }
-            deleted_indexes.clear();
+            RwLockUpgradableReadGuard::upgrade(deleted_indexes).clear();
         }
 
         // Propagate created indexes
         {
-            let mut created_indexes = self.created_indexes.write();
+            let created_indexes = self.created_indexes.upgradable_read();
             for (key, field_schema) in created_indexes.iter() {
                 write_segment.create_field_index(op_num, key, Some(field_schema))?;
             }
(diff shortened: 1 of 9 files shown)


timvisee:
Since this does segment holder lock upgrading/downgrading I still want to do more manual testing to be 100% sure this doesn't introduce some sort of deadlock. But other than that it seems to be ready for review.

There are some other improvement ideas in our design document, but that's something we can cover in follow up PRs.

timvisee:
I am confident this PR is ready now. I had some locking issues before, but these have since been resolved.

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -584,20 +593,263 @@ impl<'s> SegmentHolder {
...
+
+        // Replace all segments with proxies
+        // We cannot fail past this point to prevent only having some segments proxified
+        let mut proxies = Vec::with_capacity(new_proxies.len());
+        let mut write_segments = RwLockUpgradableReadGuard::upgrade(segments_lock);
+        for (segment_id, mut proxy) in new_proxies {
+            // Replicate field indexes a the second time, because optimized segments could have
+            // been changed. The probability is small, though, so we can afford this operation
+            // under the full collection write lock
+            let op_num = proxy.version() + 1;
```
generall, line 746: It might not matte in this case, but I would prefer to only increase op_num in WAL. This should guarantee that version number won't accumulate independently from the WAL, therefore we would guarantee that new upcomming operations always have newer version number

lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -158,6 +158,57 @@ impl ProxySegment {
...
+    /// - deleted payload indexes
+    /// - created payload indexes
+    ///
+    /// This is required if making both the wrapped segment and the writable segment available in a
+    /// shard holder at the same time. If the wrapped segment is thrown away, then this is not
+    /// required.
+    pub(super) fn propagate_to_wrapped(&self) -> OperationResult<()> {
+        let wrapped_segment = self.wrapped_segment.get();
+        let mut wrapped_segment = wrapped_segment.write();
+        let op_num = wrapped_segment.version() + 1;
```
generall, line 175: same, let's just use `wrapped_segment.version()`, it should be enough to apply operation and it is used like this in optimizer

skeptrunedev:
Ah man, we are incredibly excited about this PR!!!

generall:
Manual tests on my laptop, by running

```
docker run --rm -it --network=host  qdrant/bfb:dev ./bfb -n 1000000
```

and making snapshots in parallel, looks like this:

dev:

```
Slow upsert: 0.483429769
Slow upsert: 1.147807834
Slow upsert: 0.678137794
```

PR:

```
Slow upsert: 0.137086712
Slow upsert: 0.120869729
Slow upsert: 0.103555548
```

generall (approved):

lib/collection/src/collection_manager/holders/segment_holder.rs
timvisee: Good suggestion, I'll remove the +1's.

timvisee: I see you already did so. Thanks!

timvisee:
> Manual tests on my laptop, by running
> 
> ```
> docker run --rm -it --network=host  qdrant/bfb:dev ./bfb -n 1000000
> ```
> 
> and making snapshots in parallel, looks like this:
> 
> dev:
> 
> ```
> Slow upsert: 0.483429769
> Slow upsert: 1.147807834
> Slow upsert: 0.678137794
> ```
> 
> PR:
> 
> ```
> Slow upsert: 0.137086712
> Slow upsert: 0.120869729
> Slow upsert: 0.103555548
> ```

Thanks for giving it another test. On my machine I don't see any slowdowns after this PR. With more data the difference seems to become greater.

ffuugoo (approved):

PR #3423, opened 2024-01-18
fixes gRPC API return status code.

### LINKED ISSUE #3217: gRPC API should return AlreadyExists status code rather than InvalidArgument if collection already exists
When you try to create a collection that already exists the gRPC API returns status with `tonic::Code::InvalidArgument` code variant and I think it should return `tonic::Code::AlreadyExists`.

`tonic::Code::AlreadyExists` code is returned instead of `tonic::Code::InvalidArgument` when trying to create a collection that already exists.

The obvious alternative is using the current implementation, but I think that it is somewhat misleading to return `InvalidArgument` when the `AlreadyExists` code is also present in `tonic::Code`.

**Additional context**
I didn't dig very deep, but as far as I can tell [this is the place where the error is created](https://github.com/qdrant/qdrant/blob/97b2483660d9190835b8df13ee1b27873d370da5/lib/storage/src/content_manager/collections_ops.rs#L21), and my guess would be that there is some sort of conversion between `StorageError` and `tonic::Status`, and `StorageError` uses `BadInput` variant, as it doesn't have `AlreadyExists` equivalent.

2k4sm:
Fixes  #3217

- This PR fixes the return of AlreadyExists error instead of returning InvalidArgument in case there is an attemp to create a collection that already exists.
- To achieve this i have created a AlreadyExists error type and AlreadyExists function for StorageError enum and changed the return of StorageError from BadInput to AlreadyExists.
- This  uses tonic::Code::AlreadyExists for the returning status code.
- It also adds AlreadyExists error type for actix to handle this.
- This uses HttpResponse::Conflict for building HttpResp.
- It also adds status code and description for HttpError.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/collections_ops.rs
+++ b/lib/storage/src/content_manager/collections_ops.rs
@@ -8,13 +8,13 @@ use crate::content_manager::errors::StorageError;
 pub type Collections = HashMap<CollectionId, Collection>;
 
 pub trait Checker {
-    fn is_collection_exists(&self, collection_name: &str) -> bool;
+    fn collection_exists(&self, collection_name: &str) -> bool;
 
     async fn validate_collection_not_exists(
         &self,
         collection_name: &str,
     ) -> Result<(), StorageError> {
-        if self.is_collection_exists(collection_name) {
+        if self.collection_exists(collection_name) {
             return Err(StorageError::AlreadyExists {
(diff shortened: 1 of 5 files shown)


coszio (approved):
Looks good

PR #3408, opened 2024-01-16
Add "clock_tag" field to update operations in the WAL and to the internal gRPC API

ffuugoo:
This PR adds a `clock_tag` field to update operations stored in WAL and to the internal gRPC API.

This field will be used (in follow-up PRs) to optimize shard transfer: e.g., search relevant position in WAL on a "good" node and only transfer part of WAL that is missing on "dead" node.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #3426, opened 2024-01-19
Assign clock tags to internal update operations (and update operation responses)

generall:

This PR
- add `ClockSet` type to generate clock tags for update operations
- assign clock tags to update operations (and update operation responses)

Currently based on #3408, but will be rebased on `dev` once #3408 is merged.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/replica_set/clock_set.rs
+++ b/lib/collection/src/shards/replica_set/clock_set.rs
@@ -12,7 +12,7 @@ impl ClockSet {
         Self::default()
     }
 
-    /// Get the first available clock from this set.
+    /// Get the first available clock from the set, or create a new one.
     pub fn get_clock(&mut self) -> ClockGuard {
         for (id, clock) in self.clocks.iter().enumerate() {
             if clock.lock() {
@@ -44,19 +44,13 @@ impl ClockGuard {
         self.id
     }
 
+    #[must_use = "new clock value must be used"]
     pub fn tick_once(&mut self) -> u64 {
         self.clock.tick_once()
     }
 
     pub fn advance_to(&mut self, new_tick: u64) -> u64 {
         self.clock.advance_to(new_tick)
-    }
-
-    pub fn release(self) {
-        // Do not call `self.clock.release()` here!
-        //
-        // `Drop` trait will automatically trigger `self.clock.release()`, when `self` goes out of
-        // scope at the end of the method.
     }
 }
 
@@ -73,27 +67,28 @@ struct Clock {
 }
 
 impl Clock {
-    pub fn new_locked() -> Self {
+    fn new_locked() -> Self {
         Self {
             clock: AtomicU64::new(0),
             available: AtomicBool::new(false),
         }
     }
 
-    pub fn tick_once(&self) -> u64 {
+    #[must_use = "new clock value must be used"]
+    fn tick_once(&self) -> u64 {
         self.clock.fetch_add(1, Ordering::Relaxed) + 1
     }
 
-    pub fn advance_to(&self, new_tick: u64) -> u64 {
+    fn advance_to(&self, new_tick: u64) -> u64 {
         let current_tick = self.clock.fetch_max(new_tick, Ordering::Relaxed);
         cmp::max(current_tick, new_tick)
     }
 
-    pub fn lock(&self) -> bool {
+    fn lock(&self) -> bool {
         self.available.swap(false, Ordering::Relaxed)
     }
 
-    pub fn release(&self) {
+    fn release(&self) {
         self.available.store(true, Ordering::Relaxed);
     }
 }
--- a/lib/collection/src/shards/replica_set/clock_set.rs
+++ b/lib/collection/src/shards/replica_set/clock_set.rs
@@ -1,0 +1,91 @@
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::Arc;
+
+#[derive(Clone, Debug, Default)]
+pub struct ClockSet {
+    clocks: Vec<Arc<Clock>>,
+}
+
+impl ClockSet {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn get_clock(&mut self) -> ClockGuard {
+        for (id, clock) in self.clocks.iter().enumerate() {
+            if clock.lock() {
+                return ClockGuard::new(id, clock.clone());
+            }
+        }
+
+        let id = self.clocks.len();
+        let clock = Arc::new(Clock::new_locked());
+
+        self.clocks.push(clock.clone());
+
+        ClockGuard::new(id, clock)
+    }
+}
+
+#[derive(Debug)]
+pub struct ClockGuard {
+    id: usize,
+    clock: Arc<Clock>,
+}
+
+impl ClockGuard {
+    fn new(id: usize, clock: Arc<Clock>) -> Self {
+        Self { id, clock }
+    }
+
+    pub fn id(&self) -> usize {
+        self.id
+    }
+
+    pub fn tick_once(&mut self) -> u64 {
+        self.clock.tick_once()
+    }
+
+    pub fn release(self) {
+        // Do not call `Clock::release` explicitly!
+        //
+        // `Drop` will be called when `self` goes out of scope at the end of this method
+        // and it will call `Clock::release`.
+        //
+        // If call `Clock::release` explicitly, then `ClockGuard::release` will call
+        // `Clock::release` *twice*! (Which is a bug! :D)
+    }
+}
+
+impl Drop for ClockGuard {
+    fn drop(&mut self) {
+        self.clock.release();
+    }
+}
+
+#[derive(Debug)]
+struct Clock {
+    clock: AtomicU64,
+    available: AtomicBool,
+}
+
+impl Clock {
+    pub fn new_locked() -> Self {
+        Self {
+            clock: AtomicU64::new(0),
+            available: AtomicBool::new(false),
+        }
+    }
+
+    pub fn tick_once(&self) -> u64 {
+        self.clock.fetch_add(1, Ordering::Relaxed) + 1
+    }
+
+    pub fn lock(&self) -> bool {
+        self.available.swap(false, Ordering::Relaxed)
+    }
+
+    pub fn release(&self) {
+        self.available.store(true, Ordering::Relaxed);
+    }
+}
--- a/lib/collection/src/shards/replica_set/clock_set.rs
+++ b/lib/collection/src/shards/replica_set/clock_set.rs
@@ -46,6 +46,10 @@ impl ClockGuard {
         self.clock.tick_once()
     }
 
+    pub fn advance_to(&mut self, tick: u64) -> u64 {
+        self.clock.advance_to(tick)
+    }
+
     pub fn release(self) {
         // Do not call `Clock::release` explicitly!
         //
@@ -79,6 +83,10 @@ impl Clock {
 
     pub fn tick_once(&self) -> u64 {
         self.clock.fetch_add(1, Ordering::Relaxed) + 1
+    }
+
+    pub fn advance_to(&self, tick: u64) -> u64 {
+        self.clock.fetch_max(tick, Ordering::Relaxed)
     }
 
     pub fn lock(&self) -> bool {
--- a/lib/collection/src/shards/replica_set/clock_set.rs
+++ b/lib/collection/src/shards/replica_set/clock_set.rs
@@ -1,3 +1,4 @@
+use std::cmp;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
 
@@ -46,18 +47,15 @@ impl ClockGuard {
         self.clock.tick_once()
     }
 
-    pub fn advance_to(&mut self, tick: u64) -> u64 {
-        self.clock.advance_to(tick)
+    pub fn advance_to(&mut self, new_tick: u64) -> u64 {
+        self.clock.advance_to(new_tick)
     }
 
     pub fn release(self) {
-        // Do not call `Clock::release` explicitly!
+        // Do not call `self.clock.release()` here!
         //
-        // `Drop` will be called when `self` goes out of scope at the end of this method
-        // and it will call `Clock::release`.
-        //
-        // If call `Clock::release` explicitly, then `ClockGuard::release` will call
-        // `Clock::release` *twice*! (Which is a bug! :D)
+        // `Drop` trait will automatically trigger `self.clock.release()`, when `self` goes out of
+        // scope at the end of the method.
     }
 }
 
@@ -85,8 +83,9 @@ impl Clock {
         self.clock.fetch_add(1, Ordering::Relaxed) + 1
     }
 
-    pub fn advance_to(&self, tick: u64) -> u64 {
-        self.clock.fetch_max(tick, Ordering::Relaxed)
+    pub fn advance_to(&self, new_tick: u64) -> u64 {
+        let current_tick = self.clock.fetch_max(new_tick, Ordering::Relaxed);
+        cmp::max(current_tick, new_tick)
     }
 
     pub fn lock(&self) -> bool {
--- a/lib/collection/src/shards/replica_set/clock_set.rs
+++ b/lib/collection/src/shards/replica_set/clock_set.rs
@@ -12,6 +12,7 @@ impl ClockSet {
         Self::default()
     }
 
+    /// Get the first available clock from this set.
     pub fn get_clock(&mut self) -> ClockGuard {
         for (id, clock) in self.clocks.iter().enumerate() {
             if clock.lock() {
--- a/lib/collection/src/shards/replica_set/clock_set.rs
+++ b/lib/collection/src/shards/replica_set/clock_set.rs
@@ -71,8 +71,8 @@ struct Clock {
 impl Clock {
     fn new_locked() -> Self {
         Self {
-            next_tick: AtomicU64::new(0),
-            available: AtomicBool::new(false),
+            next_tick: 0.into(),
+            available: false.into(),
         }
     }
 
--- a/lib/collection/src/shards/replica_set/clock_set.rs
+++ b/lib/collection/src/shards/replica_set/clock_set.rs
@@ -108,7 +108,8 @@ impl Clock {
 
     /// Try to acquire exclusive lock over this clock.
     ///
-    /// Returns `true` if the lock was successfuly acquired, or `false` if the clock is already locked.
+    /// Returns `true` if the lock was successfully acquired, or `false` if the clock is already
+    /// locked.
     fn try_lock(&self) -> bool {
         self.available.swap(false, Ordering::Relaxed)
     }
--- a/lib/collection/src/shards/replica_set/clock_set.rs
+++ b/lib/collection/src/shards/replica_set/clock_set.rs
@@ -1,4 +1,3 @@
-use std::cmp;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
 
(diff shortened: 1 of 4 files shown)


lib/collection/src/collection/sharding_keys.rs
```diff
@@ -105,7 +107,12 @@ impl Collection {
                     }),
                 );
 
-                replica_set.update_local(create_index_op, true).await?;
+                // ToDo[vector-clock]: Is this correct?
+                let clock_sync = ClockSync::new(tick, 0, 0);
+                replica_set
+                    .update_local(create_index_op, true, Some(clock_sync))
+                    .await?;
+                tick += 1;
```
ffuugoo, line 115: Ditto regarding `update_all_local`. I don't think `ClockSync::new(tick, 0, 0)` makes sense here. 🤔

lib/collection/src/shards/replica_set/vector_clock.rs
```diff
@@ -0,0 +1,81 @@
...
+            availability,
+        }
+    }
+
+    pub fn get_id(&self) -> usize {
+        self.id
+    }
+
+    pub fn resolve(self) {
+        self.availability.swap(false, Ordering::Relaxed);
```
ffuugoo, line 32: 🤔

```suggestion
        self.availability.swap(true, Ordering::Relaxed);
```

lib/collection/src/shards/replica_set/vector_clock.rs
```diff
@@ -0,0 +1,81 @@
...
+            thread_id: self.id as u64,
+        }
+    }
+
+    pub fn increment(&self) {
+        self.counter.fetch_add(1, Ordering::Relaxed);
+    }
+
+    fn release(&mut self) {
+        self.availability.swap(false, Ordering::Relaxed);
```
ffuugoo, line 48: 🤔

```suggestion
        self.availability.swap(true, Ordering::Relaxed);
```

lib/collection/src/shards/replica_set/vector_clock.rs
```diff
@@ -0,0 +1,81 @@
...
+        Self {
+            counters: Vec::new(),
+            availabilities: Vec::new(),
+        }
+    }
+
+    pub fn occupy(&mut self) -> ClockGuard {
+        for (id, availability) in self.availabilities.iter().enumerate() {
+            if availability.swap(false, Ordering::SeqCst) {
+                self.counters[id].fetch_add(1, Ordering::SeqCst);
```
ffuugoo, line 69: `Ordering::Relaxed`? (I'm not 100% sure, though...)

lib/collection/src/shards/replica_set/vector_clock.rs
```diff
@@ -0,0 +1,81 @@
...
+        self.availability.swap(false, Ordering::Relaxed);
+    }
+
+    pub fn get_clock_sync(&self, peer_id: PeerId) -> ClockSync {
+        ClockSync {
+            tick: self.counter.load(Ordering::Relaxed),
+            peer_id,
+            thread_id: self.id as u64,
+        }
+    }
```
ffuugoo, line 41: I think `peer_id` should be set once during `VectorClock` construction. Way too easy to accidentally pass wrong `peer_id` into `get_clock_sync`.

lib/collection/src/shards/replica_set/shard_transfer.rs
```diff
@@ -319,7 +330,15 @@ impl ShardReplicaSet {
             proxy.remote_shard.peer_id,
         );
 
-        proxy.transfer_indexes().await
+        let peer_id = self.this_peer_id();
+        let vector_clock_guard = self.vector_clock.lock().await.occupy();
+
+        proxy
+            .transfer_indexes(|| {
+                vector_clock_guard.increment();
```
ffuugoo, line 338: Do we need an increment here, if `occupy` does `fetch_add` already?

ffuugoo (commented):
I think most of the PR make sense, I only have minor-ish comments.

ffuugoo:
Rebased on top of #3408 and added TODO to PR description.

ffuugoo:
@generall can't add you as reviewer (cause you opened the PR itself), but please review. 😂🙏

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -130,80 +131,80 @@ impl ShardReplicaSet {
...
-                        local
-                            .get()
-                            .update(operation, local_wait)
-                            .await
-                            .map(|ok| (this_peer_id, ok))
-                            .map_err(|err| (this_peer_id, err))
-                    };
+        let current_clock_tick = clock.tick_once();
+        let clock_tag = ClockTag::new(this_peer_id, clock.id() as _, current_clock_tick);
+        let operation = OperationWithClockTag::new(operation, Some(clock_tag));
```
timvisee, line 158: What about an `Operation::with_clock_tag(clock_tag)` function that constructs the `OperationWithClockTag` for us?

Just a style change, but I guess it may be a nice one.

lib/collection/src/shards/replica_set/clock_set.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    pub fn release(self) {
+        // Do not call `Clock::release` explicitly!
+        //
+        // `Drop` will be called when `self` goes out of scope at the end of this method
+        // and it will call `Clock::release`.
+        //
+        // If call `Clock::release` explicitly, then `ClockGuard::release` will call
+        // `Clock::release` *twice*! (Which is a bug! :D)
+    }
```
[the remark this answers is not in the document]
ffuugoo: But maybe I worded it awkwardly? Open to suggestions.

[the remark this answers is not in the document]
timvisee: Is there a good reason for ever calling `release` manually versus not dropping right away?

If there isn't, wouldn't it just be safer to disallow calling this?

lib/collection/src/shards/replica_set/update.rs
timvisee: I did implement this now. I also removed

timvisee: I did implement this now. I also removed `From<Operation>` implementation which silently wrapped the type without clock tag, which may go unnoticed and be undesirable in some cases.

This changes makes it more explicit, and shorter. :smile:

timvisee: I did implement this now. I also removed `From<Operation>` implementation which silently wrapped the type without clock tag, which may go unnoticed and be undesirable in some cases.

This changes makes it more explicit, and shorter. :smile:

lib/collection/src/shards/replica_set/clock_set.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    #[must_use = "new clock value must be used"]
+    fn tick_once(&self) -> u64 {
+        self.clock.fetch_add(1, Ordering::Relaxed)
+    }
+
+    fn advance_to(&self, new_tick: u64) -> u64 {
+        let current_tick = self.clock.fetch_max(new_tick + 1, Ordering::Relaxed);
+        cmp::max(current_tick, new_tick + 1)
+    }
```
ffuugoo, line 85: Ok, so, after discussion this morning, I've updated `Clock` so that it will start ticking from `0` instead of `1`... but, can you guys please help me analyze if I've updated `advance_to` correctly? 🥲🙏

Initially
- `tick_once` was `fetch_add(1) + 1`
- and `advance_to` was `fetch_max(new_tick)`

But now
- `tick_once` does `fetch_add(1)`
  - so it _does_ increment, but returns the value _before_ increment
  - so if `Clock` starts from `0`, then `tick_once` will return `0` and store `1`
  - then the next `tick_once` will return `1` and store `2`
  - etc...
- so, with that in mind, _I think_, it makes sense to change `advance_to` to `fetch_max(new_tick + 1)`
  - e.g., we restarted a node, the clock is `0`, we send first request and receive response with clock `10`
  - then we have to do `fetch_max(10 + 1)`
  - so that _the next_ `tick_once` will return `11` (and store `12`)
  - this, _kinda_, depends on how exactly we calculate the clock in update operation response
  - e.g., we could return "current version" or "current version + 1"
  - but it seems easier to return "current version"
  - and then we need to do `fetch_max(new_tick + 1)`
  - _I think_

lib/collection/src/shards/replica_set/clock_set.rs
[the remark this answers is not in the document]
ffuugoo: Sorry, I reverted most of this (cause, IMO, my version is easier to read/understand) with slight modifications:
- I kept `must_use`
- and I removed `pub` from `Clock` methods

lib/collection/src/shards/replica_set/update.rs
ffuugoo: I reverted this as well. Seems a bit too much to add and export a trait just to construct `OperationWithClockTag`.

I've also tried adding explicit methods `OperationWithClockTag::with_tag`/`OperationWithClockTag::without_tag`, but, IMO, these are also annoying to use. :/

For now, I simply reverted to using `From`, but if you're strongly against it we can add `with_tag`/`without_tag`.

lib/collection/src/shards/replica_set/clock_set.rs
ffuugoo: @generall @timvisee

lib/collection/src/shards/replica_set/clock_set.rs
```diff
@@ -0,0 +1,94 @@
...
+            available: AtomicBool::new(false),
+        }
+    }
+
+    #[must_use = "new clock value must be used"]
+    fn tick_once(&self) -> u64 {
+        self.clock.fetch_add(1, Ordering::Relaxed)
+    }
+
+    fn advance_to(&self, new_tick: u64) -> u64 {
```
generall, line 82: If I understood correctly, the main purpose of this function is to make sure, that no other thread will get ticks equal or less than `new_tick`. Return value might just create a confusion here, so I would propose to not return anything from the function

lib/collection/src/shards/replica_set/clock_set.rs
```diff
@@ -62,26 +62,29 @@ impl Drop for ClockGuard {
...
-        self.clock.fetch_add(1, Ordering::Relaxed)
+        // Fetch the next clock tick to return, and increment it by 1 after
+        self.next_tick.fetch_add(1, Ordering::Relaxed)
     }
 
     fn advance_to(&self, new_tick: u64) -> u64 {
-        let current_tick = self.clock.fetch_max(new_tick + 1, Ordering::Relaxed);
-        cmp::max(current_tick, new_tick + 1)
+        let next_tick = self.next_tick.fetch_max(new_tick + 1, Ordering::Relaxed);
+        cmp::max(next_tick, new_tick + 1)
```
ffuugoo, line 87: This does not make sense now. 😂

timvisee: According to the above comment we might actually want to remove any returns here. Otherwise please feel free to suggest something better :smile:

ffuugoo: Because when doing any `fetch_*` on `next_tick`, it still returns `current_tick` (i.e., previous value).

timvisee: > Because when doing any fetch_* on self.next_tick, it still returns current_tick (i.e., previous value).

Yeah, but the previous value represents the next tick.

lib/collection/src/shards/replica_set/clock_set.rs
timvisee: If we want to return something here because we need in in the future, then I strongly recommend to rename this function to clarify what it actually returns.

Maybe something like `advance_to_get_next_tick`

ffuugoo: It returns current value of the clock (i.e., either `new_tick`, or whatever is in the clock, if it's already "newer" than `new_tick`). I don't use it anywhere ATM, but I was using it before at some point.

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -130,80 +131,80 @@ impl ShardReplicaSet {
...
+                    .update(operation, wait)
+                    .await
+                    .map(|ok| (remote.peer_id, ok))
+                    .map_err(|err| (remote.peer_id, err))
+            };
+
+            update_futures.push(remote_update.right_future());
+        }
 
-                None => FuturesUnordered::from_iter(update_futures).collect().await,
```
generall, line 205: It feels like changes in this part of code are not strictly required for the PR.
If it is a refactoring, could we please keep it in separate PRs next time, otherwise it is quite hard to review and make sure new changes + refactoring don't break anything

ffuugoo: This is _kinda_ needed, cause everything was wrapped in a `{}` block before, but now we need to access `clock` (to update the clock during response handling), and so it has to be out of the `{}` block, so I had to restructure it.

ffuugoo: And if we just put `let mut clock = self.clock_set.lock().await.get_clock();` outside the `{}` block, then it would be locked before other locks and which may generate too many `clock_id`s. :/

lib/collection/src/shards/replica_set/clock_set.rs
ffuugoo: > `advance_and_get_next_tick`

I tried to do something like this, but this also doesn't make much sense. Like, it returns "next tick"... but then the next call to `tick_once` returns the same tick? I agree, that it's just easier to not return anything.

timvisee: > I agree, that it's just easier to not return anything.

We can add it back later if we suddenly need it :+1:

lib/collection/src/shards/replica_set/update.rs
timvisee: As a side effect of this change, we're now not dropping the `local` and `remotes` in time. We should probably drop both explicitly after `let all_res = ...` to prevent holding on to them longer than we need.

generall: ok, then we need to add `drop` to simulate prev implementation (commited)

ffuugoo: > As a side effect of this change, we're now not dropping the local and remotes in time...

Right. That's why I put them into a block in the first place. 🤦‍♀️

timvisee: I think this is resolved now, so, marking it as such. :white_check_mark:

ffuugoo:
I have another question: do we want additional safeguard around tick `0`?

Like, right now if you restart a shard and the first request (with `clock_tick = 0`) fails somehow (so it does not get updated `clock_tick` from other node), the next request would use `clock_tick = 1`. It would still correct itself, but the operation would be stored with `clock_tick = 1` on the other node.

We could add some special handling around `0`, but it will probably be rather ugly. :/

lib/collection/src/shards/replica_set/clock_set.rs
```diff
@@ -0,0 +1,112 @@
...
+        let current_tick = self.next_tick.fetch_add(1, Ordering::Relaxed);
+
+        // If `current_tick` is `0`, then "revert" `next_tick` back to `0`.
+        // We expect that `advance_to` would be used to advance "past" the initial `0`.
+        //
+        // Executing multiple atomic operations sequentially is not strictly thread-safe,
+        // but we expect that `Clock` would be "locked" before calling `tick_once`.
+        if current_tick == 0 {
+            self.next_tick.store(0, Ordering::Relaxed);
+        }
```
ffuugoo, line 93: > I have another question: do we want additional safeguard around tick 0?
> ...
> We could add some special handling around 0, but it will probably be rather ugly. :/

Disregard that. I've figured out a *very elegant* solution to this.

So, the idea is that `tick_once` only increment `next_tick` if it is already more than `0`. If `next_tick` is `0` we keep returning `0` and don't increment `next_tick`. To "advance" past this "initial `0`" you **have to** use `advance_to`. And because `advance_to` does `new_tick + 1`, even if you call `advance_to(0)` it will still increment `next_tick` to `1` or more.

So, after you restart a node, you have an empty `ClockSet` with all zeroes. Then, *until you've made at least one successful request*, `ClockSet` will stay at `0` (so that operations won't be written into WALs on other nodes with wrong `clock_tick`s).

We always call `advance_to` when collecting/processing responses from other replicas (if response has a `clock_tag`, of course), so once you received the first successful response, the `advance_to` call would update the clock past initial `0` and it will start ticking normally.

ffuugoo: `tick_once` could be implemented in a more thread-safe manner (using `load` and `compare_and_exchange` in a loop), but we explicitly lock the `Clock` anyway, and updating `next_tick` the way I currently implemented it should be more efficient.

ffuugoo:
Another question:
- should we only call `advance_to` if at least `minimal_success_count` replicas returned successful response
- or if `local_shard` is present on the node, and it returned a successful response (i.e., even if the first condition is false)

Also, what about the corner case if remote replicas succeeded, but `local_shard` failed? 🤔

lib/collection/src/shards/replica_set/clock_set.rs
```diff
@@ -0,0 +1,112 @@
...
+        // `Clock` tracks *next* tick, so we increment `next_tick` by 1 and return *previous* value
+        // of `next_tick` (which is *exactly* what `fetch_add(1)` does)
+        let current_tick = self.next_tick.fetch_add(1, Ordering::Relaxed);
+
+        // If `current_tick` is `0`, then "revert" `next_tick` back to `0`.
+        // We expect that `advance_to` would be used to advance "past" the initial `0`.
+        //
+        // Executing multiple atomic operations sequentially is not strictly thread-safe,
+        // but we expect that `Clock` would be "locked" before calling `tick_once`.
+        if current_tick == 0 {
```
timvisee, line 91: Doing a second operation on our atomic, removes the atomic guarantees. :(

Can't we just increment and keep it at 1, and then update our clock after?

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -220,6 +224,33 @@ impl ShardReplicaSet {
...
+
+        let new_clock_tick = successes
+            .iter()
+            .filter_map(|(_, result)| {
+                let echo_tag = result.clock_tag?;
+
+                if echo_tag.peer_id != clock_tag.peer_id {
+                    // TODO: `log::warn`!?
+                    return None;
+                }
```
timvisee, line 237: Is there a good reason to explicitly reject this?

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -220,6 +224,33 @@ impl ShardReplicaSet {
...
+
+                Some(echo_tag.clock_tick)
+            })
+            .max();
+
+        if let Some(new_clock_tick) = new_clock_tick {
+            if new_clock_tick > current_clock_tick || new_clock_tick == 0 {
```
timvisee, line 249: Shouldn't the second part be our old number?

```suggestion
            if new_clock_tick > current_clock_tick || clock_tag.clock_tick == 0 {
```

lib/collection/src/shards/replica_set/clock_set.rs
timvisee: We discussed this: we cannot do this.

If we set it to 1 right away, and we crash, we'd start from 1 after restart. Though, we should still be at 0 if the request never resolved.

Having this if-statement prevents this. :+1:

lib/collection/src/shards/replica_set/update.rs
timvisee: In fact, can't we just totally remove this if-statement, and always `advance_to` the `new_clock_tick` value.

timvisee (approved):

timvisee:
> should we only call `advance_to` if at least `minimal_success_count` replicas returned successful response

To answer this, always advance for now so we don't get operations with duplicate clock tags.

If a 0 operation only succeeded on one node, and that node had an old clock value, we might have a problem. But we can discuss this corner case and how to reject (silently or explicit, retry?) operations later.

generall:
LGTM

lib/collection/src/shards/replica_set/update.rs
ffuugoo: Have to agree. Removed.

lib/collection/src/shards/replica_set/update.rs
ffuugoo: It's more of a future-proofing/fail-safe.

With replication/shard proxies/write ordering it is sometimes very unobvious which code path a request will take, and it seems possible to me, that we might introduce a bug that will assign the wrong clock tag to the `UpdateResponse`?

ffuugoo:
Rebased on `dev`, cleaned up PR history, add some documentation to `clock_set::Clock`. Will merge once CI is green

PR #3435, opened 2024-01-22
Remove unused dependencies

agourlay:
Cleanup unused dependencies using [cargo-machete](https://github.com/bnjbvr/cargo-machete)


generall (approved):
If CI is ok, I am ok

coszio:
I love the name, we could go one step further and include it as part of CI, if it doesn't take much time

agourlay:
@coszio I understand why this would make sense and at the same time I think we reached the point where our CI does too much/too slow.

I am fine trimming dependencies manually once per year :hocho:

PR #3449, opened 2024-01-23
Limit automatic shard transfers

timvisee:
This PR implements limiting automatic shard transfers for shard recovery in a cluster. The limit prevents a node from taking the whole cluster down with a lot of shard transfers.

It now limits the number of incoming and outgoing transfers to 1 each per collection, and is now also configurable.

It is a soft limit. Meaning it only applies to automatically initiated transfers, and not affect user requested transfers.

The tricky part here is that we should also consider transfers that are proposed to consensus, but are not picked up yet. For this I keep track of a list of proposed transfers, where each item has a cooldown of 10 seconds. This means that if consensus fails to pick it up, a spot for another transfer will be released in 10 seconds. If we don't track this separately, a train of many transfers will go through in quick succession. If we don't apply some cooldown time here, we may block following transfers forever.

The configuration looks like this:

```yaml
storage:
  performance:
    # Soft limit for number of incoming shard transfers per collection on this node.
    # Only affects automatic shard transfers for recovery, does not affect user requested transfers.
    # The same value should be used on all nodes in a cluster.
    # If null - unlimited.
    incoming_shard_transfers_limit: 1

    # Soft limit for number of outgoing shard transfers per collection on this node.
    # Only affects automatic shard transfers for recovery, does not affect user requested transfers.
    # The same value should be used on all nodes in a cluster.
    # If null - unlimited.
    outgoing_shard_transfers_limit: 1
```

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -18,18 +18,6 @@ use crate::shards::transfer::{
 };
 use crate::shards::{transfer, CollectionId};
 
-/// Soft limit of incoming shard transfers on a node
-///
-/// To automatically initiate a new shard transfer, we must be below this limit. This is a soft
-/// limit, because we won't reject additional shard transfers requested by the user.
-pub(super) const AUTO_SHARD_TRANSFER_LIMIT_IN: usize = 1;
-
-/// Soft limit of outgoing shard transfers on a node
-///
-/// To automatically initiate a new shard transfer, we must be below this limit. This is a soft
-/// limit, because we won't reject additional shard transfers requested by the user.
-pub(super) const AUTO_SHARD_TRANSFER_LIMIT_OUT: usize = 1;
-
 /// Cooldown for proposed shard transfers, when they are not (yet) accepted by consensus
 ///
 /// After this time, a proposed shard transfer is forgotten when it is not accepted by consensus.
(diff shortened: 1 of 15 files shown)


lib/collection/src/collection/shard_transfer.rs
```diff
@@ -357,4 +373,95 @@ impl Collection {
...
+}
+
+/// A structure tracking ongoing and proposed shard transfers, usually used globally.
+///
+/// This can be used to count the number of shard transfers on some peer to rate-limit automatic
+/// transfers.
+#[derive(Debug, Default)]
+pub struct ShardTransferTracker {
+    /// Ongoing shard transfers.
+    transfers: Vec<ShardTransferKey>,
```
timvisee, line 399: This implementation currently relies on tracking the ongoing shard transfers ourselves.

I implemented the first iteration of this to be global, meaning the configured limits apply to the node across all collections. In that configuration tracking ongoing transfers here was required.

Since, I've changed this to be on collection level. Because of that, it isn't strictly required anymore to track transfers here. We could also rely on the list of transfers in the collection's shard holder.

I've not decided yet whether to replace this with ongoing transfers in the shard holder. It may be somewhat tricky to rework due to locking issues.

What do you think? Keep it as-is, or rework this to rely on our existing list of ongoing transfers?

timvisee:
I've implemented a stateless version of this, which is likely a lot better: <https://github.com/qdrant/qdrant/pull/3458>

I'll therefore draft this one. We can close this if <https://github.com/qdrant/qdrant/pull/3458> gets merged.

timvisee:
Closing this now as I don't think we'll revisit this implementation again.

PR #3454, opened 2024-01-24
chore: sort the denpendencies

ben1009:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

- run [cargo sort -g](https://crates.io/crates/cargo-sort) to sort  the  dependencies in Cargo.toml alphabetically.dependencies


Cargo.toml
```diff
@@ -34,78 +34,97 @@ tokio-tracing = ["tokio/tracing"]
...
-clap = { version = "4.4.18", features = ["derive"] }
+itertools = "0.12"
+log = "0.4"
+memory = { path = "lib/common/memory" }
+parking_lot = { version = "0.12.1", features = [
+    "deadlock_detection",
+], optional = true }
+prometheus = { version = "0.13.3", default-features = false }
+prost = "0.11.9"
+# Consensus related crates
```
timvisee, line 77: Leaving this comment here does not make sense now.

Cargo.toml
```diff
@@ -34,78 +34,97 @@ tokio-tracing = ["tokio/tracing"]
...
-cancel = { path = "lib/common/cancel" }
-memory = { path = "lib/common/memory" }
-segment = { path = "lib/segment" }
-collection = { path = "lib/collection" }
-storage = { path = "lib/storage" }
-api = { path = "lib/api" }
-actix-multipart = "0.6.1"
-constant_time_eq = "0.3.0"
 
 # Profiling
```
timvisee, line 117: Same, this is an useless comment now.

generall (changes requested):
The benefit of sorting dependencies is kinda questionable. In current version we could have some kind of grouping.
In my opinion sorting only makes sense if we want to search faster, which doesn't look like a problem in case of cargo files.

I would only keep the "remove unused" part

ben1009:
> The benefit of sorting dependencies is kinda questionable. In current version we could have some kind of grouping. In my opinion sorting only makes sense if we want to search faster, which doesn't look like a problem in case of cargo files.
> 
> I would only keep the "remove unused" part

get your point

- revert to run `cargo sort -g` to keep the group if any
- for sort besides search faster, another benefit is [Large projects tend to like for things to stay sorted to cut down the occurrence of merge conflicts between two developers appending to the end of the same unsorted list in parallel changes. ](https://github.com/dtolnay/request-for-implementation/issues/29)
- for the "remove unused" part, already done in https://github.com/qdrant/qdrant/pull/3435, as you can see, no cargo.lock change in this pr

timvisee:
> for sort besides search faster, another benefit is [Large projects tend to like for things to stay sorted to cut down the occurrence of merge conflicts between two developers appending to the end of the same unsorted list in parallel changes. ](https://github.com/dtolnay/request-for-implementation/issues/29)

Yes, I do like this part, but we can still do that with grouping.

Removing unused and sorting groups would be a good middle ground for me :+1:

PR #3458, opened 2024-01-24
Limit automatic shard transfers (stateless)

timvisee:
Alternative for <https://github.com/qdrant/qdrant/pull/3449>.

This PR implements limiting automatic shard transfers for shard recovery in a cluster. The limit prevents a node from taking the whole cluster down with a lot of shard transfers.

It now limits the number of incoming and outgoing transfers to 1 each per collection, and is now also configurable.

It is a soft limit. Meaning it only applies to automatically initiated transfers, and not affect user requested transfers.

Basically, we now purely rely on the active transfers in the consensus state. On each consensus sync we find dead shards and request shard transfers for them to recover. It may take some time for a requested transfer to appear in the consensus state though. During each sync we also keep track of how many transfers we proposed, so that we can consider this when checking for shard transfer limits. It may happen that the same transfer is requested multiple times if consensus isn't updated in time, in which case duplicate transfer requests are dropped automatically. This is arguably a lot better than the previous implementation in <https://github.com/qdrant/qdrant/pull/3449> which required extra state, while this is stateless.

The configuration looks like this:

```yaml
storage:
  performance:
    # Soft limit for number of incoming shard transfers per collection on this node.
    # Only affects automatic shard transfers for recovery, does not affect user requested transfers.
    # The same value should be used on all nodes in a cluster.
    # If null - unlimited.
    incoming_shard_transfers_limit: 1

    # Soft limit for number of outgoing shard transfers per collection on this node.
    # Only affects automatic shard transfers for recovery, does not affect user requested transfers.
    # The same value should be used on all nodes in a cluster.
    # If null - unlimited.
    outgoing_shard_transfers_limit: 1
```

This has consensus tests to: 1. ensure we don't go above the default limit. 2. ensure we reach custom limits. 3. ensure user requested transfers always go through.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/collection/src/collection/mod.rs
```diff
@@ -390,31 +390,6 @@ impl Collection {
...
-                        self.shared_storage_config
-                            .default_shard_transfer_method
-                            .unwrap_or_default(),
-                    ),
-                })
-            } else {
-                log::warn!("No alive replicas to recover shard {shard_id}");
-            }
-        }
-
```
timvisee, line 417: This removes logic for requesting transfers on dead shards in the `set_shard_replica_state` function here.

I believe this is fine, because we do the same check in `sync_local_state`.

In fact, keeping this logic here is unsupported in this stateless implementation.

agourlay:
Thanks for this second take on the problem :+1: 

Any idea how we could test this somehow?

lib/collection/src/shards/shard_holder.rs
```diff
@@ -282,6 +282,20 @@ impl ShardHolder {
         })?)
     }
 
+    /// The count of incoming and outgoing shard transfers on the given peer
+    ///
+    /// This only includes shard transfers that are in consensus for the current collection. A
+    /// shard transfer that has just been proposed may not be included yet.
+    pub fn count_shard_tranfser_io(&self, peer_id: &PeerId) -> (usize, usize) {
```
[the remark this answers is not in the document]
timvisee: Woah, good catch! :eye: :eye: 
Thanks!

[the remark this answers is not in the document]
timvisee: Patched in 2344f1e9aa7f8231dde835e966d37860d42b9eef

timvisee:
> Thanks for this second take on the problem 👍
> 
> Any idea how we could test this somehow?

I'm about to dive into that now.

Not sure yet. My plan is to see if we have any tests that do recovery transfers in our test suite. If we do, extend it with a check constantly counting ongoing transfers.

agourlay:
@timvisee What about adding a `debug_assert` when fetching the cluster state to validate that the number of transfer is indeed under the limit?
That's not a test but it will give us more confidence :)

timvisee:
> What about adding a `debug_assert` when fetching the cluster state to validate that the number of transfer is indeed under the limit?
> That's not a test but it will give us more confidence :)

Interesting idea!

I see two problems with this though:
- If we ever go over the limit somehow, we should not crash
- This is just a soft limit for automated transfers. User requests are always accepted no matter the limit. We cannot distinguish between automated and user requests.

agourlay:
> This is just a soft limit for automated transfers. User requests are always accepted no matter the limit.

I do not like this very much TBH, the system should protect itself.

> If we ever go over the limit somehow, we should not crash (I think at least)

The `debug_assert` won't make it into the release build, so it "should" be ok.

What about adding a field to the transfer object containing the source of the trigger?
Enum looking like:
- recovery
- user demand
- other

timvisee:
> I do not like this very much TBH, the system should protect itself.

The idea is that when a user is sending a request, the user knows what he's doing. Maybe his requested transfer(s) are important for building up the node again with custom management logic.

What would you do when such request comes in? We cannot queue these requests. So we either have to reject it, or block the call which can take multiple hours to resolve. I don't like either of these.

> What about adding a field to the transfer object containing the source of the trigger?

Interesting suggestion. We can definitely do that. Maybe we've to postpone this until the next minor release though.

I don't like that we crash if we reach this condition though, even if it's just in debug. Would it be better to print an error instead?

lib/collection/src/collection/mod.rs
```diff
@@ -532,6 +513,21 @@ impl Collection {
...
+                transfer_io.1,
+            );
+            if self.check_shard_transfer_limit(transfer_io) {
+                log::trace!(
+                    "Postponing automatic shard {shard_id} transfer to stay below limit on this node (i/o: {}/{})",
+                    transfer_io.0,
+                    transfer_io.1,
+                );
+                continue;
+            }
```
ffuugoo, line 529: I'm sorry, but this is unreadable.

I'm still confused regarding incoming/outgoing:
- `count_shard_transfer_io` returns `(outgoing, incoming)`
- but `check_shard_transfer_limit` expects `(incoming, outgoing)`
- but when we do `log::trace` we also use them "reversed":
  - `transfer_io.0` is `incoming`
  - and `transfer_io.1` is `outgoing`

I've made a few suggestions, without changing the order anywhere, but please clarify these. 😂

```suggestion
            let (mut outgoing, incoming = shard_holder.count_shard_transfer_io(this_peer_id);
            outgoing += proposed.get(this_peer_id).copied().unwrap_or(0);

            if self.check_shard_transfer_limit(incoming, outgoing) {
                log::trace!(
                    "Postponing automatic shard {shard_id} transfer to stay below limit on this node (i/o: {outgoing}/{incoming})"
                );
                continue;
            }
```

lib/collection/src/collection/mod.rs
```diff
@@ -551,6 +547,21 @@ impl Collection {
...
+                    transfer_io.1 + *proposed.get(&replica_id).unwrap_or(&0),
+                );
+                if self.check_shard_transfer_limit(transfer_io) {
+                    log::trace!(
+                        "Postponing automatic shard {shard_id} transfer to stay below limit on peer {replica_id} (i/o: {}/{})",
+                        transfer_io.0,
+                        transfer_io.1,
+                    );
+                    continue;
+                }
```
ffuugoo, line 563: Ditto.

```suggestion
                let (outgoing, mut incoming) = shard_holder.count_shard_transfer_io(&replica_id);
                incoming += proposed.get(&replica_id).copied().unwrap_or(0);

                if self.check_shard_transfer_limit(outgoing, incoming) {
                    log::trace!(
                        "Postponing automatic shard {shard_id} transfer to stay below limit on peer {replica_id} (i/o: {outgoing}/{incoming})"
                    );
                    continue;
                }
```

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -357,4 +356,18 @@ impl Collection {
...
+        self.shared_storage_config
+            .incoming_shard_transfers_limit
+            .map(|limit| incoming >= limit)
+            .unwrap_or(false)
+            || self
+                .shared_storage_config
+                .outgoing_shard_transfers_limit
+                .map(|limit| outgoing >= limit)
+                .unwrap_or(false)
+    }
```
ffuugoo, line 372:
```suggestion
    pub(super) fn check_shard_transfer_limit(&self, incoming: usize, outgoing: usize) -> bool {
        let incoming_shard_transfer_limit_reached = self.shared_storage_config
            .incoming_shard_transfers_limit
            .map_or(false, |limit| incoming >= limit);
            
        let outgoing_shard_transfer_limit_reached = self
            .shared_storage_config
            .outgoing_shard_transfers_limit
            .map_or(false, |limit| outgoing >= limit);

         incoming_shard_transfer_limit_reached || outgoing_shard_transfer_limit_reached
    }
```

lib/collection/src/shards/shard_holder.rs
```diff
@@ -282,6 +282,20 @@ impl ShardHolder {
...
+    /// shard transfer that has just been proposed may not be included yet.
+    pub fn count_shard_transfer_io(&self, peer_id: &PeerId) -> (usize, usize) {
+        self.shard_transfers
+            .read()
+            .iter()
+            .map(|transfer| (transfer.to == *peer_id, transfer.from == *peer_id))
+            .fold((0, 0), |(i, o), (is_outgoing, is_incoming)| {
+                (i + is_outgoing as usize, o + is_incoming as usize)
+            })
+    }
```
ffuugoo, line 297:
```suggestion
    pub fn count_shard_transfer_io(&self, peer_id: &PeerId) -> (usize, usize) {
        let mut outgoing = 0;
        let mut incoming = 0;
        
        for transfer in &self.shard_transfers {
            if transfer.to == *peer_id {
                outgoing += 1;
            }
        
            if transfer.from == *peer_id {
                incoming += 1;
            }
        }
        
        (outgoing, incoming)
    }
```

lib/collection/src/shards/transfer/mod.rs
```diff
@@ -61,7 +61,7 @@ pub struct ShardTransferKey {
 
 impl ShardTransferKey {
     pub fn check(&self, transfer: &ShardTransfer) -> bool {
-        self.shard_id == transfer.shard_id && self.from == transfer.from && self.to == transfer.to
+        transfer.key() == *self
     }
```
ffuugoo, line 65: Can't "suggest" on unmodified lines, but as `ShardTransferKey` is `Copy` now:

```rust
pub fn check(self, transfer: &ShardTransfer) -> bool {
  self == transfer.key()
}
```

lib/collection/src/collection/mod.rs
timvisee: It looks like I did mix up variables, but the result is the same.

The intent is to always do input/output because of io.

lib/collection/src/collection/shard_transfer.rs
timvisee: Fine. But this does not allow shortcutting.

lib/collection/src/shards/transfer/mod.rs
timvisee: Good catch removing the &. Thanks!

lib/collection/src/shards/shard_holder.rs
timvisee: Yours is the wrong way around. `.to` means it'll be _incoming_ on that node, while `.from` means it'll be _outgoing_ on that node.

I mixed up the variable names accidentally in my version, because I flipped the ordering around some commit back. But the result was the same.

lib/collection/src/collection/shard_transfer.rs
ffuugoo: It's a single `>=` comparison! Also, compiler should still be able to optimize it.  😂

timvisee: Alright! I thought this was doing a loop, sorry. :smile:

lib/collection/src/collection/mod.rs
```diff
@@ -532,6 +513,14 @@ impl Collection {
             // Try to find dead replicas with no active transfers
             let transfers = shard_holder.get_transfers(|_| true);
 
+            // Respect shard transfer limit, consider already proposed transfers in our counts
+            let (mut incoming, outgoing) = shard_holder.count_shard_transfer_io(this_peer_id);
+            incoming += proposed.get(this_peer_id).copied().unwrap_or(0);
+            if self.check_shard_transfer_limit(incoming, outgoing) {
+                log::trace!("Postponing automatic shard {shard_id} transfer to stay below limit on this node (i/o: {incoming}/{outgoing})");
```
ffuugoo, line 520: Final suggestion:

```suggestion
                log::trace!("Postponing automatic shard {shard_id} transfer to stay below limit on this node (incoming: {incoming}, outgoing: {outgoing})");
```

lib/collection/src/collection/mod.rs
```diff
@@ -551,6 +540,14 @@ impl Collection {
                     continue; // this transfer won't work
                 }
 
+                // Respect shard transfer limit, consider already proposed transfers in our counts
+                let (incoming, mut outgoing) = shard_holder.count_shard_transfer_io(&replica_id);
+                outgoing += proposed.get(&replica_id).copied().unwrap_or(0);
+                if self.check_shard_transfer_limit(incoming, outgoing) {
+                    log::trace!("Postponing automatic shard {shard_id} transfer to stay below limit on peer {replica_id} (i/o: {incoming}/{outgoing})");
```
ffuugoo, line 547: Ditto.

```suggestion
                    log::trace!("Postponing automatic shard {shard_id} transfer to stay below limit on peer {replica_id} (incoming: {incoming}, outgoing: {outgoing})");
```

ffuugoo (approved):

lib/collection/src/collection/mod.rs
timvisee: I had this originally but changed it to `i/o` because I thought it was quite long.

But happy to change it back :+1: 

Thanks for another review round!

timvisee:
I added three consensus tests for this functionality:
1. a test to ensure automatic transfers never go over the default limit of 1
2. a test setting 3/2 as i/o limits, which requires to reach 3 (and no more) ongoing transfers at _some_ point
3. a test with extra user requests that ensures we do go over the limit at _some_ point

Regarding the debug assertion: I suggest to postpone it and add it in the next minor release. This, because it requires API changes, while we want to include this PR in a patch release.

cc @agourlay

config/config.yaml
```diff
@@ -50,6 +51,18 @@ storage:
     # If null - auto selection.
     update_rate_limit: null
 
+    # Soft limit for number of incoming shard transfers per collection on this node.
+    # Only affects automatic shard transfers for recovery, does not affect user requested transfers.
```
coszio, line 55: I think the word `soft` might create more confusion than clarification. How about something like this?
```suggestion
    # Limit for number of incoming automatic shard transfers per collection on this node, does not affect user-requested transfers.
```

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -357,4 +356,20 @@ impl Collection {
             }
         }
     }
+
+    /// Whether we have reached the shard transfer soft limit based on the given incoming and
+    /// outgoing transfers.
+    pub(super) fn check_shard_transfer_limit(&self, incoming: usize, outgoing: usize) -> bool {
```
coszio, line 362: Probably same comment here
```suggestion
    pub(super) fn check_automatic_shard_transfer_limit(&self, incoming: usize, outgoing: usize) -> bool {
```

lib/collection/src/operations/shared_storage_config.rs
```diff
@@ -36,6 +38,8 @@ impl Default for SharedStorageConfig {
             update_concurrency: None,
             is_distributed: false,
             default_shard_transfer_method: None,
+            incoming_shard_transfers_limit: Some(1),
+            outgoing_shard_transfers_limit: Some(1),
```
coszio, line 42: Based on the description in `config.yaml`, I thought default was `None`?

Also: 

```suggestion
            incoming_shard_transfers_limit: default_io_shard_transfers_limit(),
            outgoing_shard_transfers_limit: default_io_shard_transfers_limit(),
```

lib/collection/src/shards/shard_holder.rs
```diff
@@ -282,6 +282,21 @@ impl ShardHolder {
...
+    /// shard transfer that has just been proposed may not be included yet.
+    pub fn count_shard_transfer_io(&self, peer_id: &PeerId) -> (usize, usize) {
+        let (mut incoming, mut outgoing) = (0, 0);
+
+        for transfer in self.shard_transfers.read().iter() {
+            incoming += (transfer.to == *peer_id) as usize;
+            outgoing += (transfer.from == *peer_id) as usize;
+        }
+
+        (incoming, outgoing)
```
coszio, line 297: Probably useless suggestion, but:
```suggestion
        self.shard_transfers
            .read()
            .iter()
            .fold((0, 0), |(incoming, outgoing), transfer| {
                (
                    incoming + (transfer.to == *peer_id) as usize,
                    outgoing + (transfer.from == *peer_id) as usize,
                )
            })
```
😄

ffuugoo: 😡😡😡

config/config.yaml
timvisee: Agreed. Thanks!

lib/collection/src/operations/shared_storage_config.rs
timvisee: The default is 1 indeed, not `None`. But a user could provide `None` to remove any limits.

I'll update the comments in `config.yaml` to clarify this better.

generall (approved):

PR #3462, opened 2024-01-25
Only install the main packages

mrbusche:
and ignore recommended packages; delete cache from build step
Also, fix consistency with `AS` for alias

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ COPY --from=xx / /
 # so, please, don't reorder them without prior consideration. 🥲
 
 RUN apt-get update \
-    && apt-get install --no-install-recommends -y clang lld cmake protobuf-compiler jq \
+    && apt-get install -y clang lld cmake protobuf-compiler jq \
     && rustup component add rustfmt \
     && rm -rf /var/lib/apt/lists/*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,8 +32,7 @@ COPY --from=xx / /
 
 RUN apt-get update \
     && apt-get install -y clang lld cmake protobuf-compiler jq \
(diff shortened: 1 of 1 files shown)


timvisee (approved):
I assume this is fine. Thanks! 

Lets see if our CI stays green, and then I'd be happy to merge.

timvisee (changes requested):
I'm afraid this eliminates packages we actually need. We see a bunch of ProtoBuf errors when building the Docker image in CI.

mrbusche:
> I'm afraid this eliminates packages we actually need. We see a bunch of ProtoBuf errors when building the Docker image in CI.

Very strange, I was able to build and run locally. I will take another look.

timvisee:
Maybe our CI is running a slightly different configuration.

Anyway, thank you for testing it locally first, and thanks for taking another look!

mrbusche:
I tried to mimic the build system and made one change. Still builds on my Windows PC, but we'll see what happens. It barely decrease the build size, not as impactful as I had hoped.

Dockerfile
```diff
@@ -32,7 +32,8 @@ COPY --from=xx / /
 
 RUN apt-get update \
     && apt-get install -y clang lld cmake protobuf-compiler jq \
-    && rustup component add rustfmt
+    && rustup component add rustfmt \
+    && rm -rf /var/lib/apt/lists/*
```
timvisee, line 36: Since this is an intermediate container, removing apt sources is not needed here as they're dropped anyway.

I guess it's better to remove this line then.

```suggestion
    && rustup component add rustfmt
```

timvisee (approved):
Thank you! 😁

PR #3464, opened 2024-01-25
Optimize sparse vector search with batched scoring

agourlay:
This PR optimizes the sparse vector search by using a new batched approach.

# Motivation

The existing implementation walks the posting list in lockstep by:
- finding the `min_record_id` at the head of all posting list
- score heads where `record_id == min_record_id`
- advance scored posting list by 1
- repeat until all posting lists are empty

This algorithm is slow because it jumps all over the place in memory to process intersections one by one.

# Batching

This PR introduces a new batching approach that relies on sequential access instead.

It starts by computing the range of all possible record ids to score in the posting lists.

Given that range, we can compute a maximum record id watermark according to a given batch size and then score all elements in the posting lists *up to* the watermark.

This can be done efficiently by reading the posting list sequentially and storing the intermediate scores.

Then in all posting lists, the cursors are advanced up to the watermark.

This process is repeated until all posting lists are exhausted.

I had to remove a few unit tests that were getting in the way of the refactoring.

Correctness is ensured by:
- the [sparse_vector_index_search_tests](https://github.com/qdrant/qdrant/blob/master/lib/segment/tests/integration/sparse_vector_index_search_tests.rs)
- the Python client sparse congruence tests
- the validation of ground-truth data in the performance tests

# Performance numbers

The performance gains are validated by https://github.com/qdrant/sparse-vectors-benchmark which uses a real-life dataset from BigANN.

A bit about the setup:
- 8 segments
- index.on_disk=false
- 6980 queries

```
DEV
min: 230.57 millis
50p: 1772.39 millis
95p: 3626.78 millis
99p: 4462.23 millis
999p: 5094.45 millis
max: 5411.65 millis

PR
min: 125.61 millis
50p: 374.41 millis
95p: 570.08 millis
99p: 644.55 millis
999p: 723.14 millis
max: 772.38 millis
```

DEV             |  PR
sparse_bench_small_dev  | sparse_bench_small_PR

Without the same point of reference on the y-axis the latency look likes this:

sparse_bench_full_PR_no_scaled

# Future work

The performance gets better as the batch size increases. However, we do not want to allocate a large chunk of memory for each incoming request.

A possible solution is to pool a set of larger vectors at the index level to reuse across searches to account for concurrent access.

# Things I tried that did not work out

## Bitset of all ids to traverse

Using a [roaring bitmap](https://roaringbitmap.org/) I captured all the ids in the posting lists and then simply iterate over those.

pros: 
- no ids guessing
- apply predicate filter before scoring

cons:
- building the Bitset requires to traverse the posting list for each query
- the Bitset object is expensive to build

It could be tractable if we pre-compute and persist one bitset per posting list and then simply intersect at query time.

## Bloom-filter for ids

Traverse all ids between `min_id` and `max_id` but before processing, verify that the id is present within the posting lists using a bloom filter.

pros:
- almost no ids guessing
- apply predicate before scoring 

cons:
- building the bloom-filter requires to traverse the posting list for each query
- querying the bloom filter dominates the query CPU time.


lib/sparse/src/index/search_context.rs
```diff
@@ -900,70 +738,70 @@ mod tests {
...
-                },
-                ScoredPointOffset {
-                    score: -2.0,
-                    idx: 2
-                }
-            ]
-        );
-    }
+    // #[test]
+    // fn pruning_does_not_skip_negative_score_test() {
```
generall, line 742: keeping it for the further work?

agourlay: Good catch :+1: 

Yes I want to revisit pruning later on and adapt this test.

generall (approved):
Nuke the commended test and it is good to go

lib/sparse/src/index/search_context.rs
```diff
@@ -106,72 +126,49 @@ impl<'a> SearchContext<'a> {
...
+            for element in posting.posting_list_iterator.remaining_elements() {
+                let element_id = element.record_id;
+                if element_id > batch_last_id {
+                    // reaching end of the batch
+                    break;
                 }
-            } else {
-                // mark search context for cleanup if any posting list iterator is exhausted
-                self.contains_empty_posting = true;
+                let element_score = element.weight * posting.query_weight;
```
IvanPleshkov, line 143: I cannot understand why do we skip here the filter condition? What if `element.record_id` is filtered?

IvanPleshkov: Got it, we calculate score even for deleted points. I can propose to try avoid this additional calculation with doing filter conditions before scoring. keep filter result in biset and resuse it for both scoring and `self.result_queue` update

lib/sparse/src/index/search_context.rs
```diff
@@ -247,30 +244,49 @@ impl<'a> SearchContext<'a> {
...
+            };
+
+            // compute batch range of contiguous ids for the next batch
+            let last_batch_id = min(
+                start_batch_id + ADVANCE_BATCH_SIZE as u32,
+                self.max_record_id,
+            );
+            let batch_len = last_batch_id - start_batch_id + 1;
+
+            // init batch scores
```
IvanPleshkov, line 266: I propose to move this part
```
            // init batch scores
            self.batch_scores.clear();
            self.batch_scores.resize(batch_len as usize, 0.0);
```
into the `advance_batch` method. Because `self.batch` is `advance_batch` internal cache and we don't use `self.batch` inside search function directly

IvanPleshkov (approved):
Perfect PR with perfect results. Leaved just some non-blocking questions and nitpicks

lib/sparse/src/index/search_context.rs
IvanPleshkov: after discussion with @agourlay we decided to keep as is because benchmark shown that this approach is better

lib/sparse/src/index/search_context.rs
```diff
@@ -36,35 +40,51 @@ impl<'a> SearchContext<'a> {
...
         let result_queue = FixedLengthPriorityQueue::new(top);
         // Query vectors with negative values can NOT use the pruning mechanism which relies on the pre-computed `max_next_weight`.
         // The max contribution per posting list that we calculate is not made to compute the max value of two negative numbers.
         // This is a limitation of the current pruning implementation.
         let use_pruning = query.values.iter().all(|v| *v >= 0.0);
-        // find min record id across all posting lists
-        let min_record_id = Self::next_min_id(&postings_iterators);
-        // no empty posting lists at the beginning
-        let contains_empty_posting = false;
+        // TODO pool this Vec to reuse memory across searches
```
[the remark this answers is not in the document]
IvanPleshkov: @agourlay said in a personal conversation that batch size may be much larger such a 10K for instance. It makes sense, resolve this conversation

Issue #3331, opened 2024-01-05
Support `minimum should match` clause 


generall:
**Is your feature request related to a problem? Please describe.**

Currently, qdrant only support `should` clause, where any `true` condition results in `true` for the whole `should` clause.
But sometimes it is useful to specify a minimal amount of conditions, that should be resolved to turn the whole clause into `true`.

**Describe the solution you'd like**

The API design is up to discussion. Here are a couple of options:

```json
{
    "filter": {
        "must": [
            {
                "min_should_match": {
                    "threshold / count / min_count": 2,
                    "conditions": [
                        { ... },
                        { ... },
                        { ... }
                    ]
                }
            }
        ]
    }
}
```

```json
{
    "filter": {
        "min_should": {
            "conditions": [
                { ... },
                { ... },
                { ... }
            ],
            "threshold / count / min_count": 2
        }
    }
}
```

Arguments for a specific interface variant are welcome.

**Describe alternatives you've considered**

The main decision point is to choose which interface option to implement

**Additional context**

Important parts of this task:

- to properly implement cardinality estimation for the request.  
- Implement backward compatible REST and gRPC

---

Note for contributors: Please consider this as tracking issue. If you think that it would be beneficial to split the task into multiple smaller PRs, please you are welcome to do so. Bounty will be rewarded for each PR independently

timvisee:
I like your base suggestions! I think I like the first interface example better, as it seems more versatile (also usable in `must_not` for example).

Do we also want to support a maximum, to potentially support ranges? I'm not sure if there'd be a use case for this.

Maybe we can source some inspiration from here as well:
- https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-minimum-should-match.html

ding-young:
> I like your base suggestions! I think I like the first interface example better, as it seems more versatile (also usable in `must_not` for example).
> 
> Do we also want to support a maximum, to potentially support ranges? I'm not sure if there'd be a use case for this.
> 
> Maybe we can source some inspiration from here as well:
> 
>     * https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-minimum-should-match.html
Hi, I'm currently working on this issue and have concerns about the interface.  
Besides two nice suggestions described in the initial post, Elasticsearch style interface seems like below.
```json
{
    "filter": {
        "should":  [
                { ... },
                { ... },
                { ... }
        ],
        "threshold / count / min_count": 2
    }
}
```
[...]

PR #3466, opened 2024-01-25
Support Min should clause (#3331)

ding-young:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

## Explanation
/claim #3331 

Support `min-should` clause for finding points whose payloads match more than minimal amount of conditions. Implemented by adding a field in `Filter` struct. 
- API 
```json
{
    "filter": {
        "min_should": {
            "conditions": [
                { ... },
                { ... },
                { ... }
            ],
            "min_count": 2
        }
    }
}
```
Above example is basic usage of `min_should` clause, but it can also be nested in another filter clause as follows. In that case, it is regarded as a condition so below request will return points except the ones that makes `min_should` clause true.
```json
{
    "filter": {
        "must_not": [
            {
                "min_should": {
                    "min_count": 2,
                    "conditions": [
                        { ... },
                        { ... },
                        { ... }
                    ]
                }
            }
        ]
    }
}
```

- Cardinality Estimation   
Let A, B, C, D refer to the set of points that match each condition specified in `min_should` clause, respectively. 
Since the result set that matches at least `min_count` (ex. 2) conditions is equivalent to union of possible intersections 
ex. (A ⋂ B) ∪ (A ⋂ C) ∪ (A ⋂ D) ∪ (B ⋂ C) ∪ (B ⋂ D) ∪ (C ⋂ D) , I first estimated the cardinality of each intersection set, and then combined these estimations as union. However, I'm not sure whether this approximation is appropriate.


lib/segment/src/index/query_estimator.rs
```diff
@@ -125,6 +125,27 @@ pub fn combine_should_estimations(
...
+        .iter()
+        .combinations(min_count)
+        .map(|intersection| {
+            combine_must_estimations(&intersection.into_iter().cloned().collect_vec(), total)
+        })
+        .collect_vec();
+
+    combine_should_estimations(&intersection_estimations, total)
+}
+
```
ding-young, line 148: May I ask for suggestions or feedback about cardinality estimation?

lib/segment/src/types.rs
```diff
@@ -2010,6 +2041,18 @@ impl Filter {
...
+                match (self.min_should.clone(), other.min_should.clone()) {
+                    (None, None) => None,
+                    (Some(this), None) => Some(this),
+                    (None, Some(other)) => Some(other),
+                    (Some(mut this), Some(other)) => {
+                        this.conditions.extend(other.conditions);
+                        Some(this) // min_count follows self.min_should.min_count
+                    }
+                }
+            },
```
ding-young, line 2055: I think lack understanding of merging Filters. Is there any use case when merging min_should clauses is required? The only cases that I found in source code was merging `must` or `must_not` filters.

lib/segment/src/types.rs
```diff
@@ -2010,6 +2041,18 @@ impl Filter {
...
             must: merge_component(self.must.clone(), other.must.clone()),
             must_not: merge_component(self.must_not.clone(), other.must_not.clone()),
```
ding-young, line 2057: I think I lack understanding of merging Filters. Is there any situations that merging min_should clause is required? The only use cases I found in the source code was merging `must` or `must_not` clauses in `group_by.rs`

lib/segment/src/index/query_estimator.rs
coszio: After thinking about it for a while, I believe the logic is sound. I also added an extra equivalence test for completeness.

lib/segment/src/types.rs
coszio: The usage of merging Filters in `group_by` is just for altering them from within qdrant, and the usage is pretty established by now without using min_should, so this will not break anything there. 

I don't know if there will ever be a use-case for merging min_should conditions, but my intuition says that the `min_count` should be either the min, max or the sum of both `min_count`s, to keep commutativity. From those options I vote for keeping the max.

coszio (commented):
I like the shape of this. Just going to push a few nits to mark my approval, if you don't mind.

ding-young:
Thank you for your review, @coszio ! I've checked your modifications. 
By the way, may I work on fixing the merge conflicts? I just want to tidy up the test cases in `lib/segment/src/payload_storage/query_checker.rs`

coszio (approved):

lib/segment/src/types.rs
```diff
@@ -2025,12 +2025,31 @@ pub struct WithPayload {
     pub payload_selector: Option<PayloadSelector>,
 }
 
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Validate, Clone, PartialEq, Default)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
```
generall, line 2029: Is `deny_unknown_fields` required? Usually we prefer not to use it unless it is strictly necessary. Otherwise it would be core complicated to extend this data structure further and maintain compatibility with previous versions.

generall (changes requested):
Hey @dzhao, thanks for your PR! There is a couple of minor problems, but overall everything looks good!

Especially interesting idea with cardinality estimation, at first I had a concern that `combinations` might explode, but on the second though I don't think it could happen in practice.

ding-young:
Hi, @generall @coszio . I checked test passes locally but build fails on gh action. It says that there's missing min_should field in 

`
   --> lib/segment/src/fixtures/payload_fixtures.rs:225:5

`
, but it seems that the code line number does not match the newest source code.

generall:
some kind of merge conflict, locally looks correct for me as well

generall:
Maybe next time prefer rebase over [Merge remote-tracking branch 'upstream/dev' into min-should](https://github.com/qdrant/qdrant/pull/3466/commits/24f5e8d6b112bf96b3257bc70edc2d236f9ceb26)

I will try merge as is, will see how it goes

ding-young:
Got it. Thank you for help 😃

PR #3469, opened 2024-01-26
ci: set rust-toolchain to stable

ben1009:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

- set the rust-toolchain to stable

### DIFF
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "stable"
--- a/.github/workflows/rust-lint.yml
+++ b/.github/workflows/rust-lint.yml
@@ -13,15 +13,20 @@ jobs:
   test:
 
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        # 1.75 is the MSRV, https://github.com/qdrant/qdrant/pull/3315
+        rust-version: [ "1.75", stable ]
+      fail-fast: false
     steps:
     - name: Install minimal nightly (only for fmt)
       uses: dtolnay/rust-toolchain@nightly
       with:
         components: rustfmt
-    - name: Install minimal stable
-      uses: dtolnay/rust-toolchain@stable
+    - name: Install rust
+      uses: dtolnay/rust-toolchain@master
       with:
+        toolchain: ${{ matrix.rust-version }}
         components: rustfmt, clippy
     - uses: actions/checkout@v4
     - name: Install Protoc
--- a/.github/workflows/rust-lint.yml
+++ b/.github/workflows/rust-lint.yml
@@ -13,20 +13,15 @@ jobs:
   test:
 
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        # 1.75 is the MSRV, https://github.com/qdrant/qdrant/pull/3315
-        rust-version: [ "1.75", stable ]
-      fail-fast: false
+
     steps:
     - name: Install minimal nightly (only for fmt)
       uses: dtolnay/rust-toolchain@nightly
       with:
         components: rustfmt
-    - name: Install rust
-      uses: dtolnay/rust-toolchain@master
+    - name: Install minimal stable
+      uses: dtolnay/rust-toolchain@stable
       with:
-        toolchain: ${{ matrix.rust-version }}
         components: rustfmt, clippy
     - uses: actions/checkout@v4
     - name: Install Protoc
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,6 @@ homepage = "https://qdrant.tech/"
 repository = "https://github.com/qdrant/qdrant"
 license = "Apache-2.0"
 edition = "2021"
-rust-version = "1.75"
 default-run = "qdrant"
 
 [features]


timvisee:
Sure! Simply addition, but I have no objection against this.

I don't see this matrix being used yet in the current run, do you have any idea why?
https://github.com/qdrant/qdrant/actions/runs/7663917140/job/20893663932?pr=3469

ben1009:
> Sure! Simply addition, but I have no objection against this.
> 
> 
> 
> I don't see this matrix being used yet in the current run, do you have any idea why?
> 
> https://github.com/qdrant/qdrant/actions/runs/7663917140/job/20893663932?pr=3469

Oh, currently only added to lints. Do you want add msrv check to test too?

timvisee:
> > Sure! Simply addition, but I have no objection against this.
> > I don't see this matrix being used yet in the current run, do you have any idea why?
> > https://github.com/qdrant/qdrant/actions/runs/7663917140/job/20893663932?pr=3469
> 
> Oh, currently only added to lints. Do you want add msrv check to test too?

I see! The list of runs is unordered so I had a hard time finding it.

No, I think it's all good like this!

timvisee (approved):

agourlay:
Thanks for the contribution :+1: 

I am trying to understand exactly what this change is protecting us from.
a.k.a. when will this be red?

ben1009:
> Thanks for the contribution 👍
> 
> I am trying to understand exactly what this change is protecting us from. a.k.a. when will this be red?

Currently we are testing against [stable](https://github.com/qdrant/qdrant/actions/runs/7663917140/job/20893663932?pr=3469#step:2:65) which is 1.75 for now, equals to the [msrv](https://github.com/qdrant/qdrant/blob/dev/Cargo.toml#L13). but let’s say after several months, stable become 1.78, there is a chance some of the change may not working under 1.75, but no way to detect that in current ci setup

timvisee:
> Currently we are testing against [stable](https://github.com/qdrant/qdrant/actions/runs/7663917140/job/20893663932?pr=3469#step:2:65) which is 1.75 for now, equals to the [msrv](https://github.com/qdrant/qdrant/blob/dev/Cargo.toml#L13). but let’s say after several months, stable become 1.78, there is a chance some of the change may not working under 1.75, but no way to detect that in current ci setup

To be fair, though we define our MSRV in the `Cargo.toml`, we don't strictly want to keep support in it.

Since this is not a library, we prefer to always use the latest stable Rust version instead so that we can always use the newest features.

Though this test may catch a case where where our configure MSRV mismatched, it's probably not super imporatnt.

ben1009:
> > Currently we are testing against [stable](https://github.com/qdrant/qdrant/actions/runs/7663917140/job/20893663932?pr=3469#step:2:65) which is 1.75 for now, equals to the [msrv](https://github.com/qdrant/qdrant/blob/dev/Cargo.toml?rgh-link-date=2024-01-26T15%3A53%3A10Z#L13). but let’s say after several months, stable become 1.78, there is a chance some of the change may not working under 1.75, but no way to detect that in current ci setup
> 
> To be fair, though we define our MSRV in the `Cargo.toml`, we don't strictly want to keep support in it.
> 
> Since this is not a library, we prefer to always use the latest stable Rust version instead so that we can always use the newest features.
> 
> Though this test may catch a case where where our configure MSRV mismatched, it's probably not super important.

agree, so maybe add a rust-toolchain  would be better

ffuugoo (approved):

rust-toolchain
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "stable"
```
agourlay, line 2: I do not think this is bringing much to the table.

My understanding is that the toolchain is mostly useful for pinning specific nightly revision.

What if the user has not pulled the latest stable? It could still fail without the clear information of the MSRV.

TBH I am not convinced this change is improving on the status quo.

ben1009: I am afraid i am not follow, do we come to a agreement the msrv is not need. We just want always use the latest stable?

timvisee: Yes, we don't want to pin to some MSRV and just use stable.

The `rust-toolchain` file you added doesn't guarantee that we use the latest nightly.

timvisee: Latest stable, sorry!

ben1009: So you mean we should pin to a stable/nighty version, after the latest stable released, run some tests, if everything work as expected, upgrade to that version?

timvisee: No. I mean that pinning `"stable"` here does not make much sense. It doesn't refer to the latest released stable per-se.

timvisee: To clarify, just removing `rust-version = ""` from `Cargo.toml` is probably enough.

ben1009: > No. I mean that pinning `"stable"` here does not make much sense. It doesn't refer to the latest released stable per-se.

Sorry, I am confused, if not what is the value for the **stable**

generall:
Agreed with the comments above, I don't see how this change is helpful

rust-toolchain
ben1009: Thx for the explanation, though already closed 😶

Issue #3332, opened 2024-01-05
API for checking collection exists


generall:
**Is your feature request related to a problem? Please describe.**

Currently, there are 2 ways of checking if collection exists:

- list all collections and find name in the list
- query specific collection info and check if it is 404 or 200 response code

But in some clients if would be much more convenient to have a dedicated API which would return `true` or `false` without error http code.

**Describe the solution you'd like**

API `/collections/<collection-name>/check` and similar in gRPC

API should return `true` if collection exists and `false` if not (with all the usual result wrappers, of course)

**Describe alternatives you've considered**

Make more sophisticated and extendable API, which could return some additional information as well (but not as much as collection info)

Or make an API for batch request of collection info, so that you could specify multiple names and get results for all of them

**Additional context**

API should be consistent with our current set of endpoints

---

Note for contributors: Please consider this as tracking issue. If you think that it would be beneficial to split the task into multiple smaller PRs, please you are welcome to do so. Bounty will be rewarded for each PR independently

timvisee:
To be fair, in true REST you should rely on 404 when requesting the object.

> API `/collections/<collection-name>/check` and similar in gRPC

What about `/exists`, or do you eventually want to extend this with more types of checks?

PR #3472, opened 2024-01-26
API for checking collection exists

gabrieldemian:
### All Submissions:

/claim #3332 

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/api/src/grpc/proto/collections.proto
+++ b/lib/api/src/grpc/proto/collections.proto
@@ -54,8 +54,12 @@ message CollectionExistsRequest {
   string collection_name = 1;
 }
 
-message CollectionExistsResponse {
+message CollectionExists {
   bool exists = 1;
+}
+
+message CollectionExistsResponse {
+  CollectionExists result = 1;
   double time = 2; // Time spent to process
 }
 
--- a/lib/api/src/grpc/proto/collections.proto
+++ b/lib/api/src/grpc/proto/collections.proto
@@ -56,6 +56,7 @@ message CollectionExistsRequest {
 
 message CollectionExistsResponse {
   bool exists = 1;
+  double time = 2; // Time spent to process
 }
 
 message ListCollectionsRequest {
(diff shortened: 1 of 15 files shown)


lib/api/src/grpc/proto/collections.proto
```diff
@@ -50,6 +50,14 @@ message GetCollectionInfoRequest {
   string collection_name = 1; // Name of the collection
 }
 
+message CollectionExistsRequest {
+  string collection_name = 1;
+}
+
+message CollectionExistsResponse {
+  bool exists = 1;
+}
```
coszio, line 59: Just for consistency, we should also add timing in grpc, same as we do in http

coszio (changes requested):
Good work, I kinda want to approve this already, but I have this 1 item to resolve

lib/api/src/grpc/proto/collections.proto
gabrieldemian: thank you, just did this

generall:
Hi @gabrieldemian,

it looks like the REST api response is inconsistent now:

image

generall (changes requested):
Please fix the REST output format.

Also I noticed that our CI script for checking read-only API list consistency is broken. Here is a fix attached
[fix-read-only.diff.txt](https://github.com/qdrant/qdrant/files/14072832/fix-read-only.diff.txt)

gabrieldemian:
> Hi @gabrieldemian,
> 
> it looks like the REST api response is inconsistent now:
> 
> image

I fixed this inconsistency, it's working as expected on REST. On grpc, when the collection doesn't exist, `grpcurl` result is an empty object, however, the result is actually correct (I did a simple println! of the tonic::Response). I'm not sure if this is related to grpcurl.

image
image

gabrieldemian:
> Please fix the REST output format.
> 
> Also I noticed that our CI script for checking read-only API list consistency is broken. Here is a fix attached [fix-read-only.diff.txt](https://github.com/qdrant/qdrant/files/14072832/fix-read-only.diff.txt)

fixed that, I tested both REST and grpc on my terminal and the responses were as expected, a single time value outside the result object

timvisee (approved):

generall (approved):

generall (approved):
LGMT

PR #3473, opened 2024-01-28
feat: verify snapshot integrity with checksum

### LINKED ISSUE #3372: Verify snapshot integrity with checksum in shard snapshot transfer
Depends on <https://github.com/qdrant/qdrant/issues/3371>.

In <https://github.com/qdrant/qdrant/pull/2840> we've added checksums for snapshot files. The implementation is somewhat limited however, and requires further integration to make full use of it.

Since Qdrant 1.7 different shard transfer methods are supported. [Snapshot transfers](https://qdrant.tech/documentation/guides/distributed_deployment/#shard-transfer-method) have been added to make these transfers more capable by utilizing snapshots.

One problem with this approach is that we have no integrity checks for the actual snapshot files. If such file would become corrupted, Qdrant will happily restore possibly resulting in a broken shard.

When a shard snapshot transfer happens, we should check integrity of the snapshot file by verifying the attached checksum. Since <https://github.com/qdrant/qdrant/pull/2840>, the checksum is attached to the `SnapshotDescription` object.

<https://github.com/qdrant/qdrant/issues/3371> will implement a `checksum` field in snapshot recovery endpoints. We'll have to wait for this to be implemented so that we can utilize this in the snapshot transfer progress.

The right approach is probably to pass the checksum along in the recovery call here: https://github.com/qdrant/qdrant/blob/c6a351c82d57eb786b75dac785de7b085c6f80ff/lib/collection/src/shards/transfer/snapshot.rs#L215-L228

If checksum verification on the remote node fails, we should clean up the snapshot file and return with an error. Cleaning up on the remote is probably already handled with <https://github.com/qdrant/qdrant/issues/3371>.

**Additional context**
There's other work to be done to properly integrate checksums, but that will be handled in different issues/PRs.

dracarys18:
Closes #3372 

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/remote_shard.rs
+++ b/lib/collection/src/shards/remote_shard.rs
@@ -470,6 +470,7 @@ impl RemoteShard {
         shard_id: ShardId,
         url: &Url,
         snapshot_priority: SnapshotPriority,
+        checksum: &Option<String>,
     ) -> CollectionResult<RecoverSnapshotResponse> {
         let res = self
             .with_shard_snapshots_client_timeout(
@@ -484,7 +485,7 @@ impl RemoteShard {
                             snapshot_priority: api::grpc::qdrant::ShardSnapshotPriority::from(
                                 snapshot_priority,
                             ) as i32,
-                            checksum: None,
+                            checksum: checksum.clone(),
                         })
                         .await
                 },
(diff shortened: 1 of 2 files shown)


generall:
There is another open PR for the same issue: #9535. We'll review #9535 first; the other one is closed as a duplicate to keep review in one place — thanks to both authors.

Issue #1957, opened 2023-05-24
Provide additional docker version tags for the image containing only the major or minor version


volkerstampa:
**Is your feature request related to a problem? Please describe.**

When referring to qdrant's docker image as `qdrant/qdrant` you always get the newest release which includes potentially breaking major releases. When you refer to the image as `qdrant/qdrant:v1.2.3` you do not get updates at all not even potentially security relevant patch upgrades.

**Describe the solution you'd like**

If the latest release for a major version would also be tagged as for example `v1` and the latest release for a minor would also be tagged as for example `v1.2` then one could refer to the docker image using these additional tags to indicate if one is interested in _automatic_ minor or patch upgrades.

**Describe alternatives you've considered**

- use `latest` and risk breaking changes
- use specific version and miss security fixes

**Additional context**

PAARTH2608:
@volkerstampa @generall Hi, can you please assign this to me? I would like to work on this

generall:
we don't usually assign tasks to external contributors. Feel free to create a PR, though

jayantpranjal0:
@generall ig this issue also needs to be closed

timvisee:
Implemented by <https://github.com/qdrant/qdrant/pull/6611>

PR #3474, opened 2024-01-28
ci: Implement Semantic Versioning for Docker Image Deployment

yuvraj-wale:
Added major/minor tags support for docker image workflow.

linked to issue #1957

### All Submissions:

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -18,13 +18,23 @@ jobs:
     - name: Get current tag
       id: vars
       run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+    - name: Extract Major and Minor Versions
+      id: version
+      run: |
+        FULL_VERSION=${{ steps.vars.outputs.tag }}
+        MAJOR_VERSION=$(echo $FULL_VERSION | cut -d '.' -f 1)
+        MINOR_VERSION=$(echo $FULL_VERSION | cut -d '.' -f 1,2)
+        echo ::set-output name=major::${MAJOR_VERSION}
+        echo ::set-output name=minor::${MINOR_VERSION}
     - name: Build the Docker image
       env:
         RELEASE_VERSION: ${{ steps.vars.outputs.tag }}
+        MAJOR_VERSION: ${{ steps.version.outputs.major }}
+        MINOR_VERSION: ${{ steps.version.outputs.minor }}
       run: |
         # Create build container
         docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
-        docker buildx create --use
+        docker buildx create --us
 
         # Authenticate on registries
         docker login --username generall --password ${{ secrets.DOCKERHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-image.yml
```diff
@@ -18,9 +18,19 @@ jobs:
     - name: Get current tag
       id: vars
       run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+    - name: Extract Major and Minor Versions
+      id: version
+      run: |
+        FULL_VERSION=${{ steps.vars.outputs.tag }}
+        MAJOR_VERSION=$(echo $FULL_VERSION | cut -d '.' -f 1)
+        MINOR_VERSION=$(echo $FULL_VERSION | cut -d '.' -f 1,2)
```
timvisee, line 26: I think we should prefix a `v` here. Our full version and ref name have this too, such as `v1.2.3`.

```suggestion
        MAJOR_VERSION=v$(echo $FULL_VERSION | cut -d '.' -f 1)
        MINOR_VERSION=v$(echo $FULL_VERSION | cut -d '.' -f 1,2)
```

.github/workflows/docker-image.yml
```diff
@@ -18,9 +18,19 @@ jobs:
...
+        echo ::set-output name=major::${MAJOR_VERSION}
+        echo ::set-output name=minor::${MINOR_VERSION}
```
timvisee, line 28: Could you elaborate on what these do?

timvisee (changes requested):

.github/workflows/docker-image.yml
timvisee: Of course, this is automatically included sourced from the original tag. I accidentally tested this without a prefixed `v`. So this is all good! :smile:

.github/workflows/docker-image.yml
generall: As I understood, it writes values into github "variables"

generall: not trivial to test without actually running CI, though

yuvraj-wale: yes these lines set the 'major' and 'minor' output parameters, similar to how the 'tag' parameter is set in the 'Get current tag' step. These outputs are then used in subsequent steps of the workflow.

.github/workflows/docker-image.yml
```diff
@@ -18,9 +18,19 @@ jobs:
     - name: Get current tag
       id: vars
       run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+    - name: Extract Major and Minor Versions
+      id: version
+      run: |
+        FULL_VERSION=${{ steps.vars.outputs.tag }}
+        MAJOR_VERSION=$(echo $FULL_VERSION | cut -d '.' -f 1)
```
timvisee, line 25: Thank you for elaborating on the other comments!

Before I approve, I have a final thing to decide on. I wonder if we should remove major version tagging.

I don't think it makes much sense based on our current releases.

Maybe @generall has a good opinion on this?

yuvraj-wale: Yes makes sense, but I feel maintaining major tags is a standard practice & usually beneficial long-term. Your call, though.

mac-chaffee (changes requested):

PR #3475, opened 2024-01-29
Remove doctest usage

agourlay:
This PR migrates all our doctest to regular unit tests.

It is motivated by the fact that we currently spends several minutes on CI for executing only 3 tests.

## before

before

## after

after


timvisee (approved):
I like that doctests are a thing, but I guess moving it to regular unit tests is better to allow using nextest. :+1:

timvisee:
Removing 5:40 of compile time during CI is some fantastic gain!

PR #3476, opened 2024-01-29
Disable incremental compilation on CI

agourlay:
Let's have a look at the numbers first.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
 
 jobs:
   test:
--- a/.github/workflows/rust-lint.yml
+++ b/.github/workflows/rust-lint.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
 
 jobs:
   test:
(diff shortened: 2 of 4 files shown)


agourlay:
I don't see any gains, closing.

ben1009:
> I don't see any gains, closing.

FYI, actually, the env already set, https://github.com/qdrant/qdrant/actions/runs/7693838931/job/20963364815?pr=3476#step:2:6

agourlay:
@ben1009 Thanks for having a look at this PR :)

It seems the `rust-cache` action sets the env. variable automatically on [restore](https://github.com/Swatinem/rust-cache/blob/378c8285a4eaf12899d11bea686a763e906956af/src/restore.ts#L28)

ben1009:
> @ben1009 Thanks for having a look at this PR :)
> 
> It seems the `rust-cache` action sets the env. variable automatically on [restore](https://github.com/Swatinem/rust-cache/blob/378c8285a4eaf12899d11bea686a763e906956af/src/restore.ts#L28)

oh, i miss that part, first the env set by https://github.com/qdrant/qdrant/actions/runs/7693838931/job/20963364815?pr=3476#step:2:6, then stored to cache in https://github.com/qdrant/qdrant/actions/runs/7693838931/job/20963364815?pr=3476#step:39:19, if have another run will fetch from cache by https://github.com/qdrant/qdrant/actions/runs/7693838931/job/20963364815?pr=3476#step:4:10 as part of cache key

PR #3490, opened 2024-01-30
Move common dependencies to workspace

timvisee:
This explores the idea of moving dependencies we use often to our workspace as described [here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace).

It forces us to use the same dependency definition (version, features) in all our sub crates, because we only define it once. Before this change, we had wildly different definitions across all crates.

I manually selected these to subjectively be:
- used often
- having difficult feature flags
- having inconsistent configurations across our sub-crates
- requiring strict versions across multiple crates

With this PR I'd like to **request for comment**. What do we think about this? We can include more/less dependencies on the workspace level if we do desire.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,37 +42,37 @@ rusty-hook = "^0.11.2"
 
 
 [dependencies]
-
-parking_lot = { version = "0.12.1", features = ["deadlock_detection"], optional = true }
+parking_lot = { workspace = true, optional = true }
 
 thiserror = "1.0"
 log = "0.4"
 colored = "2"
-serde = { version = "~1.0", features = ["derive"] }
-serde_json = "~1.0"
+serde = { workspace = true }
+serde_json = { workspace = true }
 chrono = { version = "~0.4", features = ["serde"] }
 rand = "0.8.5"
-schemars = { version = "0.8.16", features = ["uuid1", "preserve_order", "chrono"] }
+schemars = { workspace = true }
 itertools = "0.12"
 anyhow = "1.0.79"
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -114,6 +113,23 @@ rstack-self = { version = "0.3.0", optional = true }
...
+serde = { version = "~1.0", features = ["derive"] }
+serde_cbor = { version = "0.11.2" }
+serde_json = "~1.0"
+tokio = { version = "~1.35", features = ["full"] }
+tokio-util = "0.7"
+tonic = { version = "0.9.2", features = ["gzip", "tls"] }
+tonic-reflection = "0.9.2"
+tracing = { version = "0.1", features = ["async-await"] }
+uuid = { version = "1.7", features = ["v4", "serde"] }
+
```
timvisee, line 132: The important bit is here. We define all these dependencies here only once now.

ffuugoo: I had no idea this was possible... 🤯

ffuugoo (approved):
Considering this is built into `cargo`, I'm cool with it. Didn't even know this is an option. 😂

coszio (approved):
The only downside I can think of is that there is no check for when we add package dependencies that are already in the `[workspace.dependencies]`. However, the benefit of using a single version of each dependency seems like a win to me.

timvisee:
Thanks for the insights! I'll merge. If we later find a good reason not to do it, we can always revert.

PR #3492, opened 2024-01-30
Fix: Update Qdrant configuration to use absolute paths

### LINKED ISSUE #3370: Debian package should use absolute Qdrant directory

shankariraja:
- Previously, Qdrant used the current working directory for configuration, storage, and snapshots, potentially causing issues when starting Qdrant from different directories.
- This commit addresses the problem by updating the configuration to use absolute paths:
  - Storage directory: `/var/lib/qdrant/storage`
  - Snapshots directory: `/var/lib/qdrant/snapshots`
  - temp_path: `/var/lib/qdrant/temp`
- This change ensures that Qdrant files are consistently located, regardless of the working directory from which it is started. The new paths also follow the conventions used by similar tools.

Fixes #3370

### DIFF
--- a/QUICK_START.md
+++ b/QUICK_START.md
@@ -29,12 +29,12 @@ And once you need a fine-grained setup, you can also define a storage path and c
 
 ```bash
 docker run -p 6333:6333 \
-    -v $(pwd)/path/to/data:/qdrant/storage \
+    -v $(pwd)/path/to/data:/var/lib/qdrant/storage \
     -v $(pwd)/path/to/custom_config.yaml:/qdrant/config/production.yaml \
     qdrant/qdrant
 ```
 
-- `/qdrant/storage` - is a place where Qdrant persists all your data.
+- `/var/lib/qdrant/storage` - is a place where Qdrant persists all your data.
   Make sure to mount it as a volume, otherwise docker will drop it with the container.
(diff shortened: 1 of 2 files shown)


timvisee (changes requested):
We should only use absolute paths in the Debian package, not in all other builds we make. This is described in #3370.

The rest should be kept as is, using the current working directory.

PR #2865, opened 2023-10-23
add `order_by` param in Scroll API

kwkr:
Hi @generall !
As an exercise to get to know the codebase better, I started implementing #1763 . I was able to hack together some working solution and wanted to confirm whether this is still a required feature and whether I got the requirements right.

So basically I added the `order_by` parameter to the scroll endpoint which allows scrolling through data by using a composite key as pointed out here:
```
Qdrant will read the index sorted by payload key, and will return results sorted by combination of (paylaod_value, id).
```

Currently, there are two possible ways to scroll through the data:
- order_by asc and min param, which will return the data sorted by payload in asc order and then by id.
- order_by desc and max param, which will return the data sorted by payload in desc order and then by id.

After getting the first results, the data can be further scrolled using the combination of payload_value and id.
When no offset is specified, the data is just returned starting from an item according to the provided min/max value.
The solution utilizes the Btrees of the underlying payload indexes so it should be efficient enough.

In both cases, the `filter` param is ignored.

Please let me know if this is still relevant so I can either polish the solution or drop it.

/claim #1763

PR #3493, opened 2024-01-30
`order_by` in scroll

coszio:
Supersedes #2865.

Adds `order_by` parameter to `scroll` API, which introduces this interface:
```json
"order_by": "my_key",

// or

"order_by": {
    "key": "my_key",
    "direction": "desc",
    "start_from": 70.0
}
```

## Implementation details

It takes advantage of the `BTree` from the numeric index to make a stream that can be filtered and ordered, instead of streaming from directly from the ids.

Limitation: it can only be used for fields that have a `range` index created for them.

Non-guarantee: ordering will only take into consideration the value from the `order_by` key. So ids won't be part of the ordering if the order_by value is the same. Because of this, offset is not allowed (and disabled in the response). 

For pagination, since offset is disabled, users can use `order_by.start_from` and a `must_not` with `has_id` condition to exclude the values that have already appeared for the same `start_from` value (see integration test)

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


generall (approved):

Dhwanil-Shah:
@coszio @generall Is it possible to support using "offset" along with "order_by" instead of relying on "start_from" since that requires an anchor point value of that key, I want to skip first N records to implement a simple pagination.
```
    # Desired usage
    page = 1
    page_size = 10
    records, _ = qdrant_client.scroll(
        collection_name="<collection-name>",
        limit=page_size,
        offset=(page * page_size), # 10 for 2nd page
        order_by="<integer-field-representing-order>",    # key field has duplicates, but since offset is used, no issue
        scroll_filter=filter,
    )
```

The current limitation forces me to use a continuation token rather than a simple page number that can be used to calculate offsets.
```
    # Current implementation
    page_size = 10
    continuation_token = 10.  # a random rank
    records, _ = qdrant_client.scroll(
        collection_name="<collection-name>",
        limit=page_size,
        order_by={
              "key": "<integer-field-representing-order>",   # key field has duplicates, since using "start_from" needs to be de-duplicated
              "start_from": continuation_token,
        },
        scroll_filter=filter,
    )
```
The issue with this is, the `key` field has duplicate values, so 2 pages can have overlapping data which is undesirable. So I need to add more processing for deduplication.

If this is not possible, can you give some context as to the technical challenges in doing so?

coszio:
Hi @Dhwanil-Shah 

The problem why `offset` is not available is precisely this. `offset` in pagination is just a point ID, such that qdrant can read only starting from such point, without having to "skip" points.

For `order_by` the case is different. Since one point can have multiple values. If only an ID offset is provided, qdrant won't know which of those values it refers to. 

The solution we recommend so far is to use `start_from`, in combination with a `must_not` + `has_id` to exclude the IDs that have already appeared.

PR #3495, opened 2024-01-31
Fix lints clippy 1.76

agourlay:
You know the drill :)


timvisee (approved):

timvisee:
We released 1.7.4 just in time! Otherwise this would be problematic again.

Thanks!

Issue #804, opened 2022-07-12
[rocksdb] Investigate compression scheme for RocksDB


agourlay:
**Is your feature request related to a problem? Please describe.**
A quick profiling session on a heavy write workload shows that Snappy can take up to 2% of CPU time.
![snappy](https://user-images.githubusercontent.com/606963/178513624-622407d2-c69f-49d8-83e8-58237b5a3a02.png)

The official doc says:

```
Use options.compression to specify the compression to use. 
By default it is Snappy. We believe LZ4 is almost always better than Snappy.
```
https://github.com/facebook/rocksdb/wiki/Compression

**Describe the solution you'd like**
Investigate if it can be beneficial to switch to another compression scheme.

e.g.

```rust
options.set_compression_type(DBCompressionType::Lz4);
```

generall:
Criteria for this issue:

- It is possible to make switch to a new compression backward compatible
- How well it compresses datasets with/without payload: size, write/read speed benches

thomas-k-cameron:
I have done something similar in the past.

Here is my take:
- zstd should be faster based on some benchmarks
- Also, the official [wiki](https://github.com/facebook/rocksdb/wiki/Compression) recommends zstd

Let me try it.

thomas-k-cameron:
I bechmarked it.

Here is a table of results.

I used the benchmarking script in benches directory. 

I'm don't know how scientific it is but I think it gives you a rough idea.

To summarize, upsert performance is `zstd` is on par with `snappy` but `zstd` makes it significantly faster when it comes to search performance.

| compression | benchmark          | grpc_req_duration | iteration duration | iterations   |
| ----------- | ------------------ | ----------------- | ------------------ | ------------ |
| snappy      | run_grpc_upsert.sh | avg=21.41ms       | avg=63.29ms        | 449.381818/s |
| zstd        | run_grpc_upsert.sh | avg=20.73ms       | avg=62.6ms         | 427.646846/s |
| snappy      | run_grpc_search.sh | avg=38.5ms        | avg=60.26ms        | 248.779121/s |
| zstd        | run_grpc_search.sh | avg=27.82ms       | avg=45.5ms         | 329.470261/s |

If you want to be more adventurous, I think you can consider adopting something else like this one.
https://github.com/bytedance/terarkdb

thomas-k-cameron:
So I did some benchmarking using scripts I found on `bench/services`. 
[...]

PR #3497, opened 2024-01-31
Text index immutable state without documents data

IvanPleshkov:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
@@ -5,6 +5,7 @@ use serde::{Deserialize, Serialize};
 
 use super::posting_list::PostingList;
 use super::postings_iterator::intersect_postings_iterator;
+use crate::common::operation_error::{OperationError, OperationResult};
 use crate::index::field_index::{CardinalityEstimation, PayloadBlockCondition, PrimaryCondition};
 use crate::types::{FieldCondition, Match, MatchText, PayloadKeyType};
 
@@ -56,28 +57,33 @@ impl ParsedQuery {
     }
 }
 
-#[derive(Default)]
-pub struct InvertedIndex {
-    postings: Vec<Option<PostingList>>,
-    pub vocab: HashMap<String, TokenId>,
-    pub point_to_docs: Vec<Option<Document>>,
-    pub points_count: usize,
+pub enum InvertedIndex {
+    Mutable(MutableInvertedIndex),
+    Immutable(ImmutableInvertedIndex),
 }
 
 impl InvertedIndex {
-    pub fn new() -> InvertedIndex {
-        Default::default()
+    pub fn new(is_appendable: bool) -> InvertedIndex {
+        if is_appendable {
+            InvertedIndex::Mutable(MutableInvertedIndex::default())
+        } else {
+            InvertedIndex::Immutable(ImmutableInvertedIndex::default())
+        }
     }
 
     pub fn document_from_tokens(&mut self, tokens: &BTreeSet<String>) -> Document {
+        let vocab = match self {
+            InvertedIndex::Mutable(index) => &mut index.vocab,
+            InvertedIndex::Immutable(index) => &mut index.vocab,
+        };
         let mut document_tokens = vec![];
         for token in tokens {
             // check if in vocab
-            let vocab_idx = match self.vocab.get(token) {
+            let vocab_idx = match vocab.get(token) {
                 Some(&idx) => idx,
                 None => {
-                    let next_token_id = self.vocab.len() as TokenId;
-                    self.vocab.insert(token.to_string(), next_token_id);
+                    let next_token_id = vocab.len() as TokenId;
+                    vocab.insert(token.to_string(), next_token_id);
                     next_token_id
                 }
             };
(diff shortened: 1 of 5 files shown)


lib/segment/src/index/field_index/full_text_index/postings_iterator.rs
```diff
@@ -4,6 +4,7 @@ use super::posting_list::PostingList;
 
 pub fn intersect_postings_iterator<'a>(
     mut postings: Vec<&'a PostingList>,
+    filter: impl Fn(PointOffsetType) -> bool + 'a,
```
IvanPleshkov, line 7: In the immutable state, we don't delete anything from the posting list. Provide here the deleted filter which is empty for mutable state and deleted flag chacking for immutable one

lib/segment/src/index/field_index/full_text_index/postings_iterator.rs
```diff
@@ -14,7 +15,7 @@ pub fn intersect_postings_iterator<'a>(
     let smallest_posting = postings.remove(smallest_posting_idx);
 
     let and_iter = smallest_posting
-        .iter()
+        .iter(filter)
         .filter(move |doc_id| postings.iter().all(|posting| posting.contains(doc_id)));
```
IvanPleshkov, line 19: I didn't add deleted condition in `filter` because we calculate the set intersection here. And all deleted id's are filtered by `iter(filter)`

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -56,28 +57,40 @@ impl ParsedQuery {
     }
 }
 
-#[derive(Default)]
-pub struct InvertedIndex {
-    postings: Vec<Option<PostingList>>,
-    pub vocab: HashMap<String, TokenId>,
-    pub point_to_docs: Vec<Option<Document>>,
-    pub points_count: usize,
+pub enum InvertedIndex {
```
IvanPleshkov, line 60: Defifinion of new Inverted index. Here I aggregate all business logic (estimations, filtering, payload blocks etc). And In mutable/immutable state I store only data and how to access it

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -253,4 +275,215 @@ impl InvertedIndex {
                 }),
         )
     }
+
+    pub fn build_index(
```
IvanPleshkov, line 279: Index loading. Iterate over rocksdb and build inverted index. Because immutable state does not support points addition, I load data as mutable index and then convert to immutable when needed

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -56,28 +57,40 @@ impl ParsedQuery {
...
 
     pub fn document_from_tokens(&mut self, tokens: &BTreeSet<String>) -> Document {
+        let vocab = match self {
+            InvertedIndex::Mutable(index) => &mut index.vocab,
+            InvertedIndex::Immutable(index) => &mut index.vocab,
+        };
+        Self::document_from_tokens_impl(vocab, tokens)
+    }
+
+    fn document_from_tokens_impl(
```
IvanPleshkov, line 82: Don't like this private method but I use it in two places - in `document_from_tokens` and in `build_index`

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -253,4 +275,215 @@ impl InvertedIndex {
...
+                if let Some(posting_list) = &self.postings[query_token.unwrap() as usize] {
+                    posting_list.contains(&point_id)
+                } else {
+                    false
+                }
+            })
+    }
+}
+
+impl From<MutableInvertedIndex> for ImmutableInvertedIndex {
```
IvanPleshkov, line 476: I use this convention in index construction. Commented above

lib/segment/src/index/query_optimization/condition_converter.rs
```diff
@@ -274,9 +274,7 @@ pub fn get_match_checkers(index: &FieldIndex, cond_match: Match) -> Option<Condi
             FieldIndex::FullTextIndex(full_text_index) => {
                 let parsed_query = full_text_index.parse_query(&text);
                 Some(Box::new(move |point_id: PointOffsetType| {
-                    full_text_index
-                        .get_doc(point_id)
-                        .map_or(false, |doc| parsed_query.check_match(doc))
+                    full_text_index.check_match(&parsed_query, point_id)
```
IvanPleshkov, line 277: Here is an affected point. Previously we checked the match between the two documents. Now, we use the postings list to do this check. It is less efficient, here is random access in the posting list for each token

lib/segment/src/index/field_index/full_text_index/text_index.rs
```diff
@@ -47,7 +47,7 @@ impl FullTextIndex {
         })
     }
 
-    fn deserialize_document(data: &[u8], index: &mut InvertedIndex) -> OperationResult<Document> {
+    fn deserialize_document(data: &[u8]) -> OperationResult<BTreeSet<String>> {
```
IvanPleshkov, line 50: deserialization in my opinion has to have the same type as serialization. changed to type of serialization. Also `BTreeSet<_>` conversion avoids usage of `InvertedIndex` here and it makes the rusts borrow checker happier

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -154,21 +144,45 @@ impl InvertedIndex {
...
+                // in case of immutable index, deleted documents are still in the postings
+                let filter = move |idx| {
+                    matches!(
+                        index.point_documents_tokens.get(idx as usize),
+                        Some(Some(_))
+                    )
+                };
+                intersect_postings_iterator(postings, filter)
+            }
+        }
```
timvisee, line 163: Can we remove all of this logic into the (im)mutable index themselves, so we don't have to explicitly filter here?

It looks like the immutable structure already has al information for this, the postings and document tokens. Or am I missing an important detail here?

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -154,21 +144,45 @@ impl InvertedIndex {
...
         condition: &FieldCondition,
     ) -> CardinalityEstimation {
+        let points_count = match self {
+            InvertedIndex::Mutable(index) => index.points_count,
+            InvertedIndex::Immutable(index) => index.points_count,
+        };
+        let index_postings = match self {
+            InvertedIndex::Mutable(index) => &index.postings,
+            InvertedIndex::Immutable(index) => &index.postings,
+        };
```
timvisee, line 178: We can do a single match here that returns a tuple, like:

```rust
let (points_count, index_postings) = match self {
    InvertedIndex::Mutable(index) => (index.points_count, &index.postings),
    InvertedIndex::Immutable(index) => (index.points_count, &index.postings),
};
```

The same goes for some lines below.

If you like the current a lot better though, feel free to keep it as-is.

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -253,4 +275,215 @@ impl InvertedIndex {
...
+            postings: index.postings,
+            vocab: index.vocab,
+            point_documents_tokens: index
+                .point_to_docs
+                .iter()
+                .map(|doc| doc.as_ref().map(|doc| doc.len()))
+                .collect(),
+            points_count: index.points_count,
+        }
+    }
```
timvisee, line 488: Would it make sense here to to call [`shrink_to_fit`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.shrink_to_fit) on some of the types here, since we won't mutate them anyway?

Thinking about [this](https://github.com/qdrant/qdrant/issues/804#issuecomment-1931534082) here.

generall:
I don't have any additional comments to what @timvisee have already suggested

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
IvanPleshkov: It makes sense, fixed

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
IvanPleshkov: You are right, common logic between mutable and immutable states is only `None` conditions. Moved to separate functions, it looks better. Fixed

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -203,9 +177,9 @@ impl InvertedIndex {
         } else {
             let expected_frac: f64 = postings
                 .iter()
-                .map(|posting| posting.len() as f64 / self.points_count as f64)
+                .map(|posting| posting.len() as f64 / points_count as f64)
```
timvisee, line 180: Should we ensure here that `points_count` is not `0`? Maybe we can do an early return somewhere.

IvanPleshkov: That's a brilliant finding. Because previously we was sure that `points_count` is not zero (because we checked amount of postings list). But not we don't delete anything from inverted index and may meet the situation where postings list is not empty but `points_count` is zero. Fixed using additional check for `points_count != 0`

timvisee: Thanks for adding the check!

timvisee (approved):

PR #3498, opened 2024-01-31
allow stopping segment loading

### LINKED ISSUE #3140: Stoppable segment loading
Segment loading does not support stop flag. Loading may take a long time so I propose adding stop flag into segment loading and subsystems loading (vector storage, vector index, ect)

JojiiOfficial:
fixes #3140 

### All Submissions:

### New Feature Submissions:

Interestingly rustc panics when running `cargo clippy -all --all-features` on dev and this branch (on master it doesn't). Is this only a thing on my machine?

### DIFF
--- a/lib/collection/src/shards/local_shard.rs
+++ b/lib/collection/src/shards/local_shard.rs
@@ -2,6 +2,7 @@ use std::collections::{BTreeSet, HashMap};
 use std::mem::size_of;
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
+use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 use std::thread;
 use std::time::{Duration, Instant};
@@ -204,7 +205,7 @@ impl LocalShard {
                 thread::Builder::new()
                     .name(format!("shard-load-{collection_id}-{id}"))
                     .spawn(move || {
-                        let mut res = load_segment(&segments_path)?;
+                        let mut res = load_segment(&segments_path, &AtomicBool::new(false))?;
                         if let Some(segment) = &mut res {
                             segment.check_consistency_and_repair()?;
                         } else {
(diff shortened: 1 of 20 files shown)


lib/segment/src/segment_constructor/segment_constructor_base.rs
timvisee, line 1: Shouldn't we check for cancellation a lot more in the create/load functions, to quit as soon as possible? We only do this in two places now.

lib/collection/src/shards/local_shard.rs
```diff
@@ -204,7 +205,7 @@ impl LocalShard {
                 thread::Builder::new()
                     .name(format!("shard-load-{collection_id}-{id}"))
                     .spawn(move || {
-                        let mut res = load_segment(&segments_path)?;
+                        let mut res = load_segment(&segments_path, &AtomicBool::new(false))?;
```
IvanPleshkov, line 208: It seems changes are disabled because here we don't use a stop flag. Why did you decide to add placeholder here instead of actual working?

IvanPleshkov (changes requested):

lib/collection/src/shards/local_shard.rs
JojiiOfficial: From the issue comments it sounded to me that the focus of this segment loading stopping is the segment optimizer that uses this load_segment function indirectly which can be cancelled, and not LocalShard loading. From where exactly do you expect local shards getting cancelled during loading?

generall: @IvanPleshkov https://github.com/qdrant/qdrant/issues/3140#issuecomment-1877541306

generall: The idea was to cancel load segments during the optimization, not during the load of shards.

IvanPleshkov (approved):

Issue #3471, opened 2024-01-26
Tracking issue: Issues API


coszio:
This is a tracking issue for following the development of a new API that aims to report some configuration problems to the users:

- [ ] Internal `issues` crate: #3375 
- [ ] Create endpoint (http only): WIP
- [ ] Report missing payload index
- [ ] Report too many collections

generall:
>  Create endpoint (http only): WIP

Might it be part of telemetry?

coszio:
Probably best to make dedicated API to keep it as a beta feature

PR #3502, opened 2024-01-31
Issues API: Interface: get issues

coszio:
Needs #3375. Tracked by #3471.

Adds a dedicated `GET /issues` endpoint, only in REST to keep this functionality as a beta feature. Probably later we can include it as a part of telemetry. 

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


openapi/openapi-service.ytt.yaml
```diff
@@ -127,12 +127,21 @@ paths:
...
           content:
             text/plain:
               schema:
                 type: string
                 example: healthz check passed
-        '4XX':
+        "4XX":
           description: error
+
+  /issues:
```
timvisee, line 140: Did I get this right; we plan to remove this endpoint before release, and it is just here for testing purposes?

If so, we might not have to add this definition. Then, if we accidentally release with it, it won't be publicly exposed/documented.

timvisee (approved):
Sure!

It isn't entirely clear to me though whether we want to keep this endpoint. See my review comment below.

openapi/openapi-service.ytt.yaml
coszio: It is not just for testing, it is just marked as a beta feature to advise people to not depend on this too critically, since it might change or get added to `/telemetry` once it is stable enough.

I think it is useful to document it, to make the development of the webUI feature easier

timvisee: Check! In that case I agree, keep it here.

timvisee (approved):

PR #3506, opened 2024-02-01
Add `ClockMap` type to track versions of update operations in the local shard WAL

ffuugoo:
This PR *will* add `ClockMap` type.

Currently based on #3426, but pointing to `dev`, cause you can't create an empty PR. 😂

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/operations/mod.rs
+++ b/lib/collection/src/operations/mod.rs
@@ -67,9 +67,9 @@ impl From<CollectionUpdateOperations> for OperationWithClockTag {
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
 pub struct ClockTag {
-    peer_id: PeerId,
-    clock_id: u32,
-    clock_tick: u64,
+    pub peer_id: PeerId,
+    pub clock_id: u32,
+    pub clock_tick: u64,
 }
 
 impl ClockTag {
--- a/lib/collection/src/shards/replica_set/clock_set.rs
+++ b/lib/collection/src/shards/replica_set/clock_set.rs
@@ -1,0 +1,91 @@
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::Arc;
+
+#[derive(Clone, Debug, Default)]
+pub struct ClockSet {
+    clocks: Vec<Arc<Clock>>,
+}
+
+impl ClockSet {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn get_clock(&mut self) -> ClockGuard {
+        for (id, clock) in self.clocks.iter().enumerate() {
+            if clock.lock() {
+                return ClockGuard::new(id, clock.clone());
+            }
+        }
+
+        let id = self.clocks.len();
+        let clock = Arc::new(Clock::new_locked());
+
+        self.clocks.push(clock.clone());
+
+        ClockGuard::new(id, clock)
+    }
+}
+
+#[derive(Debug)]
+pub struct ClockGuard {
+    id: usize,
+    clock: Arc<Clock>,
+}
+
+impl ClockGuard {
+    fn new(id: usize, clock: Arc<Clock>) -> Self {
+        Self { id, clock }
+    }
+
+    pub fn id(&self) -> usize {
+        self.id
+    }
+
+    pub fn tick_once(&mut self) -> u64 {
+        self.clock.tick_once()
+    }
+
+    pub fn release(self) {
+        // Do not call `Clock::release` explicitly!
+        //
+        // `Drop` will be called when `self` goes out of scope at the end of this method
+        // and it will call `Clock::release`.
+        //
+        // If call `Clock::release` explicitly, then `ClockGuard::release` will call
+        // `Clock::release` *twice*! (Which is a bug! :D)
+    }
+}
+
+impl Drop for ClockGuard {
+    fn drop(&mut self) {
+        self.clock.release();
+    }
+}
+
+#[derive(Debug)]
+struct Clock {
+    clock: AtomicU64,
+    available: AtomicBool,
+}
+
+impl Clock {
+    pub fn new_locked() -> Self {
+        Self {
+            clock: AtomicU64::new(0),
+            available: AtomicBool::new(false),
+        }
+    }
+
+    pub fn tick_once(&self) -> u64 {
+        self.clock.fetch_add(1, Ordering::Relaxed) + 1
+    }
+
+    pub fn lock(&self) -> bool {
+        self.available.swap(false, Ordering::Relaxed)
+    }
+
+    pub fn release(&self) {
+        self.available.store(true, Ordering::Relaxed);
+    }
+}
--- a/lib/collection/src/shards/replica_set/clock_set.rs
+++ b/lib/collection/src/shards/replica_set/clock_set.rs
@@ -46,6 +46,10 @@ impl ClockGuard {
         self.clock.tick_once()
     }
 
+    pub fn advance_to(&mut self, tick: u64) -> u64 {
+        self.clock.advance_to(tick)
+    }
+
     pub fn release(self) {
         // Do not call `Clock::release` explicitly!
         //
@@ -79,6 +83,10 @@ impl Clock {
 
     pub fn tick_once(&self) -> u64 {
         self.clock.fetch_add(1, Ordering::Relaxed) + 1
+    }
+
+    pub fn advance_to(&self, tick: u64) -> u64 {
+        self.clock.fetch_max(tick, Ordering::Relaxed)
     }
 
     pub fn lock(&self) -> bool {
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -1,3 +1,4 @@
+mod clock_set;
 mod execute_read_operation;
 mod locally_disabled_peers;
 mod read_ops;
@@ -27,6 +28,7 @@ use crate::operations::types::{CollectionError, CollectionResult};
 use crate::save_on_disk::SaveOnDisk;
 use crate::shards::channel_service::ChannelService;
 use crate::shards::dummy_shard::DummyShard;
+use crate::shards::replica_set::clock_set::ClockSet;
 use crate::shards::shard::{PeerId, Shard, ShardId};
 use crate::shards::shard_config::ShardConfig;
 use crate::shards::telemetry::ReplicaSetTelemetry;
@@ -94,6 +96,7 @@ pub struct ShardReplicaSet {
     optimizer_cpu_budget: CpuBudget,
     /// Lock to serialized write operations on the replicaset when a write ordering is used.
     write_ordering_lock: Mutex<()>,
+    clock_set: Mutex<ClockSet>,
 }
 
 pub type AbortShardTransfer = Arc<dyn Fn(ShardTransfer, &str) + Send + Sync>;
@@ -181,6 +184,7 @@ impl ShardReplicaSet {
             search_runtime,
             optimizer_cpu_budget,
             write_ordering_lock: Mutex::new(()),
+            clock_set: Mutex::new(ClockSet::new()),
         })
     }
 
@@ -291,6 +295,7 @@ impl ShardReplicaSet {
             search_runtime,
             optimizer_cpu_budget,
             write_ordering_lock: Mutex::new(()),
+            clock_set: Mutex::new(ClockSet::new()),
         };
 
         if local_load_failure && replica_set.active_remote_shards().await.is_empty() {
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -1,3 +1,4 @@
+use std::cmp;
 use std::ops::Deref as _;
 use std::time::Duration;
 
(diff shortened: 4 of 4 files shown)


lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub fn new(tick: u64) -> Self {
+        Self {
+            clock: AtomicU64::new(tick),
+        }
+    }
+
+    pub fn advance_to(&self, new_tick: u64) -> u64 {
+        self.clock.fetch_max(new_tick, Ordering::Relaxed)
+    }
+}
```
timvisee, line 101: Can't we reuse the same `Clock` type here, to prevent name collisions?

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -69,6 +73,7 @@ pub struct LocalShard {
     pub(super) path: PathBuf,
     pub(super) optimizers: Arc<Vec<Arc<Optimizer>>>,
     pub(super) optimizers_log: Arc<ParkingMutex<TrackerLog>>,
+    clock_map: Arc<Mutex<ClockMap>>,
```
timvisee, line 76: I'd prefer a name there that describes what we use it for, not a name that matches the type definition.

```suggestion
    /// Clock map holding the last seen clock tick for all clocks we know about
    last_seen_clocks: Arc<Mutex<ClockMap>>,
```

or something different.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -0,0 +1,121 @@
...
+            clock.advance_to(new_tick)
+        } else {
+            self.clocks.insert(clock_id, Clock::new(new_tick));
+            0
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Deserialize, Serialize)]
+struct ClockId {
```
timvisee, line 72: I don't like this name, because we use it for two different things:
- in some places it only defines a clock index on _any_ peer
- in this place it defines a clock index on a specific peer

What about naming this `PeerClockId` or something similar?

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -245,11 +245,31 @@ impl ShardOperation for LocalShard {
...
+                    .advance_clock_and_correct_tag(clock_tag);
+
+                if new_tick > 0 && new_tick <= old_tick {
+                    return Ok(UpdateResult {
+                        operation_id: None,
+                        status: UpdateStatus::Acknowledged,
+                        clock_tag: Some(*clock_tag),
+                    });
+                }
+            }
```
timvisee, line 268: Could we maybe add a comment here, on the outer and inner if-statement, to describe what this does.

I assume the inner if-statement rejects operations with old ticks. But then I find `new_tick <= old_tick` a bit confusing. Would it maybe help to rename `old_tick` to `last_seen_tick` or `known_tick`?

lib/collection/src/update_handler.rs
```diff
@@ -634,6 +647,20 @@ impl UpdateHandler {
...
 
+            let clock_map_store_future = async {
+                let clock_map_tmp_file = tempfile::NamedTempFile::new()?;
+                let clock_map_tmp_path = clock_map_tmp_file.into_temp_path();
+
+                clock_map.lock().await.store(&clock_map_tmp_path)?;
+                tokio::fs::copy(&clock_map_tmp_path, &clock_map_path).await?;
+
+                CollectionResult::Ok(())
+            };
```
timvisee, line 658: I'd prefer to use the `AtomicFile` wrapper here instead, because we also use this in other places.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    pub fn advance_clock_and_correct_tag(&mut self, clock_tag: &mut ClockTag) -> u64 {
+        let prev_tick = self.advance_clock(clock_tag);
+
+        if clock_tag.clock_tick == 0 {
+            clock_tag.clock_tick = prev_tick;
+        }
+
+        prev_tick
+    }
```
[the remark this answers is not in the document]
ffuugoo: See [here]. I don't really like it, but I did not yet figure out how to restructure the API. Maybe just return `bool` from `advance_clock_and_correct_tag`?

[here]: https://github.com/qdrant/qdrant/pull/3506/files#diff-0d032845d818bc68dc94a1e996b506d56b8735e2b93d03ce3bdcf1222931b28aR253-R267

lib/collection/src/shards/local_shard/clock_map.rs
ffuugoo: We can't reuse them, cause they work differently. Like, `advance_to` is implemented differently on `clock_set::Clock` and `clock_map::Clock`.

Also, these are private types, so I don't think any kind of name collision is possible. %)

timvisee: Is there no way to unify them?

> Also, these are private types, so I don't think any kind of name collision is possible. %)

I meant that we have the same name for a similar-but-not-the-same type, which is very confusing to me.

lib/collection/src/shards/local_shard/clock_map.rs
ffuugoo: Sure. I just didn't want to spend too much time figuring out a proper name. I'd go for `ClockMapKey`, I think.

timvisee: > I'd go for ClockMapKey, I think.

Good one!

lib/collection/src/shards/local_shard/shard_ops.rs
ffuugoo: Yeah, sure. It's still a WIP, cause I don't like it either and want to come up with a better API/structure for this check.

lib/collection/src/update_handler.rs
ffuugoo: Totally forgot we have it. Will do.

lib/collection/src/shards/local_shard/clock_map.rs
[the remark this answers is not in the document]
timvisee: If we just want to check whether it was modified in-place, then yes, a boolean would likely be better.

But I must admit, how this works is currently not entirely clear to me yet.

lib/collection/src/shards/local_shard/clock_map.rs
ffuugoo: > Is there no way to unify them?
Maybe there is, but I don't want to, tbh.

> similar-but-not-the-same type
Yeah, exactly. In my mind they are, logically, two different types. _Somewhat_ similar, sure, but even the main method (i.e., `advance_to`) is different for both of them.

And regarding type names... I dunno. I personally prefer this approach of using simple, clear names (even if the same name is used for multiple types) and just scope them with modules.

Like, I actually *don't* like when we name our types `CollectionError` or `CollectionResult`. IMO using `collection::Error`/`collection::Result` is just better. It's the same amount of typing (or, two additional symbols for `::`), but *inside* the `collection` module you just use `Error` instead of repeating `Collection*` all the time and this way you import less stuff (cause you just do `use collection` and then do `collection::Error` instead of `use collection::CollectionError` and then also `CollectionError`).

Does this make sense?

lib/collection/src/shards/local_shard/mod.rs
ffuugoo: Sure, works for me.

lib/collection/src/shards/local_shard/clock_map.rs
timvisee: I like short names to. But with this approach you always need to know or define your current context, and you need to know what file/module your cursor is currently in. That has always been a problem for me personally, requiring me to jump/look at things a lot more when working with types like this. With unique names I don't have to think about this at all.

I have a similar problem with all the Mutex/RwLock types we're using interchangeably, to me it is extremely annoying.

But whatever, lets see if someone else has an opinion on this.

ffuugoo: > But with this approach you always need to know or define your current context, and you need to know what file/module your cursor is currently in.

Is it a problem even in case of `ClockSet`/`ClockMap`? Like, it's a private type used only as part of `ClockSet`/`ClockMap` implementation and both these modules are 100 something lines of code.

I'm not completely against renaming these to, e.g., `ClockSetClock`/`ClockMapClock`... but, tbh, it feels so "extra", given these are private types used in a very limited context. Just rubs me the wrong way. But I'm ok renaming them, if it makes it easier for everyone else to work with. 😭

> I have a similar problem with all the Mutex/RwLock types...

Same. I hate this.

IMO, typing module name explicitly (e.g., `parking_lot::Mutex` and `tokio::sync::Mutex`) solves the issue in the best way.

timvisee: We agreed to keep it like this for now. If I come up with a good example on how we might unify this, we can make the necessary changes in the future.

lib/collection/src/shards/local_shard/clock_map.rs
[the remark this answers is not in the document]
ffuugoo: > how this works is currently not entirely clear to me yet

Because it doesn't. I think there's a bug and it's just wrong. 😅

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub fn advance_clock(&mut self, clock_tag: &ClockTag) -> u64 {
+        let clock_id = ClockId::from_tag(clock_tag);
+        let new_tick = clock_tag.clock_tick;
+
+        if let Some(clock) = self.clocks.get(&clock_id) {
+            clock.advance_to(new_tick)
+        } else {
+            self.clocks.insert(clock_id, Clock::new(new_tick));
+            0
+        }
```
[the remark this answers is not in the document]
ffuugoo: Done.

lib/collection/src/shards/local_shard/clock_map.rs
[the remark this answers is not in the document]
ffuugoo: Ok, so, I've refactored `advance_clock*` methods... but it still feels convoluted. I'm open to suggestions how to improve it. :/

lib/collection/src/shards/local_shard/clock_map.rs
ffuugoo: Done. Went just with the `Key`, cause it's a private type used in a very small context, but can rename to `ClockMapKey` if you prefer it that way.

lib/collection/src/shards/local_shard/shard_ops.rs
ffuugoo: Reworked it, it's still a mess, but so far I'm not sure how to improve it much.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -0,0 +1,148 @@
...
+        // And we also *always* accept all operations with `clock_tick = 0` and *always* update their clock tags.
+
+        let operation_accepted = clock_updated || clock_tag.clock_tick == 0;
+        let update_tag = !clock_updated || clock_tag.clock_tick == 0;
+
+        let (operation_accepted, update_tag) = if clock_tag.clock_tick > 0 {
+            (clock_updated, !clock_updated)
+        } else {
+            (true, true)
+        };
```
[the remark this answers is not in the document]
ffuugoo: > I prefer to put "main" case in the first branch, but clock_tick == 0 is a special case here, but it's also super unobvious if you put clock_tick > 0 into conditional in previous example

I guess that kinda resolves it for me (I don't like both 2nd and 3rd example, so I'd go with the first), but I still want to hear what you guys think.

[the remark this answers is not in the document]
timvisee: I think I like the first one best.

That is because we check and assign the two separately, rather than in a tuple. For me that is a bit easier to understand.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        operation_accepted
+    }
+
+    pub fn advance_clock(&mut self, clock_tag: &ClockTag) {
+        self.advance_clock_impl(clock_tag);
+    }
+
+    fn advance_clock_impl(&mut self, clock_tag: &ClockTag) -> (bool, u64) {
```
timvisee, line 79: Do we want a `must_use` here, just to be sure? Or maybe on the other function(s) that actually expose this return value.

lib/collection/src/shards/local_shard/shard_ops.rs
timvisee: Better now. The `let accepted` clarifies a lot.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        // We "accept" an operation, if it has `clock_tick` that is "newer" than `current_tick` in `ClockMap`
+        // (e.g., if `advance_clock_impl` *updated* the clock and returned `clock_updated = true`).
+        //
+        // If we "reject" an operation (because it has `clock_tick` that is "older" than `current_tick` in `ClockMap`),
+        // we have to update its clock tag with `current_tick`, so that it can be "echoed" back to the node.
+        //
+        // And we also *always* accept all operations with `clock_tick = 0` and *always* update their clock tags.
+
+        let operation_accepted = clock_updated || clock_tag.clock_tick == 0;
```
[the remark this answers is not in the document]
ffuugoo: Yeah, I wanted to show two variants. I'll "fixup" it now. :D

lib/collection/src/shards/local_shard/clock_map.rs
ffuugoo: I don't think it makes much sense on the `impl` one, but it definitely does on the `and_correct_tag`.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+impl Clock {
+    pub fn new(tick: u64) -> Self {
+        Self {
+            current_tick: tick.into(),
+        }
+    }
+
+    pub fn advance_to(&self, new_tick: u64) -> (bool, u64) {
```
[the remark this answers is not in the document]
ffuugoo: Didn't have the time to "formulate" the documentation, thanks.

lib/collection/src/shards/local_shard/clock_map.rs
timvisee: I agree. :+1:

lib/collection/src/shards/local_shard/clock_map.rs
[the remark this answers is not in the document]
ffuugoo: Added documentation to main clock map methods. It's a bit awkward, I'm open to suggestions how to simplify/improve it.

lib/collection/src/shards/local_shard/clock_map.rs
ffuugoo: Added `must_use` to `impl` method as well in the end, just in case. 😂

lib/collection/src/shards/local_shard/clock_map.rs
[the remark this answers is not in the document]
ffuugoo: Remove second example from the code.

lib/collection/src/shards/local_shard/clock_map.rs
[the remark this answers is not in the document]
ffuugoo: I think I've done my best on it by now...

lib/collection/src/update_handler.rs
ffuugoo: Done.

timvisee (approved):

timvisee:
fd187cfbbf8155fc6aa4f3eed6d16fbba2fcc532 now disables the force flag on forward proxy operations and (temporarily) disables rejecting operations with old clock values.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -74,15 +74,15 @@ impl ClockMap {
         //
         // And we also *always* accept all operations with `clock_tick = 0` and *always* update their clock tags.
 
-        let operation_accepted = clock_updated || clock_tag.clock_tick == 0 || clock_tag.force;
+        let _operation_accepted = clock_updated || clock_tag.clock_tick == 0 || clock_tag.force;
         let update_tag = (!clock_updated || clock_tag.clock_tick == 0) && !clock_tag.force;
 
-        if operation_accepted && !update_tag {
+        if _operation_accepted && !update_tag {
             log::trace!("Accepting clock tag {clock_tag:?} (current tick: {current_tick})");
```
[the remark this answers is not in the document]
timvisee: Resolved in <https://github.com/qdrant/qdrant/pull/3567>

Issue #3477, opened 2024-01-29
Tracking issue: shard diff transfer - WAL delta


timvisee:
This issue tracks progress on the feature for just sending the diff as part of shard transfers. This will be based on the WAL delta between two shard replicas. It should significantly reduce the data to transfer in cases where a diff can be resolved.

### Tasks for 1.8

- [x] https://github.com/qdrant/qdrant/pull/3408
- [x] https://github.com/qdrant/qdrant/pull/3426
- [x] https://github.com/qdrant/qdrant/pull/3506
- [x] https://github.com/qdrant/qdrant/pull/3567
- [x] https://github.com/qdrant/qdrant/pull/3573
- [x] #3572
- [x] https://github.com/qdrant/qdrant/pull/3510
- [x] https://github.com/qdrant/qdrant/pull/3550
- [x] https://github.com/qdrant/qdrant/pull/3551
- [x] https://github.com/qdrant/qdrant/pull/3571
- [x] https://github.com/qdrant/qdrant/pull/3588
- [x] https://github.com/qdrant/qdrant/pull/3590
- [x] https://github.com/qdrant/qdrant/pull/3591
- [x] https://github.com/qdrant/qdrant/pull/3593
- [x] https://github.com/qdrant/qdrant/pull/3620
- [x] https://github.com/qdrant/qdrant/pull/3631
- [x] https://github.com/qdrant/qdrant/pull/3610
- [x] https://github.com/qdrant/qdrant/pull/3615
- [x] https://github.com/qdrant/qdrant/pull/3618
- [x] https://github.com/qdrant/qdrant/pull/3642
- [x] https://github.com/qdrant/qdrant/pull/3660
- [x] https://github.com/qdrant/qdrant/pull/3662
- [x] https://github.com/qdrant/qdrant/pull/3659
- [x] https://github.com/qdrant/qdrant/pull/3661
- [x] https://github.com/qdrant/qdrant/pull/3670
- [x] https://github.com/qdrant/qdrant/pull/3671
- [x] https://github.com/qdrant/qdrant/pull/3509
- [x] #3645
- [x] https://github.com/qdrant/qdrant/pull/3679
- [x] #3677
- [x] https://github.com/qdrant/qdrant/pull/3719
- [x] https://github.com/qdrant/qdrant/pull/3725
- [x] #3687
- [x] https://github.com/qdrant/qdrant/pull/3703
- [x] https://github.com/qdrant/qdrant/pull/3728
[...]

PR #3509, opened 2024-02-01
Shard diff transfer integration

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/3551>, <https://github.com/qdrant/qdrant/pull/3571>

This adds a new `WalDelta` shard transfer method to allow shard diff transfers. This variant remains unused for now and is hidden by default.

It adds a `transfer_wal_delta` function to actually drive such transfer from a source node.

The method currently utilizes a queue proxy shard because it already has all the necessary bits and bots implemented for transferring a WAL including all new incoming updates.

### Tasks
- [x] Resolve all TODOs
- [x] Prevent physical WAL from being truncated if we diff, even if it has already been truncated of the logical WAL
      We can set a `max_ack` version now, but this only limits the logical WAL. The physical WAL probably contains important data too which we should not truncate off while we still need it for transferring the diff.
 - [x] Merge <https://github.com/qdrant/qdrant/pull/3551>
 - [x] Merge <https://github.com/qdrant/qdrant/pull/3571>
 - [x] Rebase this on `dev`
 - [x] Target this PR to merge into `dev`
 - [x] Undraft this PR

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -80,7 +80,8 @@ impl Collection {
                 ShardTransferMethod::StreamRecords => ReplicaState::Partial,
                 // TODO(1.9): switch into recovery state instead
                 ShardTransferMethod::Snapshot => ReplicaState::PartialSnapshot,
-                ShardTransferMethod::WalDelta => ReplicaState::Partial,
+                // TODO(1.9): switch into recovery state instead
+                ShardTransferMethod::WalDelta => ReplicaState::PartialSnapshot,
             };
 
             // Create local shard if it does not exist on receiver, or simply set replica state otherwise
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -49,7 +49,7 @@ impl Collection {
     {
         // Select transfer method
         if shard_transfer.method.is_none() {
-            let method = ShardTransferMethod::WalDelta;
+            let method = ShardTransferMethod::default();
             log::warn!("No shard transfer method selected, defaulting to {method:?}");
             shard_transfer.method.replace(method);
         }
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -79,9 +79,9 @@ impl Collection {
             let initial_state = match shard_transfer.method.unwrap_or_default() {
                 ShardTransferMethod::StreamRecords => ReplicaState::Partial,
                 // TODO(1.9): switch into recovery state instead
-                ShardTransferMethod::Snapshot => ReplicaState::PartialSnapshot,
-                // TODO(1.9): switch into recovery state instead
-                ShardTransferMethod::WalDelta => ReplicaState::PartialSnapshot,
+                ShardTransferMethod::Snapshot | ShardTransferMethod::WalDelta => {
+                    ReplicaState::PartialSnapshot
+                }
             };
 
             // Create local shard if it does not exist on receiver, or simply set replica state otherwise
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -49,10 +49,7 @@ impl Collection {
     {
         // Select transfer method
         if shard_transfer.method.is_none() {
-            let method = self
-                .shared_storage_config
-                .default_shard_transfer_method
-                .unwrap_or_default();
+            let method = ShardTransferMethod::WalDelta;
             log::warn!("No shard transfer method selected, defaulting to {method:?}");
             shard_transfer.method.replace(method);
         }
@@ -83,6 +80,7 @@ impl Collection {
                 ShardTransferMethod::StreamRecords => ReplicaState::Partial,
                 // TODO(1.9): switch into recovery state instead
                 ShardTransferMethod::Snapshot => ReplicaState::PartialSnapshot,
+                ShardTransferMethod::WalDelta => ReplicaState::Partial,
             };
 
             // Create local shard if it does not exist on receiver, or simply set replica state otherwise
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -49,7 +49,10 @@ impl Collection {
     {
         // Select transfer method
         if shard_transfer.method.is_none() {
-            let method = ShardTransferMethod::default();
+            let method = self
+                .shared_storage_config
+                .default_shard_transfer_method
+                .unwrap_or_default();
             log::warn!("No shard transfer method selected, defaulting to {method:?}");
             shard_transfer.method.replace(method);
         }
--- a/lib/collection/src/shards/transfer/driver.rs
+++ b/lib/collection/src/shards/transfer/driver.rs
@@ -104,7 +104,7 @@ pub async fn transfer_shard(
 
                 // Temporary hack for remote shard to get from PartialSnapshot back into Partial state
                 log::trace!("Shard {shard_id} diff recovered on {} for diff transfer, switching into next stage through consensus", remote_shard.peer_id);
-                consensus
+                let result = consensus
                     .snapshot_recovered_switch_to_partial_confirm_remote(
                         &transfer_config,
                         collection_name,
@@ -115,7 +115,10 @@ pub async fn transfer_shard(
                         crate::operations::types::CollectionError::service_error(format!(
                             "Can't switch shard {shard_id} to Partial state after diff transfer: {err}"
                         ))
-                    })?;
+                    });
+                if let Err(err) = result {
+                    log::warn!("Shard {shard_id} coult not be switched from PartialSnapshot to Partial state, ignoring: {err}");
+                }
 
                 // Redo the transfer but set a different method
                 transfer_config
--- a/lib/collection/src/shards/transfer/driver.rs
+++ b/lib/collection/src/shards/transfer/driver.rs
@@ -3,6 +3,7 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::Duration;
 
+use async_recursion::async_recursion;
 use parking_lot::Mutex;
 use tokio::time::sleep;
 
@@ -27,8 +28,9 @@ pub(crate) const MAX_RETRY_COUNT: usize = 3;
 ///
 /// This function is cancel safe.
 #[allow(clippy::too_many_arguments)]
+#[async_recursion]
 pub async fn transfer_shard(
-    transfer_config: ShardTransfer,
+    mut transfer_config: ShardTransfer,
     progress: Arc<Mutex<TransferTaskProgress>>,
     shard_holder: Arc<LockedShardHolder>,
     consensus: &dyn ShardTransferConsensus,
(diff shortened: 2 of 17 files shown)


lib/collection/src/collection/shard_transfer.rs
```diff
@@ -78,6 +75,7 @@ impl Collection {
             let initial_state = match shard_transfer.method.unwrap_or_default() {
                 ShardTransferMethod::StreamRecords => ReplicaState::Partial,
                 ShardTransferMethod::Snapshot => ReplicaState::PartialSnapshot,
+                ShardTransferMethod::WalDelta => ReplicaState::Partial,
```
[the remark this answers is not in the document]
timvisee: The problem is that if we don't use `Partial`, we need to add another replica set state. @generall wanted to prevent this if possible.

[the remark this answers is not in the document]
timvisee: One interesting property is that we'll always send those "intermixed" operations as part of the WAL diff too.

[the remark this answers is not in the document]
ffuugoo: Yeah, sure, it's just pointless. Why can't we reuse `PartialSnapshot` for this, though?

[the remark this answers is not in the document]
generall: Because transition from Dead -> `Anything` is not atomic. There will be a gap, where the updates might be send to only one node, so w either need to explicitly wait in `Patrial` state, or ensure that all nodes have received consensus updates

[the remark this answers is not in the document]
ffuugoo: I think we can't use Partial while doing diff transfer. Seems like I figured a case where it may break:Here's WAL on node A:

```
clock: A-1, point id: 1, payload: false
clock: A-2, point id: 2, payload: false
...
clock: A-1000, point id: 1000, payload: false
clock: A-1001, update all matching "payload: false" to "payload: true"
```

We are doing diff transfer from node A to node B starting from clock: 1. Here's current WAL on node B:

```
clock: A-1, point id: 1, payload: false
clock: A-2, point id: 2, payload: false
...
clock: A-500, point id: 500, payload: false
(so, ATM we transferred half of the diff)
```

Now imagine that on node C we send delete all matching "payload: false".

- Node B is in Partial state
- So this operation will be "propagated" to both node A and node B
- All 500 points that are currently in the storage on node B have payload: false
- So all of them would be deleted
- Then we will receive points 501-1000 and operation A-1001 (update "payload: false" to "payload: true") and operation from the node C (delete all "payload: false")
- But by this time half of the points on node B are already deleted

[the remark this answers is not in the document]
ffuugoo: > Now imagine that on node C...
Also note, that we don't even need node C for this example. We can do this operation directly on node A, and it will work exactly the same way. (Using separate node C just seemed a bit more "natural".)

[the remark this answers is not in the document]
ffuugoo: > In this state we currently reject incoming operations, so we cannot forward operations from a queue proxy. Because of it, we cannot transfer the WAL diff in this state.

Can we refactor so that in `PartialSnapshot` we reject all operations, except if they have `force: true`? Feels like it may solve all of our problems with above scenario.

lib/collection/src/shards/transfer/driver.rs
```diff
@@ -68,6 +75,40 @@ pub async fn transfer_shard(
...
+                    "Failed to do shard diff transfer, falling back to stream records: {err}"
+                );
+                transfer_stream_records(
+                    shard_holder.clone(),
+                    shard_id,
+                    remote_shard,
+                    collection_name,
+                )
+                .await?;
+            }
```
timvisee, line 110: This is a naive fallback approach. I'll improve this in a follow up PR, and it is noted in our tracking issue.

lib/collection/src/collection/shard_transfer.rs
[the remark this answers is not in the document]
timvisee: We'll use `PartialSnapshot` for now, and migrate to the `Recovery` state in Qdrant 1.9.
Read more about this here: https://github.com/qdrant/qdrant/pull/3659

timvisee:
We may need to fine tune the way we select/use WAL diff transfers depending on compatibility concerns. But I suggest to do this in a different PR. This is being tracked in the tracking issue (<https://github.com/qdrant/qdrant/issues/3477>).

If we merge this, we can start working on some basic integration tests.

lib/collection/src/shards/transfer/mod.rs
```diff
@@ -197,10 +202,48 @@ pub trait ShardTransferConsensus: Send + Sync {
...
+
+    tokio::select! {
+        biased;
+        Ok(_) = sync_consensus => {
+            log::trace!("All peers reached consensus");
+        }
+        _ = timeout => {
+            log::warn!("All peers failed to synchronize consensus, continuing after timeout");
+        }
+    }
```
ffuugoo, line 248: Why not `tokio::time::timeout`? IMO, less confusing than a biased `select!`.

Also, with how `select!` call is structured, if `sync_consensus` *returns an error*, select will still wait for `timeout` to expire. (That's because `Ok(_)` pattern in `sync_consensus` branch in select won't match `Err(_)` result, and so `select!` will keep waiting for `timeout` branch to resolve.)

lib/collection/src/shards/transfer/stream_records.rs
```diff
@@ -62,12 +63,33 @@ pub(super) async fn transfer_stream_records(
...
+            )));
+        };
+
+        let cutoff = replica_set.shard_recovery_point().await?;
+        let result = remote_shard
+            .update_shard_cutoff_point(collection_name, shard_id, &cutoff)
+            .await;
+        if let Err(err) = result {
+            log::warn!("Failed to update cutoff point on remote shard, ignoring: {err}");
+        }
```
ffuugoo, line 90: Should we explicitly only allow "not implemented" errors here? 🤔

lib/collection/src/shards/transfer/wal_delta.rs
```diff
@@ -0,0 +1,174 @@
...
+        // Transfer queued updates to remote, transform into forward proxy
+        // This way we send a complete WAL diff
+        log::trace!("Transfer WAL diff by transferring all queue proxy updates and transform into forward proxy");
+        replica_set.queue_proxy_into_forward_proxy().await?;
+
+        // Update cutoff point on remote shard, disallow recovery before our current last seen
+        let cutoff = replica_set.shard_recovery_point().await?;
+        remote_shard
+            .update_shard_cutoff_point(collection_name, shard_id, &cutoff)
+            .await?;
```
ffuugoo, line 131: I don't think we need this for diff transfer?

Like, imagine you just transferred a diff of exactly one point. So, recovered node now has more-or-less exactly the same WAL as the current node. It will be able to resolve any diff just fine. Why forcibly update the cutoff?

timvisee: I think you're right.

Before we had this issue with ordering during the WAL delta transfer. But we basically resolved that now with an extra replica set state. I'll remove it.

lib/collection/src/shards/transfer/driver.rs
timvisee: This is currently broken because it leaves the recovering node in unexpected state. Our stream records transfer logic cannot continue just on that.

lib/collection/src/shards/transfer/mod.rs
timvisee: IIRC I had two reasons for this. First, we always prefer the actual state rather than hitting the timeout. And second:

> Also, with how select! call is structured, if sync_consensus returns an error, select will still wait for timeout to expire.

That is intended behavior. If there is a problem or some node fails to respond, we always want to wait for `CONSENSUS_META_OP_WAIT`.

So if we're good we return right away. Otherwise we always wait. This way there is enough time to assume synchronization.

ffuugoo: > First, we always prefer the actual state rather than hitting the timeout.

Same with [`tokio::time::timeout`][tokio-time-timeout-impl].

[tokio-time-timeout-impl]: https://docs.rs/tokio/1.34.0/src/tokio/time/timeout.rs.html#201

> That is intended behavior. If there is a problem or some node fails to respond, we always want to wait for CONSENSUS_META_OP_WAIT.

This is super obscure and should be properly documented. I know well enough how `select!` works, but I still can't easily tell if it's intended or not.

Or better:

```suggestion
	let instant = tokio::time::Instant::now() + CONSENSUS_META_OP_WAIT;
	let _ = tokio::time::timeout_at(sync_consensus, instant).await;
	tokio::time::sleep_until(instant).await;
```

timvisee: > This is super obscure and should be properly documented.

It is in the comment above.

ffuugoo: The usage of `select!` is still super confusing.

It's almost impossible to tell if `Ok(_)` pattern is intentional (and we intentionally want to wait for `timeout`, if `sync_consensus` fails), or if it was an error/misunderstanding of `select!`/etc. Using such a non-trivial feature requires a clarification comment.

Or we can just use my suggested snippet above, which is easier to understand.

timvisee: Yes, your above suggestion looks good. We should use that instead :+1:

lib/collection/src/shards/transfer/stream_records.rs
ffuugoo: There should be specific `tonic::Status` or something like this. IIRC, I've done it for `/readyz`.

lib/collection/src/shards/transfer/driver.rs
timvisee: I temporarily fixed this by switching the shard state from `PartialSnapshot` back to `Partial`, required for stream records transfer. But it's more a hack than a right way of doing things.

lib/collection/src/shards/transfer/stream_records.rs
timvisee: There is, but I'm afraid it is abstracted away here.

ffuugoo: Should we add a `TODO: Only ignore "API unimplemented" error!` here or something?

lib/collection/src/shards/transfer/driver.rs
```diff
@@ -79,34 +81,53 @@ pub async fn transfer_shard(
...
+                    shard_holder,
+                    consensus,
+                    collection_id,
                     collection_name,
+                    channel_service,
+                    snapshots_path,
+                    temp_dir,
                 )
-                .await?;
+                .await;
```
ffuugoo, line 130: Just do `transfer_shard(...).boxed().await` instead of `async_recursion`? :D

ffuugoo: You need `use futures::FutureExt as _` for `.boxed()`.

ffuugoo: Apparently doesn't work like this. I'd prefer to just do whatever `async_recursion` does manually, but don't care much.

ffuugoo (approved):

lib/collection/src/shards/transfer/stream_records.rs
ffuugoo: Oh, before merging, please this! 🙏

timvisee: I have implemented it like this now:

```rust
        // Warn and ignore if remote shard is running an older version, error otherwise
        match result {
            Err(err) if err.to_string().starts_with("Service internal error: Tonic status error: status: Unimplemented") => {
                log::warn!("Cannot update cutoff point on remote shard because it is running an older version, ignoring: {err}");
            }
            Err(err) => return Err(err),
            Ok(()) => {},
        }
```

I don't like it at all, and it may be fragile, but I don't see a better way.

ffuugoo: Oh, sorry, I just meant to add a `TODO`... Sorry! 🤦‍♀️

I think we can fix this in a more meaningful way later. Like, add another error variant or something.

timvisee: I'll put the TODO back to make sure we don't miss it.

timvisee:
I'll merge this with WAL delta transfers disabled by default. I've confirmed this to work with 1.7 and 1.8 nodes in a single cluster.

For now, a user could explicitly invoke a diff transfer by setting `"method": "wal_delta"`.

The plan is to make automatic WAL delta transfers happen in the 1.8 release, but there are some state problems to resolve. I'd prefer to merge this first so we have something to work with. Automatic WAL delta transfers will be worked on separately.

PR #3510, opened 2024-02-01
Add endpoint to request recovery point for remote shard

timvisee:

This adds an gRPC (internal) interface to request the recovery point of a remote shard. It also adds a convenience function to the `RemoteShard` to easily access this.

It has a new `RecoveryPoint` type, similar to the `ClockMap`, to hold the clock states. Right now it just returns the last seen clock states as recovery point.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/api/src/grpc/proto/collections_internal_service.proto
+++ b/lib/api/src/grpc/proto/collections_internal_service.proto
@@ -1,6 +1,7 @@
 syntax = "proto3";
 
 import "collections.proto";
+import "points_internal_service.proto";
 
 package qdrant;
 option csharp_namespace = "Qdrant.Client.Grpc";
@@ -47,7 +48,11 @@ message GetShardRecoveryPointRequest {
 }
 
 message GetShardRecoveryPointResponse {
-  // TODO: replace this with last seen clock map on remote
-  uint64 recovery_point = 1; // Recovery point of the shard
+  RecoveryPoint recovery_point = 1; // Recovery point of the shard
   double time = 2; // Time spent to process
+}
+
+message RecoveryPoint {
+  // TODO: use different type here, that does not have force flag
+  repeated ClockTag clocks = 1;
 }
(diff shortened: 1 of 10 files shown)


lib/api/src/grpc/proto/collections_internal_service.proto
```diff
@@ -36,3 +41,18 @@ message WaitForShardStateRequest {
...
+
+message GetShardRecoveryPointResponse {
+  RecoveryPoint recovery_point = 1; // Recovery point of the shard
+  double time = 2; // Time spent to process
+}
+
+message RecoveryPoint {
+  // TODO: use different type here, that does not have force flag
+  repeated ClockTag clocks = 1;
+}
```
timvisee, line 58: I currently reuse the `ClockTag` for this:

```proto
message ClockTag {
  uint64 peer_id = 1;
  uint32 clock_id = 2;
  uint64 clock_tick = 3;
  bool force = 4;
}
```

We need the first three fields, but the `force` flag is obsolete here. I can make it optional, or we create a new type with just the three fields.

What do you think? I prefer not to bloat our interface with a whole bunch of types.

ffuugoo: I'd add a new type. Like, we won't be able to change it once we release, and I don't want to tie `ClockTag` to the `RecoveryPoint`. What if we want to add more fields to `ClockTag` in the future?

I'd also, maybe, make it a `map<ClockKey, uint64>` or something like this, but I guess it does not matter much, cause we'll probably convert gRPC type to "collection" type anyway.

timvisee: > Like, we won't be able to change it once we release, and I don't want to tie ClockTag to the RecoveryPoint. What if we want to add more fields to ClockTag in the future?

Good point about future maintainability. I'll add a new type.

> I'd also, maybe, make it a map<ClockKey, uint64> or something like this, but I guess it does not matter much, cause we'll probably convert gRPC type to "collection" type anyway.

I tried this at first, but it is not possible. We cannot use custom types as keys, sadly.

It also seems to introduce overhead by sorting the map and things like that.

timvisee: Added in d99708e134c1582376073485e7a8ec06cf915cd1

generall (approved):

PR #3511, opened 2024-02-01
Issues API: Unindexed field issue integration

coszio:
Needs: #3375. Tracked by: #3471 

Submits and solves the `UnindexedField` issue:
- submits when a search request takes longer than some threshold (currently 300ms)
- solves after a `CreateIndex` update is successful on the same collection and field name, or the collection is deleted

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/segment/src/index/struct_payload_index.rs
```diff
@@ -309,7 +310,21 @@ impl StructPayloadIndex {
...
+                        self.get_collection_name(),
+                    )) {
+                        let description = unindexed_field.description();
+                        if issues::submit(unindexed_field) {
+                            log::warn!("Performance issue: {}", description);
+                        }
+                    }
+
+                    CardinalityEstimation::unknown(self.available_point_count())
+                }),
```
coszio, line 327: For reviewers: Here is where this issue is triggered

src/common/points.rs
```diff
@@ -547,14 +547,23 @@ pub async fn do_create_index_internal(
...
+            ordering,
+            shard_selector,
+        )
+        .await;
+
+    if res.is_ok() {
+        // We can deactivate issues related to this missing index
+        let code = segment::problems::UnindexedField::get_code(collection_name, &field_name);
+        issues::solve(code);
+    }
```
coszio, line 564: And here it is solved

lib/segment/src/problems/unindexed_field.rs
```diff
@@ -0,0 +1,177 @@
...
+impl TryFrom<(FieldCondition, String)> for UnindexedField {
+    type Error = OperationError;
+
+    /// Try to form an issue from a field condition and a collection name
+    ///
+    /// # Failures
+    ///
+    /// Will fail if the field condition cannot be used for inferring an appropriate schema.
+    /// For example, when there is no index that can be built to improve performance.
+    fn try_from(condition_collection_tuple: (FieldCondition, String)) -> Result<Self, Self::Error> {
```
timvisee, line 161: I don't think I like this too much. I find converting `(FieldCondition, String)` into an `UnindexedField` type quite arbitrary.

To use it, you'd have to construct a tuple too.

What about just:

```rust
impl UnindexedField {
    fn new(condition: FieldCondition, collection: String) { ... }
}
```

lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -58,7 +59,13 @@ impl TableOfContent {
             }
             CollectionMetaOperations::DeleteCollection(operation) => {
                 log::info!("Deleting collection {}", operation.0);
-                self.delete_collection(&operation.0).await
+                let res = self.delete_collection(&operation.0).await;
+
+                // We can solve all issues related to this collection
+                if *res.as_ref().unwrap_or(&false) {
+                    issues::solve_by_filter(|code| code.starts_with(&operation.0));
+                }
```
timvisee, line 67: Wouldn't it be better to just store the collection name in the issue, and match that directly.

I feel like matching part of a string here may be very error prone here.

For example, I could name my collection: `something.UNINDEXED_FIELD`

coszio: I guess I didn't check our validation for collection names, so maybe we can use `/` as separator instead?

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -309,7 +310,26 @@ impl StructPayloadIndex {
...
+                            crate::problems::UnindexedField::try_from((
+                                field_condition.clone(),
+                                collection_name,
+                            ))
+                            .ok()
+                        })
+                    {
+                        let description = unindexed_field.description();
+                        if issues::submit(unindexed_field) {
+                            log::warn!("Performance issue: {}", description);
```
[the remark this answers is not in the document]
coszio: it wasn't until recently that VSCode represented this formatting correctly, 'cause sometimes it wouldn't propagate squiggly lines there, so I got used to do it outside of the fmt string.

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -309,7 +310,26 @@ impl StructPayloadIndex {
             }
             Condition::Field(field_condition) => self
                 .estimate_field_condition(field_condition, nested_path)
-                .unwrap_or_else(|| CardinalityEstimation::unknown(self.available_point_count())),
+                .unwrap_or_else(|| {
```
agourlay, line 313: TBH I am unsure if it is the best spot for generating the issues :thinking: 

The cardinality estimation is used to pick the driving filter from the query, that is the one with the highest selectivity.

I am worried this generates a lot of issue objects under high read load with lots of filters.
Most of those will be discarded because they exist already in the in-memory issues store. 

An alternative would be to place it in the function `condition_converter` to catch the actual field driving the query.

Can you maybe elaborate on why you picked this approach? :)

coszio: While I agree that this might not be the best place for generating the issues. I would argue that it is more because it might have side effects for when the cardinality estimation is not directly used for searching or filtering, rather than if the issue is trying to be generated for every unindexed condition. 

My rationale is: if the search is slow already due to unindexed field, it won't affect much to try to do a read operation on the issues dashboard (when checking if the issue already exists).

After seeing if I can move this generation to the `condition_converter` function, I realized there is no easy way to include the collection name without drilling it down across many parent functions. So the question is: would it be worth it to increase the complexity for this?

coszio: Another place we could add it is in:
https://github.com/qdrant/qdrant/blob/7ac7398bb129946c7c770f882272a0917d647237/lib/segment/src/index/struct_payload_index.rs#L456-L460

So that only the primary conditions trigger the issue

cc @generall

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -331,6 +351,16 @@ impl StructPayloadIndex {
     ) -> OperationResult<()> {
         crate::rocksdb_backup::restore(snapshot_path, &segment_path.join("payload_index"))
     }
+
+    /// Hack to get collection name from the path, to avoid early refactoring on segment level
+    fn get_collection_name(&self) -> Option<String> {
+        // Root dir -> storage -> collections -> ⭐️collection_name⭐️
+        self.path
+            .components()
+            .nth(3)
```
generall, line 360: too ugly

coszio:
Pending: do time-based issue generation only

coszio:
Update: This issue is now triggered on core searches only, when the request takes more than 300ms (arbitrarily chosen) to fulfill. 

It scans the filter and checks against the current schema, so that, if the field is already indexed, it won't submit the issue.

The current `GET /issues` output for when there is a missing index on `sections` (collection name is `docs`), looks like this:
imagen

lib/segment/src/problems/unindexed_field.rs
```diff
@@ -0,0 +1,315 @@
...
+                    .ok()
+                    .and_then(|var| var.parse::<u64>().ok())
+                    .unwrap_or(300),
+            )
+        })
+    }
+
+    pub fn get_code(collection_name: &str, field_name: &JsonPathV2) -> CodeType {
+        format!("/{collection_name}/UNINDEXED_FIELD/{field_name}")
+    }
```
timvisee, line 41: Sorry, but I still don't like this string format because it  requires string based matching when automatically resolving issues. That's error prone.

Is there a good reason for not doing it in a strongly typed manner? Could that be a bad idea because various issues might be on wildly different levels, where a simple string could easily represent them all?

How about a flat structure where all issues are on the same level, having some constant type such as `UNINDEXED_FIELD` where each type can have their own respective metadata - a collection and field name in this case?

coszio: I kinda share your thoughts on this... 

How about this: If we adhere to a REST-like representation, we can avoid headaches. In this case, a better code for this issue could be `/collections/{collection_name}/points/index/{field_name}/UNINDEXED_FIELD`

What makes me reluctant to have a flat structure is that it will want to have a different implementation of the `issues::Dashboard`, in which it becomes event-based, so that we don't pollute the codebase too much by separating the event from the handling. Maybe it is not a bad thing either, but it will require at least another PR.

coszio:
Closing in favor of #4139

PR #3516, opened 2024-02-02
Datetime index: Introduce `RangeInterface`

coszio:
Initially it made sense to me to keep different types separate in `FieldCondition`, but now I propose to handle all `Range` conditions similarly and together, so here we add a new `RangeInterface` enum to wrap them together.

This change also made me realize we missed using the actual index sometimes when filtering `datetime_range` conditions (this is fixed now, but not very pretty, see `PayloadFieldIndex` implementation).

Note: this only affects the REST interface, GRPC is intact.

A bonus change I did is to use pattern matching and destructuring in the checkers section, so it will be checked by compiler to not forget any new conditions.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/numeric_index/mod.rs
+++ b/lib/segment/src/index/field_index/numeric_index/mod.rs
@@ -118,30 +118,22 @@ impl Encodable for DateTimePayloadType {
     }
 }
 
-impl Range<FloatPayloadType> {
-    pub(in crate::index::field_index::numeric_index) fn as_bounds<T: Encodable + Numericable>(
+impl<T: Encodable + Numericable> Range<T> {
+    pub(in crate::index::field_index::numeric_index) fn as_index_key_bounds(
         &self,
     ) -> (Bound<NumericIndexKey<T>>, Bound<NumericIndexKey<T>>) {
         let start_bound = match self {
-            Range { gt: Some(gt), .. } => {
-                let v: T = T::from_f64(*gt);
-                Excluded(NumericIndexKey::new(v, PointOffsetType::MAX))
-            }
+            Range { gt: Some(gt), .. } => Excluded(NumericIndexKey::new(*gt, PointOffsetType::MAX)),
             Range { gte: Some(gte), .. } => {
-                let v: T = T::from_f64(*gte);
-                Included(NumericIndexKey::new(v, PointOffsetType::MIN))
+                Included(NumericIndexKey::new(*gte, PointOffsetType::MIN))
             }
             _ => Unbounded,
         };
 
         let end_bound = match self {
-            Range { lt: Some(lt), .. } => {
-                let v: T = T::from_f64(*lt);
-                Excluded(NumericIndexKey::new(v, PointOffsetType::MIN))
-            }
+            Range { lt: Some(lt), .. } => Excluded(NumericIndexKey::new(*lt, PointOffsetType::MIN)),
             Range { lte: Some(lte), .. } => {
-                let v: T = T::from_f64(*lte);
-                Included(NumericIndexKey::new(v, PointOffsetType::MAX))
+                Included(NumericIndexKey::new(*lte, PointOffsetType::MAX))
             }
             _ => Unbounded,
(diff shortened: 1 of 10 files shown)


lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -196,26 +197,50 @@ impl<T: Encodable + Numericable + Default> NumericIndex<T> {
...
+
+                let gbound = if let Some(gte) = datetime_range.gte {
+                    Included(T::from_i64(gte.timestamp_micros()))
+                } else if let Some(gt) = datetime_range.gt {
+                    Excluded(T::from_i64(gt.timestamp_micros()))
+                } else {
+                    Unbounded
+                };
+                (lbound, gbound)
+            }
```
xzfc, line 243: You could reduce code duplication by converting to `Range<T>` first.
Also, an idea: add `to_bounds` method to the `Range<T>` structure.

```rust
impl<T: Copy> Range<T> {
    fn to_bounds(&self) -> (Bound<T>, Bound<T>) {...}
}
```

So this code could be reduced to this:
```rust
let (lbound, gbound) = match range {
    RangeInterface::Float(range) => range.map(T::from_f64),
    RangeInterface::DateTime(range) => range.map(|t| T::from_i64(i.timestamp_micros())),
}.to_bounds();
```

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -301,30 +326,62 @@ impl<T: Encodable + Numericable + Default> PayloadFieldIndex for NumericIndex<T>
...
+                        let v: T = T::from_i64(lt.timestamp_micros());
+                        Excluded(NumericIndexKey::new(v, PointOffsetType::MIN))
+                    }
+                    Range { lte: Some(lte), .. } => {
+                        let v: T = T::from_i64(lte.timestamp_micros());
+                        Included(NumericIndexKey::new(v, PointOffsetType::MAX))
+                    }
+                    _ => Unbounded,
+                };
+                (start_bound, end_bound)
```
xzfc, line 383: You could reduce code duplication by converting to `Range<T>` first.
```rust
let cond_range = match cond_range {
    RangeInterface::Float(range) => range.map(T::from_f64),
    RangeInterface::DateTime(range) => range.map(|t| T::from_i64(i.timestamp_micros())),
}
let start_bound = match cond_range {
...
```

xzfc (dismissed):

lib/segment/src/index/field_index/numeric_index/mod.rs
coszio: I actually implemented a similar thing for #3493, I'm going to reuse it here 👍

lib/segment/src/types.rs
```diff
@@ -1323,6 +1323,29 @@ impl From<Vec<IntPayloadType>> for MatchExcept {
...
+            _ => None,
+        }
+    }
+
+    pub fn to_datetime_range(self) -> Option<Range<DateTimePayloadType>> {
+        match self {
+            RangeInterface::DateTime(range) => Some(range),
+            _ => None,
+        }
+    }
```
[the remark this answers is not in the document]
coszio: Good eye 👁️, I was using these for conversions, but then reimplemented them and forgot to remove them

lib/segment/src/index/query_optimization/condition_converter.rs
```diff
@@ -128,55 +128,41 @@ pub fn field_condition_index<'a>(
...
+            ..
+        } => get_geo_polygon_checkers(index, geo_polygon.clone()),
 
-    if let Some(checker) = field_condition
-        .geo_polygon
-        .clone()
-        .and_then(|cond| get_geo_polygon_checkers(index, cond))
-    {
-        return Some(checker);
+        FieldCondition {
```
generall, line 156: this is nice!

generall (approved):

PR #3519, opened 2024-02-04
fix: segment_builder_test::test_building_cancellation

### LINKED ISSUE #2723: Flaky test `segment_builder_test::test_building_cancellation`
### System error

```text
baseline time: 9921
thread 'segment_builder_test::test_building_cancellation' panicked at 'time_later: 2813, late_stop_delay: 1984', lib\segment\tests\integration\segment_builder_test.rs:172:5
```

### Context

Date: 25.09.2023 09:59

[Flaky failure run](https://github.com/qdrant/qdrant/actions/runs/6297499749/job/process-results)

[Commit](https://github.com/qdrant/qdrant/tree/840248f6fcddcd27bef686ccaf70af4afe67f0ee)

OS: windows-latest

Branch: refs/heads/dev

#

jkcs:
close #2723

**A Novice's Attempt at Rust, If there are any issues, please point them out. Thank All.**

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/tests/integration/segment_builder_test.rs
+++ b/lib/segment/tests/integration/segment_builder_test.rs
@@ -107,7 +107,7 @@ fn estimate_build_time_with_callback(
 
     let stopped_t = stopped.clone();
 
-    std::thread::Builder::new()
+    let handle = std::thread::Builder::new()
         .name("build_estimator_timeout".to_string())
         .spawn(move || {
             build_callback(&stopped_t);


generall:
Hi @jkcs, could you please provide some explanations for the changes?

timvisee:
Hi @jkcs. I don't think I understand how this solves the flaky tests.

> In my understanding, this is essentially a thread synchronization issue. The assertion within `baseline_segment` stipulates that it must complete within 20 seconds;

No, the baseline test should run longer than 20 seconds. But it will be cancelled in 20 seconds to get some baseline on how long running for 20 seconds and cancelling takes.

> Ideally, there should be no need for a timeout duration.

That's right, but there does not seem to be any better or more reliable way to test for cancellation (within time!) without depending on some timeout.

> When executing baseline_segment, there is no need to set a timeout; we simply wait for its completion to obtain the time_baseline value.

I think we do, because it is not a baseline measurement for completion, but a baseline measurement for 20 seconds plus cancellation. The function will simply return an error if it was not cancelled.

timvisee:
@jkcs I don't see how this solves this being flaky.

All logic still seems to be the same, except for removing the _timeout_ for the baseline measurement. I don't see how using a callback here is different than implementing the timeout/cancellation in the function directly.

If I recall correctly, the flakiness actually comes from the short and long tasks taking a longer/shorter time to cancel than we expected, possibly because the CI machine is busy with a lot of other things.

Maybe I'm missing something important. If you think I do, then could you please: 1. explain what you think the problem is, in more detail. 2. how your approach specifically solves this problem.

timvisee:
> Sorry, I haven't fully grasped this issue yet. This attempt has failed, and the PR is closed.
> 

No problem. We haven't either! 😁

Thank you for giving it a shot.

Issue #3522, opened 2024-02-04
Optimize MatchAny for large amount of values


generall:
**Is your feature request related to a problem? Please describe.**

There are use-case, where we might want to specify a lot of values in the MatchAny condition.
One of those use-cases is "reddit" type of application, where user might have up to a several thousand subscriptions and we want to search in only among those subscriptions (and not all groups).

**Describe the solution you'd like**

Make sure our implementation of MatchAny is optimal enough to handle those case.

To do so, we would need:

- Benchmark dataset, which approximates mentioned scenario
- Measurements before and after the change
- Handling of the case on the segment level (most likely it could something like converting MatchAny list into hashmap if it is long enough)

**Describe alternatives you've considered**

Other potential approach is to assign user "ids" to the groups, but in this case it might require huge lists of users associated with each group record. 

**Additional context**

No interface / storage changes are expected in this PR.

haruncurak:
Hi all! I'd love to try and tackle this one as my first Qdrant issue . Would be fantastic if I could get assigned - excited to get going. 

/attempt #3522

timvisee:
> One of those use-cases is "reddit" type of application, where user might have up to a several thousand subscriptions and we want to search in only among those subscriptions (and not all groups).

This is a nitpick, but Reddit does not support this either. On your front-page, it takes 50 random subreddits you subscribed on and only shows content for them. 

Maybe it would be possible for the user actually wanting this to use a similar approach.

generall:
> This is a nitpick, but Reddit does not support this either. On your front-page, it takes 50 random subreddits you subscribed on and only shows content for them.


Maybe that's a reddit who requested this feature to be improved :man_shrugging:

subh-cs:
attempt #3522

haruncurak:
Just a quick check @generall - would the points in the approximate dataset be users with subscriptions, potentially including posts within those subscriptions? 

Something like this?
```

{
  "id": "5c56c793-69f3-4fbf-87e6-c4bf54c28c26", (user id)
  "vector": [0.9, 0.1, 0.1],
  "payload": {
    "subscriptions": [
      {
        "subscription": "subreddit_name",
        "posts": [
          {
            "title": "",
            "content": "",
          },
          {
            "...": "..."
          }
        ]
      },
      {
        "...": "..."
      }
    ]
  }
}

```

Or would the subscriptions (i.e. subreddits) themselves be points?

generall:
each post in subreddit should be a separate point.

PR #3525, opened 2024-02-05
Optimize MatchAny

JojiiOfficial:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

/claim #3522 

**Note:** in /Cargo.toml:181 I used a custom fork of schemars to allow using SmolStr directly in the API. This is unsecure and should be replaced with a fork in the qdrant organization like the tonic patch above!

# Benches
I created a new bench function `conditional-search-large-match-any` that measures the performance increase. To test the performance of the old implementation one can simply checkout the first commit of this PR.

Results for 100k points, 1k MatchAny values of which 15% can be a matching value:
Old: `1.07ms`
This PR: `561.12µs`

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -6575,14 +6575,16 @@
             "type": "array",
             "items": {
               "type": "string"
-            }
+            },
+            "uniqueItems": true
           },
           {
             "type": "array",
             "items": {
               "type": "integer",
               "format": "int64"
-            }
+            },
+            "uniqueItems": true
           }
         ]
       },
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -6583,7 +6583,8 @@
             "items": {
               "type": "integer",
               "format": "int64"
-            }
+            },
+            "uniqueItems": true
           }
         ]
       },
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -6583,8 +6583,7 @@
             "items": {
               "type": "integer",
               "format": "int64"
-            },
-            "uniqueItems": true
+            }
           }
         ]
       },
--- a/lib/segment/src/index/query_optimization/condition_converter.rs
+++ b/lib/segment/src/index/query_optimization/condition_converter.rs
@@ -306,9 +306,13 @@ pub fn get_match_checkers(index: &FieldIndex, cond_match: Match) -> Option<Condi
             (AnyVariants::Keywords(list), FieldIndex::KeywordIndex(index)) => {
                 Some(Box::new(move |point_id: PointOffsetType| {
                     index.get_values(point_id).map_or(false, |values| {
-                        values
-                            .iter()
-                            .any(|k| list.iter().any(|s| s.as_str() == k.as_ref()))
+                        if list.len() < 13 {
+                            values
+                                .iter()
+                                .any(|k| list.iter().any(|s| s.as_str() == k.as_ref()))
+                        } else {
+                            values.iter().any(|k| list.contains(k.as_str()))
+                        }
                     })
                 }))
             }
@@ -325,9 +329,13 @@ pub fn get_match_checkers(index: &FieldIndex, cond_match: Match) -> Option<Condi
             (AnyVariants::Keywords(list), FieldIndex::KeywordIndex(index)) => {
                 Some(Box::new(move |point_id: PointOffsetType| {
                     index.get_values(point_id).map_or(false, |values| {
-                        values
-                            .iter()
-                            .any(|k| !list.iter().any(|s| s.as_str() == k.as_ref()))
+                        if list.len() < 13 {
+                            values
+                                .iter()
+                                .any(|k| !list.iter().any(|s| s.as_str() == k.as_ref()))
+                        } else {
+                            values.iter().any(|k| !list.contains(k.as_str()))
+                        }
                     })
                 }))
             }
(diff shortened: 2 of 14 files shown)


JojiiOfficial:
I experienced a further increase in performance around 5-15% in most filtering benchmarks using the fnv hashing algorithm instead of the std's SipHash. 
The [fnv crate](https://github.com/servo/rust-fnv) is already part of the dependency tree, so it doesn't add any further crates that need to be compiled when compiling the qdrant binary.

generall:
Hey @JojiiOfficial, thanks for the PR.

Couple of things to consider from my side:

- Would it be possible to avoid usage of non-merged PRs? I understood that's for the SmolStr, but is it really necessary for the internal optimization? Maybe we could decompose external structures from the internal logic?
- We need to make sure, that those changes doesn't compromise the performance of the default use-case with small number of values

JojiiOfficial:
Thanks for the reply @generall 

Indeed I found a way to remove the SmolStr from the external structure. I also added a benchmark for a small number of keyword values and optimized the comparison logic. 

With the latest changes I don't get a real performance regression over multiple benchmarks runs. In contrast most other conditional-search benches got a slight increase in performance in average.

generall:
FYI I was using our benchmarking tool with the following changes: https://github.com/qdrant/bfb/commit/cb1a77e66b4dc7437505165509bbcf6b703bd446

JojiiOfficial:
You're right! Thanks for checking.
I indeed missed a spot where the filter gets checked. As you can see from the benchmark results below, It is nearly twice as fast for 200 keywords per query. Even for only 20 points we have a performance improvement of around 22%
I'll experience and try to optimize a bit more and push my changes later.

image

docs/redoc/master/openapi.json
```diff
@@ -6585,7 +6585,8 @@
             "type": "array",
             "items": {
               "type": "string"
-            }
+            },
+            "uniqueItems": true
```
coszio, line 6589: This might lead to a breaking change in clients, is it possible to remove?

lib/segment/src/index/query_optimization/condition_converter.rs
```diff
@@ -332,9 +336,13 @@ pub fn get_match_checkers(index: &FieldIndex, cond_match: Match) -> Option<Condi
             (AnyVariants::Keywords(list), FieldIndex::KeywordIndex(index)) => {
                 Some(Box::new(move |point_id: PointOffsetType| {
                     index.get_values(point_id).map_or(false, |values| {
-                        values
-                            .iter()
-                            .any(|k| !list.iter().any(|s| s.as_str() == k.as_ref()))
+                        if list.len() < 13 {
```
coszio, line 339: We should put this magic number in a constant along with a comment to explain where it comes from

lib/segment/src/payload_storage/condition_checker.rs
```diff
@@ -91,15 +91,27 @@ impl ValueChecker for Match {
...
                 (Value::Number(stored), AnyVariants::Integers(list)) => stored
                     .as_i64()
                     .map(|num| list.contains(&num))
                     .unwrap_or(false),
                 _ => false,
             },
             Match::Except(MatchExcept { except }) => match (payload, except) {
-                (Value::String(stored), AnyVariants::Keywords(list)) => !list.contains(stored),
+                (Value::String(stored), AnyVariants::Keywords(list)) => {
+                    if list.len() < 13 {
```
coszio, line 109: Same here

docs/redoc/master/openapi.json
generall: I don't think client enforces this one. I propose to ignore for now

lib/segment/src/types.rs
```diff
@@ -1185,7 +1187,7 @@ pub enum ValueVariants {
 #[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, PartialEq, Eq)]
 #[serde(untagged)]
 pub enum AnyVariants {
-    Keywords(Vec<String>),
+    Keywords(IndexSet<String, FnvBuildHasher>),
     Integers(Vec<IntPayloadType>),
```
coszio, line 1191: Why are integers not optimized?

coszio (changes requested):

lib/segment/src/index/query_optimization/condition_converter.rs
generall: +1

lib/segment/src/types.rs
JojiiOfficial: My initial benchmark showed that searching integer iteratively were faster than hashing for a reasonable sized list. I assumed this were the case as integer perfectly fit into cache while being very small. 
Doing a new quick bench however showed

coszio: Is the threshold the same for both strings and integers? 

Another quick idea (for both strings and integers) is to do binary search over a sorted list as a replacement of iterating. I bet this can increase the threshold before which hashing is better. Although, to be fair, this would add extra complexity and can be addressed later in another PR.

timvisee: Maybe we can base this threshold on number of bytes instead.

> Another quick idea (for both strings and integers) is to do binary search over a sorted list as a replacement of iterating. I bet this can increase the threshold before which hashing is better. Although, to be fair, this would add extra complexity and can be addressed later in another PR.

Yes, I've been thinking about this too. But it would make our logic a bit more complex as we likely end up with two different search methods based on some heuristic.

generall (approved):

lib/segment/src/types.rs
coszio: I played around with the comparisons between binary search, linear scan, and IndexSet for searching UUIDs. The results favor the current implementation (the one in this PR 🎉).

Before a small threshold, linear scan and binary search perform basically the same. Y-Axis is nanoseconds, X-Axis is list size.
imagen

However maybe the threshold (at least for UUIDs) might be better adjusted at size of 30 items.

TLDR: there is a really small optimization that can be made by setting the threshold based on a byte size heuristic, but I don't think it is worth the complexity.

timvisee: I guess this to be very dependent on hardware configuration.

But what about setting a threshold at 512 bytes (32 * 16 (UUID) = 512)? It might scale nicely that way.

PR #3527, opened 2024-02-05
Optimize sparse search by advancing postings without binary search

agourlay:
context: https://github.com/qdrant/qdrant/pull/3464

Optimize sparse vector search by advancing posting lists without binary search.

We know exactly where to move the cursor for the posting list when reaching the end of the batch.

No need to use `skip_to` with a point id which performs a binary search internally.

# Performance numbers

- 8 segments
- index.on_disk=false
- 6980 queries

```
DEV
min: 125.61 millis
50p: 374.41 millis
95p: 570.08 millis
99p: 644.55 millis
999p: 723.14 millis
max: 772.38 millis
```

```
PR
min: 90.75 millis
50p: 179.43 millis
95p: 310.99 millis
99p: 408.16 millis
999p: 462.66 millis
max: 533.39 millis
```

DEV             |  PR
sparse_bench_small_dev  | sparse_bench_small_PR

### DIFF
--- a/lib/sparse/src/index/search_context.rs
+++ b/lib/sparse/src/index/search_context.rs
@@ -134,9 +134,9 @@ impl<'a> SearchContext<'a> {
         filter_condition: &F,
     ) {
         for posting in self.postings_iterators.iter_mut() {
-            // index at which the posting list stops contributing to the batch (relative to the batch start)
+            // offset at which the posting list stops contributing to the batch (relative to the batch start)
             let mut posting_stopped_at = None;
-            for (index, element) in posting
+            for (offset, element) in posting
                 .posting_list_iterator
                 .remaining_elements()
                 .iter()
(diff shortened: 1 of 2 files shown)


lib/sparse/src/index/search_context.rs
```diff
@@ -134,19 +134,36 @@ impl<'a> SearchContext<'a> {
         filter_condition: &F,
     ) {
         for posting in self.postings_iterators.iter_mut() {
-            for element in posting.posting_list_iterator.remaining_elements() {
+            // index at which the posting list stops contributing to the batch (relative to the batch start)
+            let mut posting_stopped_at = None;
+            for (index, element) in posting
```
timvisee, line 139: Nitpick, but wouldn't it be better to call it `offset`. It's the _index_ in the remaining set, not in the full set. Right?

timvisee (approved):

IvanPleshkov (approved):

lib/sparse/src/index/search_context.rs
agourlay: You are right, `offset` is a better fit there :+1: 
https://github.com/qdrant/qdrant/pull/3527/commits/3d24a17ef4c2479f7821f0ab2a8fd92eeff67436

Issue #3521, opened 2024-02-04
[order-by] Tracking issue


generall:
**Is your feature request related to a problem? Please describe.**

Original issue: https://github.com/qdrant/qdrant/issues/1763

ToDo:

* [x] initial implementation #3493
* [x] Edit `order_by` description to drop ordering by id guarantee #3589 
* [x] Enable selection of efficient query execution strategy #3562
* [x] Enable `order_by` for date-time values #3540
* [x] Extend internal representation of the order-able value (currently - f64), to also be compatible with large i64 values #3533
* [x] ~Get rid of~ https://github.com/qdrant/qdrant/blob/b0aab16b0de67488e2859c8c511c806c8efdf82c/lib/segment/src/data_types/order_by.rs#L7
* [x] #3726
* [ ] Implement feature in local mode qdrant/qdrant-client#491
* [ ] Write documentation qdrant/landing_page#631

PR #3533, opened 2024-02-05
Order-by: Introduce `OrderingValue` for ordering also based on int values

coszio:
Tracked by #3521.

Currently all ordering values are transformed to `f64` to be comparable and ordered. 
With these changes int indexes won't have to transform to float, and thus large int values can still be ordered correctly.

This was solved here using an enum to wrap both ints and floats.

### All Submissions:

### Changes to Core Features:


lib/segment/src/data_types/order_by.rs
```diff
@@ -81,3 +85,60 @@ impl OrderBy {
...
+        OrderingValue::Int(value)
+    }
+}
+
+impl PartialEq for OrderingValue {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (OrderingValue::Float(a), OrderingValue::Float(b)) => a == b,
+            (OrderingValue::Int(a), OrderingValue::Int(b)) => a == b,
+            (OrderingValue::Float(a), OrderingValue::Int(b)) => a == &OrderedFloat(*b as f64),
```
xzfc, line 123: Lossy type conversion violates the transitive condition required by `PartialEq` and `PartialCmp`. (see [docs](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html))

```
let a = OrderingValue::from(9007199254740992);
let b = OrderingValue::from(9007199254740992.0);
let c = OrderingValue::from(9007199254740993);
assert!(a == b); // true
assert!(b == c); // true
assert!(a == c); // false
```

I think to solve it we could provide a function `fn(i64, f64) -> Ordering` that carefully handles all possible cases; and reuse this function in `PartialEq` and `Ord` implementations. Maybe it's already implemented in some crate.

---

Not related to your PR, but a fun trivia. A recent vulnerability found in glibc: if you pass a non-transitive comparison function to `qsort`, you'll get out-of-bounds read and writes. https://www.openwall.com/lists/oss-security/2024/01/30/7
It would be not that severe in our case, but I think it still could break sorting or other function's logic.

xzfc (changes requested):

lib/segment/src/data_types/order_by.rs
xzfc: Because of f64 rounding. It could be simplified to this:
```rust
assert!(9007199254740992.0 == 9007199254740993.0); // true
```
`2**53 + 1.0` could not be represented by f64, so it's getting rounded to `2**53`.

generall: looks like the violation of transitive condition is only a consequence of comparison which works incorrectly then

xzfc: It works correctly since `9007199254740992.0f64` has the same representation as `9007199254740993.0f64`. It's the same value bitwise. (`println!("{}", 9007199254740993.0);` prints the `…92` not `…93`)

The problem arise because of the cast (`as f64`) being used. The rounding happens in it.

generall: maybe this https://crates.io/crates/num-cmp ?

xzfc: From a quick glance and testing this crate does the right thing.

xzfc: I did update the code to follow the `NaN` handling, in the same way as `OrderingFloat` does.

xzfc (approved):

timvisee:
Late to the party, but, why not `OrderingValue -> OrderedValue` to match `OrderedFloat`?

PR #3540, opened 2024-02-05
Order-by: support datetimes

coszio:
Tracked by: #3521 

We recently added support for a datetime index, which also has range-able capabilities. This PR enables the usage of datetimes for ordering in scroll with minimal interface changes.

### Future work:
- Allow integers as `start_from` parameter. But this should also be paired with a new `RangeInterface::Int(Range<i64>)`

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -310,7 +310,8 @@ message StartFrom {
   oneof value {
     double float = 1;
     int64 integer = 2;
-    google.protobuf.Timestamp datetime = 3;
+    google.protobuf.Timestamp timestamp = 3;
+    string datetime = 4;
   }
 }
 
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -310,6 +310,7 @@ message StartFrom {
   oneof value {
     double float = 1;
     int64 integer = 2;
+    google.protobuf.Timestamp datetime = 3;
   }
 }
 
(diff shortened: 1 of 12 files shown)


coszio:
This one has now been rebased with the latest `OrderingValue` changes 😄

cc @generall @xzfc

lib/api/src/grpc/proto/points.proto
```diff
@@ -310,6 +310,7 @@ message StartFrom {
   oneof value {
     double float = 1;
     int64 integer = 2;
+    google.protobuf.Timestamp datetime = 3;
```
generall, line 313: Maybe also include string representation?

generall: yeah

lib/segment/src/data_types/order_by.rs
```diff
@@ -17,20 +18,31 @@ pub enum Direction {
...
             },
         }
     }
 }
 
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+#[serde(untagged)]
+pub enum StartFrom {
+    Float(FloatPayloadType),
+    Datetime(DateTime<Utc>),
```
generall, line 43: We might need `deserialize_with` here, so it could work with requests like:

```
POST collections/test_collection/points/scroll
{
  "limit": 2,
  "order_by": {
    "direction": "desc",
    "key": "date",
    "start_from": "2024-10-11 12:34"
  }
}
```

See https://github.com/qdrant/qdrant/pull/3549

generall (approved):

Issue #3543, opened 2024-02-06
REST API pattern match ordering problem


timvisee:
Related to <https://github.com/qdrant/qdrant-web-ui/issues/158>.

## Current Behavior
We have middleware in our REST API for API key validation. For the read only API key, we have a [list](https://github.com/qdrant/qdrant/blob/524adc3f70805b33a6b94bce410a20f3d95bfa51/src/actix/api_key.rs#L13-L25) of POST endpoints we still allow reading from. Our scroll endpoint is a good example of this.

In the following lines we grab the pattern that matches a user request, and match it against the [list](https://github.com/qdrant/qdrant/blob/524adc3f70805b33a6b94bce410a20f3d95bfa51/src/actix/api_key.rs#L13-L25) of allowed read only patterns:

https://github.com/qdrant/qdrant/blob/524adc3f70805b33a6b94bce410a20f3d95bfa51/src/actix/api_key.rs#L161-L164

However, with some debugging I found out that matching patterns is giving us wrong results breaking our logic.

For example:
- `POST /collections/benchmark/points/scroll` 
- matches: `/collections/{name}/points/{id}`
- but it should be: `/collections/{name}/points/scroll`

## Steps to Reproduce
1. Set up a read only API key
2. Wrap the above lines in `dbg!()`
3. Send a scroll request
4. See debug output in your log, it shows the wrong pattern

## Expected Behavior
Instead, it should show the correct match pattern.

## Possible Solution
Maybe we can tweak the ordering of pattern matching to always match strict paths like `/scroll` first, and `/{id}` after.

## Additional details
Fixes <https://github.com/qdrant/qdrant-web-ui/issues/158>.

PR #3544, opened 2024-02-06
Fix REST API pattern match ordering problem

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/3543>, <https://github.com/qdrant/qdrant-web-ui/issues/158>.

This reorders our actix web REST API services so that we don't have pattern matching conflicts.

As a micro optimization, I now also binary search of the list of allowed patterns.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/src/actix/api_key.rs
+++ b/src/actix/api_key.rs
@@ -10,18 +10,19 @@ use futures_util::future::LocalBoxFuture;
 
 use crate::common::auth::AuthKeys;
 
+/// List of read-only POST request paths. List MUST be sorted.
 const READ_ONLY_POST_PATTERNS: [&str; 11] = [
     "/collections/{name}/points",
     "/collections/{name}/points/count",
-    "/collections/{name}/points/search",
-    "/collections/{name}/points/scroll",
-    "/collections/{name}/points/search/groups",
-    "/collections/{name}/points/search/batch",
-    "/collections/{name}/points/recommend",
-    "/collections/{name}/points/recommend/groups",
-    "/collections/{name}/points/recommend/batch",
     "/collections/{name}/points/discover",
     "/collections/{name}/points/discover/batch",
+    "/collections/{name}/points/recommend",
+    "/collections/{name}/points/recommend/batch",
+    "/collections/{name}/points/recommend/groups",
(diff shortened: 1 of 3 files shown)


generall:
If the ordering is required, it should be enforced

generall:
also, why is it even required? Why not hashmap?

generall (approved):

src/actix/api_key.rs
```diff
@@ -10,18 +10,19 @@ use futures_util::future::LocalBoxFuture;
 
 use crate::common::auth::AuthKeys;
 
+/// List of read-only POST request paths. List MUST be sorted.
```
agourlay, line 13: is it possible to have a debug_assert! somewhere to check that it is sorted?

src/actix/api_key.rs
```diff
@@ -160,8 +161,27 @@ fn is_read_only(req: &ServiceRequest) -> bool {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn read_only_post_patterns_sorted() {
+        let mut sorted = READ_ONLY_POST_PATTERNS;
+        sorted.sort_unstable();
```
agourlay, line 181: the test should not sort the list because being sorted is an invariant :)

agourlay: sorry misread the code, I am a dummy

agourlay (approved):

ffuugoo (approved):

src/actix/api_key.rs
timvisee: We can, but I created a [test](https://github.com/qdrant/qdrant/pull/3544/commits/45dd5e4a7cdc0986f6f14c9c4189452ff233a3ac) for it instead so our CI disallows breaking sorting.

I did try to check for sorting at compile time, but `is_sorted` and related functions are not `const` nor stabilized in the stdlib.

timvisee:
We decided to merge this as is now, and investigate ways to better test this in the future.

For testing, I've opened this issue: https://github.com/qdrant/qdrant/issues/3547

Maybe someone from our community can help with this.

Merging this now. :+1:

PR #3548, opened 2024-02-06
Feat/set payload by key

### LINKED ISSUE #3524: Support JsonPath in update payload commands

There are currently two ways to update point's payload:
- [set_payload](https://qdrant.tech/documentation/concepts/payload/#set-payload) to change only the value of a top level key
- [overwrite_payload](https://qdrant.tech/documentation/concepts/payload/#overwrite-payload) replace all values under a top level key

Given the payload:

```json
 "my_property": {
    "a": "a",
    "b": "b"
  }
```

I would like to add a nested key so that I get the following payload:

```json
"my_property": {
    "a": "a",
    "b": "b",
    "c": "c"
  }
```

Currently it is necessary to perform two calls:
- delete `my_property`
- set fully `my_property`

Perform the update in a single call by making the payload update API aware of nested key using JsonPath.

```json
{
    "key": "my_property",
    "payload": {
        "c": "c",
    },
    "points": [ 1 ]
}
```

For example, this should work as well to modify deeper objects:

```json
{
    "key": "some.deep.object",
    "payload": {
        "c": "c",
    },
    "points": [ 1 ]
}
```

**Additional context**

It should follow the JsonPath convention and infrastructure introduced when we [allowed JsonPath for PayloadSelector](https://github.com/qdrant/qdrant/pull/2756)

Shylock-Hg:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

Close #3524
/claim #3524

### DIFF
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -119,6 +119,7 @@ message SetPayloadPoints {
   optional PointsSelector points_selector = 5; // Affected points
   optional WriteOrdering ordering = 6; // Write ordering guarantees
   optional ShardKeySelector shard_key_selector = 7; // Option for custom sharding to specify used shard keys
+  optional string key = 8; // Option for indicate property of payload
 }
 
 message DeletePayloadPoints {
@@ -465,6 +466,7 @@ message PointsUpdateOperation {
       map<string, Value> payload = 1;
       optional PointsSelector points_selector = 2; // Affected points
       optional ShardKeySelector shard_key_selector = 3; // Option for custom sharding to specify used shard keys
+      optional string key = 4; // Option for path to indicate property
   }
   message DeletePayload {
       repeated string keys = 1;
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -466,7 +466,7 @@ message PointsUpdateOperation {
       map<string, Value> payload = 1;
       optional PointsSelector points_selector = 2; // Affected points
       optional ShardKeySelector shard_key_selector = 3; // Option for custom sharding to specify used shard keys
-      optional string key = 4; // Option for path to indicate property
+      optional string key = 4; // Option for indicate property of payload
   }
   message DeletePayload {
       repeated string keys = 1;
--- a/lib/segment/src/common/utils.rs
+++ b/lib/segment/src/common/utils.rs
@@ -231,6 +231,106 @@ pub fn get_value_from_json_map_opt<'a>(
     }
 }
 
+pub fn set_value_to_json_map<'a>(
+    path: &str,
+    dest: &'a mut serde_json::Map<String, Value>,
+    src: &'a serde_json::Map<String, Value>,
+) {
+    // check if leaf path element
+    match path.split_once('.') {
+        Some((element, rest_path)) => {
+            // check if targeting array
+            match parse_array_path(element) {
+                Some((array_element_path, array_index)) => {
+                    set_by_array_path(array_element_path, array_index, Some(rest_path), dest, src)
+                }
+                None => {
+                    // no array notation
+                    match dest.get_mut(element) {
+                        Some(Value::Object(map)) => set_value_to_json_map(rest_path, map, src),
+                        _ => (),
+                    }
+                }
+            }
+        }
+        None => match parse_array_path(path) {
+            Some((array_element_path, array_index)) => {
+                set_by_array_path(array_element_path, array_index, None, dest, src)
+            }
+            None => {
+                if path.is_empty() {
+                    merge_map(dest, src);
+                } else {
+                    if let Some(v) = dest.get_mut(path) {
+                        if let Value::Object(map) = v {
+                            merge_map(map, src);
+                        }
+                    } else {
+
+                    }
+                }
+            },
+        },
+    }
+}
+
+// Merge source map into destination map
+pub fn merge_map(dest: &mut serde_json::Map<String, Value>, source: &serde_json::Map<String, Value>) {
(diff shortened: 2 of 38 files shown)


xzfc:
> this assertion failed, HTTP should return error when key is empty string.
>  I can't found the reason, could give me some tips?

@Shylock-Hg By default, this endpoint responds with `HTTP 202 Accepted` immediately while running its job in the background. You need to pass `wait=true` as a query parameter to get its result/error, like this:
https://github.com/qdrant/qdrant/blob/eec5022831a2b93a8d3fd6db9072d152070fa3d7/openapi/tests/openapi_integration/test_payload_operations.py#L22

generall (changes requested):
Hey @Shylock-Hg, thanks for the PR!
I think interface matches our expectation, however there are a few problems with internals:


Proposed API change breaks guarantee of [Idempotence](https://en.wikipedia.org/wiki/Idempotence).
In other words, sending same request twice produce different result and different state of the storage.

For example:

Assume point 123 have empty payload.  When we do:

```
POST collections/test_collection/points/payload?wait=true
{
  "points": [123],
  "payload": {
    "a": 1
  },
  "key": "nested1"
}
```

It will generate point with:

```
 "payload": {
  "a": 1
}
```

But sending the same request twice will produce:

```
"error": "Wrong input: Can't find payload key segment: \"nested1\"."
```

Having Idempotence API is critical for Qdrant's distributed deployment mode and we can't ignore it.


On top of that there are a few other problems I will highlight in followup comments.

lib/segment/src/types.rs
```diff
@@ -874,14 +874,47 @@ pub struct Payload(pub Map<String, Value>);
...
-                Value::Null => self.0.remove(key),
-                _ => self.0.insert(key.to_owned(), value.to_owned()),
+                Value::Null => dest.remove(key),
+                _ => dest.insert(key.to_owned(), value.to_owned()),
             };
         }
     }
 
+    fn get_object_mut(&mut self, key: &str) -> OperationResult<&mut Map<String, Value>> {
+        let segs = key.split('.').collect::<Vec<&str>>();
```
generall, line 896: Json path might be more complicated than that. See `get_value_from_json_map_opt`

lib/segment/src/types.rs
```diff
@@ -874,14 +874,47 @@ pub struct Payload(pub Map<String, Value>);
...
+        if segs.is_empty() {
+            return Err(OperationError::WrongPayloadKey {
+                description: format!("Unexpected payload key: \"{key}\""),
```
generall, line 899: Empty `key` might be considered as "overwrite" for the whole object. In this case we are covering "overwrite" API as well.

lib/segment/src/types.rs
```diff
@@ -874,14 +874,47 @@ pub struct Payload(pub Map<String, Value>);
...
+            });
+        }
+        let mut v = &mut self.0;
+        for seg in segs {
+            let v1 = v.get_mut(seg).ok_or(OperationError::WrongPayloadKey {
+                description: format!("Can't find payload key segment: \"{seg}\"."),
```
generall, line 905: I think this should create required nested object instead of faiting with error.

lib/api/src/grpc/proto/points.proto
```diff
@@ -118,6 +118,7 @@ message SetPayloadPoints {
   optional PointsSelector points_selector = 5; // Affected points
   optional WriteOrdering ordering = 6; // Write ordering guarantees
   optional ShardKeySelector shard_key_selector = 7; // Option for custom sharding to specify used shard keys
+  optional string key = 8; // Option for indicate property of payload
```
generall, line 121: Would need to come up with better description for this field. (note mostly to myself)

lib/segment/src/common/utils.rs
```diff
@@ -220,6 +220,95 @@ pub fn get_value_from_json_map_opt<'a>(
     }
 }
 
+pub fn set_value_to_json_map<'a>(
```
generall, line 223: this function requires explicit testing

generall: Having openapi integration tests is nice, but it would be better to implement unit-tests here as well

lib/segment/src/payload_storage/simple_payload_storage_impl.rs
```diff
@@ -23,6 +23,21 @@ impl PayloadStorage for SimplePayloadStorage {
...
+    fn assign_by_key(
+        &mut self,
+        point_id: PointOffsetType,
+        payload: &Payload,
+        key: &str,
+    ) -> OperationResult<()> {
+        match self.payload.get_mut(&point_id) {
+            Some(point_payload) => point_payload.merge_by_key(payload, key)?,
+            None => {
+                self.payload.insert(point_id, payload.to_owned());
```
generall, line 35: This logic path ignores `key`, that will create a situation where idempotency will not be satisfied.

generall (changes requested):

lib/segment/src/payload_storage/simple_payload_storage_impl.rs
generall: > I plan to return error in this branch

no, it is a valid request. You just need to merge empty payload with given

generall: same as we would expect if stored payload doesn't have `key1.key2`

xzfc:
> This linter seems wrong. I never modify related files in this PR.

@Shylock-Hg It probably was caused basing your PR on the `master` branch. You should target your PRs onto `dev`, not `master`. I've rebased your PR to `dev`.

generall (approved):
LGTM once tests are green

PR #3550, opened 2024-02-06
Change `max_ack_version` to be `wal_keep_from`

timvisee:
Required for: <https://github.com/qdrant/qdrant/pull/3509>

We had a field called `max_ack_version` to specify limit what the last version was that could be acknowledged/truncated from the WAL. This PR changes to specify what WAL version to keep instead, shifting the logic by one.

To accompany this, I've renamed the field from `max_ack_version` to `wal_keep_from`.

This change is important because it now allows keeping the first entry in the WAL by setting it to `0`, while this was not possible before. It is required for <https://github.com/qdrant/qdrant/pull/3509>.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -55,10 +55,10 @@ impl QueueProxyShard {
     pub fn new(
         wrapped_shard: LocalShard,
         remote_shard: RemoteShard,
-        max_ack_version: Arc<AtomicU64>,
+        wal_keep_from: Arc<AtomicU64>,
     ) -> Self {
         Self {
-            inner: Some(Inner::new(wrapped_shard, remote_shard, max_ack_version)),
+            inner: Some(Inner::new(wrapped_shard, remote_shard, wal_keep_from)),
         }
     }
 
@@ -144,7 +144,7 @@ impl QueueProxyShard {
             .inner
             .take()
             .expect("Queue proxy has already been finalized");
-        queue_proxy.set_max_ack_version(None);
+        queue_proxy.set_wal_keep_from(None);
 
         (queue_proxy.wrapped_shard, queue_proxy.remote_shard)
     }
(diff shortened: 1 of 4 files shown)


lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -373,16 +374,16 @@ impl Inner {
...
     /// truncated before all these update operations have been flushed.
-    /// Using this function we set the WAL not to truncate past the given point.
+    /// Using this function we set the WAL not to acknowledge and truncate from a specific point.
     ///
     /// Providing `None` will release this limitation.
-    fn set_max_ack_version(&self, max_version: Option<u64>) {
+    fn set_wal_keep_from(&self, max_version: Option<u64>) {
         let max_version = max_version.unwrap_or(u64::MAX);
-        self.max_ack_version.store(max_version, Ordering::Relaxed);
+        self.wal_keep_from.store(max_version, Ordering::Relaxed);
```
ffuugoo, line 386: I'd rename `max_version` (both the parameter and `let max_version`) to something else (just `version`?), cause it somehow don't make sense with the new method name, IMO.

lib/collection/src/wal.rs
```diff
@@ -162,8 +162,7 @@ impl<'s, R: DeserializeOwned + Serialize + Debug> SerdeWal<R> {
     /// # Arguments
     ///
     /// * `until_index` - the newest no longer required record sequence number
-    ///
-    pub fn ack(&mut self, until_index: u64) -> Result<()> {
+    pub(super) fn ack(&mut self, until_index: u64) -> Result<()> {
```
ffuugoo, line 165: Why not just `pub`? 🪡

ffuugoo (approved):
Some minor comments, but LGTM, I think

lib/collection/src/shards/queue_proxy_shard.rs
timvisee: Yes, that's better. I forgot about this during refactoring.

lib/collection/src/wal.rs
timvisee: We can, but I'd like to limit access to this as much as possible. We should not blindly call `ack` if this `wal_keep_from` parameter is set, because we need to consider that as well.

PR #3551, opened 2024-02-06
Allow using queue proxy on WAL in the past

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/3550>
Required for: https://github.com/qdrant/qdrant/pull/3509

This widens the usability of our queue proxy to allow using it for queuing and transferring WAL entries that are in the past. Before we could only queue and transfer new WAL entries after creating the queue proxy.

This is essential for <https://github.com/qdrant/qdrant/pull/3509>, as it will be used for transferring the WAL diff. I choose to use the queue proxy for this because it already had all the bells and whistles to handle this transfer safely. 

### Tasks
 - [x] Merge <https://github.com/qdrant/qdrant/pull/3550>
 - [x] Rebase this on `dev`
 - [x] Target this PR to merge into `dev`
 - [x] Undraft this PR

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -296,8 +296,9 @@ struct Inner {
     pub(super) wrapped_shard: LocalShard,
     /// Wrapped remote shard, to transfer operations to.
     pub(super) remote_shard: RemoteShard,
-    /// ID of the last WAL operation we consider transferred.
-    last_update_idx: AtomicU64,
+    /// ID of the WAL operation we should transfer next. We consider everything before it to be
+    /// transferred.
+    transfer_from: AtomicU64,
     /// Lock required to protect transfer-in-progress updates.
     /// It should block data updating operations while the batch is being transferred.
     update_lock: Mutex<()>,
@@ -314,18 +315,18 @@ impl Inner {
         remote_shard: RemoteShard,
         wal_keep_from: Arc<AtomicU64>,
     ) -> Self {
-        let last_idx = wrapped_shard.wal.lock().last_index();
+        let start_from = wrapped_shard.wal.lock().last_index();
 
         let shard = Self {
             wrapped_shard,
             remote_shard,
-            last_update_idx: last_idx.into(),
+            transfer_from: start_from.into(),
             update_lock: Default::default(),
             wal_keep_from,
         };
 
         // Keep all new WAL entries so we don't truncate them off when we still need to transfer
-        shard.set_wal_keep_from(Some(last_idx + 1));
+        shard.set_wal_keep_from(Some(start_from));
 
         shard
     }
@@ -339,7 +340,7 @@ impl Inner {
         let shard = Self {
             wrapped_shard,
             remote_shard,
-            last_update_idx: version.into(),
+            transfer_from: version.into(),
             update_lock: Default::default(),
             wal_keep_from,
         };
@@ -364,9 +365,9 @@ impl Inner {
     pub async fn transfer_all_missed_updates(&self) -> CollectionResult<()> {
         while !self.transfer_wal_batch().await? {}
 
-        // Set max acknowledged version for WAL to last item we transferred
-        let last_idx = self.last_update_idx.load(Ordering::Relaxed);
-        self.set_wal_keep_from(Some(last_idx + 1));
+        // Set the WAL version to keep to the next item we should transfer
+        let transfer_from = self.transfer_from.load(Ordering::Relaxed);
+        self.set_wal_keep_from(Some(transfer_from));
 
         Ok(())
     }
@@ -387,13 +388,13 @@ impl Inner {
     /// idempotent.
     async fn transfer_wal_batch(&self) -> CollectionResult<bool> {
         let mut update_lock = Some(self.update_lock.lock().await);
-        let start_index = self.last_update_idx.load(Ordering::Relaxed) + 1;
+        let transfer_from = self.transfer_from.load(Ordering::Relaxed);
 
         // Lock wall, count pending items to transfer, grab batch
         let (pending_count, batch) = {
             let wal = self.wrapped_shard.wal.lock();
-            let items_left = wal.last_index().saturating_sub(start_index - 1);
-            let batch = wal.read(start_index).take(BATCH_SIZE).collect::<Vec<_>>();
+            let items_left = wal.last_index().saturating_sub(transfer_from - 1);
+            let batch = wal.read(transfer_from).take(BATCH_SIZE).collect::<Vec<_>>();
             (items_left, batch)
         };
 
@@ -427,7 +428,7 @@ impl Inner {
             }
 
             if let Some(idx) = last_idx {
-                self.last_update_idx.store(idx, Ordering::Relaxed);
+                self.transfer_from.store(idx + 1, Ordering::Relaxed);
             }
         }
 
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -315,7 +315,7 @@ impl Inner {
         remote_shard: RemoteShard,
         wal_keep_from: Arc<AtomicU64>,
     ) -> Self {
-        let start_from = wrapped_shard.wal.lock().last_index();
+        let start_from = wrapped_shard.wal.lock().last_index() + 1;
 
         let shard = Self {
             wrapped_shard,
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -75,7 +75,7 @@ impl QueueProxyShard {
     /// # Errors
     ///
     /// This fails if the given `version` is not in bounds of our current WAL. If the given
-    /// `version` is to old or to new, queue proxy creation is rejected.
+    /// `version` is too old or too new, queue proxy creation is rejected.
     pub fn new_from_version(
         wrapped_shard: LocalShard,
         remote_shard: RemoteShard,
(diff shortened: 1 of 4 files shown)


lib/collection/src/shards/replica_set/shard_transfer.rs
```diff
@@ -74,10 +74,24 @@ impl ShardReplicaSet {
...
+    ///
     /// # Cancel safety
     ///
     /// This method is cancel safe.
-    pub async fn queue_proxify_local(&self, remote_shard: RemoteShard) -> CollectionResult<()> {
+    pub async fn queue_proxify_local(
+        &self,
+        remote_shard: RemoteShard,
+        from_version: Option<u64>,
+    ) -> CollectionResult<()> {
```
timvisee, line 94: The essential bit is here. We can now construct a queue proxy from any arbitrary WAL version (that is in range).

Our WAL diff transfer will use this with the first version of the WAL diff that needs to be transferred.

lib/collection/src/shards/replica_set/shard_transfer.rs
```diff
@@ -145,10 +159,34 @@ impl ShardReplicaSet {
...
+        match proxy_shard {
+            // All good, insert queue proxy shard
+            Ok(proxy_shard) => {
+                let _ = local.insert(Shard::QueueProxy(proxy_shard));
+                Ok(())
+            }
+            Err((local_shard, err)) => {
+                log::warn!("Failed to queue proxify shard, reverting to local shard: {err}");
+                let _ = local.insert(Shard::Local(local_shard));
+                Err(err)
```
agourlay, line 187: What will the caller of this function do in that case?

timvisee: That is up to the caller, but in our current code base we abort a shard transfer when this returns an error.

The important thing here is that reverting back to local brings it into the original state. That is required so the shard transfer (or something else) can be retried again.

lib/collection/src/shards/replica_set/shard_transfer.rs
Vivesh2911: Greetings! Excited to contribute to #3509. Enhancing the queue proxy for WAL diff transfer, ensuring efficient and secure data transmission. Considering extending queue proxy to handle past versions, enabling seamless transfer of historical WAL entries.

lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -62,6 +65,43 @@ impl QueueProxyShard {
...
+    ///
+    /// This queues all (existing) updates from a specific WAL `version` and onwards. In other
+    /// words, this will ensure we transfer updates we already have and all new updates from a
+    /// specific point in our WAL. The `version` may be in the past, but must always be within
+    /// range of the current WAL.
+    ///
+    /// # Errors
+    ///
+    /// This fails if the given `version` is not in bounds of our current WAL. If the given
+    /// `version` is to old or to new, queue proxy creation is rejected.
```
ffuugoo, line 78: Small typo.

```suggestion
    /// This fails if the given `version` is not in bounds of our current WAL. If the given
    /// `version` is too old or too new, queue proxy creation is rejected.
```

lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -62,6 +65,43 @@ impl QueueProxyShard {
...
+        version: u64,
+    ) -> Result<Self, (LocalShard, CollectionError)> {
+        // Lock WAL until we've successfully created the queue proxy shard
+        let wal = wrapped_shard.wal.clone();
+        let wal_lock = wal.lock();
+
+        // If start version is not in current WAL bounds, we cannot reliably transfer WAL
+        let (first_idx, last_idx) = (wal_lock.first_physical_index(), wal_lock.last_index());
+        if version < first_idx || version > last_idx {
+            return Err((wrapped_shard, CollectionError::service_error(format!("Cannot create queue proxy shard from version {version} because it is out of WAL bounds ({first_idx}..={last_idx})"))));
```
ffuugoo, line 92: `[{first_idx}, {last_ids}]`? Like, math notation for inclusive range?
(Just a suggestion, though. I'm cool with either.)

lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -290,6 +330,26 @@ impl Inner {
...
+        let shard = Self {
+            wrapped_shard,
+            remote_shard,
+            last_update_idx: version.into(),
+            update_lock: Default::default(),
+            wal_keep_from,
+        };
+
+        // Keep all WAL entries from `version` so we don't truncate them off when we still need to transfer
+        shard.set_wal_keep_from(Some(version));
```
ffuugoo, line 348: In `Inner::new` we keep `last_update_idx` as-is, but `set_wal_keep_from(last_idx + 1)`. Can there be an "off by 1" error hidden here? E.g., should we do `last_update_idx = version - 1`?

(Note, this is a *question*, not a suggestion. Just noted it, cause `ClockSet`/`ClockMap` do have this distinction, maybe this case is similar?)

ffuugoo (approved):
Few minor comments, but LGTM otherwise

lib/collection/src/shards/queue_proxy_shard.rs
timvisee: Good idea, I'll do that to make it more explicit.

I'll also do `(first_idx..=last_idx).contains()` to more closely match the condition.

timvisee: It'll be `(first_idx, last_idx)` by the way, as it is inclusive.

lib/collection/src/shards/queue_proxy_shard.rs
timvisee: I'd prefer to keep it this way. That is because both variables/parameters describe two different things:

- `last_update_idx` describes the last update we have actually transferred, which matches the last operation index
- `wal_keep_from` describes what version we want to keep

We can rename `last_update_idx` to `first_not_transferred_idx` or something similar, but I don't really like this. Also, we cannot change `wal_keep_from`.

Before, `wal_keep_from` was named `max_ack_version`, in which case the two matched up better.

> Also, we probably should rename `last_update_idx` to something more meaningful

What about `last_transferred_idx`?

ffuugoo: I still don't get why we do `+1` in one case, and we don't in another.

Like, my concern here, is that if there's some other block of logic (outside `new`/`new_from_version`; or maybe we will add one in the future) handling `last_update_idx` and making this assumption of "we keep WAL from `last_update_idx + 1`", we may have an annoying bug. And I'm not sure if we can't easily distinguish these two cases looking only at the `QueueProxy` fields.

ffuugoo: Though, if you sure that this is fine, then it's fine by me. At least we know where to look if we suddenly see first operation not being transferred or something. 😂

lib/collection/src/shards/queue_proxy_shard.rs
ffuugoo: `[start, end]` is inclusive. `(x, y)` is exclusive. But I think we may better leave it as `..=` then? :D
https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints

lib/collection/src/shards/queue_proxy_shard.rs
timvisee: There was a mistake indeed. In case of `new_from_version` the `last_update_idx` assignment was wrong.

I've fixed this and changed the field to `transfer_from` to make it work.

lib/collection/src/shards/queue_proxy_shard.rs
timvisee: :man_facepalming:

timvisee: Fixed.

lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -327,13 +388,13 @@ impl Inner {
...
         let mut update_lock = Some(self.update_lock.lock().await);
-        let start_index = self.last_update_idx.load(Ordering::Relaxed) + 1;
+        let transfer_from = self.transfer_from.load(Ordering::Relaxed);
 
         // Lock wall, count pending items to transfer, grab batch
         let (pending_count, batch) = {
             let wal = self.wrapped_shard.wal.lock();
-            let items_left = wal.last_index().saturating_sub(start_index - 1);
-            let batch = wal.read(start_index).take(BATCH_SIZE).collect::<Vec<_>>();
+            let items_left = wal.last_index().saturating_sub(transfer_from - 1);
```
ffuugoo, line 396: 1) Do we still need to do `transfer_from - 1`, if we don't do `start_index = self.last_update_idx + 1` anymore?
2) If we do still need `-1`, then what if `transfer_from = 0`?

ffuugoo (commented):
One more `+/-1` question... (Sorry!)

lib/collection/src/shards/queue_proxy_shard.rs
timvisee: I think it should be like this now:

```rust
let items_left = (wal.last_index() + 1).saturating_sub(transfer_from);
```

So if:
- `transfer_from` is 8
- `wal.last_index` is 12

We have `(12 + 1) - 8 = 5`. Which matches transferring 8, 9, 10, 11, 12.

What do you think?

ffuugoo: Yeah, seems reasonable. 👌

ffuugoo (approved):

PR #3555, opened 2024-02-07
Report shard transfer progress

### LINKED ISSUE #3433: Report shard transfer progress
As an operator I have no idea if shard transfers are making good progress and when they will be done.

I want to have:
- the progress of on-going shard transfers in percentage  
- an estimated time of completion assuming constant speed

Either print the progress in the logs every minute or even better output it in the shard transfers section of the cluster information.

xzfc:
Closes #3433.

The shard progress state is tracked locally inside `TransferTasksPool` (aka `Collection::transfer_tasks`).
Since it's local, only the sender node shows the progress in the result of `GET collections/{name}/cluster`.

Output of `GET collections/{name}/cluster`:
```json
{"result": {
  "shard_transfers": [
    {
      "shard_id": 1,
      "from": 3351940210608703,
      "to": 7522668664620757,
      "sync": false,
      "method": "stream_records",
      ↓ ↓ ↓   This field is added   ↓ ↓ ↓
      "comment": "Transferring records (136000/193725), started 15s ago, ETA: 6.83s"
      ↑ ↑ ↑   This field is added  ↑ ↑ ↑
    }
  ],
  "peer_id": 3351940210608703,
  "local_shards": [
    {
      "shard_id": 1,
      "points_count": 178839,
      "state": "Active"
    }, …
  ],
  "remote_shards": [
    {
      "shard_id": 1,
      "peer_id": 7522668664620757,
      "state": "Partial"
    }, …
  ], …
}, … }
```

---

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -8704,11 +8704,13 @@
             "minimum": 0
           },
           "from": {
+            "description": "Source peer id",
             "type": "integer",
             "format": "uint64",
             "minimum": 0
           },
           "to": {
+            "description": "Destination peer id",
             "type": "integer",
             "format": "uint64",
             "minimum": 0
@@ -8726,6 +8728,32 @@
                 "nullable": true
               }
             ]
+          },
+          "started_at": {
+            "description": "Time when the transfer was started. Available only on the source peer.",
+            "type": "string",
+            "format": "date-time",
+            "nullable": true
+          },
+          "points_transferred": {
+            "description": "Amount of points transferred. Available only on the source peer.",
+            "type": "integer",
+            "format": "uint",
+            "minimum": 0,
+            "nullable": true
+          },
+          "points_total": {
+            "description": "Total amount of points to transfer. Available only on the source peer.",
+            "type": "integer",
+            "format": "uint",
+            "minimum": 0,
+            "nullable": true
+          },
+          "eta_seconds": {
+            "description": "Estimated time of arrival. Available only on the source peer. Null when the transfer is stalled or not started yet.",
+            "type": "number",
+            "format": "double",
+            "nullable": true
           }
         }
       },
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -8729,30 +8729,9 @@
               }
             ]
           },
-          "started_at": {
-            "description": "Time when the transfer was started. Available only on the source peer.",
+          "comment": {
+            "description": "A human-readable report of the transfer progress. Available only on the source peer.",
             "type": "string",
-            "format": "date-time",
-            "nullable": true
-          },
-          "points_transferred": {
-            "description": "Amount of points transferred. Available only on the source peer.",
-            "type": "integer",
-            "format": "uint",
-            "minimum": 0,
-            "nullable": true
-          },
-          "points_total": {
-            "description": "Total amount of points to transfer. Available only on the source peer.",
-            "type": "integer",
-            "format": "uint",
-            "minimum": 0,
-            "nullable": true
-          },
-          "eta_seconds": {
-            "description": "Estimated time of arrival. Available only on the source peer. Null when the transfer is stalled or not started yet.",
-            "type": "number",
-            "format": "double",
             "nullable": true
           }
         }
--- a/lib/collection/src/common/eta_calculator.rs
+++ b/lib/collection/src/common/eta_calculator.rs
@@ -1,0 +1,71 @@
+use std::time::{Duration, Instant};
+
+/// A progress ETA calculator.
+/// Calculates the ETA roughly based on the last ten seconds of measurements.
+pub struct EtaCalculator {
+    ring: [(Instant, usize); Self::SIZE],
+    ring_pos: usize,
+}
+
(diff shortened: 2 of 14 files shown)


docs/redoc/master/openapi.json
```diff
@@ -8608,6 +8608,23 @@
                 "nullable": true
               }
             ]
+          },
+          "started_at": {
+            "type": "string",
+            "format": "date-time",
+            "nullable": true
+          },
+          "records_done": {
```
timvisee, line 8617: Nitpick, but `records_transferred` may be a bit better. Whichever you prefer!

lib/collection/src/shards/transfer/stream_records.rs
```diff
@@ -64,7 +73,12 @@ pub(super) async fn transfer_stream_records(
 
         if offset.is_none() {
```
timvisee, line 74: We can also do this without branching like:

```rust
let mut progress = progress.lock();
progress.records_done = max(progress.records_done + TRANSFER_BATCH_SIZE, progress.records_total);

File: lib/collection/src/shards/transfer/stream_records.rs [hunk 3]
```diff
@@ -37,7 +41,13 @@ pub(super) async fn transfer_stream_records(
             )));
         };
 
-        replica_set.proxify_local(remote_shard).await?;
+        let points_count = replica_set
+            .proxify_local(remote_shard)
+            .await?
+            .wrapped_shard
+            .read_filtered(None)?
+            .len();
```
timvisee, line 49: I'd prefer to do these two things separate. Maybe we can utilize `replica_set.count_local` for it.

```suggestion
        replica_set.proxify_local(remote_shard).await?;
            
        let points_count = replica_set.count_local(...).await?;
```

lib/collection/src/common/eta_calculator.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub fn estimate(&self, target_progress: usize) -> Duration {
+        let now = Instant::now();
+        let (_, current_progress) = self.ring[self.ring_pos];
+        let (prev_time, old_progress) = self.ring[(self.ring_pos + Self::SIZE - 1) % Self::SIZE];
+        match target_progress.checked_sub(current_progress) {
+            None => Duration::from_secs(0),
+            Some(value_diff) => {
+                let rate = value_diff as f64 / (current_progress - old_progress) as f64;
+                let time_diff = (now - prev_time).as_secs_f64();
+                Duration::try_from_secs_f64(time_diff * rate).unwrap_or(Duration::MAX)
```
timvisee, line 46: If current and old are 0, we get weird results.

What about returning `None` instead of `MAX` in such cases? That way we prevent a weird duration spike at the beginning.

timvisee (changes requested):

lib/collection/src/shards/transfer/stream_records.rs
xzfc: Fixed.

lib/collection/src/common/eta_calculator.rs
xzfc: Updated.

agourlay:
```json
"eta_seconds": 73.496774076
```

I think truncating to two decimals should be enough precision :)

lib/collection/src/common/eta_calculator.rs
```diff
@@ -0,0 +1,71 @@
+use std::time::{Duration, Instant};
+
+/// A progress ETA calculator.
+/// Calculates the ETA roughly based on the last ten seconds of measurements.
+pub struct EtaCalculator {
```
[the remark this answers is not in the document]
agourlay: I find the ring based design a bit difficult to review TBH

It is fine that you picked this, it is your call :+1: 

In that case I'd love a few unit tests for it because there is a lot of array indexing manipulations going on.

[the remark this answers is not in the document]
timvisee: Also, would it be useful to utilize any of the [`ringbuffer`](https://docs.rs/ringbuffer/)s so we don't have to manage it ourselves?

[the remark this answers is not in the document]
xzfc: Added a dependency on `ringbuffer` and some commentary so it should be more clear.
Also, added a test.

xzfc:
> I think truncating to two decimals should be enough precision :)

Done.

agourlay (approved):
Thanks for the tests and switching to the ringbuffer crate :+1:

xzfc:
After a discussion, switched to a human-readable text format, e.g.
```json
"comment": "Transferring records (136000/193725), started 15s ago, ETA: 6.83s"
```

agourlay:
> After a discussion, switched to a human-readable text format, e.g.

Can you please share why this decision was taken? :)

xzfc:
@agourlay 
The text-based format is enough for debugging purposes, and it is flexible enough to cover other transfer types.
OTOH, a strict schema will be too complicated and we don't see any meaningful way to use it in automation.

timvisee (approved):
:raised_hands:

Issue #3547, opened 2024-02-06
Add test for validating REST API routing & service order


timvisee:
**Is your feature request related to a problem? Please describe.**
The order of services we define for our REST API is important. If ordering is incorrect, there can be problems resolving a route due to pattern collisions.

Read more about this here: <https://github.com/qdrant/qdrant/issues/3543>.

The PR <https://github.com/qdrant/qdrant/pull/3544> fixes this problem, but we want to make sure that ordering isn't accidentally broken afterwards.

**Describe the solution you'd like**
To prevent this, we would like to add a test that asserts routing and ordering of services is correct.

We imagine the following:
- create app definition for actix (if required)
- set up a few dummy requests, such as `POST /collections/some/points/scroll`
- get route pattern for request, possibly using `match_pattern`
- ensure pattern matches what we expect

We imagine setting up a few dummy requests, such as to scroll, and asserting that the right route pattern is matched.

Most importantly, we want this new test to fail if ordering in <https://github.com/qdrant/qdrant/pull/3544> would be reverted.

**Describe alternatives you've considered**
We could also add some integration test to call each read-only endpoint asserting the responses. But that would require quite a bit of work, and it would be less complete than testing actix routing and service ordering directly.

**Additional context**
I tried to implement a proof of concept for this test, but sadly that was not sufficient. In the snippet below we define the app having routes, and we define a request. But the `match_pattern` function to select what route pattern is matched does not know about our `app`. There does not seem to be a way to connect the two together.

```rust
#[cfg(test)]
mod tests {
    use actix_web::{http::header::ContentType, test, App};

    use super::*;

    #[actix_web::test]
    async fn test_routing_order() {
        let app = test::init_service(App::new()
                .service(index)
                .configure(config_collections_api)
                .configure(config_snapshots_api)
                .configure(config_update_api)
                .configure(config_cluster_api)
                .configure(config_service_api)
                .configure(config_search_api)
                .configure(config_recommend_api)
                .configure(config_discovery_api)
                .configure(config_shards_api)
                .service(scroll_points)
                .service(count_points)
                .service(get_point)
                .service(get_points)).await;

        let req = test::TestRequest::post()
            .app_data(app)
            .uri("/collections/test_collection/points/scroll")
            .insert_header(ContentType::json())
            .to_srv_request();

        assert_eq!(req.match_pattern().unwrap(), "/collections/{collection}/points/scroll");
    }
}
```

Also, actix web has some testing facilities that may be helpful:
[...]

PR #3556, opened 2024-02-07
Add regex dependency and implement match_pattern fuction

kemkemG0:
should resolve: #3543 #3547

I opened an issue (https://github.com/qdrant/qdrant-web-ui/issues/158) regarding a read-only pattern matching problem. 

Actix's `match_pattern()` function doesn't distinguish between `GET` and `POST` methods, causing `GET /collections/{name}/points/{id}` to be matched when trying `req.match_pattern()` with `POST /collections/{name}/points/scroll`. 

The issue was addressed in #3544 by @timvisee, however, challenges related to testability and instability still persist.

Given Actix's `match_pattern()` limitations, as discussed in #3547, I've developed our own `match_pattern_to_request_path` function with a straightforward regex implementation. 

This function aims to address the issues discussed in the aforementioned open issues.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

/claim #3547

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3893,6 +3893,7 @@ dependencies = [
  "raft",
  "raft-proto",
  "rand 0.8.5",
+ "regex",
  "reqwest",
  "rstack-self",
  "rustls",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,8 @@ rustls-pemfile = "2.0.0"
 prometheus = { version = "0.13.3", default-features = false }
 validator = { version = "0.16", features = ["derive"] }
 
+regex = "1.8.4"
+
 # Consensus related crates
 raft = { version = "0.7.0", features = ["prost-codec"], default-features = false }
 slog = { version="2.7.0", features = ["max_level_trace", "release_max_level_debug"] }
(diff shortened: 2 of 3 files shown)


timvisee (changes requested):
Thank you for giving this a shot!

However, this replaces the route matching logic. We don't want to do that. It adds another layer of complexity and another thing to configure around routing, which we should avoid.

Also, your approach still has a similar problem:
- This: `/collections/something/points/scroll`
- Does still match: `/collections/something/points/{id}`
- And: `/collections/something/points/scroll`

Instead, we want to confirm that our existing logic for route matching returns the results we expect.

kemkemG0:
@timvisee 

Thank you for your feedback!

I understand your concerns. However, my intention was not to replace the route matching logic but rather to address the specific issue of pattern matching with `READ_ONLY_POST_PATTERNS` without altering the core routing logic. 

The challenge with the original match_pattern function lies in its method-agnostic approach, which disregards the requested HTTP method (`GET` or `POST`) and attempts to match patterns solely based on paths, leading to potential pattern collisions across different methods. 

My proposed function is designed  exclusively for matching the patterns of read-only `POST` requests within our predefined `READ_ONLY_POST_PATTERNS`, thereby avoiding namespace conflicts, as I understand it. ( It might be worth considering a re-naming this function to more clearly convey its purpose. )

While adding tests to the original implementation is indeed valuable and something I'm actively considering, I hope we can also reconsider the solution I've proposed, with the provided context and rationale in mind : )

PR #3560, opened 2024-02-07
Pool larger memory batches for sparse vector search

agourlay:
This PR improves the performance of the sparse vector search by:
- increasing the size of the underlying working batch (from 1k to 10k)
- pooling the memory for working batches between searches

The rational being that a larger batch improves performance but we want to mitigate the cost of memory allocation.

# Performance

- 8 segments
- index.on_disk=false
- 6980 queries

```
DEV
min: 90.75 millis
50p: 179.43 millis
95p: 310.99 millis
99p: 408.16 millis
999p: 462.66 millis
max: 533.39 millis
```

```
PR
min: 91.04 millis
50p: 188.84 millis
95p: 244.93 millis
99p: 272.17 millis
999p: 301.74 millis
max: 334.95 millis
```

DEV             |  PR
sparse_bench_full_dev  | sparse_bench_full_PR


lib/sparse/src/index/search_context.rs
```diff
@@ -17,28 +18,29 @@ pub struct IndexedPostingListIterator<'a> {
     query_weight: DimWeight,
 }
 
-/// Making this larger makes the search faster but uses more memory
-const ADVANCE_BATCH_SIZE: usize = 1_000;
+/// Making this larger makes the search faster but uses more (pooled) memory
+const ADVANCE_BATCH_SIZE: usize = 10_000;
```
agourlay, line 22: increasing the batch size

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -41,6 +42,7 @@ pub struct SparseVectorIndex<TInvertedIndex: InvertedIndex> {
     searches_telemetry: SparseSearchesTelemetry,
     is_appendable: bool,
     pub indices_tracker: IndicesTracker,
+    pub scores_memory_pool: ScoresMemoryPool,
```
generall, line 45: Doesn't need to be public

lib/sparse/src/index/search_context.rs
generall: 8192 ?

lib/sparse/src/common/scores_memory_pool.rs
```diff
@@ -0,0 +1,66 @@
...
+
+lazy_static! {
+    /// Max number of `PooledScores` to preserve in memory
+    ///
+    /// If there are more concurrent requests, a new temporary vec is created dynamically.
+    /// This limit is implemented to prevent memory leakage.
+    /// It matches the number of logical CPU cores, to best represent the expected number of
+    /// concurrent requests. Clamped between 16 and 128 to prevent extreme values.
+    static ref POOL_KEEP_LIMIT: usize = common::cpu::get_num_cpus().clamp(16, 128);
+}
```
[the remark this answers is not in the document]
agourlay: I like the idea of moving it to `default.rs` :+1:

lib/sparse/src/index/search_context.rs
agourlay: I'd rather not go under 10k as the performance seems to respond well to the value.

If you really want a power of 2 I'd vote for `16_384` :)

lib/sparse/src/common/scores_memory_pool.rs
```diff
@@ -0,0 +1,66 @@
...
+
+impl<'a> Drop for PooledScoresHandle<'a> {
+    fn drop(&mut self) {
+        self.pool.return_back(std::mem::take(&mut self.scores));
+    }
+}
+
+#[derive(Debug)]
+pub struct ScoresMemoryPool {
+    pool: RwLock<Vec<PooledScores>>,
```
IvanPleshkov, line 37: We use `RwLock` for write access only. May we change it to regulal `Mutex` instead?

lib/sparse/src/common/scores_memory_pool.rs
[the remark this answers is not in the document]
IvanPleshkov: This is the same as in visited pool in HNSW. May we use already defined limit with the same value instead of new one?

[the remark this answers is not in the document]
timvisee: > This is the same as in visited pool in HNSW. May we use already defined limit with the same value instead of new one?

Yes. That's basically what I suggested. :smiley:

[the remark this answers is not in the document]
IvanPleshkov: Yep, true😃 didn't read your suggestion properly =)

lib/sparse/src/index/search_context.rs
agourlay: I will bench it quickly :scientist:

timvisee (approved):

lib/sparse/src/index/search_context.rs
```diff
@@ -19,7 +19,7 @@ pub struct IndexedPostingListIterator<'a> {
 }
 
 /// Making this larger makes the search faster but uses more (pooled) memory
-const ADVANCE_BATCH_SIZE: usize = 10_000;
+const ADVANCE_BATCH_SIZE: usize = 16_384;
```
timvisee, line 22: By the way, [`pow`](https://doc.rust-lang.org/std/primitive.u64.html#method.pow) is `const` these days, so we can even do:

```rust
const ADVANCE_BATCH_SIZE: usize = 2.pow(14);
```

Totally fine to leave it as is though, I don't suggest to change it. :+1:

lib/sparse/src/index/search_context.rs
agourlay: Results of the bench:

- 16k performs overall worst than 10k
- 8k has a significant larger max value than 10k

Decision: I will stick to 10k

agourlay:
All comments are addressed :)

IvanPleshkov (approved):

PR #3562, opened 2024-02-07
Order-by: use pre-filtering for high-constraining filters

coszio:
Tracked by: #3521 

Currently all `order_by` calls use a streaming approach from the BTree of the numeric index. But this is inefficient when there is already a high-constraining filter in the request, which can potentially make the stream have to iterate through an excessive amount of values to fulfill the requested `limit`. In this case, it makes more sense to get iterate through the filtered points first, and then sort the top points only.

This is the same optimization that is done without `order_by` on scroll calls, so it reuses the same estimation logic to choose streaming vs pre-filtering.

I have confirmed locally that this yields performance improvements by manually querying with web-ui dashboard. It can be between 4-10x faster, compared to `dev` branch.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/data_types/order_by.rs
+++ b/lib/segment/src/data_types/order_by.rs
@@ -58,11 +58,13 @@ impl OrderBy {
         self.direction.unwrap_or_default()
     }
 
-    pub fn value_offset(&self) -> f64 {
-        self.start_from.unwrap_or_else(|| match self.direction() {
-            Direction::Asc => f64::NEG_INFINITY,
-            Direction::Desc => f64::INFINITY,
-        })
+    pub fn start_from(&self) -> OrderingValue {
+        self.start_from
+            .unwrap_or_else(|| match self.direction() {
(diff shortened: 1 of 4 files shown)


lib/segment/src/data_types/order_by.rs
```diff
@@ -58,11 +58,13 @@ impl OrderBy {
...
-            Direction::Asc => f64::NEG_INFINITY,
-            Direction::Desc => f64::INFINITY,
-        })
+    pub fn start_from(&self) -> OrderingValue {
+        self.start_from
+            .unwrap_or_else(|| match self.direction() {
+                Direction::Asc => f64::NEG_INFINITY,
+                Direction::Desc => f64::INFINITY,
+            })
+            .into()
```
coszio, line 67: This will have to change with #3540

generall (approved):
Some benches:
```
// Branch: "time": 0.000431166
// Dev: "time": 0.013284618
```

PR #3563, opened 2024-02-08
Text index inverted index compression

IvanPleshkov:
Compressing postings list for text field index.

Compression is implemented using `bitpacking` crate: https://github.com/quickwit-oss/bitpacking. Compression works for immutable state only. Mutable state uses old postings list implementation.

Because `bitpacking` compresses only fixed-len chunks (128 of `u32` in this PR), compressed posting list is divided into chunks, each chunk is compressed separately and compressed data is flattened on one large collection.

Because `bitpacking` provides only pack/unpack methods, we have to decompress one chunk is we want to check if value is presented in compressed posting list. For intersection of posting list there are special helping visitor, who reuse decompressed chunks and shorten search ranges.

### RAM difference

RAM difference was tested on https://storage.googleapis.com/common-datasets-snapshots/arxiv_abstracts-3083016565637815127-2023-06-02-07-26-29.snapshot

Measuring method: `/proc/<PID>/status`

VmRSS on dev: `10_116 MB`
VmRSS on branch: `9_696 MB`
There is 420MB difference in VmRSS.

RssAnon on dev: `10_041 MB`
RssAnon on branch: `9_621 MB`
There is 420MB difference in RssAnon.

### Performance difference

Performance was measured on the snapshot mentioned above.
As measurement strategy, I called using Rust client 1000 times filtered recommendation API with 2 different positive points in single thread mode. Call is like:
```
curl -X POST "http://$QDRANT_HOST/collections/text/points/recommend" \
  -H 'Content-Type: application/json' \
  --data-raw '{
      "positive": [ "000004ce-7a38-478f-8d83-2900a72e1d8d", "000004ce-7a38-478f-8d83-6700a12e173t" ],
      "limit": 10,
      "filter": {
          "should": [
              {
                  "key": "abstract",
                  "match": {
                      "text": "<TEST_TEXT>"
                  }
              }
          ]
      },
      "with_payload": false,
      "with_vector": false
    }' | jq
```

For small-cardinality `TEST_TEXT` is `Immediate benefits`.
For large-cardinality `TEST_TEXT` is `the a`.
I checked with telemetry that this texts cover small- and large-cardinality cases.

P95 search time for `Immediate benefits` request:
```
DEV: 0.000173356
PR: 0.000119321
```

P95 search time for `The a` request:
```
DEV: 0.030939524
PR: 0.02868633
```

There is a performance boost. Because such measurement is unexpected, I checked that all search results, used for benchmarking, are the same.
Performance boost can be explained for small-cardinality case, where `visitor` pattern shorten search ranges while postings intersection. For HNSW case I can only guess that doing binary search in two small collections may take much less time that binary search in one large collection - I have only this explanation

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
@@ -180,7 +180,7 @@ impl InvertedIndex {
             };
         }
         // Smallest posting is the largest possible cardinality
-        let smallest_posting = postings.iter().cloned().min().unwrap();
+        let smallest_posting = postings.iter().min().copied().unwrap();
 
         return if postings.len() == 1 {
             CardinalityEstimation {
@@ -544,12 +544,11 @@ impl ImmutableInvertedIndex {
 
 impl From<MutableInvertedIndex> for ImmutableInvertedIndex {
     fn from(mut index: MutableInvertedIndex) -> Self {
-        let mut postings: Vec<Option<CompressedPostingList>> = index
+        let postings: Vec<Option<CompressedPostingList>> = index
             .postings
             .into_iter()
             .map(|x| x.map(CompressedPostingList::new))
             .collect();
-        postings.shrink_to_fit();
         index.vocab.shrink_to_fit();
 
         ImmutableInvertedIndex {
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
@@ -3,8 +3,10 @@ use std::collections::{BTreeSet, HashMap};
 use common::types::PointOffsetType;
 use serde::{Deserialize, Serialize};
 
-use super::posting_list::PostingList;
-use super::postings_iterator::intersect_postings_iterator;
+use super::posting_list::{CompressedPostingList, PostingList};
+use super::postings_iterator::{
+    intersect_compressed_postings_iterator, intersect_postings_iterator,
+};
 use crate::common::operation_error::{OperationError, OperationResult};
 use crate::index::field_index::{CardinalityEstimation, PayloadBlockCondition, PrimaryCondition};
 use crate::types::{FieldCondition, Match, PayloadKeyType};
@@ -132,9 +134,9 @@ impl InvertedIndex {
         query: &ParsedQuery,
         condition: &FieldCondition,
     ) -> CardinalityEstimation {
-        let (points_count, index_postings) = match self {
-            InvertedIndex::Mutable(index) => (index.points_count, &index.postings),
-            InvertedIndex::Immutable(index) => (index.points_count, &index.postings),
+        let points_count = match self {
+            InvertedIndex::Mutable(index) => index.points_count,
+            InvertedIndex::Immutable(index) => index.points_count,
         };
         let postings_opt: Option<Vec<_>> = query
             .tokens
@@ -142,7 +144,20 @@ impl InvertedIndex {
             .map(|&vocab_idx| match vocab_idx {
                 None => None,
                 // unwrap safety: same as in filter()
-                Some(idx) => index_postings.get(idx as usize).unwrap().as_ref(),
+                Some(idx) => match &self {
+                    Self::Mutable(index) => index
+                        .postings
+                        .get(idx as usize)
+                        .unwrap()
+                        .as_ref()
+                        .map(|p| p.len()),
+                    Self::Immutable(index) => index
+                        .postings
+                        .get(idx as usize)
+                        .unwrap()
+                        .as_ref()
+                        .map(|p| p.len()),
+                },
             })
             .collect();
         if postings_opt.is_none() || points_count == 0 {
@@ -165,7 +180,7 @@ impl InvertedIndex {
             };
         }
         // Smallest posting is the largest possible cardinality
-        let smallest_posting = postings.iter().map(|posting| posting.len()).min().unwrap();
+        let smallest_posting = postings.iter().cloned().min().unwrap();
 
         return if postings.len() == 1 {
             CardinalityEstimation {
@@ -177,7 +192,7 @@ impl InvertedIndex {
         } else {
             let expected_frac: f64 = postings
                 .iter()
-                .map(|posting| posting.len() as f64 / points_count as f64)
+                .map(|posting| *posting as f64 / points_count as f64)
                 .product();
             let exp = (expected_frac * points_count as f64) as usize;
             CardinalityEstimation {
@@ -194,32 +209,31 @@ impl InvertedIndex {
         threshold: usize,
         key: PayloadKeyType,
     ) -> Box<dyn Iterator<Item = PayloadBlockCondition> + '_> {
+        let map_filter_condition = move |(token, postings_len): (&str, usize)| {
+            if postings_len >= threshold {
+                Some(PayloadBlockCondition {
+                    condition: FieldCondition::new_match(key.clone(), Match::new_text(token)),
+                    cardinality: postings_len,
+                })
+            } else {
+                None
+            }
+        };
+
         // It might be very hard to predict possible combinations of conditions,
         // so we only build it for individual tokens
-        let (vocab, postings) = match self {
-            InvertedIndex::Mutable(index) => (&index.vocab, &index.postings),
-            InvertedIndex::Immutable(index) => (&index.vocab, &index.postings),
-        };
-        Box::new(
-            vocab
-                .iter()
-                .filter(|(_token, &posting_idx)| postings[posting_idx as usize].is_some())
-                .filter(move |(_token, &posting_idx)| {
-                    // unwrap crash safety: all tokens that passes the first filter should have postings
-                    postings[posting_idx as usize].as_ref().unwrap().len() >= threshold
-                })
-                .map(|(token, &posting_idx)| {
-                    (
-                        token,
-                        // same as the above case
-                        postings[posting_idx as usize].as_ref().unwrap(),
-                    )
-                })
-                .map(move |(token, posting)| PayloadBlockCondition {
-                    condition: FieldCondition::new_match(key.clone(), Match::new_text(token)),
-                    cardinality: posting.len(),
-                }),
-        )
+        match &self {
+            InvertedIndex::Mutable(index) => Box::new(
+                index
+                    .vocab_with_positngs_len_iter()
+                    .filter_map(map_filter_condition),
+            ),
+            InvertedIndex::Immutable(index) => Box::new(
+                index
+                    .vocab_with_positngs_len_iter()
+                    .filter_map(map_filter_condition),
+            ),
+        }
     }
 
     pub fn build_index(
@@ -401,7 +415,7 @@ impl MutableInvertedIndex {
             // Empty request -> no matches
             return Box::new(vec![].into_iter());
         }
-        intersect_postings_iterator(postings, |_| true)
+        intersect_postings_iterator(postings)
     }
 
     fn values_count(&self, point_id: PointOffsetType) -> usize {
@@ -424,11 +438,21 @@ impl MutableInvertedIndex {
     fn get_doc(&self, idx: PointOffsetType) -> Option<&Document> {
         self.point_to_docs.get(idx as usize)?.as_ref()
     }
+
+    fn vocab_with_positngs_len_iter(&self) -> impl Iterator<Item = (&str, usize)> + '_ {
+        self.vocab.iter().filter_map(|(token, &posting_idx)| {
+            if let Some(Some(postings)) = self.postings.get(posting_idx as usize) {
+                Some((token.as_str(), postings.len()))
+            } else {
+                None
+            }
+        })
+    }
 }
 
 #[derive(Default)]
 pub struct ImmutableInvertedIndex {
-    postings: Vec<Option<PostingList>>,
+    postings: Vec<Option<CompressedPostingList>>,
     vocab: HashMap<String, TokenId>,
     point_documents_tokens: Vec<Option<usize>>,
     points_count: usize,
@@ -465,10 +489,10 @@ impl ImmutableInvertedIndex {
             return Box::new(vec![].into_iter());
         }
 
-        // deleted documents are still in the postings
+        // in case of immutable index, deleted documents are still in the postings
         let filter =
             move |idx| matches!(self.point_documents_tokens.get(idx as usize), Some(Some(_)));
-        intersect_postings_iterator(postings, filter)
+        intersect_compressed_postings_iterator(postings, filter)
     }
 
     fn values_is_empty(&self, point_id: PointOffsetType) -> bool {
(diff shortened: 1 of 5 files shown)


agourlay:
Do you have an end to end test to see if the search latency is impacting by the compression overhead?

IvanPleshkov:
> Do you have an end to end test to see if the search latency is impacting by the compression overhead?

Added section with performance measurement results in description

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -506,22 +530,30 @@ impl ImmutableInvertedIndex {
...
-            .iter_mut()
-            .filter_map(|posting| posting.as_mut())
-            .for_each(|posting| {
-                posting.shrink_to_fit();
-            });
-        index.postings.shrink_to_fit();
+            .into_iter()
+            .map(|x| x.map(CompressedPostingList::new))
+            .collect();
+        postings.shrink_to_fit();
```
agourlay, line 552: Why is `shrink_to_fit` necessary?
I'd assume the `collect` method would size the vector appropriately.

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -424,11 +438,21 @@ impl MutableInvertedIndex {
...
+                None
+            }
+        })
+    }
 }
 
 #[derive(Default)]
 pub struct ImmutableInvertedIndex {
-    postings: Vec<Option<PostingList>>,
+    postings: Vec<Option<CompressedPostingList>>,
```
agourlay, line 455: Are we saving the `ImmutableInvertedIndex` to disk with the compressed posting lists at some point?

IvanPleshkov: No, we build it from rocksdb while loading

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -165,7 +180,7 @@ impl InvertedIndex {
             };
         }
         // Smallest posting is the largest possible cardinality
-        let smallest_posting = postings.iter().map(|posting| posting.len()).min().unwrap();
+        let smallest_posting = postings.iter().cloned().min().unwrap();
```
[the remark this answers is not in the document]
timvisee: Would using `copied()` be possible here? That would be a lot cheaper.

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
timvisee: Yes, assuming size information of the iterator is exact it is unnecessary.

lib/segment/src/index/field_index/full_text_index/posting_list.rs
```diff
@@ -37,23 +40,309 @@ impl PostingList {
...
+        self.list.iter().copied()
     }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct CompressedPostingList {
+    len: usize,
+    last_doc_id: PointOffsetType,
+    data: Box<[u8]>,
+    chunks: Box<[CompressedPostingChunk]>,
```
timvisee, line 53: Can't we use a vector in these two cases instead, rather than a boxed array?

lib/segment/src/index/field_index/full_text_index/posting_list.rs
```diff
@@ -37,23 +40,309 @@ impl PostingList {
...
+        assert!(chunk_index < chunks.len());
+        if chunk_index + 1 < chunks.len() {
+            chunks[chunk_index + 1].offset as usize - chunks[chunk_index].offset as usize
+        } else {
+            data.len() - chunks[chunk_index].offset as usize
+        }
+    }
+
+    fn find_chunk(&self, doc_id: &PointOffsetType, start_chunk: Option<usize>) -> Option<usize> {
+        let start_chunk = if let Some(idx) = start_chunk { idx } else { 0 };
```
timvisee, line 164:
```suggestion
        let start_chunk = start_chunk.unwrap_or_default();
```

or

```suggestion
        let start_chunk = start_chunk.unwrap_or(0);
```

lib/segment/src/index/field_index/full_text_index/posting_list.rs
```diff
@@ -37,23 +40,309 @@ impl PostingList {
...
+            Ok(idx) => Some(start_chunk + idx),
+            // chunk idx has larger initial value than doc_id
+            // so we need the previous chunk
+            Err(idx) => {
+                if idx > 0 {
+                    Some(start_chunk + idx - 1)
+                } else {
+                    None
+                }
+            }
```
timvisee, line 176: Not necessary, but we can have conditional branches:

```suggestion
            Err(idx) if idx > 0 => Some(start_chunk + idx - 1),
            Err(_) => None,
```

lib/segment/src/index/field_index/full_text_index/posting_list.rs
```diff
@@ -37,23 +40,309 @@ impl PostingList {
...
+
+    fn decompress_chunk(
+        &self,
+        bitpacker: &BitPackerImpl,
+        chunk_index: usize,
+        decompressed: &mut [PointOffsetType],
+    ) {
+        let chunk = &self.chunks[chunk_index];
+        let chunk_size = Self::get_chunk_size(&self.chunks, &self.data, chunk_index);
+        let chunk_bits = (chunk_size * 8) / BitPackerImpl::BLOCK_LEN;
```
timvisee, line 192: Are we sure we never have any remainder here?

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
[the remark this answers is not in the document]
IvanPleshkov: fixed

lib/segment/src/index/field_index/full_text_index/posting_list.rs
IvanPleshkov: I placed boxed array to show that we cannot change size. But in general I don't have any doubts. Fixed

generall:
VmRSS on dev: 10_116 MB
VmRSS on branch: 9_696 MB

RssAnon could be more relevant

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -132,17 +134,30 @@ impl InvertedIndex {
...
+            InvertedIndex::Immutable(index) => index.points_count,
         };
         let postings_opt: Option<Vec<_>> = query
             .tokens
             .iter()
             .map(|&vocab_idx| match vocab_idx {
                 None => None,
                 // unwrap safety: same as in filter()
-                Some(idx) => index_postings.get(idx as usize).unwrap().as_ref(),
+                Some(idx) => match &self {
```
generall, line 147: `postings_opt` -> `posting_lengths`?

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -165,7 +180,7 @@ impl InvertedIndex {
             };
         }
         // Smallest posting is the largest possible cardinality
-        let smallest_posting = postings.iter().map(|posting| posting.len()).min().unwrap();
+        let smallest_posting = postings.iter().min().copied().unwrap();
```
generall, line 183: Also I think `rust-analyzer` have troubles to infer `postings_opt` type now. 

It shows me `impl Iterator<Item = Option<...>>` before the collect, but where another layer of option goes, assuming that `postings` is `Vec<usize>`

lib/segment/src/index/field_index/full_text_index/posting_list.rs
```diff
@@ -37,23 +40,305 @@ impl PostingList {
...
+    offset: u32,
+}
+
+impl CompressedPostingList {
+    pub fn new(mut posting_list: PostingList) -> Self {
+        if posting_list.list.is_empty() {
+            return Self::default();
+        }
+        let len = posting_list.len();
+        let last_doc_id = *posting_list.list.last().unwrap();
```
generall, line 68: Is it expected to be obtined before sorting?

lib/segment/src/index/field_index/full_text_index/posting_list.rs
```diff
@@ -37,23 +40,305 @@ impl PostingList {
...
 
-    pub fn shrink_to_fit(&mut self) {
-        self.list.shrink_to_fit();
+        let bitpacker = BitPackerImpl::new();
+        posting_list.list.sort_unstable();
+
+        let last = *posting_list.list.last().unwrap();
+        while posting_list.list.len() % BitPackerImpl::BLOCK_LEN != 0 {
+            posting_list.list.push(last);
```
generall, line 75: I am not sure this is a good heuristic.
If we assume that data is distributed according to Zipf, only first few thousands of the vocab tokens will have enough occurances, while the long tail of the smaller posting lists will actually be bloated with this extra "alignment".

I propose to make a Enum of postings, which can deside for it's own which impelemtation to choose.

lib/segment/src/index/field_index/full_text_index/posting_list.rs
```diff
@@ -37,23 +40,305 @@ impl PostingList {
...
+            postings,
+            decompressed_chunk: [0; BitPackerImpl::BLOCK_LEN],
+            decompressed_chunk_idx: None,
+            decompressed_chunk_start_index: 0,
+            #[cfg(test)]
+            last_checked: None,
+        }
+    }
+
+    pub fn contains(&mut self, val: &PointOffsetType) -> bool {
```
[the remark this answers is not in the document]
generall: Also docstring might help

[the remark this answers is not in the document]
agourlay: I agree, the `&mut` tripped me as well on a `contains` method :+1:

generall (changes requested):
Needs refactoring

lib/segment/src/index/field_index/full_text_index/posting_list.rs
```diff
@@ -243,7 +241,10 @@ impl<'a> CompressedPostingVisitor<'a> {
         }
     }
 
-    pub fn contains(&mut self, val: &PointOffsetType) -> bool {
+    // Check if the next value is in the compressed posting list.
+    // This function reuses the decompressed chunk to avoid unnecessary decompression.
+    // It is useful when the visitor is used to check the values in the increasing order.
+    pub fn contains_next(&mut self, val: &PointOffsetType) -> bool {
```
timvisee, line 247: Could we extend the comment to explain why `&mut` is needed?

I assume it's there to be careful about something specifically.

lib/segment/src/index/field_index/full_text_index/posting_list.rs
IvanPleshkov: We are sure about it because `chunk_size` is defined as `chunk_bits * BitPackerImpl::BLOCK_LEN / 8`, `BitPackerImpl::BLOCK_LEN` is `128`

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
IvanPleshkov: fixed

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
IvanPleshkov: Added `Vec<usize>` type

lib/segment/src/index/field_index/full_text_index/posting_list.rs
IvanPleshkov: Removed, thanks!

lib/segment/src/index/field_index/full_text_index/posting_list.rs
IvanPleshkov: Fixed. I implemented it simpler. Remainder data is uncompressed. For small postings list we contain add data as remainder

lib/segment/src/index/field_index/full_text_index/posting_list.rs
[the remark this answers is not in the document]
IvanPleshkov: This method is defined is separate help structure which reuse decompressed data to avoid unnecessary decompression. Help structure was commented. Added comments also to this function. And renamed into `contains_next `

lib/segment/src/index/field_index/full_text_index/posting_list.rs
IvanPleshkov: Commened the same question below. This method is defined in separate help structure which reuse decompressed data to avoid unnecessary decompression. Help structure was commented. Added comments also to this function. And renamed into contains_next

timvisee (approved):

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -132,20 +134,33 @@ impl InvertedIndex {
...
+                        .map(|p| p.len()),
+                    Self::Immutable(index) => index
+                        .postings
+                        .get(idx as usize)
+                        .unwrap()
+                        .as_ref()
+                        .map(|p| p.len()),
+                },
             })
             .collect();
```
generall, line 162: I probably need an extra explanation why we have double Option around `PostingList`?

And how Option of `p.len()` is collected into just `Vec<usize>`? Do we skip Nones? If we have no posting lists for token, shouldn't we consider it as length=0?

lib/segment/src/index/field_index/full_text_index/posting_list.rs
generall: `contains_next` still doesn't highlight the need for mutability.

lib/segment/src/index/field_index/full_text_index/posting_list.rs
```diff
@@ -37,23 +40,323 @@ impl PostingList {
...
+        self.postings
+            .decompress_chunk(&self.bitpacker, chunk_index, &mut self.decompressed_chunk);
+        self.decompressed_chunk_idx = Some(chunk_index);
+        self.decompressed_chunk_start_index = 0;
+
+        // check if the value is in the decompressed chunk
+        self.find_in_decompressed_chunk(val)
+    }
+
+    fn find_in_decompressed_chunk(&mut self, val: &PointOffsetType) -> bool {
```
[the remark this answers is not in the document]
IvanPleshkov: It's not a method of postings list. It's a method for help structure for intersections. It's mutable because we do binary search inside not from 0. We do it from index, which we found from previous `contains_next` check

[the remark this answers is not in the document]
IvanPleshkov: But I agree that it's incorrect naming. I have to mark in function name that I change internal state of help stucture

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
IvanPleshkov: Agree that this `Option`s are unclear. I think that we can simplify the whole inverted index and remove a lot of obsolete `Option`s. But right now I'm not sure that there are no any place where `Option` is important. I Propose to fix it as separate PR, created issue for that:
https://github.com/qdrant/qdrant/issues/3716

PR #3565, opened 2024-02-08
Configure default RocksDB column family

agourlay:
This PR configures the `default` column family like all other column families.

After https://github.com/qdrant/qdrant/pull/3557 it was [reported](https://github.com/qdrant/qdrant/pull/3557#discussion_r1482184228) that the RocksDB logs still mentioned an incorrect `write_buffer_size` once at startup.

I was able to see the same locally and after some digging I found out that RocksDB always creates a default column family just in case there are key-value pairs without column.

See https://github.com/facebook/rocksdb/wiki/Column-Families

By applying the following patch I was able to configure properly the default column family.

see `OPTIONS-00000xx` files.

```
[CFOptions "default"]
  ...
  write_buffer_size=10485760
```

I have not found how to disable this column family completely.


timvisee:
Would it be possible to entirely disable the default column instead/

timvisee (approved):

generall (approved):

IvanPleshkov (approved):

PR #3568, opened 2024-02-08
fix: flaky segment builder cancellation test

felipetrz:
### All Submissions:

### Test changes to minimize flakiness

/claim #2723

The main reason for this test's flakiness is the performance fluctuation of the cancellation teardown running in debug mode on the CI environment.

The acceptance margin for the teardown delay after cancellation being absolute makes this test too dependant on specific performance characteristics of the testing environment.

My proposed solution to the flakiness is to define the `acceptable_stopped_delay` proportionally to the `baseline_time`, decoupling the parameter from specific performance requirements. This should make the test consistent regardless of the performance characteristics of the environment it's running.

Another issue I've identified is that the indexing thread-pool sets threads to a low priority. This could make the entire test starve in a loaded shared-time environment (which is easy to verify by running the test while a CPU stress test is running in background). I've made the low thread priority only be used in release mode, which should eliminate this factor as well.

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -7,6 +7,7 @@ use std::thread;
 
 use atomic_refcell::AtomicRefCell;
 #[cfg(target_os = "linux")]
+#[cfg(profile = "release")]
 use common::cpu::linux_low_thread_priority;
 use common::cpu::CpuPermit;
 use common::types::{PointOffsetType, ScoredPointOffset};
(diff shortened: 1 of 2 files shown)


coszio:
Hi @felipetrz! Thank you for submitting this PR.

While I agree that the main reason is the performance fluctuation of Github Actions is making this test flake so often, we had the same issue even from before introducing thread priorities.

Personally I feel like the overall design of our test might not be the best, not just the current parameters.

felipetrz:
Yes @coszio, the idea of my PR is to minimize the test's sensitivity to performance and load fluctuations, while keeping the test intact (as per [this comment](https://github.com/qdrant/qdrant/issues/2723#issuecomment-1916402612)).

The intention of the test is to verify that the operation is indeed cancelled, but what ends up actually being tested is the performance of the atomic unit of work between cancellation checks, except tests run in an unoptimized debug build, inflating this delay in an unpredictable manner that doesn't actually reflect how the cancellation behaves in a production build. Turning the test into a benchmark would make more sense, but then the introduction of a cancellation-blocking path wouldn't fail validation.

felipetrz:
It seems there's no intention of merging this, so I'll close the PR.

PR #3569, opened 2024-02-08
Refactor method `wait_for_consensus`

jbajic:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/consensus_manager.rs
+++ b/lib/storage/src/content_manager/consensus_manager.rs
@@ -4,7 +4,7 @@ use std::fmt::Display;
 use std::future::Future;
 use std::ops::Deref;
 use std::sync::Arc;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 use anyhow::{anyhow, Context};
 use chrono::Utc;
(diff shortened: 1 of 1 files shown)


lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -600,10 +600,27 @@ impl<C: CollectionContainer> ConsensusManager<C> {
...
+        commit: u64,
+        term: u64,
+        consensus_tick: Duration,
+    ) -> Result<(), ()> {
+        let mut interval = tokio::time::interval(consensus_tick);
 
-        // TODO: naive approach with spinlock for waiting on commit/term, find better way
-        while start.elapsed() < timeout {
+        loop {
+            interval.tick().await;
```
timvisee, line 623: We don't want to wait on the first run.

timvisee (commented):
Hi! Could you elaborate on what these changes intend to do?

lib/storage/src/content_manager/consensus_manager.rs
jbajic: The first tick is triggered immediately as per docs (https://docs.rs/tokio/latest/tokio/time/struct.Interval.html).

jbajic:
> Hi! Could you elaborate on what these changes intend to do?
> 
> We prefer not to touch functions like this unless strictly necessary.

I wanted to resolve TODO comment. My idea was to use a more idiomatic approach. Since tokio supports the exact API functions for this use case.

timvisee:
> I wanted to resolve TODO comment. My idea was to use a more idiomatic approach. Since tokio supports the exact API functions for this use case.

Right!

The problem is, that it does not resolve the TODO. There is still a lock spinning each consensus tick.

Instead this TODO aims to implement it in a different way, probably with some event listening. But we don't have the infrastructure for this in place. And the question is whether we ever will.

jbajic:
> > I wanted to resolve TODO comment. My idea was to use a more idiomatic approach. Since tokio supports the exact API functions for this use case.
> 
> Right!
> 
> The problem is, that it does not resolve the TODO. There is still a lock spinning each consensus tick.
> 
> Instead this TODO aims to implement it in a different way, probably with some event listening. But we don't have the infrastructure for this in place. And the question is whether we ever will.

Aha, that makes sense. I misunderstood the TODO comment, I will close the PR

timvisee:
No problem! Thank you for taking the effort. :smile:

PR #3572, opened 2024-02-09
Add unit tests for the `ClockSet`/`ClockMap` workflow

ffuugoo:
This PR adds some unit tests for the `ClockMap` and a "`ClockSet`/`ClockMap` workflow test", that checks that they update each other correctly in a common scenario.

Tracked in: #3477

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -127,15 +127,12 @@ pub struct Key {
 }
 
 impl Key {
-    pub fn new(peer_id: PeerId, clock_id: u32) -> Self {
+    fn new(peer_id: PeerId, clock_id: u32) -> Self {
         Self { peer_id, clock_id }
     }
 
-    pub fn from_tag(clock_tag: ClockTag) -> Self {
-        Self {
-            peer_id: clock_tag.peer_id,
-            clock_id: clock_tag.clock_id,
-        }
+    fn from_tag(clock_tag: ClockTag) -> Self {
+        Self::new(clock_tag.peer_id, clock_tag.clock_id)
     }
 }
 
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -386,6 +386,7 @@ impl From<Error> for CollectionError {
 
 #[cfg(test)]
 mod test {
+    use proptest::prelude::*;
     use test_case::test_case;
 
     use super::*;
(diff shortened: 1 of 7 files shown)


lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -362,4 +374,217 @@ mod tests {
...
+        //
+        // Start from tick `1`, cause tick `0` is a special case that is always accepted
+        for older_tick in 1..10 {
+            helper.advance(tag(older_tick)).assert(false, 10);
+        }
+
+        // Reject current tick
+        for current_tick in [10, 10, 10, 10, 10] {
+            helper.advance(tag(current_tick)).assert(false, 10);
+        }
```
timvisee, line 436: If rejection means we won't write the operation to WAL, I'm not sure if we should do this.

What if we do want to rewrite an operation, since they're idempotent, it shouldn't matter.

Or are all cases where we might do this covered with force=true?

timvisee (approved):

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -1,4 +1,4 @@
-mod clock_set;
+pub mod clock_set;
```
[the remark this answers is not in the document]
ffuugoo: I just don't like `pub(crate)`/`pub(super)`. I think they are mostly useless. What `pub(super)` will give us, that `pub` won't?

lib/collection/src/shards/local_shard/clock_map.rs
ffuugoo: > If rejection means we won't write the operation to WAL, I'm not sure if we should do this.

Do you mean older ticks or current ticks?

I think we strictly **cannot** write *older* ticks, and we *maybe* can write *current* ticks, but I'm not sure if we should.

If you just allow writing arbitrary older ticks into WAL, you fuck up the diff resolution.

> What if we do want to rewrite an operation, since they're idempotent, it shouldn't matter.

I don't think that's something we ever do? (Or maybe I don't understand what you mean here.)

> Or are all cases where we might do this covered with force=true?

Yeah, I think they are.

We generally want to keep WAL strictly sequential (with a small exception for `0` ticks), so we explicitly only accept *newer* ticks.

Slightly artificial example: what if ops 4 and 5 arrive from forward proxy, and then we (somehow) receive very late op 4 from original shard (and op 5 is lost forever). And now we have 4, 5, 4 in WAL and diff resolution is broken.

And we have `force = true` for rare cases when we want to "force" some operations into WAL, but in such cases we *have to* make sure these "forced" operations won't break WAL properties that we rely on.

ffuugoo: > We generally want to keep WAL strictly sequential (with a small exception for 0 ticks)...

BTW, I think our diff resolution can't handle this case yet, cause when shard is restarted it will send op(s) with tick `0`, and these ops will be written into WALs with *current* tick, and so we might have multiple different ops saved in WAL with the same tick, so we'll have to pick all of them.

lib/collection/src/shards/replica_set/mod.rs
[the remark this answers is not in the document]
timvisee: > What pub(super) will give us, that pub won't?

Extra protection over who is allowed to use the module. It exists for a reason.

lib/collection/src/shards/local_shard/clock_map.rs
timvisee: Alright. Let's merge and see what happens! :+1:

lib/collection/src/shards/replica_set/mod.rs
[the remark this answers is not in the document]
ffuugoo: There is also `Iterator::fold`, doesn't mean you have to use it. 😂

All types and methods are properly scoped in this module. There's literally 1) nothing to protect here and 2) no extra protection `pub(super)` can give us over `pub`.

PR #3574, opened 2024-02-09
Add integer range condition and integer `start_from` to interface

coszio:
To finish taking advantage of the newly added range interface. This adds:
- `Range<IntPayloadType>` to `RangeInterface`
- `StartFrom::Int(IntPayloadType)` variant for order-by `scroll` requests.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/numeric_index/mod.rs
+++ b/lib/segment/src/index/field_index/numeric_index/mod.rs
@@ -417,7 +417,7 @@ impl<T: Encodable + Numericable + Default> PayloadFieldIndex for NumericIndex<T>
                         _ => None,
                     },
                 };
-                let cardinality = self.range_cardinality(&RangeInterface::Float(range.clone()));
+                let cardinality = self.range_cardinality(&RangeInterface::Float(range));
                 let condition = PayloadBlockCondition {
                     condition: FieldCondition::new_range(key.clone(), range),
                     cardinality: cardinality.exp,
@@ -542,7 +542,7 @@ where
         &self,
         range_interface: &RangeInterface,
     ) -> Box<dyn DoubleEndedIterator<Item = (T, PointOffsetType)> + '_> {
-        let range = Range::from(range_interface.clone());
+        let range = Range::from(*range_interface);
         let (start_bound, end_bound) = range.as_index_key_bounds();
 
         // map.range
--- a/lib/segment/src/index/field_index/numeric_index/mod.rs
+++ b/lib/segment/src/index/field_index/numeric_index/mod.rs
@@ -241,6 +241,7 @@ impl<T: Encodable + Numericable + Default> NumericIndex<T> {
             RangeInterface::DateTime(datetime_range) => {
                 datetime_range.map(|dt| T::from_i64(dt.timestamp()))
             }
+            RangeInterface::Int(int_range) => int_range.map(T::from_i64),
         };
 
         let lbound = if let Some(lte) = range.lte {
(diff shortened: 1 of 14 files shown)


lib/segment/src/types.rs
```diff
@@ -59,6 +59,34 @@ impl DateTimeWrapper {
     }
 }
 
+impl FromStr for DateTimePayloadType {
```
coszio, line 62: This part was just moved closer to the definition of `DateTimeWrapper`

lib/segment/src/payload_storage/condition_checker.rs
```diff
@@ -140,18 +137,34 @@ impl ValueChecker for Match {
...
+
+impl ValueChecker for Range<IntPayloadType> {
+    fn check_match(&self, payload: &Value) -> bool {
+        payload
+            .as_number()
+            .and_then(|x| x.as_i64().or_else(|| x.as_f64().map(|x| x as i64)))
+            .map_or(false, |x| self.check_range(x))
+    }
+}
+
```
coszio, line 167: For unindexed fields, we now try to convert into the type of the range to do the checks

lib/segment/src/index/query_optimization/condition_converter.rs
```diff
@@ -245,6 +246,28 @@ pub fn get_float_range_checkers(
...
+            Some(Box::new(move |point_id: PointOffsetType| {
+                num_index
+                    .get_values(point_id)
+                    .is_some_and(|values| values.iter().copied().any(|f| range.check_range(f)))
+            }))
+        }
+        _ => None,
+    }
+}
+
```
coszio, line 270: For indexed fields, we follow same approach as before, converting the range into the type of the index

coszio:
I am facing issues making appropriate conversions in the python client (in which pydantic is matching floats to ints, and viceversa). Maybe this is a good reason to **not** include these changes. 

In the meantime, I'll draft this PR to avoid merging just yet.

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -526,21 +522,27 @@ impl ValueIndexer<FloatPayloadType> for NumericIndex<FloatPayloadType> {
...
-        range: &RangeInterface,
+        range_interface: &RangeInterface,
     ) -> Box<dyn DoubleEndedIterator<Item = (T, PointOffsetType)> + '_> {
-        let range = match range {
-            RangeInterface::Float(float_range) => float_range.map(T::from_f64),
-            RangeInterface::DateTime(datetime_range) => {
-                datetime_range.map(|dt| T::from_i64(dt.timestamp()))
-            }
-        };
+        let range = Range::from(range_interface.clone());
```
xzfc, line 545: Nit: maybe we could avoid cloning if there is `impl<T: Numericable> From<&RangeInterface> for Range<T>` (note the `&`).

lib/segment/src/payload_storage/condition_checker.rs
```diff
@@ -140,18 +137,34 @@ impl ValueChecker for Match {
...
+            .and_then(|x| x.as_f64().or_else(|| x.as_i64().map(|x| x as f64)))
+            .map_or(false, |x| self.check_range(x))
+    }
+}
+
+impl ValueChecker for Range<IntPayloadType> {
+    fn check_match(&self, payload: &Value) -> bool {
+        payload
+            .as_number()
+            .and_then(|x| x.as_i64().or_else(|| x.as_f64().map(|x| x as i64)))
```
xzfc, line 163: This rounds floats down, so, e.g., `{"payload": 0.3}` would be rounded to `0i64` and then matched by `{lte: Some(0i64)}` which is not what we want.
I think a similar rounding issue could also be in the case above (`x as f64`).
I propose to do comparison using `num-cmp`, like discussed in https://github.com/qdrant/qdrant/pull/3533#pullrequestreview-1864712419. I.e. explicitly handle cases `i64<i64`, `i64<f64`, `f64<i64`, and so on.

xzfc (changes requested):

lib/segment/src/payload_storage/condition_checker.rs
coszio: You're totally right, I'll update accordingly

coszio: @xzfc I have added a new check_range implementation for the cases where we need to check different types.

For the case of `Range<f64>`, though, I've restored its original implementation, as all integers were getting converted to floats via `as f64` inside the `serde_json` crate already.

lib/segment/src/index/field_index/numeric_index/mod.rs
coszio: Yeah, I also thought about that, but I think this is a fairly inexpensive operation.

To make this more explicit, I have just pushed a change to make `Range` and `RangeInterface` `Copy`

xzfc (dismissed):

generall:
Please feel free to make another iteration, once we will have a better idea of how integrate it into the client

PR #3578, opened 2024-02-09
refactor: add DateTime wrapper

coszio:
Creates a newtype pattern for `DateTime<Utc>`, so that we can encapsulate the special deserialization and conversion to timestamp in a single entity.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -65,7 +65,7 @@ impl<'de> Deserialize<'de> for DateTimeWrapper {
         D: Deserializer<'de>,
     {
         let str_datetime = <&str>::deserialize(deserializer)?;
-        let parse_result = try_parse_datetime(str_datetime);
+        let parse_result = DateTimePayloadType::from_str(str_datetime).ok();
         match parse_result {
             Some(datetime) => Ok(datetime),
             None => Err(serde::de::Error::custom(format!(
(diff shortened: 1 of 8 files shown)


lib/segment/src/types.rs
```diff
@@ -45,7 +45,41 @@ pub type FloatPayloadType = f64;
 /// Type of integer point payload
 pub type IntPayloadType = i64;
 /// Type of datetime point payload
-pub type DateTimePayloadType = chrono::DateTime<chrono::Utc>;
+pub type DateTimePayloadType = DateTimeWrapper;
+
+/// Wraps `DateTime<Utc>` to allow more flexible deserialization
+#[derive(Clone, Copy, Serialize, JsonSchema, Debug, PartialEq, PartialOrd)]
+#[serde(transparent)]
+pub struct DateTimeWrapper(pub chrono::DateTime<chrono::Utc>);
```
xzfc, line 53: Maybe name it just as `DateTime`? And use it everywhere in the codebase instead of `chrono::DateTime`, `chrono::NaiveDateTime`, `std::time::SystemTime`, etc. This would improve the consistency.

lib/segment/src/types.rs
```diff
@@ -1380,7 +1413,7 @@ pub struct Range<T> {
 pub fn try_parse_datetime(s: &str) -> Option<DateTimePayloadType> {
```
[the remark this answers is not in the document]
coszio: I forgot about `FromStr`, I also think it is a good idea

lib/segment/src/types.rs
coszio: Not sure if we should replace `std::time::SystemTime`, but I thought about the name, and decided `DateTimeWrapper` would be a good name to make it more obvious when seeing this in inlay hints, or elsewhere outside the context of this PR. Do you find it weird?

generall: Since this name is not exposed to any API, I am ok to merge as is and change later if it causes any confusion

generall (approved):

lib/segment/src/types.rs
```diff
@@ -1376,89 +1409,34 @@ pub struct Range<T> {
...
+        // Attempt to parse the input string in the specified formats:
+        // - YYYY-MM-DD'T'HH:MM:SS (without timezone or Z)
+        // - YYYY-MM-DD HH:MM:SS
+        // - YYYY-MM-DD HH:MM
+        // - YYYY-MM-DD
+        // See: <https://github.com/qdrant/qdrant/issues/3529>
+        let datetime = chrono::NaiveDateTime::parse_from_str(s, "%Y-%m-%dT%H:%M:%S")
+            .or_else(|_| chrono::NaiveDateTime::parse_from_str(s, "%Y-%m-%d %H:%M:%S"))
+            .or_else(|_| chrono::NaiveDateTime::parse_from_str(s, "%Y-%m-%d %H:%M"))
+            .or_else(|_| chrono::NaiveDate::parse_from_str(s, "%Y-%m-%d").map(Into::into))?;
```
timvisee, line 1432: I like this one much better. Nice!

timvisee (approved):

Issue #3450, opened 2024-01-23
Deprecate `Initializing` state of the shards


generall:
**Is your feature request related to a problem? Please describe.**
`Initializing` state was used to ensure that all replicas created the collection before reporting to the user that the collection was created. But it seems that this solution is not optimal and the better one would be to explicitly await for consensus operation on other nodes ( this PR: https://github.com/qdrant/qdrant/pull/3318 )

**Describe the solution you'd like**

Step-by-step deprecate usage of `Initializing` state:

- Make `Initializing` behave as `Active` in all situations
- *release minor version*
- Create shards with `Active` status by default
- *release minor version*
- Remove all operations related to switch from Initializing state to Active
- *release minor version*
- Remove `Initializing` completely 

**Describe alternatives you've considered**

Do all in one release, but in this case there is a risk creating a collection in between updates which will never transition from `Initializing` into `Active`

**Additional context**

Continuation of https://github.com/qdrant/qdrant/pull/3318

AlexandruCihodaru:
Hello.
I have created https://github.com/qdrant/qdrant/pull/3584 for the first point in this list.
I have identified three places where the behavior is different depending on the ReplicaState:
* https://github.com/qdrant/qdrant/blob/4c5596579e0b5e8f1312126b22264e0436253765/lib/collection/src/shards/replica_set/mod.rs#L884-L894
In this method we should also return true for ReplicaState::Initializing to skip them when looking for unhealty_shards.
* https://github.com/qdrant/qdrant/blob/4c5596579e0b5e8f1312126b22264e0436253765/lib/collection/src/shards/replica_set/mod.rs#L578-L581
In here we should update the local state to Active and also notify a peer failure, as it's currently done for ReplicaState::Active and ReplicaState::Listener
* https://github.com/qdrant/qdrant/blob/4c5596579e0b5e8f1312126b22264e0436253765/lib/collection/src/shards/replica_set/update.rs#L308-L314
In here we should remove the second part of the if because there is no point in waiting for an "Initializing" shard anymore.

PR #3584, opened 2024-02-11
Change initializing behaviour

AlexandruCihodaru:
### All Submissions:

### Changes to Core Features:

* [] Have you written new tests for your core changes, as applicable?

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -913,12 +913,9 @@ impl ReplicaState {
     pub fn is_active_or_listener(self) -> bool {
         // Use explicit match, to catch future changes to `ReplicaState`
         match self {
-            ReplicaState::Active | ReplicaState::Listener => true,
+            ReplicaState::Active | ReplicaState::Listener | ReplicaState::Initializing => true,
 
-            ReplicaState::Dead
-            | ReplicaState::Initializing
-            | ReplicaState::Partial
-            | ReplicaState::PartialSnapshot => false,
+            ReplicaState::Dead | ReplicaState::Partial | ReplicaState::PartialSnapshot => false,
         }
     }
 
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -358,7 +358,7 @@ impl ShardReplicaSet {
                 continue;
             }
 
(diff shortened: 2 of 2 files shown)


generall:
Hey @AlexandruCihodaru, we unfortunately can't remove `Initializing` right away, it would create problems during cluster version upgrade, when some nodes have latest versions, while others don't.

Here is a step-by-step plan of how we should proceed: https://github.com/qdrant/qdrant/issues/3450

AlexandruCihodaru:
> Hey @AlexandruCihodaru, we unfortunately can't remove `Initializing` right away, it would create problems during cluster version upgrade, when some nodes have latest versions, while others don't.
> 
> Here is a step-by-step plan of how we should proceed: #3450

Hello @generall  I have seen that issue and was planning to follow that list. In this first PR I attempted to address the first bullet
"Make Initializing behave as Active in all situations"

generall:
>  In this first PR I attempted to address the first bullet "Make Initializing behave as Active in all situations"

Oh, ok. Let me have another look then

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -650,8 +650,10 @@ impl ShardReplicaSet {
                             .await?;
                     }
                     ReplicaState::Initializing => {
-                        self.set_local(local_shard, Some(ReplicaState::Initializing))
+                        // Same as `Active`, we report a failure to consensus
+                        self.set_local(local_shard, Some(ReplicaState::Active))
```
generall, line 654: I don't think we can set Active local state here, even if the Active and Initializing will act the same.

General rule here is that we only change state of shards if there is a consensus operation saying to do so. Otherwise there is a risk of having de-sync in shard states between nodes.


Consider the following scenario:

- Node A of a newest version receives Snapshot with `ReplicaState::Initializing` and changes it to Active.
- Node B with an older version still thinks that Node A have shard in Initializing state.
- Since Node A have local state of Active, it will never trigger consensus to change the state
- Node B will think about node A as stuck in Initializing state indefinitely,  even after update to a newer version.


Luckily it should not be a big deal to have `Initializing` state, as we have a procedure to convert Initializing into active if it happens to be a local one

AlexandruCihodaru: I see, thank you for your explication. I will revert this commit because there won't be the case to notify peer failure.

Issue #3585, opened 2024-02-12
Service internal error: Json error: key must be a string


dacox:
On the `dev` branch, we are getting a service error in the optimizer status

```
{
  "result": {
    "status": "red",
    "optimizer_status": {
      "error": "Service internal error: Json error: key must be a string"
    },
```

## Current Behavior

No error about json keys

## Steps to Reproduce

1. use `qdrant:dev`
2. create a collection
3. index a bunch of points
4. `GET` the collection and look at optimizer status

## Context (Environment)

This is for the `dev` image, not specific to a known release

timvisee:
Very interesting. We've not seen this before.

1. Does the same happen on our latest release?
2. Could you provide an example or snippet on what you're inserting exactly?

dacox:
> Very interesting. We've not seen this before.
> 
> 1. Does the same happen on our latest release?
> 2. Could you provide an example or snippet on what you're inserting exactly?

@timvisee I have not seen this on the latest release, just the `dev` image.

I've been using the same script I've been using for benchmarking memory usage on local builds

```
import os
import random
import uuid

import requests
import numpy as np

from qdrant_client import QdrantClient
from qdrant_client.http.models import PointStruct


HOST = os.getenv('HOST', 'qdrant')
PORT = os.getenv('PORT', 6333)
COLLECTION_NAME = os.getenv('COLLECTION_NAME', 'collection_123')


def get_or_create_collection():
    try:
        requests.put(f'http://{HOST}:{PORT}/collections/{COLLECTION_NAME}', json={
            "vectors": {
              "size": 1536,
              "distance": "Cosine",
              "on_disk": True
            },
            "optimizers_config": {
              "default_segment_number": 32,
              "indexing_threshold": 0,
              "max_optimization_threads": 1
            },
            "hnsw_config": {
              "on_disk": True
            },
            "quantization_config": {
              "binary": {
                "always_ram": True
              }
            },
            "on_disk_payload": False
        })
    except Exception as e:
        print(e)


def gen(batch_size):
    vectors = np.random.rand(batch_size, 1536)
    sps = [random.choices(list(range(12)), k=1)[0] for _ in range(batch_size)]
    vps = [random.choices([True, False], weights=[0.2, 0.8], k=1)[0] for _ in range(batch_size)]
    uuids = [str(uuid.uuid4()) for _ in range(batch_size)]
    points = []
    for vector, s, v, uid in zip(vectors, sps, vps, uuids):
[...]

PR #3590, opened 2024-02-12
Fix broken serialization of ClockMap

timvisee:
Fixes: <https://github.com/qdrant/qdrant/issues/3585>

The current implementation of `ClockMap` fails to serialize when there's any data in it. It holds a `HashMap<Key, Clock>`, but `serde_json` cannot serialize `Key` as it must be a string.

More specifically, when we try to serialize I'm getting this error:

```
Error("key must be a string", line: 0, column: 0)
```

This PR fixes that by utilizing [`serde_with::serde_as`](https://docs.rs/serde_with/3.6.1/serde_with/attr.serde_as.html), mapping our hash map into a `Vec<(Key, Clock)>` on (de)serialization.

We now get:

```json
{"clocks":[[{"peer_id":123,"clock_id":1},152],[{"peer_id":123,"clock_id":0},246]]}
```

If we don't like the tuple representation we may also decide to implement separate serializable types while using `#[serde(into = "SomeType", from = "SomeType")]` ([docs](https://serde.rs/container-attrs.html)). I didn't go that route as it would require more code to make it happen.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -6,6 +6,7 @@ use std::path::Path;
 use api::grpc::qdrant::RecoveryPointClockTag;
 use io::file_operations;
 use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
 
 use crate::operations::types::CollectionError;
 use crate::operations::ClockTag;
@@ -13,7 +14,9 @@ use crate::shards::shard::PeerId;
 
 #[derive(Debug, Default, Deserialize, Serialize)]
 #[serde(transparent)]
+#[serde_as]
 pub struct ClockMap {
+    #[serde_as(as = "Vec<(_, _)>")]
     clocks: HashMap<Key, Clock>,
 }
 
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -141,7 +141,6 @@ impl Key {
 }
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
-#[serde(transparent)]
 struct Clock {
     current_tick: u64,
 }
@@ -292,7 +291,6 @@ impl From<ClockMapHelper> for ClockMap {
 struct KeyClockHelper {
     #[serde(flatten)]
     key: Key,
-
     #[serde(flatten)]
     clock: Clock,
 }
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -12,7 +12,7 @@ use crate::operations::types::CollectionError;
 use crate::operations::ClockTag;
 use crate::shards::shard::PeerId;
 
-#[derive(Debug, Default, Deserialize, Serialize)]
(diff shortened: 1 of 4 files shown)


ffuugoo:
`Vec<Key, Clock>` is okay-ish, but kinda ugly. How annoying would it be to make it `[{"peer_id": 123, "clock_id": 1, "clock_tick": 10}]` or something? 🤔

timvisee:
> `Vec<Key, Clock>` is okay-ish, but kinda ugly. How annoying would it be to make it `[{"peer_id": 123, "clock_id": 1, "clock_tick": 10}]` or something? 🤔

It's not super easy to do.

We'd need an alternative type for `ClockMap` to (de)serialize from. So we'd need to implement a `SerdeClockMap` and `SerdeClockMapEntry` or something like this.

ffuugoo:
`serde_as` is cool, I love it. I just don't like current clock map representation.

> So we'd need to implement a SerdeClockMap and SerdeClockMapEntry or something along with conversion functions.

As we are using `serde_as`, I think we can do this:
- `struct Helper { peer_id: PeerId, clock_id: u32, clock_tick: u64 }`
- and then do `impl SerializeAs<Helper> for (Key, Clock)`
- and `impl DeserializeAs<Helper> for (Key, Clock)`

agourlay:
Is it possible to write tests for this?

timvisee:
> Is it possible to write tests for this?

Yes. Fill the struct with some dummy data and try to store/load.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -254,6 +254,48 @@ impl From<api::grpc::qdrant::RecoveryPoint> for RecoveryPoint {
...
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
+struct KeyClockHelper {
+    #[serde(flatten)]
+    key: Key,
+
+    #[serde(flatten)]
```
timvisee, line 283: I think this fails because `Clock` has transparent.

I guess the best would be to remove transparent, and keep flatten here.

ffuugoo: Yeah, you're probably right. Nice catch.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -294,3 +334,38 @@ impl From<Error> for CollectionError {
...
+        clock_map.advance_clock(ClockTag::new(2, 2, 12345));
+
+        let temp_dir = tempfile::tempdir().unwrap();
+        let path = temp_dir.path().join("clock_map.json");
+
+        clock_map.store(&path).unwrap();
+        let clock_map2 = ClockMap::load(&path).unwrap();
+
+        assert_eq!(clock_map, clock_map2);
+    }
```
ffuugoo, line 370: Do you mind if I switch to `serde_json::to_value`/`serde_json::from_value` to not dump stuff onto disk in these tests?

timvisee: Sure, good idea. :+1:

timvisee:
✅

ffuugoo (approved):
Obviously, LGTM from me as well. :D

PR #3593, opened 2024-02-12
Recovery point should list first unseen ticks rather than last seen

timvisee:

Before this PR, the recovery point lists the last seen clock ticks. After this PR, the recovery point lists the clock ticks right after the last seen ones.

In other words, the recovery point defines a lower bound which is now inclusive, rather than exclusive. It describes what clock ticks we want to have on recovery, not the clock ticks we already have.

We must do this so that we can extend the recovery point with `0` ticks for clocks the recovering node might not have seen yet. In this case, recovery should include operation 0 for that clock (making it inclusive). Before this change, the 0 operation would be skipped.

The WAL resolving logic always assumed the recovery point had an inclusive lower bound. Because of it, it's logic has not been changed. It now only includes new operations when resolving a diff. Before it also included the last operations the rescovering node already had, which are effectively obsolete.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -101,13 +101,21 @@ impl ClockMap {
         }
     }
 
-    /// Create a recovery point from the current clock map state
+    /// Create a recovery point based on the current clock map state, so that we can recover any
+    /// new operations with new clock values
+    ///
+    /// The recovery point will contain every clock that is in this clock map, but with a tick of
+    /// one higher. That is because we already have an operation for the current clock tick, but
+    /// would like to receive the operation with the next tick on recovery.
+    ///
+    /// In other words, the recovery point will contain the first clock tick values the clock map
+    /// has not seen yet.
     pub fn to_recovery_point(&self) -> RecoveryPoint {
         RecoveryPoint {
             clocks: self
                 .clocks
                 .iter()
-                .map(|(key, clock)| (*key, clock.current_tick))
+                .map(|(key, clock)| (*key, clock.current_tick + 1))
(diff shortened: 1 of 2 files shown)


generall:
I am not completely agree with this change. Could you please provide an example where previous logic fails?

timvisee:
> I am not completely agree with this change. Could you please provide an example where previous logic fails?

I can easily explain why we need something like this.

Our recovery point must be able to describe that a node needs clock 0. In the merged implementation you'd have to use clock -1 because it is exclusive, and we cannot do that.

An alternative would be to use an option. If a recovery point has a clock with a None tick, it needs clock 0 and upwards. But, I don't like this at all and find the suggestion in this PR much better.

Does that explanation help?

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -101,13 +101,21 @@ impl ClockMap {
...
+    /// would like to receive the operation with the next tick on recovery.
+    ///
+    /// In other words, the recovery point will contain the first clock tick values the clock map
+    /// has not seen yet.
     pub fn to_recovery_point(&self) -> RecoveryPoint {
         RecoveryPoint {
             clocks: self
                 .clocks
                 .iter()
-                .map(|(key, clock)| (*key, clock.current_tick))
```
agourlay, line 110: Can we test this new behavior? 
Off-by one changes are pretty subtle.

timvisee: Not at this time. But we will test it as part of our bigger testing efforts once all parts are in place.

generall (approved):

PR #3610, opened 2024-02-13
WAL delta resolution and recovery tests

timvisee:

A series of WAL delta resolution and recovery related tests. These are similar to what we planned in Notion.

Along with tests this PR brings a few fixes and improvements where needed.

This tests:
- WAL recovery with one operation
- WAL recovery with many operations
- WAL recovery with operations from different entrypoints intermixed
- WAL recovery with operations in a different order on different nodes
- WAL recovery error on empty recovery point
- WAL recovery error with unknown clock
- WAL recovery error with higher requested clock than on the source node
- WAL recovery error where no operation satisfies recovery point

There are more scenarios to be tested for this, but I suggest to do that in a different PR.

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/3631>
- [x] Rebase onto `dev`
- [x] Undraft this PR

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -121,7 +121,7 @@ impl ClockMap {
 }
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Deserialize, Serialize)]
-struct Key {
+pub(super) struct Key {
     peer_id: PeerId,
     clock_id: u32,
 }
@@ -167,10 +167,14 @@ impl Clock {
 /// recovering node has not seen yet.
 #[derive(Clone, Debug, Default)]
 pub struct RecoveryPoint {
-    clocks: HashMap<Key, u64>,
+    pub(super) clocks: HashMap<Key, u64>,
 }
 
 impl RecoveryPoint {
+    pub fn insert(&mut self, peer_id: PeerId, clock_id: u32, clock_tick: u64) {
+        self.clocks.insert(Key::new(peer_id, clock_id), clock_tick);
+    }
+
     pub fn is_empty(&self) -> bool {
         self.clocks.is_empty()
     }
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -171,6 +171,7 @@ pub struct RecoveryPoint {
 }
 
 impl RecoveryPoint {
+    #[cfg(test)]
     pub(crate) fn insert(&mut self, peer_id: PeerId, clock_id: u32, clock_tick: u64) {
         self.clocks.insert(Key::new(peer_id, clock_id), clock_tick);
     }
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -205,6 +205,18 @@ impl RecoveryPoint {
         })
     }
 
+    /// Check whether this recovery point has any lower clock value than `other`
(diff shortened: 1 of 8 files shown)


lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -168,14 +168,25 @@ impl Clock {
...
+    pub fn insert(&mut self, peer_id: PeerId, clock_id: u32, clock_tick: u64) {
+        self.clocks.insert(Key::new(peer_id, clock_id), clock_tick);
+    }
+
     pub fn is_empty(&self) -> bool {
         self.clocks.is_empty()
     }
 
+    /// Check whether this recovery point has any clocks that are not in `other`
+    pub fn has_clocks_not_in(&mut self, other: &Self) -> bool {
```
generall, line 184: Does self needs to be mutable?

generall (approved):
Added a test, which emulates streaming shard transfer (with no clock tags), and different consequences of this transfer.

Test only works partially and requires an update after truncation logic is implemented.
But I think this still can be merged

timvisee:
> Added a test, which emulates streaming shard transfer (with no clock tags), and different consequences of this transfer.
> 
> Test only works partially and requires an update after truncation logic is implemented.
> But I think this still can be merged

Nice!

I wanted to do it separately, but you beat me to it. Thanks!

I'll add a commit to use the mocked operation in all other tests too, and then it's ready for merge.

lib/collection/src/wal_delta.rs
```diff
@@ -128,10 +154,946 @@ fn resolve_wal_delta(
...
+        // No diff expected
+        // **************************************************************************************
+        // ToDo: This check is broken! Expected to be fixed after truncation logic is implemented
+        // **************************************************************************************
+        // assert_eq!(e_wal.wal.lock().read(delta_from).count(), 0);
+
+        assert_eq!(
+            delta_from.unwrap_err().to_string(),
+            "recovery point requests clocks this WAL does not know about",
+        );
```
timvisee, line 960: @generall I think this final part of the test may be wrong.

Above you mention:

> 8. Try to recover A from E (cannot diff, expect no diff, as both are supposed to be active)

But A and E they are not equal. Instead we get an error, because we cannot recover A from E. That is because A has clock C(1), which E does not know about.

timvisee: This should be resolved once truncation/cutoff has been implemented.

lib/collection/src/wal_delta.rs
```diff
@@ -162,18 +198,987 @@ fn resolve_wal_delta(
...
+            // + 1 operation from C_1
+            // + 1 operations from D (one missing)
+            // + 5 recovery operations
+            // + 1 operation from C during recovery
+            // + 1 duplicate operation from C
+            assert_eq!(b_wal.wal.lock().read(delta_from).count(), 11);
+
+            // Recover E from B
+            e_wal.append_from(&b_wal, delta_from).await.unwrap();
+        }
```
timvisee, line 983: This bit is now problematic.

The intent here is to recover E from B using diff transfer, but we cannot do that. B has a high cutoff point because it just received a full transfer, so it cannot provide a delta big enough to recover E.

@generall Could you check how we can change this to match your original intent?

generall: Neither A or E can be recovered from B, I guess. That's fine, let's change the scenario

ffuugoo (approved):
> 1k lines

TL;DR. LGTM.

PR #3611, opened 2024-02-13
Don't rebuild HNSW when updating non-indexed fields

### LINKED ISSUE #2705: Don't build HNSW when non-indexed field was added
Use case from users:
1. Update payload of each point
2. Add new field. But not index
3. QDrant reindex all data

In this scenario, reindexing is not required. By this issue it's required to avoid HNSW reindexing in this case.

xzfc:
Closes #2705.

This PR lets the caller of `SegmentHolder::apply_points_to_appendable()` (renamed to `apply_points_and_reindex`) to specify whether it is OK to update non-appendable segments.
It is safe when a non-indexed field is set or removed, or when the payload is being cleared in case if there are no indices at all.

---

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/segments_updater.rs
+++ b/lib/collection/src/collection_manager/segments_updater.rs
@@ -64,7 +64,7 @@ pub(crate) fn update_vectors(
             });
     let ids: Vec<PointIdType> = points_map.keys().copied().collect();
 
-    let updated_points = segments.apply_points_and_reindex(
+    let updated_points = segments.apply_points_with_conditional_move(
         op_num,
         &ids,
         |id, write_segment| {
@@ -116,7 +116,7 @@ pub(crate) fn overwrite_payload(
     payload: &Payload,
     points: &[PointIdType],
 ) -> CollectionResult<usize> {
-    let updated_points = segments.apply_points_and_reindex(
+    let updated_points = segments.apply_points_with_conditional_move(
         op_num,
         points,
         |id, write_segment| write_segment.set_full_payload(op_num, id, payload),
@@ -144,7 +144,7 @@ pub(crate) fn set_payload(
     points: &[PointIdType],
     key: &Option<JsonPath>,
 ) -> CollectionResult<usize> {
-    let updated_points = segments.apply_points_and_reindex(
+    let updated_points = segments.apply_points_with_conditional_move(
         op_num,
         points,
         |id, write_segment| write_segment.set_payload(op_num, id, payload, key),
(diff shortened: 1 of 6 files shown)


lib/collection/src/collection_manager/segments_updater.rs
```diff
@@ -130,15 +143,43 @@ pub(crate) fn set_payload(
...
+    // Note that if we have, e.g., an index on "a.c" and the payload being set is `{"a": {"b": 1}}`,
+    // it is not safe because the whole value of "a" is being replaced.
+    indexed_fields
+        .keys()
+        .all(|path| !payload.0.contains_key(get_path_head(path)))
+}
+
+fn safe_to_delete_payload_keys(
+    indexed_fields: &HashMap<PayloadKeyType, PayloadFieldSchema>,
+    keys_to_delete: &[PayloadKeyType],
```
xzfc, line 171: Note: accepting the whole `HashMap` is not necessarly (an `Iterator<Item = &PayloadKeyType>` would suffice), but I did it here explicitly so this function arguments won't get mixed up by accident.

lib/collection/src/collection_manager/segments_updater.rs
```diff
@@ -130,15 +143,43 @@ pub(crate) fn set_payload(
...
 ) -> CollectionResult<usize> {
-    let updated_points =
-        segments.apply_points_to_appendable(op_num, points, |id, write_segment| {
-            write_segment.set_payload(op_num, id, payload)
-        })?;
+    let updated_points = segments.apply_points_and_reindex(
+        op_num,
+        points,
+        |id, write_segment| write_segment.set_payload(op_num, id, payload),
+        |segment| safe_to_set_payload(&segment.get_indexed_fields(), payload),
```
xzfc, line 150: Note: calling `segment.get_indexed_fields()` is not the fastest way to do it (it builds a whole hashmap, but we only need the keys). But it's only done once per segment and the code is concise.

timvisee:
Then I think this is ready for merge after a rebase.

xzfc:
Converting to draft due to changes in #3548.

lib/collection/src/collection_manager/segments_updater.rs
```diff
@@ -131,15 +144,48 @@ pub(crate) fn set_payload(
...
 
+fn safe_to_set_payload(
+    indexed_fields: &HashMap<PayloadKeyType, PayloadFieldSchema>,
+    payload: &Payload,
+    key: &Option<String>,
+) -> bool {
+    if key.is_some() {
+        // Not supported yet
+        return false;
+    }
```
xzfc, line 166: Could be supported after switching to `JsonPath` parser.

timvisee (approved):
We'll need another approval from @generall or @IvanPleshkov to ensure this is sound. This is sensitive code. Thanks for the change, though!

generall (approved):

PR #3552, opened 2024-02-06
Prometheus histogram support

xzfc:
Part of to #2857.

<details>
<summary>Example /metrics output.</summary>

```
# HELP rest_responses_duration_seconds response duration histogram
# TYPE rest_responses_duration_seconds histogram
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.000001"} 0
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.000005"} 0
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.00001"} 0
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.00005"} 0
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.0001"} 0
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.0005"} 0
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.001"} 6
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.005"} 8
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.01"} 24
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.05"} 24
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.1"} 24
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="0.5"} 24
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="1"} 24
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="5"} 24
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="10"} 24
rest_responses_duration_seconds_bucket{method="POST",endpoint="/collections/{name}/points/payload",status="200",le="50"} 24
[...]

PR #3612, opened 2024-02-14
Refactor: introduce details level enum

xzfc:
Replace raw `usize` with an enum for telemetry details level. So now we can use "find references" feature of IDEs to find all the places where details level is significant.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/common/common/src/types.rs
+++ b/lib/common/common/src/types.rs
@@ -26,3 +26,35 @@ impl PartialOrd for ScoredPointOffset {
         Some(self.cmp(other))
     }
 }
+
+#[derive(Copy, Clone, Debug)]
+pub struct TelemetryDetail {
+    pub level: DetailsLevel,
+    pub histograms: bool,
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum DetailsLevel {
+    Level0,
+    Level1,
+    Level2,
+}
+
+impl Default for TelemetryDetail {
+    fn default() -> Self {
+        TelemetryDetail {
+            level: DetailsLevel::Level0,
+            histograms: false,
+        }
+    }
+}
+
+impl From<usize> for DetailsLevel {
+    fn from(value: usize) -> Self {
+        match value {
+            0 => DetailsLevel::Level0,
+            1 => DetailsLevel::Level1,
+            _ => DetailsLevel::Level2,
+        }
+    }
+}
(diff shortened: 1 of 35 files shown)


timvisee (commented):
I like this approach!

Though, we may better use `>=` rather than `>`. So that if we do `>= Level1` we show it at level 1 and higher. That makes more sense to me, because that way you could jump to all places in code where level 1 (or higher) is used as you note in the PR description.

xzfc:
Updated the PR, now it's ready for review.
The updated version introduces a boolean value `TelemetryDetail::histograms` and passing of `TelemetryDetail` to every place that potentially could call `OperationDurationsAggregator::get_statistics()`.
It is not used right now, but will be used after merging #3552.

> Though, we may better use `>=` rather than `>`.

Thanks for the suggestion, I've updated the code.

lib/common/common/src/types.rs
```diff
@@ -26,3 +26,35 @@ impl PartialOrd for ScoredPointOffset {
...
+
+impl From<usize> for DetailsLevel {
+    fn from(value: usize) -> Self {
+        match value {
+            0 => DetailsLevel::Level0,
+            1 => DetailsLevel::Level1,
+            _ => DetailsLevel::Level2,
+        }
+    }
+}
```
timvisee, line 60: We can derive `Default` like this, without any custom implementation:

```rust
#[derive(Copy, Clone, Debug, Default)]
pub struct TelemetryDetail {
    pub level: DetailsLevel,
    pub histograms: bool,
}

#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
pub enum DetailsLevel {
    #[default]
    Level0,
    Level1,
    Level2,
}
```

What do you think? I don't have a strong preference.

timvisee (approved):
As far as I can see, the level thresholds for everything are still the same, which would make it a great change :+1:

lib/common/common/src/types.rs
xzfc: I'd prefer not to have `Default` at all in the production code so it'll be always specified explicitly at the top-level.
However, it's convenient for unit/integration tests where concrete values are not important.

So, as a compromise I'd prefer to keep `Default` for `TelemetryDetail` (used in a test), but not for `DetailsLevel`.

timvisee: Check! But the we probably want to add `#[cfg(test)]` to the `impl Default` we have in this PR.

xzfc: Items under `#[cfg(test)]` couldn't be used in integration tests as they're separate crates. :( Only unit tests could use them.

PR #3606, opened 2024-02-13
refactor: OpenAPI tests

KShivendu:
It's better to keep all the tests together in one directory `tests`. Most projects have (OpenAPI) integration tests in a  `tests` directory at the root level.  

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

PR #3622, opened 2024-02-14
Datetime parser: Parse decimals without timezone

coszio:
This PR enables parsing a date like `2024-02-15T15:49:33.449041` in the same way as `2024-02-15T15:49:33.449041Z` (with or without the Z). Previously we introduced this flexibility (#3549) but it wouldn't parse the datetime if it had decimals in the timezone-less format.

Additional changes:
- capture qdrant pid via `$!` in integration tests
- fix the default openapi path when env variable is not provided (after moving this section in #3606), and add python cache to gitignore

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -2281,6 +2281,8 @@ mod tests {
     #[case::rfc_3339_and_decimals("2020-03-01T00:00:00.123456Z")]
     #[case::without_z("2020-03-01T00:00:00")]
     #[case::without_z_and_decimals("2020-03-01T00:00:00.123456")]
+    #[case::space_sep_without_z("2020-03-01 00:00:00")]
+    #[case::space_sep_without_z_and_decimals("2020-03-01 00:00:00.123456")]
     fn test_datetime_deserialization(#[case] datetime: &str) {
         let datetime = DateTimePayloadType::from_str(datetime).unwrap();
         let serialized = serde_json::to_string(&datetime).unwrap();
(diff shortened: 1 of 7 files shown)


lib/segment/src/types.rs
```diff
@@ -1426,8 +1426,8 @@ impl FromStr for DateTimePayloadType {
         // - YYYY-MM-DD HH:MM
         // - YYYY-MM-DD
         // See: <https://github.com/qdrant/qdrant/issues/3529>
-        let datetime = chrono::NaiveDateTime::parse_from_str(s, "%Y-%m-%dT%H:%M:%S")
-            .or_else(|_| chrono::NaiveDateTime::parse_from_str(s, "%Y-%m-%d %H:%M:%S"))
+        let datetime = chrono::NaiveDateTime::parse_from_str(s, "%Y-%m-%dT%H:%M:%S%.f")
+            .or_else(|_| chrono::NaiveDateTime::parse_from_str(s, "%Y-%m-%d %H:%M:%S%.f"))
```
generall, line 1430: Does this change means, that `2012-12-12 12:20:45` is not a valid date anymore?

coszio: It is still valid, the last section allows 0 or more decimals

generall: In this case, could you please add new cases to the tests, and do not replace the existing ones

coszio: Added the case of space-separated dates 👍

PR #3631, opened 2024-02-16
Add WAL cutoff clock map

timvisee:

Add a cutoff clock map to our recoverable WAL structure. This defines clocks below which we cannot resolve a WAL delta.

It is similar to the truncation clock map idea we had before, but without actually linking it to truncations. After further investigation, this seems to be unnecessary, as we can derive this from the current WAL and the highest clock map.

Also:
- update the WAL delta resolve function to consider this cutoff point
- add method to update this cutoff point based on a recovery point
- renamed last seen clock map to highest clock map

We still have to:
- test this in practice: I propose to merge this so we can test this in our other [testing](https://github.com/qdrant/qdrant/pull/3610) PRs
- set this cutoff point in case of full transfer: I propose to merge this first so we can test it before implementing it further

Roman suggested that we might want to represent the current clocks in a recovery point, rather than doing +1 when requesting it. We can do +1 locally at the place where we resolve the WAL delta. I guess that's a good improvement but I propose to do that in a separate PR so we at least have something to work with.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -175,6 +175,13 @@ impl RecoveryPoint {
         self.clocks.is_empty()
     }
 
+    /// Iterate over all recovery point entries as clock tags.
+    pub fn clock_tag_iter(&self) -> impl Iterator<Item = ClockTag> + '_ {
+        self.clocks
+            .iter()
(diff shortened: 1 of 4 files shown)


ffuugoo (approved):
LGTM

Issue #3613, opened 2024-02-14
Overwrite/set payload changes to a point wont reflect in filter/scroll queries if the key contains a period or dot


cn-mayank:
I updated payload of few points with key tb_9.6.1-1 and then tried to get count of points with the key which showed 0 points.
This was unexpected so i created a new collection and tested the same behavior and it worked fine.

The only difference between the two collection was the size of the collection itself and the key that i used to update the payload.
So in the new collection i introduced the same key with dot, underscore and hyphen and the issue reproduced.
Eliminating period from the key solved this in the new collection too.

## Current Behavior
Updated payload of points dont reflect in scroll/count with filters. The count of points with above key comes 0 even though that is not the case

## Steps to Reproduce
1. set up qdrant container as per env described below
2. create a test collection and add a dummy point to it with payload {'init': 'init'}
3. update payload by {"init":"init",  "tb_9.6.11": "f5a1166b-6ca3-4070-b57b-931bed1e1201"}
4. POST collections/test_payload/points/count
{  "filter": {
    "must_not": [
     {
        "is_empty": 
            { "key": "tb_9.6.11" }
      }
    
    ]
  }
}
5. you will get count as 0
6. change the key in payload to something without period and run above query again and count will be 1

## Expected Behavior
The count should be appropriate

## Context (Environment)
 running latest qdrant docker v1.7.4 locally on a windows machine

generall:
Hi @cn-mayank, thanks for reporting. We will see how we can fix that, but for this time I can only recommend to avoid using payload keys with dots

PR #3637, opened 2024-02-18
Introduce JsonPath

xzfc:
The current treatment of JSON paths is a bit messy in our code. Our JSON manipulation functions parse these paths in an ad-hoc way. This is error-prone, and is not convenient for building non-trivial logic on top of it. (e.g. #3611 would require it after "set payload by key" is introduced)

This PR introduces a special type `JsonPath` which contain parsed path.

Breakdown of important changes:
- Core change: replace old path-related functions in `lib/segment/src/common/utils.rs` with a new module `lib/segment/src/json_path`.
  Namely, the following:
  - `get_value_from_json_map_opt`
  - `get_value_from_json_map`
  - `remove_value_from_json_map`
  - `check_include_pattern`
  - `check_exclude_pattern`
  - `filter_json_values`
  - `JsonPathPayload::{extend, extend_or_new}` → `JsonPath::{extend, extend_or_new}`
- Core change: HTTP API functions now accept paths `JsonPath` instead of `String`. From the user's perspective, they are still strings, but they are now parsed and validated. Similarly for gRPC.
- Emphasis on validation: invalid paths are now rejected by the API. This was agreed by @generall.
- Note that empty string is considered as invalid path. I could change this to treat `key: ""` as `key: null`, but I believe it's a loosy behavior.
- Support for quoted paths, e.g., `foo."bar.baz".qux` to access `{"foo": {"bar.baz": {"qux": 1}}}`. (solves #3613)
- Drop `PayloadContainer::get_value_opt` (see my comment below).

Other changes are mostly mechanical, but they are the reason why this PR is so big:
- Use `smallvec::SmallVec` instead of `MultiValue`. It does the same thing but more cleanly.
- Replace `&str` and `String` with `JsonPath` across the codebase.
- In tests, replace hardcoded paths, e.g., `"path"` with `"path".parse().unwrap()`. These unwraps are only used in tests, the production code do not contain these unwraps.
- Move path-related tests from `lib/segment/src/common/utils.rs` and `lib/segment/src/types.rs` to `lib/segment/src/json_path/`.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/segment/src/types.rs
```diff
@@ -899,21 +896,9 @@ impl TryFrom<GeoPointShadow> for GeoPoint {
...
-    /// Absence of value and value `null` is NOT the same in this function.
-    fn get_value_opt(&self, path: &str) -> Option<MultiValue<&Value>>;
-
     /// Return value from payload by path.
-    /// If value is not present in the payload, returns empty value.
-    ///
-    /// # Warning
-    ///
-    /// Absence of value and value `null` is considered same in this function.
-    fn get_value(&self, path: &str) -> MultiValue<&Value>;
```
xzfc, line 916: The previous documentation didn't match the actual behavior. It doesn't have any treatment for `null` values. The actual difference between these methods was that `get_value_opt` distinguishes between a missing path and an expanding empty array. But since `get_value_opt` is only used in one place where the difference doesn't matter, I've removed it.

```rust
let p1 = Payload::new(json!({"foo": []}));
p1.get_value("foo[]"); // []
p1.get_value_opt("foo[]"); // Some([])

let p2 = Payload::new(json!({}));
p1.get_value("foo[]"); // []
p2.get_value_opt("foo[]"); // None
```

generall:
Hey @xzfc, the PR is way too big. Could you please split it into at least 3 smaller ones?

xzfc:
@generall Yes, I think I could split it into three parts:

1. Preparation: introduce `struct JsonPath(pub String);` and update the codebase to use it. Keep logic the same.
2. Switch `JsonPath` implementation and drop old functions in `lib/segment/src/common/utils.rs`.
3. Move tests to `lib/segment/src/json_path`.

However, I think, merging only the first part wouldn't make much sense.
Would you mind if we use [reviewable.io](https://reviewable.io/reviews/qdrant/qdrant/3637) to review this PR?
It allows per-commit review of a PR, so I could split this PR into three well-structured commits in the same PR for an easier review.

timvisee:
> However, I think, merging only the first part wouldn't make much sense.

How so? If that works towards the greater goal of implementing JsonPath everywhere, I do think it makes a lot of sense.

A feature doesn't have to be a single atomic PR.

> Would you mind if we use [reviewable.io](https://reviewable.io/reviews/qdrant/qdrant/3637) to review this PR?

I think we should avoid this, because a lot of us use built-in IDE tools for reviewing GitHub PRs, rather than depending on some web UI. I don't mind to do it on reviewable once, but others might disagree.

lib/segment/src/types.rs
timvisee: I don't see an immediate problem with this change.

Do we have a test to assert this behavior? If not, would you mind to add one?

xzfc: Since `get_value` is just a wrapper around `get_value_from_json_map`, I could an additional test case for `get_value_from_json_map`.

timvisee:
I cannot publish my review status nor comments through Reviewable due to limited OAuth App access.

I hope you'll be able to see the comments on there, because I cannot publish them here at this time.

In any case, my general verdict on this is that it is way to big to properly review. I'm not confident all choices/changes are correct, and it's very hard to track how things have moved.

Some problems are:
- Do we have to implement a custom `JsonPath` implementation, it's rather complex, and I'd prefer to use something existing if available
- We now try to manually parse a JsonPath through `nom`, I think we should avoid this if we can
- We removed some validation rules, while that is the place we should do validation at
- We must ensure it does not break our existing API, I cannot validate this at this time because it is too big

xzfc:
Closing as superseded by #3639, #3666, #3674, #3696, #3697, #3709.

PR #3419, opened 2024-01-18
fix: Include git commit id while building docker image if possible

KShivendu:
Try to include git commit ID while building docker image if .git is present. Otherwise, show a warning.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

PR #3640, opened 2024-02-19
fix: Include git commit id while building docker image if possible

KShivendu:
Try to include git commit ID while building docker image if .git is present. Otherwise, show a warning.

Attempt to fix https://github.com/qdrant/qdrant/pull/3419

### All Submissions:


Dockerfile
```diff
@@ -77,6 +77,7 @@ ARG RUSTFLAGS
 ARG LINKER=mold
 
 COPY --from=planner /qdrant/recipe.json recipe.json
+COPY . .
```
xzfc, line 80: Removing `.git` from `.dockerignore` and using `COPY . .` in `Dockerfile` puts the entire `.git` directory into a docker layer/context and thus it could render the docker caching ineffective.

Instead, I suggest to pass git revision to the docker either via an environment variable using a single file. E.g. [like this](https://www.github.com/moby/moby/issues/14191#issuecomment-242718433).

KShivendu: > puts the entire .git directory into a docker layer/context and thus it could render the docker caching ineffective.

Great point. I didn't consider this before. Thanks!

timvisee: @KShivendu I believe we have suggested this quite a few times now.

Also, I'm afraid the changes in the current PR break in the same way the last PR did, which is why it was reverted.

If not, please feel free to elaborate!

timvisee (approved):

Dockerfile
```diff
@@ -62,6 +62,9 @@ RUN case "$BUILDPLATFORM" in \
 ARG TARGETPLATFORM
 ENV TARGETPLATFORM=${TARGETPLATFORM:-linux/amd64}
 
+ARG GIT_COMMIT_ID
+ENV GIT_COMMIT_ID=${GIT_COMMIT_ID}
```
ffuugoo, line 66: Sorry for the necro-bump, but I noticed this change today, and as far as I can tell, this fucks up the Docker cache a bit... Please, mark me on any future changes to the Dockerfile. 🥲🙏

timvisee: You're right. Changing it invalidates all the next steps.

Any idea on how to improve upon this?

ffuugoo: Yeah, I'll open a PR with improvement (along with some other update/fixes) in a bit.

ffuugoo: Also, sorry if I sounded a bit harsh. This is not critical in any sense, it just pains me when I notice these sudden changes in the Dockerfile, after I painstakingly ordered every single instruction there. 😭

PR #3641, opened 2024-02-19
Do not set empty payload when moving points

agourlay:
Follow up on https://github.com/qdrant/qdrant/pull/3629 & https://github.com/qdrant/qdrant/pull/3633

This seems to be the last place where we create empty payloads according to my research.

move_if_payload


agourlay:
@generall WDYT?

timvisee (approved):

Issue #3324, opened 2024-01-04
Snapshot storage on S3


generall:
**Is your feature request related to a problem? Please describe.**

Related issue: https://github.com/qdrant/qdrant/issues/1703

Currently qdrant can only store snapshots in local filesystem (even if it is a mounted s3 bucket), but it might be convenient to upload and download snapshots from S3 directly.

**Describe the solution you'd like**

Ability to configure snapshot directory as S3 bucket, specifying access credentials.
It should be possible to list snapshots saved on attached S3 bucket as well.

**Describe alternatives you've considered**

Our current recommendations are to mount s3 as local fs, but it might be non-trivial to do in some environments.

**Additional context**

Important qualities of the solution:

- Solution should not change default behavior and should be backward compatible
- Additional dependency to work with S3 should be have reasonable size, do not contribute significantly into qdrant compile time & binary size
- Should work with all types of S3-compatible storage, not only aws, but also google storage, minio, e.t.c.

---

Note for contributors: Please consider this as tracking issue. If you think that it would be beneficial to split the task into multiple smaller PRs, please you are welcome to do so. Bounty will be rewarded for each PR independently

rajdip-b:
Won't take much time, would like to give it a shot in parallel! /attempt #3324

rajdip-b:
@generall hey, I think we will have to make the updates in the grpc and rest implementations if im not wrong. My query is, how should I tell the API to use local/s3 snapshots? Some insights into the implementation would be really helpful!

generall:
Hi @rajdip-b, no, this issue goesn't expect any API changes. S3 params should be configures with our standard configuration infra, and everything else is transparent for the user.

This issue is purely about storage backend for snapshots. Currently it assumes local finesystem and we need to make it work with s3

rajdip-b:
Okay, so we just need to swap local storage with S3, correct?

generall:
We need an ability to configure either local storage or S3.

rajdip-b:
Alright, so where should this configuration exist?

generall:
https://qdrant.tech/documentation/guides/configuration/

generall:
FYI: this issue is not closed, merged PR was aimed to initial refactoring.

Bounty is still to be paid in full.

gulshan-rs:
I guess it was done here https://github.com/qdrant/qdrant/pull/4150

timvisee:
You'd be correct. Thanks for noting!

[...]

PR #3643, opened 2024-02-19
Snapshot store

AlexandruCihodaru:
This PR is a first step in an attempt to fix https://github.com/qdrant/qdrant/issues/3324. 
To be able to add S3 support and other storage back ends later I defined a trait that will contain all needed methods to handle created snapshots. I took a simplifying assumption that the snapshots will first be created in a temp local place and after the snapshotting is completed we will move the snapshot and checksum.
After S3 support is added, it should be straight forward to implement support for additional storage back ends, the only needed action being implementing the trait for the new type. 

To keep the PR small and easy to review, currently it only handles create_snapshot,delete,list (collection/full/shard). If the direction is good others could be implemented in subsequent PRs without breaking compatibility.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -652,6 +652,10 @@ impl Collection {
     pub fn request_shard_transfer(&self, shard_transfer: ShardTransfer) {
         self.request_shard_transfer_cb.deref()(shard_transfer)
     }
+
+    pub fn shared_storage_config(&self) -> Arc<SharedStorageConfig> {
+        self.shared_storage_config.clone()
+    }
 }
 
 struct CollectionVersion;
--- a/lib/collection/src/collection/snapshots.rs
+++ b/lib/collection/src/collection/snapshots.rs
@@ -3,15 +3,12 @@ use std::path::{Path, PathBuf};
 
 use io::file_operations::read_json;
 use segment::common::version::StorageVersion as _;
-use tempfile::TempPath;
 use tokio::fs;
-use tokio::io::AsyncWriteExt;
 
 use super::Collection;
 use crate::collection::CollectionVersion;
-use crate::common::sha_256::hash_file;
 use crate::config::{CollectionConfig, ShardingMethod};
-use crate::operations::snapshot_ops::{self, SnapshotDescription};
+use crate::operations::snapshot_ops::SnapshotDescription;
 use crate::operations::types::{CollectionError, CollectionResult, NodeType};
 use crate::shards::local_shard::LocalShard;
 use crate::shards::remote_shard::RemoteShard;
@@ -126,33 +123,14 @@ impl Collection {
         });
         snapshot_temp_arc_file = archiving.await??;
 
-        // Move snapshot to permanent location.
-        // We can't move right away, because snapshot folder can be on another mounting point.
-        // We can't copy to the target location directly, because copy is not atomic.
-        // So we copy to the final location with a temporary name and then rename atomically.
-        let snapshot_path_tmp_move = snapshot_path.with_extension("tmp");
-
-        // Ensure that the temporary file is deleted on error
-        let _temp_path = TempPath::from_path(&snapshot_path_tmp_move);
-        fs::copy(&snapshot_temp_arc_file.path(), &snapshot_path_tmp_move).await?;
-
-        // compute and store the file's checksum before the final snapshot file is saved
-        // to avoid making snapshot available without checksum
-        let checksum_path = snapshot_ops::get_checksum_path(&snapshot_path);
-        let checksum = hash_file(&snapshot_path_tmp_move).await?;
-        let checksum_file = tempfile::TempPath::from_path(&checksum_path);
-        let mut file = tokio::fs::File::create(checksum_path.as_path()).await?;
-        file.write_all(checksum.as_bytes()).await?;
-
-        let snapshot_file = tempfile::TempPath::from_path(&snapshot_path);
-        fs::rename(&snapshot_path_tmp_move, &snapshot_path).await?;
-
-        // Snapshot files are ready now, so keep them
-        snapshot_file.keep()?;
-        checksum_file.keep()?;
-
-        log::info!("Collection snapshot {snapshot_name} completed into {snapshot_path:?}");
-        snapshot_ops::get_snapshot_description(&snapshot_path).await
+        let snapshot_manager = self.shared_storage_config.snapshot_manager();
+        snapshot_manager
+            .store_file(
+                snapshot_temp_arc_file.path(),
+                snapshot_path.as_path(),
+                false,
+            )
+            .await
     }
 
     /// Restore collection from snapshot
--- a/lib/collection/src/collection/snapshots.rs
+++ b/lib/collection/src/collection/snapshots.rs
@@ -130,10 +130,7 @@ impl Collection {
 
         let snapshot_manager = self.get_snapshots_storage_manager();
         snapshot_manager
-            .store_file(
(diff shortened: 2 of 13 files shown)


lib/storage/src/types.rs
```diff
@@ -53,7 +53,7 @@ pub struct StorageConfig {
     pub storage_path: String,
     #[serde(default = "default_snapshots_path")]
     #[validate(length(min = 1))]
-    pub snapshots_path: String,
+    pub snapshots_path: Option<String>,
```
AlexandruCihodaru, line 56: This change is needed because in the future there could be either snapshot_path for local storage either some details for the storage back end. It would have been cleaner with an enum in here but I believe it would break backwards compatibility.

lib/collection/src/collection/mod.rs
```diff
@@ -141,6 +143,7 @@ impl Collection {
             this_peer_id,
             path: path.to_owned(),
             snapshots_path: snapshots_path.to_owned(),
+            snapshot_storage_manager: Self::create_snapshot_storage_manager(snapshots_path),
             channel_service,
```
AlexandruCihodaru, line 147: I think that in the future snapshots_path could be remove so we will avoid info duplication, the path can be retrieved from the snapshot_storage_manager.

lib/collection/src/collection/mod.rs
```diff
@@ -50,6 +51,7 @@ pub struct Collection {
     this_peer_id: PeerId,
     path: PathBuf,
     snapshots_path: PathBuf,
+    snapshot_storage_manager: Box<dyn SnapshotStorage>,
```
AlexandruCihodaru, line 54: This could be static dispatch by making Collection generic over SnapshotStorage. We could switch to it if snapshotting happens often because it is cheaper.

lib/collection/src/collection/snapshots.rs
```diff
@@ -135,23 +134,10 @@ impl Collection {
...
-
-        // Snapshot files are ready now, so keep them
-        snapshot_file.keep()?;
-        checksum_file.keep()?;
-
-        log::info!("Collection snapshot {snapshot_name} completed into {snapshot_path:?}");
-        snapshot_ops::get_snapshot_description(&snapshot_path).await
+        let snapshot_storage_manager = &self.snapshot_storage_manager;
+        snapshot_storage_manager
+            .store_file(&snapshot_path_tmp_move, snapshot_path.as_path())
```
generall, line 139: Actual moving to another location (mount or S3) happens n the `fs::copy` statement above. 
Therefore it should also be habdled by `snapshot_storage_manager`

lib/collection/src/shards/transfer/snapshot.rs
```diff
@@ -188,10 +189,17 @@ pub(super) async fn transfer_snapshot(
         "Local shard must be a queue proxy"
     );
 
+    let snapshot_storage_manager = Collection::create_snapshot_storage_manager(snapshots_path);
```
generall, line 192: The fact that we are using `create_snapshot_storage_manager` from shards suggesrts that it should not be part of `Collection`.
Ideally shards should not know anything about Collection existance, otherwise it is circular dependency.

AlexandruCihodaru: Do you think that lib/common/common would be a good location?

generall: I would do it as a `::new` with config and path as arguments. Since we do have access to both in all cases

lib/collection/src/collection/snapshots.rs
generall: Not related to this PR, but checksum should be computed before copying

lib/storage/src/types.rs
generall: :check:

lib/collection/src/shards/transfer/snapshot.rs
AlexandruCihodaru: I think we don't have access to StorageConfig (which will hold the info for S3/LocaFS) from the shard holder, it does have SharedStorageConfig, so we could update that structure I think :shrug:

generall: `transferring_shard` have access to `shared_storage_config`. We can have a method to create SnapshotManager in both Collection and ShardReplicaSet

AlexandruCihodaru: Yes but SharedStorageConfig https://github.com/qdrant/qdrant/blob/d0baa5ef4a4bc2c597b28d0e1f1011262be8ed4d/lib/collection/src/operations/shared_storage_config.rs#L17 does not have the needed info for handling snapshot transfer

generall: Good point. I would add this info into `SharedStorageConfig` then

lib/collection/src/collection/snapshots.rs
AlexandruCihodaru: Right. I moved it, also moved the checksum computation before copying.

lib/storage/src/content_manager/snapshots/mod.rs
```diff
@@ -44,9 +41,15 @@ pub async fn do_delete_full_snapshot(
...
 ) -> Result<bool, StorageError> {
     let dispatcher = dispatcher.clone();
-    let snapshot_name = snapshot_name.to_string();
-    let task =
-        tokio::spawn(async move { _do_delete_full_snapshot(&dispatcher, &snapshot_name).await });
+    let storage_config = dispatcher
+        .clone()
+        .toc()
+        .storage_config
+        .to_shared_storage_config(false);
```
AlexandruCihodaru, line 48: Here I am creating the shared storage config from the storage config to avoid duplicating the *snapshot_manager* method.

AlexandruCihodaru:
Hey @generall, I have addressed your comments. In a subsequent PR we should also move the get_snapshot_description and get_snapshot/checksum_path into the trait because we will have different implementations for the LocalFS and S3.

lib/storage/src/types.rs
AlexandruCihodaru: @generall an enum would mean to also have a custom deserialization implementation and everything, overall I felt that it was cleaner how I implemented it here, but for reference this is how it would look like https://github.com/AlexandruCihodaru/qdrant/commit/887d42631996c554e419f8b54eb2b95c397914b6

lib/collection/src/collection/mod.rs
```diff
@@ -46,7 +46,7 @@ pub struct Collection {
     pub(crate) id: CollectionId,
     pub(crate) shards_holder: Arc<LockedShardHolder>,
     pub(crate) collection_config: Arc<RwLock<CollectionConfig>>,
-    pub(crate) shared_storage_config: Arc<SharedStorageConfig>,
+    pub shared_storage_config: Arc<SharedStorageConfig>,
```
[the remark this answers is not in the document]
AlexandruCihodaru: Definitely.

AlexandruCihodaru:
Hey @generall  I was wondering if the desired outcome is to use S3 like a "persistent" storage or to use it for any snapshot related actions. If we targeting the first option than I think it is possible to bring locally from S3 the snapshot dir before initializing the TableOfContent in here https://github.com/qdrant/qdrant/blob/d0baa5ef4a4bc2c597b28d0e1f1011262be8ed4d/lib/storage/src/content_manager/toc/mod.rs#L94 and I believe that there will not be a lot of additional changes after this PR.

generall (approved):

PR #3645, opened 2024-02-19
Diff transfer improvements

ffuugoo:
Minor fixes/improvements/cleanups for diff transfer related code.

Tracked in: #3477

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -176,11 +176,6 @@ pub struct RecoveryPoint {
 }
 
 impl RecoveryPoint {
-    #[cfg(test)]
-    pub(crate) fn insert(&mut self, peer_id: PeerId, clock_id: u32, clock_tick: u64) {
-        self.clocks.insert(Key::new(peer_id, clock_id), clock_tick);
-    }
-
     pub fn is_empty(&self) -> bool {
         self.clocks.is_empty()
     }
@@ -263,6 +258,11 @@ impl RecoveryPoint {
                 self.clocks.remove(&key);
             }
         }
+    }
+
+    #[cfg(test)]
+    pub(crate) fn insert(&mut self, peer_id: PeerId, clock_id: u32, clock_tick: u64) {
+        self.clocks.insert(Key::new(peer_id, clock_id), clock_tick);
     }
 }
 
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -75,7 +75,7 @@ impl ClockMap {
     /// Advance clock referenced by `clock_tag` to `clock_tick`, if it's newer than current tick.
     ///
     /// If the clock is not yet tracked by the `ClockMap`, it is initialized to
-    /// the `clock_tag.clock_tick` and added to the `ClockMap`.
+    /// the `clock_tick` and added to the `ClockMap`.
     pub fn advance_clock(&mut self, clock_tag: ClockTag) {
         let _ = self.advance_clock_impl(clock_tag);
     }
@@ -181,15 +181,17 @@ impl RecoveryPoint {
     }
 
     /// Iterate over all recovery point entries as clock tags.
-    pub fn clock_tag_iter(&self) -> impl Iterator<Item = ClockTag> + '_ {
+    pub fn iter_as_clock_tags(&self) -> impl Iterator<Item = ClockTag> + '_ {
         self.clocks
             .iter()
-            .map(|(key, tick)| ClockTag::new(key.peer_id, key.clock_id, *tick))
+            .map(|(key, &tick)| ClockTag::new(key.peer_id, key.clock_id, tick))
     }
 
     /// Increase all existing clocks in this recovery point by the given amount
-    pub fn increase_by(&mut self, amount: u64) {
-        self.clocks.values_mut().for_each(|tick| *tick += amount);
+    pub fn increase_all_clocks_by(&mut self, ticks: u64) {
+        for current_tick in self.clocks.values_mut() {
+            *current_tick += ticks;
+        }
     }
 
     /// Check whether this recovery point has any clocks that are not in `other`
@@ -199,49 +201,50 @@ impl RecoveryPoint {
             .any(|key| !other.clocks.contains_key(key))
     }
 
-    /// Check whether this recovery point has any higher clock value than `other`
+    /// Check if this recovery point has any clock that is newer than the one in the `other`.
     ///
-    /// A clock in this recovery point that is not in `other` is always considered to be higher.
-    pub fn has_any_higher(&self, other: &Self) -> bool {
-        self.clocks.iter().any(|(key, tick)| {
+    /// A clock that is present in this recovery point, but not in the `other`,
+    /// is always considered to be *newer*.
+    pub fn has_any_newer_clocks_than(&mut self, other: &Self) -> bool {
+        self.clocks.iter().any(|(key, &tick)| {
             other
                 .clocks
                 .get(key)
-                .map_or(true, |other_tick| *tick > *other_tick)
+                .map_or(true, |&other_tick| tick > other_tick)
         })
     }
 
-    /// Check whether this recovery point has any lower clock value than `other`
+    /// Check if this recovery point has any clock that is older than the one in the `other`.
     ///
-    /// A clock in this recovery point that is not in `other` is not considered lower.
-    pub fn has_any_lower(&self, other: &Self) -> bool {
-        self.clocks.iter().any(|(key, tick)| {
+    /// A clock that is present in this recovery point, but not in the `other`,
+    /// is always considered to be *newer*.
+    pub fn has_any_older_clocks_than(&self, other: &Self) -> bool {
+        self.clocks.iter().any(|(key, &tick)| {
             other
                 .clocks
                 .get(key)
-                .map_or(false, |other_tick| *tick < *other_tick)
+                .map_or(false, |&other_tick| tick < other_tick)
         })
     }
 
-    /// Extend this recovery point with new clocks from `clock_map`
+    /// Extend this recovery point with clocks that are only present in the `other`.
     ///
-    /// Clocks that we have not seen yet are added with a tick of `0`, because we must recover all
-    /// records for it.
+    /// Clocks that are not present in this recovery point are initialized to the tick `0`,
+    /// because we must recover all operations for them.
     ///
-    /// Clocks that we already have in this recovery point are not updated, regardless of their
-    /// tick value.
-    pub fn extend_with_missing_clocks(&mut self, other: &Self) {
+    /// Clocks that are already present in this recovery point are not updated.
+    pub fn initialize_clocks_missing_from(&mut self, other: &Self) {
         // Clocks known on our node, that are not in the recovery point, are unknown on the
         // recovering node. Add them here with tick 0, so that we include all records for it
-        for key in other.clocks.keys() {
-            self.clocks.entry(*key).or_insert(0);
+        for &key in other.clocks.keys() {
+            self.clocks.entry(key).or_insert(0);
         }
     }
 
-    /// Remove clocks from this recovery point, if they are equal in `other`
-    pub fn remove_equal_clocks(&mut self, other: &Self) {
-        for (key, tick) in &other.clocks {
-            if let Some(other_tick) = self.clocks.get(key) {
+    /// Remove clocks from this recovery point, that are equal to the clocks in the `other`.
+    pub fn remove_clocks_equal_to(&mut self, other: &Self) {
+        for (key, other_tick) in &other.clocks {
+            if let Some(tick) = self.clocks.get(key) {
                 if tick == other_tick {
                     self.clocks.remove(key);
                 }
@@ -249,12 +252,13 @@ impl RecoveryPoint {
         }
     }
 
-    /// Remove a clock from this recovery point, if the given `clock_tag` describes an equal or
-    /// lower clock value.
-    pub fn remove_equal_or_lower(&mut self, clock_tag: ClockTag) {
-        let key = Key::from_tag(clock_tag);
-        if let Some(tick) = self.clocks.get(&key) {
-            if clock_tag.clock_tick <= *tick {
+    /// Remove a clock referenced by the clock tag from this recovery point, if the clock is
+    /// *newer or equal* than the tick in the tag.
+    pub fn remove_clock_if_newer_than_or_equal_to_tag(&mut self, tag: ClockTag) {
+        let key = Key::from_tag(tag);
+
+        if let Some(&tick) = self.clocks.get(&key) {
+            if tick >= tag.clock_tick {
                 self.clocks.remove(&key);
             }
         }
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -959,8 +959,6 @@ impl LocalShardClocks {
 
     /// Persist clock maps to disk
     pub async fn store(&self, shard_path: &Path) -> CollectionResult<()> {
-        // TODO: Is order important here!? If it is, it should be documented with a comment!
-
         self.oldest_resolvable_clocks
             .lock()
             .await
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -512,7 +512,7 @@ impl LocalShard {
         // (`SerdeWal::read_all` may even start reading WAL from some already truncated
         // index *occasionally*), but the storage can handle it.
 
-        for (op_num, update) in wal.read_all() {
+        for (op_num, update) in wal.read_all(false) {
             if let Some(clock_tag) = update.clock_tag {
                 highest_clocks.advance_clock(clock_tag);
             }
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -165,7 +165,7 @@ impl LocalShard {
             segments: segment_holder,
             collection_config,
             shared_storage_config,
-            wal: RecoverableWal::from(locked_wal, clocks.highest_clocks, clocks.cutoff_clocks),
+            wal: RecoverableWal::new(locked_wal, clocks.highest_clocks, clocks.cutoff_clocks),
             update_handler: Arc::new(Mutex::new(update_handler)),
             update_sender: ArcSwap::from_pointee(update_sender),
             update_tracker,
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -130,15 +130,13 @@ impl LocalShard {
         optimizers: Arc<Vec<Arc<Optimizer>>>,
         optimizer_cpu_budget: CpuBudget,
         shard_path: &Path,
-        highest_clock_map: ClockMap,
-        cutoff_clock_map: ClockMap,
+        clocks: LocalShardClocks,
         update_runtime: Handle,
     ) -> Self {
         let segment_holder = Arc::new(RwLock::new(segment_holder));
         let config = collection_config.read().await;
         let locked_wal = Arc::new(ParkingMutex::new(wal));
         let optimizers_log = Arc::new(ParkingMutex::new(Default::default()));
-        let clocks = LocalShardClocks::new_at(highest_clock_map, cutoff_clock_map, shard_path);
 
         let mut update_handler = UpdateHandler::new(
             shared_storage_config.clone(),
@@ -151,6 +149,7 @@ impl LocalShard {
             config.optimizer_config.flush_interval_sec,
             config.optimizer_config.max_optimization_threads,
             clocks.clone(),
+            shard_path.into(),
         );
 
         let (update_sender, update_receiver) =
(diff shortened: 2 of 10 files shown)


lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -544,7 +544,7 @@ impl LocalShard {
         // (`SerdeWal::read_all` may even start reading WAL from some already truncated
         // index *occasionally*), but the storage can handle it.
 
-        for (op_num, update) in wal.read_all() {
+        for (op_num, update) in wal.read(wal.first_index()) {
```
timvisee, line 547: Why do we want to be more explicit here, isn't `read_all()` enough, fulfilling this purpose?

lib/collection/src/shards/shard.rs
```diff
@@ -151,7 +151,10 @@ impl Shard {
     pub async fn shard_recovery_point(&self) -> CollectionResult<RecoveryPoint> {
         match self {
             Self::Local(local_shard) => Ok(local_shard.recovery_point().await),
+
+            // TODO: Is it good idea to return `RecoveryPoint` for `ForwardProxyShard`? 🤔
```
timvisee, line 155: Maybe not, but my thought was that it's still a regular shard internally.

I guess we can assume how that we only request a recovery point if we want to recover this node, in which case it should not have a forward proxy.

So I'm fine by limiting it to be cautious, and removing this here.

lib/collection/src/shards/shard.rs
```diff
@@ -167,13 +170,16 @@ impl Shard {
     ) -> CollectionResult<Option<u64>> {
         let resolve_result = match self {
             Self::Local(local_shard) => local_shard.wal.resolve_wal_delta(recovery_point).await,
+
+            // TODO: Same as `shard_recovery_point`, is it a good idea to resolve WAL delta for `ForwardProxyShard`?
```
timvisee, line 174: Same as the above basically, I've only routed it here because we can.

lib/collection/src/wal.rs
```diff
@@ -110,8 +110,16 @@ impl<'s, R: DeserializeOwned + Serialize + Debug> SerdeWal<R> {
             .map_err(|err| WalError::WriteWalError(format!("{err:?}")))
     }
 
-    pub fn read_all(&'s self) -> impl Iterator<Item = (u64, R)> + 's {
-        self.read(self.first_index())
+    pub fn read(&self, start_index: u64) -> impl DoubleEndedIterator<Item = (u64, R)> + '_ {
+        (start_index..self.last_index()).map(move |idx| {
```
timvisee, line 114: I thought the `last_index()` is inclusive. Which is interesting, because if that is the case, it means the previous implementation is broken too.

```suggestion
        (start_index..=self.last_index()).map(move |idx| {
```

Note: I'm investigating!

lib/collection/src/wal.rs
```diff
@@ -110,8 +110,16 @@ impl<'s, R: DeserializeOwned + Serialize + Debug> SerdeWal<R> {
             .map_err(|err| WalError::WriteWalError(format!("{err:?}")))
     }
 
-    pub fn read_all(&'s self) -> impl Iterator<Item = (u64, R)> + 's {
-        self.read(self.first_index())
+    pub fn read(&self, start_index: u64) -> impl DoubleEndedIterator<Item = (u64, R)> + '_ {
```
timvisee, line 113: I like that we use `DoubleEndedIterator` here now so we have a single function. But I must admit, this way I find `start_index` very confusing.

I'd rather have a `read_all(with_acknowledged: bool)` which is also doubly ended to prevent any confusion here.

I'd strongly vote for the latter because it prevents exposing unnecessary implementation details - such as record indices - to the public.

lib/collection/src/wal_delta.rs
```diff
@@ -14,8 +14,12 @@ pub type LockedWal = Arc<ParkingMutex<SerdeWal<OperationWithClockTag>>>;
 /// A WAL that is recoverable, with operations having clock tags and a corresponding clock map.
 pub struct RecoverableWal {
     pub(super) wal: LockedWal,
+
     /// Map of all highest seen clocks for each peer and clock ID.
+    //
+    // TODO: Rename to `newest_clocks` (or `newest_*` something)?
```
timvisee, line 20: Sure! I see that you've also used _new_ in other places, so let's follow that here then.

Please also update the file name to `clock_map_newest.json`.

lib/collection/src/wal_delta.rs
```diff
@@ -24,11 +28,14 @@ pub struct RecoverableWal {
     /// This means two things:
     /// - this WAL has at least all these clock versions (same as `highest_clocks`)
     /// - this WAL cannot resolve any delta below any of these clocks
+    //
+    // TODO: Rename to `oldest_available_records` or `oldest_resolvable_records`?
+    // TODO: What does "same as `highest_clocks`" note above mean?
```
timvisee, line 33: It means that entries in `cutoff_clocks` are never newer than `highest_clocks`.

So, if we update cutoff, we also update `highest_clocks`.

lib/collection/src/wal_delta.rs
```diff
@@ -24,11 +28,14 @@ pub struct RecoverableWal {
     /// This means two things:
     /// - this WAL has at least all these clock versions (same as `highest_clocks`)
     /// - this WAL cannot resolve any delta below any of these clocks
+    //
+    // TODO: Rename to `oldest_available_records` or `oldest_resolvable_records`?
```
timvisee, line 32: Don't really like that name, but I don't have a better suggestion. I'd prefer something short.

lib/collection/src/wal_delta.rs
```diff
@@ -81,18 +92,19 @@ impl RecoverableWal {
...
+            // A recovery point has every clock one higher, here we decrease it by one again
+            //
+            // TODO: Refactor `RecoveryPoint` to transfer *current* clocks instead of *next* clocks!
+            clock_tag.clock_tick -= 1;
+
+            // Advance highest clocks we've seen
+            highest_clocks.advance_clock(clock_tag);
+
+            // Advance the cutoff point
+            cutoff_clocks.advance_clock(clock_tag);
```
timvisee, line 106: Nitpick, but after reading this a bunch of times I don't think we have to be overly explicit.

```suggestion
            clock_tag.clock_tick -= 1;
            highest_clocks.advance_clock(clock_tag);
            cutoff_clocks.advance_clock(clock_tag);
```

lib/collection/src/wal_delta.rs
```diff
@@ -192,13 +204,16 @@ fn resolve_wal_delta(
...
     log::trace!("Resolving WAL delta for: {recovery_point}");
-    let delta_from = local_wal
-        .lock()
-        .read_from_last(true)
+
+    let wal = local_wal.lock();
+
+    let delta_from = wal
+        .read(wal.first_closed_index())
+        .rev()
```
timvisee, line 212: This is where I'd prefer the following:

```suggestion
        .read_all(true)
        .rev()
```

timvisee (commented):
Some of my comments on the current state.

lib/collection/src/wal.rs
timvisee: This is the case indeed! :exploding_head: 

```rust
    /// Last index that is still available in logical WAL.
    pub fn last_index(&self) -> u64 {
        self.wal.last_index()
    }
```

which points to:

```rust
    /// The index of the last entry
    pub fn last_index(&self) -> u64 {
        let num_entries = self.num_entries();
        self.first_index() + num_entries.saturating_sub(1)
    }
```

The `num_entries.saturating_sub(1)` means the higher bound is inclusive.

So we want to loop over that records as well:

```diff
-(start_index..self.last_index())
+(start_index..=self.last_index())
```

lib/collection/src/shards/local_shard/mod.rs
timvisee: Makes sense!

Yes, I'd vote for a `read_all(with_acknowledged)` which defines `first_index` or the other internally, returning back the doubly ended iterator.

ffuugoo: Tbh, I don't like these `with_acknowledged` methods. In my opinion, it makes *very* little difference with just doing `wal.read(wal.some_specific_index())`. Especially when we only ever use `read_all` *once* and always with `with_acknowledged = false`.

But I only *slightly* disagree with this, so I'm not too against `read_all(with_acknowledged)`.

lib/collection/src/wal.rs
ffuugoo: Yeah, that's probably just a bug I've introduced. Thanks. (Maybe that's why tests failed. This PR is more of idea dump atm.)

lib/collection/src/wal_delta.rs
ffuugoo: > It means that entries in cutoff_clocks are never newer than highest_clocks.
> So, if we update cutoff, we also update highest_clocks.

I do get this... but I still don't see how "same as `highest_clocks`" can mean this. I'll just reword the quote above and replace this part later. :D

lib/collection/src/wal_delta.rs
ffuugoo: Hmmm. Maybe you're right. Ok, I'll return the `read_all`. 👍

lib/collection/src/shards/local_shard/mod.rs
timvisee: > Especially when we only ever use read_all once and always with with_acknowledged = false.

We don't.

When resolving the diff it would use `read_all(true)`.

lib/collection/src/wal.rs
ffuugoo: I think I have to agree now. I'll return the `read_all` method back. I also don't like these indices, cause they are confusing as hell.

lib/collection/src/wal_delta.rs
timvisee: It notes 'at least', but I agree it is confusing.

> I'll just reword the quote above and replace this part later. :D

Thanks!

lib/collection/src/shards/local_shard/mod.rs
ffuugoo: Yeah, I see now. Before we only used `read_all` in a single place. It makes sense now. 👍

lib/collection/src/wal.rs
ffuugoo: Fixed this, to see if it will fix CI.

ffuugoo:
Resolved all previous comments, organized history, rebased on `dev`, added a few more improvements. Also removed proto file changes (will open a separate PR in a moment).

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -934,85 +932,105 @@ impl Drop for LocalShard {
...
 /// Holds a clock map for tracking the highest clocks and the cutoff clocks.
-#[derive(Clone)]
+#[derive(Clone, Debug, Default)]
 pub struct LocalShardClocks {
-    highest_clocks: Arc<Mutex<ClockMap>>,
-    cutoff_clocks: Arc<Mutex<ClockMap>>,
-    highest_clocks_path: PathBuf,
-    cutoff_clocks_path: PathBuf,
+    newest_observed_clocks: Arc<Mutex<ClockMap>>,
+    oldest_resolvable_clocks: Arc<Mutex<ClockMap>>,
```
timvisee, line 938: I like short ones better, maybe even: `newest`, `oldest`

We can explain details in comments if we like.

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -247,15 +247,21 @@ impl ShardOperation for LocalShard {
...
+            // It is *critical* to hold `_wal_lock` while sending operation to the update handler!
+            let (operation_id, _wal_lock) = self.wal.lock_and_write(&mut operation).await?;
 
             channel_permit.send(UpdateSignal::Operation(OperationData {
                 op_num: operation_id,
                 operation: operation.operation,
                 sender: callback_sender,
                 wait,
             }));
-            drop(wal_lock);
```
timvisee, line 258: I put it here so we don't make the same mistake again :smile:

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -934,85 +932,105 @@ impl Drop for LocalShard {
...
+
         Ok(())
     }
 
     /// Delete clock data from disk at the given shard path.
     pub async fn delete_data(shard_path: &Path) -> CollectionResult<()> {
-        let highest_path = Self::highest_clock_map_path(shard_path);
-        let cutoff_path = Self::cutoff_clock_map_path(shard_path);
+        let highest_path = Self::newest_observed_clock_map_path(shard_path);
+        let cutoff_path = Self::oldest_resolvable_clock_map_path(shard_path);
```
timvisee, line 1016: Update variable names

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -934,85 +932,105 @@ impl Drop for LocalShard {
...
+
         Ok(())
     }
 
     /// Move clock data on disk from one shard path to another.
     pub async fn move_data(from: &Path, to: &Path) -> CollectionResult<()> {
-        let highest_path_from = Self::highest_clock_map_path(from);
-        let cutoff_path_from = Self::cutoff_clock_map_path(from);
+        let highest_path_from = Self::newest_observed_clock_map_path(from);
+        let cutoff_path_from = Self::oldest_resolvable_clock_map_path(from);
```
timvisee, line 998: Update variable names

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -934,85 +932,105 @@ impl Drop for LocalShard {
...
-        shard_path.join("clock_map_cutoff.json")
+        Ok(())
     }
 
     /// Copy clock data on disk from one shard path to another.
     pub async fn copy_data(from: &Path, to: &Path) -> CollectionResult<()> {
-        let highest_path_from = Self::highest_clock_map_path(from);
-        let cutoff_clock_map_path = Self::cutoff_clock_map_path(from);
+        let highest_path_from = Self::newest_observed_clock_map_path(from);
+        let cutoff_clock_map_path = Self::oldest_resolvable_clock_map_path(from);
```
timvisee, line 980: Update variable names

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -934,85 +932,105 @@ impl Drop for LocalShard {
...
+            ClockMap::load_or_default(&Self::oldest_resolvable_clock_map_path(shard_path))?;
+
+        Ok(Self::new(newest_observed_clocks, oldest_resolvable_clocks))
+    }
+
     /// Persist clock maps to disk
-    pub async fn store(&self) -> CollectionResult<()> {
-        self.cutoff_clocks
+    pub async fn store(&self, shard_path: &Path) -> CollectionResult<()> {
+        // TODO: Is order important here!? If it is, it should be documented with a comment!
```
timvisee, line 962: My reasoning was that the oldest recoverable are more important than the newest seen with regards to consistency.

But I'm not sure if that's actually true.

Before I locked both together but that would be a potential foot gun for a deadlock. I left a comment regarding this, but it's gone for some reason.

lib/collection/src/wal_delta.rs
```diff
@@ -154,10 +176,10 @@ impl RecoverableWal {
 /// If `None` - the remote WAL is already equal, and we don't have to send any records.
 /// If `Err` - no delta can be resolved.
 fn resolve_wal_delta(
+    operations: impl DoubleEndedIterator<Item = (u64, Option<ClockTag>)>,
```
timvisee, line 179: This one is potentially dangerous, because we say nothing about the order we expect.

timvisee (commented):
Some comments (and :pick:s) on a quick pass.

lib/collection/src/shards/local_shard/shard_ops.rs
ffuugoo: Ok, I kinda get your point now, but, IMO, it's not very effective and this part simply requires an explanation comment (with a lot of exclamation marks in it).

It's still super easy to just look at this code, see the `drop`, think "WTF is this here for?" and just remove the `drop` and refactor the `wal_lock` away same way we did earlier.

If we add the comment, then it will explain that the `drop` is *intentional*, and we *have to* hold the `wal_lock` for the duration of `send`... but at this point there's not much use in doing explicit `drop`, IMO.

ffuugoo:
Ok, so the failing CI was my own fault (obviously, lol). It's fixed now. I think we can give it one last pass and merge.

lib/collection/src/shards/local_shard/mod.rs
timvisee: I think we can remove this TODO.

lib/collection/src/wal.rs
```diff
@@ -110,8 +110,35 @@ impl<'s, R: DeserializeOwned + Serialize + Debug> SerdeWal<R> {
...
+            self.read(self.first_index())
+        }
+    }
+
+    pub fn read(&self, from: u64) -> impl DoubleEndedIterator<Item = (u64, R)> + '_ {
+        // We have to explicitly do `from..self.first_index() + self.len(false)`, instead of more
+        // concise `from..=self.last_index()`, because if the WAL is empty, `Wal::last_index`
+        // returns `Wal::first_index`, so we end up with `1..=1` instead of an empty range. 😕
+        //
+        // TODO: Refactor `SerdeWal::last_index`/`Wal::last_index` to return `Option<u64>`? 🤔
```
timvisee, line 129: :+1:

lib/collection/src/wal.rs
```diff
@@ -251,6 +240,8 @@ impl<'s, R: DeserializeOwned + Serialize + Debug> SerdeWal<R> {
     ///
     /// If the index is lower than `first_index`, it means we have already acknowledged it but we
     /// are still holding it in a closed segment until it gets truncated.
+    //
+    // TODO: Rename to `first_acknowledged_index`? Or to something else more meaningful than "closed"? 🤔
```
timvisee, line 244: `first_acknowledged_index` fine by me, I don't have a strong opinion on either. I do find both names still a bit confusing though.

I choose `closed` at first because we use the same terminology in other places.

lib/collection/src/shards/local_shard/shard_ops.rs
timvisee: If I see a `drop` I always think: it must be there for a reason, lets be very careful about it.

Comments don't do the same to me while refactoring.

But anyway, fine with both.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -204,66 +201,73 @@ impl RecoveryPoint {
...
 
-    /// Remove a clock from this recovery point, if the given `clock_tag` describes an equal or
-    /// lower clock value.
-    pub fn remove_equal_or_lower(&mut self, clock_tag: ClockTag) {
-        let key = Key::from_tag(clock_tag);
-        if let Some(tick) = self.clocks.get(&key) {
-            if clock_tag.clock_tick <= *tick {
+    /// Remove a clock referenced by the clock tag from this recovery point, if the clock is
+    /// *newer or equal* than the tick in the tag.
+    pub fn remove_clock_if_newer_than_or_equal_to_tag(&mut self, tag: ClockTag) {
```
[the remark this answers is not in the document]
ffuugoo: I don't see how it is clear. The argument name does not change the fact, that you can read "if newer or equal" both ways. What are we comparing to what? `clock > tag` or `tag > clock`?

It does make more sense to read "left to right", e.g., `clock > tag`, but, like, do you *absolutely always* do the most logical thing, or are you, like, human? :D

In the end, you *have to* go read the documentation/code to verify that it is indeed `clock > tag`. Using `newer_than_or_equal_to_tag` clarifies `clock > tag`. It is clear to you, because you've been working with it for the last two weeks.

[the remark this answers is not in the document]
timvisee: `remove_ge_tag`?

The only reasonable assumption I can make about that is: we remove clocks, if greater or equal to the given tag.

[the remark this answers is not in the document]
timvisee: Or `retain_old_clocks(tag)` with a comment describing the exact condition.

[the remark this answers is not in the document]
ffuugoo: > `remove_ge_tag`?
> The only reasonable assumption I can make about that is: we remove clocks, if greater or equal to the given tag.

You still can read it both ways. Like, it literally reads as "remove greater or equal tag". Which is a lot like "remove if *tag* is greater of equal". Why are you so against the conjunctions? Just add `remove_ge_to_tag`, and it's clear. Though, I'd still prefer fully verbose one.

> Or `retain_old_clocks(tag)` with a comment describing the exact condition.

That's exactly what I *don't* want. I don't want to *have to* go read the comment. I want it to be obvious when you just read the method name. :/

lib/collection/src/shards/local_shard/shard_ops.rs
ffuugoo: Hmmm...... I usually do as well, but in this case it's _kinda_ obvious that it's the end of the scope, so I usually assume it's just a leftover from a refactoring or something.

lib/collection/src/wal.rs
ffuugoo: I'd just remove the `TODO` then. Don't feel much like fixing it now, especially, because then `SerdeWall::last_index` and `Wal::last_index` will behave different.

lib/collection/src/wal.rs
ffuugoo: Decided to ignore this for now, cause I realized what confuses me about the whole `first_index`/`first_closed_index`, is that it _kinda_ makes more sense to give a more specific name to `first_index`, instead of to `first_closed_index`.

Like, `first_closed_index` *is* **the first** index of WAL. And the `first_index` is, like, some semi-arbitrary index somewhere in the WAL. I don't want to dive deeer into it RN, so I leave it as-is. (Just added it to my personal list of Qdrant `TODO`s. :D)

timvisee: Under normal circumstances you'd only use the logical (open) part of the WAL. So in that case, `first_index` probably makes sense. Its only with WAL diff that we want to read over the closed bit too, just because we have that information. Otherwise we'd significantly limit our diffing possibilities.

What about naming it `first_{open,closed}_index()` for now? That may prevent some confusion because you then have to choose a specific method.

timvisee (approved):

PR #3658, opened 2024-02-21
Test: fix shard snapshot transfer flakyness

timvisee:
This is not intended to be merged.

Tests <https://github.com/qdrant/qdrant/pull/3642> with some testing tweaks on top.


timvisee:
Closing because <https://github.com/qdrant/qdrant/pull/3642> is merged, which has the actual meat of this change.

PR #3659, opened 2024-02-21
Add replica state: Recovery

timvisee:

Add a new `ReplicaState::Recovery` state. We don't send updates to this state by default. It only accepts operations that had a force flag set, and rejects all others. We want to have this for shard diff transfer (<https://github.com/qdrant/qdrant/issues/3477>).

Sadly, we cannot use it directly due to backwards compatibility limits. Because of this we repurpose the current `PartialSnapshot` state, and will eventually deprecate it moving over to `Recovery`.

Here's a rough plan:
- Qdrant 1.8:
  - Change `PartialSnapshot` state:
    - Reject operations like before
    - But allow operations if a force flag is set
  - Add `Recovery` state
    - Only allow operations if a force flag is set
    - Cannot be used yet due to backwards compatibility
- Qdrant 1.9:
  - Deprecate `PartialSnapshot`, use `Recovery` everywhere

With the changes in 1.8, the partial snapshot and recovery state are essentially the same.

In 1.9, the recovery state will be used both for snapshot and WAL delta transfers. The partial snapshot state will be deprecated to limit the number of possible states.

Why don't we use Recovery state directly? We need a migration window. A cluster that updates nodes one by one cannot propose the recovery state in consensus, as old (unsupported) nodes will reject it.

Please feel free to suggest a name that is less generic than _recovery_, I couldn't come up with something better myself.

I've left some `// TODO(1.9):` comments to help point to the next steps to take when we'll be releasing 1.9.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/grpc/docs.md
+++ b/docs/grpc/docs.md
@@ -1493,6 +1493,7 @@ Note: 1kB = 1 vector of size 256. |
 | Initializing | 3 | Collection is being created |
 | Listener | 4 | A shard which receives data, but is not used for search; Useful for backup shards |
 | PartialSnapshot | 5 | Snapshot shard transfer is in progress; Updates should not be sent to (and are ignored by) the shard |
+| Recovery | 6 | Shard is undergoing recovered by an external node; Normally rejects updates, accepts updates if force is true |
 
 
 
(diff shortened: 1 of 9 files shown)


timvisee:
If we can remove the `PartialSnapshot` state in REST/gRPC specs of 1.9, we can simplify this more. It is only ever exposed in API responses, not in requests. If we can do that, we can simply keep the partial snapshot state but rename/alias it.

generall (approved):

PR #3660, opened 2024-02-21
Remove off-by-one logic in recovery point

timvisee:

Our recovery point was a copy of a click map, but with all clocks +1. This removes the +1, making a recovery point the same as the current clock map.

It should help reduce confusion regarding the values in the recovery point.

The existing test suite added as part of <https://github.com/qdrant/qdrant/issues/3477> proves that this change is behaving fine.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -103,18 +103,14 @@ impl ClockMap {
     /// Create a recovery point based on the current clock map state, so that we can recover any
     /// new operations with new clock values
     ///
-    /// The recovery point will contain every clock that is in this clock map, but with a tick of
-    /// one higher. That is because we already have an operation for the current clock tick, but
-    /// would like to receive the operation with the next tick on recovery.
-    ///
-    /// In other words, the recovery point will contain the first clock tick values the clock map
-    /// has not seen yet.
+    /// The recovery point contains every clock that is in this clock map. So, it represents all
+    /// the clock ticks we have.
     pub fn to_recovery_point(&self) -> RecoveryPoint {
         RecoveryPoint {
             clocks: self
                 .clocks
                 .iter()
-                .map(|(key, clock)| (*key, clock.current_tick + 1))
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -103,18 +103,14 @@ impl ClockMap {
...
-    /// has not seen yet.
+    /// The recovery point contains every clock that is in this clock map. So, it represents all
+    /// the clock ticks we have.
     pub fn to_recovery_point(&self) -> RecoveryPoint {
         RecoveryPoint {
             clocks: self
                 .clocks
                 .iter()
-                .map(|(key, clock)| (*key, clock.current_tick + 1))
+                .map(|(key, clock)| (*key, clock.current_tick))
```
timvisee, line 113: We don't do +1 here anymore.

lib/collection/src/wal_delta.rs
```diff
@@ -166,26 +158,32 @@ fn resolve_wal_delta(
...
-    if recovery_point.has_any_higher(local_recovery_point) {
+    if recovery_point.has_any_higher(&local_recovery_point) {
         return Err(WalDeltaError::HigherThanCurrent);
     }
 
+    // From this point, increase all clocks by one
+    // We must do that so we can specify clock tick 0 as needing everything from that clock
+    recovery_point.increase_by(1);
+    local_recovery_point.increase_by(1);
+    local_cutoff_point.increase_by(1);
```
timvisee, line 175: We now do +1 in a later stage of our WAL delta resolution logic. We need to do this so we can represent 0 as 'we need all clocks for this peer_id+clock_id, including tick 0'.

Without the +1 here, using 0 would mean we'd only get ticks 1 and up, missing tick 0.

Doing the +1 here is much more localized, and only internal to this function. That is a lot better than making the whole recovery point off-by-one by default like we had before.

generall (approved):

PR #3672, opened 2024-02-22
use readyz

generall:
This should prevent flakiness in some integration tests

Related experiment: https://github.com/qdrant/qdrant/pull/3668


timvisee (approved):
:tada:

PR #3674, opened 2024-02-23
Introduce JsonPathString

xzfc:
Part 1/3 of https://github.com/qdrant/qdrant/pull/3637#issuecomment-1951329675.

Roadmap/general idea: let the new and the old implementation coexist for a while. Introduce a new trait that both implementations will implement. Use the same set of unit tests for both implementations.

This PR adds a new trait `JsonPathInterface` and its implementation `JsonPathString` which behaves exactly like the old implementation.
All the code is updated to use the new interface. These are mostly mechanical changes in tests.

Additionally, this PR adds an empty stub implementation of `JsonPathV2` just to check that everything compiles if we replace the old implementation with the new one. It panics though.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/json_path/mod.rs
+++ b/lib/segment/src/json_path/mod.rs
@@ -14,24 +14,37 @@ pub use string::JsonPathString;
 pub type JsonPath = string::JsonPathString;
 
 pub trait JsonPathInterface: Sized + Clone + FromStr<Err = ()> + Display {
+    // Value operations
+
     fn value_get<'a>(&self, json_map: &'a serde_json::Map<String, Value>) -> MultiValue<&'a Value>;
+
     fn value_set<'a>(
         path: Option<&Self>,
         dest: &'a mut serde_json::Map<String, Value>,
         src: &'a serde_json::Map<String, Value>,
     );
+
     fn value_remove(&self, json_map: &mut serde_json::Map<String, Value>) -> MultiValue<Value>;
+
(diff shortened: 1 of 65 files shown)


lib/segment/src/json_path/mod.rs
timvisee, line 1: With a trait like this it seems a bit like over engineering, but I get that we only do this during the _migration period_. So then I'd be fine to keep it like this :+1:

lib/segment/src/json_path/string.rs
```diff
@@ -0,0 +1,119 @@
...
+        f.write_str(&self.0)
+    }
+}
+
+impl JsonPathInterface for JsonPathString {
+    fn value_get<'a>(&self, json_map: &'a serde_json::Map<String, Value>) -> MultiValue<&'a Value> {
+        #[allow(deprecated)]
+        crate::common::utils::get_value_from_json_map(&self.0, json_map)
+    }
+    fn value_set<'a>(
```
timvisee, line 29: Probably a nitpick, but please leave an empty line between functions or modules:

```suggestion
    }
    
    fn value_set<'a>(
```

timvisee (approved):

generall:
Please separate implementation and integration next time. 1000+ lines is still too big

generall (approved):

PR #3676, opened 2024-02-23
fix: Make thread id visible for shard loading thread

KShivendu:
Linux kernel [limits thread names to 15 chars](https://stackoverflow.com/questions/5026531/thread-name-longer-than-15-chars). 

This means thread that's loading `benchmark` collection is named as `shard-load-bench` (gets trimmed to 15 chars). This hides the thread id that we put in the custom thread name (useful for perf analysis tools). So it's (slightly) better to put the id before so it's visible to some extent.  This obviously can trim collection name. 

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -217,7 +217,7 @@ impl LocalShard {
             let segments_path = entry.unwrap().path();
             load_handlers.push(
                 thread::Builder::new()
-                    .name(format!("shard-load-{id}-{collection_id}"))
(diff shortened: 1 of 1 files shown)


agourlay:
> This hides the thread id that we put in the custom thread name (useful for perf analysis tools). So it's (slightly) better to put the id before so it's visible to some extent

The `id` here is the shard id which does not bring so much to the table.

The real information missing for the thread is the segment id which do not print because it is much longer.

I am not sure this PR brings much TBH.

timvisee:
I agree. I (also) prefer to keep it as-is.

`shard-load-*` is more valuable to me than `benchmark-*`

PR #3680, opened 2024-02-23
Drop unused Deserialize

xzfc:
A lot of structs has a `Deserialize` derive despite we never deserialize them. 

It is inconvenient for the following reasons:
- When adding new fields to structures, we need to provide validation or reason about "what if an user pass an unexpected data?". If a struct is not `Deserialize`, a developer could search and review all the places in the code base where this specific value originates. If it's `Deserialize`, this means that data could be arbitrary. E.g., my histogram bucket merging code works fine for data being generated by my code, but it's not true for arbitrary data.
- Double the work if we decide to add a custom `impl Serialize` for some field since then we have to implement a custom `Deserialize` too.

This PR drops unused `Deserialize` impls. Precisely, the following things removed:
- `#[derive(Deserialize)]`, obviously.
- `#[serde(default)]` for non-deserializable structs.
- `#[derive(Validate)]` and `#[validate]` in one place.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -9411,7 +9411,8 @@
       "OperationDurationStatistics": {
         "type": "object",
         "required": [
-          "count"
+          "count",
+          "total_duration_micros"
         ],
         "properties": {
           "count": {
@@ -9444,7 +9445,6 @@
           },
           "total_duration_micros": {
             "description": "The total duration of all operations in microseconds.",
-            "default": 0,
             "type": "integer",
             "format": "uint64",
             "minimum": 0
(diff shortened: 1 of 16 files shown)


docs/redoc/master/openapi.json
```diff
@@ -9411,7 +9411,8 @@
       "OperationDurationStatistics": {
         "type": "object",
         "required": [
-          "count"
+          "count",
+          "total_duration_micros"
```
xzfc, line 9415: It's output-only structure and this field is never skipped, so the new schema is more accurate.

generall (approved):

PR #3683, opened 2024-02-26
fix: Missing commit ID from dockerhub image builds

KShivendu:
`docker.io/qdrant/qdrant:{dev,latest}` images are built automatically by docker hub but they don't have commit ID. This PR fixes that.  

- `hooks/build` is a special file that Dockerhub looks for while building the image 
- `SOURCE_COMMIT` and `$IMAGE_NAME` are special env vars inside Dockerhub build env. 

[Reference](https://stackoverflow.com/questions/45277186/is-it-possible-to-add-environment-variables-in-automated-builds-in-docker-hub)

### All Submissions:

### DIFF
--- a/hooks/build
+++ b/hooks/build
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+docker build -t $IMAGE_NAME --build-arg GIT_COMMIT_ID=$SOURCE_COMMIT .
+


generall:
Not a big fan of extra top-level directories. Why GitHub registry is not enough?

KShivendu:
> Not a big fan of extra top-level directories

Same. But didn't have any other alternative. 

> Why GitHub registry is not enough?

We are including commit id for [dev](https://github.com/qdrant/qdrant/blob/dev/.github/workflows/dev-docker-image-build.yml#L29) and [release](https://github.com/qdrant/qdrant/blob/dev/.github/workflows/docker-image.yml#L40) build workflows on Github. It's better to be consistent everywhere.

PR #3688, opened 2024-02-26
Prototype `zero_increment_forward_proxy_inconcistency` test

ffuugoo:
Tracked in #3477.
Currently based on #3645, will be rebased on dev once #3645 is merged.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/wal_delta.rs
+++ b/lib/collection/src/wal_delta.rs
@@ -1098,6 +1098,7 @@ mod tests {
     #[case::eight_nodes(8, 1..8)]
     #[case::nine_nodes(9, 1..9)]
     #[case::ten_nodes(10, 1..10)]
+    #[tokio::test]
     async fn test_resolve_wal_delta_randomized(
         #[case] node_count: usize,
         #[case] dead_nodes_range: Range<usize>,
--- a/lib/collection/src/wal_delta.rs
+++ b/lib/collection/src/wal_delta.rs
@@ -87,14 +87,14 @@ impl RecoverableWal {
         // Lock highest and cutoff maps separately to avoid deadlocks
         {
             let mut highest_clocks = self.highest_clocks.lock().await;
-            for clock_tag in cutoff.clock_tag_iter() {
+            for clock_tag in cutoff.iter_as_clock_tags() {
                 highest_clocks.advance_clock(clock_tag);
             }
         }
 
         {
             let mut cutoff_clocks = self.cutoff_clocks.lock().await;
-            for clock_tag in cutoff.clock_tag_iter() {
+            for clock_tag in cutoff.iter_as_clock_tags() {
                 cutoff_clocks.advance_clock(clock_tag);
             }
         }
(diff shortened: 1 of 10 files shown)


lib/collection/src/wal_delta.rs
```diff
@@ -279,6 +280,119 @@ mod tests {
...
+        {
+            let mut operation = operation.clone();
+            let (_, _) = c_wal.lock_and_write(&mut operation).await.unwrap();
+            // we do not advance `a_clock_0` here
+        }
+
+        // Now, if we assume that:
+        // - we increment local clock map, when we receive an operation with `clock_tick = 0`
+        // - and we reject older operations
+        //   - (and we have to do so, otherwise we can't guarantee that our WAL has monotoncally increasing versions)
```
timvisee, line 353: > has monotoncally increasing versions

This is not the property we test. It can actually go down. We only want to make sure that the end of our WAL has the highest clocks.

lib/collection/src/wal_delta.rs
```diff
@@ -279,6 +280,119 @@ mod tests {
...
+        //
+        // Then the next operation will be *rejected* on node C, and C will be permanently missing
+        // `mock_operation(5)`.
```
timvisee, line 356: I don't think we care if a node is missing operations.

lib/collection/src/wal_delta.rs
```diff
@@ -279,6 +280,119 @@ mod tests {
...
+        for wal in [&a_wal, &b_wal] {
+            let mut operation = operation.clone();
+            let (_, _) = wal.lock_and_write(&mut operation).await.unwrap();
+            a_clock_0.advance_to(operation.clock_tag.unwrap().clock_tick);
+        }
+
+        // Then try to write operation on peer C
+        {
+            let mut operation = operation.clone();
+            // This should be rejected, but we currently accept all operations, so I can't assert this *yet* 🤷‍♀️
```
timvisee, line 375: Do we also want to reject operations with the same clock tick? Because if we do not, I don't think we have this problem. Then you'd just get operation 5 (or whichever it is) a second time. We do not have more than one forward proxy to a shard at a time.

timvisee (commented):
Some thoughts while going through the test:

lib/collection/src/wal_delta.rs
ffuugoo: That was what I meant, it was just annoying to formulate properly.

lib/collection/src/wal_delta.rs
ffuugoo: How so? That's an inconsistency between nodes. :/

lib/collection/src/wal_delta.rs
ffuugoo: If we do not reject current tick, then we'll just start having multiple operations with the same tick again. And in that case, what's the point of bothering with incrementing clock map in the first place then?

ffuugoo: Disregard that. I see that you might have a point here. In this case we *kinda* don't care about repetitions, and can just return the last operation with requred tick. Hmmmmm.......

lib/collection/src/wal_delta.rs
timvisee: > How so? That's an inconsistency between nodes. :/

That totally depends on what guarantees you want to have.

Imagine you send an operation from A to B and C, and C gets it twice. B's clock may be at 4, while C's clock may be at 5. If you send operation 6 normally now, B is missing operation 5. There is no data inconsistency.

lib/collection/src/wal_delta.rs
timvisee: Isn't that a compromise good enough? (rejecting lower, allowing the same)

Can we think of other scenarios where this incrementing may be problematic?

lib/collection/src/wal_delta.rs
ffuugoo: Basically, what happens here is node A and B have this WAL:
- clock: A0/0, operation: 1
- clock: A0/1, operation: 2
- clock: A0/2, operation: 3
- clock: A0/3, operation: 4
- clock: A0/4, operation: 5

But on node C the WAL is:
- clock: A0/0, operation: 1
- clock: A0/1, operation: 2
- clock: A0/2, operation: 3
- clock: A0/3, operation: 4
- clock: A0/4, operation: 4

So the node C is missing an operation *with clock tag 5*. It's missing the *unique operation 5*. That what I meant when I referred to `mock_operation(5)`.

ffuugoo:
I think this is also not relevant anymore

Issue #2326, opened 2023-07-25
Support batch requests to all Grouping APIs


keotic:
**Is your feature request related to a problem? Please describe.**
For production pipelines that require high throughput it is common to buffer and batch incoming requests before passing them
to the Model Encoding stage in the pipeline (i.e transforming Text / Image / Voice into a vector).
Therefore in production systems we usually need to handle parallel batches of queries. 

Currently all Grouping APIs are for single vectors, making these much appreciated and useful APIs a production bottleneck. 

**Describe the solution you'd like**
Add batching APIs for the Grouping APIs (and any new APIs in mind :))
https://qdrant.tech/documentation/concepts/search/#grouping-api

joein:
Was also asked in Discord https://discord.com/channels/907569970500743200/1134119750260236378

Snehil-Shah:
@agourlay @generall I want to work on this..
should I make an endpoint like `/collections/{collection_name}/points/search/groups/batch` ? Would really appreciate some initial guidance..

carlosb1:
Checking this... to be sure, I see  different cases:
- Improve API to permit queries with batches, to receive different types in the responses
- (from discord thread), modify from the current implementation to permit grouping queries with some  incorrect ID, then it should return only the correct ones.
Is it correct?

generall:
> (from discord thread), modify from the current implementation to permit grouping queries with some incorrect ID, then it should return only the correct ones.

this is a separate issue, it should be discussed separatelly

PR #3698, opened 2024-02-26
Add batching api for groups

acrrd:
The PR adds the batching API for groups. Closes: #2326 

This is still a draft to get some feedback on the direction of this PR.
The current change adds support for batching groups API.
GRPC, Openapi and tests are still missing.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/grouping/builder.rs
+++ b/lib/collection/src/grouping/builder.rs
@@ -18,7 +18,7 @@ where
     F: Fn(String) -> Fut + Clone,
     Fut: Future<Output = Option<RwLockReadGuard<'a, Collection>>>,
 {
-    group_by: GroupRequest,
+    group_by: Vec<GroupRequest>,
     collection: &'a Collection,
     /// `Fn` to get a collection having its name. Obligatory for recommend and lookup
     collection_by_name: F,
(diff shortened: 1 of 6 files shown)


acrrd:
@generall Do you think this is going in the right direction?

lib/collection/src/grouping/builder.rs
```diff
@@ -34,6 +34,21 @@ where
 {
     /// Creates a basic GroupBy builder
     pub fn new(group_by: GroupRequest, collection: &'a Collection, collection_by_name: F) -> Self {
```
acrrd, line 36: I left this signature as is to avoid changing too much in the tests.

lib/storage/src/content_manager/toc/point_ops.rs
```diff
@@ -176,7 +176,38 @@ impl TableOfContent {
         group_by
             .execute()
             .await
-            .map(|groups| GroupsResult { groups })
+            .map(|mut groups| GroupsResult {
+                groups: groups.pop().unwrap_or_default(),
```
acrrd, line 180: In this case, `execute` should always return one element. Would be preferable to return an error or use `unwrap()` instead of `unwrap_or_default()`?

lib/storage/src/content_manager/toc/point_ops.rs
```diff
@@ -176,7 +176,38 @@ impl TableOfContent {
         group_by
```
acrrd, line 176: We can rewrite `fn group` to use `fn group_batch`

PR #3700, opened 2024-02-27
Use WAL delta transfer for recovery if all nodes are 1.8+

timvisee:

A temporary hack to enable WAL delta transfer by default for shard recovery, only if all nodes are 1.8.

It probes all other peers to see if our whole cluster is 1.8. It probes up to 14 other nodes, otherwise it just assumes false. In later versions we'd simply rely on version information in consensus.

This would only be in version 1.8. In 1.9 it would be obsolete, in which case we'll remove it again.

We probably will not end up using this. I decided to push it anyway because I had already implemented it.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1254,6 +1254,7 @@ dependencies = [
  "pprof",
  "proptest",
  "rand 0.8.5",
+ "reqwest",
  "ringbuffer",
  "rmp-serde",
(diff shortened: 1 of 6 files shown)


timvisee:
We'll not be using this. Nor do I think this will be useful in 1.9.

PR #3701, opened 2024-02-27
Update Tokio 1.36

agourlay:
This [version](https://github.com/tokio-rs/tokio/releases/tag/tokio-1.36.0) has been released 25 days ago but was not picked up by Dependabot.

Looking at the past weeks, the weekly update quota was not reached so I do not understand why it was not proposed automatically.

Maybe the recent workspace setup has some mysterious side effects :man_shrugging:


timvisee:
> Maybe the recent workspace setup has some mysterious side effects 🤷‍♂️

Probably :(

Dependabot uses a custom dependency parser/resolver which, to be honest, is utter garbage.

timvisee (approved):

PR #3702, opened 2024-02-27
Add peer metadata to consensus, tracking Qdrant versions

timvisee:

Add peer metadata to consensus. Peers may propose new metadata once it becomes outdated.

At this time it holds the Qdrant version number. We may add more fields in the future.

Knowing the version number of all nodes is useful for feature gating. It may prevent us a lot of headaches in the future.

### Tasks
- [x] Disable updating metadata automatically until Qdrant 1.9

### All Submissions:

### New Feature Submissions

### Changes to Core Features:

### DIFF
--- a/lib/common/common/src/defaults.rs
+++ b/lib/common/common/src/defaults.rs
@@ -1,11 +1,12 @@
 use std::time::Duration;
 
 use lazy_static::lazy_static;
+use semver::Version;
 
 use crate::cpu;
 
 /// Current Qdrant version
-pub const QDRANT_VERSION: &str = "0.11.1";
+pub const QDRANT_VERSION: Version = Version::new(0, 11, 1);
 
 /// Default timeout for consensus meta operations.
 pub const CONSENSUS_META_OP_WAIT: Duration = Duration::from_secs(10);
--- a/lib/common/common/src/defaults.rs
+++ b/lib/common/common/src/defaults.rs
@@ -4,6 +4,9 @@ use lazy_static::lazy_static;
 
 use crate::cpu;
 
+/// Current Qdrant version
+pub const QDRANT_VERSION: &str = "1.7.4";
+
 /// Default timeout for consensus meta operations.
 pub const CONSENSUS_META_OP_WAIT: Duration = Duration::from_secs(10);
 
--- a/lib/common/common/src/defaults.rs
+++ b/lib/common/common/src/defaults.rs
@@ -5,7 +5,7 @@ use lazy_static::lazy_static;
 use crate::cpu;
 
 /// Current Qdrant version
-pub const QDRANT_VERSION: &str = "1.7.4";
+pub const QDRANT_VERSION: &str = "0.11.1";
 
 /// Default timeout for consensus meta operations.
 pub const CONSENSUS_META_OP_WAIT: Duration = Duration::from_secs(10);
--- a/lib/storage/src/content_manager/consensus/persistent.rs
+++ b/lib/storage/src/content_manager/consensus/persistent.rs
@@ -39,7 +39,7 @@ pub struct Persistent {
     /// Last known cluster topology
     #[serde(with = "serialize_peer_addresses")]
     pub peer_address_by_id: Arc<RwLock<PeerAddressById>>,
-    #[serde(with = "serialize_peer_metadata")]
+    #[serde(default, with = "serialize_peer_metadata")]
     pub peer_metadata_by_id: Arc<RwLock<PeerMetadataById>>,
     pub this_peer_id: PeerId,
     #[serde(skip)]
--- a/lib/storage/src/content_manager/consensus/persistent.rs
+++ b/lib/storage/src/content_manager/consensus/persistent.rs
@@ -39,7 +39,7 @@ pub struct Persistent {
     /// Last known cluster topology
     #[serde(with = "serialize_peer_addresses")]
     pub peer_address_by_id: Arc<RwLock<PeerAddressById>>,
-    #[serde(default, with = "serialize_peer_metadata")]
+    #[serde(default)]
     pub peer_metadata_by_id: Arc<RwLock<PeerMetadataById>>,
     pub this_peer_id: PeerId,
     #[serde(skip)]
@@ -307,34 +307,6 @@ mod serialize_peer_addresses {
     {
         let addresses: HashMap<u64, Uri> = serialize_peer_addresses::deserialize(deserializer)?;
         Ok(Arc::new(RwLock::new(addresses)))
-    }
-}
-
-mod serialize_peer_metadata {
-    use std::collections::HashMap;
-    use std::sync::Arc;
-
-    use parking_lot::RwLock;
-    use serde::{self, Deserialize, Deserializer, Serialize, Serializer};
-
-    use crate::types::{PeerMetadata, PeerMetadataById};
-
-    pub fn serialize<S>(
-        metadata: &Arc<RwLock<PeerMetadataById>>,
-        serializer: S,
-    ) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        metadata.read().serialize(serializer)
-    }
-
-    pub fn deserialize<'de, D>(deserializer: D) -> Result<Arc<RwLock<PeerMetadataById>>, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        let metadata: HashMap<u64, PeerMetadata> = PeerMetadataById::deserialize(deserializer)?;
-        Ok(Arc::new(RwLock::new(metadata)))
     }
 }
 
--- a/lib/storage/src/content_manager/consensus/persistent.rs
+++ b/lib/storage/src/content_manager/consensus/persistent.rs
@@ -320,13 +320,13 @@ mod serialize_peer_metadata {
     use crate::types::{PeerMetadata, PeerMetadataById};
 
     pub fn serialize<S>(
-        versions: &Arc<RwLock<PeerMetadataById>>,
+        metadata: &Arc<RwLock<PeerMetadataById>>,
         serializer: S,
     ) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
     {
-        versions.read().serialize(serializer)
+        metadata.read().serialize(serializer)
     }
 
     pub fn deserialize<'de, D>(deserializer: D) -> Result<Arc<RwLock<PeerMetadataById>>, D::Error>
--- a/lib/storage/src/content_manager/consensus/persistent.rs
+++ b/lib/storage/src/content_manager/consensus/persistent.rs
@@ -189,7 +189,7 @@ impl Persistent {
         self.peer_metadata_by_id
             .read()
             .get(&self.this_peer_id())
-            .map_or(true, |metadata| metadata.is_outdated())
+            .map_or(true, |metadata| metadata.is_different_version())
     }
 
     pub fn this_peer_id(&self) -> PeerId {
--- a/lib/storage/src/content_manager/consensus/persistent.rs
+++ b/lib/storage/src/content_manager/consensus/persistent.rs
@@ -14,7 +14,7 @@ use raft::RaftState;
 use serde::{Deserialize, Serialize};
 
 use crate::content_manager::consensus::entry_queue::{EntryApplyProgressQueue, EntryId};
-use crate::types::PeerAddressById;
+use crate::types::{PeerAddressById, PeerMetadata, PeerMetadataById};
 use crate::StorageError;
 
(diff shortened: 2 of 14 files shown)


lib/storage/src/types.rs
```diff
@@ -264,3 +266,25 @@ impl Anonymize for ClusterStatus {
         }
     }
 }
+
+/// Metadata describing extra properties for each peer
+#[derive(Debug, Hash, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct PeerMetadata {
+    /// Peer Qdrant version string
+    version: Option<String>,
+}
```
timvisee, line 275: Do we also want to keep metadata here for unknown fields, potentially from newer versions?

lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -703,6 +717,22 @@ impl<C: CollectionContainer> ConsensusManager<C> {
...
+                .send(ConsensusOperations::UpdatePeerMetadata {
+                    peer_id: self.this_peer_id(),
+                    metadata: PeerMetadata::current(),
+                });
+            if let Err(err) = result {
+                log::error!(
+                    "Failed to propose consensus peer metadata update for this peer: {err}"
+                );
+            }
+        }
```
timvisee, line 734: For compatibility reasons we must disable this logic before merging, until we release 1.9.

Having this here in 1.8 will break clusters that are still running 1.7 nodes, so we need a migration period of at least 1 minor version.

lib/common/common/src/defaults.rs
```diff
@@ -4,6 +4,9 @@ use lazy_static::lazy_static;
 
 use crate::cpu;
 
+/// Current Qdrant version
+pub const QDRANT_VERSION: &str = "1.7.4";
```
generall, line 8: Is it now another place to remember about during version upgrades?

timvisee: Yes. We cannot access the true version from code where we need it.

I do have a trick up my sleeve though: we can assert it in `src/main.rs` to make sure it matches up with the version of the main crate. How do you feel about that?

ffuugoo: We can make it a global static and set it from the main crate at startup.

E.g.:

```rust
// `defaults` crate
pub static QDRANT_VERSION: Mutex<String> = Mutex::new(String::new());

// `qdrant` crate
fn main() {
    *defaults::QDRANT_VERSION.lock().unwrap() =  env!("CARGO_PKG_VERSION").into();
}
```

Or, better yet, make it three `AtomicU32`s:

```rust
// `defaults` crate
pub static QDRANT_VERSION_MAJOR: AtomicU32 = AtomicU32::new(0);
pub static QDRANT_VERSION_MINOR: AtomicU32 = AtomicU32::new(0);
pub static QDRANT_VERSION_PATCH: AtomicU32 = AtomicU32::new(0);

// `qdrant` crate
fn main() {
    defaults::QDRANT_VERSION_MAJOR.store(env!("CARGO_PKG_VERSION_MAJOR").parse().unwrap());
    defaults::QDRANT_VERSION_MINOR.store(env!("CARGO_PKG_VERSION_MINOR").parse().unwrap());
    defaults::QDRANT_VERSION_PATCH.store(env!("CARGO_PKG_VERSION_PATCH").parse().unwrap());
}
```

lib/storage/src/content_manager/consensus/persistent.rs
```diff
@@ -275,6 +310,34 @@ mod serialize_peer_addresses {
...
+    }
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<Arc<RwLock<PeerMetadataById>>, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let metadata: HashMap<u64, PeerMetadata> = PeerMetadataById::deserialize(deserializer)?;
+        Ok(Arc::new(RwLock::new(metadata)))
+    }
+}
```
ffuugoo, line 339: There are `serde/rc` and `parking_lot/serde` features. Or these don't work somehow? 🤔

(Didn't check `serialize_peer_addresses`, but maybe we can get rid of it as well, if it's similar to `serialize_peer_metadata`.)

lib/storage/src/content_manager/consensus/persistent.rs
```diff
@@ -150,6 +154,25 @@ impl Persistent {
...
+        metadata: PeerMetadata,
+    ) -> Result<(), StorageError> {
+        if let Some(prev_metadata) = self
+            .peer_metadata_by_id
+            .write()
+            .insert(peer_id, metadata.clone())
+        {
+            log::warn!(
+                "Replaced metadata of peer {peer_id} from {prev_metadata:?} to {metadata:?}"
+            );
```
ffuugoo, line 169: `log::info` maybe? I guess metadata won't be updated _too_ often (as we will only use it for Qdrant version; _for now_), but `log::warn` just seems too verbose for a pretty regular, expected condition.

lib/storage/src/content_manager/toc/mod.rs
```diff
@@ -114,10 +114,7 @@ impl TableOfContent {
                 .path();
 
             if !CollectionConfig::check(&collection_path) {
-                log::warn!(
-                    "Collection config is not found in the collection directory: {:?}, skipping",
-                    collection_path
-                );
+                log::warn!("Collection config is not found in the collection directory: {collection_path:?}, skipping");
```
ffuugoo, line 117: 🤔

```suggestion
                log::warn!(
                    "Collection config is not found in the collection directory: {}, skipping",
                    collection_path.display()
                 );
```

lib/storage/src/types.rs
```diff
@@ -264,3 +266,25 @@ impl Anonymize for ClusterStatus {
         }
     }
 }
+
+/// Metadata describing extra properties for each peer
+#[derive(Debug, Hash, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct PeerMetadata {
+    /// Peer Qdrant version string
+    version: Option<String>,
```
ffuugoo, line 274: Should we make it three integers instead of a `String`? Or introduce `struct Version { major: u32, minor: u32, patch: u32 }`, maybe?

lib/common/common/src/defaults.rs
timvisee: I don't like that at all. It suddenly makes our sub crate dependent on our main one to work properly.

I'll quickly add an assertion in `src/main.rs` to achieve the same effect.

lib/storage/src/content_manager/consensus/persistent.rs
timvisee: > There are serde/rc and parking_lot/serde features. Or these don't work somehow? 🤔

I copied the logic for URLs. I'll check to see if I can reduce it with predefined serde implementations.

lib/storage/src/content_manager/consensus/persistent.rs
timvisee: Better, I agree.

lib/storage/src/content_manager/consensus/persistent.rs
timvisee: Adding the `serde` feature to `parking_lot` doesn't work sadly.

<details>
<summary>Cargo errors</summary>

```rust
 1  error[E0277]: the trait bound `std::sync::Arc<parking_lot::lock_api::RwLock<parking_lot::RawRwLock, s
 td::collections::HashMap<u64, types::PeerMetadata>>>: types::_::_serde::Serialize` is not satisfied
     --> lib/storage/src/content_manager/consensus/persistent.rs:27:17
     --> /home/timvisee/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.197/src/ser/mod.rs
 :1865:12
 2  error[E0277]: the trait bound `std::sync::Arc<parking_lot::lock_api::RwLock<parking_lot::RawRwLock, s
 td::collections::HashMap<u64, types::PeerMetadata>>>: types::_::_serde::Deserialize<'_>` is not satisfie
 d
     --> lib/storage/src/content_manager/consensus/persistent.rs:43:30
     --> /home/timvisee/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.197/src/de/mod.rs:
 1726:12
 3  error[E0277]: the trait bound `std::sync::Arc<parking_lot::lock_api::RwLock<parking_lot::RawRwLock, s
 td::collections::HashMap<u64, types::PeerMetadata>>>: types::_::_serde::Deserialize<'_>` is not satisfie
 d
     --> lib/storage/src/content_manager/consensus/persistent.rs:43:30
     --> /home/timvisee/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.197/src/de/mod.rs:
 1865:12
```

</details>

lib/common/common/src/defaults.rs
ffuugoo: >  It suddenly makes our sub crate dependent on our main one to work properly.

Kinda, but not really?

Like, we do set the values from `qdrant` crate, sure. But the variables are still accessible, they are just zeroes or empty strings by default... which I don't think is critical?

E.g., `collection` (and other `lib` crates) _don't_ run by itself, on it's own. We only run then as part of `qdrant` crate. The only exception, is when we run _tests_.

But let's assume we go with `const QDRANT_VERSION: &str = "1.2.3"`. How will you be able to use it in tests? You literally can't in any meaningful way _assert_ on `QDRANT_VERSION` value (except, _maybe_, that it's not an empty string). Otherwise, every time you release, you'd have to go and modify both `QDRANT_VERSION` *and* all tests that rely on `QDRANT_VERSION` *value*.

But if we do `static QDRANT_VERSION: Mutex<String> = ...`, then if you *want* to rely on it in tests, you can just *set* the value inside the test.

And if you never _check_ `QDRANT_VERSION` value inside your tests... than what the difference? It will be an empty string, sure, but, just run your tests with an empty string, it'll probably still work.

We pretty much already do this. We do the whole "dependency inversion" thing in a bunch of places, when we define implementation in `qdrant` crate and propagate it into `collection` crate. It's kinda like this, I think?

lib/storage/src/types.rs
ffuugoo: I'm not like super against it, but but how will you implement feature gates then?

Let's say we merged it, and you want to feature-gate something, only if all nodes are at least 1.9.0. We released 1.10.0, how do you implement the comparison? You can't simply compare *strings*, cause "1.10.0" < "1.9.0"

You'll *have to* use `semver::Version` and parse the string every time? :/

lib/storage/src/content_manager/consensus/persistent.rs
ffuugoo: You may have to also enable `rc` feature of `serde` crate?

Sorry if you've already tried it, I'm just totally *not* reading this logs. :D

Both `Arc` *and* `parking_lot` implement serde under a feature, so you have to enable *both* of them.

lib/storage/src/types.rs
ffuugoo: I'd better use `semver::Version` straight away, cause I don't think that string version is all that useful.

lib/storage/src/content_manager/consensus/persistent.rs
timvisee: Missed `rc`. Works now. Thanks!

timvisee: We cannot do it on the old one though. `Uri` does not implement serde traits. 

There is [`http_serde`](https://docs.rs/http-serde/latest/http_serde/), but we cannot use them because we've wrapped the types. So I'll leave the existing module in place.

src/main.rs
```diff
@@ -113,6 +114,12 @@ struct Args {
 }
 
 fn main() -> anyhow::Result<()> {
+    debug_assert_eq!(
+        defaults::QDRANT_VERSION,
+        env!("CARGO_PKG_VERSION"),
+        "crate version does not match with defaults.rs",
+    );
```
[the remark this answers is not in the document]
timvisee: I'll add a test.

A static assertion is what I tried first, but we cannot use it in this case. :(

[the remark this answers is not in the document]
timvisee: I've now added this assertion in `build.rs`.

That way it is checked at compile time, for regular builds and tests. :+1:

ffuugoo (approved):

lib/storage/src/content_manager/consensus_manager.rs
timvisee: I've commented it out now so it is mergeable.

ffuugoo (approved):

generall:
I am trying to simulate migration from previous version and my `peer_metadata_by_id` doesn't get populated even after several re-starts

generall (approved):

PR #3703, opened 2024-02-27
Add consensus operation to restart shard transfer

timvisee:

Add a consensus operation to restart a shard transfer with a different configuration.

This will make falling back to a different shard transfer method a whole lot nicer. We make sure to properly clean up and prepare for the different shard transfer method. Without it, we're manually managing and update shard replica set states to make transfers happen which is very fragile.

The operation ensures that:
- a transfer is currently ongoing
- the transfer configuration has changed (likely a different transfer method)

It is similar to calling abort/start right after each other, but in a single consensus operation.

If we'll merge this in 1.8, we'll be able to make use of it in Qdrant 1.9.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/toc/collection_meta_ops.rs
+++ b/lib/storage/src/content_manager/toc/collection_meta_ops.rs
@@ -342,11 +342,18 @@ impl TableOfContent {
                     .await?;
             }
             ShardTransferOperations::Restart(transfer) => {
-                // Validate transfer exists
-                transfer::helpers::validate_transfer_exists(
-                    &transfer.key(),
-                    &collection.state().await.transfers,
-                )?;
+                let transfers = collection.state().await.transfers;
+
+                // Transfer must exist
+                transfer::helpers::validate_transfer_exists(&transfer.key(), &transfers)?;
+
+                // Transfer must have changed configuration
+                if transfers.contains(&transfer) {
+                    return Err(StorageError::bad_request(format!(
+                        "Cannot restart transfer for shard {} from {} to {}, its configuration did not change",
+                        transfer.shard_id, transfer.from, transfer.to,
+                    )));
+                }
 
                 // Abort and start transfer
                 self.handle_transfer(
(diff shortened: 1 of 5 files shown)


lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -339,6 +341,32 @@ impl TableOfContent {
...
+                self.handle_transfer(
+                    collection_id.clone(),
+                    ShardTransferOperations::Abort {
+                        transfer: transfer.key(),
+                        reason: "restart transfer".into(),
+                    },
+                )
+                .await?;
+                self.handle_transfer(collection_id, ShardTransferOperations::Start(transfer))
+                    .await?;
```
timvisee, line 368: Here I just call abort and start right after each other, as part of a single _operation_. It keeps the implementation very simple.

Maybe there's more to it though.

ffuugoo (approved):
Looks simple enough. LGTM.

generall (approved):

lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -364,7 +372,12 @@ impl TableOfContent {
                     },
                 )
                 .await?;
-                self.handle_transfer(collection_id, ShardTransferOperations::Start(transfer))
+
+                let mut new_transfer = transfer.clone();
+                // Preserve sync flag from the old transfer
+                new_transfer.sync = old_transfer.sync;
+
```
generall, line 379: This is new stuff from my side, might require extra look

timvisee: Shall we do an OR, to sync if the old OR the new one has sync=true?

```suggestion
                // Preserve sync flag from the old transfer
                let mut new_transfer = transfer.clone();
                new_transfer.sync |= old_transfer.sync;

```

timvisee: I've accidentally hit the green button.

I'm handling this here: https://github.com/qdrant/qdrant/pull/3728

Issue #3627, opened 2024-02-15
Flaky test openapi_integration/test_snapshot


agourlay:
Flaky test observed several times on CI.

```
collection_name = 'test_collection_snapshot'

    def drop_snapshots(collection_name: str) -> None:
        """Delete all snapshots on the collection."""
        response = request_with_validation(
            api='/collections/{collection_name}/snapshots',
            method="GET",
            path_params={'collection_name': collection_name},
        )
        assert response.ok
        for snapshot in response.json()['result']:
            response = request_with_validation(
                api='/collections/{collection_name}/snapshots/{snapshot_name}',
                method="DELETE",
                path_params={'collection_name': collection_name,
                             'snapshot_name': snapshot['name']},
                query_params={'wait': 'true'},
            )
>           assert response.ok
E           assert False
E            +  where False = <Response [404]>.ok

openapi_integration/test_snapshot.py:35: AssertionError
```

e.g. https://github.com/qdrant/qdrant/actions/runs/7917662351/job/21614243565?pr=3605#step:10:11351

generall:
Another example: https://github.com/qdrant/qdrant/actions/runs/7923808268/job/21634275041?pr=3610

agourlay:
Still present https://github.com/qdrant/qdrant/actions/runs/8067272697/job/22037287976

Seems to only reproduce on CI according to my investigation.

timvisee:
<https://github.com/qdrant/qdrant/pull/3704> fixes it, which I've tested with 200 CI runs. It's now running another time with 500 runs to hopefully eliminate this once and for good.

timvisee:
Fixed in <https://github.com/qdrant/qdrant/pull/3704>. (hopefully :smile: )

PR #3704, opened 2024-02-27
Fix flaky invalid URI snapshot test

timvisee:
Fixes: <https://github.com/qdrant/qdrant/issues/3627>

Tested in <https://github.com/qdrant/qdrant/pull/3707> running it 100s of times: https://github.com/qdrant/qdrant/actions/runs/8078479540/job/22070892084?pr=3707

Our snapshot tests did not clean up snapshot files properly.
### All Submissions:

### New Feature Submissions:

### DIFF
--- a/tests/openapi/openapi_integration/test_snapshot.py
+++ b/tests/openapi/openapi_integration/test_snapshot.py
@@ -361,3 +361,6 @@ def test_snapshot_invalid_file_uri():
     )
     assert response.status_code == 400
     assert response.json()["status"]["error"] == "Bad request: Snapshot file \"/whatever.snapshot\" does not exist"
+
+    # Give file system some time to clean up temporary snapshot files
+    sleep(1)
--- a/tests/openapi/openapi_integration/test_snapshot.py
+++ b/tests/openapi/openapi_integration/test_snapshot.py
@@ -324,9 +324,6 @@ def test_snapshot_operations_non_wait():
             sleep(0.1)
             continue
 
-    # Give file system some time to clean up temporary snapshot files
-    sleep(1)
-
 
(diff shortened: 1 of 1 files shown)


agourlay:
I have troubling wrapping my head around the fix :thinking: 

The test `test_snapshot_invalid_file_uri` fails during the setup phase before yielding to the test body.
Why does cleaning at the end of the test fixes the issue>

tests/openapi/openapi_integration/test_snapshot.py
```diff
@@ -361,3 +361,6 @@ def test_snapshot_invalid_file_uri():
     )
     assert response.status_code == 400
     assert response.json()["status"]["error"] == "Bad request: Snapshot file \"/whatever.snapshot\" does not exist"
+
+    # Give file system some time to clean up temporary snapshot files
+    sleep(1)
```
agourlay, line 366: I am having one last issue with this fix :sweat_smile: 

The `test_snapshot_invalid_file_uri` test does not create any valid snapshot as far as I can tell.
Therefore it is slightly weird having to wait for cleaning the temporary snapshot files at the end.

Makes sense?

timvisee: You're right, it is very confusing.

After another look, it's not `test_snapshot_invalid_file_uri` that creates files and fails, but other snapshot tests before it. But before starting `test_snapshot_invalid_file_uri` we list/delete these same snapshot files.

I'll update the fix putting stuff in the right place so that it makes more sense.

Thanks for the careful consideration!

generall (approved):

agourlay (approved):

PR #3712, opened 2024-02-28
Clearer names for dense storage implementations

agourlay:
This PR is a bunch of mechanical renaming for dense storage related types and files.

The motivation is that we want to add soon new types of storages for multi-vector points, so it is better to makes things explicit.

### DIFF
--- a/lib/segment/src/segment.rs
+++ b/lib/segment/src/segment.rs
@@ -108,7 +108,7 @@ impl VectorData {
         };
 
         let storage_task = match &*self.vector_storage.borrow() {
-            VectorStorageEnum::Memmap(storage) => storage.prefault_mmap_pages(),
+            VectorStorageEnum::DenseMemmap(storage) => storage.prefault_mmap_pages(),
             _ => None,
         };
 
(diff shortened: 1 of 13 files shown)


timvisee (commented):
Just quickly skimmed it: are we sure this doesn't break compatibility, in terms of loading existing storage from disk?

agourlay:
> Just quickly skimmed it: are we sure this doesn't break compatibility, in terms of loading existing storage from disk?

This part is tested by our storage compatibility tests.
Only types are changes, no field names, I do not foresee any compatibility issues.

Do you maybe a specific scenario in mind?

timvisee:
> This part is tested by our storage compatibility tests. Only types are changes, no field names, I do not foresee any compatibility issues.

You're right. Brilliant! Thanks for confirming.

> Do you maybe a specific scenario in mind?

Nope! Not anything other than what you've already covered.

agourlay:
I just double checked, it seems the compatibility tests are **not** covering the appendable memap configuration.
Will check how to add it quickly.

generall:
I think it is safe to merge, cause those structs do not implement Deserialize anyway

agourlay:
@generall excellent observation :+1: 

Then feel free to approve, I will work on improving the compat. tests independently from this PR.

generall (approved):

PR #3717, opened 2024-02-28
Enforce that storage compatibility validates master data on dev

agourlay:
So far our strategy for testing storage compatibility has been to refresh from time to time a set of archives containing data and snapshots and making sure that the current `dev` branch understand those.

This PR proposes to make the contract clearer by only pulling the archives from the master branch during tests.

This makes sure we have a stable point of reference.


generall (approved):

coszio (approved):
I like this approach

PR #3719, opened 2024-02-28
Logic tweaks for WAL diff shard transfer

ffuugoo:

This PR tweaks WAL diff transfer related logic:
- [x] `ClockMap` should reject operations with `clock_tick = 0`
  - [x] update `ClockMap` unit-tests
- [x] `LocalShard::update` should honor `ClockMap::advance_*` result
  - [x] accept or *reject* operations
  - [x] and propagate updated `clock_tick` to the caller
  - [x] extend `UpdateResult` type, so that rejection can be propagated more explicitly?
  - [x] refactor `SerdeWal::lock_and_write`, so it will be more ergonomic to work with?
- [x] if *any* replica rejected the operation, `ShardReplicaSet::update` should retry the operation with new tick (on *all* replicas)
- [x] update `wal_delta` tests

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/local_shard/shard_ops.rs
+++ b/lib/collection/src/shards/local_shard/shard_ops.rs
@@ -254,9 +254,10 @@ impl ShardOperation for LocalShard {
                 Ok(id_and_lock) => id_and_lock,
 
                 Err(crate::wal::WalError::Rejected) => {
+                    // Propagate clock rejection to operation sender
                     return Ok(UpdateResult {
                         operation_id: None,
-                        status: UpdateStatus::Acknowledged,
+                        status: UpdateStatus::ClockRejected,
                         clock_tag: operation.clock_tag,
                     });
                 }
--- a/lib/collection/src/shards/local_shard/shard_ops.rs
+++ b/lib/collection/src/shards/local_shard/shard_ops.rs
@@ -253,7 +253,7 @@ impl ShardOperation for LocalShard {
             let (operation_id, _wal_lock) = match self.wal.lock_and_write(&mut operation).await {
                 Ok(id_and_lock) => id_and_lock,
 
-                Err(crate::wal::WalError::Rejected) => {
+                Err(crate::wal::WalError::ClockRejected) => {
                     // Propagate clock rejection to operation sender
                     return Ok(UpdateResult {
                         operation_id: None,
--- a/lib/collection/src/shards/local_shard/shard_ops.rs
+++ b/lib/collection/src/shards/local_shard/shard_ops.rs
@@ -247,13 +247,22 @@ impl ShardOperation for LocalShard {
             let update_sender = self.update_sender.load();
             let channel_permit = update_sender.reserve().await?;
 
-            // TODO:
-            //
-            // - Apply operation, only if it's accepted and written into the WAL...
-            // - ...or *propagate rejection to the sender node*, if it's rejected!
-
             // It is *critical* to hold `_wal_lock` while sending operation to the update handler!
-            let (operation_id, _wal_lock) = self.wal.lock_and_write(&mut operation).await?;
+            //
+            // TODO: Refactor `lock_and_write`, so this is less terrible? :/
+            let (operation_id, _wal_lock) = match self.wal.lock_and_write(&mut operation).await {
+                Ok(id_and_lock) => id_and_lock,
+
+                Err(crate::wal::WalError::Rejected) => {
+                    return Ok(UpdateResult {
+                        operation_id: None,
+                        status: UpdateStatus::Acknowledged,
+                        clock_tag: operation.clock_tag,
+                    });
+                }
+
+                Err(err) => return Err(err.into()),
+            };
 
             channel_permit.send(UpdateSignal::Operation(OperationData {
                 op_num: operation_id,
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -136,6 +136,25 @@ impl ShardReplicaSet {
         operation: CollectionUpdateOperations,
         wait: bool,
     ) -> CollectionResult<UpdateResult> {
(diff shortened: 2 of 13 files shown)


lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -247,7 +247,20 @@ impl ShardOperation for LocalShard {
...
+            // TODO: Refactor `lock_and_write`, so this is less terrible? :/
+            let (operation_id, wal_lock) = match self.wal.lock_and_write(&mut operation).await {
+                Ok(id_and_lock) => id_and_lock,
+
+                Err(crate::wal::WalError::Rejected) => {
+                    return Ok(UpdateResult {
+                        operation_id: None,
+                        status: UpdateStatus::Acknowledged,
+                        clock_tag: operation.clock_tag,
+                    });
```
ffuugoo, line 259: So, this is the "hack". I just use `operation_id: None` and `status: Acknowledged` and updated clock there. I think it's "good enough" to implement the retry code for now.

generall: Can't we introduce new `UpdateStatus`? It should be backward compatible, I think

ffuugoo: I think we can, just didn't want to bother with it _yet_.

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -136,6 +136,25 @@ impl ShardReplicaSet {
...
     ) -> CollectionResult<UpdateResult> {
+        // `ShardRepilcaSet::update_impl` is not cancel safe, so this method is not cancel safe.
+
+        loop {
+            if let Some(res) = self.update_impl(operation.clone(), wait).await? {
+                return Ok(res);
+            }
+
+            log::warn!("Operation {operation:?} was rejected by some node(s), retrying...");
+        }
```
ffuugoo, line 147: Some notes:
- I just realized I did not implement it exactly right: I release the clock between `update_impl` calls. Will fix this in a moment. 😂
- This spins and retries the operation infinitely, we probably want to limit the number of retries.
- The whole `update` is marked as not cancel safe, even though we should be safe to cancel "between" the calls to `update_impl`...
  - ...but I'm not sure if we can easily make it "cancellable between `update_impl` calls", so it might still spin in the background for a while, even if client drops the request...
  - ...which should be OK, once we limit the number of retries

ffuugoo: So, the question is: how many retries do we want? @generall @timvisee

ffuugoo:
Still WIP, but pinged you guys for review, to check if the PR seems reasonable so far, or you have objections already.

lib/collection/src/shards/local_shard/shard_ops.rs
timvisee: I've added this in <https://github.com/qdrant/qdrant/pull/3725>.

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -136,13 +136,44 @@ impl ShardReplicaSet {
...
+            let res = self
+                .update_impl(operation.clone(), wait, &mut clock)
+                .await?;
+
+            if let Some(res) = res {
+                return Ok(res);
+            }
+
+            log::warn!("Operation {operation:?} was rejected by some node(s), retrying...");
+        }
```
timvisee, line 159: Do we still need to update our own clock to the newest we've seen?

If it's done somewhere else, I've missed it.

And obviously we want to limit retries here.

ffuugoo: > If it's done somewhere else, I've missed it.

It's done in `update_impl`.

timvisee: Of course, it's just our old logic of advancing with all tags we receive. Thanks!

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -241,12 +293,18 @@ impl ShardReplicaSet {
                 let echo_tag = result.clock_tag?;
 
                 if echo_tag.peer_id != clock_tag.peer_id {
-                    // TODO: `log::warn`!?
+                    debug_assert!(
+                        false,
+                        "Echoed clock tag peer_id does not match the original"
+                    );
```
ffuugoo, line 299:
```suggestion
                    #[cfg(debug_assertions)]
                    panic!("Echoed clock tag peer_id does not match the original");
```

Not sure if this is better, though.

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -241,12 +293,18 @@ impl ShardReplicaSet {
...
                     return None;
                 }
 
                 if echo_tag.clock_id != clock_tag.clock_id {
-                    // TODO: `log::warn`!?
+                    debug_assert!(
+                        false,
+                        "Echoed clock tag clock_id does not match the original"
+                    );
```
ffuugoo, line 307: Ditto.

generall:
`test_shard_wal_delta_transfer_manual_recovery_chain` is not stable with this change (even locally)

timvisee:
> `test_shard_wal_delta_transfer_manual_recovery_chain` is not stable with this change (even locally)

Investigating

lib/collection/src/shards/replica_set/update.rs
ffuugoo: Nevermind. With `panic` it starts complaining about `return None` being unreachable. :/

timvisee:
> `test_shard_wal_delta_transfer_manual_recovery_chain` is not stable with this change (even locally)

These errors do not seem to be related to the changes here. I've also seen it happen on `dev`. I'll resolve the consensus test on dev directly.

timvisee (approved):

PR #3720, opened 2024-02-29
Introduce MultiDenseVector to models

agourlay:
This PR adds a new variant `MultiDenseVector` to our models to bootstrap the work on ColBERT embeddings.


agourlay:
The internal changes is visible in the OpenAPI because we expose our internal segment types to the users.

To do this cleanly we should first introduce a layer of abstraction (DTO) to enable us to move forward.

PR #3723, opened 2024-02-29
Add WAL delta consensus test for aborting and retrying diff transfer

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/3790>

Add a consensus test that confirms that aborting and restarting a WAL delta transfer works.

This consensus test:
- Measures WAL delta recovery baseline time
- Kills the last node
- Starts WAL delta recovery, but aborts it in the middle
- Retries WAL delta recovery
- Asserts that all data is still intact

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/tests/consensus_tests/test_shard_wal_delta_transfer.py
+++ b/tests/consensus_tests/test_shard_wal_delta_transfer.py
@@ -354,7 +354,7 @@ def test_shard_wal_delta_transfer_manual_recovery_chain(tmp_path: pathlib.Path):
 #
 # Test that data on the both sides is consistent even if replication was
 # aborted.
-def test_shard_wal_delta_transfer_abort_and_retry(tmp_path: pathlib.Path, capfd):
+def test_shard_wal_delta_transfer_abort_and_retry(tmp_path: pathlib.Path):
     assert_project_root()
 
     # Prevent automatic recovery on restarted node, so we can manually recover with a specific transfer method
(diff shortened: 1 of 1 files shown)


tests/consensus_tests/test_shard_wal_delta_transfer.py
```diff
@@ -207,6 +207,161 @@ def test_shard_wal_delta_transfer_manual_recovery(tmp_path: pathlib.Path, capfd)
...
+                "shard_id": 0,
+                "from_peer_id": from_peer_id,
+                "to_peer_id": to_peer_id,
+                "method": "wal_delta",
+            }
+        })
+    assert_http_ok(r)
+
+    # But this time abort the transfer in the middle
+    sleep(baseline / 2)
```
generall, line 299: seems to be very fragile on CI

timvisee: Any suggestion?

I don't see another way versus a fixed timeout, which is probably worse.

tests/consensus_tests/test_shard_wal_delta_transfer.py
```diff
@@ -207,6 +207,161 @@ def test_shard_wal_delta_transfer_manual_recovery(tmp_path: pathlib.Path, capfd)
...
+                "from_peer_id": from_peer_id,
+                "to_peer_id": to_peer_id,
+            }
+        })
+    assert_http_ok(r)
+
+    sleep(1)
+
+    # Confirm WAL delta transfer is aborted based on stdout logs
+    stdout, _stderr = capfd.readouterr()
```
agourlay, line 313: Asserting on application logs is giving me bad vibes TBH

Is it strictly necessary?

agourlay: I did not know we already used this before.

My worry is that we are checking some internal state that should not be visible.

> What about this: leaving it in, but if it becomes a problem we remove the extra checks?

I am Ok. with this, I certainly do not want to slow you down here :)

tests/consensus_tests/test_shard_wal_delta_transfer.py
xzfc: Maybe we could wait till a specific line appear in the log.

timvisee: We don't log anything half way, so I'm afraid we can't really do that.

Issue #3684, opened 2024-02-26
[ColBERT] Tracking issue


IvanPleshkov:
This issue tracks the necessary tasks to integrate late Interaction models into the engine.

Epic description: https://www.notion.so/qdrant/Late-Interaction-models-aka-ColBERT-4b6963314fed45909b803ee7ea19d3d8

## New vector storage
- [x] New simple vector storage for multivector data
  - https://github.com/qdrant/qdrant/pull/3718
  - https://github.com/qdrant/qdrant/pull/3885
- [x] Mmap multivector storage
  - https://github.com/qdrant/qdrant/pull/4106
  - https://github.com/qdrant/qdrant/pull/4194
  - https://github.com/qdrant/qdrant/pull/4197
- [x] Quantization support
  - https://github.com/qdrant/qdrant/pull/4202 

## Integration
- [x] Decouple `Vector` type logic. REST API's vector (and vector-related structures) has to be implemented as separate structure and not defined in `segment` crate
  - https://github.com/qdrant/qdrant/pull/3829
  - https://github.com/qdrant/qdrant/pull/3865
  - https://github.com/qdrant/qdrant/pull/4014
- [x] Extend vector type to support multivectors 
  - https://github.com/qdrant/qdrant/pull/3826
  - https://github.com/qdrant/qdrant/pull/3851
  - https://github.com/qdrant/qdrant/pull/4004
- [x] New metric scorer for multivectors
  - https://github.com/qdrant/qdrant/pull/3713
- [x] HNSW support for multivectors
  - https://github.com/qdrant/qdrant/pull/3880
- [x] Multivector support for segment
  - https://github.com/qdrant/qdrant/pull/3963
  - https://github.com/qdrant/qdrant/pull/3973
- [x] Recommendation support

## API
- [x] REST API
  - write & retrieve https://github.com/qdrant/qdrant/pull/4117
  - config improvements https://github.com/qdrant/qdrant/pull/4171
  - basic search support in REST legacy API https://github.com/qdrant/qdrant/pull/4203
  - new universal search
- [x] gRPC API 
  - write & retrieve https://github.com/qdrant/qdrant/pull/4179
  - basic search support in gRPC legacy API https://github.com/qdrant/qdrant/pull/4203
  - new universal search

## Benchmarks
- [x] internals
  - setup https://github.com/qdrant/qdrant/pull/4005
  - improve MaxSim https://github.com/qdrant/qdrant/pull/4023
- [x] end to end

## Misc.
[...]

PR #3724, opened 2024-02-29
Define segment internal vector type

IvanPleshkov:
Currently, `Vector` type is both a REST API type and internal segment structure. While working with ColBERT vectors we have to add new variant into `Vector` but we cannot because it changes public API.

After discussion, @generall  and @agourlay decided to decouple this logic and implement REST API's `Vector` as separate entity. It also includes another vector-related types which has to be moved.
Discussion thread: https://qdrant.slack.com/archives/C027ADU2611/p1709198634747509

This goal is mentioned as a step of ColBERT tracking issue: https://github.com/qdrant/qdrant/issues/3684

This PR intoduces first step of such refactoring. Here was defined an `internal::Vector` which is a type of segment API without `JsonSchema`. In this PR `internal::Vector` is already integrated into segment.

In next PR's REST API's `Vector` and related stuff (like `NamedVector`) will be moved from segment crate.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -18,8 +18,8 @@ use segment::common::operation_error::OperationError;
 use segment::data_types::groups::GroupId;
 use segment::data_types::order_by::OrderBy;
 use segment::data_types::vectors::{
-    DenseVector, Named, NamedQuery, NamedVectorStruct, QueryVector, Vector, VectorRef,
-    VectorStruct, DEFAULT_VECTOR_NAME,
+    DenseVector, Named, NamedQuery, NamedVectorStruct, QueryVector, Vector, VectorStruct,
+    DEFAULT_VECTOR_NAME,
 };
 use segment::json_path::{JsonPath, JsonPathInterface};
 use segment::types::{
@@ -1258,12 +1258,12 @@ impl Record {
         }
     }
 
-    pub fn get_vector_by_name(&self, name: &str) -> Option<VectorRef> {
+    pub fn get_vector_by_name(&self, name: &str) -> Option<Vector> {
         match &self.vector {
             Some(VectorStruct::Single(vector)) => {
-                (name == DEFAULT_VECTOR_NAME).then_some(vector.into())
+                (name == DEFAULT_VECTOR_NAME).then_some(vector.clone().into())
             }
-            Some(VectorStruct::Multi(vectors)) => vectors.get(name).map(|v| v.into()),
+            Some(VectorStruct::Multi(vectors)) => vectors.get(name).cloned(),
             None => None,
         }
     }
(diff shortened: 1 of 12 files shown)


lib/collection/src/recommendations.rs
```diff
@@ -27,14 +25,14 @@ use crate::operations::types::{
     RecommendRequestInternal, RecommendStrategy, UsingVector,
 };
 
-fn avg_vectors<'a>(vectors: impl Iterator<Item = VectorRef<'a>>) -> CollectionResult<Vector> {
+fn avg_vectors(vectors: impl Iterator<Item = Vector>) -> CollectionResult<Vector> {
```
IvanPleshkov, line 28: Downside of this PR. We take ownership here because `VectorRef` is an segment internal structure not related to `Vector` anymore

lib/segment/src/data_types/vectors.rs
```diff
@@ -12,6 +12,111 @@ use crate::vector_storage::query::context_query::ContextQuery;
 use crate::vector_storage::query::discovery_query::DiscoveryQuery;
 use crate::vector_storage::query::reco_query::RecoQuery;
 
+pub mod internal {
```
IvanPleshkov, line 15: Definition of segment internal vector is here

lib/segment/src/data_types/vectors.rs
```diff
@@ -44,10 +149,10 @@ impl Validate for Vector {
 }
 
 impl<'a> VectorRef<'a> {
-    pub fn to_vec(self) -> Vector {
+    pub fn to_vec(self) -> internal::Vector {
```
IvanPleshkov, line 152: `VectorRef` is related to internal vector, not external. I didn't allow to take `VectorRef` from `Vector`, only from `internal::Vector`

lib/segment/src/data_types/vectors.rs
```diff
@@ -472,21 +579,21 @@ impl<T: Validate> Validate for NamedQuery<T> {
 
 #[derive(Debug, Clone)]
 pub enum QueryVector {
```
IvanPleshkov, line 581: `QueryVector` is internal, it has to use internal version of vector

lib/segment/src/data_types/vectors.rs
```diff
@@ -386,13 +491,15 @@ impl NamedVectorStruct {
         }
     }
 
+    /*
```
agourlay, line 494: clean

lib/collection/src/operations/types.rs
```diff
@@ -1258,12 +1258,12 @@ impl Record {
         }
     }
 
-    pub fn get_vector_by_name(&self, name: &str) -> Option<VectorRef> {
+    pub fn get_vector_by_name(&self, name: &str) -> Option<Vector> {
```
agourlay, line 1261: what about `&Vector`?

IvanPleshkov:
We need additional eyes to review this PR. @timvisee can you help please? =)

lib/collection/src/operations/types.rs
IvanPleshkov: `self.vector` here has a type `Option<VectorStruct>`. Let's look into `VectorStruct` definition:
```
pub enum VectorStruct {
    Single(DenseVector),
    Multi(HashMap<String, Vector>),
}
```
In case of single-vector without name, we cannot convert `&DenseVector` into `&Vector` type. That's why I have taken ownership here because we can convert `DenseVector` into `Vector`

lib/segment/src/data_types/vectors.rs
timvisee: We don't really use this pattern anywhere, except for tests. But at the same time, I don't have a strong objection against it.

timvisee (approved):

lib/collection/src/operations/types.rs
timvisee: Is there really no other way to make it happen?

I can imagine these clones might be coming somewhat expensive if we do it everywhere.

lib/segment/src/data_types/vectors.rs
IvanPleshkov: it's temporary. In next PR's about this issue I gonna move REST vector to collection crate

IvanPleshkov:
After discussion with @generall and @agourlay we decided to close this PR because it's wrong designed. This PR is a step to move REST Vector definition into collection. But it's wrong, we want to move all REST definitions also from collection

PR #3728, opened 2024-02-29
When restarting shard transfer, keep sync if old transfer was sync

timvisee:
Related to: <https://github.com/qdrant/qdrant/pull/3703#discussion_r1507431125>

When restarting a shard transfer, set `sync=true` if the old transfer was also using sync.

This should have been part of <https://github.com/qdrant/qdrant/pull/3703>, but I accidentally merged too early.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/storage/src/content_manager/toc/collection_meta_ops.rs
+++ b/lib/storage/src/content_manager/toc/collection_meta_ops.rs
@@ -373,9 +373,9 @@ impl TableOfContent {
                 )
                 .await?;
 
+                // Preserve sync flag from the old or new transfer
                 let mut new_transfer = transfer.clone();
-                // Preserve sync flag from the old transfer
-                new_transfer.sync = old_transfer.sync;
+                new_transfer.sync = old_transfer.sync || new_transfer.sync;
 
                 self.handle_transfer(collection_id, ShardTransferOperations::Start(new_transfer))
                     .await?;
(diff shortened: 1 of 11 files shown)


generall (approved):

lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -355,27 +356,30 @@ impl TableOfContent {
...
                 .await?;
 
-                // Preserve sync flag from the old or new transfer
-                let mut new_transfer = transfer.clone();
-                new_transfer.sync = old_transfer.sync || new_transfer.sync;
+                let new_transfer = ShardTransfer {
+                    shard_id: transfer_restart.shard_id,
+                    from: transfer_restart.from,
+                    to: transfer_restart.to,
+                    sync: old_transfer.sync, // Preserve sync flag from the old transfer
```
ffuugoo, line 380: Just to clarify, this is desired, intended behavior, right?

(I didn't dig into it deep enough, I have a very vague idea what's going on. Just clarifying, cause it's been changed there and back a few times.)

@generall @timvisee

ffuugoo: Ok, I think I get it now.

Before we reused `ShardTransfer` struct that had `sync` flag, and now we use `RestartShardTransfer`, so there's no second `sync` flag, so we just use `old_transfer.sync` here.

ffuugoo (approved):

PR #2381, opened 2023-08-03
Replace `env_logger` with `tracing`

ffuugoo:
This PR replaces `env_logger` with `tracing`/`tracing-subscriber`/`tracing-log` as our `log` backend (and also removes half of the custom features added in #2187).

(Currently based on top of #2187. Would be rebased onto `dev` once #2187 is merged.)

__TODO:__
- [x] Fix `DEVELOPMENT.md` documentation after removing custom tracing-related features from `qdrant` crate
- [x] ~~Make `tracing` _required_ dependency of `lib/*` crates and remove custom `tracing` feature everywhere?~~
  - There's practically no downside to leaving `tracing` optional in `lib/*` crates, so I crossed it out.

__Further Improvements:__
- Add API to change log-level/filters when Qdrant is already running (#2383)
  - should be pretty trivial to do with [`tracing_subscriber::reload`]
- Add configuration parameter to [select "span events" to be printed][with-span-events]
  - (and an API to change this in runtime)
  - span events are rather verbose to enable them by default, but they are _extremely_ useful for debugging
  - adding configuration parameter should be trivial
  - adding an API - _moderately_ annoying
- Add file appender (and all required configuration/infrastructure), so that logs can be written to a file (#2392)
  - [`RollingFileAppender`] from [`tracing-appender`] crate does not _yet_ support _practical_ rollover strategies
    - https://github.com/tokio-rs/tracing/pull/2323 is merged, but not yet released
    - https://github.com/tokio-rs/tracing/pull/2497 is not yet merged
  - there's also [`tracing-rolling-file`], which looks fine, but I haven't tested it yet
- Implement slog drain that would convert slog records into tracing events, _preserving slog's "structured" KV pairs_ 
  - [`slog-stdlog`] (that we are currently using) is fine
    - but it does not preserve KV pairs
    - I'm not sure if `raft` crate actually emits any, but it would be kinda silly to use `slog` if it does not, so I assume it does
    - and so I'd want to capture/log these KV pairs
  - [`tracing-slog`] is mostly useless
    - it is extremely simplistic and does not preserve KV pairs
    - which is, like, the whole point of `slog` and any _specific_ "slog-to-tracing" bridge
    - because [`slog-stdlog`] is perfectly fine if you don't want/need KV pairs
- Add configuration parameter to select output format?
  - (and an API to change this in runtime)
  - I can already tell, that this will be __super__ annoying to implement, so not sure if it's worth it

[...]

PR #2507, opened 2023-08-25
Add `/logger` HTTP endpoint to get/update logger config at runtime

ffuugoo:
Refactor logger configuration and implement `/logger` HTTP endpoint to get/update logger config at runtime.

Logger configuration now looks like this:

```yaml
# This is "soft-deprecated" now. If both `log_level` and `logger.log_level` are specified,
# `logger.log_level` would be used and warning would be printed.
log_level: INFO

logger:
  log_level: DEBUG

  # Enables printing "fn entered"/"fn exited" logs when using `#[tracing::instrument]` on a function.
  # Very useful during debug when using `#[tracing::instrument]`, but (sometimes) very verbose.
  # But rebuilding just to change this option is very annoying.
  span_events: ["new", "close"]

  # To force-enable colors on GitHub CI. 😂
  color: true
```

I'm sorry for the write-only code, but `tracing`/`tracing-subscriber` is terrible for the runtime configuration. ¯\\\_(ツ)\_/¯

__TODO:__
- [x] add a comment, that logging can't be used from within `tracing.rs`, when logger config is being updated (it's a deadlock)
  - ~~changed the way logger is updated, so this is not critical now~~
    - ~~it does not work (and it's all `tracing-subscriber`'s fault) 😂😭~~
- [ ] ~~update OpenAPI spec? 🤔~~
  - we decided to not include `/logger` in the spec
- [x] test the API 😂

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [ ] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #2392, opened 2023-08-07
Implement logging into a file on disk

ffuugoo:
Fixes #2505.

Extends logger to support logging to a file on disk.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #3733, opened 2024-03-01
Stop uploading logs on consensus test error

agourlay:
This PR removes the upload of logs when the consensus tests fail.

The reason is that, since we changed our logs backend in https://github.com/qdrant/qdrant/pull/2381 the logs are now outputted to stdout.

This makes debugging in the GitHub action UI convenient but it also prevents logs from being written to the files.

The consequence is that we are pushing empty archives now, containing only the directories structure.

empty

I think it makes sense to sense the upload process and keep the current logging configuration.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -62,13 +62,6 @@ jobs:
       - name: Run integration tests - multiple peers - pytest
         run: pytest ./tests/consensus_tests
         timeout-minutes: 60
-      - name: upload logs in case of failure
-        uses: actions/upload-artifact@v4
-        if: failure() || cancelled()
-        with:
-          name: consensus-test-logs
-          retention-days: 90 # default max value
-          path: consensus_test_logs
 
   test-consensus-compose:
 


generall:
Maybe we should start writing logs again? Sometimes those huge stdout walls of text are annoyng

agourlay:
TBH I liked better analyzing the files locally when chasing consensus bugs.

Can we maybe have both?

ffuugoo:
> the logs are now outputted to stdout

Were they outputted to stderr before? This should be a quick fix.

Also, #2507 and #2392. I'll get back to this _soon-ish_, I promise. The last stopper on them was that it issues when changing log levels, but I've since figured out why and how to fix it.

agourlay:
> Were they outputted to stderr before?

I believe yes.

> This should be a quick fix.

Do you mean pushing logs to both outputs or you have something else in mind?

ffuugoo:
> Do you mean pushing logs to both outputs or you have something else in mind?

Switch from stdout to stderr. If we push to both, then it will output each log twice in most cases.

agourlay:
Got a :facepalm: moment, I can simply capture `stdout` instead of `stderr` in our consensus infra.
Trying https://github.com/qdrant/qdrant/pull/3753

agourlay:
Closing in favor of https://github.com/qdrant/qdrant/pull/3753

PR #3736, opened 2024-03-01
Fix state switching order in WAL delta transfer: forward proxy vs partial state

timvisee:

Fix the order of state switching in WAL delta transfer.

Before:
- Transfer all queue proxy items
- Transform queue proxy into forward proxy
- Switch shard from PartialSnapshot to Partial

After:
- Transfer all queue proxy items
- Switch shard from PartialSnapshot to Partial
- Transform queue proxy into forward proxy

Before, it was possible that the forward proxy forwarded a new operation just before switching to Partial state. The remote shard would still be in PartialSnapshot state, rejecting the operation. The sending node would see this as failure, marking the shard as dead (again).

This is the flakyness that appeared in WAL delta tests on CI. Sometimes this did happen, causing the test to fail.

With this change this cannot happen anymore. The forward proxy always forwards to Partial or Active state now, which both allow operations.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/replica_set/shard_transfer.rs
+++ b/lib/collection/src/shards/replica_set/shard_transfer.rs
@@ -363,6 +363,37 @@ impl ShardReplicaSet {
         proxy.transfer_indexes().await
     }
 
+    /// Send all queue proxy updates to remote
+    ///
+    /// This method allows to transfer queued updates at any point, fbefore the shard is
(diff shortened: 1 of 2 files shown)


timvisee:
I've ran this test 200 times locally. No failures anymore.

generall (approved):

PR #3738, opened 2024-03-01
Attempt to improve WAL delta tests, don't fall through transfers finished barrier

timvisee:
This change attempts to improve reliability of some barrier checks in WAL delta tests.

This barrier makes sure to wait until there are no more ongoing shard transfers, before we continue with the rest of the test. But, if we check the condition too early and the shard transfer has not yet started, we simply fall through it, not waiting at all.

To improve on this I've changed this barrier to wait for 2 states in order: first, wait for the specified number of ongoing transfers. Second, wait for no ongoing transfers. This prevents falling through.

This change was my first attempt to fix the recent flaky test we've seen. It didn't solve it directly, but I still think it's a good improvement. That's why I'm submitting it in a separate PR here.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/tests/consensus_tests/test_shard_wal_delta_transfer.py
+++ b/tests/consensus_tests/test_shard_wal_delta_transfer.py
@@ -68,8 +68,8 @@ def test_empty_shard_wal_delta_transfer(capfd, tmp_path: pathlib.Path):
         })
     assert_http_ok(r)
 
-    # Wait for end of shard transfer
-    wait_for_collection_shard_transfers_count(peer_api_uris[0], COLLECTION_NAME, 0)
+    # Wait for transfer to start and end
+    wait_for_collection_shard_transfers_start_complete(peer_api_uris[0], COLLECTION_NAME, 1)
 
     # Confirm empty WAL delta based on debug message in stdout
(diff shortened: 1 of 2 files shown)


generall:
> To improve on this I've changed this barrier to wait for 2 states in order: first, wait for the specified number of ongoing transfers. Second, wait for no ongoing transfers. This prevents falling through.

this just switches problem from skipping too early to basically never skipping, because transfer would be finished before we do the first check

timvisee:
> > To improve on this I've changed this barrier to wait for 2 states in order: first, wait for the specified number of ongoing transfers. Second, wait for no ongoing transfers. This prevents falling through.
> 
> this just switches problem from skipping too early to basically never skipping, because transfer would be finished before we do the first check

In my tests transfers were never faster than 1 second. That means we'd always see it in the check.

But, I'm also fine with not merging this.

generall:
> In my tests transfers were never faster than 1 second. That means we'd always see it in the check.

I have seen similar failures where transfer was finished earlier

timvisee:
I'll drop this PR then. It's not strictly necessary.

PR #3739, opened 2024-03-01
Add unique token to clock tag, detect duplicate operations

timvisee:

Add a unique token to clock tags, to uniquely identify operations having this tag.

The unique token is sent over the wire along with clock tags. Our clock map for highest seen clocks keeps track of the last token as well.

Normally we only allow higher clock tags on incoming operations, rejecting equal or lower tags. This is problematic if we receive the last operation multiple times, because we reject after the first instance. With this change we allow equal clock tags only if their unique token is exactly the same as well. This allows retransmissions and makes operations idempotent. We use the unique token to prevent a different operation with the same clock tag from being accepted in the same manner.

Now, we reject operations when:
- the clock tag is lower than the highest we have seen
- the clock tag is equal to the highest we have seen, and with a different unique token

### Tasks
- [x] Add test explicitly testing acceptance behavior

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -298,7 +298,8 @@ impl RecoveryPoint {
 
     #[cfg(test)]
     pub(crate) fn insert(&mut self, peer_id: PeerId, clock_id: u32, clock_tick: u64) {
-        self.clocks.insert(Key::new(peer_id, clock_id), clock_tick);
+        self.clocks
+            .insert(Key::new(peer_id, clock_id), (clock_tick, Uuid::new_v4()));
     }
 }
 
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -5,6 +5,7 @@ use std::path::Path;
 use api::grpc::qdrant::RecoveryPointClockTag;
 use io::file_operations;
 use serde::{Deserialize, Serialize};
+use tonic::Status;
 use uuid::Uuid;
 
 use crate::operations::types::CollectionError;
@@ -119,7 +120,7 @@ impl ClockMap {
             clocks: self
                 .clocks
                 .iter()
-                .map(|(key, clock)| (*key, clock.current_tick))
+                .map(|(key, clock)| (*key, (clock.current_tick, clock.token)))
                 .collect(),
         }
     }
@@ -200,7 +201,7 @@ impl Clock {
 /// recovering node has not seen yet.
 #[derive(Clone, Debug, Default)]
 pub struct RecoveryPoint {
-    clocks: HashMap<Key, u64>,
+    clocks: HashMap<Key, (u64, Uuid)>,
 }
 
 impl RecoveryPoint {
@@ -213,12 +214,14 @@ impl RecoveryPoint {
         self.clocks
             .iter()
             // TODO: keep clock token here? currently we generate a new unique one.
-            .map(|(key, &tick)| ClockTag::new(key.peer_id, key.clock_id, tick))
+            .map(|(key, &(tick, token))| {
+                ClockTag::new_with_token(key.peer_id, key.clock_id, tick, token)
+            })
     }
 
     /// Increase all existing clocks in this recovery point by the given amount
     pub fn increase_all_clocks_by(&mut self, ticks: u64) {
-        for current_tick in self.clocks.values_mut() {
+        for (current_tick, _token) in self.clocks.values_mut() {
             *current_tick += ticks;
         }
     }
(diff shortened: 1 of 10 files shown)


lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -149,18 +151,27 @@ impl Key {
...
     /// Advance clock to `new_tick`, if `new_tick` is newer than current tick.
     ///
     /// Returns whether the clock was updated and the current tick.
     #[must_use = "clock update status and current tick must be used"]
-    fn advance_to(&mut self, new_tick: u64) -> (bool, u64) {
+    fn advance_to(&mut self, new_tick: u64, token: Uuid) -> (bool, u64) {
+        // Accept if we receive exactly the same tick and token
+        if self.current_tick == new_tick && self.token == token {
+            return (true, self.current_tick);
+        }
```
timvisee, line 173: New early return. If the tick and the token are the same, we 'accept'.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -191,6 +202,7 @@ impl RecoveryPoint {
     pub fn iter_as_clock_tags(&self) -> impl Iterator<Item = ClockTag> + '_ {
         self.clocks
             .iter()
+            // TODO: keep clock token here? currently we generate a new unique one.
```
timvisee, line 205: Important: when updating the cutoff point we currently lose the tokens. In this place we simply generate a new unique one. Though not strictly necessary, we likely want to use the original tokens here.

timvisee: I have added this now: recovery points also hold a unique token.

It is not used for resolving the WAL delta. But it is used to set the cutoff point.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -321,15 +350,26 @@ impl From<RecoveryPoint> for api::grpc::qdrant::RecoveryPoint {
...
+                        Status::invalid_argument(format!("Malformed clock tag token: {err}"))
+                    })
+                    .map(|token: Uuid| {
+                        (Key::new(tag.peer_id, tag.clock_id), (tag.clock_tick, token))
+                    })
+            })
+            .collect::<Result<_, _>>()?;
 
-        Self { clocks }
+        Ok(Self { clocks })
```
ffuugoo, line 372: I'd _maybe_ rewrite with an explicit loop. It's super ugly and hard to read. 😅

This is better, IMO.
```rust
let mut clocks = HashMap::new();

for tag in rp.clocks {
    let token = tag.token.try_into().map_err(|err| {
        Status::invalid_argument(format!("Malformed clock tag token: {err}"))
    })?;

    clocks.insert(Key::new(tag.peer_id, tag.clock_id), (tag.clock_tick, token));
}

Ok(Self { clocks })
```

ffuugoo (approved):
Only style changes and nitpicks. Overall, LGTM. 👍

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -191,12 +213,15 @@ impl RecoveryPoint {
     pub fn iter_as_clock_tags(&self) -> impl Iterator<Item = ClockTag> + '_ {
         self.clocks
             .iter()
-            .map(|(key, &tick)| ClockTag::new(key.peer_id, key.clock_id, tick))
+            // TODO: keep clock token here? currently we generate a new unique one.
+            .map(|(key, &(tick, token))| {
+                ClockTag::new_with_token(key.peer_id, key.clock_id, tick, token)
+            })
```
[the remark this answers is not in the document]
timvisee: Yes, the TODO isn't supposed to be there anymore. Thanks!

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -149,21 +152,40 @@ impl Key {
...
+
+        // Accept equal tick if it has the same unique token
+        if self.current_tick == new_tick {
+            return (self.token == new_token, self.current_tick);
+        }
+
+        // Accept higher ticks
+        self.current_tick = new_tick;
+        self.token = new_token;
+        (true, self.current_tick)
```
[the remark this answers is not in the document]
ffuugoo: It's not really better, though. (I'm just annoyed that token makes it all ugly and confusing... 😭)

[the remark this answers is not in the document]
timvisee: I'll do yours. Less branching.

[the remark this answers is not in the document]
ffuugoo: Maybe this? 🤔

```rust
        if self.current_tick < new_tick {
            self.current_tick = new_token;
            self.token = new_token;

            (true, new_tick)
        } else if self.current_tick == new_tick {
            (self.token == new_token, self.current_tick)
        } else {
            (false, self.current_tick)
        }
```

lib/collection/src/shards/local_shard/clock_map.rs
timvisee: Sure!

Does `HashMap::with_capacity(...)` make sense here?

generall:
I have changed the token type from uuid -> u64, it removed a few ugly conversions.

P.S. I think compatibility test is failing for unrelated reason (it is using snapshot from previous dev, which is wrong)

timvisee:
> I have changed the token type from uuid -> u64, it removed a few ugly conversions.

Great idea. I like that a lot better!

lib/collection/src/shards/local_shard/clock_map.rs
ffuugoo: Yeah, absolutely!

generall (approved):

PR #3742, opened 2024-03-01
Fix: Validate `datetime_range` in `FieldCondition` validation

coszio:
We were missing to include `datetime_range.is_none()` in the validation of FieldCondition, which caused any field condition with only a datetime range to fail validation, as if no condition was set.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/api/src/grpc/validate.rs
+++ b/lib/api/src/grpc/validate.rs
@@ -4,7 +4,7 @@ use std::collections::HashMap;
 use common::validation::{validate_move_shard_different_peers, validate_range_generic};
 use validator::{Validate, ValidationError, ValidationErrors};
 
-use super::qdrant::{GeoLineString, NamedVectors};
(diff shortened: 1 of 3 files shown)


lib/api/src/grpc/validate.rs
```diff
@@ -180,14 +180,26 @@ impl Validate for crate::grpc::qdrant::condition::ConditionOneOf {
...
+            values_count,
+        } = self;
+
+        let all_fields_none = r#match.is_none()
+            && range.is_none()
+            && datetime_range.is_none()
+            && geo_bounding_box.is_none()
+            && geo_radius.is_none()
+            && geo_polygon.is_none()
+            && values_count.is_none();
```
coszio, line 202: With this destructuring it hopefully won't happen again :)

xzfc (approved):

PR #3754, opened 2024-03-04
avoid assert data1 == data2

generall:
assert condition tends to log the whole comparing structure, which generates too much logs


timvisee (approved):
Better!

PR #3757, opened 2024-03-04
Add `ShardTransferOperations::RecoveryToPartial`

ffuugoo:

- Add `RecoveryToPartial` operation (that is, basically, an alias for `SnapshotRecovered`)
- Slightly extend `SnapshotRecovered`/`RecoveryToPartial` handling...
  - so that both operations switch to `Partial` from *either* `PartialSnapshot` *or* `Recovery`
  - this is necessary, because *later*, when upgrading cluster from 1.8.* to 1.9.*, it is possible
    - for 1.8.* nodes to have a replica in `PartialSnapshot` state, and receive `RecoveryToPartial` operation
    - for 1.9.* nodes to have a replica in `Recovery` state, and receive `SnapshotRecovered` operation
  - some other options are possible as well, but allowing switching from both states seems like a simple and safe solution
- Also added `Collection::set_shard_replica_state_impl` method to be able to do the "`PartialSnapshot` *or* `Recovery`" check
  - (please feel free to suggest better method name)

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/toc/collection_meta_ops.rs
+++ b/lib/storage/src/content_manager/toc/collection_meta_ops.rs
@@ -4,7 +4,6 @@ use std::path::Path;
 use async_recursion::async_recursion;
 use collection::collection_state;
 use collection::config::ShardingMethod;
-use collection::operations::types::CollectionError;
 use collection::shards::collection_shard_distribution::CollectionShardDistribution;
 use collection::shards::replica_set::ReplicaState;
 use collection::shards::transfer::ShardTransfer;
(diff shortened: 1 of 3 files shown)


lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -392,28 +393,46 @@ impl TableOfContent {
...
+                                        transfer.to,
+                                        transfer.shard_id,
+                                        ReplicaState::PartialSnapshot,
+                                        ReplicaState::Recovery,
+                                    )))
+                                }
+                            }
+                        }
+                    )
+                    .await?;
```
timvisee, line 435: What about keeping this the same as what we had before, but requesting the current state first?

So:
1. get current replica set local shard state
2. assert it is PartialSnapshot/Recovery
3. call `set_shard_replica_state` with that `from` state

It looks like that would keep things simpler.

In 1.9 or 1.10 we likely remove `PartialSnapshot` in which case it'll only change from `Recovery` to `Partial` again.

ffuugoo: Done.

timvisee (approved):

generall (approved):

PR #3769, opened 2024-03-04
Report basic progress for WAL delta and snapshot transfer

timvisee:
Tracked in: <https://github.com/qdrant/qdrant/issues/3477>

Add basic progress reporting for WAL delta and snapshot transfer.

More specifically: enable reporting on how many records are being transferred (and how many to transfer in total) while the queue proxy is transferring updates to a remote shard.

This is not yet reported, and may be done in a separate PR later:
- process of resolving WAL delta
- snapshot creation
- snapshot transfer
- snapshot recovery

Exposing progress like this will help us improve our tests, and assertions we make in it.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

PR #3753, opened 2024-03-04
Consensus tests capture stdout

agourlay:
context: https://github.com/qdrant/qdrant/pull/3733

This PR changes the consensus tests infrastructure to capture the logs from `stdout` instead of `stderr`.

After merging this, the logs will be available via downloading the archives again.

PR #3775, opened 2024-03-05
For WAL delta test assertions, use transfer status rather than logs

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/3769>

We did depend on log messages to assert various WAL delta transfer states. It appears to be problematic, and does not work with <https://github.com/qdrant/qdrant/pull/3753>.

<https://github.com/qdrant/qdrant/pull/3769> exposes WAL delta transfer progress in the collection cluster output. These tests now depend on that instead.

I was not able to replace all assertions through this new exposed state, in which case I left a comment.

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/3769>
- [x] Rebase on `dev`
- [x] Undraft

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/tests/consensus_tests/test_shard_wal_delta_transfer.py
+++ b/tests/consensus_tests/test_shard_wal_delta_transfer.py
@@ -95,7 +95,7 @@ def test_empty_shard_wal_delta_transfer(tmp_path: pathlib.Path):
     # Wait for end of shard transfer
     wait_for_collection_shard_transfers_count(peer_api_uris[0], COLLECTION_NAME, 0)
 
-    # Assume we got an emtpy WAL delta here, logging 'Resolved WAL delta that is
+    # Assume we got an empty WAL delta here, logging 'Resolved WAL delta that is
     # empty'. But we cannot assert that at this point.
 
     # Doing it the other way around should result in exactly the same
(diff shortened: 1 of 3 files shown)


tests/consensus_tests/test_shard_wal_delta_transfer.py
```diff
@@ -409,9 +391,11 @@ def test_shard_wal_delta_transfer_fallback(capfd, tmp_path: pathlib.Path):
...
     wait_for_collection_shard_transfers_count(peer_api_uris[0], COLLECTION_NAME, 0)
 
-    # Confirm that we fall back to a different method after WAL delta fails
-    stdout, _stderr = capfd.readouterr()
-    assert "Failed to do shard diff transfer, falling back to default method" in stdout
+    # Assume that the WAL delta could not be resolved, preventing a diff
+    # transfer. 'Failed to do shard diff transfer, falling back to default
+    # method' is reported in the logs. But we cannot assert that at this point.
+    # It falls back to streaming records.
+    # TODO(1.9): assert 'streaming_records' transfer method in cluster state
```
ffuugoo, line 398: I'm a bit lost here, why should we only assert this in 1.9? Why can't we do so now?

timvisee: Because in our current implementation we don't properly fall back yet. And the transfers status doesn't show it yet either.

Currently, we simply call the stream records logic once WAL delta transfer fails. We don't change the transfer method through consensus.

In 1.9 we'll change the transfer method through consensus with the new call we implemented.

ffuugoo: Thanks for clarifying!

ffuugoo (approved):

agourlay (approved):

PR #3776, opened 2024-03-05
Try to make `/readyz` check more reliable

ffuugoo:
Try to make `/readyz` check more reliable after cluster upgrade incident last week.

__TODO:__
- [x] Explicitly handle bootstrapping *new* cluster node
  - currently, the node may report "ready" right after startup, because for a short moment before *joining* the cluster, it considers itself the *only* node in the cluster, and `/readyz` instantly returns "ready" in that case

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/common/health.rs
+++ b/src/common/health.rs
@@ -20,9 +20,18 @@ use tokio::{runtime, sync, time};
 const READY_CHECK_TIMEOUT: Duration = Duration::from_millis(500);
 const GET_CONSENSUS_COMMITS_RETRIES: usize = 2;
 
+
+/// Structure used to process health checks like `/readyz` endpoints.
 pub struct HealthChecker {
+    // The state of the health checker.
+    // Once set to `true`, it should not change back to `false`.
+    // Initially set to `false`.
     is_ready: Arc<AtomicBool>,
+    // The signal that notifies that state has changed.
+    // Comes from the health checker task.
     is_ready_signal: Arc<sync::Notify>,
+    // Signal to the health checker task, that the API was called.
+    // Used to drive the health checker task and avoid constant polling.
     check_ready_signal: Arc<sync::Notify>,
     cancel: cancel::DropGuard,
 }
@@ -90,8 +99,14 @@ impl HealthChecker {
 pub struct Task {
     toc: Arc<TableOfContent>,
     consensus_state: ConsensusStateRef,
+    // Shared state with the health checker
+    // Once set to `true`, it should not change back to `false`.
     is_ready: Arc<AtomicBool>,
+    // Used to notify the health checker service that the state has changed.
     is_ready_signal: Arc<sync::Notify>,
+    // Driver signal for the health checker task
+    // Once received, the task should proceed with an attempt to check the state.
+    // Usually comes from the API call, but can be triggered by the task itself.
     check_ready_signal: Arc<sync::Notify>,
     cancel: cancel::CancellationToken,
     wait_for_bootstrap: bool,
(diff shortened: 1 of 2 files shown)


src/common/health.rs
```diff
@@ -128,14 +128,20 @@ impl Task {
...
-            // - Check if cluster commit index *decresed* since last check
-            cluster_commit_index = cmp::min(current_cluster_commit_index, cluster_commit_index);
+            // TODO: Do we want to update `cluster_commit_index` here?
+            //
+            // I.e.:
+            // - If we *don't* update `cluster_commit_index`, then we will only wait till the node
+            //   catch up with the cluster commit index *at the moment the node has been started*
+            // - If we *do* update `cluster_commit_index`, then we will keep track of cluster
+            //   commit index updates and wait till the node *completely* catch up with the leader,
+            //   which might be hard (if not impossible) in some situations
```
ffuugoo, line 144: First question: what do we want to do here?

generall: > catch up with the cluster commit index *at the moment the node has been started*

this should be fine, I guess.

There is a middle-ground of doing this update only once, after first sync is complete

ffuugoo:
Second question:

Initially, we *had to* allow `/readyz` to ignore errors while querying other nodes for the cluster commit index. (For the backward compatibility, because older version did not support required APIs for the `/readyz`.)

But now, we can assume, that any version running in the cluster supports `/readyz`, so we can be more strict with the error handling.

So, currently (with changes in this PR!), if the node queries other nodes for the cluster commit *for the first time*, and if *all* requests happened to fail, then the node would effectively ignore cluster commit check. It will continue to the next step (wait for the shards to turn green once).

So, do we want to make this initial cluster commit check more strict? E.g.:
- Do we want to *require* the node to query *at least* 1 other node? 
- Or, maybe, we want to *require* the node to query *at least* the majority of nodes?
- Maybe one of the above, but with fixed number of retries, so it will continue to the next step, if `N` attempts to query other nodes fail?

src/common/health.rs
```diff
@@ -199,28 +220,32 @@ impl Task {
...
+        // other nodes are unavailable or they don't support `GetConsensusCommit` gRPC API).
+        //
+        // So this check is not 100% reliable and can give a false-positive result!
+        let cluster_commit_index = commit_indices
+            .into_iter()
+            .map(|resp| resp.into_inner().commit)
+            .max()
+            .unwrap_or(0);
+
+        Some(cluster_commit_index as _)
```
ffuugoo, line 248: Also, some explanation for the changes.

_For some reason_, we decided to return 0 (which means "ignore cluster commit check"), if we did not get *at least* `required_commit_indices_count` successful responses... which is silly?

Unless someone somehow added a node from *another* cluster to the existing one, or manually modified node data on disk, or *manually* break the cluster in some elaborate way, getting *any* consensus commit is better than getting none.

So I removed this nonsense logic, and I always return `max` of *any* commit indices we have. We only return 0 if *all` requests failed.

src/common/health.rs
```diff
@@ -128,14 +128,20 @@ impl Task {
...
+            // - Check if this is the only node in the cluster
+            let Some(_) = self.cluster_commit_index().await else {
                 self.set_ready();
                 return;
             };
 
-            // - Check if cluster commit index *decresed* since last check
-            cluster_commit_index = cmp::min(current_cluster_commit_index, cluster_commit_index);
+            // TODO: Do we want to update `cluster_commit_index` here?
+            //
```
ffuugoo, line 138: We also did this `min` of the `cluster_commit_index` on each iteration, but Raft guarantees that it will never decrease cluster commit index.

And so, if during any of this continuous queries, *all* requests fail (e.g., because node lost connectivity for a moment) then `cluster_commit_index()` will return 0, and `min(cluster_commit_index, 0)` *is* 0. So we will instantly go to the next step of the `/readyz` check, even though we have *some* reasonable commit already.

Basically, I think this also makes no sense and makes `/readyz` check more flaky, so I've removed the `min` call.

src/common/health.rs
ffuugoo: Ok, then it should be fine as-is.

src/common/health.rs
generall: I may speculate that we did it for the compatibility with versions that didn't support sharing of the consensus version. Anyway, I don't see the reason for keeping the old logic either.

generall (approved):
Found myself struggling to remember how some of the signals work in this scenario, so took a liberty to propose additional comments.

PR #3778, opened 2024-03-05
RBAC: Initial implementation for decoding `Claims`

coszio:
- Adds a new crate for keeping the RBAC-specific logic together.
- Includes the claims for expiration (`exp`) and write access (`w`), read access is always granted.
- Uses HMAC SHA-256 encryption.
- Validates expiration claim as part of decoding.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/rbac/src/jwt.rs
+++ b/lib/rbac/src/jwt.rs
@@ -1,6 +1,6 @@
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 
-#[derive(Deserialize)]
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
 pub struct Claims {
     /// Expiration time
     pub exp: Option<u64>,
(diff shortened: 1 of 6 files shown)


lib/rbac/src/jwt.rs
```diff
@@ -0,0 +1,13 @@
...
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+pub struct Claims {
+    /// Expiration time
+    pub exp: Option<u64>,
+
+    /// Read access
+    pub r: Option<bool>,
+
+    /// Read-write access
+    pub rw: Option<bool>,
```
generall, line 12: maybe one bool would be enough?
How it can be write access without read?

xzfc (approved):

lib/rbac/src/jwt.rs
coszio: So only read-only / write access claim? Makes sense

pkg/appimage/qdrant.desktop
```diff
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name=qdrant
 GenericName=Vector Database
-Comment=Advanced and High-Performant Vector Search Technology
```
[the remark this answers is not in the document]
coszio: :( Ok

PR #3784, opened 2024-03-06
Cancel shard transfers when shard is deleted

ffuugoo:
Currently, when "remove shard replica" consensus operation is processed, we don't handle shard transfers related to the removed replica in any way.

This PR:
- cancels all shard transfer tasks and then remove all shard transfers related to the removed shard
- when processing `ShardTransferOperations::Abort`, check all preconditions more carefully, before changing local state:
  - do *not* change replica state to `Dead`, if replica does not exist (happens if replica was deleted, and then transfer was aborted)
  - do *not* remove replica, if shard transfer does not exist (could happen due to `Option::map_or` call)
  - only revert local proxy shard, if transfer *exists* (it was possible to revert wrong shard if multiple `Abort`s were sent)

So far, @generall and I decided to *not* version gate these changes, because it's annoying to do, and it seems unlikely to trigger this code during cluster version upgrade.

__TODO:__
- [x] Can these changes introduce backward compatibility issues?
  - because local state will be slightly different on different versions
    - updated nodes will *not* have the transfer
    - but older nodes will *still* have one
  - __UPD:__ probably not, or, at least, it's non-critical, cause deleting shard replica is a *manual* operation, and it should also be possible to *manually* resolve these inconsistencies
- [x] Explicitly check that shard exists, when restarting the transfer?
  - __UPD:__ this one is still nice to have, and easy to implement, so it stays
- [ ] ~~Handle "shard not found" error during the transfer? Don't report error/don't restart the transfer automatically?~~
  - __UPD:__ seems like this is not strictly necessary, because if *transfer-receiver* node is updated, other fixes in this PR will prevent recreating the shard and restarting transfer infinitely

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/collection_ops.rs
+++ b/lib/collection/src/collection/collection_ops.rs
@@ -194,6 +194,17 @@ impl Collection {
                     }
 
                     replica_set.remove_peer(peer_id).await?;
+
+                    // Collect shard transfers related to removed shard...
+                    let transfers = read_shard_holder.get_transfers(|transfer| {
+                        transfer.from == peer_id || transfer.to == peer_id
+                    });
+
+                    // ...and cancel transfer tasks and remove transfers from internal state
+                    for transfer in transfers {
+                        self.finish_shard_transfer(transfer, Some(&read_shard_holder))
+                            .await?;
+                    }
                 }
             }
         }
--- a/lib/collection/src/collection/collection_ops.rs
+++ b/lib/collection/src/collection/collection_ops.rs
@@ -2,7 +2,9 @@ use std::cmp;
 use std::sync::Arc;
 
 use futures::{future, TryStreamExt as _};
+use lazy_static::lazy_static;
 use segment::types::QuantizationConfig;
+use semver::Version;
 
 use super::Collection;
 use crate::operations::config_diff::*;
@@ -11,6 +13,13 @@ use crate::operations::types::*;
 use crate::optimizers_builder::OptimizersConfig;
 use crate::shards::replica_set::{Change, ReplicaState};
 use crate::shards::shard::PeerId;
+
+lazy_static! {
+    /// When dropping a shard, only cancel all related shard transfers to and from it when all nodes
+    /// are running at least this version. That way, we avoid getting an inconsistent state in
+    /// consensus if some nodes are still running an older version.
+    static ref ABORT_TRANSFERS_ON_SHARD_DROP_FROM_VERSION: Version = Version::parse("1.9.0-dev").unwrap();
+}
 
 impl Collection {
     /// Updates collection params:
@@ -193,14 +202,19 @@ impl Collection {
 
             replica_set.remove_peer(peer_id).await?;
 
-            // Collect shard transfers related to removed shard...
-            let transfers = shard_holder
-                .get_transfers(|transfer| transfer.from == peer_id || transfer.to == peer_id);
+            let all_nodes_cancel_transfers = self
+                .channel_service
+                .all_peers_at_version(ABORT_TRANSFERS_ON_SHARD_DROP_FROM_VERSION.clone());
+            if all_nodes_cancel_transfers {
+                // Collect shard transfers related to removed shard...
+                let transfers = shard_holder
+                    .get_transfers(|transfer| transfer.from == peer_id || transfer.to == peer_id);
 
-            // ...and cancel transfer tasks and remove transfers from internal state
-            for transfer in transfers {
-                self.finish_shard_transfer(transfer, Some(&shard_holder))
-                    .await?;
+                // ...and cancel transfer tasks and remove transfers from internal state
+                for transfer in transfers {
+                    self.finish_shard_transfer(transfer, Some(&shard_holder))
+                        .await?;
+                }
             }
         }
         Ok(())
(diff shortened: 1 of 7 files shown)


ffuugoo:
Draft, because I'm not 100% sure about backward compatibility. Otherwise, changes in the PR should be good. Last two `TODO`s in PR description can be fixed in follow-up PRs.

It might be easier to review the changes commit-by-commit, cause I organized the history and also included a small cleanup in the PR.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -257,33 +256,37 @@ impl Collection {
...
+        let Some(replica_set) = shard_holder.get_shard(&transfer_key.shard_id) else {
+            return Err(CollectionError::bad_request(format!(
+                "Shard {} doesn't exist",
+                transfer_key.shard_id
+            )));
+        };
+
+        let Some(transfer) = shard_holder.get_transfer(&transfer_key) else {
+            return Ok(());
+        };
```
ffuugoo, line 275: As far as I can tell, it *might* be possible for a node to propose multiple `Abort` messages (see `ShardReplicaSet::sync_local_state`, so it's a good idea to only do the following operations if transfer exists. (And as far as I can tell, it's also *safe and correct* to do so.)

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -257,33 +256,37 @@ impl Collection {
...
-            replica_set.set_replica_state(&transfer_key.to, ReplicaState::Dead)?;
-        } else {
-            replica_set.remove_peer(transfer_key.to).await?;
+        if replica_set.peer_state(&transfer.to).is_some() {
+            if transfer.sync {
+                replica_set.set_replica_state(&transfer.to, ReplicaState::Dead)?;
+            } else {
+                replica_set.remove_peer(transfer.to).await?;
+            }
         }
```
ffuugoo, line 283: We only change replica state/remove peer if it *exists* now.

Before we would `set_replica_state(Dead)` even if peer is already removed, which would *re-create* the peer. And then the peer would request shard transfer, because it's `Dead`, so it was non-trivial to cancel `sync = true` shard transfer.

We also only call `remove_peer` if shard transfer *exists*.

Before it was possible to accidentally delete the wrong replica, if multiple `Abort` messages were sent for a *non* `sync = true` transfer, because `.unwrap_or(false)` would trigger `remove_peer` branch if `transfer = None`.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -257,33 +256,37 @@ impl Collection {
...
 
-        if self.this_peer_id == transfer_key.from {
-            transfer::driver::revert_proxy_shard_to_local(
-                shard_holder_guard,
-                transfer_key.shard_id,
-            )
-            .await?;
+        if transfer.from == self.this_peer_id {
+            transfer::driver::revert_proxy_shard_to_local(shard_holder, transfer.shard_id).await?;
```
ffuugoo, line 286: We only trigger this if transfer *exists* as well.

Once again, if multiple `Abort`s were sent, the first would cancel the transfer, *but another transfer might have been started from this node to some other node*, so we might have unproxified the wrong shard.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -257,33 +256,37 @@ impl Collection {
...
         }
 
-        let _finish_was_registered = shard_holder_guard.register_finish_transfer(&transfer_key)?;
+        shard_holder.register_finish_transfer(&transfer_key)?;
```
ffuugoo, line 289: Seems safe to only do this when `transfer.is_some()`, because this just removes the transfer from the local state, so if `transfer = None` already, it's just a no-op.

ffuugoo (commented):
Some explanations

lib/collection/src/collection/collection_ops.rs
```diff
@@ -193,14 +201,19 @@ impl Collection {
 
             replica_set.remove_peer(peer_id).await?;
 
-            // Collect shard transfers related to removed shard...
-            let transfers = shard_holder
-                .get_transfers(|transfer| transfer.from == peer_id || transfer.to == peer_id);
+            let all_nodes_drop_transfers = self
+                .channel_service
+                .all_peers_at_version(ABORT_TRANSFERS_ON_SHARD_DROP_FROM_VERSION);
+            if all_nodes_drop_transfers {
```
timvisee: I've added a simple version gate here. Only drop ongoing transfers to a shard we delete if all nodes are at least 1.9.

It makes the logic more robust, which is worth it in my opinion.

Though it's unlikely to actually go wrong, it'll be painful if it happens. That is because we don't have any instruments to fix such problems.

@generall I heard you're not a fan of version gating here. Is that also true in this specific use case?

timvisee (approved):

lib/collection/src/collection/collection_ops.rs
generall: I think I imagined version-gating through the request to other peers for some reason. But this version looks good to me, it is guaranteed by consensus that during execution of this operation all nodes will see the same versions of other nodes

generall (approved):

PR #3792, opened 2024-03-07
Update peer metadata through consensus for 1.9

timvisee:

Enable updating the peer metadata through consensus if it is outdated.

The groundwork for this was already implemented in <https://github.com/qdrant/qdrant/pull/3702>. But we postponed enabling it to 1.9 for backwards compatibility.

To prevent spamming consensus I've added a cooldown, limiting these updates to one per minute.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/storage/src/content_manager/consensus_manager.rs
+++ b/lib/storage/src/content_manager/consensus_manager.rs
@@ -98,7 +98,7 @@ pub struct ConsensusManager<C: CollectionContainer> {
     /// Consensus thread errors, changed by the consensus thread
     message_send_failures: RwLock<HashMap<String, MessageSendErrors>>,
     /// Last time we attempted to update the peer metadata
-    last_peer_metadata_update: Mutex<Option<Instant>>,
+    next_peer_metadata_update_attempt: Mutex<Instant>,
 }
 
 impl<C: CollectionContainer> ConsensusManager<C> {
@@ -121,7 +121,7 @@ impl<C: CollectionContainer> ConsensusManager<C> {
                 last_update: Utc::now(),
             }),
             message_send_failures: Default::default(),
-            last_peer_metadata_update: Default::default(),
+            next_peer_metadata_update_attempt: Mutex::new(Instant::now()),
         }
     }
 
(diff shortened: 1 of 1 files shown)


lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -94,6 +97,8 @@ pub struct ConsensusManager<C: CollectionContainer> {
     consensus_thread_status: RwLock<ConsensusThreadStatus>,
     /// Consensus thread errors, changed by the consensus thread
     message_send_failures: RwLock<HashMap<String, MessageSendErrors>>,
+    /// Last time we attempted to update the peer metadata
+    last_peer_metadata_update: Mutex<Option<Instant>>,
```
ffuugoo, line 101:
```suggestion
    next_peer_metadata_update_attempt: Mutex<Instant>,
```

lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -717,25 +723,41 @@ impl<C: CollectionContainer> ConsensusManager<C> {
...
+    /// Try to update our peer metadata if it's outdated
+    ///
+    /// It rate limits updating to `CONSENSUS_PEER_METADATA_UPDATE_INTERVAL`.
+    fn try_update_peer_metadata(&self) -> Result<(), StorageError> {
+        // Throttle updates to prevent spamming consensus
+        if self.last_peer_metadata_update.lock().map_or(false, |last| {
+            last.elapsed() < CONSENSUS_PEER_METADATA_UPDATE_INTERVAL
+        }) {
+            return Ok(());
+        }
```
ffuugoo, line 739:
```suggestion
        if Instant::now() < *self.next_peer_metadata_update_attempt.lock() {
            return Ok(());
        }
```

lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -717,25 +723,41 @@ impl<C: CollectionContainer> ConsensusManager<C> {
...
+            .send(ConsensusOperations::UpdatePeerMetadata {
+                peer_id: self.this_peer_id(),
+                metadata: PeerMetadata::current(),
+            });
+        if let Err(err) = result {
+            log::error!("Failed to propose consensus peer metadata update for this peer: {err}");
+        }
+        self.last_peer_metadata_update
+            .lock()
+            .replace(Instant::now());
```
ffuugoo, line 757:
```suggestion
        self.next_peer_metadata_update_attempt
            .lock()
            .replace(Instant::now() + CONSENSUS_PEER_METADATA_UPDATE_INTERVAL);
```

ffuugoo (approved):
Using `next_attempt` is a bit easier than `last_attempt`, IMO. LGTM overall. 👍

lib/storage/src/content_manager/consensus_manager.rs
timvisee: Sure. I like that we can remove the option this way.

generall (approved):

Issue #3777, opened 2024-03-05
[RBAC] Tracking issue


generall:
### Features we want

- Provide access based on token, which have more granular control than current API / read-only API keys.
- Control granularity
    - Collections
    - API endpoints: read / write / list
    - Subsets of data - by payload filters
    - Apply some arbitrary Claims
- Administration
    - TTL
    - Ability to revoke token
    - Ability to generate token on demand
    - Do not change internal storage

## How it might work

Based on https://jwt.io/ and some basic cryptography

- The original API key (the write API-Key) is known to Qdrant Service and the **Administrator**.

- Administrator confirms that the owner of the token is granted permissions to do stuff, specified in the *token* by ***signing*** the token.
- The token contains **Claims** to follow, when the token is provided.
- Qdrant service checks the signature against current configured API key (all tokens are revoked if API key is changed)
- Qdrant modifies the request and executes it according to the provided **Claims**

## Tasks

* [x] Implement basic extendable structure which will be used to de-serialize JWT token content
  * Basic Claims - TTL, read/write => #3778
* [x] Implement JWT checking middle-ware for REST and gRPC #3813
* Implement claims checking for Points API
  * [x] #3801
  * [x] filters, collection_name => #3849
  * [ ] Implement `payload` claim for payload editing APIs (grep for `CollectionUpdateOperations::PayloadOperation` in rbac code)
* [x] Implement claims checking for Collections API #3863
* [x] Implement claims checking for other APIs (cluster, service, snapshot, update) #3908
* [x] Check if token is not revoked #3874
* [x] Integration of advanced checks into API function (probably in ToC)
  * How to make sure we don't implement unprotected APIs in future? compile time / tests?
* [x] Use new claims schema #3941
* Access TOC/collections by auth object.
  * [x] Preparation: #3925, #3932
  * [x] Make `TableOfContent::get_collection` into `pub(crate)` and use `get_collection_by_pass` instead. #3972
  * [x] `Dispatcher::toc()` should take an auth object. #3972
* [x] Prepare web-ui for JWT token generation qdrant/qdrant-web-ui#175
* [x] Unit tests for `storage::rbac` module. #4006 
* [x] Integration tests (pytest) 
  - #3961 
  - #4008
  - #4017
  - #4038

PR #3794, opened 2024-03-07
Use RBAC in actix

xzfc:
Tracked by #3777.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/actix/auth.rs
+++ b/src/actix/auth.rs
@@ -4,7 +4,7 @@ use actix_web::body::{BoxBody, EitherBody};
 use actix_web::dev::{forward_ready, Service, ServiceRequest, ServiceResponse, Transform};
 use actix_web::http::header::Header;
 use actix_web::http::Method;
-use actix_web::{Error, HttpResponse};
+use actix_web::{Error, HttpMessage as _, HttpResponse};
 use actix_web_httpauth::headers::authorization::{Authorization, Bearer};
 use futures_util::future::LocalBoxFuture;
 
@@ -25,12 +25,12 @@ const READ_ONLY_POST_PATTERNS: [&str; 11] = [
     "/collections/{name}/points/search/groups",
 ];
 
-pub struct ApiKey {
+pub struct Auth {
     auth_keys: Option<AuthKeys>,
     whitelist: Vec<WhitelistItem>,
 }
 
-impl ApiKey {
+impl Auth {
     pub fn new(auth_keys: Option<AuthKeys>, whitelist: Vec<WhitelistItem>) -> Self {
         Self {
             auth_keys,
@@ -39,7 +39,7 @@ impl ApiKey {
     }
 }
 
-impl<S, B> Transform<S, ServiceRequest> for ApiKey
+impl<S, B> Transform<S, ServiceRequest> for Auth
 where
     S: Service<ServiceRequest, Response = ServiceResponse<EitherBody<B, BoxBody>>, Error = Error>,
     S::Future: 'static,
@@ -48,11 +48,11 @@ where
     type Response = ServiceResponse<EitherBody<B, BoxBody>>;
     type Error = Error;
     type InitError = ();
-    type Transform = ApiKeyMiddleware<S>;
+    type Transform = AuthMiddleware<S>;
     type Future = Ready<Result<Self::Transform, Self::InitError>>;
 
     fn new_transform(&self, service: S) -> Self::Future {
-        ready(Ok(ApiKeyMiddleware {
+        ready(Ok(AuthMiddleware {
             auth_keys: self.auth_keys.clone(),
             whitelist: self.whitelist.clone(),
             service,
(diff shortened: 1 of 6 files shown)


src/actix/auth.rs
```diff
@@ -126,33 +126,46 @@ where
...
-            let is_allowed = if let Some(ref auth_keys) = self.auth_keys {
-                auth_keys.can_write(&key) || (is_read_only(&req) && auth_keys.can_read(&key))
-            } else {
-                // This code path should not be reached
-                log::warn!("Auth for REST API is set up incorrectly. Denying access by default.");
-                false
-            };
-            if is_allowed {
+        let invalid_api_key = |req: ServiceRequest| -> Self::Future {
+            Box::pin(async {
```
coszio, line 130: Can this be its own function?

src/actix/auth.rs
```diff
@@ -126,33 +126,46 @@ where
...
+                    .into_response(HttpResponse::Forbidden().body("Invalid api-key"))
+                    .map_into_right_body())
+            })
+        };
+
+        let Some(ref auth_keys) = self.auth_keys else {
+            // This code path should not be reached
+            log::warn!("Auth for REST API is set up incorrectly. Denying access by default.");
+            return invalid_api_key(req);
+        };
```
coszio, line 141: I know this was here before, but we can fix this by setting `auth_keys: AuthKeys` in the definition (without `Option`). The constructor should only work if we have indeed AuthKeys

src/common/auth.rs
```diff
@@ -1,14 +1,18 @@
...
+#[derive(Clone)]
 pub struct AuthKeys {
     /// A key allowing Read or Write operations
     read_write: Option<String>,
 
     /// A key allowing Read operations
     read_only: Option<String>,
+
+    jwt_parser: Option<JwtParser>,
 }
```
coszio, line 16: I have a suggestion, but open to discussion:

We can keep `AuthKeys` as is, to encapsulate the raw keys (the data layer). But then also add a secondary structure `Auth` to hold the keys and the jwt parser, which can act as the validation layer. 

Something like:
```suggestion
#[derive(Clone)]
pub struct AuthKeys {
    /// A key allowing Read or Write operations
    read_write: Option<String>,

    /// A key allowing Read operations
    read_only: Option<String>,
}

pub struct Auth {
    api_keys: AuthKeys,
    jwt_parser: Option<JwtParser>,
}
```

src/actix/auth.rs
xzfc: It would require too much type boilerplate...
```rust
fn invalid_api_key<S, B>(
    req: ServiceRequest,
) -> LocalBoxFuture<'static, Result<ServiceResponse<EitherBody<S::Response, BoxBody>>, Error>>
where
    S: Service<ServiceRequest, Response = ServiceResponse<EitherBody<B, BoxBody>>, Error = Error>,
    S::Future: 'static,
    B: 'static,
{
    Box::pin(async {
        Ok(req
            .into_response(HttpResponse::Forbidden().body("Invalid api-key"))
            .map_into_right_body())
    })
}
```

coszio: 🤯 Ok

src/common/auth.rs
xzfc: Do you want to keep `can_read`/`can_write` methods in the `AuthKeys`, or move to `Auth` as well?

Also, I think that common validation logic could be placed in this validation layer. Maybe we should extract it later, when both actix and tonic middlewares would be in the same branch?

src/actix/auth.rs
```diff
@@ -126,33 +126,42 @@ where
...
+
+        let api_key = req
+            .headers()
+            .get("api-key")
+            .and_then(|key| key.to_str().ok());
+
+        let bearer_key = Authorization::<Bearer>::parse(&req);
+        let bearer_key = bearer_key.as_ref().ok().map(|auth| auth.as_ref().token());
+
+        if let Some(key) = api_key.or(bearer_key) {
```
coszio, line 147: Can we lazily parse the bearer key? such that we don't parse it if we have a valid api key from the `"api-key"` header.

src/common/auth.rs
coszio: Yeah, I also think we can generalize it better later. Let's postpone this

src/actix/auth.rs
coszio: I think the flow should be like this: 
If the api-key is incorrect, we should return an error right away. Otherwise, check the bearer token, first as an api-key, and if it fails, then as a JWT (or the other way around, open for discussion).

xzfc: I don't see a clean way to do it without duplicating code or using clones. The original version were using `to_string()` and `into()` to clone strings. Mine operates on `&str`.

```diff
-        // Grab API key from request
-        let key =
-            // Request header
-            req.headers().get("api-key").and_then(|key| key.to_str().ok()).map(|key| key.to_string())
-                // Fall back to authentication header with bearer token
-                .or_else(|| {
-                    Authorization::<Bearer>::parse(&req).ok().map(|auth| auth.as_ref().token().into())
-                });
```

coszio (changes requested):
I think this is the only pending edit to be addressed

https://github.com/qdrant/qdrant/pull/3794#discussion_r1516537132

xzfc:
Closing in favor of #3813.

PR #3798, opened 2024-03-08
Make shard diff transfer fallback to different method through consensus for 1.9

timvisee:

In 1.8 we introduced WAL delta transfer which could fall back to the default shard transfer method. It was implemented in quite a hacky way.

This PR makes the fallback happen through consensus, which has a few advantages:
- we now properly clean up the old transfer and prepare the new one, preventing unexpected state issues
- all nodes are properly aware of the current shard transfer state (and method)

The WAL delta fallback test has been updated to assert we properly fall back.

### Tasks
- [x] Respect the default transfer method as selected by the user

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -139,10 +139,6 @@ impl Collection {
         let shard_holder = self.shards_holder.clone();
         let collection_id = self.id.clone();
         let channel_service = self.channel_service.clone();
-        let default_shard_transfer_method = self
-            .shared_storage_config
-            .default_shard_transfer_method
-            .unwrap_or_default();
 
         let progress = Arc::new(Mutex::new(TransferTaskProgress::new()));
 
@@ -155,7 +151,6 @@ impl Collection {
             channel_service,
             self.snapshots_path.clone(),
             self.name(),
-            default_shard_transfer_method,
             temp_dir,
             on_finish,
             on_error,
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -139,6 +139,10 @@ impl Collection {
         let shard_holder = self.shards_holder.clone();
         let collection_id = self.id.clone();
         let channel_service = self.channel_service.clone();
+        let default_shard_transfer_method = self
+            .shared_storage_config
+            .default_shard_transfer_method
+            .unwrap_or_default();
 
         let progress = Arc::new(Mutex::new(TransferTaskProgress::new()));
 
@@ -151,6 +155,7 @@ impl Collection {
             channel_service,
             self.snapshots_path.clone(),
             self.name(),
+            default_shard_transfer_method,
             temp_dir,
             on_finish,
             on_error,
--- a/lib/collection/src/shards/transfer/driver.rs
+++ b/lib/collection/src/shards/transfer/driver.rs
@@ -90,11 +90,11 @@ pub async fn transfer_shard(
         ShardTransferMethod::WalDelta => {
             let result = transfer_wal_delta(
                 transfer_config.clone(),
-                shard_holder.clone(),
-                progress.clone(),
+                shard_holder,
+                progress,
                 shard_id,
-                remote_shard.clone(),
-                channel_service.clone(),
+                remote_shard,
+                channel_service,
                 consensus,
                 collection_name,
             )
--- a/lib/collection/src/shards/transfer/driver.rs
+++ b/lib/collection/src/shards/transfer/driver.rs
@@ -103,10 +103,7 @@ pub async fn transfer_shard(
 
             // Handle failure, fall back to default transfer method or propagate error
             if let Err(err) = result {
-                log::warn!(
-                    "Failed to do shard diff transfer, falling back to default method {:?}: {err}",
-                    ShardTransferMethod::default(),
-                );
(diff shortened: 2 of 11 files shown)


lib/collection/src/shards/transfer/driver.rs
```diff
@@ -97,97 +101,53 @@ pub async fn transfer_shard(
...
-        collection_id,
-        collection_name,
-        channel_service,
-        snapshots_path,
-        temp_dir,
-    )
-    .await
+    // Propose to restart transfer with a different method
+    transfer_config.method.replace(fallback_method);
+    consensus.restart_shard_transfer(transfer_config, collection_name.into())?;
```
timvisee, line 148: We might want to confirm consensus actually picks this up. Thoughts?

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -140,6 +140,10 @@ impl Collection {
         let shard_holder = self.shards_holder.clone();
         let collection_id = self.id.clone();
         let channel_service = self.channel_service.clone();
+        let default_shard_transfer_method = self
+            .shared_storage_config
+            .default_shard_transfer_method
+            .unwrap_or_default();
```
ffuugoo, line 146: What if you specify `wal_delta` as default shard transfer method in the config file (I believe it's `storage.shard_transfer_method` config parameter)? 🤔

I might be wrong here, but I also believe we are using the same parameter, *[when requesting shard transfers during `Collection::sync_local_state` in the first place](https://github.com/qdrant/qdrant/blob/dev/lib/collection/src/collection/mod.rs#L573-L578)*.

So:
- if you specify `shard_transfer_method: wal_delta`, then our `sync = true` transfers would use WAL diff by default... but then if WAL diff transfer fails, it will try to fallback to WAL diff transfer once again
- if you specify `shard_transfer_method: stream_records`, then WAL diff would properly fallback to stream records transfer... but then `sync = true` transfers would always use stream records, and you could only start WAL diff transfer manually with the API

I think we need to introduce *separate* parameter for WAL diff fallback method. And check that it's *not* `wal_delta` when deserializing config. 😂

Also, I think it's a bit confusing that it's `StorageConfig::shard_transfer_method`, but `SharedStorageConfig::default_shard_transfer_method`.

ffuugoo: I also see how feature gating is implemented in #3800 now.

@generall mentioned this week, that he wants a config parameter to be able to disable WAL diff transfers for `sync = true` transfers (even if supported; like, as an emergency measure, if WAL diffs happen to work poorly in practice), so I think we still want/need to introduce a few parameters, but it will take a bit more thinking how to do so properly.

timvisee: > What if you specify wal_delta as default shard transfer method in the config file (I believe it's storage.shard_transfer_method config parameter)? 🤔

> I might be wrong here, but I also believe we are using the same parameter, [when requesting shard transfers during Collection::sync_local_state in the first place](https://github.com/qdrant/qdrant/blob/dev/lib/collection/src/collection/mod.rs#L573-L578).

The same property is used, correct. But if it tries to fall back to WAL delta transfer again, it simply does not fall back at all. So there is no loop here.

Anyway. @generall mentioned that we want to fall back to streaming records in all cases for now, because snapshot transfer is not super robust yet. In that case we don't need a property to configure it, and can add this later if we so desire if snapshot transfers are up to speed again. I agree with that idea.

timvisee: I've now implemented this. We fall back to stream records, no matter what is configured in the configuration.

lib/collection/src/shards/transfer/driver.rs
ffuugoo: We have to check how restarting interacts with `sync_local_state`. If it will be eventually retried/cleaned up in `sync_local_state`, then this is, _probably_, fine. If not, then I think we either have to wait for consensus (with timeout error; as usual) or fix `sync_local_state`, IMO.

timvisee: I did think about this, but it proved very hard to do in this place. We don't have infrastructure in place here to wait on shard transfer states, nor can we access the `on_apply` callbacks in the consensus manager here.

I'll see what I can do to improve upon this.

timvisee: I found out we actually do have access to consensus. I've simply used our existing `propose_consensus_op_with_await` and added a loop for 3 retries. That should be good enough.

If retrying like this fails 3 times the failure is propagated to the current transfer, which means it will be aborted.

lib/storage/src/content_manager/toc/transfer.rs
```diff
@@ -31,6 +32,7 @@ impl ShardTransferDispatcher {
     }
 }
 
+#[async_trait]
```
ffuugoo, line 35: Ha, neat! I never realized you don't need `#[async_trait]`, if all async methods have default impls.

ffuugoo (approved):

lib/storage/src/content_manager/toc/transfer.rs
timvisee: Yeah it's a bit weird to be fair. I hope some upcoming Rust release will allow us to remove it entirely.

generall, line 35: is this still needed in latest stable?

timvisee: Sadly, yes.

generall (approved):

PR #3800, opened 2024-03-08
Use WAL delta transfer by default for shard recovery for 1.9

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/3792>

Enable WAL delta transfer by default for automatic shard recovery.

As extra safety measure I've added a check to ensure that all nodes are running 1.8+.

We only detect version 1.9 and higher through consensus. I still decided to keep the check at 1.8 and up because that's the first true version in which WAL delta transfer was available.

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/3792>
- [ ] Our dev version is currently 0.11.1, we need to change that for this to work

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -565,7 +565,7 @@ impl Collection {
 
             // Select shard transfer method, prefer user configured method or choose one now
             // If all peers are 1.8+, we try WAL delta transfer, otherwise we use the default method
-            let auto_shard_transfer_method = self
+            let shard_transfer_method = self
                 .shared_storage_config
                 .default_shard_transfer_method
                 .unwrap_or_else(|| {
@@ -587,7 +587,7 @@ impl Collection {
                     shard_id,
                     sync: true,
                     // For automatic shard transfers, always select some default method from this point on
-                    method: Some(auto_shard_transfer_method),
+                    method: Some(shard_transfer_method),
                 };
 
                 if check_transfer_conflicts_strict(&transfer, transfers.iter()).is_some() {
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -37,7 +37,7 @@ use crate::shards::shard::{PeerId, ShardId};
 use crate::shards::shard_holder::{shard_not_found_error, LockedShardHolder, ShardHolder};
 use crate::shards::transfer::helpers::check_transfer_conflicts_strict;
 use crate::shards::transfer::transfer_tasks_pool::{TaskResult, TransferTasksPool};
-use crate::shards::transfer::ShardTransfer;
+use crate::shards::transfer::{ShardTransfer, ShardTransferMethod};
 use crate::shards::{replica_set, CollectionId};
 use crate::telemetry::CollectionTelemetry;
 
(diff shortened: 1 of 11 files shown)


lib/collection/src/collection/mod.rs
```diff
@@ -563,6 +563,19 @@ impl Collection {
...
+            let all_support_wal_delta = self
+                .channel_service
+                .all_peers_at_version(Version::new(1, 8, 0));
+            let default_shard_transfer_method = if all_support_wal_delta {
+                ShardTransferMethod::WalDelta
+            } else {
+                self.shared_storage_config
+                    .default_shard_transfer_method
+                    .unwrap_or_default()
+            };
```
ffuugoo, line 577: As I've mentioned on #3798, we might want to rework this *somehow* to allow users to forcefully disable WAL diff transfers, if wanted/needed.

I think the best way would be to introduce `wal_delta_fallback_method` (or whatever name you prefer), and change this check to something like:

```suggestion
            // Select shard transfer method
            // If all peers are 1.8+, we try WAL delta transfer, otherwise use configured default
            let shard_transfer_method = self.shared_storage_config
                .default_shard_transfer_method
                .unwrap_or_default();
            
            let shard_transfer_method = match shard_transfer_method {
                ShardTransferMethod::WalDelta => {
                    let all_support_wal_delta = self
                        .channel_service
                        .all_peers_at_version(Version::new(1, 8, 0));

                    if all_support_wal_delta {
                        ShardTransferMethod::WalDelta
                    } else {
                        self.shared_storage_config
                            .wal_delta_fallback_method
                            .unwrap_or_default()
                    }
                }

                method => method,
            };
```

timvisee: What would be a good reason for users to disable this?

ffuugoo: What is we missed some testing scenario, and it corrupts data? And now whoever upgraded to 1.9 is stuck, with no way to disable WAL diff transfer.

timvisee: I think the only right solution would be to release a patch version right away.

ffuugoo: That would definitely take more time than for a customer to change config parameter and restart a node. Anyway, take it to @generall, that's his idea.

timvisee: I've changed this now.

It always uses the user configured method here. If it is `null` (default), we make a smart selection between WAL delta (if supported) and streaming records.

lib/collection/src/collection/mod.rs
```diff
@@ -563,6 +563,22 @@ impl Collection {
                 continue;
             }
 
+            // Select shard transfer method, prefer user configured method or choose one now
+            // If all peers are 1.8+, we try WAL delta transfer, otherwise we use the default method
+            let auto_shard_transfer_method = self
```
ffuugoo, line 568: 🪡: Just `shard_transfer_method`? `auto` is a bit confusing, IMO.

lib/collection/src/collection/mod.rs
```diff
@@ -571,11 +587,7 @@ impl Collection {
                     shard_id,
                     sync: true,
                     // For automatic shard transfers, always select some default method from this point on
-                    method: Some(
-                        self.shared_storage_config
-                            .default_shard_transfer_method
-                            .unwrap_or_default(),
-                    ),
+                    method: Some(auto_shard_transfer_method),
```
ffuugoo, line 590: Ditto.

ffuugoo (approved):
LGTM (with a tiny nitpick) 👍

PR #3801, opened 2024-03-08
RBAC: Route `Claims` into `TableOfContent` for points API

coszio:
Tracked by #3777 

There are some claims, like `filter` and `payload` which need to access the deserialized structures from `Filter` and `Payload`. To avoid code duplication in both tonic and actix layers, we'll handle these cases in the table of contents.

- Extracts claims from the extensions of the request in tonic and actix interfaces
- Routes them into table of contents

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/src/actix/api/count_api.rs
+++ b/src/actix/api/count_api.rs
@@ -3,11 +3,12 @@ use actix_web::{post, web, Responder};
 use actix_web_validator::{Json, Path, Query};
 use collection::operations::shard_selector_internal::ShardSelectorInternal;
 use collection::operations::types::CountRequest;
+use rbac::jwt::Claims;
 use storage::content_manager::toc::TableOfContent;
 
 use super::CollectionPath;
 use crate::actix::api::read_params::ReadParams;
-use crate::actix::auth::ActixClaims;
+use crate::actix::auth::Extension;
 use crate::actix::helpers::process_response;
 use crate::common::points::do_count_points;
 
@@ -17,7 +18,7 @@ async fn count_points(
     collection: Path<CollectionPath>,
     request: Json<CountRequest>,
     params: Query<ReadParams>,
-    ActixClaims(claims): ActixClaims,
+    claims: Extension<Claims>,
 ) -> impl Responder {
     let timing = Instant::now();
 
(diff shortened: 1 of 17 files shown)


src/actix/api/count_api.rs
```diff
@@ -35,7 +35,8 @@ async fn count_points(
         count_request,
         params.consistency,
         shard_selector,
-        // ToDo: use timeout from params
+        None, // TODO(RBAC): add claims from request
+              // ToDo: use timeout from params
```
xzfc, line 39: Style nit: don't move this comment.

src/tonic/api/points_internal_api.rs
```diff
@@ -37,9 +38,12 @@ impl PointsInternalService {
 impl PointsInternal for PointsInternalService {
     async fn upsert(
         &self,
-        request: Request<UpsertPointsInternal>,
+        mut request: Request<UpsertPointsInternal>,
     ) -> Result<Response<PointsOperationResponseInternal>, Status> {
         validate_and_log(request.get_ref());
+
+        let claims = request.extensions_mut().remove::<Claims>();
```
xzfc, line 45: I suggest to put code that calls `extensions_mut().insert()` and `extensions_mut().remove()` to the same module. Then this line should look like this:

```suggestion
        let claims = super::auth::extract_claims(&mut request);
```

xzfc (changes requested):

xzfc (approved):

coszio:
Thanks for the rebase @xzfc 🙌, I'll just do a quick refactor of `ActixClaims` into `Extension<T>` in case we want to reuse it later

src/tonic/api/points_internal_api.rs
```diff
@@ -21,6 +21,7 @@ use crate::tonic::api::points_common::{
     delete_payload, delete_vectors, get, overwrite_payload, recommend, scroll, set_payload, sync,
     update_vectors, upsert,
 };
+use crate::tonic::auth::extract_claims;
```
generall, line 24: I don't think we need Claim checks for internal API

generall (approved):

Issue #3802, opened 2024-03-08
Unsafe usage of `std::slice::from_raw_parts`


gautamg795:
`segment::common::mmap_ops::transmute_from_u8_to_slice` and `segment::common::mmap_ops::transmute_from_u8_to_mut_slice` ([link](https://github.com/qdrant/qdrant/blob/3fbe1cae6cb7f51a0c5bb4b45cfe6749ac76ed59/lib/common/memory/src/mmap_ops.rs#L106-L118)) call `std::slice::from_raw_parts` with `&[u8]` slice whose start is potentially unaligned for a slice of `&[T]`.

We've observed panics in `segment::index::hnsw_index::graph_links::GraphLinksConverter::serialize_to` — the panic comes from the [precondition assertion in the standard library](https://doc.rust-lang.org/src/core/slice/raw.rs.html#97).

We only observed this when updating from the `nightly-2023-11-01` toolchain to `nightly-2024-03-05`, though the panic also reproduced on various toolchains in between. I believe the issue is still present on older toolchains, but the precondition check was altered slightly in between to make it detect this case. I believe the code is potentially invoking undefined behavior regardless, as the pointer is never guaranteed to be aligned here.

In the specific case of `GraphLinksConverter::serialize_to`, it's relatively easy to fix — the transmuted slice is only used for the convenience of serializing a `&[T]` into a `&[u8]` mmap slice. I have a patch prepared that removes the undefined behavior from this function by transmuting `&[T]` into `&[u8]` and copying that transmuted slice into the `&[u8]` mmap, rather than trying to do the opposite.

However, there are many other usages of these `transmte_from_u8_*` methods throughout the `segment` crate that are vulnerable to the same issue.

## Current Behavior
Panics in `std::slice::from_raw_parts` caused by `GraphLinksConverter::serialize_to`.

<details>
<summary>Backtrace</summary>

```
thread '<unnamed>' panicked at library/core/src/panicking.rs:155:5:
unsafe precondition(s) violated: slice::from_raw_parts_mut requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`
stack backtrace:
   0:        0x107f100e0 - std::backtrace_rs::backtrace::libunwind::trace::he983ffd396d11726
                               at /rustc/d18480b84fdbf1efc34f62070951334aa833d761/library/std/src/../../backtrace/src/backtrace/libunwind.rs:104:5
[...]

PR #3806, opened 2024-03-11
Fix `GraphLinksConverter::serialize_to` data alignment issues

ffuugoo:
This PR fix data alignment issue in `GraphLinksConverter::serialize_to` that was discovered in #3802.

It also adds a few asserts to the `transmute_from_u8*` functions, to try to catch similar alignment mismatches in the future.

Fixes #3802. Includes #3807.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #3807, opened 2024-03-11
Fix alignment in graph links file

IvanPleshkov:
This PR fixes https://github.com/qdrant/qdrant/issues/3802 with debug assertions, introduced here https://github.com/qdrant/qdrant/pull/3806.

Structure of hnsw links file:
```
1. Header. 4 u64 values (alignment 8 bytes)
2. Reserved bytes for extending header, 32 bytes
3. An array of u64 values describes the size of each HNSW level (alignment 8 bytes)
4. An array with reindexing point id into the internal index, u32 with alignment 4 bytes
5. An array of all flattened links data, u32 with alignment 4 bytes
6. An array of points offset in flattened links data, u64 with alignment 8 bytes
```

All parts definition with the calculation of ranges inside file you can find in `GraphLinksFileHeader` methods:
https://github.com/qdrant/qdrant/blob/master/lib/segment/src/index/hnsw_index/graph_links.rs#L113

The problem is between `5` and `6` parts of file. Part `5` has alignment 4 bytes but part `6` has alignment 8 bytes.

This PR did 2 things:
1. Add additional padding between parts `5` and `6` while construction to guarantee the right aligments
2. Avoid transmutation into slice for part `6` and use conversion `u64::from_ne_bytes`

Technically, only avoiding transmutation may be the correct fix without changing file construction. But I propose to fix the alignment anyway.

This pr adds new field to header - padding size. Adding new header field is safe. Additional header space is defined here:
https://github.com/qdrant/qdrant/blob/master/lib/segment/src/index/hnsw_index/graph_links.rs#L116

And reserved space is zeroed. File allocation is here:
https://github.com/qdrant/qdrant/blob/master/lib/segment/src/index/hnsw_index/graph_links.rs#L275
`set_len` function zeroes file by function documentation.

I expect that avoiding transmutation for wrong-aligned data does not have a performance impact. While HNSW search, we read from wrong-aligned area only 2 `u64` values per search candidate to calculate links slice.

### All Submissions:

### New Feature Submissions:

7. [ ] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
8. [ ] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_links.rs
+++ b/lib/segment/src/index/hnsw_index/graph_links.rs
@@ -52,6 +52,7 @@ struct GraphLinksFileHeader {
     pub levels_count: u64,
     pub total_links_len: u64,
     pub total_offsets_len: u64,
+    pub offsets_padding: u64,
 }
 
 fn get_reindex_slice<'a>(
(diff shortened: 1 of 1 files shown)


lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -103,6 +132,7 @@ impl GraphLinksFileHeader {
             levels_count: arr[1],
             total_links_len: arr[2],
             total_offsets_len: arr[3],
+            offsets_padding: arr[4],
```
ffuugoo, line 135: Feels like this might need some kind of migration?

Or do we have a fixed number of bytes (which is more than `5 * size_of::<u64>()` in this case) allocated for the header, and all unused bytes are set to 0, and so if we try to load an older file, with new code, `arr[4]` will be 0, and all will be good? 🤔

IvanPleshkov: Answered if PR description. There is a reserved data for future header extension

generall: image

generall (approved):

PR #3715, opened 2024-02-28
Cover mmap storages in compatibility tests

agourlay:
I have noticed that our storage compatibility tests do not cover our mmap storage implementations.

This PR tries to mitigate the issue by introducing new collections with mmap configurations.

The archive size grows quickly so it becomes difficult to test all combinations.

Also the refreshed archive now contains information about clock maps.

e.g.

``` ./storage/collections/test_collection_vector_on_disk/
./storage/collections/test_collection_vector_on_disk/payload_index.json
./storage/collections/test_collection_vector_on_disk/0/
./storage/collections/test_collection_vector_on_disk/0/segments/
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/OPTIONS-000035
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/OPTIONS-000021
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/000027.sst
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/000032.log
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/LOCK
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/LOG
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/000025.sst
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/000026.sst
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/config.json
./storage/collections/test_collection_vector_on_disk/0/segments/440ad0d4-61dc-4605-bd05-747b8ed28094/payload_index/IDENTITY
[...]

PR #3816, opened 2024-03-12
Update storage compatibility version

agourlay:
This PR updates our storage compatibility test to `v1.8.1`.

I have generated the data from master and uploaded it to our GCP bucket.

The storage archives contains [two new collections](https://github.com/qdrant/qdrant/pull/3715) with various mmap configuration.

EDIT: I also added the support for testing against the last minor and patch version,

### DIFF
--- a/tests/storage-compat/README.md
+++ b/tests/storage-compat/README.md
@@ -2,7 +2,7 @@
 
 In order to detect quickly breakage of storage compatibility, we check that the current code understands the storage format from the latest stable release.
 
-To not burden the git repository with tracking the large binary files, we are pushing storage archives to our [S3 bucket](https://storage.googleapis.com/qdrant-backward-compatibility/).
+To not burden the git repository with tracking the large binary files, we are pushing storage archives to our [GCP bucket](https://storage.googleapis.com/qdrant-backward-compatibility/).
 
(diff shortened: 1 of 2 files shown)


generall:
I would prefer to test backward compatibility for 1.8.x against 1.7.3. We can switch to 1.8.1 after we release 1.9

agourlay:
Maybe we should test against both:
- 1.8.1
- 1.7.4

That is for any version with test against:
- the previous patch version
- the previous minor version

WDYT?

generall:
> Maybe we should test against both

Is it only because we added more scenarios? Can we simply generate those using 1.7.4?

agourlay:
> Is it only because we added more scenarios?

It is an important part of the motivation.

> Can we simply generate those using 1.7.4?

I will have to backport the test change locally on an old revision. Doable but kinda hacky.

Do you have any concerns with the patch & minor testing scheme I proposed?

generall:
> Do you have any concerns with the patch & minor testing scheme I proposed?

Apart from increased testing time, should be fine.

agourlay:
Alright I have pushed an additional change to check against minor and patch.
Did a bit of bash script refactoring to reuse functions :hammer:

generall (approved):

PR #3832, opened 2024-03-14
Fix optimizations config

generall:
- fix updating of the max_optimization_threads param
-

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -587,6 +587,7 @@ impl LocalShard {
         );
         update_handler.optimizers = new_optimizers;
         update_handler.flush_interval_sec = config.optimizer_config.flush_interval_sec;
+        update_handler.max_optimization_threads = config.optimizer_config.max_optimization_threads;
         update_handler.run_workers(update_receiver);
         self.update_sender.load().send(UpdateSignal::Nop).await?;
 
--- a/lib/collection/src/update_handler.rs
+++ b/lib/collection/src/update_handler.rs
@@ -86,6 +86,7 @@ pub struct UpdateHandler {
     /// Assigns CPU permits to tasks to limit overall resource utilization.
     optimizer_cpu_budget: CpuBudget,
     /// How frequent can we flush data
+    /// This parameter depends on the optimizer config and should be updated accordingly.
     pub flush_interval_sec: u64,
     segments: LockedSegmentHolder,
     /// Process, that listens updates signals and perform updates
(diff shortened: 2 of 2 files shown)


agourlay:
`on_optimizer_config_update` would benefit from a unit test asserting that the config values are propagated to the update handler. 

Not sure how difficult it is in practice though.

lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
```diff
@@ -182,27 +186,36 @@ impl IndexingOptimizer {
...
         let (selected_segment_id, selected_segment_size) = *selected_segment.unwrap();
 
+        let number_of_segments = segments_read_guard.len();
+
+        // If the number of segments if equal or bigger than the default_segments_number
+        // We want to make sure that we at least do not increase number of segments after optimization
+
+        if number_of_segments < self.default_segments_number {
+            return vec![selected_segment_id];
+        }
```
timvisee, line 209: I wonder how well this works if we have multiple indexation jobs (such as 5) at the same time. This is easily possible now with CPU saturation on big CPUs.

lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
```diff
@@ -182,27 +186,36 @@ impl IndexingOptimizer {
...
+            .max_by_key(|(_, vector_size_bytes)| *vector_size_bytes);
         if selected_segment.is_none() {
             return vec![];
         }
         let (selected_segment_id, selected_segment_size) = *selected_segment.unwrap();
 
+        let number_of_segments = segments_read_guard.len();
+
+        // If the number of segments if equal or bigger than the default_segments_number
+        // We want to make sure that we at least do not increase number of segments after optimization
```
timvisee, line 205: Extended a sentence to make a better point.

```suggestion
        // We want to make sure that we at least do not increase number of segments after optimization, thus we take more than one segment to optimize
```

timvisee: If we spawn a bunch of those jobs below or above the limit, we might get a disbalance in number of segments again. But I'm afraid we just have to accept that.

The alternative would be to calculate how the number of segments will change considering all ongoing optimizations, so we can make a better estimated guess here. But that's probably more complex than helpful.

lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
timvisee: If we spawn a bunch of those jobs below or above the limit, we might get a disbalance in number of segments again. But I'm afraid we just have to accept that.

The alternative would be to calculate how the number of segments will change considering all ongoing optimizations, so we can make a better estimated guess here. But that's probably more complex than helpful.

timvisee (approved):

timvisee (approved):

PR #3835, opened 2024-03-15
Only flush clock maps to disk if they have changed

timvisee:

Prevent having a lot of IO writes by flushing clock maps constantly. Only write to disk if they have changed.

Note that I didn't check how this affects IO on syscall level.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -16,7 +16,6 @@ use crate::shards::shard::PeerId;
 pub struct ClockMap {
     clocks: HashMap<Key, Clock>,
     /// Whether this clock map has changed since the last time it was persisted.
-    #[serde(skip, default)]
     changed: bool,
 }
 
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -390,9 +390,7 @@ impl From<ClockMapHelper> for ClockMap {
     fn from(helper: ClockMapHelper) -> Self {
         Self {
             clocks: helper.clocks.into_iter().map(Into::into).collect(),
-            // We get a new clock map with non-default state, assume changed so we store it next
-            // time
-            changed: true,
+            changed: false,
         }
     }
 }
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -43,7 +43,7 @@ impl ClockMap {
         Ok(())
     }
 
-    pub fn store_changed(&mut self, path: &Path) -> Result<()> {
+    pub fn store_if_changed(&mut self, path: &Path) -> Result<()> {
         if self.changed {
             self.store(path)?;
         }
--- a/lib/collection/src/shards/local_shard/clock_map.rs
+++ b/lib/collection/src/shards/local_shard/clock_map.rs
@@ -57,7 +57,7 @@ impl ClockMap {
     /// and applied to the storage, or rejected.
     #[must_use = "operation accept status must be used"]
     pub fn advance_clock_and_correct_tag(&mut self, clock_tag: &mut ClockTag) -> bool {
-        let (clock_updated, current_tick) = self.advance_clock_impl(*clock_tag);
+        let (clock_accepted, current_tick) = self.advance_clock_impl(*clock_tag);
 
         // We *accept* an operation, if its `clock_tick` is *newer* than `current_tick`.
         //
@@ -73,7 +73,7 @@ impl ClockMap {
         //
         // TODO: Should we *reject* operations with `force = true`, *if* `clock_tick = 0`!?
 
-        let operation_accepted = clock_updated || clock_tag.force;
+        let operation_accepted = clock_accepted || clock_tag.force;
 
         if !operation_accepted {
             clock_tag.clock_tick = current_tick;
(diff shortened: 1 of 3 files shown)


lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -376,6 +390,9 @@ impl From<ClockMapHelper> for ClockMap {
     fn from(helper: ClockMapHelper) -> Self {
         Self {
             clocks: helper.clocks.into_iter().map(Into::into).collect(),
+            // We get a new clock map with non-default state, assume changed so we store it next
+            // time
+            changed: true,
```
timvisee, line 395: Please confirm whether my train of thought is correct here. :smile:

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -88,6 +99,9 @@ impl ClockMap {
     /// Returns whether the clock was updated (or initialized) and the current tick.
     #[must_use = "clock update status and current tick must be used"]
     fn advance_clock_impl(&mut self, clock_tag: ClockTag) -> (bool, u64) {
+        // To keep things simple, assume we always change state here
+        self.changed = true;
```
timvisee, line 103: We can consider being smarter about the changed state here. Currently I just assume we change the state in all cases.

I didn't want to make things more complicated by passing 'changed' booleans all over the place to make this work.

Thoughts/opionions?

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -35,8 +38,16 @@ impl ClockMap {
         Ok(clock_map)
     }
 
-    pub fn store(&self, path: &Path) -> Result<()> {
+    pub fn store(&mut self, path: &Path) -> Result<()> {
```
generall, line 41: maybe AtomicBool instead?

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -35,8 +38,16 @@ impl ClockMap {
...
         file_operations::atomic_save_json(path, self)?;
+        self.changed = false;
+        Ok(())
+    }
+
+    pub fn store_changed(&mut self, path: &Path) -> Result<()> {
+        if self.changed {
```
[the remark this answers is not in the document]
ffuugoo: I think it's okay-ish. We're named the method "store-if-changed", it's your fault if you changed the path. We only call it in a single place and path is always the same, so, I'd say it's good enough.

lib/collection/src/shards/local_shard/clock_map.rs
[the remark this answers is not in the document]
timvisee: Yes, just wanted to come back to this change. Should be `false` indeed, since this is only used when loading from disk.

lib/collection/src/shards/local_shard/clock_map.rs
timvisee: We only store from mutable context. It also prevents us from calling it multiple times concurrently.

I'd suggest to change to an atomic boolean if we ever need this from immutable context.

lib/collection/src/shards/local_shard/clock_map.rs
timvisee: It is not correct. Changed to `false`.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -15,6 +15,9 @@ use crate::shards::shard::PeerId;
 #[serde(from = "ClockMapHelper", into = "ClockMapHelper")]
 pub struct ClockMap {
     clocks: HashMap<Key, Clock>,
+    /// Whether this clock map has changed since the last time it was persisted.
+    #[serde(skip, default)]
```
generall: default should be `true`?

timvisee: How so?

It's either from disk or a new one. In both cases I'd consider it unchanged, because if we load it in the same way we'd get the exact same state.

generall: If it is new, it should be `changed=true`, otherwise how we save it? I think. No?

lib/collection/src/shards/local_shard/clock_map.rs
timvisee: Our documentation is a bit off here.

It returns whether it was accepted, not whether it is updated. Those are not the same. See:

https://github.com/qdrant/qdrant/blob/c2aa229e59c58c5d690b9dacb97e72d8dd7d27b2/lib/collection/src/shards/local_shard/clock_map.rs#L173-L182

Shall we just use this accepted flag then to distinguish between changed or not? If we do that, we still set the changed flag on duplicate clock tags that don't actually change any state. But I guess that's a good enough middle ground.

timvisee: I've implemented this now. We only set `changed = true` if some clock tag is _accepted_.

ffuugoo: Yeah, I think it's still a bit better. Some false positives, but a bit less than with just `changed = true` always.

timvisee: By the way, I've updated documentation around the functions accordingly. They now mention that it returns whether the clock tag was accepted.

lib/collection/src/shards/local_shard/clock_map.rs
timvisee: We could do it to be on the safe side. But the above mentioned function makes it unnecessary in the current implementation.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -35,8 +37,16 @@ impl ClockMap {
...
     }
 
-    pub fn store(&self, path: &Path) -> Result<()> {
+    pub fn store(&mut self, path: &Path) -> Result<()> {
         file_operations::atomic_save_json(path, self)?;
+        self.changed = false;
+        Ok(())
+    }
+
+    pub fn store_if_changed(&mut self, path: &Path) -> Result<()> {
```
agourlay, line 46: Doing extra profiling, I can see that building the path with `oldest_clock_path` shows up although we do not use it.

Could we build the path only if required?

generall: I don't think it worth it...

lib/collection/src/shards/local_shard/clock_map.rs
generall: If we are Ok, with not saving default, then let's consider it resolved.

generall (approved):

ffuugoo (approved):

PR #3849, opened 2024-03-18
Implement `collection` claims for points API

xzfc:
Tracked by #3777.
Depends on #3801.

In this PR changes are made for the Points API:
- The new claim `collections` are checked.
- The contents of the new `payload` claim is added to the filters.
- For the payload editing API (e.g. set payload/delete payload), added stubs that reject this action when the `payload` claim is present. Later we could implement query rewrite/vaidation for them.

### DIFF
--- a/lib/storage/src/content_manager/claims.rs
+++ b/lib/storage/src/content_manager/claims.rs
@@ -28,7 +28,6 @@ pub fn check_collection_name(
     })
 }
 
-#[allow(private_bounds)]
 pub fn check_points_op(
     collections: &Option<Vec<String>>,
     payload: &Option<PayloadClaim>,
@@ -43,7 +42,7 @@ pub fn check_points_op(
     Ok(())
 }
 
-trait PointsOpClaimsChecker {
+pub trait PointsOpClaimsChecker {
     /// An iterator over the collection names used in the operation, for checking `collections`
     /// claim.
     fn collections_used(&self) -> impl Iterator<Item = &str>;
--- a/lib/storage/src/content_manager/claims.rs
+++ b/lib/storage/src/content_manager/claims.rs
@@ -17,7 +17,7 @@ use segment::types::{Condition, ExtendedPointId, FieldCondition, Filter, Match, 
 use super::errors::StorageError;
 
 pub fn check_collection_name(
-    collections: &Option<Vec<String>>,
+    collections: Option<&Vec<String>>,
     collection_name: &str,
 ) -> Result<(), StorageError> {
     let ok = collections
@@ -29,8 +29,8 @@ pub fn check_collection_name(
 }
 
 pub fn check_points_op(
-    collections: &Option<Vec<String>>,
-    payload: &Option<PayloadClaim>,
+    collections: Option<&Vec<String>>,
+    payload: Option<&PayloadClaim>,
     op: &mut impl PointsOpClaimsChecker,
 ) -> Result<(), StorageError> {
     for collection in op.collections_used() {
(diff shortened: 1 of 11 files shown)


lib/storage/src/content_manager/toc/point_ops.rs
```diff
@@ -31,13 +32,17 @@ impl TableOfContent {
...
         shard_selector: ShardSelectorInternal,
-        _claims: Option<Claims>,
+        claims: Option<Claims>,
         timeout: Option<Duration>,
     ) -> Result<Vec<ScoredPoint>, StorageError> {
-        // TODO(RBAC): handle claims
+        if let Some(claims) = claims.as_ref() {
+            check_collection_name(claims, collection_name)?;
+            check_points_op(claims, &mut request)?;
+        }
```
coszio, line 44: Let's destructure `Claims` instead, so that we get nice compiler errors when adding/changing fields.

coszio: Something like this

```suggestion
        if let Some(claims) = claims {
            let Claims {
                exp: _,
                w: _,
                collections,
                payload,
            } = claims;

            check_collection_name(collections, collection_name)?;
            check_points_op(payload, &mut request)?;
        }
```

lib/storage/src/content_manager/claims.rs
```diff
@@ -0,0 +1,306 @@
...
+    let ok = claims
+        .collections
+        .as_ref()
+        .map_or(true, |c| c.iter().any(|c| c == collection_name));
+    ok.then_some(()).ok_or_else(|| {
+        StorageError::unauthorized(format!("Collection '{collection_name}' is not allowed"))
+    })
+}
+
+#[allow(private_bounds)]
```
coszio, line 29: I think we can use explicit visibility instead of allowing the lint

lib/storage/src/content_manager/claims.rs
```diff
@@ -0,0 +1,306 @@
...
+                FieldIndexOperations::CreateIndex(_) => failed_payload_claim(),
+                FieldIndexOperations::DeleteIndex(_) => failed_payload_claim(),
+            },
+        }
+    }
+}
+
+/// Helper function to indicate that the operation is not allowed when `payload` claim is present.
+/// Usually used when point IDs are involved.
+fn failed_payload_claim<T>() -> Result<T, StorageError> {
```
coszio, line 252:
```suggestion
fn incompatible_with_payload_claim<T>() -> Result<T, StorageError> {
```

lib/storage/src/content_manager/toc/point_ops.rs
xzfc: Updated.

But OTOH, now we have `check_collection_name` accepting `&Option<Vec<String>>` instead of `&Claims`, and I think it's less expressive and a bit more error-prone. So I'm not sure that this is a strict improvement.

coszio (changes requested):
`claims` section looks complicated, but I am not proposing any better solution that will make it simpler. 

We also need to handle `discover_batch`. Other than that it LGTM

xzfc:
> `lib/storage/src/content_manager/claims.rs` file looks complicated, but I am not proposing any better solution that will make it simpler.

Yes, it's bulky, but the idea is that once we extend the `PointsOpClaimsChecker` trait to support additional claims, the compiler will be forced to implement it on all of the operations.

> We also need to handle `discover_batch`.

Fixed.

coszio (approved):

PR #3851, opened 2024-03-18
Add MultiDense vector to CowVector

agourlay:
Follow up on https://github.com/qdrant/qdrant/pull/3826

Mostly mechanical to enable later storage implementations.


lib/segment/src/data_types/named_vectors.rs
```diff
@@ -82,13 +88,20 @@ impl<'a> From<&'a [VectorElementType]> for CowVector<'a> {
     }
 }
 
+impl<'a> From<&'a [DenseVector]> for CowVector<'a> {
+    fn from(v: &'a [DenseVector]) -> Self {
+        CowVector::MultiDense(Cow::Owned(v.into()))
```
IvanPleshkov, line 93: Here sounds `Cow::Borrowed` option to avoid unnecessary clones

IvanPleshkov (changes requested):

IvanPleshkov (approved):

PR #3864, opened 2024-03-19
Remove duplicated VectorRef conversion

agourlay:
This PR removes the manual `to_vec_ref` converter to use the existing standard `From` implementation instead.


IvanPleshkov (approved):
Good refactoring!

PR #3870, opened 2024-03-19
On manual shard transfers, automatically choose WAL delta if suitable

timvisee:

Shard transfers manually triggered by a user may not have a transfer method specified. At the call site, select an appropriate transfer method based on the cluster state.

In practice, this allows WAL delta to be used automatically if a user manually initiates a shard transfer. WAL delta transfer is already used for automatic shard recovery, this extends it to manual transfers too.

### All Submissions:

### New Feature Submissions:


timvisee:
We decided that this change is obsolete.

We only expect WAL delta transfer to make sense when recovering an existing shard on a node. Manually invoking a transfer is not the way to do that. Instead, automatic shard recovery can utilize WAL delta transfer and that already happens.

Closing it now.

PR #3874, opened 2024-03-19
RBAC: stateful validation with `value_exists`

coszio:
Tracked by #3777

Adds `value_exists` claim, which provides a collection name and some key with a value. This combination must exist in order for the JWT to be valid.

Tonic and Actix middlewares have been refactored to use an async function, which now performs the extra lookup.

Open questions:
- Should we include shard key selection? Right now it is implemented as a read on all shards operation.

TODO: 
- [x] Tests!

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


tests/consensus_tests/requirements.txt
```diff
@@ -0,0 +1,3 @@
+pytest
```
xzfc, line 1: We need to pin dependency versions so it won't break when the upstream gets updated.
Check `requirements.txt`/`requirements-freeze.txt` in other test dirs for the reference.

Also, style nit: missing trailing newline.

coszio: This was just moved from the workflow into a `requirements` file. But I agree it is better to use a `requirements-freeze` file instead

src/common/auth.rs
```diff
@@ -45,33 +55,68 @@ impl AuthKeys {
...
+                let res = self
+                    .toc
+                    .scroll(
+                        &value_exists.collection,
+                        scroll_req,
+                        None,
+                        ShardSelectorInternal::All,
+                        None,
+                    )
+                    .await
```
xzfc: I'd prefer to extract this code to a separate function to keep `validate_request()` simple.

tests/consensus_tests/test_jwt_validation.py
```diff
@@ -0,0 +1,148 @@
...
+import pytest
+import requests
+from consensus_tests.assertions import assert_http_ok
+from consensus_tests.fixtures import create_collection, drop_collection, upsert_random_points
+
+from .utils import encode_jwt, make_peer_folder, start_first_peer, wait_for
+
+N_PEERS = 1
+N_REPLICA = 1
+N_SHARDS = 1
```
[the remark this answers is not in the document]
coszio: Not really, I just wanted to follow the other consensus tests

xzfc (approved):

lib/rbac/src/jwt.rs
```diff
@@ -18,6 +18,25 @@ pub struct Claims {
...
+    /// Validate this token by looking for a value inside a collection.
+    pub value_exists: Option<ValueExists>,
 }
 
 pub type PayloadClaim = HashMap<JsonPath, ValueVariants>;
+
+#[derive(Serialize, Deserialize, PartialEq, Clone, Debug)]
+pub struct ValueExists {
+    pub collection: String,
+    pub key: JsonPath,
```
generall, line 31: We might want to check for multiple keys and values. The example would be is that we want to check for the user exists AND they have required access level

coszio: @generall Ok for me, but we can also do it later. For now, this can be done via concatenation from the user side if refers to same collection. When we do it, we should allow different collections too.

PR #3894, opened 2024-03-22
[RBAC] allow multiple mathes in jwt validation check

generall:
We might want to check for multiple keys and values. The example would be is that we want to check for the user exists AND they have required access level

### DIFF
--- a/lib/rbac/src/jwt.rs
+++ b/lib/rbac/src/jwt.rs
@@ -26,17 +26,43 @@ pub struct Claims {
 pub type PayloadClaim = HashMap<JsonPath, ValueVariants>;
 
 #[derive(Serialize, Deserialize, PartialEq, Clone, Debug)]
+pub struct KeyValuePair {
+    key: JsonPath,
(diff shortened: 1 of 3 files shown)


lib/rbac/src/jwt.rs
```diff
@@ -25,18 +25,44 @@ pub struct Claims {
...
+    }
+}
+
 #[derive(Serialize, Deserialize, PartialEq, Clone, Debug)]
 pub struct ValueExists {
-    pub collection: String,
-    pub key: JsonPath,
-    pub value: ValueVariants,
+    collection: String,
+    matches: Vec<KeyValuePair>,
```
xzfc, line 46: Maybe use `HashMap<JsonPath, ValueVariants>` to match with the `payload` claim?

generall: Not the same. We might want to check multiple values of the same field. Cause field can be array of values

xzfc (approved):

PR #3896, opened 2024-03-22
[RBAC] Access control inside table of content

generall:
- Introduces AccessToken data structure, which should be used to
  - check claims inside our code
  - decouple Claims (interface structure) from check implementation
  - Ensure that conditions are validated


xzfc:
How do you suggest to get `PayloadClaim` from `CollectionsAccessToken`?

xzfc (approved):

PR #3905, opened 2024-03-25
make sure that collection files are deleted even if the collection is not loaded

generall:
There were several reports, that it is impossible to create collection, even if it doesn't exist.

This change doesn't eliminate the root problem (if any), but allows to simply re-try deletion even if the collection is not loaded.

### DIFF
--- a/lib/storage/src/content_manager/toc/collection_meta_ops.rs
+++ b/lib/storage/src/content_manager/toc/collection_meta_ops.rs
@@ -191,6 +191,14 @@ impl TableOfContent {
             });
             Ok(true)
         } else {
+            let path = self.get_collection_path(collection_name);
+            if path.exists() {
+                log::warn!(
+                    "Collection {} is not loaded, but its directory still exists. Deleting it.",
+                    collection_name
+                );
+                tokio::fs::remove_dir_all(path).await?;
+            }
             Ok(false)
         }
     }


lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -191,6 +191,14 @@ impl TableOfContent {
...
             Ok(true)
         } else {
+            let path = self.get_collection_path(collection_name);
+            if path.exists() {
+                log::warn!(
+                    "Collection {} is not loaded, but its directory still exists. Deleting it.",
+                    collection_name
+                );
+                tokio::fs::remove_dir_all(path).await?;
+            }
```
timvisee, line 201: Are we 100% sure this won't delete by accident? That would be painful.

Can we possibly improve it by checking if any locks are held inside the collection directory?

I can't think of a situation where this happens, but I'd prefer to be as careful as possible here.

generall: well, collection is explicitly requested to be deleted. If it existed - it would have been deleted anyway.

I am not sure about locks and how can it potentially help

timvisee: I misread. I thought this was in the collection creation call, as a mechanism to help with collection creation.

I see now. No objections. Thanks!

timvisee (approved):

PR #3907, opened 2024-03-25
Refactor actix response handlers

xzfc:
This PR refactors actix endpoints to use `helpers::{time, time_or_accept, time_or_accept_with_handle}` more, instead of calling `process_response` manually.

The updated code is more concise, particularly because of using `?` instead of more verbose `match`es.

A new helper function `time_or_accept_with_handle` is added, for handlers consisting of two stages: mandatory (various checks), and optional (runs in a background if `?wait=false`).

### DIFF
--- a/src/actix/helpers.rs
+++ b/src/actix/helpers.rs
@@ -1,35 +1,13 @@
 use std::fmt::Debug;
 use std::future::Future;
-use std::io;
 
 use actix_web::rt::time::Instant;
-use actix_web::{error, http, Error, HttpResponse};
+use actix_web::{http, HttpResponse, ResponseError};
 use api::grpc::models::{ApiResponse, ApiStatus};
 use collection::operations::types::CollectionError;
 use serde::Serialize;
 use storage::content_manager::errors::StorageError;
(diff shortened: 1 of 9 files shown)


src/actix/helpers.rs
```diff
@@ -186,33 +225,21 @@ impl actix_web::ResponseError for HttpError {
...
-            StorageError::ChecksumMismatch { .. } => {
-                (http::StatusCode::BAD_REQUEST, err.to_string())
-            }
-            StorageError::Unauthorized { description } => {
-                (http::StatusCode::UNAUTHORIZED, description)
-            }
+        let status_code = match &err {
+            StorageError::BadInput { .. } => http::StatusCode::BAD_REQUEST,
+            StorageError::NotFound { .. } => http::StatusCode::NOT_FOUND,
+            StorageError::ServiceError { .. } => http::StatusCode::INTERNAL_SERVER_ERROR,
```
generall, line 231: ServiceError also contains stack-trace which we don't want't to sent over API

generall (approved):

Issue #3333, opened 2024-01-05
Try to allow `f16` vectors in storage


generall:
**Is your feature request related to a problem? Please describe.**

Currently qdrant uses hard-coded f32 as a data type for all vectors (except quantized).
But in many applications storing f32 might be too expensive and redundant. 

It would be interesting to try to eventually migrate to dynamic switch between f16 and f32 precisions.

**Describe the solution you'd like**

As a first step of the migration, we'd like to experiment with compile-time switch first.

To do this, we would need to introduce a feature-flag, that would switch from f32 to f16.

Make sure that all qdrant functions work properly with f16 type enabled, such as:

- quantization
- mmaps
- querying
- indexing
- ...

**Describe alternatives you've considered**

Make storage type generic, but this easily might be too much effort for a single PR, so it is better to postpone for another iteration.

**Additional context**

We would need to pay extra attention for query types, as interface (REST and gRPC) should not change once we switch to f16.

---

Note for contributors: Please consider this as tracking issue. If you think that it would be beneficial to split the task into multiple smaller PRs, please you are welcome to do so. Bounty will be rewarded for each PR independently

xzfc:
Sent a PR #3373. Not sure if I am on the right track, so there are a few questions:

1. What kind of 16-bit floats do you want: `fp32` or `bf32`? The `half` crate has the both.

2. Which types exactly do we need to update? I've stumbled upon these candidates:
- `segment::data_types::vectors::VectorElementType` — for sure, we need to update this one. Am I right?
- `sparse::common::types::DimWeight` — I think we need to update this too.
- `common::common::types::ScoreType` — what about this?
I did not see any `f32` mentions in `lib/storage`, so, it's not the crate that we are supposed to update?

3. I've started by changing the `segment` crate (PR #3373). Is this the right place to fix? Do you have a suggestion on what should be changed next?

PranayB003:
> sparse::common::types::DimWeight — I think we need to update this too.

I was thinking this needs to be changed as well. Hadn't figured out the rest so didn't implement anything yet. Overall it seems like there are lots of changes to be made (as evident from your PR), do you think it will be a good idea to split them among us?

coszio:
I believe `ScoreType` should remain the same, this is issue refers to the vector representation only.

Also yes, sparse values (`sparse::common::types::DimWeight`) also count as vector elements, so updating them should be included. This probably would need a separate PR from the dense vectors, though.

xhjkl:
[...]

PR #3922, opened 2024-03-26
(WIP) Generic float

maksimryndin:
Issue https://github.com/qdrant/qdrant/issues/3333

Approach - use a generic Float type (trait-based) with zero-cost Newtype pattern with the same memory layout, float size to be set per collection

Design goals
* Be sure not to impact performance or current logic
* Extendable for different float type (not only f16 but also f64 and f128 for high precision applications)
* Backward-compatibe api changes
* Flexible for user 

Pros
* Straightforward (no change for any logic) and natural (type aliases (VectorElementType, ScoreType) were already introduced)
* No performance penalties for runtime
* Extendable for other possible Float types (e.g. f64, and also https://rust-lang.github.io/rfcs/3453-f16-and-f128.html  and https://github.com/rust-lang/rust/issues/116909) as opposed to “features creep” (https://effective-rust.com/features.html). f16 from half crate also implements Float (https://docs.rs/half/latest/half/struct.f16.html#impl-Float-for-f16) 
* Allows for future dynamic switching as opposed to features
* Use num-traits which is already in the dependencies tree so no new dependency is introduced
* Num-traits is widely used in the ecosystem and its traits are implemented by related libs (like half)

Cons
* Slower compilation time (to be measured)
* Binary size increases (to be measured)

Steps for this PR

- [x] Introduce generic VectorElementType and ScoreType with default f32 - the binary compiles 🎉
- [ ] Accommodate tests

Roadmap

- [ ] Introduce Float trait with a default type (f32) - to show a POC, goal - minimal changes and a digestible PR - code compiles and runs as before <-- this PR
- [ ] Make DimWeight generic for SparseVector
- [ ] Extend quantization crate to work with f16
- [ ] Neon cpu specific should be extended to work with f16
- [ ] Add generic params over the whole codebase, accommodate tests - at this stage compile-switched binary should work (Float is selected for all collections)
- [ ] Add support for API to setup float size per collection
- [ ] Duplicate boilerplate trait impls for ScoreType, DimWeight and VectorElementType can be abstracted with macros

Future investigations
* Float type uniformity check across distributed nodes
* User can upgrade (via VectorParamsDiff) to the bigger-sized floats (create collection, update collection)
* (Backlog) Quantization and neon support for f64, f128
* (Backlog) Protobuf API only support f32 and f64 so any greater precision (e.g. f128) requires additions/changes to API

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


generall:
Hey @maksimryndin, thanks for the PR! I see that you changes have a significant impact on everything from storage to API (I have seen that you change types for Score). 

This is goes a bit beyond the original intent of the issue to only change stored types and keep interface intact. So most likely we will not merge it.

For the reference: here is a PR that makes storage generic against datatype https://github.com/qdrant/qdrant/pull/3900

maksimryndin:
> Hey @maksimryndin, thanks for the PR! I see that you changes have a significant impact on everything from storage to API (I have seen that you change types for Score).
> 
> This is goes a bit beyond the original intent of the issue to only change stored types and keep interface intact. So most likely we will not merge it.
> 
> For the reference: here is a PR that makes storage generic against datatype #3900

Hi @generall ! The HTTP and gRPC API interface should be the same (I run the server and use the client as before), lib interfaces have changed a bit (for example for Rust via mere Deref) which is handled by version bump.

`ScoreType` can preserved for now as `f32` while introducing `f16` only.
I thought about the issue in a generic setting to allow for different floats and to make it dynamic per collection in the future (feature-based doesn't allow for that).

Should I close this PR and take another issue?

generall:
Closing in favor of https://github.com/qdrant/qdrant/pull/4122

PR #3924, opened 2024-03-27
implement conversions for primitive byte storage type

generall:


lib/segment/src/data_types/primitive.rs
```diff
@@ -25,11 +25,17 @@ impl PrimitiveVectorElement for VectorElementType {
 }
 
 impl PrimitiveVectorElement for VectorElementTypeByte {
-    fn from_vector_ref(_vector: VectorRef) -> OperationResult<Cow<[Self]>> {
-        unimplemented!("VectorElementUnsignedByte is not implemented")
+    fn from_vector_ref(vector: VectorRef) -> OperationResult<Cow<[Self]>> {
+        let vector_ref: &[VectorElementType] = vector.try_into()?;
+        let byte_vector = vector_ref.iter().map(|&x| x as u8).collect::<Vec<u8>>();
+        Ok(Cow::from(byte_vector))
```
timvisee, line 31: Can't we do this without copy/allocation?

timvisee: I see we basically need to do that because we want to transform a f32 into a byte.

timvisee (approved):

PR #3931, opened 2024-03-27
Speedup consensus thread

ffuugoo:
This PR restructure consensus thread loop to be more efficient.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -931,7 +931,7 @@ impl RaftMessageBroker {
     }
 
     fn message_sender(&self) -> (RaftMessageSender, RaftMessageSenderHandle) {
-        let (messages_tx, messages_rx) = tokio::sync::mpsc::channel(128);
+        let (messages_tx, messages_rx) = tokio::sync::mpsc::channel(256);
         let (heartbeat_tx, heartbeat_rx) = tokio::sync::watch::channel(Default::default());
 
         let task = RaftMessageSender {
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -501,7 +501,6 @@ impl Consensus {
         let mut updates = 0;
         let mut timeout_at = previous_tick + tick_period;
 
-
         // We need to limit the batch size, as application of one batch should be limited in time.
         const RAFT_BATCH_SIZE: usize = 128;
 
@@ -528,7 +527,6 @@ impl Consensus {
                 Err(_) => break,
             };
 
-
             // Those messages should not be batched, so we interrupt the loop if we see them.
             // Motivation is: if we change the peer, it should be done immediately,
             //  otherwise we loose the update on this new peer
@@ -550,7 +548,10 @@ impl Consensus {
             updates += 1;
             timeout_at = Instant::now() + wait_timeout_for_consecutive_messages;
 
-            if previous_tick.elapsed() >= tick_period || updates >= RAFT_BATCH_SIZE || is_conf_change {
+            if previous_tick.elapsed() >= tick_period
+                || updates >= RAFT_BATCH_SIZE
+                || is_conf_change
+            {
                 break;
             }
         }
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -453,6 +453,9 @@ impl Consensus {
         let mut previous_tick = Instant::now();
 
         loop {
+            // Apply in-memory changes to the Raft State Machine
+            // If updates = None, we need to skip this step due to timing limits
+            // If updates = Some(0), means we didn't receive any updates explicitly
             let updates = self.advance_node(previous_tick, tick_period)?;
 
             let mut elapsed = previous_tick.elapsed();
@@ -465,12 +468,15 @@ impl Consensus {
             }
 
             if self.node.has_ready() {
+                // Persist AND apply changes, which were committed in the Raft State Machine
                 let stop_consensus = self.on_ready()?;
 
                 if stop_consensus {
                     return Ok(());
                 }
             } else if updates == Some(0) {
+                // Assume consensus is up-to-date, we can sync local state
+                // Which involves resoling inconsistencies and trying to recover data marked as dead
                 self.try_sync_local_state()?;
             }
         }
@@ -495,12 +501,37 @@ impl Consensus {
         let mut updates = 0;
         let mut timeout_at = previous_tick + tick_period;
 
+
+        // We need to limit the batch size, as application of one batch should be limited in time.
+        const RAFT_BATCH_SIZE: usize = 128;
+
+        let wait_timeout_for_consecutive_messages = tick_period / 10;
+
+        // This loop batches incoming messages, so we would need to "apply" them only once.
+        // The "Apply" step is expensive, so it is done for performance reasons.
+
+        // But on the other hand, we still want to react to rare
+        // individual messages as fast as possible.
+        // To fulfill both requirements, we are going the following way:
+        //   1. Wait for the first message for full tick period.
+        //   2. If the message is received, wait for the next message only for 1/10 of the tick period.
         loop {
+            // This queue have 2 types of events:
+            // - Messages from the leader, like pings, requests to add logs, acks, etc.
+            // - Messages from users, like requests to start shard transfers, etc.
+            //
+            // Timeout defines how long can we wait for the next message.
+            // Since this thread is sync, we can't wait indefinitely.
+            // Timeout is set up to be about the time of tick.
             let message = match self.recv_update(timeout_at) {
                 Ok(message) => message,
                 Err(_) => break,
             };
 
+
+            // Those messages should not be batched, so we interrupt the loop if we see them.
+            // Motivation is: if we change the peer, it should be done immediately,
+            //  otherwise we loose the update on this new peer
             let is_conf_change = matches!(
                 message,
                 Message::FromClient(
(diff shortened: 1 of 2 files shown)


src/consensus.rs
```diff
@@ -584,30 +603,37 @@ impl Consensus {
...
-        let status = self.node.status();
-        // Promote only if leader
-        if status.ss.raft_state != StateRole::Leader {
+
+        let learner = if let Some(learner) = self.find_learner_to_promote() {
+            learner
+        } else {
             return Ok(false);
-        }
+        };
```
timvisee, line 624: Shall we also use the fancy let else syntax here?

src/consensus.rs
```diff
@@ -446,38 +449,54 @@ impl Consensus {
...
-                let have_changes = self.propose_updates(timeout)?;
-                if !have_changes {
-                    self.try_sync_local_state()?;
+                let mut timeout = tick_period.saturating_sub(elapsed);
+
+                if elapsed < tick_period && self.propose_updates(timeout)? {
+                    elapsed = previous_tick.elapsed();
+                    proposed += 1;
+
+                    timeout = Duration::from_millis(5);
```
timvisee, line 470: Where does this 5ms come from?

ffuugoo: My head? :D

The idea here is to "wait for the first operation available on the channel" and then "batch up to 128 of the *already queued* operations, as well as everything that *will be* queued shortly (next 5ms in this case)".

timvisee: My thinking was, would it possibly make more sense to base this on the tick rate? Would it possibly make more sense to use something like `self.config.tick_period_ms / 2`?

ffuugoo: Maybe, but, IMO, that's not all too important in this case. We're gonna loop back into this shortly anyway. If we timeout by some fraction of a tick, then we batch more, if we do fixed timeout, then we (theoretically, at least) lower the latency, because we do `on_ready` faster and send messages faster.

lib/storage/src/content_manager/consensus/persistent.rs
```diff
@@ -120,9 +120,8 @@ impl Persistent {
         self.apply_progress_queue.current()
     }
 
-    pub fn entry_applied(&mut self) -> Result<(), StorageError> {
+    pub fn entry_applied(&mut self) {
         self.apply_progress_queue.applied();
-        self.save()
```
ffuugoo, line 125: This single `save` call, when called after each entry application, slowed the whole consensus sync by up to 7.5 times...

lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -348,11 +348,15 @@ impl<C: CollectionContainer> ConsensusManager<C> {
...
-                .entry_applied()
-                .context("Failed to save new state of applied entries queue")?;
+
+            self.persistent.write().entry_applied();
         }
+
+        self.persistent
+            .write()
+            .save()
+            .context("Failed to save new state of applied entries queue")?;
```
ffuugoo, line 358: I think it should be acceptable to only save this state *once*, after applying all entries.

This is a simple operation, that just dumps a JSON on disk. In the worst case, if the node crashed *after* applying some operations, but before persisting the state, it will cause the node to *reapply* some operations. The only issue with reapplying operations, that I can think of, is reapplying delete collection/drop shard operations.

E.g.:

Let's assume we have this WAL:

1. delete collection "OOPS"
2. create collection "OOPS"

We applied *both* operations, but crashed before persisting the change.

So now we will reapply these operations, and so we will delete *a possibly non-empty* collection, and then create *an empty* one in place of it.

However, I still think that this is not an issue *in practice*. The `apply_entries` method (and its loop) is not very hot most of the time. We usually don't create and delete thousands of collections every second, but only apply a *handful* of operations. In most cases the loop will run for 10 or so operations and then persist the state right after.

The window for the scenario above to happen would be tiny. 

Another case is when the node is lagging behind a lot and the loop is very hot and applying hundreds of operations per `apply_entries` call... but if we crash when syncing that is even less of an issue, cause the node have to sync with consensus first, and then its shards would be marked as dead at the end of sync anyway.

ffuugoo: Also, we can tweak this, by calling `save` after *some* operations that might lose data if reapplied improperly (like delete collection/drop shard/etc), but not all.

ffuugoo: Also, we can revert this change for this PR and decide what to do with this later. It was just convenient for the demonstration.

ffuugoo:
Mentioned this on Slack, but by now I'm even more convinced, that offloading application of committed operations to a separate thread (see [`raft::RawNode::advance_append`] and [`raft::RawNode::advance_apply_to`]) *might* give a huge speedup to the consensus sync.

What happens now, is we are using simpler flow, where we *apply* operations as part of the main consensus loop. This introduces noticeable "delays" in the consensus loop, which cause delays in *message exchange* between nodes.

If we offload application to a separate thread, then we would be able to do both of these things in parallel: receive and persist new messages from the leader, and apply earlier received messages.

Without delays in message exchange, we should be able to receive a *constant* stream of updates from the leader, so at the very least, *receiving* operations should take seconds instead of minutes or hours. And as long as we don't have thousands and thousands of very heave operations, we should be able to *apply* a batch of previously received operations while receiving the next one, so, so far, I'd expect a substantial speedup.

It all kinda depends on how exactly raft crate works when using this flow, cause it's not perfectly documented, but so far it almost always does reasonable stuff, so I'd expect it to do reasonable stuff (i.e., be fast) in this case as well. :D

[`raft::RawNode::advance_append`]: https://docs.rs/raft/latest/raft/raw_node/struct.RawNode.html#method.advance_append
[`raft::RawNode::advance_apply_to`]: https://docs.rs/raft/latest/raft/raw_node/struct.RawNode.html#method.advance_apply_to

ffuugoo:
This is 90% done, I'll give a last try to further simplify the consensus loop and then it's ready for review.

timvisee (approved):

src/consensus.rs
```diff
@@ -448,144 +448,175 @@ impl Consensus {
...
+
+            if let Err(err) = self.advance_node_impl(message) {
+                log::warn!("{err}");
+                continue;
+            }
+
+            updates += 1;
+            timeout_at = Instant::now() + tick_period / 10;
+
+            if previous_tick.elapsed() >= tick_period || updates >= 128 || is_conf_change {
```
generall, line 519: Is 128 a batch size?

src/consensus.rs
```diff
@@ -448,144 +448,175 @@ impl Consensus {
...
+                ),
+            );
+
+            if let Err(err) = self.advance_node_impl(message) {
+                log::warn!("{err}");
+                continue;
+            }
+
+            updates += 1;
+            timeout_at = Instant::now() + tick_period / 10;
```
generall, line 517: why `tick_period / 10`?

src/consensus.rs
ffuugoo: In some sense? `advance_node` does two things: it *propose* new messages to consensus and it *steps* the node with incoming messages from the leader. So this does not directly correlate with how many messages would be send back from this node.

My logic here was:
- we want to tick the node more-or-less reliably every 100ms
- we also want to push as much data into consensus at all times
- every `propose` and every `step` potentially generates some work to be done during `on_ready`
- if we push too many operations (e.g., the node is behind on consensus, and there are a 1000 messages in the `advane_node` queue, and we push all of them into `RawNode`), we might generate too much work to be done in the next 100ms
- so this 128 makes sure we call `on_ready` at least every 128 "events" (propose or step)

src/consensus.rs
ffuugoo: TL;DR: No particular reason for it to be `tick_period / 10`, it just have to be a reasonably short interval relative to `tick_period`.

We kinda have two different situations with consensus:
- sometimes the nodes are lagging behind, so it gets a constant stream of updates from leader, and so we get a lot of messages in the `advance_node` queue
- but most of the time the nodes are up-to-date, and so updates are rather sparse
- when the node is lagging behind we want to push as many updates as fast as possible, so that's why we "batch" multiple `advance_node` calls
- but when updates are sparse (e.g., we only get pings and occasional consensus operation), we also don't want to wait 100ms before calling `on_ready`, because most operations in `raft` crate *does not* depend on `tick`, and if we do `advance_node`/`on_ready` in quick succession, consensus (on a whole cluster level) would progress *faster*

So what this whole `timeout_at` manipulation does is:
- on each iteration of an outer loop in consensus thread, we wait for the *first* event for up to `tick_period`
- but for each event *after* the first one, we only wait for `tick_period / 10`
- this way, when there are a lot of updates queued/incoming, we would pull a lot of them each iteration of outer loop
- but if when updates are sparse, we would wait for `tick_period / 10` for the next update and if there is none, we would go and call `on_ready`, so that the delay between `advance_node` and `on_ready` would be shorter when the load on consensus is low, so that consensus resolves faster

But overall, there is no particular reason for this timeout to be `tick_period / 10`. I used `Duration::from_millis(5)` initially. It could even be `Duration::ZERO` (with some additional tweaks). It just have to be something "relatively short compared to `tick_period`".

generall (approved):

PR #3932, opened 2024-03-27
RBAC: introduce CollectionPass

xzfc:
Part of #3777.

This PR introduces the concept of `CollectionPass`.
It is used to get `Collection` object from `TableOfContent`.
Instead of getting collection by raw string name, now we are forced to perform validation check using `Access` object.

```rust
// Old way:
let collection = toc.get_collection(collection_name /*: &str */)?;

// New way:
let collection_pass = access.check_whole_collection_rights(collection_name)?;
let collection = toc.get_collection_by_pass(&collection_pass /*: &CollectionPass */)?;
```

Additionally, this PR moves `Access` object deconstruction into its methods so the switch to the new claims schema would be easier.

There are still a few places where the old approach is used, we could cover them after updating the claims schema.

### DIFF
--- a/lib/storage/src/content_manager/snapshots/recover.rs
+++ b/lib/storage/src/content_manager/snapshots/recover.rs
@@ -8,13 +8,12 @@ use collection::shards::shard_config::ShardType;
 use collection::shards::shard_versioning::latest_shard_paths;
 use tokio::task::JoinHandle;
 
-use crate::content_manager::claims::check_manage_rights;
 use crate::content_manager::collection_meta_ops::{
     CollectionMetaOperations, CreateCollectionOperation,
 };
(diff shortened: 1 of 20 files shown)


lib/storage/src/content_manager/snapshots/recover.rs
```diff
@@ -56,18 +55,18 @@ pub fn do_recover_from_snapshot(
     access: Access,
     client: reqwest::Client,
 ) -> Result<JoinHandle<Result<bool, StorageError>>, StorageError> {
-    check_manage_rights(&access)?;
+    let multipass = access.check_manage_rights()?;
```
generall, line 58: image

generall (approved):

PR #3941, opened 2024-03-30
RBAC: new JWT schema

xzfc:
Tracked in #3777.

# Primary change

The schema of the claims object is updated to allow fine per-collection access control.
Dropped the `w` claim as redundant.
The rbac matching logic is rehauled to support the new schema.

Examples of the new claims payload.
```js
// Full cluster-wide access
{"access": "rw"}

// Full cluster-wide read-only access
{"access": "r"}

// Per-collection access
{"access": [
  {"collections": ["col-name"],
   "access": "rw"
  },
  {"collections": ["another-col-name"],
   "access": "r",
   "payload": {"foo": "bar"}
  }
]}

// All supported claims after this PR:
// The `exp` and `value_exists` are still the same.
{"exp": 1234567890, "access": "rw", "value_exists": …}
```

# Misc changes

- Code in `storage::rbac::*` cleaned up.

- Method `Access::full()` now accepts mandatory commentary argument (e.g. `Access::full("Internal API")`).
  Now, every place where this method is called has a commentary.

- Drop `impl Deref for Dispatcher`. Now we need explicitly call `dispatcher.toc()` to access ToC.
  In subsequent PRs we'll add mandatory argument of type `Access` to the `toc()` method.


lib/storage/src/rbac/mod.rs
```diff
@@ -1,9 +1,209 @@
...
+    Read,
+
+    /// Read and write access to a collection, with some restrictions.
+    #[serde(rename = "rw")]
+    ReadWrite,
+
+    /// Read and write access to a collection, with no restrictions.
+    #[serde(rename = "m")]
+    Manage,
+}
```
coszio, line 62: We can add collection `manage` access later. For now, and to allow this extensibility without (hopefully) a confusing interface, we might want to rename full cluster access to `Manage`.


```suggestion
pub enum ClusterAccessMode {
    /// Read-only access to the whole cluster.
    #[serde(rename = "r")]
    Read,

    /// Read and write access to the whole cluster.
    #[serde(rename = "m")]
    Manage,
}

#[derive(Serialize, Deserialize, Eq, PartialEq, Ord, PartialOrd, Copy, Clone, Debug)]
pub enum CollectionAccessMode {
    /// Read-only access to a collection.
    #[serde(rename = "r")]
    Read,

    /// Read and write access to a collection, with some restrictions.
    #[serde(rename = "rw")]
    ReadWrite,
}
```

lib/storage/src/rbac/mod.rs
```diff
@@ -1,9 +1,209 @@
...
+    pub access: CollectionAccessMode,
+
+    /// Payload constraints.
+    /// An object where each key is a JSON path, and each value is JSON value.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub payload: Option<PayloadConstraint>,
+}
+
+#[derive(Serialize, Deserialize, Eq, PartialEq, Ord, PartialOrd, Copy, Clone, Debug)]
+pub enum ClusterAccessMode {
```
coszio, line 39: Why is this name specific to the Cluster concept?
```suggestion
pub enum GlobalAccessMode {
```

lib/storage/src/rbac/mod.rs
```diff
@@ -1,9 +1,209 @@
...
+            return Err(StorageError::unauthorized(format!(
+                "Collection {} is not allowed",
+                self.collection,
+            )));
+        }
+        Ok(())
+    }
+}
+
+pub struct CollectionMultipass;
```
coszio, line 175:
```suggestion
/// Creates [CollectionPass]es for all collections
pub struct CollectionMultipass;
```

lib/storage/src/rbac/mod.rs
```diff
@@ -1,9 +1,209 @@
...
+use crate::content_manager::errors::StorageError;
+
+/// A structure that defines access rights.
+#[derive(Serialize, Deserialize, PartialEq, Clone, Debug)]
+#[serde(untagged)]
+pub enum Access {
+    /// Cluster-wide access.
+    Cluster(ClusterAccessMode),
+    /// Access to specific collections.
+    Collection(AccessCollection),
```
coszio, line 19: Other renaming suggestions: 
`AccessCollection` -> `CollectionAccesses`
`AccessCollectionRule` -> `CollectionAccess`

and to avoid repeating `access` in claims:
`AccessCollectionRule.access` -> `AccessCollectionRule.mode`?

lib/storage/src/rbac/mod.rs
```diff
@@ -1,9 +1,209 @@
...
+            Access::Cluster(ClusterAccessMode::Read) => {
+                if mode != CollectionAccessMode::Read {
+                    return Err(StorageError::unauthorized(
+                        "Only read-only access is allowed",
+                    ));
+                }
+            }
+            Access::Cluster(ClusterAccessMode::ReadWrite) => (),
+            Access::Collection(rules) => {
+                let view = rules.find_view(collection_name)?;
```
coszio, line 111: This search is not exhaustive, as there can be more than 1 rule for the same collection

coszio (changes requested):
- https://github.com/qdrant/qdrant/pull/3941#discussion_r1546555334
- https://github.com/qdrant/qdrant/pull/3941#discussion_r1546646718

src/common/auth/claims.rs
```diff
@@ -1,17 +1,13 @@
...
 #[derive(Serialize, Deserialize, PartialEq, Clone, Debug)]
 pub struct Claims {
     /// Expiration time (seconds since UNIX epoch)
     pub exp: Option<u64>,
 
-    /// Write access, default is false. Read access is always enabled with a valid token.
-    pub w: Option<bool>,
-
-    #[serde(flatten)]
     pub access: Access,
```
coszio, line 11: Maybe it is a good idea to keep this full access as default value?

Related to:
https://github.com/qdrant/qdrant/blob/cd7e99d7982cdcc544908fc414a67e614d6ba2ec/tests/consensus_tests/test_jwt_validation.py#L102

lib/storage/src/rbac/mod.rs
xzfc: Could you provide a use-case/reasoning for that?

I was thinking in a different way: rules should be additive, meaning that adding a new rule to the list should not impose new restrictions but rather lift them.

As for multiple rules with the same collection, we could use any rule of the matching type, e.g. use the first one for read-only APIs, and the second for writes:
```json
"access": [
  {"collections":["a"], "access": "r"},
  {"collections":["a"], "access": "rw", "payload": {"x": 0}}
]
```
But it would be complicated, so in this PR I decided to ignore this and just always use the first rule.

coszio: > I was thinking in a different way: rules should be additive, meaning that adding a new rule to the list should not impose new restrictions but rather lift them.

This is also possible, but after some thought, it is not trivial to combine 

My reasoning is, for this example: 
```json
"access": [
  {"collections":["a", "b"], "access": "rw"},
  {"collections":["a", "c"], "access": "r", "payload": {"x": 0}}
]
```

The view for collection "a" generated for this can be:
```json
// Option 1: additive (permissive)
{ "collection": "a", "access": "rw" }

// Option 2: subtractive (restrictive)
{ "collection": "a", "access": "r", "payload": { "x": 0 } }

// Option 3: hierarchical (global access > collection access > payload constraints)
{ "collection": "a", "access": "rw", "payload": { "x": 0 } }
```

Now, if we also have a third config like `{"collections":["a"], "access": "rw", "payload": {"y": 0}}`. How would payload combination work?

I believe the easiest solution is to validate the access struct itself to make sure there aren't multiple configs for the same collection, wdyt?

xzfc: > I believe the easiest solution is to validate the access struct itself to make sure there aren't multiple configs for the same collection, wdyt?

LGTM

lib/storage/src/rbac/mod.rs
```diff
@@ -1,9 +1,209 @@
...
+
+impl<'a> CollectionPass<'a> {
+    pub fn name(&'a self) -> &'a str {
+        &self.0
+    }
+
+    pub fn into_static(self) -> CollectionPass<'static> {
+        CollectionPass(Cow::Owned(self.0.into_owned()))
+    }
+}
```
timvisee, line 194: Doesn't implementing [`to_owned`](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html) instead make more sense?

src/common/auth/jwt_parser.rs
```diff
@@ -59,22 +62,22 @@ mod tests {
...
+                        ("field3".parse().unwrap(), ValueVariants::Bool(true)),
                     ]
                     .into_iter()
                     .collect(),
-                ),
-            },
+                )),
+            }])),
             value_exists: None,
         };
```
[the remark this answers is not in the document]
coszio: we don't yet, but it might be a _nice to have_

lib/storage/src/rbac/mod.rs
xzfc: This would require to have two distinct types (e.g. `CollectionPassOwned` and `CollectionPassRef<'a>`). Having a single type with a `Cow` inside is more elegant, IMO.

lib/storage/src/rbac/mod.rs
xzfc: Renamed to `CollectionAccess`/`CollectionAccessList`/`CollectionAccessView`.

I'd keep `AccessCollectionRule.access` because that's how it's named in JSON.

lib/storage/src/rbac/mod.rs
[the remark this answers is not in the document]
xzfc: Added explicit `allows_as_much_as` method instead of deriving `Ord`.

src/common/auth/jwt_parser.rs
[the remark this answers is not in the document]
xzfc: I'll think about it when adding tests. For now, I'd keep it as is since it's only used in a single test, and it's not supposed to be created outside of `serde::Deserealize` and tests.

lib/storage/src/rbac/mod.rs
timvisee: Having a :cow2: is fine. Though nitpicky, maybe we can rename the function to `to_owned()`  then to follow naming conventions.

lib/storage/src/rbac/mod.rs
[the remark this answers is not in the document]
timvisee: Is there a good reason for not using `is_read()` and `is_write()` functions instead? Because this would still be incompatible if we'd add a new variant that does not fit in this linear state of permission.

src/common/auth/jwt_parser.rs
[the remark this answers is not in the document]
timvisee: Yeah. Since we only use it here; fine by me!

src/common/collections.rs
```diff
@@ -54,7 +55,8 @@ pub async fn do_get_collection(
     name: &str,
     shard_selection: Option<ShardId>,
 ) -> Result<CollectionInfo, StorageError> {
-    let collection_pass = access.check_partial_collection_rights(name)?;
+    let collection_pass =
+        access.check_collection_access(name, false, CollectionAccessMode::Read)?;
```
generall, line 59: I think it should require access to the whole collection to execute this request

generall: Otherwise total number of points can leak

generall (approved):

lib/storage/src/rbac/mod.rs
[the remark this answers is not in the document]
xzfc: Replaced `allows_as_much_as` with this structure:
 
```rust
pub struct AccessRequrements {
    /// Write access is required.
    pub write: bool,
    /// Manage access is required, implies write access.
    pub manage: bool,
    /// If true, the access should be not limited by a payload restrictions.
    pub whole: bool,
}
```

So it could be checked like this:
```rust
access.check_collection_access(&collection_name, AccessRequrements::new().write().whole())?;
```

lib/storage/src/rbac/mod.rs
```diff
@@ -70,21 +71,30 @@ impl Access {
...
         match self {
-            Access::Cluster(mode) if *mode >= min_mode => Ok(CollectionMultipass),
-            _ => Err(StorageError::unauthorized("Cluster access is not allowed")),
+            Access::Global(GlobalAccessMode::Manage) => Ok(CollectionMultipass),
+            Access::Global(GlobalAccessMode::Read) => {
+                if min_mode == GlobalAccessMode::Read {
+                    Ok(CollectionMultipass)
+                } else {
+                    Err(StorageError::unauthorized(
+                        "Only read-only access is allowed",
```
coszio: This is a confusing message. 

If `min_mode` is higher than `read`, and the provided access mode is `read`, maybe it should say something like `"Read-only access is not enough"`

timvisee (approved):
Once all the open review comments are resolved, I'm happy :smile:

lib/storage/src/rbac/mod.rs
```diff
@@ -70,21 +71,30 @@ impl Access {
...
+                    ))
+                }
+            }
+            _ => Err(StorageError::unauthorized("Global access is not allowed")),
```
coszio: Same thing: if the provided access is not `Access::Global` but the user needs global, the message should be the other way around.

```suggestion
            _ => Err(StorageError::unauthorized("JWT needs global access")),
```

coszio (changes requested):
LGTM, except for the backward error messages

lib/storage/src/rbac/mod.rs
```diff
@@ -210,44 +203,61 @@ impl std::fmt::Display for CollectionPass<'_> {
...
-enum AccessMode {
-    Read,
-    ReadWrite,
-    Manage,
+#[derive(Default, Debug, Copy, Clone)]
+pub struct AccessRequrements {
+    /// Write access is required.
+    pub write: bool,
+    /// Manage access is required, implies write access.
+    pub manage: bool,
```
coszio, line 211: Why not use the `Access` enum itself?

xzfc: It's introduced to address this comment: https://github.com/qdrant/qdrant/pull/3941#discussion_r1549325924
Precisely, `Requirements` is added to avoid comparing two enums.

coszio (approved):
Approving to keep momentum forward

PR #3945, opened 2024-03-31
keep jemalloc dependency only for builds on aarch64 and amd64 platforms, fixes #2474

### LINKED ISSUE #2474: <jemalloc>: Unsupported system page size on ARM64 system
I am trying to start `qdrant/qdrant:v1.4.1` docker image on a ARM64 machine (a QNAP NAS  with AL324 CPU and 16 Gigabyte RAM). 
Fails with memory allocation error.

## Current Behavior
Start fails with memory allocation errors.

## Steps to Reproduce

Start qdrant on an ARM64 machine using `docker run qdrant/qdrant:v1.4.1`
 
```
<jemalloc>: Unsupported system page size
<jemalloc>: Unsupported system page size
memory allocation of 5 bytes failed
./entrypoint.sh: line 25:     7 Aborted                 ./qdrant $@
```

## Expected Behavior
Should properly start.

## Possible Solution

Found a similar problem with redis, indicating that this might be a problem of the page size used at build time: https://github.com/docker-library/redis/issues/208

abhishek818:
Fixes: #2474 
/claim #2474 
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,7 +113,7 @@ actix-web-extras = "0.1.0"
 [target.'cfg(target_os = "linux")'.dependencies]
 rstack-self = { version = "0.3.0", optional = true }
 
-[target.'cfg(any(target_arch = "x86_64", target_arch = "aarch64"))'.dependencies]
+[target.'cfg(all(not(target_env = "msvc"), any(target_arch = "x86_64", target_arch = "aarch64")))'.dependencies]
 tikv-jemallocator = "0.5"
 
 [workspace.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,7 +113,7 @@ actix-web-extras = "0.1.0"
 [target.'cfg(target_os = "linux")'.dependencies]
 rstack-self = { version = "0.3.0", optional = true }
 
-[target.'cfg(not(target_env = "msvc"))'.dependencies]
+#['cfg(all(not(target_env = "msvc"), any(target_arch = "x86_64", target_arch = "aarch64")))'.dependencies]
 tikv-jemallocator = "0.5"
 
 [workspace.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,7 +113,7 @@ actix-web-extras = "0.1.0"
 [target.'cfg(target_os = "linux")'.dependencies]
 rstack-self = { version = "0.3.0", optional = true }
 
-[target.'cfg(any(target_arch = "powerpc64le", target_arch = "x86_64", target_arch = "aarch64"))'.dependencies]
+[target.'cfg(any(target_arch = "x86_64", target_arch = "aarch64"))'.dependencies]
 tikv-jemallocator = "0.5"
 
 [workspace.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,7 +113,7 @@ actix-web-extras = "0.1.0"
 [target.'cfg(target_os = "linux")'.dependencies]
 rstack-self = { version = "0.3.0", optional = true }
 
-#['cfg(all(not(target_env = "msvc"), any(target_arch = "x86_64", target_arch = "aarch64")))'.dependencies]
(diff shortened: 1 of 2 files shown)


abhishek818:
@generall Up for review.

Morriz:
@generall any idea why this older (than #3946) PR is not yet approved to run tests? Does @abhishek818 need to call any other commands?

generall:
same here https://github.com/qdrant/qdrant/pull/3946#issuecomment-2028776778

abhishek818:
@generall , have addressed your comment.

abhishek818:
@generall  , have addressed your comment.

timvisee (changes requested):
This does not resolve the issue, nor does it implement the suggested solution.

The original [issue](https://github.com/qdrant/qdrant/issues/2474) describes that jemalloc doesn't work on ARM64 due to an mismatch in page size. This still enables jemalloc without any changes on ARM64.

timvisee (changes requested):
Could you enable it for PowerPC on Linux as well?

See: https://github.com/tikv/jemallocator?tab=readme-ov-file#platform-support

Cargo.toml
```diff
@@ -113,7 +113,7 @@ actix-web-extras = "0.1.0"
 [target.'cfg(target_os = "linux")'.dependencies]
 rstack-self = { version = "0.3.0", optional = true }
 
-[target.'cfg(not(target_env = "msvc"))'.dependencies]
+#['cfg(all(not(target_env = "msvc"), any(target_arch = "x86_64", target_arch = "aarch64")))'.dependencies]
```
Morriz, line 116: @abhishek818 your logic is rather convoluted:
```py
all(not(target_env = "msvc"), any(target_arch = "x86_64", target_arch = "aarch64"))
```
as the first part becomes redundant by adding the second part. This is enough:
```py
any(target_arch = "x86_64", target_arch = "aarch64")
```

Morriz: same goes for the other instances of your logic

Morriz:
@generall like I said in https://github.com/qdrant/qdrant/pull/3946#issuecomment-2028813412, would you be ok with this PR setting the page size to 16 toes how the tests perform?

generall:
> would you be ok with this PR setting the page size to 16 to show the tests perform?

Not on this PR. We would need to test it separetelly and make sure it doesn't have negative effects

Morriz:
just trying with `JEMALLOC_SYS_WITH_LG_PAGE==16` might work (see https://github.com/tikv/jemallocator/blob/master/jemalloc-sys/README.md)

generall:
> just trying with `JEMALLOC_SYS_WITH_LG_PAGE==16` might work (see https://github.com/tikv/jemallocator/blob/master/jemalloc-sys/README.md)

I am afraid this link won't be enough to convince me. We need a proper testing on real machine

Morriz:
I presume PRs will trigger your tests? What is your concern?

abhishek818:
@Morriz  Did some research on the with_lg_page approach, seems to be mixed opinion on this. Refer [link1 ](https://github.com/facebook/buck2/issues/91)and [link2](https://github.com/jemalloc/jemalloc/issues/467).

generall:
> I presume PRs will trigger your tests? What is your concern?


We don't have performance benchmarks for difference architectures in CI

Morriz:
@abhishek818 this PR is pointless as it does not satisfy the objective when the issue is not solved.

Morriz:
> We don't have performance benchmarks for difference architectures in CI

@generall I think we should then add such a test first?

generall:
> @generall I think we should then add such a test first?

That would be, of course, preferred. But also not very trivial, since github doesn't provide arm action runners and we would need to do something custom. So far we are only testing perf regularly on x86

generall:
Hey @abhishek818, it looks like `PowerPC` doesn't actually work. Let's rollback to just aarch and x86

generall:
@abhishek818 sorry for poking you again, but now the windows build doesn't work. I think `msvc` should not be removed

timvisee (approved):

PR #3946, opened 2024-03-31
fix: include jemalloc only for ARM64/AMD64 architectures

Dhanus3133:
Fixes: #2474
/claim #2474

### All Submissions:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -32,7 +32,7 @@ use storage::content_manager::consensus_manager::{ConsensusManager, ConsensusSta
 use storage::content_manager::toc::transfer::ShardTransferDispatcher;
 use storage::content_manager::toc::TableOfContent;
 use storage::dispatcher::Dispatcher;
-#[cfg(not(target_env = "msvc"))]
+#[cfg(all(not(target_env = "msvc"), not(target_arch = "aarch64")))]
 use tikv_jemallocator::Jemalloc;
 
 use crate::common::helpers::{
@@ -47,7 +47,7 @@ use crate::settings::Settings;
 use crate::snapshots::{recover_full_snapshot, recover_snapshots};
 use crate::startup::{remove_started_file_indicator, touch_started_file_indicator};
 
-#[cfg(not(target_env = "msvc"))]
+#[cfg(all(not(target_env = "msvc"), not(target_arch = "aarch64")))]
 #[global_allocator]
 static GLOBAL: Jemalloc = Jemalloc;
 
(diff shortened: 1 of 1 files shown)


generall:
i think it should work for aarch64. At least we don't observe any problems running it.

I think it should be excluded for all architectures other than aarch64 and amd64

Morriz:
> i think it should work for aarch64. At least we don't observe any problems running it.
> 
> I think it should be excluded for all architectures other than aarch64 and amd64

@generall A lot of people are seeing issues with jemalloc on certain arm64 platforms (like raspberry Pi, but also others, just google "jemalloc aarch64"), so dismissing those issues seems not the right way forward...is the bounty still on for making it conditional (exclude msvc and aarch64)? Personally I think that the best solution would be to set better defaults for page size depending on platform and to override it if needed with some env var. What do you think?

generall:
Changing page size would require additional testing and is not as simple as just opt-out for specific architectures.

The current version works well enough on aws aarch64 machines which we use a lot.

generall:
@Morriz but i am open to suggestions of how to implement it. If you could create a separate issue with the description of how to test it and make sure it won't break anything that's already working, i am happy to assign a bounty to it as well.

Morriz:
It would be up to the user to override the page size, so imo that is none of your concern other than guidance + warnings by example in the docs. Why not just make the page size configurable and set sane defaults (RPi uses 16k, OSX also, so checking linux/darwin is a start). Your test suite should not test jemalloc and may just follow their contracts (if you design by contract that is which I promote), and they offer that parameter so why not?

Morriz:
After reading this: https://github.com/redis/redis/pull/11407#issuecomment-1300766392,
I suggest adding `ARG JEMALLOC_PAGE_SIZE` and compiling like this:
```
JEMALLOC_CONFIGURE_OPTS="--with-lg-page=${JEMALLOC_PAGE_SIZE:-64}"
```

Morriz:
After reading this: https://github.com/redis/redis/pull/11407#issuecomment-1300766392,
I am sad to understand the binaries use static linking and so the page size can't be made configurable. But would it be too much to ask to build one extra target prefixing the tag with `-p16` (I assume your current build is for page size 64?)

Morriz:
I see home assistant choosing to do this: https://github.com/home-assistant/docker-base/pull/248

Morriz:
just trying with `JEMALLOC_SYS_WITH_LG_PAGE==16` might work (see https://github.com/tikv/jemallocator/blob/master/jemalloc-sys/README.md)

generall:
Sorry, I will close this in favor of https://github.com/qdrant/qdrant/pull/3945

It was the first and also more correct solution.

PR #3960, opened 2024-04-02
Decouple SegmentConfig

agourlay:
Relies on a hack to anonymize the new API types


agourlay:
No need for decoupling, we are not guaranteeing backward compatibility on the telemetry types.

PR #3961, opened 2024-04-02
RBAC: [tests] exhaustive access tests

coszio:
Tracked by #3777 

These tests aim to cover all of REST and gRPC surface with different access tokens, such that we make sure that they adhere to what we define as `read`, `read-write`, or `manage` privileges when using JWT for authorization.

Exhaustiveness is ensured for:
- REST - by comparing against the `openapi.json` spec
- gRPC - by comparing against the advertised methods in the reflection server

Exhaustiveness for within one endpoint (the ones supporting multiple operations) is not ensured yet.

Sorry for the huge PR! The main thing to review here is the `ACTION_ACCESS` table, which states what is supposed to happen with each of the permissions

### All Submissions:


timvisee (approved):
:raised_hands: 

I must admit that I skimmed all the individual test functions, but the overall structure looks well and good! I like that you've included a test ensuring all endpoints are covered.

tests/consensus_tests/test_jwt_validation.py
```diff
@@ -92,66 +570,1124 @@ def scroll_with_token(uri: str, collection: str, token: str) -> requests.Respons
...
+    if should_succeed:
+        assert res.status_code < 500 and res.status_code not in [
+            401,
+            403,
+        ], f"{method} {path} failed with {res.status_code}: {res.text}"
+    else:
+        assert res.status_code in [
+            401,
+            403,
+        ], f"{method} {path} failed with {res.status_code}: {res.text}"
```
xzfc, line 670: No need to check for 401 Unauthorized, this is for wrong/missing creds.
Check only for 403 Forbidden.
You are the one who suggested to distinguish them in the first place. 😅

Also, I suggest to reword the latter assertion message to indicate that it is *not* failed (but it should).

```suggestion
    if should_succeed:
        assert (
            res.status_code < 500 and res.status_code != 403
        ), f"{method} {path} failed with {res.status_code}: {res.text}"
    else:
        assert (
            res.status_code != 403
        ), f"{method} {path} was not forbidden with {res.status_code}: {res.text}"
```

xzfc (changes requested):

tests/consensus_tests/test_jwt_validation.py
coszio: I already removed the 401 check in #4008, I think it is fine to merge this as is to save some conflict resolution afterwards :)

I agree with the rewording, though. I'll edit the message in #4008 too.

xzfc (approved):

PR #3962, opened 2024-04-03
Add grey collection status

timvisee:
Add a 'grey' collection status color to identify:
- pending optimizations
- but no optimizations are running

This happens while an instance is restarted while optimizations are ongoing. In previous versions we triggered optimizations on start, but we don't do this anymore to prevent OOM crash loops. A new update operation is required to trigger optimizations once more.

Currently we just show a green collection status, which may be confusing. This grey status should help identify the above described situation, so that users can act accordingly.

### Compatibility
Due to compatibility reasons the grey status is not used right away. It is not compatible with our current clients. Until that is possible, we set an optimization error instead:

image

In Qdrant 1.10 we'll actually set the grey status, see the task below.

### Tasks
- [ ] Add documentation describing how to resolve this state (<https://github.com/qdrant/landing_page/pull/779>)
- [ ] Qdrant 1.10: Actually set grey status (see `// TODO(1.10):`)

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -885,7 +885,13 @@ impl LocalShard {
 
         // If still green while optimization conditions are triggered, mark as grey
         if self.update_handler.lock().await.has_pending_optimizations() {
-            status = CollectionStatus::Grey;
+            // TODO(1.10): enable grey status in Qdrant 1.10+
+            // status = CollectionStatus::Grey;
+            if optimizer_status == OptimizersStatus::Ok {
+                optimizer_status = OptimizersStatus::Error(
+                    "optimizations pending, awaiting update operation".into(),
+                );
+            }
         }
 
         CollectionInfoInternal {
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -884,7 +884,9 @@ impl LocalShard {
         }
 
         // If still green while optimization conditions are triggered, mark as grey
-        if self.update_handler.lock().await.has_pending_optimizations() {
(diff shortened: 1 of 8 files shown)


lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -846,39 +846,53 @@ impl LocalShard {
...
-        }
-        if !segments.failed_operation.is_empty() || segments.optimizer_errors.is_some() {
-            status = CollectionStatus::Red;
         }
 
-        let optimizer_status = match &segments.optimizer_errors {
-            None => OptimizersStatus::Ok,
-            Some(error) => OptimizersStatus::Error(error.to_string()),
-        };
+        // If still green while optimization conditions are triggered, mark as grey
```
generall, line 886: check for status is missing
Screenshot from 2024-04-03 16-26-04

generall (approved):

fastfists:
We had a cluster in gray status but didn't realize until we started adding more data into it. On restart it seemed like it restored itself since the status showed green.This is great thanks!

timvisee:
> We had a cluster in gray status but didn't realize until we started adding more data into it. This is great thanks!

We did not release this yet, so your cluster will not be showing a grey status until we release 1.9 or 1.10.

However, even before 1.9 you can trigger this situation if you restart your nodes while they were optimizing, but it won't be clearly reported. In that case it'll just show a green status, like you have above.

I did start on documentation for it. We'll release that part of the documentation along with Qdrant 1.9. Since it may be helpful for you, see the two paragraphs at the following link: https://deploy-preview-779--condescending-goldwasser-91acf0.netlify.app/documentation/concepts/collections/#collection-info

PR #3963, opened 2024-04-03
Multivec knob for SegmentConfig

agourlay:
This PR adds a configuration field to the internal `SegmentConfig` type in order to pick later the multi vector storage backend.

Although the type is internal, it is visible through the telemetry API, therefore this PR changes the openapi specification.

This consequence is acceptable because the telemetry schema is only used for documentation purposes.


lib/segment/src/types.rs
```diff
@@ -771,6 +771,9 @@ pub struct VectorDataConfig {
     pub index: Indexes,
     /// Vector specific quantization config that overrides collection config
     pub quantization_config: Option<QuantizationConfig>,
+    /// Enable multiple vectors per point
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub multivec: Option<bool>,
```
IvanPleshkov, line 776: I want to discuss `bool` type here. I know that this option is documented in the design document. But we can discuss the way to use some `enum` instead of bool for future extensions.

lib/segment/tests/integration/sparse_discover_test.rs
```diff
@@ -138,6 +138,7 @@ fn sparse_index_discover_test() {
                 storage_type: VectorStorageType::Memory,
                 index: Indexes::Plain {},
                 quantization_config: None,
+                multivec: None,
```
IvanPleshkov, line 141: Because we have a lot of default values here, I propose to use `..Default::default()` everywhere in this PR for all default fields

IvanPleshkov (approved):
I like this PR. approved. I proposed some design changes while we have the last chance to do it )

lib/segment/src/types.rs
agourlay: What kind of extensions do you have in mind?

IvanPleshkov: For instance,
```
pub struct VectorDataConfig {
    ...
    pub multivec: Option<MultivectorConfig>,
}

pub enum MultivectorConfig {
    #[default]
    Colbert(), // or some another placeholder name
}
```
It allows us to extend multivector parameters if needed.
Sorry that I propose it after opening this PR, I just realized it right now.

agourlay: Thanks for the suggestion, I have added an enum to configure the similarity for multivectors :+1:

lib/segment/tests/integration/sparse_discover_test.rs
agourlay: I am not sure about this TBH.
It is annoying but useful to see all the places impacted by a new field.

lib/segment/src/vector_storage/query_scorer/multi_custom_query_scorer.rs
```diff
@@ -57,8 +57,9 @@ impl<
 
     #[inline]
     fn score(&self, against: &MultiDenseVector) -> ScoreType {
-        self.query
-            .score_by(|example| score_multi::<TMetric>(example, against))
+        self.query.score_by(|example| {
```
IvanPleshkov, line 60: I like your providing config into scorer

IvanPleshkov (approved):

Issue #1519, opened 2023-03-03
Investigate Raft message batching


agourlay:
The raft library [raft.rs](https://github.com/tikv/raft-rs) supports the batching of `MsgAppend` which could have a positive impact on large clusters.

The feature is not explicitly documented, see https://github.com/tikv/raft-rs/issues/18

```
When a proposal is made, leader will broadcast it to all follower immediately. 
This can make proposal reach follower as fast as possible.
However, the message is only sent out when a ready is handled.
So this may cause too many MsgAppend messages in high load, which is very not friendly to channels.
```

It is easy to enable in `consensus.rs`

```diff
diff --git a/src/consensus.rs b/src/consensus.rs
@@ -191,6 +191,8 @@ impl Consensus {
             log::debug!("Local raft state found - skipping initialization");
         };
         let mut node = Node::new(&raft_config, state_ref.clone(), logger)?;
+        // whether to batch append msg at runtime.
+        node.set_batch_append(true);
         // Before consensus has started apply any unapplied committed entries
         // They might have not been applied due to unplanned Qdrant shutdown
         let _stop_consensus = state_ref.apply_entries(&mut node)?;
```

The change needs to be validated on clusters of different size to make sure it does not have a negative impact.

agourlay:
Tackled in https://github.com/qdrant/qdrant/pull/3964

PR #3964, opened 2024-04-03
Enable raft message batching (#1519)

ffuugoo:
This PR enables built-in raft message batching. (See #1519.)

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -969,7 +969,7 @@ impl RaftMessageBroker {
     }
 
     fn message_sender(&self) -> (RaftMessageSender, RaftMessageSenderHandle) {
-        let (messages_tx, messages_rx) = tokio::sync::mpsc::channel(256);
+        let (messages_tx, messages_rx) = tokio::sync::mpsc::channel(128);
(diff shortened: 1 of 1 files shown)


agourlay:
Could you elaborate a bit on the observable effects of this change? :chart_with_upwards_trend:

timvisee:
> The only possible concern would be to check if we have any limit on maximal size of a single gRPC message, maybe we should increase it.

I thought we didn't have a limit on the internal API, but we should confirm it.

ffuugoo:
> > The only possible concern would be to check if we have any limit on maximal size of a single gRPC message, maybe we should increase it.
> 
> I thought we didn't have a limit on the internal API, but we should confirm it.

Seems like [we don't](https://github.com/qdrant/qdrant/blob/9292dedb5c075231d99680d1ed27e91d2d18dc89/src/tonic/mod.rs#L345).

agourlay (approved):

timvisee (approved):

generall (approved):
I don't think would I have any insights without giving it a run on CI & chaos

PR #3977, opened 2024-04-05
test of stream shard transfer consistency

generall:
This integration test highlights the problem with stream shard transfer. Not 100% of the times, but reproducible locally

### DIFF
--- a/lib/collection/src/collection/point_ops.rs
+++ b/lib/collection/src/collection/point_ops.rs
@@ -117,7 +117,9 @@ impl Collection {
         if let Some(result) = result {
             Ok(result)
         } else {
-            Err(CollectionError::service_error(format!(
+            // Special error type needed to handle creation of partial shards
+            // In all other scenarios, equivalent to `service_error`
+            Err(CollectionError::pre_condition_failed(format!(
                 "No target shard {shard_selection} found for update"
             )))
         }
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -104,7 +104,9 @@ impl Collection {
 
                 replica_set.set_local(shard, Some(initial_state)).await?;
             } else {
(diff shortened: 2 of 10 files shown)


generall:
Found the reason for this test failing: when creating a shard transfer into a new replica, operation create State Partial for the remote shard, but doesn't create remote shard object itself. The fix is relatively simple, but creates a problem on it's own. 
If we create a remote shard on a new replica, which didn't exist before - it may produce a situation, where node A already received consensus operation, but node B, which should receive transfer, didn't. So it might be possible that we will send updates to a shard which doesn't have any local shard. If that happens - node will report an error and terminate the transfer, even though everything was working as expected.

Here are some of the possible ways to address this issue:

- ignore all errors from partial replicas (may potentially create inconsistency if failure matches the transition from partial to active)
- ignore only a specific type of errors from the partial (requires introduction of a new type of errors)
- introduce a special state that should be used before partial (too complicated)
- Force creating shard in dead (or some new) state before requesting transfer
- wait on all nodes for the receiver side to switch first (requires long operation in consensus loop)
- ???

generall:
> What about force-creating the shard in the initiate_transfer call. We currently use that to explicitly wait for the proper shard state on the remote. We can extend it to also ensure the actual shard is there.

Do you propose to make a call to all other nodes and make sure they created remote shard?

timvisee:
On another thought, I'm a bit puzzled this only happens with the stream transfer method.

Shouldn't the other transfer test, which the stream transfer test was based upon, suffer from the same issue?

generall:
> Shouldn't the other transfer test, which the stream transfer test was based upon, suffer from the same issue? Our other methods are switched into Recovery first, but once we switch into Partial another node may become Partial before our target node.


No, other transfer methods are guaranteed to have remote shard object in place, as they use two-step transition. This was the only place where we forgot to create remote shard object

timvisee (approved):

PR #3980, opened 2024-04-05
force wait for updates in forward proxy

generall:
This is required to make streaming shard transfer more reliable in cases where we send updates during transfer and those updates should be present in stream batch


timvisee (approved):
🤞

PR #3982, opened 2024-04-06
Applying the optimization options

kozistr:
### All Submissions:

### Change Logs

I made changes to the optimization option for release.

* using a single `codegen-unit`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -162,6 +162,13 @@ members = [
 
 [profile.release]
 lto = "fat"
+opt-level = 3
+codegen-units = 1
+strip = "symbols"
+panic = "abort"
+debug-assertions = false
+overflow-checks = false
+rpath = false
 
 # Inherit from release, because we are not rebuilding often,
 # and we don't want the huge binary sizes from debug builds.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -162,13 +162,7 @@ members = [
 
 [profile.release]
 lto = "fat"
-opt-level = 3
 codegen-units = 1
-strip = "symbols"
-panic = "abort"
-debug-assertions = false
-overflow-checks = false
-rpath = false
 
 # Inherit from release, because we are not rebuilding often,
 # and we don't want the huge binary sizes from debug builds.


generall:
Hey @kozistr, could you please explain why setting those options explicitly is required? AFAIK, most of them are already used by default.

kozistr:
> Hey @kozistr, could you please explain why setting those options explicitly is required? AFAIK, most of them are already used by default.

you're right. most of them are already used by default except `codegen-units`. I just initially thought applying those options explicitly would be good, but never mind. I'll remove them!

changed https://github.com/qdrant/qdrant/pull/3982/commits/fce1e2b5869f74c6edaf20abe78730789b55f18a

generall:
Still, could you please provide some explanation why we would want `codegen-unit` and what are the possible cons

kozistr:
> Still, could you please provide some explanation why we would want `codegen-unit` and what are the possible cons

sure! AFAIK, setting `codegen-units` to 1 could bring the performance gain or reduce the binary size although taking a longer compilation time (followed by [here](https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units)). I guess it might help to boost the Qdrant performance.

Maybe, if you've already considered this option and thought it wasn't suitable for the Qdrant project, please let me know. (I didn't run lots of tests, or know how `codegen-units` works deeply, so i might be wrong)

timvisee:
Reducing code-gen units to one does indeed open the door for more performance optimizations. The problem is that it'll make builds extremely slow. Qdrant builds already take a significant amount of time, and I don't think we want to see that as side effect.

generall:
> The problem is that it'll make building extremely slow.

How extremely? If we need to build it once per release, it is not that bad

AS1100K (approved):
Adding `codegen-units = 1` definitely will increase the performance but it comes with a cost of compile time.

For release profile this extra time is indeed not a big deal as it's not build frequently.

Compile time can also be reduced by modifying the value of `codegen-units`, but according to me for this codebase `codegen-units = 1` is just fine and is balanced between performance and compile time.

/claim #3982

timvisee:
On my machine I'm seeing 414 seconds versus 519 with a single codegen unit. It's less of a difference than I had anticipated.

AS1100K:
That's a good point about compile times varying based on system resources. It's possible GitHub Actions might have a larger difference due to their build environment.

Using `codegen-units=1` can improve performance, and small optimizations can add up. Here's something else to consider for future optimizations: exploring potential improvements to multi-threading.

generall:
> /claim https://github.com/qdrant/qdrant/pull/3982

I am mostly interested in search perf benchmarks

kozistr:
> > /claim #3982
> 
> I am mostly interested in search perf benchmarks

I ran some tests with the [bfb](https://github.com/qdrant/bfb)

* Environment
   * OS: Windows 10 (WSL 2)
   * CPU: i7-7700K 
   * Disk: Samsung SSD 970 EVO Plus 1TB
   * RAM: DDR4

* treatment: Qdrant 1.8.4
* experimental: Qdrant 1.8.4 with `codegen-units = 1`

## building index

* command line: `docker run -it --rm --network host bfb ./bfb --dim 1024`
    * dimension : `1024`

| group | times |
| :---: | :---: |
| treatment | 85.193107507 seconds |
| experimental | 76.164150314 seconds |

## Search

* Distance: Dot
* on-disk: (payload, index, vectors)

* command line : `docker run -it --rm --network host bfb ./bfb --dim 1024 --search --distance Dot --on-disk-payload --on-disk-index true --on-disk-vectors true`
    * Dimension: 1024 
    * Distance: Dot
    * on-disk: (payload, index, vectors)
    * The reason that I set it up like this is because it's similar to my production environment (on-disk & 1024 dims).

### Search timings

| group | min | avg | median | p95 | p99 | max |
| :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| treatment | 0.005384548 | 0.009714346868400054 | 0.009783974 | 0.01135127 | 0.012353613 | 0.019784106 |
| experimental | 0.00537742 | 0.009676725693100133 | 0.009701624 | 0.011258946 | 0.012252325 | 0.019721841 |

### Server timings

| group | min | avg | median | p95 | p99 | max |
| :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| treatment | 0.004888853 | 0.008754061682599942 | 0.008803368 | 0.010377731 | 0.011456847 | 0.019127279 |
| experimental | 0.00486561 | 0.008729234132379823 | 0.008736518 | 0.01030292 | 0.011354128 | 0.018932535 |

### RPS

| group | min | avg | median | max |
| :---: | :---: | :---: | :---: | :---: |
| treatment | 201.41468678122897 | 203.95397667696304 | 204.1001233179035 | 327.0230449873457 |
| experimental | 202.41011500729388 | 204.7785899851906 | 204.85674918334945 | 348.1159128477496 |

Although there are many configs to test and the result depends on the environment, it seems that `codegen-units = 1` gives a slightly better result in my experiments.

I'll do more tests if I'm available.

generall (approved):

PR #3984, opened 2024-04-07
[RBAC] remove option to specify multiple collection names in one rule

generall:
Simplify claim definition


xzfc (approved):
For the context, rationale behind this PR:
- The current structure is hard to represent in the UI.
- Multiple collections per rule were initially added to avoid bloating the JWT token, but there is little use in having multiple rules with the same `"payload"` rule.

PR #3990, opened 2024-04-08
Fix lock ordering in proxy segment

generall:
change order of locking to prevent a situation where point is marked as removed before moved


timvisee:
I did attempt this in my testing, but it does not solve the problem of missing points.

timvisee (approved):
Though it doesn't fix the issue we're currently hunting, we agreed that it's probably a good change to keep anyway.

PR #3996, opened 2024-04-09
On read operations, go over non-appendable segments first

timvisee:
Extends <https://github.com/qdrant/qdrant/pull/3995>

All segment reading operations must read over non-appendable segments first.

Points may be moved from non-appendable to appendable, because we don't lock all segments together read ordering is very important here.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -249,6 +249,22 @@ impl<'s> SegmentHolder {
         segments
     }
 
+    pub fn collect_non_appendable_segments(&self) -> Vec<LockedSegment> {
+        self.segments
+            .values()
+            .cloned()
+            .filter(|segment| !segment.get().read().is_appendable())
+            .collect()
+    }
+
+    pub fn collect_appendable_segments(&self) -> Vec<LockedSegment> {
+        self.segments
+            .values()
+            .cloned()
+            .filter(|segment| segment.get().read().is_appendable())
+            .collect()
+    }
+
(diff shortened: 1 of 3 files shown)


lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -115,9 +115,10 @@ impl LocalShard {
 
         let read_handles: Vec<_> = {
             let segments_guard = segments.read();
-            segments_guard
-                .iter()
-                .map(|(_, segment)| {
+            let segments = segments_guard.non_appendable_then_appendable_segments();
```
timvisee, line 118: We must enforce ordering here across `spawn_blocking`.

ffuugoo: Try to resolve this by making reads on all non-appendable segments first, and then on all appendable. Not 100% sure if that's the best/right solution, but feels like it should work for what we try to achieve here?

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -249,6 +249,15 @@ impl<'s> SegmentHolder {
         segments
     }
 
+    pub fn collect_non_appendable_and_appendable_segments(
+        &self,
+    ) -> (Vec<LockedSegment>, Vec<LockedSegment>) {
+        self.segments
+            .values()
+            .cloned()
+            .partition(|segment| !segment.get().read().is_appendable())
```
ffuugoo, line 258: 👨‍🍳🤌

ffuugoo (approved):

generall (approved):

PR #3998, opened 2024-04-09
Fix indexed-only behavior for proxy shard

generall:
This is an attempt to rethink a suggestion proposed in https://github.com/qdrant/qdrant/pull/3186

Instead of introducing a new parameter, we simply fix the problem of excluding proxy shard with too many changes from the search by allowing each individual raw segment to decide if it can do search or not.

P.S.
There is a follow-up PR with post-refactoring of simple `search` method.

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -300,6 +300,7 @@ impl SegmentEntry for ProxySegment {
         top: usize,
         params: Option<&SearchParams>,
         is_stopped: &AtomicBool,
+        search_optimized_threshold_kb: usize,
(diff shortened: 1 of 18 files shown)


keremnc:
@generall thanks for reviving this discussion.
I think there may still be a shortcoming with this approach - a search request with an extremely restrictive filter could skip a large plain segment, even though the cost of performing the plain search would be very little.
Maybe it makes sense to additionally consider the cardinality estimate of the query here?

timvisee (approved):

generall:
Indeed, moving it to the level of index would be a better option. Also resolves the bug with sparse vectors we had.

WDYT about the current version @keremnc ?

keremnc:
Thanks @generall, this looks good to me

timvisee (approved):

PR #4000, opened 2024-04-09
RBAC: Disallow write ops when there's a payload constraint

coszio:
Tracked in #3777 

When testing in #3961, I realized there is only a small subset of point update operations that are allowed, making it inconvenient to use. 

Here I propose to generalize to disallowing these operations for now. Making a collection **`rw`** access with `payload` constraint the same as a collection **`r`** access with `payload` constraint

### All Submissions:

### Changes to Core Features:


timvisee (approved):

coszio:
Closing: after discussion with the other feature peers, we decided to still allow a subset of operations

PR #4004, opened 2024-04-10
Flatten representation for MultiDenseVectors

agourlay:
Use a flatten representation for `MultiDenseVector` for mechanical sympathy.

It will help in the future for the mmap and quantization work as well.


IvanPleshkov (changes requested):
Request changes because of concern with `MultiDenseVector::default()`

IvanPleshkov (approved):

PR #4005, opened 2024-04-10
Introduce multi vector search benchmark

agourlay:
A new benchmark to validate the performance of the multivector search.

### DIFF
--- a/lib/segment/benches/multi_vector_search.rs
+++ b/lib/segment/benches/multi_vector_search.rs
@@ -1,0 +1,109 @@
+use std::collections::HashMap;
+use std::sync::atomic::AtomicBool;
+use std::sync::Arc;
+
+use common::cpu::CpuPermit;
+use criterion::{criterion_group, criterion_main, Criterion};
+use rand::prelude::StdRng;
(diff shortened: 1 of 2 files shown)


lib/segment/benches/multi_vector_search.rs
```diff
@@ -0,0 +1,109 @@
...
+        vector_storage.clone(),
+        quantized_vectors.clone(),
+        segment.payload_index.clone(),
+        hnsw_config,
+    )
+    .unwrap();
+
+    hnsw_index.build_index(permit.clone(), &stopped).unwrap();
+
+    let query = random_multi_vector(&mut rnd, VECTOR_DIM, NUM_VECTORS_PER_POINT).into();
```
IvanPleshkov, line 88: I don't know, is it a good idea to use the same query vector each time. There are pros and cons.
Pros: the same codepath can produce accurate and stable benchmark
Cons: same query may affect some reusages of caches etc
WDYT? I think that we can keep it as is if HNSW index is large enough

IvanPleshkov (approved):

PR #4006, opened 2024-04-10
RBAC: add tests

xzfc:
Tracked in #3777.

This PR adds units tests for operations check/rewrite rules.

Additionally this PR fixes a bug causing `payload` restriction to be ignored.

### DIFF
--- a/lib/storage/src/rbac/ops_checks.rs
+++ b/lib/storage/src/rbac/ops_checks.rs
@@ -14,6 +14,7 @@ use collection::operations::types::{
 use collection::operations::vector_ops::VectorOperations;
 use collection::operations::CollectionUpdateOperations;
 use segment::types::{Condition, ExtendedPointId, FieldCondition, Filter, Match, Payload};
+use strum::IntoEnumIterator as _;
 
 use super::{
     incompatible_with_payload_constraint, Access, AccessRequirements, CollectionAccessList,
@@ -525,8 +526,10 @@ mod tests_ops {
     use std::fmt::Debug;
 
     use api::rest::{BatchVectorStruct, VectorStruct};
+    use collection::operations::payload_ops::PayloadOpsDiscriminants;
     use collection::operations::point_ops::{
-        Batch, PointInsertOperationsInternal, PointStruct, PointSyncOperation,
+        Batch, PointInsertOperationsInternal, PointInsertOperationsInternalDiscriminants,
+        PointStruct, PointSyncOperation,
     };
     use collection::operations::types::{
         OrderByInterface, QueryEnum, RecommendStrategy, SearchRequestInternal, UsingVector,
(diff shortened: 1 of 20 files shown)


lib/storage/src/rbac/ops_checks.rs
```diff
@@ -457,3 +452,709 @@ impl PayloadConstraint {
...
+                .add("col", false, false)
+                .into(),
+            |op| {
+                op.filter = Some(PayloadConstraint::new_test("col").to_filter());
+            },
+        );
+    }
+
+    #[test]
+    fn test_collection_update_operations_upsert_points() {
```
coszio, line 976: How about taking advantage of having an ops enum + the [strum](https://docs.rs/strum_macros/latest/strum_macros/derive.EnumIter.html) crate to ensure all variants are tested?

xzfc: Added usage of `strum` to cover all variants of `CollectionUpdateOperations` and its nested enums.

lib/storage/src/rbac/ops_checks.rs
```diff
@@ -457,3 +452,773 @@ impl PayloadConstraint {
...
+                ..op.clone()
+            },
+            &AccessCollectionBuilder::new()
+                .add("col", false, true)
+                .into(),
+        );
+    }
+
+    #[test]
+    fn test_scroll_request_internal() {
```
coszio, line 688:
```suggestion
    fn test_point_request_internal() {
```

lib/storage/src/rbac/ops_checks.rs
```diff
@@ -457,3 +452,773 @@ impl PayloadConstraint {
...
+                ..op.clone()
+            },
+            &AccessCollectionBuilder::new()
+                .add("col", false, true)
+                .into(),
+        );
+    }
+
+    #[test]
+    fn check_test_scroll_request_internal() {
```
coszio, line 951:
```suggestion
    fn test_scroll_request_internal() {
```

lib/storage/src/rbac/ops_checks.rs
```diff
@@ -457,3 +452,773 @@ impl PayloadConstraint {
...
+    }
+
+    /// Tests for [`CollectionUpdateOperations::FieldIndexOperation`].
+    fn test_collection_update_operations_field_index() {
+        let op = CollectionUpdateOperations::FieldIndexOperation(
+            FieldIndexOperations::CreateIndex(CreateIndex {
+                field_name: "path".parse().unwrap(),
+                field_schema: None,
+            }),
+        );
```
coszio, line 1210: We can also use the same `strum` approach here: we have create and delete operations in the enum

lib/storage/src/rbac/ops_checks.rs
```diff
@@ -457,3 +452,773 @@ impl PayloadConstraint {
...
+            }
+            CollectionUpdateOperationsDiscriminants::VectorOperation => {
+                test_collection_update_operations_update_vectors()
+            }
+            CollectionUpdateOperationsDiscriminants::PayloadOperation => {
+                test_collection_update_operations_payload()
+            }
+            CollectionUpdateOperationsDiscriminants::FieldIndexOperation => {
+                test_collection_update_operations_field_index()
+            }
```
coszio, line 996: I suggest to use `check_` prefix instead of `test_` for these sub_tests, since they aren't technically "standalone" tests

coszio (changes requested):
Requesting changes only because `test_collection_update_operations_field_index` can also take advantage of the `strum` approach

coszio (approved):

PR #4007, opened 2024-04-10
Use idiomatic Rust naming

agourlay:
Rust rover is complaining that methods starting with `into_*` are by convention consuming their input.

This PR follows the naming convention and rename those methods


coszio (approved):
Something good about rust rover 😄

PR #4014, opened 2024-04-11
Decouple named vector struct

agourlay:
This PR decouples the `NamedVectorStruct` segment type from the API.

A new REST only type is introduced with the necessary conversions.

### DIFF
--- a/lib/segment/src/data_types/vectors.rs
+++ b/lib/segment/src/data_types/vectors.rs
@@ -410,9 +410,7 @@ pub struct NamedSparseVector {
     pub vector: SparseVector,
 }
 
-#[derive(Debug, Deserialize, Serialize, Clone)]
-#[serde(rename_all = "snake_case")]
-#[serde(untagged)]
+#[derive(Debug, Clone)]
 pub enum NamedVectorStruct {
     Default(DenseVector),
     Dense(NamedVector),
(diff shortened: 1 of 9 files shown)


lib/segment/src/data_types/vectors.rs
```diff
@@ -399,28 +410,14 @@ pub struct NamedSparseVector {
...
-/// or named mode:
-///
-/// {
-///   "vector": {
-///     "vector": [1.0, 2.0, 3.0],
-///     "name": "image-embeddings"
-///   }
-/// }
-#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+#[derive(Debug, Deserialize, Serialize, Clone)]
```
IvanPleshkov, line 413: Do we need `Deserialize` and `Serialize`?

IvanPleshkov (approved):

lib/segment/src/data_types/vectors.rs
agourlay: very good catch :+1: 

https://github.com/qdrant/qdrant/pull/4014/commits/ff0b47db1a5c8f2475a1889253fef6d17087fe95

IvanPleshkov (approved):

PR #4017, opened 2024-04-11
RBAC: Restructure jwt tests

coszio:
Purely a refactoring change: restructures `test_jwt_validation.py` into two parts:
- one for keeping only the access table and its tests
- another one for regular validation (`value_exists`, `payload` constraint, etc)

This will also make it easier to move the `api-key` tests into `tests/consensus_tests/auth_tests`


tests/consensus_tests/auth_tests/test_jwt_validation.py
```diff
@@ -1,623 +1,37 @@
...
-
-    fixtures.upsert_random_points(
-        REST_URI, 100, COLL_NAME, shard_key=SHARD_KEY, headers=API_KEY_HEADERS
-    )
+    fixtures.upsert_random_points(REST_URI, 100, COLL_NAME, headers=API_KEY_HEADERS)
 
     yield peer_api_uris, peer_dirs, bootstrap_uri
 
     fixtures.drop_collection(REST_URI, COLL_NAME, headers=API_KEY_HEADERS)
+    kill_all_processes()
```
xzfc, line 34: We need to kill processes even if `drop_collection` throws an exception.

Also, maybe it's worth to put a fixture with `start_jwt_protected_cluster()` + `kill_all_processes()` into `utils.py`? Then this `setup()` could look like this:

```python
from .utils import jwt_setup
@pytest.fixture(scope="module", autouse=True)
def setup(
    jwt_setup # force the execution order
):
    fixtures.create_collection(
        REST_URI,
        collection=COLL_NAME,
        headers=API_KEY_HEADERS,
    )

    fixtures.upsert_random_points(REST_URI, 100, COLL_NAME, headers=API_KEY_HEADERS)
    yield
    fixtures.drop_collection(REST_URI, COLL_NAME, headers=API_KEY_HEADERS)
    # processes are killed inside jwt_setup fixture
```

xzfc (changes requested):

tests/consensus_tests/auth_tests/test_jwt_validation.py
coszio: I had tried something similar, which didn't work because of temp files. But now I've found a better and cleaner solution with your suggestion :)

xzfc (approved):

PR #4028, opened 2024-04-12
Update docs to mention safer production practices

mac-chaffee:
### All Submissions:

Currently, the docs refer to unauthenticated instances of Qdrant as "for production" when we actually discourage that: https://qdrant.tech/documentation/guides/security/

So in the README and QUICK_START guides, I've removed those mentions of "production" and instead included mentions of safer ways to run Qdrant in production.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ client = QdrantClient(path="path/to/db")  # Persists changes to disk, fast proto
 
 ### Client-Server
 
-This is the recommended method for production usage. To run the container, use the command:
+To run the container, use the command:
 
 ```bash
 docker run -p 6333:6333 qdrant/qdrant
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -50,7 +50,7 @@ client = QdrantClient(path="path/to/db")  # Persists changes to disk, fast proto
 
 ### Client-Server
 
-This is the recommended method for production usage. To run the container, use the command:
+To run the container, use the command:
```
NirantK, line 53: I'd prefer to have some sort of recommendation e.g. prefer client-server for production usage because we do get questions on why is ":memory:" mode not fast enough or similar

NirantK (approved):
Looks good to me, no specific changes that I'd insist on

README.md
mac-chaffee: I could make a separate "Production Deployment" section right underneath the "Client-Server" section, would that help?

Or I could leave the "recommended for production" phrase and just update this existing section to match QUICK_START.md, which mentions enabling authentication and `--read-only`.

Do you have a preference?

NirantK: I have a mild preference for a separate Production Deployment section — but both solutions work for the problem to be solved here.

mac-chaffee: I tried a mixture of the two ideas, since it seemed like a new section would be easily overlooked. Let me know if it's better!

NirantK (approved):

PR #4031, opened 2024-04-12
Support for an HTTP publish port

pnivlek:
When running Qdrant inside an orchestrator, the port published for clients to connect to a peer can be different than the host port bound to, i.e. `docker run -p 26333:6333`. 

Before v1.7.0 and shard snapshot transfers, this was fine, because Qdrant never referenced its own HTTP port towards itself. However, [shard snapshot transfers refer to a peers own port](https://github.com/qdrant/qdrant/blob/v1.8.4/lib/collection/src/shards/transfer/snapshot.rs#L174) when transferring data to another snapshot, so without a publish port a remote shard is unable to be transferred.

A gRPC port is not needed as of now because a node never references its own gRPC server. p2p is not needed because nodes connect to each other using --uri, which simply contains the published p2p port.

I've tested this PR to confirm replication still works with and without the published port.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -166,6 +166,9 @@ service:
   # HTTP(S) port to bind the service on
   http_port: 6333
 
+  # HTTP(S) port to publish to the world
+  http_publish_port: 6333
+
   # gRPC port to bind the service on.
   # If `null` - gRPC is disabled. Default: null
   # Comment to disable gRPC:
@@ -193,7 +196,7 @@ service:
   #
   # Uncomment to enable.
   # api_key: your_secret_api_key_here
-   
+
   # Set an api-key for read-only operations.
   # If set, all requests must include a header with the api-key.
   # example header: `api-key: <API-KEY>`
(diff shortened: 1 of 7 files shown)


generall:
Hey @pnivlek, thanks for highlighting this issue!

I have one question: since we still require to change configuration for published port, why won't we just change the default http port to match the published one?

Which advantages does this new configuration option provides?

pnivlek:
Hi @generall - good point, I'd completely overlooked that.

To confirm we're on the same page (or on the off chance someone has the same issue in the future): since you can change the http port (especially with an environment variable), the dynamic port can be the internal port as well. This would mean from my example `-p 26333:26333` and `QDRANT__SERVICE__HTTP_PORT=26333`, with 26333 randomly generated by a separate program/the orchestrator. The only scenario I can see the config option being needed then is if a cluster is running behind separate load balancers per peer, which seems like a really really bad idea. Feel free to close unless you can think of another scenario.

PR #4033, opened 2024-04-13
Fix snapshot tar symlinks validation

generall:
add validation for snapshot archives for not containing any symlinks

### DIFF
--- a/lib/segment/src/common/validate_snapshot_archive.rs
+++ b/lib/segment/src/common/validate_snapshot_archive.rs
@@ -1,0 +1,40 @@
+use std::fs::File;
+use std::path::Path;
+
+use tar::Archive;
+
+use crate::common::operation_error::{OperationError, OperationResult};
+
+pub fn open_snapshot_archive_with_validation<P: AsRef<Path>>(
+    snapshot_path: P,
+) -> OperationResult<Archive<File>> {
+    let path = snapshot_path.as_ref();
+    {
+        let archive_file = File::open(path).map_err(|err| {
+            OperationError::service_error(format!(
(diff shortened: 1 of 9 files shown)


lib/segment/src/common/validate_snapshot_archive.rs
```diff
@@ -0,0 +1,40 @@
...
+        let archive_file = File::open(path).map_err(|err| {
+            OperationError::service_error(format!(
+                "failed to open segment snapshot archive {path:?}: {err}"
+            ))
+        })?;
+        let mut ar = Archive::new(archive_file);
+
+        for entry in ar.entries_with_seek()? {
+            let entry = entry?;
+            if entry.header().entry_type() == tar::EntryType::Symlink {
```
timvisee, line 22: This may still contain a hard link and other weird file types.

Though I don't know what possible attack vectors they may introduce, should we just allow [`Regular`](https://docs.rs/tar/latest/tar/enum.EntryType.html#variant.Regular) and [`Directory`](https://docs.rs/tar/latest/tar/enum.EntryType.html#variant.Directory) instead to stay on the safe side?

timvisee (changes requested):

timvisee:
I've also just confirmed that [`tar`](https://docs.rs/tar/) does not support `..` in paths :+1:

lib/segment/src/common/validate_snapshot_archive.rs
timvisee: @generall I've pushed another commit on top to handle this

timvisee (approved):

mac-chaffee (approved):

PR #4038, opened 2024-04-15
RBAC: Make read-only api key behave the same as global read-only token

coszio:
Tracked by #3777 

There were some endpoints in gRPC, which were not allowed by read-only api key, but allowed by the global `read` access token. 

With this we make them behave the same, and we get prepared to remove the `is_read_only` implementation entirely which can be substituted it with an Access struct

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/src/actix/auth.rs
+++ b/src/actix/auth.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use actix_web::body::{BoxBody, EitherBody};
 use actix_web::dev::{forward_ready, Service, ServiceRequest, ServiceResponse, Transform};
 use actix_web::http::Method;
-use actix_web::{Error, FromRequest, HttpMessage as _, HttpResponse, ResponseError};
+use actix_web::{Error, FromRequest, HttpMessage, HttpResponse, ResponseError};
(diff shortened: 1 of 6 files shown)


src/actix/auth.rs
```diff
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use actix_web::body::{BoxBody, EitherBody};
 use actix_web::dev::{forward_ready, Service, ServiceRequest, ServiceResponse, Transform};
 use actix_web::http::Method;
-use actix_web::{Error, FromRequest, HttpMessage as _, HttpResponse, ResponseError};
```
xzfc, line 8: :pick: Nitpick: I write `use … as _` for traits to avoid polluting the namespace with unused identifiers. But I don't have a strong opinion about that.

xzfc (approved):

src/actix/auth.rs
coszio: I think this was from some `fmt`. But my opinion is that it should not matter unless there are conflicts or a shorter name is more ergonomic.

PR #4047, opened 2024-04-16
Fix average sparse vector recommendation

agourlay:
The recommendation API must sort the sparse vectors before computing the average because the `combine_aggregate` function requires sorted vectors.

Issue discovered while running the congruence tests locally with debug assertions enabled.

### DIFF
--- a/lib/collection/src/recommendations.rs
+++ b/lib/collection/src/recommendations.rs
@@ -46,7 +46,7 @@ fn avg_vectors<'a>(vectors: impl Iterator<Item = VectorRef<'a>>) -> CollectionRe
             }
             VectorRef::Sparse(vector) => {
                 sparse_count += 1;
-                avg_sparse = vector.combine_aggregate(&avg_sparse, |v1, v2| v1 + v2);
+                avg_sparse.combine_aggregate(vector, |v1, v2| v1 + v2);
             }
             VectorRef::MultiDense(_) => {
                 // TODO(colbert)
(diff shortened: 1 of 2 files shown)


lib/sparse/src/common/sparse_vector.rs
```diff
@@ -87,6 +87,7 @@ impl SparseVector {
     }
 
     /// Construct a new vector that is the result of performing all indices-wise operations
+    /// Warning: Expects both vectors to be sorted by indices.
```
IvanPleshkov, line 90: we can check while iteration that indices are growing

lib/collection/src/recommendations.rs
```diff
@@ -290,15 +290,15 @@ fn recommend_by_avg_vector(
 
     let lookup_collection_name = lookup_from.as_ref().map(|x| &x.collection);
 
-    let positive_vectors = convert_to_vectors(
-        positive.iter(),
+    let positive_vectors = convert_to_vectors_owned(
```
IvanPleshkov, line 293: It affects regular dense vector and do unnecessary copying. We can fix this bug by changing `combine_aggregate` and allow to do this function with any order of indices. In that case we can avoid any changes in this file

IvanPleshkov (changes requested):
Great finding! Thanks for fixing, it was my fault (
I requested changes because PR affects dense vectors case

lib/collection/src/recommendations.rs
agourlay: Which exact copying are you referring to?
Here I am passing ownership of the vector and consuming it.

IvanPleshkov: `convert_to_vectors` returns `impl Iterator<Item = VectorRef<'a>>` without ownership and we use it in dense case for avg calculation with allocation for result vector only. Here I see a copying of each positive vector in dense case where it's not required

lib/sparse/src/common/sparse_vector.rs
IvanPleshkov: I find that there are debug_assertion for that. @agourlay do you know why the debug assertions wasn't triggered?

agourlay: the assertion is triggered on a debug build for the Python congruence tests :)

agourlay:
The new approach is all green here but breaks the congruence test.

```
FAILED tests/congruence_tests/test_sparse_recommend.py::test_simple_recommend - AssertionError: point1[0].id = 151, point2[0].id = 394
```

I need to investigate :mag:

IvanPleshkov:
> @IvanPleshkov I pushed a new version that does not break the congruence tests.
> 
> TBH I could not find the exact issue, so I went with a simpler implementation without in-place mutation for the result in the argument.

okay, approve. the easiest way to fix

IvanPleshkov (approved):

PR #4049, opened 2024-04-16
Byte storage integration into segment

IvanPleshkov:
This PR adds byte vector storage support into segment.

Main changes in this PR to do integration:
1. Add `datatype` field into segment config (into `VectorDataConfig`). It allows storing the type of storage and loading segment properly
2. Add new fields into `VectorStorageEnum`: `DenseSimpleByte`, `DenseMemmapByte` and `DenseAppendableMemmapByte`.
3. Add in segment constructor support of changed config and of new vector storages types.
4. Remove `Distance::preprocess_vector` and `Distance::similarity`. `Distance` does not describe which type of vector is processing so it cannot trigger functions from the generic `Metric` trait.
5. Removing `Distance::preprocess_vector` triggers large refactor in quantization scorers.
6. Changes in raw scorer constructors, adding byte type support.
7. Unit test which checks plain search, HNSW with filters, discovery search, recommendations and compare search over byte-storaged segment with original one

### All Submissions:

### New Feature Submissions:

8. [ ] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
9. [ ] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:


lib/segment/src/vector_storage/query_scorer/metric_query_scorer.rs
```diff
@@ -30,8 +30,9 @@ impl<
         query: TypedDenseVector<VectorElementType>,
         vector_storage: &'a TVectorStorage,
     ) -> Self {
+        let preprocessed_vector = TMetric::preprocess(query);
         Self {
-            query: TMetric::preprocess(query),
+            query: TElement::slice_from_float_cow(&preprocessed_vector).to_vec(),
```
generall, line 35: It is good to know, that Cow won't actually invoke copy for owned vectors

IvanPleshkov: Fixed: https://github.com/qdrant/qdrant/pull/4049/commits/0edd6304ebbcecfe2bb52b428f464e70d04306f5
Now we don't do unnecessary copying. In this place for float and non-cosine distance there is no any reallocation anymore

generall (approved):

PR #4053, opened 2024-04-16
refactor segment holder to separate appendable and non-appendable seg…

generall:
…ments in advance

I find it a bit better than doing filter each time.
In particular, new approach doesn't require locking of all segments in query time to determine which of them are append-able or not.

### DIFF
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -6,7 +6,6 @@ use std::sync::Arc;
 use std::thread::sleep;
 use std::time::Duration;
 
-use itertools::Itertools;
 use parking_lot::{RwLock, RwLockReadGuard, RwLockUpgradableReadGuard, RwLockWriteGuard};
 use rand::seq::SliceRandom;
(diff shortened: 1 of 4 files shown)


lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -174,34 +177,47 @@ impl<'s> SegmentHolder {
...
 
+    fn _insert(&mut self, segment: LockedSegment) -> SegmentId {
+        let key = self.generate_new_key();
+        if segment.get().read().is_appendable() {
+            self.appendable_segments.insert(key, segment);
+        } else {
+            self.non_appendable_segments.insert(key, segment);
+        }
+        key
+    }
```
timvisee, line 199: What about putting this just in `add()` rather than in a separate internal method. The function signature is the same, `add()` doesn't have any special logic.

timvisee (approved):
Great idea!

lib/collection/src/collection_manager/holders/segment_holder.rs
generall: :thinking:

PR #4054, opened 2024-04-16
[order-by] Fix remote shard calling

### LINKED ISSUE #4024: order_by is not working in distributed deployment (multi-node cluster)
## Current Behavior

in V1.8.0, qdrant introduced order_by in scroll api, tested in one node and official cloud demo cluster, it worked well,
while when deployed a distributed deployment with multiple nodes (e.g. 5 nodes), the scroll api retured points is not restrictly sorted by the order_by fields, when reduce the cluster's node number, the result became more sequential though.

## Steps to Reproduce

1. deploy a multiple node deployment
2. create a test collection
3. add index, to one field (e.g. test_int_index)
4. add 10 points
5. call scroll api with order_by="index", the results is not sorted as expected.

## Expected Behavior

the scroll results should be sorted by the sort_by field.

coszio:
Fixes #4024

The root cause of the issue is that the remote shards were not actually sending the `____ordered_with____` value, because we were removing it right before sending it.

This fixes 2 main things:
- deserialization of this value into `OrderingValue`
- detects when it is a remote shard call to not erase the `____ordered_with____` value early. Then uglily adds the requested `with_payload` after merging from the shards.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/data_types/order_by.rs
+++ b/lib/segment/src/data_types/order_by.rs
@@ -93,8 +93,8 @@ impl OrderBy {
                 StartFrom::Datetime(dt) => OrderingValue::Int(dt.timestamp()),
             })
             .unwrap_or_else(|| match self.direction() {
-                Direction::Asc => OrderingValue::Int(std::i64::MIN),
-                Direction::Desc => OrderingValue::Int(std::i64::MAX),
+                Direction::Asc => OrderingValue::MIN,
+                Direction::Desc => OrderingValue::MAX,
             })
     }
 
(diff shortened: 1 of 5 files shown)


tests/consensus_tests/test_order_by.py
```diff
@@ -0,0 +1,145 @@
...
+        },
+    ).raise_for_status()
+
+    res = requests.post(
+        f"{uri}/collections/{COLL_NAME}/points/scroll", json={"limit": 10, "order_by": "index", "with_payload": False}
+    )
+    
+    res.raise_for_status()
+    
+    ids = [point["id"] for point in res.json()['result']["points"]]
```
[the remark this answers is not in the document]
tellet-q: @coszio you can view suggested changes in https://github.com/qdrant/qdrant/pull/4056

tellet-q (changes requested):

coszio:
Silly mistake, fixed the default `with_payload` value in https://github.com/qdrant/qdrant/pull/4054/commits/c3f1fcf645b8f4d914ce47c913845e964f738268

tellet-q (approved):

lib/segment/src/data_types/order_by.rs
```diff
@@ -109,22 +109,28 @@ impl OrderBy {
...
 }
 
+#[derive(Debug)]
 pub enum OrderingValue {
     Float(FloatPayloadType),
     Int(IntPayloadType),
 }
 
+impl OrderingValue {
+    const MAX: Self = Self::Float(f64::NAN);
```
generall, line 130: NAN -> MAX?

coszio: According to out `Ord` implementation it is the highest value possible

generall (approved):

PR #4058, opened 2024-04-17
Always flush both `wrapped` and `write` segments during `ProxySegment::flush`

ffuugoo:
This should fix the issue, when `ProxySegment::flush` returns `self.wrapped_segment.get().read().version()` as flushed version, which is *not* guaranteed to be actually flushed.

Returned flushed version is used to truncate the WAL, so in this case, if Qdrant is restarted, it may skip some operations from `wrapped_segment` that weren't flushed (but were reported *as* flushed).

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -1,4 +1,4 @@
-use std::cmp::max;
+use std::cmp;
 use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::AtomicBool;
@@ -247,7 +247,7 @@ impl ProxySegment {
 
 impl SegmentEntry for ProxySegment {
     fn version(&self) -> SeqNumberType {
-        max(
+        cmp::max(
             self.wrapped_segment.get().read().version(),
             self.write_segment.get().read().version(),
         )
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -703,27 +703,22 @@ impl SegmentEntry for ProxySegment {
...
         } else {
-            let _write_segment_version = self.write_segment.get().read().flush(sync)?;
-            // If intermediate state is not empty - that is possible that some changes are not persisted
-            Ok(self
-                .last_flushed_version
-                .read()
-                .unwrap_or_else(|| self.wrapped_segment.get().read().version()))
-        }
+            cmp::min(write_segment_version, wrapped_version)
+        };
```
ffuugoo, line 717: I'm not 100% sure if this is required, but my logic was:

If we have *no* data in `deleted_points`/`deleted_indexes`/etc, then there are no pending changes to `wrapped_segment`/`optimized_segment`. We can safely `ack` `write_segment_version`, because all changes were applied to `write_segment` only, `wrapped_segment` weren't affected at all.

If we *do* have in-memory data, it means some of the applied operations applied to proxy segment should have modified `wrapped_segment` (we just couldn't apply them, because `wrapped_segment` is locked by optimizer), so if we crash, we might want to reapply these operations to the `wrapped_segment`, so we ack `wrapped_version` (which should always be older than `write_segment_version`).

An example of second case is a simple `point_delete` operation, that would just delete the point from `wrapped_segment`. If we always ack `write_segment_version`, then after restart we will still have the point in `wrapped_segment`, but we already "acked" `point_delete` operation, so it won't be replayed on the `wrapped_segment`.

timvisee (approved):

generall (approved):

PR #4060, opened 2024-04-17
Test point version handling and moving correctness across segments

timvisee:
Add two tests to assert correctness of point version handling and moving points across segments, typically with an old operation ID and a newer segment version.

Before <https://github.com/qdrant/qdrant/pull/4057>, these did show wrong behavior when doing a point operation with an operation ID that is older than the current segment version.

In the case of moving points across segments, it resulted in missing points. A move happens if we want to update a point in a non-appendable segment. We copy to a new segment, and delete it from the old segment. If the copy was (silently) rejected due to a version mismatch, the point is deleted and ends up missing.

The tests assert correctness in two places to make sure the same doesn't happen in the future.

The tests now succeed, while they fail before <https://github.com/qdrant/qdrant/pull/4057>.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/segment.rs
+++ b/lib/segment/src/segment.rs
@@ -2542,4 +2542,66 @@ mod tests {
                 .unwrap();
         }
     }
+
+    /// Test handling point versions
+    ///
+    /// Apply if the point version is equal or higher. Always apply if the point does not exist
+    /// yet.
+    #[test]
+    fn test_handle_point_version() {
+        // Create base segment with a single point
+        let dir = Builder::new().prefix("segment_dir").tempdir().unwrap();
+        let dim = 4;
+        let config = SegmentConfig {
+            vector_data: HashMap::from([(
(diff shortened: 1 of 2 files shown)


lib/segment/src/segment.rs
```diff
@@ -2542,4 +2542,70 @@ mod tests {
...
+
+        // Do handle operation on existing point when providing the current version
+        let applied = segment
+            .handle_point_version(100, Some(0), |_segment| Ok((true, None)))
+            .unwrap();
+        assert!(applied);
+
+        // Do handle operation on existing point when providing a newer version
+        let applied = segment
+            .handle_point_version(100, Some(0), |_segment| Ok((true, None)))
```
generall, line 2593:
```suggestion
            .handle_point_version(101, Some(0), |_segment| Ok((true, None)))
```

?

PR #4068, opened 2024-04-18
gRPC API for multivector

agourlay:
This PR adds support for multivector in the gRPC API.

### DIFF
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -47,8 +47,9 @@ message SparseIndices {
 }
 
 message Vector {
-  repeated float data = 1;
-  optional SparseIndices indices = 2;
+  repeated float data = 1; // Vector data (flatten for multi-dimensional vectors)
+  optional SparseIndices indices = 2; // Sparse indices for sparse vectors
+  optional uint32 tokens_count = 3; // Number of tokens in the vector
 }
 
 // ---------------------------------------------
@@ -267,6 +268,7 @@ message SearchPoints {
   optional uint64 timeout = 13; // If set, overrides global timeout setting for this request. Unit is seconds.
   optional ShardKeySelector shard_key_selector = 14; // Specify in which shards to look for the points, if not specified - look in all shards
   optional SparseIndices sparse_indices = 15;
+  optional uint32 tokens_count = 16;
 }
 
 message SearchBatchPoints {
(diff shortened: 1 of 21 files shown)


lib/api/src/grpc/proto/points.proto
```diff
@@ -47,8 +47,9 @@ message SparseIndices {
 }
 
 message Vector {
-  repeated float data = 1;
-  optional SparseIndices indices = 2;
+  repeated float data = 1; // Vector data (flatten for multi-dimensional vectors)
+  optional SparseIndices indices = 2; // Sparse indices for sparse vectors
+  optional uint32 tokens_count = 3; // Number of tokens in the vector
```
generall, line 52: tokens is ColBERT specific terminology. It is possible that this multi-vector functionality will be used for other purposes then ColBERT (e.g. jsut storing embeddings of multiple images of the same object), so let's call it more generic

lib/api/src/grpc/proto/points.proto
```diff
@@ -300,6 +302,7 @@ message SearchPointGroups {
   optional uint64 timeout = 14; // If set, overrides global timeout setting for this request. Unit is seconds.
   optional ShardKeySelector shard_key_selector = 15; // Specify in which shards to look for the points, if not specified - look in all shards
   optional SparseIndices sparse_indices = 16;
+  optional uint32 tokens_count = 17;
```
generall, line 305: let's avoid changing search API

lib/api/src/grpc/proto/points.proto
```diff
@@ -267,6 +268,7 @@ message SearchPoints {
   optional uint64 timeout = 13; // If set, overrides global timeout setting for this request. Unit is seconds.
   optional ShardKeySelector shard_key_selector = 14; // Specify in which shards to look for the points, if not specified - look in all shards
   optional SparseIndices sparse_indices = 15;
+  optional uint32 tokens_count = 16;
```
IvanPleshkov, line 271: Colbert design document covered this message and `SearchPointGroups`. The user has to send a flattened vector and `SearchPoint`/`SearchPointGroups` have to be unchanged. It's possible to convert from flattened `vector` into colbert vectors using collection DIM

IvanPleshkov (changes requested):

agourlay:
Thanks for the feedback, I am closing this PR because it won't be merged as is.

We have decided to not implement multivector support in search for the existing REST and gRPC API.

I will propose a new PR with the write and retrieve paths only REST and gRPC.

PR #4086, opened 2024-04-22
Issues API: Event-driven issue framework

coszio:
Tracked by #3471 

This PR introduces type safety into the `issues` internals in two ways:

1. All issues now have to be matched on their type, so that `UnindexedField` issues can only remove `UnindexedField` issues from the dashboard.
2. Introduction of an `EventBroker`, which keeps track of subscribers to specific events, and notifies them when such event has occured. This helps keeping the codebase less polluted from this feature.

### All Submissions:

### New Feature Submissions:

5. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### DIFF
--- a/lib/common/issues/src/broker.rs
+++ b/lib/common/issues/src/broker.rs
@@ -79,23 +79,13 @@ mod tests {
     use super::*;
     use crate::issue::DummyIssue;
 
-    struct DummySubscriber<E> {
-        _phantom: std::marker::PhantomData<E>,
-    }
+    struct DummySubscriber;
 
     struct DummyEvent {
         pub collection_id: String,
     }
 
-    impl<E> DummySubscriber<E> {
-        pub fn new() -> Self {
-            Self {
-                _phantom: std::marker::PhantomData,
-            }
-        }
-    }
-
-    impl Subscriber<DummyEvent> for DummySubscriber<DummyEvent> {
+    impl Subscriber<DummyEvent> for DummySubscriber {
         fn notify(&self, event: Arc<DummyEvent>) {
             let issue = DummyIssue::new(event.collection_id.clone());
             crate::submit(issue);
@@ -106,7 +96,7 @@ mod tests {
         collection_id: String,
     }
 
-    impl Subscriber<CollectionDeletedEvent> for DummySubscriber<CollectionDeletedEvent> {
+    impl Subscriber<CollectionDeletedEvent> for DummySubscriber {
         fn notify(&self, event: Arc<CollectionDeletedEvent>) {
             crate::solve_by_filter::<DummyIssue, _>(|code| code == &event.collection_id);
         }
@@ -116,8 +106,8 @@ mod tests {
     fn test_basic_use() {
         let mut broker = EventBroker::new();
 
-        broker.add_subscriber(Box::new(DummySubscriber::<DummyEvent>::new()));
-        broker.add_subscriber(Box::new(DummySubscriber::<CollectionDeletedEvent>::new()));
+        broker.add_subscriber::<DummyEvent>(Box::new(DummySubscriber));
+        broker.add_subscriber::<CollectionDeletedEvent>(Box::new(DummySubscriber));
 
         broker.notify(DummyEvent {
             collection_id: "dummy".to_string(),
--- a/lib/common/issues/src/broker.rs
+++ b/lib/common/issues/src/broker.rs
@@ -84,7 +84,7 @@ mod tests {
 
     impl Subscriber<CollectionDeletedEvent> for DummySubscriber {
         fn notify(&self, event: Arc<CollectionDeletedEvent>) {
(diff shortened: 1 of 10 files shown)


lib/common/issues/src/broker.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        assert!(
+            crate::all_issues()
+                .iter()
+                .all(|issue| issue.code != "dummy"),
+            "{:?}",
+            crate::all_issues()
+        );
+    }
+}
```
coszio, line 142: This is how the usage of these changes look like

lib/common/issues/src/broker.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+    pub fn add_subscriber<E: 'static>(&mut self, subscriber: DynSubscriber<E>) {
+        self.subscriptions.push(subscriber);
+    }
+
+    /// Notify all subscribers of the event. Spawns one tokio task per subscriber, so that it can be completed in the background.
+    pub fn notify_async<E: Send + Sync + 'static>(&self, event: E) {
+        if let Some(subscribers) = self.subscriptions.get::<E>() {
+            let event = Arc::new(event);
+            for sub in subscribers {
```
agourlay, line 52: How many subscribers do we expect in general?

lib/common/issues/src/dashboard.rs
```diff
@@ -7,22 +9,32 @@ use crate::issue::{CodeType, Issue, IssueRecord};
 #[derive(Default)]
 struct Dashboard {
     pub issues: DashMap<CodeType, IssueRecord>,
+    inverted_index: DashMap<TypeId, HashSet<CodeType>>,
```
agourlay, line 12: I think this inverted index is a bit overkill given that there will be often a very low number of issues.

coszio: How about using a composite `Code` type instead of a plain string (https://github.com/qdrant/qdrant/pull/4086/commits/3827ebbcb47bad353abb6423d6e18ebec2d41255)?

agourlay: I like the structured `Code` better :+1:

lib/common/issues/src/broker.rs
```diff
@@ -0,0 +1,132 @@
...
+        Self {
+            subscriptions: SubscriberMap::new(),
+        }
+    }
+    pub fn add_subscriber<E: 'static>(&mut self, subscriber: DynSubscriber<E>) {
+        self.subscriptions.push(subscriber);
+    }
+
+    /// Notify all subscribers of the event. Spawns one tokio task per subscriber, so that it can be completed in the background.
+    pub fn publish_async<E: Send + Sync + 'static>(&self, event: E) {
```
agourlay, line 49: will you remove `publish_async` after our conversation?

it seems all `notify` can run async. computation in the background.

coszio: Yeah I think we can remove it for now. Ideally I'd want this kind of `publish_async` to return a handle so that we can choose whether we want to await for all the notifications to be handled or not. But that sounds like a refactor/discussion we can have later.

We are not using this function anywhere right now :)

lib/common/issues/src/issue.rs
```diff
@@ -4,30 +4,53 @@ use chrono::{DateTime, Utc};
...
+use crate::dashboard::Code;
 use crate::solution::Solution;
 
-/// Type of the issue code
-pub type CodeType = String;
-
 pub trait Issue {
-    fn code(&self) -> CodeType;
+    /// Differentiates issues of the same type
+    fn distinctive(&self) -> &str;
```
agourlay, line 12: I think we need a better name here, `distinctive` is not working out for my brain.

What about `issue_instance_id`?
The idea being that they are class of issues with different particular instances.
WDYT?

lib/common/issues/src/issue.rs
```diff
@@ -4,30 +4,53 @@ use chrono::{DateTime, Utc};
...
+
+    /// Actionable solution to the issue
     fn solution(&self) -> Solution;
+
+    /// Internal code for the issue
+    fn code(&self) -> Code
+    where
+        Self: std::marker::Sized + 'static,
+    {
+        Code::new::<Self>(self.distinctive())
```
agourlay, line 28: it feels a bit weird to me to have a child field embedding part of the parent object.

coszio: Good point. I'll move this to the `Code` implementation

lib/common/issues/src/issue.rs
coszio: After thinking about it I could not come up with a better name than `instance_id`, renamed 👍

agourlay (approved):
This PR represents a nice improvement in the architecture :+1: 

I think it fine to continue further discussion regarding the String based identity based on the integration in #3511

PR #4098, opened 2024-04-23
Fix debug assertion, replaced points may be less than points to update

timvisee:
Fix a debug assertion that fails quite often locally. This happens on stream records transfer with concurrent upsertions.

Our sync points logic is a multi step process. We select a list of points to be replaced, and eventually update all of them. When updating, we expect the number of replaced points to match the list of points we selected.

When concurrent upsertions are ongoing, some of these points may get updated in between this process. Those points will have a newer version and will not be updated. That results in a lower number of replaced points.

Roman and I investigated this some time ago. We believe this is expected, and we should relax the debug assertion as done in this PR.

### All Submissions:

### New Feature Submissions:


ffuugoo (approved):
Original assert didn't make sense to me, this seems more reasonable, IMO.

Issue #4102, opened 2024-04-24
Geo Bounding Box Antimeridian Failure


almostimplemented:
When a bounding box crosses the 180 meridian line, the simple `check_point` implementation fails.

## Current Behavior + Steps to Reproduce
We noticed this error for user queries where "USA" was predicted as a location filter constraint from user input. 

We use Google Maps geocoding API. In Qdrant's top-left / bottom-right syntax:

```python
((74.071028, 167), (18.7763, -66.885417))
```

Here is the current logic:

```rust
impl GeoBoundingBox {
    pub fn check_point(&self, point: &GeoPoint) -> bool {
        (self.top_left.lon < point.lon)
            && (point.lon < self.bottom_right.lon)
            && (self.bottom_right.lat < point.lat)
            && (point.lat < self.top_left.lat)
    }
}
```

Consider this GPS coordinate in Los Angeles, CA: 
```python
(34.010714, -118.232922) 
```

This will fail because `167` is not less than `-118.232922`. 

## Expected Behavior
This should return True, Los Angeles is within this bounding box.

## Possible Solution
The situation of a bounding box crossing the antimeridian requires special handling. I will open a PR with a fix.

agourlay:
released in [1.9.1](https://github.com/qdrant/qdrant/releases/tag/v1.9.1)

PR #4103, opened 2024-04-24
Handle bounding box crossing the 180th meridian

almostimplemented:
### All Submissions:

* [x] Does your submission pass tests?
* [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
* [ ] Have you checked your code using `cargo clippy --all --all-features` command?

I am not a Rust programmer. I would appreciate community assistance in adding a unit test file! We can use the reproduction example in #4102

### DIFF
--- a/lib/segment/src/index/field_index/geo_hash.rs
+++ b/lib/segment/src/index/field_index/geo_hash.rs
@@ -583,6 +583,51 @@ mod tests {
     }
 
     #[test]
+    fn rectangle_hashes_crossing_antimeridian() {
+        // conversion to lon/lat http://geohash.co/
+        // "ztnv2hjxn03k"
+        let top_left = GeoPoint {
+            lat: 74.071028,
+            lon: 167.0,
+        };
+
+        // "dr5ru7c02wnv"
+        let bottom_right = GeoPoint {
+            lat: 40.75798,
+            lon: -73.991516,
+        };
+
+        let crossing_usa_rectangle = GeoBoundingBox {
+            top_left,
+            bottom_right,
+        };
+
+        let usa_hashes_result = rectangle_hashes(&crossing_usa_rectangle, 200);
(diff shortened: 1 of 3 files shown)


agourlay:
@almostimplemented thanks for the contribution :+1: 

Regarding adding a unit test, please have a look at the file `geo_hash.rs`, it contains some practical example of intersection and box checking using real cities.

generall:
FYI: check_point is not the only place where this logic will be handled. We also have geohash-based index, which should be  tested (and modified) accordingly.

almostimplemented:
@generall Can you provide some guidance for the necessary changes for the geohash code? Is that fix tightly coupled with this change? 

My team will be detecting this edge-case and splitting the bounding box into two rectangles (on either side of the 180th meridian), and I am not sure I have the bandwidth to address the geohash index changes.

generall:
hey @almostimplemented, sure this is understandable, we don't expect you to do all the work;) my comment was addressed to whoever reviews this PR to make sure it is not merged without full support.

agourlay:
@almostimplemented Thanks for you work, I will double check the geohash indexing and add some tests.

agourlay:
@almostimplemented  I took the liberty of pushing tests into your PR with https://github.com/qdrant/qdrant/pull/4103/commits/97d1040e6465e71318266b267dbf83431fe712ad

My conclusion is that your fix the bug globally.

Our Geohash index links geohashes to the recorded points and uses the `GeoBoundingBox::check_point` function to filter points present in the corresponding geo hashes.

See https://github.com/qdrant/qdrant/blob/b99d507496a40b0895fff76eb2fa7a5cfdb701a9/lib/segment/src/index/field_index/geo_index/mod.rs#L358-L373

agourlay (approved):

timvisee (approved):

PR #4105, opened 2024-04-24
tests: Add test on low disk

tellet-q:
1. Add a new test with the following scenario: "Run Qdrant container with limited disk amount and verify that it doesn't crash when disk is running low during points insertion."
2. Rename CI workflow `test-low-ram` into `test-low-resources`. Run the new test as a part of it. Note: the test fails right now as Qdrant container doesn't handle OOD properly. 

Checklist


generall:
merged as a part of https://github.com/qdrant/qdrant/pull/4165

PR #4106, opened 2024-04-24
Mmap multivector storage

IvanPleshkov:
This PR introduces memory map storage for multidense vectors.

It extends `VectorStorageEnum` with new variant:
```
MultiDenseAppendableMemmap(AppendableMmapMultiDenseVectorStorage<VectorElementType>)
```
This is a generic appendable mmap with `f32` type. `u8` type is possible but will be introduced later as a separate PR with a supporting byte multivector.

We are going to release an appendable mmap only without regular mmap. Appendable mmap covers all use cases of regular mmap but with more file descriptors. We find it ok.

### Changes in existing code
I reused `ChunkedMmapVectors` structure from regular appendable mmap. This structure works with regular dense vectors. I extended the API of this structure and add support to read/write bunch of flattened vectors.
https://github.com/qdrant/qdrant/pull/4106/files#diff-cded970cc52c5614f1c851876649bcc6a35127062395a8bc753b6e11b805b2daR162
https://github.com/qdrant/qdrant/pull/4106/files#diff-cded970cc52c5614f1c851876649bcc6a35127062395a8bc753b6e11b805b2daR219
https://github.com/qdrant/qdrant/pull/4106/files#diff-cded970cc52c5614f1c851876649bcc6a35127062395a8bc753b6e11b805b2daR204

### Work of `AppendableMmapMultiDenseVectorStorage`
Because each multivector has different amount of inner vector, we cannot use `ChunkedMmapVectors` storage directly. We need to have a remap from internal ID into ID of start vector in `ChunkedMmapVectors` and retrieve full flattened inner data.
This remap is a metadata of multivector. Also, as multivector metadata we have size (inner vectors amount) and allocated vectors amount (description is above).
All this metadata is stored in another appendable mmap file (see `AppendableMmapMultiDenseVectorStorage.offsets`).

When user adds new vector, we just extend `ChunkedMmapVectors` and add new metadata. The allocated size is the same as amount of inner vectors.
When user rewrites the multivector, we can rewrite vector data in the same place when the new vector is not larger that the currently allocated place. Size will be changed, allocated size won't. If new vector is larger, do as for new vector and become hole in file. It should be shrunk after optimizers.

In PR comments, I marked the place, where we guarantee that flattened multivector data stores in one chunk only.
Also, we have no shrinking in `update_from` method. I will introduce shrinking in the next PR.

### Unit tests
Unit tests are the same as for RAM multivector storage. I find tests for RAM great and reused it, thanks @agourlay


agourlay:
@IvanPleshkov please add a PR description with the important choices you made and the things I might pay extra attention to :pray:

lib/segment/src/vector_storage/chunked_mmap_vectors.rs
```diff
@@ -178,15 +216,32 @@ impl<T: PrimitiveVectorElement + 'static> ChunkedMmapVectors<T> {
         Ok(new_id)
     }
 
-    pub fn get<TKey>(&self, key: TKey) -> &[T]
+    pub fn get<TKey>(&self, key: TKey) -> Option<&[T]>
```
IvanPleshkov, line 219: I made it optional because user may send the wrong key. Internal we just call new API for many vectors

lib/segment/src/vector_storage/chunked_mmap_vectors.rs
```diff
@@ -150,10 +149,39 @@ impl<T: PrimitiveVectorElement + 'static> ChunkedMmapVectors<T> {
         Ok(())
     }
 
-    pub fn insert(&mut self, key: PointOffsetType, vector: &[T]) -> OperationResult<()> {
-        let key = key as usize;
-        let chunk_idx = self.get_chunk_index(key);
-        let chunk_offset = self.get_chunk_offset(key);
+    pub fn insert<TKey>(&mut self, key: TKey, vector: &[T]) -> OperationResult<()>
```
IvanPleshkov, line 152: I changed implementation only (reuse many vectors API) and made it more generic

lib/segment/src/vector_storage/chunked_mmap_vectors.rs
```diff
@@ -150,10 +149,39 @@ impl<T: PrimitiveVectorElement + 'static> ChunkedMmapVectors<T> {
...
+    where
+        TKey: num_traits::cast::AsPrimitive<usize>,
+    {
+        assert!(vector.len() == self.config.dim, "Vector size mismatch");
+        let is_inserted = self.insert_many(key, vector, 1)?;
+        assert!(is_inserted, "Insertion failed");
+        Ok(())
+    }
+
+    pub fn insert_many<TKey>(
```
IvanPleshkov, line 162: Implementation is modified one from `insert`

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -0,0 +1,271 @@
...
+    stopped: &AtomicBool,
+) -> OperationResult<AppendableMmapMultiDenseVectorStorage<T>> {
+    create_dir_all(path)?;
+
+    let vectors_path = path.join(VECTORS_DIR_PATH);
+    let offsets_path = path.join(OFFSETS_DIR_PATH);
+    let deleted_path = path.join(DELETED_DIR_PATH);
+
+    let vectors = ChunkedMmapVectors::open(&vectors_path, dim)?;
+    let offsets = ChunkedMmapVectors::open(&offsets_path, 1)?;
```
IvanPleshkov, line 78: Ugly moment. I want to reuse extendable mmap for `offsets`. But we have only 1 offset for each multivector. So, I use then `ChunkedMmapVectors` too but with dimension `1`

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -0,0 +1,271 @@
...
+        let multi_vector = multi_vector.as_ref();
+        assert_eq!(multi_vector.dim, self.vectors.dim());
+
+        let mut offset = self
+            .offsets
+            .get(key as usize)
+            .map(|x| x.first().copied().unwrap_or_default())
+            .unwrap_or_default();
+
+        if multi_vector.len() > offset.reserved as usize {
```
IvanPleshkov, line 185: here we check, reuse old allocated place or make a hole and allocate new one

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -0,0 +1,271 @@
...
+        distance,
+        multi_vector_config,
+        deleted_count,
+    })
+}
+
+impl<T: PrimitiveVectorElement + 'static> AppendableMmapMultiDenseVectorStorage<T> {
+    /// Set deleted flag for given key. Returns previous deleted state.
+    #[inline]
+    fn set_deleted(&mut self, key: PointOffsetType, deleted: bool) -> OperationResult<bool> {
```
IvanPleshkov, line 106: Same as in dense appendable mmap storage

lib/segment/src/vector_storage/tests/test_appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -0,0 +1,267 @@
...
+        4,
+        Distance::Dot,
+        MultiVectorConfig::default(),
+        &AtomicBool::new(false),
+    )
+    .unwrap();
+}
+
+#[test]
+fn test_update_from_delete_points_simple_multi_dense_vector_storage() {
```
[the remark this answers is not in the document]
IvanPleshkov: I thinks so, let me try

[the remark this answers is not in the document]
IvanPleshkov: done

agourlay:
adding Tim as a reviewer because I am not familiar with the internals of the `ChunkedMmapVectors` :)

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
agourlay: Do you think using `ChunkedMmapVectors` with a dimension of 1 is actually a hack?
What would be an alternative? Building a simpler appendable mmap abstraction without any chunking?

IvanPleshkov:
```
Building a simpler appendable mmap abstraction without any chunking?
```
Yep, it's an alternative

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -0,0 +1,271 @@
...
+    let deleted: DynamicMmapFlags = DynamicMmapFlags::open(&deleted_path)?;
+
+    let mut deleted_count = 0;
+
+    for i in 0..num_vectors {
+        if deleted.get(i) {
+            deleted_count += 1;
+        }
+        check_process_stopped(stopped)?;
+    }
```
timvisee, line 91: I'm not sure if we have the same logic in a different place, but we might want to optimize this with a more efficient call for bit counting.

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
timvisee: I must admit I don't really like this approach, but I don't have a better suggestion.

`ChunkedMmapVectors` could probably benefit from another round of refactoring to make it a bit more generic (and not vector specific according to it's name). That would probably make it better. It's not something to worry about in this PR though.

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
IvanPleshkov: It's a similar approach like in dense appendable mmap storage:
https://github.com/qdrant/qdrant/blob/dev/lib/segment/src/vector_storage/dense/appendable_mmap_dense_vector_storage.rs#L81
I guess better fix it later (in two places) if we notice that storage loading is a bottleneck

timvisee: > if we notice that storage loading is a bottleneck

Well, we are trying to profile slow loading storage now. I did notice this when using mmap's, though I'm not sure if it's related to this specifically.

generall (approved):

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
timvisee: I have explored this idea here: https://github.com/qdrant/qdrant/pull/4176

If we merge it before this PR, we can simply change this to `deleted.count_flags()`.

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -0,0 +1,271 @@
...
+
+const VECTORS_DIR_PATH: &str = "vectors";
+const OFFSETS_DIR_PATH: &str = "offsets";
+const DELETED_DIR_PATH: &str = "deleted";
+
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+struct MultivectorMmapOffset {
+    offset: PointOffsetType,
+    count: PointOffsetType,
+    reserved: PointOffsetType,
```
agourlay, line 32: `capacity` maybe?

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -0,0 +1,271 @@
...
+                offset: new_key as PointOffsetType,
+                count: multi_vector.len() as PointOffsetType,
+                reserved: multi_vector.len() as PointOffsetType,
+            };
+        } else {
+            // use existing place to insert vector
+            offset.count = multi_vector.len() as PointOffsetType;
+        }
+
+        self.offsets.insert(key as usize, &[offset])?;
```
agourlay, line 203: what happened if we crash between updating `offsets` and `vectors`

lib/segment/src/vector_storage/tests/test_appendable_multi_dense_vector_storage.rs
```diff
@@ -217,62 +226,65 @@ fn do_test_update_from_delete_points(storage: Arc<AtomicRefCell<VectorStorageEnu
...
-    let _storage = open_simple_multi_dense_vector_storage(
-        db,
-        DB_VECTOR_CF,
-        4,
-        Distance::Dot,
-        MultiVectorConfig::default(),
-        &AtomicBool::new(false),
-    )
-    .unwrap();
+    let _storage = create_vector_storage(storage_type, dir.path());
```
agourlay, line 289: can we retrieve the number of points before the drop and then retrieve after re-open to make sure that the persistence works

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
timvisee: The above has been merged, so this one can be updated.

lib/segment/src/vector_storage/chunked_mmap_vectors.rs
```diff
@@ -150,10 +149,39 @@ impl<T: PrimitiveVectorElement + 'static> ChunkedMmapVectors<T> {
...
 
-    pub fn insert(&mut self, key: PointOffsetType, vector: &[T]) -> OperationResult<()> {
-        let key = key as usize;
-        let chunk_idx = self.get_chunk_index(key);
-        let chunk_offset = self.get_chunk_offset(key);
+    pub fn insert<TKey>(&mut self, key: TKey, vector: &[T]) -> OperationResult<()>
+    where
+        TKey: num_traits::cast::AsPrimitive<usize>,
+    {
+        assert!(vector.len() == self.config.dim, "Vector size mismatch");
```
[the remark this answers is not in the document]
IvanPleshkov: Done

lib/segment/src/vector_storage/chunked_mmap_vectors.rs
```diff
@@ -150,10 +149,39 @@ impl<T: PrimitiveVectorElement + 'static> ChunkedMmapVectors<T> {
...
+        let is_inserted = self.insert_many(key, vector, 1)?;
+        assert!(is_inserted, "Insertion failed");
```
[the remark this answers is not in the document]
IvanPleshkov: I don't use `bool` from `insert_many`. So I removed this check and `insert_many` returns error instead of `false` if vectors don't fit chunk

lib/segment/src/vector_storage/chunked_mmap_vectors.rs
```diff
@@ -162,14 +190,24 @@ impl<T: PrimitiveVectorElement + 'static> ChunkedMmapVectors<T> {
...
 
         if new_len > self.status.len {
             self.status.len = new_len;
         }
-        Ok(())
+        Ok(true)
+    }
+
+    // returns how many vectors can be inserted starting from key
+    pub fn get_chunk_left_keys<TKey>(&mut self, start_key: TKey) -> usize
```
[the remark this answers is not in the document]
IvanPleshkov: Thanks, done

lib/segment/src/vector_storage/chunked_mmap_vectors.rs
```diff
@@ -178,15 +216,32 @@ impl<T: PrimitiveVectorElement + 'static> ChunkedMmapVectors<T> {
...
-    pub fn get<TKey>(&self, key: TKey) -> &[T]
+    pub fn get<TKey>(&self, key: TKey) -> Option<&[T]>
+    where
+        TKey: num_traits::cast::AsPrimitive<usize>,
+    {
+        self.get_many(key, 1)
+    }
+
+    // returns count flattened vectors starting from key. if chunk boundary is crossed, returns None
+    pub fn get_many<TKey>(&self, start_key: TKey, count: usize) -> Option<&[T]>
```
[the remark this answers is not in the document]
IvanPleshkov: Done

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
IvanPleshkov: Thanks for your great improvement. Changed here and use your counting now

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
IvanPleshkov: Done

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
IvanPleshkov: Corrupted storage will happend. I changed the order of insertion. I insert into `vectors` first and then into `offsets`. Thanks for this finding!

Issue #3957, opened 2024-04-02
Add properties to a collection


Waffleboy:
**Is your feature request related to a problem? Please describe.**
If theres any information that applies table-wide, theres currently no way to add it in unless we add it as a metadata to every row. This causes space wastage

**Describe the solution you'd like**
Ability to add any properties to the table. eg,

```python
collection = client.get_collection("example_tablename")
print(collection.properties) # --> gives dictionary of properties. default {}

collection.set_property("main_column", "column a") --> sets it in property

print(collection.properties) # --> {"main_column":"column a"}
```

**Describe alternatives you've considered**
Currently adding as a column metadata instead (so every row has a static value)

**Additional context**

generall:
Hey @Waffleboy, could you please describe your use-case a bit more - why would you need those values collection-wide?

Also, would it be possible to use special collection which would hold per-collection info instead?

Waffleboy:
Hi @generall , 

There are some metadata we want to add to the table, for example which is the main date column to filter by.

generall:
Ok, have you considered making another collection with collection-level-metadata? It seems that you would need to make 2 requests to the service to make this filtering requests anyway

Waffleboy:
How would that work? The collection has one row with the payload being the collection-level-metadata?

Rows need a vector, so some dummy vector?

Seems very inelegant, and there wouldnt be a duplicate call as Im already calling `collection = lient.get_collection(table_name)` in code, I expect the properties to be inside that collection object?

timvisee:
> How would that work? The collection has one row with the payload being the collection-level-metadata?

Yes.

> Rows need a vector, so some dummy vector?

You can configure any vector size, and upload a point with a `null` vector.

> Seems very inelegant, and there wouldnt be a duplicate call as Im already calling collection=client.get_collection(table_name) in code, I expect the properties to be inside that collection object?

It is, but its what is already supported in our current releases.

Since we've seen requests for this multiple times, it's something we might add later down the line. But currently we have higher priority tasks to focus on, so it's not on our road map in the short term.

generall:
> You can configure any vector size, and upload a point with a null vector.


you can actually configure no vectors at all.

generall:
tbh, I don't see any advantages of the per-collection meta. Maybe you could provide a sample usage with "pseudocode" of how  you would expect it to work

gshashi:
[...]

PR #4107, opened 2024-04-24
add 'comment' field to collection apis (#3957)

abhishek818:
helps in providing table metadata

fixes: #3957 
/claim #3957 
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/api/src/grpc/proto/collections.proto
+++ b/lib/api/src/grpc/proto/collections.proto
@@ -322,7 +322,7 @@ message CreateCollection {
   optional QuantizationConfig quantization_config = 14; // Quantization configuration of vector
   optional ShardingMethod sharding_method = 15; // Sharding method
   optional SparseVectorConfig sparse_vectors_config = 16; // Configuration for sparse vectors
-  optional string comment = 17; // Properties/metadata of the collection
+  optional string comment = 17; // Properties or metadata of the collection
 }
 
 message UpdateCollection {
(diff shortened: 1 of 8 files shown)


abhishek818:
@generall  Up for review.

lib/api/src/grpc/proto/collections.proto
```diff
@@ -316,6 +317,7 @@ message UpdateCollection {
   optional VectorsConfigDiff vectors_config = 6; // New vector parameters
   optional QuantizationConfigDiff quantization_config = 7; // Quantization configuration of vector
   optional SparseVectorConfig sparse_vectors_config = 8; // New sparse vector parameters
+  map<string, string> properties = 9; // New properties of the collection
```
generall, line 320: I would prefer to have a single optional string called `comment`

abhishek818: resolved.

lib/collection/src/operations/types.rs
```diff
@@ -129,6 +129,8 @@ pub struct CollectionInfo {
     pub config: CollectionConfig,
     /// Types of stored payload
     pub payload_schema: HashMap<PayloadKeyType, PayloadIndexInfo>,
+    /// Properties or metadata of the collection
+    pub comment: String,
```
generall, line 133: To make it backward compatible, let's make this field optional and skip serialization if none

generall:
Hey @abhishek818, I believe the last step would be to trigger API schema gen https://github.com/qdrant/qdrant/blob/master/docs/DEVELOPMENT.md#rest

timvisee:
@generall Do you think a comment field would be sufficient? I'd expect an interface that is a bit more fully-fledged; something like the payload API but on collection level.

generall:
> I'd expect an interface that is a bit more fully-fledged; something like the payload API but on collection level.

I would prefer to not over-complicate this. As I understood - the main use-case is o give more information about the collection than it would normally fit into the name. 

If we would need payload-like data in the collection, we can make another change

generall (changes requested):
This PR requires ability to update the comment. And it doesn't look the comment even persisted

generall:
Will re-iterate over this later

Issue #3531, opened 2024-02-05
Better error response for wrong datetime format in REST filter


timvisee:
**Is your feature request related to a problem? Please describe.**
We recently merged <https://github.com/qdrant/qdrant/pull/3395> which adds a datetime payload index.

Currently, the datetime parser is very strict, only allowing [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339) formats in our REST API. A common format such as `YYYY-MM-DD HH:MM:SS` is currently not accepted.

If you'd send the following request:

```json
POST collections/test_collection/points/scroll
{
  "limit": 10,
  "filter": {
    "must": [
      {
        "key": "updated",
        "range": {
          "gt": "2014-01-01T00:00:00BAD"
        }
      }
    ]
  }
}
```

You'd receive the following error:

```
Format error in JSON body: data did not match any variant of untagged enum Condition at line 1 column 96
```

This error is very confusing.

**Describe the solution you'd like**
We'd strongly prefer a more descriptive error message instead.

Something like this would be a lot better:

```
Format error in JSON body: '2014-01-01T00:00:00' does not match any accepted datetime format
```

**Describe alternatives you've considered**
_None_

**Additional context**
Related issue: <https://github.com/qdrant/qdrant/issues/3529>

kartikchoubisa:
Much needed. I'd rather have it parse multiple formats, but for now I don't mind having at least a better error message.
I could take up this issue

generall:
@kartikchoubisa it is already merged

Pushkarm029:
Can I work on this one? I think @gsaivinay is no longer interested in working on it.

Pushkarm029:
Since, no one is responding. I think i am free to work on this.

/attempt #3531

timvisee:
Yes, please feel free!

Note that this seems to be quite difficult to achieve. We don't want to make significant changes in our deserializing logic just to make this happen.

But if you think this can be done without much changes, please feel free to give it a shot in a PR.

Pushkarm029:
It seems this is an upstream issue: collecting errors when deserializing untagged enums is not currently supported by `serde-rs`, and there are no plans to add this feature by the `serde-rs` maintainers.

I also tried using [serde_path_to_error](https://docs.rs/serde_path_to_error/latest/serde_path_to_error/), but it didn't work as expected.

If this issue becomes a higher priority in the future, we could consider adding tags for these enums to work around the limitation.

Closing my PR #5164.

Reference:
[...]

PR #4111, opened 2024-04-24
fix: Better error response for wrong datetime format in REST filter

gsaivinay:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

/claim #3531


lib/segment/src/types.rs
```diff
@@ -1877,6 +1942,51 @@ pub enum Condition {
...
+            serde_json::Value::Object(map) => {
+                if map.contains_key("key") {
+                    FieldCondition::deserialize(serde_json::Value::Object(map))
+                        .map(Condition::Field)
+                        .map_err(serde::de::Error::custom)
+                } else if map.contains_key("is_empty") {
+                    IsEmptyCondition::deserialize(serde_json::Value::Object(map))
+                        .map(Condition::IsEmpty)
+                        .map_err(serde::de::Error::custom)
+                } else if map.contains_key("is_null") {
```
generall, line 1961: with if-else statements it will be very easy to forget to implement some parts of the deserializer if we introduce new field here, as compiler won't show any errors

gsaivinay: @generall - fair enough, I'll try to change it, do you suggest any approach?

timvisee: A `match` statement with strict enum variants for example. Though to be honest, I don't clearly see how such approach can be used here.

gsaivinay: >  I don't clearly see how such approach can be used here

yes that is true
.


> strict enum variants

But isn't that requires introducing a new enum, and we might face the `did not match any variant` again ?

generall (changes requested):
See comment above. The variants of the enum should be traceable by the compiler

PR #4037, opened 2024-04-15
Median based TopK for sparse vectors scoring

agourlay:
Experiment median TopK based on the article from the nice folks at Quickwit https://quickwit.io/blog/top-k-complexity.

# Performance

## Criterion 

```
sparse-vector-search-group/inverted-index-search
                        time:   [566.19 µs 575.56 µs 584.71 µs]
                        change: [-9.0722% -7.3381% -5.5563%] (p = 0.00 < 0.05)
                        Performance has improved.
```

## End to end

Running https://github.com/qdrant/sparse-vectors-benchmark with full dataset

```
DEV
min: 72.6 millis
50p: 157.3 millis
95p: 210.72 millis
99p: 235.89 millis
999p: 261.57 millis
max: 338.19 millis
```

```
PR
min: 62.32 millis
50p: 148.66 millis
95p: 197.56 millis
99p: 222.33 millis
999p: 246.36 millis
max: 273.89 millis
```

# Validation

- checked congruence tests
- checked ground truth values in benchmarks

# Concerns

We might want to constraint the `limit` argument in search to avoid having someone reserving a huge amount of memory?

PR #4114, opened 2024-04-25
feat: Introduce top-k algo to HNSW

KShivendu:
HNSW equivalent of https://github.com/qdrant/qdrant/pull/4037 PR.

### New Feature Submissions:

### Todo:
- [ ] Benchmark against prev implementation

### DIFF
--- a/lib/common/common/src/top_k.rs
+++ b/lib/common/common/src/top_k.rs
@@ -38,18 +38,22 @@ impl TopK {
         self.threshold
     }
 
+    pub fn prune_elements(&mut self) -> Vec<ScoredPointOffset> {
+        let (_, median_el, _) = self
+            .elements
+            .select_nth_unstable_by(self.k - 1, |a, b| b.cmp(a));
+        self.threshold = median_el.score;
+        self.elements.split_off(self.k)
+    }
+
     // Push an element to the top k.
-    // Sometimes it ejects (returns) pushed elements.
+    // It ejects (returns) low scoring elements when full
     pub fn push(&mut self, element: ScoredPointOffset) -> Option<Vec<ScoredPointOffset>> {
         if element.score > self.threshold {
             self.elements.push(element);
-            // eject half the elements if full
+            // prune half the elements if full
             if self.elements.len() == self.k * 2 {
-                let (_, median_el, _) = self
-                    .elements
-                    .select_nth_unstable_by(self.k - 1, |a, b| b.cmp(a));
-                self.threshold = median_el.score;
-                return Some(self.elements.split_off(self.k));
+                return Some(self.prune_elements());
             }
             return None;
         }
@@ -62,8 +66,9 @@ impl TopK {
         }
     }
 
-    pub fn into_vec(self) -> Vec<ScoredPointOffset> {
-        self.into_iter().collect()
+    pub fn into_vec(mut self) -> Vec<ScoredPointOffset> {
+        self.prune_elements();
+        self.elements
     }
 }
 
@@ -80,7 +85,8 @@ impl IntoIterator for TopK {
     type Item = ScoredPointOffset;
     type IntoIter = std::vec::IntoIter<ScoredPointOffset>;
 
-    fn into_iter(self) -> Self::IntoIter {
+    fn into_iter(mut self) -> Self::IntoIter {
+        self.prune_elements();
         self.elements.into_iter()
     }
 }
--- a/lib/common/common/src/top_k.rs
+++ b/lib/common/common/src/top_k.rs
@@ -45,31 +45,55 @@ impl TopK {
             self.elements.push(element);
             // eject half the elements if full
             if self.elements.len() == self.k * 2 {
-                let position_to_sort = self.k.min(self.elements.len());
                 let (_, median_el, _) = self
                     .elements
-                    .select_nth_unstable_by(position_to_sort - 1, |a, b| b.cmp(a));
+                    .select_nth_unstable_by(self.k - 1, |a, b| b.cmp(a));
                 self.threshold = median_el.score;
-                return Some(self.elements.split_off(position_to_sort));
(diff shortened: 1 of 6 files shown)


lib/common/common/src/top_k.rs
```diff
@@ -11,7 +11,7 @@ use crate::types::{ScoreType, ScoredPointOffset};
 #[derive(Default)]
 pub struct TopK {
     k: usize,
-    elements: Vec<Reverse<ScoredPointOffset>>,
+    pub elements: Vec<Reverse<ScoredPointOffset>>,
```
generall, line 14: I would prefer not to expose this.
You can either return reference to a slice in a method, or deconstruct the whole thing into vector if you need ownership

lib/common/common/src/top_k.rs
```diff
@@ -34,27 +34,37 @@ impl TopK {
...
 
-    pub fn into_vec(mut self) -> Vec<ScoredPointOffset> {
+    pub fn truncate(&mut self) {
         self.elements.sort_unstable();
         self.elements.truncate(self.k);
+    }
+
+    pub fn into_vec(mut self) -> Vec<ScoredPointOffset> {
+        self.truncate();
         self.elements.into_iter().map(|Reverse(x)| x).collect()
```
generall, line 68: I am not sure if compiler is able to optimize this, but just to be extra sure, I would prefer to avoid using Reverse. There is a `select_nth_unstable_by` function, which should help

lib/common/common/src/top_k.rs
KShivendu: Fixed.

lib/common/common/src/top_k.rs
KShivendu: Done

lib/segment/src/index/hnsw_index/search_context.rs
```diff
@@ -24,18 +22,17 @@ impl SearchContext {
...
 
     /// Updates search context with new scored point.
     /// If it is closer than existing - also add it to candidates for further search
     pub fn process_candidate(&mut self, score_point: ScoredPointOffset) {
         let was_added = match self.nearest.push(score_point) {
             None => true,
-            Some(removed) => removed.idx != score_point.idx,
+            Some(removed) => removed
+                .iter()
+                .any(|removed_point| removed_point.idx == score_point.idx),
```
KShivendu, line 35: I think it would be faster to not check this and just let it go through next iteration so that:
```
if candidate.score < searcher.lower_bound() {
    break;
}
```
takes care of it. Wdyt?

lib/segment/src/index/hnsw_index/graph_layers_builder.rs
```diff
@@ -419,6 +420,7 @@ impl GraphLayersBuilder {
                             }
                         }
                     } else {
+                        search_context.nearest.prune_elements();
```
KShivendu, line 423: Couldn't figure out a simple way to make this happen inside `iter()`. Do you have any ideas?

lib/segment/src/index/hnsw_index/search_context.rs
generall: what's a `candidate`?

KShivendu: Point whose links are to be explored.

lib/segment/src/index/hnsw_index/graph_layers_builder.rs
generall: I don't see too much problems with it

KShivendu: Okay.

lib/segment/src/index/hnsw_index/search_context.rs
```diff
@@ -24,19 +22,13 @@ impl SearchContext {
...
     }
 
     /// Updates search context with new scored point.
     /// If it is closer than existing - also add it to candidates for further search
     pub fn process_candidate(&mut self, score_point: ScoredPointOffset) {
-        let was_added = match self.nearest.push(score_point) {
-            None => true,
-            Some(removed) => removed.idx != score_point.idx,
-        };
+        let was_added = self.nearest.push(score_point);
```
agourlay, line 31: why is the deduplication based on `removed.idx != score_point.idx,` not required anymore?

lib/segment/src/index/hnsw_index/search_context.rs
```diff
@@ -24,19 +22,13 @@ impl SearchContext {
     }
 
     pub fn lower_bound(&self) -> ScoreType {
-        match self.nearest.top() {
-            None => ScoreType::min_value(),
-            Some(worst_of_the_best) => worst_of_the_best.score,
-        }
+        self.nearest.threshold()
```
agourlay, line 25: Why is it Ok. to use threshold here given that it is lazy and might not represent the full information of the top values?

lib/segment/src/index/hnsw_index/search_context.rs
generall: We can't return index of the point if it was discarded or not, because we would need to return a whole list in this case. So the TopK return type is different and it is just bool

lib/segment/src/index/hnsw_index/search_context.rs
generall: This might be not Ok, it is up to benchmarks

generall:
It doesn't look like the change gives us a benefit: 
image

Even though the TopK might improve candidate selection performance, it looks like the change in search_context update logic have more negative effect on overall results

KShivendu:
We see small gains (<10%) in latency with lower dims datasets like glove-100 but it also costs us some precision. So we decided to not merge this. 

glove-100-latency

generall:
I seems this PR is not longer neeeded. Closing it

PR #4117, opened 2024-04-25
REST API multivector for write and retrieve

agourlay:
This PR adds the support for multivector in the REST API for writes and retrieve use case.

An important point is that I decided to keep the conversion between REST Vectors and Segment Vectors infallible because handling error would create a gigantic change through the codebase.

The assumption in this PR is that the input vectors will have gone through validation first.

I have integration added tests showing that:
- the validation kicks in properly
- writes work
- retrieves work
- search is not supported

### DIFF
--- a/lib/api/src/rest/conversions.rs
+++ b/lib/api/src/rest/conversions.rs
@@ -6,9 +6,8 @@ impl From<segment::data_types::vectors::Vector> for Vector {
         match value {
             segment::data_types::vectors::Vector::Dense(vector) => Vector::Dense(vector),
             segment::data_types::vectors::Vector::Sparse(vector) => Vector::Sparse(vector),
-            segment::data_types::vectors::Vector::MultiDense(_vector) => {
-                // TODO(colbert)
-                unimplemented!()
+            segment::data_types::vectors::Vector::MultiDense(vector) => {
+                Vector::MultiDense(vector.into_multi_vectors())
             }
         }
     }
@@ -19,6 +18,13 @@ impl From<Vector> for segment::data_types::vectors::Vector {
         match value {
             Vector::Dense(vector) => segment::data_types::vectors::Vector::Dense(vector),
             Vector::Sparse(vector) => segment::data_types::vectors::Vector::Sparse(vector),
+            Vector::MultiDense(vector) => {
+                // the REST vectors have been validated already
+                // we can use an internal constructor
+                segment::data_types::vectors::Vector::MultiDense(
+                    segment::data_types::vectors::MultiDenseVector::new_validated(vector),
+                )
+            }
         }
     }
 }
--- a/lib/api/src/rest/conversions.rs
+++ b/lib/api/src/rest/conversions.rs
@@ -22,7 +22,7 @@ impl From<Vector> for segment::data_types::vectors::Vector {
                 // the REST vectors have been validated already
                 // we can use an internal constructor
                 segment::data_types::vectors::Vector::MultiDense(
-                    segment::data_types::vectors::MultiDenseVector::new_validated(vector),
+                    segment::data_types::vectors::MultiDenseVector::new_unchecked(vector),
                 )
             }
         }
--- a/lib/api/src/rest/schema.rs
+++ b/lib/api/src/rest/schema.rs
@@ -6,11 +6,15 @@ use serde::{Deserialize, Serialize};
 /// Type for dense vector
 pub type DenseVector = Vec<segment::data_types::vectors::VectorElementType>;
 
+/// Type for multi dense vector
+pub type MultiDenseVector = Vec<DenseVector>;
+
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize, JsonSchema)]
 #[serde(untagged, rename_all = "snake_case")]
 pub enum Vector {
     Dense(DenseVector),
     Sparse(sparse::common::sparse_vector::SparseVector),
+    MultiDense(MultiDenseVector),
 }
 
 /// Full vector data per point separator with single and multiple vector modes
@@ -29,6 +33,7 @@ impl VectorStruct {
             VectorStruct::Multi(vectors) => vectors.values().all(|v| match v {
                 Vector::Dense(vector) => vector.is_empty(),
                 Vector::Sparse(vector) => vector.indices.is_empty(),
+                Vector::MultiDense(vector) => vector.is_empty(),
             }),
         }
     }
@@ -123,4 +128,5 @@ pub enum NamedVectorStruct {
     Default(segment::data_types::vectors::DenseVector),
     Dense(segment::data_types::vectors::NamedVector),
     Sparse(segment::data_types::vectors::NamedSparseVector),
+    // No support for multi-dense vectors in search
 }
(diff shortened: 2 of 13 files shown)


lib/api/src/rest/conversions.rs
```diff
@@ -19,6 +18,13 @@ impl From<Vector> for segment::data_types::vectors::Vector {
         match value {
             Vector::Dense(vector) => segment::data_types::vectors::Vector::Dense(vector),
             Vector::Sparse(vector) => segment::data_types::vectors::Vector::Sparse(vector),
+            Vector::MultiDense(vector) => {
+                // the REST vectors have been validated already
+                // we can use an internal constructor
+                segment::data_types::vectors::Vector::MultiDense(
+                    segment::data_types::vectors::MultiDenseVector::new_validated(vector),
```
agourlay, line 25: Without this trick, I would be force to use a `TryFrom` here and propagate it to all callsites of this conversions.
The change was massive so I decided to assume that the input vectors are valid at this point based on my tests.

tests/openapi/openapi_integration/test_multi_vector.py
```diff
@@ -0,0 +1,199 @@
...
+                    "id": 1,
+                    "vector": {
+                        "my-multivec": [0.19, 0.81, 0.75, 0.11]
+                    }
+                }
+            ]
+        }
+    )
+    assert not response.ok
+    assert 'Wrong input: Conversion between multi and regular vectors failed' in response.json()["status"]["error"]
```
agourlay, line 134: this is a storage error because we send a `DenseVector` to the `MultiDenseStorage`

lib/segment/src/data_types/vectors.rs
```diff
@@ -197,13 +197,25 @@ pub type MultiDenseVector = TypedMultiDenseVector<VectorElementType>;
 
 impl<T: PrimitiveVectorElement> TypedMultiDenseVector<T> {
     pub fn new(flattened_vectors: TypedDenseVector<T>, dim: usize) -> Self {
-        assert!(flattened_vectors.len() % dim == 0, "Invalid vector length");
+        assert_eq!(flattened_vectors.len() % dim, 0, "Invalid vector length");
         Self {
             inner_vector: flattened_vectors,
```
coszio, line 202: Worth renaming the field itself to `flat_vector`?

lib/segment/src/data_types/vectors.rs
```diff
@@ -197,13 +197,25 @@ pub type MultiDenseVector = TypedMultiDenseVector<VectorElementType>;
 
 impl<T: PrimitiveVectorElement> TypedMultiDenseVector<T> {
     pub fn new(flattened_vectors: TypedDenseVector<T>, dim: usize) -> Self {
-        assert!(flattened_vectors.len() % dim == 0, "Invalid vector length");
+        assert_eq!(flattened_vectors.len() % dim, 0, "Invalid vector length");
```
coszio, line 200:
```suggestion
        debug_assert_eq!(flattened_vectors.len() % dim, 0, "Invalid vector length");
```

lib/segment/src/data_types/vectors.rs
```diff
@@ -221,6 +233,15 @@ impl<T: PrimitiveVectorElement> TypedMultiDenseVector<T> {
         self.inner_vector.chunks_exact_mut(self.dim)
     }
 
+    /// Consumes the multi vector and returns the underlying individual vectors
+    pub fn into_multi_vectors(self) -> Vec<Vec<T>> {
+        let mut chunks = vec![];
+        for chunk in self.inner_vector.into_iter().chunks(self.dim).into_iter() {
+            chunks.push(chunk.collect());
+        }
+        chunks
```
coszio, line 242: 🥢 Unnecessary nit: since it is already using `chunks`... 
```suggestion
        self.inner_vector
            .into_iter()
            .chunks(self.dim)
            .into_iter()
            .map(Iterator::collect)
            .collect()
```

tests/openapi/openapi_integration/test_multi_vector.py
```diff
@@ -0,0 +1,216 @@
...
+        api='/collections/{collection_name}/points',
+        method="PUT",
+        path_params={'collection_name': collection_name},
+        query_params={'wait': 'true'},
+        body={
+            "points": [
+                {
+                    "id": 1,
+                    "vector": {
+                        "my-multivec": [
```
IvanPleshkov, line 187: Also, you can add a test with with different dimension inner vectors

lib/collection/src/operations/types.rs
```diff
@@ -1304,6 +1305,9 @@ pub struct VectorParams {
 
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub datatype: Option<Datatype>,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub multi_vec_config: Option<MultiVectorConfig>,
```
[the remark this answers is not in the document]
IvanPleshkov: We need some flag to say that these vectors are multivectors. But we found just `bool` a bad idea because multivectors may be configurable in future (for instance, we may introduce new multivector scorers, not only `MaxSim`)

lib/common/common/src/validation.rs
```diff
@@ -142,6 +142,44 @@ pub fn validate_sha256_hash_option(value: &Option<impl AsRef<str>>) -> Result<()
...
+    if multivec.is_empty() {
+        let mut errors = ValidationErrors::default();
+        let mut err = ValidationError::new("empty_multi_vector");
+        err.add_param(Cow::from("message"), &"multi vector must not be empty");
+        errors.add("data", err);
+        return Err(errors);
+    }
+
+    // check all individual vectors non-empty
+    if multivec.iter().any(|v| v.is_empty()) {
```
IvanPleshkov, line 156: It's just a small discussion, not a proposal. We can check only the first vector for non-empty because later we check that all vectors have the same dim. It will produce different error messages but we do less validation job. I think it doesn't worth it, but we can change this check.

lib/api/src/rest/schema.rs
```diff
@@ -123,4 +128,5 @@ pub enum NamedVectorStruct {
     Default(segment::data_types::vectors::DenseVector),
     Dense(segment::data_types::vectors::NamedVector),
     Sparse(segment::data_types::vectors::NamedSparseVector),
+    // No support for multi-dense vectors in search
```
IvanPleshkov, line 131: What's about advanced search API like Discovery? They have another endpoint and I don't understand why we won't support multivector for discovery/recommend/etc do untrivial enum here with mismatch

IvanPleshkov: Got it, mark as resolved

lib/segment/src/data_types/vectors.rs
agourlay: I like `flattened_vectors` because the convey the plural aspect of multivectors.
Why do you prefer `flat_vector`? :)

lib/collection/src/operations/types.rs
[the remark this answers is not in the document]
agourlay: we need this field to enable multivector support when creating a collection

lib/segment/src/data_types/vectors.rs
coszio: Yeah, `flattened_vectors` too, but I meant renaming `inner_vector`

coszio (approved):

lib/common/common/src/validation.rs
agourlay: given they the number of vectors per multivector is rather low, I'd rather keep a clear error message instead of optimizing :)

PR #4119, opened 2024-04-26
Optimize NumericIndex value_range

agourlay:
Tried to optimize the immutable numeric index but it does not improve things according to new benchmarks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -180,6 +180,7 @@ opt-level = 0
 [profile.bench]
 lto = false
 debug = true
+codegen-units = 256 # restore default value for faster compilation
 
 # Profile for performance testing, which is faster to build than release.
 [profile.perf]
--- a/lib/segment/benches/range_filtering.rs
+++ b/lib/segment/benches/range_filtering.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 
 use atomic_refcell::AtomicRefCell;
 use common::types::PointOffsetType;
-use criterion::{criterion_group, criterion_main, Criterion};
+use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
 use rand::prelude::StdRng;
 use rand::{Rng, SeedableRng};
 use segment::fixtures::payload_context_fixture::FixtureIdTracker;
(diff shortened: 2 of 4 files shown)


agourlay:
@IvanPleshkov WDYT about this?

Is the change still valuable although I can't prove it is better?
Should I just push the benchmarks?

Cargo.toml
```diff
@@ -180,6 +180,7 @@ opt-level = 0
 [profile.bench]
 lto = false
 debug = true
+codegen-units = 256 # restore default value for faster compilation
```
IvanPleshkov, line 183: Is it possible to to remove this line and add check in `[profile.release]` if it is a non-bench build?

lib/segment/benches/range_filtering.rs
```diff
@@ -0,0 +1,137 @@
...
+    // make sure all points are indexed
+    assert_eq!(index.indexed_points(&FLT_KEY.parse().unwrap()), NUM_POINTS);
+    assert_eq!(index.indexed_points(&INT_KEY.parse().unwrap()), NUM_POINTS);
+
+    let mut result_size = 0;
+    let mut query_count = 0;
+
+    group.bench_function("float-mutable-index", |b| {
+        b.iter(|| {
+            let filter = random_range_filter(&mut rng, FLT_KEY);
```
IvanPleshkov, line 90: I don't know how much random filter generation costs. Can you try to build large predefined filters before measurement and check if the measurement result doesn't change? I guess randomness costs enough to move random generator outside the measured scope

lib/segment/src/index/field_index/numeric_index/immutable_numeric_index.rs
```diff
@@ -151,13 +150,19 @@ impl<T: Encodable + Numericable> NumericKeySortedVec<T> {
         }
     }
 
-    fn find_end_index(&self, bound: Bound<NumericIndexKey<T>>) -> usize {
+    /// Find the end index of the range.
+    ///
+    /// Scans only starting from `start` to avoid repeated work.
+    fn find_end_index(&self, start: usize, bound: Bound<NumericIndexKey<T>>) -> usize {
```
IvanPleshkov, line 156: Instead of a documentation comment, I propose to add asserts like `assert!(start <= self.data.len() && start <= bound.start)`

lib/segment/src/index/field_index/numeric_index/immutable_numeric_index.rs
```diff
@@ -151,13 +150,19 @@ impl<T: Encodable + Numericable> NumericKeySortedVec<T> {
...
         match bound {
-            Bound::Included(bound) => match self.data.binary_search(&bound) {
-                Ok(idx) => idx + 1,
-                Err(idx) => idx,
+            Bound::Included(bound) => match self.data[start..].binary_search(&bound) {
```
IvanPleshkov, line 158: What if start is `0` and `self.data` is empty? is it possible to meet such a situation? Maybe we have a bug here

IvanPleshkov (approved):

lib/segment/benches/range_filtering.rs
ffuugoo: I think I've used [`Bencher::iter_batched`](https://docs.rs/criterion/latest/criterion/struct.Bencher.html#method.iter_batched) for this before.

Cargo.toml
IvanPleshkov: cause I had to open rust documentation to check the default value

lib/segment/src/index/field_index/numeric_index/immutable_numeric_index.rs
```diff
@@ -151,13 +150,19 @@ impl<T: Encodable + Numericable> NumericKeySortedVec<T> {
         }
     }
 
-    fn find_end_index(&self, bound: Bound<NumericIndexKey<T>>) -> usize {
+    fn find_end_index(&self, start: usize, bound: Bound<NumericIndexKey<T>>) -> usize {
+        if start >= self.data.len() {
+            return self.data.len();
```
IvanPleshkov, line 155: I guess start index has to be always not larger that end index. I propose to return `start` instead of `self.data.len()`
```suggestion
            return start;
```

PR #4121, opened 2024-04-26
Make `LocalShard` aware of its collection id

coszio:
[Issues API] For implementing `UnindexedField` issues, we want to trigger them based on execution time. Since timing it at `Collection` level and above works, the best place to time it would be at the `LocalShard`, to exclude any network latencies.

Without this PR it is not possible to do the above, because we need to publish an event which includes the collection name + filters.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -66,6 +66,7 @@ const WAL_LOAD_REPORT_EVERY: Duration = Duration::from_secs(60);
 ///
 /// Holds all object, required for collection functioning
 pub struct LocalShard {
+    pub(super) collection_id: CollectionId,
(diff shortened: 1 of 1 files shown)


generall:
Looks like an abstraction leak to me. I would rather do it completely on collection level.

coszio:
@generall No concerns with false positives? a `POST /issues/clear` endpoint will be even more important

generall:
But higher timing is not enough to trigger the event, right? You also need wrong configuration.

coszio:
Yes, but not every payload key used to filter needs to be indexed, is it?

generall:
> Yes, but not every payload key used to filter needs to be indexed, is it?

not every, but if the latency is high, pretty much for sure it is part of the problem

coszio:
Ok, closing.

Issue #4110, opened 2024-04-24
Support for f16 vector metrics


generall:
**Is your feature request related to a problem? Please describe.**

Currently Qdrant supports storage of raw vectors with 2 datatypes: `f32` and `uint8`.
In some applications it is also desirable to have `f16` datatype support.

Qdrant `VectorStorage` itself is generic to the datatype and technically can store vectors of any type with no additional changes.

There is, however, a component which depends on datatype and requires specialized implementation - that is metrics.

**Describe the solution you'd like**

Provide SIMD-optimized implementations for the following:

```
impl Metric<f16> for DotProductMetric {
impl Metric<f16> for CosineMetric {
impl Metric<f16> for EuclidMetric {
impl Metric<f16> for ManhattanMetric {
```

for `avx2`, `sse` and `neon` architectures in a similar way as it is implemented for f32.

Please also include unit-tests to ensure correctness of the implementation.

**Describe alternatives you've considered**

- add a compile-time option to switch `VectorElementType` between f32 and f16, but this approach is less flexible and requires more changes

**Additional context**
-

TheQuantumFractal:
Happy to help with this. You just want SIMD implementations and tests for f16? (ie. code doesn't have to be refactored to account for f16 datatype correct?)
/attempt #4110

generall:
> Happy to help with this.

Thanks :smiley: 

> code doesn't have to be refactored to account for f16 datatype correct?

yeah, integration of f16 into API and storage would be a separate issue.

abhishek818:
Found a [crate](https://github.com/ashvardanian/simsimd#using-simsimd-in-rust) based on simd for f16.

abhishek818:
Great work @TheQuantumFractal !

PR #4122, opened 2024-04-26
Half-precision float vector metrics

TheQuantumFractal:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

I built out a SIMD implementation with testing for Neon, AVX2, SSE2 on euclidean, manhattan, and dot similarity. Something to note is that float16 SIMD operations are not supported on most ISAs (ARM32/64 processors are able to handle it, and AVX512 recently announced some hardware support but most machines on AVX2 and SSE2 do not support it). F16C is an x86 instruction set extension supported on most x86 modern machines that supports conversion between half- and single-precision floating point formats. Essentially, to run the metrics on AVX2 or SSE2, f16 vectors need to be converted to f32 then processed with f32 SIMD accordingly. My implementations are as such. I also wrote out a separate C / assembly file that enables Neon f16 SIMD operations since Rust does not currently support ARM f16 SIMD operations.

The AVX2 / SSE2 SIMD was tested on a Intel(R) Xeon(R) CPU while the Neon SIMD was tested on an Apple M1 Pro.

As for cosine similarity, the current cosine similarity preprocess step accepts float32 DenseVectors and simply normalizes them. You can similarly normalize the float16 vectors by computing dot product between the vector and itself using the dot similarity SIMD implementation. The actual metric after preprocessing would use the same SIMD dot similarity implementation.

/claim #4110


lib/segment/src/spaces/metric_f16/arm.s
generall, line 1: I don't feel too confident about including ASM files into the project as-is.
 @TheQuantumFractal could you please elaborate why you decided to do it like this instead of directly linking C?
 
 FYI in qunatizations repo we have an example https://github.com/qdrant/quantization/tree/master/quantization/cpp
 
 If you think ASM is strictly necessary, could you please include an instruction of how to generate it from C

TheQuantumFractal: Yes, there isn't really a reason to include asm files. Directly linking C is a better solution. I can just set up linking for neon in qdrant/lib/segment then?

generall: That would help, thanks!

IvanPleshkov: @TheQuantumFractal you can find the example how to integrate C code here:
https://github.com/qdrant/quantization/blob/master/quantization/build.rs

IvanPleshkov: Please use this example because here we solved cross-compilation issues (for instance, build on x64 host binary for arm target)

lib/segment/src/spaces/metric_f16/simple.rs
```diff
@@ -0,0 +1,201 @@
...
+        #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+        {
+            if is_x86_feature_detected!("sse") && v1.len() >= MIN_DIM_SIZE_SIMD {
+                return unsafe { euclid_similarity_sse(v1, v2) };
+            }
+        }
+
+        #[cfg(all(target_arch = "aarch64", target_feature = "neon"))]
+        {
+            if std::arch::is_aarch64_feature_detected!("neon") && v1.len() >= MIN_DIM_SIZE_SIMD {
```
IvanPleshkov, line 63: float 16-bit neon was introduced in `armv8.2`, for `armv7.*` and `armv8.0` this code will produce segfault because of calling of non-exists CPU instruction. We need to check if float 16 is supported on device

IvanPleshkov: the same for each metric

lib/segment/src/spaces/mod.rs
```diff
@@ -8,6 +8,7 @@ pub mod simple_sse;
 #[cfg(target_arch = "x86_64")]
 pub mod simple_avx;
 
+pub mod metric_f16;
```
IvanPleshkov, line 11: Reorganise please the `mod` like `metric_unit`. `half`-typed vectors should have the same design as `byte`-typed vectors.
Please move all implementations like in this folder:
https://github.com/qdrant/qdrant/tree/dev/lib/segment/src/spaces/metric_uint

lib/segment/src/data_types/primitive.rs
```diff
@@ -67,6 +68,61 @@ impl PrimitiveVectorElement for VectorElementType {
     }
 }
 
+impl PrimitiveVectorElement for VectorElementTypeHalf {
+    fn slice_from_float_cow(vector: Cow<[VectorElementType]>) -> Cow<[Self]> {
+        Cow::Owned(vector.iter().map(|&x| f16::from_f32(x)).collect())
```
IvanPleshkov, line 73: you convert `f16-vectors` several times. I propose to define a separate function for `f16`-vector conversions

lib/segment/src/spaces/metric_f16/arm.c
```diff
@@ -0,0 +1,135 @@
...
+
+
+float32_t dotProduct_half_4x4(const float16_t* pSrcA, const float16_t* pSrcB, uint32_t blockSize)
+{
+    float32_t dotProduct = 0.0f;
+    float16x8_t sum1 = vdupq_n_f16(0.0f);
+    float16x8_t sum2 = vdupq_n_f16(0.0f);
+    float16x8_t sum3 = vdupq_n_f16(0.0f);
+    float16x8_t sum4 = vdupq_n_f16(0.0f);
+    uint32_t i = 0;
```
IvanPleshkov, line 15: Why do you want to define iterator here instead of `for (int i, ...)`?

lib/segment/src/spaces/metric_f16/arm.c
```diff
@@ -0,0 +1,135 @@
...
+    float32_t euclideanDistance = 0.0f;
+    float16x8_t sum1 = vdupq_n_f16(0.0f);
+    float16x8_t sub1 = vdupq_n_f16(0.0f);
+    float16x8_t sum2 = vdupq_n_f16(0.0f);
+    float16x8_t sub2 = vdupq_n_f16(0.0f);
+    float16x8_t sum3 = vdupq_n_f16(0.0f);
+    float16x8_t sub3 = vdupq_n_f16(0.0f);
+    float16x8_t sum4 = vdupq_n_f16(0.0f);
+    float16x8_t sub4 = vdupq_n_f16(0.0f);
+    uint32_t i = 0;
```
IvanPleshkov, line 56: Why not inside `for` definition?

lib/segment/src/spaces/metric_f16/arm.c
```diff
@@ -0,0 +1,135 @@
...
+}
+
+float32_t manhattanDist_half_4x4(const float16_t* pSrcA, const float16_t* pSrcB, uint32_t blockSize)
+{
+    float32_t manhattanDistance = 0.0f;
+    float16x8_t sum1 = vdupq_n_f16(0.0f);
+    float16x8_t sum2 = vdupq_n_f16(0.0f);
+    float16x8_t sum3 = vdupq_n_f16(0.0f);
+    float16x8_t sum4 = vdupq_n_f16(0.0f);
+    uint32_t i = 0;
```
IvanPleshkov, line 102: Why not inside `for` definition?

lib/segment/src/spaces/metric_f16/arm.c
```diff
@@ -0,0 +1,135 @@
...
+    sum1 = vaddq_f16(sum1, sum4);
+
+    float32x4_t sum = vcvt_f32_f16(vget_high_f16(sum1));
+    sum = vaddq_f32(sum, vcvt_f32_f16(vget_low_f16(sum1)));
+
+    manhattanDistance = vaddvq_f32(sum);
+    float32_t tmp = 0.0f;
+    for (i=0; i < (blockSize % 32); i++) {
+        tmp = (*pSrcA - *pSrcB);
+        manhattanDistance += tmp > 0 ? tmp : -tmp;
```
IvanPleshkov, line 129: Why not `abs` instead?

lib/segment/src/spaces/metric_f16/simple_avx.rs
```diff
@@ -0,0 +1,319 @@
...
+    let x128: __m128 = _mm_add_ps(_mm256_extractf128_ps(x, 1), _mm256_castps256_ps128(x));
+    let x64: __m128 = _mm_add_ps(x128, _mm_movehl_ps(x128, x128));
+    let x32: __m128 = _mm_add_ss(x64, _mm_shuffle_ps(x64, x64, 0x55));
+    _mm_cvtss_f32(x32)
+}
+
+#[target_feature(enable = "avx")]
+#[target_feature(enable = "fma")]
+#[target_feature(enable = "f16c")]
+pub(crate) unsafe fn euclid_similarity_avx(
```
IvanPleshkov, line 21: One comment for all namings. `euclid_similarity_avx` is already presented. Please, rename `euclid_similarity_avx` into `avx_euclid_similarity_half` like it was named for byte type:
https://github.com/qdrant/qdrant/blob/dev/lib/segment/src/spaces/metric_uint/avx2/euclid.rs#L9

IvanPleshkov: Do this please for all simd functions

IvanPleshkov (changes requested):
Also, it would be nice to add `f16` scoring benchmarks. You can do it here where `byte` scoring defined
https://github.com/qdrant/qdrant/blob/dev/lib/segment/benches/metrics.rs

lib/segment/src/spaces/metric_f16/simple.rs
```diff
@@ -0,0 +1,201 @@
...
+impl Metric<VectorElementTypeHalf> for EuclidMetric {
+    fn distance() -> Distance {
+        Distance::Euclid
+    }
+
+    fn similarity(v1: &[VectorElementTypeHalf], v2: &[VectorElementTypeHalf]) -> ScoreType {
+        #[cfg(target_arch = "x86_64")]
+        {
+            if is_x86_feature_detected!("avx")
+                && is_x86_feature_detected!("fma")
```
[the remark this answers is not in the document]
generall: Do you have a suggestion of how to make this check?

lib/segment/src/spaces/metric_f16/arm.s
TheQuantumFractal: Added build.rs to link the C file.

lib/segment/src/spaces/metric_f16/simple.rs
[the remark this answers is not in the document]
TheQuantumFractal: Resolved these.

lib/segment/src/spaces/metric_f16/simple_sse.rs
```diff
@@ -0,0 +1,272 @@
...
+use std::arch::x86_64::*;
+
+use common::types::ScoreType;
+use half::f16;
+
+use crate::data_types::vectors::VectorElementTypeHalf;
+
+#[target_feature(enable = "sse")]
+#[allow(clippy::missing_safety_doc)]
+pub unsafe fn hsum128_ps_sse(x: __m128) -> f32 {
```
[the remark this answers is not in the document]
TheQuantumFractal: sse code was resolved to not use f16c (it just converts to f32 then calls the f32 simd)

lib/segment/src/spaces/metric_f16/simple_avx.rs
```diff
@@ -0,0 +1,319 @@
...
+
+use common::types::ScoreType;
+use half::f16;
+
+use crate::data_types::vectors::VectorElementTypeHalf;
+
+#[target_feature(enable = "avx")]
+#[target_feature(enable = "fma")]
+#[allow(clippy::missing_safety_doc)]
+pub unsafe fn hsum256_ps_avx(x: __m256) -> f32 {
```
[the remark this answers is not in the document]
TheQuantumFractal: resolved

lib/segment/src/spaces/metric_f16/simple_avx.rs
```diff
@@ -0,0 +1,319 @@
...
+    let mut sum256_1: __m256 = _mm256_setzero_ps();
+    let mut sum256_2: __m256 = _mm256_setzero_ps();
+    let mut sum256_3: __m256 = _mm256_setzero_ps();
+    let mut sum256_4: __m256 = _mm256_setzero_ps();
+
+    let mut addr1s: __m128i;
+    let mut addr2s: __m128i;
+    let mut i: usize = 0;
+    while i < m {
+        addr1s = _mm_loadu_si128(ptr1);
```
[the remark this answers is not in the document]
TheQuantumFractal: I believe so. The [f16c operations](https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-9/mm256-cvtph-ps.html) require input registers to be __m128i.

lib/segment/src/spaces/metric_f16/simple_avx.rs
```diff
@@ -0,0 +1,319 @@
...
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_spaces_avx() {
+        use super::*;
+        use crate::spaces::metric_f16::simple::*;
+
+        if is_x86_feature_detected!("avx") && is_x86_feature_detected!("fma") {
+            let v1: Vec<f16> = vec![
```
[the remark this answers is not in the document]
TheQuantumFractal: resolved

lib/segment/src/spaces/metric_f16/simple_sse.rs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_spaces_sse() {
+        use super::*;
+        use crate::spaces::metric_f16::simple::*;
+
+        if is_x86_feature_detected!("sse") {
```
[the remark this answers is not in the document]
TheQuantumFractal: resolved

lib/segment/src/spaces/metric_f16/simple_sse.rs
```diff
@@ -0,0 +1,272 @@
...
+#[allow(clippy::missing_safety_doc)]
+pub unsafe fn hsum128_ps_sse(x: __m128) -> f32 {
+    let x64: __m128 = _mm_add_ps(x, _mm_movehl_ps(x, x));
+    let x32: __m128 = _mm_add_ss(x64, _mm_shuffle_ps(x64, x64, 0x55));
+    _mm_cvtss_f32(x32)
+}
+
+#[target_feature(enable = "sse")]
+#[target_feature(enable = "f16c")]
+pub(crate) unsafe fn euclid_similarity_sse(
```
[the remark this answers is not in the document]
TheQuantumFractal: resolved

lib/segment/src/spaces/metric_f16/simple.rs
TheQuantumFractal: Added a check in build.rs to account for this.

lib/segment/src/spaces/mod.rs
TheQuantumFractal: Done

lib/segment/src/spaces/metric_f16/arm.c
TheQuantumFractal: Added the iterator into the for loop

lib/segment/src/spaces/metric_f16/arm.c
TheQuantumFractal: Using the arm f16 abs operation now

lib/segment/src/spaces/metric_f16/simple_avx.rs
TheQuantumFractal: Done

IvanPleshkov:
@TheQuantumFractal rebase please to the latest dev, CI is red

generall (approved):

generall:
Hey @TheQuantumFractal thanks a lot for the contribution! We will take it from here and finish the integration as a separate PR.

TheQuantumFractal:
Sounds good! Happy to help :)

PR #4123, opened 2024-04-26
Issues API: Listen for issues events in main app

coszio:
Needs #4086 

This PR setups and demonstrates how issue subscribers listen to issue-able events. 

While this PR won't submit nor solve any `Issue`, it slims down further PRs, for easier reviewability.

### DIFF
--- a/lib/storage/src/issues_subscribers.rs
+++ b/lib/storage/src/issues_subscribers.rs
@@ -1,0 +1,41 @@
+use std::sync::Arc;
+
+use collection::events::{CollectionDeletedEvent, IndexCreatedEvent, SlowQueryEvent};
+use issues::broker::Subscriber;
+
+use crate::content_manager::toc::TableOfContent;
+use crate::dispatcher::Dispatcher;
+use crate::rbac::Access;
+
+#[allow(dead_code)] // TODO: remove
+#[derive(Clone)]
+pub struct UnindexedFieldSubscriber {
+    toc: Arc<TableOfContent>,
+    access: Access,
+}
+
+impl UnindexedFieldSubscriber {
+    pub fn new(dispatcher: Arc<Dispatcher>) -> Self {
+        let access = Access::full("For Issues API");
+        let toc = dispatcher.toc(&access).clone();
+        Self { toc, access }
+    }
+}
+
+impl Subscriber<SlowQueryEvent> for UnindexedFieldSubscriber {
+    fn notify(&self, _event: Arc<SlowQueryEvent>) {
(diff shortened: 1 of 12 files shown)


lib/storage/src/issues_subscribers.rs
```diff
@@ -0,0 +1,41 @@
+use std::sync::Arc;
+
+use collection::events::{CollectionDeletedEvent, IndexCreatedEvent, SlowQueryEvent};
+use issues::broker::Subscriber;
+
+use crate::content_manager::toc::TableOfContent;
+use crate::dispatcher::Dispatcher;
+use crate::rbac::Access;
+
+#[allow(dead_code)] // TODO: remove
```
agourlay, line 10: still necessary?

src/main.rs
```diff
@@ -393,6 +394,9 @@ fn main() -> anyhow::Result<()> {
         log::info!("Telemetry reporting disabled");
     }
 
+    // Setup subscribers to listen for issue-able events
+    issues_setup::setup_subscribers(dispatcher_arc.clone());
```
agourlay, line 398: Can we maybe protect the feature behind a configuration flag?
It would be great to be able to disable it if necessary.

lib/storage/src/issues_subscribers.rs
```diff
@@ -0,0 +1,41 @@
...
+
+impl Subscriber<CollectionDeletedEvent> for UnindexedFieldSubscriber {
+    fn notify(&self, _event: Arc<CollectionDeletedEvent>) {
+        // TODO: solve all unindexed field issues related to the deleted collection
+    }
+}
+
+impl Subscriber<IndexCreatedEvent> for UnindexedFieldSubscriber {
+    fn notify(&self, _event: Arc<IndexCreatedEvent>) {
+        // TODO: solve the missing index issue
```
agourlay, line 39: You would still need to check that the index was created on the right field, right?

src/main.rs
coszio: Yes, one of the advantages of having the event broker. I'll leave it enabled by default

lib/storage/src/issues_subscribers.rs
coszio: It will be removed in #4139, we are not accessing `toc` and `access` here yet

src/main.rs
agourlay: Sorry, I meant a runtime configuration with a new config field.
The idea being that we can experiment with it easily and disable it if it creates problems.
WDYT?

agourlay (approved):

PR #4125, opened 2024-04-26
Issues API: Add `clear_issues` REST endpoint

coszio:
Adds a `POST /issues/clear` endpoint to reset the `issues::Dashboard` in case the user wants to remove false positives.

Tests will be added later once we have any issue submission merged.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/src/actix/api/issues_api.rs
+++ b/src/actix/api/issues_api.rs
@@ -1,5 +1,5 @@
 use actix_web::rt::time::Instant;
-use actix_web::{get, web, Responder};
+use actix_web::{get, post, web, Responder};
 use collection::operations::types::IssuesReport;
 use storage::rbac::AccessRequirements;
 
@@ -17,7 +17,20 @@ async fn get_issues(ActixAccess(access): ActixAccess) -> impl Responder {
     process_response(response, timing)
 }
 
+#[post("/issues/clear")]
+async fn clear_issues(ActixAccess(access): ActixAccess) -> impl Responder {
+    let timing = Instant::now();
+    let response = access
+        .check_global_access(AccessRequirements::new().manage())
+        .map(|_| {
+            issues::clear();
+            true
+        });
+    process_response(response, timing)
+}
+
 // Configure services
 pub fn config_issues_api(cfg: &mut web::ServiceConfig) {
     cfg.service(get_issues);
+    cfg.service(clear_issues);
 }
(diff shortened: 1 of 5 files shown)


src/actix/api/issues_api.rs
```diff
@@ -17,7 +17,20 @@ async fn get_issues(ActixAccess(access): ActixAccess) -> impl Responder {
     process_response(response, timing)
 }
 
+#[post("/issues/clear")]
+async fn clear_issues(ActixAccess(access): ActixAccess) -> impl Responder {
```
timvisee, line 21: How about `DELETE /issues`?

src/actix/api/issues_api.rs
```diff
@@ -17,7 +17,20 @@ async fn get_issues(ActixAccess(access): ActixAccess) -> impl Responder {
...
+    let timing = Instant::now();
+    let response = access
+        .check_global_access(AccessRequirements::new().manage())
+        .map(|_| {
+            issues::clear();
+            true
+        });
+    process_response(response, timing)
+}
```
xzfc, line 30: Could be simplified as:
```suggestion
fn clear_issues(ActixAccess(access): ActixAccess) -> impl Future<Output = HttpResponse> {
    crate::actix::helpers::time(async move {
        access.check_global_access(AccessRequirements::new().manage())?;
        issues::clear();
        Ok(true)
    })
}
```

src/actix/api/issues_api.rs
coszio: I thought `DELETE` verb was only for single resources, but after a little digging, it is normal to do it for the resource collection as well

src/actix/api/issues_api.rs
coszio: Nice! looks cleaner

agourlay (approved):

tests/openapi_consistency_check.sh
```diff
@@ -35,8 +35,8 @@ fi
 # Cleanup
 rm -f ./docs/redoc/master/.diff.openapi.json
 
-NUMBER_OF_APIS=$(cat ./docs/redoc/master/openapi.json | jq '.paths | length')
-EXPECTED_NUMBER_OF_APIS=52
+NUMBER_OF_APIS=$(cat ./docs/redoc/master/openapi.json | jq '[.paths[] | length] | add')
+EXPECTED_NUMBER_OF_APIS=65
```
coszio, line 39: we were only counting the amount of paths, not the combination of verb + path

PR #4126, opened 2024-04-26
Sparse idf dot

generall:
This PR introduces a `modifier` param, which can be configured for individual sparse vectors.

It is also introduces a single non-empty value for the modifier, called `idf`.

Usage example:

```
PATCH collections/test_sparse_idf
{
    "sparse_vectors": {
        "text": {
            "modifier": "idf"
        }
    }
}
```

Modifier alters the weights of sparse vectors, so the resulting score (and order) are different.

The `idf` modifier multiplies weights of the sparse vectors to account for "rarity" of the vector elements in the current collection. Less frequent the element is, more impact on the scoring it will have. 

Exact formula looks like this: 

```
((n - df + 0.5) / (df + 0.5) + 1.).ln()
```

where `n` - number of points in shard, `df` - number of points with given element.
This component is computed at query time and multiplied with query before performing the actual search. 

Similar scoring component is used in Okapi BM25, which allows usage of sparse vectors as a replacement for Bm25 text search without the necessity of pre-computing of the corpus statistics.

### DIFF
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -25,7 +25,7 @@ use crate::collection_manager::probabilistic_segment_search_sampling::find_searc
 use crate::collection_manager::search_result_aggregator::BatchResultAggregator;
 use crate::config::CollectionConfig;
 use crate::operations::query_enum::QueryEnum;
-use crate::operations::types::{CollectionResult, CoreSearchRequestBatch, Record};
+use crate::operations::types::{CollectionResult, CoreSearchRequestBatch, Modifier, Record};
 use crate::optimizers_builder::DEFAULT_INDEXING_THRESHOLD_KB;
 
 type BatchOffset = usize;
@@ -174,7 +174,7 @@ impl SegmentsSearcher {
                 .params
                 .get_sparse_vector_params_opt(vector_name)
             {
-                if sparse_vector_params.idf.unwrap_or_default()
+                if sparse_vector_params.modifier == Some(Modifier::Idf)
                     || !idf_vectors.contains(&vector_name)
                 {
                     idf_vectors.push(vector_name);
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -245,7 +245,7 @@ impl SegmentsSearcher {
             // - segments are not empty
             let use_sampling = sampling_enabled
                 && segments_lock.len() > 1
-                && query_context_acr.get_available_point_count() > 0;
+                && query_context_acr.available_point_count() > 0;
 
             segments
                 .map(|segment| {
@@ -575,7 +575,7 @@ fn execute_batch_search(
             search_params.top,
             ef_limit,
             segment_points,
-            query_context.get_available_point_count(),
+            query_context.available_point_count(),
         )
     } else {
         search_params.top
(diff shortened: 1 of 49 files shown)


tests/openapi/openapi_integration/test_sparse_idf.py
```diff
@@ -0,0 +1,164 @@
...
+from .helpers.helpers import request_with_validation
+
+collection_name = 'test_sparse_idf'
+
+
+@pytest.fixture(autouse=True)
+def setup():
+    idf_collection_setup(collection_name=collection_name)
+    yield
+    # drop_collection(collection_name=collection_name)
```
xzfc, line 13: This line is commented. Is this intended?

lib/sparse/src/common/sparse_vector.rs
```diff
@@ -17,6 +19,95 @@ pub struct SparseVector {
...
+#[derive(Debug, PartialEq, Clone, Default, Serialize, Deserialize)]
+pub struct RemappedSparseVector {
+    /// indices must be unique
+    pub indices: Vec<DimOffset>,
+    /// values and indices must be the same length
+    pub values: Vec<DimWeight>,
+}
+
+/// Sort two arrays by the first array.
+fn double_sort<T: Ord + Copy, V: Copy>(indices: &mut [T], values: &mut [V]) {
```
xzfc, line 33: 📝 Memo: Recently, I was trying to find an implementation of two-arrays-sort for Rust without additional memory allocation, but I couldn't find any.
There is a crate [permutation](https://crates.io/crates/permutation) that could be used, but it also allocates memory.

Shame on Rust since C++ apparently has it. https://stackoverflow.com/a/53785022

xzfc (changes requested):

lib/collection/src/collection_manager/segments_searcher.rs
```diff
@@ -254,10 +307,9 @@ impl SegmentsSearcher {
                         search_in_segment(
                             segment,
                             partial_batch_request,
-                            0,
```
agourlay, line 257: We used to pass explicitly `zero` in that case vs `available_point_count` for the query above.

Now both `search_in_segment` will use the same value from the `query_context`.

Not sure of the consequences yet.

lib/collection/src/operations/types.rs
```diff
@@ -1367,19 +1327,36 @@ impl Anonymize for VectorParams {
     }
 }
 
+/// If used, include weight modification, which will be applied to sparse vectors in query time:
```
agourlay, line 1330:
```suggestion
/// If used, include weight modification, which will be applied to sparse vectors at query time:
```

tests/openapi/openapi_integration/test_sparse_idf.py
generall: forgot after debugging, thanks!

lib/sparse/src/common/sparse_vector.rs
generall: Yeah, I didn't want to bring this additional crate just for this

lib/collection/src/collection_manager/segments_searcher.rs
generall: None, if the heuristic is not used, it won't use this parameter

agourlay:
FYI the sparse vector congruence tests are still green with this branch :ok_hand:

lib/segment/src/data_types/query_context.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    pub fn get_available_point_count(&self) -> usize {
+        self.available_point_count
+    }
+
+    pub fn get_search_optimized_threshold_kb(&self) -> usize {
+        self.search_optimized_threshold_kb
+    }
```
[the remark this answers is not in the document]
generall: fixed

agourlay (approved):

xzfc (approved):

Issue #4131, opened 2024-04-29
Optimization task panicked after a collection is recovered, causing search API timeout


no7dw:
## Current Behavior
we have a collection report error:
optimizer_status": {
      "error": "Service internal error: Optimization task panicked: called `Option::unwrap()` on a `None` value"
    },

which seems causing slow performance usually timeout on search .

we check the docker logs 
show as :

```
2024-04-29T09:04:57.585769Z  INFO actix_web::middleware::logger: 103.152.220.15 "GET /collections/new_text HTTP/1.1" 200 761 "http://34.142.201.204:6333/dashboard" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/122.0.0.0 Safari/537.36" 0.001342

2024-04-29T09:04:57.632442Z ERROR qdrant::startup: Panic backtrace:
   0: qdrant::startup::setup_panic_hook::{{closure}}
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/alloc/src/boxed.rs:2029:9
   2: std::panicking::rust_panic_with_hook
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:785:13
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:659:13
   4: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/sys_common/backtrace.rs:171:18
   5: rust_begin_unwind
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:647:5
   6: core::panicking::panic_fmt
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/panicking.rs:72:14
   7: <segment::vector_storage::vector_storage_base::VectorStorageEnum as segment::vector_storage::vector_storage_base::VectorStorage>::is_deleted_vector
   8: segment::segment::Segment::vector_by_offset
[...]

PR #4132, opened 2024-04-29
Use safe reading of deleted flags

generall:
Might be related to: https://github.com/qdrant/qdrant/issues/4131
but there are no reproduction scenario yet


coszio (approved):

timvisee (approved):
Definitely worth to merge, even if we don't have a clear repro yet.

PR #4139, opened 2024-04-29
Issues API: Unindexed fields

coszio:
Tracked by #3471 
Needs #4123
Supersedes #3511 

This PR submits and solves `UnindexedField` issues

Publishes these events:
- `SlowQueryEvent`: a search than ran longer than an arbitrary threshold (currently 300ms). Only triggered on core searches for now.
- `IndexCreatedEvent`: an index was created
- `CollectionDeletedEvent`: a collection was deleted

When a `SlowQueryEvent` happens, the subscriber will process the filters used, and infer which indexes could improve the search, comparing against the current schema.

When `IndexCreatedEvent` or `CollectionDeleteEvent` happen, the subscriber will solve the corresponding issues.

#### Open Questions
- The slow search threshold is currently configurable with a hidden `QDRANT_SLOW_SEARCH_SECS` environment variable. Should we expose this?
- Should we also publish slow queries for `scroll` and `count`?

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/segment/src/problems/unindexed_field.rs
```diff
@@ -0,0 +1,330 @@
...
+        FieldCondition {
+            key: _,
+            r#match: None,
+            range: None,
+            geo_bounding_box: None,
+            geo_radius: None,
+            geo_polygon: None,
+            values_count: _,
+        } => vec![],
+    }
```
xzfc, line 229: :pick: Nitpick: could be rewritten in this style. This way, `cargo check` would complain about unused variables if you miss handling of some field.

```rust
let FieldCondition {
    r#match,
    range,
    geo_bounding_box,
    …
} = field_condition;

if let Some(r#match) = r#match {
    vec![…]
} else if let Some(range) = range {
    vec![…]
} … {
    vec![…]
} else {
    vec![]
}
```

lib/storage/src/issues_subscribers.rs
```diff
@@ -2,40 +2,42 @@ use std::sync::Arc;
...
+
+        tokio::spawn(async move {
+            for filter in &event.filters {
+                segment::problems::UnindexedField::submit_possible_suspects(
+                    filter,
+                    &event.schema,
+                    event.collection_id.clone(),
+                )
+            }
+        });
```
xzfc, line 25: The inner function is synchronous and doesn't have any `.await`s. Does it make sense to put it into `tokio::spawn`?

coszio: You're right, it previously did some async work but not anymore, and the processing it does now is not so expensive. We can remove it

lib/segment/src/problems/unindexed_field.rs
```diff
@@ -0,0 +1,330 @@
...
+        static SLOW_SEARCH_THRESHOLD: OnceLock<Duration> = OnceLock::new();
+
+        *SLOW_SEARCH_THRESHOLD.get_or_init(|| {
+            Duration::from_millis(
+                std::env::var("QDRANT_SLOW_SEARCH_THRESHOLD")
+                    .ok()
+                    .and_then(|var| var.parse::<u64>().ok())
+                    .unwrap_or(300),
+            )
+        })
```
timvisee, line 38: A nitpick: I think we should use seconds here instead, to prevent using different units everywhere. We can do `0.3` as a float.

What do you think?

timvisee (approved):

timvisee:
To answer your open questions:

> The slow search threshold is currently configurable with a hidden QDRANT_SLOW_SEARCH_THRESHOLD environment 
variable. Should we expose this?

I think its fine to expose it this way, partially because we use it in testing too.

I don't think we should document it though to prevent users from depending on it.

If users ever ask about this specifically, we can decide on what the best way is to properly expose it.

> Should we also publish slow queries for scroll and count?

I think we do, but probably as a different type. Scroll and counts with filters are a bit different as far as I'm aware.

lib/segment/src/problems/unindexed_field.rs
coszio: rewritten :)

lib/segment/src/problems/unindexed_field.rs
coszio: I think it is a good suggestion, changed

lib/segment/src/problems/unindexed_field.rs
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    pub fn get_instance_id(collection_name: &str, field_name: &JsonPathV2) -> String {
+        format!("{collection_name}/{field_name}")
+    }
+
+    pub fn get_collection_name(code: &Code) -> &str {
+        debug_assert!(code.issue_type == TypeId::of::<Self>());
+        code.instance_id.split('/').next().unwrap() // Code format is always the same
```
agourlay, line 47: Is this `unwrap` happening on the user's request thread?
If yes it may make sense to not fail hard.

lib/segment/src/problems/unindexed_field.rs
```diff
@@ -0,0 +1,352 @@
...
+    instance_id: String,
+}
+
+impl UnindexedField {
+    pub fn slow_search_threshold() -> Duration {
+        static SLOW_SEARCH_THRESHOLD: OnceLock<Duration> = OnceLock::new();
+
+        *SLOW_SEARCH_THRESHOLD.get_or_init(|| {
+            Duration::from_secs_f32(
+                std::env::var("QDRANT_SLOW_SEARCH_THRESHOLD")
```
agourlay, line 33: the env. variable does not specify the expected unit

tests/consensus_tests/test_issues_api.py
```diff
@@ -0,0 +1,137 @@
...
+    issues = get_issues(uri)
+    assert expected_issue_code not in [issue["id"] for issue in issues]
+
+    search_with_city_filter(uri)
+
+    issues = get_issues(uri)
+
+    # check the issue is now active
+    issue_present = False
+    for issue in issues:
```
agourlay, line 98: do we expect other issues in the array?

tests/consensus_tests/test_issues_api.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # check the timestamp has not changed
+    issue_present = False
+    for issue in issues:
+        if issue["id"] == expected_issue_code:
+            issue_present = True
+            assert timestamp == issue["timestamp"]
+            break
+    assert issue_present
+
```
agourlay, line 119: I would love to have asserts on the content of the `solution` before unpacking it.

It would us to detect when we change the format.

lib/collection/src/collection/search.rs
```diff
@@ -156,8 +162,19 @@ impl Collection {
...
+            .merge_from_shards(
+                all_searches_res,
+                Arc::clone(&request),
+                !shard_selection.is_shard_id(),
+            )
+            .await;
+
+        let filters_refs = request.searches.iter().map(|req| req.filter.as_ref());
+
+        self.post_process_if_slow_request(instant.elapsed(), filters_refs);
```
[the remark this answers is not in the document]
coszio: It is possible yes, but we cannot do it at `LocalShard` because we don't have the collection name anymore (#4121)

tests/consensus_tests/test_issues_api.py
coszio: We don't right now, as per the test structure, but this makes it resilient if we add other tests.

lib/segment/src/problems/unindexed_field.rs
coszio: -> `QDRANT_SLOW_SEARCH_SECS`?

tests/consensus_tests/test_issues_api.py
coszio: Asserted

lib/collection/src/collection/search.rs
```diff
@@ -266,4 +283,22 @@ impl Collection {
...
+
+    fn post_process_if_slow_request<'a>(
+        &self,
+        duration: Duration,
+        filters: impl Iterator<Item = Option<&'a Filter>>,
+    ) {
+        if duration > segment::problems::UnindexedField::slow_search_threshold() {
+            let filters = filters.flatten().cloned().collect::<Vec<_>>();
+
+            let schema = self.payload_index_schema.read().schema.clone();
```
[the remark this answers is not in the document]
coszio: I wasn't aware of that, but thanks for the clarification

lib/segment/src/problems/unindexed_field.rs
```diff
@@ -0,0 +1,357 @@
...
+impl UnindexedField {
+    pub fn slow_search_threshold() -> Duration {
+        static SLOW_SEARCH_THRESHOLD: OnceLock<Duration> = OnceLock::new();
+
+        *SLOW_SEARCH_THRESHOLD.get_or_init(|| {
+            Duration::from_secs_f32(
+                std::env::var("QDRANT_SLOW_SEARCH_SECS")
+                    .ok()
+                    .and_then(|var| var.parse::<f32>().ok())
+                    .unwrap_or(0.3),
```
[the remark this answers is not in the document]
coszio: That's fine for me

generall (approved):

PR #4140, opened 2024-04-30
Proactively fix lints for Clippy 1.78

agourlay:
Rust 1.78 is coming out in two days time https://releases.rs/docs/1.78.0/

Let's fix new Clippy lints proactively to ensure a smooth transition.


timvisee (approved):
And they are already green too! :raised_hands:

PR #4142, opened 2024-04-30
Remove unecessary Clippy allows

agourlay:
Maybe we changed our code or maybe Clippy changed the lints.

In any case we should not keep those escape hatches in place.


timvisee (approved):
I feel like there's a bunch more allows that are not needed anymore. But its quite labor intensive to test all of them. :smile:

agourlay:
@timvisee I am doing a partial scan at each Clippy release. We will slowly get there :+1:

PR #4148, opened 2024-05-01
Faster deleted filter in proxy segments

generall:
There is a hypothesis, that in case if we have a lot of modifications handled by ProxySegment,
query suffers from the overhead of the modified filtering:

1. It requires copying of the deleted list into the query on each request
2. Checking id presence during the search might be expensive if the deleted points hash-set is large

This PR introduce a new mechanics, which allows to modify deleted list via QueryContext, so that ProxySegment can keep the most relevant version of the deleted list and propagate it into the search via QueryContext object

Bonus: move is_stopped flag into query context, so it is less function arguments


generall:
Bench results:

Screenshot from 2024-05-05 10-36-55

Setup:

200K points indexed into single segment. `max_indexing_threads=1` to minimize impact of the optimizer.

2 parallel bfb processes, all single threaded:

- Upsert 100k points (overwrite existing) to trigger indexation and fill proxy segment with deleted points
- Search with `indexed_only=true` - so, no full-scan happening

Overall:

Max RPS without upsertion: ~220
Min RPS during upsertion:
- release: ~10
- branch: ~80

timvisee (approved):

Issue #4154, opened 2024-05-02
Optimizer panic 


generall:
Related to https://github.com/qdrant/qdrant/issues/4131

```
2024-05-02 10:57:46.344	
2024-05-02T08:57:46.344159Z  INFO actix_web::middleware::logger: 10.100.20.2 "GET /dashboard HTTP/1.1" 200 922 "https://cloud.qdrant.io/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36" 0.034867    
2024-05-02 10:57:41.923	
2024-05-02T08:57:41.923141Z  WARN collection::update_handler: Optimization task panicked, collection may be in unstable state: called `Option::unwrap()` on a `None` value    
2024-05-02 10:57:35.680	
2024-05-02T08:57:35.679919Z ERROR qdrant::startup: Panic occurred in file lib/segment/src/segment_constructor/segment_builder.rs at line 139: called `Option::unwrap()` on a `None` value    
2024-05-02 10:57:35.680	
    
2024-05-02 10:57:35.680	
  19: <unknown>
2024-05-02 10:57:35.680	
  18: <unknown>
2024-05-02 10:57:35.680	
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/sys/pal/unix/thread.rs:108:17
2024-05-02 10:57:35.680	
  17: std::sys::pal::unix::thread::Thread::new::thread_start
2024-05-02 10:57:35.680	
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/alloc/src/boxed.rs:2015:9
2024-05-02 10:57:35.680	
  16: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
2024-05-02 10:57:35.680	
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/alloc/src/boxed.rs:2015:9
2024-05-02 10:57:35.680	
  15: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
2024-05-02 10:57:35.680	
  14: core::ops::function::FnOnce::call_once{{vtable.shim}}
2024-05-02 10:57:35.680	
  13: std::sys_common::backtrace::__rust_begin_short_backtrace
2024-05-02 10:57:35.680	
[...]

PR #4156, opened 2024-05-02
Clean-up unversioned points

generall:
…oned points after WAL recovery

Fixes: https://github.com/qdrant/qdrant/issues/4154

### DIFF
--- a/lib/segment/src/segment_constructor/segment_builder.rs
+++ b/lib/segment/src/segment_constructor/segment_builder.rs
@@ -137,14 +137,15 @@ impl SegmentBuilder {
                         continue;
                     };
 
-                let other_version = other_id_tracker
-                    .internal_version(old_internal_id)
-                    .unwrap_or_else(|| {
+                let other_version = match other_id_tracker.internal_version(old_internal_id) {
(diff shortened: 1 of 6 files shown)


lib/segment/src/segment_constructor/segment_builder.rs
```diff
@@ -136,7 +136,15 @@ impl SegmentBuilder {
...
-                let other_version = other_id_tracker.internal_version(old_internal_id).unwrap();
+
+                let other_version = other_id_tracker
+                    .internal_version(old_internal_id)
+                    .unwrap_or_else(|| {
+                        log::debug!(
+                            "Internal version not found for internal id {old_internal_id}, using 0"
+                        );
+                        0
+                    });
```
timvisee, line 147: A nitpick, but in this case I'd rather use a match structure.

timvisee (approved):

lib/segment/src/segment_constructor/segment_builder.rs
generall: clippy wants if let

PR #4157, opened 2024-05-02
Missed lints for clippy 1.78

agourlay:
```
error: assigning the result of `Clone::clone()` may be inefficient
   --> lib/collection/src/collection_manager/holders/segment_holder.rs:126:9
    |
126 |         *self = source.clone();
    |         ^^^^^^^^^^^^^^^^^^^^^^ help: use `clone_from()`: `self.clone_from(source)`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#assigning_clones
    = note: `-D clippy::assigning-clones` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::assigning_clones)]`
```

This PR removes the function because it is equivalent to the default implementation in the trait.

### DIFF
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -121,10 +121,6 @@ impl Clone for LockedSegment {
             LockedSegment::Proxy(x) => LockedSegment::Proxy(x.clone()),
         }
     }
-
-    fn clone_from(&mut self, source: &Self) {
-        self.clone_from(source);
-    }
 }
 
 impl From<Segment> for LockedSegment {
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -123,7 +123,7 @@ impl Clone for LockedSegment {
     }
 
     fn clone_from(&mut self, source: &Self) {
-        *self = source.clone();
+        self.clone_from(source);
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -123,7 +123,7 @@ impl Clone for LockedSegment {
     }
 
     fn clone_from(&mut self, source: &Self) {
-        *self = source.clone();
+        self.clone_from(source);
```
coszio, line 126: This is a recursive call 🙈

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -123,7 +123,7 @@ impl Clone for LockedSegment {
...
     }
 }
```
timvisee, line 128: I also caught this one locally.

I think we should just remove this function, it is provided by the trait: https://doc.rust-lang.org/std/clone/trait.Clone.html#method.clone_from

timvisee (changes requested):

lib/collection/src/collection_manager/holders/segment_holder.rs
agourlay: I was too quick there :face_with_head_bandage:

coszio (approved):

timvisee (approved):

PR #4158, opened 2024-05-02
Instrument upsert points, optimization and flushing operations

ffuugoo:
This PR adds `tracing` instrumentation to upsert points operations, optimizer and flush workers.

Intended use is to enable this additional logs on chaos-testing, so that we can debug consistency failures.

`tracing` filtering [is a clusterfuck][tracing-filter-clusterfuck], so I've added a custom `inner = true` field to all added spans and events, so that we can turn them on/off easily (and independently of other log messages).

All added spans and events are disabled by default (so they would not spam the console; this logging *is* extensive).
They can be enabled with the following filters:
- `[{inner}]=info` - enable *all* added instrumentation
- `[upsert_points/insert]=info,[upsert_points/update]=info,[upsert_points/move]=info` - enable instrumentation related to upsert points operations
- `[optimize]=info` - enable instrumentation related to optimization
- `[flush_all]=info,[flush]=info` - enable instrumentation related to flushing
- E.g.: `QDRANT__LOG_LEVEL=error,[upsert_points/insert]=info,[optimize]=info,[flush]=info`

[tracing-filter-clusterfuck]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -180,6 +180,11 @@ impl ProxySegment {
 
         self.set_deleted_offset(point_offset);
 
+        tracing::info!(
+            tracing.target = "upsert_points",
+            "moved point {point_id} from wrapped into temporary segment",
+        );
+
         Ok(true)
     }
 
@@ -822,6 +827,9 @@ impl SegmentEntry for ProxySegment {
         let deleted_indexes_guard = self.deleted_indexes.read();
         let created_indexes_guard = self.created_indexes.read();
 
+        let _span = tracing::info_span!("flush", segment.id = self.id(), tracing.target = "flush",)
+            .entered();
+
         let wrapped_version = self.wrapped_segment.get().read().flush(sync, force)?;
         let write_segment_version = self.write_segment.get().read().flush(sync, force)?;
 
@@ -836,6 +844,11 @@ impl SegmentEntry for ProxySegment {
         };
 
         let _ = self.last_flushed_version.write().insert(flushed_version);
+
+        tracing::info!(
+            tracing.target = "flush",
+            "proxy segment flushed: {flushed_version}",
+        );
 
         Ok(flushed_version)
     }
@@ -991,6 +1004,14 @@ impl SegmentEntry for ProxySegment {
             .get()
             .read()
             .fill_query_context(query_context)
+    }
+
+    fn is_proxy(&self) -> bool {
+        true
+    }
+
+    fn tmp_path(&self) -> PathBuf {
+        self.write_segment.get().read().data_path()
     }
 }
 
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -45,6 +45,8 @@ pub struct ProxySegment {
     created_indexes: LockedFieldsMap,
     last_flushed_version: Arc<RwLock<Option<SeqNumberType>>>,
     wrapped_config: SegmentConfig,
+    data_path: PathBuf,
+    tmp_path: PathBuf,
 }
 
 impl ProxySegment {
@@ -63,7 +65,15 @@ impl ProxySegment {
             }
             LockedSegment::Proxy(_) => None,
         };
-        let wrapped_config = segment.get().read().config().clone();
+
+        let (wrapped_config, data_path) = {
+            let segment = segment.get();
+            let segment = segment.read();
+            (segment.config().clone(), segment.data_path().to_path_buf())
+        };
+
+        let tmp_path = write_segment.get().read().data_path().to_path_buf();
+
         ProxySegment {
             write_segment,
             wrapped_segment: segment,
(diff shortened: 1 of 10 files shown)


lib/segment/src/entry/entry_point.rs
```diff
@@ -212,4 +212,50 @@ pub trait SegmentEntry {
...
+        false
+    }
+
+    /// Returns path to the inner `write_segment`, if this segment is a `ProxySegment`.
+    /// Otherwise, returns empty [`PathBuf`].
+    ///
+    /// This default impl is overridden by `ProxySegment`.
+    fn tmp_path(&self) -> PathBuf {
+        PathBuf::new()
+    }
```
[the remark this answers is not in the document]
ffuugoo: It only makes sense for `ProxySegment`s. Regular segments don't have any write/tmp segment inside them, so I just return empty `PathBuf`.

ffuugoo:
<details>
<summary>Example of new instrumentation</summary>

Screenshot 2024-05-02 at 17 10 07

</details>

lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -863,6 +872,14 @@ impl SegmentEntry for ProxySegment {
...
             .fill_query_context(query_context)
     }
+
+    fn is_proxy(&self) -> bool {
+        true
+    }
+
+    fn tmp_path(&self) -> PathBuf {
+        self.write_segment.get().read().data_path()
+    }
```
xzfc, line 882: The result of this function is not stored anywhere and is used immediately. Would it make sense to return a non-owning type here to avoid clones? E.g. something like this: `fn tmp_path(&self) -> MappedRwLockReadGuard<'_, Path<'_>>`. It would require a similar change for `fn data_path()`.

lib/segment/src/entry/entry_point.rs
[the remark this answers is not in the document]
xzfc: A pair `(se.is_proxy(), se.tmp_path())` always is either `(false, "")` or `(true, "non-empty-path")`. Maybe it would make sense to combine these methods into one returning an `Option`?

lib/collection/src/collection_manager/holders/proxy_segment.rs
ffuugoo: Yeah, I think I'll make just that. Initially this was made for debug session, and I didn't care to refactor `SegmentEntry::data_path`, so I just rolled with `PathBuf`s everywhere.

But if we gonna merge this, I don't want to have allocating calls in the `upsert_points` path. I've tried to implement a check with `tracing` if particular span/event would be logged, and if not, avoid calling `SegmentEntry::id`... but neither of my approaches worked, so far. :/

lib/segment/src/entry/entry_point.rs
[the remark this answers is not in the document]
ffuugoo: Yeah, `Option` kinda works better, it was just easier not to deal with the `Option` during debug. I'll fix this.

lib/collection/src/collection_manager/holders/proxy_segment.rs
timvisee: So, having a `tracing::info!("{}", something())` is not enough to not call the `something()` if it is disabled?

ffuugoo: So, I haven't strictly tested all of this, but:
- for message format arguments, it *might* be enough
- but I'm not using `SegmentEntry::id` as a format argument, I use it as a "field"
- and `tracing` allows using fields for *filtering*
- so, I assume, that fields *have to* always be evaluated
  - well, maybe except when the span/event is disabled by the plain log-level
  - but I explicitly use fields to filter tracing data in this PR
  - so I assume `segment.id = segment.id()` in the tracing stuff is always evaluated

To make it even worse:
- I tried a few approaches to pre-filter instrumentation
- e.g., query "will this span/event be enabled given it has this log-level and these fields"
- but in all cases `tracing` *always* reported instrumentation as *enabled*
  - even though instrumentation was *not* logged, i.e., it *was* filtered out
- so, I'd assume, that with how PR is structured right now, `tracing` evaluates *all* expressions first
  - and filtering applied later, once everything is already evaluated

ffuugoo: _Partially_ fixed this. `SegmentEntry::id` is cheap to call now, though, I assume `segment.id = %segment.id()` in instrumentation still allocates.

ffuugoo: Refactored my "persistent" `SegmentId` type and related methods. It should not require allocations now. At least, not in most "usual" cases.

ffuugoo:
I'm using this branch for "missing payload" bug debug now. I think, once I'm done (just in case I'd like to introduce some additional changes), we can revisit and merge this PR. I've refactored `SegmentId` type once again and I think it's solid now.

Issue #4108, opened 2024-04-24
Handle Out-Of-Disk gracefully


generall:
**Is your feature request related to a problem? Please describe.**

Currently, there are situations, when Qdrant service can crush if it is not enough disk space to perform update operation.
This is sub-optimal behavior, as it should still be possible to respond to the search requests in this case.

**Describe the solution you'd like**

Add improve the handling of the situation, where qdrant faces out-of-disk problem. Instead of crashing, it should answer 500 to the user and still be able to process incoming search requests.

**Describe alternatives you've considered**

Block requests if the disk usage is above some threshold. This would require configuration of the arbitrary threshold and overall less desirable.

**Additional context**

We prepared an automated test scenario - https://github.com/qdrant/qdrant/pull/4105
Solution of this issue should include a PR into `test/low-disk-tests` branch, which makes the OOD test pass.

Rutik7066:
@generall I would like to solve this issue. Could you please assign me?

harishn1408:
### Proposed Solution

**Analysis**

The root cause of the issue is that Qdrant service crashes when it runs out of disk space during an update operation. This is because the service does not have a mechanism to handle out-of-disk situations gracefully, resulting in a crash. To improve this, we need to implement a feature that allows Qdrant to respond with a 500 error to the user while still processing incoming search requests.

**Implementation Plan**

1. **Add a disk usage monitoring mechanism**: Create a new module `qdrant/services/disk_monitor.py` that periodically checks the disk usage and reports it to the Qdrant service.
2. **Implement a disk usage threshold**: In `qdrant/services/disk_monitor.py`, define a threshold (e.g., 80%) above which the service should consider itself out of disk space.
3. **Create a handler for out-of-disk situations**: In `qdrant/services/handlers.py`, add a new handler `OutOfDiskHandler` that catches the `OutOfDiskError` exception raised by the disk monitor and returns a 500 error to the user.
4. **Modify the update operation to check disk space**: In `qdrant/services/update.py`, add a check before performing the update operation to see if the disk space is sufficient. If not, raise an `OutOfDiskError` exception.
5. **Update the test scenario**: In `test/low-disk-tests`, update the test to pass by mocking the disk monitor and handler.

**Code Changes**

* `qdrant/services/disk_monitor.py`:
```python
import psutil

class DiskMonitor:
    def __init__(self):
        self.threshold = 0.8  # 80% threshold

    def check_disk_usage(self):
[...]

PR #4165, opened 2024-05-03
Handle Out-Of-Disk gracefully 

kemkemG0:
fix: #4108
/claim #4108

### Description 
- Set a disk usage threshold, and perform a threshold(disk_buffer) check for Create and Update operations.
I haven't decided what value to set for `disk_buffer` , but `disk_buffer = wal_capacity_mb` might be a good idea, as I searched here (https://github.com/qdrant/qdrant/issues/4108#issuecomment-2092944158)

- I introduced a new crate (https://github.com/danburkert/fs2-rs) for getting free disk space where the specified path exists.

- Also, I updated the test created in #4105 so that it ensures it can search while out of disk. 

### TODO ?

I wasn't sure If I should fix  `collection create`, `vector create` and `snapshot create` in this PR, but i'm willing to fix them if needed : )

- [x] Merged #4105 into this branch
- [ ] collection create ?
- [x] collection update
- [ ] vector create ?
- [x] vector update
- [ ] snapshot create ?

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


tests/low-disk/create_and_search_items.py
```diff
@@ -0,0 +1,90 @@
...
+        exit(-3)
+    return resp.json()
+
+
+def initialize_qdrant(qdrant_host, collection_name, points_amount):
+    create_collection(qdrant_host, collection_name)
+    for points_batch in generate_points(points_amount):
+        insert_points(qdrant_host, collection_name, points_batch)
+        search_point(qdrant_host, collection_name)
+
```
kemkemG0, line 75: Ensure that search works after insertion

tests/low-disk/create_and_search_items.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    EXPECTED_ERROR_MESSAGE = "No space left on device"
+    if resp.status_code != 200:
+        if resp.status_code == 500 and EXPECTED_ERROR_MESSAGE in resp.text:
+            requests.put(f"{qdrant_host}/collections/{collection_name}/points?wait=true", json=batch_json)
+        else:
+            error_response = resp.json()
+            print(f"Points insertions failed with response body:\n{error_response}")
+            exit(-2)
+
```
kemkemG0, line 47: Either `status_code:200` or `status_code:500 + EXPECTED_ERROR_MESSAGE` is expected

kemkemG0:
@generall @timonv 
Hi, I think I'm ready for review ! :)

generall:
Hey @kemkemG0, thanks for the PR!

Before proceeding with the merge, I have a couple of follow-up questions regarding the proposed solution:

- It seems like fs2 depends on libc: https://github.com/danburkert/fs2-rs?tab=readme-ov-file#platforms, does it mean that our MUSL builds won't work? If so, is it possible to make this dependency and overall disk check optional?

- I tested the latency of the check - on my system (local SSD, ext4) it takes about `10µs` average, which is fine. Is there an estimation what should we expect on different systems? Especially network-mounted disks. It it might have impact on the overall performance, is there a way to make those checks less frequent?

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -213,6 +213,47 @@ impl LocalShard {
...
+    /// # Returns
+    /// A result indicating success (`Ok(())`) if there is sufficient disk space,
+    /// or a `CollectionError::service_error` if the disk space is insufficient or cannot be retrieved.
+    ///
+    /// # Errors
+    /// This function returns an error if:
+    /// - The disk space retrieval fails, detailing the failure reason.
+    /// - The available space is less than the configured WAL buffer size, specifying both the available and required space.
+    async fn ensure_sufficient_disk_space(&self) -> CollectionResult<()> {
+        let disk_free_space_bytes = fs2::available_space(self.path.as_path()).map_err(|err| {
```
generall, line 232: calling sync `available_space` in async context might be not so good performance-wise

generall:
I would take a look at https://github.com/al8n/fs4-rs

kemkemG0:
Hi @generall !

I took a look at fs4 and it seems great; it uses rustix and should be able to build with MUSL as well. I'll proceed with switching from fs2 to fs4. : )

While I'm not an expert on file systems or the `statvfs` syscall, I believe the increase in response time can be considered as the sum of `one RTT (Round-Trip Time)` + `one disk lookup` and I assume these statistics are typically cached in memory?

Also, quoting from a blog by tokio maintainer ryhl:
> "A good rule of thumb is no more than 10 to 100 microseconds between each .await."

Your benchmark shows about `10µs`, and it can be expected that responses on other systems will also come back within 100µs. Therefore, I believe that keeping it synchronous might be fine, but wrapping it with `tokio::task::spawn_blocking` might also be a good option.

As a means to reduce these checks, I thought about changing the `WAL flush` function so that checks are only performed when flushing. This would definitely reduce the number of calls, but since the `WAL` is used as an external crate, it could not be modified. If it were to be changed, it would involve forking and modifying the flush function to include the checks, but I'm worried this might be too complex for what we want to achieve.

Please let me know your thoughts or if there are any other aspects we should consider.

generall:
> As a means to reduce these checks, I thought about changing the WAL flush function so that checks are only performed when wal's flushing. 

Do you assume that the main source of the failures if WAL creating new segments?

timvisee:
> Do you assume that the main source of the failures if WAL creating new segments?

IIRC I've also seen OOD failures in RocksDB. But checking for at least WAL capacity probably prevents that from happening too.

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -213,6 +213,54 @@ impl LocalShard {
...
+            * 1024;
+
+        if disk_free_space_bytes < disk_buffer_bytes.try_into().unwrap_or_default() {
+            return Err(CollectionError::service_error(
+                "No space left on device: WAL buffer size exceeds available disk space".to_string(),
+            ));
+        }
+
+        Ok(())
+    }
```
timvisee, line 263: Requesting the available space may fail. In my opinion, we shouldn't fail but return early with `Ok(())` if that happens. That would prevent making Qdrant unusable if some platform or environment doesn't support this call properly.

kemkemG0:
@generall @timvisee 

I've been considering two possible scenarios:

1. The write to database itself crush.
2. After nearly exhausts the disk space, and then the OS issues a system call using the disk and crashes. In fact, when debugging with `println!`, the function unrelated to the write continues to work until just before it is called, but it doesn't print right after the call, suggesting it's not a panic but a crash.

- - - - 

> which gives almost 15% slowdown with the check.

> I am afraid that would be too much and we would need to find some better way to implement this

Additionally, I mentioned 

> I thought about changing the WAL flush function so that checks are only performed when wal's flushing.

This is because in the current my implementation, a `statvfs(3)` call is made with each update request, but by calling the statvfs syscall only when the wal is flushing, I believe it could be sped up by a factor of `wal_size/request_size`.

timvisee:
Looking at the above performance test, maybe we can make it less intensive:

- periodically query free disk space in a background job
- or query on something like 1% of the update operations

Once we hit the condition we can query on 100% of the update requests to immediately recover once disk space is available.

@generall Thoughts?

generall:
One alternative: make this check less frequent by using timeout and number of operations.

So for example, check the disk space only if it was more than one second or more than N operations was performed since the last check

xzfc:
> but since the WAL is used as an external crate, it could not be modified.

Wal is a Qdrant crate, you could send a PR to it. https://www.github.com/qdrant/wal

kemkemG0:
@generall @timvisee 
To summarize:

1. Integrate a disk usage check as part of the flush loop to regularly monitor disk usage.
2. Also, Perform the disk usage check every N iterations.
3. Save the checked usage by mutating a shared variable.
4. When updating or inserting into the database, refer to this variable.

I will try proceeding with the implementation based on this plan :)

@generall 
However, I feel that `1` might be sufficient, but why is `2` necessary?

kemkemG0:
@general @timvisee 

I implemented `DiskUsageWathcer` , could you check it please?

----

BenchMark:  125MBps, 1100IOPS  
`docker run --rm -it --network host qdrant/bfb:dev ./bfb -n 500000 -b 10 -d 128 -p 2 --uri 'http://127.0.0.1:6334' --skip-wait-index`
- this branch:
    - 65, 71, 64, 60, 64, 63 sec
- dev: 
    - 64, 65, 63, 62, 65, 68 sec

There is some variation, but they appear to be within the same range.

timvisee:
> However, I feel that `1` might be sufficient, but why is `2` necessary?

If the upsertion rate in terms of bytes is greater than the WAL size per flush interval, then it might fall through this check.

What are the above numbers? Do the represent the number of seconds for consecutive runs?

kemkemG0:
@timvisee 
I see, then I wonder what the N should be, since every request can different size of insertion and can't decided fixed N :(

> What are the above numbers? Do the represent the number of seconds for consecutive runs?
Yeah, they're the benchmarks on my end! I would like you to double check the benchmark as well to confirm if this branch's not that slow.

kemkemG0:
Another idea that might be realized by extending this implementation is to vary the frequency of updates based on free disk space.

For example, as far as I remember, the limit per request is 30MB, so how about an implementation where if there is more than 50MB(maybe 100MB is safer???) remaining, updates are made only during the flush loop, and if it's below 50MB, updates are made every time?

timvisee:
Good question. The request size can be measured to determine this, but that may be over-engineering it.

Maybe we can pick N=1000 for now to refresh every 1000 checks. A thousand vectors with dimensionality 1500 are 6 megabytes. I'd suggest implementing that first. We can fine tune this later.

generall (approved):

generall:
Hey @kemkemG0, we are going to merge it as-is now. Will also apply some minor updates later.

I like you idea about changing the check interval based on free disk size, I think we are going to follow it

PR #4167, opened 2024-05-03
Multivector flattened ram storage

IvanPleshkov:
This PR uses ideas from https://github.com/qdrant/qdrant/pull/4106 to refactor multivector RAM storage.

Currently, each multivector uses a separate allocated `Vec`. This PR uses `ChunkedVectors` from simple dense vector storage to store all multivector data in large flattened chunks.

All ideas and techniques as the same like in mmap storage.

No new unit test, multivector storage is already covered by tests

### DIFF
--- a/lib/segment/src/vector_storage/simple_multi_dense_vector_storage.rs
+++ b/lib/segment/src/vector_storage/simple_multi_dense_vector_storage.rs
@@ -9,6 +9,7 @@ use common::types::PointOffsetType;
 use parking_lot::RwLock;
 use rocksdb::DB;
 
+use super::chunked_vectors::ChunkedVectors;
 use crate::common::operation_error::{check_process_stopped, OperationError, OperationResult};
 use crate::common::rocksdb_wrapper::DatabaseColumnWrapper;
 use crate::common::Flusher;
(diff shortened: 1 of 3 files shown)


lib/segment/src/vector_storage/simple_multi_dense_vector_storage.rs
```diff
@@ -64,10 +75,25 @@ pub fn open_simple_multi_dense_vector_storage(
...
-        if point_id_usize >= vectors.len() {
-            vectors.resize(point_id_usize + 1, TypedMultiDenseVector::placeholder(dim));
+        if point_id_usize >= vectors_metadata.len() {
+            vectors_metadata.resize(point_id_usize + 1, Default::default());
         }
-        vectors[point_id_usize] = stored_record.vector;
+
+        let metadata = &mut vectors_metadata[point_id_usize];
+        metadata.size = stored_record.vector.flattened_vectors.len();
+        metadata.capacity = metadata.size;
```
agourlay, line 84: Why is the capacity equals to size here?
I thought it was the capacity left in the vector :thinking:

IvanPleshkov: `capacity` here is a size of allocated place in `ChunkedVectors`. When we rewrite vector, we can use old allocated place if new size is not larger than capacity

IvanPleshkov: Here it's equal because we allocate exact `size` vectors in `ChunkedVectors` container. When user rewrites vector with smaller multivector, size will be changed but capacity won't

agourlay (approved):

PR #4170, opened 2024-05-03
Remove pointer alignment checks redundant since Rust 1.78

timvisee:
We have some pointer alignment checks which are now redundant since the release of Rust 1.78.

Release notes: https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#asserting-unsafe-preconditions

Same assertion shown in the standard library: https://doc.rust-lang.org/src/core/slice/raw.rs.html#95

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/common/memory/src/mmap_ops.rs
+++ b/lib/common/memory/src/mmap_ops.rs
@@ -119,18 +119,6 @@ pub fn transmute_to_u8<T>(v: &T) -> &[u8] {
 pub fn transmute_from_u8_to_slice<T>(data: &[u8]) -> &[T] {
     debug_assert_eq!(data.len() % size_of::<T>(), 0);
 
-    debug_assert_eq!(
-        data.as_ptr().align_offset(align_of::<T>()),
-        0,
-        "transmuting byte slice {:p} into slice of {}: \
-         required alignment is {} bytes, \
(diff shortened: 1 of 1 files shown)


ffuugoo (approved):

ffuugoo:
Our assertions are more informative, so I'd still leave them. But PR itself is LGTM.

agourlay (approved):

timvisee:
> Our assertions are more informative, so I'd still leave them. But PR itself is LGTM.

That's true, though I don't expect to hit them again. If we get a panic from this later we can easily point to it and add it back if desired.

I'm just not super happy about doing this check twice, also because it makes those functions twice as big.

Wdyt? Keep or remove?

ffuugoo:
> > Our assertions are more informative, so I'd still leave them. But PR itself is LGTM.
> 
> That's true, though I don't expect to hit them again. If we get a panic from this later we can easily point to it and add it back if desired.
> 
> I'm just not super happy about doing this check twice, also because it makes those functions twice as big.
> 
> Wdyt? Keep or remove?

Personally, I'd leave them. But I'm fine either way.

timvisee:
Two people I asked both preferred to leave the extra checks, so I'm closing this issue

PR #4171, opened 2024-05-03
Simplify MaxSim configuration

agourlay:
This PR turns the `MultiVectorConfiguration` into an object to be able to extend it in the future.

I don't think the max similarity object will receive any configurations so it becomes a regular string enum.


docs/redoc/master/openapi.json
```diff
@@ -5724,15 +5724,11 @@
...
       },
       "MultiVectorConfig": {
-        "anyOf": [
-          {
-            "$ref": "#/components/schemas/MaxSimConfig"
-          }
+        "type": "string",
+        "enum": [
+          "max_sim"
         ]
```
coszio, line 5730: I currently don't know, but is this still compatible with `anyOf`? such that if we add more variants -which can be objects- can still be represented fine? Off the top of my mind I can't remember if we have something like this already

agourlay: I don't think we could have other non string type in the enum with this encoding.

agourlay: after thinking about your comment, I will push a different encoding to make sure we can add new fields to the `MultiVectorConfiguration` :)

tests/openapi/openapi_integration/test_multi_vector.py
```diff
@@ -30,7 +30,7 @@ def multivector_collection_setup(collection_name='test_collection'):
                     "size": 4,
                     "distance": "Dot",
                     "multi_vec_config": {
-                        "max_sim": {}
+                        "comparator": "max_sim"
```
agourlay, line 33: naming is hard, feel free to recommend something better for the field name :)

lib/segment/src/types.rs
```diff
@@ -738,20 +737,19 @@ pub enum VectorStorageDatatype {
...
+    /// How to compare multivector points
+    pub comparator: MultiVectorComparator,
 }
 
 #[derive(Debug, Default, Deserialize, Serialize, JsonSchema, Eq, PartialEq, Copy, Clone, Hash)]
-pub struct MaxSimConfig {}
+#[serde(rename_all = "snake_case")]
+pub enum MultiVectorComparator {
+    #[default]
+    MaxSim,
```
coszio, line 751: Smol description?
```suggestion
    /// Compare all individual vectors, use the maximum similarity
    #[default]
    MaxSim,
```

docs/redoc/master/openapi.json
coszio: I like the new approach, analog to the `distance` param

tests/openapi/openapi_integration/test_multi_vector.py
coszio: technically the name would be `distance` or `similarity` too, but that would be confusing. I cannot think of a better name than `comparator` right now.

lib/segment/src/types.rs
generall: this going to break OpenAPI generator

tests/openapi/openapi_integration/test_multi_vector.py
generall: the `multi_vec_config` also might be better. Rule of thumb: don't use more then one `_`

agourlay: I can propose `multivec_config` to have a single underscore.

agourlay: I am not a super fan of `comparator` either, it feels a bit too close to programming.

I remember @IvanPleshkov proposing `metric` instead of `comparator`.

That's one I could get behind if it gather votes.

generall (approved):

agourlay:
Decided to merge this PR as it improves the situation significantly.
We still have time to bikeshed naming in the future.

PR #4178, opened 2024-05-06
Chore: remove deprecated internal search API

coszio:
We had previously changed the response of this endpoint to always return an error. Since this only affects internal API, it is safe to remove it by now.


generall (approved):

agourlay (approved):

timvisee (approved):
image

Great!

PR #4192, opened 2024-05-07
Introduce Cargo feature "testing"

xzfc:
The Rust's limitation of an attribute `#[cfg(test)]` is that it makes items available only for unit tests in this crate, but not for integration tests, benchmarks, and unit tests from other crates.

A way to have these items in all tests/benchmarks, but not in the production builds is to hide them behind Cargo feature, e.g. `#[cfg(feature = "testing")]`.

This PR introduces feature `testing` for crates `collection`, `common`, `segment`. Additionally, and it hides several items under this attribute.


timvisee (approved):
Thanks for picking this up! It looks a lot better than I initially expected.

I believe you did, but I would prefer to keep using `#[cfg(test)]` in all possible cases.

xzfc:
Yes, strictly internal entries are still `#[cfg(test)]`.

.github/workflows/rust-lint.yml
```diff
@@ -36,4 +36,8 @@ jobs:
     - name: Check code formatting
       run: cargo +nightly fmt --all -- --check
     - name: Check cargo clippy warnings
+      run: cargo clippy --workspace -- -D warnings
+    - name: Check cargo clippy warnings for all targets
+      run: cargo clippy --workspace --all-targets -- -D warnings
+    - name: Check cargo clippy warnings for all targets and features
       run: cargo clippy --workspace --all-targets --all-features -- -D warnings
```
ffuugoo, line 43: Do we need all three of them? Doesn't `--workspace --all-targets` *strictly* supersedes `--workspace`? 🤔

IvanPleshkov: Step without `--all-features` allows us to find places where we try to use testing code in a non-testing environment. Clippy may fail in this case by non-existsting entities

ffuugoo: > Step without `--all-features`...

Yeah, this one I get. I'm asking about `--workspace` vs  `--workspace --all-targets`. It seems to me, like `--workspace` without `--all-targets` is not required.

xzfc: > --all-targets
> Check all targets. This is equivalent to specifying --lib --bins --tests --benches --examples.

Apparently, it enables features from `[dev-dependencies]` not just only for `--tests` and `--benches`, but also for `--lib` and `--bins`. I've added it to check that `--lib`/`--bins` compile with this feature disabled.

ffuugoo: Ok, got it. Thanks.

PR #4203, opened 2024-05-08
Allow basic multivec search on legacy API

agourlay:
This enables the support of basic multivec search on the current search API without changing the schema.

The trick is to expand a single dense vector into a multivector during conversion instead of returning an error.

A new search API will later be introduced to search using full fledged multivectors.

### DIFF
--- a/lib/segment/src/common/operation_error.rs
+++ b/lib/segment/src/common/operation_error.rs
@@ -17,8 +17,8 @@ pub const PROCESS_CANCELLED_BY_SERVICE_MESSAGE: &str = "process cancelled by ser
 #[derive(Error, Debug, Clone)]
 #[error("{0}")]
 pub enum OperationError {
-    #[error("Vector inserting error: expected dim: {expected_dim}, got {received_dim}")]
-    WrongVector {
+    #[error("Vector dimension error: expected dim: {expected_dim}, got {received_dim}")]
+    WrongVectorDimension {
         expected_dim: usize,
         received_dim: usize,
     },
--- a/lib/segment/src/data_types/vectors.rs
+++ b/lib/segment/src/data_types/vectors.rs
@@ -69,7 +69,7 @@ impl<'a> TryFrom<VectorRef<'a>> for &'a MultiDenseVector {
 
     fn try_from(value: VectorRef<'a>) -> Result<Self, Self::Error> {
         match value {
-            VectorRef::Dense(_) => Err(OperationError::WrongMulti),
+            VectorRef::Dense(_) => Err(OperationError::WrongMulti), // VectorRef::Dense cannot be converted to &MultiDense
             VectorRef::Sparse(_v) => Err(OperationError::WrongSparse),
             VectorRef::MultiDense(v) => Ok(v),
         }
@@ -116,7 +116,7 @@ impl TryFrom<Vector> for MultiDenseVector {
 
     fn try_from(value: Vector) -> Result<Self, Self::Error> {
         match value {
-            Vector::Dense(_) => Err(OperationError::WrongMulti),
+            Vector::Dense(v) => Ok(MultiDenseVector::new(v, 1)), // expand single dense vector
             Vector::Sparse(_) => Err(OperationError::WrongSparse),
             Vector::MultiDense(v) => Ok(v),
         }
@@ -369,7 +369,7 @@ impl<'a> TryInto<&'a MultiDenseVector> for &'a Vector {
 
     fn try_into(self) -> Result<&'a MultiDenseVector, Self::Error> {
         match self {
-            Vector::Dense(_) => Err(OperationError::WrongMulti),
+            Vector::Dense(_) => Err(OperationError::WrongMulti), // &Dense vector cannot be converted to &MultiDense
             Vector::Sparse(_) => Err(OperationError::WrongSparse),
             Vector::MultiDense(v) => Ok(v),
         }
--- a/lib/segment/src/data_types/vectors.rs
+++ b/lib/segment/src/data_types/vectors.rs
@@ -270,7 +270,7 @@ impl<T: PrimitiveVectorElement> TryFrom<Vec<TypedDenseVector<T>>> for TypedMulti
         let dim = value[0].len();
         // assert all vectors have the same dimension
         if let Some(bad_vec) = value.iter().find(|v| v.len() != dim) {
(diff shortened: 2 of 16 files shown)


lib/segment/src/common/operation_error.rs
```diff
@@ -17,8 +17,8 @@ pub const PROCESS_CANCELLED_BY_SERVICE_MESSAGE: &str = "process cancelled by ser
 #[derive(Error, Debug, Clone)]
 #[error("{0}")]
 pub enum OperationError {
-    #[error("Vector inserting error: expected dim: {expected_dim}, got {received_dim}")]
-    WrongVector {
+    #[error("Vector dimension error: expected dim: {expected_dim}, got {received_dim}")]
+    WrongVectorDimension {
```
IvanPleshkov, line 21: I like your renaming. We had to do it a long time ago =)

lib/segment/src/data_types/vectors.rs
```diff
@@ -68,7 +68,7 @@ impl<'a> TryFrom<VectorRef<'a>> for &'a MultiDenseVector {
 
     fn try_from(value: VectorRef<'a>) -> Result<Self, Self::Error> {
         match value {
-            VectorRef::Dense(_) => Err(OperationError::WrongMulti),
+            VectorRef::Dense(_) => Err(OperationError::WrongMulti), // VectorRef::Dense cannot be converted to &MultiDense
```
IvanPleshkov, line 71: Why `Vector` can be converted and `VectorRef` cannot be?

lib/segment/src/data_types/vectors.rs
```diff
@@ -68,7 +68,7 @@ impl<'a> TryFrom<VectorRef<'a>> for &'a MultiDenseVector {
 
     fn try_from(value: VectorRef<'a>) -> Result<Self, Self::Error> {
```
IvanPleshkov, line 69: I propose to remove this method. We have `TypedMultiDenseVectorRef<f32>` which replace `&MultiDenseVector` and can be converted from `VectorRef::Dense`

tests/basic_multivector_grpc_test.sh
```diff
@@ -242,4 +242,42 @@ if [[ $response != *"vectors count must be greater than 0"* ]]; then
...
+      "vector_name": "my-multivec"
+    }' $QDRANT_HOST qdrant.Points/Search 2>&1
+)
+
+if [[ $response != *"Wrong input: Vector dimension error: expected dim: 4, got 3"* ]]; then
+    echo Unexpected response, expected validation error: "$response"
+    exit 1
+fi
+
+response=$(
```
IvanPleshkov, line 260: Do we actually need to check `dim=3` and `dim=5` both? I find it equal

IvanPleshkov (changes requested):
I like that this feature was implemented by changing conventions only. Requested changes because I want to discuss `impl<'a> TryFrom<VectorRef<'a>> for &'a MultiDenseVector`

tests/basic_multivector_grpc_test.sh
agourlay: No, just wanted to make sure less and greater values were handled.

I can remove it if it bugs you :)

IvanPleshkov: Honestly, it bugs me a little bit. But it's so small nitpick that you may leave it as is

agourlay: https://github.com/qdrant/qdrant/pull/4203/commits/f714a8ef2def4bd0b0636d29b86c3e5a155e0734

lib/segment/src/data_types/vectors.rs
agourlay: Here we are implementing `impl<'a> TryFrom<VectorRef<'a>> for &'a MultiDenseVector`

A `VectorRef::Dense` contains a `&[VectorElement]` but `MultiDenseVector` requires a own vector.
This could be solved with a reallocation.

However the final type is `&MultiDenseVector` with an outer lifetime.

```
error[E0515]: cannot return value referencing temporary value
  --> lib/segment/src/data_types/vectors.rs:71:36
   |
71 |             VectorRef::Dense(v) => Ok(&MultiDenseVector::new(v.to_owned(), 1)),
   |                                    ^^^^--------------------------------------^
   |                                    |   |
   |                                    |   temporary value created here
   |                                    returns a value referencing data owned by the current function
```

I can't return a reference to the temporary object.

WDYT?

IvanPleshkov: We don't need `&MultiDenseVector`, use `TypedMultiDenseVectorRef<f32>` instead. It allows such conversion. And you can do `impl<'a> From<&'a MultiDenseVector> for TypedMultiDenseVectorRef<'a, f32>`

lib/segment/src/data_types/vectors.rs
```diff
@@ -115,7 +115,7 @@ impl TryFrom<Vector> for MultiDenseVector {
 
     fn try_from(value: Vector) -> Result<Self, Self::Error> {
         match value {
-            Vector::Dense(_) => Err(OperationError::WrongMulti),
+            Vector::Dense(v) => Ok(MultiDenseVector::new(v, 1)), // expand single dense vector
```
agourlay, line 118: this line is actually incorrect!
the second argument is the vector dimension and not the vector count :face_with_head_bandage:

IvanPleshkov: Cool that you noticed it )

agourlay: https://github.com/qdrant/qdrant/pull/4203/commits/0bfe07b6faed467c267306f39223961fe7afdc89

lib/segment/src/data_types/vectors.rs
IvanPleshkov: I tried to do it in a separate PR:
https://github.com/qdrant/qdrant/pull/4224
The result is sad. If `VectorRef` is converted into `MultiDenseVectorRef`, the user would use dense vectors as multivectors everywhere like in vector insertion.
On another hand, I cannot approve different conventions from `Vector` and `VectorRef`. It's too unexplicit

agourlay:
I have added a test showing that we support the single dense vector expansion on upserts as well.


Will merge on green :green_heart:

PR #4198, opened 2024-05-07
universal-query: Sketch request types pipeline

coszio:
Main idea: Each user request will be converted into a new type at each step of the search: API -> ToC -> Collection -> Shards.

This PR adds the types used at Shards level only, as well as a PoC of how the request will be handled at `LocalShard`.

Note that a significant amount of changed lines come from moving the `LocalShard` implementations of `do_search` `scroll` and `query` to their own files

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

PR #4209, opened 2024-05-09
universal-query: Introduce new `RawVector` and `VectorInput` messages

coszio:
Needs #4198 

Future-proofs the grpc query vector type to allow any shape of vector.

- `RawVector`: Only bare vector shapes
- `VectorInput`: Can include point ids

### All Submissions:

### New Feature Submissions:


lib/api/src/grpc/proto/points.proto
```diff
@@ -46,12 +46,36 @@ message SparseIndices {
...
   optional uint32 vectors_count = 3; // Number of vectors per multi vector
 }
 
+message DenseVector {
+  repeated float data = 1;
+}
+
+message SparseVector {
+  repeated float data = 1;
+  optional SparseIndices indices = 2;
```
agourlay, line 62: not optional maybe?

coszio: resolved :)

lib/api/src/grpc/proto/points.proto
```diff
@@ -46,12 +46,36 @@ message SparseIndices {
   repeated uint32 data = 1;
 }
 
+// Legacy vector format, which determines the vector type by the configuration of its fields.
```
agourlay, line 49: `Legacy` is always the system that pays the bills :older_man:

agourlay (approved):

lib/segment/a.out
ffuugoo, line 1: I've noticed that this PR pushed a binary into the repo. Is this something useful or an accident?

ffuugoo: Noticed, because I *somehow* also had this file on disk (but not checked into the repo) and `git pull` complained that this file will be overridden.

I guess, the fact that it was on disk on my machine, means it was produced (by build scripts?) at some point?

agourlay: https://github.com/qdrant/qdrant/pull/4243

Issue #4213, opened 2024-05-10
Tracking issue: resharding


timvisee:
This issue tracks progress on the resharding feature. It allows to dynamically grow the number of shards in a Qdrant collection.

### Tasks

- [x] #4216
- [x] #4236
- [x] #4255
- [x] #4238
- [x] #4239
- [x] #4217
- [x] #4305
- [x] #4301
- [x] #3499
- [x] #4290
- [x] #4346
- [x] #4349
- [x] #4351
- [x] #4256
- [x] #4355
- [x] #4344
- [x] #4373
- [x] #4379
- [x] #4381
- [x] #4480
- [x] #4417
- [x] #4491
- [x] #4506
- [x] #4501
- [x] #4507
- [x] #4503
- [x] #4493
- [x] #4509
- [x] #4515
- [x] #4516
- [x] #4524
- [x] #4561
- [x] #4565
- [x] #4597
- [x] #4598
- [x] #4319
- [x] #4608
- [x] #4609
- [x] #4642
- [x] #4617
- [x] #4681
- [x] #4666
- [x] #4682
- [x] #4687
- [x] #4704
- [x] #4725
- [x] #4707
- [x] #4699
- [x] #4697
- [x] #5005
- [x] #5019
- [x] #5034
- [x] #4724
- [x] #4736
- [x] #4741
- [x] #5522
- [x] https://github.com/qdrant/qdrant/pull/5736
- [x] #4599
- [x] #5718

### Backlog

A rough outline of new tasks on this feature:

- [x] deduplicate count during resharding
- [x] implement resharding "driver"
  - [x] dedicated task running on receiver node is driving resharding (@timvisee proposal)
- [x] add resharding filter for query API
- [x] on resharding start, clear any existing state
- [x] abort/cancel resharding
  - [x] when participating peer is removed
  - [x] is it possible to remove the only/last replica of a shard? if so, then we also need to cancel resharding when we remove shard replicas
  - [x] when resharding transfer fails
  - [x] when new shard is marked as `Dead`
    - [ ] ensure new shard *can* be marked as `Dead`, even if it's the *only* replica of a shard
- [x] report resharding in cluster info
- [x] report correct number of points transferred in resharding transfer
- [x] fix exact counts while resharding
- [x] ensure compatibility with 1.9.x patch versions that already persisted default resharding state on disk
- [x] extend shard transfer key, include target shard to uniquely identify resharding transfers to
- [x] support scaling down
[...]

PR #4217, opened 2024-05-10
Add shard transfer method for resharding

timvisee:

Adds a `ReshardingStreamRecords` shard transfer method that will be utilized as part of resharding.

### Tasks
- [x] Propagate new hash ring configuration into shard method
- [x] Only transfer points that fall into the new shard based on the hashring

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/transfer/driver.rs
+++ b/lib/collection/src/shards/transfer/driver.rs
@@ -6,6 +6,7 @@ use std::time::Duration;
 use parking_lot::Mutex;
 use tokio::time::sleep;
 
+use super::resharding_stream_records::transfer_resharding_stream_records;
 use super::snapshot::transfer_snapshot;
 use super::stream_records::transfer_stream_records;
 use super::transfer_tasks_pool::TransferTaskProgress;
@@ -60,6 +61,18 @@ pub async fn transfer_shard(
         // Transfer shard record in batches
         ShardTransferMethod::StreamRecords => {
             transfer_stream_records(
+                shard_holder.clone(),
+                progress,
+                shard_id,
+                remote_shard,
+                collection_name,
+            )
+            .await?;
+        }
+
+        // Transfer shard record in batches for resharding
+        ShardTransferMethod::ReshardingStreamRecords => {
+            transfer_resharding_stream_records(
                 shard_holder.clone(),
                 progress,
                 shard_id,
--- a/lib/collection/src/shards/transfer/driver.rs
+++ b/lib/collection/src/shards/transfer/driver.rs
@@ -45,13 +45,12 @@ pub async fn transfer_shard(
     temp_dir: &Path,
 ) -> CollectionResult<bool> {
     // The remote might target a different shard ID depending on the shard transfer type
-    let shard_id = transfer_config
-        .to_shard_id
-        .unwrap_or(transfer_config.shard_id);
+    let local_shard_id = transfer_config.shard_id;
+    let remote_shard_id = transfer_config.to_shard_id.unwrap_or(local_shard_id);
 
     // Initiate shard on a remote peer
     let remote_shard = RemoteShard::new(
-        shard_id,
+        remote_shard_id,
         collection_id.clone(),
         transfer_config.to,
         channel_service.clone(),
@@ -66,7 +65,7 @@ pub async fn transfer_shard(
             transfer_stream_records(
                 shard_holder.clone(),
                 progress,
-                shard_id,
+                local_shard_id,
                 remote_shard,
                 collection_name,
             )
@@ -78,7 +77,7 @@ pub async fn transfer_shard(
             transfer_resharding_stream_records(
                 shard_holder.clone(),
                 progress,
-                shard_id,
+                local_shard_id,
                 remote_shard,
                 collection_name,
             )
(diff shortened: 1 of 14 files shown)


timvisee:
@ffuugoo I believe all the necessary bits have been implemented. If you think this is sufficient, I'd suggest to merge and polish this further down the line as necessary.

lib/collection/src/shards/transfer/driver.rs
```diff
@@ -69,6 +73,18 @@ pub async fn transfer_shard(
...
+        ShardTransferMethod::ReshardingStreamRecords => {
+            transfer_resharding_stream_records(
+                shard_holder.clone(),
+                progress,
+                shard_id,
+                remote_shard,
+                collection_name,
+            )
+            .await?;
+        }
```
ffuugoo, line 86: Even thought it is checked in `validate_transfer`, I'd still make missing `to_shard_id` a hard error for `ReshardingStreamRecords` transfer:

```suggestion
        // Transfer shard record in batches for resharding
        ShardTransferMethod::ReshardingStreamRecords => {
            transfer_resharding_stream_records(
                shard_holder.clone(),
                progress,
                transfer_config.to_shard_id.ok_or_else(|| CollectionError::service_error("to_shard_id is None")),
                remote_shard,
                collection_name,
            )
            .await?;
        }
```

lib/collection/src/shards/transfer/driver.rs
```diff
@@ -43,7 +44,10 @@ pub async fn transfer_shard(
     snapshots_path: &Path,
     temp_dir: &Path,
 ) -> CollectionResult<bool> {
-    let shard_id = transfer_config.shard_id;
+    // The remote might target a different shard ID depending on the shard transfer type
+    let shard_id = transfer_config
+        .to_shard_id
+        .unwrap_or(transfer_config.shard_id);
```
ffuugoo, line 50: This does not seem right.

- we initialize `shard_id` to `to_shard_id`, and initialize `remote_shard` with it, which is correct...
- ...but then we pass the same `shard_id` (which is initialized to `to_shard_id`) into `transfer_resharding_stream_records`, which treats `shard_id` as a *source* shard id
- e.g., `transfer_resharding_stream_records` expects `shard_id` to be a local shard that has to be proxified, but `shard_id` is initialized to `to_shard_id` in this case, so we try to proxify a remote shard to itself :D

lib/collection/src/shards/transfer/resharding_stream_records.rs
```diff
@@ -0,0 +1,148 @@
...
+            .count_local(Arc::new(CountRequestInternal {
+                filter: None,
+                exact: true,
+            }))
+            .await?
+        else {
+            return Err(CollectionError::service_error(format!(
+                "Shard {shard_id} not found"
+            )));
+        };
```
ffuugoo, line 71:
```suggestion
        // TODO(resharding): Fix resharding points count?

        let Some(count_result) = replica_set
            .count_local(Arc::new(CountRequestInternal {
                filter: None,
                exact: true,
            }))
            .await?
        else {
            return Err(CollectionError::service_error(format!(
                "Shard {shard_id} not found"
            )));
        };
```

lib/collection/src/shards/transfer/resharding_stream_records.rs
```diff
@@ -0,0 +1,148 @@
...
+            .transfer_batch(offset, TRANSFER_BATCH_SIZE, Some(&hashring))
+            .await?;
+
+        {
+            let mut progress = progress.lock();
+            let transferred =
+                (progress.points_transferred + TRANSFER_BATCH_SIZE).min(progress.points_total);
+            progress.points_transferred = transferred;
+            progress.eta.set_progress(transferred);
+        }
```
ffuugoo, line 103:
```suggestion
        {
            // TODO(resharding): Fix resharding transfer progress?

            let mut progress = progress.lock();
            let transferred =
                (progress.points_transferred + TRANSFER_BATCH_SIZE).min(progress.points_total);
            progress.points_transferred = transferred;
            progress.eta.set_progress(transferred);
        }
```

ffuugoo (changes requested):
LGTM overall, but with a few mostly minor comments.

lib/collection/src/shards/transfer/driver.rs
ffuugoo: Disregard this. There's a further comment that obsoletes this one. Forgot to remove this one before posting a review.

lib/collection/src/shards/transfer/driver.rs
timvisee: Fixed in 782d78b99a3a7c739b6149015b1100621e621c6a

lib/collection/src/shards/transfer/resharding_stream_records.rs
ffuugoo: Ok, whatever then. :D

ffuugoo (approved):

lib/collection/src/shards/transfer/resharding_stream_records.rs
timvisee: I'll list it in our backlog so we can visit it later, when there's less pressing issues.

PR #4222, opened 2024-05-13
refactor disk usage checks

generall:
Instead of checking on flush, we would prefer to check _before_ inserting into WAL, but with smart heuristics.


generall:
Confirmed, that there are no performance degradation compared to the current dev

PR #4216, opened 2024-05-10
Add basic resharding types

ffuugoo:
Tracked in #4213.

This PR adds a few basic types/variants/fields that are required as the most fundamental building blocks for resharding.

__TODO:__
- [x] add basic structure with resharding state to `Collection`
- [x] add second hashring to `ShardHolder::rings`
- [x] add `Resharding` replica state

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #4236, opened 2024-05-14
Combine (re)sharding hash rings in a single type, update split-by-shard logic

timvisee:
Tracked in #4213.

This combines (re)sharding hash rings in a single type, replacing the single hash ring we had before. It provides a consistent interface for all scenarios, whether we have a single hash ring or multiple while resharding. It keeping most of the existing logic intact, and allows us to extend and/or tweak behavior in a single place in the future.

Here I propose to replace the initial approach with a separate hash ring on a shared resharding key introduced in <https://github.com/qdrant/qdrant/pull/4216>. It seems more future proof, easier to manage, and remains on the shard key level. When working with shards, we always namespace to a specific shard key and so it makes sense to combine all necessary state in a single type with a single lookup.

It doesn't only replace the way of storing a resharding hash ring, but also updates all point-to-shard routing to spread to two shards in case of resharding.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/operations/mod.rs
+++ b/lib/collection/src/operations/mod.rs
@@ -208,14 +208,25 @@ impl Validate for CollectionUpdateOperations {
     }
 }
 
-fn point_to_shard(point_id: ExtendedPointId, ring: &ShardHashRing) -> ShardId {
-    *ring
-        .get(&point_id)
-        .expect("Hash ring is guaranteed to be non-empty")
+/// Get the shards for a point ID
+///
+/// Normally returns a single shard ID. Might return multiple if resharding is currently in
+/// progress.
+///
+/// # Panics
+///
+/// Panics if the hash ring is empty and there is no shard for the given point ID.
+fn point_to_shards(point_id: &ExtendedPointId, ring: &ShardHashRing) -> Vec<ShardId> {
+    let shard_ids = ring.get(point_id);
+    assert!(
+        !shard_ids.is_empty(),
+        "Hash ring is guaranteed to be non-empty",
+    );
+    shard_ids
 }
 
 /// Split iterator of items that have point ids by shard
-fn split_iter_by_shard<I, F, O>(
+fn split_iter_by_shard<I, F, O: Clone>(
     iter: I,
     id_extractor: F,
     ring: &ShardHashRing,
@@ -226,8 +237,12 @@ where
 {
     let mut op_vec_by_shard: HashMap<ShardId, Vec<O>> = HashMap::new();
     for operation in iter {
-        let shard_id = point_to_shard(id_extractor(&operation), ring);
-        op_vec_by_shard.entry(shard_id).or_default().push(operation);
+        for shard_id in point_to_shards(&id_extractor(&operation), ring) {
+            op_vec_by_shard
+                .entry(shard_id)
+                .or_default()
+                .push(operation.clone());
+        }
     }
     OperationToShard::by_shard(op_vec_by_shard)
 }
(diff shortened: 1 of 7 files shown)


lib/collection/src/shards/shard_holder.rs
```diff
@@ -37,44 +39,78 @@ pub const SHARD_KEY_MAPPING_FILE: &str = "shard_key_mapping.json";
...
 
+pub type LockedShardHolder = RwLock<ShardHolder>;
+
 impl ShardHolder {
     pub fn new(collection_path: &Path) -> CollectionResult<Self> {
-        let mut rings = HashMap::new();
-        rings.insert(RingsKey::Default, HashRing::fair(HASH_RING_SHARD_SCALE));
+        let rings = HashMap::from([(
+            None,
+            ShardHashRing::from(HashRing::fair(HASH_RING_SHARD_SCALE)),
```
ffuugoo, line 112: I'd add `ShardHashRing::new` or `ShardHashRing::single` method for this. It's used in a bunch of places, seems annoying to have to type this every time.

ffuugoo (approved):
Overall, LGTM! Thanks for taking over it! 👍

My only real comment would be: we might *not* want to *always* split-by-shard to new hashring, when doing resharding. E.g., we might want to introduce explicit stages when we do (or do not) split-by-shard to the new shard.

I'd go and add explicit `.take(1)` to the `point_to_shards` results, so that this PR won't change current behavior *yet*.

And then we will introduce another PR that will update split-by-shard logic, so that it would properly split only when required.

lib/collection/src/operations/mod.rs
```diff
@@ -208,39 +208,54 @@ impl Validate for CollectionUpdateOperations {
...
 ) -> OperationToShard<Vec<O>>
 where
     I: IntoIterator<Item = O>,
     F: Fn(&O) -> ExtendedPointId,
 {
     let mut op_vec_by_shard: HashMap<ShardId, Vec<O>> = HashMap::new();
     for operation in iter {
-        let shard_id = point_to_shard(id_extractor(&operation), ring);
-        op_vec_by_shard.entry(shard_id).or_default().push(operation);
+        for shard_id in point_to_shards(&id_extractor(&operation), ring) {
```
ffuugoo, line 240: Basically, I propose we do this (in all places we use `point_to_shards`):

```suggestion
        for shard_id in point_to_shards(&id_extractor(&operation), ring).take(1) {
```

And then even if we start resharding (and initialize `ShardHashRing::Resharding`) it will not change split-by-shard behavior *yet*, it will still only split-by-shard to the old hashring.

(Maybe slight change to `ShardHashRing::get` required. I'm not sure if `dedup` preserves `old, new` order.)

ffuugoo: And also to `select_shards`, so that read operations are not redirected to new shard.

lib/collection/src/shards/shard_holder.rs
timvisee: Sounds good.

lib/collection/src/operations/mod.rs
timvisee: It is not entirely clear to me what you suggest.

Are you suggesting to just route to the original shard for now, and enable multiple shards at a later point in time?

If so it may be easier to change `get()` to just return a single shard ID all the time. That requires a change in just one place, versus in a few places we potentially miss.

lib/collection/src/shards/shard_holder.rs
timvisee: Implemented in ddb2066ee41c4b09a6dfdaa0d316c2793c4d9cb8

lib/collection/src/operations/mod.rs
timvisee: Implemented in 52b1a22f4f691bb5b2c4d9fb675b8cdcd390eeca

PR #4238, opened 2024-05-14
Add `ReshardingOperation::Start` consensus message

ffuugoo:
Tracked in #4213.

This PR adds `ReshardingOperation::Start` consensus message. And also (in the most trivial way) disables splitting read operations to the resharding shard.

Important TODO to be fixed in some future PR: consensus snapshot does not handle resharding in any way *yet*.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -1,3 +1,4 @@
+use segment::types::ShardKey;
 use serde::{Deserialize, Serialize};
 
 use crate::shards::shard::{PeerId, ShardId};
@@ -6,11 +7,16 @@ use crate::shards::shard::{PeerId, ShardId};
 pub struct State {
     pub peer_id: PeerId,
     pub shard_id: ShardId,
+    pub shard_key: Option<ShardKey>,
 }
 
 impl State {
     #[allow(dead_code)]
-    pub fn new(peer_id: PeerId, shard_id: ShardId) -> Self {
-        Self { peer_id, shard_id }
+    pub fn new(peer_id: PeerId, shard_id: ShardId, shard_key: Option<ShardKey>) -> Self {
+        Self {
+            peer_id,
+            shard_id,
+            shard_key,
+        }
     }
 }
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -4,13 +4,13 @@ use serde::{Deserialize, Serialize};
 use crate::shards::shard::{PeerId, ShardId};
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct State {
+pub struct ReshardingState {
     pub peer_id: PeerId,
     pub shard_id: ShardId,
     pub shard_key: Option<ShardKey>,
 }
 
-impl State {
+impl ReshardingState {
     #[allow(dead_code)]
     pub fn new(peer_id: PeerId, shard_id: ShardId, shard_key: Option<ShardKey>) -> Self {
         Self {
--- a/lib/collection/src/hash_ring.rs
+++ b/lib/collection/src/hash_ring.rs
@@ -1,4 +1,5 @@
 use std::hash::Hash;
+use std::mem;
 
 #[derive(Clone)]
 pub enum HashRing<T: Hash + Copy> {
@@ -10,6 +11,28 @@ pub enum HashRing<T: Hash + Copy> {
 }
 
 impl<T: Hash + Copy> HashRing<T> {
+    // `hashring::HashRing` does not implement `Clone`, so we have to do this... 🤦‍♀️
+    pub fn clone(&mut self) -> Self {
+        match self {
+            HashRing::Raw(ring) => {
+                let (this, other) = duplicate(mem::take(ring));
+                *ring = this;
+
+                Self::Raw(other)
+            }
+
+            HashRing::Fair { ring, scale } => {
+                let (this, other) = duplicate(mem::take(ring));
+                *ring = this;
+
+                Self::Fair {
+                    ring: other,
+                    scale: *scale,
+                }
+            }
+        }
+    }
+
     pub fn raw() -> Self {
         Self::Raw(hashring::HashRing::new())
     }
@@ -70,6 +93,22 @@ impl<T: Hash + Copy> HashRing<T> {
             HashRing::Fair { ring, .. } => ring.is_empty(),
         }
     }
+}
+
+// `hashring::HashRing` does not implement `Clone`, so we have to do this... 🤦‍♀️
+fn duplicate<T>(ring: hashring::HashRing<T>) -> (hashring::HashRing<T>, hashring::HashRing<T>)
+where
+    T: Clone + Hash,
+{
+    let nodes: Vec<_> = ring.into_iter().collect();
+
+    let mut first = hashring::HashRing::new();
+    first.batch_add(nodes.clone());
+
+    let mut second = hashring::HashRing::new();
+    second.batch_add(nodes);
+
+    (first, second)
 }
 
 #[cfg(test)]
--- a/lib/collection/src/hash_ring.rs
+++ b/lib/collection/src/hash_ring.rs
@@ -7,19 +7,16 @@ use crate::shards::shard::ShardId;
 const HASH_RING_SHARD_SCALE: u32 = 100;
 
 #[derive(Clone)]
-pub enum HashRing {
+pub enum HashRing<T = ShardId> {
     /// Single hashring
-    Single(Inner<ShardId>),
+    Single(Inner<T>),
 
     /// Two hashrings when transitioning during resharding
     /// Depending on the current resharding state, points may be in either or both shards.
-    Resharding {
-        old: Inner<ShardId>,
-        new: Inner<ShardId>,
-    },
+    Resharding { old: Inner<T>, new: Inner<T> },
 }
 
-impl HashRing {
+impl<T: Hash + Copy> HashRing<T> {
     /// Create a new single hashring.
     ///
     /// The hashring is created with a fair distribution of points and `HASH_RING_SHARD_SCALE` scale.
@@ -30,13 +27,13 @@ impl HashRing {
     /// Create a new resharding hashring, with resharding shard already added into `new` hashring.
     ///
     /// The hashring is created with a fair distribution of points and `HASH_RING_SHARD_SCALE` scale.
-    pub fn resharding(shard_id: ShardId) -> Self {
+    pub fn resharding(shard: T) -> Self {
         let mut ring = Self::Resharding {
             old: Inner::fair(HASH_RING_SHARD_SCALE),
             new: Inner::fair(HASH_RING_SHARD_SCALE),
         };
 
-        ring.add_resharding(shard_id);
+        ring.add_resharding(shard);
 
         ring
     }
@@ -52,7 +49,7 @@ impl HashRing {
         matches!(self, Self::Resharding { .. })
     }
 
-    pub fn add(&mut self, shard: ShardId) {
+    pub fn add(&mut self, shard: T) {
         match self {
             Self::Single(ring) => ring.add(shard),
             Self::Resharding { old, new } => {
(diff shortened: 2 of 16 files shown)


lib/collection/src/shards/shard_holder.rs
```diff
@@ -67,28 +70,65 @@ pub enum ShardHashRing {
...
         match self {
             Self::Single(ring) => ring.add(shard),
-            // When resharding, update the new ring
-            Self::Resharding { old: _, new } => new.add(shard),
+            Self::Resharding { old, new } => {
+                if new.get(&shard).is_none() {
+                    old.add(shard);
+                    new.add(shard);
+                }
+            }
```
ffuugoo, line 113: So, this design is the simplest way to make `ShardHashRing::add` work properly with `ShardHolder::load_shards`/`ShardHolder::add_shard` with minimal changes.

With this implementation, we can add our resharding shards to the hashring (with `ShardHashRing::add_resharding`) *before* calling `load_shards`/`add_shard`.

And then, we can just call `load_shards`/`add_shard` as usual (without any changes to signature/logic/order of calls/etc) and it will properly add all shards to *both* hashrings, and simply "skip" any hashring changes for the resharding shards that are already in the hashring.

lib/collection/src/shards/shard_holder.rs
```diff
@@ -67,28 +70,65 @@ pub enum ShardHashRing {
...
+
+            *self = Self::Resharding { old, new };
+        }
+
+        let Self::Resharding { new, .. } = self else {
+            unreachable!();
+        };
+
+        new.add(shard);
+    }
```
ffuugoo, line 130: - switch from `ShardHashRing::Single` to `ShardHashRing::Resharding`, _if required_
- add shard to the `new` hashring _only_

lib/collection/src/shards/shard_holder.rs
```diff
@@ -110,11 +150,16 @@ impl ShardHashRing {
 pub type LockedShardHolder = RwLock<ShardHolder>;
 
 impl ShardHolder {
-    pub fn new(collection_path: &Path) -> CollectionResult<Self> {
-        let rings = HashMap::from([(None, ShardHashRing::single())]);
+    pub fn new(collection_path: &Path, resharding: Option<ShardId>) -> CollectionResult<Self> {
+        let default_ring = resharding.map_or_else(ShardHashRing::single, ShardHashRing::resharding);
```
ffuugoo, line 154: `resharding` argument is propagated from `Collection`.

If resharding is in progress and `Collection` loaded some resharding state, then this single line will restore resharding hashring and resharding shard to the `new` hashring.

It's a bit unobvious, because `clippy` forced me to pass methods instead of closures. It's this:

```rust
let default_ring = resharding.map_or_else(
    // if `resharding` is `None` - initialize empty `single` hashring
    || ShardHashRing::single(),
    // if `resharding` is `Some(shard)` - initialize `resharding` hashring with `shard` already added to `new` hashring
    |shard| ShardHashRing::resharding(shard),
);
```

lib/collection/src/shards/shard_holder.rs
```diff
@@ -412,8 +505,12 @@ impl ShardHolder {
                 debug_assert!(false, "Do not expect empty shard selector")
             }
             ShardSelectorInternal::All => {
-                for (shard_id, shard) in self.shards.iter() {
-                    let shard_key = self.shard_id_to_key_mapping.get(shard_id);
+                for (&shard_id, shard) in self.shards.iter() {
+                    if self.resharding == Some(shard_id) {
+                        continue;
+                    }
```
ffuugoo, line 511: Ignore resharding shard when doing read operations. Otherwise, once you initialized the new shard, all read requests would be broken, because we did not fix read operations to work with resharding yet.

lib/collection/src/shards/shard_holder.rs
```diff
@@ -155,6 +201,51 @@ impl ShardHolder {
         Ok(())
     }
 
+    pub fn start_resharding(
+        &mut self,
+        shard_id: ShardId,
+        shard: ShardReplicaSet,
+    ) -> Result<(), CollectionError> {
+        let Some(default_ring) = self.rings.get_mut(&None) else {
```
timvisee, line 209: Let's pass the sharding key in here right away, instead of defaulting to `None` here.

lib/collection/src/shards/shard_holder.rs
timvisee: Instead I'd prefer to depend on the `Resharding` replica set state. I feel like that more closely matches the periods we do and do not want to read from the new shard.

lib/storage/src/content_manager/collection_meta_ops.rs
```diff
@@ -291,6 +291,11 @@ pub struct ChangeAliasesOperation {
 #[serde(rename_all = "snake_case")]
 pub struct DeleteCollectionOperation(pub String);
 
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Deserialize, Serialize)]
+pub enum ReshardingOperation {
+    Start { peer_id: PeerId, shard_id: ShardId },
```
timvisee, line 296: We probably do not want to provide the shard ID here, but select it automagically.

But I guess we can keep it for the time being.

IIRC the shard number **must** be the next one so all are in sequence, otherwise loading shards will fail on restart.

timvisee: This is not exposed to the user of course :man_facepalming: 

I see we have a good place to do it in https://github.com/qdrant/qdrant/pull/4239

lib/collection/src/hash_ring.rs
```diff
@@ -71,6 +94,22 @@ impl<T: Hash + Copy> HashRing<T> {
     }
 }
 
+// `hashring::HashRing` does not implement `Clone`, so we have to do this... 🤦‍♀️
```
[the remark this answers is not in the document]
ffuugoo: No reason, other than "this use-case was not considered". E.g., https://github.com/jeromefroe/hashring-rs/pull/20.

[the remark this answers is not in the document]
ffuugoo: > Also, can't we take a reference instead of having to do `mem` hacks?

As far as I can tell, no. There is no way to access nodes added into `HashRing`, other than *consume* it into iterator. There's no `fn iter(&self)` or something. Only `impl IntoIterator for HashRing`.

And there's also no `Clone`, so I can't *clone* it and then call `into_iter` on the clone.

lib/collection/src/shards/shard_holder.rs
ffuugoo: There's no sharding key, because we don't support resharding custom sharding keys.

Personally, I would prefer to implement resharding for default sharding key first, and then extend it for custom sharding if necessary.

lib/storage/src/content_manager/collection_meta_ops.rs
ffuugoo: Yeah, that's the consensus message and user don't have any control over the value of `shard_id` field.

lib/collection/src/shards/shard_holder.rs
ffuugoo: I don't think that's practical (or even possible).

Also, this is pure WIP code to disable querying resharded shard, so that read operations still work while we implement proper read operation handling.

lib/collection/src/shards/shard_holder.rs
ffuugoo: Another point would be: it does not *strictly* starts resharding. When we are loading collection/shard holder, resharding might already be in progress, so in that case we would call `add_resharding` during loading as well.

But I won't fight more over it. I'm cool with either.

lib/collection/src/hash_ring.rs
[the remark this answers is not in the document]
timvisee: I see the user in the linked PR having the same problem.

I've added a comment there to hopefully reinstate it.

lib/collection/src/shards/shard_holder.rs
timvisee: > because we don't support resharding custom sharding keys.

Why?

It's essentially the same, but on a different scope.

lib/collection/src/shards/shard_holder.rs
ffuugoo: Technically, yes, but `shard` (that is in scope in the loop) is a `ShardReplicaSet`.

So to query replica set state you have to reach into `ShardReplicaSet::replica_state` (which is `SaveOnDisk`, so you have to read-lock it first), and then you have to either specify `peer_id` (and I don't think we actually have one here to specify) or you have to loop over the whole replica set state hashmap.

ffuugoo: When doing resharding you *should* have only a single replica in the replica set. And this single replica *should* be in `Resharding` state (at least, it is for now). But it does seem a lot easier to just duplicate this info for now.

Maybe we come up with a better understanding what exactly should be "the source of truth" for resharding state later, I vote we leave this as-is for now.

lib/collection/src/shards/shard_holder.rs
ffuugoo: Because I did not implement it yet? :D

Also, I don't fully understand how custom sharding works (and if that would add any annoying corner cases to resharding),.

timvisee: > Also, I don't fully understand how custom sharding works

You can use any arbitrary key to create a different set of shards. 

In other words, each shard key, whether `None` or `Some(Key)` has their own set of shards. In all cases resharding would work exactly the same, but it does need to know about the current sharding key.

ffuugoo: Done.

lib/collection/src/shards/shard_holder.rs
```diff
@@ -155,6 +206,52 @@ impl ShardHolder {
...
+            self.resharding.is_none(),
+            "shard holder does not contain resharding hashring, but resharding field is {:?}",
+            self.resharding
+        );
+
+        if self.shards.contains_key(&shard_id) {
+            // TODO: `CollectionError::service_error`? 🤔
+            return Err(CollectionError::bad_request(format!(
+                "shard holder already contains shard {shard_id} replica set"
+            )));
```
timvisee, line 245: Yes, a service error is better in all of these cases. It's something internal.

ffuugoo: I kinda agree, but the trick is that `start_resharding` is called by consensus thread (when `Resharding::Start` message is applied). And if `start_resharding` returns *service* error, it will make consensus thread to panic/exit.

So I'm not *yet* 100% sure if any/some/all of these errors can be somehow safeguarded on another level (e.g., some validation when initially proposing `Resharding::Start`, or something).

The first two, most likely, can be made service errors, but I'm not sure about the third one. I propose to leave it as-is for now, and once we implement more of resharding, 1) we might restructure the code so these error conditions aren't there at all anymore, or 2) we get better understanding, which conditions can or cannot *ever* happen.

timvisee: > it will make consensus thread to panic/exit.

:disappointed:

lib/collection/src/shards/shard_holder.rs
```diff
@@ -67,26 +70,63 @@ pub enum ShardHashRing {
 }
 
 impl ShardHashRing {
```
generall, line 72: Maybe a good time to move to a separate file

lib/collection/src/collection/resharding.rs
```diff
@@ -6,11 +7,16 @@ use crate::shards::shard::{PeerId, ShardId};
 pub struct State {
```
generall, line 7: Let's maybe call it ReshardingState. Navigation and refactoring will be easier

ffuugoo: Just use `resharding::State`.... (but OK 🥲)

lib/collection/src/shards/shard_holder.rs
ffuugoo: Do you mind if I move this into `collection/src/hash_ring.rs` and rename `HashRing` (that is in `hash_ring.rs` already) into something like `RawOrFair` or just `Inner`?

ffuugoo: Done.

lib/collection/src/collection/resharding.rs
ffuugoo: Done.

Issue #4233, opened 2024-05-14
Issues API test failure: broker::tests::test_basic_use


timvisee:
## Current Behavior
One of the issue API tests seems to fail. The behavior is somewhat weird, because it only happens in certain contexts. When running the test exclusively all is right. When running it along with other tests, it fails. I expect some global state to be messing things up.

## Steps to Reproduce
1. `cargo test --workspace`
2. observe the test failure

## Expected Behavior
The test should fail, no matter in what context it is run.

## Possible Solution
Cleaning up the global state, or isolating the test. I've tried to add `crate::clear()` but that didn't solve it.

timvisee:
Fixed in <https://github.com/qdrant/qdrant/pull/4240>

PR #4240, opened 2024-05-14
Make `Broker` test independent of a global dashboard

coszio:
I could not reproduce #4233 locally, but making it not global-state dependent is a better idea


timvisee (approved):
I haven't seen this problem since. I like that it doesn't depend on global state anymore.

PR #4242, opened 2024-05-14
universal-query: Test and fix `PlannedQuery` conversion

coszio:
Needs #4245 

- Tests and fixes the `TryFrom<ShardQueryRequest> for PlannedQuery` implementation
- Adds `ScoringQuery::Rrf` variant

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/operations/universal_query/planned_query.rs
+++ b/lib/collection/src/operations/universal_query/planned_query.rs
@@ -58,42 +58,48 @@ impl TryFrom<ShardQueryRequest> for PlannedQuery {
 
     fn try_from(request: ShardQueryRequest) -> CollectionResult<Self> {
         let ShardQueryRequest {
+            prefetches,
             query,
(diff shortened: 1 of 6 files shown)


lib/collection/src/operations/universal_query/planned_query.rs
```diff
@@ -131,8 +162,8 @@ fn recurse_prefetches(
                         params,
                         limit,
                         offset: 0,
-                        with_payload: None,
-                        with_vector: None,
+                        with_payload: Some(WithPayloadInterface::Bool(false)),
```
agourlay, line 165: `None` is not good enough?

coszio: I figured it would be better to make it explicit :)

timvisee (approved):

agourlay (approved):

PR #4247, opened 2024-05-16
Introduce RRF scoring

agourlay:
This PR introduces the scoring infrastructure for Reciprocal Rank Fusion (RRF).

I followed the implementation of our [client](https://github.com/qdrant/qdrant-client/blob/master/qdrant_client/hybrid/fusion.py) to enable the future congruence tests.


lib/segment/src/common/reciprocal_rank_fusion.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    // track RFF scores per point id
+    let mut scores_by_id: HashMap<ExtendedPointId, f32> = HashMap::new();
+
+    // track scored points by id
+    let mut points_by_id: HashMap<ExtendedPointId, ScoredPoint> = HashMap::new();
+
+    for response in responses {
+        for (pos, score) in response.into_iter().enumerate() {
+            let rff_score = position_score(pos);
```
generall, line 38:
```suggestion
            let rrf_score = position_score(pos);
```

lib/segment/src/common/reciprocal_rank_fusion.rs
```diff
@@ -0,0 +1,168 @@
...
+            match scores_by_id.entry(score.id) {
+                Entry::Occupied(mut entry) => {
+                    // accumulate score
+                    *entry.get_mut() += rff_score;
```
generall, line 42:
```suggestion
                    *entry.get_mut() += rrf_score;
```

lib/segment/src/common/reciprocal_rank_fusion.rs
```diff
@@ -0,0 +1,168 @@
...
+                }
+                Entry::Vacant(entry) => {
+                    // init score
+                    entry.insert(rff_score);
```
generall, line 46:
```suggestion
                    entry.insert(rrf_score);
```

lib/segment/src/common/reciprocal_rank_fusion.rs
```diff
@@ -0,0 +1,168 @@
...
+                    // track point
+                    points_by_id.insert(score.id, score);
+                }
+            }
+        }
+    }
+
+    // sort by tracked scores
+    let mut scores = scores_by_id.into_iter().collect::<Vec<_>>();
+    scores.sort_by(|a, b| OrderedFloat(b.1).cmp(&OrderedFloat(a.1)));
```
coszio, line 56: Might be a little better to allocate the sorted vec instead.
```suggestion
    let mut scores = scores_by_id
        .into_iter()
        .sorted_by(|a, b| OrderedFloat(b.1).cmp(&OrderedFloat(a.1)))
        .collect::<Vec<_>>();
```

**Additional note**
For congruence tests, does it matter if we choose unstable sorting? With RRF we expect same-scored points to occur often

lib/segment/src/common/reciprocal_rank_fusion.rs
```diff
@@ -0,0 +1,168 @@
...
+        .all(|r| r.windows(2).all(|w| w[0].score >= w[1].score)));
+    fn position_score(position: usize) -> f32 {
+        1.0 / (position as f32 + RFF_RANKING_K)
+    }
+
+    // track RFF scores per point id
+    let mut scores_by_id: HashMap<ExtendedPointId, f32> = HashMap::new();
+
+    // track scored points by id
+    let mut points_by_id: HashMap<ExtendedPointId, ScoredPoint> = HashMap::new();
```
coszio, line 34: Why is it better to have parallel hashmaps instead of setting and updating the score in points_by_id?

lib/segment/src/common/reciprocal_rank_fusion.rs
```diff
@@ -0,0 +1,168 @@
...
+/// Each response must be sorted by score in descending order.
+///
+/// The output is a single sorted list of ScoredPoint.
+/// Does not break ties.
+#[allow(dead_code)]
+pub fn rrf_scoring(responses: Vec<Vec<ScoredPoint>>, limit: usize) -> Vec<ScoredPoint> {
+    // ensure each input is sorted
+    debug_assert!(responses
+        .iter()
+        .all(|r| r.windows(2).all(|w| w[0].score >= w[1].score)));
```
coszio, line 25: This part might need to take into account Order::LargeBetter, or Order::SmallBetter

lib/segment/src/common/reciprocal_rank_fusion.rs
agourlay: Great catch, this was an artifact from a previous approach without full ownership :+1:

lib/segment/src/common/reciprocal_rank_fusion.rs
agourlay: I removed this assert as it does not bring much value.

lib/segment/src/common/reciprocal_rank_fusion.rs
agourlay: I agree that there will be a lot of same scored points.

Going with unstable will make it faster but will make the congruence tests less easy.

I think we should go for unstable.

lib/segment/src/common/reciprocal_rank_fusion.rs
```diff
@@ -0,0 +1,168 @@
...
+use ahash::{HashMap, HashMapExt};
+use ordered_float::OrderedFloat;
+
+use crate::types::{ExtendedPointId, ScoredPoint};
+
+/// Reciprocal Rank Fusion (RRF) is a method for combining rankings from multiple sources.
+/// See https://plg.uwaterloo.ca/~gvcormac/cormacksigir09-rrf.pdf
+
+/// Mitigates the impact of high rankings by outlier systems
+const RFF_RANKING_K: f32 = 2.0;
```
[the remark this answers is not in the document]
coszio: According to the paper it does a little 
imagen

lib/segment/src/common/reciprocal_rank_fusion.rs
```diff
@@ -0,0 +1,153 @@
...
+    for response in responses {
+        for (pos, mut score) in response.into_iter().enumerate() {
+            let rrf_score = position_score(pos);
+            match points_by_id.entry(score.id) {
+                Entry::Occupied(mut entry) => {
+                    // accumulate score
+                    entry.get_mut().score += rrf_score;
+                }
+                Entry::Vacant(entry) => {
+                    score.score = rrf_score;
```
coszio, line 39: nit: maybe rename to `point` so that it says `point.score`?

coszio (approved):

PR #4251, opened 2024-05-16
universal-query: New `QueriedPoint` struct with extendable `Score` enum

coszio:
- Introduces a `QueriedPoint` struct, which is equivalent to a `ScoredPoint`, but has a `Score` enum, which can be used to be the output of scroll-like requests, like when specifying `order_by`.
- Changes the output elements of `query` to be of `QueriedPoint` type.
- Also adds this types as external grpc types and in internal grpc service

### DIFF
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -683,7 +683,7 @@ message QueriedPoint {
   PointId id = 1; // Point id
   map<string, Value> payload = 2; // Payload
   optional Score score = 3; // Score to order against other points, if not present, the point is ordered by its ID.
-  uint64 version = 4; // Last update operation applied to this point
+  optional uint64 version = 4; // Last update operation applied to this point
   optional Vectors vectors = 5; // Vectors to search
   optional ShardKey shard_key = 6; // Shard key
 }
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -674,17 +674,15 @@ message UpdateBatchResponse {
 
 message Score {
   oneof variant {
-    PointId id = 1;
-    float similarity = 2;
-    int64 int_value = 3;
-    double float_value = 4;
+    int64 int = 1;
+    double float = 2;
   }
 }
 
 message QueriedPoint {
   PointId id = 1; // Point id
   map<string, Value> payload = 2; // Payload
-  Score score = 3; // Score to order against other points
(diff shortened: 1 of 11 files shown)


lib/segment/src/types.rs
```diff
@@ -233,6 +234,89 @@ impl PartialEq for ScoredPoint {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub enum Score {
+    Id(PointIdType),
+    Similarity(ScoreType),
+    IntValue(IntPayloadType),
+    FloatValue(FloatPayloadType),
```
generall, line 242: I am not sure I agree that Id and Similarity should be separate. 
In my understanding any score should be possible to represent as a number. Even if we order by time, the score value should be timestamp as a number.

We can consider enum of i64 and float though, cause float64 by itself not always have enough capacity.

lib/api/src/grpc/proto/points.proto
```diff
@@ -672,6 +672,22 @@ message UpdateBatchResponse {
...
+    int64 int = 1;
+    double float = 2;
+  }
+}
+
+message QueriedPoint {
+  PointId id = 1; // Point id
+  map<string, Value> payload = 2; // Payload
+  optional Score score = 3; // Score to order against other points, if not present, the point is ordered by its ID.
+  uint64 version = 4; // Last update operation applied to this point
```
generall, line 686: let's also make this one optional

lib/segment/src/types.rs
```diff
@@ -233,6 +234,79 @@ impl PartialEq for ScoredPoint {
...
+
+impl Ord for QueriedPoint {
+    /// A point with a score is always better than point without a score
+    fn cmp(&self, other: &Self) -> Ordering {
+        match (&self.score, &other.score) {
+            (Some(a), Some(b)) => a.cmp(b),
+            (None, Some(_)) => Ordering::Less,
+            (Some(_), None) => Ordering::Greater,
+            (None, None) => self.id.cmp(&other.id),
+        }
```
timvisee, line 294: `Option` implements `Ord` where `None` is always less, so you could do this with a bit less branching:
 
```suggestion
        if self.score.is_some() || other.score.is_some() {
            self.score.cmp(&other.score)
        } else {
            self.id.cmp(&other.id)
        }
```

Or maybe you want this instead:

```rust
self.score.cmp(&other.score).then_with(|| self.id.cmp(&other.id))
```

coszio:
Closing in favor of #4291

timvisee (commented):
It got closed right when I was done :smiling_face_with_tear:

PR #4252, opened 2024-05-17
Rescore universal query

agourlay:
This PR adds the implementation for re-scoring with RRF or a query at the shard level.

I took extra care make this testable by setting up test fixtures for shard tests.

### DIFF
--- a/lib/collection/src/shards/local_shard/query.rs
+++ b/lib/collection/src/shards/local_shard/query.rs
@@ -34,19 +34,15 @@ impl LocalShard {
 
         // TODO(universal-query): implement batch scrolling
         let scrolls = &vec![];
-        let merged_results = self
-            .recurse_prefetch(
-                &request.merge_plan,
-                &core_results,
-                scrolls,
-                search_runtime_handle,
-                timeout,
-            )
-            .await?;
-
-        // TODO(universal-query): Implement with_vector and with_payload
-        let results = vec![merged_results];
-        Ok(results)
+        self.recurse_prefetch(
+            &request.merge_plan,
+            &core_results,
+            scrolls,
+            search_runtime_handle,
+            timeout,
+            0, // initial depth
+        )
+        .await
     }
 
     fn recurse_prefetch<'shard, 'query>(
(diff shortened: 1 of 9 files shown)


lib/collection/src/tests/shard_query.rs
```diff
@@ -0,0 +1,147 @@
...
+    let sources_scores = shard
+        .query(Arc::new(query), &current_runtime)
+        .await
+        .unwrap();
+
+    // only one inner result in absence of prefetches
+    assert_eq!(sources_scores.len(), 1);
+    // number of results is limited by the outer limit for rescoring
+    assert_eq!(sources_scores[0].len(), outer_limit);
+}
```
coszio, line 147: I would love to see a the case of 2+ prefetches with RRF rescore, to show that the output has two lists, not just one.

lib/collection/src/shards/local_shard/query.rs
```diff
@@ -25,18 +27,26 @@ impl LocalShard {
...
+                &request.merge_plan,
+                &core_results,
+                search_runtime_handle,
+                timeout,
+            )
+            .await?;
 
         // TODO(universal-query): Implement with_vector and with_payload
-        todo!()
+        let results = vec![merged_results];
```
coszio, line 40: This means that we are always returning 1 list, regardless of RRF

agourlay:
@coszio I have pushed a change to keep intermediate results in case of top level RRF rescoring.

I applied the merging properly at the lower level by tracking the depth of the recursion.

WDYT?

lib/collection/src/tests/shard_query.rs
```diff
@@ -0,0 +1,261 @@
...
+        .await
+        .unwrap();
+
+    // one score per prefetch
+    assert_eq!(sources_scores.len(), 1);
+
+    // in case of top level RFF we need to propagate intermediate results
+    // so the number of results is not limited by the outer limit at the shard level
+    // the first source returned all its inner results
+    assert_eq!(sources_scores[0].len(), inner_limit);
```
coszio, line 101: Good check :+1:

lib/collection/src/shards/local_shard/query.rs
```diff
@@ -30,65 +32,164 @@ impl LocalShard {
...
+                        )
+                        .await?
                     }
                 };
-                sources.push(vec);
+                sources.extend(vec);
             }
 
-            self.merge_prefetches(sources, &prefetch.merge).await
+            if depth == 0 && prefetch.merge.rescore == Some(ScoringQuery::Fusion(Fusion::Rrf)) {
```
coszio, line 90: TBH I thought it was going to be handled at `do_planned_query`, but this way there is less repetition. I like this.

coszio (approved):

PR #4256, opened 2024-05-17
Send updates to the new shard during resharding

ffuugoo:
Tracked in #4213.

This PR enables sending update to the created shard and adds basic error handling for failed updates.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -333,16 +333,15 @@ impl Collection {
             replica_set.peer_state(&peer_id),
         );
 
+        let current_state = replica_set.peer_state(&peer_id);
+
         // Validation:
         // 0. Check that `from_state` matches current state
 
-        if from_state.is_some() {
-            let current_state = replica_set.peer_state(&peer_id);
-            if current_state != from_state {
-                return Err(CollectionError::bad_input(format!(
-                    "Replica {peer_id} of shard {shard_id} has state {current_state:?}, but expected {from_state:?}"
-                )));
-            }
+        if from_state.is_some() && current_state != from_state {
+            return Err(CollectionError::bad_input(format!(
+                "Replica {peer_id} of shard {shard_id} has state {current_state:?}, but expected {from_state:?}"
+            )));
         }
 
         // 1. Do not deactivate the last active replica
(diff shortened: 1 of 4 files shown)


ffuugoo:
This still feels a bit "hacky", IMO, but I don't want to spend more time on this RN. I propose we merge it, and then improve on it once we started working on the resharding driver.

lib/collection/src/collection/mod.rs
```diff
@@ -392,11 +391,29 @@ impl Collection {
...
+                .cloned();
+
+            drop(shard_holder);
+
+            self.abort_resharding(peer_id, shard_id, shard_key).await?;
+
+            // TODO(resharding): Abort all resharding transfers!?
+
+            return Ok(());
+        }
```
timvisee, line 408: How about doing this in `sync_local_state`, like we do with shard transfers?

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -432,10 +446,12 @@ impl ShardReplicaSet {
                 continue;
             };
 
-            if peer_state != ReplicaState::Active && peer_state != ReplicaState::Initializing {
-                continue;
+            match peer_state {
+                ReplicaState::Active | ReplicaState::Initializing | ReplicaState::Resharding => (),
+                _ => continue,
             }
```
timvisee, line 452: Since we now have a match here anyway, shall we make it exhaustive?

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -432,10 +446,12 @@ impl ShardReplicaSet {
...
 
+            // TODO(resharding): 🤔
```
timvisee, line 454: Can we skip this for resharding, because our _driver_ arranges proper synchronization?

lib/collection/src/collection/mod.rs
ffuugoo: Can you elaborate? `abort_resharding` does not send any consensus messages. It actually aborts resharding locally. We also abort shard transfers on `Dead` state in the same way (just a few lines below selected snippet).

timvisee (approved):

lib/collection/src/collection/mod.rs
ffuugoo: Either user aborts explicitly, or the driver. When update fails we just propose peer `Dead` and abort when applying this state change.

Issue #4225, opened 2024-05-13
Tracking issue universal query


agourlay:
- internal gRPC API definition https://github.com/qdrant/qdrant/pull/4198
- gather all vectors from ids (with automatic batching)
- receive new gRPC query types (conversion)
- Toc -> ShardHolder -> Local & Remote (shardOp trait)
- send remote new gRPC query types (shard holder level)
- plan query from input query (local shard level)
  - testable
- execute plan query
- merging from local & remote (collection level)

agourlay:
The group API has been added, this issue can be closed :heavy_check_mark:

PR #4261, opened 2024-05-17
universal-query: Add `OrderBy` scoring, and add `scrolls` to `PlannedQuery`

coszio:
Tracked in #4225 

- Adds `ScoringQuery::OrderBy` variant
- Modifies `PlannedQuery` to also store a list of scroll alongside the core search batch
- Adjusts the try_from conversion accordingly
- Internal grpc `QueryShardPoints` also gets the order-by variant

### DIFF
--- a/lib/api/src/grpc/proto/points_internal_service.proto
+++ b/lib/api/src/grpc/proto/points_internal_service.proto
@@ -241,7 +241,7 @@ message QueryShardPoints {
     oneof score {
       RawQuery vector = 1; // (re)score against a vector query
       bool rrf = 2; // Reciprocal Rank Fusion
-      // TODO(universal-query): Add order-by
+      OrderBy order_by = 3; // Order by a field
     }
   }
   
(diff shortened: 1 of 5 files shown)


lib/api/src/grpc/proto/points_internal_service.proto
```diff
@@ -241,7 +241,7 @@ message QueryShardPoints {
     oneof score {
       RawQuery vector = 1; // (re)score against a vector query
       bool rrf = 2; // Reciprocal Rank Fusion
```
generall, line 243: Not related to this PR, but it might be better to use 
```
enum Fusion {
  RRF
}
```
instead of boolean?

coszio: I considered that but I thought it would be too much nesting. Considering I wasn't aware if the other fusion techniques are configurable

generall: booleans are much harder to extend

generall (approved):

PR #4262, opened 2024-05-18
[universal-search] refactoring suggestions

generall:
- Rename some internal structure
 - Use enum instead of boolean value for fusion

### DIFF
--- a/lib/collection/src/operations/universal_query/planned_query.rs
+++ b/lib/collection/src/operations/universal_query/planned_query.rs
@@ -12,8 +12,8 @@ use crate::operations::types::{
 };
 
 pub struct PlannedQuery {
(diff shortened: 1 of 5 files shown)


lib/collection/src/operations/universal_query/planned_query.rs
```diff
@@ -12,8 +12,8 @@ use crate::operations::types::{
 };
 
 pub struct PlannedQuery {
-    pub merge_plan: PrefetchPlan,
-    pub batch: Arc<CoreSearchRequestBatch>,
+    pub merge_plan: MergePlan,
```
coszio, line 15: I like this rename

coszio (approved):
Looks ok to me 👌. Thanks for the suggestions. RRF as enum is better too

PR #4267, opened 2024-05-20
Add test for OOD during indexing

tellet-q:
Automate a test scenario for OutOfDisk error during indexing.

Scenario:
Limit Qdrant container disk to ~10 MB
1. create collection with "indexing_threshold": 0 (disabled indexing)
2. add points
3. enable indexing
4. wait for indexing
5. send a search request

Expected behavior: indexing should gracefully fail without crushing a container and post a log message, the container should continue running and accept search requests. 

### All Submissions:

### DIFF
--- a/tests/low-disk/create_and_search_items.py
+++ b/tests/low-disk/create_and_search_items.py
@@ -40,19 +40,17 @@ def update_collection(qdrant_host, collection_name, collection_json: dict):
 
 
 def wait_for_status(qdrant_host, collection_name, status: str):
-    counter = 0
     curr_status = ""
-    while counter < 30:
+    for i in range(30):
         resp = requests.get(
             f"{qdrant_host}/collections/{collection_name}")
-        counter += 1
         curr_status = resp.json()["result"]["status"]
         if resp.status_code != 200:
             print(f"Collection info fetching failed with response body:\n{resp.json()}")
             exit(-1)
         if curr_status == status:
             print(f"Status {status}: OK")
-            return
+            return "ok"
         sleep(1)
         print(f"Wait for status {status}")
     print(f"After 30s status is not {status}, found: {curr_status}. Stop waiting.")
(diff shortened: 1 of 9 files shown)


tests/low-disk/create_and_search_items.py
```diff
@@ -1,44 +1,72 @@
...
+
+
+def wait_for_status(qdrant_host, collection_name, status: str):
+    counter = 0
+    curr_status = ""
+    while counter < 30:
+        resp = requests.get(
+            f"{qdrant_host}/collections/{collection_name}")
+        counter += 1
+        curr_status = resp.json()["result"]["status"]
```
timvisee, line 49: This is very much a nitpick: if we want to loop a limited number of times for retrying, maybe we can use a for loop to prevent having to update the counter yourself.

```suggestion
    for counter in range(30):
        resp = requests.get(
            f"{qdrant_host}/collections/{collection_name}")
        curr_status = resp.json()["result"]["status"]
```

tests/low-disk/create_and_search_items.py
```diff
@@ -67,22 +95,78 @@ def search_point(qdrant_host, collection_name):
...
+    collection_json = {
+        "vectors": {
+            "size": VECTOR_SIZE,
+            "distance": "Cosine"
+        }
+    }
+    create_collection(qdrant_host, collection_name, collection_json)
     for points_batch in generate_points(points_amount):
         insert_points(qdrant_host, collection_name, points_batch)
         search_point(qdrant_host, collection_name)
```
timvisee, line 108: Would it make more sense to insert all points first, and search after, using two separate loops? Or is this order required for this test?

I know this has not been changed in this PR, but I only noticed this now.

tests/low-disk/create_and_search_items.py
```diff
@@ -67,22 +95,78 @@ def search_point(qdrant_host, collection_name):
...
+        "wal_config": {
+            "wal_capacity_mb": 1
+        }
+    }
+    create_collection(qdrant_host, collection_name, collection_json)
+    for points_batch in generate_points(points_amount):
+        res = insert_points(qdrant_host, collection_name, points_batch, quit_on_ood=True)
+        if res == "ood":
+            break
+        search_point(qdrant_host, collection_name)
```
timvisee, line 139: I guess a similar comment as above, insert first then search?

timvisee (approved):

tests/low-disk/create_and_search_items.py
tellet-q: I don't think we even need a loop for the search.

tests/low-disk/create_and_search_items.py
tellet-q: see the reply above

PR #4282, opened 2024-05-21
Bump Rustls 0.23

agourlay:
This PR is an update to rustls 0.23 which is enabled by https://github.com/qdrant/qdrant/pull/4274

The [0.23 change log](https://github.com/rustls/rustls/releases/tag/v%2F0.23.0) mentions a new default  cryptography provider:

> Default cryptography provider changed to [aws-lc-rs](https://crates.io/crates/aws-lc-rs). Note that this has some implications on [platform support and build-time tool requirements](https://aws.github.io/aws-lc-rs/requirements/index.html) such as cmake on all platforms and nasm on Windows.
Support for ring continues to be available: set the ring crate feature.

I decided to stick to `ring` for the time being and keep the potential migration to `aws-ls-rs` disjoint to avoid potential build issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -213,7 +213,7 @@ dependencies = [
  "pin-project-lite 0.2.12",
  "rustls-pki-types",
  "tokio 1.37.0",
- "tokio-rustls 0.25.0",
+ "tokio-rustls 0.26.0",
  "tokio-util",
  "tracing",
 ]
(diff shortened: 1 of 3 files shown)


timvisee:
@agourlay it does fail on Windows CI at this time, so I assume the above quoted text is important.

https://github.com/qdrant/qdrant/actions/runs/9172171821/job/25218093190?pr=4282#step:7:679

agourlay:
Looks like I need to disable `aws_lc_rs` from the default feature set :+1:

agourlay:
The default features are:

- aws_lc_rs
- logging
- std
- tls12 

So I disabled the default features and added all but `aws_lc_rs`

PR #4283, opened 2024-05-21
Fix wrong multivectors count in storage

IvanPleshkov:
Vector storage should return vectors count, not all inner vectors count. Because `0..self.total_vector_count()` is a range of valid internal ID's. Also removed obsolete `OrderedFloat` usage


agourlay:
Can we cover this change by a test?
Like adding an assert on for the count on an existing test is fine for me.

IvanPleshkov:
> Can we cover this change by a test? Like adding an assert on for the count on an existing test is fine for me.

Done

agourlay (approved):

PR #4285, opened 2024-05-21
Promote anyhow to workspace

agourlay:
Centralize `anyhow` version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ chrono = { workspace = true }
 rand = "0.8.5"
 schemars = { workspace = true }
 itertools = { workspace = true }
-anyhow = "1.0.86"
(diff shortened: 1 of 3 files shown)


generall:
I would rather prefer to get rid of anyhow inside the storage instead

agourlay:
> I would rather prefer to get rid of anyhow inside the storage instead

It seems to be used only within the `consensus_manager` to add context messages on the consensus persistence application.

Do you prefer having a dedicated error variant instead?

generall:
> Do you prefer having a dedicated error variant instead?

I am thinking about it, yeah

agourlay:
The workspace promotion makes less sense if we are trying to decouple our library crates from anyhow.

PR #4290, opened 2024-05-21
Add `ReshardingOperation::Abort` consensus message

ffuugoo:
Tracked in #4213.

This PR:
- adds `ReshardingOperation::Abort` consensus message
- and `AbortResharding` operation to `update_collection_cluster` REST API
  - (currently not exposed in the OpenAPI spec)
- and also fixes incorrect serialization format of `StartResharding` operation for `update_collection_cluster` API
  - `#[serde(untagged)]` is annoying to work with 😒

Abort operation is structured so that it can be applied to a partial/inconsistent resharding state, so that you can cleanup a broken state if something went wrong.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -700,6 +700,41 @@ impl Collection {
         Ok(())
     }
 
+    pub async fn abort_resharding(
+        &self,
+        peer_id: PeerId,
+        shard_id: ShardId,
+        shard_key: Option<ShardKey>,
+    ) -> CollectionResult<()> {
+        let mut shard_holder = self.shards_holder.write().await;
+
+        if let Some(state) = self.resharding_state.read().deref() {
+            let is_in_progress = state.peer_id == peer_id
+                && state.shard_id == shard_id
+                && state.shard_key == shard_key;
+
+            if !is_in_progress {
+                return Err(CollectionError::bad_request(format!("TODO")));
+            }
+        } else {
+            log::warn!("resharding of collection {} is not in progress", self.id);
+        };
+
+        if shard_holder.get_shard(&shard_id).is_none() {
+            log::warn!("shard {shard_id} does not exist in collection {}", self.id);
+        }
+
+        shard_holder
+            .abort_resharding(shard_id, peer_id, shard_key.clone())
+            .await?;
+
+        self.resharding_state.write(|state| {
+            *state = None;
+        })?;
+
+        Ok(())
+    }
+
     pub async fn get_telemetry_data(&self, detail: TelemetryDetail) -> CollectionTelemetry {
         let (shards_telemetry, transfers) = {
             let mut shards_telemetry = Vec::new();
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -731,7 +731,7 @@ impl Collection {
             log::warn!(
                 "aborting resharding of collection {} ({peer_id}/{shard_id}/{shard_key:?}), \
                  but shard {shard_id} does not exist in collection",
-                 self.id,
+                self.id,
             );
         }
 
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -714,20 +714,33 @@ impl Collection {
                 && state.shard_key == shard_key;
 
             if !is_in_progress {
-                return Err(CollectionError::bad_request(format!("TODO")));
+                return Err(CollectionError::bad_request(format!(
+                    "resharding of collection {} is not in progress",
+                    self.id,
+                )));
             }
         } else {
-            log::warn!("resharding of collection {} is not in progress", self.id);
+            log::warn!(
+                "aborting resharding of collection {} ({peer_id}/{shard_id}/{shard_key:?}),\
+                 but resharding is not in progress",
+                self.id,
+            );
         };
 
         if shard_holder.get_shard(&shard_id).is_none() {
-            log::warn!("shard {shard_id} does not exist in collection {}", self.id);
+            log::warn!(
+                "aborting resharding of collection {} ({peer_id}/{shard_id}/{shard_key:?}), \
+                 but shard {shard_id} does not exist in collection",
+                 self.id,
+            );
         }
 
+        // TODO: Contextualize errors? 🤔
         shard_holder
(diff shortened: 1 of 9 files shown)


lib/collection/src/collection/mod.rs
```diff
@@ -699,6 +703,54 @@ impl Collection {
...
+                self.id,
+            );
+        }
+
+        // TODO: Contextualize errors? 🤔
+        shard_holder
+            .abort_resharding(shard_id, peer_id, shard_key.clone())
+            .await?;
+
+        // TODO: Contextualize errors? 🤔
```
timvisee, line 746:
```suggestion
```

lib/collection/src/operations/cluster_ops.rs
```diff
@@ -196,3 +204,12 @@ pub struct AbortShardTransfer {
     pub to_peer_id: PeerId,
     pub from_peer_id: PeerId,
 }
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Validate)]
+pub struct StartResharding {
+    pub peer_id: Option<PeerId>,
```
timvisee, line 210: Is this peer ID the preferred target peer for resharding? Where it would otherwise just pick a semi-random one?

lib/collection/src/operations/cluster_ops.rs
```diff
@@ -196,3 +204,12 @@ pub struct AbortShardTransfer {
...
+    pub shard_key: Option<ShardKey>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Validate)]
+pub struct AbortResharding {}
```
timvisee, line 215: Don't we need `shard_id`, `peer_id` and `shard_key` fields here, or at least some of them?

lib/collection/src/shards/shard_holder.rs
```diff
@@ -165,6 +165,57 @@ impl ShardHolder {
...
+                Some(ReplicaState::Resharding) => {
+                    log::debug!("removing peer {peer_id} from {shard_id} replica set");
+                    shard.remove_peer(peer_id).await?;
+                }
+
+                Some(state) => {
+                    return Err(CollectionError::bad_request(format!(
+                        "peer {peer_id} is in {state:?} state"
+                    )));
+                }
```
timvisee, line 195: We probably also want to allow dead state?

lib/collection/src/operations/cluster_ops.rs
timvisee: I now see we just take these from the current resharding state. At this point we only accept one of them to happen at a time.

We might add those fields if we eventually support multiple operations together.

lib/collection/src/operations/cluster_ops.rs
ffuugoo: Yes.

lib/collection/src/operations/cluster_ops.rs
ffuugoo: I did want to add explicit parameters, so that you can specify stuff explicitly to safe-guard from accidents and/or try to fix broken state (because internally you *have to* specify peer/shard/shard key to call `abort_resharding` methods)... but it was just way too annoying to handle, so I decided to add the most basic abort first, and I'll maybe extend this later.

lib/collection/src/shards/shard_holder.rs
ffuugoo: Hmmm, good point. I'd expect `Dead` peers to automatically stop resharding, but one of the points of `Abort` is to be able to resolve broken state.

I'll see if there's a **neat/safe** way to add this. You would not want to accidentally yeet *any* random `Dead` peer using this.

timvisee: If that shard ID is in the resharding state, I think it should be fine to yeet it in dead state.

ffuugoo: Done. Might have overcomplicated this check a bit, but we store resharding state in a bunch of places, so I had to come up with something.

lib/collection/src/collection/mod.rs
ffuugoo: I'd leave this for now. Will add `TODO(resharding)` tag though.

lib/collection/src/hash_ring.rs
```diff
@@ -121,10 +121,11 @@ impl<T: Hash + Copy> HashRing<T> {
...
                 );
 
-                // TODO(resharding): `return` here?
+                false
             }
-        }
+        };
 
+        // TODO(resharding): Improve old/new hashrings equality check!?
         if old.len() == new.len() {
```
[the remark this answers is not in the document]
ffuugoo: Well, we can, but it will be ugly af. Same issue as with `Clone` and `iter`: neither is implemented on `hashring::HashRing`. So to implement `PartialEq` we'd have to clone both hashrings, then call `into_iter` on them and compare elements of the iterators.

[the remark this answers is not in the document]
timvisee: Let's upstream it!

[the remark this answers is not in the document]
ffuugoo: Yeah, my thoughts as well. Seems easier to just send a PR with `#[derive(PartialEq)]`.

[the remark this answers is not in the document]
timvisee: I've submitted a PR here: https://github.com/jeromefroe/hashring-rs/pull/27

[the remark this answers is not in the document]
timvisee: I temporarily patched this in 3f7a5e1ff39bfdd547000bf14a6f61e5038cfe28

lib/storage/src/content_manager/collection_meta_ops.rs
```diff
@@ -298,6 +298,12 @@ pub enum ReshardingOperation {
         shard_id: ShardId,
         shard_key: Option<ShardKey>,
     },
+
+    Abort {
+        peer_id: PeerId,
```
generall, line 303: Do we need this params only to be sure we are not cancelling something outdated?

ffuugoo: Yes.

generall (approved):
This is an explicit check of the state, I must say.
I hope it would make debugging easier.

PR #4278, opened 2024-05-20
Migrate score field to `Score` enum

coszio:
Supersedes #4251 (potentially)

- fix ordering in groups api for discover and best_score reco
- decouple `GroupsResult` from rest interface
- use `Option<Score>` for `ScoredPoint`
- prepare grpc scored point for migration

PR #4291, opened 2024-05-21
universal-query: Introduce `order_value` field in `ScoredPoint`

coszio:
Tracked in #4225 
Supersedes #4251 #4278 

- Fix ordering in group-by for custom scoring
- Rename `OrderingValue` -> `OrderedValue`
- Introduce `order_value` field in `ScoredPoint` (internal, grpc, and rest)

### DIFF
--- a/docs/grpc/docs.md
+++ b/docs/grpc/docs.md
@@ -150,6 +150,7 @@
     - [NamedVectors.VectorsEntry](#qdrant-NamedVectors-VectorsEntry)
     - [NestedCondition](#qdrant-NestedCondition)
     - [OrderBy](#qdrant-OrderBy)
+    - [OrderedValue](#qdrant-OrderedValue)
     - [PayloadExcludeSelector](#qdrant-PayloadExcludeSelector)
     - [PayloadIncludeSelector](#qdrant-PayloadIncludeSelector)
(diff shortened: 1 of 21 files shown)


docs/grpc/docs.md
```diff
@@ -3203,6 +3220,7 @@ For example, if `oversampling` is 2.4 and `limit` is 100, then 240 vectors will
 | version | [uint64](#uint64) |  | Last update operation applied to this point |
 | vectors | [Vectors](#qdrant-Vectors) | optional | Vectors to search |
 | shard_key | [ShardKey](#qdrant-ShardKey) | optional | Shard key |
+| order_value | [OrderedValue](#qdrant-OrderedValue) | optional | Order by value |
```
agourlay, line 3223: Can we make the field name and the field type consistent?
Like order vs ordered?

generall: I have slight preference towards `order_value`

coszio: I don't have a strong opinion to keeping different names, but it made sense for me to name it after OrderedFloat

generall: I think it is a bit different semantics here. We are not making Value ordered, we are using Value as order

agourlay (approved):

PR #4299, opened 2024-05-22
Add vectors and quantization options and defaults

JojiiOfficial:
Extends the configs parameters for some collection related defaults

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -163,21 +163,20 @@ storage:
 
   # Default parameters for collections
   collection:
+    # Number of replicas of each shard that network tries to maintain
+    replication_factor: 1
+
+    # How many replicas should apply the operation for us to consider it successful
+    write_consistency_factor: 1
+
     # Default parameters for vectors.
     vectors:
       # Whether vectors should be stored in memory or on disk.
       on_disk: null
 
-    consensus:
-      # Number of replicas of each shard that network tries to maintain
-      replication_factor: 1
-
-      # How many replicas should apply the operation for us to consider it successful
-      write_consistency_factor: 1
-
-  # Default quantization configuration.
-  # More info: https://qdrant.tech/documentation/guides/quantization
-  quantization: null
+    # Default quantization configuration.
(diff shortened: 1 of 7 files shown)


config/config.yaml
```diff
@@ -241,6 +250,12 @@ cluster:
     # We encourage you NOT to change this parameter unless you know what you are doing.
     tick_period_ms: 100
 
+    # Number of replicas of each shard that network tries to maintain
+    replication_factor: 1
+
+    # How many replicas should apply the operation for us to consider it successful
+    write_consistency_factor: 1
```
generall, line 257: Not sure if those should be part of the `ConsensusConfig`. It is not a part of `StorageConfig`, so it would require to change a lot of code to pass it deep enough to use in https://github.com/qdrant/qdrant/blob/28dee3821f473a4cc15875ef8e4652e52e4cf9ed/lib/storage/src/content_manager/toc/create_collection.rs#L25

config/config.yaml
```diff
@@ -161,6 +161,15 @@ storage:
   # More info: https://qdrant.tech/documentation/guides/distributed_deployment/#shard-transfer-method
   shard_transfer_method: null
 
+  # Default parameters for vectors.
+  vectors_config:
+    # Whether vectors should be stored in memory or on disk.
+    on_disk: false
```
timvisee, line 167: Thought it would effectively be the same, we might want to prefer `null` as default here to not set it at all.

Wdyt? @generall

config/config.yaml
timvisee: How about putting them under `storage.collection` or `storage.collection_params`?

config/config.yaml
generall: +

generall (approved):
Some changes from my side:

- remove redundant derives `JsonSchema` and `Serialize`
- use unique names for config structures to avoid confusion
- use single source of truth for default values
- Add shard_number parameter

timvisee (approved):

PR #4300, opened 2024-05-22
Query fetches payload and vector if necessary

agourlay:
Hydrate query scores with `payload` and `vector` if necessary.

### DIFF
--- a/lib/collection/src/shards/local_shard/query.rs
+++ b/lib/collection/src/shards/local_shard/query.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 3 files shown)


lib/collection/src/shards/local_shard/query.rs
```diff
@@ -32,15 +35,60 @@ impl LocalShard {
...
+        // fetch payload and/or vector for scored points if necessary
+        if request.with_payload.is_required() || request.with_vector.is_enabled() {
+            // ids to retrieve
+            let mut point_ids = Vec::new();
+            for scored_point in scored_points.iter().flatten() {
+                point_ids.push(scored_point.id);
+            }
+            // remove duplicates
+            point_ids.sort_unstable();
+            point_ids.dedup();
```
coszio, line 58: isn't it better to use a `HashSet` in this case?

agourlay: The number of points here is rather small, so I decided to go with the `vec` to avoid creating a lot of small `hashset` which consume more memory.

coszio: If we do group-by in the same high-level way we are doing it right now, it will be normal to have a high limit. It will be `limit` * `group_size` in each iteration

agourlay: Alright, I switched to a HashSet :)

lib/collection/src/shards/local_shard/query.rs
```diff
@@ -32,15 +33,43 @@ impl LocalShard {
...
+                self.segments(),
+                &point_ids,
+                &WithPayload::from(&request.with_payload),
+                &request.with_vector,
+            )?;
+
+            // update scored points in place
+            for (scored_point, record) in scored_points.iter_mut().flatten().zip(records.iter()) {
+                scored_point.payload.clone_from(&record.payload);
+                scored_point.vector.clone_from(&record.vector);
```
coszio, line 68: Is it necessary to clone?

coszio (approved):

PR #4301, opened 2024-05-22
Deduplicate points for shard reads

timvisee:

Deduplicate points based on their point ID in all reads from shards.

This is necessary for resharding, where multiple shards may have a single point ID.

The idea was to only deduplicate this while resharding is active, but after a discussion with @generall we decided to enable it at all times. We might benefit from this in other cases too, such as when we have the same point ID across multiple shard keys. At this time it is unclear yet what performance impact that might have, so that's yet to be tested.

### Tasks
- [x] Add tests for every deduplication scenario

### Open questions
- Do we need to handle read errors in any special way?
  - e.g., *ignore* read errors from new shard?

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/point_ops.rs
+++ b/lib/collection/src/collection/point_ops.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use futures::stream::FuturesUnordered;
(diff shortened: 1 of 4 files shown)


coszio (approved):
I assume there is no way to dedup in `count` since the response  from each shard is just a number

timvisee:
@coszio is right that count should be updated too. I propose to do that in a separate PR though, as it requires a different approach than what I've done in this PR.

PR #4303, opened 2024-05-22
universal-query: user-facing gRPC and REST request types

coszio:
- Add gRPC `QueryPoints` type
  - Move `Fusion` to `points.proto`
- Add rest `QueryRequest` type in collection. Not in `api::rest` because it needs types from `collection` crate.
  - Not added to the schema generator yet because we have not added it to actix (an unused warning appears if we include it at this point).


agourlay:
@coszio please check the openapi consistency check :)

agourlay:
> Add rest QueryRequest type in collection. Not in api::rest because it needs types from collection crate. 

Is this a large number of types?
Can the be copied over into `api::rest` for decoupling?

coszio:
Might be possible, I'll check

lib/api/src/rest/schema.rs
```diff
@@ -132,3 +138,173 @@ pub enum NamedVectorStruct {
...
+    pub query: Option<QueryInterface>,
+
+    /// Define which vector to use for querying. If missing, the default vector is used.
+    pub using: Option<String>,
+
+    /// Filter conditions - return only those points that satisfy the specified conditions.
+    pub filter: Option<Filter>,
+
+    /// Search params for when there is no prefetch
+    pub params: Option<SearchParams>,
```
agourlay, line 182: what are we doing when the top level `params` is defined AND prefetch is defined?

lib/api/src/rest/schema.rs
```diff
@@ -132,3 +138,173 @@ pub enum NamedVectorStruct {
...
+    /// Query to perform. If missing, returns points ordered by their IDs.
+    pub query: Option<QueryInterface>,
+
+    /// Define which vector to use for querying. If missing, the default vector is used.
+    pub using: Option<String>,
+
+    /// Filter conditions - return only those points that satisfy the specified conditions.
+    pub filter: Option<Filter>,
+
+    /// Search params for when there is no prefetch
```
agourlay, line 245: this one is a prefetch :)

lib/api/src/rest/schema.rs
coszio: We ignore it

lib/collection/src/operations/types.rs
```diff
@@ -536,22 +516,13 @@ impl From<u64> for RecommendExample {
...
-///   If the `max_neg_score` is chosen then it is squared and negated, otherwise it is just
-///   the `max_pos_score`.
-#[derive(Debug, Deserialize, Serialize, JsonSchema, Default, PartialEq, Clone, Copy)]
-#[serde(rename_all = "snake_case")]
-pub enum RecommendStrategy {
-    #[default]
-    AverageVector,
-    BestScore,
-}
+// /// How to use positive and negative examples to find the results, default is `average_vector`:
```
agourlay, line 519: I think this one can go now?

coszio: 🧹

agourlay (approved):
Thanks for the effort on the decoupling front :+1:

PR #4307, opened 2024-05-23
Fix overflow panic if scroll limit is usize::MAX

timvisee:
The scroll limit might overflow if providing `usize::MAX` because we add one to it if no ordering is used:

```rust
thread 'tests::snapshot_test::test_shard_dedup' panicked at lib/collection/src/collection/point_ops.rs:260:13:
attempt to add with overflow
```

I've changed this to `saturating_add(1)`. It means we won't add one if the limit is `usize::MAX`, but I think it's unreasonable to reach that limit anyway. Better be safe than panic.

### All Submissions:

### New Feature Submissions:


agourlay (approved):
Ready for very large pages :elephant:

PR #4309, opened 2024-05-23
Add config option shard_nr_per_node

JojiiOfficial:
Adds `shard_number_per_node` option to the config

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -174,6 +174,8 @@ storage:
       # Whether vectors should be stored in memory or on disk.
       on_disk: null
 
+    # shard_number_per_node: 1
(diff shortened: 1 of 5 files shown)


timvisee (approved):

generall (changes requested):
This PR doesn't work as intended.

When I configure a collection on 3-node cluster with `shard_number_per_node: 2`,
I expect to get 6-shards collection by default. But the actual number of shards I get is 3.

Please test solutions before submitting PR;)

timvisee (approved):
Thanks!

Confirmed and tested in a call. :heavy_check_mark:

PR #4312, opened 2024-05-24
Use correct vector storage size

IvanPleshkov:
In a lot of places storage size in bytes is hardoced as `vectors_count * vector_dimension * 4`. But after sparse vectors, multivectors and byte/f16 storages, this behavior is incorrect. This PR removes such vector storage size calculation, and removes `dimension` from segment and vector storage APIs.


CODE_OF_CONDUCT.md
```diff
@@ -5,7 +5,7 @@
 We as members, contributors, and leaders pledge to make participation in our
 community a harassment-free experience for everyone, regardless of age, body
 size, visible or invisible disability, ethnicity, sex characteristics, gender
-identity and expression, level of experience, education, socio-economic status,
+identity and expression, level of experience, education, socioeconomic status,
```
IvanPleshkov, line 8: codespell error

lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -695,6 +695,29 @@ impl SegmentEntry for ProxySegment {
...
+            .get()
+            .read()
+            .available_vectors_size_in_bytes(vector_name)?;
+        let write_count = self.write_segment.get().read().available_point_count();
+        let deleted_points_count = self.deleted_points.read().len();
+        let stored_points = wrapped_count + write_count;
+        let avaliable_points = stored_points.saturating_sub(deleted_points_count);
+        // because we don't know the exact size of deleted vectors, we assume that they are the same avg size as the wrapped ones
+        Ok(
+            ((wrapped_size as u128 + write_size as u128) * avaliable_points as u128
```
IvanPleshkov, line 716: `u128` is required here because of probability to overflow `usize`. Here we estimate size using proportion. We know `available_vectors_size_in_bytes` from write and wrapped segments and we know externally deleted amount. Calculate as `(wrapper_size + write_size) * (total_count - external_deleted_count) / total_count`

lib/segment/src/vector_storage/simple_sparse_vector_storage.rs
```diff
@@ -175,6 +155,11 @@ impl VectorStorage for SimpleSparseVectorStorage {
         self.total_vector_count
     }
 
+    fn available_size_in_bytes(&self) -> usize {
+        // multiply by 2 to account for indices & values
+        self.total_sparse_size * 2
```
IvanPleshkov, line 160: Discussion point. Do we need to multiply by two here? @agourlay
Also, is the total size correct here?

lib/segment/src/vector_storage/tests/async_raw_scorer.rs
```diff
@@ -69,43 +69,42 @@ fn test_async_raw_scorer(
         let mutable_storage = open_simple_dense_vector_storage(
             db,
             rocksdb_wrapper::DB_VECTOR_CF,
-            4,
+            dim,
```
IvanPleshkov, line 72: I don't know how this test worked properly with dimension mismatch =)

lib/segment/tests/integration/byte_storage_hnsw_test.rs
```diff
@@ -82,14 +82,14 @@ fn compare_search_result(result_a: &[Vec<ScoredPointOffset>], result_b: &[Vec<Sc
 }
 
 #[rstest]
-#[case::nearest(QueryVariant::Nearest, VectorStorageDatatype::Uint8, 32, 10)]
-#[case::nearest(QueryVariant::Nearest, VectorStorageDatatype::Float16, 32, 10)]
-#[case::discovery(QueryVariant::Discovery, VectorStorageDatatype::Uint8, 128, 20)]
+#[case::nearest(QueryVariant::Nearest, VectorStorageDatatype::Uint8, 32, 20)]
```
IvanPleshkov, line 85: I decreased required acc here. Don't know actually why. But after updating size estimation in HNSW, this test did low cardinality searches instead of HNSW. I changed the filter generation to restore search over HNSW but accuracy decreases after it

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -94,8 +94,13 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
         } else {
             let vector_storage = vector_storage.borrow();
             let available_vectors = vector_storage.available_vector_count();
-            let full_scan_threshold = hnsw_config.full_scan_threshold.saturating_mul(BYTES_IN_KB)
-                / (vector_storage.vector_dim() * VECTOR_ELEMENT_SIZE);
+            let full_scan_threshold = if available_vectors > 0 {
```
IvanPleshkov, line 97: Discussion point. What to do if storage is empty and is this estimation correct? It affects HNSW search

lib/segment/src/index/plain_payload_index.rs
```diff
@@ -232,18 +232,23 @@ impl PlainIndex {
...
         let vector_storage = self.vector_storage.borrow();
-        let vector_size_bytes = vector_storage.vector_dim() * VECTOR_ELEMENT_SIZE;
-        let indexing_threshold_bytes = search_optimized_threshold_kb * BYTES_IN_KB;
-
-        if let Some(payload_filter) = filter {
-            let payload_index = self.payload_index.borrow();
-            let cardinality = payload_index.estimate_cardinality(payload_filter);
-            let scan_size = vector_size_bytes.saturating_mul(cardinality.max);
-            scan_size <= indexing_threshold_bytes
+        let available_vector_count = vector_storage.available_vector_count();
```
IvanPleshkov, line 235: Here `vector_size_bytes` was changed only. If storage is empty, the function returns `true` - empty storage is small enough for unindexed search

lib/collection/src/collection_manager/holders/proxy_segment.rs
timvisee: How about keeping a u64, and [saturating](https://doc.rust-lang.org/std/primitive.u64.html#method.saturating_mul) up to `u64::MAX`.

Isn't `u64::MAX` is already bigger than the largest amount of data we ever expect?

lib/segment/src/entry/entry_point.rs
```diff
@@ -147,6 +145,21 @@ pub trait SegmentEntry {
...
+    fn max_available_vectors_size_in_bytes(&self) -> OperationResult<usize> {
+        let mut max_size = 0;
+        for vector_name in self.vector_names() {
+            max_size = std::cmp::max(
+                max_size,
+                self.available_vectors_size_in_bytes(&vector_name)?,
+            );
+        }
+        Ok(max_size)
+    }
```
timvisee, line 161: We can simplify this a bit with [`max`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max).

lib/segment/src/fixtures/index_fixtures.rs
```diff
@@ -68,6 +68,10 @@ impl<TMetric: Metric<VectorElementType>> VectorStorage for TestRawScorerProducer
         self.vectors.len()
     }
 
+    fn available_size_in_bytes(&self) -> usize {
+        self.available_vector_count() * self.vector_dim() * std::mem::size_of::<VectorElementType>()
+    }
```
timvisee, line 73: Though I don't see us reaching u64 limits here per se, should we stay on the safe side and use [`saturating_mul`](https://doc.rust-lang.org/std/primitive.usize.html#method.saturating_mul)?

Maybe in the other places we do such calculation too.

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -94,8 +94,13 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
...
+                hnsw_config.full_scan_threshold.saturating_mul(BYTES_IN_KB)
+                    / (vector_storage.available_size_in_bytes()
+                        / vector_storage.available_vector_count())
```
timvisee, line 100:
```suggestion
                        / available_vectors)
```

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -94,8 +94,13 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
         } else {
             let vector_storage = vector_storage.borrow();
             let available_vectors = vector_storage.available_vector_count();
-            let full_scan_threshold = hnsw_config.full_scan_threshold.saturating_mul(BYTES_IN_KB)
-                / (vector_storage.vector_dim() * VECTOR_ELEMENT_SIZE);
+            let full_scan_threshold = if available_vectors > 0 {
+                hnsw_config.full_scan_threshold.saturating_mul(BYTES_IN_KB)
+                    / (vector_storage.available_size_in_bytes()
```
timvisee, line 99: Are we sure this one is never 0?

lib/segment/src/index/plain_payload_index.rs
```diff
@@ -232,18 +232,23 @@ impl PlainIndex {
...
+        if available_vector_count > 0 {
+            let vector_size_bytes =
+                vector_storage.available_size_in_bytes() / vector_storage.available_vector_count();
```
timvisee, line 238:
```suggestion
                vector_storage.available_size_in_bytes() / available_vector_count;
```

lib/segment/src/vector_storage/dense/appendable_mmap_dense_vector_storage.rs
```diff
@@ -137,6 +137,10 @@ impl<T: PrimitiveVectorElement> VectorStorage for AppendableMmapDenseVectorStora
         self.vectors.len()
     }
 
+    fn available_size_in_bytes(&self) -> usize {
+        self.available_vector_count() * self.vector_dim() * std::mem::size_of::<T>()
```
timvisee, line 141: Another `saturating_mul` candidate?

lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
```diff
@@ -162,6 +162,10 @@ impl<T: PrimitiveVectorElement> VectorStorage for MemmapDenseVectorStorage<T> {
         self.mmap_store.as_ref().unwrap().num_vectors
     }
 
+    fn available_size_in_bytes(&self) -> usize {
+        self.available_vector_count() * self.vector_dim() * std::mem::size_of::<T>()
```
timvisee, line 166: Another `saturating_mul` candidate?

lib/segment/src/vector_storage/dense/simple_dense_vector_storage.rs
```diff
@@ -214,6 +214,10 @@ impl<T: PrimitiveVectorElement> VectorStorage for SimpleDenseVectorStorage<T> {
         self.vectors.len()
     }
 
+    fn available_size_in_bytes(&self) -> usize {
+        self.available_vector_count() * self.vector_dim() * std::mem::size_of::<T>()
```
timvisee, line 218: Another `saturating_mul` candidate?

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -181,6 +181,15 @@ impl<T: PrimitiveVectorElement> VectorStorage for AppendableMmapMultiDenseVector
         self.offsets.len()
     }
 
+    fn available_size_in_bytes(&self) -> usize {
+        let total_size = self.vectors.len() * self.vector_dim() * std::mem::size_of::<T>();
```
timvisee, line 185: Another `saturating_mul` candidate?

lib/segment/src/vector_storage/multi_dense/simple_multi_dense_vector_storage.rs
```diff
@@ -308,6 +308,15 @@ impl<T: PrimitiveVectorElement> VectorStorage for SimpleMultiDenseVectorStorage<
         self.vectors_metadata.len()
     }
 
+    fn available_size_in_bytes(&self) -> usize {
+        let total_size = self.vectors.len() * self.vector_dim() * std::mem::size_of::<T>();
```
timvisee, line 312: Another `saturating_mul` candidate?

lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -695,6 +695,33 @@ impl SegmentEntry for ProxySegment {
...
+        let write_count = self.write_segment.get().read().available_point_count();
+        let deleted_points_count = self.deleted_points.read().len();
+        let stored_points = wrapped_count + write_count;
+        let avaliable_points = stored_points.saturating_sub(deleted_points_count);
+        // because we don't know the exact size of deleted vectors, we assume that they are the same avg size as the wrapped ones
+        if stored_points > 0 {
+            Ok(
+                ((wrapped_size as u128 + write_size as u128) * avaliable_points as u128
+                    / stored_points as u128) as usize,
+            )
```
timvisee, line 719: Another `saturating_add` and `saturating_mul` candidate?

lib/segment/src/entry/entry_point.rs
IvanPleshkov: Thanks for fixing it!

lib/segment/src/index/hnsw_index/hnsw.rs
IvanPleshkov: There is check for 0. If 0, `full_scan_threshold` will be 1 as a placeholder value. It's a case of indexed empty segment. It may happen when user has multiple names but didn't set any vector value for some name. Any placeholder value is valid id guess

lib/collection/src/collection_manager/holders/proxy_segment.rs
IvanPleshkov: I propose to use here `u128` because here we multiply size and point amount. For instance, we have segment with 100M OpenAI embeddings. It's `1536dim * 4bytes * 100M count = 614GB = 2^39`. Here we may multiply `2^39` by `100M=2^27` and get some about `2^39 * 2^27=2^66` as an intermediate result before deletion

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -94,8 +94,12 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
         } else {
             let vector_storage = vector_storage.borrow();
             let available_vectors = vector_storage.available_vector_count();
-            let full_scan_threshold = hnsw_config.full_scan_threshold.saturating_mul(BYTES_IN_KB)
-                / (vector_storage.vector_dim() * VECTOR_ELEMENT_SIZE);
+            let full_scan_threshold = if available_vectors > 0 {
+                hnsw_config.full_scan_threshold.saturating_mul(BYTES_IN_KB)
+                    / (vector_storage.available_size_in_bytes() / available_vectors)
```
generall, line 99: In theory, `vector_storage.available_size_in_bytes()` might be zero even if available_vectors is not. E.g. if we have empty sparse vectors or something like this. It might not happen in practice, but I would do the check anyway.

Also, please consider `checked_div`

IvanPleshkov: Good catch!

lib/segment/src/fixtures/index_fixtures.rs
IvanPleshkov: I guess it's safe here because size in bytes on disk cannot be larger than `u64::max`

lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
IvanPleshkov: I guess it's safe here because size in bytes on disk cannot be larger than `u64::max`

lib/segment/src/vector_storage/dense/simple_dense_vector_storage.rs
IvanPleshkov: I guess it's safe here because size in bytes on disk cannot be larger than `u64::max`

lib/segment/src/index/hnsw_index/hnsw.rs
timvisee: In this case, we can probably do:

```rust
let full_scan_threshold = hnsw_config
    .full_scan_threshold
    .saturating_mul(BYTES_IN_KB)
    .checked_div(
        vector_storage
            .available_size_in_bytes()
            .checked_div(available_vectors)
            .unwrap_or_default(),
    )
    .unwrap_or(1);
```

or the other way around:

```rust
let full_scan_threshold = vector_storage
    .available_size_in_bytes()
    .checked_div(available_vectors)
    .and_then(|div| {
        hnsw_config
            .full_scan_threshold
            .saturating_mul(BYTES_IN_KB)
            .checked_div(div)
    })
    .unwrap_or(1);
```

lib/segment/src/vector_storage/dense/simple_dense_vector_storage.rs
timvisee: Yes, sounds reasonable :+1:

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -181,6 +181,15 @@ impl<T: PrimitiveVectorElement> VectorStorage for AppendableMmapMultiDenseVector
...
+        if self.total_vector_count() > 0 {
+            total_size * self.available_vector_count() / self.total_vector_count()
```
generall, line 187: why don't we just do `self.available_vector_count() * self.vector_dim() * std::mem::size_of::<T>()` ?

lib/segment/src/index/hnsw_index/hnsw.rs
IvanPleshkov: done, renamed `div` in snippet into `avg_vector_size`

timvisee: Better. Thanks!

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
IvanPleshkov: We cannot do it for multivectors. Your proposal will work only if all multivectors have 1 vector inside. We estimate here available vectors size as a proportion. We guess here that average vector size between existing and deleted vectors are the same

lib/collection/src/collection_manager/holders/proxy_segment.rs
IvanPleshkov: answered here: https://github.com/qdrant/qdrant/pull/4312#discussion_r1616029582
It's possible to get `u64` overflow and this overflow is a correct case, we need `u128` here

lib/segment/tests/integration/byte_storage_hnsw_test.rs
```diff
@@ -233,8 +233,8 @@ fn test_byte_storage_hnsw(
     for i in 0..attempts {
         let query = random_query(&query_variant, &mut rnd, dim);
 
-        let range_size = 40;
-        let left_range = rnd.gen_range(0..400);
+        let range_size = 160;
```
generall, line 236: I believe you changed that cause otherwise the quality were too bad, rigth?

IvanPleshkov: It's a `u8`-vectors test. This PR affects the full scan threshold in HNSW for `u8` and `f16` vectors. I changed the filter here because the old range has a small cardinality in terms of affected full scan

lib/segment/tests/integration/byte_storage_hnsw_test.rs
generall: I propose to change `full_scan_threshold` to 0, so it would be fair for the HNSW

lib/segment/tests/integration/byte_storage_hnsw_test.rs
IvanPleshkov: Reverted this change and set `full_scan_threshold = 0` instead

lib/segment/tests/integration/byte_storage_hnsw_test.rs
IvanPleshkov: Reverted this change and set `full_scan_threshold = 0` instead

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
IvanPleshkov: here we cannot overflow usize because it's a size in bytes of storage. But we can overflow it next line, I added `u128` usage. The reason is the same as here https://github.com/qdrant/qdrant/pull/4312#discussion_r1616029582

lib/segment/src/vector_storage/multi_dense/simple_multi_dense_vector_storage.rs
IvanPleshkov: here we cannot overflow usize because it's a size in bytes of storage. But we can overflow it next line, I added `u128` usage. The reason is the same as here https://github.com/qdrant/qdrant/pull/4312#discussion_r1616029582

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
IvanPleshkov: Your true, fixed it after already for all multivector storages
https://github.com/qdrant/qdrant/pull/4312/commits/f8400e223b2a0efd0d9744fa91e9f3a8d149e5cb

PR #4313, opened 2024-05-24
Recreate sparse vector index

xzfc:
Instead of implementing index migration, we could just drop and re-create the index from scratch when it is not compatible or fails to load. This PR implements this approach.

Pros of this approach over hand-written migration:
- Simpler and more robust.
- We can implement arbitrary changes in the index format as many times as we want without fear of losing/corrupting the data.
- It allows rolling back to the previous index version for free.

Cons:
- It takes 25 seconds of CPU time on my machine to rebuild an index in a single segment with 1.3M of points.
- Increased memory consumption during the first startup. Might significantly increase loading time if the system is swapping.

### DIFF
--- a/lib/segment/src/index/sparse_index/sparse_vector_index.rs
+++ b/lib/segment/src/index/sparse_index/sparse_vector_index.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use std::fs::{create_dir_all, remove_dir_all};
+use std::fs::{create_dir_all, remove_dir_all, rename};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
@@ -9,11 +9,12 @@ use common::cpu::CpuPermit;
 use common::types::{PointOffsetType, ScoredPointOffset, TelemetryDetail};
 use io::storage_version::{StorageVersion as _, VERSION_FILE};
 use itertools::Itertools;
+use semver::Version;
 use sparse::common::scores_memory_pool::ScoresMemoryPool;
 use sparse::common::sparse_vector::SparseVector;
 use sparse::common::types::DimId;
 use sparse::index::inverted_index::inverted_index_ram_builder::InvertedIndexBuilder;
-use sparse::index::inverted_index::InvertedIndex;
+use sparse::index::inverted_index::{InvertedIndex, INDEX_FILE_NAME, OLD_INDEX_FILE_NAME};
 use sparse::index::migrate::SparseVectorIndexVersion;
 use sparse::index::search_context::SearchContext;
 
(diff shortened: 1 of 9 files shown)


lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -101,19 +102,35 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
+                    || (),
+                )?;
+
+                remove_dir_all(path)?;
+                create_dir_all(path)?;
+
+                SparseVectorIndexVersion::save(path)?;
+                config.save(&config_path)?;
+                inverted_index.save(path)?;
+                indices_tracker.save(path)?;
```
generall, line 126: I am not sure if this order of saving will reliable once we assume that service can be killed at any moment.

Assume the following scenario:

- we migrate from the current version
- we do the save 

```
SparseVectorIndexVersion::save(path)?;
config.save(&config_path)?;
inverted_index.save(path)?;
```

but the service got killed with SIGKILL

during the

```
indices_tracker.save(path)?;
```

In this case next time we will load the index, it will already see the new version and won't try to apply migration, even though the `indices_tracker` was only partially saved.

It might be, that partially saved index will cause the error and re-building of the index on it's own, but there are no explicit guarantees about that.

Our suggestion would be to change the order of saving and save version file always as a last step. In this way we can guarantee that the migration is completed if we see the newest version in the storage.

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -132,6 +149,24 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         })
     }
 
+    fn try_load(
+        path: &Path,
+    ) -> OperationResult<(SparseIndexConfig, TInvertedIndex, IndicesTracker)> {
+        let stored_version = SparseVectorIndexVersion::load(path)?;
```
generall, line 155: WDYT if we add a "default_version" here, to prevent migration before we actually enable the new type of posting lists?

lib/segment/src/index/sparse_index/sparse_vector_index.rs
xzfc: > It might be, that partially saved index will cause the error and re-building of the index on it's own, but there are no explicit guarantees about that.

It will fail during `try_load()` which requires all of the files to be present.
Note: previously, the missing indices tracker file was optional, falling back to trivial mapping (`indices_tracker.remap_index(i) == Some(i)`). In this PR this file is required to be present.

I think "all files are required" is easier to reason than considering the order of saves.

Perhaps, except, for the version file, to address your review comment below. It still should not be the last file to save. Otherwise, in following scenario the resulting files would be implicitly incorrectly treated as the old format due to the missing version file.
```rs
config.save(); inverted_index.save(); indices_tracker.save();
SIGKILL
version.save(); // version not saved
```

I'll state this approach explicitly in the code comments.

generall: > It will fail during try_load() which requires all of the files to be present.

This is only true if the `indices_tracker.save(path)` is atomic. But in current implementation it is only atomic _by chance_. 
Nobody would stop us if we would want to implement it in a way like `inverted_index.save` is implemented, and this issue will be very hard to find. 


> Otherwise, in following scenario the resulting files would be implicitly incorrectly treated as the old format due to the missing version file.

There are 2 options here:

1.  use different names for new version files (which also will make it easier to rollback to previous version in case of failure and don't delete old index before save)

2. Delete version file before saving index - easy, but not as elegant

generall (approved):

PR #4317, opened 2024-05-24
Add optimizer_overwrite config option

JojiiOfficial:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


timvisee (commented):
Could you add a - possibly commented out - section in `config/config.yaml` to show how this would be configured in practice?

config/config.yaml
```diff
@@ -129,6 +129,10 @@ storage:
     # If 0 - no optimization threads, optimizations will be disabled.
     max_optimization_threads: null
 
+  # This section has the same options as 'optimizers' above. All values specified here will overwrite the collections
+  # optimizers configs regardless of the config above and the options specified at collection creation.
+  optimizers_overwrite: null
```
timvisee, line 134:
```suggestion
  optimizers_overwrite: {}
```

Can we do this instead as empty object, with no changed parameters?

JojiiOfficial: Just curious but what exactly is different compared to `null`?

timvisee: It's subtle.

To me `null` looks like you'd specify a single value in that field. While `{}` hints it being an object in which you can specify multiple parameters.

timvisee: Maybe it would be even better to do it like this:

```suggestion
  # optimizers_overwrite:
  #   indexing_threshold_kb: 20000
  #   max_optimization_threads: null
```

What do you think?

JojiiOfficial: Giving it a second thought, it's indeed easier to set parameters when you just have to uncomment them, so I think it's a good idea!

timvisee (approved):

lib/storage/src/content_manager/toc/create_collection.rs
```diff
@@ -156,11 +156,16 @@ impl TableOfContent {
...
-        let optimizers_config = match optimizers_config_diff {
+        let mut optimizers_config = match optimizers_config_diff {
             None => self.storage_config.optimizers.clone(),
             Some(diff) => diff.update(&self.storage_config.optimizers)?,
         };
 
+        // Apply global optimizers-config overwriting if configured
+        if let Some(optimizers_overwrite) = &self.storage_config.optimizers_overwrite {
+            optimizers_config = optimizers_overwrite.clone().update(&optimizers_config)?;
+        }
```
generall, line 167: This change will create collection with "optimizers_overwrite" parameters, which will be just persisted everywhere.

But what we need is different: we need to use `optimizers_overwrite` _even though_ the collection is configured with different values.
That means, for example, that we should be able to restart service with differnt config, and the already existing collection should act differently, without changes of the persisted collection config

generall (changes requested):

lib/collection/src/shards/shard_holder.rs
```diff
@@ -585,6 +586,7 @@ impl ShardHolder {
         collection_path: &Path,
         collection_id: &CollectionId,
         collection_config: Arc<RwLock<CollectionConfig>>,
+        optimizers_config: OptimizersConfig,
```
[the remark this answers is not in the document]
JojiiOfficial: Yeah, indeed a good idea!

generall (approved):

PR #4323, opened 2024-05-24
Reset update buffers for simple storages

agourlay:
This PR makes sure we are not re-using the data of the previously inserted vector when inserting a new deleted vector in a rocksdb simple storage.


agourlay:
Looks like I need to this a bit :)

timvisee:
I agree with @generall, this should not be necessary. When implementing this approach our logic was: performance is more important here than cleanliness.

agourlay:
I will close this PR and focus later on removing the performance optimization to improve readability :+1:

PR #4326, opened 2024-05-26
Handle SIGABRT crash on search with huge limit

### LINKED ISSUE #4321: SIGABRT crash on search with huge limit
## Current Behavior
When searching with a huge limit such as 99999999999999999, Qdrant crashes with a SIGABRT.

Such limit is obviously unrealistic, but that doesn't mean we should accept the hard crash.

## Steps to Reproduce
1. Create collection with 100k points: `bfb -d 2`
2. Do search request with a huge limit:
   ```json
   POST collections/benchmark/points/search 
   {
     "limit": 99999999999999999,
     "vector": [0.1, 0.2]
   }
   ```
3. Watch Qdrant crash:
   ```bash
   memory allocation of memory allocation of 800000000000000008800000000000000008 bytes failed
    bytes failed
   fish: Job 1, './target/perf/qdrant' terminated by signal SIGABRT (Abort)
   ```

## Expected Behavior
No matter he input, we prevent crashing in all cases. When using this on a cluster, you might take down 100% of the nodes instantly with one request.

Ideally we'd simply respond with the complete set of points as search result.

If that isn't possible, we should respond with an error explaining the specified limit is problematic.

## Possible Solution
Here are some potential solutions, we probably want to pick one:

1. Memory allocation fails. We either want to test for memory availability or catch the allocation failure somehow so we can return an error to the user.

2. We could internally clamp the limit to the total number of points available in a collection. You cannot return more results than points you have. The total point count is somewhat unreliable though, which makes this hard to implement correctly.

3. We could implement a different code path for when a huge limit is set, this code path would sort search results in a different way not requiring us to allocate capacity for _limit_ items.

4. We can also set an artificial upper bound for the limit, but I'd rather see any of the above solutions instead.

## Detailed Description
As far as I understand, memory allocation fails in the [fixed length priority queue](https://github.com/qdrant/qdrant/blob/76c77092b6645605cd72de31d74bc76e9ab95f82/lib/common/common/src/fixed_length_priority_queue.rs) we use during searches. It tries to allocate capacity for 99999999999999999 items, for which we do not have enough memory.

We had something else before, maybe it is better suited for huge limits not requiring such huge allocations.

kid-116:
/claim #4321

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -811,10 +811,15 @@ impl LocalShard {
         filter: Option<&'a Filter>,
     ) -> CollectionResult<BTreeSet<PointIdType>> {
         let segments = self.segments().read();
-        let all_points: BTreeSet<_> = segments
-            .non_appendable_then_appendable_segments()
-            .flat_map(|segment| segment.get().read().read_filtered(None, None, filter))
-            .collect();
+        // let all_points: BTreeSet<_> = segments
+        //     .non_appendable_then_appendable_segments()
+        //     .flat_map(|segment| segment.get().read().read_filtered(None, None, filter))
+        //     .collect();
+        let mut all_points = BTreeSet::new();
+        for segment in segments.non_appendable_then_appendable_segments() {
+            let points = segment.get().read().read_filtered(None, None, filter)?;
+            all_points.extend(points);
+        }
         Ok(all_points)
     }
 
--- a/lib/collection/src/shards/local_shard/scroll.rs
+++ b/lib/collection/src/shards/local_shard/scroll.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use futures::future::try_join_all;
 use itertools::Itertools as _;
+use segment::common::operation_error::OperationError;
 use segment::data_types::order_by::{Direction, OrderBy, OrderValue};
 use segment::types::{
     ExtendedPointId, Filter, ScoredPoint, WithPayload, WithPayloadInterface, WithVector,
@@ -131,8 +132,19 @@ impl LocalShard {
             })
         };
 
+        let unwrap_vec = |vec: Vec<Result<Vec<ExtendedPointId>, OperationError>>| -> Result<Vec<Vec<ExtendedPointId>>, OperationError> {
+            for ele in vec.iter() {
+                if let Err(err) = ele {
+                    return Err(err.clone());
+                }
+            }
+            Ok(vec.into_iter().map(|x| x.unwrap()).collect())
+        };
+
         let non_appendable = try_join_all(non_appendable.into_iter().map(read_filtered)).await?;
+        let non_appendable = unwrap_vec(non_appendable)?;
         let appendable = try_join_all(appendable.into_iter().map(read_filtered)).await?;
+        let appendable = unwrap_vec(appendable)?;
 
         let point_ids = non_appendable
             .into_iter()
--- a/lib/common/common/src/fixed_length_priority_queue.rs
+++ b/lib/common/common/src/fixed_length_priority_queue.rs
@@ -24,7 +24,10 @@ impl<T: Ord> FixedLengthPriorityQueue<T> {
     /// Creates a new queue with the given length
     /// Panics if length is 0
     pub fn new(length: usize) -> Self {
-        let heap = BinaryHeap::with_capacity(length + 1);
+        let mut heap = BinaryHeap::new();
+        heap.try_reserve_exact(length + 1)
+            .expect("Failed to reserve memory for heap");
+        let heap = heap;
         let length = NonZeroUsize::new(length).expect("length must be > 0");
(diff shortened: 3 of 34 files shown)


kid-116:
@generall The current fix, prevents a complete crash and the following response is received:

```json
{
    "status": {
        "error": "Service internal error: 1 of 1 read operations failed:\n  Service internal error: task 165 panicked"
    },
    "time": 1.055431089
}
```

What should be my next steps? I believe if we wish to display a better message, we can move start by implementing a `pub enum CommonError::AllocationFailure` and propagate the error through the layers. Otherwise, I could start working on adding tests.

lib/common/common/src/fixed_length_priority_queue.rs
```diff
@@ -24,7 +24,10 @@ impl<T: Ord> FixedLengthPriorityQueue<T> {
     /// Creates a new queue with the given length
     /// Panics if length is 0
     pub fn new(length: usize) -> Self {
-        let heap = BinaryHeap::with_capacity(length + 1);
+        let mut heap = BinaryHeap::new();
+        heap.try_reserve_exact(length + 1)
+            .expect("Failed to reserve memory for heap");
```
generall, line 29: This can still crash the whole service under some conditions. 
While following this strategy, the proper solution would require to propagate error to the API via result

kid-116: Right

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -131,8 +132,19 @@ impl LocalShard {
...
+                    return Err(err.clone());
+                }
+            }
+            Ok(vec.into_iter().map(|x| x.unwrap()).collect())
+        };
+
         let non_appendable = try_join_all(non_appendable.into_iter().map(read_filtered)).await?;
+        let non_appendable = unwrap_vec(non_appendable)?;
         let appendable = try_join_all(appendable.into_iter().map(read_filtered)).await?;
+        let appendable = unwrap_vec(appendable)?;
```
timvisee, line 147: We can probably simplify this by collecting like this: `collect::<Result<Vec<Vec<ExtendedPointId>>>()`

It takes out the results short cutting early for us.

lib/common/common/src/fixed_length_priority_queue.rs
```diff
@@ -16,17 +16,19 @@ pub struct FixedLengthPriorityQueue<T: Ord> {
...
 
 impl<T: Ord> FixedLengthPriorityQueue<T> {
     /// Creates a new queue with the given length
     /// Panics if length is 0
-    pub fn new(length: usize) -> Self {
-        let heap = BinaryHeap::with_capacity(length + 1);
+    pub fn new(length: usize) -> Result<Self, TryReserveError> {
+        let mut heap = BinaryHeap::new();
+        heap.try_reserve_exact(length + 1)?;
+        let heap = heap;
```
timvisee, line 29:
```suggestion
```

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -811,10 +811,15 @@ impl LocalShard {
...
     ) -> CollectionResult<BTreeSet<PointIdType>> {
         let segments = self.segments().read();
-        let all_points: BTreeSet<_> = segments
-            .non_appendable_then_appendable_segments()
-            .flat_map(|segment| segment.get().read().read_filtered(None, None, filter))
-            .collect();
+        // let all_points: BTreeSet<_> = segments
+        //     .non_appendable_then_appendable_segments()
+        //     .flat_map(|segment| segment.get().read().read_filtered(None, None, filter))
+        //     .collect();
```
timvisee, line 817:
```suggestion
```

generall:
Closing in favor of https://github.com/qdrant/qdrant/pull/4328

timvisee (changes requested):
Thank you for picking this up!

A different approach was also attempted in <https://github.com/qdrant/qdrant/pull/4328>, since it was so simple I merged it right away.

However, I do like that you're properly propagating errors. I therefore propose to merge both, including your PR. That does require a few changes though.

I like the idea of preallocating at most 1M capacity, as implemented <https://github.com/qdrant/qdrant/pull/4328>. I think we should keep that but catch allocation errors with your logic. Along with that we then also need to catch reallocation errors when the structure grows. I'd suggest to also use `try_reserve(1)` catching errors each time we push new elements into the binary heap.

Could you do the following?
- Resolve the current open review comments
- Rebase on the latest dev
- Pre-allocate at most 1M, catching allocation errors
- When inserting new elements into the binary heap, use `try_reserve(1)` first to catch allocation errors along the way

timvisee:
On second thought, we will be closing it.

Though the above changes might sound good, in such scenario it is very likely to crash in a different place first. I didn't think about that. That would eliminate the effect of these changes. Resolving that will require humongous refactoring which is out of scope for this issue/PR.

Anyway, thank you for giving this a shot!

kid-116:
Cool

PR #2136, opened 2023-06-22
delete points from id tracker on flush only

generall:
fix flush ordering of the removed points during copy-on-write

PR #4329, opened 2024-05-27
Rocksdb atomic flush

IvanPleshkov:
RocksDB documentation about atomic flush:
https://github.com/facebook/rocksdb/wiki/Atomic-flush

Possible solution of https://github.com/qdrant/qdrant/issues/4318

RocksDB has an internal background flush which is a black box. Maybe bug was appeared because background flush did IDTracker flush first (because of small size of changes or something else) and background flush of Storage didn't finish because of killing machine.

If so, rocksdb has an option `atomic_flush` wich guarantee consistency between column families in background flush.


IvanPleshkov:
Close because it seems https://github.com/qdrant/qdrant/pull/2136 covers described behaviour

PR #4331, opened 2024-05-27
universal-query: Impl of `query_internal` in collection

coszio:
Tracked in #4225 

- Moves `ScoredPointTies` to segment, as it is now used for `cmp` when merging from shards
- Implements the internal version of `query` which merges intermediate results from shards. This function will only be called from the internal service (when the instance is a remote)


lib/segment/tests/integration/utils/mod.rs
```diff
@@ -1 +1 @@
-pub mod scored_point_ties;
```
agourlay, line 1: empty file ? :)

lib/collection/src/collection/query.rs
```diff
@@ -0,0 +1,122 @@
...
+            let order = collection_params.scoring_order(intermediate_info.scoring_query)?;
+
+            let intermediate_result = match order {
+                Order::LargeBetter => Either::Left(
+                    intermediate_result_for_shard
+                        .kmerge_by(|a, b| ScoredPointTies(a) > ScoredPointTies(b)),
+                ),
+                Order::SmallBetter => Either::Right(
+                    intermediate_result_for_shard
+                        .kmerge_by(|a, b| ScoredPointTies(a) < ScoredPointTies(b)),
```
coszio, line 110: To discuss: we need this to merge points which have equal scoring, like those coming from scrolls. Should we just add this comparison by ID to the `Ord` impl of `ScoredPoint`?

lib/segment/src/utils/scored_point_ties.rs
```diff
@@ -0,0 +1,39 @@
...
+        let res = self.0.cmp(other.0);
+
+        // for identical scores, we fallback to sorting by ids to have a stable output
+        if res == Ordering::Equal {
+            self.0.id.cmp(&other.0.id)
+        } else {
+            res
+        }
+    }
+}
```
timvisee, line 25: We can do [`then_with`](https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.then_with):

(untested):
```suggestion
    fn cmp(&self, other: &Self) -> Ordering {
        self.0.cmp(other.0).then_with(|| self.0.id.cmp(&other.0.id))
    }
}
```

lib/collection/src/collection/query.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+impl Collection {
+    async fn query_shards_concurrently(
+        &self,
+        request: Arc<ShardQueryRequest>,
+        read_consistency: Option<ReadConsistency>,
+        shard_selection: &ShardSelectorInternal,
+    ) -> CollectionResult<Vec<ShardQueryResponse>> {
+        // query all shards concurrently
```
timvisee, line 29: Do we want _some_ limiting here in case there is a huge number of shards?

lib/collection/src/collection/query.rs
timvisee: Not sure I'd like that. Comparing in that way might not always make sense when you have a scored point, or does it?

The construction of `ScoredPointTies(a)` is cheap, so I wouldn't worry about it from a performance perspective.

Do you have a specific reason on why you might prefer implementing Ord on the scored point directly for this?

timvisee (approved):

lib/collection/src/collection/query.rs
coszio: > The construction of ScoredPointTies(a) is cheap, so I wouldn't worry about it from a performance perspective.

My same take on it.

I was only suggesting if it should become a part of the behavior of ScoredPoint, but we might not want that.

lib/collection/src/config.rs
```diff
@@ -395,4 +396,28 @@ impl CollectionParams {
             Ok(Default::default())
         }
     }
+
+    pub fn scoring_order(&self, scoring_query: Option<&ScoringQuery>) -> CollectionResult<Order> {
```
generall, line 400: not sure if this function should be part of the collection configuration. I would rather prefer it independent

lib/collection/src/operations/universal_query/shard_query.rs
```diff
@@ -56,6 +58,38 @@ impl ScoringQuery {
             _ => None,
         }
     }
+
+    /// Return the order of results, depending on the type of query
+    pub fn order(
```
generall, line 63: is this logic duplicated?

lib/collection/src/collection/query.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        let mut all_shards_results = self
+            .query_shards_concurrently(Arc::clone(&request), read_consistency, shard_selection)
+            .await?;
+
+        let (num_intermediate_responses, intermediate_infos) =
+            if let Some(ScoringQuery::Fusion(Fusion::Rrf)) = request.query {
+                (
+                    request.prefetches.len(),
+                    Either::Left(
```
generall, line 72: what's the difference between left and right in this context?

lib/collection/src/collection/query.rs
```diff
@@ -0,0 +1,122 @@
...
+                        }
+                    }
+                    Ok(records)
+                })
+        });
+        future::try_join_all(all_searches).await
+    }
+
+    /// This is a remote shard implementation to querying the shards. Only used for the internal service
+    pub async fn query_internal(
```
generall, line 56: tbh, I am struggling to understand what's going on in this function. it might require some more comments/decomposition

lib/collection/src/collection/query.rs
coszio: It is only to avoid collecting, but it should be cheap to collect as well

lib/collection/src/collection/query.rs
coszio: This is what will be called on the remote shard. So that, if it is a fusion request, it sends the merged prefetches instead of the main query. 

I'll make it easier to look at

lib/collection/src/operations/universal_query/shard_query.rs
coszio: It was! I will remove the one in CollectionConfig

lib/collection/src/collection/query.rs
```diff
@@ -65,41 +65,66 @@ impl Collection {
...
+            // Either::Left and Either::Right are used to allow type inference to work.
+            //
             let intermediate_result = match order {
                 Order::LargeBetter => Either::Left(
-                    same_result_per_shard.kmerge_by(|a, b| ScoredPointTies(a) > ScoredPointTies(b)),
+                    shards_results
+                        .into_iter()
+                        .kmerge_by(|a, b| ScoredPointTies(a) > ScoredPointTies(b)),
                 ),
                 Order::SmallBetter => Either::Right(
```
generall, line 120: I was trying to get rid of those Either things, but it looks like type checker doesn't really like it, alternative implementations seems to have performance overhead.

So instead I added comment of how ideally I would like this code to look like, if it were able to compile like that

generall (approved):

PR #4332, opened 2024-05-27
Fix missing segments, use correct path for new segment created during snapshot

timvisee:
Fix issue where segments did go missing.

We created a new segments in the wrong directory, which means they would not be loaded on restart.

This also refactors some parameters, function names and documentation to remove a lot of confusion around what type of paths are used where.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -658,7 +658,7 @@ impl<'s> SegmentHolder {
     /// sourced from it.
     pub fn proxy_all_segments_and_apply<F>(
         segments: LockedSegmentHolder,
-        shard_path: &Path,
+        segments_path: &Path,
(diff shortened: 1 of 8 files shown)


lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -753,7 +753,7 @@ impl<'s> SegmentHolder {
         };
 
         // Create temporary appendable segment to direct all proxy writes into
-        let tmp_segment = LockedSegment::new(build_segment(collection_path, &config, false)?);
+        let tmp_segment = LockedSegment::new(build_segment(segments_path, &config, false)?);
```
timvisee, line 756: Most importantly, we built new segments in the collection path before, instead of the segments path.

ffuugoo (commented):
Great job figuring this out! 👍

generall (approved):

PR #4338, opened 2024-05-27
universal-query: Fix order in `LocalShard` vector rescoring

coszio:
Tracked in #4225 

- create a single core search request for all ids to rescore. Correct ordering is already handled within `do_search`
~- optimize taking from sources so that no extra cloning is made~

### DIFF
--- a/lib/collection/src/shards/local_shard/query.rs
+++ b/lib/collection/src/shards/local_shard/query.rs
@@ -148,12 +148,11 @@ impl LocalShard {
             }
             ScoringQuery::Vector(query_enum) => {
                 // create single search request for rescoring query
-                let point_ids = sources.into_iter().fold(HashSet::new(), |mut acc, source| {
-                    for scored_point in source {
-                        acc.insert(scored_point.id);
-                    }
-                    acc
-                });
+                let point_ids = sources
+                    .into_iter()
+                    .flatten()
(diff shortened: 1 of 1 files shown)


generall (approved):

lib/collection/src/shards/local_shard/query.rs
```diff
@@ -147,45 +147,44 @@ impl LocalShard {
...
-                let top = core_results
-                    .into_iter()
-                    .flatten()
-                    .sorted_unstable()
-                    .take(limit)
-                    .collect();
+                    .await?
+                    // One search request is sent. We expect only one result
+                    .pop()
+                    .unwrap_or_default();
```
timvisee, line 186: I wonder if panicing with a descriptive error may be better here :thinking:

timvisee (approved):

lib/collection/src/shards/local_shard/query.rs
coszio: Maybe not panic but logging?

timvisee: Better.

The main point was, just taking the default here if something goes wrong may create unexpected problems down the line that are hard to debug.

coszio: It now returns a service error

PR #4347, opened 2024-05-28
universal-query: Clone less in local shard

coszio:
- Introduce a custom iterator abstraction for going over the results of core searches, scrolls, and new merges.
- Reduce cloning by using `Cow`, since cloning is only necessary for RRF.

### DIFF
--- a/lib/collection/src/shards/local_shard/query.rs
+++ b/lib/collection/src/shards/local_shard/query.rs
@@ -201,11 +201,14 @@ impl LocalShard {
                 Ok(top_rrf)
             }
             ScoringQuery::OrderBy(order_by) => {
-                let point_ids = sources
-                    .into_iter()
-                    .flatten()
-                    .map(|scored_point| scored_point.id)
-                    .collect::<HashSet<_>>();
+                // create single scroll request for rescoring query
+                let mut point_ids = HashSet::new();
+
+                for source in sources {
+                    for point in source.iter() {
+                        point_ids.insert(point.id);
+                    }
+                }
 
                 let filter = Filter::new_must(segment::types::Condition::HasId(
                     HasIdCondition::from(point_ids),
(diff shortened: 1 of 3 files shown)


lib/collection/src/shards/local_shard/query.rs
```diff
@@ -92,41 +147,38 @@ impl LocalShard {
...
                     }
-                };
-                sources.extend(vec);
+                }
             }
 
-            if depth == 0 && prefetch.merge.rescore == Some(ScoringQuery::Fusion(Fusion::Rrf)) {
+            let sources = prefetch_holder.iter_sources(search_indices, scroll_indices, merged_list);
+
+            if depth == 0 && merge_plan.merge.rescore == Some(ScoringQuery::Fusion(Fusion::Rrf)) {
```
generall, line 176:
```suggestion
            if depth == 0 && merge_plan.merge.rescore == Some(ScoringQuery::Fusion(_)) {
```

generall (approved):

lib/collection/src/shards/local_shard/query.rs
```diff
@@ -92,41 +147,38 @@ impl LocalShard {
...
-                    }
-                    PrefetchSource::ScrollsIdx(idx) => {
-                        // TODO(universal-query): don't clone, by using something like a hashmap instead of a vec
-                        let scrolled = scrolls.get(idx).cloned().unwrap_or_default();
-                        vec![scrolled]
-                    }
+            let mut search_indices = Vec::new();
+            let mut scroll_indices = Vec::new();
+
+            let mut merged_list = Vec::new();
```
timvisee, line 153: Not strictly necessary, but can we make some sane guesstimations here for preallocating capacity?

timvisee (approved):

lib/collection/src/shards/local_shard/query.rs
coszio: Maybe better to turn into an exhaustive match, or does every fusion need all intermediate results?

lib/collection/src/shards/local_shard/query.rs
coszio: We might preallocate all of them to the length of `merge_plan.sources`.

PR #4356, opened 2024-05-30
Add IdTrackerEnum

JojiiOfficial:
Introduceds enum dispatch for IdTracker. Currently falls back to `SimpleIdTracker` for ImmutableIdTracker as this will follow in a separater PR.

### DIFF
--- a/lib/segment/src/id_tracker/id_tracker_base.rs
+++ b/lib/segment/src/id_tracker/id_tracker_base.rs
@@ -147,15 +147,17 @@ pub trait IdTracker {
 pub type IdTrackerSS = dyn IdTracker + Sync + Send;
 
 pub enum IdTrackerEnum {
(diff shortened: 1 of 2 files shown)


lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -144,3 +145,146 @@ pub trait IdTracker {
 }
 
 pub type IdTrackerSS = dyn IdTracker + Sync + Send;
+
+pub enum IdTrackerEnum {
+    SimpleIdTracker(SimpleIdTracker),
+    ImmutableIdTracker(SimpleIdTracker),
```
timvisee, line 151: We might want to call the variants `Mutable` and `Immutable` for the time being.

timvisee (approved):

generall (approved):

PR #4359, opened 2024-05-30
Add optional post WAL data consistency check

agourlay:
This PR adds a new post WAL application data consistency check.

The core assumption is that there must be no data consistency issue **after** the outstanding operations from the WAL have been applied.

The data consistency check looks for:
- dangling internal ids
- dangling external ids
- internal ids without version
- internal ids without stored vector

The proposal is to perform this new check behind a new feature flag in order for us to test internally our storage without impacting users.

I have added the feature flag in our CI pipelines where I believe it makes sense.

When an inconsistency is detected, the impacted shard cannot be loaded properly.

The server is still up but without the shards with corrupted segments.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ tracing-tracy = ["tracing", "dep:tracing-tracy"]
 tokio-tracing = ["tokio/tracing"]
 stacktrace = ["rstack-self"]
 chaos-testing = []
+data-consistency-check = ["collection/data-consistency-check"]
 
 [dev-dependencies]
 serde_urlencoded = "0.7"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -33,6 +33,7 @@ tracing-tracy = ["tracing", "dep:tracing-tracy"]
 tokio-tracing = ["tokio/tracing"]
 stacktrace = ["rstack-self"]
 chaos-testing = []
+data-consistency-check = ["collection/data-consistency-check"]
```
timvisee, line 36: This is very much a nitpick, but I can imagine us potentially adding more of these in the future.

Shall we prefix them all with `debug-` to organize it a bit?

agourlay: This sounds sensible.
Do you want me to rename everything in this PR or later?

generall (approved):

generall:
I am going to merge now, so we won't wait for CI tomorrow

PR #4360, opened 2024-05-30
universal-query: convert `CollectionQueryRequest` into `ShardQueryRequest` (resolve ids into vectors)

coszio:
Needs #4331 

- make `VectorQuery` generic
- retrieve ids from `CollectionQueryRequest` and convert into vectors
- implement `RetrieveRequest` trait for it
- calculate avg vector with logic from `recommendations.rs`
- convert `CollectionQueryRequest` into `ShardQueryRequest`
- begin implementing `query()` fn in collection

### DIFF
--- a/lib/collection/src/common/fetch_vectors.rs
+++ b/lib/collection/src/common/fetch_vectors.rs
@@ -15,6 +15,7 @@ use crate::operations::shard_selector_internal::ShardSelectorInternal;
 use crate::operations::types::{
     CollectionError, CollectionResult, PointRequestInternal, RecommendExample, Record,
 };
+use crate::operations::universal_query::collection_query::VectorInput;
 
 pub async fn retrieve_points(
     collection: &Collection,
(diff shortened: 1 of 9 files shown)


timvisee:
@coszio I see this is drafted. Would you still like us to review it?

coszio:
@timvisee Yes, it was only drafted because it depended on a previous PR. Marking ready for review now

lib/collection/src/common/fetch_vectors.rs
```diff
@@ -105,17 +106,32 @@ impl ReferencedVectors {
...
+        match lookup_collection_name {
             None => self.default_mapping.get(&point_id),
             Some(collection) => {
                 let collection_mapping = self.collection_mapping.get(*collection)?;
                 collection_mapping.get(&point_id)
             }
         }
     }
+
+    pub fn convert_to_vectors_owned<'a>(
```
generall, line 121: It might be more comprehensive, if we just conver single vector at a time, and do the iteration outside of this function

generall (approved):

PR #4364, opened 2024-05-30
Make sparse vectors generic

xzfc:
This PR makes sparse vector types ant traits generic over the weight type.

Two implementations are provided: `half::f16` and `f32`. We'll add `u8` later since its logic significantly differs. The `half::f16` is not used anywhere yet and serves as a reference.

# Benchmark results

### basic

task | ram | ramc32 | ramc16 | diff | mmap | mmapc32 | mmapc16 | diff
| - | - | - | - | - | - | - | - | -
random-50k | 0.75228 | 0.98497 | 1.1265 | +30% / +49% | 0.75071 | 0.9944 | 1.1692 | +32% / +55%
random-500k | 6.8732 | 8.7098 | 10.608 | +26% / +54% | 6.9610 | 8.9881 | 10.189 | +29% / +46%
neurips2023-1M | 11.313 | 13.294 | 14.593 | +17% / +28% | 11.489 | 13.479 | 14.594 | +17% / +27%
neurips2023-full-25pct | 26.420 | 30.656 | 33.720 | +16% / +27% | 26.353 | 31.023 | 34.479 | +17% / +30%
movies | 3.3224 | 3.9881 | 4.5630 | +20% / +37% | 3.2512 | 3.9569 | 4.5507 | +21% / +39%

### hottest
task | ram | ramc32 | ramc16 | diff | mmap | mmapc32 | mmapc16 | diff
| - | - | - | - | - | - | - | - | -
random-50k | 0.1857 | 0.19823 | 0.19553 | +6% / +5% | 0.18572 | 0.19651 | 0.1977 | +5% / +6%
random-500k | 1.5728 | 1.6731 | 1.6958 | +6% / +7% | 1.5822 | 1.7087 | 1.6814 | +7% / +6%
neurips2023-1M | 8.1935 | 9.0012 | 9.2419 | +9% / +12% | 8.2457 | 8.9325 | 9.1735 | +8% / +11%
neurips2023-full-25pct | 18.308 | 20.002 | 20.337 | +9% / +11% | 18.399 | 19.969 | 20.246 | +8% / +10%
movies | 0.17718 | 0.1958 | 0.23379 | +10% / +31% | 0.17592 | 0.19475 | 0.23019 | +10% / +30%


lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -39,7 +39,7 @@ use crate::vector_storage::{
     check_deleted_condition, new_stoppable_raw_scorer, VectorStorage, VectorStorageEnum,
 };
 
-pub struct SparseVectorIndex<TInvertedIndex: InvertedIndex> {
+pub struct SparseVectorIndex<TInvertedIndex: InvertedIndex<VectorElementType>> {
```
xzfc, line 42: Further generalization probably would start here.

lib/api/src/rest/schema.rs
```diff
@@ -19,7 +20,7 @@ pub type MultiDenseVector = Vec<DenseVector>;
 #[serde(untagged, rename_all = "snake_case")]
 pub enum Vector {
     Dense(DenseVector),
-    Sparse(sparse::common::sparse_vector::SparseVector),
+    Sparse(sparse::common::sparse_vector::SparseVector<VectorElementType>),
```
generall, line 23: I don't think we want to make `Vector` itself generic against datatype. It would require too much changes.

generall (changes requested):
I think we should follow the pattern of dense vector storage and _only_ change the data type inside the storage itself. We have already tried the approach with generalizing everything up to the interface, this proves to be a huge pain and not really helpful at the end.

It is fine if the input vectors will have f32 datatype, we can always convert them into f16 internally, during scoring.

generall (approved):

PR #4365, opened 2024-05-31
Improve tools/*.sh

xzfc:
1. Speedup `tools/generate_openapi_models.sh` by reducing the amount of `docker run` invocations.
    - On my machine, it's now faster by 5 seconds.
    - When `ytt` and `yq` are installed locally, it's faster by 7 seconds.
2. Clean up temp files in `tools/generate_grpc_docs.sh`. No need to pollute `/tmp`. Also, use a meaningful temporary directory name.
3. Do not call `sudo chmod` in `tools/generate_grpc_docs.sh`. [The causing issue](https://www.github.com/pseudomuto/protoc-gen-doc/issues/383) is solved.
4. Use proper shell quoting in `tools/generate_grpc_docs.sh`.

### DIFF
--- a/tools/generate_grpc_docs.sh
+++ b/tools/generate_grpc_docs.sh
@@ -6,28 +6,27 @@ set -e
 cd "$(dirname "$0")/../"
 
 # Create a temporary directory and store its name in a variable.
-TEMPD=$(mktemp -d)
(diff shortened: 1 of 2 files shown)


timvisee (approved):
I like that we removed the root requirement and that it can fall back to a local installation :smile:

PR #4369, opened 2024-05-31
Fix hnsw full scan threshold

IvanPleshkov:
Fix performance degradation on H&M dataset.

After this PR https://github.com/qdrant/qdrant/pull/4312, full scan threshold depends on vectors count in vector storage. The problem is, while `HNSWIndex.open` is running, vector storage may be empty - it's a case from optimizer.

The solution is like indexed vectors counting: update full scan threshold later (indexed point count updating: https://github.com/qdrant/qdrant/blob/dev/lib/segment/src/index/hnsw_index/hnsw.rs#L851)

Test coverage: low cardinality search check was added to the filterable hnsw unit test to check full scan threshold.

### DIFF
--- a/lib/segment/src/index/hnsw_index/config.rs
+++ b/lib/segment/src/index/hnsw_index/config.rs
@@ -20,7 +20,7 @@ pub struct HnswGraphConfig {
     ///
     /// Note: this is number of vectors, not KiloBytes.
     #[serde(alias = "indexing_threshold")]
-    pub full_scan_threshold: Option<usize>,
+    pub full_scan_threshold: usize,
     #[serde(default)]
     pub max_indexing_threads: usize,
     #[serde(default)]
@@ -35,19 +35,21 @@ impl HnswGraphConfig {
     pub fn new(
         m: usize,
         ef_construct: usize,
+        full_scan_threshold: usize,
         max_indexing_threads: usize,
         payload_m: Option<usize>,
+        indexed_vector_count: usize,
     ) -> Self {
         HnswGraphConfig {
             m,
             m0: m * 2,
             ef_construct,
             ef: ef_construct,
-            full_scan_threshold: None,
+            full_scan_threshold,
             max_indexing_threads,
             payload_m,
             payload_m0: payload_m.map(|v| v * 2),
-            indexed_vector_count: None,
+            indexed_vector_count: Some(indexed_vector_count),
         }
     }
 
--- a/lib/segment/src/index/hnsw_index/config.rs
+++ b/lib/segment/src/index/hnsw_index/config.rs
@@ -20,7 +20,7 @@ pub struct HnswGraphConfig {
     ///
     /// Note: this is number of vectors, not KiloBytes.
     #[serde(alias = "indexing_threshold")]
-    pub full_scan_threshold: usize,
+    pub full_scan_threshold: Option<usize>,
     #[serde(default)]
     pub max_indexing_threads: usize,
     #[serde(default)]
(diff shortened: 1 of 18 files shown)


lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -692,7 +680,19 @@ impl<TGraphLinks: GraphLinks> VectorIndex for HNSWIndex<TGraphLinks> {
...
             permit.num_cpus,
         );
-        let indexing_threshold = self.config.full_scan_threshold;
+        let available_vectors = vector_storage.available_vector_count();
+        let indexing_threshold = vector_storage
+            .available_size_in_bytes()
+            .checked_div(available_vectors)
+            .and_then(|avg_vector_size| {
+                self.hnsw_config
+                    .full_scan_threshold
```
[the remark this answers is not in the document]
IvanPleshkov: I think we can do this in a separate PR because it's not bug-related

lib/segment/src/index/hnsw_index/config.rs
```diff
@@ -35,21 +35,19 @@ impl HnswGraphConfig {
...
             m0: m * 2,
             ef_construct,
             ef: ef_construct,
-            full_scan_threshold,
+            full_scan_threshold: None,
             max_indexing_threads,
             payload_m,
             payload_m0: payload_m.map(|v| v * 2),
-            indexed_vector_count: Some(indexed_vector_count),
+            indexed_vector_count: None,
```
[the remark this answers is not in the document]
timvisee: Thanks! I was just being cautious.

lib/segment/src/index/hnsw_index/hnsw.rs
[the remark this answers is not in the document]
timvisee: Agreed.

Would you mind to make the change? I believe it's a serde structure, so we would need a `#[serde(rename = "full_scan_threshold")]` attribute above it.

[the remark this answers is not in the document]
IvanPleshkov: this changes have been reverted

lib/segment/src/segment_constructor/segment_builder.rs
```diff
@@ -191,34 +229,113 @@ impl SegmentBuilder {
...
+                temp_path,
+                indexed_fields,
+            } = self;
+
+            let appendable_flag = segment_config.is_appendable();
+
+            payload_storage.flusher()()?;
+            let payload_storage_arc = Arc::new(AtomicRefCell::new(payload_storage));
+
+            id_tracker.mapping_flusher()()?;
```
IvanPleshkov, line 256: We don't need to flush `id_tracker` here. it's already flushed in `update_from`

lib/segment/src/segment_constructor/segment_builder.rs
```diff
@@ -191,34 +229,113 @@ impl SegmentBuilder {
...
-            for (field, payload_schema) in &self.indexed_fields {
-                segment.create_field_index(segment.version(), field, Some(payload_schema))?;
-                check_process_stopped(stopped)?;
+                let Some(vector_storage) = vector_storages.remove(vector_name) else {
+                    return Err(OperationError::service_error(format!(
+                        "Vector storage for vector name {vector_name} not found on segment build"
+                    )));
+                };
+
+                vector_storage.flusher()()?;
```
IvanPleshkov, line 297: I propose to flush vector storages in `update_from` like for id tracker. because `update_from` is a place where vector storage is constructed. If you want to keep all flushes here in `build`, I propose to do it all in one place before dropping

lib/segment/src/segment_constructor/segment_builder.rs
generall: Better safe than sorry

IvanPleshkov: Yea, thats why I find id_tracker flush in `update_from` save - it's flushed even if `build` is not called. As alternative, I added another comment with proposal to do all flushes in one place just before dropping

generall: Id tracker might accumulate changes in RAM if not flushed. I would prefer to keep as is

lib/segment/src/segment_constructor/segment_builder.rs
generall: If we do it before dropping, we would need to go through Arc and RefCell stuff

generall: My idea was that flush should happen before we start using those objects for building index

PR #4370, opened 2024-05-31
Rocksdb remove should use the same options as put

agourlay:
For the sake of consistency, we should apply the same options to `remove` and `put` operations against RocksDB.

```rust
fn get_write_options() -> WriteOptions {
        let mut write_options = WriteOptions::default();
        write_options.set_sync(false); // default to false
        write_options.disable_wal(true); // default to false
        write_options
    }
```

The current setup might cause some storage inconsistencies which are difficult to debug given that we control precisely the exact moment of the flushing.

### DIFF
--- a/lib/segment/src/common/rocksdb_wrapper.rs
+++ b/lib/segment/src/common/rocksdb_wrapper.rs
@@ -166,9 +166,10 @@ impl DatabaseColumnWrapper {
     {
         let db = self.database.read();
         let cf_handle = self.get_column_family(&db)?;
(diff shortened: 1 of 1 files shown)


timvisee (approved):

agourlay:
As expected, this did not help with the inconsistency issue.

The following run https://github.com/qdrant/crasher/actions/runs/9319413588/job/25653927949 took place on this commit `603efeda059acda5f9350f098a455df0d1c2e44d`.

PR #4371, opened 2024-05-31
Log failed segment flush on drop

agourlay:
It is a good practice to not ignore silently persistence errors.

 This PR proposes to log at debug level failed segment flush on drop, this might help us one day.

### DIFF
--- a/lib/segment/src/segment.rs
+++ b/lib/segment/src/segment.rs
@@ -1923,7 +1923,9 @@ impl SegmentEntry for Segment {
 
 impl Drop for Segment {
     fn drop(&mut self) {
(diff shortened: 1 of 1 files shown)


lib/segment/src/segment.rs
timvisee, line 1: I assume we don't want a warning here because this isn't directly useful to users. Because even if we'd fail, we expect our WAL to recover it transparently. Am I right?

timvisee (approved):

lib/segment/src/segment.rs
agourlay: Yes, we put all our money on the WAL :+1:

PR #4375, opened 2024-05-31
Fix inplace updates for sparse index

generall:
In the current dev version, the procedure of updating mutable sparse index didn't take into account, that there might be existing vectors with non-overlapping elements.
Those elements were not removed from the posting lists on insertion.

This PR fixes the insertion by introducing an additional optional parameter `old_vector`, which is used to properly cleaned the storage.
Dense index doesn't require old vector to do proper updates, so it is only retrieved when we are updating sparse index in-place.

### DIFF
--- a/lib/segment/src/data_types/vectors.rs
+++ b/lib/segment/src/data_types/vectors.rs
@@ -352,14 +352,6 @@ impl<'a> VectorRef<'a> {
             VectorRef::MultiDense(v) => Vector::MultiDense(v.to_owned()),
         }
     }
-
-    pub fn is_sparse(&self) -> bool {
-        match self {
-            VectorRef::Dense(_) => false,
-            VectorRef::Sparse(_) => true,
-            VectorRef::MultiDense(_) => false,
-        }
-    }
 }
 
 impl<'a> TryInto<&'a [VectorElementType]> for &'a Vector {
--- a/lib/segment/src/data_types/vectors.rs
+++ b/lib/segment/src/data_types/vectors.rs
@@ -352,6 +352,14 @@ impl<'a> VectorRef<'a> {
             VectorRef::MultiDense(v) => Vector::MultiDense(v.to_owned()),
         }
     }
+
+    pub fn is_sparse(&self) -> bool {
+        match self {
+            VectorRef::Dense(_) => false,
+            VectorRef::Sparse(_) => true,
+            VectorRef::MultiDense(_) => false,
+        }
+    }
 }
 
 impl<'a> TryInto<&'a [VectorElementType]> for &'a Vector {
--- a/lib/segment/src/index/plain_payload_index.rs
+++ b/lib/segment/src/index/plain_payload_index.rs
@@ -371,6 +371,7 @@ impl VectorIndex for PlainIndex {
             vector_storage.insert_vector(id, vector)?;
         } else {
             if id as usize >= vector_storage.total_vector_count() {
+                debug_assert!(id as usize == vector_storage.total_vector_count());
                 // Vector doesn't exist in the storage
                 // Insert default vector to keep the sequence
                 let default_vector = vector_storage.default_vector();
--- a/lib/segment/src/index/plain_payload_index.rs
+++ b/lib/segment/src/index/plain_payload_index.rs
@@ -16,7 +16,7 @@ use crate::common::operation_time_statistics::{
 };
 use crate::common::{Flusher, BYTES_IN_KB};
 use crate::data_types::query_context::VectorQueryContext;
-use crate::data_types::vectors::{QueryVector, Vector, VectorRef};
+use crate::data_types::vectors::{QueryVector, VectorRef};
 use crate::id_tracker::IdTrackerSS;
 use crate::index::field_index::{CardinalityEstimation, PayloadBlockCondition};
 use crate::index::payload_config::PayloadConfig;
@@ -362,10 +362,23 @@ impl VectorIndex for PlainIndex {
 
     fn update_vector(
         &mut self,
-        _id: PointOffsetType,
-        _vector: VectorRef,
-        _old_vector: Option<Vector>,
+        id: PointOffsetType,
+        vector: Option<VectorRef>,
     ) -> OperationResult<()> {
+        let mut vector_storage = self.vector_storage.borrow_mut();
+
+        if let Some(vector) = vector {
+            vector_storage.insert_vector(id, vector)?;
+        } else {
+            if id as usize >= vector_storage.total_vector_count() {
+                // Vector doesn't exist in the storage
+                // Insert default vector to keep the sequence
+                let default_vector = vector_storage.default_vector();
+                vector_storage.insert_vector(id, VectorRef::from(&default_vector))?;
+            }
+            vector_storage.delete_vector(id)?;
+        }
+
         Ok(())
     }
 }
--- a/lib/segment/src/index/plain_payload_index.rs
+++ b/lib/segment/src/index/plain_payload_index.rs
@@ -16,7 +16,7 @@ use crate::common::operation_time_statistics::{
 };
 use crate::common::{Flusher, BYTES_IN_KB};
 use crate::data_types::query_context::VectorQueryContext;
-use crate::data_types::vectors::{QueryVector, VectorRef};
+use crate::data_types::vectors::{QueryVector, Vector, VectorRef};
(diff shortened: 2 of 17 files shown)


lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -584,19 +584,28 @@ impl<TInvertedIndex: InvertedIndex> VectorIndex for SparseVectorIndex<TInvertedI
...
                 "Cannot update vector in non-appendable index",
             ));
         }
 
         let vector: &SparseVector = vector.try_into()?;
+        let old_vector: Option<SparseVector> =
+            old_vector.map(SparseVector::try_from).transpose()?;
+
         // do not upsert empty vectors into the index
         if !vector.is_empty() {
```
xzfc, line 604: We need to delete the old vector even the new vector is empty.

lib/sparse/src/index/posting_list.rs
```diff
@@ -29,6 +29,18 @@ impl PostingList {
...
+            .elements
+            .binary_search_by_key(&record_id, |e| e.record_id);
+        if let Ok(found_index) = index {
+            self.elements.remove(found_index);
+            if found_index < self.elements.len() {
+                self.propagate_max_next_weight_to_the_left(found_index);
+            }
+        }
+    }
+
```
xzfc, line 43: Randomized test show that resulting `max_next_weight`s are not the same as if we recreate the same index from scratch without the deleted element.

But since the resulting `max_next_weight`s are greater or equal than the proper values, it'll not break the pruning logic, but may make it less efficient.

Not sure if it worth fixing giving that we probably get rid of of the pruning soon. But maybe add a commend and a test checking that `max_next_weight[i] >= max(weight[i..])`.

<details>
<summary>quick and dirty test code</summary>

```rust
#[test]
fn test_random_delete() {
    use rand::Rng;
    use rand::seq::SliceRandom;
    let mut rng = rand::thread_rng();
    for _ in 0..1000 {
        let mut ids = Vec::new();
        let mut cur_id = 0;
        for _ in 0..32 {
            cur_id += rng.gen_range(1..10);
            ids.push(cur_id);
        }
        ids.shuffle(&mut rng);
        let random_id = ids[rng.gen_range(0..ids.len())];

        let mut builder1 = PostingBuilder::new();
        let mut builder2 = PostingBuilder::new();
        for id in ids {
            let val = rng.gen_range(0..100) as f32 / 10.0;
            builder1.add(id, val);
            if id != random_id {
                builder2.add(id, val);
            }
        }

        let mut posting_list1 = builder1.build();
        posting_list1.delete(random_id);
        let posting_list2 = builder2.build();

        // Ok
        assert_eq!(
            posting_list1.elements.iter().map(|e| e.record_id).collect_vec(),
            posting_list2.elements.iter().map(|e| e.record_id).collect_vec(),
        );
        assert_eq!(
            posting_list1.elements.iter().map(|e| e.weight).collect_vec(),
            posting_list2.elements.iter().map(|e| e.weight).collect_vec(),
        );

        // Fail
        assert_eq!(
            posting_list1.elements.iter().map(|e| e.max_next_weight).collect_vec(),
            posting_list2.elements.iter().map(|e| e.max_next_weight).collect_vec(),
        );

        // Ok (at least they won't break pruning logic)
        assert!(
            std::iter::zip(
                &posting_list1.elements,
                &posting_list2.elements,
            ).all(|(e1, e2)| e1.max_next_weight >= e2.max_next_weight),
        );
    }
}
```

</details>

lib/segment/src/vector_storage/simple_sparse_vector_storage.rs
```diff
@@ -105,15 +100,16 @@ impl SimpleSparseVectorStorage {
         vector: Option<&SparseVector>,
     ) -> OperationResult<()> {
         // Write vector state to buffer record
-        let record = &mut self.update_buffer;
-        record.deleted = deleted;
+        let record = StoredSparseVector {
+            deleted,
+            vector: vector.cloned().unwrap_or_default(),
+        };
```
xzfc, line 106: This change is not related to the issue being fixed?

(not an objection to include it, just trying to follow)

lib/segment/src/segment.rs
```diff
@@ -181,15 +194,32 @@ impl Segment {
...
         for (vector_name, new_vector) in vectors {
             let vector_data = &self.vector_data[vector_name.as_ref()];
+            let old_vector = if new_vector.as_vec_ref().is_sparse() {
+                // Sparse vector index requires an old vector to be passed
+                // to update operation, so it can clean up old index entries properly
+                //
+                // So we retrieve it here.
+                vector_data
+                    .vector_storage
+                    .borrow()
```
xzfc, line 204: Performance nit: maybe borrow `vector_storage` once per loop instead of borrowing it twice for each loop iteration?

lib/segment/src/index/sparse_index/sparse_vector_index.rs
generall: Good catch!

lib/segment/src/vector_storage/simple_sparse_vector_storage.rs
generall: yes, not directly related (sorry)

lib/sparse/src/index/posting_list.rs
generall: Fixed the correctness, but the code doesn't look nice.

Let's remove this heuristic in the next PR

lib/sparse/src/index/posting_list.rs
```diff
@@ -86,13 +105,7 @@ impl PostingList {
         for element in self.elements[..up_to_index].iter_mut().rev() {
             // update max_next_weight for element
             element.max_next_weight = max_next_weight;
-            if element.weight >= max_next_weight {
```
agourlay, line 89: why did you change the logic here?

generall: because not we need to not only propagate weights when they are inserted (increase only) but also when they are removed (decrease). But this logic is up to removal anyway as per discussion here https://github.com/qdrant/qdrant/pull/4375#discussion_r1623000524

lib/segment/src/data_types/vectors.rs
```diff
@@ -352,6 +352,14 @@ impl<'a> VectorRef<'a> {
             VectorRef::MultiDense(v) => Vector::MultiDense(v.to_owned()),
         }
     }
+
+    pub fn is_sparse(&self) -> bool {
```
agourlay, line 356: do you still need this one?

lib/segment/src/index/vector_index_base.rs
```diff
@@ -51,7 +51,11 @@ pub trait VectorIndex {
     fn indexed_vector_count(&self) -> usize;
 
     /// Update index for a single vector
-    fn update_vector(&mut self, id: PointOffsetType, vector: VectorRef) -> OperationResult<()>;
+    fn update_vector(
```
agourlay, line 54: please add a comment regarding the meaning of the optional vector :)

generall: Done

agourlay (approved):

lib/segment/src/index/plain_payload_index.rs
```diff
@@ -360,7 +360,25 @@ impl VectorIndex for PlainIndex {
...
+        if let Some(vector) = vector {
+            vector_storage.insert_vector(id, vector)?;
+        } else {
+            if id as usize >= vector_storage.total_vector_count() {
+                // Vector doesn't exist in the storage
+                // Insert default vector to keep the sequence
+                let default_vector = vector_storage.default_vector();
+                vector_storage.insert_vector(id, VectorRef::from(&default_vector))?;
+            }
+            vector_storage.delete_vector(id)?;
```
timvisee, line 379: Is inserting one here always enough? Or do we want a loop until the ID matches the vector count?

timvisee (approved):

lib/segment/src/index/plain_payload_index.rs
generall: this logic is not new, it migrated from `insert_new_vectors` function of the segment. In practice `PointOffsetType` is always auto-incremented, we should not have gaps in the vector storage

xzfc (approved):

PR #4381, opened 2024-06-03
Replicate new shard to match replication factor in resharding driver

timvisee:
Tracked in: #4213 
Depends on: <https://github.com/qdrant/qdrant/pull/4379>

Implements the third stage in the resharding driver. It keeps replicating our new shard until we match the desired replication factor. This is dynamic and will follow the latest configured replication factor, even while it changes during this process. If we don't have enough peers to allow that number of replicas it continues early.

This does not implement the following yet, marked as TODOs:
- using the preferred shard transfer method, as configured by the user
- balance the shard distribution across the peers

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/resharding/driver.rs
+++ b/lib/collection/src/shards/resharding/driver.rs
@@ -488,16 +488,16 @@ fn stage_finalize() -> CollectionResult<()> {
     todo!()
 }
 
-/// Await for a resharding shard trnasfer to succeed.
+/// Await for a resharding shard transfer to succeed.
 ///
-/// Yields on a successfull transfer.
+/// Yields on a successful transfer.
 ///
 /// Returns an error if:
 /// - the transfer failed or got aborted
 /// - the transfer timed out
 /// - no matching transfer is ongoing; it never started or went missing without a notification
 ///
-/// Yields on a succesful transfer. Returns an error if an error occurred or if the global timeout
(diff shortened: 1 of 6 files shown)


lib/collection/src/shards/resharding/driver.rs
```diff
@@ -363,3 +487,79 @@ fn stage_propagate_deletes() -> CollectionResult<()> {
...
+/// Yields on a succesful transfer. Returns an error if an error occurred or if the global timeout
+/// is reached.
+async fn await_transfer_success(
+    reshard_key: &ReshardKey,
+    transfer: &ShardTransfer,
+    shard_holder: &Arc<LockedShardHolder>,
+    collection_id: &CollectionId,
+    consensus: &dyn ShardTransferConsensus,
+    await_transfer_end: impl Future<Output = CollectionResult<Result<(), ()>>>,
+) -> CollectionResult<()> {
```
timvisee, line 509: Though this and it's usage may seem a bit complex, there's three important things it does:

1. we subscribe to shard transfer changes before we actually start transfers, to ensure we don't miss notifications
2. it enforces a timeout, we don't wait longer than 24 hours for a transfer to complete
3. it does a periodic sanity check to ensure the transfer is still active in consensus, if it isn't we assume it has failed (and panic in debug mode because this cannot happen)

ffuugoo (approved):
LGTM. Sorry for taking so long to review this. 🙈

PR #4394, opened 2024-06-04
Skip serializing null fields in `Record` and `ScoredPoint` (REST)

coszio:
With the intention of making the interface simpler, I propose to skip serializing the optional fields in `Record` and `ScoredPoint`.

From my POV, this has two advantages:
- Less noise in the response when `with_payload` and `with_vector` are not requested.
- Less confusion from users thinking their points don't have payload/vector ("It says payload is null!")

Python client will keep deserializing them into their classes as `None` fields, though, so the advantages only apply when doing plain REST requests.

### DIFF
--- a/lib/api/src/rest/schema.rs
+++ b/lib/api/src/rest/schema.rs
@@ -97,13 +97,16 @@ pub struct ScoredPoint {
     /// Points vector distance to the query vector
     pub score: common::types::ScoreType,
     /// Payload - values assigned to the point
+    #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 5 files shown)


coszio:
@Anush008 Do you know if this would a breaking change with other clients/integrations?

Anush008:
I will have to take a look. Will update here.

Anush008:
The integrations should all be good. Java and C# clients use gRPC. So shouldn't affect those as well.

coszio:
Drafting to prevent merging before a patch release

generall (approved):

PR #4399, opened 2024-06-05
Add immutable ID tracker

JojiiOfficial:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/id_tracker/id_tracker_base.rs
+++ b/lib/segment/src/id_tracker/id_tracker_base.rs
@@ -1,3 +1,5 @@
+use std::path::Path;
+
 use bitvec::prelude::BitSlice;
 use common::types::PointOffsetType;
 use rand::rngs::StdRng;
@@ -5,6 +7,7 @@ use rand::{Rng, SeedableRng};
 
 use crate::common::operation_error::OperationResult;
 use crate::common::Flusher;
+use crate::id_tracker::immutable_id_tracker::ImmutableIdTracker;
 use crate::id_tracker::simple_id_tracker::SimpleIdTracker;
 use crate::types::{PointIdType, SeqNumberType};
 
@@ -112,6 +115,10 @@ pub trait IdTracker {
     /// Check whether the given point is soft deleted
     fn is_deleted_point(&self, internal_id: PointOffsetType) -> bool;
 
+    fn make_immutable(&self, save_path: &Path) -> OperationResult<IdTrackerEnum>;
+
+    fn name(&self) -> &'static str;
+
     /// Iterator over `n` random IDs which are not deleted
     ///
     /// A [`BitSlice`] of deleted vectors may optionally be given to also consider deleted named
@@ -148,7 +155,7 @@ pub type IdTrackerSS = dyn IdTracker + Sync + Send;
 
 pub enum IdTrackerEnum {
     MutableIdTracker(SimpleIdTracker),
-    ImmutableIdTracker(SimpleIdTracker),
+    ImmutableIdTracker(ImmutableIdTracker),
 }
 
 impl IdTracker for IdTrackerEnum {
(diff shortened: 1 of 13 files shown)


lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -287,6 +293,22 @@ impl IdTracker for IdTrackerEnum {
...
     }
 
+    fn make_immutable(&self, save_path: &Path) -> OperationResult<IdTrackerEnum> {
+        match self {
+            IdTrackerEnum::MutableIdTracker(id_tracker) => id_tracker.make_immutable(save_path),
+            IdTrackerEnum::ImmutableIdTracker(id_tracker) => {
+                Ok(id_tracker.make_immutable(save_path)?)
+            }
+        }
+    }
```
timvisee, line 303: Instead of `make_immutable` on all implementations, I think I'd prefer `ImmutableIdTracker::from(other_id_tracker)`.

What do you think?

lib/segment/src/id_tracker/simple_id_tracker.rs
```diff
@@ -405,6 +434,41 @@ impl IdTracker for SimpleIdTracker {
...
+               };
+
+               let immutable_tracker =
+                   ImmutableIdTracker::new(save_path.to_path_buf(), self.deleted.clone(), mappings);
+
+               // Make it persistent
+               immutable_tracker.save()?;
+
+               Ok(IdTrackerEnum::ImmutableIdTracker(immutable_tracker))
+        */
```
timvisee, line 465: Rather than commented-out bits I'd prefer not having them here at all, and adding them in a separate commit (in a new PR).

Do you think that's possible without too much effort?

If that does take a lot of effort I'm fine with leaving them in this time.

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,429 @@
...
+            Self::save_file(&path, &deleted)?;
+            *dirty.write() = false;
+            Ok(())
+        })
+    }
+
+    /// Not implemented for immutable id tracker.
+    fn versions_flusher(&self) -> Flusher {
+        Box::new(|| Ok(()))
+    }
```
timvisee, line 265: We will need this eventually.

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,429 @@
...
+
+        let deleted = self.deleted.clone();
+        let path = self.deleted_file_path();
+        let dirty = self.dirty.clone();
+        Box::new(move || {
+            Self::save_file(&path, &deleted)?;
+            *dirty.write() = false;
+            Ok(())
+        })
+    }
```
timvisee, line 260: If we move to the approach using memory mapping, we can remove the dirty flag and simply return it's associated flusher.

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,429 @@
...
+            mappings,
+        }
+    }
+
+    pub(super) fn save(&self) -> OperationResult<()> {
+        Self::save_file(&self.deleted_file_path(), &self.deleted)?;
+        Self::save_file(&self.mappings_file_path(), &self.mappings)?;
+        *self.dirty.write() = false;
+        Ok(())
+    }
```
timvisee, line 74: Is there a specific reason we need this?

I expect we want to save once when creating this immutable structure. But then I'd prefer to use the `::from(other)` approach, automatically saving once, keeping it internal only.

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,429 @@
...
+            .get(internal_id as usize)
+            .copied()
+    }
+
+    fn set_internal_version(
+        &mut self,
+        _internal_id: PointOffsetType,
+        _version: SeqNumberType,
+    ) -> OperationResult<()> {
+        panic!("Trying to call a mutating function (`set_internal_version`) of an immutable id tracker");
```
timvisee, line 117: We will need this eventually. Or well, I believe we need to be able to use this to update the version on delete.

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,429 @@
...
+                .filter(|i| i.0.is_uuid())
+                .map(|(k, v)| (k.into(), *v))
+        };
+        let offset_uuid_iter = |offset: Uuid| {
+            self.mappings
+                .external_to_internal
+                .range(StoredPointId::Uuid(offset)..)
+                .filter(|i| i.0.is_uuid())
+                .map(|(k, v)| (k.into(), *v))
+        };
```
timvisee, line 217: Two separate types for numeric and UUID IDs are probably better because this may potentially be inefficient. We now have to go over the full list twice, with a more expensive iterator.

lib/segment/src/id_tracker/id_tracker_base.rs
JojiiOfficial: I think this is not possible. [Here](https://github.com/qdrant/qdrant/blob/dd326e9091bf18db73230344ee3cd998818f8614/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs#L380) we only have access to the `dyn IdTracker` trait, not the simple id tracker value. This is why I added this conversion function on trait level.

lib/segment/src/id_tracker/immutable_id_tracker.rs
JojiiOfficial: Fixed in the other PR

Issue #4199, opened 2024-05-08
Some collections stop responding while the server seems ok


hugh2slowmo:
## Current Behavior
version: 1.9.1
We facing some weird not response issue yesterday, some of the collections stop working on any request including /search /search/batch /scroll, even /collections/collection_name, i try to get collection status to give me some insight, but it hangs, i also try using curl on server itself so there may not any kind of network issues, and there are no any timeout logs, we also check the cpu, mem, disk io, but everything looks quite well, for make our app back in life quickly, we choose to reboot qdrant then.
Some related machine info shows below, the peek in the middle is cause by a short query benchmark to make sure it works fine after we reboot:
![image](https://github.com/qdrant/qdrant/assets/27076640/e785f198-0631-4794-bd76-632db7c3ed18)
![image](https://github.com/qdrant/qdrant/assets/27076640/81547e54-b94f-4e10-967d-31622fa1b235)
![image](https://github.com/qdrant/qdrant/assets/27076640/54a66754-dd1c-425f-830b-96526c6d9014)
![image](https://github.com/qdrant/qdrant/assets/27076640/73c52d29-edc0-48e0-ba99-99e46b594ace)

## Steps to Reproduce
Every thing looks good since we restarted it yesterday, don't know how to reproduce.
Through digging around logs i found one thing that i don't understand but might lead to some issues:
```
2024-05-07T04:23:15.180072Z  WARN collection::collection_manager::holders::segment_holder: Trying to read-lock all collection segments is taking a long time. This could be a deadlock and may block new updates.    
2024-05-07T04:23:28.601938Z  WARN collection::collection_manager::holders::segment_holder: Trying to read-lock all collection segments is taking a long time. This could be a deadlock and may block new updates.    
2024-05-07T04:23:55.445584Z  WARN collection::collection_manager::holders::segment_holder: Trying to read-lock all collection segments is taking a long time. This could be a deadlock and may block new updates.    
2024-05-07T04:24:49.132774Z  WARN collection::collection_manager::holders::segment_holder: Trying to read-lock all collection segments is taking a long time. This could be a deadlock and may block new updates.    
2024-05-07T04:26:36.507051Z  WARN collection::collection_manager::holders::segment_holder: Trying to read-lock all collection segments is taking a long time. This could be a deadlock and may block new updates.    
[...]

PR #4402, opened 2024-06-06
Fix deadlock caused by concurrent snapshot and optimization

timvisee:
Fix a deadlock that could occur when a snapshot happens during an active optimization.

With our non-blocking snapshot implementation, all segments are proxied for snapshotting. When unproxying, all segments get a new unique ID assigned. This causes the active optimization to loose track of the segments, because IDs don't match up anymore. If the optimizer unproxies, we get a broken state.

More specifically, when this happens we end up with both the proxy and the wrapped segment in the segment holder. The segment is effectively placed twice in the holder now. It causes a double read, causing a deadlock.

This PRs does a few things to mitigate the problem:
- when unproxing after a snapshot, put the segment back with their original ID
- change segment ID generation to be auto incrementing, prevent potential conflicts when putting back old IDs
- add a (debug) sanity check when unproxying in the optimizer, make sure we remove the expected number of proxies
- function names have been changed to clarify the segment ID behavior better
- extend a snapshot test to assert correct behavior

This also fixes the following message that popped up due to a double read (e.g. <https://github.com/qdrant/qdrant/issues/4199>):

> Trying to read-lock all collection segments is taking a long time. This could be a deadlock and may block new updates.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/fixtures.rs
+++ b/lib/collection/src/collection_manager/fixtures.rs
@@ -207,8 +207,8 @@ pub fn build_test_holder(path: &Path) -> RwLock<SegmentHolder> {
 
     let mut holder = SegmentHolder::default();
 
-    let _sid1 = holder.add(segment1);
-    let _sid2 = holder.add(segment2);
+    let _sid1 = holder.add_new(segment1);
+    let _sid2 = holder.add_new(segment2);
 
     RwLock::new(holder)
 }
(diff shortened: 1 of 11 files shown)


timvisee:
> Bug is not longer reproducing on my setup

Neither is it on mine. Thanks for checking!

Along with the above explanation this seems like a solid case closed.

generall (approved):

PR #4403, opened 2024-06-06
Persistent immutable id tracker

JojiiOfficial:
replaces #4399 

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -3,6 +3,7 @@ use std::path::Path;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
+use atomic_refcell::AtomicRefCell;
 use common::cpu::CpuPermit;
 use common::disk::dir_size;
 use io::storage_version::StorageVersion;
@@ -431,7 +432,6 @@ pub trait SegmentOptimizer {
 
         let mut optimized_segment: Segment = segment_builder.build(permit, stopped)?;
 
-        /* TODO: uncomment this when we integrate the new immutable id tracker
         // Make the IdTracker immutable if built segment is not appendable.
         if !optimized_segment.appendable_flag {
             let immutable = optimized_segment
@@ -440,7 +440,6 @@ pub trait SegmentOptimizer {
                 .make_immutable(&optimized_segment.current_path)?;
             optimized_segment.id_tracker = Arc::new(AtomicRefCell::new(immutable));
         }
-         */
 
         // Delete points in 2 steps
         // First step - delete all points with read lock
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -3,7 +3,6 @@ use std::path::Path;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
-use atomic_refcell::AtomicRefCell;
 use common::cpu::CpuPermit;
 use common::disk::dir_size;
 use io::storage_version::StorageVersion;
@@ -432,6 +431,7 @@ pub trait SegmentOptimizer {
 
         let mut optimized_segment: Segment = segment_builder.build(permit, stopped)?;
 
+        /* TODO: uncomment this when we integrate the new immutable id tracker
         // Make the IdTracker immutable if built segment is not appendable.
         if !optimized_segment.appendable_flag {
             let immutable = optimized_segment
@@ -440,6 +440,7 @@ pub trait SegmentOptimizer {
                 .make_immutable(&optimized_segment.current_path)?;
             optimized_segment.id_tracker = Arc::new(AtomicRefCell::new(immutable));
         }
+         */
 
         // Delete points in 2 steps
         // First step - delete all points with read lock
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -3,6 +3,7 @@ use std::path::Path;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
+use atomic_refcell::AtomicRefCell;
 use common::cpu::CpuPermit;
 use common::disk::dir_size;
 use io::storage_version::StorageVersion;
@@ -431,6 +432,15 @@ pub trait SegmentOptimizer {
 
         let mut optimized_segment: Segment = segment_builder.build(permit, stopped)?;
 
+        // Make the IdTracker immutable if built segment is not appendable.
+        if !optimized_segment.appendable_flag {
+            let immutable = optimized_segment
+                .id_tracker
+                .borrow()
+                .make_immutable(&optimized_segment.current_path)?;
+            optimized_segment.id_tracker = Arc::new(AtomicRefCell::new(immutable));
+        }
+
         // Delete points in 2 steps
         // First step - delete all points with read lock
         // Second step - delete all the rest points with full write lock
@@ -676,6 +686,7 @@ pub trait SegmentOptimizer {
                 tmp_segment.drop_data()?;
             }
         }
+
         timer.set_success(true);
         Ok(true)
     }
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -3,7 +3,6 @@ use std::path::Path;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
-use atomic_refcell::AtomicRefCell;
 use common::cpu::CpuPermit;
 use common::disk::dir_size;
 use io::storage_version::StorageVersion;
@@ -431,15 +430,6 @@ pub trait SegmentOptimizer {
         }
 
         let mut optimized_segment: Segment = segment_builder.build(permit, stopped)?;
-
-        // Make the IdTracker immutable if built segment is not appendable.
-        if !optimized_segment.appendable_flag {
-            let immutable = optimized_segment
-                .id_tracker
-                .borrow()
-                .make_immutable(&optimized_segment.current_path)?;
-            optimized_segment.id_tracker = Arc::new(AtomicRefCell::new(immutable));
-        }
 
         // Delete points in 2 steps
         // First step - delete all points with read lock
--- a/lib/segment/src/id_tracker/id_tracker_base.rs
+++ b/lib/segment/src/id_tracker/id_tracker_base.rs
@@ -8,6 +8,7 @@ use rand::{Rng, SeedableRng};
 
 use crate::common::operation_error::OperationResult;
 use crate::common::Flusher;
+use crate::id_tracker::immutable_id_tracker::ImmutableIdTracker;
 use crate::id_tracker::simple_id_tracker::SimpleIdTracker;
 use crate::types::{PointIdType, SeqNumberType};
 
@@ -156,7 +157,7 @@ pub type IdTrackerSS = dyn IdTracker + Sync + Send;
 #[derive(Debug)]
 pub enum IdTrackerEnum {
     MutableIdTracker(SimpleIdTracker),
-    ImmutableIdTracker(SimpleIdTracker),
+    ImmutableIdTracker(ImmutableIdTracker),
 }
 
 impl IdTracker for IdTrackerEnum {
--- a/lib/segment/src/id_tracker/id_tracker_base.rs
+++ b/lib/segment/src/id_tracker/id_tracker_base.rs
@@ -8,7 +8,6 @@ use rand::{Rng, SeedableRng};
 
 use crate::common::operation_error::OperationResult;
 use crate::common::Flusher;
-use crate::id_tracker::immutable_id_tracker::ImmutableIdTracker;
 use crate::id_tracker::simple_id_tracker::SimpleIdTracker;
 use crate::types::{PointIdType, SeqNumberType};
 
@@ -157,7 +156,7 @@ pub type IdTrackerSS = dyn IdTracker + Sync + Send;
 #[derive(Debug)]
 pub enum IdTrackerEnum {
     MutableIdTracker(SimpleIdTracker),
-    ImmutableIdTracker(ImmutableIdTracker),
+    ImmutableIdTracker(SimpleIdTracker),
 }
 
 impl IdTracker for IdTrackerEnum {
--- a/lib/segment/src/id_tracker/id_tracker_base.rs
+++ b/lib/segment/src/id_tracker/id_tracker_base.rs
@@ -1,4 +1,5 @@
 use std::fmt;
+use std::path::Path;
 
 use bitvec::prelude::BitSlice;
 use common::types::PointOffsetType;
@@ -7,6 +8,7 @@ use rand::{Rng, SeedableRng};
 
 use crate::common::operation_error::OperationResult;
 use crate::common::Flusher;
+use crate::id_tracker::immutable_id_tracker::ImmutableIdTracker;
 use crate::id_tracker::simple_id_tracker::SimpleIdTracker;
 use crate::types::{PointIdType, SeqNumberType};
 
@@ -114,6 +116,10 @@ pub trait IdTracker: fmt::Debug {
     /// Check whether the given point is soft deleted
     fn is_deleted_point(&self, internal_id: PointOffsetType) -> bool;
 
+    fn make_immutable(&self, save_path: &Path) -> OperationResult<IdTrackerEnum>;
+
+    fn name(&self) -> &'static str;
+
     /// Iterator over `n` random IDs which are not deleted
     ///
     /// A [`BitSlice`] of deleted vectors may optionally be given to also consider deleted named
@@ -151,7 +157,7 @@ pub type IdTrackerSS = dyn IdTracker + Sync + Send;
 #[derive(Debug)]
 pub enum IdTrackerEnum {
     MutableIdTracker(SimpleIdTracker),
-    ImmutableIdTracker(SimpleIdTracker),
+    ImmutableIdTracker(ImmutableIdTracker),
 }
 
 impl IdTracker for IdTrackerEnum {
@@ -287,6 +293,22 @@ impl IdTracker for IdTrackerEnum {
             IdTrackerEnum::ImmutableIdTracker(id_tracker) => {
                 id_tracker.is_deleted_point(internal_id)
             }
+        }
+    }
+
+    fn make_immutable(&self, save_path: &Path) -> OperationResult<IdTrackerEnum> {
+        match self {
+            IdTrackerEnum::MutableIdTracker(id_tracker) => id_tracker.make_immutable(save_path),
+            IdTrackerEnum::ImmutableIdTracker(id_tracker) => {
+                Ok(id_tracker.make_immutable(save_path)?)
+            }
+        }
+    }
+
+    fn name(&self) -> &'static str {
+        match self {
+            IdTrackerEnum::MutableIdTracker(id_tracker) => id_tracker.name(),
+            IdTrackerEnum::ImmutableIdTracker(id_tracker) => id_tracker.name(),
         }
     }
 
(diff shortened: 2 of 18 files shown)


lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -371,6 +372,15 @@ pub trait SegmentOptimizer {
...
         let mut optimized_segment: Segment = segment_builder.build(permit, stopped)?;
 
+        // Make the IdTracker immutable if built segment is not appendable.
+        if !optimized_segment.appendable_flag {
+            let immutable = optimized_segment
+                .id_tracker
+                .borrow()
+                .make_immutable(&optimized_segment.current_path)?;
+            optimized_segment.id_tracker = Arc::new(AtomicRefCell::new(immutable));
+        }
```
generall, line 382: This should rather happen inside `segment_builder.build`

generall: `segment_builder` should persist all files for immutable id_tracker, and then load segment with proper id_tracker based on configuration. 

In other words, content of the segment should always be defined by it's config, and we should not patch is afterwards

lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -112,6 +115,10 @@ pub trait IdTracker {
     /// Check whether the given point is soft deleted
     fn is_deleted_point(&self, internal_id: PointOffsetType) -> bool;
 
+    fn make_immutable(&self, save_path: &Path) -> OperationResult<IdTrackerEnum>;
```
generall, line 118: Not sure this belongs to the trait

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,446 @@
...
+use uuid::Uuid;
+
+use crate::common::mmap_type::{MmapBitSlice, MmapSlice};
+use crate::common::operation_error::{OperationError, OperationResult};
+use crate::common::Flusher;
+use crate::id_tracker::simple_id_tracker::StoredPointId;
+use crate::id_tracker::{IdTracker, IdTrackerEnum};
+use crate::types::{PointIdType, SeqNumberType};
+
+pub const DELETED_FILE_NAME: &str = "id_tracker.deleted";
```
generall, line 21: Since we create new files, it should be used in `take_snapshot` of segment.

I propose to do similar to other components and put `fn files(&self) -> Vec<PathBuf>` into IdTracker trait

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,446 @@
...
+pub const VERSION_MAPPING_FILE_NAME: &str = "id_tracker.versions";
+
+pub struct ImmutableIdTracker {
+    deleted: MmapBitSlice,
+    internal_to_version: MmapSlice<SeqNumberType>,
+    mappings: PointMappings,
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
+pub(super) struct PointMappings {
```
generall, line 32: Serializing the whole structure is redundant. To recreate the structure we only need a list if (StoredPointId, PointOffsetType) pairs.

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,446 @@
...
+                description: format!("id tracker can't be serialized: {:?}", err),
+            }
+        })?;
+        Ok(())
+    }
+}
+
+/// Returns the required mmap filesize for a `BitSlice`.
+fn bitmap_mmap_size(deleted: &BitSlice) -> usize {
+    let num_bytes = deleted.len().div_ceil(8); // used bytes
```
generall, line 130: Don't use magic constants

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,446 @@
...
+    }
+
+    fn drop(&mut self, external_id: PointIdType) -> OperationResult<()> {
+        let internal_id = match external_id {
+            PointIdType::NumId(num) => self.mappings.external_to_internal_num.get(&num).copied(),
+            PointIdType::Uuid(uuid) => self.mappings.external_to_internal_uuid.get(&uuid).copied(),
+        };
+
+        if let Some(internal_id) = internal_id {
+            self.deleted.set(internal_id as usize, true);
```
generall, line 186: Similar to simple ID tracker, we need to ensure, that data on disk is not changed before flush. We would need to use something similar to `DatabaseColumnScheduledUpdateWrapper`

generall: Same goes to version mmap

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -282,12 +283,20 @@ pub(crate) fn create_payload_storage(
     Ok(payload_storage)
 }
 
-pub(crate) fn create_id_tracker(database: Arc<RwLock<DB>>) -> OperationResult<IdTrackerEnum> {
+pub(crate) fn create_mutable_id_tracker(
+    database: Arc<RwLock<DB>>,
+) -> OperationResult<IdTrackerEnum> {
```
generall, line 288: Would be more convenient, if this function returns just `SimpleIdTracker`, not enum

generall (changes requested):

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,532 @@
...
+                }
+                .unwrap();
+                (external, internal_id)
+            })
+            .collect();
+        Self::serialize_file(path, &mappings)
+    }
+
+    fn load_mappings(path: &Path) -> OperationResult<PointMappings> {
+        let mappings: Vec<(StoredPointId, PointOffsetType)> = Self::deserialize_file(path)?;
```
generall, line 151: I think at some scale this can be a bottleneck, when we reading all points in memory and then build data structure.

Consider 1B points (something we ideally want to run on a single machine, assuming we have MMAP for everything). 

For the PointsMapping structure it would be `1_000_000_000 x "(24 + 16 + 4)" / 1024 / 1024 / 1024 ~= 40Gb RAM`

If we need double of than to just load from disk, that might be just to expensive. 
Potential downside - reading one-by-one might be slower, cause we would need to call deserialize function for each point (not sure how significant would it be)


Probably need a second opinion @timvisee

JojiiOfficial: I agree. We could also try to use a custom deserializer, that directly builds the required `PointMappings` Struct in [visit_seq](https://docs.rs/serde/1.0.203/serde/de/trait.Visitor.html#method.visit_seq) so we don't allocate a `Vec` first. I'd need to test and verify whether this really is faster and doesn't allocate too much memory but I can imagine this to work.

JojiiOfficial: TL;DR: This approach is ~15% faster and ~2.5x more memory efficient.

I've implemented and analyzed this approach which confirms that the custom deserializer uses way less memory. Furthermore I tested execution time (point count: 763000 and 3203400) and the new approach is ~15%  faster as well.
I'm not 100% sure if the amount of memory reported on the screenshots are only from loading itself because there seems to be some overhead of the profiler but with those metrics the new approach uses ~2.5 times less memory. This also confirms that there is no other large allocation internally.

I think this is the most performant approach that can be done with serde at the moment. The next step would be to do it our own and write/read the bytes directly into/from the file.
If you're interested in that I can sure try it out and measure its performance.

image
image

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,582 @@
...
+        }
+    }
+
+    fn iter_ids(&self) -> Box<dyn Iterator<Item = PointOffsetType> + '_> {
+        self.iter_internal()
+    }
+
+    /// Creates a flusher function, that writes the deleted points bitvec to disk.
+    fn mapping_flusher(&self) -> Flusher {
+        self.deleted_wrapper.flusher()
```
timvisee, line 302: Am I correct that we only flush the deletions here because the mappings are constant?

I'd prefer to add a comment to make this clear:

```suggestion
        // Only flush deletions because mappings are immutable
        self.deleted_wrapper.flusher()
```

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,582 @@
...
+        self.internal_to_version_wrapper.flusher()
+    }
+
+    fn total_point_count(&self) -> usize {
+        self.mappings.internal_to_external.len()
+    }
+
+    fn available_point_count(&self) -> usize {
+        self.mappings.external_to_internal_num.len() + self.mappings.external_to_internal_uuid.len()
+    }
```
timvisee, line 316: I don't think this is correct. It should decrease if we mark a point is deleted.

Or does that indeed happen and am I missing an important step somewhere else?

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,582 @@
...
+
+    fn deleted_point_count(&self) -> usize {
+        self.total_point_count() - self.available_point_count()
+    }
```
timvisee, line 320: IIRC, in other cases we keep track of the number of deletions because it's a bit more reliable.

The available point count then becomes: `total - deleted`

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,582 @@
...
+    pub(super) struct PointMappingVisitor;
+
+    impl<'de> Visitor<'de> for PointMappingVisitor {
+        type Value = PointMappings;
+
+        fn expecting(&self, formatter: &mut Formatter) -> std::fmt::Result {
+            formatter.write_str("a sequence of tuples")
+        }
+
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
```
timvisee, line 421:
```suggestion
        // Custom deserializer to prevent allocating a humongous intermediate list of points
        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
```

lib/segment/src/id_tracker/immutable_id_tracker.rs
timvisee: Nice! We're also interested in less allocations too, but I guess that's an automatic side effect of this approach.

I haven't inspected the code around this yet, but the results look promising.

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,582 @@
...
+
+pub const DELETED_FILE_NAME: &str = "id_tracker.deleted";
+pub const MAPPINGS_FILE_NAME: &str = "id_tracker.mappings";
+pub const VERSION_MAPPING_FILE_NAME: &str = "id_tracker.versions";
+
+pub struct ImmutableIdTracker {
+    path: PathBuf,
+
+    deleted: BitVec,
+    deleted_wrapper: MmapBitSliceBufferedUpdateWrapper,
```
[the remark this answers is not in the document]
generall: We need to have separate in-memory copy to be able to write only on flush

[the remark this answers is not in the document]
JojiiOfficial: > This means we have the deleted bitvec in memory twice

Do memory maps really copy/cache their entire content into memory? I haven't seen this prefaulting and writing is combined with directly flushing it to the file. So in theory it shouldn't use that much memory - but I can be wrong.

> Am I correct that we need this because the RwLock is in the way, and we cannot remove because we need to prevent concurrent access?

Exactly. We need to be able to dereference it [here](https://github.com/qdrant/qdrant/blob/0a375583bfdbb19376c64e4235c0eb9f275af143/lib/segment/src/id_tracker/immutable_id_tracker.rs#L321C1-L324C6)

lib/segment/src/id_tracker/immutable_id_tracker.rs
generall: Looks correct to be, since new in-memory representation is equivalent to the SimpleIdTracker

generall: hm, maybe we should modify it then?

JojiiOfficial: Then we'd also need to flush and memory map them too. What do you think about counting the amount of deleted points and just subtract them? We could simply store this value as header in the memory mapped `.deleted` file.
I think this would be more efficient and cleaner.

lib/segment/src/id_tracker/immutable_id_tracker.rs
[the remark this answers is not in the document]
timvisee: > Do memory maps really copy/cache their entire content into memory?
In this case, yes, because you make a copy of it so it needs to load all of it into memory. These pages may eventually be dropped through.

But, lets keep it like this.

lib/segment/src/id_tracker/immutable_id_tracker.rs
timvisee: I would like that approach better.

We have another case where we do this. There we don't store the count, but count again on start because we may not be able to reliably store it: https://github.com/qdrant/qdrant/blob/8f1d84e40427f8bfd946edfac7a1b03656e37d41/lib/segment/src/vector_storage/dense/appendable_mmap_dense_vector_storage.rs#L77-L78

JojiiOfficial: Oh you're right. I'd still prefer the deleted-count tracker, unless of course you prefer deleting points from the BTreeMap.

JojiiOfficial:
I added more tests and found a few serious bugs, which are also fixed now. 
With the latest changes we have a pretty high test coverage for these changes, as you can see in the screenshots below.

image
image

lib/segment/src/id_tracker/simple_id_tracker.rs
```diff
@@ -190,6 +219,38 @@ impl SimpleIdTracker {
...
+            external_to_internal_num,
+            external_to_internal_uuid,
+            internal_to_external: self.internal_to_external.iter().map(Into::into).collect(),
+        };
+
+        let mut internal_to_version = self.internal_to_version.clone();
+        if internal_to_version.len() < self.internal_to_external.len() {
+            internal_to_version.resize(self.internal_to_external.len(), 0);
+        }
+
```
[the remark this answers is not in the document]
JojiiOfficial: We're comparing `internal_to_version` with `..to_external` here and they can be of different size (at least they were in a test case).
I'm not sure if we want to shrink `.._to_version` here too, if it is larger than `.._to_external`, so I added the check.

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,879 @@
...
+        deleted: &BitSlice,
+        internal_to_version: &[SeqNumberType],
+        mappings: PointMappings,
+    ) -> OperationResult<Self> {
+        // Create mmap file for deleted bitvec
+        let deleted_filepath = Self::deleted_file_path(path);
+        {
+            let deleted_size = bitmap_mmap_size(deleted);
+            create_and_ensure_length(&deleted_filepath, deleted_size)?;
+        }
```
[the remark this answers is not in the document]
JojiiOfficial: `create_and_ensure_length` returns a file-handle. To prevent issues when opening the memory-map again in the next line, I thought flushing/dropping it first would be safer.
It is also scoped here: https://github.com/qdrant/qdrant/blob/master/lib/segment/src/vector_storage/chunked_mmap_vectors.rs#L55
But maybe it is not really needed?

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,879 @@
...
+    pub(crate) external_to_internal_num: BTreeMap<u64, PointOffsetType>,
+    pub(crate) external_to_internal_uuid: BTreeMap<Uuid, PointOffsetType>,
+}
+
+/// Used endianness for storing PointMapping-files.
+type FileEndianess = LittleEndian;
+
+impl PointMappings {
+    const EXTERNAL_ID_NUMBER_BYTE: u8 = 0;
+    const EXTERNAL_ID_UUID_BYTE: u8 = 1;
```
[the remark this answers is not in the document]
JojiiOfficial: As far as I know, we can't properly convert from a number to this enum, without unsafe or a custom `From` implementation that then also is a source of truth. 
What do you think?

lib/segment/src/id_tracker/immutable_id_tracker.rs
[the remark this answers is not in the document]
timvisee: It'll be dropped right away anyway because we don't store the resulting value in a variable.

But feel free to leave it like this to _emphasize_ this a bit.

timvisee (approved):

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,880 @@
...
+        }
+
+        // Check that the file has ben fully read.
+        #[cfg(debug_assertions)] // Only for dev builds
+        {
+            let mut buf = vec![];
+            let read_bytes = reader.read_to_end(&mut buf).unwrap();
+            assert_eq!(buf.len(), 0);
+            assert_eq!(read_bytes, 0);
+        }
```
ffuugoo, line 101: Maybe replace `read_to_end` with [`Read::bytes`] and [`Iterator::count`]? This, most likely, won't ever be a problem in practice, but the fix is trivial.

```suggestion
        // Check that the file has been fully read.
        debug_assert_eq!(
            reader.bytes().map(Result::unwrap).count(),
            0,
        );
```

[`Read::bytes`]: https://doc.rust-lang.org/std/io/trait.Read.html#method.bytes
[`Iterator::count`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,880 @@
...
+        // Deserialize the list entries
+        for i in 0..len {
+            let (internal_id, external_id) = Self::read_entry(&mut reader)?;
+
+            // Need to push this regardless of point deletion as the vecs index represents the internal id
+            // which would become wrong if we leave out entries.
+            internal_to_external.push(external_id);
+
+            let deleted = filter
+                .as_ref()
```
ffuugoo, line 76: `filter` is already `Option<&BitSlice>`, so `as_ref` is not strictly necessary.

```suggestion
```

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,880 @@
...
+        let mut external_to_internal_num: BTreeMap<u64, PointOffsetType> = BTreeMap::new();
+        let mut external_to_internal_uuid: BTreeMap<Uuid, PointOffsetType> = BTreeMap::new();
+
+        // Deserialize the list entries
+        for i in 0..len {
+            let (internal_id, external_id) = Self::read_entry(&mut reader)?;
+
+            // Need to push this regardless of point deletion as the vecs index represents the internal id
+            // which would become wrong if we leave out entries.
+            internal_to_external.push(external_id);
```
ffuugoo, line 73: AFAIK, `SimpleIdTracker` use `u64::MAX` to mark deleted internal-to-external mappings, so we might want to handle this similarly for `ImmutableIdTracker`?

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -0,0 +1,880 @@
...
+        // Create mmap file for deleted bitvec
+        let deleted_filepath = Self::deleted_file_path(path);
+        {
+            let deleted_size = bitmap_mmap_size(deleted);
+            create_and_ensure_length(&deleted_filepath, deleted_size)?;
+        }
+
+        let deleted_mmap = open_write_mmap(&deleted_filepath)?;
+        let mut deleted_new = MmapBitSlice::try_from(deleted_mmap, 0)?;
+        deleted_new[..deleted.len()].copy_from_bitslice(deleted);
```
ffuugoo, line 243: Just `..`?

```suggestion
        deleted_new[..].copy_from_bitslice(deleted);
```

ffuugoo (approved):
Some minor comments, otherwise great job @JojiiOfficial!

generall (approved):

PR #4404, opened 2024-06-06
universal-query: Fix local shard ordering

coszio:
Needs #4380 

- adds an `order` field in `planned_query::ResultsMerge` to indicate the desired direction of merging, and merges accordingly
  - converting from `ShardQueryRequest` to `PlannedQuery` now needs the collection params to determine the required order depending on the vector config
- handles previously unhandled `filter` and `score_threshold` in `LocalShard::merge_prefetches()`, either by passing it to the core_search/scroll rescoring or during regular merging

PR #4405, opened 2024-06-06
universal-query: Exclude referenced ids

coszio:
Needs #4404 

In the Query API, all vector inputs can be ids, for convenience, but we were missing to exclude those ids from the response.


generall (approved):

coszio:
### Merge activity

* **Jun 10, 6:06 PM EDT**: @coszio started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/qdrant/qdrant/4405).

PR #4414, opened 2024-06-07
Dockerfile updates

ffuugoo:
- bump `mold` version to [2.31.0](https://github.com/rui314/mold/releases/tag/v2.31.0)
- move `ARG GIT_COMMIT_ID` down the build step to improve Docker layer caching
- cleanup the Dockerfile a bit
- add `PACKAGES` and `SOURCES` build-args, to allow trivially including debugger into the Docker container
  - this one is arguable for inclusion, but I've been using this for a while, and I'm tired of keeping it in the local branch 🥲

### All Submissions:

### New Feature Submissions:


Dockerfile
```diff
@@ -102,18 +101,45 @@ RUN PKG_CONFIG="/usr/bin/$(xx-info)-pkg-config" \
...
+# Copy Qdrant source files into the container. Useful for debugging.
+#
+# To enable, set `SOURCES` to *any* non-empty string. E.g., 1/true/enable/whatever.
+# (Note, that *any* non-empty string would work, so 0/false/disable would enable the option as well.)
+ARG SOURCES
 
-RUN mkdir -p "$APP"
+# Dockerfile does not support conditional `COPY` instructions (e.g., it's impossible to do something
+# like `if [ -n "$SOURCES" ]; then COPY ...; fi`), so we *hack* conditional `COPY` by abusing
+# parameter expansion and `COPY` wildcards support. 😎
```
ffuugoo, line 126: If source path of a `COPY` instruction uses a wildcard, and the wildcard does not match anything, then `COPY` instruction is silently ignored without producing an error.

If `SOURCES` build-arg is not set (or set to an empty string), then each `${DIR:-/null?}` would expand to `/null?` (which contains `?` wildcard!), and because there's no path matching `/null?` in the `builder` step, each `COPY` command would be silently ignored.

(And, of course, if `SOURCES` arg is set to *anything*, then each `${DIR:-/null?}` expansion would expand to the directory path and be copied properly.)

timvisee (approved):

generall (approved):

PR #4416, opened 2024-06-07
Ensure we have any segment within capacity, otherwise add new one

timvisee:
Extend our optimizer loop to add a new empty segment if all segments are over capacity. The capacity check is based off `max_segment_size_kb` as configured by the user.

Routing writes to segments with capacity is done in the next PR: <https://github.com/qdrant/qdrant/pull/4420>

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/update_handler.rs
+++ b/lib/collection/src/update_handler.rs
@@ -1,6 +1,6 @@
 use std::cmp::min;
 use std::collections::HashSet;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
 
@@ -81,10 +81,6 @@ pub enum OptimizerSignal {
 /// Structure, which holds object, required for processing updates of the collection
 pub struct UpdateHandler {
     shared_storage_config: Arc<SharedStorageConfig>,
-    /// Collection parameters if a new segment needs to be built.
-    pub(super) collection_params: CollectionParams,
-    /// Collection thresholds configuration to determine whether segments are over capacity.
-    pub(super) thresholds_config: OptimizerThresholds,
     /// List of used optimizers
     pub optimizers: Arc<Vec<Arc<Optimizer>>>,
     /// Log of optimizer statuses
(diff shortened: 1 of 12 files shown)


lib/collection/src/update_handler.rs
```diff
@@ -257,6 +260,22 @@ impl UpdateHandler {
...
+                &segments,
+                optimizer.segments_path(),
+                &optimizer.collection_params(),
+                optimizer.threshold_config(),
+            );
+            if let Err(err) = result {
+                log::error!("Failed to ensure there are appendable segments with capacity: {err}");
+                panic!("Failed to ensure there are appendable segments with capacity: {err}");
+            }
+        }
```
timvisee, line 277: Here - in the optimization loop - we now ensure we always have at least one segment with capacity. We check this before running any of the optimizers and their conditions.

timvisee:
Tested on <https://github.com/qdrant/qdrant/pull/4420>, this seems to have quite a significant effect on upload performance.

Though this is an unscientific test, the results are a bit worrying:

- dev branch: 99s (100k op/s)
- PR with very high `max_segment_size`: 104s (96k op/s)
- PR with default `max_segment_size`: 130s (77k op/s)

Tested with:

```bash
bfb -n 10000000 -d 32 --skip-wait-index
```

generall (approved):

lib/collection/src/update_handler.rs
```diff
@@ -376,6 +380,48 @@ impl UpdateHandler {
...
+                .filter_map(|segment_id| segments_read.get(segment_id))
+                .all(|segment| {
+                    let max_vector_size_bytes = segment
+                        .get()
+                        .read()
+                        .max_available_vectors_size_in_bytes()
+                        .unwrap_or_default();
+                    let max_segment_size_bytes = thresholds_config
+                        .max_segment_size_kb
+                        .saturating_add(segment::common::BYTES_IN_KB);
```
ffuugoo, line 409: Is this right? Should it be `saturating_mul`? 🤔

ffuugoo: Would be kinda fun if this fix perf degradation you've seen in your tests... 😂

generall: Nice catch!

timvisee: :clown_face: 

Thanks!

generall: It can explain poor benchmark speed, if segment max size is low, we would create too many segments

ffuugoo (approved):

PR #4417, opened 2024-06-07
Rework `CommitHashRing` consensus message into `CommitRead`/`CommitWrite`/`Finish`

ffuugoo:
Tracked in #4213.

This PR reworks `CommitHashRing` into three separate operations: `CommitRead`, `CommitWrite` and `Finish`.
We need all three to make resharding process seem "transparent" to the user.

We don't (yet) properly track resharding state, so this PR adds some "stubs" and `TODO`s in places where we will introduce state later.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -962,7 +962,11 @@ fn conditions_helper_to_grpc(conditions: Option<Vec<segment::types::Condition>>)
             if conditions.is_empty() {
                 vec![]
             } else {
-                conditions.into_iter().map(|c| c.into()).collect()
+                conditions
+                    .into_iter()
+                    .filter(|c| !matches!(c, segment::types::Condition::Resharding(_))) // TODO(resharding)!?
+                    .map(|c| c.into())
+                    .collect()
             }
         }
     }
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -1060,6 +1060,10 @@ impl From<segment::types::Condition> for Condition {
             segment::types::Condition::Nested(nested) => {
                 ConditionOneOf::Nested(nested.nested.into())
             }
+
+            segment::types::Condition::Resharding(_) => {
+                unimplemented!()
+            }
         };
 
         Self {
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -964,7 +964,7 @@ fn conditions_helper_to_grpc(conditions: Option<Vec<segment::types::Condition>>)
             } else {
                 conditions
                     .into_iter()
-                    .filter(|c| !matches!(c, segment::types::Condition::Resharding(_))) // TODO(resharding)!?
+                    .filter(|c| !c.is_local_only()) // TODO(resharding)!?
                     .map(|c| c.into())
                     .collect()
             }
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -87,8 +87,25 @@ impl Collection {
         Ok(())
     }
 
-    pub async fn commit_hashring(&self, reshard: ReshardKey) -> CollectionResult<()> {
-        self.shards_holder.write().await.commit_hashring(reshard)
+    pub async fn commit_read_hashring(&self, resharding_key: ReshardKey) -> CollectionResult<()> {
+        self.shards_holder
+            .write()
+            .await
+            .commit_read_hashring(resharding_key)
+    }
+
+    pub async fn commit_write_hashring(&self, resharding_key: ReshardKey) -> CollectionResult<()> {
+        self.shards_holder
+            .write()
+            .await
+            .commit_write_hashring(resharding_key)
+    }
+
+    pub async fn finish_resharding(&self, resharding_key: ReshardKey) -> CollectionResult<()> {
+        self.shards_holder
+            .write()
+            .await
+            .finish_resharding(resharding_key)
     }
 
     pub async fn abort_resharding(&self, reshard_key: ReshardKey) -> CollectionResult<()> {
--- a/lib/collection/src/hash_ring.rs
+++ b/lib/collection/src/hash_ring.rs
@@ -76,7 +76,7 @@ impl<T: Hash + Copy + PartialEq> HashRing<T> {
         new.add(shard);
     }
 
-    pub fn commit(&mut self) -> bool {
+    pub fn commit_resharding(&mut self) -> bool {
         let Self::Resharding { new, .. } = self else {
             log::warn!("committing resharding hashring, but hashring is not in resharding mode");
             return false;
--- a/lib/collection/src/hash_ring.rs
+++ b/lib/collection/src/hash_ring.rs
@@ -1,7 +1,9 @@
-use std::fmt;
 use std::hash::Hash;
+use std::{any, fmt};
 
 use itertools::Itertools as _;
+use segment::index::field_index::CardinalityEstimation;
+use segment::types::{PointIdType, ReshardingCondition};
 use smallvec::SmallVec;
 
 use crate::shards::shard::ShardId;
@@ -245,6 +247,61 @@ impl<T: Hash + Copy> Inner<T> {
             Inner::Raw(ring) => ring.is_empty(),
             Inner::Fair { ring, .. } => ring.is_empty(),
         }
+    }
+
+    pub fn len(&self) -> usize {
+        match self {
+            Inner::Raw(ring) => ring.len(),
+            Inner::Fair { ring, scale } => ring.len() / *scale as usize,
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Filter<T = ShardId> {
+    ring: Inner<T>,
+    filter: T,
+}
+
+impl<T> Filter<T> {
+    pub fn new(ring: Inner<T>, filter: T) -> Self {
+        Self { ring, filter }
+    }
+
+    pub fn check(&self, point_id: PointIdType) -> bool
+    where
+        T: Hash + PartialEq + Copy,
+    {
+        self.ring.get(&point_id) == Some(&self.filter)
+    }
+}
+
+impl<T> ReshardingCondition for Filter<T>
+where
+    T: fmt::Debug + Hash + PartialEq + Copy + 'static,
+{
+    fn check(&self, point_id: PointIdType) -> bool {
+        self.check(point_id)
+    }
+
+    fn estimate_cardinality(&self, points: usize) -> CardinalityEstimation {
+        CardinalityEstimation {
+            primary_clauses: vec![],
+            min: 0,
+            exp: points / self.ring.len(),
+            max: points,
+        }
+    }
+
+    fn eq(&self, other: &dyn ReshardingCondition) -> bool {
+        match other.as_any().downcast_ref::<Self>() {
+            Some(other) => self == other,
+            None => false,
+        }
+    }
+
+    fn as_any(&self) -> &dyn any::Any {
+        self
     }
 }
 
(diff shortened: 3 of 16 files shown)


ffuugoo:
I've got somewhat stuck with an alternative idea (to filter *incoming updates*), and I had this PR half-implemented already, so I've decided to push this to a working prototype.

In my simple tests - it works:
- run a cluster (or just a single node in a cluster mode)
- populate collection with `bfb`
- ```
  POST collections/benchmark/cluster
  {
    "start_resharding": {}
  }
  ```
- ```
  POST collections/benchmark/cluster
  {
    "commit_read": {}
  }
  ```
  - added this operation temporarily, strictly for debug purposes
  - also disabled Tim's code that start resharding driver, so that it won't interfere with debugging
- `POST collections/benchmark/points/count`
  - should return less points than expected, cause they are considered to be moved to the new shard, and so filtered from the result

I suggest we take a look at the PR and consider, once again, which of the two approaches to implement.

lib/collection/src/shards/shard_holder.rs
```diff
@@ -323,6 +361,37 @@ impl ShardHolder {
...
+        let Some(ring) = self.rings.get(&state.shard_key) else {
+            return None; // TODO(resharding)!?
+        };
+
+        let HashRing::Resharding { new, .. } = ring else {
+            return None; // TODO(resharding)!?
+        };
+
+        let filter = {
+            let ring = new.clone();
```
timvisee, line 384: I cannot imagine how this would have been without `Clone` being there :face_holding_back_tears:

lib/collection/src/collection/resharding.rs
```diff
@@ -47,47 +47,28 @@ impl Collection {
...
 
-            active_reshard_tasks.add_task(
-                reshard_key,
-                ReshardTaskItem {
-                    task: spawned_task,
-                    started_at: chrono::Utc::now(),
-                    progress,
-                },
-            );
-        }
```
timvisee, line 84: Where did this move to?

ffuugoo: Oops. Temporarily deleted to be able to easily test filtering. I'll revert this. :D

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -308,6 +308,30 @@ impl StructPayloadIndex {
...
+                    .collect();
+
+                let num_ids = ids.len();
+
+                CardinalityEstimation {
+                    primary_clauses: vec![PrimaryCondition::Ids(ids)],
+                    min: num_ids,
+                    exp: num_ids,
+                    max: num_ids,
+                }
```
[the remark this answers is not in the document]
ffuugoo: I kinda copied it from `HasId` branch above, but now that I think about it, this should be a reasonable estimation? (I'd have to refactor a bit so that I can access `hash_ring.len()`, but it's trivial. Just introduce a custom trait.)

```rust
{
	min: 0, 
	max: id_tracker.external_id.len(),
	exp: id_tracker.external_id.len() * (hash_ring.len() - 1) / hash_ring.len(),
}
```

Not sure about `primary_clauses`, though.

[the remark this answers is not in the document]
timvisee: I'd believe so.

Though, this estimation is specific to using a hashring. It does not work for a general purpose 'abstract' filter.

Maybe we can rename it to specialize it for filtering by hashring, while keeping the abstract closure.

lib/collection/src/shards/shard_holder.rs
ffuugoo: Annoying af. 😂 Thanks for pushing `Clone` into upstream!

lib/segment/src/index/struct_payload_index.rs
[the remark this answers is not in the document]
ffuugoo: Though, I'd want @generall's opinion on this.

Initially, I've copied `Condition::Abstract`/`Condition::Resharding` implementation from `Condition::HasId` implementation. I've tracked how `primary_clauses` are used, but it's still a bit over my head. I can't decisively reason if initial (e.g., similar to `HasId`) or current implementation would be better for `Condition::Resharding`.

I can only note that during last stages of resharding, `Condition::Resharding` would be applied to **all** read operations. And it may have to be applied for a while, until we delete "resharded" points from the old shards. Or, at best, until we delete "resharded" points *on a particular node*.

lib/collection/src/hash_ring.rs
```diff
@@ -246,6 +248,54 @@ impl<T: Hash + Copy> Inner<T> {
...
+where
+    T: fmt::Debug + Hash + PartialEq + Copy + 'static,
+{
+    fn check(&self, point_id: PointIdType) -> bool {
+        self.ring.get(&point_id) == Some(&self.filter)
+    }
+
+    fn estimate_cardinality(&self, points: usize) -> CardinalityEstimation {
+        CardinalityEstimation {
+            primary_clauses: vec![],
```
timvisee, line 282: Not sure about this one.

lib/segment/src/problems/unindexed_field.rs
```diff
@@ -323,6 +323,7 @@ impl<'a> Extractor<'a> {
             }
             // No index needed
             Condition::HasId(_) => return,
+            Condition::Resharding(_) => return, // TODO(resharding): 🤔🤔🤔
```
timvisee, line 326: Seems correct

```suggestion
            Condition::Resharding(_) => return,
```

lib/collection/src/shards/shard_holder.rs
```diff
@@ -323,6 +361,30 @@ impl ShardHolder {
...
+            return None;
+        }
+
+        let Some(ring) = self.rings.get(&state.shard_key) else {
+            return None; // TODO(resharding)!?
+        };
+
+        let HashRing::Resharding { new, .. } = ring else {
+            return None; // TODO(resharding)!?
+        };
```
timvisee, line 381: I would say, error, as we don't expect this.

lib/segment/src/index/query_optimization/condition_converter.rs
```diff
@@ -121,6 +121,17 @@ pub fn condition_converter<'a>(
...
+            let external_ids: HashMap<_, _> = id_tracker
+                .iter_internal()
+                .filter_map(|internal_id| Some((internal_id, id_tracker.external_id(internal_id)?)))
+                .collect();
+
+            Box::new(move |point_id| match external_ids.get(&point_id) {
+                Some(&point_id) => cond.check(point_id),
+                None => false,
+            })
+        }
```
timvisee, line 134: How about pre-filtering here? Checking the hashring condition when building the external ID list.

lib/collection/src/hash_ring.rs
ffuugoo: Andrey said, `Resharding` condition does not need one. 🤷‍♀️

Or maybe we can benchmark this.

lib/segment/src/index/query_optimization/condition_converter.rs
ffuugoo: Hmmmmmm. Maybe. Let me think about this for a moment.

lib/collection/src/hash_ring.rs
timvisee: I see your comment on it now. Thanks!

lib/segment/src/index/query_optimization/condition_converter.rs
ffuugoo: Done.

lib/segment/src/problems/unindexed_field.rs
ffuugoo: Done.

lib/collection/src/shards/resharding/mod.rs
```diff
@@ -33,6 +33,7 @@ pub struct ReshardState {
     pub peer_id: PeerId,
     pub shard_id: ShardId,
     pub shard_key: Option<ShardKey>,
+    pub filter_read_operations: bool, // TODO(resharding): Add proper resharding state!
```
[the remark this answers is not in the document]
ffuugoo: We will soon have more state stored in `ReshardState` anyway, so I don't see reason to further duplicate this into `HashRing`.

lib/api/src/grpc/conversions.rs
```diff
@@ -939,7 +939,11 @@ fn conditions_helper_to_grpc(conditions: Option<Vec<segment::types::Condition>>)
             if conditions.is_empty() {
                 vec![]
             } else {
-                conditions.into_iter().map(|c| c.into()).collect()
+                conditions
+                    .into_iter()
+                    .filter(|c| !matches!(c, segment::types::Condition::Resharding(_))) // TODO(resharding)!?
+                    .map(|c| c.into())
+                    .collect()
```
ffuugoo, line 946: A bit hacky, but this was *much* easier, than refactoring all read operations, so that we only propagate `Condition::Resharding` to local read operations, but not to remote read operations.

lib/collection/src/hash_ring.rs
```diff
@@ -246,6 +248,61 @@ impl<T: Hash + Copy> Inner<T> {
...
+    fn eq(&self, other: &dyn ReshardingCondition) -> bool {
+        match other.as_any().downcast_ref::<Self>() {
+            Some(other) => self == other,
+            None => false,
+        }
+    }
+
+    fn as_any(&self) -> &dyn any::Any {
+        self
+    }
```
ffuugoo, line 305: Might be a bit of an overkill, but this is required to implement `PartialEq` on `Condition`.

lib/segment/src/types.rs
```diff
@@ -1947,6 +1950,24 @@ pub enum Condition {
...
+impl PartialEq for Condition {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Field(this), Self::Field(other)) => this == other,
+            (Self::IsEmpty(this), Self::IsEmpty(other)) => this == other,
+            (Self::IsNull(this), Self::IsNull(other)) => this == other,
+            (Self::HasId(this), Self::HasId(other)) => this == other,
+            (Self::Nested(this), Self::Nested(other)) => this == other,
+            (Self::Filter(this), Self::Filter(other)) => this == other,
+            (Self::Resharding(this), Self::Resharding(other)) => this.eq(other.deref()),
```
ffuugoo, line 1967: Without `ReshardingCondition::eq` and `ReshardingCondition::as_any`, we can always return `false` here, but this will kinda break equality for `Condition`.

lib/api/src/grpc/conversions.rs
timvisee: Would a `condition.is_local_only()` make sense here?

lib/collection/src/shards/shard_holder.rs
```diff
@@ -323,6 +361,34 @@ impl ShardHolder {
         Ok(())
     }
 
+    pub fn resharding_filter(&self) -> Option<Filter> {
```
timvisee, line 364: How about something like this, to clarify the semantics a bit?

```suggestion
    pub fn resharding_filter_exclude_migrated(&self) -> Option<Filter> {
```

timvisee (approved):
Though I don't see any pressing issues, I do se a lot of TODO's. What do we do about those?

ffuugoo:
> If we merge all of the we get a growing number of TODOs which we'll never resolve. I'd suggest as many as possible right away, or remove them if it's something not strictly necessary.

All marked with `TODO(resharding)`, we can remove them later, when finalizing resharding.

lib/collection/src/shards/shard_holder.rs
ffuugoo: Dunno. Does not make much difference, IMO. Probably better to add documentation to the method.

timvisee:
```suggestion
    /// Create a filter that excludes points migrated to a different shard, as part of resharding.
    pub fn resharding_filter(&self) -> Option<Filter> {
```

timvisee (approved):

PR #4420, opened 2024-06-10
When selecting a segment for writing, select one that has capacity

timvisee:
Builds on top of: <https://github.com/qdrant/qdrant/pull/4416>

Direct writes to segments with capacity.

When selecting an appendable segment for writing, consider the capacity and prefer segments that still have capacity.

Checking capacity requires us to read lock segments. In this implementation I tried to find the right balance between correctness and performance. For example, when selecting an appendable segment for writing it may be over capacity if the segments having capacity currently could not be read-locked. Nor do we always return the segment with minimum capacity because that would require us to read lock every segment all the time. I tried to keep our existing structures mostly intact.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/collection_ops.rs
+++ b/lib/collection/src/collection/collection_ops.rs
@@ -229,10 +229,11 @@ impl Collection {
     /// Partially blocking. Stopping existing optimizers is blocking. Starting new optimizers is
     /// not blocking.
     pub async fn recreate_optimizers_blocking(&self) -> CollectionResult<()> {
+        let effective_config = self.effective_optimizers_config().await?;
         let shard_holder = self.shards_holder.read().await;
         let updates = shard_holder
             .all_shards()
-            .map(|replica_set| replica_set.on_optimizer_config_update());
+            .map(|replica_set| replica_set.on_optimizer_config_update(effective_config.clone()));
         future::try_join_all(updates).await?;
         Ok(())
     }
--- a/lib/collection/src/collection_manager/fixtures.rs
+++ b/lib/collection/src/collection_manager/fixtures.rs
@@ -33,6 +33,22 @@ pub fn empty_segment(path: &Path) -> Segment {
     build_simple_segment(path, 4, Distance::Dot).unwrap()
 }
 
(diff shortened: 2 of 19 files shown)


generall (changes requested):
I think this PR is a bit more complicated than it needs to be.

I don't think SegmentHolder should know about `OptimizerThresholds`, instead it would be enough if `random_appendable_segment_with_capacity` would simply select the smallest one, even if it means we need to iterate over all of them. I am pretty sure the overhead is going to be minimal.

`aloha_random_write` In my optimion, is not required to have this logic at all. In theory it is nice to have, but as a prise we have non-trivial function with potential deadlocks doubled in size. It is only used if the segments are indexed, so not related to the case we are trying to optimize.

timvisee:
> I think this PR is a bit more complicated than it needs to be.
> 
> I don't think SegmentHolder should know about `OptimizerThresholds`, instead it would be enough if `random_appendable_segment_with_capacity` would simply select the smallest one, even if it means we need to iterate over all of them. I am pretty sure the overhead is going to be minimal.
> 
> `aloha_random_write` In my optimion, is not required to have this logic at all. In theory it is nice to have, but as a prise we have non-trivial function with potential deadlocks doubled in size. It is only used if the segments are indexed, so not related to the case we are trying to optimize.

You know what, I fully agree on it being more complicated than it needs to be. I like that the suggested approach is a lot simpler. And as I understand it it still covers the case we're actually looking for.

Now, you mention trying  to lock all segments. Since the effect of this PR is already noticeable in a simple upsertion benchmark I'm not sure I agree it is so minimal. But I'll give it a proper go anyway. We can give it a test if we like.

I'll implement your suggestion in a separate PR so that we can compare the two. This one is already quite noisy anyway.

Thanks for the suggestion and insight!

timvisee:
I've implemented the alternative here: <https://github.com/qdrant/qdrant/pull/4440>

timvisee:
Superseded by <https://github.com/qdrant/qdrant/pull/4440>

PR #4424, opened 2024-06-10
Fix multivector simple storage metadata

agourlay:
This PR fixes a bug in the multivector simple storage where the metadata recorded was incorrect.

This patch also introduces a series of improvements created during the investigation:
- improve unit test with larger inner vector sizes
- remove `update_buffer` mechanism
- clarify naming `vector_count` vs `vector_len`
- return option in `get_many` to play nice with data consistency check

### DIFF
--- a/lib/segment/src/data_types/vectors.rs
+++ b/lib/segment/src/data_types/vectors.rs
@@ -263,11 +263,7 @@ impl<T: PrimitiveVectorElement> TypedMultiDenseVector<T> {
         self.flattened_vectors.is_empty()
     }
 
-    pub fn len(&self) -> usize {
-        self.flattened_vectors.len()
-    }
-
-    pub fn vector_count(&self) -> usize {
+    pub fn vectors_count(&self) -> usize {
         self.flattened_vectors.len() / self.dim
     }
 }
@@ -315,7 +311,7 @@ impl<'a, T: PrimitiveVectorElement> TypedMultiDenseVectorRef<'a, T> {
         self.flattened_vectors.is_empty()
     }
 
-    pub fn vector_count(self) -> usize {
+    pub fn vectors_count(self) -> usize {
         self.flattened_vectors.len() / self.dim
     }
 
--- a/lib/segment/src/vector_storage/multi_dense/simple_multi_dense_vector_storage.rs
+++ b/lib/segment/src/vector_storage/multi_dense/simple_multi_dense_vector_storage.rs
@@ -134,19 +134,19 @@ fn open_simple_multi_dense_vector_storage_impl<T: PrimitiveVectorElement>(
         }
 
         let metadata = &mut vectors_metadata[point_id_usize];
-        metadata.inner_vectors_count = stored_record.vector.vector_count();
+        metadata.inner_vectors_count = stored_record.vector.vectors_count();
         metadata.inner_vector_capacity = metadata.inner_vectors_count;
         metadata.id = point_id;
 
         metadata.start = vectors.len() as PointOffsetType;
         let left_keys = vectors.get_chunk_left_keys(metadata.start);
-        if stored_record.vector.vector_count() > left_keys {
+        if stored_record.vector.vectors_count() > left_keys {
             metadata.start += left_keys as PointOffsetType;
         }
         vectors.insert_many(
             metadata.start,
             &stored_record.vector.flattened_vectors,
-            stored_record.vector.vector_count(),
+            stored_record.vector.vectors_count(),
         )?;
 
         check_process_stopped(stopped)?;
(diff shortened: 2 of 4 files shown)


lib/segment/src/vector_storage/multi_dense/simple_multi_dense_vector_storage.rs
```diff
@@ -134,19 +134,19 @@ fn open_simple_multi_dense_vector_storage_impl<T: PrimitiveVectorElement>(
         }
 
         let metadata = &mut vectors_metadata[point_id_usize];
-        metadata.size = stored_record.vector.flattened_vectors.len();
-        metadata.capacity = metadata.size;
+        metadata.inner_vectors_count = stored_record.vector.vector_count();
```
agourlay, line 137: the bug was at this line, assigning the flatten len as a size for the vector count

lib/segment/src/data_types/vectors.rs
```diff
@@ -264,6 +264,10 @@ impl<T: PrimitiveVectorElement> TypedMultiDenseVector<T> {
     }
 
     pub fn len(&self) -> usize {
```
IvanPleshkov, line 266: Do we actually need this `len`? I propose to get flattened vectors len from flattened vectors

lib/segment/src/data_types/vectors.rs
```diff
@@ -311,7 +315,7 @@ impl<'a, T: PrimitiveVectorElement> TypedMultiDenseVectorRef<'a, T> {
         self.flattened_vectors.is_empty()
     }
 
-    pub fn len(self) -> usize {
+    pub fn vector_count(self) -> usize {
```
IvanPleshkov, line 318:
```suggestion
    pub fn vectors_count(self) -> usize {
```

lib/segment/src/data_types/vectors.rs
```diff
@@ -264,6 +264,10 @@ impl<T: PrimitiveVectorElement> TypedMultiDenseVector<T> {
     }
 
     pub fn len(&self) -> usize {
+        self.flattened_vectors.len()
+    }
+
+    pub fn vector_count(&self) -> usize {
```
IvanPleshkov, line 270:
```suggestion
    pub fn vectors_count(&self) -> usize {
```

lib/segment/src/vector_storage/multi_dense/simple_multi_dense_vector_storage.rs
```diff
@@ -192,9 +188,10 @@ impl<T: PrimitiveVectorElement> SimpleMultiDenseVectorStorage<T> {
         deleted: bool,
         vector: Option<TypedMultiDenseVectorRef<T>>,
     ) -> OperationResult<()> {
-        // Write vector state to buffer record
-        let record = &mut self.update_buffer;
-        record.deleted = deleted;
+        let mut record = StoredMultiDenseVector {
```
IvanPleshkov, line 191: If temp buffer doesn't affect performance, I agree to remove it

lib/segment/src/vector_storage/multi_dense/simple_multi_dense_vector_storage.rs
```diff
@@ -271,18 +268,22 @@ impl<T: PrimitiveVectorElement> MultiVectorStorage<T> for SimpleMultiDenseVector
...
     fn get_multi_opt(&self, key: PointOffsetType) -> Option<TypedMultiDenseVectorRef<T>> {
-        self.vectors_metadata
-            .get(key as usize)
-            .map(|metadata| TypedMultiDenseVectorRef {
-                flattened_vectors: self.vectors.get_many(metadata.start, metadata.size),
+        self.vectors_metadata.get(key as usize).map(|metadata| {
+            let flattened_vectors = self
+                .vectors
+                .get_many(metadata.start, metadata.inner_vectors_count)
+                .unwrap_or_else(|| panic!("Vectors does not contain data for {:?}", metadata));
```
IvanPleshkov, line 275: do we need this panic here? I guess we can return None here

IvanPleshkov (approved):

lib/segment/src/vector_storage/multi_dense/simple_multi_dense_vector_storage.rs
agourlay: My reasoning here is that `vectors_metadata` and `vectors` are inconsistent if the key exists in `metadata` but NOT in `vectors`.
Therefore this is a broken invariant.

WDYT?

IvanPleshkov: No concerns

PR #4427, opened 2024-06-10
universal-query: Propagate filters to leaf queries

coszio:
This PR changes the planned query conversion to propagate every filter from the root of the query to each of the subqueries.

The motivation is so that every condition is already satisfied when merging multiple prefetches, and prevent post-filtering behavior.


agourlay:
Is it possible to have some basic tests here to make sure it still works?

coszio:
@agourlay Tests updated to reflect this change 👍

agourlay (approved):

PR #4428, opened 2024-06-10
better default max segment size estimation

generall:
Since we potentially use less CPUs for indexing, default max segment size should reflect that


timvisee (approved):
Thanks!

PR #4442, opened 2024-06-11
Hnsw update entry point after linking

IvanPleshkov:
This PR is a subset of changes from https://github.com/qdrant/qdrant/pull/4412.

It's a minimal change to fix flacky `test_graph_connectivity`.

There is a bug while HNSW construction. We change entry points before linking. In multithreaded environment is cause such problem: point may take entry with no links.

The solution is updating entry points after linking.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
@@ -308,7 +308,7 @@ impl GraphLayersBuilder {
         let entry_point_opt = self
             .entry_points
             .lock()
-            .new_point(point_id, level, |point_id| {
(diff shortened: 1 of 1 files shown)


timvisee:
Will this finally fix [917 failures](https://github.com/qdrant/qdrant/issues/2875)? 🥹
 
image

IvanPleshkov:
> Will this finally fix [917 failures](https://github.com/qdrant/qdrant/issues/2875)? 🥹
> 
> image

Right. It's a minimal required change from https://github.com/qdrant/qdrant/pull/4412 to fix flacky test

timvisee (approved):

generall (approved):

PR #4453, opened 2024-06-12
Integrate compressed posting list (attempt 2)

xzfc:
This PR integrates new compressed posting lists into `VectorIndexEnum`.
It doesn't affect the actual behavior (non-compressed index is still used non-conditionally), so it should be safe to include this into a patch release.

### DIFF
--- a/lib/segment/src/index/sparse_index/sparse_vector_index.rs
+++ b/lib/segment/src/index/sparse_index/sparse_vector_index.rs
@@ -16,7 +16,6 @@ use sparse::common::sparse_vector::SparseVector;
 use sparse::common::types::DimId;
 use sparse::index::inverted_index::inverted_index_ram_builder::InvertedIndexBuilder;
 use sparse::index::inverted_index::{InvertedIndex, INDEX_FILE_NAME, OLD_INDEX_FILE_NAME};
-use sparse::index::migrate::SparseVectorIndexVersion;
 use sparse::index::search_context::SearchContext;
 
 use super::indices_tracker::IndicesTracker;
@@ -132,7 +131,7 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
                 // version file. To distinguish between index in original format and partially
                 // written index in the current format, the index file name is changed from
                 // `inverted_index.data` to `inverted_index.dat`.
-                SparseVectorIndexVersion::save(path)?;
+                TInvertedIndex::Version::save(path)?;
 
                 if vector_storage.borrow().total_vector_count() != 0 {
                     log::info!("Successfully rebuilt");
@@ -161,22 +160,23 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
     fn try_load(
         path: &Path,
     ) -> OperationResult<(SparseIndexConfig, TInvertedIndex, IndicesTracker)> {
-        let mut stored_version = SparseVectorIndexVersion::load(path)?;
+        let mut stored_version = TInvertedIndex::Version::load(path)?;
 
         // Simple migration mechanism for 0.1.0.
-        // TODO: Drop this code on the next version bump.
         let old_path = path.join(OLD_INDEX_FILE_NAME);
-        assert_eq!(SparseVectorIndexVersion::current(), Version::new(0, 1, 0));
-        if stored_version.is_none() && old_path.exists() {
+        if TInvertedIndex::Version::current() == Version::new(0, 1, 0)
+            && stored_version.is_none()
+            && old_path.exists()
+        {
             rename(old_path, path.join(INDEX_FILE_NAME))?;
-            SparseVectorIndexVersion::save(path)?;
-            stored_version = Some(SparseVectorIndexVersion::current());
+            TInvertedIndex::Version::save(path)?;
+            stored_version = Some(TInvertedIndex::Version::current());
         }
 
-        if stored_version != Some(SparseVectorIndexVersion::current()) {
+        if stored_version != Some(TInvertedIndex::Version::current()) {
             return Err(OperationError::service_error(format!(
                 "Index version mismatch, expected {}, found {}",
-                SparseVectorIndexVersion::current(),
+                TInvertedIndex::Version::current(),
                 stored_version.map_or_else(|| "none".to_string(), |v| v.to_string()),
             )));
         }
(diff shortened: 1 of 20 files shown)


lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -333,40 +333,14 @@ pub(crate) fn create_vector_index(
...
-        SparseIndexType::MutableRam => VectorIndexEnum::SparseRam(SparseVectorIndex::open(
-            sparse_vector_config.index,
-            id_tracker.clone(),
-            vector_storage.clone(),
-            payload_index.clone(),
-            vector_index_path,
-            stopped,
-        )?),
+    let vector_index = match args.config.index_type {
+        SparseIndexType::MutableRam => VectorIndexEnum::SparseRam(SparseVectorIndex::open(args)?),
```
IvanPleshkov, line 339: Are we going to create a compressed index later? Do we want to keep both compressed and original indices in the future?

lib/sparse/src/index/inverted_index/inverted_index_ram.rs
```diff
@@ -3,13 +3,22 @@ use std::cmp::max;
...
 use crate::common::types::DimId;
 use crate::index::inverted_index::InvertedIndex;
 use crate::index::posting_list::{PostingList, PostingListIterator};
 use crate::index::posting_list_common::PostingElementEx;
 
+pub struct Version;
+
+impl StorageVersion for Version {
+    fn current_raw() -> &'static str {
+        panic!("InvertedIndexRam is not supposed to be versioned");
```
IvanPleshkov, line 18: Just a nitpick to use `unreachable` instead of `panic`

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -161,22 +174,23 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
 
         // Simple migration mechanism for 0.1.0.
-        // TODO: Drop this code on the next version bump.
         let old_path = path.join(OLD_INDEX_FILE_NAME);
-        assert_eq!(SparseVectorIndexVersion::current(), Version::new(0, 1, 0));
-        if stored_version.is_none() && old_path.exists() {
+        if TInvertedIndex::Version::current() == Version::new(0, 1, 0)
+            && stored_version.is_none()
+            && old_path.exists()
+        {
```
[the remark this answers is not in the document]
xzfc: Recently we did introduce a version file for sparse index. This file should be written last to indicate that the write is complete. But how to distinguish between the following cases?
1. Data were written by an old Qdrant release. (`A` in the diagram)
2. A crash occurred during index file write. (`D` in the diagram)

Both have index data and no `version.info`, but only in the second case we need to rebuild from scratch.

```mermaid
flowchart LR
    A["A (Qdrant ≤v1.9)\nno version.info\ninverted_index.data"]
    B["B (current, 1.9.x)\nversion.info=0.1.0\ninverted_index.dat"]
    C["C (compressed, 1.10+)\nversion_info=0.2.0\ninverted_index.dat"]
    D["D (Partially written B or C)\nno version.info\ninverted_index.dat"]

    A -->|simple rename| B
    B -->|rebuild| C
    A -->|rebuild| C

    D -->|rebuild| B
    D -->|rebuild| C
```

Rename allows us to distinguish between `A` and `D`, and to perform migration from `A` to `B` without a full rebuild.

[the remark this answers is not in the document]
IvanPleshkov: Excellent! I got the idea and found the solution great. I would ask only to add this explanation as a comment into the code

lib/segment/src/segment_constructor/segment_constructor_base.rs
xzfc: Plan to keep those for v1.10:
- MutableRam: uncompressed since updating compressed index is costly; and `f32` only. (tho maybe we'll add f16/u8 if it would be worthy)
- Compressed immutable: Ram/Mmap; f32/f16/u8. (so, 6 combinations total)

lib/sparse/src/index/inverted_index/inverted_index_compressed_mmap.rs
```diff
@@ -29,14 +30,22 @@ use crate::index::posting_list_common::GenericPostingElement;
 const POSTING_HEADER_SIZE: usize = size_of::<PostingListFileHeader>();
 const INDEX_CONFIG_FILE_NAME: &str = "inverted_index_config.json";
 
+pub struct Version;
+
+impl StorageVersion for Version {
+    fn current_raw() -> &'static str {
+        "0.2.0"
```
timvisee, line 37: Do we manually define a version number here on purpose?

xzfc: Yes. There are two versions defined in this PR: `0.1.0` for the current uncompressed format and `0.2.0` for the new.

Previously I've used the crate version as the format version (like how it is done in other parts of the codebase). But since we're going to merge support for two format versions in a single crate, it's not convenient. Assuming that there is a single version is messy and blocks me from further PRs on this feature.

lib/segment/src/segment_constructor/segment_constructor_base.rs
IvanPleshkov: Got it, you did it as a next step
https://github.com/qdrant/qdrant/pull/4454/files#diff-e1db48b48e08d3c200e7f47f4080abb0316f015bdb75aedfa865e63bfaf64289:~:text=lib/segment/src/segment_constructor/segment_constructor_base.rs

lib/sparse/src/index/inverted_index/inverted_index_compressed_mmap.rs
IvanPleshkov: As I remember from @xzfc explanation, he cannot use the crate version as a version here because he has to define both current and previous and the crate allows to set only one version. Am I right, @xzfc ?

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -79,16 +79,27 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
     }
 }
 
+pub struct OpenArgs<'a> {
```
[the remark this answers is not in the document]
xzfc: Renamed to `SparseVectorIndexOpenArgs`.

As for the builder, I'd prefer to keep it as simple as it is now. I've introduced this structure to reduce the amount of boilerplate (see `create_sparse_vector_index` in this PR and in #4454); and as a poor's man named arguments for a function. Builder pattern won't help here.

lib/sparse/src/index/inverted_index/inverted_index_compressed_mmap.rs
IvanPleshkov: We can set initial version as any value. Can we set the last version the same as the crate version? And hardcode only the obsolete one?

IvanPleshkov (approved):

PR #4454, opened 2024-06-12
Add datatype option for sparse vector index

xzfc:
Depends on #4453.

This PR adds an option `datatype` with possible values of `float16` and `float32`.
It's hidden in the OpenAPI docs for now, and is not usable until `USE_COMPRESSED` is false (see below).

The switch `pub const USE_COMPRESSED: bool = false;` is added as a feature gate until the v1.10 release. When it's false, the old index format is used.

### DIFF
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -1324,7 +1324,6 @@ pub struct SparseIndexParams {
     pub on_disk: Option<bool>,
     /// Datatype used to store weights in the index.
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    #[schemars(skip) /* TODO(1.10): remove this in the next minor release to expose the compressed sparse index */]
     pub datatype: Option<Datatype>,
 }
 
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -24,6 +24,7 @@ use segment::data_types::groups::GroupId;
 use segment::data_types::vectors::{
     DenseVector, QueryVector, VectorRef, VectorStruct, DEFAULT_VECTOR_NAME,
 };
+use segment::index::sparse_index::sparse_index_config::SparseVectorIndexDatatype;
 use segment::types::{
     Distance, Filter, MultiVectorConfig, Payload, PayloadIndexInfo, PayloadKeyType, PointIdType,
     QuantizationConfig, SearchParams, SeqNumberType, ShardKey, VectorStorageDatatype,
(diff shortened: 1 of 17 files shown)


lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -39,6 +39,10 @@ use crate::vector_storage::{
     check_deleted_condition, new_stoppable_raw_scorer, VectorStorage, VectorStorageEnum,
 };
 
+// TODO(compressed_sparse_index): remove this once feature is released
+/// Whether to use the new compressed format.
+pub const USE_COMPRESSED: bool = false;
```
timvisee, line 44: Not strictly necessary, but maybe we can use this format. We use it in other places as well.

```suggestion
// TODO(1.10): remove this in the next minor release to expose the compressed sparse index
/// Whether to use the new compressed format.
pub const USE_COMPRESSED: bool = false;
```

lib/segment/src/index/sparse_index/sparse_index_config.rs
```diff
@@ -49,22 +67,33 @@ pub struct SparseIndexConfig {
     pub full_scan_threshold: Option<usize>,
     /// Type of sparse index
     pub index_type: SparseIndexType,
+    /// Datatype used to store weights in the index.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "SparseVectorIndexDatatype::should_hide")]
+    #[schemars(skip)] // TODO(compressed_sparse_index): remove this once feature is released
```
timvisee, line 73:
```suggestion
    #[schemars(skip)] // TODO(1.10): remove this once feature is released
```

lib/segment/src/index/sparse_index/sparse_index_config.rs
```diff
@@ -39,8 +40,25 @@ impl SparseIndexType {
...
 
+/// Storage types for vectors
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Copy, Clone, PartialEq, Eq, Default)]
+pub enum SparseVectorIndexDatatype {
+    // Single-precision floating point
+    #[default]
+    Float32,
+    // Half-precision floating point
+    Float16,
+}
```
timvisee, line 51: Do we want [`#[serde(rename_all = "snake_case")]`](https://serde.rs/container-attrs.html#rename_all) here?

lib/collection/src/operations/types.rs
```diff
@@ -1366,16 +1386,22 @@ impl SparseIndexParams {
         SparseIndexParams {
             full_scan_threshold,
             on_disk,
+            datatype: None, // TODO
```
timvisee, line 1389: Shall we propagate it here from parameters?

xzfc: Fixed. (this method was not used, so I've deleted it at all)

generall (approved):

PR #4457, opened 2024-06-12
universal-query: limit max depth for prefetches

agourlay:
This PR adds a validation for the depth of the prefetches of universal queries.

This is a protection mechanism to avoid processing very large nested queries as it could be very expensive and potentially render the server irresponsive.


coszio (approved):
Good tests!

PR #4459, opened 2024-06-12
universal-query: Execute batch queries in `LocalShard`

coszio:
Edits the `do_planned_query` function into a `do_planned_query_batch`, to generalize into a handling batch queries.

Most of the function is the same (doing searches and scrolls), but introduces extra complexity when fetching payload and vectors. 

This complexity comes from the fact that points may be deleted between search/scroll and after merging, so we now ignore those points which are not there anymore.

### DIFF
--- a/lib/collection/src/shards/local_shard/query.rs
+++ b/lib/collection/src/shards/local_shard/query.rs
@@ -132,6 +132,19 @@ impl LocalShard {
                 .map(|scored_point| scored_point.id)
                 .collect::<Vec<PointIdType>>();
 
+            // Assert that the flattened list is in the same order as iterating over each of the inner lists.
+            debug_assert!({
+                let mut point_ids_iter = point_ids.iter();
+                query_response.iter().all(|intermediate| {
+                    intermediate.iter().all(|point| {
+                        point_ids_iter
+                            .next()
+                            .map(|id| *id == point.id)
+                            .unwrap_or(false)
+                    })
+                })
+            });
+
             let mut records_iter = SegmentsSearcher::retrieve(
                 self.segments(),
                 &point_ids,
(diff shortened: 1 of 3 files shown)


lib/collection/src/shards/local_shard/query.rs
```diff
@@ -75,45 +75,100 @@ impl LocalShard {
...
+
+            query_response
+                .into_iter()
+                .map(|intermediate| {
+                    intermediate
+                        .into_iter()
+                        // Points might get deleted between search and retrieve.
+                        // But it's not a problem, because we just filter them out.
+                        .filter_map(|mut scored_point| {
+                            match records_iter.peeking_next(|record| record.id == scored_point.id) {
```
coszio, line 152: This part assumes that the flattened iterator over ids above, has the same order as:
1. The response of `SegmentsSearcher::retrieve` (excluding those points which were not found)
2. Iterating over each intermediate response without flattening.

agourlay: Can we have a debug_assert! for detecting when the order is different?

timvisee: And maybe also if the records iterator contains a point we will not be covering anymore? Or does that not make sense?

If that would happen, this logic would break.

timvisee (approved):
LGTM once the above review comment is resolved.

coszio:
### Merge activity

* **Jun 14, 11:47 AM EDT**: @coszio started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/qdrant/qdrant/4459).

generall:
This PR needs refactoring, it degrades performance compared to the current implementation, as the filters are not reused in batch

PR #4462, opened 2024-06-13
Build perf with up to 256 codegen units to make building fast

timvisee:
The `perf` profile is supposed to build fast. However, since it now inherits 1 codegen unit from the release profile this is not the case anymore.

This sets the number of codegen units on the profile back to the default, making builds significantly faster.

On my machine:
- without this patch, building takes 412 seconds
- with this patch, building takes 187 seconds

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -197,6 +197,7 @@ codegen-units = 256 # restore default value for faster compilation
 inherits = "release"
 lto = false
 opt-level = 3
+codegen-units = 256 # restore default value for faster compilation
 
 [patch.crates-io]
 # Temporary patch until <https://github.com/hyperium/tonic/pull/1401> is merged


agourlay:
`profile.bench` and `profile.perf` seem to be equivalent now, can we just use one?

timvisee:
> `profile.bench` and `profile.perf` seem to be equivalent now, can we just use one?

I would vote to set `codegen-units = 1` for `bench`. I believe our bench profile is supposed to be fast at runtime, in which case that would make sense.

What do you think?

agourlay:
I did the change for setting `codegen-units = 256` in `profile.bench` in first place because it was taking me way too long to iterate on performance improvements.

I will just approve this PR so that you do not waste more time.

agourlay (approved):

timvisee:
> I did the change for setting `codegen-units = 256` in `profile.bench` in first place because it was taking me way too long to iterate on performance improvements.
> 
> I will just approve this PR so that you do not waste more time.

But isn't that what `perf` is for, and not `bench`?

agourlay:
I was iterating on Criterion benches at the time so I optimized that profile for quicker turn around.
Are you saying I can use a different profile for Criterion benchmarks?

ffuugoo:
TIL about codegen units 😅

ffuugoo (approved):

timvisee:
> I think @timvisee's point is that [`codegen-units = 1` may produce more optimized code](https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units), which we probably prefer when running benchmarks.

Correct.

However, there's a balance here: getting faster code means pulling your hair out when benchmarking due to long build times.

In a call we decided to keep both with the current configuration in this PR. We consider bench a high number of codegen units to be good enough. And even though they're similar now, we also keep perf because that's just what we're used to while testing.

PR #4463, opened 2024-06-13
universal-query: API testing with multi named vectors

agourlay:
This PR introduces test for the new query API using several named vectors of different types:
- dense
- sparse
- multivector

The goal is to provide a larger safety net as we start to change the internals significantly.

### DIFF
--- a/tests/openapi/openapi_integration/test_query_full.py
+++ b/tests/openapi/openapi_integration/test_query_full.py
@@ -458,3 +458,46 @@ def test_rrf():
         assert expected["id"] == result["id"]
         assert expected["payload"] == result["payload"]
         assert isclose(expected["score"], result["score"], rel_tol=1e-5)
+
+    # test with inner filters instead of root filter
+    response = request_with_validation(
+        api="/collections/{collection_name}/points/query",
+        method="POST",
+        path_params={"collection_name": collection_name},
+        body={
+            "prefetch": [
+                {
+                    "query": [0.1, 0.2, 0.3, 0.4],
+                    "using": "dense-image",
+                    "filter": filter,
+                },
+                {
+                    "query": [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8],
+                    "using": "dense-text",
+                    "filter": filter,
+                },
+                {
+                    "query": {
+                        "indices": [63, 65, 66],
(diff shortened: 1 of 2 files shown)


tests/openapi/openapi_integration/test_query_full.py
```diff
@@ -0,0 +1,551 @@
...
+        method="POST",
+        path_params={"collection_name": collection_name},
+        body={
+            "order_by": "count",
+        },
+    )
+    assert response.ok
+    scroll_result = response.json()["result"]["points"]
+
+    query_result = root_and_rescored_query({"order_by": "count"}, "dense-image", with_payload=True)
```
coszio, line 348: `"dense-image"`?

tests/openapi/openapi_integration/test_query_full.py
```diff
@@ -0,0 +1,551 @@
...
+    search_result_3 = response.json()["result"]
+
+    response = request_with_validation(
+        api="/collections/{collection_name}/points/search",
+        method="POST",
+        path_params={"collection_name": collection_name},
+        body={
+            "vector": {
+                "name": "dense-multi",
+                "vector": [3.05, 3.61, 3.76, 3.74], # legacy API expands single vector to multiple vectors
```
coszio, line 424: nice check 👍

tests/openapi/openapi_integration/test_query_full.py
```diff
@@ -0,0 +1,551 @@
...
+            "query": {"fusion": "rrf"}
+        },
+    )
+    assert response.ok, response.json()
+    rrf_result = response.json()["result"]
+
+    def get_id(x):
+        return x["id"]
+
+    # rrf order is not deterministic with same scores, so we need to sort by id
```
coszio, line 472: Non-related to this PR: We could make it deterministic by using `ScoredPointTies`, but do we want to?

coszio (approved):
Thanks for the safety net 🥅

tests/openapi/openapi_integration/test_query_full.py
coszio: I meant specifying "using" for order by

agourlay: Good catch, forgot that one :face_with_head_bandage: 
Will fix with the next batch of tests :)

PR #4464, opened 2024-06-13
check f16 neon in runtime

IvanPleshkov:
Check neon presense in runtime and remove unused cfg feature and non-simd code pass


xzfc (approved):

timvisee (commented):
Happy to see `std::arch::is_aarch64_feature_detected!("fp16")` is a runtime thing :smile:

PR #4471, opened 2024-06-14
universal-query: cluster test for query API

agourlay:
Tests for the query API in a cluster setup to make sure the inter shards communication works properly.

### DIFF
--- a/tests/consensus_tests/test_points_query.py
+++ b/tests/consensus_tests/test_points_query.py
@@ -213,7 +213,39 @@ def test_points_query(tmp_path: pathlib.Path):
                 },
                 "limit": 5,
             })
+        ),
+        # request filter & source filters
+        (
+            ("query",{
+                "prefetch": [
+                    {
+                        "query": [0.2, 0.1, 0.9, 0.7],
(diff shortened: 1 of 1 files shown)


coszio (changes requested):
Good testing! I would love to also see two things more:
1. Setting read_consistency
2. Order_by in asc and desc

agourlay:
failing for scroll & query equivalence
```
scroll
{'filter': {'must_not': [{'key': 'city', 'match': {'value': 'Berlin'}}]}, 'limit': 5, 'order_by': 'count', 'direction': 'asc'}
[5, 9, 10, 6, 7]
VS
query
{'filter': {'must_not': [{'key': 'city', 'match': {'value': 'Berlin'}}]}, 'limit': 5, 'order_by': 'count', 'direction': 'asc'}
[4, 5, 6, 7, 8]
```

agourlay:
@coszio I have removed the broken scroll tests for the time being.

I propose to merge this to increase our safety net and then uncomment later the test cases when introducing a fix :)

coszio (approved):

PR #4476, opened 2024-06-16
universal-search: local-shard refactoring

generall:
- rename some stuff here and there
- attempt to remove unnecessary abstractions


generall:
ToDo: Remove `PlannedQueryBatch` and `WeakPlannedQuery` and extend `PlannedQuery` instead

agourlay (approved):

lib/collection/src/operations/universal_query/planned_query.rs
```diff
@@ -72,10 +74,16 @@ pub struct MergePlan {
...
-impl TryFrom<ShardQueryRequest> for PlannedQuery {
-    type Error = CollectionError;
+impl PlannedQuery {
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self {
+            root_plans: Vec::with_capacity(capacity),
+            searches: Vec::with_capacity(capacity),
+            scrolls: Vec::with_capacity(capacity),
+        }
+    }
```
generall, line 84: It is unlikely that all 3 are going to have same capacity, isn't it?

generall: I propose to just init empty, don't think it is critical here

coszio: 🤷 micro-optimization, I am fine either way

coszio:
I just fixed a bug where we would ignore the `with_payload` and `with_vector` params when dealing with a fusion request.

This was introduced because we are not fetching these as a last step anymore. To fix it I decided it would be the responsibility of the `PlannedQuery` to decide when to fetch it. This allowed further simplification in `do_planned_query`

agourlay (approved):

PR #4477, opened 2024-06-17
Fix building cancellation flaky test

IvanPleshkov:
Next attempt to fix build cancellation flaky test.

Issue https://github.com/qdrant/qdrant/issues/2723

In this PR we don't have a magic number of required building time. The building time of the non-canceled segment is not a subject of this test.


timvisee (approved):

timvisee:
@IvanPleshkov you drafted it, but it seems ready for merge. What do you think?

PR #4478, opened 2024-06-17
Max size of multivector to fix chunk in storage

IvanPleshkov:
This PR adds max size limit for multivector.

Problem: we cannot use unlimited size of multivector because vector storage is divided into chunks and multivector may be larger that vector storage shunk.

Solution: add validation to multivector, add additional checks in vector storages, and cover it by test.

### DIFF
--- a/lib/common/common/src/validation.rs
+++ b/lib/common/common/src/validation.rs
@@ -3,6 +3,9 @@ use std::borrow::Cow;
 use serde::Serialize;
 use validator::{Validate, ValidationError, ValidationErrors};
 
+// Multivector sholud be small enough to fir the chunk of vector storage
+pub const MAX_MULTIVECTOR_FLATTENED_LEN: usize = 1_000_000;
+
 #[allow(clippy::manual_try_fold)] // `try_fold` can't be used because it shortcuts on Err
 pub fn validate_iter<T: Validate>(iter: impl Iterator<Item = T>) -> Result<(), ValidationErrors> {
     let errors = iter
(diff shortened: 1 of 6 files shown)


lib/common/common/src/validation.rs
```diff
@@ -3,6 +3,9 @@ use std::borrow::Cow;
 use serde::Serialize;
 use validator::{Validate, ValidationError, ValidationErrors};
 
+// Multivector sholud be small enough to fir the chunk of vector storage
+pub const MAX_MULTIVECTOR_FLATTENED_LEN: usize = 1_000_000;
```
generall, line 7:
```suggestion
pub const MAX_MULTIVECTOR_FLATTENED_LEN: usize = 1024 * 1024;
```

generall: Also, `_LEN` means it will be the same for bytes / half-floats / floats?

generall: Also, this should be documented

IvanPleshkov: We validate at the point where we don't know vector storage type. So I have chosen the worse option - `f32` and set 1M elements

generall: Why not set the value to be in bytes?

generall (approved):

lib/common/common/src/validation.rs
```diff
@@ -3,6 +3,9 @@ use std::borrow::Cow;
 use serde::Serialize;
 use validator::{Validate, ValidationError, ValidationErrors};
 
+// Multivector sholud be small enough to fir the chunk of vector storage
```
timvisee, line 6:
```suggestion
// Multivector should be small enough to fit the chunk of vector storage
```

timvisee (approved):

lib/common/common/src/validation.rs
timvisee: I see I'm late to the party. I'll submit a quick PR to patch this.

timvisee: https://github.com/qdrant/qdrant/pull/4481

PR #4479, opened 2024-06-17
universal-query: add lookups to query API definition

agourlay:
This PR adds the `lookup_from` field on the query API both at the root and prefetch level.

To that end, a number of types have been lifted up from the `collection` crate into the `api` crate.
It is necessary because the query definition lives in the `api` crate which does not have access to the lookup types in `collection`.

I also lifted the types for grouping which will be required later.

A future PR will add the actual implementation for `lookup_from` for queries.

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -11417,11 +11417,12 @@
               }
             ]
           },
-          "with_lookup": {
-            "description": "Options for specifying where to lookup vector ids from during recommendation",
+          "lookup_from": {
+            "description": "The location used to lookup vectors. If not specified - use current collection. Note: the other collection should have the same vector size as the current collection",
+            "default": null,
             "anyOf": [
               {
-                "$ref": "#/components/schemas/WithLookupInterface"
+                "$ref": "#/components/schemas/LookupLocation"
               },
               {
                 "nullable": true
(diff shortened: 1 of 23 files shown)


lib/api/src/rest/schema.rs
```diff
@@ -350,3 +360,135 @@ impl ContextPair {
...
+
+    /// Options for specifying which vectors to include (or not)
+    #[serde(alias = "with_vector")]
+    #[serde(default)]
+    pub with_vectors: Option<WithVector>,
+}
+
+const fn default_with_payload() -> Option<WithPayloadInterface> {
+    Some(WithPayloadInterface::Bool(true))
+}
```
coszio, line 382: Does it have to be at the same level of all other structs?

docs/redoc/master/openapi.json
```diff
@@ -11416,6 +11416,18 @@
                 "nullable": true
               }
             ]
+          },
+          "lookup_from": {
+            "description": "The location used to lookup vectors. If not specified - use current collection. Note: the other collection should have the same vector size as the current collection",
```
coszio, line 11421: More specifically, same as the `using` vector

coszio (commented):
LGTM. Just to recap:
- moved to `api` crate: 
  - `SearchRequestInternal`
  - `WithLookupInterface`
  - `WithLookup`
  - `LookupLocation`
  - `BaseGroupRequest`
  - `SearchGroupsRequestInternal`

The only thing I'd request to change is the description for the `lookup_from` parameter to link it to the `using` param

agourlay:
@coszio thanks for the nice recap, I should provided it in first place :memo: 

Regarding improving the doc between `lookup_from` and `using` I pushed the following https://github.com/qdrant/qdrant/pull/4479/commits/b4cc220679b8c09133115f018d7dc23a2d735633

coszio (approved):

PR #4480, opened 2024-06-17
Properly integrate replica set state switching in resharding transfers

timvisee:

In <https://github.com/qdrant/qdrant/pull/4373> I implemented a first iteration of point migration. In practice, transferring still appeared to be broken because a lot more had to be done for replica set state switching and shard routing.

This PR does all that needs to be done in terms of shard transfers to make resharding work. Specifically, to actually migrate points from all shards to the new target shard, and to replicate the new shard matching the configured replication factor.

Most importantly, this does:
- support transfers with a different source/target shard (for resharding)
- extend the unique shard key with a target shard field (for resharding)
- resharding transfer must merge points in target shard, it should not replace them
- the resharding driver state must be aware of the peers we know about
- use correct replica set state switching for staring and finishing resharding transfers

What doesn't work yet and will be resolved in a separate PR:
- migrate points within the same node

I've extensively tested this locally with various different scenarios, such as different shard numbers, replication factors and target nodes. All works well.

We cannot (automatically) test this yet, because the consensus operation for finalizing resharding is missing. Once that's there, I'll add an integration test to assert correct behavior.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -369,7 +369,9 @@ impl Collection {
                 let shard_transfer_registered = shards_holder.shard_transfers.wait_for(
                     |shard_transfers| {
                         shard_transfers.iter().any(|shard_transfer| {
-                            shard_transfer.shard_id == shard_id && shard_transfer.to == this_peer_id
+                            (shard_transfer.shard_id == shard_id
+                                || shard_transfer.to_shard_id == Some(shard_id))
+                                && shard_transfer.to == this_peer_id
                         })
                     },
                     Duration::from_secs(60),
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -211,10 +211,14 @@ impl Collection {
         // Unwrap forward proxy into local shard, or replace it with remote shard
         // depending on the `sync` flag.
         if self.this_peer_id == transfer.from {
+            let activate_shard = transfer
+                .method
+                .map_or(true, |method| !method.is_resharding());
             let proxy_promoted = transfer::driver::handle_transferred_shard_proxy(
                 shard_holder,
                 transfer.shard_id,
                 transfer.to,
+                activate_shard,
                 transfer.sync,
             )
             .await?;
@@ -223,7 +227,11 @@ impl Collection {
 
         // Should happen on receiving side
         // Promote partial shard to active shard
-        if self.this_peer_id == transfer.to {
+        if self.this_peer_id == transfer.to
+            && !transfer
+                .method
+                .map_or(false, |method| method.is_resharding())
+        {
             let shard_promoted =
                 transfer::driver::finalize_partial_shard(shard_holder, transfer.shard_id).await?;
             log::debug!(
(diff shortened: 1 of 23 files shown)


lib/collection/src/collection/shard_transfer.rs
```diff
@@ -57,27 +57,28 @@ impl Collection {
...
+                .to_shard_id
+                .unwrap_or(shard_transfer.shard_id);
+
             let shards_holder = self.shards_holder.read().await;
+            let from_replica_set = shards_holder.get_shard(&from_shard_id).ok_or_else(|| {
+                CollectionError::service_error(format!("Shard {from_shard_id} doesn't exist"))
+            })?;
+            let to_replica_set = shards_holder.get_shard(&to_shard_id).ok_or_else(|| {
+                CollectionError::service_error(format!("Shard {to_shard_id} doesn't exist"))
+            })?;
```
timvisee, line 77: Because we now have transfers with a different source and target shard ID (in case resharding), we need a separate source and target replica set in our shard transfer logic. This theme will repeat throughout this PR in various places.

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -809,17 +809,17 @@ impl ShardReplicaSet {
...
-                Some(peer_id)
-            } else {
-                None
-            }
+        // List the peers in active or resharding state
+        // If resharding, there will be only one shard in resharding state and we should not
+        // consider all to be dead
+        // TODO(resharding): accept resharding state as active like below?
+        let active_or_resharding_peers = peers.iter().filter_map(|(&peer_id, &state)| {
+            matches!(state, ReplicaState::Active | ReplicaState::Resharding).then_some(peer_id)
```
timvisee, line 817: When resharding, we only have one shard in the `Resharding` state, and none in `Active` state.

This prevents messing with our locally disabled peers list. I'm not too happy with it though, maybe there's a better place where we can prevent this behavior.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -340,7 +369,9 @@ impl Collection {
                 let shard_transfer_registered = shards_holder.shard_transfers.wait_for(
                     |shard_transfers| {
                         shard_transfers.iter().any(|shard_transfer| {
-                            shard_transfer.shard_id == shard_id && shard_transfer.to == this_peer_id
+                            (shard_transfer.shard_id == shard_id
+                                || shard_transfer.to_shard_id == Some(shard_id))
+                                && shard_transfer.to == this_peer_id
```
ffuugoo, line 374: I think we also have to explicitly check for `st.to_shard_id.is_none()`, e.g.:

```rust
(st.shard_id == shard_id && st.to_shard_id.is_none()) || st.to_shard_id == Some(shard_id)
```

Cleaned-up suggestion:

```suggestion
                            let to_this_shard = match shard_transfer.to_shard_id {
                                Some(to_shard_id) => to_shard_id == shard_id,
                                None => shard_transfer.shard_id == shard_id,
                            };

                            let to_this_peer = shard_transfer.to == this_peer_id;

                            to_this_shard && to_this_peer
```

ffuugoo (approved):

PR #4482, opened 2024-06-17
Fix multivector for unnamed vectors

generall:
currently we can't upload / return multi-vectors if we are using unnamed vectors in Rest API.

This PR is supposed to fix this


generall:
ToDo:

- check conversions from GRPC
- apply same for BatchVectorStruct
- Use reference in `new_unchecked` of Multivector

IvanPleshkov (approved):

PR #4491, opened 2024-06-18
While resharding, support migrating points on a node locally

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/4480>

Allow resharding migration transfers within the same node. It's a special case where the source and target shard are on the same machine, while our shard transfer mechanism doesn't support this.

This scenario must be supported if there's only one shard replica, exclusively on the node having the new target shard.

It's an initial approach, still relying on some building blocks of shard transfers and remote shards. It still talks over gRPC channels, but to itself. Eventually I'd like to improve upon this to migrate points from one shard replica to another directly instead of relying on transfers. But that requires significant refactoring as various components, such as our proxy shards, don't support this. I propose to attempt that later.

Note that this PR also refactors the point migration function overall. The idea is that it better shows the possible code paths.

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/4480>
- [x] Rebase on `dev`
- [x] Undraft PR

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -351,6 +351,17 @@ impl ShardReplicaSet {
         self.replica_state.read().get_peer_state(peer_id).copied()
     }
 
+    /// List the peer IDs on which this shard is active, both the local and remote peers.
+    pub async fn active_shards(&self) -> Vec<PeerId> {
+        let replica_state = self.replica_state.read();
+        replica_state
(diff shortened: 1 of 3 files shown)


timvisee:
@ffuugoo In today's daily we agreed on keeping it like this for a while, and to not focus on trying to make it proper local because it's just not worth the effort.

ffuugoo (approved):

PR #4495, opened 2024-06-18
universal-query: gRPC query API

agourlay:
This PR adds a new gRPC endpoint for the query API.

### DIFF
--- a/lib/api/build.rs
+++ b/lib/api/build.rs
@@ -298,10 +298,8 @@ fn configure_validation(builder: Builder) -> Builder {
             ("CountPointsInternal.count_points", ""),
             ("SyncPointsInternal.sync_points", ""),
             ("SyncPoints.collection_name", "length(min = 1, max = 255)"),
-            ("QueryPointsInternal.collection_name", "length(min = 1, max = 255)"),
-            ("QueryPointsInternal.filter", ""),
-            ("QueryPointsInternal.params", ""),
-            ("QueryPointsInternal.timeout", "custom = \"crate::grpc::validate::validate_u64_range_min_1\""),
+            ("QueryBatchPointsInternal.collection_name", "length(min = 1, max = 255)"),
+            ("QueryBatchPointsInternal.timeout", "custom = \"crate::grpc::validate::validate_u64_range_min_1\""),
         ], &[])
         // Service: raft_service.proto
         .validates(&[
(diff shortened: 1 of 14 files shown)


lib/api/build.rs
```diff
@@ -259,6 +259,10 @@ fn configure_validation(builder: Builder) -> Builder {
             ("DatetimeRange.gt", "custom = \"crate::grpc::validate::validate_timestamp\""),
             ("DatetimeRange.lte", "custom = \"crate::grpc::validate::validate_timestamp\""),
             ("DatetimeRange.gte", "custom = \"crate::grpc::validate::validate_timestamp\""),
+            ("QueryPoints.collection_name", "length(min = 1, max = 255)"),
+            ("QueryPoints.filter", ""),
+            ("QueryPoints.params", ""),
+            ("QueryPoints.timeout", "custom = \"crate::grpc::validate::validate_u64_range_min_1\""),
```
coszio, line 265: Let's validate limit > 1 too

lib/api/build.rs
```diff
@@ -294,6 +298,10 @@ fn configure_validation(builder: Builder) -> Builder {
             ("CountPointsInternal.count_points", ""),
             ("SyncPointsInternal.sync_points", ""),
             ("SyncPoints.collection_name", "length(min = 1, max = 255)"),
+            ("QueryPointsInternal.collection_name", "length(min = 1, max = 255)"),
+            ("QueryPointsInternal.filter", ""),
+            ("QueryPointsInternal.params", ""),
+            ("QueryPointsInternal.timeout", "custom = \"crate::grpc::validate::validate_u64_range_min_1\""),
```
coszio, line 304: This will be substituted with batch version

lib/api/src/grpc/proto/points_service.proto
```diff
@@ -109,4 +109,8 @@ service Points {
   Perform multiple update operations in one request
   */
   rpc UpdateBatch (UpdateBatchPoints) returns (UpdateBatchResponse) {}
+  /*
+  Universal query interface for search, query, scroll and rescoring.
```
coszio, line 113: Let's agree on a single description in rest and grpc

lib/api/src/grpc/proto/points_service.proto
```diff
@@ -109,4 +109,8 @@ service Points {
...
+  */
+  rpc Query (QueryPoints) returns (SearchResponse) {}
```
coszio, line 115: We want to extend the response, so a unique type would be better

tests/integration-tests.sh
```diff
@@ -57,3 +57,5 @@ fi
 ./tests/basic_sparse_grpc_test.sh
 
 ./tests/basic_multivector_grpc_test.sh
+
+./tests/basic_query_basic_grpc_test.sh
```
coszio, line 61: add trailing line :)

coszio (approved):
LGTM

PR #4500, opened 2024-06-19
universal-query: Fix query order_by tests

agourlay:
This PR enables tests for the query API using `order_by` that were broken.

Actually, there was no bug on the server side but the tests were not using the query API properly :face_with_head_bandage:

### DIFF
--- a/tests/consensus_tests/test_points_query.py
+++ b/tests/consensus_tests/test_points_query.py
@@ -262,49 +262,56 @@ def test_points_query(tmp_path: pathlib.Path):
                 "score_threshold": 0.5
             })
         ),
-        # TODO(universal-query) uncomment when order_by aggregation is fixed
-        # (
-        #     # scroll
-        #     ("scroll", "points.id", {
-        #         "filter": filter,
-        #         "limit": 5,
-        #     }),
-        #     ("query", "id", {
-        #         "filter": filter,
-        #         "limit": 5,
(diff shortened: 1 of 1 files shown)


tests/consensus_tests/test_points_query.py
```diff
@@ -262,49 +262,56 @@ def test_points_query(tmp_path: pathlib.Path):
...
+                "filter": filter,
+                "limit": 5,
+                "order_by": "count",
+                "direction": "asc",
+            }),
+            ("query", "id", {
+                "filter": filter,
+                "limit": 5,
+                "query": {
+                    "order_by": {
```
agourlay, line 289: example here of proper usage of order_by through the query field

generall (approved):

tests/consensus_tests/test_points_query.py
```diff
@@ -262,49 +262,56 @@ def test_points_query(tmp_path: pathlib.Path):
...
+                "with_payload": True,
+            }),
+        ),
+        (
+            # scroll order by `asc`
+            ("scroll", "points.id", {
+                "filter": filter,
+                "limit": 5,
+                "order_by": "count",
+                "direction": "asc",
```
coszio, line 283: This and below is still incorrect
```suggestion
                "order_by": {
                    "key: "count",
                    "direction": "asc",
                }
```

PR #4501, opened 2024-06-19
Propagate deletes for resharding, delete migrated points

timvisee:

Implement the 5th resharding stage: delete migrated points from their old shards.

This is a naive implementation, but it is functional. We can improve upon this later.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/resharding/driver.rs
+++ b/lib/collection/src/shards/resharding/driver.rs
@@ -635,7 +635,7 @@ async fn stage_propagate_deletes(
                 ))
             })?;
 
-            // Take batch of pionts, if full, pop the last entry as next batch offset
+            // Take batch of points, if full, pop the last entry as next batch offset
             let mut batch = replica_set
                 .scroll_by(
                     offset,
--- a/lib/collection/src/shards/resharding/driver.rs
+++ b/lib/collection/src/shards/resharding/driver.rs
@@ -625,10 +625,11 @@ async fn stage_propagate_deletes(
     };
 
     while let Some(source_shard_id) = block_in_place(|| state.read().shards_to_delete().next()) {
-        let mut offset = None;
+        let mut offset = Some(0.into());
 
-        while {
+        while offset.is_some() {
             let shard_holder = shard_holder.read().await;
+
             let replica_set = shard_holder.get_shard(&source_shard_id).ok_or_else(|| {
                 CollectionError::service_error(format!(
                     "Shard {source_shard_id} not found in the shard holder for resharding",
@@ -636,7 +637,7 @@ async fn stage_propagate_deletes(
             })?;
 
             // Take batch of points, if full, pop the last entry as next batch offset
-            let mut batch = replica_set
+            let mut points = replica_set
                 .scroll_by(
                     offset,
                     DELETE_BATCH_SIZE + 1,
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/resharding/driver.rs
```diff
@@ -514,16 +530,92 @@ fn stage_commit_hashring() -> CollectionResult<()> {
...
+                .map(|record| record.id)
+                .collect::<Vec<_>>();
+            let operation =
+                CollectionUpdateOperations::PointOperation(PointOperations::DeletePoints { ids });
+            replica_set
+                .update_with_consistency(operation, offset.is_some(), WriteOrdering::Weak)
+                .await?;
+
+            offset.is_some()
+        } {}
```
ffuugoo, line 607: This loop is cursed. 😂

Why not either:
```rust
let mut spin = true;

while spin {
  // ...

  spin = offset.is_some();
}
```

Or:
```rust
loop {
  // ...

  if offset.is_none() {
    break;
  }
}
```

timvisee: It's a do-while loop, this is how Rust imagined it :smile: 

Fine with both your alternatives though!

lib/collection/src/shards/resharding/driver.rs
```diff
@@ -514,16 +530,92 @@ fn stage_commit_hashring() -> CollectionResult<()> {
...
+                    &false.into(),
+                    &false.into(),
+                    // TODO(resharding): directly apply hash ring filter here
+                    None,
+                    None,
+                    false,
+                    None,
+                )
+                .await?;
+            offset = (batch.len() >= DELETE_BATCH_SIZE).then(|| batch.pop().unwrap().id);
```
ffuugoo, line 593: I think this should be `>` instead of `>=`:

```rust
batch.len() > DELETE_BATCH_SIZE
```

(Not critical though. I'll fix this along with while loop fix.)

lib/collection/src/shards/resharding/driver.rs
ffuugoo: Fixed.

lib/collection/src/shards/resharding/driver.rs
ffuugoo: Fixed.

ffuugoo:
@timvisee I've rebased on `dev` and pushed my fixes for the loop. Uncursed `while` condition and changed `>=` to `>` in the batch length check.

lib/collection/src/shards/resharding/driver.rs
timvisee: Good catch. Thanks!

lib/collection/src/shards/resharding/driver.rs
```diff
@@ -625,18 +625,19 @@ async fn stage_propagate_deletes(
     };
 
     while let Some(source_shard_id) = block_in_place(|| state.read().shards_to_delete().next()) {
-        let mut offset = None;
+        let mut offset = Some(0.into());
```
timvisee, line 628: I think having `None` is better, because we might also have UUIDs. UUIDs should be ordered after integers, but still.

I'll change it to a loop with an explicit break check at the end if you don't mind.

ffuugoo: Ok.

PR #4507, opened 2024-06-19
Commit read/write hashrings while resharding

timvisee:
Tracked in: #4213 

Implement the hashring commit stage in the resharding driver.

Commit the read and write hashrings separately and sync the cluster in both cases after doing so.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/mod.rs
+++ b/lib/collection/src/shards/mod.rs
@@ -24,7 +24,11 @@ mod test;
 
 use std::path::{Path, PathBuf};
 
-use shard::ShardId;
+use channel_service::ChannelService;
+use common::defaults;
+use shard::{PeerId, ShardId};
+use tokio::time::{sleep_until, timeout_at};
+use transfer::ShardTransferConsensus;
 
 use crate::operations::types::{CollectionError, CollectionResult};
 use crate::shards::shard_versioning::versioned_shard_path;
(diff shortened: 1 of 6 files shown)


lib/collection/src/shards/mod.rs
```diff
@@ -52,3 +56,38 @@ pub async fn create_shard_dir(
...
+            sleep_until(wait_until).await;
+        }
+        // Reached timeout, assume consensus is synchronized
+        Err(err) => {
+            log::warn!(
+                "All peers failed to synchronize consensus, continuing after timeout: {err}"
+            );
+        }
+    }
+}
```
timvisee, line 93: We had exactly this already, I just moved it here so we can use it in resharding.

lib/collection/src/shards/resharding/driver.rs
```diff
@@ -559,15 +568,41 @@ async fn stage_replicate(
...
+    consensus
+        .commit_read_hashring_confirm_and_retry(collection_id, reshard_key)
+        .await?;
+    await_consensus_sync(consensus, channel_service).await;
+
+    // Commit write hashring and sync cluster
+    consensus
+        .commit_write_hashring_confirm_and_retry(collection_id, reshard_key)
+        .await?;
+    await_consensus_sync(consensus, channel_service).await;
```
timvisee, line 599: The important part is here. Here we commit the read and write hashring in two separate steps. In both cases we sync all peers in the cluster right after to ensure everybody reached the same state.

The other changes in this PR is mostly to route the consensus messages and to confirm/retry.

ffuugoo (approved):

PR #4508, opened 2024-06-19
universal-query: lookup_from implementation

agourlay:
This PR implements the logic for the `lookup_from` field on the query API.

Most of the complexity comes from resolving ids from different collections.

This change also makes sure `lookup_from` behave according to the user token.

### DIFF
--- a/lib/collection/src/common/retrieve_request_trait.rs
+++ b/lib/collection/src/common/retrieve_request_trait.rs
@@ -105,7 +105,7 @@ impl RetrieveRequest for DiscoverRequestInternal {
     }
 }
 
-impl RetrieveRequest for CollectionQueryResolveRequest {
+impl<'a> RetrieveRequest for CollectionQueryResolveRequest<'a> {
     fn get_lookup_collection(&self) -> Option<&String> {
         self.lookup_from.as_ref().map(|x| &x.collection)
     }
--- a/lib/collection/src/common/retrieve_request_trait.rs
+++ b/lib/collection/src/common/retrieve_request_trait.rs
@@ -4,7 +4,8 @@ use segment::types::PointIdType;
 
 use crate::operations::types::{DiscoverRequestInternal, RecommendRequestInternal, UsingVector};
 use crate::operations::universal_query::collection_query::{
-    self, CollectionPrefetch, CollectionQueryRequest, VectorInput, VectorQuery,
+    self, CollectionPrefetch, CollectionQueryRequest, CollectionQueryResolveRequest, VectorInput,
+    VectorQuery,
 };
 
 const EMPTY_SHARD_KEY_SELECTOR: Option<ShardKeySelector> = None;
@@ -105,9 +106,11 @@ impl RetrieveRequest for DiscoverRequestInternal {
     }
 }
 
+/// Disclaimer: the lookup info corresponds ONLY to the root query, not the prefetches.
+/// Whereas the `get_referenced_point_ids` method will return the point ids of the root query AND all its prefetches.
 impl RetrieveRequest for CollectionQueryRequest {
     fn get_lookup_collection(&self) -> Option<&String> {
-        None // TODO(universal-query): Change this when we add lookup_from to CollectionQueryRequest
+        self.lookup_from.as_ref().map(|x| &x.collection)
     }
 
     fn get_referenced_point_ids(&self) -> Vec<PointIdType> {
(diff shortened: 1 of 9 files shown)


tests/openapi/openapi_integration/test_query_full.py
```diff
@@ -217,39 +220,367 @@ def test_recommend_avg():
...
+            "limit": 10,
+            "using": "dense-image",
+            "lookup_from": {
+                "collection": lookup_collection_name,
+                "vector": "other"
+            }
+        },
+    )
+    assert not response.ok, response.text
+    assert response.json()["status"]["error"] == "Not found: No point with id 3 found"
```
coszio, line 309: Is it hard to include the collection name in the error?

lib/collection/src/operations/universal_query/collection_query.rs
```diff
@@ -46,6 +47,15 @@ impl CollectionQueryRequest {
     pub const DEFAULT_WITH_PAYLOAD: WithPayloadInterface = WithPayloadInterface::Bool(false);
 }
 
+/// Lightweight representation of a query request to implement the [RetrieveRequest] trait.
+#[derive(Debug)]
+pub struct CollectionQueryResolveRequest {
+    pub vector_query: VectorQuery<VectorInput>,
```
coszio, line 53: nice

coszio: Is it possible to make this a reference so that we don't have to clone raw vectors?

lib/collection/src/common/retrieve_request_trait.rs
```diff
@@ -105,9 +106,11 @@ impl RetrieveRequest for DiscoverRequestInternal {
     }
 }
 
+/// Disclaimer: the lookup info corresponds ONLY to the root query, not the prefetches.
+/// Whereas the `get_referenced_point_ids` method will return the point ids of the root query AND all its prefetches.
 impl RetrieveRequest for CollectionQueryRequest {
```
coszio, line 111: I believe we can remove this implementation now :)

tests/openapi/openapi_integration/test_query_full.py
```diff
@@ -217,39 +220,367 @@ def test_recommend_avg():
...
+                    "using": "dense-image",
+                    "lookup_from": {
+                        "collection": lookup_collection_name,
+                        "vector": "other"
+                    }
+                }
+            ],
+            "limit": 10,
+            "using": "dense-image",
+            "query": {"fusion": "rrf"}
```
coszio, line 559: let's have a validation error for "using" + fusion

lib/collection/src/operations/universal_query/collection_query.rs
```diff
@@ -302,8 +314,8 @@ impl CollectionPrefetch {
             .map(|query| {
                 query.try_into_scoring_query(
                     ids_to_vectors,
-                    lookup_vector_name,
-                    lookup_collection,
+                    parent_lookup_vector_name,
+                    parent_lookup_collection,
```
coszio, line 318: We shouldn't propagate "using" into the prefetches. They should only work on their level

coszio (changes requested):
See above comments 😄

lib/collection/src/operations/universal_query/collection_query.rs
```diff
@@ -533,6 +544,16 @@ impl CollectionQueryRequest {
                 _ => {}
             }
         }
+
+        // Check that fusion queries are not combined with a using vector name
+        if let Some(Query::Fusion(_)) = query {
+            if using != DEFAULT_VECTOR_NAME {
+                return Err(CollectionError::bad_request(
+                    "Fusion queries can only be used with the default vector name.",
```
coszio, line 525: Better error message?

lib/collection/src/operations/universal_query/collection_query.rs
```diff
@@ -296,6 +296,16 @@ impl CollectionPrefetch {
...
 
+    fn get_lookup_vector_name(&self) -> String {
+        match &self.lookup_from {
+            None => self.using.to_owned(),
+            Some(lookup_from) => match &lookup_from.vector {
+                None => self.using.to_owned(),
+                Some(vector_name) => vector_name.clone(),
+            },
+        }
+    }
```
coszio, line 307: good candidate for `Option::and_then`

coszio (changes requested):
Nice!
Let's just fix the error message https://github.com/qdrant/qdrant/pull/4508#discussion_r1653060829

coszio (approved):

PR #4510, opened 2024-06-19
Fix wrong consistency results ordering

coszio:
One assumption we were making when using `kmerge`, is that the results provided by each resolved shard were already ordered appropriately. But it turns out this is not the case for when we actually resolve consistency for a replica set. 

A mistake we were making is to sort the resolved consistency results always in ascending order, without considering if the order should be descending.

To fix this, in this PR I have:
- stopped sorting when resolving replica results
- replaced usages of `kmerge` with `sorted_unstable_by`

This will most likely have a slight perf hit, but at least results will be correct.

An additional thing this also fixes is that, at collection, we were deduping `order_by` results by id only, while multivalue payloads can make the same id appear as many times as the amount of values it has.

### Future work
In a later PR I want to introduce changes so that we communicate the desired order to the replica resolver, so that we can keep using `kmerge` or similar in the subsequent layers.

### DIFF
--- a/lib/collection/src/collection/point_ops.rs
+++ b/lib/collection/src/collection/point_ops.rs
@@ -293,7 +293,7 @@ impl Collection {
             Some(order_by) => {
                 retrieved_iter
                     // Extract and remove order value from payload
-                    .map(|records| {
+                    .flat_map(|records| {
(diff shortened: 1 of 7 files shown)


coszio:
Closing in favor of #4511

PR #4511, opened 2024-06-20
fix: Communicate ordering to replica resolver

coszio:
Supersedes #4510 

Fixes same issue as #4510, but now this is what I consider a proper fix. Now the resolver knows which way to order.

### DIFF
--- a/lib/collection/src/collection/search.rs
+++ b/lib/collection/src/collection/search.rs
@@ -268,11 +268,16 @@ impl Collection {
 
             let results_from_shards = all_searches_res
                 .iter_mut()
-                .map(|res| mem::take(&mut res[batch_index]));
+                .map(|res| mem::take(&mut res[batch_index]))
+                .flatten();
 
             let merged_iter = match order {
-                Order::LargeBetter => Either::Left(results_from_shards.kmerge_by(|a, b| a > b)),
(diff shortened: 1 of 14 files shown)


timvisee (approved):

lib/collection/src/collection/search.rs
```diff
@@ -243,36 +244,33 @@ impl Collection {
...
-            let results_from_shards = all_searches_res
-                .iter_mut()
-                .map(|res| mem::take(&mut res[batch_index]));
+            let results = results_from_shards.into_iter().map(|(res, _)| res);
 
             let merged_iter = match order {
-                Order::LargeBetter => Either::Left(results_from_shards.kmerge_by(|a, b| a > b)),
-                Order::SmallBetter => Either::Right(results_from_shards.kmerge_by(|a, b| a < b)),
+                Order::LargeBetter => Either::Left(results.kmerge_by(|a, b| a > b)),
+                Order::SmallBetter => Either::Right(results.kmerge_by(|a, b| a > b)),
```
generall, line 273: Is the change of comparision correct?

coszio: It is incorrect, thanks

coszio:
FYI: @generall and I just discussed an alternative approach which does not require communicating ordering

coszio:
Closing in favor of #4688

PR #4516, opened 2024-06-20
Remove clock tags when forwarding operations during reshading shard transfer

ffuugoo:
Resharding shard transfer initializes `ForwardProxy` that forwards incoming updates from local shard `X` to remote shard `Y`.

Our clock tag logic does not work well enough in such case (though, it still _kinda_ works!), so we just remove clock tags during `ForwardProxy::update`, if `wrapped_shard` and `remote_shard` IDs are different.

**TODO (for a follow-up PR):**

When `ForwardProxy` forwards updates from one shard to *another* one, it should also "split" the operation with a hashring and only forward relevant *part* of the operation to the remote shard. (Or, alternatively, *remote* shard should split incoming operations and only apply relevant part.)

It seems a lot more annoying to implement, though, so I'll leave it for a separate PR.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/forward_proxy_shard.rs
+++ b/lib/collection/src/shards/forward_proxy_shard.rs
@@ -12,6 +12,7 @@ use segment::types::{
 use tokio::runtime::Handle;
 use tokio::sync::Mutex;
 
+use super::shard::ShardId;
 use super::update_tracker::UpdateTracker;
 use crate::hash_ring::HashRing;
 use crate::operations::point_ops::{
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -207,6 +210,10 @@ impl ShardOperation for ForwardProxyShard {
         // the transfer needs to have access to the latest version of points.
         let mut result = self.wrapped_shard.update(operation.clone(), true).await?;
 
+        if self.shard_id != self.remote_shard.id {
+            operation.clock_tag = None;
+        }
```
timvisee, line 215:
```suggestion
        // Strip clock tag if forwarding to a different shard
        // Each shard has their own clock tags and they are incompatible with other shards
        if self.shard_id != self.remote_shard.id {
            operation.clock_tag = None;
        }
```

timvisee (approved):

PR #4517, opened 2024-06-20
Fix web UI index path http security headers

Rendez:
The web-ui `/dashboard` endpoint can currently be embedded as an iframe. Adding the header `X-Frame-Options: DENY` shall protect against it.

In terms of the approach taken with actix_web, it would have been ideal if `actix_files::Files` allowed to set HTTP headers, but it only allows a handful of flags like `etag`, `last_modified` or `content_disposition`, which fall short of our needs.

This is why for the root path reading the `index.html` file a custom handler has been introduced: `web_ui_index`.

Awaiting for feedback in the approach before writing any kind of tests.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/actix/mod.rs
+++ b/src/actix/mod.rs
@@ -12,8 +12,10 @@ use std::sync::Arc;
 
 use ::api::grpc::models::{ApiResponse, ApiStatus, VersionInfo};
 use actix_cors::Cors;
+use actix_files::NamedFile;
 use actix_multipart::form::tempfile::TempFileConfig;
 use actix_multipart::form::MultipartFormConfig;
+use actix_web::http::header::{self, HeaderValue};
 use actix_web::middleware::{Compress, Condition, Logger};
 use actix_web::{error, get, web, App, HttpRequest, HttpResponse, HttpServer, Responder};
 use actix_web_extras::middleware::Condition as ConditionEx;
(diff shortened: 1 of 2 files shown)


Rendez:
I'd like to move the file read to be async:

```rust
async fn web_ui_index(
    req: HttpRequest,
    web_ui_settings: web::Data<WebUISettings>,
) -> impl Responder {
    let path = Path::new(&web_ui_settings.static_folder).join("index.html");

    match NamedFile::open_async(path).await {
        Ok(file) => {
            let mut res = file.respond_to(&req);
            res.headers_mut()
                .insert(header::X_FRAME_OPTIONS, HeaderValue::from_static("DENY"));

            res
        }
        Err(err) => HttpResponse::from_error(err),
    }
}
```

 ... But despite .await returning a Result, I keep getting the inner error:

```No such file or directory (os error 2)```

File: src/actix/mod.rs [hunk 1]
```diff
@@ -183,7 +213,22 @@ pub fn init(
...
+                                .route(actix_web::web::get().to(web_ui_index)),
+                        )
+                        .service(
+                            actix_web::web::resource("/")
+                                .route(actix_web::web::get().to(web_ui_index)),
+                        )
+                        .service(
+                            actix_files::Files::new("/", &static_folder)
+                                .path_filter(|path, _| Path::new("index.html").ne(path)),
+                        ),
```
xzfc, line 231: You could simplify the implementation by using `DefaultHeaders` middleware. It will append the header on all files, not just `index.html` but I guess there is no harm in that.

```rust
app = app.service(
    actix_web::web::scope(WEB_UI_PATH)
        .wrap(
            DefaultHeaders::new()
                .add(("X-Frame-Options", HeaderValue::from_static("DENY"))),
        )
        .service(
            actix_files::Files::new("/", &static_folder).index_file("index.html"),
        ),
)
```

generall: not only on all files, but on all API calls as well

xzfc: This part of code affects only the `/dashboard` path and its descendants.

generall: hm, really? In this case it is a better option indeed

timvisee: I agree that the suggested approach is a lot better, and so I've pushed it onto this branch.

I've confirmed the header is set exclusively on dashboard responses.

timvisee (approved):

src/actix/mod.rs
Rendez: Though I agree this simplifies things, I explicitly did it the way I did it to prevent setting a useless header for all other static assets.

PR #4518, opened 2024-06-20
Switch to poetry

xzfc:
This PR replaces a bunch of `requirements.txt` files with a single `pyproject.toml`.


timvisee:
I'm not sure what our final verdict was on this.

Do we want to keep it for installing dependencies, but not more than that? If so, is the current state of this PR ready?

generall (approved):

.github/workflows/integration-tests.yml
```diff
@@ -49,15 +57,19 @@ jobs:
...
+          poetry -C tests install --no-root
       - name: Build
         run: cargo build --features "service_debug data-consistency-check"
       - name: Run integration tests - 1 peer
-        run: ./tests/integration-tests.sh distributed
+        run: poetry -C tests run ./tests/integration-tests.sh distributed
         shell: bash
       - name: Run integration tests - multiple peers - pytest
-        run: pytest ./tests/consensus_tests
+        run: poetry -C tests run pytest tests/consensus_tests
```
timvisee, line 72: @generall These are different. But I assume you're fine with this because you can still invoke `pytest` directly locally, am I right?

timvisee (approved):

.github/workflows/integration-tests.yml
generall: yes, exactly. And I can use my own env as well

PR #4520, opened 2024-06-21
Fixing various API docs

agourlay:
gathering some API doc fixes


lib/collection/src/operations/types.rs
```diff
@@ -314,7 +314,7 @@ pub struct ScrollRequestInternal {
     /// Select which payload to return with the response. Default: All
     pub with_payload: Option<WithPayloadInterface>,
```
coszio, line 315: Nitpicking on unchanged lines: default for scroll is `true`, while other endpoints is `false`, not `All` and `None`

coszio (approved):

PR #4524, opened 2024-06-21
Split-by-shard when forwarding operations during resharding shard transfer

ffuugoo:
Tracked in #4213.
Follow-up to #4516.

Resharding shard transfer initializes `ForwardProxyShard` that forwards incoming updates from local shard `X` to remote shard `Y`.

In such case, `ForwardProxyShard` should additionally split operation by shard, and only forward a *part* of operation, that belongs to the remote shard.

E.g.:
- we are resharding from 2 shards to 3 shards
- there's a resharding shard transfer from shard 1 to shard 3
  - and so, there's a `ForwardProxyShard` from shard 1 to shard 3
- user sends `upsert 42, 69, 128, 228, 666, 1337`
- we split-by-shard on the collection/shard-holder level and get
  - shard 1: `upsert 42, 128, 228`
  - shard 2: `upsert 69, 666, 1337`
  - shard 3: `upsert 42, 69`
- now, imagine if we go down the callstack, to the local shard of shard 1 replica set
- it is a `ForwardProxyShard`, so `ForwardProxyShard::update` is called
- the operation is already split-by-shard on collection/shard-holder level
- so we will be forwarding `upsert 42, 128, 228`
- but in case of *resharding* shard transfer, we only have to forward `upsert 42`
- so we have to, once again, split-by-shard this operation, to only forward relevant part of it to remote shard

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/forward_proxy_shard.rs
+++ b/lib/collection/src/shards/forward_proxy_shard.rs
@@ -42,7 +42,7 @@ pub struct ForwardProxyShard {
     shard_id: ShardId,
     pub(crate) wrapped_shard: LocalShard,
     pub(crate) remote_shard: RemoteShard,
-    hash_ring: Option<HashRing>,
+    resharding_hash_ring: Option<HashRing>,
     /// Lock required to protect transfer-in-progress updates.
     /// It should block data updating operations while the batch is being transferred.
     update_lock: Mutex<()>,
@@ -53,13 +53,13 @@ impl ForwardProxyShard {
         shard_id: ShardId,
         wrapped_shard: LocalShard,
         remote_shard: RemoteShard,
-        hash_ring: Option<HashRing>,
+        resharding_hash_ring: Option<HashRing>,
     ) -> Self {
         Self {
             shard_id,
             wrapped_shard,
             remote_shard,
-            hash_ring,
+            resharding_hash_ring,
             update_lock: Mutex::new(()),
         }
     }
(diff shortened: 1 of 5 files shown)


ffuugoo:
Draft, because this feels *super* hacky. But at least it was rather trivial to implement.

Initially, I hoped this can be done without any forward proxies at all, because we are doing full-cluster syncs during resharding, and can simply synchronize so that every shard will be aware of the new shard and would include it into `split_by_shard`. But `ForwardProxyShard` also ensures that when doing (resharding) stream records transfer we deliver all updates to the transfer receiver *in correct order*, and I think we still need `ForwardProxyShard`s for that. :/

@generall I'd appreciate some input here, maybe you can come up with a better idea how to implement this.

lib/collection/src/shards/replica_set/shard_transfer.rs
```diff
@@ -20,7 +20,11 @@ impl ShardReplicaSet {
     /// # Cancel safety
     ///
     /// This method is cancel safe.
-    pub async fn proxify_local(&self, remote_shard: RemoteShard) -> CollectionResult<()> {
+    pub async fn proxify_local(
+        &self,
+        remote_shard: RemoteShard,
+        hash_ring: Option<HashRing>,
```
generall, line 26: let's mention, that it is a new hashring

ffuugoo: Renamed to `resharding_hash_ring`.

lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -210,31 +218,71 @@ impl ShardOperation for ForwardProxyShard {
...
+            //
+            // In such case `split_by_shard` call above would be a no-op, and we can preserve the
+            // clock tag.
+            let tag = if self.shard_id != self.remote_shard.id {
+                None
+            } else {
+                operation.clock_tag
+            };
+
+            op.map(|op| OperationWithClockTag::new(op, tag))
```
timvisee, line 252: I don't see a reason to keep the other branch, as we even assert we point to a different shard a few lines above.

```suggestion
            // Strip the clock tag from the operation, because clock tags are incompatible between
            // different shards.
            OperationWithClockTag::new(op, None)
```

timvisee (approved):

lib/collection/src/shards/forward_proxy_shard.rs
ffuugoo: This is kinda intended.

If we ever accidentally pass `Some` hashring into a non-resharding forward proxy... it would still work as-is. Split by shard would just be a no-op in this case.

So Andrey wanted to structure this something like this:

```rust
let mut forward_operation = if let Some(ring) = hash_ring {
	// ...
} else {
    Some(operation)
};

if self.shard_id != self.remote_shard.id {
  forward_operation.as_mut().map(|op| op.operation.tag = None);
}
```

But, IMO, it makes `let Some(ring) = hash_ring` branch harder to understand, so I've moved `self.shard_id != self.remote_shard.id` under hashring branch.

I agree that it's kinda (like, 99%) pointless, but it is possible to miss `debug_assert`, and it does not hurt/degrade anything, so I don't mind having this.

I'll add a `log::warn` under that branch, though.

PR #4536, opened 2024-06-24
universal-query: Add validations to query input

agourlay:
Adding some missing validations to the query API input.

### DIFF
--- a/lib/collection/src/operations/universal_query/collection_query.rs
+++ b/lib/collection/src/operations/universal_query/collection_query.rs
@@ -294,6 +294,13 @@ impl CollectionPrefetch {
             ));
         }
 
+        // Check no score_threshold without a query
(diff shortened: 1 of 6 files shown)


lib/collection/src/operations/universal_query/collection_query.rs
```diff
@@ -294,6 +294,13 @@ impl CollectionPrefetch {
             ));
         }
 
+        // Check no score_threshold without a query
+        if self.score_threshold.is_some() && self.query.is_none() {
+            return Err(CollectionError::bad_request(
+                "A query is needed to use the score_threshold. Can't have score_threshold without defining a query.",
+            ));
+        }
```
coszio, line 302: But also no score_threshold for `order_by`

coszio (changes requested):

coszio (approved):

PR #4537, opened 2024-06-24
Histogram save load

IvanPleshkov:
This PR is a part of on-disk-payload-feature.

For the on-disk numeric index, it was decided that the histogram would be still on RAM because it's pretty small. But we need persistence anyway to avoid a full scan while loading.

This PR adds save-load for histogram and use it in histogram unit test.

### DIFF
--- a/lib/segment/src/index/field_index/histogram.rs
+++ b/lib/segment/src/index/field_index/histogram.rs
@@ -138,8 +138,7 @@ impl<T: Numericable + Serialize + DeserializeOwned> Histogram<T> {
     }
 
     #[allow(dead_code)]
-    pub fn load(path: impl AsRef<Path>) -> OperationResult<Self> {
-        let path = path.as_ref().to_owned();
+    pub fn load(path: &Path) -> OperationResult<Self> {
         let config_path = path.join(CONFIG_PATH);
         let borders_path = path.join(BORDERS_PATH);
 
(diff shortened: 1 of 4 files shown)


lib/segment/src/index/field_index/tests/histogram_tests.rs
```diff
@@ -249,6 +251,15 @@ fn test_build_histogram() {
...
     let (histogram, points_index) = build_histogram(max_bucket_size, precision, points);
+
+    // test persistence
+    let dir = tempfile::Builder::new()
+        .prefix("histogram_dir")
+        .tempdir()
+        .unwrap();
+    histogram.save(&dir).unwrap();
+    let histogram = Histogram::<f64>::load(&dir).unwrap();
+
```
xzfc, line 262: I suggest to not to mix things under the test. Instead, split into two parts:
In the first, run the old test without saving/loding.
In the second, do load/save and compare histograms using `assert_eq!()`.

IvanPleshkov: fixed. I reverted this change and added a separate unit test for test save-load

xzfc (approved):

lib/segment/src/index/field_index/histogram.rs
```diff
@@ -123,6 +137,47 @@ impl<T: Numericable> Histogram<T> {
         }
     }
 
+    #[allow(dead_code)]
+    pub fn load(path: impl AsRef<Path>) -> OperationResult<Self> {
+        let path = path.as_ref().to_owned();
```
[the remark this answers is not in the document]
IvanPleshkov: Fixed

PR #4538, opened 2024-06-24
universal-query: Add farthest query

coszio:
Adds a `{ query: { farthest: vector_input } }` shorthand for `{query: { recommend: { negative: [vector_input], strategy: "best_score" } } }` for better discoverability of features

### DIFF
--- a/lib/api/src/rest/schema.rs
+++ b/lib/api/src/rest/schema.rs
@@ -235,6 +235,9 @@ pub enum Query {
 
     /// Return points that live in positive areas.
     Context(ContextQuery),
+    
+    /// Return points that are the farthest apart from this vector. Shorthand for recommend best_score with only one negative input
+    Farthest(FarthestQuery),
 
     /// Order the points by a payload field.
     OrderBy(OrderByQuery),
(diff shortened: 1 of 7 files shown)


lib/api/src/rest/schema.rs
```diff
@@ -267,6 +270,12 @@ pub struct ContextQuery {
     pub context: ContextInput,
 }
 
+#[derive(Debug, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct FarthestQuery {
+    pub furthest: VectorInput,
```
agourlay, line 276: the struct says `farthest` but the field says `furthest` :)

agourlay:
TBH I am unsure about the value of the shortcut.

```json
{ 
  "query": { 
    "farthest": "vector_input"
   }
}
```

vs

```json
{
  "query": { 
    "recommend": { 
      "negative":  ["vector_input"],
      "strategy": "best_score"
    }
  }
}
```

It does not remove so much input for committing us on new public API forever.

Have you seen users confuse about using recommendation this way? Is it maybe a documentation issue?

coszio:
> It does not remove so much input for committing us on new public API forever.

It is in the same spirit of ` "query": <vector_input>` for nearest search of providing a less cluttered interface for the simple cases. But it is indeed a commit to consider.

agourlay:
Still on the fence for the decision, I'd live to have other members express their opinion :)

PR #4542, opened 2024-06-24
Avg multivector

IvanPleshkov:
Fix the missing average multivector and integration test for it

### DIFF
--- a/lib/collection/src/recommendations.rs
+++ b/lib/collection/src/recommendations.rs
@@ -5,8 +5,8 @@ use std::time::Duration;
 use api::rest::RecommendStrategy;
 use itertools::Itertools;
 use segment::data_types::vectors::{
-    DenseVector, NamedQuery, NamedVectorStruct, Vector, VectorElementType, VectorRef,
-    DEFAULT_VECTOR_NAME,
+    DenseVector, NamedQuery, NamedVectorStruct, TypedMultiDenseVector, Vector, VectorElementType,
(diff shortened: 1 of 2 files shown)


lib/collection/src/recommendations.rs
```diff
@@ -51,36 +53,43 @@ fn avg_vectors<'a>(vectors: impl IntoIterator<Item = VectorRef<'a>>) -> Collecti
...
+        (0, _, 0) => {
             for item in &mut avg_sparse.values {
                 *item /= sparse_count as VectorElementType;
             }
-            Ok(avg_sparse.into())
+            Ok(Vector::from(avg_sparse))
         }
-        (_, _) => Err(CollectionError::bad_input(
-            "Can't average dense and sparse vectors together".to_owned(),
+        (0, 0, _) => Ok(Vector::from(avg_multi.unwrap())),
```
agourlay, line 90: maybe we could do it without the `unwrap` :)

agourlay (approved):

lib/collection/src/recommendations.rs
IvanPleshkov: There are no default multivector to replace unwrap. And placeholder cannot be constructed because if `avg_multi` is empty, dimension is unknown here. I can only return `service_error` instead

IvanPleshkov: fixed

PR #4551, opened 2024-06-24
Fix id exclusion when recommending from a different collection

agourlay:
This PR fixes a bug in the current recommendation (discovered while working on https://github.com/qdrant/qdrant/pull/4508)

The recommendation API excludes from the results the vector ids used in the query (positives & negatives).
However, this should not happen when using `lookup_from` as the ids are pointing to a different collection.

### DIFF
--- a/tests/openapi/openapi_integration/test_multicollection_reco.py
+++ b/tests/openapi/openapi_integration/test_multicollection_reco.py
@@ -239,3 +239,86 @@ def test_recommend_from_another_collection(on_disk_vectors):
         path_params={'collection_name': collection_name2},
     )
     assert response.ok
+
+
+def test_recommend_lookup():
+    # delete lookup collection if exists
+    response = request_with_validation(
+        api='/collections/{collection_name}',
+        method="DELETE",
+        path_params={'collection_name': collection_name2},
+    )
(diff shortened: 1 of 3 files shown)


tests/openapi/openapi_integration/test_multicollection_reco.py
```diff
@@ -239,3 +239,86 @@ def test_recommend_from_another_collection(on_disk_vectors):
...
+            "negative": [[0.0, 0.0, 0.0, 2.0]],
+            "limit": 10,
+        },
+    )
+    assert response.ok, response.text
+    recommend_result_by_vector = response.json()["result"]
+
+    # results should be equivalent
+    assert recommend_result_by_id == recommend_result_by_vector,\
+        f"recommend_result_by_id: {recommend_result_by_id}, recommend_result_by_vector: {recommend_result_by_vector}"
```
coszio, line 324: No trailing empty line 🙈

coszio (approved):
Thanks for the fix! We should also fix this for discover api

agourlay:
@coszio Thanks for the quick review, I will fix the discovery API as well :+1:

timvisee (approved):
Thanks for adding a test too!

PR #4553, opened 2024-06-25
Fix discovery API id exclusion and lookup

agourlay:
Similar to https://github.com/qdrant/qdrant/pull/4551 but for the discovery API.

Additionally the `lookup_from` was wrongly used for the `using` field of the search query.

### DIFF
--- a/lib/collection/src/discovery.rs
+++ b/lib/collection/src/discovery.rs
@@ -96,20 +96,22 @@ fn discovery_into_core_search(
         _ => referenced_ids,
     };
 
-    let filter = {
+    let filter = if reference_vectors_ids_to_exclude.is_empty() {
+        request.filter
+    } else {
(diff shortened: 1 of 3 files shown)


lib/collection/src/discovery.rs
```diff
@@ -77,19 +80,25 @@ fn discovery_into_core_search(
         // Target with/without pairs => Discovery
         (Some(target), pairs) => QueryEnum::Discover(NamedQuery {
             query: DiscoveryQuery::new(target, pairs),
-            using: Some(lookup_vector_name),
```
agourlay, line 80: bonus bug :)

coszio: Dang

coszio (approved):
Thanks for the fix!

PR #4554, opened 2024-06-25
Fix handling of OOD issues in optimizer

### LINKED ISSUE #4297: Implement better handling of OOD issues in optimizer

Currently, if the optimizer is used, it creates a new temporary segment with a copy of data from those segments, which are currently under optimization. 
This process requires sometimes significant amount of disk space and if it is not enough, current implementation might panic.

In optimization process, we can check if there are enough disk space to continue indexing or not.
A good space to do it is https://github.com/qdrant/qdrant/blob/ccf7f1d24f37dcfe82dad85fa66fb91352bd7a21/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs#L100

Indexed segment is potentially a plain structure. There are not too many places where we can't handle the OOD error.
Alternatively, we can check if we can actually handle OOD errors in this case, instead of checking for empty space each time.

**Additional context**

We created a test that simulates the scenario of disk overflow https://github.com/qdrant/qdrant/pull/4267

The solution for this issue should resolve this test scenario in the most general way reasonable.

the-johnwick:
fixes #4297
/claim #4297

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -117,6 +117,7 @@ pub trait SegmentOptimizer {
         //     image_vectors: 10100 * dim * VECTOR_ELEMENT_SIZE
         // }
         let mut bytes_count_by_vector_name = HashMap::new();
+        let mut total_required_space: u64 = 0;
 
         for segment in optimizing_segments {
             let segment = match segment {
@@ -129,11 +130,37 @@ pub trait SegmentOptimizer {
             };
             let locked_segment = segment.read();
 
+            match fs4::total_space(locked_segment.data_path()) {
+                Ok(size) => total_required_space += size,
+                Err(error) => log::error!(
+                    "Failed to fetch available space for path: {} due to: {}",
+                    self.temp_path().display(),
+                    error.to_string(),
+                ),
+            }
+
             for vector_name in locked_segment.vector_names() {
                 let vector_size = locked_segment.available_vectors_size_in_bytes(&vector_name)?;
                 let size = bytes_count_by_vector_name.entry(vector_name).or_insert(0);
                 *size += vector_size;
             }
+        }
+
+        let mut available_space: u64 = 0;
+
+        match fs4::available_space(self.temp_path()) {
+            Ok(size) => available_space = size,
+            Err(error) => log::error!(
+                "Failed to fetch available space for path: {} due to: {}",
+                self.temp_path().display(),
+                error.to_string(),
+            ),
+        }
+
+        if available_space < total_required_space {
+            return Err(CollectionError::service_error(
+                "Not enough disk space for optimization".to_string(),
+            ));
         }
 
         // Example: maximal_vector_store_size_bytes = 10200 * dim * VECTOR_ELEMENT_SIZE
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -195,10 +195,14 @@ jobs:
         working-directory: ./tests/low-ram
         shell: bash
         run: ./low-ram.sh
-      - name: Run low Disk test
+      - name: Run low Disk test - search
         working-directory: ./tests/low-disk
         shell: bash
         run: ./low-disk.sh
+      - name: Run low Disk test - indexing
+        working-directory: ./tests/low-disk
+        shell: bash
+        run: ./low-disk.sh indexing
 
   test-snapshot-operations-s3-minio:
     runs-on: ubuntu-latest
--- a/tests/low-disk/create_and_search_items.py
+++ b/tests/low-disk/create_and_search_items.py
@@ -1,8 +1,12 @@
 import argparse
 import random
 import uuid
+from time import sleep
 
 import requests
+
+
+VECTOR_SIZE = 4
 
 
 def generate_points(amount: int):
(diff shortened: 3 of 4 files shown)


the-johnwick:
@generall  waiting for your feedback whether it is going in right direction or not.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -143,6 +156,17 @@ pub trait SegmentOptimizer {
             .copied()
             .unwrap_or(0);
 
+        // Estimate required disk space
+        let required_space = maximal_vector_store_size_bytes;
```
generall, line 160: I would argue, that required_space should also include the paylaod size.

I propose to take sum of optimizing segments disk size as a reference instead.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -38,6 +39,18 @@ pub struct OptimizerThresholds {
     pub indexing_threshold_kb: usize,
 }
 
+fn available_disk_space(path: &Path) -> u64 {
+    let mut sys = System::new_all();
+    sys.refresh_disks_list();
+
+    for disk in sys.disks() {
```
generall, line 46: We are already using `fs4` for he similar purpose https://github.com/qdrant/qdrant/blob/1efec29f65c7569a3ae3b76bbd511410556c53c6/lib/collection/src/shards/local_shard/disk_usage_watcher.rs#L106

I would prefer to keep the consistency

generall (changes requested):
Hey @the-johnwick, thanks for the PR! Overall I agree with the approach, but there are a couple of thing I would prefer to change. Please see the comments

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
the-johnwick: Sure, but I am confused about calculating payload's total size.

Rutik7066:
@generall  added the fs4 to check available space and the payload size to the required size. It may not be the right way to calculate payload size. Please let me know .

the-johnwick:
@generall added the fs4 to check available space and the payload size to the required size. It may not be the right way to calculate payload size. Please let me know.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -1,3 +1,4 @@
+use std::borrow::Borrow;
```
generall, line 1: this doesn't look related, is it?

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -128,11 +130,11 @@ pub trait SegmentOptimizer {
                 }
             };
             let locked_segment = segment.read();
-
             for vector_name in locked_segment.vector_names() {
                 let vector_size = locked_segment.available_vectors_size_in_bytes(&vector_name)?;
                 let size = bytes_count_by_vector_name.entry(vector_name).or_insert(0);
                 *size += vector_size;
+                total_payload_size_bytes += vector_size
```
generall, line 137: I don't think it is a correct way to estimate payload.

I think what we need to do is to simply check what's the current disk usage of the segment is, not try to estimate it with some form of proxy value.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -143,6 +145,21 @@ pub trait SegmentOptimizer {
...
             .unwrap_or(0);
 
+        let mut available_space;
+        match fs4::available_space(self.temp_path()) {
+            Some(size) => available_space = size,
+            Err(error) => CollectionError::service_error(format!(
+                "Failed to fetch available space for path: {} due to: {}",
+                self.temp_path().to_owned(),
+                error.to_string()
+            )),
```
generall, line 155: Some file systems might not support it, so we would prefer to go on with the optimization instead of failing. 
So logging the error should be enough

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
the-johnwick: Yes, I think it is added by mistake. I will remove it.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
the-johnwick: Sure I will update the changes.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -101,6 +101,30 @@ pub trait SegmentOptimizer {
...
+        match fs4::available_space(self.temp_path()) {
+            Some(size) => available_space = size,
+            Err(error) => log::error!(
+                "Failed to fetch available space for path: {} due to: {}",
+                self.temp_path().to_owned(),
+                error.to_string(),
+            ),
+        }
+        let mut total_space;
+        match fs4::total_space(self.segments_path()) {
```
generall, line 114: it should be sum of `optimizing_segments`, not all segments in overall

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -128,6 +128,14 @@ pub trait SegmentOptimizer {
                 }
             };
             let locked_segment = segment.read();
+            match fs4::total_space(locked_segment.data_path()) {
+                Some(size) => total_space = size,
```
generall, line 132:
```suggestion
                Some(size) => total_space += size,
```

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -117,7 +117,7 @@ pub trait SegmentOptimizer {
         //     image_vectors: 10100 * dim * VECTOR_ELEMENT_SIZE
         // }
         let mut bytes_count_by_vector_name = HashMap::new();
-
+        let mut total_required_space;
```
generall, line 120:
```suggestion
        let mut total_required_space = 0;
```

the-johnwick:
@generall  apologies for such silly mistakes.

generall:
Hey @the-johnwick, could you please rebase on top of https://github.com/qdrant/qdrant/pull/4267
So we can be sure that the test case is fixed

the-johnwick:
@generall  Sure.

Cargo.toml
```diff
@@ -70,7 +70,7 @@ tokio = { workspace = true }
 
 actix-cors = "0.7.0"
 actix-files = "0.6.6"
-actix-web = { version = "4.7.0", optional = true, features = ["rustls-0_23", "actix-tls"] }
```
generall, line 73: how is this change related to the issue?

the-johnwick: No. actually I have messed up while trying to rebase

the-johnwick: Do you know, how to undo the changes?

the-johnwick:
@generall Can we merge this.

generall:
Close in favor of https://github.com/qdrant/qdrant/pull/4578

PR #2415, opened 2023-08-10
Limit snapshot recovery file URIs to snapshots directory

timvisee:
Depends on <https://github.com/qdrant/qdrant/pull/2414>.

Fixes a potential security issue where arbitrary files from the file system could be read as snapshot.

This limits recovering snapshots with the `file://` URI to the snapshots directory.

```json
curl -v -X PUT --header 'Content-Type: application/json' --data '{"location": "file:///etc/shadow"}' http://localhost:6333/collections/whatever/snapshots/recover
{"status":{"error":"Forbidden: Snapshot file \"/etc/shadow\" must be inside snapshots dir"},"time":0.00030464}
```

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your core changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #4572, opened 2024-06-27
Limit snapshot recovery file uri to snapshots dir

timvisee:
Re-implements: <https://github.com/qdrant/qdrant/pull/2415>

Fixes a security issue where arbitrary files from the file system could be read as snapshot.

This limits recovering snapshots with the `file://` URI to the snapshots directory.

We've opened this change before for discussion but it became stale. This reopens it. I don't see a problem with this limitation and it's good practice anyway.

Even though `file://` normally doesn't support relative paths, I've added support for this to keep backward compatibility. Of course, these files are jailed within the snapshot directory as well.

This:

```json
PUT /collections/pentest/snapshots/recover
{
  "location": "file:///etc/passwd"
}
```

now returns:

```json
{
  "error": "Forbidden: Snapshot file \"/etc/passwd\" must be inside snapshots directory"
}
```

With support for relative files this also works if the given snapshot file exists:

```json
PUT /collections/pentest/snapshots/recover
{
  "location": "file://./my_snapshot.snapshot"
}
```

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/snapshots/download.rs
+++ b/lib/storage/src/content_manager/snapshots/download.rs
@@ -58,15 +58,26 @@ async fn download_file(
     Ok(temp_path)
 }
 
-/// Download a snapshot from the given URI.
+/// Download a snapshot file an URI, return a file path to it
+///
+/// For remote resources such as an URL, this downloads the given file and puts it in
+/// `snapshots_dir`. A path to the downloaded file is returned.
+///
+/// For `file://` URIs a direct path is returned.
 ///
 /// May returen a `TempPath` if a file was downloaded from a remote source. If it is dropped the
 /// downloaded file is deleted automatically. To keep the file `keep()` may be used.
+///
+/// # Security
+///
+/// A `file://` URI may point to arbitrary files on the file system, which could be a security
+/// concern. Set `strict_file` to `true` to enforce a local file to be inside `snapshots_dir`.
 #[must_use = "may return a TempPath, if dropped the downloaded file is deleted"]
 pub async fn download_snapshot(
     client: &reqwest::Client,
     url: Url,
     snapshots_dir: &Path,
+    only_snapshot_dir: bool,
 ) -> Result<(PathBuf, Option<TempPath>), StorageError> {
     match url.scheme() {
         "file" => {
@@ -81,18 +92,19 @@ pub async fn download_snapshot(
                 )));
             }
 
-            // Try to canonicalize paths so we can more reliably compare with required prefix
-            let snapshots_dir = snapshots_dir
-                .canonicalize()
-                .unwrap_or_else(|_| snapshots_dir.to_path_buf());
             let local_path = local_path.canonicalize().unwrap_or(local_path);
 
             // Prevent using arbitrary files from our file system, enforce the file to be in the
             // snapshots directory
-            if !local_path.starts_with(snapshots_dir) {
-                return Err(StorageError::forbidden(format!(
-                    "Snapshot file {local_path:?} must be inside snapshots directory",
-                )));
+            if only_snapshot_dir {
+                let snapshots_dir = snapshots_dir
+                    .canonicalize()
+                    .unwrap_or_else(|_| snapshots_dir.to_path_buf());
+                if !local_path.starts_with(snapshots_dir) {
+                    return Err(StorageError::forbidden(format!(
+                        "Snapshot file {local_path:?} must be inside snapshots directory",
+                    )));
+                }
             }
 
             Ok((local_path, None))
--- a/lib/storage/src/content_manager/snapshots/download.rs
+++ b/lib/storage/src/content_manager/snapshots/download.rs
@@ -10,6 +10,9 @@ use uuid::Uuid;
 
 use crate::StorageError;
 
+/// Special file:// URI prefix for relative paths
+const URI_FILE_RELATIVE_PREFIX: &str = "file://./";
+
 fn random_name() -> String {
     format!("{}.snapshot", Uuid::new_v4())
 }
@@ -81,11 +84,7 @@ pub async fn download_snapshot(
 ) -> Result<(PathBuf, Option<TempPath>), StorageError> {
     match url.scheme() {
         "file" => {
-            let local_path = url.to_file_path().map_err(|_| {
-                StorageError::bad_request(
-                    "Invalid snapshot URI, file path must be absolute or on localhost",
-                )
-            })?;
+            let local_path = resolve_uri_file_path(&url, snapshots_dir)?;
             if !local_path.exists() {
                 // Report user provided URL here to prevent leaking the local path
                 return Err(StorageError::bad_request(format!(
@@ -128,4 +127,33 @@ pub async fn download_snapshot(
             url.scheme()
         ))),
     }
+}
+
+/// Resolve a file:// URI to a local path
+///
+/// This supports both absolute and relative paths. If the path is relative, it is resolved within
+/// the given `workdir`.
+///
+/// # Security
+///
+/// This may point to arbitrary files. The resolved file may not exist.
+pub fn resolve_uri_file_path(url: &Url, workdir: &Path) -> Result<PathBuf, StorageError> {
+    // Must be a file URI
+    if url.scheme() != "file" {
+        return Err(StorageError::service_error(
+            "provided URI is not a file:// URI",
+        ));
+    }
+
+    // Parse relative path with specific prefix, normally not supported
+    if let Some(relative_path) = url.to_string().strip_prefix(URI_FILE_RELATIVE_PREFIX) {
+        let workdir = workdir
+            .canonicalize()
+            .unwrap_or_else(|_| workdir.to_path_buf());
+        return Ok(workdir.join(relative_path));
+    }
+
+    // Parse absolute path
+    url.to_file_path()
+        .map_err(|_| StorageError::bad_request("Malformed file URI"))
 }
--- a/lib/storage/src/content_manager/snapshots/download.rs
+++ b/lib/storage/src/content_manager/snapshots/download.rs
@@ -114,6 +114,11 @@ pub async fn download_snapshot(
         "http" | "https" => {
             let download_to = snapshots_dir.join(snapshot_name(&url));
 
+            log::debug!(
+                "Downloading snapshot from {url} to {}",
+                snapshots_dir.display(),
+            );
+
             let temp_path = download_file(client, &url, &download_to).await?;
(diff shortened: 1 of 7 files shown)


src/actix/api/snapshot_api.rs
```diff
@@ -211,6 +211,8 @@ async fn upload_snapshot(
             &collection.name,
             snapshot_recover,
             access,
+            // No need to limit to snapshot directory, only uploaded file is used
+            false,
```
mac-chaffee, line 215: Is it necessary to set this to `false`? Couldn't we set it to `true` just to be safe, since `do_save_uploaded_snapshot` already tries to force the file to be in the snapshot directory? https://github.com/qdrant/qdrant/blob/5c29cad76edc4fe33882d385c02633352666d67e/src/actix/api/snapshot_api.rs#L110

Currently we use string validation to block relative paths in the filename in the `multipart/form-data` request body and collection name, which could mean slight changes could make the vuln come back, so I like the idea of checking the path twice if it doesn't break other stuff

timvisee: > Is it necessary to set this to false? Couldn't we set it to true just to be safe, since do_save_uploaded_snapshot already tries to force the file to be in the snapshot directory? 

Yes that is required, in the current implementation. When uploading it might store snapshots in a different directory if the user configured it that way, in which case the check would fail.

You're right that we use string-ish validation, but we can not really represent file paths in a different way. To add a bit of nuance here, we use [`Path`](https://doc.rust-lang.org/std/path/struct.Path.html) from the standard library here. I believe we can assume this to be safe for what we're trying to do here - strip everything but a file name from an arbitrary path. Note that if this step fails, we pick a random UUID.

The only solution I can think of is to use something like [`cap-std`](https://docs.rs/cap-std/latest/cap_std/) which enforces directory namespacing. However, it is not compatible with our current snapshot logic. I did attempt to integrate this before, but had to decide it was not worth the effort.

So what's your opinion on this after the above?

mac-chaffee: > When uploading it might store snapshots in a different directory if the user configured it that way, in which case the check would fail.

If the user configured a different snapshots directory, couldn't we read that configuration and enforce the file is being saved there? Or am I misunderstanding?

Overall this PR looks good, just wanted to ask if flipping that boolean could make us a little safer for free. Sounds like it's not that simple :)

mac-chaffee (approved):

lib/storage/src/content_manager/snapshots/download.rs
```diff
@@ -95,3 +128,32 @@ pub async fn download_snapshot(
...
+        return Err(StorageError::service_error(
+            "provided URI is not a file:// URI",
+        ));
+    }
+
+    // Parse relative path with specific prefix, normally not supported
+    if let Some(relative_path) = url.to_string().strip_prefix(URI_FILE_RELATIVE_PREFIX) {
+        let workdir = workdir
+            .canonicalize()
+            .unwrap_or_else(|_| workdir.to_path_buf());
```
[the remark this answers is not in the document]
ffuugoo: Also, I'm not sure if `canonicalize` is necessary here. We'll canonicalize during `only_snapshots_dir` check in `download_snapshot` anyway.

src/common/snapshots.rs
```diff
@@ -128,8 +128,13 @@ pub async fn recover_shard_snapshot(
                     let client = client.client(api_key.as_deref())?;
 
                     let (snapshot_path, snapshot_temp_path) =
-                        snapshots::download::download_snapshot(&client, url, download_dir.path())
-                            .await?;
+                        snapshots::download::download_snapshot(
+                            &client,
+                            url,
+                            download_dir.path(),
+                            false,
```
ffuugoo, line 135: We explicitly only allow `http://` and `https://` URLs for shard snapshots (and local files are handled explicitly, without relying on `file://` URLs), so I'd enforce `only_snapshots_dir` here, just in case.

```suggestion
                            true,
```

ffuugoo (changes requested):
I'm sorry I had to do this, @timvisee. 🥲

src/actix/api/snapshot_api.rs
timvisee: We don't move the file, but we do directly store it in the snapshots directory now.

IIRC we documented it differently, like I described above. We either have to update documentation, or store in the proper directory.

lib/storage/src/content_manager/snapshots/download.rs
```diff
@@ -58,33 +61,63 @@ async fn download_file(
...
             if !local_path.exists() {
+                // Report user provided URL here to prevent leaking the local path
                 return Err(StorageError::bad_request(format!(
-                    "Snapshot file {local_path:?} does not exist"
+                    "Snapshot file {:?} does not exist",
+                    url.to_string(),
                 )));
             }
+
+            let local_path = local_path.canonicalize().unwrap_or(local_path);
```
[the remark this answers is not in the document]
timvisee: > path does not exist.

This prevents breakage if the snapshots directory does not exist somehow. I came across this during testing.

lib/storage/src/content_manager/snapshots/download.rs
```diff
@@ -58,33 +61,63 @@ async fn download_file(
...
+
+            // Prevent using arbitrary files from our file system, enforce the file to be in the
+            // snapshots directory
+            if only_snapshot_dir {
+                let snapshots_dir = snapshots_dir
+                    .canonicalize()
+                    .unwrap_or_else(|_| snapshots_dir.to_path_buf());
```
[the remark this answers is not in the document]
timvisee: It prevents an extra error if the file does not exist.

But in this case it doesn't really matter whether we error here or error later. I'd be fine by not `unwrap_or_else` here. It doesn't change the final behavior.

lib/storage/src/content_manager/snapshots/download.rs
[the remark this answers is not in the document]
timvisee: Two reasons:
- the working directory might not exist (like the above)
- a relative path is not a valid URI, so `url.to_file_path()` breaks if the path is not canonicalized.

Note that we do support relative paths ourselves for backward compatibility reasons, but `url` itself does not.

src/actix/api/snapshot_api.rs
ffuugoo: Am I missing something?

- https://github.com/qdrant/qdrant/blob/dev/src/actix/api/snapshot_api.rs#L196-L197
- https://github.com/qdrant/qdrant/blob/dev/src/actix/api/snapshot_api.rs#L89-L112

src/common/snapshots.rs
timvisee: Sure. Lets wait with this one until we've fixed actually downloading to the correct directory.

src/actix/api/snapshot_api.rs
ffuugoo: Also:
- https://github.com/qdrant/qdrant/blob/dev/src/actix/mod.rs#L136
- https://github.com/qdrant/qdrant/blob/dev/src/actix/mod.rs#L66-L69
- https://github.com/qdrant/qdrant/blob/master/lib/storage/src/content_manager/toc/temp_directories.rs#L122-L128

lib/storage/src/content_manager/snapshots/download.rs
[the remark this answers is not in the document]
ffuugoo: IMO, we should just create the snapshots directory if it does not exist.

lib/storage/src/content_manager/snapshots/download.rs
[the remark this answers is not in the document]
ffuugoo: IMO, we should just create snapshots directory if it does not exist.

lib/storage/src/content_manager/snapshots/download.rs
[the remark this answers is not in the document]
ffuugoo: IMO, we should just create workdir. (We are providing workdir path anyway, it's not controlled by client.)

src/actix/api/snapshot_api.rs
ffuugoo: Yes, but this whole thread is about `upload_snapshot`?

timvisee:
@ffuugoo With the latest changes I should now have resolved everything. Mind another look over this?

I've removed the `only_snapshots_dir` boolean because it ended up always being `true`. It now always does the check.

lib/storage/src/content_manager/snapshots/download.rs
```diff
@@ -93,29 +91,37 @@ pub async fn download_snapshot(
...
-                        "Snapshot file {local_path:?} must be inside snapshots directory",
-                    )));
-                }
+            let inside_snapshots_dir = snapshots_dir
+                .canonicalize()
+                .map_or(false, |snapshots_dir| local_path.starts_with(snapshots_dir));
+            if !inside_snapshots_dir {
+                return Err(StorageError::forbidden(format!(
+                    "Snapshot file {local_path:?} must be inside snapshots directory",
+                )));
```
ffuugoo, line 114:
```suggestion
                return Err(StorageError::forbidden(format!(
                    "Snapshot file {} must be inside snapshots directory",
                    local_path.display(),
                )));
```

ffuugoo (approved):

ffuugoo:
Quick update to the review: should we consider using [`std::path::absolute`](https://doc.rust-lang.org/stable/std/path/fn.absolute.html) instead of `canonicalize`?

`absolute` seems simpler than `canonicalize`. E.g., it does not require FS access and it does not require path to exist. And personally, I strictly prefer "logical" path resolution over "physical" (what `canonicalize` does, as far as I understand). Also, _I think_, that logical resolution is less prone to "chroot-escape-like" exploits.

@timvisee @mac-chaffee do you guys have any thoughts on this, or am I overthinking?

timvisee:
Cool. I wasn't aware [`absolute`](https://doc.rust-lang.org/stable/std/path/fn.absolute.html) was added recently.

But, I do not agree. `absolute` does not resolve symlinks, which means its not reliable for actual directory checking. Having a symlink inside your snapshots directory would allow you to break out. [`canonicalize`](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) does not suffer from this problem.

Now, there may be a case where we can use [`absolute`](https://doc.rust-lang.org/stable/std/path/fn.absolute.html) if there is no security concern as it is a bit cheaper, but I don't think that's important here.

timvisee:
Superseded by <https://github.com/qdrant/qdrant/pull/8341>

PR #4578, opened 2024-06-27
Fail early when encountering out-of-storage during optimization

xhjkl:
/claim #4297

### All Submissions:

What this does is checks if there is twice as much free space on the volume as what the collection being optimized already takes.

I'm not happy with this, and I'm not sure this submission worth the hassle as it is, but I'd like to compare opinions.

## Why it's bad
* The free space checking works synchronously; even though the optimizations themselves take up background threads, I feel uneasy about it.
* The checking happens in the moment the optimization is about to start, not well beforehand.
* We still might run into OOD down the line because FS is non-atomic.
* It still logs the error, perhaps we want to special-case it as we do with `CollectionError::Cancelled`.

## What else have I tried
* Capture write error from `db.put_cf_opt`; it first happens somewhere around `lib/segment/src/segment_constructor/segment_builder.rs:197`, but I could not find a way to cleanly rollback a partially done optimization.
* Utilize the existing `DiskUsageWatcher` to do the checking before the optimization starts, but it does not have a dynamic threshold, so far as I understand it, and overall it gets a larger change that doing how it's done now.
* Add a method `check_environment` to the trait `SegmentOptimizer` in the same spirit as the existing `SegmentOptimizer::check_condition` so that all the implementors could tailor it to their taste, but for now this approach adds no value.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5310,6 +5310,7 @@ dependencies = [
  "criterion",
  "dataset",
  "fnv",
+ "fs4",
  "fs_extra",
  "generic-tests",
  "geo",
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -4,10 +4,10 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use common::cpu::CpuPermit;
+use common::disk::dir_size;
 use io::storage_version::StorageVersion;
 use itertools::Itertools;
 use parking_lot::{Mutex, RwLock, RwLockUpgradableReadGuard};
-use segment::common::dir_size;
 use segment::common::operation_error::check_process_stopped;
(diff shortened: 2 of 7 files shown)


xhjkl:
@generall, @timvisee, I've rebased my changes upon the branch with the test — the test seems to pass now.
Also, I dropped all the styling changes so that the diff is minimal now; hope this helps.

generall:
> The free space checking works synchronously; even though the optimizations themselves take up background threads, I feel uneasy about it.

I don't see problem with it. Async FS is fake anyway.

> We still might run into OOD down the line because FS is non-atomic.

There are no guarantees about it indeed, some external process can occupy the disk and we can do nothing about it

generall (approved):

xhjkl:
> 
> I don't see problem with it. Async FS is fake anyway.
> 
Understood. Thank you for commenting on that 🙏

> 
> There are no guarantees about it indeed, some external process can occupy the disk and we can do nothing about it
>
Yeah, that's my thoughts precisely: that's why I made it `2 * x`, just do be doubly sure.

PR #4580, opened 2024-06-27
order_by: Order by value, then id. Dedup by both

coszio:
Once implementing resharding deduplication, we introduced deduping by id. However, this broke one behavior for `order_by` in scroll: to repeat points if they had multiple values to order from.

With these changes now we output consistent (value, id) order, and we de-duplicate on the combination too.

### DIFF
--- a/lib/collection/src/collection/point_ops.rs
+++ b/lib/collection/src/collection/point_ops.rs
@@ -319,12 +319,16 @@ impl Collection {
                         })
                     })
                     // Get top results
-                    .kmerge_by(|(value_a, _), (value_b, _)| match order_by.direction() {
-                        Direction::Asc => value_a <= value_b,
-                        Direction::Desc => value_a >= value_b,
+                    .kmerge_by(|(value_a, record_a), (value_b, record_b)| {
+                        match order_by.direction() {
+                            Direction::Asc => (value_a, record_a.id) < (value_b, record_b.id),
+                            Direction::Desc => (value_a, record_a.id) > (value_b, record_b.id),
+                        }
                     })
-                    // Add each point only once, deduplicate point IDs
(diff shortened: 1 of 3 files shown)


timvisee:
Note: we want this because we've documented this behavior:

> Note: for payloads with more than one value (such as arrays), the same point may show up more than once. Each point can appear as many times as the number of elements in the array. For example, if you have a point payload with a timestamp key, and the value for the key is an array of 3 elements, the same point will appear 3 times in the results, one for each timestamp.

via: https://qdrant.tech/documentation/concepts/points/#order-points-by-payload-key

timvisee (approved):
Thanks!

timvisee:
> (...) if there are any other reason to include point multiple times

Currently only because we've documented it that way.

Which do you prefer?
- deduplicate point IDs and update the documentation
- or deduplicate point ID+value pairs like in this PR

generall (approved):

timvisee:
@coszio In a call we decided to do deduplication on just point IDs, and break the old behavior. This PR has been updated accordingly.

lib/collection/src/collection/point_ops.rs
```diff
@@ -325,10 +325,8 @@ impl Collection {
                             Direction::Desc => (value_a, record_a.id) > (value_b, record_b.id),
                         }
                     })
-                    // Add each point only once, deduplicate on same (value, IDs) combination
-                    .dedup_by(|(value_a, record_a), (value_b, record_b)| {
-                        (value_a, record_a.id) == (value_b, record_b.id)
-                    })
+                    // Only keep the point with the most "valuable" order value
```
coszio, line 328: So, only repeat when there are interleaving values between multiple same ids?

PR #4581, opened 2024-06-28
Add logs for JWT RBAC configuration

superseb:
### All Submissions:

### New Feature Submissions:

I ran into these while checking out JWT, let me know if this is something we want to add (or not) and what should be changed according to how we do things. Also Rust noob so open to any suggestions how to do things differently.

- When JWT RBAC is enabled but there is no API key set, without at least an error, user does not know why it does not work. I added logging an error cause we are not failing at this currently, but we can change if we want it to fail. At least with this change the user can see from the log something is not going to work.
- Recommended secret size for HMAC-SHA256 is 32 bytes, I added logging a recommendation if JWT RBAC is enabled and the secret size is less than 32 bytes. Reference: https://www.rfc-editor.org/rfc/rfc7518#section-3.2 ```A key of the same size as the hash output (for instance, 256 bits for
   "HS256") or larger MUST be used with this algorithm.```. Larger seems to be unneeded but we can change the wording to "at least" if wanted.


src/main.rs
```diff
@@ -397,6 +397,33 @@ fn main() -> anyhow::Result<()> {
...
+        // Log if JWT RAC is enabled, API key is set but smaller than recommended size for JWT secret
+        } else if settings.service.api_key.clone().unwrap_or_default().len()
+            < JWT_RECOMMENDED_SECRET_LENGTH
+        {
+            log::warn!(
+                "It is highly recommended to use an API key of {} bytes when JWT RBAC is enabled",
+                JWT_RECOMMENDED_SECRET_LENGTH
+            )
+        }
+    }
```
generall, line 425: It looks like this check only depends on `settings`. Might be better to move it to `validate_and_warn`

superseb: I moved the code, let me know if it needs to be wrapped in a function or in some other way.

src/settings.rs
```diff
@@ -219,6 +219,27 @@ impl Settings {
...
+        //
+        // JWT RBAC
+        //
+        // Using HMAC-SHA256, recommended secret size is 32 bytes
+        const JWT_RECOMMENDED_SECRET_LENGTH: usize = 256 / 8;
+
+        // Log if JWT RBAC is enabled but no API key is set
+        if self.service.jwt_rbac.unwrap_or_default() {
+            if self.service.api_key.clone().unwrap_or_default().is_empty() {
+                log::error!("JWT RBAC enabled but no API key set")
```
timvisee, line 231: I'd prefer to demote this to a `warn` instead, they both have the same visibility in logs anyway.

```suggestion
                log::warn!("JWT RBAC enabled but no API key set")
```

timvisee (approved):

src/settings.rs
superseb: Ok, the reason I used error here is that the requested functionality (enabling JWT RBAC) is not applied so users might complain that something is not working but there are no errors in the log.

superseb: Updated to `warn` and changed wording a bit to make it more clear to the user.

coszio (approved):

superseb:
@timvisee Is it okay for me to merge this or do you coordinate this? If this is no longer needed/wanted, it can also be closed.

timvisee:
> @timvisee Is it okay for me to merge this or do you coordinate this? If this is no longer needed/wanted, it can also be closed.

Since we approved, feel free to merge! 

I'll take the opportunity now :smile:

PR #4583, opened 2024-06-28
Atomic create_and_ensure_length

IvanPleshkov:
There is a possibility to fails with file creation. Crash may happen between file creation and settings size. Solution: create temporary file and rename when ready

### DIFF
--- a/lib/common/memory/src/mmap_ops.rs
+++ b/lib/common/memory/src/mmap_ops.rs
@@ -17,7 +17,7 @@ pub fn create_and_ensure_length(path: &Path, length: usize) -> io::Result<File> 
         let file = OpenOptions::new()
             .read(true)
             .write(true)
-            .create(true)
+            .create(false)
             // Don't truncate because we explicitly set the length later
             .truncate(false)
             .open(path)?;
--- a/lib/common/memory/src/mmap_ops.rs
+++ b/lib/common/memory/src/mmap_ops.rs
@@ -26,14 +26,18 @@ pub fn create_and_ensure_length(path: &Path, length: usize) -> io::Result<File> 
         Ok(file)
     } else {
         let temp_path = path.with_extension(TEMP_FILE_EXTENSION);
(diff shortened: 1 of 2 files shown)


lib/segment/src/vector_storage/chunked_utils.rs
```diff
@@ -26,7 +27,8 @@ pub fn read_mmaps<T: Sized>(directory: &Path) -> OperationResult<Vec<MmapChunk<T
     for entry in directory.read_dir()? {
         let entry = entry?;
         let path = entry.path();
-        if path.is_file() {
+        if path.is_file() && path.extension().map(OsStr::to_str) != Some(Some(TEMP_FILE_EXTENSION))
```
xzfc, line 30: Nit: maybe we need to clean up such resides?

Also, `.map(…) == Some(Some(…))` could be rewritten as `.and_then(…) == Some(…)`.

xzfc: Also, note that later in this code we check that the name conforms the format of `chunk_<number>.mapp`, so checking for `.tmp` is not necessary.

lib/common/memory/src/mmap_ops.rs
```diff
@@ -10,17 +10,27 @@ use memmap2::{Mmap, MmapMut};
...
         .write(true)
         .create(true)
         // Don't truncate because we explicitly set the length later
         .truncate(false)
-        .open(path)?;
+        .open(&temp_path)?;
     file.set_len(length as u64)?;
 
-    Ok(file)
+    std::fs::rename(&temp_path, path)?;
```
generall, line 26: potentially we might want to flush and destroy file before rename

lib/segment/src/vector_storage/chunked_utils.rs
IvanPleshkov: Your right, reverted this file changes

generall (approved):

PR #4601, opened 2024-07-02
Fusions: add score fusion internals

coszio:
Adds an internal method for making parametrized score fusions, currently with:

- Normalizations
  - MinMax
  - Distribution-based
- Combinators
  -  Sum
- Weighting
- Ordering

### DIFF
--- a/lib/segment/src/common/score_fusion.rs
+++ b/lib/segment/src/common/score_fusion.rs
@@ -51,14 +51,13 @@ pub fn score_fusion(
         })
         // weight each list of points
         .zip(weights)
-        .map(|(points, weight)| {
+        .flat_map(|(points, weight)| {
             points.into_iter().map(move |p| ScoredPoint {
                 score: p.score * weight,
                 ..p
             })
         })
         // combine to deduplicate
-        .flatten()
         .fold(
             HashMap::<PointIdType, ScoredPoint>::new(),
             |mut acc, point| {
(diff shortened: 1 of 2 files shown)


lib/segment/src/common/score_fusion.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub method: CombMethod,
+    /// Defines how to normalize the scores in each list
+    pub norm: Normalization,
+    /// Multipliers for each list of scores
+    pub weights: Vec<f32>,
+    /// Final ordering of the results
+    pub order: Order,
+}
+
+pub enum CombMethod {
```
agourlay, line 21: `CombMethod` is rather exotic, what about `Aggregation` instead?

lib/segment/src/common/score_fusion.rs
```diff
@@ -0,0 +1,186 @@
...
+            Order::SmallBetter => a.cmp(b),
+            Order::LargeBetter => b.cmp(a),
+        })
+        .collect()
+}
+
+/// Normalizes the scores of the given points between 0.0 and 1.0, using the given minimum and maximum scores as extremes.
+fn norm(mut points: Vec<ScoredPoint>, min: ScoreType, max: ScoreType) -> Vec<ScoredPoint> {
+    points.iter_mut().for_each(|p| {
+        p.score = (p.score - min) / (max - min);
```
agourlay, line 85: does it make sense to add protection against dividing by zero here?

lib/segment/src/common/score_fusion.rs
coszio: I didn't think too deep about it, but `Aggregation` makes sense. I was just following the naming used in the resources I read

agourlay (approved):

PR #4605, opened 2024-07-03
Promote log dep to workspace

agourlay:
The `log` crate is a good candidate for promotion.


lib/sparse/Cargo.toml
```diff
@@ -27,7 +27,6 @@ ordered-float = "4.2"
 rand = "0.8.5"
 validator = { workspace = true }
 itertools = { workspace = true }
-log = "0.4"
```
agourlay, line 30: removed because it was unused :broom:

timvisee (approved):

PR #4608, opened 2024-07-03
Fix resharding local transfer leaving proxy

timvisee:
Tracked in: #4213 

While migration points as part of resharding we might do a 'local transfer' if the source and target shard are on the same node.

This logic did not clean up the forward proxy that is created. This PR fixes that.

Normally consensus takes care of this for us, but we don't use consensus here.

This also fixes randomly deleting points from the newly created resharded-shard during the delete propagation stage.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/resharding/driver.rs
+++ b/lib/collection/src/shards/resharding/driver.rs
@@ -531,13 +531,25 @@ async fn migrate_local(
 
     let progress = Arc::new(Mutex::new(TransferTaskProgress::new()));
     transfer_resharding_stream_records(
-        shard_holder,
+        Arc::clone(&shard_holder),
         progress,
         source_shard_id,
         target_shard,
         collection_id,
     )
     .await?;
+
+    // Unproxify forward proxy on local shard we just transferred
+    // Normally consensus takes care of this, but we don't use consensus here
+    {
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/resharding/driver.rs
```diff
@@ -531,14 +531,26 @@ async fn migrate_local(
 
     let progress = Arc::new(Mutex::new(TransferTaskProgress::new()));
     transfer_resharding_stream_records(
-        shard_holder,
+        Arc::clone(&shard_holder),
```
ffuugoo, line 534:
```suggestion
        shard_holder.clone(),
```

ffuugoo (approved):
LGTM, but what if `transfer_resharding_stream_records` fails? Should we unproxify on error as well?

lib/collection/src/shards/resharding/driver.rs
timvisee: Good call. There are other bits in code that'll recover it. But it won't hurt to be safe here as well.

I'll add it and will merge after.

timvisee: 0e4b2a794512d2397a4d214eab9782e1d464d542

PR #4611, opened 2024-07-03
universal-query: resolve offset at collection level only

coszio:
Fixes a bug with offset, where it was being applied twice. 

Adds a basic test that was failing before the fix


generall (approved):

agourlay (approved):

agourlay:
@coszio thanks for the fix :clap: 

Please double check that the local mode is correct.

coszio:
Double checked, local mode works fine

Issue #4502, opened 2024-06-19
[On disk payload index] Tracking issue


IvanPleshkov:
This issue tracks the necessary tasks to integrate the on-disk-payload-index feature into the engine.

## Keyword Index
- [ ] Mmap multi hash map
- [ ] Mmap keyword field index enum variant

## Numeric Index
- [ ] Persisted histogram
- [ ] Mmap numeric field index enum variant

## Boolean Index
- [ ] Mmap boolean field index enum variant

## Geo Index
- [ ] Mmap geo field index enum variant

## Integration
- [ ] Propagate on-disk flag into payload index
- [ ] REST API
- [ ] gRPC API

PR #4612, opened 2024-07-03
Add MmapHashMap

xzfc:
Tracking issue: #4502.
This PR adds `MmapHashMap`, an on-disk hash map baked by a memory-mapped file.

Some design decisions:
- Use `zerocopy` as a sound/safe alternative to our [`mmap_ops`](https://github.com/qdrant/qdrant/blob/v1.10.0/lib/common/memory/src/mmap_ops.rs#L124-L188).
- Put the implementation into the `lib/common/common` crate. It wouldn't make compilation of this crate noticeable slower.
- Make `MmapHashMap::create()` take `BTreeMap<String, BTreeSet<u32>>`, not `impl Iterator<…>`. Since this structure would be used in a single place (`segment`), we could hardcode it to take a specific data structure for better compile times.

### DIFF
--- a/lib/common/common/src/mmap_hashmap.rs
+++ b/lib/common/common/src/mmap_hashmap.rs
@@ -1,0 +1,233 @@
+use std::collections::{BTreeMap, BTreeSet};
+use std::fs::File;
+use std::io::{self, Cursor, Write as _};
+use std::mem::size_of;
+use std::path::Path;
+use std::str;
+
+use memmap2::Mmap;
+use ph::fmph::Function;
+#[cfg(test)]
+use rand::rngs::StdRng;
+#[cfg(test)]
+use rand::{Rng as _, SeedableRng as _};
+use zerocopy::{AsBytes, FromBytes, FromZeroes};
+
+/// On-disk hash map baked by a memory-mapped file.
+pub struct MmapHashMap {
+    mmap: Mmap,
+    header: Header,
+    phf: Function,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone, Debug, AsBytes, FromBytes, FromZeroes)]
+struct Header {
+    buckets_pos: u64,
+    buckets_count: u64,
+}
+
+const PADDING_SIZE: usize = 4096;
+
+impl MmapHashMap {
+    /// Save `map` contents to `path`.
+    pub fn create(path: &Path, map: BTreeMap<String, BTreeSet<u32>>) -> io::Result<()> {
+        let phf = Function::from(map.keys().collect::<Vec<_>>().as_slice());
+
+        // == First pass ==
+
+        let mut file_size = 0;
+        // 1. Header
+        file_size += size_of::<Header>();
+
+        // 2. PHF
+        file_size += phf.write_bytes();
+
+        // 3. Padding
+        let padding_len = (PADDING_SIZE - (file_size % PADDING_SIZE)) % PADDING_SIZE;
+        file_size += padding_len;
+
+        // 4. Buckets
+        let buckets_pos = file_size;
+        file_size += map.len() * size_of::<u64>();
+
+        // 5. Data
+        let mut buckets = vec![0u64; map.len()];
+        let mut last_bucket = 0;
+        for (k, v) in map.iter() {
+            buckets[phf.get(k).expect("Key not found in phf") as usize] = last_bucket as u64;
+            last_bucket += entry_size(k, v);
+        }
+        file_size += last_bucket;
+        _ = file_size;
+
+        // == Second pass ==
+
+        let mut bufw = io::BufWriter::new(File::create(path)?);
+
+        // 1. Header
+        let header = Header {
+            buckets_pos: buckets_pos as u64,
+            buckets_count: map.len() as u64,
+        };
+        bufw.write_all(header.as_bytes())?;
+
+        // 2. PHF
+        phf.write(&mut bufw)?;
+
+        // 3. Padding
+        bufw.write_all(zeroes(padding_len))?;
+
+        // 4. Buckets
+        bufw.write_all(buckets.as_bytes())?;
+
+        // 5. Data
+        let mut pos = 0;
+        for (k, v) in map.iter() {
+            buckets.push(pos as u64);
+            pos += entry_size(k, v);
+
+            // Lengths
+            bufw.write_all((k.len() as u64).as_bytes())?;
+            bufw.write_all((v.len() as u64).as_bytes())?;
+
+            // Key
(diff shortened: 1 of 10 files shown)


lib/common/common/src/mmap_hashmap.rs
```diff
@@ -0,0 +1,233 @@
...
+        for (k, v) in map.iter() {
+            buckets[phf.get(k).expect("Key not found in phf") as usize] = last_bucket as u64;
+            last_bucket += entry_size(k, v);
+        }
+        file_size += last_bucket;
+        _ = file_size;
+
+        // == Second pass ==
+
+        let mut bufw = io::BufWriter::new(File::create(path)?);
```
IvanPleshkov, line 66: We calculated size already. I propose to create file using `create_and_ensure_length` which creates a file with provided size

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -0,0 +1,233 @@
...
+struct Header {
+    buckets_pos: u64,
+    buckets_count: u64,
+}
+
+const PADDING_SIZE: usize = 4096;
+
+impl MmapHashMap {
+    /// Save `map` contents to `path`.
+    pub fn create(path: &Path, map: BTreeMap<String, BTreeSet<u32>>) -> io::Result<()> {
```
IvanPleshkov, line 34: Can we implement hashmap generic over key? Map field index is defined for `String` and for `IntPayloadType`

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -0,0 +1,233 @@
...
+        ))?;
+
+        Ok(MmapHashMap { mmap, header, phf })
+    }
+
+    /// Get the values associated with `key`. It will return `None` in the following cases:
+    /// - The key is not present in this hash map.
+    /// - The data is corrupted, e.g. invalid UTF-8 or wrong alignment. (should not happen during
+    ///   normal operation)
+    pub fn get(&self, key: &str) -> Option<&[u32]> {
```
IvanPleshkov, line 132: Corrupted storage is not the same as unexist key. I propose to add error handling and return here `OperationResult<Option<&[u32]>>`

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -0,0 +1,233 @@
...
+
+fn entry_size(key: &str, values: &BTreeSet<u32>) -> usize {
+    2 * size_of::<u64>()
+        + key.as_bytes().len().next_multiple_of(4)
+        + values.len() * size_of::<u32>()
+}
+
+/// Returns a reference to a slice of zeroes of length `len`.
+#[inline]
+fn zeroes(len: usize) -> &'static [u8] {
```
IvanPleshkov, line 177: Because we use this and other methods only for tests in this file, I propose to move `zeroes`, `random_map`, `rand_ident` into `mod tests` scope

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -0,0 +1,233 @@
...
+        let hash = self.phf.get(key)?;
+
+        let buckets = u64::slice_from(self.mmap.get(
+            self.header.buckets_pos as usize
+                ..self.header.buckets_pos as usize
+                    + self.header.buckets_count as usize * size_of::<u64>(),
+        )?)?;
+        let bucket_val = *buckets.get(hash as usize)?;
+
+        let entry = &self.mmap.get(
```
IvanPleshkov, line 142: Interesting that Clippy is happy. We don't need `&` here, `get` returns already a slice

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        let key_len = u64::read_from(entry.get(0..size_of::<u64>())?)? as usize;
+        let values_len =
+            u64::read_from(entry.get(size_of::<u64>()..2 * size_of::<u64>())?)? as usize;
+
+        let hash_key =
+            str::from_utf8(entry.get(2 * size_of::<u64>()..2 * size_of::<u64>() + key_len)?)
+                .ok()?;
+
+        if hash_key != key {
```
IvanPleshkov, line 156: I propose to avoid the utf8 conversion check and compare raw data from file with `key.as_bytes()`

lib/common/common/src/mmap_hashmap.rs
IvanPleshkov, line 34: Why `io::Result` instead of our internal `OperationResult`?

lib/common/common/src/mmap_hashmap.rs
xzfc: `zeroes` are used in the implementation, not tests.
As for `random_map`, `random_ident`, I have plans to use them later in the benchmarks, meaning that they should not be inside `mod tests`.

IvanPleshkov: Okay, resolved

lib/common/common/src/mmap_hashmap.rs
xzfc: Yep, I have a plan to make it generic and store
1. For strings, `size: u64` + `contents: [u8; size]`. (implemented in this PR)
2. For ints and other fixed-size values, just `data: T`.

But haven't decided a best way to implement it (either by introducing a trait, or maybe using some existing crate). I want to postpone it for now.

xzfc: `OperationResult` is defined in the `segment` crate. So, I've used `io::Result`, just like in the `sparse` crate.

IvanPleshkov: Okay, for me it's fine to do it later and make this PR simpler, resolved

IvanPleshkov: got it!

lib/common/common/src/mmap_hashmap.rs
IvanPleshkov: Thanks!

lib/common/common/src/mmap_hashmap.rs
IvanPleshkov: Thanks!

lib/common/common/src/mmap_hashmap.rs
IvanPleshkov: Thanks!

IvanPleshkov (approved):

lib/common/common/src/mmap_hashmap.rs
xzfc: #4648

PR #4616, opened 2024-07-04
universal-query: Grouping REST API

agourlay:
This PR adds a new endpoint for a grouping feature on the query API.

I decided to reuse the existing `grouping` infrastructure which is well tested and to adapt it to the query request.

This means the existing group search will be executed by the query runner.

This approach enables to leverage the existing tests to validate the equivalence of search vs query when grouping.

The corresponding gRPC API will be added in subsequent PR.

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -12169,9 +12169,7 @@
       "QueryGroupsRequest": {
         "type": "object",
         "required": [
-          "group_by",
-          "group_size",
-          "limit"
+          "group_by"
         ],
         "properties": {
           "shard_key": {
@@ -12274,16 +12272,18 @@
             "minLength": 1
           },
           "group_size": {
-            "description": "Maximum amount of points to return per group",
+            "description": "Maximum amount of points to return per group. Default is 3.",
             "type": "integer",
-            "format": "uint32",
-            "minimum": 1
+            "format": "uint",
+            "minimum": 1,
+            "nullable": true
           },
           "limit": {
-            "description": "Maximum amount of groups to return",
+            "description": "Maximum amount of groups to return. Default is 10.",
             "type": "integer",
-            "format": "uint32",
-            "minimum": 1
+            "format": "uint",
+            "minimum": 1,
+            "nullable": true
           },
           "with_lookup": {
             "description": "Look for points in another collection using the group ids",
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -5342,6 +5342,101 @@
           }
         }
       }
+    },
+    "/collections/{collection_name}/points/query/groups": {
+      "post": {
+        "tags": [
+          "points"
+        ],
+        "summary": "Query points, grouped by a given payload field",
+        "description": "Universally query points, grouped by a given payload field",
+        "operationId": "query_points_groups",
+        "requestBody": {
+          "description": "Describes the query to make to the collection",
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/QueryGroupsRequest"
+              }
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "collection_name",
+            "in": "path",
+            "description": "Name of the collection to query",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "consistency",
+            "in": "query",
+            "description": "Define read consistency guarantees for the operation",
+            "required": false,
+            "schema": {
+              "$ref": "#/components/schemas/ReadConsistency"
+            }
+          },
+          {
+            "name": "timeout",
+            "in": "query",
+            "description": "If set, overrides global timeout for this request. Unit is seconds.",
+            "required": false,
+            "schema": {
+              "type": "integer",
+              "minimum": 1
+            }
+          }
+        ],
+        "responses": {
+          "default": {
+            "description": "error",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ErrorResponse"
+                }
+              }
+            }
+          },
+          "4XX": {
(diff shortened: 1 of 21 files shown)


lib/collection/src/recommendations.rs
```diff
@@ -172,6 +173,16 @@ pub fn recommend_into_core_search(
     request: RecommendRequestInternal,
     all_vectors_records_map: &ReferencedVectors,
 ) -> CollectionResult<CoreSearchRequest> {
+    let query_request =
+        recommend_into_query_search(collection_name, request, all_vectors_records_map)?;
+    CoreSearchRequest::try_from(query_request)
+}
+
+pub fn recommend_into_query_search(
```
coszio, line 181: This implementation means that for ungrouped recommend, we will transform from recommend -> query -> core. 

Wouldn't it be an easier mental model to not touch the recommend endpoint, and for groups, we can convert from recommend -> core -> query?

lib/collection/src/operations/universal_query/collection_query.rs
```diff
@@ -20,7 +20,7 @@ use crate::operations::types::{CollectionError, CollectionResult};
 use crate::recommendations::avg_vector_for_recommendation;
 
 /// Internal representation of a query request, used to converge from REST and gRPC. This can have IDs referencing vectors.
-#[derive(Debug)]
+#[derive(Clone, Debug, PartialEq)]
```
agourlay, line 23: `Clone` + `PartialEq` is a requirement of our test logic for grouping, not for actual runtime.

tests/consensus_tests/auth_tests/test_jwt_access.py
```diff
@@ -530,6 +530,13 @@ def __init__(self, r, coll_rw, m, rest_endpoint, grpc_endpoint=None, **kwargs):
     "query_batch_points": EndpointAccess(
         True, True, True, "POST /collections/{collection_name}/points/query/batch", "qdrant.Points/QueryBatch"
     ),
+    "query_points_groups": EndpointAccess(
+        True,
+        True,
+        True,
+        "POST /collections/{collection_name}/points/query/groups",
+        #"qdrant.Points/QueryGroups",
```
agourlay, line 538: for the next PR :)

docs/redoc/master/openapi.json
```diff
@@ -11970,6 +12065,138 @@
             }
           }
         }
+      },
+      "QueryGroupsRequest": {
+        "type": "object",
+        "required": [
+          "group_by",
+          "group_size",
+          "limit"
```
coszio, line 12074: Let's keep the required params minimal., in this case only require to set `group_by`. Some reasonable defaults would be limit=10 and group_size=3, wdyt?

lib/collection/src/grouping/group_by.rs
```diff
@@ -224,21 +249,68 @@ impl From<RecommendGroupsRequestInternal> for GroupRequest {
...
+
+    if let Some(query) = &request.source.query {
+        if query.is_distance_scored() {
+            if let Some(vector_name) = query.get_vector_name() {
+                let collection_params = collection.collection_config.read().await;
+                let distance = collection_params.params.get_distance(vector_name)?;
+                score_ordering = distance.distance_order()
+            }
+        }
     };
```
coszio, line 314: We also need to consider `order_by` direction. There is a [`ScoringQuery::order`](https://github.com/qdrant/qdrant/blob/dev/lib/collection/src/operations/universal_query/shard_query.rs#L83) function which can be used in here :)

lib/collection/src/operations/universal_query/collection_query.rs
coszio: If you would like to [compile it conditionally](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute), it would look something like this
```suggestion
#[derive(Debug)]
#[cfg_attr(test, derive(Clone, PartialEq)]
```

But I think the extra derives are small enough to not even bother

lib/collection/src/operations/universal_query/shard_query.rs
```diff
@@ -79,6 +79,14 @@ impl ScoringQuery {
...
     }
 
+    /// Only when the distance is the scoring, this will return true.
+    pub fn is_distance_scored(&self) -> bool {
+        match self {
+            ScoringQuery::Vector(query) => query.is_distance_scored(),
+            ScoringQuery::Fusion(_) | ScoringQuery::OrderBy(_) => false,
+        }
+    }
+
```
coszio, line 89: > https://github.com/qdrant/qdrant/pull/4616/files#r1667088059

we can avoid adding this function

tests/openapi/openapi_integration/test_query_full.py
coszio, line 1: I would love to see at least one case per scoring query type:
- nearest ✅ 
- recommend avg
- recommend best_score
- discovery (I think it is fine to not test context)
- order_by asc and desc

lib/collection/src/grouping/aggregator.rs
```diff
@@ -95,20 +94,24 @@ impl GroupsAggregator {
...
             // Insert score if better than the group best score
-            self.group_best_scores
+            self.groups_best_point
                 .entry(group_key.clone())
-                .and_modify(|e| {
-                    *e = match self.order {
-                        Order::LargeBetter => point.score.max(*e),
-                        Order::SmallBetter => point.score.min(*e),
+                .and_modify(|p_id| {
+                    let other_score: &ScoredPoint = self.groups[&group_key].get(p_id).unwrap();
```
coszio, line 100: Having to dereference the point id from the groups and then unwrapping makes the logic to be more complicated than it should, IMO. 

If we instead store the actual `ScoredPoint` as a value in the `groups_best_point` hashmap, we won't need any dereferencing, and it wouldn't hurt memory too much, since it is only one per group.

coszio (approved):

PR #4619, opened 2024-07-05
random-query: Implement random reads from segments up to local shard

coszio:
Implements reading randomly from segments.

The way it reads is not entirely optimal for large limits (it has rejection sampling to avoid repetition), but since we will not promise stable random output, offset won't affect the limit. In other words, it should be "fine" for most use-cases.

The other thing changed here is that I've edited the `QueryScrollRequestInternal` so that ordering is always specified, as by id, by field, or random.

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -663,6 +663,34 @@ impl SegmentEntry for ProxySegment {
         Ok(read_points)
     }
 
+    fn read_random_filtered<'a>(
+        &'a self,
+        limit: usize,
+        filter: Option<&'a Filter>,
+    ) -> Vec<PointIdType> {
+        let deleted_points = self.deleted_points.read();
+        let mut read_points = if deleted_points.is_empty() {
+            self.wrapped_segment
+                .get()
+                .read()
+                .read_random_filtered(limit, filter)
+        } else {
+            let wrapped_filter =
+                self.add_deleted_points_condition_to_filter(filter, &deleted_points);
+            self.wrapped_segment
+                .get()
+                .read()
+                .read_random_filtered(limit, Some(&wrapped_filter))
+        };
+        let mut write_segment_points = self
+            .write_segment
+            .get()
+            .read()
+            .read_random_filtered(limit, filter);
+        read_points.append(&mut write_segment_points);
+        read_points
+    }
+
     /// Read points in [from; to) range
     fn read_range(&self, from: Option<PointIdType>, to: Option<PointIdType>) -> Vec<PointIdType> {
         let deleted_points = self.deleted_points.read();
--- a/lib/collection/src/shards/local_shard/scroll.rs
+++ b/lib/collection/src/shards/local_shard/scroll.rs
@@ -3,6 +3,7 @@ use std::time::Duration;
 
 use futures::future::try_join_all;
 use itertools::Itertools as _;
+use rand::seq::IteratorRandom;
 use segment::data_types::order_by::{Direction, OrderBy, OrderValue};
 use segment::types::{
     ExtendedPointId, Filter, ScoredPoint, WithPayload, WithPayloadInterface, WithVector,
@@ -13,7 +14,7 @@ use super::LocalShard;
 use crate::collection_manager::holders::segment_holder::LockedSegment;
 use crate::collection_manager::segments_searcher::SegmentsSearcher;
 use crate::operations::types::{
-    CollectionError, CollectionResult, QueryScrollRequestInternal, Record,
+    CollectionError, CollectionResult, QueryScrollRequestInternal, Record, ScrollOrder,
 };
 
 impl LocalShard {
@@ -51,7 +52,7 @@ impl LocalShard {
             limit,
             with_vector,
             filter,
-            order_by,
+            scroll_order,
             with_payload,
         } = request;
 
@@ -59,10 +60,8 @@ impl LocalShard {
 
         let offset_id = None;
 
-        let order_by = order_by.clone().map(OrderBy::from);
-
-        let point_results = match order_by {
-            None => self
+        let point_results = match scroll_order {
+            ScrollOrder::ById => self
                 .scroll_by_id(
                     offset_id,
                     limit,
@@ -83,7 +82,7 @@ impl LocalShard {
                     order_value: None,
                 })
                 .collect(),
-            Some(order_by) => {
+            ScrollOrder::ByField(order_by) => {
                 let (records, values) = self
                     .scroll_by_field(
                         limit,
@@ -91,7 +90,7 @@ impl LocalShard {
                         with_vector,
                         filter.as_ref(),
                         search_runtime_handle,
-                        &order_by,
+                        order_by,
                     )
                     .await?;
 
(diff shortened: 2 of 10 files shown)


lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -662,6 +662,34 @@ impl SegmentEntry for ProxySegment {
         Ok(read_points)
     }
 
+    fn read_random_filtered<'a>(
```
timvisee, line 665: It is very similar to the other two read implementations. I wonder if it would make sense to merge them.

I assume you didn't merge it with `read_ordered_filtered` using a 'Random` order variant because this one is not ordered nor deterministic at all?

lib/segment/src/fixtures/payload_context_fixture.rs
```diff
@@ -119,6 +119,14 @@ impl IdTracker for FixtureIdTracker {
         )
     }
 
+    /// Not actually random, but it's enough for tests
+    fn iter_random(&self) -> Box<dyn Iterator<Item = (PointIdType, PointOffsetType)> + '_> {
+        Box::new(
+            self.iter_ids()
+                .map(|id| (PointIdType::NumId(id as u64), id)),
+        )
+    }
```
timvisee, line 128: Why not proper random?

lib/segment/src/id_tracker/simple_id_tracker.rs
```diff
@@ -343,6 +345,28 @@ impl IdTracker for SimpleIdTracker {
...
+        let rng = rand::thread_rng();
+        let available = self.internal_to_external.len();
+        let uniform = rand::distributions::Uniform::new(0, available);
+        let iter = Distribution::sample_iter(uniform, rng)
+            // TODO: this is not efficient if `available` is large and we iterate over most of them,
+            // but it's good enough for low limits.
+            //
+            // We could improve it by using a variable-period PRNG to adjust depending on the number of available points.
+            .unique()
+            .take(available)
```
timvisee, line 358: A simple alternative may be, for each iteration:
- generate a random number
- if we haven't seen it yet, use it
- if we did see it, pick the next unseen number

It'll be faster, at the cost of being a bit less random.

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -219,4 +242,47 @@ impl LocalShard {
...
+                    .read()
+                    .read_random_filtered(limit, filter.as_ref())
+            })
+        };
+
+        let all_reads = try_join_all(
+            non_appendable
+                .into_iter()
+                .map(read_filtered)
+                .chain(appendable.into_iter().map(read_filtered)),
```
timvisee, line 273: We can probably map once after the chain.

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -219,4 +242,47 @@ impl LocalShard {
...
+        )
+        .await?;
+
+        let point_ids = all_reads
+            .into_iter()
+            .flatten()
+            .choose_multiple(&mut rand::thread_rng(), limit);
```
timvisee, line 280: IIRC, though this works, it can be very expensive as it might evaluate all possible values in the iterator. I must admit that I may be wrong here, as the details on it's implementation are quite confusing.

I wonder if the following would be better: create a list of all random readers, so you have one for each segment. Fill our final list by constantly picking from a random reader. This will break uniformity though with segments that don't have equal size.

Also, we should deduplicate here. It is possible for different segments to have the same point ID. Ideally we'd only take the one with the highest version, but I don't see how we can do that here.

Do you have any thoughts on this based on the above?

lib/segment/src/fixtures/payload_context_fixture.rs
coszio: tbh, laziness, but I can just copy the implementation from the id-tracker :)

lib/segment/src/id_tracker/simple_id_tracker.rs
coszio: that's exactly what this is doing. The thing I am pointing out is that, let's say, the segment has 10k points, and you request a limit of 10k too. For finding the last number, it would have a 1/10k probability of having a random number which we haven't seen before. 

Ideally the best algorithm would be one which is pseudo-random, and doesn't repeat values until all of them have appeared once.

lib/collection/src/shards/local_shard/scroll.rs
coszio: Thank you, I did forget to deduplicate 👍.

Implementation of `choose_multiple` is indeed `O(n)`. I did try to do the streaming solution, but was kind of defeated for the day. I'll give it another go, it should be better than fetching the whole limit from every shard, since we don't need any kind of "best result" here.

lib/segment/src/id_tracker/simple_id_tracker.rs
timvisee: > that's exactly what this is doing

It does not. I meant something different.

Your implementation keeps generating random numbers until we have a lucky unseen number.

I meant looping through the seen list in-order to find the next (ordered) unseen number without any randomness.

More like this, (but proper, and wrapping):
```rust
let num = random();
if is_already_seen {
    return (num + 1..).filter(|n| !seen.contains(n)).next().unwrap();
}
```

lib/segment/src/fixtures/payload_context_fixture.rs
coszio: left it at unimplemented

lib/segment/src/id_tracker/simple_id_tracker.rs
coszio: Ah, I see what you mean now. However this would tend to have runs of sequential outputs, doesn't it?

timvisee: > However this would tend to have runs of sequential outputs, doesn't it?

Yes.

We might need some heuristic here to dynamically select the right approach. Because the 10k from 10k example you give would be extremely expensive.

@generall may have a good suggestion here.

lib/segment/src/segment.rs
```diff
@@ -1384,6 +1411,22 @@ impl SegmentEntry for Segment {
...
     }
 
+    fn read_random_filtered(&self, limit: usize, filter: Option<&Filter>) -> Vec<PointIdType> {
+        match filter {
+            None => self.read_by_random_id(limit),
+            Some(condition) => {
+                if self.should_pre_filter(condition, Some(limit)) {
+                    let mut points = self.filtered_read_by_index(None, Some(limit), condition);
+                    let mut rng = thread_rng();
+                    points.shuffle(&mut rng);
```
agourlay, line 1421: If I read this correctly, we always return the same set of points in case of filter, but just in a different order.

We could also increase randomly the `limit` to try to surface different points at each call.

coszio: Yeah, that I missed to think of that. Fixed in https://github.com/qdrant/qdrant/pull/4619/commits/f79d705dd0deadea16ce67b461f31d706f2d7c6b

timvisee (approved):

agourlay (approved):

lib/collection/src/collection_manager/holders/proxy_segment.rs
coszio: They are pretty similar, yes, but the assumptions are different. I don't discard the possibility to merge them, though

generall (approved):

PR #4623, opened 2024-07-05
Remove dead code

xzfc:
This PR removes dead public structs/methods/etc that I found while experimenting with `rust-analyzer scip`.


timvisee (approved):
Nice!

Carefully gone over all of them. I don't see any issues in us having missed some of these code paths.

PR #4630, opened 2024-07-08
universal-query: handle score threshold at collection level

coszio:
We forgot to handle the `score_threshold` parameter when it was present in the root query. This PR fixes it.

We only need to handle it for fusions because it is the only case where we generate a new scoring at collection level.
For the other cases, it should already be resolved from shard level


agourlay:
Can you please add a small integration test for this :pray:

agourlay (approved):

Issue #4629, opened 2024-07-08
Update with strong/medium ordering might try to deactivate last active shard 


generall:
## Current Behavior

Currently, if update request fails with strong/medium ordering guarantee, it is always try to deactivate "leader" shard.
In case of this leader shard is the only one, it will create an inconsistency, where node will try to deactivate last peer,
but consensus won't allow it.

It will result in a following error, if locally deactivated peer is queried:

```
{
  "error": "Service internal error: 0 of 0 read operations failed"
}
```

## Steps to Reproduce

1. Run distributed cluster with 2 nodes
2. Create collection with 1 shard upload some data
3. Shut down the peer with local shard
4. Try to make an update with strong consistency guarantee (it will fail)
5. Bring peer back online
6. Try to read collection info from the peer which was not restarted

## Expected Behavior

Peer should not be locally deactivated.

## Possible Solution

Make additional checks in the logic of deactivation.

timvisee:
This problem is resolved by <https://github.com/qdrant/qdrant/pull/4631>, which will be included in the upcoming patch release.

PR #4631, opened 2024-07-08
Don't disable last peer locally

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/4629>

Prevents a shard becoming unresponsive because the last replica was locally disabled. This was possible when using medium/strong ordering. Now it does not allow locally disabling the last replica anymore.

I believe is the simplest approach and covers both scenarios we spoke about, for having just one or multiple replicas.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -825,24 +825,23 @@ impl ShardReplicaSet {
             .active_or_resharding_peers()
             .filter(|id| id != &peer_id);
 
+        let mut locally_disabled_peers_guard = self.locally_disabled_peers.upgradable_read();
+
         // Prevent disabling last peer in consensus
         {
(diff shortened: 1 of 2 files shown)


generall (approved):

generall:
I think it might be a potential double-read deadlock:

https://github.com/qdrant/qdrant/blob/489a93a37e8781b5e67d1fe6110f1d9c5948e451/lib/collection/src/shards/replica_set/update.rs#L334

and 

https://github.com/qdrant/qdrant/blob/3a2e06405fbc66f214f5090027e03bff3f8b687c/lib/collection/src/shards/replica_set/mod.rs#L355

Both read `replica_state`, bu one inside the other.

timvisee:
Good catch! I've resolved this by passing the replica set reference we already have into the locally disable function.

generall (approved):

PR #4632, opened 2024-07-08
Derive/implement `fmt::Debug` for `Segment`

ffuugoo:
This PR derives/implements `fmt::Debug` for `Segment`/`ProxySegment`.

It's a leftover from the "missing payloads" debug. It was pretty useful to be able to look into in-memory state of the segment.

The PR is trivial, just bulky.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -30,6 +30,7 @@ type LockedFieldsMap = Arc<RwLock<HashMap<PayloadKeyType, PayloadFieldSchema>>>;
 /// This object is a wrapper around read-only segment.
 /// It could be used to provide all read and write operations while wrapped segment is being optimized (i.e. not available for writing)
 /// It writes all changed records into a temporary `write_segment` and keeps track on changed points
+#[derive(Debug)]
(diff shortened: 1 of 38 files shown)


ffuugoo:
> The only thing I don't like with these manual implementations is that we might _forget_ new fields when we add them.

True, but this is still better than no debug, IMO. These won't be used too often, I think, but it's 100 times less annoying when it's already implemented (just missing a field or two), than when you need it and have to go drop `#[derive(Debug)]` on 50 different types. 😂

Also, most of manual impls can be switched to derive if we go sprinkle `#[derive(Debug)]` on `quantization` repo.

xzfc:
> The only thing I don't like with these manual implementations is that we might forget new fields when we add them.

We could use destructuring syntax so we got a compile error unless we'll update the debug impl too.

```rust
let DynamicMmapFlags {
    flags,
    flags_flusher: _, // not Debug
    status,
    directory,
} = self;
```

Also, once `derive_more` 1.0.0 come out of beta, we could use their [`#[debug(skip)]`](https://docs.rs/derive_more/1.0.0-beta.6/derive_more/derive.Debug.html) attribute.

PR #4641, opened 2024-07-09
Mmap points to values map

IvanPleshkov:
This PR introduces a mmap analog of `ImmutablePointToValues`, which is required for mmap field indexes.

`MmapPointToValues` has almost the same API. The main difference is that the method returns an iterator of values instead of a slice. It's necessary because mmap file cannot contain slice of strings, we have to parse it.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your core changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #4644, opened 2024-07-09
Field index values as iterator instead of slice

IvanPleshkov:
Tracking issue: #4502.

Integration of this PR https://github.com/qdrant/qdrant/pull/4641.

This PR changes signature of `[ImmutablePointToValues::get_values](https://github.com/qdrant/qdrant/blob/field-index-values-as-iterator-instead-of-slice/lib/segment/src/index/field_index/immutable_point_to_values.rs#L36)`

We cannot use slice anymore because we cannot store slice of strings in mmap file. Instead, `get_values` returns values iterator.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/map_index/immutable_map_index.rs
+++ b/lib/segment/src/index/field_index/map_index/immutable_map_index.rs
@@ -141,7 +141,7 @@ impl<N: Hash + Eq + Clone + Display + FromStr + Default> ImmutableMapIndex<N> {
             }
             self.values_count = self
                 .values_count
-                .checked_div(removed_values_count)
+                .checked_sub(removed_values_count)
                 .unwrap_or_default();
         }
         self.point_to_values.remove_point(idx);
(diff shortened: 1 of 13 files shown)


lib/segment/src/index/field_index/map_index/immutable_map_index.rs
```diff
@@ -137,7 +133,16 @@ impl<N: Hash + Eq + Clone + Display + FromStr + Default> ImmutableMapIndex<N> {
...
+                removed_values_count += 1;
+            }
+
+            if removed_values_count > 0 {
+                self.indexed_points -= 1;
             }
+            self.values_count = self
+                .values_count
+                .checked_div(removed_values_count)
+                .unwrap_or_default();
```
timvisee, line 145: I don't think this matches with `self.values_count -= removed_values.len();` from before. Is that intentional?

xzfc:
I see you've added `get_values_count()` method. Have you considered implementing `ExactSizeIterator` instead?

timvisee (approved):
Looks good once we fix the current open issues.

IvanPleshkov:
> I see you've added `get_values_count()` method. Have you considered using `ExactSizeIterator` instead?

PR with point-to-values map https://github.com/qdrant/qdrant/pull/4641/files cannot implement `ExactSizeIterator` because in mmap case we scan values in only forward direction

lib/segment/src/index/field_index/map_index/immutable_map_index.rs
IvanPleshkov: thanks! it's a bug

PR #4647, opened 2024-07-10
remove schemars examples from the enum

generall:


joein:
I've tried out generating python client from this branch, tests work ✅

joein (approved):

PR #4610, opened 2024-07-03
Basic defragmentation

JojiiOfficial:
Adds the basic functionality of grouping vectors in vector storage together based on the payload value of a given payload.

#### Features not part of this PR:
 - Allow picking custom key
 - Benefiting from already ordered numeric index 

#### Design decisions:
For generic grouping, we use the hash of a value as criteria to sort the vectors. This allows a proper order, even if the payload values have different types. Later we can extend this by using the payload index directly.

The vector storages now do `update_from` on iterators of vectors directly, instead of IDs and a `VectorStorageEnum`.
This allows us to pass vectors more dynamically, eg from multiple different segments in a custom order.

Deduping is done before inserting into the new vector storage, in order to allow ordering the vectors in vector storage properly. The new `update` function can be called multiple times and will still deduplicate the points but a strong order of the vectors can't be enforced when doing so.

PR #4652, opened 2024-07-10
Add defragmentation configuration to the API

JojiiOfficial:
Depends on #4610 

Adds defragmentation keys to the optimizers configuration.
I decided to use an array so we can extend it later with multiple keys for sorting.

### DIFF
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -1,5 +1,6 @@
 use std::collections::{HashMap, HashSet};
 use std::path::Path;
+use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
@@ -39,6 +40,11 @@ pub struct OptimizerThresholds {
     pub indexing_threshold_kb: usize,
 }
 
+#[derive(Debug, Clone)]
+pub struct DefragmentationConfig {
+    pub keys: Vec<String>,
+}
+
 /// SegmentOptimizer - trait implementing common functionality of the optimizers
 ///
 /// It provides functions which allow to re-build specified segments into a new, better one.
@@ -68,6 +74,8 @@ pub trait SegmentOptimizer {
 
     /// Get thresholds configuration for the current optimizer
     fn threshold_config(&self) -> &OptimizerThresholds;
+
+    fn defragmentation_config(&self) -> Option<&DefragmentationConfig>;
 
     /// Checks if segment optimization is required
     fn check_condition(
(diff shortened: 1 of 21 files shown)


lib/collection/src/optimizers_builder.rs
```diff
@@ -77,6 +79,15 @@ pub struct OptimizersConfig {
     /// If 0 - no optimization threads, optimizations will be disabled.
     #[serde(default)]
     pub max_optimization_threads: Option<usize>,
+
+    /// Payload keys to use for defragmentation. The keys configured will be used
+    /// to sort vectors internally to take advantage from memory/storage locality.
+    /// The keys are applied in order, where the first key is most important.
+    /// (For now, only the firts one is applied)
+    ///
+    /// Note: The keys need to be indexed.
```
timvisee, line 88: This may be a nice nuance:

```suggestion
    /// Note: The keys need to be indexed with a type that supports ordering.
```

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -39,6 +40,11 @@ pub struct OptimizerThresholds {
     pub indexing_threshold_kb: usize,
 }
 
+#[derive(Debug, Clone)]
+pub struct DefragmentationConfig {
+    pub keys: Vec<String>,
+}
+
```
timvisee, line 47: I don't think we want a separate type for this. It may be better to put it in the optimizers configuration instead. Just to prevent adding a million different config types.

@generall thoughts? Also, how do you feel about exposing this as part of a collection config?

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -422,20 +430,32 @@ pub trait SegmentOptimizer {
...
+                    .field_indexes
+                    .iter()
+                    .next()
+                    .map(|(k, _)| k.clone());
+
+                if let Some(defragment_key) = defragment_key {
+                    segment_builder.set_defragment_keys(vec![defragment_key]);
+                    break;
+                }
             }
```
timvisee, line 459: I like that we have an automatic function, assuming we can come up with a good default.

However, I wonder if we also want to be able to fully disable it. I'm not sure yet.

We can possibly do this with `null`, `[]`, `["some", "keys]` as defragmentation keys. Where an empty list would mean no defragging at all.

JojiiOfficial: I had the same idea but it seems like protobuf doesn't support optional+repeated fields, so we can't differentiate between `null` and `[]`, at least not without a workaround (please correct me if I'm wrong).
Also I can imagine this becoming a bit confusing, because `null` can also mean "no defragmentation" to some users.
Lastly I couldn't think of a reason why we want to disable it, especially because we don't enforce any order currently.

This were my thoughts when implementing, what do you think?

timvisee: > (...) so we can't differentiate between null and [] (...)

Yeah we have a similar-ish problem in Python actually. There we cannot distinguish between a field being `None` or non-existent.

> Lastly I couldn't think of a reason why we want to disable it, especially because we don't enforce any order currently.

If defragmentation doesn't work for some use case, it only adds overhead during optimization time.

JojiiOfficial: > If defragmentation doesn't work for some use case, it only adds overhead during optimization time.

You're right, but looking at the execution metrics I figured we can trade between complexity and a few ms that it takes longer for sorting, even if it is not needed. Or do you disagree?

PR #4656, opened 2024-07-10
universal-query: gRPC group API

agourlay:
Follow up on https://github.com/qdrant/qdrant/pull/4616 for the gRPC API.

I introduced `CollectionQueryGroupsRequest` as a convergence type for REST and gRPC.

### DIFF
--- a/lib/collection/src/operations/universal_query/collection_query.rs
+++ b/lib/collection/src/operations/universal_query/collection_query.rs
@@ -63,15 +63,15 @@ pub struct CollectionQueryResolveRequest<'a> {
 pub struct CollectionQueryGroupsRequest {
     pub prefetch: Vec<CollectionPrefetch>,
     pub query: Option<Query>,
-    pub using: Option<String>,
+    pub using: String,
     pub filter: Option<Filter>,
     pub params: Option<SearchParams>,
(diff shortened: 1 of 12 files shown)


lib/collection/src/operations/universal_query/collection_query.rs
```diff
@@ -56,6 +58,23 @@ pub struct CollectionQueryResolveRequest<'a> {
...
+    pub filter: Option<Filter>,
+    pub params: Option<SearchParams>,
+    pub score_threshold: Option<ScoreType>,
+    pub with_vector: Option<WithVector>,
+    pub with_payload: Option<WithPayloadInterface>,
+    pub group_by: JsonPath,
+    pub group_size: Option<usize>,
+    pub limit: Option<usize>,
+    pub with_lookup: Option<WithLookup>,
+}
```
coszio, line 76: Just to be consistent with `CollectionQueryRequest`, I think we can resolve these default values when converting to this type:
- using
- with_vector
- with_payload
- group_size
- limit

coszio (commented):
Other than the review comment, it LGTM 👍

coszio (approved):

PR #4657, opened 2024-07-10
universal-query: Expose random sampling query

coszio:
Finishes implementation from collection to rest and grpc API, effectively exposing it.

Sample REST request:
```
POST collections/{collection_name}/points/query
{
  "query": { "sample": "random" }
}
```

It will produce a random, unstable output. If used without prefetches, offset will be ignored.

### DIFF
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -508,6 +508,15 @@ enum Fusion {
     DBSF = 1; // Distribution-Based Score Fusion
 }
 
+/// Sample points from the collection
+///
+/// Available sampling methods:
+///
+/// * `random` - Random sampling
+enum Sample {
+    RANDOM = 0;
+}
+
 message Query {
   oneof variant {
     VectorInput nearest = 1; // Find the nearest neighbors to this vector.
@@ -516,6 +525,7 @@ message Query {
     ContextInput context = 4; // Return points that live in positive areas.
     OrderBy order_by = 5; // Order the points by a payload field.
     Fusion fusion = 6; // Fuse the results of multiple prefetches.
+    Sample sample = 7; // Sample points from the collection.
   }
 }
 
--- a/tests/openapi/openapi_integration/test_query.py
+++ b/tests/openapi/openapi_integration/test_query.py
@@ -345,10 +345,11 @@ def test_basic_random_query():
         assert len(points) == 10
         assert set(point["id"] for point in points) == set(range(1, 11))
         
-        # check the order of ids are different every time
         ids = str([point["id"] for point in points])
-        assert ids not in ids_lists
         ids_lists.add(ids)
+        
+    # check the order of ids are different at least once
+    assert len(ids_lists) > 1
     
 
 def test_basic_rrf():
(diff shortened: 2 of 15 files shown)


lib/api/src/grpc/proto/points.proto
```diff
@@ -507,6 +507,15 @@ enum Fusion {
     RRF = 0; // Reciprocal Rank Fusion
 }
 
+/// Sample points from the collection
+///
+/// Available sampling methods:
+///
+/// * `random` - Random sampling
+enum Sample {
+    RANDOM = 0;
```
agourlay, line 516: Random thought here, is there a use case for exposing to precise underlying distribution used to sample ids?

```rust
enum Sample {
  RAND_UNIFORM = 0;
  RAND_NORMAL = 1;
  ...
} 
```

timvisee: I don't think that makes a lot of sense.

First of all, if I assume that 'normal' is the standard `rand` distribution, then it basically is the same as uniform.

Second, we pick random points directly from storage slices. Not from an ordered list of point IDs. And storage is modified all the time. I therefore think different distributions are meaningless. Correct me if I'm wrong though!

tests/openapi/openapi_integration/test_query.py
```diff
@@ -328,6 +328,29 @@ def test_basic_order_by():
...
+        assert response.ok, response.text
+        
+        points = response.json()["result"]["points"]
+        assert len(points) == 10
+        assert set(point["id"] for point in points) == set(range(1, 11))
+        
+        # check the order of ids are different every time
+        ids = str([point["id"] for point in points])
+        assert ids not in ids_lists
+        ids_lists.add(ids)
```
timvisee, line 351: If this ever becomes flaky I think we could also do a few requests and just make sure that some differ, not necessarily all.

lib/api/src/grpc/proto/points.proto
coszio: > we pick random points directly from storage slices. Not from an ordered list of point IDs. And storage is modified all the time. I therefore think different distributions are meaningless. Correct me if I'm wrong though!

Basically this. But also a normal distribution assumes some mean to use as center, which also doesn't makes sense for me.

tests/openapi/openapi_integration/test_query.py
coszio: good idea, it might be possible since our dataset is very small

coszio: solved

timvisee (approved):

lib/api/src/grpc/proto/points.proto
agourlay: If other distributions do not make sense here, any idea if this enum could actually receive additional variants in the future?

coszio: @agourlay Now that I think about it, we could be adding **vector**-based distributions/approaches like:
- normal distribution (I think I missed that this is what you meant 🙈 )
- minimum similarity (aka diversity). Returning points which are furthest from all other returned points.

agourlay: I am fine for keeping it open until we figure something else out :+1: 

Thanks for thinking about it.

agourlay (approved):

generall (approved):

PR #4658, opened 2024-07-11
In memory IdTracker

JojiiOfficial:
Depends on #4403 

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/id_tracker/immutable_id_tracker.rs
+++ b/lib/segment/src/id_tracker/immutable_id_tracker.rs
@@ -32,16 +32,16 @@ pub const VERSION_MAPPING_FILE_NAME: &str = "id_tracker.versions";
 pub struct ImmutableIdTracker {
     path: PathBuf,
 
-    deleted: BitVec,
+    pub(crate) deleted: BitVec,
     deleted_wrapper: MmapBitSliceBufferedUpdateWrapper,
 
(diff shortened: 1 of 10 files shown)


lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -29,16 +29,16 @@ pub const VERSION_MAPPING_FILE_NAME: &str = "id_tracker.versions";
...
-    deleted: BitVec,
+    pub(crate) deleted: BitVec,
     deleted_wrapper: MmapBitSliceBufferedUpdateWrapper,
 
-    internal_to_version: Vec<SeqNumberType>,
+    pub(crate) internal_to_version: Vec<SeqNumberType>,
     internal_to_version_wrapper: MmapSliceBufferedUpdateWrapper<SeqNumberType>,
 
-    mappings: PointMappings,
+    pub(crate) mappings: PointMappings,
```
timvisee, line 38: Do we need `pub(crate)` just for conversions? (I haven't properly checked yet).

If so, maybe we can do `pub(super)` instead to limit it a bit more.

timvisee: Thanks. Better like this.

By the way, please re-request review if you think it's ready for another round.

generall (approved):

PR #4659, opened 2024-07-11
Fix examples for scored point

kartik-gupta-ij:
Output: 
Screenshot 2024-07-11 at 5 22 49 PM


generall:
@joein could you please verify this changes don't break python client generation

generall (approved):

PR #4662, opened 2024-07-12
Introduce iterator for DatabaseColumnScheduledDeleteWrapper

xzfc:
This PR amends #4653. It implements an alternative to the approach described in https://github.com/qdrant/qdrant/pull/4653#discussion_r1673626968.
I suggest to review it per-commit.


timvisee (approved):
Nice! Basically follows the same approach as the other non-scheduled iterator. That makes a lot of sense.

Thanks!

Let's merge as soon as CI is green.

ffuugoo:
I'm a bit late to this, but great job @xzfc! `iter_pending_deletes` hack was ugly as hell. 😂

PR #4663, opened 2024-07-12
Fix full payload update

generall:
Current full payload update deletes existing payload and then writes a new one, which is:

- sub-optimal
- may lead to the lost payload if the service is killed with partial flush (not verified)


timvisee (approved):
:crossed_fingers:

PR #4664, opened 2024-07-12
perf: Use `Mutex` in `PrefetchResults`

coszio:
Changes the `PrefetchResults` struct to use an ~~`RwLock`~~`Mutex` and `mem::take` its contents. Drops the `Cow` usage to avoid duplicating results when returning them.

Based on criterion benchmarks, these are some of the approximate changes (with_payload=true):
- nearest, limit 10: -5% (still 1% slower than `core_search`)
- rrf limit 10, from 2 searches limit 100: -19%
- rescore limit 10, from search limit 100: no noticeable change

if we increase limits, this becomes even more noticeable:
- nearest, limit 100: -14% (still 1% slower than `core_search`)
- rrf limit 100, from 2 searches limit 1000: -28%
- rescore limit 100, from search limit 1000: -2%


generall:
A bit more per on top: 
image


The reasoning is:

- We can use simple mutex instead of RwLock, as we actually never read.
- Get operation is fairly short, it doesn't even touch heap, so using async lock might have too high overhead

generall (approved):

agourlay (approved):
Thanks for the benchmarks :+1:

timvisee (approved):

PR #4666, opened 2024-07-15
Resume resharding driver on node restart

timvisee:
Tracked in: #4213 
Depends on: <https://github.com/qdrant/qdrant/pull/4617>

Allow the resharding driver to resume at any arbitrary point. This allows the driving node to restart as if nothing has happened.

There is more to be done in terms of resumption and the ability to abort, but that'll be for separate PRs.

A new test is added interrupting the process at various stages, asserting the end result is what we expect.

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/4617>
- [x] Rebase on `dev`
- [x] Undraft

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -56,7 +56,7 @@ impl Collection {
                 let mut config = self.collection_config.write().await;
 
                 if config.params.sharding_method.unwrap_or_default() == ShardingMethod::Auto {
-                    debug_assert_eq!(config.params.shard_number.get(), resharding_key.shard_id,);
+                    debug_assert_eq!(config.params.shard_number.get(), resharding_key.shard_id);
                 }
 
                 config.params.shard_number = config
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -68,7 +68,6 @@ impl Collection {
     /// If no resharding is active, this returns early without error.
     pub async fn resume_resharding_unchecked<T, F>(
         &self,
-        // resharding_key: ReshardKey,
         consensus: Box<dyn ShardTransferConsensus>,
         temp_dir: PathBuf,
         on_finish: T,
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -55,9 +55,8 @@ impl Collection {
             {
                 let mut config = self.collection_config.write().await;
 
-                #[cfg(debug_assertions)]
                 if config.params.sharding_method.unwrap_or_default() == ShardingMethod::Auto {
-                    assert_eq!(config.params.shard_number.get(), resharding_key.shard_id);
+                    debug_assert_eq!(config.params.shard_number.get(), resharding_key.shard_id,);
                 }
 
                 config.params.shard_number = config
@@ -194,11 +193,10 @@ impl Collection {
         {
             let mut config = self.collection_config.write().await;
 
-            #[cfg(debug_assertions)]
             if config.params.sharding_method.unwrap_or_default() == ShardingMethod::Auto {
-                assert_eq!(
+                debug_assert_eq!(
                     config.params.shard_number.get() - 1,
-                    resharding_key.shard_id
+                    resharding_key.shard_id,
                 );
             }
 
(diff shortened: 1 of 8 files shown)


lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -533,6 +526,16 @@ impl ShardHolder {
...
             ShardingMethod::Auto => {
+                let mut shard_number = shard_number;
+
+                // If resharding is active, we should load an extra shard
+                // TODO(resharding): prevent error if extra shard cannot be loaded?
+                // TODO(resharding): may be needed if resharding starts and crashes before migrate
+                let is_resharding = self.resharding_state.read().is_some();
+                if is_resharding {
+                    shard_number += 1;
+                }
```
ffuugoo, line 537: I'm not 100% sure about this one. Should we, maybe, increment `shard_number` at the point when we *create* resharding shard (i.e., when we start resharding)?

And add code to decrement it if we abort.

Seems easier this way, I think.

lib/collection/src/collection/resharding.rs
```diff
@@ -106,7 +160,25 @@ impl Collection {
...
+                debug_assert_eq!(
+                    config.params.shard_number.get() - 1,
+                    resharding_key.shard_id,
+                );
+            }
+
+            if let Err(err) = config.save(&self.path) {
+                log::error!("Failed to update and save collection config during resharding: {err}");
+            }
+        }
```
ffuugoo, line 181: Ditto.

ffuugoo (approved):

lib/collection/src/shards/shard_holder/mod.rs
timvisee: Yes, that's probably better.

On abort, we also need to explicitly remove that extra shard. That's probably a better place to decrease the count.

timvisee: a47baaa095e8f670e0d184e95a0877e193b1f7c9

lib/collection/src/collection/resharding.rs
```diff
@@ -162,30 +181,31 @@ impl Collection {
...
-            // If using automatic sharding, the reshard key shard ID must correspond with number
+            #[cfg(debug_assertions)]
             if config.params.sharding_method.unwrap_or_default() == ShardingMethod::Auto {
-                debug_assert_eq!(
+                assert_eq!(
                     config.params.shard_number.get() - 1,
-                    resharding_key.shard_id,
+                    resharding_key.shard_id
                 );
             }
```
ffuugoo, line 200: I'd actually leave it as it was before in `finish_resharding`. E.g., just `debug_assert_eq` under conditional, without explicit `#[cfg(debug_assertions)]`. I'd expect the compiler to optimize the conditional away in release.

lib/collection/src/collection/resharding.rs
```diff
@@ -49,6 +50,24 @@ impl Collection {
...
 
             shard_holder.start_resharding_unchecked(resharding_key.clone(), replica_set)?;
+
+            // Increase the persisted shard count, loads new shard on restart
+            {
+                let mut config = self.collection_config.write().await;
+
+                #[cfg(debug_assertions)]
+                if config.params.sharding_method.unwrap_or_default() == ShardingMethod::Auto {
+                    assert_eq!(config.params.shard_number.get(), resharding_key.shard_id,);
```
ffuugoo, line 60: Trailing comma.

```suggestion
                    assert_eq!(config.params.shard_number.get(), resharding_key.shard_id);
```

ffuugoo (approved):

lib/collection/src/collection/resharding.rs
```diff
@@ -162,30 +181,31 @@ impl Collection {
...
-                debug_assert_eq!(
+                assert_eq!(
                     config.params.shard_number.get() - 1,
-                    resharding_key.shard_id,
+                    resharding_key.shard_id
                 );
             }
 
+            config.params.shard_number = NonZeroU32::new(config.params.shard_number.get() - 1)
+                .expect("cannot have zero shards after aborting resharding");
```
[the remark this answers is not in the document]
timvisee: That does not work.

It is a `NonZeroU32` which does not have [`_sub`](https://doc.rust-lang.org/std/num/type.NonZeroU32.html) methods. Calling `checked_sub(1)` on the raw `u32` does not make sense, because it will always be 1 or more.

PR #4677, opened 2024-07-16
Add JsonPath::filename()

xzfc:
Tracking issue: #4502.
This PR adds the method `JsonPath::filename()`. It would be used as a directory name for on-disk indices.

### DIFF
--- a/lib/segment/src/json_path/mod.rs
+++ b/lib/segment/src/json_path/mod.rs
@@ -1,10 +1,13 @@
 use std::fmt::{Display, Formatter};
 
+use data_encoding::BASE32_DNSSEC;
+use itertools::Itertools as _;
 use schemars::gen::SchemaGenerator;
 use schemars::schema::Schema;
 use schemars::JsonSchema;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde_json::Value;
+use sha2::{Digest as _, Sha256};
 
 use crate::common::anonymize::Anonymize;
 use crate::common::utils::{merge_map, MultiValue};
@@ -274,6 +277,36 @@ impl JsonPath {
                 }
             }
         }
+    }
+
+    /// Convert the path into a string suitable for use as a filename by adhering to the following
+    /// restrictions: max length, limited character set, but still being unique.
+    pub fn filename(&self) -> String {
+        const MAX_LENGTH: usize = 64;
+        const HASH_LENGTH: usize = 32; // In base32 characters, i.e. 5 bits per character.
+
+        let text = self.to_string();
+        let mut result = String::with_capacity(MAX_LENGTH);
+
+        BASE32_DNSSEC.encode_append(
+            &Sha256::digest(text.as_bytes()).as_slice()[0..(HASH_LENGTH * 5).div_ceil(8)],
+            &mut result,
+        );
+        debug_assert_eq!(result.len(), HASH_LENGTH);
+
+        result.push('-');
+
+        text.chars()
+            .map(|c| match c {
+                'a'..='z' | 'A'..='Z' | '0'..='9' | '-' | '_' => c.to_ascii_lowercase(),
+                _ => '_',
+            })
+            .dedup_by(|&a, &b| a == '_' && b == '_')
+            .take(MAX_LENGTH - result.len())
+            .for_each(|c| result.push(c));
+
+        debug_assert!(result.len() <= MAX_LENGTH);
+        result
     }
 }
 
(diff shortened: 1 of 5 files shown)


lib/segment/src/json_path/mod.rs
```diff
@@ -1310,4 +1335,25 @@ mod tests {
         ); // empty object left
         assert_eq!(payload, Default::default());
     }
+
+    #[test]
+    fn test_filename() {
+        assert_eq!(
+            JsonPath::new("foo").filename(),
+            "LCa0a2j_xo_5m0U8HTBBNBNCLXBkg7-g-YpeiGJm564-foo",
```
generall, line 1343: wdyt, about making hash lowercase (same as https://github.com/qdrant/qdrant/blob/6cfcce764d768df67b742107d559b66f75a3132b/lib/collection/src/tests/sha_256_test.rs#L14) and putting name first? 
Might be a bit more readable and easy to search, and we won't have concern about those windows users which don't recognize capitalization

Otherwise lgtm

timvisee (approved):

IvanPleshkov (approved):

lib/segment/src/json_path/mod.rs
xzfc: Update:
- Switched hash part to base32 (it's all lowercase alphanumeric, no `_` and `-`).
- Converted path part to lowercase (`fooBar` → `foobar`).
- Truncated hash to 32 chars (160 bits).
- Truncated the total length to 64 chars (32 for a hash, a dash, and 31 for a path).

---

As for putting the name first, I believe that hash being first is more readable. Compare:

```
5gjb8qr8vv38vucr8ku1qc216g9k4bbg-foo
jkdf7941qdcjau3f7mgqrb4ka2pua7p0-foo_bar_baz_qoox_0_
c5b17acv8pn0s0te8pg7molpa1h09prh-really_looooooooooooooooooooooo
4huj4rn1fflrtriqo0tieqhhu6hp5lci-m_esli
oelovs9n43kar427rkskcqpsh5qeb4m2-foobar
ivfjb25lkfp4nau3gkdjebobksd9rjeu-foobarbaz
2ogqcsltkpvl4g94iob9unnjeiimoiuo-foobarbazqoox
```

```
foo-5gjb8qr8vv38vucr8ku1qc216g9k4bbg
foo_bar_baz_qoox_0_-jkdf7941qdcjau3f7mgqrb4ka2pua7p0
really_looooooooooooooooooooooo-c5b17acv8pn0s0te8pg7molpa1h09prh
m_esli-4huj4rn1fflrtriqo0tieqhhu6hp5lci
foobar-oelovs9n43kar427rkskcqpsh5qeb4m2
foobarbaz-ivfjb25lkfp4nau3gkdjebobksd9rjeu
foobarbazqoox-2ogqcsltkpvl4g94iob9unnjeiimoiuo
```

IvanPleshkov: As I remember, windows does not allow to create directory with more than 256 symbols. WDYT, is it a problem for us?

IvanPleshkov: Checked, windows does not support long directory names. Do we need to shorten the folder name using more characters?

xzfc: By default, Windows apps has `MAX_PATH` limitation of 260 characters, but it could be lifted [\[1\]]. Need to investigate whether it's a problem for modern Rust apps, but it surely could break other apps that peeking into qdrant storage dir (file managers, archivers, etc).

I've created some made-up path and it's 223 symbols long, which is 85% of 260. So a chance to hit it is high.
```
D:\projects\qdrant\goods_categorization_demo\storage\collections\goods\0\segments\2e9e753e-6955-4e26-8cfb-f649fb99d6b3\mmap_payload_index\c5b17acv8pn0s0te8pg7molpa1h09prh-really_looooooooooooooooooooooo\point_to_values.bin
^                                           ^                                    ^                                    ^                  ^                                                                ^                   ^
1                                           45                                   82                                  119                138                                                              203                 223
```

So, open questions:
1. Do we care if qdrant would break on deeply nested working directories on Windows?
2. Do we care if qdrant would not break on long paths (if we manage to circumvent that), but other apps peeking into the storage dir would?



[\[1\]]: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry
See also:
- https://www.github.com/BurntSushi/ripgrep/commit/db6bb21a629d5b1ec1bfe89c693b280497c9eedc
- https://www.github.com/rust-lang/rust/pull/89174

xzfc:
OK, long names on Windows could be problematic.
Merging it for now (to unblock futher work on this feature), but adding a TODO entry into the tracking issue.

PR #4680, opened 2024-07-16
sample in O(limit)

generall:
The suggested algorithm:
- guarantees to finish in O(limit)  - assuming the selected points are unique
- subjectively easier to understand

### DIFF
--- a/lib/collection/src/shards/local_shard/scroll.rs
+++ b/lib/collection/src/shards/local_shard/scroll.rs
@@ -4,6 +4,7 @@ use std::time::Duration;
 
 use futures::future::try_join_all;
 use itertools::Itertools as _;
+use rand::distributions::WeightedIndex;
 use rand::Rng;
 use segment::data_types::order_by::{Direction, OrderBy, OrderValue};
 use segment::types::{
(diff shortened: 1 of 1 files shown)


lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -280,51 +281,57 @@ impl LocalShard {
...
-
-            if !rng.gen_bool(*weight) {
-                continue;
+        let mut random_points = HashSet::with_capacity(limit);
+
+        // This loop iterates <= LIMIT times, and either quits if we have enough points,
+        // or if some of the segments are exhausted.
+        //
+        // If the segments are exhausted, we will fill up the rest of the points from other segments.
+        // In total, the complexity is guaranteed to be O(limit).
```
timvisee, line 303: I feel like the two parts contradict each other:
- the first says it stops if a segment is exhausted
- the second says we'll fill up with other segments

Seeing the implementation I believe this is intentional. Maybe this helps clarify a bit though:

```rust
        // Randomly sample points in two stages
        //
        // 1. This loop iterates <= LIMIT times, and either breaks early if we
        // have enough points, or if some of the segments are exhausted.
        //
        // 2. If the segments are exhausted, we will fill up the rest of the
        // points from other segments. In total, the complexity is guaranteed to
        // be O(limit).
```

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -280,51 +281,57 @@ impl LocalShard {
...
-            };
-
-            // Add point to result
-            random_points.push(point);
-            if random_points.len() == limit {
-                break;
             }
         }
 
+        let selected_points: Vec<_> = random_points.into_iter().collect();
```
timvisee, line 330: Is it possible for the output to be somewhat deterministic here due to the hash set? I honestly don't know.

If it is, I think that's undesirable.

timvisee (approved):

lib/collection/src/shards/local_shard/scroll.rs
timvisee: Assuming each set has an unique [`RandomState`](https://doc.rust-lang.org/std/collections/hash_map/struct.RandomState.html), it does not have this problem.

coszio (approved):
I agree this easier to reason about. Thanks.

PR #4684, opened 2024-07-17
Defragmentation in api

JojiiOfficial:
Replaces #4652 

In this version we don't automatically decide on a defragmentation key and only defragment if a user configured Qdrant to do so.
If we still agree on doing otherwise and intelligently try to automatically find a key to use for defragmentation, we can still implement this in a separate PR.

Also note that this PR only uses the first found key configured as tenant. To fix this, we need to agree on an algorithm for applying multiple keys, which I also suggest to use in a separate PR.

### DIFF
--- a/docs/grpc/docs.md
+++ b/docs/grpc/docs.md
@@ -858,6 +858,7 @@
 | ----- | ---- | ----- | ----------- |
 | lookup | [bool](#bool) |  | If true - support direct lookups. |
 | range | [bool](#bool) |  | If true - support ranges filters. |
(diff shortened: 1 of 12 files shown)


timvisee (commented):
The implementation seems good.

The naming still seems to be a bit controversial though. I prefer something different than `is_primary`. Let's discuss in Slack and settle on something there.

generall:
I will merge this PR, as it blocks the original one from being merged (this one actually disables defrag by default, which is desired behaviour).

Naming can be discussed separately.

PR #4686, opened 2024-07-17
universal-queries: Size prefetches according to number of groups

agourlay:
When using the grouping API with prefetches, it is necessary to size the inner `limit` according to the number of groups to expand the space of values inspected.

### DIFF
--- a/lib/collection/src/grouping/group_by.rs
+++ b/lib/collection/src/grouping/group_by.rs
@@ -123,7 +123,7 @@ impl GroupRequest {
             source: query_search,
             group_by: self.group_by,
             group_size: self.group_size,
-            limit: self.limit,
+            groups: self.limit,
             with_lookup: self.with_lookup,
         })
     }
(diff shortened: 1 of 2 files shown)


lib/collection/src/grouping/group_by.rs
```diff
@@ -123,7 +123,7 @@ impl GroupRequest {
             source: query_search,
             group_by: self.group_by,
             group_size: self.group_size,
-            limit: self.limit,
+            groups: self.limit,
```
agourlay, line 126: renamed to `groups` for more clarity

lib/collection/src/grouping/group_by.rs
```diff
@@ -144,7 +144,11 @@ impl QueryGroupRequest {
     ) -> CollectionResult<Vec<ScoredPoint>> {
         let mut request = self.source.clone();
 
-        request.limit = self.limit * self.group_size;
+        // Adjust limit to fetch enough points to fill groups
+        request.limit = self.groups * self.group_size;
+        request.prefetches.iter_mut().for_each(|prefetch| {
+            prefetch.limit *= self.groups;
+        });
```
coszio, line 151: we must do this recursively, not only to the first layer of prefetches :)

coszio (approved):

PR #4688, opened 2024-07-17
Replica resolver sorts by overall index

coszio:
Fixes same issue as #4511:

The order of results is not always the same, as the implementation had been considering. Nearest search can have different order direction depending on the metric (Euclidean vs Dot), and order-by can select its direction too.

~To preserve this ordering, I am introducing an extra `precedence` field in `ResolverRecord` to aggregate the sum of the positions of the same item in different results.~

~By having this number, in the end we can just sort in ascending order, and the resolved list should be ordered correctly.~

~I have not written any new tests yet, but the [`resolve_scored_points_batch_4_majority`](https://github.com/qdrant/qdrant/blob/dev/lib/collection/src/shards/resolve.rs#L397) test failed once we [stopped sorting `Vec<Vec<ScoredPoint>>`](https://github.com/qdrant/qdrant/compare/dev...replica-resolver-also-resolves-order?body=&expand=1#diff-c0decfb6f5c7e6dd8e2f779deee8d7dc971600d003faa3a9da2e79d1a2cc81f4L62-L68), and now it passes again with the new approach.~

Edit: it does not always work.

Edit2: I have now solved the issue by creating a custom iterator which takes a vec of filtered rows, and advances them in relation to their heads. So that they advance once enough heads have the same item.

### DIFF
--- a/lib/collection/src/shards/resolve.rs
+++ b/lib/collection/src/shards/resolve.rs
@@ -1,7 +1,9 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::hash;
+use std::iter::Peekable;
+use std::rc::Rc;
 
-use itertools::Itertools;
+use itertools::{Itertools, Position};
 use segment::types::{Payload, ScoredPoint};
 use tinyvec::TinyVec;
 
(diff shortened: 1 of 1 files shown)


lib/collection/src/shards/resolve.rs
```diff
@@ -208,9 +202,14 @@ where
...
 struct ResolverRecord<'a, T> {
     item: Option<&'a T>,
+    /// Offset within all results
     row: usize,
+    /// Offset within a set of results
     index: usize,
+    /// Keeps track of the amount of times we see this same item
     count: usize,
+    /// Keeps track of the index of the same item in different results. The lower, the earlier.
+    precedence: usize,
```
timvisee, line 212: How about just calling it order? Or do you want to prevent confusing it with order by?

coszio: I mean it is not exactly order, it is just a heuristic for how early it should appear. It is the sum of the positions in each response

coszio:
Just found a way to make this approach fail :( again

timvisee:
> Just found a way to make this approach fail :( again

Does your last commit patch this problem?

lib/collection/src/shards/resolve.rs
coszio: Outdated :)

generall:
I am too stupid to understand what's happening there

generall (approved):

PR #4690, opened 2024-07-18
Make ValueIndexer trait non-generic

xzfc:
```rust
// This PR replaces this trait (old):
trait ValueIndexer<T> {
    …
}

// With this trait (new):
trait ValueIndexer {
    type ValueType;
    …
}
```

# Why/general direction
We have this implementation of `build_field_indexes`: https://github.com/qdrant/qdrant/blob/v1.10.1/lib/segment/src/index/struct_payload_index.rs#L172-L191
How it build both appendable and non-appendable segments: it creates mutable index no matter what kind of segments it builds, and appends point into it. It assumes ("the assumption") that we have two kind of payload indices: mutable and immutable; and both of them use the same on-disk representation (aka rocksdb).

The on-disk payload index doesn't fit into this model: it's immutable and it doesn't have a mutable counterpart.

I am trying to untangle this code by moving "the assumption" from the `build_field_indexes` method into implementations of a `trait PayloadIndexBuilder` (yet to be introduced). One particular roadblock for this is `trait ValueIndexer<T>`, particularly the generic parameter `T`. The type parameter `T` is an implementation detail that exposes what kind of internal representation a particular index uses. So, this PR moves this generic parameter into the associated type so users of this trait can stop caring what internal representation is used.

# Implementation details
Previously the same `NumericIndex` were used to handle multiple kind of payloads (thus multiple impls of `ValueIndexer<T>` for different `T`). Now `NumericIndex` is a wrapper parametrized with a payload type, so it knows the type of payload it works with. But internally it's still the same payload-agnostic index.

```rust
// Old
pub enum FieldIndex {
    IntIndex(NumericIndex<IntPayloadType>), // same
    DatetimeIndex(NumericIndex<IntPayloadType>), // same
    FloatIndex(NumericIndex<FloatPayloadType>),
    …
}

pub enum NumericIndex<T: Encodable + Numericable + Default> {
    Mutable(MutableNumericIndex<T>),
    Immutable(ImmutableNumericIndex<T>),
}
```

```rust
// New
pub enum FieldIndex {
    IntIndex(NumericIndex<IntPayloadType, IntPayloadType>),
    DatetimeIndex(NumericIndex<IntPayloadType, DateTimePayloadType>),
    FloatIndex(NumericIndex<FloatPayloadType, FloatPayloadType>),
    …
}

pub struct NumericIndex<T: Encodable + Numericable + Default, P> {
    inner: NumericIndexInner<T>,
    _phantom: PhantomData<P>,
}

pub enum NumericIndexInner<T: Encodable + Numericable + Default> {
    Mutable(MutableNumericIndex<T>),
    Immutable(ImmutableNumericIndex<T>),
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1604,6 +1604,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "delegate"
+version = "0.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4e018fccbeeb50ff26562ece792ed06659b9c2dae79ece77c4456bb10d9bf79b"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.48",
+]
+
+[[package]]
 name = "derive_builder"
 version = "0.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5419,6 +5430,7 @@ dependencies = [
  "criterion",
  "data-encoding",
  "dataset",
+ "delegate",
  "fnv",
  "fs_extra",
  "generic-tests",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,6 +124,7 @@ tikv-jemallocator = "0.6"
 [workspace.dependencies]
 chrono = { version = "0.4.38", features = ["serde"] }
 data-encoding = { version = "2.6.0" }
+delegate = "0.12.0"
 fnv = "1.0"
 futures = "0.3.30"
 futures-util = "0.3.30"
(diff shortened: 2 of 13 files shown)


ffuugoo (approved):
Good design. Makes more sense, IMO.

coszio (approved):
I don't have any counter against this other the nit of needing to create this extra `*Inner` type if the trait needs to be implemented for a different value type. However, I like that it simplifies the usage of `ValueIndexer` elsewhere.

IvanPleshkov (approved):

PR #4692, opened 2024-07-18
Enable new idtracker

JojiiOfficial:
Enables the `ImmutableIdTracker` and `InMemroyIdTracker`.
Marked as draft until next release.

### DIFF
--- a/lib/segment/src/id_tracker/immutable_id_tracker.rs
+++ b/lib/segment/src/id_tracker/immutable_id_tracker.rs
@@ -252,7 +252,7 @@ impl ImmutableIdTracker {
         })
     }
 
-    pub(super) fn new(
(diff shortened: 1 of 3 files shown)


timvisee:
While it does enable the in-memory ID tracker, this does not seem to enable the immutable ID tracker in its current state. It only allows opening it from disk, but we never actually create it.

JojiiOfficial:
We create it here!
https://github.com/qdrant/qdrant/pull/4692/files#diff-48d69fee7ec36ad38673917fe953159ed23e135b9207465b30050379f4e115e4R412

timvisee:
Thanks for pointing to it. It appears 'show references' was broken, because it didn't show anything for me :smile:

PR #4699, opened 2024-07-18
Abort resharding when relevant consensus events happen

ffuugoo:
Tracked in #4213.

This PR adds two main features:
- forbid to abort resharding, if it reached `ReshardStage::WriteHashRingCommitted` stage
- automatically cancel/abort resharding, when certain consensus operations applied, that affect collection or shards that is involved in resharding:
	- if collection is deleted, resharding is automatically aborted
	- if shard key is deleted, resharding is automatically aborted
	- if peer is force-removed from cluster, resharding is automatically aborted
- a few other relevant operations does not require special handling:
	- it's impossible to delete last repilca of a shard, so no special handling required before new shard is replicated
	- and it seems like if either replica of new shard is removed *during* replication, it should restart the whole resharding, which is good-enough, IMO
		- however, there seem to be some bugs related to removing replicas involved in stream-records transfer 😬
- tests are implemented for all scenarios
	- they are super flakey though
	- tests for removing shard replicas are also implemented, but currently disabled due to bugs

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -530,6 +530,8 @@ impl Collection {
     }
 
     pub async fn remove_shards_at_peer(&self, peer_id: PeerId) -> CollectionResult<()> {
+        // Abort resharding, if shards are removed from peer driving resharding
+        // (which *usually* means the *peer* is being removed from consensus)
         let resharding_state = self
             .resharding_state()
             .await
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -398,7 +398,25 @@ impl Collection {
             }
         }
 
-        // TODO(resharding): 🤔
+        // Abort resharding, if resharding shard is marked as `Dead`.
+        //
+        // This branch should only be triggered, if resharding is currently at
+        // `ReshardStage::MigratingPoints` stage, because resharding shard should be marked as
+        // `Active` when all resharding transfers are successfully completed, and so the check
+        // *right above* this one should be triggered.
+        //
+        // So, if resharding reached `ReshardingStage::ReadHashRingCommitted`, this branch *won't*
+        // be triggered, and in this case, resharding *won't* be cancelled. Though, the update
+        // request should *fail* with "failed to update all replicas of a shard" error.
+        //
+        // If resharding reached `ReshardingStage::WriteHashRingCommitted`, and this branch is
+        // triggered *somehow*, then `Collection::abort_resharding` call should return an error,
+        // so no special handling is needed for `ReshardingStage::WriteHashRingCommitted`.
+        //
+        // TODO(resharding):
+        //
+        // Abort resharding, if resharding shard is (being) marked as `Dead` and resharding is at
+        // `ReshardingStage::ReadHashRingCommitted` stage!? 🤔
         if current_state == Some(ReplicaState::Resharding) && state == ReplicaState::Dead {
             let shard_key = shard_holder
                 .get_shard_id_to_key_mapping()
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -407,11 +407,14 @@ impl Collection {
 
             drop(shard_holder);
 
-            self.abort_resharding(ReshardKey {
-                peer_id,
-                shard_id,
-                shard_key,
-            })
+            self.abort_resharding(
+                ReshardKey {
+                    peer_id,
+                    shard_id,
+                    shard_key,
+                },
+                false,
+            )
             .await?;
 
             // TODO(resharding): Abort all resharding transfers!?
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -530,6 +530,20 @@ impl Collection {
     }
 
     pub async fn remove_shards_at_peer(&self, peer_id: PeerId) -> CollectionResult<()> {
+        let resharding_state = self
+            .resharding_state()
+            .await
+            .filter(|state| state.peer_id == peer_id);
+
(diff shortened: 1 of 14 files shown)


ffuugoo:
I'll do some tests tomorrow, fix stuff that might not work yet and undraft PR. Though, I'd appreciate it if you can take a look and see if my *logic* (in comments and implementations) seem solid, or you have some comments already.

lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -340,7 +351,7 @@ impl TableOfContent {
             }
 
             ReshardingOperation::Abort(key) => {
-                collection.abort_resharding(key).await?;
+                collection.abort_resharding(key, false).await?;
```
timvisee, line 354: Do we want to expose the boolean to the user? Or possibly, hide it from the user, but allow us to specify it if we ever want to manually recover some state?

lib/collection/src/collection/mod.rs
```diff
@@ -407,11 +425,14 @@ impl Collection {
...
+                ReshardKey {
+                    peer_id,
+                    shard_id,
+                    shard_key,
+                },
+                false,
+            )
             .await?;
 
             // TODO(resharding): Abort all resharding transfers!?
```
timvisee, line 438: I think we can remove this line

lib/collection/src/collection/mod.rs
```diff
@@ -509,6 +530,20 @@ impl Collection {
...
+            .filter(|state| state.peer_id == peer_id);
+
+        if let Some(state) = resharding_state {
+            if let Err(err) = self.abort_resharding(state.key(), true).await {
+                log::error!(
+                    "Failed to abort resharding {} while removing peer {peer_id}: {err}",
+                    state.key(),
+                );
+            }
+        }
```
timvisee, line 545: I was a bit confused by this one. A simple comment might help.

```suggestion
        // If shards are removed from peer driving resharding, abort it now
        let resharding_state = self
            .resharding_state()
            .await
            .filter(|state| state.peer_id == peer_id);

        if let Some(state) = resharding_state {
            if let Err(err) = self.abort_resharding(state.key(), true).await {
                log::error!(
                    "Failed to abort resharding {} while removing peer {peer_id}: {err}",
                    state.key(),
                );
            }
        }
```

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -176,15 +176,57 @@ impl ShardHolder {
...
+        // - it's safe to run, if write hash ring was not committed yet
+        if state.stage < ReshardStage::WriteHashRingCommitted {
+            return Ok(());
+        }
+
+        // - but resharding can't be aborted, after write hash ring has been committed
+        Err(CollectionError::bad_request(format!(
+            "can't abort resharding {resharding_key}, \
+             because write hash ring has been committed already, \
+             resharding have to be driven to completion",
```
timvisee, line 203:
```suggestion
             resharding must be completed",
```

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -176,15 +176,57 @@ impl ShardHolder {
...
         } = resharding_key;
 
         let is_in_progress = match self.resharding_state.read().deref() {
-            Some(state) if state.matches(&resharding_key) => true,
+            Some(state) if state.matches(&resharding_key) => {
+                if !force && state.stage >= ReshardStage::WriteHashRingCommitted {
+                    return Err(CollectionError::bad_request(format!(
+                        "can't abort resharding {resharding_key}, \
+                         because write hash ring has been committed already, \
+                         resharding have to be driven to completion",
```
timvisee, line 224:
```suggestion
                        "can't abort resharding {resharding_key}, \
                         because write hash ring has been committed already, \
                         resharding must be completed",
```

timvisee (approved):
Nice! I see no logic problems. :raised_hands: 

I was thinking about what happens if resharding errors past the point from which we can abort. In that case it basically stops the driver, and retries up to three times. On retry it simply resumes the driver so it'll continue where it left off. After three retries it'll stop the driver and invoke the 'abort resharding' operation. If it's past the abort point this will fail. That means we're left with: 1. resharding active in consensus. 2. no running driver.

I think this is actually good. It means we don't infinitely retry if things are broken. We can restart a node to give it another 3 attempts.

I wonder if it'd be worth it to add some kind of resume operation to allow us to retry an ongoing resharding operation without resharding the node. What do you think?

lib/collection/src/collection/mod.rs
ffuugoo: Sure, but it's not "shards removed from peer", but "the peer".

We can't remove last replica of a shard, so we can't remove the one-and-only new shard (at least, not until it was replicated).

lib/storage/src/content_manager/toc/collection_meta_ops.rs
ffuugoo: Yeah, that's exactly what I wanted to do, but, I think, I'll add it later in a separate PR. Requires fiddling with consensus messages.

timvisee: Sounds good.

ffuugoo:
Added tests, but they seem to be super flaky, suggestions welcome. :/

ffuugoo:
@timvisee discovered that Qdrant sometimes locks-up on resharding abort (not sure/can't recall if on explicit abort only or in generall). I disabled abort resharding test temporarily (to reduce flakiness) and propose to merge this PR first, and then I fix this deadlock post-merge.

ack @generall

PR #4700, opened 2024-07-18
fixup: Implement `iter_random` for `ImmutableIdTracker`

coszio:
I just merged #4619 which conflicted with #4403, this PR fixes it + fixes the case where a segment is empty, it would crash the Uniform distribution builder.


generall (approved):

timvisee:
Thanks for patching it quickly!

PR #4706, opened 2024-07-19
universal-search: Query group API should support lookup_from

agourlay:
This PR adds the support for `lookup_from` in the query group API which was missing from https://github.com/qdrant/qdrant/pull/4656

It enables to reproduce the legacy recommend groups API.

### DIFF
--- a/lib/collection/src/grouping/group_by.rs
+++ b/lib/collection/src/grouping/group_by.rs
@@ -264,6 +264,7 @@ impl From<CollectionQueryGroupsRequest> for GroupRequest {
             score_threshold,
             with_vector,
             with_payload,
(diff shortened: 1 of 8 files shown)


lib/collection/src/grouping/group_by.rs
```diff
@@ -281,7 +282,7 @@ impl From<CollectionQueryGroupsRequest> for GroupRequest {
             params,
             with_vector,
             with_payload,
-            lookup_from: None,
+            lookup_from,
```
agourlay, line 285: :face_with_head_bandage:

coszio (approved):

PR #4712, opened 2024-07-20
New Subset Matching in Array Fields

### LINKED ISSUE #4679: Filtering: Support for Subset Matching in Array Fields
Hello,

Today, I encountered a use case that seems not to be covered by the current filtering capabilities of Qdrant: I would like to express something like: "Given a match query for a field with an array, match all the points that have a subset of the array query for this field."

For example, let's consider I have the following points in my collection:
```
[
  { "id": 1, "city": "London", "colors": ["green", "blue"] },
  { "id": 2, "city": "London", "colors": ["red", "blue"] },
  { "id": 3, "city": "London", "colors": ["blue", "yellow"] },
  { "id": 4, "city": "Berlin", "colors": ["red"] },
  { "id": 5, "city": "Moscow", "colors": ["green", "yellow"] },
  { "id": 6, "city": "Moscow", "colors": ["blue"] }
]
```

Now considering a query like this (which does not exist in Qdrant):
```
POST /collections/{collection_name}/points/scroll
{
    "filter": {
        "must": [
            { "key": "colors", "match_subset": { "value": ["blue", "green", "yellow"] } }
        ]
    }
    ...
}
```

The result would be:
```
[
  { "id": 1, "city": "London", "colors": ["green", "blue"] },
  { "id": 3, "city": "London", "color": ["blue", "yellow"] },
  { "id": 5, "city": "Moscow", "color": ["green", "yellow"] },
  { "id": 6, "city": "Moscow", "color": ["blue"] }
]
```

Add a new filtering clause that allows filtering as shown in the example above. Alternatively, add an argument to the match clause to support this use case (cf. how Elasticsearch handles it below).

- **Compute all possible combinations and combine them with exact matches**. However, this approach has two issues:
    - It could be expensive if we have many elements in the query (10 elements in the query produce 1023 combinations).
    - It seems it is not possible to exact match a query array with a point field array (maybe by combining MUST with a simple match clause?).
- **Apply post-filtering**, but this would require retrieving more points from Qdrant (or performing more queries if we don't get any points in the first call). Also, the built-in filtering is one of the main reasons we chose Qdrant.
- **Switch to another engine**; [it seems Elasticsearch supports this use case](https://stackoverflow.com/questions/72004393/how-to-return-all-documents-that-has-a-subset-of-input-array-in-opensearch).

**Additional context**
/

mautini:
Following the feature request I opened here: #4679, I tried to implement a new subset filter that we could express like this:

> Given a match query for a field with an array, match all the points that have a subset of the array query for this field.

(See the feature request for more information about this new feature.)

- As this is my first contribution to Qdrant and Rust is not my primary programming language, please let me know if I need to change anything.
- I created a PR to update the documentation to include this new filter (https://github.com/qdrant/landing_page/pull/1038).
- Since this filter is crucial for our use case, please let me know if you could consider merging this PR. If not, we could quickly consider alternative solutions.
- I'm not sure if the client libraries are autogenerated from the server or if we need to modify them to include this new feature. If you approve this PR and we need to change all the clients, I would be willing to do it 🙂

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

resolve #4679

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -7121,6 +7121,9 @@
           },
           {
             "$ref": "#/components/schemas/MatchExcept"
+          },
+          {
+            "$ref": "#/components/schemas/MatchSubset"
           }
         ]
       },
@@ -7201,6 +7204,18 @@
         ],
         "properties": {
           "except": {
+            "$ref": "#/components/schemas/AnyVariants"
+          }
+        }
+      },
+      "MatchSubset": {
+        "description": "Should have all values matching any of the given values",
+        "type": "object",
+        "required": [
+          "subset"
+        ],
+        "properties": {
+          "subset": {
             "$ref": "#/components/schemas/AnyVariants"
           }
         }
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -1557,16 +1557,16 @@ impl TryFrom<Match> for segment::types::Match {
                 MatchValue::Integers(ints) => ints.integers.into(),
                 MatchValue::ExceptKeywords(kwds) => {
                     segment::types::Match::Except(kwds.strings.into())
-                },
+                }
                 MatchValue::ExceptIntegers(ints) => {
                     segment::types::Match::Except(ints.integers.into())
-                },
+                }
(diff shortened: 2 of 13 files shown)


mautini:
Closing as we can achieve the same behavior using `must_not` and `except`

PR #4717, opened 2024-07-22
Introduce FieldIndexBuilder

xzfc:
Tracking issue: #4502. Continuation of #4690.

This PR introduces the following items:
- `trait FieldIndexBuilderTrait`.
- `enum FieldIndexBuilder` (contains all `FieldIndexBuilderTrait` implementations).
- `index_builder_selector()` which is similar to `index_selector()`.
- An implementation of `FieldIndexBuilderTrait` for each index type.

Right now, the builders are merely single-line methods that implement the same old logic. In subsequent PRs they will be extended to support mmap indices.

### DIFF
--- a/lib/segment/src/index/field_index/field_index_base.rs
+++ b/lib/segment/src/index/field_index/field_index_base.rs
@@ -1,10 +1,14 @@
 use std::fmt::Formatter;
 
 use common::types::PointOffsetType;
+use delegate::delegate;
 use serde_json::Value;
 
-use super::map_index::MapIndex;
-use super::numeric_index::{NumericIndex, StreamRange};
+use super::binary_index::BinaryIndexBuilder;
+use super::full_text_index::text_index::FullTextIndexBuilder;
+use super::geo_index::GeoMapIndexBuilder;
+use super::map_index::{MapIndex, MapIndexBuilder};
+use super::numeric_index::{NumericIndex, NumericIndexBuilder, StreamRange};
 use crate::common::operation_error::OperationResult;
 use crate::common::Flusher;
 use crate::data_types::order_by::OrderValue;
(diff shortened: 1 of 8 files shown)


lib/segment/src/index/field_index/field_index_base.rs
```diff
@@ -343,6 +347,62 @@ impl FieldIndex {
...
 
+pub trait FieldIndexBuilderTrait {
+    /// The resulting type of the index
+    type FieldIndexType;
+
+    fn init(&mut self) -> OperationResult<()>;
+
+    fn add_point(&mut self, id: PointOffsetType, payload: &[&Value]) -> OperationResult<()>;
+
+    fn finalize(self) -> OperationResult<Self::FieldIndexType>;
```
IvanPleshkov, line 358: I propose to use build function as a construction method as a more common case

timvisee (approved):

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -175,19 +177,23 @@ impl StructPayloadIndex {
         payload_schema: PayloadFieldSchema,
     ) -> OperationResult<Vec<FieldIndex>> {
         let payload_storage = self.payload.borrow();
-        let mut field_indexes = index_selector(field, &payload_schema, self.db.clone(), true);
-        for index in &field_indexes {
-            index.recreate()?;
+        let mut builders = index_builder_selector(field, &payload_schema, self.db.clone());
+        for index in &mut builders {
+            index.init()?;
```
IvanPleshkov, line 182: We can remove `recreate` from all field indices now

IvanPleshkov (approved):

lib/segment/src/index/struct_payload_index.rs
IvanPleshkov: resolve this thread because we need this changes asap and `recreate` exists in tests

PR #4724, opened 2024-07-22
Fix hash ring rebuilding and loading while resharding

timvisee:
Tracked in: #4213 
Depends on: #4697

Update hash ring rebuilding to incorporate the resharding state.

Also fixes a few other hash ring state issues when reloading from disk.

### Tasks
- [x] Merge #4697
- [x] Rebase on `dev`
- [x] Undraft

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/hash_ring.rs
+++ b/lib/collection/src/hash_ring.rs
@@ -281,8 +281,8 @@ impl<T: Hash + Copy> HashRing<T> {
         T: std::cmp::PartialEq,
     {
         match self {
-            Inner::Raw(ring) => ring.clone().into_iter().any(|node| node == *shard),
-            Inner::Fair { ring, .. } => ring.clone().into_iter().any(|(node, _)| node == *shard),
+            HashRing::Raw(ring) => ring.clone().into_iter().any(|node| node == *shard),
+            HashRing::Fair { ring, .. } => ring.clone().into_iter().any(|(node, _)| node == *shard),
         }
     }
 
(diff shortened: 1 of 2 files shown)


lib/collection/src/hash_ring.rs
```diff
@@ -58,15 +58,27 @@ impl<T: Hash + Copy + PartialEq> HashRing<T> {
     pub fn add(&mut self, shard: T) {
         match self {
             Self::Single(ring) => ring.add(shard),
-            Self::Resharding { old, new } => {
-                if new.get(&shard).is_none() {
-                    old.add(shard);
+            Self::Resharding { new, old: _ } => {
+                if !new.contains_shard(&shard) {
```
timvisee, line 62: This check was incorrect.

`get(point_id) -> target_shard` returns the target shard, and we gave it the shard ID as source. I've updated this with a new method actually checking whether the shard is already in the ring.

In practice we should never reach this branch while resharding, so I've added a debug assertion below.

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -207,6 +207,35 @@ impl ShardHolder {
                 .add(*shard_id);
         }
 
+        // Restore resharding hash ring if resharding is active and haven't reached WriteHashRingCommitted yet
```
timvisee, line 210: Above, we construct new hash rings as if no resharding is happening.

When resharding, we expect them to reach a different state though. That's done below.

When resharding up, we should have a resharding hash ring. The old ring should not include the new resharding shard.

When resharding down, we should have a resharding hash ring. The new hash ring should not include the shard that is about to be removed.

timvisee (commented):
Annotated a few important parts:

timvisee:
Superseded by <https://github.com/qdrant/qdrant/pull/5038>

PR #4734, opened 2024-07-23
remove testing non-defragmented segments for non-defragmentation

JojiiOfficial:
Fixes flaky test #4708 

When not defragmented, points in a newly built segment get placed in random order, which means it is possible they are already ordered - especially for segments with only a few points. However we assert that a non-defragmented segment always is unsorted.
This PR removes this part.

I think we can safely remove this part because:
- It was initially just to test that the defragmentation checking-logic was working properly.
- The important part, checking defragmented segments are sorted is untouched.
- I don't think another fix, addressing the actual issue, is worth the work that would be needed to fix it.

If you think it is still important to test the checking logic too, I'll create a fix that preserves the test.

### DIFF
--- a/lib/segment/tests/integration/segment_builder_test.rs
+++ b/lib/segment/tests/integration/segment_builder_test.rs
@@ -91,11 +91,6 @@ fn test_building_new_segment() {
     );
 
     assert_eq!(merged_segment.point_version(3.into()), Some(100));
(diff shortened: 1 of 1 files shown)


timvisee (approved):
Makes sense. Thanks for elaborating!

PR #4738, opened 2024-07-23
UUID payload index

JojiiOfficial:
Adds a specially crafted payload index for UUIDs to improve memory usage.


lib/segment/src/index/field_index/uuid_index.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::common::Flusher;
+use crate::index::field_index::histogram::Numericable;
+use crate::index::field_index::PrimaryCondition;
+use crate::telemetry::PayloadIndexTelemetry;
+use crate::types::{
+    FieldCondition, Match, MatchValue, PayloadKeyType, UuidPayloadKeyType, UuidPayloadType,
+    ValueVariants,
+};
+
+pub struct UuidPayloadIndex {
```
IvanPleshkov, line 25: Are you sure that we need separate structure for uuid? You can use `NumericIndex<u128, UuidPayloadKeyType>` and don't copy builder, delegations etc

JojiiOfficial: The reason I created a separate structure is because [here](https://github.com/qdrant/qdrant/pull/4738/files#diff-3fd7336e71c9c53184049c40a48f91408e1dd7ad63aa49ee2cf91d56d0b1ace7R82) and in estimation we try to cast a keyword into a UUID, which makes sense for UUID index but not for numeric index. 
I thought it would be cleaner to separate this behavior, especially because if a user accidentally passes a valid UUID to a number index, it would be parsed and the key would be compared to the numbers in the number index which doesn't make sense at all.

IvanPleshkov: We have already such behaviour (conversion from string) for datetime index:
https://github.com/qdrant/qdrant/blob/dev/lib/segment/src/index/field_index/numeric_index/mod.rs#L563
It seems you may only implement `impl ValueIndexer for NumericIndex<u128, UuidPayloadKeyType>` instead of new entity

lib/segment/src/index/field_index/field_index_base.rs
```diff
@@ -114,6 +114,7 @@ pub enum FieldIndex {
     GeoIndex(GeoMapIndex),
     FullTextIndex(FullTextIndex),
     BinaryIndex(BinaryIndex),
+    UuidIndex(NumericIndex<UuidPayloadKeyType, UuidPayloadType>),
```
coszio, line 117: After digging down into the encoding part I understood where the `Key` name came from, but in my mind it was conflicting with `PayloadKeyType`. Maybe a better name would be `UuidIntType`/`IntUuidType`?

docs/grpc/docs.md
```diff
@@ -1446,6 +1448,21 @@ Note: 1kB = 1 vector of size 256. |
...
 
+<a name="qdrant-UuidIndexParams"></a>
+
+### UuidIndexParams
+
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| is_tenant | [bool](#bool) | optional | If true - used for tenant optimization. |
```
skeptrunedev, line 1459: Amazing!

IvanPleshkov (dismissed):

agourlay:
> Adds a specially crafted payload index for UUIDs to improve memory usage.

Please share some performance numbers to validate that it does what it expected :)

JojiiOfficial:
## Setup
- 100k vectors with unique Uuids

## Allocations after startup
### Uuid index
image

### Keyword index
image


## Flamegraph 
### Uuid index
image

### Keyword index
image

## Bfb following

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -254,7 +269,7 @@ impl<T: Encodable + Numericable + Default> NumericIndexInner<T> {
         let range = match range {
             RangeInterface::Float(float_range) => float_range.map(T::from_f64),
             RangeInterface::DateTime(datetime_range) => {
-                datetime_range.map(|dt| T::from_i64(dt.timestamp()))
+                datetime_range.map(|dt| T::from_i128(i128::from(dt.timestamp())))
```
generall, line 272: why DateTime index is affected by this change?

JojiiOfficial: Because I updated [this](https://github.com/qdrant/qdrant/pull/4738/files#diff-5adcfd0e240fb6450bb128fad52ac82295ebb05f83cb4c55c05de70593e69db9R57) trait function in order to be able to pass UUIDs. The DateTime change is just to make it compatible with the updated trait.

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -325,6 +340,43 @@ impl<T: Encodable + Numericable + Default> NumericIndexInner<T> {
...
+                    (None, _) => 0,
+                }
+            }
+            NumericIndexInner::Immutable(immutable) => {
+                let start_index = immutable.map.find_start_index(start);
+                let end_index = immutable.map.find_end_index(start_index, end);
+                let range_size = end_index - start_index;
+                let avg_values_per_point =
+                    self.total_unique_values_count() as f32 / self.get_points_count() as f32;
+                (range_size as f32 / avg_values_per_point).round() as usize
```
IvanPleshkov, line 376: When `range_size` is `1` and `avg_values` is `3`, the result will be `0`. I propose to avoid zero returning if `range_size` is not zero

lib/segment/src/segment_constructor/segment_builder.rs
```diff
@@ -191,6 +191,14 @@ impl SegmentBuilder {
                     }
                     break;
                 }
+                FieldIndex::UuidIndex(index) => {
+                    if let Some(ids) = index.get_values(internal_id) {
+                        for id in ids {
+                            ordering = ordering.wrapping_add(id as u64);
```
generall, line 197: It looks like i128 -> u64 conversion goes by lower bytes. In case of UUID it might be better to order by higher, more valuable bytes.

Experiments: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bfaa47e0247696885bf6d7a93f1b66aa

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -325,6 +340,43 @@ impl<T: Encodable + Numericable + Default> NumericIndexInner<T> {
     pub fn values_is_empty(&self, idx: PointOffsetType) -> bool {
         self.values_count(idx) == 0
     }
+
+    pub fn values_by_key(&self, key: &T) -> Box<dyn Iterator<Item = PointOffsetType> + '_> {
```
generall, line 344: Naming might be confusing.

We iterate points by payload values, so why don't we call it `point_ids_by_value` instead?

lib/segment/src/segment_constructor/segment_builder.rs
JojiiOfficial: Are you sure, the first bits of a Uuid aren't the ones storing the timestamp?
Given the table in [Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier#Binary_wire_format) and the uuid [crates source code](https://docs.rs/uuid/latest/src/uuid/lib.rs.html#627), it seems to be the case as well.
Please correct me if I'm reading those sources wrong.

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -325,6 +340,46 @@ impl<T: Encodable + Numericable + Default> NumericIndexInner<T> {
...
+        let start = Bound::Included(Point::new(*key, PointOffsetType::MIN));
+        let end = Bound::Included(Point::new(*key, PointOffsetType::MAX));
+        match &self {
+            NumericIndexInner::Mutable(mutable) => Box::new(mutable.values_range(start, end)),
+            NumericIndexInner::Immutable(immutable) => Box::new(immutable.values_range(start, end)),
+        }
+    }
+
+    /// Tries to estimate the amount of points for a given key.
+    pub fn estimate_points(&self, key: &T) -> usize {
```
IvanPleshkov, line 354: I guess some unit tests for numeric index index changes are suitable. Not required for this PR IMHO but we need python integration test at least

lib/segment/src/segment_constructor/segment_builder.rs
generall: The alternative would be to change ordering value to 128 bit

generall: Also ordering by timestamp seems logical, if we have option between ordering by random or by timestamp, I would prefer timestamp

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -325,6 +340,46 @@ impl<T: Encodable + Numericable + Default> NumericIndexInner<T> {
     pub fn values_is_empty(&self, idx: PointOffsetType) -> bool {
         self.values_count(idx) == 0
     }
+
+    pub fn point_ids_by_value(&self, key: &T) -> Box<dyn Iterator<Item = PointOffsetType> + '_> {
```
generall, line 344:
```suggestion
    pub fn point_ids_by_value(&self, value: &T) -> Box<dyn Iterator<Item = PointOffsetType> + '_> {
```

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -325,6 +340,46 @@ impl<T: Encodable + Numericable + Default> NumericIndexInner<T> {
...
+        let start = Bound::Included(Point::new(*key, PointOffsetType::MIN));
```
generall, line 345:
```suggestion
        let start = Bound::Included(Point::new(*value, PointOffsetType::MIN));
```

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -325,6 +340,46 @@ impl<T: Encodable + Numericable + Default> NumericIndexInner<T> {
...
+        let end = Bound::Included(Point::new(*key, PointOffsetType::MAX));
```
generall, line 346:
```suggestion
        let end = Bound::Included(Point::new(*value, PointOffsetType::MAX));
```

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -325,6 +340,46 @@ impl<T: Encodable + Numericable + Default> NumericIndexInner<T> {
...
+        match &self {
+            NumericIndexInner::Mutable(mutable) => Box::new(mutable.values_range(start, end)),
+            NumericIndexInner::Immutable(immutable) => Box::new(immutable.values_range(start, end)),
+        }
+    }
+
+    /// Tries to estimate the amount of points for a given key.
+    pub fn estimate_points(&self, key: &T) -> usize {
```
generall, line 354:
```suggestion
    pub fn estimate_points(&self, value: &T) -> usize {
```

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -325,6 +340,46 @@ impl<T: Encodable + Numericable + Default> NumericIndexInner<T> {
...
+        let start = Bound::Included(Point::new(*key, PointOffsetType::MIN));
```
generall, line 355:
```suggestion
        let start = Bound::Included(Point::new(*value, PointOffsetType::MIN));
```

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -325,6 +340,46 @@ impl<T: Encodable + Numericable + Default> NumericIndexInner<T> {
...
+        let end = Bound::Included(Point::new(*key, PointOffsetType::MAX));
```
generall, line 356:
```suggestion
        let end = Bound::Included(Point::new(*value, PointOffsetType::MAX));
```

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -465,20 +520,32 @@ impl<T: Encodable + Numericable + Default> PayloadFieldIndex for NumericIndexInn
...
         condition: &FieldCondition,
     ) -> Option<Box<dyn Iterator<Item = PointOffsetType> + '_>> {
+        if let Some(Match::Value(MatchValue {
+            value: ValueVariants::Keyword(keyword),
+        })) = &condition.r#match
+        {
+            let keyword = keyword.as_str();
+
+            if let Ok(uuid) = Uuid::from_str(keyword) {
+                let key = T::from_i128(uuid.as_u128() as UuidIntType);
```
generall, line 530:
```suggestion
                let value = T::from_i128(uuid.as_u128() as UuidIntType);
```

lib/segment/src/segment_constructor/segment_builder.rs
JojiiOfficial: Yeah I agree. I'll update it to using `get_timestamp` (and  `as_fields` as fallback).

generall (approved):

PR #4741, opened 2024-07-24
When resharding down, migrate points with shard transfers

timvisee:
Tracked in: #4213 
Depends on: #4697

_WIP_

Revise point migration transfers when resharding down. Rather than a manual scroll+upsert loop, this utilizes shard transfers to make the point migrations. The catch here is that it needs a transfer to every replica separately.

Resharding down with 3 shards and 3 replicas on each would mean 2*3 migration transfers.

Another problem is that the these transfers currently switch the active replicas into `Resharding` state. That means that **all** shards changed into this state at some point during resharding. We can't leave them `Active`, because our shard transfer logic does not support that right now. I'm not sure if this is easily fixable. Ideally we don't touch their state at all and leave them `Active`.

### Open problems
- [ ] Shard holder reading for getting replicas to migrate is currently badly implemented
- [ ] Target replicas are switched into `Resharding` state during transfer

### Tasks
- [ ] Merge #4697
- [ ] Rebase on `dev`
- [ ] Undraft

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1126,6 +1126,7 @@ dependencies = [
  "serde",
  "serde_cbor",
  "serde_json",
+ "serde_variant",
  "sha2",
  "smallvec",
  "sparse",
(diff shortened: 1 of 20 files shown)


timvisee:
Not needed anymore since we move most of this logic into cluster manager.

PR #4742, opened 2024-07-24
Use filed index builder in numeric index tests

IvanPleshkov:
This PR changes the construction of the numeric index in tests. It's a preparation step to add mmap numeric index tests which have to be constructed using builder

### DIFF
--- a/lib/segment/src/index/field_index/numeric_index/tests.rs
+++ b/lib/segment/src/index/field_index/numeric_index/tests.rs
@@ -10,56 +10,90 @@ use crate::json_path::JsonPath;
 
 const COLUMN_NAME: &str = "test";
 
-fn get_index() -> (TempDir, NumericIndexInner<f64>) {
+#[derive(Clone, Copy)]
+enum IndexType {
+    Mutable,
+    Immutable,
+}
+
+enum IndexBuilder {
+    Mutable(NumericIndexBuilder<FloatPayloadType, FloatPayloadType>),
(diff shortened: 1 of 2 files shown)


lib/segment/src/index/field_index/numeric_index/tests.rs
```diff
@@ -209,40 +246,27 @@ fn test_payload_blocks_small(#[case] immutable: bool) {
...
-        index
-    };
+    values.into_iter().enumerate().for_each(|(idx, values)| {
+        let values = values.iter().map(|v| Value::from(*v)).collect_vec();
+        let values = values.iter().collect_vec();
+        index_builder
+            .add_point(idx as PointOffsetType + 1, &values)
+            .unwrap();
+    });
+    let index = index_builder.finalize().unwrap();
```
xzfc, line 256: Maybe it would be more convenient/concise to create a helper like this and use it instead of calling `add_point()`/`finalize()` in each test.

```rust
pub trait FieldIndexBuilderTrait {
    …
    #[cfg(test)]
    fn make_from_iter(mut self, it: impl Iterator<…>) -> OperationResult<Self::FieldIndexType> { … }
}
```

xzfc (approved):

lib/segment/src/index/field_index/numeric_index/tests.rs
IvanPleshkov: We may do it if other indices require such helper too

PR #4743, opened 2024-07-24
Remove NumericIndexKey and use histogram point

IvanPleshkov:
This PR removes `NumericIndexKey` structure and replace it by `Point` from histogram. The main goal of this PR is to unify type of id+value pair and don't define such pair multiple times with conversions to each other. All indices (Mutable+Immutable+MMap) will use only one type of id+value pair

### DIFF
--- a/lib/segment/src/index/field_index/histogram.rs
+++ b/lib/segment/src/index/field_index/histogram.rs
@@ -23,7 +23,7 @@ pub struct Counts {
     pub right: usize,
 }
 
-#[derive(PartialEq, Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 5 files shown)


lib/segment/src/index/field_index/histogram.rs
```diff
@@ -24,16 +24,23 @@ pub struct Counts {
 }
 
 #[derive(PartialEq, PartialOrd, Debug, Clone, Serialize, Deserialize)]
+#[repr(C)]
```
IvanPleshkov, line 27: The goal is using `Point` in all numeric indices even in mmap

xzfc (approved):

PR #4744, opened 2024-07-24
Remove vec u8 from mutable numeric index

IvanPleshkov:
This PR removes `Vec<u8>` as numeric type in `MutableNumericIndex`. It's a part of mmap index because mmap builder will use the mutable index as data collector. Use `Point<T>` as unified type for id+value in numeric index

### DIFF
--- a/lib/segment/src/index/field_index/numeric_index/immutable_numeric_index.rs
+++ b/lib/segment/src/index/field_index/numeric_index/immutable_numeric_index.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::BTreeSet;
 use std::ops::Bound;
 use std::sync::Arc;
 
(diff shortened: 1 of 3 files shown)


lib/segment/src/index/field_index/numeric_index/immutable_numeric_index.rs
```diff
@@ -311,37 +304,23 @@ impl<T: Encodable + Numericable + Default> ImmutableNumericIndex<T> {
...
     use std::ops::Bound;
-    use std::ops::Bound::{Excluded, Included, Unbounded};
 
     use super::*;
     use crate::types::FloatPayloadType;
 
     fn check_range(
         key_set: &NumericKeySortedVec<FloatPayloadType>,
-        encoded_map: &BTreeMap<Vec<u8>, PointOffsetType>,
+        encoded_map: &BTreeSet<Point<FloatPayloadType>>,
```
IvanPleshkov, line 313: Main change here

lib/segment/src/index/field_index/numeric_index/immutable_numeric_index.rs
```diff
@@ -437,71 +416,30 @@ mod tests {
...
         let deleted_key = Point::new(-5.0, 1);
         set_keys.remove(deleted_key.clone());
-        set_byte.remove(&deleted_key.val.encode_key(deleted_key.idx));
+        set_byte.remove(&deleted_key);
 
         check_ranges(&set_keys, &set_byte);
     }
-
-    #[test]
-    fn test_numeric_index_key_sorting() {
```
IvanPleshkov, line 476: This test is not required anymore. We don't need to test btreeset and btreeset

xzfc (approved):

PR #4753, opened 2024-07-25
Facets in segment

coszio:
Adds implementation of internals for getting counts of each value in a map index.

Currently enabled for:
- Keyword index

### DIFF
--- a/lib/segment/src/segment.rs
+++ b/lib/segment/src/segment.rs
@@ -1509,7 +1509,7 @@ impl SegmentEntry for Segment {
             let filter_cardinality = payload_index.estimate_cardinality(filter);
 
             let hits_map = payload_index
-                .iter_filtered_points(filter, &id_tracker, &filter_cardinality)
+                .iter_filtered_points(filter, &*id_tracker, &filter_cardinality)
                 .fold(HashMap::new(), |mut map, point_id| {
                     for value in facet_index.get_values(point_id) {
                         if let Some(counter) = map.get_mut(&value) {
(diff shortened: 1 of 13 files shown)


lib/segment/src/segment.rs
```diff
@@ -1492,6 +1495,33 @@ impl SegmentEntry for Segment {
...
+            .get(&request.key)
+            .and_then(|index| index.iter().find_map(|index| index.as_map_index()))
+            .ok_or_else(|| OperationError::MissingMapIndexForFacet {
+                key: request.key.to_string(),
+            })?;
+
+        let hits = if let Some(filter) = &request.filter {
+            let struct_filter_context = payload_index.struct_filtered_context(filter);
+            let check_counter = |iter: IdRefIter| {
+                iter.filter(|point_id| struct_filter_context.check(**point_id))
```
generall, line 1512: This approach will check each point in the storage regardless of the filter cardinality, right? 

I propose to do it other way: read all ids by filter, retrieve associated payload value from payload index, group payload values into hashmap counter.

In this way, scanning of the whole dataset will only be a worst case, if filter doesn't restrict anything.

Reading of the payload values from index should be fast enough.

coszio:
Already flaky? no way 🔨

lib/segment/src/segment.rs
coszio: Good proposal, I implemented it now in https://github.com/qdrant/qdrant/pull/4753/commits/a68e8033750ea0db50beb5a282de427a536de0b9.

As a note we discussed in chat: this forces us to create a hashmap possibly as big as the cardinality of values we have in the collection for each call. This might be improved with other probabilistic structure, like min-count sketch

generall (approved):

PR #4754, opened 2024-07-26
Pass is_appendable to index_builder_selector()

xzfc:
This PR adds `is_appendable` to `index_builder_selector()`. Currently unused, but it will be used in subsequent PRs: the `on_disk` parameter should be ignored for appendable indices.


xzfc:
I found out that it could be simpler than that.

PR #4755, opened 2024-07-26
Add {Integer,Float,Datetime}IndexParams::on_disk to the API

xzfc:
This PR enables `on_disk` boolean parameter for the payload types affected by map and numeric indices. Namely, for Keyword, Integer, Float, Datetime.

Additionally, this PR adds `PayloadFieldSchema::is_on_disk()` method to get this value regardless of field schema type.

### DIFF
--- a/lib/api/src/grpc/proto/collections.proto
+++ b/lib/api/src/grpc/proto/collections.proto
@@ -387,15 +387,18 @@ enum TokenizerType {
 
 message KeywordIndexParams {
     optional bool is_tenant = 3; // If true - used for tenant optimization.
+    optional bool on_disk = 4; // If true - store index on disk.
 }
 
 message IntegerIndexParams {
   bool lookup = 1; // If true - support direct lookups.
   bool range = 2; // If true - support ranges filters.
   optional bool is_tenant = 3; // If true - used for tenant optimization.
+  optional bool on_disk = 4; // If true - store index on disk.
 }
 
 message FloatIndexParams {
+  optional bool on_disk = 1; // If true - store index on disk.
 }
 
 message GeoIndexParams {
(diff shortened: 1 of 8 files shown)


lib/api/src/grpc/proto/collections.proto
```diff
@@ -387,15 +387,18 @@ enum TokenizerType {
 
 message KeywordIndexParams {
     optional bool is_tenant = 3; // If true - used for tenant optimization.
+    optional bool on_disk = 4; // If true - store index on disk.
```
IvanPleshkov, line 390: While keyword is not ready, I propose to remove it for this PR and add later

lib/segment/src/types.rs
```diff
@@ -1184,6 +1184,18 @@ impl PayloadSchemaParams {
...
+    pub fn is_on_disk(&self) -> bool {
+        match self {
+            PayloadSchemaParams::Keyword(i) => i.on_disk.unwrap_or_default(),
+            PayloadSchemaParams::Integer(i) => i.on_disk.unwrap_or_default(),
+            PayloadSchemaParams::Float(i) => i.on_disk.unwrap_or_default(),
+            PayloadSchemaParams::Geo(_) => false,
+            PayloadSchemaParams::Text(_) => false,
+            PayloadSchemaParams::Bool(_) => false,
+            PayloadSchemaParams::Datetime(i) => i.on_disk.unwrap_or_default(),
+        }
```
IvanPleshkov, line 1197: I propose to use `_ => false,` for Geo, Text and Bool

IvanPleshkov (approved):
I'm not sure if merging API changes without a ready function is a good idea.
But I place approve because we need this feature ASAP

PR #4764, opened 2024-07-29
Mmap hashmap iterate over keys

IvanPleshkov:
This PR adds iterator over keys

### DIFF
--- a/lib/common/common/src/mmap_hashmap.rs
+++ b/lib/common/common/src/mmap_hashmap.rs
@@ -88,8 +88,9 @@ impl<K: Key + ?Sized> MmapHashMap<K> {
 
         // 5. Data
         let mut buckets = vec![0 as BucketOffset; keys_count];
-        let mut last_bucket = 0;
+        let mut last_bucket = 0usize;
         for (k, v) in map.clone() {
+            last_bucket = last_bucket.next_multiple_of(K::align());
             buckets[phf.get(k).expect("Key not found in phf") as usize] =
                 last_bucket as BucketOffset;
             last_bucket += Self::entry_bytes(k, v.len()).0;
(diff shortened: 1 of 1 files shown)


lib/common/common/src/mmap_hashmap.rs
```diff
@@ -4,7 +4,7 @@ use std::fs::File;
 use std::hash::Hash;
 use std::io::{self, Cursor, Write};
 use std::marker::PhantomData;
-use std::mem::size_of;
+use std::mem::{align_of, size_of};
```
xzfc: Nit: starting from Rust 1.80, we don't need to import both of these.
https://releases.rs/docs/1.80.0/#libraries

I've pushed the fix.

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -186,31 +193,20 @@ impl<K: Key + ?Sized> MmapHashMap<K> {
         self.header.buckets_count as usize
     }
 
+    pub fn keys(&self) -> impl Iterator<Item = &K> {
+        (0..self.keys_count()).filter_map(|i| {
+            let entry = self.get_entry(i).ok()?;
```
xzfc, line 197: Maybe we can squeeze a bit more performance if we'll seek over entries contents only rather than looking at the buckets table.

But we can do it later as a separate PR as further optimization.

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -294,6 +318,15 @@ impl Key for str {
         //    validity of `str`/`String` types.
         buf.get(..self.len()) == Some(AsBytes::as_bytes(self)) && buf.get(self.len()) == Some(&0xFF)
     }
+
+    fn align() -> usize {
```
xzfc: Nit: we can replace this method it with an associated constant.

I've pushed the fix.

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -186,31 +192,20 @@ impl<K: Key + ?Sized> MmapHashMap<K> {
         self.header.buckets_count as usize
     }
 
+    pub fn keys(&self) -> impl Iterator<Item = &K> {
```
xzfc, line 195: We need to add this to the benchmark.

xzfc (approved):

lib/common/common/src/mmap_hashmap.rs
IvanPleshkov: I propose to keep just test coverage, I find this part not performance critical

PR #4779, opened 2024-07-30
Mmap map index

IvanPleshkov:
Mmap map index with it's builder

### DIFF
--- a/lib/common/common/src/mmap_hashmap.rs
+++ b/lib/common/common/src/mmap_hashmap.rs
@@ -440,6 +440,11 @@ mod tests {
         assert_eq!(mmap.keys_count(), map.len());
         assert_eq!(mmap.keys().count(), map.len());
 
+        for (k, v) in mmap.iter() {
+            let v = v.iter().copied().collect::<BTreeSet<_>>();
+            assert_eq!(map.get(&from_ref(k)).unwrap(), &v);
+        }
+
         // Existing keys should return the correct values
         for (k, v) in map {
             assert_eq!(
--- a/lib/common/common/src/mmap_hashmap.rs
+++ b/lib/common/common/src/mmap_hashmap.rs
@@ -200,6 +200,15 @@ impl<K: Key + ?Sized> MmapHashMap<K> {
         })
     }
 
+    pub fn iter(&self) -> impl Iterator<Item = (&K, &[PointOffsetType])> {
+        (0..self.keys_count()).filter_map(|i| {
+            let entry = self.get_entry(i).ok()?;
+            let key = K::from_bytes(entry)?;
+            let values = Self::get_values_from_entry(entry, key).ok()?;
+            Some((key, values))
+        })
+    }
+
     /// Get the values associated with the `key`.
     pub fn get(&self, key: &K) -> io::Result<Option<&[PointOffsetType]>> {
         let Some(hash) = self.phf.get(key) else {
@@ -207,11 +216,16 @@ impl<K: Key + ?Sized> MmapHashMap<K> {
         };
 
         let entry = self.get_entry(hash as usize)?;
-        let entry_start = entry.as_ptr() as usize;
 
         if !key.matches(entry) {
             return Ok(None);
         }
+
+        Ok(Some(Self::get_values_from_entry(entry, key)?))
+    }
+
+    fn get_values_from_entry<'a>(entry: &'a [u8], key: &K) -> io::Result<&'a [PointOffsetType]> {
+        let entry_start = entry.as_ptr() as usize;
         let entry = &entry[key.write_bytes()..];
 
         let padding = {
@@ -229,7 +243,7 @@ impl<K: Key + ?Sized> MmapHashMap<K> {
             )
             .ok_or(io::ErrorKind::InvalidData)?;
         let result = PointOffsetType::slice_from(entry).ok_or(io::ErrorKind::InvalidData)?;
-        Ok(Some(result))
+        Ok(result)
     }
 
     fn get_entry(&self, index: usize) -> io::Result<&[u8]> {
--- a/lib/segment/src/index/field_index/map_index/mmap_map_index.rs
+++ b/lib/segment/src/index/field_index/map_index/mmap_map_index.rs
@@ -1,4 +1,6 @@
+use std::borrow::Borrow;
 use std::iter;
+use std::mem::size_of;
 use std::path::{Path, PathBuf};
 
 use ahash::HashMap;
@@ -79,7 +81,7 @@ impl<N: MapIndexKey + Key + ?Sized> MmapMapIndex<N> {
             &hashmap_path,
             values_to_points
                 .iter()
-                .map(|(value, ids)| (<N as MapIndexKey>::from_owned(value), ids.iter().copied())),
+                .map(|(value, ids)| (value.borrow(), ids.iter().copied())),
         )?;
 
         MmapPointToValues::<N>::from_iter(
@@ -89,19 +91,18 @@ impl<N: MapIndexKey + Key + ?Sized> MmapMapIndex<N> {
                     idx as PointOffsetType,
                     values
                         .iter()
-                        .map(|value| N::into_referenced(N::from_owned(value))),
+                        .map(|value| N::into_referenced(value.borrow())),
                 )
             }),
         )?;
 
         {
-            const BITS_IN_BYTE: usize = 8;
             let deleted_flags_count = point_to_values.len();
             let deleted_file = create_and_ensure_length(
                 &deleted_path,
-                BITS_IN_BYTE
-                    * BITS_IN_BYTE
-                    * deleted_flags_count.div_ceil(BITS_IN_BYTE * BITS_IN_BYTE),
+                deleted_flags_count
+                    .div_ceil(u8::BITS as usize)
+                    .next_multiple_of(size_of::<usize>()),
             )?;
             let mut deleted_mmap = unsafe { MmapMut::map_mut(&deleted_file)? };
             deleted_mmap.fill(0);
@@ -141,9 +142,11 @@ impl<N: MapIndexKey + Key + ?Sized> MmapMapIndex<N> {
 
     pub fn remove_point(&mut self, idx: PointOffsetType) {
         let idx = idx as usize;
-        if idx < self.deleted.len() && !self.deleted.get(idx).unwrap_or(true) {
-            self.deleted.set(idx, true);
-            self.deleted_count += 1;
+        if let Some(deleted) = self.deleted.get(idx) {
+            if !deleted {
+                self.deleted.set(idx, true);
+                self.deleted_count += 1;
+            }
         }
     }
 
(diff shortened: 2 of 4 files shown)


lib/segment/src/index/field_index/map_index/mmap_map_index.rs
```diff
@@ -0,0 +1,226 @@
...
+                    idx as PointOffsetType,
+                    values
+                        .iter()
+                        .map(|value| N::into_referenced(N::from_owned(value))),
+                )
+            }),
+        )?;
+
+        {
+            const BITS_IN_BYTE: usize = 8;
```
xzfc, line 98: Use `u8::BITS`.

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
```diff
@@ -0,0 +1,226 @@
...
+            let deleted_flags_count = point_to_values.len();
+            let deleted_file = create_and_ensure_length(
+                &deleted_path,
+                BITS_IN_BYTE
+                    * BITS_IN_BYTE
+                    * deleted_flags_count.div_ceil(BITS_IN_BYTE * BITS_IN_BYTE),
```
xzfc, line 104: Not sure what happen here. Maybe you want one of the following?

```rust
deleted_flags_count.div_ceil(u8::BITS)
// or
deleted_flags_count.div_ceil(u8::BITS).next_multiple_of(size_of::<usize>())
```

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
```diff
@@ -0,0 +1,226 @@
...
+            self.path.join(DELETED_PATH),
+            self.path.join(CONFIG_PATH),
+        ];
+        files.extend(self.point_to_values.files());
+        files
+    }
+
+    pub fn remove_point(&mut self, idx: PointOffsetType) {
+        let idx = idx as usize;
+        if idx < self.deleted.len() && !self.deleted.get(idx).unwrap_or(true) {
```
xzfc, line 144: Nit: I think the first check could be dropped?

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
```diff
@@ -0,0 +1,226 @@
...
+            self.deleted.set(idx, true);
+            self.deleted_count += 1;
+        }
+    }
+
+    pub fn check_values_any(&self, idx: PointOffsetType, check_fn: impl Fn(&N) -> bool) -> bool {
+        if self.deleted.get(idx as usize).unwrap_or(true) {
```
xzfc, line 151: Nit: would `.unwrap_or(false)` produce the same result? And same for two methods below.

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
```diff
@@ -0,0 +1,226 @@
...
+            false
+        }
+    }
+
+    pub fn get_values(
+        &self,
+        idx: PointOffsetType,
+    ) -> Option<Box<dyn Iterator<Item = N::Referenced<'_>> + '_>> {
+        if !self.deleted.get(idx as usize).unwrap_or(true) {
+            Some(Box::new(self.point_to_values.get_values(idx)?))
```
xzfc, line 164: Style nit: you could use `bool_value.then(|| …)` instead of `if bool_value { Some(…) } else { None }`.

lib/segment/src/index/field_index/map_index/mod.rs
```diff
@@ -482,7 +575,11 @@ impl PayloadFieldIndex for MapIndex<IntPayloadType> {
     }
 
     fn clear(self) -> OperationResult<()> {
-        self.get_db_wrapper().recreate_column_family()
+        match self {
+            MapIndex::Mutable(index) => index.get_db_wrapper().recreate_column_family(),
+            MapIndex::Immutable(index) => index.get_db_wrapper().recreate_column_family(),
+            MapIndex::Mmap(index) => index.clear(),
+        }
```
xzfc, line 582: Nit: it could be deduplicated into a common code for both `str` and `IntPayloadType`. Same for remove_point.

xzfc (changes requested):

lib/segment/src/index/field_index/map_index/mod.rs
```diff
@@ -645,47 +756,90 @@ mod tests {
...
+
+    fn save_map_index<N>(
+        data: &[Vec<N::Owned>],
+        path: &Path,
+        index_type: IndexType,
+        into_value: impl Fn(&N::Owned) -> Value,
+    ) where
+        N: MapIndexKey + MmapValue + ?Sized,
         MapIndex<N>: PayloadFieldIndex + ValueIndexer,
+        <MapIndex<N> as ValueIndexer>::ValueType: Into<N::Owned>,
```
xzfc, line 782: A lot of type conversion happen here... Like you're using `String` and `SmolStr` at the same time. Maybe you could use just one of them?

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -200,18 +200,32 @@ impl<K: Key + ?Sized> MmapHashMap<K> {
         })
     }
 
+    pub fn iter(&self) -> impl Iterator<Item = (&K, &[PointOffsetType])> {
```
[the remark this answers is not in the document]
IvanPleshkov: added

lib/segment/src/index/field_index/map_index/mod.rs
```diff
@@ -32,14 +35,17 @@ use crate::types::{
...
 
 pub type IdRefIter<'a> = Box<dyn Iterator<Item = &'a PointOffsetType> + 'a>;
 
-pub trait MapIndexKey: Key + Eq + Display + Debug {
+pub trait MapIndexKey: Key + MmapValue + Eq + Display + Debug {
     type Owned: Borrow<Self> + Hash + Eq + Clone + FromStr + Default;
 
     fn to_owned(&self) -> Self::Owned;
+
+    fn from_owned(owned: &Self::Owned) -> &Self;
```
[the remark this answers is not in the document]
IvanPleshkov: thanks! didn't realise that `N` has `Borrow`. done

lib/segment/src/index/field_index/map_index/mod.rs
```diff
@@ -56,14 +66,19 @@ impl MapIndexKey for IntPayloadType {
...
         *self
     }
+
+    fn from_owned(owned: &Self::Owned) -> &Self {
+        owned
+    }
 }
 
-pub enum MapIndex<N: MapIndexKey + ?Sized> {
+pub enum MapIndex<N: MapIndexKey + MmapValue + ?Sized> {
```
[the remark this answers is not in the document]
IvanPleshkov: removed unnecessary restriction everywhere

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
IvanPleshkov: done

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
IvanPleshkov: done

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
IvanPleshkov: we check `len` because we call `set`. refactored to make it more clear

lib/segment/src/index/field_index/map_index/mod.rs
IvanPleshkov: There is a conversion chain: `Value -> ValueType -> N::Owned`. `Value` is json, `ValueType` is `String`, and `N::Owned` is `SmolStr`. We use `SmolStr` here to avoid a lot of conversion while retrieve values and checks

lib/segment/src/index/field_index/map_index/mod.rs
xzfc: `clear()`:
- https://github.com/qdrant/qdrant/blob/63f6cd3dbec8f18942cf9d7ae6ba9d9f47a1503b/lib/segment/src/index/field_index/map_index/mod.rs#L463-L469
- https://github.com/qdrant/qdrant/blob/63f6cd3dbec8f18942cf9d7ae6ba9d9f47a1503b/lib/segment/src/index/field_index/map_index/mod.rs#L577-L583

`remove_point()`:
- https://github.com/qdrant/qdrant/blob/63f6cd3dbec8f18942cf9d7ae6ba9d9f47a1503b/lib/segment/src/index/field_index/map_index/mod.rs#L704-L713
- https://github.com/qdrant/qdrant/blob/63f6cd3dbec8f18942cf9d7ae6ba9d9f47a1503b/lib/segment/src/index/field_index/map_index/mod.rs#L742-L751

I suggest to add these two methods to `impl<N: MapIndexKey + MmapValue + ?Sized> MapIndex<N>`:
https://github.com/qdrant/qdrant/blob/63f6cd3dbec8f18942cf9d7ae6ba9d9f47a1503b/lib/segment/src/index/field_index/map_index/mod.rs#L81

IvanPleshkov: Thanks! Got it, done

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
IvanPleshkov: Refactored to avoid `if`

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
IvanPleshkov: refactored and removed `unwrap_or`

xzfc (approved):

lib/segment/src/index/field_index/map_index/mod.rs
```diff
@@ -362,6 +419,47 @@ where
...
+    }
+
+    fn add_point(&mut self, id: PointOffsetType, payload: &[&Value]) -> OperationResult<()> {
+        let mut flatten_values: Vec<_> = vec![];
+        for value in payload.iter() {
+            let payload_values = <MapIndex<N> as ValueIndexer>::get_values(value);
+            flatten_values.extend(payload_values);
+        }
+        let flatten_values: Vec<N::Owned> = flatten_values.into_iter().map(Into::into).collect();
+        self.point_to_values.resize_with(id as usize + 1, Vec::new);
```
JojiiOfficial, line 446: This part makes building Mmap index very slow if there are many points to insert.
Simply passing the max point id to this function gave me a speedup from ~15-20 minutes to a few seconds.
I'd propose to pass a hint to this function with the largest point id and directly allocating the right size.
If you agree, I can make a PR!

IvanPleshkov: I agree, you can provide max id into the `init` method. You can also create your branch from this to avoid future merge conflicts
https://github.com/qdrant/qdrant/pull/4809

JojiiOfficial: Will do!

PR #4782, opened 2024-07-30
More filtering of points without preallocating

coszio:
Needs #4753 

In #4753 we added `iter_filtered_points` to the payload index struct, which allows us to return the filtered iterator.

This PR switches the use of `query_points` to the new `iter_filterd_points` when possible, to avoid an early filtered points allocation.


JojiiOfficial (approved):

generall (approved):
Avoiding `collect` sometimes leads to performance degradation, we observed in inside HNSW on very hot loops. But in this case, I believe it should be fine.
If not, we should see it on CI benches

PR #4783, opened 2024-07-30
Non blocking filtered deletion - Refactoring suggestions

generall:
- a bit shorter
- do not use unwrap
- use segment ids


generall:
Disregard that, it will lock the same segment over and over, which is essentially what we are trying to avoid.

PR #4785, opened 2024-07-30
Facets in local shard

coszio:
Aggregates faceting from multiple segments inside local shard

### DIFF
--- a/lib/collection/src/shards/local_shard/facet.rs
+++ b/lib/collection/src/shards/local_shard/facet.rs
@@ -1,0 +1,51 @@
+use std::sync::Arc;
+
+use futures::future::try_join_all;
+use itertools::{process_results, Itertools};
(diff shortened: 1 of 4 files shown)


lib/collection/src/shards/local_shard/facet.rs
```diff
@@ -0,0 +1,51 @@
...
+
+                read_segment.facet(&request)
+            })
+        };
+
+        let all_reads = {
+            let segments_lock = self.segments().read();
+
+            try_join_all(
+                segments_lock
```
agourlay, line 33: There is no timeout nor cancellation support here for the spawned computations.
Is this something for later?

coszio: I am leaving it for later :)

agourlay (approved):

PR #4788, opened 2024-07-31
Reduce collection name length limit

### LINKED ISSUE #4568: Cannot delete collection with long name
I'm able to create a collection with long name, but when i try to delete that collection, it throws an error saying "Unexpected Response: 500 (Internal Server Error)
Raw response content:
b'{"status":{"error":"Service internal error: File name too long (os error 36)"},"time":0.064506633}'".

JojiiOfficial:
Fixes #4568 

We can't limit the length of other collection name request parameters as this might make some existing collections not work at all.
To prevent issues in future, we only limit the collection name length to 218

### DIFF
--- a/lib/api/build.rs
+++ b/lib/api/build.rs
@@ -116,7 +116,7 @@ fn configure_validation(builder: Builder) -> Builder {
         .validates(&[
             ("GetCollectionInfoRequest.collection_name", "length(min = 1, max = 255)"),
             ("CollectionExistsRequest.collection_name", "length(min = 1, max = 255)"),
-            ("CreateCollection.collection_name", "length(min = 1, max = 255), custom = \"common::validation::validate_collection_name\""),
+            ("CreateCollection.collection_name", "length(min = 1, max = 218), custom = \"common::validation::validate_collection_name\""),
             ("CreateCollection.hnsw_config", ""),
             ("CreateCollection.wal_config", ""),
             ("CreateCollection.optimizers_config", ""),
--- a/lib/api/src/grpc/qdrant.rs
+++ b/lib/api/src/grpc/qdrant.rs
@@ -435,7 +435,7 @@ pub struct CreateCollection {
     /// Name of the collection
     #[prost(string, tag = "1")]
     #[validate(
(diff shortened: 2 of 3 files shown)


generall:
this is technically a breaking change

timvisee:
I'd also argue that 218 is somewhat arbitrary, and might break again if we have a different base directory.

Instead I'd prefer to investigate what options we have for actually supporting long names. IIRC Windows has some API to toggle support for it, but I don't know what side effects it might have.

gulshan-rs:
Yeah, windows do have support long-names but AFAIK, there's no such support in Linux (for good)

I have below proposals. 

[1]  Hash the long-name using SHA-256, which will produce a fixed 64-character hexadecimal hash.  There are very rare chances of collision post hashing, still, we could append UUID to guarantee uniqueness.   

The only caveat is that the hashing irreversible and debugging for a bug in deletion log becomes hard.

```
get_hashed_deleted_path(storage_path: &str, collection_name: &str) -> PathBuf {

    let removed_collections_path = Path::new(storage_path).join(".deleted");

    let mut hasher = Sha256::new();
    hasher.update(collection_name);
    let hashed_name = format!("{:x}", hasher.finalize());

    let uuid = Uuid::new_v4().to_string();
    removed_collections_path.join(hashed_name).with_extension(uuid)
}
```

[2] We can create sub-paths including UUID as a sub-directory f.e ```./deleted/<uuid>/my_long_file_name``` . The disadvantage would be slower async deletes with increased dir depth.

gulshan-rs:
@generall , could you please review the proposals, i can send a my first PR to this repo :)

timvisee:
If this is just about the file path we temporary use while deleting the collection, I'd vote to just use a random string, since we don't really care about the real name anymore at that point.

For example: `/storage/collections/.deleted/5d41402a`

gulshan-rs:
thanks @timvisee for the proposal, +1 for the simplicity. I opened a PR for this https://github.com/qdrant/qdrant/pull/5439

timvisee:
Superseded by <https://github.com/qdrant/qdrant/pull/5439>.

PR #4789, opened 2024-07-31
Rename tenant to principal for float,int,datetime

JojiiOfficial:
Renames `is_tenant` to `is_principal` in the API.


lib/api/src/grpc/proto/collections.proto
```diff
@@ -394,13 +394,13 @@ message KeywordIndexParams {
 message IntegerIndexParams {
   bool lookup = 1; // If true - support direct lookups.
   bool range = 2; // If true - support ranges filters.
-  optional bool is_tenant = 3; // If true - used for tenant optimization.
+  optional bool is_principal = 3; // If true - used for optimization similar to tenants.
```
generall, line 397:
```suggestion
  optional bool is_principal = 3; // If true - use this key to organize storage of the collection data. This option assumes that this key will be used in majority of filtered requests.
```

generall (approved):

generall (approved):

PR #4792, opened 2024-07-31
Support timeout in Facets

coszio:
Needs #4790 

Support timeout in facet requests, which carries stoppage all the way into segment. 

A new iterator struct has been created to avoid loading an `AtomicBool` excessively on a hot path.


agourlay (approved):

lib/common/common/src/iterator_ext.rs
```diff
@@ -0,0 +1,69 @@
+pub trait IteratorExt: Iterator {
```
agourlay, line 1: Very neat abstraction, gonna use it as soon as it is merged :clap: 

Nitpick: I would love to have a unit test for it for a quick sanity check

PR #4793, opened 2024-08-01
Distance matrix API internals

agourlay:
This PR introduces the internals of the new distance matrix API.

Given a `sample_size` and a `limit_per_sample`, it:
- computes `sample_size` samples
- for each sample, compute the nearest neighbors within the sample
- return all scores

## Future work

- design various output formats
- create new REST and gRPC API to expose the matrix


lib/collection/src/collection/distance_matrix.rs
```diff
@@ -0,0 +1,142 @@
...
+
+use crate::collection::Collection;
+use crate::operations::query_enum::QueryEnum;
+use crate::operations::shard_selector_internal::ShardSelectorInternal;
+use crate::operations::types::{
+    CollectionResult, CoreSearchRequest, CoreSearchRequestBatch, PointRequestInternal,
+};
+
+#[derive(Debug, Default)]
+struct DistanceMatrixResult {
```
coszio, line 18: nit: let's use `*Response` for success response types to not confuse with `*Result` which can have an `Err` variant

lib/collection/src/collection/distance_matrix.rs
```diff
@@ -0,0 +1,142 @@
...
+        limit: usize,
+        filter: Option<&Filter>,
+        using: String,
+        timeout: Option<Duration>,
+    ) -> CollectionResult<DistanceMatrixResult> {
+        if limit == 0 || sample_count == 0 {
+            return Ok(Default::default());
+        }
+
+        let shard_selector = ShardSelectorInternal::All;
```
coszio, line 36: `TODO: implement shard selection`?

lib/collection/src/collection/distance_matrix.rs
```diff
@@ -0,0 +1,142 @@
...
+struct DistanceMatrixResult {
+    sample_ids: Vec<PointIdType>,   // sampled point ids
+    nearest: Vec<Vec<ScoredPoint>>, // nearest points for each sampled point
+}
+
+impl Collection {
+    async fn distance_matrix(
+        &self,
+        sample_count: usize,
+        limit: usize,
```
coszio, line 27: 1. What is the difference between `sample_count` and `limit`?

2. nit: maybe `sample_size` is a better name?

lib/collection/src/collection/distance_matrix.rs
```diff
@@ -0,0 +1,142 @@
...
+                break;
+            }
+            let guard = self.shards_holder.read().await;
+            if let Some(shard) = guard.get_shard(&local_shard) {
+                let sampled = shard
+                    .sample_filtered_points(sample_count, filter, timeout)
+                    .await?;
+                sampled_point_ids.extend(sampled);
+            }
+        }
```
coszio, line 57: Why not use the query api with `ScoringQuery::Sample(Sample::Random)`?

This current implementation also has preference towards the first shard it chooses

lib/collection/src/collection/distance_matrix.rs
agourlay: 1. sample count = number of points to select randomly
    limit = number of neighbors per sample to compute
    
2. maybe :+1:

lib/collection/src/collection/distance_matrix.rs
agourlay: - we want to sample on the local shard only to minimize network access

- regarding the preference of the first shard, I agree, an alternative implementation is to query all shards and then merge, shuffle and truncate the results

agourlay: I decided to use the Query API instead of local sampling + retrieval.
It makes the code much simpler and less weird.

Thank you for the suggestion :+1:

lib/collection/src/collection/distance_matrix.rs
```diff
@@ -0,0 +1,121 @@
...
+                    VectorStructInternal::MultiDense(v) => {
+                        let multi = NamedMultiDenseVector {
+                            name: using.clone(),
+                            vector: v,
+                        };
+                        NamedVectorStruct::from(multi)
+                    }
+                    VectorStructInternal::Named(named_vectors) => {
+                        // expect is safe because we just retrieved the vector
+                        let v = named_vectors.get(&using).expect("Vector not found").clone();
```
coszio, line 75: Not necessarily, since a vector can have (or not) its named vector, isn't it?

I think we need to introduce a `HasVector` condition

agourlay: Good catch :+1:

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -281,10 +281,10 @@ impl LocalShard {
 
         let (availability, mut segments_reads): (Vec<_>, Vec<_>) = all_reads.into_iter().unzip();
 
-        if availability.is_empty() {
+        // Shortcut if all segments are empty
+        if availability.iter().all(|&count| count == 0) {
```
agourlay, line 285: found this bug while testing querying an empty collection

coszio (approved):

PR #4810, opened 2024-08-02
Cleanup `ClusterOperations` enum structure and ser/de implementation

ffuugoo:
`ClusterOperations` uses `#[serde(untagged)]` in a bit of nonsensical way: each variant defines a custom structure with a single field, and this field name is unique across all structures.

Basically, it's just serde's default externally-tagged enum representation with 1) extra steps, 2) worse deserialization performance and 3) terrible error messages.

I've simplified `ClusterOperations` enum back to externally-tagged representation and moved around declarations in `cluster_ops` module to be a bit more organized.

The only potential downside (stopper?) is that even if from Rust POV this new representation is exactly the same (i.e., it serializes to and deserializes from exactly the same JSON), it still changes OpenAPI spec slightly, and I'm not sure if this change would be backward compatible. 🤔

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -10723,40 +10723,99 @@
         }
       },
       "ClusterOperations": {
-        "anyOf": [
+        "oneOf": [
           {
-            "$ref": "#/components/schemas/MoveShardOperation"
+            "description": "Move shard to a different peer",
+            "type": "object",
+            "required": [
+              "move_shard"
+            ],
+            "properties": {
+              "move_shard": {
+                "$ref": "#/components/schemas/MoveShard"
+              }
+            },
+            "additionalProperties": false
           },
           {
-            "$ref": "#/components/schemas/ReplicateShardOperation"
+            "description": "Replicate shard to a different peer",
+            "type": "object",
+            "required": [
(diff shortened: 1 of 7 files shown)


generall (changes requested):
This structure is intentional, as it generates better OpenAPI structures.
We follow this pattern in all untagged enums observed from external interfaces.

This PR breaks OpenAPI and will cause ugly anonymous objects in docs and in clients
Screenshot from 2024-08-07 19-48-36

vs

Screenshot from 2024-08-07 19-48-28

ffuugoo:
> This structure is intentional, as it generates better OpenAPI structures. We follow this pattern in all untagged enums observed from external interfaces.
> 
> This PR breaks OpenAPI and will cause ugly anonymous objects in docs and in clients

Life is a prison 🥲

timvisee:
This is a painful one, OpenAPI hitting us again.

Based on the above I think we can close this. Or @ffuugoo, does this contain something else that is worth merging?

ffuugoo:
Yeah, I think we can close this for now. It's so annoying to work with this god-awful `untagged` enums. 🙄

PR #4815, opened 2024-08-02
Fix reco query tests flakiness

coszio:
Fixes them by relaxing float comparison

### DIFF
--- a/lib/segment/src/vector_storage/query/reco_query.rs
+++ b/lib/segment/src/vector_storage/query/reco_query.rs
@@ -127,9 +127,20 @@ mod test {
         }
     }
 
+    fn ulps_eq(a: f32, b: f32, ulps: u32) -> bool {
+        if a.signum() != b.signum() {
+            return false;
+        }
+
+        let a = a.to_bits();
+        let b = b.to_bits();
+
+        a.abs_diff(b) <= ulps
+    }
+
     /// Relaxes the comparison of floats to allow for a some difference in units of least precision
     fn float_cmp(a: f32, b: f32) -> Ordering {
-        if (a - b).abs() < 0.001 {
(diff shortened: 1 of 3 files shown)


lib/segment/src/vector_storage/query/reco_query.rs
```diff
@@ -125,18 +128,30 @@ mod test {
         }
     }
 
+    /// Relaxes the comparison of floats to allow for a some difference in units of least precision
+    fn float_cmp(a: f32, b: f32) -> Ordering {
+        if approx_eq!(f32, a, b, ulps = 80) {
```
generall, line 133:
```suggestion
        if (a - b).abs() < 0.0001 {
```

coszio: IMO (after just learning about them) units of least precision make more sense, since they are less sensitive to the magnitude of the floats.

generall: 1. I have to google about ulps to understand this
2. `ulps = 80` is as arbitrary as `0.0001`
3. This small function doesn't justify including another crate

coszio: Ok

coszio: I had to relax it to < 0.001 to make the inputs `a = 62.9219, b = 62.921734` pass

agourlay (approved):

agourlay:
CI is still complaining for the macos-build :(

coszio:
So the test is now failing because large magnitude numbers that are not considered Equal, get mapped into the range (0,1) which then get considered as Equal. 

I'll do the ulp implementation without adding the extra crate.

agourlay (approved):

PR #4821, opened 2024-08-03
Add `shell.nix`

xzfc:
This PR adds `shell.nix` which is used to set up a development shell containing all dependencies to develop Qdrant. I'm using it as my primary development environment, and it could be helpful for external contributors who are using Nix.

## What's inside

- Rust toolchain.
- Rust native dependenies (`libunwind`, `protobuf`, and so on).
- Python and its dependenies. It reads `pyproject.toml` and `poetry.lock`, no `poetry install` is needed.
- Dependencies used by various other scripts (`jq`, `curl`, and so on).
- Various hacks to make rebuilds faster:
    - [`mold`](https://github.com/rui314/mold) linker.
    - [`ccache`](https://ccache.dev) for C/C++. Yes, `librocksdb-sys`, I'm looking at you.
    - [`sccache`](https://github.com/mozilla/sccache) for Rust.
    - A variable `LINDERA_CACHE` for `lindera-unidic`.

## How to use

Install nix and run `nix-shell` in the root of the repo.

To update the toolchain and dependencies run `tools/nix/update.py` within that shell.

## What about Nix flake?

Initially, I was using a flake version of this (see https://github.com/xzfc/qdrant-tools). But after moving it to the main repo I've found that `nix develop` wants to copy the whole working directory into the `/nix/store` (about 13 MB). While it's not a stopper, I find it a bit inconvenient for a development shell, as the cache would be invalidated every time you change any of the files. See https://www.github.com/NixOS/nix/issues/3121, https://www.github.com/NixOS/nix/issues/4097, https://www.github.com/NixOS/nix/pull/6530.
`shell.nix` is a more classic approach than `flake.nix`.


generall (approved):
I can only review the readme part of this PR

PR #4823, opened 2024-08-05
Add more missing APIs to the metrics allow list

agourlay:
Initially I wanted to just add the missing entry for the new Query Groups API.

However I realized that there were more API missing so I decided to add a few more for completeness.


KShivendu (approved):
Compared with https://api.qdrant.tech and LGTM

generall (approved):

PR #4825, opened 2024-08-05
On-disk Uuid hybrid-index

JojiiOfficial:
Implements hybrid (numeric+map) Uuid index based on the storage type (disk or memory).
Also enables the `on_disk` api parameter for Uuid index.

### DIFF
--- a/lib/segment/src/index/field_index/index_selector.rs
+++ b/lib/segment/src/index/field_index/index_selector.rs
@@ -149,13 +149,19 @@ impl<'a> IndexSelector<'a> {
                     FieldIndexBuilder::DatetimeMmapIndex,
                 )]
             }
-            PayloadSchemaParams::Uuid(_) => {
-                vec![self.map_builder(
-                    field,
-                    FieldIndexBuilder::UuidIndex,
-                    FieldIndexBuilder::UuidMmapIndex,
-                )]
-            }
+            PayloadSchemaParams::Uuid(_) => match self {
+                IndexSelector::RocksDb(rocksdb) => {
+                    vec![FieldIndexBuilder::UuidIndex(NumericIndex::builder(
+                        rocksdb.db.clone(),
+                        &field.to_string(),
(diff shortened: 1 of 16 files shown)


lib/segment/src/index/struct_payload_index.rs
```diff
@@ -441,6 +441,8 @@ impl PayloadIndex for StructPayloadIndex {
         self.build_and_save(field, payload_schema)?;
         self.save_config()?;
 
+        println!("Indexing done");
```
generall, line 444: debug code

lib/segment/src/index/field_index/index_selector.rs
```diff
@@ -81,7 +81,14 @@ impl<'a> IndexSelector<'a> {
...
                 vec![FieldIndex::DatetimeIndex(self.numeric_new(field)?)]
             }
-            PayloadSchemaParams::Uuid(_) => vec![FieldIndex::UuidIndex(self.numeric_new(field)?)],
+            PayloadSchemaParams::Uuid(_) => match self {
+                IndexSelector::RocksDb(_) => {
+                    vec![FieldIndex::UuidIndex(self.numeric_new(field)?)]
+                }
+                IndexSelector::OnDisk(_) => {
+                    vec![FieldIndex::UuidMapIndex(self.map_new(field)?)]
+                }
```
generall, line 90: I think there was a misunderstanding. I thought that by hybrid index we mean the same approach we use in `int` index, where we use numeric for ranges and map for match.

This approach creates numeric index, this will work for ranges _partially_.

Some segments might be on disk, others are not, so the range request will be executed only on some segments. I don't think user will expect that.

I will submit a review commit, which will only allow `UuidMapIndex` in all cases.
Initially I though to include numeric option as well, but it can be done later and won't give us much benefits, cause we don't have UUID support in range request yet anyway.

generall (approved):

Issue #4797, opened 2024-08-01
Server freezing on large data uploads


jlkravitz:
Large uploads cause server to freeze, perhaps due to read-lock issues.

## Current Behavior

When uploading large collections using the [python qdrant client](https://github.com/qdrant/qdrant-client), the upload generally freezes before finishing. When I look at the output of the qdrant server, I see many messages that say `WARN collection::collection_manager::holders::segment_holder: Trying to read-lock all collection segments is taking a long time. This could be a deadlock and may block new updates.`.

When I try to view the dashboard, it becomes unusable (does not respond).

I then have to restart the qdrant server, which takes a long time to do because we have so much data to recover before it can start accepting API requests.
 
## Steps to Reproduce

1. 
2.
3.
4.

## Expected Behavior

It should not freeze the server, and uploading should not stop before finishing.

## Context (Environment)

I am trying to upload a large dataset to qdrant, and it is becoming exceedingly frustrating to do so.

generall:
Hey @jlkravitz,
could you please describe your hardware configuration and data volume you are talking about (number of vectors, dimentionality, payload, e.t.c.)

jlkravitz:
Hey @generall, hardware info below.  We're on Ubuntu Linux.

In terms of volume, we are uploading ~40M vectors of size 1280. Payload is just an internal ID, stored as a string that is up to 109 characters.

Please let me know if any additional information would be helpful!

<details>
<summary>
<code>CPU</code>
</summary>

```
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         46 bits physical, 57 bits virtual
  Byte Order:            Little Endian
CPU(s):                  224
  On-line CPU(s) list:   0-223
Vendor ID:               GenuineIntel
  Model name:            Intel(R) Xeon(R) Platinum 8480+
    CPU family:          6
    Model:               143
    Thread(s) per core:  2
    Core(s) per socket:  56
    Socket(s):           2
    Stepping:            8
    BogoMIPS:            4000.00
```
</details>

<details>
<summary>
<code>RAM</code>
</summary>

```
               total        used        free      shared  buff/cache   available
Mem:           2.0Ti       1.0Ti       5.0Gi       556Mi       997Gi       992Gi
Swap:          8.0Gi        81Mi       7.9Gi
```
</details>

generall:
Hey @jlkravitz, I think I was able to reproduce the issue.

[...]

PR #4828, opened 2024-08-05
Use mmap lock as default vector storage

generall:
Addresses issue reported in https://github.com/qdrant/qdrant/issues/4797

This solution have a downside:

Then migrated from unix -> windows machine the storage will fallback into just mmap implementation (without forcing of all data in RAM), which might result in high latency of cold requests.

But I assume that unix -> windows migration via snapshots is a rare case, it doesn't cause errors, and it can be fixed by forcing re-indexing, I consider this problem minor. So if benchmark results are satisfactory, I propose to use it.

---

UPD.

Previously described approach have an unforeseen problem of mlock limit, configured by default in linux systems. It seems that default ulimit configuration is quite restrictive, and even assuming we can change that, it most likely will create more unpredictable problems.

So instead of memlocking, we are trying to just have mmap structure for persistence and Chuncked Vectors for reading.

ToDo:

- benchmark
- How to avoid 32mb allocation for (almost)empty segment?
-  Experiment with MEM ADVICE


lib/collection/src/config.rs
```diff
@@ -354,7 +354,11 @@ impl CollectionParams {
                         // Default to in memory storage
                         storage_type: if params.on_disk.unwrap_or_default() {
                             VectorStorageType::ChunkedMmap
+                        } else if cfg!(unix) {
+                            VectorStorageType::LockedChunkedMmap
```
generall, line 358: this is enabling of `LockedChunkedMmap` by default on unix

generall:
Benchmarks:

- upload time

```
docker run --rm -it --network=host qdrant/bfb:dev ./bfb -n 2000000 -d 1280 --indexing-threshold 0
```

old: 5:24
new: 4:20 (same as `on_disk: true`)

- search

within the margin of error - same

- Memory usage

old rss anon: 906Mb
old rss: 990Mb

new rss anon: 650Mb
new rss: 1490Mb

generall:
ToDo: use TKey in ChunkedVectorStorage instead of PointOffset

lib/segment/src/vector_storage/chunked_vectors.rs
```diff
@@ -32,7 +27,7 @@ impl<T: Copy + Clone + Default> ChunkedVectors<T> {
     pub fn new(dim: usize) -> Self {
         assert_ne!(dim, 0, "The vector's dimension cannot be 0");
         let vector_size = dim * mem::size_of::<T>();
-        let chunk_capacity = max(MIN_CHUNK_CAPACITY, CHUNK_SIZE / vector_size);
+        let chunk_capacity = CHUNK_SIZE / vector_size;
```
IvanPleshkov, line 30: This `min` covers case when `CHUNK_SIZE / vector_size` is `0`. How very  large DIM produces errors because of empty `chunk_capacity`. I propose to do the same for mmap chunks

lib/segment/src/vector_storage/in_ram_persisted_vectors.rs
```diff
@@ -0,0 +1,117 @@
...
+
+impl<T: Sized + Copy + Clone + Default + 'static> InRamPersistedVectors<T> {
+    pub fn open(directory: &Path, dim: usize) -> OperationResult<Self> {
+        let mmap_storage = ChunkedMmapVectors::open(directory, dim, Some(false))?;
+
+        let mut vectors = ChunkedVectors::new(dim);
+
+        let total_vectors = mmap_storage.len();
+
+        for key in 0..total_vectors {
```
IvanPleshkov, line 26: because chunk size is equal, can we directly copy the whole file to the chunk without `push`? It might be faster on load

lib/segment/src/vector_storage/in_ram_persisted_vectors.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn push(&mut self, vector: &[T]) -> OperationResult<PointOffsetType> {
+        self.mmap_storage.push(vector)?;
+        let key = self.vectors.push(vector)?;
+        Ok(key)
+    }
+
+    #[inline]
+    fn insert(&mut self, key: PointOffsetType, vector: &[T]) -> OperationResult<()> {
+        self.mmap_storage.insert(key, vector)?;
+        self.vectors.insert(key, vector)?;
```
IvanPleshkov, line 79: I propose to insert to RAM first. If we meet out of RAM, we do actually write to the file here.

lib/segment/src/vector_storage/in_ram_persisted_vectors.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn insert_many<TKey>(
+        &mut self,
+        start_key: TKey,
+        vectors: &[T],
+        count: usize,
+    ) -> OperationResult<()>
+    where
+        TKey: num_traits::cast::AsPrimitive<usize>,
+    {
+        self.mmap_storage.insert_many(start_key, vectors, count)?;
```
IvanPleshkov, line 93: I propose to write RAM first. The same reason as for `insert`

IvanPleshkov (approved):
approve, added some nits

lib/segment/src/vector_storage/chunked_utils.rs
```diff
@@ -48,6 +48,16 @@ pub fn read_mmaps<T: Sized>(directory: &Path) -> OperationResult<Vec<MmapChunk<T
...
+        // If unix, lock the memory
+        #[cfg(unix)]
+        if mlock {
+            mmap.lock()?;
+        }
+        // If not, log warning and continue
+        #[cfg(not(unix))]
+        if mlock {
+            log::warn!("Can't lock vectors in RAM, is not supported on this platform");
+        }
```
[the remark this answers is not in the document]
generall: I can tolerate problems on windows, but large mlock is problematic on unix as well

lib/segment/src/vector_storage/chunked_vectors.rs
generall: But we don't allow individual vectors to be larger than chunk. If we take validation into account, vectors are guaranteed to be smaller than chunk size

lib/segment/src/vector_storage/in_ram_persisted_vectors.rs
generall: Theoretically yes, but it will break abstraction

lib/segment/src/vector_storage/chunked_vectors.rs
IvanPleshkov: Do we have DIM limit? I'm sure we have limit for multivectors. Also, we can add assert at least
```
assert_ne!(chunk_capacity, 0);
```

generall: > Do we have DIM limit?

yes, 65k

lib/segment/src/vector_storage/chunked_utils.rs
```diff
@@ -48,6 +48,16 @@ pub fn read_mmaps<T: Sized>(directory: &Path) -> OperationResult<Vec<MmapChunk<T
             ))
         })?;
         let mmap = open_write_mmap(&mmap_file)?;
+        // If unix, lock the memory
+        #[cfg(unix)]
+        if mlock {
+            mmap.lock()?;
```
timvisee, line 54: I believe I've seen this fail in some cases, for example when low in memory. I therefore think we should return a warning here rather than failing.

PR #4838, opened 2024-08-05
refactor: Abstract away automatic resharding filter on reads

coszio:
Introduces a `ReshardableReadRequest` type to use on every read request which implement a new `EditFilter` trait. This reduces the cognitive complexity of internal read apis like:

- batch query
- scroll
- count
- batch search

And prepares for easy integration with facets, which will use this as soon as this is merged

### DIFF
--- a/lib/collection/src/shards/shard_holder/reshardable_read_request.rs
+++ b/lib/collection/src/shards/shard_holder/reshardable_read_request.rs
@@ -7,103 +7,119 @@ use crate::operations::types::{CoreSearchRequest, CoreSearchRequestBatch, CountR
 use crate::operations::universal_query::shard_query::ShardQueryRequest;
 use crate::shards::shard::ShardId;
 
-pub trait EditFilter {
-    fn edit_filter<E>(&mut self, edit: E)
+pub enum ReshardableReadRequest<T> {
+    Filtered {
+        resharding_shard_id: ShardId,
+        filtered: T,
+        original: T,
+    },
+
+    Normal {
+        request: T,
+    },
+}
+
+impl<T> ReshardableReadRequest<T> {
+    pub fn new(resharding_shard_id: ShardId, resharding_filter: Filter, request: T) -> Self
     where
-        E: Clone + FnMut(Option<Filter>) -> Option<Filter>;
-}
+        T: Clone + MergeFilter<Filter>,
+    {
+        let mut filtered = request.clone();
+        filtered.merge_filter(resharding_filter);
 
-// Preserves original request, and optionally holds an edited one with the resharding filter.
-//
-// The edited request should be used on all shards, except the new resharding shard
-pub struct ReshardableReadRequest<T> {
-    /// Original request
-    pub original: Arc<T>,
-
-    /// Tuple of edited request and resharding shard id
-    pub edited: Option<(Arc<T>, u32)>,
-}
-
-impl<T: EditFilter + Clone> ReshardableReadRequest<T> {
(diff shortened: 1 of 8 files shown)


agourlay:
I am in favor of adding this abstraction as the resharding filters are creeping in many places.

Not in position to decide whether this is good for the resharding work group :)

ffuugoo (approved):
Yoooo, this is good! @coszio is better at programming than @timvisee and me. 🥲

I think I'd want to merge this, cause it will simplify the next ~~hack~~ feature I'll be working on for resharding.

The only thing I'm not sure about is naming, but I'll try to come up with something. Also, happy to rebase this on `dev`, cause I'm gonna need this anyway.

coszio:
Thanks @ffuugoo, definitely a better name is possible. I'll fix conflicts and merge

lib/collection/src/shards/shard_holder/reshardable_read_request.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /// If available, use edited request on all shards, except the new resharding shard
+    pub fn get_request(&self, shard_id: ShardId) -> Arc<T> {
+        if let Some((edited, resharding_id)) = self.edited.clone() {
+            // Use edited request on all shards, except the new resharding shard
+            if resharding_id != shard_id {
+                return edited;
+            }
+        }
```
timvisee, line 58: We can probably eliminate this clone, by checking the shard ID as a reference first.

timvisee (approved):
Looks great indeed! Definitely a lot better than what we did before.

It may be worth to test this implementation by temporarily enabling the resharding tests we had, and the resharding driver, just to make sure this implementation is sound.

ffuugoo:
Took me a while to figure out how to structure this properly... 🙄

lib/collection/src/shards/shard_holder/reshardable_read_request.rs
ffuugoo: Done as part of my changes.

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo, line 1: Why though? Does not make sense to me, especially when use as `From::from`.

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -374,11 +374,11 @@ impl ShardHolder {
         T: Clone + MergeFilter,
     {
         let Some(state) = self.resharding_state() else {
-            return request.into();
+            return From::from(request);
```
ffuugoo, line 377: Why though? Does not make sense to me. Especially as `From::from` (instead of, e.g., `ReshardableRequest::from`).

ffuugoo: `Into` exists to make your code less annoying, use it!

coszio: Except that it has terrible navigation

ffuugoo:
Though, sad to tell that we gonna nuke this. Filters turned out to be way too slow. 💀

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo: Is rust-analyzer clever enough to jump to correct implementation of `From` when using `From::from`? 🤔

timvisee:
I'll draft this to not accidentally merge it until we're sure we actually drop it.

Thanks for your effort!

timvisee:
It looks like we won't be going with this approach because we'll be filtering updates instead.

PR #4842, opened 2024-08-06
Add and fix missing timeout in the retrieve APIs

agourlay:
This PR adds and fixes missing timeout in the retrieve APIs such as:
- ignoring the timeout provided by the user (REST)
- not enabling the user the provide a timeout (gRPC)

This has been done for:
- scroll
- count
- get_points

The fixes follow the same strategy:
- cap async computations with `tokio::time::timeout`
- propagate timeout to other cluster node
- apply timeout to inter node gRPC connections
- update timeout when several operations are performed sequentially

All shard transfer operations relying on `scrolls` are now using a `None` timeout to keep the previous behavior.

### DIFF
--- a/lib/collection/src/collection/search.rs
+++ b/lib/collection/src/collection/search.rs
@@ -45,6 +45,7 @@ impl Collection {
         shard_selection: ShardSelectorInternal,
         timeout: Option<Duration>,
     ) -> CollectionResult<Vec<Vec<ScoredPoint>>> {
+        let start = Instant::now();
         // shortcuts batch if all requests with limit=0
         if request.searches.iter().all(|s| s.limit == 0) {
             return Ok(vec![]);
@@ -100,6 +101,8 @@ impl Collection {
                     timeout,
                 )
                 .await?;
+            // update timeout
+            let timeout = timeout.map(|t| t - start.elapsed());
             let filled_results = without_payload_results
                 .into_iter()
                 .zip(request.clone().searches.into_iter())


lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -169,7 +172,11 @@ impl ShardOperation for LocalShard {
...
     }
 
-    async fn count(&self, request: Arc<CountRequestInternal>) -> CollectionResult<CountResult> {
+    async fn count(
+        &self,
+        request: Arc<CountRequestInternal>,
+        _timeout: Option<Duration>,
+    ) -> CollectionResult<CountResult> {
         let total_count = if request.exact {
             let all_points = self.read_filtered(request.filter.as_ref())?;
```
coszio, line 181: I think it makes sense to enable cancellation of `read_filtered`

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -184,6 +191,7 @@ impl ShardOperation for LocalShard {
         request: Arc<PointRequestInternal>,
         with_payload: &WithPayload,
         with_vector: &WithVector,
+        _timeout: Option<Duration>,
```
coszio, line 194: Maybe also here

lib/collection/src/shards/transfer/stream_records.rs
```diff
@@ -48,10 +50,13 @@ pub(super) async fn transfer_stream_records(
...
-            .count_local(Arc::new(CountRequestInternal {
-                filter: None,
-                exact: true,
-            }))
+            .count_local(
+                Arc::new(CountRequestInternal {
+                    filter: None,
+                    exact: true,
+                }),
+                timeout,
```
coszio, line 58: I think we can cut the propagation here, no need to specify timeout for transfer_stream_records, I believe

lib/collection/src/shards/transfer/resharding_stream_records.rs
```diff
@@ -95,7 +98,7 @@ pub(crate) async fn transfer_resharding_stream_records(
         };
 
         offset = replica_set
-            .transfer_batch(offset, TRANSFER_BATCH_SIZE, Some(&hashring), true)
+            .transfer_batch(offset, TRANSFER_BATCH_SIZE, Some(&hashring), true, None)
```
coszio, line 101: Also here, we might not need to know about timeouts for this function

lib/collection/src/collection/search.rs
```diff
@@ -110,6 +110,7 @@ impl Collection {
                         req.with_vector.unwrap_or_default(),
                         read_consistency,
                         &shard_selection,
+                        timeout,
```
coszio, line 113: We can update the timeout here as well :)

src/common/auth/mod.rs
```diff
@@ -126,6 +126,7 @@ impl AuthKeys {
                 value_exists.get_collection(),
                 scroll_req,
                 None,
+                None, // no timeout
```
coszio, line 129: I'm explicitly ignoring this part too :D

coszio (approved):

lib/collection/src/shards/local_shard/shard_ops.rs
agourlay: It actually requires a bit more work to do it properly, I propose to do this in a subsequent PR.

lib/collection/src/shards/local_shard/shard_ops.rs
agourlay: Same proposal

lib/collection/src/shards/local_shard/shard_ops.rs
agourlay: https://github.com/qdrant/qdrant/pull/4842/commits/81b42890a14d5849f7c137b41a59f47de10f9c0d

lib/collection/src/shards/local_shard/shard_ops.rs
agourlay: https://github.com/qdrant/qdrant/pull/4842/commits/81b42890a14d5849f7c137b41a59f47de10f9c0d

lib/collection/src/shards/local_shard/shard_ops.rs
coszio: Ok

PR #4843, opened 2024-08-06
Add local shard points get/scroll/count/delete HTTP API

ffuugoo:
`delete` API is required to move resharding driver to external service. `get`/`scroll`/`count` are not strictly required, but seems useful for future debugging and troubleshouting.

__TODO:__
- disable automatic resharding filter for local shard points APIs
	- currently, after read hash ring was committed, resharding filter is automatically applied to all read requests
	- e.g., `get`/`scroll`/`count` start return empty/zero results...
	- and `delete` stops working, because it relies on `scroll` requests internally
	- it is still possible to test the API (just don't commit read hash ring), so this fix will be done in a follow-up PR
- integration tests
	- I'd prefer to add tests in a follow-up PR as well, once everything is fixed and fully ready

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/hash_ring.rs
+++ b/lib/collection/src/hash_ring.rs
@@ -10,7 +10,7 @@ use smallvec::SmallVec;
 use crate::operations::cluster_ops::ReshardingDirection;
 use crate::shards::shard::ShardId;
 
-const HASH_RING_SHARD_SCALE: u32 = 100;
+pub const HASH_RING_SHARD_SCALE: u32 = 100;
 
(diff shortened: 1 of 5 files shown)


ffuugoo:
@generall @timvisee Alternative approach to `hash_ring_filter` for "local shard points" API. Instead of specifying hashring explicitly in the request, we only specify `expected_shard_id` and use *current* hashring for the filter.

timvisee (approved):

PR #4844, opened 2024-08-07
Non blocking retrieve with timeout and cancellation support

agourlay:
Performing a retrieve operation at the segment level is a blocking operation:
- the segment's read lock needs to be acquired
- the vectors are fetched from storage
- the payload are fetched from storage

This PR makes retrieving asynchronous & multithreaded so it does not block anymore Tokio threads which can cause various responsiveness issues.

It also applies additional measures to:
- honor user timeout where possible
- ensure cancellation of tasks

### DIFF
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -1,11 +1,11 @@
 use std::collections::HashMap;
+use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 
 use common::types::ScoreType;
 use futures::future::try_join_all;
 use itertools::Itertools;
 use ordered_float::Float;
-use parking_lot::RwLock;
 use segment::common::operation_error::OperationError;
 use segment::data_types::named_vectors::NamedVectors;
 use segment::data_types::query_context::QueryContext;
@@ -19,7 +19,7 @@ use tokio::runtime::Handle;
 use tokio::task::JoinHandle;
 
 use super::holders::segment_holder::LockedSegmentHolder;
-use crate::collection_manager::holders::segment_holder::{LockedSegment, SegmentHolder};
+use crate::collection_manager::holders::segment_holder::LockedSegment;
 use crate::collection_manager::probabilistic_segment_search_sampling::find_search_sampling_over_point_distribution;
 use crate::collection_manager::search_result_aggregator::BatchResultAggregator;
 use crate::common::stopping_guard::StoppingGuard;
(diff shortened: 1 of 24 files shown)


lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -192,11 +193,22 @@ impl ShardOperation for LocalShard {
...
         with_vector: &WithVector,
-        _timeout: Option<Duration>,
+        search_runtime_handle: &Handle,
+        timeout: Option<Duration>,
     ) -> CollectionResult<Vec<Record>> {
-        // TODO run in spawn_blocking + timeout + cancellation
-        let records_map =
-            SegmentsSearcher::retrieve(self.segments(), &request.ids, with_payload, with_vector)?;
+        let timeout = timeout.unwrap_or(self.shared_storage_config.search_timeout);
+        let records_map = tokio::time::timeout(
```
agourlay, line 200: where the rubber meets the road

lib/collection/src/collection_manager/segments_searcher.rs
```diff
@@ -342,58 +342,90 @@ impl SegmentsSearcher {
...
+    pub async fn retrieve(
+        segments: LockedSegmentHolder,
+        points: &[PointIdType],
+        with_payload: &WithPayload,
+        with_vector: &WithVector,
+        runtime_handle: &Handle,
+    ) -> CollectionResult<HashMap<PointIdType, Record>> {
+        let stopping_guard = StoppingGuard::new();
+        runtime_handle
+            .spawn_blocking({
```
timvisee, line 354: For reviewers, important part: here we now spawn blocking tasks for each segment.

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -338,7 +367,7 @@ impl LocalShard {
             ))
         })?;
 
-        let mut rng = rand::thread_rng();
+        let mut rng = StdRng::from_entropy();
```
timvisee, line 370: What's the reason for switching from `rand::thread_rng` to `StdRng::from_entropy` here?

I'd expect the previous one to be faster because of reusing seeded material (though I haven't tested this).

timvisee (approved):

lib/collection/src/collection_manager/segments_searcher.rs
agourlay: Having one task per segment is possible optimization (marked by a TODO below)

For now there is one task per `segment_holder`, so just one per request :)

lib/collection/src/shards/local_shard/scroll.rs
agourlay: Thanks for pointing this, I forgot to add a comment.

The `rand::thread_rng` is not `Send` and got me into compilation issues.

timvisee: Makes sense, thanks!

timvisee: I think we can also just put `rand::thread_rng()` in the place we use `rng`

agourlay: So you mean calling `rand::thread_rng()` at every loop iteration?

Do you believe it will be faster than a single `StdRng::from_entropy` because of memoization or something?

No strong opinion on my side :)

timvisee: Based on a quick skim I thought the number of iterations would be very small, but that may not be the case.

Let's keep it like this. Sorry for the noise.

agourlay: No worries, it is always better to be thorough!

PR #4849, opened 2024-08-08
Non blocking exact count with timeout and cancellation support

agourlay:
Similar to https://github.com/qdrant/qdrant/pull/4844 for exact counts.

Exact counts can be very expensive, this PR make those non-blocking with the regular benefits on the Tokio runtime.

### DIFF
--- a/lib/collection/src/shards/local_shard/shard_ops.rs
+++ b/lib/collection/src/shards/local_shard/shard_ops.rs
@@ -176,11 +176,19 @@ impl ShardOperation for LocalShard {
     async fn count(
         &self,
         request: Arc<CountRequestInternal>,
-        _timeout: Option<Duration>,
+        search_runtime_handle: &Handle,
+        timeout: Option<Duration>,
     ) -> CollectionResult<CountResult> {
-        // TODO run in spawn_blocking + timeout + cancellation
         let total_count = if request.exact {
-            let all_points = self.read_filtered(request.filter.as_ref())?;
+            let timeout = timeout.unwrap_or(self.shared_storage_config.search_timeout);
+            let all_points = tokio::time::timeout(
+                timeout,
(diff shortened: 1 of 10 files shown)


lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -176,11 +176,19 @@ impl ShardOperation for LocalShard {
...
-        // TODO run in spawn_blocking + timeout + cancellation
         let total_count = if request.exact {
-            let all_points = self.read_filtered(request.filter.as_ref())?;
+            let timeout = timeout.unwrap_or(self.shared_storage_config.search_timeout);
+            let all_points = tokio::time::timeout(
+                timeout,
+                self.read_filtered(request.filter.as_ref(), search_runtime_handle),
+            )
+            .await
+            .map_err(|_: Elapsed| {
```
timvisee, line 189:
```suggestion
            .map_err(|_| {
```

timvisee (approved):

lib/collection/src/shards/local_shard/shard_ops.rs
agourlay: It seems we ascribed the `Elapsed` type on all `tokio::time::timeout` usages in the code base.
Maybe an artifact of previous type inference issues.
I will remove all of those in a single PR right after this one.

agourlay: Maybe we did that to avoid issues during refactoring in order to make sure the ignored value has this very type.

coszio: For me it didn't work without specifying it for some reason

timvisee: It did build for me without the specification. Weird!

PR #4864, opened 2024-08-12
Limit JsonPath::filename() to 33 characters

xzfc:
This PR limits JsonPath::filename() to from 64 to 33 characters (24 for hash + 1 for underscore + 8 for human-readable field name).

Motivation: make it less like likely to hit 260 characters path length limit for Windows. This limitation should not affect modern Rust code, but it might affect legacy apps peeking into Qdrant storage directory (e.g. archivers).


timvisee (approved):
Since it is shorter now. Shall we also add a `.dat` suffix, I believe we do that for some similar files?

xzfc:
It's used as a directory name, not regular file path.

PR #4867, opened 2024-08-12
Filter update operations during resharding

ffuugoo:
This PR adds a pre-filter to update operations during resharding.

There are two different (but similar in implementation) filters introduced:

During `MigratingPoints` stage we forward relevant update operations to "new" shard. However, what if we forward, e.g., `SetPayload` operation for the point that was not yet migrated to the new shard? In such case, we can pre-filter update operation and remove modifications to points that does not (yet) exist in the shard.

Starting at `CommitReadHashRing` stage we want to apply pre-filter to update operations on "old" shards. We want to apply updates to points that are present both in old and new shards, but we don't want to *insert* new (or previously deleted) points that are hashed into new shard. This allows us to delete resharded points from old shards, but also keep the data consistent between old and new shards while we do these deletions.

**TODO:**
- [x] think through/handle resharding down?
- [ ] tests

### All Submissions:

### New Feature Submissions:

3. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
4. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

### DIFF
--- a/lib/api/src/rest/schema.rs
+++ b/lib/api/src/rest/schema.rs
@@ -80,19 +80,19 @@ pub enum BatchVectorStruct {
 }
 
 impl BatchVectorStruct {
-    pub fn swap_remove(&mut self, vector_index: usize) {
+    pub fn remove(&mut self, index: usize) {
         match self {
             Self::Single(vectors) => {
-                vectors.swap_remove(vector_index);
+                vectors.remove(index);
             }
 
             Self::MultiDense(vectors) => {
-                vectors.swap_remove(vector_index);
+                vectors.remove(index);
             }
 
             Self::Named(vector_batches) => {
                 for vector_batch in vector_batches.values_mut() {
-                    vector_batch.swap_remove(vector_index);
+                    vector_batch.remove(index);
                 }
             }
         }
--- a/lib/api/src/rest/schema.rs
+++ b/lib/api/src/rest/schema.rs
@@ -79,6 +79,26 @@ pub enum BatchVectorStruct {
     Named(HashMap<String, Vec<Vector>>),
 }
 
+impl BatchVectorStruct {
+    pub fn swap_remove(&mut self, vector_index: usize) {
+        match self {
+            Self::Single(vectors) => {
+                vectors.swap_remove(vector_index);
+            }
+
+            Self::MultiDense(vectors) => {
+                vectors.swap_remove(vector_index);
+            }
+
+            Self::Named(vector_batches) => {
+                for vector_batch in vector_batches.values_mut() {
+                    vector_batch.swap_remove(vector_index);
+                }
+            }
+        }
+    }
+}
+
 #[derive(Debug, Deserialize, Serialize, Clone, JsonSchema, PartialEq)]
 #[serde(untagged)]
 pub enum ShardKeySelector {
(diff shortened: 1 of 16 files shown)


ffuugoo:
The name `ReshardingUpdatePreFilter` is intentionally hideous, cause I can't come up with a good enough one. Please help! 🥲🙏

@generall @timvisee

lib/collection/src/operations/point_ops.rs
```diff
@@ -216,6 +216,34 @@ pub enum PointInsertOperationsInternal {
...
+        }
+    }
+
+    pub fn remove_point_ids(&mut self, point_ids: &HashSet<ExtendedPointId>) {
+        match self {
+            Self::PointsBatch(batch) => {
+                for index in (0..point_ids.len()).rev() {
+                    if point_ids.contains(&batch.ids[index]) {
+                        batch.ids.swap_remove(index);
+                        batch.vectors.swap_remove(index);
```
timvisee, line 233: Operations must remain ordered, in case we update the same point ID twice in a single batch.

lib/api/src/rest/schema.rs
```diff
@@ -79,6 +79,26 @@ pub enum BatchVectorStruct {
...
+                vectors.swap_remove(vector_index);
+            }
+
+            Self::Named(vector_batches) => {
+                for vector_batch in vector_batches.values_mut() {
+                    vector_batch.swap_remove(vector_index);
+                }
+            }
+        }
+    }
```
timvisee, line 99: Same here as mentioned above. I don't think we can do this.

We can make this a bit smarter, to use `swap_remove` only if the same ID doesn't follow. But I'm not sure if that's worth it.

lib/collection/src/collection/resharding_update_pre_filter.rs
```diff
@@ -0,0 +1,96 @@
...
+
+impl ReshardingUpdatePreFilter {
+    pub fn new(shard_holder: &ShardHolder, shard_id: ShardId) -> Option<Self> {
+        let state = shard_holder.resharding_state()?;
+
+        if state.shard_id == shard_id && state.stage == ReshardStage::MigratingPoints {
+            Some(Self::NewShard)
+        } else if state.shard_id != shard_id && state.stage >= ReshardStage::ReadHashRingCommitted {
+            shard_holder.resharding_filter().map(Self::OldShard)
+        } else {
```
timvisee, line 27: These should be inverted if resharding down

lib/collection/src/collection/resharding_update_pre_filter.rs
```diff
@@ -0,0 +1,96 @@
...
+        operation: &mut CollectionUpdateOperations,
+        local_shard: &(dyn ShardOperation + Sync),
+        runtime: &tokio::runtime::Handle,
+    ) -> CollectionResult<()> {
+        // There's no need to pre-filter point delete operations
+        if operation.is_delete_points() {
+            return Ok(());
+        }
+
+        let points_to_filter = match self {
```
timvisee, line 43: Points that are not moved as part of resharding should not be affected at all.

In other words, we should only check the condition on point IDs here that are distributed to another shard according to the hash rings. All other point IDs should be kept no matter what. I believe we had a `is_point_moved` (or something) helper for that.

ffuugoo: > Points that are not moved as part of resharding should not be affected at all.

That's exactly how it works already? Can you clarify, cause I can't see what's wrong currently.

ffuugoo: I think I got it. It matters when pre-filtering on receiver shard during resharding down. 🤦‍♀️

ffuugoo:
Closing, cause we implemented a different approach in #4928.

PR #4881, opened 2024-08-13
InRamChunkedMmap: optimize using madvise

xzfc:
Follow-up for #4828.

### DIFF
--- a/lib/segment/src/vector_storage/chunked_mmap_vectors.rs
+++ b/lib/segment/src/vector_storage/chunked_mmap_vectors.rs
@@ -1,6 +1,7 @@
 use std::cmp::max;
 use std::fs::{create_dir_all, OpenOptions};
 use std::io::Write;
+use std::iter::zip;
 use std::path::{Path, PathBuf};
 
 use memmap2::MmapMut;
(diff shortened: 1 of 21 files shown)


lib/segment/src/vector_storage/chunked_mmap_vectors.rs
```diff
@@ -221,6 +225,42 @@ impl<T: Sized + Copy + 'static> ChunkedMmapVectors<T> {
...
+        let mut it = 0..self.len();
+
+        for (chunk_idx, chunk) in self.chunks.iter().enumerate() {
+            for (i, _) in zip(0..self.config.chunk_size_vectors, &mut it) {
+                let idx = i * self.config.dim;
+                f(&chunk[idx..idx + self.config.dim])?;
+            }
+
+            chunk.flusher()()?;
+            fn _assert_t_is_copy<T: Copy>() {}
```
timvisee, line 242: What is this line for?

lib/sparse/src/index/inverted_index/inverted_index_mmap.rs
```diff
@@ -196,8 +195,7 @@ impl InvertedIndexMmap {
         let file_header: InvertedIndexFileHeader = read_json(&config_file_path)?;
         // read index data into mmap
         let file_path = Self::index_file_path(path.as_ref());
-        let mmap = open_read_mmap(file_path.as_ref())?;
-        madvise::madvise(&mmap, madvise::Advice::Normal)?;
+        let mmap = open_read_mmap(file_path.as_ref(), Advice::Normal.into())?;
```
generall, line 198: For the sake of better code navigation, let's use something like this:

```suggestion
        let mmap = open_read_mmap(file_path.as_ref(), AdviceSetting::from(Advice::Normal))?;
```

generall (approved):
LGTM, with minor nitpick.

lib/segment/src/vector_storage/chunked_mmap_vectors.rs
timvisee: I see now. Interesting hack to enforce the `Copy` trait :smile:

timvisee (approved):

PR #4883, opened 2024-08-13
Fix openapi operationId inconsistency

agourlay:
The proper naming scheme is `*_point_groups`.

Do not merge before 1.12.0.

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -5464,7 +5464,7 @@
         ],
         "summary": "Query points, grouped by a given payload field",
         "description": "Universally query points, grouped by a given payload field",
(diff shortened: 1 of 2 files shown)


timvisee (commented):
I'm afraid we have to keep both, and deprecate the wrong naming. That'll allow us to have a migration period of a few versions.

Or are you suggesting to remove it from the spec here, but keep the old endpoint in code?

agourlay:
AFAIK this `operationId` is only used for:
- the API reference docs
- the Python REST client

In the next minor version we could deprecate the Python function `query_points_groups` and add `query_point_groups`

agourlay:
This fell off my radar and I am unsure of the best approach.

PR #4884, opened 2024-08-13
Search distance matrix REST API

agourlay:
**context**: https://qdrant.notion.site/shared-Distance-Matrix-API-07cc5decbbb6435da8a87b18cf5c1305
**on top of**: https://github.com/qdrant/qdrant/pull/4793

This PR adds a new way to query points by enabling users to retrieve the distance matrix of sampled points.

There are different ways to represent a distance sparse matrix therefore I implemented 3 different endpoints with different output shapes to target different use cases.

The goal of this PR is to decide which ones make sense for ours users.

I called those output shapes:
- offsets based
- rows based
- pairs based


lib/collection/src/collection/distance_matrix.rs
```diff
@@ -12,26 +16,140 @@ use crate::operations::types::{CollectionResult, CoreSearchRequest, CoreSearchRe
 use crate::operations::universal_query::shard_query::{Sample, ScoringQuery, ShardQueryRequest};
 
 #[derive(Debug, Default)]
-pub struct DistanceMatrixResponse {
-    pub sample_ids: Vec<PointIdType>,   // sampled point ids
-    pub nearest: Vec<Vec<ScoredPoint>>, // nearest points for each sampled point
+pub struct CollectionSearchMatrixResponse {
+    pub sample_ids: Vec<PointIdType>,    // sampled point ids
+    pub nearests: Vec<Vec<ScoredPoint>>, // nearest points for each sampled point
```
coszio, line 21: If order of ids is not important as an internal representation, this can be represented as a HashMap. It would make it easier to understand, IMO

lib/api/src/rest/schema.rs
```diff
@@ -651,3 +653,70 @@ pub struct QueryGroupsRequest {
...
 }
+
+#[derive(Serialize, Deserialize, JsonSchema, Validate, Debug, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub struct SearchMatrixRequestInternal {
+    /// Look only for points which satisfies this conditions
+    #[validate]
+    pub filter: Option<Filter>,
+    /// How many points to select and search within.
+    #[validate(range(min = 1))]
```
coszio, line 664: I assume a sane minimum sample size would be 2, since 1 won't have another vector to compare against

coszio (commented):
While having multiple output formats is helpful for wiring out easily to the consumers, I believe that having 3 separate endpoints is a bit overkill. 

I would like to justify each format with a known consumer of it

generall:
> I would like to justify each format with a known consumer of it

Agree. Here is my reasoning:

* offsets based - is the native format for `csr_matrix` in scipy: https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html So majority of the popular python-based libraries will work with this format

* pairs based - I refactored into a bit more human-readable format, it not avoids any offsets and just assigns a score to each known pair in a flatten manner. This is also convenient for graph rendering, where you can just iterate over pairs and do `graph.add_link(a,b, score)`

* rows based - I think is the most confusing one, I propose to drop it.

generall (approved):

PR #4887, opened 2024-08-14
[Strict-Mode] Basic implementation

JojiiOfficial:
Introduces a new request verification trait, for checking  collection-level requests (eg. for strict mode).
This trait is designed to be simply extendable for other checks we might want to add later. 

The newly introduced `VerificationPass` will later be passed to the `.toc()` function of `Dispatcher` like `access`, so we don't forget to implement for new types.


coszio:
Is it something separate from `Access`? or will it unify with it? I think the latter would be cleaner

JojiiOfficial:
It is separate from `Access`. I considered unifying it with Access but I think they both try to achieve a different goal and It would be cleaner to have them separated

lib/collection/src/operations/verification/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+            self.request_timeout(),
+            strict_mode_config.max_timeout,
+            "timeout",
+        )
+    }
+
+    // Checks all filters use indexed fields only.
+    fn check_request_filter(
+        &self,
+        collection: &RwLockReadGuard<'_, Collection>,
```
generall, line 85: I don't think we need `RwLockReadGuard` as argument type

lib/collection/src/operations/config_diff.rs
```diff
@@ -85,6 +86,68 @@ pub struct HnswConfigDiff {
...
+    pub search_max_oversampling: Option<f64>,
+
+    // Group-by
+    pub group_by_max_group_size: Option<usize>,
+
+    // Recommend
+    pub recommend_max_examples: Option<usize>,
+
+    // Discovery
+    pub discovery_max_context_size: Option<usize>,
```
generall, line 110: Let's skip serialization for None for all fields

lib/collection/src/operations/config_diff.rs
```diff
@@ -85,6 +86,68 @@ pub struct HnswConfigDiff {
     pub payload_m: Option<usize>,
 }
 
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Validate, Clone, Merge)]
+pub struct StrictModeConfigDiff {
+    // Global
+    pub enabled: Option<bool>,
```
generall, line 92: _Maybe_ make enabled non-optional

lib/collection/src/operations/conversions.rs
```diff
@@ -760,6 +776,7 @@ impl TryFrom<api::grpc::qdrant::CollectionConfig> for CollectionConfig {
                     None
                 }
             },
+            strict_mode_config: config.strict_mode_config.map(|i| i.into()),
```
generall, line 779: Instead of .into, let's use Struct::from, for better code navigation

lib/storage/src/dispatcher.rs
```diff
@@ -38,10 +39,18 @@ impl Dispatcher {
...
     /// of the [Access] object.
+    /// TODO: Replace this function with `toc_new` everywhere.
     pub fn toc(&self, _access: &Access) -> &Arc<TableOfContent> {
         &self.toc
     }
 
+    /// Get the table of content.
+    /// The `_access` and `_pass` parameters are not used, but are required to verify caller's possession
+    /// of the [Access] and [VerificationPass] object.
+    pub fn toc_new(&self, _access: &Access, _pass: &VerificationPass) -> &Arc<TableOfContent> {
```
generall, line 50: Maybe store `Access` in `VerificationPass` and avoid extra function

lib/storage/src/content_manager/collection_verification.rs
```diff
@@ -0,0 +1,80 @@
...
+        &self,
+        collection: &RwLockReadGuard<'_, Collection>,
+        strict_mode_config: &StrictModeConfigDiff,
+    ) -> Result<(), StorageError>;
+}
+
+/// Returns the `TableOfContents` from `dispatcher` without needing a validity check.
+/// Caution: Do only use this to obtain a `VerificationPass`!
+/// Don't make public!
+fn get_toc_unchecked<'a>(dispatcher: &'a Dispatcher, access: &Access) -> &'a Arc<TableOfContent> {
```
generall, line 62: rename to "without_verification" to avoid confusion with Access token validation

lib/collection/src/operations/verification/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+use crate::collection::Collection;
+
+// Creates a new `VerificationPass` for successful verifications.
+// Don't use this, unless you know what you're doing!
+pub fn new_pass() -> VerificationPass {
+    VerificationPass { inner: () }
+}
+
+/// A pass, created on successful verification.
+pub struct VerificationPass {
```
generall, line 18: Let's rollback verification pass for now, and keep config and `StrictModeVerification` trait with implementation.

Changing Access logic might be non-trivial enough to make a separate discussion about it

lib/collection/src/collection/payload_index_schema.rs
```diff
@@ -84,4 +84,37 @@ impl Collection {
...
+            .must
+            .iter()
+            .flatten()
+            .chain(filter.must_not.iter().flatten())
+            .chain(filter.should.iter().flatten())
+            .chain(filter.min_should.iter().flat_map(|i| &i.conditions))
+            .find_map(|condition| self.has_condition_without_index(condition))
+    }
+
+    fn has_condition_without_index(&self, condition: &Condition) -> Option<JsonPath> {
```
generall, line 101: This function doesn't see to take index type into account.

E.g. if index is built for integer and query uses keyword

lib/collection/src/collection/payload_index_schema.rs
```diff
@@ -84,4 +84,37 @@ impl Collection {
...
+
+        match condition {
+            Condition::Field(field) => {
+                if !has_key(&field.key) {
+                    Some(field.key.clone())
+                } else {
+                    None
+                }
+            }
+            Condition::Nested(nested) => self.has_filter_without_index(nested.filter()),
```
generall, line 112: It would be nice to see extra tests for nested fields, as they use relative keys

lib/collection/src/collection/payload_index_schema.rs
```diff
@@ -84,4 +84,37 @@ impl Collection {
...
+            Condition::Filter(filter) => self.has_filter_without_index(filter),
+            Condition::IsEmpty(_)
+            | Condition::IsNull(_)
```
generall, line 115: Is Empty and IsNull should be indexed as wll

generall (changes requested):
I think it should be possible to share functions for checking payload index with issue API. It seems like there are a lot of overlap there and it might reduce code duplication.

See https://github.com/qdrant/qdrant/blob/644b01ee89a1f53e175da63bd2f3adbaed31e06a/lib/segment/src/problems/unindexed_field.rs#L273

lib/segment/src/problems/unindexed_field.rs
```diff
@@ -236,14 +236,14 @@ fn infer_schema_from_field_condition(field_condition: &FieldCondition) -> Vec<Pa
     inferred
 }
 
-struct Extractor<'a> {
+pub struct Extractor<'a> {
     payload_schema: &'a HashMap<PayloadKeyType, PayloadFieldSchema>,
```
JojiiOfficial, line 240: I'm not completely happy with the way we're reusing `Extractor` here but I couldn't come up with a better idea.
If you have a suggestion on how to improve this, please let me know!

The advantages are, we're automatically testing `Extractor` too and don't have any code duplication.

lib/collection/src/operations/config_diff.rs
```diff
@@ -85,6 +86,79 @@ pub struct HnswConfigDiff {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub search_max_oversampling: Option<f64>,
+
+    // Group-by
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub group_by_max_group_size: Option<usize>,
+
+    // Recommend
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub recommend_max_examples: Option<usize>,
```
coszio, line 117: Recommend for average vector with multiple examples is not too expensive, does this apply only for `best_score`? If it is the case, name should be something like: `recommend_best_score_max_inputs`

lib/collection/src/operations/config_diff.rs
```diff
@@ -85,6 +86,79 @@ pub struct HnswConfigDiff {
     pub payload_m: Option<usize>,
 }
 
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Validate, Clone, Merge)]
+pub struct StrictModeConfig {
```
coszio, line 90: Let's add some descriptions to each field

lib/collection/src/operations/verification/mod.rs
```diff
@@ -0,0 +1,160 @@
+mod search;
+
+use std::fmt::Display;
+
+use segment::types::Filter;
+
+use super::config_diff::StrictModeConfig;
+use crate::collection::Collection;
+
+/// Trait to verify strict mode for requests. All functions in this trait will default to an 'empty' implementation,
```
coszio, line 10: I don't like this idea too much. It makes it more ergonomic to implement, but it may shoot us in the foot when maintaining it. 

IMO it is better to explicitly ignore irrelevant methods for each implementor and let the compiler complain if something isn't considered.

lib/collection/src/operations/verification/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+    ))
+}
+
+pub(crate) fn check_limit_opt<T: PartialOrd + Display>(
+    value: Option<T>,
+    limit: Option<T>,
+    name: &str,
+) -> Result<(), String> {
+    let Some(limit) = limit else { return Ok(()) };
+    let Some(value) = value else { return Ok(()) };
```
coszio, line 151:
```suggestion
    let (Some(limit), Some(value)) = (limit, value) else { return Ok(()) };
```

lib/collection/src/operations/verification/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+/// Trait to verify strict mode for requests. All functions in this trait will default to an 'empty' implementation,
+/// which means every value that needs to be checked must be implemented.
+/// This trait ignores the `enabled` parameter in `StrictModeConfig`.
+pub trait StrictModeVerification {
+    /// Implementing this method allows adding a custom check for request specific values.
+    fn check_custom(
+        &self,
+        _collection: &Collection,
+        _strict_mode_config: &StrictModeConfig,
+    ) -> Result<(), String> {
```
coszio, line 19: I think we could either add variants in `CollectionError` or create a new kind of error implement conversion into `CollectionError`

lib/segment/src/problems/unindexed_field.rs
coszio: #4945

lib/segment/src/types.rs
```diff
@@ -2340,6 +2340,15 @@ impl Filter {
             (Some(this), Some(other)) => Some(this.merge_owned(other)),
         }
     }
+
+    pub fn iter_conditions(&self) -> impl Iterator<Item = &Condition> {
```
coszio, line 2344: nice improvement, thanks

lib/storage/src/content_manager/conversions.rs
```diff
@@ -62,6 +62,7 @@ impl TryFrom<api::grpc::qdrant::CreateCollection> for CollectionMetaOperations {
                     .sharding_method
                     .map(sharding_method_from_proto)
                     .transpose()?,
+                strict_mode_config: value.strict_mode_config.map(|i| i.into()),
```
coszio, line 65: `From::from` has better navigation from rust-analyzer :)
```suggestion
                strict_mode_config: value.strict_mode_config.map(From::from),
```

src/migrations/single_to_cluster.rs
```diff
@@ -62,6 +62,7 @@ pub async fn handle_existing_collections(
                 optimizers_config: Some(collection_state.config.optimizer_config.into()),
                 init_from: None,
                 quantization_config: collection_state.config.quantization_config,
+                strict_mode_config: None,
```
coszio, line 65: why `None`?

JojiiOfficial: Good catch! I was adding this a bunch of tests and accidentally also set this to none.

lib/collection/src/operations/config_diff.rs
```diff
@@ -85,6 +86,79 @@ pub struct HnswConfigDiff {
...
+    // Global
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub enabled: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_filter_limit: Option<usize>,
```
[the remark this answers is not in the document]
JojiiOfficial: I didn't call it 'limit' so we can have different limits for different contexts in future.
I agree with `filter_limit` sounding a bit confusing but since we're sharing this with eg. discovery, `search_limit` might also be confusing as well.
What do you think?

lib/collection/src/operations/config_diff.rs
JojiiOfficial: CC @generall

lib/collection/src/operations/config_diff.rs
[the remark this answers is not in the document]
coszio: `max_query_limit`? And specify the apis that it applies to

lib/collection/src/operations/config_diff.rs
JojiiOfficial: I just discussed this with @generall and he prefers to keep the amount of parameters low. 
To address that, I've merged `recommend_max_examples` and `discovery_max_context_size` into `max_input_examples` to make it more generic.

generall (approved):

lib/collection/src/config.rs
```diff
@@ -161,6 +161,9 @@ pub struct CollectionConfig {
     pub wal_config: WalConfig,
     #[serde(default)]
     pub quantization_config: Option<QuantizationConfig>,
+    #[serde(default)]
+    #[schemars(skip)]
+    pub strict_mode_config: Option<StrictModeConfig>,
```
coszio, line 166: TODO: Let's add some validation here and in `build.rs` for grpc. Specially for the fields that would be problematic to set to `0` like max limit and max timeout

coszio (commented):
A few extra comments :)

generall (approved):

lib/collection/src/operations/config_diff.rs
```diff
@@ -85,6 +86,74 @@ pub struct HnswConfigDiff {
...
+        self.max_query_limit.hash(state);
+        self.max_timeout.hash(state);
+        self.unindexed_filtering_retrieve.hash(state);
+        self.unindexed_filtering_update.hash(state);
+        self.search_max_hnsw_ef.hash(state);
+        self.search_allow_exact.hash(state);
+        self.search_max_oversampling
+            .map(|i| i.to_le_bytes())
+            .hash(state);
+    }
```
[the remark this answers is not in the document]
generall: This is :+1:

lib/collection/src/operations/config_diff.rs
```diff
@@ -85,6 +86,74 @@ pub struct HnswConfigDiff {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_timeout: Option<usize>,
+
+    /// Allow usage of unindexed fields in retrieval based (eg. search) filters.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub unindexed_filtering_retrieve: Option<bool>,
+
+    /// Allow usage of unindexed fields in filtered updates (eg. delete by payload).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub unindexed_filtering_update: Option<bool>,
```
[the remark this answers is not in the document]
coszio: And what does `None` mean?

PR #4893, opened 2024-08-15
Alternative for: Filter update operations during resharding

timvisee:
Alternative resharding pre filter implementation on top of the current PR (<https://github.com/qdrant/qdrant/pull/4867>).

Reimplements the resharding pre filter.

This functions more as a request for comment, rather than a direct replacement.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection/resharding_update_pre_filter.rs
+++ b/lib/collection/src/collection/resharding_update_pre_filter.rs
@@ -23,8 +23,11 @@ pub struct ReshardingUpdatePreFilter {
     /// - *non* target shards during resharding *down*
     is_receiver_shard: bool,
 
-    /// Hashring filter, that selects points that are hashed into *target* shard
-    filter: hash_ring::HashRingFilter,
+    /// Filter that selects point IDs that are moved, based on resharding hash rings.
+    filter: hash_ring::HashRingMovedFilter,
+
+    /// Resharding direction
+    direction: ReshardingDirection,
 }
 
 impl ReshardingUpdatePreFilter {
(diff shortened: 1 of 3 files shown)


lib/collection/src/collection/resharding_update_pre_filter.rs
```diff
@@ -62,32 +65,33 @@ impl ReshardingUpdatePreFilter {
...
-        // - on *receiver* shards during `MigratingPoints` stage
-        // - and on *sender* shards during `ReadHashRingCommitted` stage when resharding *up*
-        let should_filter = (is_receiver_shard && state.stage == ReshardStage::MigratingPoints)
-            || (is_sender_shard
-                && state.stage >= ReshardStage::ReadHashRingCommitted
-                && state.direction == ReshardingDirection::Up);
-
-        if !should_filter {
+        // After write hash ring commit, we stop filtering
+        if state.stage >= ReshardStage::WriteHashRingCommitted {
```
ffuugoo, line 69: I never explained this explicitly, but I don't think we can do this. E.g.:

- Node A already received `WriteHashRingCommitted` message from consensus
  - and so it does not apply pre-filter locally anymore
- but Node B did not yet received `WriteHashRingCommitted` from consensus
  - and so it will still send all operations to new and *old* shards both
- now, user sends `upsert point 1337` to Node B
  - Node B forwards this update to *old* shard on Node A
  - and Node A inserts this point into *old* shard, because it does not apply pre-filter anymore

Basically, I think we should apply pre-filter until resharding is finished.

ffuugoo:
Tbh, I don't see much difference with #4867.

lib/collection/src/collection/resharding_update_pre_filter.rs
timvisee: > Basically, I think we should apply pre-filter until resharding is finished.

Fine by me. We can eventually fine tune it if desired.

timvisee:
> Tbh, I don't see much difference with #4867.

I agree it's not very different. I'd like to have seen a more significant change.

But I do see this:
- easier to read apply function
- slightly different decision tree better representing what we want

ffuugoo:
Closing, cause we implemented a different approach in https://github.com/qdrant/qdrant/pull/4928.

PR #4894, opened 2024-08-15
Update Validator 0.18

agourlay:
This update our `Validator` infrastructure to the [0.18.1](https://github.com/Keats/validator/blob/master/CHANGELOG.md) version.

The important changes are:
- `#[validate]` becomes `#[validate(nested)]` for fields implementing `Validate`.
- new syntax for validation with custom function
- a validation on an `Option<_>` field will be executed on the contained type if the option is `Some`.


lib/api/src/grpc/validate.rs
```diff
@@ -40,19 +40,6 @@ where
     }
 }
 
-impl<V> ValidateExt for Vec<V>
```
agourlay, line 43: An implement for `Vec<V>` is now in the library creating conflicts.
https://github.com/Keats/validator/blob/0872516333bf365668a22e522c57dac64e60d398/validator/src/traits.rs#L50

lib/api/src/grpc/validate.rs
```diff
@@ -349,19 +336,12 @@ pub fn validate_geo_polygon_line_helper(line: &grpc::GeoLineString) -> Result<()
...
-) -> Result<(), ValidationError> {
-    match line {
-        Some(l) => {
-            if l.points.is_empty() {
-                return Err(ValidationError::new("not_empty"));
-            }
-            validate_geo_polygon_line_helper(l)?;
-            Ok(())
-        }
-        _ => Err(ValidationError::new("not_empty")),
```
agourlay, line 363: the None case is checked with an extra validation in https://github.com/qdrant/qdrant/blob/6336395685187dffdcd786e1b3f6f5233541195a/lib/api/src/grpc/conversions.rs#L1475

lib/storage/src/rbac/mod.rs
```diff
@@ -23,13 +23,15 @@ pub enum Access {
...
+
 #[derive(Serialize, Deserialize, Validate, PartialEq, Clone, Debug)]
+#[validate(context = ExistingCollections, mutable)]
 pub struct CollectionAccess {
     /// Collection names that are allowed to be accessed
-    #[validate(custom(
-        function = "validate_unique_collections",
-        arg = "&'v_a mut HashSet<String>"
-    ))]
+    #[validate(custom(function = "validate_unique_collections", use_context))]
```
agourlay, line 34: the new macro refused to work with `HashSet<String>` and forced me to create a new struct.

See context in https://github.com/Keats/validator?tab=readme-ov-file#custom

lib/api/src/grpc/validate.rs
timvisee: This is fantastic. This implementation has caused some hidden panics before. :smile:

lib/storage/src/rbac/mod.rs
timvisee: :disappointed: 

Since this struct doesn't leak to other places I'm totally fine with it.

PR #4895, opened 2024-08-15
shell.nix: Allow to build on x86 darwin

SevInf:
I've tried to check out qdrant source code and build it on x86 mac laptop, however, i found out a couple of issues with current `shell.nix`.

- Couple of dependencies were missing.
- Default `CXXFLAGS` are not compatible with rocks-db (needs higher deployment target).
- Mold setup did not work with clang out of the box.

This PR fixes all of the issues and allows to build `qdrant` from within nix shell on Intel macOS. I don't have access to ARM Mac at the moment, so I can not say for sure if this configuration will work there.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -38,10 +38,8 @@ let
     preferWheels = true; # wheels speed up building of the environment
   };
 
-  # Use mold linker to speed up builds
-  mkShellMold = pkgs.mkShell.override { stdenv = pkgs.stdenvAdapters.useMoldLinker pkgs.stdenv; };
 in
-mkShellMold {
+pkgs.mkShell {
   buildInputs = [
     # Rust toolchain
     cargo-wrapper # should be before rust-combined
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -60,6 +58,10 @@ mkShellMold {
     pkgs.jq # used in ./tests and ./tools
     pkgs.nixfmt-rfc-style # to format this file
     pkgs.npins # used in tools/nix/update.py
+    pkgs.iconv
+    pkgs.ccache # compile cache for c/cxx code
+    pkgs.sccache # compile cache for rust code
```
timvisee, line 63: I don't think we want this by default. Any specific reason for setting it up here?

SevInf: It's already used in a shell hook below unconditionally for `RUST_WRAPPER`. 
I can either install nix package on darwin only, remove ccache/sccache usage from shell hook or make it conditional too. What would you prefer?

arthurk:
I've tried a build on this branch using an old intel mac and it worked. Linux works fine without this branch too, so maybe the `iconv` pkg is not necessary there.

shell.nix
xzfc: It was there before, and I prefer not to remove it.

The point of `shell.nix` is not just to have a bare minimum to build Qdrant, but rather to make local development comfortable. So it's a bit opinionated and includes stuff like rust-analyzer, build caching tricks, and stuff to make out git pre-push hooks work.

shell.nix
```diff
@@ -60,6 +58,10 @@ mkShellMold {
     pkgs.jq # used in ./tests and ./tools
     pkgs.nixfmt-rfc-style # to format this file
     pkgs.npins # used in tools/nix/update.py
+    pkgs.iconv
```
xzfc, line 61: Could you please leave a short comment on why `iconv` is needed / which dependency wants it?

SevInf: Done

xzfc (approved):

PR #4898, opened 2024-08-15
Fix: allow to specify shard_key in `with_lookup`

### LINKED ISSUE #4425: `with_lookup` lacks of the `shard_key` parameter
## Current Behavior

The [`with_lookup`](https://api.qdrant.tech/master/api-reference/search/recommend-point-groups#request.body.with_lookup) parameter misses a `with_lookup.shard_key` sub-parameter.
Instead, the shard key of the original table is used.

In other words, this method tries to reuse the `SHARD_KEY_1` both for the `COL1` and `COL2`.
```
POST collections/COL1/points/recommend/groups
{
  …,
  "shard_key": "SHARD_KEY_1",
  "with_lookup": { "collection": "COL2" }
}
```

## Steps to Reproduce
Prepare:
```
DELETE collections/t1?wait=true

DELETE collections/t2?wait=true

PUT collections/t1?wait=true
{
  "vectors": {"size": 1, "distance": "Dot"},
  "sharding_method": "custom"
}

PUT collections/t2?wait=true
{"vectors": {"size": 1, "distance": "Dot"}}

PUT collections/t1/shards?wait=true
{
  "shard_key": "sk1"
}

PUT collections/t1/points?wait=true
{
  "points": [
    {"id": 0, "vector": [0], "payload": {"p": 0}},
    {"id": 1, "vector": [1], "payload": {"p": 1}}
  ],
  "shard_key": "sk1"
}

PUT collections/t2/points?wait=true
{
  "points": [{"id": 0, "vector": [0]}, {"id": 1, "vector": [1]}]
}
```

Run:
```
POST collections/t1/points/search/groups
{
  "vector": [2],
  "limit": 30,
  "group_by": "p",
  "group_size": 10,
  "shard_key": "sk1",
  "with_lookup": {"collection": "t2"}
}
```

This produces the following error because `t2` lacks of the `sk1` key.
```json
{
  "error": "Bad request: Shard key \"sk1\" not found"
}
```

## Expected Behavior

It should not try to use the shard key `sk1` for the collection `t2`.

## Possible Solution

Implement either the first one, or both.

1. Ignore the outer `.shard_key` for the collection mentioned in `.with_lookup`. I consider this is a bug.
2. Add a new parameter `.with_lookup.shard_key`. I consider this is a missing feature.

SevInf:
This PR introduces following changes to group by endpoint:

1. Stop re-using outer collection `shard_key` for lookup of the nested collection. Lookup will now default to searching on all shards.
2. Allows to specify `shard_key` in `with_lookup` parameter of group by to perform the lookup only within specific shard(s).

Adds python integration tests that verifies that both parts work as expected.

Fix #4425

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -11491,6 +11491,17 @@
                 "nullable": true
               }
             ]
+          },
+          "shard_key": {
+            "description": "Shard key for lookup. If not specified, all shards would be searched",
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/ShardKeySelector"
+              },
+              {
+                "nullable": true
+              }
+            ]
           }
         }
       },
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -2220,6 +2220,7 @@ impl TryFrom<WithLookup> for rest::WithLookup {
                 .transpose()?
                 .or_else(with_default_payload),
             with_vectors: value.with_vectors.map(|wv| wv.into()),
+            shard_key: value.shard_key_selector.map(Into::into),
         })
     }
 }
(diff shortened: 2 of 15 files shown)


timvisee:
I've only skimmed this yet.

I'm afraid your first noted change is breaking, and am therefore not sure if we
can change it. An alternative may be to explicitly specify what shard key we use
for nested requests [inherit, all, specific].

Also, I'm curious, do you have a specific use case that would benefit from this
change?

PR #4901, opened 2024-08-16
Wrap handler body in helpers::time

JojiiOfficial:
Wraps all cancel-safe operations in actix handlers in helpers::time.
This allows early-returning using `?` and makes the code cleaner.

### DIFF
--- a/src/actix/api/collections_api.rs
+++ b/src/actix/api/collections_api.rs
@@ -43,10 +43,8 @@ fn get_collections(
 async fn get_aliases(
     dispatcher: web::Data<Dispatcher>,
     ActixAccess(access): ActixAccess,
-) -> impl Responder {
-    let timing = Instant::now();
-    let response = do_list_aliases(dispatcher.toc(&access), access).await;
-    process_response(response, timing)
+) -> HttpResponse {
+    helpers::time(async { do_list_aliases(dispatcher.toc(&access), access).await }).await
 }
 
 #[get("/collections/{name}")]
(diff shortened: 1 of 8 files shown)


src/actix/api/collections_api.rs
```diff
@@ -43,44 +43,44 @@ fn get_collections(
 async fn get_aliases(
     dispatcher: web::Data<Dispatcher>,
     ActixAccess(access): ActixAccess,
-) -> impl Responder {
-    let timing = Instant::now();
-    let response = do_list_aliases(dispatcher.toc(&access), access).await;
-    process_response(response, timing)
+) -> HttpResponse {
+    helpers::time(async { do_list_aliases(dispatcher.toc(&access), access).await }).await
```
generall, line 47:
```suggestion
    helpers::time(do_list_aliases(dispatcher.toc(&access), access)).await
```

And in other places

src/actix/api/count_api.rs
```diff
@@ -19,28 +18,27 @@ async fn count_points(
...
+        } = request.into_inner();
 
-    let CountRequest {
-        count_request,
-        shard_key,
-    } = request.into_inner();
+        let shard_selector = match shard_key {
+            None => ShardSelectorInternal::All,
+            Some(shard_keys) => ShardSelectorInternal::from(shard_keys),
+        };
```
generall, line 30: This might be outside of time function, so we can avoid `async {}` frapper

src/actix/api/recommend_api.rs
generall, line 1: Also propose to do this

JojiiOfficial: Since we add strict mode checking (which is an async function) soon anyways, we're probably going to add most async blocks back again we remove now, so I'd suggest to keep them.
What do you think?

generall: could you pls share an example of how it might look like?

JojiiOfficial: For searching points something like this:
```rust
async fn search_points(
    dispatcher: web::Data<Dispatcher>,
    collection: Path<CollectionPath>,
    request: Json<SearchRequest>,
    params: Query<ReadParams>,
    ActixAccess(access): ActixAccess,
) -> HttpResponse {
    helpers::time(async {
        let verification_pass = request
            .check(&dispatcher, &access, &collection.name)
            .await?;

        let SearchRequest {
            search_request,
            shard_key,
        } = request.into_inner();

        let shard_selection = match shard_key {
            None => ShardSelectorInternal::All,
            Some(shard_keys) => shard_keys.into(),
        };

        do_core_search_points(
        ...
```

generall (approved):

PR #4911, opened 2024-08-19
tests: fix flaky `correct_positive_order`

coszio:
Last fix (#4815) didn't account for when the input would result in `before = Ordering::Greater, after = Ordering::Equal`. 

To fix this, we now do `total_cmp` on the "after", and skip the cases where the relaxed cmp on the "before" is Equal

### DIFF
--- a/lib/segment/src/vector_storage/query/reco_query.rs
+++ b/lib/segment/src/vector_storage/query/reco_query.rs
@@ -177,13 +177,17 @@ mod test {
 
             let ordering_before = float_cmp(dummy_similarity(&a), dummy_similarity(&b));
 
+            // Too similar scores can get compressed to the same value by the sigmoid function.
+            // This would make the test useless, so we skip those cases.
+            prop_assume!(ordering_before != Ordering::Equal);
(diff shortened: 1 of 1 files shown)


lib/segment/src/vector_storage/query/reco_query.rs
```diff
@@ -177,13 +177,17 @@ mod test {
 
             let ordering_before = float_cmp(dummy_similarity(&a), dummy_similarity(&b));
 
+            // Too similar scores can get compressed to the same value by the sigmoid function.
+            // This would make the test useless, so we skip those cases.
+            prop_assume!(ordering_before != Ordering::Equal);
```
timvisee, line 182: I wasn't aware of this macro. Cool stuff.

timvisee (approved):
Thanks!

lib/segment/src/vector_storage/query/reco_query.rs
coszio: I also learned about it today :D

PR #4923, opened 2024-08-19
Allow explicit populate of mmap

generall:
- Instead of in-ram + mmap we can just force mmap in ram with populate
- Should solve double memory usage of experimental vector storage along with cold start and OOM problems
- required benchmarks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -249,3 +249,5 @@ codegen-units = 256 # restore default value for faster compilation
 [patch.crates-io]
 # Temporary patch until our PRs are merged.
 tar = { git = "https://github.com/qdrant/tar-rs", branch="main" }
+# Temporary patch until memmap2 0.9.5 is released.
+memmap2 = { git = "https://github.com/RazrFalcon/memmap2-rs" }
--- a/lib/collection/src/config.rs
+++ b/lib/collection/src/config.rs
@@ -359,7 +359,7 @@ impl CollectionParams {
                         storage_type: if params.on_disk.unwrap_or_default() {
                             VectorStorageType::ChunkedMmap
                         } else {
-                            VectorStorageType::Memory
+                            VectorStorageType::InRamChunkedMmap
                         },
                         multivector_config: params.multivector_config,
                         datatype: params.datatype.map(VectorStorageDatatype::from),
--- a/lib/collection/src/config.rs
+++ b/lib/collection/src/config.rs
@@ -359,7 +359,7 @@ impl CollectionParams {
                         storage_type: if params.on_disk.unwrap_or_default() {
                             VectorStorageType::ChunkedMmap
                         } else {
-                            VectorStorageType::InRamChunkedMmap
+                            VectorStorageType::Memory
                         },
                         multivector_config: params.multivector_config,
                         datatype: params.datatype.map(VectorStorageDatatype::from),
--- a/lib/common/memory/src/madvise.rs
+++ b/lib/common/memory/src/madvise.rs
@@ -1,9 +1,12 @@
 //! Platform-independent abstractions over [`memmap2::Mmap::advise`]/[`memmap2::MmapMut::advise`]
 //! and [`memmap2::Advice`].
 
+use std::hint::black_box;
 use std::io;
 
 use serde::Deserialize;
+
+const PAGE_SIZE: usize = 4096;
 
 /// Global [`Advice`] value, to trivially set [`Advice`] value
 /// used by all memmaps created by the `segment` crate.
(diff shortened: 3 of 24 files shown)


generall:
image

generall:
depends on https://github.com/qdrant/qdrant/pull/4943

lib/collection/src/config.rs
```diff
@@ -355,7 +355,7 @@ impl CollectionParams {
                         storage_type: if params.on_disk.unwrap_or_default() {
                             VectorStorageType::ChunkedMmap
                         } else {
-                            VectorStorageType::Memory
+                            VectorStorageType::InRamChunkedMmap
```
generall, line 358:
```suggestion
                            VectorStorageType::Memory
```

generall: Disable for now

lib/common/memory/src/mmap_ops.rs
```diff
@@ -107,14 +120,15 @@ where
...
-
-    for chunk in mmap.chunks(dst.len()) {
-        dst[..chunk.len()].copy_from_slice(chunk);
+    if let Err(err) = mmap.populate() {
+        log::warn!(
+            "Failed to populate mmap{separator}{path:?} to cache: {err}",
+            separator = separator,
+            path = path,
+            err = err
+        );
```
xzfc, line 129: Style nit:

```suggestion
        log::warn!(
            "Failed to populate mmap{separator}{path:?} to cache: {err}"
        );
```

---

Also, It was already before this PR, but I don't think this `separator` logic works as expected. If `path` is None, it will print this: `Failed to populate mmap"" to cache: …`. Maybe it should be like this:

```rust
let path = path.map_or_else(|| "".to_string(), |path| format!(" {path:?}"));
…
log::warn!("Failed to populate mmap{path} to cache: {err}");
```

lib/common/memory/src/madvise.rs
```diff
@@ -117,4 +142,22 @@ impl Madviseable for memmap2::MmapMut {
         log::debug!("Ignore {advice:?} on this platform");
         Ok(())
     }
+
+    fn populate(&self) -> io::Result<()> {
+        #[cfg(target_os = "linux")]
+        self.advise(memmap2::Advice::PopulateRead)?;
```
xzfc, line 148: This option requires Linux 5.14+. Some long-term supported distros (e.g. RHEL 8 and Ubuntu 20.04) still have older kernels. Maybe it would make sense to check beforehand if it is supported and fallback to the loop implementation if it isn't. Quote from [`madvise(2)`](https://man7.org/linux/man-pages/man2/madvise.2.html):
> _madvise(0, 0, advice)_ will return zero iff _advice_ is supported by the kernel and can be relied on to probe for support.

xzfc: The check could be like this:
```rust
fn populate_read_is_supported() -> bool {
    use std::sync::OnceLock;
    static IS_SUPPORTED: OnceLock<bool> = OnceLock::new();
    *IS_SUPPORTED.get_or_init(|| unsafe {
        libc::madvise(std::ptr::null_mut(), 0, Advice::PopulateRead as libc::c_int) == 0
    })
}
```

timvisee: If we're running 5.14 (at least) in the cloud I don't think we should care too much about older versions.

We can, but I prefer to keep things simpler.

timvisee: In some cases we do use an older kernel version. So, lets follow your suggestion with the runtime check.

@xzfc I've tagged you in Slack where we've discussed this.

xzfc: Recap: we still have older versions somewhere.

And even if we didn't, performing this tiny check shouldn't complicate things much, since we already have the fallback code for non-Linux systems.

xzfc: I've implemented that in the last commit to this PR. It uses a new `mmap2::Advise::is_supported()` method added in https://www.github.com/RazrFalcon/memmap2-rs/pull/122.

timvisee (approved):
LGTM assuming the open review comments get resolved

Cargo.toml
```diff
@@ -249,3 +249,5 @@ codegen-units = 256 # restore default value for faster compilation
 [patch.crates-io]
 # Temporary patch until our PRs are merged.
 tar = { git = "https://github.com/qdrant/tar-rs", branch="main" }
+# Temporary patch until memmap2 0.9.5 is released.
+memmap2 = { git = "https://github.com/RazrFalcon/memmap2-rs" }
```
timvisee, line 253: We bump this to the latest release in <https://github.com/qdrant/qdrant/pull/5085>.

PR #4928, opened 2024-08-20
Split operations into `update_all`/`update_existing` during resharding

ffuugoo:
This is _yet another_ alternative to filtering operations on read/write during resharding. The (main) goal is the same as with filtering: prevent any "stale" data to show up in read request results during `CommitReadHashRing`/`CommitWriteHashRing` transition.

This PR reuses a lot of #4867 code, but this time it's not used to filter-out point IDs from update operations, but to "split" update operations into "update all points"/"update existing points" operations (conceptually similar to `split_by_shard`). These split operations are then applied to the replica set (mostly) as usual. The only difference is that for "update existing points" operations, we simply ignore any "No point with id found" errors.

There is one moderate caveat with this PR: if we want to apply "upsert" operation in "update existing points" mode, we have to convert it to a sequence of "update vector"/"overwrite payload" operations. Qdrant can only "overwrite payload" of a single point at a time, so, e.g., a single "upsert" of a 100 points (with a payload) would be converted to a 101 operation (1 "update vector" and 100 separate "set payload" operations), which then have to be propagated over network, which tanks performance somewhat. In a simple test I see 50% performance hit (e.g., 100k points take 30 seconds without splitting and ~45 seconds with splitting). This can be improved in a follow-up PR, though, so it should not be a blocker for this PR.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/operations/point_ops.rs
+++ b/lib/collection/src/operations/point_ops.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::iter;
 
 use api::rest::{BatchVectorStruct, ShardKeySelector, VectorStruct};
 use itertools::izip;
@@ -12,9 +13,15 @@ use serde::{Deserialize, Serialize};
 use strum::{EnumDiscriminants, EnumIter};
 use validator::Validate;
 
-use super::{point_to_shards, split_iter_by_shard, OperationToShard, SplitByShard};
+use super::payload_ops::SetPayloadOp;
+use super::vector_ops::{PointVectors, UpdateVectorsOp};
+use super::{
+    point_to_shards, split_iter_by_shard, CollectionUpdateOperations, OperationToShard,
+    SplitByShard,
+};
 use crate::hash_ring::HashRingRouter;
 use crate::operations::types::Record;
+use crate::operations::{payload_ops, vector_ops};
 use crate::shards::shard::ShardId;
 
 /// Defines write ordering guarantees for collection operations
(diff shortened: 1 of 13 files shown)


lib/collection/src/shards/replica_set/update.rs
```diff
@@ -128,7 +129,7 @@ impl ShardReplicaSet {
 
         peer_ids
             .into_iter()
-            .filter(|peer_id| self.peer_is_active(peer_id)) // re-acquire replica_state read lock
+            .filter(|peer_id| self.peer_is_active_or_resharding(peer_id)) // re-acquire replica_state read lock
```
ffuugoo, line 132: Small fix: allows selecting replicas in `Resharding` state as a leader peer for `Medium` write ordering.

When resharding "up" we only have a single replica, and without this fix updates with `Medium` ordering would fail until all resharding transfers are finished and replica is switched into `Active`.

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -450,14 +455,18 @@ impl ShardReplicaSet {
...
+                && err.is_pre_condition_failed()
             {
                 // Handles a special case where transfer receiver haven't created a shard yet.
                 // In this case update should be handled by source shard and forward proxy.
                 continue;
             }
 
+            if update_existing && err.is_missing_point() {
+                continue;
+            }
```
ffuugoo, line 468: Explicitly ignore "missing point" errors for "update existing" operations.

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -364,6 +369,107 @@ impl ShardHolder {
...
+            OperationsByMode::default().update_existing(operation)
+        } else {
+            let mut update_all = operation.clone();
+            update_all.retain_point_ids(|point_id| !target_point_ids.contains(point_id));
+
+            let mut update_existing = operation;
+            update_existing.retain_point_ids(|point_id| target_point_ids.contains(point_id));
+
+            OperationsByMode::from(update_all).update_existing(update_existing)
+        }
```
ffuugoo, line 470: We need to extract all "target" point IDs from the operation into a separate "update existing" operation.

Conceptually, this is a lot like `split_by_shard`, but it was too annoying to copy-and-tweak `split_by_shard` as `split_by_mode` or to try to generalize `split_by_shard`/`split_by_mode` into single method.

So I've made a less optimal, but simpler impl based on `get_point_ids` and `retain_point_ids` methods: if operation have to be split in two, we just clone it and then do `retain_point_ids` on both copies.

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -391,6 +497,39 @@ impl ShardHolder {
...
+}
+
+impl OperationsByMode {
+    pub fn update_existing(mut self, operation: CollectionUpdateOperations) -> Self {
+        match operation {
+            CollectionUpdateOperations::PointOperation(
+                point_ops::PointOperations::UpsertPoints(operation),
+            ) => {
+                self.update_existing = operation.into_update_only();
+            }
```
ffuugoo, line 513: Convert `upsert` operation into "update vector"/"overwrite payload", if we want to apply `upsert` in "update existing" mode.

src/main.rs
```diff
@@ -348,9 +348,9 @@ fn main() -> anyhow::Result<()> {
             }
         });
 
-        runtime_handle.block_on(async {
-            toc_arc.resume_resharding_tasks().await;
-        });
+        // runtime_handle.block_on(async {
+        //     toc_arc.resume_resharding_tasks().await;
+        // });
```
ffuugoo, line 353: Disable resuming resharding driver task when starting Qdrant. I've missed it when disabling resharding driver earlier. 😅

ffuugoo (commented):
Some notes and pointers for the reviewers.

lib/collection/src/operations/types.rs
```diff
@@ -984,6 +984,18 @@ impl CollectionError {
...
         }
     }
+
+    pub fn is_pre_condition_failed(&self) -> bool {
+        matches!(self, Self::PreConditionFailed { .. })
+    }
+
+    pub fn is_missing_point(&self) -> bool {
+        match self {
+            CollectionError::NotFound { what } => what.contains("No point with id"),
```
timvisee, line 994: I would prefer some strict type checking here, but I understand that may not really be possible right now.

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -71,6 +71,7 @@ impl ShardReplicaSet {
         operation: CollectionUpdateOperations,
         wait: bool,
         ordering: WriteOrdering,
+        update_existing: bool,
```
timvisee, line 74: I was a bit confused by this name when I first came across it, especially because we pass it along a few times.

Would `only_update_existing` be a bit better?

I couldn't really come up with any other good name, and don't like `ignore_non_exists_errors`.

lib/collection/src/operations/point_ops.rs
```diff
@@ -216,6 +223,200 @@ pub enum PointInsertOperationsInternal {
...
+                if let Some(payload) = &mut batch.payloads {
+                    retain_with_index(payload, |index, _| retain_indices.contains(&index));
+                }
+            }
+
+            Self::PointsList(points) => points.retain(|point| filter(&point.id)),
+        }
+    }
+
+    pub fn into_update_only(self) -> Vec<CollectionUpdateOperations> {
```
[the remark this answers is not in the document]
timvisee: We can make it less bulky with type conversions or by extracting the `PointsBatch` and `PointsList` variant branches into separate methods.

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -364,6 +369,107 @@ impl ShardHolder {
         Ok(())
     }
 
+    pub fn split_by_mode(
```
timvisee, line 372:
```suggestion
    /// Split collection update operations in updating all and just existing points.
    ///
    /// Depends on the current resharding state. If resharding is not active operations are not split.
    pub fn split_by_mode(
```

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -364,6 +369,107 @@ impl ShardHolder {
...
+            && state.direction == ReshardingDirection::Up;
+
+        if !should_split_receiver && !should_split_sender {
+            return operation.into();
+        }
+
+        // There's no point splitting delete operations
+        if operation.is_delete_points() {
+            return operation.into();
+        }
```
timvisee, line 440: Do we return an error if deletion fails? If so, I think we should split it to prevent this.

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -364,6 +369,107 @@ impl ShardHolder {
...
+
+        let filter = self
+            .resharding_filter()
+            .expect("resharding filter is available when resharding is in progress");
```
timvisee, line 444: We don't hold a lock here on the resharding state. So it may be possible to update it in between.

If so, we shouldn't panic but return here.

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -391,6 +497,39 @@ impl ShardHolder {
...
 }
 
+#[derive(Clone, Debug, Default)]
+pub struct OperationsByMode {
+    pub update_all: Vec<CollectionUpdateOperations>,
+    pub update_existing: Vec<CollectionUpdateOperations>,
+}
+
+impl OperationsByMode {
+    pub fn update_existing(mut self, operation: CollectionUpdateOperations) -> Self {
```
timvisee, line 507: Nitpick, how about `with_update_existing`?

lib/collection/src/shards/shard_holder/resharding.rs
timvisee: It does not error.

src/main.rs
ffuugoo: Done.

lib/collection/src/operations/types.rs
ffuugoo: Yeah, too much refactoring to propagate this error properly. :/

lib/collection/src/shards/replica_set/update.rs
ffuugoo: We could not come up with a unanimously better name, so we decided to leave it as is for now. :/

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo: Added a more extensive documentation, wdyt?

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo: Good point. Fixed.

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo: Done.

lib/collection/src/shards/shard_holder/resharding.rs
timvisee: Is this still relevant?

ffuugoo: It is. We don't have any other way to handle this rn.

timvisee: I believe it is handled in `into_update_only()`

timvisee (approved):

lib/collection/src/operations/types.rs
```diff
@@ -1019,6 +1019,18 @@ impl CollectionError {
...
```
generall, line 1029: Maybe use something more reliable?

generall (approved):
LGTM, except for the error matching logic. Please create a task to fix that it following PRs

ffuugoo:
> LGTM, except for the error matching logic. Please create a task to fix that it following PRs

https://github.com/qdrant/qdrant/issues/5063

lib/collection/src/operations/types.rs
ffuugoo: Opened an issue for that https://github.com/qdrant/qdrant/issues/5063

PR #4929, opened 2024-08-20
fix mismatch of index parametrization and type

coszio:
We had a bug where selecting `field_type: Bool` and `field_index_params: { bool_index_params: {} }` would result in a mismatch error, and same for Geo type. 

With these changes we fix it and also we make sure to not forget it in the future via a compiler error.


timvisee (approved):

joein (approved):
Checked with python, works now 👍 
Thanks

PR #4935, opened 2024-08-22
[Strict-mode] Verification check in API

JojiiOfficial:
Adds a trait for request verification and implements it for the search REST API.

To ensure we don't forget to implement checks for future endpoints, a new `toc_new` function gets introduced that requires a new, empty type created when verification was successful. 

We also need to check for access rights in strict mode, since it gets executed before the operation, which usually does the check. An alternative would be to combine `Access` and `Verification`, let `CollectionRequestVerification` do the access check and use only `Verification` as parameter for `toc(..)`.

### DIFF
--- a/src/actix/api/search_api.rs
+++ b/src/actix/api/search_api.rs
@@ -8,13 +8,14 @@ use collection::operations::types::{
 };
 use futures::TryFutureExt;
 use itertools::Itertools;
+use storage::content_manager::collection_verification::CollectionRequestVerification;
 use storage::dispatcher::Dispatcher;
 use tokio::time::Instant;
 
 use super::read_params::ReadParams;
 use super::CollectionPath;
 use crate::actix::auth::ActixAccess;
-use crate::actix::helpers::{self, process_response};
+use crate::actix::helpers::{self, process_response, process_response_error};
 use crate::common::points::{
     do_core_search_points, do_search_batch_points, do_search_point_groups, do_search_points_matrix,
 };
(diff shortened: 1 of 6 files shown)


src/actix/api/search_api.rs
```diff
@@ -27,6 +28,11 @@ async fn search_points(
     params: Query<ReadParams>,
     ActixAccess(access): ActixAccess,
 ) -> HttpResponse {
+    let pass = match request.check(&dispatcher, &access, &collection.name).await {
+        Ok(pass) => pass,
+        Err(err) => return process_response_error(err, Instant::now()),
```
JojiiOfficial, line 33: In order to prevent the usage of `async {}`, we return a 'dummy' time value here. 
If the timing information is important for errors too, I'll move it into an async block.

timvisee: Can't we just instantiate the time right at the start of the function, and pass it in here, without async?

src/actix/api/search_api.rs
timvisee, line 1: Am I correct that this currently only integrates it into search, and that we cover all the other interfaces later?

src/actix/api/search_api.rs
JojiiOfficial: You mean simply putting the `Instant::now()` call before strict-mode checking to measure it?
What about `helpers::time`, should we replace this with a custom implementation and pass the `Instant` from the start of the function too?

PR #4938, opened 2024-08-22
Expose resharding operations for collection in telemetry

timvisee:
Tracked in: #4213 

Expose resharding operations of a collection in telemetry.

This is needed so cluster manager can assess this state.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -10129,6 +10129,7 @@
           "config",
           "id",
           "init_time_ms",
+          "resharding",
           "shards",
           "transfers"
         ],
@@ -10154,6 +10155,12 @@
             "type": "array",
             "items": {
               "$ref": "#/components/schemas/ShardTransferInfo"
+            }
+          },
+          "resharding": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/ReshardingInfo"
             }
           }
         }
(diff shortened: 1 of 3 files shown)


ffuugoo (approved):

ffuugoo (approved):
The code is LGTM, but it does not make much practical sense, cause `ReshardingInfo` depends on *resharding tasks/drivers*... but we disabled resharding driver, and so the infos will always be empty.

I think we should refactor this to
- return `ReshardStage` from `ResrhardState` (cause this *will* be useful/required for `cluster-manager`, I think)
- return status on all peers, not just the (former) driver-peer (cause the only resharding state in Qdrant is consensus-state now, and so it's available on all peers)

timvisee:
> The code is LGTM, but it does not make much practical sense, cause `ReshardingInfo` depends on _resharding tasks/drivers_... but we disabled resharding driver, and so the infos will always be empty.

IIRC that's only for the `comment`.

It currently reports the state we need for cluster manager correctly when it is being driven by CM.

With regards to the status comment, we'll probably remove it when removing the driver once it's fully implemented in cluster manager. Because then the field will become useless.

> I think we should refactor this to
> 
>     * return `ReshardStage` from `ReshardState` (cause this _will_ be useful/required for `cluster-manager`, I think)
> 
>     * return status on all peers, not just the (former) driver-peer (cause the only resharding state in Qdrant is consensus-state now, and so it's available on all peers)

We'll probably remove the stages here. They are for the driver in CM, and we'll be removing the driver here. This will then only report the current consensus state we have on it.

> (Approved, cause we can still merge this without much downside, and committed code would at least mark places that would require refactoring.)

Yes, sound good.

ffuugoo:
> IIRC that's only for the `comment`.

You're right. Missed that.

>> I think we should refactor this to

Then this is even simpler: just replace `comment` with `stage` (i.e., the one in consensus) and be done with it.

PR #4940, opened 2024-08-22
Introduce grey collection status and expose shard status in telemetry

KShivendu:
Introduces a `status` field in telemetry response for each of the local shard. It will have same possible values as the collection status (green, yellow, grey, red). So shard telemetry will look like:
```jsonc
{
    "shards": [
        {
            "id": 0,
            "local": {
                "variant_name": "local shard",
                "status": "green", // --------------------> new addition
                "segments": [ "..." ],
                "optimizations": {
                    "status": "ok",
                    "optimizations": {
                        "count": 0,
                        "total_duration_micros": 0,
                        "last_responded": "2024-08-23T08:57:39.410Z"
                    },
                    "log": []
                }
            },
            "remote": [],
            "replicate_states": {
                "2762498644323566": "Active"
            }
        }
    ],
    "transfers": []
}
``` 

This feature will be used to assess if a replica is indexed enough.

This PR does the following:
- Introduces the grey collection status that was supposed to be released with `v1.10`. It means that shard is available, but some segments have pending optimization. 
- Renames `CollectionInfoInternal` to `ShardInfoInternal` to better represent that the struct represents shard level data. The same struct is eventually used to build `CollectionInfo` which is exposed by the APIs.
- Introduces a dedicated `ShardStatus` to be used in shard telemetry. This is also converted into `CollectionStatus` when representating collection status (say for `/collection/<collection>` API)

### All Submissions:

### DIFF
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -64,7 +64,7 @@ pub enum CollectionStatus {
     Green,
     // Collection is available, but some segments might be under optimization
     Yellow,
-    // Collection is available, but some segments are pending optimization
+    // Collection is available, but some segments are non-optimal and still optimizations aren't running
     Grey,
     // Something is not OK:
     // - some operations failed and was not recovered
@@ -80,7 +80,7 @@ pub enum ShardStatus {
     Green,
     // Shard is available, but some segments might be under optimization
     Yellow,
-    // Shard is available, but some segments are pending optimization
+    // Shard is available, but some segments are non-optimal and still optimizations aren't running
     Grey,
     // Something is not OK:
     // - some operations failed and was not recovered
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -76,11 +76,11 @@ pub enum CollectionStatus {
 #[derive(Debug, Serialize, JsonSchema, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
 #[serde(rename_all = "snake_case")]
 pub enum ShardStatus {
-    // Collection if completely ready for requests
+    // Shard if completely ready for requests
     Green,
-    // Collection is available, but some segments might be under optimization
+    // Shard is available, but some segments might be under optimization
     Yellow,
-    // Collection is available, but some segments are pending optimization
+    // Shard is available, but some segments are pending optimization
     Grey,
     // Something is not OK:
     // - some operations failed and was not recovered
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -90,10 +90,10 @@ pub enum ShardStatus {
 impl From<ShardStatus> for CollectionStatus {
     fn from(info: ShardStatus) -> Self {
         match info {
-            ShardStatus::Green => CollectionStatus::Green,
-            ShardStatus::Yellow => CollectionStatus::Yellow,
-            ShardStatus::Grey => CollectionStatus::Grey,
-            ShardStatus::Red => CollectionStatus::Red,
+            ShardStatus::Green => Self::Green,
+            ShardStatus::Yellow => Self::Yellow,
+            ShardStatus::Grey => Self::Grey,
+            ShardStatus::Red => Self::Red,
         }
     }
 }
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -60,7 +60,7 @@ use crate::wal::WalError;
 #[derive(Debug, Serialize, JsonSchema, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
 #[serde(rename_all = "snake_case")]
 pub enum CollectionStatus {
-    // Collection if completely ready for requests
+    // Collection is completely ready for requests
     Green,
     // Collection is available, but some segments might be under optimization
     Yellow,
@@ -76,7 +76,7 @@ pub enum CollectionStatus {
 #[derive(Debug, Serialize, JsonSchema, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
 #[serde(rename_all = "snake_case")]
 pub enum ShardStatus {
-    // Shard if completely ready for requests
+    // Shard is completely ready for requests
     Green,
     // Shard is available, but some segments might be under optimization
     Yellow,
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -56,7 +56,7 @@ use crate::shards::transfer::ShardTransferMethod;
 use crate::wal::WalError;
 
 /// Current state of the collection.
-/// `Green` - all good. `Yellow` - optimization is running, `Red` - some operations failed and was not recovered
+/// `Green` - all good. `Yellow` - optimization is running, 'Grey' - optimizations are possible but not triggered, `Red` - some operations failed and was not recovered
 #[derive(Debug, Serialize, JsonSchema, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
 #[serde(rename_all = "snake_case")]
 pub enum CollectionStatus {
@@ -64,7 +64,7 @@ pub enum CollectionStatus {
     Green,
     // Collection is available, but some segments might be under optimization
     Yellow,
-    // Collection is available, but some segments are non-optimal and still optimizations aren't running
+    // Collection is available, but some segments can be optimized still optimizations aren't triggered
     Grey,
     // Something is not OK:
     // - some operations failed and was not recovered
@@ -72,7 +72,7 @@ pub enum CollectionStatus {
 }
 
 /// Current state of the shard (supports same states as the collection)
-/// `Green` - all good. `Yellow` - optimization is running, `Red` - some operations failed and was not recovered
+/// `Green` - all good. `Yellow` - optimization is running, 'Grey' - optimizations are possible but not triggered, `Red` - some operations failed and was not recovered
 #[derive(Debug, Serialize, JsonSchema, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
 #[serde(rename_all = "snake_case")]
 pub enum ShardStatus {
@@ -80,7 +80,7 @@ pub enum ShardStatus {
     Green,
     // Shard is available, but some segments might be under optimization
     Yellow,
-    // Shard is available, but some segments are non-optimal and still optimizations aren't running
+    // Shard is available, but some segments can be optimized still optimizations aren't triggered
     Grey,
     // Something is not OK:
     // - some operations failed and was not recovered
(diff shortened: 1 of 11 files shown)


lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -969,13 +970,13 @@ impl LocalShard {
         vector_size * info.points_count
     }
 
-    pub async fn local_shard_info(&self) -> CollectionInfoInternal {
+    pub async fn local_shard_info(&self) -> ShardInfoInternal {
```
KShivendu, line 973: Should I just create a dedicated function for deciding shard status:
```rs
pub async fn local_shard_status(&self) -> ShardStatus { ... }
```

It won't have to compute `vectors_count`, `schema`, etc. However, it will have the part of the code from `local_shard_info` and will have to be kept in sync.

KShivendu:
```rs
pub async fn local_shard_status(&self) -> ShardStatus {
        let segments = self.segments().read();
        let mut status = ShardStatus::Green;
        let mut optimizer_status = OptimizersStatus::Ok;

        for (idx, segment) in segments.iter() {
            if segment_info.segment_type == SegmentType::Special {
                status = ShardStatus::Yellow;
            }

            if !segments.failed_operation.is_empty() || segments.optimizer_errors.is_some() {
                status = ShardStatus::Red;
            }

            if let Some(error) = &segments.optimizer_errors {
                optimizer_status = OptimizersStatus::Error(error.to_string());
            }
        }

        // If still green while optimization conditions are triggered, mark as grey
        if status == ShardStatus::Green
            && self.update_handler.lock().await.has_pending_optimizations()
            && optimizer_status == OptimizersStatus::Ok
        {
            status = ShardStatus::Grey;
            // optimizer_status =
            //     OptimizersStatus::Error("optimizations pending, awaiting update operation".into());
        }

        status
}
```

lib/collection/src/operations/types.rs
```diff
@@ -71,6 +71,33 @@ pub enum CollectionStatus {
     Red,
 }
 
+/// Current state of the shard (supports same states as the collection)
+/// `Green` - all good. `Yellow` - optimization is running, `Red` - some operations failed and was not recovered
+#[derive(Debug, Serialize, JsonSchema, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
+#[serde(rename_all = "snake_case")]
+pub enum ShardStatus {
```
generall, line 78: Do I get it right, that you don't want to reuse CollectionStatus, cause we don't want to change names exposed to the clients? (make sense, btw)

lib/collection/src/shards/local_shard/mod.rs
generall: separate function is preferred, as it looks like it would allow us to keep `get_telemetry_data` sync

lib/collection/src/operations/types.rs
KShivendu: Yeah that plus it makes more semantic sense to call it `ShardStatus` than `CollectionStatus`

timvisee: Isn't it only in telemetry? I believe we don't really care about changes there, because we don't guarantee stable typing there.

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -970,6 +970,39 @@ impl LocalShard {
         vector_size * info.points_count
     }
 
+    pub async fn shard_status(&self) -> ShardStatus {
+        // Note: Should be kept in sync with parts of local_shard_info
```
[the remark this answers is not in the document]
KShivendu: Just more computation of aggregating more stats like `vector_count`, `indexed_vector_count`, etc. But now that I think about it. It's won't make of a difference.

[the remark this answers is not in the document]
KShivendu: Did what you suggested. Reusing `local_shard_info`

lib/collection/src/operations/types.rs
KShivendu: It's still better semantics to call it ShardStatus?

lib/collection/src/shards/shard.rs
```diff
@@ -63,9 +63,13 @@ impl Shard {
...
 
-    pub fn get_telemetry_data(&self, detail: TelemetryDetail) -> LocalShardTelemetry {
+    pub async fn get_telemetry_data(&self, detail: TelemetryDetail) -> LocalShardTelemetry {
         let mut telemetry = match self {
-            Shard::Local(local_shard) => local_shard.get_telemetry_data(detail),
+            Shard::Local(local_shard) => {
+                let mut shard_telemetry = local_shard.get_telemetry_data(detail);
+                shard_telemetry.status = Some(local_shard.local_shard_info().await.status);
+                shard_telemetry
+            }
```
timvisee, line 72: We only do it this way because we want to expose the status to cluster manager, am I right?

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -1009,19 +1010,16 @@ impl LocalShard {
         }
 
         // If still green while optimization conditions are triggered, mark as grey
-        if status == CollectionStatus::Green
+        if status == ShardStatus::Green
             && self.update_handler.lock().await.has_pending_optimizations()
+            && optimizer_status == OptimizersStatus::Ok
```
timvisee, line 1015: Let's swap these because `has_pending_optimizations` can be a bit expensive and even takes a lock:

```suggestion
        if status == ShardStatus::Green
            && optimizer_status == OptimizersStatus::Ok
            && self.update_handler.lock().await.has_pending_optimizations()
```

Good suggestion by the way, to add the optimizer status check.

lib/collection/src/shards/local_shard/mod.rs
timvisee: I would actually vote to just keep `local_shard_info` without a separate function for the status. The reason is that there is quite a bit of overlap and we'll most likely fail to keep it properly in sync.

The only overhead I see in the current function is some number additions. Those are negledable.

timvisee: On second thought, I have just discussed this with @generall: 

Let's do separate functions. But the `local_shard_info` we already have will simply call `local_shard_status` to get the color. That way we don't have duplication. There is a bit of overhead, but it's good enough here.

lib/collection/src/operations/types.rs
ffuugoo: > It's still better semantics to call it ShardStatus...

IMO, yes, `ShardStatus` is better.

lib/collection/src/operations/types.rs
```diff
@@ -71,6 +71,33 @@ pub enum CollectionStatus {
...
+    // Shard if completely ready for requests
```
ffuugoo, line 79:
```suggestion
    // Shard is completely ready for requests
```

lib/collection/src/shards/shard.rs
ffuugoo: Looks kinda ugly... 😬

I'd add a comment explaining why is it structured like this (e.g., why not just move `local_shard_info` inside `get_telemetry_data`).

ffuugoo (approved):
LGTM with a very minor comments. Good job, @KShivendu! 👍

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -969,15 +970,48 @@ impl LocalShard {
...
+            let segments = self.segments().read();
+            for (_, segment) in segments.iter() {
+                let segment_info = segment.get().read().info();
+                if segment_info.segment_type == SegmentType::Special {
+                    status = ShardStatus::Yellow;
+                }
+
+                if !segments.failed_operation.is_empty() || segments.optimizer_errors.is_some() {
+                    status = ShardStatus::Red;
+                }
```
timvisee, line 987: We probably need to break here to not overwrite a red status with something else after.

```suggestion
                if !segments.failed_operation.is_empty() || segments.optimizer_errors.is_some() {
                    status = ShardStatus::Red;
                    break;
                }
```

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -969,15 +970,48 @@ impl LocalShard {
...
+        }
+
+        // If still green while optimization conditions are triggered, mark as grey
+        if status == ShardStatus::Green
+            && optimizer_status == OptimizersStatus::Ok
+            && self.update_handler.lock().await.has_pending_optimizations()
+        {
+            status = ShardStatus::Grey;
+            optimizer_status =
+                OptimizersStatus::Error("optimizations pending, awaiting update operation".into());
```
timvisee, line 1002: This was only here temporary while we couldn't use the grey status. I suggest to remove this now.

```suggestion
```

timvisee (changes requested):

lib/collection/src/shards/local_shard/mod.rs
KShivendu: Looks like segment is a list. I did some refactoring

lib/collection/src/shards/local_shard/mod.rs
KShivendu: done

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -976,30 +976,34 @@ impl LocalShard {
...
+            } else {
+                let has_special_segments = segments
+                    .iter()
+                    .map(|(_, segment)| segment.get().read().info().segment_type)
+                    .any(segment_type == SegmentType::Special);
+
+                // Special segment means it's a proxy segment and is being optimized, mark as yellow
+                if has_special_segments {
+                    status = ShardStatus::Yellow;
+                }
```
timvisee, line 995: Now that I think of it, this would also trigger if we'd just create a snapshot. Snapshotting creates temporary proxy segments.

Let's keep it like this for now though.

timvisee (approved):

lib/collection/src/shards/local_shard/mod.rs
KShivendu: added a comment

lib/collection/src/operations/types.rs
```diff
@@ -56,31 +56,31 @@ use crate::shards::transfer::ShardTransferMethod;
...
+/// `Green` - all good. `Yellow` - optimization is running, 'Grey' - optimizations are possible but not triggered, `Red` - some operations failed and was not recovered
 #[derive(Debug, Serialize, JsonSchema, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
 #[serde(rename_all = "snake_case")]
 pub enum CollectionStatus {
     // Collection is completely ready for requests
     Green,
     // Collection is available, but some segments might be under optimization
     Yellow,
-    // Collection is available, but some segments are pending optimization
+    // Collection is available, but some segments can be optimized still optimizations aren't triggered
```
timvisee, line 67: Nit: I liked the previous better

PR #4941, opened 2024-08-22
fix: Non-blocking payload index building

### LINKED ISSUE #4934: Non-blocking payload index construction
I have quite a large collection: 60M points and rising. Sometimes it is necessary to add new payload indices to the collection or update existing ones. Currently, each such request effectively freezes the database (both read and write queries are blocked) which causes significant downtime.

Payload index modification happens completely in the background. It should be possible to upsert and read data from the DB while the index is being constructed. Ideally, there are also performance benefits from an incomplete index.

I haven't tested it but would it help to have several shards in the collection? Currently, there is only one shard. Other workarounds would also be appreciated.

coszio:
Fixes #4934 

Currently, when we build a payload index, we hold a write lock the entire time the index is being built. This might be acceptable for small collections, but on bigger ones this means that searches will be completely interrupted (and even make them timeout) during building.

With these changes I've separated the building and applying sections, so that we only need a read lock when building, and upgrades to a write lock when applying it to the segment.

I've tested this manually with bfb + dashboard, and now searches keep responding normally during index building, with only a really small hiccup at the end.

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -14,7 +14,7 @@ use segment::data_types::order_by::OrderValue;
 use segment::data_types::query_context::{QueryContext, SegmentQueryContext};
 use segment::data_types::vectors::{QueryVector, Vector};
 use segment::entry::entry_point::SegmentEntry;
-use segment::index::field_index::CardinalityEstimation;
+use segment::index::field_index::{CardinalityEstimation, FieldIndex};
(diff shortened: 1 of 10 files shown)


coszio:
Note: Since these changes, I've had to add `?wait=true` to some of our tests, because now it truly happens on the background, not only for responding quickly. 

Our tests worked because the index building locked the server while it finished, so it behaved similarly as if `wait` was always `true`.

generall (approved):
Nice!

timvisee:
Sweet! Thanks for implementing it so quickly.

PR #4946, opened 2024-08-23
Integer and UUID facets

coszio:
Enables faceting with the other existing map indexes: integer and uuid

Note: I decided to not add a special case for UUID in REST, and just use Keyword variant as output. This should be transparent when serializing, and we do not consume it in any form.

Note2: I edited the trait and implementations of `MmapValue` to allow references of `i64` and `u128`

### DIFF
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -603,6 +603,7 @@ message FacetCounts {
 message FacetValue {
     oneof variant {
         string string_value = 1; // String value from the facet
+        int64 integer_value = 2; // Integer value from the facet
     }
 }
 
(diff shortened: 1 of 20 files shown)


lib/api/src/grpc/proto/points.proto
```diff
@@ -603,6 +603,8 @@ message FacetCounts {
 message FacetValue {
     oneof variant {
         string string_value = 1; // String value from the facet
+        int64 integer_value = 2; // Integer value from the facet
+        string uuid_value = 3; // UUID value from the facet
```
generall, line 607: Why uuid is not just a `string_value`?

coszio: Because this is the same structure we use for internal service, we need this to be able to deserialize it into uuid again. We could make a separate internal structure though

generall: Yes, let's make a separate structures for consistency with REST

timvisee (commented):
:+1:

timvisee (approved):
Better. Thanks for the change!

Issue #4949, opened 2024-08-26
Collection with multiple vectors creates dummy embeddings to all vectors


keviveks:
We have defined multiple named vector in our collections to try out with different embedding model.
Usually we create a point in qdrant with payload and one vector embedding, recently we are observing some of the points filled with vector ([1,...]) placeholder embeddings to other vectors and these points are not efficient in search.

## Current Behavior
<img width="156" alt="Screenshot 2024-08-21 at 3 48 57 PM" src="https://github.com/user-attachments/assets/dc81cee9-4df2-4995-af04-e61826be6a60">
<img width="156" alt="Screenshot 2024-08-21 at 3 49 12 PM" src="https://github.com/user-attachments/assets/0d33c8aa-5ab2-4c30-9270-da0acc7b9b26">
<img width="150" alt="Screenshot 2024-08-21 at 3 49 20 PM" src="https://github.com/user-attachments/assets/673004f3-55a6-4846-b2c1-6018007d6364">

## Steps to Reproduce
1. Create collection with multiple named vector
```
{
  "vectors": {
    "embedding-1024": {
      "size": 1024,
      "distance": "Cosine",
      "on_disk": true
    },
    "embedding-1536": {
      "size": 1536,
      "distance": "Cosine",
      "on_disk": true
    },
    "embedding-3072": {
      "size": 3072,
      "distance": "Cosine",
      "on_disk": true
    },
    "embedding-4096": {
      "size": 4096,
      "distance": "Cosine",
      "on_disk": true
    },
    "embedding-768": {
      "size": 768,
      "distance": "Cosine",
      "on_disk": true
    }
  },
  "shard_number": 1,
  "replication_factor": 2,
  "write_consistency_factor": 1,
  "on_disk_payload": true
}
```
2. Upsert point with one vector
```
PUT collections/<collection_name>/points 
{
  "wait": true,
  "points": [
    {
      "id": 1,
      "vector": {
        "embedding-1536": [
[...]

PR #4952, opened 2024-08-26
Fix point deletions on mmap segment optimization

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/4949>.

Fix `memmap_dense_vector_storage` incorrectly taking over vector deletions on segment optimization. This resulted in deleted vectors suddenly appearing again once segments get optimized.

This problem has existed for a long time but only came to light recently. Our input data on `update_from` may now contain points for multiple segments, in which case the previous logic breaks.

All other storage types implemented it correctly. This implementation has now been updated to match the others, fixing the problem.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
+++ b/lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
@@ -175,7 +175,7 @@ impl<T: PrimitiveVectorElement> VectorStorage for MemmapDenseVectorStorage<T> {
 
     fn update_from<'a>(
         &mut self,
-        other_ids: &'a mut impl Iterator<Item = (PointOffsetType, CowVector<'a>, bool)>,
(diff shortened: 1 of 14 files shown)


lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
```diff
@@ -191,7 +191,7 @@ impl<T: PrimitiveVectorElement> VectorStorage for MemmapDenseVectorStorage<T> {
         // Extend vectors file, write other vectors into it
```
timvisee, line 191: Here is the important bit. We don't use the provided point offset, but regenerate our own offset sequence.

JojiiOfficial (approved):

generall (approved):

generall (approved):

PR #4961, opened 2024-08-26
Implement better handling of UUID index

generall:
- Implement MatchAny and Except


lib/segment/src/index/field_index/map_index/mod.rs
```diff
@@ -620,36 +646,140 @@ impl PayloadFieldIndex for MapIndex<UuidIntType> {
...
+                            .flat_map(|uuid| self.get_iterator(&uuid).copied())
+                            .unique(),
+                    ))
+                }
+                AnyVariants::Integers(integers) => {
+                    if integers.is_empty() {
+                        Some(Box::new(iter::empty()))
+                    } else {
+                        None
+                    }
```
generall, line 679: this and similar cases reasoning: we can't guaratnee which type will be used for de-serialization of empty list, as it is ambiguous. Having check for empty condition is a fail-safe

coszio (approved):

PR #4964, opened 2024-08-27
Packed geohash

IvanPleshkov:
This PR replaces `GeoHash` implementation from `SmolStr` into `u64`. It's possible because each string char can be one of 32 values (5bits), max chars 12 (5*12=60bits) and size 4 bit.

### DIFF
--- a/lib/segment/src/index/field_index/geo_hash.rs
+++ b/lib/segment/src/index/field_index/geo_hash.rs
@@ -1,4 +1,4 @@
-use std::ops::Range;
+use std::ops::{Index, Range};
 
 use geo::algorithm::haversine_distance::HaversineDistance;
 use geo::{Coord, Intersects, LineString, Point, Polygon};
@@ -9,7 +9,20 @@ use smol_str::SmolStr;
 use crate::common::operation_error::{OperationError, OperationResult};
 use crate::types::{GeoBoundingBox, GeoPoint, GeoPolygon, GeoRadius};
 
-pub type GeoHash = SmolStr;
+#[derive(Default, Clone, Copy, Debug, PartialEq, Hash, Ord, PartialOrd, Eq)]
+pub struct GeoHash {
+    packed: u64,
+}
+
+// code from geohash crate
+// the alphabet for the base32 encoding used in geohashing
+#[rustfmt::skip]
+const BASE32_CODES: [char; 32] = [
+    '0', '1', '2', '3', '4', '5', '6', '7',
+    '8', '9', 'b', 'c', 'd', 'e', 'f', 'g',
+    'h', 'j', 'k', 'm', 'n', 'p', 'q', 'r',
+    's', 't', 'u', 'v', 'w', 'x', 'y', 'z',
+];
 
 /// Max size of geo-hash used for indexing. size=12 is about 6cm2
 const GEOHASH_MAX_LENGTH: usize = 12;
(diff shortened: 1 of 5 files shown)


lib/segment/src/index/field_index/geo_hash.rs
```diff
@@ -18,6 +31,107 @@ const LON_RANGE: Range<f64> = -180.0..180.0;
...
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    pub fn len(&self) -> usize {
+        (self.packed & 0b1111) as usize
+    }
+
+    pub fn cut(&self, new_len: usize) -> Self {
```
timvisee, line 104: This only keeps the most significant bits, and removes the least significant ones.

In that case I'd prefer the name `truncate`. In my opinion that better clarifies this removes the end.

lib/segment/src/index/field_index/geo_hash.rs
```diff
@@ -67,22 +181,27 @@ fn sphere_lat(lat: f64) -> f64 {
...
+    let (coord, lon_err, lat_err) = decode(hash_str.as_str())?;
     let (dlat, dlng) = direction.to_tuple();
     let lon = sphere_lon(coord.x + 2f64 * lon_err.abs() * dlng);
     let lat = sphere_lat(coord.y + 2f64 * lat_err.abs() * dlat);
 
     let neighbor_coord = Coord { x: lon, y: lat };
-    encode(neighbor_coord, hash_str.len()).map(Into::into)
+    encode(neighbor_coord, hash_str.len())
+        .map(Into::<SmolStr>::into)
+        .map(Into::into)
```
timvisee, line 194: Let's also implement `impl From<String> for GeoHash` so we don't have to convert into an intermediate `SmolStr`

lib/segment/src/index/field_index/geo_hash.rs
```diff
@@ -67,22 +181,27 @@ fn sphere_lat(lat: f64) -> f64 {
...
-    encode((lon, lat).into(), GEOHASH_MAX_LENGTH).map(Into::into)
+    encode((lon, lat).into(), GEOHASH_MAX_LENGTH)
+        .map(Into::<SmolStr>::into)
+        .map(Into::into)
 }
 
-pub fn geo_hash_to_box(geo_hash: &GeoHash) -> GeoBoundingBox {
-    let rectangle = decode_bbox(geo_hash).unwrap();
+pub fn geo_hash_to_box(geo_hash: GeoHash) -> GeoBoundingBox {
+    let rectangle = decode_bbox(SmolStr::from(geo_hash).as_str()).unwrap();
```
timvisee, line 204: And we might also benefit from a direct to `String` conversion, versus an intermediate `SmolStr`

lib/segment/src/index/field_index/geo_hash.rs
xzfc: At this point we could just re-implement the `encode` method and avoid using strings (especially heap-allocated ones).

It's not that big, and it uses similar `u64` packing internally already.

lib/segment/src/index/field_index/geo_hash.rs
```diff
@@ -18,6 +31,107 @@ const LON_RANGE: Range<f64> = -180.0..180.0;
...
+    pub fn len(&self) -> usize {
+        (self.packed & 0b1111) as usize
+    }
+
+    pub fn cut(&self, new_len: usize) -> Self {
+        assert!(new_len <= self.len());
+        let mut packed = self.packed;
+        for i in new_len..self.len() {
+            packed &= !(0b11111 << Self::shift_value(i));
+        }
```
xzfc, line 109: Performance nit: It could be done without a loop:
`packed = !0x1111 << 5 * (self.len() - new_len);` (draft)

lib/segment/src/index/field_index/geo_index/mod.rs
```diff
@@ -103,8 +104,9 @@ impl GeoMapIndex {
         format!("{field}_geo")
     }
 
-    fn encode_db_key(value: &str, idx: PointOffsetType) -> String {
-        format!("{value}/{idx}")
+    fn encode_db_key(value: GeoHash, idx: PointOffsetType) -> String {
```
xzfc, line 107: Nit: this method could return `-> SmolStr` to avoid allocation.

agourlay:
I would love to see performance numbers here to assess if the complexity is worth it :pray:

lib/segment/src/index/field_index/geo_hash.rs
```diff
@@ -18,6 +31,107 @@ const LON_RANGE: Range<f64> = -180.0..180.0;
...
+            } else {
+                true
+            }
+        } else {
+            false
+        }
+    }
+
+    fn shift_value(i: usize) -> usize {
+        // first 4 bits is size, then 5 bits per character in reverse order (for lexicographical order)
```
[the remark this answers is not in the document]
IvanPleshkov: Thanks for this great example! I added doc comment to the `GeoHash` struct

lib/segment/src/index/field_index/geo_hash.rs
```diff
@@ -18,6 +31,107 @@ const LON_RANGE: Range<f64> = -180.0..180.0;
...
+        let mut packed = self.packed;
+        for i in new_len..self.len() {
+            packed &= !(0b11111 << Self::shift_value(i));
+        }
+        packed &= !(0b1111);
+        packed |= new_len as u64;
+        Self { packed }
+    }
+
+    pub fn starts_with(&self, other: GeoHash) -> bool {
```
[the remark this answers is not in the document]
IvanPleshkov: Done! Thanks for mention that, there are no any explicit tests for this function

lib/segment/src/index/field_index/geo_index/mod.rs
xzfc: The `format!` could be replaced with `write!`.

```rust
use std::fmt::Write as _;
let mut b = smol_str::SmolStrBuilder::new();
write!(b, "{value}/{idx}").unwrap();
b.finish()
```

Using unwrap here is fine (aka no worse than using `format!`), see
- [`format!` doc on when it panics](https://doc.rust-lang.org/std/macro.format.html#panics).
- [`impl fmt::Write for SmolStrBuilder`](https://docs.rs/smol_str/0.3.1/src/smol_str/lib.rs.html#801-807) never returns an `Error`.

xzfc: Ah, even better, there is [`format_smolstr!`](https://docs.rs/smol_str/latest/smol_str/macro.format_smolstr.html) macro.

lib/segment/src/index/field_index/geo_hash.rs
```diff
@@ -18,6 +31,107 @@ const LON_RANGE: Range<f64> = -180.0..180.0;
...
+impl From<GeoHash> for SmolStr {
+    fn from(hash: GeoHash) -> Self {
+        hash.iter().collect()
+    }
+}
+
+pub struct GeoHashIterator {
+    hash: GeoHash,
+    index: usize,
+    len: usize,
```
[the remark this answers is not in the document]
IvanPleshkov: Did another implementation based on your idea to use only one field
```
fn next(&mut self) -> Option<Self::Item> {
    let len = self.packed_chars & 0b1111;
    if len > 0 {
        // take first character from the packed value
        let char_index = (self.packed_chars >> 59) & 0b11111;

        // shift packed value to the left to get the next character
        self.packed_chars = (self.packed_chars << 5) | (len - 1);

        // get character from the base32 alphabet
        Some(BASE32_CODES[char_index as usize])
    } else {
        None
    }
}
```

lib/segment/src/index/field_index/geo_index/mod.rs
IvanPleshkov: Great! Thanks for googling it for me, fixed!

lib/segment/src/index/field_index/geo_hash.rs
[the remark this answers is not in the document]
xzfc: Updated the test to be more extensive (6*6 variants, including negatives).

xzfc (approved):

generall (approved):

PR #4968, opened 2024-08-27
Check that peer still exist, when applying finish transfer operation

ffuugoo:
This PR fixes a possible bug, when peer is removed right before shard transfer **to** removed peer is finished, which may create an "orphan" shard: a shard, that exists in shard replica state, but the **peer** this shard should belong to is already removed from cluster.

The fix is rather simple: when applying `ShardTransfer::Finish` conensus message, only mark replica as `Active`, if *peer* still exists in the cluster.

This PR also restructures/simplifies `Collection::finish_shard_transfer`, cause it was too hard to work with before.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::future::Future;
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -182,6 +183,7 @@ impl Collection {
     pub async fn finish_shard_transfer(
         &self,
         transfer: ShardTransfer,
+        all_peers: &HashSet<PeerId>,
         shard_holder: Option<&ShardHolder>,
     ) -> CollectionResult<()> {
         let transfer_result = self
@@ -209,9 +211,14 @@ impl Collection {
             // Normally we promote the shard to become active, in case of resharding we do not.
             // For resharding we have multiple transfers in sequence, during which the shard should
             // remain in the resharding state. Once all are done, the shard is manually promoted to active.
-            let activate_shard = transfer
+            let dst_peer_exists = all_peers.contains(&transfer.to);
+
+            let is_resharding_transfer = transfer
                 .method
-                .map_or(true, |method| !method.is_resharding());
+                .map_or(false, |method| method.is_resharding());
+
+            let activate_shard = dst_peer_exists && !is_resharding_transfer;
+
             let proxy_promoted = transfer::driver::handle_transferred_shard_proxy(
                 shard_holder,
                 transfer.shard_id,
@@ -220,6 +227,7 @@ impl Collection {
                 transfer.sync,
             )
             .await?;
+
             log::debug!("proxy_promoted: {proxy_promoted}");
         }
 
@@ -254,6 +262,13 @@ impl Collection {
             } else {
                 ReplicaState::Active
             };
+
+            let state = if all_peers.contains(&transfer.to) {
+                Some(state)
+            } else {
+                None
+            };
+
             let remote_shard_rerouted = transfer::driver::change_remote_shard_route(
                 shard_holder,
                 transfer.shard_id,
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -1,4 +1,3 @@
-use std::collections::HashSet;
 use std::future::Future;
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -181,7 +180,6 @@ impl Collection {
     pub async fn finish_shard_transfer(
         &self,
         transfer: ShardTransfer,
-        all_peers: &HashSet<PeerId>,
         shard_holder: Option<&ShardHolder>,
     ) -> CollectionResult<()> {
         let transfer_result = self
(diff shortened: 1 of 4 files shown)


lib/collection/src/collection/shard_transfer.rs
```diff
@@ -182,6 +183,7 @@ impl Collection {
     pub async fn finish_shard_transfer(
         &self,
         transfer: ShardTransfer,
+        all_peers: &HashSet<PeerId>,
         shard_holder: Option<&ShardHolder>,
     ) -> CollectionResult<()> {
```
timvisee, line 188: Here, can't we just check whether the peer ID is in the shard holder?

Because it's just when finishing a shard transfer, I think that should be fine. We don't expect to create shards on new peers here.

ffuugoo: Hmmmm.... this should work as well, I think.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -202,72 +198,56 @@ impl Collection {
...
-                transfer.from,
-                transfer.to,
-                state,
-                transfer.sync,
-            )
-            .await?;
-            log::debug!("remote_shard_rerouted: {remote_shard_rerouted}");
+                // If transfer is *not* sync (and *not* resharding), we *remove* source replica
+
+                if transfer.from == self.this_peer_id {
```
generall, line 220: Activation of target now depends on `replica_set.peer_state(&transfer.to).is_some()` condition. We should not remove source if target is not activated. 

Propose to handle target first and then return flag from there

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -202,72 +198,56 @@ impl Collection {
...
-        let finish_was_registered = shard_holder.register_finish_transfer(&transfer.key())?;
-        log::debug!("finish_was_registered: {finish_was_registered}");
+
+        // Handle *destination* replica
+        let dest_replica_set =
+            shard_holder.get_shard(&transfer.to_shard_id.unwrap_or(transfer.shard_id));
+
+        if let Some(replica_set) = dest_replica_set {
+            if replica_set.peer_state(&transfer.to).is_some() && !is_resharding_transfer {
+                // Promote *destination* replica/shard to `Active` if:
```
generall, line 234: shouldn't we also do

```
    if to_replica_set.this_peer_id() != new_peer_id {
        to_replica_set.add_remote(new_peer_id, state).await?;
    }
```

here?

generall (changes requested):
I like the refactoring, `finish_shard_transfer` is now much more readable.

lib/collection/src/collection/shard_transfer.rs
ffuugoo: Good point. Will do.

lib/collection/src/collection/shard_transfer.rs
ffuugoo: I've checked `ShardTransfer::Start`/`ShardTransfer::Restart` and `Resharding::Start`, and I don't think we need this. We create shards (local on receiver, and remote on all other nodes) when handling `(Re)Start` messages, and so by the time we receive `ShardTransfer::Finish`, the peer should already exist in the replica set.

ffuugoo: Maybe there is/was some special case that needs `add_remote`, but the tests are still green without it.

lib/collection/src/collection/shard_transfer.rs
ffuugoo: Done.

lib/collection/src/collection/shard_transfer.rs
generall: let's maybe do it anyway, just for a good measure. I assume there are no real action will happen if remote already exist?

ffuugoo: Meh. Sure.

ffuugoo: Last question:

Original impl effectively did:

```rust
let state = if is_resharding {
    ReplicaState::Resharding
} else {
    ReplicaState::Active
};

replica_set.add_remote(transfer.to, state).await?;
```

Do we want to recreate this exactly, it's fine to just do:

```rust
if !is_resharding {
    replica_set.add_remote(transfer.to, ReplicaState::Active).await?;
}
```

(If too deep/you don't have an answer right away, just answer "let's recreate exactly".)

ffuugoo: @generall 👆

generall: I don't have answer right away. I assume Resharding state is set by some other operations before transfer.

timvisee: I believe I put that there. It's so the new shard stays in resharding state. Otherwise the first transfer that finishes activates it, which we don't want.

timvisee (approved):

lib/collection/src/collection/shard_transfer.rs
ffuugoo: Tweaked a bit differently, we do `set_state`/`add_remote` always now, but at least it's still readable.

I still think this is worse than initial refactoring, though. I'd prefer this to only do absolutely minimal required changes, e.g., only _set_ (not add) replicas to `Active` in case of non-resharding transfer, and _nothing_ more. Less change of re-introducing removed replicas in the future, if/when we tweak some pre-condition checks or something.

generall (approved):

generall:
let's merge after release of 1.11.3 patch

PR #4969, opened 2024-08-27
Facets: add consistency param to openapi

coszio:
Adds the missing read consistency query parameter to the openapi spec for the facets endpoint

### DIFF
--- a/lib/segment/src/index/field_index/field_index_base.rs
+++ b/lib/segment/src/index/field_index/field_index_base.rs
@@ -373,8 +373,8 @@ impl FieldIndex {
         match self {
             FieldIndex::KeywordIndex(index) => Some(FacetIndex::Keyword(index)),
             FieldIndex::IntMapIndex(index) => Some(FacetIndex::Int(index)),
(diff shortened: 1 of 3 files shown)


timvisee (approved):

lib/segment/src/index/field_index/field_index_base.rs
```diff
@@ -373,8 +373,8 @@ impl FieldIndex {
         match self {
             FieldIndex::KeywordIndex(index) => Some(FacetIndex::Keyword(index)),
             FieldIndex::IntMapIndex(index) => Some(FacetIndex::Int(index)),
-            | FieldIndex::UuidMapIndex(_) // TODO(facets): use uuid index too
-            | FieldIndex::UuidIndex(_)
+            FieldIndex::UuidMapIndex(index) => Some(FacetIndex::Uuid(index)),
```
agourlay, line 376: that's unrelated but on purpose?

coszio: It is related only in the "Things I forgot to do for facets" thematic

PR #4971, opened 2024-08-28
Add tests on uuid payload index queries with filters

tellet-q:
* Parameterize existing tests with different filters 
* Also add tests for query batches with and without filters.

### All Submissions:

### DIFF
--- a/tests/openapi/test_query_full.py
+++ b/tests/openapi/test_query_full.py
@@ -255,7 +255,7 @@ def test_uuid_index_filtered_query():
             method="POST",
             path_params={"collection_name": collection_name},
             body={
-                "vector": [0.1, 0.2, 0.3, 0.4],
+                "vector": {"vector": [0.1, 0.2, 0.3, 0.4], "name": "dense-image"},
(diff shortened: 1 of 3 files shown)


tests/openapi/test_query_full.py
generall, line 1: AFAIK those tests ensure consistency of results between old search API and new query API.
But they don't validate if the result itself is correct.

Since filter index shares implementation in both - old and new requests, error in the index will also be same.

let's create a test which would check against some reference output instead (or we can create test which validates that keyword and UUID indexes behave same), but I think it should be a dedicated file

generall (changes requested):

tests/openapi/test_query_full.py
tellet-q: I've added a dedicated test file that compares results between keyword index and uuid index. Right now the results for match-except filter differ between those two.

tellet-q:
Merge after https://github.com/qdrant/qdrant/pull/4977

generall (approved):

PR #4985, opened 2024-08-29
Check that peer still exist, when applying set shard replica set operation

ffuugoo:
Similar to #4968, this PR fixes a potential bug, when peer is removed right before state of a replica on removed peer is changes, which may create an "orphan" shard.

Fix is conceptually the same as in #4968: check that peer still exists in the cluster before applying `SetShardReplicaSet` operation.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -372,16 +372,30 @@ impl Collection {
         let current_state = replica_set.peer_state(&peer_id);
 
         // Validation:
-        // 0. Check that `from_state` matches current state
+        //
+        // 1. Check that peer exists in the cluster (peer might *not* exist, if it was removed from
+        //    the cluster right before `SetShardReplicaSet` was proposed)
+        let peer_exists = self
+            .channel_service
+            .id_to_address
+            .read()
+            .contains_key(&peer_id);
 
+        if !peer_exists {
+            return Err(CollectionError::bad_input(format!(
+                "Can't set replica {peer_id}:{shard_id} state to {state:?}, \
+                 because peer {peer_id} is not part of the cluster"
+            )));
+        }
+
+        // 2. Check that `from_state` matches current state
         if from_state.is_some() && current_state != from_state {
             return Err(CollectionError::bad_input(format!(
                 "Replica {peer_id} of shard {shard_id} has state {current_state:?}, but expected {from_state:?}"
             )));
         }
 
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection/mod.rs
```diff
@@ -372,16 +372,30 @@ impl Collection {
...
+        // 1. Check that peer exists in the cluster (peer might *not* exist, if it was removed from
+        //    the cluster right before `SetShardReplicaSet` was proposed)
+        let peer_exists = self
+            .channel_service
+            .id_to_address
+            .read()
+            .contains_key(&peer_id);
 
+        if !peer_exists {
+            return Err(CollectionError::bad_input(format!(
```
generall, line 385: We might still want to change replica state if replica already exists.
The only thing we don't want to do it to create new replica if none registered yet.

This might be only needed if something goes wrong, so I propose to include debug assert in this case

ffuugoo: > The only thing we don't want to do it to create new replica if none registered yet.

Not sure if this is true. We might actually rely on this somewhere. What we don't want to do, is create new replica *if peer does not exist*.

ffuugoo: Also, the fact that a lot of tests exploded already seems very sus... 😂

ffuugoo: > We might actually rely on this somewhere

I mean, we literally call `replica_set.ensure_replica_with_state(...)`, not just `replica_set.set_replica_state(...)`. Makes it reasonable to assume, that when `SetShardReplicaSet` is applied, it *creates* replica, if it does not exist yet.

ffuugoo: I'll change the check to `!peer_exists && !replica_exists` though. This way we only prevent (re)creating *new* replica, if peer is gone, but allow to change state of *existing* replica (even if peer is gone).

generall (approved):

lib/collection/src/collection/mod.rs
timvisee: I like that we still allow changing it if the replica already exists.

I wonder if we can clarify the semantics of this, and prevent calling this to spawn an unknown replica from thin air. It would require a multi-version effort though.

timvisee (approved):

PR #4987, opened 2024-08-29
[Facet] Do not include values with count 0

coszio:
Inside approximate faceting, there was a case where we would output hits with count=0, which isn't desirable and is incongruent with other cases.

This fixes those cases.

### DIFF
--- a/lib/segment/src/segment/facet.rs
+++ b/lib/segment/src/segment/facet.rs
@@ -66,7 +66,8 @@ impl Segment {
 
                 let iter = facet_index
                     .iter_filtered_counts_per_value(&context)
-                    .check_stop(|| is_stopped.load(Ordering::Relaxed));
+                    .check_stop(|| is_stopped.load(Ordering::Relaxed))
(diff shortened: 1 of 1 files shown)


agourlay:
Is this tested via the Python congruence tests?

coszio:
Yep, that's how it came up

agourlay (approved):

timvisee:
Can't we assume all are at least 1?

I mean, if we estimate a key has a count of 0. Can't we assume there is at least one occurrence of that key, because otherwise it wouldn't be listed at all.

Or is this with filtering, where it's definitely possible for there being zero occurrences even though the key exists _somewhere_.

coszio:
> Or is this with filtering, where it's definitely possible for there being zero occurrences even though the key exists somewhere.

This is the case. When filtering over the values->ids map, we can definitely have values where none of their ids match the filter

timvisee:
> This is the case. When filtering over the values->ids map, we can definitely have values where none of their ids match the filter

Thanks for confirming!

PR #4992, opened 2024-08-30
Create Telemetry client only once

agourlay:
Creating a Reqwest client is a rather expensive operation as it involves setting up a connection pool, configuring DNS resolver and setting up TLS things.

This PR proposes to create a single client for the telemetry reporter which is reused enabling to leverage connection pooling if possible.

Even if our reporting interval is high, it does not mean we should be wasteful with resources :earth_americas:


timvisee (approved):
:ok_man:

PR #4994, opened 2024-09-01
fix: OOM when uploading a large snapshot to S3

### LINKED ISSUE #4991: OOM when uploading a large snapshot to S3
## Current Behavior
The database OOMs when attempting to upload a large snapshot archive file (larger than the available RAM) to S3. The OOM happens at the uploading stage specifically (when the snapshot is already archived). 

## Possible Solution
I did some digging in code and it appears that the following lines are the cause 
https://github.com/qdrant/qdrant/blob/43fa913929b763438aa09afdeac5ecd1906dee59/lib/collection/src/operations/snapshot_storage_ops.rs#L119-L125 The `write.write(buffer)` part will copy data from the `buffer` and will create a new tokio task returning immediately. Since there is no call to `write.wait_for_capacity` - a new chunk from the file will be read into RAM immediately and so on. This will eventually cause an OOM.

Separately, there appears to be a bug in the following lines https://github.com/qdrant/qdrant/blob/43fa913929b763438aa09afdeac5ecd1906dee59/lib/collection/src/operations/snapshot_storage_ops.rs#L92-L95 I think the formula is meant to be `(file_size - 1) / MAX_PART_NUMBER + 1`.

## Context (Environment)
- Qdrant `Version: 1.11.3, build: 9fa86106`
- snapshot size - 227648366080 bytes
- RAM - 64gb (around 47gb free when running DB).

varshith257:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

fixes #4991 
/claim #4991

### DIFF
--- a/lib/collection/src/operations/snapshot_storage_ops.rs
+++ b/lib/collection/src/operations/snapshot_storage_ops.rs
@@ -2,6 +2,7 @@ use std::fs::File;
 use std::io::{BufReader, Read};
 use std::path::{Path, PathBuf};
 
+use common::cpu::CpuBudget;
 use futures::StreamExt;
 use object_store::WriteMultipart;
 use tokio::io::AsyncWriteExt;
@@ -90,7 +91,7 @@ pub async fn get_appropriate_chunk_size(local_source_path: &Path) -> CollectionR
     // check if the file size exceeds the maximum part number
     // if so, adjust the chunk size to fit the maximum part number
     if file_size > DEFAULT_CHUNK_SIZE * MAX_PART_NUMBER {
-        let chunk_size = (file_size - 1 / MAX_PART_NUMBER) + 1; // ceil(file_size / MAX_PART_NUMBER)
+        let chunk_size = ((file_size - 1) / MAX_PART_NUMBER) + 1; // ceil((file_size) / MAX_PART_NUMBER)
         return Ok(chunk_size);
     }
     Ok(DEFAULT_CHUNK_SIZE)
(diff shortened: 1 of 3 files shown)


varshith257:
This PR is ready for review and feedback. Thank you for your time and consideration.

lib/collection/src/operations/snapshot_storage_ops.rs
```diff
@@ -113,6 +114,10 @@ pub async fn multipart_upload(
     let mut reader = BufReader::new(file);
     let mut buffer = vec![0u8; chunk_size];
 
+    // Initialize CpuBudget to manage concurrency
+    let cpu_budget = CpuBudget::default();
+    let max_concurrency = cpu_budget.available_cpu_budget();
+
```
Vasniktel, line 120: I will defer to repo maintainers here but I'm curious whether it is better to use available RAM here as a constraining factor instead of the number of available CPUs.

For example, a hard cap of 2GiB can be put on the uploading process (or alternatively, compute this threshold based on the available RAM) and then infer the concurrency limit based on the chunk size. A care would have to be taken not to exceed the limit when passing a `buffer` to the `write` method:
```rust
let mut copy = std::mem::take(&mut buffer);
copy.truncate(bytes_read);
write.put(copy.into()); // 1. write.write() is sync but a worker thread is spawned internally.
write.wait_for_capacity(...).await?;
buffer.resize(chunk_size, 0u8);
```

varshith257: Makes sense!

varshith257: I am thinking of @generall /  @timvisee review on the above suggestion. But still the `wait_for_capacity`  expects for max concurrency param.

generall (approved):
Hey @varshith257, sorry for the late review. Changes look fine to me. Regarding limiting memory - it might be a second step. Current version already looks better then previous.

PR #5000, opened 2024-09-02
Track number of points optimized and expose in telemetry

KShivendu:
Expose number of points optimized in the optimizations log:

```json
[
    {
        "name": "indexing",
        "segment_ids": [
            9,
            8
        ],
        "points_optimized": null,
        "status": "optimizing",
        "start_at": "2024-09-02T09:41:35.241746919Z",
        "end_at": null
    },
    {
        "name": "indexing",
        "segment_ids": [
            0,
            1
        ],
        "points_optimized": 34200, // ----> newly added
        "status": "done",
        "start_at": "2024-09-02T09:37:54.923944282Z",
        "end_at": "2024-09-02T09:37:55.062533570Z"
    }
]
```

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/collection/src/update_handler.rs
```diff
@@ -329,7 +345,12 @@ impl UpdateHandler {
                             ) {
                                 // Perform some actions when optimization if finished
                                 Ok(result) => {
-                                    tracker_handle.update(TrackerStatus::Done);
+                                    if result == false {
+                                        num_points_optimized = 0;
+                                    }
+
+                                    tracker_handle
+                                        .update(TrackerStatus::Done, Some(num_points_optimized));
```
KShivendu, line 353: Alternatively, we can just return num points optimized from the `optimize()` function instead of a bool.

KShivendu: Did this.

timvisee (commented):
I don't think this is what we need in cluster manager.

This exposes the number of optimized points in each optimization separately.

I believe we just want to know the total number of points optimized across all optimizations.

In CM we want to be able to detect whether we've optimized 100% of the point count in that ~collection~ shard. That's where this global total comes in handy. It means that if the optimized count is equal or higher to the point count, then we can assume it is optimized enough.

I would suggest to not expose this number here, as we don't have a clear use case for it now. But to count it globally and expose it as part of the shard info.

KShivendu:
> we've optimized 100% of the point count in that collection

I think you meant shard here (instead of collection). But got your point.

KShivendu:
I don't like the name `points_indexed_once`. But I couldn't think of something better that someone wouldn't confuse with `num_indexed_points`

timvisee:
> I don't like the name `points_indexed_once`. But I couldn't think of something better that someone wouldn't be confused with `num_indexed_points`. Lmk if you have any better ideas.

`total_indexed_points` or `indexed_points_sum`?

lib/collection/src/shards/telemetry.rs
```diff
@@ -30,6 +30,7 @@ pub struct RemoteShardTelemetry {
 #[derive(Serialize, Clone, Debug, JsonSchema)]
 pub struct LocalShardTelemetry {
     pub variant_name: Option<String>,
+    pub points_indexed_once: usize,
```
timvisee, line 33:
```suggestion
    /// Total number of indexed points since last start.
    pub total_indexed_points: usize,
```

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -541,9 +541,14 @@ pub trait SegmentOptimizer {
                 .iter()
                 .all(|s| matches!(s, LockedSegment::Original(_)));
 
+        let num_points_to_optimize = optimizing_segments
```
timvisee, line 544: Here we count the number of points from the original segments. I wonder if it may be better to count the number of points in the final optimized segment. It would more closely reflect the number of points actually being optimized. Wdyt?

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -86,6 +86,7 @@ pub struct LocalShard {
     pub(super) path: PathBuf,
     pub(super) optimizers: Arc<Vec<Arc<Optimizer>>>,
     pub(super) optimizers_log: Arc<ParkingMutex<TrackerLog>>,
+    pub(super) points_indexed_once: Arc<ParkingMutex<usize>>,
```
timvisee, line 89: Let's do an `AtomicUsize` here. A mutex is a bit overkill.

lib/collection/src/shards/telemetry.rs
```diff
@@ -55,6 +56,7 @@ impl Anonymize for LocalShardTelemetry {
     fn anonymize(&self) -> Self {
         LocalShardTelemetry {
             variant_name: self.variant_name.clone(),
+            points_indexed_once: 0,
```
timvisee, line 59: Can we use `anonymize()` here?

```suggestion
            points_indexed_once: self.points_indexed_once.anonymize(),
```

lib/collection/src/tests/mod.rs
```diff
@@ -64,10 +64,12 @@ async fn test_optimization_process() {
     let optimizers = Arc::new(vec![merge_optimizer, indexing_optimizer]);
 
     let optimizers_log = Arc::new(Mutex::new(Default::default()));
+    let points_indexed_once = Arc::new(Mutex::new(0_usize));
```
timvisee, line 67:
```suggestion
    let points_indexed_once = Arc::new(Mutex::new(0));
```

lib/collection/src/update_handler.rs
```diff
@@ -328,7 +335,11 @@ impl UpdateHandler {
...
                                 // Perform some actions when optimization if finished
-                                Ok(result) => {
+                                Ok(num_points_optimized) => {
+                                    let mut points_indexed_in_shard = points_indexed_once.lock();
+                                    *points_indexed_in_shard += num_points_optimized;
+
+                                    let result = num_points_optimized > 0;
                                     tracker_handle.update(TrackerStatus::Done);
                                     callback(result);
                                     result
```
timvisee, line 345: Let's give `result` a name here

```suggestion
                                Ok(num_points_optimized) => {
                                    *points_indexed_once.lock() += num_points_optimized;
                                    let is_optimized = num_points_optimized > 0;
                                    tracker_handle.update(TrackerStatus::Done);
                                    callback(is_optimized);
                                    is_optimized
```

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
KShivendu: > It would more closely reflect the number of points actually being optimized. Wdyt?

result of segment optimization is binary, right? i.e. either all points of a segment are optimized or none of them are optimized.

timvisee: No. Different segments having the same point ID will now be merged into one. That means the number of points in the final segment can be a bit less.

lib/collection/src/shards/telemetry.rs
KShivendu: Done.

docs/redoc/master/openapi.json
```diff
@@ -10232,6 +10233,11 @@
               }
             ]
           },
+          "points_indexed_once": {
+            "type": "integer",
+            "format": "uint",
+            "minimum": 0
+          },
```
timvisee, line 10240: This requires updating

lib/collection/src/update_handler.rs
```diff
@@ -328,10 +335,16 @@ impl UpdateHandler {
                                 stopped,
                             ) {
                                 // Perform some actions when optimization if finished
-                                Ok(result) => {
+                                Ok(num_points_optimized) => {
+                                    // FIXME: Decide on ordering
+                                    // ToDo: Only increment if it's an indexing optimization
+                                    total_indexed_points
+                                        .fetch_add(num_points_optimized, Ordering::Relaxed);
```
timvisee, line 342: To be more specific: all optimization variants can index. So we're specifically interested in whether the segment has an HNSW index or not. IIRC we can distinguish this by `Indexed` segment type.

lib/collection/src/update_handler.rs
```diff
@@ -336,10 +336,10 @@ impl UpdateHandler {
...
                                 // Perform some actions when optimization if finished
                                 Ok(num_points_optimized) => {
-                                    // FIXME: Decide on ordering
-                                    // ToDo: Only increment if it's an indexing optimization
-                                    total_indexed_points
-                                        .fetch_add(num_points_optimized, Ordering::Relaxed);
+                                    if optimizer.name() == "indexing" {
+                                        total_indexed_points
+                                            .fetch_add(num_points_optimized, Ordering::Relaxed);
+                                    }
```
timvisee, line 342: This is incorrect. All optimizer types may index points, so we don't want to consider the optimizer type here.

I believe we want to look at the end result instead and probe whether any points were actually indexed.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -543,7 +543,9 @@ pub trait SegmentOptimizer {
 
         if !all_segments_ok {
             // Cancel the optimization
-            return Ok(false);
+            return Err(CollectionError::Cancelled {
+                description: "Some segments are not ready for optimization".to_string(),
+            });
```
timvisee, line 548: I don't think we want to remove the boolean here. We need to propagate it through the callback in the match statement, and I don't think your alternative gives us exactly the same result.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -543,9 +543,7 @@ pub trait SegmentOptimizer {
 
         if !all_segments_ok {
             // Cancel the optimization
-            return Err(CollectionError::Cancelled {
-                description: "Some segments are not ready for optimization".to_string(),
-            });
+            return Ok((false, 0));
```
timvisee, line 546: This cannot be zero, because we have no segment.

```suggestion
            return Ok((false, None));
```

KShivendu: But the return type is `(bool, usize)`.

timvisee: Then we need to change it :+1:

timvisee:
@ffuugoo Could you also do a quick check of this PR?

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
ffuugoo: I guess I'm a bit late to the party, by why not change it to simply `Option<SegmentId>`? As far as I can see, it's always either `(false, None)` or `(true, Some(segment_id))`. So the meaning would be: `None` - optimization cancelled/not-performed, `Some(...)` - optimization finished successfully.

Though, I don't want to prolong this anymore, so it's fine to merge as-is. We can fix this later (or never :D).

ffuugoo (approved):

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -514,7 +514,7 @@ pub trait SegmentOptimizer {
         ids: Vec<SegmentId>,
         permit: CpuPermit,
         stopped: &AtomicBool,
-    ) -> CollectionResult<bool> {
+    ) -> CollectionResult<(bool, Option<SegmentId>)> {
```
generall, line 517: Returning bool is redundant

lib/collection/src/update_handler.rs
```diff
@@ -328,10 +335,25 @@ impl UpdateHandler {
...
                                 // Perform some actions when optimization if finished
-                                Ok(result) => {
+                                Ok((is_optimized, segment_id)) => {
+                                    if let Some(segment_id) = segment_id {
+                                        if let Some(segment) = segments.read().get(segment_id) {
+                                            // Any kind of optimization could have triggered indexing
+                                            let segment_ref = segment.get();
+                                            let segment_info = segment_ref.read();
+
+                                            if segment_info.segment_type() == SegmentType::Indexed {
```
generall, line 345: Optimized doesn't necessary creates Indexed segments

KShivendu: We only want to count number of points indexed (not optimized). Do you still see a problem?

generall: What if indexed threshold is too high and it will never create any indexed segments? Optimization will still running

generall (approved):

lib/collection/src/update_handler.rs
timvisee: Hmm, didn't think about that situation.

Then it sounds like we want to count optimized points, rather than indexed points. :face_exhaling:

generall: Do you propose to rename the param?

timvisee: > Do you propose to rename the param?

If we don't count indexed points, yes.

KShivendu: Right.

timvisee:
Now we also don't need to return the segment ID anymore, and can simply return the number of points.

That saves us some additional locking.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -707,7 +707,8 @@ pub trait SegmentOptimizer {
 
             timer.set_success(true);
 
-            if let Some(new_segment) = segments_lock.get(new_segment_id) {
+
+            if let Some(new_segment) = segments.read().get(new_segment_id) {
                 return Ok(new_segment.get().read().available_point_count());
             }
```
timvisee: We don't have to do this. We can count the numbers in `optimized_segment` before calling `swap_new`. That way we don't have to keep track of the new segment ID and don't have to do locking.

KShivendu: Right!

PR #4990, opened 2024-08-30
[Strict mode] Finish REST requests

JojiiOfficial:
### All Submissions:

* [ ] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [ ] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [ ] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #5003, opened 2024-09-02
[Strict mode] GRPC verification

JojiiOfficial:
Depends on #4990

### DIFF
--- a/src/tonic/api/points_api.rs
+++ b/src/tonic/api/points_api.rs
@@ -72,7 +72,7 @@ impl Points for PointsService {
         let access = extract_access(&mut request);
 
         delete(
-            self.dispatcher.toc(&access).clone(),
+            StrictModeCheckedProvider::new(&self.dispatcher),
             request.into_inner(),
             None,
             None,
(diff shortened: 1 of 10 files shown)


src/tonic/api/points_api.rs
```diff
@@ -71,7 +72,7 @@ impl Points for PointsService {
         let access = extract_access(&mut request);
 
         delete(
-            self.dispatcher.toc(&access).clone(),
+            StrictModeCheckedTocProvider::new(&self.dispatcher),
```
timvisee, line 75: I don't really like that we have to color all gRPC functions and pass in a `StrictModeCheckedTocProvider` or the unchecked variant.

But I don't think there is a nice alternative is there?

timvisee (approved):

src/tonic/api/points_api.rs
JojiiOfficial: The only alternative I can think of is to have each function twice, checked and unchecked. But I think the current approach with `StrictModeCheckedTocProvider` is much better.

PR #5005, opened 2024-09-02
Only omit new shard from shard selection when resharding up

timvisee:

We don't include the target shard while resharding in shard selection. We should only do this when resharding up. While resharding down, the target shard already has all relevant points from the start.

### All Submissions:

### New Feature Submissions:


generall (approved):

ffuugoo (approved):
LGTM, but I'm thinking, will it still work if we remove this filtering completely? (E.g., always send read requests to all shards, including the new shard.)

PR #5011, opened 2024-09-03
gRPC API for Distance Matrix

agourlay:
context: https://github.com/qdrant/qdrant/milestone/11

This PR adds gRPC endpoints to the new distance matrix API

### DIFF
--- a/docs/grpc/docs.md
+++ b/docs/grpc/docs.md
@@ -3821,7 +3821,7 @@ For example, if `oversampling` is 2.4 and `limit` is 100, then 240 vectors will 
 
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
-| offsets_row | [uint64](#uint64) | repeated | Row coordinates of the matrix (COO) |
+| offsets_row | [uint64](#uint64) | repeated | Row coordinates of the matrix |
 | offsets_col | [uint64](#uint64) | repeated | Column coordinates ids of the matrix |
 | scores | [float](#float) | repeated | Scores associate with coordinates |
 | ids | [PointId](#qdrant-PointId) | repeated | Ids of the points in order |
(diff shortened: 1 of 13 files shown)


lib/api/build.rs
```diff
@@ -274,7 +274,13 @@ fn configure_validation(builder: Builder) -> Builder {
             ("QueryBatchPoints.timeout", "custom(function = \"crate::grpc::validate::validate_u64_range_min_1\")"),
             ("FacetCounts.collection_name", "length(min = 1, max = 255)"),
             ("FacetCounts.key", "length(min = 1)"),
-            ("FacetCounts.timeout", "custom(function = \"crate::grpc::validate::validate_u64_range_min_1\")")
+            ("FacetCounts.filter", ""),
```
agourlay, line 277: added missing validation on `FacetCounts.filter`

docs/redoc/master/openapi.json
```diff
@@ -13011,16 +13007,18 @@
             ]
           },
           "sample": {
-            "description": "How many points to select and search within.",
+            "description": "How many points to select and search within. Default is 10.",
```
agourlay, line 13010: I decided to make `sample` and `limit` optional for the sake of consistency with other endpoints.

lib/api/build.rs
coszio: Thank you!

docs/grpc/docs.md
```diff
@@ -3807,6 +3813,110 @@ For example, if `oversampling` is 2.4 and `limit` is 100, then 240 vectors will
...
 
+<a name="qdrant-SearchMatrixOffsets"></a>
+
+### SearchMatrixOffsets
+
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| offsets_row | [uint64](#uint64) | repeated | Row coordinates of the CRS matrix |
```
coszio, line 3824: After some investigation I realized that CRS stands for something else. What we are doing with offsets is more similar to the COO representation

agourlay: Good observation :+1: 
I will get this fixed.

lib/api/src/grpc/proto/points.proto
```diff
@@ -612,6 +612,35 @@ message FacetHit {
...
+
+message SearchMatrixPair {
+  PointId a = 1; // first id of the pair
+  PointId b = 2; // second id of the pair
+  float score = 3; // score of the pair
+}
+
+message SearchMatrixOffsets {
+  repeated uint64 offsets_row = 1; // Row coordinates of the matrix
+  repeated uint64 offsets_col = 2; // Column coordinates ids of the matrix
```
coszio, line 638: How about 
```suggestion
  repeated uint64 row_indices = 1; // Row indices of the matrix
  repeated uint64 col_indices = 2; // Column indices of the matrix
```

And same for REST

PR #5013, opened 2024-09-03
Approximate facets if filter has big cardinality

coszio:
Currently each segment produces a very accurate response, even at high filter cardinalities. But this is at the cost of slow responses.

To mitigate this, this PR actually takes advantage of doing approximate faceting. When the cardinality is above ~16k points, it switches from counting exactly, to providing estimations based on the filter and the amount of points for each value.

This will keep low counts accurate, while high counts will be approximate but fast.

### DIFF
--- a/lib/common/common/src/math.rs
+++ b/lib/common/common/src/math.rs
@@ -17,7 +17,11 @@ pub fn scaled_fast_sigmoid(x: ScoreType) -> ScoreType {
     0.5 * (fast_sigmoid(x) + 1.0)
 }
 
-/// Calculates the ideal sample size for a given population size and proportion at confidence level 0.99.
+/// Sample size for an unlimited population, confidence level 0.99 and margin of error 0.01.
+pub const SAMPLE_SIZE_CL99_ME01: usize = 16641;
+
+/// Calculates the ideal sample size for a given population size and proportion
+/// at confidence level 0.99 and margin of error 0.01.
 ///
 /// If population is `None`, considers an unlimited population.
 pub fn ideal_sample_size(population_size: Option<usize>) -> usize {
@@ -47,7 +51,7 @@ pub fn ideal_sample_size(population_size: Option<usize>) -> usize {
 
 #[cfg(test)]
 mod tests {
-    use crate::math::ideal_sample_size;
+    use crate::math::{ideal_sample_size, SAMPLE_SIZE_CL99_ME01};
 
     /// Tests equivalence against https://www.calculator.net/sample-size-calculator.html
     #[test]
@@ -57,6 +61,6 @@ mod tests {
         assert_eq!(ideal_sample_size(Some(100000)), 14267);
         assert_eq!(ideal_sample_size(Some(1000000)), 16369);
         assert_eq!(ideal_sample_size(Some(10000000)), 16614);
-        assert_eq!(ideal_sample_size(None), 16641);
+        assert_eq!(ideal_sample_size(None), SAMPLE_SIZE_CL99_ME01);
     }
 }
--- a/lib/common/common/src/math.rs
+++ b/lib/common/common/src/math.rs
@@ -16,3 +16,47 @@ pub fn fast_sigmoid(x: ScoreType) -> ScoreType {
 pub fn scaled_fast_sigmoid(x: ScoreType) -> ScoreType {
     0.5 * (fast_sigmoid(x) + 1.0)
 }
+
+/// Calculates the ideal sample size for a given population size and proportion at confidence level 0.99.
+///
+/// If population is `None`, considers an unlimited population.
+pub fn ideal_sample_size(population_size: Option<usize>) -> usize {
+    // z-score for confidence level 0.99
+    const Z_SCORE: f32 = 2.58;
+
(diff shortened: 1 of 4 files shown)


coszio:
@timvisee It does make it orders of magnitude faster on large datasets

lib/segment/src/segment/facet.rs
```diff
@@ -58,16 +53,18 @@ impl Segment {
...
-                // This is more similar to a full-scan, but we won't be hashing so many times.
-                context = payload_index.struct_filtered_context(filter);
+                // Use cardinality estimations for each value to estimate the counts
+                let cardinality_ratio =
+                    filter_cardinality.exp as f32 / (self.available_point_count() + 1) as f32;
 
+                // Go over each value in the map index
                 let iter = facet_index
-                    .iter_filtered_counts_per_value(&context)
+                    .iter_counts_per_value()
```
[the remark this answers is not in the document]
agourlay: Approving since there is actually a test for this path :)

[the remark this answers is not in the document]
coszio: That is a good point @agourlay, I think it is still possible for count to be 0 here. I will handle that case.

There is actually no test exercising this path right now, only manual tests because we need a large dataset. I'll try to make one which covers this

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
```diff
@@ -198,11 +198,12 @@ impl<N: MapIndexKey + Key + ?Sized> MmapMapIndex<N> {
...
     pub fn get_count_for_value(&self, value: &N) -> Option<usize> {
-        self.value_to_points
-            .get(value)
-            .ok()
-            .flatten()
-            .map(|p| p.len())
+        self.value_to_points.get(value).ok().flatten().map(|ids| {
+            ids.iter()
+                .filter(|idx| !(self.deleted.get(**idx as usize).unwrap_or(false)))
+                .unique()
```
generall, line 204: Is there a possibility that we actually store non-unique point-id for a value?

lib/common/common/src/math.rs
```diff
@@ -16,3 +16,51 @@ pub fn fast_sigmoid(x: ScoreType) -> ScoreType {
...
 }
+
+/// Sample size for an unlimited population, confidence level 0.99 and margin of error 0.01.
+pub const SAMPLE_SIZE_CL99_ME01: usize = 16641;
+
+/// Calculates the ideal sample size for a given population size and proportion
+/// at confidence level 0.99 and margin of error 0.01.
+///
+/// If population is `None`, considers an unlimited population.
+pub fn ideal_sample_size(population_size: Option<usize>) -> usize {
```
generall, line 27: Is this function used?

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
coszio: The tests were failing without this, so I assume yes.

lib/common/common/src/math.rs
coszio: Well, not actually but I didn't want to drop the implementation in case we would need it later

lib/segment/src/segment/facet.rs
```diff
@@ -58,15 +75,18 @@ impl Segment {
...
+                    filter_cardinality.exp as f32 / self.available_point_count() as f32;
 
+                // Go over each value in the map index
                 let iter = facet_index
-                    .iter_filtered_counts_per_value(&context)
+                    .iter_counts_per_value()
                     .check_stop(|| is_stopped.load(Ordering::Relaxed))
+                    .map(move |hit| FacetHit {
+                        value: hit.value,
+                        count: (hit.count as f32 * cardinality_ratio).round() as usize,
```
generall, line 88: Do I understand this correct, that this approach assumes that filter doesn't correlate with facet value?

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
generall: not sure how this is possible, maybe we should unique before building hashmap

lib/segment/src/segment/facet.rs
coszio: Hmm good point. It does not take into consideration that the facet key may be included in the filter

lib/segment/src/segment/facet.rs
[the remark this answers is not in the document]
coszio: exercised this path in https://github.com/qdrant/qdrant/pull/5013/commits/19575bf92d743b599c4c3bb378f7fb1c37a9da39

lib/segment/src/segment/facet.rs
generall: Not only that. Facet values can be biased towards filtering. I don't think we can make independence assumption in this case.

I would rather make sampling like this:

- sample random points -> filter them -> compute facets -> repeat until sample size is either big enough, or reach max attempts

coszio: And then adjust the sampled counts with the cardinality estimation?

PR #5020, opened 2024-09-04
Abort resharding when aborting resharding shard transfer

ffuugoo:
This PR aborts resharding, when *resharding* shard transfer is aborted.

This is the most basic error handling strategy possible, but it's trivial to implement and we will revisit it later if required/desired.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -275,7 +275,6 @@ impl Collection {
     pub async fn abort_shard_transfer(
         &self,
         transfer_key: ShardTransferKey,
-        shard_holder: Option<&ShardHolder>,
     ) -> CollectionResult<()> {
         // TODO: Ensure cancel safety!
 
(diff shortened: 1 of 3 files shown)


timvisee:
I'm seeing some weird behavior while testing this with CM. Not sure if it's caused by this, or if the problem is from earlier. I'll dive deeper into it tomorrow morning.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -314,11 +325,22 @@ impl Collection {
...
+        if is_resharding_transfer {
+            let resharding_state = shard_holder.resharding_state.read().clone();
+
+            // `abort_resharding` locks `shard_holder`!
+            drop(shard_holder);
+
+            if let Some(state) = resharding_state {
+                self.abort_resharding(state.key(), false).await?;
+            }
+        }
```
timvisee, line 342: How about removing the clone and just taking the key. Then we can drop the shard holder and pass it into the abort function.

Obviously a nitpick, but I noticed this when messing with the code.

timvisee (approved):
We've identified the problem above and are fixing it separately.

With those fixes in place this works as expected during tests. :+1:

PR #5021, opened 2024-09-04
Cleanup shards when aborting scale-down resharding

ffuugoo:
Error handling for scale-up resharding is rather trivial: when resharding is aborted, we simply delete "new" shard. However, in case of scale-down resharding, we transfer points into already existing shards, and so when resharding is aborted, we need to remove all points that might have already been transferred into existing shards.

This PR executes filtered delete when aborting scale-down resharding.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -795,11 +795,13 @@ impl ShardReplicaSet {
         let mut ids = Vec::new();
 
         while let Some(current_offset) = next_offset {
+            const BATCH_SIZE: usize = 1000;
+
             let mut points = local_shard
                 .get()
                 .scroll_by(
                     Some(current_offset),
-                    1001,
+                    BATCH_SIZE + 1,
                     &false.into(),
                     &false.into(),
                     Some(&filter),
@@ -809,7 +811,7 @@ impl ShardReplicaSet {
                 )
                 .await?;
 
-            if points.len() > 100 {
+            if points.len() > BATCH_SIZE {
                 next_offset = points.pop().map(|points| points.id);
             } else {
                 next_offset = None;
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -15,6 +15,7 @@ use std::time::Duration;
 use common::cpu::CpuBudget;
 use common::types::TelemetryDetail;
 use schemars::JsonSchema;
+use segment::types::{ExtendedPointId, Filter};
 use serde::{Deserialize, Serialize};
 use tokio::runtime::Handle;
 use tokio::sync::{Mutex, RwLock};
@@ -27,8 +28,10 @@ use super::CollectionId;
 use crate::collection::payload_index_schema::PayloadIndexSchema;
 use crate::common::snapshots_manager::SnapshotStorageManager;
 use crate::config::CollectionConfig;
+use crate::operations::point_ops::{self};
 use crate::operations::shared_storage_config::SharedStorageConfig;
-use crate::operations::types::{CollectionError, CollectionResult};
+use crate::operations::types::{CollectionError, CollectionResult, UpdateResult, UpdateStatus};
+use crate::operations::CollectionUpdateOperations;
 use crate::optimizers_builder::OptimizersConfig;
 use crate::save_on_disk::SaveOnDisk;
 use crate::shards::channel_service::ChannelService;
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -777,6 +780,61 @@ impl ShardReplicaSet {
...
+            return Err(CollectionError::bad_request("TODO"));
+        };
+
+        let mut next_offset = Some(ExtendedPointId::NumId(0));
+        let mut ids = Vec::new();
+
+        while let Some(current_offset) = next_offset {
+            let mut points = local_shard
+                .get()
+                .scroll_by(
```
timvisee, line 796: I don't see parameter names in the GitHub UI, but it looks like we're using different limits: 100 and 1k

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -777,6 +780,61 @@ impl ShardReplicaSet {
...
+        }
+
+        drop(local_shard_guard);
+
+        let op =
+            CollectionUpdateOperations::PointOperation(point_ops::PointOperations::DeletePoints {
+                ids,
+            });
+
+        // TODO(resharding): Assign clock tag to the operation!? 🤔
```
timvisee, line 832: I still can't think of this being problematic

lib/collection/src/shards/replica_set/mod.rs
ffuugoo: Nice catch, thx.

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -284,6 +285,30 @@ impl ShardHolder {
...
+                }
+
+                // We only cleanup local shards
+                if !shard.is_local().await {
+                    continue;
+                }
+
+                let filter = self.hash_ring_filter(id).expect("hash ring filter");
+                let filter = Filter::new_must(Condition::CustomIdChecker(Arc::new(filter)));
+                shard.cleanup_local_shard(filter).await?;
```
timvisee, line 308: Shards may also be left in `Resharding` state, and it is probably best to 'clean that up' here too.

I imagine that if the shard is local and in resharding state, we propose to consensus to activate the shard.

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -284,6 +285,30 @@ impl ShardHolder {
...
+                if id == shard_id {
+                    continue;
+                }
+
+                // We only cleanup local shards
+                if !shard.is_local().await {
+                    continue;
+                }
+
+                let filter = self.hash_ring_filter(id).expect("hash ring filter");
```
timvisee, line 306: Shall we not expect/unwrap here?

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo: Good point. I'll take a look how to add this.

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo: I think my logic was that we check resharding state right above, and resharding can only be finished/aborted through consensus, and we apply consensus ops sequentially, so it *has to* be available.

lib/collection/src/shards/shard_holder/resharding.rs
timvisee: What will be annoying is that we don't guarantee the state is actually applied through consensus.

I think we can add a fallback for this in cluster manager itself: if there are resharding shards but there is no resharding operation, clean things up.

lib/collection/src/shards/shard_holder/resharding.rs
timvisee: Fair enough. The `is_in_progress` is good enough for me.

I'm just cautious because I never want to resolve broken consensus again :clown_face:

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo: Done.

timvisee (approved):

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -289,6 +290,38 @@ impl ShardHolder {
...
+
+                // We only cleanup local shards
+                if !shard.is_local().await {
+                    continue;
+                }
+
+                // Remove any points that might have been transferred from target shard
+                let filter = self.hash_ring_filter(id).expect("hash ring filter");
+                let filter = Filter::new_must(Condition::CustomIdChecker(Arc::new(filter)));
+                shard.cleanup_local_shard(filter).await?;
```
timvisee, line 321: Judging by the change I made [here](https://github.com/qdrant/qdrant/pull/5024#discussion_r1762580792), we might need `new_must_not` here.

timvisee (changes requested):

timvisee (approved):

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo: Yeah, I also think that we do. Done.

(And anyway, I'm working on tests for this as part of #5069, so I'll fix/tweak it later if needed.)

PR #5024, opened 2024-09-05
Refactor `local_shard_points::delete_points` API into `cleanup_shard` API

ffuugoo:
This PR refactors and simplifies `local_shard_points::delete_points` HTTP API into `cleanup_shard`, that reuses `ShardReplicaSet::delete_local_points` method added in #5021.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/shard_holder/resharding.rs
+++ b/lib/collection/src/shards/shard_holder/resharding.rs
@@ -525,7 +525,7 @@ impl ShardHolder {
         }
 
         let filter = self.hash_ring_filter(shard_id).expect("hash ring filter");
-        let filter = Filter::new_must(Condition::CustomIdChecker(Arc::new(filter)));
+        let filter = Filter::new_must_not(Condition::CustomIdChecker(Arc::new(filter)));
         shard.delete_local_points(filter).await
     }
 
--- a/lib/collection/src/shards/shard_holder/resharding.rs
+++ b/lib/collection/src/shards/shard_holder/resharding.rs
@@ -8,7 +8,7 @@ use segment::types::{Condition, CustomIdCheckerCondition as _, Filter, ShardKey}
 use super::ShardHolder;
 use crate::hash_ring::{self, HashRingRouter};
 use crate::operations::cluster_ops::ReshardingDirection;
-use crate::operations::types::{CollectionError, CollectionResult};
+use crate::operations::types::{CollectionError, CollectionResult, UpdateResult};
 use crate::operations::{point_ops, CollectionUpdateOperations};
 use crate::shards::replica_set::{ReplicaState, ShardReplicaSet};
 use crate::shards::resharding::{ReshardKey, ReshardStage, ReshardState};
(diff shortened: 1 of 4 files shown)


timvisee (approved):
While I did like a generic delete endpoint I do get t he value here, it was quite specific and this is easier to use. Above all, we don't have to specify an additional shard ID anymore.

src/actix/api/local_shard_api.rs
```diff
@@ -153,78 +139,18 @@ async fn count_points(
     .await
 }
 
-#[post("/collections/{collection}/shards/{shard}/points/delete")]
-async fn delete_points(
+#[post("/collections/{collection}/shards/{shard}/cleanup")]
```
timvisee, line 142: Since we changed this and don't strictly take a body anymore, is POST still a good candidate here? :thinking:

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -513,6 +513,22 @@ impl ShardHolder {
...
+
+        if !shard.is_local().await {
+            return Err(CollectionError::bad_shard_selection(format!(
+                "shard {shard_id} is not a local shard"
+            )))?;
+        }
+
+        let filter = self.hash_ring_filter(shard_id).expect("hash ring filter");
+        let filter = Filter::new_must(Condition::CustomIdChecker(Arc::new(filter)));
+        shard.cleanup_local_shard(filter).await
```
timvisee, line 529: This appears to clean incorrectly. Based on testing I assume it may be deleting the inverse.

Here I used clean on shard 0. Total point count was 1000:

image

timvisee (changes requested):
Requesting changes because the implementation appears to be incorrect:

lib/collection/src/shards/shard_holder/resharding.rs
timvisee: I patched this in https://github.com/qdrant/qdrant/pull/5024/commits/a008b47da9b2e35aafebd3a2e27d1a758a09be3c which seems to give the correct result.

1000 points over 3 shards after cleanup:
image

Could you check whether using `must_not` is preferred versus inverting the condition in the `CustomIdChecker`?

timvisee (approved):

timvisee:
<https://github.com/qdrant/cluster-manager/pull/125> implements using this change in cluster manager.

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo: _I think_ you're right, this should be `must_not`. Thanks!

src/actix/api/local_shard_api.rs
ffuugoo: Yeah, I think it is. It's definitely not `GET`, and `PUT`/`PATCH`/`DELETE` does not feel better as well.

timvisee (approved):

PR #5027, opened 2024-09-05
Improve logging

KShivendu:
Calling telemetry API now checks for pending optimizations while fetching shard status. So it looks like Qdrant is now repeatedly logging `Merge candidates: [...]` which gives a false appearance that Qdrant merge optimizer is stuck 

### All Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/optimizers/merge_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/merge_optimizer.rs
@@ -139,7 +139,6 @@ impl SegmentOptimizer for MergeOptimizer {
         if candidates.len() < 3 {
             return vec![];
         }
-        log::trace!("Possible merge candidates: {:?}", candidates);
         candidates
     }
 
--- a/lib/collection/src/collection_manager/optimizers/merge_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/merge_optimizer.rs
@@ -139,7 +139,7 @@ impl SegmentOptimizer for MergeOptimizer {
         if candidates.len() < 3 {
             return vec![];
         }
-        log::debug!("Possible merge candidates: {:?}", candidates);
+        log::trace!("Possible merge candidates: {:?}", candidates);
         candidates
     }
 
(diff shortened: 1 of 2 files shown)


lib/collection/src/collection_manager/optimizers/merge_optimizer.rs
```diff
@@ -139,7 +139,7 @@ impl SegmentOptimizer for MergeOptimizer {
         if candidates.len() < 3 {
             return vec![];
         }
-        log::debug!("Merge candidates: {:?}", candidates);
+        log::debug!("Possible merge candidates: {:?}", candidates);
```
timvisee, line 142: Let's do trace because we still have a lot of logging now:

```suggestion
        log::trace!("Possible merge candidates: {:?}", candidates);
```

timvisee (approved):

agourlay:
My expectation is that calling the telemetry is cheap and has no side effect with the optimizer.

I personally used that very `debug` log in that past during testing so I am sad to see it go away.

timvisee:
Based on what we said in the daily, @agourlay raised two concerns:
1. he has used this very line during testing many times
2. is it any good that calling /telemetry reaches this deep into optimizers?

A solution for the first point may be to remove this log statement from the condition check, and move it into the place where we actually start (merge) optimizing the segments. As a fallback we could resort to the optimizer log in telemetry which would should the same, including segment IDs.

I'm not quite sure about the second part. I agree that it would be better to not reach this deep. But I'm unsure if we can easily solve the problems without it, namely: 1. detecting grey state, being pending optimizations 2. prevent status blinking (with possible optimizer blockers such as CPU budget).

Thoughts?

agourlay:
@timvisee thank you for the write up!

I am fine with the solution for the first point, so that we keep the log and avoid the noise.
The architectural aspects can be tackled later.

KShivendu:
I think @generall's solution is clean. We can go ahead with that. wdyt?

agourlay (approved):
:+1:

timvisee (approved):
Thanks!

PR #5029, opened 2024-09-05
initial structures for reading mmap text index

generall:
Continue working on mmap text index

### DIFF
--- a/lib/common/memory/src/mmap_type.rs
+++ b/lib/common/memory/src/mmap_type.rs
@@ -269,10 +269,7 @@ impl<T> MmapSlice<T> {
         self.mmap.flusher()
     }
 
-    pub fn create(path: &Path, slice: &[T]) -> Result<()>
-    where
-        T: Copy,
-    {
+    pub fn create(path: &Path, mut iter: impl ExactSizeIterator<Item = T>) -> Result<()> {
         let file = OpenOptions::new()
             .read(true)
             .write(true)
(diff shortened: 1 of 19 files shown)


lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mod.rs
```diff
@@ -1,3 +1,53 @@
...
+
+const POSTINGS_FILE: &str = "postings.dat";
+const VOCAB_FILE: &str = "vocab.dat";
+const POINT_TO_TOKENS_COUNT_FILE: &str = "point_to_tokens_count.dat";
+
+pub struct MmapInvertedIndex {
+    path: PathBuf,
+    postings: MmapPostings,
+    vocab: MmapHashMap<str>,
+    point_to_tokens_count: MmapSliceBufferedUpdateWrapper<Option<usize>>,
```
generall, line 22: Would it be possible to avoid Option? Value = None means deleted? If so, let's use bitmap instead.
`size_of::<Option<usize>>()` = 16 bytes.

generall:
ToDo: write tests

lib/common/memory/src/mmap_type.rs
```diff
@@ -365,17 +370,26 @@ impl MmapBitSlice {
...
 
-        let len = bitslice.len();
-        file.set_len(len as u64)?;
+        let bits_count = bitslice.len();
 
-        let mut mmap = unsafe { MmapMut::map_mut(&file)? };
-        mmap.fill(0);
+        let bytes_count = bits_count.div_ceil(u8::BITS as usize);
+
+        let bytes_count = max(Self::MIN_FILE_SIZE, bytes_count.next_power_of_two());
```
generall, line 377: Why `next_power_of_two`? Doesn't this mean we would have 50% overhead worst case?

coszio: right, I think it should be `.next_multiple_of(size_of::<usize>())` instead.

coszio:
Update: I just wrote a test with several cases to check the equivalence of immutable to mmap conversion, as well as fixing bugs in `MmapBitSlice::create`

coszio (approved):

PR #5033, opened 2024-09-06
Remove unecessary clippy allow deadcode

agourlay:
Found unnecessary clippy allow statements that we can get rid of.

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -51,7 +51,6 @@ use crate::shards::{replica_set, CollectionId};
 use crate::telemetry::CollectionTelemetry;
 
 /// Collection's data is split into several shards.
-#[allow(dead_code)]
 pub struct Collection {
     pub(crate) id: CollectionId,
     pub(crate) shards_holder: Arc<LockedShardHolder>,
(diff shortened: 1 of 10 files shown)


timvisee (approved):
Thanks!

Maybe we can replace remaining cases with the new `#[expect(dead_code)]` to show a lint error when they have become obsolete.

agourlay:
Yes I used the new `expect` to find some of those.
My plan was to follow up with the migration :)

timvisee:
Fantastic!

agourlay:
It seems the `expect` feature is suffering from false positive.

```
error: this lint expectation is unfulfilled
   --> lib/segment/src/index/field_index/geo_hash.rs:105:15
    |
105 |      #[expect(dead_code)]
    |               ^^^^^^^^^
    |
    = note: `-D unfulfilled-lint-expectations` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unfulfilled_lint_expectations)]`
```

but when removed

```
error: field `south_east` is never read
   --> lib/segment/src/index/field_index/geo_hash.rs:105:5
    |
102 | struct GeohashBoundingBox {
    |        ------------------ field in this struct
...
105 |     south_east: GeoHash, // field is not involved in the calculations, but is kept for symmetry
    |     ^^^^^^^^^^
    |
    = note: `GeohashBoundingBox` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
    = note: `-D dead-code` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(dead_code)]`
``` 

I will postpone the migration for the time being :(

timvisee:
I guess that may be because we have different compiler targets, some trigger the lint, others do not.

PR #5035, opened 2024-09-06
Better json path parsing errors

generall:


JojiiOfficial (approved):

generall:
> This doesn't patch anything, but only helps us understand any errors caused by it better, am I right?

yes

xzfc (approved):

PR #5037, opened 2024-09-06
[Strict mode] default config

JojiiOfficial:
Depends on #5003

Since we need the strict mode config struct to be in `/lib/segment` but also need a struct that derives `Merge`, I decided to create two structs for strict mode: StrictModeDiff and StrictMode.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -91,7 +91,7 @@ storage:
     # Only supported on Linux, must be enabled in your kernel.
     # See: <https://qdrant.tech/articles/io_uring/#and-what-about-qdrant>
     #async_scorer: false
-    
+
   optimizers:
     # The minimal fraction of deleted vectors in a segment, required to perform segment optimization
     deleted_threshold: 0.2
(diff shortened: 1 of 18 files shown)


lib/api/src/grpc/conversions.rs
```diff
@@ -1754,6 +1755,36 @@ impl From<HnswConfigDiff> for segment::types::HnswConfig {
...
+            max_query_limit: value.max_query_limit.map(|i| i as usize),
+            max_timeout: value.max_timeout.map(|i| i as usize),
+            unindexed_filtering_retrieve: value.unindexed_filtering_retrieve,
+            unindexed_filtering_update: value.unindexed_filtering_update,
+            search_max_hnsw_ef: value.search_max_hnsw_ef.map(|i| i as usize),
+            search_allow_exact: value.search_allow_exact,
+            search_max_oversampling: value.search_max_oversampling.map(f64::from),
+        }
+    }
+}
```
timvisee, line 1771: If we can construct a regular config from a diff, I don't see much reason to also have a diff type.

lib/segment/src/types.rs
```diff
@@ -661,6 +664,47 @@ impl From<BinaryQuantizationConfig> for QuantizationConfig {
     }
 }
 
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Validate, Clone, PartialEq, Default)]
+pub struct StrictModeConfig {
```
timvisee, line 668: At this time the -Diff struct of this appears to be exactly the same. In that case, does it make sense to use a diff structure at all?

config/config.yaml
```diff
@@ -197,6 +197,32 @@ storage:
...
+      unindexed_filtering_update: null
+
+      # Max HNSW value allowed in search parameters.
+      search_max_hnsw_ef: null
+
+      # Whether exact search is allowed or not.
+      search_allow_exact: null
+
+      # Max oversampling value allowed in search.
+      search_max_oversampling: null
```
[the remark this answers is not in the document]
JojiiOfficial: It's not necessary. I added them to make it easier to configure.
Do you prefer them to be commented out, or have them removed completely?

timvisee (approved):

PR #5038, opened 2024-09-06
Fix bugs in and refactor `hash_ring` module

ffuugoo:
This PR fixes a few bugs in `hash_ring` module:

- make `HashRing::add`/`HashRingRouter::add` idempotent
  - `hashring::HashRing::add` is not idempotent
  - adding same node multiple times adds multiple instances of the node to the inner state
  - and each call to `remove` only removes a single instance of the node
  - so, to ensure `HashRing::add` is idempotent, we add a `HashSet` into `HashRing`, to explicitly track nodes added to the hashring
- correctly switch hashring into `Resharding` mode
  - before, when switching into `Resharding` mode, we only ensured new node is added to/removed from `new` hashring
  - however, when recovering `Resharding` hashring state from disk, we
    - initialize hashring in `Single` mode
    - add *all* shards (including *resharding target shard*) to the hashring
    - switch hashring into `Resharding` mode
    - and so, it this case, target shard would *already* be in `old` hashring, and `new.add(shard)` would be a no-op
- forbid adding new shards to `HashRingRouter::Resharding`
  - it's confusing and error-prone

This PR also incorporates @timvisee's changes from #4724, to correctly rebuild hashrings when resharding is in progress.

And finally, this PR cleanup `hash_ring` module: minimizes trait bounds on `impl` blocks, rename some methods, improve interfaces, improve error messages, etc.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/hash_ring.rs
+++ b/lib/collection/src/hash_ring.rs
@@ -59,10 +59,8 @@ impl<T: Hash + Copy + PartialEq> HashRingRouter<T> {
         match self {
             Self::Single(ring) => ring.add(shard),
             Self::Resharding { old, new } => {
-                if new.get(&shard).is_none() {
-                    old.add(shard);
-                    new.add(shard);
-                }
+                old.add(shard);
+                new.add(shard);
             }
         }
     }
--- a/lib/collection/src/hash_ring.rs
+++ b/lib/collection/src/hash_ring.rs
@@ -253,7 +253,7 @@ impl<T: Hash + Copy> HashRing<T> {
             HashRing::Fair { ring, scale } => {
                 let mut removed = false;
                 for i in 0..*scale {
-                    if ring.remove(&(*shard, i)).is_some() {
+                    while ring.remove(&(*shard, i)).is_some() {
                         removed = true;
                     }
                 }
--- a/lib/collection/src/hash_ring.rs
+++ b/lib/collection/src/hash_ring.rs
@@ -12,7 +12,7 @@ use crate::shards::shard::ShardId;
 
 pub const HASH_RING_SHARD_SCALE: u32 = 100;
 
-#[derive(Clone, PartialEq, Eq, Debug)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum HashRingRouter<T: Eq + Hash = ShardId> {
     /// Single hashring
     Single(HashRing<T>),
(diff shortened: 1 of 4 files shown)


lib/collection/src/hash_ring.rs
```diff
@@ -195,105 +176,188 @@ impl<T: Hash + Copy + PartialEq> HashRingRouter<T> {
...
     }
 }
 
+impl<T: Eq + Hash> PartialEq for HashRing<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.scale() == other.scale() && self.nodes() == other.nodes()
+    }
+}
+
+impl<T: Eq + Hash> Eq for HashRing<T> {}
```
timvisee, line 359: Can't we just derive PartialEq here as well? I believe that should be fine now since this made add/remove idempotent by the attached nodes hash set.

lib/collection/src/hash_ring.rs
```diff
@@ -44,24 +44,24 @@ impl<T: Hash + Copy + PartialEq> HashRingRouter<T> {
...
+
+                if !old.contains(&shard) && !new.contains(&shard) {
                     old.add(shard);
                     new.add(shard);
+                    true
+                } else {
+                    false
                 }
             }
         }
```
timvisee, line 67: I don't understand how this works.

Why wouldn't we only add in the new hash ring here?

lib/collection/src/hash_ring.rs
```diff
@@ -195,53 +195,104 @@ impl<T: Hash + Copy + PartialEq> HashRingRouter<T> {
...
+
+        nodes.clone()
     }
 }
 
+impl<T: Copy + Eq + Hash> PartialEq for HashRingRouter<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.single_ring() == other.single_ring()
+            && self.old_ring() == other.old_ring()
+            && self.new_ring() == other.new_ring()
```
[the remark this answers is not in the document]
timvisee: Done in https://github.com/qdrant/qdrant/pull/5038/commits/ed9ee597fe862dd316d04665a0df7969e2dfc6eb

lib/collection/src/hash_ring.rs
ffuugoo: We decided to simply forbid adding shards into `HashRingRouter::Resharding` 🤷‍♀️

lib/collection/src/hash_ring.rs
ffuugoo: Hmmmm.... You're right, it should be safe now.

ffuugoo: Done.

lib/collection/src/hash_ring.rs
```diff
@@ -22,48 +22,18 @@ pub enum HashRingRouter<T = ShardId> {
...
+    pub fn add(&mut self, shard: T) -> bool {
         match self {
             Self::Single(ring) => ring.add(shard),
-            Self::Resharding { old, new } => {
-                if new.get(&shard).is_none() {
-                    old.add(shard);
-                    new.add(shard);
-                }
-            }
+            Self::Resharding { .. } => panic!("can't add shards to hashring in resharding mode"),
```
timvisee, line 36: Not sure I like this hard panic here.

The reason is that it might be very annoying to resolve in a customers' cluster if it ends up in WAL.

I'd prefer to do a debug assert instead, and do some best-effort approach in release builds.

Wdyt?

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -208,6 +207,17 @@ impl ShardHolder {
...
+        // Restore resharding hash ring if resharding is active and haven't reached
+        // `WriteHashRingCommitted` stage yet
+        if let Some(state) = self.resharding_state.read().deref() {
+            if state.stage < ReshardStage::WriteHashRingCommitted {
+                rings
+                    .get_mut(&state.shard_key)
+                    .expect("must have hash ring for current resharding shard key")
+                    .start_resharding(state.shard_id, state.direction);
+            }
+        }
```
timvisee, line 219: I don't think this is good enough in case of sharding down.

Imagine the following situation:
- sharding down
- just committed write hash ring
- shard still exists because we only drop it at the very end of resharding

If we now rebuild the hash ring it wrongfully includes the last shard we'll be dropping at the end of resharding.

lib/collection/src/hash_ring.rs
ffuugoo: I don't think it can ever end up in WAL, though.

timvisee: I hit this panic right away during testing, when resharding is started through consensus.

ffuugoo: We agreed to revert this to initial behavior of `add`, because it requires a more thorough refactoring, which we can introduce later.

timvisee (approved):

PR #5039, opened 2024-09-06
use atomic write to initially create alias mapping

generall:
It is possible that service if shut down after file is created, but before data is fsynced, which may leave us with empty file
(this reproduces on filesystem-level snapshots)

If we use atomic writes, it should be impossible to end up with corrupted file


timvisee:
Nice find. Thanks for the patch.

PR #5043, opened 2024-09-09
Fix shard/collection status blinking

KShivendu:
We should mark shard status as:
- Yellow if optimizers are suboptimal and we have run at least one since starting the node
- Grey if optimizers are suboptimal and we haven't run any since starting the node

This makes the status consistent

[Slack discussion](https://qdrant.slack.com/archives/C077CCW8NU8/p1725609571955249?thread_ts=1725568515.703859&cid=C077CCW8NU8)

### All Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -994,10 +994,14 @@ impl LocalShard {
 
         // If status looks green/ok but some optimizations are suboptimal
         if status == ShardStatus::Green && optimizer_status == OptimizersStatus::Ok {
-            let (has_triggered_any_optimizers, has_suboptimal_optimizers) =
-                self.update_handler.lock().await.pending_optimizations();
+            let (has_triggered_any_optimizers, has_suboptimal_optimizers) = self
+                .update_handler
+                .lock()
+                .await
+                .check_suboptimal_optimizations();
 
             if has_suboptimal_optimizers {
+                // Check if any optimizations were triggered after starting the node
                 if has_triggered_any_optimizers {
                     status = ShardStatus::Yellow;
                 } else {
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -992,14 +992,20 @@ impl LocalShard {
             }
         }
 
-        // If status looks green/ok but optimizations can be triggered, mark as grey
-        if status == ShardStatus::Green
-            && optimizer_status == OptimizersStatus::Ok
-            && self.update_handler.lock().await.has_non_optimal_segments()
-        {
-            // This can happen when a node is restarted (crashed), because we don't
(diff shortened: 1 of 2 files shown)


lib/collection/src/update_handler.rs
```diff
@@ -433,18 +433,18 @@ impl UpdateHandler {
     /// Checks conditions for all optimizers and returns whether any is satisfied
     ///
     /// In other words, if this returns true we have pending optimizations.
-    pub(crate) fn has_non_optimal_segments(&self) -> bool {
-        // If we did trigger optimizers at least once, we do not consider to be pending
-        if self.has_triggered_optimizers.load(Ordering::Relaxed) {
-            return false;
-        }
+    pub(crate) fn check_suboptimal_optimizations(&self) -> (bool, bool) {
```
timvisee, line 436: I find `check_suboptimal_optimizations` rather confusing and would prefer to call it something like `check_optimizer_conditions`. Wdyt?

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -992,14 +992,24 @@ impl LocalShard {
...
+            if has_suboptimal_optimizers {
+                // Check if any optimizations were triggered after starting the node
+                if has_triggered_any_optimizers {
+                    status = ShardStatus::Yellow;
+                } else {
+                    // This can happen when a node is restarted (crashed), because we don't
+                    // automatically trigger optimizations on restart to avoid a crash loop
+                    status = ShardStatus::Grey;
+                }
+            }
```
timvisee, line 1012: I like that we only report grey here if we had no optimizations before :+1:

lib/collection/src/update_handler.rs
KShivendu: Sounds better. thanks!

lib/collection/src/shards/local_shard/mod.rs
KShivendu: thanks to Andrey

generall (approved):

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -992,14 +992,24 @@ impl LocalShard {
...
```
timvisee, line 1012: By the way, we can do something like this. Whichever you prefer :+1: 

```suggestion
            match (has_suboptimal_optimizers, has_triggered_any_optimizers) {
                // Optimizations are queued but other optimizations must finish first
                (true, true) => status = ShardStatus::Yellow,
                // Optimizations are pending any update operation
                // Can happen if node crashed, don't automatically trigger to prevent crash loop
                (true, false) => status = ShardStatus::Grey,
                (false, _) => {},
            }
```

timvisee (approved):

lib/collection/src/shards/local_shard/mod.rs
KShivendu: It's fine like this imo

PR #5060, opened 2024-09-10
make resharding_operations optional, so OpenAPI is properly generated

generall:


lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -445,10 +445,12 @@ impl ShardHolder {
                 shard_key: resharding_state.shard_key.clone(),
                 comment: status.map(|p| p.comment),
             });
+        } else {
+            return None;
```
Taym95, line 449: I think you can use `?` in `let resharding_state = &*self.resharding_state.read()?;` and avoid returning `None` here

timvisee: Not really, because it's a parking_lot RwLockGuard that doesn't implement Try itself.

But I've moved the return up:

```rust
        let Some(resharding_state) = &*self.resharding_state.read() else {
            return None;
        };
```

Thanks!

timvisee (approved):

joein (approved):
works like a charm

PR #5064, opened 2024-09-11
Make computing snapshot checksums faster in debug

timvisee:
Computing file checksums is extremely slow in our debug builds. I was experimenting with snapshots again today and decided to fix this because it was quite annoying.

This enables the release optimizations in all profiles for the sha2 sub crate to always benefit from the faster implementation.

Before: checksum takes 15.6s
After: checksum takes 0.4s

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/common/sha_256.rs
+++ b/lib/collection/src/common/sha_256.rs
@@ -110,4 +110,37 @@ mod tests {
         assert!(hashes_equal("0123abc", "0123abc "));
         assert!(!hashes_equal("0123abc", "0123abd"));
     }
+
+    /// Test consistency between built-in sha256 hasher and sha256sum binary
+    #[tokio::test]
+    #[cfg(all(debug_assertions, unix))]
+    async fn test_hash_file_bin() {
+        use std::io::Write;
+
+        use rand::{thread_rng, Rng};
+
+        // Create a temporary file with random data
+        let mut temp_file = tempfile::Builder::new()
+            .prefix("test-sha256sum")
+            .tempfile()
+            .unwrap();
+        let mut buf = [0u8; 128];
+        thread_rng().fill(&mut buf[..]);
+        temp_file.write_all(&buf).unwrap();
+
+        let hash_builtin = hash_file_builtin(temp_file.path()).await.unwrap();
+        let hash_bin = match hash_file_bin(temp_file.path()).await {
+            Ok(hash) => hash,
+            // May error if the sha256bin binary is not available on this system
+            Err(err) => {
+                log::warn!("Failed to compute checksum with sha256sum, skipping test: {err}");
+                return;
+            }
+        };
+
+        assert_eq!(
+            hash_builtin, hash_bin,
+            "built-in hasher and sha256sum binary must produce the same hash",
+        );
+    }
 }
--- a/lib/collection/src/common/sha_256.rs
+++ b/lib/collection/src/common/sha_256.rs
@@ -5,25 +5,10 @@ use bytes::BytesMut;
 use sha2::{Digest, Sha256};
 use tokio::io::AsyncReadExt;
 
-const SHA256_HASH_LENGTH: usize = 64;
-
 /// Compute sha256 hash for the given file
 pub async fn hash_file(file_path: &Path) -> io::Result<String> {
     log::debug!("Computing checksum for file: {file_path:?}");
 
-    // On Linux in debug builds, get hash through sha256sum because it's significantly faster
-    #[cfg(all(debug_assertions, target_os = "linux"))]
-    match hash_file_bin(file_path).await {
-        Ok(hash) => return Ok(hash),
-        Err(err) => log::warn!(
-            "Failed to compute checksum with sha256sum, falling back to built-in hasher: {err}",
-        ),
-    };
-
-    hash_file_builtin(file_path).await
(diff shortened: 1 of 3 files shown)


ffuugoo:
On macOS it's just `shasum`... 😂
Screenshot 2024-09-11 at 11 39 40

Though, I don't think we care about macOS much.

timvisee:
> On macOS it's just `shasum`... 😂 Screenshot 2024-09-11 at 11 39 40
> 
> Though, I don't think we care about macOS much.

Oh noes!

Thanks for checking. I'll validate on my end how likely it is for `sha256sum` to exist. Changing the compile time conditions is probably good enough.

timvisee:
I asked in our internal channels, and all Linux users do seem to have it.

Linux does also have `shasum` but it's about 2.5 times slower in my quick tests.

I'll just enable it on Linux only then to keep things simple. It's not very significant anyway and falls back if the binary is not there.

lib/collection/src/common/sha_256.rs
```diff
@@ -25,6 +41,52 @@ pub async fn hash_file(file_path: &Path) -> io::Result<String> {
...
+/// Compute sha256 hash for the given file using the `sha256sum` binary
+///
+/// May fail if the binary does not exist on the system.
+#[cfg(all(debug_assertions, unix))]
+async fn hash_file_bin(file_path: &Path) -> io::Result<String> {
+    use tokio::process::Command;
+
+    let output = match Command::new("sha256sum")
+        .arg("-b")
+        .arg(file_path.as_os_str())
```
ffuugoo, line 53: I think just `file_path` would work here.

```suggestion
        .arg(file_path)
```

lib/collection/src/common/sha_256.rs
```diff
@@ -25,6 +41,52 @@ pub async fn hash_file(file_path: &Path) -> io::Result<String> {
...
+            ));
+        }
+        // Binary must have been called successfully
+        Err(err) => {
+            return Err(io::Error::new(
+                io::ErrorKind::Other,
+                format!("Failed to compute checksum with sha256sum: {err}"),
+            ));
+        }
+        Ok(command) => command,
```
ffuugoo, line 74: I'd use `Ok(command) if command.status.success() => command` at the top.

```suggestion
        Ok(command) if command.status.success() => command,
        Ok(command) => {
            return Err(io::Error::new(
                io::ErrorKind::Other,
                format!(
                    "Failed to compute checksum with sha256sum: {}",
                    String::from_utf8(command.stderr).unwrap(),
                ),
            ));
        }
        Err(err) => {
            return Err(io::Error::new(
                io::ErrorKind::Other,
                format!("Failed to compute checksum with sha256sum: {err}"),
            ));
        }
```

ffuugoo (approved):
Feel free to ignore comments, cause they are bikesheddiest of bikesheddings.

generall (approved):
`sha256sum` exists even in our docker container (which is not even required) 
So I assume this this fine.

An alternative I would consider is to just disable checksum in dev

lib/collection/src/common/sha_256.rs
```diff
@@ -25,6 +41,50 @@ pub async fn hash_file(file_path: &Path) -> io::Result<String> {
...
+/// Compute sha256 hash for the given file using the `sha256sum` binary
+///
+/// May fail if the binary does not exist on the system.
+#[cfg(all(debug_assertions, target_os = "linux"))]
+async fn hash_file_bin(file_path: &Path) -> io::Result<String> {
+    use tokio::process::Command;
+
+    let output = match Command::new("sha256sum")
+        .arg("-b")
+        .arg(file_path)
```
xzfc, line 53: Security/correctness: prepend `--` argument when passing arbitrary filenames so weird filenames like `--help` won't be treated as options.

```suggestion
    let output = match Command::new("sha256sum")
        .arg("-b")
        .arg("--")
        .arg(file_path)
```

ffuugoo:
> We could go with [`Cargo.toml` overrides](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides), i.e. by adding this to our top-level `Cargo.toml`:
> 
> ```toml
> [profile.dev.package.sha2]
> opt-level = 3
> ```

🤯

I vote to test this one first, then.

lib/collection/src/common/sha_256.rs
timvisee: Thanks. Didn't think about that one!

timvisee:
> > Ideally we enable release optimizations for the built-in hasher at all times. But I don't see a proper way to achieve this.
> 
> We could go with [`Cargo.toml` overrides](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides), i.e. by adding this to our top-level `Cargo.toml`:
> 
> ```toml
> [profile.dev.package.sha2]
> opt-level = 3
> ```

Woah! I wasn't aware of this 🥹

Let me give it a quick shot. I'll definitely go with that if that works.

timvisee:
Brilliant, works like a charm!

Latest test for hashing duration:
- before: 15.6s
- with `sha256sum`: 0.3s (got to love GNU coretutils)
- with release mode: 0.4s

That's obviously good enough. I've updated the PR to enable release optimizations for the sub-crate.

PR #5067, opened 2024-09-12
Refactor `FullTextIndex` as enum

coszio:
- Refactors `InvertedIndex` from an enum into a trait
- Changes `FullTextIndex` from a struct into an enum which now holds mutable and immutable versions, but soon will also hold Mmap version. This also follows same pattern as current map and numeric indices.

All of this is done just because mmap text index doesn't need a db associated to it. So this refactor is necessary for integrating it.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
@@ -2,13 +2,12 @@ use std::collections::HashMap;
 
 use common::types::PointOffsetType;
 
+use super::inverted_index::InvertedIndex;
 use crate::common::operation_error::{OperationError, OperationResult};
 use crate::index::field_index::full_text_index::compressed_posting::compressed_posting_list::CompressedPostingList;
 use crate::index::field_index::full_text_index::inverted_index::{ParsedQuery, TokenId};
 use crate::index::field_index::full_text_index::mutable_inverted_index::MutableInvertedIndex;
 use crate::index::field_index::full_text_index::postings_iterator::intersect_compressed_postings_iterator;
-
-use super::inverted_index::InvertedIndex;
 
 #[cfg_attr(test, derive(Clone))]
 #[derive(Default)]
@@ -116,7 +115,7 @@ impl InvertedIndex for ImmutableInvertedIndex {
         self.points_count
     }
 
-    fn get_token(&self, token: &str) -> Option<TokenId> {
+    fn get_token_id(&self, token: &str) -> Option<TokenId> {
         self.vocab.get(token).copied()
     }
 }
--- a/lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
@@ -98,17 +98,16 @@ impl InvertedIndex for ImmutableInvertedIndex {
     }
 
     fn values_is_empty(&self, point_id: PointOffsetType) -> bool {
-        if self.point_to_tokens_count.len() <= point_id as usize {
-            return true;
-        }
-        self.point_to_tokens_count[point_id as usize].is_none()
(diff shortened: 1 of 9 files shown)


lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
```diff
@@ -83,12 +97,26 @@ impl ImmutableInvertedIndex {
...
-        self.vocab.iter().filter_map(|(token, &posting_idx)| {
-            self.postings
-                .get(posting_idx as usize)
-                .map(|posting| (token.as_str(), posting.len()))
-        })
+    fn values_is_empty(&self, point_id: PointOffsetType) -> bool {
+        if self.point_to_tokens_count.len() <= point_id as usize {
+            return true;
+        }
+        self.point_to_tokens_count[point_id as usize].is_none()
```
Taym95, line 104:
```suggestion        
        self.point_to_tokens_count
        .get(point_id as usize)
        .map_or(true, |count| count.is_none())
```

lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
```diff
@@ -83,12 +97,26 @@ impl ImmutableInvertedIndex {
...
+    }
+
+    fn values_count(&self, point_id: PointOffsetType) -> usize {
+        if self.point_to_tokens_count.len() <= point_id as usize {
+            return 0;
+        }
+        self.point_to_tokens_count[point_id as usize].unwrap_or(0)
```
Taym95, line 111:
```suggestion
       self. point_to_tokens_count
        .get(point_id as usize)
        .and_then(|&count| count)  
        .unwrap_or(0)
```

lib/segment/src/index/field_index/full_text_index/mutable_inverted_index.rs
```diff
@@ -135,34 +144,47 @@ impl MutableInvertedIndex {
         intersect_postings_iterator(postings)
     }
 
-    pub fn values_count(&self, point_id: PointOffsetType) -> usize {
-        // Maybe we want number of documents in the future?
-        self.get_doc(point_id).map(|x| x.len()).unwrap_or(0)
+    fn get_posting_len(&self, token_id: TokenId) -> Option<usize> {
+        self.postings
+            .get(token_id as usize)
+            // unwrap safety: token_id is a valid index. Same case as in `filter` method.
```
xzfc, line 150: After extracting this snippet of code from `estimate_cardinality` into a separate trait method, this assumption is more error prone as it assumes the only callsite calls it correctly.

I suggest either to make it return an `Err` or `None` (maybe with a `debug_assert`) instead of a panicking unwrap, or at least put a warning into the trait method doc, e.g.:

```rust
pub trait InvertedIndex {
    …

    /// # Panics
    /// Panics when `token_id` is not a valid index.
    fn get_posting_len(&self, token_id: TokenId) -> Option<usize>;

    …
}
```

xzfc (approved):
LGTM besides remarks mentioned above.

lib/segment/src/index/field_index/full_text_index/mutable_inverted_index.rs
coszio: Tbh the trait was not strictly necessary since we never request a trait object or an impl Trait. But it does make implementing all dispatched methods easier. 

I agree that we should document the panic, but even better, we can just return `None` :)

lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
coszio: Added, thanks!

lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
coszio: Added, thanks!

PR #5070, opened 2024-09-12
Improve immutable map index performance: sort point IDs, binary search and shift on delete

timvisee:
Alternative to: <https://github.com/qdrant/qdrant/pull/5072>

See <https://github.com/qdrant/qdrant/pull/5072> for details.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

PR #5072, opened 2024-09-12
Improve immutable map index performance: sort point IDs, binary search and deleted flags

timvisee:
Alternative to: <https://github.com/qdrant/qdrant/pull/5070>

_WIP_

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/map_index/immutable_map_index.rs
+++ b/lib/segment/src/index/field_index/map_index/immutable_map_index.rs
@@ -4,6 +4,7 @@ use std::iter;
 use std::ops::Range;
 use std::sync::Arc;
 
+use bitvec::vec::BitVec;
 use common::types::PointOffsetType;
 use parking_lot::RwLock;
 use rocksdb::DB;
@@ -16,8 +17,11 @@ use crate::common::rocksdb_wrapper::DatabaseColumnWrapper;
 use crate::index::field_index::immutable_point_to_values::ImmutablePointToValues;
 
 pub struct ImmutableMapIndex<N: MapIndexKey + ?Sized> {
-    value_to_points: HashMap<N::Owned, Range<u32>>,
+    value_to_points: HashMap<N::Owned, (Range<u32>, u32)>,
+    /// Container holding a slice of point IDs per value. `value_to_point` holds the range per value.
+    /// Each slice MUST be sorted so that we can binary search over it.
     value_to_points_container: Vec<PointOffsetType>,
+    deleted_value_to_points_container: BitVec,
     point_to_values: ImmutablePointToValues<N::Owned>,
     /// Amount of point which have at least one indexed payload value
     indexed_points: usize,
@@ -35,6 +39,7 @@ impl<N: MapIndexKey + ?Sized> ImmutableMapIndex<N> {
         Self {
             value_to_points: Default::default(),
             value_to_points_container: Default::default(),
+            deleted_value_to_points_container: Default::default(),
             point_to_values: Default::default(),
             indexed_points: 0,
             values_count: 0,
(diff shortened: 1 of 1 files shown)


lib/segment/src/index/field_index/map_index/immutable_map_index.rs
```diff
@@ -43,13 +48,15 @@ impl<N: MapIndexKey + ?Sized> ImmutableMapIndex<N> {
     }
 
     /// Return mutable slice of a container which holds point_ids for given value.
+    ///
+    /// The returned slice is sorted and does contain deleted values.
     fn get_mut_point_ids_slice<'a>(
-        value_to_points: &mut HashMap<N::Owned, Range<u32>>,
+        value_to_points: &HashMap<N::Owned, (Range<u32>, u32)>,
```
generall, line 54: I would rather prefer to use bitmap, seems like more straightforward and already implemented in other places.

lib/segment/src/index/field_index/map_index/immutable_map_index.rs
```diff
@@ -16,8 +17,11 @@ use crate::common::rocksdb_wrapper::DatabaseColumnWrapper;
 use crate::index::field_index::immutable_point_to_values::ImmutablePointToValues;
 
 pub struct ImmutableMapIndex<N: MapIndexKey + ?Sized> {
-    value_to_points: HashMap<N::Owned, Range<u32>>,
+    value_to_points: HashMap<N::Owned, (Range<u32>, u32)>,
```
xzfc, line 20: Maybe it's worth converting `(Range<u32>, u32)` into a struct with named fields.

timvisee: Agreed. Done!

xzfc (approved):
LGTM.

I think it's also possible to implement another approach:

```rust
enum PointsSet {
   // Up to three points could be stored directly in within this type
   Inline([u32; 3]),
   // 3..32 points could be stored in value_to_points_container.
   // Do a linear search during deletion, but for up to 16 elements it would be neglectible.
   Container(Range<u32>),
   // Greater amount of points stored in a heap-allocated structure.
   // The memory overhead should only be a concern for small amount of points which is covered by previous variant.
   // Enables `O(log(n))` deletion times.
   BTreeSet(BTreeSet<u32>),
}
```

timvisee:
> LGTM.
> 
> I think it's also possible to implement another approach:
> 
> ```rust
> enum PointsSet {
>    // Up to three points could be stored directly in within this type
>    Inline([u32; 3]),
>    // 3..32 points could be stored in value_to_points_container.
>    // Do a linear search during deletion, but for up to 16 elements it would be neglectible.
>    Container(Range<u32>),
>    // Greater amount of points stored in a heap-allocated structure.
>    // The memory overhead should only be a concern for small amount of points which is covered by previous variant.
>    // Enables `O(log(n))` deletion times.
>    BTreeSet(BTreeSet<u32>),
> }
> ```

Definitely something to investigate.

Out of curiosity, why did you choose 3? Is it because it's the maximum size without growing the enum in terms of memory?

xzfc:
> Out of curiosity, why did you choose 3? Is it because it's the maximum size without growing the enum in terms of memory?

Yep, `sizeof::<BTreeSet<u32>>()` is 3.

generall (approved):

generall:
> LGTM.
> 
> I think it's also possible to implement another approach:
> 
> ```rust
> enum PointsSet {
>    // Up to three points could be stored directly within this type
>    Inline([u32; 3]),
>    // 4..24 points could be stored in value_to_points_container.
>    // Do a linear search during deletion, but for up to 24 elements it would be neglectible.
>    Container(Range<u32>),
>    // Greater amount of points stored in a heap-allocated structure.
>    // The memory overhead should only be a concern for small amount of points which is covered by previous variant.
>    // Enables `O(log(n))` deletion times.
>    BTreeSet(BTreeSet<u32>),
> }
> ```

This structure takes 32 bytes in my test

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4716c988b3f3b34c0aad9c08e7bf23f2

Issue #5076, opened 2024-09-13
Potential consensus desync on delete shards during transfer


generall:
## Current Behavior

```
- call the move_shard api from 7238360501313588 -> 6016877126863533  for shard_id 0
- observed that it is still in status mode in 6016877126863533  for shard_id 0
- called the drop_shard api, for 6016877126863533  for shard_id 0
- collection fails with not-enough local replicas
```

## Expected Behavior

Collection state should be same on all nodes

PR #5079, opened 2024-09-13
Align delete replica logic with transfer cancellation

generall:
abort shard transfer on delete of replica + check that we don't delete last active

Related to: https://github.com/qdrant/qdrant/issues/5076

### DIFF
--- a/lib/collection/src/collection/collection_ops.rs
+++ b/lib/collection/src/collection/collection_ops.rs
@@ -194,13 +194,12 @@ impl Collection {
                 });
             }
 
-            if peers.len() == 1 {
+            // Check that we are not removing the
+            if replica_set.is_last_active_replica(peer_id) {
                 return Err(CollectionError::BadRequest {
-                    description: format!("Shard {shard_id} must have at least one replica"),
+                    description: format!("Shard {shard_id} must have at least one active replica after removing {peer_id}"),
                 });
             }
-
-            replica_set.remove_peer(peer_id).await?;
 
             let all_nodes_cancel_transfers = self
                 .channel_service
@@ -212,10 +211,12 @@ impl Collection {
 
                 // ...and cancel transfer tasks and remove transfers from internal state
                 for transfer in transfers {
-                    self.finish_shard_transfer(transfer, Some(&shard_holder))
+                    self.abort_shard_transfer(transfer.key(), Some(&shard_holder))
                         .await?;
                 }
             }
+
+            replica_set.remove_peer(peer_id).await?;
 
             // We can't remove the last repilca of a shard, so this should prevent removing
             // resharding shard, because it's always the *only* replica.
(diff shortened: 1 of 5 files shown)


lib/collection/src/collection/shard_transfer.rs
```diff
@@ -275,6 +275,7 @@ impl Collection {
     pub async fn abort_shard_transfer(
         &self,
         transfer_key: ShardTransferKey,
+        shard_holder: Option<&ShardHolder>,
```
generall, line 278: required to avoid double read lock

lib/collection/src/collection/collection_ops.rs
```diff
@@ -212,11 +211,13 @@ impl Collection {
 
                 // ...and cancel transfer tasks and remove transfers from internal state
                 for transfer in transfers {
-                    self.finish_shard_transfer(transfer, Some(&shard_holder))
+                    self.abort_shard_transfer(transfer.key(), Some(&shard_holder))
                         .await?;
                 }
             }
 
+            replica_set.remove_peer(peer_id).await?;
```
generall, line 219: It feels safer to remove shard after the transfer is aborted

lib/collection/src/collection/collection_ops.rs
```diff
@@ -212,11 +211,13 @@ impl Collection {
 
                 // ...and cancel transfer tasks and remove transfers from internal state
                 for transfer in transfers {
-                    self.finish_shard_transfer(transfer, Some(&shard_holder))
+                    self.abort_shard_transfer(transfer.key(), Some(&shard_holder))
```
generall, line 214: No idea why it was finished before

timvisee (approved):
Looks good except for dropping the shard on transfer abort. See my review [comment](https://github.com/qdrant/qdrant/pull/5079#discussion_r1759077165).

lib/collection/src/collection/collection_ops.rs
generall: This happens if the "drop_replica" operation is requested. In this PR we don't introduce new logic of removing shard, we simply change form removing shard before cancelling transfers to after.

timvisee: I see now. Thanks!

lib/collection/src/collection/collection_ops.rs
```diff
@@ -194,14 +194,13 @@ impl Collection {
                 });
             }
 
-            if peers.len() == 1 {
+            // Check that we are not removing the
```
ffuugoo, line 197: It's just weird to not finish a comment like that 😂

```suggestion
            // Check that we are not removing the last active replica
```

ffuugoo (approved):

lib/collection/src/collection/collection_ops.rs
ffuugoo: My only concern would be if we are in the middle of upgrade, where some nodes run `finish_shard_transfer` and other nodes run `abort_shard_transfer`, but I guess we don't care?

Issue #5084, opened 2024-09-15
Error when loading large amounts of points with on_disk=True


Jack-ZC8:
Hello,

I encountered an issue when loading a large number of points with `on_disk=True`. The error displayed in the dashboard's optimizer_status is as follows:


**Service internal error: Optimization task panicked: called `Result::unwrap()` on an `Err` value: SizeExact(8, 4096)**

This error occurs when I repeatedly add points using the example from [Qdrant hybrid-search](https://qdrant.tech/documentation/tutorials/hybrid-search-fastembed/). Specifically, the error appears after adding points for the third time with the following configuration:

```json
{
  "params": {
    "vectors": {
      "fast-all-minilm-l6-v2": {
        "size": 384,
        "distance": "Cosine",
        "on_disk": true
      }
    },
    "shard_number": 1,
    "replication_factor": 1,
    "write_consistency_factor": 1,
    "on_disk_payload": true,
    "sparse_vectors": {
      "fast-sparse-splade_pp_en_v1": {
        "index": {
          "on_disk": true
        }
      }
    }
  },
  "hnsw_config": {
    "m": 16,
    "ef_construct": 100,
    "full_scan_threshold": 10000,
    "max_indexing_threads": 0,
    "on_disk": false
  },
  "optimizer_config": {
    "deleted_threshold": 0.2,
    "vacuum_min_vector_number": 1000,
    "default_segment_number": 0,
    "max_segment_size": null,
    "memmap_threshold": null,
    "indexing_threshold": 20000,
    "flush_interval_sec": 5,
    "max_optimization_threads": null
  },
  "wal_config": {
    "wal_capacity_mb": 32,
    "wal_segments_ahead": 0
  },
  "quantization_config": null
}
```
Currently, the status is as shown in the attached image.
![image](https://github.com/user-attachments/assets/63b341fd-1f28-4b46-aab8-e579efe088d7)

Interestingly, changing the vector storage to on_disk=False will not encounter the error.

Then I close the qdrant and open it again, I got the following log:
```
Version: 1.11.3, build: 9fa86106                                                                                              
Access web UI at http://localhost:6333/dashboard                                                                              
                                                                                                                              
2024-09-15T17:04:37.867187Z  WARN qdrant::settings: Config file not found: config/config                                      
2024-09-15T17:04:37.871841Z  WARN qdrant::settings: Config file not found: config/development                                 
[...]

PR #5088, opened 2024-09-16
mmap_to_type_unbounded -> mmap_prefix_to_type_unbounded

generall:
Related to: https://github.com/qdrant/qdrant/issues/5084

This PR fixes reading of the "corrupted" files, but doesn't prevent "corruption". 
So we assume that corrupted files are just declared normal.
(reason for extension of the file size with zeros is still unclear)

### DIFF
--- a/lib/common/memory/src/mmap_type.rs
+++ b/lib/common/memory/src/mmap_type.rs
@@ -99,7 +99,7 @@ where
     ///
     /// - panics when the size of the mmap doesn't match size `T`
     /// - panics when the mmap data is not correctly aligned for type `T`
-    /// - See: [`mmap_to_type_unbounded`]
+    /// - See: [`mmap_prefix_to_type_unbounded`]
     pub unsafe fn from(mmap_with_type: MmapMut) -> Self {
         Self::try_from(mmap_with_type).unwrap()
     }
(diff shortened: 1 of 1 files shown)


timvisee (approved):

lib/common/memory/src/mmap_type.rs
```diff
@@ -432,26 +434,36 @@ pub enum Error {
...
+    if mmap.len() != size_t {
+        log::warn!(
+            "Mmap length {} is not equal to size of type {}",
+            mmap.len(),
+            size_t,
+        );
+    }
+
+    #[cfg(debug_assertions)]
     if bytes.len() != mem::size_of::<T>() {
```
xzfc, line 467: This check never fails. If we want to return `Error::SizeExact` on debug builds, we need to adjust the check:
```suggestion
    if mmap.len() != mem::size_of::<T>() {
```

generall: Then it would be the same as a previous check. 
It is kinda redundant, but may serve explanation functionality

xzfc: After discussion, decided to keep it as is.

PR #5090, opened 2024-09-16
Clean temporary segments if optimization is cancelled

### LINKED ISSUE #2978: Temporary snapshots are not cleaned if the optimization if cancelled
Temporary snapshots are not cleaned if the optimization if cancelled

## Steps to Reproduce

1. Run optimization
2. Cancel it with, e.g. changing of the optimizations param (`PATCH /collections/<name>`)

Temporary segment is still in the `temp_segments` folder

## Expected Behavior

Clean `temp_segments`

dkellner:
This uses `TempDir` and comes with its caveats, e.g. the temporary segment directory will not be deleted if the process exits (see https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#resource-leaking). This should not be a problem in practice, as all temporary segments get deleted when the shard is loaded (see PR #2319).

To reproduce the issue I've created a collection with a sample dataset and repeatedly changed the HNSW configuration:

```
curl -X PATCH "$qdrant_url/collections/test" \
     -d '{"hnsw_config":{"m": 24}}'
curl -X PATCH "$qdrant_url/collections/test" \
     -d '{"hnsw_config":{"m": 32}}'
curl -X PATCH "$qdrant_url/collections/test" \
     -d '{"hnsw_config":{"m": 8}}'
```

Before applying this fix, this results in segment directories in `storage/collections/test/0/temp_segments` that will not get deleted until qdrant is restarted/the shard gets loaded again. After this fix, `temp_segments` is empty as expected.

Fixes #2978.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/segment_constructor/segment_builder.rs
+++ b/lib/segment/src/segment_constructor/segment_builder.rs
@@ -11,6 +11,7 @@ use bitvec::macros::internal::funty::Integral;
 use common::cpu::CpuPermit;
 use common::types::PointOffsetType;
 use io::storage_version::StorageVersion;
+use tempfile::TempDir;
 use uuid::Uuid;
 
 use super::{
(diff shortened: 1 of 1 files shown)


generall:
Great improvement, thanks @dkellner :pray:  !

generall (approved):

PR #5100, opened 2024-09-17
Facets: Support for `bool` payload 

coszio:
Adds support for boolean payloads with `bool` index in Facet API. It is a small change to fill a missing hole. 

Now it makes sense to say that any index which supports `MatchValue` conditions can be used for faceting.

cc @joein

### DIFF
--- a/lib/segment/src/index/field_index/binary_index.rs
+++ b/lib/segment/src/index/field_index/binary_index.rs
@@ -227,9 +227,13 @@ impl BinaryIndex {
         ]
         .into_iter()
     }
-    
+
     pub fn iter_counts_per_value(&self) -> impl Iterator<Item = (bool, usize)> + '_ {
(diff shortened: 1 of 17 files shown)


lib/segment/src/index/field_index/binary_index.rs
```diff
@@ -218,6 +219,22 @@ impl BinaryIndex {
...
+        ]
+        .into_iter()
+    }
+
+    pub fn iter_counts_per_value(&self) -> impl Iterator<Item = (bool, usize)> + '_ {
+        vec![
+            (false, self.memory.falses_count()),
+            (true, self.memory.trues_count()),
+        ]
+        .into_iter()
```
timvisee, line 236: Do we need to filter items with a count of 0 here?

coszio: It is already filtered later :)

timvisee (approved):

PR #5104, opened 2024-09-18
Remapping tags to api docs 

kartik-gupta-ij:
The remapping of tags in the  [openapi.json](https://github.com/qdrant/qdrant/blob/master/docs/redoc/master/openapi.json) to align with the updated tag structure as defined in the [API reference](https://github.com/qdrant/api-reference/blob/main/fern/api-description-overwrites.yml). The primary goal is to test compatibility with both the remapped and original tag versions in clients.


timvisee (approved):

generall:
Please merge only after we have a confirmation that this doesn't break clients

Anush008:
Note to use title case tags when adding APIs in the future. They are case-sensitive.

generall (approved):

PR #5107, opened 2024-09-18
Make BitVec portable, set fixed bit layout in memory

timvisee:
We make use of [`BitVec`](https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html) and its default parameters in many places throughout our code base.

The problem is that this definition is not portable. The container type and bit order are system dependent. Because we persist these bit vectors on disk and potentially transfer them to other machines, this may be problematic.

To resolve this, we now explicitly set our container type and bit order. That way we're sure the persisted data is consistent across all system configurations. To simplify usage of this, I've defined these explicit types centrally. See the review comment below.

Note that this assumes we've only been used on little endian systems thus far. Big endian systems, having different byte order, will break with this update. The idea is that we better patch and use a fixed configuration now, versus letting this escalate further beyond control.

One interesting performance quirk is that using the native bit/byte ordering on the platform likely will be faster at runtime. Because of this I've also exposed `MemoryBitVec`. It is intended to use in places where we never persist the bit vector, as portability is not needed there. My plan is to apply this type in more places where it seems fit, but in a separate PR.

<https://github.com/qdrant/qdrant/pull/5180> goes along with this one to potentially help debugging remote systems.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/common/common/src/bitvec.rs
+++ b/lib/common/common/src/bitvec.rs
@@ -1,0 +1,11 @@
+use bitvec::slice::BitSlice as RawBitSlice;
+pub use bitvec::store::BitStore;
+pub use bitvec::vec::BitVec as RawBitVec;
+
+// BitVec type we use across the codebase. Has consistent bit layout, suitable for persisting.
+pub type BitSliceElement = u64;
+pub type BitSlice = RawBitSlice<BitSliceElement, bitvec::order::Lsb0>;
+pub type BitVec = RawBitVec<BitSliceElement, bitvec::order::Lsb0>;
+
+// BitVec type we use for in-memory bit vectors. Can be more efficient than BitVec.
+pub type MemoryBitVec = bitvec::vec::BitVec;
--- a/lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
+++ b/lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
@@ -4,11 +4,11 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::{fmt, fs};
 
+use common::bitvec::{BitSlice, BitSliceElement};
 use memmap2::MmapMut;
 use memory::madvise::{self, AdviceSetting, Madviseable as _};
 use memory::mmap_ops::{create_and_ensure_length, open_write_mmap};
-use memory::mmap_type::BitSlice;
-use memory::mmap_type::{BitSliceElement, MmapBitSlice, MmapFlusher, MmapType};
+use memory::mmap_type::{MmapBitSlice, MmapFlusher, MmapType};
 use parking_lot::Mutex;
 
 use crate::common::operation_error::{OperationError, OperationResult};
--- a/lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
+++ b/lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
@@ -4,11 +4,11 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::{fmt, fs};
 
-use bitvec::prelude::BitSlice;
 use memmap2::MmapMut;
 use memory::madvise::{self, AdviceSetting, Madviseable as _};
 use memory::mmap_ops::{create_and_ensure_length, open_write_mmap};
-use memory::mmap_type::{MmapBitSlice, MmapFlusher, MmapType};
+use memory::mmap_type::BitSlice;
+use memory::mmap_type::{BitSliceElement, MmapBitSlice, MmapFlusher, MmapType};
 use parking_lot::Mutex;
 
 use crate::common::operation_error::{OperationError, OperationResult};
@@ -68,9 +68,14 @@ impl fmt::Debug for DynamicMmapFlags {
 
 /// Based on the number of flags determines the size of the mmap file.
 fn mmap_capacity_bytes(num_flags: usize) -> usize {
-    let number_of_bytes = num_flags.div_ceil(u8::BITS as usize);
+    let mut number_of_bytes = num_flags.div_ceil(u8::BITS as usize);
 
-    max(MINIMAL_MMAP_SIZE, number_of_bytes.next_power_of_two())
+    // Exponentially grow file
+    number_of_bytes = max(MINIMAL_MMAP_SIZE, number_of_bytes.next_power_of_two());
+
+    // Bit slice uses multi-byte type as backing storage, so we must use a multiple of it
+    let align_bytes = BitSliceElement::BITS / u8::BITS;
+    number_of_bytes.next_multiple_of(align_bytes as usize)
 }
 
 /// Based on the current length determines how many flags can fit into the mmap file without resizing it.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1157,7 +1157,6 @@ dependencies = [
  "arc-swap",
  "async-trait",
  "atomicwrites",
- "bitvec",
  "bytes",
  "cancel",
  "chrono",
@@ -1232,6 +1231,7 @@ dependencies = [
 name = "common"
 version = "0.0.0"
 dependencies = [
+ "bitvec",
  "common",
  "criterion",
  "lazy_static",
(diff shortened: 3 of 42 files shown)


lib/common/common/src/bitvec.rs
```diff
@@ -0,0 +1,11 @@
...
+pub use bitvec::store::BitStore;
+pub use bitvec::vec::BitVec as RawBitVec;
+
+// BitVec type we use across the codebase. Has consistent bit layout, suitable for persisting.
+pub type BitSliceElement = u64;
+pub type BitSlice = RawBitSlice<BitSliceElement, bitvec::order::Lsb0>;
+pub type BitVec = RawBitVec<BitSliceElement, bitvec::order::Lsb0>;
+
+// BitVec type we use for in-memory bit vectors. Can be more efficient than BitVec.
+pub type MemoryBitVec = bitvec::vec::BitVec;
```
timvisee, line 11: Explicitly specify container type and bit order of `BitVec` and `BitSlice` types in a central place here.

lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
```diff
@@ -68,9 +68,14 @@ impl fmt::Debug for DynamicMmapFlags {
...
+    let mut number_of_bytes = num_flags.div_ceil(u8::BITS as usize);
 
-    max(MINIMAL_MMAP_SIZE, number_of_bytes.next_power_of_two())
+    // Exponentially grow file
+    number_of_bytes = max(MINIMAL_MMAP_SIZE, number_of_bytes.next_power_of_two());
+
+    // Bit slice uses multi-byte type as backing storage, so we must use a multiple of it
+    let align_bytes = BitSliceElement::BITS / u8::BITS;
+    number_of_bytes.next_multiple_of(align_bytes as usize)
 }
```
timvisee, line 79: Update mmap capacity bytes calculation here to ensure we always fit our `BitSlice` container type.

xzfc: Nit: I think it is not strictly unnecessary because of the previous line.
```rust
number_of_bytes = max(MINIMAL_MMAP_SIZE, number_of_bytes.next_power_of_two());
```
It will be always one of `0x100000`, `0x200000`, `0x400000`, `0x800000`, and so on, which is sure divisible by `size_of<usize>()`.

But if we want to be extra sure/explicit, I suggest to align to the same size (e.g. 8 bits) on all systems unconditionally.
So it won't be a problem when a file got created on a 32-bit machine and then transferred on 64-bit one.

lib/common/common/src/bitvec.rs
xzfc: An idea: It would be helpful if the type system would prevent converting `&[u8]`/mmaps into this type and vice versa.

timvisee:
> I might be wrong, but I think `BitVec<u64, Lsb0>` is still not compatible between little- and big-endian systems in the same sense as `&[u64]` isn't compatible. If so, I guess this PR won't change behavior on them.

You may be right. I believe the crate is container type agnostic and so it doesn't matter, but now I'm not so sure anymore.

I suggest to draft this for now and take a proper look into it on Monday. I think we would benefit from a test here too, that asserts we read raw bytes the way we expect.

timvisee:
The current bitorder and bitstore is indeed problematic.

What we need is `<u8, Msb0>`. However, this is not compatible with our current persisted data and would require conversion. That's really painful, and I'm not sure if we're ever able to do this migration properly.

See: https://ferrilab.github.io/bitvec/memory-representation.html#little-endian-byte-ordered-machines

lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
timvisee: You're right. But I've just been cautious here in case we decrease the minimum size.

> But if we want to be extra sure/explicit, I suggest to align to the same size (e.g. 8 bytes) on all systems unconditionally.
So it won't be a problem when a file got created on a 32-bit machine and then transferred on 64-bit one.

This was currently achieved by using the u64 bit store type explicitly.

However, as it turns out we might need to use `u8` instead in which case this extra alignment isn't needed anymore.

ffuugoo:
I think, alternatively, we can introduce custom [BitStore](https://docs.rs/bitvec/latest/bitvec/store/trait.BitStore.html) implementation that will be layout-compatible with `u64` on LE systems? Seems _a bit_ annoying to do, but I think it should not be terribly complicated to implement? We could probably copy most of the implementation from `impl BitStore for u64` from the crate, just replace `u64` with `BackwardCompatibleStore([u8; 8])` and ensure that we index same way as `u64` would be indexed on LE system?

xzfc:
Reiterating my suggestion that I did in private.

We can switch to `BitVec<u8, Lsb0>`. It implements the exact same layout as our current `BitVec<usize, Lsb0>` (when on 64-bit LE systems such as `x86_64`), so no conversion steps are required for major platforms.

For the sake of performance, we can still use `BitVec<u64, Lsb0>` on major platforms and fall back to `BitVec<u8, Lsb0>` for portability on exotic systems.

In code, 
```rust
// Our current approach: non-portable, fast.
type BitVec = bitvec::vec::BitVec<usize, Lsb0>;

// Approach 1: portable, slow.
// On x86_64 compatible with our current approach.
type BitVec = bitvec::vec::BitVec<u8, Lsb0>;

// Approach 2: same as the approach 1, but special-case on x86_64 for the sake of performance.
// Portable, fast on x86_64, slow on ppc64.
#[cfg(all(target_endian = "little", target_pointer_width="64"))]
type BitVec = bitvec::vec::BitVec<u64, Lsb0>;
#[cfg(not(all(target_endian = "little", target_pointer_width="64")))]
type BitVec = bitvec::vec::BitVec<u8, Lsb0>;
```

### Proof

This snippet prints a layout of all possible `BitStore`/`BitOrder` combinations. It could be cross-compiled using [`cross`](https://github.com/cross-rs/cross) for for powerpc64 and ran using `qemu-ppc64` (user space emulator, aka runs a single binary, no full powerpc64 installation required).

<details><summary>code</summary>

```rust
use std::any::type_name;

use bitvec::order::{BitOrder, Lsb0, Msb0};
use bitvec::store::BitStore;
use bitvec::vec::BitVec;

fn main() {
    foo::<u8, Lsb0>();
    foo::<u8, Msb0>();
    foo::<u64, Lsb0>();
    foo::<u64, Msb0>();
}

fn foo<T: BitStore, O: BitOrder>() {
    let primes =
        [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61];

    let mut v = BitVec::<T, O>::repeat(false, 64);
    for i in &primes {
        v.set(*i, true);
    }

    let slice_t = v.as_raw_slice();
    let slice_u8 = unsafe {
        std::slice::from_raw_parts(
            slice_t.as_ptr() as *const u8,
            slice_t.len() * size_of::<T>(),
        )
    };

    for i in slice_u8 {
        print!("{:02x} ", i);
    }
    println!(
        " {:>3} {} {}",
        type_name::<T>(),
        type_name::<O>().split("::").last().unwrap(),
        std::env::consts::ARCH,
    );
}
```

</details>

Results for `x86_64` and `powerpc64`, sorted and grouped:
```
14 04 51 04 05 51 14 35  u64 Msb0 x86_64

28 20 8a 20 a0 8a 28 ac  u64 Lsb0 powerpc64

35 14 51 05 04 51 04 14   u8 Msb0 powerpc64
35 14 51 05 04 51 04 14   u8 Msb0 x86_64
35 14 51 05 04 51 04 14  u64 Msb0 powerpc64

ac 28 8a a0 20 8a 20 28   u8 Lsb0 powerpc64   # !
ac 28 8a a0 20 8a 20 28   u8 Lsb0 x86_64      # !
ac 28 8a a0 20 8a 20 28  u64 Lsb0 x86_64      # !
```

The last group is what we are using right now on `x86_64`.

timvisee:
Closing this since this currently doesn't provide a clear path forward while already having a lot of conflicts.

Let's revisit this idea a bit later.

PR #5111, opened 2024-09-19
Fix race condition getting optimizers stuck, leaving collection in yellow state

timvisee:
This PR fixes a race condition which would allow the optimizers for a collection to get stuck. The result is a collection that remains in yellow state forever, unless a new update operation is send to trigger the remaining optimizations.

The PR has two changes:
1. patch race condition (with an extra branch)
2. remove a level of nesting in the match statement. The bulk of the logic now happens after the match rather than inside it, other branches use `continue`/`break`.

The data race is caused by an ordering problem between three things:
1. condition that checks max allowed handles (max 1 in this case)
2. clean up of handles that are finished
3. signal at the end of optimization to trigger new optimizations

Basically, we trigger the optimizers again before cleaning the finished handle up (through a callback). If checking the maximum allowed handles on a new run, we'd still be at the maximum, cancelling the new optimization. There are no other ongoing optimizations to trigger the optimizer loop once again, getting it stuck. In other words, the problem is that we retrigger optimizations before cleaning up the handles. Sadly that's not an easy thing to fix.

This PR patches the problem by triggering optimizations even if we clean up any finished handles. This means we pass the handle limit condition, which prevents it getting stuck. We have tested this patch on our benchmark server. It has shown to solve the problem.

I'm not super satisfied with the way the optimization loops works right now. I suggest to merge this PR, and attempt to refactor it separately to make this data race impossible.

<details>
<summary>Details</summary>

We did not catch this before because we have no optimization handle limit by default. In case of our benchmark server we have a limit of 1 set however.

Detailed scenario, in order:
1. Thread 1: one optimization is running
2. Thread 1: optimization is at the end, sends optimizer signal
3. Thread 2: optimization worker triggered, received signal
4. Thread 2: optimization worker fails condition, already at max of 1 handle
5. Thread 1: optimization is finished
6. _some time passes_
7. Thread 2: optimization worker triggered by clean interval
8. Thread 2: optimization worker cleans finished handle
9. _now we're stuck, we won't receive new signals_

This PR patches point 8.

</details>

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/update_handler.rs
+++ b/lib/collection/src/update_handler.rs
@@ -495,9 +495,11 @@ impl UpdateHandler {
     /// It also propagates any panics (and unknown errors) so we properly handle them if desired.
     ///
     /// It is essential to call this every once in a while for handling panics in time.
+    ///
+    /// Returns true if any optimization handle was finished, joined and removed.
     async fn cleanup_optimization_handles(
         optimization_handles: Arc<TokioMutex<Vec<StoppableTaskHandle<bool>>>>,
-    ) {
+    ) -> bool {
         // Remove finished handles
         let finished_handles: Vec<_> = {
             let mut handles = optimization_handles.lock().await;
@@ -510,10 +512,14 @@ impl UpdateHandler {
                 .collect()
         };
 
+        let finished_any = !finished_handles.is_empty();
+
         // Finalize all finished handles to propagate panics
         for handle in finished_handles {
             handle.join_and_handle_panic().await;
         }
+
+        finished_any
     }
 
     #[allow(clippy::too_many_arguments)]
--- a/lib/collection/src/update_handler.rs
+++ b/lib/collection/src/update_handler.rs
@@ -548,10 +548,22 @@ impl UpdateHandler {
 
         loop {
             let receiver = timeout(OPTIMIZER_CLEANUP_INTERVAL, receiver.recv());
-            let result = receiver.await;
+            let mut result = receiver.await;
 
             // Always clean up on any signal
-            Self::cleanup_optimization_handles(optimization_handles.clone()).await;
+            let cleaned_any =
+                Self::cleanup_optimization_handles(optimization_handles.clone()).await;
+
+            // If we cleaned any optimization handles, always trigger Nop
+            // TODO(timvisee): implement this in a better way
+            if let Err(Elapsed { .. }) = &result {
+                if cleaned_any {
+                    log::warn!(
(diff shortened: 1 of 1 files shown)


tellet-q (approved):
lgtm, we ran a test to check if the log line appears and it did

lib/collection/src/update_handler.rs
```diff
@@ -506,14 +508,18 @@ impl UpdateHandler {
                 .collect::<Vec<_>>()
                 .into_iter()
                 .rev()
-                .map(|i| handles.remove(i))
+                .map(|i| handles.swap_remove(i))
```
agourlay, line 511: Using `swap_removes` means losing ordering of handles.
Just want to confirm that it is not something we care about :)

lib/collection/src/update_handler.rs
```diff
@@ -541,95 +547,110 @@ impl UpdateHandler {
...
             }
+
+            // Determine optimization handle limit based on max handles we allow
+            // Not related to the CPU budget, but a different limit for the maximum number
+            // of concurrent concrete optimizations per shard as configured by the user in
+            // the Qdrant configuration.
+            // Skip if we reached limit, an ongoing optimization that finishes will trigger this loop again
+            let limit = max_handles.saturating_sub(optimization_handles.lock().await.len());
+            if limit == 0 {
+                log::trace!("Skipping optimization check, we reached optimization thread limit");
```
agourlay, line 639: I never use `trace` because it is way too verbose.
Would it make sense to bump this one to `debug`?

lib/collection/src/update_handler.rs
timvisee: It doesn't matter, but thanks for confirming!

We only keep the handles to count the tasks that are still ongoing, and to catch any panics. Tasks themselves run asynchronously in their own thread. We don't care in what order or at what time we clean them up.

lib/collection/src/update_handler.rs
timvisee: It can happen a lot and cause a lot of log spam in debug. A lot as in, with every update operation you send. That's why we demoted it a long time ago to trace.

It actually was useful in our benchmarking setup a few days ago where we changed logging to trace just to confirm some branches were hit.

Do you think setting it to debug would be worth the log spam?

agourlay: Let's keep it to `trace` if it is noisy :+1:

agourlay (approved):
Thank you very much for the care put in the explanation and thought process.

This LGTM but not gonna lie the state machine is getting difficult to grasp.

KShivendu (approved):

PR #5112, opened 2024-09-20
Update `tar_ext::BuilderExt`

xzfc:
This PR updates `tar_ext::BuilderExt` with the following changes:

1. Generalize: instead of taking `std::fs::File`, take `impl Write + Seek`. Later this requirement will be reduced further to `impl Write` to support both `File` and an async stream (`SyncIoBridge`).
2. Add a workaround for drop: `tar::Builder<W>::drop` might panic if the underlying writer panics when using in async context. See a comment for `BuilderBox` and newly added unit tests for more details on the issue.

### DIFF
--- a/lib/common/common/src/tar_ext.rs
+++ b/lib/common/common/src/tar_ext.rs
@@ -1,10 +1,11 @@
 //! Extensions for the `tar` crate.
 
-use std::fs::File;
-use std::io::Write;
+use std::io::{Seek, Write};
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
+use tap::Tap;
 use tokio::sync::Mutex;
 use tokio::task::JoinError;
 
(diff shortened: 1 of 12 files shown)


lib/common/common/src/tar_ext.rs
```diff
@@ -13,16 +14,92 @@ use tokio::task::JoinError;
...
+
+impl Drop for BuilderBox {
+    fn drop(&mut self) {
+        self.enabled.store(false, Ordering::Relaxed);
+    }
+}
+
+impl Write for WriteBox {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        if !self.enabled.load(Ordering::Relaxed) {
```
timvisee, line 61: Don't we need release/acquire here to prevent reordering incorrectly?

timvisee, line 61: Do I understand it correctly: the tar impl will attempt to write the final 1K zero bytes. But this write will be rejected here if we unset the enabled flag, preventing the panic?

If we explicitly finalize ourselves the tar impl doesn't write anything and we don't hit this branch at all (during drop).

xzfc: Yes, that's correct.

generall (approved):

PR #4855, opened 2024-08-09
Storage improvement

JojiiOfficial:
### All Submissions:

* [ ] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [ ] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [ ] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #5113, opened 2024-09-20
Fix set payload by key on non appendable segment

agourlay:
fixes: https://github.com/qdrant/qdrant-client/issues/780

The bug affects set payload operation using `key` on non appendable segments.

It manifest itself by the apparition of unexpected payload values at the root level, which are supposed to be added under `key`.

The bug was introduced in [1.11.1](https://github.com/qdrant/qdrant/releases/tag/v1.11.1) with https://github.com/qdrant/qdrant/pull/4855/files#diff-fe6fcf41822e7bcc71c22cfb9067f6e80e36228f69d2727724e4cc51abc12e07R193

The fix is to merge the payloads using the `key`.

I have tested the fix on a user snapshot file.

### DIFF
--- a/lib/collection/src/collection_manager/segments_updater.rs
+++ b/lib/collection/src/collection_manager/segments_updater.rs
@@ -190,7 +190,10 @@ pub(crate) fn set_payload(
             op_num,
             chunk,
             |id, write_segment| write_segment.set_payload(op_num, id, payload, key),
-            |_, _, old_payload| old_payload.merge(payload),
+            |_, _, old_payload| match key {
+                Some(key) => old_payload.merge_by_key(payload, key),
+                None => old_payload.merge(payload),
+            },
             |segment| {
                 segment.get_indexed_fields().keys().all(|indexed_path| {
                     !indexed_path.is_affected_by_value_set(&payload.0, key.as_ref())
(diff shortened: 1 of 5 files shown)


lib/collection/src/collection_manager/segments_updater.rs
```diff
@@ -190,7 +190,10 @@ pub(crate) fn set_payload(
             op_num,
             chunk,
             |id, write_segment| write_segment.set_payload(op_num, id, payload, key),
-            |_, _, old_payload| old_payload.merge(payload),
+            |_, _, old_payload| match key {
```
agourlay, line 193: the fix

lib/segment/src/types.rs
```diff
@@ -986,9 +986,8 @@ impl Payload {
         utils::merge_map(&mut self.0, &value.0)
     }
 
-    pub fn merge_by_key(&mut self, value: &Payload, key: &JsonPath) -> OperationResult<()> {
+    pub fn merge_by_key(&mut self, value: &Payload, key: &JsonPath) {
```
agourlay, line 989: removing the unnecessary `Result` type to align with the expected closure

timvisee (approved):

JojiiOfficial (commented):
Sorry for that one. Missed the separate `merge_by_key` function.

agourlay:
Merging the fix to expedite the release.

I will work on a unit test to enforce non regression in a different PR :memo:

agourlay:
Important information, the bug is present only when there is an indexed on the target payload key.
This is make it a bit less widespread I guess.

Issue #5114, opened 2024-09-23
Failed to populate memmap


ava57r:
Migration existing collection from v1.10.1 to v1.11.4

## Current Behavior
I see errors into log about memmap when I upgrade qdrant version from v1.10.1 to v1.11.4

## Steps to Reproduce
1. I have collection for testing updates v1.10.1
2. I just update bin qdrant to v1.11.4

## Context (Environment)

OS: Oracle Linux 8.6
rust version: rustc 1.81.0

LOG:

```
Sep 23 10:46:20 : INFO collection::shards::local_shard: Recovering collection test-test: 0/1 (0%)
Sep 23 10:46:20 : INFO collection::shards::local_shard: Recovered collection test-test: 1/1 (100%)
Sep 23 10:46:20 : WARN memory::mmap_ops: Failed to populate mmap "./storage/collections/test-test/3/segments/32bc911f-23ec-40a9-8f55-cf3cd33722fb/vector_index" to cache: Invalid argument (os error 22)
Sep 23 10:46:20 : WARN memory::mmap_ops: Failed to populate mmap "./storage/collections/test-test/3/segments/32bc911f-23ec-40a9-8f55-cf3cd33722fb/vector_storage/matrix.dat" to cache: Invalid argument (os error 22)
```

ava57r:
OL8.6 has kernel `kernel-uek-5.4.17-2136.307.3`

>= OL8.7 has `kernel-uek-5.15.0-3.60.5.1`

timvisee:
Thanks!

Would it be possible for you to update to OL8.7 and see if the issue persists? Or do you prefer to stick with OL8.6 instead?

timvisee:
We just release [Qdrant 1.11.5](https://github.com/qdrant/qdrant/releases/tag/v1.11.5) with a hotfix for this problem.

Though it'll still show a warning message similar to what you see above, it'll fall back to populate the memmap pages using a different method.

We plan to release a proper fix a bit later, which will also remove the warnings.

ava57r:
> Thanks!
> 
> Would it be possible for you to update to OL8.7 and see if the issue persists? Or do you prefer to stick with OL8.6 instead?

I upgraded it to OL8.7
Now It is working without warrings.

timvisee:
Fantastic! Thanks for being flexible :smile:

timvisee:
[...]

PR #5123, opened 2024-09-23
Do not return error if madvise(PopulateRead) fails

xzfc:
Fixup for #5114.


lib/common/memory/src/madvise.rs
```diff
@@ -112,13 +112,15 @@ impl Madviseable for memmap2::Mmap {
...
     }
 
-    fn populate(&self) -> io::Result<()> {
+    fn populate(&self) {
         #[cfg(target_os = "linux")]
         if *POPULATE_READ_IS_SUPPORTED {
-            return self.advise(memmap2::Advice::PopulateRead);
+            match self.advise(memmap2::Advice::PopulateRead) {
+                Ok(()) => return,
+                Err(err) => log::warn!("Failed to populate with MADV_POPULATE_READ: {err}"),
```
timvisee, line 120: Maybe it's worth to describe it's falling back:

```suggestion
                Err(err) => log::warn!("Failed to populate with MADV_POPULATE_READ, falling back to naive approach: {err}"),
```

Same goes for the other instance.

What do you think?

xzfc: Updated.

timvisee: Thanks!

timvisee (approved):

PR #5124, opened 2024-09-23
[Strict mode] Unit tests

JojiiOfficial:
Depends on #5120

### DIFF
--- a/lib/collection/src/operations/verification/mod.rs
+++ b/lib/collection/src/operations/verification/mod.rs
@@ -339,6 +339,12 @@ mod test {
             exact: true,
         };
         assert_strict_mode_error(request, collection).await;
+
+        let request = CountRequestInternal {
+            filter: None,
+            exact: false,
+        };
+        assert_strict_mode_success(request, collection).await;
     }
 
     async fn assert_strict_mode_error<R: StrictModeVerification>(
--- a/lib/collection/src/operations/verification/mod.rs
+++ b/lib/collection/src/operations/verification/mod.rs
@@ -251,3 +251,184 @@ impl StrictModeVerification for SearchParams {
         None
     }
 }
+
+#[cfg(test)]
+mod test {
+    use std::sync::Arc;
+
+    use common::cpu::CpuBudget;
+    use segment::types::{
+        Condition, FieldCondition, Filter, Match, SearchParams, StrictModeConfig,
+    };
+    use tempfile::Builder;
+
+    use super::StrictModeVerification;
+    use crate::collection::{Collection, RequestShardTransfer};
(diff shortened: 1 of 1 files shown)


lib/collection/src/operations/verification/mod.rs
```diff
@@ -251,3 +251,184 @@ impl StrictModeVerification for SearchParams {
...
+    use crate::optimizers_builder::OptimizersConfig;
+    use crate::shards::channel_service::ChannelService;
+    use crate::shards::collection_shard_distribution::CollectionShardDistribution;
+    use crate::shards::replica_set::{AbortShardTransfer, ChangePeerState};
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_strict_mode_verification_trait() {
+        let collection = fixture().await;
+
+        test_query_limit(&collection).await;
```
JojiiOfficial, line 282: The timeout function seems to be obsolete, and completely replaced by the timeout parameter in `check_strict_mode`. Therefore there is no test for timeout here.

I've kept the function in the trait, since we might need it in future.
If you think we should rather remove it and add it back in case we need it, I can do that in a following PR.

timvisee: > If you think we should rather remove it and add it back in case we need it, I can do that in a following PR.

If we don't use it, let's remove it. We can always put it back later.

Just to be clear, we do validate it as part of strict mode, but at the endpoint level. Not on the internal structures through the trait.

lib/collection/src/operations/verification/mod.rs
```diff
@@ -251,3 +251,184 @@ impl StrictModeVerification for SearchParams {
...
+        test_search_params(&collection).await;
+        test_filter_read(&collection).await;
+        test_filter_write(&collection).await;
+        test_request_exact(&collection).await;
+    }
+
+    async fn test_query_limit(collection: &Collection) {
+        assert_strict_mode_error(discovery_fixture(Some(10), None, None), collection).await;
```
timvisee, line 290: Shall we also assert a good example here, rather than just a bad example? (also for the others)

lib/collection/src/operations/verification/mod.rs
timvisee: > If you think we should rather remove it and add it back in case we need it, I can do that in a following PR.

If we don't use it, let's remove it. We can always put it back later.

Just to be clear, we do validate it as part of strict mode, but at the endpoint level. Not on the internal structures through the trait.

JojiiOfficial: I'll do this in a following PR since there is no overlap with unit tests.

timvisee (approved):

PR #5132, opened 2024-09-23
Add unit test for set_payload bug

agourlay:
This PR adds a non regression test for https://github.com/qdrant/qdrant/pull/5113

The conditions are:
- set_payload op
- targeting nested key
- key must be indexed
- the point must be on a non appendable segment


JojiiOfficial (approved):

timvisee (approved):

timvisee:
Thanks for taking care of it!

PR #5137, opened 2024-09-24
[Strict mode] Merge StrictMode config types

JojiiOfficial:
Depends on #5124 

Merges `StrictModeConfig` and `StrictModeConfigDiff` into a single struct, as they have the same fields.

### DIFF
--- a/lib/segment/Cargo.toml
+++ b/lib/segment/Cargo.toml
@@ -99,7 +99,7 @@ macro_rules_attribute = "0.2.0"
 generic-tests = { workspace = true }
 nom = "7.1.3"
 half = { workspace = true }
-merge = "0.1.0"
+merge = { workspace = true }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 cgroups-rs = "0.3"
--- a/lib/segment/Cargo.toml
+++ b/lib/segment/Cargo.toml
@@ -99,6 +99,7 @@ macro_rules_attribute = "0.2.0"
 generic-tests = { workspace = true }
 nom = "7.1.3"
 half = { workspace = true }
+merge = "0.1.0"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 cgroups-rs = "0.3"
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -716,17 +716,19 @@ impl Hash for StrictModeConfig {
             unindexed_filtering_update,
             search_max_hnsw_ef,
             search_allow_exact,
-            search_max_oversampling,
+            // We skip hashing this field because we cannot reliably hash a float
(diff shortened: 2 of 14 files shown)


lib/segment/Cargo.toml
```diff
@@ -99,6 +99,7 @@ macro_rules_attribute = "0.2.0"
 generic-tests = { workspace = true }
 nom = "7.1.3"
 half = { workspace = true }
+merge = "0.1.0"
```
timvisee, line 102: Since we now have this dependency in multiple places, could you promote it to a workspace dependency?

lib/segment/src/types.rs
```diff
@@ -705,6 +706,30 @@ pub struct StrictModeConfig {
...
+        enabled.hash(state);
+        max_query_limit.hash(state);
+        max_timeout.hash(state);
+        unindexed_filtering_retrieve.hash(state);
+        unindexed_filtering_update.hash(state);
+        search_max_hnsw_ef.hash(state);
+        search_allow_exact.hash(state);
+        search_max_oversampling.map(|i| i.to_le_bytes()).hash(state);
+    }
+}
```
timvisee, line 731: Can't we derive hash, or does `search_max_oversampling` make life difficult?

Otherwise we might want to not hash it to prevent float rounding errors from breaking hashes.

timvisee (changes requested):
I do think this is better, having just a single type!

lib/segment/src/types.rs
JojiiOfficial: Yes `search_max_oversampling` is the problem here.
We need `StrictModeConfig` to implement `Hash`, since it's [a field in this struct](https://github.com/qdrant/qdrant/blob/strict_mode_test/lib/storage/src/content_manager/collection_meta_ops.rs#L172), which derives `Hash`.

timvisee: Check.

Let's skip hashing `search_max_oversampling` because we cannot reliably do that:

```suggestion
impl Hash for StrictModeConfig {
    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
        let Self {
            enabled,
            max_query_limit,
            max_timeout,
            unindexed_filtering_retrieve,
            unindexed_filtering_update,
            search_max_hnsw_ef,
            search_allow_exact,
            // We skip hashing this field because we cannot reliably hash a float
            search_max_oversampling: _,
        } = self;

        (
            enabled,
            max_query_limit,
            max_timeout,
            unindexed_filtering_retrieve,
            unindexed_filtering_update,
            search_max_hnsw_ef,
            search_allow_exact,
        )
            .hash(state);
    }
}
```

timvisee (approved):

PR #5140, opened 2024-09-24
Trigger optimizers when uploading snapshots

### LINKED ISSUE #5139: Optimizers not triggered when recovering unoptimized snapshot
## Current Behavior
If recovering an unoptimized snasphot, optimizers will not be triggered. Instead the collection will remain in grey status until an [update](https://qdrant.tech/documentation/concepts/collections/#grey-collection-status) is sent.

## Steps to Reproduce
Preparation, create an unoptimized snapshot:

1. create a collection with `bfb -n 1000 --indexing-threshold 1`
2. when optimizations are started, restart the Qdrant instance
3. the collection is now has a grey status
4. create a collection snapshot
5. you now have an unoptimized snapshot

Reproduce, recover unoptimized snapshot:

6. recover the unoptimized snapshot
7. the collection will remain in grey status

## Expected Behavior
When recovering a snapshot optimizers should be triggered, so that we never stick with a grey status.

That means that after recovery a collection should always have either yellow or green status. Eventually it must always become green.

A user could always explicitly [trigger](https://qdrant.tech/documentation/concepts/collections/#grey-collection-status) it, but to me it is more reasonable to do this for them.

## Possible Solution

Explicitly trigger optimizers with a `Nop` signal once a collection snapshot is recovered.

## Detailed Description

There's multiple ways to get to the same problem.

Some parameters that affect optimizations have default values that are system dependent, like `default_segment_number`:

https://github.com/qdrant/qdrant/blob/dee106d74625e891256ebe06c2c2e0515650e67e/config/config.yaml#L102-L110

That means that if a collection is green one one system. It may need to be optimized again on a different system having a different system configuration. To me that makes sense.

For example:
1. create collection on 32 CPU machine, wait until it is green
2. snapshot collection
3. recover snapshot on 2 CPU machine
4. collection will be in grey status, we merge to lower segment number due to less CPUs

JojiiOfficial:
Fixes #5139

### DIFF
--- a/lib/collection/src/shards/proxy_shard.rs
+++ b/lib/collection/src/shards/proxy_shard.rs
@@ -83,8 +83,9 @@ impl ProxyShard {
         self.wrapped_shard.on_optimizer_config_update().await
     }
 
-    pub async fn trigger_optimizers(&self) -> CollectionResult<()> {
-        self.wrapped_shard.trigger_optimizers().await
+    pub fn trigger_optimizers(&self) {
+        // TODO: we might want to defer this trigger until we unproxy
+        self.wrapped_shard.trigger_optimizers();
     }
 
     pub async fn reinit_changelog(&self) -> CollectionResult<()> {
--- a/lib/collection/src/shards/proxy_shard.rs
+++ b/lib/collection/src/shards/proxy_shard.rs
@@ -83,6 +83,10 @@ impl ProxyShard {
         self.wrapped_shard.on_optimizer_config_update().await
     }
 
+    pub async fn trigger_optimizers(&self) -> CollectionResult<()> {
+        self.wrapped_shard.trigger_optimizers().await
+    }
+
     pub async fn reinit_changelog(&self) -> CollectionResult<()> {
         // Blocks updates in the wrapped shard.
         let mut changed_points_guard = self.changed_points.write().await;
--- a/lib/collection/src/shards/shard_holder/mod.rs
+++ b/lib/collection/src/shards/shard_holder/mod.rs
@@ -63,12 +63,10 @@ pub struct ShardHolder {
 pub type LockedShardHolder = RwLock<ShardHolder>;
 
 impl ShardHolder {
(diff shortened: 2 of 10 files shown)


lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -63,6 +63,14 @@ pub struct ShardHolder {
 pub type LockedShardHolder = RwLock<ShardHolder>;
 
 impl ShardHolder {
+    pub async fn trigger_optimizers(&self) -> CollectionResult<()> {
+        for shard in self.shards.values() {
+            shard.trigger_optimizers().await?;
```
JojiiOfficial, line 68: I'm not sure if we really need to call this for each shard to be honest.

timvisee (changes requested):

lib/collection/src/shards/shard_holder/mod.rs
timvisee: Yes.

I mistakenly described the problem at collection level. But you're right. Each shard has their own optimizers so we need to trigger all of them.

lib/collection/src/shards/proxy_shard.rs
```diff
@@ -83,6 +83,10 @@ impl ProxyShard {
         self.wrapped_shard.on_optimizer_config_update().await
     }
 
+    pub async fn trigger_optimizers(&self) -> CollectionResult<()> {
+        self.wrapped_shard.trigger_optimizers().await
```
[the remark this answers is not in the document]
timvisee: It turns out that we don't actually use it right now, and we have no place where we actually apply collected updates.

Let's keep it like you had but hint to this idea with a TODO:

```suggestion
        // TODO: we might want to defer this trigger until we unproxy
        self.wrapped_shard.trigger_optimizers().await
```

I think we'll eventually remove this proxy all together.

timvisee (approved):

PR #5141, opened 2024-09-25
When initializing new cluster, propose origin peer to consensus

### LINKED ISSUE #5138: Consensus: origin node not registered in consensus log, breaks if origin is removed
Joining a new peer to consensus is broken if the first peer is removed from the cluster. Adding new peers currently depend on the first node still being a member, which means we cannot remove the first peer until this is fixed.

## Current Behavior

Once the first peer is removed from the cluster, joining a new peer becomes problematic. Though it looks like the new peer joins without problems, the cluster state may be left in a corrupt state.

Note that bootstrapping from any peer is fine, as long as the first peer still exists. So the problem is not about the bootstrapping target, but rather about whether the original peer is available while bootstrapping.

## Steps to Reproduce

1. create a 2 peer cluster
2. create a collection with two shards
3. move the shard from the first peer to the second, putting both on the second
4. remove the first peer (from consensus)
5. clear storage for the first peer
6. rejoin first peer by bootstrapping from the second peer
7. the new peer joins the cluster but has inconsistent state, missing one of the shards on the newly joined peer

Just removing the first peer and joining any new one should be enough. But the above steps allow you to easily confirm the problem has occurred by checking the collection/shard state.

## Expected Behavior

Joining an existing cluster should always work fine, even if the first peer is removed. With the above scenario we could confirm this by seeing the collection/shard state being consistent.

In fact, joining a cluster should always work no matter what the current peer configuration is like.

## Possible Solution

It looks like the core problem is that we don't write an operation to the consensus (WAL) log for the first peer joining the cluster. Sadly we cannot update the start of the WAL in already created clusters.

Thus far we think we need to implement the following three things:
1. newly created clusters should write the first peer joining to the WAL
2. existing nodes should store their current 'origin peer' to their raft state, when bootstrapping we should include this origin peer (either from the WAL or from the bootstrap state)
3. properly implement consensus snapshots, which should bypass this problem

ffuugoo:
Fixes #5138 for new clusters.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/consensus_manager.rs
+++ b/lib/storage/src/content_manager/consensus_manager.rs
@@ -3,6 +3,7 @@ use std::error::Error;
 use std::fmt::Display;
 use std::future::Future;
 use std::ops::Deref;
+use std::str;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
@@ -386,11 +387,30 @@ impl<C: CollectionContainer> ConsensusManager<C> {
         for single_change in &change.changes {
             match single_change.change_type() {
                 ConfChangeType::AddNode => {
-                    debug_assert!(
-                        self.peer_address_by_id()
-                            .contains_key(&single_change.node_id),
-                        "Peer should be already known"
-                    )
+                    let context = entry.get_context();
+
+                    if !context.is_empty() {
+                        let peer_uri = str::from_utf8(context)
+                            .map_err(|err| {
+                                StorageError::service_error(format!(
+                                    "failed to parse peer URI: {err}"
+                                ))
+                            })?
+                            .parse()
+                            .map_err(|err| {
+                                StorageError::service_error(format!(
+                                    "failed to parse peer URI: {err}"
+                                ))
+                            })?;
+
+                        self.add_peer(single_change.node_id, peer_uri)?;
+                    } else {
+                        debug_assert!(
+                            self.peer_address_by_id()
+                                .contains_key(&single_change.node_id),
+                            "Peer should be already known"
+                        )
+                    }
                 }
                 ConfChangeType::RemoveNode => {
                     log::debug!("Removing node {}", single_change.node_id);
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -207,7 +207,7 @@ impl Consensus {
             Self::init(
                 &state_ref,
                 bootstrap_peer.clone(),
-                uri,
+                uri.clone(),
                 p2p_port,
                 &config,
                 tls_config.clone(),
@@ -492,6 +492,13 @@ impl Consensus {
         }
 
         if self
+            .try_add_origin()
+            .context("failed to propose origin peer URI to consensus")?
+        {
+            return Ok(Some(1));
+        }
+
+        if self
             .try_promote_learner()
             .context("failed to promote learner")?
         {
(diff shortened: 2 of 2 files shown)


lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -386,11 +387,30 @@ impl<C: CollectionContainer> ConsensusManager<C> {
...
+                            })?;
+
+                        self.add_peer(single_change.node_id, peer_uri)?;
+                    } else {
+                        debug_assert!(
+                            self.peer_address_by_id()
+                                .contains_key(&single_change.node_id),
+                            "Peer should be already known"
+                        )
+                    }
```
timvisee, line 413: Could you elaborate on why this change is needed?

Am I right that it allows us to add the very first peer, because it needs to take this path, while the others don't? And why can context be empty?

src/consensus.rs
```diff
@@ -491,6 +491,13 @@ impl Consensus {
             return Ok(None);
         }
 
+        if self
+            .try_add_origin()
+            .context("failed to propose origin peer URI to consensus")?
+        {
+            return Ok(Some(1));
+        }
```
timvisee, line 499: Shouldn't we just add the origin in the `init` function, for the non-bootstrap branch?

lib/storage/src/content_manager/consensus_manager.rs
ffuugoo: (Discussed outside GH, posting comment here for the context.)

Yes, this is required for origin peer.

When we add other peers (i.e., other than the origin peer) to the cluster, we use `AddLearnerNode`, which carries peer URI. Once new peer synchronized with consensus, we use `AddNode` (see `try_promote_learner`) without a URI (because peer URI was already added to consensus by initial `AddLearnerNode`) to finalize the process.

I haven't tested this, but I'm almost sure that `AddLernerNode` workflow won't work for origin peer, so I extended `AddNode` handling, so that it also contains an optional URI.

src/consensus.rs
ffuugoo: (Discussed outside GH, posting comment here for the context.)

This was my initial idea: to simply push a single `AddPeer` message to consensus channel, even before starting consensus thread/loop, so that it will be processed once consensus started. But unfortunately, it's never that easy with Raft. 😭

When origin node is started for the first time (i.e., when you start a new cluster), it needs to finish a self-election before you can propose/commit anything to consensus (it will return an error before leader is elected). So you need to spin consensus loop few times (to tick Raft node and process election) before you can propose and commit origin peer.

So I decided to organize `try_add_origin` a bit like `try_promote_learner` and add it as part of the consensus loop.

timvisee:
We decided to wait with merging for a bit.

We likely want to merge the other patch first: the patch that fixes this on existing clusters by persisting/sharing the origin peer.

src/consensus.rs
```diff
@@ -491,11 +491,33 @@ impl Consensus {
...
+
+            // Origin peer is not a leader yet, wait for the next tick and return to consensus loop
+            // to tick Raft node
+            Err(err @ TryAddOriginError::NotLeader) => {
+                log::debug!("{err}");
+
+                let timeout_at = previous_tick + tick_period;
+
+                self.runtime
+                    .block_on(async { tokio::time::sleep_until(timeout_at.into()).await });
```
timvisee, line 511: How about just [`std::thread::sleep`](https://doc.rust-lang.org/std/thread/fn.sleep.html), isn't that the same here?

ffuugoo: Good point. I guess, I was not thinking much here, just copied stuff from `recv_update`. 🤦‍♀️

ffuugoo: Fixed. (Thanks!)

src/consensus.rs
```diff
@@ -498,6 +498,34 @@ impl Consensus {
...
+            Ok(true) => return Ok(Some(1)),
+
+            // Origin peer is not a leader yet, wait for the next tick and return to consensus loop
+            // to tick Raft node
+            Err(err @ TryAddOriginError::NotLeader) => {
+                log::debug!("{err}");
+
+                let next_tick = previous_tick + tick_period;
+                let duration_until_next_tick = next_tick.saturating_duration_since(Instant::now());
+                thread::sleep(duration_until_next_tick);
```
timvisee, line 517: Is this still sound along with <https://github.com/qdrant/qdrant/pull/5173>?

tests/consensus_tests/test_cluster_rejoin.py
```diff
@@ -73,3 +75,137 @@ def test_rejoin_cluster(tmp_path: pathlib.Path):
...
+    tmp_path: pathlib.Path,
+    start_cluster: Callable[[pathlib.Path, int], tuple[list[str], list[pathlib.Path], str]],
+    raft_state: Callable[[dict[str, Any], int], Any | None],
+    collection: str = "test_collection",
+    peers: int = 3,
+    shards: int = 3,
+    expected_shards: int = 3,
+):
+    """
+    Parameterized test body, that tests adding new peer after origin peer was removed from the cluster.
```
timvisee, line 111: Not necessary, but maybe we can link the issue here because this roughly implements the described repro.

```suggestion
    Parameterized test body, that tests adding new peer after origin peer was removed from the cluster.
    See: <https://github.com/qdrant/qdrant/issues/5138>
```

timvisee (approved):

src/consensus.rs
ffuugoo: I think it is? Should be a no-op, but I'd keep it just in case, so committing origin would be "self-sufficient", if we revert #5173 for some reason.

tests/consensus_tests/test_cluster_rejoin.py
ffuugoo: Sure!

generall (approved):

PR #5142, opened 2024-09-26
[Strict mode] Remove obsolete timeout function

JojiiOfficial:
Depends on #5137 

The `timeout()` function in the `StrictModeVerification` trait is not used anymore since we already check the timeout value in the `check_strict_mode` function. Therefore we can remove it from the trait definition.


timvisee (approved):
image
 :ok_hand:

PR #5147, opened 2024-09-27
Persist and propagate origin peer when bootstrapping new nodes

### LINKED ISSUE #5138: Consensus: origin node not registered in consensus log, breaks if origin is removed

ffuugoo:
Partially fixes #5138 for existing clusters (if origin peer is not yet removed).

**TODO:**
- [ ] recover origin peer ID from the WAL on existing nodes if it was not persisted earlier

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/consensus/persistent.rs
+++ b/lib/storage/src/content_manager/consensus/persistent.rs
@@ -38,6 +38,8 @@ pub struct Persistent {
     /// Operations to applied, consensus consider them committed, but this peer didn't apply them yet
     #[serde(default)]
     pub apply_progress_queue: EntryApplyProgressQueue,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub first_voter: Option<PeerId>,
     /// Last known cluster topology
     #[serde(with = "serialize_peer_addresses")]
     pub peer_address_by_id: Arc<RwLock<PeerAddressById>>,
@@ -200,6 +202,18 @@ impl Persistent {
         self.apply_progress_queue.get_last_applied()
     }
 
+    pub fn first_voter(&self) -> PeerId {
+        match self.first_voter {
+            Some(id) => id,
+            None => self.this_peer_id(),
+        }
+    }
+
+    pub fn set_first_voter(&mut self, id: PeerId) -> Result<(), StorageError> {
+        self.first_voter = Some(id);
+        self.save()
+    }
+
     pub fn peer_address_by_id(&self) -> PeerAddressById {
         self.peer_address_by_id.read().clone()
     }
@@ -244,6 +258,7 @@ impl Persistent {
                 conf_state: ConfState::from((voters, vec![])),
             },
             apply_progress_queue: Default::default(),
+            first_voter: if first_peer { Some(this_peer_id) } else { None },
             peer_address_by_id: Default::default(),
             peer_metadata_by_id: Default::default(),
             cluster_metadata: Default::default(),
--- a/lib/storage/src/content_manager/consensus_manager.rs
+++ b/lib/storage/src/content_manager/consensus_manager.rs
@@ -91,9 +91,6 @@ pub struct ConsensusManager<C: CollectionContainer> {
     /// Sends messages to the consensus thread, which is defined externally, outside of the state.
     /// (e.g. in the `src/consensus.rs`)
     propose_sender: OperationSender,
-    /// Defines if this peer is a first peer of the consensus,
-    /// which might affect the init logic
-    first_voter: RwLock<Option<PeerId>>,
     /// Status of the consensus thread, changed by the consensus thread
     consensus_thread_status: RwLock<ConsensusThreadStatus>,
     /// Consensus thread errors, changed by the consensus thread
@@ -117,7 +114,6 @@ impl<C: CollectionContainer> ConsensusManager<C> {
             toc,
             on_consensus_op_apply: Default::default(),
             propose_sender,
-            first_voter: Default::default(),
             consensus_thread_status: RwLock::new(ConsensusThreadStatus::Working {
                 last_update: Utc::now(),
             }),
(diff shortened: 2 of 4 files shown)


ffuugoo:
> ```
> error: redundant closure
>    --> src/tonic/api/raft_api.rs:120:29
>     |
> 120 |             .or_insert_with(|| Uri::default());
>     |                             ^^^^^^^^^^^^^^^^^ help: replace the closure with the function itself: `Uri::default`
> ```

`cargo fix` generated this code (from `if contains_key { insert }`) itself. So annoying... 🙄

lib/storage/src/content_manager/consensus/persistent.rs
```diff
@@ -38,6 +38,8 @@ pub struct Persistent {
     /// Operations to applied, consensus consider them committed, but this peer didn't apply them yet
     #[serde(default)]
     pub apply_progress_queue: EntryApplyProgressQueue,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub first_voter: Option<PeerId>,
```
timvisee, line 42: Why not call this `origin_peer_id`?

I feel like first voter doesn't strictly represent the very first peer ID we ever see.

src/consensus.rs
```diff
@@ -442,7 +446,7 @@ impl Consensus {
         // Only first peer has itself as a voter in the initial conf state.
         // This needs to be propagated manually to other peers as it is not contained in any log entry.
         // So we skip the learner phase for the first peer.
-        state_ref.set_first_voter(all_peers.first_peer_id);
+        state_ref.set_first_voter(all_peers.first_peer_id)?;
         state_ref.set_conf_state(ConfState::from((vec![all_peers.first_peer_id], vec![])))?;
```
timvisee, line 450: Maybe it's worth to combine both, so we only write a single time.

lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -194,14 +190,90 @@ impl<C: CollectionContainer> ConsensusManager<C> {
...
+        };
+
+        if first_entry.index != 1 {
+            log::warn!("TODO: compacted WAL");
+            return Ok(Some(PeerId::MAX));
+        }
+
+        let mut peers: HashSet<_> = self.peers().into_iter().collect();
+
+        for index in first_entry.index..last_entry.index + 1 {
```
timvisee, line 235: I was a bit confused by this at first. Maybe we can add some comment to describe it.

```suggestion
        // Drain added peers from list, if we have one left that'd be our origin peer
        let mut peers: HashSet<_> = self.peers().into_iter().collect();
        for index in first_entry.index..last_entry.index + 1 {
```

I figured, maybe we can do it the other way around. To find a remove for which we don't have a add entry. But I assume you're not doing that because we also store garbage removes in the WAL for peers that never existed.

ffuugoo: Added a comment.

ffuugoo:
These tests were a nightmare. :/

lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -788,6 +802,87 @@ impl<C: CollectionContainer> ConsensusManager<C> {
...
+
+    if peers.len() > 1 {
+        log::warn!(
+            "Failed to recover first voter peer: \
+             found multiple peers without ConfChange entry in WAL: \
+             {peers:?}"
+        );
+
+        return Ok(Some(PeerId::MAX));
+    }
```
timvisee, line 881: Also check and if the list is empty?

I think we can report:

> Failed to recover first voter peer: first peer has already been removed from the cluster

tests/consensus_tests/test_cluster_rejoin_data/collections_meta_wal/open-1
timvisee, line 1: These (`open-{1,2}`) currently are 2x 32MB. Would it be possible to shrink them to just their entries, and/or even omit the second file?

lib/storage/src/content_manager/consensus_manager.rs
ffuugoo: If `peers` is empty by the end of the test, it means first peer **is** in the WAL. It's "handled" by the `peers.into_iter().next()` (which will return `None`) at the end of the method.

Though, maybe I should change it to return `PeerId::MAX`, so we don't try `recover_first_voter` again.

timvisee: > it means first peer is in the WAL. It's "handled" by the peers.into_iter().next()

Ah of course, in cases where we did add the origin peer as first entry.

tests/consensus_tests/test_cluster_rejoin_data/collections_meta_wal/open-1
generall, line 1: can we have smaller files for tests? Or maybe compress this 2?

timvisee: I did shrink `open-1` to just 256 bytes and removed `open-2`.

That's from 64MB down to 256 bytes. Tests still work as expected.

Did modified the commit in which they were appended to not drag these files along in our git history forever. I still have the original once if you need them @ffuugoo.

timvisee (approved):

tests/consensus_tests/test_cluster_rejoin_data/collections_meta_wal/open-1
ffuugoo: Thanks! 🙏

generall (approved):

PR #5150, opened 2024-09-29
`set_payload`: for `null` args, set the key to `null` instead of removing it from the payload

duci9y:
### All Submissions:

### Changes to Core Features:

I've added support for explicitly setting a key path to `null` within a payload, which differs from the current behavior of completely removing the key path when `null` is assigned. This is important for my use case, where setting a keypath to `null` carries specific meaning distinct from the key path being missing from the map. An alternative could be using a custom representation like `{ "type": "null" }`, but that feels awkward and inconsistent with how typical JSON implementations do it.

### DIFF
--- a/lib/segment/src/common/utils.rs
+++ b/lib/segment/src/common/utils.rs
@@ -40,10 +40,7 @@ pub fn merge_map(
     source: &serde_json::Map<String, Value>,
 ) {
     for (key, value) in source {
-        match value {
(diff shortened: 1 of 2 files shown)


timvisee:
> which differs from the current behavior of completely removing the key path when null is assigned

I understand this may be beneficial. But I'm afraid we can't merge this as it'd be a breaking API change.

PR #5163, opened 2024-10-01
mmap geo index api

IvanPleshkov:
REST/grpc api and integration of mmap geo index


generall:
I think `counts_per_hash` and `points_values_count` should be updated on deletes. It doesn't seems to be much problematic with mmaps

generall (approved):

PR #5165, opened 2024-10-01
Inference interface in REST and gRPC

generall:
Introduces `Document`, `Image` and `InferenceObject` into gRPC and REST API. 
Those data structures may be used as a replacement for vectors for insertion and query. Whenever one of those structures is used, it should be replaced internally with inferenced vector.

Inference itself if out of scope for this PR and will be introduced separately.

Some changes made to make it work:

- grpc `Vector` is modified to include additional one_of field `vector`. It is similar to existing `VectorInput`, put doesn't allow usage of Ids. The idea is to migrate `Vector` from current hacky representation of vectors (which require un-flattening of multivectors) into `one_of` representation. In order to do this all components should learn how to read `one_of` field and prefer it if it exists. On the later iterations we will start to output using this field and mark existing ones as deprecated

- Previously we used the same vector structures to describe input and output, but we can't do it anymore, as input now have additional options. This PR introduces `*Output` structures, which are the copies of older version common structures with no extra fields. That required writing additional conversions

- Existing conversion rely on TryFrom trait, but inference require `async` function with extra parameters, so TryFrom implementation is not enough


lib/collection/src/operations/point_ops.rs
```diff
@@ -103,6 +103,52 @@ impl From<VectorPersisted> for VectorInternal {
...
+//                       Only Vectors             │      Only Vectors
+//                       Implement JsonSchema     │      Optimized for search
+//                                                │
+//                                                │
+//                      ┌─────────────────────┐   │
+//                      │ grpc::VectorsOutput ◄───┘
+//                      └───────────┬─────────┘
+//                                  │
+//                                  ▼
+//                              gPRC Response
```
generall, line 149: This is a general idea of data structures refactoring

generall, line 149: This is a general idea of data structures refactoring

timvisee (approved):
We did go over this in a pair review. Surprisingly: no complaints. :smile: 

Drafting so we don't accidentally merge it too soon.

PR #5173, opened 2024-10-03
Instantly self-elect Raft leader, when cluster contains single peer

ffuugoo:
Currently, when starting first peer of a new cluster, we have to wait for a short while (a second or two) until the first peer self-elects itself as a leader. Before the leader is elected, we can't bootstrap other nodes of the cluster or create a collection.

This PR checks if we are the only peer in the cluster and if so, ticks the node few times, so that it self-elects itself as a leader instantly.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -448,6 +448,13 @@ impl Consensus {
     }
 
     pub fn start(&mut self) -> anyhow::Result<()> {
+        // When starting first peer of a new cluster, tick a few times to establish leader instantly
+        if self.node.store().is_new_deployment() && self.node.store().peer_count() == 1 {
+            while !self.node.has_ready() {
+                self.node.tick();
+            }
+        }
+
         let tick_period = Duration::from_millis(self.config.tick_period_ms);
 
         let mut previous_tick = Instant::now();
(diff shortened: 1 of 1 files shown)


timvisee (approved):
Great idea. I don't see any problems with this.

I just tested this with some log messages, and it works as expected.

timvisee:
Should we also do this on restart, if we don't have any leader yet and we're still the only peer?

src/consensus.rs
```diff
@@ -448,8 +448,15 @@ impl Consensus {
     }
 
     pub fn start(&mut self) -> anyhow::Result<()> {
-        let tick_period = Duration::from_millis(self.config.tick_period_ms);
+        // If this is the only peer in the cluster, tick Raft node a few times to instantly
+        // self-elect itself as Raft leader
+        if self.node.store().peer_count() == 1 {
```
timvisee, line 453: Does it make sense to make this check part of the loop?

I'm not sure if we might collect more peers before `has_ready() == true`.

ffuugoo: > Does it make sense to make this check part of the loop?

I don't think so.

> I'm not sure if we might collect more peers before has_ready() == true

We can't. We either collect them when bootstrapping new node/loading existing node from disk, or we have to apply some committed entry, which happens during `on_ready` call (under `has_ready` in the loop).

PR #5175, opened 2024-10-03
Introduce SnapshotFormat

xzfc:
This PR introduces `enum SnapshotFormat`. The usage in this PR is minimal:
- Passed across `create_snapshot()` invocations.
- The name appears in log output during unpacking.

In subsequent PRs it will be used to control the format of resulting snapshots.

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -2420,6 +2420,73 @@ impl Filter {
     }
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum SnapshotFormat {
+    /// Created by Qdrant `<0.11.0`.
+    ///
+    /// The collection snapshot contains nested tar archives for segments.
+    /// Segment tar archives contain a plain copy of the segment directory.
+    ///
+    /// ```plaintext
+    /// ./0/segments/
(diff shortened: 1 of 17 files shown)


lib/segment/src/types.rs
```diff
@@ -2420,6 +2420,73 @@ impl Filter {
...
+    /// ├── 0b31e274-dc65-40e4-8493-67ebed4bcf10.tar
+    /// │   ├── segment.json
+    /// │   ├── CURRENT
+    /// │   ├── 000009.sst
+    /// │   ├── 000010.sst
+    /// │   └── …
+    /// ├── 1d6c96ec-7965-491a-9c45-362d55361e9b.tar
+    /// └── …
+    /// ```
+    Ancient,
```
timvisee, line 2441: 🦖🏺📜🗿

Do we still support unpacking these? If we do, let's keep it. Otherwise I'd suggest to remove this.

xzfc: Not sure if it works, but the code is still there.
https://github.com/qdrant/qdrant/blob/dee106d74625e891256ebe06c2c2e0515650e67e/lib/segment/src/segment/segment_ops.rs#L471-L474

PR #5187, opened 2024-10-07
Add strict mode config to update collection API + Populate strict mode in API

JojiiOfficial:
Allows changing strict mode configuration using the update collection API.
Also makes strict mode API docs visible.

Integration tests: #5189

### DIFF
--- a/lib/storage/src/content_manager/toc/collection_meta_ops.rs
+++ b/lib/storage/src/content_manager/toc/collection_meta_ops.rs
@@ -162,7 +162,6 @@ impl TableOfContent {
         if let Some(changes) = replica_changes {
             collection.handle_replica_changes(changes).await?;
         }
-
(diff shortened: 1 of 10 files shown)


lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -162,6 +163,10 @@ impl TableOfContent {
             collection.handle_replica_changes(changes).await?;
         }
 
+        if let Some(strict_mode) = strict_mode {
+            collection.update_strict_mode_config(strict_mode).await?;
+        }
+
```
timvisee, line 169:
```suggestion
        if let Some(strict_mode) = strict_mode {
            collection.update_strict_mode_config(strict_mode).await?;
        }

```

timvisee (approved):

generall (approved):

PR #5188, opened 2024-10-07
enable mmap vector storage by default

generall:


timvisee (approved):
Tested with 1k and 1m bfb vectors.

image

PR #5217, opened 2024-10-11
Refactor `ConsensusOpWal`

ffuugoo:
While working on WAL compaction I've refactored `ConsensusOpWal` (cause it was too confusing to work with) and decided to split it into separate PR. The behavior after refactoring should be exactly the same, just with (hopefully 🤞) cleaner code and some trivial optimizations.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/consensus/consensus_wal.rs
+++ b/lib/storage/src/content_manager/consensus/consensus_wal.rs
@@ -47,6 +47,10 @@ impl ConsensusOpWal {
         self.entry_by_wal_index(wal_index)
     }
 
+    /// Get all entries in the given range
+    ///
+    /// The end of the range is exclusive (`[from_raft_index, until_raft_index)`).
+    /// A specified `max_size_bytes` will be ignored for the first entry.
     pub fn entries(
         &self,
         from_raft_index: u64,
--- a/lib/storage/src/content_manager/consensus/consensus_wal.rs
+++ b/lib/storage/src/content_manager/consensus/consensus_wal.rs
@@ -128,6 +128,7 @@ impl ConsensusOpWal {
 
                 // Check that new entry index is sequential (it's not greater than next WAL index),
                 // or truncate entries at the tail of WAL, if it overwrites some
+                #[allow(clippy::comparison_chain)] // stupid ahh diagnostics 🙄
                 if new_entry_wal_index > next_wal_index {
                     return Err(StorageError::service_error(format!(
                         "Can't append entry with Raft index {} (expected WAL index {}), \
--- a/lib/storage/src/content_manager/consensus/consensus_wal.rs
+++ b/lib/storage/src/content_manager/consensus/consensus_wal.rs
@@ -1,203 +1,288 @@
-use std::fs::create_dir_all;
 use std::path::Path;
+use std::{cmp, fs};
 
-use itertools::Itertools;
 use prost_for_raft::Message;
 use protobuf::Message as _;
 use raft::eraftpb::Entry as RaftEntry;
 use wal::Wal;
 
 use crate::content_manager::consensus_manager;
-use crate::{ConsensusOperations, StorageError};
+use crate::content_manager::consensus_ops::ConsensusOperations;
+use crate::StorageError;
 
 const COLLECTIONS_META_WAL_DIR: &str = "collections_meta_wal";
 
-pub struct ConsensusOpWal(Wal);
+#[derive(Debug)]
+pub struct ConsensusOpWal {
+    wal: Wal,
+}
 
 impl ConsensusOpWal {
     pub fn new(storage_path: &str) -> Self {
         let collections_meta_wal_path = Path::new(storage_path).join(COLLECTIONS_META_WAL_DIR);
-        create_dir_all(&collections_meta_wal_path)
-            .expect("Can't create Collections meta Wal directory");
-        let wal = Wal::open(collections_meta_wal_path).expect("Can't open Collections meta Wal");
-        ConsensusOpWal(wal)
+
+        fs::create_dir_all(&collections_meta_wal_path)
+            .expect("Can't create consensus WAL directory");
+
+        let wal = Wal::open(collections_meta_wal_path).expect("Can't open consensus WAL");
+        Self { wal }
     }
 
     pub fn clear(&mut self) -> Result<(), StorageError> {
-        Ok(self.0.clear()?)
+        self.wal.clear()?;
+        Ok(())
     }
 
-    pub fn entry(&self, id: u64) -> raft::Result<RaftEntry> {
+    pub fn entry(&self, raft_index: u64) -> raft::Result<RaftEntry> {
         // Raft entries are expected to have index starting from 1
-        if id < 1 {
+        if raft_index < 1 {
             return Err(raft::Error::Store(raft::StorageError::Unavailable));
         }
-        let first_entry = self
-            .first_entry()
-            .map_err(consensus_manager::raft_error_other)?
-            .ok_or_else(|| raft::Error::Store(raft::StorageError::Unavailable))?;
-        if id < first_entry.index {
-            return Err(raft::Error::Store(raft::StorageError::Compacted));
-        }
-        // Due to snapshots there might be different offsets between wal index and raft entry index
-        let offset = first_entry.index - self.0.first_index();
-        <RaftEntry as prost_for_raft::Message>::decode(
-            self.0
-                .entry(id - offset)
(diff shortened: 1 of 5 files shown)


ffuugoo:
Let's see if tests pass first. 😬🙏

lib/storage/src/content_manager/consensus/consensus_wal.rs
```diff
@@ -1,203 +1,288 @@
...
-            .try_collect()?;
+        let offset = self.index_offset()?;
+
+        // Map requested Raft indices to WAL indices
+        let from_wal_index = offset.raft_to_wal(from_raft_index);
+        let until_wal_index = offset.raft_to_wal(until_raft_index);
+
+        // Bound mapped WAL indices between first/last WAL indices
+        let from_wal_index = cmp::max(from_wal_index, offset.wal_index);
+        let until_wal_index = cmp::min(until_wal_index, offset.wal_index + self.wal.num_entries());
```
timvisee, line 64: Is this the same? If so I think I'd find that a bit easier to read.

```suggestion
        let from_wal_index = cmp::max(from_wal_index, self.wal.first_index());
        let until_wal_index = cmp::min(until_wal_index, self.wal.last_index());
```

lib/storage/src/content_manager/consensus/consensus_wal.rs
```diff
@@ -1,203 +1,288 @@
...
+
+        let wal_index = self
+            .index_offset()?
+            .try_raft_to_wal(raft_index)
+            .ok_or_else(|| raft::Error::Store(raft::StorageError::Compacted))?;
+
+        self.entry_by_wal_index(wal_index)
     }
 
     pub fn entries(
```
timvisee, line 50: Is `until` inclusive or exclusive? I thought it was inclusive before, and it looks like it's implemented to be exclusive now.

Either way, let's document it in a comment here.

lib/storage/src/content_manager/consensus/consensus_wal.rs
```diff
@@ -1,203 +1,288 @@
...
+            let entry = self.entry_by_wal_index(wal_index)?;
+
+            size_bytes += u64::from(entry.compute_size());
+            entries.push(entry);
+
+            if let Some(max_size_bytes) = max_size_bytes {
+                if size_bytes >= max_size_bytes {
+                    break;
+                }
+            }
```
timvisee, line 90: This likely goes over the limit in many cases.

Shouldn't we try to stay below the limit unless we don't have any items yet?

Maybe we can move the if-statement before `entries.push()`: `if !entries.is_empty() && size_bytes >= max_size_bytes`

lib/storage/src/content_manager/consensus/consensus_wal.rs
```diff
@@ -1,203 +1,288 @@
...
+                        "Can't append entry with Raft index {}, \
+                         because WAL is already compacted at Raft index {}",
+                        new_entry.index, offset.raft_index,
+                    )));
+                };
 
-            let index = entry.index;
-            let current_index = self.0.last_index();
-            let index_offset = self.index_offset()?;
+                let next_wal_index = self.wal.last_index() + 1;
```
timvisee, line 127: Note that if we have either 0 or 1 items in our WAL, `last_index()` produces `0` in both cases.

In other words, it returns a nonsensical value if the WAL is still empty. I'm not sure if that's desirable.

Maybe we want to return an option there so we have to handle this case explicitly.

lib/storage/src/content_manager/consensus/consensus_wal.rs
ffuugoo: Hmmm. Maybe. But I've implemented it like this intentionally, so that we always include at least one entry if `max_size_bytes` is 0 (or some other small size, that does not fit a single entry).

lib/storage/src/content_manager/consensus/consensus_wal.rs
ffuugoo: This whole branch is only triggered when `current_index_offset` is `Some`, which means there is at least one entry in the WAL, which means `Wal::last_index` should be valid.

ffuugoo: This whole branch is only triggered when `current_index_offset` is `Some`, which means there is at least one entry in the WAL, which means `Wal::last_index` should be valid.

lib/storage/src/content_manager/consensus/consensus_wal.rs
timvisee: I understand, but wouldn't the above mentioned alternative achieve the same?

lib/storage/src/content_manager/consensus/consensus_wal.rs
ffuugoo: It's currently exclusive, and _I think_ I've made it that way, cause original implementation was exclusive. But I'll verify.

lib/storage/src/content_manager/consensus/consensus_wal.rs
timvisee: Makes sense, thanks!

lib/storage/src/content_manager/consensus/consensus_wal.rs
timvisee: > But I'll verify.

Thanks! :pray:

lib/storage/src/content_manager/consensus/consensus_wal.rs
ffuugoo: It's not, _if we consider `until` to be exclusive_. `first_index + num_entries` always returns one-after-the-last-index.

timvisee: Yes, makes sense. Thanks!

lib/storage/src/content_manager/consensus/consensus_wal.rs
ffuugoo: Almost, except if you repeatedly request entries with a very small `max_size_bytes` it might never return you anything. Though, I'll verify how original impl handled this and tweak it accordingly.

timvisee: > Almost, except if you repeatedly request entries with a very small max_size_bytes it might never return you anything.

How? Doesn't the `!entries.is_empty()` take care of that?

lib/storage/src/content_manager/consensus/consensus_wal.rs
```diff
@@ -1,203 +1,289 @@
...
-        low: u64,
-        high: u64,
-        max_size: Option<u64>,
+        from_raft_index: u64,
+        until_raft_index: u64,
+        max_size_bytes: Option<u64>,
     ) -> raft::Result<Vec<RaftEntry>> {
-        let mut size = 0;
-
-        let entries = (low..high)
```
ffuugoo, line 61: I think previous implementation of `entries` definitely was exlcusive, cause we take `low..high` here.

lib/storage/src/content_manager/consensus/consensus_wal.rs
ffuugoo: [I think tt was exclusive.](https://github.com/qdrant/qdrant/pull/5217#discussion_r1796838494)

lib/storage/src/content_manager/consensus/consensus_wal.rs
ffuugoo: Hmmmm. And you once again right, I think. Ok then. 🤦‍♀️😂

ffuugoo: Fixed.

lib/storage/src/content_manager/consensus/consensus_wal.rs
```diff
@@ -1,203 +1,290 @@
...
+
+        let wal_index = self
+            .index_offset()?
+            .try_raft_to_wal(raft_index)
+            .ok_or_else(|| raft::Error::Store(raft::StorageError::Compacted))?;
+
+        self.entry_by_wal_index(wal_index)
     }
 
     pub fn entries(
```
timvisee, line 50:
```suggestion
    /// Get all entries in the given range
    ///
    /// The end of the range is exclusive (`[from_raft_index, until_raft_index)`).
    /// A specified `max_size_bytes` will be ignored for the first entry.
    pub fn entries(
```

timvisee (approved):

generall (approved):

PR #5219, opened 2024-10-11
[perf] Do not capture backtrace on get_sparse_vector_opt

agourlay:
On a `perf` optimized build.

sparse-opt-perf

The problem is that we are capturing backtraces at a high frequency when trying to fetch non existent sparse vectors from RocksDB.

The proposed fix is to have a specialised function for getting optional value from RocksDB.

## Benchmark.

`./bfb --vectors-per-point 0 --sparse-vectors-per-point 1  --sparse-vectors 0.3 --num-vectors 100000`

- before: upload 28s
- after: upload 22s

### DIFF
--- a/lib/segment/src/common/operation_error.rs
+++ b/lib/segment/src/common/operation_error.rs
@@ -61,6 +61,8 @@ pub enum OperationError {
 }
 
 impl OperationError {
+    /// Create a new service error with a description and a backtrace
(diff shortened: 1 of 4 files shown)


timvisee (approved):
The difference in upsertion time is huge. Nice find!

generall (approved):

PR #5222, opened 2024-10-11
handle the case of empty geo hashes

generall:
Without this fix service panics if `radius: 0`

### DIFF
--- a/lib/segment/src/index/field_index/geo_index/mod.rs
+++ b/lib/segment/src/index/field_index/geo_index/mod.rs
@@ -204,7 +204,9 @@ impl GeoMapIndex {
             return CardinalityEstimation::exact(0);
         }
 
-        let common_hash = common_hash_prefix(values);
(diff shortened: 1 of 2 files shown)


agourlay (approved):

lib/segment/src/index/field_index/geo_index/mod.rs
```diff
@@ -204,7 +204,9 @@ impl GeoMapIndex {
             return CardinalityEstimation::exact(0);
         }
 
-        let common_hash = common_hash_prefix(values);
+        let Some(common_hash) = common_hash_prefix(values) else {
```
IvanPleshkov, line 207: I propose to check directly here if `values` are empty. For `common_hash_prefix` return error is `values` are empty

generall: all other logic is sound if values are empty

PR #5228, opened 2024-10-14
Hash/lookup point version only once, not 3 times

timvisee:
In the segments searcher retrieve logic we loop over points and keep their latest version.

This PR improves it by only hashing/looking up once, rather than three times. For this we use the entry API.

Note that this doesn't fix bad behavior, and it's just low hanging fruit in terms of performance improvement.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -1,3 +1,4 @@
+use std::collections::hash_map::Entry;
 use std::collections::{BTreeSet, HashMap};
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection_manager/segments_searcher.rs
```diff
@@ -389,44 +390,49 @@ impl SegmentsSearcher {
                 let version = segment.point_version(id).ok_or_else(|| {
                     OperationError::service_error(format!("No version for point {id}"))
                 })?;
-                // If this point was not found yet or this segment have later version
-                if !point_version.contains_key(&id) || point_version[&id] < version {
-                    point_records.insert(
+
+                // If we already have the latest point version, keep that and continue
+                let version_entry = point_version.entry(id);
+                if matches!(version_entry, Entry::Occupied(ref entry) if *entry.get() >= version) {
```
ffuugoo, line 396: Eeeew, `ref`! 🤢

```suggestion
                if matches!(&version_entry, Entry::Occupied(entry) if *entry.get() >= version) {
```

ffuugoo (approved):

lib/collection/src/collection_manager/segments_searcher.rs
ffuugoo: Hopefully this works, or I'll look silly 🤡

timvisee: It does

agourlay (approved):

PR #5230, opened 2024-10-14
Remove unnecessary clone when reordering records

timvisee:
Remove an unnecessary clone. Can be beneficial on large point records.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/scroll.rs
+++ b/lib/collection/src/shards/local_shard/scroll.rs
@@ -202,7 +202,7 @@ impl LocalShard {
         let with_payload = WithPayload::from(with_payload_interface);
         // update timeout
         let timeout = timeout.saturating_sub(start.elapsed());
-        let records_map = tokio::time::timeout(
+        let mut records_map = tokio::time::timeout(
             timeout,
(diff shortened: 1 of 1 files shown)


ffuugoo (approved):

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -303,7 +304,7 @@ impl LocalShard {
 
         let ordered_records = point_ids
             .iter()
-            .filter_map(|point| records_map.get(point).cloned())
+            .filter_map(|point_id| records_map.get(point_id).cloned())
```
agourlay, line 307: why not do the same trick here?

timvisee: We do return the same point ID multiple times there. It happens if you have a point with multiple payload values, and it returns it for each value.

On the first push I actually used the same trick here, and tests started failing because of the above.

agourlay: Makes sense :+1:

agourlay (approved):

PR #5237, opened 2024-10-14
Clean up temporary snapshot files

xzfc:
Continuation of #5227.

This PR fixes the following bug: when s3 storage is used, a leftover temporary file appears after creating a shard snapshot.
```console
$ ls snapshots/tmp/                                                  
upload

$ curl -X POST http://localhost:6333/collections/q/shards/0/snapshots
{"result":{"name":"q-shard-0-2024-10-14-23-03-54.snapshot","creation_time":"2024-10-14T23:03:55","size":425984},"status":"ok","time":0.863852705}

$ ls snapshots/tmp/                                                  
q-shard-0-2024-10-14-23-03-54.snapshot-38We4I.tar  upload
```

Additionally, this PR makes the local (non-s3) case more robust: temporary files now getting cleaned up in case of an error.

### DIFF
--- a/lib/collection/src/common/file_utils.rs
+++ b/lib/collection/src/common/file_utils.rs
@@ -30,30 +30,46 @@ pub async fn move_dir(from: impl Into<PathBuf>, to: impl Into<PathBuf>) -> Colle
     Ok(())
 }
 
+/// Move file from one location to another.
+/// Handles the case when the source and destination are on different filesystems.
 pub async fn move_file(from: impl AsRef<Path>, to: impl AsRef<Path>) -> CollectionResult<()> {
-    // Try to rename first and fallback to copy to prevent TOCTOU
     let from = from.as_ref();
     let to = to.as_ref();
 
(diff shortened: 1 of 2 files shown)


lib/collection/src/common/file_utils.rs
```diff
@@ -30,30 +30,48 @@ pub async fn move_dir(from: impl Into<PathBuf>, to: impl Into<PathBuf>) -> Colle
...
-        })?;
+    // Try to rename first and fallback to copy to prevent TOCTOU.
+    if let Ok(()) = tokio::fs::rename(from, to).await {
+        return Ok(());
+    }
+
+    // If rename failed, try to copy.
+    // It is possible that the source and destination are on different filesystems.
+    if let Err(err) = tokio::fs::copy(from, to).await {
+        cleanup_file(from).await;
```
generall, line 48: Are you sure that we want to remove both files if the move failed?

generall (approved):

PR #5239, opened 2024-10-15
Add HardwareCounterCell + CPU measurement for sparse vector search

JojiiOfficial:
Adds a new type for hardware usage measurement and implements measuring CPU usage of sparse vector search.

### Design choices
- `std::cell::Cell` for better performance when measuring, also in places where we don't have mutable access to the counter variable.
- `HardwareCounterCell`-type to group hardware measurements together. Also prevents having each measurement counter in each (search) function signature individually which would result in too many arguments.
- `HardwareCounterCell` in (search) function signatures to allow populating measurements to calling functions.
  This is due to some structures requiring to be `Sync + Send`, which `HardwareCounterCell` doesn't implement and because we don't want to have measurement counters in structures like `VectorIndex`.

### DIFF
--- a/lib/segment/src/index/plain_payload_index.rs
+++ b/lib/segment/src/index/plain_payload_index.rs
@@ -4,7 +4,6 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use atomic_refcell::AtomicRefCell;
-use common::counter::hardware_counter::HardwareCounterCell;
 use common::types::{PointOffsetType, ScoredPointOffset, TelemetryDetail};
 use parking_lot::Mutex;
 use schemars::_serde_json::Value;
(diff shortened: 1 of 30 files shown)


lib/segment/src/index/plain_payload_index.rs
```diff
@@ -278,6 +279,7 @@ impl VectorIndex for PlainIndex {
         top: usize,
         params: Option<&SearchParams>,
         query_context: &VectorQueryContext,
+        _hardware_counter: &HardwareCounterCell,
```
generall, line 282: WDYT about putting `HardwareCounterCell` into `SegmentQueryContext`?

It is also per-segment, we should not have problems with sync/send, and it is already propagated everywhere (via VectorQueryContext). 

Overall it should have less impact on the code changes in tests

generall (approved):

generall (approved):

PR #5240, opened 2024-10-15
Inference Service Integration

n0x29a:


src/common/inference/service.rs
```diff
@@ -0,0 +1,145 @@
...
+        })?;
+
+        if status == reqwest::StatusCode::BAD_REQUEST {
+            let error_json: serde_json::Value =
+                serde_json::from_str(&response_body).map_err(|e| {
+                    StorageError::service_error(format!("Failed to parse error response: {e}"))
+                })?;
+
+            if let Some(error_message) = error_json["error"].as_str() {
+                return Err(StorageError::service_error(error_message.to_string()));
```
generall, line 128: If the user provided wrong model, proxy should reply with `BAD_REQUEST`, and qdrant should also answer bad_input or something, not `service_error`. Are there any cases when `BAD_REQUEST` of inference proxy is qdrant's fault?

src/common/inference/service.rs
```diff
@@ -0,0 +1,145 @@
...
+                    "Failed to send inference request: {e}, body: {error_body}"
+                ))
+            })?;
+
+        let status = response.status();
+        let response_body = response.text().await.map_err(|e| {
+            StorageError::service_error(format!("Failed to read response body: {e}"))
+        })?;
+
+        if status == reqwest::StatusCode::BAD_REQUEST {
```
generall, line 121: need more checks for statuses. I would prefer `match` here.

generall (changes requested):

generall (approved):

PR #5241, opened 2024-10-15
Add CPU measurement for multivectors

JojiiOfficial:
Depends on #5239 

Adds measurement for multi-vector search and some sparse vector search implementations.

### DIFF
--- a/lib/segment/src/vector_storage/query_scorer/multi_metric_query_scorer.rs
+++ b/lib/segment/src/vector_storage/query_scorer/multi_metric_query_scorer.rs
@@ -1,5 +1,6 @@
 use std::marker::PhantomData;
 
+use common::counter::hardware_counter::HardwareCounterCell;
 use common::types::{PointOffsetType, ScoreType};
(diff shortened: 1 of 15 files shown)


lib/segment/src/vector_storage/query_scorer/multi_metric_query_scorer.rs
```diff
@@ -49,6 +52,12 @@ impl<
         multi_dense_a: TypedMultiDenseVectorRef<TElement>,
         multi_dense_b: TypedMultiDenseVectorRef<TElement>,
     ) -> ScoreType {
+        println!("Score multi");
```
generall, line 55: debug code

generall (approved):
Idea for further PRs:

we might need to have some guarantees that we accounted for all computations and didn't miss anything. For this I propose to introduce `debug_assert` that hardware counters are consumed (eq. to 0) on Drop. 
This is not an absolute guarantee, but better than nothing

PR #5248, opened 2024-10-16
[WIP] Multithreaded sparse index build

agourlay:
The sparse index is currently built on a single thread which happens:
- when opening a RAM sparse index on startup
- when optimizing existing segments

Building a sparse index means populating posting list from payload storage and then sorting those at the end.

This PR explores the possibility to sort the posting list using multi-thread by leveraging the existing CPU permit infrastructure.

The efficiency depends on the size of the posting lists, it does not make sense to use a dedicated thread for sorting a short list.

## TODO
- [ ] benchmarks with real data

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6069,6 +6069,7 @@ dependencies = [
  "parking_lot",
  "pprof 0.13.0",
  "rand 0.8.5",
+ "rayon",
  "schemars",
  "serde",
  "serde_json",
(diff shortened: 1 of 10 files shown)


agourlay:
This change does not improve the current situation.

The bottleneck is filling up the mutable index with `sparse::index::posting_list::PostingList::upsert`.
Merging existing segment leverages the sparse index builder which is efficient enough.

PR #5251, opened 2024-10-16
Fix eager backtrace generation

agourlay:
similar: https://github.com/qdrant/qdrant/pull/5219

While doing jemalloc profiling I found out that we are eagerly generating backtraces during points updates.

eager


generall:
:clap:

coszio (approved):

timvisee (approved):

PR #5254, opened 2024-10-17
Do not create errors eagerly

agourlay:
This PR cleans up call sites that were creating error values eagerly to be dismissed right away.

I tried to find the places which were allocating and left the very cheap one untouched.


timvisee:
Nice! Did you just replace all of them, or only the errors that potentially create expensive backtraces?

agourlay:
There is no backtraces in this one.
I searched for all the `_or(` usages and promoted the ones which are not using constants or very cheap values.

generall (approved):
LGTM

timvisee (approved):

PR #5256, opened 2024-10-17
Only fall back to different shard transfer method for specific transfers

timvisee:
Adjust our shard transfer fallback mechanism to only fall back on specific shard transfers.

Fallback:
- stream records: yes
- wal delta: yes, needed in case a diff cannot be resolved
- snapshot: no, fallback method is likely expensive because it doesn't contain index
- resharding stream records: no, special transfer type incompatible with other methods

In other words, do not fall back anymore on snapshot or resharding stream record transfers.

### All Submissions:

### New Feature Submissions:


timvisee:
We currently only go into the retry logic if a WAL delta transfer fails. 

The change in this PR therefore does not make any difference.

Instead of adjusting I'll just close this as I don't think it's worth the extra work.

PR #5261, opened 2024-10-18
Cpu measurements for dense vector + custom sparse scorer

JojiiOfficial:


lib/sparse/src/common/sparse_vector.rs
```diff
@@ -97,14 +102,41 @@ impl RemappedSparseVector {
...
+
     /// Score this vector against another vector using dot product.
     /// Warning: Expects both vectors to be sorted by indices.
     ///
     /// Return None if the vectors do not overlap.
     pub fn score(&self, other: &RemappedSparseVector) -> Option<ScoreType> {
         debug_assert!(self.is_sorted());
         debug_assert!(other.is_sorted());
-        score_vectors(&self.indices, &self.values, &other.indices, &other.values)
+        let hardware_counter = HardwareCounterCell::new();
```
JojiiOfficial, line 132: I'm not sure which is the best practical way of handling this properly.
Having the score function twice, one metered and one not is unpractical in my opinion. However measuring and throwing the measurements away like here isn't clean either.
Feel free to suggest which approach is preferred

lib/sparse/src/common/sparse_vector.rs
```diff
@@ -71,10 +74,12 @@ fn score_vectors<T: Ord + Eq>(
                 score += self_values[i] * other_values[j];
                 i += 1;
                 j += 1;
+                cpu_counter += 1;
```
generall, line 77: Putting the counter here would mean, that computing score between 2 huge vectors which have no intersection would require 0 cpu resources. I don't think this is correct, I would prefer to cost of this operations and just a sum: `self_indices.len() + other_indices.len()`

lib/sparse/src/common/sparse_vector.rs
generall: Why exactly do we need unmeasured function?

JojiiOfficial: Because [here](https://github.com/qdrant/qdrant/blob/dev/lib/sparse/src/index/search_context.rs#L131) we already do measurement manually and don't need the `score` function to measure anything.
[Here](https://github.com/qdrant/qdrant/blob/more_scorer_measurements/lib/segment/src/vector_storage/query_scorer/sparse_custom_query_scorer.rs#L51) we do need the score function to measure though.

PR #5263, opened 2024-10-18
Refactor replica set update

timvisee:
Refactor the replica set update logic to make the possible branches easier to understand. Make tiny adjustments alongside it.

Most notably this refactors the two separate if-blocks into a single if{if-else} block to clarify both branches are mutually exclusive.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -365,7 +365,7 @@ impl ShardReplicaSet {
                 );
 
                 // Wait for replica failures to be accepted, otherwise return consistency error
-                if wait && wait_for_deactivation && !failures.is_empty() {
+                if wait && wait_for_deactivation {
                     // ToDo: allow timeout configuration in API
                     let timeout = DEFAULT_SHARD_DEACTIVATION_TIMEOUT;
 
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -336,7 +336,7 @@ impl ShardReplicaSet {
         let failure_error = if let Some((peer_id, collection_error)) = failures.first() {
             format!("Failed peer: {peer_id}, error: {collection_error}")
         } else {
-            "".to_string()
+            String::new()
         };
 
         if !failures.is_empty() {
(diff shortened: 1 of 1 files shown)


agourlay:
needs a rebase :)

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -322,91 +322,92 @@ impl ShardReplicaSet {
...
         }
 
-        if !failures.is_empty() && successes.len() < minimal_success_count {
-            self.handle_failed_replicas(
-                failures
-                    .iter()
-                    .filter(|(peer_id, _)| self.peer_is_resharding(*peer_id)),
-                &self.replica_state.read(),
-                update_only_existing,
-            );
```
timvisee, line 384: Removing this is problematic. Keeping it seems to be problematic too.

Currently discussing what's best to do here with @ffuugoo. Will mark this as draft in the meanwhile.

timvisee: I reverted this bit to match the old behavior.

It may still not be ideal for resharding, but our plan is to handle it in a separate PR. Resharding is not released yet anyway.

timvisee:
> needs a rebase :)

@agourlay Done!

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -322,91 +322,102 @@ impl ShardReplicaSet {
...
+                    &failures,
+                    &self.replica_state.read(),
+                    update_only_existing,
+                );
 
-            // completely failed - report error to user
-            let (_peer_id, err) = failures.into_iter().next().expect("failures is not empty");
-            return Err(err);
+                // Wait for replica failures to be accepted, otherwise return consistency error
+                if wait && wait_for_deactivation && !failures.is_empty() {
```
ffuugoo, line 368:
```suggestion
                if wait && wait_for_deactivation {
```

`!failures.is_empty()` is redundant here, because we check for it in the outer block

ffuugoo (changes requested):
I'd tweak a few things:

- I'd put `successes.len() >= minimal_success_count` branch first, and `successes.len() < minimal_success_count` second
	- `>= minimal_success_count` is a "happy-path", and it's easier to read it that way, IMO
- then I'd actually replace `successes.len() < minimal_success_count` with a simple `else` to `successes.len() >= minimal_success_count`
  - because they are mutually exclusive anyway

timvisee:
@ffuugoo Good suggestions. Thanks. Applied them now!

ffuugoo (approved):

agourlay (approved):

PR #5265, opened 2024-10-20
Reinit consensus

generall:


lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -789,7 +793,8 @@ impl<C: CollectionContainer> ConsensusManager<C> {
 
         let first_unapplied_index = applied_index + 1;
 
-        debug_assert!(first_unapplied_index <= first_entry.index);
+        // ToDo: it seems like a mistake, need to check if it's correct
+        // debug_assert!(first_unapplied_index <= first_entry.index);
```
generall, line 797: This part need extra review

tests/consensus_tests/test_reinit_consensus.py
```diff
@@ -0,0 +1,98 @@
...
+    compare_points(peer_urls_new[0])
+
+    print("Add one more peer")
+    peer_dir_additional = make_peer_folder(tmp_path, N_PEERS)
+    peer_dirs.append(peer_dir_additional)
+    peer_url_additional = start_peer(peer_dir_additional, f"peer_{N_PEERS}_additional.log", bootstrap_uri)
+    wait_peer_added(bootstrap_api_uri, expected_size=3)
+    peer_urls_new.append(peer_url_additional)
+    wait_for_uniform_cluster_status(peer_urls_new, leader)
+    wait_collection_exists_and_active_on_all_peers(collection_name=COLLECTION_NAME, peer_api_uris=peer_urls_new)
```
timvisee, line 98: Let's do another `compare_points(peer_urls_new[1])` here.

lib/storage/src/content_manager/consensus/persistent.rs
```diff
@@ -235,10 +258,14 @@ impl Persistent {
...
-    fn init(path: PathBuf, first_peer: bool) -> Result<Self, StorageError> {
+    fn init(
+        path: PathBuf,
+        first_peer: bool,
+        this_peer_id: Option<PeerId>,
+    ) -> Result<Self, StorageError> {
         // Do not generate too big peer ID, to avoid problems with serialization
         // (especially in json format)
-        let this_peer_id = rand::random::<PeerId>() % (1 << 53);
+        let this_peer_id = this_peer_id.unwrap_or_else(|| rand::random::<PeerId>() % (1 << 53));
```
timvisee, line 268: Unrelated but, do we do `% (1 << 53)` here to not use numbers higher than JavaScript supports?

src/consensus.rs
```diff
@@ -179,7 +181,24 @@ impl Consensus {
...
+
+        // On the bootstrap-ed peers during reinit of the consensus
+        // we want to make sure only the bootstrap peer will hold the true state
+        // Therefore we clear the WAL on the bootstrap peer to force it to request a snapshot
+        let clear_wal = reinit && bootstrap_peer.is_some();
+
+        if clear_wal {
+            log::debug!("Clearing WAL on the bootstrap peer to force snapshot transfer");
+            state_ref.clear_wal()?;
+        }
```
timvisee, line 200: This seems a bit dangerous in case we _accidentally_ start with `--bootstrap` on the first peer. But, I don't see a good way to safeguard this.

tests/consensus_tests/test_reinit_consensus.py
generall: :white_check_mark:

lib/storage/src/content_manager/consensus_manager.rs
timvisee: I agree. I think this is only valid if we always compact all entries.

Doesn't it need to be like this?

```suggestion
        // debug_assert!(first_unapplied_index >= first_entry.index);
```

lib/storage/src/content_manager/consensus/persistent.rs
generall: exactly because of JavaScript

src/consensus.rs
generall: intended to be used on copies of the data, not the main database

lib/storage/src/content_manager/consensus/persistent.rs
generall: Ideally we should have used a uuid, but migration overhead is too big now

src/consensus.rs
timvisee: Fair enough. No concerns then.

timvisee (approved):

PR #5271, opened 2024-10-21
Fix data race in WAL delta transfer causing inconsistencies

timvisee:
Fix a critical problem in WAL delta transfers which can cause missing updates, resulting in data inconsistencies.

First of all, during a WAL delta transfer, the target replica switches between these states:
1. Active: _regular behavior, before transfer_
2. Recovery: _get queued updates from transfer source, skip other updates_
3. Partial: _get last queued updates from transfer source, get all updates from other replicas_
4. Active: _regular behavior, after transfer_

The problem was in how different delta resolutions were handled. If the target shard was outdated and we resolved a diff, then everything was fine. If the target shard was already up-to-date and we resolved an empty diff, then problems occurred.

Imagine the target shard was up-to-date and we resolved an empty diff. In this case, we did not create a queue proxy at all. We did just fall through proxying and continued with the rest of the transfer logic. It did go through the replica states and would finish the transfer shortly after. The assumption was that a queue proxy is not needed because the target already has all the necessary updates. That assumption is wrong however. 

Even though the target shard is up-to-date at the time of resolution, it does not mean it's still up-to-date moments after. During resolution the target shard is in Recovery state. After resolution of an empty diff, we switch to Partial state. The key thing here is that new updates are not send to replicas in Recovery state. If we receive a new update between resolution and switching to Partial, it'll be lost on the target shard. Updates can fall in between.

That means that the following events result in inconsistencies:
- WAL delta transfer is started
- target shard switches into Recovery state
- we resolve WAL delta, it is empty because the target shard is up-to-date
- a new update comes in, the target shard does not receive it because it's still in Recovery state
- we switch the target shard into Partial state
- we finish the transfer and switch the target shard into active state
- the transfer is done, the target shard **never** received the above update

To fix the problem I now create a queue proxy in all cases. If the target shard is up-to-date and our resolved delta is empty, we start queuing/forwarding all new updates from that point on. All updates that fall in between the state switch will still reach the target shard because of the queue proxy, preventing missing updates.

With this fix the events would look like this:
- WAL delta transfer is started
- target shard switches into Recovery state
- we resolve WAL delta, it is empty because the target shard is up-to-date
- we create queue proxy, queueing all new updates
- a new update comes in, the target shard does not receive it yet because it's still in Recovery state
- we forward all queued items to the target shard, the target shard now receives the update
- we switch the target shard into Partial state
- we forward any new items on the queue and convert into a forward proxy
- we finish the transfer and switch the target shard into active state
- the transfer is done, the target shard did receive the above update

I'll attempt to write an integration test for this in a separate PR. I've already tested the update with 1000 subsequent WAL delta transfers, and the problem never appeared again.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -433,6 +433,10 @@ impl Inner {
             let items_left = (wal.last_index() + 1).saturating_sub(transfer_from);
             let items_total = (transfer_from - self.started_at) + items_left;
             let batch = wal.read(transfer_from).take(BATCH_SIZE).collect::<Vec<_>>();
+            debug_assert!(
+                batch.len() <= items_left as usize,
+                "batch cannot be larger than items_left",
+            );
             (items_left, items_total, batch)
         };
 
--- a/lib/collection/src/shards/replica_set/shard_transfer.rs
+++ b/lib/collection/src/shards/replica_set/shard_transfer.rs
@@ -486,4 +486,15 @@ impl ShardReplicaSet {
 
         local_shard.resolve_wal_delta(recovery_point).await
     }
+
+    pub async fn wal_version(&self) -> CollectionResult<Option<u64>> {
+        let local_shard_read = self.local.read().await;
+        let Some(local_shard) = local_shard_read.deref() else {
+            return Err(CollectionError::service_error(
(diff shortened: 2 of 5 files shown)


generall (approved):

timvisee:
Merging now so we can evaluate it in chaos testing ASAP.

PR #5286, opened 2024-10-22
Minor improvements around the WAL delta fix

timvisee:
Minor improvements on <https://github.com/qdrant/qdrant/pull/5271>.

This is mostly refactoring which I didn't have the time for yesterday. It contains no significant logic changes.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -1062,7 +1062,7 @@ pub enum ReplicaState {
     //
     // Snapshot shard transfer is in progress, updates aren't sent to the shard
     // Normally rejects updates. Since 1.8 it allows updates if force is true.
-    // TODO(1.10): remove PartialSnapshot state entirely?
+    #[schemars(skip)]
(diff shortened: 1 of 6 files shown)


lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -1062,7 +1062,7 @@ pub enum ReplicaState {
     //
     // Snapshot shard transfer is in progress, updates aren't sent to the shard
     // Normally rejects updates. Since 1.8 it allows updates if force is true.
-    // TODO(1.10): remove PartialSnapshot state entirely?
```
agourlay, line 1065: Why is now a good time to do this change?

timvisee: My original plan was to remove the variant entirely. But, I don't think we can as some of our logic still relies on understanding this variant.

That's why I removed the comment for removing it entirely and did just hide it from our API spec.

We could have done this before two versions ago, but we simply forgot about it.

PR #5287, opened 2024-10-22
Inference Service, billing-id

n0x29a:
Add `billing_id` for inference service.


config/development.yaml
```diff
@@ -3,6 +3,7 @@ log_level: DEBUG
 inference:
   address: "http://localhost:2114/api/v1/infer"
   timeout: 10
+  billing_id: "98eb568c-dea5-4347-a3a5-583478983bb9"
```
generall, line 6: let's call it `token`

generall (approved):

PR #5288, opened 2024-10-22
Hw counter utilization checks

JojiiOfficial:
Depends on #5261

Allows to easily detect losing hardware measurements by panicing (in debug mode) and logging a warning (in release mode) when measurements don't get processed.
This helps keeping track of many different `HardwareCounterCells` across the code and in new code.

### DIFF
--- a/lib/segment/src/data_types/query_context.rs
+++ b/lib/segment/src/data_types/query_context.rs
@@ -1,6 +1,5 @@
 use std::collections::HashMap;
 use std::ops::Deref;
-use std::rc::Rc;
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 
@@ -84,7 +83,7 @@ impl QueryContext {
         SegmentQueryContext {
             query_context: self,
             deleted_points: None,
-            hardware_counter: Rc::new(HardwareCounterCell::new()),
+            hardware_counter: HardwareCounterCell::new(),
         }
     }
 }
@@ -96,11 +95,11 @@ impl Default for QueryContext {
 }
 
 /// Defines context of the search query on the segment level
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct SegmentQueryContext<'a> {
     query_context: &'a QueryContext,
     deleted_points: Option<&'a BitSlice>,
-    hardware_counter: Rc<HardwareCounterCell>,
+    hardware_counter: HardwareCounterCell,
 }
 
 impl<'a> SegmentQueryContext<'a> {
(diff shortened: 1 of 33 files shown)


lib/segment/src/data_types/query_context.rs
```diff
@@ -83,7 +84,7 @@ impl QueryContext {
         SegmentQueryContext {
             query_context: self,
             deleted_points: None,
-            hardware_counter: HardwareCounterCell::default(),
+            hardware_counter: Rc::new(HardwareCounterCell::new()),
```
generall, line 87: Why RC is now needed?

JojiiOfficial: To not clone `SegmentQueryContext` and accidentally create multiple instances of `hardware_counter`, of which we consume one and drop the others causing a debug panic or (wrong) warning.
It also can't be a simple reference, because [here](https://github.com/qdrant/qdrant/blob/7e39692d80acfbf10b527c8664d2d4e79667578e/lib/segment/src/data_types/query_context.rs#L87) we need to create a `SegmentQueryContext` with a new instance of `HardwareCounterCell`.

If you have another suggestion, please let me know!

generall: Maybe forbid clone?

JojiiOfficial: We currently use to clone it [here](https://github.com/qdrant/qdrant/blob/master/lib/collection/src/collection_manager/holders/proxy_segment.rs#L355) and might need it in future in similar cases too, especially as `SegmentQueryContext` only holds borrowed values (except HardwareCounterCell).
Is there a specific reason you don't think RC is a good fit here?

generall: I mean, forbid clone of the `HardwareCounterCell` and use reference

JojiiOfficial: We can't use a reference because then we won't be able to create a new `SegmentQueryContext` [here](https://github.com/qdrant/qdrant/blob/7e39692d80acfbf10b527c8664d2d4e79667578e/lib/segment/src/data_types/query_context.rs#L87) as it would return a reference to the HardwareCounterCell which is only a temporary value

generall: `SegmentQueryContext` exists strictly inside a single thread, while QueryContext is shared.
If we want to get results from multiple segments, it should be aggregated

JojiiOfficial: In that case I think it is not avoidable to use `RC` here, unless we make `SegmentQueryContext` non-clone and find a workaround for the existing code calling `clone`.
I prefer the RC approach because it gives us more flexibility without real drawbacks.
What do you think?

generall: > In that case I think it is not avoidable to use RC here, unless we make SegmentQueryContext non-clone and find a workaround for the existing code calling clone.

something like this:
https://github.com/qdrant/qdrant/pull/5288/commits/e269f68cb567193ecba6df0cc8ed37988716739b

generall (approved):
Note: counter for quantized scorer needs to be implemented (in another PR)

PR #5297, opened 2024-10-23
Vulkan api wrapper crate

IvanPleshkov:
First step to integrate GPU acceleration into dev branch. This PR adds a new workspace member `gpu` which is a wrapper around Vulkan API and helps to avoid complicated and boilerplate work with Vulkan API.


Cargo.toml
```diff
@@ -37,6 +37,7 @@ tokio-tracing = ["tokio/tracing"]
 stacktrace = ["rstack-self"]
 chaos-testing = []
 data-consistency-check = ["collection/data-consistency-check"]
+gpu = ["gpu/gpu"]
```
IvanPleshkov, line 40: GPU cargo feature is defined here

lib/gpu/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+authors = [
+    "Andrey Vasnetsov <vasnetsov93@gmail.com>",
+    "Ivan Pleshkov <pleshkov.ivan@gmail.com>",
+    "Qdrant Team <info@qdrant.tech>",
+]
+license = "Apache-2.0"
+edition = "2021"
+
+[features]
+gpu = [
```
IvanPleshkov, line 13: I decided to include crate anyway but if `gpu` feature is disabled, all codes and dependencies are not presented, crate will be empty. It allows to skip GPU for regular build and allow to use cargo chef without errors (cargo prepare wants to include this crate even if feature is disabled)

lib/gpu/src/context.rs
```diff
@@ -0,0 +1,408 @@
...
+                // This approach may cause memory leaks and used gpu kernels,
+                // but it's better than potential segfault.
+                log::error!("Failed to wait for GPU context to finish");
+
+                // Error was logged, do memory leak to keep the gpu running.
+                let resources = self.resources.clone();
+                self.resources.clear();
+                for resource in resources.into_iter() {
+                    // !!!!!!!!!
+                    std::mem::forget(resource);
```
IvanPleshkov, line 371: I need additional eyes here. If a shader has an infinity loop, I allow to run it and leak all gpu related resources with proper log error.

lib/gpu/src/basic_test.rs
```diff
@@ -0,0 +1,176 @@
+static SHADER_CODE: &str = "
```
IvanPleshkov, line 1: Test that shows how to work with this GPU-wrapper crate

lib/gpu/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "gpu"
+version = "0.0.1"
+authors = [
+    "Andrey Vasnetsov <vasnetsov93@gmail.com>",
```
generall, line 5:
```suggestion
```

generall:
`cargo test --features "gpu"` fails with

```
called `Result::unwrap()` on an `Err` value: Other("Layer VK_LAYER_KHRONOS_validation not found")
```

All instructions I found by googling are windows-related.

generall:
Test seems to start working after installing some of those:

```
sudo apt install vulkan-tools
lunarg-via
lunarg-vkconfig
lunarg-vulkan-layers
vulkancapsviewer
vulkan-headers
libvulkan-dev
```

however installing `vulkan-sdk` didn't work because of dependency conflict on pop-os

generall:
While running test, I can't notice any process mentioned in `nvidia-smi`, is that expected? If so, how is memory accounted?

Additionally, I tried to change `let numbers_count = 512;` to some higher number like `2048`, but the test just stuck forever without any errors. Is that expected error handling in gpu?

IvanPleshkov:
> `cargo test --features "gpu"` fails with
> 
> ```
> called `Result::unwrap()` on an `Err` value: Other("Layer VK_LAYER_KHRONOS_validation not found")
> ```
> 
> All instructions I found by googling are windows-related.

This error means that you have OK drivers but you need Vulkan SDK which has `VK_LAYER_KHRONOS_validation` extension. This extension is check that all API calls are correct, NVidia does not provide this extension by default, it's only presented in Vulkan SDK.

You may find instructions here: [VK_LAYER_KHRONOS_validation](https://vulkan.lunarg.com/sdk/home#linux)
Vulkan SDK is developed by LunarG company and basically linux package manager does not have it in repositories (exceptions: ubuntu 20/22).

IvanPleshkov:
> Test seems to start working after installing some of those:
> 
> ```
> sudo apt install vulkan-tools
> lunarg-via
> lunarg-vkconfig
> lunarg-vulkan-layers
> vulkancapsviewer
> vulkan-headers
> libvulkan-dev
> ```
> 
> however installing `vulkan-sdk` didn't work because of dependency conflict on pop-os

As mentioned in another comment, SDK is developed by LunarG and can be found here as a `.tar.xz` archive: https://vulkan.lunarg.com/sdk/home#linux

IvanPleshkov:
> While running test, I can't notice any process mentioned in `nvidia-smi`, is that expected? If so, how is memory accounted?
> 
> Additionally, I tried to change `let numbers_count = 512;` to some higher number like `2048`, but the test just stuck forever without any errors. Is that expected error handling in gpu?

Thanks for checking! I already fixed this stuck in magic branch but didbn't cherry-picked it to this PR. Updated, now it returns error and don't stuck. Error is fine here, test creates one thread per number and a threads limit may appear. Also 2048 will work fine because I changed bounding checking too.

IvanPleshkov:
> While running test, I can't notice any process mentioned in `nvidia-smi`, is that expected? If so, how is memory accounted?

This may happen if test is running under GPU emulator on CPU (LLVM emulator is included into `libvulkan-dev`). In this case I find it ok on current stage, it's a host setup, not an incorrect test - test try to use any found GPU even emulated. On hnsw branch of course I skip emulated and integrated gpus by default.

I can run this test on all found GPUs. If you have a set of devices (discrete, integrated, cpu-emulated), we may run test on all of them.

generall (approved):
I think it is fine to merge from the build pipeline point of view, I have no expertise to evaluate GPU-related code

timvisee:
I had a comment and commit pending :smile: 

https://github.com/qdrant/qdrant/commit/301cb71019d5302f0c1985a424549f2210e5cda9

They're not very significant though.

IvanPleshkov:
> I had a comment and commit pending 😄
> 
> [301cb71](https://github.com/qdrant/qdrant/commit/301cb71019d5302f0c1985a424549f2210e5cda9)
> 
> They're not very significant though.

I applied your suggestions in the next PR with GPU vector storage:
https://github.com/qdrant/qdrant/pull/5339/commits/007722c5ce202f1524927e0296a56852785fc32e

PR #5299, opened 2024-10-23
Tiny refactoring during data race investigation

timvisee:
Tiny improvements I made on top of <https://github.com/qdrant/qdrant/pull/5298>.

To prevent cluttering the main PR - making the problem harder to understand - I moved them to a separate PR.

These changes are not significant and don't include any logic changes.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/collection_updater.rs
+++ b/lib/collection/src/collection_manager/collection_updater.rs
@@ -20,8 +20,7 @@ impl CollectionUpdater {
             Ok(_) => {
                 if !segments.read().failed_operation.is_empty() {
                     // If this operation failed before, remove it because it got fixed now
-                    let mut write_segments = segments.write();
-                    write_segments.failed_operation.remove(&op_num);
+                    segments.write().failed_operation.remove(&op_num);
                 }
             }
             Err(collection_error) => {
--- a/lib/collection/src/collection_manager/collection_updater.rs
+++ b/lib/collection/src/collection_manager/collection_updater.rs
@@ -19,11 +19,9 @@ impl CollectionUpdater {
         match operation_result {
             Ok(_) => {
                 if !segments.read().failed_operation.is_empty() {
+                    // If this operation failed before, remove it because it got fixed now
                     let mut write_segments = segments.write();
-                    if write_segments.failed_operation.contains(&op_num) {
-                        // Failed operation successfully fixed
-                        write_segments.failed_operation.remove(&op_num);
-                    }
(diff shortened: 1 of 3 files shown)


lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -305,7 +305,7 @@ impl ShardOperation for ForwardProxyShard {
             let tick = result.clock_tag.map(|tag| tag.clock_tick);
             let remote_tick = remote_result.clock_tag.map(|tag| tag.clock_tick);
 
-            if remote_tick > tick || tick.is_none() {
```
agourlay, line 308: why is this change fine?

timvisee: The simplified condition behaves slightly different on inputs:

| `remote_tick` | `tick`   | before | after |
|---------------|----------|------  |-------|
| `None`          | `None`     | true   | false |
| `Some`          | `None`     | true   | true  |
| `None`          | `Some`     | false  | false |
| `Some(1)`       | `Some(2)`  | false  | false |
| `Some(2)`       | `Some(1)`  | true   | true  |
| `Some(1)`       | `Some(1)`  | false  | false |

There's one discrepancy as you can see. However, in that case the clock tag itself will be exactly the same, and so the end result is exactly the same.

Even though it's fine, I'll revert my change because the previous version may be a bit easier to understand.

lib/collection/src/collection_manager/collection_updater.rs
```diff
@@ -19,11 +19,9 @@ impl CollectionUpdater {
         match operation_result {
             Ok(_) => {
                 if !segments.read().failed_operation.is_empty() {
+                    // If this operation failed before, remove it because it got fixed now
                     let mut write_segments = segments.write();
-                    if write_segments.failed_operation.contains(&op_num) {
-                        // Failed operation successfully fixed
-                        write_segments.failed_operation.remove(&op_num);
-                    }
+                    write_segments.failed_operation.remove(&op_num);
```
KShivendu, line 24: Nitpicking but maybe call it something like `segments_write_guard`. But upto you :)

KShivendu (approved):

lib/collection/src/collection_manager/collection_updater.rs
timvisee: I removed the intermediate write guard entirely, so no naming is needed anymore :+1:

PR #5301, opened 2024-10-23
[perf] Do not reallocate RocksDB write_options

agourlay:
Do not reallocate a native RocksDB `write_options` object on every `put` call.

According to a benchmark, it takes up to 32% of the memory allocation during payload + index upsert.

write-alloc

### DIFF
--- a/lib/segment/src/common/rocksdb_wrapper.rs
+++ b/lib/segment/src/common/rocksdb_wrapper.rs
@@ -1,3 +1,4 @@
+use std::fmt::Debug;
 use std::path::Path;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


timvisee (approved):
>  it takes up to 32% of the memory allocation during payload + index upsert.

:open_mouth:

lib/segment/src/common/rocksdb_wrapper.rs
```diff
@@ -230,7 +242,7 @@ impl DatabaseColumnWrapper {
         Ok(db.cf_handle(&self.column_name).is_some())
     }
 
-    fn get_write_options() -> WriteOptions {
+    fn make_write_options() -> WriteOptions {
```
coszio, line 245: Better semantic! 👍

coszio (approved):

agourlay:
Here is the flamegraph after. The allocations are completely gone.
The system feels faster too but I did not benchmark it formally.

after

PR #5302, opened 2024-10-23
Fix off by one in WAL delta causing operation to be sent twice

timvisee:
Improvement on <https://github.com/qdrant/qdrant/pull/5271>.

If we resolve an empty WAL delta we still create a queue proxy to transmit new updates.

The queue proxy expects the version to start sending updates from. Before this PR we gave it our last version, which means the last operation would be sent a second time.

This is not necessary and we can bump the version to only transmit new updates.

I've tested this change locally on 1000 WAL delta transfers.

### All Submissions:

### New Feature Submissions:


agourlay (approved):

agourlay:
> I've tested this change locally on 1000 WAL delta transfers.

Can we somehow add those tests to our integration tests?

PR #5305, opened 2024-10-24
[perf] No backtrace on sparse vector index version mismatch

agourlay:
The sparse vector index is created if it does not exist when trying to open it.

When the data path does not exist, an error regarding mismatched version is created,

The failure is not exceptional and does not require an expensive backtrace which is not returned to the user.

### DIFF
--- a/lib/segment/src/common/operation_error.rs
+++ b/lib/segment/src/common/operation_error.rs
@@ -69,6 +69,14 @@ impl OperationError {
             backtrace: Some(Backtrace::force_capture().to_string()),
         }
     }
+
+    /// Create a new service error with a description and no backtrace
+    pub fn service_error_light(description: impl Into<String>) -> OperationError {
+        OperationError::ServiceError {
+            description: description.into(),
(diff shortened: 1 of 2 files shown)


lib/segment/src/common/operation_error.rs
```diff
@@ -69,6 +69,14 @@ impl OperationError {
             backtrace: Some(Backtrace::force_capture().to_string()),
         }
     }
+
+    /// Create a new service error with a description and no backtrace
+    pub fn service_error_light(description: impl Into<String>) -> OperationError {
```
timvisee, line 74: How about naming this `service_error` and the other `service_error_with_backtrace`? Or are there just too many.

timvisee (approved):

lib/segment/src/common/operation_error.rs
agourlay: I thought about this but there are 155 calls, quite the noisy PR.

We could try to migrate some call sites to the light version over time.

PR #5306, opened 2024-10-24
Add hardware measurements in API

JojiiOfficial:
Depends on #5288

Adds information about hardware usage to the API and a config to toggle this.

### DIFF
--- a/lib/api/src/grpc/models.rs
+++ b/lib/api/src/grpc/models.rs
@@ -43,6 +43,14 @@ pub struct ApiResponse<D> {
     pub result: Option<D>,
     pub status: ApiStatus,
     pub time: f64,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub hardware_usage: Option<HardwareUsage>,
+}
+
+#[derive(Debug, Serialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct HardwareUsage {
+    pub cpu: usize,
 }
 
 #[derive(Debug, Serialize, JsonSchema)]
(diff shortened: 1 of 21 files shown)


src/settings.rs
```diff
@@ -50,6 +50,10 @@ pub struct ServiceConfig {
 
     /// How much time is considered too long for a query to execute.
     pub slow_query_secs: Option<f32>,
+
+    /// Whether to enable reporting of measured hardware utilization in API responses.
+    #[serde(default)]
+    pub hardware_reporting: bool,
```
generall, line 56: Let's make it optional, usually it helps for maintaining param compatibility in future

lib/api/src/grpc/models.rs
```diff
@@ -43,6 +43,14 @@ pub struct ApiResponse<D> {
     pub result: Option<D>,
     pub status: ApiStatus,
     pub time: f64,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub hardware_usage: Option<HardwareUsage>,
+}
+
+#[derive(Debug, Serialize, JsonSchema)]
```
generall, line 50: Not for this PR, but it looks like misplacement of models. Since models implement `JsonSchema`, they shouldn't be in grpc subfolder

src/actix/mod.rs
```diff
@@ -142,6 +146,7 @@ pub fn init(
                 .app_data(validate_json_config)
                 .app_data(TempFileConfig::default().directory(&upload_dir))
                 .app_data(MultipartFormConfig::default().total_limit(usize::MAX))
+                .app_data(hardware_reporting_settings.clone())
```
generall, line 149: Extra argument just for this bool seems like an overkill.

How about we either put it into global static, or pass the whole settings object?

generall (approved):

PR #5307, opened 2024-10-24
Use Jemalloc unprefixed_malloc_on_supported_platform

agourlay:
With this feature, both Rust code and linked C/C++ code will use jemalloc for memory allocation.

This would explain why the memory reported by jemalloc was too low, because it is missing RocksDB related allocations.

I can see the impact of this change in Heaptrack which seems unable to track Jemalloc allocations.
Before this change I see only RocksDB allocations and after I see nothing :)


generall (approved):

agourlay:
With this change I now see the memory allocated by RocksDB in the jemalloc profile.
This confirms that this PR as unified the allocator.

rock-mem

PR #5313, opened 2024-10-25
Distance matrix does not need to use client's filter when looking up by ids

agourlay:
It is not needed to build a merged filter using the client's filter when selecting by point ids.

We have already applied this filter to select the point ids.

This PR also adds a TODO for the next possible optimization.


coszio (approved):
Haha! Good catch, we don't actually need to re-filter after the first sampling

PR #5314, opened 2024-10-25
Leverage shared filter in distance matrix API

agourlay:
In order to fully leverage the batch search, all search requests should use the same filter.

In this PR the distance matrix API is changed to not create a dedicated specialized filter for scoring each sample.

A single filter using all point ids is reused with additional oversampling to account for scoring against itself.

At the end a postprocessing step is required to filter the results.

### DIFF
--- a/lib/collection/src/collection/distance_matrix.rs
+++ b/lib/collection/src/collection/distance_matrix.rs
@@ -214,29 +214,28 @@ impl Collection {
         sampled_points.truncate(sample_size);
         // sort by id for a deterministic order
         sampled_points.sort_unstable_by(|(id1, _), (id2, _)| id1.cmp(id2));
+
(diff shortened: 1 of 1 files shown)


coszio (approved):
Now batching can do its work better 👍

generall:
Confirm, that now batch request re-uses filters: internal counter reports one event per matrix request

```
                        "filtered_plain": {
                          "count": 1,
                          "avg_duration_micros": 11362,
                          "min_duration_micros": 11362,
                          "max_duration_micros": 11362,
                          "total_duration_micros": 11362,
                          "last_responded": "2024-10-25T15:47:23.560Z"
                        }
```

Previously it was same as sample size

generall (approved):

PR #5328, opened 2024-10-29
Hardware utilization information in endpoints

JojiiOfficial:
Depends on #5308
Adds hardware information for requests to more endpoints in the REST API

### DIFF
--- a/lib/collection/src/collection/collection_ops.rs
+++ b/lib/collection/src/collection/collection_ops.rs
@@ -1,6 +1,7 @@
 use std::cmp;
 use std::sync::Arc;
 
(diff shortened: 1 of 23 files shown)


lib/collection/src/collection/collection_ops.rs
```diff
@@ -341,7 +342,7 @@ impl Collection {
                     .copied()
                     .unwrap_or(ReplicaState::Dead);
                 let count_result = replica_set
-                    .count_local(count_request.clone(), None)
+                    .count_local(count_request.clone(), None, HwMeasurementAcc::new())
```
generall, line 345: Does it mean that we ignore measurement here?

PR #5331, opened 2024-10-29
batch increment of counter

generall:
Benchmarks incomming

### DIFF
--- a/lib/sparse/src/index/search_context.rs
+++ b/lib/sparse/src/index/search_context.rs
@@ -148,12 +148,6 @@ impl<'a, 'b, T: PostingListIter> SearchContext<'a, 'b, T> {
         self.pooled.scores.clear(); // keep underlying allocated memory
         self.pooled.scores.resize(batch_len as usize, 0.0);
 
-        {
-            // Measure CPU usage of indexed sparse search.
-            let cpu_counter = self.hardware_counter.cpu_counter_mut();
-            cpu_counter.incr_delta(batch_len as usize);
-        }
-
         for posting in self.postings_iterators.iter_mut() {
             posting.posting_list_iterator.for_each_till_id(
                 batch_last_id,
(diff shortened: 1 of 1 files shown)


lib/sparse/src/index/search_context.rs
```diff
@@ -258,6 +253,17 @@ impl<'a, 'b, T: PostingListIter> SearchContext<'a, 'b, T> {
...
         }
+
+        {
+            // Measure CPU usage of indexed sparse search.
+            // Assume the complexity of the search as total volume of the posting lists
+            // that are traversed in the batched search.
+            let cpu_counter = self.hardware_counter.cpu_counter_mut();
+            for posting in self.postings_iterators.iter() {
+                cpu_counter.incr_delta_mut(posting.posting_list_iterator.len_to_end());
+            }
```
timvisee, line 264: The original does count on `for_each_till_id(batch_last_id, ...)`. The new implementation counts `.len_to_end()`. Are they the same?

generall: According to tests, yes. But overall I think it is correct to estimate search complexity as a sum of posting length

lib/sparse/src/index/search_context.rs
```diff
@@ -161,9 +159,6 @@ impl<'a, 'b, T: PostingListIter> SearchContext<'a, 'b, T> {
                     // SAFETY: `id` is within `batch_start_id..=batch_last_id`
                     // Thus, `local_id` is within `0..batch_len`.
                     *unsafe { scores.get_unchecked_mut(local_id) } += element_score;
-
-                    // Measure CPU usage of indexed sparse search.
-                    cpu_counter.incr_mut();
```
xzfc, line 166: This is a hot loop indeed and removing this line gives ~4-10% percents on `search/mmap/random-50k` on my machine.

timvisee (approved):

PR #5338, opened 2024-10-30
Use secure base image

KShivendu:
Should make the container more secure.

### All Submissions:


KShivendu:
Simply changing the base image failed. Will try by building everything with debian and copying to the secure image in the end.

timvisee:
IMO we should create an image from scratch, with no base image at all. That way the only vulnerabilities it can contain are the ones in the code being used.

The downside is that it'll remove all utilities from the image that are useful during debug sessions, such as `bash`, `sh` or `curl`.

PR #5343, opened 2024-10-31
Fix killing replicas too earily on state switches

timvisee:
We kill other replicas too eagerly if we think there's a problem.

We mark a replica as Dead if we get an error while sending an update to it (such as a precondition failure) and we think the replica is Active. The idea here is that Active replicas should never give errors.

The key thing here is that peers might be slightly out of sync, meaning that consensus operations are not applied at exactly the same time across a fleet.

So while we think we get an error on an Active replica, the actual replica might already be in Recovery state. In that case we expect to get an error back and should not escalate to killing the replica like we do now.

To solve this we now send the current state we expect to consensus when killing a replica. If it is in that state, we still mark it as dead. If it is not, we ignore the message because our peer was lagging behind.

Before this PR I saw a replica getting marked as dead about every 2 seconds while sending high load to it with concurrent shard transfers. After this PR I do not see dead replicas at all anymore. This is a great stability improvement and may prevent a lot of unnecessary replica recoveries.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -39,7 +39,9 @@ use crate::shards::channel_service::ChannelService;
 use crate::shards::collection_shard_distribution::CollectionShardDistribution;
 use crate::shards::local_shard::clock_map::RecoveryPoint;
 use crate::shards::replica_set::ReplicaState::{Active, Dead, Initializing, Listener};
-use crate::shards::replica_set::{ChangePeerState, ReplicaState, ShardReplicaSet};
+use crate::shards::replica_set::{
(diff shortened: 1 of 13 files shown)


generall (approved):
Why I think this change is safe to accept, and we should not get a situation, where storage gets inconsistent and we fail to report that:

- Local state will be erased is state changes by consensus. Meaning that local state will either be applied, or be replaced with actual state from the consensus.

- Change only have effect if expected state is one of: 

```
            ReplicaState::Active
            | ReplicaState::Dead
            | ReplicaState::Initializing
            | ReplicaState::Listener => false,
```

Which means that it can only transition into state, after which the recovery will follow. 

(ToDo: what would happen if we have Active -> Listener transition? Do we even care?)

PR #5346, opened 2024-11-01
Fix immutable map index mishandling point deletions

timvisee:
Fix a problem where the immutable map index is mishandling point deletions. This was introduced in <https://github.com/qdrant/qdrant/pull/5072>.

The problem is that while deleting points, the index of a subslice is used. We should use the global index on the container instead.

### All Submissions:

### New Feature Submissions:


generall (approved):

timvisee:
Thanks for the test!

PR #5349, opened 2024-11-01
Experiment: ignore clock tags when replica is in partial state

timvisee:
It seems that clock tags can be problematic if a replica is in partial state. That is because we might receive updates in arbitrary order from any node, resulting in clock gaps.

This creates two problems:
- during this state we may reject operations we do need
- if we abort an ongoing shard transfer our recovery point may be incorrect

To fix this we now simply ignore clock tags in partial state, and accept all operations.

I've implemented this in quite a naive way. I'd like to merge it into our dev branch so we can test it over the weekend. I plan to implement it in a better way next week.

In my local testing this seems to fix a critical problem with stream records and WAL delta transfers. Before the PR I got a failure every few minutes. With the PR I've been running the test for two hours no without problems.

While I do expect this to improve chaos testing, I'm not confident this fixes all the issues we're seeing.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -554,10 +554,12 @@ mod tests {
         // at build time the replicas are all dead, they need to be activated
         assert_eq!(rs.highest_alive_replica_peer_id(), None);
 
-        rs.set_replica_state(1, ReplicaState::Active).unwrap();
-        rs.set_replica_state(3, ReplicaState::Active).unwrap();
-        rs.set_replica_state(4, ReplicaState::Active).unwrap();
-        rs.set_replica_state(5, ReplicaState::Partial).unwrap();
+        rs.set_replica_state(1, ReplicaState::Active).await.unwrap();
+        rs.set_replica_state(3, ReplicaState::Active).await.unwrap();
+        rs.set_replica_state(4, ReplicaState::Active).await.unwrap();
+        rs.set_replica_state(5, ReplicaState::Partial)
+            .await
+            .unwrap();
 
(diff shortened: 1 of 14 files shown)


lib/collection/src/wal_delta.rs
```diff
@@ -146,6 +154,11 @@ impl RecoverableWal {
         }
         Ok(())
     }
+
+    // TODO(timvisee): provide a better interface for this, pass param into `lock_and_write` instead?
```
timvisee, line 158: Left this here because I want to implement this in a better way once I have the time for it. I plan to do so next week.

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -422,12 +422,9 @@ impl ShardReplicaSet {
...
             Some(ReplicaState::Listener) => true,
-            // Recovery: keep sending updates to prevent a data race
-            // The replica on the peer may still be active for some time if its consensus is slow.
-            // The peer may respond to read requests until it switches to recovery state too. We
-            // must keep sending updates to prevent those reads being stale.
-            // See: <https://github.com/qdrant/qdrant/pull/5298>
-            Some(ReplicaState::Recovery | ReplicaState::PartialSnapshot) => true,
+            // We must not send updates to replicas in recovery state.
+            // If we do we might create gaps in WAL clock tags.
+            Some(ReplicaState::Recovery | ReplicaState::PartialSnapshot) => false,
```
generall, line 427: This is the reversion of <https://github.com/qdrant/qdrant/pull/5298> for good

generall (approved):

PR #5350, opened 2024-11-04
Fix consensus sync peers waiting on wrong value

timvisee:
When synchronizing the consensus commit and peer we were waiting on the wrong commit value.

We waited on the commit in hard state, but we should wait on the last applied entry instead. The last applied entry better reflects what collection metadata we're actually seeing, while the hard state commit might be somewhere in the future.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/storage/src/content_manager/consensus/persistent.rs
+++ b/lib/storage/src/content_manager/consensus/persistent.rs
@@ -225,6 +225,13 @@ impl Persistent {
         self.apply_progress_queue.get_last_applied()
     }
 
+    /// Get the last applied commit and term, reflected in our current state.
+    pub fn applied_commit_term(&self) -> (u64, u64) {
+        let hard_state = &self.state().hard_state;
+        let last_commit = self.last_applied_entry().unwrap_or(hard_state.commit);
+        (last_commit, hard_state.term)
+    }
+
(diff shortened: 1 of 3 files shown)


KShivendu (commented):
Seems correct code-wise. But I have the same doubt as Andrey, why is it not better to wait for a higher commit value? In the worse case, there will be a timeout but even that it was waiting for the same time 🤔

Maybe we should pick one of the scenarios (one with snapshot or wal delta transfer) to see if this actually makes a difference?

timvisee:
> But I have the same doubt as Andrey, why is it not better to wait for a higher commit value? In the worse case, there will be a timeout but even that it was waiting for the same time 🤔

I'm quite confused by this as well, and have no answer to this yet. What I can say is that making this change did have a clear effect on my repro.

Even though this is not entirely clear yet I do believe it's better to change it because this is how it should have been from the start.

> Maybe we should pick one of the scenarios (one with snapshot or wal delta transfer) to see if this actually makes a difference?

If you mean - testing in chaos testing - then yes I definitely vote for that.

ffuugoo (approved):

generall (approved):
I got it. It is what we ask other nodes to wait for, not what other nodes use to wait on

PR #5351, opened 2024-11-04
Do not sort sparse vector before remapping it

agourlay:
Sparse vectors indices are remapped and sorted by internal indices right before the search.

Therefore it is not necessary to sort the initial sparse vector.

Running with CI benchmark does not show any improvement sadly.

### DIFF
--- a/lib/segment/src/index/sparse_index/sparse_vector_index.rs
+++ b/lib/segment/src/index/sparse_index/sparse_vector_index.rs
@@ -320,7 +320,7 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
 
     pub fn search_plain(
         &self,
-        sparse_vector: &SparseVector,
+        sparse_vector: SparseVector,
         filter: &Filter,
         top: usize,
         prefiltered_points: &mut Option<Vec<PointOffsetType>>,
@@ -349,7 +349,7 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         .filter(|&idx| check_deleted_condition(idx, deleted_vectors, deleted_point_bitslice))
         .collect_vec();
 
-        let sparse_vector = self.indices_tracker.remap_vector(sparse_vector.to_owned());
+        let sparse_vector = self.indices_tracker.remap_vector(sparse_vector);
         let memory_handle = self.scores_memory_pool.get();
         let mut search_context = SearchContext::new(
             sparse_vector,
@@ -366,7 +366,7 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
     // search using sparse vector inverted index
     fn search_sparse(
         &self,
-        sparse_vector: &SparseVector,
+        sparse_vector: SparseVector,
         filter: Option<&Filter>,
         top: usize,
         vector_query_context: &VectorQueryContext,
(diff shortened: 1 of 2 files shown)


lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -334,7 +334,7 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
 
         let deleted_point_bitslice = vector_query_context
             .deleted_points()
-            .unwrap_or(id_tracker.deleted_point_bitslice());
+            .unwrap_or_else(|| id_tracker.deleted_point_bitslice());
```
timvisee, line 337: This probably is not necessary.

The function just returns a reference and we don't do any business logic.

Since we use LTO we can assume it yeets all the intermediate function calls.

However, since we use LTO I don't expect it to be any worse either because I expect it to also remove the closure.

I have a slight preference to keep the previous. Thoughts?

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -349,7 +349,7 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         .filter(|&idx| check_deleted_condition(idx, deleted_vectors, deleted_point_bitslice))
         .collect_vec();
 
-        let sparse_vector = self.indices_tracker.remap_vector(sparse_vector.to_owned());
+        let sparse_vector = self.indices_tracker.remap_vector(sparse_vector.clone());
```
timvisee, line 352: While I'm fine with this change, I believe it's exactly the same on assembly level.

timvisee (commented):
It is unclear to me what this actually changes. See my above comments. Would you mind to elaborate?

More specifically, it's unclear where sorting happened before, and how we prevent it now.

lib/segment/src/index/sparse_index/sparse_vector_index.rs
agourlay: Yes the calls are equivalent.
I prefer calling explicitly `clone` in that case because we do not depend on the `ToOwned` trait abstraction.

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -424,8 +424,6 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         if vector.is_empty() {
             return Ok(vec![]);
         }
-        let mut vector = vector.clone();
```
agourlay, line 427: @timvisee here is the meaningful change

timvisee: Woah, that clearly shows 'sort' as well! The rest of the diff just distracted me.

Thanks!

lib/segment/src/index/sparse_index/sparse_vector_index.rs
timvisee: One argument may be: this will prevent making it costly if we ever do add business logic in those functions in the future. But, I'm not sure if that will ever happen.

timvisee (approved):

lib/segment/src/index/sparse_index/sparse_vector_index.rs
agourlay: My bad, I did not push a nice diff on this one, sorry.

PR #5352, opened 2024-11-04
Experiment: also don't write clock tags to WAL in partial state

timvisee:
Extends experiment in <https://github.com/qdrant/qdrant/pull/5349>.

Don't just ignore the clock tag in partial state, but also prevent writing it in the WAL.

Otherwise we'll still bump our last seen clocks with the tags replayed from the WAL.

### All Submissions:


generall (approved):

generall:
This change is potentially have impact on the ability to use wal-delta transfer multiple times in a row, so we should not release version with this patch

timvisee:
I'll prepare a revert PR so we can easily undo the experiment.

PR #5353, opened 2024-11-04
Revert experiment: ignore WAL clocks in partial state

timvisee:
This reverts an experiment, consisting of these PRs:
- https://github.com/qdrant/qdrant/pull/5349
- https://github.com/qdrant/qdrant/pull/5352
- https://github.com/qdrant/qdrant/pull/5372
- https://github.com/qdrant/qdrant/pull/5379
- https://github.com/qdrant/qdrant/pull/5375

It does keep the change to not send updates to replicas in recovery state ([comment](https://github.com/qdrant/qdrant/pull/5353#discussion_r1827777721)).

Some of these changes are significant, and so the plan is to re-implement them a bit later. Reverting them now allows us to do a patch release excluding these changes.

### DIFF
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -39,29 +39,19 @@ impl ShardReplicaSet {
         if let Some(local_shard) = local.deref() {
             match self.peer_state(self.this_peer_id()) {
                 Some(
-                    state @ (ReplicaState::Active
+                    ReplicaState::Active
                     | ReplicaState::Partial
                     | ReplicaState::Initializing
-                    | ReplicaState::Resharding),
-                ) => {
-                    let ignore_local_clocks = state.is_ignore_local_clocks();
-                    Ok(Some(
-                        local_shard
-                            .get()
-                            .update(operation, wait, ignore_local_clocks)
-                            .await?,
-                    ))
+                    | ReplicaState::Resharding,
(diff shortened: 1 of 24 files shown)


lib/collection/src/shards/replica_set/update.rs
```diff
@@ -424,9 +424,12 @@ impl ShardReplicaSet {
...
             Some(ReplicaState::Listener) => true,
-            // We must not send updates to replicas in recovery state.
-            // If we do we might create gaps in WAL clock tags.
-            Some(ReplicaState::Recovery | ReplicaState::PartialSnapshot) => false,
+            // Recovery: keep sending updates to prevent a data race
+            // The replica on the peer may still be active for some time if its consensus is slow.
+            // The peer may respond to read requests until it switches to recovery state too. We
+            // must keep sending updates to prevent those reads being stale.
+            // See: <https://github.com/qdrant/qdrant/pull/5298>
+            Some(ReplicaState::Recovery | ReplicaState::PartialSnapshot) => true,
```
generall, line 432: Do you think we should revert this part as well?

timvisee: I'm indecisive thus far, I left a comment about this in my notes.

Based on what we've discussed before I'd be happy to keep it at false.

timvisee: Kept this at false - we still do not send updates to replicas in recovery state.

generall (approved):

PR #5336, opened 2024-10-30
VisitedList: Vec<usize> -> Vec<u8>

xzfc:
This PR optimizes `VisitedList` by switching counters from `usize` to `u8` (and memset them every 256 iterations). Surprisingly, this change improves not only the memory usage, but also the search/build performance.

```console
$ cargo bench -p segment --bench hnsw_search_graph # old
hnsw-index-search-group/hnsw_search
                        time:   [634.32 µs 640.56 µs 646.44 µs]
…

$ cargo bench -p segment --bench hnsw_search_graph # new
hnsw-index-search-group/hnsw_search
                        time:   [596.08 µs 601.57 µs 606.93 µs]
```

Large graphs are also improved. Graph build time (dim=4, npoints=10_000_000): 892s → 640s.

PR #5365, opened 2024-11-05
Fixups for VistedPool

xzfc:
The PR #5336 introduced a in `VisitedListHandle` bug which is reproducible after two `u8` wraparounds.
This PR fixes it by reseting `visit_counters` every `255` iterations, not every `256`.


agourlay:
What about having a unit test for this case?

generall:
There were a test in python client, which cough that. It would be better to have unit test of course.

lib/segment/src/index/visited_pool.rs
```diff
@@ -142,7 +149,7 @@ mod tests {
             assert!(visited_list.check_and_update_visited(0));
             assert!(visited_list.check(0));
 
-            for _ in 0..260 {
+            for _ in 0..(u8::MAX as usize * 2 + 10) {
```
generall, line 152: Oh, I think this is a test already

generall (approved):

PR #5367, opened 2024-11-05
Update collection config, when applying Raft snapshot

ffuugoo:
Currently, when applying Raft snapshot, we only update collection optimizer config from the snapshot. This PR extends this, so that we update collection config in full.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/state_management.rs
+++ b/lib/collection/src/collection/state_management.rs
@@ -4,7 +4,7 @@ use crate::collection::payload_index_schema::PayloadIndexSchema;
 use crate::collection::Collection;
 use crate::collection_state::{ShardInfo, State};
 use crate::config::CollectionConfig;
-use crate::operations::types::CollectionResult;
+use crate::operations::types::{CollectionError, CollectionResult};
 use crate::shards::replica_set::ShardReplicaSet;
 use crate::shards::shard::{PeerId, ShardId};
 use crate::shards::shard_holder::{ShardKeyMapping, ShardTransferChange};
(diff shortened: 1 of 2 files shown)


lib/collection/src/collection/state_management.rs
```diff
@@ -64,18 +72,66 @@ impl Collection {
...
-            config.params.replication_factor = new_config.params.replication_factor;
-            config.params.write_consistency_factor = new_config.params.write_consistency_factor;
+
+            if let Err(err) = config.params.check_compatible(&new_config.params) {
+                // Stop consensus with a service error, if new config is incompatible with current one.
+                //
+                // We expect that `apply_config` is only called when configs are compatible, otherwise
+                // collection have to be *recreated*.
+                return Err(CollectionError::service_error(err.to_string()));
+            }
```
timvisee, line 86: Just to be sure: we report an error, but don't rely on this error to actually recreate the collection am I right? This is more of a sanity check so that we don't actually use this function if it wouldn't work.

lib/collection/src/config.rs
```diff
@@ -116,6 +116,56 @@ impl CollectionParams {
...
+            vec![]
+        };
+        this_sparse_vectors.sort();
+
+        let mut other_sparse_vectors: Vec<_> = if let Some(sparse_vectors) = sparse_vectors {
+            sparse_vectors.keys().collect()
+        } else {
+            vec![]
+        };
+        other_sparse_vectors.sort();
```
timvisee, line 146: A hash set may be more fitting. No sorting required.

lib/collection/src/collection/state_management.rs
timvisee: Confirmed myself: correct.

timvisee (approved):

lib/collection/src/config.rs
generall: agree

generall (approved):

lib/collection/src/config.rs
```diff
@@ -116,6 +116,54 @@ impl CollectionParams {
...
+            HashSet::new()
+        };
+
+        if this_sparse_vectors != other_sparse_vectors {
+            return Err(CollectionError::bad_input(format!(
+                "sparse vectors are incompatible: \
+                 origin sparse vectors: {this_sparse_vectors:?}, \
+                 while other sparse vectors: {other_sparse_vectors:?}",
+            )));
+        }
```
ffuugoo, line 152: Ooops. Missed sparse vectors check. Thanks! 😬

PR #5333, opened 2024-10-29
Add hardware usage to grpc API

JojiiOfficial:
Depends on #5328

Adds basic hardware utilization information for most GRPC endpoints

PR #5369, opened 2024-11-05
Hardware counting for quantization

JojiiOfficial:
Depends on #5333


generall (approved):

generall:
It is a bit disappointing that our structure won't allow to avoid passing counter as function param. I tried to refactor it a bit, but the change was too big.

PR #5373, opened 2024-11-05
Update is_on_disk to restore original default behavior of quantization 

generall:
After the change in #5188 default vector storage is now mmap. This leads to the (undesirable) change of default behavior of quantization storage: instead of being loaded in RAM, it is not loaded on-disk by default, which leads to several undesirable consequences:

- Creation of the quantized vectors with default original vector configuration leads to unexpected performance degradation
- Quantized and original vectors start to compete for RAM, if disk cache is not large enough, so default configuration will degrade in performance and it is hard to diagnose in real life scenario.

Temporary workaround: always specify `always_ram` explicitly.


timvisee (approved):
Interesting side effect! Thanks for the explanation.

Approved, though note that I did not do a performance test on the before and after.

PR #5374, opened 2024-11-06
GraphLinksMmap: remove Option

xzfc:
This PR removes unnecessary usage of `Option` in `GraphLinksMmap`. It was required because this field was under `#[serde(skip)]`.

Before this PR:
- `GraphLinksMmap::mmap` is `Option<Arc<Mmap>>`.
- The trait `GraphLinks` requires implementing `Default`.
- `struct GraphLayers` implements `Serialize`/`Deserealize`, but skips some fields using `#[serde(skip)]`.

After this PR:
- `GraphLinksMmap::mmap` is `Arc<Mmap>`.
- The trait `GraphLinks` no longer requires implementing `Default`.
- Serializable fields of `GraphLayers` are moved into `GraphLayerData`.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_layers.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::cmp::max;
 use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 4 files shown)


lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -280,12 +293,21 @@ where
     }
 
     pub fn save(&self, path: &Path) -> OperationResult<()> {
```
generall, line 295: Not related to this PR, but I think we better rename this method to something like `save_params` to avoid confusion

generall (approved):

timvisee (approved):

PR #5376, opened 2024-11-06
Fix consistency check in test

agourlay:
This consistency check is specialized to find discrepancy in point ids.

There are two problems:
- Python issue not accessing dictionary element properly `x.id` which fails, should be `x['id']`
- When the point ids are the same but the payloads are different, the error message is not helpful

This PR fixes the syntax issue and reuse the logic from another test to build a diff of the content.

### DIFF
--- a/tests/consensus_tests/test_shard_wal_delta_transfer.py
+++ b/tests/consensus_tests/test_shard_wal_delta_transfer.py
@@ -1,6 +1,7 @@
 import multiprocessing
 import pathlib
 import random
+from operator import itemgetter
 from time import sleep
 
 from .fixtures import upsert_random_points, create_collection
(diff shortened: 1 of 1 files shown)


tests/consensus_tests/test_shard_wal_delta_transfer.py
```diff
@@ -39,12 +40,17 @@ def check_data_consistency(data):
         data_j = data[j]
 
         if data_i != data_j:
-            ids_i = set(x.id for x in data_i["points"])
-            ids_j = set(x.id for x in data_j["points"])
+            ids_i = set(x['id'] for x in data_i["points"])
+            ids_j = set(x['id'] for x in data_j["points"])
```
timvisee, line 44: It is not entirely clear to me why `.id` worked in most cases (?), and suddenly started to fail in your branch.

timvisee (approved):

tests/consensus_tests/test_shard_wal_delta_transfer.py
agourlay: No idea either :shrug: 

I suspect the code has not been exercised in a very long time.
This object is now clearly a dictionary.

PR #5377, opened 2024-11-06
Introduce GraphLinksRef

xzfc:
Depends on #5374.

In this PR:
1. The common logic of `GraphLinksRam` and `GraphLinksMmap` extracted into `GraphLinksView`. Now both implementations just merely provide the storage (`Vec<u8>` or `Mmap`), and all the logic operates on `&[u8]`.
2. Removed a bunch of methods from `GraphLinks` which expose implementation details.
3. The method `GraphLinks::links()` is changed to return an iterator instead of a slice. This change would allow implementations to iterate over compressed values.

The idea for 2 and 3 is taken from the branch https://github.com/qdrant/qdrant/tree/try-to-compress-hnsw-links by @IvanPleshkov.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_layers.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers.rs
@@ -161,7 +161,7 @@ impl<TGraphLinks: GraphLinks> GraphLayersBase for GraphLayers<TGraphLinks> {
         F: FnMut(PointOffsetType),
     {
         for link in self.links.links(point_id, level) {
-            f(*link);
+            f(link);
         }
     }
 
@@ -469,7 +469,7 @@ mod tests {
         assert_eq!(main_entry.level, num_levels);
 
         let total_links_0 = (0..num_vectors)
-            .map(|i| graph_layers.links.links(i as PointOffsetType, 0).len())
+            .map(|i| graph_layers.links.links(i as PointOffsetType, 0).count())
             .sum::<usize>();
 
         eprintln!("total_links_0 = {total_links_0:#?}");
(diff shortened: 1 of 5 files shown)


lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -553,60 +426,125 @@ impl GraphLinks for GraphLinksMmap {
...
+        self.view().point_level(point_id)
     }
+}
 
-    fn get_links_range(&self, idx: usize) -> Range<usize> {
-        let offsets_range = self.header.get_offsets_range();
-        let mmap: &[u8] = if let Some(mmap) = &self.mmap {
-            &mmap[offsets_range]
+#[derive(Debug)]
+struct GraphLinksView<'a> {
```
generall, line 463: Let's move this into a separate file for easier navigation

lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -449,7 +469,7 @@ mod tests {
         assert_eq!(main_entry.level, num_levels);
 
         let total_links_0 = (0..num_vectors)
-            .map(|i| graph_layers.links.links(i as PointOffsetType, 0).len())
+            .map(|i| graph_layers.links.links(i as PointOffsetType, 0).count())
```
timvisee, line 472: Nit: if we can return an exact size iterator here, we can simply do [len()](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html#method.len). 

That may be a bit nicer because it'll prevent evaluating each entry. If we add decompression to the mix it'll prevent decompressing entries on getting the length.

Count does not have this _optimization_ and will evaluate all entries

Wdyt? Would that be worth it?

generall:
Trying to run bench .. https://github.com/qdrant/vector-db-benchmark/actions/runs/11820840867/job/32940643776

PR #5378, opened 2024-11-06
Add UUID to collection config

ffuugoo:
When applying Raft snapshot, we receive a list of collections that currently exist in the cluster. However, it is _generally_ impossible to determine, if a collection was deleted and then another collection with the same name was created in its place.

E.g., imagine that
- collection "example" is created
- node A is temporarily shut down
- collection "example" is deleted
- and *another* collection, also named "example", with exactly the same configuration is created
- now node A is restarted, and receives Raft snapshot to synchronize consensus state
- this Raft snapshot will contain collection "example" with exactly the same parameters as the collection that exists on node A, and it's impossible to detect, that collection has been recreated

This PR adds a UUID field to collection config, so that it is possible to compare UUIDs of existing collection, and collection in the Raft snapshot, to determine if the collection needs to be recreated from scratch.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/state_management.rs
+++ b/lib/collection/src/collection/state_management.rs
@@ -77,6 +77,17 @@ impl Collection {
         {
             let mut config = self.collection_config.write().await;
 
+            if config.uuid != new_config.uuid {
+                return Err(CollectionError::service_error(format!(
+                    "collection {} UUID mismatch: \
+                     UUID of existing collection is different from UUID of collection in Raft snapshot: \
+                     existing collection UUID: {:?}, Raft snapshot collection UUID: {:?}",
+                    self.id,
+                    config.uuid,
+                    new_config.uuid,
+                )));
+            }
+
             if let Err(err) = config.params.check_compatible(&new_config.params) {
                 // Stop consensus with a service error, if new config is incompatible with current one.
                 //
@@ -96,6 +107,7 @@ impl Collection {
                 wal_config,
                 quantization_config,
                 strict_mode_config,
+                uuid: _,
             } = &new_config;
 
             let is_core_config_updated = params != &config.params
--- a/lib/collection/src/collection/state_management.rs
+++ b/lib/collection/src/collection/state_management.rs
@@ -3,7 +3,7 @@ use std::collections::{HashMap, HashSet};
 use crate::collection::payload_index_schema::PayloadIndexSchema;
 use crate::collection::Collection;
 use crate::collection_state::{ShardInfo, State};
-use crate::config::CollectionConfig;
+use crate::config::CollectionConfigInternal;
 use crate::operations::types::{CollectionError, CollectionResult};
 use crate::shards::replica_set::ShardReplicaSet;
 use crate::shards::shard::{PeerId, ShardId};
@@ -11,7 +11,10 @@ use crate::shards::shard_holder::{ShardKeyMapping, ShardTransferChange};
 use crate::shards::transfer::ShardTransfer;
 
 impl Collection {
-    pub async fn check_config_compatible(&self, config: &CollectionConfig) -> CollectionResult<()> {
+    pub async fn check_config_compatible(
+        &self,
+        config: &CollectionConfigInternal,
+    ) -> CollectionResult<()> {
         self.collection_config
             .read()
             .await
@@ -71,7 +74,7 @@ impl Collection {
         Ok(())
     }
 
-    async fn apply_config(&self, new_config: CollectionConfig) -> CollectionResult<()> {
+    async fn apply_config(&self, new_config: CollectionConfigInternal) -> CollectionResult<()> {
         let recreate_optimizers;
 
         {
@@ -100,7 +103,7 @@ impl Collection {
             // complain, if new field is added to `CollectionConfig` struct, but not destructured
             // explicitly. We have to explicitly compare config fields, because we want to compare
             // `wal_config` and `strict_mode_config` independently of other fields.
-            let CollectionConfig {
+            let CollectionConfigInternal {
                 params,
                 hnsw_config,
                 optimizer_config,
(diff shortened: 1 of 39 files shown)


ffuugoo:
Ready for review. Figured, that we don't really need UUID in gRPC at all, and it actually has to be excluded from OpenAPI spec. 🤦‍♀️

lib/collection/src/operations/conversions.rs
```diff
@@ -777,6 +777,7 @@ impl TryFrom<api::grpc::qdrant::CollectionConfig> for CollectionConfig {
                 }
             },
             strict_mode_config: config.strict_mode_config.map(StrictModeConfig::from),
+            uuid: None,
```
ffuugoo, line 780: This is conversion from gRPC `CollectionConfig` into Qdrant `CollectionConfig`.

As far as I can tell, it's only used as part of `CollectionInfo` (which is transferred over gRPC), but never for anything important (like consensus related stuff), so it's fine not to include it into gRPC.

The only concern is if we **do** start using it for something important in the future, it might be easy to overlook. But I don't think it's very likely.

lib/storage/src/content_manager/conversions.rs
```diff
@@ -68,6 +68,7 @@ impl TryFrom<api::grpc::qdrant::CreateCollection> for CollectionMetaOperations {
                     .map(sharding_method_from_proto)
                     .transpose()?,
                 strict_mode_config: value.strict_mode_config.map(strict_mode_from_api),
+                uuid: None,
```
ffuugoo, line 71: This is conversion from gRPC `CreateCollection` into Qdrant `CreateCollection`.

IMO, the user should not be allowed to explicitly specify collection UUID (it should be randomly generated), so it makes sense to **exclude** UUID from gRPC.

As far as I can tell gRPC `CreateCollection` is only used as part of **public** `create_collection` gRPC API, but not in internal gRPC API. When Raft propagates `CreateCollection` message to other nodes, it does not use gRPC directly, but serialize *Qdrant* `CreateCollection` structure (which contains UUID) to CBOR.

lib/storage/src/dispatcher.rs
```diff
@@ -118,6 +118,18 @@ impl Dispatcher {
...
                     }
+
+                    if let Some(uuid) = &op.create_collection.uuid {
+                        log::warn!(
+                            "Collection UUID {uuid} explicitly specified, \
+                             when proposing create collection {} operation, \
+                             new random UUID will be generated instead",
+                            op.collection_name,
+                        );
+                    }
```
ffuugoo, line 129: We have to include UUID into Qdrant's `CreateCollection` structure, because the same type is used both in consensus (to be serialized to CBOR and written into WAL) and in REST API, so I've excluded UUID from OpenAPI spec and added this warning.

lib/storage/src/content_manager/collection_meta_ops.rs
```diff
@@ -171,6 +172,9 @@ pub struct CreateCollection {
     /// Strict-mode config.
     #[validate(nested)]
     pub strict_mode_config: Option<StrictModeConfig>,
+    #[serde(default)]
+    #[schemars(skip)]
+    pub uuid: Option<Uuid>,
```
ffuugoo, line 177: Excluded UUID from OpenAPI spec (see other comments).

lib/collection/src/collection/state_management.rs
```diff
@@ -77,6 +77,17 @@ impl Collection {
         {
             let mut config = self.collection_config.write().await;
 
+            if config.uuid != new_config.uuid {
```
timvisee, line 80: How about this edge case:

Imagine a user created a collection in the middle of a rolling upgrade to a Qdrant version including this change.

In that case I expect that only half of the peers have the UUID set.

If we'd then restore a consensus snapshot, it'll wipe the collection (None vs UUID).

Shall we be a bit more careful here and just compare Some's?

```suggestion
            if config.uuid.zip(new_config.uuid).map_or(false, |(a, b)| a != b) {
```

lib/collection/src/operations/conversions.rs
timvisee: What is the reason _here_ for not including it in gRPC. Just because that would expose it to end users?

lib/storage/src/dispatcher.rs
timvisee: Do we strictly have to regenerate here? Can't we just use what is provided, if we ever want to explicitly use this (as hidden API)?

lib/storage/src/content_manager/toc/collection_container.rs
```diff
@@ -134,12 +134,29 @@ impl TableOfContent {
 
             for (id, state) in &data.collections {
                 if let Some(collection) = collections.get(id) {
-                    if let Err(err) = collection.check_config_compatible(&state.config).await {
+                    let collection_uuid = collection.uuid().await;
+
+                    let recreate_collection = if collection_uuid != state.config.uuid {
```
timvisee, line 139: Probably the same comment as I have above. Ignore None's here?

lib/collection/src/collection/state_management.rs
ffuugoo: I think it might be too much of an edge case, but good thinking. @generall wdyt?

lib/collection/src/operations/conversions.rs
ffuugoo: Yeah, pretty much. Seems easier to simply omit it here.

lib/storage/src/dispatcher.rs
ffuugoo: Not really, I just felt it's safer this way, but sure, we can allow it.

timvisee (approved):
Looks good.

Note that I haven't practically tested it yet though.

generall:
In the current version I see UUID is displayed in collection info (though, OpenAPI is somehow not changed?)

Would it be possible to not render UUID and keep it internal only?
image

generall:
My attempt to fix https://github.com/qdrant/qdrant/pull/5452

PR #5380, opened 2024-11-06
Add logging for abort shard transfer and fix a small typo

KShivendu:
Should help with debugging

### All Submissions:

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -279,7 +279,7 @@ impl Collection {
     ) -> CollectionResult<()> {
         // TODO: Ensure cancel safety!
 
-        log::debug!("Starting abort shard transfer {transfer_key:?}");
+        log::debug!("Aborting shard transfer {transfer_key:?}");
 
         let _transfer_result = self
             .transfer_tasks


lib/collection/src/collection/shard_transfer.rs
```diff
@@ -279,6 +279,8 @@ impl Collection {
     ) -> CollectionResult<()> {
         // TODO: Ensure cancel safety!
 
+        log::debug!("Starting abort shard transfer {transfer_key:?}");
+
```
timvisee, line 283: Aborting is not a long running process, and so I'd prefer to not log a separate start/stop, but to only log the event once.

I think it's best to place the message right before the bigger if-block on line 308.

timvisee (changes requested):

lib/collection/src/collection/shard_transfer.rs
KShivendu: It's not just to track time but also if we get any error in between. 

In case of error, only start will be printed but not stop. So let's keep this. Wdyt?

ffuugoo (approved):

timvisee (approved):

generall (approved):

PR #5387, opened 2024-11-06
Manual rocksdb's WAL flush

coszio:
We do not depend on rocksdb's WAL for durability, so this PR disables its WAL to be flushed on every write. Instead, we flush it at the same time as the other manual flushing.

Note that these changes also make the test suite green on #5190 without https://github.com/qdrant/qdrant/pull/5190/commits/0e70a454961eb9433c368486122cd66083e61b99. Although, tbh, it also passes without triggering the manual WAL flush in the `flusher()` implementation 🤷 

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/common/rocksdb_wrapper.rs
+++ b/lib/segment/src/common/rocksdb_wrapper.rs
@@ -61,6 +61,7 @@ pub fn make_db_options() -> Options {
     options.create_missing_column_families(true);
     options.set_max_open_files(DB_MAX_OPEN_FILES as i32);
     options.set_compression_type(rocksdb::DBCompressionType::Lz4);
+    options.set_manual_wal_flush(true);
 
     // Qdrant relies on it's own WAL for durability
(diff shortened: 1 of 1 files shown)


generall:
rocksdb flush was enabled for a reason https://github.com/qdrant/qdrant/pull/4411

timvisee:
> rocksdb flush was enabled because of consistency problems #4411

Yes, but this doesn't fully disable it (like before). This only synchronizes it's WAL flushing with our flushes. That's different.

timvisee:
I'm not really sure what hidden side effects this might have. Maybe we can redo the tests we did in <https://github.com/qdrant/qdrant/pull/4411> with this change and see if data still remains consistent. That would be enough for me to approve this.

agourlay:
This PR can be closed for now but we should definitely keep in mind that trick :+1:

PR #5391, opened 2024-11-07
Simplify logic in /readyz for checking if we have just one peer

timvisee:
What we had in place seems overly complicated for what it tries to achieve, and may send a lot of extra unnecessary requests while catching up.

This simplifies the check.

### All Submissions:

### DIFF
--- a/src/common/health.rs
+++ b/src/common/health.rs
@@ -212,14 +212,15 @@ impl Task {
         }
 
         // Get *cluster* commit index
-        let this_peer_id = self.toc.this_peer_id;
-        let transport_channel_pool = &self.toc.get_channel_service().channel_pool;
-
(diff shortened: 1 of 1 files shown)


src/common/health.rs
```diff
@@ -163,10 +163,10 @@ impl Task {
             // If not:
             //
             // - Check if this is the only node in the cluster
-            let Some(_) = self.cluster_commit_index().await else {
+            if self.consensus_state.peer_count() <= 1 {
```
generall, line 166: Is that always true when we bootstrap the peer? Assuming  that before consensus is in sync the current peer doesn't know about others?

ffuugoo: https://github.com/qdrant/qdrant/blob/dev/src/common/health.rs#L139-L148

ffuugoo (approved):

PR #5395, opened 2024-11-07
Per collection hardware metrics

JojiiOfficial:
Aggregate hardware measurements in the corresponding `Collection` for all measured operations.

Design choices:
  - Utilizes a new wrapper proving by design that no measurements can be dropped without noticing.
  - `HwMeasurementAcc` now also has checks when dropping values in test+debug builds with zero overhead in release, thanks to being completely conditionally compiled.

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -10984,7 +10984,6 @@
         "type": "object",
         "required": [
           "config",
-          "hardware_usage",
           "id",
           "init_time_ms",
           "resharding",
@@ -11022,7 +11021,14 @@
             }
           },
           "hardware_usage": {
-            "$ref": "#/components/schemas/HardwareInfo"
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/HardwareInfo"
+              },
+              {
+                "nullable": true
+              }
+            ]
           }
         }
       },
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -10984,6 +10984,7 @@
         "type": "object",
         "required": [
           "config",
+          "hardware_usage",
           "id",
           "init_time_ms",
           "resharding",
@@ -11019,6 +11020,9 @@
             "items": {
               "$ref": "#/components/schemas/ReshardingInfo"
             }
+          },
+          "hardware_usage": {
+            "$ref": "#/components/schemas/HardwareInfo"
           }
         }
       },
@@ -11758,6 +11762,19 @@
           },
           "updates": {
             "$ref": "#/components/schemas/OperationDurationStatistics"
+          }
+        }
+      },
+      "HardwareInfo": {
+        "type": "object",
+        "required": [
+          "cpu"
+        ],
+        "properties": {
+          "cpu": {
+            "type": "integer",
+            "format": "uint",
+            "minimum": 0
           }
         }
       },
--- a/lib/collection/src/collection/collection_ops.rs
+++ b/lib/collection/src/collection/collection_ops.rs
@@ -341,7 +341,11 @@ impl Collection {
                     .copied()
                     .unwrap_or(ReplicaState::Dead);
                 let count_result = replica_set
-                    .count_local(count_request.clone(), None, HwMeasurementAcc::new())
+                    .count_local(
+                        count_request.clone(),
+                        None,
+                        HwMeasurementAcc::new_unchecked(),
+                    )
                     .await
                     .unwrap_or_default();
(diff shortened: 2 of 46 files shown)


lib/common/common/src/counter/hardware_accumulator.rs
```diff
@@ -1,23 +1,42 @@
...
+    #[cfg(any(debug_assertions, test))]
+    applied: Arc<AtomicBool>,
 }
 
 impl HwMeasurementAcc {
     pub fn new() -> Self {
         Self::default()
     }
 
+    pub fn new_unchecked() -> Self {
```
timvisee, line 25: I see that we use this one a lot in this PR. Why can't we use the `new` variant instead for any of them?

lib/common/common/src/counter/hardware_accumulator.rs
```diff
@@ -26,24 +45,125 @@ impl HwMeasurementAcc {
...
+
     /// Consumes and accumulates the values from `hw_counter_cell` into the accumulator.
     pub fn merge_from_cell(&self, hw_counter_cell: impl Into<HardwareCounterCell>) {
         let HardwareCounterCell { ref cpu_counter } = hw_counter_cell.into();
 
         self.cpu_counter
             .fetch_add(cpu_counter.take(), Ordering::Relaxed);
     }
+
+    pub fn has_values(&self) -> bool {
```
timvisee, line 81: Nitpick: how about `is_zero()` (which would be the inverse)?

docs/redoc/master/openapi.json
```diff
@@ -10984,6 +10984,7 @@
         "type": "object",
         "required": [
           "config",
+          "hardware_usage",
```
timvisee, line 10987: Shouldn't this be optional?

lib/common/common/src/counter/hardware_accumulator.rs
JojiiOfficial: In the Drop-function of this type we have a check that helps us not to accidentally lose measurements in debug mode and tests.
This function makes it clear that when creating an instance of `HwMeasurementAcc`, we don't want it to have this check, which makes writing tests or hardware unrelated code easier to work with.

docs/redoc/master/openapi.json
JojiiOfficial: This is the telemetry endpoint and as far as I understood, we wanted it to be able to turn off for each API that gets measured (like search or matrix), but not telemetry.
I could be wrong here though.

Maybe @generall can give his opinion, whether we want it always to be in telemetry endpoint, or only in a certain condition.

generall: It should be option at least for backward compatibility of the clients. If new client reads old telemetry output it should not fail

lib/collection/src/collection/collection_ops.rs
```diff
@@ -341,7 +341,11 @@ impl Collection {
                     .copied()
                     .unwrap_or(ReplicaState::Dead);
                 let count_result = replica_set
-                    .count_local(count_request.clone(), None, HwMeasurementAcc::new())
+                    .count_local(
+                        count_request.clone(),
+                        None,
+                        HwMeasurementAcc::new_unchecked(),
```
generall, line 347: This needs a comment why `unchecked` is acceptable.

generall: Since this is not a test and not a bench

lib/collection/src/collection/mod.rs
```diff
@@ -176,6 +178,7 @@ impl Collection {
             update_runtime: update_runtime.unwrap_or_else(Handle::current),
             search_runtime: search_runtime.unwrap_or_else(Handle::current),
             optimizer_cpu_budget,
+            hardware_usage: HwMeasurementAcc::new_unchecked(),
```
generall, line 181: Why `unchecked` is acceptable?

lib/collection/src/collection/mod.rs
```diff
@@ -285,6 +288,7 @@ impl Collection {
...
```
generall, line 291: Why `unchecked` is acceptable?

lib/collection/src/shards/transfer/resharding_stream_records.rs
```diff
@@ -70,7 +70,7 @@ pub(crate) async fn transfer_resharding_stream_records(
                     exact: true,
                 }),
                 None,
-                HwMeasurementAcc::new(),
+                HwMeasurementAcc::new_unchecked(),
```
generall, line 73: needs a comment

JojiiOfficial:
Closing in favor of #5453

PR #5396, opened 2024-11-07
Make `default_on_disk_payload` consistent with config file

coszio:
We currently have our code to interpret a missing `on_disk_payload` as `false`, but it does not usually get exercised because the default config file for storage sets it to true:

https://github.com/qdrant/qdrant/blob/dev/config/config.yaml#L36

This PR:
- makes the default `on_disk_payload` to `true`, to match the default config value
- changes the default `PayloadStorageType` to `OnDisk`
- removes duplicate `default_on_disk_payload` implementation


agourlay (approved):
Thanks for tracking this down :+1: 

I believe this change is fine, we should not promote the current naive implementation of in memory payload.

generall (approved):

PR #5398, opened 2024-11-07
Backward compatibility for mmap payload storage

coszio:
Adds the integration necessary to be able to load and use a segment with mmap storage, but never actually creates it. 
The intention is so that we can release a patch version which can handle a rollback from a newer minor version, e.g. to let `1.13.0` able to be rolled back into `1.12.3`, if the new storage turns out to be buggy.
 
Also centralizes the `on_disk_payload` translation into `PayloadStorageType` so that it becomes easier to update in the future.

Just for reference, [this](https://github.com/qdrant/qdrant/blob/d5b7f22748b2784956fd6df4ba70394c05b3649c/lib/segment/src/segment_constructor/segment_constructor_base.rs#L690-L692) is where the segment config is read. This PR makes it able to handle the `PayloadStorageType::Mmap` variant.

**Note**: Thanks to #5190 (@agourlay) from which I basically git-checkout'ed the integration part

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,19 +4,19 @@ version = 3
 
 [[package]]
 name = "actix-codec"
-version = "0.5.0"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "57a7559404a7f3573127aab53c08ce37a6c6a315c374a31070f3c91cd1b4a7fe"
+checksum = "5f7b0a21988c1bf877cf4759ef5ddaac04c1c9fe808c9142ecb78ba97d97a28a"
 dependencies = [
- "bitflags 1.3.2",
+ "bitflags 2.6.0",
  "bytes",
  "futures-core",
  "futures-sink",
- "log",
  "memchr",
  "pin-project-lite",
  "tokio",
  "tokio-util",
+ "tracing",
 ]
 
 [[package]]
@@ -44,7 +44,7 @@ dependencies = [
  "actix-service",
  "actix-utils",
  "actix-web",
- "bitflags 2.4.1",
+ "bitflags 2.6.0",
  "bytes",
  "derive_more",
  "futures-core",
@@ -59,9 +59,9 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.7.0"
+version = "3.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4eb9843d84c775696c37d9a418bbb01b932629d01870722c0f13eb3f95e2536d"
+checksum = "d48f96fc3003717aeb9856ca3d02a8c7de502667ad76eeacd830b48d2e91fac4"
 dependencies = [
  "actix-codec",
  "actix-rt",
@@ -69,8 +69,8 @@ dependencies = [
  "actix-tls",
  "actix-utils",
  "ahash",
- "base64 0.22.0",
- "bitflags 2.4.1",
+ "base64 0.22.1",
+ "bitflags 2.6.0",
  "brotli",
  "bytes",
  "bytestring",
@@ -79,7 +79,7 @@ dependencies = [
  "flate2",
  "futures-core",
  "h2 0.3.26",
- "http 0.2.9",
+ "http 0.2.12",
  "httparse",
  "httpdate",
  "itoa",
@@ -99,12 +99,12 @@ dependencies = [
 
 [[package]]
 name = "actix-macros"
-version = "0.2.3"
+version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "465a6172cf69b960917811022d8f29bc0b7fa1398bc4f78b3c466673db1213b6"
+checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 1.0.107",
+ "syn 2.0.87",
 ]
 
 [[package]]
@@ -153,7 +153,7 @@ checksum = "13d324164c51f63867b57e73ba5936ea151b8a41a1d23d1031eeb9f70d0236f8"
 dependencies = [
  "bytestring",
  "cfg-if",
- "http 0.2.9",
+ "http 0.2.12",
  "regex",
  "regex-lite",
  "serde",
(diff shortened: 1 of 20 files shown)


Cargo.lock
```diff
@@ -4,19 +4,19 @@ version = 3
 
 [[package]]
 name = "actix-codec"
-version = "0.5.0"
```
agourlay, line 7: there are a lot of unrelated dependency changes here, did you run cargo update on the all repo?

lib/segment/src/payload_storage/mmap_payload_storage.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    fn wipe(&mut self) -> OperationResult<()> {
+        self.storage.write().wipe();
+        Ok(())
+    }
+
+    fn flusher(&self) -> Flusher {
+        let storage = self.storage.clone();
+        Box::new(move || {
+            storage.write().flush().unwrap(); // TODO: error conversion
```
agourlay, line 144: I think we need to address this TODO

lib/segment/src/payload_storage/mmap_payload_storage.rs
```diff
@@ -0,0 +1,172 @@
...
+    fn iter<F>(&self, mut callback: F) -> OperationResult<()>
+    where
+        F: FnMut(PointOffsetType, &Payload) -> OperationResult<bool>,
+    {
+        self.storage.read().iter(|point_id, payload| {
+            match callback(point_id, payload) {
+                Ok(true) => Ok(true),
+                Ok(false) => Ok(false),
+                Err(e) => {
+                    // TODO return proper error
```
agourlay, line 158: I think we need to address this TODO

agourlay:
Thanks for pushing the integration forward, I added a few comments that I try to handle myself.

lib/common/common/src/cpu.rs
```diff
@@ -208,7 +208,7 @@ pub enum ThreadPriorityError {
     #[error("Failed to set thread priority: {0:?}")]
     SetThreadPriority(thread_priority::Error),
     #[error("Failed to parse thread priority value: {0}")]
-    ParseNice(&'static str),
+    ParseNice(String),
```
timvisee, line 211: Why is this necessary? Due to a dependency bump?

lib/segment/src/payload_storage/mmap_payload_storage.rs
```diff
@@ -0,0 +1,172 @@
...
+            Ok(())
+        })
+    }
+
+    fn iter<F>(&self, mut callback: F) -> OperationResult<()>
+    where
+        F: FnMut(PointOffsetType, &Payload) -> OperationResult<bool>,
+    {
+        self.storage.read().iter(|point_id, payload| {
+            match callback(point_id, payload) {
```
timvisee, line 154: How about `callback(...).map_err(|e| std::io::Error::new(...))` here?

Cargo.lock
coszio: I did 😬

lib/common/common/src/cpu.rs
coszio: I assume it was a dependency bump, yes. But let's see if we can avoid it by not updating the whole project

lib/segment/Cargo.toml
```diff
@@ -88,6 +88,8 @@ zerocopy = { workspace = true }
 sysinfo = "0.32"
 charabia = { version = "0.9.1", default-features = false, features = ["greek", "hebrew", "thai"] }
 
+blob_store = { git = "https://github.com/qdrant/mmap-payload-storage" }
```
agourlay, line 91: we just should lock the revision like we do for the `wal` crate

lib/segment/src/payload_storage/mmap_payload_storage.rs
```diff
@@ -0,0 +1,172 @@
...
+            Ok(Self { storage })
+        } else {
+            Err(OperationError::service_error(
+                "Failed to open mmap payload storage",
+            ))
+        }
+    }
+
+    fn new(path: PathBuf) -> Self {
+        let storage = BlobStore::new(path, StorageOptions::default()).unwrap();
```
agourlay, line 59: this unwrap should go :)

lib/segment/src/payload_storage/mmap_payload_storage.rs
```diff
@@ -0,0 +1,172 @@
...
+            // create folder if it does not exist
+            std::fs::create_dir_all(&path).map_err(|_| {
+                OperationError::service_error("Failed to create mmap payload storage directory")
+            })?;
+            Ok(Self::new(path))
+        }
+    }
+
+    fn open(path: PathBuf) -> OperationResult<Self> {
+        if let Some(storage) = BlobStore::open(path) {
```
agourlay, line 48: we should have an error here explaining exactly why it can't opened

coszio: https://github.com/qdrant/qdrant/pull/5398/commits/8f697c26c0e14aa7c417c58c1bcee5e260f11431

lib/segment/src/payload_storage/mmap_payload_storage.rs
coszio: https://github.com/qdrant/qdrant/pull/5398/commits/8f697c26c0e14aa7c417c58c1bcee5e260f11431

lib/segment/Cargo.toml
coszio: https://github.com/qdrant/qdrant/pull/5398/commits/8f697c26c0e14aa7c417c58c1bcee5e260f11431

lib/segment/src/payload_storage/mmap_payload_storage.rs
coszio: Yeah, better

lib/segment/src/payload_storage/tests.rs
```diff
@@ -0,0 +1,106 @@
+use serde_json::json;
+
+use super::mmap_payload_storage::MmapPayloadStorage;
+use super::on_disk_payload_storage::OnDiskPayloadStorage;
+use super::simple_payload_storage::SimplePayloadStorage;
+use super::PayloadStorage;
+use crate::common::rocksdb_wrapper::open_db;
+use crate::types::Payload;
+
+fn test_trait_impl<S: PayloadStorage>(mut storage: S) {
```
agourlay, line 10: WDYT about adding `open` to the `PayloadStorage` interface?

This way we could test persistence here by dropping the storage and  then reopening it to check that the payloads are still present.

agourlay:
Looking very good, one last ask to improve testing then it is good to go for me :+1:

agourlay (approved):

timvisee (approved):

PR #5403, opened 2024-11-08
update Image.image type to Value instead of string

joein:
Just str type is not enough for local inference, since it does not cover all the types we'd like to support in local inference. This PR changes Image.image type to Value in order to support any value.
The decision to make it any was taken due to the desire to support objects which are not supported in openapi schema (e.g. `PIL.Image`

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/grpc/docs.md
+++ b/docs/grpc/docs.md
@@ -2813,7 +2813,7 @@ Additionally, the first and last points of each GeoLineString must be the same.
 
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
-| image | [string](#string) |  | Image data, either base64 encoded or URL |
+| image | [Value](#qdrant-Value) |  | Image data, either base64 encoded or URL |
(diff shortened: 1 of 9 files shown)


timvisee (commented):
Am I correct that we do not deem an URL or base64 string to be enough here?

I imagine sending a byte array with image data, for example. But in that case we'd need an extra field to specify the data type, as otherwise it'd be very arbitrary.

joein:
> Am I correct that we do not deem an URL or base64 string to be enough here?
> 
> I imagine sending a byte array with image data, for example. But in that case we'd need an extra field to specify the data type, as otherwise it'd be very arbitrary.

At the moment we support only base64 and url in cloud inference, I think that byte support might be added later once it is required
This change is mostly required for local inference, when we want to utilise the same structure, but we can't do it with string type, since we want to be able to use some python specific objects like `PIL.Image`

generall (approved):

PR #5405, opened 2024-11-08
Make /readyz catch up to latest consensus commit

timvisee:
Make `/readyz` a more strict and let it catch up to the latest current consensus commit, instead of just the latest commit at the time the peer started.

In other words, it now constantly polls the latest commit and tries to catch up to that, instead of just trying to catch up to the latest commit we saw at the time of starting the peer.

This is important because if a replica is marked as dead and is then recovered, the active state of that replica might blink on the restarted peer while it is already accepting updates. That could result in missing updates.

Some potential negative side effects of this change are:
- it might take longer to be ready
- if there is a constant stream of new consensus operations we may never catch up
- we have more calls across the cluster to synchronize /readyz

But, ensuring data consistency is more important than the above three points.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/src/common/health.rs
+++ b/src/common/health.rs
@@ -154,32 +154,32 @@ impl Task {
         self.check_ready_signal.notify_one();
 
         // Get *cluster* commit index, or check if this is the only node in the cluster
-        let Some(cluster_commit_index) = self.cluster_commit_index().await else {
+        let Some(mut cluster_commit_index) = self.cluster_commit_index().await else {
             self.set_ready();
             return;
         };
 
-        // Check if *local* commit index >= *cluster* commit index...
-        while self.commit_index() < cluster_commit_index {
-            // Wait for `/readyz` signal
-            self.check_ready_signal.notified().await;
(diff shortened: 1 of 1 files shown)


src/common/health.rs
```diff
@@ -153,33 +153,33 @@ impl Task {
         // This allows to check the happy path without waiting for the first call.
         self.check_ready_signal.notify_one();
 
-        // Get *cluster* commit index, or check if this is the only node in the cluster
-        let Some(cluster_commit_index) = self.cluster_commit_index().await else {
+        // Get estimate of current cluster commit so we can wait for it
+        let Some(mut cluster_commit_index) = self.cluster_commit_index(true).await else {
```
timvisee, line 157: Minor change: the first consensus commit we wait on is just requested from a single node, which may not necessarily be the latest. It is fine because we'll always do another set of requests to all nodes later.

In an ideal situation this reduces requests to other nodes by (50% - 1).

generall (approved):

PR #5425, opened 2024-11-12
Optimize matrix API result rendering

agourlay:
Decrease memory allocations when building large matrix API results.

### DIFF
--- a/lib/collection/src/collection/distance_matrix.rs
+++ b/lib/collection/src/collection/distance_matrix.rs
@@ -96,7 +96,8 @@ impl From<CollectionSearchMatrixResponse> for SearchMatrixPairsResponse {
             nearests,
         } = response;
 
-        let mut pairs = Vec::with_capacity(sample_ids.len() * nearests.len());
+        let pairs_len = nearests.iter().map(|n| n.len()).sum();
+        let mut pairs = Vec::with_capacity(pairs_len);
 
         for (a, scored_points) in sample_ids.into_iter().zip(nearests.into_iter()) {
             for scored_point in scored_points {
--- a/lib/collection/src/collection/distance_matrix.rs
+++ b/lib/collection/src/collection/distance_matrix.rs
@@ -77,8 +77,8 @@ impl From<CollectionSearchMatrixResponse> for SearchMatrixOffsetsResponse {
             }
         }
         let scores = nearests
-            .iter()
-            .flat_map(|row| row.iter().map(|p| p.score))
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection/distance_matrix.rs
```diff
@@ -96,7 +96,7 @@ impl From<CollectionSearchMatrixResponse> for SearchMatrixPairsResponse {
             nearests,
         } = response;
 
-        let mut pairs = vec![];
+        let mut pairs = Vec::with_capacity(sample_ids.len() * nearests.len());
```
timvisee, line 99: Shouldn't this be `nearest.iter().map(|n| n.len()).sum()` instead?

Nearest is a list of lists, and we add all the items for the sublists.

Or is `sample_ids.len() * nearests.len()` magically the same because it is a square?

lib/collection/src/collection/distance_matrix.rs
```diff
@@ -245,17 +245,15 @@ impl Collection {
...
+            if let Some(sample_pos) = scores.iter().position(|p| p.id == *sample_id) {
+                scores.remove(sample_pos);
+            } else {
+                // if not found pop lowest score
+                if scores.len() == limit_per_sample + 1 {
+                    // if we have enough results, remove the last one
+                    scores.pop();
+                }
             }
-            *scores = filtered;
```
timvisee, line 258: The new behavior is different, is that correct?

agourlay: How is it different?

timvisee: The new implementation pops an item at a specific position, the previous implementation did not.

timvisee: I see now, the previous implementation did this as well by reconstructing a filtered list.

Your new approach is better, thanks!

lib/collection/src/collection/distance_matrix.rs
agourlay: In the nominal case it should be a square but I think it is better to go with the most precise sizing, thanks :+1:

timvisee: Yeah, though, if it's almost correct we may _just_ have an extra reallocation. And if it should be a square we may benefit from a debug assertion, so we can adjust it once we have a case that isn't square.

timvisee (approved):

PR #5428, opened 2024-11-12
Leverage ahash in search results

agourlay:
On large result sets, the SIP hasher shows up in profiling when combining search results from different segments & shards.

This PR replaces the default hasher with `Ahash` which is more efficient on smaller input like ids.
After the change, hashing does not appear in profiling anymore.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1167,6 +1167,7 @@ dependencies = [
  "actix-files",
  "actix-web",
  "actix-web-validator",
+ "ahash",
  "api",
  "approx",
  "arc-swap",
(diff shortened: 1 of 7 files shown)


JojiiOfficial (approved):

timvisee (approved):
Maybe there are more places where we might benefit from this change.

Did you only change it here because it did show up during profiling?

That'd be totally valid by the way, I'm just curious. :smiley:

agourlay:
> Maybe there are more places where we might benefit from this change.

I also believe this, but as you guessed, I only changed the ones I could observe during profiling to keep the PR laser focused :+1:

PR #5430, opened 2024-11-12
Optimize merge shards results

agourlay:
Avoid allocating intermediary results when merging results from shards.

I do not have clear performance results but the extra `collect` are gone from the profile.

### DIFF
--- a/lib/collection/src/collection/search.rs
+++ b/lib/collection/src/collection/search.rs
@@ -260,7 +260,7 @@ impl Collection {
         let collection_params = self.collection_config.read().await.params.clone();
 
         // Merge results from shards in order and deduplicate based on point ID
-        let mut top_results: Vec<Vec<ScoredPoint>> = Vec::with_capacity(batch_size);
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection/search.rs
timvisee, line 1: Aren't both exactly the same in terms of allocations?

Before:
- `top_results` is correctly sized, but has no entries yet
- `.collect()` creates sub-vec with one allocation

After:
- `top_results`: is correctly sized, and has empty vec's
- `.extend()` allocates sub-vec with one allocation

timvisee: > we would need to set the capacity of the inner vectors for a real impact.

I think that's already handled.

A `vec![]` does no heap allocation at all and will only allocate on first use.

`.collect()` correctly guesses the length and will allocate once, same goes for `.extend()`.

What could improve it is to not use sub vectors and allocate the whole thing in one contiguous slice. That would remove all the allocations needed in inner vectors. But I'm not sure if that's doable here.

PR #5435, opened 2024-11-12
Add WriteZerosExt to write zeros

xzfc:
Add a small helper extension trait to write zeros into `std::io::Writer`.

### DIFF
--- a/lib/common/common/src/mmap_hashmap.rs
+++ b/lib/common/common/src/mmap_hashmap.rs
@@ -16,6 +16,8 @@ use rand::rngs::StdRng;
 use rand::Rng as _;
 use zerocopy::{AsBytes, FromBytes, FromZeroes};
 
+use crate::zeros::WriteZerosExt as _;
+
 type ValuesLen = u32;
 
 /// On-disk hash map backed by a memory-mapped file.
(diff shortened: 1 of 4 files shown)


lib/common/common/src/mmap_hashmap.rs
```diff
@@ -80,7 +80,7 @@ impl<K: Key + ?Sized, V: Sized + AsBytes + FromBytes> MmapHashMap<K, V> {
         file_size += phf.write_bytes();
 
         // 3. Padding
-        let padding_len = file_size.next_multiple_of(PADDING_SIZE) - file_size;
+        let padding_len = file_size.next_multiple_of(4096) - file_size;
```
generall, line 83: why const?

lib/common/common/src/zeros.rs
```diff
@@ -0,0 +1,17 @@
+use std::io::{Result, Write};
+
+static ZEROS: [u8; 8096] = [0u8; 8096];
```
timvisee, line 3: Out of curiosity, why 8k and not 4k?

Is it because of this IO constant? https://github.com/rust-lang/rust/blob/ec239b888f3945a6ce043db639f1741d14f08874/library/std/src/sys_common/io.rs#L3

timvisee (approved):

lib/common/common/src/zeros.rs
xzfc: It's a coincidence, tho it's a nice observation.

My considerations:
- I've increased it a bit for possible future use. (right now we don't need more than 4k).
- It's a static array, its size contributes to the resulting on-disk binary size as it goes to the [`.data` section](https://en.wikipedia.org/wiki/Data_segment). So it' shouldn't be too big.
- A single module with a single 8k buffer is not more wasteful than a bunch of 4k buffers in every module that wants to write zeros.

PR #5437, opened 2024-11-12
use more stable computation of vectors_size_bytes + refactoring

generall:
Current calculation of the `vectors_size_bytes` have some problems:

- It accumulates same vectors multiple times, if proxy segment is used
- It suffers from vectors not explicitly marked as deleted, even if the point is deleted

This PR fixes those shortcomings by changing the way we compute vector size:

1. We compute average size of all vectors
2. We estimate segment size by multiplying number of **points** to the average
3. We do not take temp segments into account

This PR also contains renaming for `size_in_bytes` -> `size_of_available_vectors_in_bytes` so the name is more explicit

+ bonus, PR makes sparse vector storage compute only size of available points.

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -11181,6 +11181,7 @@
             "minimum": 0
           },
           "vectors_size_bytes": {
+            "description": "An ESTIMATION of effective amount of bytes used for vectors Do NOT rely on this number unless you know what you are doing",
             "type": "integer",
             "format": "uint",
             "minimum": 0
--- a/lib/segment/src/vector_storage/simple_sparse_vector_storage.rs
+++ b/lib/segment/src/vector_storage/simple_sparse_vector_storage.rs
@@ -7,6 +7,7 @@ use common::types::PointOffsetType;
 use parking_lot::RwLock;
 use rocksdb::DB;
 use sparse::common::sparse_vector::SparseVector;
+use sparse::common::types::DimWeight;
 
 use super::SparseVectorStorage;
 use crate::common::operation_error::{check_process_stopped, OperationError, OperationResult};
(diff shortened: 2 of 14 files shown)


KShivendu (commented):
> do not take temp segments into account

But what if new point IDs are inserted into temp segments during optimizations? Do you intend to make it behave such that transfers will happen after optimizations if required in such case?

docs/redoc/master/openapi.json
```diff
@@ -11181,6 +11181,7 @@
             "minimum": 0
           },
           "vectors_size_bytes": {
+            "description": "An ESTIMATION of effective amount of bytes used for vectors Do NOT rely on this number unless you know what you are doing",
```
timvisee, line 11184: > ESTIMATION

:rofl:

lib/segment/src/vector_storage/simple_sparse_vector_storage.rs
```diff
@@ -162,8 +163,14 @@ impl VectorStorage for SimpleSparseVectorStorage {
...
 
-    fn size_in_bytes(&self) -> usize {
-        self.total_sparse_size * (std::mem::size_of::<f32>() + std::mem::size_of::<u32>())
+    fn size_of_available_vectors_in_bytes(&self) -> usize {
+        if self.total_vector_count == 0 {
+            return 0;
+        }
+        let available_fraction =
+            (self.total_vector_count - self.deleted_count) as f32 / self.total_vector_count as f32;
+        let available_size = (self.total_sparse_size as f32 * available_fraction) as usize;
```
timvisee, line 172: How about f64 here as this will only accurately represent counts up to 16.7 million. Or do we not care from that point onwards?

timvisee (approved):

lib/segment/src/vector_storage/simple_sparse_vector_storage.rs
timvisee: On second thought, I don't expect any real problems here.

generall:
> But what if new point IDs are inserted into temp segments during optimizations? Do you intend to make it behave such that transfers will happen after optimizations if required in such case?

Points inserted into temp segment are ignored until the indexation is complete. As well as if you delete points, it won't be reflected in this value during the optimization. Effectively, it is the volume of vectors, accessible with `indexed_only` param.

PR #5439, opened 2024-11-13
Fix the file name too long error during collection deletion

gulshan-rs:
Fixes https://github.com/qdrant/qdrant/issues/4568

Based on suggestion https://github.com/qdrant/qdrant/pull/4788#issuecomment-2468444870, we just not use the collection-name in the folder struct of the collections to be deleted; which prevents us in getting into risks of too-long-file-name error.

### DIFF
--- a/lib/storage/src/content_manager/toc/collection_meta_ops.rs
+++ b/lib/storage/src/content_manager/toc/collection_meta_ops.rs
@@ -8,7 +8,7 @@ use collection::shards::collection_shard_distribution::CollectionShardDistributi
 use collection::shards::replica_set::ReplicaState;
 use collection::shards::transfer::ShardTransfer;
 use collection::shards::{transfer, CollectionId};
-use uuid::Uuid;
+use tempfile::Builder;
 
 use super::TableOfContent;
 use crate::content_manager::collection_meta_ops::*;
@@ -199,13 +199,20 @@ impl TableOfContent {
             drop(removed);
 
             // Move collection to ".deleted" folder to prevent accidental reuse
-            let uuid = Uuid::new_v4().to_string();
+            // the original collection path will be moved atomically within this
+            // directory.
             let removed_collections_path =
                 Path::new(&self.storage_config.storage_path).join(".deleted");
             tokio::fs::create_dir_all(&removed_collections_path).await?;
-            let deleted_path = removed_collections_path
-                .join(collection_name)
-                .with_extension(uuid);
+
+            let deleted_path = Builder::new()
+                // Limit the file name to be on a lower side to avoid running into too-long
+                // file names.
+                // Even if the chosen randomness factor poses chances of collision, the library
+                // prevents creation of duplicate files within the chosen directory.
+                .rand_bytes(8)
+                .tempdir_in(removed_collections_path)?;
+
             tokio::fs::rename(path, &deleted_path).await?;
 
             // Solve all issues related to this collection
--- a/lib/storage/src/content_manager/toc/collection_meta_ops.rs
+++ b/lib/storage/src/content_manager/toc/collection_meta_ops.rs
@@ -227,7 +227,7 @@ impl TableOfContent {
                 if let Err(error) = tokio::fs::remove_dir_all(&deleted_path).await {
                     log::error!(
                         "Can't delete collection {} from disk. Error: {}",
-                        deleted_path.display(),
+                        deleted_path.as_ref().display(),
                         error
                     );
                 }
--- a/lib/storage/src/content_manager/toc/collection_meta_ops.rs
+++ b/lib/storage/src/content_manager/toc/collection_meta_ops.rs
@@ -211,6 +211,7 @@ impl TableOfContent {
                 // Even if the chosen randomness factor poses chances of collision, the library
                 // prevents creation of duplicate files within the chosen directory.
                 .rand_bytes(8)
+                .prefix("")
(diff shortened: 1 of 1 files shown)


lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -199,13 +200,23 @@ impl TableOfContent {
...
+            // Generate a random alphanumeric string to append to the deleted storage directory.
+            let random_str: String = rand::thread_rng()
+                .sample_iter(&Alphanumeric)
+                // Limit the generated string to 8 characters to avoid issues
+                // with file name length restrictions. This size provides a
+                // very low probability of collision (1 in 62^8).
+                .take(8)
+                .map(char::from)
+                .collect();
+            let deleted_path = removed_collections_path.join(random_str);
```
timvisee, line 218: Though the chance for collision is low, I'd still prefer to totally eliminate it from happening.

Maybe we can use [`TempDir`](https://docs.rs/tempfile/latest/tempfile/) here instead, we already use that crate across the codebase.

It'll generate a random directory name for us and ensures uniqueness. As a bonus it'll also help us clean up the directory if we some how failed to do so in the tokio spawn loop below.

I'd suggest to use the same `.deleted` directory like now so you might have to use [`tempdir_in`](https://docs.rs/tempfile/latest/tempfile/fn.tempdir_in.html).

What do you think?

timvisee (changes requested):
Thanks for picking this up!

Please see my review comment below as I think that might improve it a bit.

lib/storage/src/content_manager/toc/collection_meta_ops.rs
gulshan-rs: makes sense to me. both reasoning are strong; I will send a new revision to it. 

fyi: ```tempdir``` will produce random file names of 12 chars, which extend the collision factor to 1 in 62^12 (that's 3.23 quintillion)

timvisee: > tempdir will produce random file names of 12 chars, which extend the collision factor to 1 in 62^12 (that's 3.23 quintillion)

tempdir also checks for collision internally, pick a new random name if it already exists. So no concern there.

If we can reduce the random characters to 8 for tempdir that's fine by me. No strong opinions there.

gulshan-rs:
Ready for review.

lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -199,13 +199,25 @@ impl TableOfContent {
...
-                .join(collection_name)
-                .with_extension(uuid);
+
+            let temp_file_binding = Builder::new()
+                // Limit the file name to be on a lower side to avoid running into too-long
+                // file names.
+                // Even if the chosen randomness factor poses chances of collision, the library
+                // prevents creation of duplicate files within the chosen directory.
+                .rand_bytes(8)
+                .tempfile_in(removed_collections_path)?;
```
timvisee, line 214: I think you need tempdir instead:

```suggestion
                .tempdir_in(removed_collections_path)?;
```

As far as I understand the `tokio::fs::rename` will simply replace it if the directory is still empty.

gulshan-rs: > I think you need tempdir instead:

both ```tempdir``` and ```tempfile``` crates are similar.  In qdrant, we use tempfile, here's a [github search](https://github.com/search?q=repo%3Aqdrant%2Fqdrant+tempfile+path%3A*.toml&type=code); so i didn't feel like adding a new dependency.

timvisee: > oh i think you meant, using tempdir_in instead of tempfile_in using the tempfile crate, let me check on it.

Correct :smile: 

Just because it better matches with the end result, having a directory there (not a file).

gulshan-rs: You are right.  I misjudged that the ```path``` that is being renamed to be a file, its actually a directory.

timvisee: Thanks for the patch!

timvisee:
Final thing: could you point this PR to the dev branch instead?

You have to rebase your branch locally first and can do that with the following commands:
```
git fetch
git rebase --onto origin/dev 1a575c7500de258ee8aa82414ebcba344110571a^
git push --force-with-lease
```

timvisee:
I did a basic patch to remove the default `.tmp` prefix on the directories in https://github.com/qdrant/qdrant/pull/5439/commits/26cb9ef29ce88509d4f530d38215f27a15111b6a.

It now picks a directory like: `./storage/.deleted/9hRdT0Td`

timvisee (approved):
Thanks a lot for implementing this! I think it's a good fix for <https://github.com/qdrant/qdrant/issues/4568>. :raised_hands: 

I tested it locally and it behaves as expected.

gulshan-rs:
I actually wanted the ```tmp``` prefix, hinting it's a temp dir and might get auto-cleaned (too) -- but, not a huge deal.

timvisee:
> I actually wanted the `tmp` prefix, hinting it's a temp dir

Fair enough. I thought the `.deleted/` prefix would already give that away. No strong opinion on this :smile:

generall (approved):

PR #5440, opened 2024-11-13
Optimize mmap sequential access for large scoring

agourlay:
When scoring vectors from a plain segment, all vectors need to be retrieved.

However the underlying mmap storage uses `MADV_RANDOM` which is not working well with a sequential access pattern. 

This issue is particularly visible when using network mounted disk, where each read ends up hitting the disk.

This PR introduces a new way to retrieve vectors in a batch fashion from storage for this use case.

We can leverage the fact that a batch of vectors is contiguous to prefetch the memory on a second mmap file using `MADV_SEQUENTIAL`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -245,7 +245,7 @@ opt-level = 0
 # Inherits by default from release
 [profile.bench]
 lto = "off"
-#debug = true
+debug = true
 codegen-units = 256 # restore default value for faster compilation
 
 # Profile for performance testing, which is faster to build than release.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -245,7 +245,7 @@ opt-level = 0
 # Inherits by default from release
 [profile.bench]
 lto = "off"
-debug = true
+#debug = true
 codegen-units = 256 # restore default value for faster compilation
 
 # Profile for performance testing, which is faster to build than release.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -244,7 +244,7 @@ opt-level = 0
 
 # Inherits by default from release
 [profile.bench]
-lto = false
+lto = "off"
 debug = true
 codegen-units = 256 # restore default value for faster compilation
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -244,7 +244,7 @@ opt-level = 0
 
 # Inherits by default from release
 [profile.bench]
-lto = "off"
+lto = false
 debug = true
 codegen-units = 256 # restore default value for faster compilation
 
--- a/lib/segment/src/vector_storage/async_raw_scorer.rs
+++ b/lib/segment/src/vector_storage/async_raw_scorer.rs
@@ -203,6 +203,10 @@ where
         pq.into_vec()
     }
 
+    fn peek_top_all_old(&self, top: usize) -> Vec<ScoredPointOffset> {
+        self.peek_top_all(top)
+    }
+
     fn take_hardware_counter(&self) -> HardwareCounterCell {
         self.query_scorer.take_hardware_counter()
     }
--- a/lib/segment/src/vector_storage/async_raw_scorer.rs
+++ b/lib/segment/src/vector_storage/async_raw_scorer.rs
@@ -203,10 +203,6 @@ where
         pq.into_vec()
     }
 
-    fn peek_top_all_old(&self, top: usize) -> Vec<ScoredPointOffset> {
-        self.peek_top_all(top)
-    }
-
     fn take_hardware_counter(&self) -> HardwareCounterCell {
         self.query_scorer.take_hardware_counter()
     }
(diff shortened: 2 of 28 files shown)


Cargo.toml
```diff
@@ -244,8 +244,8 @@ opt-level = 0
 
 # Inherits by default from release
 [profile.bench]
-lto = false
-debug = true
+lto = "off"
```
agourlay, line 247: > false: Performs “thin local LTO” which performs “thin” LTO on the local crate only across its [codegen units](https://doc.rust-lang.org/cargo/reference/profiles.html#codegen-units).

https://doc.rust-lang.org/cargo/reference/profiles.html#lto

It is almost impossible for me to iterate otherwise due to compile times

lib/segment/src/vector_storage/async_raw_scorer.rs
```diff
@@ -203,6 +203,10 @@ where
         pq.into_vec()
     }
 
+    fn peek_top_all_old(&self, top: usize) -> Vec<ScoredPointOffset> {
```
agourlay, line 206: TODO remove before merge

lib/segment/src/vector_storage/raw_scorer.rs
```diff
@@ -87,6 +88,7 @@ pub trait RawScorer {
     ) -> Vec<ScoredPointOffset>;
 
     fn peek_top_all(&self, top: usize) -> Vec<ScoredPointOffset>;
+    fn peek_top_all_old(&self, top: usize) -> Vec<ScoredPointOffset>;
```
agourlay, line 91: TODO remove before merge

lib/segment/src/vector_storage/raw_scorer.rs
```diff
@@ -915,16 +916,22 @@ where
...
 
     fn peek_top_all(&self, top: usize) -> Vec<ScoredPointOffset> {
-        let scores = (0..self.point_deleted.len() as PointOffsetType)
+        let point_ids = (0..self.point_deleted.len() as PointOffsetType)
             .take_while(|_| !self.is_stopped.load(Ordering::Relaxed))
             .filter(|point_id| self.check_vector(*point_id))
-            .map(|point_id| {
-                let point_id = point_id as PointOffsetType;
-                ScoredPointOffset {
+            .chunks(64); // batch points to leverage storage sequential access
```
agourlay, line 922: Ideally the batch size should be a function of the vector size.

lib/segment/src/vector_storage/raw_scorer.rs
```diff
@@ -915,16 +916,22 @@ where
...
+
+        let scores = point_ids.into_iter().flat_map(|point_ids| {
+            let point_ids: Vec<PointOffsetType> = point_ids.collect();
```
generall, line 925: might be possible to avoid collect here?

agourlay: One constraint is checking that the ids are consecutive to pick the batching strategy. This requires a full traversal.

lib/segment/src/vector_storage/raw_scorer.rs
coszio: Even when batched in chunks of 64 vectors, the "must be consecutive requirement" could be relaxed to have some proportion of it missing and still be able to leverage the sequential reads. E.g. allow some small percentage of the IDs to not be consecutive. WDYT?

agourlay:
converting to draft because `MADV_WILLNEED` is not helping

I will apply a couple refactoring to at least ship the batching logic

lib/segment/src/vector_storage/query_scorer/mod.rs
```diff
@@ -58,3 +63,45 @@ fn score_multi<T: PrimitiveVectorElement, TMetric: Metric<T>>(
...
 }
+
+/// Check if ids are rather contiguous to enable further optimizations
+/// The threshold is 20% of sparsity
+pub fn check_ids_rather_contiguous(ids: &[PointOffsetType]) -> bool {
+    if ids.len() < 2 {
+        return false;
+    }
+    // check if sorted
+    if ids.windows(2).any(|w| w[0] >= w[1]) {
```
generall, line 74: Check for sorted is `O(N)`, but for the same price we can just find min & max and don't really care if it is strictly sorted

lib/segment/src/vector_storage/query_scorer/custom_query_scorer.rs
```diff
@@ -98,6 +98,16 @@ impl<
         self.score(stored)
     }
 
+    fn score_stored_batch(&self, ids: &[PointOffsetType]) -> Vec<ScoreType> {
+        if check_ids_rather_contiguous(ids) {
```
generall, line 102: I seems to me that this check should happen on the storage level, cause it should know better how vectors are stored exactly.

generall:
```
Current dev:
// 5k vectors 1536d
// seq full scan: 5+ seconds 30M cpu counters

This PR


// 5k vectors 1536d

// seq full scan: 0.2s 30M cpu counters
// rr hnsw ef 10: 0.97s 3.5M cpu countres

// seq hnsw ef 10: 0.4s 3.5M cpu countres


// 50k vectors 512d
// seq full scan: 0.74s 100M cpu counters
// rr hnsw ef 10: 0.61s 1.3M cpu countres
// seq hnsw ef 10: 0.81s 1.3M cpu comparision
```

generall (approved):

PR #5447, opened 2024-11-14
allow inference for Documents in points (grpc + rest)

n0x29a:
This PR allows inference for points queries for rest and grpc

### DIFF
--- a/lib/api/src/rest/validate.rs
+++ b/lib/api/src/rest/validate.rs
@@ -50,16 +50,7 @@ impl Validate for VectorInput {
             VectorInput::DenseVector(_dense) => Ok(()),
             VectorInput::SparseVector(sparse) => sparse.validate(),
             VectorInput::MultiDenseVector(multi) => validate_multi_vector(multi),
-            VectorInput::Document(_) => {
-                let mut errors = ValidationErrors::default();
-                let mut err = ValidationError::new("not_supported_inference");
-                err.add_param(
-                    Cow::from("message"),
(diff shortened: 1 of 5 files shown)


lib/api/src/rest/validate.rs
```diff
@@ -50,16 +50,7 @@ impl Validate for VectorInput {
...
-                let mut errors = ValidationErrors::default();
-                let mut err = ValidationError::new("not_supported_inference");
-                err.add_param(
-                    Cow::from("message"),
-                    &"Document inference is not implemented, please use vectors instead",
-                );
-                errors.add("text", err);
-                Err(errors)
-            }
+            VectorInput::Document(_) => Ok(()),
```
agourlay, line 53: Is there nothing to validate on the `Document` object?
Like non empty `text` and `model` maybe?

n0x29a: It should follow a pattern matching approach: if there is a combination of text and model, it is considered a Document. At this point, we assume the necessary fields are already present.

Checking the content isn’t straightforward, as it could be used for third-party services like Jina.ai. If the text is empty, the response will be an empty vector, which might be expected behavior rather than an error.

Currently, the inference service validates an empty model. If the model is missing, it responds with: Unsupported model "".

In summary, I think model + text validation could occur outside the inference service, with Qdrant simply proxying the request. 

Please let me know what you think and if it will be a better approach to add additional checks here.

generall (approved):

PR #5453, opened 2024-11-15
Per collection hardware measurements 

JojiiOfficial:
Supersedes #5395 

A bit different approach to accumulate hardware measurements per collection.
I ended up splitting it into multiple related commits instead, so reviewing by commit would probably be easier.

Changes over the other PR:
  - `HwMeasurementAcc` doesn't implement clone but can be 'duplicated' for multithreading/async by utilizing a new `HwMeasurementCollector` type which doesn't implement a drop-check and can be moved across threads since it's owned.
  - `HwMeasurementAcc` is now passed by reference.
  - Per collection measurements is stored in `Toc` instead of the collection itself.


src/tonic/api/points_api.rs
```diff
@@ -278,14 +292,20 @@ impl Points for PointsService {
     ) -> Result<Response<SearchResponse>, Status> {
         validate(request.get_ref())?;
         let access = extract_access(&mut request);
-        search(
+        let collection_name = request.get_ref().collection_name.clone();
+        let hw_metrics = HwMeasurementAcc::new();
```
generall, line 296: What if request fails(times out) before metric is reported?

lib/common/common/src/counter/hardware_accumulator.rs
```diff
@@ -47,3 +101,46 @@ impl Default for HwMeasurementAcc {
...
+///
+/// This collector automatically applies all measurements to the `HwMeasurementAcc` that was used when creating this collector on drop.
+/// To ensure all values get applied, before reading from a `HwMeasurementAcc`, this type must be dropped first.
+pub struct HwMeasurementCollector {
+    /// Shared reference of CPU usage with the given Parent.
+    cpu_counter: Arc<AtomicUsize>,
+
+    /// A new temporary measurement accumulator, which values get applied to the parents
+    /// ones on drop.
+    collector: HwMeasurementAcc,
```
generall, line 117: I like this idea, but now we have a circular dependency between structures. Both of `HwMeasurementCollector` and `HwMeasurementAcc` require each other. I would prefer to avoid that.

lib/common/common/src/counter/hardware_accumulator.rs
```diff
@@ -47,3 +101,46 @@ impl Default for HwMeasurementAcc {
...
+        let collector = HwMeasurementAcc::new();
+        Self {
+            cpu_counter,
+            collector,
+        }
+    }
+}
+
+impl Deref for HwMeasurementCollector {
+    type Target = HwMeasurementAcc;
```
generall, line 131: same as implicit conversion, it is quite impossible to find where exactly this deref is used.

generall (approved):
Here is a quick summary of my changes on top of existing PR:

- Remove collector logic and replace it with drain logic. 

This allows to address 2 problems: remove circular dependency in data structures and avoid usage of deref from one structure to another

- Remove some functions like `clear` which was only privately used by `discard` and make some other function private

- Restructure reporting of the usage to collection accumulator, instead of reporting at the end of operation we register drain immediately and report usage on drop. This prevents unregistered usage by timing out the request

- Implement special logic for internal requests - if they are not reported to API response, the usage is registered on the local machine. Needed for the same reason of accounting for timing-out requests, but now in distributed setup.

-  Implemented collection in REST api, removed `helpers::time_and_hardware_opt` as it creates problems with lifetime and overall doesn't look too convenient

PR #5457, opened 2024-11-17
Fixes for rescoring param usage and HNSW

generall:
This PR contains 2 improvement, which when combined, make usage of the colpali models a bit less miserable.

1. Propagate search parameters to re-scoring search.

Previously, if we were using re-scoring (query + prefetch), the re-scoring request was used with empty param, assuming that since we had fixed list of elements to re-score, we would always do a full-scan. 
This is, however, not true for ColPali vectors, where only ~20 vectors overwhelm  full-scan threshold and force us to use HNSW (which is not efficient at this stage)

Additionally, empty parameter forced us to ignore quantization, while sometimes it might be useful. Now params from the query are fully propagated.

2. Detection if the HNSW is disabled. 

Since ColPali vectors are enormous, building HNSW for them barely makes sense. Comparing 2 multi-vectors require (1030 ** 2) vector comparisons. Instead full-size vectors should be used only for re-scoring. Therefore, HNSW should be disabled. We, however, previously ignored disabled HNSW and continued to search in HNSW graph even if it is not built.

This PR explicitly checks if HNSW is enabled and falls back to full scan (with quantization) if it is.

Those changes might be useful not only for colpali models.

### DIFF
--- a/lib/collection/src/operations/universal_query/planned_query.rs
+++ b/lib/collection/src/operations/universal_query/planned_query.rs
@@ -1,7 +1,7 @@
 //! Types used within `LocalShard` to represent a planned `ShardQueryRequest`
 
 use common::types::ScoreType;
-use segment::types::{Filter, WithPayloadInterface, WithVector};
+use segment::types::{Filter, SearchParams, WithPayloadInterface, WithVector};
 
 use super::shard_query::{SampleInternal, ScoringQuery, ShardPrefetch, ShardQueryRequest};
 use crate::operations::types::{
@@ -44,6 +44,9 @@ pub struct RescoreParams {
 
     /// The payload to return
     pub with_payload: WithPayloadInterface,
+
+    /// Parameters for the rescore search request
+    pub params: Option<SearchParams>,
 }
 
 #[derive(Debug, PartialEq)]
@@ -138,6 +141,7 @@ impl PlannedQuery {
                         score_threshold,
                         with_vector,
                         with_payload,
+                        params,
                     }),
                 }
             }
@@ -281,6 +285,7 @@ fn recurse_prefetches(
                     score_threshold,
                     with_vector: with_vector.clone(),
                     with_payload: with_payload.clone(),
+                    params,
                 }),
             };
 
@@ -441,7 +446,10 @@ mod tests {
             score_threshold: None,
             limit: 10,
             offset: 0,
-            params: None,
+            params: Some(SearchParams {
+                exact: true,
+                ..Default::default()
+            }),
             with_vector: WithVector::Bool(true),
             with_payload: WithPayloadInterface::Bool(true),
         };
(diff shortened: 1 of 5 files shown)


lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -877,23 +912,20 @@ impl<TGraphLinks: GraphLinks> VectorIndex for HNSWIndex<TGraphLinks> {
...
-                if exact {
-                    let exact_params = params.map(|params| {
-                        let mut params = *params;
-                        params.quantization = Some(QuantizationSearchParams {
-                            ignore: true,
-                            rescore: Some(false),
-                            oversampling: None,
-                        }); // disable quantization for exact search
-                        params
+                if exact || is_hnsw_disabled {
```
timvisee, line 915: Do we also want this branch if `vector_storage.available_vector_count() < self.config.full_scan_threshold`, similar to above?

lib/collection/src/operations/universal_query/planned_query.rs
```diff
@@ -55,7 +58,7 @@ pub enum Source {
     ScrollsIdx(usize),
 
     /// A nested prefetch
-    Prefetch(MergePlan),
+    Prefetch(Box<MergePlan>),
```
timvisee, line 61: Interesting clippy warning!

```
 1  warning: large size difference between variants
   --> lib/collection/src/operations/universal_query/planned_query.rs:53:1
    |
 53 | / pub enum Source {
 54 | |     /// A reference offset into the main search batch
 55 | |     SearchesIdx(usize),
    | |     ------------------ the second-largest variant contains at least 8 bytes
 ...  |
 61 | |     Prefetch(MergePlan),
    | |     ------------------- the largest variant contains at least 240 bytes
 62 | | }
    | |_^ the entire enum is at least 240 bytes
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.htm
 l#large_enum_variant
    = note: `#[warn(clippy::large_enum_variant)]` on by default
 help: consider boxing the large fields to reduce the total size of the enum
    |
 61 |     Prefetch(Box<MergePlan>),
    |              ~~~~~~~~~~~~~~
```

Based on that I do agree with the change. Not to make `SearchesIdx` and `ScrollsIdx` very heavy.

timvisee (approved):

lib/collection/src/operations/universal_query/planned_query.rs
generall: My reasoning was that MergePlan only exists for queries with prefetch, so they will be heavy anyways.

lib/segment/src/index/hnsw_index/hnsw.rs
generall: This branch is a shortcut to avoid unneeded `estimate_cardinality` call. For unfiltered requests it is not needed at all, but here we can split in 2 cases

PR #5458, opened 2024-11-18
Fix large variant for fulltext index

agourlay:
I do not think we need to make an exception here for sizing the `FullTextIndex` enum.

```console
error: large size difference between variants
  --> lib/segment/src/index/field_index/full_text_index/text_index.rs:28:1
   |
28 | / pub enum FullTextIndex {
29 | |     Mutable(MutableFullTextIndex),
   | |     ----------------------------- the second-largest variant contains at least 0 bytes
30 | |     Immutable(ImmutableFullTextIndex),
31 | |     Mmap(MmapFullTextIndex),
   | |     ----------------------- the largest variant contains at least 352 bytes
32 | | }
   | |_^ the entire enum is at least 0 bytes
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
   = note: `-D clippy::large-enum-variant` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::large_enum_variant)]`
help: consider boxing the large fields to reduce the total size of the enum
   |
31 |     Mmap(Box<MmapFullTextIndex>),
   |          ~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/mmap_text_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/mmap_text_index.rs
@@ -147,6 +147,6 @@ impl FieldIndexBuilderTrait for FullTextMmapIndexBuilder {
             config,
         };
 
-        Ok(FullTextIndex::Mmap(mmap_index))
+        Ok(FullTextIndex::Mmap(Box::new(mmap_index)))
     }
 }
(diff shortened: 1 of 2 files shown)


timvisee (approved):
I'm a bit surprised this turns out to be so big.

`MmapInvertedIndex` is 312 bytes, partially because the `MmapHashMap` inside it is already 104 bytes :open_mouth:

PR #5460, opened 2024-11-18
Log debug message on operation rejection for missing force flag

timvisee:
An operation may be rejected on a replica in `Recovery` state if no `force` flag is provided.

This now shows a debug message when that happens.

For rejections caused by old clocks we already have a warning and automatic retry in place here: https://github.com/qdrant/qdrant/blob/d6f266a4ddbf3f57a3d308b6dce1fed83b867917/lib/collection/src/shards/replica_set/update.rs#L175-L178

I expect this debug message to be a bit noisy if a lot of replica state changes are happening. Let's see how it behaves and revert if needed. This will at least allow us to detect rejections on chaos testing for the time being.

### All Submissions:

### DIFF
--- a/lib/collection/src/operations/mod.rs
+++ b/lib/collection/src/operations/mod.rs
@@ -160,12 +160,12 @@ impl CollectionUpdateOperations {
         )
     }
 
-    pub fn point_ids(&self) -> Vec<PointIdType> {
+    pub fn point_ids(&self) -> Option<Vec<PointIdType>> {
         match self {
             Self::PointOperation(op) => op.point_ids(),
             Self::VectorOperation(op) => op.point_ids(),
             Self::PayloadOperation(op) => op.point_ids(),
-            Self::FieldIndexOperation(_) => Vec::new(),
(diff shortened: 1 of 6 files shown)


timvisee:
I've adjusted this a bit, making this less noisy at the cost of leaving out details.

If the operation specifies point IDs it'll just list the affected point IDs. If the operation does not specify point IDs, with delete by filter for example, it'll still print the full operation.

For example:

```
DEBUG: Operation affecting point IDs [NumId(5184), NumId(5185), NumId(5186)] rejected on this peer, force flag required in recovery state
DEBUG: Operation affecting point IDs [NumId(5199), NumId(5200), NumId(5201)] rejected on this peer, force flag required in recovery state
DEBUG: Operation affecting point IDs [NumId(5202), NumId(5203), NumId(5204)] rejected on this peer, force flag required in recovery state
```

It is a bit of a naive implementation because it's specialized to just this log message.

generall (approved):

PR #5462, opened 2024-11-18
RocksDB buffers use Ahash

agourlay:
The RocksDB's buffering is hashing ids using the SIP hasher.

sipping

This PR changes to Ahash which is better for small inputs.


timvisee (approved):
Again assuming this shows a measurable difference.

agourlay:
I don't see a clear difference in the results, closing for now.

PR #5473, opened 2024-11-18
Merge quantization integration tests

agourlay:
This PR merges the integration tests from the `quantization` crate into a single test binary in order to:
- decrease disk usage of compiled artifacts
- decrease compilation time

similar to https://github.com/qdrant/qdrant/pull/2033


generall (approved):

timvisee (approved):

agourlay:
For reference, the gain this time is around 350MB of binaries.

PR #5475, opened 2024-11-19
Fix: Timeout Precision in Remote Shard Queries

### LINKED ISSUE #5463: Timeout error on query if timeout=1 in a distributed deployment
## Current Behavior
When querying the Qdrant Recommendation API, if the collection is distributed (across multiple shards) and the timeout is set to 1 second, Qdrant may return a timeout error during request distribution, even if the request execution time is less than 1 second. Without a timeout specified, or with timeout=2, the request succeeds.

## Steps to Reproduce
1.	Create a Qdrant cluster with multiple nodes using the official Helm Chart.
2.	Create a collection with multiple shards.
3.	Execute a simple recommend query like the one below (with timeout=1):
```
POST /collections/<collection_name>/points/query?timeout=1
{
  "query": {
    "recommend": {
      "positive": [
        "c026bde1-8399-488d-954a-c777a42ab74a"
      ]
    }
  },
  "with_payload": false,
  "with_vector": false
}
```
4. Qdrant output the following error
```
{
  "error": "Service internal error: 2 of 2 read operations failed:\n  Timeout error: Deadline Exceeded: status: DeadlineExceeded, message: \"Timeout: Timeout error: Operation 'Search' timed out after 0 seconds\", details: [], metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"date\": \"Mon, 18 Nov 2024 17:28:13 GMT\", \"content-length\": \"0\"} }\n  Timeout error: Deadline Exceeded: status: DeadlineExceeded, message: \"Timeout: Timeout error: Operation 'Search' timed out after 0 seconds\", details: [], metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"date\": \"Mon, 18 Nov 2024 17:28:13 GMT\", \"content-length\": \"0\"} }"
}
```

## Expected Behavior
The request should return a result without any errors. (Running the same request without a timeout set returns a response in less than 1 second.)

## Possible Solution
The issue seems to stem from the timeout settings between nodes (shards), as indicated by the message timed out after 0 seconds. This might be due to subtracting 1 (an integer) instead of a float (less than 1). The issue could be related to the following commits:
- https://github.com/qdrant/qdrant/pull/4844
- https://github.com/qdrant/qdrant/pull/4849 

## Context (Environment)
- **Deployment**: Qdrant deployed in a private Kubernetes cluster using the official Helm Chart.
- **Affected Versions**: At least Qdrant 1.11.5 and 1.12.3. It worked fine with 1.10.1
- **Reproducibility**: The issue is not related to any specific client, as it can be reproduced using Python or Curl.
- **Scope**: The issue occurs only with distributed clusters (multiple shards).

cc @evelynegroen

varshith257:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

This PR addresses **#5463**: Timeout errors when querying distributed Qdrant deployments with a 1-second timeout. The issue stems from timeout calculations during query distribution where time subtraction and conversion to integers result in 0-second timeouts for remote shards, causing validation errors.

### Root Cause
- Timeout calculations used `Duration::as_secs()` or equivalent, truncating sub-second timeouts to 0 seconds after subtracting elapsed time.
- Remote shard queries received timeouts below the minimum required threshold (1 second), leading to `Deadline Exceeded` errors.

### Changes

   - Introduced the `calculate_timeout` function to handle precise floating-point timeout values, ensuring timeout clamping to a minimum of 1 second.
   - The function now supports both `std::time::Instant` and `tokio::time::Instant` for flexibility across codebases.
   - Using clamping, enforced a minimum timeout of 1 second for distributed queries.
   - Added validation checks to avoid passing invalid timeouts to remote shards.
   
   
Closes #5463.
/claim #5463

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6101,9 +6101,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.132"
+version = "1.0.133"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d726bfaff4b320266d395898905d0eba0345aae23b54aee3a737e260fd46db03"
+checksum = "c7fceb2473b9166b2294ef05efcb65a3db80803f0b03ef86a5fc88a2b85ee377"
 dependencies = [
  "indexmap 2.6.0",
  "itoa",


timvisee:
@agourlay I agree. I'd rather keep using the std [`Duration`](https://doc.rust-lang.org/std/time/struct.Duration.html) API everywhere possible. It already has sufficient arithmetic functions.

> timeout.map(|t| t.as_secs().max(1)

You suggest lower-capping at at least 1.

How about using `t.as_sec_f32().ceil() as usize` instead?

timvisee:
> What about handling timeouts in milliseconds (ms) between nodes?

I'd prefer seconds as floating point.

Either way it'd require breaking API changes in some places. So it's not so trivial to implement.

mautini:
Sorry, I wasn’t clear! I’m totally fine with using seconds as a floating-point value. My point was more “we cannot floor or ceil to the nearest second; we need greater precision than an int.”

zeyaddeeb:
I'm actually seeing this issue with 5 second timeouts as well

agourlay:
The issue can happen for other timeout values where the pre-processing steps prior to reaching the remote shard decrease the value to less than 1 second.

I believe the course of action I described is good enough to also tackle the precision concern given that we cannot introduce breaking changes.

We do set the original timeout on the gRPC connection for each remote shard call with

```rust
if let Some(timeout) = timeout {
  request.set_timeout(timeout);
}
```

In that case the connection will be closed which in turn will drop the future object being processed.

timvisee:
@varshith257 Hi! Would you mind to bump your PR to apply the other suggestion.

Basically just update the calculation mentioned [here](https://github.com/qdrant/qdrant/issues/5463#issuecomment-2484279402). And update it to something like `t.as_sec_f32().ceil() as usize`.

I think that'll be good enough.

lib/collection/src/collection/query.rs
```diff
@@ -217,7 +217,12 @@ impl Collection {
         .await?;
 
         // update timeout
-        let timeout = timeout.map(|timeout| timeout.saturating_sub(start.elapsed()));
+        let timeout = timeout.map(|timeout| {
+            // timeout is calculated with sufficient precision and avoids flooring
+            (timeout.as_secs_f32() - start.elapsed().as_secs_f32())
+                .ceil()
+                .max(1.0) as usize
```
timvisee, line 224: Removing this should be fine.

```suggestion
```

In that case it'll always be at least 1 unless it was 0 already. I think that makes sense.

timvisee, line 224: I think this would be better:

```suggestion
            // avoid flooring the timeout to 0
            (timeout.saturating_sub(start).as_secs_f32().ceil() as usize
```

We prevent going below zero. It'll only be 0 if enough time has already passed, and otherwise it's always at least 1.

timvisee (changes requested):
Please do one more round of reformatting, and then it should be good :smile: 

```bash
cargo +nightly fmt --all -- --check
```

agourlay:
FYI I have implemented my suggestions in https://github.com/qdrant/qdrant/pull/5495

PR #5477, opened 2024-11-19
HardwareCounterCell::drop: avoid nested panic

xzfc:
Nested panics caused by `HardwareCounterCell::drop()` abort the program which is inconvenient when debugging failed tests.
To avoid this, this PR adds a check suggested in the Rust docs: https://doc.rust-lang.org/std/ops/trait.Drop.html#panics.


JojiiOfficial (approved):
Nice addition, Thanks!

JojiiOfficial:
We're panicking in the drop impl of `HwMeasurementAcc` as well, so I added the same check to this type too

generall (approved):

PR #5479, opened 2024-11-19
Do not build wal_inspector by default

agourlay:
This PR prevents the `wal_inspector` binary to be built by default, it is now hidden behind the existing `service_debug` feature.

Why?

- The binary is used very rarely.

- It is rather large `979M wal_inspector-e63716334eafa7cb`.

- It takes a significant amount of time to build.

The `service_debug` feature is used in our integration tests so we should be able to catch a breakage early.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -224,6 +224,7 @@ name = "wal_inspector"
 path = "src/wal_inspector.rs"
 test = false
 bench = false
+required-features = ["service_debug"]
 
 [workspace]
 members = [


agourlay:
There is also the `schema_generator` binary but it is used more often. 
Hiding it behind a feature flag might make the dev process more confusing.

timvisee:
> It is rather large 979M wal_inspector-e63716334eafa7cb.

:exploding_head: 

For the sake of quicker builds this is perfectly acceptable to me.

If you ask me, I'd also put the schema generator behind a feature flag. The only way I ever use it is through the `tools/generate_openapi_models.sh` script which can provide that flag by default.

timvisee (approved):

agourlay:
@timvisee you make a good point about adding the flag to `generate_openapi_models.sh` for the other binary.

I will explore it in the different PR to keep things clean :+1:

Issue #5483, opened 2024-11-19
Panic on nearest neighbour search with high limit


Michael-JB:
## Current Behavior

In Qdrant v1.12.4, nearest neighbour searches/queries with a high limit (e.g., `u64::MAX`) cause Qdrant to panic. It seems Qdrant now tries to allocate O(limit) memory for nearest neighbour searches.

This is a regression in v1.12.4. This bug is not present in v1.12.3.

## Steps to Reproduce

The following test fails in v1.12.4:

```rust
#[tokio::test]
async fn nearest_neighbour_limit() {
    let collection_name = "test_collection";
    let query_embedding = vec![1.0; 128];

    let qdrant = Qdrant::new(QdrantConfig::from_url(qdrant_url().as_str())).unwrap();

    qdrant
        .create_collection(
            CreateCollectionBuilder::new(collection_name)
                .vectors_config(VectorParamsBuilder::new(128, Distance::Cosine)),
        )
        .await
        .unwrap();

    qdrant
        .query(
            QueryPointsBuilder::new(collection_name)
                .query(Query::new_nearest(query_embedding))
                .limit(u64::MAX)
                .build(),
        )
        .await
        .unwrap();
}
```

Running this test causes the following Qdrant panic:

```
2024-11-19T15:46:34.643154Z  INFO storage::content_manager::toc::collection_meta_ops: Creating collection test_collection
2024-11-19T15:46:34.700389Z ERROR qdrant::startup: Panic backtrace:
   0: std::backtrace::Backtrace::create
   1: qdrant::startup::setup_panic_hook::{{closure}}
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::begin_panic_handler::{{closure}}
   4: std::sys::backtrace::__rust_end_short_backtrace
   5: rust_begin_unwind
   6: core::panicking::panic_fmt
   7: hashbrown::raw::Fallibility::capacity_overflow
   8: hashbrown::raw::RawTable<T,A>::with_capacity_in
   9: collection::shards::local_shard::search::<impl collection::shards::local_shard::LocalShard>::do_search::{{closure}}
  10: collection::shards::local_shard::shard_ops::<impl collection::shards::shard_trait::ShardOperation for collection::shards::local_shard::LocalShard>::query_batch::{{closure}}
  11: collection::shards::replica_set::read_ops::<impl collection::shards::replica_set::ShardReplicaSet>::query_batch::{{closure}}::{{closure}}::{{closure}}
[...]

PR #5484, opened 2024-11-19
Fix panic on query with huge limit

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/5483>

[Here](https://github.com/qdrant/qdrant/pull/5428/files#diff-9a15df0820bef38d5804ec1676101484e170367b109289cb02410109db341d2aR17) we started to preallocate memory in the search result aggregator based on the user provided limit. If a user provides a huge limit it will panic due to an allocation failure.

This PR fixes the problem by limiting the size of preallocating to one million items, exactly like `FixedLengthPriorityQueue` does. I'm not entirely sure if one million is reasonable here, and it might be too big. Lets discuss here what we think may be reasonable.

I've also fixed a limit overflow in `FixedLengthPriorityQueue` if a `limit` of `u64::MAX` is given, and an overflow in search.

### All Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/search_result_aggregator.rs
+++ b/lib/collection/src/collection_manager/search_result_aggregator.rs
@@ -5,6 +5,9 @@ use common::fixed_length_priority_queue::FixedLengthPriorityQueue;
 use common::types::ScoreType;
 use segment::types::{PointIdType, ScoredPoint, SeqNumberType};
 
+/// Avoid excessive memory allocation and allocation failures on huge limits
+const LARGEST_REASONABLE_ALLOCATION_SIZE: usize = 1_048_576;
+
 pub struct SearchResultAggregator {
     queue: FixedLengthPriorityQueue<ScoredPoint>,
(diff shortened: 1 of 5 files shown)


lib/collection/src/collection_manager/search_result_aggregator.rs
```diff
@@ -5,6 +5,9 @@ use common::fixed_length_priority_queue::FixedLengthPriorityQueue;
 use common::types::ScoreType;
 use segment::types::{PointIdType, ScoredPoint, SeqNumberType};
 
+/// Avoid excessive memory allocation and allocation failures on huge limits
+const LARGEST_REASONABLE_ALLOCATION_SIZE: usize = 1_048_576;
```
agourlay, line 9: I think a lower value should be enough if we want to protect ourselves.

Something like `8,192`

generall (approved):

agourlay (approved):
Thanks for the tests :pray:

PR #5486, opened 2024-11-20
Drop `GraphLayersBackwardCompatibility`

xzfc:
This PR drops the support for reading/converting of the old `graph.bin` file format used before Qdrant 0.8.4.
This old format stores everything in a single `graph.bin` file while newer versions of qdrant store the graph in two files, `graph.bin`+`links.bin`.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_layers.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers.rs
@@ -4,7 +4,7 @@ use std::path::{Path, PathBuf};
 
 use common::fixed_length_priority_queue::FixedLengthPriorityQueue;
 use common::types::{PointOffsetType, ScoredPointOffset};
-use io::file_operations::{atomic_save_bin, read_bin, FileStorageError};
+use io::file_operations::{atomic_save_bin, read_bin};
 use itertools::Itertools;
 use memory::mmap_ops;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -34,17 +33,6 @@ struct GraphLayerData<'a> {
...
 }
 
-/// Contents of the `graph.bin` file (Qdrant 0.8.4).
-#[derive(Deserialize, Serialize, Debug)]
-struct GraphLayersBackwardCompatibility {
-    max_level: usize,
-    m: usize,
-    m0: usize,
-    ef_construct: usize,
-    links_layers: Vec<LayersContainer>,
```
IvanPleshkov, line 44: Can you also delete please
```
pub type LayersContainer = Vec<LinkContainer>;
```
It's unused in this PR.

IvanPleshkov (approved):

timvisee (approved):
Out of curiosity. What happens if a user still has this file? Will we just ignore it, or do we crash?

IvanPleshkov:
> Out of curiosity. What happens if a user still has this file? Will we just ignore it, or do we crash?
> 
> I'd prefer if we could ignore it if that isn't too much work so that we can build a fresh HNSW graph later on.

I agree with this question. We don't have a conversion from the old format. Such situations are possible even if a user updates qdrant correctly (minor-by-minor version). We probably need to implement a converter and remove everything in the next minor release. Let's discuss it. To avoid spontaneously merging before the decision, I mark this PR as changes required.

IvanPleshkov (changes requested):
Mark as requested changes while old version support discussion

timvisee:
> Such situations are possible even if a user updates qdrant correctly (minor-by-minor version).

Good point.

IvanPleshkov:
> > Such situations are possible even if a user updates qdrant correctly (minor-by-minor version).
> 
> Good point.

I'm reading current `dev` and it seems we have conversion and we can merge this PR. @xzfc do you have a chance to check manually if the old format will work correctly?

IvanPleshkov (approved):

xzfc:
The conversion happens at `0.11.1` → `1.0.0` with the following message:
```
[segment::segment_constructor::segment_constructor_base] Migrating segment 0.4.0 -> 0.5.0
```

So, this was a dead code already.

PR #5487, opened 2024-11-20
Add common::bitpacking

xzfc:
This PR adds a new module `common::bitpacking` with two structs: `BitPacker` and `BitUnpacker`. They would be used for the HNSW graph links compression in subsequent PRs.

Comparison with other implementations:
- [`bitpacking`](https://docs.rs/bitpacking/latest/bitpacking/): it's fast, SIMD-optimized, but application is limited: blocks should be of size of 32/128/256 integers, all stored integers within a block should have the same bitsize.
- [`bitstream_io`](https://docs.rs/bitstream-io/latest/bitstream_io/): the idea is somewhat similar to this PR, but this crate resulted to be slower for links compression/in-memory use-case.
- [`tantivy_bitpacker`](https://docs.rs/tantivy-bitpacker/latest/tantivy_bitpacker/): optimized for a random access reads within a packed array.

The most important function is `BitUnpacker::read()` as it would be used in a hot loop. The inner buffer size is 8 bytes (a single `u64`) and it's optimized to read by whole 8 bytes at once.

The intended use-case for this module would look like the following (example numbers are arbitrary):
- Store header: 5 bit.
- Store first 32 values, each of size 12 bits. (e.g. sorted/delta-encoded for the main graph).
- Store last 20 values, each of size 15 bits. (e.g. additional links).
Total size of the example: $\lceil \frac{5 + 32 \times 12 + 20 \times 15}{8} \rceil = 87 \ \text{bytes}$.

### DIFF
--- a/lib/common/common/src/bitpacking.rs
+++ b/lib/common/common/src/bitpacking.rs
@@ -1,0 +1,138 @@
+pub struct BitPacker<'a> {
+    output: &'a mut Vec<u8>,
+    buf: u64,
+    buf_bits: u8,
+}
+
+impl<'a> BitPacker<'a> {
+    #[inline]
+    pub fn new(output: &'a mut Vec<u8>) -> Self {
+        output.clear();
+        Self {
+            output,
+            buf: 0,
+            buf_bits: 0,
+        }
+    }
+
+    #[inline]
+    pub fn push(&mut self, value: u32, bits: u8) {
+        if self.buf_bits + bits >= 64 {
+            self.buf |= u64::from(value) << self.buf_bits;
+            self.output.extend_from_slice(&self.buf.to_le_bytes());
+            self.buf = u64::from(value) >> (64 - self.buf_bits);
+            self.buf_bits = self.buf_bits + bits - 64;
+        } else {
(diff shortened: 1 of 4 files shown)


timvisee (approved):
I'm not confident I understand all the nitty-gritty. However, I'm assuming that you've already extensively tested this. Based on a quick skim I don't have any remarks, and I like that you split it into a sub crate.

lib/common/common/src/bitpacking.rs
```diff
@@ -0,0 +1,138 @@
...
+        output.clear();
+        Self {
+            output,
+            buf: 0,
+            buf_bits: 0,
+        }
+    }
+
+    #[inline]
+    pub fn push(&mut self, value: u32, bits: u8) {
```
generall, line 19: Could you please provide some examples of how to use this function in real-life scenario. For instance, where `bits` is coming from? I see in the provided tests, that bits are used to generate value. Does it means we can derive `bits` from `value` inside the push function?

generall (approved):
ToDo: comment that the Bitpacker abstraction expects some external entity to keep track of value bit lengths

generall (approved):

PR #5488, opened 2024-11-20
Improve timeout in distance matrix

agourlay:
Computing the distance matrix requires several stages of querying.

The timeout should be updated mid-way to reflect the time already elapsed.


timvisee (approved):
Nice!

PR #5489, opened 2024-11-20
GraphLinksConverter: take `m` and `compressed` arguments

xzfc:
This PR performs some integration work required for HNSW links compression.
- `GraphLinksConverter::new()` now takes `m: usize` and `compression: bool`. It is assumed that compression algorithm might sort the first `m`/`2*m` links for each node.
- This PR doesn't include the compression algorithm, so these parameters are ignored for now.
- Some unit tests are updated to test both compressed and uncompressed implementations.
- A new undocumented environment var is added: `__QDRANT_COMPRESSED_LINKS`. When set, it will enable creation of compressed graphs for testing/debugging purposes. By default it's disabled.

Assumption: `m0 == 2*m`.

Depends on #5486, ignore the first commit during review.

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -94,6 +94,25 @@ pub struct HnswIndexOpenArgs<'a> {
     pub stopped: &'a AtomicBool,
 }
 
+fn use_links_compression(existing: bool) -> bool {
+    let name = "__QDRANT_COMPRESSED_LINKS";
+    match std::env::var(name).as_deref() {
+        // 0: Only read compressed links, but never write.
+        Ok("0") | Err(std::env::VarError::NotPresent) => false,
+        // 1: Enable for newly created graphs.
+        Ok("1") => !existing,
+        // 2: Same as 1 + convert existing graphs.
+        Ok("2") => true,
+        _ => {
+            log::warn!(
+                "Unknown value for {name}={:?}, defaulting to 0",
+                std::env::var_os(name),
+            );
+            false
+        }
+    }
+}
+
 impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
     pub fn open(args: HnswIndexOpenArgs<'_>) -> OperationResult<Self> {
         let HnswIndexOpenArgs {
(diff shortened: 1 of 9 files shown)


lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -412,8 +412,13 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
         config.indexed_vector_count.replace(indexed_vectors);
 
         let graph_links_path = GraphLayers::<TGraphLinks>::get_links_path(path);
-        let graph: GraphLayers<TGraphLinks> =
-            graph_layers_builder.into_graph_layers(Some(&graph_links_path))?;
+        let use_compression =
+            std::env::var_os("__QDRANT_COMPRESSED_LINKS").map_or(false, |x| !x.is_empty());
```
timvisee, line 416: Clippy for Rust 1.83 will start suggesting this:

```suggestion
            std::env::var_os("__QDRANT_COMPRESSED_LINKS").is_some_and(|x| !x.is_empty());
```

timvisee (approved):

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -94,6 +94,25 @@ pub struct HnswIndexOpenArgs<'a> {
     pub stopped: &'a AtomicBool,
 }
 
+fn use_links_compression(existing: bool) -> bool {
```
IvanPleshkov, line 97: Not sure about `existing` parameter here. How do we handle the case when we create compressed links and reload them? Do we disable enabled compression?

IvanPleshkov (changes requested):
Requested change to refactor hardcoded `m0 = m * 2` logic. I find important to move it as a separate function

lib/segment/src/index/hnsw_index/hnsw.rs
xzfc: Rewritten to make it more clear.

(Read-only) usage of the compression during search is determined based the on-disk file contents. It's not configurable, we can read both formats regardless of the setting.

This setting control when to write it: never/only for new indices/migrate old indices during startup.

IvanPleshkov (approved):

PR #5490, opened 2024-11-20
GraphLinks::for_each_link

xzfc:
This PR slightly updates the `GraphLinks` trait:
- Removed: `fn links()`; it was `iter()`-like, i.e. returns an iterator.
- Added instead: `fn for_each_link()`; it's `for_each()`-like, i.e. takes a closure.
- Also, added `#[cfg(test)] fn links_vec()`; returns a vector. Obliviously, it's slow, but it used only in tests.

The reason is that it would be easier for a compiler to generate inline-able code. (and the the subsequent PRs depend on this change)

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -419,11 +419,9 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
         {
             for (idx, deleted) in deleted_bitslice.iter().enumerate() {
                 if *deleted {
-                    debug_assert!(graph
-                        .links
-                        .links(idx as PointOffsetType, 0)
-                        .next()
-                        .is_none());
+                    graph.links.for_each_link(idx as PointOffsetType, 0, |_| {
+                        panic!("Deleted point in the graph");
+                    });
                 }
             }
         }
--- a/lib/segment/src/index/hnsw_index/tests/test_graph_connectivity.rs
+++ b/lib/segment/src/index/hnsw_index/tests/test_graph_connectivity.rs
@@ -96,7 +96,7 @@ fn test_graph_connectivity() {
         let links = hnsw_index
             .graph()
             .links
-            .links(point_id as PointOffsetType, 0);
+            .links_vec(point_id as PointOffsetType, 0);
         for link in links {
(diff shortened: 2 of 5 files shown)


lib/segment/src/index/hnsw_index/tests/test_graph_connectivity.rs
```diff
@@ -96,7 +96,7 @@ fn test_graph_connectivity() {
         let links = hnsw_index
             .graph()
             .links
-            .links(point_id as PointOffsetType, 0);
+            .links_vec(point_id as PointOffsetType, 0);
```
timvisee, line 99: Are you saying that a change like this allows compile time inlining?

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -419,11 +419,9 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
         {
             for (idx, deleted) in deleted_bitslice.iter().enumerate() {
                 if *deleted {
-                    debug_assert!(graph
-                        .links
-                        .links(idx as PointOffsetType, 0)
-                        .next()
-                        .is_none());
+                    graph.links.for_each_link(idx as PointOffsetType, 0, |_| {
+                        panic!("Deleted point in the graph");
```
timvisee, line 423: I'd prefer to keep a debug-level panic here. You never know.

timvisee (approved):

lib/segment/src/index/hnsw_index/hnsw.rs
xzfc: It's behind `#[cfg(debug_assertions)]`, check 9 lines above.

lib/segment/src/index/hnsw_index/tests/test_graph_connectivity.rs
xzfc: For `links_vec()` — no (and probably it will make the performance worse). But it's only for tests.

For `links()` → `for_each_link()`, yes.

lib/segment/src/index/hnsw_index/hnsw.rs
timvisee: Thanks! The GitHub diff hid it from me :face_holding_back_tears:

PR #5495, opened 2024-11-20
Improve timeout handling in remote shards

agourlay:
Some queries require to perform several I/O action in a sequence before responding to the client.

Between each action, we decrease the timeout in order to track precisely the elapsed time to honor the timeout input from the client.

This logic triggers two different failure modes for the remote shard when:
- the timeout is zero:
  -  the remote shard does not accept timeout=0 as input
- the timeout is between 0 and 1 sec
  - given that the value is floored, it triggers the previous case as well

This PR proposes to:
- gate the access to remote shard with non zero timeout to shortcut the query processing
- round up the timeout value to the next greater integer. 
  - e.g `0.5 sec` would become `1 sec` instead of timing out
- unify our timeout param. value with the one used for the `gRPC` connection timeout

This gives the following change:

| Code   | Qdrant timeout   | gRPC timeout   |
|------------|------------|------------|
| dev | floored sec | untouched |
| PR | round up sec | round up sec |

The new approach is more lenient as it gives more time to finish to all requests.

### DIFF
--- a/lib/collection/src/shards/remote_shard.rs
+++ b/lib/collection/src/shards/remote_shard.rs
@@ -630,6 +630,16 @@ impl RemoteShard {
 
         Ok(())
     }
+
+    /// Check if the timeout is zero and return an error if it is.
+    /// Used before making a request to the remote shard to avoid useless traffic.
+    pub fn check_timeout(timeout: Option<Duration>, operation: &str) -> CollectionResult<()> {
+        if timeout.map(|t| t.is_zero()).unwrap_or(false) {
+            Err(CollectionError::timeout(0, operation))
+        } else {
+            Ok(())
+        }
+    }
 }
 
 // New-type to own the type in the crate for conversions via From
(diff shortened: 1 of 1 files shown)


lib/collection/src/shards/remote_shard.rs
```diff
@@ -630,6 +630,16 @@ impl RemoteShard {
 
         Ok(())
     }
+
+    /// Check if the timeout is zero and return an error if it is.
+    /// Used before making a request to the remote shard to avoid useless traffic.
+    pub fn check_timeout(timeout: Option<Duration>, operation: &str) -> CollectionResult<()> {
+        if timeout.map(|t| t.is_zero()).unwrap_or(false) {
+            Err(CollectionError::timeout(0, operation))
```
agourlay, line 638: we do not have the original timeout so the error is not great for the user here

generall (approved):

PR #5496, opened 2024-11-20
Fix: Timeout Precision in Remote Shard Queries

### LINKED ISSUE #5463: Timeout error on query if timeout=1 in a distributed deployment

varshith257:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

Fixes #5463 
/claim #5463 

Follow up to #5475 Discuusions

### DIFF
--- a/lib/collection/src/collection/query.rs
+++ b/lib/collection/src/collection/query.rs
@@ -217,7 +217,10 @@ impl Collection {
         .await?;
 
         // update timeout
-        let timeout = timeout.map(|timeout| timeout.saturating_sub(start.elapsed()));
+        let timeout = timeout.map(|timeout| {
+            // avoid flooring the timeout to 0
+            (timeout.saturating_sub(start).as_secs_f32().ceil() as usize)
+        });
 
         // Check we actually fetched all referenced vectors from the resolver requests
         for (resolver_req, _) in &resolver_requests {
--- a/lib/collection/src/collection/query.rs
+++ b/lib/collection/src/collection/query.rs
@@ -219,9 +219,9 @@ impl Collection {
         // update timeout
         let timeout = timeout.map(|timeout| {
             // avoid flooring the timeout to 0
-            Duration::from_secs_f32(
-                timeout.saturating_sub(start.elapsed()).as_secs_f32().ceil() as usize
-            )
+            let remaining_secs =
+                timeout.saturating_sub(start.elapsed()).as_secs_f32().ceil() as usize;
+            Duration::from_secs(remaining_secs as u64)
         });
 
         // Check we actually fetched all referenced vectors from the resolver requests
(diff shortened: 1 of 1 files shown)


varshith257:
@generall @timvisee Any reviews on this?

generall:
Hey @varshith257, thanks for the PR! Unfortunately, I think it doesn't cover all cases. For example request like

```
POST collections/test_collection/points/query/groups?timeout=1
{
    "group_by": "field-a",
    "query": 1,
    "with_lookup": {
        "collection": "test_collection"
    }
}
```

would still unexpectedly fail with the timeout. 
For this reason I would prefer to go with this PR instead: https://github.com/qdrant/qdrant/pull/5495 it covers the timeout rounding in a more systematic way, and therefore it would be easier to spot the problem like this in future.

PR #5501, opened 2024-11-21
Strict mode max collection vector size

JojiiOfficial:
Adds a new config option to strict mode that allows setting a max vector storage size.

This PR focuses only on single-node setups which will be addressed in a separate PR.


lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -1004,6 +1004,27 @@ impl ShardReplicaSet {
...
+        self.local
+            .read()
+            .await
+            .as_ref()
+            .map(|i| match i {
+                Shard::Local(local) => local
+                    .segments()
+                    .read()
+                    .iter()
+                    .map(|i| i.1.get().read().info().vectors_size_bytes)
```
agourlay, line 1019: Are we going to call `info()` on all segments for each request?
Have you checked how performant it is or am I needlessly worrying :)

JojiiOfficial: Benchmark results:

With 2 payload indices:
```
segment-info            time:   [285.36 ns 285.57 ns 285.96 ns]
```

With 3 payload indices:
```
segment-info            time:   [364.87 ns 365.16 ns 365.58 ns]
```

Without payload index:
```
segment-info            time:   [100.33 ns 100.49 ns 100.75 ns]
```

Without payload index and calling `.clone()`:
```
segment-info            time:   [93.023 ns 93.194 ns 93.426 ns]
```

The performance of `info()` depends on the amount of indexed fields, since the schema [gets cloned](https://github.com/qdrant/qdrant/blob/dev/lib/segment/src/segment/entry.rs#L436).
I think this is still fast enough for our use case.

timvisee: This definitely is expensive, partially because of all the locking that needs to happen and because it collects much more than we actually need.

I'm not confident this is worth it on each request, especially if a lot of other operations are going on at the same time. I think we should be looking at a different approach.

Maybe this works: a variable on the shard level holding the current size estimate, which we update on each flush.

generall: Original idea was to keep cache and update it every ~32 update operations. So info should not be called on each request

agourlay: Thanks for chiming in, I would prefer if we introduce the amortized approach right away :pray:

JojiiOfficial:
With the latest changes the values get cached using `AtomicUsize` types (which are faster than rw-locking) and updated every 32 requests.
I also added a new function `size_info()` that doesn't clone the payload storage in normal segments.

Benches for the new `size_info()` function:
```
segment-size-info       time:   [87.147 ns 87.203 ns 87.267 ns]
```

lib/collection/src/collection/mod.rs
```diff
@@ -283,6 +287,7 @@ impl Collection {
             update_runtime: update_runtime.unwrap_or_else(Handle::current),
             search_runtime: search_runtime.unwrap_or_else(Handle::current),
             optimizer_cpu_budget,
+            local_stats_cache: LocalDataStatsCache::default(),
```
agourlay, line 290: What about computing the stats once on collection load as well?
Otherwise it seems the 32 first requests will not benefit from the cache?
WDYT?

lib/collection/src/operations/verification/update.rs
```diff
@@ -56,6 +56,18 @@ impl StrictModeVerification for DeleteVectors {
 }
 
 impl StrictModeVerification for SetPayload {
+    // TODO: Payload storage size limit
```
agourlay, line 59: this one can now be implemented in a different PR

I can take care of it later to get use to the `StrictModeVerification` infra :)

lib/collection/src/collection/mod.rs
JojiiOfficial: The cache gets loaded on the first request since [this](https://github.com/qdrant/qdrant/pull/5501/files#diff-f91d53257ec619a1403e6122ec4c7f004fffe1f32bd91286f6950dbcab39c3a5R20) check evaluates to `true` when called the first time.
We however can still load the stats after loading the collection to improve on performance 👍🏻

docs/grpc/docs.md
```diff
@@ -1431,6 +1431,7 @@ Note: 1kB = 1 vector of size 256. |
 | search_allow_exact | [bool](#bool) | optional |  |
 | search_max_oversampling | [float](#float) | optional |  |
 | upsert_max_batchsize | [uint64](#uint64) | optional |  |
+| max_collection_vector_size | [uint64](#uint64) | optional |  |
```
agourlay, line 1434: I think this one should have clear unit in the name.

Also it is possible that our users prefer working with megabytes right away.

WDYT?

lib/segment/src/segment/entry.rs
```diff
@@ -501,11 +491,28 @@ impl SegmentEntry for Segment {
             ram_usage_bytes: 0, // ToDo: Implement
             disk_usage_bytes: 0, // ToDo: Implement
             is_appendable: self.appendable_flag,
-            index_schema: schema,
+            index_schema: HashMap::new(),
```
agourlay, line 494: I'd love a small comment here explaining that it is left empty done on purpose.

agourlay: Just found the comment on the trait, I guess you can discard my nitpick :+1:

docs/grpc/docs.md
agourlay: I was reminded by someone that this is an internal field, manipulated by our custom UI most likely.
Sticking to bytes is good enough then :+1:

agourlay (approved):

lib/segment/src/segment/entry.rs
```diff
@@ -430,17 +430,7 @@ impl SegmentEntry for Segment {
...
-        let schema = payload_index
-            .indexed_fields()
-            .into_iter()
-            .map(|(key, index_schema)| {
-                let points_count = payload_index.indexed_points(&key);
-                (key, PayloadIndexInfo::new(index_schema, points_count))
-            })
-            .collect();
-
+    fn size_info(&self) -> SegmentInfo {
```
timvisee, line 433: We might want a separate type here to prevent accidental misuse?

tests/openapi/test_strictmode.py
```diff
@@ -461,3 +462,144 @@ def search_request():
...
+    set_strict_mode(collection_name, {
+        "enabled": True,
+        "max_collection_vector_size_bytes": 240,
+    })
+
+    for _ in range(33):
+        upsert_points([6, 7, 8, 9, 10]).raise_for_status()
+
+    # Max limit has been reached and one of the next requests must fail. Due to cache it might not be the first call!
+    for _ in range(33):
```
timvisee, line 596: Shouldn't 32 be enough? (also in other places)

```suggestion
    for _ in range(32):
```

lib/collection/src/collection/mod.rs
```diff
@@ -782,6 +795,39 @@ impl Collection {
...
+        self.local_stats_cache.get()
+    }
+
+    /// Returns the estimated local vector storage size for this collection, cached and auto-updated.
+    pub async fn estimated_local_vector_storage_size(&self) -> usize {
+        if let Some(shard_stats) = self.check_and_update_local_size_stats().await {
+            return shard_stats.vector_storage_size;
+        }
+        self.local_stats_cache.get_vector_storage()
+    }
```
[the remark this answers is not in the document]
JojiiOfficial: We don't check both in a request. The first function was intended for places where we need all data eg. if we want to sync those stats with other nodes in a cluster or check all statistics in a request and is currently not in use.
The second function only returns vector storage size to prevent fetching more atomic values than we actually need for less performance impact.

If you still think we should remove it and add when needed, I'll remove it from this PR.

lib/segment/src/segment/entry.rs
JojiiOfficial: Since we're not differentiating between `info` and `size_info()` in proxy segments and this type being used directly in the API, I tried to avoid changing this type and thought a proper comment would suffice.

If you prefer having a separate type, I'll implement it and try to not break the API (which could be possible using `#[serde(flatten)]` but haven't checked)

lib/collection/src/collection/mod.rs
```diff
@@ -261,6 +269,10 @@ impl Collection {
 
         let locked_shard_holder = Arc::new(LockedShardHolder::new(shard_holder));
 
+        let local_stats_cache = LocalDataStatsCache::new_with_values(
+            Self::load_segment_stats(&locked_shard_holder).await,
+        );
```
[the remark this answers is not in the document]
JojiiOfficial: I agree
We could make this load optional for collections that have strict mode disabled. But since we can afford this check in the API, the performance overhead at collection initialization should not be too high.

agourlay:
Can we get this one merged? :)

lib/collection/src/operations/verification/update.rs
agourlay: Started in https://github.com/qdrant/qdrant/pull/5588

lib/collection/src/collection/mod.rs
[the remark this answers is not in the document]
timvisee: If you're not 99% but 100% sure you'll use it later, then I'm happy :smiley:

lib/segment/src/segment/entry.rs
timvisee: > If you prefer having a separate type, I'll implement it and try to not break the API (which could be possible using #[serde(flatten)] but haven't checked)

I'd also be fine with some overlap of fields between two types, that may be easier without flattening.

But I don't have a strong preference on having a separate type or not. I'd also be happy to keep it for now.

timvisee (approved):

PR #5506, opened 2024-11-22
better handle non-transient errors

generall:
During stream shard transfer in ForwardShardProxy we modify `wait` parameter for internal and forwarded requests.
This can lead to inconsistent responses across replicas, as wait=true/false affects when the validation of request is happening. It might be, that some replicas accept the request, while the one with shard transfer doesn't. 

This inconsistency triggers the consistency resolution mechanism, which previously ignored the difference between transient and non-transient errors. 

This PR makes consistency resolution more careful with non-transient errors and only deactivate the replica if there are others who completely applied update operation

### DIFF
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -363,12 +363,30 @@ impl ShardReplicaSet {
                 );
             }
 
+            // If there is at least one full-complete operation, we can't ignore non-transient errors (4xx)
+            // And we must deactivate failed replicas to ensure consistency
+            let has_full_completed_updates = successes.iter().any(|(_, res)| match res.status {
+                UpdateStatus::Completed => true,
+                UpdateStatus::Acknowledged => false,
+                UpdateStatus::ClockRejected => false,
+            });
+
             if successes.len() >= minimal_success_count {
                 // If there are enough successes, deactivate failed replicas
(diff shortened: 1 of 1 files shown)


lib/collection/src/shards/replica_set/update.rs
```diff
@@ -363,12 +363,30 @@ impl ShardReplicaSet {
                 );
             }
 
+            // If there is at least one full-complete operation, we can't ignore non-transient errors (4xx)
+            // And we must deactivate failed replicas to ensure consistency
```
timvisee, line 367: I don't fully understand why this is the case.

Isn't exactly the same true for all acknowledged updates, because we can assume they'll be applied (eventually) as well?

Could you provide an example which was problematic and has improved now, and potentially a counter example on why it isn't necessary for just acknowledged updates?

generall: image

https://github.com/qdrant/qdrant/blob/5b578c4f34188f0474f901e49d4726213596433d/lib/collection/src/shards/forward_proxy_shard.rs#L243

timvisee (approved):

ffuugoo (approved):
I still don't like it and feel like there should be a better way to handle this, but I did not figure it out yet, so I propose we merge it and see how it works, and I'll try to revisit it later. 😔

PR #5508, opened 2024-11-22
Also limit the batch-size for vector updates

JojiiOfficial:
Depends on #5501 

Also limit vector update operation with `upsert_max_batchsize` strict mode config.

### DIFF
--- a/lib/api/src/rest/schema.rs
+++ b/lib/api/src/rest/schema.rs
@@ -1032,3 +1032,13 @@ impl Validate for PointInsertOperations {
         }
     }
 }
+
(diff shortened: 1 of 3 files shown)


lib/api/src/rest/schema.rs
```diff
@@ -1032,3 +1032,17 @@ impl Validate for PointInsertOperations {
         }
     }
 }
+
+impl PointInsertOperations {
+    /// Amonut of vectors in the operation request.
```
timvisee, line 1037:
```suggestion
    /// Amount of vectors in the operation request.
```

PR #5509, opened 2024-11-22
Expose storage size of RocksDB payload storage

agourlay:
Introduces a new feature to retrieve the RocksDB storage size of a given column family.

I tried various approach before settling on this one, it seems to give a good proxy for what is being persisted.

This function is introduced mostly for consistency sake as we want to support this feature on the payload storage trait.
It is unlikely that it will be used meaningfully in practice.


timvisee (approved):
I like that it is 'as simple' as getting the `size` property. Nice!

Issue #5243, opened 2024-10-15
Can't return collection config to default null value


ffletcherr:
I changed `max_optimization_threads` from `null` to `2`. using this command in qdrant dashboard console:

```
PATCH /collections/test
{
    "optimizers_config": {
        "max_optimization_threads": 2
    }
}
```

when run this to set it `null` again the value doesn't change (without any error or complain):

```
PATCH /collections/test
{
    "optimizers_config": {
        "max_optimization_threads": null
    }
}
```

qdrant version : 1.12.1

generall:
This is known behaviour. From the protocol level, there are no difference between null and empty, so we qdrant can't distinguish those.

generall:
Potential fix could be to replace `max_optimization_threads` type from integer to Enum of {Auto, int}, so we could explicitly set it to null

PR #5520, opened 2024-11-25
Allow max-optimization-thread config to be set to null on update

gulshan-rs:
### All Submissions:

Fixes https://github.com/qdrant/qdrant/issues/5243

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/operations/conversions.rs
+++ b/lib/collection/src/operations/conversions.rs
@@ -14,7 +14,7 @@ use api::grpc::qdrant::update_collection_cluster_setup_request::{
 use api::grpc::qdrant::{CreateShardKey, Vectors};
 use api::rest::schema::ShardKeySelector;
 use api::rest::BaseGroupRequest;
-use common::types::ScoreType;
+use common::types::{MaxOptimizationThreads, ScoreType};
 use itertools::Itertools;
 use segment::common::operation_error::OperationError;
 use segment::data_types::vectors::{
@@ -426,10 +426,11 @@ impl From<CollectionInfo> for api::grpc::qdrant::CollectionInfo {
                         .indexing_threshold
                         .map(|x| x as u64),
                     flush_interval_sec: Some(config.optimizer_config.flush_interval_sec),
-                    max_optimization_threads: config
-                        .optimizer_config
-                        .max_optimization_threads
-                        .map(|n| n as u64),
+                    max_optimization_threads: match config.optimizer_config.max_optimization_threads
+                    {
+                        MaxOptimizationThreads::Auto => None,
+                        MaxOptimizationThreads::Threads(n) => Some(n as u64),
+                    },
                 }),
                 wal_config: config
                     .wal_config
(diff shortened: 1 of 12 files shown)


lib/collection/src/operations/conversions.rs
```diff
@@ -499,7 +500,9 @@ impl From<api::grpc::qdrant::OptimizersConfigDiff> for OptimizersConfig {
             flush_interval_sec: optimizer_config.flush_interval_sec.unwrap_or_default(),
             max_optimization_threads: optimizer_config
                 .max_optimization_threads
-                .map(|n| n as usize),
+                .map_or(MaxOptimizationThreads::Auto, |n| {
+                    MaxOptimizationThreads::Threads(n as usize)
+                }),
```
coszio, line 505: This means that even if `max_optimization_threads` is not specified in a grpc client, it will set it to `Auto`, which is not necessarily the intention.

coszio (changes requested):
Nice job, @gulshan-rs! However, there are interface changes still to be made for this to work as expected.

1. GRPC interface needs to be updated as well with a new field (old field can be renamed, but can't change its numbering). 
Surely it will need deprecation, but to allow zero-downtime upgrades, for now we will need to be able to read both fields (new and old). We can take care of safely removing the old field after a releasing a later version.
1. Right now the generated openapi model for `MaxOptimizationThreads` accepts `null` or a number, but we want `"auto"` or a number.
1. Please run `./tools/generate_openapi_models.sh` (for REST) and `./tools/generate_grpc_docs.sh` (for grpc) after making the respective interface updates.

lib/collection/src/operations/conversions.rs
gulshan-rs: Thanks for the review and detailed comments. I will take another pass at it. 

My initial intention was to not make any change in the grpc client so as to avoid any need of deprecation (tags) and worry about backward compatibility. Hence I decided to only change the inners i.e the ```OptimizersConfig``` and handle remaps via ```conversions``` --  but it seems broken. I will rerun integration tests locally to see what is broken or follow what you have suggested.

timvisee:
I'm afraid we'll not be able to make this change, because as far as I understand is this will introduce a breaking change on the API level.

Also note that we cannot use `null` in the REST interface here because some of our clients cannot distinguish between the existence of a field or it being `null`.

gulshan-rs:
> Also note that we cannot use null in the REST interface here. 

Ok. I am more concerned with this one, yes this will be breaking changes for existing customers. I am going to pause work and discard this PR. thankyou both for all your quick feedback.

timvisee:
Thank you so much for giving it a shot though!

I'll let you know if I think of another way to potentially approach this.

lib/collection/src/operations/conversions.rs
coszio: Unfortunately we don't have a test which would break with this change yet, nor an easy framework for testing grpc interface. Most grpc changes are just validated with the compiler.

For REST, though, a test could be made to validate that setting a number or `"auto"` works ok.

PR #5521, opened 2024-11-26
SegmentSearcher stream results

agourlay:
Similar to https://github.com/qdrant/qdrant/pull/5505 but for segments during search.

Process the results from segments out of order as soon as possible.

Tested against CI benchmarks does not show any visible improvements.
I assume it shines with a large number of segments which are not balanced.

### DIFF
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -5,7 +5,8 @@ use std::sync::Arc;
 
 use common::counter::hardware_accumulator::HwMeasurementAcc;
 use common::types::ScoreType;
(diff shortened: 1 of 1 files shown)


timvisee (approved):
What is the expected benefit here?

Just that we digest the search results right away, instead of keeping all response states until the end?

agourlay:
> Just that we digest the search results right away, instead of keeping all response states until the end?

Yes and we do not allocate the outer vector result form

```rust
let search_results_per_segment_res = searches.await?;
```

The change is a nice to have, ready to close the PR if you are not convinced :)

timvisee:
Totally fine with the change. I only wanted to confirm our expectations match.

:heavy_check_mark:

PR #5522, opened 2024-11-26
Expose shard keys in telemetry

timvisee:
Expose shard keys for each shard in telemetry. This is necessary for resharding.

Rather than listing the key mapping, this lists the assigned key in each shard instead. I think that is a better interface to work with. Because of it I've copied the shard key into the replica set struct.

See `"key": "test"`:

```json
{
  "collections": {
    "number_of_collections": 1,
    "collections": [
      {
        "id": "benchmark",
        "init_time_ms": 800,
        "config": { ... },
        "shards": [
          {
            "id": 1,
            "key": "test",
            "local": { ... },
            "remote": [],
            "replicate_states": {
              "6469369128247073": "Active"
            }
          }
        ],
        "transfers": [],
        "resharding": []
      }
    ]
  }
}
```

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -11126,6 +11126,16 @@
             "format": "uint32",
             "minimum": 0
           },
+          "key": {
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/ShardKey"
+              },
+              {
+                "nullable": true
+              }
+            ]
+          },
           "local": {
             "anyOf": [
               {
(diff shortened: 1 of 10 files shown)


ffuugoo (approved):

agourlay:
> This is necessary for resharding.

Does it mean the resharding process queries the telemetry endpoint?
What if telemetry is disabled?

generall:
> What if telemetry is disabled?

we can't disable endpoint, only the data collection

timvisee:
> > This is necessary for resharding.
> 
> Does it mean the resharding process queries the telemetry endpoint? What if telemetry is disabled?

We discussed it in our daily, but for formality: CM already depends on telemetry as it fetches all the details about the peer/cluster state from it. It therefore makes sense the resharding key is exposed there as well. The alternative would be to make a separate request to cluster info along side it.

PR #5523, opened 2024-11-26
Add retrieving storage size to payload storage trait

agourlay:
This PR adds `get_storage_size_bytes` to the payload storage trait for us to leverage later on the serverless infra.

### DIFF
--- a/lib/segment/src/payload_storage/in_memory_payload_storage_impl.rs
+++ b/lib/segment/src/payload_storage/in_memory_payload_storage_impl.rs
@@ -91,6 +91,19 @@ impl PayloadStorage for InMemoryPayloadStorage {
     fn files(&self) -> Vec<PathBuf> {
         vec![]
     }
+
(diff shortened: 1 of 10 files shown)


lib/segment/src/payload_storage/in_memory_payload_storage_impl.rs
```diff
@@ -91,6 +91,19 @@ impl PayloadStorage for InMemoryPayloadStorage {
     fn files(&self) -> Vec<PathBuf> {
         vec![]
     }
+
+    fn get_storage_size_bytes(&self) -> OperationResult<usize> {
```
agourlay, line 95: I came up with a super simple implementation for the in-memory payload storage as it is only used in tests.
Feel free to suggest a better one if you want :)

timvisee: I was already writing a huge comment about performance until I noticed

> only used in tests

coszio (approved):

PR #5527, opened 2024-11-26
Points inconsistency during cancel of optimization

generall:
- test to demonstrate inconsistency failure during optimization


timvisee:
Implemented a fix for `apply_points` which also fixes this test in <https://github.com/qdrant/qdrant/pull/5528>

@generall has another patch for proxy segments pending.

timvisee (approved):

PR #5530, opened 2024-11-27
Fix path to new payload storage for optimized segments

agourlay:
Fix the path for the mmap payload storage in the segment builder.

Currently the storage is built at the level of `segments` and is not propagated properly.
We did not discover the issue earlier because we create the storage folder if it does not exist on load.

I added a simple making sure we do not add the folder at the wrong spot.

### DIFF
--- a/lib/segment/src/segment_constructor/segment_builder.rs
+++ b/lib/segment/src/segment_constructor/segment_builder.rs
@@ -54,13 +54,13 @@ pub struct SegmentBuilder {
     temp_dir: TempDir,
     indexed_fields: HashMap<PayloadKeyType, PayloadFieldSchema>,
 
-    // Payload key to deframent data to
+    // Payload key to defragment data to
     defragment_keys: Vec<PayloadKeyType>,
 }
 
 impl SegmentBuilder {
     pub fn new(
-        segment_path: &Path,
+        segments_path: &Path,
         temp_dir: &Path,
         segment_config: &SegmentConfig,
     ) -> OperationResult<Self> {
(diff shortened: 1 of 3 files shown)


lib/segment/src/segment_constructor/segment_builder.rs
```diff
@@ -54,13 +54,13 @@ pub struct SegmentBuilder {
...
     indexed_fields: HashMap<PayloadKeyType, PayloadFieldSchema>,
 
-    // Payload key to deframent data to
+    // Payload key to defragment data to
     defragment_keys: Vec<PayloadKeyType>,
 }
 
 impl SegmentBuilder {
     pub fn new(
-        segment_path: &Path,
```
agourlay, line 63: rename for clairty

lib/segment/tests/integration/segment_builder_test.rs
```diff
@@ -385,3 +387,54 @@ fn test_building_cancellation() {
...
+
+    let new_segment: Segment = builder.build(permit, &stopped).unwrap();
+    assert_eq!(
+        new_segment.segment_config.payload_storage_type,
+        PayloadStorageType::Mmap
+    );
+
+    let new_segment_count = segment_dir.path().read_dir().unwrap().count();
+
+    assert_eq!(new_segment_count, 2);
```
agourlay, line 439: failed previously with `3` given the `payload_storage` directory was misplaced

coszio (approved):
The importance of good naming 🙌

PR #5531, opened 2024-11-27
preserve the update version before doing move_if_exists in proxy

generall:


lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -153,11 +149,42 @@ impl ProxySegment {
...
+            // - Already moved version is less than what we have in current proxy -> overwrite
+            // - Already moved version is higher than the current one -> mark local as removed
+            // - Local point doesn't exist or already removed locally -> do nothing
+
+            let local_version = match (local_version_opt, deleted_points_guard.get(&point_id)) {
+                (Some(local_version), Some(&deleted_version)) => {
+                    if local_version > deleted_version {
+                        // If local version is higher than what we already moved,
+                        // We might want to overwrite it again
+                        local_version
```
generall, line 172: This is the code-path, which we didn't allow before

lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -443,13 +472,21 @@ impl SegmentEntry for ProxySegment {
             LockedSegment::Original(raw_segment) => {
                 let point_offset = raw_segment.read().get_internal_id(point_id);
                 if point_offset.is_some() {
-                    was_deleted = self.deleted_points.write().insert(point_id);
+                    was_deleted = self
+                        .deleted_points
+                        .write()
+                        .insert(point_id, op_num)
+                        .is_some();
```
timvisee, line 479: Condition is accidentally inverted.

```suggestion
                        .is_none();
```

`insert` returns `None` now if it's newly inserted. It did return `true` before if something was newly inserted. We only want this to be `true` for new insertions to only consider those as actual deletions.

lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -443,13 +472,21 @@ impl SegmentEntry for ProxySegment {
...
                 point_offset
             }
             LockedSegment::Proxy(proxy) => {
                 if proxy.read().has_point(point_id) {
-                    was_deleted = self.deleted_points.write().insert(point_id);
+                    was_deleted = self
+                        .deleted_points
+                        .write()
+                        .insert(point_id, op_num)
+                        .is_some();
```
timvisee, line 489:
```suggestion
                        .is_none();
```

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -459,8 +459,11 @@ pub trait SegmentOptimizer {
         // Second step - delete all the rest points with full write lock
         //
         // Use collection copy to prevent long time lock of `proxy_deleted_points`
-        let deleted_points_snapshot: Vec<PointIdType> =
-            proxy_deleted_points.read().iter().cloned().collect();
+        let deleted_points_snapshot: Vec<PointIdType> = proxy_deleted_points
+            .read()
+            .iter()
+            .map(|(point_id, _)| *point_id)
+            .collect();
```
timvisee, line 466:
```suggestion
        let deleted_points_snapshot: Vec<PointIdType> =
            proxy_deleted_points.read().keys().copied().collect();
```

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -625,8 +629,11 @@ pub trait SegmentOptimizer {
         // - save already removed points while avoiding long read locks
         // - exclude already removed points from post-optimization removing
         let already_remove_points = {
-            let mut all_removed_points: HashSet<_> =
-                proxy_deleted_points.read().iter().cloned().collect();
+            let mut all_removed_points: HashSet<_> = proxy_deleted_points
+                .read()
+                .iter()
+                .map(|(point_id, _)| *point_id)
+                .collect();
```
timvisee, line 636:
```suggestion
            let mut all_removed_points: HashSet<_> =
                proxy_deleted_points.read().keys().copied().collect();
```

timvisee (approved):
I simplified the match tree a bit with two flattened if-statements (60154427ea35903b52f2b2ecbdc05b98c2927d9a). Feel free to drop the commit if you like the previous better - it contains no logic change.

It also looked like two boolean conditions were accidentally reverted, I added a commit to patch this.

Other than that, the implementation looks good. The change makes a lot of sense. Great find!

PR #5532, opened 2024-11-27
Restructure sparse storages module

coszio:
Purely a restructuring PR. Now with the new mmap sparse vector storage, we have two implementations that should live in its own module.

Follows the same pattern as the dense vectors storages.


agourlay (approved):
thank you for keeping things tidy :saluting_face:

PR #5533, opened 2024-11-28
Fast loading of mmap sparse storage

coszio:
Addresses https://github.com/qdrant/qdrant/pull/5454#pullrequestreview-2465392680

We want mmap storages to not waste any extra time loading, so this PR refactors `MmapSparseVectorStorage` to get rid of loading and use `DynamicMmapFlags` for deleted flags

### DIFF
--- a/lib/blob_store/src/tracker.rs
+++ b/lib/blob_store/src/tracker.rs
@@ -293,7 +293,6 @@ impl Tracker {
             self.pending_updates
                 .insert(point_offset, PointerUpdate::Unset(pointer));
         }
-        self.next_pointer_offset = self.next_pointer_offset.max(point_offset + 1);
 
         pointer_opt
     }
--- a/lib/common/memory/src/mmap_type.rs
+++ b/lib/common/memory/src/mmap_type.rs
@@ -389,6 +389,38 @@ impl MmapBitSlice {
 
         Ok(())
     }
+
+    /// Reopens the mmap file with a larger size.
+    ///
+    /// # Arguments
+    /// - `path`: Path to the mmap file
+    /// - `total_capacity`: New total capacity of the bitslice, in amount of bits it should fit.
+    ///
+    /// # Safety
+    /// Marked unsafe because the path must be the same as the original mmap file.
+    pub unsafe fn extend(
+        &mut self,
+        path: &Path,
+        total_capacity: usize,
+        advice: AdviceSetting,
+        populate: bool,
+    ) -> Result<()> {
+        if total_capacity <= self.len() {
+            return Ok(());
+        }
+
+        let new_length = total_capacity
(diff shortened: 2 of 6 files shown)


lib/blob_store/src/tracker.rs
```diff
@@ -293,7 +293,6 @@ impl Tracker {
             self.pending_updates
                 .insert(point_offset, PointerUpdate::Unset(pointer));
         }
-        self.next_pointer_offset = self.next_pointer_offset.max(point_offset + 1);
```
coszio, line 296: This used to fix the missing persistence of the deleted flags, but now they get persisted on their own, and I believe it shouldn't be here anymore. 

The persistence of the last deleted id should be responsibility of the higher layer, not this one.

lib/common/memory/src/mmap_type.rs
```diff
@@ -389,6 +389,38 @@ impl MmapBitSlice {
...
+
+    /// Reopens the mmap file with a larger size.
+    ///
+    /// # Arguments
+    /// - `path`: Path to the mmap file
+    /// - `total_capacity`: New total capacity of the bitslice, in amount of bits it should fit.
+    ///
+    /// # Safety
+    /// Marked unsafe because the path must be the same as the original mmap file.
+    pub unsafe fn extend(
```
generall, line 401: I reacal we have similar functionality somewhere, which didn't require unsafe function

cc @xzfc

xzfc: https://github.com/qdrant/qdrant/blob/1beab3d8aeea53886816ba82615fa3ef5222af3d/lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs#L163-L170

coszio: TIL about `DynamicMmapFlags`, thanks @xzfc!

lib/blob_store/src/tracker.rs
agourlay: It is weird that we still increment it L286 above though in the `set` method.

Do we have a test showing that this is correct? :thinking:

coszio: I don't think we have a specific test for this, but the reasoning is:

If we insert a value, it might be a new id, so we need to advance the `next_pointer_offset`.

If we remove a value, since we don't shrink, we don't care if it is higher or lower than the `next_pointer_offset`.

generall:
I have some concerns about usage of `next_point_offset`, while the same information should be available in the storage itself. What happens if storage is flushed, while metadata is not?

coszio:
Status: drafting, while we merge #5587

coszio:
#5587 is now merged, so this one does not need to calculate total sparse size now ⚡ 

Another review pass? @generall

PR #5536, opened 2024-11-28
Return complete error for sparse vector validation

agourlay:
Always return the precise validation error for sparse vectors.

### DIFF
--- a/lib/api/src/conversions/vectors.rs
+++ b/lib/api/src/conversions/vectors.rs
@@ -464,12 +464,11 @@ impl TryFrom<grpc::Vector> for VectorInternal {
         // sparse vector
         if let Some(indices) = vector.indices {
             return Ok(VectorInternal::Sparse(
-                sparse::common::sparse_vector::SparseVector::new(indices.data, vector.data)
-                    .map_err(|e| {
-                        Status::invalid_argument(format!(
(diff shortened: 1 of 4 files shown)


lib/api/src/conversions/vectors.rs
```diff
@@ -464,12 +464,11 @@ impl TryFrom<grpc::Vector> for VectorInternal {
         // sparse vector
         if let Some(indices) = vector.indices {
             return Ok(VectorInternal::Sparse(
-                sparse::common::sparse_vector::SparseVector::new(indices.data, vector.data)
```
agourlay, line 467: the correct error pattern was already present here, in that diff I just removed the unnecessary path :)

tests/basic_sparse_grpc_test.sh
```diff
@@ -175,7 +175,7 @@ response=$(
     "limit": 3
   }' $QDRANT_HOST qdrant.Points/Search 2>&1
 )
-if [[ $response != *"Sparse indices does not match sparse vector conditions"* ]]; then
+if [[ $response != *"Sparse indices does not match sparse vector conditions: values: Validation error: must be the same length as indices [{}]"* ]]; then
```
timvisee, line 178: Ouch this is an ugly error. But I like that we expose more details now!

timvisee (approved):

PR #5537, opened 2024-11-28
Better debug print for vectors

generall:


timvisee (approved):
Much better! I like that it only takes the first four elements.

PR #5543, opened 2024-11-28
it is easy if you know where to look

generall:
What lessons should we learn from this bug?

- Do test coverage?
- Rely more on the pattern matching and force compiler to make sure all fields are used?
- Do not access fields of the structure directly?

### DIFF
--- a/lib/segment/src/segment_constructor/segment_builder.rs
+++ b/lib/segment/src/segment_constructor/segment_builder.rs
@@ -241,23 +241,22 @@ impl SegmentBuilder {
         for (segment_index, segment) in segments.iter().enumerate() {
             for external_id in segment.iter_points() {
                 let version = segment.point_version(external_id).unwrap_or(0);
+                let internal_id = segment.get_internal_id(external_id).unwrap();
                 merged_points
                     .entry(external_id)
                     .and_modify(|entry| {
                         if entry.version < version {
                             entry.segment_index = segment_index;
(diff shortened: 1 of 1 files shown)


lib/segment/src/segment_constructor/segment_builder.rs
```diff
@@ -241,23 +241,22 @@ impl SegmentBuilder {
...
             for external_id in segment.iter_points() {
                 let version = segment.point_version(external_id).unwrap_or(0);
+                let internal_id = segment.get_internal_id(external_id).unwrap();
                 merged_points
                     .entry(external_id)
                     .and_modify(|entry| {
                         if entry.version < version {
                             entry.segment_index = segment_index;
                             entry.version = version;
+                            entry.internal_id = internal_id;
```
timvisee, line 251: To clarify: here we mixed up internal IDs because we'd use the internal index of segment A in segment B. When assigning the new index we should also assign the new internal ID of that index.

timvisee (approved):
Hah, this is definitely a fun one. Fantastic find!

lib/segment/src/segment_constructor/segment_builder.rs
agourlay: This in place mutation field by field is very error prone indeed. 

Can we validate this invariant with tests now?
Are there other places where we do something similar?

PR #5546, opened 2024-11-28
Better `QueryInterface` untagged enum message

coszio:
Previously: 
`"Format error in JSON body: data did not match any variant of untagged enum QueryInterface at line 1 column 16"`

Now:
`"Format error in JSON body: Expected some form of vector, id, or a type of query at line 1 column 16"`

I did try to use `strum` crate to list the exact variants that are allowed, but `#[serde(expecting = "...")]` only allows string literals, so even when I could construct a `const &str` which was descriptive enough, I could not use it without reimplementing `Deserialize` for the type.

### DIFF
--- a/lib/api/src/rest/schema.rs
+++ b/lib/api/src/rest/schema.rs
@@ -435,6 +435,7 @@ pub struct QueryResponse {
 
 #[derive(Debug, Serialize, Deserialize, JsonSchema)]
 #[serde(untagged)]
+#[serde(expecting = "Expected some form of vector, id, or a type of query")]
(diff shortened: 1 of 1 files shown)


ffuugoo:
I had no idea `expecting` exists. Great find! 🤯🎉

ffuugoo (approved):

coszio:
I learned about it today!

timvisee:
> I had no idea `expecting` exists. Great find! 🤯🎉

Dang! That must be a new thing! :raised_hands:

PR #5550, opened 2024-11-29
Keep temporary segment if not empty

timvisee:
Be a bit more explicit on when we keep the temporary write segment of a segment proxy.

This doesn't solve a clear issue but I think the improvement is worth it and may save us in the future.

We try to be smart about whether to keep it. The idea is that if there's nothing in it, we can throw it away to not bloat the segment holder. This happens after optimization, if cancelling an optimization, and after proxying all segments for taking a snapshot.

Before we relied on the available point count. That count is not very reliable however (especially with a proxy), which may be problematic. It also doesn't consider deleted points which we may actually want to keep for the recovery procedure because they still have a version attached.

This PR makes the decision more explicit by only throwing the temporary segment away if it's completely empty.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -1174,10 +1174,11 @@ impl<'s> SegmentHolder {
 
         // Finalize temporary segment we proxied writes to
         // Append a temp segment to collection if it is not empty or there is no other appendable segment
-        let available_points = tmp_segment.get().read().available_point_count();
-        let has_appendable_segment = write_segments.has_appendable_segment();
-        if available_points > 0 || !has_appendable_segment {
-            log::trace!("Keeping temporary segment with {available_points} points");
+        if !write_segments.has_appendable_segment() && !tmp_segment.get().read().is_empty() {
+            log::trace!(
+                "Keeping temporary segment with {} points",
+                tmp_segment.get().read().available_point_count(),
+            );
             write_segments.add_new_locked(tmp_segment);
         } else {
             log::trace!("Dropping temporary segment with no changes");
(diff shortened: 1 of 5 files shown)


lib/segment/src/segment/entry.rs
```diff
@@ -375,6 +375,10 @@ impl SegmentEntry for Segment {
         self.id_tracker.borrow().internal_id(point_id).is_some()
     }
 
+    fn is_empty(&self) -> bool {
+        self.id_tracker.borrow().total_point_count() > 0
```
JojiiOfficial, line 379: Just want to check but this is intended, right? Naming feel as a bit off given the code.

timvisee: Yes.

Could you elaborate on why you think it's a bit off?

timvisee: :man_facepalming: Thanks!

generall: +1

timvisee: Patched.

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -1174,10 +1174,11 @@ impl<'s> SegmentHolder {
 
         // Finalize temporary segment we proxied writes to
         // Append a temp segment to collection if it is not empty or there is no other appendable segment
-        let available_points = tmp_segment.get().read().available_point_count();
-        let has_appendable_segment = write_segments.has_appendable_segment();
-        if available_points > 0 || !has_appendable_segment {
-            log::trace!("Keeping temporary segment with {available_points} points");
+        if !write_segments.has_appendable_segment() && !tmp_segment.get().read().is_empty() {
```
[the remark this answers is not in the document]
timvisee: Another :man_facepalming: Apparently I don't know what I'm doing.

Yes, it should be an OR condition in which case it is more likely to keep the segment.

Thanks once again!

JojiiOfficial (approved):

agourlay (approved):

PR #5552, opened 2024-11-29
Expose payload storage size in segment info

agourlay:
Adding the payload storage size at the segment info makes it visible in the telemetry.

Tested manually

e.g.

- empty segment

```json
  "num_vectors": 0,
  "num_points": 0,
  "num_indexed_vectors": 0,
  "num_deleted_vectors": 0,
  "vectors_size_bytes": 0,
  "payloads_size_bytes": 0,
  "ram_usage_bytes": 0,
  "disk_usage_bytes": 0,
```

- non empty

```json
  "num_vectors": 6,
  "num_points": 6,
  "num_indexed_vectors": 0,
  "num_deleted_vectors": 0,
  "vectors_size_bytes": 96,
  "payloads_size_bytes": 1183,
  "ram_usage_bytes": 0,
  "disk_usage_bytes": 0,
```

### DIFF
--- a/lib/segment/src/segment/entry.rs
+++ b/lib/segment/src/segment/entry.rs
@@ -491,14 +491,22 @@ impl SegmentEntry for Segment {
 
         let vectors_size_bytes = total_average_vectors_size_bytes * num_points;
 
+        // Unwrap and default to 0 here because the RocksDB storage is the only faillible one, and we will remove it eventually.
+        let payloads_size_bytes = self
+            .payload_storage
+            .borrow()
+            .get_storage_size_bytes()
+            .unwrap_or(0);
+
         SegmentInfo {
             segment_type: self.segment_type,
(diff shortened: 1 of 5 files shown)


lib/segment/src/segment/entry.rs
```diff
@@ -491,14 +491,22 @@ impl SegmentEntry for Segment {
 
         let vectors_size_bytes = total_average_vectors_size_bytes * num_points;
 
+        // Unwrap and default to 0 here because the RocksDB storage is the only faillible one, and we will remove it eventually.
+        let payloads_size_bytes = self
```
agourlay, line 495: the alternative is to change the entire call stack to the telemetry endpoint to use `CollectionResult` which is very noisy IMHO

JojiiOfficial (approved):

PR #5553, opened 2024-11-29
Fix missing flush on proxy segment if sharing write segment

timvisee:
Fix a problem where we could potentially miss flushing deletes on segments wrapped in a proxy.

Imagine the following events:
- flush proxy A
- reinsert point into proxy A
- flush proxy B

If proxy A and B share the same write segment, this write segment will be updated even though proxy B is yet to be flushed. The write segment affects the segment version of the proxy, so the version of proxy B is bumped even though it's still pending flush. We normally don't expect this to happen as we hold a read lock on proxy B.

Also, in our flush logic we normally acknowledge the highest version we've seen to the write ahead log.

If we add versions to the above actions, the problem becomes visible:
- flush proxy A (version 10)
- reinsert point into proxy A (version 11)
- flush proxy B (version 11)

We did not flush the new change having version 11 in proxy A because we flushed before the update came in. We do acknowledge version 11 in the WAL because it was the highest as reported by proxy B which does include the new version (through proxy A). If we crash now, the WAL won't fix this for us because it'll only start replaying from version 12 onward since version 11 was already acknowledged.

As far as I can see this can only cause us missing flushing deletes. Nor should it cause major issues because we already take point versions into account everywhere. Still I think it's better to prevent this scenario from happening to make our storage more robust.

The key problem here is that we can have multiple proxy segments sharing the same write segment. They are flushed at different times and individually release immediately after flushing. Because of it we can affect a segment we still have a read-lock on.

Note that it's fine to write into non-proxy segments that have just been flushed, even if other segments are still pending flush. The reason is that writing into these will not affect the segment version of the segments pending flush.

To fix the problem from happening I delay releasing the lock for all proxy segments until all segments have been flushed. This will prevent access to shared segments if some part is already flushed.

A potential downside can be that accepting a write takes longer if all segments are proxified. I don't see this as real issue though.

This one is hard to explain, let me know if I need to elaborate further.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -20,7 +20,9 @@ use segment::data_types::named_vectors::NamedVectors;
 use segment::entry::entry_point::SegmentEntry;
 use segment::segment::{Segment, SegmentVersion};
 use segment::segment_constructor::build_segment;
-use segment::types::{Payload, PointIdType, SegmentConfig, SeqNumberType, SnapshotFormat};
+use segment::types::{
+    Payload, PointIdType, SegmentConfig, SegmentType, SeqNumberType, SnapshotFormat,
+};
 
 use crate::collection::payload_index_schema::PayloadIndexSchema;
 use crate::collection_manager::holders::proxy_segment::ProxySegment;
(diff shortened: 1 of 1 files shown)


generall (approved):

ffuugoo (approved):
Tbh, I'm still not convinced this is a proper fix, but it seems safe enough (and easy to revert), so let's gooooo! 🫡

PR #5559, opened 2024-12-02
inference: simple connection pool

n0x29a:
Implements a simple connection pool to manage multiple HTTP clients for inference requests.

### DIFF
--- a/src/common/inference/config.rs
+++ b/src/common/inference/config.rs
@@ -6,15 +6,11 @@ pub struct InferenceConfig {
     #[serde(default = "default_inference_timeout")]
     pub timeout: u64,
     pub token: Option<String>,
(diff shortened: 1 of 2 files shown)


generall (approved):

agourlay:
Why writing an explicit of pool of `Client` instead of using the integrated connection pool mechanism from Reqwest with [pool_max_idle_per_host](https://docs.rs/reqwest/0.12.9/reqwest/struct.ClientBuilder.html#method.pool_max_idle_per_host)?

n0x29a:
since [reqwest already supports a connection pool internally](https://docs.rs/reqwest/latest/reqwest/struct.Client.html), we can close this pr 

```
The Client holds a connection pool internally, so it is advised that you create one and reuse it.
```

PR #5560, opened 2024-12-02
Enforce clippy match_wildcard_for_single_variants

agourlay:
This PR fixes and enforces the Clippy lint [clippy::match_wildcard_for_single_variants](https://rust-lang.github.io/rust-clippy/master/index.html#/wildcard%20matches%20only).

This only checks the usage of wildcard enum matches for a single variant.
There is another lint for multiple matches but it would be pretty large to fix.


timvisee (approved):
Good lint change. Thanks!

PR #5561, opened 2024-12-02
Enforce clippy needless_pass_by_ref_mut

agourlay:
This PR fixes and enforces clippy [needless_pass_by_ref_mut](https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut)

### DIFF
--- a/lib/segment/src/segment/segment_ops.rs
+++ b/lib/segment/src/segment/segment_ops.rs
@@ -72,6 +72,7 @@ impl Segment {
     /// # Warning
     ///
     /// Available for appendable segments only.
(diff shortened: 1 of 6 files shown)


lib/segment/src/segment/segment_ops.rs
```diff
@@ -72,6 +72,7 @@ impl Segment {
     /// # Warning
     ///
     /// Available for appendable segments only.
+    #[allow(clippy::needless_pass_by_ref_mut)] // ensure single access to AtomicRefCell vector_index
```
agourlay, line 75: the important change

timvisee: Nice. I was looking for cases like this, thank you for carefully reviewing.

PR #5573, opened 2024-12-03
In proxy segment, apply proxied point delete with its operation version

timvisee:
At the end of segment optimization, we apply all proxied point deletes to the optimized segment.

When applying these deletes we reused the current segment version on those operations before. The segment likely considers that version to already be persisted. Subsequent flushes will not write the deletes to disk because of it, the segment thinks there are no new changes.

I've changed this to apply the deletes with the operation version that caused the delete to happen. Those will bump the segment version and ensure proper flushing.

I've added a debug assertion to ensure the delete versions are always newer than the segment version.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -254,8 +254,14 @@ impl ProxySegment {
             let deleted_points = self.deleted_points.upgradable_read();
             if !deleted_points.is_empty() {
                 wrapped_segment.with_upgraded(|wrapped_segment| {
-                    for point_id in deleted_points.keys() {
-                        wrapped_segment.delete_point(op_num, *point_id)?;
+                    for (point_id, version) in deleted_points.iter() {
+                        // Delete points here with their operation version, that'll bump the optimized
+                        // segment version and will ensure we flush the new changes
+                        debug_assert!(
+                            *version >= op_num,
+                            "proxied point deletes should have newer version than segment",
+                        );
+                        wrapped_segment.delete_point(*version, *point_id)?;
                     }
                     OperationResult::Ok(())
                 })?;
@@ -266,6 +272,8 @@ impl ProxySegment {
                 // Points are still marked as deleted in two places, which is fine
             }
         }
+
+        let op_num = wrapped_segment.version();
 
         // Propagate deleted indexes
         // Ordering is important here and must match the flush function to prevent a deadlock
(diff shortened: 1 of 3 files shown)


lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -646,12 +646,17 @@ pub trait SegmentOptimizer {
...
+                // Delete points here with their operation version, that'll bump the optimized
+                // segment version and will ensure we flush the new changes
+                debug_assert!(
+                    optimized_segment_version >= version,
+                    "proxied point deletes should have newer version than segment",
+                );
+                optimized_segment.delete_point(version, point_id).unwrap();
             }
 
             for deleted_field_name in proxy_deleted_indexes.read().iter() {
```
timvisee, line 662: We also delete/create payload indices here. We probably need to use newer operation versions here as well.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -646,12 +646,17 @@ pub trait SegmentOptimizer {
...
+                // Delete points here with their operation version, that'll bump the optimized
+                // segment version and will ensure we flush the new changes
+                debug_assert!(
+                    version >= optimized_segment_version,
+                    "proxied point deletes should have newer version than segment",
+                );
+                optimized_segment.delete_point(version, point_id).unwrap();
             }
 
             for deleted_field_name in proxy_deleted_indexes.read().iter() {
```
timvisee, line 662: We also apply payload index changes here with the same segment version. We should apply the same change to those, using newer operation versions.

I'll do that in a separate PR as it'll require more significant changes, and it is not important for our current chaos testing efforts.

It'll already be fine with this change if there are any deletes.

generall (approved):

ffuugoo (approved):

ffuugoo (approved):

lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -463,7 +478,13 @@ impl SegmentEntry for ProxySegment {
                     was_deleted = self
                         .deleted_points
                         .write()
-                        .insert(point_id, op_num)
+                        .insert(
+                            point_id,
+                            ProxyDeletedPoint {
+                                local_version: op_num,
+                                operation_version: op_num,
+                            },
```
timvisee, line 486: For explicit delete operations I use `op_num` for both versions. It prevents moving the point into the write segment through another proxy, if it has a point version in between this local one and this operation.

generall (approved):

PR #5582, opened 2024-12-04
Rate limiting for shard operations

agourlay:
This PR enables rate limiting through strict mode for read and write requests.

Integration tests are present to validate the limit kicks in with the correct error message.

I validated the precision of the limit with an end to end example for 1000 search req/s.

```http
PUT collections/benchmark
{
  "vectors": {
    "size": 4,
    "distance": "Dot"
  },
  "strict_mode_config": {
    "enabled": true,
    "read_rate_limit_per_sec": 1000
  }
}
```

Shoot with [oha](https://github.com/hatoo/oha) with 10 concurrent workers for 1 minutes

```bash
oha -m POST  \
 -d "{ \"vector\": [0.2, 0.1, 0.9, 0.7], \"limit\": 4 }" \
 -T application/json \
 -A application/json  \
 -z 1m \
 -c 10 \
 http://127.0.0.1:6333/collections/benchmark/points/search
```

Expected results `1000 rps * 60s = 60k` valid responses per minute.

```
Status code distribution:
  [200] 60949 responses
  [429] 38885 responses
```

### DIFF
--- a/docs/grpc/docs.md
+++ b/docs/grpc/docs.md
@@ -1432,6 +1432,8 @@ Note: 1kB = 1 vector of size 256. |
 | search_max_oversampling | [float](#float) | optional |  |
 | upsert_max_batchsize | [uint64](#uint64) | optional |  |
 | max_collection_vector_size_bytes | [uint64](#uint64) | optional |  |
+| read_rate_limit_per_sec | [uint32](#uint32) | optional |  |
+| write_rate_limit_per_sec | [uint32](#uint32) | optional |  |
 
 
 
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -17,7 +17,7 @@ use std::time::{Duration, Instant};
 
 use arc_swap::ArcSwap;
 use common::cpu::CpuBudget;
-use common::rate_limiting::RateLimiter;
+use common::rate_limiting::{Rate, RateLimiter};
 use common::types::TelemetryDetail;
 use common::{panic, tar_ext};
 use indicatif::{ProgressBar, ProgressStyle};
@@ -96,8 +96,8 @@ pub struct LocalShard {
     update_runtime: Handle,
     pub(super) search_runtime: Handle,
     disk_usage_watcher: DiskUsageWatcher,
-    read_rate_limiter: Option<ParkingMutex<RateLimiter>>,
-    write_rate_limiter: Option<ParkingMutex<RateLimiter>>,
+    read_rate_limiter: ParkingMutex<Option<RateLimiter>>,
+    write_rate_limiter: ParkingMutex<Option<RateLimiter>>,
 }
 
 /// Shard holds information about segments and WAL.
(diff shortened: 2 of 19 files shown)


timvisee:
> [200] 60949 responses

Nice! I think the ~1000 extra are because we start with a capacity of 1000 and are replenishing in real time.

lib/segment/src/types.rs
```diff
@@ -707,6 +707,14 @@ pub struct StrictModeConfig {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     pub upsert_max_batchsize: Option<usize>,
+
+    /// Max number of read operations per second
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub read_rate_limit_per_sec: Option<usize>,
+
+    /// Max number of write operations per second
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub write_rate_limit_per_sec: Option<usize>,
```
timvisee, line 717: Not sure about this one, but I think it may be useful to be a bit more explicit on this one:

```suggestion
    /// Max number of read operations per second per shard per peer
    #[serde(skip_serializing_if = "Option::is_none")]
    pub read_rate_limit_per_sec: Option<usize>,

    /// Max number of write operations per second per shard per peer
    #[serde(skip_serializing_if = "Option::is_none")]
    pub write_rate_limit_per_sec: Option<usize>,
```

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -741,6 +757,28 @@ impl LocalShard {
         Ok(())
     }
 
+    /// Apply shard's strict mode configuration update
+    /// - Update read and write rate limiters
+    pub async fn on_strict_mode_config_update(&self) {
+        let config = self.collection_config.read().await;
+
+        if let Some(strict_mode_config) = &config.strict_mode_config {
```
timvisee, line 765: Would it be worth it to create a function that builds and returns the new `RateLimiter`s. We'd be able to use it in the constructor above, and here, sharing code.

tests/openapi/test_strictmode.py
```diff
@@ -461,3 +461,83 @@ def search_request():
...
+                ]
+            }
+        )
+        if not response.ok:
+            assert response.status_code == 429
+            assert "Rate limiting exceeded: Write rate limit exceeded, retry later" in response.json()['status']['error']
+            failed = True
+            break
+
+    assert failed, "Rate limiting did not work"
```
timvisee, line 543: Shall we make it a bit more strict and make `failed` a counter? And assert we've failed at least 5 (50%) times?

timvisee (approved):
Thanks! :pray: 

Left some suggestions. Don't take all of them too serious :smile:

JojiiOfficial (approved):
Nothing to add to @timvisee's points

docs/grpc/docs.md
```diff
@@ -1432,6 +1432,8 @@ Note: 1kB = 1 vector of size 256. |
 | search_max_oversampling | [float](#float) | optional |  |
 | upsert_max_batchsize | [uint64](#uint64) | optional |  |
 | max_collection_vector_size_bytes | [uint64](#uint64) | optional |  |
+| read_rate_limit_per_sec | [uint32](#uint32) | optional |  |
+| write_rate_limit_per_sec | [uint32](#uint32) | optional |  |
```
generall, line 1436: rate limit per second won't allow us to configure bursts. Could we switch to requests per minute?

agourlay: Internally the rate limiter computes the number of requests allowed per second.

```rust
let tokens_per_sec = rate.requests_num() as f64 / rate.period().as_secs_f64();
```

Having an input of `10 req/s` is equivalent to `600 req/m`.

Those tokens are replenished in real time based on the elapsed time since the last check.
We do start at full capacity to allow burst on start.

Here is the impl:

```rust
pub fn check(&mut self) -> bool {
        let now = Instant::now();
        let elapsed = now.duration_since(self.last_check);
        self.last_check = now;

        // Refill tokens based on elapsed time.
        self.tokens += self.tokens_per_sec * elapsed.as_secs_f64();
        if self.tokens > self.capacity as f64 {
            self.tokens = self.capacity as f64;
        }

        if self.tokens >= 1.0 {
            self.tokens -= 1.0; // Consume one token.
            true // Request allowed.
        } else {
            false // Request denied.
        }
    }
```

Which kind of burst scenario are you concerned about?

generall (approved):
ToDO in other PR: switch to per-minute

docs/grpc/docs.md
agourlay: Discussed with Andrey regarding his concerns, he wants to be able to have an actual burst capacity of one minute worth of tokens and not one second.

I will merge this PR as it is because it is tidy and has been reviewed.
Then propose the adjusted implementation in a new PR for clarity.

lib/segment/src/types.rs
timvisee: On another thought `per replica` may be shorter to `per shard per peer`

agourlay:
Here is the followup https://github.com/qdrant/qdrant/pull/5597

PR #5585, opened 2024-12-04
Fix point deduplication

timvisee:
Fix a bug in point deduplication that caused mishandling of point versions.

The 'last point version' was not properly cleared when moving to the next point ID, meaning irrelevant point versions from the previous point leaked into the current one. Because of this it was possible we deleted newer point versions rather than old ones.

I've added two unit tests:
- the first tests this specific problem
- the second inserts random points into segments and asserts we always keep the highest point version

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -1377,6 +1377,7 @@ mod tests {
     use std::fs::File;
     use std::str::FromStr;
 
+    use rand::Rng;
     use segment::data_types::vectors::VectorInternal;
     use segment::json_path::JsonPath;
     use segment::segment_constructor::simple_segment_constructor::build_simple_segment;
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -1364,6 +1362,7 @@ impl<'s> SegmentHolder {
             } else {
                 last_point_id_opt = Some(point_id);
                 last_segment_id_opt = Some(segment_id);
+                last_point_version_opt = None;
```
timvisee, line 1365: The fix: properly clear point version when moving to the next point ID.

Thanks @generall for catching this one!

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -1688,6 +1688,168 @@ mod tests {
...
+        assert_eq!(
+            holder
+                .get(sid2)
+                .unwrap()
+                .get()
+                .read()
+                .available_point_count(),
+            2,
+        );
+    }
```
timvisee, line 1770: Test covering the bug we fixed.

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -1688,6 +1688,168 @@ mod tests {
...
+                found_versions.len(),
+                1,
+                "point version must be maximum known version",
+            );
+            assert_eq!(
+                found_versions[0], max,
+                "point version must be maximum known version",
+            );
+        }
+    }
```
timvisee, line 1851: Test using a thousand random points to assert correct behavior.

IvanPleshkov (approved):

PR #5587, opened 2024-12-04
Prefer vector index for `size_of_available_vectors_in_bytes`

coszio:
In #5533 we realized that calculating `total_sparse_vectors_size` on load was inevitable, but we don't want it, reason is:
- to compute it on load, we need to scan the whole storage, which we want to avoid for a mmap storage.
- if we put it in a mmapped file, replayed operations are not idempotent, which leads to a counter which we can't trust.

To solve this, we add `VectorIndex::size_of_searchable_vectors_in_bytes` as a primary source for `VectorStorage::size_of_available_vectors_in_bytes`. All indices have access to the vector storage, so it is easy to use the storage as a fallback if the index does not know it.

Regarding the precision of making this number come from the sparse index, a mutable segment will return an up-to-date number. For the immutable counterpart, we will rely on the vacuum optimizer, but it should still be representative of the work it takes to search through the all the vectors.

This will allow us to load the mmap sparse vector storage quicker.

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -163,7 +163,6 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
                 let available_vectors = vector_storage.available_vector_count();
                 let full_scan_threshold = vector_storage
                     .size_of_available_vectors_in_bytes()
-                    .expect("Dense vector storages always know their size")
                     .checked_div(available_vectors)
                     .and_then(|avg_vector_size| {
                         hnsw_config
@@ -268,7 +267,6 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
 
         let full_scan_threshold = vector_storage
             .size_of_available_vectors_in_bytes()
-            .expect("Dense vector storages always know their size")
             .checked_div(total_vector_count)
             .and_then(|avg_vector_size| {
                 hnsw_config
@@ -1328,9 +1326,10 @@ impl<TGraphLinks: GraphLinks> VectorIndex for HNSWIndex<TGraphLinks> {
             .unwrap_or_else(|| self.graph.num_points())
     }
 
-    fn size_of_searchable_vectors_in_bytes(&self) -> Option<usize> {
-        // get the size from the vector storage
-        None
+    fn size_of_searchable_vectors_in_bytes(&self) -> usize {
+        self.vector_storage
+            .borrow()
+            .size_of_available_vectors_in_bytes()
     }
 
     fn update_vector(
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -163,6 +163,7 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
                 let available_vectors = vector_storage.available_vector_count();
                 let full_scan_threshold = vector_storage
                     .size_of_available_vectors_in_bytes()
+                    .expect("Dense vector storages always know their size")
                     .checked_div(available_vectors)
                     .and_then(|avg_vector_size| {
                         hnsw_config
@@ -267,6 +268,7 @@ impl<TGraphLinks: GraphLinks> HNSWIndex<TGraphLinks> {
 
         let full_scan_threshold = vector_storage
             .size_of_available_vectors_in_bytes()
+            .expect("Dense vector storages always know their size")
             .checked_div(total_vector_count)
             .and_then(|avg_vector_size| {
                 hnsw_config
@@ -1324,6 +1326,11 @@ impl<TGraphLinks: GraphLinks> VectorIndex for HNSWIndex<TGraphLinks> {
             .indexed_vector_count
             // If indexed vector count is unknown, fall back to number of points
             .unwrap_or_else(|| self.graph.num_points())
+    }
+
+    fn size_of_searchable_vectors_in_bytes(&self) -> Option<usize> {
+        // get the size from the vector storage
+        None
     }
 
     fn update_vector(
--- a/lib/segment/src/segment/mod.rs
+++ b/lib/segment/src/segment/mod.rs
@@ -110,10 +110,13 @@ impl VectorData {
         storage: &VectorStorageEnum,
         index: &VectorIndexEnum,
     ) -> usize {
-        storage
+        let size = storage
             .size_of_available_vectors_in_bytes()
-            .or_else(|| index.size_of_searchable_vectors_in_bytes())
-            .unwrap_or(0)
+            .or_else(|| index.size_of_searchable_vectors_in_bytes());
+
+        debug_assert!(size.is_some());
+
+        size.unwrap_or(0)
     }
 }
 
--- a/lib/segment/src/segment/mod.rs
+++ b/lib/segment/src/segment/mod.rs
@@ -24,11 +24,11 @@ use rocksdb::DB;
 use crate::common::operation_error::{OperationResult, SegmentFailedState};
 use crate::id_tracker::IdTrackerSS;
 use crate::index::struct_payload_index::StructPayloadIndex;
-use crate::index::{VectorIndex, VectorIndexEnum};
+use crate::index::VectorIndexEnum;
 use crate::payload_storage::payload_storage_enum::PayloadStorageEnum;
 use crate::types::{SegmentConfig, SegmentType, SeqNumberType, VectorName};
 use crate::vector_storage::quantized::quantized_vectors::QuantizedVectors;
-use crate::vector_storage::{VectorStorage, VectorStorageEnum};
+use crate::vector_storage::VectorStorageEnum;
 
 pub const SEGMENT_STATE_FILE: &str = "segment.json";
 
@@ -104,19 +104,6 @@ impl VectorData {
         };
 
         index_task.into_iter().chain(storage_task)
-    }
-
-    pub fn size_of_available_vectors_in_bytes(
-        storage: &VectorStorageEnum,
-        index: &VectorIndexEnum,
-    ) -> usize {
-        let size = storage
-            .size_of_available_vectors_in_bytes()
-            .or_else(|| index.size_of_searchable_vectors_in_bytes());
-
-        debug_assert!(size.is_some());
-
-        size.unwrap_or(0)
     }
 }
 
--- a/lib/segment/src/segment/mod.rs
+++ b/lib/segment/src/segment/mod.rs
@@ -24,11 +24,11 @@ use rocksdb::DB;
 use crate::common::operation_error::{OperationResult, SegmentFailedState};
 use crate::id_tracker::IdTrackerSS;
 use crate::index::struct_payload_index::StructPayloadIndex;
-use crate::index::VectorIndexEnum;
+use crate::index::{VectorIndex, VectorIndexEnum};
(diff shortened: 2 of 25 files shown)


lib/segment/src/segment/mod.rs
```diff
@@ -105,6 +105,16 @@ impl VectorData {
...
+
+    pub fn size_of_available_vectors_in_bytes(
+        storage: &VectorStorageEnum,
+        index: &VectorIndexEnum,
+    ) -> usize {
+        storage
+            .size_of_available_vectors_in_bytes()
+            .or_else(|| index.size_of_searchable_vectors_in_bytes())
+            .unwrap_or(0)
+    }
```
coszio, line 117: This is where the fallback is done

lib/segment/src/index/plain_payload_index.rs
```diff
@@ -249,10 +249,14 @@ impl PlainIndex {
         filter: Option<&Filter>,
     ) -> bool {
         let vector_storage = self.vector_storage.borrow();
+        let Some(size_of_available_vectors) = vector_storage.size_of_available_vectors_in_bytes()
+        else {
+            // In case this gets called with sparse vectors, which are always indexed.
```
generall, line 254: It seems like an abstraction leak. There are no guarantees in the interface of the vector storage, that if vector size is none => segment is always indexed. If we can have another type of vectors in future, which doesn't satisfy this assumption. If we assume that this situation should never happen in a Plain segment, it should be debug assertion

lib/segment/src/segment/mod.rs
```diff
@@ -105,6 +105,16 @@ impl VectorData {
...
     }
+
+    pub fn size_of_available_vectors_in_bytes(
+        storage: &VectorStorageEnum,
+        index: &VectorIndexEnum,
+    ) -> usize {
+        storage
+            .size_of_available_vectors_in_bytes()
+            .or_else(|| index.size_of_searchable_vectors_in_bytes())
+            .unwrap_or(0)
```
generall, line 116: Debug assert that at least one of those should have the value?

lib/sparse/src/index/inverted_index/inverted_index_compressed_mmap.rs
```diff
@@ -39,8 +39,12 @@ impl StorageVersion for Version {
...
 #[derive(Debug, Default, Clone, Serialize, Deserialize)]
 pub struct InvertedIndexFileHeader {
-    pub posting_count: usize, // number of posting lists
-    pub vector_count: usize,  // number of unique vectors indexed
+    /// Number of posting lists
+    pub posting_count: usize,
+    /// Number of unique vectors indexed
+    pub vector_count: usize,
+    /// Total size of all searchable sparse vectors in bytes
+    pub total_sparse_size: usize,
```
generall, line 47: We can't just add a required field to serializable structure. This will break backward compatibility

lib/sparse/src/index/inverted_index/inverted_index_mmap.rs
```diff
@@ -35,8 +35,12 @@ impl StorageVersion for Version {
...
 #[derive(Debug, Default, Clone, Serialize, Deserialize)]
 pub struct InvertedIndexFileHeader {
-    pub posting_count: usize, // number oof posting lists
-    pub vector_count: usize,  // number of unique vectors indexed
+    /// Number of posting lists
+    pub posting_count: usize,
+    /// Number of unique vectors indexed
+    pub vector_count: usize,
+    /// Total size of all searchable sparse vectors in bytes
+    pub total_sparse_size: usize,
```
generall, line 43: same

lib/sparse/src/index/inverted_index/inverted_index_compressed_immutable_ram.rs
```diff
@@ -101,13 +103,18 @@ impl<W: Weight> InvertedIndex for InvertedIndexCompressedImmutableRam<W> {
         Ok(InvertedIndexCompressedImmutableRam {
             postings,
             vector_count: ram_index.vector_count,
+            total_sparse_size: ram_index.total_sparse_size,
```
generall, line 106: Compressed index may apply quantization, which should be reflected in index size. So I propose to compute just number of elements in mutable ram index, and adjust it according to `size_of::<W>()` on the getter implementation.

generall (changes requested):
See comments above

lib/sparse/src/index/inverted_index/inverted_index_compressed_mmap.rs
agourlay: I am curious, why are the data compatibility tests on CI not failing already for this PR?

coszio: I just checked, the inverted index that is currently on the compatibility snapshots is pre 0.2.0 (compressed index), so it detects the version mismatch and just rebuilds from scratch. This is why it doesn't fail. 

However if we try to load a 0.2.0 index, this new field should make it fail.

lib/segment/src/index/plain_payload_index.rs
coszio: Added a debug assertion

lib/sparse/src/index/inverted_index/inverted_index_mmap.rs
coszio: We can ditch this implementation in a separate PR

lib/sparse/src/index/inverted_index/inverted_index_compressed_immutable_ram.rs
coszio: I just found out there is a handy `CompressedPostingListStoreSize`. I used it in the new commits.

Shout-out to @xzfc 🙌

lib/segment/src/segment/mod.rs
```diff
@@ -105,6 +105,19 @@ impl VectorData {
...
+        storage: &VectorStorageEnum,
+        index: &VectorIndexEnum,
+    ) -> usize {
+        let size = storage
+            .size_of_available_vectors_in_bytes()
+            .or_else(|| index.size_of_searchable_vectors_in_bytes());
+
+        debug_assert!(size.is_some());
+
+        size.unwrap_or(0)
```
agourlay, line 119: If we were to go with this design, we would need a log warning here to know something is wrong

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -1023,6 +1025,11 @@ impl<TGraphLinks: GraphLinks> VectorIndex for HNSWIndex<TGraphLinks> {
             .unwrap_or_else(|| self.graph.num_points())
     }
 
+    fn size_of_searchable_vectors_in_bytes(&self) -> Option<usize> {
+        // get the size from the vector storage
+        None
```
agourlay, line 1030: Can the dense vector dense do the calculation?
Like call the function from dense storage which we know is defined?

lib/segment/src/index/plain_payload_index.rs
```diff
@@ -375,6 +383,11 @@ impl VectorIndex for PlainIndex {
         0
     }
 
+    fn size_of_searchable_vectors_in_bytes(&self) -> Option<usize> {
+        // get the size from the vector storage
+        None
```
agourlay, line 388: having implementation returning None makes it very confusing the juggle with the fallback

lib/segment/src/vector_storage/vector_storage_base.rs
```diff
@@ -352,6 +352,58 @@ impl VectorStorageEnum {
...
+            }
+            VectorStorageEnum::DenseAppendableInRamHalf(v) => {
+                v.size_of_available_vectors_in_bytes()
+            }
+            VectorStorageEnum::SparseSimple(v) => v.size_of_available_vectors_in_bytes(),
+            VectorStorageEnum::SparseMmap(_v) => {
+                unreachable!(
+                    "Mmap sparse storage does not know its total size, get from index instead"
+                )
+            }
```
coszio, line 383: Update: we now only make the exception here. For the places where we need the size of sparse vectors, we can use `index.size_of_searchable_vectors_in_bytes()`.

`size_of_available_vectors_in_bytes()` has been moved from `VectorStorage` trait to `DenseVectorStorage` and `MultiVectorStorage` traits

lib/segment/src/index/plain_payload_index.rs
coszio: https://github.com/qdrant/qdrant/pull/5587/commits/7733772afafa8195dd85dedd84db843af5857685 :D

lib/segment/src/segment/mod.rs
coszio: Ditched this thang

lib/segment/src/vector_storage/vector_storage_base.rs
agourlay: looks like CI enters this unreachable block

https://github.com/qdrant/qdrant/actions/runs/12375987460/job/34542060519?pr=5587

coszio: oopsie, fixed

lib/sparse/src/index/inverted_index/inverted_index_compressed_immutable_ram.rs
```diff
@@ -98,16 +100,24 @@ impl<W: Weight> InvertedIndex for InvertedIndexCompressedImmutableRam<W> {
             }
             postings.push(new_posting_list.build());
         }
+
+        let total_sparse_size = postings.iter().map(|p| p.view().store_size().total).sum();
```
agourlay, line 104: Are we sure this is not slow?
Otherwise we moved the problem from storage to index :)

lib/sparse/src/index/inverted_index/inverted_index_immutable_ram.rs
```diff
@@ -12,6 +12,7 @@ use crate::index::posting_list::{PostingList, PostingListIterator};
 
 /// A wrapper around [`InvertedIndexRam`].
 /// Will be replaced with the new compressed implementation eventually.
+// TODO: Remove this inverted index implementation, it is no longer used
```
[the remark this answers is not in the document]
coszio: yes, we can delay this indefinitely :D

lib/sparse/src/index/inverted_index/inverted_index_compressed_mmap.rs
```diff
@@ -280,12 +297,26 @@ impl<W: Weight> InvertedIndexCompressedMmap<W> {
...
+            index.file_header.total_sparse_size = Some(index.calculate_total_sparse_size());
+            atomic_save_json(&config_file_path, &index.file_header)?;
+        }
+
+        Ok(index)
+    }
+
+    fn calculate_total_sparse_size(&self) -> usize {
+        (0..self.file_header.posting_count as DimId)
+            .filter_map(|id| self.get(&id).map(|posting| posting.store_size().total))
```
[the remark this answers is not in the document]
coszio: same as https://github.com/qdrant/qdrant/pull/5587#discussion_r1890467418

agourlay (approved):

lib/sparse/src/index/inverted_index/inverted_index_compressed_mmap.rs
```diff
@@ -280,12 +297,26 @@ impl<W: Weight> InvertedIndexCompressedMmap<W> {
...
+            .sum()
```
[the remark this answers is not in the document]
coszio: isn't it the same thing in the end? It still needs to deserialize the posting

PR #5588, opened 2024-12-05
Strict mode for payload storage

agourlay:
Validate the payload storage size when using the strict mode the same way it was done for `max_collection_vector_size_bytes`.

### DIFF
--- a/lib/api/src/grpc/proto/collections.proto
+++ b/lib/api/src/grpc/proto/collections.proto
@@ -324,6 +324,7 @@ message StrictModeConfig {
   optional uint64 max_collection_vector_size_bytes  = 10;
   optional uint32 read_rate_limit = 11; // Max number of read operations per minute per replica
   optional uint32 write_rate_limit = 12; // Max number of write operations per minute per replica
+  optional uint64 max_collection_payload_size_bytes  = 13;
 }
 
 message CreateCollection {
--- a/lib/api/src/grpc/proto/collections.proto
+++ b/lib/api/src/grpc/proto/collections.proto
@@ -324,7 +324,7 @@ message StrictModeConfig {
   optional uint64 max_collection_vector_size_bytes  = 10;
   optional uint32 read_rate_limit = 11; // Max number of read operations per minute per replica
   optional uint32 write_rate_limit = 12; // Max number of write operations per minute per replica
-  optional uint64 max_collection_payload_size_bytes  = 13;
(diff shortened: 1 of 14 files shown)


tests/openapi/test_strictmode.py
```diff
@@ -642,6 +644,134 @@ def test_strict_mode_read_rate_limiting(collection_name):
...
+            continue
+        assert "Max payload storage size" in failed_upsert.json()['status']['error']
+        assert not failed_upsert.ok
+        return
+
+    assert False, "Upserting should have failed but didn't"
+
+
+def test_strict_mode_max_collection_payload_size_upsert_batch(collection_name):
+    # TODO: Update this test when payload rocksDB storage has been replaced by mmap payload storage!
```
agourlay, line 710: I am trying to improve the test here for RocksDB but I also experience delays in the updating of  the `payloads_size_bytes` value.
Checking how we can mitigate it :thinking:

agourlay: I sadly can't find a way around it :(

I vote for going with the current test and migrate it once we have the mmap payload storage available in the API.

lib/api/src/grpc/proto/collections.proto
```diff
@@ -324,6 +324,7 @@ message StrictModeConfig {
   optional uint64 max_collection_vector_size_bytes  = 10;
   optional uint32 read_rate_limit = 11; // Max number of read operations per minute per replica
   optional uint32 write_rate_limit = 12; // Max number of write operations per minute per replica
+  optional uint64 max_collection_payload_size_bytes  = 13;
```
timvisee, line 327:
```suggestion
  optional uint64 max_collection_payload_size_bytes = 13;
```

tests/openapi/test_strictmode.py
timvisee, line 1: There are some `TODO: remove`'s in this file. Do we want to remove those now?

JojiiOfficial: Those will be removed when the default payload storage implementation has been migrated to mmap. They are workarounds to make the tests work with rocksdb.

JojiiOfficial: Done: #5669

agourlay: Thanks for pushing the review to the end line :+1: 
Merging now!

PR #5589, opened 2024-12-05
Pass inference API token from request to inference service

n0x29a:
This change enables dynamic token propagation from API requests to the inference service. 
Key changes:

- Add InferenceToken type to hold API tokens from requests
- Modify auth middleware to extract and propagate tokens from request headers 
- Update inference service to accept tokens per-request (when possible)
- Remove requirement for token in config (now optional)
- Thread token through query and update operations
- Fall back to config token if request token is not provided

The main goal is to allow per-request inference tracking with the user specific token rather than using a single global token. 

Technical details:
- Inference token is extracted during auth and stored in request extensions
- Token is propagated through all operations that need inference (queries, updates, etc)
- InferenceService.infer() now accepts a token parameter
- Default to config token if no request token is provided

### DIFF
--- a/src/actix/api/query_api.rs
+++ b/src/actix/api/query_api.rs
@@ -209,8 +209,7 @@ async fn query_points_groups(
             Some(shard_keys) => shard_keys.into(),
         };
         let query_group_request =
-            convert_query_groups_request_from_rest(search_group_request, inference_token.clone())
-                .await?;
+            convert_query_groups_request_from_rest(search_group_request, inference_token).await?;
 
         do_query_point_groups(
             dispatcher.toc(&access, &pass),
--- a/src/actix/api/query_api.rs
+++ b/src/actix/api/query_api.rs
@@ -1,4 +1,4 @@
-use actix_web::{post, web, HttpMessage, HttpRequest, Responder};
+use actix_web::{post, web, Responder};
 use actix_web_validator::{Json, Path, Query};
 use api::rest::{QueryGroupsRequest, QueryRequest, QueryRequestBatch, QueryResponse};
 use collection::operations::shard_selector_internal::ShardSelectorInternal;
@@ -14,10 +14,10 @@ use super::read_params::ReadParams;
 use super::CollectionPath;
 use crate::actix::auth::ActixAccess;
 use crate::actix::helpers::{self, get_request_hardware_counter, process_response_error};
-use crate::common::auth::ApiKey;
 use crate::common::inference::query_requests_rest::{
     convert_query_groups_request_from_rest, convert_query_request_from_rest,
 };
+use crate::common::inference::InferenceToken;
 use crate::common::points::do_query_point_groups;
 use crate::settings::ServiceConfig;
 
@@ -29,7 +29,7 @@ async fn query_points(
     params: Query<ReadParams>,
     service_config: web::Data<ServiceConfig>,
     ActixAccess(access): ActixAccess,
-    req: HttpRequest,
+    inference_token: InferenceToken,
 ) -> impl Responder {
     let QueryRequest {
         internal: query_request,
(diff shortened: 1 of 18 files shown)


src/common/inference/mod.rs
```diff
@@ -6,3 +12,41 @@ pub mod query_requests_grpc;
 pub mod query_requests_rest;
 pub mod service;
 pub mod update_requests;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct InferenceToken(pub Option<String>);
```
agourlay, line 17: Why did you decide for this design instead of having an `Option<InferenceToken>` instead?

n0x29a: We discussed with @generall that it could be an option. The reason is to simplify the code. If it's not the best approach, share your ideas to make it better.

src/actix/api/query_api.rs
```diff
@@ -59,9 +61,9 @@ async fn query_points(
         Some(shard_keys) => shard_keys.into(),
     };
     let hw_measurement_acc = request_hw_counter.get_counter();
-
     let result = async move {
-        let request = convert_query_request_from_rest(query_request).await?;
+        let request =
+            convert_query_request_from_rest(query_request, inference_token.clone()).await?;
```
agourlay, line 66: I do not think this clone is necessary :)

src/actix/api/query_api.rs
```diff
@@ -129,7 +132,8 @@ async fn query_points_batch(
                 shard_key,
             } = request;
 
-            let request = convert_query_request_from_rest(internal).await?;
+            let request =
+                convert_query_request_from_rest(internal, inference_token.clone()).await?;
```
agourlay, line 136: same here about clone

src/actix/api/query_api.rs
```diff
@@ -204,9 +209,9 @@ async fn query_points_groups(
             None => ShardSelectorInternal::All,
             Some(shard_keys) => shard_keys.into(),
         };
-
         let query_group_request =
-            convert_query_groups_request_from_rest(search_group_request).await?;
+            convert_query_groups_request_from_rest(search_group_request, inference_token.clone())
```
agourlay, line 213: +1

src/common/inference/service.rs
```diff
@@ -138,12 +139,6 @@ impl InferenceService {
     pub fn init_global(config: InferenceConfig) -> Result<(), StorageError> {
         let mut inference_service = INFERENCE_SERVICE.write();
 
-        if config.token.is_none() {
```
agourlay, line 141: So now we rely solely on this code for the token extraction?

```rust
  let token = inference_token
            .0
            .unwrap_or_else(|| self.config.token.clone().unwrap_or_default());
```

Does the inference service fail with a 4xx error with an empty token?

src/common/inference/mod.rs
```diff
@@ -6,3 +12,41 @@ pub mod query_requests_grpc;
...
+
+impl InferenceToken {
+    pub fn new(key: String) -> Self {
```
agourlay, line 20: Is there nothing to validate about the token String itself?

src/tonic/auth.rs
```diff
@@ -29,6 +30,8 @@ async fn check(auth_keys: Arc<AuthKeys>, mut req: Request) -> Result<Request, St
         })?;
 
     let previous = req.extensions_mut().insert::<Access>(access);
+    req.extensions_mut()
+        .insert(InferenceToken::new(api_key.to_string()));
```
agourlay, line 34: Maybe we can `assert_debug` that the insertion returns None?
As far I understand, there should never be already an inference token on the request.
WDYT?

src/common/inference/service.rs
n0x29a: yes, the inference service fail with a 4xx error with an empty token. it's a change for upcoming release.

src/tonic/auth.rs
n0x29a: The inference token should be updated per request, because requests could land to the inference service from different instances. Does it make sense?

timvisee: In that case, wouldn't each request start with a clean slate? The state we define here is not shared across requests, or is it?

Arnaud suggests a debug assertion as sanity check to ensure we do indeed start with a clean slate for every request. That's a valid suggestion to me, as accidentally mixing tokens across requests will be a nightmare.

src/common/inference/mod.rs
timvisee: Do we want to allow an empty string?

timvisee (approved):
Please wait for at least one other approval before merging :smile:

agourlay (approved):

PR #5590, opened 2024-12-05
Improve point deduplication loop

timvisee:
The logic for finding point duplicates was quite convoluted and was hard to understand. It even hid a significant bug for a long time (<https://github.com/qdrant/qdrant/pull/5585>)!

In this PR I attempt to simplify the loop to replace the error prone behavior we had before.

Note: this still needs a benchmark!

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -1285,8 +1285,8 @@ impl<'s> SegmentHolder {
             .map(|(segment_id, locked_segment)| (*segment_id, locked_segment.read()))
             .collect::<BTreeMap<_, _>>();
 
-        // Iterator produces segment:point tuples in order of point ID from all segments
-        let ordered_iter = locked_segments
+        // Iterator produces groups of points with the same point ID
+        let point_group_iter = locked_segments
             .iter()
             .map(|(&segment_id, locked_segment)| {
                 locked_segment
@@ -1297,15 +1297,13 @@ impl<'s> SegmentHolder {
                         version: None,
                     })
             })
-            .kmerge_by(|a, b| a.point_id < b.point_id);
-
-        // Iterator produces groups of point IDs
-        let chunked_iter = ordered_iter.chunk_by(|entry| entry.point_id);
+            .kmerge_by(|a, b| a.point_id < b.point_id)
+            .chunk_by(|entry| entry.point_id);
 
         let mut points = Vec::new();
         let mut points_to_remove: HashMap<SegmentId, Vec<PointIdType>> = HashMap::new();
 
-        for (point_id, group_iter) in &chunked_iter {
+        for (_, group_iter) in &point_group_iter {
             // Fill buffer with points of current chunk, need at least 2 points to deduplicate
             points.clear();
             points.extend(group_iter);
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -1270,6 +1270,12 @@ impl<'s> SegmentHolder {
     }
 
     fn find_duplicated_points(&self) -> HashMap<SegmentId, Vec<PointIdType>> {
+        struct DedupPoint {
+            segment_id: SegmentId,
+            point_id: PointIdType,
+            version: Option<SeqNumberType>,
+        }
+
         let segments = self
             .iter()
             .map(|(&segment_id, locked_segment)| (segment_id, locked_segment.get()))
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -1296,73 +1284,58 @@ impl<'s> SegmentHolder {
...
-                        .or_default()
-                        .push(point_id);
-                } else {
-                    log::debug!("Selected point {point_id} in segment {last_segment_id} for deduplication (version {last_point_version:?} versus {point_version:?} in segment {segment_id})");
-
-                    points_to_remove
-                        .entry(last_segment_id)
-                        .or_default()
-                        .push(point_id);
+            // Attach point version to
```
coszio, line 1314: nit: ... to each point?

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -1296,73 +1284,58 @@ impl<'s> SegmentHolder {
...
+                segment_id,
+                point_id,
+                version,
+            } in &points[1..]
+            {
+                log::debug!(
+                    "Selected point {point_id} in segment {segment_id} for deduplication (version {version:?} versus {:?} in segment {})",
+                    points[0].version,
+                    points[0].segment_id,
+                );
```
coszio, line 1334: I see it was also `debug!` before, but... For a segment with lots of old versions, this can create a lot of noise, doesn't it? 

I'd consider using `trace!` instead 🤔

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -1296,73 +1284,58 @@ impl<'s> SegmentHolder {
...
-                    last_point_version_opt = Some(point_version);
-                }
-            } else {
-                last_point_id_opt = Some(point_id);
-                last_segment_id_opt = Some(segment_id);
-                last_point_version_opt = None;
+            // Sort points from highest to lowest version
+            points.sort_unstable_by(|a, b| b.version.cmp(&a.version));
+
+            // Keep the first point, remove all others which are older
```
coszio, line 1323: Since we don't actually check the other's versions, we assume all other ones are older.

Just to double check, it is not possible to have two points with the same version, is it?

coszio (commented):
If performance does not degrade, I'd say I had an easier time reading this than before 👍

lib/collection/src/collection_manager/holders/segment_holder.rs
timvisee: Yes, the plan was to demote this again with <https://github.com/qdrant/qdrant/pull/5586> (as it was promoted to debug for chaos testing). But it looks like that one is already merged.

I'll create a separate PR instead to demote theses, so it doesn't depend on these changes.

lib/collection/src/collection_manager/holders/segment_holder.rs
timvisee: > Since we don't actually check the other's versions, we assume all other ones are older.

Yes, but only because they have an older version number. It's the only think that can tell us what point is newer, we cannot rely on any other data to tell us this.

> Just to double check, it is not possible to have two points with the same version, is it?

It is possible, though I'm not sure if it'll ever happen in practice with normal usage.

But even if it does happen, we can assume picking either of the points (having the same version) is fine. :+1:

lib/collection/src/collection_manager/holders/segment_holder.rs
timvisee: https://github.com/qdrant/qdrant/pull/5604

lib/collection/src/collection_manager/holders/segment_holder.rs
coszio: Sorry, I selected the wrong line. Still the questions are answered, but my concern was the operation of removing `[1..]`

timvisee: Check!

Yeah, so it should be fine because:
- index 0 is always the point we want to keep, we consider that to be the highest
- there are always at least 2 points in that loop

timvisee:
I ran some basic benchmarks.

Setup:
- 5 segments
- 100k points
- ~8600 duplicate points

Before: `[5.2221 ms 5.2353 ms 5.2494 ms]`
After: `[6.0726 ms 6.0842 ms 6.0970 ms]`

With 100 segments (and ~11k duplicates) the gap closes:

Before: `[8.8116 ms 8.8251 ms 8.8405 ms]`
After: `[9.0606 ms 9.0709 ms 9.0820 ms]`

Given that this new version is much easier to reason about and maintain I'd still like to merge it, even though it performs slightly slower.

The criterion benchmark is included in this PR.

coszio (approved):

coszio:
Thanks for providing the benchmark!

PR #5592, opened 2024-12-05
[Strict Mode] Max collection size in distributed setup

JojiiOfficial:
Depends on #5588

Allows limiting collection size (vector and payload storage separately) in distributed setups with strict mode.

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -83,7 +83,7 @@ pub struct Collection {
     // Search runtime handle.
     search_runtime: Handle,
     optimizer_cpu_budget: CpuBudget,
-    // Cached statistics of collection size, may be outdatetd.
+    // Cached statistics of collection size, may be outdated.
     collection_stats_cache: CollectionSizeStatsCache,
 }
 
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -27,8 +27,10 @@ use tokio::sync::{Mutex, RwLock, RwLockWriteGuard};
 
 use crate::collection::payload_index_schema::PayloadIndexSchema;
 use crate::collection_state::{ShardInfo, State};
+use crate::common::collection_size_stats::{
+    CollectionSizeAtomicStats, CollectionSizeStats, CollectionSizeStatsCache,
+};
 use crate::common::is_ready::IsReady;
-use crate::common::local_data_stats::{LocalDataAtomicStats, LocalDataStats, LocalDataStatsCache};
 use crate::config::CollectionConfigInternal;
 use crate::operations::config_diff::{DiffConfig, OptimizersConfigDiff};
 use crate::operations::shared_storage_config::SharedStorageConfig;
@@ -81,8 +83,8 @@ pub struct Collection {
     // Search runtime handle.
     search_runtime: Handle,
     optimizer_cpu_budget: CpuBudget,
-    // Cached stats over all local shards used in strict mode, may be outdated
-    local_stats_cache: LocalDataStatsCache,
+    // Cached statistics of collection size, may be outdatetd.
+    collection_stats_cache: CollectionSizeStatsCache,
 }
 
 pub type RequestShardTransfer = Arc<dyn Fn(ShardTransfer) + Send + Sync>;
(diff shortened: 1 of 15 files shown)


lib/collection/src/collection/mod.rs
```diff
@@ -81,8 +83,8 @@ pub struct Collection {
     // Search runtime handle.
     search_runtime: Handle,
     optimizer_cpu_budget: CpuBudget,
-    // Cached stats over all local shards used in strict mode, may be outdated
-    local_stats_cache: LocalDataStatsCache,
+    // Cached statistics of collection size, may be outdatetd.
```
agourlay, line 86:
```suggestion
    // Cached statistics of collection size, may be outdated.
```

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -1119,13 +1119,37 @@ impl ShardHolder {
...
-    pub async fn calculate_local_shards_stats(&self) -> LocalDataStats {
-        let mut stats = LocalDataStats::default();
+    /// Estimates the collections size based on local shard data. Returns `None` if no shard for the collection was found locally.
+    pub async fn estimate_collection_size_stats(&self) -> Option<CollectionSizeStats> {
+        if self.is_distributed().await {
+            // In distributed, we estimate the whole collection size by using a single local shard and multiply by amount of shards in the collection.
+            for shard in self.shards.iter() {
+                if let Some(shard_stats) = shard.1.calculate_local_shard_stats().await {
+                    // Project the single shards size to the full collection.
+                    let collection_estimate = shard_stats.multiplied_with(self.shards.len());
```
agourlay, line 1129: @timvisee would this approach break while running resharding?

I am afraid the estimate would be much larger/smaller while we transfer data to new shards.

timvisee: > would this approach break while running resharding?

Yes.

It'll not be that easy to do while resharding is going on. Depending on the current state of resharding, the calculation will be different. If there are multiple resharding operations going on at the same time, this will become more complex.

Would it be possible to sum each shard separately for now, and improve on this in a separate PR?

agourlay: @timvisee The multiplication approach relies on the fact that all shards have the assumption that all shards have more or less the same size. Are there maybe some resharding state we could leverage to refine our estimation? 

Without the multiplication trick, we have a hole in the estimation because we will only account for what is present locally.

@JojiiOfficial I am now a bit confused with the naming `estimate_collection_size_stats`.
Are we just interested in what is "local" on the node or are trying to have actual estimate of the full collection size?

For the full collection size case, we could leverage our `ReplicaSet` abstraction and multiply the size of the local shard by the number of remote replicas.

I understand that If we have zero local shard, we assume the internal API on the remote shard will be responsible to computing a the collection size.

timvisee: In a call we agreed it might be best to estimate like this: `shard_size * shard IDs * replication factor`

We use the replication factor and now the real number of replicas because the 'cost for extra replicas during transfer' is on us. The replication factor better represents what data we care about.

generall: I would vote for not including replication factor at all

agourlay: @generall Could you please elaborate on why you'd rather not account for replication factor?

My guess is that the replication factor an implementation details in the serverless setup.
But what about the non serverless use case then?

agourlay:
The error happens when the shard is not local on `peer_urls[0]`.
This means the strict mode is not properly enforced through the internal upsert API.

lib/collection/src/shards/shard_holder/mod.rs
generall: we might want to temporary create additional replicas for the purpose of data migration. In serverless, user have no control over replication factor. I would consider this size limitation as a limitation for the size of unique data, which doesn't include duplication from segments or shards

JojiiOfficial:
However all tests do pass on my machine. I'm working on it!

agourlay:
@JojiiOfficial looks like I fixed it :crossed_fingers:

agourlay:
In a next PR we can do the max payload size in distributed setup by wiring the internal payload API to the `StrictModeCheckedInternalTocProvider`

lib/collection/src/shards/shard_holder/mod.rs
timvisee: > I would consider this size limitation as a limitation for the size of unique data, which doesn't include duplication from segments or shards

:+1:

JojiiOfficial: The approach of this PR doesn't include replicas. I've updated the strict mode options documentation to make this more clear.
We'll adjust this estimation in future to eg. make it compatible with resharding.

PR #5603, opened 2024-12-09
Update pprof-0.14

agourlay:
There is a Dependabot security alert but it is not being applied.

https://github.com/qdrant/qdrant/security/dependabot/102

Let's get rid of it manually even if it is not a concern for our usage of `pprof`.


timvisee (approved):
Thanks!

PR #5613, opened 2024-12-09
Fix Debian package release build

timvisee:
Building a Debian package on release [current](https://github.com/qdrant/qdrant/actions/runs/12238399544/) fails since <https://github.com/qdrant/qdrant/pull/5479>.

This patches the release logic to fix the build in future releases.

`cargo deb` does not have many toggles, and providing `--features=service_debug` to bypass the error seems to be the best approach here. I do prefer to keep that explicit feature flag to speed up other builds as described in the [PR](<https://github.com/qdrant/qdrant/pull/5479>) that implemented it.

### All Submissions:

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -51,7 +51,7 @@ jobs:
         if: matrix.target == 'x86_64-unknown-linux-musl'
         run: |
           cargo install cargo-deb
-          cargo deb --no-strip --target ${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


generall:
maybe exclude schema_generator from deb instead?

generall:
Propose an alternative https://github.com/qdrant/qdrant/pull/5630

timvisee:
> maybe exclude schema_generator from deb instead?

It wasn't entirely clear how to achieve that, there are no `--bin` flags for example.

Your suggested alternative seems nice. I'll cover it shortly.

PR #5628, opened 2024-12-09
Atomic save chunked mmap config

IvanPleshkov:
Save chunked mmap config atomically.
Just flush is not enough because it doesn't guarantee presence on disk. This situation was caught on a real cluster.

Also, this file can be generated from another config. Do it if loading fails.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/vector_storage/chunked_mmap_vectors.rs
+++ b/lib/segment/src/vector_storage/chunked_mmap_vectors.rs
@@ -75,48 +75,56 @@ impl<T: Sized + Copy + 'static> ChunkedMmapVectors<T> {
         populate: Option<bool>,
     ) -> OperationResult<ChunkedMmapConfig> {
         let config_file = Self::config_file(directory);
-        if let Some(config) = Self::load_config(&config_file) {
-            if config.dim != dim {
-                return Err(OperationError::service_error(format!(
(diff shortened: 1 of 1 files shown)


lib/segment/src/vector_storage/chunked_mmap_vectors.rs
```diff
@@ -88,27 +98,18 @@ impl<T: Sized + Copy + 'static> ChunkedMmapVectors<T> {
...
-                )));
-            }
+        }
+    }
 
-            Ok(config)
+    fn load_config(config_file: &Path) -> Option<ChunkedMmapConfig> {
+        if config_file.exists() {
+            let file = std::fs::File::open(config_file).ok()?;
+            let config: ChunkedMmapConfig = serde_json::from_reader(file).ok()?;
```
generall, line 109: I would prefer to keep & log error in case of failure

IvanPleshkov: WHYM about keeping? As I remember, the decision was to create a new file if prev one is corrupted

IvanPleshkov: Fixed

PR #5629, opened 2024-12-09
use fsync instead of flush

generall:


IvanPleshkov (approved):

xzfc (approved):

lib/quantization/src/encoded_storage.rs
```diff
@@ -52,7 +52,7 @@ impl EncodedStorage for Vec<u8> {
     fn save_to_file(&self, path: &Path) -> std::io::Result<()> {
         let mut buffer = File::create(path)?;
         buffer.write_all(self.as_slice())?;
-        buffer.flush()?;
+        buffer.sync_all()?;
```
timvisee, line 55: Are we sure that `sync_all` implies `flush`, and that there is no need for both calls?

timvisee: Flush appears to be a noop for [Unix](https://github.com/rust-lang/rust/blob/ff7906bfe1ed264bf9c4d3abe1940e357b7e61dd/library/std/src/sys/pal/unix/fs.rs#L1426-L1429) and [Windows](https://github.com/rust-lang/rust/blob/ff7906bfe1ed264bf9c4d3abe1940e357b7e61dd/library/std/src/sys/pal/windows/fs.rs#L602-L604), so in that case it should be fine. But I'm not sure about other platform specifics.

PR #5630, opened 2024-12-09
Explict configuration for deb package build

generall:
- **explicit configuration of deb package**
- **use special build feature**

Bonus: now works with web-ui

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ tower = { version = "0.5.1", features = ["util"] }
 tower-layer = "0.3.3"
 reqwest = { workspace = true }
 # rustls minor version must be synced with actix-web
-rustls = { version = "0.23.19", default-features = false, features = [ "logging", "std", "tls12", "ring"] }
+rustls = { version = "0.23.19", default-features = false, features = ["logging", "std", "tls12", "ring"] }
 rustls-pki-types = "1.10.0"
 rustls-pemfile = "2.2.0"
 prometheus = { version = "0.13.4", default-features = false }
@@ -174,7 +174,7 @@ indicatif = "0.17.9"
 itertools = "0.13.0"
 log = "0.4.22"
 memmap2 = "0.9.5"
-nix = { version = "0.29", features = ["fs"]}
+nix = { version = "0.29", features = ["fs"] }
 num-traits = "0.2.19"
 ordered-float = "4.5"
 rayon = "1.10.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -38,6 +39,7 @@ stacktrace = ["rstack-self"]
 chaos-testing = []
 data-consistency-check = ["collection/data-consistency-check"]
 gpu = ["gpu/gpu", "segment/gpu"]
+deb = []
```
timvisee, line 42: Don't we need at least `default` in here?

Cargo.toml
```diff
@@ -282,4 +284,21 @@ opt-level = 3
...
+section = "database"
+priority = "optional"
+extended-description = """\
+Qdrant is a high-performance vector search engine. It is designed to index and search large collections of high-dimensional vectors.\
+"""
+assets = [
+    ["target/release/qdrant", "usr/bin/", "755"],
+    ["README.md", "usr/share/doc/qdrant/README", "644"],
+    ["static/*/*", "/etc/qdrant/static", "644"],
+    ["config/deb.yaml", "etc/qdrant/config.yaml", "644"],
```
timvisee, line 303: Only one of the lines has the `/` prefix. Is that intended?

Cargo.toml
```diff
@@ -282,4 +284,21 @@ opt-level = 3
 
 [patch.crates-io]
 # Temporary patch until our PRs are merged.
-tar = { git = "https://github.com/qdrant/tar-rs", branch="main" }
+tar = { git = "https://github.com/qdrant/tar-rs", branch = "main" }
+
+[package.metadata.deb]
+maintainer = "Qdrant Team <team@qdrant.com>"
+depends = "$auto"
+features = ["deb"]
```
timvisee, line 292: This is the default, but it makes it a bit more explicit for me that that default features are used as well.

```suggestion
default-features = true
features = ["deb"]
```

Cargo.toml
generall: All examples are actually using path without `/`, so I think it is safe to remove

timvisee: It should be a relative path to properly resolve it when installing the package.

```suggestion
    ["target/release/qdrant", "usr/bin/", "755"],
    ["README.md", "usr/share/doc/qdrant/README", "644"],
    ["static/*/*", "etc/qdrant/static", "644"],
    ["config/deb.yaml", "etc/qdrant/config.yaml", "644"],
```

timvisee (approved):
Builds fine locally. Thanks!

PR #5634, opened 2024-12-10
Enable resetting `max_optimization_threads` to automatic

### LINKED ISSUE #5243: Can't return collection config to default null value

coszio:
Fixes #5243

Continues the work from @gulshan-rs (#5520) to enable setting `max_optimization_threads` to automatic mode.

The main problem is that serde has little distinction between setting to `null` and not including the field. So, to make it explicit, this PR switches the config diff to be an enum of `["auto" | usize ]`, but only on the API level

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1311,6 +1311,7 @@ dependencies = [
  "ordered-float 4.5.0",
  "ph",
  "rand 0.8.5",
+ "schemars",
  "semver",
  "serde",
  "tap",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1311,7 +1311,6 @@ dependencies = [
  "ordered-float 4.5.0",
  "ph",
  "rand 0.8.5",
- "schemars",
  "semver",
  "serde",
  "tap",
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -5,6 +5,7 @@ use std::time::Instant;
 use chrono::{NaiveDateTime, Timelike};
 use common::counter::hardware_accumulator::HwMeasurementAcc;
 use common::counter::hardware_counter::HardwareCounterCell;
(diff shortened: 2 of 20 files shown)


lib/api/src/grpc/conversions.rs
```diff
@@ -1607,6 +1607,76 @@ impl From<segment::data_types::order_by::StartFrom> for StartFrom {
     }
 }
 
+impl TryFrom<MaxOptimizationThreads> for rest::MaxOptimizationThreads {
```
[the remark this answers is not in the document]
coszio: That's actually a nice idea. Since it is a common pattern in all the conversions, it might be a good idea to generalize it into a conversion error which takes the type it failed to convert from, and has the same error message for all of them.

It would be best to do it in a separate PR, though :)

lib/common/common/Cargo.toml
```diff
@@ -31,6 +31,7 @@ memmap2 = { workspace = true }
 semver = { workspace = true }
 zerocopy = { workspace = true }
 log = { workspace = true }
+schemars = { workspace = true }
```
generall, line 34: I don't think this is needed here

```suggestion
```

Cargo.lock
```diff
@@ -1311,6 +1311,7 @@ dependencies = [
  "ordered-float 4.5.0",
  "ph",
  "rand 0.8.5",
+ "schemars",
```
generall, line 1314:
```suggestion
```

not needed as well

generall (approved):

PR #5639, opened 2024-12-12
zerocopy: 0.7 -> 0.8

xzfc:
This PR bumps usage of `zerocopy` from `0.7` to `0.8`.

It's brings some incompatible changes to the API, but the new API neatly simplifies the code that reads items/slices in a sequential manner. Check the updated `MmapPostings::get_reader()`.

A list of changes and migration guide is listed here: https://github.com/google/zerocopy/discussions/1680.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
+++ b/lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
@@ -193,14 +193,12 @@ impl MmapPostings {
         let path = path.into();
         let mmap = open_read_mmap(&path, AdviceSetting::Advice(Advice::Normal), populate)?;
 
-        let header = PostingsHeader::read_from_prefix(&mmap)
-            .map_err(|_| {
-                io::Error::new(
-                    io::ErrorKind::InvalidData,
-                    format!("Invalid header deserialization in {}", path.display()),
-                )
-            })?
-            .0;
+        let (header, _) = PostingsHeader::read_from_prefix(&mmap).map_err(|_| {
+            io::Error::new(
+                io::ErrorKind::InvalidData,
+                format!("Invalid header deserialization in {}", path.display()),
+            )
+        })?;
 
         Ok(Self { path, mmap, header })
     }
--- a/lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
+++ b/lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
@@ -7,7 +7,7 @@ use common::zeros::WriteZerosExt;
 use memmap2::Mmap;
 use memory::madvise::{Advice, AdviceSetting};
 use memory::mmap_ops::open_read_mmap;
-use zerocopy::{AsBytes, FromBytes, FromZeroes};
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
 use crate::index::field_index::full_text_index::compressed_posting::compressed_chunks_reader::ChunkReader;
 use crate::index::field_index::full_text_index::compressed_posting::compressed_common::CompressedPostingChunksIndex;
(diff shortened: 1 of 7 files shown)


lib/segment/src/index/field_index/mmap_point_to_values.rs
```diff
@@ -351,9 +347,9 @@ impl<T: MmapValue + ?Sized> MmapPointToValues<T> {
         if point_id < self.header.points_count as PointOffsetType {
             let range_offset = (self.header.ranges_start as usize)
                 + (point_id as usize) * std::mem::size_of::<MmapRange>();
-            self.mmap
-                .get(range_offset..range_offset + std::mem::size_of::<MmapRange>())
-                .and_then(MmapRange::read_from)
+            MmapRange::read_from_prefix(self.mmap.get(range_offset..)?)
```
IvanPleshkov, line 350: As I understand, `read_from_prefix` truncates additional bytes and we can provide longer byte slice

lib/segment/src/index/field_index/mmap_point_to_values.rs
```diff
@@ -263,11 +259,11 @@ impl<T: MmapValue + ?Sized> MmapPointToValues<T> {
     pub fn open(path: &Path) -> OperationResult<Self> {
         let file_name = path.join(POINT_TO_VALUES_PATH);
         let mmap = open_write_mmap(&file_name, AdviceSetting::Global, false)?;
-        let header = Header::read_from_prefix(mmap.as_ref()).ok_or_else(|| {
-            OperationError::InconsistentStorage {
+        let header = Header::read_from_prefix(mmap.as_ref())
```
IvanPleshkov, line 262: I propose `let (header, _) = `instead of `.0`. not critical

lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
```diff
@@ -92,36 +91,17 @@ impl MmapPostings {
...
-
-        let data = self.mmap.get(data_offset..alignment_offset)?;
-
-        let remainder_mem = self
-            .mmap
-            .get(remainder_offset..remainder_offset + remainder_size_bytes)?;
-        let remainder_postings = u32::slice_from(remainder_mem)?;
+        let bytes = self.mmap.get(header.offset as usize..)?;
+        let (last_doc_id, bytes) = PointOffsetType::read_from_prefix(bytes).ok()?;
+        let (chunks, bytes) = <[CompressedPostingChunksIndex]>::ref_from_prefix_with_elems(
```
IvanPleshkov, line 96: TBH, `slice_from` in prev version of zerocopy looks easier

IvanPleshkov (approved):

lib/segment/src/index/field_index/mmap_point_to_values.rs
xzfc: In the old code we truncate additional bytes manually (line 355).

lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
xzfc: In short, now they don't provide specialized method for slices, instead you just write `[T]` and use regular methods.

Old (two methods):
- `T::ref_from(…)` for `&T`
- `T::slice_from(…)` for `&[T]`

New (same method):
- `T::ref_from_bytes(…)` for `&T`
- `<[T]>::ref_from_bytes(…)` for `&[T]`

---

As for variations like `ref_from_prefix_with_elems`, they are consistent with the "sequential reading" approach. I.e., from a high level perspective it's seen as this:
```rust
let (some_field_1, bytes) = read_field_1_from_prefix(bytes)?;
let (some_field_2, bytes) = read_field_2_from_prefix(bytes)?;
let (some_field_3, bytes) = read_field_4_from_prefix(bytes, length)?;
let (some_field_4, bytes) = read_field_5_from_prefix(bytes)?;
…
```

It resembles the approach that's often used in parsers, e.g. [this](https://github.com/biomejs/biome/blob/c2dda66cc7533019bbd6c76960fcd68cd94cf1be/xtask/coverage/src/symbols/msts.rs#L244-L317). You don't fiddle with offsets, you read things in a sequence.

---

The name `ref_from_prefix_with_elems` is scary because it supports arbitrary [DST](https://doc.rust-lang.org/nomicon/exotic-sizes.html#dynamically-sized-types-dsts)s, not just slices.

timvisee (approved):

PR #5650, opened 2024-12-16
Simplify gRPC validation

timvisee:
In gRPC we had a lot of custom validation rules because earlier validator versions had limited support for numbers wrapped in options.

Since <https://github.com/qdrant/qdrant/pull/4894> we can use the built-in range and length validators on numbers wrapped in options.

I've removed the custom validation functions we now don't need anymore to greatly simplify validation in gRPC.

### All Submissions:


lib/api/build.rs
```diff
@@ -191,9 +191,10 @@ fn configure_validation(builder: Builder) -> Builder {
             ("UpsertPoints.points", ""),
             ("DeletePoints.collection_name", "length(min = 1, max = 255)"),
             ("UpdatePointVectors.collection_name", "length(min = 1, max = 255)"),
-            ("UpdatePointVectors.vectors", "custom(function = \"crate::grpc::validate::validate_named_vectors_not_empty\", message = \"must specify vectors to update\")"),
+            ("UpdatePointVectors.points", ""),
```
agourlay, line 194: What?! The build script does not fail on non existing field?

timvisee: Sadly not. It's a limitation in the code generator of tonic/prost.

agourlay (approved):
Very neat change :clap:

PR #5651, opened 2024-12-16
GraphLinks: replace trait with enum

xzfc:
~Depends on #5666 and #5492. Read only the last commit during the review.~

This PR improves the performance of graph links by merging `impl GraphLinks`, `struct GraphLinksRam` and `struct GraphLinksMmap` into a single `struct GraphLinks`.

The new structure is self-referential (using [`self_cell`](https://docs.rs/self_cell/1.1.0) crate), and contains `GraphLinksView` with precalculated slices.
Enum dispatching is not required: once `GraphLinksView` is created during loading, it doesn't care about the underlying storage type.

At a high level, the new structure looks like this:
```rust
self_cell::self_cell! {
    pub struct GraphLinks {
        owner: GraphLinksEnum,     // owns the data
        #[covariant]
        dependent: GraphLinksView, // slice references to the data
    }
}

enum GraphLinksEnum {
    Ram(Vec<u8>),
    Mmap(Arc<Mmap>),
}

struct GraphLinksView<'a> {
    reindex: &'a [PointOffsetType],
    offsets: &'a [u64],
    compression: CompressionInfo<'a>,
    level_offsets: Vec<u64>,
}

enum CompressionInfo<'a> { … }
```

The new implementation is more efficient, because slices are created only once during loading and not recalculated on every access, so we can avoid additional indirections and bounds checks. For comparison:
```rust
// old
let offsets_range: Range<usize> = self.info.offsets_range();
let offsets: &[u64] = <[u64]>::ref_from_bytes(&self.data[offsets_range]).unwrap(); // indirection + bounds check
let links_start: Range<usize> = offsets[idx] as usize..offsets[idx + 1] as usize;

// new
let links_range: Range<usize> = self.offsets[idx] as usize..self.offsets[idx + 1] as usize;
```

On the test introduced in #5666, the new implementation makes the HNSW search 2-3% faster; and makes compressed version faster than the uncompressed one. I'll provide more benchmark results later.

Besides the performance improvements, replacing a trait with a enum made it possible to simplify some code.

### DIFF
--- a/lib/segment/benches/hnsw_search_graph.rs
+++ b/lib/segment/benches/hnsw_search_graph.rs
@@ -13,7 +13,6 @@ use rayon::iter::{IntoParallelIterator as _, ParallelIterator as _};
 use segment::fixtures::index_fixtures::{random_vector, FakeFilterContext, TestRawScorerProducer};
 use segment::index::hnsw_index::graph_layers::GraphLayers;
 use segment::index::hnsw_index::graph_layers_builder::GraphLayersBuilder;
-use segment::index::hnsw_index::graph_links::GraphLinksRam;
 use segment::index::hnsw_index::point_scorer::FilteredScorer;
 use segment::spaces::simple::CosineMetric;
 
@@ -52,8 +51,7 @@ fn hnsw_benchmark(c: &mut Criterion) {
 
         if graph_fname.exists() && links_fname.exists() {
             eprintln!("Loading cached links from {links_fname:?}");
-            graph_layers =
-                GraphLayers::<GraphLinksRam>::load(&graph_fname, &links_fname, false).unwrap();
+            graph_layers = GraphLayers::load(&graph_fname, &links_fname, false, false).unwrap();
         } else {
             let mut graph_layers_builder =
                 GraphLayersBuilder::new(NUM_VECTORS, M, M * 2, EF_CONSTRUCT, 10, USE_HEURISTIC);
@@ -78,7 +76,7 @@ fn hnsw_benchmark(c: &mut Criterion) {
                 });
 
             graph_layers = graph_layers_builder
-                .into_graph_layers::<GraphLinksRam>(Some(&links_fname), false)
+                .into_graph_layers(&links_fname, false, false)
                 .unwrap();
             graph_layers.save(&graph_fname).unwrap();
         }
@@ -96,7 +94,7 @@ fn hnsw_benchmark(c: &mut Criterion) {
         })
     });
 
-    graph_layers.compress();
+    graph_layers.compress_ram();
     let mut rng = StdRng::seed_from_u64(42);
     group.bench_function("compressed", |b| {
         b.iter(|| {
--- a/lib/segment/src/index/hnsw_index/graph_links.rs
+++ b/lib/segment/src/index/hnsw_index/graph_links.rs
@@ -2,15 +2,14 @@ use std::cmp::Reverse;
 use std::fs::{File, OpenOptions};
 use std::io::{Read as _, Write};
 use std::mem::take;
-use std::ops::Range;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::sync::Arc;
 
 use common::bitpacking::packed_bits;
 use common::bitpacking_links::{for_each_packed_link, pack_links, MIN_BITS_PER_VALUE};
 use common::types::PointOffsetType;
 use common::zeros::WriteZerosExt as _;
-use itertools::Either;
+use itertools::{Either, Itertools as _};
 use memmap2::Mmap;
 use memory::{madvise, mmap_ops};
 use zerocopy::little_endian::U64;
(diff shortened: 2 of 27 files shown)


lib/segment/src/index/vector_index_base.rs
```diff
@@ -58,8 +57,7 @@ pub trait VectorIndex {
 #[derive(Debug)]
 pub enum VectorIndexEnum {
     Plain(PlainIndex),
-    HnswRam(HNSWIndex<GraphLinksRam>),
-    HnswMmap(HNSWIndex<GraphLinksMmap>),
+    Hnsw(HNSWIndex),
```
IvanPleshkov, line 60: Honestly, it looks inconsistent to me. Each variant has `Ram`/`Mmap`/`Compressed` versions and only HNSW index is an exception with internal definitions. Like quantization and vector storage, we do the same: expand all possible options in a single enum. Future developers may be confused and think that hnsw does not have ram/mmap versions. I won't say it's a blocker, but let's discuss it. Probably we can inline all `HNSWIndex` variants here

lib/segment/benches/hnsw_search_graph.rs
```diff
@@ -1,60 +1,116 @@
...
-        GraphLayersBuilder::new(NUM_VECTORS, M, M * 2, EF_CONSTRUCT, 10, USE_HEURISTIC);
-    for idx in 0..(NUM_VECTORS as PointOffsetType) {
-        let added_vector = vector_holder.vectors.get(idx as VectorOffsetType).to_vec();
-        let raw_scorer = vector_holder.get_raw_scorer(added_vector).unwrap();
-        let scorer = FilteredScorer::new(raw_scorer.as_ref(), Some(&fake_filter_context));
-        let level = graph_layers_builder.get_random_layer(&mut rng);
-        graph_layers_builder.set_levels(idx, level);
-        graph_layers_builder.link_new_point(idx, scorer);
+    // Note: make sure that vector generation is deterministic.
+    let vector_holder =
```
IvanPleshkov, line 35: Are this changes from https://github.com/qdrant/qdrant/pull/5666? Did you add something new or is it copy-paste from another PR?

generall:
@xzfc could you please resolve merge conflict?

lib/segment/benches/hnsw_search_graph.rs
xzfc: Yes, this PR is based on #5666. Only the last commit is related to this PR. Check the commit tree.

lib/segment/src/index/vector_index_base.rs
xzfc: TBH I don't see an issue here. Sparse is already a different beast than HNSW/dense, and now they are different just a bit more. Whether an enum is expanded here I see an implementation detail of each index.

I did two things here:
- Removed a generic parameter from `HNSWIndex`. Pros: lesser compiler times and binary sizes.
- Removed an enum variant from `VectorIndexEnum` as it duplicates an internal runtime flag of `HNSWIndex`. Among other things, this change makes an [Illegal State Unrepresentable](https://corrode.dev/blog/illegal-state/), i.e. we can't have `on_disk: true` with `HnswRam` by accident.

IvanPleshkov: @timvisee needs additional opinion. After a personal discussion with @xzfc , we don't have a strong vision in the same direction

coszio: My personal unsolicited take: @xzfc makes good arguments, for me it makes sense to go with this new design, which partitions vector indices by nature, then by implementation. 

Regarding consistency, I would argue that all the other `Sparse*` variants should undergo a similar refactoring (joining them into a separate enum)

coszio (commented):
I like the refactoring into an enum 👍

However, I am failing to see the benefit of the `self_cell` macro. Why can't it just be a `try_new` implementation? I see two cons of using it here:
- it is only used as a pattern rather than a tool for maintaining less code.
- Additional complexity: I did not know about the crate before, and had to go into its docs to understand what it was about.

lib/segment/src/index/vector_index_base.rs
IvanPleshkov: I agree that @xzfc proposal is good. But I believe we won't do the same for all variants and all enums with old design

IvanPleshkov: Anyway thanks @coszio! I find your comment as "looks good", resolve this thread

lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -556,94 +507,40 @@ impl GraphLinks for GraphLinksMmap {
...
-            );
-        } else {
-            let all_links = <[PointOffsetType]>::ref_from_bytes(all_links).unwrap();
-            for &link in &all_links[links_range] {
-                f(link);
-            }
-        }
+    pub fn links_vec(&self, point_id: PointOffsetType, level: usize) -> Vec<PointOffsetType> {
+        let mut links = Vec::new();
+        self.for_each_link(point_id, level, |link| links.push(link));
```
timvisee, line 516: Can we potentially preallocate this vec?

lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -556,94 +507,40 @@ impl GraphLinks for GraphLinksMmap {
...
-                    // incorrect level because point_id is out of range
-                    return level - 1;
-                }
-            } else {
-                // incorrect level because this level is larger that available levels
-                return level - 1;
+    /// Convert the graph links to a vector of edges, suitable for passing into
+    /// [`GraphLinksConverter::new`] or using in tests.
+    pub fn into_edges(self) -> Vec<Vec<Vec<PointOffsetType>>> {
+        let mut edges = Vec::new();
```
timvisee, line 523:
```suggestion
        let mut edges = Vec::with_capacity(self.num_points());
```

lib/segment/src/index/vector_index_base.rs
timvisee: Thanks for bringing this to our attention Ivan, it's good to discuss changes with this significance.

I agree that it would be nice to have the same architecture in multiple places, but in my opinion xzfc gives two compelling arguments. Based on that I also vote to keep the design in this PR.

timvisee (commented):
On a first review pass I see no problems in this PR.

I'd like to do a second review pass tomorrow to carefully digest all the small details in this PR, after which I'll leave my stamp of approval.

coszio:
> It is to allow self reference without implementing this from scratch.

Yeah, no, after looking deeper into it, I realize why it is needed. It is not possible to _just_ create a struct which references itself. Sorry for the noise!

coszio (approved):

IvanPleshkov (approved):

timvisee (approved):
Thank you for your work on this! :pray:

PR #5653, opened 2024-12-16
Fix rate limiting of internal update operations

agourlay:
This PR ensures that we do not rate limit internal update operations used during transfers.

This is achieved by lifting the write rate limiter up from local shard to the replica set.
This new location has access to the replica state to decide whether the update should be rate limited.

I have added an integration tests to demonstrate that a shard can be transferred when the write limiting budget is used up.

Interestingly I found a couple of spots where read requests involved in shard transfers are wrongly rate limited.
I have annotated those for a future PR.

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -101,7 +101,6 @@ pub struct LocalShard {
     pub(super) search_runtime: Handle,
     disk_usage_watcher: DiskUsageWatcher,
     read_rate_limiter: ParkingMutex<Option<RateLimiter>>,
-    write_rate_limiter: ParkingMutex<Option<RateLimiter>>,
 }
 
 /// Shard holds information about segments and WAL.
@@ -206,13 +205,6 @@ impl LocalShard {
             .and_then(|strict_mode| strict_mode.read_rate_limit.map(RateLimiter::new_per_minute));
         let read_rate_limiter = ParkingMutex::new(read_rate_limiter);
 
-        let write_rate_limiter = config.strict_mode_config.as_ref().and_then(|strict_mode| {
-            strict_mode
-                .write_rate_limit
-                .map(RateLimiter::new_per_minute)
-        });
-        let write_rate_limiter = ParkingMutex::new(write_rate_limiter);
-
         drop(config); // release `shared_config` from borrow checker
 
         Self {
@@ -232,7 +224,6 @@ impl LocalShard {
             total_optimized_points,
             disk_usage_watcher,
             read_rate_limiter,
-            write_rate_limiter,
         }
     }
 
@@ -770,13 +761,6 @@ impl LocalShard {
                 read_rate_limiter_guard
                     .replace(RateLimiter::new_per_minute(read_rate_limit_per_sec));
             }
-
-            // update write rate limiter
-            if let Some(write_rate_limit_per_sec) = strict_mode_config.write_rate_limit {
-                let mut write_rate_limiter_guard = self.write_rate_limiter.lock();
-                write_rate_limiter_guard
-                    .replace(RateLimiter::new_per_minute(write_rate_limit_per_sec));
-            }
         }
     }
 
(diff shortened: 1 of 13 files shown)


tests/consensus_tests/test_shard_transfer_rate_limiting.py
```diff
@@ -0,0 +1,245 @@
...
+        if not r.ok:
+            rate_limited = True
+            assert r.status_code == 429, f"Failure body {r.json()}"
+            assert "Rate limiting exceeded: Write rate limit exceeded, retry later" in r.json()['status']['error']
+            break
+
+    assert rate_limited, "Expected write rate limit to be reached"
+
+    # Move shard `shard_id` to peer `target_peer_id` without write budget
+    r = requests.post(
```
agourlay, line 153: this operation failed with rate limits before the fix

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -101,7 +101,6 @@ pub struct LocalShard {
     pub(super) search_runtime: Handle,
     disk_usage_watcher: DiskUsageWatcher,
     read_rate_limiter: ParkingMutex<Option<RateLimiter>>,
-    write_rate_limiter: ParkingMutex<Option<RateLimiter>>,
```
JojiiOfficial, line 104: Is there a specific reason why we only move `write_rate_limiter` and not both?

JojiiOfficial (approved):

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -111,6 +112,7 @@ pub struct ShardReplicaSet {
     write_ordering_lock: Mutex<()>,
     /// Local clock set, used to tag new operations on this shard.
     clock_set: Mutex<ClockSet>,
+    write_rate_limiter: parking_lot::Mutex<Option<RateLimiter>>,
```
coszio, line 115: Shouldn't this be an `Option<Mutex<RateLimiter>>`? For the regular case of no limiting, this might introduce overhead when locking

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -730,11 +754,36 @@ impl ShardReplicaSet {
...
+            }
+        }
+        Ok(())
+    }
+
+    /// Check if the write rate limiter allows the operation to proceed
+    ///
+    /// Returns an error if the rate limit is exceeded.
+    fn check_write_rate_limiter(&self) -> CollectionResult<()> {
+        if let Some(rate_limiter) = self.write_rate_limiter.lock().as_mut() {
```
coszio, line 780: why `.as_mut()`?

tests/consensus_tests/test_shard_transfer_rate_limiting.py
```diff
@@ -0,0 +1,245 @@
...
+        f"{source_uri}/collections/test_collection/cluster", json={
+            "move_shard": {
```
timvisee, line 155: Just to be clear, move shard does not recreate the entire replica on the target peer and keeps the existing rate limiter in place?

I'm asking because some of our shard operations may recreate the replica, in which case it may also recreate and replenish the rate limiter.

timvisee (approved):

lib/collection/src/shards/local_shard/mod.rs
agourlay: I will move the read limiter as well if it is required to fix potential miscounts.

I also prefer splitting my work in small PRs :smile_cat:

lib/collection/src/shards/replica_set/mod.rs
agourlay: Thanks for the idea :+1: 

I applied it in https://github.com/qdrant/qdrant/pull/5653/commits/a5cbe8f6c6f16c8d00c8614f69d257c17db6df73 and even found a bug regarding disabling rate limiting on the way!

If you are happy with this I will apply the same transformation and fix to the read limiter in a follow up PR

tests/consensus_tests/test_shard_transfer_rate_limiting.py
agourlay: Move sets up the the local shard in the existing target peer's replica set.

But we track only operations done to the local shard anyway, initializing a new local shard will always result in a new full rate limiter.

lib/collection/src/shards/replica_set/mod.rs
agourlay: Renamed it to `check_and_update`.

tests/openapi/test_strictmode.py
```diff
@@ -643,6 +643,23 @@ def test_strict_mode_read_rate_limiting(collection_name):
     # loose check, as the rate limiting might not be exact
     assert failed_count > 5, "Rate limiting did not work"
 
+    # TODO(ratelimite) test read limit can be disabled
```
timvisee, line 646:
```suggestion
    # TODO(ratelimiting) test read limit can be disabled
```

lib/collection/src/shards/replica_set/mod.rs
timvisee: Yes, this looks good!

I like that we now prevent locking if disabled. And the mutable self references we now need to update the rate limiter seem fine to me.

Thanks for adding support for disabling as well.

PR #5655, opened 2024-12-16
Fix debug assertion for propagating segment proxy deletes

timvisee:
Fix the debug assertion panic causing flaky tests that were implemented in <https://github.com/qdrant/qdrant/pull/5573>.

Point delete operations don't have to be newer than the segment version, but they should not be older than the point version of the wrapped segment.

Example failures:
- > proxied point deletes should have newer version than segment
- https://github.com/qdrant/qdrant/actions/runs/12207755405/job/34059745505
- https://github.com/qdrant/qdrant/actions/runs/12254635448/job/34185980363

### All Submissions:


agourlay:
@timvisee Maybe we should also adjust the two assert messages to reflect the new logic?

timvisee:
Good idea. Updated. Thanks!

agourlay (approved):
:sheep:

PR #5571, opened 2024-12-02
Integrate mmap bool index

coszio:
Needs #5526 

Adds the option to set `on_disk=true` for boolean index.

Since this index is mutable, it is also enabled for appendable segments, when the `on_disk` setting is true.

For the sake of compatibility we will support both indices for a version. We can set the new one as the only boolean index at a later version.

PR #5673, opened 2024-12-17
Introduce facet index trait

coszio:
Refactors the faceting infrastructure to make all facet-enabled indices implement the same trait.

This also fixes a bug introduced in #5571 ([failed test run](https://github.com/qdrant/qdrant-client/actions/runs/12373060583/job/34533583784?pr=866)) which would iterate over the whole index, instead of just a single point's values


joein (approved):
python client tests are green, thanks 👍

PR #5675, opened 2024-12-18
Extract plain vector index

agourlay:
This PR extracts the plain vector index into a dedicated module and renames it for clarity.

I am proposing this change because I got confused while reviewing https://github.com/qdrant/qdrant/pull/5587 when seeing changes in the `plain_payload_index.rs` file which were not related to payloads.

### DIFF
--- a/lib/segment/src/segment_constructor/segment_constructor_base.rs
+++ b/lib/segment/src/segment_constructor/segment_constructor_base.rs
@@ -22,7 +22,7 @@ use crate::id_tracker::simple_id_tracker::SimpleIdTracker;
 use crate::id_tracker::{IdTracker, IdTrackerEnum, IdTrackerSS};
 use crate::index::hnsw_index::gpu::gpu_devices_manager::LockedGpuDevice;
 use crate::index::hnsw_index::hnsw::{HNSWIndex, HnswIndexOpenArgs};
-use crate::index::plain_payload_index::PlainIndex;
+use crate::index::plain_vector_index::PlainVectorIndex;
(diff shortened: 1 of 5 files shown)


agourlay:
@coszio I hope to not make your work on https://github.com/qdrant/qdrant/pull/5587 more difficult with this change. 
We can sync regarding merging order :)

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -376,9 +376,11 @@ pub(crate) fn create_vector_index(
...
 ) -> OperationResult<VectorIndexEnum> {
     let vector_index = match &vector_config.index {
-        Indexes::Plain {} => {
-            VectorIndexEnum::Plain(PlainIndex::new(id_tracker, vector_storage, payload_index))
-        }
+        Indexes::Plain {} => VectorIndexEnum::Plain(PlainVectorIndex::new(
+            id_tracker,
+            vector_storage,
+            payload_index,
+        )),
```
coszio, line 383: I don't get why this changed the formatting, am I missing something? 🦮

coszio (approved):

lib/segment/src/segment_constructor/segment_constructor_base.rs
agourlay: the struct name went from `PlainIndex` to `PlainVectorIndex` which makes the line longer

coszio:
Let's merge yours first, it should be easier to make the compiler complain in my rebasing than to do careful copy pasting

PR #5677, opened 2024-12-19
Add task to clean local shard in the background

timvisee:
_WIP_

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/collection/src/collection/clean.rs
```diff
@@ -0,0 +1,318 @@
...
+        shard_id: ShardId,
+        wait: bool,
+        timeout: Option<Duration>,
+    ) -> CollectionResult<UpdateStatus> {
+        let mut tasks = self.tasks.upgradable_read();
+
+        // Await existing task if not cancelled or failed
+        if let Some(task) = tasks
+            .get(&shard_id)
+            .filter(|task| !task.is_cancelled_or_failed())
```
agourlay, line 58: Are failed tasks eventually pruned from the Map?

What about logging the task error as well?

lib/collection/src/collection/clean.rs
```diff
@@ -0,0 +1,318 @@
...
+                Err(_) => return Ok(UpdateStatus::Acknowledged),
+            }
+        }
+    }
+
+    /// Cancel cleaning of a shard and mark it as dirty
+    pub(super) fn invalidate(&self, shard_id: ShardId) {
+        let removed = self.tasks.write().remove(&shard_id);
+        if let Some(task) = removed {
+            task.abort();
```
agourlay, line 129: Here we are not `awaiting` the aborted task.
Is this intended?

lib/collection/src/collection/clean.rs
```diff
@@ -0,0 +1,318 @@
...
+            if !shard.is_local().await {
+                break ShardCleanStatus::Failed(format!("Shard {shard_id} is not a local shard"));
+            }
+
+            // Scroll batch of points with hash ring filter
+            let filter = shard_holder
+                .hash_ring_filter(shard_id)
+                .expect("hash ring filter");
+            let filter = Filter::new_must_not(Condition::CustomIdChecker(Arc::new(filter)));
+            let mut ids = match shard
```
agourlay, line 217: can you please add a `TODO(ratelimiter) this scroll should not be rate limited` for me :)

lib/collection/src/collection/clean.rs
```diff
@@ -0,0 +1,318 @@
...
+            .statuses()
+            .into_iter()
+            .map(|(shard_id, status)| (shard_id, status.into()))
+            .collect()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(super) enum ShardCleanStatus {
+    Started,
```
agourlay, line 303: maybe we could track the starting time in this variant?

lib/collection/src/collection/clean.rs
```diff
@@ -0,0 +1,318 @@
...
+    Done,
+    Failed(String),
+    Cancelled,
```
agourlay, line 306: Are there not specific cancelling reason to capture?

lib/collection/src/collection/clean.rs
timvisee: Can do!

I actually plan to add a `Progress(something)` state which keeps being bumped until we eventually finish. I think I like that better as it'd show it's actually doing something.

If we implement the above we'd loose the Started variant almost immediately, also loosing the start time. So based on that I don't think it's worth it to include a starting time.

lib/collection/src/collection/clean.rs
timvisee: I initially thought - we don't really care.

But on second thought it may be better to await, thanks!

If we change the hash rings (and invalidate here), we don't want it to delete the wrong points based on the old hash ring. Awaiting here would resolve that.

That does mean it might await on a non-async lock, so I might replace it with a tokio RwLock instead.

I'll think about it for a bit and implement a fix.

lib/collection/src/collection/clean.rs
timvisee: > Are failed tasks eventually pruned from the Map?

Yes. If a task failed it passes this if-block and creates a new task below it. The new task will replace this one, removing the failed task.

If we don't call the clean endpoint again it'll remain until either of:
- we restart Qdrant
- we delete/recreate the collection
- we invalidate this specific shard due to some other reason (like resharding)

> What about logging the task error as well?

Good idea.

lib/collection/src/collection/clean.rs
timvisee: I've just implemented the progress signal:

https://github.com/user-attachments/assets/49994794-4ca3-40fe-b2b0-0406d1e9ff4a

lib/collection/src/collection/clean.rs
timvisee: After another look it isn't super obvious on how to implement this.

We abort the task through a [`CancellationToken`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html), which is basically just a `cancelled: bool`. I'm not sure I like attaching extra payload to it just to pass the cancellation reason.

docs/redoc/master/openapi.json
```diff
@@ -12000,6 +12007,60 @@
...
+                  "deleted_points": {
+                    "type": "integer",
+                    "format": "uint",
+                    "minimum": 0
+                  }
+                }
+              }
+            },
+            "additionalProperties": false
+          },
```
coszio, line 12041: I have suspicion that the python client will struggle to generate these anonymous variants. Might be better to just add extra structs for these cases

timvisee: Hmm. These types are exclusively used in the telemetry API though, and not really intended to be parsed/used directly by our users. I therefore think it's fine to have them like this.

@generall Thoughts?

timvisee: I shortly mentioned this in our daily. I'll add proper types here. I have a separate time for internal vs telemetry anyway, so this is easy to pull off.

It's not strictly necessary as a user will never use this as input, but it's easy to fix so I'll implement it anyway.

timvisee: Fixed in <https://github.com/qdrant/qdrant/pull/5677/commits/60d86d1953161187a7ea62f7afdf8de431b69ccd>

ffuugoo (approved):

Issue #5669, opened 2024-12-17
Migrate Strict Mode tests


JojiiOfficial:
Migrate strict mode [openapi](https://github.com/qdrant/qdrant/blob/7394f65b2d29387436a7ab6f0c1d65b53ea980ad/tests/openapi/test_strictmode.py#L647) and [consensus](https://github.com/qdrant/qdrant/pull/5679/files#diff-baa704569daab5bf3c29925da4c59db4e6a423d9e49987690599f69f0a57f22fR120) tests when the default payload storage type has been migrated to mmap.

PR #5679, opened 2024-12-19
[Strict Mode] distributed max collection payload size

JojiiOfficial:


agourlay (approved):

src/tonic/api/points_internal_api.rs
```diff
@@ -258,7 +258,7 @@ impl PointsInternal for PointsInternalService {
             .ok_or_else(|| Status::invalid_argument("UpdateVectors is missing"))?;
 
         update_vectors(
-            UncheckedTocProvider::new_unchecked(&self.toc),
+            StrictModeCheckedInternalTocProvider::new(&self.toc),
```
timvisee, line 261: I'm not super happy about this type name because it is so huge. But I do see that it's not strictly related to this PR.

Let's not bother with it in this PR and keep it like this for now.

tests/consensus_tests/test_strict_mode.py
```diff
@@ -116,3 +116,84 @@ def test_strict_mode_upsert_local_shard(tmp_path: pathlib.Path):
...
+        "max_collection_payload_size_bytes": 8000,
+    }
+    create_collection(peer_urls[0], collection=COLLECTION_NAME, shard_number=4, replication_factor=N_REPLICAS, strict_mode=strict_mode)
+
+    wait_collection_exists_and_active_on_all_peers(collection_name=COLLECTION_NAME, peer_api_uris=peer_urls)
+
+    # Insert points into leader
+    for i in range(10):
+        upsert_random_points(peer_urls[0], 100, collection_name=COLLECTION_NAME, offset=i*100)
+        time.sleep(5)  # TODO: Remove
```
timvisee, line 136: Can we remove it now (and others)? :smiley:

JojiiOfficial: This is related to #5669 as well and can be removed when the payload storage has been migrated to mmap

PR #5686, opened 2024-12-20
[Strict Mode] max upsert payload size

JojiiOfficial:
Limitation of payload size in a single request. Doesn't take the update by filter into account.

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -726,6 +726,7 @@ pub struct StrictModeConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub max_collection_payload_size_bytes: Option<usize>,
 
+    /// Max payload size per request in bytes.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub upsert_max_payload_size: Option<usize>,
 }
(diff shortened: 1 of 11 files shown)


lib/segment/src/types.rs
```diff
@@ -1132,6 +1135,27 @@ impl Payload {
...
+
+        size
+    }
+}
+
+fn value_size(value: &Value) -> usize {
+    match value {
+        Value::Null => 0,
+        Value::Bool(_) => 1,
+        Value::Number(_) => 8,
```
timvisee, line 1154: Though it's effectively the same, I do prefer this:

```suggestion
        Value::Null => std::mem::size_of::<()>(),
        Value::Bool(_) => std::mem::size_of::<bool>(),
        Value::Number(_) => std::mem::size_of::<f64>(),
```

tests/openapi/test_strictmode.py
```diff
@@ -46,6 +46,43 @@ def strict_mode_enabled(collection_name) -> bool:
...
+
+    set_strict_mode(collection_name, {
+        "enabled": True,
+        "upsert_max_payload_size": 3,
+    })
+
+    search_fail = upsert_request({"key": "abc"})
+
+    assert "payload limit" in search_fail.json()['status']['error']
+    assert not search_fail.ok
```
timvisee, line 83: Based on this it only counts values, not fields.

Shouldn't we include the length of fields too? Possibly just by iterating over the map keys as well.

generall:
I would prefer to limit size of request instead, not a big fan of this per-field recursive calculation

JojiiOfficial:
Closing since not needed anymore.

PR #5687, opened 2024-12-22
use 64-bit dim-id for storing sparse vectors

generall:
Eventually we want to allow 64bit `DimId` for sparse vectors. This is one of the steps towards.

### DIFF
--- a/lib/blob_store/src/blob_store.rs
+++ b/lib/blob_store/src/blob_store.rs
@@ -8,7 +8,7 @@ use parking_lot::RwLock;
 
 use crate::bitmask::Bitmask;
 use crate::blob::Blob;
-use crate::config::{StorageConfig, StorageOptions};
+use crate::config::{Compression, StorageConfig, StorageOptions};
 use crate::page::Page;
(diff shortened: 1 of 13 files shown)


generall:
ToDo: do we need to regen storage compatibility?

generall:
Also todo: maybe instead of bincode we do @xzfc 's bitpacking?

generall:
Another ToDo: maybe make compression in blob_storage optional, and rely on the "bitpacking" only

generall:
I haven't figure out how to replace bincode completely. But compression seems to work

agourlay:
I will run some benchmarks with realistic data to see how delta+binpacking perform here.

The storage compatibility will have to be regen also.

agourlay:
My criterion benchmark of the storage layer show that the new implementation is slightly faster than the compression based for reads and writes.

I will push a sanitized version of the benchmark which does not expose the internal compression scheme.

agourlay:
I will regen the storage compat for `v1.12.5` and merge right after to not block other PRs.

PR #5688, opened 2024-12-23
bitpacking: add tango benchmark

xzfc:
This PR adds a `tango` variant of the `bitpacking` benchmark. The logic is identical to the existing `criterion` benchmark.

Unlike `criterion`, `tango` runs two versions of the benchmark at once. So, background system load, CPU temperature, throttling and scaling mechanisms affect both versions equally, making it easier to compare e.g. performance of two commits.

https://www.bazhenov.me/posts/paired-benchmarking/


generall:
Please add an instruction of how to run tango benchmarks

generall (approved):

PR #5704, opened 2024-12-28
Separate file for compressed links

generall:
Depends on https://github.com/qdrant/qdrant/pull/5700 

1. Fixes `on_disk` parameter for loading of the HNSW links
2. Always loads mmaps and change `populate` if the user wants `on_disk` experience
3. Decide compression based on file name, rather than guessing from version.

Notes for the future: 

- I would prefer to have completely separate structures for compressed and non-compressed implementations, probably even allow those structures to own data to be completely independent. Current version works, but with internal enums the implementation looks very complicated and hard to understand. Any further extension (e.g. compression v2) will make the code unreadable.

- `graph_links.rs` contains a number of structures and enums and is hard to navigate, in future it should be split into multiple files.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_layers.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers.rs
@@ -317,7 +317,7 @@ impl GraphLayers {
     #[cfg(feature = "testing")]
     pub fn compress_ram(&mut self) {
         use crate::index::hnsw_index::graph_links::GraphLinksConverter;
-        assert!(!self.links.compressed() && !self.links.on_disk());
+        assert!(!self.links.compressed());
         let dummy = GraphLinksConverter::new(Vec::new(), false, 0, 0).to_graph_links_ram();
         let links = std::mem::replace(&mut self.links, dummy);
         self.links = GraphLinksConverter::new(links.into_edges(), true, self.m, self.m0)
@@ -429,6 +429,7 @@ mod tests {
     }
 
     #[rstest]
+    #[case::uncompressed((false, false))]
     #[case::converted((false, true))]
     #[case::compressed((true, false))]
     fn test_save_and_load(#[case] (compressed, converted): (bool, bool)) {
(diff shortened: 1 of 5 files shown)


lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -453,33 +489,22 @@ enum GraphLinksEnum {
...
 }
 
 impl GraphLinks {
-    pub fn load_from_file(path: &Path, on_disk: bool) -> OperationResult<Self> {
-        let file = OpenOptions::new()
-            .read(true)
-            .write(false)
-            .create(false)
-            .open(path)?;
-        if on_disk {
```
generall, line 472: This looks like a blunder

xzfc:
Yeah, I've missed that when replacing `in_ram` with `on_disk`, negating it through the entire code.

I wonder, whether "Invert Boolean﻿" refactoring is available in RustRover...

lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -429,6 +429,7 @@ mod tests {
     }
 
     #[rstest]
+    #[case::uncompressed((false, false))]
```
generall, line 432: @xzfc this parameter doesn't seem to be used.

lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -301,7 +317,7 @@ impl GraphLayers {
     #[cfg(feature = "testing")]
     pub fn compress_ram(&mut self) {
         use crate::index::hnsw_index::graph_links::GraphLinksConverter;
-        assert!(!self.links.compressed() && !self.links.on_disk());
```
xzfc, line 304: Note: I've removed this check because the mmap+populate change breaks hnsw_search_graph.

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -181,13 +189,22 @@ impl HNSWIndex {
                 )
             };
 
+            let do_convert = LinkCompressionExperimentalSetting::from_env().convert_existing;
+
+            let graph_links_path = if !is_links_compressed && do_convert {
+                GraphLayers::convert_to_compressed(&graph_path, &regular_links, &compressed_links)?;
```
xzfc, line 195: @generall
There were an error here (`graph_links_path` == `graph_links.bin` vs `graph_path` == `graph.bin`). It did not caught by unit tests, but I got it testing manually by running qdrant with `__QDRANT_COMPRESSED_LINKS=2`.

I think these indirections with filename passing are bulky and error-prone, so I made an alternative approach here: #5749.

Issue #5451, opened 2024-11-14
Cannot change `.qdrant-initialized` destination


miguelaeh:
It looks like one can change the storage and snapshot directory path but not the `.qdrant-initialized` file path.

## Current Behavior

The `.qdrant-initialized` is generated on the working directory

## Steps to Reproduce

1. Run the qdrant binary
2. List the files on the working directory

## Expected Behavior

I would like to configure the destination of that file, just like snapshots and storage directories. 

## Context (Environment)

Ubuntu 22.04

## Detailed Description

A configuration option just like `storage_path` or `snapshots_path`

miguelaeh:
Hi @generall ,

I embedded qdrant in a desktop application and the initialized file gets created from where the app is called instead of in a consistent directory.

jayantpranjal0:
@generall could you look at my PR?

PR #5714, opened 2024-12-31
Added feature to configure file path of .qdrant-initialized file

jayantpranjal0:
Feat: #5451

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -9,6 +9,8 @@ log_level: INFO
 #     log_file: path/to/log/file.log
 #     log_level: INFO
 
+init_file_path: .qdrant-initialized
+
 storage:
   # Where to store all the data
   storage_path: ./storage
--- a/src/startup.rs
+++ b/src/startup.rs
@@ -5,13 +5,12 @@ use std::panic;
 use std::path::PathBuf;
 
 use crate::common::error_reporting::ErrorReporter;
+use crate::settings::Settings;
 
-const DEFAULT_INITIALIZED_FILE: &str = ".qdrant-initialized";
-
-fn get_init_file_path() -> PathBuf {
+fn get_init_file_path(settings: &Settings) -> PathBuf {
     std::env::var("QDRANT_INIT_FILE_PATH")
         .map(PathBuf::from)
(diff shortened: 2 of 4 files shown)


generall:
logic here https://github.com/qdrant/qdrant/blob/master/tools/entrypoint.sh should be updated. And overall we would need to check if this change is backward compatible. 

Overall, taking into account that the original version allowed env to overwrite the path and questionable usefulness of the whole entrypoint script, I have doubts if investing more time into this config is worth the effort

config/config.yaml
```diff
@@ -9,6 +9,8 @@ log_level: INFO
 #     log_file: path/to/log/file.log
 #     log_level: INFO
 
+init_file_path: ./.qdrant-initialized
```
timvisee, line 12:
```suggestion
init_file_path: .qdrant-initialized
```

src/startup.rs
```diff
@@ -5,13 +5,10 @@ use std::panic;
...
 
-const DEFAULT_INITIALIZED_FILE: &str = ".qdrant-initialized";
-
-fn get_init_file_path() -> PathBuf {
-    std::env::var("QDRANT_INIT_FILE_PATH")
-        .map(PathBuf::from)
-        .unwrap_or_else(|_| DEFAULT_INITIALIZED_FILE.into())
+fn get_init_file_path(settings: &Settings) -> PathBuf {
+    PathBuf::from(&settings.init_file_path)
 }
```
timvisee, line 12: Let's prefer the `QDRANT_INIT_FILE_PATH` here over the configuration file if set.

That'll ensure backwards compatibility.

timvisee (changes requested):
I agree with @generall.

We'd somehow have to make `entrypoint.sh` aware of the configured path. That will not be trivial.

I did leave some minor suggestions below, but that will not resolve the above problem.

jayantpranjal0:
I agree that it wouldn't be trivial to make entrypoint.sh aware of configured init file path, but entrypoint.sh would be working as it was supposed to be working before this change, since most priority in configuration for init_file_path is given to the enviroment variable which is set in the script.

jayantpranjal0:
@timvisee @generall how should I move forward with this issue?

PR #5716, opened 2025-01-02
Disable default-feature for config crate

agourlay:
We only need to parse `yaml` files therefore we can avoid all the dependencies required for other formats.

https://docs.rs/crate/config/0.15.4/features


timvisee (approved):
Thanks! :pray:

KShivendu (approved):

ffuugoo (approved):
Best PR of 2025! 🤯🎉

PR #5717, opened 2025-01-02
Bump rocksdb 0.23.0

agourlay:
Test driving https://github.com/rust-rocksdb/rust-rocksdb/releases/tag/v0.23.0


agourlay:
More work is required to install the missing packages in all workflows.
Otherwise it compiles for me locally :tm:

PR #5728, opened 2025-01-03
Optimize consensus loop

ffuugoo:
This PR fixes two non-critical performance bugs in consensus loop:

<details>
<summary>Node that receives constant stream of consensus updates will trigger excessive Raft elections</summary>

Our consensus loop is structured in the following way:
- [take messages from Raft message queue, and propagate them into Raft node][advance-node]
- [tick node _as many times as necessary_ since last tick][tick]
- [process `on_ready`][process-on-ready]

There's a corner case with this loop structure:
1. Imagine we have a node that receives a constant stream of updates. E.g., it's a new node, or it was offline and needs to catch up with the cluster.
2. Such node receives a constant stream of updates from the leader, so on each iteration of consensus loop we would propagate a full batch of messages to Raft node, and then process `on_ready` for all these messages.
3. Processing `on_ready` will often take more time, than election timeout.
4. So, on next iteration of the loop, we will tick the node too many times and trigger leader election.
5. And this may happen multiple times while the node is catching up with consensus. It's not critical for correctness or consistency, but it slows consensus down a lot.

To fix this issue, we check if any new Raft messages were received since last iteration of the loop, and limit the number of times we tick Raft node.

From my tests, Raft crate is not super reliant on ticks: it works even if you never tick any node, as long as you established a leader and all nodes are online, so this should be safe.

[advance-node]: https://github.com/qdrant/qdrant/blob/ffda0b90c8c44fc43c99adab518b9787fe57bde6/src/consensus.rs#L496
[tick]: https://github.com/qdrant/qdrant/blob/ffda0b90c8c44fc43c99adab518b9787fe57bde6/src/consensus.rs#L509
[process-on-ready]: https://github.com/qdrant/qdrant/blob/ffda0b90c8c44fc43c99adab518b9787fe57bde6/src/consensus.rs#L509
</details>

<details>
<summary>`try_sync_local_state` is not always triggered reliably</summary>

While working on the first bug in the PR, I discovered that sometimes the `try_sync_local_state` might not be triggered for multiple seconds on a node.

1. We [trigger `try_sync_local_state`][sync-local-state], if during an iteration of consensus loop, no messages were received and no `Ready` state produced by Raft node.
2. This is a valid approach, but it's overly strict.
3. Raft nodes exchange heartbeats every 2 ticks. So, every 2 iterations of consensus loop, a heartbeat message is received and `Ready` state is produced to send a heartbeat-response message.
4. And I've seen cases, when consensus loop would get into a "lockstep" with heartbeats: every iteration of the loop would take **2 ticks** (not sure why exactly, maybe because `on_ready` would take a bit too long to process, or due to timeouts not being perfectly precise), and so for multipe iterations of the loop, *each* iteration would receive a message and/or produce `Ready` state, and `try_sync_local_state` might not be triggered *for seconds*.

To fix this, I reworked how we trigger `try_sync_local_state:
1. `on_ready` now returns `is_idle` flag.
2. Raft node is "idle", if during `on_ready` Raft state was not updated. E.g., no new entries appended to WAL, no entries committed, no Raft snapshot applied. The only idle operation during `on_ready` is sending outgoing Raft messages.
3. We trigger `try_sync_local_state`, if the node was idle for a few consecutive iterations of consensus loop.

This way, we can reliably detect when node caught up with consensus (and heartbeats don't affect this check).

[sync-local-state]: https://github.com/qdrant/qdrant/blob/6260eb7d74e8c5303b2e346eb2d3367d6345d698/src/consensus.rs#L507-L518
</details>

### All Submissions:

### New Feature Submissions:

7. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
8. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ homepage = "https://qdrant.tech/"
 repository = "https://github.com/qdrant/qdrant"
 license = "Apache-2.0"
 edition = "2021"
-rust-version = "1.77"
+rust-version = "1.80"
 default-run = "qdrant"
 
 [lints]
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -14,7 +14,6 @@ use collection::shards::channel_service::ChannelService;
 use collection::shards::shard::PeerId;
 #[cfg(target_os = "linux")]
 use common::cpu::linux_high_thread_priority;
-use common::defaults;
 use raft::eraftpb::Message as RaftMessage;
 use raft::prelude::*;
 use raft::{SoftState, StateRole, INVALID_ID};
@@ -208,16 +207,6 @@ impl Consensus {
             ..Default::default()
         };
         raft_config.validate()?;
-        let op_wait = defaults::CONSENSUS_META_OP_WAIT;
-        // Commit might take up to 4 ticks as:
-        // 1 tick - send proposal to leader
-        // 2 tick - leader sends append entries to peers
-        // 3 tick - peer answers leader, that entry is persisted
-        // 4 tick - leader increases commit index and sends it
-        if 4 * Duration::from_millis(config.tick_period_ms) > op_wait {
-            log::warn!("With current tick period of {}ms, operation commit time might exceed default wait timeout: {}ms",
-                 config.tick_period_ms, op_wait.as_millis())
-        }
         // bounded channel for backpressure
         let (sender, receiver) = tokio::sync::mpsc::channel(config.max_message_queue_size);
         // State might be initialized but the node might be shutdown without actually syncing or committing anything.
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -3,7 +3,7 @@ use std::str::FromStr;
 use std::sync::{mpsc, Arc};
 use std::thread::JoinHandle;
 use std::time::{Duration, Instant};
-use std::{fmt, thread};
+use std::{cmp, fmt, thread};
 
 use anyhow::{anyhow, Context as _};
 use api::grpc::dynamic_channel_pool::make_grpc_channel;
@@ -55,6 +55,7 @@ pub struct Consensus {
     /// ToDo: Make if many
     config: ConsensusConfig,
     broker: RaftMessageBroker,
+    raft_config: Config,
 }
 
 impl Consensus {
@@ -285,6 +286,7 @@ impl Consensus {
             runtime,
             config,
             broker,
+            raft_config,
         };
 
         if !state_ref.is_new_deployment() {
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -11,7 +11,7 @@ homepage = "https://qdrant.tech/"
 repository = "https://github.com/qdrant/qdrant"
 license = "Apache-2.0"
 edition = "2021"
-rust-version = "1.77"
+rust-version = "1.80"
```
ffuugoo, line 14: Required for [`div_duration_f32`] that I now use in consensus loop.

[`div_duration_f32`]: https://doc.rust-lang.org/std/time/struct.Duration.html#method.div_duration_f32

src/consensus.rs
```diff
@@ -833,29 +834,31 @@ impl Consensus {
     }
 
     /// Returns `true` if consensus should be stopped, `false` otherwise.
```
timvisee, line 836: Let's update the docs

timvisee (commented):
I did go through this, but I find it quite hard to keep track on what all the changes do exactly.

I agree that testing this on choas tesing for a while is a good idea, just to be sure there's no obvious problem. We can start testing after releasing 1.13.

src/consensus.rs
```diff
@@ -486,149 +486,150 @@ impl Consensus {
             }
         }
 
+        // If this is the origin peer of the cluster, try to add origin peer to consensus
+        if let Err(err) = self.try_add_origin() {
```
generall, line 490: Why didn't we put `try_add_origin` in here in the first version? Is that related to refactoring, or just better?

ffuugoo: IIRC, when I added `try_add_origin`, I did not yet implement the "instantly self-elect itself as a leader" feature (right above `try_add_origin` call), so I had to put it inside the consensus loop.

ffuugoo:
Qdrant 1.13 is released, so we can review and merge this now.

ack @generall @timvisee

src/consensus.rs
```diff
@@ -486,149 +486,150 @@ impl Consensus {
...
-            let updates = self.advance_node(previous_tick, tick_period)?;
+            // Wait (for up to `tick_period`) for incoming client requests and Raft messages
+            let raft_messages = self.advance_node(tick_period)?;
 
-            let mut elapsed = previous_tick.elapsed();
-
-            while elapsed > tick_period {
-                self.node.tick();
+            // Calculate how many ticks passed since the last one
+            let elapsed_ticks = previous_tick.elapsed().div_duration_f32(tick_period) as u32;
```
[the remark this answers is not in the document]
ffuugoo: I think it's safe enough to ignore (like, don't put nonsensical values in the config, duh). And there's `#[validate(range(min = 1))]` annotation on `ConsensusConfig::tick_period_ms` field already.

src/consensus.rs
ffuugoo: Done.

src/consensus.rs
```diff
@@ -486,149 +477,154 @@ impl Consensus {
...
+                let max_elapsed_ticks =
+                    cmp::max(1, self.raft_config.election_tick.saturating_sub(5));
 
-            if self.node.has_ready() {
-                // Persist AND apply changes, which were committed in the Raft State Machine
-                let stop_consensus = self.on_ready()?;
+                cmp::min(elapsed_ticks, max_elapsed_ticks as u32)
+            } else {
+                elapsed_ticks
+            };
```
timvisee, line 519: Should we have an upper bound? 

I mean, what if this node has been stuck (somehow) for a long while and we still report just 15 ticks? In that case I do expect new election.

Or am I missing an important detail here?

src/consensus.rs
```diff
@@ -832,30 +828,34 @@ impl Consensus {
...
-    fn on_ready(&mut self) -> anyhow::Result<bool> {
+    /// Returns two boolean flags: `stop_consensus` and `is_idle`.
+    /// If `stop_consensus` is true, then we should exit consensus loop and stop consensus.
+    /// If `is_idle` is true, it means that no on-disk state was updated during this `on_ready` call.
+    fn on_ready(&mut self) -> anyhow::Result<(bool, bool)> {
         if !self.node.has_ready() {
             // No updates to process
-            return Ok(false);
+            return Ok((false, true));
         }
```
timvisee, line 838: Are we idle if we're not ready?

timvisee (approved):
There's a **lot** going on here. Though I don't see any obvious problems I'm not confident I got all possible edge cases here in these changes.

As said before, let's test this on chaos testing for a while. We can merge this once we've released 1.13.1 which will be in one or two days.

src/consensus.rs
ffuugoo: The election will be triggered on the next few iterations of consensus loop then.

src/consensus.rs
ffuugoo: Yes, we are.

generall (approved):

ffuugoo:
Rebased on `dev` (just in case), will merge once green ✅

PR #5729, opened 2025-01-03
Consider filter cardinality in write ratelimiter for update-by-filter requests

JojiiOfficial:


tests/openapi/test_strictmode.py
```diff
@@ -852,4 +852,45 @@ def test_strict_mode_write_rate_limiting(collection_name):
...
+        api='/collections/{collection_name}/points/payload',
+        method="PUT",
+        path_params={'collection_name': collection_name},
+        body={
+            "payload": {"City": "Not London"},
+            "filter": {"must": [{"key": "City", "match": {"value": "London"}}]}
+        }
+    )
+    assert response.status_code == 429
+    assert "Rate limiting exceeded: Write rate limit exceeded, retry later" in response.json()['status']['error']
```
generall, line 896: Considering that we are assigning different rate limits to different operations, this information should be represented in the response error. Something like `operation require X units, but Y availble`, otherwise it would be confusing to debug

lib/collection/src/operations/mod.rs
```diff
@@ -180,6 +180,40 @@ impl CollectionUpdateOperations {
             Self::FieldIndexOperation(_) => (),
         }
     }
+
+    pub fn filter(&self) -> Option<&Filter> {
```
generall, line 184: We already have this https://github.com/qdrant/qdrant/blob/c95b2448d9231ac4488798af30d718965766d0d8/lib/collection/src/operations/operation_effect.rs#L28

Which might be both: a better fit (cause also includes points, which we need to count as individual updates as well), and require some upgrade, cause we might not want to clone `Filter`

generall: This operation effect area is only used in ProxyShard, but ProxyShard itself is not used anywhere, so, if needed, we can change some stuff in the signatures if needed

generall (approved):

agourlay (approved):
Nice cleanup on the clones :+1:

PR #5737, opened 2025-01-06
Fix unaligned reads for graph links offsets

xzfc:
The change #5491 (included in version v1.12.5) introduced the following crash when reading some old snapshots.

Repro:
```
PUT collections/test/snapshots/recover?wait=true
{ "location": "https://snapshots.qdrant.io/qdrant-web-site-docs-2024-04-05.snapshot" }

POST /collections/test/points/recommend
{ "positive": [0], "limit": 10 }
```
Results in the following crash in [`graph_links.rs:490`](https://github.com/qdrant/qdrant/blob/v1.12.5/lib/segment/src/index/hnsw_index/graph_links.rs#L490):
> 'Service internal error: 1 of 1 read operations failed:
>  Service internal error: task 128 panicked with message "called `Option::unwrap()` on a `None` value"

Reason: old versions of qdrant (before v1.8.2) did store graph links offsets in a non-aligned array, but the new code assumes it's aligned. For the reference, non-faulty code snippet in `v0.12.4` uses `u64::from_ne_bytes`.
https://github.com/qdrant/qdrant/blob/5b578c4f34188f0474f901e49d4726213596433d/lib/segment/src/index/hnsw_index/graph_links.rs#L517-L524

This PR replaces usage of `u64` with `zerocopy::native_endian::U64` which doesn't require alignment. This is a fix for the `master` branch, there will be a separate PR for the `dev` branch.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_links.rs
+++ b/lib/segment/src/index/hnsw_index/graph_links.rs
@@ -9,6 +9,7 @@ use common::types::PointOffsetType;
 use common::zeros::WriteZerosExt as _;
 use memmap2::Mmap;
 use memory::{madvise, mmap_ops};
+use zerocopy::native_endian::U64 as NativeU64;
 use zerocopy::{AsBytes, FromBytes, FromZeroes};
 
 use crate::common::operation_error::{OperationError, OperationResult};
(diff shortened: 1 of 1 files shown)


generall:
please rebase to latest dev, there are fixes for poetry version

xzfc:
This is a PR for `master`. The code in `dev` diverges significantly, so it would require a different fix/PR.

Perhaps, we need to cherry-pick that poetry fix to the `master`.

generall:
If we are going to patch release, it will likely be from dev, I don't think we need changes in master

PR #5745, opened 2025-01-06
Impletement a special type of access to collection suitable for serverless

generall:
A special type of collection access `prw` - Points Read Write, which forbids:

- cluster info
- snapshots
- payload indexes

### DIFF
--- a/lib/storage/src/rbac/mod.rs
+++ b/lib/storage/src/rbac/mod.rs
@@ -72,6 +72,11 @@ pub enum CollectionAccessMode {
     /// Read and write access to a collection, with some restrictions.
     #[serde(rename = "rw")]
     ReadWrite,
+
+    /// Points read and write - access to update and modify points in the collection,
+    /// but not snapshots or payload indexes.
+    #[serde(rename = "prw")]
+    PointsReadWrite,
 }
 
 #[derive(Serialize, Deserialize, PartialEq, Clone, Debug)]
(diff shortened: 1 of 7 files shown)


n0x29a (approved):

agourlay:
There is a failing test with a super weird mention of Rocksdb data.

lib/storage/src/rbac/mod.rs
```diff
@@ -230,6 +245,8 @@ pub struct AccessRequirements {
     pub manage: bool,
     /// If true, the access should be not limited by a payload restrictions.
     pub whole: bool,
+    /// Require access to collection extras, like snapshots, payload indexes, cluster info.
+    pub extras: bool,
```
agourlay, line 249: This `extra` field feels a bit like a shortcut to cram things together.
Would it make sense to have dedicated flags for `snapshots`, `payload_index` and `cluster_info`?

generall: I though about it, but if I we do it like this, the next question would be - why not do it for everything?
This about it from the other side: we don't want to block a list of things, we want to allow access to points only. If we would have some new entity inside the collection, we won't need to come up with a new name, we just put it under extras.

Anyway, this is an internal logic, so I propose to keep it like this until we would need to change it.

agourlay (approved):

PR #5754, opened 2025-01-07
[Strict mode] limits for filter and conditions

JojiiOfficial:
Adds 2 new limits to strict mode:
- Max amount of total conditions in the filter
- Max elements in a condition

### DIFF
--- a/lib/collection/src/operations/verification/mod.rs
+++ b/lib/collection/src/operations/verification/mod.rs
@@ -108,22 +108,27 @@ pub trait StrictModeVerification {
         let check_filter = |filter: Option<&Filter>,
                             allow_unindexed_filter: Option<bool>|
          -> Result<(), CollectionError> {
-            if let Some(read_filter) = filter {
-                if allow_unindexed_filter == Some(false) {
-                    if let Some((key, schemas)) = collection.one_unindexed_key(read_filter) {
-                        let possible_schemas_str = schemas
-                            .iter()
-                            .map(|schema| schema.to_string())
-                            .collect::<Vec<_>>()
-                            .join(", ");
+            let Some(filter) = filter else {
+                return Ok(());
+            };
 
-                        return Err(CollectionError::strict_mode(
+            // Check for filter indices
+            if allow_unindexed_filter == Some(false) {
+                if let Some((key, schemas)) = collection.one_unindexed_key(filter) {
+                    let possible_schemas_str = schemas
+                        .iter()
+                        .map(|schema| schema.to_string())
+                        .collect::<Vec<_>>()
+                        .join(", ");
+
+                    return Err(CollectionError::strict_mode(
                             format!("Index required but not found for \"{key}\" of one of the following types: [{possible_schemas_str}]"),
                             "Create an index for this key or use a different filter.",
                         ));
-                    }
                 }
             }
+
+            check_filter_limits(&filter, strict_mode_config)?;
 
             Ok(())
         };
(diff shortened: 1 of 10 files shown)


tests/openapi/test_strictmode.py
```diff
@@ -852,4 +851,81 @@ def test_strict_mode_write_rate_limiting(collection_name):
             }
         )
 
-        assert response.ok, "Rate limiting should be disabled now"
\ No newline at end of file
+        assert response.ok, "Rate limiting should be disabled now"
+
```
agourlay, line 855: Can you please add a basic test with a nested filter for sanity check

tests/openapi/helpers/helpers.py
```diff
@@ -43,6 +43,7 @@ def request_with_validation(
         query_params: dict = None,
         body: dict = None
 ) -> requests.Response:
+    print(body)
```
agourlay, line 46: :broom:

lib/segment/src/types.rs
```diff
@@ -725,6 +725,14 @@ pub struct StrictModeConfig {
     /// Max size of a collections payload storage in bytes
     #[serde(skip_serializing_if = "Option::is_none")]
     pub max_collection_payload_size_bytes: Option<usize>,
+
+    /// Max conditions a filter can have at max.
```
agourlay, line 729:
```suggestion
    /// Max conditions a filter can have.
```

lib/segment/src/types.rs
```diff
@@ -2060,6 +2090,18 @@ impl FieldCondition {
...
+
+    fn input_size(&self) -> usize {
+        if self.r#match.is_none() {
+            return 0;
+        }
+
+        match self.r#match.as_ref().unwrap() {
+            Match::Any(match_any) => match_any.any.len(),
+            Match::Except(match_except) => match_except.except.len(),
+            _ => 0,
```
agourlay, line 2102: can we have an exhaustive match instead please

agourlay (commented):
Great job, I have left some minor comments :+1:

tests/openapi/test_strictmode.py
JojiiOfficial: Done: https://github.com/qdrant/qdrant/pull/5754/files#diff-70aaa118e65b662af3397f4a9fb93751d947b331c0e9bc441dfaf49531dfbc09R934

lib/collection/src/operations/verification/mod.rs
```diff
@@ -158,6 +163,33 @@ pub trait StrictModeVerification {
...
+    filter: &Filter,
+    strict_mode_config: &StrictModeConfig,
+) -> Result<(), CollectionError> {
+    if !check_custom(
+        || Some(filter.total_conditions_count()),
+        strict_mode_config.filter_max_conditions,
+    ) {
+        return Err(CollectionError::strict_mode(
+            "Filter condition limit reached",
+            "Reduce the amount of conditions of your filter.",
```
agourlay, line 176: I think the error reporting might frustrating for users.

They are told to reduce the number of conditions but there is no exact number to target.

Would it be difficult to return something clearer here?

lib/collection/src/operations/verification/mod.rs
```diff
@@ -158,6 +163,33 @@ pub trait StrictModeVerification {
...
+        ));
+    }
+
+    if !check_custom(
+        || Some(filter.max_condition_input_size()),
+        strict_mode_config.condition_max_size,
+    ) {
+        return Err(CollectionError::strict_mode(
+            "Condition size limit reached",
+            "Reduce the size of your condition.",
```
agourlay, line 186: same point about error reporting :)

lib/collection/src/operations/verification/mod.rs
```diff
@@ -199,6 +238,21 @@ pub(crate) fn check_limit_opt<T: PartialOrd + Display>(
...
+) -> bool {
+    let Some(limit) = limit else {
+        return true;
+    };
+
+    let Some(value) = value_fn() else {
+        return true;
+    };
+
+    value <= limit
```
timvisee, line 253: We _can_ simplify this to:

```rust
value_fn()
    .zip(limit)
    .is_none_or(|(value, limit)| value <= limit)
```

timvisee (approved):
Thanks for adding the tests!

lib/collection/src/operations/verification/mod.rs
JojiiOfficial: Note that the `value_fn()` won't be evaluated lazily anymore then. For our current implementation this doesn't change anything but I think it is nice to have a lazily evaluated function that only checks when strict mode is configured.

PR #5756, opened 2025-01-08
Fix GPU sorting bug for low-end devices

IvanPleshkov:
Fix bitonic sort bug. No test is required, this bug was found on GPU CI on an emulated GPU device:
https://github.com/qdrant/qdrant/actions/runs/12658322098/job/35274994845

### DIFF
--- a/lib/segment/src/index/hnsw_index/gpu/shaders/shared_buffer.comp
+++ b/lib/segment/src/index/hnsw_index/gpu/shaders/shared_buffer.comp
@@ -58,7 +58,9 @@ void _bitonic_cmp(uint offset, uint count, uint b_i, uint b_j) {
     }
 }
 
+// Sort the shared buffer in place using bitonic sort
 void sort(uint offset, uint count) {
+    // To increase performance, sort blocks of size `SUBGROUP_SIZE` using bubble sort.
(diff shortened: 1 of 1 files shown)


lib/segment/src/index/hnsw_index/gpu/shaders/shared_buffer.comp
```diff
@@ -78,7 +80,8 @@ void sort(uint offset, uint count) {
     }
     subgroupMemoryBarrierShared();
 
-    for (uint k = 64; (k >> 1) < count; k <<= 1) {
+    // Start bitonic sort. Start from `k = 2 * SUBGROUP_SIZE` instead of `1` because we have already sorted blocks of size `SUBGROUP_SIZE`.
+    for (uint k = 2 * SUBGROUP_SIZE; (k >> 1) < count; k <<= 1) {
```
timvisee, line 84: `k` changed from `64` to `2 * SUBGROUP_SIZE`.

With a default subgroup size of 32 that was valid. But the new approach parameterized it in case the subgroup size is different.

This means that it's effectively the same on 32-group hardware. But it got improved on hardware having a different sub group size.

PR #5763, opened 2025-01-09
Add `VectorName` type alias

xzfc:
This PR replaces usages like `vector_name: &str` to use the following type aliases instead of raw `str` or `String`:

```rust
/// Name of a vector
pub type VectorName = str;
/// Name of a vector (owned variant)
pub type VectorNameBuf = String;
```

The naming is an analogy to `Path`/`PathBuf`.

Later these aliases could be replaced with newtypes. To aid with that, this PR updates the code that assumes strings to use more generic methods, e.g., `name.to_string()` → `name.to_owned()`.

### DIFF
--- a/lib/segment/src/segment/search.rs
+++ b/lib/segment/src/segment/search.rs
@@ -12,6 +12,8 @@ use crate::data_types::vectors::VectorStructInternal;
 #[cfg(feature = "testing")]
 use crate::entry::entry_point::SegmentEntry;
 #[cfg(feature = "testing")]
+use crate::types::VectorName;
+#[cfg(feature = "testing")]
 use crate::types::{Filter, SearchParams};
 use crate::types::{ScoredPoint, WithPayload, WithVector};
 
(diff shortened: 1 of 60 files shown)


lib/segment/src/segment/search.rs
```diff
@@ -91,7 +91,7 @@ impl Segment {
     #[cfg(feature = "testing")]
     pub fn search(
         &self,
-        vector_name: &str,
+        vector_name: &crate::types::VectorName,
```
timvisee, line 94: Used the qualified path here on purpose?

timvisee (approved):

agourlay:
Not against the change, just curious about the vision here :)

Do we want to have specific utility functions for our Vector names?

lib/segment/src/segment/search.rs
xzfc: Moved to the top of the file for consistency. (it was here because it's behind `#[cfg]`)

xzfc:
@agourlay

My intent is to make the code clearer for readers. Initially, I've stumbled into this:
```rust
vector_storages: HashMap<String, VectorStorageEnum>,
```
and it was not immediately obvious to me that `String` is a vector name.

---

As for newtypes, I've used them to find all the relevant places to change.
But I did not include to make it easy of merging and reviewing; and I'm not sure if we want them yet.
You might check not-yet-cleaned-up version in the `vector-name-newtype` branch if you are interested.

PR #5766, opened 2025-01-09
Fix Clippy unused import on Windows

agourlay:
Spotted on CI.

```
warning: unused import: `AccessRequirements`
 --> src\common\telemetry_ops\memory_telemetry.rs:4:29
  |
4 | use storage::rbac::{Access, AccessRequirements};
  |                             ^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```

### DIFF
--- a/src/common/telemetry_ops/memory_telemetry.rs
+++ b/src/common/telemetry_ops/memory_telemetry.rs
@@ -1,7 +1,12 @@
 use schemars::JsonSchema;
 use segment::common::anonymize::Anonymize;
 use serde::Serialize;
-use storage::rbac::{Access, AccessRequirements};
+use storage::rbac::Access;
+#[cfg(all(
+    not(target_env = "msvc"),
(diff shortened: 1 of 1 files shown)


src/common/telemetry_ops/memory_telemetry.rs
```diff
@@ -1,7 +1,12 @@
 use schemars::JsonSchema;
 use segment::common::anonymize::Anonymize;
 use serde::Serialize;
-use storage::rbac::{Access, AccessRequirements};
+use storage::rbac::Access;
+#[cfg(all(
+    not(target_env = "msvc"),
+    any(target_arch = "x86_64", target_arch = "aarch64")
+))]
+use storage::rbac::AccessRequirements;
```
timvisee, line 9: Alternatively we can move this import into the function itself, or use the fully qualified path there, because we only use it in a single place.

timvisee (approved):

PR #5767, opened 2025-01-09
Update config 0.15.4

agourlay:
This PR updates to config [0.15.4](https://github.com/rust-cli/config-rs/blob/main/CHANGELOG.md#change-log)

The most important changes is that the configuration keys are now case sensitive.

We are only impacted for the `SnapshotsStorageConfig` where our default configuration uses lower case.

The solution is to apply `#[serde(rename_all = "lowercase")]` to the enum as usual.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1270,13 +1270,13 @@ dependencies = [
 
 [[package]]
 name = "config"
-version = "0.14.1"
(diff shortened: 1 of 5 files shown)


timvisee (approved):
Thanks for the work! Happy to see the serde rename worked out.

PR #5771, opened 2025-01-10
Estimate payload size

JojiiOfficial:
Extend `Payload` type with a new function that estimates the payloads size. This will be in IO usage estimation for payload storage.

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -3,6 +3,7 @@ use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
 use std::fmt::{self, Display, Formatter};
 use std::hash::Hash;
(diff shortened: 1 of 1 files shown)


generall:
why do we need to estimate IO usage before we apply io operations?
I mean, can't blob storage report IO usage instead of us guessing what the IO will be by estimating payload size?

JojiiOfficial:
We can. I figured it to be cleaner to not put measurements too deep into the core structures. Also this implementation is more generic and can be used in other places too when we want to estimate a points payloads size.

If you prefer the storage to report the operations cost directly I'm fine with implementing this instead 👍🏻

generall:
> If you prefer the storage to report the operations cost directly I'm fine with implementing this instead

yeah, let's try this way first

PR #5773, opened 2025-01-10
Measure payload read IO

JojiiOfficial:
This PR introduces measuring read IO for payload index for several endpoints. Collected values are currently only reported in metrics/telemetry API. Per endpoint reporting, tests, distributed support and TODOs will be done in separate PRs to not make this even larger.

Since it's mostly function signature changes, splitting into multiple PRs without merging half baked things is hard. This is why I decided to create a larger PR that can be reviewed on per commit level.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -831,6 +831,7 @@ version = "0.1.0"
 dependencies = [
  "bitvec",
  "bustle",
+ "common",
  "criterion",
  "csv",
  "dataset",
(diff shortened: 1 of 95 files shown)


generall (approved):
ToDo in other PRs:
- measurement for payload storage reads inside Optimized query filter (aka nested fields)
- measurement for vector reads
- measurement for hnsw index reads
- measurement for payload index reads
- measurement for sparse vectors reads (and index)

PR #5780, opened 2025-01-13
Update storage compat 1.11 & 1.12

agourlay:
We want to make sure that we are compatible with the latest minor version and previous patch.

`v1.12.6-rocksdb+blob_store-pre_release` did not contain proper blob store data as the script did not activate the feature properly.

I fixed this in a new archive and also documented how to do it.

I validated that the archive uses the new blob store for payload storage and sparse vector storage.

blob-store-payload
blob-store-sparse


agourlay:
back to draft, I need to sort out something for the new blob store data

timvisee (approved):
:+1:

PR #5768, opened 2025-01-10
Experiment: enable bindgen-runtime for RocksDB

timvisee:
Potential improvement after <https://github.com/qdrant/qdrant/pull/5765>.

This might relax the build time requirements for RocksDB.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

PR #5785, opened 2025-01-13
Revert "Experiment: enable bindgen-runtime for RocksDB (#5768)"

KShivendu:
Github CI has starting going OOM all the time when generating coverage reports and we have seen that normal workflows are [also](https://github.com/qdrant/qdrant/actions/runs/12747175346/attempts/1?pr=5782) [failing](https://github.com/qdrant/qdrant/actions/runs/12745867106/attempts/1?pr=5770). I suspect it has something to do with recent RocksDB changes. 

Just trying to revert and see if it makes any difference. In the meantime, I'll try to find something to profile the coverage test command. 

TODO:
- Push 10 empty commits with `gcmsg "Trigger CI ${i}" --allow-empty` and see if any of them fail with 143 again.
- This will prove that revert works


KShivendu:
It's kinda better after reverting. It [almost completed](https://github.com/qdrant/qdrant/actions/runs/12748813062/job/35529836081) when it reached 100% RAM usage and got OOMed. Previously it got killed much earlier. But obviously reverting RocksDB is not something we'd want to do. I'll try once by removing/modifying other steps.

timvisee:
Do you still want to give this experiment a shot?

In the end I don't think that downgrading RocksDB is a solution.

KShivendu:
> downgrading RocksDB is a solution

Seems to be making a difference though. I didn't get any `143` exit failures on CI after reverting :thinking:

timvisee:
> Seems to be making a difference though. I didn't get any 143 exit failures on CI after reverting. Last 4 are runs are green. 🤔

Having a newer RocksDB version and compiling with with a coverage harness both consume more memory. I don't agree that downgrading RocksDB is a solution just to make compiling coverage tests work.

Testing locally shows a marginal difference in memory usage during compilation for both RocksDB versions.

KShivendu:
We decided to not revert rocksdb version since we have released it with recent versions of Qdrant. The CI failures are occasional and we plan to replace rocksdb soon.

timvisee:
> we plan to replace rocksdb soon.

RocksDB will stay in our code base for at least another year :face_holding_back_tears:

PR #5765, opened 2025-01-09
Update rocksdb 0.23.0

agourlay:
### All Submissions:

* [ ] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [ ] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [ ] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #5791, opened 2025-01-14
Remove unecessary docker build dependency

agourlay:
Those deps were introduced in https://github.com/qdrant/qdrant/pull/5765/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557.

But then partially reverted in https://github.com/qdrant/qdrant/pull/5768/files

The docker part was missing.


timvisee (approved):
Thanks! I missed it.

PR #5792, opened 2025-01-14
Resharding: fix abort on resharding up with replicas in active state

timvisee:
When resharding up, migrated replicas will be in active state.

This fixes a problem where aborting resharding failed because it did not expect the active state.

### All Submissions:

### DIFF
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -210,6 +210,10 @@ impl Collection {
             },
         }
 
+        shard_holder
+            .abort_resharding(resharding_key.clone(), force)
+            .await?;
+
         // Decrease the persisted shard count, ensures we don't load dropped shard on restart
         if resharding_key.direction == ReshardingDirection::Up {
             let mut config = self.collection_config.write().await;
@@ -235,8 +239,6 @@ impl Collection {
                 ShardingMethod::Custom => {}
             }
         }
-
-        shard_holder.abort_resharding(resharding_key, force).await?;
 
         Ok(())
     }
--- a/lib/collection/src/shards/shard_holder/resharding.rs
+++ b/lib/collection/src/shards/shard_holder/resharding.rs
@@ -344,10 +344,13 @@ impl ShardHolder {
         }
 
         // Remove new shard if resharding up
-        if direction == ReshardingDirection::Up {
+        if is_in_progress && direction == ReshardingDirection::Up {
             if let Some(shard) = self.get_shard(shard_id) {
                 match shard.peer_state(peer_id) {
-                    Some(ReplicaState::Resharding) => {
+                    // Valid replica states:
+                    // - in resharding state during migration transfer
+                    // - in active state after migration transfer
+                    Some(ReplicaState::Resharding | ReplicaState::Active) => {
                         log::debug!("removing peer {peer_id} from {shard_id} replica set");
                         shard.remove_peer(peer_id).await?;
                     }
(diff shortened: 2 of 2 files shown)


lib/collection/src/collection/resharding.rs
```diff
@@ -210,6 +210,10 @@ impl Collection {
             },
         }
 
+        shard_holder
+            .abort_resharding(resharding_key.clone(), force)
+            .await?;
+
```
timvisee, line 216: Additionally: abort before persisting the shard number on disk.

This way we don't persist wrong state if aborting itself fails.

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -344,51 +344,55 @@ impl ShardHolder {
...
+                        // Valid replica states:
+                        // - in resharding state during migration transfer
+                        // - in active state after migration transfer
+                        // - in partial state during replication
+                        // - in dead state during errors
+                        ReplicaState::Resharding
+                        | ReplicaState::Active
+                        | ReplicaState::Partial
+                        | ReplicaState::Dead => {}
+
```
timvisee, line 361: Question: maybe we want to allow any state if `force == true`?

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -344,51 +344,55 @@ impl ShardHolder {
         }
 
         // Remove new shard if resharding up
-        if direction == ReshardingDirection::Up {
+        if (is_in_progress || force) && direction == ReshardingDirection::Up {
```
ffuugoo, line 347: I don't think we can use `force` here. It will basically allow you to remove any shard, if you abort resharding with proper `reshard_key` and a `force` flag.

ffuugoo: The `force` is intended to allow aborting if we are at `ReadHashRingCommitted`/`WriteHashRingCommitted` stage.

timvisee: > It will basically allow you to remove any shard, if you abort resharding with proper reshard_key and a force flag.

Isn't that kind of the point. The reshard key should only match a new shard that is newly created with resharding, and it can therefore not be used to remove an arbitrary shard ID.

> The force is intended to allow aborting if we are at ReadHashRingCommitted/WriteHashRingCommitted stage.

I agree that that was the original intention. But I was looking at it from a 'will this help us resolve corrupt state' perspective. In that case I think allowing this as well will be useful.

What do you think?

lib/collection/src/shards/shard_holder/resharding.rs
timvisee: > I think, if the whole condition is under is_in_progress, we can allow removing replicas in any state.

Fine by me.

I figured this being a bit more strict will prevent problems if something else is interfering with the replicas (some user transfer potentially).

lib/collection/src/shards/shard_holder/resharding.rs
ffuugoo: > The reshard key should only match a new shard that is newly created with resharding, and it can therefore not be used to remove an arbitrary shard ID.

As far as I can tell, there's nothing in `abort_resharding` that prevents calling it with garbage `ReshardKey` (e.g., when resharding is not in progress or with incorrect `shard_id`). And it such case, we will just purge the whole shard no questions asked.

timvisee: You're right. With the current state in the PR it is possible to bypass the shard key check with the force flag.

I agree that isn't desirable so I'll adjust.

ffuugoo (approved):

Issue #3184, opened 2023-12-08
Validate unique name between dense & sparse vectors


agourlay:
## Current Behavior
It is currently possible to create a collection with a dense and sparse vector sharing the same name.

```json
 "vectors": {
        "image": {
            "size": 2,
            "distance": "Euclid"
        }
    },
    "sparse_vectors": {
        "image": {}
    }
```

The collection creation goes fine.

However it is impossible to add sparse vectors later on into the collection because our internal validation checks the dense vectors first.

It finds `image` registered as a dense vectors and rejects the point upload with a client side error because it is malformed.

## Steps to Reproduce

```python
collection_name = 'test_sparse_vector_validation'

def test_sparse_dense_vector_naming_validations():
    response = request_with_validation(
        api='/collections/{collection_name}',
        method="DELETE",
        path_params={'collection_name': collection_name},
    )
    assert response.ok

    response = request_with_validation(
        api='/collections/{collection_name}',
        method="PUT",
        path_params={'collection_name': collection_name},
        body={
            "vectors": {
                "image": {
                    "size": 2,
                    "distance": "Euclid"
                }
            },
            "sparse_vectors": {
                "image": {}
            }
        }
    )
    assert response.ok
```

## Expected Behavior

The collection creation should fail.

## Possible Solution

Enrich collection creation validation to make sure that dense and sparse vectors have different names.

tokhirasulov:
can you Assign me?

timvisee:
> can you Assign me?

We don't assign issues to specific people on GitHub. But feel free to give this a shot and pick it up.

agourlay:
Fix in https://github.com/qdrant/qdrant/releases/tag/v1.13.0

PR #5808, opened 2025-01-15
API validation for unique dense and sparse vector names

agourlay:
Tackling https://github.com/qdrant/qdrant/issues/3184

The new validation kicks in for both REST and gRPC.

### DIFF
--- a/src/migrations/single_to_cluster.rs
+++ b/src/migrations/single_to_cluster.rs
@@ -65,7 +65,8 @@ pub async fn handle_existing_collections(
                 strict_mode_config: collection_state.config.strict_mode_config,
                 uuid: collection_state.config.uuid,
             },
-        );
+        )
+        .expect("Failed to create collection operation");
 
         let mut consensus_operations = Vec::new();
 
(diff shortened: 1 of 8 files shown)


src/migrations/single_to_cluster.rs
```diff
@@ -65,7 +65,8 @@ pub async fn handle_existing_collections(
                 strict_mode_config: collection_state.config.strict_mode_config,
                 uuid: collection_state.config.uuid,
             },
-        );
```
agourlay, line 68: this one is called from main

generall (approved):
Assuming tests are green

src/migrations/single_to_cluster.rs
agourlay: the alternative is to skip the collection

timvisee (approved):
I assume it isn't easily possible to integrate this check in our existing [`validator`](https://docs.rs/validator/latest/validator/) infrastructure, is it? I believe that'd require us to implement a custom validator which messes with the other validation rules.

Assuming the above isn't viable - I'm good with this :heavy_check_mark:

agourlay:
@timvisee it is indeed difficult to use our `validator` infrastructure because two independent fields must be inspected for this validation.

PR #5816, opened 2025-01-16
Io measurement rename functions

JojiiOfficial:
Renames `*_measured` functions back to their original + more signature changes

### DIFF
--- a/lib/segment/src/payload_storage/payload_storage_base.rs
+++ b/lib/segment/src/payload_storage/payload_storage_base.rs
@@ -15,7 +15,12 @@ pub trait PayloadStorage {
     fn overwrite(&mut self, point_id: PointOffsetType, payload: &Payload) -> OperationResult<()>;
 
     /// Set payload for point_id. If payload already exists, merge it with existing
-    fn set(&mut self, point_id: PointOffsetType, payload: &Payload) -> OperationResult<()>;
+    fn set(
+        &mut self,
+        point_id: PointOffsetType,
+        payload: &Payload,
+        hw_counter: &HardwareCounterCell,
+    ) -> OperationResult<()>;
 
     /// Set payload to a point_id by key. If payload already exists, merge it with existing
     fn set_by_key(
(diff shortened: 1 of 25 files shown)


lib/segment/src/payload_storage/payload_storage_base.rs
```diff
@@ -15,28 +15,35 @@ pub trait PayloadStorage {
...
-    fn delete(&mut self, point_id: PointOffsetType, key: &JsonPath) -> OperationResult<Vec<Value>>;
+    fn delete(
+        &mut self,
+        point_id: PointOffsetType,
+        key: &JsonPath,
+        hw_counter: &HardwareCounterCell,
+    ) -> OperationResult<Vec<Value>>;
 
     /// Clear all payload of the point
     fn clear(&mut self, point_id: PointOffsetType) -> OperationResult<Option<Payload>>;
```
generall, line 49: if delete has `hw_counter`, clear and wipe should as well

lib/segment/src/payload_storage/query_checker.rs
```diff
@@ -248,6 +249,7 @@ impl SimpleConditionChecker {
 #[cfg(feature = "testing")]
 impl ConditionChecker for SimpleConditionChecker {
     fn check(&self, point_id: PointOffsetType, query: &Filter) -> bool {
+        let hw_counter = HardwareCounterCell::disposable(); // TODO(io_measurement): Propagate this value to caller!
```
generall, line 252: maybe not to the caller, but to the creator of `SimpleConditionChecker`

lib/segment/src/payload_storage/payload_storage_base.rs
JojiiOfficial: Yes they will! I tried to keep the PR smaller this time. Unless you want me to implement it in this PR, I'd suggest to add it in the next one.

generall: Either implement, or keep track somewhere of the leftovers

JojiiOfficial: Already implemented. It'll be included in the next PR 👍🏻

generall (approved):

timvisee (approved):

ffuugoo:
I think this PR _somehow_ broke compilation of some tests? E.g., `test_set_by_key_consistency` fails to build, because `hw_counter` is not propagated into `set` call? Or am I doing something wrong?

JojiiOfficial:
I pushed something into this PR https://github.com/qdrant/qdrant/pull/5822
@ffuugoo can you check if this PR resolves your issue?

ffuugoo:
@JojiiOfficial I've opened #5860 one with the fix for the broken test.

PR #5819, opened 2025-01-17
[Bugfix] Fix set_by_key not persisted between reload

JojiiOfficial:
Fixes a bug where in memory payload storage doesn't persist after restart of Qdrant when upserted using `set_by_key`


agourlay (approved):
Very good catch :+1:

timvisee (approved):
Thanks! 

Too bad it didn't make it in the 1.13 release :smiling_face_with_tear: 

Thanks for adding the test too!

timvisee:
Based on a quick git bisect this issue appears to have existed since Qdrant 1.8 onward.

PR #5822, opened 2025-01-17
Implement more IO measurements for PayloadStorage

JojiiOfficial:
Depends on #5816

### DIFF
--- a/lib/blob_store/src/tracker.rs
+++ b/lib/blob_store/src/tracker.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 
+use common::counter::hardware_counter::HardwareCounterCell;
 use memmap2::MmapMut;
 use memory::madvise::{Advice, AdviceSetting};
 use memory::mmap_ops::{
@@ -287,11 +288,24 @@ impl Tracker {
     }
 
     /// Unset the value at the given point offset and return its previous value
-    pub fn unset(&mut self, point_offset: PointOffset) -> Option<ValuePointer> {
+    pub fn unset(
+        &mut self,
+        point_offset: PointOffset,
+        hw_counter: &HardwareCounterCell,
+    ) -> Option<ValuePointer> {
         let pointer_opt = self.get(point_offset);
+
+        hw_counter
+            .io_read_counter()
+            .incr_delta(size_of::<Option<ValuePointer>>());
+
         if let Some(pointer) = pointer_opt {
             self.pending_updates
                 .insert(point_offset, PointerUpdate::Unset(pointer));
+
+            hw_counter
+                .io_write_counter()
+                .incr_delta(size_of::<Option<ValuePointer>>());
         }
 
         pointer_opt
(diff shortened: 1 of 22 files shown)


lib/blob_store/src/tracker.rs
```diff
@@ -287,11 +288,24 @@ impl Tracker {
     }
 
     /// Unset the value at the given point offset and return its previous value
-    pub fn unset(&mut self, point_offset: PointOffset) -> Option<ValuePointer> {
+    pub fn unset(
+        &mut self,
+        point_offset: PointOffset,
+        hw_counter: &HardwareCounterCell,
```
generall, line 294: Why only `unset` has `hw_counter` and `set` does not?

lib/segment/src/index/query_optimization/payload_provider.rs
```diff
@@ -52,7 +52,7 @@ impl PayloadProvider {
             // Rewrite condition checking code to support error reporting.
             // Which may lead to slowdown and assumes a lot of changes.
             PayloadStorageEnum::OnDiskPayloadStorage(s) => s
-                .read_payload(point_id)
+                .read_payload(point_id, &hw_counter)
```
generall, line 55: `hw_counter` should be initialized in the proper place

lib/segment/src/payload_storage/simple_payload_storage.rs
```diff
@@ -41,15 +42,25 @@ impl SimplePayloadStorage {
...
 
-    pub(crate) fn update_storage(&self, point_id: PointOffsetType) -> OperationResult<()> {
+    pub(crate) fn update_storage(
+        &self,
+        point_id: PointOffsetType,
+        hw_counter: &HardwareCounterCell,
+    ) -> OperationResult<()> {
+        let point_id_serialized = serde_cbor::to_vec(&point_id).unwrap();
+        hw_counter
+            .cpu_counter()
```
generall, line 52: Why CPU is involved here?

lib/blob_store/src/tracker.rs
JojiiOfficial: The place where `set` is called is already measured, `unset` can only be measured here. I tried to measure things as early as possible in the call stack to avoid too many function changes.
However thinking about it again it might be beneficial to measure functions like `set` too, in order to not forget measurements later.

lib/segment/src/index/query_optimization/payload_provider.rs
JojiiOfficial: What do you mean? There is a TODO for propagating the measured value which will be done in a separate PR.
If you want me to implement it here, I can do that too. I'm just trying to keep PRs small

lib/segment/src/payload_storage/simple_payload_storage.rs
JojiiOfficial: Good catch!

lib/blob_store/src/tracker.rs
JojiiOfficial: Changed!

lib/segment/src/index/query_optimization/payload_provider.rs
generall: the PR is called

> Finish io measurement for payload storage

Either we need to implement it, or rename the PR

JojiiOfficial: Renamed it 👍🏻

generall (approved):

timvisee (approved):

PR #5837, opened 2025-01-21
FixedLengthPriorityQueue: explicit sorted/unsorted methods

xzfc:
`FixedLengthPriorityQueue` API is confusing. Some iterators/conversions are ordered, some are not. Furthermore, these two implementations of `IntoIterator` contradict each other.

```rust
// unordered
impl<'a, T: Ord> IntoIterator for &'a FixedLengthPriorityQueue<T> { … }
// ordered
impl<T: Ord> IntoIterator for FixedLengthPriorityQueue<T> { … }
```

Giving that they are called implicitly, it makes things confusing:
```rust
let mut q = FixedLengthPriorityQueue::new(5);
for i in (0..5).rev() {
    q.push(i);
}

let mut a1 = Vec::new();
for &i in &q { // unordered
    a1.push(i);
}

let mut a2 = Vec::new();
for i in q { // ordered
    a2.push(i);
}

assert_eq!(a1, a2); // panic: [2, 4, 3, 1, 0] != [4, 3, 2, 1, 0]
```

This PR makes API more explicit:
- Implementations of `IntoIterator` are removed in favor of explicit `iter_unsorted` and `into_iter_sorted` methods.
- The method `into_vec` is renamed into `into_sorted_vec`.

Also, unlike old consuming `IntoIterator`, the new `into_iter_sorted` doesn't allocate intermediate vector, making it more efficient with partial iterators (i.e. `.take()`). Its based on a backported code from unstable Rust feature ([src](https://github.com/rust-lang/rust/blob/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/alloc/src/collections/binary_heap/mod.rs#L1616-L1657)).

### DIFF
--- a/lib/common/common/src/fixed_length_priority_queue.rs
+++ b/lib/common/common/src/fixed_length_priority_queue.rs
@@ -1,10 +1,9 @@
 use std::cmp::Reverse;
-use std::collections::binary_heap::Iter as BinaryHeapIter;
 use std::collections::BinaryHeap;
-use std::iter::Rev;
 use std::num::NonZeroUsize;
 use std::vec::IntoIter as VecIntoIter;
 
+use bytemuck::{TransparentWrapper as _, TransparentWrapperAlloc as _};
 use serde::{Deserialize, Serialize};
 
 /// To avoid excessive memory allocation, FixedLengthPriorityQueue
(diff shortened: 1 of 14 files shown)


lib/common/common/src/fixed_length_priority_queue.rs
```diff
@@ -59,20 +58,25 @@ impl<T: Ord> FixedLengthPriorityQueue<T> {
...
             .into_sorted_vec()
             .into_iter()
             .map(|Reverse(x)| x)
             .collect()
     }
 
-    /// Returns an iterator over the elements in the queue.
-    pub fn iter(&self) -> Iter<'_, T> {
-        Iter {
-            it: self.heap.iter().rev(),
```
xzfc, line 74: Also, I don't see a reason why `.rev()` was used.

It would place the smallest element at the end, but that's it; it doesn't fix the order order of all but the smallest element.

I've removed it.

timvisee (approved):
Thanks! I think this is a fair clean-up.

I really hope I've covered all the changes correctly, because this can cause nasty bugs. :smiley:

PR #5840, opened 2025-01-21
Remove setting OpenOptions flags to false

coszio:
As I was auditing the options combinations, I noticed that all `OpenOptions` flags are [false when creating the struct](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.new), so there is no need to specify them explicitly.

### DIFF
--- a/lib/common/memory/src/mmap_ops.rs
+++ b/lib/common/memory/src/mmap_ops.rs
@@ -15,7 +15,6 @@ pub fn create_and_ensure_length(path: &Path, length: usize) -> io::Result<File> 
         let file = OpenOptions::new()
             .read(true)
             .write(true)
-            .create(false)
             // Don't truncate because we explicitly set the length later
             .truncate(false)
             .open(path)?;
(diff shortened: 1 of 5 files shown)


lib/common/memory/src/mmap_ops.rs
```diff
@@ -39,22 +38,15 @@ pub fn create_and_ensure_length(path: &Path, length: usize) -> io::Result<File>
...
 }
 
 pub fn open_read_mmap(path: &Path, advice: AdviceSetting, populate: bool) -> io::Result<Mmap> {
     let file = OpenOptions::new()
         .read(true)
-        .write(false)
         .append(true)
         .create(true)
-        .truncate(false)
         .open(path)?;
```
coszio, line 50: Any specific reason why we need `append` and `create` here?

timvisee: I _think_ it was copy pasted from the read/write implementation and we forgot to move these options.

The [documentation](https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html#method.map) makes me think that only having the read flag is good enough.

I cannot think of another reason for us needing these two mutability flags.

timvisee (approved):

lib/common/memory/src/mmap_ops.rs
timvisee: Based on the above I think we should remove the `append` and `create` options. Then we can merge!

PR #5842, opened 2025-01-21
Avoid flushing a zero-sized mmapped file

coszio:
Discovered when running crasher 🔥 

The problem arose when trying to convert an empty `ImmutableInvertedIndex` into a `MmapInvertedIndex`, which would create zero-sized files.

There seems to be no problem mmapping them, but, when we try to `msync` during flushing, it panics with an [`EINVAL` error](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/msync.2.html#//apple_ref/doc/man/2/msync).

I have done a simple fix where we just don't flush when the file length is zero.

### DIFF
--- a/lib/common/memory/src/mmap_type.rs
+++ b/lib/common/memory/src/mmap_type.rs
@@ -165,7 +165,10 @@ where
         Box::new({
             let mmap = self.mmap.clone();
             move || {
-                mmap.flush()?;
+                // flushing a zero-sized mmap can cause panicking on some systems
+                if mmap.len() > 0 {
+                    mmap.flush()?;
+                }
                 Ok(())
             }
         })
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
@@ -285,6 +285,7 @@ mod tests {
     #[case(1111, 1110)]
     #[case(1111, 0)]
     #[case(10, 2)]
+    #[case(0, 0)]
     #[test]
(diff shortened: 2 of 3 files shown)


lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -285,6 +285,7 @@ mod tests {
     #[case(1111, 1110)]
     #[case(1111, 0)]
     #[case(10, 2)]
+    #[case(0, 0)]
```
coszio, line 288: If we had included this case, we would've discovered this much earlier 📝

ffuugoo (approved):

lib/common/memory/src/mmap_type.rs
```diff
@@ -165,7 +165,10 @@ where
...
             let mmap = self.mmap.clone();
             move || {
-                mmap.flush()?;
+                // flushing a zero-sized mmap can cause panicking on some systems
+                if mmap.len() > 0 {
+                    mmap.flush()?;
+                }
                 Ok(())
             }
         })
```
timvisee, line 174: Can we move the length check outside of the box, and return an empty closure if it's empty?

That may be a bit nicer so we don't clone the mmap structure (even though it's empty).

timvisee (approved):

lib/common/memory/src/mmap_type.rs
coszio: Is it ensured that a mmap won't grow its size? just thinking that this returns a closure which will be run at a later time

Issue #5834, opened 2025-01-20
Panic on search after GPU indexing


generall:
## Current Behavior

Search fails after index was build with GPU

## Steps to Reproduce

Build index on the machine with GPU

```
docker run --rm -it -e --network=host qdrant/bfb:dev ./bfb -n 14000000 -d 768 -p 10 -t 2 --quantization-in-ram true --quantization binary --on-disk-payload --on-disk-vectors true --max-segment-size 2000000
```

After indexing, start searches:

```
docker run --rm -it --network=host qdrant/bfb:dev ./bfb -d 768 -n 10000 --quantization binary --skip-create --skip-upload --skip-wait-index --quantization-rescore false --search
```

Search randomly crashes with

```
2025-01-20T20:41:18.667923Z ERROR qdrant::startup: Panic occurred in file lib/segment/src/index/hnsw_index/graph_links.rs at line 233: called `Option::unwrap()` on a `None` value

2025-01-20T20:41:18.805262Z DEBUG collection::shards::replica_set::execute_read_operation: Read operation failed: Service internal error: task 555390 panicked with message "called `Option::unwrap()` on a `None` value"
2025-01-20T20:41:18.806298Z ERROR qdrant::tonic::logging: gRPC /qdrant.Points/Search unexpectedly failed with Internal error "Service internal error: 1 of 1 read operations failed:\n  Service internal error: task 555390 panicked with message \"called `Option::unwrap()` on a `None` value\"" 0.142194
2025-01-20T20:41:18.809868Z ERROR qdrant::startup: Panic backtrace:
   0: qdrant::startup::setup_panic_hook::{{closure}}
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys::backtrace::__rust_end_short_backtrace
   4: rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::panicking::panic
   7: core::option::unwrap_failed
   8: segment::index::hnsw_index::hnsw::HNSWIndex::search_with_graph
   9: <core::iter::adapters::GenericShunt<I,R> as core::iter::traits::iterator::Iterator>::next
  10: segment::index::hnsw_index::hnsw::HNSWIndex::search_vectors_with_graph
  11: <segment::index::vector_index_base::VectorIndexEnum as segment::index::vector_index_base::VectorIndex>::search
[...]

PR #5848, opened 2025-01-22
Filter broken gpu links

IvanPleshkov:
Fix https://github.com/qdrant/qdrant/issues/5834

It's a quick fix which includes additional sync in shaders and results filtering.
It's not a proper solution but it fixes panic.


generall:
Precision is Ok on my test, I get around 10 broken links per 200k segment. Tested on both - random and read data

generall (approved):

PR #5849, opened 2025-01-22
Enable workspace lints for `gpu`

xzfc:
This PR enables our default set of lints for the `gpu` crate and fixes them.
Fixes are performed automatically using the following command:
```console
cargo clippy -p gpu --all-features --all-targets
```

---
For the reference, a command to find crates without lints enabled:
```console
find -name Cargo.toml | xargs grep -L '^workspace = true$'
```
(currently only `gpu` is missing)

### DIFF
--- a/lib/gpu/Cargo.toml
+++ b/lib/gpu/Cargo.toml
@@ -7,6 +7,9 @@ authors = [
 license = "Apache-2.0"
 edition = "2021"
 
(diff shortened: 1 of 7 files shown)


IvanPleshkov (approved):

IvanPleshkov:
Thanks for this change!

PR #5851, opened 2025-01-22
shell.nix: add vulkan-loader to LD_LIBRARY_PATH

xzfc:
This PR updates `shell.nix` to let Qdrand find `libvulkan.so.1`.

Tested only on NixOS. Not sure how it would affect nix without NixOS on Linux/macOS/Windows.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -109,6 +109,10 @@ mkShell {
       export MACOSX_DEPLOYMENT_TARGET="10.13"
     fi
 
+    # GPU bindings require libvulkan.so.1 during runtime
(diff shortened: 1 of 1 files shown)


IvanPleshkov (commented):
For development puproses also required tools. It includes device extensions like validator etc
You may adopt these bash commands from our CI:
```
sudo apt-get install clang
sudo apt-get install mesa-vulkan-drivers libvulkan1 vulkan-tools vulkan-validationlayers
```

Check status:
```
/usr/bin/vulkaninfo --summary
```

PR #5844, opened 2025-01-21
Propagate hardware counter for more functions

JojiiOfficial:
Depends on #5822 and #5816

PR #5863, opened 2025-01-23
[WIP] Io measurement for auth middlewares

JojiiOfficial:
WIP

Depends on #5844


generall:
I am not sure we need this. Serverless user normally doesn't have control over JWT token structure

PR #5866, opened 2025-01-23
id_tracker filters out points without version to honor flushing sequence

agourlay:
A point without a version did not go through the full flushing sequence because the version is flushed last.

An incomplete point should not be visible when merging points from the different segments.

The current behavior creates false positive for the data consistency check, where points without vectors get a synthetic version `Some(0)` generated.

Example of broken storage created by [crasher](https://github.com/qdrant/crasher)

```
025-01-23T14:40:08.473815Z ERROR segment::segment::segment_ops: Vector storage 'sparse-vector-float16-1' is missing point Some(NumId(750)) point_offset: 1600 version: Some(0)    
2025-01-23T14:40:08.473900Z ERROR segment::segment::segment_ops: Vector storage 'sparse-vector-float16-2' is missing point Some(NumId(750)) point_offset: 1600 version: Some(0)    
2025-01-23T14:40:08.473911Z ERROR segment::segment::segment_ops: Vector storage 'sparse-vector-index-memory-2' is missing point Some(NumId(750)) point_offset: 1600 version: Some(0)    
2025-01-23T14:40:08.473936Z ERROR segment::segment::segment_ops: Vector storage 'sparse-vector-float16-1' is missing point Some(NumId(740)) point_offset: 1601 version: Some(0)    
2025-01-23T14:40:08.473947Z ERROR segment::segment::segment_ops: Vector storage 'sparse-vector-float16-2' is missing point Some(NumId(740)) point_offset: 1601 version: Some(0)
```

### DIFF
--- a/lib/segment/src/id_tracker/mod.rs
+++ b/lib/segment/src/id_tracker/mod.rs
@@ -27,6 +27,8 @@ pub struct MergedPointId {
 }
 
 /// Calls a closure for each unique point from multiple ID trackers.
+///
+/// Discard points that have no version.
 pub fn for_each_unique_point<'a>(
     id_trackers: impl Iterator<Item = &'a (impl IdTracker + ?Sized + 'a)>,
     mut f: impl FnMut(MergedPointId),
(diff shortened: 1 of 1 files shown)


coszio (approved):

timvisee (approved):
I do think this change is correct.

I'm still wondering though if there's any possible side effect introduced by this. I suggest to keep an eye on chaos testing to see if behavior changes there.

PR #5871, opened 2025-01-24
update filters out points without version to honor flushing sequence

agourlay:
Same motivation as https://github.com/qdrant/qdrant/pull/5866

Points without version are not well defined and should not be updated, only deleted.


timvisee (approved):
This is quite a fundamental change.

I totally agree that we should not rely on/use points that have no version, as I'd consider that bad practice.

I cannot think of a clear scenario where applying this change becomes problematic. In case of our copy on write operations, we always flush appendable segments first which should prevent failing to persist the moved point version before deleting the old point.

Let's merge, but keep this change in mind if we see any inconsistency problems in the future.

PR #5873, opened 2025-01-24
Simplify tar ext

ffuugoo:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


xzfc:
Also, please test it on interrupted snapshot streaming; check if it doesn't panic. It's  the situation that these unit tests are trying to recreate.

E.g. run on a non-empty collecton:
```sh
curl -v 127.0.0.1:6333/collections/collection_name/shards/0/snapshot --max-time 2 --limit-rate 10k --output /dev/null
```

lib/common/common/src/tar_ext.rs
```diff
@@ -390,39 +301,39 @@ mod tests {
...
-        let tar = BuilderExt::new_seekable_owned(DummyBridgeWriter(false, Arc::clone(&data)));
+        let tar = BuilderExt::new_seekable_owned(DummyBridgeWriter(false, data.clone()));
         assert!(tar.finish().await.is_ok());
         assert_eq!(data.lock().await.len(), 1024);
     }
 
     #[tokio::test]
     async fn test_dummy_finish_fail() {
         let data = Arc::new(Mutex::new(Vec::new()));
-        let tar = BuilderExt::new_seekable_owned(DummyBridgeWriter(true, Arc::clone(&data)));
```
generall, line 409: I like the explicit one better

ffuugoo: Why though? 🤔

xzfc: It's recommended in the [Rust Book](https://doc.rust-lang.org/book/ch15-04-rc.html#using-rct-to-share-data):

> We could have called `a.clone()` rather than `Rc::clone(&a)`, but Rust’s convention is to use `Rc::clone` in this case.

And once it was [recommended in the `std` doc](https://doc.rust-lang.org/1.38.0/std/sync/struct.Arc.html#cloning-references):

> The `Arc::clone(&from)` syntax is the most idiomatic because it conveys more explicitly the meaning of the code. In the example above, this syntax makes it easier to see that this code is creating a new reference rather than copying the whole content of foo.

On the other hand, it's debatable whether it's more idiomatic or just opinionated; and they've [removed](https://www.github.com/rust-lang/rust/pull/63252) this recommendation from the `std` docs, and [demoted](https://www.github.com/rust-lang/rust-clippy/pull/2355) the corresponding clippy lint (`clone_on_ref_ptr`) from being enabled by default.

---

My preference is to keep it explicit.

lib/common/common/src/tar_ext.rs
```diff
@@ -325,38 +177,97 @@ impl<W: Send + Write + Seek + 'static> BuilderExt<W> {
...
 
+/// A wrapper that drops inner type `T` inside [`tokio::task::block_in_place`].
+///
+/// This is a is a workaround for the inconvenient [`tar::Builder`] behavior: dropping a [`tar::Builder`]
+/// might cause a final write of archive footer. This behavior is problematic for [`Write`]
+/// implementations that could panic when used in an async context, such as [`SyncIoBridge`].
+///
+/// Note, that [`tokio::task::block_in_place`] **panics** when called under "current-thread" Tokio runtime,
+/// and so this wrapper only works under "multi-threaded" runtime.
+struct BlockInPlaceOnDrop<T>(Option<T>);
```
timvisee, line 216: Would it be worth to put this in a separate module, because it's a generic type that may be used elsewhere?

timvisee (commented):
image :ok_hand:

lib/common/common/src/tar_ext.rs
ffuugoo: Yeah, I remember it being recommended, but I never liked and followed that convention. I'll revert, but I see very little merit in this rule. It's just noise. 🙄

ffuugoo:
Also, stop reviewing draft PRs, you're stressing me out 😬

lib/common/common/src/tar_ext.rs
```diff
@@ -325,38 +177,109 @@ impl<W: Send + Write + Seek + 'static> BuilderExt<W> {
...
+
+        if rt.runtime_flavor() != tokio::runtime::RuntimeFlavor::CurrentThread {
+            tokio::task::block_in_place(move || drop(inner));
+        } else {
+            // We can't use `block_in_place` with current-thread runtime, so just drop `inner` as-is.
+            // It might block the runtime and `SyncIoBridge` will still panic, but it's the best we can do. 🤷‍♀️
+            drop(inner);
+        }
+    }
+}
```
ffuugoo, line 246: I'm not super happy about the whole `BlockInPlaceOnDrop`, because it still panics if run on a current-thread Tokio runtime. And I'm ready to drop this change, if everyone prefer `BlowFuseOnDrop`, but hear me out:

Cons:
- panics if run on current-thread Tokio runtime
  - _I think_, we mostly use multi-thread runtimes in Qdrant, so it should not be a major issue, I think 🤔
  - at worst we'll have to use `#[tokio::test(flavor = "multi_thread")]` on some of our tests

Pros:
- it's simpler than `BlowFuseOnDrop`
- it's more **correct**, because it's easy to introduce subtle bugs with `BlowFuseOnDrop`

ffuugoo: > it's easy to introduce subtle bugs with BlowFuseOnDrop

If we look how we *currently* do `Regular` snapshots:

https://github.com/qdrant/qdrant/blob/a72a5f43e8574ad7810dc45967c393cf6acf4163/lib/segment/src/segment/entry.rs#L820-L824

In this snippet, we create and write an *inner* tar archive (for the segment) directly into the outer tar archive of the collection/shard snapshot.

The issue is that **we forgot to call `BuilderExt::finish`**. And because of `BlowFuseOnDrop`, `tar::Builder::finish` *won't* be called when we drop `BuilderExt`. Basically, when creating snapshots in `Regular` format we don't correctly finalize inner segment archives. 💁‍♀️

```rust
tar.blocking_write_fn(Path::new(&format!("{segment_id}.tar")), |writer| {
    // We create "nested" `tar_ext::BuilderExt`, that writes inner archive directly into outer archive,
    // without creating any additional files on disk
    let tar = tar_ext::BuilderExt::new_streaming_borrowed(writer);
    let tar = tar.descend(Path::new(SNAPSHOT_PATH))?;
    snapshot_files(self, temp_path, &tar) // we write data into inner archive... but we don't call `tar.finish()` after
    // so, `tar` variable is dropped here, `BlowFuseOnDrop` "disables" `tar::Builder::finish`, and the tail of the archive is not properly written
})??;
```

It's not critical in this case (@xzfc said that `tar::Builder::finish` only pads the tail of the archive with zero-bytes, so it's not super important?), but it's a live example, that it's *easy* to introduce a *potential* bug.

`BlockInPlaceOnDrop` tries to fix this in the most generic way possible, with a single caveat that it panics when using current-thread runtime.

ffuugoo: The bug example is relevant, but `block_in_place` is not good enough in practice. :/

ffuugoo:
I've removed more extreme changes from the PR, it's ready for review.

lib/common/common/src/tar_ext.rs
ffuugoo: Decided to remove this change from the PR.

xzfc:
At first glance, LGTM, will check thoughtfully a bit later.

xzfc (approved):
Looks neat, thanks. Sorry for long wait, got piled by a backlog.

PR #5876, opened 2025-01-27
Bump itertools 0.14

agourlay:
Itertools 0.14.0 has been out for a month but for some reason Dependabot is not picking it up.

https://github.com/rust-itertools/itertools/blob/master/CHANGELOG.md#0140

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -493,7 +493,7 @@ version = "1.13.2-dev"
 dependencies = [
  "chrono",
  "common",
- "itertools 0.13.0",
+ "itertools 0.14.0",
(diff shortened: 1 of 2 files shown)


timvisee (approved):
> but for some reason Dependabot is not picking it up

Our ['10 updates per week'](https://github.com/qdrant/qdrant/blob/a72a5f43e8574ad7810dc45967c393cf6acf4163/.github/dependabot.yml#L8) may be too restrictive.

agourlay:
It seems Dependabot only created 6 PRs last week, it had the capacity :shrug:

PR #5881, opened 2025-01-27
Add payload_json! macro

xzfc:
We have this impl:
https://github.com/qdrant/qdrant/blob/a72a5f43e8574ad7810dc45967c393cf6acf4163/lib/segment/src/types.rs#L1243-L1250

A `From` implementation that panics violates [this trait basic assumption](https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from). To avoid that, we could replace it with `TryFrom` and use `.try_into().unwrap()` instead of `.into()`. But in our code, this conversion is used solely to convert from `json!({…})` literals. So, a cleaner approach: introduce a macro `payload_json!` that couldn't fail.

This PR introduces a new macro `payload_json!` and replaces replaces usages like `json!({ "some": "thing" }).into()` with `payload_json! { "some": "thing" }`. The `From` impl is removed.

### DIFF
--- a/lib/collection/src/collection_manager/collection_updater.rs
+++ b/lib/collection/src/collection_manager/collection_updater.rs
@@ -78,6 +78,7 @@ mod tests {
     };
     use segment::entry::entry_point::SegmentEntry;
     use segment::json_path::JsonPath;
+    use segment::payload_json;
     use segment::types::PayloadSchemaType::Keyword;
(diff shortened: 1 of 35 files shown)


timvisee (approved):
In my opinion this is a great use case for a new macro. :ok_hand: 

I like the infallible construction!

PR #5879, opened 2025-01-27
Add more tests for find_fitting_gap in blob store

timvisee:
Two more tests that cover finding a gap in our blob store implementation.

These tests try to find a number of blocks covering one or more regions. It especially tests the way `find_fitting_gap` is implemented with the `windows` iterator.

Two sets of assertions fail on our current branch, marked with a TODO.

### All Submissions:

* [ ] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [ ] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [ ] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #5882, opened 2025-01-27
Fix gap finding logic

coszio:
There was a case, where we were missing to scan a section of the gaps to find a spot to write. 

Even when having enough space in the last region + next to last region, the logic was exiting earlier and causing a panic because there were "allegedly" no more pages even after giving the opportunity to create them.

This PR rewrites the logic and fixes it by ensuring all cases are covered before responding that there is no gap large enough. Now it does one pass with the minimum amount of regions, and another one with the maximum if needed.

### DIFF
--- a/lib/blob_store/src/bitmask/gaps.rs
+++ b/lib/blob_store/src/bitmask/gaps.rs
@@ -261,7 +261,7 @@ mod tests {
     use tempfile::tempdir;
 
     use super::*;
-    use crate::config::{StorageOptions, DEFAULT_BLOCK_SIZE_BYTES, DEFAULT_REGION_SIZE_BLOCKS};
+    use crate::config::{StorageOptions, DEFAULT_REGION_SIZE_BLOCKS};
 
     prop_compose! {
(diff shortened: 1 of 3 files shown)


timvisee (changes requested):
I did write some tests for `find_fitting_gaps` on the side in <https://github.com/qdrant/qdrant/pull/5879>.

These changes don't make the test pass.

More specifically: if we have 3 empty regions, and we try to find blocks covering about 2.1 regions, we still fail.

coszio:
Thank you for the extra tests @timvisee!

I missed asserting that the middle regions were also empty when traversing the windows. Just pushed the fix. 

Would you mind giving it another look?

agourlay:
@coszio did you try running our blobstore benchmark for sanity check?

coszio:
@agourlay 
Just ran them now. No noticeable degradation was found

timvisee (approved):
As discussed, this now checks both window sizes. I'm happy with this as a patch.

Thanks for including both sets of tests.

PR #5884, opened 2025-01-27
Use dynamic windows with sliding head/tail in find_fitting_gap

timvisee:
Improvement on <https://github.com/qdrant/qdrant/pull/5882>
Alternative for <https://github.com/qdrant/qdrant/pull/5883>

Tweaks the `find_fitting_gap` and implements the dynamic sliding window with head/tail pointer as discussed in a call.

This _seeks_ the head and tail pointer forward until a range of regions is found with enough blocks to satisfy the requested size.

On each iteration, this scans backwards from the head (higher index) to the tail (lower index). This way we can jump the tail forward multiple steps if we find a region that is not fully empty.

All new tests added in <https://github.com/qdrant/qdrant/pull/5882> are included and succeed.

I did not benchmark this (yet).

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/blob_store/src/bitmask/gaps.rs
+++ b/lib/blob_store/src/bitmask/gaps.rs
@@ -194,18 +194,18 @@ impl BitmaskGaps {
             return None;
         }
 
-        let (mut head, mut tail) = (0, 0);
+        let (mut head, mut tail) = (min_regions_needed - 1, 0);
 
         // Slide head/tail over regions until we find a gap
         'slider: while head < self.len() {
             // If we need more/less regions, advance head/tail
             let region_count = head.saturating_sub(tail) + 1;
             if region_count < min_regions_needed {
-                head += 1;
+                head += min_regions_needed - region_count;
                 continue;
             }
             if region_count > max_regions_needed {
-                tail += 1;
+                tail += region_count - max_regions_needed;
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


lib/blob_store/src/bitmask/gaps.rs
```diff
@@ -186,72 +186,70 @@ impl BitmaskGaps {
...
-                // make sure the middle regions are all free
-                let middle_regions = &gaps[1..window_size - 1];
-                if middle_regions
-                    .iter()
-                    .any(|gap| gap.max as usize != self.config.region_size_blocks)
-                {
-                    return None;
+                if blocks >= num_blocks {
+                    return Some(region_index as u32..head as u32 + 1);
                 }
```
coszio, line 231: This check only applies until `(head - region_index) >= min_regions_needed`, but it is fine since we cannot avoid to check the region's `trailing` field, and the check is super cheap.

coszio (commented):
After spending quite some time of analyzing this logic, I agree it is sound and efficient, specially for the case of a value which spans 2 or more regions. Although I don't think such large values should be normal.

I would love to see some benchmarking against current implementation, just to get some numbers on the speedup this can generate, even with a synthetic dataset. It would help to build a case for (IMO) increasing the cognitive load of the method.

timvisee:
> I would love to see some benchmarking against current implementation, just to get some numbers on the speedup this can generate, even with a synthetic dataset.

I did a basic benchmark using our `real_data_bench` benchmark. It shows a +40% regression with this implementation.

```
Benchmarking insert real payload (large): Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 75.8s, or reduce sample count to 10.
insert real payload (large)
                        time:   [1.5283 s 1.7568 s 2.0280 s]
                        change: [+12.821% +40.400% +76.542%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) high mild
  7 (7.00%) high severe
```

> It would help to build a case for (IMO) increasing the cognitive load of the method.

I like the approach in this PR better because there's a single loop with a few conditions that does everything. Versus branching into different blocks or functions based on what we're looking for like currently in dev.

What's currently in dev clearly performs better. I don't want to go through the effort of tuning this, so I'll close the PR. It was worth the shot!

PR #5892, opened 2025-01-27
bump and migrate to `rand` 0.9.0

coszio:
Supersedes #5885 

Also bumps `rand_distr` to 0.5.0 to match it.

Performs a rather lengthy migration of renames and one slightly breaking change: now the implementation for generating random `usize` values is gone, so we must choose between `u32` or `u64`. I think this is for good reproducibility reasons.

Here is the [migration guide](https://rust-random.github.io/book/update-0.9.html)


timvisee (approved):
There's an insane amount of rng in our codebase :exploding_head: 

Thanks for the busy work!

I've added the AVX2, SSE and GPU conversions as well.

timvisee:
There's a unit test regression:

```
thread 'sparse_vector_index_search_tests::sparse_vector_index_fallback_plain_search' panicked at lib/segment/tests/integration/sparse_vector_index_search_tests.rs:115:13:
assertion `left == right` failed: query = SparseVector {
    indices: [
        4091,
    ],
    values: [
        -44.223434,
    ],
}, filter_result = [
    ScoredPointOffset {
        idx: 0,
        score: 0.0,
    },
] no_filter_result = []
  left: 1
 right: 0
```

I see no obvious reason yet on why this is failing. Changes to the random function around this unit test seem sound.

Could it be an rng seeding change which now magically hits an errorrous code path?

agourlay:
I will have a look at the failing sparse vector test :eyes:

agourlay (approved):
:clap:

PR #5897, opened 2025-01-28
Turn for_each_packed_link into iterator

xzfc:
Initially, a simple interface of `for_each_packed_link()` was enough for a search use case. However, it's not convenient for other cases (e.g. incremental HNSW); a proper `Iterator` is more convenient.

This PR replaces `fn for_each_packed_link()` with `fn iterate_packed_links()` which returns an `Iterator`.

<table><th>Old<th>New<tr><td>

```rust
pub fn for_each_packed_link(
    links: &[u8],
    bits_per_unsorted: u8,
    sorted_count: usize,
    mut f: impl FnMut(u32),
) {                               
```

<td>

```rust
pub fn iterate_packed_links(
    links: &[u8],
    bits_per_unsorted: u8,
    sorted_count: usize,
) -> PackedLinksIterator {        
```

</table>

A special care is taken to make the performance not worse by overriding the provided method `Iterator::fold()` instead of relying just on `Iterator::next()`.

Additionally, tests for `common::bitpacking_links` module are extended to cover more cases.

This PR could be reviewed on a per-commit basis.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1255,6 +1255,7 @@ dependencies = [
  "ordered-float 4.6.0",
  "ph",
  "rand 0.9.0",
+ "rstest",
  "self_cell",
  "semver",
  "serde",
--- a/lib/common/common/Cargo.toml
+++ b/lib/common/common/Cargo.toml
@@ -39,6 +39,7 @@ log = { workspace = true }
 common = { path = ".", features = ["testing"] }
 criterion = { workspace = true }
 itertools = { workspace = true }
+rstest = { workspace = true }
 self_cell = "1.1.0"
 tango-bench = "0.6.0"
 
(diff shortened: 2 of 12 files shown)


generall:
I see `fold` is implemented, but not yet used in this PR, is that correct? If so, should we expect degradation on benchmarks once this PR is merged and before followup with actual usage?

xzfc:
> I see `fold` is implemented, but not yet used in this PR, is that correct?

It is used but indirectly: the default implementation of `Iterator::for_each` is expressed through `fold()`:
https://github.com/rust-lang/rust/blob/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/iter/traits/iterator.rs#L790-L801

I don't expect performance degradation after merging this PR.

generall (approved):

timvisee (approved):

PR #5898, opened 2025-01-29
HwCounter rename IO metrics

JojiiOfficial:


generall (approved):
Assuming it is safe to rename since we don't enable it by default, therefore clients should not be affected

PR #5908, opened 2025-01-29
Fix flushing of RocksDB sparse vector storage

agourlay:
The `SimpleSparseVectorStorage` based on RocksDB is not honoring the flushing sequence as it persist eagerly changes.

This causes discrepancies in testing with the new `MmapSparseStorage` which behaves properly.

With this patch, both implementation trigger the same consistency issue in Crasher which is unrelated to the sparse storage.

### DIFF
--- a/lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
+++ b/lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
@@ -3,7 +3,7 @@ use std::mem;
 
 use parking_lot::Mutex;
 
-use crate::common::operation_error::OperationResult;
+use crate::common::operation_error::{OperationError, OperationResult};
 use crate::common::rocksdb_wrapper::{DatabaseColumnWrapper, LockedDatabaseColumnWrapper};
 use crate::common::Flusher;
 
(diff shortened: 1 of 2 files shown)


lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
```diff
@@ -77,4 +77,40 @@ impl DatabaseColumnScheduledUpdateWrapper {
...
+    pub fn get<K>(&self, key: K) -> OperationResult<Vec<u8>>
+    where
+        K: AsRef<[u8]>,
+    {
+        if let Some(value) = self.insert_pending_persistence.lock().get(key.as_ref()) {
+            return Ok(value.clone());
+        }
+        if self
+            .deleted_pending_persistence
+            .lock()
```
timvisee, line 90: It looks like we basically always lock these two mutexes together.

How about using a single mutex for both pending insertions and deletions?

timvisee (approved):
Good catch!

I've a comment on the mutex we use, not strictly related to this change, but it may be a good time to improve upon now.

coszio (approved):

lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
agourlay: I think a single Mutex is indeed better perf. wise :+1: 

I will do it in a follow up request to get a cleaner diff.

timvisee: https://github.com/qdrant/qdrant/pull/5915

PR #5909, opened 2025-01-29
Relax consistency check for WAL replay

agourlay:
The feature `data-consistency-check` enables an optional data consistency check after the WAL run.

It validates that segments are consistent on their own which can be overly restrictive because a shard can still be consistent after merging points from underlying segments.

This patch ignores the check for a particular case involving unflushed points behind resurrected by the WAL replay.

### DIFF
--- a/lib/segment/src/id_tracker/simple_id_tracker.rs
+++ b/lib/segment/src/id_tracker/simple_id_tracker.rs
@@ -164,7 +164,8 @@ impl SimpleIdTracker {
                 internal_to_version[internal_id as usize] = version;
             } else {
                 log::debug!(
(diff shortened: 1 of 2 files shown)


coszio (approved):
I just got a very good explanation of why `Some(0)` is a side-effect of our structure for internal id to versions.

`internal_to_version` is just an array of integers. Say it has a length of 10, and we want to add a version for offset 20. When we resize it, offsets 10..20 will need to be initialized to 0, and only offset 20 will have a "real" version.

Until we create another implementation which can distinguish between "not actually initialized" and "newly initialized", the relaxation in this PR will help to not report false positives.

timvisee (approved):

PR #5911, opened 2025-01-30
Temporarily run code coverage on merging to dev

KShivendu:
Coverage action didn't get scheduled or show up in https://github.com/qdrant/qdrant/actions because it's not part of master yet. I think triggering it once from dev branch should bring it there. Don't want to wait for next release

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -4,6 +4,8 @@ on:
   workflow_dispatch: # Trigger from UI for selected branch
   schedule:
     - cron: "0 0 * * *"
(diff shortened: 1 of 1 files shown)


timvisee:
> I think triggering it once from dev branch should bring it there. Don't want to wait for next release

Scheduled workflows only run on the default branch, which means it must be merged into `master` first.

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

timvisee (approved):

PR #5751, opened 2025-01-07
Setup code coverage reports

KShivendu:
### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

PR #5913, opened 2025-01-30
Use cargo --locked on CI for reproducible builds

agourlay:
Without --locked, Cargo will:

1. Use the Cargo.lock file if it exists
2. Update the lock file if necessary to resolve dependencies
3. Create a new lock file if one doesn't exist

With --locked, Cargo will:

1. Use the exact versions specified in the existing Cargo.lock file
2. Fail if the lock file is missing
3. Fail if any changes to the lock file are required

The main difference is that --locked prevents any modifications to the lock file, ensuring deterministic builds, which is particularly useful in CI environments.

Without --locked, Cargo may update the lock file if needed, but it will still use the lock file as a starting point for dependency resolution.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -35,7 +35,7 @@ jobs:
           poetry -C tests check --lock
           poetry -C tests install --no-root
       - name: Build
-        run: cargo build --features "service_debug data-consistency-check"
+        run: cargo build --features "service_debug data-consistency-check" --locked
       - name: Run integration tests
         run: poetry -C tests run ./tests/integration-tests.sh
         shell: bash
@@ -65,7 +65,7 @@ jobs:
           poetry -C tests check --lock
           poetry -C tests install --no-root
       - name: Build
-        run: cargo build --features "service_debug data-consistency-check"
+        run: cargo build --features "service_debug data-consistency-check" --locked
       - name: Run integration tests - 1 peer
         run: poetry -C tests run ./tests/integration-tests.sh distributed
         shell: bash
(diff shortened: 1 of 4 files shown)


KShivendu:
@agourlay is the `cargo build` step required for everything? Asking because [coverage tests](https://github.com/qdrant/qdrant/pull/5751) don't do it.

agourlay:
`cargo nextest` will build the repo before the running the tests.

I do not think we need `--locked` everywhere, just a few places to catch issues should be good enough.

timvisee:
I'm a bit afraid that this will break CI runs on older bits of code. For example, on the `master` branch, or from some previous version. We do sometimes run jobs again.

Any thoughts on that?

Other than that I'm happy to give this a shot, as we can revert if any problems arise.

xzfc:
@agourlay
> I do not think we need `--locked` everywhere, just a few places to catch issues should be good enough.

I think adding it in an exactly one<sup>†</sup> place would be enough for tests. I.e. a step `cargo check --frozen --dry-run` at [rust-lint.yml](https://github.com/qdrant/qdrant/blob/80bfc03aa0daef98709cd0c95fdf90f62c4f83d5/.github/workflows/rust-lint.yml). Reason: "a check that `Cargo.lock` is consistent" conceptually is a lint. We'll still catch this error on CI, but it will be a single red cross mark rather than every test failing.

† But keep it for `release-artifacts.yml` as a safeguard, as it's not just merely a test, but a job that uploads binaries.

---

@timvisee
> I'm a bit afraid that this will break CI runs on older bits of code. For example, on the master branch, or from some previous version.

It shouldn't break on old commits, unless these commits have an inconsistency between their `Cargo.lock` and `Cargo.toml`. If we want to be really sure, we can run `cargo check --frozen --dry-run` on every commit in the master. (only need to do it only once)

timvisee:
> It shouldn't break on old commits, unless these commits have an inconsistency between their Cargo.lock and Cargo.toml. If we want to be really sure, we can run cargo check --frozen --dry-run on every commit in the master. (only need to do it only once)

I was thinking about yanked packages. But I'm actually not sure if it would bump the lockfile, and the likeliness of this happening would probably be very very low anyway.

PR #5915, opened 2025-01-30
Optimize RocksDB buffered wrapper with single Mutex

agourlay:
The two mutexes are always acquired together so they can be merged.

### DIFF
--- a/lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
+++ b/lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
@@ -16,16 +16,20 @@ use crate::common::Flusher;
 #[derive(Debug)]
 pub struct DatabaseColumnScheduledUpdateWrapper {
     db: DatabaseColumnWrapper,
-    deleted_pending_persistence: Mutex<HashSet<Vec<u8>>>,
-    insert_pending_persistence: Mutex<HashMap<Vec<u8>, Vec<u8>>>,
+    pending_operations: Mutex<PendingOperations>, // in-flight operations persisted on flush
+}
+
+#[derive(Debug, Default)]
+struct PendingOperations {
+    deleted: HashSet<Vec<u8>>,
+    inserted: HashMap<Vec<u8>, Vec<u8>>,
 }
 
 impl DatabaseColumnScheduledUpdateWrapper {
     pub fn new(db: DatabaseColumnWrapper) -> Self {
         Self {
             db,
-            deleted_pending_persistence: Mutex::new(HashSet::new()),
-            insert_pending_persistence: Mutex::new(HashMap::new()),
+            pending_operations: Mutex::new(PendingOperations::default()),
         }
     }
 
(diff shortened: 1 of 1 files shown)


lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
```diff
@@ -47,17 +51,16 @@ impl DatabaseColumnScheduledUpdateWrapper {
...
+        pending_guard.deleted.insert(key.to_vec());
         Ok(())
     }
 
     pub fn flusher(&self) -> Flusher {
-        let ids_to_insert = mem::take(&mut *self.insert_pending_persistence.lock());
-        let ids_to_delete = mem::take(&mut *self.deleted_pending_persistence.lock());
+        let pending_operations = mem::take(&mut *self.pending_operations.lock());
+        let ids_to_insert = pending_operations.inserted;
+        let ids_to_delete = pending_operations.deleted;
```
timvisee, line 63: Nit: how about struct destructing here?

timvisee, line 63: Nit: how about struct destructing here?

timvisee, line 63: Nit: how about struct destructing here?

timvisee, line 63: Nit: how about struct destructing here?

timvisee, line 63: Nit: how about struct destructing here?

timvisee (approved):
Nice! This is better!

timvisee (approved):
Nice! This is better!

timvisee (approved):
Nice! This is better!

timvisee (approved):
Nice! This is better!

timvisee (approved):
Nice! This is better!

timvisee (approved):
Nice! This is better!

lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
agourlay: very good point, it will please Andrey :)

lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
```diff
@@ -16,16 +16,20 @@ use crate::common::Flusher;
...
     db: DatabaseColumnWrapper,
-    deleted_pending_persistence: Mutex<HashSet<Vec<u8>>>,
-    insert_pending_persistence: Mutex<HashMap<Vec<u8>, Vec<u8>>>,
+    pending_operations: Mutex<PendingOperations>, // in-flight operations persisted on flush
+}
+
+#[derive(Debug, Default)]
+struct PendingOperations {
+    deleted: HashSet<Vec<u8>>,
+    inserted: HashMap<Vec<u8>, Vec<u8>>,
```
coszio, line 25: do we even need parallel structures here? Wouldn't it be easier to represent as hashmap of enum?

KShivendu: Do you mean something like this? :thinking: 

```rs
enum Operation {
    Inserted(Vec<u8>),
    Deleted,
}

type PendingOperations = HashMap<Vec<u8>, Operation>;
```

timvisee: Yes.

Insertions have priority over deletions, so a delete entry can just be replaced. If we really want we can add the same debug assertion in a different place.

agourlay: I'd prefer just tackling the Mutex refactoring in the current PR. 
Using a single data structure is a different transformation that should be done in isolation.

PR #5917, opened 2025-01-30
Retry-After on rate limiter hit for REST API

agourlay:
The rate limiter now computes a `retry-after` field in its error to feed to the `Retry-After` HTTP header.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

### DIFF
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -1120,7 +1120,10 @@ impl CollectionError {
                     tokens_available,
                     retry_after,
                 } = retry_error;
-                let description = format!("{rate_limiter_type} rate limit exceeded: Operation requires {cost} tokens but only {tokens_available} were available. Retry later");
+                let description = format!(
+                    "{rate_limiter_type} rate limit exceeded: Operation requires {cost} tokens but only {tokens_available} were available. Retry after {}s",
+                    retry_after.as_secs_f32().ceil() as u32,
+                );
                 (description, Some(retry_after))
             }
         };
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -4,7 +4,7 @@ use std::error::Error as _;
 use std::fmt::{Debug, Write as _};
 use std::iter;
 use std::num::NonZeroU64;
-use std::time::SystemTimeError;
+use std::time::{Duration, SystemTimeError};
 
 use api::grpc::transport_channel_pool::RequestError;
 use api::rest::{
@@ -13,7 +13,7 @@ use api::rest::{
 };
 use common::defaults;
 use common::ext::OptionExt;
-use common::rate_limiting::RateLimitError;
+use common::rate_limiting::{RateLimitError, RetryError};
 use common::types::ScoreType;
 use common::validation::validate_range_generic;
 use io::file_operations::FileStorageError;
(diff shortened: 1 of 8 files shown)


tests/openapi/test_strictmode.py
```diff
@@ -1010,7 +1013,10 @@ def test_strict_mode_write_rate_limiting(collection_name):
         if not response.ok:
             failed_count += 1
             assert response.status_code == 429
-            assert "Rate limiting exceeded: Write rate limit exceeded, retry later" in response.json()['status']['error']
+            assert "Rate limiting exceeded: Write rate limit exceeded" in response.json()['status']['error']
+            assert response.headers['Retry-After'] is not None
+            # need to wait 59s for the single token available to be replenished
+            assert int(response.headers['Retry-After']) is 59
```
timvisee, line 1019: I vote to relax the requirement of 59 a bit. Maybe we can pick the range [50,60]? 60 because of rounding up as described below.

```suggestion
            # need to wait 59s for the single token available to be replenished
            assert int(response.headers['Retry-After']) is 59
```

Maybe we want another rate limit test that asserts we get a low retry-after value if:
- our read limit is high
- we've exhausted reads
- we'll have enough tokens for the next read almost immediately

Wdyt?

src/actix/helpers.rs
```diff
@@ -141,6 +150,38 @@ fn log_service_error(err: &StorageError) {
...
+        match &self.0 {
+            StorageError::RateLimitExceeded {
+                description: _,
+                retry_after,
+            } => {
+                if let Some(retry_after) = retry_after {
+                    // Retry-After is expressed in seconds `https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After`
+                    headers.insert(
+                        header::RETRY_AFTER,
+                        header::HeaderValue::from(retry_after.as_secs()),
```
timvisee, line 165: `as_secs()` rounds down. I don't think that is desired, and we should ceil instead.

Maybe:

```suggestion
                        header::HeaderValue::from(retry_after.as_secs_f32().ceil() as u32),
```

Or something like:

```suggestion
                        header::HeaderValue::from(retry_after.as_secs() + (retry_after.subsec_nanos() > 0) as u64),
```

I find [MDNs page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After#directives) quite unclear on whether the header supports subsecond values. Based on [this](https://httpwg.org/specs/rfc9110.html#field.retry-after) I don't think it does, and we should ceil.

lib/collection/src/operations/types.rs
```diff
@@ -1100,31 +1103,31 @@ impl CollectionError {
...
-                    format!("{rate_limiter_type} rate limit exceeded: Operation requires {} tokens but only {token_available} were available. Retry later", cost.unwrap())
-                } else {
-                    format!("{rate_limiter_type} rate limit exceeded, retry later")
-                }
+            RateLimitError::Retry(retry_error) => {
+                let RetryError {
+                    tokens_available,
+                    retry_after,
+                } = retry_error;
+                let description = format!("{rate_limiter_type} rate limit exceeded: Operation requires {cost} tokens but only {tokens_available} were available. Retry later");
```
timvisee, line 1123: Not sure about this suggestion:

```suggestion
                let description = format!("{rate_limiter_type} rate limit exceeded: Operation requires {cost} tokens but only {tokens_available} were available. Retry after {:?retry_after}");
```

tests/openapi/test_strictmode.py
timvisee: Implemented in a range from 55 to 60.

src/actix/helpers.rs
timvisee: Implemented the first suggestion.

timvisee (approved):
Made a few changes to push it to the finish line. See the (now resolved) comments above.

PR #5918, opened 2025-01-30
Rename `blob_store` as `gridstore`

coszio:
We've been preparing an article with will expose the project publicly. 

I don't love `blob_store` name so much because it has Binary Large OBject baggage, which is not exactly what we tried to mean. "blob" was in there just because we built it for variable-sized values. 

Instead, I am proposing to rename it as `gridstore`. I think this is a better name because it highlights the main idea of its architecture: having fixed-sized blocks of information which we can scope out using other structures, as if it was a grid of data.


agourlay (approved):

timvisee:
Can we now use the waffle emoji for this?

image

PR #5928, opened 2025-02-03
Prototype partial segment snapshots and `SegmentManifest` format

ffuugoo:
This PR prototypes `Segment::take_partial_snapshot` implementation and `SegmentManifest` (structure that is used to calculate which files to include into partial snapshot) format.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/entry/entry_point.rs
+++ b/lib/segment/src/entry/entry_point.rs
@@ -12,6 +12,7 @@ use crate::data_types::named_vectors::NamedVectors;
 use crate::data_types::order_by::{OrderBy, OrderValue};
 use crate::data_types::query_context::{QueryContext, SegmentQueryContext};
 use crate::data_types::vectors::{QueryVector, VectorInternal};
+use crate::entry::partial_snapshot_entry::PartialSnapshotEntry;
 use crate::index::field_index::{CardinalityEstimation, FieldIndex};
 use crate::json_path::JsonPath;
 use crate::telemetry::SegmentTelemetry;
@@ -25,7 +26,7 @@ use crate::types::{
 ///
 /// Assume all operations are idempotent - which means that no matter how many times an operation
 /// is executed - the storage state will be the same.
-pub trait SegmentEntry {
+pub trait SegmentEntry: PartialSnapshotEntry {
     /// Get current update version of the segment
     fn version(&self) -> SeqNumberType;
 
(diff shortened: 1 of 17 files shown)


ffuugoo:
Forgot to `[skip-ci]` this... 🙄

lib/segment/src/segment/entry.rs
```diff
@@ -879,25 +964,108 @@ impl SegmentEntry for Segment {
...
+        if is_updated {
+            updated.insert(path.clone());
+        }
+    }
+
+    updated
+}
+
+const ROCKS_DB_VIRT_FILE: &str = "__ROCKS_DB";
+const PAYLOAD_INDEX_ROCKS_DB_VIRT_FILE: &str = "__PAYLOAD_INDEX_ROCKS_DB";
```
ffuugoo, line 1042: "Fake" files to track RocksDB version in `SegmentManifest`

lib/segment/src/segment/entry.rs
```diff
@@ -837,18 +838,102 @@ impl SegmentEntry for Segment {
...
+                );
+            }
+        }
+
+        // TODO: Version RocksDB!? 🤯
+        files.insert(PathBuf::from(ROCKS_DB_VIRT_FILE), FileVersion::Unversioned);
+        files.insert(
+            PathBuf::from(PAYLOAD_INDEX_ROCKS_DB_VIRT_FILE),
+            FileVersion::Unversioned,
+        );
```
ffuugoo, line 933: Here we insert RocksDB "virtual" files as `Unversioned` into `SegmentManifest`

lib/segment/src/segment/entry.rs
```diff
@@ -879,25 +964,108 @@ impl SegmentEntry for Segment {
...
-        .borrow()
-        .take_database_snapshot(&payload_index_db_backup_path)?;
+    if include_if(PAYLOAD_INDEX_ROCKS_DB_VIRT_FILE.as_ref()) {
+        let payload_index_db_backup_path = temp_path.join(PAYLOAD_DB_BACKUP_PATH);
+
+        segment
+            .payload_index
+            .borrow()
+            .take_database_snapshot(&payload_index_db_backup_path)?;
+    }
```
ffuugoo, line 1069: And here we check if we should include RocksDB backups into partial snapshot.

ffuugoo (commented):
Example of how we can track RocksDB changes for partial snapshots.

The only two places where we special-case anything for RocksDB are:
- when we add RocksDB virtual file into `SegmentManifest`
- and when we check if we should include RocksDB into partial snapshot
  - but that one is so trivial, it's barely worth mention, IMO

We would also need to track when we make changes into RocksDB, but that logic would be pretty much the same to how we would track changes into any other on-disk file, so there won't be anything "special" to RocksDB there.

The logic in `updated_files`, that compares two `SegmentManifest`s and calculates which files should be included into partial snapshot does not require any special casing for RocksDB: it just compares strings and numbers.

ffuugoo:
The whole 2fef99d commit is a _partial_ example, how we could implement per-file version tracking: by doing painstakingly annoying 1000s-of-lines-of-boilerplate-changes refactorings, that propagate `op_num` deep into each operation for each on-disk structure, so that they can track their updates.

I'm not sure if there's a better way to do this, but we've done incredibly annoying changes like this before, so this could be done. 🥲

ffuugoo:
CI is green, I think we can review and merge this then. I think I'll be looking at `take_partial_snapshot`/`get_segment_manifest` for `ProxySegment` next, cause it might need some tweaks to handle proxy segments.

lib/segment/src/segment/entry.rs
timvisee: Random thought: would it be worth to use a character prefix that's even less likely to be used as file name by accident?

Like `::PAYLOAD_INDEX_ROCKS_DB`. Or do we not care?

ffuugoo: Good idea, we can do that.

lib/segment/src/entry/entry_point.rs
```diff
@@ -316,8 +316,74 @@ pub trait SegmentEntry {
...
+mod test {
+    use super::*;
+
+    #[test]
+    fn file_version_serde() {
+        test_file_version_serde(FileVersion::Version(42), "42");
+        test_file_version_serde(FileVersion::Unversioned, "null");
+    }
+
+    fn test_file_version_serde(version: FileVersion, json: &str) {
```
timvisee, line 378: I was a tiny bit confused on what this function did, may be a comment helps:

```suggestion
    /// Test that file version variants are uniquely represented in JSON
    fn test_file_version_serde(version: FileVersion, json: &str) {
```

lib/segment/src/entry/entry_point.rs
```diff
@@ -316,8 +316,74 @@ pub trait SegmentEntry {
...
+    fn get_segment_manifest(&self) -> OperationResult<SegmentManifest>;
+
     // Get collected telemetry data of segment
     fn get_telemetry_data(&self, detail: TelemetryDetail) -> SegmentTelemetry;
 
     fn fill_query_context(&self, query_context: &mut QueryContext);
 }
+
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+pub struct SegmentManifest {
```
generall, line 335: please move to another file

lib/segment/src/segment/entry.rs
```diff
@@ -879,25 +998,108 @@ impl SegmentEntry for Segment {
     }
 }
 
+impl Segment {
```
generall, line 1001: separate file?

generall (approved):

timvisee (approved):

lib/segment/src/segment/entry.rs
ffuugoo: Done.

lib/segment/src/entry/entry_point.rs
ffuugoo: Done.

ffuugoo:
Made some minor cleanups: renamed or moved stuff around (a little bit), added or reworded some comments. No logic changes.

PR #5938, opened 2025-02-03
[mmap sparse vector storage] return cancelled error if stopped

coszio:
This fixes a bug that happened when updating a mmap sparse vector storage is cancelled. 

An invariant was broken when handling the `stopped` flag. All other implementations stop midway and return `OperationError::Cancelled`, but this implementation just stopped iterating, and returned a partial updated range. 

This caused the range check to fail, don't handle cancellation nicely, and send a service error back all the way to the user.

### DIFF
--- a/lib/segment/src/vector_storage/sparse/mmap_sparse_vector_storage.rs
+++ b/lib/segment/src/vector_storage/sparse/mmap_sparse_vector_storage.rs
@@ -13,7 +13,7 @@ use parking_lot::RwLock;
 use sparse::common::sparse_vector::SparseVector;
 
 use super::simple_sparse_vector_storage::SPARSE_VECTOR_DISTANCE;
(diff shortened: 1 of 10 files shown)


lib/segment/src/vector_storage/sparse/mmap_sparse_vector_storage.rs
```diff
@@ -240,6 +240,10 @@ impl VectorStorage for MmapSparseVectorStorage {
             let vector = (!other_deleted).then_some(other_vector);
             self.update_stored(new_id, vector, &hw_counter)?;
         }
+
+        // return cancelled error if stopped
+        check_process_stopped(stopped)?;
+
```
coszio, line 246: Here is the fix

ffuugoo (approved):

PR #5944, opened 2025-02-04
Measure io writes for vector upsertions

JojiiOfficial:
Depends on #5922

Adds measurement for vector storage upsertions

### DIFF
--- a/lib/segment/src/vector_storage/dense/appendable_dense_vector_storage.rs
+++ b/lib/segment/src/vector_storage/dense/appendable_dense_vector_storage.rs
@@ -130,7 +130,7 @@ impl<T: PrimitiveVectorElement, S: ChunkedVectorStorage<T>> VectorStorage
         stopped: &AtomicBool,
     ) -> OperationResult<Range<PointOffsetType>> {
         let start_index = self.vectors.len() as PointOffsetType;
-        let disposed_hw = HardwareCounterCell::disposable(); // This function is only used for operations.
+        let disposed_hw = HardwareCounterCell::disposable(); // This function is only used for internal operations.
         for (other_vector, other_deleted) in other_vectors {
             check_process_stopped(stopped)?;
             // Do not perform preprocessing - vectors should be already processed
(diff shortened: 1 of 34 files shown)


lib/segment/src/vector_storage/dense/appendable_dense_vector_storage.rs
```diff
@@ -124,11 +130,12 @@ impl<T: PrimitiveVectorElement, S: ChunkedVectorStorage<T>> VectorStorage
         stopped: &AtomicBool,
     ) -> OperationResult<Range<PointOffsetType>> {
         let start_index = self.vectors.len() as PointOffsetType;
+        let disposed_hw = HardwareCounterCell::disposable(); // This function is only used for operations.
```
timvisee, line 133: Should it be like this?

```suggestion
        let disposed_hw = HardwareCounterCell::disposable(); // This function is only used during optimization.
```

lib/segment/src/vector_storage/sparse/simple_sparse_vector_storage.rs
```diff
@@ -111,11 +113,15 @@ impl SimpleSparseVectorStorage {
             }
         }
 
+        let key_enc = bincode::serialize(&key).unwrap();
+        let record_enc = bincode::serialize(&record).unwrap();
+
+        hw_counter
+            .vector_io_write_counter()
+            .incr_delta(key_enc.len() + record_enc.len());
```
timvisee, line 121: Are we sure we want to count the key here as well? As far as I'm aware we don't do this for dense vectors.

lib/segment/src/vector_storage/dense/appendable_dense_vector_storage.rs
JojiiOfficial: It should've been "internal operations". Fixed

lib/segment/src/vector_storage/sparse/simple_sparse_vector_storage.rs
JojiiOfficial: We do for dense vectors too: https://github.com/qdrant/qdrant/blob/measure_vector_insert_io_writes/lib/segment/src/vector_storage/dense/simple_dense_vector_storage.rs#L182

I measured it because it is part of what we store and have to write to disk.
Do you think it would be better to not measure it?

timvisee (approved):

PR #5947, opened 2025-02-04
[flushing] take update buffers by reference

coszio:
Our flushing sequence is very clearly defined, and it is important that some components get flushed before others. However, there are some buffered updates which introduce a bit of a different behavior. When we call `flusher()` of a particular component, we expect a closure which will be run at a different time. 

However, as @timvisee pointed out to me by chat, some of those flushers would capture the pending updates at the time the closure is created. This means that the closure they return is returning the state at which `flusher()` is called, not exactly a function which will flush the component at any time.

To fix this, this PR makes them return a reference to their buffers, so that we know for sure that when the closure is called, all its pending changes are also flushed up to that moment.

There are 4 structures which behave this way, and are updated in this PR:
- `MmapBitSliceBufferedUpdateWrapper`
- `MmapSliceBufferedUpdateWrapper`
- `DatabaseColumnScheduledDeleteWrapper`
- `DatabaseColumnScheduledUpdateWrapper`

These changes have already passed one crasher run. Currently testing with another 2 runs, one of 1hr and another one of 6hrs.

### DIFF
--- a/lib/segment/src/common/mmap_bitslice_buffered_update_wrapper.rs
+++ b/lib/segment/src/common/mmap_bitslice_buffered_update_wrapper.rs
@@ -14,7 +14,7 @@ use crate::common::Flusher;
 pub struct MmapBitSliceBufferedUpdateWrapper {
     bitslice: Arc<RwLock<MmapBitSlice>>,
     len: usize,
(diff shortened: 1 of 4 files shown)


coszio:
Second crasher run failed with same error

PR #5951, opened 2025-02-06
IO read measurements for most Payload indices

JojiiOfficial:
Depends on #5949

Basic Payload IO measurement implementation.
Some conditions and indices (eg. bool index) are not fully implemented yet. Those will be addressed in a separate PR.

### DIFF
--- a/lib/api/src/grpc/qdrant.rs
+++ b/lib/api/src/grpc/qdrant.rs
@@ -6490,8 +6490,10 @@ pub struct HardwareUsage {
     #[prost(uint64, tag = "3")]
     pub payload_io_write: u64,
     #[prost(uint64, tag = "4")]
-    pub vector_io_read: u64,
+    pub payload_index_io_read: u64,
     #[prost(uint64, tag = "5")]
+    pub vector_io_read: u64,
+    #[prost(uint64, tag = "6")]
     pub vector_io_write: u64,
 }
 #[derive(serde::Serialize)]
--- a/lib/common/common/src/mmap_hashmap.rs
+++ b/lib/common/common/src/mmap_hashmap.rs
@@ -16,6 +16,7 @@ use rand::Rng as _;
 use rand::rngs::StdRng;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
+use crate::counter::hardware_counter::HardwareCounterCell;
 use crate::zeros::WriteZerosExt as _;
 
 type ValuesLen = u32;
@@ -241,12 +242,20 @@ impl<K: Key + ?Sized, V: Sized + FromBytes + Immutable + IntoBytes + KnownLayout
     }
 
     /// Get the values associated with the `key`.
-    pub fn get(&self, key: &K) -> io::Result<Option<&[V]>> {
+    pub fn get(&self, key: &K, hw_counter: &HardwareCounterCell) -> io::Result<Option<&[V]>> {
         let Some(hash) = self.phf.get(key) else {
             return Ok(None);
         };
 
+        hw_counter
+            .payload_index_io_read_counter()
+            .incr_delta(size_of::<BucketOffset>());
+
         let entry = self.get_entry(hash as usize)?;
+
+        hw_counter
+            .payload_index_io_read_counter()
+            .incr_delta(size_of_val(entry));
 
         if !key.matches(entry) {
             return Ok(None);
@@ -508,9 +517,10 @@ mod tests {
         let mmap = MmapHashMap::<K, u32>::open(&tmpdir.path().join("map")).unwrap();
 
         // Non-existing keys should return None
+        let hw_counter = HardwareCounterCell::new();
         for _ in 0..1000 {
             let key = repeat_until(|| generator(&mut rng), |key| !map.contains_key(key));
-            assert!(mmap.get(as_ref(&key)).unwrap().is_none());
+            assert!(mmap.get(as_ref(&key), &hw_counter).unwrap().is_none());
         }
 
         // check keys iterator
@@ -529,7 +539,7 @@ mod tests {
         // Existing keys should return the correct values
         for (k, v) in map {
             assert_eq!(
-                mmap.get(as_ref(&k)).unwrap().unwrap(),
+                mmap.get(as_ref(&k), &hw_counter).unwrap().unwrap(),
                 &v.into_iter().collect::<Vec<_>>()
             );
         }
@@ -541,6 +551,7 @@ mod tests {
         let tmpdir = tempfile::Builder::new().tempdir().unwrap();
 
         let mut map: HashMap<i64, BTreeSet<u64>> = Default::default();
+        let hw_counter = HardwareCounterCell::new();
 
         for key in 0..10i64 {
             map.insert(key, (0..100).map(|_| rng.random_range(0..=1000)).collect());
@@ -556,11 +567,11 @@ mod tests {
 
         for (k, v) in map {
             assert_eq!(
-                mmap.get(&k).unwrap().unwrap(),
+                mmap.get(&k, &hw_counter).unwrap().unwrap(),
                 &v.into_iter().collect::<Vec<_>>()
             );
         }
 
-        assert!(mmap.get(&100).unwrap().is_none())
+        assert!(mmap.get(&100, &hw_counter).unwrap().is_none())
     }
 }
--- a/lib/common/common/src/mmap_hashmap.rs
+++ b/lib/common/common/src/mmap_hashmap.rs
@@ -58,6 +58,8 @@ struct Header {
 
 const PADDING_SIZE: usize = 4096;
 
+pub const BUCKET_OFFSET_OVERHEAD: usize = size_of::<BucketOffset>();
+
 type BucketOffset = u64;
 
 impl<K: Key + ?Sized, V: Sized + FromBytes + Immutable + IntoBytes + KnownLayout>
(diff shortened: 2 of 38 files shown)


lib/common/common/src/mmap_hashmap.rs
```diff
@@ -241,13 +244,21 @@ impl<K: Key + ?Sized, V: Sized + FromBytes + Immutable + IntoBytes + KnownLayout
...
 
+        hw_counter
+            .payload_index_io_read_counter()
+            .incr_delta(BUCKET_OFFSET_OVERHEAD);
+
         let entry = self.get_entry(hash as usize)?;
 
+        hw_counter
+            .payload_index_io_read_counter()
+            .incr_delta(size_of_val(entry));
```
timvisee, line 260: Since it's just a `&[u8]` we can take it's length directly.

```suggestion
            .incr_delta(entry.len());
```

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
```diff
@@ -198,7 +207,8 @@ impl<N: MapIndexKey + Key + ?Sized> MmapMapIndex<N> {
     }
 
     pub fn get_count_for_value(&self, value: &N) -> Option<usize> {
-        match self.value_to_points.get(value) {
+        let hw_counter = HardwareCounterCell::disposable(); // TODO(io_measurement): Maybe propagate?
```
timvisee, line 210: Yeah I think we need to count this, IIRC its used in the facet API and it can be expensive.

lib/segment/src/index/query_optimization/condition_converter/match_converter.rs
```diff
@@ -9,38 +10,47 @@ use crate::types::{
...
+            let hw_counter = hw_acc.get_counter_cell();
             Some(Box::new(move |point_id: PointOffsetType| {
-                index.check_values_any(point_id, |i| *i == value)
+                index.check_values_any(point_id, &hw_counter, |i| *i == value)
             }))
         }
         (ValueVariants::Bool(is_true), FieldIndex::BoolIndex(index)) => {
             Some(Box::new(move |point_id: PointOffsetType| {
+                // TODO(io_measurement): implement measurement
                 index.check_values_any(point_id, is_true)
```
timvisee, line 54: Can't we count just 1 for a boolean check?

In the description I see that you're handling this in a separate PR, fine by me. :+1:

timvisee (approved):

lib/segment/src/index/field_index/map_index/immutable_map_index.rs
```diff
@@ -247,9 +250,23 @@ impl<N: MapIndexKey + ?Sized> ImmutableMapIndex<N> {
...
+        check_fn: impl Fn(&N) -> bool,
+    ) -> bool {
+        // Overhead of accessing the index.
+        hw_counter
+            .payload_index_io_read_counter()
+            .incr_delta(MMAP_PTV_ACCESS_OVERHEAD);
+
+        self.point_to_values.check_values_any(idx, |v| {
+            let v: &N = v.borrow();
+            let size = <N as MmapValue>::mmapped_size((*v).as_referenced());
```
coszio, line 266: Can't we get this size outside the hot path?

coszio (commented):
I am not sure if we should consider entire slices when we iterate only a few of their values. 

Just in case, I just made this iterator adapter (#5985) which should make it easier (and faster considering it is not needed to have any smart pointers in the hot path) to measure the actual amount of iterations used.

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -57,6 +58,8 @@ struct Header {
 
 const PADDING_SIZE: usize = 4096;
 
+pub const BUCKET_OFFSET_OVERHEAD: usize = size_of::<BucketOffset>();
```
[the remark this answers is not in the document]
JojiiOfficial: I'm using this value in multiple places and thought all having the same source of truth would be easier to extend in future

lib/segment/src/index/query_optimization/condition_converter/match_converter.rs
JojiiOfficial: In theory it is more than 1 byte though because we load a full page into memory, even if we only check one byte.
Can we maybe go probabilistic here and say 99% of the time we access the same page so we use
`0.01 *  PAGE_SIZE` (4kb) as constant measurement.
@generall wdyt?

lib/segment/src/index/field_index/map_index/immutable_map_index.rs
JojiiOfficial: Not really. If `N` is a `str`, it may always have a different size so either we do it here or iterate over the values twice.

lib/segment/src/index/query_optimization/condition_converter/match_converter.rs
JojiiOfficial: Looks like it's already implemented with 1 byte for boolean here: https://github.com/qdrant/qdrant/pull/5954/files#diff-2bee7d830875853065cb32b763bc13bc85fca02f0dbab25f7018d251fd192d20R186

I think it's still worth reconsidering to use > 1 byte.

lib/segment/src/index/field_index/map_index/immutable_map_index.rs
agourlay: Any idea how expensive this could get?
Looking around I could not find a bench for the immutable map index unfortunately

lib/api/src/grpc/qdrant.rs
```diff
@@ -6490,8 +6490,10 @@ pub struct HardwareUsage {
     #[prost(uint64, tag = "3")]
     pub payload_io_write: u64,
     #[prost(uint64, tag = "4")]
-    pub vector_io_read: u64,
+    pub payload_index_io_read: u64,
     #[prost(uint64, tag = "5")]
+    pub vector_io_read: u64,
+    #[prost(uint64, tag = "6")]
```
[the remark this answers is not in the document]
JojiiOfficial: Not used anywhere yet so it's safe for us to change them

lib/segment/src/index/field_index/bool_index/mmap_bool_index.rs
```diff
@@ -367,15 +368,19 @@ impl PayloadFieldIndex for MmapBoolIndex {
...
-                self.get_slice_for(*value)
-                    .iter_ones()
-                    .map(|x| x as PointOffsetType),
-            )),
+            })) => {
+                let slice = self.get_slice_for(*value);
+                hw_counter
+                    .payload_index_io_read_counter()
+                    .incr_delta(slice.len()); // We have to iterate over the whole slice
+                Some(Box::new(slice.iter_ones().map(|x| x as PointOffsetType)))
```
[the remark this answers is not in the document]
JojiiOfficial: Fixed in later PR!

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
```diff
@@ -212,13 +222,24 @@ impl<N: MapIndexKey + Key + ?Sized> MmapMapIndex<N> {
...
+                hw_counter
+                    .payload_index_io_read_counter()
+                    .incr_delta(size_of_val(slice));
+
+                Box::new(
+                    slice
+                        .iter()
+                        .filter(|idx| !self.deleted.get(**idx as usize).unwrap_or(false)),
+                )
+            }
```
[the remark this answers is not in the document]
JojiiOfficial: We're passing the slice, not a reference to the slice. Thus it measures the same as `len*element-size`.
It's actually a [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#manual_slice_size_calculation) and changing it to this suggestion will cause a linter error.

lib/segment/src/index/field_index/geo_index/mod.rs
```diff
@@ -181,12 +183,13 @@ impl GeoMapIndex {
...
         check_fn: impl Fn(&GeoPoint) -> bool,
     ) -> bool {
         match self {
             GeoMapIndex::Mutable(index) => index.check_values_any(idx, check_fn),
-            GeoMapIndex::Immutable(index) => index.check_values_any(idx, check_fn),
-            GeoMapIndex::Mmap(index) => index.check_values_any(idx, check_fn),
+            GeoMapIndex::Immutable(index) => index.check_values_any(idx, hw_counter, check_fn),
+            GeoMapIndex::Mmap(index) => index.check_values_any(idx, hw_counter, check_fn),
         }
     }
```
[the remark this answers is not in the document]
JojiiOfficial: Implemented in later PR

lib/segment/src/index/field_index/map_index/immutable_map_index.rs
JojiiOfficial: I haven't measured it directly but the size calculation is as I said unavoidable since `N` can be of type `str`.

We probably can improve performance a bit by using an integer for collecting the total size and accumulate this value into the HwCounter afterwards but I'd suggest to address this in a separate PR and focus on merging the PRs for now.

lib/segment/src/index/field_index/bool_index/mmap_bool_index.rs
```diff
@@ -367,15 +368,19 @@ impl PayloadFieldIndex for MmapBoolIndex {
...
-            })) => Some(Box::new(
-                self.get_slice_for(*value)
-                    .iter_ones()
-                    .map(|x| x as PointOffsetType),
-            )),
+            })) => {
+                let slice = self.get_slice_for(*value);
+                hw_counter
+                    .payload_index_io_read_counter()
+                    .incr_delta(slice.len()); // We have to iterate over the whole slice
```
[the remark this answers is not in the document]
JojiiOfficial: Fixed in #6132

coszio (approved):

agourlay (approved):

PR #5953, opened 2025-02-07
Tiny refactoring to reduce some boilerplate in update API

ffuugoo:
Some trivial tweaks to update API to slash a few lines of boilerplate:

- simplify `wait` and `ordering` query parameters deserialization in REST API
  - this does not affect OpenAPI spec, because these parameters are *explicitly* described there
- use `common::inference::extract_token` in internal gRPC API
  - (as far as I can tell, it's *exactly* the same, as a copy-pasted snippet that was used in internal gRPC API before)
- destructure `UpdateVectorsInternal`/`DeleteVectorsInternal` requests instead of accessing fields individually
- add generic `extract_internal_request` helper

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/actix/api/update_api.rs
+++ b/src/actix/api/update_api.rs
@@ -30,9 +30,11 @@ struct FieldPath {
 }
 
 #[derive(Deserialize, Serialize, JsonSchema, Validate)]
-pub struct UpdateParam {
-    pub wait: Option<bool>,
-    pub ordering: Option<WriteOrdering>,
+pub struct UpdateParams {
+    #[serde(default)]
+    pub wait: bool,
+    #[serde(default)]
+    pub ordering: WriteOrdering,
 }
 
 #[put("/collections/{name}/points")]
@@ -40,7 +42,7 @@ async fn upsert_points(
     dispatcher: web::Data<Dispatcher>,
     collection: Path<CollectionPath>,
     operation: Json<PointInsertOperations>,
-    params: Query<UpdateParam>,
+    params: Query<UpdateParams>,
     service_config: web::Data<ServiceConfig>,
     ActixAccess(access): ActixAccess,
     inference_token: InferenceToken,
@@ -52,8 +54,7 @@ async fn upsert_points(
         };
 
     let operation = operation.into_inner();
-    let wait = params.wait.unwrap_or(false);
-    let ordering = params.ordering.unwrap_or_default();
+    let UpdateParams { wait, ordering } = params.into_inner();
 
     let request_hw_counter = get_request_hardware_counter(
         &dispatcher,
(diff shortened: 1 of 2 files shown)


src/actix/api/update_api.rs
```diff
@@ -30,17 +30,19 @@ struct FieldPath {
 }
 
 #[derive(Deserialize, Serialize, JsonSchema, Validate)]
-pub struct UpdateParam {
-    pub wait: Option<bool>,
-    pub ordering: Option<WriteOrdering>,
+pub struct UpdateParams {
```
generall, line 33: This will cause changes in OpenAPI

timvisee: It doesn't because these params are explicitly defined. The consistency check CI job passes :heavy_check_mark:

timvisee (approved):

src/actix/api/update_api.rs
```diff
@@ -30,17 +30,19 @@ struct FieldPath {
...
+    #[serde(default)]
```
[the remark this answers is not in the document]
ffuugoo: Why though?

- we unwrap it exactly the same way everywhere
  - and it *actually* ensures, that each API handles parameters *the same* way
- it does not change current behavior at all.
- if we ever *need* to handle `None`, we can simply put `Option`s back
- if any specific API needs to handle `None` or defaults differently, we can add `SpecialNeedsUpdateParams` and change single API endpoint

It's literally 100% positive improvement over `Option`s.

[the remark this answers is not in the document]
coszio: Python REST client generation looks fine as well 👍

src/tonic/api/points_internal_api.rs
```diff
@@ -273,18 +254,15 @@ impl PointsInternal for PointsInternalService {
...
-        let clock_tag = request.clock_tag;
-
-        let delete_point_vectors = request
-            .delete_vectors
-            .ok_or_else(|| Status::invalid_argument("DeleteVectors is missing"))?;
+        let DeleteVectorsInternal {
+            delete_vectors: delete_vectors_req,
+            shard_id,
+            clock_tag,
+        } = request.into_inner();
```
coszio, line 261: Nice destructuring improvement

coszio (approved):

src/actix/api/update_api.rs
generall: Name of the structure in Rust vs name of the structure if OpenAPI doesn't match?

ffuugoo: It's not even referenced by OpenAPI. It's not in the `schema_generator`.

https://github.com/qdrant/qdrant/blob/master/openapi/openapi-points.ytt.yaml#L88-L99.

generall: why `#[derive(..., JsonSchema, ..)]` then?

ffuugoo: 🤷‍♀️

ffuugoo: I've removed `JsonSchema` derive from `UpdateParams`, CI is still green.

src/actix/api/update_api.rs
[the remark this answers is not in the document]
generall: My concerns are mostly applied to parameters we actually store, like collection config e.t.c.
In this case it might be fine

generall (approved):

PR #5955, opened 2025-02-07
[id_tracker] Don't introduce synthetic versions

coszio:
Our id trackers currently can't discriminate between version 0 as real version, and 0 as a not-yet-set version. 

The latter case happens when the `internal_to_version` vector gets resized. It gets filled with zeros, and there is no indication whether they represent actual versions or not. 

To fix this, I propose to grow the vector with `u64::MAX`, so that version 0 is still valid and don't break anything, but also to know which versions are not actually set yet. This special number is safe to use because I've introduced a wrapper type for this purpose:
```rust
#[repr(transparent)]
pub struct VersionType(SeqNumberType);
```

This PR swaps the type of `internal_to_version` to `Vec<VersionType>`, so that the newtype checks whether it has been synthetically initialized, or it is an actual version.

NOTE: this is potentially problematic if there is a qdrant rollback, as these versions would be considered as the latest.

### DIFF
--- a/lib/segment/src/id_tracker/immutable_id_tracker.rs
+++ b/lib/segment/src/id_tracker/immutable_id_tracker.rs
@@ -651,7 +651,7 @@ pub(super) mod test {
 
             assert_eq!(
                 id_tracker.internal_version(internal_id),
-                Some((*expect_version).into())
+                Some(*expect_version)
             );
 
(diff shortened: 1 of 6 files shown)


coszio:
> this is potentially problematic if there is a qdrant rollback, as these versions would be considered as the latest.

PR #5956, opened 2025-02-07
Apply point deletions to all point versions

timvisee:
Fix a critical problem which could result in us reviving (partial) already deleted points.

A point may exist in multiple segments, having multiple versions. This is normal behavior and can happen due to a myriad of reasons.

Normally, point operations are only [applied](https://github.com/qdrant/qdrant/blob/997ef849ae983282e7bbf804c8aba702c47da6bb/lib/collection/src/collection_manager/holders/segment_holder.rs#L487-L511) to the latest point version. This is done on [purpose](https://github.com/qdrant/qdrant/blob/997ef849ae983282e7bbf804c8aba702c47da6bb/lib/collection/src/collection_manager/holders/segment_holder.rs#L464-L466), so that any copy-on-write operation always uses the latest point data as source.

The problem is in deleting points though. If we only delete the latest point version, we may leave older point versions behind. We never use these older versions, but we might now accidentally revive them if we forget the newer deleted version. To prevent this problem, we now delete all point versions.

To accommodate this change, I've extracted the point/segment selection logic into a separate function. It takes care of either selecting all point versions, or just the latest.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/segments_updater.rs
+++ b/lib/collection/src/collection_manager/segments_updater.rs
@@ -127,8 +127,8 @@ pub(crate) fn delete_vectors(
                 }
                 Ok(res)
             },
-            // Apply delete to all point versions
-            true,
+            // Only apply operation to latest point versions, operation does not delete points
+            false,
         )?;
 
         total_deleted_points += deleted_points;
--- a/lib/collection/src/collection_manager/segments_updater.rs
+++ b/lib/collection/src/collection_manager/segments_updater.rs
@@ -51,6 +51,8 @@ pub(crate) fn delete_points(
             batch,
             |_| (),
             |id, _idx, write_segment, ()| write_segment.delete_point(op_num, id, hw_counter),
+            // Apply point delete to all point versions
+            false,
         )?;
 
         total_deleted_points += deleted_points;
@@ -125,6 +127,8 @@ pub(crate) fn delete_vectors(
                 }
                 Ok(res)
             },
+            // Apply delete to all point versions
+            false,
         )?;
 
         total_deleted_points += deleted_points;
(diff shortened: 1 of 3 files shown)


lib/collection/src/collection_manager/segments_updater.rs
```diff
@@ -51,6 +51,8 @@ pub(crate) fn delete_points(
             batch,
             |_| (),
             |id, _idx, write_segment, ()| write_segment.delete_point(op_num, id, hw_counter),
+            // Apply point delete to all point versions
+            true,
```
timvisee, line 55: Critical bit: for point delete operations we use `true` to apply the delete to all point versions.

For other operations this flag is false, meaning it only applies to the latest version as expected.

coszio (approved):

lib/collection/src/collection_manager/segments_updater.rs
```diff
@@ -125,6 +127,8 @@ pub(crate) fn delete_vectors(
                 }
                 Ok(res)
             },
+            // Apply delete to all point versions
+            true,
```
generall, line 131: I think this should be `false`. Points with different versions might have different payloads, deleting vectors will create unresolvable conflict on merge

generall (changes requested):
see comment

lib/collection/src/collection_manager/segments_updater.rs
timvisee: Do you mean in case a payload filter is used?

I'm not too worried about it as we should still only use the latest point version for retrieval.

But since this is not strictly related to point deletion, and since this on its own cannot revive old keys, I'm happy with changing this one to false 👍

timvisee:
Three crasher runs:
- https://github.com/qdrant/crasher/actions/runs/13210732687
- https://github.com/qdrant/crasher/actions/runs/13213841462
- https://github.com/qdrant/crasher/actions/runs/13213865925

lib/collection/src/collection_manager/segments_updater.rs
timvisee: f492b6c3f86e309134dda2f8ebe43eecd2c16a02

PR #5959, opened 2025-02-07
[id tracker] mark skipped versions as deleted

coszio:
Our id trackers currently can't discriminate between version 0 as real version, and 0 as a not-yet-set version.

The latter case happens when the internal_to_version vector gets resized. It gets filled with zeros, and there is no indication whether they represent actual versions or not.

To fix it, this PR makes sure that the corresponding skipped points are marked as deleted, so that a zero version can be known to be synthetic if the bitvec says it is deleted.

### DIFF
--- a/lib/segment/src/id_tracker/simple_id_tracker.rs
+++ b/lib/segment/src/id_tracker/simple_id_tracker.rs
@@ -147,7 +147,8 @@ impl SimpleIdTracker {
             }
         }
 
-        let mut internal_to_version: Vec<SeqNumberType> = Default::default();
+        let mut internal_to_version: Vec<SeqNumberType> =
+            Vec::with_capacity(internal_to_external.len());
         let versions_db_wrapper = DatabaseColumnScheduledUpdateWrapper::new(
             DatabaseColumnWrapper::new(store, DB_VERSIONS_CF),
         );
--- a/lib/segment/src/id_tracker/simple_id_tracker.rs
+++ b/lib/segment/src/id_tracker/simple_id_tracker.rs
@@ -10,6 +10,7 @@ use rocksdb::DB;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
+use super::ensure_versions_len;
 use crate::common::Flusher;
 use crate::common::operation_error::OperationResult;
 use crate::common::rocksdb_buffered_update_wrapper::DatabaseColumnScheduledUpdateWrapper;
@@ -158,9 +159,7 @@ impl SimpleIdTracker {
                 PointIdType::Uuid(uuid) => external_to_internal_uuid.get(&uuid).copied(),
             };
             if let Some(internal_id) = internal_id {
-                if internal_id as usize >= internal_to_version.len() {
-                    internal_to_version.resize(internal_id as usize + 1, 0);
-                }
+                ensure_versions_len(internal_id, &mut internal_to_version, &mut deleted);
                 internal_to_version[internal_id as usize] = version;
             } else {
                 log::debug!(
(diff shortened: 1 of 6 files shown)


coszio:
Corresponding 6hr crasher run:
https://github.com/qdrant/crasher/actions/runs/13205653137

coszio:
Crasher still show same error

coszio:
Ok, now there is a corresponding [6hr green crasher run](https://github.com/qdrant/crasher/actions/runs/13244116271). 

This time I did two more things:
- drop the mappings too, when resizing the vec on `set_internal_version`.
- on load, populate `internal_to_version` by iterating through `internal_to_external`. This way we are sure these two are in sync. If there is no corresponding version, it marks it as deleted.

lib/segment/src/id_tracker/simple_id_tracker.rs
```diff
@@ -150,25 +151,26 @@ impl SimpleIdTracker {
...
-                    internal_to_version.resize(internal_id as usize + 1, 0);
-                }
-                internal_to_version[internal_id as usize] = version;
-            } else {
-                log::debug!(
-                    "Found version: {} without internal id, external id: {}",
-                    version,
-                    external_id
-                );
+        for (internal, external) in internal_to_external.iter().enumerate() {
```
timvisee, line 155: If we extend the version slice, we drop mappings that don't have any version. When loading the simple ID tracker we mark mappings missing a version as deleted.

Why is the action we take in both cases different? Can't we drop the mappings in case of loading the simple ID tracker as well?

timvisee (changes requested):

timvisee:
@generall This disables points having no version (drops mapping or marks as deleted). Can you imagine a scenario where this is problematic?

lib/segment/src/id_tracker/simple_id_tracker.rs
coszio: Yeah, it is a good point. I did not drop anything here because it affects the persistent state, while the method is only supposed to load. 

Do you think we should perform repair as well on load?

coszio: I decided to drop the in-memory mapping without removing it from rocksdb

lib/segment/src/id_tracker/simple_id_tracker.rs
```diff
@@ -154,9 +154,9 @@ impl SimpleIdTracker {
         );
         for (internal, external) in internal_to_external.iter_mut().enumerate() {
             if deleted[internal] {
-                log::debug!("Adding synthetic version 0 to a deleted point, external id: {external}, internal id: {internal}");
                 // Add synthetic version 0
                 internal_to_version.push(0);
+                log::debug!("Added synthetic version 0 to a deleted point, external id: {external}, internal id: {internal}");
```
timvisee, line 159: Maybe its better to remove this one, I can imagine it to be noisy.

We don't want 100k log lines if we have 1 million points with 10% deleted.

```suggestion
```

timvisee (approved):

lib/segment/src/id_tracker/simple_id_tracker.rs
coszio: Good point

timvisee:
CI is being funny again. Retriggered it just now.

coszio:
@generall what are your thoughts on this? This PR might need to be updated in case we want to merge it since we now have the new compressed structures

coszio:
Since there is uncertainty of whether these changes are desired or not, I will close the PR. We can revisit it later if there comes a good argument to include it.

PR #5962, opened 2025-02-10
Always delete old point versions when applying point operation

timvisee:
Continuation after <https://github.com/qdrant/qdrant/pull/5956> as suggested by <https://github.com/qdrant/qdrant/pull/5956#issuecomment-2645360891>.

An initial implementation for always deleting older point versions when an operation is applied. It cleans up duplicate points and prevents users retrieving old point data.

This is a first iteration, and I'm not super happy with the implementation yet. However, it is good enough to be included in a patch release.

This PR has the following properties:
- does not delete other point instances having the same version
- deletes old points with their existing version

In a future PR I'd like to change these properties to:
- do delete other point instances having the same version, keeping just one instance
- delete old point versions with the operation version (to also bump the segment version)

However, implementing these extra changes appears to be entirely non trivial. Naively implementing it makes many weird problem appear, such as losing points if we have multiple proxies using the same write segment.

I've added a test and adjusted an existing test to assert correct behavior of this change.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -590,6 +590,17 @@ impl<'s> SegmentHolder {
 
         let segment_points = self.find_points_to_update_and_delete(ids);
 
+        // Delete old points first
+        for (&segment_id, points) in &segment_points {
+            let segment = self.get(segment_id).unwrap();
+            let segment_arc = segment.get();
+            let mut write_segment = segment_arc.write();
+
+            for &point_id in &points.to_delete {
+                point_delete_operation(point_id, &mut write_segment)?;
+            }
+        }
+
         // Apply point operations to selected segments
         let mut applied_points = 0;
         for (segment_id, points) in segment_points {
@@ -602,10 +613,6 @@ impl<'s> SegmentHolder {
                 let is_applied =
                     point_operation(point_id, segment_id, &mut write_segment, &segment_data)?;
                 applied_points += usize::from(is_applied);
-            }
-
-            for point_id in points.to_delete {
-                point_delete_operation(point_id, &mut write_segment)?;
             }
         }
 
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -404,46 +404,46 @@ impl<'s> SegmentHolder {
             .collect()
     }
 
-    /// Select what point IDs are in what segments.
+    /// Select what point IDs to update in each segment
     ///
-    /// For each external id, picks the segment which has the latest version of such point.
-    fn latest_segments_points(
-        &self,
-        ids: &[PointIdType],
-    ) -> (
-        AHashMap<SegmentId, Vec<PointIdType>>,
-        AHashMap<SegmentId, Vec<PointIdType>>,
-    ) {
-        // Find in which segments latest point versions are located
-        let mut points: AHashMap<PointIdType, (SeqNumberType, SegmentId)> =
+    /// For each external id, pick the segment which has the latest version of such point.
+    /// Older point versions are also returned and marked to be deleted.
+    ///
+    /// Points that are already soft deleted are not included.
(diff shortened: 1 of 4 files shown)


lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -547,25 +582,32 @@ impl<'s> SegmentHolder {
...
+        // Delete old points first
+        for (segment_id, points) in to_delete {
+            let segment = self.get(segment_id).unwrap();
+            let segment_arc = segment.get();
+            let mut write_segment = segment_arc.write();
+
+            for point_id in points {
+                point_delete_operation(point_id, &mut write_segment)?;
+            }
+        }
```
timvisee, line 600: Here we now delete older versions before applying operations to the newest point version.

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -404,61 +405,95 @@ impl<'s> SegmentHolder {
...
     ///
-    /// For each external id, picks the segment which has the latest version of such point.
-    fn latest_segments_points(&self, ids: &[PointIdType]) -> AHashMap<SegmentId, Vec<PointIdType>> {
-        // Find in which segments latest point versions are located
-        let mut points: AHashMap<PointIdType, (SeqNumberType, SegmentId)> =
+    /// This finds all point versions and groups them per segment. The newest point versions are
+    /// selected to be updated, all older versions are marked to be deleted.
+    ///
+    /// Points that are already soft deleted are not included.
+    fn find_points_to_update_and_delete(
```
timvisee, line 416: Function is adjusted to now collect both the latest point versions, and all older point versions.

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -404,61 +405,95 @@ impl<'s> SegmentHolder {
...
+                        if old_version < point_version {
+                            to_delete
+                                .entry(old_segment_id)
+                                .or_default()
+                                .push(segment_point);
                         }
-                        // Point we have seen before is newer, do nothing
-                        Entry::Occupied(_) => {}
+                    }
+                    // Ignore points with the same version, only update one of them
```
generall, line 456: That means that after update operation, one of the points became outdated, right?

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -547,25 +582,32 @@ impl<'s> SegmentHolder {
...
 
-        // Select what points to update in what segments
-        let segment_points = if all_point_versions {
-            self.all_segments_points(ids)
-        } else {
-            self.latest_segments_points(ids)
-        };
+        let (to_update, to_delete) = self.find_points_to_update_and_delete(ids);
+
+        // Delete old points first
```
generall, line 591:
```suggestion
// Delete old points first, because we want to handle copy-on-write in multiple proxy segments properly
```

lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -532,12 +567,12 @@ impl<'s> SegmentHolder {
...
     /// The `segment_data` function is called no more than once for each segment and its result is
     /// passed to `point_operation`.
-    pub fn apply_points<T, O, D>(
+    pub fn apply_points<T, D, O, OD>(
         &self,
         ids: &[PointIdType],
         mut segment_data: D,
         mut point_operation: O,
-        all_point_versions: bool,
+        mut point_delete_operation: OD,
```
generall, line 575: Why can't we just delete points without external function?

lib/collection/src/collection_manager/holders/segment_holder.rs
timvisee: Correct. We changed this to update all points having the highest version.

lib/collection/src/collection_manager/holders/segment_holder.rs
timvisee: I've removed the closure, it now deletes internally

generall (approved):

PR #5966, opened 2025-02-11
WIP: Delete old points with current operation version

timvisee:
Continuation of <https://github.com/qdrant/qdrant/pull/5962>

_WIP_

### Tasks

- [ ] Finish PR
- [x] Merge <https://github.com/qdrant/qdrant/pull/5962>
- [x] Rebase on `dev`
- [ ] Undraft

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -164,11 +164,13 @@ impl ProxySegment {
                 return Ok(false);
             };
 
+            let local_version = cmp::max(local_version, op_num);
+
             // Equal or higher point version is already moved into write segment - delete from
             // wrapped segment and do not move it again
             if deleted_points_guard
                 .get(&point_id)
-                .is_some_and(|&deleted| deleted.local_version >= local_version)
+                .is_some_and(|&deleted| deleted.local_version > local_version)
             {
                 drop(deleted_points_guard);
                 self.set_deleted_offset(point_offset);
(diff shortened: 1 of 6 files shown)


timvisee:
We have a flaky test due to these changes:

```bash
pytest -p no:cacheprovider ./tests/consensus_tests/test_snapshot_consistency.py -k test_shard_wal_delta_transfer_manual_recovery
```

It is unclear to me why this fails. But it seems to happen because we create a whole bunch of proxies while snapshotting segments.

lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -164,11 +164,13 @@ impl ProxySegment {
...
             };
 
+            let local_version = cmp::max(local_version, op_num);
+
             // Equal or higher point version is already moved into write segment - delete from
             // wrapped segment and do not move it again
             if deleted_points_guard
                 .get(&point_id)
-                .is_some_and(|&deleted| deleted.local_version >= local_version)
+                .is_some_and(|&deleted| deleted.local_version > local_version)
```
timvisee, line 173: Needs further work, but the idea is that we don't rely on the local point versions anymore. Instead we're interested in whether our point is newer than the current operation version.

timvisee:
This change appears to be problematic when multiple proxies are involved. It is unclear what the exact problem is.

The [failing test](https://github.com/qdrant/qdrant/pull/5966#issuecomment-2653726404) shows things are not sound yet.

We therefore decided to drop this idea for now, and keep the current implementation as worked on in <https://github.com/qdrant/qdrant/pull/5962>.

We don't think not bumping the segment versions in which we apply deletions is a huge problem. In the very worst case, deletes aren't picked up (in incremental snapshots for example), but we have other things in place to safe guard this.

PR #5967, opened 2025-02-11
Add strict mode for max points count

agourlay:
New strict mode config for the max number of points per collection.

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -7338,6 +7338,13 @@
             "minimum": 0,
             "nullable": true
           },
+          "max_points_count": {
+            "description": "Max number of points in a collection",
+            "type": "integer",
+            "format": "uint",
+            "minimum": 1,
+            "nullable": true
+          },
           "filter_max_conditions": {
             "description": "Max conditions a filter can have.",
             "type": "integer",
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -7339,7 +7339,7 @@
             "nullable": true
           },
           "max_points_count": {
-            "description": "Max number of points in a collection",
+            "description": "Max number of points estimated in a collection",
             "type": "integer",
             "format": "uint",
             "minimum": 1,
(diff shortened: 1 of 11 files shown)


lib/collection/src/common/collection_size_stats.rs
```diff
@@ -104,31 +114,38 @@ pub struct CollectionSizeStats {
     pub vector_storage_size: usize,
     /// Estimated amount of payload storage size.
     pub payload_storage_size: usize,
+    /// Estimated amount of points.
+    pub points_count: usize,
```
coszio, line 118: Good to mention it is estimated

docs/redoc/master/openapi.json
```diff
@@ -7338,6 +7338,13 @@
             "minimum": 0,
             "nullable": true
           },
+          "max_points_count": {
+            "description": "Max number of points in a collection",
```
coszio, line 7342: Maybe we should mention that it is an estimated number here too

coszio (approved):

lib/collection/src/operations/verification/update.rs
```diff
@@ -239,6 +240,24 @@ async fn check_collection_size_limit(
         check_collection_payload_size_limit(payload_storage_size_limit_bytes, stats)?;
     }
 
+    if let Some(points_count_limit) = point_limit {
```
timvisee, line 243: Would it make sense to include the number of points the current request might add? As far as I can see we're not doing that.

Imagine we set a point limit of 1. It looks like the first request is able to insert a 100 points in one operation, and only then we're limiting point count. Or am I wrong on that?

agourlay: You are right, it would be better to not assume each request is equivalent to adding one point.
I will see what  I can do here :+1:

timvisee: @agourlay I'd be fine with handling that in a separate PR, so we keep this one simple and clean.

I don't deem this super critical, but rather a nice to have.

timvisee (approved):

PR #5978, opened 2025-02-12
Extend partial segment snapshot to support `ProxySegment`s

ffuugoo:
Based on #5928. Changes `SegmentEntry::take_partial_snapshot` (and related methods) signature and implementation a bit, to support `ProxySegment`s.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/data_types/segment_manifest.rs
+++ b/lib/segment/src/data_types/segment_manifest.rs
@@ -5,6 +5,7 @@ use crate::types::SeqNumberType;
 
 #[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
 pub struct SegmentManifest {
+    pub segment_id: String,
     pub segment_version: SeqNumberType,
     pub file_versions: HashMap<PathBuf, FileVersion>,
 }
--- a/lib/segment/src/data_types/segment_manifest.rs
+++ b/lib/segment/src/data_types/segment_manifest.rs
@@ -3,7 +3,42 @@ use std::path::PathBuf;
 
 use crate::types::SeqNumberType;
 
-#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[derive(Clone, Debug, Default)]
+pub struct SegmentManifests {
+    manifests: HashMap<String, SegmentManifest>,
+}
+
+impl SegmentManifests {
+    pub fn version(&self, segment_id: &str) -> Option<SeqNumberType> {
+        self.manifests
+            .get(segment_id)
+            .map(|manifest| manifest.segment_version)
+    }
+
+    pub fn get(&self, segment_id: &str) -> Option<&SegmentManifest> {
+        self.manifests.get(segment_id)
+    }
+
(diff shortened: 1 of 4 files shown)


lib/segment/src/data_types/segment_manifest.rs
```diff
@@ -3,8 +3,44 @@ use std::path::PathBuf;
...
+    pub fn add(&mut self, new_manifest: SegmentManifest) -> bool {
+        let Some(current_manifest) = self.manifests.get_mut(&new_manifest.segment_id) else {
+            self.manifests
+                .insert(new_manifest.segment_id.clone(), new_manifest);
+
+            return true;
+        };
+
+        debug_assert_eq!(current_manifest.segment_id, new_manifest.segment_id);
+
```
timvisee, line 31: I'd consider this one obsolete

```suggestion
        debug_assert_eq!(current_manifest.segment_id, new_manifest.segment_id);

```

lib/segment/src/data_types/segment_manifest.rs
```diff
@@ -3,8 +3,44 @@ use std::path::PathBuf;
...
+            return true;
+        }
+
+        false
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct SegmentManifest {
+    pub segment_id: String,
```
timvisee, line 43: Can't we remove this, and just rely on map keys? It is not used outside of the `SegmentManifests` struct.

lib/segment/src/segment/partial_snapshot.rs
```diff
@@ -22,32 +22,62 @@ impl PartialSnapshotEntry for Segment {
...
+        manifests.add(self.get_segment_manifest()?);
+        Ok(())
+    }
+}
+
+impl Segment {
+    fn segment_id(&self) -> OperationResult<&str> {
+        self.current_path
+            .file_stem()
+            .and_then(|segment_dir| segment_dir.to_str())
```
timvisee, line 70: Shall we assert here it's a UUID, just in case we're doing some funny stuff we don't expect?

If so we might rename this to `segment_uuid()` to make it a bit more explicit.

timvisee: I'll add a debug insertion then, just in case!

lib/segment/src/data_types/segment_manifest.rs
timvisee, line 31: I consider this to be obsolete

```suggestion
```

lib/segment/src/data_types/segment_manifest.rs
timvisee: I expect you added it here so that we persist it in the manifest JSON. While it's not strictly necessary, I have no objection to keeping it.

I assume that's also why we have the debug assertion below, which would otherwise become obsolete.

timvisee (approved):

Issue #5983, opened 2025-02-13
Malformed snapshot can cause qdrant to panic on restart


generall:
Snapshot recovery procedure doesn't clean collection directory in case if snapshot recovery failure.
As a result, folder with collection config is created, which is later recognized as malformed collection and causes panic on service restart.

```
qdrant-0f57ce39-3e79-4a7b-a31d-2ecdd3c9474c-0 2025-02-12T22:20:42.084449Z ERROR qdrant::startup: Panic occurred in file /qdrant/lib/collection/src/shards/shard_holder/mod.rs at line 607: called `Result::unwrap()` on an `Err` value: ServiceError { error: "No shard found: 0 at ./storage/collections/radio_row_temp", backtrace: Some("   0: std::backtrace::Backtrace::create\n   1: collection::operations::types::CollectionError::service_error\n   2: collection::shards::shard_holder::ShardHolder::load_shards::{{closure}}\n   3: collection::collection::Collection::load::{{closure}}\n   4: storage::content_manager::toc::TableOfContent::new\n   5: qdrant::main\n   6: std::sys::backtrace::__rust_begin_short_backtrace\n   7: main\n   8: <unknown>\n   9: __libc_start_main\n  10: _start\n") }    
qdrant-0f57ce39-3e79-4a7b-a31d-2ecdd3c9474c-0     
qdrant-0f57ce39-3e79-4a7b-a31d-2ecdd3c9474c-0   16: _start
qdrant-0f57ce39-3e79-4a7b-a31d-2ecdd3c9474c-0   15: __libc_start_main
qdrant-0f57ce39-3e79-4a7b-a31d-2ecdd3c9474c-0   14: <unknown>
qdrant-0f57ce39-3e79-4a7b-a31d-2ecdd3c9474c-0   13: main
qdrant-0f57ce39-3e79-4a7b-a31d-2ecdd3c9474c-0   12: std::sys::backtrace::__rust_begin_short_backtrace
qdrant-0f57ce39-3e79-4a7b-a31d-2ecdd3c9474c-0   11: qdrant::main
qdrant-0f57ce39-3e79-4a7b-a31d-2ecdd3c9474c-0   10: storage::content_manager::toc::TableOfContent::new
qdrant-0f57ce39-3e79-4a7b-a31d-2ecdd3c9474c-0    9: collection::collection::Collection::load::{{closure}}
[...]

PR #5984, opened 2025-02-13
Delete collection dir on failed snapshot restore

agourlay:
Tackle https://github.com/qdrant/qdrant/issues/5983

In case of failed snapshot restore via the API, a malformed collection can be left behind preventing the service to properly restart.

This happens because in case of a restore of a non existing collection, we first create a collection and then restore shards into it.

We usually assume that all collection directory with configuration files are well formed but it is not the case anymore here.

Failing to restore a shard leaves behind a malformed collection.

The fix is to delete the whole collection directory IF it was absent before the recovery process.

### DIFF
--- a/lib/storage/src/content_manager/snapshots/recover.rs
+++ b/lib/storage/src/content_manager/snapshots/recover.rs
@@ -140,7 +140,9 @@ async fn _do_recover_from_snapshot(
     let snapshot_config = CollectionConfigInternal::load(tmp_collection_dir.path())?;
     snapshot_config.validate_and_warn();
 
-    let collection = match toc.get_collection(&collection_pass).await.ok() {
+    let collection = toc.get_collection(&collection_pass).await.ok();
+    let collection_did_not_exist = collection.is_none();
+    let collection = match collection {
         Some(collection) => collection,
         None => {
             log::debug!("Collection {collection_pass} does not exist, creating it");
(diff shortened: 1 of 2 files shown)


timvisee (approved):
Would it be possible to add a test for this?

lib/storage/src/content_manager/snapshots/recover.rs
```diff
@@ -229,11 +231,28 @@ async fn _do_recover_from_snapshot(
...
+                Err(err) => {
+                    log::error!("Failed to recover shard {}: {err}", shard_id);
+                    // delete whole collection dir if it was created to not leave a valid config behind
+                    if collection_did_not_exist {
+                        let collection_path = toc.get_collection_path(&collection_pass.to_string());
+                        log::debug!(
+                            "Deleting collection directory for failed snapshot recovery: {:?}",
+                            collection_path
+                        );
+                        tokio::fs::remove_dir_all(&collection_path).await?;
```
generall, line 252: I don't think we can do this in distributed environment. Consider cluster of 3 machines: one machine restores snapshot, 2 other create empty collection. If snapshot recovery fails, we only remove collection from one machine, 2 other machines will still think it exists.

generall (changes requested):
There 2 problems:

1. Proposed fix doesn't resolve problem if service is killed during snapshot recovery
2. In will create inconsistency in distributed environment.

As an alternative I would propose some options:
- make `restore_local_replica_from` cancel-safe.
- tolerate non-existing shards on load and create empty ones instead
- introduce a special type of file which would indicate that shard is not ready for load and should be ignored on load

PR #5985, opened 2025-02-13
`OnFinalCount` iterator adapter

coszio:
This PR adds an iterator adapter to make hardware usage counting easier when we are measuring iterators.

This adapter plugs in easily to an iterator. When the iterator gets dropped, it will execute the callback to do something with the final number of iterations the iterator was used for.

For example:
```rust
index.value_to_points.get_iter(value).on_final_count(|nexts| {
    hw_counter
        .payload_index_io_read_counter()
        .incr_delta(nexts * size_of::<PointOffsetType>)
});
```

I am not sure about the name, though. Feel free to propose another one :)

### DIFF
--- a/lib/common/common/src/iterator_ext/mod.rs
+++ b/lib/common/common/src/iterator_ext/mod.rs
@@ -2,8 +2,10 @@
 use std::fmt::Debug;
 
 use check_stopped::CheckStopped;
+use on_final_count::OnFinalCount;
 
 mod check_stopped;
+mod on_final_count;
 
 pub trait IteratorExt: Iterator {
     /// Periodically check if the iteration should be stopped.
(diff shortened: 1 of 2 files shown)


JojiiOfficial (approved):
This is a nice addition. Thanks!

lib/common/common/src/iterator_ext/mod.rs
```diff
@@ -26,6 +28,20 @@ pub trait IteratorExt: Iterator {
...
         self.check_stop_every(500, f)
     }
+
+    /// Will execute the callback when the iterator is dropped.
+    ///
+    /// The callback receives the total number of times `.next()` was called on the iterator,
+    /// including the final one where it usually returns `None`.
+    ///
+    /// Consider subtracting 1 if the final `None` is not needed.
+    fn on_final_count<F>(self, f: F) -> OnFinalCount<Self, F>
```
timvisee, line 38: A bit verbose, but how about `count_callback_on_drop`? It might give a better hint on what it does without looking at the function.

I've no strong opinion on either.

cc @JojiiOfficial

timvisee (approved):

lib/common/common/src/iterator_ext/mod.rs
JojiiOfficial: I'm fine with both but agree this being more self-explanatory

PR #5991, opened 2025-02-14
Do not deactivate partial/recovery replica on missing point

timvisee:
A replica that is receiving a shard transfer, might not have all points yet. A replica could have missed a bunch of insertions, which might be the reason why the shard transfer is ongoing - to recover the replica.

Our replica deactivation logic was a bit to eager, immediately marking a replica as dead if a point is not found.

Marking as dead cascades into other problems. If we mark the target shard of a transfer as dead, the transfer is immediately aborted. If we have a continuous stream of operations modifying points that aren't there yet, the transfer will never progress because it constantly gets aborted. It is exactly the scenario we've been seeing in a real cluster.

This PR relaxes the condition. It accepts a missing point error while the replica is undergoing recovery. We should accept it because it might not have all points yet. It prevents us from marking the replica as dead too eagerly, cascading into other problems. This change is safe because the ongoing shard transfer will take care of ensuring data consistency. If our recovering shard might not apply an incoming update now, it will apply it (again) later as part of the transfer.

I've added a test that asserts the new behavior. The test fails when running it without the changes in this PR. That shows the new behavior works as expected.

### Tasks

- [x] Decide whether this change is correct
- [x] Add test

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -424,6 +424,15 @@ impl ShardReplicaSet {
                         .collect()
                 } else {
                     failures
+                        .into_iter()
+                        .filter(|(peer_id, err)| {
+                            let not_partial_recovery = self
+                                .peer_state(*peer_id)
+                                .is_none_or(|state| !state.is_partial_or_recovery());
+                            let is_transient = err.is_transient();
+                            not_partial_recovery || is_transient
+                        })
+                        .collect()
                 };
 
                 let wait_for_deactivation = self.handle_failed_replicas(
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -424,15 +424,6 @@ impl ShardReplicaSet {
                         .collect()
                 } else {
                     failures
-                        .into_iter()
-                        .filter(|(peer_id, err)| {
-                            let not_partial_recovery = self
-                                .peer_state(*peer_id)
-                                .is_none_or(|state| !state.is_partial_or_recovery());
-                            let is_transient = err.is_transient();
-                            not_partial_recovery || is_transient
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/replica_set/update.rs
```diff
@@ -580,6 +580,16 @@ impl ShardReplicaSet {
...
 
+            // Ignore missing point errors if replica is in partial or recovery state
+            // Partial or recovery state indicates that the replica is receiving a shard transfer,
+            // it might not have received all the points yet
+            let partial_or_recovery = self
+                .peer_state(*peer_id)
+                .is_some_and(|state| state.is_partial_or_recovery());
+            if partial_or_recovery && err.is_missing_point() {
+                continue;
+            }
```
ffuugoo, line 591: We actually have `peer_state` in scope already 😂

```suggestion                
            if peer_state.is_partial_or_recovery() && err.is_missing_point() {
                continue;
            }
```

timvisee: :clown_face:

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -580,6 +580,14 @@ impl ShardReplicaSet {
                 continue;
             }
 
+            // Ignore missing point errors if replica is in partial or recovery state
+            // Partial or recovery state indicates that the replica is receiving a shard transfer,
+            // it might not have received all the points yet
+            // See: <https://github.com/qdrant/qdrant/pull/5991>
+            if peer_state.is_partial_or_recovery() && err.is_missing_point() {
+                continue;
+            }
```
timvisee, line 589: The change is simple: do not mark a replica as dead on a missing point error if it's in partial/recovery state.

generall (approved):

PR #5992, opened 2025-02-14
Sparse vectors log

KShivendu:
Sparse vectors was causing OOM on chaos testing and it was not clear what's happening because of not having relevant logs. 

I propose to log it on a debug level similar to how we log `building HNSW for 11806 vectors with 1 CPUs`

### DIFF
--- a/lib/sparse/src/index/inverted_index/inverted_index_ram_builder.rs
+++ b/lib/sparse/src/index/inverted_index/inverted_index_ram_builder.rs
@@ -47,6 +47,14 @@ impl InvertedIndexBuilder {
 
     /// Consumes the builder and returns an InvertedIndexRam
     pub fn build(self) -> InvertedIndexRam {
+        if self.posting_builders.is_empty() {
+            return InvertedIndexRam {
+                postings: vec![],
+                total_sparse_size: self.total_sparse_size,
+                vector_count: self.vector_count,
+            };
+        }
+
         let num_posting_lists = self.posting_builders.len();
         debug!(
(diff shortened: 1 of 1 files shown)


KShivendu:
> count of sparse vectors is zero

Well it's printing:
```
2025-02-14T14:09:29.166564Z DEBUG sparse::index::inverted_index::inverted_index_ram_builder: building inverted index with 0 posting lists
```

I can avoid it by simply adding a check for `num_posting_lists > 0`

lib/sparse/src/index/inverted_index/inverted_index_ram_builder.rs
```diff
@@ -46,11 +47,33 @@ impl InvertedIndexBuilder {
...
+            "building inverted index with {} posting lists",
+            num_posting_lists
+        );
+        let mut new_vectors_count = 0;
+
         let mut postings = Vec::with_capacity(self.posting_builders.len());
         for posting_builder in self.posting_builders {
-            postings.push(posting_builder.build());
+            let posting_list = posting_builder.build();
+            new_vectors_count += posting_list.elements.len();
```
agourlay, line 68: the sum of posting length is not equal to the vector count

points have their dimension spread over several posting lists

KShivendu: ohh right! makes sense. Each posting list represents a dimension and we already have `self.vector_count` which was incremented by `add()`

agourlay (approved):

PR #6010, opened 2025-02-17
Ca cert for snapshots

generall:
I have a single node hybrid cloud cluster. I configured TLS directly in the DB (no P2P TLS) which sets this:

```
service:
  # Enable HTTPS for the REST and gRPC API
  enable_tls: true

# TLS configuration.
# Required if either service.enable_tls or cluster.p2p.enable_tls is true.
tls:
  # Server certificate chain file
  cert: ./tls/cert.pem

  # Server private key file
  key: ./tls/key.pem
```  

This works correctly when connecting to Qdrant.
But when I try to load a snapshots from a remote HTTP location:
```
PUT /collections/midjourney/snapshots/recover
{
  "location": "http://snapshots.qdrant.io/midlib.snapshot"
}
```
I get the following error returned from Qdrant
```
{
  "error": "Service internal error: failed to initialize HTTP(S) client: failed to read HTTPS client CA certificate file ./tls/cacert.pem: No such file or directory (os error 2)"
}
```
Why is a ./tls/cacert.pem mandatory for this operation?

---

Resolution:

Indeed, `./tls/cacert.pem` is not required. Not our config file can accept `null` as `ca_cert`. 
To covert for the case where default configuration is used, the download will no longer error out if certificate is not found.

### DIFF
--- a/config/development.yaml
+++ b/config/development.yaml
@@ -28,11 +28,3 @@ storage:
     default_segment_number: 2
 
   handle_collection_load_errors: true
-
-
-tls:
-  # Certificate chain file
-  cert: ./tests/tls/cert/cert.pem
-
-  # Private key file
-  key: ./tests/tls/cert/key.pem
--- a/config/development.yaml
+++ b/config/development.yaml
@@ -28,3 +28,11 @@ storage:
     default_segment_number: 2
 
   handle_collection_load_errors: true
+
+
+tls:
+  # Certificate chain file
+  cert: ./tests/tls/cert/cert.pem
+
+  # Private key file
+  key: ./tests/tls/cert/key.pem
--- a/src/common/http_client.rs
+++ b/src/common/http_client.rs
@@ -65,11 +65,9 @@ fn https_client(
 
     // Configure TLS root certificate and validation
     if let Some(tls_config) = tls_config {
-        if let Some(ca_cert) = tls_config.ca_cert.clone() {
-            match https_client_ca_cert(ca_cert.as_ref()) {
-                Ok(ca_cert) => {
-                    builder = builder.add_root_certificate(ca_cert);
-                }
+        if let Some(ca_cert) = &tls_config.ca_cert {
+            match https_client_ca_cert(ca_cert) {
+                Ok(ca_cert) => builder = builder.add_root_certificate(ca_cert),
                 Err(err) => {
                     // I think it might be Ok to not fail here, if root certificate is not found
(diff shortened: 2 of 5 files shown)


config/development.yaml
```diff
@@ -28,3 +28,11 @@ storage:
...
 
   handle_collection_load_errors: true
+
+
+tls:
+  # Certificate chain file
+  cert: ./tests/tls/cert/cert.pem
+
+  # Private key file
+  key: ./tests/tls/cert/key.pem
```
timvisee, line 38: Do we just put this here to ensure we can parse `TlsConfig` without `ca_cert` defined?

src/common/http_client.rs
```diff
@@ -65,7 +65,27 @@ fn https_client(
...
+                    // There are 2 possible scenarios:
+                    //
+                    // 1. Server TLS is either not used or it uses some other CA certificate (like global one)
+                    // 2. Server TLS is using self-signed certificate, and we should have it.
+                    //
+                    // In first case, we don't need to load the CA certificate, everything will work in either case.
+                    // In second case, we should have the CA certificate, request will fail because of invalid certificate.
+                    //
+                    // So both scenarios work exactly the same way if we fail early or not.
+                    // Warning message is needed for easier debugging in case of second scenario.
```
timvisee, line 82: @generall I do prefer to fail here, just to make things a bit more explicit, especially since this is security critical.

Either a user configures the CA, or the user does not.

If the user does configure the CA, we fail if:
- we fail to read the file
- we fail to parse the CA

I'd consider both cases to be valid failures.

If it is by accident the user can adjust the configuration and remove the CA.

Wdyt?

timvisee (commented):
A question about error handling before we merge:

config/development.yaml
generall: It was a debug change, forgot to roll back

src/common/http_client.rs
generall: In ideal situation I would agree, but out existing configuration, unfrtunatelly, specified path by default. So if user wants to disable it, they should explicitly overwrite default value. It might be a problem for cloud and existing users

generall: What is the reason to fail, if we could verify the connection by other existing certificates?

timvisee: > So if user wants to disable it, they should explicitly overwrite default value. 

That makes a lot of sense. Let's keep it.

timvisee (approved):

PR #6013, opened 2025-02-18
Add codecov threshold

KShivendu:
Avoid marking branch as failures if coverage is above 50%. Current is `55.26%` (only with unit tests)

Have triggered a run to see results: https://github.com/qdrant/qdrant/actions/runs/13386901276/job/37385582869

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 50
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -3,3 +3,4 @@ coverage:
     project:
       default:
         target: 50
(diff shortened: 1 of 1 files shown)


.github/codecov.yml
```diff
@@ -0,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 50
```
KShivendu, line 5: If it drops below 50% we mark as CI failure.

timvisee (approved):
Thanks!

If it ever becomes a problem again, we can adjust the value.

KShivendu:
Triggered https://github.com/qdrant/qdrant/actions/runs/13387084122. It should pass and generated report shouldn't include patch diff.

PR #6015, opened 2025-02-18
IO resource usage permit

generall:
Optimization is running in multiple stages. One stage is IO-heavy, another is CPU-heavy.
We want to utilize all types of resources at the same time, so we introduce a new type of resource budget, that allows other parallel processes to pick up IO-heavy phase without awaiting for CPU-heavy one


timvisee (commented):
Reviewed in call.

We found two points of improvement:
1. Use the same count in IO budget, and acquire the same amount of IO vs CPU. That keeps both in sync.
2. `replace_with` can deadlock when re-acquiring the same resource

lib/collection/src/update_handler.rs
```diff
@@ -618,14 +621,16 @@ impl UpdateHandler {
             // Otherwise skip now and start a task to trigger the optimizer again once CPU
             // budget becomes available
             let desired_cpus = num_rayon_threads(max_indexing_threads);
-            if !optimizer_cpu_budget.has_budget(desired_cpus) {
+            let desired_io = 0; // TODO: We need at least one IO thread for each optimization
```
timvisee, line 624: Does this TODO need updating?

lib/common/common/src/budget.rs
```diff
@@ -0,0 +1,359 @@
...
+                return Some(permit);
+            } else {
+                std::thread::sleep(delay);
+                delay = (delay * 2).min(Duration::from_secs(2));
+            }
+        }
+        None
+    }
+
+    pub fn replace_with(
```
timvisee, line 154: This can still 'deadlock' if our requested number of CPUs or IO is higher than the total budget/capacity.

Maybe an assertion would suffice here, as we don't suspect to get into that scenario in practice?

lib/storage/src/types.rs
```diff
@@ -39,6 +39,11 @@ pub struct PerformanceConfig {
     /// If positive - use this absolute number of CPUs.
     #[serde(default)]
     pub optimizer_cpu_budget: isize,
+    /// IO budget, how many parallel IO operations to allow for an optimization job.
```
timvisee, line 42: This might need updating since we don't take just one permit/token for each job anymore.

timvisee (approved):

lib/collection/src/update_handler.rs
generall: this is a good catch

timvisee (approved):
:raised_hands:

agourlay:
Are there some benchmark results to share to validate this patch?

generall:
Not yet. I wanted to make them before merging

timvisee:
Sorry! :clown_face:

generall:
Improvement is visible on large-scale indexation. In my experiment I was indexing 400M 512d vectors, and current release take about 40 hours on 8 CPU machine, while dev with this change finishes in 28 hours.

Also attaching CPU/Disk usage proviles, which clearly indicate an improvement:

NEW
Screenshot from 2025-02-22 12-54-02

OLD
Screenshot from 2025-02-22 12-52-43

PR #6020, opened 2025-02-19
Handle malformed shard after incomplete restore

agourlay:
Alternative to https://github.com/qdrant/qdrant/pull/5984

This PR makes sure that a failed snapshot restore does not prevent the service from restarting.

This is achieved by making the restore process crash safe using a flag file.
The flag is created before the shard restore and deleted afterwards.

The presence of a flag at startup means that either:
- the restore process crashed midway
- the restore process failed and left the flag behind

When starting up, the presence of the flag is used to load an empty shard marked as dead.
This empty shard can be the target of a new snapshot restore.

There is an E2E integration test that generates a corrupted snapshot to prove the existence of the flag and the correct startup of the service.

## Status

At this point, it is possible to restart the service & try another non-corrupted snapshot to fix the shard.

However we want to be able to resync the node "automatically" which requires a shard transfer.
The absence of a proper replica state makes it difficult to achieve without hacks.

### DIFF
--- a/lib/collection/src/shards/mod.rs
+++ b/lib/collection/src/shards/mod.rs
@@ -41,6 +41,11 @@ pub fn shard_path(collection_path: &Path, shard_id: ShardId) -> PathBuf {
     collection_path.join(format!("{shard_id}"))
 }
 
+/// Path to a shard directory
+pub fn shard_initialized_flag_path(collection_path: &Path, shard_id: ShardId) -> PathBuf {
+    collection_path.join(format!("shard_{shard_id}.initialized"))
+}
+
 /// Verify that a shard exists by loading its configuration.
 /// Returns the path to the shard if it exists.
 pub async fn check_shard_path(
--- a/lib/collection/src/shards/shard_holder/mod.rs
+++ b/lib/collection/src/shards/shard_holder/mod.rs
@@ -601,8 +601,6 @@ impl ShardHolder {
         };
 
         for shard_id in shard_ids_list {
-            let shard_key = self.get_shard_id_to_key_mapping().get(&shard_id);
-
             // Check if shard is fully initialized on disk
             // The initialization flag should be absent for a well-formed replica set
             let initialized_flag = shard_initialized_flag_path(collection_path, shard_id);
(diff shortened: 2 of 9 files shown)


lib/collection/src/shards/mod.rs
```diff
@@ -41,6 +41,11 @@ pub fn shard_path(collection_path: &Path, shard_id: ShardId) -> PathBuf {
     collection_path.join(format!("{shard_id}"))
 }
 
+/// Path to a shard directory
+pub fn shard_initialized_flag_path(collection_path: &Path, shard_id: ShardId) -> PathBuf {
+    collection_path.join(format!("shard_{shard_id}.initialized"))
```
coszio, line 46:
```suggestion
    collection_path.join(format!("shard_{shard_id}.initializing"))
```

To indicate that the file is temporary

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -601,52 +601,94 @@ impl ShardHolder {
...
+                .await
+                .unwrap();
+                self.add_shard(shard_id, empty_replica_set, shard_key.cloned())
+                    .unwrap();
+                // Delete the initialized flag
+                tokio::fs::remove_file(&initialized_flag)
+                    .await
+                    .unwrap_or_else(|e| {
+                        log::error!("Failed to remove initialized flag for shard {collection_id}:{shard_id}: {e}");
+                    });
```
coszio, line 643: Should we delete the directory as well? Otherwise I suspect the service can be restarted again and load a corrupt shard

timvisee: I don't think we can just remove the directory. The replica (local shard) must exist, because consensus tells us it should.

If we just remove the directory, we must remove the replica from consensus (and confirm that, which is not trivial).

You make a good point though. Just removing the initialized file here may be incorrect, because another crash will not get it into the same state on restart.

I think we must do either of:
- load dummy shard, mark shard as dead (if not done already), confirm with consensus, only then remove initialized file
- remove replica entirely, confirm with consensus, remove directory

I strongly prefer the first, because I see two more problems with the second approach:
- we might accidentally remove the last  replica of some shard
- we lose the replica set state, which we should always have, even if not having a local replica

agourlay:
superseded by https://github.com/qdrant/qdrant/pull/6038

PR #6021, opened 2025-02-19
Gpu add memory barriers

IvanPleshkov:
This PR adds a vulkan memory barrier. Memory barrier is a sync primitive that adds guarantees about execution order. After this barrier, all changes must be available for downloading without race conditions or unflushed caches.

### DIFF
--- a/lib/gpu/src/context.rs
+++ b/lib/gpu/src/context.rs
@@ -112,6 +112,13 @@ impl Context {
         Ok(())
     }
 
+    /// Barrier for buffers.
+    /// It's used to synchronize access to buffers between different shaders.
+    /// By Vulkan specification, resources between different shaders/transfers must be synchronized.
+    /// Example of compute shader sync:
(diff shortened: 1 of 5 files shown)


lib/gpu/src/context.rs
```diff
@@ -112,6 +112,45 @@ impl Context {
...
 
+    pub fn barrier_buffers(&mut self, buffers: &[Arc<Buffer>]) -> GpuResult<()> {
+        if self.vk_command_buffer == vk::CommandBuffer::null() {
+            self.init_command_buffer()?;
+        }
+
+        let buffer_memory_barriers = buffers
+            .iter()
+            .map(|buffer| {
+                self.resources.push(buffer.clone() as Arc<dyn Resource>);
```
xzfc, line 123: It seems that each call to `barrier_buffers` would extend the `self.resources` vec.

xzfc (approved):

PR #6022, opened 2025-02-19
Optimize immutable id tracker

generall:


lib/segment/src/id_tracker/compressed/versions_store.rs
```diff
@@ -0,0 +1,74 @@
...
+///
+/// Implements a required subset of the Vec API:
+///
+/// * get by index
+/// * set by index
+/// * [] operator
+/// * len
+/// * push
+#[derive(Debug)]
+pub struct CompressedVersionsStore {
```
coszio, line 16: `Store` sounds like it is persisted, maybe just omit it? 
```suggestion
pub struct CompressedVersions {
```

coszio (approved):

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -369,13 +370,13 @@ impl IdTracker for ImmutableIdTracker {
         version: SeqNumberType,
     ) -> OperationResult<()> {
         if self.external_id(internal_id).is_some() {
-            let old_version = self.internal_to_version.get_mut(internal_id as usize);
+            let old_version = self.internal_to_version.get(internal_id as usize);
```
timvisee, line 373: Implement `has` here, which would be more efficient

lib/segment/src/id_tracker/compressed/versions_store.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    pub fn get(&self, index: usize) -> Option<SeqNumberType> {
+        self.lower_bytes.get(index).map(|&lower| {
+            let upper = *self.upper_bytes.get(&(index as u32)).unwrap_or(&0);
+            Self::version_from_parts(lower, upper)
+        })
+    }
+
+    pub fn set(&mut self, index: usize, value: SeqNumberType) {
```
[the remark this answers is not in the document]
timvisee: Handled in <https://github.com/qdrant/qdrant/pull/6036>

Issue #6025, opened 2025-02-19
qdrant::startup: Panic occurred in file / Failed to load local shard


itinance:
After running qdrant since half a year on the same hard drive, it suddenly stopped to boot properly, complaining about a missing directory.
Directly after booting the qdrant container, it stops early with a panic error, that a directory for a collection does not exist, although it does.

```
qdrant_core_2  |   __ _  __| |_ __ __ _ _ __ | |_  
qdrant_core_2  |  / _` |/ _` | '__/ _` | '_ \| __| 
qdrant_core_2  | | (_| | (_| | | | (_| | | | | |_  
qdrant_core_2  |  \__, |\__,_|_|  \__,_|_| |_|\__| 
qdrant_core_2  |     |_|                           
qdrant_core_2  | 
qdrant_core_2  | Version: 1.12.5, build: 27260abd
qdrant_core_2  | Access web UI at http://localhost:6333/dashboard
qdrant_core_2  | 
qdrant_core_2  | 2025-02-19T20:11:50.238705Z  INFO storage::content_manager::consensus::persistent: Loading raft state from ./storage/raft_state.json    
qdrant_core_2  | 2025-02-19T20:11:50.239813Z  INFO storage::content_manager::toc: Loading collection: test    
Recovering collection test [00:00:00] █████████████████████████████ 0/0 (eta:0s)2025-02-19T20:11:50.293646Z  INFO storage::content_manager::toc: Loading collection: checkpoints    
Recovering collection checkpoints [00:00:00] ██████████████████████ 1/1 (eta:0s)2025-02-19T20:11:50.359342Z  INFO storage::content_manager::toc: Loading collection: embeddings_title_ada    
qdrant_core_2  | 2025-02-19T20:11:50.374996Z ERROR qdrant::startup: Panic backtrace: 
qdrant_core_2  |    0: std::backtrace::Backtrace::create
qdrant_core_2  |    1: qdrant::startup::setup_panic_hook::{{closure}}
qdrant_core_2  |    2: std::panicking::rust_panic_with_hook
qdrant_core_2  |    3: std::panicking::begin_panic_handler::{{closure}}
qdrant_core_2  |    4: std::sys::backtrace::__rust_end_short_backtrace
qdrant_core_2  |    5: rust_begin_unwind
qdrant_core_2  |    6: core::panicking::panic_fmt
qdrant_core_2  |    7: collection::shards::shard_holder::ShardHolder::load_shards::{{closure}}
[...]

PR #6032, opened 2025-02-20
Improve segment loading, don't load invalid segments

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/6025>

When loading a local shard, only load segments from valid directory paths.

Before, we'd try to load every entry in the segments directory as if it was a segment. This caused a panic during segment loading if macOS places a `.DS_Store` file in there.

More specifically, we now ignore:
- non-directories
- hidden directories prefixed with a `.`
- `.DS_Store`

We now also walk over all segment path entries before trying to load them. That is significant because we now catch any IO errors before spawning threads. Before we would leak threads if crashing in the middle.

Log example when trying to load such invalid segment:

```
2025-02-20T18:17:10.241651Z DEBUG collection::shards::local_shard: Segments path entry prefixed with a period, ignoring: ./storage/collections/embeddings_title_ada/0/segments/.DS_Store
```

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -275,7 +275,32 @@ impl LocalShard {
         // let semaphore = Arc::new(parking_lot::Mutex::new(()));
 
         for entry in segment_dirs {
-            let segments_path = entry.unwrap().path();
+            let entry = entry.map_err(|err| {
+                CollectionError::service_error(format!(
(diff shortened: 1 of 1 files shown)


agourlay (approved):

itinance:
When will this be released?

timvisee:
> When will this be released?

We didn't settle on a day yet. It may be sometime in the next two weeks.

PR #6034, opened 2025-02-20
Fix flaky `test_keyword_facet` test

coszio:
This flaky test didn't get enough love for a looong time, sorry. 

There was a bug hidden in the map mmap index, which didn't consider the deleted bitslice when iterating over the point offsets of each value.

This fixes it and improves the tests.

### DIFF
--- a/lib/segment/src/id_tracker/compressed/compressed_point_mappings.rs
+++ b/lib/segment/src/id_tracker/compressed/compressed_point_mappings.rs
@@ -30,9 +30,10 @@ pub type FileEndianess = LittleEndian;
 
 #[derive(Clone, PartialEq, Default, Debug)]
 pub struct CompressedPointMappings {
-    // `deleted` specifies which points of internal_to_external was deleted.
-    // It is possible that `deleted` can be longer or shorter than `internal_to_external`.
-    // - if `deleted` is longer, then extra bits should be set to `false` and ignored.
+    /// `deleted` specifies which points of internal_to_external was deleted.
+    /// It is possible that `deleted` can be longer or shorter than `internal_to_external`.
+    /// - if `deleted` is longer, then extra bits should be set to `false` and ignored.
+    /// - if `deleted` is shorter, then extra indices are as if the bits were set to `true`.
     deleted: BitVec,
     internal_to_external: CompressedInternalToExternal,
 
--- a/lib/segment/src/id_tracker/compressed/compressed_point_mappings.rs
+++ b/lib/segment/src/id_tracker/compressed/compressed_point_mappings.rs
@@ -31,9 +31,7 @@ pub type FileEndianess = LittleEndian;
 #[derive(Clone, PartialEq, Default, Debug)]
 pub struct CompressedPointMappings {
     /// `deleted` specifies which points of internal_to_external was deleted.
-    /// It is possible that `deleted` can be longer or shorter than `internal_to_external`.
-    /// - if `deleted` is longer, then extra bits should be set to `false` and ignored.
-    /// - if `deleted` is shorter, then extra indices are as if the bits were set to `true`.
+    /// Its size is exactly the same as `internal_to_external`.
     deleted: BitVec,
     internal_to_external: CompressedInternalToExternal,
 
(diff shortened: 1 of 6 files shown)


lib/segment/src/index/field_index/map_index/mmap_map_index.rs
```diff
@@ -247,8 +247,15 @@ impl<N: MapIndexKey + Key + ?Sized> MmapMapIndex<N> {
...
-        self.value_to_points
-            .iter()
-            .map(|(k, v)| (k, Box::new(v.iter().copied()) as IdIter))
+        self.value_to_points.iter().map(|(k, v)| {
+            (
+                k,
+                Box::new(
+                    v.iter()
+                        .copied()
+                        .filter(|idx| !self.deleted.get(*idx as usize).unwrap_or(true)),
```
coszio, line 256: This is the line which was missing

lib/segment/src/id_tracker/compressed/compressed_point_mappings.rs
```diff
@@ -30,9 +30,10 @@ pub type FileEndianess = LittleEndian;
 
 #[derive(Clone, PartialEq, Default, Debug)]
 pub struct CompressedPointMappings {
-    // `deleted` specifies which points of internal_to_external was deleted.
-    // It is possible that `deleted` can be longer or shorter than `internal_to_external`.
-    // - if `deleted` is longer, then extra bits should be set to `false` and ignored.
+    /// `deleted` specifies which points of internal_to_external was deleted.
+    /// It is possible that `deleted` can be longer or shorter than `internal_to_external`.
+    /// - if `deleted` is longer, then extra bits should be set to `false` and ignored.
+    /// - if `deleted` is shorter, then extra indices are as if the bits were set to `true`.
```
timvisee, line 36: This is incorrect. In this structure the deleted flags are exactly sized. See [this](https://github.com/qdrant/qdrant/blob/c8898a19a4d63480543d5f05bbe33b4ddd35f716/lib/segment/src/id_tracker/compressed/compressed_point_mappings.rs#L50-L52) a bit further down.

I understand the comment was already there, but I only notice it now so let's change it here.

```suggestion
    /// `deleted` specifies which points of internal_to_external was deleted.
    /// Its size is exactly the same as `internal_to_external`.
```

lib/segment/src/id_tracker/point_mappings.rs
```diff
@@ -25,9 +25,10 @@ pub type FileEndianess = LittleEndian;
 
 #[derive(Clone, PartialEq, Default, Debug)]
 pub struct PointMappings {
-    // `deleted` specifies which points of internal_to_external was deleted.
-    // It is possible that `deleted` can be longer or shorter than `internal_to_external`.
-    // - if `deleted` is longer, then extra bits should be set to `false` and ignored.
+    /// `deleted` specifies which points of internal_to_external was deleted.
+    /// It is possible that `deleted` can be longer or shorter than `internal_to_external`.
+    /// - if `deleted` is longer, then extra bits should be set to `false` and ignored.
+    /// - if `deleted` is shorter, then extra indices are as if the bits were set to `true`.
```
timvisee, line 31: Here it can indeed be dynamically sized, so we can keep this :+1:

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
timvisee, line 256: Shouldn't it be unwrap or false? If our deleted bitvec is not big enough, we assume the point is not deleted yet.

```suggestion
                        .filter(|idx| !self.deleted.get(*idx as usize).unwrap_or(false)),
```

timvisee (changes requested):

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
coszio: It is not possible that the deleted slice is shorter, since it gets created at the specific length and never resizes. So if we check for a larger idx, it should mean it is not there

timvisee: But, it'll also never be longer, right? And we control what points we query from it.

Maybe 

```rust
!self.deleted.get[*idx as usize]
```

will suffice.

timvisee (approved):

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
coszio: I don't trust direct indexing at this point 😅

timvisee: Okay! 😆

PR #6038, opened 2025-02-21
Keep initial shard configuration on failed restore

agourlay:
Alternative to https://github.com/qdrant/qdrant/pull/6020

This PR makes sure that a failed snapshot restore does not prevent the service from restarting.

This is achieved by performing two things:
- keeping the shard's folder inner configuration in case of failed recovery, instead of deleting it completely
- making the restore process crash safe using a flag file

Not deleting the shard folder pleases the `ShardHolder`  which expects one folder per shard.
Without a WAL found for the shard, it is turned into a `DummyShard` at load time.

```
ERROR collection::shards::replica_set: Failed to load local shard "./storage/collections/test_collection/0", initializing "dummy" shard instead: Service internal error: Wal error: Can't init WAL: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```
 
More importantly, keeping the shard configuration provides the latest valid `replica_state.json` to access the remote shards. 

Concerning the crash safety, a file flag is created before the shard restore and deleted afterwards.

The presence of a flag at startup means that either:
- the restore process crashed midway
- the restore process failed and left the flag behind

When starting up, the presence of the flag is used to delete the potentially corrupted segments data and WAL.
This dummy shard can be the target of a new snapshot restore OR manual shard replication to fix it.

There is an E2E integration test that generates a corrupted snapshot to prove the existence of the flag, the correct startup of the service and the recovery of the shard.

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -108,6 +108,7 @@ pub struct LocalShard {
 
 /// Shard holds information about segments and WAL.
 impl LocalShard {
+    /// Moves `wal`, `segments` and `clocks` data from one path to another.
     pub async fn move_data(from: &Path, to: &Path) -> CollectionResult<()> {
         let wal_from = Self::wal_path(from);
         let wal_to = Self::wal_path(to);
--- a/lib/collection/src/shards/shard_holder/mod.rs
+++ b/lib/collection/src/shards/shard_holder/mod.rs
@@ -39,11 +39,12 @@ use crate::operations::{OperationToShard, SplitByShard};
 use crate::optimizers_builder::OptimizersConfig;
 use crate::save_on_disk::SaveOnDisk;
 use crate::shards::channel_service::ChannelService;
+use crate::shards::local_shard::LocalShard;
 use crate::shards::replica_set::{ReplicaState, ShardReplicaSet};
 use crate::shards::shard::{PeerId, ShardId};
 use crate::shards::shard_config::ShardConfig;
 use crate::shards::transfer::{ShardTransfer, ShardTransferKey};
-use crate::shards::{CollectionId, check_shard_path};
+use crate::shards::{CollectionId, check_shard_path, shard_initializing_flag_path, shard_path};
 
 const SHARD_TRANSFERS_FILE: &str = "shard_transfers";
 const RESHARDING_STATE_FILE: &str = "resharding_state.json";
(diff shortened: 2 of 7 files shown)


lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -601,6 +602,28 @@ impl ShardHolder {
...
+                    shard_id
+                );
+                let shard_path = shard_path(collection_path, shard_id);
+                LocalShard::clear(&shard_path).await.unwrap();
+                // Delete the initialized flag now that the shard is clean
+                tokio::fs::remove_file(&initializing_flag)
+                    .await
+                    .unwrap_or_else(|e| {
+                        log::error!("Failed to remove initializing flag for shard {collection_id}:{shard_id}: {e}");
+                    });
```
timvisee, line 624: I don't think this is correct.

Here we totally remove our local shard, but as far as consensus is concerned, the local shard should still be there. That's a consensus de-sync, and creates problems.

Until we delete or re-recover the replica, we must keep it in dummy state. Even on a restart or crash, we should load the dummy shard again. Removing the initializing file here will not allow that.

I therefore think we should keep this file until the replica itself is properly replaced or deleted. I expect that to already work for restoring the shard through a snapshot as it'll use the same flag file. For other transfer methods, such as stream records, we probably need to explicitly drop the flag when recovery is finished.

Alternatively we can remove the replica from consensus, confirm it, and then remove it locally like we do now. But it's fragile, and I definitely prefer the other option above.

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -108,6 +108,7 @@ pub struct LocalShard {
 
 /// Shard holds information about segments and WAL.
 impl LocalShard {
+    /// Moves `wal`, `segments` and `clocks` data from one path to another.
     pub async fn move_data(from: &Path, to: &Path) -> CollectionResult<()> {
```
timvisee, line 112: In `move_data`, lets move the initializing file marker as well if it exists. Because after moving, we should still consider it to be not initialized.

timvisee (changes requested):

lib/collection/src/shards/local_shard/mod.rs
agourlay: One issue with this approach is that the marker file does not live inside the shard folder but outside.

lib/collection/src/shards/shard_holder/mod.rs
agourlay: > Here we totally remove our local shard

The function `LocalShard::clear` deletes the segment data and WAL but not the shard folder.

This way, the rest of the loading sequence will load the shard as a Dummy if it is not well formed.

timvisee: I wasn't aware that simply deleting our local shard data already ensures that we'll be loading a dummy shard every time. Thanks for elaborating!

lib/collection/src/shards/local_shard/mod.rs
timvisee: Would it be worth to move it inside, to support this?

timvisee (approved):

PR #6039, opened 2025-02-22
gpu: zerocopy instead of transmute

xzfc:
In this PR transmutes in the GPU-related code are replaced with the usage of `zerocopy`. See https://github.com/qdrant/qdrant/pull/5297/files/be387be795977cfe99fabe77694ba31c66226f32#r1836929158.

Note: `half::f16` is not compatible with the latest version of `zerocopy`, so `bytemuck` is used to convert it to `&[u8]`.

Also, since every usage of `download_slice()` was used to write into a newly allocated vector, I've introduced `download_vec()` to make usage a bit more convenient.

### DIFF
--- a/lib/common/memory/src/mmap_ops.rs
+++ b/lib/common/memory/src/mmap_ops.rs
@@ -137,12 +137,6 @@ pub fn transmute_to_u8<T>(v: &T) -> &[u8] {
     unsafe { std::slice::from_raw_parts(ptr::from_ref::<T>(v).cast::<u8>(), mem::size_of_val(v)) }
 }
 
-pub fn transmute_to_u8_mut<T>(v: &mut T) -> &mut [u8] {
-    unsafe {
-        std::slice::from_raw_parts_mut(ptr::from_mut::<T>(v).cast::<u8>(), mem::size_of_val(v))
-    }
-}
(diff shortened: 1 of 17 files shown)


lib/common/memory/src/mmap_ops.rs
```diff
@@ -137,12 +137,6 @@ pub fn transmute_to_u8<T>(v: &T) -> &[u8] {
     unsafe { std::slice::from_raw_parts(ptr::from_ref::<T>(v).cast::<u8>(), mem::size_of_val(v)) }
 }
 
-pub fn transmute_to_u8_mut<T>(v: &mut T) -> &mut [u8] {
```
IvanPleshkov, line 140: Great you found that this and `transmute_to_u8_slice_mut` are used only in gpu 🕵️‍♂️

IvanPleshkov (commented):
Will approve right after this CI:
https://github.com/qdrant/qdrant/actions/runs/13495790236

timvisee (approved):

IvanPleshkov (approved):
GPU CI is green:
https://github.com/qdrant/qdrant/actions/runs/13495790236/job/37703896368

PR #6040, opened 2025-02-22
notify optimizers scheduler of the budget change

generall:
Fixes https://github.com/qdrant/qdrant/pull/6015 for the case, where there are no running updates.

### DIFF
--- a/lib/common/common/src/budget.rs
+++ b/lib/common/common/src/budget.rs
@@ -173,8 +173,10 @@ impl ResourceBudget {
         permit.merge(extra_acquired);
 
         // Release excess resources we now have
-        permit.release_cpu_count(permit.num_cpus.saturating_sub(new_desired_cpus as u32));
-        permit.release_io_count(permit.num_io.saturating_sub(new_desired_io as u32));
+        permit.release(
+            permit.num_cpus.saturating_sub(new_desired_cpus as u32),
+            permit.num_io.saturating_sub(new_desired_io as u32),
+        );
 
         Ok(permit)
     }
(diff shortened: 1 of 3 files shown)


lib/segment/src/segment_constructor/segment_builder.rs
```diff
@@ -552,13 +552,6 @@ impl SegmentBuilder {
             #[cfg(not(feature = "gpu"))]
             let gpu_device = None;
 
-            // If GPU is enabled, release all CPU cores except one.
-            if let Some(_gpu_device) = &gpu_device {
-                if permit.num_cpus > 1 {
-                    permit.release_cpu_count(permit.num_cpus - 1);
-                }
-            }
```
generall, line 560: This behavior was broken and would have created problems with changes of this PR. E.g. we don't want to schedule CPU optimization if GPU is available. More correct implementation should involve dedicated GPU permit.

lib/common/common/src/budget.rs
```diff
@@ -318,21 +330,22 @@ impl ResourcePermit {
...
             num_io: 0,
             io_permit: None,
+            on_release: None,
         }
     }
 
     /// Release CPU permit, giving them back to the semaphore.
-    pub fn release_cpu(&mut self) {
+    fn release_cpu(&mut self) {
         self.cpu_permit.take();
```
timvisee, line 339:
```suggestion
        self.num_cpus = 0;
        self.cpu_permit.take();
```

lib/common/common/src/budget.rs
```diff
@@ -318,21 +330,22 @@ impl ResourcePermit {
...
     }
 
     /// Release IO permit, giving them back to the semaphore.
-    pub fn release_io(&mut self) {
+    fn release_io(&mut self) {
         self.io_permit.take();
```
timvisee, line 344:
```suggestion
        self.num_io = 0;
        self.io_permit.take();
```

timvisee (approved):

generall:
@coderabbitai review

PR #6054, opened 2025-02-24
[score boosting] add more expressions

coszio:
Adds the following operations to expression:
- sqrt
- pow
- exp
- log (base 10)
- ln
- abs

### DIFF
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -587,6 +587,12 @@ message Expression {
         SumExpression sum = 6; // Sum
         DivExpression div = 7; // Divide
         Expression neg = 8; // Negate
+        Expression abs = 9; // Absolute value
+        Expression sqrt = 10; // Square root
+        PowExpression pow = 11; // Power
+        Expression exp = 12; // Exponential
+        Expression log10 = 13; // Logarithm
+        Expression ln = 14; // Natural logarithm
     }
 }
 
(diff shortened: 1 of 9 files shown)


lib/api/src/grpc/proto/points.proto
```diff
@@ -609,6 +615,11 @@ message DivExpression {
     optional float by_zero_default = 3;
 }
 
+message PowExpression {
+    Expression base = 1;
```
timvisee, line 619: Make this optional and default to 2?

coszio: Why, though? I am trying to keep close to [`std`'s powf](https://doc.rust-lang.org/std/primitive.f32.html#method.powf), otherwise it would look more similar to [exp2](https://doc.rust-lang.org/std/primitive.f32.html#method.exp)

lib/segment/src/index/query_optimization/rescore_formula/formula_scorer.rs
```diff
@@ -155,24 +155,46 @@ impl FormulaScorer<'_> {
             }
             ParsedExpression::Sqrt(expr) => {
                 let value = self.eval_expression(expr, point_id)?;
-                Ok(value.sqrt())
+                let sqrt_value = value.sqrt();
+                if sqrt_value.is_nan() {
+                    // Undefined: Send this point to the bottom of results
+                    Ok(f32::NEG_INFINITY)
```
timvisee, line 161: Does this work for euclid as well, because we invert later?

timvisee (approved):

lib/segment/src/index/query_optimization/rescore_formula/formula_scorer.rs
coszio: All score boosting will have `Order::LargerIsBetter`, this will be added to documentation. For the case of euclidean distance, users will have to negate the scores, to be explicit.

PR #6056, opened 2025-02-24
Shard-level limit undersampling

generall:
Apply the same under-sampling as we do for segments, but on the shard level.
Motivation: large-scale deployments with high-limit queries could benefit from this optimization, as data transfer between shards becoming a bottleneck.

Some considerations:

- Unlike segments, shards are guaranteed to be randomly and independently distributed
- Re-requests in shards are too expensive, so we don't make them
- Without this optimization, high limit queries can not efficiently scale horizontally


lib/collection/src/collection/query.rs
```diff
@@ -73,6 +140,22 @@ impl Collection {
...
+        // Auto-sharding happens when we are only querying shards with _the_same_ shard key.
+        // It either might be when we are querying a specific shard key
+        // OR when we are querying all shards with no shard keys specified.
+        let is_auto_sharding = num_unique_shard_keys == 1;
+
+        let batch_request = Self::modify_shard_query_for_undersampling_limits(
+            batch_request,
+            target_shards.len(),
+            is_auto_sharding,
+        );
```
generall, line 157: I was trying to find a better place for this modification, but was not able to, because:

- Higher level functions require to know original limit and offset of the query for proper merge operation, so we can't modify original query higher in the call stack
- Modification of the limit lower in the call stack also problematic, as we need to know "context" which depends on the shard selection. And we should not know about shard selector inside a shard.

generall:
How to test:

create collection with 3 shards (can be local) and 10000 points. Disable hnsw index.

Compare exact search vs non-exact search with limit > 128, the difference should be within 0.999.

lib/collection/src/collection/query.rs
```diff
@@ -60,6 +61,83 @@ impl Collection {
...
+    /// If there are many independent shards, and we need a very high limit, we can do an optimization.
+    /// Instead of querying all shards with the same limit, we can query each shard with a smaller limit
+    /// and then merge the results. Since shards are independent and data is randomly distributed, we can
+    /// apply probability estimation to make sure we query enough points to get the desired number of results.
+    ///
+    /// Same optimization we already apply on segment level, but here it seems to be even more reliable
+    /// because auto-sharding guarantee random and independent distribution of data.
+    ///
+    /// Unlike segments, however, the cost of re-requesting the data is much higher for shards.
+    /// So we "accept" the risk of not getting enough results.
```
timvisee, line 85: If it's so unlikely, can we still do it in case we don't get enough results? 

I imagine a latency spike to be a bit better than getting less results. Getting less results might make the user incorrectly think they have 'reached the end' or 'exhausted all'.

If we basically never expect this to happen, a latency spike might be a 'safer' option.

Wdyt?

lib/collection/src/collection/query.rs
```diff
@@ -60,6 +61,83 @@ impl Collection {
...
+
+        if max_limit < Self::SHARD_QUERY_SUBSAMPLING_LIMIT {
+            return batch_request;
+        }
+
+        let mut new_requests = Vec::with_capacity(batch_request.len());
+
+        for request in batch_request.iter() {
+            let mut new_request = request.clone();
+            let request_limit = new_request.limit + new_request.offset;
```
timvisee, line 115: Should we not go lower than the offset, to not destabilize results before our offset?

I understand that this wasn't 100% accurate or deterministic before, but I wonder if this might make it worse, amplifying inaccuracies, making the results all over the place.

timvisee (commented):
I have two high level questions:

lib/collection/src/collection/query.rs
generall: We can, it is a bit annoying to implement with our batching structure, would be at least twice as many changes

lib/collection/src/collection/query.rs
generall: To compensate the accuracy user might increase hnsw_ef, it will have the same effect as higher limits, but won't cause network/IO saturation as bad as limit

generall:
> Do you maybe have some performance numbers to share based on a large realistic test deployment?

yes, 10 machines cluster with 10k limit takes 2 seconds per request, with 1k limit it takes 0.1 second, everything in RAM (including payload)

lib/collection/src/collection/query.rs
```diff
@@ -60,6 +61,83 @@ impl Collection {
         Ok(results.into_iter().next().unwrap())
     }
 
+    /// If the query limit above this value, it will be a subject to undersampling.
+    const SHARD_QUERY_SUBSAMPLING_LIMIT: usize = 128;
```
coszio, line 65: Why specifically this number? For segments we use `params.hnsw_ef`

generall: `hnsw_ef` was right choise for segment, cause we wanted to minimize HNSW overhead for large limits. This is not relevant for shard-level, where bottleneck is the networking

generall:
@agourlay added a debug message

timvisee (approved):

agourlay (approved):

PR #6057, opened 2025-02-25
[score boosting] Add formula to internal gRPC API

coszio:
Adds the formula query to the internal gRPC API, as well as the necessary conversions.

I decided to convert from `ParsedFormula` to `grpc::Formula` by "unparsing" the structure, rather than adding a parsed equivalent in the internal interface surface to avoid leaking the parsing implementation.

### DIFF
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -11,6 +11,9 @@ use segment::data_types::index::{
     KeywordIndexType, TextIndexType, UuidIndexType,
 };
 use segment::data_types::{facets as segment_facets, vectors as segment_vectors};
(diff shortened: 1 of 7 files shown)


lib/api/src/grpc/conversions.rs
```diff
@@ -2333,3 +2338,100 @@ impl From<HwMeasurementAcc> for HardwareUsage {
...
+            .collect();
+
+        Formula {
+            expression: Some(expression),
+            defaults,
+        }
+    }
+}
+
+fn unparse_expression(
```
timvisee, line 2370: This function name made me laugh :smile:

timvisee (approved):

PR #6061, opened 2025-02-25
Use common actions pattern in workflows

tellet-q:
### All Submissions:


IvanPleshkov (changes requested):
Is this change applicable to GPU CI? There is also `dev-docker-image-build-gpu.yml`, where we build a docker image. I guess it should be changed too. I place a blocker while this question is not resolved

tellet-q:
@IvanPleshkov this PR is to address a new workflow in https://github.com/qdrant/vector-db-benchmark/pull/226, to compare versions using benchmark. I don't think we have any benchmarks for gpu, so I see no reason in comparing gpu images, but I agree that `dev-docker-image-build-gpu.yml` may be updated to using common action.

IvanPleshkov:
> @IvanPleshkov this PR is to address a new workflow in [qdrant/vector-db-benchmark#226](https://github.com/qdrant/vector-db-benchmark/pull/226), to compare versions using benchmark. I don't think we have any benchmarks for gpu, so I see no reason in comparing gpu images, but I agree that `dev-docker-image-build-gpu.yml` may be updated to using common action.

I mean using common code where applicable. Sure I don't propose to include GPU into benches. Remove blocker

IvanPleshkov (approved):

PR #6062, opened 2025-02-25
Struct deconstruction in gRPC conversions

agourlay:
This PR migrates the file `grpc/conversions.rs` to deconstruct the input systematically.

This is desirable to make sure we will not forget new fields in the future like we did in the past :)

By doing so, I found a couple places where some fields were not used.

Future PRs will tackle the remaining grpc and rest conversion files.

### DIFF
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -2408,7 +2408,7 @@ impl TryFrom<SearchPointGroups> for rest::SearchGroupsRequestInternal {
             with_payload,
             with_vector,
             score_threshold,
-        } = search_points.try_into()?;
+        } = rest::SearchRequestInternal::try_from(search_points)?;
 
         Ok(Self {
             vector,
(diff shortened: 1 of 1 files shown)


lib/api/src/grpc/conversions.rs
```diff
@@ -2125,21 +2354,39 @@ impl TryFrom<SearchPointGroups> for rest::SearchGroupsRequestInternal {
...
+            score_threshold,
+            vector_name,
             limit: 0,
             offset: None,
-            collection_name: String::new(),
-            read_consistency: None,
-            timeout: None,
-            shard_key_selector: None,
-            sparse_indices: value.sparse_indices,
+            collection_name,
```
agourlay, line 2385: `collection_name`, `read_consistency` and `shard_key_selector` were not propagated properly.

agourlay:
Fixing flaky test in https://github.com/qdrant/qdrant/pull/6063

lib/api/src/grpc/conversions.rs
timvisee: Good call. Moving the value is practically free, so no problems there either.

timvisee (approved):
Thank you for using the `from` conversions where possible too :pray:

PR #6064, opened 2025-02-25
[score boosting] openapi test for formula

coszio:
Adds a sanity check before proceeding to implement local mode... the double implementation testing


timvisee (approved):

timvisee:
If possible, I might like another test with some illegal or otherwise weird values. Random things that come to mind:

- missing payload value
- negative sqrt (and others)
- euclid distance (with inverse score)
- multiply by very large value (to get float inf)
- divide by zero
- huge neg chain

PR #6065, opened 2025-02-25
remove warning about bootstrap URI if it is not provided

generall:


timvisee (approved):
Fun edge case :smiley:

PR #6072, opened 2025-02-26
Add retry-after header for gRPC

agourlay:
Follow up to https://github.com/qdrant/qdrant/pull/5917 for the gRPC API.

The `retry-after` header is passed using the standard gRPC [metadata structure](https://grpc.io/docs/guides/metadata/).

The value can easily be extracted my clients. (e.g. Rust [client](https://github.com/qdrant/rust-client/pull/212))

This PR makes sure that the `retry-after` header is properly propagated during inter node communication.

There is a test to validate the end to end error propagation when querying an empty node.

### DIFF
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -1131,7 +1131,7 @@ impl CollectionError {
                     retry_after,
                 } = retry_error;
                 let description = format!(
-                    "{rate_limiter_type} rate limit exceeded: Operation requires {cost} tokens but only {tokens_available} were available. Retry after {}s",
+                    "{rate_limiter_type} rate limit exceeded: Operation requires {cost} tokens but only {tokens_available:.1} were available. Retry after {}s",
                     retry_after.as_secs_f32().ceil() as u32,
                 );
                 (description, Some(retry_after))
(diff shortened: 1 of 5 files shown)


lib/collection/src/operations/types.rs
```diff
@@ -1131,7 +1131,7 @@ impl CollectionError {
                     retry_after,
                 } = retry_error;
                 let description = format!(
-                    "{rate_limiter_type} rate limit exceeded: Operation requires {cost} tokens but only {tokens_available} were available. Retry after {}s",
+                    "{rate_limiter_type} rate limit exceeded: Operation requires {cost} tokens but only {tokens_available:.1} were available. Retry after {}s",
```
agourlay, line 1134: no need to display more than one digit precision

lib/collection/src/operations/types.rs
```diff
@@ -1345,10 +1345,27 @@ impl From<tonic::Status> for CollectionError {
...
+                // extract retry-after from metadata
+                // the value is passed as a String containing an integer number of seconds
+                let retry_after = err.metadata().get("retry-after").and_then(|v| {
+                    v.to_str()
+                        .inspect_err(|e| log::info!("Failed to parse retry-after header: {e}"))
+                        .ok()
+                        .and_then(|v| {
+                            v.parse::<u64>()
+                                .inspect_err(|e| {
+                                    log::info!("Failed to parse retry-after header: {e}")
```
timvisee, line 1358: Nit:

```suggestion
                                    log::info!("Failed to parse retry-after value in gRPC metadata (value: {v}): {e}")
```

May point to the problem a bit better. Also, the error does not include the malformed value, so it might be useful to include it explicitly.

timvisee (approved):

PR #6075, opened 2025-02-27
Make sure the shard initialization flag is fsynced

agourlay:
This PR makes sure the initialization flag is synced to disk before proceeding with the snapshot restore.

### DIFF
--- a/lib/collection/src/collection/snapshots.rs
+++ b/lib/collection/src/collection/snapshots.rs
@@ -246,7 +246,8 @@ impl Collection {
         // set shard_id initialization flag
         // the file is removed after full recovery to indicate a well-formed shard
         let shard_flag = shard_initializing_flag_path(&self.path, shard_id);
-        tokio::fs::write(&shard_flag, b"").await?;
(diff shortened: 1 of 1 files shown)


timvisee (approved):
Yes, that makes sense!

Do we potentially need the same when removing the file?

agourlay:
> Do we potentially need the same when removing the file?

I check it looks rather annoying, apparently the parent directory needs to be fsync.

```rust
 // Delete the file
    fs::remove_file(file_path).await?;

    // Open the parent directory to sync it
    if let Some(parent) = file_path.parent() {
        let dir = OpenOptions::new().read(true).open(parent).await?;
        dir.sync_all().await?; // Ensure the deletion is persisted
    }
```

If the file deletion is lost due to a crash, the shard will be considered not well formed and will be cleaned up and loaded as dummy.

I don't really want to sync the whole collection folder.
WDYT?

timvisee:
@agourlay I didn't realize that we'd have to sync the whole shard folder, as otherwise, we might have flushed the flag deletion but other files still aren't written durably.

I agree, let's not bother.

PR #6084, opened 2025-02-28
Improve shard transfer tracker, show fraction in case of resharding

timvisee:
Extension of <https://github.com/qdrant/qdrant/pull/6074>

Improve tracking of records to transfer in shard transfers, specifically for stream records and resharding stream records transfer.

The main two improvements are:
- count records actually being transferred, not batch size
- in case of resharding, estimate number of records to transfer based on resharding fraction

For example, if we're resharding from one to two shards, a resharding transfer will only transfer ~50% of the records, not all of them.

The current resharding implementation still filters points with the hash ring after scrolling, not during scrolling itself. Using the hash ring filter during scrolling itself is very expensive and makes the transfer **significantly** slower, bumping transfer time from 28 to 500 seconds.
Branch with it implemented: https://github.com/qdrant/qdrant/commits/improve-transfer-progress-tracker-hashring-filter/

Example collection cluster info during resharding transfer (now shows 250k and correct record count):
```json
{
  "peer_id": 6294971774567105,
  "shard_count": 2,
  "local_shards": [
    {
      "shard_id": 0,
      "points_count": 500000,
      "state": "Active"
    },
    {
      "shard_id": 1,
      "points_count": 145399,
      "state": "Resharding"
    }
  ],
  "shard_transfers": [
    {
      "shard_id": 0,
      "to_shard_id": 1,
      "from": 6294971774567105,
      "to": 6294971774567105,
      "sync": true,
      "method": "resharding_stream_records",
      "comment": "Transferring records (145399/250000), started 16s ago, ETA: 11.68s"
    }
  ],
  // ...
}
```

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/transfer/transfer_tasks_pool.rs
+++ b/lib/collection/src/shards/transfer/transfer_tasks_pool.rs
@@ -22,8 +22,8 @@ pub struct TransferTaskItem {
 }
 
 pub struct TransferTaskProgress {
-    pub points_transferred: usize,
-    pub points_total: usize,
+    points_transferred: usize,
+    points_total: usize,
     pub eta: EtaCalculator,
 }
 
@@ -48,6 +48,18 @@ impl TransferTaskProgress {
             eta: EtaCalculator::new(),
         }
     }
+
+    pub fn add(&mut self, delta: usize) {
+        self.points_transferred += delta;
+        self.points_total = max(self.points_total, self.points_transferred);
+        self.eta.set_progress(self.points_transferred);
+    }
+
+    pub fn set(&mut self, transferred: usize, total: usize) {
+        self.points_transferred = transferred;
+        self.points_total = total;
+        self.eta.set_progress(transferred);
+    }
 }
 
 impl TransferTasksPool {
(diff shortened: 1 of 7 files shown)


lib/collection/src/shards/transfer/transfer_tasks_pool.rs
```diff
@@ -48,6 +48,18 @@ impl TransferTaskProgress {
             eta: EtaCalculator::new(),
         }
     }
+
+    pub fn add(&mut self, delta: usize) {
+        self.points_transferred += delta;
+        self.points_total = max(self.points_total, self.points_transferred);
```
agourlay, line 54: Why is `points_total` changing during the resharding?
Is to account for some ongoing updates?

timvisee: The total count is an estimate, as it's **very** expensive to do exact counting here.

It therefore is expected to overshoot the estimated total about 50% of the time.

agourlay (approved):

PR #6085, opened 2025-02-28
Optimize cleanup of shards, used in resharding

timvisee:
Optimize the background task that takes care of cleaning up shards, specifically deleting points from it that don't belong in it anymore.

This change makes the cleanup process a lot faster, which means it'll complete in less wall clock time.

The primary change is about how we select what points to delete. Before, we used a hash ring filter during the scroll request to only scroll points that don't belong into the shard anymore. Now, we simply scroll all point IDs and filter  the list after the scroll request. The key thing here is that the hash ring check is very expensive. Before, the hash ring check was done a lot more often across all segments in the shard. Now we only check each point ID once.

To summarize, this PR does two things:
1. move hash ring check outside of scroll operation
2. use wait=false on all delete operations, except the last one

I don't have a fancy graph, but I do have performance numbers. This is on a single node, with 5 million points, resharding from 1 to 2 shards, with 2 or 100 segments.

Time for shard cleanup to complete:

0. Before this PR:
    - 2 segments: 115 seconds
    - 100 segments: 72 seconds
1. With hash ring check outside scroll operation:
    - 2 segments: 35 seconds
    - 100 segments: 53 seconds
2. With wait=false:
    - 2 segments: 20 seconds
    - 100 segments: 43 seconds

### All Submissions:

### New Feature Submissions:

4. [x] Does your submission pass tests?

### DIFF
--- a/lib/collection/src/collection/clean.rs
+++ b/lib/collection/src/collection/clean.rs
@@ -6,7 +6,7 @@ use std::time::{Duration, Instant};
 use cancel::{CancellationToken, DropGuard};
 use common::counter::hardware_accumulator::HwMeasurementAcc;
 use parking_lot::RwLock;
-use segment::types::{Condition, Filter};
+use segment::types::ExtendedPointId;
 use tokio::sync::watch::{Receiver, Sender};
 use tokio::task::JoinHandle;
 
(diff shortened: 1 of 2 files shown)


agourlay:
Out of curiosity, why is a check using the Hashring so expensive?

generall:
I am not entirely understand why number of comparisons is different. In scroll we do not check all points, but only points which are candidates to our limit+offset pool. Is that because we search top-N results in each segment independently?

generall (approved):

generall:
CPU usage before and after this fix: 
image

timvisee:
> Is that because we search top-N results in each segment independently?

Yes

> CPU usage before and after this fix:
> image

Awesome! Thank you for sharing the results. The difference seems bigger than I expected.

I assume that part of the spikes in the new part of the graph are just because of indexing.

PR #6092, opened 2025-03-03
[score boosting] Only accept finite numbers

coszio:
We decided to only handle finite numbers in the formula evaluation. 

With these changes, we now return a new bad request error when the evaluation leads to infinite or NaN.

Test cases were added to assert the errors

### DIFF
--- a/lib/segment/src/index/query_optimization/rescore_formula/formula_scorer.rs
+++ b/lib/segment/src/index/query_optimization/rescore_formula/formula_scorer.rs
@@ -141,15 +141,19 @@ impl FormulaScorer<'_> {
                 if left == 0.0 {
                     return Ok(0.0);
                 }
-                let right = self.eval_expression(right, point_id)?;
-                // avoid division by zero
+                let mut right = self.eval_expression(right, point_id)?;
+
                 if right == 0.0 {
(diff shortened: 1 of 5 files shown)


lib/segment/src/index/query_optimization/rescore_formula/formula_scorer.rs
```diff
@@ -144,7 +144,9 @@ impl FormulaScorer<'_> {
                 let right = self.eval_expression(right, point_id)?;
                 // avoid division by zero
                 if right == 0.0 {
-                    Ok(*by_zero_default)
+                    by_zero_default.ok_or_else(|| OperationError::NonFiniteNumber {
+                        expression: format!("{left}/{right}"),
```
timvisee, line 148: Would it make sense to include the result, being NaN or Inf?

```suggestion
                        expression: format!("{left}/{right}={}", left / right),
```

timvisee (approved):

lib/segment/src/index/query_optimization/rescore_formula/formula_scorer.rs
coszio: Probably yes, so that it makes it easier to choose a sane default

PR #6107, opened 2025-03-04
Remove validation on CollectionInfo response type

agourlay:
This PR removes the validation on the `CollectionInfo` type.

The motivation is that this type is used only in responses and never constructed by the user.

The validation is never called on it at the API level.


lib/collection/src/collection_state.rs
```diff
@@ -16,9 +15,8 @@ pub struct ShardInfo {
     pub replicas: HashMap<PeerId, ReplicaState>,
 }
 
-#[derive(Debug, Serialize, Deserialize, Validate, Clone, PartialEq)]
```
agourlay, line 19: `State` is never validated

generall (approved):

timvisee (approved):

PR #6113, opened 2025-03-05
Use Struct deconstruction in config diffs

agourlay:
Seems like an oversight from the recent migration to struct deconstruction.


lib/collection/src/operations/config_diff.rs
```diff
@@ -1,4 +1,3 @@
-use std::hash::Hash;
```
agourlay, line 1: we use the fully qualified path everywhere

timvisee (approved):

PR #6114, opened 2025-03-05
Separate StrictModeConfig output structure

agourlay:
Separate the whole struct tree of `StrictModeConfig` used in the `CollectionInfo` in order to enable the Python REST client to to tag those types as accepting extra keys during deserialization.

This is required to keep the [backward compatibility](https://github.com/qdrant/qdrant-client/issues/906) between the Python client 1.13.2 and the upcoming 1.14.0 version.

We have used a similar pattern in the past for `OptimizerConfig` and `OptimizerConfigDiff` for the same issue.

As a final test I validated that the new output classes are tagged properly in the [client generation](https://github.com/qdrant/qdrant-client/pull/916)

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -11416,7 +11416,7 @@
             "minimum": 0
           },
           "config": {
-            "$ref": "#/components/schemas/CollectionConfigInternal"
+            "$ref": "#/components/schemas/CollectionConfigTelemetry"
           },
           "shards": {
             "type": "array",
@@ -11444,7 +11444,7 @@
           }
         }
       },
-      "CollectionConfigInternal": {
+      "CollectionConfigTelemetry": {
         "type": "object",
         "required": [
           "hnsw_config",
@@ -11479,7 +11479,7 @@
           "strict_mode_config": {
             "anyOf": [
               {
-                "$ref": "#/components/schemas/StrictModeConfig"
+                "$ref": "#/components/schemas/StrictModeConfigOutput"
               },
               {
                 "nullable": true
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -6704,7 +6704,7 @@
           "strict_mode_config": {
             "anyOf": [
               {
-                "$ref": "#/components/schemas/StrictModeConfig"
+                "$ref": "#/components/schemas/StrictModeConfigOutput"
               },
               {
                 "nullable": true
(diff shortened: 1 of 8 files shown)


lib/collection/src/operations/types.rs
```diff
@@ -229,7 +229,7 @@ impl From<CollectionConfigInternal> for CollectionConfig {
             optimizer_config,
             wal_config: Some(wal_config),
             quantization_config,
-            strict_mode_config,
+            strict_mode_config: strict_mode_config.map(StrictModeConfigOutput::from),
```
generall, line 232: `CollectionConfigInternal` is used in collection telemetry, and, therefore, should either be splitted into Output structure, or contain `StrictModeConfigOutput`

generall (changes requested):

docs/redoc/master/openapi.json
```diff
@@ -7253,7 +7253,7 @@
           }
         }
       },
-      "StrictModeConfig": {
+      "StrictModeConfigOutput": {
```
timvisee, line 7256: This means generated clients will get a different type name, and that might need migration by users. Is there no other way to achieve the same without splitting into two types?

timvisee (commented):
> Extra inputs are not permitted

I'd argue that pydantic should ignore unknown field instead.

docs/redoc/master/openapi.json
agourlay: To recap about the blast radius of the change:
- this is a change only for the REST clients (Python & js)
- this is an output type that users never construct

The content of the objects are identical, so it will be observable only by users explicitly type checking against the `StrictModeConfig` type which seems unlikely in Python & js.

I do not see another way of pleasing the current Pydantic setup without changing it.

timvisee: > The content of the objects are identical

In that case I don't see why we need a different object at all.

> I do not see another way of pleasing the current Pydantic setup without changing it.

What do you think about my above comment? Don't you agree pydantic should ignore unknown fields? If it doesn't, it'll be very hard to add new fields (across the API) in the future too.

generall: > In that case I don't see why we need a different object at all.

For backward compatibility purposes. 

> I'd argue that pydantic should ignore unknown field instead.

This is undesirable. We want validation of inputs. I don't remember the exact case why, but it was strong enough to go through this separation of input-output in the first place

generall (approved):

agourlay:
I am merging this PR because I have been able to show that:
- the infinite recursion is due to the new `Expression` type used in scoring and NOT the new Strict Mode Config output types
- the infinite recursion can be fixed anyway (still needs Georges' validation) https://github.com/qdrant/qdrant-client/pull/919

PR #6116, opened 2025-03-05
[score boosting] fix `by_zero_default` evaluation

coszio:
small fix for the way the division default is used.

`by_zero_default` is the default result, not denominator


timvisee (approved):

agourlay:
Is this tested only with the local mode?

PR #6117, opened 2025-03-06
Cardinality estimation IO measurements

JojiiOfficial:
Depends on #5954.

There are places where we would benefit from #5985 which have been left out and marked with a TODO.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
@@ -44,12 +44,11 @@ impl InvertedIndex for ImmutableInvertedIndex {
         true
     }
 
-    fn filter(
-        &self,
-        query: &ParsedQuery,
-        hw_counter: &HardwareCounterCell,
-    ) -> Box<dyn Iterator<Item = PointOffsetType> + '_> {
-        let hw_counter = hw_counter.payload_index_io_read_counter();
+    fn filter<'a>(
+        &'a self,
+        query: ParsedQuery,
+        hw_counter: &'a HardwareCounterCell,
+    ) -> Box<dyn Iterator<Item = PointOffsetType> + 'a> {
         let postings_opt: Option<Vec<_>> = query
             .tokens
             .iter()
@@ -58,10 +57,6 @@ impl InvertedIndex for ImmutableInvertedIndex {
                 // if a ParsedQuery token was given an index, then it must exist in the vocabulary
                 Some(idx) => {
                     let postings = self.postings.get(idx as usize);
-                    hw_counter.incr_delta(
-                        size_of::<Option<CompressedPostingList>>()
-                            + postings.map(|i| i.len()).unwrap_or(0) * size_of::<PointOffsetType>(),
-                    );
                     postings
                 }
             })
@@ -73,7 +68,10 @@ impl InvertedIndex for ImmutableInvertedIndex {
             _ => return Box::new(vec![].into_iter()),
         };
 
-        let posting_readers: Vec<_> = postings.iter().map(|posting| posting.reader()).collect();
+        let posting_readers: Vec<_> = postings
+            .iter()
+            .map(|posting| posting.reader(hw_counter))
+            .collect();
 
         // in case of immutable index, deleted documents are still in the postings
         let filter =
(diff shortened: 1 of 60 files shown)


lib/segment/src/index/field_index/geo_index/mod.rs
```diff
@@ -544,7 +544,11 @@ impl PayloadFieldIndex for GeoMapIndex {
         None
     }
 
-    fn estimate_cardinality(&self, condition: &FieldCondition) -> Option<CardinalityEstimation> {
+    fn estimate_cardinality(
+        &self,
+        condition: &FieldCondition,
+        _hw_counter: &HardwareCounterCell, // Nothing to measure here.
```
coszio, line 550: Are you sure about this? I see that at least we read `points_of_hash` from the index, and we have mutable, immutable, and mmap variants

lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
```diff
@@ -82,8 +82,16 @@ impl InvertedIndex for ImmutableInvertedIndex {
...
-        self.postings.get(token_id as usize).map(|p| p.len())
+    fn get_posting_len(
+        &self,
+        token_id: TokenId,
+        hw_counter: &HardwareCounterCell,
+    ) -> Option<usize> {
+        let len = self.postings.get(token_id as usize).map(|p| p.len());
+        hw_counter
+            .payload_index_io_read_counter()
+            .incr_delta(size_of::<Option<CompressedPostingList>>() + len.unwrap_or(0));
```
coszio, line 93: `CompressedPostingList::len()` only reads the len of two vecs, does that still count toward io reads?

lib/segment/src/index/field_index/full_text_index/mutable_inverted_index.rs
```diff
@@ -161,12 +161,21 @@ impl InvertedIndex for MutableInvertedIndex {
...
+        let len = self
+            .postings
             .get(token_id as usize)
             .and_then(|posting| posting.as_ref())
             .as_ref()
-            .map(|x| x.len())
+            .map(|x| x.len());
+        hw_counter
+            .payload_index_io_read_counter()
+            .incr_delta(size_of::<Option<PostingList>>() + len.unwrap_or(0));
```
coszio, line 177: Same here, this `len` only reads the len of 1 vec

lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
```diff
@@ -60,7 +60,7 @@ impl InvertedIndex for ImmutableInvertedIndex {
                     let postings = self.postings.get(idx as usize);
                     hw_counter.incr_delta(
                         size_of::<Option<CompressedPostingList>>()
-                            + postings.map(|i| i.len()).unwrap_or(0),
+                            + postings.map(|i| i.len()).unwrap_or(0) * size_of::<u32>(),
```
agourlay, line 63: why do we multiply by `size_of::<u32>()` here?

JojiiOfficial: `postings` is of type `Vec<PointOffsetType>`. To calculate the size of `postings`, we have to multiply the length by the size of each item.
Giving it a second thought, we can simply use `size_of_val` inside the `map(..)` function instead

agourlay (approved):

lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
```diff
@@ -92,11 +92,11 @@ impl MmapPostings {
...
     /// remainder_postings: &'a [PointOffsetType],
     /// ```
-    fn get_reader(
-        &self,
-        header: &PostingListHeader,
-        hw_counter: &HardwareCounterCell,
-    ) -> Option<ChunkReader<'_>> {
+    fn get_reader<'a>(
+        &'a self,
+        header: &'a PostingListHeader,
```
JojiiOfficial, line 97: `header` doesn't need a lifetime here, since it is not part of the returned value.

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -640,11 +641,11 @@ impl<T: Encodable + Numericable + MmapValue + Default> PayloadFieldIndex for Num
...
     }
 
-    fn filter(
-        &self,
-        condition: &FieldCondition,
-        hw_acc: HwMeasurementAcc,
-    ) -> Option<Box<dyn Iterator<Item = PointOffsetType> + '_>> {
+    fn filter<'a>(
+        &'a self,
+        condition: &'a FieldCondition,
```
JojiiOfficial, line 646: No lifetime needed for `condition`

timvisee (commented):
Should we be concerned about double-counting, where we count some operation twice during the cardinality estimation and also while doing the operation itself?

I did some basic test with this, but definitely was not able to test all possible paths. Since this covers so much, it's very hard to keep track of what branches there are, and which we do or do not count.

generall (approved):
I made test measurements on keyword and text indexes, after some fixes they now match my expectations. Also, found a potential place for optimization - we can try to exclude primary condition from second stage filtering.

I propose to merge this and proceed with further steps: create a checklist and some docs, which would allow external people to understand what to expect from measurement and how to test them.

PR #6118, opened 2025-03-07
Do not rate limit scroll operations for shard transfers

agourlay:
Do not apply read rate limit to scroll operation issued for shard transfers.

The fix relies on leveraging the work done by tagging internal operations with disposable hardware measurements.

Now all operations with disposable hardware measurements by-pass the read rate limiter.

### DIFF
--- a/tests/consensus_tests/test_shard_transfer_rate_limiting.py
+++ b/tests/consensus_tests/test_shard_transfer_rate_limiting.py
@@ -82,13 +82,13 @@ def test_shard_transfer_rate_limiting(tmp_path: pathlib.Path):
     shard_id = collection_cluster_info["local_shards"][0]["shard_id"]
     source_peer_id = collection_cluster_info["peer_id"]
 
-    # set rate limiting
+    # Set rate limiting
     r = requests.patch(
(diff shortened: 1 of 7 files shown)


tests/consensus_tests/test_shard_transfer_rate_limiting.py
```diff
@@ -82,13 +82,13 @@ def test_shard_transfer_rate_limiting(tmp_path: pathlib.Path):
...
 
-    # set rate limiting
+    # Set rate limiting
     r = requests.patch(
         f"{source_uri}/collections/test_collection", json={
             "strict_mode_config": {
                 "enabled": True,
                 "write_rate_limit": 1,
-                # TODO(ratelimit) validate read limits as well "read_rate_limit": 1
+                "read_rate_limit": 1,
```
agourlay, line 91: without the fix, enabling this line makes the test fail due to read rate limit reached :)

timvisee (approved):
I like the idea of the disposable flag, to just reuse the HwMeasurement type.

I explicitly checks this is sound for all shard transfer types. :+1:

JojiiOfficial (approved):

PR #6127, opened 2025-03-10
Bump Ring 0.17.13

agourlay:
Replacing https://github.com/qdrant/qdrant/pull/6122 for `dev`


timvisee (approved):
Thanks!

I don't deem the DoS risk important enough to port it to master right away. So lets keep it on `dev` until we release.

PR #6131, opened 2025-03-10
Remove code duplication for recommendation conversion

agourlay:
Stumbled upon that one while performing the struct deconstruction migration.

Decided to finally get rid of the TODO.


coszio (approved):
Thank you! 🧹

PR #6132, opened 2025-03-10
Payload Index Io measurements with Iterator

JojiiOfficial:
Depends on #6117

### DIFF
--- a/lib/common/common/src/counter/hardware_counter.rs
+++ b/lib/common/common/src/counter/hardware_counter.rs
@@ -17,6 +17,7 @@ pub struct HardwareCounterCell {
     pub(super) vector_io_read_counter: CounterCell,
     pub(super) vector_io_write_counter: CounterCell,
     pub(super) accumulator: HwMeasurementAcc,
+    disposable: bool,
 }
 
 impl HardwareCounterCell {
@@ -31,6 +32,7 @@ impl HardwareCounterCell {
             vector_io_read_counter: CounterCell::new(),
             vector_io_write_counter: CounterCell::new(),
             accumulator: HwMeasurementAcc::new(),
+            disposable: false,
         }
     }
 
@@ -47,27 +49,7 @@ impl HardwareCounterCell {
             vector_io_read_counter: CounterCell::new(),
             vector_io_write_counter: CounterCell::new(),
             accumulator: HwMeasurementAcc::disposable(),
-        }
-    }
-
-    #[cfg(feature = "testing")]
-    pub fn new_with(
-        cpu: usize,
-        payload_io_read: usize,
-        payload_io_write: usize,
-        payload_index_io_read: usize,
-        vector_io_read: usize,
-        vector_io_write: usize,
-    ) -> Self {
-        Self {
-            cpu_multiplier: 1,
-            cpu_counter: CounterCell::new_with(cpu),
-            payload_io_read_counter: CounterCell::new_with(payload_io_read),
-            payload_io_write_counter: CounterCell::new_with(payload_io_write),
-            payload_index_io_read_counter: CounterCell::new_with(payload_index_io_read),
-            vector_io_read_counter: CounterCell::new_with(vector_io_read),
-            vector_io_write_counter: CounterCell::new_with(vector_io_write),
-            accumulator: HwMeasurementAcc::new(),
+            disposable: true,
         }
     }
 
@@ -80,6 +62,7 @@ impl HardwareCounterCell {
             payload_index_io_read_counter: CounterCell::new(),
             vector_io_read_counter: CounterCell::new(),
             vector_io_write_counter: CounterCell::new(),
+            disposable: accumulator.is_disposable(),
             accumulator,
         }
     }
@@ -101,6 +84,7 @@ impl HardwareCounterCell {
             vector_io_read_counter: CounterCell::new(),
             vector_io_write_counter: CounterCell::new(),
             accumulator: self.accumulator.clone(),
+            disposable: self.disposable,
         }
     }
 
@@ -183,6 +167,7 @@ impl HardwareCounterCell {
             vector_io_read_counter,
             vector_io_write_counter,
             accumulator: _,
+            disposable: _,
         } = self;
 
         HardwareData {
--- a/lib/common/common/src/counter/hardware_counter.rs
+++ b/lib/common/common/src/counter/hardware_counter.rs
@@ -17,7 +17,6 @@ pub struct HardwareCounterCell {
     pub(super) vector_io_read_counter: CounterCell,
     pub(super) vector_io_write_counter: CounterCell,
     pub(super) accumulator: HwMeasurementAcc,
-    disposable: bool,
 }
 
 impl HardwareCounterCell {
@@ -32,7 +31,6 @@ impl HardwareCounterCell {
             vector_io_read_counter: CounterCell::new(),
             vector_io_write_counter: CounterCell::new(),
             accumulator: HwMeasurementAcc::new(),
-            disposable: false,
         }
     }
 
@@ -49,7 +47,6 @@ impl HardwareCounterCell {
             vector_io_read_counter: CounterCell::new(),
             vector_io_write_counter: CounterCell::new(),
             accumulator: HwMeasurementAcc::disposable(),
-            disposable: true,
         }
     }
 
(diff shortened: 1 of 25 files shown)


lib/segment/src/index/field_index/map_index/mutable_map_index.rs
```diff
@@ -187,11 +193,16 @@ impl<N: MapIndexKey + ?Sized> MutableMapIndex<N> {
...
+        hw_acc: HwMeasurementAcc,
     ) -> impl Iterator<Item = (&N, IdIter<'_>)> + '_ {
-        self.map
-            .iter()
-            .map(|(k, v)| (k.borrow(), Box::new(v.iter().copied()) as IdIter))
+        let hw_cell = hw_acc.get_counter_cell();
+        self.map.iter().map(move |(k, v)| {
+            hw_cell
+                .payload_index_io_read_counter()
+                .incr_delta(v.len() + N::mmapped_size(MmapValue::as_referenced(k.borrow())));
```
coszio, line 202: Also this. In this case, I think it should be `v.len() * size_of<PointOffsetType>()`

coszio:
I realized we could use the precomputed counts more in mmap bool index. Added the change in https://github.com/qdrant/qdrant/pull/6132/commits/a75d952dcf30f35aa57cf8040d3cf407ff13b9e6

lib/segment/src/index/field_index/map_index/mmap_map_index.rs
```diff
@@ -273,9 +278,15 @@ impl<N: MapIndexKey + Key + ?Sized> MmapMapIndex<N> {
...
-        _hw_acc: HwMeasurementAcc, // TODO(io_measurement): Collect values.
+        hw_acc: HwMeasurementAcc,
     ) -> impl Iterator<Item = (&N, IdIter<'_>)> + '_ {
-        self.value_to_points.iter().map(|(k, v)| {
+        let hw_counter = hw_acc.get_counter_cell();
+
+        self.value_to_points.iter().map(move |(k, v)| {
+            hw_counter
+                .payload_index_io_read_counter()
+                .incr_delta(v.len() + k.write_bytes());
```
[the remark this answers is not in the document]
JojiiOfficial: Yeah you're right, we should use `size_of_val()` instead of `len()` here to account for the size of elements in `v`.
If I'm not mistaken, `v` is of type `PointOffsetType`, which is always the size of `u32`. Or did you mean we need to measure something else?

coszio (approved):

agourlay (approved):

lib/segment/src/index/field_index/numeric_index/mmap_numeric_index.rs
```diff
@@ -204,6 +204,11 @@ impl<T: Encodable + Numericable + Default + MmapValue> MmapNumericIndex<T> {
         check_fn: impl Fn(&T) -> bool,
         hw_counter: &HardwareCounterCell,
     ) -> bool {
+        // Measuring self.deleted read.
+        hw_counter
+            .payload_index_io_read_counter()
+            .incr_delta(size_of::<bool>());
```
generall, line 210: seems insignificant

lib/segment/src/index/field_index/map_index/mutable_map_index.rs
```diff
@@ -176,22 +176,39 @@ impl<N: MapIndexKey + ?Sized> MutableMapIndex<N> {
     pub fn get_count_for_value(
         &self,
         value: &N,
-        _hw_counter: &HardwareCounterCell, // TODO(io_measurement): Collect values
+        hw_counter: &HardwareCounterCell,
     ) -> Option<usize> {
-        self.map.get(value).map(|p| p.len())
+        let counter = hw_counter.payload_index_io_read_counter();
+        counter.incr_delta(BUCKET_OFFSET_OVERHEAD);
```
generall, line 182: `BUCKET_OFFSET_OVERHEAD` only makes sense for mmap hashmap

lib/segment/src/index/field_index/bool_index/mmap_bool_index.rs
```diff
@@ -202,12 +211,18 @@ impl MmapBoolIndex {
...
+    pub fn iter_values_map<'a>(
+        &'a self,
+        hw_counter: &'a HardwareCounterCell,
+    ) -> impl Iterator<Item = (bool, IdIter<'a>)> + 'a {
         [
             (false, Box::new(self.falses_slice.iter_trues()) as IdIter),
             (true, Box::new(self.trues_slice.iter_trues()) as IdIter),
         ]
         .into_iter()
+        .measure_hw_with_cell(hw_counter, size_of::<usize>(), |i| {
```
generall, line 223: why do we count usize for bool values?

lib/segment/src/index/field_index/bool_index/simple_bool_index.rs
```diff
@@ -170,10 +173,34 @@ mod memory {
...
         }
 
+        pub fn iter_has_false_measured<'a>(
+            &'a self,
+            hw_counter: &'a HardwareCounterCell,
+        ) -> impl Iterator<Item = PointOffsetType> + 'a {
+            self.falses
+                .iter_ones()
+                .map(|v| v as PointOffsetType)
+                .measure_hw_with_cell_and_fraction(hw_counter, u8::BITS as usize, |i| {
```
generall, line 199: Fot this we should have 1/8 of a byte per value, not the other way around

JojiiOfficial: the `*_and_fraction` function calculates `1/fraction` and not `*fraction`, it's a different function than in other places

lib/common/common/src/counter/hardware_counter.rs
```diff
@@ -17,6 +17,7 @@ pub struct HardwareCounterCell {
     pub(super) vector_io_read_counter: CounterCell,
     pub(super) vector_io_write_counter: CounterCell,
     pub(super) accumulator: HwMeasurementAcc,
+    disposable: bool,
```
generall, line 20: What's the purpose? Doesn't seem to be used anywhere

lib/common/common/src/iterator_ext/mod.rs
```diff
@@ -42,6 +46,68 @@ pub trait IteratorExt: Iterator {
...
         OnFinalCount::new(self, f)
     }
+
+    /// Measures the hardware usage of an iterator.
+    ///
+    /// # Arguments
+    /// - `hw_acc`: accumulator holding a counter cell
+    /// - `multiplier`: multiplies the number of iterations by this factor.
+    /// - `f`: Closure to get the specific counter to increase from the cell inside the accumulator.
+    fn measure_hw_with_acc<R>(
```
generall, line 56: It doesn't feel like Hw related functions should be in the pretty low-level `IteratorExt`. I would prefer to have it on higher level of abstraction, similar to our discussion about MmapHashmap

lib/common/common/src/iterator_ext/mod.rs
```diff
@@ -42,6 +46,68 @@ pub trait IteratorExt: Iterator {
...
+        multiplier: usize,
+        mut f: R,
+    ) -> OnFinalCount<Self, impl FnMut(usize)>
+    where
+        Self: Sized,
+        R: FnMut(&HardwareCounterCell) -> &CounterCell,
+    {
+        OnFinalCount::new(self, move |i| {
+            let hw_counter = hw_acc.get_counter_cell();
+            // Subtract 1 to not account for the latest `None` call.
```
generall, line 68: Does it mean that `i` in this context is not a number of values in the iterator?

If so, shouldn't it be fixed inside `OnFinalCount`? Or if it is intended, what's a true meaning of `i`? I would prefer a more explicit naming if possible.

JojiiOfficial: `i` represents how many times we called `.next()` on an iterator, which might be less than the actual amount of items in the iterator e.g. if we stopped early.
The call of `next()` that returns `None` is counted too in `i`. This behavior is documented on the `on_final_count` function of this trait. Because we don't want this measurement, we subtract `1` here.

lib/segment/src/index/field_index/bool_index/mmap_bool_index.rs
JojiiOfficial: I confused underlying datatype (usize) with the actual bit type. Fixed now 👍🏻

timvisee (approved):

generall (approved):

PR #6133, opened 2025-03-10
Declare shard green if there can be no optimization threads

KShivendu:
One of the user clusters got stuck in replica wait on indexing stage. This happened because they had configured `max_optimization_threads == 0` which returns Yellow status for the collection as optimizations are possible but can't be triggered because of not having any optimization threads.

This PR makes Qdrant returns Green status in such scenario so we don't get stuck.

### All Submissions:

### TODO:
- [ ] Add test

### DIFF
--- a/lib/collection/src/tests/mod.rs
+++ b/lib/collection/src/tests/mod.rs
@@ -405,13 +405,14 @@ async fn test_shard_status_based_on_optimization(
         collection_config_guard
             .optimizer_config
             .default_segment_number = 1;
+        drop(collection_config_guard);
+
+        let info = shard.local_shard_info().await;
+        assert_eq!(info.status, ShardStatus::Green);
+
+        // This recreates optimizers based on updated collection config
+        shard.on_optimizer_config_update().await.unwrap();
     }
-
-    let info = shard.local_shard_info().await;
-    assert_eq!(info.status, ShardStatus::Green);
-
-    // This recreates optimizers based on updated collection config
-    shard.on_optimizer_config_update().await.unwrap();
 
     let info = shard.local_shard_info().await;
     assert_eq!(info.status, expected_shard_status);
--- a/lib/collection/src/tests/mod.rs
+++ b/lib/collection/src/tests/mod.rs
@@ -9,20 +9,26 @@ mod snapshot_test;
 mod sparse_vectors_validation_tests;
 mod wal_recovery_test;
 
+use std::panic;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time::Duration;
 
 use common::budget::ResourceBudget;
+use common::counter::hardware_accumulator::HwMeasurementAcc;
 use common::counter::hardware_counter::HardwareCounterCell;
+use fixtures::{create_collection_config, upsert_operation};
 use futures::future::join_all;
 use itertools::Itertools;
 use parking_lot::{Mutex, RwLock};
 use rand::Rng;
+use rstest::rstest;
 use segment::data_types::vectors::only_default_vector;
 use segment::index::hnsw_index::num_rayon_threads;
 use segment::types::{Distance, PointIdType};
 use tempfile::Builder;
+use tokio::runtime::Handle;
+use tokio::sync::RwLock as TokioRwLock;
 use tokio::time::{Instant, sleep};
 
 use crate::collection::Collection;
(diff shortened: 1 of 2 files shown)


KShivendu:
> Green status means all the points have been processed and the index are fresh

It just means that collection is ready to use. See [docs](https://qdrant.tech/documentation/concepts/collections/)

Please also check [this](https://qdrant.slack.com/archives/C077CCW8NU8/p1741614209370129?thread_ts=1741602865.964849&cid=C077CCW8NU8) and [this](https://qdrant.slack.com/archives/C077CCW8NU8/p1741614376536909?thread_ts=1741602865.964849&cid=C077CCW8NU8) and lmk if you don't agree :)

timvisee:
@agourlay 

I agree that green is better than grey here. Green means we're done with indexing defined by how the user configured it. Grey would imply that there's still things left to do. It might give the false impression that sending an update will trigger optimizations again (which it does not).

Alternatively we could add another status color, but I'd like to prevent that if possible.

Wdyt?

lib/collection/src/update_handler.rs
```diff
@@ -465,6 +465,12 @@ impl UpdateHandler {
         // Check if Qdrant triggered any optimizations since starting at all
         let has_triggered_any_optimizers = self.has_triggered_optimizers.load(Ordering::Relaxed);
 
+        if self.max_optimization_threads == Some(0) {
+            // Collection cannot have optimization threads. So we cannot have suboptimal optimizers
+            // This eventually leads to ShardStatus::Green
+            return (has_triggered_any_optimizers, false);
+        }
```
timvisee, line 472: Just return `true` here for `has_triggered_any_optimizers` (I think).

In this specific case it doesn't mater whether optimizers are triggered once or not, and the current implementation still shows an intermediate Grey status.

I think it's better to show Green right away.

```suggestion
        if self.max_optimization_threads == Some(0) {
            // Collection cannot have optimization threads. So we cannot have suboptimal optimizers
            // This eventually leads to ShardStatus::Green
            return (true, false);
        }
```

lib/collection/src/tests/mod.rs
```diff
@@ -342,3 +351,68 @@ fn check_version_upgrade() {
...
+        .await
+        .unwrap();
+
+    // This should trigger segment merge optimizer if we have optimizer threads (i.e. != Some(0))
+    {
+        let mut collection_config_guard = collection_config.write().await;
+        collection_config_guard
+            .optimizer_config
+            .default_segment_number = 1;
+    }
```
timvisee, line 408: This by itself will not trigger the optimizer, because it doesn't fire an update.

We should probably do the operation shown [here](https://qdrant.tech/documentation/concepts/collections/#grey-collection-status) as well to actually _trigger_.

KShivendu:
> Alternatively we could add another status color, but I'd like to prevent that if possible.

No please. Too much complexity for user.

lib/collection/src/tests/mod.rs
KShivendu: > optimizer, because it doesn't fire an update

Yes but I'm doing this afterwards and tests are passing:
```rs
// This recreates optimizers based on updated collection config
shard.on_optimizer_config_update().await.unwrap();
```

lib/collection/src/update_handler.rs
KShivendu: > and the current implementation still shows an intermediate Grey status

I don't think so. How? :thinking: 

> Just return true here for has_triggered_any_optimizers

I think they are independent variables. `has_triggered_any_optimizers` can be both true or false even it doesn't matter for now because `has_suboptimal_optimizers = false`.

agourlay:
@timvisee Ok, with green, the following sentence convinced me.

> Green means we're done with indexing defined by how the user configured it. 

We just need to make sure it is clear for the users as well :+1:

generall:
I am not sure about showing green in this scenario. I would either prefer Gray or keep as is.
Downside of Gray is that we have web-ui logic to trigger optimizers, and it won't have any effect if indexing threads are disabled.

And Green will make it harder to spot the problem

lib/collection/src/tests/mod.rs
timvisee: > I wanted to drop the guard automatically instead of calling `drop()` myself so I kept that part inside a block. Do you know any better syntax for this?

Yes, just this:

```rust
    // Change configuration and trigger merge optimizer
    collection_config.write().await.optimizer_config.default_segment_number = 1;
    let info = shard.local_shard_info().await;
    assert_eq!(info.status, ShardStatus::Green);
    shard.on_optimizer_config_update().await.unwrap();
```

KShivendu:
> I would either prefer Gray or keep as is.

Replica wait on indexing will get stuck in such collections then. CM doesn't have a UI for customers to understand that it's waiting on indexing. They'll remain confused why cluster is not scaling.

> Downside of Gray is that we have web-ui logic to trigger optimizers, and it won't have any effect if indexing threads are disabled.

Yeah and that's even more confusing for the user. If user configured `max_optimization_threads = 0`, imo it's correct assume that it's the desired behaviour. 

> Green means we're done with indexing defined by how the user configured it.

I think we are still aligned with this definition. It's a good reason to keep it green.

KShivendu:
Alternatively, we can make it Grey and warn user that collection `max_optimization_threads` is set to 0 when they press "Trigger Optimizers" button.

timvisee:
Alright! There are multiple perspectives to this, and I'm also fine with keeping it as-is. At least we agree we should not add another status color just for this.

So: let's keep it like it is already, and handle it within CM instead :+1:

PR #6148, opened 2025-03-11
WIP: Mmap in-ram payload indexes without RocksDB

generall:
We are trying to get rid of RocksDB everywhere, this is another place

### DIFF
--- a/lib/segment/src/index/struct_payload_index.rs
+++ b/lib/segment/src/index/struct_payload_index.rs
@@ -14,9 +14,7 @@ use schemars::_serde_json::Value;
 
 use super::field_index::FieldIndexBuilderTrait as _;
 use super::field_index::facet_index::FacetIndexEnum;
-use super::field_index::index_selector::{
-    IndexSelector, IndexSelectorOnDisk, IndexSelectorRocksDb,
-};
+use super::field_index::index_selector::{IndexSelector, IndexSelectorMmap, IndexSelectorRocksDb};
 use crate::common::Flusher;
 use crate::common::operation_error::{OperationError, OperationResult};
 use crate::common::rocksdb_wrapper::open_db_with_existing_cf;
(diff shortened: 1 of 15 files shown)


lib/segment/src/index/struct_payload_index.rs
```diff
@@ -156,11 +160,30 @@ impl StructPayloadIndex {
         let config = if config_path.exists() {
             PayloadConfig::load(&config_path)?
         } else {
+            // ToDo(mmap-paylaod-index): uncomment before minor release
```
agourlay, line 163: Did you manually test the `skip_rocksdb` part?

generall: there is one green CI test pass with this option enabled and one (latest) where this option is disabled

agourlay (approved):

lib/segment/src/vector_storage/dense/appendable_dense_vector_storage.rs
```diff
@@ -289,7 +291,8 @@ pub fn open_appendable_in_ram_vector_storage_impl<T: PrimitiveVectorElement>(
 
     let vectors = InRamPersistedVectors::<T>::open(&vectors_path, dim)?;
 
-    let deleted: DynamicMmapFlags = DynamicMmapFlags::open(&deleted_path)?;
+    let populate = true;
+    let deleted: DynamicMmapFlags = DynamicMmapFlags::open(&deleted_path, populate)?;
```
timvisee, line 295: Populating memory is of course not the same as forcibly keeping in memory. Is just populating indeed good enough?

timvisee (approved):
Other than the two review suggestions below - all good!

lib/segment/src/vector_storage/dense/appendable_dense_vector_storage.rs
generall: What I know is that forcing in-ram causing many deployments to go OOM

generall: Also previous behavior didn't even try to populate  `DynamicMmapFlags`

PR #6152, opened 2025-03-11
Write rate limit batch update per point count

agourlay:
This PR makes sure all update operation are priced according to the content of their content.

It leverages an existing infrastructure for computing the cost of updates from https://github.com/qdrant/qdrant/pull/5729.

### DIFF
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -10,7 +10,7 @@ use super::{ReplicaSetState, ReplicaState, ShardReplicaSet, clock_set};
 use crate::operations::point_ops::WriteOrdering;
 use crate::operations::types::{CollectionError, CollectionResult, UpdateResult, UpdateStatus};
 use crate::operations::{ClockTag, CollectionUpdateOperations, OperationWithClockTag};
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/replica_set/update.rs
```diff
@@ -54,8 +54,7 @@ impl ShardReplicaSet {
         let result = match state {
             ReplicaState::Active => {
                 // Rate limit update operations on Active replica
-                // TODO(ratelimits) determine cost of update based on operation
```
agourlay, line 57: used to be always 1 token per update

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -296,19 +295,11 @@ impl ShardReplicaSet {
 
                 if self.peer_is_active(this_peer_id) {
                     // Check write rate limiter before proceeding if replica active
-                    // TODO(ratelimits) determine cost of update based on operation
-
-                    self.check_write_rate_limiter_lazy(&hw_measurement_acc, || {
```
agourlay, line 301: this path already had the fix in place, just needed to refactor to share it

JojiiOfficial (approved):

Issue #6157, opened 2025-03-12
Tracking issue: mutable ID tracker without RocksDB


timvisee:
This issue tracks the progress of implementing a mutable ID tracker without backing RocksDB.

### Tasks
- [x] https://github.com/qdrant/qdrant/pull/6150
- [x] https://github.com/qdrant/qdrant/pull/6158
- [x] https://github.com/qdrant/qdrant/pull/6166
- [x] https://github.com/qdrant/qdrant/pull/6169
- [x] https://github.com/qdrant/qdrant/pull/6173
- [x] https://github.com/qdrant/qdrant/pull/6174
- [x] https://github.com/qdrant/qdrant/pull/6203
- [x] https://github.com/qdrant/qdrant/pull/6207
- [x] https://github.com/qdrant/qdrant/pull/6211
- [x] https://github.com/qdrant/qdrant/pull/6267
- [x] https://github.com/qdrant/qdrant/pull/6268

### Backlog
- [x] improve storage format (possibly use format from immutable ID tracker)
- [x] more resilient against flush failures (don't crash if partially flushed)
- [x] store internal point versions as map, not as appendable list
- [x] truncate mapping changes file if partially persisted
- [x] rename files on disk to not collide with other ID tracker
- [x] integrate, use as default

timvisee:
Released in [Qdrant 1.14](https://github.com/qdrant/qdrant/releases/tag/v1.14.0)

PR #6158, opened 2025-03-12
Add mutable ID tracker tests

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/6150>

Add mutable ID tracker tests, ported from [immutable ID tracker](https://github.com/qdrant/qdrant/blob/6a52a8728d9e778c2ba4737368b83ae2c5646ee0/lib/segment/src/id_tracker/immutable_id_tracker.rs#L512).

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/6150>
- [x] Rebase on `dev`
- [x] Undraft

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/id_tracker/point_mappings.rs
+++ b/lib/segment/src/id_tracker/point_mappings.rs
@@ -103,6 +103,11 @@ impl PointMappings {
             PointIdType::Uuid(uuid) => self.external_to_internal_uuid.remove(&uuid),
         };
 
+        // Also reset inverse mapping
+        if let Some(internal_id) = internal_id {
+            self.internal_to_external[internal_id as usize] = PointIdType::NumId(u64::MAX);
+        }
+
         if let Some(internal_id) = &internal_id {
             self.deleted.set(*internal_id as usize, true);
         }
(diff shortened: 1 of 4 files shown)


lib/segment/src/id_tracker/point_mappings.rs
```diff
@@ -103,6 +103,11 @@ impl PointMappings {
             PointIdType::Uuid(uuid) => self.external_to_internal_uuid.remove(&uuid),
         };
 
+        // Also reset inverse mapping
+        if let Some(internal_id) = internal_id {
+            self.internal_to_external[internal_id as usize] = PointIdType::NumId(u64::MAX);
+        }
```
timvisee, line 109: Also resetting the inverse mapping is not strictly necessary. However, I think it's nice to do because that'll put the mapping in exactly the same state as before inserting the point.

Adding this here makes the new test `test_point_deletion_persists_reload()` happy. The alternative would be to not reset the inverse mapping and relax the test a bit to not assert raw values.

JojiiOfficial (approved):

agourlay (approved):

PR #6159, opened 2025-03-12
Write rate limit cost always lazy

agourlay:
There are two reasons for unifying the implementation:
- the function with eager cost evaluation is never used directly
- the lazy function duplicates the guarding logic

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -802,7 +802,7 @@ impl ShardReplicaSet {
         cost_fn: F,
     ) -> CollectionResult<()>
     where
-        F: Fn() -> usize,
+        F: FnOnce() -> usize,
     {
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -792,19 +792,24 @@ impl ShardReplicaSet {
...
-    fn check_write_rate_limiter(
+    fn check_write_rate_limiter<F>(
         &self,
-        cost: usize,
         hw_measurement_acc: &HwMeasurementAcc,
-    ) -> CollectionResult<()> {
+        cost_fn: F,
+    ) -> CollectionResult<()>
+    where
+        F: Fn() -> usize,
```
JojiiOfficial, line 805: A small nit but why not make this FnOnce?

JojiiOfficial (approved):

PR #6160, opened 2025-03-12
Add Anonymize macro

xzfc:
This PR introduces a derive macro for the `Anonymize` trait and replaces most of the hand-written implementations with it.

No change in the logic/behavior are made.

# Showcase

So, instead of having this:
```rust
#[derive(Serialize, Clone, Debug, JsonSchema)]
pub struct SegmentTelemetry {
    pub info: SegmentInfo,
    pub config: SegmentConfig,
    pub vector_index_searches: Vec<VectorIndexSearchesTelemetry>,
    pub payload_field_indices: Vec<PayloadIndexTelemetry>,
}

impl Anonymize for SegmentTelemetry {
    fn anonymize(&self) -> Self {
        Self {
            info: self.info.anonymize(),
            config: self.config.anonymize(),
            vector_index_searches: self.vector_index_searches.anonymize(),
            payload_field_indices: self.payload_field_indices.anonymize(),
        }
    }
}
```

We will have this:
```rust
//                                            ↓↓↓↓↓↓↓↓↓
#[derive(Serialize, Clone, Debug, JsonSchema, Anonymize)]
pub struct SegmentTelemetry {
    pub info: SegmentInfo,
    pub config: SegmentConfig,
    pub vector_index_searches: Vec<VectorIndexSearchesTelemetry>,
    pub payload_field_indices: Vec<PayloadIndexTelemetry>,
}
```

# No-op implementations

This PR adds a bunch of no-op `Anonymize` implementations in places where `.clone()` was used before. See #6115 description for a rationale.

E.g., previously:
```rust
pub struct SparseVectorParams {
    pub index: Option<SparseIndexParams>,
    pub modifier: Option<Modifier>,
}
pub enum Modifier {
    None,
    Idf,
}

impl Anonymize for SparseVectorParams {
    fn anonymize(&self) -> Self {
        Self {
            index: self.index.anonymize(),
            modifier: self.modifier.clone(), // clone
        }
    }
}
```

After this PR:
```rust
#[derive(Anonymize)] // added
pub struct SparseVectorParams {
    pub index: Option<SparseIndexParams>,
    pub modifier: Option<Modifier>,
}
#[derive(Anonymize)] // added
pub enum Modifier {
    None,
    Idf,
}

// Generated code
impl Anonymize for SparseVectorParams {
    fn anonymize(&self) -> Self {
        Self {
            index: self.index.anonymize(),
            modifier: self.modifier.anonymize(), // a call to no-op anonymize implementation
        }
    }
}
// Generated code (no-op)
impl Anonymize for Modifier {
    fn anonymize(&self) -> Self {
        match self {
            Self::None => Self::None,
            Self::Idf => Self::Idf,
        }
    }
}
```


timvisee (approved):
Nice! I like hit.

But I wonder what @generall thinks of it :smiley:

src/common/telemetry_ops/cluster_telemetry.rs
```diff
@@ -46,28 +47,35 @@ impl From<&Settings> for ClusterConfigTelemetry {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     pub status: Option<ClusterStatusTelemetry>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub config: Option<ClusterConfigTelemetry>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[anonymize(false)]
     pub peers: Option<HashMap<PeerId, PeerInfo>>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[anonymize(false)]
     pub metadata: Option<HashMap<String, serde_json::Value>>,
```
generall, line 79: I think we should anonymize this as well.

timvisee: Good catch. Using `#[anonymize(value = None)]` is probably good enough to not include it at all.

generall: I would still prefer to see if there is <something> in metadata, might be indication of a problem

PR #6161, opened 2025-03-12
Refactor `GraphLayersBuilder::link_new_point`

xzfc:
The `GraphLayersBuilder::link_new_point` method is 156 lines, making it inconvenient to work with. This PR splits it into a set of smaller methods:
- `pub fn link_new_point()`
- `fn link_new_point_on_level()`
- `fn link_with_heuristic()`
- `fn link_without_heuristic()`


timvisee (approved):
Am I right that this does not have any logic changes at all? In that case I'm happy to approve :+1:

IvanPleshkov (approved):

PR #6166, opened 2025-03-13
Implement byte storage format for mutable ID tracker mappings

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/6158>

This implements a custom byte-based storage format for persisting mutable ID tracker mappings. It also improves how we handle reading/writing/loading/storing these mappings in general.

Note that this does not really touch point version storage yet. That'll be done in the following PR.

The following comment in code describes the storage format:

```rust
/// Each change entry has a variable size. We first write a 1-byte header to define the change
/// type. The change type implies how long the entry is.
///
/// Insertion changes are serialized as follows:
///
/// +-----------------------+-----------------------+------------------+
/// | MappingChangeType: u8 | Number/UUID: u64/u128 | Internal ID: u32 |
/// +-----------------------+-----------------------+------------------+
///
/// Deletion changes are serialized as follows:
///
/// +-----------------------+-----------------------+
/// | MappingChangeType: u8 | Number/UUID: u64/u128 |
/// +-----------------------+-----------------------+
```

This includes tests to ensure serializing and deserialization is sound, even in case of corrupted data.

The reader ignores only the very last mapping in the file if it's not fully written. This way we can still recover with the WAL on start. If any other entries are corrupted an error is returned.

### Tasks
- [ ] Merge <https://github.com/qdrant/qdrant/pull/6158>
- [ ] Rebase on `dev`
- [x] Undraft

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -368,20 +328,94 @@ fn versions_path(segment_path: &Path) -> PathBuf {
     segment_path.join(FILE_VERSIONS)
 }
 
-fn load_mappings(
-    mappings_path: &Path,
-    deleted: &mut BitVec,
-    internal_to_external: &mut Vec<PointIdType>,
-    external_to_internal_num: &mut BTreeMap<u64, PointOffsetType>,
-    external_to_internal_uuid: &mut BTreeMap<Uuid, PointOffsetType>,
+/// Store new mapping changes, appending them to the given file
```
timvisee, line 331: The meat for all store/load logic begins here.

It's all separated in functions like:
- store (append) mappings to file
- write mappings to writer
- write single mapping to writer
- load mappings from file
- read mappings from reader
- read single mapping from reader

JojiiOfficial (approved):

ffuugoo (approved):
LGTM, though, why not use `serde_cbor` for serialization instead of hand-rolled format? Would be a bit less efficient, but easier to maintain, and we already use CBOR in number of places. 🤔

timvisee:
> LGTM, though, why not use `serde_cbor` for serialization instead of hand-rolled format? Would be a bit less efficient, but easier to maintain, and we already use CBOR in number of places. 🤔

I didn't look at CBOR specifically, but did look at bincode.

But when I look at CBOR now, I can list a few arguments on why I'd still prefer the current approach:

- CBOR does not support u128 to store a UUID
- with CBOR, we'd need to know the length of each item to read which would require a length header prefix
- this custom implementation is still relatively simple (imo)
- this custom implementation is smaller

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -368,20 +328,94 @@ fn versions_path(segment_path: &Path) -> PathBuf {
...
+/// An error item is returned if reading a mapping change fails due to malformed data. Then the
+/// iterator will not produce any more items.
+fn read_mappings_iter<R: Read>(
+    mut reader: R,
+) -> impl FusedIterator<Item = OperationResult<MappingChange>> {
+    // Keep reading until end of file or error
+    std::iter::from_fn(move || match read_entry(&mut reader) {
+        Ok(entry) => Some(Ok(entry)),
+        // Done reading if end of file is reached
+        Err(err) if err.kind() == io::ErrorKind::UnexpectedEof => None,
```
generall, line 400: What would happen if we have a corrupted file, in which we ignored last entry error, in case we continue to write new records into it?

timvisee: Thanks! That would indeed be a problem. I had it on my task list, but forgot it in this PR.

I'll try to add some logic to truncate it.

timvisee: I'll handle this in a separate PR, as it isn't trivial to do and I don't want to make this PR more complicated. I've added it in the [tracking issue](https://github.com/qdrant/qdrant/issues/6157).

timvisee:
Merging now. The review comments are handled in a follow up PR.

PR #6169, opened 2025-03-14
Mutable ID tracker: implement storage for point versions

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/6166>

Implement the new storage for point versions.

This changes the mapping of versions from external ID to version to internal ID to version.

Points are kept in memory for fast access. On load we read the full list of versions. On flush we store we write all changed versions directly to the versions file.

Writing changes is optimized to properly buffer all writes. This has a great effect if writing a big batch of versions, to use a lot less intermediate syscalls.

In terms of the file format, the file on disk is just a contiguous list of version numbers similar to an array.

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/6166>
- [x] Rebase on `dev`
- [x] Undraft

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/id_tracker/mutable_id_tracker.rs
+++ b/lib/segment/src/id_tracker/mutable_id_tracker.rs
@@ -4,7 +4,6 @@ use std::io::{self, BufReader, BufWriter, Read, Seek, Write};
 use std::mem;
 use std::path::{Path, PathBuf};
 
-use ahash::AHashMap;
 use bitvec::prelude::{BitSlice, BitVec};
 use byteorder::{ReadBytesExt, WriteBytesExt};
 use common::types::PointOffsetType;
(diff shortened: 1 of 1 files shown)


ffuugoo (approved):

coszio (approved):

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -604,7 +604,12 @@ fn store_version_changes(
     })?;
 
     // Explicitly fsync file contents to ensure durability
-    let file = writer.into_inner().unwrap();
+    let file = writer.into_inner().map_err(|err| {
+        OperationError::service_error(format!(
+            "Failed to close ID tracker point versions write buffer: {}",
+            err.into_error()
+        ))
+    })?;
```
timvisee, line 612: @coszio Thanks for the review.

Though, this one is unnecessary. We've already completely flushed the writer right before it, so this cannot panic on unwrap. Sadly I need the unwrap here so that I can touch `sync_all()`.

ffuugoo: TBH, I'd prefer `?`. It's excessive, but it prevents accidental panics when someone would (intentionally or not) *remove* explicit flush, but miss your comment and unwrap.

PR #6170, opened 2025-03-14
[score boosting] fix proxy segment leakage

coszio:
The signature for `rescore_with_formula` included a deleted bitslice only because the proxy segment would introduce it to handle the write and wrapped results. 

In this PR, we remove the abstraction leak, and handle the logic to prefer write results only within the proxy segment.

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -4,8 +4,8 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
 
+use ahash::AHashMap;
 use bitvec::prelude::BitVec;
-use bitvec::slice::BitSlice;
 use common::counter::hardware_counter::HardwareCounterCell;
 use common::tar_ext;
 use common::types::{PointOffsetType, TelemetryDetail};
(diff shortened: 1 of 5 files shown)


lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -454,36 +454,44 @@ impl SegmentEntry for ProxySegment {
...
-        // Just join both results, they will be deduplicated and top-k'd later
-        write_results.append(&mut wrapped_results);
+        // Merge results
+        if self.deleted_points.read().is_empty() {
+            // Just join both results, they will be deduplicated and top-k'd later
+            wrapped_results.append(&mut write_results);
+        } else {
+            // Create a hashmap of write results by point ID for faster lookup
+            // We expect write results to be shorter than wrapped results
+            let mut write_results_map: AHashMap<_, _> = write_results
```
generall, line 480: I would prefer to avoid allocating HashMaps when possible. This is potentially an expensive operation.

generall (approved):

lib/collection/src/collection_manager/holders/proxy_segment.rs
coszio: Duh, this is much better, thanks

PR #6173, opened 2025-03-17
Mutable ID tracker: truncate mappings file if last entry is partial

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/6169>

If the last entry in the mappings file is partial, truncate the file.

It can happen if we crash during flush. We must truncate the file so that newly appended mappings don't get corrupted.

I don't particularly like the implementation, but the tools here are a bit limited. An alternative approach would require to load the full file in memory, or it would require to read over the file twice, which is why I implemented the current approach instead.

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/6169>
- [x] Rebase on `dev`
- [x] Undraft

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/id_tracker/mutable_id_tracker.rs
+++ b/lib/segment/src/id_tracker/mutable_id_tracker.rs
@@ -366,13 +366,37 @@ fn write_mapping_changes<W: Write>(
 }
 
 /// Load point mappings from the given file
+///
+/// If the file ends with an incomplete entry, it is truncated from the file.
 fn load_mappings(mappings_path: &Path) -> OperationResult<PointMappings> {
     let file = File::open(mappings_path)?;
+    let file_len = file.metadata()?.len();
     let reader = BufReader::new(file);
-    read_mappings(reader)
+
+    let (mappings, bytes_read) = read_mappings(reader)?;
+
+    // Truncate file if it ends with an incomplete entry
+    // It can happen on crash while flushing. We must truncate the file here to not corrupt new
+    // entries we append to the file
+    if bytes_read < file_len {
+        log::warn!(
+            "Mutable ID tracker mappings file ends with incomplete entry, removing last {} bytes and assuming WAL recovery",
+            file_len - bytes_read
+        );
+        let file = File::options()
+            .write(true)
+            .truncate(false)
+            .open(mappings_path)?;
+        file.set_len(bytes_read)?;
+        file.sync_all()?;
+    }
+
+    Ok(mappings)
 }
 
 /// Iterate over mapping changes from the given reader
+///
+/// Each non-errorous item is a tuple of the mapping change and the number of bytes read so far.
 ///
 /// The iterator ends when the end of the file is reached, or when an error occurred.
 ///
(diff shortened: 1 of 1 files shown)


lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -382,10 +406,15 @@ fn load_mappings(mappings_path: &Path) -> OperationResult<PointMappings> {
...
+
     // Keep reading until end of file or error
     std::iter::from_fn(move || match read_entry(&mut reader) {
-        Ok(entry) => Some(Ok(entry)),
+        Ok((entry, read_bytes)) => {
+            position += read_bytes;
+            Some(Ok((entry, position)))
+        }
         // Done reading if end of file is reached
         Err(err) if err.kind() == io::ErrorKind::UnexpectedEof => None,
```
ffuugoo, line 419: Why not do something like this:

```rust
let mut position = reader.stream_position()?;

std::iter::from_fn(move || match read_entry(&mut reader) {
	Ok((entry, read_bytes)) => {
		position += read_bytes;
		Some(Ok(entry))
	}
	// Done reading if end of file is reached
	Err(err) if err.kind() == io::ErrorKind::UnexpectedEof => {
		reader.seek(SeekFrom::Start(position))?;
		None
	}
	Err(err) => Some(Err(err.into())),
}
```

It will require a few more syscalls, but one extra when starting reading a file and a few extras when handling an error should be acceptable, IMO.

timvisee: Fine suggestion. I like that it keeps the return type a bit more clean. Thanks.

timvisee: Not necessary, I can simply check if we've exhausted the reader or not.

timvisee: @ffuugoo done in <https://github.com/qdrant/qdrant/pull/6173/commits/40c187d1ce2710eaa0b0d4b5e3cc7adf225c79c1>

ffuugoo (approved):

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -502,22 +549,28 @@ fn read_entry<R: Read>(reader: &mut R) -> io::Result<MappingChange> {
...
+            ))
         }
         MappingChangeType::InsertUuid => {
             let external_id =
                 PointIdType::Uuid(Uuid::from_u128_le(reader.read_u128::<FileEndianess>()?));
             let internal_id = reader.read_u32::<FileEndianess>()? as PointOffsetType;
-            Ok(MappingChange::Insert(external_id, internal_id))
+            Ok((
+                MappingChange::Insert(external_id, internal_id),
+                (8 + 128 + 32) / 8,
```
generall, line 563: Let's use named constants

generall (approved):

lib/segment/src/id_tracker/mutable_id_tracker.rs
timvisee: I like your addition. Thanks!

I've added a test to ensure the operation size we return matches the number of bytes read from a stream.

PR #6174, opened 2025-03-17
Mutable ID tracker integration

timvisee:
Depends on: <https://github.com/qdrant/qdrant/pull/6173>

Add support for the new mutable ID tracker.

It is currently locked behind a feature flag, only enabled in development builds. In the next Qdrant release the new tracker will enabled globally.

- Feature flag disabled _(default in production build)_
  RocksDB based tracker is used by default, unless new mappings are stored on disk
- Feature flag enabled _(default in development build)_
  New tracker is used by default, unless there are RocksDB mappings on disk

Set `QDRANT__FEATURE_FLAGS__USE_MUTABLE_ID_TRACKER_WITHOUT_ROCKSDB=true` to prematurely enable the new ID tracker.

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/6173>
- [x] Rebase on `dev`
- [x] Undraft

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/config/development.yaml
+++ b/config/development.yaml
@@ -2,6 +2,7 @@ log_level: DEBUG
 
 feature_flags:
   use_new_shard_key_mapping_format: false
+  use_mutable_id_tracker_without_rocksdb: true
 
 inference:
   address: "http://localhost:2114/api/v1/infer"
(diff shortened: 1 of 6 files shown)


generall (approved):
Did some manual tests to check that id_tracked doesn't grow indefinitelly:

- upsert points in a loop with overwrite
- create and delete the same point 1k times

In both cases files either don't grow (because we re-use same internal id), or optimizer kicks-in and vacuums the segment.
So works as expected.

One unpleasant side-effect: if new id tracker is recovered in old version, which doesn't know about it, it will be treated as if there are no points in segment, rather than a panic. (might be possible to address in separate PR)

PR #6190, opened 2025-03-18
Writeback counter cell

JojiiOfficial:
Introduces a new type that counts inside a `usize` and writes the results into a `CounterCell` on drop.
This reduces potential errors when manually counting inside a hotpath and applying measurements afterwards


coszio (approved):

agourlay (approved):

lib/common/common/src/counter/counter_cell.rs
```diff
@@ -83,4 +83,73 @@ impl CounterCell {
...
+    #[test]
+    fn test_write_back_counter() {
+        let cell = CounterCell::new();
+
+        {
+            let mut wb_counter = cell.write_back_counter();
+            wb_counter.incr_delta(4);
+        }
+
+        assert_eq!(cell.get(), 4)
```
timvisee, line 153: How do you feel about adding this assertion?

It's not necessary, but it would prove the type behaves like we tell it does.

```suggestion
        {
            let mut wb_counter = cell.write_back_counter();
            wb_counter.incr_delta(4);
            assert_eq!(cell.get(), 0);
        }

        assert_eq!(cell.get(), 4);
```

timvisee (approved):

generall (approved):

PR #6191, opened 2025-03-18
fix loading of on-disk payload indexes

generall:


agourlay (approved):
Tested locally :heavy_check_mark:

PR #6193, opened 2025-03-18
Debug assert a missing column family or not loaded  payload index

agourlay:
We should never enter this condition during development as it could actually hide a different problem https://github.com/qdrant/qdrant/pull/6191

### DIFF
--- a/lib/segment/src/index/struct_payload_index.rs
+++ b/lib/segment/src/index/struct_payload_index.rs
@@ -145,6 +145,7 @@ impl StructPayloadIndex {
         }
         if !is_loaded {
             debug!("Index for `{field}` was not loaded. Building...");
+            debug_assert!(false, "Index should not need to be loaded during testing");
             // todo(ivan): decide what to do with indexes, which were not loaded
             indexes = self.build_field_indexes(
                 field,
@@ -754,6 +755,10 @@ impl PayloadIndex for StructPayloadIndex {
                         // Still we want to log this event, for potential debugging.
                         log::warn!(
                             "Flush: RocksDB cf_handle error: Cannot find column family {name}. Assume index is removed.",
+                        );
+                        debug_assert!(
+                            false,
+                            "Missing column family should not happen during testing"
                         );
                     }
                     Err(err) => {
--- a/tests/storage-compat/storage-compatibility.sh
+++ b/tests/storage-compat/storage-compatibility.sh
@@ -126,9 +126,3 @@ test_version $PREV_PATCH_QDRANT_VERSION
 
 # Test previous minor version
 test_version $PREV_MINOR_QDRANT_VERSION
-
-# Test gridstore storage generated manually with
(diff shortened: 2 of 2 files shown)


generall:
I would also include debug assert here: https://github.com/qdrant/qdrant/blob/1fff66693928fc4fff546b48f50dccddc98c1365/lib/segment/src/index/struct_payload_index.rs#L145

agourlay:
The assertions make the storage compat. test is failing for a boolean payload which is on_disk.
When digging deeper, it seems that the column family for the `SimpleBoolIndex` is not found.

agourlay:
There is actually a warning when loading the segment.

```
Segment has incorrect payload index for boolean_field, recreating it now (current: FieldParams(Bool(BoolIndexParams { type: Bool, on_disk: Some(true) })), configured: FieldParams(Bool(BoolIndexParams { type: Bool, on_disk: None })))
```

timvisee:
> There is actually a warning when loading the segment.
> 
> ```
> Segment has incorrect payload index for boolean_field, recreating it now (current: FieldParams(Bool(BoolIndexParams { type: Bool, on_disk: Some(true) })), configured: FieldParams(Bool(BoolIndexParams { type: Bool, on_disk: None })))
> ```

I did see this under semi-normal behavior - when restarting a peer that is currently in the progress of creating a new index.

agourlay:
It seems the `boolean_field` column family is really missing from the rocksdb data in the snapshot.

```
Existing column families: [
"default", 
"integer_field_map", 
"integer_field_numeric", 
"float_field_numeric",
"text_field_fts", 
"keyword_field_map",
"geo_field_geo"
] for path "./storage/collections/test_collection_mmap_field_index/0/segments/94131989-8c06-4ebc-9837-10cd9094800b/payload_index"
```

agourlay:
The existing storage archive use in the compatibility tests are causing too much friction here.
I will revisit in `v1.14.0` once all problematic archives are purged.

agourlay:
Reopening now that the old archives are gone https://github.com/qdrant/qdrant/pull/6422

agourlay:
Rebasing and removing the old gridstore test snapshot made that branch green :green_circle:

coszio:
Although I agree the custom gridstore compat test is not necessary now, I want to point out that we don't have a test which asserts that we can still open a rocksdb-based payload storage.

If a user had its data from pre-gridstore (or any other new structure for that matter), there is nothing which guarantees that upgrading minor by minor will migrate all their segments to use the newer structures. We are only relying on the optimizers to do the migration when the segment needs optimization.

So, I suggest we either:
- force migration from legacy structures to new ones once there is some amount of versions' difference.
- keep trying to open older compat snapshots, even if we need to open them version-by-version.

I think the latter option is easier for us, and is consistent with our backward compatibility policy. For instance, we could take the snapshot from 1.12, and first open it with 1.13, then with 1.14, then with PR, and assert that there are no errors

coszio (approved):

timvisee (approved):

PR #6194, opened 2025-03-18
Add global feature flags

timvisee:
First iteration of a structure for configuring feature flags.

This allows us to explicitly set some flags during testing to enable new and upcoming features. It helps us test the features before releasing them, without having to change source code. I'd like to unify the interface for this. It's intentionally kept very simple.

For now this includes just one flag. I'd like to add a flag for using the [new immutable ID tracker](https://github.com/qdrant/qdrant/pull/6174) once merged.

In debug builds, all flags are enabled by default (as per `config/development.yaml`). In chaos testing we'll probably enable all features by setting the appropriate environment variables.

I chose to use this global structure because it turns out to be extremely cumbersome to propagate the flags to all places we might use it. I want to prevent having to propagate it (adding a lot of changes), having to undo all changes one or two  versions later when we remove the flag. While I normally hate a global/static structure, I think this is a fair exception.

Let's use this PR as a place to discuss this feature.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/config/development.yaml
+++ b/config/development.yaml
@@ -1,5 +1,8 @@
 log_level: DEBUG
 
+feature_flags:
+  use_new_shard_key_mapping_format: true
+
 inference:
   address: "http://localhost:2114/api/v1/infer"
   timeout: 10
--- a/config/development.yaml
+++ b/config/development.yaml
@@ -1,7 +1,7 @@
 log_level: DEBUG
 
 feature_flags:
-  use_new_shard_key_mapping_format: true
+  use_new_shard_key_mapping_format: false
 
 inference:
   address: "http://localhost:2114/api/v1/infer"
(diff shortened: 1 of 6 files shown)


coszio (approved):
Other than figuring out if the CI error is related, I agree this is a good way to:
- keep temporary flags in one place
- not having to propagate them in the code
- being able to change them without recompiling

JojiiOfficial (approved):
I like this idea!
I created some minor suggestions in #6198 but they're not necessary.
✅

config/development.yaml
```diff
@@ -1,5 +1,8 @@
 log_level: DEBUG
 
+feature_flags:
+  use_new_shard_key_mapping_format: false
+
```
timvisee, line 5: This idea immediately paid off, enabling the new shard key mapping logic turns out to be problematic with recent changes. A few consensus tests fail and it's not immediately clear to me what the problem is.

I'll disable it for now, and resolve the problems in a separate PR.

We can merge this one :+1:

generall (approved):

PR #6196, opened 2025-03-18
[score boosting] support datetime expressions

coszio:
Adds ` { "datetime": <datetime constant> }` and `{ "datetime_key": <payload key> }`  expressions which can be used to parse strings into f64's ~microseconds~ seconds.

A necessary additional change is to switch expression evaluation type from f32 to f64, so that we don't lose precision on timestamp microseconds right away, it will still be converted to f32 but only after finishing the evaluation. This is specially important since timestamp will be most certainly used within a decay function (#6154) to compress its range.


lib/segment/src/types.rs
```diff
@@ -92,7 +90,46 @@ impl<'de> Deserialize<'de> for DateTimeWrapper {
     }
 }
 
-impl From<chrono::DateTime<chrono::Utc>> for DateTimeWrapper {
+impl FromStr for DateTimePayloadType {
```
coszio, line 93: this one was moved closer to the type definition, but no changes were made

lib/segment/src/index/query_optimization/rescore_formula/parsed_formula.rs
```diff
@@ -76,11 +77,30 @@ impl VariableId {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // try as date
+        DateTimePayloadType::from_str(s)
+            .map(DateTimeExpression::Constant)
+            // else as payload key
+            .or_else(|_| JsonPath::from_str(s).map(DateTimeExpression::PayloadVariable))
+            .map_err(|_| format!("Invalid date time expression: {s}"))
+    }
+}
```
coszio, line 97: Since the expression is a string, it will try to parse as datetime first, then as a payload key

coszio:
Rebased on top of dev, and included fixes for `PreciseScore` in decay functions

timvisee:
> f64's microseconds

Why not seconds? It's a float anyhow. I usually don't like it if we diverge from the SI base units.

lib/collection/src/operations/universal_query/shard_query.rs
```diff
@@ -397,6 +397,7 @@ impl TryFrom<grpc::Expression> for ExpressionInternal {
                     .map_err(|_| tonic::Status::invalid_argument("invalid payload key"))?;
                 ExpressionInternal::GeoDistance { origin, to }
             }
+            Variant::Datetime(dt_str) => ExpressionInternal::DateTime(dt_str),
```
timvisee, line 400: It looks like we have two different formats: `DateTime` and `Datetime`

I prefer the first everywhere if possible, as it's much more commonly used.

lib/segment/src/index/query_optimization/rescore_formula/parsed_formula.rs
timvisee: It is not possible to properly distinguish between the two types, right? Not a fan of this, since it's definitely possible for a user to use date time strings as keys.

Maybe like this:

```json
"history": {
  "2025-03-24": 128,
  "2025-03-23": 110,
  "2025-03-22": 283,
  "2025-03-21": 10,
  "2025-03-20": 123,
  "2025-03-19": 28,
  "2025-03-18": 387
}
```

coszio: That is a good point, I didn't think of the case where datetime strings are used as keys, but it does make sense

lib/collection/src/operations/universal_query/shard_query.rs
coszio: At the same time, we also have `DatetimeRange` and `FieldTypeDatetime` in grpc, and their counterparts in REST. This is why I stuck with the single word representation. 

I can homogenize into `Datetime` in the internal variants too.

lib/segment/src/index/query_optimization/rescore_formula/parsed_formula.rs
coszio: Solved by introducing a separate expression: `{ "datetime_key": <payload key> }`

timvisee: Thanks. That works!

coszio:
Just updated the description, the PRs uses datetime as float seconds

timvisee (approved):

PR #6198, opened 2025-03-19
Suggestions for global feature flags

JojiiOfficial:
Two minor suggestions for global feature flags in #6194:

- Private static Flags to have better control of initialization
- Public function for accessing global flags with fallback to default values to prevent unwrapping everywhere

This is just a suggestion and I'm happy to discuss or discard if not needed!


timvisee (approved):
Good suggestion. Thank you!

PR #6199, opened 2025-03-19
Flush after moving snapshots

KShivendu:
In chaos testing, we found nodes be panicing when force killed during (streaming) snapshot transfers. We found that some of the files sometimes go missing from `/qdrant/storage/collections/<collection-name>/<shard-id>/segments/<segment-id>/vector_storage-<vector-name>` (happened with both dense + sparse vectors). This includes:
- `vectors/chunk_*.mmap`
- `deleted/{status,flags_a}.dat`
- `quantized.config.json`

These all files are however clearly present if we unarchive the received snapshot file found in `/qdrant/snapshots/tmp/<snapshot-name>.download`. This means that the streaming download is working fine. After that we restructure the files in place in `/qdrant/storage/tmp` directory. And then we finally move to shard directory at `/qdrant/storage/collections/<collection-name>/<shard-id>`. We suspect that the last move operation if terminated in between will lead to lost files because of not flushing to disk. So this PR adds flushing after that step.

<details>
<summary>
More details
</summary>

We found this in logs just before the node was killed:
```
2025-03-07T13:40:43.324481Z  INFO segment::segment::segment_ops: Snapshot format: Streamable
2025-03-07T13:40:43.396846Z  INFO segment::segment::segment_ops: Snapshot format: Streamable
2025-03-07T13:40:43.427133Z DEBUG collection::update_handler: Stopping flush worker.
```

While I could only see one of the segments in the `/qdrant/storage/collections/benchmark/2/segments` dir. One of the was entirely missing and another one was partially present with missing files (as described above)

This means the segment level snapshot restructing/unpacking step probably worked and persisted for both the segments. Till this step they are stored in `/qdrant/snapshots/tmp` and are moved to the `/qdrant/storage/collections/benchmark/2` which is a different network disk and I suspect that it's this movement between disks that wasn't persisted properly because of kill. So flushing there is important. 
</details>

TODO:
- [ ] Manually test that flushing logic works as expected

### All Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -120,6 +120,8 @@ impl LocalShard {
 
         LocalShardClocks::move_data(from, to).await?;
 
+        std::fs::File::open(to)?.sync_all()?;
+
         Ok(())
     }
 
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -24,6 +24,7 @@ use common::rate_limiting::RateLimiter;
 use common::types::TelemetryDetail;
 use common::{panic, tar_ext};
 use indicatif::{ProgressBar, ProgressStyle};
+use io::fsync::fsync_dir_recursive;
 use itertools::Itertools;
 use parking_lot::{Mutex as ParkingMutex, RwLock};
 use segment::data_types::vectors::VectorElementType;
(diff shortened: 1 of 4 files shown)


KShivendu:
We agreed that this is probably not the root cause because we have the shard initializing flag file. 

Instead we found that, we move shards to `storage` in two different places and only one of those place have the flag initializing logic. So maybe the other flow without the flag is being called in chaos testing workflow and that causes missing files. I'll update that 2nd function instead and how it goes in chaos testing.

PR #6208, opened 2025-03-20
dont anonymize shard transfers

generall:


lib/collection/src/operations/types.rs
```diff
@@ -345,43 +345,53 @@ pub struct CollectionClusterInfo {
...
     pub sync: bool,
 
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[anonymize(false)]
     pub method: Option<ShardTransferMethod>,
 
     /// A human-readable report of the transfer progress. Available only on the source peer.
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[anonymize(false)]
     pub comment: Option<String>,
```
timvisee, line 380: Sure about this one? I'm not sure if we rely on it anywhere else where we also anonymize.

lib/collection/src/shards/transfer/mod.rs
```diff
@@ -107,7 +108,9 @@ impl ShardTransferKey {
 }
 
 /// Methods for transferring a shard from one node to another.
-#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Default, Serialize, Deserialize, JsonSchema)]
+#[derive(
+    Debug, Clone, Copy, Hash, PartialEq, Eq, Default, Serialize, Deserialize, JsonSchema, Anonymize,
+)]
```
timvisee, line 113: We skip anonymization for this type so we can remove the derive.

```suggestion
#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Default, Serialize, Deserialize, JsonSchema)]
```

lib/collection/src/shards/transfer/mod.rs
```diff
@@ -3,6 +3,7 @@ use std::time::Duration;
 use async_trait::async_trait;
 use common::defaults::{self, CONSENSUS_CONFIRM_RETRIES};
 use schemars::JsonSchema;
+use segment::common::anonymize::Anonymize;
```
timvisee, line 6:
```suggestion
```

PR #6210, opened 2025-03-20
Fix consensus snapshot not applying resharding state

timvisee:
The resharding state in a consensus snapshot was simply not applied. This fixes the issue.

We've been seeing a wild range of resharding state errors in chaos testing, which we've yet to have an explanation for. I expect this to resolve all those problems.

A proper test for this will be done separately, as this is part of a wider effort fixing issues.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection/state_management.rs
+++ b/lib/collection/src/collection/state_management.rs
@@ -6,6 +6,7 @@ use crate::collection_state::{ShardInfo, State};
 use crate::config::CollectionConfigInternal;
 use crate::operations::types::{CollectionError, CollectionResult};
 use crate::shards::replica_set::ShardReplicaSet;
+use crate::shards::resharding::ReshardState;
 use crate::shards::shard::{PeerId, ShardId};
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection/state_management.rs
```diff
@@ -29,12 +30,22 @@ impl Collection {
         this_peer_id: PeerId,
         abort_transfer: impl FnMut(ShardTransfer),
     ) -> CollectionResult<()> {
-        self.apply_config(state.config).await?;
-        self.apply_shard_transfers(state.transfers, this_peer_id, abort_transfer)
+        let State {
+            config,
+            shards,
+            resharding,
```
timvisee, line 36: We totally missed handling `resharding` here. Struct destruction reveals the unused field.

KShivendu (approved):
Thanks! I **almost** came to a similar conclusion with logs and had just started diving in to the code and you already found it! Next time I will be faster! :smile:

ffuugoo (approved):

PR #6214, opened 2025-03-21
(Potentially) Fix missing CF on flush on payload index change

generall:


lib/segment/src/common/rocksdb_wrapper.rs
```diff
@@ -220,15 +220,9 @@ impl DatabaseColumnWrapper {
...
             let Some(column_family) = db.cf_handle(&column_name) else {
-                // It is possible, that the index was removed during the flush by user or another thread.
-                // In this case, non-existing column family is not an error, but an expected behavior.
-
-                // Still we want to log this event, for potential debugging.
-                log::warn!(
-                    "Flush: RocksDB cf_handle error: Cannot find column family {}. Ignoring",
-                    &column_name
-                );
-                return Ok(()); // ignore error
```
generall, line 231: We don't want to ignore it here for 2 reasons:

- `rocksdb_wrapper` is used in other structures as well, not only in payload index. Some of them can't be removed, so ignoring error is not correct.
- There are other functions in payload index flush (e.g. `remove` in `DatabaseColumnScheduledDeleteWrapper`) which are trying to read column family. They can fail in the same way, but previously we didn't ignore those other errors. 


Instead, we not handle this type error in the segment level flush for the whole `payload_index` structure, and only for `payload_index`

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -502,6 +502,13 @@ pub trait SegmentOptimizer {
 
         let mut optimized_segment: Segment = segment_builder.build(indexing_permit, stopped)?;
 
+        // Delete points
+        let deleted_points_snapshot = proxy_deleted_points
+            .read()
+            .iter()
+            .map(|(point_id, versions)| (*point_id, *versions))
+            .collect::<Vec<_>>();
+
```
timvisee, line 511: What is the reason for moving this up? It does not have any affect on business logic does it?

timvisee (approved):

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
generall: I had suspicions I could not confirm. I propose to keep it as in the worst case it just doesn't change anything

PR #6220, opened 2025-03-21
Simplify `ShardKeyMapping` implementation

ffuugoo:
This PR simplifies new `ShardKeyMapping` format support introduced in #6209.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/snapshots.rs
+++ b/lib/collection/src/collection/snapshots.rs
@@ -184,7 +184,6 @@ impl Collection {
                 if !mapping_path.exists() {
                     Vec::new()
                 } else {
-                    // Use wrapper type to support both formats
                     let shard_key_mapping: ShardKeyMapping = read_json(&mapping_path)?;
                     shard_key_mapping.shard_ids()
                 }
--- a/lib/collection/src/collection/snapshots.rs
+++ b/lib/collection/src/collection/snapshots.rs
@@ -23,7 +23,7 @@ use crate::shards::remote_shard::RemoteShard;
 use crate::shards::replica_set::ShardReplicaSet;
 use crate::shards::shard::{PeerId, ShardId};
 use crate::shards::shard_config::{self, ShardConfig};
-use crate::shards::shard_holder::shard_mapping::ShardKeyMappingWrapper;
+use crate::shards::shard_holder::shard_mapping::ShardKeyMapping;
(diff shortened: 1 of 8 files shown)


ffuugoo:
Spray and pray PR, really bet on those integration tests 🤞✅

lib/collection/src/shards/shard_holder/shard_mapping.rs
```diff
@@ -1,293 +1,121 @@
...
+    Old(HashMap<ShardKey, HashSet<ShardId>>),
 }
 
-impl From<ShardKeyMapping> for ShardKeyMappingWrapper {
+impl From<ShardKeyMapping> for SerdeHelper {
     fn from(mapping: ShardKeyMapping) -> Self {
-        // If any shard key is a number, always prefer new persisted format
-        // The old format is broken and fails to deserialize shard key numbers
-        let any_number = mapping.keys().any(|key| matches!(key, ShardKey::Number(_)));
+        if feature_flags().use_new_shard_key_mapping_format {
```
timvisee, line 103: This is incorrect. It must also pick the new format if there are any numeric values, because we lose type information for them in the old format.

ffuugoo: It will break backward compatibility with previous version. Not sure which one is better: broken shard keys or broken consensus. 😂

timvisee: Correct, but that's intended.

The [initial](https://github.com/qdrant/qdrant/pull/5838) fix for this problem landed in 1.13.1.

Having a numeric value is an opt-in for the new format. You're correct that it breaks backwards compatibility, but that's better than a cluster that does not function at all.

This PR should match the implementation of <https://github.com/qdrant/qdrant/pull/5838>. We have some customers that are actively using this.

This fix not being properly included in consensus snapshots was an oversight. But the best we can do is to bring it up to speed with the other fix, versus not handling it at all. It's the only way we can stabilize it. If we use the old format here, it remains broken in existing and upcoming releases. If we opt-in to the new format here, we at least fix new deployments.

lib/collection/src/collection/snapshots.rs
```diff
@@ -185,7 +185,7 @@ impl Collection {
                     Vec::new()
                 } else {
                     // Use wrapper type to support both formats
```
ffuugoo, line 187:
```suggestion
```

timvisee:
I consider this to be done. Thanks! :pray:

PR #6228, opened 2025-03-21
Add `all` feature flag, report enabled flags on start

timvisee:
Extends <https://github.com/qdrant/qdrant/pull/6194>

This extends feature flags a bit, adding two useful improvements:

1. add magic `all` flag that enables everything
2. if any feature flags are set, log them on start

The `all` flag is especially useful on chaos testing to always enable the feature flags we set up. This prevents us from constantly having to bump environment variables for new flags we add. It also keeps our development configuration file a bit cleaner.

The log entry for enabled feature flags is rather basic and looks like this:

```log
2025-03-21T15:53:59.394753Z DEBUG qdrant: Feature flags: FeatureFlags { all: true, use_new_shard_key_mapping_format: true, use_mutable_id_tracker_without_rocksdb: true }
```

I've again tried to keep this as simple as possible.

### All Submissions:

### New Feature Submissions:

5. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### DIFF
--- a/lib/common/common/src/flags.rs
+++ b/lib/common/common/src/flags.rs
@@ -8,7 +8,9 @@ static FEATURE_FLAGS: OnceLock<FeatureFlags> = OnceLock::new();
 #[derive(Default, Debug, Deserialize, Clone, Copy)]
 pub struct FeatureFlags {
     /// Magic feature flag that enables all features.
-    pub all: bool,
+    ///
+    /// Note that this will only be applied to all flags when passed into [`init_feature_flags`].
+    all: bool,
 
     /// Whether to use the new format to persist shard keys
     ///
--- a/lib/common/common/src/flags.rs
+++ b/lib/common/common/src/flags.rs
@@ -7,6 +7,9 @@ static FEATURE_FLAGS: OnceLock<FeatureFlags> = OnceLock::new();
 
 #[derive(Default, Debug, Deserialize, Clone, Copy)]
 pub struct FeatureFlags {
+    /// Magic feature flag that enables all features.
+    pub all: bool,
+
     /// Whether to use the new format to persist shard keys
     ///
     /// The old format fails to persist shard key numbers correctly, converting them into strings on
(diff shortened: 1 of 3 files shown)


lib/common/common/src/flags.rs
```diff
@@ -25,10 +28,34 @@ pub struct FeatureFlags {
...
-    let res = FEATURE_FLAGS.set(*flags);
+pub fn init_feature_flags(mut flags: FeatureFlags) {
+    let FeatureFlags {
+        all,
+        use_new_shard_key_mapping_format,
+        use_mutable_id_tracker_without_rocksdb,
+    } = &mut flags;
+
+    // If all is set, explicitly set all feature flags
+    if *all {
```
JojiiOfficial, line 53: Tbh I don't like this approach too much because setting `all` to `true` later (somehow) won't do anything.
But I agree that the alternative isn't any better as it would require us to add getter function for each feature that additionally checks for the `all` flag directly or indirectly.

Wdyt about adding a note to the `all` attribute hinting that later mutation won't have any affect?
I'm aware that currently it is not possible but in case we change things later, it might help if it is explicitly stated somewhere.

lib/common/common/src/flags.rs
```diff
@@ -25,10 +28,34 @@ pub struct FeatureFlags {
...
 
+impl FeatureFlags {
+    /// Check if no flag is set at all
+    pub fn is_empty(self) -> bool {
+        let FeatureFlags {
+            all,
+            use_new_shard_key_mapping_format,
+            use_mutable_id_tracker_without_rocksdb,
+        } = self;
+        !all && !use_new_shard_key_mapping_format && !use_mutable_id_tracker_without_rocksdb
```
JojiiOfficial, line 39: If we go with the "apply `all` on init" approach of this PR, I think we should remove the check for `all` here since it doesn't serve any purpose but makes things more complex and might give the wrong feeling of the `all` attribute.

JojiiOfficial (approved):
Left two comments with suggestions but all in all lgtm 👍🏻

lib/common/common/src/flags.rs
coszio: I can suggest a few solutions:
- remove `pub` visibility
- use a Deserialize shadow type which includes the `all` field. (A bit overkill)
- use getters (also too verbose, even with a macro to generate them)

JojiiOfficial: Thanks! I think it's good enough ✅

coszio (approved):

PR #6230, opened 2025-03-24
Properly use shard snapshot initializing flag while recovering

KShivendu:
We introduced shard initailizing flag to clean dirty snapshots after a failed shard snapshot recovery (for example: files missing after move/rename step in shard snapshot recovery due to node crash)

However, there are 2 functions that call `ShardHolder::recover_local_shard_from()` and we were creating/deleting this flag in only one of them which led to panics in chaos testing. This PR makes Qdrant use the snapshot flag in both places by making it a part of `ShardHolder::recover_local_shard_from` so Qdrant cleans up the dirty shard on its own.

More context of this can be found in https://github.com/qdrant/qdrant/pull/6199#issuecomment-2743350845

### All Submissions:

### DIFF
--- a/lib/collection/src/shards/replica_set/snapshots.rs
+++ b/lib/collection/src/shards/replica_set/snapshots.rs
@@ -81,6 +81,7 @@ impl ShardReplicaSet {
     pub async fn restore_local_replica_from(
         &self,
         replica_path: &Path,
+        shard_initializing_flag_path: &Path,
         cancel: cancel::CancellationToken,
     ) -> CollectionResult<bool> {
         // `local.take()` call and `restore` task have to be executed as a single transaction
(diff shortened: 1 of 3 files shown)


agourlay:
Were you able to test this change?

agourlay:
The assertions make the storage compat. test is failing for a boolean payload which is on_disk.
When digging deeper, it seems that the column family for the `SimpleBoolIndex` is not found.

KShivendu:
> Were you able to test this change?

Hard to reproduce locally because it's related to fsync. My idea was to test this in chaos testing since it reproduced there again just today. Open to any suggestions that can reproduce it locally.

timvisee (approved):

agourlay:
Looks like it breaks CI for `test-shard-snapshot-api-s3-minio`

lib/collection/src/shards/replica_set/snapshots.rs
```diff
@@ -112,6 +119,9 @@ impl ShardReplicaSet {
 
             LocalShard::move_data(replica_path, &self.shard_path).await?;
 
+            // remove shard_id initialization flag because shard is fully recovered
```
generall, line 122: We don't know if it is fully recovered before we actually load the shard. It might be recovered from corrupted snapshot

timvisee (approved):

PR #6232, opened 2025-03-24
Merge `take_snapshot` and `take_partial_snapshot` methods

ffuugoo:
This PR merges `ShardReplicaSet::take_snapshot` and `ShardReplicaSet::take_partial_snapshot` methods. They are very similar to each other, and it's easier (on the HTTP API level) to call a single method instead of two separate ones.

This PR adds `manifest` parameter to `take_snapshot` and logic that creates a partial snapshot (instead of a regular one) based on this parameter, but it's not yet propagated into `take_snapshot` (it will be done in yet another PR, that will add partial snapshot API endpoints), so it always creates a regular snapshot.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/segment/src/segment/snapshot.rs
```diff
@@ -47,6 +49,42 @@ impl SnapshotEntry for Segment {
...
+                let tar = tar.descend(Path::new(&segment_id))?;
+                tar.blocking_append_data(
+                    &updated_manifest_json,
+                    Path::new("segment_manifest.json"),
+                )?;
+
+                let mut empty_manifest = None;
+                let request_manifest = manifest
+                    .get(segment_id)
+                    .unwrap_or_else(|| empty_manifest.insert(SegmentManifest::empty(segment_id)));
```
timvisee, line 71: I don't really get how this works. Shouldn't we take out the `empty_manifest` in between changing it to:

```rust
                let request_manifest = manifest
                    .get(segment_id)
                    .unwrap_or_else(|| SegmentManifest::empty(segment_id));
```

timvisee (approved):

lib/segment/src/segment/snapshot.rs
ffuugoo: `manifest` is `Option<&SegmentManifest>`. We can't return `&SegmentManifest::empty(...)` from `unwrap_or_else`, so we either have to clone `manifest` or use this trick to _conditionally_ initialize empty manifest (into `empty_manifest` variable) and return reference to this empty manifest from `unwrap_or_else` call.

timvisee: Ah, I get it, it's because of the reference.

Construction of `SegmentManifest::empty(segment_id)` is cheap though, it's only a `&str -> String`.

PR #6233, opened 2025-03-24
[order_by] Deprecate using `order_value` from payload, dedup by `(order_value, id)` tuple

coszio:
Initially, we started inserting the order_by value in the payload. Since version 1.11 we can already deprecate this behavior because now we have a dedicated field in the point struct.

This PR removes the legacy code which inserted and removed the order_by value in the payload.

### DIFF
--- a/lib/collection/src/collection/point_ops.rs
+++ b/lib/collection/src/collection/point_ops.rs
@@ -343,7 +343,7 @@ impl Collection {
                     // Get top results
                     .kmerge_by(|a, b| match order_by.direction() {
                         Direction::Asc => (a.order_value, a.id) < (b.order_value, b.id),
-                        Direction::Desc => (b.order_value, b.id) > (a.order_value, a.id),
+                        Direction::Desc => (a.order_value, a.id) > (b.order_value, b.id),
                     })
                     // Only keep the point with the most "valuable" order value
                     // This means that the same point will be repeated only if there is an alternating order value with another point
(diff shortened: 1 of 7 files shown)


lib/collection/src/collection/point_ops.rs
```diff
@@ -340,42 +340,26 @@ impl Collection {
...
+                    // { id: 1, values: [11, 21]}
+                    // { id: 2, values: [12, 22, 32]}
+                    //
+                    // The output will be
+                    // { id: 1, order_value: 11 }
+                    // { id: 2, order_value: 12 }
+                    // { id: 1, order_value: 21 }
+                    // { id: 2, order_value: 22 }
+                    // (no order_value 32)
+                    .dedup_by(|record_a, record_b| record_a.id == record_b.id)
```
timvisee, line 361: Is this desirable behavior?

Could we other wise dedup by the tuple `(record_a.id, record_a.order_value)`?

coszio: I have the same opinion, here I only documented the current behavior.

IIRC, the intention for this is that the case when a point has many values, to only use one of them if they are consecutive, and there is no interleaving with another point. But tbh, it looks more like a bug, like it puzzled even ourselves.

@generall thoughts?

timvisee: Thanks. I think this is better!

timvisee (approved):

PR #6234, opened 2025-03-24
Add fallback for missing all feature_flag

KShivendu:
If you pass:

```yaml
feature_flags:
  use_new_shard_key_mapping_format: true
  use_mutable_id_tracker_without_rocksdb: true
  payload_index_skip_rocksdb: true
```

Qdrant throws error because `all` field is not passed. 

```
Error: missing field `all` for key `feature_flags`
```

We should add a fallback for it (false) so other vars can be set independently. This was detected by chaos testing.

### All Submissions:


timvisee (approved):
Blunder from my side for forgetting this one. Thanks! 😁

PR #6245, opened 2025-03-24
Move filtering logic from RawScorer to FilteredScorer

xzfc:
# Current version

Currently, we have the following traits/structs:
- `RawScorer` filters points by deletion flags, then scores them.
- `FilteredScorer` filters points by `FilterContext`, then pass them to `RawScorer` to score.
  Some methods of `FilteredScorer` simply call the corresponding methods of `RawScorer`.

Both `RawScorer` and `FilteredScorer` filter points, and both of them perform the scoring.

# In this PR

In this PR, I moved a lot of logic from `RawScorer` to `FilteredScorer`.

- `RawScorer::score_points()` doesn't filter points anymore.
- `point_deleted`, `vec_deleted` bitslices are moved to `FilteredScorer`.
   Consequence: `FilteredScorer::check_vector()` now should be a bit faster.
    - It no longer performs a dyn trait call to `RawScorer`.
    - Now these bitslices are checked before checking `FilterContext`, not after.
- `peek_top_iter()` and `peek_top_all()` methods are moved to `FilteredScorer`.
- A lot of tests that use `RawScorer` updated to use `FilteredScorer` instead.

## ~~Removed temporary buffer `FilteredScorer::points_buffer`~~

Not in this version of PR.

<details><summary>What was in this PR previously?</summary>

<strike>
Minor change: this PR changes the interface of the scorer and the filterer to avoid the need for a temporary buffer.

Previously:
1. Caller prepares a vector of `PointOffsetType` and passes it to `FilteredScorer`.
2. `FilteredScorer`
   - filters in-place this vector (destroying the original values)
   - allocates a buffer of type `Vec<ScoredPointOffset>` (reusable, stored in `FilteredScorer`)
   - calls `RawScorer::score_points` with both of these vectors.
3. `RawScorer::score_points` also performs its own filtering. It returns a number of scored points.
4. `FilteredScorer` returns buffer slice truncated by this number to the caller.

In this PR:
1. Caller prepares a vector of `ScoredPointOffset`.
   Only the `.idx` field should be set; the `.score` field should be set to an arbitrary meaningless value.
2. `PointsFilterer` filters this vector in-place using `Vec::retain()`.
3. `RawScorer::score_points` fills the `.score` values in-place.
</strike>

</details>

<details><summary>Why it is no longer in this PR?</summary>

In the current version of the PR, `RawScorerImpl::score_points` calls `QueryScorer::score_stored_batch`, which accepts two separate slices.

Doing the same for `QueryScorer` would be complicated (but possible I guess).
Illustrated: if you replace `ids: &[PointOffsetType], scores: &mut [ScoreType]` with `ids_with_scores: &mut [ScoredPointOffset]`, then you can't just pass `ids` to `get_dense_batch` in the following code:

https://github.com/qdrant/qdrant/blob/23b5f352f13e52fa002799e101d4e843ca4edac6/lib/segment/src/vector_storage/query_scorer/metric_query_scorer.rs#L77-L81

</details>

# Performance

BFB on small (32 element) vectors: 3612 RPS -> 4721 RPS.
No significant changes on 1.5k vectors.

https://github.com/qdrant/vector-db-benchmark/actions/runs/14606832003

## Criterion benchmark results

A lot of benchmarks got a huge boost, but the benchmark logic it's not exactly the same, so take it with a grain of salt:
- Benchmarks that use `TestRawScorerProducer` no longer use `FakeFilterContext` (so one less dyn call in `FilteredScorer` for them).
- `FilteredScorer::score_points()` now returns an iterator, not a slice.

<details><summary>Results</summary>

`hnsw_build_graph`:
```
Benchmarking hnsw-index-build-group/hnsw_index: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 9.2s.
hnsw-index-build-group/hnsw_index
                        time:   [915.87 ms 920.81 ms 926.33 ms]
                        change: [-3.5310% -2.8506% -2.0601%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) high mild
```

`hnsw_build_asymptotic`:
```
hnsw-index-build-asymptotic/build-n-search-hnsw-5k
                        time:   [43.190 µs 43.231 µs 43.275 µs]
                        change: [-5.8648% -5.5461% -5.1992%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
hnsw-index-build-asymptotic/build-n-search-hnsw-1M
                        time:   [143.51 µs 143.76 µs 144.02 µs]
                        change: [-10.849% -10.650% -10.448%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
hnsw-index-build-asymptotic/build-n-search-hnsw-1M-score-point
                        time:   [68.044 µs 68.103 µs 68.162 µs]
                        change: [-18.860% -18.676% -18.505%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

scoring-vector/score-point
                        time:   [19.259 µs 19.275 µs 19.291 µs]
                        change: [-23.060% -22.902% -22.749%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
scoring-vector/score-point-10x
                        time:   [19.414 µs 19.448 µs 19.485 µs]
                        change: [-25.670% -25.517% -25.364%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  6 (6.00%) high mild
scoring-vector/score-point-50x
                        time:   [55.111 µs 55.168 µs 55.227 µs]
                        change: [-23.943% -23.653% -23.373%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild

scoring-vector/basic-score-point
                        time:   [144.57 µs 144.70 µs 144.82 µs]
                        change: [-0.8233% -0.6519% -0.4308%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe
scoring-vector/basic-score-point-10x
                        time:   [155.73 µs 155.89 µs 156.04 µs]
                        change: [-1.2111% -0.9589% -0.7167%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```

`hnsw_search_graph`:
```
hnsw-search-graph/uncompressed
                        time:   [357.41 µs 358.14 µs 358.86 µs]
                        change: [-17.125% -16.858% -16.586%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) low mild
hnsw-search-graph/compressed
                        time:   [360.08 µs 360.91 µs 361.75 µs]
                        change: [-16.297% -16.001% -15.667%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
hnsw-search-graph/plain time:   [19.076 ms 19.113 ms 19.156 ms]
                        change: [-5.7350% -5.5325% -5.3217%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
```

</details>

## Performance on glove-100

Made 30k unique batch search requests with 10 searches in each. Times for each batch are in ms.

<table><tr><td>

No filter:
|        | before | after  |
| ------ | -----: | -----: |
| min    | 10.017 |  8.731 |
| max    | 20.559 | 22.528 |
| mean   | 12.683 | 11.020 |
| median | 12.380 | 10.792 |
| p95    | 15.680 | 13.445 |
| p99    | 17.600 | 15.077 |

<td>

With no-op filter:
|        | before | after  |
| ------ | -----: | -----: |
| min    | 10.724 |  9.445 |
| max    | 20.958 | 22.160 |
| mean   | 13.595 | 11.986 |
| median | 13.267 | 11.700 |
| p95    | 16.967 | 14.755 |
| p99    | 19.192 | 16.051 |

</table>

Request for no filter:
```
POST collections/{collection}/points/search/batch
{
  "searches": [
    {"vector": <vector>, "limit": 10 },
    … (9 more) …
  ]
}
```

Request for no-op filter:
```
POST collections/{collection}/points/search/batch
{
  "searches": [
    {"vector": <vector>, "limit": 10, "filter": {"must_not": {"has_id": [1]}} },
    … (9 more) …
  ]
}
```

# Why

The new interface would be useful in incremental HNSW build as we would need to search without filtering.

Another potential improvement that this PR would allow described below. Suppose we perform an unfiltered search within a graph that has additional links. See this code:
https://github.com/qdrant/qdrant/blob/7abc684361fb81d8b62cf1554d8bf4fb65a706d7/lib/segment/src/index/hnsw_index/graph_layers.rs#L73-L80
This logic here is NOT: "use the first `m` neighbors aka main links";
Instead, the implemented logic is: "use the first `m` non-visited neighbors, i.e. take additional links if the main ones have already been visited".
It seems to me that the implemented logic imposes additional unnecessary work. I believe that we should flip filters like this (pseudo code):

```
// old
point_ids.filter(is_not_visited).filter(is_not_deleted).take(m)
// new
point_ids.filter(is_not_deleted).take(m).filter(is_not_visited)
```

### DIFF
--- a/lib/segment/src/index/hnsw_index/point_scorer.rs
+++ b/lib/segment/src/index/hnsw_index/point_scorer.rs
@@ -1,30 +1,119 @@
-use common::types::{PointOffsetType, ScoreType, ScoredPointOffset};
+use std::ops::Deref;
+use std::sync::atomic::AtomicBool;
 
+use bitvec::slice::BitSlice;
+use common::counter::hardware_counter::HardwareCounterCell;
+use common::fixed_length_priority_queue::FixedLengthPriorityQueue;
+use common::types::{PointOffsetType, ScoreType, ScoredPointOffset};
+use itertools::Itertools as _;
+
+use crate::common::operation_error::{CancellableResult, OperationResult, check_process_stopped};
+use crate::data_types::vectors::QueryVector;
 use crate::payload_storage::FilterContext;
-use crate::vector_storage::RawScorer;
+use crate::vector_storage::common::VECTOR_READ_BATCH_SIZE;
+use crate::vector_storage::quantized::quantized_vectors::QuantizedVectors;
+use crate::vector_storage::{
+    RawScorer, VectorStorage, VectorStorageEnum, check_deleted_condition, new_raw_scorer,
+};
 
 pub struct FilteredScorer<'a> {
-    pub raw_scorer: &'a dyn RawScorer,
-    pub filter_context: Option<&'a dyn FilterContext>,
-    points_buffer: Vec<ScoredPointOffset>,
+    raw_scorer: Box<dyn RawScorer + 'a>,
+    filter_context: Option<BoxCow<'a, dyn FilterContext + 'a>>,
+
+    /// Point deleted flags should be explicitly present as `false`
+    /// for each existing point in the segment.
+    /// If there are no flags for some points, they are considered deleted.
+    /// [`BitSlice`] defining flags for deleted points (and thus these vectors).
+    point_deleted: &'a BitSlice,
+    /// [`BitSlice`] defining flags for deleted vectors in this segment.
+    vec_deleted: &'a BitSlice,
+
+    scores_buffer: Vec<ScoreType>,
 }
 
 impl<'a> FilteredScorer<'a> {
+    /// Create a new filtered scorer.
+    ///
+    /// If present, `quantized_vectors` will be used for scoring, otherwise `vectors` will be used.
     pub fn new(
-        raw_scorer: &'a dyn RawScorer,
-        filter_context: Option<&'a dyn FilterContext>,
+        query: QueryVector,
+        vectors: &'a VectorStorageEnum,
+        quantized_vectors: Option<&'a QuantizedVectors>,
+        filter_context: Option<BoxCow<'a, dyn FilterContext + 'a>>,
+        point_deleted: &'a BitSlice,
+        hardware_counter: HardwareCounterCell,
+    ) -> OperationResult<Self> {
+        let raw_scorer = match quantized_vectors {
+            Some(quantized_vectors) => quantized_vectors.raw_scorer(query, hardware_counter)?,
+            None => new_raw_scorer(query, vectors, hardware_counter)?,
+        };
+        Ok(FilteredScorer {
+            raw_scorer,
+            filter_context,
+            point_deleted,
+            vec_deleted: vectors.deleted_vector_bitslice(),
+            scores_buffer: Vec::new(),
+        })
+    }
+
+    /// Create a new filtered scorer from a raw scorer.
+    #[cfg(feature = "testing")]
+    pub fn new_from_raw(
+        raw_scorer: Box<dyn RawScorer + 'a>,
+        point_deleted: &'a BitSlice,
+        vec_deleted: &'a BitSlice,
     ) -> Self {
         FilteredScorer {
             raw_scorer,
-            filter_context,
-            points_buffer: Vec::new(),
+            filter_context: None,
+            point_deleted,
+            vec_deleted,
+            scores_buffer: Vec::new(),
         }
     }
 
+    /// Create a new filtered scorer for testing purposes.
+    ///
+    /// # Panics
+    ///
+    /// Panics if [`new_raw_scorer`] fails.
+    #[cfg(feature = "testing")]
+    pub fn new_for_test(
+        vector: QueryVector,
+        vector_storage: &'a VectorStorageEnum,
+        point_deleted: &'a BitSlice,
+    ) -> Self {
+        let raw_scorer =
+            new_raw_scorer(vector, vector_storage, HardwareCounterCell::new()).unwrap();
+        FilteredScorer::new_from_raw(
+            raw_scorer,
+            point_deleted,
+            vector_storage.deleted_vector_bitslice(),
+        )
+    }
+
+    /// Return true if vector satisfies current search context for given point:
+    /// exists, not deleted, and satisfies filter context.
     pub fn check_vector(&self, point_id: PointOffsetType) -> bool {
-        match self.filter_context {
-            None => self.raw_scorer.check_vector(point_id),
-            Some(f) => f.check(point_id) && self.raw_scorer.check_vector(point_id),
+        match &self.filter_context {
+            None => self.simple_check_vector(point_id),
+            Some(f) => f.check(point_id) && self.simple_check_vector(point_id),
+        }
+    }
+
+    fn simple_check_vector(&self, point_id: PointOffsetType) -> bool {
+        check_deleted_condition(point_id, self.vec_deleted, self.point_deleted)
+    }
+
+    pub fn rescore_points(&mut self, points: &mut [ScoredPointOffset]) {
+        let point_ids = points.iter().map(|p| p.idx).collect_vec();
+        if self.scores_buffer.len() < point_ids.len() {
+            self.scores_buffer.resize(point_ids.len(), 0.0);
+        }
+        self.raw_scorer
+            .score_points(&point_ids, &mut self.scores_buffer[..point_ids.len()]);
+        for (point, score) in std::iter::zip(points, &self.scores_buffer) {
+            point.score = *score;
         }
     }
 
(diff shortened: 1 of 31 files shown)


lib/segment/src/vector_storage/async_raw_scorer.rs
```diff
@@ -69,45 +70,17 @@ impl<TQueryScorer> RawScorer for AsyncRawScorerImpl<'_, TQueryScorer>
...
-            })
-            .unwrap();
-
-        // ToDo: io_uring is experimental, it can fail if it is not supported.
-        // Instead of silently falling back to the sync implementation, we prefer to panic
-        // and notify the user that they better use the default IO implementation.
-
-        processed
-    }
+        let scores = Cell::from_mut(scores).as_slice_of_cells();
```
generall, line 77: why do we need a Cell here?

lib/segment/src/vector_storage/async_raw_scorer.rs
```diff
@@ -69,45 +70,17 @@ impl<TQueryScorer> RawScorer for AsyncRawScorerImpl<'_, TQueryScorer>
 where
     TQueryScorer: QueryScorer<[VectorElementType]>,
 {
-    fn score_points(&self, points: &[PointOffsetType], scores: &mut [ScoredPointOffset]) -> usize {
+    fn score_points(&self, scores: &mut [ScoredPointOffset]) {
         if self.is_stopped.load(Ordering::Relaxed) {
-            return 0;
+            return;
```
generall, line 75: Not sure if this is correct: previous implementation explicitly returned no scored points. Current implementation, looks like, returns points with unassigned scores

lib/segment/src/index/hnsw_index/point_filterer.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// [`RawScorer::score_points`]: crate::vector_storage::RawScorer::score_points
+    /// [`score`]: ScoredPointOffset::score
+    /// [`idx`]: ScoredPointOffset::idx
+    pub fn filter_scores(&self, scores: &mut Vec<ScoredPointOffset>, limit: usize) {
+        match self.filter_context {
+            None => scores.retain(|score| self.simple_filterer.check_vector(score.idx)),
+            Some(f) => scores
+                .retain(|score| self.simple_filterer.check_vector(score.idx) && f.check(score.idx)),
+        }
+        if limit != 0 {
```
generall, line 75: Why this condition is needed?

lib/segment/src/vector_storage/raw_scorer.rs
```diff
@@ -951,43 +938,13 @@ where
...
-        if self.is_stopped.load(Ordering::Relaxed) {
-            return vec![];
-        }
-        let mut scores = vec![];
-        for point_id in points {
-            scores.push(ScoredPointOffset {
-                idx: point_id,
-                score: self.query_scorer.score_stored(point_id),
-            });
+            return;
```
generall, line 943: same here: original semantic is that we return no scored points to prevent further search

lib/segment/src/vector_storage/raw_scorer.rs
```diff
@@ -53,21 +53,8 @@ use crate::vector_storage::query_scorer::multi_metric_query_scorer::MultiMetricQ
...
-    /// * `points` - points to score
-    ///
-    /// # Returns
-    ///
-    /// Vector of scored points
-    fn score_points_unfiltered(
-        &self,
-        points: &mut dyn Iterator<Item = PointOffsetType>,
-    ) -> Vec<ScoredPointOffset>;
+    /// Set [`ScoredPointOffset::score`] field for each point in the given slice.
```
generall, line 56: Documentation above should be updated

generall:
After review in-call:

- Try to merge Filterer and RawScorer in the following way:

FIlteredScorer uses RawScorer to combine into a single call
```
            filterer.filter_scores(&mut scores, limit);
            scorer.score_points(&mut scores);
```

If RawScorer is needed without filters, it can be extracted (used from) FilteredScorer


- `is_stopped` flag should be moved FilteredScorer
- Move `peek_top_iter` and `peek_top_all` into FilteredScorer


---

For a separate PR:

- Try to use `query_scorer.score_stored_batch` in `score_points` of RawScorerImpl

lib/common/common/src/types.rs
```diff
@@ -15,6 +15,15 @@ pub struct ScoredPointOffset {
     pub score: ScoreType,
 }
 
+impl ScoredPointOffset {
+    pub fn new_unset(idx: PointOffsetType) -> Self {
```
timvisee, line 19: How about this?

```suggestion
    pub fn new_zero(idx: PointOffsetType) -> Self {
```

lib/segment/src/index/hnsw_index/point_filterer.rs
```diff
@@ -0,0 +1,79 @@
+use bitvec::slice::BitSlice;
+use common::types::{PointOffsetType, ScoredPointOffset};
+
+use crate::id_tracker::IdTracker;
+use crate::payload_storage::FilterContext;
+use crate::vector_storage::{VectorStorage, check_deleted_condition};
+
+#[derive(Clone, Copy)]
+pub struct SimplePointsFilterer<'a> {
```
timvisee, line 9: I think this name would define it's purpose a bit better:

```suggestion
pub struct DeletedPointsFilter<'a> {
```

Simple implies to me that it is a naive filter, similar to all the other 'simple's we use across our code base.

lib/segment/src/index/hnsw_index/point_filterer.rs
timvisee: Shall we put it at the beginning of the function body without the condition?

xzfc: Relevant code in the current version of this PR:
https://github.com/qdrant/qdrant/blob/e2bce5a09fe5e61db118953493f27899cfc930dd/lib/segment/src/index/hnsw_index/point_scorer.rs#L133-L141

---

`limit == 0` means "don't truncate scores". I also don't like, but this logic was there before this PR. I've added a comment in `FilteredScorer::score_points` about this.

(maybe `limit: usize` can be replaced with `Option<NonZeroUsize>`, but I don't want more bloat in this PR)

> Shall we put it at the beginning of the function body without the condition?

No, retain-then-retain may produce a different result than truncate-then-retain.

lib/segment/src/index/hnsw_index/point_scorer.rs
```diff
@@ -1,75 +1,154 @@
...
-        match self.filter_context {
-            None => self.raw_scorer.check_vector(point_id),
-            Some(f) => f.check(point_id) && self.raw_scorer.check_vector(point_id),
-        }
+        check_deleted_condition(point_id, self.vec_deleted, self.point_deleted)
+            && self
+                .filter_context
+                .as_ref()
+                .is_none_or(|f| f.check(point_id))
     }
```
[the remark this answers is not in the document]
xzfc: > As‑is, calls to .is_none_or(...) will fail to compile.

It doesn't. Please don't report issues that could be caught by `cargo check`.

lib/segment/src/vector_storage/async_raw_scorer.rs
```diff
@@ -63,60 +45,15 @@ impl<TQueryScorer> RawScorer for AsyncRawScorerImpl<'_, TQueryScorer>
...
-            .read_vectors_async(points_stream, |idx, point_id, other_vector| {
-                scores[idx] = ScoredPointOffset {
-                    idx: point_id,
-                    score: self.query_scorer.score(other_vector),
-                };
-                processed += 1;
+            .read_vectors_async(points_stream, |idx, _point_id, other_vector| {
+                scores[idx] = self.query_scorer.score(other_vector);
             })
             .unwrap();
```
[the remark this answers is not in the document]
xzfc: I've restored the comment about explicit panic in async scorer.

generall (approved):

PR #6256, opened 2025-03-25
`sum_scores` recommendation strategy

coszio:
Adds a new strategy for recommendation: `sum_scores`

Formula for this is:

$score = \sum s(v^+) - \sum s(v^-)$

where $v^+$ are positive vectors and $v^-$ are negative ones

### DIFF
--- a/lib/segment/src/data_types/vectors.rs
+++ b/lib/segment/src/data_types/vectors.rs
@@ -680,7 +680,8 @@ impl NamedQuery<RecoQuery<VectorInternal>> {
 #[derive(Debug, Clone)]
 pub enum QueryVector {
     Nearest(VectorInternal),
-    Recommend(RecoQuery<VectorInternal>),
+    RecommendBestScore(RecoQuery<VectorInternal>),
+    RecommendSumScores(RecoQuery<VectorInternal>),
     Discovery(DiscoveryQuery<VectorInternal>),
     Context(ContextQuery<VectorInternal>),
 }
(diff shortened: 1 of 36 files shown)


lib/segment/src/vector_storage/query/reco_query.rs
```diff
@@ -37,42 +37,89 @@ impl<T, U> TransformInto<RecoQuery<U>, T, U> for RecoQuery<T> {
...
+        // Sum all positive vectors scores
+        let positive_score: ScoreType = self.0.positives.iter().map(&similarity).sum();
+
+        // Sum all negative vectors scores
+        let negative_score: ScoreType = self.0.negatives.iter().map(&similarity).sum();
+
+        // Subtract
+        positive_score - negative_score
     }
 }
```
coszio, line 124: This is the scoring implementation

lib/segment/src/data_types/vectors.rs
```diff
@@ -692,7 +693,9 @@ impl TransformInto<QueryVector, VectorInternal, VectorInternal> for QueryVector
     {
         match self {
             QueryVector::Nearest(v) => f(v).map(QueryVector::Nearest),
-            QueryVector::Recommend(v) => Ok(QueryVector::Recommend(v.transform(&mut f)?)),
+            QueryVector::RecommendBestScore(v) | QueryVector::RecommendSumScores(v) => {
+                Ok(QueryVector::RecommendSumScores(v.transform(&mut f)?))
+            }
```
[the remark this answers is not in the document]
coszio: Thanks, I missed this 🙌

lib/segment/src/vector_storage/query/reco_query.rs
```diff
@@ -37,42 +37,89 @@ impl<T, U> TransformInto<RecoQuery<U>, T, U> for RecoQuery<T> {
...
+
+        // get max similarity to positives and max to negatives
+        let max_positive = positive_similarities
+            .max_by(|a, b| a.total_cmp(b))
+            .unwrap_or(ScoreType::NEG_INFINITY);
 
-        merge_similarities(positive_similarities, negative_similarities)
+        let max_negative = negative_similarities
+            .max_by(|a, b| a.total_cmp(b))
+            .unwrap_or(ScoreType::NEG_INFINITY);
```
timvisee, line 79: Is this indeed correct or should it be the inverse like this?

```suggestion
        let max_negative = negative_similarities
            .min_by(|a, b| a.total_cmp(b))
            .unwrap_or(ScoreType::INFINITY);
```

coszio: It is not min_by, because we are calculating whether the closest example is a negative or positive example. This means we want to know the highest similarity in both cases

generall (approved):

PR #6260, opened 2025-03-26
Resharding: only read vectors/payload of points to transfer

timvisee:
_WIP_

An optimization in resharding shard transfers - only read vectors and payload of points that actually need to be transferred.

This can result in a significant performance improvement if:
- a small portion needs to be transferred (when having a large number of shards)
- when vector/payload data is on disk
- when memory pressure is high

### Tasks
- [ ] Split batching logic for use with/without hash ring filter to make the logic more clear

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/forward_proxy_shard.rs
+++ b/lib/collection/src/shards/forward_proxy_shard.rs
@@ -131,17 +131,26 @@ impl ForwardProxyShard {
         debug_assert!(batch_size > 0);
         let limit = batch_size + 1;
         let _update_lock = self.update_lock.lock().await;
+
+        // If scrolling with data, we fetch all point data right in the scroll request
+        // If we scroll without data, we fetch only point IDs and defer fetching all point data
+        // with an explicit retrieve operation
+        //
+        // If using a hash ring filter, we defer reading data so that we only read vectors and
+        // payload for points that are actually transferred.
+        let scroll_with_data = hashring_filter.is_none();
+
         let mut batch = self
             .wrapped_shard
             .scroll_by(
                 offset,
                 limit,
-                &WithPayloadInterface::Bool(true),
-                &true.into(),
+                &WithPayloadInterface::Bool(scroll_with_data),
+                &WithVector::Bool(scroll_with_data),
                 None,
                 runtime_handle,
                 None,
-                None,                           // no timeout
+                None,                           // No timeout
                 HwMeasurementAcc::disposable(), // Internal operation, no need to measure hardware here.
             )
             .await?;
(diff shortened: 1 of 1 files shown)


ffuugoo (approved):

lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -196,6 +168,146 @@ impl ForwardProxyShard {
...
+        // Oversample batch size to account for points that will be filtered out by the hash ring
+        let oversample_factor = match &hashring_filter {
+            HashRingRouter::Single(_) => 1,
+            // - resharding: 1 -> 2, transfer 50%,  factor 2
+            // - resharding: 2 -> 3, transfer 33%,  factor 3
+            // - resharding: 3 -> 4, transfer 25%,  factor 4
+            // - resharding: 2 -> 1, transfer 100%, factor 1
+            // - resharding: 3 -> 2, transfer 50%,  factor 2
+            // - resharding: 4 -> 3, transfer 33%,  factor 3
+            HashRingRouter::Resharding { old: _, new } => new.len().max(1),
```
KShivendu, line 250:
```suggestion
            HashRingRouter::Resharding { old: _, new } => new.len(),
```

What's the use of `.max(1)` here?  `new.len()` will always >= 1. This will limit the factor to 1 and defeat the purpose of oversampling.

KShivendu (approved):
I think there's a small mistake. Rest LGTM

lib/collection/src/shards/forward_proxy_shard.rs
KShivendu: Okay nvm.`max(1)` will ensure that the final value remains >= 1. What I described would be the behaviour of `.min(1)`.

timvisee: :+1: Yeah, I'm also always confused by min/max in practice.

Even though we should never reach 0 shards, it's just there to stay on the safe side.

ffuugoo: I always use `cmp::max`/`cmp::min` explicitly, cause `.max(...)` confuses me. :/

```suggestion
            HashRingRouter::Resharding { old: _, new } => cmp::max(1, new.len()),
```

timvisee:
I did some testing, results are about what I did expect.

Setup:
- 4 nodes
- 3 shards, resharding to 4
- replication factor 1
- 512d vectors
- integer and large text payload
- everything on-disk
- size of collection is different per type of storage, and not super relevant

I did measure the wall clock time for the three resharding transfers that happen when resharding up.

Result:
- On ramdisk:
  - Before PR: 17s, 23s, 30s
  - With PR: 17s, 23s, 30s
- On NVMe drive:
  - Before PR: 16s, 17s, 17s
  - With PR: 10s, 12s, 16s
- On old SATA HDD:
  - Before PR: 81s, 110s, 116s
  - With PR: 81s, 85s, 88s

This is definitely not a scientific test, but this naive approach already shows an improvement. I expect (but did not test) the effect to be a lot greater with networked storage and/or with high memory pressure.

bfb command to set up collection:
```bash
bfb --uri http://127.0.0.1:6334/,http://127.0.0.2:6334/,http://127.0.0.3:6334/,http://127.0.0.4:6334/ -t3 -p3 -n 500000 -d 512 --shards 3 --replication-factor 1 --indexing-threshold 0 --skip-wait-index --int-payloads 100 --text-payloads --text-payload-length 512 --timestamp-payload --retry 3 --on-disk-payload --on-disk-payload-index --on-disk-index true --on-disk-vectors true --timeout 20
```

PR #6279, opened 2025-03-31
[query] Optimize fetching payloads/vectors at local shard

coszio:
Changes the structure of the planned query to make sure we fetch payloads and vectors only at the last step before returning the results from a shard.

This improvement makes sure that the impact of fetching many large payloads is mitigated.

For instance, before this change, a single search of limit 100 in 500 segments would mean that 50,000 payloads would be fetched, while we only needed 100 of them

### DIFF
--- a/lib/collection/src/operations/universal_query/planned_query.rs
+++ b/lib/collection/src/operations/universal_query/planned_query.rs
@@ -18,7 +18,7 @@ pub struct PlannedQuery {
     /// This retains the recursive structure of the original queries.
     ///
     /// One per each query in the batch
-    pub root_plans: Vec<MergePlan>,
+    pub root_plans: Vec<RootPlan>,
 
     /// All the leaf core searches
     pub searches: Vec<CoreSearchRequest>,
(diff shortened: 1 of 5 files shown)


lib/collection/src/operations/universal_query/planned_query.rs
```diff
@@ -153,9 +161,9 @@ impl PlannedQuery {
                         query,
                         filter,
                         score_threshold,
-                        with_vector: Some(with_vector),
-                        with_payload: Some(with_payload),
-                        offset: 0, // offset is handled at collection level
+                        with_vector: Some(false.into()), // will be fetched after aggregating from segments
+                        with_payload: Some(false.into()), // will be fetched after aggregating from segments
```
coszio, line 165: we can look later into removing these params from `CoreSearchRequest` and `QueryScrollRequest`, but for now we can just disable them here.

lib/segment/src/segment/entry.rs
```diff
@@ -99,7 +97,7 @@ impl SegmentEntry for Segment {
             hw_counter,
         )?;
 
-        self.process_search_result(internal_results, with_payload, with_vector, hw_counter)
+        self.process_search_result(internal_results, &false.into(), &false.into(), hw_counter)
```
timvisee, line 100: It was a parameter before, and it's false now.

Can we use false here, because if a user specified true, it'll be fetched in the separate step that is implemented in this PR?

timvisee (approved):
Thanks!

Tested the scenario I had before, using hardware counters to measure the number of reads. This fixes the scenario, reading the expected amount of payloads. :raised_hands:

generall (approved):

PR #6288, opened 2025-03-31
Include `version` field in `Record` structs

coszio:
As part of a further refactoring, and to be consistent with query interface, I propose to include the point version for scrolls too. 

This is simple to add, since we were already resolving it, but we were discarding it right away.
This also fixes the fact that scroll-like queries would return a synthetic version 0 in the response.

This PR adds a `version` field to `RecordInternal`, `rest::Record` and `grpc::RetrievedPoint`

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -7898,11 +7898,17 @@
         "description": "Point data",
         "type": "object",
         "required": [
-          "id"
+          "id",
+          "version"
         ],
         "properties": {
           "id": {
             "$ref": "#/components/schemas/ExtendedPointId"
+          },
+          "version": {
+            "type": "integer",
+            "format": "uint64",
+            "minimum": 0
           },
           "payload": {
             "description": "Payload - values assigned to the point",
@@ -9319,6 +9325,7 @@
             "example": [
               {
                 "id": 40,
+                "version": 12,
                 "payload": {
                   "city": "London",
                   "color": "green"
@@ -9332,6 +9339,7 @@
               },
               {
                 "id": 41,
+                "version": 13,
                 "payload": {
                   "city": "Paris",
                   "color": "red"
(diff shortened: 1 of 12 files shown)


docs/redoc/master/openapi.json
```diff
@@ -7898,12 +7898,18 @@
...
-          "id"
+          "id",
+          "version"
         ],
         "properties": {
           "id": {
             "$ref": "#/components/schemas/ExtendedPointId"
           },
+          "version": {
+            "type": "integer",
```
KShivendu, line 7909: Maybe add a description here.

KShivendu (approved):
```
POST /collections/benchmark/points/scroll
{
  "limit": 1
}
```
returns
```
{
  "result": {
    "points": [
      {
        "id": 0,
        "version": 0,
        "payload": {}
      }
    ],
    "next_page_offset": 1
  },
  "status": "ok",
  "time": 0.001338581
}
```

It's working as expected with the HTTP APIs. However, it's suddenly a bit more bulkier. I'm wondering if we should make it optional by default by introducing a `with_version`. Wdyt?

KShivendu (changes requested):
Please update integration tests :)

generall:
Clients might not like this change

generall:
If not strictly required, I would prefer not to do that

timvisee:
> If not strictly required, I would prefer not to do that

I agree I don't think it's strictly required. I would be fine with adding it behind a flag, but I don't see a real use case for it yet.

We should definitely not add it by default. It makes scrolling point IDs twice as bulky in terms of bytes per points in the REST interface.

coszio:
Actually, if we wanted to use versions to ensure correct payload for searched points we can do it inside a local shard, so we'd only have to change `RecordInternal`, which avoids changing the interface.

I will draft for now

PR #6289, opened 2025-03-31
Scroll lock: make sure no segments are modified between scroll and retrieve

generall:
This PR fixes rare but possible loss of points in scroll results:

due to 2-stage search+retrieve approach in scroll, we can't guarantee that points retrieved on the first stage will be available and/or unchanged on the second stage. This might create a problem of either returning changed payload, which doesn't satisfy filters anymore, or compromise number of returned points.

It is especially critical for scroll request, which relies on number of returned points to generate a "next_page" offset, and is used in internal operations like shard streaming.

What I have tried and didn't work:

- use segments lock and https://amanieu.github.io/parking_lot/parking_lot/struct.RwLock.html#method.read_recursive - didn't work because it would require to have sync locks which live between await blocks.
- Read version along with point ids and check version on retrieve - won't help with  deleted points
- Move read points with payloads right away - reading payloads in advance is too expensive disk-wise

Current approach drawbacks:

- Usage of the scroll-lock is not enforced by the compiler. Have to rely on carefulness
- Update operations might be slightly delayed compared to original version

### DIFF
--- a/lib/collection/src/collection_manager/collection_updater.rs
+++ b/lib/collection/src/collection_manager/collection_updater.rs
@@ -47,8 +47,7 @@ impl CollectionUpdater {
         let scroll_lock = segments.read().scroll_read_lock.clone();
 
         // Use block_in_place here to avoid blocking the current async executor
-        let _scroll_lock =
-            tokio::task::block_in_place(|| tokio::sync::RwLock::write_owned(scroll_lock));
+        let _scroll_lock = tokio::task::block_in_place(|| scroll_lock.blocking_write());
 
         let operation_result = match operation {
             CollectionUpdateOperations::PointOperation(point_operation) => {
--- a/lib/collection/src/collection_manager/collection_updater.rs
+++ b/lib/collection/src/collection_manager/collection_updater.rs
@@ -45,7 +45,10 @@ impl CollectionUpdater {
         // Allow only one update at a time, ensure no data races between segments.
         // let _lock = self.update_lock.lock().unwrap();
         let scroll_lock = segments.read().scroll_read_lock.clone();
-        let _scroll_lock = futures::executor::block_on(scroll_lock.write());
+
+        // Use block_in_place here to avoid blocking the current async executor
+        let _scroll_lock =
+            tokio::task::block_in_place(|| tokio::sync::RwLock::write_owned(scroll_lock));
 
         let operation_result = match operation {
             CollectionUpdateOperations::PointOperation(point_operation) => {
--- a/lib/collection/src/collection_manager/collection_updater.rs
+++ b/lib/collection/src/collection_manager/collection_updater.rs
@@ -44,6 +44,9 @@ impl CollectionUpdater {
     ) -> CollectionResult<usize> {
         // Allow only one update at a time, ensure no data races between segments.
         // let _lock = self.update_lock.lock().unwrap();
+        let scroll_lock = segments.read().scroll_read_lock.clone();
+        let _scroll_lock = futures::executor::block_on(scroll_lock.write());
+
         let operation_result = match operation {
             CollectionUpdateOperations::PointOperation(point_operation) => {
                 process_point_operation(segments, op_num, point_operation, hw_counter)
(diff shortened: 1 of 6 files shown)


timvisee:
> Tried but failed: Read version along with point ids and check version on retrieve - won't help with deleted points

Could you elaborate on why this doesn't help with deleted points? Can't we detect deleted points by comparing point count between the two calls?

The only way we could resolve the situation in that case is to re-fetch. I don't suggest that to be a good idea, but I'm just curious on why it didn't work out.

coszio:
> Could you elaborate on why this doesn't help with deleted points? Can't we detect deleted points by comparing point count between the two calls?

I think this solves the problem of returning different payload than what the initial selection stated. However, it doesn't solve the problem of returning less points than initially selected. In regular queries it shouldn't be problematic, because we don't rely on the last point to provide a `next_page_offset`, but in scroll we do.

lib/collection/src/collection_manager/collection_updater.rs
```diff
@@ -44,6 +44,12 @@ impl CollectionUpdater {
     ) -> CollectionResult<usize> {
         // Allow only one update at a time, ensure no data races between segments.
         // let _lock = self.update_lock.lock().unwrap();
+        let scroll_lock = segments.read().scroll_read_lock.clone();
+
+        // Use block_in_place here to avoid blocking the current async executor
+        let _scroll_lock =
+            tokio::task::block_in_place(|| tokio::sync::RwLock::write_owned(scroll_lock));
```
coszio: This returns a future, does it mean that it will not make the update wait until the scroll write lock gets acquired?

timvisee: Thanks! I only half-committed it seems. Bumped it just now in https://github.com/qdrant/qdrant/pull/6289/commits/3b65d006b72994f9294b712381c9fcad858f5035.

coszio (approved):
Since we rely on scrolling for transfers, we need them to be accurate, so I cannot oppose this change.

However, we should be aware that this can hurt some performance under specific workloads.

In manual testing, I created a collection with
```bash
bfb --segments 100 -n 100000 --on-disk-payload --text-payloads --text-payload-length 500
```

Then I ran scrolls and upserts simultaneously

- Upsert: `bfb --skip-create -n 100000 --text-payloads --text-payload-length 500 --batch-size 1 --throttle 1000`.
    - this didn't trigger optimizations, so that it reflects only the upsert effects
- Scroll: `bfb --skip-create --skip-upload --skip-wait-index --scroll`

I observed some degradation in RPS
| Branch | Upsert RPS (thottled to 1000) | Scroll RPS |
| --- | --- | --- |
| dev | 999.97 | 1,996.85 | 
| scroll-lock | 866.80 | 1,722.63 | 

As a baseline, scrolls without upserting simultaneously reached up to around 2,300 RPS

timvisee:
> Since we rely on scrolling for transfers, we need them to be accurate, so I cannot oppose this change.
> 
> However, we should be aware that this can hurt some performance under specific workloads.
> 
> In manual testing, using `perf` builds, I created a collection with
> 
> ```shell
> bfb --segments 100 -n 100000 --on-disk-payload --text-payloads --text-payload-length 500
> ```
> 
> Then I ran scrolls and upserts simultaneously
> 
>     * Upsert: `bfb --skip-create -n 100000 --text-payloads --text-payload-length 500 --batch-size 1 --throttle 1000`.
>       
>       * this didn't trigger optimizations, so that it reflects only the upsert effects
> 
>     * Scroll: `bfb --skip-create --skip-upload --skip-wait-index --scroll`
> 
> 
> I observed some degradation in RPS
> Branch 	Upsert RPS (thottled to 1000) 	Scroll RPS
> dev 	999.97 	1,996.85
> scroll-lock 	866.80 	1,722.63
> 
> As a baseline, scrolls without upserting simultaneously reached up to around 2,300 RPS

Thank you very much for the careful review. It'd be nice to not have this impact, but consistency is more important here so we don't really have a choice.

Maybe some of our future efforts to clean locks around this area up a bit will have a positive effect :crossed_fingers:

ffuugoo (approved):

PR #6292, opened 2025-04-01
Test behaviour of Qdrant with shard initializing flag

KShivendu:
Add a new integration test that ensures that Qdrant is able to load a dirty shard (shard initializing flag still present). This test is expected to fail currently because Qdrant isn't doing the right thing. There will be another PR on top for this that fixes the test failure.

It currently loads a dummy shard and stays like that (since we have `handle_collection_load_errors: true` in our `development.yaml`). Ideally it should initiate a transfer and get the local shard fixed.

### All Submissions:

### DIFF
--- a/tests/consensus_tests/test_failed_snapshot_recovery.py
+++ b/tests/consensus_tests/test_failed_snapshot_recovery.py
@@ -1,6 +1,7 @@
 import pathlib
 import subprocess
 import tempfile
+import shutil
 import requests
 from time import sleep
 
@@ -74,6 +75,16 @@ def corrupt_snapshot(snapshot_path: pathlib.Path, segment_name: str):
 
             # Replace the original snapshot tar with the corrupted one
             os.rename(new_segment_archive, snapshot_path)
+
+def corrupt_shard_dir(shard_path: pathlib.Path):
+    segements_path = shard_path / "segments"
+    segment_names = filter(lambda x: x.is_dir(), pathlib.Path(segements_path).iterdir())
+
+    first_segment_path = next(segment_names)
+    shutil.rmtree(first_segment_path)
+
+    wal_path = shard_path / "wal"
+    shutil.rmtree(wal_path)
 
 def remove_file_from_tar(original_tar, file_to_remove, new_tar):
     file_to_remove = file_to_remove.replace(os.sep, "/")
@@ -218,11 +229,15 @@ def test_failed_snapshot_recovery(tmp_path: pathlib.Path):
 @pytest.mark.parametrize("transfer_method", ["snapshot", "stream_records", "wal_delta"])
 def test_dirty_shard_handling_with_active_replicas(tmp_path: pathlib.Path, transfer_method: str):
     assert_project_root()
+    extra_env = {
+        "QDRANT__STORAGE__SHARD_TRANSFER_METHOD": transfer_method,
+        "QDRANT__STORAGE__HANDLE_COLLECTION_LOAD_ERRORS": "false"
+    }
 
     peer_api_uris, peer_dirs, bootstrap_uri = start_cluster(
         tmp_path,
         N_PEERS,
-        extra_env={"QDRANT__STORAGE__SHARD_TRANSFER_METHOD": transfer_method },
+        extra_env=extra_env,
     )
 
     create_collection(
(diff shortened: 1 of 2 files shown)


generall:
I assume the CI failure is intended?

tests/consensus_tests/test_failed_snapshot_recovery.py
```diff
@@ -212,4 +212,84 @@ def test_failed_snapshot_recovery(tmp_path: pathlib.Path):
...
+    Path(flag_path).touch()
+
+    # Kill last peer
+    p = processes.pop()
+    p.kill()
+
+    # Restart same peer
+    peer_api_uris[-1] = start_peer(
+        peer_dirs[-1], f"peer_{N_PEERS}_restarted.log", bootstrap_uri
+    )
```
timvisee, line 260: IMO we should also try corrupting the files on disk, and totally removing the local shard data. We want to make sure it still properly starts in those scenarios.

tests/consensus_tests/test_failed_snapshot_recovery.py
```diff
@@ -212,4 +212,84 @@ def test_failed_snapshot_recovery(tmp_path: pathlib.Path):
...
+
+    # Assert that the local shard is active and not empty
+    local_shards = get_local_shards(peer_api_uris[-1])
+    assert len(local_shards) == 1
+    assert local_shards[0]["shard_id"] == 0
+    assert local_shards[0]["state"] == "Active"
+    assert local_shards[0]["points_count"] == 1000  # FAILS because of dummy shard
+
+    # Qdrant loads a dummy shard (since handle_collection_load_errors=true) so we get 0 points instead of 1000 and test fails
+    # Ideally Qdrant should try to recover this shard automatically using other replicas
```
timvisee, line 276: `handle_collection_load_errors` is only `true` when the development configuration is applied, because it is in `config/development.yaml`.

I think we should explicitly disable the flag in this case, because that would better mimic a real scenario. I believe the code paths are a little different then.

tests/consensus_tests/test_failed_snapshot_recovery.py
```diff
@@ -212,4 +212,84 @@ def test_failed_snapshot_recovery(tmp_path: pathlib.Path):
...
+    # Restart same peer
+    peer_api_uris[-1] = start_peer(
+        peer_dirs[-1], f"peer_{N_PEERS}_restarted.log", bootstrap_uri
+    )
+
+    # Wait for end of shard transfer
+    wait_for_collection_shard_transfers_count(peer_api_uris[0], COLLECTION_NAME, 0)
+
+    # Wait for all replicas to be active on the receiving peer
+    wait_for_all_replicas_active(peer_api_uris[-1], COLLECTION_NAME)
```
timvisee, line 266: It might not start a transfer instantly, and so it can fall through here.

Maybe we should do this

```
wait_for_collection_shard_transfers_count(peer_api_uris[0], COLLECTION_NAME, 1)
wait_for_collection_shard_transfers_count(peer_api_uris[0], COLLECTION_NAME, 0)
```

though we might need more points to make the transfer longer so it has the time to detect it.

tests/consensus_tests/test_failed_snapshot_recovery.py
```diff
@@ -212,4 +212,84 @@ def test_failed_snapshot_recovery(tmp_path: pathlib.Path):
...
+
+    # Assert that the local shard is active and not empty
+    local_shards = get_local_shards(peer_api_uris[-1])
+    assert len(local_shards) == 1
+    assert local_shards[0]["shard_id"] == 0
+    assert local_shards[0]["state"] == "Active"
+    assert local_shards[0]["points_count"] == 1000  # FAILS because of dummy shard
```
timvisee, line 273: Why does it fail? Isn't it supposed to be recovered now?

timvisee: I see the comment is removed in <https://github.com/qdrant/qdrant/pull/6293> :+1:

tests/consensus_tests/test_failed_snapshot_recovery.py
KShivendu: Doesn't really matter when you have a dummy shard, but added.

tests/consensus_tests/test_failed_snapshot_recovery.py
KShivendu: Done.

tests/consensus_tests/test_failed_snapshot_recovery.py
KShivendu: Done. Thanks!

KShivendu:
The same commits are now part of https://github.com/qdrant/qdrant/pull/6293. So closing this now.

PR #6293, opened 2025-04-01
Recover dirty shards using other replicas when marked Dead

KShivendu:
Fixes the bug demonstrated in https://github.com/qdrant/qdrant/pull/6292

### All Submissions:

TODO:
- [x] Delete the shard initializing flag after initiating transfer to the dummy shard for recovery
- ~Try to load shards even if they are dirty instead of directly recreating as empty shard - especially important when we have 0 replicas?~ Not desirable anymore.

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -681,9 +681,11 @@ impl Collection {
                 continue;
             }
 
-            if !(this_peer_state == Some(Dead) || replica_set.is_dirty().await) {
+            if !(this_peer_state == Some(Dead) && !replica_set.is_recovery().await) {
                 continue; // All good
             }
+
+            // Reaches here only if replica is Dead but not in recovery mode
 
             // Try to find dead replicas with no active transfers
             let transfers = shard_holder.get_transfers(|_| true);
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -683,7 +683,7 @@ impl Collection {
 
             #[allow(clippy::nonminimal_bool)]
             {
-                if !(this_peer_state == Some(Dead) && replica_set.is_dirty().await) {
+                if !(this_peer_state == Some(Dead) && !replica_set.is_dirty().await) {
                     continue; // All good
                 }
             }
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -683,7 +683,7 @@ impl Collection {
 
             #[allow(clippy::nonminimal_bool)]
             {
-                if !(this_peer_state == Some(Dead) && !replica_set.is_dirty().await) {
+                if !(this_peer_state == Some(Dead) && replica_set.is_dirty().await) {
                     continue; // All good
                 }
             }
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -681,9 +681,8 @@ impl Collection {
                 continue;
             }
 
-            // Don't recover our replicas if started in recovery mode
-            let is_recovery_mode = self.shared_storage_config.recovery_mode.is_some();
-            if is_recovery_mode {
+            // Don't automatically recover replicas if started in recovery mode
+            if self.shared_storage_config.recovery_mode.is_some() {
                 continue;
             }
 
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -681,11 +681,14 @@ impl Collection {
                 continue;
             }
 
-            if !(this_peer_state == Some(Dead) && !replica_set.is_recovery().await) {
-                continue; // All good
+            #[allow(clippy::nonminimal_bool)]
+            {
+                if !(this_peer_state == Some(Dead) && replica_set.is_dirty().await) {
+                    continue; // All good
+                }
             }
 
-            // Reaches here only if replica is Dead but not in recovery mode
+            // Reaches here only if replica is Dead OR dirty
 
             // Try to find dead replicas with no active transfers
             let transfers = shard_holder.get_transfers(|_| true);
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -686,12 +686,11 @@ impl Collection {
                 continue;
             }
 
+            // Don't recover replicas if not dead
             let is_dead = this_peer_state == Some(Dead);
-            if !(is_dead || replica_set.is_dummy().await) {
-                continue; // All good
+            if !is_dead {
+                continue;
             }
-
-            // Reaches here only if replica is Dead OR dirty (dummy but not in recovery)
 
             // Try to find dead replicas with no active transfers
             let transfers = shard_holder.get_transfers(|_| true);
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -681,9 +681,11 @@ impl Collection {
                 continue;
             }
 
-            if this_peer_state != Some(Dead) || replica_set.is_dummy().await {
+            if this_peer_state != Some(Dead) {
                 continue; // All good
             }
+
+            // TODO: Skip transfers for dummy shards that aren't dirty
 
             // Try to find dead replicas with no active transfers
             let transfers = shard_holder.get_transfers(|_| true);
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -681,14 +681,18 @@ impl Collection {
                 continue;
             }
 
-            #[allow(clippy::nonminimal_bool)]
-            {
-                if !(this_peer_state == Some(Dead) || replica_set.is_dirty().await) {
-                    continue; // All good
-                }
+            // Don't recover our replicas if started in recovery mode
+            let is_recovery_mode = self.shared_storage_config.recovery_mode.is_some();
+            if is_recovery_mode {
+                continue;
             }
 
-            // Reaches here only if replica is Dead OR dirty
+            let is_dead = this_peer_state.is_none_or(|state| state == Dead);
+            if !(is_dead || replica_set.is_dummy().await) {
+                continue; // All good
+            }
+
+            // Reaches here only if replica is Dead OR dummy
 
             // Try to find dead replicas with no active transfers
             let transfers = shard_holder.get_transfers(|_| true);
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -685,8 +685,6 @@ impl Collection {
                 continue; // All good
             }
 
-            // TODO: Skip transfers for dummy shards that aren't dirty
-
             // Try to find dead replicas with no active transfers
             let transfers = shard_holder.get_transfers(|_| true);
 
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -686,12 +686,12 @@ impl Collection {
                 continue;
             }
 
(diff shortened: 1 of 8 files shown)


lib/collection/src/shards/replica_set/snapshots.rs
```diff
@@ -150,26 +150,7 @@ impl ShardReplicaSet {
                     "Failed to restore local replica",
                 )));
 
-                // Mark this peer as "locally disabled"...
-                //
-                // `active_remote_shards` includes `Active` and `ReshardingScaleDown` replicas!
-                let has_other_active_peers = self.active_remote_shards().is_empty();
```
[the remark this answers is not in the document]
timvisee: I agree this is wrong. Could you open a quick PR on dev first with a one line change, just to add the ! in front so we can merge that fix right away?

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -363,9 +370,34 @@ impl ShardReplicaSet {
...
         }
 
+        if is_dirty_shard {
+            replica_set.mark_local_as_dead();
+        }
+
         replica_set
     }
 
+    pub fn mark_local_as_dead(&self) {
```
timvisee, line 380: Yes, let's remove this and use `add_locally_disabled` instead.

lib/collection/src/collection/mod.rs
```diff
@@ -681,7 +681,7 @@ impl Collection {
                 continue;
             }
 
-            if this_peer_state != Some(Dead) || replica_set.is_dummy().await {
+            if !(this_peer_state == Some(Dead) || replica_set.is_dirty().await) {
```
timvisee, line 684: I wonder, we might always want to recover dummy shards here. In that case we can remove the dirty flag, and just use `is_dummy()`. If indeed possible, I'd prefer to remove it to keep our state simpler.

This might conflict with recovery mode though.

Let me have a bit of a thought about this.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -418,7 +420,15 @@ impl Collection {
...
+                // Check if shard was dirty before init_empty_local_shard
+                let was_dirty = replica_set.is_dirty().await;
+                // TODO: If dirty, still try to load the shard and init empty shard only if it's not recoverable?
                 replica_set.init_empty_local_shard().await?;
+
+                if was_dirty {
+                    let shard_flag = shard_initializing_flag_path(&collection_path, shard_id);
+                    tokio::fs::remove_file(&shard_flag).await?;
+                }
             }
```
timvisee, line 432: Yes, we should initialize an empty shard here as cannot do our recovery process on top of a corrupt shard, and so we must empty it here.

```suggestion
            if replica_set.is_dummy().await {
                // If shard was dirty, remove initializing flag after initializing empty
                let was_dirty = replica_set.is_dirty().await;
                replica_set.init_empty_local_shard().await?;
                if was_dirty {
                    let shard_flag = shard_initializing_flag_path(&collection_path, shard_id);
                    tokio::fs::remove_file(&shard_flag).await?;
                }
            }
```

It's fine to empty it here, because at this point we're sure that there's some other replica that still has our data. We therefore have no data loss.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -418,7 +420,15 @@ impl Collection {
...
 
             if replica_set.is_dummy().await {
+                // Check if shard was dirty before init_empty_local_shard
+                let was_dirty = replica_set.is_dirty().await;
+                // TODO: If dirty, still try to load the shard and init empty shard only if it's not recoverable?
                 replica_set.init_empty_local_shard().await?;
```
timvisee, line 426: But! Inside `init_empty_local_shard` we must do `LocalShard::clear()` to actually remove the corrupt files (if there are any). There's a TODO in there which is probably the right place to put it.

tests/consensus_tests/test_failed_snapshot_recovery.py
```diff
@@ -134,7 +134,8 @@ def test_failed_snapshot_recovery(tmp_path: pathlib.Path):
 
     # Assert storage contains initialized flag
     flag_path = shard_initializing_flag(peer_dirs[-1], COLLECTION_NAME, 0)
-    assert os.path.exists(flag_path)
+    flag_exists = os.path.exists(flag_path)
+    print("Shard initializing flag exists: {}".format(flag_exists)) # It may or may not exist because we recover very quickly
```
timvisee, line 138: Maybe we can make the recovery process longer?

tests/consensus_tests/test_failed_snapshot_recovery.py
```diff
@@ -159,15 +160,15 @@ def test_failed_snapshot_recovery(tmp_path: pathlib.Path):
...
     flag_path = shard_initializing_flag(peer_dirs[-1], COLLECTION_NAME, 0)
     assert not os.path.exists(flag_path)
 
     # Assert that the local shard is dead and empty
     local_shards = get_local_shards(peer_api_uris[-1])
     assert len(local_shards) == 1
     assert local_shards[0]["shard_id"] == 0
-    assert local_shards[0]["state"] == "Partial"
+    assert local_shards[0]["state"] == "Recovery"
     assert local_shards[0]["points_count"] == 0
```
timvisee, line 172: Lets assert it's either `Partial` OR `Recovery`. It can be either depending on what shard transfer method is selected. We cannot properly guarantee here that WAL delta is always used.

lib/collection/src/collection/mod.rs
KShivendu: > This might conflict with recovery mode though.

Exactly why I avoided it.

timvisee: We should probably disable shard transfers all together when in recovery mode.

If we do that, we can simply check `is_dummy()` here without problems.

lib/collection/src/shards/replica_set/snapshots.rs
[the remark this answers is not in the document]
KShivendu: Merged in https://github.com/qdrant/qdrant/pull/6297

lib/collection/src/collection/shard_transfer.rs
KShivendu: > because at this point we're sure that there's some other replica that still has our data. We therefore have no data loss

Because there was a shard initializing flag? Hmm, makes sense.

lib/collection/src/collection/shard_transfer.rs
KShivendu: Done inside `init_empty_local_shard()` since it was only used here.

tests/consensus_tests/test_failed_snapshot_recovery.py
KShivendu: It  will elongate tests quite a bit. Is there a way to slow down transfers with some env var?

tests/consensus_tests/test_failed_snapshot_recovery.py
```diff
@@ -170,28 +172,27 @@ def test_failed_snapshot_recovery(tmp_path: pathlib.Path):
...
-    assert not os.path.exists(flag_path)
-
     # Assert that the local shard is dead and empty
     local_shards = get_local_shards(peer_api_uris[-1])
     assert len(local_shards) == 1
     assert local_shards[0]["shard_id"] == 0
-    assert local_shards[0]["state"] == "Partial"
-    assert local_shards[0]["points_count"] == 0
+    assert local_shards[0]["state"] in ["Active", "Dead", "Partial", "Recovery"]
+    assert local_shards[0]["points_count"] <= n_points
```
timvisee, line 180: I think we can remove these assertions - just check for the flag and assert that it becomes active after.

tests/consensus_tests/test_failed_snapshot_recovery.py
```diff
@@ -269,7 +271,8 @@ def test_dirty_shard_handling_with_active_replicas(tmp_path: pathlib.Path, trans
     # Delete some of the files from the shard:
     shard_path = Path(peer_dirs[-1]) / "storage" / "collections" / COLLECTION_NAME / f"{shard_id}"
     assert shard_path.exists()
-    # corrupt_shard_dir(shard_path)
+
+    corrupt_shard_dir(shard_path)
```
timvisee, line 275: Inside this function: we cannot detect a removed segment in anyway. So let's not remove a segment, but do s till remove some WAL directory.

tests/consensus_tests/test_failed_snapshot_recovery.py
```diff
@@ -281,24 +284,31 @@ def test_dirty_shard_handling_with_active_replicas(tmp_path: pathlib.Path, trans
...
     # Wait for start of shard transfer
     wait_for_collection_shard_transfers_count(peer_api_uris[0], COLLECTION_NAME, 1)
 
+    # Assert that the local shard is in partial/recovery state
+    [local_shard] = get_local_shards(peer_api_uris[-1])
+    assert local_shard["shard_id"] == 0
+    shard_state = local_shard["state"]
+    assert shard_state in ["Active", "Dead", "Partial", "Recovery"]
+    assert local_shard["points_count"] <= n_points, local_shard
+
```
timvisee, line 298: I don't think asserting all these states makes sense.

lib/collection/src/collection/mod.rs
timvisee: I don't see checking a dirty flag as a stable way to see if recovery mode is enabled.

Instead we should explicitly check for recovery mode in the place where we drive transfers.

So I still prefer to:
- just check `is_dummy()` here
- explicitly check recovery mode

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -418,7 +420,24 @@ impl Collection {
...
+                    replica_set.shard_id,
+                    was_dirty
+                );
                 replica_set.init_empty_local_shard().await?;
+
+                if was_dirty {
+                    // TODO: Shouldn't we do this after we start the transfer?
+                    let shard_flag = shard_initializing_flag_path(&collection_path, shard_id);
+                    tokio::fs::remove_file(&shard_flag).await?;
+                }
```
timvisee, line 440: Let's remove the dirty flag and remove the initializing flag unconditionally (or more specifically, if it exists)

KShivendu:
> So I still prefer to:
> - just check is_dummy() here
> - explicitly check recovery mode


Dropped `is_dirty` bool from the DummyShard struct. But still keeping a `is_dirty()` in `ShardReplicaSet` since it keeps the code cleaner.

timvisee:
> > So I still prefer to:
> > 
> > * just check is_dummy() here
> > * explicitly check recovery mode
> 
> Okay. Dropped `is_dirty` bool from the DummyShard struct. But still keeping a `is_dirty()` in `ShardReplicaSet` since it keeps the code cleaner.

I don't agree. `sync_local_state` must never initiate a shard transfer if in recovery mode, and so we shall explicitly check it.

Then the dirty flag becomes obsolete.

Also, I think we should recover dummy replicas both if dead or not dead.

The dirty flag magically acting different based on external state is not obvious and a potential footgun. I strongly prefer explicit checks here.

lib/collection/src/collection/mod.rs
```diff
@@ -681,10 +681,19 @@ impl Collection {
...
             }
 
-            if this_peer_state != Some(Dead) || replica_set.is_dummy().await {
+            // Don't recover our replicas if started in recovery mode
+            let is_recovery_mode = self.shared_storage_config.recovery_mode.is_some();
+            if is_recovery_mode {
+                continue;
+            }
+
+            let is_dead = this_peer_state.is_none_or(|state| state == Dead);
```
KShivendu, line 690: When can the `this_peer_state` be `None`?

generall (approved):

PR #6295, opened 2025-04-02
Add ConditionedCounter

JojiiOfficial:
Adds a new wrapper for `HwCounterCell` that allows conditioned counting.
This is needed in places where we can't use a multiplier of 0 to disable counting, like mmap bool index.

An example of how this can be used can be found in #6296.

### DIFF
--- a/lib/common/common/src/counter/conditioned_counter.rs
+++ b/lib/common/common/src/counter/conditioned_counter.rs
@@ -1,0 +1,114 @@
+use std::ops::Deref;
+
+use super::hardware_counter::HardwareCounterCell;
+use super::hardware_data::HardwareData;
+
+/// A counter that measures or disposes measurements based on a condition.
+/// This is needed in places where we need to decide at runtime whether to measure or not.
+/// Implements `Deref<Target=HardwareCounterCell>` so it can be directly passed to functions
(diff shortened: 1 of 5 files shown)


lib/common/common/src/counter/conditioned_counter.rs
```diff
@@ -0,0 +1,114 @@
+use std::ops::Deref;
+
+use super::hardware_counter::HardwareCounterCell;
+use super::hardware_data::HardwareData;
+
+/// A counter that measures or disposes measurements based on a condition.
+/// This is needed in places where we need to decide at runtime whether to measure or not.
+/// Implements `Deref<Target=HardwareCounterCell>` so it can be directly passed to functions
+/// that need a `HardwareCounterCell` as parameter.
+pub struct ConditionedCounter<'a> {
```
timvisee, line 10: This does not seem to be used yet, or is it?

timvisee: Nevermind, I see it is used in a sub-PR: https://github.com/qdrant/qdrant/pull/6296

generall (approved):

agourlay (approved):

PR #6296, opened 2025-04-02
Only measure HW if on disk for bool index

JojiiOfficial:
Depends on #6295 

Only measure bool index io if index is on disk

### DIFF
--- a/lib/segment/src/index/field_index/bool_index/mmap_bool_index.rs
+++ b/lib/segment/src/index/field_index/bool_index/mmap_bool_index.rs
@@ -1,6 +1,7 @@
 use std::path::{Path, PathBuf};
 
 use bitvec::slice::BitSlice;
+use common::counter::conditioned_counter::ConditionedCounter;
 use common::counter::hardware_counter::HardwareCounterCell;
 use common::counter::iterator_hw_measurement::HwMeasurementIteratorExt;
 use common::types::PointOffsetType;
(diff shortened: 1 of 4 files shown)


lib/segment/src/index/field_index/bool_index/mmap_bool_index.rs
```diff
@@ -79,21 +81,32 @@ impl MmapBoolIndex {
...
         })
     }
 
+    fn make_conditioned_counter<'a>(
+        &self,
+        hw_counter: &'a HardwareCounterCell,
+    ) -> ConditionedCounter<'a> {
+        let on_disk = !self.populated; // Measure if on disk.
+        ConditionedCounter::new(on_disk, hw_counter)
+    }
```
timvisee, line 94: Can't we return a dummy counter here somehow, without a boolean condition inside the counter itself?

JojiiOfficial: We probably can but we'll use this in more places so I thought having a separate logic with known semantics makes the code more readable and easier to implement.

JojiiOfficial: Also there are places where we would run into lifetime issues with local disposable hardware counters.
An example for this is the `ChunkReader` in #6309

generall (approved):

PR #6298, opened 2025-04-02
Postprocess scores in max similarity

### LINKED ISSUE #6294: MAX_SIM is computed incorrectly with euclidean distance
## Current Behavior
Instead of summing up euclidean distances in MAX_SIM, qdrant is summing up negative squared distances
I believe, that after it accumulates `sum` of such distances, it just takes sqrt(abs(sum)).
This sum is not equal to a sum of euclidean distances, which leads to a wrong distance value.

## Steps to Reproduce
```python
from qdrant_client import QdrantClient, models

if __name__ == "__main__":
    cl = QdrantClient()
    cl.create_collection(
        "test",
        vectors_config=models.VectorParams(
            size=3,
            distance=models.Distance.EUCLID,
            multivector_config=models.MultiVectorConfig(
                comparator=models.MultiVectorComparator.MAX_SIM
            ),
        ),
    )
    cl.upsert(
        'test',
        points=[models.PointStruct(id=2, vector=[[3,3,3], [4,2,1]])]
    )
    print(cl.query_points('test', query=[[1,2,3], [3,3,3], [4,5,6]], with_vectors=True).points[0].score)
```
And the output is

`4.358899`

## Expected behaviour 
The same as in local mode (and here are some prints from local mode):

```python
Query multivec:
[[1. 2. 3.]
 [3. 3. 3.]
 [4. 5. 6.]]

Doc multivec:
[[3. 3. 3.]
 [4. 2. 1.]]

Distance matrix:
[[2.23606798 3.60555128]
 [0.         2.44948974]
 [3.74165739 5.83095189]]

Result:
5.977725364273731  # (2.23.. + 0. + 3.74..)
```

## Observations:

If we square the matrix values one by one, we'll get

```python
[[ 5. 13.]
 [ 0.  6.]
 [14. 34.]]
```

And sqrt(5+0+14) = sqrt(19) = ~4.36, which is exactly the score we're getting with hosted qdrant

agourlay:
This PR fixes https://github.com/qdrant/qdrant/issues/6294

The max similarity currently does not run the post-processing phase on the distance scoring.

This creates a discrepancy for the `Euclidean` metric where the processing is significant.

The proposed solution is to run the postprocessing step as part of the max similarity computation instead of at the very end.

I included a new benchmarks, the impact of this change is not visible for other distance metrics but Euclidean gets 20% slower on multivector.

TODO before shipping the PR:
- [ ] check Python congruence tests

### DIFF
--- a/lib/segment/src/vector_storage/query_scorer/mod.rs
+++ b/lib/segment/src/vector_storage/query_scorer/mod.rs
@@ -3,7 +3,7 @@ use common::types::{PointOffsetType, ScoreType};
 use crate::data_types::primitive::PrimitiveVectorElement;
 use crate::data_types::vectors::TypedMultiDenseVectorRef;
 use crate::spaces::metric::{Metric, MetricPostProcessing};
-use crate::types::{MultiVectorComparator, MultiVectorConfig};
+use crate::types::{MultiVectorComparator, MultiVectorConfig, Order};
 use crate::vector_storage::chunked_vector_storage::VectorOffsetType;
 
 pub mod custom_query_scorer;
@@ -38,13 +38,24 @@ pub fn score_max_similarity<
     debug_assert!(!multi_dense_a.is_empty());
     debug_assert!(!multi_dense_b.is_empty());
     let mut sum = 0.0;
+    let distance = TMetric::distance().distance_order();
+    let init_value = match distance {
+        Order::LargeBetter => ScoreType::NEG_INFINITY,
+        Order::SmallBetter => ScoreType::INFINITY,
+    };
+    let order = |a: ScoreType, b: ScoreType| -> bool {
+        match distance {
+            Order::LargeBetter => a > b,
+            Order::SmallBetter => a < b,
+        }
+    };
     for dense_a in multi_dense_a.multi_vectors() {
-        let mut max_sim = ScoreType::NEG_INFINITY;
+        let mut max_sim = init_value;
         // manual `max_by` for performance
         for dense_b in multi_dense_b.multi_vectors() {
             let sim = TMetric::similarity(dense_a, dense_b);
             let sim = TMetric::postprocess(sim);
-            if sim > max_sim {
+            if order(sim, max_sim) {
                 max_sim = sim;
             }
         }
@@ -140,6 +151,6 @@ mod tests {
         let b = MultiDenseVectorInternal::try_from(vec![vec![3.0, 3.0, 3.0], vec![4.0, 2.0, 1.0]])
             .unwrap();
         let score = score_max_similarity::<f32, EuclidMetric>((&a).into(), (&b).into());
-        assert_eq!(score, 11.885993);
+        assert_eq!(score, 5.9777255);
     }
 }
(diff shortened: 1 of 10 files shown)


lib/segment/src/vector_storage/query_scorer/mod.rs
```diff
@@ -40,6 +43,7 @@ pub fn score_max_similarity<T: PrimitiveVectorElement, TMetric: Metric<T>>(
         // manual `max_by` for performance
         for dense_b in multi_dense_b.multi_vectors() {
             let sim = TMetric::similarity(dense_a, dense_b);
+            let sim = TMetric::postprocess(sim);
```
agourlay, line 46: here is the postprocessing call

lib/segment/src/vector_storage/query_scorer/mod.rs
```diff
@@ -28,19 +28,34 @@ pub trait QueryScorer<TVector: ?Sized> {
...
+    let init_value = match distance {
+        Order::LargeBetter => ScoreType::NEG_INFINITY,
+        Order::SmallBetter => ScoreType::INFINITY,
+    };
+    let order = |a: ScoreType, b: ScoreType| -> bool {
+        match distance {
+            Order::LargeBetter => a > b,
+            Order::SmallBetter => a < b,
+        }
+    };
```
coszio, line 51: crutching further the unorthodox code path (euclidean with multivectors) 💀

coszio (approved):

joein (changes requested):
In case of euclidean distance the smaller the distance the better, however, it seems that here we have the opposite

Example:

```python
from qdrant_client import QdrantClient, models

if __name__ == "__main__":

    query_vector = [[1,2,3], [3,3,3], [4,5,6]]
    data_vector_0 = [[1,2,3], [3,3,3], [4,5,6]]
    data_vector_1 = [[3,3,3], [4,2,1],]


    for cl in QdrantClient(":memory:"), QdrantClient():
        if cl.collection_exists("test"):
            cl.delete_collection('test')
        cl.create_collection(
            "test",
            vectors_config=models.VectorParams(
                size=3,
                distance=models.Distance.EUCLID,
                multivector_config=models.MultiVectorConfig(
                    comparator=models.MultiVectorComparator.MAX_SIM
                ),
            ),
        )
        cl.upsert(
            'test',
            points=[models.PointStruct(id=0, vector=data_vector_0), models.PointStruct(id=1, vector=data_vector_1)],
        )
        print([point.score for point in cl.query_points('test', query=query_vector, with_vectors=True).points])
```

Output:
```
[0.0, 5.977725364273731]  # local mode
[5.9777255, 0.0]  # server
```

`data_vector_0` has a score of 0.0 (since it is the same vector as a query vector), and it should be the first one in the response

agourlay:
I decided to not invest more time on this issue :no_good_man: 

The score values in `dev` are not identical to the Colbert paper but the ordering is correct for the users.
Fixing the initial issue introduces a large performance regression and a significant amount of complexity in managing sort orders in various location in the code base.

Therefore I am proposing  to close this PR and change the local mode implementation to fit the server side.

I will capture the tests, benchmark and additional comments about our learning in a different PR.

PR #6308, opened 2025-04-02
[recommend avg_vector] combine multivectors into a single averaged vector

coszio:
This PR changes the implementation from concatenating all the multivectors in the input, to averaging them into a single one, same as we do for dense vectors.

For example , instead of combining `[ [],[] ]` and `[ [],[],[] ]` into a `[ [],[],[],[],[] ]` multivector, we now average all the individual vectors and produce an averaged multivector of just one vector: `[ [] ]`.

The idea behind this is that average vector strategy should combine all examples into a single one for a more optimized search. In the case of multivectors, we will still do MaxSim against the documents, but now only with a single vector in the multivector query.

### DIFF
--- a/lib/collection/src/recommendations.rs
+++ b/lib/collection/src/recommendations.rs
@@ -500,38 +500,64 @@ fn recommend_by_custom_score(
 
 #[cfg(test)]
 mod tests {
-    use segment::data_types::vectors::{VectorInternal, VectorRef};
+    use segment::data_types::vectors::{VectorInternal, VectorInternalDiscriminants, VectorRef};
     use sparse::common::sparse_vector::SparseVector;
+    use strum::IntoEnumIterator;
 
     use super::avg_vectors;
 
+    fn check_avg_vectors(variant: VectorInternalDiscriminants) {
+        match variant {
+            VectorInternalDiscriminants::Dense => {
+                let vectors: Vec<VectorInternal> = vec![
+                    vec![1.0, 2.0, 3.0].into(),
+                    vec![1.0, 2.0, 3.0].into(),
+                    vec![1.0, 2.0, 3.0].into(),
+                ];
+                assert_eq!(
+                    avg_vectors(vectors.iter().map(VectorRef::from)).unwrap(),
+                    vec![1.0, 2.0, 3.0].into(),
+                );
+            }
+            VectorInternalDiscriminants::Sparse => {
+                let vectors: Vec<VectorInternal> = vec![
+                    SparseVector::new(vec![0, 1, 2], vec![0.0, 0.1, 0.2])
+                        .unwrap()
+                        .into(),
+                    SparseVector::new(vec![0, 1, 2], vec![0.0, 1.0, 2.0])
+                        .unwrap()
+                        .into(),
+                ];
+                assert_eq!(
+                    avg_vectors(vectors.iter().map(VectorRef::from)).unwrap(),
+                    SparseVector::new(vec![0, 1, 2], vec![0.0, 0.55, 1.1])
+                        .unwrap()
+                        .into(),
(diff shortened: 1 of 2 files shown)


lib/collection/src/recommendations.rs
```diff
@@ -36,10 +36,10 @@ fn avg_vectors<'a>(
 ) -> CollectionResult<VectorInternal> {
     let mut avg_dense = DenseVector::default();
     let mut avg_sparse = SparseVector::default();
-    let mut avg_multi: Option<TypedMultiDenseVector<VectorElementType>> = None;
+    let mut avg_multi: Option<DenseVector> = None;
```
agourlay, line 39: I think we want to keep the existing behavior for single multivector.
That is not average it at all.

coszio: Ah, I see, I forgot we don't do `pos + pos - neg` if there are no negatives.
I will change it so that single multivector is just passed as-is

coszio: https://github.com/qdrant/qdrant/pull/6308/commits/029a30b039eedb65c71a98a86d4b4365943452a7

agourlay: I'd love to have a short unit test showing the new strategy in action.
One positive example, one negative example etc... :)

agourlay:
Before merging this we need to make sure the congruence tests are green for single example tests.
The implementation in local mode will have to be adapted for the multi vector case.

lib/collection/src/recommendations.rs
coszio: in https://github.com/qdrant/qdrant/pull/6308/commits/0aa7f77ef74b55b1c27d98df9c38dafae25547da I added a test to make sure that the averaging is the same as for dense, but flattened

lib/collection/src/recommendations.rs
```diff
@@ -118,10 +131,17 @@ fn merge_positive_and_negative_avg(
...
-        (VectorInternal::MultiDense(mut positive), VectorInternal::MultiDense(negative)) => {
-            // merge positive and negative vectors as concatenated vectors with negative vectors negated
-            positive.flattened_vectors.extend(negative.flattened_vectors.into_iter().map(|x| -x));
-            Ok(VectorInternal::MultiDense(positive))
+        (VectorInternal::MultiDense(positive), VectorInternal::MultiDense(negative)) => {
+            // averaged positive and negative multivectors consist of just one vector each
+            debug_assert_eq!(positive.len(), 1);
+            debug_assert_eq!(negative.len(), 1);
+
+            let vector = positive.flattened_vectors.iter().zip(negative.flattened_vectors.iter()).map(|(pos, neg)| pos + pos - neg).collect();
```
joein, line 139: Just to clarify

If multivectors have different number of vectors, then we take only `len(short_multivector)` vectors from both multivectors and dropping the remainder?

lib/collection/src/recommendations.rs
```diff
@@ -56,22 +56,28 @@ fn avg_vectors<'a>(
...
+                    in_multi_count += 1;
+                    avg_multi = Some(avg_multi.map_or_else(
+                        || vector.to_owned(),
+                        |mut avg_multi| {
+                            for i in 0..vector.len() {
+                                if i >= avg_multi.len() {
+                                    avg_multi.push(vector[i])
+                                } else {
+                                    avg_multi[i] += vector[i];
+                                }
```
joein, line 70: the second dimensionality in multivectors is configured on a collection level, it should not be possible to get into else branch

coszio:
Update, we just had a discussion where we resolved that the original way of concatenating all multivectors made the most sense. I will close this PR for now

PR #6309, opened 2025-04-02
Don't measure in-memory hw for full text indices

JojiiOfficial:
Depends on #6296


lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
```diff
@@ -121,7 +122,7 @@ impl MmapPostings {
             chunks,
             data,
             remainder_postings,
-            hw_counter,
+            ConditionedCounter::always(hw_counter),
```
generall, line 125: This shoudl depend on `populate` parameter on initialization

JojiiOfficial: Good catch, thanks!

timvisee (approved):

generall (approved):

PR #6310, opened 2025-04-02
Add a tool to view percentage of mmaps cache

generall:
A tool which can be used to see mmaps disk cache inside running qdrant process.

```
Usage examples:
  # Basic usage - show aggregated cache percentages for all vector index files
  python measure.py $(pidof qdrant) '.*/vector_index/.*'

  # Verbose mode - show individual files larger than 1MB
  python measure.py $(pidof qdrant) '.*/vector_storage/.*' -v

Example output:
    Cache percentages for pattern '.vector_storage.*':
    r--s: 0.00%
    rw-s: 0.06%
```

### DIFF
--- a/tools/smaps-view.py
+++ b/tools/smaps-view.py
@@ -1,0 +1,194 @@
+"""
+Usage examples:
+  # Basic usage - show aggregated cache percentages for all vector index files
+  python measure.py $(pidof qdrant) '.*/vector_index/.*'
+
+  # Verbose mode - show individual files larger than 1MB
+  python measure.py $(pidof qdrant) '.*/vector_storage/.*' -v
+
+Example output:
+    Cache percentages for pattern '.vector_storage.*':
+    r--s: 0.00%
+    rw-s: 0.06%
+"""
+
+import re
+import sys
+import argparse
+from typing import Dict, Tuple, List
+from dataclasses import dataclass
+
+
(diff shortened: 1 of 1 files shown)


tools/smaps-view.py
```diff
@@ -0,0 +1,194 @@
+"""
+Usage examples:
+  # Basic usage - show aggregated cache percentages for all vector index files
+  python smaps-view.py $(pidof qdrant) '.*/vector_index/.*'
+
+  # Verbose mode - show individual files larger than 1MB
+  python smaps-view.py $(pidof qdrant) '.*/vector_storage/.*' -v
+
```
KShivendu, line 8:
```suggestion
  # Verbose mode - show vector storage files larger than 1MB
  python smaps-view.py $(pidof qdrant) '.*/vector_storage/.*' -v

  # Verbose mode - show all files larger than 1MB
  python smaps-view.py $(pidof qdrant) '.*' -v
```

I tested on a small collection. Needed `python smaps-view.py $(pidof qdrant) '.*' -v` to see file stats. Might be better to highlight that in docs.

KShivendu (approved):

tools/smaps-view.py
generall: what is the point? This "match_anything" pattern will include a lot of garbage

PR #6314, opened 2025-04-03
Pytest run failing first

KShivendu:
This would speed up development cycle.

### All Submissions:


.github/workflows/integration-tests.yml
```diff
@@ -69,9 +69,20 @@ jobs:
       - name: Run integration tests - 1 peer
         run: poetry -C tests run ./tests/integration-tests.sh distributed
         shell: bash
+      - name: Restore pytest cache
+        uses: actions/cache@v3
+        with:
+          path: .pytest_cache
+          key: pytest-cache-${{ github.run_id }}
+          restore-keys: pytest-cache-
```
timvisee, line 77: In [CI documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#cache-hits-and-misses) I read this:

> `restore-keys`: (...)  
If there are no exact matches, the action searches for partial matches of the restore keys. When the action finds a partial match, the most recent cache is restored to the path directory.

Would it make sense to also include the current branch name in this cache key? That may prevent taking caches from other branches/PRs that are unrelated.

timvisee (approved):

agourlay:
> This would speed up development cycle.

Do you have performance number to back this up?

KShivendu:
The caching didn't work as expected. I'll get back to this soon. Have higher priority tasks to wrap up.

timvisee:
I'd also like to see some performance number for this one.

Since we now include the branch name in the cache key, does it mean that the initial CI run on a PR might take much longer?

PR #6316, opened 2025-04-03
Fix measuring in memory inverted index

JojiiOfficial:
Depends on #6311

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mod.rs
+++ b/lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mod.rs
@@ -283,11 +283,11 @@ impl InvertedIndex for MmapInvertedIndex {
     }
 
     fn get_token_id(&self, token: &str, hw_counter: &HardwareCounterCell) -> Option<TokenId> {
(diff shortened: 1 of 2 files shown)


lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mod.rs
```diff
@@ -273,6 +283,8 @@ impl InvertedIndex for MmapInvertedIndex {
     }
 
     fn get_token_id(&self, token: &str, hw_counter: &HardwareCounterCell) -> Option<TokenId> {
+        let hw_counter = self.make_conditioned_counter(hw_counter);
```
generall, line 286: too much overhead for such a simple function, isn't it?

generall: I woulf rather have `if self.is_on_disk` for such cases

JojiiOfficial: Yeah good point. Fixed!

generall (approved):

PR #6317, opened 2025-04-03
[strict mode | score boosting] Strict mode for formula queries

coszio:
Implements the check for unindexed fields in formula expressions.

Even if formulas are only allowed as a reranking step, there is no guarantee that the preselection will be small enough for making reading payloads directly negligible. 

With this check, only correctly-indexed payload fields will be allowed in the formula expressions.

A needed change was to move the `problems` module from `segment` to the `collection` crate, because we needed to evaluate variables inside `ExpressionInternal`.

Also realized that we could use `JsonPath` directly as the type for REST's `datetime_key` expression.


lib/collection/src/operations/verification/query.rs
```diff
@@ -105,6 +106,35 @@ impl StrictModeVerification for QueryGroupsRequestInternal {
 }
 
 impl StrictModeVerification for CollectionQueryRequest {
+    async fn check_custom(
+        &self,
+        collection: &Collection,
+        strict_mode_config: &StrictModeConfig,
+    ) -> CollectionResult<()> {
+        // check for unindexed fields in formula
+        if strict_mode_config.unindexed_filtering_retrieve == Some(false) {
```
JojiiOfficial, line 115: This and `CollectionQueryGroupsRequest` `check_customt` function seems duplicated code. Can we extract it into a separate function?

JojiiOfficial (approved):

lib/collection/src/operations/verification/query.rs
coszio: it is duplicated indeed

coszio: https://github.com/qdrant/qdrant/pull/6317/commits/70cad84e1329500a82f2d62046cc5ad07707040e

agourlay:
Thank you very much for adding this :+1: 

Please add an integration test in `tests/openapi/test_strictmode.py`

agourlay (approved):

PR #6319, opened 2025-04-03
Only implement verification for `CollectionQuery` types

coszio:
I noticed there was duplication in the implementation of `StrictModeVerification` trait for query request types.

We already have a converging type for rest and grpc for query requests, so we can implement the trait only for this converging type.

By unifying, this also fixes a hole where we were not checking the prefetches when using the grpc interface.


JojiiOfficial (approved):
Thanks!

agourlay (approved):
:trophy:

PR #6323, opened 2025-04-04
disk cache hygiene

generall:
Done:

- explicit functions for populate and clear mmap cache for: Vector storage, Vector Index, Payload Storage, Payload Index, QuantizedVectorStorage
- Remove delayed mmap population, as it seems it pollutes cache
- Implement cache clearing for dropping segment

How it was tested:

1. Create collection with bfb:

```
docker run --rm -it --network=host qdrant/bfb:dev ./bfb -n 500000 --segments 20 --quantization scalar --quantization-in-ram true -d 128 --indexing-threshold 100 --on-disk-vectors true --on-disk-payload
```

2. Reload service with clear caches and mem limit:

```
sudo bash -c 'sync; echo 3 > /proc/sys/vm/drop_caches'
docker run --rm -it -m 600Mb --network=host -v $(pwd)/qdrant-storage:/qdrant/storage qdrant-cache-test
```

(here caches are full)

```
sudo python3 smaps-view.py $(pidof qdrant) '.*/vector_index/.*' -v
```

3. Run single-threaded indexation:

```
PATCH collections/benchmark
{
  "hnsw_config": {
    "ef_construct": 101
  },
  "optimizers_config": {
    "max_optimization_threads": 1
  }
}
```

4. Wait indexation finish, observe loss of disk cache:

```
sudo python3 smaps-view.py $(pidof qdrant) '.*/vector_index/.*' -v

> Cache percentages for pattern '.*/vector_index/.*':
>  r--s: 33.23%
```

With this PR cache is not lost.

### DIFF
--- a/lib/common/memory/src/madvise.rs
+++ b/lib/common/memory/src/madvise.rs
@@ -193,4 +193,6 @@ pub fn clear_disk_cache(file_path: &Path) -> io::Result<()> {
             Ok(())
         }
     }
+    #[cfg(not(target_os = "linux"))]
+    Ok(())
 }
--- a/lib/common/memory/src/madvise.rs
+++ b/lib/common/memory/src/madvise.rs
@@ -175,7 +175,7 @@ fn populate_simple(slice: &[u8]) {
 ///
 /// If `posix_fadvise` is not supported, this function does nothing.
 pub fn clear_disk_cache(file_path: &Path) -> io::Result<()> {
-    #[cfg(unix)]
+    #[cfg(target_os = "linux")]
     {
         use std::os::unix::io::AsRawFd;
 
--- a/lib/common/memory/src/madvise.rs
+++ b/lib/common/memory/src/madvise.rs
@@ -1,6 +1,7 @@
 //! Platform-independent abstractions over [`memmap2::Mmap::advise`]/[`memmap2::MmapMut::advise`]
 //! and [`memmap2::Advice`].
 
+use std::fs::File;
 use std::hint::black_box;
 use std::io;
 use std::num::Wrapping;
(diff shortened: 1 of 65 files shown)


lib/common/memory/src/madvise.rs
```diff
@@ -169,3 +170,29 @@ fn populate_simple(slice: &[u8]) {
...
+
+        if ret != 0 {
+            Err(io::Error::from_raw_os_error(ret))
+        } else {
+            Ok(())
+        }
+    }
+    #[cfg(not(target_os = "linux"))]
+    Ok(())
+}
```
xzfc, line 198: We have a similar function here:
https://github.com/qdrant/qdrant/blob/4db98ecdbf66bedbce268ec26865ea446d0bba7b/lib/common/io/src/file_operations.rs#L34-L35

generall: `advise_dontneed` doesn't seem to be used anywhere. I propose to remove it and keep a new one.

lib/segment/src/vector_storage/quantized/quantized_mmap_storage.rs
```diff
@@ -2,12 +2,19 @@ use std::path::Path;
...
 #[derive(Debug)]
 pub struct QuantizedMmapStorage {
     mmap: Mmap,
 }
 
+impl QuantizedMmapStorage {
+    pub fn populate(&self) {
+        self.mmap.populate();
+    }
+}
```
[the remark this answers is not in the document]
generall: This is done like this because Clippy was complaining

lib/common/memory/src/madvise.rs
xzfc: Then I propose to copy the old definition:
- It uses safe wrapper `nix::fcntl` instead of unsafe `libc` call.
- It works on any platform that support `posix_fadvise`, not just Linux.

generall: Done

timvisee (approved):

PR #6325, opened 2025-04-04
Incremental HNSW index building: append-only case 

xzfc:
This PR implements incremental HNSW building for the append-only case.

A new feature flag is added, `incremental_hnsw_building`, disabled by default.

When enabled, it will try to reuse old graph as a starting point for building the new graph. When reusing, point levels and old graph links are copied as-is (but with reassigned IDs) before the regular HNSW point insertion process starts.

What is not supported as of now:
- GPU building. Only point levels are copied.
- Point/vector removal/overwriting. If a point or vector is removed, the old graph is discarded. Same if vector value is changed.

When old index is reused, the following log messages are printed:
```
DEBUG segment::index::hnsw_index::hnsw: building HNSW for 518500 vectors with 8 CPUs
DEBUG segment::index::hnsw_index::hnsw: Reusing 457700 points from the old index
```

Benchmarked using `bfb --indexing-threshold 400 --dim 32 --num-vectors 2000000 --throttle 500`.
Upload time: 40 seconds.
Upload+index (incremental): 40+17 = 57 seconds.
Upload+index (non-incremental): 40+84 = 124 seconds.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
@@ -403,12 +403,40 @@ impl GraphLayersBuilder {
             // New point is a new empty entry (for this filter, at least)
             // We can't do much here, so just quit
         }
-        self.ready_list.write().set(point_id as usize, true);
+        let was_ready = self.ready_list.write().replace(point_id as usize, true);
+        assert!(!was_ready);
         self.entry_points
             .lock()
             .new_point(point_id, level, |point_id| {
                 points_scorer.check_vector(point_id)
             });
+    }
+
+    /// Add a new point using pre-existing links.
+    /// Mutually exclusive with [`Self::link_new_point`].
+    pub fn add_new_point(&self, point_id: PointOffsetType, levels: Vec<Vec<PointOffsetType>>) {
+        let level = self.get_point_level(point_id);
+        assert_eq!(levels.len(), level + 1);
+
+        for (level, neighbours) in levels.iter().enumerate() {
+            let mut links = self.links_layers[point_id as usize][level].write();
+            links.clear();
+            links.extend_from_slice(neighbours);
+        }
+
+        let was_ready = self.ready_list.write().replace(point_id as usize, true);
+        assert!(!was_ready);
+        self.entry_points
+            .lock()
+            .new_point(point_id, level, |_| true);
+    }
+
+    pub fn link_new_point_complete_indexing(&self, point_id: PointOffsetType) {
+        let level = self.get_point_level(point_id);
+        self.ready_list.write().set(point_id as usize, true);
+        self.entry_points
+            .lock()
+            .new_point(point_id, level, |_| true);
     }
 
     /// Link a new point on a specific level.
--- a/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
@@ -431,14 +431,6 @@ impl GraphLayersBuilder {
             .new_point(point_id, level, |_| true);
     }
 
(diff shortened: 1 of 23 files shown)


generall:
@xzfc please resolve conflict

timvisee (approved):

timvisee:
I was only able to partially test this locally. So thanks for adding those tests! The results actually look fantastic. I wonder how this would affect bulk upsertions without disabling HNSW building.

lib/segment/src/index/hnsw_index/graph_layers_builder.rs
```diff
@@ -403,14 +403,42 @@ impl GraphLayersBuilder {
             // New point is a new empty entry (for this filter, at least)
             // We can't do much here, so just quit
         }
-        self.ready_list.write().set(point_id as usize, true);
+        let was_ready = self.ready_list.write().replace(point_id as usize, true);
+        assert!(!was_ready);
```
agourlay, line 407: debug assert or do we want to crash here?

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -239,6 +241,19 @@ impl HNSWIndex {
             total_vector_count,
         );
 
+        let old_index = old_indices
+            .iter()
+            .filter_map(|old_index| {
+                feature_flags.incremental_hnsw_building.then_some(())?;
```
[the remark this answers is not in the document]
timvisee: Though it's cheap (just a `bool`) I agree it might be better to not iterate at all if the flag is not set.

> Can it be changed without restarting the service?

No, at least not in the duration of this function.

agourlay:
> A new feature flag is added, incremental_hnsw_building, disabled by default.

Where you able to run the all CI suite with the flag enabled?

lib/segment/src/index/hnsw_index/hnsw.rs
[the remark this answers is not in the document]
xzfc: It's optimized for easy removal. We might want make it unconditional soon. So, removing this check would be an one-line change.

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -349,13 +370,23 @@ impl HNSWIndex {
...
+            } else {
+                ids.extend(id_tracker_ref.iter_ids_excluding(deleted_bitslice));
+            }
 
-            let first_few_ids: Vec<_> = ids_iterator
-                .by_ref()
-                .take(SINGLE_THREADED_HNSW_BUILD_THRESHOLD)
-                .collect();
-            let ids: Vec<_> = ids_iterator.collect();
+            let first_few_ids = ids.split_off(
```
generall, line 386: `first_few_ids` is only relevant if the graph is empty. Potentially we can skip this step if we have already added enough points with `add_new_point`

lib/segment/src/index/hnsw_index/graph_layers_builder.rs
xzfc: Replaced with `debug_assert`.

lib/segment/src/index/hnsw_index/hnsw.rs
xzfc: Updated.

generall (approved):
LGTM, please make sure to enable immediate indexing on CI-benchmarks

xzfc:
@agourlay
> Where you able to run the all CI suite with the flag enabled?

Checked in a temporary branch (look for `17 / 17` in the GitHub UI to see CI): https://github.com/qdrant/qdrant/commit/42cfdb10d905646f7e7f51bfb51f77b375d433fb

Will enable it by default in a separate PR.

PR #6327, opened 2025-04-04
Rate limit retrieve based on number of ids

agourlay:
Large retrieve operation should consume more tokens in the rate limiter.

### DIFF
--- a/tests/openapi/test_strictmode.py
+++ b/tests/openapi/test_strictmode.py
@@ -1530,7 +1530,6 @@ def test_strict_mode_retrieve_read_rate_limiting(collection_name):
         path_params={"collection_name": collection_name},
         body={
             "ids": [1, 2, 3, 4, 5],
-
         },
     )
     assert response.ok, response.text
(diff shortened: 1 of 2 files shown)


tests/openapi/test_strictmode.py
```diff
@@ -1521,3 +1521,58 @@ def test_strict_mode_recommendation_best_score_read_rate_limiting(collection_nam
...
+        method="POST",
+        path_params={"collection_name": collection_name},
+        body={
+            "ids": [1, 2, 3, 4, 5],
+        },
+    )
+    assert response.status_code == 429
+    assert "Read rate limit exceeded, request larger than rate limiter capacity, please try to split your request" in response.json()['status']['error']
+
+    set_strict_mode(collection_name, {
```
JojiiOfficial, line 1566: Suggestion:
We're already testing that disabling rate limit aligns with disabling strict mode and here we're testing for the same logic again. Maybe we can test for an allowed number of IDs instead, wdyt?

JojiiOfficial (approved):

PR #6330, opened 2025-04-07
Enable integer overflow checks in perf mode

agourlay:
We often use the `perf` mode for local performance testing.

Enabling integer overflow check on this profile is a low hanging fruit to catch issues that cannot be observed in release mode.

Rust will panic in debug mode in case of overflow, but in release mode it will silently wrap around.

On a quick and dirty local test it seems to be more than 10% slower (bfb upsert) 

source: https://corrode.dev/blog/pitfalls-of-safe-rust/#protect-against-integer-overflow

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -282,6 +282,7 @@ inherits = "release"
 lto = false
 opt-level = 3
 codegen-units = 256 # restore default value for faster compilation
+overflow-checks = true # Enable integer overflow checks
 
 # Override for fast sha256 hashing in dev builds
 [profile.dev.package.sha2]


agourlay:
@timvisee What do you think about this? I know you are a big user of `perf` mode. 
I am not super happy about the slowdown, maybe you get better numbers than me :man_shrugging:

timvisee:
@agourlay Based on your measurement I don't think we should merge this.

A 10% slowdown seems rather significant. The main goal of `perf` is to be as close as possible to a release build, but without the very lengthy lto step.

Since the overflow check is enabled in debug builds and in all tests, I'm not sure it's really necessary here.

Thank you for giving it a shot though! Thoughts? :smiley:

PR #6349, opened 2025-04-08
refactor `QueryEnum::Nearest` for using `NamedQuery`

coszio:
Homogenizes the `QueryEnum` enum for using `NamedQuery<T>` for all variants. The only different variant was `Nearest`. Essentially this PR makes this change:

```diff
pub enum QueryEnum {
-   Nearest(NamedVectorStruct),
+   Nearest(NamedQuery<VectorInternal>),
    RecommendBestScore(NamedQuery<RecoQuery<VectorInternal>>),
    RecommendSumScores(NamedQuery<RecoQuery<VectorInternal>>),
    Discover(NamedQuery<DiscoveryQuery<VectorInternal>>),
    Context(NamedQuery<ContextQuery<VectorInternal>>),
}
```

PR #6350, opened 2025-04-08
[strict mode] verify number and size of vectors in query

coszio:
Builds on top of #6349 for easier handling of `VectorInternal`

This PR:
- enforces `max_vectors` for each multivector in the query
- enforces `max_length` for each sparse vector in the query
- introduces and enforces `max_query_vectors`, which counts the total amount of vectors in a query
  - each vector in a multivector is accounted individually
  - covers all vector inputs in special queries like recommend, discover, context, etc.


coszio:
After an in-call discussion, this same logic can be applied to the rate-limiter, which will cover the validation.

Closing this PR, but will repurpose the changes for using them within rate-limiter

PR #6351, opened 2025-04-09
Strict mode prevents fullscan for multivector

agourlay:
This PR makes sure the multivector search will not use the full scan strategy as it is very expensive.

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -933,6 +933,11 @@ impl HNSWIndex {
         params: Option<&SearchParams>,
         vector_query_context: &VectorQueryContext,
     ) -> OperationResult<Vec<ScoredPointOffset>> {
+        if vector_query_context.strict_mode_enabled() && vector.is_multivector() {
+            return Err(OperationError::validation_error(
+                "Multivector full scan search not allowed in strict mode",
+            ));
+        }
         let id_tracker = self.id_tracker.borrow();
         let vector_storage = self.vector_storage.borrow();
         let quantized_vectors = self.quantized_vectors.borrow();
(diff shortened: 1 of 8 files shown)


lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -939,6 +939,11 @@ impl HNSWIndex {
         params: Option<&SearchParams>,
         vector_query_context: &VectorQueryContext,
     ) -> OperationResult<Vec<ScoredPointOffset>> {
+        if vector_query_context.strict_mode_enabled() && vector.is_multivector() {
+            return Err(OperationError::validation_error(
+                "Multivector full scan search not allowed in strict mode",
+            ));
+        }
```
timvisee, line 946: Can we move both of these if-statements into the shared `search_plain_iterator` instead?

tests/openapi/test_strictmode.py
```diff
@@ -1630,3 +1647,109 @@ def scroll_request(condition_count: int):
...
+
+
+def test_strict_mode_fullscan_multivector(collection_name):
+    # Clear collection to not depend on other tests
+    response = request_with_validation(
+        api='/collections/{collection_name}',
+        method="DELETE",
+        path_params={'collection_name': collection_name},
+    )
+    assert response.ok
```
timvisee, line 1659: How about just using a different collection name?

timvisee (commented):
Thanks for the test!

Please see my review remarks below:

tests/openapi/test_strictmode.py
agourlay: If I create a collection with a different name, it won't be automatically deleted at the end of test (success or failure).

timvisee: We can add it in the fixture up-top:

```
@pytest.fixture(autouse=True)
def setup(collection_name):
    basic_collection_setup(collection_name=collection_name)
    yield
    drop_collection(collection_name=collection_name)
```

agourlay: The test uses multivectors which are not setup by `basic_collection_setup`.
I think I am missing something here, what is your concern? :)

generall (changes requested):
I don't think this is a correct understanding of the original problem.

Original problem is: we don't cover all cases if `search_allow_exact: false` is enabled.
We do NOT want to disable full-scan specifically for multi-vectors if strict mode is enabled.

Multivectors are only related to this problem as an example of a use-case, where we disable global HNSW index, cause we only expect multivectors to be used as rescoring part of query. But in general, it is applicable for all kinds of vectors (and we can assume that sparse vectors are always indexed).

PR #6356, opened 2025-04-09
[rate limits] consider length of multivectors for query cost

coszio:
Considers the size of multivectors and total amount of examples in special queries like recommend, discover, etc. in the cost of a single core_search. 

This effectively limits requests that would be too expensive according to the rate-limiter setting.

### DIFF
--- a/lib/segment/src/data_types/vectors.rs
+++ b/lib/segment/src/data_types/vectors.rs
@@ -16,7 +16,7 @@ use crate::types::{VectorName, VectorNameBuf};
 use crate::vector_storage::query::{ContextQuery, DiscoveryQuery, RecoQuery, TransformInto};
 
 /// How many dimensions of a sparse vector are considered to be a single unit for cost estimation.
-const SPARSE_VECTOR_COST: usize = 64;
+const SPARSE_DIMS_COST_UNIT: usize = 64;
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum VectorInternal {
@@ -30,9 +30,7 @@ impl VectorInternal {
     pub fn similarity_cost(&self) -> usize {
         match self {
             VectorInternal::Dense(_dense) => 1,
-            // TODO(ratelimits): Currently one sparse vector counts as one, regardless of number of dimensions.
-            //                   We should come up with a formula for adjusting its cost.
-            VectorInternal::Sparse(sparse) => sparse.indices.len().div_ceil(SPARSE_VECTOR_COST),
+            VectorInternal::Sparse(sparse) => sparse.indices.len().div_ceil(SPARSE_DIMS_COST_UNIT),
             VectorInternal::MultiDense(multivec) => multivec.len(),
         }
     }
--- a/lib/segment/src/data_types/vectors.rs
+++ b/lib/segment/src/data_types/vectors.rs
@@ -22,6 +22,19 @@ pub enum VectorInternal {
     MultiDense(MultiDenseVectorInternal),
 }
 
+impl VectorInternal {
+    /// Returns the estimated cost of using this vector in terms of the number of how many similarity comparisons vector will make against one point.
+    pub fn similarity_cost(&self) -> usize {
+        match self {
+            VectorInternal::Dense(_dense) => 1,
+            // TODO(ratelimits): Currently one sparse vector counts as one, regardless of number of dimensions.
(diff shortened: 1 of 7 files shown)


agourlay:
This needs an integration test to validate that the rate limiter is now triggered.
The multivector case should be enough.

tests/openapi/test_strictmode.py
```diff
@@ -1631,100 +1637,76 @@ def scroll_request(condition_count: int):
     assert response.status_code == 429
     assert "Read rate limit exceeded: Operation requires 7 tokens" in response.json()['status']['error']
 
+def test_read_rate_limiter_many_vectors(full_collection_name):
+    collection_name = full_collection_name
 
-def test_strict_mode_group_limits(collection_name):
```
agourlay, line 1635: Did you just nuke my tests? :smile:

coszio: Sorry, looks like I mishandled conflicts when rebasing

coszio: https://github.com/qdrant/qdrant/pull/6356/commits/d2cf3856673c1bce700128b31b96efa74f3b6f50

lib/segment/src/data_types/vectors.rs
```diff
@@ -15,13 +15,29 @@ use crate::common::utils::transpose_map_into_named_vector;
...
 }
 
+impl VectorInternal {
+    /// Returns the estimated cost of using this vector in terms of the number of how many similarity comparisons vector will make against one point.
+    pub fn similarity_cost(&self) -> usize {
+        match self {
+            VectorInternal::Dense(_dense) => 1,
+            // TODO(ratelimits): Currently one sparse vector counts as one, regardless of number of dimensions.
+            //                   We should come up with a formula for adjusting its cost.
+            VectorInternal::Sparse(sparse) => 1.max(sparse.indices.len() / SPARSE_VECTOR_COST),
```
timvisee, line 35: How about [`div_ceil`](https://doc.rust-lang.org/std/primitive.usize.html#method.div_ceil)?

```suggestion
            VectorInternal::Sparse(sparse) => sparse.indices.len().div_ceil(SPARSE_VECTOR_COST),
```

| Size | Cost max/div | Cost div_ceil | Comment |
|---:|---:|---:|:---|
| 0 | 1 | 0 | Should not be possible, cost 0 seems fine
| 1 | 1 | 1 |
| 64 | 1 | 1 |
| 65 | 1 | 2 | I think 2 is better than 1
| 128 | 2 | 2 |
| 129 | 2 | 3 | I think 3 is better than 2

tests/openapi/test_strictmode.py
```diff
@@ -1727,4 +1733,79 @@ def test_strict_mode_distance_matrix_limits(collection_name):
...
+                "limit": 5
+            }
+        )
+        check_response(search_response, should_succeed)
+
+    # check without strict mode
+    check_multivector_query_raw(should_succeed=True)
+    check_multivector_query_id(should_succeed=True)
+
+    ##### Set strict mode with very low read_rate_limit, it should not succeed
```
timvisee, line 1782:
```suggestion
    # Set strict mode with very low read_rate_limit, it should not succeed
```

tests/openapi/test_strictmode.py
```diff
@@ -1727,4 +1733,79 @@ def test_strict_mode_distance_matrix_limits(collection_name):
...
+
+    # reset rate limiter for next request
+    set_strict_mode(collection_name, {
+        "enabled": True,
+        "read_rate_limit": 2 # multivector has 3 vectors
+    })
+    check_multivector_query_id(should_succeed=False)
+
+
+    #### Set strict mode with just enough read_rate_limit, it should succeed
```
timvisee, line 1797:
```suggestion
    # Set strict mode with just enough read_rate_limit, it should succeed
```

timvisee (commented):
Minor nit on how we estimate cost for sparse vectors:

lib/segment/src/data_types/vectors.rs
coszio: Good idea, thanks

coszio:
I had edited a consensus test's imports which resulted in not killing the cluster when the test finished, and then the next test would fail. It is fixed now.

PR #6357, opened 2025-04-09
[query api] don't propagate offset into prefetches

coszio:
One overlooked quirk when developing prefetches is that we should not propagate the main query offset into them. 

In general terms, offsets are not compatible with prefetches. Let's look at one example:

Imagine we have a vector search prefetch (limit=5), and an order_by main query (limit=3). For simplicity, let's assume letters are point IDs, and numbers are the order_by values
```
with offset 0
prefetch candidates: [a: 5, b: 8, c: 3, d: 9, e: 2]

main query ordering: [e: 2, c: 3, a: 5, b: 8, d: 9]
main query results:  [e: 2, c: 3, a: 5]

----------------------------------------
with offset 3 (propagated)
prefetch candidates: [a: 5, b: 8, c: 3, d: 9, e: 2, f: 6, g: 1, h: 4]

main query ordering: [g: 1, e: 2, c: 3, h: 4, a: 5, f: 6, b: 8, d: 9]
main query results:  [h: 4, a: 5, f: 6] 
```

As you can see, the second query, with an offset is not consistent with the first one. Result `g` will never appear.

With this PR, offset is not propagated anymore, so the limit in the prefetch is the total amount of offsetting that can happen for the request. If the user keeps paginating with offset and a fixed limit, results will be stable, but will eventually lead to an empty result.

### DIFF
--- a/lib/common/memory/src/madvise.rs
+++ b/lib/common/memory/src/madvise.rs
@@ -1,7 +1,6 @@
 //! Platform-independent abstractions over [`memmap2::Mmap::advise`]/[`memmap2::MmapMut::advise`]
 //! and [`memmap2::Advice`].
 
-use std::fs::File;
 use std::hint::black_box;
 use std::io;
 use std::num::Wrapping;
@@ -187,6 +186,7 @@ pub fn clear_disk_cache(file_path: &Path) -> io::Result<()> {
         target_env = "uclibc",
     ))]
     {
+        use std::fs::File;
         use std::os::fd::AsRawFd as _;
 
         use nix::fcntl;
--- a/lib/common/memory/src/madvise.rs
+++ b/lib/common/memory/src/madvise.rs
@@ -1,7 +1,6 @@
 //! Platform-independent abstractions over [`memmap2::Mmap::advise`]/[`memmap2::MmapMut::advise`]
 //! and [`memmap2::Advice`].
 
-use std::fs::File;
 use std::hint::black_box;
 use std::io;
 use std::num::Wrapping;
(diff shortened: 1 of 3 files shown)


lib/common/memory/src/madvise.rs
```diff
@@ -1,7 +1,6 @@
 //! Platform-independent abstractions over [`memmap2::Mmap::advise`]/[`memmap2::MmapMut::advise`]
 //! and [`memmap2::Advice`].
 
-use std::fs::File;
```
agourlay, line 4: maybe unecessary on Mac but breaks the build for everyone else :)

timvisee (commented):
Could you add a test that asserts this behavior?

Prefetch with and without an offset, and assert that we eventually run out of results. That'll ensure the behavior we document actually happens in practice.

lib/common/memory/src/madvise.rs
timvisee: It was indeed 'unused' on macOS because it's only usage is gated in

```
    #[cfg(any(
        target_os = "linux",
        target_os = "freebsd",
        target_os = "android",
        target_os = "fuchsia",
        target_os = "emscripten",
        target_os = "wasi",
        target_env = "uclibc",
    ))]
```

I've moved the import into that scope to resolve the problem.

https://github.com/qdrant/qdrant/pull/6357/commits/72cbf76f39a5eea3cc120b5c6d6094e9c827b0a8

coszio:
@timvisee There was actually a failing test (https://github.com/qdrant/qdrant/pull/6357/commits/c5fbbad6830d5309956cdc27be53ae0b98cafe51) that I have now edited to assert the new behavior.

agourlay (approved):

agourlay:
This may break the local mode congruence tests :crystal_ball:

PR #6280, opened 2025-03-31
Rename GHA job names, so that `rust-cache` keys are unique for all workflows/jobs

ffuugoo:
We have three jobs with the name `test` in `rust.yml`, `integration-tests.yml` and `storage-compat.yml` workflows.

The issue is that `rust-cache` action, that we use in these jobs, uses job name (e.g., `test`) as a key for GHA cache, and so this three jobs "conflict" on the same cache key. `rust.yml` job builds Rust uni-tests (`cargo build --tests`/`cargo test`), but `integration-tests.yml` and `storage-compat.yml` build Qdrant binary (`cargo build`/`cargo run`), and so `/target` dir between them are not fully reusable. E.g., if you do `cargo build` and then `cargo build --tests`, the second build would rebuild ~200 crates. Due to all of this, basically on every CI run, either `rust.yml` or `integration-tests.yml`/`storage-compat.yml` uses suboptimal cache, and waste extra time rebuilding stuff.

This PR tweaks job names in our workflows, so that they won't collide anymore.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #6362, opened 2025-04-10
Optimize GitHub Actions workflows

ffuugoo:
This PR optimizes our GitHub Actions workflows:

- use explicit `rust-cache` key for a bunch of integration test workflows, so that we store a single GHA cache file instead of 5 copies of it
- disable debug symbols for `dev` builds
	- as far as I can tell, we pretty much never need debug symbols (at least, not on CI)
	- on my (pretty fast!) machine, disabling debug symbols speeds up builds 20%
	- and also reduces `target` dir size 2-3 times
	  - which should make `rust-cache` 2-3 times smaller
	  - which should make cache restore faster (which currently takes 20-30s)
- remove unnecessary Rust toolchain setup steps from workflows that only use Dockerized builds

(Currently based on #6280.)

__TODO:__
- another optimization that we can do, is to change Docker layer cache storage from GHA to GHCR
  - basically, upload/download Docker layer cache to GitHub Container Registry
  - which would likely be _faster_ purely due to faster download speeds from GHCR
  - but I might need help setting this up from someone who have access to Qdrant GHCR?

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -76,9 +76,9 @@ jobs:
       - name: Run integration tests - multiple peers - pytest
         run: poetry -C tests run pytest tests/consensus_tests --durations=10
         timeout-minutes: 60
-      - name: upload logs in case of failure
+      - name: upload logs
         uses: actions/upload-artifact@v4
-        if: failure() || cancelled()
+        if: always()
         with:
           name: consensus-test-logs
           retention-days: 90 # default max value
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -17,13 +17,15 @@ jobs:
 
     steps:
       - name: Install minimal stable
-        uses: dtolnay/rust-toolchain@stable 
+        uses: dtolnay/rust-toolchain@stable
       - uses: actions/checkout@v4
       - name: Install Protoc
         uses: arduino/setup-protoc@v3
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: integration-tests
       - uses: actions/setup-python@v5
         with:
           python-version: '3.11'
@@ -53,6 +55,8 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: integration-tests
       - name: Install dependencies
         run: sudo apt-get install clang
       - uses: actions/setup-python@v5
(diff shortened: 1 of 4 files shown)


ffuugoo:
__We can also look at `rust-cache` sizes:__

<details>
<summary>test-consensus job in e617b9f</summary>

- e617b9f is the next commit after a534e95 above, so it uses cache stored by a534e95
- cache size (with debug symbols) is 1295 MB
- cache load time is 27s

Screenshot 2025-04-10 at 16 09 26
</details>

<details>
<summary>integration-tests-consensus job for 9293e6f (this PR; same job, just renamed; second run attempt)</summary>

- _first_ attempt failed, so the _second_ attempt picked up new cache _without_ debug symbols
  - stored by some other successful job in this PR, because they all use the same cache key now
- cache size (without debug symbols) is 669 MB
- cache load time is 10s
- so, 2 times smaller cache, 3 times faster load
- (also, the build using this cache seems to be correct: only Qdrant workspace crates are rebuilt)

Screenshot 2025-04-10 at 16 32 01
</details>

ffuugoo:
Pretty much all jobs are faster by a minute or few:

a534e95:
Screenshot 2025-04-10 at 16 39 42

9293e6f:
Screenshot 2025-04-10 at 16 39 52

.github/workflows/integration-tests.yml
```diff
@@ -72,9 +76,9 @@ jobs:
       - name: Run integration tests - multiple peers - pytest
         run: poetry -C tests run pytest tests/consensus_tests --durations=10
         timeout-minutes: 60
-      - name: upload logs in case of failure
+      - name: upload logs
         uses: actions/upload-artifact@v4
-        if: failure() || cancelled()
+        if: always()
```
coszio, line 81: Could you elaborate why we want to have logs even from successful runs?

ffuugoo: I needed logs from successful run to debug my failing runs, and it's super annoying to tweak the workflow and push when you need them. The logs are just 1.8mb, I think it should be fine to always upload them? We can also reduce retention period, I don't think anyone would ever need logs from more than 10 days ago.

ffuugoo: I don't care too much about it, though. Just thought I'll see how everyone else feels about this.

coszio: If we reduce the retention time, then LGTM. I just don't want to store everything and generate a humongous storage "just because".

ffuugoo: My brother in Christ, I've just reduced rust caches size like 10 times (like 10+ GB) each dependency bump. We have space for logs. 😂

PR #6363, opened 2025-04-10
Fix strict mode unindexed group_by path

agourlay:
Fix strict mode for grouping APIs to make sure `unindexed_filtering_retrieve` applies to the field grouped on.

### DIFF
--- a/lib/collection/src/operations/verification/query.rs
+++ b/lib/collection/src/operations/verification/query.rs
@@ -127,6 +127,15 @@ impl StrictModeVerification for CollectionQueryGroupsRequest {
             // check for unindexed fields in formula
             query.check_strict_mode(collection, strict_mode_config)?
         }
+        // check for unindexed fields targeted by group_by
+        if strict_mode_config.unindexed_filtering_retrieve == Some(false)
+            && !collection.payload_key_is_indexed(&self.group_by)
+        {
+            return Err(CollectionError::strict_mode(
+                format!("Index required but not found for \"{}\"", self.group_by,),
+                "Create an index for this key.",
+            ));
+        }
         Ok(())
     }
 
--- a/lib/collection/src/operations/verification/query.rs
+++ b/lib/collection/src/operations/verification/query.rs
@@ -1,6 +1,6 @@
 use segment::types::StrictModeConfig;
 
-use super::StrictModeVerification;
+use super::{StrictModeVerification, check_grouping_field};
 use crate::collection::Collection;
 use crate::operations::types::{CollectionError, CollectionResult};
 use crate::operations::universal_query::collection_query::{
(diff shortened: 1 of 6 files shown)


coszio (changes requested):

lib/collection/src/operations/verification/query.rs
```diff
@@ -127,6 +127,15 @@ impl StrictModeVerification for CollectionQueryGroupsRequest {
             // check for unindexed fields in formula
             query.check_strict_mode(collection, strict_mode_config)?
         }
+        // check for unindexed fields targeted by group_by
+        if strict_mode_config.unindexed_filtering_retrieve == Some(false)
+            && !collection.payload_key_is_indexed(&self.group_by)
+        {
+            return Err(CollectionError::strict_mode(
+                format!("Index required but not found for \"{}\"", self.group_by,),
+                "Create an index for this key.",
```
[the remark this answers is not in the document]
agourlay: I was not able to reuse the `unindexed field extractor` because I don't have a filter to start with.
Instead I made something manually looking up the schema.

lib/segment/src/types.rs
```diff
@@ -1711,6 +1711,20 @@ impl PayloadSchemaType {
...
+            Self::Keyword => true,
+            Self::Integer => true,
+            Self::Bool => true,
+            Self::Uuid => true,
+            Self::Float => false,
+            Self::Geo => false,
+            Self::Text => false,
+            Self::Datetime => false,
+        }
+    }
```
coszio, line 1727: integer and uuid can be parametrized without map index, which would make them not support match conditions

coszio (changes requested):
> I was not able to reuse the unindexed field extractor because I don't have a filter to start with.
Instead I made something manually looking up the schema.

One solution to reuse the unindexed field infra is to create a synthetic match filter to pass to it :)

agourlay:
> One solution to reuse the unindexed field infra is to create a synthetic match filter to pass to it :)

I thought about it but decided against it because I felt that coming up with a synthetic match value was too hacky.
Is the type of the synthetic match value  irrelevant for this trick?

lib/segment/src/types.rs
agourlay: I handled the integer the index here https://github.com/qdrant/qdrant/pull/6363/commits/d6fbc6d2fd0a83c22cc03187acb1abce91cd6055

But I do not think the UUID lookup can be disabled.

coszio:
> I can't do this trick, the unindexed check only that the index has the the same type as the inferred matched value

Then we should fix the unindexed extractor, regular filter checking for strict mode depends on this IIRC.

agourlay:
> Then we should fix the unindexed extractor, regular filter checking for strict mode depends on this IIRC.

I'd rather look at this in a different PR given that is is orthogonal to my current work.

coszio (approved):

agourlay:
In a follow up PR I will fix the existing unindexed logic used by strict mode by leveraging the new match support detection.

agourlay:
Here is the follow up https://github.com/qdrant/qdrant/pull/6496

PR #6364, opened 2025-04-10
Abort resharding if any Resharding replica is to be marked dead

KShivendu:
Nodes can go down while resharding and we found an edge case where it can cause problems. 

Whenever a node goes down, the leader detects this (on next update request) because of `sync_local_state()`, and it notifies other peers by appending 
```
CollectionMeta(SetShardReplicaState(SetShardReplicaState { collection_name: "benchmark", shard_id: 0, peer_id: XYZ, state: Dead, from_state: None }))
```
to the consensus WAL. Now all online nodes are expected to apply this operation locally and **try** to mark the resharding replica as `Dead`. 

However, this logic to mark `Resharding/ReshardingScaleDown` state replicas as `Dead` behaves like this:
- If node that went down was the main resharding replica (`resharding_state.peer_id =  peer_that_went_down`), we abort resharding (this [also](https://github.com/qdrant/qdrant/blob/master/lib/collection/src/shards/shard_holder/resharding.rs#L317-L323) marks the shard as `Active`) 
- If other nodes with `Resharding/ReshardingScaleDown` replica goes down, we ignore the error and hence the leader keeps on trying forever to mark the `Resharding/ReshardingScaleDown` replica as `Dead` (bloats the consensus)

After this PR this logic change to:
- ~If **any** node containing `Resharding/ReshardingScaleDown` replica goes down, we first mark it as `Dead`, and then we abort resharding (doesn't mark the shard as `Active` anymore since it's `Dead` instead of `Resharding/ReshardingScaleDown`)~ 
- If **any** node containing `Resharding/ReshardingScaleDown` replica goes down, we first abort resharding and then mark it as `Dead`.

Reason for this is the following edge case:
- Imagine a node with `Resharding/ReshardingScaleDown` replica goes down
- An update X comes in and the replica down so it's not applied by this node.
- However, we currently mark the replica as `Active` after aborting resharding and `return Ok()`. `Active` replica state means it's expected to be in sync (won't ask for updates after it comes back online).
- Hence, it's better to mark it as `Dead` after aborting resharding so we don't miss any updates. 

Also, it's better to trigger abort if **any** node with resharding replica goes down . Imagine if node is marked as `Dead` then on recovery, it will ask take updates from other replicas hence revert all the work done by resharding. But CM will simply ignore this problem since it relies on `replicas_migrated` (which already includes this node) and doesn't retry resharding for such replicas leaving the replicas out of sync.

TODO:
- [ ] Add tests

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -436,64 +436,49 @@ impl Collection {
             )));
         }
 
-        // Abort resharding, if resharding shard is marked as `Dead`.
-        //
-        // This branch should only be triggered, if resharding is currently at `MigratingPoints`
-        // stage, because target shard should be marked as `Active`, when all resharding transfers
-        // are successfully completed, and so the check *right above* this one would be triggered.
-        //
-        // So, if resharding reached `ReadHashRingCommitted`, this branch *won't* be triggered,
-        // and resharding *won't* be cancelled. The update request should *fail* with "failed to
-        // update all replicas of a shard" error.
-        //
-        // If resharding reached `ReadHashRingCommitted`, and this branch is triggered *somehow*,
-        // then `Collection::abort_resharding` call should return an error, so no special handling
-        // is needed.
-        let is_resharding = matches!(
-            current_state,
-            Some(ReplicaState::Resharding | ReplicaState::ReshardingScaleDown)
-        );
-
-        // If *any* of re-sharding replicas are dead, we need to abort re-sharding before marking them as Dead
-        if is_resharding && new_state == ReplicaState::Dead {
-            drop(shard_holder); // need to release lock so abort_resharding() can take write lock
-
-            if let Some(state) = self.resharding_state().await {
-                self.abort_resharding(state.key(), false).await?;
-            }
-
-            // Actually mark as Dead
-            {
-                let shard_holder = self.shards_holder.read().await;
-                let replica_set = shard_holder
-                    .get_shard(shard_id)
-                    .ok_or_else(|| shard_not_found_error(shard_id))?;
-
-                replica_set
-                    .ensure_replica_with_state(peer_id, ReplicaState::Dead)
-                    .await?;
-            }
-
-            return Ok(());
-        }
-
         // Update replica status
         replica_set
             .ensure_replica_with_state(peer_id, new_state)
             .await?;
 
         if new_state == ReplicaState::Dead {
-            // TODO(resharding): Abort all resharding transfers!?
-
-            // Terminate transfer if source or target replicas are now dead
+            let resharding_state = shard_holder.resharding_state.read().clone();
             let related_transfers = shard_holder.get_related_transfers(shard_id, peer_id);
 
-            // `abort_shard_transfer` locks `shard_holder`!
+            // Functions below lock `shard_holder`!
             drop(shard_holder);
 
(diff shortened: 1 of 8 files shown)


lib/collection/src/collection/mod.rs
```diff
@@ -454,13 +454,15 @@ impl Collection {
...
+            .await?;
+
         if is_resharding && new_state == ReplicaState::Dead {
             drop(shard_holder);
 
-            let resharding_state = self
-                .resharding_state()
-                .await
-                .filter(|state| state.peer_id == peer_id);
+            let resharding_state = self.resharding_state().await;
```
timvisee, line 465: I suggest to add this as minor remark, the plan is to eventually support multiple concurrent operations.

```suggestion
            // Pick the resharding state that is related to this replica
            // Currently we can have one operation at a time thus we pick the main one
            let resharding_state = self.resharding_state().await;
```

ffuugoo:
Aborting resharding and marking is `Active` was intentional, because
- in case of resharding-up, if we abort resharding, we simply remove the whole shard, we don't even need to bother with `Dead`
- in case of resharding-down, my logic was that we migrate *additional* points into existing shard, so if we abort resharding, the node still have all *original* points, so we can switch it back to `Active`
- but seems like I did not think it completely through 😅

> An update X comes in and the replica down so it's not applied by this node

Good point. I think I mostly thought about "resharding transfer" and "resharded/migrated points updates" when I was doing scale-down, but forgot to think how regular updates would interact with it... 😬

Anyway, I'll take a closer look at the PR now.

lib/collection/src/collection/mod.rs
ffuugoo: I'd leave it out for now, cause this is more confusing than helpful (imo) and we might *not* add concurrent resharding support for the next 2 years for all we know, and it will just stay there. 😂

ffuugoo (approved):
LGTM. I think that's how it should have looked from the start. Great catch, and good fix, thanks!

KShivendu:
Regarding the previous edge case:

- Imagine a node with Resharding/ReshardingScaleDown replica goes down
- An update X comes in and the replica down so it's not applied by this node so leader wants to mark all replicas on that resharding peer as `Dead`
- So we abort resharding (mark all replica except as `Active`) and then marks it as `Dead`
- This means when the node comes back online, it will request for updates

However I just realised that there's a problem with this. There's a small window between shard being marked Active after aborting resharding and then it getting marked as `Dead`, if node goes down between this again, we'll keep it as Active.

timvisee (changes requested):
Activating and then marking the replica as dead again is definitely a problem.

Because we release the shard holder lock before aborting, the replica state might blink to be active for a split second if there is another reader. That means that readers might rely on the replica to read corrupt state.

I'm not entirely sure yet what the best way of resolving this is, as there is a few different approaches. I _may_ be to not blindly activate replicas on abort. I'd have to think about it for a second.

> the replica down so it's not applied by this node so leader wants to mark all replicas on that resharding peer as Dead

Minor correction: the operation only marks one (the current) replica as dead, not all replicas on that node.

timvisee:
I suggest different handling and a few fixes. I've implemented them in this sub-PR here: https://github.com/qdrant/qdrant/pull/6394

timvisee (approved):
With the merged changes from <https://github.com/qdrant/qdrant/pull/6394> I think this is good to go.

PR #6365, opened 2025-04-10
Fix some wrong vector-read IO measurements

JojiiOfficial:
Fixes some wrong measurements for vector storage IO reads


lib/segment/src/vector_storage/sparse/simple_sparse_vector_storage.rs
```diff
@@ -179,7 +179,7 @@ impl VectorStorage for SimpleSparseVectorStorage {
     }
 
     fn is_on_disk(&self) -> bool {
-        true
+        false
```
agourlay, line 182: I think the `SimpleSparseVectorStorage` is on disk.
The data is only managed by RocksDB without in memory replication.

JojiiOfficial: Fixed, Thanks!

lib/segment/src/vector_storage/quantized/quantized_scorer_builder.rs
```diff
@@ -40,8 +40,12 @@ impl<'a> QuantizedScorerBuilder<'a> {
         vec_deleted: &'a BitSlice,
         distance: &'a Distance,
         datatype: VectorStorageDatatype,
-        hardware_counter: HardwareCounterCell,
+        mut hardware_counter: HardwareCounterCell,
     ) -> Self {
+        if !quantized_storage.is_on_disk() {
+            hardware_counter.set_vector_io_read_multiplier(0);
+        }
```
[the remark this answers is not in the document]
JojiiOfficial: Fixed!

timvisee (approved):

PR #6383, opened 2025-04-14
Fix update wrapper

generall:
It looks like we had a broken flush procedure in all db wrappers.

Old logic does the following:

- accumulate updates in tmp structure in ram
- On flush - move tmp structure in flush thread
- Write changes from tmp structure to disk

But, what if we read from the structure in the moment after flush thread created, but before it is actually written - we can have inconsistent results.

This PR: makes a copy of changes for the thread and only removes pending updates if they are persisted.

### DIFF
--- a/lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
+++ b/lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
@@ -1,7 +1,6 @@
-use std::collections::{HashMap, HashSet};
-use std::mem;
-
 use parking_lot::Mutex;
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
 
 use crate::common::Flusher;
 use crate::common::operation_error::{OperationError, OperationResult};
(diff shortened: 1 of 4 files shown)


lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
```diff
@@ -57,22 +57,50 @@ impl DatabaseColumnScheduledUpdateWrapper {
...
         debug_assert!(
             inserted.keys().all(|key| !deleted.contains(key)),
             "Key to marked for insertion is also marked for deletion!"
         );
         let wrapper = self.db.clone();
+        let pending_operations_arc = self.pending_operations.clone();
+
         Box::new(move || {
-            for id in deleted {
+            for id in deleted.iter() {
```
IvanPleshkov, line 90: Do we need to cover an execution order? If a user inserts a value and deletes it after, the deletion will be performed first.

timvisee: No, this is not a problem. The operation is listed in insertion xor deletion, not in both. The above debug assertion covers this.

If the operation for a point would be both in the insert/delete list, then we would have an ordering problem, but that is not the case.

It is possible that the current flush closure inserts point 10, and then the user deletes point 10 right away. But this is also properly covered. Yes, the current flush will persist the first operation, but we will not remove the second operation from the pending list. The following flush invocation will properly persist the second operation.

timvisee (approved):
:heavy_check_mark:

lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
IvanPleshkov: Thanks for a detailed explanation!

PR #6385, opened 2025-04-15
Speed up pending changes lookups using ahash

timvisee:
A very simple change - replace the standard [`HashMap`](https://doc.rust-lang.org/std/collections/struct.HashMap.html) with [`AHashMap`](https://docs.rs/ahash/latest/ahash/struct.AHashMap.html). Or more specifically, hash using [`ahash`](https://docs.rs/ahash).

It changes the hashmap in the mmap/rocksdb update buffer types where it lists changes pending flush. Its also applied to some surrounding types/functions. There are more places we can benefit from this, but lets take it step by step.

Ahash is much faster, at least on small numeric keys like we're using in many places. Here's a generic benchmark of <ins>sip13</ins> (used by standard HashMap) versus <ins>ahash</ins>:

image
<small>[source](https://github.com/tkaitchuck/aHash/blob/7d5c661a74b12d5bc5448b0b83fdb429190db1a3/README.md#comparison-with-other-hashers)</small>

I've added a criterion benchmark that tests the change in the buffered mmap bitslice. I expect the effect to be similar for the other types.

Here's a bench for 100 iterations on look ups with 1 million pending changes:

```bash
# Before PR
$ cargo bench --bench mmap_bitslice_buffered_update_wrapper
mmap-bitslice-buffered-update-wrapper/lookup-with-pending-changes
                        time:   [43.298 ms 43.611 ms 43.949 ms]

# After PR
$ cargo bench --bench mmap_bitslice_buffered_update_wrapper
mmap-bitslice-buffered-update-wrapper/lookup-with-pending-changes
                        time:   [30.691 ms 30.881 ms 31.113 ms]
                        change: [-29.886% -29.189% -28.441%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Graph for 100 criterion samples

Red shows the lookup timing distribution for the type before this PR. Blue shows the timings after this PR, being about 30% faster.

Note that in the case of the pending changes list, it may be empty a lot of the time. In that case there is no time difference. The improvement in lookup time will get more and more significant as the pending changes list grows.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/benches/mmap_bitslice_buffered_update_wrapper.rs
+++ b/lib/segment/benches/mmap_bitslice_buffered_update_wrapper.rs
@@ -1,0 +1,72 @@
+use std::fs::File;
+use std::iter;
+
+use criterion::{Criterion, black_box, criterion_group, criterion_main};
+use memmap2::MmapMut;
+use memory::mmap_type::MmapBitSlice;
+use rand::prelude::*;
+use rand::rngs::StdRng;
+use segment::common::mmap_bitslice_buffered_update_wrapper::MmapBitSliceBufferedUpdateWrapper;
+use tempfile::tempdir;
+
+const SIZE: usize = 4 * 1024 * 1024;
+const FLAG_COUNT: usize = 1_000_000;
+const LOOKUP_COUNT: usize = 1_000_000;
+
+fn mmap_bitslice_buffered_update_wrapper(c: &mut Criterion) {
+    let mut rng = StdRng::seed_from_u64(42);
+    let dir = tempdir().unwrap();
+    let path = dir.path().join("bitslice.mmap");
+
+    let file = File::create_new(path).unwrap();
+    file.set_len(SIZE as u64).unwrap();
+    file.sync_all().unwrap();
+
+    let mmap_mut = unsafe { MmapMut::map_mut(&file).unwrap() };
+    let mmap_bitslice = MmapBitSlice::from(mmap_mut, 0);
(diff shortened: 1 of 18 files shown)


generall (approved):

lib/segment/benches/mmap_bitslice_buffered_update_wrapper.rs
```diff
@@ -0,0 +1,72 @@
...
+    let mut group = c.benchmark_group("mmap-bitslice-buffered-update-wrapper");
+
+    let lookups: Vec<_> = iter::repeat_with(|| rng.random::<u64>() as usize % SIZE)
+        .take(LOOKUP_COUNT)
+        .collect();
+
+    group.bench_function("lookup-without-pending-changes", |b| {
+        b.iter(|| {
+            for lookup in &lookups {
+                black_box(mmap_bitslice_buffered_update_wrapper.get(*lookup).unwrap());
```
JojiiOfficial, line 46: I'm not sure if we need this black_box here, because I think it would only hide the return value from the compiler. I can be wrong though

timvisee: Hmm, yeah, you might be right. I just wanted to prevent it completely optimizing out this getter.

JojiiOfficial: I don't have a strong preference and am fine with keeping it. I just wanted to mention it ✅

PR #6390, opened 2025-04-15
Telemetry improvements

generall:
Main idea: for automated tools like Cluster Manager it should be enough to work with level-3. Level-4 is quite expensive in terms of request size and should only be required for in-depth performance debug 

- allow shard states in anonymize telemetry (with hashed peer ids)
- split shard and segments detail levels

--- 

- Make sure cluster manager can use Level3 of telemetry

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -12093,8 +12093,7 @@
       "OperationDurationStatistics": {
         "type": "object",
         "required": [
-          "count",
-          "total_duration_micros"
+          "count"
         ],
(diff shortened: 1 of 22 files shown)


coszio (changes requested):
The only relevant change I'd like is to not output an empty list for `"shards"` when details_level < 3. 

I made a PR (https://github.com/qdrant/qdrant/pull/6398) to fix this and the same for `"resharding"` and `"transfers"`. I didn't commit directly in case we want to keep these two others as basic shards info.

PR #6391, opened 2025-04-15
[mutable text index] No need to wrap vec in option

coszio:
Builds on top of #6264 

I think this is an unnecessary wrapping into `Option`, if a posting list is empty, let it just be an empty `PostingList`.
`Option<PostingList>` and `PostingList` both use 24 bytes, even when the option is `None`.

Main change is this:
```diff
pub struct MutableInvertedIndex {
-    pub(in crate::index::field_index::full_text_index) postings: Vec<Option<PostingList>>,
+    pub(in crate::index::field_index::full_text_index) postings: Vec<PostingList>,
    pub(in crate::index::field_index::full_text_index) vocab: HashMap<String, TokenId>,
    pub(in crate::index::field_index::full_text_index) point_to_docs: Vec<Option<Document>>,
    pub(in crate::index::field_index::full_text_index) points_count: usize,
}
```

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/immutable_inverted_index.rs
@@ -141,9 +141,8 @@ impl From<MutableInvertedIndex> for ImmutableInvertedIndex {
             .postings
             .into_iter()
             .enumerate()
-            .filter_map(|(orig_token, posting)| match posting {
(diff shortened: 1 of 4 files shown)


generall:
Flacky test seems to be related to the change in the PR

coszio:
Actually it comes from the PR this builds on top of (#6391). Investigating...

generall (approved):
:heavy_check_mark:  as soon as conflict is resolved

PR #6393, opened 2025-04-16
criterion hnsw filtered search

IvanPleshkov:
Adding a criterion benchmark for HNSW filtered search.

It covers HNSW and plain search with a numeric payload index with immutable and mmap options.


IvanPleshkov:
Close because of the non-requirement of this bench. If needed, we can create a new one

PR #6394, opened 2025-04-16
Abort resharding after marking as dead

timvisee:
Fixes on top of <https://github.com/qdrant/qdrant/pull/6364>

The main things this solves is:
- if resharding abort fails, don't forget to mark replica as dead
- if aborting resharding, don't forget to also cancel related shard transfers
- first mark replica as dead, then abort resharding
- no more replica state blinking, don't mark as active first and then dead
- actually delete points in replica even if in dead state

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -437,24 +437,6 @@ impl Collection {
             )));
         }
 
-        // Abort resharding, if resharding shard is marked as `Dead`.
-        //
-        // This branch should only be triggered, if resharding is currently at `MigratingPoints`
-        // stage, because target shard should be marked as `Active`, when all resharding transfers
-        // are successfully completed, and so the check *right above* this one would be triggered.
-        //
-        // So, if resharding reached `ReadHashRingCommitted`, this branch *won't* be triggered,
-        // and resharding *won't* be cancelled. The update request should *fail* with "failed to
-        // update all replicas of a shard" error.
-        //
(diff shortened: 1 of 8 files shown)


lib/collection/src/collection/mod.rs
```diff
@@ -436,64 +437,57 @@ impl Collection {
...
+                    abort_resharding_result = self.abort_resharding(state.key(), false).await;
+                }
+                // If resharding up, abort if we kill target shard
+                (_, Some(state))
+                    if state.direction == ReshardingDirection::Up
+                        && state.peer_id == peer_id
+                        && state.shard_id == shard_id =>
+                {
+                    abort_resharding_result = self.abort_resharding(state.key(), false).await;
+                }
```
ffuugoo, line 473: This is kinda wrong. We have to abort if target shard is in `Resharding` state, but we don't need to abort once it's `Active`. (Especially if we already replicated target shard.)

ffuugoo: IMO, previous logic, when we relied purely on `Resharding`/`ReshardingScaleDown` was fine. Works the same but easier to read and understand.

ffuugoo (approved):
Small comment regarding when to trigger `abort_resharding`, LGTM overall

KShivendu (approved):
Code LGTM. I'm testing it locally rn to find any potential problems.

timvisee:
Thanks! Also doing extensive testing now.

PR #6399, opened 2025-04-18
Improve telemetry logic and test

KShivendu:
Some improvements on top of https://github.com/qdrant/qdrant/pull/6390. 

Also checked latency of telemetry requests with level 3 vs 10 in chaos testing.

Before: (level 10)
```
p50 0.8347805
p99 1.5783391899999997
```

After: (level 3)
```
p50 0.7328605
p99 0.93947373
```

That's a 40% speed up in this case :)

### All Submissions:

### DIFF
--- a/lib/collection/src/shards/telemetry.rs
+++ b/lib/collection/src/shards/telemetry.rs
@@ -58,8 +58,8 @@ pub struct LocalShardTelemetry {
     /// Do NOT rely on this number unless you know what you are doing
     #[serde(skip_serializing_if = "Option::is_none")]
     pub num_vectors: Option<usize>,
-    #[serde(skip_serializing_if = "Vec::is_empty")]
-    pub segments: Vec<SegmentTelemetry>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub segments: Option<Vec<SegmentTelemetry>>,
     pub optimizations: OptimizerTelemetry,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub async_scorer: Option<bool>,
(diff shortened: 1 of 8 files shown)


lib/collection/src/shards/telemetry.rs
```diff
@@ -58,6 +58,7 @@ pub struct LocalShardTelemetry {
     /// Do NOT rely on this number unless you know what you are doing
     #[serde(skip_serializing_if = "Option::is_none")]
     pub num_vectors: Option<usize>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
```
generall, line 61: Empty vectors will break OpenAPI validation - it should be `Option<Vec<SegmentTelemetry>>` instead

tests/openapi/test_service.py
```diff
@@ -58,8 +60,42 @@ def test_telemetry_detailed():
...
+            ]
+        elif level > 3:
+            assert list(local_shard.keys()) == [
+                'variant_name', 'status', 'total_optimized_points', 'vectors_size_bytes',
+                'payloads_size_bytes', 'num_points', 'num_vectors', 'segments', 'optimizations', 'async_scorer'
+            ]
+
+    if level >= 4:
+        segment = local_shard['segments'][0]
+        assert list(segment.keys()) == ['info', 'config', 'vector_index_searches', 'payload_field_indices']
```
generall, line 101: I am not sure we care that much about those levels. It makes this test kind of fragile. Also, `.keys()` ordering is not guaranteed

KShivendu: > Also, .keys() ordering is not guaranteed

It returns ordered_dict and Qdrant always returns them in same order. it's not gonna be flaky.

> I am not sure we care that much about those levels

yeah it's not critical piece of code. Better to have tests than not have them?

generall: > It returns ordered_dict and Qdrant always returns them in same order. 

nothing really enforces that

KShivendu: Using `set` to avoid order change issues now.

generall (approved):

PR #6403, opened 2025-04-20
careful index creation

generall:
Current implementation of payload index building had a problem: if payload index already existed and we tried to overwrite it with a new one, and the service was stopped during the process, we could end up in the situation where one index was created in storage, while another index was configured. That caused panic on start:

```
2025-04-20T09:21:25.054723Z  INFO storage::content_manager::toc: Loading collection: hist_vector_100D_40
2025-04-20T09:21:25.439576Z ERROR qdrant::startup: Panic backtrace:
   0: std::backtrace::Backtrace::create
   1: qdrant::startup::setup_panic_hook::{{closure}}
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::begin_panic_handler::{{closure}}
   4: std::sys::backtrace::__rust_end_short_backtrace
   5: rust_begin_unwind
   6: core::panicking::panic_fmt
   7: core::result::unwrap_failed
   8: segment::index::field_index::numeric_index::mutable_numeric_index::MutableNumericIndex<T>::load
   9: segment::index::field_index::numeric_index::NumericIndexInner<T>::load
  10: segment::index::struct_payload_index::StructPayloadIndex::open
  11: segment::segment_constructor::segment_constructor_base::create_segment
  12: segment::segment_constructor::segment_constructor_base::load_segment
  13: std::sys::backtrace::__rust_begin_short_backtrace
  14: core::ops::function::FnOnce::call_once{{vtable.shim}}
  15: std::sys::pal::unix::thread::Thread::new::thread_start
  16: <unknown>
  17: <unknown>
2025-04-20T09:21:25.439604Z ERROR qdrant::startup: Panic occurred in file lib/segment/src/index/key_encoding.rs at line 120: called `Result::unwrap()` on an `Err` value: TryFromSliceError(())
```

To prevent this error, this PR changes index creation function in a way, that we always drop incompatible index before trying to build a new one.

Additionally, this PR removes obsolete auto-inference of the index type, which is not used anyway.

### DIFF
--- a/lib/collection/src/collection_manager/holders/proxy_segment.rs
+++ b/lib/collection/src/collection_manager/holders/proxy_segment.rs
@@ -12,6 +12,7 @@ use common::types::{PointOffsetType, TelemetryDetail};
 use itertools::Itertools;
 use parking_lot::{RwLock, RwLockUpgradableReadGuard};
 use segment::common::operation_error::{OperationResult, SegmentFailedState};
+use segment::data_types::build_index_result::BuildFieldIndexResult;
 use segment::data_types::facets::{FacetParams, FacetValue};
 use segment::data_types::named_vectors::NamedVectors;
 use segment::data_types::order_by::OrderValue;
@@ -286,6 +287,11 @@ impl ProxySegment {
                             }
                             ProxyIndexChange::Delete(version) => {
                                 wrapped_segment.delete_field_index(*version, field_name)?;
+                            }
+                            ProxyIndexChange::DeleteIfIncompatible(version, schema) => {
+                                wrapped_segment.delete_field_index_if_incompatible(
+                                    *version, field_name, schema,
+                                )?;
                             }
                         }
                     }
@@ -1169,15 +1175,36 @@ impl SegmentEntry for ProxySegment {
             .delete_field_index(op_num, key)
     }
 
+    fn delete_field_index_if_incompatible(
+        &mut self,
+        op_num: SeqNumberType,
+        key: PayloadKeyTypeRef,
+        field_schema: &PayloadFieldSchema,
+    ) -> OperationResult<bool> {
+        if self.version() > op_num {
+            return Ok(false);
+        }
+
+        self.changed_indexes.write().insert(
+            key.clone(),
+            ProxyIndexChange::DeleteIfIncompatible(op_num, field_schema.clone()),
+        );
+
+        self.write_segment
+            .get()
+            .write()
+            .delete_field_index_if_incompatible(op_num, key, field_schema)
+    }
+
     fn build_field_index(
         &self,
         op_num: SeqNumberType,
         key: PayloadKeyTypeRef,
-        field_type: Option<&PayloadFieldSchema>,
+        field_type: &PayloadFieldSchema,
         hw_counter: &HardwareCounterCell,
-    ) -> OperationResult<Option<(PayloadFieldSchema, Vec<FieldIndex>)>> {
+    ) -> OperationResult<BuildFieldIndexResult> {
         if self.version() > op_num {
-            return Ok(None);
+            return Ok(BuildFieldIndexResult::SkippedByVersion);
         }
 
         self.write_segment
@@ -1224,6 +1251,13 @@ impl SegmentEntry for ProxySegment {
                 }
                 ProxyIndexChange::Delete(_) => {
                     indexed_fields.remove(field_name);
+                }
+                ProxyIndexChange::DeleteIfIncompatible(_, schema) => {
+                    if let Some(existing_schema) = indexed_fields.get(field_name) {
+                        if existing_schema != schema {
+                            indexed_fields.remove(field_name);
+                        }
+                    }
                 }
             }
         }
@@ -1412,6 +1446,7 @@ impl ProxyIndexChanges {
 pub enum ProxyIndexChange {
     Create(PayloadFieldSchema, SeqNumberType),
     Delete(SeqNumberType),
+    DeleteIfIncompatible(SeqNumberType, PayloadFieldSchema),
 }
 
 impl ProxyIndexChange {
@@ -1419,6 +1454,7 @@ impl ProxyIndexChange {
         match self {
             ProxyIndexChange::Create(_, version) => *version,
             ProxyIndexChange::Delete(version) => *version,
+            ProxyIndexChange::DeleteIfIncompatible(version, _) => *version,
         }
     }
 }
--- a/lib/segment/src/data_types/build_index_result.rs
+++ b/lib/segment/src/data_types/build_index_result.rs
@@ -1,0 +1,16 @@
+use crate::index::field_index::FieldIndex;
+use crate::types::PayloadFieldSchema;
+
+pub enum BuildFieldIndexResult {
+    /// Index was not built, as operation version is lower than segment version
+    SkippedByVersion,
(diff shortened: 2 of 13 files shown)


lib/segment/src/data_types/build_index_result.rs
```diff
@@ -0,0 +1,16 @@
+use crate::index::field_index::FieldIndex;
+use crate::types::PayloadFieldSchema;
+
+pub enum BuildFieldIndexResult {
```
generall, line 4: This result structure enforces the caller to handle each case, including the `IncompatibleSchema`. It makes sure we don't forget to follow the protocol of removing existing index (otherwise it is impossible to handle `IncompatibleSchema` case)

lib/segment/src/entry/entry_point.rs
```diff
@@ -297,13 +306,35 @@ pub trait SegmentEntry: SnapshotEntry {
...
+                BuildFieldIndexResult::AlreadyExists => {
+                    return Ok(false);
+                }
+                BuildFieldIndexResult::IncompatibleSchema => {
+                    // This is a service error, as we should have just removed the old index
+                    // So it should not be possible to get this error
+                    return Err(OperationError::service_error(format!(
+                        "Incompatible schema for field index on field {key}",
+                    )));
+                }
```
generall, line 333: This handling only makes sense since we called `delete_field_index_if_incompatible` before

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -483,18 +482,18 @@ impl PayloadIndex for StructPayloadIndex {
...
         if let Some(prev_schema) = self.config.indexed_fields.get(field) {
             // the field is already indexed with the same schema
             // no need to rebuild index and to save the config
-            if prev_schema == payload_schema {
-                return Ok(None);
-            }
+            return if prev_schema == payload_schema {
+                Ok(BuildIndexResult::AlreadyBuilt)
+            } else {
+                Ok(BuildIndexResult::IncompatibleSchema)
```
generall, line 492: We can't handle `IncompatibleSchema` inside `build_index` function, cause this function only has read-only access to the segment. This is important, since we want to split index building and index registration. Index building is a long process, and we don't want to build it while holding write lock. Therefore we have to process `IncompatibleSchema` externally

lib/segment/src/entry/entry_point.rs
```diff
@@ -297,13 +306,35 @@ pub trait SegmentEntry: SnapshotEntry {
...
+                BuildFieldIndexResult::Built { schema, indexes } => (schema, indexes),
+            };
+
+        self.apply_field_index(op_num, key.to_owned(), schema, indexes)
     }
```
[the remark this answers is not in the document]
generall: If it already exists, means it is already registered. Also, it is how existing logic works, this PR didn't change it.

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -483,18 +482,18 @@ impl PayloadIndex for StructPayloadIndex {
...
+            };
         }
-
         let indexes = self.build_field_indexes(field, payload_schema, hw_counter)?;
-
-        Ok(Some(indexes))
+        Ok(BuildIndexResult::Built(indexes))
     }
```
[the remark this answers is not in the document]
generall: This doesn't make sense, cause `let Some(prev_schema) = self.config.indexed_fields.get(field)` already guarantees that the key exist

lib/collection/src/collection_manager/holders/proxy_segment.rs
```diff
@@ -1169,15 +1175,36 @@ impl SegmentEntry for ProxySegment {
...
+        self.changed_indexes.write().insert(
+            key.clone(),
+            ProxyIndexChange::DeleteIfIncompatible(op_num, field_schema.clone()),
+        );
+
+        self.write_segment
+            .get()
+            .write()
+            .delete_field_index_if_incompatible(op_num, key, field_schema)
+    }
```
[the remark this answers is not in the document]
generall: this is intentional

lib/segment/src/index/plain_payload_index.rs
```diff
@@ -105,11 +103,31 @@ impl PayloadIndex for PlainPayloadIndex {
...
 
+    fn set_indexed(
+        &mut self,
+        _field: PayloadKeyTypeRef,
+        _payload_schema: impl Into<PayloadFieldSchema>,
+        _hw_counter: &HardwareCounterCell,
+    ) -> OperationResult<()> {
+        // there are no indexes in the plain index
+        Ok(())
+    }
```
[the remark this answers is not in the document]
generall: ok

lib/segment/src/index/payload_index_base.rs
```diff
@@ -12,10 +12,19 @@ use crate::index::field_index::{CardinalityEstimation, PayloadBlockCondition};
...
 
+pub enum BuildIndexResult {
+    /// Index was built
+    Built(Vec<FieldIndex>),
+    /// Index was already built
+    AlreadyBuilt,
+    /// Field Index already exists, but incompatible schema
+    /// Requires extra actions to remove the old index.
+    IncompatibleSchema,
+}
```
coszio, line 26: This looks too similar to `BuildFieldIndexResult`, is it possible to keep just one? or what is the reason for a separate enum?

lib/segment/src/entry/entry_point.rs
```diff
@@ -297,13 +306,35 @@ pub trait SegmentEntry: SnapshotEntry {
...
-            self.build_field_index(op_num, key, field_schema, hw_counter)?
-        else {
-            return Ok(false);
+        let Some(field_schema) = field_schema else {
+            // Legacy case, where we tried to automatically detect the schema for the field.
+            // We don't do this anymore, as it is not reliable.
+            return Err(OperationError::TypeInferenceError {
+                field_name: key.clone(),
+            });
         };
```
coszio, line 315: Can't we enforce this at the interface level?

lib/segment/src/index/payload_index_base.rs
generall: Similar, but the other one have one extra field.

lib/segment/src/entry/entry_point.rs
generall: That would be a breaking change for the interface.

agourlay: I think it could be done later in a different PR to clean things up :+1:

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -512,6 +511,36 @@ impl PayloadIndex for StructPayloadIndex {
...
 
+    fn set_indexed(
+        &mut self,
+        field: PayloadKeyTypeRef,
+        payload_schema: impl Into<PayloadFieldSchema>,
+        hw_counter: &HardwareCounterCell,
+    ) -> OperationResult<()> {
+        let payload_schema = payload_schema.into();
+
+        self.drop_index_if_incompatible(field, &payload_schema)?;
```
agourlay, line 522: Did you manually test that the index is properly created on restart if we crash just after deleting the incompatible index?

generall: I did

agourlay (approved):

PR #6410, opened 2025-04-21
Fix GPU tests

generall:


generall:
GPU test CI: https://github.com/qdrant/qdrant/actions/runs/14581304011

lib/segment/src/index/hnsw_index/gpu/gpu_level_builder.rs
```diff
@@ -130,9 +130,11 @@ mod tests {
         )
         .unwrap();
 
-        let graph_layers_builder =
+        let mut graph_layers_builder =
```
IvanPleshkov, line 133: Can you please move `graph_layers_builder.fill_ready_list();` inside `create_graph_layers_builder` to avoid code duplication and to hide details of `GraphLayersBuilder` here?

IvanPleshkov (approved):

lib/segment/src/index/hnsw_index/gpu/gpu_level_builder.rs
generall: No, I can not. For the very reason this PR exists.

timvisee (approved):
:ballot_box_with_check:

PR #6413, opened 2025-04-22
Fix inconsistent point deduplication in search/query

timvisee:
Point deduplication was inconsistent in the search and query API.

This changes our deduplication logic in both the search and query API to match our documentation.

It now deduplicates (non-sequential) points by:
- point ID
- shard key - we expect to see the same point more than once if in different shard keys
- order value - we expect to see the same point more than once if it matched multiple payload values (array) on the same payload key

The added tests now confirm consistency.

### Tasks
- [ ] There's a `clone` in the `key()` function which may be expensive, ideally we replace this with just a reference but it's not obvious to do
      (<https://github.com/qdrant/qdrant/pull/6420> and https://github.com/qdrant/qdrant/pull/6426 makes cloning the shard key significantly cheaper, that may be good enough)

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/tests/points_dedup.rs
+++ b/lib/collection/src/tests/points_dedup.rs
@@ -218,6 +218,76 @@ async fn test_scroll_dedup() {
         );
         assert!(record.order_value.is_some());
     }
+
+    // Scroll all points using query API
+    let points = collection
+        .query(
+            ShardQueryRequest {
+                query: None,
+                prefetches: vec![],
+                filter: None,
+                params: Some(SearchParams {
+                    exact: true,
+                    ..Default::default()
+                }),
+                limit: 100,
+                offset: 0,
+                with_payload: WithPayloadInterface::Bool(false),
+                with_vector: WithVector::Bool(false),
+                score_threshold: None,
+            },
+            None,
+            ShardSelectorInternal::All,
+            None,
+            HwMeasurementAcc::disposable(),
+        )
+        .await
+        .expect("failed to scroll");
+    assert!(!points.is_empty(), "expected some points");
+
+    let mut seen = HashSet::new();
+    for point_id in points.iter().map(|point| point.id) {
+        assert!(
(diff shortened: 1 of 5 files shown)


lib/segment/src/types.rs
```diff
@@ -324,6 +324,36 @@ pub struct ScoredPoint {
...
+            shard_key,
+            order_value,
+            version: _,
+            score: _,
+            payload: _,
+            vector: _,
+        } = self;
+
+        // TODO: this is NOT efficient due to clone and may have a significant performance impact
+        (*id, shard_key.clone(), *order_value)
```
timvisee, line 351: I think we should deduplicate points by these three fields.

**But**, this implementation still clones the shard key, which might impact search performance. I think we should take by reference instead, but that isn't obvious to do here.

lib/collection/src/tests/points_dedup.rs
```diff
@@ -284,6 +287,87 @@ async fn test_search_dedup() {
...
+    assert!(!points.is_empty(), "expected some points");
+
+    let mut seen = HashSet::new();
+    for point_id in points.iter().map(|point| point.id) {
+        assert!(
+            seen.insert(point_id),
+            "got point id {point_id} more than once, they should be deduplicated",
+        );
+    }
+}
```
timvisee, line 329: This new test failed before all the other changes in this PR.

We expect the point deduplication behavior of the search and query API to be the same, this test showed that was not the case.

lib/segment/src/types.rs
timvisee: The expensive clone is potentially tacked by <https://github.com/qdrant/qdrant/pull/6420>.

It will make cloning cheap for shard keys of 23 bytes or less.

timvisee:
Closing this once since I don't have any strong evidence for us needing it.

PR #6414, opened 2025-04-22
WIP: Introduce coverage reports for integration tests

KShivendu:
### All Submissions:


timvisee:
Do I see correctly that we still run the full set of integration tests both with and without coverage testing?

Because I think we must keep both. We must also do integration tests with `kill`ing, so we can simply keep both.

generall (approved):

PR #6420, opened 2025-04-23
Use smol string as shard key

timvisee:
Replace the user-defined shard key `String` type to [`SmolStr`](https://docs.rs/smol_str/latest/smol_str/struct.SmolStr.html). It remains exactly the same size (24-bytes), but puts small strings on the stack to eliminate a heap allocation.

I feel like this is a great candidate for the type change because: shard keys can easily be 23-bytes or less, shard keys are cloned all over the place.

Specifically, this change is desired for <https://github.com/qdrant/qdrant/pull/6413> which also wants to extensively clone the shard key to allow point deduplication.

The type changes does not affect our OpenAPI schema.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -4337,8 +4337,11 @@ fn shard_key_number_example() -> u64 {
 #[derive(Deserialize, Serialize, JsonSchema, Anonymize, Debug, Clone, PartialEq, Eq, Hash)]
 #[serde(untagged)]
 pub enum ShardKey {
-    #[schemars(example = "shard_key_string_example")]
-    Keyword(String),
+    #[schemars(
+        schema_with = "String::json_schema",
+        example = "shard_key_string_example"
+    )]
+    Keyword(SmolStr),
     #[schemars(example = "shard_key_number_example")]
     #[anonymize(false)]
     Number(u64),
@@ -4346,13 +4349,19 @@ pub enum ShardKey {
 
 impl From<String> for ShardKey {
     fn from(s: String) -> Self {
+        ShardKey::Keyword(s.into())
+    }
+}
+
+impl From<SmolStr> for ShardKey {
+    fn from(s: SmolStr) -> Self {
         ShardKey::Keyword(s)
     }
 }
 
 impl From<&str> for ShardKey {
     fn from(s: &str) -> Self {
-        ShardKey::Keyword(s.to_owned())
+        ShardKey::Keyword(s.into())
     }
 }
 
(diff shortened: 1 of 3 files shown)


lib/segment/src/types.rs
```diff
@@ -4337,22 +4337,31 @@ fn shard_key_number_example() -> u64 {
 #[derive(Deserialize, Serialize, JsonSchema, Anonymize, Debug, Clone, PartialEq, Eq, Hash)]
 #[serde(untagged)]
 pub enum ShardKey {
-    #[schemars(example = "shard_key_string_example")]
-    Keyword(String),
+    #[schemars(
+        schema_with = "String::json_schema",
+        example = "shard_key_string_example"
+    )]
+    Keyword(SmolStr),
```
timvisee, line 4344: Key change is here!

Both the stack allocation size of the enum type, and the OpenAPI schema remain the same.

lib/segment/src/types.rs
```diff
@@ -4337,22 +4337,31 @@ fn shard_key_number_example() -> u64 {
...
     #[schemars(example = "shard_key_number_example")]
     #[anonymize(false)]
     Number(u64),
 }
 
 impl From<String> for ShardKey {
     fn from(s: String) -> Self {
+        ShardKey::Keyword(s.into())
```
agourlay, line 4352: nitpick: I'd prefer `SmolStr::from`

lib/segment/src/types.rs
```diff
@@ -4337,22 +4337,31 @@ fn shard_key_number_example() -> u64 {
...
+impl From<SmolStr> for ShardKey {
+    fn from(s: SmolStr) -> Self {
         ShardKey::Keyword(s)
     }
 }
 
 impl From<&str> for ShardKey {
     fn from(s: &str) -> Self {
-        ShardKey::Keyword(s.to_owned())
+        ShardKey::Keyword(s.into())
```
agourlay, line 4364: nitpick: I'd prefer `SmolStr::from`

coszio (approved):

xzfc:
Did you consider using [other similar crates](https://github.com/rosetta-rs/string-rosetta-rs) instead of `smol_str`?

Even the author of `smol_str` [recomends switching to a different crate](https://www.reddit.com/r/rust/comments/117ksvr/ecow_compact_cloneonwrite_vector_and_string/j9eh35d/). It seems the reason is that `smol_str` is optimized to serve for rust-analyzer use-cases (e.g. special handling of white-space-only strings) rather than for a general use-case.

timvisee:
> Did you consider using [other similar crates](https://github.com/rosetta-rs/string-rosetta-rs) instead of `smol_str`?
> 
> Even the author of `smol_str` [recomends switching to a different crate](https://www.reddit.com/r/rust/comments/117ksvr/ecow_compact_cloneonwrite_vector_and_string/j9eh35d/). It seems the reason is that `smol_str` is optimized to serve for rust-analyzer use-cases (e.g. special handling of white-space-only strings) rather than for a general use-case.

I did not consider others, and picked this one because it was already listed as dependency.

Happy to switch to another one though! 

Noticed your comment too late and already merged this one. I'll file an issue to take a look at it a little bit later, I don't see it as super significant right now.

PR #6425, opened 2025-04-23
Use smol string as shard key - review suggestions

timvisee:
Applies <https://github.com/qdrant/qdrant/pull/6420#discussion_r2056145876>, which I forgot in <https://github.com/qdrant/qdrant/pull/6420>.

Thought I had applied the review suggestions before merging, but I now discovered I didn't properly push.


agourlay (approved):
Merci :)

PR #6429, opened 2025-04-24
Fix flaky test formula

agourlay:
The test is flaky on slow machines.

```
FAILED tests/openapi/test_query_formula.py::test_formula[True-formula0-<lambda>] - AssertionError: {'status': {'error': 'Wrong input: Expected number value for price in the payload and/or in the formula defaults. Error: Value is not a number'}, 'time': 0.002502803}
```

We need to make sure the payload index has been built.


coszio (approved):
Oopsie! Thanks for looking into it

PR #6434, opened 2025-04-25
Fix warning regarding missing wait param in openapi tests

agourlay:
Fix the following type of warning.

```
 qdrant/tests/openapi/helpers/helpers.py:81: UserWarning: Delete call for /collections/{collection_name}/points/delete missing wait=true param, adding it
```


timvisee (approved):
Thanks! :pray:

PR #6435, opened 2025-04-25
Regen and update poetry lock

agourlay:
This PR cleans up the side effect of the security Dependabot https://github.com/qdrant/qdrant/pull/6430

Namely:
- it upgraded the Poetry version of our lock file which can cause issues
- downgraded a bunch of packages

This PR restores the correct Poetry version and bumps a bunch of versions to keep things fresh.


tests/poetry.lock
```diff
@@ -529,114 +505,106 @@ testing = ["protobuf (>=4.21.9)"]
...
 ]
 
 [package.dependencies]
-grpcio = ">=1.67.1"
-protobuf = ">=5.26.1,<6.0dev"
+grpcio = ">=1.62.3"
+protobuf = ">=4.21.6"
 
 [[package]]
 name = "h11"
```
agourlay, line 603: still the new version https://github.com/qdrant/qdrant/pull/6430

timvisee (approved):

PR #6441, opened 2025-04-25
Only initialize segment RocksDB if it's used by the segment

ffuugoo:
Basically, subject. Currently, we always initialize RocksDB when creating a segment, even if it's later completely unused by any structure in the segment. This PR makes the field optional and tweaks segment initialization, so that we only open it if it's required.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/segment_constructor/rocksdb_builder.rs
+++ b/lib/segment/src/segment_constructor/rocksdb_builder.rs
@@ -1,0 +1,87 @@
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use parking_lot::{RwLock, RwLockReadGuard};
+
+use super::segment_constructor_base::get_vector_name_with_prefix;
+use crate::common::operation_error::{OperationError, OperationResult};
+use crate::common::rocksdb_wrapper;
+use crate::types::{SegmentConfig, SparseVectorStorageType};
+
+#[derive(Debug)]
+pub struct RocksDbBuilder {
+    path: PathBuf,
+    column_families: Vec<String>,
+    rocksdb: Option<Arc<RwLock<rocksdb::DB>>>,
+    is_required: bool,
+}
+
+impl RocksDbBuilder {
+    pub fn new(path: impl Into<PathBuf>, config: &SegmentConfig) -> OperationResult<Self> {
+        let path = path.into();
+
+        let vector_cfs = config.vector_data.keys().map(|vector_name| {
+            get_vector_name_with_prefix(rocksdb_wrapper::DB_VECTOR_CF, vector_name)
+        });
+
+        let sparse_vector_cfs =
+            config
+                .sparse_vector_data
+                .iter()
+                .filter_map(|(vector_name, config)| {
+                    if matches!(config.storage_type, SparseVectorStorageType::OnDisk) {
+                        Some(get_vector_name_with_prefix(
+                            rocksdb_wrapper::DB_VECTOR_CF,
+                            vector_name,
+                        ))
+                    } else {
+                        None
+                    }
+                });
+
+        let column_families: Vec<_> = vector_cfs.chain(sparse_vector_cfs).collect();
+
+        let rocksdb = if rocksdb_wrapper::check_db_exists(&path) {
+            Some(open_db(&path, &column_families)?)
+        } else {
+            None
+        };
+
+        Ok(Self {
+            path,
+            column_families,
+            rocksdb,
+            is_required: false,
+        })
+    }
+
+    pub fn read(&self) -> Option<RwLockReadGuard<'_, rocksdb::DB>> {
+        self.rocksdb.as_ref().map(|db| db.read())
+    }
+
+    pub fn require(&mut self) -> OperationResult<Arc<RwLock<rocksdb::DB>>> {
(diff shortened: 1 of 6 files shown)


lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -306,45 +306,96 @@ pub(crate) fn open_vector_storage(
...
+            Some(db) => db,
+            None => self
+                .rocksdb
+                .insert(Self::open_db(&self.path, &self.column_families)?),
+        };
+
+        self.is_required = true;
+
+        Ok(db.clone())
+    }
```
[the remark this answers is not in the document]
ffuugoo: stupid robot can't see how green CI is, because it has no eyes 😵

comment irrelevant, dismissed

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -306,45 +306,96 @@ pub(crate) fn open_vector_storage(
...
-    segment_path: &Path,
-    config: &SegmentConfig,
-) -> OperationResult<Arc<RwLock<DB>>> {
-    let vector_db_names: Vec<String> = config
-        .vector_data
-        .keys()
-        .map(|vector_name| get_vector_name_with_prefix(DB_VECTOR_CF, vector_name))
-        .chain(
+#[derive(Debug)]
+pub struct RocksDbBuilder {
```
generall, line 310: Please move into dedicated file

generall:
It seems merging this feature can make snapshots incompatible with 1.14.0. Let's put  it  under a feature flag

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -545,25 +597,27 @@ fn create_segment(
     let id_tracker = if mutable_id_tracker {
         // Determine whether we use the new (file based) or old (RocksDB) mutable ID tracker
         // Decide based on the feature flag and state on disk
-        let use_new_mutable_tracker = {
+        let use_new_mutable_tracker = if let Some(db) = db_builder.read() {
             // New ID tracker is enabled by default, but we still use the old tracker if we have
             // any mappings stored in RocksDB
+            //
             // TODO(1.15 or later): remove this check and use new mutable ID tracker unconditionally
```
generall, line 604: Not related to this PR, but it seems like we can't remove this check, as we never actually do migration to the new format. And even though, this function is called `create_segment`, it is also used during loading.

generall:
> It seems merging this feature can make snapshots incompatible with 1.14.0. Let's put it under a feature flag

Actually I tested, and it works fine. Old version just creates empty RocksDB which is not used anyway

lib/segment/src/segment_constructor/segment_constructor_base.rs
ffuugoo: Done.

generall (approved):

lib/segment/src/segment_constructor/rocksdb_builder.rs
```diff
@@ -0,0 +1,87 @@
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use parking_lot::{RwLock, RwLockReadGuard};
+
+use super::segment_constructor_base::get_vector_name_with_prefix;
+use crate::common::operation_error::{OperationError, OperationResult};
+use crate::common::rocksdb_wrapper;
+use crate::types::{SegmentConfig, SparseVectorStorageType};
+
```
timvisee, line 10: A comment like this might be helpful:

```suggestion
/// Struct to optionally create and open a RocksDB instance in a lazy way.
/// Used as helper to eventually completely remove RocksDB.
```

timvisee (approved):
Better than I expected. Nice! Thanks!

lib/segment/src/segment_constructor/rocksdb_builder.rs
```diff
@@ -0,0 +1,88 @@
...
+                .insert(open_db(&self.path, &self.column_families)?),
+        };
+
+        self.is_required = true;
+
+        Ok(db.clone())
+    }
+
+    pub fn build(self) -> Option<Arc<RwLock<rocksdb::DB>>> {
+        self.rocksdb.filter(|_| self.is_required)
```
KShivendu, line 77:
```suggestion
        self.is_required.then(self.rocksdb)
```

a little more readable :)

KShivendu (approved):
Very neat refactor. Learning so much from your PRs. Thanks!

lib/segment/src/segment_constructor/rocksdb_builder.rs
ffuugoo: Would have been nice, but `bool::then`/`bool::then_some` takes `T` and returns `Option<T>`, so in this case it would return `Option<Option<DB>>`, and then you have to "flatten" it and it's not as great as it could have been. 🥲

ffuugoo: Maybe `self.is_required.then(self.rocksdb?)`, but I'd prefer `filter` in this case, tbh.

PR #6473, opened 2025-04-30
Strict mode detects full scan on query

agourlay:
This PR implements fully the strict mode for `search_allow_exact` to make sure no search query can be issued against a disabled HNSW index.

The check takes into account the presence of prefetches to still allow the rescoring use case anyway.

### DIFF
--- a/lib/collection/src/operations/verification/query.rs
+++ b/lib/collection/src/operations/verification/query.rs
@@ -8,7 +8,7 @@ use crate::operations::universal_query::collection_query::{
 };
 
 impl Query {
-    fn check_strict_mode(
+    async fn check_strict_mode(
         &self,
         collection: &Collection,
         strict_mode_config: &StrictModeConfig,
@@ -35,6 +35,54 @@ impl Query {
         }
         Ok(())
     }
+
+    /// Check that the query does not perform a fullscan based on the collection configuration.
+    async fn check_fullscan(
+        &self,
+        using: &str,
+        collection: &Collection,
+        strict_mode_config: &StrictModeConfig,
+    ) -> CollectionResult<()> {
+        // Check only applies on `search_allow_exact`
+        if strict_mode_config.search_allow_exact == Some(false) {
+            match &self {
+                Query::Fusion(_) | Query::OrderBy(_) | Query::Formula(_) | Query::Sample(_) => (),
+                Query::Vector(_) => {
+                    let config = collection.collection_config.read().await;
+
+                    // ignore sparse vectors
+                    let query_targets_sparse = config
+                        .params
+                        .sparse_vectors
+                        .as_ref()
+                        .is_some_and(|sparse| sparse.contains_key(using));
+                    if query_targets_sparse {
+                        // sparse vectors are always indexed
+                        return Ok(());
+                    }
+
+                    // check HNSW configuration for vector
+                    let vector_hnsw_config = &config
+                        .params
+                        .vectors
+                        .get_params(using)
+                        .and_then(|param| param.hnsw_config.as_ref());
+
+                    let vector_hnsw_m = vector_hnsw_config.and_then(|hnsw| hnsw.m);
+                    // TODO(strict-mode) check also payload_m if if there is a filter by tenant/principal
+                    if vector_hnsw_m == Some(0) {
+                        return Err(CollectionError::strict_mode(
+                            format!(
+                                "Fullscan forbidden on '{using}' – vector indexing is disabled (hnsw_config.m = 0)"
+                            ),
+                            "Enable vector indexing or use a prefetch query before rescoring",
+                        ));
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
 }
 
 impl StrictModeVerification for CollectionQueryRequest {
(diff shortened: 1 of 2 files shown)


lib/collection/src/operations/verification/query.rs
```diff
@@ -35,6 +35,50 @@ impl Query {
...
+                // check HNSW configuration for vector
+                let vector_hnsw_config = &config
+                    .params
+                    .vectors
+                    .get_params(using)
+                    .and_then(|param| param.hnsw_config);
+
+                let vector_hnsw_m = vector_hnsw_config.and_then(|hnsw| hnsw.m);
+                let vector_hnsw_payload_m = vector_hnsw_config.and_then(|hnsw| hnsw.payload_m);
+
```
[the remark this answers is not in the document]
agourlay: done

lib/collection/src/operations/verification/query.rs
```diff
@@ -35,6 +35,52 @@ impl Query {
...
+    /// Check that the query does not perform a fullscan based on the collection configuration.
+    async fn check_fullscan(
+        &self,
+        using: &str,
+        collection: &Collection,
+        strict_mode_config: &StrictModeConfig,
+    ) -> CollectionResult<()> {
+        // Check only applies on `unindexed_filtering_retrieve`
+        if strict_mode_config.unindexed_filtering_retrieve == Some(false) {
```
JojiiOfficial, line 47: Nit but maybe we can early return here to reduce amount of indents.

lib/collection/src/operations/verification/query.rs
```diff
@@ -35,6 +35,52 @@ impl Query {
...
+            if let Query::Vector(_) = &self {
+                let config = collection.collection_config.read().await;
```
JojiiOfficial, line 49: Maybe a bit too cautious but are you sure that this block call is not too slow for getting called in every request?
I don't know if there is a proper workaround though.

JojiiOfficial (approved):

lib/collection/src/operations/verification/query.rs
agourlay: I prefer keeping it this to make the scope clearer when add new checks in the custom checker.

lib/collection/src/operations/verification/query.rs
agourlay: In my experience the async lock on the collection configuration is not an issue.

lib/collection/src/operations/verification/query.rs
```diff
@@ -35,6 +35,52 @@ impl Query {
...
     }
+
+    /// Check that the query does not perform a fullscan based on the collection configuration.
+    async fn check_fullscan(
+        &self,
+        using: &str,
+        collection: &Collection,
+        strict_mode_config: &StrictModeConfig,
+    ) -> CollectionResult<()> {
+        // Check only applies on `unindexed_filtering_retrieve`
```
generall, line 46: Shouldn't it be `search_allow_exact` instead?

lib/collection/src/operations/verification/query.rs
```diff
@@ -35,6 +35,52 @@ impl Query {
...
+        if strict_mode_config.unindexed_filtering_retrieve == Some(false) {
+            if let Query::Vector(_) = &self {
```
generall, line 48: Let's use `match` here, explicitly handling each case

lib/collection/src/operations/verification/query.rs
```diff
@@ -35,6 +35,52 @@ impl Query {
...
+
+                // check HNSW configuration for vector
+                let vector_hnsw_config = &config
+                    .params
+                    .vectors
+                    .get_params(using)
+                    .and_then(|param| param.hnsw_config.as_ref());
+
+                let vector_hnsw_m = vector_hnsw_config.and_then(|hnsw| hnsw.m);
+                let vector_hnsw_payload_m = vector_hnsw_config.and_then(|hnsw| hnsw.payload_m);
```
generall, line 70: Payload_m is only an excuse if there is a filter by tenant/principal

lib/collection/src/operations/verification/query.rs
```diff
@@ -89,8 +145,14 @@ impl StrictModeVerification for CollectionPrefetch {
...
 
         if let Some(query) = self.query.as_ref() {
+            // check if prefetch can perform a fullscan
+            query
+                .check_fullscan(&self.using, collection, strict_mode_config)
+                .await?;
             // check for unindexed fields in formula
-            query.check_strict_mode(collection, strict_mode_config)?
+            query
+                .check_strict_mode(collection, strict_mode_config)
```
generall, line 154: Function naming in this case becomes confusing.
`check_fullscan` is followed by `check_strict_mode`. Isn't `check_fullscan` part of strict mode?

generall (changes requested):
Please address commends

lib/collection/src/operations/verification/query.rs
agourlay: Good catch!
You are right, got confused with my ongoing work on group by filtering :facepalm:

lib/collection/src/operations/verification/query.rs
agourlay: Got it, removing it from the check and will tackle multitenancy in a different PR.

lib/collection/src/operations/verification/query.rs
agourlay: `check_strict_mode` on the `Query` does not have the knowledge whether there are prefetches.
This information is necessary to know if it is a rescoring use case.

For `CollectionPrefetch` it should always be checked.
But for `CollectionQueryRequest` only if there are no prefetches.

This is why the calls are not unified.

agourlay:
More review is necessary here :pray:

lib/collection/src/operations/verification/query.rs
```diff
@@ -35,6 +35,54 @@ impl Query {
...
+
+                    // check HNSW configuration for vector
+                    let vector_hnsw_config = &config
+                        .params
+                        .vectors
+                        .get_params(using)
+                        .and_then(|param| param.hnsw_config.as_ref());
+
+                    let vector_hnsw_m = vector_hnsw_config.and_then(|hnsw| hnsw.m);
+                    // TODO(strict-mode) check also payload_m if if there is a filter by tenant/principal
```
generall, line 72: Just FYI: we can't do releases without this todo implemented. Otherwise it is too restrictive

generall (approved):

PR #6475, opened 2025-04-30
Disable nextest retries

agourlay:
The current retry setup for failed tests is too comfortable.

It allowed us to not think too much about the problem of flaky tests and have issues piling on to the side 

Exhibit A: https://github.com/qdrant/qdrant/issues?q=is%3Aissue%20state%3Aopen%20Flaky

I propose to remove the retry mechanism to force us to fix our tests for real.


agourlay:
baby step https://github.com/qdrant/qdrant/pull/6480

agourlay:
closing for now as we need to fix more tests before taking such a drastic action

PR #6477, opened 2025-04-30
Deterministic HNSW builder

coszio:
One source of variation during tests is the creation of hnsw, this PR allows a random number generator to be passed to the builder.

Locally, after running some of the flaky tests a few hundred times, they stopped failing. Let's see if CI agrees over time

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -17,6 +17,7 @@ use itertools::{EitherOrBoth, Itertools as _};
 use log::debug;
 use memory::fadvise::clear_disk_cache;
 use parking_lot::Mutex;
+use rand::Rng;
 use rayon::ThreadPool;
 use rayon::prelude::*;
 
@@ -186,9 +187,9 @@ impl HNSWIndex {
         self.quantized_vectors.clone()
     }
 
-    pub fn build(
+    pub fn build<R: Rng + ?Sized>(
         open_args: HnswIndexOpenArgs<'_>,
-        build_args: VectorIndexBuildArgs<'_>,
+        build_args: VectorIndexBuildArgs<'_, R>,
     ) -> OperationResult<Self> {
         if HnswGraphConfig::get_config_path(open_args.path).exists()
             || GraphLayers::get_path(open_args.path).exists()
@@ -213,6 +214,7 @@ impl HNSWIndex {
             permit,
             old_indices,
             gpu_device,
+            rng,
             stopped,
             feature_flags,
         } = build_args;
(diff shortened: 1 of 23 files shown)


lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -260,7 +262,6 @@ impl HNSWIndex {
             .max_by_key(|old_index| old_index.valid_points);
 
         // Build main index graph
-        let mut rng = rand::rng();
         let deleted_bitslice = vector_storage_ref.deleted_vector_bitslice();
```
coszio, line 266: Main change is here.

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -330,7 +331,7 @@ impl HNSWIndex {
             let level = old_index
                 .as_ref()
                 .and_then(|old_index| old_index.point_level(vector_id))
-                .unwrap_or_else(|| graph_layers_builder.get_random_layer(&mut rng));
+                .unwrap_or_else(|| graph_layers_builder.get_random_layer(rng));
             graph_layers_builder.set_levels(vector_id, level);
```
coszio, line 335: This is the only place during building that I found which was using a random number generator. 

For hnsw gurus: Was I thourough? Are there any other places to pass the `rng` down to?

xzfc: Another possible source of non-determinism is multi-threading, i.e. calling `insert_point()` in `permit.num_cpus` threads.

xzfc (approved):

lib/segment/src/segment_constructor/segment_builder.rs
```diff
@@ -455,10 +456,11 @@ impl SegmentBuilder {
         Ok(true)
     }
 
-    pub fn build(
+    pub fn build<R: Rng + ?Sized>(
```
xzfc, line 459: Usage of generics could make build times/binary size worse

Is it possible to use a non-generic type here? (e.g. `StdRng` or `&mut dyn RngCore`).

agourlay:
Looks like our manual retries on CI finally made it fail.

```
  thread 'multivector_filtrable_hnsw_test::test_multi_filterable_hnsw::case_8_recosumscores_multi' panicked at lib\segment\tests\integration\multivector_filtrable_hnsw_test.rs:217:5:
    hits: 89/100 (expected less than 10 failures)
```

https://github.com/qdrant/qdrant/actions/runs/14763677517/job/41564052730?pr=6477

I am still in favor of this change though :+1:

lib/segment/src/segment_constructor/segment_builder.rs
coszio: From what I saw, we only use `ThreadRng` and `StdRng`. Do you know how much impact it would have if we switch to dyn?

coszio:
> Another possible source of non-determinism is multi-threading, i.e. calling insert_point() in permit.num_cpus threads.

I will try to limit this into single-threaded build. It shouldn't make the test suite slower since other threads can be used for the rest of tests

agourlay:
> I will try to limit this into single-threaded build.

Please validate this hypothesis afterwards :)

coszio:
Ok I am happy with the successful runs

Issue #6423, opened 2025-04-23
Very high LA on server after update to v1.14.0


ava57r:
I updated my test server from v1.13.6 to v1.14.0
I got high Server LA

My use case is
1. search input vectors (batching) (~ 100 - 1000 queries with limit 1)
2. process search results
3. update vectors with blocking API (~ 100 - 2000 points)

## Current Behavior
It picture is  from monitoring.
1 section - qdrant v1.13.6. from: 6:00 to: 11:30
2. section - qdrant v1.14.0. from: 11:30 to: 16:50
3. section - revert qdrant v1.13.6

![Image](https://github.com/user-attachments/assets/5935f221-ae27-4234-ace2-bb57b70caeeb)

## Context (Environment)
server has 56 CPUs

I tried
1. set `incremental_hnsw_building: false` in config
2. change value for collection `optimizer_config.max_optimization_threads`. `null -> 2`
3. change value for collection `hnsw_config.max_indexing_threads`. `0` -> `8`
4. change value in config file `performance.update_rate_limit`. `null` -> `32`
5. change value in config file `service.max_workers`. `0` -> `32`

It did not work.

timvisee:
Could you share the following?
1. collection configuration (`GET /collections/name`)
2. collection distribution (if you have multiple peers) (`GET /collections/name/cluster`)
3. an example of one of your search requests

Could you try the following:
1. is using much smaller batches (10 queries) faster?

It could potentially be related to <https://github.com/qdrant/qdrant/pull/6326>.

ava57r:
> Could you share the following?
> 
>     1. collection configuration (`GET /collections/name`)
> 
>     2. collection distribution (if you have multiple peers) (`GET /collections/name/cluster`)
> 
>     3. an example of one of your search requests
> 
>     4. how many batches are you sending concurrently, just one or multiple?
> 
>     5. what is QPS like in 1.13.6 and in 1.14.0?
> 
> 
> Could you try the following:
> 
>     1. is using much smaller batches (10 queries) faster?
> 
> 
> It could potentially be related to [#6326](https://github.com/qdrant/qdrant/pull/6326).

1.

```
{
  "result": {
    "status": "green",
    "optimizer_status": "ok",
    "indexed_vectors_count": 97450000,
    "points_count": 91000000,
    "segments_count": 300,
    "config": {
      "params": {
        "vectors": {
          "size": 384,
          "distance": "Cosine"
        },
        "shard_number": 12,
[...]

PR #6478, opened 2025-05-01
Don't batch queries if having more segments than search threads

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/6423> regression
Extension of <https://github.com/qdrant/qdrant/pull/6326>

While the condition change is not perfect, it does help significantly with the issue as described above.

This just prevents batching for a specific case that is known to perform bad: when having a high number of segments versus search threads.

### All Submissions:

### New Feature Submissions:


generall (approved):

agourlay:
How did you validate this change?

generall:
Discovered another reason to not do mini-batching:

assume we have very large on-disk collection, and we want to do a full-scan over it.

It would be much faster if we go segment-by-segment, computing full batch of points right away. In this case we only need to load from disk once.

PR #6485, opened 2025-05-05
Fix missing gRPC query validations

agourlay:
This PR fixes a few missing validation for the query APIs:
- check prefetches
- missing group query API


agourlay:
The validation of the `Query` object is still missing but it is properly validated in the REST API.
I will check this in a different PR because it could be large.

generall (approved):

PR #6486, opened 2025-05-05
[phrase match] Introduce ordered document

coszio:
Introduces a new optional field to full-text inverted indices, where the entire document is stored in the form of token ids.

Up to now, a `Document` meant the unique tokens that exist in the document, but now, we need to rethink the concept.

In this PR:
- `Document` is the list of tokens, in the order they appear in the text
- `TokenSet` is the unique tokens in the text.

The new field is not yet populated, it will be done in a later PR

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
@@ -10,12 +10,15 @@ use crate::types::{FieldCondition, Match, PayloadKeyType};
 
 pub type TokenId = u32;
 
+/// Contains the set of tokens that are in a document.
+///
+/// Internally, it keeps them unique and sorted, so that we can binary-search over them
 #[derive(Default, Serialize, Deserialize, Debug, Clone)]
-pub struct Document {
+pub struct TokenSet {
     tokens: Vec<TokenId>,
 }
 
-impl Document {
+impl TokenSet {
     pub fn new(mut tokens: Vec<TokenId>) -> Self {
         tokens.sort_unstable();
         Self { tokens }
@@ -33,7 +36,7 @@ impl Document {
         &self.tokens
     }
 
-    pub fn check(&self, token: &TokenId) -> bool {
+    pub fn contains(&self, token: &TokenId) -> bool {
         self.tokens.binary_search(token).is_ok()
     }
 }
@@ -44,7 +47,7 @@ pub struct ParsedQuery {
 }
 
 impl ParsedQuery {
-    pub fn check_match(&self, document: &Document) -> bool {
+    pub fn check_match(&self, tokens: &TokenSet) -> bool {
         if self.tokens.is_empty() {
             return false;
         }
@@ -52,17 +55,17 @@ impl ParsedQuery {
         // Check that all tokens are in document
         self.tokens
             .iter()
-            .all(|query_token| document.check(query_token))
+            .all(|query_token| tokens.contains(query_token))
     }
 }
 
 pub trait InvertedIndex {
     fn get_vocab_mut(&mut self) -> &mut HashMap<String, TokenId>;
 
-    fn document_from_tokens(&mut self, tokens: &BTreeSet<String>) -> Document {
+    fn token_ids(&mut self, str_tokens: &BTreeSet<String>) -> TokenSet {
         let vocab = self.get_vocab_mut();
-        let mut document_tokens = vec![];
-        for token in tokens {
+        let mut token_ids = vec![];
+        for token in str_tokens {
             // check if in vocab
             let vocab_idx = match vocab.get(token) {
                 Some(&idx) => idx,
@@ -72,16 +75,16 @@ pub trait InvertedIndex {
                     next_token_id
                 }
             };
-            document_tokens.push(vocab_idx);
+            token_ids.push(vocab_idx);
         }
 
-        Document::new(document_tokens)
+        TokenSet::new(token_ids)
     }
 
-    fn index_document(
+    fn index_tokens(
         &mut self,
         idx: PointOffsetType,
-        document: Document,
+        tokens: TokenSet,
         hw_counter: &HardwareCounterCell,
     ) -> OperationResult<()>;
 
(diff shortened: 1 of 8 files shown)


lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -10,31 +12,64 @@ use crate::types::{FieldCondition, Match, PayloadKeyType};
...
+
+/// Contains the token ids that make up a document, in the same order that appear in the document.
+#[derive(Clone, Debug)]
+pub struct Document(Vec<TokenId>);
+
+impl Document {
+    pub fn new(tokens: Vec<TokenId>) -> Self {
+        Self(tokens)
+    }
+}
```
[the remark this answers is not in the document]
coszio: Not actually, but thanks for pointing it out. I will remove the ser/de from `TokenSet`, since we have a custom implementation for it

lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mod.rs
```diff
@@ -195,6 +243,9 @@ impl InvertedIndex for MmapInvertedIndex {
             // Only if the index is within bounds of `point_to_tokens_count`, we decrement the active points count.
             self.active_points_count -= 1;
         }
+        if let Some(_point_to_doc) = &mut self.point_to_doc {
+            // Cannot modify a MmapHashMap
```
generall, line 247: Is it Ok that we don't remove it here?

coszio:
I will exclude the changes in mmap variant, since we want a different representation there, like a compressed posting list instead of plain documents

lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mod.rs
coszio: just removed this, we'll integrate with a postinglist

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -1,40 +1,74 @@
...
+/// Internally, it keeps them unique and sorted, so that we can binary-search over them
+#[derive(Default, Debug, Clone)]
+pub struct TokenSet(Vec<TokenId>);
 
-impl Document {
-    pub fn new(mut tokens: Vec<TokenId>) -> Self {
-        tokens.sort_unstable();
-        Self { tokens }
+impl TokenSet {
+    pub fn new(tokens: AHashSet<TokenId>) -> Self {
```
agourlay, line 21: I think this one could be a `From` impl. instead

lib/segment/src/index/field_index/full_text_index/mutable_inverted_index.rs
```diff
@@ -109,27 +115,63 @@ impl InvertedIndex for MutableInvertedIndex {
...
+
+        let mut hw_cell_wb = hw_counter
+            .payload_index_io_write_counter()
+            .write_back_counter();
+
+        // Ensure container has enough capacity
+        if point_to_doc.len() <= point_id as usize {
+            let new_len = point_id as usize + 1;
+
+            hw_cell_wb.incr_delta((new_len - point_to_doc.len()) * size_of::<Option<Document>>());
```
agourlay, line 142: maybe we could batch all measurements into a single call

coszio:
Drafting while we integrate #6528

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -1,40 +1,66 @@
...
+            .dedup()
+            .collect::<Vec<_>>();
 
-    pub fn check(&self, token: &TokenId) -> bool {
-        self.tokens.binary_search(token).is_ok()
+        Self(tokens)
+    }
+}
+
+/// Contains the token ids that make up a document, in the same order that appear in the document.
```
timvisee, line 58: Also note in the comment that this can contain duplicates, in contrast to the `TokenSet`.

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -44,25 +70,32 @@ pub struct ParsedQuery {
...
     fn get_vocab_mut(&mut self) -> &mut HashMap<String, TokenId>;
 
-    fn document_from_tokens(&mut self, tokens: &BTreeSet<String>) -> Document {
+    /// Translate the string tokens into token ids.
+    /// If it is an unseen token, it is added to the vocabulary and a new token id is generated.
+    ///
+    /// The order of the tokens is preserved.
+    fn token_ids<'a>(
+        &mut self,
+        str_tokens: impl IntoIterator<Item = &'a String> + 'a,
```
timvisee: Can we use `&'a str` here?

lib/segment/src/index/field_index/full_text_index/mutable_inverted_index.rs
```diff
@@ -109,27 +115,63 @@ impl InvertedIndex for MutableInvertedIndex {
...
 
-    fn remove_document(&mut self, idx: PointOffsetType) -> bool {
-        if self.point_to_docs.len() <= idx as usize {
+    fn remove(&mut self, idx: PointOffsetType) -> bool {
+        if self.point_to_tokens.len() <= idx as usize {
             return false; // Already removed or never actually existed
         }
 
-        let Some(removed_doc) = std::mem::take(&mut self.point_to_docs[idx as usize]) else {
+        let Some(removed_token_set) = std::mem::take(&mut self.point_to_tokens[idx as usize])
```
timvisee: Nit:

```suggestion
        let Some(removed_token_set) = self.point_to_tokens[idx as usize].take()
```

timvisee (approved):

lib/segment/src/index/field_index/full_text_index/mutable_inverted_index.rs
coszio: the `WritebackCounterCell` already handles this logic internally 👍

PR #6493, opened 2025-05-05
[phrase matching] populate `point_to_doc` structures

coszio:
Builds on top of #6602 

Adds a `phrase_matching` bool param in the text index parameterization which controls how to store the tokens of each text.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
@@ -30,6 +30,10 @@ impl TokenSet {
         &self.0
     }
 
+    pub fn inner(self) -> Vec<TokenId> {
+        self.0
+    }
+
     pub fn contains(&self, token: &TokenId) -> bool {
         self.0.binary_search(token).is_ok()
     }
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index.rs
@@ -116,24 +116,26 @@ pub trait InvertedIndex {
     /// The order of the tokens is preserved.
     fn register_tokens<'a>(
         &mut self,
-        str_tokens: impl IntoIterator<Item = &'a str> + 'a,
+        str_tokens: impl IntoIterator<Item = impl AsRef<str>> + 'a,
     ) -> Vec<TokenId> {
-        let vocab = self.get_vocab_mut();
-        let mut token_ids = vec![];
-        for token in str_tokens {
-            // check if in vocab
-            let vocab_idx = match vocab.get(token) {
-                Some(&idx) => idx,
-                None => {
-                    let next_token_id = vocab.len() as TokenId;
-                    vocab.insert(token.to_string(), next_token_id);
-                    next_token_id
-                }
-            };
-            token_ids.push(vocab_idx);
-        }
+        str_tokens
+            .into_iter()
+            .map(|token| self.register_token(token.as_ref()))
+            .collect()
+    }
 
-        token_ids
+    /// Translate the string tokens into token ids.
+    /// If it is an unseen token, it is added to the vocabulary and a new token id is generated.
+    fn register_token(&mut self, token_str: &str) -> TokenId {
+        let vocab = self.get_vocab_mut();
+        match vocab.get(token_str) {
+            Some(&idx) => idx,
+            None => {
+                let next_token_id = vocab.len() as TokenId;
+                vocab.insert(token_str.to_string(), next_token_id);
+                next_token_id
+            }
+        }
     }
 
     fn index_tokens(
@@ -322,8 +324,8 @@ mod tests {
         for idx in 0..indexed_count {
             // Generate 10 to 30-word documents
             let doc_len = rand::rng().random_range(10..=30);
-            let str_tokens: BTreeSet<String> = (0..doc_len).map(|_| generate_word()).collect();
-            let token_ids = index.register_tokens(str_tokens.iter().map(String::as_str));
+            let tokens: BTreeSet<String> = (0..doc_len).map(|_| generate_word()).collect();
+            let token_ids = index.register_tokens(&tokens);
             let token_set = TokenSet::from_iter(token_ids);
             index.index_tokens(idx, token_set, &hw_counter).unwrap();
         }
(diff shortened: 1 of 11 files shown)


lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -97,28 +97,32 @@ pub trait InvertedIndex {
...
-            token_ids.push(vocab_idx);
+            token_ids.push(self.token_id(token));
         }
 
         token_ids
     }
 
+    /// Translate the string tokens into token ids.
+    /// If it is an unseen token, it is added to the vocabulary and a new token id is generated.
+    fn token_id(&mut self, token_str: &str) -> TokenId {
```
generall, line 114: nit: maybe rename this function into something which would reflect the mutable nature. Like `register_token`.

generall:
This PR is expected to only affect the mutable index, right?

coszio:
> This PR is expected to only affect the mutable index, right?

Mutable and immutable, since immutable loads from db too.

lib/segment/src/index/field_index/full_text_index/text_index.rs
```diff
@@ -213,21 +213,44 @@ impl FullTextIndex {
...
     }
 
-    pub(super) fn serialize_document_tokens(tokens: BTreeSet<String>) -> OperationResult<Vec<u8>> {
+    pub(super) fn serialize_token_set(tokens: BTreeSet<String>) -> OperationResult<Vec<u8>> {
         #[derive(Serialize)]
-        struct StoredDocument {
+        struct StoredTokens {
+            tokens: BTreeSet<String>,
+        }
+        let doc = StoredTokens { tokens };
```
timvisee, line 240: Could you check if we can simply use a `Vec` here. IIRC, the serialized format is compatible. If possible, we probably want an assertion here that asserts ordering just in case.

timvisee (approved):

timvisee (changes requested):
Could you add a congruence test for the phrase index functionality.

Also, maybe it's worth to add a basic unit test that just asserts Document/TokenSet are handled correctly when filtering/checking - in one we use ordering, in the other we don't.

coszio:
@timvisee 
> Could you add a congruence test for the phrase index functionality.

I have now added cases with phrase enabled in https://github.com/qdrant/qdrant/pull/6493/commits/9bf8e90e5b306deb91d8b6492b15773d3eaea670

> Also, maybe it's worth to add a basic unit test that just asserts Document/TokenSet are handled correctly when filtering/checking - in one we use ordering, in the other we don't.

We haven't implemented phrase checking yet, but will definitely include tests for it.

lib/segment/src/index/field_index/full_text_index/text_index.rs
coszio: That's correct, I have added a test and `cfg(test)` attributes for this in https://github.com/qdrant/qdrant/pull/6493/commits/017a70f82ff4c120dd8c4349c9827d6f7cc06fb5. We can remove the extra implementations in a cleanup PR

lib/segment/src/index/field_index/full_text_index/mutable_inverted_index.rs
```diff
@@ -162,7 +184,7 @@ impl InvertedIndex for MutableInvertedIndex {
             .write_back_counter();
 
         // Ensure container has enough capacity
-        if point_to_doc.len() <= point_id as usize {
+        if point_id as usize >= point_to_doc.len() {
```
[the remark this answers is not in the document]
coszio: I find it easier to read like this 😅

lib/segment/src/index/field_index/full_text_index/inverted_index.rs
```diff
@@ -116,24 +120,26 @@ pub trait InvertedIndex {
...
     ) -> Vec<TokenId> {
+        str_tokens
+            .into_iter()
+            .map(|token| self.register_token(token.as_ref()))
+            .collect()
+    }
+
+    /// Translate the string tokens into token ids.
+    /// If it is an unseen token, it is added to the vocabulary and a new token id is generated.
+    fn register_token(&mut self, token_str: &str) -> TokenId {
```
timvisee: Nit: if using `impl AsRef<str>` above, maybe also use it here and just pass `token` along.

lib/segment/src/index/field_index/full_text_index/mutable_text_index.rs
```diff
@@ -234,8 +242,15 @@ impl MutableFullTextIndex {
...
         let db_idx = FullTextIndex::store_key(idx);
 
-        let db_document =
-            FullTextIndex::serialize_document_tokens(str_tokens.into_iter().collect())?;
+        let tokens_to_store = if phrase_matching {
+            // store ordered tokens
+            str_tokens
+        } else {
+            // store unique tokens
+            BTreeSet::from_iter(str_tokens).into_iter().collect()
```
timvisee:
```rust
str_tokens.sort_unstable();
str_tokens.dedup();`
```

is probably faster.

Alternatively we can use `HashSet`, or do these need to be sorted?

lib/segment/src/index/field_index/full_text_index/text_index.rs
```diff
@@ -226,21 +227,54 @@ impl FullTextIndex {
...
+            .map_err(|e| {
+                OperationError::service_error(format!("Failed to deserialize document: {e}"))
+            })
+            .map(|doc| doc.tokens)
+    }
+
+    pub(super) fn serialize_document(tokens: Vec<String>) -> OperationResult<Vec<u8>> {
+        #[derive(Serialize)]
+        struct StoredDocument {
+            tokens: Vec<String>,
```
timvisee, line 266: Let's add a debug assertion that asserts [`is_sorted()`](https://doc.rust-lang.org/std/primitive.slice.html#method.is_sorted). Just in case.

timvisee (approved):
Nice!

No significant remarks, only nits.

lib/segment/src/index/field_index/full_text_index/mutable_text_index.rs
coszio: That may be true. I opted for BTreeSet because it was converted first in previous implementation

timvisee: Yeah it'd be a micro optimization anyway, which is probably unreasonable.

lib/segment/src/index/field_index/full_text_index/text_index.rs
coszio: In this case we don't want to assert this because tokensets and documents are both serialized with this function. Documents are not sorted

PR #6496, opened 2025-05-06
Fix UnindexedField infra to handle lookup&range index requirements

agourlay:
Follow up to https://github.com/qdrant/qdrant/pull/6363#issuecomment-2847187070

This PR fills a hole in the missing index detection where the `lookup` & `range` capabilities were not taken into account.

Also no fix required for `Formula` because of https://github.com/qdrant/qdrant/pull/6494

### DIFF
--- a/lib/collection/src/problems/unindexed_field.rs
+++ b/lib/collection/src/problems/unindexed_field.rs
@@ -366,7 +366,7 @@ impl<'a> Extractor<'a> {
                 );
                 return;
             }
-            // Any index will suffice
+            // Any index will suffice to get the satellite null index
             Condition::IsEmpty(is_empty) => {
                 key = &is_empty.is_empty.key;
                 required_index = all_indexes().collect();
@@ -396,7 +396,7 @@ impl<'a> Extractor<'a> {
         match self.payload_schema.get(key) {
             Some(index_info) => {
                 // check if the index present has the right capabilities
-                let index_field_types = schema_capacities(index_info);
+                let index_field_types = schema_capabilities(index_info);
                 let already_indexed = required_indexes
                     .iter()
                     .any(|required| index_field_types.contains(required));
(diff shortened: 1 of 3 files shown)


coszio:
This looks hacky. It seems like we are missing one abstraction.

We have:
- `PayloadSchemaType` - which describes the type of data in payload
- `PayloadSchemaParams` - which describes the specific details of each index (both in functionality, like lookup and range; and in performance, like on-disk, is-tenant, etc)

Just because we have customizable integer index, I feel like we should have one which only captures functionality, following the internal `FieldIndex` enum. So that it can be like:

```rust
enum FieldIndexType {
    Int,
    Datetime,
    IntMap,
    Keyword,
    Float,
    Geo,
    FullText,
    Bool,
    Uuid,
    UuidMap,
    Null
}
```

This way we should be able to select one of these when checking the filter, and then validate against schema if it is fulfilled or not. If it is not, it should be trivial to implement a conversion to a payload schema.

agourlay:
@coszio can you please have a look and tell me if the approach is what you had in mind :pray:

lib/collection/src/problems/unindexed_field.rs
```diff
@@ -533,11 +507,117 @@ impl<'a> Extractor<'a> {
...
+                };
+                PayloadFieldSchema::FieldParams(PayloadSchemaParams::Integer(params))
+            }
+            FieldIndexType::IntRange => {
+                let params = IntegerIndexParams {
+                    range: Some(true),
+                    ..Default::default()
+                };
+                PayloadFieldSchema::FieldParams(PayloadSchemaParams::Integer(params))
+            }
```
coszio, line 604: I think we should treat the parametrized versions as advanced usage. Since we don't issue a recommendation for on_disk/is_tenant/etc settings, I'd say we can just convert to unparametrized type index here.

Moreover, this change would make it more obvious to convert to `PayloadSchemaType` instead.

Wdyt?

coszio (approved):

lib/collection/src/problems/unindexed_field.rs
agourlay: Done in https://github.com/qdrant/qdrant/pull/6496/commits/8bc30336e22f5e26b93e3dd9e7250a058606719d :+1:

PR #6497, opened 2025-05-06
Add partial snapshot recovery and read operation locks

ffuugoo:
This PR adds
- `partial_snapshot_recovery_lock`
  - which only allows recovering a single partial snapshot request at a time
- and `partial_snapshot_read_lock`
  - which explicitly rejects read operations while partial snapshot recovery is in progress

__TODO:__
- [x] check partial snapshot recovery lock with `?wait=false`
- [x] add tests
- [ ] ~~report status of the last recovered partial snapshot~~
  - _will be implemented in a separate PR_

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/replica_set/execute_read_operation.rs
+++ b/lib/collection/src/shards/replica_set/execute_read_operation.rs
@@ -122,7 +122,7 @@ impl ShardReplicaSet {
             )));
         };
 
-        self.check_partial_snapshot_read_lock()?;
+        let _partial_snapshot_search_lock = self.try_take_search_read_lock()?;
 
         read_operation(local.get()).await
     }
@@ -157,7 +157,7 @@ impl ShardReplicaSet {
 
         let local_operation = if local_is_active {
             let local_operation = async {
-                self.check_partial_snapshot_read_lock()?;
+                let _partial_snapshot_search_lock = self.try_take_search_read_lock()?;
 
                 let Some(local) = local else {
                     return Err(CollectionError::service_error(format!(
--- a/lib/collection/src/shards/replica_set/execute_read_operation.rs
+++ b/lib/collection/src/shards/replica_set/execute_read_operation.rs
@@ -122,6 +122,8 @@ impl ShardReplicaSet {
             )));
         };
 
+        self.check_partial_snapshot_read_lock()?;
+
         read_operation(local.get()).await
     }
 
@@ -155,6 +157,8 @@ impl ShardReplicaSet {
 
         let local_operation = if local_is_active {
             let local_operation = async {
+                self.check_partial_snapshot_read_lock()?;
+
                 let Some(local) = local else {
                     return Err(CollectionError::service_error(format!(
                         "Local shard {} not found",
--- a/lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
+++ b/lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
@@ -2,18 +2,15 @@ use std::sync::Arc;
 
 use crate::operations::types::{CollectionError, CollectionResult};
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct PartialSnapshotMeta {
     recovery_lock: Arc<tokio::sync::Mutex<()>>,
-    read_lock: Arc<tokio::sync::Semaphore>,
+    search_lock: Arc<tokio::sync::RwLock<()>>,
 }
 
 impl PartialSnapshotMeta {
     pub fn new() -> Self {
-        Self {
-            recovery_lock: Arc::new(tokio::sync::Mutex::new(())),
-            read_lock: Arc::new(tokio::sync::Semaphore::new(1)),
-        }
+        Self::default()
     }
 
     pub fn try_take_recovery_lock(&self) -> CollectionResult<tokio::sync::OwnedMutexGuard<()>> {
@@ -22,22 +19,19 @@ impl PartialSnapshotMeta {
         })
     }
 
-    pub fn try_take_read_lock(&self) -> CollectionResult<tokio::sync::OwnedSemaphorePermit> {
-        self.read_lock.clone().try_acquire_owned().map_err(|_| {
-            CollectionError::bad_request("partial snapshot recovery is already in progress")
-        })
+    pub async fn take_search_write_lock(&self) -> tokio::sync::OwnedRwLockWriteGuard<()> {
+        self.search_lock.clone().write_owned().await
     }
 
-    pub fn check_read_lock(&self) -> CollectionResult<()> {
-        let read_operation_permits = self.read_lock.available_permits();
-
-        if read_operation_permits > 0 {
-            Ok(())
-        } else {
-            Err(CollectionError::ServiceError {
+    pub fn try_take_search_read_lock(
+        &self,
+    ) -> CollectionResult<tokio::sync::OwnedRwLockReadGuard<()>> {
+        self.search_lock
+            .clone()
+            .try_read_owned()
+            .map_err(|_| CollectionError::ServiceError {
                 error: "shard unavailable, partial snapshot recovery is in progress".into(),
                 backtrace: None,
             })
-        }
     }
 }
--- a/lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
+++ b/lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
@@ -1,0 +1,43 @@
+use std::sync::Arc;
+
+use crate::operations::types::{CollectionError, CollectionResult};
+
+#[derive(Debug)]
+pub struct PartialSnapshotMeta {
+    recovery_lock: Arc<tokio::sync::Mutex<()>>,
+    read_lock: Arc<tokio::sync::Semaphore>,
+}
+
+impl PartialSnapshotMeta {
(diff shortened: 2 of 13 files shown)


ffuugoo:
Note: `partia_snapshot_*` locks are moved into separate struct in #6545.

tests/consensus_tests/test_partial_snapshot.py
```diff
@@ -31,6 +32,34 @@ def test_partial_snapshot(tmp_path: pathlib.Path, bootstrap_points: int, recover
...
+    recover_future = executor.submit(recover_partial_snapshot_from, read_peer, write_peer)
+
+    is_search_rejected = False
+    while not recover_future.done():
+        response = try_search_random(read_peer)
+        if response.status_code == 500:
+            is_search_rejected = True
+            break
+
+    assert is_search_rejected, "Search requests have to be rejected during partial snapshot recovery"
```
KShivendu, line 61: Affecting search availability might be not be a good idea? Although, I agree that we need to take it down to load the new temp directory snapshot :thinking:

KShivendu (approved):
Dropped some minor suggestions and a question. Overall LGTM!

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -0,0 +1,43 @@
...
+pub struct PartialSnapshotMeta {
+    recovery_lock: Arc<tokio::sync::Mutex<()>>,
+    read_lock: Arc<tokio::sync::Semaphore>,
+}
+
+impl PartialSnapshotMeta {
+    pub fn new() -> Self {
+        Self {
+            recovery_lock: Arc::new(tokio::sync::Mutex::new(())),
+            read_lock: Arc::new(tokio::sync::Semaphore::new(1)),
```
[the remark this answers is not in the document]
generall: Why does it even require count? Why not mutex?

lib/collection/src/shards/replica_set/snapshots.rs
```diff
@@ -60,6 +60,22 @@ impl ShardReplicaSet {
...
+        self.partial_snapshot_meta.take_recovery_lock()
+    }
+
+    pub fn take_partial_snapshot_read_lock(
+        &self,
+    ) -> CollectionResult<tokio::sync::OwnedSemaphorePermit> {
+        self.partial_snapshot_meta.take_read_lock()
+    }
+
+    pub fn check_partial_snapshot_read_lock(&self) -> CollectionResult<()> {
```
generall, line 75: Does it have to be partial-snapshot specific, or we can do it for any kind of snapshots?

src/actix/api/snapshot_api.rs
```diff
@@ -675,7 +675,31 @@ async fn recover_partial_snapshot_from(
     // nothing to verify
     let pass = new_unchecked_verification_pass();
 
+    let take_recovery_lock_future = async {
```
generall, line 678: same logic should be in regular `recover` method as well

generall: Maybe it would also me nice to make it a function and re-use

lib/collection/src/shards/replica_set/execute_read_operation.rs
```diff
@@ -122,6 +122,8 @@ impl ShardReplicaSet {
             )));
         };
 
+        self.check_partial_snapshot_read_lock()?;
```
generall, line 125: I could not reproduce the problem here, but It feels like we should better check the lock before trying to real local shard.

generall: nevermind, I see now that we actually want to check only after real lock is aquired.

lib/collection/src/shards/replica_set/snapshots.rs
ffuugoo: I think we can easily reuse it for all snapshot types, I just make it partial snapshot specific so far

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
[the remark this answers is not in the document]
ffuugoo: Semaphore, basically, allows you to check `is_locked` without actually **acquiring** the lock. With mutex we can only do `try_lock` which will acquire lock, which we then have to instantly release.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    pub fn take_recovery_lock(&self) -> CollectionResult<tokio::sync::OwnedMutexGuard<()>> {
```
generall, line 19:
```suggest
    pub fn try_take_recovery_lock(&self) -> CollectionResult<tokio::sync::OwnedMutexGuard<()>> {
```

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -0,0 +1,43 @@
...
+        self.recovery_lock.clone().try_lock_owned().map_err(|_| {
+            CollectionError::bad_request("partial snapshot recovery is already in progress")
+        })
+    }
+
+    pub fn take_read_lock(&self) -> CollectionResult<tokio::sync::OwnedSemaphorePermit> {
```
generall, line 25:
```suggestion
    pub fn try_take_read_lock(&self) -> CollectionResult<tokio::sync::OwnedSemaphorePermit> {
```

tests/consensus_tests/test_partial_snapshot.py
ffuugoo: That's by design. We want to reject read requests while local replica is unavailable due to partial snapshot being recovered.

src/actix/api/snapshot_api.rs
generall: Also, would it be possible to move this lock into Collection, so we won't even need to change API handlers?

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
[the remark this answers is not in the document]
generall: So, kind of atomic bool with auto-release?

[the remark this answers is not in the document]
generall: Why don't we use a single mutex and actually hold it while read operation is going?

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -0,0 +1,43 @@
...
+        })
+    }
+
+    pub fn check_read_lock(&self) -> CollectionResult<()> {
+        let read_operation_permits = self.read_lock.available_permits();
+
+        if read_operation_permits > 0 {
+            Ok(())
+        } else {
+            Err(CollectionError::ServiceError {
```
generall, line 37: I understand that we need to have transient error here, but it also generates a lot of `Error` level logs, some users might freak out

src/actix/api/snapshot_api.rs
ffuugoo: Not really (at least I don't see how), because we want to check this lock with `wait=false`, **before** we spawn recover_shard_snapshot future. It *can* be done, but will require refactoring recover snapshot methods.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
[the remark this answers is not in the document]
ffuugoo: Then all read operations would become congested on a single mutex. 😂

We can switch from `Semaphore` to `AtomicBool`... but then we'll have to make auto-release guard around `AtomicBool`, and `Semaphore` already exists.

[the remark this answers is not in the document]
generall: RwLock then?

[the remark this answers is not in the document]
ffuugoo: That might work.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
ffuugoo: Done.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
ffuugoo: Done.

src/actix/api/snapshot_api.rs
ffuugoo: Done.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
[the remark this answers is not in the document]
ffuugoo: Done, but I liked `Semaphore` better.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
ffuugoo: Done.

ffuugoo:
Thanks for the review everyone, will merge once green! 🎉

PR #6498, opened 2025-05-06
Strict mode allows fullscan for multitenant payload index

agourlay:
In a previous [PR](https://github.com/qdrant/qdrant/pull/6473), fullscan was prohibited in strict mode when using `search_allow_exact=false`.

However we want to allow querying when no global HNSW is present if the query leverage a multitenant HNSW based payload index.

### DIFF
--- a/lib/collection/src/operations/verification/query.rs
+++ b/lib/collection/src/operations/verification/query.rs
@@ -118,7 +118,7 @@ impl Query {
                     // HNSW disabled AND no filters
                     return Err(CollectionError::strict_mode(
                         format!(
-                            "Fullscan forbidden{vector_error_label} – vector indexing is disabled (hnsw_config.m = 0)"
+                            "Fullscan forbidden{vector_error_label} because vector indexing is disabled (hnsw_config.m = 0)"
                         ),
                         "Enable vector indexing or use a prefetch query before rescoring",
                     ));
--- a/lib/collection/src/operations/verification/query.rs
+++ b/lib/collection/src/operations/verification/query.rs
@@ -90,10 +90,17 @@ impl Query {
                         if uses_multitenant_filter {
                             // no need for HNSW
                             return Ok(());
+                        } else {
+                            return Err(CollectionError::strict_mode(
+                                format!(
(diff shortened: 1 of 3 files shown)


lib/collection/src/operations/verification/query.rs
```diff
@@ -69,15 +70,43 @@ impl Query {
...
+                    if let Some(filter) = filter {
+                        let uses_multitenant_filter = filter
+                            .iter_conditions()
+                            .filter_map(|c| c.targeted_key())
+                            .filter_map(|key| collection.payload_key_index_schema(&key))
+                            .any(|index_schema| index_schema.is_tenant());
+
+                        if uses_multitenant_filter {
+                            // no need for HNSW
+                            return Ok(());
```
generall, line 92: We still want to make sure that payload_m is > 0

agourlay:
Dear reviewers, please review this PR :pray:

lib/segment/src/types.rs
```diff
@@ -2756,6 +2756,17 @@ impl Condition {
             | Condition::HasVector(_) => 0,
         }
     }
+
+    pub fn targeted_key(&self) -> Option<PayloadKeyType> {
```
JojiiOfficial, line 2760: Can't we return a reference here?

JojiiOfficial (approved):
Only one minor nit but otherwise ✅

lib/segment/src/types.rs
JojiiOfficial: I see. We _could_ make it return a `Cow<PayloadKeyType>` but this is probably not worth the effort.

generall (changes requested):
see test

generall (commented):
see tests

generall (changes requested):
see tests

generall (changes requested):
```
PATCH collections/benchmark
{
    "strict_mode_config": {
        "enabled": true,
        "search_allow_exact": false
    }
}


PATCH collections/benchmark
{
    "vectors": {
        "": {
            "hnsw_config": {
                "ef_construct": 128
            }
        }
    }
}

{
  "query": 1
}
```

generall (approved):

PR #1968, opened 2023-05-24
Optimized HNSW heuristic

IvanPleshkov:
### All Submissions:

* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [ ] Have you formatted your code locally using `cargo fmt` command prior to submission?
3. [ ] Have you checked your code using `cargo clippy` command?

### Changes to Core Features:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #6501, opened 2025-05-07
Heuristic optimization

xzfc:
This PR optimizes heuristic computation.

When building the HNSW graph (particularly when adding incoming links), the following loop pattern happen:
A few links are added to the container, then this container is pruned using the heuristic and the loop starts again.
Illustrated by pseudo code:
```
links += new_points
links = heuristic(links)
links += new_points
links = heuristic(links)
…
```

Running the heuristic requires `O(n*n)` vector comparisons, thus this part of the code could take a significant percentage of the graph building time.

However, some vector comparisons could be avoided if part of the input has already been processed by the heuristic.
This PR implements this optimization by introducing a new struct `LinksContainer` that tracks the number of points that are processed by the heuristic and uses this information to speed up the next heuristic call.

A similar change was proposed two years ago in #1965 and #1968, and described in a [notion page](https://www.notion.so/qdrant/Heuristic-optimization-ee4ff1ee35cc4ce2828ef87d09dd300b).

## Performance

Graph building time measured for a single segment (not the full dataset). Times in seconds.

| Dataset    | Quantization   | Points |   `dev` | This PR |  Change |
| ---------- | -------------- | -----: | ------: | ------: | ------: |
| dbpedia    |                |  25.6k |    7.49 |    7.38 |  -1.5 % |
| dbpedia    |                | 243.8k |  120.00 |  110.93 |  -7.6 % |
| dbpedia    | Binary         | 243.8k |   39.71 |   34.82 | -12.3 % |
| dbpedia    | Scalar u8 0.99 | 243.8k |   66.14 |   60.51 |  -8.5 % |
| dbpedia    | Product x16    | 243.8k | 1289.36 |  789.30 | -38.8 % |
| glove 100  |                |   300k |   57.40 |   50.84 | -11.4 % |
| random 128 |                |   100k |   39.16 |   24.68 | -37.0 % |

## Further improvements

Some ideas (not implemented in this PR):

1. Cache scores, i.e. add `scores: Vec<ScoreType>` into `LinksContainer`. I expect it to cut the number of comparisons by 33% at the cost of using 2x RAM.

   Not implemented in this PR as it imposes a different trade-off. (as opposed to current implementation which I believe is strictly better)

2. Use smarter way to merge old and new scores. In this PR, I use regular sort with comparison function that could skip scores calculation when comparing two old links ([src](https://github.com/qdrant/qdrant/blob/f1a5b5e18e5bbb8a170479ab62b647bb20bb52c4/lib/segment/src/index/hnsw_index/links_container.rs#L175-L181)). This slightly reduces amount of comparisons, but still not ideal. A better approach would be to use merging/sorting algorithm that minimizes number of comparisons. E.g. adding exactly one link should result in a binary search. Check the description of [`binary-merge`](https://docs.rs/binary-merge/0.1.2/binary_merge/index.html) crate. (I'm not proposing to use this particular crate, but their problem statement perfectly describes what I mean).

   Not implemented in this PR as I think it would be an overkill at this moment.

### DIFF
--- a/lib/segment/src/index/hnsw_index/links_container.rs
+++ b/lib/segment/src/index/hnsw_index/links_container.rs
@@ -174,22 +174,38 @@ impl LinksContainer {
         });
         items.0.sort_unstable_by(|a, b| {
             if let (Some(a_order), Some(b_order)) = (a.order, b.order) {
+                // Both items processed by the heuristic, compare their order
+                // to avoid recomputing the score.
                 return a_order.cmp(&b_order);
             }
-            b.score(target_point_id, &mut score)
-                .total_cmp(&a.score(target_point_id, &mut score))
+            b.cached_score(target_point_id, &mut score)
+                .total_cmp(&a.cached_score(target_point_id, &mut score))
         });
 
         self.links.clear();
 
+        // The code below is similar to `fill_from_sorted_with_heuristic` with
+        // two notable differences:
+        //
+        // (A) If both items have already been processed by the heuristic, the
+        // score check is skipped as it is known to pass.
+        //
+        // (B) Instead of having separate input iterator (`candidates`) and
+        // intermediate vector for the processed items (`self.links`), this
+        // implementation reads and updates the same vector in-place:
+        // - `items[read]` is the next candidate to be processed.
+        // - `items[0..write]` are already processed items.
+        // Since `read` ≤ `write`, there are no collisions, so this approach is
+        // sound.
         let mut write = 0;
         'outer: for read in 0..items.0.len() {
             let candidate = items.0[read].clone();
             for existing in &items.0[0..write] {
                 if candidate.order.is_some() && existing.order.is_some() {
-                    continue;
+                    continue; // See (A).
                 }
-                if score(candidate.idx, existing.idx) > candidate.score(target_point_id, &mut score)
+                if score(candidate.idx, existing.idx)
+                    > candidate.cached_score(target_point_id, &mut score)
                 {
                     continue 'outer;
                 }
@@ -214,11 +230,13 @@ pub struct ItemsBuffer(Vec<Item>);
 struct Item {
     idx: PointOffsetType,
     score: Cell<Option<ScoreType>>,
+    /// Order is used to avoid recomputing the score for items known be sorted.
     order: Option<NonZeroU32>,
 }
 
 impl Item {
-    fn score<F>(&self, query: PointOffsetType, score: F) -> ScoreType
+    /// Get the score. This value is lazy/cached: it's computed no more than once.
+    fn cached_score<F>(&self, query: PointOffsetType, score: F) -> ScoreType
     where
         F: FnOnce(PointOffsetType, PointOffsetType) -> ScoreType,
     {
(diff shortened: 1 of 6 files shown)


lib/segment/src/index/hnsw_index/links_container.rs
```diff
@@ -0,0 +1,402 @@
...
+        for x in selected_candidates.iter() {
+            eprintln!("selected_candidates = {x}");
+        }
+    }
+
+    #[test]
+    fn test_connect_new_point() {
+        let m = 6;
+
+        // See illustration in docs
```
IvanPleshkov, line 296: Which docs?

IvanPleshkov:
GPU test is green
https://github.com/qdrant/qdrant/actions/runs/14903570575
It means not only GPU is fine. It also means that cpu-gpu equivalency is green and this PR didn't change heuristic result

lib/segment/src/index/hnsw_index/links_container.rs
```diff
@@ -0,0 +1,402 @@
...
+            }
+        }
+
+        if self.links.len() < level_m {
+            self.links.insert(id_to_insert, new_point_id);
+        } else if id_to_insert != self.links.len() {
+            self.links.pop();
+            self.links.insert(id_to_insert, new_point_id);
+        }
+    }
```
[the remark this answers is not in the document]
xzfc: This is not expected to happen, and it's not worth checking for performance reasons.

lib/segment/src/index/hnsw_index/links_container.rs
xzfc: Added an illustration in a comment.

lib/segment/src/index/hnsw_index/links_container.rs
```diff
@@ -0,0 +1,434 @@
...
+        items.0.push(Item {
+            idx: new_point_id,
+            score: Cell::new(None),
+            order: None,
+        });
+        items.0.sort_unstable_by(|a, b| {
+            if let (Some(a_order), Some(b_order)) = (a.order, b.order) {
+                return a_order.cmp(&b_order);
+            }
+            b.score(target_point_id, &mut score)
```
generall, line 179: Do we assume that re-computing scores for already inserted points once is negligible?

lib/segment/src/index/hnsw_index/links_container.rs
```diff
@@ -0,0 +1,434 @@
...
+                .total_cmp(&a.score(target_point_id, &mut score))
+        });
+
+        self.links.clear();
+
+        let mut write = 0;
+        'outer: for read in 0..items.0.len() {
```
generall, line 186: It was not trivial to read this part. I am not sure which kind of comments would help here, but I had to use copilot to generate some explanations

generall (approved):

lib/segment/src/index/hnsw_index/links_container.rs
xzfc: No, but I decided to not to preserve because it would impose 2x RAM increase (see "Further improvements" point 1 in the PR description). If we decide to save it, I suggest to have it in a separate PR.

Also, `.score()` is lazy/cached, the second call would return the saved result.

And calling `.score()` is skipped if both `a` and `b` are already processed by heuristics.

lib/segment/src/index/hnsw_index/links_container.rs
xzfc: Added an explanation comment.

PR #6504, opened 2025-05-07
Use null index more

coszio:
Currently null index was only working for `is_empty=false` and `is_null=true` cases.

By making the index aware of the `max_point_offset` available, we can also make it work for `is_empty=true` and `is_null=false`.

This PR also builds the index even if the `IndexSelector` chooses RocksDB-based index. Now the null index is chosen regardless of appendable and on-disk parameters.

### Perf gains
In some local manual tests, I can see about 1.8x difference between dev and branch. It is quite small (**~2.5ms vs ~1.4ms**) but noticeable
```shell
# upload 1M points without payload
bfb -n 1000000 --dim 64

# this will apply int payloads to ~10% of points, randomly
bfb -n 100000 --dim 64 --skip-create --int-payloads 1000 --max-id 1000000
```
then created the index through dashboard
```http
PUT /collections/benchmark/index
{
  "field_name": "c",
  "field_schema": {
    "type": "integer",
    "on_disk": true
  }
}
```

then did this for each build, watching the server timings in the logs, not the wall time registered by hyperfine
```shell
hyperfine "curl -X POST -H \"Content-Type: application/json\" -d '{\"query\": 3, \"filter\": {\"must\": {\"key\": \"c\", \"is_empty\": true}}}' localhost:6333/collections/benchmark/points/query" --warmup 200
```

### DIFF
--- a/lib/segment/src/index/field_index/null_index/mmap_null_index.rs
+++ b/lib/segment/src/index/field_index/null_index/mmap_null_index.rs
@@ -209,7 +209,7 @@ impl MmapNullIndex {
 
 impl PayloadFieldIndex for MmapNullIndex {
     fn count_indexed_points(&self) -> usize {
-        self.has_values_slice.count_flags()
+        self.has_values_slice.len()
     }
 
     fn load(&mut self) -> OperationResult<bool> {
--- a/lib/segment/src/index/field_index/null_index/mmap_null_index.rs
+++ b/lib/segment/src/index/field_index/null_index/mmap_null_index.rs
@@ -79,7 +79,10 @@ impl MmapNullIndex {
         })
     }
 
-    pub fn open_if_exists(path: &Path, max_point_offset: PointOffsetType) -> OperationResult<Option<Self>> {
+    pub fn open_if_exists(
+        path: &Path,
+        max_point_offset: PointOffsetType,
+    ) -> OperationResult<Option<Self>> {
         if !path.is_dir() {
             return Ok(None);
         }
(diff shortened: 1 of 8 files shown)


coszio:
There are some errors to solve regarding cardinality estimation

lib/segment/src/index/field_index/null_index/mmap_null_index.rs
```diff
@@ -202,7 +209,7 @@ impl MmapNullIndex {
 
 impl PayloadFieldIndex for MmapNullIndex {
     fn count_indexed_points(&self) -> usize {
-        self.has_values_slice.len()
+        self.has_values_slice.count_flags()
```
generall, line 212: If there a strong reason to change this?

This function is called on INFO, so it might be too expensive to run it every time

generall: Also, we aggregate indexed count for each index and choose the minimal one. So this will likely cause the index to be displayed as 0 in almost all cases

generall (changes requested):
I am not sure those big changes are actually needed. Let's discuss in a call

lib/segment/tests/integration/payload_index_test.rs
```diff
@@ -477,7 +477,7 @@ fn build_test_segments_nested_payload(path_struct: &Path, path_plain: &Path) ->
 
     for (_field, indexes) in struct_segment.payload_index.borrow().field_indexes.iter() {
         for index in indexes {
-            assert!(index.count_indexed_points() < num_points as usize);
+            assert!(index.count_indexed_points() <= num_points as usize);
```
coszio, line 480: this is modified because now the null index is built for `on_disk=false` as well, and null index reports the max id that has had a value, even if it was deleted (`has_values_slice.len()`)

lib/segment/src/index/field_index/null_index/mmap_null_index.rs
coszio: Reverted 👍

generall (approved):

PR #6507, opened 2025-05-08
Abort transfer on force peer delete

KShivendu:
Force deletes to a peer should also abort all shards transfers from this peer. Otherwise, they can remain stuck there forever.

TODO:
- [ ] Think we should abort `to` transfers for this force deleted peer

### All Submissions:

### DIFF
--- a/lib/storage/src/content_manager/toc/mod.rs
+++ b/lib/storage/src/content_manager/toc/mod.rs
@@ -461,6 +461,26 @@ impl TableOfContent {
         false
     }
 
+    pub async fn abort_peer_transfers(&self, peer_id: PeerId) -> CollectionResult<()> {
+        for collection in self.collections.read().await.values() {
+            let related_transfers = collection
+                .shards_holder()
+                .read()
+                .await
+                .get_transfers(|transfer| transfer.from == peer_id);
+
+            for transfer in related_transfers {
+                collection
+                    .shards_holder()
+                    .write()
+                    .await
+                    .register_abort_transfer(&transfer.key())?;
+            }
+        }
+
+        Ok(())
+    }
+
     pub async fn get_telemetry_data(
         &self,
         detail: TelemetryDetail,
--- a/src/actix/api/cluster_api.rs
+++ b/src/actix/api/cluster_api.rs
@@ -79,6 +79,12 @@ fn remove_peer(
             });
         }
 
+        // If we reach here, either there are no shards or force removal is requested
+        if has_shards {
+            log::warn!("Aborting peer transfers for peer {peer_id} before force removal");
+            toc.abort_peer_transfers(peer_id).await?;
+        }
+
         match dispatcher.consensus_state() {
             Some(consensus_state) => {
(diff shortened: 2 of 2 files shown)


lib/storage/src/content_manager/toc/mod.rs
```diff
@@ -461,6 +461,26 @@ impl TableOfContent {
         false
     }
 
+    pub async fn abort_peer_transfers(&self, peer_id: PeerId) -> CollectionResult<()> {
+        for collection in self.collections.read().await.values() {
```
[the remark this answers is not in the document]
KShivendu: > Because removing it from the state here, does not stop the ongoing transfer task

Ahh right. I didn't realise that `register_abort_transfer()` is local and it's not pushing it to consensus. Thanks :)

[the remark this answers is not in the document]
ffuugoo: `remove_peer` is a consensus operation itself, and so it will be called on all peers through consensus, so we don't need to `submit_collection_meta_op`, we can just remove transfer related data/stop transfer tasks *locally*.

Arguably, `submit_collection_meta_op` here is wrong, because it will (I think?) cause *every* peer to submit a copy of `ShardTransferOperations::Abort`.

src/actix/api/cluster_api.rs
```diff
@@ -79,6 +79,13 @@ fn remove_peer(
             });
         }
 
+        // If we reach here, either there are no shards or force removal is requested
+        if has_shards {
+            log::warn!("Aborting peer transfers for peer {peer_id} before force removal");
+            toc.abort_peer_transfers(dispatcher.clone(), access, peer_id)
+                .await?;
+        }
```
ffuugoo, line 87: IMO, this should be done differently. Instead of explicitly submitting `ShardTransferOperations::Abort` in the API, we should modify the code that is called when `RemovePeer` consensus operation is applied, and make it abort transfers as part of `RemovePeer` operation.

lib/collection/src/collection/mod.rs
```diff
@@ -580,6 +580,10 @@ impl Collection {
             }
         }
 
+        for transfer in self.get_related_transfers(peer_id).await {
+            self.abort_shard_transfer(transfer.key(), None).await?;
```
KShivendu, line 584: Edge case: What if you read about a transfer and it finished just before you try to delete it. Has happened in previous test failure. https://github.com/qdrant/qdrant/actions/runs/14967895398/job/42041945647?pr=6507

timvisee (approved):
I like this a lot better! Simple and effective.

lib/collection/src/collection/mod.rs
timvisee: It's fine, it'll be ignored deeper down: https://github.com/qdrant/qdrant/blob/9c723aa48083a4bc5e2d9124b827405b2c35b398/lib/collection/src/collection/shard_transfer.rs#L331-L333

PR #6508, opened 2025-05-08
Make reads sequential in BatchedVectorReader

JojiiOfficial:
Depends on #6487

Improves optimization when copying points from different segments into a new one.

Benchmark:

**On the same machine** with storage on a Hetzner Network Disk (5M points, 20=>10 segments, vectors on disk)
Dev: 200s
PR: 3s

**SSH-FS mounted** Hetzner Network Disk (100k points, 10=>5 segments, vectors on disk)
Dev: ~450s
PR: 240s

PR #6510, opened 2025-05-08
Make gridstore Advise::Random and add function for sequential read

JojiiOfficial:
Depends on #6508


lib/gridstore/src/gridstore.rs
```diff
@@ -218,6 +218,32 @@ impl<V: Blob> Gridstore<V> {
         raw_sections
     }
 
+    /// Read raw value from the pages. Considering that they can span more than one page.
+    fn read_from_pages_sequential(
```
coszio, line 222: This code duplication makes it harder to maintain, I feel like it should be easier to parametrize which memmap to use by introducing an enum and sticking it in the function. Something like:

```rust
enum MmapRead {
    Sequential,
    Random,
}
```

JojiiOfficial: I agree. The reason I still decided to go with code duplication was to prevent introducing (more) branching into this very hot method (most likely a premature optimization though)

I'd rather suggest to use a closure to specify which underlying read-method should be used (which is also not 100% clean) but might prevent adding overhead.

Wdyt?

coszio: A closure sounds fine by me too

generall (approved):

PR #6518, opened 2025-05-12
Fix Clippy 1.87 bis

agourlay:
Follow up to https://github.com/qdrant/qdrant/pull/6516 with non controversial changes.

### DIFF
--- a/lib/collection/src/update_handler.rs
+++ b/lib/collection/src/update_handler.rs
@@ -61,6 +61,7 @@ pub struct OperationData {
 
 /// Signal, used to inform Updater process
 #[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
 pub enum UpdateSignal {
     /// Requested operation to perform
     Operation(OperationData),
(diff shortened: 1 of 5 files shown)


lib/collection/src/update_handler.rs
```diff
@@ -61,6 +61,7 @@ pub struct OperationData {
 
 /// Signal, used to inform Updater process
 #[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
```
agourlay, line 64: `Operation` is the largest variant but also the most used by far, I don't think it is necessary to Box here

lib/common/issues/src/solution.rs
```diff
@@ -7,7 +7,7 @@ use serde_json::Value;
 #[serde(rename_all = "snake_case")]
 pub enum Solution {
     /// A solution that can be applied immediately
-    Immediate(ImmediateSolution),
+    Immediate(Box<ImmediateSolution>),
```
agourlay, line 10: large variant fix

timvisee (approved):

lib/collection/src/update_handler.rs
timvisee: Completely agree :heavy_check_mark:

PR #6527, opened 2025-05-12
Faster stream records transfer

AugustKarlstedt:
Following in the same vein as the changes in #6458. I've observed faster shard recovery doing similar for stream_records. In the case where wal_delta falls back to stream_records without this change, there is a chance the shard could never recover (depending on write load). With this change, we were able to successfully recover with much higher write RPS than before.

### All Submissions:

* [x ] Contributions should target the `dev` branch. Did you create your branch from `dev`?

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/forward_proxy_shard.rs
+++ b/lib/collection/src/shards/forward_proxy_shard.rs
@@ -1,5 +1,6 @@
 use std::path::Path;
 use std::sync::Arc;
+use std::sync::LazyLock;
 use std::time::Duration;
 
 use async_trait::async_trait;
@@ -15,6 +16,7 @@ use segment::types::{
     ExtendedPointId, Filter, PointIdType, ScoredPoint, SizeStats, SnapshotFormat, WithPayload,
     WithPayloadInterface, WithVector,
 };
+use semver::Version;
 use tokio::runtime::Handle;
 use tokio::sync::Mutex;
 
@@ -23,6 +25,7 @@ use super::update_tracker::UpdateTracker;
 use crate::hash_ring::HashRingRouter;
 use crate::operations::point_ops::{
     PointInsertOperationsInternal, PointOperations, PointStructPersisted, PointSyncOperation,
+    WriteOrdering,
 };
 use crate::operations::types::{
     CollectionError, CollectionInfo, CollectionResult, CoreSearchRequestBatch,
@@ -38,6 +41,9 @@ use crate::shards::local_shard::LocalShard;
 use crate::shards::remote_shard::RemoteShard;
 use crate::shards::shard_trait::ShardOperation;
 use crate::shards::telemetry::LocalShardTelemetry;
+
+static MINIMAL_VERSION_FOR_BATCH_TRANSFER: LazyLock<Version> =
+    LazyLock::new(|| Version::parse("1.14.1-dev").unwrap());
 
 /// ForwardProxyShard
 ///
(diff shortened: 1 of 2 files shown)


generall (changes requested):
Hey @AugustKarlstedt, thanks for the contribution!

I see you changed the default batch size quite significantly. I think this is the main reason why you see the performance difference in your case. This is, unfortunately, won't work as good for other cases, where size of the vectors is much larger.

All other changes regarding using a different API method for batching are likely have no effect, as we already do batching inside `SyncPoints` operation.

We will try to take into account your case and come up with more suitable solution.

AugustKarlstedt:
> We will try to take into account your case and come up with more suitable solution.

Sounds good - yeah this is just me trying some stuff to improve recovery times for us. The first commit essentially just increases the batch size, but the second commit does add batching more similarly to the wal_delta improvements. Increasing the batch size by itself did show some speedup (we were able to recover a node in ~40 minutes), but the batching I added in the second commit sped things up considerably (recovered in 6 minutes). 

> I see you changed the default batch size quite significantly. I think this is the main reason why you see the performance difference in your case. This is, unfortunately, won't work as good for other cases, where size of the vectors is much larger.
> 
> All other changes regarding using a different API method for batching are likely have no effect, as we already do batching inside `SyncPoints` operation.

Hmm, I don't quite understand where `SyncPoints` is doing batching. Pseudocode:

```
loop {
  read from shard holder
  transfer batch (just 100 points by default)
  call remote shard update with point sync operation
  makes gRPC request to `/qdrant.PointsInternal/Sync`
}
```

It definitely doesn't seem like it is if so 🤔 

`RemoteShard::update` does not batch but the new methods you added for wal-delta do, that's why I swapped to them.

generall:
> Hmm, I don't quite understand where SyncPoints is doing batching. Pseudocode:

Batch happens here https://github.com/qdrant/qdrant/blob/438fcf86cbef8f09a737a28ff8d957087e3eaff9/lib/collection/src/shards/forward_proxy_shard.rs#L151

All points of the batch are transferred in a single `SyncPoints` update request. I don't see why would multiple  `SyncPoints` operations be different from one large operation.

AugustKarlstedt:
Ah I see what you mean. I have tried increasing the default batch size before but it didn't seem to help very much. Will be curious of your solution!

timvisee:
> Increasing the batch size by itself did show some speedup (we were able to recover a node in ~40 minutes), but the batching I added in the second commit sped things up considerably (recovered in 6 minutes).

That difference is humongous. I'm really curious why it's so much faster.

I agree with Andrey, we already consider `SyncPoints` to be a batch. `SyncPoints(300)` should be the same as `[SyncPoints(100), SyncPoints(100), SyncPoints(100)]`.

Could the difference just be that your first commit transferred 1000 points, while the second commit transfers 5000 points (5*1000)?

AugustKarlstedt:
> > Increasing the batch size by itself did show some speedup (we were able to recover a node in ~40 minutes), but the batching I added in the second commit sped things up considerably (recovered in 6 minutes).
> 
> That difference is humongous. I'm really curious why it's so much faster.
> 
> I agree with Andrey, we already consider `SyncPoints` to be a batch. `SyncPoints(300)` should be the same as `[SyncPoints(100), SyncPoints(100), SyncPoints(100)]`.
> 
> Could the difference just be that your first commit transferred 1000 points, while the second commit transfers 5000 points (5*1000)?

Quite possibly. I did test simply increasing the batch size to a much larger number but didn't see as fast of recovery. I think I tried 100k points, probably too large of an increase.

PR #6532, opened 2025-05-13
Refactor: move HardwareUsage to the Usage map

n0x29a:
The hardware usage was moved into dedicated map in order to make it eventually extendable

### DIFF
--- a/src/actix/helpers.rs
+++ b/src/actix/helpers.rs
@@ -5,7 +5,7 @@ use actix_web::http::header;
 use actix_web::http::header::HeaderMap;
 use actix_web::rt::time::Instant;
 use actix_web::{HttpResponse, ResponseError, http};
-use api::rest::models::{ApiResponse, ApiStatus, HardwareUsage};
+use api::rest::models::{ApiResponse, ApiStatus, HardwareUsage, Usage};
 use collection::operations::types::CollectionError;
 use common::counter::hardware_accumulator::HwMeasurementAcc;
 use serde::Serialize;
@@ -33,7 +33,9 @@ pub fn accepted_response(timing: Instant, hardware_usage: Option<HardwareUsage>)
         result: None,
         status: ApiStatus::Accepted,
         time: timing.elapsed().as_secs_f64(),
-        usage: hardware_usage,
+        usage: Option::from(Usage {
+            hardware: hardware_usage,
+        }),
     })
 }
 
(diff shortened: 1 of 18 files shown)


src/tonic/api/points_internal_api.rs
```diff
@@ -605,17 +609,19 @@ impl PointsInternal for PointsInternalService {
                     }
                 },
             };
-            let mut response = result.into_inner();
+            let response = result.into_inner();
 
-            if let Some(usage) = response.usage.take() {
+            if let Some(usage) = response.clone().usage.unwrap_or_default().hardware.take() {
```
JojiiOfficial, line 614: I'd prefer something like this to prevent unnecessary cloning.
```suggestion
            let mut response = result.into_inner();
            let mut usage = response.usage.take();

            if let Some(usage) = usage.as_mut().and_then(|i| i.hardware.take()) {
                total_usage.add(usage);
            }

```

src/actix/helpers.rs
```diff
@@ -33,7 +33,9 @@ pub fn accepted_response(timing: Instant, hardware_usage: Option<HardwareUsage>)
         result: None,
         status: ApiStatus::Accepted,
         time: timing.elapsed().as_secs_f64(),
-        usage: hardware_usage,
+        usage: Option::from(Usage {
```
JojiiOfficial, line 36: Why using `Option::from` here (and below) instead of `Some(..)`?
Also in `src/tonic/api/points_internal_api.rs`

n0x29a: Option::from is just a more generic approach. It works when the type isn't explicitly Option, or when writing generic code. In this case, Some(..) probably would be more useful, but I used Option::from mostly because of Clippy suggestion. Would you like to change to Some()?

JojiiOfficial:
Could you please rebase this PR onto the latest dev?

src/tonic/api/points_internal_api.rs
```diff
@@ -605,12 +607,7 @@ impl PointsInternal for PointsInternalService {
                     }
                 },
             };
-            let mut response = result.into_inner();
-
-            if let Some(usage) = response.usage.take() {
-                total_usage.add(usage);
-            }
```
generall, line 612: Why is this removed?

generall (approved):

PR #6538, opened 2025-05-15
Fix dev branch coverage

KShivendu:
Coverage reports seems to be getting uploaded for `master` instead of `dev`

Logs don't have warnings or errors so the root cause is not 100% clear. But I suspect the problem is that we don't do enough commit info in `upload-and-merge` stage.  My hypothesis is that it's attaching the commit to the wrong branch because default fetch depth is 1 and hence it doesn't have full context and falls back to registering it as master branch. So this is similar to https://github.com/qdrant/qdrant/pull/6345

### All Submissions:

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -87,8 +87,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


timvisee (approved):

timvisee:
Cherry-picked into master: https://github.com/qdrant/qdrant/commit/beccbfd36300e1166d0074bd79309a856bed8e50

PR #6543, opened 2025-05-16
HNSW healing

xzfc:
This PR implements graph healing algorithm, allowing HNSW index to reuse the old graph even if some of the vectors are updated or deleted.

This is done by the new graph search function `GraphLayersBuilder::search_shortcuts_on_level` which finds shortcut links around the deleted points. (check it's documentation comment for more info)

The main graph building is still done in two stages (migrate phase, and main phase).
But in this PR,
- Point migration not just copies links as-is, but it could replace links to deleted nodes with newly found shortcuts around them..
- Migration is done in parallel, similar to the main phase.

By default, this feature is hidden behind the new `hnsw_healing` feature flag.
The threshold is 10% of missing points (if more than that, then the old index is discarded).

# Benchmark on random data

```console
$ bfb --segments 1 --indexing-threshold 1 -n 50000 --dim 100
$ curl -X PATCH 0:6333/collections/benchmark -d '{"hnsw_config":{"ef_construct":101}}' # Force index rebuild
$ bfb --skip-create --dim 100 -n 1000 # update 2%
$ bfb --skip-create --dim 100 -n 2000 # update 4%
$ bfb --skip-create --dim 100 -n 4000 # update 8%
```

Index build time from scratch: 4519 ms
Index build time with healing (2% updated): 117+363 = 480 ms (117 ms for migration and 363 ms for main phase)
Index build time with healing (4% updated): 245+506 = 751 ms
Index build time with healing (8% updated): 660+725 = 1385 ms

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -12,6 +12,7 @@ use common::counter::hardware_counter::HardwareCounterCell;
 #[cfg(target_os = "linux")]
 use common::cpu::linux_low_thread_priority;
 use common::ext::BitSliceExt as _;
+use common::flags::FeatureFlags;
 use common::types::{PointOffsetType, ScoredPointOffset, TelemetryDetail};
 use itertools::{EitherOrBoth, Itertools as _};
 use log::debug;
@@ -27,6 +28,7 @@ use super::gpu::gpu_insert_context::GpuInsertContext;
 #[cfg(feature = "gpu")]
 use super::gpu::gpu_vector_storage::GpuVectorStorage;
 use super::graph_links::GraphLinksFormat;
+use super::links_container::LinksContainer;
 use crate::common::BYTES_IN_KB;
 use crate::common::operation_error::{OperationError, OperationResult, check_process_stopped};
 use crate::common::operation_time_statistics::{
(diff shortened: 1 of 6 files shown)


lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -1454,24 +1473,47 @@ impl<'a> OldIndexCandidate<'a> {
...
-                        // FIXME: Not supported yet.
-                        return None;
+                        missing_points += 1;
+                        feature_flags.hnsw_healing.then_some(())?;
                     }
                 }
             }
         }
 
+        for old_offset in 0..old_to_new.len() {
```
generall, line 1483: This section requires a comment

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -1501,19 +1547,59 @@ impl OldIndex<'_> {
...
+        &self,
+        src_old: PointOffsetType,
+        builder: &GraphLayersBuilder,
+    ) -> OperationResult<()> {
         let links = &self.index.graph.links;
         let point_level = links.point_level(src_old);
 
+        let hardware_counter = HardwareCounterCell::disposable(); // Internal operation. No measurements needed
+        let vector_storage = self.index.vector_storage.borrow();
+        let old_scorer = new_raw_scorer(
```
generall, line 1565: this requires comment

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -1501,19 +1547,59 @@ impl OldIndex<'_> {
...
+            vector_storage.get_vector(src_old).as_vec_ref().into(),
+            &vector_storage,
+            hardware_counter,
+        )?;
+
         let links = (0..=point_level).map(|level| {
```
generall, line 1571: `links` -> `links_per_level`, otherwise it is not clear without a typehint

generall (commented):
There are a lot of thing which require more detailed explanation. I propose to schedule a call to comment stuff together.

generall (approved):

PR #6545, opened 2025-05-16
Add `partial_snapshot_create_requests` and `*_recovery_timestamp` to telemetry

ffuugoo:
This PR adds two partial snapshot fields to replica set telemetry:
- `ongoing_create_snapshot_requests ` – counts concurrent create partial snapshot requests
- `recovery_timestamp ` – tracks UNIX timestamp of the last successfully recovered partial snapshot

These fields are needed so that cluster-manager can apply limits on partial snapshot creation/recovery.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
+++ b/lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
@@ -84,7 +84,7 @@ impl PartialSnapshotMeta {
     }
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Debug, Default)]
 pub struct RequestTracker {
     requests: Arc<AtomicUsize>,
 }
--- a/lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
+++ b/lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
@@ -1,4 +1,6 @@
 use std::sync::Arc;
+use std::sync::atomic::{self, AtomicU64, AtomicUsize};
+use std::time::{Duration, SystemTime};
 
 use crate::operations::types::{CollectionError, CollectionResult};
 
@@ -25,13 +27,20 @@ use crate::operations::types::{CollectionError, CollectionResult};
 ///
 #[derive(Debug, Default)]
 pub struct PartialSnapshotMeta {
+    ongoing_create_snapshot_requests_tracker: RequestTracker,
     recovery_lock: Arc<tokio::sync::Mutex<()>>,
     search_lock: Arc<tokio::sync::RwLock<()>>,
+    recovery_timestamp: AtomicU64,
 }
 
 impl PartialSnapshotMeta {
-    pub fn new() -> Self {
-        Self::default()
+    pub fn ongoing_create_snapshot_requests(&self) -> usize {
+        self.ongoing_create_snapshot_requests_tracker.requests()
+    }
+
+    pub fn track_create_snapshot_request(&self) -> RequestGuard {
+        self.ongoing_create_snapshot_requests_tracker
+            .track_request()
     }
 
     pub fn try_take_recovery_lock(&self) -> CollectionResult<tokio::sync::OwnedMutexGuard<()>> {
@@ -54,5 +63,52 @@ impl PartialSnapshotMeta {
                 error: "shard unavailable, partial snapshot recovery is in progress".into(),
                 backtrace: None,
             })
+    }
+
+    pub fn recovery_timestamp(&self) -> u64 {
+        self.recovery_timestamp.load(atomic::Ordering::Relaxed)
+    }
+
+    pub fn snapshot_recovered(&self) {
+        let timestamp = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap_or(Duration::ZERO)
+            .as_secs();
+
+        self.recovery_timestamp
+            .store(timestamp, atomic::Ordering::Relaxed);
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct RequestTracker {
+    requests: Arc<AtomicUsize>,
+}
+
+impl RequestTracker {
+    pub fn requests(&self) -> usize {
+        self.requests.load(atomic::Ordering::Relaxed)
+    }
+
+    pub fn track_request(&self) -> RequestGuard {
+        RequestGuard::new(self.requests.clone())
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct RequestGuard {
+    requests: Arc<AtomicUsize>,
+}
+
+impl RequestGuard {
+    fn new(requests: Arc<AtomicUsize>) -> Self {
+        requests.fetch_add(1, atomic::Ordering::Relaxed);
+        Self { requests }
+    }
+}
+
+impl Drop for RequestGuard {
+    fn drop(&mut self) {
+        self.requests.fetch_sub(1, atomic::Ordering::Relaxed);
     }
 }
--- a/lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
+++ b/lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
@@ -27,9 +27,13 @@ use crate::operations::types::{CollectionError, CollectionResult};
 ///
 #[derive(Debug, Default)]
 pub struct PartialSnapshotMeta {
+    /// Tracks ongoing operations of creation of partial snapshots.
+    /// There might be multiple parallel requests to create partial snapshots,
+    /// so we track them with a counter.
     ongoing_create_snapshot_requests_tracker: RequestTracker,
     recovery_lock: Arc<tokio::sync::Mutex<()>>,
     search_lock: Arc<tokio::sync::RwLock<()>>,
+    /// Timestamp of the last successful snapshot recovery.
     recovery_timestamp: AtomicU64,
 }
 
--- a/lib/collection/src/shards/telemetry.rs
+++ b/lib/collection/src/shards/telemetry.rs
@@ -21,6 +21,10 @@ pub struct ReplicaSetTelemetry {
     pub remote: Vec<RemoteShardTelemetry>,
     #[anonymize(with = anonymize_collection_with_u64_hashable_key)]
     pub replicate_states: HashMap<PeerId, ReplicaState>,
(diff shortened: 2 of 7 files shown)


ffuugoo:
Here's some pseudo-code that shows how these fields can be used in CM to track/control the number of partial snapshot recovery requests across the nodes:

<details>
<summary>Example</summary>

```rust
// Get telemetry for all nodes in read cluster
let read_cluster_telemetry = get_read_cluster_telemetry().await?;

// Filter local replicas
let mut read_replicas: Vec<_> = read_cluster_telemetry
    .into_iter()
    .filter(|replica_set_tele| replica_set_tele.local.is_some())
    .collect();

// Sort local replicas by `partial_snapshot_recovery_timestamp`
read_replicas
    .sort_by_key(|replica_set_tele| replica_set_tele.partial_snapshot_recovery_timestamp);


// Get telemetry for all nodes in write cluster
let write_cluster_telemetry = get_write_cluster_telemetry().await?;

// Filter local replicas that are not creating partial snapshot already
let mut available_write_replicas: HashSet<_> = write_cluster_telemetry
    .into_iter()
    .filter(|replica_set_tele| {
        replica_set_tele.local.is_some() && replica_set_tele.partial_snapshot_create_requests
    })
    .collect();


// Read-replica to write-replica mapping
let read_write_map: HashMap<(peer_id, shard_id), (peer_id, shard_id)> = todo!();

// Start `recover_from` requests
for (read_peer_id, read_shard_id) in &read_replicas {
    let (write_peer_id, write_shard_id) = read_write_map
        .get(&(read_peer_id, read_shard_id))
        .expect("read/write mapping exist for all replicas");
    
    if !available_write_replicas.contains(&(write_peer_id, write_shard_id)) {
        // Write replica is creating partial snapshot already
        continue;
    }
    
    // Call `recover_from?wait=false` on read peer
    let result = try_recover_from(read_peer_uri, write_peer_uri);
    
    match result {
        Ok(()) => {
            // Recovery started successfully, remove write peer from list of available write peers
            available_write_replicas.remove(&(write_peer_id, write_shard_id));
        }
        
        Err(err) if err.is_partial_snapshot_recovery_already_in_progress() => {
            // Partial snapshot recovery is already in progress on read peer
            continue;
        }
        
        Err(err) => {
            // Other error
            return Err(err);
        }
    }
}
```

</details>

lib/collection/src/shards/telemetry.rs
```diff
@@ -21,6 +21,10 @@ pub struct ReplicaSetTelemetry {
     pub remote: Vec<RemoteShardTelemetry>,
     #[anonymize(with = anonymize_collection_with_u64_hashable_key)]
     pub replicate_states: HashMap<PeerId, ReplicaState>,
+    #[anonymize(false)]
+    pub partial_snapshot_create_requests: usize,
+    #[anonymize(false)]
+    pub partial_snapshot_recovery_timestamp: u64,
```
ffuugoo, line 27: LMK if I should move these fields around or rename them.

lib/collection/src/shards/telemetry.rs
```diff
@@ -21,6 +21,10 @@ pub struct ReplicaSetTelemetry {
     pub remote: Vec<RemoteShardTelemetry>,
     #[anonymize(with = anonymize_collection_with_u64_hashable_key)]
     pub replicate_states: HashMap<PeerId, ReplicaState>,
+    #[anonymize(false)]
+    pub ongoing_create_partial_snapshot_requests: usize,
```
generall, line 25: It should be optional, so that new version of generated client would still be capable of processing response of the old version where those fields are missing.

lib/collection/src/shards/telemetry.rs
```diff
@@ -21,6 +21,10 @@ pub struct ReplicaSetTelemetry {
...
+    #[anonymize(false)]
+    pub partial_snapshot_recovery_timestamp: u64,
```
generall, line 27: same, should be optional

lib/collection/src/shards/telemetry.rs
ffuugoo: Noted. Will fix.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -40,4 +49,61 @@ impl PartialSnapshotMeta {
...
+            .as_secs();
+
+        self.recovery_timestamp
+            .store(timestamp, atomic::Ordering::Relaxed);
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct RequestTracker {
+    requests: Arc<AtomicUsize>,
```
generall, line 76: how it can be more than 1?

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -1,29 +1,38 @@
 use std::sync::Arc;
+use std::sync::atomic::{self, AtomicU64, AtomicUsize};
+use std::time::{Duration, SystemTime};
 
 use crate::operations::types::{CollectionError, CollectionResult};
 
 #[derive(Debug)]
 pub struct PartialSnapshotMeta {
+    ongoing_create_snapshot_requests_tracker: RequestTracker,
```
generall, line 9: Feels like an overkill. I would rather make state boolean and use single read lock as a source of truth

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
ffuugoo: `create_partial_snapshot` is **not** limited to one request at a time. Only `recover_from` is.

If you call `recover_from` from two different read nodes (with the same write node url), two create_partial_snapshot requests may be called on write node at the same time.

lib/collection/src/shards/telemetry.rs
ffuugoo: Done.

lib/collection/src/shards/telemetry.rs
ffuugoo: Done.

lib/collection/src/shards/telemetry.rs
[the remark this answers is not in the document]
ffuugoo: Done.

generall (approved):

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
ffuugoo: Discussed on call.

TL;DR: This tracks **create** partial snapshot requests (*not* recover or recover-from), and there could be multiple ongoing create partial snapshot requests at the same time (there are no locks in create snapshot path), so we track with a counter. In practice, this field will be used by cluster-manager to ensure we only create 1 (or other configured number?) partial snapshot at a time, but it still have to be a *counter* (not a bool) unless we add a lock to create partial snapshot API as well.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
ffuugoo: See answer to the next comment.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -55,4 +68,51 @@ impl PartialSnapshotMeta {
...
+    requests: Arc<AtomicUsize>,
+}
+
+impl RequestGuard {
+    fn new(requests: Arc<AtomicUsize>) -> Self {
+        requests.fetch_add(1, atomic::Ordering::Relaxed);
+        Self { requests }
+    }
+}
+
```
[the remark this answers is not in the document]
ffuugoo: Thanks! Good bot!

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -55,4 +68,51 @@ impl PartialSnapshotMeta {
...
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap_or(Duration::ZERO)
+            .as_secs();
+
+        self.recovery_timestamp
+            .store(timestamp, atomic::Ordering::Relaxed);
+    }
+}
+
+#[derive(Clone, Debug, Default)]
```
ffuugoo, line 87:
```suggestion
#[derive(Debug, Default)]
```

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -55,4 +68,51 @@ impl PartialSnapshotMeta {
...
+            .as_secs();
+
+        self.recovery_timestamp
+            .store(timestamp, atomic::Ordering::Relaxed);
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct RequestTracker {
+    requests: Arc<AtomicUsize>,
```
timvisee, line 89: We _can_ also use [`Arc`](https://doc.rust-lang.org/std/sync/struct.Arc.html) for this, give out weak references, and count using [`weak_count`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.weak_count). This way we don't reinvent the wheel and don't make mistakes.

ffuugoo: Valid idea, but it's more obscure than explicit atomic, IMO.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -55,4 +68,51 @@ impl PartialSnapshotMeta {
...
+    pub fn requests(&self) -> usize {
+        self.requests.load(atomic::Ordering::Relaxed)
+    }
+
+    pub fn track_request(&self) -> RequestGuard {
+        RequestGuard::new(self.requests.clone())
+    }
+}
+
+#[derive(Clone, Debug)]
```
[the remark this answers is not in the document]
ffuugoo: Fixed.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
timvisee: Fair enough. I'd also consider using Arc here more of a hack.

timvisee (approved):

PR #6548, opened 2025-05-19
Add setting to limit max number of collections

generall:
```
QDRANT__STORAGE__MAX_COLLECTIONS=1000
```

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -238,6 +238,10 @@ storage:
       # Max oversampling value allowed in search.
       search_max_oversampling: null
 
+  # Maximum number of collections allowed to be created
+  # If null - no limit.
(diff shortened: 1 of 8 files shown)


timvisee (approved):
:+1: 

I also added the property to the configuration (`null`). If you don't like it, you can drop the last commit :smiley:

KShivendu (approved):

agourlay:
Can we somehow test this with the consensus test infra by passing the env. variable?

generall:
> Can we somehow test this with the consensus test infra by passing the env. variable?

Tested manually, works as expected, assuming that `QDRANT__STORAGE__MAX_COLLECTIONS` value is the same on all nodes

PR #6549, opened 2025-05-19
Always run cancellation handle for interrupted optimizations

agourlay:
Seems to fix Crasher, will let CI chew on it.

This PR makes sure we handle the on going proxy segments when failing with any error to build a new optimized segment.

### DIFF
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -675,10 +675,7 @@ pub trait SegmentOptimizer {
         ) {
             Ok(segment) => segment,
             Err(error) => {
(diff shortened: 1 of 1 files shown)


timvisee (commented):
There's still some places where we might error without handling cancellation, for example: https://github.com/qdrant/qdrant/blob/d36e0dcb9a540fd5a23ba57c05cedbcbcb846345/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs#L713-L728

I think it'd be best to handle those too, possibly by moving everything that works with proxies into a sub-function.

timvisee:
I suggest this: https://github.com/qdrant/qdrant/pull/6557

That way we handle cancellation on all errors in a single place.

timvisee (approved):

Issue #6539, opened 2025-05-15
Flaky `test_shard_wal_delta_transfer_abort_and_retry`


agourlay:
This test is failing often on CI.

```
 FAILED tests/consensus_tests/test_shard_wal_delta_transfer.py::test_shard_wal_delta_transfer_abort_and_retry - Exception: Timeout waiting for condition check_collection_shard_transfer_progress to be satisfied in 30 seconds
```

```
=================================== FAILURES ===================================
________________ test_shard_wal_delta_transfer_abort_and_retry _________________

tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-1/test_shard_wal_delta_transfer_2')

    def test_shard_wal_delta_transfer_abort_and_retry(tmp_path: pathlib.Path):
        assert_project_root()
    
        # Prevent automatic recovery on restarted node, so we can manually recover with a specific transfer method
        env={
            "QDRANT__STORAGE__PERFORMANCE__INCOMING_SHARD_TRANSFERS_LIMIT": "0",
            "QDRANT__STORAGE__PERFORMANCE__OUTGOING_SHARD_TRANSFERS_LIMIT": "0",
        }
    
        # seed port to reuse the same port for the restarted nodes
        peer_api_uris, peer_dirs, bootstrap_uri = start_cluster(tmp_path, 3, 20000, extra_env=env)
    
        create_collection(peer_api_uris[0], shard_number=1, replication_factor=3)
        wait_collection_exists_and_active_on_all_peers(
            collection_name=COLLECTION_NAME,
            peer_api_uris=peer_api_uris
        )
    
        transfer_collection_cluster_info = get_collection_cluster_info(peer_api_uris[0], COLLECTION_NAME)
        receiver_collection_cluster_info = get_collection_cluster_info(peer_api_uris[2], COLLECTION_NAME)
    
        from_peer_id = transfer_collection_cluster_info['peer_id']
        to_peer_id = receiver_collection_cluster_info['peer_id']
    
        # Start pushing points to the cluster
        upload_process_1 = run_update_points_in_background(peer_api_uris[0], COLLECTION_NAME, init_offset=100000, throttle=True)
        upload_process_2 = run_update_points_in_background(peer_api_uris[1], COLLECTION_NAME, init_offset=200000, throttle=True)
        upload_process_3 = run_update_points_in_background(peer_api_uris[2], COLLECTION_NAME, init_offset=300000, throttle=True)
    
        sleep(1)
    
[...]

PR #6550, opened 2025-05-19
Fix WAL delta transfer flakiness

timvisee:
Attempts to fix <https://github.com/qdrant/qdrant/issues/6539>

Fixes flakiness in the WAL delta transfer tests.

In the test, transfers are very short. The transfer may already be completed when we're waiting on a specific state to synchronize the test. This was causing flakiness.

Instead of waiting for specific shard transfer progress, just wait for any transfer to be active.

I can't be 100% this will fix flakiness in CI, but I've been seeing good results locally. Over 100 runs I now have 0 failures. If it still appears to be flaky, we can take further action.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/tests/consensus_tests/test_shard_wal_delta_transfer.py
+++ b/tests/consensus_tests/test_shard_wal_delta_transfer.py
@@ -202,7 +202,7 @@ def test_shard_wal_delta_transfer_manual_recovery(tmp_path: pathlib.Path):
     assert_http_ok(r)
 
     # Assert WAL delta transfer progress, and wait for it to finish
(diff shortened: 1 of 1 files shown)


KShivendu:
> I can't be 100% this will fix flakiness in CI, but I've been seeing good results locally. Over 100 runs I now have 0 failures. If it still appears to be flaky, we can take further action

Run this particular test on CI 100 times by temporarily modifying the GH workflow/script. It should prove that it's fixed on CI as well.

generall (approved):

PR #6560, opened 2025-05-20
Strict mode error is now 400 instead of 403

generall:


timvisee (approved):
Motivation for this change:

By returning HTTP 400 on strict mode errors, we can distinguish it from API key errors which uses HTTP 401/403.

PR #6561, opened 2025-05-20
Add volatile dense vector storage, use in tests

timvisee:
Add a new dense vector storage type. The storage is volatile, purely in memory, and is intended for use in testing.

This brings us a tiny step closer to fully removing RocksDB from our code base.

We were using the RocksDB based dense vector storage as baseline in many of our tests. It must be replaced so that RocksDB can be removed at some point.

It is possible to replace them with the appendable mmap storage. However, I prefer this volatile storage because it is much simpler and doesn't require any IO.

If we deem this acceptable, I'll also implement this for other storage types if necessary.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/vector_storage/dense/mod.rs
+++ b/lib/segment/src/vector_storage/dense/mod.rs
@@ -3,3 +3,5 @@ pub mod dynamic_mmap_flags;
 pub mod memmap_dense_vector_storage;
 pub mod mmap_dense_vectors;
 pub mod simple_dense_vector_storage;
+#[cfg(test)]
+pub mod volatile_dense_vector_storage;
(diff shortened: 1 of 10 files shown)


lib/segment/src/vector_storage/dense/mod.rs
```diff
@@ -3,3 +3,5 @@ pub mod dynamic_mmap_flags;
 pub mod memmap_dense_vector_storage;
 pub mod mmap_dense_vectors;
 pub mod simple_dense_vector_storage;
+#[cfg(test)]
+pub mod volatile_dense_vector_storage;
```
timvisee, line 7: New storage is marked with `#[cfg(test)]` so that we cannot use it in a release build.

generall (approved):

PR #6563, opened 2025-05-20
Add inference usage tracking to REST API

n0x29a:


lib/api/src/rest/conversions.rs
```diff
@@ -7,8 +7,25 @@ use super::{
     FacetRequestInternal, FacetResponse, FacetValue, FacetValueHit, NearestQuery, OrderByInterface,
     Query, QueryInterface, VectorOutput, VectorStructOutput,
 };
+use crate::grpc::{InferenceUsage as GrpcInferenceUsage, ModelUsage};
```
coszio, line 10: nit: for consistency, I'd rather use namespace alias, as it is less verbose when dealing with multiple types. This way it is not necessary to add one alias per type
```rust
use crate::grpc as grpc;

impl From<InferenceUsage> for grpc::InferenceUsage {
...
```

lib/api/src/grpc/proto/points.proto
```diff
@@ -1171,9 +1171,22 @@ message GeoPoint {
...
+// ------------ Inference measurements ----------
+// ---------------------------------------------
+
+message InferenceUsage {
+  map<string, ModelUsage> model = 1;
+}
+
+message ModelUsage {
+  uint64 tokens = 1;
+}
```
coszio, line 1188: Is this only intended for text? how will we represent measurements about other modalities like images?

coszio (changes requested):
I didn't notice before for a lack of being in the loop, but I believe there is no need to duplicate the inference handling twice (one for grpc and one for rest) since we already have `CollectionQueryRequest`, which acts as a direct converging type from both interfaces, exactly with this kind of purpose.

Previously, it was introduced to handle ID -> vector resolution. I think it is the same case for Object -> vector inference.

I'm fine with dismissing this change request if there a reason it is being handled directly on the API types, I just want to make it explicit.

src/actix/api/query_api.rs
```diff
@@ -50,7 +52,8 @@ async fn query_points(
     };
     let hw_measurement_acc = request_hw_counter.get_counter();
     let result = async move {
-        let request = convert_query_request_from_rest(query_request, &inference_token).await?;
+        let CollectionQueryRequestWithUsage { request, usage } =
+            convert_query_request_from_rest(query_request, &inference_token).await?;
 
         let pass = check_strict_mode(
```
generall, line 58: Strict mode should be checked before inference.

generall: it will require implementing StrictModeVerification Trait for external type

src/actix/api/query_api.rs
```diff
@@ -80,11 +83,21 @@ async fn query_points(
             .map(api::rest::ScoredPoint::from)
             .collect_vec();
 
-        Ok(QueryResponse { points })
+        Ok((QueryResponse { points }, usage))
```
generall, line 86: we also want to return usage even in a case when operation failed on later stages.

coszio:
After discussion in chat, it makes sense to keep current approach of handling at API layer with the purpose of not introducing inference responsibility into Collection level. This dismisses my previous review 👍

lib/api/src/rest/conversions.rs
```diff
@@ -7,8 +9,27 @@ use super::{
     FacetRequestInternal, FacetResponse, FacetValue, FacetValueHit, NearestQuery, OrderByInterface,
     Query, QueryInterface, VectorOutput, VectorStructOutput,
 };
+use crate::grpc;
+use crate::rest::models::InferenceUsage;
 use crate::rest::{DenseVector, NamedVectorStruct};
 
+impl From<InferenceUsage> for grpc::InferenceUsage {
+    fn from(value: InferenceUsage) -> Self {
+        let mut grpc_usage_models = HashMap::default();
```
JojiiOfficial, line 18: Minor nit but lets use `HashMap::with_capacity` here

src/actix/api/query_api.rs
```diff
@@ -110,13 +123,19 @@ async fn query_points_batch(
...
             let QueryRequest {
                 internal,
                 shard_key,
-            } = request;
+            } = request_item;
+
+            let CollectionQueryRequestWithUsage { request, usage } =
+                convert_query_request_from_rest(internal, &inference_token).await?;
+
+            inference_usages.push(usage);
```
JojiiOfficial, line 137: Why are we collecting all usages, if we're only using only the last `InferenceUsage` item?

src/actix/api/query_api.rs
```diff
@@ -189,33 +220,44 @@ async fn query_points_groups(
...
             hw_measurement_acc,
         )
-        .await
+        .await?;
+        Ok((query_result, usage))
     }
     .await;
 
-    helpers::process_response(result, timing, request_hw_counter.to_rest_api())
+    let (result, inference_usage) = match result {
```
JojiiOfficial, line 250: This statement seems to be used a lot. Can we maybe extract it into a function?

src/common/inference/service.rs
```diff
@@ -215,23 +217,15 @@ impl InferenceService {
...
             reqwest::StatusCode::OK => {
-                let inference_response: InferenceResponse = serde_json::from_str(response_body)
+                serde_json::from_str(response_body)
                     .map_err(|e| {
                         StorageError::service_error(format!(
                             "Failed to parse successful inference response: {e}. Response body: {response_body}",
                         ))
-                    })?;
-
-                if inference_response.embeddings.is_empty() {
```
JojiiOfficial, line 228: Why don't we need this error anymore?

src/common/update.rs
```diff
@@ -662,12 +686,17 @@ pub async fn do_batch_update_points(
...
+                .await?;
+                (update_res, None)
             }
-        }?;
-        results.push(result);
+        };
+
+        results.push(current_update_result);
+        if inference_usage.is_none() && current_operation_usage_opt.is_some() {
+            inference_usage = current_operation_usage_opt;
```
JojiiOfficial, line 696: Why do we only take the first `usage` as `inference_usage` here?

lib/api/src/rest/models.rs
```diff
@@ -55,12 +56,17 @@ pub struct ApiResponse<D> {
 pub struct Usage {
     #[serde(skip_serializing_if = "Option::is_none")]
```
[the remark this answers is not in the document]
n0x29a: it will be not enough to add same condition for inference because there is no guarantee that it will be exactly None instead of Some. Because of that [there is check here](https://github.com/qdrant/qdrant/blob/d043cd1e0e0931f84e584bee4c78d5ef2e627cab/lib/api/src/rest/models.rs#L48)

src/actix/helpers.rs
```diff
@@ -28,21 +28,32 @@ pub fn get_request_hardware_counter(
     )
 }
 
-pub fn accepted_response(timing: Instant, hardware_usage: Option<HardwareUsage>) -> HttpResponse {
+pub fn accepted_response(
+    timing: Instant,
+    hardware_usage: Option<HardwareUsage>,
+    inference_usage: Option<InferenceUsage>,
+) -> HttpResponse {
+    let usage = {
```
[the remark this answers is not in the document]
n0x29a: it can't be used directly because usage_or_none expects grpc types

timvisee:
In a call we agreed that the current approach isn't very sustainable. Returning usage in all ok/err returns adds a lot of boilerplate.

We agreed to try a different approach instead, one that follows the design of our current hardware usage counters. Both concepts seem quite similar. That way we don't need to return usages everywhere, and we handle ok/error statuses in exactly the same way.

src/actix/api/query_api.rs
n0x29a: These two lines don't justify extracting a standalone function we'll never reuse, I think

generall (approved):

PR #6568, opened 2025-05-21
Fix force delete peer flaky test

KShivendu:
~I did these these~:
- ~Flipped the initial vs later upsert count to avoid extending test duration. Upserting more points in the 2nd upsert ensures that we catch the transfer while it's still ongoing~ 
- ~Solving [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) by returning last transfers after the wait fn~

Simplified the test so we don't need `get_collection_cluster_info` or `wait_for_collection_shard_transfers_count` which depend on transient states and lead to flaky behaviour. Original intention was to ensure that if peer is force deleted while transfers are ongoing, any ongoing transfers from/to the peer must be aborted.

### All Submissions:

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -36,6 +36,11 @@ jobs:
 
           poetry -C tests check --lock
           poetry -C tests install --no-root
+      - name: Build
+        run: cargo build --features "service_debug data-consistency-check" --locked
+      - name: Run integration tests
+        run: poetry -C tests run ./tests/integration-tests.sh
+        shell: bash
 
   integration-tests-consensus:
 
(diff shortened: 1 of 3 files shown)


KShivendu:
Test passed 130+ times [on CI](https://github.com/qdrant/qdrant/actions/runs/15157455379/job/42615502075). Gonna revert back commit that runs only that test on loop.

generall:
> Solving [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) by returning last transfers after the wait fn.

How exactly this solves TOCTOU if we still expecting 1 transfer?

agourlay (approved):

timvisee (approved):

PR #6571, opened 2025-05-21
Improve connectivity check in HNSW index

JojiiOfficial:
Improves connectivity checking for additional payload links in HNSW.
The changes:
 - [x] Skip connectivity measure for first payload index
 - [x] Improve BFS checking by going through the whole graph
 - [ ] Budget for checking instead of skipping first node.

### H&M Dataset
DEV:
  - Benchmark time: 4m9s
  - Mean precision: 0.99994 (vector-db-benchmark)
  - Precision in Qdrant UI: 0.945 +- 0.0766
  
PR:
  - Benchmark time: ~2m40s
  - Mean precision: 0.9928 (vector-db-benchmark)
  - Precision in Qdrant UI: 0.946 +- 0.0713

### arXiv titles

Dev:
  - Benchmark time: 16m7s
  - Mean precision: 0.999256 (vector-db-benchmark)
  - Precision in Qdrant UI: 0.989 ± 0.0313

PR:
  - Benchmark time: 14m30s
  - Mean precision 0.990795 (vector-db-benchmark)
  - Precision in Qdrant UI: 0.981 ± 0.0392


lib/segment/src/index/hnsw_index/graph_layers_builder.rs
```diff
@@ -82,85 +86,110 @@ impl GraphLayersBuilder {
...
+        loop {
+            // Points visited in the previous layer (Get used as entry point in the iteration over the next layer)
+            let mut previous_visited_points = vec![entry_point];
 
-        while let Some(current_point) = queue.pop() {
-            let links = self.links_layers[current_point as usize][layer].read();
+            // For each layer in HNSW...
+            for current_layer in (0..=entry_layer).rev() {
+                // Set entry points to visited points of previous layer.
+                if previous_visited_points.is_empty() {
```
generall, line 145: let's refactor to make sure `previous_visited_points` actually never shrinks, so we always use point as entry if we ever visited it on previous layers

generall (approved):

PR #6577, opened 2025-05-21
Switch to ENTRYPOINT instead of CMD

ii343hbka:
This way all the flags and arguments can be passed to entrypoint.sh#L21 just by docker run qdrant --example-flag arg


generall:
Hey @ii343hbka, thanks for the suggestion, but we unfortunately rely on existing variant in our infra, so we can't merge this at the moment

PR #6579, opened 2025-05-22
Actively migrate away from RocksDB based ID tracker

timvisee:
Actively migrate away from ID trackers that still use RocksDB. On segment load, it is replaced with our new [mutable ID tracker](https://github.com/qdrant/qdrant/issues/6157).

The RocksDB ID tracker has been disabled for a long time. However, there may be very old deployments with a static collection that still use it. An optimization run on all segments was required to migrate it, with this change we will now actively migrate it on segment load.

This is part of a bigger effort to fully remove RocksDB from our code base.

The migration is behind a runtime feature flag (`migrate_rocksdb_id_tracker`). We'll likely enable it by default in one of the upcoming releases.

### Tasks
- [ ] Test

### All Submissions:

### New Feature Submissions:


lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -745,9 +745,15 @@ fn create_segment_id_tracker(
...
-            )));
+            let id_tracker = create_rocksdb_id_tracker(db_builder.require()?)?;
+
+            // Actively migrate RocksDB based ID tracker into mutable ID tracker
+            if feature_flags().migrate_rocksdb_id_tracker {
+                let id_tracker = migrate_rocksdb_id_tracker_to_mutable(id_tracker, segment_path)?;
+                return Ok(sp(IdTrackerEnum::MutableIdTracker(id_tracker)));
+            }
+
+            return Ok(sp(IdTrackerEnum::RocksDbIdTracker(id_tracker)));
```
timvisee, line 756: Note: this is the only place left where we load the RocksDB based ID tracker. It is loaded if it still exists on disk.

Since this is the only site, we can migrate the ID tracker here.

lib/segment/src/common/rocksdb_buffered_update_wrapper.rs
```diff
@@ -133,4 +133,8 @@ impl DatabaseColumnScheduledUpdateWrapper {
         }
         self.db.get_opt(key)
     }
+
```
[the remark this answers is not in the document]
timvisee: I don't think a comment is necessary. We're well aware and use the same mechanism in multiple places.

JojiiOfficial (approved):

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -942,3 +948,40 @@ fn load_segment_state_v5(segment_path: &Path) -> OperationResult<SegmentState> {
...
+    let mut new_id_tracker = create_mutable_id_tracker(segment_path)?;
+    assert_eq!(
+        new_id_tracker.total_point_count(),
+        0,
+        "new mutable ID tracker must be empty",
+    );
+
+    // Copy all mappings into it
+    for (external_id, internal_id) in old_id_tracker.iter_from(None) {
+        let version = old_id_tracker.internal_version(internal_id).unwrap_or(0);
```
agourlay, line 974: Why are we creating a synthetic version here?
If the version is missing it means the point is not well formed I believe.

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -745,9 +745,15 @@ fn create_segment_id_tracker(
...
 
         if use_rocksdb_mutable_tracker {
-            return Ok(sp(IdTrackerEnum::RocksDbIdTracker(
-                create_rocksdb_id_tracker(db_builder.require()?)?,
-            )));
+            let id_tracker = create_rocksdb_id_tracker(db_builder.require()?)?;
+
+            // Actively migrate RocksDB based ID tracker into mutable ID tracker
+            if common::flags::feature_flags().migrate_rocksdb_id_tracker {
+                let id_tracker = migrate_rocksdb_id_tracker_to_mutable(id_tracker, segment_path)?;
```
agourlay, line 752: what happens if we fail here?

I assume the segment won't be loaded at all instead of keeping the current rocksdb id tracker.
Is it what we want?
Can this crash loop on restart?

timvisee: Correct. The collection would fail to load.

The goal here is to be sure we don't leave anything RocksDB behind.

I'm not sure if simply chugging along on failure is better. If we'd log the problem, I'm not sure if it's ever seen. It would be a major problem if we drop the RocksDB dependency in the next release.

We could pick a phased approach:
- if migration fails, still load old tracker and print a warning
- in the next minor release: force migrate and crash if it fails
- in the minor release after: fully drop RocksDB dependency

It would prolong the whole process of removing RocksDB though, which I'm not a fan of.

Alternatively we can force optimization here. If migration fails we keep the RocksDB index, then we force the optimizer to pick up all not-yet-migrated-segments which would also move segments away from the RocksDB ID tracker. Though this is less reliable, and it feels like a hack. 

Thoughts?

timvisee: To improve it a bit I've added additional cleanup in https://github.com/qdrant/qdrant/pull/6579/commits/a1802b0e8dccc78ae72f3a6c472b2fad9c67c5dd

If migrating to the mutable ID tracker failed, it now cleans up the files it created as part of it.

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -942,3 +948,63 @@ fn load_segment_state_v5(segment_path: &Path) -> OperationResult<SegmentState> {
...
+    ) -> OperationResult<MutableIdTracker> {
+        // Construct mutable ID tracker
+        let mut new_id_tracker = create_mutable_id_tracker(segment_path)?;
+        assert_eq!(
+            new_id_tracker.total_point_count(),
+            0,
+            "new mutable ID tracker must be empty",
+        );
+
+        // Copy all mappings into it
```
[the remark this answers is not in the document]
timvisee: Handled in https://github.com/qdrant/qdrant/pull/6579/commits/39461fafdccea04bcec71360a022fe689ed2aae7

generall (approved):

agourlay (approved):

lib/segment/src/segment_constructor/segment_constructor_base.rs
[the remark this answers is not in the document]
timvisee: I agree, it's better. Thanks!

PR #6589, opened 2025-05-23
Invert feature_flags.hnsw_healing

xzfc:
The PR #6543 (commit https://github.com/qdrant/qdrant/pull/6543/commits/1481fd086ec1e3a128d0f4a61446212fd587e806) introduced a blunder typo: the healing is *enabled* by default, and `QDRANT__FEATURE_FLAGS__HNSW_HEALING=true` *disables* it. The expected behavior is other way around.

This PR fixes it.


timvisee (approved):
Good catch. Thanks!

PR #6609, opened 2025-05-28
Add mutable numeric payload index on top of Gridstore

timvisee:
Depends on <https://github.com/qdrant/qdrant/pull/6614>

As part of our bigger effort to [remove RocksDB](https://github.com/qdrant/qdrant/milestone/16), we need a replacement for our mutable payload indices. All of them currently still rely on RocksDB.

This PR introduces a mutable numeric payload index on top of Gridstore. If benchmark results are satisfactory, the other payload index types will follow in separate PRs.

It adds the `payload_index_skip_mutable_rocksdb` feature flag skip using RocksDB for mutable payload indices and start using Gridstore. The flag is disabled by default for now. Note that this PR only implements the numeric index, and all the other index types will still fallback to RocksDB.

### Tasks
- [x] Actually use Gridstore based index
- [x] Benchmark payload index on Gridstore ([see here](https://github.com/qdrant/qdrant/pull/6609#issuecomment-2930408130))
- [x] Tune Gridstore options to match values we store on it
- [x] Add test
- [x] Merge <https://github.com/qdrant/qdrant/pull/6614>
- [x] Rebase on `dev`
- [x] Undraft
- [x] ~Explicitly drop Gridstore caches after load?~

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
```diff
@@ -1,28 +1,47 @@
...
 use crate::index::field_index::mmap_point_to_values::MmapValue;
 
-pub struct MutableNumericIndex<T: Encodable + Numericable> {
-    db_wrapper: DatabaseColumnScheduledDeleteWrapper,
+const GRIDSTORE_OPTIONS: StorageOptions = StorageOptions {
+    // Largest numeric value used
+    block_size_bytes: Some(size_of::<u128>()),
+    // Compressing numeric values is unreasonable
+    compression: Some(gridstore::config::Compression::None),
+    page_size_bytes: None,
```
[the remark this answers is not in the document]
timvisee: I had yet to tune and tweak these values, but this is very helpful. Thanks!

I actually planned to create a function (not a const) for these options, so that we can pass the actual size of the values in here. In that case i64 would just take 8 bytes, as 16 is unreasonable.

I completely agree that keeping gridstore for indices as small as possible is a good target. It will only be used for loading and persisting, but still!

timvisee:
I've done a basic end to end benchmark to validate the approach in this PR is acceptable. I measure upsert time, hot/cold load time, disk usage and memory consumption for RocksDB and Gridstore storage.

#### Benchmark

- `bfb -d 4 --int-payloads 100 --int-payloads-range --max-int-payloads 10 --indexing-threshold 0 -n 1000000`:  
  _Tests 1M points on 2 unindexed segments, each point has 1 to 10 integer values in payload_
  - RocksDB:
    - Upsert:    37s, 37s, 37s (avg 37s)
    - Disk size: 35M + 35M = 70M
    - Hot load:  2.669s, 2.717s, 2.705s (avg 2.697s)
    - Cold load: 2.725s, 2.797s, 2.755s (avg 2.759s)

  - Gridstore (block size 8 bytes):
    - Upsert:    25s, 24s, 24s (avg 24s)
    - Disk size: 29M + 29M = 58M
    - Hot load:  2.050s, 2.043s, 2.040s (avg 2.044s)
    - Cold load: 2.726s, 2.734s, 2.713s (avg 2.724s)
  - Gridstore (block size 16 bytes):
    - Upsert:    25s, 25s, 24s (avg 25s)
    - Disk size: 31M + 31M = 62M
    - Hot load:  2.035s, 2.067s, 2.076s (avg 2.059s)
    - Cold load: 2.741s, 2.752s, 2.745s (avg 2.746s)
  - Gridstore (block size 64 bytes):
    - Upsert:    24s, 24s, 25s (avg 24s)
    - Disk size: 45M + 45M = 89M
    - Hot load:  2.135s, 2.148s, 2.079s (avg 2.121s)
    - Cold load: 3.152s, 3.017s, 3.014s (avg 3.061s)
  - Gridstore (block size 128 bytes):
    - Upsert:    25s, 25s, 25s (avg 25s)
    - Disk size: 69M + 69M = 138M
    - Hot load:  2.097s, 2.133s, 2.087s (avg 2.106s)
    - Cold load: 3.505s, 3.539s, 3.507s (avg 3.517s)

- `bfb -d 4 --int-payloads 100 --int-payloads-range --indexing-threshold 0 -n 2500000`  
  _Tests 2.5M points on 2 unindexed segments, each point has exactly 1 integer value in payload_
  - RocksDB:
    - Upsert:    24s, 24s, 25s (avg 25s)
    - Disk size: 20M + 20M = 40M
    - Hot load:  1522.038ms, 1515.548ms, 1531.981 (avg 1523.189ms)
    - Cold load: 1552.669ms, 1523.629ms, 1534.047ms (avg 1536.782ms)
  - Gridstore (block size 8 bytes):
    - Upsert:    20s, 21s, 21s (avg 21s)
    - Disk size: 29M + 29M = 58M
    - Hot load:  858.112ms, 852.818ms, 850.455ms (avg 853.795ms)
    - Cold load: 1503.801ms, 1496.652ms, 1520.620ms (avg 1507.024ms)

#### Memory usage

Memory usage when loading 1M points, each having 1 to 10 integer values in payload. I've measured RocksDB storage, and Gridstore with block size 8, 16, 64 and 128 bytes.

chart-1

<details>
<summary>bfb command</summary>

```bash
bfb -d 4 --int-payloads 100 --int-payloads-range --max-int-payloads 10 --indexing-threshold 0 -n 1000000
```
</details>

#### Conclusion

Looking at Gridstore, the smallest possible block size for a value type behaves best. It has faster upsert/load times and less disk usage. It makes sense because data is more compact and therefore less IO is needed.

When comparing RocksDB versus Gridstore (block size 8), the new Gridstore implementation performs better across the board though a cold load performs similar.

The memory usage measurement during load shows both implementations are equal within margin of error, so there's no unexpected memory spikes. You can see each line shifted a bit because of slightly different load times.

Based on these results I conclude that we can go through with Gridstore as backing storage. I'll pick the value size as block size (8 bytes for numbers, 16 bytes for UUID).

Let me know if I should test more scenarios, and/or if I should plot the results.

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -687,7 +711,71 @@ where
...
+
+    fn init(&mut self) -> OperationResult<()> {
+        assert!(
+            self.index.is_none(),
+            "index must be initialized exactly once",
+        );
+        self.index
+            .replace(NumericIndex::new_gridstore(self.dir.clone())?);
+        Ok(())
+    }
```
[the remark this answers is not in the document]
timvisee: I'd argue a hard assertion is fine here, since it currently is not possible to construct this in a way where it would panic.

lib/gridstore/src/blob.rs
```diff
@@ -3,3 +3,57 @@ pub trait Blob {
...
+    fn to_bytes(&self) -> Vec<u8> {
+        self.iter().flat_map(|v| v.to_bytes()).collect()
+    }
+
+    fn from_bytes(data: &[u8]) -> Self {
+        data.chunks(size_of::<T>())
+            .map(|v| T::from_bytes(v))
+            .collect()
+    }
+}
```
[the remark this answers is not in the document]
coszio: Maybe we can change this automatic implementation for the implementors of zerocopy's `FromBytes` and `IntoBytes` instead

lib/gridstore/src/blob.rs
```diff
@@ -3,3 +3,57 @@ pub trait Blob {
...
+        self.to_le_bytes().to_vec()
+    }
+
+    fn from_bytes(data: &[u8]) -> Self {
+        f64::from_le_bytes(
+            data.try_into()
+                .expect("Slice with incorrect length for f64"),
+        )
+    }
+}
```
coszio: Looks like implementation would have less repetition with a macro
```suggestion
macro_rules! impl_blob_le_bytes {
    ($type:ty) => {
        impl Blob for $type {
            fn to_bytes(&self) -> Vec<u8> {
                self.to_le_bytes().to_vec()
            }

            fn from_bytes(data: &[u8]) -> Self {
                <$type>::from_le_bytes(
                    data.try_into()
                        .expect(&format!("Slice with incorrect length for {}", stringify!($type))),
                )
            }
        }
    }
}

impl_blob_le_bytes!(i64);
impl_blob_le_bytes!(f64);
impl_blob_le_bytes!(u128);
```

coszio: Although, actually, if it is always going to be used within a `Vec<T>`, the macro can already be augmented by implementing for the vec directly. This would avoid intermediate `to_vec()` calls

lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
```diff
@@ -1,28 +1,53 @@
...
 use crate::common::operation_error::{OperationError, OperationResult};
 use crate::common::rocksdb_buffered_delete_wrapper::DatabaseColumnScheduledDeleteWrapper;
 use crate::common::rocksdb_wrapper::DatabaseColumnWrapper;
 use crate::index::field_index::histogram::{Histogram, Numericable, Point};
 use crate::index::field_index::mmap_point_to_values::MmapValue;
 
-pub struct MutableNumericIndex<T: Encodable + Numericable> {
-    db_wrapper: DatabaseColumnScheduledDeleteWrapper,
+/// Default page size used in Gridstore
+pub const DEFAULT_PAGE_SIZE_BYTES: usize = 4 * 1024 * 1024; // 4 MiB = 16 * 8192 * 32
```
coszio: I think where the number comes from could be a bit clearer 😅 
```suggestion
pub const DEFAULT_PAGE_SIZE_BYTES: usize = 16 * 8192 * 32; // 4 MiB = block_size * region_blocks * regions
```

lib/segment/src/index/field_index/numeric_index/tests.rs
```diff
@@ -14,13 +14,15 @@ const COLUMN_NAME: &str = "test";
...
 enum IndexType {
     Mutable,
+    MutableGridstore,
     Immutable,
     Mmap,
     RamMmap,
 }
 
 enum IndexBuilder {
     Mutable(NumericIndexBuilder<FloatPayloadType, FloatPayloadType>),
```
coszio, line 24: Is this another good place to tag with `#[feature("rocksdb")]`?

lib/gridstore/src/blob.rs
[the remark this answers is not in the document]
timvisee: The example coderabbit gave is bad, and doesn't show the problem. With the current implemented types there is no issue, but I agree this is a potential footgun in the future if serialized types have a different size.

I like the idea of using zerocopy as extra safeguard because it ensures the serialized length is the same. I'll give a shot on implementing it. Thanks!

lib/segment/src/index/field_index/numeric_index/tests.rs
timvisee: Eventually, yes. Nothing in the payload indices is feature flagged yet, that will be done later. More specifically, once all payload indices have a mutable counterpart that doesn't rely on RocksDB.

lib/gridstore/src/blob.rs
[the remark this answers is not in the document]
timvisee: @coszio there's a bunch of problems with relying on zerocopy here. There trait bounds make it difficult to implement it. I don't want to add a whole bunch of trait bounds to all structs because that would add severe bloat.

I came up with an intermediate trait as solution. It's definitely not great, but it does prevent the footgun explained above.

See: https://github.com/qdrant/qdrant/pull/6609/commits/22e35117f26f0e947dee14696925e56eeb476ae1

I suggest to rework these traits in a separate PR, and/or when all mutable payload indices have been implemented. Please feel free to give it a shot if you like.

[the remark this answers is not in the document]
timvisee: I changed it once again: https://github.com/qdrant/qdrant/pull/6609/commits/7bfa4c234b0de68d9afe748880c0f37fee8721a9

This simplifies quite a lot by changing the trait bounds. Now we have no additional trait.

Another changes was required because the [previous](https://github.com/qdrant/qdrant/pull/6609#discussion_r2123290868) idea could not be implemented on variable sized types. It's something I need for the map index.

coszio (approved):

lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
```diff
@@ -205,41 +234,68 @@ impl<T: Encodable + Numericable + Default> InMemoryNumericIndex<T> {
...
+            Storage::RocksDb(_) => self.load_rocksdb(),
+            Storage::Gridstore(_) => Ok(self.load_gridstore()),
+        }
+    }
+
+    /// Load from RocksDB storage
+    ///
+    /// Loads in-memory index from RocksDB storage.
+    fn load_rocksdb(&mut self) -> OperationResult<bool> {
+        let Storage::RocksDb(db_wrapper) = &self.storage else {
```
agourlay, line 291: Why would we call `load_rocksdb` on a `Storage::Gridstore`?
This looks like it should be an error maybe.

timvisee: I agree, an error makes sense here. Thanks!

agourlay: Sweet, there is the same pattern here https://github.com/qdrant/qdrant/pull/6609/files#diff-90e24743aa6da58b43b43a13e2480a9447b2fb927fb54aaa10ac6dc313aa7ffeR323

timvisee: I replaced all of them here: https://github.com/qdrant/qdrant/pull/6609/commits/b96b235c72b63fe111b38c840721bf0de75f111e

lib/common/common/src/flags.rs
```diff
@@ -35,6 +41,7 @@ impl Default for FeatureFlags {
         FeatureFlags {
             all: false,
             payload_index_skip_rocksdb: true,
+            payload_index_skip_mutable_rocksdb: false,
```
agourlay, line 48: `payload_index_skip_mutable_rocksdb: false,`

Have you tried setting this to `true` and run all tests on our CI?

lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
```diff
@@ -1,28 +1,57 @@
...
-    db_wrapper: DatabaseColumnScheduledDeleteWrapper,
+/// Default options for Gridstore storage
+const fn default_gridstore_options<T: Sized>() -> StorageOptions {
+    let block_size = size_of::<T>();
+    StorageOptions {
+        // Size of numeric values in index
+        block_size_bytes: Some(block_size),
+        // Compressing numeric values is unreasonable
+        compression: Some(gridstore::config::Compression::None),
+        page_size_bytes: Some(block_size * 8192 * 32), // 4 to 8 MiB = block_size * region_blocks * regions,
```
agourlay, line 34: Why having pages of 4 to 8MB instead of default of 32MB used for payload storage & sparse vectors?

lib/common/common/src/flags.rs
timvisee: IIRC it is true in tests already, but I can validate that to be sure.

Either way, I did already run all tests locally with the flag explicitly enabled.

lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
timvisee: Because the block size is smaller, the idea was to keep the same number of blocks in a page, rather than to put more blocks in a page and keep the same size. If I understand it correctly we now have less regions per page, this way we prevent long searches through gaps.

To be honest, these values are still a bit arbitrary. The idea is to 'tune' them once Gridstore has been implemented for all payload indices. There's still a lot of things changing, so if we tune it now, we might not get the best result in the end. Would you be okay with that?

timvisee: For each index we create, we allocate at least one page on disk. If you have an empty mutable segment, it still creates this first page. If you have many payload indices the problem becomes more noticeable.

So therefore we want to find a balance:
- the smallest possible page size to not waste disk space
- the largest possible page size to have less overhead

In this case we just scaled down the page size together with the block size.

For payload storage and sparse vectors I'm a bit less worried about it. We generally have less of these, and the data they store is larger than what we're working with here.

To be honest, these values are still a bit arbitrary. The idea is to 'tune' them once Gridstore has been implemented for all payload indices. There's still a lot of things changing, so if we tune it now, we might not get the best result in the end. Would you be okay with that?

agourlay: > For each index we create, we allocate at least one page on disk. If you have an empty mutable segment, it still creates this first page. If you have many payload indices the problem becomes more noticeable. 
In this case we just scaled down the page size together with the block size.

Thanks this makes sense!
Can you please add a comment to capture this decision please :pray:

lib/common/common/src/flags.rs
agourlay: > IIRC it is true in tests already, 

Do you mean that all unit tests run with all the features enabled?
What about integration tests?

I believe you told me once that the CI benchmark uses all feature flag, right?

lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
```diff
@@ -270,26 +419,56 @@ impl<T: Encodable + Numericable + Default> MutableNumericIndex<T> {
...
+                let hw_counter_ref = hw_counter.ref_payload_index_io_write_counter();
+                store
+                    .write()
+                    .put_value(idx, &values, hw_counter_ref)
+                    .map_err(|err| {
+                        OperationError::service_error(format!(
+                            "failed to put value in mutable numeric index gridstore: {err}"
+                        ))
+                    })?;
+                hw_cell_wb.incr_delta(size_of_val(&idx) + size_of::<T>() * values.len());
```
generall: This might be extra?

generall (approved):
Please fix hardware counter, otherwise LGTM

PR #6610, opened 2025-05-29
Tweak partial snapshot `recover` and `recover_from` response

ffuugoo:
Tweak partial snapshot `recover` and `recover_from` response to return `status: already_in_progress` instead of error.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/tests/consensus_tests/test_partial_snapshot.py
+++ b/tests/consensus_tests/test_partial_snapshot.py
@@ -80,15 +80,17 @@ def test_partial_snapshot(
     recover_partial_snapshot_from(read_peer, write_peer)
     assert_consistency(read_peer, write_peer)
 
-def test_partial_snapshot_recovery_lock(tmp_path: pathlib.Path):
+@pytest.mark.parametrize("wait", [True, False])
+def test_partial_snapshot_recovery_lock(tmp_path: pathlib.Path, wait: bool):
     assert_project_root()
 
     write_peer, read_peer = bootstrap_peers(tmp_path, 100_000)
 
     executor = concurrent.futures.ThreadPoolExecutor(max_workers = 3)
-    futures = [executor.submit(try_recover_partial_snapshot_from, read_peer, write_peer, wait = False) for _ in range(3)]
+    futures = [executor.submit(try_recover_partial_snapshot_from, read_peer, write_peer, wait) for _ in range(3)]
     responses = [future.result() for future in concurrent.futures.as_completed(futures)]
 
+    # Single partial snapshot recovery request allowed at the same time
     assert any(response.status_code == 503 for response in responses), "Subsequent partial snapshot recovery requests have to be rejected during partial snapshot recovery"
 
 def test_partial_snapshot_read_lock(tmp_path: pathlib.Path):
@@ -102,7 +104,9 @@ def test_partial_snapshot_read_lock(tmp_path: pathlib.Path):
     is_search_rejected = False
     while not recover_future.done():
         response = try_search_random(read_peer)
-        if response.status_code == 500 or response.status_code == 503:
+
+        # Shard is unavailable during partial snapshot recovery
+        if response.status_code == 500:
             is_search_rejected = True
             break
 


ffuugoo:
Should we change `ApiStatus` to `#[serde(rename_all = "kebab-case")]`?

```rust
#[derive(Debug, Serialize, JsonSchema)]
pub enum ApiStatus {
    Ok,
    Error(String),
    Accepted,
    AlreadyInProgress,
}
```

tests/consensus_tests/test_partial_snapshot.py
```diff
@@ -54,7 +54,7 @@ def test_partial_snapshot_read_lock(tmp_path: pathlib.Path):
     is_search_rejected = False
     while not recover_future.done():
         response = try_search_random(read_peer)
-        if response.status_code == 500:
+        if response.status_code == 500 or response.status_code == 503:
```
KShivendu:
```suggestion
        # can't take search read lock during partial snapshot recovery
        if response.status_code == 500:
```

There aren't multiple partial snapshot recoveries in this particular test (unlike the previous one). So we can only expect status 500. Also, would be good to add a comment.

KShivendu (approved):
LGTM with small suggestions. Thanks! 

> Should we change ApiStatus to #[serde(rename_all = "kebab-case")]? Currently, the status is already_in_progress

No. imo it's good

tests/consensus_tests/test_partial_snapshot.py
```diff
@@ -41,7 +41,7 @@ def test_partial_snapshot_recovery_lock(tmp_path: pathlib.Path):
     futures = [executor.submit(try_recover_partial_snapshot_from, read_peer, write_peer, wait = False) for _ in range(3)]
     responses = [future.result() for future in concurrent.futures.as_completed(futures)]
 
-    assert any(response.status_code == 400 for response in responses), "Subsequent partial snapshot recovery requests have to be rejected during partial snapshot recovery"
+    assert any(response.status_code == 503 for response in responses), "Subsequent partial snapshot recovery requests have to be rejected during partial snapshot recovery"
```
KShivendu, line 94:
```suggestion
    # can't take (another) shard recovery lock during partial snapshot recovery
    assert any(response.status_code == 503 for response in responses), "Subsequent partial snapshot recovery requests have to be rejected during partial snapshot recovery"
```

KShivendu: Please add test for both `wait=true/true` using `@pytest.mark.parametrize` in all the partial snapshot tests.

KShivendu (commented):
Dropped one more suggestion.

tests/consensus_tests/test_partial_snapshot.py
ffuugoo: I added 503 as a future proofing, but sure, I'll remove it. Done.

tests/consensus_tests/test_partial_snapshot.py
ffuugoo: > Please add test for both wait=true/true

I can only parameterize `test_partial_snapshot_recovery_lock`, because other two tests rely on `wait=true` for timings. E.g., they need to wait until partial snapshot recovery is completed to do some check and with `wait=false` we'd need to add a sleep or add some other check, and it won't add much to these tests, IMO.

I've added `wait=true` to `test_partial_snapshot_recovery_lock`, let's see if tests pass now. 😂

tests/consensus_tests/test_partial_snapshot.py
```diff
@@ -80,16 +80,18 @@ def test_partial_snapshot(
...
     write_peer, read_peer = bootstrap_peers(tmp_path, 100_000)
 
     executor = concurrent.futures.ThreadPoolExecutor(max_workers = 3)
-    futures = [executor.submit(try_recover_partial_snapshot_from, read_peer, write_peer, wait = False) for _ in range(3)]
+    futures = [executor.submit(try_recover_partial_snapshot_from, read_peer, write_peer, wait) for _ in range(3)]
     responses = [future.result() for future in concurrent.futures.as_completed(futures)]
 
-    assert any(response.status_code == 400 for response in responses), "Subsequent partial snapshot recovery requests have to be rejected during partial snapshot recovery"
+    # Single partial snapshot recovery request allowed at the same time
+    assert any(response.status_code == 503 for response in responses), "Subsequent partial snapshot recovery requests have to be rejected during partial snapshot recovery"
```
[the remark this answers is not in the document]
ffuugoo: garbage ai slop 🙄

PR #6618, opened 2025-05-30
[phrase match] opt: Introduce `PositionDeltas`

coszio:
The recent changes in posting lists are to be able to store a token id's positions within a document.

For example, for "the fox jumps over the dog" would map to:
```
{
  the: [0, 4],
  fox: [1],
  jumps: [2],
  over: [3],
  dog: [5],
}
```

Each list of positions will be stored in its corresponding token posting list, assigned to the point id.

Here we introduce PositionDeltas to try to compress these lists of positions as delta-encoded with simple bitpacking, by choosing between u8, u16, or u32 representation dynamically.

The format is: `| bit_width (u8) | data ([u8]) ... |`

After reading the biwidth, the data gets interpreted as an array of one of the three options.

**Note**: I implemented this, but I am not sure how much advantage this gives over just storing the plain `[u32]`, considering that an `[u32]` is zero-copy friendly. Thoughts?

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/mod.rs
+++ b/lib/segment/src/index/field_index/full_text_index/mod.rs
@@ -3,6 +3,7 @@ mod inverted_index;
 mod mmap_inverted_index;
 pub mod mmap_text_index;
 mod mutable_text_index;
+mod position_deltas;
 mod posting_list;
 mod postings_iterator;
 pub mod text_index;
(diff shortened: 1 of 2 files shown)


timvisee (commented):
@coszio and I agreed to give this a shot once all the other bits are implemented. Then we can more easily measure how the index behaves, to see whether implementing this makes sense.

This PR is quite self contained so that should be doable. :+1:

coszio:
This PR is no longer actual, and such an optimization would require building a new version of the index. 

Will close it for now

PR #6617, opened 2025-05-30
[phrase match] `ParsedQuery` as enum

coszio:
Builds on top of #6486 

This is merely a refactoring to allow (with less changes) introducing phrase filtering as a separate way to use the full-text index.

PR #6619, opened 2025-05-30
[phrase match] Generic mmap postings

coszio:
Builds on top of #6617

Make `MmapPostings` generic over the stored value along with the ids. 

To achieve this, it was necessary to create a trait specific to full text index (`MmapPostingValue`), so that it is still compatible with previous implementation, but can also store positions if the right type is used.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
+++ b/lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
@@ -13,7 +13,7 @@ use memory::madvise::{Advice, AdviceSetting, Madviseable};
 use memory::mmap_ops::open_read_mmap;
 use posting_list::{
     PostingChunk, PostingList, PostingListComponents, PostingListView, PostingValue,
-    RemainderPosting, ValueHandler,
+    RemainderPosting, SizedTypeFor, ValueHandler,
 };
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout, Unaligned};
 
@@ -21,8 +21,6 @@ use crate::index::field_index::full_text_index::inverted_index::TokenId;
 use crate::index::field_index::full_text_index::positions::Positions;
 
 const ALIGNMENT: usize = 4;
-
-pub type JustIds = ();
 
 /// Trait marker to enrich [`posting_list::PostingValue`] for handling mmap files with the posting list.
 pub(in crate::index::field_index::full_text_index) trait MmapPostingValue:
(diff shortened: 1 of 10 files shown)


lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
```diff
@@ -10,14 +12,98 @@ use memmap2::Mmap;
...
+/// Trait marker to enrich [`posting_list::PostingValue`] for handling mmap files with the posting list.
+pub(in crate::index::field_index::full_text_index) trait MmapPostingValue:
+    PostingValue<
+    Handler: ValueHandler<Sized: FromBytes + Immutable + IntoBytes + KnownLayout + Unaligned>
+                 + Clone
+                 + Debug,
+>
+{
+    /// Length in bytes of the variable size data section.
+    fn var_size_data_len(var_size_data: &<Self::Handler as ValueHandler>::VarSizeData) -> usize;
```
generall: As I understood from the call review, `VarSizeData` can either be `()` or `[u8]`, and if we always use `[u8]`, the price is that we spend extra 16bytes per posting list. But I think it is worth paying, cause otherwise complexity of the code is quite significant.

lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
```diff
@@ -83,45 +176,56 @@ impl MmapPostings {
...
 
         let (last_doc_id, bytes) = PointOffsetType::read_from_prefix(bytes).ok()?;
 
-        counter.incr_delta(size_of::<PostingChunk<()>>());
+        counter.incr_delta(size_of::<PostingChunk<<V::Handler as ValueHandler>::Sized>>());
         let (chunks, bytes) =
-            <[PostingChunk<()>]>::ref_from_prefix_with_elems(bytes, header.chunks_count as usize)
-                .ok()?;
-        let (data, bytes) = bytes.split_at(header.data_bytes_count as usize);
+            <[PostingChunk<<V::Handler as ValueHandler>::Sized>]>::ref_from_prefix_with_elems(
```
generall: If possible, let's try to use type alias for `V::Handler as ValueHandler>::Sized>` as it becomes bulky. Also, it would be nic eto mention in comments next to the alias definition an example of typical type which is actually going to be used there, as it simplifies intuition

lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
coszio: simplified in https://github.com/qdrant/qdrant/pull/6619/commits/a06ab78081cc5a037af2528df5a98ba8bfd8d0a4

lib/segment/src/index/field_index/full_text_index/mmap_inverted_index/mmap_postings.rs
coszio: Also handled in https://github.com/qdrant/qdrant/pull/6619/commits/a06ab78081cc5a037af2528df5a98ba8bfd8d0a4

PR #6620, opened 2025-05-30
[phrase matching] expose `phrase_matching` flag in rest and grpc

coszio:
Builds on top of #6670  

This exposes the phrase matching feature, so let's merge only until necessary.


coszio:
Okay, gonna merge into https://github.com/qdrant/qdrant/pull/6670, it should still leave a manageable size.

PR #6622, opened 2025-06-02
Fix vector_io_read for multivector recommendation

agourlay:
This PR fixes the `vector_io_read` hardware counter when using a recommendation query with several examples against a quantized multivector storage.

The root of the issue is the multivector storage wrapping a quantized storage where the hardware measurement is taking place.

However, the quadratic scoring happening for multivectors is triggering an excess of IO measurement in the quantized storage.

Long story short, the same vectors are fetched several times in a row but won't show up in the kernel IO measurement because of caching.

The proposed fix is to perform the hardware measurement within the multivector storage instead.

## Example 

A recommend `average_score` query with 5 positives examples which are getting merged..

Each example is a multivector with 4 vectors forming an internal query of 20 vectors.

Each sub-vector from the query will be executed against each sub-vector of the target point (80 times in total).

It means each sub-vector of the target point is fetched 20 times.

I have seen this effect in action while scoring 5080 points from this example which used to be reported as 52MB `vector_io_read`.

It is now back to the expected value `5080 multivec * 4 sub-vec * 128 dim * 1 bytes (unint8) = 2,6 MB`.

## Pricer model

The issue has been originally reported by [Pricer](https://github.com/qdrant/pricer).

Below the results for 10k experiments before and after the fix where the outliers on the left hand disappear.

| Before        | After       |
| ------------- | ------------- |
|  |  |

### DIFF
--- a/lib/segment/src/vector_storage/quantized/quantized_multi_custom_query_scorer.rs
+++ b/lib/segment/src/vector_storage/quantized/quantized_multi_custom_query_scorer.rs
@@ -1,25 +1,31 @@
-use std::borrow::Cow;
 use std::marker::PhantomData;
 
 use common::counter::hardware_counter::HardwareCounterCell;
 use common::types::{PointOffsetType, ScoreType};
 
+use crate::data_types::named_vectors::CowMultiVector;
 use crate::data_types::primitive::PrimitiveVectorElement;
-use crate::data_types::vectors::{DenseVector, TypedDenseVector};
+use crate::data_types::vectors::{MultiDenseVectorInternal, TypedMultiDenseVector};
 use crate::spaces::metric::Metric;
 use crate::types::QuantizationConfig;
 use crate::vector_storage::query::{Query, TransformInto};
 use crate::vector_storage::query_scorer::QueryScorer;
 
-pub struct QuantizedCustomQueryScorer<'a, TElement, TMetric, TEncodedQuery, TEncodedVectors, TQuery>
-where
+pub struct QuantizedMultiCustomQueryScorer<
+    'a,
+    TElement,
+    TMetric,
+    TEncodedQuery,
+    TEncodedVectors,
+    TQuery,
+> where
     TElement: PrimitiveVectorElement,
     TMetric: Metric<TElement>,
     TEncodedVectors: quantization::EncodedVectors<TEncodedQuery>,
     TQuery: Query<TEncodedQuery>,
 {
     query: TQuery,
-    quantized_storage: &'a TEncodedVectors,
+    quantized_multivector_storage: &'a TEncodedVectors,
     phantom: PhantomData<TEncodedQuery>,
     metric: PhantomData<TMetric>,
     element: PhantomData<TElement>,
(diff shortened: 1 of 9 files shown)


agourlay:
Updated new results measurements to 10k experiments for a fair comparison.

lib/segment/src/vector_storage/quantized/quantized_multivector_storage.rs
```diff
@@ -172,13 +172,26 @@ where
...
             let mut max_sim = ScoreType::NEG_INFINITY;
             // manual `max_by` for performance
             for i in 0..offset.count {
-                let sim =
-                    self.quantized_storage
-                        .score_point(inner_query, offset.start + i, hw_counter);
+                let sim = self.quantized_storage.score_point(
+                    inner_query,
+                    offset.start + i,
+                    &HardwareCounterCell::disposable(), // compute hardware cost outside to avoid duplicated io cost
```
JojiiOfficial: Let's put this disposed HwCounterCell into a variable outside of the loop and share the reference across iterations. `HardwareCounterCell::disposable()` has a slight overhead and we don't need to create a new one on each iteration here.

lib/segment/src/vector_storage/query_scorer/sparse_custom_query_scorer.rs
```diff
@@ -60,6 +60,7 @@ impl<TVectorStorage: SparseVectorStorage, TQuery: Query<SparseVector>> QueryScor
             .get_sparse(idx)
             .expect("Failed to get sparse vector");
 
+        // not exactly correct for Gridstore where the indices are compressed into u8
```
JojiiOfficial, line 63: Do we not care about the wrong measurements here or is the comment suggesting that we need to fix this in a separate PR?

agourlay: I found discrepancy in sparse vector hardware measurements because of this fact.

Took me a while to pin-point so I decided to capture the knowledge there.

Not sure if we want to fix it already for the time being.

generall:
I would wait for the version without disposable, if it is possible

agourlay:
I do have a version without disposable locally which works fine for the quantized multivectors case.
But I am afraid it breaks the metrics for the dense vector quantization case.

Investigating right now :mag:

lib/segment/src/vector_storage/quantized/quantized_multi_custom_query_scorer.rs
```diff
@@ -0,0 +1,127 @@
...
+                let original_vector_prequantized = TElement::quantization_preprocess(
+                    quantization_config,
+                    TMetric::distance(),
+                    &original_vector.flattened_vectors,
+                );
+                Ok(quantized_multivector_storage.encode_query(&original_vector_prequantized))
+            })
+            .unwrap();
+
+        hardware_counter.set_cpu_multiplier(size_of::<TElement>());
```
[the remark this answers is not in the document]
agourlay: `std::mem::size_of ` is in the prelude in Rust 2024 edition

agourlay:
The latest commit https://github.com/qdrant/qdrant/pull/6622/commits/81374af76220ab5c31298400855357f3fb5b0668 is implementing the fix without the use of a disposable counter.

The increment of IO counters is removed from all non multivector quantized storage.
To not break measurements, it now happens instead above, at the level of the query scorer.

`QuantizedQueryScorer (IO)` -> `QuantizedStorage (CPU)`

The quantized storage are  now responsible for exposing the quantized vector size

However, part of the fix is having the multivector quantized storage reports the IO counter as well.

To avoid double IO counting at the query scorer level, a new type of specialized quantized multivector query scorer is introduced where no measurement is performed.

`QuantizedMultiQueryScorer` -> `MultivectorQuantizedStorage (IO)` -> `QuantizedStorage (CPU)`

This was tested manually.

lib/segment/src/vector_storage/quantized/quantized_scorer_builder.rs
```diff
@@ -201,7 +203,7 @@ impl<'a> QuantizedScorerBuilder<'a> {
 
     fn new_multi_quantized_scorer<TElement, TMetric, TEncodedQuery>(
         self,
-        quantized_storage: &'a impl EncodedVectors<TEncodedQuery>,
+        quantized_multivector_storage: &'a impl EncodedVectors<TEncodedQuery>,
```
agourlay, line 206: renamed for clarity

generall (approved):

PR #6635, opened 2025-06-04
Add stopwords support

n0x29a:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/segment/src/index/field_index/full_text_index/stop_words.rs
```diff
@@ -0,0 +1,296 @@
...
+            for word in ENGLISH_STOPWORDS {
+                stopwords.insert((*word).to_string());
+            }
+        }
+        // todo: Add support for other languages from https://huggingface.co/Qdrant/bm25/tree/main
+    }
+}
+
+// English stopwords todo: replace with https://huggingface.co/Qdrant/bm25/tree/main
+const ENGLISH_STOPWORDS: &[&str] = &[
```
generall: please make it a separate file

lib/segment/src/index/field_index/full_text_index/stop_words.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    /// Check if a token is a stopword
+    pub fn is_stopword(&self, token: &str) -> bool {
+        self.stopwords.contains(&token.to_lowercase())
+    }
+
+    /// Add stopwords for a specific language
+    fn add_language_stopwords(stopwords: &mut HashSet<String>, language: &str) {
+        // Add common English stopwords
+        if language.to_lowercase() == "english" {
```
generall: please use enum instead of str

lib/segment/src/data_types/index.rs
```diff
@@ -194,6 +194,25 @@ pub struct TextIndexParams {
...
+#[serde(untagged)]
+pub enum StopwordsOption {
+    Language(String),
+    Set(StopwordsSet),
+}
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+pub struct StopwordsSet {
+    #[serde(default)]
+    pub languages: Vec<String>,
```
generall: Use enum here, it will be a free validation

lib/segment/src/data_types/index.rs
```diff
@@ -194,6 +194,25 @@ pub struct TextIndexParams {
...
     pub on_disk: Option<bool>,
+
+    #[serde(default)]
+    pub stopwords: Option<StopwordsOption>,
+}
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+#[serde(untagged)]
+pub enum StopwordsOption {
+    Language(String),
```
generall: same, use enum

lib/segment/src/data_types/index.rs
```diff
@@ -194,6 +194,25 @@ pub struct TextIndexParams {
     /// If true, store the index on disk. Default: false.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub on_disk: Option<bool>,
+
+    #[serde(default)]
+    pub stopwords: Option<StopwordsOption>,
```
generall: If None, should skip serialization for backward compatibility

lib/segment/src/index/field_index/full_text_index/stop_words_container.rs
```diff
@@ -0,0 +1,9744 @@
+// Arabic stopwords
+pub(crate) const ARABIC_STOPWORDS: &[&str] = &[
```
coszio: An alternative would be to have a folder with files for each language, and use [`include_str`](https://doc.rust-lang.org/std/macro.include_str.html) or [`include`](https://doc.rust-lang.org/std/macro.include.html) macros, just so we have the languages more organized

lib/segment/src/index/field_index/full_text_index/stop_words.rs
```diff
@@ -0,0 +1,206 @@
...
+
+use super::stop_words_container::{
+    ARABIC_STOPWORDS, AZERBAIJANI_STOPWORDS, BASQUE_STOPWORDS, BENGALI_STOPWORDS,
+    CATALAN_STOPWORDS, CHINESE_STOPWORDS, DANISH_STOPWORDS, DUTCH_STOPWORDS, ENGLISH_STOPWORDS,
+    FINNISH_STOPWORDS, FRENCH_STOPWORDS, GERMAN_STOPWORDS, GREEK_STOPWORDS, HEBREW_STOPWORDS,
+    HINGLISH_STOPWORDS, HUNGARIAN_STOPWORDS, INDONESIAN_STOPWORDS, ITALIAN_STOPWORDS,
+    KAZAKH_STOPWORDS, NEPALI_STOPWORDS, NORWEGIAN_STOPWORDS, PORTUGUESE_STOPWORDS,
+    ROMANIAN_STOPWORDS, RUSSIAN_STOPWORDS, SLOVENE_STOPWORDS, SPANISH_STOPWORDS, SWEDISH_STOPWORDS,
+    TAJIK_STOPWORDS, TURKISH_STOPWORDS,
+};
```
coszio: nit: could be a good place to use * imports
```suggestion
use super::stop_words_container::*;
```

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +198,59 @@ pub struct TextIndexParams {
...
+#[derive(Debug, Serialize, Deserialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+#[serde(untagged)]
+pub enum StopwordsOption {
+    Language(Language),
+    Set(StopwordsSet),
+}
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+#[serde(rename_all = "lowercase")]
+pub enum Language {
```
coszio, line 244: To discuss: should we add their ISO abbreviations as aliases?

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +198,59 @@ pub struct TextIndexParams {
...
     #[serde(skip)]
     pub phrase_matching: Option<bool>,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub stopwords: Option<StopwordsOption>,
+}
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+#[serde(untagged)]
+pub enum StopwordsOption {
```
coszio: nit: There are other places where we make a simplified and parametrized options like in `MatchInterface`, `QueryInterface`, `RangeInterface`, `OrderByInterface`. So, for consistency, we could also do this an `Interface`.
```suggestion
pub enum StopwordsInterface {
```

lib/segment/src/index/field_index/full_text_index/stop_words_container.rs
n0x29a: we agreed with @generall that const is fine. should we change to include_str?

lib/segment/src/index/field_index/full_text_index/stop_words.rs
n0x29a: depends... now it explicitly shows all available languages. maybe rename <lang>_stopwords to <lang> + using ISO will allow us to see all used languages explicitly?

lib/segment/src/index/field_index/full_text_index/stop_words_container.rs
generall: I am fine with const, syntax overhead is minimal and code simpler

generall:
gRPC interface is also required

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +198,136 @@ pub struct TextIndexParams {
...
+    Spanish,
+    #[serde(alias = "sv")]
+    Swedish,
+    #[serde(alias = "tg")]
+    Tajik,
+    #[serde(alias = "tr")]
+    Turkish,
+}
+
+impl<'de> serde::Deserialize<'de> for Language {
```
n0x29a: to have more flexible error message

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +198,136 @@ pub struct TextIndexParams {
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
```
generall: This custom deserialization doesn't really help with better error message.
It is getting eaten by Enum ambiguity. If you want nicer error message, you need to change `PayloadSchemaParams`

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +198,136 @@ pub struct TextIndexParams {
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        let lowercase_s = s.to_lowercase();
+
+        match lowercase_s.as_str() {
+            "unspecified" | "unspecifiedlanguage" => Ok(Language::UnspecifiedLanguage),
```
generall: What's the expected behaviour of `unspecified`? What's the difference with None?

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +198,136 @@ pub struct TextIndexParams {
...
+                "Unsupported language: {s}. Please use one of the supported languages.",
+            ))),
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+pub struct StopwordsSet {
+    #[serde(default)]
+    pub languages: Vec<Language>,
```
generall: I would make it HashSet (or maybe BTreeSet)

Also, I would allow singular values.

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +198,136 @@ pub struct TextIndexParams {
...
+
+    #[serde(default)]
+    pub custom: Vec<String>,
```
generall: BTreeSet here as well

lib/segment/src/data_types/index.rs
n0x29a: it's following the [proto best practices](https://protobuf.dev/best-practices/dos-donts/) which tells us that the first enum parameter should be defined as `unspecified`.

the common reason is following (not necessary our use case):
 - Prevents gRPC zero-value bugs: Ensures clients that default an enum field to 0 get a safe no-op instead of an arbitrary language.
 - Explicit override: Allows unambiguous disabling of stopwords in layered configurations, where an omitted field (None) would mean "inherit from parent scope".
 - Clear PATCH semantics: Provides a canonical way to reset/nullify the field, which is distinct from omitting it (no-op).

lib/api/src/grpc/proto/collections.proto
```diff
@@ -463,12 +463,58 @@ message GeoIndexParams {
   optional bool on_disk = 1; // If true - store index on disk.
 }
 
+enum Language {
```
generall: I am not sure about using this big enum in gRPC, I would like to ask for a second opinion.

Keep in mind, that gRPC generates global constants in clients for enums like that.

lib/api/src/grpc/proto/collections.proto
```diff
@@ -463,12 +463,58 @@ message GeoIndexParams {
...
+  repeated string custom = 2; // List of custom stopwords
+}
+
+message StopwordsInterface {
+  oneof stopwords {
+    Language language = 1; // Single language stopwords
+    StopwordsSet set = 2; // Multiple languages and custom stopwords
+  }
+}
+
```
generall: `StopwordsInterface` is nice for REST, but too complex for gRPC. It will significantly increase complexity of client-side code and won't provide any meaningful UX improvements. I propose to use `StopwordsSet` directly and implement proper `Builders` in clients, so we can construct this object from just  e.g. "English"

lib/segment/src/index/field_index/full_text_index/stop_words/catalan.rs
```diff
@@ -0,0 +1,280 @@
...
+    "som",
+    "son",
+    "sons",
+    "sota",
+    "sou",
+    "sóc",
+    "són",
+    "t'ha",
+    "t'han",
+    "t'he",
```
timvisee, line 235: How are these tokenized? If these become multiple tokens, do we filter them out later, or, what do we do with them?

lib/segment/src/index/field_index/full_text_index/stop_words/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+            Language::Russian => RUSSIAN_STOPWORDS,
+            Language::Slovene => SLOVENE_STOPWORDS,
+            Language::Spanish => SPANISH_STOPWORDS,
+            Language::Swedish => SWEDISH_STOPWORDS,
+            Language::Tajik => TAJIK_STOPWORDS,
+            Language::Turkish => TURKISH_STOPWORDS,
+        };
+
+        for &word in stopwords_array {
+            stopwords.insert(word.to_lowercase());
```
timvisee: Just lower casing strings is not safe to normalize languages!

Ideally we use case-insensitive comparisons here. But I understand that's not easily doable with a hashset.

Some examples:
- `"ß"` becomes `"SS"` when upper cased
- In Turkish: upper case `"İ"` is lower case `"i"`
- In Turkish: upper case `"I"` is lower case `"ı"`
- In non-Turkish: upper case `"I"` is lower case `"i"`

Generally speaking, upper casing is seen as the better approach. But you still cannot consider that proper normalization.

I see that our German list contains both variants:
https://github.com/qdrant/qdrant/blob/73554e53f5ed7e295c780614416a1cd81c5b1823/lib/segment/src/index/field_index/full_text_index/stop_words/german.rs#L38-L39

But Turkish does not:
https://github.com/qdrant/qdrant/blob/73554e53f5ed7e295c780614416a1cd81c5b1823/lib/segment/src/index/field_index/full_text_index/stop_words/turkish.rs#L4

Point of discussion: do we want proper case-insensitive comparisons here?

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +200,139 @@ pub struct TextIndexParams {
...
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+pub struct StopwordsSet {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub language: Option<Language>,
+
+    #[serde(default, skip_serializing_if = "BTreeSet::is_empty")]
+    pub languages: BTreeSet<Language>,
```
timvisee: Why do we have both `language` and `languages`? I think we should stick with one.

If this is about supporting both 'language' and 'languages' keys, we can do:

```rust
#[serde(default, alias = "languages", skip_serializing_if = "BTreeSet::is_empty")]
pub language: BTreeSet<Language>,
```

lib/segment/src/data_types/index.rs
timvisee: I don't agree with this approach. This feels like a gRPC implementation detail that bleeds throughout the rest of our code base.

Let's only keep it in the gRPC API as suggested by the best practice.

In the rest of our code base we can remove that variant, and have only supported languages. Our gRPC front-end can simply disallow the unspecified variant when it gets converted into our internal structures.

If you talk about patch/unset semantics, the user could provide `language: []`.

lib/segment/src/data_types/index.rs
n0x29a: we have both to have some clarity about single and multiple values. customer could use languages/language without being confused using `languages: "english"`

timvisee: My above suggestion is how that is normally implemented, not with two explicit fields of different types.

lib/segment/src/index/field_index/full_text_index/stop_words/catalan.rs
n0x29a: stop words should be filtered out. [here is a test](https://github.com/qdrant/qdrant/blob/stop-words/lib/segment/src/index/field_index/full_text_index/tokenizers.rs#L338-L339) for "I'd", "t'ha" but we can add more words like that

lib/segment/src/index/field_index/full_text_index/stop_words/mod.rs
n0x29a: we could use [unicode-normalization](https://crates.io/crates/unicode-normalization), for example. But we didn't touched that point.

timvisee: Conclusion in call, let's keep it simple for now:

- don't normalize
- don't lowercase stopword list
- only lower case user query filter if `"lowercase": true` in the index definition

docs/redoc/master/openapi.json
```diff
@@ -7639,6 +7639,16 @@
...
+          "stopwords": {
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/StopwordsInterface"
+              },
+              {
+                "nullable": true
+              }
+            ]
           }
```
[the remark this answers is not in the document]
coszio: This is auto-generated

docs/redoc/master/openapi.json
```diff
@@ -7657,6 +7667,71 @@
...
+          },
+          {
+            "$ref": "#/components/schemas/StopwordsSet"
+          }
+        ]
+      },
+      "Language": {
+        "type": "string",
+        "enum": [
+          "unspecifiedlanguage",
```
coszio: I don't think `"unspecifiedlanguage"` should be a part of the enum in openapi, is there a way to hide it here?

lib/api/src/grpc/conversions.rs
```diff
@@ -214,15 +215,38 @@ impl From<segment::data_types::index::TextIndexParams> for PayloadIndexParams {
...
+            }
+            segment::data_types::index::StopwordsInterface::Set(set) => {
+                let languages = set.languages.iter().map(|lang| lang.to_string()).collect();
+
+                StopwordsSet {
+                    languages,
+                    custom: set.custom.into_iter().collect(),
+                }
+            }
+        });
```
coszio: If `StopwordsSet` from segment and grpc are equivalent, I think there should be these From implementations:
- grpc StopwordsSet <-> segment StopwordsSet
- segment StopwordsSet <-> segment StopwordsInterface

lib/api/src/grpc/conversions.rs
```diff
@@ -410,7 +434,45 @@ impl TryFrom<TextIndexParams> for segment::data_types::index::TextIndexParams {
             min_token_len,
             max_token_len,
             on_disk,
+            stopwords,
         } = params;
+
+        // Convert stopwords if present
+        let stopwords_converted = if let Some(set) = stopwords {
```
coszio, line 487: Same here

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +202,177 @@ pub struct TextIndexParams {
...
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+#[serde(untagged)]
+pub enum StopwordsInterface {
+    Language(Language),
+    Set(StopwordsSet),
+}
+
+#[derive(Debug, Serialize, JsonSchema, Clone, PartialEq, PartialOrd, Ord, Hash, Eq)]
+#[serde(rename_all = "lowercase")]
```
coszio:
```suggestion
#[serde(rename_all = "snake_case")]
```

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +202,177 @@ pub struct TextIndexParams {
...
+pub enum Language {
+    #[serde(alias = "unspecified")]
```
coszio: There is `#[serde(other)]` if you want it to match with anything else

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +202,177 @@ pub struct TextIndexParams {
...
+    UnspecifiedLanguage,
+    #[serde(alias = "ar")]
```
coszio, line 245: These aliases are not doing anything if there is a custom `Deserialize` implementation

lib/api/src/grpc/conversions.rs
n0x29a: in this case I would prefer to keep it as is because tis conversion is fallible

.github/workflows/codespell.yml
```diff
@@ -18,5 +18,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
       - name: Codespell
         uses: codespell-project/actions-codespell@v2
+        with:
+          skip: '**/stop_words/*.rs'
```
coszio, line 25: There is a `.codespellrc` file on the project root, can we add this there?

lib/api/src/grpc/conversions.rs
coszio: There is `TryFrom` for fallible conversions, just in case

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +202,129 @@ pub struct TextIndexParams {
...
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+#[serde(untagged)]
+pub enum StopwordsInterface {
+    Language(Language),
+    Set(StopwordsSet),
+}
+
+impl StopwordsInterface {
+    #[cfg(feature = "testing")]
```
JojiiOfficial, line 218: What exactly is the feature "testing" for? What advantage gives us on top of `#[cfg(test)]`?

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +202,129 @@ pub struct TextIndexParams {
...
+    pub fn new_custom(custom: &[&str]) -> Self {
+        StopwordsInterface::Set(StopwordsSet {
+            languages: None,
+            custom: Some(custom.iter().map(|s| (*s).to_string()).collect()),
```
JojiiOfficial, line 222: I don't think we need to dereference the `&&str` here, the compiler should do that for us.
```suggestion
            custom: Some(custom.iter().map(|s| s.to_string()).collect()),
```

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +202,129 @@ pub struct TextIndexParams {
...
+    #[cfg(feature = "testing")]
+    pub fn new_language(language: Language) -> Self {
+        StopwordsInterface::Language(language)
+    }
+
+    #[cfg(feature = "testing")]
+    pub fn new_set(languages: &[Language], custom: &[&str]) -> Self {
+        StopwordsInterface::Set(StopwordsSet {
+            languages: Some(languages.iter().cloned().collect()),
+            custom: Some(custom.iter().map(|s| (*s).to_string()).collect()),
```
JojiiOfficial, line 235:
```suggestion
            custom: Some(custom.iter().map(|s| s.to_string()).collect()),
```

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +202,129 @@ pub struct TextIndexParams {
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let json_string = serde_json::to_string(self).map_err(|_| fmt::Error)?;
+        f.write_str(json_string.trim_matches('"'))
+    }
+}
+
+impl FromStr for Language {
+    type Err = serde_json::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
JojiiOfficial, line 314: Not necessarily something for this PR, but maybe we can give [strum](https://github.com/Peternator7/strum) a try. This string conversion seems quite hacky and allocates on the heap, which we could easily avoid.

lib/segment/src/data_types/index.rs
```diff
@@ -198,6 +202,129 @@ pub struct TextIndexParams {
...
+    Swedish,
+    #[serde(alias = "tg")]
+    Tajik,
+    #[serde(alias = "tr")]
+    Turkish,
+}
+impl fmt::Display for Language {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let json_string = serde_json::to_string(self).map_err(|_| fmt::Error)?;
+        f.write_str(json_string.trim_matches('"'))
```
JojiiOfficial, line 307: Maybe it is a bit cleaner this way?
```suggestion
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "{self:?}")
```

lib/segment/src/index/field_index/full_text_index/stop_words/mod.rs
```diff
@@ -0,0 +1,299 @@
...
+pub use russian::RUSSIAN_STOPWORDS;
+pub use slovene::SLOVENE_STOPWORDS;
+pub use spanish::SPANISH_STOPWORDS;
+pub use swedish::SWEDISH_STOPWORDS;
+pub use tajik::TAJIK_STOPWORDS;
+pub use turkish::TURKISH_STOPWORDS;
+
+#[derive(Debug, Clone, Default)]
+pub struct StopwordsFilter {
+    stopwords: HashSet<String>,
```
JojiiOfficial: Since we refactored it recently, maybe it would be more consistent and slightly more performant with using Ahash here?

```suggestion
    stopwords: AHashSet<String>,
```

lib/segment/src/index/field_index/full_text_index/stop_words/mod.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+impl StopwordsFilter {
+    pub fn new(option: &Option<StopwordsInterface>, lowercase: bool) -> Self {
+        let mut stopwords = HashSet::new();
```
JojiiOfficial: Some minor optimization suggestion, but since we know the size, we could preallocate here.
This especially helps if we create `StopwordsFilter` a lot.

JojiiOfficial (approved):

lib/segment/src/data_types/index.rs
coszio: It was introduced to use across crates

JojiiOfficial: Alright, as long as this has been chosen over `#[cfg(test]` on purpose, I'm fine with it!

lib/segment/src/index/field_index/full_text_index/stop_words/mod.rs
```diff
@@ -0,0 +1,299 @@
...
+pub use tajik::TAJIK_STOPWORDS;
+pub use turkish::TURKISH_STOPWORDS;
+
+#[derive(Debug, Clone, Default)]
+pub struct StopwordsFilter {
+    stopwords: HashSet<String>,
+}
+
+impl StopwordsFilter {
+    pub fn new(option: &Option<StopwordsInterface>, lowercase: bool) -> Self {
```
JojiiOfficial, line 71: What exactly is semantic of the `lowercase` parameter?
Should it only import treat the list as lowercase or should it enforce lowercase during checking as well?

The test suggest that the second is the case but the code only implements the first one.

lib/segment/src/data_types/index.rs
generall: clippy complained about it

lib/segment/src/data_types/index.rs
generall: clippy complained about it

lib/segment/src/data_types/index.rs
generall: I did look into strum, but decided against it. Introducing a dependency for an operation which only happens once per index creation seems like an overkill. serde_json guarantees that we will have exactly same representation of strings as in API, though it does look a bit strange in code.

lib/segment/src/data_types/index.rs
generall: no, this way we will loose `#[serde(rename_all = "snake_case")]` and string representations won't be compatible anymore

lib/segment/src/index/field_index/full_text_index/stop_words/mod.rs
generall: :heavy_check_mark:

lib/segment/src/index/field_index/full_text_index/stop_words/mod.rs
generall: > Should it only import treat the list as lowercase or should it enforce lowercase during checking as well?

It should only enforce lowercase during creation. For lowercasing during indexing/querying it is a responsibility of tokenizer, otherwise we would need to create copy of each token twice.

lib/segment/src/index/field_index/full_text_index/stop_words/mod.rs
generall: we don't actually know size before iterating over StopwordsInterface and de-duplicating stopwords across all languages. We theoretically can estimate it, but it would make code 2x more complex and probably won't be measurable on benchmarks.

lib/segment/src/index/field_index/full_text_index/text_index.rs
```diff
@@ -603,3 +638,470 @@ impl FieldIndexBuilderTrait for FullTextGridstoreIndexBuilder {
...
+        match &mut builder {
+            #[cfg(feature = "rocksdb")]
+            IndexBuilder::Mutable(builder) => builder.init().unwrap(),
+            IndexBuilder::MutableGridstore(builder) => builder.init().unwrap(),
+            #[cfg(feature = "rocksdb")]
+            IndexBuilder::Immutable(builder) => builder.init().unwrap(),
+            IndexBuilder::Mmap(builder) => builder.init().unwrap(),
+            IndexBuilder::RamMmap(builder) => builder.init().unwrap(),
+        }
+        CreatedBuilder {
```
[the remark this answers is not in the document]
generall: It is only called from inside the test module

generall (approved):

lib/segment/src/index/field_index/full_text_index/stop_words/catalan.rs
generall: This test is actually not testing what is intended. Instead of filtering out stopwords, it just splits them differently.

generall: But the stopword lists are compiled the way, that they contain both: full form like `you'll` as well as its components `you` and `ll`

docs/redoc/master/openapi.json
```diff
@@ -7657,6 +7667,71 @@
...
+          "norwegian",
+          "portuguese",
+          "romanian",
+          "russian",
+          "slovene",
+          "spanish",
+          "swedish",
+          "tajik",
+          "turkish"
+        ]
```
timvisee, line 7712: Should aliases be in here?

lib/segment/src/data_types/index.rs
timvisee: Note that we already have `strum` in our dependency tree.

I agree that this is a bit hacky, but I like that it ensures (de)serialization is compatible.

lib/segment/src/index/field_index/full_text_index/tokenizers.rs
```diff
@@ -1,38 +1,97 @@
...
+        text: &str,
+        lowercase: bool,
+        stopwords_filter: &StopwordsFilter,
+        mut callback: C,
+    ) {
+        for token in text.split_whitespace() {
+            if token.is_empty() {
+                continue;
+            }
+
```
timvisee, line 21: According to the docs, this already eliminates empty results.

https://doc.rust-lang.org/std/primitive.str.html#method.split_whitespace

Example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=89b6012aff7a45372cf8e3dfd7f76c62

```suggestion
```

Or if we want to be safe:

```suggestion
            debug_assert!(!token.is_empty());
```

lib/segment/src/index/field_index/full_text_index/immutable_text_index.rs
```diff
@@ -13,12 +13,16 @@ use crate::common::Flusher;
...
 use crate::common::rocksdb_buffered_delete_wrapper::DatabaseColumnScheduledDeleteWrapper;
+#[cfg(feature = "rocksdb")]
 use crate::data_types::index::TextIndexParams;
 use crate::index::field_index::full_text_index::inverted_index::mmap_inverted_index::mmap_postings_enum::MmapPostingsEnum;
+use crate::index::field_index::full_text_index::tokenizers::Tokenizer;
 
 pub struct ImmutableFullTextIndex {
     pub(super) inverted_index: ImmutableInvertedIndex,
+    #[cfg(feature = "rocksdb")]
     pub(super) config: TextIndexParams,
```
timvisee, line 24: I am confused - why would we only use the config with RocksDB enabled? Does it have a strict dependency on it, while our other index type does not?

(more of a note to self, so that I'll check this in a bit :smiley:)

timvisee (approved):

docs/redoc/master/openapi.json
timvisee: Turns out not to be supported yet:

- https://github.com/GREsau/schemars/issues/25
- https://github.com/GREsau/schemars/issues/338

PR #6636, opened 2025-06-04
Remove option from immutable_inverted_index

JojiiOfficial:
Converts `Vec<Option<usize>>` to `Vec<usize>`, using `0` as replacement for `None` to make immutable_inverted_index consistent with mmap_inverted_index


timvisee:
Thanks for picking this up!

PR #6638, opened 2025-06-04
[phrase match] Restructure `full_text_index::inverted_index` module

coszio:
Builds on top of #6493 

Since more files and logic is being added to full-text-index, navigability was getting more complex. This PR aims to simplify it by:
- moving all inverted_index specific files to one module
- moving a chunky congruence test into its own file
- removing `old_mmap_postings` and `compressed_postings` modules


timvisee (approved):
Thanks!

Am I correct that there is no logic change in here. At least I didn't see any.

If not, let's merge! If there is, please let me know and I'll revisit that part again.

coszio:
There's no logic changes here, correct

PR #6640, opened 2025-06-04
Fix enormous memory usage in Distance Matrix API on high sample size

JojiiOfficial:
Fixes extremely high memory usage in Distance Matrix API on certain datasets with a large sample rate (>= ~50k).

Dev:
image

PR:
image

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -2625,7 +2625,7 @@ impl From<VectorNameBuf> for HasVectorCondition {
 }
 
 /// Threshold determining when to use an `Arc` in `HasIdCondition` if the condition includes many points.
-/// Since we're cloning filters quite alot, using an Arc for larger conditions reduces risk of memory leaks
+/// Since we're cloning filters quite a lot, using an Arc for larger conditions reduces risk of memory leaks
 /// and potentially improves performance in some places.
 const HAS_ID_CONDITION_ARC_THRESHOLD: usize = 1_000;
 
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -38,6 +38,7 @@ use crate::index::sparse_index::sparse_index_config::SparseIndexConfig;
 use crate::json_path::JsonPath;
 use crate::spaces::metric::MetricPostProcessing;
 use crate::spaces::simple::{CosineMetric, DotProductMetric, EuclidMetric, ManhattanMetric};
+use crate::utils::maybe_arc::MaybeArc;
 
 pub type PayloadKeyType = JsonPath;
 pub type PayloadKeyTypeRef<'a> = &'a JsonPath;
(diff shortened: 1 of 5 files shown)


KShivendu (commented):
Nice! Can you please share script to repro the bug/results? :)

lib/segment/src/types.rs
```diff
@@ -2608,7 +2609,7 @@ impl From<JsonPath> for IsEmptyCondition {
 #[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, PartialEq, Eq)]
 pub struct HasIdCondition {
     #[schemars(schema_with = "HashSet::<PointIdType>::json_schema")]
-    pub has_id: AHashSet<PointIdType>,
```
agourlay, line 2611: Why not always have an `Arc`?

timvisee: I think the answer lies here:

```rust
/// Threshold determining when to use an `Arc` in `HasIdCondition` if the condition includes many points.
/// Since we're cloning filters quite a lot, using an Arc for larger conditions reduces risk of memory leaks
/// and potentially improves performance in some places.
const HAS_ID_CONDITION_ARC_THRESHOLD: usize = 1_000;
```

meaning, a balance between performance and memory usage.

Though, I'm also curious how always using an Arc behaves. @JojiiOfficial  Would it be possible to benchmark it?

JojiiOfficial: Here is a comparison between cloning raw+arc and creating an arc using `Arc::new()` of the same hashset used in each iteration.
image

I suggest that we set the `HAS_ID_CONDITION_ARC_THRESHOLD` to something around `60` (probably 64 due to power of 2 allocation logic of rust) to not introduce overhead in places where we might create a lot of small filters without much cloning. Otherwise we might introduce an unexpected performance regression.

Wdyt?

coszio (approved):

lib/segment/src/utils/maybe_arc.rs
```diff
@@ -0,0 +1,111 @@
...
+// Structure that acts as `T` most of the time but allows to interchange being wrapped within an `Arc` or not.
+// This is helpful, when a variable can become memory-intensive but must remain the ability to get cloned.
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, PartialEq, Eq)]
+#[serde(untagged)] // Make this type transparent when de/serializing and always deserialize as `NoArc`, since it's the first enum kind that matches.
+pub enum MaybeArc<T> {
+    NoArc(T),
+    Arc(Arc<T>),
+}
+
+impl<T> MaybeArc<T> {
```
coszio, line 16: A nice to have would be to choose arc/no_arc automatically based on the `size_of_value()` of the inner value as a constructor function of this type.

Based on the provided chart, and considering `ExtendedPointId` is 24 bytes, the threshold could be around 1536 bytes (24 * 64)

JojiiOfficial: `size_of_value` doesn't work with HashSet though. When using `MaybeArc` wrapped around HashMaps or Vec, `size_of_value` would only return the stack size of the type, which is always constant regardless of the amount of items.
Therefore we can only apply this heuristic at each usage of `MaybeArc` instead inside a constructor, sadly.

PR #6651, opened 2025-06-06
Experiment: use m instead of m0 for initial links

xzfc:
Might improve the accuracy on some datasets.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
@@ -532,6 +532,7 @@ impl GraphLayersBuilder {
                 }
             }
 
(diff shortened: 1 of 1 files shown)


xzfc:
It improves the accuracy for the laion dataset, but makes it worse for everything else (dbpedia, deep-image, gist, glove-100).

The reason it was better on laion:
When building the HNSW index by sequentially adding points, points added early get more backlinks than points added later. Thus, the heuristic is triggered less frequently for later points.
The laion dataset (at least with the parameters I tested) got really unlucky: many of the later points had the heuristic triggered only once (which in some cases reduces the number of their links to only one) and never got repopulated by backlinks again. Many points have only one link => worse accuracy.

PR #6653, opened 2025-06-09
Buffer when writing into mmap dense vector storage

timvisee:
Minor change, buffer writes when putting vectors on disk.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
+++ b/lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
@@ -239,12 +239,10 @@ impl<T: PrimitiveVectorElement> VectorStorage for MemmapDenseVectorStorage<T> {
                 deleted_ids.push(start_index as PointOffsetType + offset as PointOffsetType);
             }
         }
-        vectors_file.flush()?;
         vectors_file
             .into_inner()
-            // unwrap safety: never fails, we flushed above
-            .unwrap()
-            .sync_all()?;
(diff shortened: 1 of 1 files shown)


lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
```diff
@@ -239,8 +239,12 @@ impl<T: PrimitiveVectorElement> VectorStorage for MemmapDenseVectorStorage<T> {
...
             }
         }
-        vectors_file.sync_all()?;
-        drop(vectors_file);
+        vectors_file.flush()?;
+        vectors_file
+            .into_inner()
+            // unwrap safety: never fails, we flushed above
+            .unwrap()
+            .sync_all()?;
```
ffuugoo: Why not just `into_inner`? It's _kinda_ a bit less explicit than `flush`, but also less ugly? Also, do we need `sync_all` or just `sync_data`?

```suggestion
        vectors_file
            .into_inner()
            .map_err(|err| err.into_error())?
            .sync_data()?;
```

timvisee: I agree that `sync_data` is fine here. I don't really mind about the other one.

I've changed both. It makes the error a bit ugly, but it removes the unwrap which is nice.

```rust
        vectors_file
            .into_inner()
            .map_err(io::IntoInnerError::into_error)?
            .sync_data()?;
```

PR #6654, opened 2025-06-09
[phrase matching] Handle phrase filtering

coszio:
Builds on top of #6638 

Adds the `ParsedQuery::Phrase` variant, and handles it for filter, match, and estimate cardinality


lib/segment/src/index/field_index/full_text_index/mutable_text_index.rs
```diff
@@ -38,6 +38,26 @@ pub(super) enum Storage {
...
+                })
+                .unwrap()
+            }
+            Storage::Gridstore(gridstore) => gridstore
+                .read()
+                .get_value(idx, &HardwareCounterCell::disposable())
+                .map(|bytes| FullTextIndex::deserialize_document(&bytes).unwrap()),
+        }
+    }
+}
```
timvisee: I vote to move this into `MutableFullTextIndex` instead, where all the other functions are placed too.

Wdyt?

lib/segment/src/index/field_index/full_text_index/inverted_index/postings_iterator.rs
```diff
@@ -57,6 +61,114 @@ pub fn intersect_compressed_postings_iterator<'a, V: PostingValue + 'a>(
...
+
+    let Some(mut postings) = postings_opt else {
+        // There are unseen tokens -> no matches
+        return Box::new(std::iter::empty());
+    };
+
+    if postings.is_empty() {
+        // Empty request -> no matches
+        return Box::new(std::iter::empty());
+    }
```
timvisee: Can we move this one up top?

If `phrase` is empty, return.

lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
```diff
@@ -47,7 +47,13 @@ impl TokenSet {
         self.0.binary_search(token).is_ok()
     }
 
+    /// Checks if the current set contains all given tokens.
+    ///
+    /// Returns false if the subset is empty
     pub fn has_subset(&self, subset: &TokenSet) -> bool {
+        if subset.is_empty() {
+            return false;
+        }
```
timvisee, line 56: Is this required? It seems counter intuitive to me, and it doesn't look like it's necessary for the callers.

lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
```diff
@@ -86,6 +92,33 @@ impl Document {
...
+    pub fn has_phrase(&self, phrase: &Document) -> bool {
+        let doc = self.0.as_slice();
+        let phrase = phrase.0.as_slice();
+
+        if doc.is_empty() {
+            return false;
+        }
+
+        // simple check for tokens in the same order as phrase
+        doc.windows(phrase.len()).any(|window| window == phrase)
```
timvisee, line 120: We can potentially benefit from something like [KMP](https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm) here / search from the back and jump the whole length forward on mismatch. I understand it's tokens here, but the same applies.

We could even implement it in a generic way because I believe we use it in more places.

Just noting it here as potential side quest :nerd_face: 

Edit: as far as I understand this can be quite hot, so it might make sense anyway.

timvisee, line 120: `phrase.len()` must not be 0, otherwise we panic!

lib/segment/src/index/field_index/full_text_index/inverted_index/positions.rs
```diff
@@ -33,3 +50,57 @@ impl UnsizedValue for Positions {
...
+    pub fn new(mut tokens_positions: Vec<TokenPosition>) -> Self {
+        tokens_positions.sort_by_key(|tok_pos| tok_pos.position);
+        Self(tokens_positions)
+    }
+
+    /// Returns true if any sequential window of tokens match the given phrase.
+    pub fn has_phrase(&self, phrase: &Document) -> bool {
+        match phrase.tokens() {
+            // no tokens in query -> no match
+            [] => false,
```
timvisee, line 74: Also mentioned above, but is this desired?

lib/segment/src/index/field_index/full_text_index/tests/test_congruence.rs
```diff
@@ -379,6 +388,49 @@ fn test_congruence(
             );
         }
 
+        if phrase_matching {
```
timvisee, line 391: Can we also explicitly check phrases with duplicate words here? Maybe we can artificially add them in the test case.

If already handled, great! Then I didn't find it so quickly.

timvisee (approved):
:+1: 

Could you also take a look at the test failures?

lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
coszio: It is a potential side-quest, yes 😄 . 

AFAIK KMP works best when the needle and the haystack are long, which may not necessarily be the case when it is already-tokenized. Moreover, it adds overhead that is not justified for small sequences.

lib/segment/src/index/field_index/full_text_index/inverted_index/positions.rs
coszio: It is the current behavior for `text` queries, which I don't think we should break for `phrase`

lib/segment/src/index/field_index/full_text_index/tests/test_congruence.rs
coszio: It's not checked yet, good idea. But I'll check that somewhere else.

Edit, this somewhere else is [next PR](https://github.com/qdrant/qdrant/pull/6670)

lib/segment/src/index/field_index/full_text_index/tests/test_congruence.rs
```diff
@@ -379,6 +388,49 @@ fn test_congruence(
...
+                assert_eq!(
+                    index_a
+                        .filter_query(parsed_query_a, &hw_counter)
+                        .collect::<HashSet<_>>(),
+                    index_b
+                        .filter_query(parsed_query_b, &hw_counter)
+                        .collect::<HashSet<_>>(),
+                );
+            }
+        }
```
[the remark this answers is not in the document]
coszio: Tested in next PR

lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
coszio: Same as https://github.com/qdrant/qdrant/pull/6654#discussion_r2145441324

PR #6670, opened 2025-06-10
[phrase matching] expose phrase condition

coszio:
- Adds a `Match::Phrase` variant to specify that a text must match by phrase.
- Adds a `phrase_matching` flag to the text index params so that text index registers positions.

Exposes it on rest and grpc

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/text_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/text_index.rs
@@ -32,7 +32,7 @@ use crate::index::field_index::{
     ValueIndexer,
 };
 use crate::telemetry::PayloadIndexTelemetry;
-use crate::types::{FieldCondition, Match, MatchText, PayloadKeyType};
+use crate::types::{FieldCondition, Match, MatchPhrase, MatchText, PayloadKeyType};
 
 pub enum FullTextIndex {
     Mutable(MutableFullTextIndex),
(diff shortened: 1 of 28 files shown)


lib/segment/src/index/field_index/full_text_index/text_index.rs
```diff
@@ -312,13 +312,26 @@ impl FullTextIndex {
         }
     }
 
+    /// Tries to parse a phrase query. If there are any unseen tokens, returns `None`
```
timvisee, line 281: > If there are any unseen tokens, returns `None`

Is this correct?

tests/openapi/test_phrase_matching.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+@pytest.fixture(scope="module")
+def phrase_collection():
+    collection_name = setup_phrase_collection()
+    yield collection_name
+    drop_collection(collection_name)
+
+
+@pytest.mark.parametrize(
```
timvisee, line 95: Can we also add a case here that tests a duplicate word, like `"the the"`

timvisee (approved):
Please see all review remarks

lib/segment/src/index/field_index/full_text_index/text_index.rs
coszio: Yes, if a token hasn't been indexed, it means the query will return 0 results in this segment

tests/openapi/test_phrase_matching.py
coszio: Added in https://github.com/qdrant/qdrant/pull/6670/commits/2bfb949be1ac47db533f3986b7d78e9c2d69e20b

timvisee: very very very very :clown_face:

PR #6675, opened 2025-06-11
Run unit tests in CI without RocksDB flag

timvisee:
Add a temporary CI job to also run all our unit tests without the `rocksdb` feature flag.

I'd like to test the RocksDB-less code too, as we'll be mainlining it in the future.

I didn't just replace the existing tests, because we'll still be releasing a few versions that include RocksDB. Therefore we must still test our RocksDB code until we actually remove it.

Once we disable RocksDB by default, we can remove this extra test job.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -110,3 +110,42 @@ jobs:
         with:
           filename: .github/ISSUE_TEMPLATE/flaky_test.md
           update_existing: true
+
+  # Temporary job that runs tests without the RocksDB feature flag
+  # We will remove this once RocksDB will be disabled by default
+  rust-tests-no-rocksdb:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ ubuntu-latest ]
+
+    steps:
+    - name: Install minimal stable
+      uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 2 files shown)


ffuugoo (approved):

.github/workflows/rust.yml
```diff
@@ -110,3 +110,44 @@ jobs:
...
+    - name: Build
+      run: cargo build --tests --workspace --locked --no-default-features
+    - name: Run tests
+      # Profile "ci" is configured in .config/nextest.toml
+      run: cargo nextest run --workspace --profile ci --locked --no-default-features
+    - name: Upload test report
+      uses: actions/upload-artifact@v4
+      with:
+        name: junit-${{ matrix.os }}-no-rocksdb.xml
+        path: target/nextest/ci/junit.xml
```
coszio: This is used by `process_results` job, if we don't want to update the job to account for this extra test, then it should be fine to remove this step

timvisee: My thinking was that it may still be useful when manually downloading it from [assets](https://github.com/qdrant/qdrant/actions/runs/15588016966).

But on second thought, we'll probably never do that.

I'll remove it.

PR #6677, opened 2025-06-12
Partial snapshot is recovering telemetry

KShivendu:
Add new telemetry field `is_recovering` to avoid triggering re-triggering partial snapshots on the same read replica while one is already ongoing. This is in addition to `status: already_in_progress` response introduced in previous PR so we avoid the API call altogether.

ToDo:
- [ ] Consider benchmarking telemetry perf degrdation because of lock? 

### All Submissions:


KShivendu:
@ffuugoo @timvisee do you think I should do a benchmark of telemetry perf? Or it's fine?

ffuugoo:
> @ffuugoo @timvisee do you think I should do a benchmark of telemetry perf? Or it's fine?

Should be fine, IMO.

timvisee:
> @ffuugoo @timvisee do you think I should do a benchmark of telemetry perf? Or it's fine? 

Do you mean of telemetry parsing perf in general? We've had problems with it before (on huge clusters), so it's good to be cautious. 

We can do it separately, but I don't consider it high priority right now.

ffuugoo:
> Do you mean of telemetry parsing perf in general?

I think @KShivendu is concerned because telemetry needs to `try_lock` a mutex that is used by recover partial snapshot API.

I personally don't think it's a major concern.

KShivendu:
> because telemetry needs to try_lock a mutex that is used by recover partial snapshot API.. I personally don't think it's a major concern

Yes, this. Do we have already have a telemetry perf bench? I couldn't find one with a quick search.

ffuugoo:
> Do we have already have a telemetry perf bench? I couldn't find one with a quick search.

Honestly, I'd not bother with benchmark (yet). IMO, we can merge this and if we notice perf degradation in cluster-manager we can investigate further, but I personally think it should not matter much.

The only side effect I can think of, is if you make a lot of parallel requests to telemetry on the same machine... but why would you?

PR #6681, opened 2025-06-12
Fix broken dev without default features

coszio:
We recently merged #6638 which didn't include the recently added test without rocksdb in CI.

This fixes the compilation error


coszio:
Closing in favor of #6680

PR #6683, opened 2025-06-13
Add Japanese stopwords

JojiiOfficial:
Depends on #6635


timvisee:
Where did you source these from? I think it'd be good to use a stable source that we also link in a comment in code.

PR #6689, opened 2025-06-13
[gridstore] Priority queue for max gaps

coszio:
This is an optimization that changes insertion from taking `O(n)`, to taking `O(log(n))` (bait headline 🙊 🎣).

In reality finding the region with a fitting gap, although being strictly a linear scan, was really fast, because it is already scoped out enough to not be noticeable with decently-sized workloads (under 2M values).

However, the problem starts to become apparent on larger datasets.

This PR trades finding the right region in `O(n)` and updating it in `O(1)`, with finding it in `O(1)` and updating it in `O(log(n))`. We still have the full-scan fallback in case there is no big enough `max` gap and we need to analyze merging in leading and trailing gaps, but it should be the vast minority of the times.

An added benefit from this, considering that analyzing a particular region is one of the more expensive operations of writes, is that this change allows us to use smaller regions to improve performance.

Here is some basic benchmark results, assuming a dataset of ~8M values in the `bustle_bench`:

| build | region_size_blocks | read_heavy workload (time/op) | insert_heavy workload (time/op) | update_heavy workload (time/op) |
| - | - | - | - | - |
| dev | 8192 (default) | 653ns | 1.743µs | 1.48µs |
| PR | 8192 (default) | 631ns | 1.518µs | 1.474µs |
| dev | 4096 | 630ns | 1.807µs | 1.304µs |
| PR | 4096 | 618ns | 1.442µs | 1.215µs |

### DIFF
--- a/lib/gridstore/Cargo.toml
+++ b/lib/gridstore/Cargo.toml
@@ -34,6 +34,7 @@ common = { path = "../common/common" }
 
 # this is not on dev-dependencies because dev-dependencies cannot be optional :(
 rocksdb = { version = "0.23.0", optional = true }
+priority-queue = "2.5.0"
 
 [dev-dependencies]
 criterion = { workspace = true }
--- a/lib/gridstore/benches/bustle_bench/main.rs
+++ b/lib/gridstore/benches/bustle_bench/main.rs
@@ -15,11 +15,12 @@ type PayloadStorage = Gridstore<Payload>;
 
 fn default_opts(workload: &mut Workload) -> &mut Workload {
     let seed = [42; 32];
-    workload.initial_capacity_log2(21).seed(seed)
+    workload.initial_capacity_log2(23).seed(seed)
 }
 
 fn main() {
-    for num_threads in [1, 2] {
+    #[allow(clippy::single_element_loop)]
+    for num_threads in [1] {
         println!("------------ {num_threads} thread(s) -------------");
         // Read heavy
         println!("**read_heavy** with prefill_fraction 0.95");
--- a/lib/gridstore/src/bitmask/gaps.rs
+++ b/lib/gridstore/src/bitmask/gaps.rs
@@ -6,6 +6,7 @@ use memory::fadvise::clear_disk_cache;
 use memory::madvise::{Advice, AdviceSetting};
 use memory::mmap_ops::{create_and_ensure_length, open_write_mmap};
 use memory::mmap_type::{self, MmapSlice};
+use priority_queue::PriorityQueue;
 
 use super::{RegionId, StorageConfig};
 
@@ -79,6 +80,7 @@ pub(super) struct BitmaskGaps {
     pub path: PathBuf,
     config: StorageConfig,
     mmap_slice: MmapSlice<RegionGaps>,
+    pq: PriorityQueue<RegionId, u16>,
 }
 
 impl BitmaskGaps {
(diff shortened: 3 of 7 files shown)


lib/gridstore/Cargo.toml
```diff
@@ -34,6 +34,7 @@ common = { path = "../common/common" }
 
 # this is not on dev-dependencies because dev-dependencies cannot be optional :(
 rocksdb = { version = "0.23.0", optional = true }
+priority-queue = "2.5.0"
```
timvisee, line 37: Naive question, can we use the native binary heap instead?

https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html

coszio: No, because we need to be able to update any gaps, not just the top one. Particularly when flushing, the top region might not be the same as the one which must free some gaps.

BinaryHeap and PriorityQueue are different because the latter lets you access and modify any item, while the first works best when we are only interested in the top one.

timvisee: That makes a lot of sense. Thanks for elaborating

lib/gridstore/benches/bustle_bench/main.rs
```diff
@@ -15,11 +15,12 @@ type PayloadStorage = Gridstore<Payload>;
...
 fn default_opts(workload: &mut Workload) -> &mut Workload {
     let seed = [42; 32];
-    workload.initial_capacity_log2(21).seed(seed)
+    workload.initial_capacity_log2(23).seed(seed)
 }
 
 fn main() {
-    for num_threads in [1, 2] {
+    #[allow(clippy::single_element_loop)]
+    for num_threads in [1] {
```
agourlay, line 23: why was this changed?

agourlay:
I am challenging a bit this PR to avoid yet another dependency to the repo :)

> However, the problem starts to become apparent on larger datasets.

Is this something you have witnessed on an end to end benchmark? 
The result seems barely visible on a micro benchmark.

My gut feeling is that the `latest_written_region` optimization is enough to amortize the linear scan most of the time.
I do not think we really need to pay the price for `change_priority` on every update.

lib/gridstore/benches/bustle_bench/main.rs
coszio: We are almost always interested in single-threaded, and this allows us to opt-in to check multithreaded easier.

coszio:
@agourlay 

The price for `change_priority` is minimal (`O(log(n)`) compared to the linear scan.

> The result seems barely visible on a micro benchmark.

Well, not barely, it is a 15% boost on `insert_heavy`.

However, I ran several e2e tests which could not conclude any clear advantage with default settings. This may be more noticeable at a larger scale (segment with > 8M points), but up to 8M, it behaves very similary.

lib/gridstore/src/bitmask/gaps.rs
```diff
@@ -79,6 +80,11 @@ pub(super) struct BitmaskGaps {
     pub path: PathBuf,
     config: StorageConfig,
     mmap_slice: MmapSlice<RegionGaps>,
+
+    /// A priority queue which tracks the region with the largest gap.
+    pq: PriorityQueue<RegionId, u16, ahash::RandomState>,
```
timvisee, line 85: I wonder, naive thought. This prioritizes the largest gap. Don't we actually want to prioritize the smallest gap - that still fits?

Won't that help put data on disk in a more compact way, potentially requiring less pages in the end?

Also, don't we risk filling large gaps with small entries here, so that there may be no large enough gap left when trying to fit some large entry?

Let me know if we already walked this path, and/or if this idea makes no sense at all.

coszio: It makes sense, but this would need a different data-structure, something where we can look for the closest greater or equal. 

I didn't really walk the path as you are saying. The pattern I expect this creates is to fill up pages by largest free space first. Only once there are no pages left with enough free space, it will create a new page

timvisee (approved):

coszio:
Let's not do this for the time being

PR #6713, opened 2025-06-17
[gridstore] update a selection instead of a range of regions

coszio:
When doing the optimization for updating bitmask in batch, we assumed that a range of regions was a good interface to select which regions should be re-analyzed for gaps. 

However, when profiling ingestion with many gridstore-backed payload indices, the amount of time `calculate_gaps` takes is very considerable, specially when flushing.

I had the suspicion that the range of region ids was including more regions than actually needed. The changes in this PR switch the method of selecting the dirty regions from a range of start_id..end_id, to a hashset of dirty regions.

Using this command,
```
bfb -n 5000000 --int-payloads 100 --int-payloads 100 --int-payloads 100 --keywords 100 --vectors-per-point 0 --max-id 2000000
```
this PR achieves a stable upsert rate of **~44k RPS, vs ~34k RPS** in dev

### DIFF
--- a/lib/gridstore/src/bitmask/mod.rs
+++ b/lib/gridstore/src/bitmask/mod.rs
@@ -387,9 +387,9 @@ impl Bitmask {
             let start_region_id =
                 (bitmask_range.start / self.config.region_size_blocks) as RegionId;
             let end_region_id =
-                ((bitmask_range.end - 1) / self.config.region_size_blocks) as RegionId;
+                bitmask_range.end.div_ceil(self.config.region_size_blocks) as RegionId;
 
-            dirty_regions.extend(start_region_id..=end_region_id);
+            dirty_regions.extend(start_region_id..end_region_id);
         }
 
         self.update_region_gaps(dirty_regions);


lib/gridstore/src/bitmask/mod.rs
```diff
@@ -198,9 +199,6 @@ impl Bitmask {
             vec![RegionGaps::all_free(self.config.region_size_blocks as u16); new_regions];
         self.regions_gaps.extend(new_gaps.into_iter())?;
 
-        // update the previous last region gaps
-        self.update_region_gaps(previous_bitslice_len - 1..previous_bitslice_len + 2);
```
agourlay, line 202: why can this one be removed?

coszio: The page size is always a multiple of region size, so it does not make sense to update the trailing region before the new page. 

When scanning for an available gap, the leading gap in the new page can be combined with the trailing of the previous page just fine

generall (approved):

timvisee:
I tested a `perf` build on a ramdisk, with the above bfb command I get these results:

- Before PR: 3:12 upload time, 25.5k/s points upserted at the end
- With PR: 2:00 upload time, 40.3k/s points upserted at the end

So that is a nice improvement :+1: 

I do notice that one core is getting a constant 100% CPU. Maybe it's worth to dive into it more in the future to see if there's other potential improvements. Note that that's with just two segments, with more segments I expect we _better_ spread the load.

lib/gridstore/src/bitmask/mod.rs
```diff
@@ -379,27 +377,27 @@ impl Bitmask {
...
             let bitmask_range = (range.start + page_start)..(range.end + page_start);
             self.bitslice[bitmask_range.clone()].fill(used);
 
-            lowest_offset = lowest_offset.min(bitmask_range.start);
-            highest_offset = highest_offset.max(bitmask_range.end);
-        }
+            let start_region_id =
+                (bitmask_range.start / self.config.region_size_blocks) as RegionId;
+            let end_region_id =
+                ((bitmask_range.end - 1) / self.config.region_size_blocks) as RegionId;
```
timvisee: Are we sure about the `-1` here?

Just curious because `bitmask_range = (range.start + page_start)..(range.end + page_start)` is exclusive already.

coszio: This is a bit confusing, yes. I did this so that we could use an inclusive range below (`start_region_id..=end_region_id`). But a more readable form might be using `div_ceil` with an exclusive range

timvisee: Thanks for having a look at it, and for the change!

PR #6714, opened 2025-06-17
Applying a different madvise after PopulateRead cancels its effect

agourlay:
Based on my manual testing, setting a different madvise right after setting `PopulateRead` cancels its effect.
It appears the kernel discards the pages loaded. 

When using a vector storage `on_disk = false`  with this patch:
- The `read_bytes` from `proc/pid/io`  is much higher on startup than `dev`
- No page faults reported during querying using `perf stat -e major-faults -p $(pidof qdrant)` unlike `dev`


agourlay:
We found a different reason for the IO load discrepancy which is unrelated to `PopulateRead`.

I still think that our usage of `PopulateRead` is not bullet proof but I cannot prove it without doubt.

Closing for the time being.

PR #6716, opened 2025-06-17
Use absolute qdrant directory in debian package

jayantpranjal0:
Closes: #3370 

### All Submissions:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -110,7 +110,7 @@ struct Args {
     /// Format: <config_file_path>
     ///
     /// Default path: config/config.yaml
-    #[arg(long, value_name = "PATH")]
+    #[arg(long, value_name = "PATH", env = "QDRANT_CONFIG_PATH")]
     config_path: Option<String>,
 
     /// Disable telemetry sending to developers
--- a/config/deb.yaml
+++ b/config/deb.yaml
@@ -1,2 +1,9 @@
+storage:
+  # Where to store all the data
+  storage_path: /var/lib/qdrant/storage
+
+  # Where to store snapshots
+  snapshots_path: /var/lib/qdrant/snapshots
+
 service:
   static_content_dir: /etc/qdrant/static


jayantpranjal0:
@timvisee can you check out this pr?

src/main.rs
```diff
@@ -110,7 +110,7 @@ struct Args {
     /// Format: <config_file_path>
     ///
     /// Default path: config/config.yaml
-    #[arg(long, value_name = "PATH")]
+    #[arg(long, value_name = "PATH", env = "QDRANT_CONFIG_PATH")]
```
timvisee, line 113: Do we set the correct absolute path as default anywhere here?

I don't think we do. In which case Qdrant will not use the above config when installed by default, and it will not use the configured absolute paths.

jayantpranjal0: No, we set the default value in default config value settings.rs file. In linux after default installation user would have to set up the environment variable to use any other files, which in the debian package is already copied to etc/qdrant/config.yaml.

jayantpranjal0: @timvisee can you let me know if you think anything should be changed. From what I have understood just setting an environment variables by using export in shell environment would do the work. There just adding an environment variables here does the job

jayantpranjal0: @timvisee can you please let me know your thoughts

timvisee: > In linux after default installation user would have to set up the environment variable to use any other files, which in the debian package is already copied to etc/qdrant/config.yaml.

I don't think this is the right approach. It would require additional configuration by the user, but we want to change default behavior here.

If we ship some default configuration and put it on disk, which we do. Qdrant should use the configuration at that path by default.

If we don't do that, the absolute paths this PR sets will not be used in deployments by default.

We should therefore make Qdrant look for the configuration at `/etc/qdrant/config.yaml` by default, but only when packaged for Debian.

timvisee:
Thanks for working on this!

Superseded by <https://github.com/qdrant/qdrant/pull/6958>

PR #6604, opened 2025-05-28
Migrate away from RocksDB based multi dense vector storage

timvisee:
Depends on <https://github.com/qdrant/qdrant/pull/6603>.

In the same vein as <https://github.com/qdrant/qdrant/pull/6603>, but for multi dense vector storage.

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/6603>
- [x] Rebase on `dev`
- [x] Undraft

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

PR #6718, opened 2025-06-18
Fix vector storage migration error

timvisee:
In <https://github.com/qdrant/qdrant/pull/6603>, <https://github.com/qdrant/qdrant/pull/6604>, <https://github.com/qdrant/qdrant/pull/6607> we migrated away from RocksDB based vector storage.

However, it forgot to update the storage type in the segment level configuration. Because of it Qdrant would still try to load the old configuration even though it was already migrated.

The logic for loading segments and their storages is nested deeply. I've implemented this in a way where I keep this nesting intact, while allowing the least amount of possible footguns.

In short: before we migrated each storage individually right after they were loaded. Now we load the full segment, and migrate each storage as part of the full segment. This allows us to access/adjust/store the segment configuration along with it.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/segment_constructor/segment_constructor_base.rs
+++ b/lib/segment/src/segment_constructor/segment_constructor_base.rs
@@ -1,7 +1,6 @@
 use std::collections::HashMap;
 use std::fs::File;
 use std::io::Read;
-use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
@@ -725,8 +724,10 @@ pub fn load_segment(path: &Path, stopped: &AtomicBool) -> OperationResult<Option
         SegmentVersion::save(path)?
     }
 
+    #[cfg_attr(not(feature = "rocksdb"), expect(unused_mut))]
     let mut segment_state = Segment::load_state(path)?;
 
+    #[cfg_attr(not(feature = "rocksdb"), expect(unused_mut))]
     let mut segment = create_segment(
         segment_state.initial_version,
         segment_state.version,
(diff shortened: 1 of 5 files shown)


lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -961,14 +940,85 @@ pub fn migrate_rocksdb_id_tracker_to_mutable(
...
+
+        // Update storage type
+        log::warn!("Updated segment config after data migration");
+        segment_state
+            .config
+            .vector_data
+            .get_mut(vector_name)
+            .unwrap()
+            .storage_type = VectorStorageType::InRamChunkedMmap;
+        Segment::save_state(segment_state, path)?;
```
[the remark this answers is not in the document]
timvisee: Multiple writes are desired. If we fail half way, we still need to have persisted the first half of them.

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -1141,25 +1184,66 @@ pub fn migrate_rocksdb_multi_dense_vector_storage_to_mmap(
...
+}
+
+#[cfg(feature = "rocksdb")]
+fn migrate_all_rocksdb_sparse_vector_storages(
+    path: &Path,
+    segment: &mut Segment,
+    segment_state: &mut SegmentState,
+) -> OperationResult<()> {
+    for (vector_name, data) in &mut segment.vector_data {
+        // Only convert simple dense and multi dense vector storages
```
coszio:
```suggestion
        // Only convert simple sparse vector storages
```

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -752,16 +725,22 @@ pub fn load_segment(path: &Path, stopped: &AtomicBool) -> OperationResult<Option
         SegmentVersion::save(path)?
     }
 
-    let segment_state = Segment::load_state(path)?;
+    let mut segment_state = Segment::load_state(path)?;
```
coszio, line 728: The compiler complains here without default features :)

generall (approved):

timvisee:
@generall suggested to bump the segment version here. While not strictly necessary for compatibility, it would automatically show a incompatibility warning when we eventually remove RocksDB. I'll attempt this idea in a separate PR.

PR #6722, opened 2025-06-19
Add in-memory payload storage on Gridstore

timvisee:
Add an in-memory payload storage type on top of Gridstore.

| On disk | With RocksDB | Without RocksDB |
|---|---|---|
| `true` | `OnDisk` | `Mmap` |
| `false` | `Memory` | `InRamMmap` 🆕 |

It is naively implemented: it is the same as `Mmap`, but it prefetches everything into memory on load. We think that is good enough.

If people report it being problematic due to memory eviction we can later adjust it to use mlock, or proper in-memory structures.

Note: it is not enabled by default yet and put behind the `payload_storage_skip_rocksdb` feature flag. We need to do this to still allow users to downgrade. We can enable the feature flag by default in Qdrant 1.15.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


coszio (changes requested):
There is a difference in terms of performance between in-memory vs on-disk. Even when everything in gridstore is nicely loaded in memory, it still needs to decompress and deserialize every time it is needed (we don't have a cache in place).

I think we need a layer on top of gridstore in the same way we do with in-memory rocksdb payload storage.

generall:
> I think we need a layer on top of gridstore in the same way we do with in-memory rocksdb payload storage.

the idea here is that actually we don't want to have in-ram payload in a long run. We can't deprecate it right away, so that is a middle-ground option. 

As far as I know, there are no real use-cases which could justify using in-ram payload anyway.

coszio:
Okay, just wanted to make the decompression + deserialization overhead explicit.

docs/redoc/master/openapi.json
```diff
@@ -11392,6 +11392,11 @@
             "default": false,
             "type": "boolean"
           },
+          "payload_storage_skip_rocksdb": {
+            "description": "Whether to skip usage of RocksDB for new payload storages.\n\nNew on-disk payload storages were already using Gridstore. In-memory payload storages still choose RocksDB when this flag is not set.\n\nFirst implemented in Qdrant 1.14.2.",
```
coszio: Still targeted for 1.14.2?

coszio (approved):
A remark. I manually checked the time it takes to create a new payload index, and the `in_memory` is about 40% faster than `in_ram_mmap` (0.6s vs 1.05s for 500k points).

Payload indexing is a setup operation and is still much faster than vector indexing, but it's worth noting that this introduces a bit of a regression.

timvisee:
> A remark. I manually checked the time it takes to create a new payload index, and the in_memory is about 40% faster than in_ram_mmap (0.6s vs 1.05s for 500k points).
>
> Payload indexing is a setup operation and is still much faster than vector indexing, but it's worth noting that this introduces a bit of a regression.

Thanks for the test!

The opposite is true for all other storage types where I implemented Gridstore. I'm a bit puzzled on how this could be slower.

Note that this only prefaults memory pages. It does not decompress payload values. Based on that alone I'd expect the RocksDB variant to be slower, because it does decompress all values.

Maybe it's just because blocks are relatively large compared to the payloads we put into it, needing more IO.

I'll dive into it to see if there's anything obvious.

docs/redoc/master/openapi.json
timvisee: 1.15 now. I'll bump it. Thanks!

timvisee:
@coszio Are you sure about the measurements?

I created a collection with: 

```bash
bfb -n 500000 --int-payloads 100 --text-payloads --text-payload-length 10 --uuid-payloads
```

RocksDB storage:
- cold: 612ms
- hot: 549ms

Gridstore storage:
- cold: 316ms
- hot: 5.2ms

timvisee:
> A remark. I manually checked the time it takes to create a new payload index, and the `in_memory` is about 40% faster than `in_ram_mmap` (0.6s vs 1.05s for 500k points).
> 
> Payload indexing is a setup operation and is still much faster than vector indexing, but it's worth noting that this introduces a bit of a regression.

I now notice you're talking about index creation sourced from payloads on Gridstore :man_facepalming:

generall:
Index creation does require de-serializing of the whole payload storage. So it is not surprising that in-ram rocksdb is faster than gridstore in this case. But I don't think it is critical, because:

- you pay the same price, but in a different time
- we don't recommend to create index after uploading data

timvisee:
Agreed.

I'll merge this.

Thanks for the reviews!

PR #6725, opened 2025-06-19
Improve read request handling during partial-snapshot recovery

ffuugoo:
Get rid of `PartialSnapshotMeta::search_lock` reimplemented read request handling during partial recovery based on `ShardReplicaSet::local` and `PartialSnapshotMeta::recovery_lock`.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/replica_set/execute_read_operation.rs
+++ b/lib/collection/src/shards/replica_set/execute_read_operation.rs
@@ -142,7 +142,7 @@ impl ShardReplicaSet {
             None => self.remotes.read().await,
         };
 
-        // We don't need to explicitly check partial snapshot recovery lock, because\
+        // We don't need to explicitly check partial snapshot recovery lock, because
         // - partial snapshot recovery *write-locks* `local` shard when applying partial snapshot
         // - this method *tries* to read-lock `local` shard, and if it's unavailable, fan-out
         //   request to other replicas


lib/collection/src/shards/replica_set/execute_read_operation.rs
```diff
@@ -144,20 +144,29 @@ impl ShardReplicaSet {
...
+        let (local, is_local_ready, update_watcher, partial_snapshot_search_lock) = match local_read
+        {
             Some(local) => {
                 let update_watcher = local.watch_for_update();
-                let is_local_ready = !local.is_update_in_progress();
-                (Some(local), is_local_ready, Some(update_watcher))
+                let partial_snapshot_search_lock = self.try_take_search_read_lock();
+
+                let is_local_ready =
+                    !local.is_update_in_progress() && partial_snapshot_search_lock.is_ok();
```
timvisee: Can't we just use the very same lock to also 'block' when recovering the snapshot? We can't read/write to it anyway while it is happening.

ffuugoo: Yeah, maybe. I'm trying to figure this out rn.

ffuugoo: Got rid of search lock entirely. 🔒🔨

KShivendu:
@ffuugoo have started a docker image build https://github.com/qdrant/qdrant/actions/runs/15821075811

Should be deployed at `ghcr.io/qdrant/qdrant:partial-snapshot-local-shard-read-lock`. Will try to deploy in our testing clusters for you to test.

lib/collection/src/shards/replica_set/execute_read_operation.rs
```diff
@@ -122,8 +134,6 @@ impl ShardReplicaSet {
             )));
         };
 
-        let _partial_snapshot_search_lock = self.try_take_search_read_lock()?;
```
ffuugoo, line 125: This check was useless/harmful.

Partial snapshot recovery takes `local.write()`. If we already managed to take `local.read()` during `execute_local_read_operation`, there's no need to `try_take_search_read_lock`. In worst case it will make us reject read request that we could have successfully performed.

lib/collection/src/shards/replica_set/execute_read_operation.rs
```diff
@@ -141,24 +151,29 @@ impl ShardReplicaSet {
...
             }
+
             None => (None, false, None),
         };
 
         let local_is_active = self.peer_is_active(self.this_peer_id());
 
         let local_operation = if local_is_active {
             let local_operation = async {
-                let _partial_snapshot_search_lock = self.try_take_search_read_lock()?;
```
ffuugoo, line 160: This check was useless/harmful.

Similar to `try_take_search_read_lock` above. If we already managed to take `local.read()`, this check may force us to reject read request, that we could have successfully served.

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -27,12 +27,15 @@ use crate::operations::types::{CollectionError, CollectionResult};
...
+    /// Tracks ongoing *create* partial snapshot requests. There might be multiple parallel
+    /// create partial snapshot requests, so we track them with a counter.
     ongoing_create_snapshot_requests_tracker: RequestTracker,
-    recovery_lock: Arc<tokio::sync::Mutex<()>>,
-    search_lock: Arc<tokio::sync::RwLock<()>>,
+
+    /// Limits parallel *recover* partial snapshot requests. We are using `RwLock`, so that multiple
+    /// read requests can check if recovery is in progress (by doing `try_read`) without blocking
+    /// each other.
+    recovery_lock: Arc<tokio::sync::RwLock<()>>,
```
ffuugoo, line 37: We now (sometimes) check if `recovery_lock` is locked during `execute_local_read_operation`, and so I've swapped `Mutex` for `RwLock`, so that concurrent read operations can do `try_read` without blocking each other.

lib/collection/src/shards/replica_set/execute_read_operation.rs
```diff
@@ -113,7 +113,19 @@ impl ShardReplicaSet {
...
         F: Fn(&(dyn ShardOperation + Send + Sync)) -> BoxFuture<'_, CollectionResult<Res>>,
     {
-        let local = self.local.read().await;
+        let local = match self.local.try_read() {
+            Ok(local) => local,
+            Err(_) => {
+                // Reject read request, if
+                // - `local` shard is not readily available
+                // - and partial snapshot recovery is in progress
+                self.check_partial_snapshot_recovery_lock()?;
```
generall: Technically it is possible that `snapshot_recovery_lock` will be acquired right after our check, and we end up waiting for the  shart recovery here

generall: If it is not possible, please extend comment why

ffuugoo: Recovery lock is taken before local shard lock, so generally this should not be a problem. But it probably might happen under some special circumstances, if local shard is write locked by something other than partial snapshot. 😔

ffuugoo: Should be fixed now.

lib/collection/src/shards/replica_set/execute_read_operation.rs
```diff
@@ -141,24 +142,29 @@ impl ShardReplicaSet {
             None => self.remotes.read().await,
         };
 
+        // We don't need to explicitly check partial snapshot recovery lock, because\
```
timvisee:
```suggestion
        // We don't need to explicitly check partial snapshot recovery lock, because
```

timvisee (approved):
Thanks!

This makes sense. I find the use of a watch channel to cancel a waiting read lock a bit convoluted, but I get why we're using it here. Another lock would not give us the same flexibility.

IIRC you mentioned in call this has the potential for a race, but I don't see a concrete problem.

ffuugoo:
> Thanks!
> 
> This makes sense. I find the use of a watch channel to cancel a waiting read lock a bit convoluted, but I get why we're using it here. Another lock would not give us the same flexibility.
> 
> IIRC you mentioned in call this has the potential for a race, but I don't see a concrete problem.

No more race, but I've made yet another update because a simple `RwLock` should be enough in the end. 🤦‍♀️

lib/collection/src/shards/replica_set/snapshots.rs
```diff
@@ -216,6 +206,13 @@ impl ShardReplicaSet {
         //   Check that shard snapshot is compatible with the collection
         //   (see `VectorsConfig::check_compatible_with_segment_config`)
 
+        let _partial_snapshot_search_lock = match recovery_type {
+            RecoveryType::Full => None,
+            RecoveryType::Partial => {
+                Some(self.partial_snapshot_meta.take_search_write_lock().await)
+            }
+        };
```
timvisee, line 214: Why not do the same on full recovery?

timvisee (approved):

lib/collection/src/shards/replica_set/snapshots.rs
ffuugoo: 1) we don't necessarily want to reject operations during regular recovery, waiting for recovery to end during search ops is fine in that case (but we do need to reject on partial, so that R/W seg works properly)
2) change of existing behavior

timvisee: Fair enough. I get that we assume use cases for full/partial recovery are different. Partial recoveries are ongoing and should not interrupt searches. While we don't expect full recoveries to be common, and they may block reads.

PR #6726, opened 2025-06-19
check if the quantization config changed in order to evaluate if we can re-use HNSW index

generall:
…an re-use HNSW index


IvanPleshkov (approved):
Great finding! Not sure if the whole comparison `old_quantization_config != *new_quantization_config` is necessary (like we can heal hnsw if the user changed quantile parameter in SQ), but I find it okay. Probably, in the future we may do a separate function to check evaluation conditions, I'm sure we can do it later if needed

xzfc (approved):
I've added a commit with a few style nipicks to make it consistent with the existing code.
1. `get_config` renamed to `config`: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter
2. There already `configuration_mismatch` boolean and check below. I've extended it.
3. Compare references, not clones.
4. Put `old_quantization_config` and `new_quantization_config` together.

PR #6728, opened 2025-06-19
Asymmetric binary quantization

IvanPleshkov:
This PR introduces an internal implementation for an upcoming feature: asymmetric binary quantization.
API PR is here: https://github.com/qdrant/qdrant/pull/6201

Motivation
Asymmetric quantization is a quantization where vector and query have different quantization algorithms. In this PR we introduce only one asymmetric option: BQ with query encoded as SQ.

Such methods allow for increased accuracy because a more accurate query produces a more accurate score. But scoring will be more complicated, as not each use case is suitable for asymmetric mechanisms.

An example of BQ quantization with SQ query is a [RaBitQ quantization](https://arxiv.org/pdf/2405.12497). We want to have something similar.

We don't use RaBitQ alg, encoding is different, and only scoring is the same. We have our vision instead. As a pros, we don't need to store additional precomputed constants for each vector, and our approach is compatible with `score_internal` method. As a cons, we require negative numbers in vectors to make it work like in a regular BQ.

We don't use asymmetric BQ in HNSW construction because experiments have shown that HNSW accuracy is poor. No GPU integration is required.

Results on test laion 400M:
```
Regular BQ: Average precision: 62.37%
Asymmetric BQ 8bit query: Average precision: 70.46%
```

### DIFF
--- a/lib/quantization/benches/binary.rs
+++ b/lib/quantization/benches/binary.rs
@@ -4,7 +4,7 @@ use common::counter::hardware_counter::HardwareCounterCell;
 use criterion::{Criterion, criterion_group, criterion_main};
 use permutation_iterator::Permutor;
 use quantization::encoded_vectors::{DistanceType, EncodedVectors, VectorParameters};
-use quantization::encoded_vectors_binary::{EncodedVectorsBin, Encoding};
+use quantization::encoded_vectors_binary::{EncodedVectorsBin, Encoding, QueryEncoding};
 use rand::{Rng, SeedableRng};
 
 fn generate_number(rng: &mut rand::rngs::StdRng) -> f32 {
@@ -40,6 +40,7 @@ fn binary_bench(c: &mut Criterion) {
             invert: false,
         },
         Encoding::OneBit,
+        QueryEncoding::SameAsStorage,
         &AtomicBool::new(false),
     )
     .unwrap();
@@ -80,6 +81,7 @@ fn binary_bench(c: &mut Criterion) {
             invert: false,
         },
         Encoding::OneBit,
+        QueryEncoding::SameAsStorage,
         &AtomicBool::new(false),
     )
     .unwrap();
--- a/lib/quantization/cpp/sse.c
+++ b/lib/quantization/cpp/sse.c
@@ -4,6 +4,11 @@
 
 #include "export_macro.h"
 
+#ifdef _MSC_VER
+#include <intrin.h>
+#define __builtin_popcount __popcnt
+#endif
+
 #define HSUM128_PS(X, R) \
     float R = 0.0f; \
     { \
--- a/lib/quantization/cpp/sse.c
+++ b/lib/quantization/cpp/sse.c
@@ -175,28 +175,19 @@ EXPORT uint32_t impl_xor_popcnt_scalar4_sse_uint128(
         uint64_t v_1 = *v_ptr;
         uint64_t v_2 = *(v_ptr + 1);
 
-        uint64_t q_1_1 = *q_ptr;
-        uint64_t q_2_1 = *(q_ptr + 1);
-        uint64_t q_1_2 = *(q_ptr + 2);
-        uint64_t q_2_2 = *(q_ptr + 3);
-        uint64_t q_1_3 = *(q_ptr + 4);
-        uint64_t q_2_3 = *(q_ptr + 5);
-        uint64_t q_1_4 = *(q_ptr + 6);
-        uint64_t q_2_4 = *(q_ptr + 7);
-
         __m128i popcnt1 = _mm_set_epi32(
-            _mm_popcnt_u64(v_1 ^ q_1_1),
-            _mm_popcnt_u64(v_1 ^ q_1_2),
-            _mm_popcnt_u64(v_1 ^ q_1_3),
-            _mm_popcnt_u64(v_1 ^ q_1_4)
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 0)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 2)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 4)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 6))
         );
         sum = _mm_add_epi32(sum, popcnt1);
 
         __m128i popcnt2 = _mm_set_epi32(
-            _mm_popcnt_u64(v_2 ^ q_2_1),
-            _mm_popcnt_u64(v_2 ^ q_2_2),
-            _mm_popcnt_u64(v_2 ^ q_2_3),
-            _mm_popcnt_u64(v_2 ^ q_2_4)
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 1)),
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 3)),
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 5)),
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 7))
         );
         sum = _mm_add_epi32(sum, popcnt2);
 
--- a/lib/quantization/cpp/sse.c
+++ b/lib/quantization/cpp/sse.c
@@ -105,6 +105,110 @@ EXPORT uint32_t impl_xor_popcnt_sse_uint32(
     return (uint32_t)result;
 }
 
+EXPORT uint32_t impl_xor_popcnt_scalar8_sse_uint128(
+    const uint8_t* query_ptr,
+    const uint8_t* vector_ptr,
+    uint32_t count
+) {
+    const uint64_t* v_ptr = (const uint64_t*)vector_ptr;
+    const uint64_t* q_ptr = (const uint64_t*)query_ptr;
+
+    __m128i sum1 = _mm_set1_epi32(0);
+    __m128i sum2 = _mm_set1_epi32(0);
+    for (uint32_t _i = 0; _i < count; _i++) {
+        uint64_t v_1 = *v_ptr;
+        uint64_t v_2 = *(v_ptr + 1);
+
+        __m128i popcnt1 = _mm_set_epi32(
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 0)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 2)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 4)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 6))
+        );
+        sum1 = _mm_add_epi32(sum1, popcnt1);
+
+        __m128i popcnt2 = _mm_set_epi32(
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 8)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 10)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 12)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 14))
+        );
+        sum2 = _mm_add_epi32(sum2, popcnt2);
+
+        __m128i popcnt3 = _mm_set_epi32(
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 1)),
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 3)),
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 5)),
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 7))
+        );
+        sum1 = _mm_add_epi32(sum1, popcnt3);
+
+        __m128i popcnt4 = _mm_set_epi32(
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 9)),
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 11)),
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 13)),
+            _mm_popcnt_u64(v_2 ^ *(q_ptr + 15))
+        );
+        sum2 = _mm_add_epi32(sum2, popcnt4);
+
+        v_ptr += 2;
+        q_ptr += 16;
+    }
+    __m128i factor1 = _mm_set_epi32(1, 2, 4, 8);
+    __m128i factor2 = _mm_set_epi32(16, 32, 64, 128);
+    __m128 result1_mm128 = _mm_cvtepi32_ps(_mm_mullo_epi32(sum1, factor1));
+    __m128 result2_mm128 = _mm_cvtepi32_ps(_mm_mullo_epi32(sum2, factor2));
+    HSUM128_PS(_mm_add_ps(result1_mm128, result2_mm128), mul_scalar);
+    return (uint32_t)mul_scalar;
+}
+
+EXPORT uint32_t impl_xor_popcnt_scalar4_sse_uint128(
+    const uint8_t* query_ptr,
+    const uint8_t* vector_ptr,
+    uint32_t count
+) {
+    const uint64_t* v_ptr = (const uint64_t*)vector_ptr;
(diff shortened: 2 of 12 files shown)


lib/quantization/cpp/sse.c
```diff
@@ -105,6 +105,110 @@ EXPORT uint32_t impl_xor_popcnt_sse_uint32(
...
+        uint64_t v_2 = *(v_ptr + 1);
+
+        uint64_t q_1_1 = *q_ptr;
+        uint64_t q_2_1 = *(q_ptr + 1);
+        uint64_t q_1_2 = *(q_ptr + 2);
+        uint64_t q_2_2 = *(q_ptr + 3);
+        uint64_t q_1_3 = *(q_ptr + 4);
+        uint64_t q_2_3 = *(q_ptr + 5);
+        uint64_t q_1_4 = *(q_ptr + 6);
+        uint64_t q_2_4 = *(q_ptr + 7);
```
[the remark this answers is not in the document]
IvanPleshkov: order is correct, simplification is done, thanks!

lib/quantization/cpp/sse.c
```diff
@@ -105,6 +105,110 @@ EXPORT uint32_t impl_xor_popcnt_sse_uint32(
...
+    for (uint32_t _i = 0; _i < count; _i++) {
+        uint64_t v_1 = *v_ptr;
+        uint64_t v_2 = *(v_ptr + 1);
+
+        __m128i popcnt1 = _mm_set_epi32(
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 0)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 2)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 4)),
+            _mm_popcnt_u64(v_1 ^ *(q_ptr + 6))
+        );
```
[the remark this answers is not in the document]
IvanPleshkov: Order is correct here, by numbers I didnt mean position in simd register. It is logically connected with `factor` variable where 1 is shift by 1, 2 is shift by 2,..

lib/quantization/cpp/sse.c
```diff
@@ -105,6 +110,365 @@ EXPORT uint32_t impl_xor_popcnt_sse_uint32(
...
+            *(q_ptr + 19),
+            *(q_ptr + 23),
+            *(q_ptr + 27),
+            *(q_ptr + 31),
+        };
+
+        uint64_t values1_u64 = *((const uint64_t*)values1);
+        uint64_t values2_u64 = *((const uint64_t*)values2);
+        uint64_t values3_u64 = *((const uint64_t*)values3);
+        uint64_t values4_u64 = *((const uint64_t*)values4);
```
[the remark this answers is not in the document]
IvanPleshkov: I don't want to use memcpy here because I hope that the compiler can better optimize the `=` statement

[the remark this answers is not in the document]
IvanPleshkov: Are you sure that using `memcpy` doesn't decrease performance?

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -537,11 +807,23 @@ impl<TBitsStoreType: BitsStoreType, TStorage: EncodedStorage> EncodedVectors
 
         let vector_data_usize_1 = transmute_from_u8_to_slice(vector_data_1);
 
-        hw_counter
-            .cpu_counter()
-            .incr_delta(query.encoded_vector.len());
+        hw_counter.cpu_counter().incr_delta(vector_data_1.len());
```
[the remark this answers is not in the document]
IvanPleshkov: Previously, the length of the vector and query were equal. And it didn't matter which `len` to use. Now query is longer and we want to measure acces to vector data instead of query

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -53,14 +80,21 @@ pub trait BitsStoreType:
...
+    + num_traits::cast::ToPrimitive
+    + std::fmt::Debug
 {
     /// Xor vectors and return the number of bits set to 1
     ///
     /// Assume that `v1` and `v2` are aligned to `BITS_STORE_TYPE_SIZE` with both with zeros
     /// So it does not affect the resulting number of bits set to 1
     fn xor_popcnt(v1: &[Self], v2: &[Self]) -> usize;
 
+    fn xor_popcnt_scalar(v1: &[Self], v2: &[Self], bits_count: usize) -> usize;
```
generall: Please provide a docstring with description of what we expect to see in `v1` and `v2`

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -35,16 +35,43 @@ impl Encoding {
...
+    Scalar4bits(EncodedScalarVector<TBitsStoreType>),
+    Scalar8bits(EncodedScalarVector<TBitsStoreType>),
+}
+
 pub struct EncodedBinVector<TBitsStoreType: BitsStoreType> {
     encoded_vector: Vec<TBitsStoreType>,
 }
 
+pub struct EncodedScalarVector<TBitsStoreType: BitsStoreType> {
+    encoded_vector: Vec<TBitsStoreType>,
```
generall, line 100: Please describe the structure of `encoded_vector`. Ideally with ASCII diagram or similar

lib/quantization/tests/integration/test_binary_encodings.rs
```diff
@@ -74,6 +76,110 @@ mod tests {
...
+        }
+
+        let query_encodings = [
+            QueryEncoding::SameAsStorage,
+            QueryEncoding::Scalar4bits,
+            QueryEncoding::Scalar8bits,
+        ];
+
+        let encoded: Vec<_> = query_encodings
+            .iter()
```
xzfc: Style nit: instead of hard-coding values, consider using [`strum::EnumIter`](https://docs.rs/strum/0.27.1/strum/derive.EnumIter.html).

```suggestion
        let encoded: Vec<_> = QueryEncoding::iter()
```

---

Also, the same suggestion to other test like `test_binary_dot_inverted_asymetric`, `test_binary_dot_asymetric`, `test_binary_scalar_internal`, etc.

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -120,6 +154,60 @@ impl BitsStoreType for u8 {
         result
     }
 
+    fn xor_popcnt_scalar(v1: &[Self], v2: &[Self], bits_count: usize) -> usize {
```
xzfc: Performance: would it be beneficial to make `bits_count` checks compile-time rather than run-time?

A simpler (but probably not the cleaner) way is to turn it into const generic.

```rust
fn xor_popcnt_scalar<const BITS_COUNT: usize>(v1: &[Self], v2: &[Self]) -> usize {
    const {
        assert!(BITS_COUNT == 4 || BITS_COUNT == 8);
    };
    
    ...
    
    if BITS_COUNT == 4 {
       ...
    }
```

This check will be compiled out even in debug builds.

---

Maybe with this approach you could merge `xor_popcnt` and `xor_popcnt_scalar` without scarifying the performance.

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -448,12 +706,19 @@ impl<TBitsStoreType: BitsStoreType, TStorage: EncodedStorage>
         vector_stats_path.push("vector_stats.json");
         Some(vector_stats_path)
     }
+
+    pub fn encode_internal_query(&self, point_id: u32) -> EncodedQueryBQ<TBitsStoreType> {
+        let encoded_data = self.get_quantized_vector(point_id).to_vec();
+        EncodedQueryBQ::Binary(EncodedBinVector {
+            encoded_vector: transmute_from_u8_to_slice(&encoded_data).to_vec(),
+        })
```
xzfc: Please use `zerocopy` or `bytemuck` for the new code instead of `transmute_from_u8_to_slice`/`transmute_to_u8_slice`.

Also, the first `.to_vec()` could be avoided.

xzfc:
Not related to this particular PR, but I've got the following errors when compiling this revision:

```
  = note: mold: error: undefined symbol: impl_xor_popcnt_scalar8_sse_uint128
          mold: error: undefined symbol: impl_xor_popcnt_scalar8_sse_u8
          mold: error: undefined symbol: impl_xor_popcnt_scalar8_avx_uint128
          mold: error: undefined symbol: impl_xor_popcnt_scalar4_sse_uint128
          mold: error: undefined symbol: impl_xor_popcnt_scalar4_sse_u8
          mold: error: undefined symbol: impl_xor_popcnt_scalar4_avx_uint128
          collect2: error: ld returned 1 exit status
error: could not compile `quantization` (test "integration") due to 1 previous error
```

Building from scratch by removing the `target` directory helps. It seems that something like this should be added to `build.rs`:

```rust
// Option 1: list each file individually
println!("cargo:rerun-if-changed=cpp/export_macro.h");
if target_arch == "x86_64" {
    println!("cargo:rerun-if-changed=cpp/sse.c");
    println!("cargo:rerun-if-changed=cpp/avx2.c");
    ...
} else if target_arch == "aarch64" {
    println!("cargo:rerun-if-changed=cpp/neon.c");
    ...
}


// Option 2: or just put the whole dir
println!("cargo:rerun-if-changed=cpp");
```

lib/quantization/benches/binary.rs
```diff
@@ -40,6 +40,7 @@ fn binary_bench(c: &mut Criterion) {
             invert: false,
         },
         Encoding::OneBit,
+        QueryEncoding::SameAsStorage,
```
xzfc, line 43: Maybe add benchmarks for new variants?

---

Also, I wonder, how good LLVM auto-vectorization is for this particular case, compared to hand-written intrinsics in C.

Related story: https://www.reddit.com/r/rust/comments/1ha7uyi/comment/m16ke06/

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -377,6 +542,89 @@ impl<TBitsStoreType: BitsStoreType, TStorage: EncodedStorage>
...
+    ) -> EncodedScalarVector<TBitsStoreType> {
+        match encoding {
+            Encoding::OneBit => Self::_encode_scalar_query_vector(query, bits_count),
+            Encoding::TwoBits => {
+                let mut extended_query = query.to_vec();
+                extended_query.extend_from_slice(query);
+                Self::_encode_scalar_query_vector(&extended_query, bits_count)
+            }
+            Encoding::OneAndHalfBits => {
+                let mut extended_query = query.to_vec();
```
xzfc: Performance nit:

```suggestion
                let mut extended_query = Vec::with_capacity(query.len() + query.len().div_ceil(2));
                extended_query.extend(query.iter());
```

And the same for `Encoding::TwoBits`.

Maybe it worth to extract `extended_dim` from `get_quantized_vector_size_from_params` to a separate function and use it here.

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -377,6 +542,89 @@ impl<TBitsStoreType: BitsStoreType, TStorage: EncodedStorage>
...
+    }
+
+    fn _encode_scalar_query_vector(
+        query: &[f32],
+        bits_count: usize,
+    ) -> EncodedScalarVector<TBitsStoreType> {
+        let encoded_query_size = TBitsStoreType::get_storage_size(query.len().max(1)) * bits_count;
+        let mut encoded_query: Vec<TBitsStoreType> = vec![Default::default(); encoded_query_size];
+
+        let max_abs_value = query.iter().map(|x| x.abs()).fold(0.0, f32::max);
```
xzfc, line 644: In the code above you were using `if v[0] > v[1] { v[0] } else { v[1] }`, and here you are using `f32::max`.

Any particular reason why use both?

The former is faster, the latter does extra work to handle `NaN` values.

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -173,6 +302,81 @@ impl BitsStoreType for u128 {
...
+        }
+
+        let mut result = 0;
+        for (&b1, b2_chunk) in vector.iter().zip(query.chunks_exact(query_bits_count)) {
+            for (i, &b2) in b2_chunk.iter().enumerate() {
+                result += (b1 ^ b2).count_ones() << i;
+            }
+        }
+        result as usize
+    }
```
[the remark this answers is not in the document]
IvanPleshkov: fixed

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -120,6 +195,60 @@
...
```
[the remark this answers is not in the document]
IvanPleshkov: fixed

lib/quantization/src/encoded_vectors_binary.rs
IvanPleshkov: Changed `if` into `f32::max`, performance is not very critical in query encoding, we do it once per search

lib/quantization/src/encoded_vectors_binary.rs
IvanPleshkov: Fixed, thanks!

lib/quantization/src/encoded_vectors_binary.rs
IvanPleshkov: Fixed, thanks!

lib/quantization/tests/integration/test_binary_encodings.rs
IvanPleshkov: Thanks for suggestion, added `EnumIter`

IvanPleshkov:
> Not related to this particular PR, but I've got the following errors when compiling this revision:
> 
> ```
>   = note: mold: error: undefined symbol: impl_xor_popcnt_scalar8_sse_uint128
>           mold: error: undefined symbol: impl_xor_popcnt_scalar8_sse_u8
>           mold: error: undefined symbol: impl_xor_popcnt_scalar8_avx_uint128
>           mold: error: undefined symbol: impl_xor_popcnt_scalar4_sse_uint128
>           mold: error: undefined symbol: impl_xor_popcnt_scalar4_sse_u8
>           mold: error: undefined symbol: impl_xor_popcnt_scalar4_avx_uint128
>           collect2: error: ld returned 1 exit status
> error: could not compile `quantization` (test "integration") due to 1 previous error
> ```
> 
> Building from scratch by removing the `target` directory helps. It seems that something like this should be added to `build.rs`:
> 
> ```rust
> // Option 1: list each file individually
> println!("cargo:rerun-if-changed=cpp/export_macro.h");
> if target_arch == "x86_64" {
>     println!("cargo:rerun-if-changed=cpp/sse.c");
>     println!("cargo:rerun-if-changed=cpp/avx2.c");
>     ...
> } else if target_arch == "aarch64" {
>     println!("cargo:rerun-if-changed=cpp/neon.c");
>     ...
> }
> 
> 
> // Option 2: or just put the whole dir
> println!("cargo:rerun-if-changed=cpp");
> ```

Thanks! I chose option 2, done!

lib/quantization/src/encoded_vectors_binary.rs
IvanPleshkov: Done

lib/quantization/src/encoded_vectors_binary.rs
IvanPleshkov: Done

lib/quantization/benches/binary.rs
IvanPleshkov: Autovectorization works great; you're right. But in this case, I added hand-written instrics because they do a measurable boost, We may check if rust can autovectorize popcnt from SSE4.2.
I will add benchmarks as the next PR. This PR is too big already

lib/quantization/src/encoded_vectors_binary.rs
IvanPleshkov: Good suggestion! May do it as a separate PR with benchmark if you don't mind

generall (approved):

PR #6729, opened 2025-06-19
Don't encode quantization query while hnsw build

IvanPleshkov:


lib/quantization/src/encoded_vectors.rs
```diff
@@ -45,6 +45,8 @@ pub trait EncodedVectors: Sized {
     fn score_internal(&self, i: u32, j: u32, hw_counter: &HardwareCounterCell) -> f32;
 
     fn quantized_vector_size(&self) -> usize;
+
+    fn encode_internal_vector(&self, id: u32) -> Option<Self::EncodedQuery>;
```
IvanPleshkov, line 51: Create query from the internal vector. Query is always in RAM

lib/quantization/src/encoded_vectors_pq.rs
```diff
@@ -574,4 +574,8 @@ impl<TStorage: EncodedStorage> EncodedVectors for EncodedVectorsPQ<TStorage> {
     fn quantized_vector_size(&self) -> usize {
         self.metadata.vector_division.len()
     }
+
+    fn encode_internal_vector(&self, _id: u32) -> Option<EncodedQueryPQ> {
+        None
```
IvanPleshkov, line 580: We cannot create query in PQ from quantized vector without LUT accuracy loss.

lib/segment/src/vector_storage/quantized/quantized_query_scorer.rs
```diff
@@ -1,41 +1,45 @@
...
     query: TEncodedVectors::EncodedQuery,
     quantized_data: &'a TEncodedVectors,
-    metric: PhantomData<TMetric>,
-    element: PhantomData<TElement>,
     hardware_counter: HardwareCounterCell,
 }
 
-impl<'a, TElement, TMetric, TEncodedVectors>
-    QuantizedQueryScorer<'a, TElement, TMetric, TEncodedVectors>
+pub enum QuantizedInternalScorerResult<'a, TEncodedVectors>
```
IvanPleshkov, line 23: Result of internal scorer creation. It can be constructed scorer or nothing but we want to get hardware counter ownership back

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -507,6 +507,16 @@ impl<TStorage: EncodedStorage> EncodedVectors for EncodedVectorsU8<TStorage> {
     fn quantized_vector_size(&self) -> usize {
         self.metadata.vector_parameters.dim
     }
+
+    fn encode_internal_vector(&self, id: u32) -> Option<EncodedQueryU8> {
+        let (query_offset, q_ptr) = self.get_vec_ptr(id);
```
generall, line 512: ToDo for another PR: avoid usage of pointers. Pretty suze zero-copy can handle it just fine

generall (approved):
This changes are not exactly pretty, but I don't have any better suggestions

PR #6730, opened 2025-06-19
[phrase matching] Text index fixes

coszio:
Builds on top of #6670.

Two fixes for text index:
- Allow immutable text index with rocks-db storage
- Phrases with repeated tokens were being incorrectly handled when checking specific ids.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index/postings_iterator.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index/postings_iterator.rs
@@ -157,6 +157,7 @@ pub fn check_compressed_postings_phrase<'a>(
     token_to_posting: impl Fn(&TokenId) -> Option<PostingListView<'a, Positions>>,
 ) -> bool {
     let Some(mut posting_iterators): Option<Vec<_>> = phrase
+        .to_token_set()
         .tokens()
         .iter()
         .map(|token_id| token_to_posting(token_id).map(|posting| (*token_id, posting.into_iter())))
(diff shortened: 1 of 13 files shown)


lib/segment/src/index/field_index/full_text_index/text_index.rs
```diff
@@ -676,470 +667,3 @@ impl FieldIndexBuilderTrait for FullTextGridstoreIndexBuilder {
         Ok(index)
     }
 }
-
-#[cfg(test)]
-mod tests {
```
coszio, line 681: This (with a few other diffs) sneaked  back in after being moved in #6638, it is now in `test_congruence.rs` file

lib/segment/src/index/field_index/index_selector.rs
```diff
@@ -363,14 +363,14 @@ impl IndexSelector<'_> {
...
-                &field.to_string(),
-            )),
+            IndexSelector::RocksDb(IndexSelectorRocksDb { db, is_appendable }) => {
+                FieldIndexBuilder::FullTextIndex(FullTextIndex::builder_rocksdb(
+                    Arc::clone(db),
+                    config,
+                    &field.to_string(),
+                    *is_appendable,
+                ))
+            }
```
coszio, line 373: This is where it was selecting an immutable index, but silently creating a mutable one

lib/segment/src/index/field_index/full_text_index/inverted_index/postings_iterator.rs
```diff
@@ -157,6 +157,7 @@ pub fn check_compressed_postings_phrase<'a>(
     token_to_posting: impl Fn(&TokenId) -> Option<PostingListView<'a, Positions>>,
 ) -> bool {
     let Some(mut posting_iterators): Option<Vec<_>> = phrase
+        .to_token_set()
```
coszio, line 160: Not deduplicating here was the main bug

PR #6734, opened 2025-06-20
Actively migrate payload storage from RocksDB to Gridstore

timvisee:
_WIP_

### Tasks
- [ ] Merge <https://github.com/qdrant/qdrant/pull/6722>
- [ ] Also migrate in-memory RocksDB to in-memory Gridstore

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -1315,7 +1321,134 @@ pub fn migrate_rocksdb_sparse_vector_storage_to_mmap(
...
+    }
+
+    // Actively migrate away from RocksDB
+    let new_storage =
+        migrate_rocksdb_payload_storage_to_mmap(segment.payload_storage.borrow().deref(), path)?;
+
+    let old_storage = std::mem::replace(&mut *segment.payload_storage.borrow_mut(), new_storage);
+
+    // Update storage type
+    // TODO(in-memory-mmap-payload-storage): once in-memory mmap payload storage is merged, select correct type here
```
coszio: outdated comment, it seems

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -1315,7 +1321,134 @@ pub fn migrate_rocksdb_sparse_vector_storage_to_mmap(
...
+        new_storage.flusher()()?;
+
+        Ok(new_storage)
+    }
+
+    let new_storage = match migrate(old_storage, segment_path) {
+        Ok(new_storage) => new_storage,
+        // On migration error, clean up and remove all new storage files
+        Err(err) => {
+            let files = find_storage_files(segment_path);
```
coszio: this deletes files, but do we care about directories?

coszio: Also, what happens if migration is aborted? e.g. User tries to cancel the migration by killing the process.

I think the files will still be there, so next time it `open_or_create` it will open instead, but everything will be rewritten... It should be fine.

I guess I answered my own question 😄

coszio (approved):

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -1315,7 +1321,134 @@ pub fn migrate_rocksdb_sparse_vector_storage_to_mmap(
...
+    segment_path: &Path,
+) -> OperationResult<PayloadStorageEnum> {
+    use common::counter::hardware_counter::HardwareCounterCell;
+
+    use crate::payload_storage::PayloadStorage;
+    use crate::payload_storage::mmap_payload_storage::find_storage_files;
+
+    log::info!(
+        "Migrating {} bytes of payload storage from RocksDB into new format",
+        old_storage.get_storage_size_bytes().unwrap_or(0),
```
coszio: Actually, can we report in human-readable format, the same way we do for vector storages? e.g. using GB, MB, KB, etc

lib/segment/src/segment_constructor/segment_constructor_base.rs
timvisee: You're correct.

I can be a bit more aggressive with cleaning here, by deleting the whole directory. The same is not true for the other places I implemented a migration, because it shared the directory.

Let me think for a bit if I like removing the whole directory better.

In practice there should be no difference, as you've also stated.

lib/segment/src/segment_constructor/segment_constructor_base.rs
timvisee: I didn't want to pull in another dependency just for formatting here, and was too lazy to invent a function for it myself :sweat_smile: 

But I'll add one now!

timvisee:
```
2025-06-26T08:03:58.035325Z  INFO storage::content_manager::toc: Loading collection: benchmark
2025-06-26T08:03:59.227097Z  INFO segment::segment_constructor::segment_constructor_base: Migrating 13.93 MiB of payload storage from RocksDB into new format
2025-06-26T08:03:59.265885Z  INFO segment::segment_constructor::segment_constructor_base: Migrating 13.92 MiB of payload storage from RocksDB into new format
```

lib/segment/src/segment_constructor/segment_constructor_base.rs
timvisee: I now drop the full directory: https://github.com/qdrant/qdrant/pull/6734/commits/61f277df6c668abc2ee68ae0d5b2694f67c47d2e

lib/segment/src/segment_constructor/segment_constructor_base.rs
coszio: I vaguely remembered the vector storage migrations to do this kind of human-readable size, I wasn't expecting a new dependency. Looks like the other storages report the number of points

coszio (approved):

PR #6736, opened 2025-06-22
Optimize shard WAL recovery: batch progress-bar updates to cut recovery time

### LINKED ISSUE #6642: Recovering shard taking long period of time
Hello,

I am having an issue with recovering shard from the initialization of qdrant process.
The collection on the file system is about 18GB, and it takes about 10 minutes to process each shard (total 100 shard) as shown below.

Image

We are currently hosting this as docker image on MacOS with the following command:
`docker run -p 6333:6333 -v /Users/user/qdrant/storage:/qdrant/storage qdrant/qdrant`

Created collection with the following config (we use someone's library to create, please let me know if additional info is needed):
`{"m": 32, "ef_construct": 128}`

Could you please guide how this can be optimized to speed up the recovery process?

Thank you!

MagellaX:
Suggested PR Description
Closes #6642.
Problem
During shard startup we replay every WAL record and increment the recovery progress-bar for each operation.
On large collections this can mean millions of updates, which:
acquire the progress-bar’s internal lock every record
emit frequent log / stdout writes in non-TTY environments
noticeably inflate the “Recovering shard…” wall-time (observed 10 min/shard for 18 GB collection, 100 shards)
What’s changed
Introduce WAL_PROGRESS_UPDATE_EVERY (default = 1 000)
Controls how many WAL records are processed before the bar is updated.
Batch progress reporting
For TTY: call bar.inc() only once per WAL_PROGRESS_UPDATE_EVERY processed records.
For non-TTY (hidden bar): keep existing 60-s textual report but compute progress from the new counter.
Ensure accurate final state
After replay finishes, set the bar’s position explicitly to processed_ops to avoid off-by-batch.
Outcome
Dramatically reduces locking and I/O overhead while replaying WAL.
Shard recovery time drops significantly for large WALs (internal test: ~3× faster on 100-shard, 18 GB dataset).
No changes to on-disk data, APIs, or user-facing configuration.
Migration / Compatibility
None required. Behavior is identical except for faster startup.

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -70,6 +70,14 @@ use crate::wal_delta::{LockedWal, RecoverableWal};
 /// If rendering WAL load progression in basic text form, report progression every 60 seconds.
 const WAL_LOAD_REPORT_EVERY: Duration = Duration::from_secs(60);
 
+/// The number of WAL operations to process between progress bar updates.
+/// Updating the progress bar for every single operation adds noticeable
+/// overhead during shard recovery, especially for large collections with
+/// millions of operations.  Batching updates dramatically reduces the
+/// cost of rendering while still providing an accurate sense of progress
+/// to the user.
+const WAL_PROGRESS_UPDATE_EVERY: usize = 1_000;
+
 const WAL_PATH: &str = "wal";
 
 const SEGMENTS_PATH: &str = "segments";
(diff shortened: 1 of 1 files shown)


timvisee:
Thanks for implementing this, though, I don't think it makes much of a difference.

If the problem is just the rapid TTY updates, we can change the refresh rate on the [default](https://docs.rs/indicatif/latest/indicatif/struct.ProgressBar.html#method.new) progress bar. Though I don't see a big advantage there.

If you think the problem is locking inside the progress bar, I'd have to disagree. It mainly uses atomics, which aren't locks at all.

There's a lot more locking happening inside a segment when applying all operations, making any locks here insignificant.

If you think this is still a problem, please provide us a before and after benchmark that shows a clear difference. Then I'd be happy to merge.

> Closes https://github.com/qdrant/qdrant/issues/6642

This PR doesn't solve the issue described, and it's unrelated.

The [issue](https://github.com/qdrant/qdrant/issues/6642) shows it doesn't even use the progress bar. The problem is likely just very expensive/slow operations that are being executed one by one.

PR #6737, opened 2025-06-22
Introduce gated float8 groundwork (E4M3 / E5M2) behind experimental_float8 feature

MagellaX:
This PR lays the foundation for native float8 support without impacting existing builds or collections.
What’s included
Enum extension
Added Float8E4M3 and Float8E5M2 variants to VectorStorageDatatype (segment types).
Variants are compiled only when the new Cargo feature experimental_float8 is enabled.
Conversion helpers
New module lib/common/src/data_types/float8.rs implements fast lookup-table encoders/decoders between the two IEEE-754-2023 float8 formats and f32.
Includes round-trip unit test.

Cargo wiring
Root Cargo.toml defines feature flag:

   ```
[features]
   experimental_float8 = ["dep:once_cell"]

   [dependencies.once_cell]
   version = "1.19"
   optional = true

```

Why split into phases?
Adding the enum & helpers first keeps master green and reduces review surface.
Storage, serialization, and search-layer work will follow in incremental PRs, each building on this gated scaffolding.

Verification

```
cargo check                              # unchanged
cargo check --features experimental_float8   # new paths compile
cargo test  --features experimental_float8   # includes float8 round-trip test
```

No behavioral change when the feature flag is off, so it’s fully backward-compatible and safe to merge.
Closes the initial plumbing step for Issue #6581.

This is Step 1 of N toward full float8 support.
What’s in this PR (Step 1)
Enum extension
Added Float8E4M3 and Float8E5M2 variants to VectorStorageDatatype, gated by the experimental_float8 Cargo feature.
Conversion helpers
lib/common/src/data_types/float8.rs implements fast encode/​decode LUTs for both formats plus a unit test.
Cargo wiring
New feature flag in root Cargo.toml (experimental_float8 = ["dep:once_cell"]).
Default build path is unchanged; enabling the flag compiles the new code.
Why split the work?
Adding the enum touches many exhaustive match statements; gating it prevents widespread refactors until the rest of the stack is ready, keeping master stable and reviews focused.
Next Steps (planned follow-up PRs)
Serialization & Schemas
Extend gRPC/REST Datatype enum and OpenAPI docs behind the same feature flag.
Ensure snapshots, dumps, and replicas tolerate the new enum value.
Disk Persistence
Encode incoming f32/​f16 vectors to float8 on write; decode to f32 on segment load.
Validate segment file format versioning for back-compat.
Search Path Integration
Modify vector storages to accept buffer slices in float8 and perform on-the-fly decode, or load-time decode for RAM-backed storages.
Bench to keep search latency within target bounds.
Collection API Updates
Allow users to specify data_type: Float8E4M3 | Float8E5M2 when creating collections (flag-gated).
Update validation, strict-mode rules, and error messages.
Robust Testing & Benchmarks
Extensive integration tests (CRUD, search accuracy, snapshot round-trip).
Performance benchmarks vs. float16 baseline.
Stabilization & Flag Removal
After sufficient soak time, lift the experimental_float8 gate and migrate existing collections with an on-startup transcode path if necessary.

How to try it today

```
cargo check --features experimental_float8
cargo test  --features experimental_float8
```

Merging this PR unblocks parallel work on the subsequent steps while guaranteeing no behavior changes for users until the feature is explicitly enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,8 @@ deb = []
 # More work is still required to eventually completely disable and remove it
 rocksdb = ["segment/rocksdb"]
 
+experimental_float8 = ["dep:once_cell"]
+
 [dev-dependencies]
 serde_urlencoded = "0.7"
 sealed_test = "1.1.0"
@@ -326,3 +328,7 @@ assets = [
     ["static/*/*", "etc/qdrant/static", "644"],
     ["config/deb.yaml", "etc/qdrant/config.yaml", "644"],
 ]
+
+[dependencies.once_cell]
+version = "1.19"
+optional = true
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -70,6 +70,14 @@ use crate::wal_delta::{LockedWal, RecoverableWal};
 /// If rendering WAL load progression in basic text form, report progression every 60 seconds.
 const WAL_LOAD_REPORT_EVERY: Duration = Duration::from_secs(60);
 
+/// The number of WAL operations to process between progress bar updates.
+/// Updating the progress bar for every single operation adds noticeable
+/// overhead during shard recovery, especially for large collections with
+/// millions of operations.  Batching updates dramatically reduces the
+/// cost of rendering while still providing an accurate sense of progress
+/// to the user.
+const WAL_PROGRESS_UPDATE_EVERY: usize = 1_000;
+
 const WAL_PATH: &str = "wal";
 
 const SEGMENTS_PATH: &str = "segments";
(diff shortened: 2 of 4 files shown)


generall:
Hey @MagellaX, thanks for your PR, but have you tried to benchmark float8?

Not so long ago I made this simple repo https://github.com/qdrant/float8-experiment to validate the float8 precision claims and was disappointed.

Do you have other data? If so, it would be much easier to validate the approach separately before actually integrating it into qdrant core.

MagellaX:
Did u read what i said, it's a startup of follow-up PR's that i'll do later, it's just an experimental one, and yes i have to benchmark.. thanks for pointing that!!

generall:
Benchmarking inside already pretty complex codebase might be a non-trivial task. I would prefer to have external benchmarks before introducing more complexity into the core.

generall:
I am closing this PR for now, please feel free to re-open once you have benchmarks

PR #6749, opened 2025-06-23
[phrase match | strict mode] Allow phrase condition when enabled in index

coszio:
Builds on top of #6670 

Fixes the check for phrase index, so that strict mode allows phrase conditions when the text index has `phrase_matching` enabled

### DIFF
--- a/lib/segment/Cargo.toml
+++ b/lib/segment/Cargo.toml
@@ -48,7 +48,6 @@ anyhow = "1.0.98"
 pprof = { workspace = true }
 
 [dependencies]
-bitpacking = "0.9.2"
 bytemuck = { workspace = true }
 data-encoding = { workspace = true }
 delegate = { workspace = true }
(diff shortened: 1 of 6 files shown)


lib/segment/Cargo.toml
```diff
@@ -45,7 +45,6 @@ anyhow = "1.0.98"
 pprof = { workspace = true }
 
 [dependencies]
-bitpacking = "0.9.2"
```
coszio, line 51: Additionally, removes bitpacking dep here

coszio:
@generall in https://github.com/qdrant/qdrant/pull/6749/commits/71e78a58baf9b50bbaa587cb0a1f367df88cd6d3 I have modified the Display impl to account for phrase matching, and parametrized integer index. Although the integer index is still recommended unparametrized

This is the new error:
```json
{
  "error": "Bad request: Index required but not found for \"t\" of one of the following types: [text (with phrase_matching: true)]. Help: Create an index for this key or use a different filter."
}
```

generall (approved):

PR #6750, opened 2025-06-24
[gridstore] use seq mmap in iter

coszio:
Most changes are related to a refactor to use const generics, which removes a bit of code duplication by drilling down a const generic.

Main change is to use the seq mmap in the `Gridstore::iter` function. Looks like we missed this when introducing the seq mmap

### DIFF
--- a/lib/gridstore/src/gridstore.rs
+++ b/lib/gridstore/src/gridstore.rs
@@ -199,52 +199,28 @@ impl<V: Blob> Gridstore<V> {
     }
 
     /// Read raw value from the pages. Considering that they can span more than one page.
-    fn read_from_pages(
-        &self,
-        start_page_id: PageId,
-        block_offset: BlockOffset,
-        length: u32,
-    ) -> Vec<u8> {
-        self.read_from_pages_with_read_fn(
-            start_page_id,
-            block_offset,
-            length,
-            |page, block_offset: BlockOffset, length: u32, block_size_bytes: usize| {
-                page.read_value(block_offset, length, block_size_bytes)
-            },
-        )
-    }
-
-    /// Read raw value from the pages. Considering that they can span more than one page.
-    fn read_from_pages_sequential(
-        &self,
-        start_page_id: PageId,
-        block_offset: BlockOffset,
-        length: u32,
-    ) -> Vec<u8> {
-        self.read_from_pages_with_read_fn(
-            start_page_id,
-            block_offset,
-            length,
-            |page, block_offset: BlockOffset, length: u32, block_size_bytes: usize| {
-                page.read_value_sequential(block_offset, length, block_size_bytes)
-            },
-        )
-    }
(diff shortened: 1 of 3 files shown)


lib/gridstore/src/gridstore.rs
```diff
@@ -543,7 +520,7 @@ impl<V: Blob> Gridstore<V> {
                 length,
             } = pointer;
 
-            let raw = self.read_from_pages(page_id, block_offset, length);
+            let raw = self.read_from_pages::<true>(page_id, block_offset, length);
```
coszio, line 523: This is the main change (which could've been done without the refactor, but it was a quick change 😸)

generall (approved):

agourlay (approved):

lib/gridstore/src/gridstore.rs
timvisee: Yes! I also put it [here](https://github.com/qdrant/qdrant/pull/6734/commits/9d332d0282356c1371b95a5143be5b4e97dda633#diff-ff9404262bc6103b41d148da4cb172961dd2325998527fc4e8a725b73cf3af92R546).

lib/gridstore/src/page.rs
```diff
@@ -88,38 +88,36 @@ impl Page {
...
     /// - None if the value is not within the page
     /// - Some(slice) if the value was successfully read
     ///
     /// # Panics
     ///
-    /// When the `block_offset` starts after the page ends
-    ///
-    pub fn read_value(
+    /// If the `block_offset` starts after the page ends.
+    pub fn read_value<const READ_SEQUENTIAL: bool>(
```
timvisee, line 100: Any reason for making this boolean generic? I don't like it that much I must admit.

My assumption is that the compiler will produce exactly the same code when we pass a boolean as argument, because it by itself it constant and we have optimizations enabled.

We can probably check it with [Godbolt](https://rust.godbolt.org/).

timvisee (approved):

lib/gridstore/src/page.rs
timvisee: Yes, it is unnecessary:

https://rust.godbolt.org/z/heEPzY36n

Putting it in a regular parameter produces exactly the same ASM.

I therefore vote to put it back into parameters.

timvisee: Scrap that, it is not the same! Made a small mistake not including the return.

https://rust.godbolt.org/z/TeMdb6ahq

timvisee: In our daily we agreed to keep it as is. Changing it is not significant enough. And the current approach actually guarantees that we must provide a constant (so the compiler can optimize it away).

coszio: Today I learned about this tool, thanks!

PR #6751, opened 2025-06-24
Bump Cargo.lock, syn 2.0.104

timvisee:
The `syn` crate is aggressively bumped to a newer version in our `Cargo.lock`.

For example, `cargo check --workspace` will create this same change.

### All Submissions:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3239,7 +3239,7 @@ checksum = "03343451ff899767262ec32146f6d559dd759fdadf42ff0e227c7c48f72594b4"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.103",
+ "syn 2.0.104",
 ]
 
 [[package]]


agourlay (approved):

timvisee:
I'm not exactly sure why it is bumped, because 2.0.103 is [not yanked](https://crates.io/crates/syn/versions). :man_shrugging:

xzfc:
> I'm not exactly sure why it is bumped

1. The `dev` as for aadb0bd7e653e4be5e6c25a99dcee474e5390b2b were depending on `syn 2.0.103`.
2. https://github.com/qdrant/qdrant/pull/6745 transitively removed `syn 2.0.103` and added `syn 2.0.104`.
3. https://github.com/qdrant/qdrant/pull/6748 added new transitive dependency `jiff-static` > `syn 2.0.103`.

Separately, 2 and 3 were consistent, but after merging them both, `jiff-static` started depending on a `syn` version that was no longer present in the lockfile.

---

Some ideas to possibly prevent similar cases:
- Ask `@dependabot rebase` or `@dependabot recreate` explicitly. (not sure if it will help)
- Add lightweight CI that calls `cargo metadata --locked --no-deps` on merges into `dev` and opens a PR when it's inconsistent.
- Set `rebase-strategy: auto` in dependabot config. (but I suspect it can cause quadratic number of CI builds)
- Configure dependabot to open one big PR instead of ten small PRs: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--

timvisee:
Thanks for the investigation!

I'm not to worried about this happening again, so we can leave things as-is if you ask me. If we're seeing this again a few times, we can adjust.

I also expect that `@dependabot rebase` would fix the issue. But it would make merging all dependabot PRs very cumbersome, because it can easily take 4 hours.

I don't like enabling the feature for one big PR, because there are quite some PR suggestions we don't want to merge. It'll be annoying to manage if it's within a single big PR.

PR #6752, opened 2025-06-24
Add validation for legacy collection names

timvisee:
A long time ago we added strict validation for new collection names:

https://github.com/qdrant/qdrant/blob/b83ace82f090856add0f2d6a7d6184fd8d443ded/src/actix/api/mod.rs#L30-L33

But, we kept very limited validation for all collection names that could already exist to not break existing deployments:

https://github.com/qdrant/qdrant/blob/b83ace82f090856add0f2d6a7d6184fd8d443ded/src/actix/api/mod.rs#L44

One thing we can guarantee, is that existing collections could never have had `/` or `\0` characters, because those are [disallowed](https://stackoverflow.com/a/31976060/1000145) on both Linux and Windows platforms.

This implements the above idea as additional validation on APIs that use existing collection names.

My main motivation for this is to be even more sure that `/` can never reach cluster manager in a collection name, as it would break it's functionality.

### All Submissions:

### New Feature Submissions:


lib/common/common/src/validation.rs
```diff
@@ -77,6 +77,29 @@ pub fn validate_collection_name(value: &str) -> Result<(), ValidationError> {
...
+        Some(c) => {
+            let mut err = ValidationError::new("does_not_contain");
+            err.add_param(Cow::from("pattern"), &c);
+            err.message
+                .replace(format!("collection name cannot contain \"{c}\" char").into());
+            Err(err)
+        }
+        None => Ok(()),
+    }
+}
```
timvisee, line 101: New validation function is here

timvisee:
CI is currently blocked by <https://github.com/qdrant/qdrant/pull/6751>

lib/api/build.rs
KShivendu, line 1: There are multiple lines in this file where have `.collection_name` but not `custom(function = \"common::validation::validate_collection_name_legacy\")`

Some examples would be `UpdatePointVectors.collection_name`, `SearchPoints.collection_name`. Or was there a reason you didn't add the validation there? :smiley:

KShivendu (changes requested):
Thanks for the initiative! I think some validations were missed. Please have a look :D

lib/api/build.rs
timvisee: That's indeed a mistake. It appears my find replace somehow didn't work properly.

Fixed it now. Thanks!

PR #6755, opened 2025-06-24
check if we try to deactivate last initializing replica

generall:
There is a bug, which I can't reproduce locally, but it was observed on practice multiple times:

If pod was somehow killed during collection creation or there was an error during creating a collection (due to file descriptors or something like that), it might be possible that some shards of the collection have inconsistent state between `initializing` and `dead`.

Local shard thinks the shard is `dead` while other machines in the cluster consider it `initializing`. 

Since local shard status it `dead` it needs to recover it from somewhere, but it is also the only shard in the cluster. So cluster is stuck in this inconsistent state without ability to recover (except for collection deletion).

This PR extends our check for `is_last_active_replica` and handles the case of no active replicas in more details.

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -411,25 +411,20 @@ impl ShardReplicaSet {
         self.replica_state.read().peers()
     }
 
-    pub fn is_last_active_replica(&self, peer_id: PeerId) -> bool {
+    /// Checks if the current replica contains a unique source of truth and should never
+    /// be deactivated or removed.
+    /// If current replica is the only "alive" replica, it is considered the last source of truth.
+    ///
+    /// This includes `initializing` replica, as even if it contains no data,
+    /// as if it is deactivated, it can never be recovered.
+    pub fn is_last_source_of_truth_replica(&self, peer_id: PeerId) -> bool {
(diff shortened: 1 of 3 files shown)


lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -411,27 +411,46 @@ impl ShardReplicaSet {
...
-    pub fn is_last_active_replica(&self, peer_id: PeerId) -> bool {
+    /// Checks if the current replica contains a unique source of truth and should never
+    /// be deactivated or removed.
+    /// If current replica is the only "alive" replica, it is considered the last source of truth.
+    ///
+    /// This includes `initializing` replica, as even if it contains no data,
+    /// as if it is deactivated, it can never be recovered.
+    ///
+    /// Same logic applies to `Listener` replicas, as they are not recoverable if
+    /// there are no other active replicas.
```
timvisee: Nit: I find this a bit more concrete, also noting when it might happen:

```suggestion
    /// If our replica is `Initializing`, we consider it to be the last source of truth if there is
    /// no other active replicas. If we would deactivate it, it will be impossible to recover the
    /// replica later. This may happen if we got killed or crashed during collection creation.
    ///
    /// Same logic applies to `Listener` replicas, as they are not recoverable if there are no
    /// other active replicas.
```

PR #6756, opened 2025-06-25
Configurable HNSW healing threshold

xzfc:
This PR makes HNSW healing treshold configurable through a new setting.

Particularly:
1. HNSW healing is enabled by default.
2. The default healing threshold is changed from 10% to 30%.
3. Added float setting `QDRANT__STORAGE__HNSW_GLOBAL_CONFIG__HEALING_THRESHOLD`, (`0.3` by default).
4. Removed boolean setting `QDRANT__FEATURE_FLAGS__HNSW_HEALING`.
   To disable healing completely, set the new setting to `0.0`.

The new setting value is exposed in telemetry when `details_level >= 1`:
```json
{
  "id": "2ad2b301-494c-4160-bdc7-ac786d9fc79a",
  "app": {
    "name": "qdrant",
    "version": "1.14.2-dev",
    "features": {
      "debug": false,
      "service_debug_feature": false,
      "recovery_mode": false,
      "gpu": false,
      "rocksdb": true
    },
    "runtime_features": {
      "all": false,
      "payload_index_skip_rocksdb": true,
      "payload_index_skip_mutable_rocksdb": false,
      "incremental_hnsw_building": true,
      "hnsw_healing": true,
      "migrate_rocksdb_id_tracker": false,
      "migrate_rocksdb_vector_storage": false
    },
    "hnsw_global_config": {
      "healing_threshold": 0.3        // ←
    },
```

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -573,6 +573,24 @@ impl HnswConfig {
     }
 }
 
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Validate, Anonymize, Clone)]
+#[serde(rename_all = "snake_case", default)]
+#[anonymize(false)]
+pub struct HnswGlobalConfig {
+    /// Enable HNSW healing if the ratio of missing points is no more than this value.
+    /// To disable healing completely, set this value to `0.0`.
+    #[validate(range(min = 0.0, max = 1.0))]
+    pub healing_threshold: f64,
+}
+
+impl Default for HnswGlobalConfig {
+    fn default() -> Self {
+        Self {
+            healing_threshold: 0.3,
+        }
+    }
+}
+
 const fn default_max_indexing_threads() -> usize {
     0
 }
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -11362,6 +11362,16 @@
               }
             ]
           },
+          "hnsw_global_config": {
+            "anyOf": [
+              {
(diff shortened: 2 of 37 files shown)


lib/segment/src/types.rs
```diff
@@ -573,6 +573,24 @@ impl HnswConfig {
...
+    pub healing_threshold: f64,
+}
+
+impl Default for HnswGlobalConfig {
+    fn default() -> Self {
+        Self {
+            healing_threshold: 0.3,
+        }
+    }
+}
```
timvisee, line 592: Do we create a separate type here, just to prevent persisting it and to hide it from users?

If we'd put it inside `HnswConfig`, can we use `#[serde(skip)]` to achieve the same, or will that be annoying to use?

xzfc: > Do we create a separate type here, just to prevent persisting it and to hide it from users?

Yes.

> If we'd put it inside `HnswConfig`, can we use `#[serde(skip)]` to achieve the same, or will that be annoying to use?

You mean, deserealize it as usual, but inject this value after deserealizing/converting from other types? Seem annoying to me, and it'll be easy to miss some cases.

Even if we use `#[serde(skip_serializing)]`, we'd still need to inject it as it could come from consensus/read from disk.

timvisee: Yes, makes sense. Injecting is not nice either. Let's keep it like this :+1:

timvisee (approved):

PR #6757, opened 2025-06-25
Enable pod role based auth for s3 snapshots

silverjam:
We'd like to be able to run back-ups from EKS pods that use role based authentication to our S3 buckets.  To support this, it seems qdrant must use the [AmazonS3Builder::from_env](https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html#method.from_env) method to populate the builder with information from the environment per this feature request from the [object-store](https://github.com/apache/arrow-rs-object-store/issues/282) library.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/common/snapshots_manager.rs
+++ b/lib/collection/src/common/snapshots_manager.rs
@@ -60,7 +60,7 @@ impl SnapshotStorageManager {
                 Ok(SnapshotStorageManager::LocalFS(SnapshotStorageLocalFS))
             }
             SnapshotsStorageConfig::S3 => {
-                let mut builder = AmazonS3Builder::new();
+                let mut builder = AmazonS3Builder::from_env();
                 if let Some(s3_config) = &snapshots_config.s3_config {
                     builder = builder.with_bucket_name(&s3_config.bucket);
 


lib/collection/src/common/snapshots_manager.rs
```diff
@@ -60,8 +61,13 @@ impl SnapshotStorageManager {
                 Ok(SnapshotStorageManager::LocalFS(SnapshotStorageLocalFS))
             }
             SnapshotsStorageConfig::S3 => {
-                let mut builder = AmazonS3Builder::new();
-                if let Some(s3_config) = &snapshots_config.s3_config {
+                let builder = if let Some(s3_config) = &snapshots_config.s3_config {
```
silverjam: Possibly could just replace `AmazonS3Builder::new` with `AmazonS3Builder::from_env` method, but not sure if there are cases where this would not be desirable behavior.

silverjam:
I ran `cargo test --workspace` and it looks like everything is passing.

lib/collection/src/common/snapshots_manager.rs
timvisee: I cannot think of a case where we would not want to source details from the environment. Let's remove `with_env` from our config, and always use [`from_env`](https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html#method.from_env).

If this ever becomes an issue in the future, we can always add the config option back.

timvisee (changes requested):
Happy to merge after we resolve [this](https://github.com/qdrant/qdrant/pull/6757#discussion_r2166160942).

lib/collection/src/common/snapshots_manager.rs
silverjam: Fixed, thanks!

silverjam:
For posterity: we tested this internally and it definitely works for our use case, thank you for the quick turnaround.

timvisee:
Awesome. Thanks for giving it a shot! And thanks for the straight-to-the-point PR.

timvisee:
Good news! We've just released [Qdrant 1.15](https://github.com/qdrant/qdrant/releases/tag/v1.15.0). It includes this change.

silverjam:
Wonderful, thank you!

PR #6759, opened 2025-06-25
inference errors

generall:
- **fix error handling in inference**
- **fmt**

### DIFF
--- a/src/common/inference/service.rs
+++ b/src/common/inference/service.rs
@@ -214,7 +214,7 @@ impl InferenceService {
             }
             Err(error) => {
                 if let Some(status) = error.status() {
-                    (format!("{error}"), status)
+                    (error.to_string(), status)
                 } else {
                     return Err(StorageError::service_error(format!(
                         "Failed to send inference request: {error}"
--- a/src/common/inference/service.rs
+++ b/src/common/inference/service.rs
@@ -13,8 +13,8 @@ use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use storage::content_manager::errors::StorageError;
 
-use crate::common::inference::InferenceToken;
 use crate::common::inference::config::InferenceConfig;
+use crate::common::inference::InferenceToken;
 
 const DOCUMENT_DATA_TYPE: &str = "text";
 const IMAGE_DATA_TYPE: &str = "image";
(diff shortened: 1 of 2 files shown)


src/common/inference/service.rs
```diff
@@ -193,24 +198,30 @@ impl InferenceService {
...
+                    Err(err) => {
+                        return Err(StorageError::service_error(format!(
+                            "Failed to read inference response body: {err}"
+                        )));
+                    }
+                }
+            }
+            Err(error) => {
+                if let Some(status) = error.status() {
+                    (format!("{error}"), status)
```
timvisee:
```suggestion
                    (error.to_string(), status)
```

timvisee (approved):

src/common/inference/service.rs
```diff
@@ -193,24 +198,30 @@ impl InferenceService {
...
-        let response = self
-            .client
-            .post(url)
-            .json(&request)
-            .send()
-            .await
-            .map_err(|e| {
-                let error_body = e.to_string();
-                StorageError::service_error(format!(
-                    "Failed to send inference request to {url}: {e}, error details: {error_body}",
```
n0x29a, line 205: The idea if this string was to split it: return just generic error to UI and log the detailed message. Something like that:

log::error!("Inference request error: {e}");
StorageError::service_error("Inference request failed.")

src/common/inference/service.rs
```diff
@@ -63,6 +63,11 @@ pub enum InferenceData {
     Object(InferenceObject),
 }
 
+#[derive(Debug, Deserialize)]
+struct InferenceError {
+    pub error: String,
```
n0x29a, line 68: it will be nice to add more control with inference communication. For example, it can return '429 Too Many Requests'. In this case instead of returning the error string will be better to add actionable point to provide a resolution of this error to our customer. 
It could be `pub retry_after: Option<u64>,` field which will tell explicitly what to do with this error on the API level.

n0x29a (approved):

src/common/inference/service.rs
generall: why?

src/common/inference/service.rs
generall: This makes sense, but maybe better done in a separate PR

src/common/inference/service.rs
timvisee: I like propagating the error here, because it gives the user some context rather than getting a vague message. It's not obvious to look into server logs just to get an idea of what is happening. I vote to keep it like this.

PR #6760, opened 2025-06-26
remove max indexing threads from config

generall:
Two reasons for it:

- option is obsolete and there are no cases to use it anymore. It's functionality is completely replaces by `optimizer_cpu_budget`
- this option might be confusing, as its name overlaps with per-collection configuration, which have different semantics and different behavior for 0 value

- **remove max_optimization_threads from config.yaml**
- **fmt**

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -64,11 +64,6 @@ storage:
     # Number of parallel threads used for search operations. If 0 - auto selection.
     max_search_threads: 0
 
-    # Max number of threads (jobs) for running optimizations across all collections, each thread runs one job.
-    # If 0 - have no limit and choose dynamically to saturate CPU.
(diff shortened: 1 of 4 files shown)


generall:
> Back in the day we kept it for backwards compatibility, but I'm okay with removing it in 1.15.

I would argue that we don't want this functionality to be adjustable anymore

Adit2607:
in the release, it says max_indexing_threads is deprecated, but in the PR can see max_optimization_threads are removed from config. Are these linked? should we use optimizer_cpu_budget for indexing threads also?

timvisee:
> in the release, it says max_indexing_threads is deprecated, but in the PR can see max_optimization_threads are removed from config. Are these linked?

It is `max_optimization_threads` that is removed. I'll update the release notes.

> should we use optimizer_cpu_budget for indexing threads also?

Yes

PR #6763, opened 2025-06-26
Fix sq internal vector encoding

IvanPleshkov:
This PR fixes SQ accuracy lost after https://github.com/qdrant/qdrant/pull/6729.

The problem was with the calculation of the vector's `offset` value (what is the `offset` value for an SQ vector? You can find it here: https://qdrant.tech/articles/scalar-quantization/). 

This value is dirrerent between vector and query. For query (in Dot metric) it's
```
values.iter().cloned().sum::<f32>()
    * self.metadata.alpha
    * self.metadata.offset
```

For the vector it's
```
dim as f32 * offset * offset + values.iter().cloned().sum::<f32>()
    * self.metadata.alpha
    * self.metadata.offset
```

This PR fixes difference in query internal method. Test included.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/quantization/src/encoded_vectors_u8.rs
+++ b/lib/quantization/src/encoded_vectors_u8.rs
@@ -518,14 +518,13 @@ impl<TStorage: EncodedStorage> EncodedVectors for EncodedVectorsU8<TStorage> {
                 self.metadata.actual_dim as f32 * self.metadata.offset * self.metadata.offset
             }
         };
-        let offset_difference = if self.metadata.vector_parameters.invert {
-            -offset_difference
-        } else {
-            offset_difference
-        };
 
         let (query_offset, q_ptr) = self.get_vec_ptr(id);
-        let query_offset = query_offset + offset_difference;
+        let query_offset = if self.metadata.vector_parameters.invert {
+            query_offset + offset_difference
+        } else {
+            query_offset - offset_difference
+        };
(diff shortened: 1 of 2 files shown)


lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -509,7 +509,22 @@ impl<TStorage: EncodedStorage> EncodedVectors for EncodedVectorsU8<TStorage> {
...
 
     fn encode_internal_vector(&self, id: u32) -> Option<EncodedQueryU8> {
+        let offset_difference = match self.metadata.vector_parameters.distance_type {
+            DistanceType::Dot => {
+                self.metadata.actual_dim as f32 * self.metadata.offset * self.metadata.offset
+            }
+            DistanceType::L1 => 0.0,
+            DistanceType::L2 => {
+                self.metadata.actual_dim as f32 * self.metadata.offset * self.metadata.offset
+            }
```
timvisee, line 519: These are the same. We can merge them if we like:

```suggestion
            DistanceType::Dot | DistanceType::L2 => {
                self.metadata.actual_dim as f32 * self.metadata.offset * self.metadata.offset
            }
            DistanceType::L1 => 0.0,
```

generall (approved):

PR #6766, opened 2025-06-26
Optimize Text payload matcher

agourlay:
The initial goal of this PR was to fix an over measurement of `payload_index_io_read` for the `Text` index.

In practice, the final fix is a performance optimization decreasing the hardware counter as a side effect.

In my tests, the `payload_index_io_read` value for `Text` matching is almost 3x higher than what is reported by the `read_bytes` of the process. 

In case of high cardinality `Text` matching, the search is not driven directly by the payload index.
Meaning a potentially large number of points need to be tested against the payload index to check if the condition matches.

The current approach in `dev` lazily recomputes the set of points which match the query tokens for each point repeatedly.
This set is a constant for all points tested.

This inflates the number of `payload_index_io_read` of ops artificially as the mmap slice is cached. 

The proposed fix is to pre-calculate the intersection of points for the query tokens to reuse it for each point test.

I was able to show locally that this change:
- fixes the `payload_index_io_read` to be much closer to `read_bytes`
- makes the query slightly faster

The trade-off is to use more memory once to avoid repeated IO.

The set of points kept in memory is the intersection of the tokens posting lists, therefore it should not grow out of proportion when adding tokens to the query. 

## Future work

Phrase matching does not leverage the optimization yet.
Hopefully we get to follow a similar patterns where positions are handled properly.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
@@ -317,7 +317,7 @@ pub trait InvertedIndex {
             .filter_map(map_filter_condition)
     }
 
-    // TODO(phrase-matching)
+    // TODO(phrase-matching) leverage pre-computed intersection for tokens
     fn check_match(
         &self,
         parsed_query: &ParsedQuery,
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
@@ -321,8 +321,42 @@ pub trait InvertedIndex {
         &self,
         parsed_query: &ParsedQuery,
         point_id: PointOffsetType,
+        covered_points: &[AHashSet<PointOffsetType>],
         hw_counter: &HardwareCounterCell,
     ) -> bool;
+
+    fn check_has_subset(
+        &self,
+        point_id: PointOffsetType,
+        covered_points: &[AHashSet<PointOffsetType>],
+    ) -> bool {
+        // check presence of the document
+        if self.values_is_empty(point_id) {
+            return false;
+        }
+
+        // Check that all tokens are in document
+        covered_points
+            .iter()
+            .all(|points_for_token| points_for_token.contains(&point_id))
+    }
+
+    fn query_token_point_ids(
+        &self,
+        parsed_query: &ParsedQuery,
+        hw_counter: &HardwareCounterCell,
+    ) -> Vec<AHashSet<PointOffsetType>> {
+        match parsed_query {
+            ParsedQuery::Tokens(tokens) => self.token_point_ids(&tokens.0, hw_counter),
+            ParsedQuery::Phrase(phrase) => self.token_point_ids(&phrase.0, hw_counter),
+        }
+    }
+
+    fn token_point_ids(
+        &self,
+        token_ids: &[PointOffsetType],
+        hw_counter: &HardwareCounterCell,
+    ) -> Vec<AHashSet<PointOffsetType>>;
 
     fn values_is_empty(&self, point_id: PointOffsetType) -> bool;
 
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
@@ -339,6 +339,12 @@ pub trait InvertedIndex {
         points_for_token.contains(&point_id)
     }
 
+    fn iter_point_ids<'a>(
+        &'a self,
+        token_id: TokenId,
+        hw_counter: &'a HardwareCounterCell,
+    ) -> Option<Box<dyn Iterator<Item = PointOffsetType> + 'a>>;
+
     fn query_token_point_ids(
         &self,
         parsed_query: &ParsedQuery,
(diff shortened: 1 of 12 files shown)


lib/segment/tests/integration/payload_index_test.rs
```diff
@@ -66,6 +66,7 @@ macro_rules! ensure {
 
 const DIM: usize = 5;
 const ATTEMPTS: usize = 20;
+const RNG_SEED: u64 = 42;
```
agourlay: had some failing tests there that I had trouble fixing due to the lack of determinism

lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
```diff
@@ -333,6 +371,48 @@ pub trait InvertedIndex {
     fn get_token_id(&self, token: &str, hw_counter: &HardwareCounterCell) -> Option<TokenId>;
 }
 
+/// Intersect a collection of sorted iterators
+fn intersect_sorted<I, T>(mut iterators: Vec<I>) -> Vec<T>
```
coszio: Could you elaborate why this was preferred over reusing the [intersection iterator](https://github.com/qdrant/qdrant/blob/9d35b84c7ed9a5d8d409649d0e18691f6cc4e374/lib/segment/src/index/field_index/full_text_index/inverted_index/postings_iterator.rs#L28)?

lib/segment/src/index/field_index/full_text_index/text_index.rs
```diff
@@ -197,22 +197,48 @@ impl FullTextIndex {
         }
     }
 
-    pub fn check_match(
+    /// Build the set intersection of point ids from the query tokens
+    pub fn query_tokens_posting_intersection(
```
coszio: That's a mouthful of terms 😅 
```suggestion
    pub fn collect_intersection(
```

lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
agourlay: I did not know about those custom intersection functions.
Let me check if I can apply those :+1:

lib/segment/src/index/field_index/full_text_index/inverted_index/mutable_inverted_index.rs
```diff
@@ -268,4 +270,14 @@ impl InvertedIndex for MutableInvertedIndex {
...
+
+    fn iter_point_ids<'a>(
+        &'a self,
+        token_id: TokenId,
+        _hw_counter: &'a HardwareCounterCell,
+    ) -> Option<Box<dyn Iterator<Item = PointOffsetType> + 'a>> {
+        self.get_tokens(token_id).map(|token_set| {
+            Box::new(token_set.0.clone().into_iter()) as Box<dyn Iterator<Item = PointOffsetType>>
+        })
+    }
```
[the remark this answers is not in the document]
generall: Agree with AI here. token_id is used as `PointOffsetType`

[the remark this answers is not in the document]
agourlay: Excellent catch :+1: 

I will double check why no tests failed in that case.

lib/segment/src/index/field_index/full_text_index/inverted_index/mod.rs
agourlay: @coszio I was able to leverage a lot more existing utilities through the `filter` function on the inverted index.
Please check again :)

coszio (approved):

generall (changes requested):
Benchmarks are controversial

lib/segment/src/index/query_optimization/condition_converter/match_converter.rs
```diff
@@ -267,8 +268,13 @@ fn get_match_text_checker<const IS_PHRASE: bool>(
...
 
+            // Build the set intersection of point ids from the query tokens
+            let points_for_token = full_text_index
+                .filter_query(parsed_query.clone(), &hw_counter)
+                .collect::<AHashSet<_>>();
+
             Some(Box::new(move |point_id: PointOffsetType| {
-                full_text_index.check_match(&parsed_query, point_id, &hw_counter)
+                points_for_token.contains(&point_id)
             }))
```
timvisee, line 278: Naive question: if we can do this here, can't we just do the same inside `full_text_index.check_match(..)`?

timvisee (commented):
In your comment you mention we lazily compute a list of points 3 times. Could you point to the exact place where we do this? That is not immediately clear to me from this PR.

agourlay:
The optimization is working fine when a lot of points are being tested in the matcher.

Otherwise, the cost of of preparing the intersection up front could be more expensive than the savings.

To make the matter worse, we actually build the `filter_context` more than once per query.
e.g:
- sampling: https://github.com/qdrant/qdrant/blob/530430fac2a3ca872504f276d2c91a5c91f43fa0/lib/segment/src/index/hnsw_index/hnsw.rs#L1310C21-L1310C35
- per query:: https://github.com/qdrant/qdrant/blob/530430fac2a3ca872504f276d2c91a5c91f43fa0/lib/segment/src/index/hnsw_index/hnsw.rs#L940
- leaf call: https://github.com/qdrant/qdrant/blob/530430fac2a3ca872504f276d2c91a5c91f43fa0/lib/segment/src/index/hnsw_index/hnsw.rs#L890

Creating a single `filter_context` would improve the performance and maybe make the tradeoff of this PR much more of a win.

But making this change is not trivial so I will close the current approach as it will never work as long as `filter_context` are assumed to be cheap objects.

PR #6768, opened 2025-06-27
[WIP] MMR

coszio:
Work in progress to add Maximum Marginal Relevance as a kind of query to join prefetches.

### DIFF
--- a/lib/collection/src/operations/universal_query/shard_query.rs
+++ b/lib/collection/src/operations/universal_query/shard_query.rs
@@ -66,6 +66,15 @@ pub enum SampleInternal {
     Random,
 }
 
+/// Maximum Marginal Relevance configuration
+#[derive(Debug, Clone, PartialEq)]
+pub struct MmrInternal {
+    /// Vector name to use for similarity computation, defaults to empty string (default vector)
+    pub vector_name: VectorNameBuf,
+    /// Lambda parameter controlling diversity vs relevance trade-off (0.0 = full diversity, 1.0 = full relevance)
+    pub lambda: f32,
+}
+
 /// Same as `Query`, but with the resolved vector references.
 #[derive(Debug, Clone, PartialEq)]
 pub enum ScoringQuery {
--- a/lib/collection/src/shards/local_shard/query.rs
+++ b/lib/collection/src/shards/local_shard/query.rs
@@ -4,12 +4,24 @@ use std::time::Duration;
 
 use ahash::AHashSet;
 use common::counter::hardware_accumulator::HwMeasurementAcc;
+use common::counter::hardware_counter::HardwareCounterCell;
+use common::top_k::TopK;
+use common::types::{ScoreType, ScoredPointOffset};
 use futures::FutureExt;
 use futures::future::BoxFuture;
+use itertools::Itertools;
 use parking_lot::Mutex;
+use segment::common::operation_error::OperationResult;
 use segment::common::reciprocal_rank_fusion::rrf_scoring;
 use segment::common::score_fusion::{ScoreFusion, score_fusion};
-use segment::types::{Filter, HasIdCondition, ScoredPoint, WithPayloadInterface, WithVector};
+use segment::data_types::vectors::{QueryVector, VectorInternal};
+use segment::types::{
+    Filter, HasIdCondition, ScoredPoint, VectorNameBuf, WithPayloadInterface, WithVector,
+};
+use segment::vector_storage::dense::volatile_dense_vector_storage::new_volatile_dense_vector_storage;
+use segment::vector_storage::multi_dense::volatile_multi_dense_vector_storage::new_volatile_multi_dense_vector_storage;
+use segment::vector_storage::sparse::volatile_sparse_vector_storage::new_volatile_sparse_vector_storage;
+use segment::vector_storage::{VectorStorage, new_raw_scorer};
 use tokio::runtime::Handle;
 use tokio::time::error::Elapsed;
 
@@ -23,7 +35,7 @@ use crate::operations::universal_query::planned_query::{
     MergePlan, PlannedQuery, RescoreParams, RootPlan, Source,
 };
 use crate::operations::universal_query::shard_query::{
-    FusionInternal, SampleInternal, ScoringQuery, ShardQueryResponse,
+    FusionInternal, MmrInternal, SampleInternal, ScoringQuery, ShardQueryResponse,
 };
 
 pub enum FetchedSource {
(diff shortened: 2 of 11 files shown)


lib/collection/src/collection/mmr.rs
```diff
@@ -0,0 +1,710 @@
...
+    using: VectorNameBuf,
+    search_runtime_handle: &Handle,
+    timeout: Duration,
+    hw_measurement_acc: HwMeasurementAcc,
+) -> CollectionResult<Vec<ScoreType>> {
+    let num_vectors = vectors.len();
+
+    // if we have less than 2 points, we can't build a matrix
+    if num_vectors < 2 {
+        panic!("There should be at least two vectors to calculate max similarities")
```
generall: debug assert & return error?

lib/collection/src/collection/mmr.rs
```diff
@@ -0,0 +1,710 @@
...
+                    typed_multi_dense_vector.dim,
+                    distance,
+                    multivec_config,
+                )
+            }
+        }
+    };
+
+    // Populate storage with vectors
+    let hw_counter = HardwareCounterCell::disposable();
```
generall: I don't think it should be `disposable` here.

coszio: It doesn't actually measure anything, since it is a volatile storage. The one which measures is the raw scorer. But it costs nothing to pass it here, so let's pass it.

coszio:
@generall noticed that there is a mistake in the way mmr is calculated. Drafting while I fix this

lib/collection/src/collection/mmr.rs
```diff
@@ -0,0 +1,800 @@
...
+
+        CollectionResult::Ok((query_similarities, similarity_matrix))
+    };
+
+    let (query_similarities, similarity_matrix) = tokio::time::timeout(
+        timeout,
+        search_runtime_handle.spawn_blocking(compute_similarities),
+    )
+    .await
+    .map_err(|_| CollectionError::timeout(timeout.as_secs() as usize, "mmr"))???;
```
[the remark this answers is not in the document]
coszio: weirdly enough, this is valid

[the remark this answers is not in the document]
coszio: One is the wrapped function result, then the result from joining the handle,  then the result from timeout

lib/segment/src/vector_storage/query_scorer/sparse_metric_query_scorer.rs
```diff
@@ -0,0 +1,91 @@
...
+impl QueryScorer for SparseMetricQueryScorer<'_> {
+    type TVector = SparseVector;
+
+    #[inline]
+    fn score_stored(&self, idx: PointOffsetType) -> ScoreType {
+        let stored = self
+            .vector_storage
+            .get_sparse(idx)
+            .expect("Sparse vector not found");
+        self.hardware_counter.vector_io_read().incr();
```
JojiiOfficial: Do we just read 1 byte for a vector? If yes, could you maybe add a comment why this is the case here?

lib/segment/src/vector_storage/query_scorer/sparse_metric_query_scorer.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    #[inline]
+    fn score(&self, v2: &SparseVector) -> ScoreType {
+        self.score_ref(v2)
+    }
+
+    fn score_stored_batch(&self, ids: &[PointOffsetType], scores: &mut [ScoreType]) {
+        debug_assert_eq!(ids.len(), scores.len());
+
+        self.hardware_counter.vector_io_read().incr_delta(ids.len());
```
JojiiOfficial: Same thing here, this means one vector ID is one byte

lib/segment/src/vector_storage/query_scorer/sparse_metric_query_scorer.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn score_internal(&self, point_a: PointOffsetType, point_b: PointOffsetType) -> ScoreType {
+        let v1 = self
+            .vector_storage
+            .get_sparse(point_a)
+            .expect("Sparse vector not found");
+        let v2 = self
+            .vector_storage
+            .get_sparse(point_b)
+            .expect("Sparse vector not found");
+
```
JojiiOfficial, line 89: VectorIO hw measurement seems missing here. `self.score_sparse` only accounts for CPU.

lib/collection/src/collection/mmr.rs
```diff
@@ -0,0 +1,800 @@
...
+                    typed_multi_dense_vector.dim,
+                    distance,
+                    multivec_config,
+                )
+            }
+        }
+    };
+
+    // Populate storage with vectors
+    for (key, vector) in (0..).zip(vectors) {
```
JojiiOfficial, line 151:
```suggestion
    for (key, vector) in vectors.into_iter().enumerate() {
```

lib/collection/src/collection/mmr.rs
```diff
@@ -0,0 +1,800 @@
...
+    lambda: f32,
+    limit: usize,
+) -> Vec<ScoredPoint> {
+    let num_candidates = candidates.len();
+    if num_candidates == 0 || limit == 0 {
+        return Vec::new();
+    }
+
+    let mut selected_indices = Vec::with_capacity(limit);
+    let mut remaining_indices: AHashSet<usize> = (0..num_candidates).collect();
```
JojiiOfficial: Alternatively we could use `IndexSet` here since it's better optimized for iterating - which we seem to do a lot.

lib/collection/src/collection/mmr.rs
coszio: I like doing this range zipping because it provides the enumeration in the desired type (`u32` in this case) so that we don't need to do `as u32` where it is used 😄

lib/segment/src/vector_storage/query_scorer/sparse_metric_query_scorer.rs
coszio: Thanks for pointing this out. This lead to the decision of not measuring `vector_io_read` in this scorer since it directly works with `VolatileSparseVectorStorage` which is always in memory.

I'll add a comment on the constructor why we are not considering it, contrary to other scorers

lib/collection/src/collection/mmr.rs
JojiiOfficial: I find `enumerate()` easier to read but don't have a strong preference, so fine by me 👍🏻

lib/collection/src/collection/mmr.rs
coszio: https://github.com/qdrant/qdrant/pull/6768/commits/662b82649b40f5e42fcfd36b498804fdb7a92591

JojiiOfficial (approved):

PR #6770, opened 2025-06-27
Implement stemming

JojiiOfficial:
Add stemming support for all tokenizers.

Configuring the API now looks like this:

```
PUT collections/benchmark/index
{
  "field_name": "t",
  "field_schema": {
    "type": "text",
    "stemmer": {
      "type": "snowball",
      "language": "en"
    }
  }
}

```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5264,8 +5264,8 @@ dependencies = [
 
 [[package]]
 name = "rust-stemmers"
-version = "1.2.0"
-source = "git+https://github.com/qdrant/rust-stemmers.git?branch=master#9b118a2b05d88fc96a54c8e0b1f7062a39964e5e"
+version = "1.2.1"
+source = "git+https://github.com/qdrant/rust-stemmers.git?tag=v1.2.1#aee4c73b4012230b1163bf82d086cbf4b3f1102e"
 dependencies = [
  "serde",
  "serde_derive",
--- a/lib/segment/Cargo.toml
+++ b/lib/segment/Cargo.toml
@@ -96,7 +96,7 @@ tap = { workspace = true }
 zerocopy = { workspace = true }
 lazy_static = "1.5.0"
 vaporetto = { version = "0.6.5" }
-rust-stemmers = { git = "https://github.com/qdrant/rust-stemmers.git", branch = "master" }
+rust-stemmers = { git = "https://github.com/qdrant/rust-stemmers.git", tag = "v1.2.1" }
 sysinfo = "0.35"
 charabia = { version = "0.9.3", default-features = false, features = [
     "greek",


lib/api/src/grpc/proto/collections.proto
```diff
@@ -484,10 +480,21 @@ message TextIndexParams {
...
+
+message StemmingAlgorithm {
+    oneof stemming_params{
+        SnowballStemmingParams snowball_params = 1; // Parameters for snowball stemming
+    }
+}
+
+message SnowballStemmingParams{
+    string language = 1; // Which language the algorithm should stem.
 }
```
coszio: nit: why is this indented differently than all the other indent changes? 🙈

lib/segment/src/index/field_index/full_text_index/tokenizers/mod.rs
```diff
@@ -252,43 +253,61 @@ impl Tokenizer {
...
+pub struct TokenizerConfig {
+    pub(crate) lowercase: bool,
+    pub(crate) stopwords_filter: StopwordsFilter,
+    pub(crate) stemmer: Option<Stemmer>,
+}
+
+impl TokenizerConfig {
+    /// Applies stemming if enabled and applies the configured stemming algorithm. Does nothing if
+    /// stemming is disabled.
+    pub fn stem_if_applicable<'a>(&self, input: Cow<'a, str>) -> Cow<'a, str> {
```
coszio: nit: to avoid ambiguity with applicability to the word
```suggestion
    pub fn stem_if_enabled<'a>(&self, input: Cow<'a, str>) -> Cow<'a, str> {
```

lib/segment/src/index/field_index/full_text_index/tokenizers/mod.rs
```diff
@@ -69,25 +70,26 @@ struct PrefixTokenizer;
...
                 } else {
                     Cow::Borrowed(word)
                 };
 
-                if stopwords_filter.is_stopword(&word_cow) {
+                if config.stopwords_filter.is_stopword(&word_cow) {
                     return;
                 }
 
+                let word_cow = config.stem_if_applicable(word_cow);
```
coszio: Prefix tokenizer behaves funnily with stemmer.

An example (using https://snowballstem.org/demo.html#English):
| Input | stemmed |
| - | - |
| `format` | `format` |
| `formati` | `formati` |
| `formatio` | `formatio` |
| `formation` |  `format` |

When typing the `i` and `o`, there will be no results, since the document got indexed as the stemmed version. But then results will re-appear when `n` is typed.

 I guess it is a decision of the user, but we should consider documenting this at least.

coszio: The alternative would be to stem after getting `ngrams`

lib/segment/src/index/field_index/full_text_index/tokenizers/multilingual.rs
```diff
@@ -20,95 +16,46 @@ const DEFAULT_NORMALIZER: NormalizerOption = NormalizerOption {
...
-        // TODO(multilingual): Replace this with a value from `config`!
-        let stem = false;
+        let lowercase = config.lowercase;
+        let stopwords_filter = &config.stopwords_filter;
 
         // If the script of the input is latin and we don't need to stem early, tokenize as-is.
         // This skips language detection, reduces overhead, and improves performance.
-        if script_is_latin(script) && !stem {
-            Self::tokenize_charabia(input, stopwords_filter, lowercase, cb);
+        if script_is_latin(script) && config.stemmer.is_none() {
```
coszio: Looks like we don't need to check if there's no stemmer here. Or am I missing something?

generall:
> ```
> PUT collections/benchmark/index
> {
>   "field_name": "t",
>   "field_schema": {
>     "type": "text",
>     "normalize": true,
>     "stemmer": {
>       "type": "snowball",
>       "language": "en"
>     }
>   }
> }
> ```

What is `"normalize": true,`?

lib/segment/src/data_types/index.rs
```diff
@@ -210,6 +210,101 @@ pub struct TextIndexParams {
...
     pub on_disk: Option<bool>,
+
+    /// Algorithm for stemming. Default: disabled.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub stemmer: Option<StemmingAlgorithm>,
+}
+
+/// Different stemming algorithms with their configs.
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+#[serde(tag = "type", rename_all = "snake_case")]
```
generall: Using serde tag will generate anonymous object, which is very hard to use in python client. Instead, we use e.g. `pub r#type: FloatIndexType,` and untagged enum

lib/segment/src/index/field_index/full_text_index/tokenizers/mod.rs
generall: Using stemmer with prefix tokenizer is already strange enough, so I don't think we should come up with any custom logic to prevent "strange" cases. I would rather prefer to have behavior simple

lib/segment/src/index/field_index/full_text_index/tokenizers/multilingual.rs
generall: +1, I also don't see a reason

lib/segment/src/index/field_index/full_text_index/tokenizers/mod.rs
JojiiOfficial: I assumed the input of `tokenize` would always be proper words, so in my eyes, stemming before breaking into ngrams made sense, even in prefixtokenizer.
Apparently this is not the case, so I assume removing stemming support in prefixtokenizer is the best option here, right?

JojiiOfficial:
> What is `"normalize": true,`?

It's leftovers from the previous approach, that I accidentally copied into here. This is not part of this PR and can be ignored.

lib/segment/src/index/field_index/full_text_index/tokenizers/mod.rs
generall: my opinion is to keep current implementation. It might be sub-optimal in some specific scenarios, but we are not forcing users to enable stemmer.

lib/api/src/grpc/proto/collections.proto
JojiiOfficial: Apparently formatter didn't run on save. Fixed now!

lib/segment/src/index/field_index/full_text_index/tokenizers/multilingual.rs
JojiiOfficial: Yeah was part of the previous language auto-detection approach. Fixed now 👍🏻

PR #6771, opened 2025-06-27
Log when aborting resharding

KShivendu:
We found a bug in debug chaos testing cluster where all the replicas go dead when a pod is restarted. This means none of the replicas can be queried or recovered.

My hypothesis is that it's happening because now we abort resharding whenever a node with resharding replicas is dead. This cleans up collection's `resharding_state`. And later on, it receives Resharding finish operation like `Finish(ShardTransfer { shard_id: 3, to_shard_id: Some(2), from: 4905613015804057, to: 8026666460338483, sync: true, method: Some(ReshardingStreamRecords) })` by that time, we have ongoing resharding shard transfers but `resharding_state` was already dropped from the collection. And we seem to be [marking the replica as Dead](https://github.com/qdrant/qdrant/blob/bfc570645025e8e3ceef2fee5fe6e7cb0deabce1/lib/collection/src/collection/shard_transfer.rs#L241-L255) in this case. All the nodes receive this operation and hence all nodes with shard 2 will end up marking it as Dead - including the last active replica.

But while I reproduce this locally and create a PR to fix it, I'd like to have a log line that clearly tells us that we are aborting resharding. Because right now this happens silently. 

### All Submissions:

### DIFF
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -177,7 +177,9 @@ impl Collection {
         resharding_key: ReshardKey,
         force: bool,
     ) -> CollectionResult<()> {
-        log::warn!("Invalidating local cleanup tasks and aborting resharding {resharding_key} (force: {force})");
+        log::warn!(
+            "Invalidating local cleanup tasks and aborting resharding {resharding_key} (force: {force})"
+        );
 
         let mut shard_holder = self.shards_holder.write().await;
 


timvisee (approved):
> This cleans up collection's resharding_state. And later on, it receives Resharding finish operation

Shouldn't the clean up also abort the transfers, so that we won't be getting a 'finished' operation after?

Either way, happy to merge this if it helps :+1:

KShivendu:
> Shouldn't the clean up also abort the transfers, so that we won't be getting a 'finished' operation after?

Hmm, I don't see the logic that does it. Do you?

It should be part of [this](https://github.com/qdrant/qdrant/blob/bfc570645025e8e3ceef2fee5fe6e7cb0deabce1/lib/collection/src/collection/resharding.rs#L175) function or its nested calls afaik since we [call it](https://github.com/qdrant/qdrant/blob/bfc570645025e8e3ceef2fee5fe6e7cb0deabce1/lib/collection/src/collection/mod.rs#L468-L475) when a resharding replica goes Dead.

### Update: 

I found that we abort transfers [afterwards](https://github.com/qdrant/qdrant/blob/bfc570645025e8e3ceef2fee5fe6e7cb0deabce1/lib/collection/src/collection/mod.rs#L480) but only for the pod that was killed (n0).  However the (problematic) future finish transfer operations are for n1 -> n2 resharding points migrate transfer. So those resharding related transfers aren't aborted. 

This could be a buggy behaviour. It makes sense to abort all resharding related transfers not only the ones related to the node that gets killed.

PR #6785, opened 2025-07-01
[text index] don't optimize when loading a mmap-backed immutable index

coszio:
In `dev`, loading a mmap-backed immutable index involved optimizing in the same way as when mutable turns immutable. This removed posting lists which have no items, and tries to reduce vocabulary size.

However, in the case of mmap->immutable, I don't think this is critical, and requires practically rebuilding the index on load. 

This PR, makes it possible to just turn the mmap structures into in-memory immutable ones by copying the content.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/immutable_text_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/immutable_text_index.rs
@@ -125,8 +125,6 @@ impl ImmutableFullTextIndex {
         };
         self.inverted_index = ImmutableInvertedIndex::from(&index.inverted_index);
 
-        index.inverted_index.clear_cache()?;
-
         // Index is now loaded into memory, clear cache of backing mmap storage
         if let Err(err) = index.clear_cache() {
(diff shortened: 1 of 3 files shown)


generall (approved):

generall:
After loading from mmap into ram we should clear mmap cache

lib/segment/src/index/field_index/full_text_index/immutable_text_index.rs
```diff
@@ -125,6 +125,8 @@ impl ImmutableFullTextIndex {
         };
         self.inverted_index = ImmutableInvertedIndex::from(&index.inverted_index);
 
+        index.inverted_index.clear_cache()?;
+
         // Index is now loaded into memory, clear cache of backing mmap storage
         if let Err(err) = index.clear_cache() {
             log::warn!("Failed to clear mmap cache of ram mmap full text index: {err}");
```
coszio: This was already done the line below 😄

generall: :man_facepalming:

PR #6786, opened 2025-07-01
[MMR] resolve at collection level

coszio:
Builds on top of #6768 

- Adds MMR as a scoring query in `ShardQueryRequest` and `CollectionQueryRequest`. 
- Plans the query to include the MMR vector if it is MMR is the top query, so that it gets resolved at collection level without re-fetching
- Handles MMR at collection level
  - MMR vector gets removed afterwards if it wasn't requested

MMR is still not exposed at this point.


lib/collection/src/operations/universal_query/shard_query.rs
```diff
@@ -575,6 +584,16 @@ impl ScoringQuery {
                     Status::invalid_argument(format!("failed to parse formula: {e}"))
                 })?,
             ),
+            grpc::query_shard_points::query::Score::Mmr(grpc::MmrInternal { vector, lambda }) => {
+                let vector =
+                    vector.ok_or_else(|| Status::invalid_argument("missing field: mmr.vector"))?;
+                let vector = VectorInternal::try_from(vector)?;
+                ScoringQuery::Mmr(MmrInternal {
+                    vector,
+                    using: using.unwrap_or(DEFAULT_VECTOR_NAME.to_string()),
```
JojiiOfficial: nit: Let's use `unwrap_or_else(|| .. )` here to avoid allocating the `DEFAULT_VECTOR_NAME`  string here.

generall (approved):

PR #6789, opened 2025-07-02
Optimize payload filtering to avoid redundant matching

agourlay:
Detects when the primary clause covers all necessary ids to avoid performing an additional matching.

This speeds up all payload filter requests with a single condition that is being used as a primary clause.

In a future PR I want to investigate a more generalized approach that works when there is more than one condition.

### DIFF
--- a/lib/segment/src/index/struct_payload_index.rs
+++ b/lib/segment/src/index/struct_payload_index.rs
@@ -458,8 +458,9 @@ impl StructPayloadIndex {
                 Some(move |id| struct_filtered_context.check(id))
             };
 
-            // add a full scan from id_tracker to compensate unhandled conditions
+            // If even one iterator is None, we should replace the whole thing with an iterator over all ids.
             if require_full_scan_fallback {
+                primary_iters.clear();
                 let full_scan = Box::new(id_tracker.iter_ids().measure_hw_with_cell(
                     hw_counter,
                     size_of::<PointOffsetType>(),
--- a/lib/segment/src/index/struct_payload_index.rs
+++ b/lib/segment/src/index/struct_payload_index.rs
@@ -435,54 +435,49 @@ impl StructPayloadIndex {
             // CPU-optimized strategy here: points are made unique before applying other filters.
             let mut visited_list = self.visited_pool.get(id_tracker.total_point_count());
 
-            let struct_filtered_check =
-                if Self::primary_clause_covers_filter(filter, query_cardinality) {
-                    // the primary clause will select the correct point ids right away
-                    None
-                } else {
-                    // there are other conditions to consider
-                    let struct_filtered_context = self.struct_filtered_context(filter, hw_counter);
-                    Some(move |id| struct_filtered_context.check(id))
-                };
-
-            let iter = query_cardinality
+            let mut primary_iters: Vec<_> = query_cardinality
                 .primary_clauses
                 .iter()
-                .flat_map(move |clause| {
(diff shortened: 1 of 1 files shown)


lib/segment/src/index/struct_payload_index.rs
```diff
@@ -437,6 +435,16 @@ impl StructPayloadIndex {
...
             let mut visited_list = self.visited_pool.get(id_tracker.total_point_count());
 
+            let struct_filtered_check: Box<dyn Fn(PointOffsetType) -> bool> =
+                if Self::primary_clause_covers_filter(filter, query_cardinality) {
+                    // the primary clause will select the correct point ids right away
+                    Box::new(|_id| true)
+                } else {
+                    // there are other conditions to consider
+                    let struct_filtered_context = self.struct_filtered_context(filter, hw_counter);
+                    Box::new(move |id| struct_filtered_context.check(id))
```
generall: Might it be possible to use something like `Either` to avoid box?

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -451,13 +459,28 @@ impl StructPayloadIndex {
...
+        filter: &'a Filter,
+        query_cardinality: &'a CardinalityEstimation,
+    ) -> bool {
+        let one_condition = filter.total_conditions_count() == 1;
+        let one_primary_clause = query_cardinality.primary_clauses.len() == 1;
+        // the payload index does not know about vector names
+        let no_vector_storage_needed = match query_cardinality.primary_clauses[0] {
+            PrimaryCondition::Condition(_) | PrimaryCondition::Ids(_) => true,
+            PrimaryCondition::HasVector(_) => false,
+        };
```
generall: Not sure I understand why those conditions are different. If `HasVector` can be primary (can it?) why we can't use it?

agourlay: `StructPayloadIndex::query_field` does not handle `PrimaryCondition::HasVector(_)` when it is a primary condition as it can't use a field index for it.

Therefore the caller falls back to iterating over all ids instead.

https://github.com/qdrant/qdrant/blob/5c54f437ab76aa608425af12228bc66825538687/lib/segment/src/index/struct_payload_index.rs#L444-L451

Therefore we need to apply the matchers.

generall: I see, but I have concerns that this check introduces second source of truth, which is not immediately obvious how to work with. I am going to try to re-organise it a little, but if it won't help - will merge as-is.

generall: For instance, `query_field` function doesn't guarantee that ` PrimaryCondition::Condition` will have an iterator.

agourlay: Good observation about the source of truth :heavy_plus_sign: 

Then we have to call `query_field` first and use the resulting option as the signal for triggering the optimization I believe.

agourlay: I have restructured it in https://github.com/qdrant/qdrant/pull/6789/commits/a605b74b4dad5e135344804205f06875623f9cdd to use properly the results from `query_field`.

As a side effect, it also generates a single full scan iterator when a `query_field` can't do its job instead of potentially one per primary clause.

agourlay:
we are going with the alternative implementation in https://github.com/qdrant/qdrant/pull/6794 because it handles all the different cases explicitly

PR #6791, opened 2025-07-02
Debug coverage

KShivendu:
A bit frustrating but looks like https://github.com/qdrant/qdrant/pull/6538 didn't solve the problem

### All Submissions:


timvisee (commented):
Let's test this in a different (temporary) repo, where we have a short feedback loop. Because I'm not really confident we won't be having more PRs like this in the future.

PR #6794, opened 2025-07-02
Optimize detect when primary condition is sufficient review

generall:
An alternative implementation to https://github.com/qdrant/qdrant/pull/6789

### DIFF
--- a/lib/segment/src/index/struct_payload_index.rs
+++ b/lib/segment/src/index/struct_payload_index.rs
@@ -7,8 +7,8 @@ use ahash::AHashSet;
 use atomic_refcell::AtomicRefCell;
 use common::counter::hardware_counter::HardwareCounterCell;
 use common::counter::iterator_hw_measurement::HwMeasurementIteratorExt;
+use common::either_variant::EitherVariant;
 use common::types::PointOffsetType;
-use itertools::Either;
 use log::debug;
 use schemars::_serde_json::Value;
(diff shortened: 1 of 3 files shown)


agourlay:
I like it except of the new Either variant abstraction.
Trying to tweak mine with the new ideas I got there.

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -423,38 +423,61 @@ impl StructPayloadIndex {
...
+                .map(move |clause| self.query_field(clause, hw_counter))
+                .collect();
+
+            if let Some(primary_iterators) = primary_clause_iterators {
+                let num_primary_iterators = primary_iterators.len();
+                let joined_primary_iterator = primary_iterators.into_iter().flatten();
+
+                return if num_primary_iterators == filter.total_conditions_count() {
+                    // All conditions are primary clauses,
+                    // We can avoid post-filtering
```
agourlay, line 452: I am not sure this is correct.

We need the intersection of the iterators not the union.

The `struct_filtered_context.check` is still required to ensure that a given point id matches all conditions.

generall: Multiple `primary_iterators` can only be generated with multiple `should` conditions. So it actually it is a union

agourlay (approved):

PR #6798, opened 2025-07-03
more io budget

generall:
- **try to run more IO jobs**


lib/collection/src/update_handler.rs
```diff
@@ -636,7 +636,7 @@ impl UpdateHandler {
             // Otherwise skip now and start a task to trigger the optimizer again once resource
             // budget becomes available
             let desired_cpus = 0;
-            let desired_io = num_rayon_threads(max_indexing_threads);
+            let desired_io = (num_rayon_threads(max_indexing_threads) / 2).max(1); // At least one IO unit
```
timvisee, line 639: Happy to try it, but...

doesn't this come down to tuning which is highly dependent on the setup? For this reason we expose the `optimizer_{cpu,io}_budget` config properties. Maybe we can double the IO value in our bench to see how it behaves.

generall: this PR for test only

PR #6800, opened 2025-07-03
Fix bug that causes all replicas to die if node is restarted during resharding

KShivendu:
Fixing the bug found in chaos testing debug cluster (reproduced in a test [here](https://github.com/qdrant/cluster-manager/pull/345)). It led to a scenario where all the replicas of a shard get marked as Dead on all nodes and they can't recover from each other since every replica is Dead. The root cause was  exactly what I hypothesised [here](https://github.com/qdrant/qdrant/pull/6771#issuecomment-3013167593). 

Put simply if you:
- Have resharding shard streaming transfer between: n1 shard 3 -> n2 shard 2
- Then if node n0 dies and also had an `Resharding/ReshardingScaleDown` replica (say shard 1), we abort resharding and abort any streaming resharding transfers related to n0 (but we don't abort the above mentioned n1:3 -> n2:2 transfer)
- When the n1:3 -> n2:2 is finished, every node sees that resharding isn't ongoing anymore and hence each node marks their Shard 2 as Dead
- Optional: Even if n0 now comes back online, it will also redo the same operation and also mark it's shard 2 as Dead (if it exists locally)

This PR adds a simple fix that after ensuring basic checks (like the cluster hasn't gone beyond `ReadHashRingCommitted` resharding stage), it will abort any resharding transfer (`method == ReshardingStreamRecords`) ongoing in the cluster.

### All Submissions:

### DIFF
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -216,12 +216,9 @@ impl Collection {
 
         // Abort all resharding transfer related to this specific resharding operation
         let resharding_transfers =
-            shard_holder.get_transfers(|t| t.is_specific_resharding(&resharding_key));
+            shard_holder.get_transfers(|t| t.is_related_to_resharding(&resharding_key));
         for transfer in resharding_transfers {
-            let _is_resharding_transfer = self
-                .abort_shard_transfer_and_report_resharding(transfer.key(), &shard_holder)
-                .await?;
-            // We don't need to abort resharding again since we are already aborting it
+            self.abort_shard_transfer(transfer, &shard_holder).await?;
         }
 
         drop(shard_holder); // drop the read lock before acquiring write lock
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -251,8 +251,10 @@ impl Collection {
         let resharding_transfers = shard_holder
             .get_transfers(|t| t.method == Some(ShardTransferMethod::ReshardingStreamRecords));
         for transfer in resharding_transfers {
-            self.abort_shard_transfer(transfer.key(), Some(&shard_holder))
+            let _is_resharding_transfer = self
+                .abort_shard_transfer_and_report_resharding(transfer.key(), &shard_holder)
                 .await?;
+            // We don't need to abort resharding again since we are already aborting it in next step
         }
 
         shard_holder
--- a/lib/collection/src/collection/resharding.rs
+++ b/lib/collection/src/collection/resharding.rs
@@ -6,10 +6,10 @@ use super::Collection;
 use crate::config::ShardingMethod;
 use crate::hash_ring::HashRingRouter;
 use crate::operations::cluster_ops::ReshardingDirection;
-use crate::operations::types::CollectionResult;
+use crate::operations::types::{CollectionError, CollectionResult};
 use crate::shards::replica_set::ReplicaState;
-use crate::shards::resharding::{ReshardKey, ReshardState};
-use crate::shards::transfer::ShardTransferConsensus;
+use crate::shards::resharding::{ReshardKey, ReshardStage, ReshardState};
+use crate::shards::transfer::{ShardTransferConsensus, ShardTransferMethod};
 
 impl Collection {
     pub async fn resharding_state(&self) -> Option<ReshardState> {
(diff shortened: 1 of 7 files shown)


KShivendu:
> When the n1:3 -> n2:2 is finished, every node sees that resharding isn't ongoing anymore and hence each node marks their Shard 2 as Dead

This also seems like a problematic logic, other nodes shouldn't mark their own local shard 2 replicas as `Dead` when the transfer was only finished for `n2:2`. I'll cross check once.

lib/collection/src/collection/resharding.rs
```diff
@@ -214,8 +214,49 @@ impl Collection {
...
+
+            None => {
+                log::warn!(
+                    "aborting resharding {resharding_key}, \
+                     but resharding is not in progress"
+                );
+
+                false
+            }
+        };
```
timvisee: This is already covered by `check_abort_resharding` above. It may not be exactly the same, but I don't see a significant thing that adds value.

lib/collection/src/collection/resharding.rs
```diff
@@ -214,8 +214,49 @@ impl Collection {
...
+
+        // We need to abort all resharding transfers
+        let resharding_transfers = shard_holder
+            .get_transfers(|t| t.method == Some(ShardTransferMethod::ReshardingStreamRecords));
+        for transfer in resharding_transfers {
+            self.abort_shard_transfer(transfer.key(), Some(&shard_holder))
+                .await?;
+        }
```
timvisee: No, only the transfers that belong to our resharding key.

I also vote to do it inside the shard holder if possible, not here.

KShivendu: > vote to do it inside the shard holder if possible, not here

Shard holder can't access its collection and call `Collection::abort_shard_transfer` so I had to do this.

I checked if can move `Collection::abort_shard_transfer` to `ShardHolder::abort_shard_transfer` entirely but there were other functions that needed access to Collection, so I wasn't sure if that's a good idea.

timvisee: You're right. Since the transfer abort function is inside the collection already, let's keep it there. Missed that.

lib/collection/src/collection/resharding.rs
KShivendu: Created https://github.com/qdrant/qdrant/pull/6806 to resolve this.

lib/collection/src/collection/resharding.rs
```diff
@@ -214,6 +214,19 @@ impl Collection {
...
 
+        // We need to abort all resharding transfers
+        let resharding_transfers = shard_holder
+            .get_transfers(|t| t.method == Some(ShardTransferMethod::ReshardingStreamRecords));
+        for transfer in resharding_transfers {
+            let _is_resharding_transfer = self
+                .abort_shard_transfer_and_report_resharding(transfer.key(), &shard_holder)
+                .await?;
+            // We don't need to abort resharding again since we are already aborting it in next step
+        }
```
timvisee: I suggest to only abort resharding transfers here that belong to the current resharding operation.

In the future we might support multiple concurrent resharding operations (at least, that is the plan). I'm afraid we'll forget about this then if we don't handle it now.

timvisee: Is it necessary to also 'report to resharding' here? Aren't we already running the abort resharding logic here?

timvisee: https://github.com/qdrant/qdrant/pull/6800/commits/b4ef5ff511498f42a65e55f0c9c1e4539a7a2094

KShivendu: The same function in used [here](https://github.com/qdrant/qdrant/pull/6800/files#diff-a1ea59ccb49f8173215cfbf07d8e752124d6cf00b8fb3c9ce1d37b425de57529R384-R386) where we need this info.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -299,19 +299,13 @@ impl Collection {
...
 
-    /// Handles abort of the transfer
-    ///
-    /// 1. Unregister the transfer
-    /// 2. Stop transfer task
-    /// 3. Unwrap the proxy
-    /// 4. Remove temp shard, or mark it as dead
-    pub async fn abort_shard_transfer(
+    /// Return if it was a resharding transfer so it can be handled correctly (aborted or ignored)
+    pub async fn abort_shard_transfer_and_report_resharding(
```
timvisee: Is this only called `_and_report_resharding` because it returns a boolean defining whether it was a resharding transfer?

KShivendu: Yes.

timvisee (approved):
Looks good, together with <https://github.com/qdrant/qdrant/pull/6881>.

@KShivendu is doing another test on the side. Once it passes we can merge.

KShivendu:
CI passed in our CM test https://github.com/qdrant/cluster-manager/pull/354. Merging this! :partying_face:

PR #6804, opened 2025-07-03
Optimize has_id with ahash

agourlay:
This `has_id` implementation is is a prime candidate for `AHashSet` because it tests the membership in a set of `u32`.

With some naive benchmarking, I get a 10% latency improvements on large distance matrix requests.

e.g.

```http
POST /collections/benchmark/points/search/matrix/pairs
{
  "sample": 500,
  "limit": 1000
}
```


generall (approved):

timvisee:
>  I get a 10% latency improvements on large distance matrix requests.

That is huge! Nice find.

PR #6805, opened 2025-07-03
Enable Gridstore in mutable payload indices, don't eagerly create storage

timvisee:
The main goal of this PR is to enable support for Gridstore in mutable payload indices unconditionally. It doesn't mean that Gridstore is always used, _yet_. It does mean Gridstore _can_ always be used. Along with this, some other necessary adjustments are made.

So, first, this enableds Gridstore on the mutable payload indices. Before, Gridstore was only supported when compiling without RocksDB.

Initialization of mutable payload indices has been adjusted. It prevents Gridstore from creating files on disk, even if we just try to load an existing index. Before, trying to load when no index exists results in loading an empty index instead, breaking our index.

Now we detect an index load failure when Gridstore is used. When no data exists on disk, we fail on load, and trigger index rebuilding. We now follow the same semantics as our other payload index storage backends.

Respect the `payload_index_skip_mutable_rocksdb` runtime feature flag. Setting it to true will now result in Gridstore being used in all new mutable segments. We used `payload_index_skip_rocksdb` for both mutable and immutable segments before, which was incorrect.

I added a boolean `create_if_missing` parameter to payload index opening. When this is false, we prevent creating files on disk if Gridstore is used. I don't think this implementation is great, but it keeps the payload index interface the same. I think this part of the code - that creates/opens indices - would benefit from refactoring. But we better do that after the release, when there's less time pressure.

In future PRs we'll implement proper migrations, and we'll make payload index loading more robust by explicitly persisting what type of index storage is used.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
+++ b/lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
@@ -52,7 +52,7 @@ where
 {
     #[cfg(feature = "rocksdb")]
     RocksDb(DatabaseColumnScheduledDeleteWrapper),
-    Gridstore(Arc<RwLock<Gridstore<Vec<T>>>>),
+    Gridstore(Option<Arc<RwLock<Gridstore<Vec<T>>>>>),
 }
 
 // Numeric Index with insertions and deletions without persistence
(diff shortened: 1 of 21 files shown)


lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
```diff
@@ -285,7 +304,8 @@ where
         match self.storage {
             #[cfg(feature = "rocksdb")]
             Storage::RocksDb(_) => self.load_rocksdb(),
-            Storage::Gridstore(_) => self.load_gridstore(),
+            Storage::Gridstore(Some(_)) => self.load_gridstore(),
```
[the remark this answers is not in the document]
timvisee: I'd argue that using the option here is the simplest option available, which is why I used it.

I don't think that adding another enum variant is the right approach.

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -187,7 +188,12 @@ impl StructPayloadIndex {
...
-                if common::flags::feature_flags().payload_index_skip_rocksdb && !is_appendable {
+                let skip_rocksdb = if is_appendable {
+                    common::flags::feature_flags().payload_index_skip_mutable_rocksdb
+                } else {
+                    common::flags::feature_flags().payload_index_skip_rocksdb
+                };
+                if skip_rocksdb {
                     new_config.skip_rocksdb = Some(true);
                 }
                 new_config
```
coszio, line 199: Context for other reviewers: 
This only applies when the payload indexes config does not exist. So if `payload_index/config.json` already exists without the `skip_rocksdb` flag, this will be what's loaded, and a rocksdb-backed index will be used, even if the index for a field is newly created.

When we migrate the indices, we need to make sure we handle the `skip_rocksdb:null` case when loading

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -782,7 +785,14 @@ pub fn build_segment(
...
 
-    let segment = create_segment(None, None, &segment_path, config, &AtomicBool::new(false))?;
+    let segment = create_segment(
+        None,
+        None,
+        &segment_path,
+        config,
+        &AtomicBool::new(false),
+        true,
+    )?;
```
coszio, line 795: Context # 2: This is the main reason for `create_if_missing`. When loading a segment, we want to fail if the index's backing storage is different. But when building a new segment, we want to create the files so they can be populated from scratch.

@timvisee can you confirm this is correct?

coszio (approved):

lib/segment/src/segment_constructor/segment_constructor_base.rs
timvisee: Correct.

agourlay (approved):
The comments from @coszio were very helpful to understand the diff :clap:

timvisee:
> The comments from @coszio were very helpful to understand the diff 👏

Sorry for this one being a bit all over the place.

PR #6812, opened 2025-07-04
[WIP] Store index type in payload index config

JojiiOfficial:
WIP

Depends on #6805


lib/segment/src/index/payload_config.rs
```diff
@@ -34,3 +34,106 @@ impl PayloadConfig {
...
+    Immutable(StorageType),
+    Mmap { is_on_disk: bool },
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub enum StorageType {
+    Gridstore,
+    RocksDB,
+    Mmap { is_on_disk: bool },
+}
```
timvisee: Note to others: the key types, defining payload index/storage variants, are here.

lib/segment/src/index/field_index/index_selector.rs
```diff
@@ -57,6 +58,98 @@ pub struct IndexSelectorGridstore<'a> {
...
+            (FullPayloadIndexType::IntIndex(_), PayloadSchemaParams::Integer(params)) => {
+                // IntIndex only gets created if `range` is true. This will only throw an error if storage is corrupt.
+                //
+                // Note that `params.range == None` means the index was created without directly specifying these parameters.
+                // In those cases it defaults to `true` so we don't need to cover this case.
+                if params.range == Some(false) {
+                    return Err(OperationError::service_error(
+                        "Inconsistent payload schema: Int index configured but schema.range is false",
+                    ));
+                }
```
timvisee: Let's throw a warning here instead. Also in the other one below.

lib/segment/src/index/field_index/index_selector.rs
```diff
@@ -57,6 +58,98 @@ pub struct IndexSelectorGridstore<'a> {
...
+                };
+                FieldIndex::NullIndex(null_index)
+            }
+
+            // Storage inconsistency. Should never happen.
+            (index_type, schema) => {
+                return Err(OperationError::service_error(format!(
+                    "Payload index storage inconsistent. Schema defines {schema:?} but storage is {index_type:?}"
+                )));
+            }
```
timvisee, line 145: Note: if the index type and schema (both are persisted on disk in the same place) don't match, we can reach this branch. This is a hard error. Since we never expect this to happen under normal circumstances, I think this is fine.

lib/segment/src/index/plain_payload_index.rs
```diff
@@ -86,15 +91,20 @@ impl PayloadIndex for PlainPayloadIndex {
...
+                schema: payload_schema.clone(),
+                index_types: field_index
+                    .iter()
+                    .map(|i| i.get_full_index_type())
+                    .collect(),
+            },
+        ) {
             // the field is already present with the same schema, no need to save the config
-            if prev_schema == payload_schema {
+            if prev_schema.schema == payload_schema {
```
timvisee: Let's also make sure that `prev_schema.index_types` is the same. Only if both are equal, we return without saving.

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -528,6 +553,55 @@ impl StructPayloadIndex {
...
+                            | StorageType::RocksDbNonAppendable(db)) = &self.storage_type
+                            else {
+                                return Err(OperationError::service_error(
+                                    "Loading payload index failed: Configured storage type and payload schema mismatch!",
+                                ));
+                            };
+
+                            return Ok(IndexSelector::RocksDb(IndexSelectorRocksDb {
+                                db,
+                                is_appendable: false,
```
timvisee: true if `RocksDbAppendable`, false if `RocksDbNonAppendable`

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -133,23 +134,48 @@ impl StructPayloadIndex {
...
+
+            // Special null index complements every index.
+            if let Some(null_index) =
+                IndexSelector::new_null_index(&self.path, field, total_point_count)?
+            {
+                // todo: This means that null index will not be built if it is not loaded here.
+                //       Maybe we should set `is_loaded` to false to trigger index building.
+                indexes.push(null_index);
+            }
+
```
timvisee, line 187: Let's persist the chosen index type here. This way we only have to guess once on load. When we restart Qdrant, we will fall into the other branch (below) and don't have to guess again.

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -133,23 +134,48 @@ impl StructPayloadIndex {
...
+            indexes
```
timvisee, line 193: Let's also add one test:
- create and load a segment
- ensure the index type is persisted in the config

We don't have to do this for every type, because there are many of them. One is enough. We can manually test all the other ones.

lib/segment/src/index/payload_config.rs
```diff
@@ -34,3 +34,106 @@ impl PayloadConfig {
...
+    DatetimeIndex(IndexMutability),
+    IntMapIndex(IndexMutability),
+    KeywordIndex(IndexMutability),
+    FloatIndex(IndexMutability),
+    GeoIndex(IndexMutability),
+    FullTextIndex(IndexMutability),
+    BoolIndex(IndexMutability),
+    UuidIndex(IndexMutability),
+    UuidMapIndex(IndexMutability),
+    NullIndex(IndexMutability),
```
timvisee: I think we want to have an 'unsupported' variant here. Just to be a bit more forward compatible. If we introduce a new variant in the future and go one version back, we'd fall into the 'unsupported' variant.

Even though with true new index type we don't really support backwards compatibility, it might still be nicer to handle it in some what rather than crashing on deserialization.

Let's not do it in this PR, but separately. Because this one is already big enough. I might pick it up at the end of the day, and (I hope) the other bits and bobs are merged.

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -122,34 +123,80 @@ impl StructPayloadIndex {
...
             field_indexes.insert(field.clone(), field_index);
         }
+
+        self.config.indexed_fields = indexed_fields;
+
+        // If any payload_schema didn't have an index type assigned, it has now
+        // and therefore we need to store it.
+        if had_index_without_type {
+            self.save_config()?;
+        }
```
timvisee, line 157: If the index has changed in any way, we should save the config. Also if we migrated to a new type, when another type was already in place.

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -528,6 +574,58 @@ impl StructPayloadIndex {
...
+                            let (StorageType::RocksDbAppendable(db)
+                            | StorageType::RocksDbNonAppendable(db)) = &self.storage_type
+                            else {
+                                return Err(OperationError::service_error(
+                                    "Loading payload index failed: Configured storage type and payload schema mismatch!",
+                                ));
+                            };
+
+                            let is_appendable =
+                                matches!(self.storage_type, StorageType::RocksDbAppendable(..));
```
timvisee: Let's add an `is_appendable()` to the storage type.

lib/segment/src/index/struct_payload_index.rs
```diff
@@ -899,3 +1008,96 @@ impl PayloadIndex for StructPayloadIndex {
...
+        payload_config.save(&payload_config_path).unwrap();
+
+        // Load once and drop.
+        {
+            load_segment(&full_segment_path, &AtomicBool::new(false))
+                .unwrap()
+                .unwrap();
+        }
+
+        // Check that index type has been written to disk again.
```
timvisee, line 1098:
```suggestion
        // Check that index type has been written to disk again.
        // Proves we'll always persist the exact index type if it wasn't known yet at that time
```

timvisee (commented):
Thanks for all the adjustments!

I've some minor things, I'll resolve them.

lib/segment/src/index/struct_payload_index.rs
JojiiOfficial: Behavior changed in https://github.com/qdrant/qdrant/compare/store-payload-index-type...migrate-payload-index-naive-fix-mmap-migration-fixed?expand=1

timvisee: Thanks for the pointer. Will check it out in a second. I'm fine with changing it in a later PR (and not in this one).

timvisee:
This currently fails to load the 'Qdrant Web Documentation' snapshot sample from web UI.

It fails at deserialization here:
https://github.com/qdrant/qdrant/blob/1da134fddd044ed4c39500fd01a367f232f8b994/lib/segment/src/index/struct_payload_index.rs#L245

It looks like serde flattening isn't working properly.

Edit: it fails to deserialize the main index schema types: https://github.com/qdrant/qdrant/blob/137d71ada86cd0b8903ed788e1eaa29634c4c208/lib/segment/src/types.rs#L1885

---

The implementation in this PR is incompatible with what we have already released.

When creating a field index (without parameters), the structures are different:

```json
// before
"myfield": "keyword",

// after
"myfield": {
  "keyword": null,
  "index_types": [
    {
      "KeywordIndex": {
        "Mutable": "RocksDB"
      }
    },
    {
      "NullIndex": {
        "Mmap": {
          "is_on_disk": true
        }
      }
    }
  ]
}
```

We must take the new index types [field](https://github.com/qdrant/qdrant/blob/95b8ade4e998a7fc804a98cd7010b2094f97452c/lib/segment/src/index/payload_config.rs#L44) out of the index map, and store it separately at the [root](https://github.com/qdrant/qdrant/blob/95b8ade4e998a7fc804a98cd7010b2094f97452c/lib/segment/src/index/payload_config.rs#L16) level.

@JojiiOfficial we can decide tomorrow on who takes care of this. :+1:

lib/segment/src/index/payload_config.rs
```diff
@@ -34,3 +34,106 @@ impl PayloadConfig {
...
+        }
+    }
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, PartialEq)]
+pub enum IndexMutability {
+    Mutable(StorageType),
+    Immutable(StorageType),
+    Mmap { is_on_disk: bool },
+}
```
coszio: I don't think `Mmap` variant belongs here. Correct me if I am wrong, but it should be possible to represent mmap indices as mutable or immutable with a mmap storage, isn't it?

```suggestion
pub enum IndexMutability {
    /// For appendable segments. Supports appending, updating, and deleting values
    Mutable(StorageType),
    /// For non-appendable segments. Only supports deleting values.
    Immutable(StorageType),
}
```

This should be enough for mutability, I believe. 

Also, I'm 99% sure there are no immutable indices which don't support deletions. Can anyone confirm?

coszio: Following this idea, looks like we need a product type (a struct), rather than a sum type (an enum, or in this case, a nested enum). What about this representation?

```rust
pub enum IndexMutability {
    /// For appendable segments. Supports appending, updating and deleting values
    Mutable,
    /// For non-appendable segments. Only supports deleting values.
    Immutable,
}

pub struct IndexKind {
    mutability: IndexMutability,
    storage_type: StorageType,
}
```

timvisee (changes requested):
[comment](https://github.com/qdrant/qdrant/pull/6812#issuecomment-3057928785)

lib/segment/src/index/payload_config.rs
```diff
@@ -34,3 +34,106 @@ impl PayloadConfig {
...
+
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, PartialEq)]
+pub enum StorageType {
+    Gridstore,
+    RocksDB,
+    Mmap { is_on_disk: bool },
```
coszio, line 188: If this is going to be in a stored config file, we should probably also include a version number field for `Mmap`, in case we want to update/migrate to a new incompatible mmap implementation

coszio (changes requested):
Happy to dismiss if I missed something in [this comment](https://github.com/qdrant/qdrant/pull/6812/files#r2198158394)

lib/segment/src/index/payload_config.rs
timvisee: I agree that using a product type, like you describe, may be more reasonable.

Are you then suggesting that StorageType should hold: `RocksDb`, `Gridstore`, `Mmap` (and any future ones)?

Edit: good comment from @JojiiOfficial: the mmap boolean index is both mutable and immutable. As in, there is no difference between the two. That's why IndexMutability::Mmap{} was chosen.

> Also, I'm 99% sure there are no immutable indices which don't support deletions. Can anyone confirm?

Correct. But that doesn't make them mutable.

lib/segment/src/index/payload_config.rs
timvisee: Not sure if that is strictly necessary.

We _could_ always add a version number later (no version means first version). And/or add an extra `Mmap2` (or better) variant. We also store a config file along with the mmap files, which may be enough to version if needed within each index itself.

While I don't strictly object against a version number - it can be useful after all - I'd like to keep this structure as simple as possible.

Thoughts?

timvisee:
I've tackled the problem explained [here](https://github.com/qdrant/qdrant/pull/6812#issuecomment-3057928785) in <https://github.com/qdrant/qdrant/pull/6852>

Edit: it is now merged into this PR :+1:

timvisee (approved):
[comment](https://github.com/qdrant/qdrant/pull/6812#issuecomment-3057928785)

lib/segment/src/index/field_index/full_text_index/text_index.rs
```diff
@@ -405,6 +406,25 @@ impl FullTextIndex {
             FullTextIndex::Mmap(index) => index.clear_cache(),
         }
     }
+
+    pub fn get_mutability_type(&self) -> IndexMutability {
+        match self {
+            FullTextIndex::Mutable(_) => IndexMutability::Mutable,
+            FullTextIndex::Immutable(_) => IndexMutability::Immutable,
+            // Mmap index can be both mutable and immutable, so we pick mutable
+            FullTextIndex::Mmap(_) => IndexMutability::Mutable,
```
coszio, line 415: For text index, mmap is immutable

PR #6813, opened 2025-07-04
[MMR] Optimize SymmetricMatrix

JojiiOfficial:
Builds on #6797 

Optimizes the diagonal matrix used in MMR to only store half of the data in a single, flattened vector.

### DIFF
--- a/lib/collection/src/collection/mmr.rs
+++ b/lib/collection/src/collection/mmr.rs
@@ -222,14 +222,14 @@ fn similarity_matrix(
     // Compute similarities only for upper triangle to optimize (i < j)
     // Since similarity is symmetric: sim(i,j) == sim(j,i), we can avoid duplicate computation
 
-    let mut scores = vec![0.0; num_vectors - 1];
+    let mut scores_buf = vec![0.0; num_vectors];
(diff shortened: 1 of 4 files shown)


lib/collection/src/collection/mmr.rs
```diff
@@ -213,19 +218,20 @@ fn similarity_matrix(
...
+    let mut scores = vec![0.0; num_vectors - 1];
+
+    for (i, raw_scorer) in raw_scorers.iter().enumerate().take(num_vectors) {
         // Only compute scores for the upper triangle
         let upper_offsets: Vec<u32> = ((i + 1)..num_vectors).map(|j| j as u32).collect();
 
         if !upper_offsets.is_empty() {
-            let mut scores = vec![0.0; upper_offsets.len()];
-            raw_scorers[i].score_points(&upper_offsets, &mut scores);
+            scores.resize(upper_offsets.len(), 0.0);
```
JojiiOfficial: Also added a minor optimization here to reduce allocations and share `scores` across loops.

Note: we don't need to clear it because we always overwrite all items before using them.

PR #6816, opened 2025-07-07
Exclude primary condition from filter to avoid duplicated matching

agourlay:
The following happens when performing a payload filter:

1. compute primary condition based on cardinality estimation
2. query using primary condition (that is the most restrictive condition)
3. run filter matcher on all resulting points to make sure they match all conditions in the filter (not only the primary one)

Effectively we are checking the primary condition twice, because the primary condition is not excluded from the matcher in the filtering phase.

e.g.

```
{
    "query": 2,
    "filter": {
        "must": [
            {
                "key": "t",
                "match": {
                    "text": "word_2 word_10"
                }
            }
        ]
    }
}
```

In that case the primary condition is the full text filter and the secondary condition is a synthetic `must not have id = 2` for the recommendation.

This PR proposes to not rerun the fulltext validation and simply match point based on the remaining `HasId` condition.
Intersecting posting list for each point id is pretty expensive.

Therefore this change yields a nice performance gain for the plain filtered vector search.

### Before

```
Search with HNSW and high cardinality: 2.063 millis
Search with small cardinality: 21.316 millis  <---
Text search as secondary condition with high cardinality: 4.173 millis
Text search as secondary condition with low cardinality: 4.006 millis
```

### After

```
Search with HNSW and high cardinality: 1.985 millis
Search with small cardinality: 11.503 millis  <---
Text search as secondary condition with high cardinality: 4.215 millis
Text search as secondary condition with low cardinality: 3.944 millis
```

## Review notes

The implementation could not modify the initial filter as all inputs are tied to the same lifetime.
The solution adapted here relies on filtering out primary conditions before the filter optimization phase

Also I only applied this transformation to `must` clause as I doing it as well on `should` clauses breaks a test.

### DIFF
--- a/lib/segment/src/index/field_index/mod.rs
+++ b/lib/segment/src/index/field_index/mod.rs
@@ -84,4 +84,17 @@ impl CardinalityEstimation {
     pub const fn equals_min_exp_max(&self, other: &Self) -> bool {
         self.min == other.min && self.exp == other.exp && self.max == other.max
     }
+
+    // TODO input should be a Condition
+    pub fn has_primary_condition(&self, condition: &FieldCondition) -> bool {
+        self.primary_clauses
+            .iter()
+            .any(|primary_condition| match primary_condition {
+                PrimaryCondition::Condition(field_condition) => {
+                    field_condition.as_ref() == condition
+                }
+                PrimaryCondition::Ids(_) => false,
+                PrimaryCondition::HasVector(_) => false,
+            })
+    }
 }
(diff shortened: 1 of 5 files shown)


generall (changes requested):
Unfortunately, I don't agree that this approach is correct.

Having something as a primary condition a) doesn't make this condition always true - there might be multiple conditions b) is not equivalent for removing the conditions

Also, same condition can be used twice, which causes incorrect evals. See example:

```
{
  "query": 1,
  "with_payload": true,
  "filter": {
    "must": [
      {
        "key": "a",
        "match": {
          "value": "keyword_3"
        }
      }
    ],
    "must_not": [
      {
        "key": "a",
        "match": {
          "value": "keyword_3"
        }
      }
    ]
  }
}
```

agourlay:
Thanks for the review, I will revisit this approach when time permits :+1:

PR #6817, opened 2025-07-07
Wait until replicas are activated, when creating custom shards

ffuugoo:
Follow-up to #6778.

After #6778, we initialize custom shard replicas in `Initializing` state, and so when Qdrant returns response to the client, it does not yet guarantee that custom shard is ready to accept write requests. There will be a small delay while shards switch from `Initializing` into `Active` state and fully available.

This PR adds explicit `wait_for_state` check, so that we ensure all replicas are `Active` before returning response to the client. I've also very slightly tweaked `test_shard_consistency` test, so that it should cover this check as part of existing test.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -549,8 +549,9 @@ impl ShardReplicaSet {
             if status {
                 Ok(())
             } else {
-                Err(CollectionError::service_error(
-                    "Failed to wait for replica set state, timed out",
+                Err(CollectionError::timeout(
+                    timeout.as_secs() as usize,
+                    "wait for replica set state",
                 ))
             }
         }
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -380,6 +380,10 @@ impl ShardReplicaSet {
         replica_set
     }
 
+    pub fn shard_key(&self) -> Option<&ShardKey> {
+        self.shard_key.as_ref()
+    }
+
     pub fn this_peer_id(&self) -> PeerId {
         self.replica_state.read().this_peer_id
     }
@@ -503,17 +507,16 @@ impl ShardReplicaSet {
     /// # Cancel safety
     ///
     /// This method is cancel safe.
-    pub async fn wait_for_state(
+    pub fn wait_for_state(
         &self,
         peer_id: PeerId,
         state: ReplicaState,
         timeout: Duration,
-    ) -> CollectionResult<()> {
+    ) -> impl Future<Output = CollectionResult<()>> + 'static {
         self.wait_for(
             move |replica_set_state| replica_set_state.get_peer_state(peer_id) == Some(state),
             timeout,
         )
-        .await
     }
 
     /// Wait for a replica set state condition to be true.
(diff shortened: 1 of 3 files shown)


lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -523,29 +526,34 @@ impl ShardReplicaSet {
     /// # Cancel safety
     ///
     /// This method is cancel safe.
-    async fn wait_for<F>(&self, check: F, timeout: Duration) -> CollectionResult<()>
+    pub fn wait_for<F>(
+        &self,
+        check: F,
+        timeout: Duration,
+    ) -> impl Future<Output = CollectionResult<()>> + 'static
```
generall, line 533: What was a reason for this change?

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -523,29 +526,34 @@ impl ShardReplicaSet {
...
+            let status = task.await.map_err(|err| {
+                CollectionError::service_error(format!(
+                    "Failed to wait for replica set state: {err}"
+                ))
+            })?;
+
+            if status {
+                Ok(())
+            } else {
+                Err(CollectionError::service_error(
```
generall: I checked usages, it should be safe to convert this error to timeout. Both Service Error and timeout are transient, also it looks like all actual places where it is used either don't care or force convert error into service error.

lib/storage/src/dispatcher.rs
```diff
@@ -226,23 +261,69 @@ impl Dispatcher {
...
+                .get_collection(&CollectionMultipass.issue_pass(&collection_name))
+                .await?
+                .shards_holder()
+                .read_owned()
+                .await;
+
+            for replica_set in shard_holder.all_shards() {
+                if replica_set.shard_key() != Some(&shard_key) {
+                    continue;
+                }
```
timvisee, line 306: Shall we create a getter in the shard holder to which we pass the key? That way we keep the shard key selection logic inside the shard holder.

We have some methods related to shard keys there, to get the shard IDs related to a shard key for example. But for this we'd probably require a new one.

lib/collection/src/shards/replica_set/mod.rs
ffuugoo: **Without** this change, `wait_for` was borrowing/lifetime-bound to `ReplicaSet`, so you *had to* hold a lock to `ReplicaSet` while await-ing `wait_for`.

**With** this change, `wait_for` creates a future that is completely independent of `ReplicaSet`, so that you can call `wait_for`, release the lock to `ReplicaSet`, and then await `wait_for` future.

lib/storage/src/dispatcher.rs
```diff
@@ -205,40 +220,12 @@ impl Dispatcher {
...
-                            ));
-                        }
-                    }
-                }
-
-                while let Some(result) = wait_for_active.next().await {
-                    result?;
-                }
+                let remaining_timeout =
+                    wait_timeout.map(|timeout| timeout.saturating_sub(start.elapsed()));
```
ffuugoo, line 226: Why?

generall: same logic, as for `do_sync_nodes`. Or what do you mean?

ffuugoo: why decrease the timeout. but I looked around, it makes sense, I guess

PR #6819, opened 2025-07-07
Track payload index schema version

ffuugoo:
Track payload index schema version, so that we can exclude unchanged payload index files from partial snapshots.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/segment/version_tracker.rs
+++ b/lib/segment/src/segment/version_tracker.rs
@@ -18,6 +18,14 @@ pub struct VersionTracker {
 
     /// Tracks version of *immutable* files inside payload index.
     /// Should be updated when payload index *schema* of the field is modified.
+    ///
+    /// Generally, we can rely on `Segment::initial_version` to filter immutable files.
+    /// E.g., HNSW index is created when creating immutable segment, and so immutable files
+    /// of HNSW index stays immutable for the whole lifetime of the segment.
+    ///
+    /// However, payload indices can be updated at any moment even for immutable segments.
+    /// E.g., payload index can be created *after* immutable segment is created, and so we have to
+    /// track payload index schema version separately from `Segment::initial_version`.
     payload_index_schema: HashMap<JsonPath, SeqNumberType>,
 }
 


generall:
Should index version be cleared here as well? https://github.com/qdrant/qdrant/blob/a7ee4ceb512e4d220c3254b9540e098eee609fed/lib/segment/src/segment/segment_ops.rs#L489

lib/segment/src/segment/version_tracker.rs
```diff
@@ -1,11 +1,15 @@
...
+use std::hash::Hash;
 
+use crate::json_path::JsonPath;
 use crate::types::{SeqNumberType, VectorNameBuf};
 
 #[derive(Clone, Debug, Default)]
 pub struct VersionTracker {
     vector_storage: HashMap<VectorNameBuf, SeqNumberType>,
     payload_storage: Option<SeqNumberType>,
+    payload_index: HashMap<JsonPath, SeqNumberType>,
```
generall: Is it right, that we track here the version of index Config, not the content of payload index?

ffuugoo:
> Should index version be cleared here as well?

Not really? We track version when index was *created*, and we don't recreate indices here?

lib/segment/src/segment/version_tracker.rs
ffuugoo: We only track when particular index was *created*, not when it was modified. And we use this version to include/exclude *immutable* files of the index.

lib/segment/src/segment/snapshot.rs
```diff
@@ -160,6 +160,23 @@ impl Segment {
...
 
+        let payload_index_files = self
+            .payload_index
+            .borrow()
+            .immutable_files()
+            .into_iter()
+            .map(|(field, file)| {
+                let version = FileVersion::from(
+                    self.version_tracker
+                        .get_payload_index(&field)
```
timvisee: A field may have multiple indices, of which only some are updated. Do we want to consider that case as well?

For example, we have a numeric and null index on `"a"`.

We currently have: `{"a": 123}`

If we update it to `{"a": 456}`, only the numeric index files are changed. The null index remains the same.

lib/segment/src/segment/version_tracker.rs
```diff
@@ -54,3 +52,25 @@ fn bump(current: Option<SeqNumberType>, new: Option<SeqNumberType>) -> Option<Se
...
+    let Some(current) = map.get_mut(key) else {
+        map.insert(key.to_owned(), new);
+        return;
+    };
+
+    if *current < new {
+        *current = new;
+    } else {
+        map.remove(key);
+    }
```
timvisee, line 92: Why not use the [`entry`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry) API here?

lib/segment/src/segment/snapshot.rs
ffuugoo: For payload *indices* (not *storage*!) we only track version of index *(re)creation*. E.g., we only track when we change payload index config.

So, in your example, we'll simply include more data (e.g., potentially unchanged `null` index) into the snapshot, *if you updated payload index schema* since last partial snapshot, which I think is fine.

And even if further optimization required, I propose we do it in a follow up PR.

lib/segment/src/segment/version_tracker.rs
ffuugoo: Less unnecessary `key.to_owned()` allocations and I think explicit `let Some(...)` and `if` reads better.

ffuugoo:
> Please add more documentation comments into the VersionTracker

Done.

PR #6829, opened 2025-07-08
Remove fallback eager allocation

agourlay:
Avoid minor unnecessary allocations.

### DIFF
--- a/lib/collection/src/collection/distance_matrix.rs
+++ b/lib/collection/src/collection/distance_matrix.rs
@@ -53,7 +53,7 @@ impl From<SearchMatrixRequestInternal> for CollectionSearchMatrixRequest {
             limit_per_sample: limit
                 .unwrap_or(CollectionSearchMatrixRequest::DEFAULT_LIMIT_PER_SAMPLE),
             filter,
-            using: using.unwrap_or(DEFAULT_VECTOR_NAME.to_owned()),
+            using: using.unwrap_or_else(|| DEFAULT_VECTOR_NAME.to_owned()),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


lib/collection/src/shards/conversions.rs
```diff
@@ -359,7 +359,7 @@ pub fn internal_create_index(
             ),
         })
         .map(|(field_type, field_params)| (Some(field_type), field_params))
-        .unwrap_or((None, None));
+        .unwrap_or_else(|| (None, None));
```
timvisee: The previous is probably cheaper. It's const and without a function call:

```suggestion
        .unwrap_or((None, None));
```

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -369,7 +369,7 @@ impl LocalShard {
                         true,
                     )
                 })
-                .unwrap_or(Ok(()))?;
+                .unwrap_or_else(|| Ok(()))?;
```
timvisee:
```suggestion
                .unwrap_or(Ok(()))?;
```

lib/segment/src/index/query_optimization/condition_converter.rs
```diff
@@ -62,7 +62,7 @@ impl StructPayloadIndex {
 
                 let (primary_null_index, fallback_index) = field_indexes
                     .map(|field_indexes| get_is_empty_indexes(field_indexes))
-                    .unwrap_or((None, None));
+                    .unwrap_or_else(|| (None, None));
```
timvisee:
```suggestion
                    .unwrap_or((None, None));
```

timvisee (approved):

PR #6832, opened 2025-07-08
[grpc docs] use PWD for temporary directory. Drop dockerfile generation

coszio:
After trying to debug my local setup, I failed to make the current `tools/generate_grpc_docs.sh` work. It worked before, so it must be some update to docker which fails to mount a folder within `/var/folders` (which `mktemp` creates), but I couldn't figure out how (it is supposed to be allowed in docker settings).

The fix is simple though, just create the temporary folder in the root. `trap` command will erase it afterward anyways.

With this PR, the temporary directory created by the script looks like this `./qdrant_docs.MF8dC4dhsU`, where the suffix is random.

This PR also ditches the `generate_grpc_docs.dockerfile*` files which produced an invalid documentation.


timvisee:
Can you try to see what happens if you add these two entries in `~/Library/Group\ Containers/group.com.docker/settings.json`, and then fully restart the Docker daemon:

```
“filesharingDirectories” : [
  ...
  “\/var\/folders”,
  “\/private\/var\/folders”
],
```

Because I'm not a huge fan of this just because of macOS shenanigans :face_holding_back_tears:

coszio:
It works!! :old_man_yells_at_docker:

PR #6833, opened 2025-07-09
Do not track HW measurement when matching over a text index

agourlay:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


generall:
Is it strictly necessary to disable measurement for the whole matching? Could we maybe disable measurement only for reading posting list?

agourlay:
> Could we maybe disable measurement only for reading posting list?

As far as I can see there are no other operations using the HW counters during the text index matching.
This is why I removed the arguments completely.

Reading the posting lists outside of the matching process is fine.
Only the matchers exhibit the pathological case.

agourlay:
:bell: gentle ping to get this one reviewed and merged :)

timvisee:
> Due to the current architecture, the query token posting lists are repeatedly read from mmap which skews the hardware statistics incorrectly.

Do we assume here that we either: don't read the data at all (already cached), or we read it exactly once (cache on first read)? In which case not counting this at all gives us better results that counting on each iteration?

Edit: I see that we now explicitly state that (and in other places): https://github.com/qdrant/qdrant/blob/92c488e3f70c72c71320cbd05e46fb05ddf53db4/lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/mod.rs#L246-L248

lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/mod.rs
```diff
@@ -247,12 +242,15 @@ impl MmapInvertedIndex {
             postings: &MmapPostings<V>,
             tokens: &TokenSet,
             point_id: PointOffsetType,
-            hw_counter: &HardwareCounterCell,
         ) -> bool {
+            // Do not track the HW counter for reading the query posting list because:
+            // - it is repeated many times (once per matched point_id)
+            // - it is most likely mmap cached after the first call
+            let disposable = HardwareCounterCell::disposable();
```
timvisee, line 249: If we assume that we'll cache exactly once now. Does it make sense to count _something_ for each token outside of this function? That way we don't count each iteration, but once each call to `check_has_subset`.

Maybe we can count the sum of this:
- for each token:
  - posting list header size:
    <https://github.com/qdrant/qdrant/blob/7c911abae008b379dcc828a2f3112ca5662e79ca/lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/mmap_postings.rs#L171-L173>
  - point offset size:
    <https://github.com/qdrant/qdrant/blob/7c911abae008b379dcc828a2f3112ca5662e79ca/lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/mmap_postings.rs#L129>
  - size of a chunk:
    <https://github.com/qdrant/qdrant/blob/7c911abae008b379dcc828a2f3112ca5662e79ca/lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/mmap_postings.rs#L133>

Or does that still badly skew the results?

agourlay: Thanks for the alternative :+1: 

So `check_has_subset` is called once per matched point_id.

To implement your idea, we would have to perform the synthetic IO tracking in the filter optimization, just outside of the matching closure.

https://github.com/qdrant/qdrant/blob/c6e38d16a7e85ff641b19cdf8aee3d8600aba1e9/lib/segment/src/index/query_optimization/condition_converter/match_converter.rs#L260-L269

I think technically it would work, I am just not a super fan of it :smile_cat: 

The resulting code would be a bit convoluted with a large comment explaining why it exists.

Moreover, doing the sum once would result in only a few added bytes which do not change much the total cost of the request.

For those reasons, I'd rather not measure it at all as it solves the over measurement issue without introducing a large discrepancy with the reality.

timvisee: I agree. Coming up with an artificial amount to count is not obvious.

PR #6840, opened 2025-07-09
fix check for pure primary condition filter

generall:
- Fix `total_conditions_count`, so it doesn't count nested must conditions as multiple conditions
- Change `PrimaryCondition` type to store external ids for: a) easier comparison with original condition b) avoiding mapping of the conditions, which might not be even used
- Change criteria for checking if filter consists only of primary conditions. Previous problem: `min_should` can generate more primary conditions then there are conditions defined (by duplicating conditions)

### DIFF
--- a/lib/segment/src/index/struct_payload_index.rs
+++ b/lib/segment/src/index/struct_payload_index.rs
@@ -1,9 +1,9 @@
 use std::collections::HashMap;
 use std::fs::create_dir_all;
+use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
-use ahash::AHashSet;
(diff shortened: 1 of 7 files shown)


lib/segment/src/index/struct_payload_index.rs
```diff
@@ -42,7 +42,7 @@ use crate::types::{
 use crate::vector_storage::{VectorStorage, VectorStorageEnum};
 
 #[derive(Debug)]
-#[expect(clippy::enum_variant_names)]
+#[allow(clippy::enum_variant_names)]
```
generall, line 44: Clippy with `--no-default-features` was complaining, I couldn't find a better way to fix it. Not directly related to PR

PR #6844, opened 2025-07-10
adjsut default values for better performance

generall:
* `indexing_threshold`: 20k -> 10k
* default `max_segment_size`: 200mb -> 256mb per CPU core
* Default number of segments: 1 -> 2 per CPU

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -129,11 +129,11 @@ storage:
     memmap_threshold_kb: null
 
     # Maximum size (in KiloBytes) of vectors allowed for plain index.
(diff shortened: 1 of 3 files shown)


timvisee (approved):
Untested, but these adjustments look sane to me.

agourlay:
Are we expecting an improvement on the CI benchmark after this PR?

generall:
> Are we expecting an improvement on the CI benchmark after this PR?

No, CI benchmarks have explicitly configured values

coszio (approved):
I think PR description is misleading. Default segment number becomes half of available cpus, so that there are 2 cpus per segment

PR #6845, opened 2025-07-10
Refactor bash tests: low-disk

tellet-q:
This PR creates a new test suite e2e_tests. It is a first PR in a series to replace some of the bash tests. The goal is to reduce the number of jobs run in Integration tests workflow and unify running of the tests. 

This replaces tests/low-disk

To run the new test suite install the tests dependencies and in project root execute pytest tests/e2e_tests or pytest -n 4 tests/e2e_tests (to run in 4 parallel workers)


tests/e2e_tests/utils.py
```diff
@@ -0,0 +1,17 @@
...
+    """Wait for Qdrant service to be ready."""
+    start_time = time.time()
+    while time.time() - start_time < timeout:
+        try:
+            response = requests.get(f"http://localhost:{port}/readyz")
+            if response.status_code == 200:
+                return True
+        except requests.exceptions.ConnectionError:
+            pass
+        time.sleep(1)
```
timvisee: Nit: we can make it a bit faster


```suggestion
        time.sleep(0.2)
```

Probably does not have a significant effect now, but it might have in the future when we have a lot more tests.

.github/workflows/integration-tests.yml
```diff
@@ -325,3 +207,75 @@ jobs:
...
+        with:
+          lfs: true
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - name: Install and set up git-lfs
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y git-lfs
+          git lfs install
```
timvisee: Just curious, do we have anything in lfs?

I know we had some snapshots in there before, but not sure if they are still there.

tests/e2e_tests/conftest.py
```diff
@@ -0,0 +1,422 @@
...
+        self.network = network
+        self.network_name = network_name
+    
+    def __repr__(self) -> str:
+        return f"QdrantCluster(leader={self.leader.name}, followers={len(self.followers)}, network='{self.network_name}')"
+
+
+@dataclass
+class QdrantContainerConfig:
+    """Configuration for creating Qdrant containers with proper typing and defaults."""
```
timvisee: Do we have types like this so that our IDEs suggest better things when writing a test?

timvisee (approved):

.github/workflows/integration-tests.yml
tellet-q: The current test suite doesn't have anything. All the data is stored as binary right now.

tellet-q: Let's remove these steps for now

tests/e2e_tests/conftest.py
tellet-q: I use class object instead of a dict, just so that type definitions are clear

PR #6848, opened 2025-07-10
Fix HW vector io for scalar quantization

agourlay:
Minor fix to make sure we are reporting the correct vector IO count for scalar quantization.


coszio (approved):
Looks sane to me

generall (approved):

IvanPleshkov (approved):
Cool! Also triggered gpu tests to make sure that SQ GPU tests passed
https://github.com/qdrant/qdrant/actions/runs/16205736260

PR #6849, opened 2025-07-11
Bump http 1.3.1

agourlay:
For some obscure reason Dependabot has been ignoring this one for the past 4 months.

https://github.com/hyperium/http/blob/master/CHANGELOG.md


lib/posting_list/Cargo.toml
```diff
@@ -16,4 +16,3 @@ zerocopy = { workspace = true }
 
 [dev-dependencies]
 rand = { workspace = true }
-tempfile = { workspace = true }
```
agourlay, line 19: unused :broom:

PR #6810, opened 2025-07-04
Naive migration of payload indices from RocksDB to Gridstore

timvisee:
This PR is now a collection of:
- https://github.com/qdrant/qdrant/pull/6810
- https://github.com/qdrant/qdrant/pull/6837 (will drop when [#6860](https://github.com/qdrant/qdrant/pull/6860) is merged)
- https://github.com/qdrant/qdrant/pull/6847

---

Depends on <https://github.com/qdrant/qdrant/pull/6812>

A naive implementation for migrating payload indices based on RocksDB into Gridstore.

Rather than carefully migrating, this simply builds the payload indices from scratch.

Naive migration is triggered when a RocksDB based payload index is loaded and the `migrate_rocksdb_payload_indices` runtime feature flag is set.

In a later PR we can dive into actually migrating. I'm quite sure a proper migration will be beneficial on large deployments.

Some other changes:
- restructure payload index `clear` and `wipe` functions:
  - _clear_ empties the index but keeps files on disk, allows reusing the index
  - _wipe_ completely removes the index, also wiping files from disk
- Gridstore based payload indices now take a weak reference to storage in flushers
- Gridstore based payload indices now take ownership of storage when wiping

### Tasks

- [x] Ensure column families are properly removed
- [x] Merge <https://github.com/qdrant/qdrant/pull/6805>
- [x] Rebase on `dev`
- [x] Merge <https://github.com/qdrant/qdrant/pull/6837> into this
- [x] Merge <https://github.com/qdrant/qdrant/pull/6812>
- [x] Rebase on `dev`
- [x] Undraft

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

PR #6860, opened 2025-07-14
Fix mmap payload index load, load in stages

timvisee:
Exactly the same as <https://github.com/qdrant/qdrant/pull/6837>, but merging into `dev` asynchronously to make <https://github.com/qdrant/qdrant/pull/6810> smaller.

---

Applies the same change as in <https://github.com/qdrant/qdrant/pull/6805>, but to our mmap payload indices.

This separates loading of our mmap payload indices into two stages. First is opening them. Second is loading. Only loading should fail if files don't exist on disk. This is important to make the interface consistent.

While this currently isn't very well implemented - using `Option`'s to wrap mmap types - I'd consider this good enough for now. It currently is possible to have these mmap payload indices in illegal state (functioning index with no files on disk). But we try to prevent this in practice by failing on load.

This PR does not change the runtime behavior of indices, and in practice, they should behave exactly the same. It does change behavior of opening/loading indices, but that is only when loading a segment. Other than that, the indices remain the same.

The plan is to do significant refactoring of payload index creation and opening. That will make the interfaces much simpler and will prevent the illegal state.

### Tasks
- [x] do same treatment to bool mmap index?
- [x] do same treatment to null mmap index?
- [x] fix warning about payload index cache clearing (if mmap files don't exist)

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/common/memory/src/fadvise.rs
+++ b/lib/common/memory/src/fadvise.rs
@@ -14,15 +14,23 @@ fn fadvise(f: &impl std::os::unix::io::AsRawFd, advise: PosixFadviseAdvice) -> i
 
 /// For given file path, clear disk cache with `posix_fadvise`
 ///
-/// If `posix_fadvise` is not supported, this function does nothing.
+/// Does nothing if:
+/// - the file does not exist
+/// - `posix_fadvise` is not supported on this platform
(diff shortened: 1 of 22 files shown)


timvisee:
This one is obsolete now. We merged this into `dev` as part of <https://github.com/qdrant/qdrant/pull/6810>.

PR #6862, opened 2025-07-14
Fix read latency spikes during partial snapshot recovery

ffuugoo:
TL;DR:
- I've noticed that all latency spikes that we observe in RW-seg cluster follow the same pattern
- following the logs/pattern I found that we call `std::thread::JoinHandle::join` ***on the async runtime*** during partial snapshot recovery
- I suspect, that the spikes are caused by this "blocking call on async runtime"
- this PR implements the most trivial workaround, so that we can quickly merge it and see if it fixes the issue on RW-seg cluster

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -241,7 +241,7 @@ impl LocalShard {
     /// Recovers shard from disk.
     #[allow(clippy::too_many_arguments)]
     pub async fn load(
-        id: ShardId,
+        _id: ShardId,
         collection_id: CollectionId,
         shard_path: &Path,
         collection_config: Arc<TokioRwLock<CollectionConfigInternal>>,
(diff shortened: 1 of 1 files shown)


generall (approved):

timvisee (approved):

ffuugoo:
So, here's a breakdown:

Screenshot 2025-07-14 at 14 33 41

- partial snapshot recovery is started (bottom log line)
- there's a `Local shard 1 not found` message
  - which simply means that local replica is unavailable, because partial snapshot is being recovered on the node
  - (the wording is a bit confusing in this case)
- however, by the time this message is logged in `execute_cluster_read_operation`, all "locks" required to execute read-request should have been acquired already
  - so, the only thing that is being executed during these 2 seconds should be network requests to remote nodes
  - (and preparing response to the user, but this is pure CPU work that should be practically instant)
- judging by the timestamp of the message, it correlates *exactly* with the total response time
  - e.g., `07:55:58.294 - 07:55:56.006 = 2.288s`
- we can also see that there are `Recovering shard...` and `Recovered collection...` logs right between `Local shard 1 not found` and the top log line
  - and **all** spikes are like these: start recovery, local shard not found, recovering/recovered, search finished (with timestamps correlating exactly)
- the network on rw-seg is *fast* and remote replicas should not be blocked by anything
  - (partial snapshot is recovered *on the same* node, that is executing the client request)
- but if there's something blocking the async runtime, that might explain why requests are stalling...
  - and lo and behold, there is a blocking call right in the code path that we know is executed in parallel with our request 🙄

PR #6864, opened 2025-07-14
Fix payload index migration config and edge case handling

timvisee:
Fix two problems in payload index migrations that has recently been merged in <https://github.com/qdrant/qdrant/pull/6810>.

The first fix is in configuration handling. This config describes the exact type of payload index being used, it is used when loading an index. The implementation was saving changes a bit too aggressively, saving an incomplete configuration in some cases. It must only persist at the end based on a dirty flag. It looks like we lost this logic due to a bad conflict resolution.

The second fix is to make payload index loading more robust. It ensures that we still open RocksDB if any of the indices on disk still relies on it, no matter if we enable migrations or not. Without the fix loading could fail resulting in a dummy segment.

Here's a case that failed before, which is now fixed in this PR:
- start Qdrant with `QDRANT__FEATURE_FLAGS__ALL=false`
- restore 'Qdrant Web Documentation' from sample snapshots in web UI
- restart Qdrant with `QDRANT__FEATURE_FLAGS__ALL=true`
- watch it fail to load in the middle of migration

The second problem was about 

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/index/payload_config.rs
+++ b/lib/segment/src/index/payload_config.rs
@@ -56,6 +56,19 @@ impl PayloadIndices {
         self.fields.values().any(|index| index.types.is_empty())
     }
 
+    /// Check if any payload field used RocksDB
+    ///
+    /// Returns false if empty.
+    #[cfg(feature = "rocksdb")]
+    pub fn any_is_rocksdb(&self) -> bool {
+        self.fields.values().any(|index| {
+            index
+                .types
+                .iter()
+                .any(|t| t.storage_type == StorageType::RocksDb)
+        })
+    }
+
     pub fn to_schemas(&self) -> HashMap<PayloadKeyType, PayloadFieldSchema> {
         self.fields
             .iter()
--- a/lib/segment/src/index/struct_payload_index.rs
+++ b/lib/segment/src/index/struct_payload_index.rs
@@ -254,7 +254,7 @@ impl StructPayloadIndex {
             }
         }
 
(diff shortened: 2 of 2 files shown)


coszio (approved):
Manually tested the fix 👍 

Some rocksdb files are still present. For indices we removed column families, but the rest of db files are leftover. I believe this can be done in another PR

timvisee:
> Some rocksdb files are still present. For indices we removed column families, but the rest of db files are leftover. I believe this can be done in another PR

We don't delete RocksDB files, because we don't know what files belong to it.

Related issue I posted some time back: https://github.com/facebook/rocksdb/issues/13445

Since the payload index directory is shared and it contains other things, we cannot just drop the full directory.

I don't consider this too big of a problem. The files will eventually be deleted with another optimization round.

timvisee:
@coszio and cc to previous comment

I gave it another look, and for deletion, it turns out it is possible :smile: 

- https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ#:~:text=Q%3A%20What%20is%20the%20correct%20way%20to%20delete%20the%20DB%3F%20Can%20I%20simply%20call%20DestroyDB()%20on%20a%20live%20DB%3F
- https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.destroy

Not exactly sure how it behaves, but I'll give it a shot in a separate PR now.

Edit: implemented in <https://github.com/qdrant/qdrant/pull/6873>

PR #6872, opened 2025-07-15
Enable migration from RocksDB to mutable ID tracker by default

timvisee:
Following our [planning](https://www.notion.so/qdrant/Milestone-Remove-RocksDB-211674779d33805f99a2f178e8d7c0f6), we enable the migration from RocksDB to mutable ID tracker by default in Qdrant 1.15.0.

The TODOs for other feature flags have also been adjusted to follow our planning.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -11538,8 +11538,8 @@
             "type": "boolean"
           },
           "migrate_rocksdb_id_tracker": {
-            "description": "Whether to actively migrate RocksDB based ID trackers into a new format.",
-            "default": false,
+            "description": "Whether to actively migrate RocksDB based ID trackers into a new format.\n\nEnabled by default in Qdrant 1.15.0.",
+            "default": true,
             "type": "boolean"
           },
           "migrate_rocksdb_vector_storage": {


ffuugoo:
@timvisee can you, please, explain the migration plan in some more details?

E.g., what `payload_index_skip_mutable_rocksdb` flag does? Prefer non-rocks payload index impls when creating *new* indices?

So, I guess:
- we already committed support for non-rocks id tracker in prev release?
- and so we enable *migration* of *existing* id trackers to non-rocks in 1.15 and you can still downgrade to prev version even after migration?
- and similar for all other structures? we'll release implementations in 1.15, then *prefer* these impls for *new* structs in 1.16 and then *migrate* in 1.16.1?

ffuugoo (approved):

timvisee:
> E.g., what `payload_index_skip_mutable_rocksdb` flag does? Prefer non-rocks payload index impls when creating _new_ indices?

Yes.

> So, I guess:
> 
>   * we already committed support for non-rocks id tracker in prev release?
>   * and so we enable _migration_ of _existing_ id trackers to non-rocks in 1.15 and you can still downgrade to prev version even after migration?
>   * and similar for all other structures? we'll release implementations in 1.15, then _prefer_ these impls for _new_ structs in 1.16 and then _migrate_ in 1.16.1?

Yes.

The flags and their behavior is elaborated upon [here](https://github.com/qdrant/qdrant/blob/f95190c82e1aef51a38e2adf6ab553a3aef6d0cb/lib/common/common/src/flags.rs#L9-L58). But I agree, the information is a bit sparse and we could add some more details there.

Let's talk about the ID tracker specifically. The [mutable ID tracker](https://github.com/qdrant/qdrant/blob/master/lib/segment/src/id_tracker/mutable_id_tracker.rs) has been implemented as a replacement for the [RocksDB based ID tracker](https://github.com/qdrant/qdrant/blob/master/lib/segment/src/id_tracker/simple_id_tracker.rs). It was released in 1.13.6.

In Qdrant 1.14 we enabled the new ID tracker by default (<https://github.com/qdrant/qdrant/pull/6268>). It means that all new (or optimized) segments will start using the mutable ID tracker rather than the RocksDB based one. This is fine because people could still downgrade to 1.13.6.

In Qdrant 1.15 we are now actively migrating away from the RocksDB based ID tracker. It'll migrate the ones that were not picked up yet by the optimizer. This way we prevent keeping the RocksDB based variant forever.

All of these follow a similar pattern:
1. implement and release new variant of some component, that does not use RocksDB
2. in the next minor version, start using it for all new segments (we can downgrade one minor version)
3. in the next minor version, actively migrate components which are still untouched
4. in the next minor version, remove old component variant

There are multiple components that rely on RocksDB. Replacements for them have been implemented at different times, which is why we enable certain flags at different times too. See [this](https://www.notion.so/qdrant/Milestone-Remove-RocksDB-211674779d33805f99a2f178e8d7c0f6) document.

Note that we keep 'use new variant for new instance' and 'actively migrate old to new variant' separate. Generally speaking, these migrations may become very expensive on humongous deployments. By keeping these two things separate, we give customers time to slowly transform their dataset into the new variants, while keeping the migration impact as small as possible.

ffuugoo:
Thanks for the explanation! 🙏

Issue #6880, opened 2025-07-16
On v1.14.1 search queries using an is_empty filter don't return recently added vectors


IllDepence:
## Current Behavior

* Search queries using an is_empty filter don't return recently added vectors (indices exist and payloads fit the query filter)
* Removing the is_empty filter from the query makes it return the recently added vectors.
* Restarting the cluster resolves the problem (=makes queries with the is_empty filter return the recently added vectors).

## Steps to Reproduce

Create a v1.14.1 cluster on Qdrant Cloud

Run the following PoC
https://gist.github.com/IllDepence/433cb58cc8148a0d9b9e3a9df1b6ef39

Output
```
$ npx ts-node qudrant_v1.14.1_is_empty_bug.ts
Created collection: test_collection
Created index: redacted_c
Created index: redacted_h
Added points
Querying with is_empty filter
Results (0)
Querying without is_empty filter
Results (2)
id: 1
score: 0.9
payload: {"redacted_c":"foo"}
id: 2
score: 0.8
payload: {"redacted_c":"foo"}
```

## Expected Behavior

Search queries using an is_empty filter should return recently added vectors.

(For the PoC above: a query with `{ is_empty: { key: "redacted_h" } }` should return vectors whose payload does not have the filed `redacted_h`.)

## Possible Solution

\-

## Context (Environment)

Newly created clusters currently require restarting whenever new vectors are added to a cluster to make them searchable.

## Detailed Description

The bug described above should be fixed.

## Possible Implementation

\-

timvisee:
Thanks for the clear report! I just confirmed this is a bug. Sorry for the inconvenience.

I can reproduce this in 1.14.1 (latest release), and also in our latest development version.

Here's a minimal reproducible snippet for use in our web UI:

```json
PUT /collections/test
{
  "vectors": {
    "size": 3,
    "distance": "Dot"
  },
  "shard_number": 1
}

PUT /collections/test/index
{
  "field_name": "a",
  "field_schema": "keyword"
}

PUT /collections/test/points
{
    "points": [
        {
            "id": 1,
            "vector": [0.9, 0.1, 0.1]
        },
        {
            "id": 2,
            "vector": [0.1, 0.9, 0.1]
        },
        {
            "id": 3,
            "vector": [0.1, 0.1, 0.9]
        }
    ]
}

POST /collections/test/points/query
{
  "filter": {
    "must": [
      {
        "is_empty": {
          "key": "a"
        }
      }
    ]
  },
[...]

PR #6882, opened 2025-07-16
Fix broken is_empty filter on new points

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/6880>

When upserting new points, the `is_empty` filter did not work properly. I did not always include the newly upserted points during filtering. 

More specifically, this problem appears when upserting points in which the payload field (of the null index) is absent. The null index relies on knowing the exact (segment) point count to reliably filter points. When upserting a point in which the payload field is absent, we don't add the point to the index and therefore don't bump the point count.

When the payload field is absent, instead of adding ([set](https://github.com/qdrant/qdrant/blob/4f9884de3ace92cf0fe99a9443d411f35f3810ce/lib/segment/src/index/struct_payload_index.rs#L1030), [overwrite](https://github.com/qdrant/qdrant/blob/4f9884de3ace92cf0fe99a9443d411f35f3810ce/lib/segment/src/index/struct_payload_index.rs#L994)) the point to the index, we actually remove ([set](https://github.com/qdrant/qdrant/blob/4f9884de3ace92cf0fe99a9443d411f35f3810ce/lib/segment/src/index/struct_payload_index.rs#L1034), [overwrite](https://github.com/qdrant/qdrant/blob/4f9884de3ace92cf0fe99a9443d411f35f3810ce/lib/segment/src/index/struct_payload_index.rs#L998)) it from the index. To fix the filtering problem, I now also bump the point count when removing a point. This will ensure that new upsertions without the field still set the correct point count.

This PR fixes two problems:
- not setting the total point count (described above)
- loading the null index from the wrong path (in some scenarios)

I also add a test to assert correct behavior.

Existing deployments do not have permanent damage because of this bug. All (existing) null indices are fixed on restart.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/index/field_index/null_index/mmap_null_index.rs
+++ b/lib/segment/src/index/field_index/null_index/mmap_null_index.rs
@@ -184,7 +184,7 @@ impl MmapNullIndex {
             .is_null_slice
             .set_with_resize(id, is_null, hw_counter_ref)?;
 
-        // Update total_points to track the highest point offset seen
+        // Bump total points
         self.total_point_count = std::cmp::max(self.total_point_count, id as usize + 1);
 
         Ok(())
@@ -204,6 +204,14 @@ impl MmapNullIndex {
         storage
             .is_null_slice
             .set_with_resize(id, false, disposed_hw)?;
+
+        // Bump total points
+        // We MUST bump the total point count when removing a point too
+        // On upsert without this respective field, remove point is called rather than add point
+        // Bumping the total point count ensures we correctly iterate over all empty points
+        // Bug: <https://github.com/qdrant/qdrant/issues/6880>
+        self.total_point_count = std::cmp::max(self.total_point_count, id as usize + 1);
+
         Ok(())
     }
 
--- a/lib/segment/src/index/field_index/null_index/mmap_null_index.rs
+++ b/lib/segment/src/index/field_index/null_index/mmap_null_index.rs
@@ -93,6 +93,12 @@ impl MmapNullIndex {
         })
     }
 
+    /// Open a null index at the given path, only if it exists.
+    ///
+    /// # Arguments
+    /// - `path` - The directory where the index files should live, must be exclusive to this index.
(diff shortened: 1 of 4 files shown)


lib/segment/src/index/field_index/null_index/mmap_null_index.rs
```diff
@@ -198,6 +204,14 @@ impl MmapNullIndex {
         storage
             .is_null_slice
             .set_with_resize(id, false, disposed_hw)?;
+
+        // Bump total points
+        // We MUST bump the total point count when removing a point too
+        // On upsert without this respective field, remove point is called rather than add point
+        // Bumping the total point count ensures we correctly iterate over all empty points
+        // Bug: <https://github.com/qdrant/qdrant/pull/6882>
+        self.total_point_count = std::cmp::max(self.total_point_count, id as usize + 1);
```
JojiiOfficial, line 213: Shouldn't be this `-1` instead of `+1` because the removed point ID doesn't exist anymore after this function.

Edit: `-1` is most likely wrong because that would be a `-2` on the total value. So if `+1` is wrong, just the addition must be removed I think.

timvisee: I believe this is correct.

A point is never actually removed from a live segment. It is only soft deleted. It may only be 'skipped' by the optimizer when moving it to a new segment.

Removing a point here should not decrease the total point count.

If we upsert point ID 99 with payload `{}`, we call `remove_point(99, {})`. It will set the total point count to 100 here, which is correct.

If we had point 99 point already, it being the last one, and we remove it now, we still keep 100 total points.

Note that these are internal point IDs, not external (user provided) ones.

JojiiOfficial: Thanks for the clarification! 
I wasn't fully aware of the context and usage of `total_point_count` but this looked wrong.
If this +1 is intentional, fine with me ✅

JojiiOfficial (approved):

PR #6883, opened 2025-07-16
Fix trying to delete rocksDb files when creating new segments

JojiiOfficial:
When building new segments, we sometimes create empty rocksDb databases for payload indices.
When loading the empty `StructPayloadIndex` directly after creating, we previously attempted to delete it again because it was detect as "not used".
This threw an error because it was actually used.

This PR fixes this warning and the attempt to delete used rocksDb database files.

### DIFF
--- a/lib/segment/src/index/payload_config.rs
+++ b/lib/segment/src/index/payload_config.rs
@@ -69,6 +69,19 @@ impl PayloadIndices {
         })
     }
 
+    /// Check if any payload field used RocksDB or could be using rocksDB.
+    /// This means that if we have no data about an indexes' type, we still return `true`.
+    #[cfg(feature = "rocksdb")]
+    pub fn any_could_be_rocksdb(&self) -> bool {
+        self.fields.values().any(|index| {
+            index
+                .types
+                .iter()
+                .any(|t| t.storage_type == StorageType::RocksDb)
+                || index.types.is_empty()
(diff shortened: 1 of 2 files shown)


generall:
If I understood correctly, we tried to delete RocksDB which might not be migrated yet. And it caused warning, that RocksDB is still used, right?

Could it potentially cause data loss?

lib/segment/src/index/payload_config.rs
```diff
@@ -69,6 +69,19 @@ impl PayloadIndices {
         })
     }
 
+    /// Check if any payload field used RocksDB or could be using rocksDB.
+    /// This means that if we have no data about an indexes' type, we still return `true`.
```
KShivendu, line 73:
```suggestion
    /// This means that if we have no data about an index's type, we still return `true`.
```
I have very little context of this PR. But nit :)

timvisee: Actually, I think 
```suggestion
    /// This means that if we have no data about an index type, we still return `true`.
```
may be better

timvisee (changes requested):
I don't think this is correct.

I suggest this instead: https://github.com/qdrant/qdrant/pull/6888

JojiiOfficial:
Thanks for the suggestion! 
Closing in favor of #6888

PR #6884, opened 2025-07-16
Remove hardware measurement from fulltext index

agourlay:
Follow up on https://github.com/qdrant/qdrant/pull/6833

We have decided to simply remove the hardware measurement in the fulltext index via the posting list infrastructure.

The reasons are:
- we do not want to introduce yet another `disposable` metric black hole
- the measured values are small and do not bring much in steady state


agourlay:
gentle nudge @timvisee to review this old PR which fell through the cracks :)

timvisee (approved):

PR #6886, opened 2025-07-16
Update collection-level payload index schema after recovering partial snapshot

ffuugoo:
When recovering shard snapshot, we check that indices inside recovered segments match payload index schema (that is set on collection level) and (re)create missing or mismatched payload indices.

However, this does not really work with RW segregation, e.g.:
- we recover shard 0 on read node
- we update collection payload schema on write node
- we recover shard 1 on read node
- now, shard 0 and shard 1 on read node have different set of payload indices...
- ...and we have to *somehow* figure which payload schema to use...
-  ...and we also don't want to *(re)create* any indices on the read cluster

So, in this PR we introduce the following behavior: after recovering partial snapshot, we set collection-level payload index schema to the **intersection** of payload index schemas of **all** shards that are present on the node. Payload field index should exist in *all* shards and have exactly the same *type and parameters* to be included into "intersection".

The logic for this is, in RW segregation cluster we only update read-cluster using partial-snapshots taken from write-cluster. Including creating or removing payload indices. And so we expect, that if we updated payload index schema on write cluster, then after we recovered each shard on particular read-cluster node, payload field indices in all these shards should converge to the payload index schema that is set on the write-cluster.

<details>
<summary>examples</summary>

- if shard 0 has the following indices
  - timestamp - date-time
  - random - integer
  - city - keyword
- and shard 1 has
  - timestamp - date-time
  - random - float
  - country - keyword
- then collection payload index schema would be set to
  - timestamp - date-time

- if we later recover another partial snapshot on shard 1 which would update field indices to
  - timestamp - date-time
  - random - integer
  - city - keyword
  - country - keyword
- then collection payload index schema would be set to
  - timestamp - date-time
  - random - integer
  - city - keyword

</details>

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/segment/src/types.rs
```diff
@@ -1879,13 +1879,35 @@ impl Validate for PayloadSchemaParams {
...
 
-#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, PartialEq, Hash, Eq)]
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
 #[serde(untagged, rename_all = "snake_case")]
 pub enum PayloadFieldSchema {
     FieldType(PayloadSchemaType),
     FieldParams(PayloadSchemaParams),
 }
 
+impl cmp::Eq for PayloadFieldSchema {}
```
timvisee: Why not derive `Eq`?

lib/segment/src/types.rs
```diff
@@ -1879,13 +1879,35 @@ impl Validate for PayloadSchemaParams {
...
+
+impl cmp::PartialEq for PayloadFieldSchema {
```
timvisee: `PartialEq` is part of the standard prelude: https://doc.rust-lang.org/std/prelude/index.html

```suggestion
impl PartialEq for PayloadFieldSchema {
```

If you apply this, it also needs adjusting of imports to satisfy clippy.

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -1061,6 +1070,74 @@ impl ShardHolder {
...
+            if shard_idx == 0 {
+                schema = shard_schema;
+            } else {
+                schema.retain(|field, schema| {
+                    let Some(shard_schema) = shard_schema.get(field) else {
+                        return false;
+                    };
+
+                    schema == shard_schema
+                });
```
timvisee, line 1130: How about using [`intersection()`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.intersection) on a hash set? With `(field, schema)` values.

lib/collection/src/shards/replica_set/snapshots.rs
```diff
@@ -392,6 +392,7 @@ impl ShardReplicaSet {
                 self.optimizers_config.clone(),
                 self.shared_storage_config.clone(),
                 self.payload_index_schema.clone(),
+                recovery_type.is_full(),
```
timvisee, line 395: Shouldn't this be inverted? So that we only apply the payload index intersection when using partial snapshots?

timvisee (approved):

lib/collection/src/shards/shard_holder/mod.rs
timvisee: It wouldn't allow us to easily eliminate duplicate fields with different schemas, so never mind.

lib/segment/src/types.rs
ffuugoo: I'm stoopid.

lib/collection/src/shards/replica_set/snapshots.rs
ffuugoo: `rebuild_payload_index` (the argument highlighted in the comment) enables [re-creating the actual indices][rebuild-payload-index] inside the segments when loading local shard, if indices don't match payload index schema config. So for partial snapshots we want it `false`, so that we *don't* modify indices at all.

Updating payload index schema config is handled [here][update-payload-index-schema].

[rebuild-payload-index]: https://github.com/qdrant/qdrant/pull/6886/files#diff-b160a7fc7dd6eea120a03469ece643a035986c1ef8263e0c73949cbf52637500R338-R341
[update-payload-index-schema]: https://github.com/qdrant/qdrant/pull/6886/files#diff-c7b57d056e900423c6f53dbc990d5b93c9f73efea12f92b8c957fea462648b20R1035-R1041

PR #6887, opened 2025-07-16
Replace some Box<T> occurrences with `Either`

JojiiOfficial:
Replaces some `Box<dyn Iterator...` with `impl Iterator` by returning `Either` variants.
In some cases we were wrapping an iterator multiple times inside `Box<>`, which wasn't necessary.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/mod.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/mod.rs
@@ -214,7 +214,10 @@ impl MmapInvertedIndex {
                 return Box::new(std::iter::empty());
             }
 
-            intersect_compressed_postings_iterator(posting_readers, filter)
+            Box::new(intersect_compressed_postings_iterator(
+                posting_readers,
+                filter,
+            ))
         }
 
         match &storage.postings {
@@ -273,9 +276,9 @@ impl MmapInvertedIndex {
         &'a self,
         phrase: Document,
         hw_counter: &'a HardwareCounterCell,
-    ) -> Box<dyn Iterator<Item = PointOffsetType> + 'a> {
+    ) -> impl Iterator<Item = PointOffsetType> + 'a {
         let Some(storage) = &self.storage else {
-            return Box::new(std::iter::empty());
+            return Either::Left(std::iter::empty());
         };
 
         // in case of mmap immutable index, deleted points are still in the postings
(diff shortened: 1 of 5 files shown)


lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/mod.rs
```diff
@@ -214,7 +214,10 @@ impl MmapInvertedIndex {
                 return Box::new(std::iter::empty());
             }
 
-            intersect_compressed_postings_iterator(posting_readers, filter)
+            Box::new(intersect_compressed_postings_iterator(
+                posting_readers,
+                filter,
+            ))
```
timvisee, line 220: Is using Either not possible here?

lib/segment/src/index/field_index/full_text_index/inverted_index/mutable_inverted_index.rs
```diff
@@ -209,7 +209,7 @@ impl InvertedIndex for MutableInvertedIndex {
         _hw_counter: &HardwareCounterCell,
     ) -> Box<dyn Iterator<Item = PointOffsetType> + '_> {
         match query {
-            ParsedQuery::Tokens(tokens) => self.filter_has_subset(tokens),
+            ParsedQuery::Tokens(tokens) => Box::new(self.filter_has_subset(tokens)),
```
timvisee, line 212: And here?

timvisee (approved):

lib/segment/src/index/field_index/full_text_index/inverted_index/mutable_inverted_index.rs
JojiiOfficial: `filter` is a trait function which require the return type to be specified. Therefore we can't return `impl Iterator` here.

lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/mod.rs
JojiiOfficial: It's possible but it would not have any benefits.
The box returned here gets directly returned a few lines below at the end of `filter_has_subset`. So it would just 'move' the box to a different place in the code. 
Since `intersection` is a function inside another function `filter_has_subset` it can't be reused somewhere else in the code so I didn't see a reason to add `Either` here.

lib/segment/src/index/field_index/full_text_index/inverted_index/mutable_inverted_index.rs
KShivendu: I feel this makes the PR a bit inconsistent. imo it's better to do it consistently across those files. Are there any clear perf benefits from this change?

timvisee: Trait cannot return `impl`, of course. Thanks!

@KShivendu Yes. There is a benefit, especially on chunk iterators.

Either is optimized at compile time (sort of), and box is something runtime. It dumps your data on the heap, and so it introduces some (unnecessary) indirection.

KShivendu (commented):
Interesting approach. I learnt something new.  :)

Have dropped a comment regarding inconsistency of usage of `Either` since it makes the code a bit harder to read/maintain.

PR #6890, opened 2025-07-17
Minor inference code cleanups

JojiiOfficial:
This PR doesn't introduce logic changes, only a suggestion for minor cleanups in inference code.

### DIFF
--- a/src/common/inference/service.rs
+++ b/src/common/inference/service.rs
@@ -146,13 +146,15 @@ impl InferenceService {
     pub fn init_global(config: InferenceConfig) -> Result<(), StorageError> {
         let mut inference_service = INFERENCE_SERVICE.write();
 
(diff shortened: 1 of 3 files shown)


src/common/inference/service.rs
```diff
@@ -192,11 +189,11 @@ impl InferenceService {
             token,
         };
 
-        let url = self.config.address.as_ref().ok_or_else(|| {
-            StorageError::service_error(
```
KShivendu, line 196: I think this one was was better previously.

JojiiOfficial: I personally find closures harder to read but I don't have a strong preference here. Do you prefer this change to be reverted?

timvisee (approved):

src/common/inference/service.rs
KShivendu: Upto you. I like '?' operator :)

PR #6892, opened 2025-07-17
Deprecations for Qdrant 1.15.0

timvisee:
Deprecates:
- `init_from` in the collection API
- `memmap_threshold` in the optimizers config
- [`/locks`](https://qdrant.tech/documentation/guides/administration/#locking) API
- payload filter in RBAC

See each individual commit for the specific changes related to each deprecation.

### All Submissions:

### DIFF
--- a/lib/collection/benches/batch_query_bench.rs
+++ b/lib/collection/benches/batch_query_bench.rs
@@ -55,6 +55,7 @@ fn setup() -> (TempDir, LocalShard) {
             vacuum_min_vector_number: 1000,
             default_segment_number: 2,
             max_segment_size: Some(100_000),
+            #[expect(deprecated)]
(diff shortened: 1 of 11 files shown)


lib/collection/src/operations/config_diff.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(deprecated)] // hack to remove warning for memmap_threshold deprecation below
+
```
timvisee, line 2: A bit of a hack, sadly.

This is required to satisfy the compiler for the `memmap_threshold` field deprecation. If we remove this, it'll complain about the field itself.

The alternative would be to not mark the field as deprecated. But that would also remove the deprecation flag in OpenAPI.

PR #6891, opened 2025-07-17
BM25 in the Core

JojiiOfficial:
This PR uncouples our `Tokenizer` from `TextIndexParams` using the already existing `TokenizerConfig`.
This is required to create tokenizers from a different context than text indices.
No logic changes introduced.

PR #6893, opened 2025-07-17
Add BM25 options to inference config

JojiiOfficial:
Depends on #6891

Adds configuration for BM25 vectorizer "models" in inference. One example model with the name `bm25` added in development config too.
Highlights:
  - Supports multiple models with different configs
  - Reuse of our tokenizers with full customization support

### DIFF
--- a/config/development.yaml
+++ b/config/development.yaml
@@ -26,10 +26,11 @@ inference:
           # Whether to convert all tokens to lowercase.
           lowercase: false
           # Use predefined set of stopwords for english.
-          stopwords_filter: english
-          # Add custom words to the stopword filter.
-          custom_stopwords:
-            - hello
+          stopwords_filter: 
+            languages: 
+              - english
+            custom:
+              - hello
           # Enables stemming using the English ruleset and the Snowball stemming algorithm.
           stemmer:
             type: snowball
--- a/src/common/inference/bm25.rs
+++ b/src/common/inference/bm25.rs
@@ -25,8 +25,6 @@ pub struct TokenizerConfig {
     pub lowercase: bool,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub stopwords_filter: Option<StopwordsInterface>,
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub custom_stopwords: Vec<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub stemmer: Option<StemmingAlgorithm>,
     pub min_token_len: Option<usize>,
--- a/src/common/inference/config.rs
+++ b/src/common/inference/config.rs
@@ -54,7 +54,12 @@ impl CustomModels {
         let mut duplicates = vec![];
         let mut seen_model_names = HashSet::new();
 
-        for i in self.bm25.iter() {
+        let CustomModels {
+            model_prefix: _,
+            bm25,
+        } = &self;
+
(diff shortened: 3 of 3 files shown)


config/development.yaml
coszio, line 1: These parameters will also be included in `config/config.yaml` as official parameters once the feature is complete, right?

config/development.yaml
```diff
@@ -7,6 +7,36 @@ inference:
   address: "http://localhost:2114/api/v1/infer"
   timeout: 10
   token: "98eb568c-dea5-4347-a3a5-583478983bb9"
+  # Define custom models/vectorizer, which get handled directly in Qdrant. These do not require an `address` to be configured.
```
coszio, line 10: Maybe they can, but it is up to the specific model params. In the case of BM25 it is not needed bc it is computed in qdrant directly.

src/common/inference/bm25.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// Bm25 tokenizer configurations.
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct TokenizerConfig {
+    #[serde(default)]
+    pub lowercase: bool,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub stopwords_filter: Option<StopwordsInterface>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub custom_stopwords: Vec<String>,
```
coszio: `StopwordsInterface` already allows specifying a "language" string, or an object with language and/or stopwords set

src/common/inference/config.rs
```diff
@@ -1,11 +1,73 @@
...
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub bm25: Vec<Bm25Config>,
+}
+
+impl CustomModels {
+    pub fn duplicate_model_names(&self) -> Vec<String> {
+        let mut duplicates = vec![];
+        let mut seen_model_names = HashSet::new();
+
+        for i in self.bm25.iter() {
```
coszio: Destructure here so that when adding a different custom model it becomes more evident to have to check the name for duplicates
```suggestion
        let CustomModels {
            model_prefix: _,
            bm25,
        } = self;
        for i in bm25.iter() {
```

config/development.yaml
JojiiOfficial: Yes!

config/development.yaml
JojiiOfficial: What do you mean? `custom_models` is supposed to only handle 'models' that are handled directly in Qdrant, like BM25.

src/common/inference/bm25.rs
JojiiOfficial: Good catch!
I didn't notice that `StopwordsSet` already supports custom stopwords.

generall (changes requested):
I am not convinced inference config is the right place for this settings.
Inference settings are global, while we would want to have it per-collection.

I propose to postpone this change and make it part of custom collection parameters instead.

config/development.yaml
coszio: I was thinking that maybe in the future users might be able to plug in their own embedding services via a custom config like this. In this case another default namespace might be more fitting like `bm25/bm25-for-paragraphs`, or `native/bm25`

JojiiOfficial:
Closing in favor of https://github.com/qdrant/qdrant/pull/6939

PR #6899, opened 2025-07-17
wait-for instead of validate

generall:


tests/consensus_tests/test_consensus_compaction.py
```diff
@@ -100,7 +100,7 @@ def test_consensus_compaction_shard_keys(tmp_path: pathlib.Path):
 
     # Validate shard keys on all peers
     wait_collection_exists_and_active_on_all_peers(collection_name="test_collection", peer_api_uris=peer_api_uris)
```
ffuugoo, line 102: Do we even need this if we rely on `wait_for_shard_keys`? 🤔

tests/consensus_tests/test_consensus_compaction.py
```diff
@@ -121,11 +121,15 @@ def test_consensus_compaction_shard_keys(tmp_path: pathlib.Path):
     peer_api_uris.append(new_url)
     wait_all_peers_up([new_url])
 
+    # New node might need time to catch up with consensus state
+    # So we await for the collection to be ready on all peers
+    wait_collection_exists_and_active_on_all_peers(collection_name="test_collection", peer_api_uris=peer_api_uris)
+
```
ffuugoo, line 127: Ditto.

PR #6902, opened 2025-07-18
Fix `test_create_shard_key_read_availability` flakiness

ffuugoo:
`test_create_shard_key_read_availability` is dependent on "peer metadata" to propagate through cluster, so this PR exposes peer metadata in telemetry and add explicit `wait_for_peer_metadata` (instead of simple `sleep`) to the test.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -12784,6 +12784,13 @@
             },
             "nullable": true
           },
+          "peer_metadata": {
+            "type": "object",
+            "additionalProperties": {
+              "$ref": "#/components/schemas/PeerMetadata"
+            },
+            "nullable": true
+          },
           "metadata": {
             "type": "object",
             "additionalProperties": true,
@@ -12902,6 +12909,19 @@
             "type": "integer",
             "format": "uint64",
             "minimum": 0
+          }
+        }
+      },
+      "PeerMetadata": {
+        "description": "Metadata describing extra properties for each peer",
+        "type": "object",
+        "required": [
+          "version"
+        ],
+        "properties": {
+          "version": {
+            "description": "Peer Qdrant version",
+            "type": "string"
           }
         }
       },
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -2131,9 +2131,10 @@ pub struct IssuesReport {
 }
 
 /// Metadata describing extra properties for each peer
-#[derive(Debug, Hash, Serialize, Deserialize, Clone, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Deserialize, Serialize, JsonSchema)]
 pub struct PeerMetadata {
     /// Peer Qdrant version
+    #[schemars(schema_with = "String::json_schema")]
     pub(crate) version: Version,
 }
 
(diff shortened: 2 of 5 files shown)


lib/storage/src/types.rs
```diff
@@ -150,6 +150,10 @@ pub struct PeerInfo {
     pub uri: String,
     // ToDo: How long ago was the last communication? In milliseconds
     // pub last_responded_millis: usize
+    #[anonymize(false)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schemars(skip)]
+    pub metadata: Option<PeerMetadata>,
```
ffuugoo: `PeerMetadata` is not exposed anywhere in the API, so I had to add it to telemetry to implement `wait_for_peer_metadata`.

So far, I've simply put it into `PeerInfo` structure (with `#[schemars(skip)]`, but I'll fix the schema in a bit). It might not be the perfect place to put it, but it's easy to move around and tweak the test, so LMK what you think.

ffuugoo: @generall @timvisee 👆

KShivendu: > add it to telemetry

Just wanted to highlight that It also got added to `GET /cluster` response. It's bloating but useful info. do we wanna keep it there? 

```json
"7379954254502439": {
    "uri": "http://127.0.0.1:6335/",
    "metadata": {
        "version": "1.14.2-dev"
    }
},
"3915067487828787": {
    "uri": "http://127.0.0.3:6335/",
    "metadata": {
        "version": "1.14.2-dev"
    }
},
"3362008934670689": {
    "uri": "http://127.0.0.2:6335/",
    "metadata": {
        "version": "1.14.2-dev"
    }
}
```

ffuugoo: > It also got added to GET /cluster response

Good catch. I'd say probably not? I think I'll try to move this around or at least remove from `/custer` then.

generall: I would prefer to add it to the schema, even if it is never going to be displayed.

ffuugoo: Sure, can do. Though, is it OK if it will be in `/cluster` response *schema*? Currently, I've tweaked the code, so that we exclude metadata from `/cluster` (and removed it from schema). But if we add it to schema then it will be a bit confusing, because it will appear in `/cluster` *schema*, but not in the response.

I can also just move `metadata` from `PeerInfo` into a separate `ClusterTelemetry::metadata` field.

lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -234,6 +235,11 @@ impl<C: CollectionContainer> ConsensusManager<C> {
                     peer_id,
                     PeerInfo {
                         uri: uri.to_string(),
+                        metadata: if peer_metadata {
+                            metadata.get(&peer_id).cloned()
+                        } else {
+                            None
+                        },
```
timvisee: _Can_ do this

```suggestion
                        metadata: metadata.get(&peer_id).filter(|_| peer_metadata).cloned(),
```

ffuugoo: I like plain `if` more...

lib/storage/src/types.rs
ffuugoo: Moved into separate `ClusterTelemetry::peer_metadata` field.

PR #6912, opened 2025-07-21
Refactor token processing in `TokenizerConfig`

coszio:
Depends on #6891 

This is a nice-to-have refactor of token processing, which centralizes the processing of all tokenizer options in a single place which applies all the configured ones.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/tokenizers/mod.rs
+++ b/lib/segment/src/index/field_index/full_text_index/tokenizers/mod.rs
@@ -20,7 +20,7 @@ impl WhiteSpaceTokenizer {
         mut callback: C,
     ) {
         for token in text.split_whitespace() {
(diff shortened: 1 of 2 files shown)


lib/segment/src/index/field_index/full_text_index/tokenizers/mod.rs
```diff
@@ -212,59 +190,29 @@ impl Tokenizer {
...
-                return;
-            }
-            if self
-                .config
-                .max_token_len
-                .map(|max_len| token.len() > max_len && token.chars().count() > max_len)
-                .unwrap_or(false)
-            {
-                return;
-            }
```
coszio, line 235: The only functional thing that is being changed here is that this PR checks token length only with `token.chars().count()`, instead of checking `token.len()` first.

JojiiOfficial (approved):

PR #6916, opened 2025-07-21
UUID index bug

generall:
MmapHashMap had a problem with Entry alightment, in case if key value was u128. It required 16bit alightment, but
storage assumed, apparently, that nothing can need more than 8 bit alignment.

This PR introduces 2 unit tests + fix.
Fix is another alignment block inside MmapHashmap. If Required alignment is <8, there are no change and format exactly same as before. If required alignment >8, we put extra empty bytes __before__ bucket offsets, so that
(new_alignment + buckets) % ALIGNMENT = 0.

Compatibility: luckily (thanks to @xzfc) bucket offset is stored explicitly in the header of the file.
So the same code should be able to read both: files with and without offset, even if old files were created for u128 key.

### DIFF
--- a/lib/common/common/src/mmap_hashmap.rs
+++ b/lib/common/common/src/mmap_hashmap.rs
@@ -550,6 +550,9 @@ mod tests {
             assert_eq!(map.get(&from_ref(k)).unwrap(), &v);
         }
 
+        let keys: Vec<_> = mmap.keys().collect();
+        assert_eq!(keys.len(), map.len());
+
         // Existing keys should return the correct values
         for (k, v) in map {
             assert_eq!(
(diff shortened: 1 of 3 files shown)


coszio (approved):

generall:
> Old files without this fix create a panic on startup, which initializes the shard as dummy

this is not a new behavior, but it is a behavior of the debug build. Previous version also paniked, now it will panic(in debug) earlier, and print error messages in production.

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -98,8 +98,12 @@ impl<K: Key + ?Sized, V: Sized + FromBytes + Immutable + IntoBytes + KnownLayout
         file_size += padding_len;
 
         // 4. Buckets
+        let buckets_size = keys_count * size_of::<BucketOffset>();
+        let bucket_align = buckets_size % K::ALIGN;
```
xzfc: Nit: works for `u64` and `u128`, but won't work for hypothetical `K::ALIGN == 32`.

```suggestion
        let bucket_align = buckets_size.next_multiple_of(K::ALIGN) - buckets_size;
```

lib/common/common/src/mmap_hashmap.rs
```diff
@@ -134,6 +138,10 @@ impl<K: Key + ?Sized, V: Sized + FromBytes + Immutable + IntoBytes + KnownLayout
         bufw.write_zeros(padding_len)?;
 
         // 4. Buckets
+        let bucket_bytes = buckets.as_bytes();
+        let buckets_len = bucket_bytes.len();
+        // Align the buckets to `K::ALIGN`, to make sure Entry.key is aligned.
+        bufw.write_zeros(buckets_len % K::ALIGN)?;
```
xzfc: See comment above; and you don't need to re-calculate `bucket_align` again.

```suggestion
        bufw.write_zeros(bucket_align)?;
```

xzfc:
Also, not related to this particular PR, but it's better not to bake `align_of` into on-disk structures: `align_of` is platform and compiler version dependent. See https://blog.rust-lang.org/2024/03/30/i128-layout-update/.

I.e., in `lib/common/common/src/mmap_hashmap.rs`:

```diff
 impl Key for u128 {
-    const ALIGN: usize = align_of::<u128>();
+    const ALIGN: usize = size_of::<u128>();
```

PR #6923, opened 2025-07-22
MADV_POPULATE_READ on sequential mmaps

xzfc:
It turns out `MADV_POPULATE_READ` performance is influenced by the `MADV_RANDOM` and `MADV_SEQUENTIAL` flags.
Obvious in hindsight, but still surprising that the kernel won't enable readahead unconditionlly given that it knows that it will be reading the whole file.

```c
// Slow, many 4K IOPS
madvise(ptr, size, MADV_RANDOM);
madvise(ptr, size, MADV_POPULATE_READ);
```

```c
// Fast, as uses large readahead
madvise(ptr, size, MADV_SEQUENTIAL);
madvise(ptr, size, MADV_POPULATE_READ);
```

# Fix

Since a lot of our files mmaped twice (for sequential and for random), the fix in this PR is trivial: let `populate()` use the sequential mmap if it available.

Fixed:
- Dense vector storages, except deleted flags
- Gridstore pages (used in sparse vectors)

Some mmaps are not duplicated, so these are not fixed:
- `MmapBitSlice` (deleted flags).
- Gridstore bitmasks (`bitmask.dat` and `gaps.dat`; used in sparse vectors)
- Quantized vector storage. (not used)
- Payload storages. (not used)

CPU usage/saturation on 400M benchmark (left is before this commit, right is after):

### DIFF
--- a/lib/common/memory/src/chunked_utils.rs
+++ b/lib/common/memory/src/chunked_utils.rs
@@ -44,7 +44,7 @@ impl<T: Sized + 'static> UniversalMmapChunk<T> {
     }
 
     pub fn populate(&self) -> io::Result<()> {
-        self.mmap.populate()
+        self.mmap_seq.populate()
     }
 }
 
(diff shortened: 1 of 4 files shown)


agourlay:
Awesome work :clap: 

I spent some time investigating some page fault around `populate` a while back.
It might be somewhat related https://github.com/qdrant/qdrant/pull/6714

PR #6924, opened 2025-07-22
Avoid creating partial snapshot if snapshot manifest contains no changes

ffuugoo:
Small optimization for partial snapshots: instead of creating an "empty" partial snapshot, when snapshot manifest contains no changes (e.g., when snapshot manifest provided by "read" replica is the same as current data on "write" replica), we simply return `NOT MODIFIED` response and "short-circuit" the whole recovery.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/conversions.rs
+++ b/lib/storage/src/content_manager/conversions.rs
@@ -46,6 +46,7 @@ impl From<StorageError> for Status {
                 tonic::Code::ResourceExhausted
             }
             StorageError::ShardUnavailable { .. } => tonic::Code::Unavailable,
+            StorageError::EmptyPartialSnapshot { .. } => tonic::Code::FailedPrecondition,
         };
         let mut status = Status::new(error_code, format!("{error}"));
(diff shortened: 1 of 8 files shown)


lib/storage/src/content_manager/conversions.rs
```diff
@@ -46,6 +46,7 @@ impl From<StorageError> for Status {
                 tonic::Code::ResourceExhausted
             }
             StorageError::ShardUnavailable { .. } => tonic::Code::Unavailable,
+            StorageError::EmptyPartialSnapshot { .. } => tonic::Code::FailedPrecondition,
```
ffuugoo, line 49: Not sure which status would be the best here, but this error type should never be returned from gRPC anyway...

timvisee: I don't see a better option since we don't have a dedicated channel to return extra metadata from errors.

timvisee (approved):

PR #6926, opened 2025-07-22
[map index] use roaring bitmap in mutable map index

coszio:
Similar to #6685, swaps the container of ids per value in mutable map index from `BTreeSet<u32>` to `RoaringBitmap`.

This is not too noticeable on "normal" workloads, but when there is many values per point, it does provide a bit of advantage on insertion. 

For example, with the following command (100k points):

```
bfb -n 300000 --max-keywords 1000 --keywords 10000  --search --match-any 900
```

This is the performance improvement (edit: reran with release build of bfb and perf builds of qdrant)

| build | final upsert RPS | total upsert time | median search RPS | 
| --- | --- | --- | --- |
| dev (perf build) | 1685 | 00:02:57 | 526
| PR (perf build) | 3159 | 00:01:34 | 525

And, with just 1 keyword per point

```
bfb -n 300000 --keywords 10000 --search --match-any 900
```

| build | final upsert RPS | total upsert time | median search RPS | 
| --- | --- | --- | --- | 
| dev (perf build) | 147385 | 00:00:02 | 618
| PR (perf build) | 161625 | 00:00:01 | 628

### DIFF
--- a/lib/segment/src/index/field_index/map_index/immutable_map_index.rs
+++ b/lib/segment/src/index/field_index/map_index/immutable_map_index.rs
@@ -18,7 +18,7 @@ use rocksdb::DB;
 #[cfg(feature = "rocksdb")]
 use super::MapIndex;
 use super::mmap_map_index::MmapMapIndex;
-use super::{IdIter, IdRefIter, MapIndexKey};
+use super::{IdIter, MapIndexKey};
 use crate::common::Flusher;
 use crate::common::operation_error::{OperationError, OperationResult};
 #[cfg(feature = "rocksdb")]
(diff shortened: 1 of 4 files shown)


generall:
Could you please also include benchmark on "normal" workloads, even if they are not improving

coszio:
@generall I made a mistake when benchmarking. When compared to perf build and all feature flags on, the difference is not as dramatic. I have updated the description accordingly.

Still increases upsert performance.

agourlay (approved):

timvisee (approved):
👏

PR #6930, opened 2025-07-23
Do no delete payload index on point deletion to avoid data inconsistency

agourlay:
This PR fixes a subtle data inconsistency issue happening when crashing while a segment optimization is running.

When running with a proxy segment, the points deletion on the wrapped segment should not be executed until the proxy segment is promoted.

Otherwise the data is lost if crash occurs in between.

For this reason we use storages with lazy writes which are capturing pending changes and applying those only on explicit flush.

Importantly, a point deletion is also deleting the payload index values as well.

However some payload indexes are not buffering writes:
- `null-index`
- `mmap-bool-index`
- more maybe?

This PR proposes an easy workaround similar to the one what we did the vector deletion to avoid data inconsistency.

A better fix would be to make those payload index buffer writes but it might incur a large performance cost.

### DIFF
--- a/lib/segment/src/segment/entry.rs
+++ b/lib/segment/src/segment/entry.rs
@@ -129,7 +129,7 @@ impl SegmentEntry for Segment {
         &mut self,
         op_num: SeqNumberType,
         point_id: PointIdType,
-        hw_counter: &HardwareCounterCell,
+        _hw_counter: &HardwareCounterCell,
     ) -> OperationResult<bool> {
         let internal_id = self.id_tracker.borrow().internal_id(point_id);
         match internal_id {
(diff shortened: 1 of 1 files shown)


lib/segment/src/segment/entry.rs
```diff
@@ -138,19 +138,19 @@ impl SegmentEntry for Segment {
             Some(internal_id) => {
                 self.handle_point_version_and_failure(op_num, Some(internal_id), |segment| {
                     // Mark point as deleted, drop mapping
-                    segment
-                        .payload_index
-                        .borrow_mut()
-                        .clear_payload(internal_id, hw_counter)?;
                     segment.version_tracker.set_payload(Some(op_num));
```
coszio, line 141: Also don't set the payload version?

agourlay:
This approach requires to clean a bunch of tests which are relying on the fact that the payload is deleted.
Not a big fan of it TBH

timvisee:
Since the proxy segment itself already buffers deletes, are deletes in the wrapped segment actually ever called?

As far as I'm aware the deletions are only propagated at promotion, in which case I don't see a clear issue.

Could you potentially elaborate a bit more to help me understand?

agourlay:
This is too hacky. Closing while we find a better solution.

timvisee:
As far as I'm aware this problem is fixed now by other means because we now defer writes in these structures.

PR #6931, opened 2025-07-23
Allow health check endpoints to bypass authentication

craig-rueda:
## Summary
This PR removes the authentication requirement for gRPC health check endpoints, allowing them to be accessed without API keys or JWT tokens. This change enables proper health monitoring and load balancer checks in production environments where authentication credentials may not be available or appropriate (we're using ALB's in AWS and can't really add keys)

### Modified Files
- **`src/tonic/auth.rs`**: Updated the authentication middleware to bypass validation for health check endpoints

### Implementation Details
- Added path-based routing logic in the `check()` function to identify health check requests
- Health check endpoints now bypass the standard `auth_keys.validate_request()` flow
- Endpoints are granted full access with appropriate `Access` and `InferenceToken` extensions
- Early return prevents unnecessary authentication processing for these endpoints

## Affected Endpoints
The following gRPC endpoints now work without authentication:

1. **Qdrant-specific health check**: `/qdrant.Qdrant/HealthCheck`
2. **Standard gRPC health protocol**: `/grpc.health.v1.Health/Check`

## Testing
To test these changes:

```bash
# Build and run Qdrant locally
cargo build --release --bin qdrant
./target/release/qdrant

# Test health check endpoints without authentication
grpcurl -plaintext localhost:6334 qdrant.Qdrant/HealthCheck
grpcurl -plaintext localhost:6334 grpc.health.v1.Health/Check

# Verify other endpoints still require authentication
grpcurl -plaintext localhost:6334 qdrant.Collections/List  # Should fail without auth
```

### DIFF
--- a/src/tonic/auth.rs
+++ b/src/tonic/auth.rs
@@ -19,6 +19,19 @@ pub struct AuthMiddleware<S> {
 }
 
 async fn check(auth_keys: Arc<AuthKeys>, mut req: Request) -> Result<Request, Status> {
+    // Allow health check endpoints to bypass authentication
+    let path = req.uri().path();
+    if path == "/qdrant.Qdrant/HealthCheck" || path == "/grpc.health.v1.Health/Check" {
+        // Set default full access for health check endpoints
+        let access = Access::full("Health check endpoints have full access without authentication");
+        let inference_token = crate::common::inference::InferenceToken(None);
+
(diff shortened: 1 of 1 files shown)


craig-rueda:
Thanks, @generall. Any idea when this will be merged / released?

generall:
> Thanks, @generall. Any idea when this will be merged / released?

1.15.1

PR #6932, opened 2025-07-23
Fix PointId Hash

generall:
UUID crate changed Hash implementation, which silently broke our `HashRing` routing

https://github.com/qdrant/qdrant/pull/6595

> Replace derive(Hash) with manual impl in Uuid by @​diopoex in uuid-rs/uuid#824

This PR adds manual implementation of Hash for `ExtendedPointId` and unit-test which pins shard mapping.

### DIFF
--- a/lib/collection/src/operations/point_ops.rs
+++ b/lib/collection/src/operations/point_ops.rs
@@ -950,7 +950,7 @@ mod tests {
         let points: Vec<_> = all_ids
             .iter()
             .map(|id| PointStructPersisted {
-                id: id.clone(),
+                id: *id,
                 vector: VectorStructPersisted::from(vec![0.1, 0.2, 0.3]),
                 payload: None,
             })


lib/segment/src/types.rs
```diff
@@ -155,6 +155,20 @@ pub enum ExtendedPointId {
     Uuid(Uuid),
 }
 
+impl Hash for ExtendedPointId {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        // WARN: Changing this implementation might break backward compatibility
+        // Of hash-ring and shard routing
+        let __self_discr = std::mem::discriminant(self);
```
generall, line 162: I have doubts about this: I used "Expand Macro" from Rust Playground, but it generates nightly. This version is copilot suggestion of how to fix Nightly build

xzfc (approved):

lib/segment/src/types.rs
xzfc: Rust 1.87 generates the same (`std::mem::discriminant`).

PR #6935, opened 2025-07-24
Appendable quantization storage

IvanPleshkov:
This PR adds `push_vector` into quantization storage to make it appendable.
Also, it deprecates `count` in quantization config.
All changes are forward and back compatible with `v1.15`.

### DIFF
--- a/lib/quantization/src/encoded_vectors_binary.rs
+++ b/lib/quantization/src/encoded_vectors_binary.rs
@@ -408,7 +408,7 @@ impl<TBitsStoreType: BitsStoreType, TStorage: EncodedStorage>
         Self {
             encoded_vectors,
             metadata: Metadata {
-                vector_parameters: vector_parameters.clone(),
+                vector_parameters,
                 encoding,
                 query_encoding,
                 vector_stats: None,


generall:
Could you please explain why `count` is deprecated?

IvanPleshkov:
> Could you please explain why `count` is deprecated?

Sure! If we keep this count, we need to rewrite config each flush or vector push. For `ChunkedVectors<u8>` and for `Mmap` there is only one source of true - container size. For upcoming PR with using `ChunkedMmap` as a quantization storage, the size is `ChunkedMmap::len`, not a config value.

xzfc:
Do we need to persist these newly pushed vectors?

If so, how to persist them? I assume the intended usage is to call `EncodedStorage::save_to_file()` after pushing. Am I right? (I expect to see write/append to file somewhere in the implementation)

lib/quantization/src/encoded_vectors.rs
```diff
@@ -15,9 +15,14 @@ pub enum DistanceType {
...
     pub dim: usize,
-    pub count: usize,
     pub distance_type: DistanceType,
     pub invert: bool,
+
+    // Deprecated, use `EncodedVectors::vectors_count` from quantization instead.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "count")]
+    pub deprecated_count: Option<usize>,
```
xzfc, line 25: Don't we still need to write this field for forward compatibility? I.e. Qdrant 1.15.1 would want to read it from `quantized.meta.json` written by Qdrant 1.16+.

If so, then this shouldn't be an `Option`.

generall: ~It should also not be renamed~

I see there is an alias

IvanPleshkov:
> Do we need to persist these newly pushed vectors?
> 
> If so, how to persist them? I assume the intended usage is to call `EncodedStorage::save_to_file()` after pushing. Am I right? (I expect to see write/append to file somewhere in the implementation)

Not in this PR. Next PR is using `ChunkedMmap` as a quantization storage and all required persists are already in `ChunkedMmap`, will also add `flush` and `files` methods in as a next step

lib/quantization/src/encoded_vectors.rs
IvanPleshkov: For non-appendable segments this field has `Some` with correct value, for non-appendable segments, where we want to add quantization, this field is `None` to avoid problems with persistent

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -399,6 +399,24 @@ impl<TBitsStoreType: BitsStoreType, TStorage: EncodedStorage>
...
+    pub fn new_empty_appendable(
+        encoded_vectors: TStorage,
+        vector_parameters: VectorParameters,
+        encoding: Encoding,
+        query_encoding: QueryEncoding,
+    ) -> Self {
+        Self {
+            encoded_vectors,
+            metadata: Metadata {
+                vector_parameters: vector_parameters.clone(),
```
xzfc: Nit: redundant clone.
```suggestion
                vector_parameters,
```

lib/segment/src/vector_storage/chunked_vectors.rs
```diff
@@ -176,20 +177,19 @@ impl quantization::EncodedStorage for ChunkedVectors<u8> {
...
+    fn from_file(path: &Path, quantized_vector_size: usize) -> std::io::Result<Self> {
         let mut vectors = Self::new(quantized_vector_size);
-        vectors
-            .try_set_capacity_exact(vectors_count)
-            .map_err(|err| {
-                std::io::Error::new(
-                    std::io::ErrorKind::OutOfMemory,
-                    format!("Failed to load quantized vectors from file: {err}"),
-                )
-            })?;
```
xzfc, line 192: Why this code is removed?

IvanPleshkov: Vectors count is unknown now while loading, because we cannot rely on config value anymore.
But it's not a problem here, `ChunkedVectors` do an allocation for the whole chunk size while `push`. It means, the amount of allocation is the same as before

IvanPleshkov:
> Do we need to persist these newly pushed vectors?
> 
> If so, how to persist them? I assume the intended usage is to call `EncodedStorage::save_to_file()` after pushing. Am I right? (I expect to see write/append to file somewhere in the implementation)

Quantization is implemented for `Vec<u8>` and `ChunkedVectors<u8>`, they both don't have persistence. Persistency is a separate PR https://github.com/qdrant/qdrant/pull/6986/files, where quantization is implemented for `ChunkedMmapVectors<u8>`

PR #6904, opened 2025-07-18
[WIP] Execute Bm25 embedding in inference

JojiiOfficial:
Depends on #6893

WIP

PR #6939, opened 2025-07-25
[Bm25] Execution implementation

JojiiOfficial:
Supersedes #6893 and #6904
Depends on https://github.com/qdrant/qdrant/pull/6891
Implements Bm25 by parsing the inference object's options.

## Request example
A possible request could look like this:

```json
{
  "query": {
    "model": "qdrant/bm25",  // Must be set but value can be arbitrary.
    "text": "Some input to embedd with bm25",
    "options": {
      "use_bm25": true,        // Required to parse this options as BM25 configuration.
      "k": 1.0,                // Optional hyperparameters. Same for `b` and `avg_len`.
      "tokenizer": "word",     // Optional 
      "lowercase": true       // Optional
    }
  },
  "using": "sparse"
}
```

## Error handling
In case the option "use_bm25" is enabled but the configuration is invalid, a detailed error is returned:
```
{
  "error": "Wrong input: Invalid BM25 config: Error(\"unknown variant `ord`, expected one of `prefix`, `whitespace`, `word`, `multilingual`\", line: 0, column: 0)"
}
```

## Test coverage
There are also tests for the new BM25 and remote-inference logic that almost cover all parts of `infer()` and >97% of bm25.rs. "Almost" because returning errors are not covered.
These tests 'mock' an inference server and ensure that we correctly merge bm25 and remote items together.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,8 @@ prometheus = { version = "0.14.0", default-features = false }
 validator = { workspace = true }
 jsonwebtoken = "9.3.1"
 
+murmur3 = { git = "https://github.com/stusmall/murmur3", rev = "2c39087" }
+
 tempfile = { workspace = true }
 
 # Consensus related crates
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,8 @@ sealed_test = "1.1.0"
 tempfile = { workspace = true }
 rusty-hook = "^0.11.2"
 
+mockito = "1.7"
+
 
 [dependencies]
 parking_lot = { workspace = true }
--- a/lib/collection/src/operations/point_ops.rs
+++ b/lib/collection/src/operations/point_ops.rs
@@ -17,6 +17,7 @@ use segment::data_types::vectors::{
 };
 use segment::types::{Filter, Payload, PointIdType, VectorNameBuf};
 use serde::{Deserialize, Serialize};
+use sparse::common::types::{DimId, DimWeight};
 use strum::{EnumDiscriminants, EnumIter};
 use validator::{Validate, ValidationErrors};
 
@@ -59,7 +60,7 @@ pub enum VectorPersisted {
 }
 
 impl VectorPersisted {
-    pub fn new_sparse(indices: Vec<u32>, values: Vec<f32>) -> Self {
+    pub fn new_sparse(indices: Vec<DimId>, values: Vec<DimWeight>) -> Self {
         Self::Sparse(sparse::common::sparse_vector::SparseVector { indices, values })
     }
 
--- a/lib/collection/src/operations/point_ops.rs
+++ b/lib/collection/src/operations/point_ops.rs
@@ -58,6 +58,16 @@ pub enum VectorPersisted {
     MultiDense(MultiDenseVector),
 }
 
+impl VectorPersisted {
+    pub fn new_sparse(indices: Vec<u32>, values: Vec<f32>) -> Self {
+        Self::Sparse(sparse::common::sparse_vector::SparseVector { indices, values })
+    }
+
+    pub fn empty_sparse() -> Self {
+        Self::new_sparse(vec![], vec![])
+    }
+}
+
 impl Debug for VectorPersisted {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
--- a/src/common/inference/bm25.rs
+++ b/src/common/inference/bm25.rs
@@ -1,0 +1,147 @@
+use std::borrow::Cow;
+use std::collections::{BTreeMap, HashMap};
+
+use collection::operations::point_ops::VectorPersisted;
+use murmur3::murmur3_32_of_slice;
+use segment::data_types::index::{StemmingAlgorithm, StopwordsInterface, TokenizerType};
+use segment::index::field_index::full_text_index::stop_words::StopwordsFilter;
+use segment::index::field_index::full_text_index::tokenizers::{
+    Stemmer, Tokenizer, TokenizerConfig,
+};
+use serde::{Deserialize, Serialize};
+
+/// Configuration of the local bm25 models.
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct Bm25Config {
+    #[serde(default = "default_model_name")]
+    pub model_name: String,
+    #[serde(default = "default_k")]
+    pub k: f64,
+    #[serde(default = "default_b")]
+    pub b: f64,
+    #[serde(default = "default_avg_len")]
+    pub avg_len: f64,
+    #[serde(default)]
+    pub tokenizer: TokenizerType,
+    #[serde(default)]
+    pub tokenizer_config: Option<InferenceTokenizerConfig>,
+}
+
+/// Bm25 tokenizer configurations.
+#[derive(Debug, Deserialize, Serialize, Clone, Default)]
+pub struct InferenceTokenizerConfig {
+    #[serde(default)]
+    pub lowercase: bool,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub stopwords_filter: Option<StopwordsInterface>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub stemmer: Option<StemmingAlgorithm>,
+    pub min_token_len: Option<usize>,
+    pub max_token_len: Option<usize>,
+}
+
+fn default_model_name() -> String {
+    "bm25".to_string()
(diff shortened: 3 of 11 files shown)


src/common/inference/service.rs
```diff
@@ -282,3 +384,218 @@ impl InferenceService {
...
+            .enumerate()
+            .filter(|(_, item)| item.try_parse_bm25_config().ok().flatten().is_none())
+            .map(|(index, _)| {
+                let values = vec![0.0; index];
+                VectorPersisted::Dense(values)
+            })
+            .collect();
+
+        // Create an HTTP mock
+        let mock = server
```
JojiiOfficial, line 470: Mocking an inference server here to fully end2end test the logic of this PR.

Cargo.toml
```diff
@@ -93,6 +95,8 @@ prometheus = { version = "0.14.0", default-features = false }
 validator = { workspace = true }
 jsonwebtoken = "9.3.1"
 
+murmur3 = { git = "https://github.com/stusmall/murmur3", rev = "2c39087" }
```
generall, line 98: Should we fork it?

src/common/inference/bm25.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[serde(default = "default_k")]
+    pub k: f64,
+    #[serde(default = "default_b")]
+    pub b: f64,
+    #[serde(default = "default_avg_len")]
+    pub avg_len: f64,
+    #[serde(default)]
+    pub tokenizer: TokenizerType,
+    #[serde(default)]
+    pub tokenizer_config: Option<InferenceTokenizerConfig>,
```
generall: I don't think it should be called `tokenizer_config`, in fact, I don't think it should be called anything. I would prefer to use `#[serde(flatten)]` here.

src/common/inference/bm25.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/// Bm25 tokenizer configurations.
+#[derive(Debug, Deserialize, Serialize, Clone, Default)]
+pub struct InferenceTokenizerConfig {
```
generall: Stemmer is not part of tokenizer. Let's call it TextPreProcessingConfig, it will be an internal name anyway

lib/collection/src/operations/point_ops.rs
```diff
@@ -58,6 +58,16 @@ pub enum VectorPersisted {
     MultiDense(MultiDenseVector),
 }
 
+impl VectorPersisted {
+    pub fn new_sparse(indices: Vec<u32>, values: Vec<f32>) -> Self {
```
generall: There are special types: DimId and DimWeight

src/common/inference/service.rs
```diff
@@ -42,15 +45,37 @@ pub struct InferenceRequest {
...
+impl InferenceInput {
+    /// Tries to parse the input's options into BM25 config and returns Ok(Some(..)) if found and Ok(None) if not.
+    ///
+    /// Returns an error if bm25 has been explicitly enabled but the options could not be deserialized properly into bm25 config.
+    pub fn try_parse_bm25_config(&self) -> Result<Option<Bm25Config>, StorageError> {
+        let Some(options) = self.options.as_ref() else {
+            return Ok(None);
+        };
+
+        if options.get("use_bm25") != Some(&Value::Bool(true))
```
generall: Why is this here and why it has hard-coded strings?

src/common/inference/service.rs
```diff
@@ -180,24 +205,72 @@ impl InferenceService {
         inference_type: InferenceType,
         inference_token: InferenceToken,
     ) -> Result<InferenceResponse, StorageError> {
+        let (bm25_inference_inputs, inference_inputs): (Vec<_>, Vec<_>) = inference_inputs
```
generall: I would prefer to keep `infer` function agnostic to specialized implementations.

Instead of splitting inputs into `bm25_inference_inputs, inference_inputs` I propose to split into `local_inference_inputs, remote_inference_inputs`.

Also, I propose to move all bm25 config parsing and inference into a separate file.

src/common/inference/service.rs
```diff
@@ -222,7 +295,40 @@ impl InferenceService {
...
+
+        let remote_res = Self::handle_inference_response(status, &response_body)?;
+        Ok(Self::merge_bm25_and_remote_result(
+            bm25_results,
+            remote_res,
+            remote_pos,
+        ))
+    }
+
+    fn merge_bm25_and_remote_result(
```
generall: same, avoid bm25 specifics here, more Bm25-related code in a dedicated file

src/common/inference/service.rs
```diff
@@ -180,24 +205,72 @@ impl InferenceService {
...
-            token,
-        };
-
         let url = self.config.address.as_ref().ok_or_else(|| {
             StorageError::service_error(
                 "InferenceService URL not configured - please provide valid address in config",
             )
         })?;
 
+        let (remote_pos, inference_inputs): (Vec<_>, Vec<_>) = inference_inputs
```
generall: this can be part of ` local_inference_inputs, remote_inference_inputs` split

generall (changes requested):

Cargo.toml
JojiiOfficial: I don't think that would have notable advantages. We'd have to manually update the fork when updating.
The reason it's pinned to a commit is that it doesn't have any tagged versions. Pinning or forking doesn't make much difference I believe.

src/common/inference/service.rs
JojiiOfficial: You're right, it should be a constant instead of directly coded into the code.

I think it's better to explicitly state that we want to use bm25, instead of "guessing" and always trying to parse the `options` into Bm25Config. Therefore we have this additional check when trying to read BM25 configuration from the inputs `options`.

Alternatively, we could rename it to something like "custom_model" or "local_model" and check the value against "bm25". This makes it more extendible for other 'local' model implementations in future.

src/common/inference/bm25.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    /// Embedds the given input using the Bm25 algorithm and configured options/hyperparameters.
+    pub fn search_embed(&self, input: &str) -> VectorPersisted {
+        let tokens = self.tokenize(input);
+
+        if tokens.is_empty() {
+            return VectorPersisted::empty_sparse();
+        }
+
+        let indices: Vec<u32> = tokens
```
generall, line 97: `DimId`?

src/common/inference/bm25.rs
```diff
@@ -0,0 +1,163 @@
...
+            .into_iter()
+            .map(|token| Self::compute_token_id(&token))
+            .unique()
+            .collect();
+
+        let values: Vec<f32> = vec![1.0; indices.len()];
```
generall, line 103: `DimWeight` ?

src/common/inference/inference_input.rs
```diff
@@ -0,0 +1,108 @@
...
+            .map_err(|err| StorageError::bad_input(format!("Invalid BM25 config: {err:#?}")))
+    }
+
+    /// Attempts to parse the input's options into a local model config.
+    pub fn try_parse_local_model_input(&self) -> Result<Option<LocalModelConfig>, StorageError> {
+        let Some(options) = self.options.as_ref() else {
+            return Ok(None);
+        };
+
+        let Some(local_model_field) = options.get(LOCAL_MODEL_KEY) else {
```
generall: I still don't understand why this is needed

JojiiOfficial: If we add a different local 'model', next to bm25, which also has tokenizer configs and a user only specifies those (so no Bm25 specific parameters), we can't distinguish which model should be used.
This requires the user to specify which local model to use.

We could use some dedicated model name e.g. 'local/bm25' to specify which local model to use. However I thought hard-coding some specific model name to be potentially confusing, especially if something like 'local/bm25' works but 'bm25' throws an error about some 'address' not being found. If you prefer this approach though, I'm totally fine with changing it.

src/common/inference/bm25.rs
JojiiOfficial: We don't have access to `DimId` and `DimWeight` in `src/common`. I don't think we should add a dependency, just for this type alias.
If you prefer having some sort of alias here, I can add a new type alias here, but I think the semantics is already known from the variable names and context from the function.

The same applies to the other place in line 103

Cargo.toml
generall: ok

generall (approved):

PR #6940, opened 2025-07-25
Introduce StableHash

xzfc:
Follow-up for #6932. A cleaned up/more explicit version.

For hashes to be stable, we need two things: a stable `Hash` implementation for our types, and a stable `Hasher` implementation.

The default `Hasher` used in the `hashring` crate is `SipHash24`. This PR makes it explicit.

We can't assume that `Hash` is stable neither for types from `std` [\[1\]](https://doc.rust-lang.org/std/hash/trait.Hash.html#portability), nor from third-party crates. Thus, this PR implements hashes by hand via the new trait `StableHash`, without depending on stdandard or third-party implementations. The new `StableHashed` adapter is provided to allow using the `StableHash` implementation in contexts where `Hash` is expected, e.g., in the `hashring` crate. The default `Hash` implementation is kept unstable, enabling whatever optimizations the `uuid` crate (or future Rust release) implements in contexts where `StableHash` is not used.

These changes do not affect produced hashes used in the hashring. The test `split_point_operations` ensures that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1168,6 +1168,7 @@ dependencies = [
  "async-trait",
  "atomicwrites",
  "bitvec",
+ "bytemuck",
  "bytes",
  "cancel",
  "chrono",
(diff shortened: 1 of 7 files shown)


generall (approved):
LGTM

PR #6945, opened 2025-07-28
QueryScorerBytes

xzfc:
This PR implements scorers required for the vectors-in-graph feature.

It introduces a new trait `QueryScorerBytes` that allows scoring encoded query against vector encoded as `&[u8]`.

This new trait is integrated enough to provide two variations of `FilteredScorer`:
- One (old one) is regular `FilteredScorer` that is used to score against PointOffsetType (retrieved from the vector storage). No changes in logic here.
- Another (new one) is `FilteredQuantizedScorer` to score against vector encoded as `&[u8]`. Later these vectors would be stored alongside graph links.

Supported quantizations: binary, scalar, and product.
Supported only singular vectors, not multi-vectors.

### DIFF
--- a/lib/segment/src/index/hnsw_index/point_scorer.rs
+++ b/lib/segment/src/index/hnsw_index/point_scorer.rs
@@ -12,6 +12,7 @@ use crate::payload_storage::FilterContext;
 use crate::vector_storage::common::VECTOR_READ_BATCH_SIZE;
 use crate::vector_storage::quantized::quantized_query_scorer::InternalScorerUnsupported;
 use crate::vector_storage::quantized::quantized_vectors::QuantizedVectors;
+use crate::vector_storage::query_scorer::QueryScorerBytes;
 use crate::vector_storage::{
     RawScorer, VectorStorage, VectorStorageEnum, check_deleted_condition, new_raw_scorer,
 };
@@ -36,8 +37,12 @@ use crate::vector_storage::{
 ///                                                              - Scoring logic
 ///                                                              - Complex queries
 /// ```
-pub struct FilteredScorer<'a> {
-    raw_scorer: Box<dyn RawScorer + 'a>,
+pub type FilteredScorer<'a> = FilteredScorerImpl<'a, Box<dyn RawScorer + 'a>>;
+
+pub type FilteredQuantizedScorer<'a> = FilteredScorerImpl<'a, Box<dyn QueryScorerBytes + 'a>>;
+
+pub struct FilteredScorerImpl<'a, TScorer> {
+    raw_scorer: TScorer,
     filter_context: Option<BoxCow<'a, dyn FilterContext + 'a>>,
     /// Point deleted flags should be explicitly present as `false`
     /// for each existing point in the segment.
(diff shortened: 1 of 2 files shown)


lib/quantization/src/encoded_vectors.rs
```diff
@@ -52,6 +52,15 @@ pub trait EncodedVectors: Sized {
...
 }
 
+pub trait EncodedVectorsBytes: EncodedVectors {
+    fn score_point_vs_bytes(
+        &self,
+        query: &Self::EncodedQuery,
+        bytes: &[u8],
+        hw_counter: &HardwareCounterCell,
+    ) -> f32;
+}
```
xzfc, line 62: We have a bunch of methods in the `EncodedVectors` trait, but not all quantizations implement all of the methods; some just `panic!` for unsupported methods. I don't like this approach, and for this new method I took a different one.

Instead of adding a new, potentially panicking method, I've added a new trait. Not all quantizations implement this trait, but those that do don't panic.

Not completely happy with this approach either, but let's see how would it go.

lib/segment/src/vector_storage/query_scorer/mod.rs
```diff
@@ -37,6 +37,10 @@ pub trait QueryScorer {
     fn score_internal(&self, point_a: PointOffsetType, point_b: PointOffsetType) -> ScoreType;
 }
 
+pub trait QueryScorerBytes {
+    fn score_bytes(&self, bytes: &[u8]) -> ScoreType;
+}
```
xzfc, line 42: We have a bunch of methods in the `QueryScorer` trait, but not all scorers implement all of the methods; some just `panic!` for unsupported methods. I don't like this approach, and for this new method I took a different one.

Instead of adding a new, potentially panicking method, I've added a new trait. Not all scorers implement this trait, but those that do don't panic.

Not completely happy with this approach either, but let's see how would it go.

IvanPleshkov: The basic idea here is to use the custom provided data instead of `VectorStorage`. I propose to make it generic and implement this trait also for regular dense vectors.

lib/segment/src/vector_storage/quantized/quantized_scorer_builder.rs
```diff
@@ -130,6 +178,41 @@ impl<'a> QuantizedScorerBuilder<'a> {
...
+            QuantizedVectorStorage::PQMmap(storage) => {
+                self.new_quantized_scorer2::<TElement, TMetric>(storage)
+            }
+            QuantizedVectorStorage::BinaryRam(storage) => {
+                self.new_quantized_scorer2::<TElement, TMetric>(storage)
+            }
+            QuantizedVectorStorage::BinaryMmap(storage) => {
+                self.new_quantized_scorer2::<TElement, TMetric>(storage)
+            }
+            QuantizedVectorStorage::ScalarRamMulti(_)
```
IvanPleshkov, line 207: Why do we skip multivector support? Is it temporary? In case of multivectors, even if we need to make additional disk reads, a new upcoming feature may be more efficient than search in mmap vector storage using regular hnsw

lib/segment/src/index/hnsw_index/point_scorer.rs
```diff
@@ -36,8 +37,12 @@ use crate::vector_storage::{
 ///                                                              - Scoring logic
 ///                                                              - Complex queries
 /// ```
-pub struct FilteredScorer<'a> {
-    raw_scorer: Box<dyn RawScorer + 'a>,
+pub type FilteredScorer<'a> = FilteredScorerImpl<'a, Box<dyn RawScorer + 'a>>;
+
+pub type FilteredQuantizedScorer<'a> = FilteredScorerImpl<'a, Box<dyn QueryScorerBytes + 'a>>;
```
IvanPleshkov, line 42: Can we redesign RawScorer to avoid this definition? We did a great job of keeping all our abstractions in the strong abstraction layer and breaking them here. We don't have here a separate definition for multivectors, for Discovery/Context search, quantizations, etc. Even if the price of such abstraction is a bunch of functions with panics, it still keeps all abstraction layers together and allows here to forget about the kinds of scorers

IvanPleshkov: I find this topic critical enough to place merge blocker

IvanPleshkov (changes requested):

lib/segment/src/vector_storage/query_scorer/mod.rs
IvanPleshkov: Not necessary, just an idea if it's easy to implement. Not necessary, just for the art of the qdrant design =)

lib/segment/src/index/hnsw_index/point_scorer.rs
IvanPleshkov: I read from the description that it's a temporary solution. Are you going to fully remove it after feature implementation?

lib/segment/src/index/hnsw_index/point_scorer.rs
```diff
@@ -50,7 +55,55 @@ pub struct FilteredScorer<'a> {
...
+        Ok(FilteredQuantizedScorer {
+            raw_scorer: query_scorer,
+            filter_context,
+            point_deleted,
+            vec_deleted: vectors.deleted_vector_bitslice(),
+            scores_buffer: Vec::new(),
+        })
+    }
+
+    pub fn score_points(
```
generall, line 89: Is this function used? I can't find any references

lib/segment/src/vector_storage/quantized/quantized_scorer_builder.rs
```diff
@@ -201,6 +284,77 @@ impl<'a> QuantizedScorerBuilder<'a> {
         }
     }
 
+    fn new_quantized_scorer2<TElement, TMetric>(
```
generall: What's the difference with `new_quantized_scorer`? (maybe requires a commend and better naming)

lib/segment/src/index/hnsw_index/point_scorer.rs
IvanPleshkov: @generall need your help here. I guess we want to have support for new scoring in the `FilteredScorer` for exact search/low cardinality search etc, and won't remove this even after finishing the milestone.

generall: I don't have a strong opinion, please resolve among yourself.

IvanPleshkov: After the discussion, we decided to keep it as is because there is no better idea

PR #6946, opened 2025-07-28
Clean up `GraphLayers::search_entry_on_level` a bit

xzfc:


lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -124,55 +124,33 @@ pub trait GraphLayersBase {
...
+                self.search_entry_on_level(current_point, level, points_scorer, &mut links_buffer);
         }
         Ok(current_point)
     }
 
-    #[cfg(test)]
-    #[cfg(feature = "gpu")]
     fn search_entry_on_level(
         &self,
-        entry_point: PointOffsetType,
```
generall, line 167: It seems to me like the original version is more intuitive

lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -124,55 +124,33 @@ pub trait GraphLayersBase {
...
+        entry_point: ScoredPointOffset,
         level: usize,
         points_scorer: &mut FilteredScorer,
+        links: &mut Vec<PointOffsetType>,
```
generall, line 158: This parameter requires a comment.

lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -124,55 +124,33 @@ pub trait GraphLayersBase {
         points_scorer: &mut FilteredScorer,
         is_stopped: &AtomicBool,
     ) -> CancellableResult<ScoredPointOffset> {
-        let mut links: Vec<PointOffsetType> = Vec::with_capacity(2 * self.get_m(0));
-
+        let mut links_buffer = Vec::new();
```
generall, line 130: Why not pre-allocating is better?

generall (changes requested):
This PR needs more explanations

lib/segment/src/index/hnsw_index/graph_layers.rs
IvanPleshkov: Because this function is a wrapper now, instead of greedy search, we do a call of nested `search_entry_on_level`, where we do preallocation: `links.reserve(limit);`.
But I see the preallocated size is different now.

lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -124,55 +124,33 @@ pub trait GraphLayersBase {
...
-                    });
-            }
+            current_point =
+                self.search_entry_on_level(current_point, level, points_scorer, &mut links_buffer);
         }
         Ok(current_point)
     }
 
-    #[cfg(test)]
-    #[cfg(feature = "gpu")]
```
IvanPleshkov, line 164: Good catch to use test function as a part of a real one!!

lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -124,55 +124,33 @@ pub trait GraphLayersBase {
...
     ) -> ScoredPointOffset {
         let limit = self.get_m(level);
-        let mut links: Vec<PointOffsetType> = Vec::with_capacity(2 * self.get_m(0));
-        let mut current_point = ScoredPointOffset {
-            idx: entry_point,
-            score: points_scorer.score_point(entry_point),
-        };
+
+        links.clear();
+        links.reserve(limit);
```
IvanPleshkov: Why do we change the preallocated size? I see that filling of `links` is done by `links_map`, which may give more than `M` values in case of filtered search - we apply filter later. I don't remember why we use such estimation, but I propose to keep it as is

IvanPleshkov: Mark PR as "request changes" because of this topic

IvanPleshkov (changes requested):

lib/segment/src/index/hnsw_index/graph_layers.rs
IvanPleshkov: added

lib/segment/src/index/hnsw_index/graph_layers.rs
IvanPleshkov: reverted

lib/segment/src/index/hnsw_index/graph_layers.rs
IvanPleshkov: fixed

generall (approved):

PR #6952, opened 2025-07-29
Fix `DateTimeWrapper` parsing

ffuugoo:
`DateTimeWrapper` could not parse it's own `to_string` output. Oops. Fixed now.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/operations/universal_query/formula.rs
+++ b/lib/collection/src/operations/universal_query/formula.rs
@@ -81,11 +81,13 @@ impl ExpressionInternal {
                 ParsedExpression::new_geo_distance(origin, to)
             }
             ExpressionInternal::Datetime(dt_str) => {
-                ParsedExpression::Datetime(DatetimeExpression::Constant(
-                    dt_str
-                        .parse()
-                        .map_err(|err: chrono::ParseError| err.to_string())?,
-                ))
+                ParsedExpression::Datetime(DatetimeExpression::Constant(dt_str.parse().map_err(
+                    |err: chrono::ParseError| {
+                        CollectionError::bad_input(format!(
+                            "failed to parse date-time {dt_str:?}: {err}"
+                        ))
+                    },
+                )?))
             }
             ExpressionInternal::DatetimeKey(json_path) => {
                 payload_vars.insert(json_path.clone());
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -3510,6 +3510,13 @@ mod tests {
     }
 
     #[test]
+    fn test_datetime_wrapper_transcoding() {
+        let expected = DateTimeWrapper(chrono::Utc::now());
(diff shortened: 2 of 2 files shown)


lib/collection/src/operations/universal_query/formula.rs
```diff
@@ -81,11 +81,13 @@ impl ExpressionInternal {
...
-                        .parse()
-                        .map_err(|err: chrono::ParseError| err.to_string())?,
-                ))
+                ParsedExpression::Datetime(DatetimeExpression::Constant(dt_str.parse().map_err(
+                    |err: chrono::ParseError| {
+                        CollectionError::bad_input(format!(
+                            "failed to parse date-time {dt_str:?}: {err}"
+                        ))
+                    },
+                )?))
```
ffuugoo, line 90: This change is optional, but would have been nice to have better errors during parsing. Would have trivialized solving this issue even more.

lib/segment/src/types.rs
```diff
@@ -99,6 +99,8 @@ impl FromStr for DateTimePayloadType {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         // Attempt to parse the input string in RFC 3339 format
         if let Ok(datetime) = chrono::DateTime::parse_from_rfc3339(s)
+            // Attempt to parse default to-string format
+            .or_else(|_| chrono::DateTime::from_str(s))
```
ffuugoo, line 103: According to [`chrono` documentation][chrono-date-time-from-str], `FromStr` implementation:

> Accepts a relaxed form of RFC3339. A space or a ‘T’ are accepted as the separator between the date and time parts. Additional spaces are allowed between each component.

Which seems exactly what we want here? Maybe we can even put this as the first parser we try, before even `parse_from_rfc3339`.

[chrono-date-time-from-str]: https://docs.rs/chrono/latest/chrono/struct.DateTime.html#impl-FromStr-for-DateTime%3CFixedOffset%3E

lib/segment/src/types.rs
```diff
@@ -99,6 +99,8 @@ impl FromStr for DateTimePayloadType {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         // Attempt to parse the input string in RFC 3339 format
         if let Ok(datetime) = chrono::DateTime::parse_from_rfc3339(s)
```
timvisee, line 101: How about also using [`to_rfc3339`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.to_rfc3339) at the time we convert it into a string?

ffuugoo: Sure. But tbh I prefer `to_string` output. More readable. 😂

timvisee: I don't see any recommendations stated from chrono for transferring a date time over the wire. So let's keep it as is :+1:

timvisee (approved):

PR #6954, opened 2025-07-29
New mutable null index

coszio:
Current null index does not buffer updates at all, so there is no synchronization as to when data is written to disk.

This mutable null index keeps in-memory values in a roaring bitmap for efficient reads, and stores in a buffered dynamic mmap flags.


lib/segment/src/common/buffered_dynamic_flags.rs
```diff
@@ -0,0 +1,573 @@
...
+
+    /// Pending changes to the storage flags.
+    buffer: Arc<RwLock<AHashMap<PointOffsetType, bool>>>,
+
+    /// In-memory bitmap of true flags.
+    // Potential optimization: add a secondary bitmap for false values for faster iter_falses implementation.
+    bitmap: RoaringBitmap,
+
+    /// Total length of the flags, including the trailing ones which have been set to false
+    cached_len: usize,
```
timvisee: Let's just rename this to `len`. Wdyt?

lib/segment/src/common/buffered_dynamic_flags.rs
```diff
@@ -0,0 +1,573 @@
...
+
+    pub fn files(&self) -> Vec<PathBuf> {
+        self.storage.lock().files()
+    }
+
+    pub fn flusher(&self) -> Flusher {
+        // take pending changes
+        let buffer_arc = self.buffer.clone();
+        let flags_arc = self.storage.clone();
+
```
timvisee, line 123: We should probably take a copy of changes to write now, not at time of calling the closure.

lib/segment/src/common/buffered_dynamic_flags.rs
```diff
@@ -0,0 +1,573 @@
...
+            // resize if needed
+            if required_len > flags_guard.len() {
+                flags_guard.set_len(required_len)?;
+            }
+
+            for (index, value) in updates {
+                flags_guard.set(index as usize, value);
+            }
+
+            flags_guard.flusher()()?;
```
timvisee, line 138: We should update this flusher to not use the single-use handle. Just directly flushing the mmap instead should be fine.

lib/segment/src/common/buffered_dynamic_flags.rs
```diff
@@ -0,0 +1,573 @@
...
+            let mut buffered_flags = BufferedDynamicFlags::new(mmap_flags);
+
+            // Initial state should match
+            assert_eq!(buffered_flags.count_trues(), 2);
+            assert_eq!(buffered_flags.len(), 3);
+
+            // Set flags beyond current length - this should grow the cached length
+            buffered_flags.set(5, true);
+            buffered_flags.set(7, true);
+            buffered_flags.set(1, true); // Update existing
```
timvisee: Note to self: if we set a higher false, we must bump/persist the length too so that we can correctly iterate. Or do we set the number of points at time of loading?

I suggest to set `(8, false)` and ensure the length becomes 9.

lib/segment/src/common/buffered_dynamic_flags.rs
```diff
@@ -0,0 +1,573 @@
...
+        Box::new(move || {
+            // take all updates from buffer.
+            let updates = std::mem::take(&mut *buffer_arc.write());
+
+            // lock for the entire flushing process
+            let mut flags_guard = flags_arc.lock();
+
+            // Calculate required length based on updates
+            let max_index = updates.keys().map(|&k| k as usize).max().unwrap_or(0);
+            let required_len = max_index + 1;
```
timvisee: We should probably source the length from `cached_len`. It makes sure the number of persisted flags is exact. It is important for our iterators to cover all points.

lib/segment/src/index/field_index/null_index/mutable_null_index.rs
```diff
@@ -283,15 +219,16 @@ impl MmapNullIndex {
...
-            .as_ref()
-            .map_or(0, |storage| storage.has_values_slice.len())
+        self.has_values_flags()
+            .map(|flags| flags.len())
+            .unwrap_or(0)
     }
 
     fn load(&mut self) -> OperationResult<bool> {
-        let is_loaded = self.storage.is_some();
+        // todo: should this load and potentially create new files?
```
timvisee: No, it must not create files.

In that case it must fail, and we handle (re)creation if this index elsewhere.

lib/segment/src/index/field_index/null_index/mutable_null_index.rs
```diff
@@ -175,104 +148,67 @@ impl MmapNullIndex {
...
-            .has_values_slice
-            .set_with_resize(id, has_values, hw_counter_ref)?;
-        storage
-            .is_null_slice
-            .set_with_resize(id, is_null, hw_counter_ref)?;
-
-        // Bump total points
-        self.total_point_count = std::cmp::max(self.total_point_count, id as usize + 1);
+        // Account for I/O cost as if we were writing to disk now
+        hw_counter.payload_index_io_write_counter().incr_delta(2);
```
timvisee, line 154: We **must** also bump the total point count here, like we did before. It affects accuracy of cardinality estimations.

```rust
// Bump total points
self.total_point_count = std::cmp::max(self.total_point_count, id as usize + 1);
```

lib/segment/src/index/field_index/null_index/mutable_null_index.rs
```diff
@@ -175,104 +148,67 @@ impl MmapNullIndex {
...
 
-        // Bump total points
-        // We MUST bump the total point count when removing a point too
-        // On upsert without this respective field, remove point is called rather than add point
-        // Bumping the total point count ensures we correctly iterate over all empty points
-        // Bug: <https://github.com/qdrant/qdrant/pull/6882>
-        self.total_point_count = std::cmp::max(self.total_point_count, id as usize + 1);
+        // Account for I/O cost as if we were writing to disk now
+        let hw_counter = HardwareCounterCell::disposable();
+        hw_counter.payload_index_io_write_counter().incr_delta(2);
```
timvisee, line 174: We **must** also bump the total point count here, like we did before. It affects accuracy of cardinality estimations.

```rust
// Bump total points
// We MUST bump the total point count when removing a point too
// On upsert without this respective field, remove point is called rather than add point
// Bumping the total point count ensures we correctly iterate over all empty points
// Bug: <https://github.com/qdrant/qdrant/pull/6882>
self.total_point_count = std::cmp::max(self.total_point_count, id as usize + 1);
```

lib/segment/src/index/field_index/null_index/mutable_null_index.rs
```diff
@@ -530,6 +404,8 @@ impl FieldIndexBuilderTrait for MmapNullIndexBuilder {
     }
 
     fn finalize(self) -> OperationResult<Self::FieldIndexType> {
+        // Flush any remaining buffered updates
+        self.0.flusher()()?;
```
timvisee, line 423: Note to self: do we do this in other builders too? We should probably remain consistent.

lib/segment/src/common/buffered_dynamic_flags.rs
coszio: but the flusher should be able to flush at any time, isn't it? So that we collect all flushers once and then be able to reuse them?

lib/segment/src/common/buffered_dynamic_flags.rs
coszio: good test case

lib/segment/src/common/buffered_dynamic_flags.rs
coszio: Ok, just confirmed here that we get one flusher every time we flush

https://github.com/qdrant/qdrant/blob/0f7ab94035d20e1cd508fbe4a791f06eac5eeda0/lib/segment/src/segment/entry.rs#L646

But I still have one question: if we take read/write locks when flushing, why would it make a difference to "snapshot" pending updates or not?

lib/segment/src/common/buffered_dynamic_flags.rs
coszio: Initially I considered the updates as a "snapshot" of what will be persisted. But copying the len value will be essentially the same 👍

lib/segment/src/index/field_index/null_index/mutable_null_index.rs
coszio: Good catch!

lib/segment/src/index/field_index/null_index/mutable_null_index.rs
coszio: Thanks for clarifying 👍

lib/segment/src/index/field_index/null_index/mutable_null_index.rs
coszio: Thank you

lib/segment/src/common/buffered_dynamic_flags.rs
coszio: handled by #6966

lib/segment/src/common/buffered_dynamic_flags.rs
coszio: anyway, I updated the PR to take updates once per flusher, not per flush

timvisee: Correct. Since it's a `FnOnce` we're sure we create a new flusher 'task' every time.

> if we take read/write locks when flushing, why would it make a difference to "snapshot" pending updates or not?

Ideally we flush everything at **exactly** the same time. But because IO takes time, that is impossible. This way, by taking a 'snapshot' at time of creating the flusher, all changes are as close as possible in terms of time.

We remember the last operation version right before creating all flushers. After flushing, we confirm to the WAL with that version. That's why taking a 'snapshot' like this is fine. Any new operations that come in at time or after the 'snapshot' will be replayed next time.

Two alternatives (which we don't want to use) are:
1. Don't take a snapshot and grab changes at time of flushing:
   Updates will be further apart in terms of time, because now IO is involved, we don't like that. Because we run every flush operation in sequence, the flush time for each component will be different.
2. Hold a lock staring at flusher creation, which we release after flush:
   This would block incoming reads/writes during the flush stage

In this case I mostly suggested this to remain consistent with all our other flushers. In other flushers we also use this 'snapshot' approach rather than working with live data.

lib/segment/src/index/field_index/null_index/mutable_null_index.rs
```diff
@@ -242,7 +242,9 @@ impl PayloadFieldIndex for MutableNullIndex {
     }
 
     fn cleanup(self) -> OperationResult<()> {
-        std::fs::remove_dir_all(&self.base_dir)?;
+        if let Some(base_dir) = &self.base_dir {
+            std::fs::remove_dir_all(base_dir)?;
+        }
```
timvisee: I think I'd like

```rust
if self.base_dir.is_dir() {
    std::fs::remove_dir_all(self.base_dir)?;
}
```

a bit better without the Option. Just to keep the struct itself a bit simpler.

Wdyt?

lib/segment/src/common/buffered_dynamic_flags.rs
timvisee: You're right that deriving it from the list of updates would achieve the same. Though, this would ensure we always have the highest length value, and not just the highest length we can derive from the current set of updates.

lib/segment/src/index/field_index/null_index/mutable_null_index.rs
timvisee: It looks like some do flush, and some don't. It makes sense to me to flush all of them, we can tackle that separately.

lib/segment/src/index/field_index/null_index/mutable_null_index.rs
timvisee: I will change it to my above suggestion.

It will ensure we also properly clean up if we only have some of the necessary files for the index, which would cause a load failure.

timvisee: Done in 4593c7163d30a182adc884dbe98915d83dc31c56

coszio: 👍 looks good

PR #6965, opened 2025-07-31
TestRawScorerProducer to use VectorStorageEnum

xzfc:
This PR makes `TestRawScorerProducer` to use `VectorStorageEnum::DenseVolatile` instead of rolling its own `impl DenseVectorStorage for TestRawScorerProducer`.

It reduces some boilerplate, and, more important, this change would simplify unit testing the upcoming hnsw-graph-with-vectors feature. (particularly, having an instance of `VectorStorageEnum` would simplify extending some existing tests to use quantization).

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers_builder.rs
@@ -738,7 +738,7 @@ mod tests {
 
         let top = 5;
         let query = random_vector(&mut rng, dim);
-        let scorer = vector_holder.get_scorer(query.clone());
+        let scorer = vector_holder.scorer(query.clone());
         let mut reference_top = FixedLengthPriorityQueue::new(top);
         for idx in 0..vector_holder.storage().total_vector_count() as PointOffsetType {
             let score = scorer.score_point(idx);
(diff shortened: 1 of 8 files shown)


lib/segment/src/index/hnsw_index/graph_layers_builder.rs
```diff
@@ -741,19 +738,16 @@ mod tests {
...
         let mut reference_top = FixedLengthPriorityQueue::new(top);
-        for idx in 0..vector_holder.vectors.len() as PointOffsetType {
-            let vec = &vector_holder.vectors.get(idx as VectorOffsetType);
-            reference_top.push(ScoredPointOffset {
-                idx,
-                score: M::similarity(vec, &processed_query),
-            });
+        for idx in 0..vector_holder.storage().total_vector_count() as PointOffsetType {
+            let score = scorer.score_point(idx);
+            reference_top.push(ScoredPointOffset { idx, score });
```
timvisee, line 745: Didn't we lose preprocessing here (and in other places)?

xzfc: `vector_holder.scorer` (aka `TestRawScorerProducer::scorer`) goes into `FilteredScorerImpl` then into `new_raw_scorer` then all the way into `MetricQueryScorer::new()` which preprocesses the query.

So, no.

timvisee: Thanks!

PR #6966, opened 2025-07-31
Fix flushing of dynamic mmap flags

timvisee:
Fix flushing behavior in the dynamic mmap flags structure, affecting the null and boolean indices.

This makes two changes:
1. on flush, directly flush flags (don't use single-use flusher that noop's on the next flush)
2. before resizing mmap, explicitly flush dirty pages (looks like we lose dirty pages if we don't)

### All Submissions:

### New Feature Submissions:

3. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
4. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### DIFF
--- a/lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
+++ b/lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
@@ -167,6 +167,9 @@ impl DynamicMmapFlags {
         let current_capacity = mmap_max_current_size(self.status.len);
 
         if new_len > current_capacity {
+            // Flush the current mmaps before resizing
+            self.flags.flusher()()?;
+
             // Don't read the whole file on resize
             let populate = false;
             let flags = Self::open_mmap(new_len, &self.directory, populate)?;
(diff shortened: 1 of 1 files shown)


lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
```diff
@@ -173,16 +168,15 @@ impl DynamicMmapFlags {
         let current_capacity = mmap_max_current_size(self.status.len);
 
         if new_len > current_capacity {
+            // Explicitly flush so we don't lose dirty pages
+            self.flusher()()?;
```
timvisee: - because all set functions require `&mut` we are guaranteed to have exclusive access here, we're therefore safe to flush
- we now flush both flags and the status file, we probably only need to flush flags

Note that we should assume that we can flush this structure at any time. New writes to the bool/null index should not be written to this structure directly and should be deferred until we flush, that will be implemented in a separate PR.

timvisee:
Let's finalize and merge <https://github.com/qdrant/qdrant/pull/6954/files> first, and see what parts of this remain relevant.

timvisee:
Note: we should still merge this to satisfy <https://github.com/qdrant/qdrant/pull/6954#discussion_r2251735916>

PR #6967, opened 2025-07-31
HardwareUsage -> Usage

generall:
Make OpenAPI consistent with actual response format


joein (approved):
works on the client side 👍

PR #6972, opened 2025-08-01
Change FnOnce to Fn in MmapFlusher

coszio:
`DynamicMmapFlags` only actually flushed the flags once per flusher, as it was taking the flusher out of the `Option`:

https://github.com/qdrant/qdrant/blob/bd34d52578399a368f203ffe029780a37d817136/lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs#L264-L266

By making `MmapFlusher` a `Fn` rather than `FnOnce`, we can reuse the flusher and avoid taking it from option.

I believe this is safe, since this is the implementation of a mmap flusher:

https://github.com/qdrant/qdrant/blob/bd34d52578399a368f203ffe029780a37d817136/lib/common/memory/src/mmap_type.rs#L160-L174

### DIFF
--- a/lib/common/memory/src/mmap_type.rs
+++ b/lib/common/memory/src/mmap_type.rs
@@ -36,7 +36,7 @@ use crate::mmap_ops;
 /// Result for mmap errors.
 type Result<T> = std::result::Result<T, Error>;
 
-pub type MmapFlusher = Box<dyn FnOnce() -> Result<()> + Send>;
+pub type MmapFlusher = Box<dyn Fn() -> Result<()> + Send>;
 
 /// Type `T` on a memory mapped file
 ///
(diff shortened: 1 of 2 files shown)


xzfc (approved):

timvisee (commented):
While I don't object to this change, I don't see why it is necessary.

We can still implement the dynamic mmap flags flusher in a way where we only flush the last instance. I gave this a shot locally, but the data consistency problems remained.

I'd argue a better 'fix' for this is to properly defer writes to the bitslice until we flush. Then we can use our conventional way of flushing, without having this separate flush handle at all.

coszio:
`Flusher` type is also `FnOnce`, so this change makes the flushing type inconsistent. In my mental model the flusher should be able to be reusable, but in reality the flushers are generated every time we want to flush.

Let's do your approach instead @timvisee, the one in #6966

PR #6975, opened 2025-08-04
Cleanup versions on segment repair

coszio:
We currently cleanup version after replaying WAL, but we should be cleaning them up BEFORE we reapply operations.

There is currently a problem where WAL replay would assign `version=Some(0)` to points which had `version=None`. If we cleanup versions before the replay, we don't need to worry about this

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -687,20 +687,6 @@ impl LocalShard {
         {
             let segments = self.segments.read();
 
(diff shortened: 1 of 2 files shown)


timvisee:
I think we originally had this behavior to still use partially persisted points when replaying the WAL. It was previously possible to CoW a point to an appendable segment, while only flushing the the deletion from the original segment. We've revisited the segment flush order since, always flushing appendable segments first now. I agree that this change seems sound now, and cannot think of a problematic situation.

PR #6976, opened 2025-08-04
Allow retaining more closed segments

KShivendu:
Introduces a new wal config called `wal_retain_closed`. It basically makes wal delta transfers (faster) more likely in large clusters over streaming records (slower). The default value is 1 to maintain backwards compatibility. 

It can be configured using these methods:

1. Env var `export QDRANT__STORAGE__WAL__WAL_RETAIN_CLOSED=5`

2. Qdrant YAML Config

```yaml
storage:
  wal:
    wal_retain_closed: 5
```

Or while **creating the collection** you can pass:
```http
PUT /collections/{name}
{
  // ...
  "wal_config": {
    "wal_retain_closed": 5
  }
}
```

Note that it **can't be modified** after creating collections (same as other wal configs)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -266,7 +266,7 @@ tonic-reflection = "0.11.0"
 tracing = { version = "0.1", features = ["async-await"] }
 uuid = { version = "1.17", features = ["v4", "serde"] }
 validator = { version = "0.18.1", features = ["derive"] }
-wal = { git = "https://github.com/qdrant/wal.git", branch = "max-closed-to-preserve" }
+wal = { git = "https://github.com/qdrant/wal.git", rev = "fad05934581585c18ff84712d6d4f5783b9b6a58" }
 zerocopy = { version = "0.8.26", features = ["derive"] }
 atomic_refcell = "0.1.13"
 byteorder = "1.5.0"
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -53,9 +53,6 @@ storage:
     # Number of WAL segments to create ahead of actual data requirement
     wal_segments_ahead: 0
 
-    # Number of closed WAL segments to keep.
-    wal_retain_closed: 1
-
   # Normal node - receives all updates and answers all queries
   node_type: "Normal"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7178,7 +7178,7 @@ dependencies = [
 [[package]]
 name = "wal"
 version = "0.1.2"
-source = "git+https://github.com/qdrant/wal.git?branch=max-closed-to-preserve#c41800904deb040e5a1037ca7d8fc76022ad121d"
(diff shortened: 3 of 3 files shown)


config/config.yaml
```diff
@@ -53,6 +53,9 @@ storage:
     # Number of WAL segments to create ahead of actual data requirement
     wal_segments_ahead: 0
 
+    # Number of closed WAL segments to keep.
+    wal_retain_closed: 1
```
timvisee: I vote to not add it to our `config.yaml` by default and to keep it a bit more hidden because it is quite specific. If people actively appear to use it, we can always expose it here later.

Wdyt?

Cargo.toml
```diff
@@ -266,7 +266,7 @@ tonic-reflection = "0.11.0"
 tracing = { version = "0.1", features = ["async-await"] }
 uuid = { version = "1.17", features = ["v4", "serde"] }
 validator = { version = "0.18.1", features = ["derive"] }
-wal = { git = "https://github.com/qdrant/wal.git", rev = "c4b26b9c0ccc0e06ba7391189e4c8eac051ca531" }
+wal = { git = "https://github.com/qdrant/wal.git", branch = "max-closed-to-preserve" }
```
timvisee: Let's switch this back to a commit hash once merged in `qdrant/wal`.

timvisee (changes requested):
Works as I expect.

It currently is not possible for users to change this number in existing collections. It is persisted on collection creation, and our collection update API does not expose it. Maybe that's something we can add in a separate PR. I don't think it's an immediate problem and am fine by merging this as is.

Marked 'request changes' until we bump the WAL dependency.

config/config.yaml
KShivendu: Fine. Closed vs open segments is a bit deeper in terms of abstraction. So it's fine to not expose it right away. We can always do it later.

agourlay:
I think one caveat of this approach is that it makes snapshots larger by default, especially in setup with a high segments count.

Currently we have 2 WAL segments (1 open & 1 closed) = 32mb + 32mb = 64mb per data segment.

With  ` "wal_retain_closed": 5`  we would have (1open & 5 closed) = 32mb + 5*32mb = 192mb per data segment.

Not a blocker for sure but something to keep in mind during operations.

WDYT?

timvisee:
@agourlay the [default](https://github.com/qdrant/qdrant/blob/1217f22aac0fb1cde35c90f1cc87f8e74ad14c5b/lib/collection/src/config.rs#L41-L49) is still just 1 closed segment, which is the same as before. Snapshots should therefore be exactly the same size.

The mention of `export QDRANT__STORAGE__WAL__WAL_RETAIN_CLOSED=5` is purely an example of how it may be configured by a user.

If a user chooses to set a higher number of segments to maintain, then we keep more data. And yes, by definition, that means larger snapshots. But there's no way around that. In that case having more history is actually desired.

PR #6977, opened 2025-08-05
Buffered mutable bool index

coszio:
Builds on top of #6954 

Same approach implemented in #6954, but for bool index. Since bool index also used unbuffered updates, this PR switches their flags component to leverage `BufferedDynamicFlags`.

### DIFF
--- a/lib/segment/src/index/field_index/bool_index/mod.rs
+++ b/lib/segment/src/index/field_index/bool_index/mod.rs
@@ -1,6 +1,6 @@
 use common::counter::hardware_counter::HardwareCounterCell;
 use common::types::PointOffsetType;
-use mmap_bool_index::MmapBoolIndex;
+use mutable_bool_index::MutableBoolIndex;
 #[cfg(feature = "rocksdb")]
 use simple_bool_index::SimpleBoolIndex;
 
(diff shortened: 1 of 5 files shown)


timvisee (approved):
✔️

lib/segment/src/index/field_index/bool_index/mod.rs
```diff
@@ -78,12 +78,12 @@ impl BoolIndex {
...
         point_id: PointOffsetType,
         is_true: bool,
-        hw_counter: &HardwareCounterCell,
+        _hw_counter: &HardwareCounterCell,
     ) -> bool {
         match self {
             #[cfg(feature = "rocksdb")]
             BoolIndex::Simple(index) => index.check_values_any(point_id, is_true),
-            BoolIndex::Mmap(index) => index.check_values_any(point_id, is_true, hw_counter),
+            BoolIndex::Mmap(index) => index.check_values_any(point_id, is_true),
```
agourlay, line 86: This new index impl. does not support hardware counter at all? Is it mmap though?

timvisee: Reads/writes are purely done in memory through the roaring bitmap. The mmap is only used for loading into memory, and for writing new changes.

We can count payload index writes when a new value is added, but I'm not sure if it's worth it since these values are quite small.

Since you're more into these counters, whats your opinion on this?

agourlay: Thanks for the explanation.

I do not think it is worth tracking as well for the reasons you mentioned :+1:

agourlay:
Did we benchmark this change with bfb for a sanity check?

coszio:
@agourlay @timvisee 

Just benchmarked it on `boolean_filtering` bench, and I'd say result is same within noise. This is vs `dev`

```
boolean-query-points/binary-index
                        time:   [1.4627 ms 1.4669 ms 1.4729 ms]
                        change: [−4.7812% −3.7910% −2.9575%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
  ```

agourlay (approved):

PR #6978, opened 2025-08-05
Cancel test jobs fast 

tellet-q:
This adds a new step in GitHub job that cancels any running jobs in the same workflow according to the specified conditions. This behavior ensures the workflow cancels as soon as  any image build failed.

```
      - name: Docker build
        id: docker-build
        uses: docker/build-push-action@v6
        with:
          context: .
          ...
      - name: Cancel other jobs in a run if Docker build failed
        if: failure() && steps.docker-build.outcome == 'failure'
        uses: ./.github/actions/cancel-workflow-run
        with:
          github-token: ${{ secrets.GITHUB_TOKEN }}
```

### All Submissions:

### DIFF
--- a/tools/sync-web-ui.sh
+++ b/tools/sync-web-ui.sh
@@ -8,7 +8,32 @@ OPENAPI_FILE=${OPENAPI_DIR:-"./docs/redoc/master/openapi.json"}
 # Download `dist.zip` from the latest release of https://github.com/qdrant/qdrant-web-ui and unzip given folder
 
 # Get latest dist.zip, assume jq is installed
-DOWNLOAD_LINK=$(curl --retry 5 --retry-all-errors --silent "https://api.github.com/repos/qdrant/qdrant-web-ui/releases/latest" | jq -r '.assets[] | select(.name=="dist-qdrant.zip") | .browser_download_url')
+API_RESPONSE=$(curl --retry 5 --retry-all-errors --silent -w "\n%{http_code}" --connect-timeout 10 --max-time 30 "https://api.github.com/repos/qdrant/qdrant-web-ui/releases/latest")
+HTTP_CODE=$(echo "$API_RESPONSE" | tail -n1)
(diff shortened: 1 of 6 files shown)


timvisee:
Random thought: would it be possible to build the (shared?) image in a single job first, and then run all jobs that need it in a second stage. If the build fails, it will not run the second stage, potentially simplifying this workflow config. Would something like that work?

tools/sync-web-ui.sh
```diff
@@ -8,7 +8,32 @@ OPENAPI_FILE=${OPENAPI_DIR:-"./docs/redoc/master/openapi.json"}
...
+fi
+
+DOWNLOAD_LINK=$(echo "$API_RESPONSE" | jq -r '.assets[] | select(.name=="dist-qdrant.zip") | .browser_download_url' 2>/dev/null)
+
+if [ -z "$DOWNLOAD_LINK" ] || [ "$DOWNLOAD_LINK" = "null" ]; then
+    echo "Error: Could not find dist-qdrant.zip in the latest release"
+    echo "Available assets:"
+    echo "$API_RESPONSE" | jq -r '.assets[]' 2>/dev/null || echo "Failed to parse assets"
+    exit 1
+fi
```
xzfc: We can avoid using JSON GitHub API at all and download from there:
`https://github.com/qdrant/qdrant-web-ui/releases/latest/download/dist-qdrant.zip`

tellet-q: If the link never changes, then yes, it would be simpler

tellet-q:
> Random thought: would it be possible to build the (shared?) image in a single job first, and then run all jobs that need it in a second stage. If the build fails, it will not run the second stage, potentially simplifying this workflow config. Would something like that work?

@timvisee I see your point, but at least one of the impacted tests requires specific context to build the image (context: tools/schema2openapi), so single job would not be enough. We'd have to build all the required images in parallel during some kind of preparation stage and make other tests (even though they do not require an image) dependent on it.

PR #6980, opened 2025-08-05
Partial snapshots update recovery ts

KShivendu:
Minor improvement on top of https://github.com/qdrant/qdrant/pull/6924

We want to set recovery timestamp even on empty partial snapshots since it's used to decide the replicas to sync.

### DIFF
--- a/src/actix/api/snapshot_api.rs
+++ b/src/actix/api/snapshot_api.rs
@@ -12,7 +12,8 @@ use collection::operations::snapshot_ops::{
 };
 use collection::operations::verification::new_unchecked_verification_pass;
 use collection::shards::replica_set::snapshots::RecoveryType;
-use collection::shards::{shard::ShardId, shard_holder::shard_not_found_error};
+use collection::shards::shard::ShardId;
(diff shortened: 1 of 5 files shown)


src/actix/api/snapshot_api.rs
```diff
@@ -750,6 +751,13 @@ async fn recover_partial_snapshot_from(
                 .error_for_status()?;
 
             if response.status() == reqwest::StatusCode::NOT_MODIFIED {
+                let shard_holder = collection.shards_holder();
+                let shard_holder = shard_holder.read().await;
+                let replica_set = shard_holder
+                    .get_shard(shard_id)
+                    .ok_or_else(|| shard_not_found_error(shard_id))?;
+
+                replica_set.partial_snapshot_meta.snapshot_recovered();
```
timvisee, line 762: Let's add a comment here to explain why we do this


```suggestion
                // Our replica is up to date so we bump the recovered timestamp
                // This prevents cluster manager from immediately trying to recover again
                replica_set.partial_snapshot_meta.snapshot_recovered();
```

timvisee (approved):
Thanks for the test!

PR #6981, opened 2025-08-05
Fix Clippy 1.89

agourlay:
Clippy fixes for Rust [1.89](https://releases.rs/docs/1.89.0/)

Most of the diff is caused by

> [Add a warn-by-default mismatched_lifetime_syntaxes lint.](https://github.com/rust-lang/rust/pull/138677) This lint detects when the same lifetime is referred to by different syntax categories between function arguments and return values, which can be confusing to read, especially in unsafe code.

### DIFF
--- a/lib/api/src/rest/validate.rs
+++ b/lib/api/src/rest/validate.rs
@@ -219,14 +219,14 @@ impl Validate for Batch {
             BatchVectorStruct::Image(_) => {}
             BatchVectorStruct::Object(_) => {}
         }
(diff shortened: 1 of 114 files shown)


ffuugoo:
> Most of the diff is caused by
>> https://github.com/rust-lang/rust/pull/138677

Finally. I *hate* that Rust allows omitted default lifetime params in types. I absolutely *do* want to know that the type is lifetime-bound. 🙄

PR #6982, opened 2025-08-06
Implement GraphLinksFormat::CompressedWithVectors

xzfc:
This PR:

1. Adds `GraphLinksFormat::CompressedWithVectors`/`links_comp_vec.bin` format.
2. Implements both writing (`GraphLinksSerializer`) and reading (`GraphLinksView`).
3. Adds `GraphLayersWithVectors::for_each_link_with_vector()` trait method.
   1. Similar to `GraphLayersBase::for_each_link()` (formerly, `GraphLayersBase::links_map()`), but iterates over `(PointOffsetType, &[u8])` pairs.
   2. It is unused for now, but it will be used later in a `GraphLayersBase::_search_on_level()`-like function.
4. Updates existing tests to cover the new format. Particularly, imbues `TestRawScorerProducer` with quantized vectors.
5. Simplifies some tests.

The main commit is `Implement GraphLinksFormat::CompressedWithVectors`, everything else is a preparation work and could be reviewed separately.

# Format

For comparison, the old format (`links_compressed.bin`) has a section called `compressed_links` which looks like this:
```
[cccccccccccccccccc][cccccccccccccccccc][cccccccccccccccccc]...
[ links for node 0 ][ links for node 1 ][ links for node 2 ]...
```

The new format (`links_comp_vec.bin`) is based on that, but this section now looks like this:
```
[LLcccccccccccccQQQQQQQQQQQQ][LLcccccccccccccQQQQQQQQQQQQ][LLcccccccccccccQQQQQQQQQQQQ]...
[ links and vecs for node 0 ][ links and vecs for node 1 ][ links and vecs for node 2 ]...
```

- `c` are compressed links (i.e. it represents `Vec<u32>` but compressed).
- `Q` are encoded quantized vectors (i.e. `Vec<Vec<u8>>`).
- `L` is a varint-encoded length. Number of links == number of quantized vectors == this value.

It assumes that every quantized vector has the same byte length. Consequently, multivectors are not supported.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_links/serializer.rs
+++ b/lib/segment/src/index/hnsw_index/graph_links/serializer.rs
@@ -110,8 +110,6 @@ impl GraphLinksSerializer {
                 _ => (hnsw_m.m, Either::Right(back_index[..count].iter().copied())),
             };
 
-            let mut compressed_links_buf = Vec::new();
-
             iter.try_for_each(|id| {
                 let mut raw_links = take(&mut edges[id][level]);
                 match format_param {
@@ -137,23 +135,20 @@ impl GraphLinksSerializer {
                         // Unwrap safety: `impl Write for Vec<u8>` never fails.
                         neighbors.write_varint(raw_links.len() as u64).unwrap();
 
-                        // 2. Padding to align vectors (`_` in the doc).
-                        let padding = neighbors.len().next_multiple_of(vector_layout.align())
-                            - neighbors.len();
-                        neighbors.write_zeros(padding).unwrap();
-
-                        // Prepare compressed links.
-                        // NOTE: we can avoid using intermediate buffer if we
-                        // split `pack_links` into two steps: prepare and write.
-                        // But it's not a bottleneck for now.
+                        // 2. Compressed links (`c` in the doc)
                         pack_links(
-                            &mut compressed_links_buf,
+                            &mut neighbors,
                             &mut raw_links,
                             bits_per_unsorted,
                             sorted_count,
                         );
 
-                        // 3. Vectors (`V` in the doc).
+                        // 3. Padding to align vectors (`_` in the doc).
+                        let padding = neighbors.len().next_multiple_of(vector_layout.align())
+                            - neighbors.len();
+                        neighbors.write_zeros(padding).unwrap();
+
+                        // 4. Vectors (`V` in the doc).
                         // Write them in the same order as `raw_links`.
                         for i in raw_links {
                             let vector = vectors.get_vector(i)?;
@@ -162,10 +157,6 @@ impl GraphLinksSerializer {
                             }
                             neighbors.extend(vector);
                         }
-
-                        // 4. Compressed links (`c` in the doc)
-                        neighbors.extend_from_slice(&compressed_links_buf);
-                        compressed_links_buf.clear();
 
                         offsets.push(neighbors.len() as u64);
                     }
--- a/lib/segment/src/index/hnsw_index/graph_links/view.rs
+++ b/lib/segment/src/index/hnsw_index/graph_links/view.rs
@@ -2,7 +2,9 @@ use std::iter::{Copied, Zip};
 use std::num::NonZero;
 
 use common::bitpacking::packed_bits;
-use common::bitpacking_links::{MIN_BITS_PER_VALUE, PackedLinksIterator, iterate_packed_links};
+use common::bitpacking_links::{
+    MIN_BITS_PER_VALUE, PackedLinksIterator, iterate_packed_links, packed_links_size,
+};
 use common::bitpacking_ordered;
 use common::types::PointOffsetType;
 use integer_encoding::VarInt as _;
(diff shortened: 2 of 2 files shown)


xzfc:
@coderabbitai full review

lib/segment/src/index/hnsw_index/graph_links/view.rs
```diff
@@ -127,6 +177,62 @@ impl GraphLinksView<'_> {
...
+                bits_per_unsorted,
+                mut vector_size_bytes,
+            } => {
+                let start = offsets.get(idx).unwrap() as usize;
+                let end = offsets.get(idx + 1).unwrap() as usize;
+
+                let (vectors_count, vectors_count_size) =
+                    u64::decode_var(&compressed_links[start..end]).unwrap();
+                let links_start = start + vectors_count_size;
+                let vectors_start = end - vectors_count as usize * vector_size_bytes;
```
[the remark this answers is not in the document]
xzfc: Decline. This implementation is consistent with `GraphLinksView::links()` witch also could panic (and doesn't use Result for performance reasons). I might cleanup this method but only after doing benchmarks, but I need to flesh out other parts of the feature first.

lib/segment/src/fixtures/index_fixtures.rs
```diff
@@ -33,25 +40,59 @@ impl TestRawScorerProducer {
                 .unwrap();
         }
 
+        let quantized_vectors = use_quantization.then(|| {
+            QuantizedVectors::create(
+                &storage,
+                &ScalarQuantizationConfig {
```
generall, line 46: nit: new hnsw layout is mostly about binary quantization

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -34,15 +39,38 @@ enum Kind {
...
+        edges: Vec<Vec<Vec<PointOffsetType>>>,
+        format_param: GraphLinksFormatParam<&'a QuantizedVectors>,
+        hnsw_m: HnswM,
+    ) -> OperationResult<Self> {
+        let get_quantized_vector;
+        let format = match format_param {
+            GraphLinksFormatParam::Plain => GraphLinksFormatParam::Plain,
+            GraphLinksFormatParam::Compressed => GraphLinksFormatParam::Compressed,
+            GraphLinksFormatParam::CompressedWithVectors(qv) => {
+                get_quantized_vector = |i| qv.get_quantized_vector(i);
```
generall: I would try to make it universal, and not just quantized

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -34,15 +39,38 @@ enum Kind {
...
+
+        Self::new_inner(edges, format, hnsw_m)
+    }
+
+    /// Unwrapped version of [`Self::new`] that accepts `&dyn Fn` rather than
+    /// `&QuantizedVectors`. This unwrapped version is intended for testing.
+    pub(super) fn new_inner<'a>(
         mut edges: Vec<Vec<Vec<PointOffsetType>>>,
-        format: GraphLinksFormat,
+        format_param: GraphLinksFormatParam<&dyn Fn(PointOffsetType) -> &'a [u8]>,
```
generall: since we actually want to only have function inside enum and nothing else, let's try to avoid `&dyn Fn` as a generic param

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -78,37 +106,65 @@ impl GraphLinksSerializer {
         let bits_per_unsorted = packed_bits(u32::try_from(edges.len().saturating_sub(1)).unwrap())
             .max(MIN_BITS_PER_VALUE);
 
+        let mut quantized_vector_size: Option<usize> = None;
```
generall: -> vector_size_bytes

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -78,37 +106,65 @@ impl GraphLinksSerializer {
...
                         links.extend_from_slice(raw_links.as_bytes());
                         offsets.push((links.len() as u64) / size_of::<PointOffsetType>() as u64);
                     }
+                    GraphLinksFormatParam::CompressedWithVectors(get_quantized_vector) => {
+                        // unwrap safety: `impl Write for Vec<u8>` never fails.
+                        links.write_varint(raw_links.len() as u64).unwrap();
+
+                        pack_links(&mut links, &mut raw_links, bits_per_unsorted, sorted_count);
+                        for i in raw_links {
+                            let vector = get_quantized_vector(i);
```
generall: -> get_vector_bytes

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -34,15 +39,38 @@ enum Kind {
     Compressed {
         compressed_offsets: Vec<u8>,
         offsets_parameters: bitpacking_ordered::Parameters,
+        quantized_vector_size: Option<usize>,
```
generall: -> vector_size

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -78,37 +106,65 @@ impl GraphLinksSerializer {
...
+                    GraphLinksFormatParam::Plain => {
                         links.extend_from_slice(raw_links.as_bytes());
                         offsets.push((links.len() as u64) / size_of::<PointOffsetType>() as u64);
                     }
+                    GraphLinksFormatParam::CompressedWithVectors(get_quantized_vector) => {
+                        // unwrap safety: `impl Write for Vec<u8>` never fails.
+                        links.write_varint(raw_links.len() as u64).unwrap();
+
+                        pack_links(&mut links, &mut raw_links, bits_per_unsorted, sorted_count);
+                        for i in raw_links {
```
generall, line 153: We likely need an alignment here, in case we will store non-byte version of vectors here

lib/segment/src/index/hnsw_index/graph_links/view.rs
```diff
@@ -41,6 +51,13 @@ pub(super) enum CompressionInfo<'a> {
         hnsw_m: HnswM,
         bits_per_unsorted: u8,
     },
+    CompressedWithVectors {
+        compressed_links: &'a [u8],
```
generall: -> compressed_links_and_vectors

lib/segment/src/fixtures/index_fixtures.rs
xzfc: Some of the tests are too fragile to changes in accuracy. So, I prefer to use less aggressive quantization for them rather than tuning test parameters making them too slow¹ or too complicated².

¹ if using the same parameters for both quantized and non-quantized variants of the same test
² if choosing parameters based on whether the quantization is used in a test variant

PR #6984, opened 2025-08-06
bump rust 1.88

voyager0003:
fix build error if use rust 1.87

if use rust 1.87 as Cargo.toml in latest master branch, will get an error when build from source
cargo build --release --bin qdrant
QQ_1754457564273

after update rust to 1.88, the error is fixed.

btw, the rust version in the dockerfile is also 1.88.


timvisee (approved):
Thanks!

I'll target your branch to `dev` instead to follow our regular development workflow. It'll eventually land in the main branch as well.

PR #6988, opened 2025-08-06
Refactor `Collection::restore_shard_snapshot`

ffuugoo:
This PR makes two improvements to `Collection::restore_shard_snapshot`:

First, it makes it cancel-safe, so that it's a bit safer to use. I don't think we have any issues with this method as-is, but it's just nice to have.

Second, restructures `restore_shard_snapshot` as a "two-step" future. It now returns *additional* future. E.g.:

```rust
// `restore_shard_snapshot` is `async fn`, that returns a future, that perfoms *validation* of snapshot archive
let validate = collection.restore_shard_snapshot(...);

// when you `await` it, it returns *another* future, that performs snapshot restore *in background*
// (we can await first future to validate snapshot with `wait=false`, and return an error *before* returning `accepted`)
let restore = validate.await?;

// and you can await this future *if you want to* to get result of recovery
// (we can await this to return response with `wait=true`)
let result = restore.await?;
```

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/actix/api/snapshot_api.rs
+++ b/src/actix/api/snapshot_api.rs
@@ -1,5 +1,6 @@
 use std::path::Path;
 
+use ::common::tempfile_ext::MaybeTempPath;
 use actix_multipart::form::MultipartForm;
 use actix_multipart::form::tempfile::TempFile;
 use actix_web::{Responder, Result, delete, get, post, put, web};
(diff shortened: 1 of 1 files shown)


src/actix/api/snapshot_api.rs
```diff
@@ -790,7 +790,7 @@ async fn recover_partial_snapshot_from(
             dispatcher.toc(&access, &pass),
             &collection,
             shard_id,
-            &partial_snapshot_temp_path,
+            partial_snapshot_temp_path.into(),
```
generall: please use `from` instead

timvisee (approved):

src/actix/api/snapshot_api.rs
ffuugoo: Done.

PR #6990, opened 2025-08-06
`BitvecFlags`: bitvec container with dynamic mmap flags storage

coszio:
This PR extracts the `BufferedDynamicFlags` logic into its own structure, which buffers writes and can only write to disk manually. 

To handle reads, two new structures have been created:
- `RoaringFlags`: uses the `BufferedDynamicFlags` as storage only, and keeps a `RoaringBitmap` in memory for reads.
- `BitvecFlags`: same `BufferedDynamicFlags` storage, but with a `BitVec` in memory for reads.

This enables the flexibility of choosing the in-memory structure. For example, vector storages need a reference to a bitslice, which can only be provided by `BitvecFlags`. On the other hand, null index benefits from lower memory overhead and does not need a bitslice explicitly.

### DIFF
--- a/lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
+++ b/lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
@@ -261,6 +261,8 @@ impl DynamicMmapFlags {
     }
 
     pub fn get_bitslice(&self) -> &BitSlice {
+        // Take subslice with actual length, bitslice may be larger due to extra allocated capacity
+        // See `mmap_capacity_bytes`
         &self.flags[..self.len()]
     }
 


lib/segment/src/common/bitvec_flags.rs
```diff
@@ -0,0 +1,181 @@
...
+    bitvec: BitVec,
+
+    /// Total length of the flags, including the trailing ones which have been set to false
+    len: usize,
+}
+
+impl BitvecFlags {
+    pub fn new(mmap_flags: DynamicMmapFlags) -> Self {
+        // load flags into memory
+        let bitvec = BitVec::from_bitslice(mmap_flags.get_bitslice());
```
generall, line 33: `mmap_flags` should be cleaned from cache here

lib/segment/src/common/roaring_flags.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// Total length of the flags, including the trailing ones which have been set to false
+    len: usize,
+}
+
+impl RoaringFlags {
+    pub fn new(mmap_flags: DynamicMmapFlags) -> Self {
+        // load flags into memory
+        let bitmap = RoaringBitmap::from_sorted_iter(mmap_flags.iter_trues())
+            .expect("iter_trues iterates in sorted order");
+
```
generall, line 35: same, `mmap_flags` should be cleared from cache

lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
```diff
@@ -261,7 +261,7 @@ impl DynamicMmapFlags {
     }
 
     pub fn get_bitslice(&self) -> &BitSlice {
-        &self.flags
+        &self.flags[..self.len()]
```
timvisee, line 266: Note to self: Are we sure this is always in bounds? And, can't we just return the full bitslice here?

lib/segment/src/common/bitvec_flags.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// Returns the previous value of the flag.
+    pub fn set(&mut self, index: PointOffsetType, value: bool) {
+        // queue write in buffer
+        self.storage.buffer_set(index, value);
+
+        // update length if needed
+        let index_usize = index as usize;
+        if index_usize >= self.len {
+            self.len = index_usize + 1;
+            self.bitvec.resize_with(self.len, |_| false);
```
timvisee: Note to self: is there a cheaper resize method?

timvisee (approved):
Looks good. Thanks!

I carefully checked the way we handle the length of structures here, as it's important for reliable iterators. While there's a bit of indirection there, everything looks sound.

lib/segment/src/common/bitvec_flags.rs
timvisee: [`resize`](https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html#method.resize) uses a constant and eliminates the (hot) closure. Not sure if there actually is a performance difference after compilation, but at least the method is simpler.

https://github.com/qdrant/qdrant/pull/6990/commits/194352365d8de9208ab9b9e61a0021178c429ddd

lib/segment/src/vector_storage/dense/dynamic_mmap_flags.rs
timvisee: Good catch! We do this because the bitslice may be larger due to extra allocated capacity.

Added a comment here: https://github.com/qdrant/qdrant/pull/6990/commits/c97b6aa63ec1ca3e29fa264df3fb149f10348281

Issue #6956, opened 2025-07-30
Log is not written to the log file


rikuto-imai:
## Current Behavior
Although the configuration file is set to create a log file, nothing is written to the file.  
Logs are correctly output to standard output.

## Steps to Reproduce
1. Modify `config/config.yaml` as below:  
   ```yaml
   log_level: INFO
   logger:
     format: text
     on_disk:
       enabled: true
       log_file: /qdrant/log/qdrant.log
       log_level: INFO
       format: text
   ```
2. Start Qdrant:  
   ```bash
   docker run -p 6333:6333 \
     -v $(pwd)/config/config.yaml:/qdrant/config/config.yaml \
     -v $(pwd)/log:/qdrant/log \
     qdrant/qdrant
   ```
3. Confirm that logs are printed to standard output  
4. Confirm that `log/qdrant.log` remains 0B in size  
5. Even after stopping Qdrant with Ctrl+C, `log/qdrant.log` still remains 0B  

## Expected Behavior
The logs that are printed to standard output should also be written to the specified log file.

## Possible Solution
I’ve confirmed that `std::io::BufWriter` is used for file output:  
https://github.com/qdrant/qdrant/blob/af7ab5b192f01c9d9cd7fff119eab2ae73755cc0/src/tracing/on_disk.rs#L80  

By default, `BufWriter` has an 8KiB buffer and does not flush output to the file until the buffer is full:  
https://doc.rust-lang.org/std/io/struct.BufWriter.html#impl-BufWriter%3CW%3E  

You can control this buffer size using `with_capacity`.  
After rebuilding Qdrant with the buffer size set to 100B, the logs were properly written to the file.  

It would be helpful if the buffer size could be made configurable via the configuration file.

## Context (Environment)
In a production environment, if logs are missing from the log file, it makes troubleshooting during failures very difficult.

Qdrant Version: 1.15.1

ffuugoo:
Thanks to @nibblerenush for implementing configuration parameter. I'll close the issue for now.

rikuto-imai:
Thank you for the fix. I adjusted the buffer size and confirmed that the logs can now be output.
I believe this will be included in v1.15.2, but is it safe to assume it will be released within August?

P.S.
When working on a fix, I would appreciate it if you could include a brief comment such as “This is a bug, so fix it.”

timvisee:
> I believe this will be included in v1.15.2, but is it safe to assume it will be released within August?

[...]

PR #6993, opened 2025-08-06
Add buffer_size_b to on_disk

nibblerenush:
Fix this issue: https://github.com/qdrant/qdrant/issues/6956

### DIFF
--- a/src/tracing/on_disk.rs
+++ b/src/tracing/on_disk.rs
@@ -17,6 +17,7 @@ pub struct Config {
     pub log_level: Option<String>,
     pub format: Option<config::LogFormat>,
     pub span_events: Option<HashSet<config::SpanEvent>>,
+    pub buffer_size_b: Option<usize>,
 }
 
 impl Config {
@@ -27,6 +28,7 @@ impl Config {
             log_level,
             span_events,
             format,
+            buffer_size_b,
         } = other;
 
         self.enabled.replace_if_some(enabled);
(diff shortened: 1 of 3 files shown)


src/tracing/on_disk.rs
```diff
@@ -17,6 +17,7 @@ pub struct Config {
     pub log_level: Option<String>,
     pub format: Option<config::LogFormat>,
     pub span_events: Option<HashSet<config::SpanEvent>>,
+    pub buffer_size_b: Option<usize>,
```
ffuugoo:
```suggestion
    pub buffer_size_bytes: Option<usize>,
```

ffuugoo (changes requested):
LGTM, I'd only ask to rename `buffer_size_b`  to `buffer_size_bytes`

src/tracing/on_disk.rs
nibblerenush: Finished and ready for review

timvisee (commented):
Could you check for us what behavior is like when you set the buffer size to 0 or 1?

nibblerenush:
> Could you check for us what behavior is like when you set the buffer size to 0 or 1 bytes?

I use the following configuration:
```
logger:
  # Logging format, supports `text` and `json`
  format: text
  on_disk:
    enabled: true
    log_file: qdrant.log
    log_level: TRACE
    # Logging format, supports `text` and `json`
    format: text
    buffer_size_bytes: <0 or 1>
```
In both cases, the expected behavior is observed, the qdrant does not fail, the information is written to the log file

nibblerenush:
Thanks for the opportunity!

PR #6994, opened 2025-08-06
Use `BitvecFlags` in vector storages

coszio:
To appropriately buffer writes to the deleted bitslice in the vector storages, we need to have a structure like `BitvecFlags` which buffers updates, but also keeps a fast in-memory copy of the bitslice.

In #6990 we introduced the structure. This PR does two things:
1. moves all the flags files into a single module
1. integrates the `BitvecFlags` where `DynamicMmapFlags` was being used directly.


lib/segment/src/vector_storage/dense/appendable_dense_vector_storage.rs
```diff
@@ -39,27 +40,29 @@ pub struct AppendableMmapDenseVectorStorage<T: PrimitiveVectorElement, S: Chunke
 impl<T: PrimitiveVectorElement, S: ChunkedVectorStorage<T>> AppendableMmapDenseVectorStorage<T, S> {
     /// Set deleted flag for given key. Returns previous deleted state.
     #[inline]
-    fn set_deleted(&mut self, key: PointOffsetType, deleted: bool) -> OperationResult<bool> {
+    fn set_deleted(&mut self, key: PointOffsetType, deleted: bool) -> bool {
         if !deleted && self.vectors.len() <= key as usize {
-            return Ok(false);
+            return false;
         }
```
timvisee, line 50: When we make this optimization, it makes the length of the deleted structure unreliable. Thus we must not depend on it anywhere.

timvisee (approved):

PR #6995, opened 2025-08-07
Refactor bash tests: tls

tellet-q:
This is a next PR in a series of PRs to replace some of the old bash tests. The goal is to reduce the number of jobs run in Integration tests workflow and unify running of the tests.

This replaces tests/tls


tests/e2e_tests/test_tls.py
```diff
@@ -0,0 +1,230 @@
...
+        """
+        config = {"compose_file": "tls-compose.yaml", "wait_for_ready": False}
+        container_info, cleanup = run_docker_compose(docker_client, qdrant_image, test_data_dir, config)
+
+        # Should be a list of 2 nodes
+        assert isinstance(container_info, list)
+        assert len(container_info) == 2
+
+        # Wait for cluster to stabilize
+        time.sleep(15)
```
KShivendu, line 35: Consider adding a health check instead of hardcoded time.

tests/e2e_tests/utils.py
```diff
@@ -239,3 +242,195 @@ def extract_archive(archive_file: Path, extract_to: Path, cleanup_archive: bool
...
+        container=container,
+        host="127.0.0.1",
+        name=container.name,
+        http_port=http_port,
+        grpc_port=grpc_port,
+        compose_project=project_name
+    )
+
+
+def run_docker_compose(docker_client, qdrant_image, test_data_dir, config):
```
KShivendu, line 321: Can we return something like `QdrantCluster` here? It can have a dedicated clean up method that can be called. Wdyt?

tests/e2e_tests/conftest.py
```diff
@@ -247,3 +247,57 @@ def test_something(storage_from_archive):
...
+
+@pytest.fixture(scope="function")
+def qdrant_compose(docker_client, qdrant_image, test_data_dir, request):
+    """
+    Fixture for creating Qdrant containers using docker-compose.yaml files from test_data folder.
+
+    Usage with parametrization:
+        @pytest.mark.parametrize("qdrant_compose", [
+            {"compose_file": "docker-compose.yaml"}
+        ], indirect=True)
```
KShivendu, line 260: What does `indirect=True` do?

tests/e2e_tests/test_data/cert.cfg
```diff
@@ -21,3 +21,5 @@ subjectAltName       = @alt_names
 [alt_names]
 DNS.1 = node1.qdrant
 DNS.2 = node2.qdrant
+DNS.3 = localhost
+IP.1 = 127.0.0.1
```
KShivendu, line 25: Okay. But why exactly do we need to add this now?

KShivendu (changes requested):
Cool. Have suggested a few changes and have a few questions :)

tests/e2e_tests/conftest.py
tellet-q: it's a part of pytest parametrization, it allows to parametrize fixtures

tests/e2e_tests/test_tls.py
tellet-q: since it's a fixture, I'd like to keep it this way, so that only the test actually asserts the cluster

tests/e2e_tests/conftest.py
```diff
@@ -247,3 +247,57 @@ def test_something(storage_from_archive):
     extract_archive(archive_file, extract_to)
 
     return temp_storage_dir
+
+
+@pytest.fixture(scope="function")
+def qdrant_compose(docker_client, qdrant_image, test_data_dir, request):
```
KShivendu, line 253: I think this isn't used anywhere.

KShivendu (approved):
Please look at my comment. Rest LGTM. Thanks!

tests/e2e_tests/conftest.py
tellet-q: yeah, I added it for future use, let's keep for now

PR #7002, opened 2025-08-08
Improve h-sum in dot_avx calculation

JojiiOfficial:


lib/segment/src/spaces/simple_avx.rs
```diff
@@ -6,13 +6,25 @@ use super::tools::is_length_zero_or_normalized;
...
 pub unsafe fn hsum256_ps_avx(x: __m256) -> f32 {
-    let x128: __m128 = _mm_add_ps(_mm256_extractf128_ps(x, 1), _mm256_castps256_ps128(x));
-    let x64: __m128 = _mm_add_ps(x128, _mm_movehl_ps(x128, x128));
-    let x32: __m128 = _mm_add_ss(x64, _mm_shuffle_ps(x64, x64, 0x55));
-    _mm_cvtss_f32(x32)
+    let lr_sum: __m128 = _mm_add_ps(_mm256_extractf128_ps(x, 1), _mm256_castps256_ps128(x));
+    let hsum = _mm_hadd_ps(lr_sum, lr_sum);
+    let p1 = _mm_extract_ps(hsum, 0);
+    let p2 = _mm_extract_ps(hsum, 1);
+    f32::from_bits(p1 as u32) + f32::from_bits(p2 as u32)
```
JojiiOfficial, line 15: The new implementation uses 1 instruction less: https://godbolt.org/z/5no3W3s9K

lib/segment/src/spaces/simple_avx.rs
```diff
@@ -6,13 +6,25 @@ use super::tools::is_length_zero_or_normalized;
...
+}
+
+/// Calculates the hsum (horizontal sum) of four 32 byte registers.
+#[target_feature(enable = "avx")]
+#[allow(clippy::missing_safety_doc)]
+pub unsafe fn four_way_hsum(a: __m256, b: __m256, c: __m256, d: __m256) -> f32 {
```
timvisee, line 21: Does adding `#[inline]` here make any difference? (I expect not)

PR #7008, opened 2025-08-11
Refactor bash test for snapshots recovery

tellet-q:
This is a next PR in a series of PRs to replace some of the old bash tests. The goal is to reduce the number of jobs run in Integration tests workflow and unify running of the tests.

This replaces tests/snapshots


tests/e2e_tests/test_snapshots_recovery.py
```diff
@@ -0,0 +1,130 @@
...
+                    'mode': 'ro'
+                }
+            }
+        )
+        container_info = qdrant_container_factory(config)
+        
+        client = ClientUtils(host=container_info.host, port=container_info.http_port)
+        
+        # Create collection and insert points
+        collection_config = {"vectors": {"size": 4, "distance": "Dot"}}
```
KShivendu:
```suggestion
        client = ClientUtils(host=container_info.host, port=container_info.http_port)
        client.wait_for_server()
        
        # Create collection and insert points
        collection_config = {"vectors": {"size": 4, "distance": "Dot"}}
```

We need to wait like the bash script.

tests/e2e_tests/test_snapshots_recovery.py
```diff
@@ -0,0 +1,130 @@
...
+        config = QdrantContainerConfig(
+            network_mode="host",
+            volumes={
+                str(config_file): {
+                    'bind': '/qdrant/config/config.yaml',
+                    'mode': 'ro'
+                }
+            }
+        )
+        container_info = qdrant_container_factory(config)
```
KShivendu, line 70: I'm a bit confused if this does this steps:

```sh
docker volume create snapshots
docker volume create tempdir
docker volume create storage

declare DOCKER_IMAGE_NAME=qdrant-snapshots
#declare DOCKER_IMAGE_NAME=qdrant/qdrant:snapshots
declare CONTAINER_NAME=qdrant-snapshots-container

docker buildx build --build-arg=PROFILE=ci --load ../../ --tag=$DOCKER_IMAGE_NAME

docker run \
    --rm -d  --network host \
    -p 6333:6333 -p 6334:6334 \
    -v snapshots:/qdrant/snapshots \
    -v tempdir:/qdrant/tempdir \
    -v storage:/qdrant/storage \
    -e QDRANT__STORAGE__TEMP_PATH=/qdrant/tempdir \
    --name ${CONTAINER_NAME} \
    $DOCKER_IMAGE_NAME
```

Can you please show me how it does that?

tests/e2e_tests/test_snapshots_recovery.py
```diff
@@ -0,0 +1,130 @@
...
+        
+        client = ClientUtils(host=container_info.host, port=container_info.http_port)
+        
+        # Create collection and insert points
+        collection_config = {"vectors": {"size": 4, "distance": "Dot"}}
+        client.create_collection("test_collection", collection_config)
+        for points_batch in client.generate_points(2):
+            client.insert_points("test_collection", points_batch)
```
KShivendu:
```suggestion
            client.insert_points("test_collection", points_batch, wait=True)
```

I see that it uses `wait=True` internally inside `insert_points`. But I think it should be an explicit param.

tests/e2e_tests/test_snapshots_recovery.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # Create collection snapshot
+        snapshot_name = client.create_snapshot("test_collection")
+        assert snapshot_name, "Snapshot creation failed"
```
KShivendu, line 84:
```py
def create_snapshot(self, collection_name: str = "test_collection") -> str:
        """Create a snapshot of the collection."""
        snapshot_info = self.client.create_snapshot(collection_name=collection_name)
        return snapshot_info.name
```

The code needs to be adjusted to return `None` or it will raise exception that property doesn't exist which still fails the test but is wrong from the API point of view.

Something like

```py
def create_snapshot(self, collection_name: str = "test_collection") -> Optional[str]:
        """Create a snapshot of the collection."""
        snapshot_info = self.client.create_snapshot(collection_name=collection_name)
        return snapshot_info.name if snapshot_info else None
```

tests/e2e_tests/test_snapshots_recovery.py
```diff
@@ -0,0 +1,130 @@
...
+        snapshot_file = tmp_path / "test_collection.snapshot"
+        with open(snapshot_file, 'wb') as f:
+            f.write(snapshot_content)
+        
+        # Test 1: Recover from URL
+        snapshot_url = f"http://{client.host}:{client.port}/collections/test_collection/snapshots/{snapshot_name}"
+        client.recover_snapshot_from_url("test_collection_recovered_1", snapshot_url, snapshot_checksum)
+        
+        # Verify recovered collection exists
+        client.verify_collection_exists("test_collection_recovered_1")
```
KShivendu: Maybe check `points_count`, `vectors_count`, `config`, etc returned by this? We can do the same for `recovered_2`

tests/e2e_tests/test_snapshots_recovery.py
```diff
@@ -0,0 +1,130 @@
...
+        shard_snapshot_url = f"http://{client.host}:{client.port}/collections/test_collection/shards/0/snapshots/{shard_snapshot_name}"
+        client.recover_shard_snapshot_from_url("test_collection_recovered_1", 0, shard_snapshot_url)
+        
+        # Test 4: Upload shard snapshot as file
+        client.upload_shard_snapshot_file("test_collection_recovered_2", 0, shard_snapshot_content)
+        
+        # Final verification - ensure all collections are accessible
+        for collection_name in ["test_collection_recovered_1", "test_collection_recovered_2"]:
+            collection_info = client.verify_collection_exists(collection_name)
+            assert collection_info["result"]["status"] == "green", f"Collection {collection_name} is not in green status"
```
KShivendu: Consider verifying other params as well. Like I suggested in https://github.com/qdrant/qdrant/pull/7008/files#r2303193358

tests/e2e_tests/test_snapshots_recovery.py
tellet-q: The python code doesn't exactly mimic this behavior, but it prepares config file just like the bash code, builds the image (qdrant_image fixture) and then runs the container with the provided config (qdrant_container_factory fixture). There is no docker mounts as bash test didn't have any related assertions, so I decided to completely omit the volumes.

KShivendu (approved):
Thanks!

PR #7009, opened 2025-08-11
BQ features without vector stats

IvanPleshkov:
Appendable BQ cannot have vector statistics. To support 2bit BQ and asymmetric queries in appendable quantization, this PR adds support for these features even if there is no vector statistics in BQ.

Additional tests are not required here, current tests cover changes in query encoding, 2bit encoding case is too trivial to have a separate test for it

### DIFF
--- a/lib/quantization/src/encoded_vectors_binary.rs
+++ b/lib/quantization/src/encoded_vectors_binary.rs
@@ -518,12 +518,8 @@ impl<TBitsStoreType: BitsStoreType, TStorage: EncodedStorage>
 
         let bits_count = u8::BITS as usize * std::mem::size_of::<TBitsStoreType>();
         let one = TBitsStoreType::one();
-        for (i, &v) in vector.iter().enumerate() {
-            let mean = vector_stats.elements_stats[i].mean;
-            let sd = vector_stats.elements_stats[i].stddev;
-
-            let (b1, b2) = Self::encode_two_bits_value(v, mean, sd);
-
+        for i in 0..vector.len() {
+            let (b1, b2) = Self::encode_two_bits_value(vector, i, vector_stats);
             if b1 {
                 encoded_vector[i / bits_count] |= one << (i % bits_count);
             }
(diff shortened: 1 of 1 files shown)


lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -617,34 +611,28 @@ impl<TBitsStoreType: BitsStoreType, TStorage: EncodedStorage>
                 EncodedQueryBQ::Binary(Self::encode_vector(query, vector_stats, encoding))
             }
             QueryEncoding::Scalar8bits => EncodedQueryBQ::Scalar8bits(
-                Self::encode_scalar_query_vector(query, vector_stats, encoding, u8::BITS as usize),
+                Self::encode_scalar_query_vector(query, encoding, u8::BITS as usize),
+            ),
+            QueryEncoding::Scalar4bits => EncodedQueryBQ::Scalar4bits(
+                Self::encode_scalar_query_vector(query, encoding, (u8::BITS / 2) as usize),
             ),
```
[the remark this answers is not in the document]
IvanPleshkov: As you can see in changes:
```
        let Some(vector_stats) = vector_stats else {
            if value > 0.0 {
                return (true, true);
            } else {
                return (false, false);
            }
        };
```
If vector statistics are presented, there is an another code pass. In this case, if vector stats are presented, final accuracy will be higher

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -575,7 +554,11 @@ impl<TBitsStoreType: BitsStoreType, TStorage: EncodedStorage>
         }
     }
 
-    fn encode_two_bits_value(value: f32, mean: f32, sd: f32) -> (bool, bool) {
+    fn encode_two_bits_value(
+        vector: &[f32],
+        i: usize,
```
generall: why not just `f32`? It is only 4 bytes on stack, instead of 4 + 4 for current version

IvanPleshkov: Thanks for fixing!

PR #7032, opened 2025-08-12
Allow updating WAL config in collection after creation

KShivendu:
Allow directly modifying WAL config via API call even after collection is created:

```http
PATCH /collections/benchmark
{
  "wal_config": {
    "wal_capacity_mb": "20",
    "wal_segments_ahead": "10",
    "wal_retain_closed": 3
  }
}
```

This makes it more likely for Qdrant to do WAL delta transfers (which is faster and preferred for large clusters)

Note that changes are effectively only after restarting Qdrant node.

Related to https://github.com/qdrant/qdrant/pull/6976#pullrequestreview-3088049521

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -10323,6 +10323,17 @@
               }
             ]
           },
+          "wal_config": {
+            "description": "WAL parameters for the collection. Changes are effective only on Qdrant restart. If none - it is left unchanged.",
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/WalConfigDiff"
+              },
+              {
+                "nullable": true
+              }
+            ]
+          },
           "quantization_config": {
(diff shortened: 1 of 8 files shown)


KShivendu:
@generall suggested that this is a bit risky especially when user wants to decrease the WAL capacity. We can lose some data (if it's not applied yet). Also, nodes must be restarted for changes to be effective, and restarts can take significant time.

Possible improvements:
- Only allow increasing the values, not decreasing?
- Mainly allow changing `wal_retain_closed` and that's always gonna be >= 1

We have seen this issue only with one large customer so we can postpone this PR till it's actually required in more real world clusters.

cc @timvisee

xzfc:
Merging #7475 introduced merge conflicts in this branch, so I rebased it.

KShivendu:
We think it's better to do it like https://github.com/qdrant/qdrant/pull/7834 and hence closing this PR.

PR #7043, opened 2025-08-12
remove quantization save

IvanPleshkov:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/quantization/src/encoded_storage.rs
+++ b/lib/quantization/src/encoded_storage.rs
@@ -3,6 +3,8 @@ use std::fs::File;
 #[cfg(feature = "testing")]
 use std::io::{Read, Write};
 use std::path::Path;
+#[cfg(feature = "testing")]
+use std::path::PathBuf;
 
 use common::counter::hardware_counter::HardwareCounterCell;
 #[cfg(feature = "testing")]
@@ -14,8 +16,6 @@ pub trait EncodedStorage {
     fn from_file(path: &Path, quantized_vector_size: usize) -> std::io::Result<Self>
     where
         Self: Sized;
-
-    fn save_to_file(&self, path: &Path) -> std::io::Result<()>;
 
     fn is_on_disk(&self) -> bool;
 
@@ -66,13 +66,6 @@ impl EncodedStorage for TestEncodedStorage {
         Ok(Self { data: buffer })
     }
 
-    fn save_to_file(&self, path: &Path) -> std::io::Result<()> {
-        let mut buffer = File::create(path)?;
-        buffer.write_all(self.data.as_slice())?;
-        buffer.sync_all()?;
-        Ok(())
-    }
-
     fn is_on_disk(&self) -> bool {
         false
     }
@@ -88,19 +81,16 @@ impl EncodedStorage for TestEncodedStorage {
 #[cfg(feature = "testing")]
 pub struct TestEncodedStorageBuilder {
     data: Vec<u8>,
+    path: Option<PathBuf>,
 }
 
 #[cfg(feature = "testing")]
 impl TestEncodedStorageBuilder {
-    pub fn new() -> Self {
-        Self { data: Vec::new() }
-    }
-}
-
-#[cfg(feature = "testing")]
-impl Default for TestEncodedStorageBuilder {
-    fn default() -> Self {
-        Self::new()
+    pub fn new(path: Option<&std::path::Path>) -> Self {
+        Self {
+            data: Vec::new(),
+            path: path.map(PathBuf::from),
+        }
     }
 }
 
@@ -109,6 +99,14 @@ impl EncodedStorageBuilder for TestEncodedStorageBuilder {
     type Storage = TestEncodedStorage;
 
     fn build(self) -> std::io::Result<Self::Storage> {
+        if let Some(path) = &self.path {
+            if let Some(dir) = path.parent() {
+                std::fs::create_dir_all(dir)?;
+            }
+            let mut file = File::create(path)?;
+            file.write_all(&self.data)?;
+            file.sync_all()?;
+        }
         Ok(TestEncodedStorage { data: self.data })
     }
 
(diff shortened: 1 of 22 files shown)


lib/quantization/src/encoded_storage.rs
```diff
@@ -88,19 +81,16 @@ impl EncodedStorage for TestEncodedStorage {
...
-    pub fn new() -> Self {
-        Self { data: Vec::new() }
-    }
-}
-
-#[cfg(feature = "testing")]
-impl Default for TestEncodedStorageBuilder {
-    fn default() -> Self {
-        Self::new()
+    pub fn new(path: Option<&std::path::Path>) -> Self {
```
timvisee, line 89: Self-check: `Into<PathBuf>` instead?

lib/segment/src/vector_storage/quantized/quantized_vectors.rs
```diff
@@ -860,9 +851,9 @@ impl QuantizedVectors {
                 vector_parameters,
                 inner_vectors_count,
                 scalar_config.quantile,
+                Some(meta_path.as_path()),
                 stopped,
             )?;
-            let offsets_path = path.join(QUANTIZED_OFFSETS_PATH);
             create_offsets_file_from_iter(&offsets_path, vectors_count, offsets)?;
```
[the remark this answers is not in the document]
IvanPleshkov: In this case, we create a mmap file and load the mmapped file as a storage for multivector offsets

lib/segment/src/vector_storage/quantized/quantized_mmap_storage.rs
```diff
@@ -82,8 +77,16 @@ impl QuantizedMmapStorageBuilder {
...
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "quantized_vector_size must be > 0",
+            ));
+        }
         let encoded_storage_size = quantized_vector_size * vectors_count;
-        path.parent().map(std::fs::create_dir_all);
+        if let Some(dir) = path.parent() {
+            std::fs::create_dir_all(dir)?;
+        }
```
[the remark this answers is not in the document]
IvanPleshkov: This PR https://github.com/qdrant/qdrant/pull/7048 has already these changes, which are proposed by you

lib/segment/src/vector_storage/quantized/quantized_ram_storage.rs
```diff
@@ -71,20 +59,34 @@ impl quantization::EncodedStorage for QuantizedRamStorage {
...
+            path: path.to_path_buf(),
+        })
     }
 }
 
 impl quantization::EncodedStorageBuilder for QuantizedRamStorageBuilder {
     type Storage = QuantizedRamStorage;
 
     fn build(self) -> std::io::Result<QuantizedRamStorage> {
+        self.path.parent().map(std::fs::create_dir_all);
```
timvisee: Let's properly catch the error here in both these cases:
- if we don't have a parent path, the path is invalid
- if we fail to create the parent directories

lib/segment/src/vector_storage/quantized/quantized_mmap_storage.rs
```diff
@@ -82,8 +77,16 @@ impl QuantizedMmapStorageBuilder {
...
     ) -> std::io::Result<Self> {
+        if quantized_vector_size == 0 {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "quantized_vector_size must be > 0",
+            ));
+        }
         let encoded_storage_size = quantized_vector_size * vectors_count;
-        path.parent().map(std::fs::create_dir_all);
+        if let Some(dir) = path.parent() {
```
timvisee: Should we also error if the parent is `None`?

lib/quantization/src/encoded_storage.rs
timvisee: Tried it locally, but the compiler starts complaining that we need to define the type in all cases we pass `None`. That isn't nice, so let's skip this idea.

lib/segment/src/vector_storage/quantized/quantized_mmap_storage.rs
IvanPleshkov: I don't know. I don't expect in practice if the path has no parent, and I don't have a strong opinion on what to do. I can return `Err` if you think we should do it. Here, I do parent creation if we can, and file creation will fail if the path is incorrect

lib/segment/src/vector_storage/quantized/quantized_ram_storage.rs
IvanPleshkov: Good catch! Fixed like in mmap storage:
```
if let Some(dir) = self.path.parent() {
    std::fs::create_dir_all(dir)?;
}
```

Strangely, CodeRabbit didn't mention it. Bot triggers on these mistakes.

lib/segment/src/vector_storage/quantized/quantized_mmap_storage.rs
timvisee: Docs: https://doc.rust-lang.org/std/path/struct.Path.html#method.parent

It'll be `None` if the path is `/`. Which I consider to be invalid.

I'd argue it's fine to disallow that here.

IvanPleshkov: Fixed in all places related to quantization

lib/segment/src/vector_storage/quantized/quantized_multivector_storage.rs
```diff
@@ -371,7 +337,14 @@ pub(super) fn create_offsets_file_from_iter(
...
 ) -> OperationResult<()> {
-    path.parent().map(std::fs::create_dir_all);
+    path.parent()
+        .ok_or_else(|| {
+            std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "Path must have a parent directory",
+            )
+        })
+        .and_then(std::fs::create_dir_all)?;
```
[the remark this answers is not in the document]
IvanPleshkov: In this conversation, we decided to return an error if the path does not have a parent:
https://github.com/qdrant/qdrant/pull/7043#discussion_r2278962896

Issue #6826, opened 2025-07-08
Fix criterion sparse_index_build


agourlay:
```bash
 RUST_BACKTRACE=1 cargo bench --all -p segment --bench sparse_index_build -- --nocapture
```

```
Benchmarking sparse-vector-build-group/build-ram-index: Warming up for 3.0000 s
thread 'main' panicked at lib/sparse/src/index/inverted_index/inverted_index_ram.rs:18:9:
InvertedIndexRam is not supposed to be versioned
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
   2: <sparse::index::inverted_index::inverted_index_ram::Version as io::storage_version::StorageVersion>::current_raw
             at /home/agourlay/Workspace/qdrant/lib/sparse/src/index/inverted_index/inverted_index_ram.rs:18:9
   3: io::storage_version::StorageVersion::current
             at /home/agourlay/Workspace/qdrant/lib/common/io/src/storage_version.rs:20:9
   4: segment::index::sparse_index::sparse_vector_index::SparseVectorIndex<TInvertedIndex>::try_load
             at ./src/index/sparse_index/sparse_vector_index.rs:174:12
   5: segment::index::sparse_index::sparse_vector_index::SparseVectorIndex<TInvertedIndex>::open
             at ./src/index/sparse_index/sparse_vector_index.rs:118:13
   6: sparse_index_build::sparse_vector_index_build_benchmark::{{closure}}::{{closure}}
             at ./benches/sparse_index_build.rs:87:17
   7: criterion::bencher::Bencher<M>::iter
             at /home/agourlay/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/criterion-0.6.0/src/bencher.rs:91:23
   8: sparse_index_build::sparse_vector_index_build_benchmark::{{closure}}
             at ./benches/sparse_index_build.rs:85:9
   9: criterion::benchmark_group::BenchmarkGroup<M>::bench_function::{{closure}}
             at /home/agourlay/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/criterion-0.6.0/src/benchmark_group.rs:253:60
[...]

PR #7044, opened 2025-08-12
fix criterion sparse_index_build #6826

guglielmo-monici:
sparse_index_build: use is_on_disk (Mmap) instead of is_persisted (Mmap, ImmutableRam) in indices_tracker and inverted_tracker save logic (#6826)

# Description:
When **SparseVectorIndex** open's method is call

| sparse_vector_index.rs
immagine

there is a condition evaluation on **config.index_type.is_persisted()**

when it's false the method save on disk these files:
- config
- inverted_index 
- inverted_tracker

immagine

the bench panic because **sparse_index_ram** must not be save and hasn't a semver version.

| inverted_index_ram.rs
immagine

I search on git blame and I have found this change: **InvertedIndexImmutableRam and index migrations #4220** by xzfc 
that added a new type **InvertedIndexImmutableRam**

before this change the saving's logic was evaluated on **if !self.is_appendable**
and **appendable** was equal to **MutableRam**:
```
        let is_appendable = config.index_type == SparseIndexType::MutableRam;
```

```
        // save inverted index
        if self.config.index_type.is_persisted() {
            self.indices_tracker.save(&self.path)?;
            self.inverted_index.save(&self.path)?;
        }
```

So I understand that the previous logic saved it only for **Mmap** type.

but after #4220 change, it was added the type **ImmutableRam** and the logic is evaluated on **config.index_type.is_persisted()** that become equal to **Mmap || ImmutableRam**
so the method try to save also when is ImmutableRam that panic for semver version.

I change the condition to **config.index_type.is_on_disk()** that is equal to **Mmap**
This is also more clear with a positive logic instead of negative

I also change the shell.nix file due to an error of lib compatibility by run test and bench.
Before change and after start nix-shell I get these errors:
```
  qdrant@1.15.3-dev requires rustc 1.88
  sysinfo@0.37.0 requires rustc 1.88
```

### All Submissions:


lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -103,18 +103,7 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
...
-            let (inverted_index, indices_tracker) = Self::build_inverted_index(
-                &id_tracker,
-                &vector_storage,
-                path,
-                stopped,
-                tick_progress,
-            )?;
-            (config, inverted_index, indices_tracker)
-        } else {
+        let (config, inverted_index, indices_tracker) = if config.index_type.is_on_disk() {
```
coszio: This change will never create an `ImmutableRam` variant.

coszio (changes requested):
After reproducing the panic in benchmark, the error seems to be that we are requesting an [`ImmutableRam`](https://github.com/qdrant/qdrant/blob/1b06abf0633efe09f7e01e61d1593046db9e2a59/lib/segment/benches/sparse_index_build.rs#L78) index in config, but using [`SparseVectorIndex<InvertedIndexRam>`](https://github.com/qdrant/qdrant/blob/1b06abf0633efe09f7e01e61d1593046db9e2a59/lib/segment/benches/sparse_index_build.rs#L87) as type for the loader. 
This would try to use the configured variant's version, but with the behavior of the specified generic type.

I believe the right fix would be to request `MutableRam` variant in the config for the benchmark, as the other benches also depend on a mutable variant of the index.

shell.nix
```diff
@@ -33,9 +33,9 @@ let
     in
     fenix.combine [
       nightly.rustfmt # should be the first
-      stable.rust
-      stable.rust-analyzer
-      stable.rust-src
+      nightly.rust
+      nightly.rust-analyzer
+      nightly.rust-src
```
timvisee: I believe we only want to use `rustfmt` from nightly.

cc @xzfc

timvisee: I'd be happy to merge if you remove these Nix changes, as they are unrelated to the benchmark problem.

xzfc: > I also change the shell.nix file due to an error of lib compatibility by run test and bench.
> Before change and after start nix-shell I get these errors:
> ```
>   qdrant@1.15.3-dev requires rustc 1.88
>   sysinfo@0.37.0 requires rustc 1.88
> ```

It's caused by pinned Rust version not being updated (sorry for that). I made a switch to `rustup` (no more pinned Rust versions) in #7049.

PR #7048, opened 2025-08-13
EncodedStorage update vector

IvanPleshkov:
This PR changes signature of vector updating in quantization:

```
From
EncodedStorage::push_vector

Into
EncodedStorage::update_vector
```

Push is not enough for an appendable segment where the user may update the vector. If `id` is not existing, quantization storage should create new one.

Also, this PR simplifies `EncodedStorage` functions and removes `quantized_vector_size` argument where possible.


lib/quantization/src/encoded_storage.rs
```diff
@@ -20,13 +20,14 @@ pub trait EncodedStorage {
 
     fn is_on_disk(&self) -> bool;
 
-    fn push_vector(
+    fn update_vector(
         &mut self,
+        id: u32,
```
timvisee: Please use `PointOffsetType` everywhere instead, which is an alias for `u32`

lib/quantization/src/encoded_storage.rs
```diff
@@ -116,10 +131,24 @@ impl EncodedStorageBuilder for TestEncodedStorageBuilder {
...
+        match self.quantized_vector_size {
+            Some(size) => {
+                if other.len() != size {
+                    panic!("Data length is not equal to quantized vector size");
+                }
+            }
+            None => {
+                self.quantized_vector_size = Some(other.len());
+            }
+        }
```
timvisee: Can't we somehow enforce to specify the quantized size when creating this structure, so we don't have to derive it now?

lib/quantization/src/encoded_storage.rs
xzfc: Also, `index` in `get_vector_data()` is `usize`.
Shouldn't they be consistent (both `usizes` OR both `u32`/`PointOffsetType`)?

lib/quantization/src/encoded_storage.rs
```diff
@@ -20,13 +20,14 @@ pub trait EncodedStorage {
 
     fn is_on_disk(&self) -> bool;
 
-    fn push_vector(
+    fn update_vector(
```
xzfc: Naming nit: I think names `set_vector()` or `upsert_vector()` would make more sense. (and same for `update_offset()`)

Also, it worth add a comment that setting/updating past the container size resizes it. (otherwise I could assume it's an error)

lib/quantization/src/encoded_storage.rs
IvanPleshkov: I agree with you. Can I do it as a separate PR? Your proposal causes changes in each quantization test.

lib/quantization/src/encoded_storage.rs
IvanPleshkov: I agree with you. But the whole quantization crate avoids `PointOffsetType`, It should be a separate PR with all replacements `u32 -> PointOffsetType`

IvanPleshkov: I can do such replacement as a separate change

lib/quantization/src/encoded_storage.rs
IvanPleshkov: For now I'm going to use an advice from CodeRabbit as a tempopary solution:
https://github.com/qdrant/qdrant/pull/7048#discussion_r2273197692

lib/quantization/src/encoded_storage.rs
IvanPleshkov: https://github.com/qdrant/qdrant/pull/7058

lib/quantization/src/encoded_storage.rs
```diff
@@ -42,31 +43,41 @@ pub trait EncodedStorageBuilder {
...
-        self.data.try_reserve(vector.len())?;
-        self.data.extend_from_slice(vector);
+        if id as usize >= self.data.len() / self.quantized_vector_size {
+            self.data
+                .resize((id as usize + 1) * self.quantized_vector_size, 0);
+        }
+        self.data[id as usize * self.quantized_vector_size
+            ..(id as usize + 1) * self.quantized_vector_size]
+            .copy_from_slice(vector);
         Ok(())
```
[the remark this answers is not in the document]
IvanPleshkov: fixed

lib/segment/src/vector_storage/quantized/quantized_ram_storage.rs
```diff
@@ -16,18 +16,19 @@ pub struct QuantizedRamStorage {
...
+        id: u32,
         vector: &[u8],
         _hw_counter: &HardwareCounterCell,
     ) -> std::io::Result<()> {
         // Skip hardware counter increment because it's a RAM storage.
         self.vectors
-            .push(vector)
+            .insert(id as usize, vector)
             .map_err(|err| std::io::Error::new(std::io::ErrorKind::OutOfMemory, err.to_string()))?;
         Ok(())
```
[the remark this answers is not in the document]
IvanPleshkov: fixed

lib/quantization/src/encoded_storage.rs
```diff
@@ -42,31 +43,61 @@ pub trait EncodedStorageBuilder {
...
```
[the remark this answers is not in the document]
IvanPleshkov: It's too complicated for the test storage. Test storage might panic because it's not used in production

lib/quantization/src/encoded_storage.rs
IvanPleshkov: Merged and fixed for `update_vector` here

lib/quantization/src/encoded_vectors_binary.rs
```diff
@@ -884,29 +877,27 @@ impl<TBitsStoreType: BitsStoreType, TStorage: EncodedStorage> EncodedVectors
...
         hw_counter: &HardwareCounterCell,
     ) -> std::io::Result<()> {
         let encoded_vector = Self::encode_vector(vector, &None, self.metadata.encoding);
-        self.encoded_vectors.push_vector(
+        self.encoded_vectors.update_vector(
+            id,
             bytemuck::cast_slice(encoded_vector.encoded_vector.as_slice()),
             hw_counter,
         )
     }
```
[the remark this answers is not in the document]
IvanPleshkov: Good catch! It's expected that vector stats are empty but I agree that It's better to use struct field even if we are sure that field is None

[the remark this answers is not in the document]
IvanPleshkov: fixed

lib/segment/src/vector_storage/quantized/quantized_mmap_storage.rs
```diff
@@ -21,32 +22,49 @@ impl QuantizedMmapStorage {
...
     fn from_file(
         path: &Path,
-        _quantized_vector_size: usize,
+        quantized_vector_size: usize,
     ) -> std::io::Result<QuantizedMmapStorage> {
         let file = std::fs::OpenOptions::new().read(true).open(path)?;
         let mmap = unsafe { Mmap::map(&file)? };
         madvise::madvise(&mmap, madvise::get_global())?;
-        Ok(Self { mmap })
+        if quantized_vector_size == 0 || mmap.len() % quantized_vector_size != 0 {
```
xzfc: Nit: [`.is_multiple_of()`](https://doc.rust-lang.org/std/primitive.usize.html#method.is_multiple_of) could be used.

lib/segment/src/vector_storage/quantized/quantized_mmap_storage.rs
```diff
@@ -58,8 +76,11 @@ impl quantization::EncodedStorage for QuantizedMmapStorage {
         true
     }
 
-    fn vectors_count(&self, quantized_vector_size: usize) -> usize {
-        self.mmap.len() / quantized_vector_size
+    fn vectors_count(&self) -> usize {
+        self.mmap
+            .len()
+            .checked_div(self.quantized_vector_size)
+            .unwrap_or_default()
```
xzfc: Nit: we can avoid these checks if `quantized_vector_size` be `NonZero<usize>`.

lib/quantization/src/encoded_storage.rs
```diff
@@ -43,31 +44,65 @@ pub trait EncodedStorageBuilder {
...
         // Skip hardware counter increment because it's a RAM storage.
-        self.data.try_reserve(vector.len())?;
-        self.data.extend_from_slice(vector);
+        if id as usize >= self.data.len() / self.quantized_vector_size {
+            self.data
+                .resize((id as usize + 1) * self.quantized_vector_size, 0);
+        }
+        self.data[id as usize * self.quantized_vector_size
+            ..(id as usize + 1) * self.quantized_vector_size]
+            .copy_from_slice(vector);
```
xzfc: Style nit:

```suggestion
        let off = id as usize * self.quantized_vector_size;
        if id as usize >= self.vectors_count() {
            self.data
                .resize(off + self.quantized_vector_size, 0);
        }
        self.data[off..off + self.quantized_vector_size]
            .copy_from_slice(vector);
```

lib/segment/src/vector_storage/quantized/quantized_multivector_storage.rs
```diff
@@ -340,8 +349,9 @@ where
         Some(query)
     }
 
-    fn push_vector(
+    fn update_vector(
         &mut self,
+        id: u32,
```
xzfc: Nit: `u32` → `PointOffsetType` here and in few other places.

```console
$ git diff --merge-base origin/dev | grep '^+.*u32' -C2
+    fn update_vector(
         &mut self,
+        id: u32,
         vector: &[f32],
--
+    fn update_vector(
         &mut self,
+        _id: u32,
         _vector: &[u8],
         _hw_counter: &HardwareCounterCell,
--
+    fn update_vector(
         &mut self,
+        id: u32,
         vector: &[f32],
--
+    fn update_vector(
         &mut self,
+        id: u32,
         vector: &[u8],
         _hw_counter: &HardwareCounterCell,
```

IvanPleshkov: Thanks! Fixed!

lib/quantization/src/encoded_storage.rs
IvanPleshkov: Fixed

lib/segment/src/vector_storage/quantized/quantized_mmap_storage.rs
IvanPleshkov: Good idea, also applied in the test storage

lib/segment/src/vector_storage/quantized/quantized_mmap_storage.rs
IvanPleshkov: Done

lib/quantization/src/encoded_storage.rs
IvanPleshkov: Not a nit. `upsert_vector` is better because later here: https://github.com/qdrant/qdrant/pull/7060
We call this function from `VectorIndex::upsert_vector`

IvanPleshkov: Fixed

lib/quantization/src/encoded_storage.rs
IvanPleshkov: Fixed, it increases x2 times the amount of changes in this PR, but all additions are trivial where we call `get_quantized_vector_size` and specify the quantized size into builder

lib/segment/src/vector_storage/quantized/quantized_mmap_storage.rs
```diff
@@ -21,32 +23,54 @@ impl QuantizedMmapStorage {
...
+        Ok(Self {
+            mmap,
+            quantized_vector_size: NonZeroUsize::new(quantized_vector_size).ok_or_else(|| {
+                std::io::Error::new(
+                    std::io::ErrorKind::InvalidInput,
+                    "`quantized_vector_size` must be non-zero",
+                )
+            })?,
+        })
     }
```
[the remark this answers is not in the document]
IvanPleshkov: `usize::is_multiple_of` does not trigger division-by-zero panic. But I still refactored this code to check if `quantized_vector_size` is zero first.

lib/quantization/benches/binary.rs
```diff
@@ -75,9 +80,14 @@ fn binary_bench(c: &mut Criterion) {
...
+    let quantized_vector_size =
+        EncodedVectorsBin::<u128, TestEncodedStorage>::get_quantized_vector_size_from_params(
+            vector_dim,
+            Encoding::OneBit,
+        );
     let encoded_u8 = EncodedVectorsBin::<u8, _>::encode(
         vectors.iter(),
-        TestEncodedStorageBuilder::new(),
+        TestEncodedStorageBuilder::new(quantized_vector_size),
         &VectorParameters {
```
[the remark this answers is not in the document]
IvanPleshkov: good catch! fixed

lib/quantization/benches/binary.rs
```diff
@@ -186,9 +201,14 @@
...
```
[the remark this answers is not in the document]
IvanPleshkov: good catch! fixed

lib/quantization/tests/integration/empty_storage.rs
```diff
@@ -99,9 +106,14 @@ mod tests {
...
 
+        let quantized_vector_size =
+            EncodedVectorsBin::<u128, TestEncodedStorage>::get_quantized_vector_size_from_params(
+                vector_dim,
+                quantization::encoded_vectors_binary::Encoding::OneBit,
+            );
         let encoded = EncodedVectorsBin::<u8, _>::encode(
             vector_data.iter(),
-            TestEncodedStorageBuilder::new(),
+            TestEncodedStorageBuilder::new(quantized_vector_size),
```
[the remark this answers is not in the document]
IvanPleshkov: good catch! fixed

lib/quantization/src/encoded_vectors_pq.rs
```diff
@@ -580,22 +572,22 @@ impl<TStorage: EncodedStorage> EncodedVectors for EncodedVectorsPQ<TStorage> {
...
         _hw_counter: &HardwareCounterCell,
     ) -> std::io::Result<()> {
-        debug_assert!(false, "PQ does not support push_vector",);
+        debug_assert!(false, "PQ does not support update_vector",);
         Err(std::io::Error::new(
             std::io::ErrorKind::Unsupported,
-            "PQ does not support push_vector",
+            "PQ does not support update_vector",
         ))
     }
```
timvisee:
```suggestion
        debug_assert!(false, "PQ does not support upsert_vector",);
Err(std::io::Error::new(
std::io::ErrorKind::Unsupported,
            "PQ does not support upsert_vector",
))
}
```

lib/quantization/src/encoded_storage.rs
```diff
@@ -43,31 +46,73 @@ pub trait EncodedStorageBuilder {
...
         &mut self,
+        id: PointOffsetType,
         vector: &[u8],
         _hw_counter: &HardwareCounterCell,
     ) -> std::io::Result<()> {
+        if vector.len() != self.quantized_vector_size.get() {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                format!(
+                    "update_vector: payload length {} != quantized_vector_size {}",
```
timvisee:
```suggestion
                    "upsert_vector: payload length {} != quantized_vector_size {}",
```

lib/quantization/src/encoded_storage.rs
```diff
@@ -43,31 +46,73 @@ pub trait EncodedStorageBuilder {
...
+        }
+        Ok(Self {
+            data: buffer,
+            quantized_vector_size: NonZeroUsize::new(quantized_vector_size).ok_or_else(|| {
+                std::io::Error::new(
+                    std::io::ErrorKind::InvalidInput,
+                    "`quantized_vector_size` must be non-zero",
+                )
+            })?,
+        })
```
timvisee: Nit: I consider the conditions a bit better defined like this:


```suggestion
        let mut file = OneshotFile::open(path)?;
        let mut buffer = Vec::new();
        file.read_to_end(&mut buffer)?;
        file.drop_cache()?;
        if !buffer.len().is_multiple_of(quantized_vector_size) {
            return Err(std::io::Error::new(
                std::io::ErrorKind::InvalidData,
                format!(
                    "TestEncodedStorage: buffer size ({}) not divisible by quantized_vector_size ({})",
                    buffer.len(),
                    quantized_vector_size,
                ),
            ));
        }
        let quantized_vector_size = NonZeroUsize::new(quantized_vector_size).ok_or_else(|| {
            std::io::Error::new(
                std::io::ErrorKind::InvalidInput,
                "`quantized_vector_size` must be non-zero",
            )
        })?;
        Ok(Self {
            data: buffer,
            quantized_vector_size,
        })
```

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -505,23 +501,21 @@ impl<TStorage: EncodedStorage> EncodedVectors for EncodedVectorsU8<TStorage> {
...
         _vector: &[f32],
         _hw_counter: &HardwareCounterCell,
     ) -> std::io::Result<()> {
-        debug_assert!(false, "SQ does not support push_vector",);
+        debug_assert!(false, "SQ does not support update_vector",);
         Err(std::io::Error::new(
             std::io::ErrorKind::Unsupported,
-            "SQ does not support push_vector",
+            "SQ does not support update_vector",
         ))
```
timvisee:
```suggestion
        debug_assert!(false, "SQ does not support upsert_vector",);
Err(std::io::Error::new(
std::io::ErrorKind::Unsupported,
            "SQ does not support upsert_vector",
))
```

lib/quantization/src/encoded_vectors_pq.rs
IvanPleshkov: Good catch! Fixed

PR #7050, opened 2025-08-13
Always load InferenceService

JojiiOfficial:
Fixes BM25 inference, which failed due to InferenceService not being loaded.

Also adds a test for this case.

### DIFF
--- a/src/common/inference/service.rs
+++ b/src/common/inference/service.rs
@@ -76,7 +76,8 @@ pub struct InferenceService {
 static INFERENCE_SERVICE: RwLock<Option<Arc<InferenceService>>> = RwLock::new(None);
 
 impl InferenceService {
-    pub fn new(config: InferenceConfig) -> Self {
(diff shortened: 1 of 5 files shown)


src/common/inference/service.rs
```diff
@@ -299,7 +300,8 @@ impl InferenceService {
     }
 
     fn is_address_valid(&self) -> bool {
-        !self.config.address.as_ref().is_none_or(|i| i.is_empty())
+        self.config.address.is_none() // In BM25 we don't need an address so we allow InferenceService to have an empty address.
+        || self.config.address.as_ref().is_some_and(|i| !i.is_empty())
```
timvisee, line 304: Shall we do _more_ validation, and verify it's a valid host?

generall: It looks like extra functionality for another PR

timvisee (approved):
Thanks for the test!

PR #7051, opened 2025-08-13
Add warning to development.yaml config

JojiiOfficial:
Depends on #7050 

Adds a warning comment to development.yaml to make clear that the changes there should be tested on release profile as well.


timvisee (approved):
I don't think this is necessary, but, I also don't object to the change

PR #7056, opened 2025-08-13
Fix bm25 params

generall:


src/common/inference/bm25.rs
```diff
@@ -147,17 +98,40 @@ impl Bm25 {
...
+
+    let stopwords_config = match stopwords {
+        None => {
+            // Try to create from the language
+            Language::from_str(&language)
+                .ok()
+                .map(StopwordsInterface::Language)
+        }
+        Some(stopwords_interface) => Some(stopwords_interface),
+    };
```
generall, line 128: main change is here

lib/api/src/rest/schema.rs
```diff
@@ -161,22 +164,184 @@ impl Hash for Options {
...
+            DocumentOptions::Bm25(bm25) => bm25.to_options(),
+        }
+    }
+}
+
+impl Hash for DocumentOptions {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let options = match self {
+            DocumentOptions::Common(options) => Cow::Borrowed(options),
+            DocumentOptions::Bm25(bm25) => Cow::Owned(bm25.to_options()),
```
generall, line 307: Don't expect `Bm25` branch to ever be used, as we don't deserialize into `Bm25`

lib/api/src/rest/schema.rs
```diff
@@ -161,22 +164,184 @@ impl Hash for Options {
...
+        // Bm25 option is used only for schema, actual deserialization will happen in specialized code
+        assert!(matches!(options, DocumentOptions::Common(_)));
+    }
+}
+
+impl DocumentOptions {
+    pub fn into_options(self) -> HashMap<String, Value> {
+        match self {
+            DocumentOptions::Common(options) => options,
+            DocumentOptions::Bm25(bm25) => bm25.to_options(),
```
JojiiOfficial, line 298: If we're never going to serialize the bm25 variant, I'd suggest to make this clear by using `unreachable!("BM25 variant never gets created");` so we will notice when we accidentally create it.
The same below

lib/api/src/rest/schema.rs
```diff
@@ -161,22 +164,184 @@ impl Hash for Options {
...
+
+impl Hash for DocumentOptions {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let options = match self {
+            DocumentOptions::Common(options) => Cow::Borrowed(options),
+            DocumentOptions::Bm25(bm25) => Cow::Owned(bm25.to_options()),
+        };
+
+        // Order of keys in the map should not affect the hash
+        let mut keys: Vec<_> = options.keys().collect();
```
JojiiOfficial, line 311: Can we maybe just collect the options into a `BTreeMap` instead? I think this custom hashing approach might cause hard to detect bugs in future.
Alternatively we could just replace the initial `HashMap` with a `BTreeMap`, the APIs should be the same.

lib/api/src/rest/schema.rs
```diff
@@ -161,22 +164,184 @@ impl Hash for Options {
...
+        let mut options = HashMap::new();
+        options.insert("k".to_string(), Value::from(k.into_inner()));
+        options.insert("b".to_string(), Value::from(b.into_inner()));
+        options.insert("avg_len".to_string(), Value::from(avg_len.into_inner()));
+        options.insert(
+            "tokenizer".to_string(),
+            serde_json::to_value(tokenizer)
+                .expect("conversion of internal structure to JSON should never fail"),
+        );
+        options.insert(
```
JojiiOfficial: The `text_preprocessing_config` is flattened in the original config. I think we should flatten here as well, so the values get applied correctly.

lib/api/src/rest/schema.rs
```diff
@@ -161,22 +164,184 @@ impl Hash for Options {
...
+    pub min_token_len: Option<usize>,
+    /// Maximum token length to keep. If token is longer than this, it will be discarded.
+    /// Default is `None`, which means no maximum length.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub max_token_len: Option<usize>,
+}
+
+impl Bm25Config {
+    pub fn to_options(&self) -> HashMap<String, Value> {
+        let Self {
```
JojiiOfficial: I think this approach might not be the best. Manually serializing the fields into a HashMap makes it very easy to forget new fields or changed attributes in the original struct.

Also if I'm seeing it correctly, this function is never called because we always prefer `DocumentOptions::Common` over `DocumentOptions::Bm25` and this function is only called for `DocumentOptions::Bm25`

lib/api/src/rest/schema.rs
```diff
@@ -161,22 +164,184 @@ impl Hash for Options {
...
+}
+
+/// Bm25 tokenizer configurations.
+#[derive(Debug, Deserialize, Serialize, Clone, Default, JsonSchema, PartialEq, Eq)]
+pub struct TextPreprocessingConfig {
+    /// Defines which language to use for text preprocessing.
+    /// This parameter is used to construct default stopwords filter and stemmer.
+    /// To disable language-specific processing, set this to `none`.
+    /// If not specified, English is assumed.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
JojiiOfficial, line 220: How are we differentiating between "not-specified" and "None" here?
To me it looks like these two cases always result in `None` and we _always_ apply the default value

src/common/inference/bm25.rs
```diff
@@ -147,17 +98,40 @@ impl Bm25 {
...
+        lowercase,
+        stopwords,
+        stemmer,
+        min_token_len,
+        max_token_len,
+    } = value;
+
+    let lowercase = lowercase.unwrap_or(true);
+
+    let language = language.unwrap_or_else(|| DEFAULT_LANGUAGE.to_string());
```
JojiiOfficial, line 113: Doesn't this always result in the default language being applied? How should users disable stemming and stopwords?

timvisee: It can be specified through the options parameter, can't it?

JojiiOfficial: Sorry for the confusion. I mean it can't be disabled, it'll always be the users value or the default value.
This results in stemmer and stopwords always being enabled, either for english (`DEFAULT_LANGUAGE`) or the users parameter. I think this is not intended as users might want to disable stemming and stopwords.

The docs for this `language` parameter say that we can disable it entirely. This is why I think this code might need to be improved.

lib/api/src/rest/schema.rs
timvisee: I agree this isn't great. But we do have struct-destructing here as safe guard. If we add a new field, the compiler will tell us.

We could use [`serde_with`](https://crates.io/crates/serde_with) here and use:

```rust
#[serde_as(as = "Map<String, Value>")]
```

JojiiOfficial: Good point, I missed the destructing. Still I think this can be removed entirely as we never call this function as far as I can see. (Please correct me if I'm wrong)

lib/api/src/rest/schema.rs
generall: yeah, this is true

lib/api/src/rest/schema.rs
generall: we can't differentiate empty and None.

src/common/inference/bm25.rs
generall: Disabling default language can be done via `"language": "none"` or any other non-existing language. 

I see you want to have this consistency of None = no language, but this is not expected behavior and not how users would expect to use it. English stays default.

lib/api/src/rest/schema.rs
generall: unreachable is a panic. If we do something wrong here, I would prefer it to degrade in production, instead of failing completely.

lib/api/src/rest/schema.rs
JojiiOfficial: Yes, however the documentation is stating this
`/// To disable language-specific processing, set this to none`
Which sounds to me that we can disable stemmers+stopwords by setting this to `None`.
But the code uses "english" if it's `None` and stemmers+stopwords aren't disabled. I'm fine with either of both but I want to make sure that the documentation is correct.

src/common/inference/bm25.rs
JojiiOfficial: Ah, got it. "none" as `&str` and not as `Option::None`. This is where I misunderstood the documentation because it stated `None`. 
I'm fine with default being english, I was just confused by the documentation which I understood as `Option::None` and not "none" to disable this feature.

Let's maybe add a hint or better explanation regarding this in the documentation to prevent similar confusion in future

lib/api/src/rest/schema.rs
generall: documentation is technically correct. In openAPI there is no `none`.  There is `null`, so if you use `"none"` it will work.

generall: I will make it more explicit

JojiiOfficial: Yeah you're absolutely right. I was just focusing on the code and not the API.
Thanks for making it more explicit!

lib/api/src/rest/schema.rs
timvisee: Solved by https://github.com/qdrant/qdrant/pull/7056/commits/5aec908423e921a88550c88f594628d789569c54

lib/api/src/rest/schema.rs
timvisee: How about this? 

```rust
// Order of keys in the map should not affect the hash
let mut keys: Vec<(_, _)> = options.iter().collect();
keys.sort_by_key(|(key, _)| *key);
keys.hash(state);
```

I'm not too worried about it because it does not have to be portable.

timvisee: I'll merge this now so I can roll the release. We can handle this one separately if we like.

timvisee (approved):

PR #7063, opened 2025-08-14
Introduce replicate points action

KShivendu:
Interface for copying points between two replicas. Part of scalable multi-tenancy design.

It can be used like this:

You need to create cutom shard keys first 
```http
PUT /collections/tenant
{
  "sharding_method": "custom"
}

PUT /collections/tenant/shards
{
  "shard_key": "default",
  "shards_number": 1
}

PUT /collections/tenant/shards
{
  "shard_key": "tenant",
  "shards_number": 1
}
```

```http
POST /collections/benchmark/cluster
{
  "replicate_points": {
    "filter": {"should": {"key": "tenant_id", "match": {"value": "tenant"}}}, // filters will be rejected for now
    "from_shard_key": "default",
    "to_shard_key": "tenant",
  }
}
```

Note that this will only transfer points that will match both the hashring and the filter

ToDo:
- [x] Think if we should restrict the filters. It bloats the consensus entries in memory:
```console
2025-10-07T14:43:07.070522Z DEBUG storage::content_manager::toc::collection_meta_ops: Transfer shard Start(ShardTransfer { shard_id: 1, to_shard_id: Some(2), from: 3959645297537920, to: 3959645297537920, sync: true, method: Some(FilteredStreamRecords), filter: Some(Filter { should: Some([Field(FieldCondition { key: JsonPath { first_key: "tenant_id", rest: [] }, match: Some(Value(MatchValue { value: String("orgX") })), range: None, geo_bounding_box: None, geo_radius: None, geo_polygon: None, values_count: None, is_empty: None, is_null: None })]), min_should: None, must: None, must_not: None }) }) of tenant
```
No we don't need to restrict filters or anything. Andrey suggested to just improve logging so it prints serialized values instead.
- [ ] Investigate why [this](https://github.com/qdrant/qdrant/pull/7063#discussion_r2432639021) didn't work

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -354,38 +354,6 @@ impl Collection {
             .collect()
     }
 
-    pub async fn get_shard_ids(&self, shard_key: &ShardKey) -> CollectionResult<Vec<ShardId>> {
-        self.shards_holder
-            .read()
-            .await
-            .get_shard_key_to_ids_mapping()
-            .get(shard_key)
-            .map(|ids| ids.iter().cloned().collect())
-            .ok_or_else(|| {
-                CollectionError::bad_input(format!(
-                    "Shard key {shard_key} does not exist for collection {}",
-                    self.name()
-                ))
-            })
-    }
-
-    pub async fn get_replicas(
-        &self,
-        shard_key: &ShardKey,
-    ) -> CollectionResult<Vec<(ShardId, PeerId)>> {
-        let shard_ids = self.get_shard_ids(shard_key).await?;
-        let shard_holder = self.shards_holder.read().await;
-        let mut replicas = Vec::new();
-        for shard_id in shard_ids {
-            if let Some(replica_set) = shard_holder.get_shard(shard_id) {
-                for (peer_id, _) in replica_set.peers() {
-                    replicas.push((shard_id, peer_id));
-                }
-            }
-        }
-        Ok(replicas)
-    }
-
     /// Return a list of local shards, present on this peer
     pub async fn get_local_shards(&self) -> Vec<ShardId> {
         self.shards_holder.read().await.get_local_shards().await
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -3547,8 +3547,6 @@ pub struct Filter {
     pub must_not: Option<Vec<Condition>>,
 }
 
-impl Eq for Filter {}
-
 impl Filter {
     pub fn new() -> Self {
         Filter {
--- a/src/common/collections.rs
+++ b/src/common/collections.rs
@@ -331,19 +331,23 @@ pub async fn do_update_collection_cluster(
                 to_shard_key,
             } = replicate_points;
 
+            let from_shard_ids = collection.get_shard_ids(&from_shard_key).await?;
+
+            // Temporary, before we support multi-source transfers
+            if from_shard_ids.len() != 1 {
+                return Err(StorageError::BadRequest {
+                    description: format!(
+                        "Only replicating from shard keys with exactly one shard is supported. Shard key {from_shard_key} has {} shards",
+                        from_shard_ids.len()
+                    ),
+                });
+            }
+
             // validate shard key exists
             let from_replicas = collection.get_replicas(&from_shard_key).await?;
             let to_replicas = collection.get_replicas(&to_shard_key).await?;
 
-            // Allow replication only shard keys with exactly one source (for now) and destination replica
-            if from_replicas.len() != 1 {
-                return Err(StorageError::BadRequest {
-                    description: format!(
-                        "Only replicating from shard keys with exactly one replica is supported. Shard key {from_shard_key} has {} replicas",
-                        from_replicas.len()
-                    ),
-                });
-            }
+            debug_assert!(!from_replicas.is_empty());
 
             if to_replicas.len() != 1 {
                 return Err(StorageError::BadRequest {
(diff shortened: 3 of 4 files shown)


lib/collection/src/operations/cluster_ops.rs
```diff
@@ -185,6 +188,35 @@ pub struct FinishReshardingOperation {
...
+}
+
+impl Validate for ReplicatePoints {
+    fn validate(&self) -> Result<(), ValidationErrors> {
+        // validate_shard_different_peers(
+        //     self.from_peer_id,
+        //     self.to_peer_id,
+        //     self.from_shard_id,
+        //     Some(self.to_shard_id),
+        // )
```
coszio: todo?

lib/segment/src/types.rs
```diff
@@ -3555,6 +3555,8 @@ pub struct Filter {
     pub must_not: Option<Vec<Condition>>,
 }
 
+impl Eq for Filter {}
+
```
coszio: This one can also be a derive macro :)

lib/collection/src/shards/transfer/mod.rs
```diff
@@ -142,6 +147,9 @@ pub enum ShardTransferMethod {
     /// Shard transfer for resharding: stream all records in batches until all points are
     /// transferred.
     ReshardingStreamRecords,
+    /// Shard transfer for : stream all records in batches until all points satisfying the filter are
+    /// transferred.
+    FilteredStreamRecords,
```
coszio:
```suggestion
    /// Shard transfer for filtered resharding: stream all records in batches until all points satisfying the filter are
    /// transferred.
    FilteredStreamRecords,
```

lib/collection/src/shards/transfer/mod.rs
```diff
@@ -45,6 +46,10 @@ pub struct ShardTransfer {
     /// Method to transfer shard with. `None` to choose automatically.
     #[serde(default)]
     pub method: Option<ShardTransferMethod>,
+
+    // Optional filter to apply when transferring points
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub filter: Option<Filter>,
```
coszio, line 52: If this structure is decoupled from grpc api, would it make more sense to add the filter directly to the `ShardTransferMethod::FilteredStreamRecords` variant?

lib/api/src/grpc/proto/collections.proto
```diff
@@ -727,6 +727,7 @@ enum ShardTransferMethod {
   Snapshot = 1; // Snapshot the shard and recover it on the target peer
   WalDelta = 2; // Resolve WAL delta between peers and transfer the difference
   ReshardingStreamRecords = 3; // Stream shard records in batches for resharding
+  FilteredStreamRecords = 4; // Stream shard records that match a filter in batches
```
coszio: I'm not seeing how we communicate the `filter` in grpc 🤔

KShivendu: Not doing it right now. Will add once it's stable. Want to keep PRs small.

lib/collection/src/shards/transfer/mod.rs
KShivendu: used a different variant. but thanks for pointing out! :D

lib/collection/src/operations/cluster_ops.rs
KShivendu: dropped. can't think of a value level validation in this. I'm checking the shard key exists and stuff later in the code. thanks!

lib/api/src/grpc/proto/collections.proto
KShivendu: Noted in [milestone](https://github.com/qdrant/qdrant/milestone/37). Thanks!

lib/segment/src/types.rs
KShivendu: It doesn't work if I do that. Some of its attributes somehow let me compile. Will figure out the reason :thinking:

lib/collection/src/shards/transfer/mod.rs
KShivendu: Maybe not. We render the transfer method in `GET /collections/benchmark/cluster` which will look confusing why method has a filter inside.

lib/collection/src/collection/mod.rs
```diff
@@ -354,6 +354,38 @@ impl Collection {
             .collect()
     }
 
+    pub async fn get_shard_ids(&self, shard_key: &ShardKey) -> CollectionResult<Vec<ShardId>> {
```
generall: please move into `sharding_keys.rs`

lib/collection/src/collection/mod.rs
```diff
@@ -354,6 +354,38 @@ impl Collection {
...
+            .map(|ids| ids.iter().cloned().collect())
+            .ok_or_else(|| {
+                CollectionError::bad_input(format!(
+                    "Shard key {shard_key} does not exist for collection {}",
+                    self.name()
+                ))
+            })
+    }
+
+    pub async fn get_replicas(
```
generall: please move into `sharding_keys.rs`

lib/segment/src/types.rs
generall: forcing compiler to accept those traits is a footgun waiting to cause problems.

I did a boring part to implement Eq properly https://github.com/qdrant/qdrant/pull/7419

src/common/collections.rs
```diff
@@ -312,6 +316,73 @@ pub async fn do_update_collection_cluster(
...
+                from_shard_key,
+                to_shard_key,
+            } = replicate_points;
+
+            // validate shard key exists
+            let from_replicas = collection.get_replicas(&from_shard_key).await?;
+            let to_replicas = collection.get_replicas(&to_shard_key).await?;
+
+            // Allow replication only shard keys with exactly one source (for now) and destination replica
+            if from_replicas.len() != 1 {
```
generall: if there are more than one replica, can't we just pick random one?

PR #7067, opened 2025-08-15
Introduce shard key fallback that can be used to route read/write requests

KShivendu:
If `shard_key` specified by an upsert doesn't exist, points will be upserted into a `fallback` shard key (if that exists). Otherwise, we throw error. Same will happen with read requests (scroll, query, count, etc)

So if you create a fallback custom shard like this:
```http
PUT /collections/benchmark/shards
{
  "shard_key": "fallback"
}

GET /collections/benchmark/cluster
# Response:
{
  "result": {
    "peer_id": 5190255624632758,
    "shard_count": 1,
    "local_shards": [
      {
        "shard_id": 1,
        "shard_key": "fallback",
        "points_count": 0,
        "state": "Active"
      }
    ],
    "remote_shards": [],
    "shard_transfers": []
  },
  "status": "ok",
  "time": 0.000217938
}
```

Example upsert
```http
PUT /collections/benchmark/points
{
    "points": [
        {
            "id": 1,
            "vector": [0.1, 0.2, 0.3]
        }
    ],
    "shard_key": "org1"
}
```

Now if you do a scroll (read) request

```http
POST /collections/benchmark/points/scroll
{
  "shard_key": "org1"
}

# Response:
{
  "result": {
    "points": [
      {
        "id": 1,
        "shard_key": "fallback"
      }
    ],
    "next_page_offset": null
  },
  "status": "ok",
  "time": 0.001306022
}
``` 

Same happens with query (and other read requests)

```http
{
  "query": [0.1, 0.2, 0.3],
  "shard_key": "org1"
}

## Response:

{
  "result": {
    "points": [
      {
        "id": 1,
        "version": 0,
        "score": 0.9999998,
        "shard_key": "fallback"
      }
    ]
  },
  "status": "ok",
  "time": 0.002821458
}
```

### DIFF
--- a/lib/collection/src/shards/shard_holder/mod.rs
+++ b/lib/collection/src/shards/shard_holder/mod.rs
@@ -63,6 +63,7 @@ pub struct ShardHolder {
     key_mapping: SaveOnDisk<ShardKeyMapping>,
     // Duplicates the information from `key_mapping` for faster access, does not use locking
     shard_id_to_key_mapping: HashMap<ShardId, ShardKey>,
+    fallback_shard_key: Option<ShardKey>,
 }
 
 pub type LockedShardHolder = RwLock<ShardHolder>;
@@ -103,6 +104,7 @@ impl ShardHolder {
             rings,
             key_mapping,
             shard_id_to_key_mapping,
+            fallback_shard_key: Some(ShardKey::Keyword("fallback".into())), // ToDo: Take from collection params
         })
     }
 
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -103,6 +104,7 @@ impl ShardHolder {
             rings,
             key_mapping,
             shard_id_to_key_mapping,
+            fallback_shard_key: Some(ShardKey::Keyword("fallback".into())), // ToDo: Take from collection params
```
generall: Does this mean that fallback shard is activated by default?

If so, let's remove it so it won't be a breaking change

generall (approved):

lib/collection/src/shards/shard_holder/mod.rs
KShivendu: Yes. It should be None by default. 

I'm creating PR for that part. Wanted to keep this one small.

timvisee:
I think we should bump this one to switch to the new approach we chose.

It may be best to open a fresh PR for it to not clutter the PR history, while using some of the changes from here.

Please re-tag me if we'll keep this one :smiley:

PR #7069, opened 2025-08-15
Fallback shard key config

KShivendu:
Allow configuring `fallback_shard_key` introduced in https://github.com/qdrant/qdrant/pull/7067 as collection level config

```http
# While creating the collection:
PUT /collections/benchmark
{
   "fallback_shard_key": "fallback"
}

# If collection already exists:
PUT /collections/benchmark
{
   "fallback_shard_key": "fallback"
}
```


lib/collection/src/operations/config_diff.rs
```diff
@@ -103,10 +104,42 @@ pub struct WalConfigDiff {
...
+    pub key: ShardKey,
+}
+
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, PartialEq, Eq, Hash)]
+#[serde(untagged)]
+#[serde(rename_all = "snake_case")]
+pub enum FallbackShardKeyDiff {
+    Key(FallbackShardKey),
+    Disabled(Disabled),
+}
```
KShivendu, line 119: Need some help with this. It works fine when I write a unit test to parse from raw json string. But the same thing doesn't work when I call it from dashboard (i.e. integrated in the API)


```http
# This doesn't work:
PATCH /collections/test
{
  "params": {
    "fallback_shard_key": {
      "key": "ggg"
    }
  }
}

# Response:
{
  "error": "Service internal error: Json error: data did not match any variant of untagged enum ShardKey"
}

# Howver, this works:
PATCH /collections/test
{
  "params": {
    "fallback_shard_key": "Disabled"
  }
}
```

ffuugoo: You need to change `CollectionParams::fallback_shard_key` type from `ShardKey` to `FallbackShardKey`.

KShivendu: Why? `CollectionParamsDiff` can be different no? I want final struct to remain simple shard key (not `{key: xxx}`) and i'm converting diff into normal config [here](https://github.com/qdrant/qdrant/pull/7069/files#diff-af0f39f882a19c9785ef60863b1de8e50b8d4a933637c99eb11fc5e42a73cfaeR37-R48)

ffuugoo: Good point, but [`CollectionParams` is updated through `serde`][update-config]. E.g., `CollectionParams` should be directly deserializable from `CollectionParamsDiff` JSON.

[update-config]: https://github.com/qdrant/qdrant/blob/a7d21f168097494acc2cbfb2489cfd2da40581ac/lib/collection/src/operations/config_diff.rs#L326-L335

ffuugoo: I guess what might work is:

  ```rust
  struct CollectionParams {
      #[serde(deserialize_with = "deserialize_fallback_shard_key", serialize_with = "serialize_fallback_shard_key")
      fallback_shard_key: Option<ShardKey>,
  }

  struct CollectionParamsDiff {
	  // I'm not sure if you can reuse `deserialize_fallback_shard_key` directly,
	  // or you'll have to make ANOTHER `deserialize_*_opt` function, that returns `Option<Option<ShardKey>>`
	  //
	  // probably the latter
      #[serde(default, deserialize_with = "deserialize_fallback_shard_key_opt", serialize_with = "..."]
      fallback_shard_key: Option<Option<ShardKey>>,
  }

  fn deserialize_fallback_shard_key<'de, D>(deserializer: D) -> Result<Option<ShardKey>, D::Error>
  where
      D: serde::Deserializer<'de>,
  {
      enum FallbackShardKey { /* copy current `FallbackShardKey` here */ }
      let fallback_key = FallbackShardKey::deserialize(deserializer)?;

      let key = match key {
          FallbackShardKey::Key(key) => Some(key),
          FallbackShardKey::Disabled(_) => None,
      };
      
      Ok(key)
  }
  ```

KShivendu: > Good point, but [CollectionParams is updated through serde](https://github.com/qdrant/qdrant/blob/a7d21f168097494acc2cbfb2489cfd2da40581ac/lib/collection/src/operations/config_diff.rs#L326-L335). 

Good point. I was doing [this](https://github.com/qdrant/qdrant/pull/7069/files#diff-af0f39f882a19c9785ef60863b1de8e50b8d4a933637c99eb11fc5e42a73cfaeR39-R49) after that step but looks like this `key` will merged into the JSON and will cause problems. Better do this explictly without the merge function. Will update!

KShivendu: > fallback_shard_key: Option<Option<ShardKey>>,

Andrey suggested using the `fallback_shard_key: "Disabled"` interface instead of `fallback_shard_key: null` for consistency.

KShivendu: I guess I'll just do a custom serializer. But weird why serde isn't working as expected. Will try using debugger to investigate once.

ffuugoo: > Andrey suggested using the fallback_shard_key: "Disabled" interface instead of fallback_shard_key: null for consistency.

I'm not suggesting to use `null`. Serde will still use `"disabled" and { "key": "whatever" }`. It's just Rust types used in `Params`/`ParamsDiff` will be clean `Option` and `Option<Option>` instead of `FallbackShardKey` (which is ugly).

> I guess I'll just do a custom serializer. But weird why serde isn't working as expected. Will try using debugger to investigate once.

It's not serde issue. The issue is that 1) we use serde to implement `update_config` and 2) `Params` use totally different type that uses totally different serde format from `ParamsDiff`. Serde is working exactly as it should: `fallback_shard_key: Option<ShardKey>` says "it can be either `null`/not specified or `ShardKey` (which is a string)". And when doing `update_config` with `ParamsDiff` you give it `{ "key": "something" }`.

timvisee:
Closing this since we'll use a different approach based on our recent design decisions.

More specifically, we don't configure shards in a special way and use our existing custom sharding.

Instead, we'll specify a target/default shard at the request level (for both reads and writes).

PR #7073, opened 2025-08-15
do not skip empty histogram values + remove too small buckets

generall:


lib/segment/src/common/operation_time_statistics.rs
```diff
@@ -436,26 +415,34 @@ mod tests {
...
+            vec![(0., 0), (1., 0), (2., 0), (3., 0), (4., 0), (5., 0)],
         );
 
         // With all zeroes except the total count
         assert_eq!(
-            convert_histogram(&[0., 1., 2., 3., 4., 5.], &[0, 0, 0, 0, 0, 0], 100),
-            vec![(5., 0)],
+            convert_histogram(&[0., 1., 2., 3., 4., 5.], &[0, 0, 0, 0, 0, 0]),
+            vec![(0., 0), (1., 0), (2., 0), (3., 0), (4., 0), (5., 0)],
         );
```
KShivendu: There's no difference between this test case and the previous one. Please drop this one.

PR #7094, opened 2025-08-19
Remove `async` from `SegmentHolder::deduplicate_points`

ffuugoo:
Small refactoring for Qdrant on Edge. We need to move `SegmentHolder` into `shard` crate, but we don't want `async` in `shard`, and `deduplicate_points` is the only `async` method of `SegmentHolder`.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection_manager/holders/segment_holder.rs
+++ b/lib/collection/src/collection_manager/holders/segment_holder.rs
@@ -11,7 +11,6 @@ use ahash::{AHashMap, AHashSet};
 use common::counter::hardware_counter::HardwareCounterCell;
 use common::iterator_ext::IteratorExt;
 use common::tar_ext;
-use futures::future::try_join_all;
 use io::storage_version::StorageVersion;
 use parking_lot::{RwLock, RwLockReadGuard, RwLockUpgradableReadGuard, RwLockWriteGuard};
 use rand::seq::IndexedRandom;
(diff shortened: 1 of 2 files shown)


lib/collection/src/collection_manager/holders/segment_holder.rs
```diff
@@ -1321,15 +1320,15 @@ impl SegmentHolder {
     /// Checks all segments and removes duplicated and outdated points.
     /// If two points have the same id, the point with the highest version is kept.
     /// If two points have the same id and version, one of them is kept.
-    pub async fn deduplicate_points(&self) -> OperationResult<usize> {
+    pub fn deduplicate_points_task(&self) -> Vec<impl Fn() -> OperationResult<usize> + 'static> {
```
timvisee: Nit: does it make sense for us to combine everything into a single function, so that we have just one thing to call?

timvisee: Nope I see we still want the individual tasks in the async function.

But then I vote to call this `deduplicate_points_tasks`

ffuugoo: Fixed.

ffuugoo:
Rebased on top of storage size test fix (thanks @agourlay!), will merge when green.

PR #7095, opened 2025-08-19
Retrieve files list from quantization storage

IvanPleshkov:
This PR refactors `QuantizedVectors::files` function. In `dev`, `QuantizedVectors::files` returns hardcoded filenames and does not use actual info from quantization storage. It was a part of an old design where quantization had a save/load function and `QuantizedVectors` owned files list.

Now it's incorrect, and `QuantizedVectors` files ownership does not allow the use `ChunkedVectorsMmap` as quantization storage because `ChunkedVectorsMmap` generates files by itself.

This PR provides `files` function down to quantization storage implementation and quantization storages know its file if it was saved to disk.

### DIFF
--- a/lib/quantization/src/encoded_vectors_pq.rs
+++ b/lib/quantization/src/encoded_vectors_pq.rs
@@ -5,7 +5,7 @@ use std::arch::aarch64::*;
 use std::arch::x86_64::*;
 use std::iter::repeat_with;
 use std::ops::Range;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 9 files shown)


lib/quantization/src/encoded_vectors_pq.rs
```diff
@@ -29,6 +29,7 @@ pub const CENTROIDS_COUNT: usize = 256;
 pub struct EncodedVectorsPQ<TStorage: EncodedStorage> {
     encoded_vectors: TStorage,
     metadata: Metadata,
+    metadata_path: Option<PathBuf>,
```
timvisee, line 32: Probably not, but, does it make sense to move `metadata_path` into `metadata` too?

IvanPleshkov: metadata is serializable, we will get in json the path of this json

PR #7109, opened 2025-08-20
In proxy segment, don't return deleted point versions

timvisee:
More specifically, don't return a point version from the wrapped segment if that point is already marked for deletion within the proxy.

In some places we rely on `point_version` to tell us if (and what) point (version) exists in a segment. When a proxy segment marks a wrapped point for deletion, we should not expose its point version anymore.

The old behavior broke the following places (and potentially others):
- https://github.com/qdrant/qdrant/blob/5ffe70d3217790cf09a1f2406803de4bdbfa6931/lib/collection/src/collection_manager/holders/segment_holder.rs#L395
- https://github.com/qdrant/qdrant/blob/5ffe70d3217790cf09a1f2406803de4bdbfa6931/lib/collection/src/collection_manager/holders/segment_holder.rs#L564

I'm currently still evaluating if stuff around this code is also sound.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/shard/src/proxy_segment/segment_entry.rs
+++ b/lib/shard/src/proxy_segment/segment_entry.rs
@@ -36,7 +36,17 @@ impl SegmentEntry for ProxySegment {
             .get()
             .read()
             .point_version(point_id)
-            .or_else(|| self.wrapped_segment.get().read().point_version(point_id))
+            .or_else(|| {
+                // Ignore point from wrapped segment if already marked for deletion
+                // By `point_version` semantics we don't expect to get a version if the point
+                // is deleted. This also prevents `move_if_exists` from moving an old point
+                // into the write segment again.
+                if self.deleted_points.read().contains_key(&point_id) {
+                    return None;
+                }
+
+                self.wrapped_segment.get().read().point_version(point_id)
+            })
     }
 
     fn search_batch(
--- a/lib/shard/src/proxy_segment/segment_entry.rs
+++ b/lib/shard/src/proxy_segment/segment_entry.rs
@@ -31,27 +31,28 @@ impl SegmentEntry for ProxySegment {
     }
 
     fn point_version(&self, point_id: PointIdType) -> Option<SeqNumberType> {
-        // Write version is always higher if present
-        self.write_segment
-            .get()
+        // Use write segment version if present, we assume it's always higher
+        if let Some(version) = self.write_segment.get().read().point_version(point_id) {
+            return Some(version);
+        }
+
+        // Use wrapped segment version, if absent we have no version at all
+        let wrapped_version = self.wrapped_segment.get().read().point_version(point_id)?;
+
+        // Ignore point from wrapped segment if already marked for deletion with newer version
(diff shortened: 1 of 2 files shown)


lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -36,7 +36,17 @@ impl SegmentEntry for ProxySegment {
...
             .read()
             .point_version(point_id)
-            .or_else(|| self.wrapped_segment.get().read().point_version(point_id))
+            .or_else(|| {
+                // Ignore point from wrapped segment if already marked for deletion
+                // By `point_version` semantics we don't expect to get a version if the point
+                // is deleted. This also prevents `move_if_exists` from moving an old point
+                // into the write segment again.
+                if self.deleted_points.read().contains_key(&point_id) {
+                    return None;
```
agourlay: Don't we need to check that the deleted_version is > than the version found in the wrapped segment?

timvisee: Nice catch, sounds good!

I don't think it'll make a difference in practice. But I'd argue your idea is more correct, so let's apply it.

agourlay: My suggestion breaks two tests which is weird. Need to dig deeper.

```
   FAIL [   0.100s] collection collection_manager::tests::test_proxy_shared_updates
   FAIL [   0.088s] collection collection_manager::tests::test_proxy_shared_updates_same_version
```

agourlay: BTW my local change is

```rust
let wrapped_version = self.wrapped_segment.get().read().point_version(point_id);
let deleted_version = self.deleted_points.read().get(&point_id).cloned();
wrapped_version.zip(deleted_version).and_then(|(wrapped_version, deleted_version)| {
    if deleted_version.local_version < wrapped_version {
        Some(wrapped_version)
    } else {
        None
    }
})
```

timvisee: I don't think that's correct. The `deleted_version` will be `None` if it wasn't marked for deletion yet. Then `zip` will iterate nothing.

I think the full function should be like this:

```rust
  fn point_version(&self, point_id: PointIdType) -> Option<SeqNumberType> {
      // Write version is always higher if present
      self.write_segment
          .get()
          .read()
          .point_version(point_id)
          .or_else(|| {
              self.wrapped_segment
                  .get()
                  .read()
                  .point_version(point_id)
                  // Ignore point from wrapped segment if already marked for deletion
                  // By `point_version` semantics we don't expect to get a version if the point
                  // is deleted. This also prevents `move_if_exists` from moving an old point
                  // into the write segment again.
                  .filter(|version| {
                      self.deleted_points
                          .read()
                          .get(&point_id)
                          .is_none_or(|delete| version > &delete.local_version)
                  })
          })
  }
```

It still satisfies those tests.

Committed it to this branch now. Thoughts?

PR #7118, opened 2025-08-22
Graph links serializer: avoid large intermediate allocations

xzfc:
# Problem

The current [`GraphLinksSerializer`](https://github.com/qdrant/qdrant/blob/v1.15.3/lib/segment/src/index/hnsw_index/graph_links/serializer.rs#L51) flow consists of two steps:

1. The `new()` method serializes each section (`reindex`, `neighbors`, `offsets`, etc.) into its own `Vec`.
2. Then `serialize_to_writer()` writes these sections in sequence.

The drawback of this flow is that the `neighbors` section for the `CompressedWithVectors` format can consume a lot of RAM. During testing, I had the process killed by the OOM killer during graph conversion.

# Solution

This PR replaces the `GraphLinksSerializer` struct and its methods with a new function, `serialize_graph_links`, which serializes to the writer directly without an intermediate in-memory representation.

# Details

* Previously, `GraphLinksSerializer::to_graph_links_ram` did [proper total size calculation] for `Vec::with_capacity()`. In this PR, this logic is dropped (otherwise it would require two compression passes: the first time to get length, the second time to write bytes). This is compensated by calling `.shrink_to_fit()` afterwards, and while this is still a bit sloppy, I expect that the overall amount of reallocations/peak memory usage is not worse than before.
* The `GraphLinksSerializer::save_as()` logic is moved to a new general-purpose function, `io::file_operations::atomic_save()`, which can be combined with `serialize_graph_links()`.
* The `GraphLinksSerializer::to_graph_links_ram()` logic is replaced with `GraphLinks::new_from_edges()`.

[proper total size calculation]: https://github.com/qdrant/qdrant/blob/v1.15.3/lib/segment/src/index/hnsw_index/graph_links/serializer.rs#L219-L246


lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -191,14 +192,32 @@ impl GraphLinks {
...
         })
     }
 
     fn view(&self) -> &GraphLinksView<'_> {
         self.borrow_dependent()
     }
 
+    pub fn as_bytes(&self) -> &[u8] {
+        self.borrow_owner().as_bytes()
+    }
```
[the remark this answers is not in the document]
xzfc: It's not related to this PR. Please remove this method in a separate PR.

lib/segment/src/index/hnsw_index/graph_layers_builder.rs
```diff
@@ -189,15 +190,17 @@ impl GraphLayersBuilder {
...
 
-        let serializer =
-            Self::links_layers_to_serializer(self.links_layers, format_param, self.hnsw_m)?;
-        serializer.save_as(&links_path)?;
-
-        let links = if on_disk {
-            GraphLinks::load_from_file(&links_path, true, format_param.as_format())?
+        let edges = Self::links_layers_to_edges(self.links_layers);
+        let links;
+        if on_disk {
```
generall, line 195: it feels like `on_disk` param requires a comment.

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -1,356 +1,214 @@
...
-                } => {
-                    size_of::<HeaderCompressedWithVectors>()
-                        + neighbors_padding
-                        + compressed_offsets.len()
+        iter.try_for_each(|id| {
+            let mut raw_links = std::mem::take(&mut edges[id][level]);
+            match format_param {
+                GraphLinksFormatParam::Plain => {
+                    writer.write_all(raw_links.as_bytes())?;
+                    offset += raw_links.len();
```
IvanPleshkov, line 107: I like simplification here, prev implementation was much harder:
```
offsets.push((neighbors.len() as u64) / size_of::<PointOffsetType>() as u64);
```

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -1,356 +1,214 @@
...
+                    writer.write_all(&links_buf)?;
+                    offset += links_buf.len();
+                }
+                GraphLinksFormatParam::CompressedWithVectors(vectors) => {
+                    // Unwrap safety: `vector_layout` is `Some` for `CompressedWithVectors`.
+                    let vector_layout = vector_layout.unwrap();
+
+                    // 1. The varint-encoded length (`N` in the doc).
+                    writer.write_varint(raw_links.len() as u64)?;
+                    offset += u64::required_space(raw_links.len() as u64);
```
IvanPleshkov: I don't see this logic
```
offset += u64::required_space(raw_links.len() as u64);
```
In dev, can you please explain why you use `required_space` here instead of direct `raw_links.len()`? I expect you collect all writed bytes count into writer.

lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -191,14 +192,32 @@ impl GraphLinks {
...
+
+    pub fn new_from_edges(
+        edges: Vec<Vec<Vec<PointOffsetType>>>,
+        format_param: GraphLinksFormatParam<'_>,
+        hnsw_m: HnswM,
+    ) -> OperationResult<Self> {
+        let mut cursor = Cursor::new(Vec::<u8>::new());
+        serialize_graph_links(edges, format_param, hnsw_m, &mut cursor)?;
+        let mut bytes = cursor.into_inner();
+        bytes.shrink_to_fit();
```
IvanPleshkov, line 207: You mentioned that it's a little bit sloppy. But I find it okay and have a suggestion of how to avoid this. 
Actually, we don't need HNSW links after construction. We call `drop` for indexed segment in segment builder, move the segment data directory, and then reload segment to use it. We can just save links and keep RAM links empty - if we can design it such solution properly

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
xzfc: It's a varint thing:

```rust
// assuming raw_links.len() is not a big number and fits within a single byte
writer.write_varint(raw_links.len() as u64)?; // writes a single byte
offset += u64::required_space(raw_links.len() as u64); // u64::required_space() == 1;
                                                       // i.e., `offset += 1`
```

In `dev`, I use `neighbors.len()` as a counter. In this PR I can't use this, so I track written bytes explicitly using `offset` variable:
```rust
offsets.push(neighbors.len() as u64); // in `dev`
offsets.push(offset as u64);          // this PR
```

---

I've replaced `u64::required_space` with `VarInt::required_space` to make it less confusing.

IvanPleshkov: Thanks for the clarification

lib/segment/src/index/hnsw_index/graph_links.rs
xzfc: Good catch. I think we don't need creating `GraphLinksEnum::Ram` at all then and we can use `GraphLinksEnum::Mmap` everywhere except tests. Worth investigating, but perhaps in a different PR.

lib/segment/src/index/hnsw_index/graph_layers_builder.rs
xzfc: Added a comment.

Also, @IvanPleshkov noted that [we don't need HNSW links after construction](https://github.com/qdrant/qdrant/pull/7118#discussion_r2293606782). So, probably this part will be gone in this or subsequent PRs.

IvanPleshkov: I propose a separate PR to make it more readable for reviewers and avoid blocking these changes to merge

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -1,356 +1,214 @@
...
-        // `reindex` is map from point id to index in `Self.offsets`
-        let mut reindex = vec![0; back_index.len()];
-        for i in 0..back_index.len() {
-            reindex[back_index[i]] = i as PointOffsetType;
-        }
+    let levels_count = back_index.first().map_or(0, |&id| edges[id].len());
+    let mut point_count_by_level = vec![0; levels_count];
+    for point in &edges {
+        point_count_by_level[point.len() - 1] += 1;
+    }
```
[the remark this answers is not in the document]
xzfc: We don't build graphs with points with no levels. And when we do for some reason, the unit tests will catch it early. So, I'd prefer to keep it simple for now.

generall (approved):

PR #7123, opened 2025-08-24
implement custom collection metadata

generall:
```
PATCH /collections/star_charts
{
  "metadata": {
    "any_values": "go-like-this"
  }
}
```

https://github.com/qdrant/qdrant/issues/3957

### DIFF
--- a/tests/openapi/test_collection_metadata.py
+++ b/tests/openapi/test_collection_metadata.py
@@ -39,6 +39,14 @@ def test_collection_exists(collection_name):
     assert response.ok
 
     response = request_with_validation(
+        api="/collections/{collection_name}",
+        method="PATCH",
+        path_params={"collection_name": collection_name},
+        body={"metadata": {"new_meta_2": None}}
+    )
+    assert response.ok
+
+    response = request_with_validation(
         api='/collections/{collection_name}',
         method="GET",
         path_params={'collection_name': collection_name},
@@ -46,4 +54,4 @@ def test_collection_exists(collection_name):
     assert response.ok
     result = response.json()["result"]
 
-    assert result["config"]["metadata"] == {"new_meta": "value3", "new_meta_2": "value2"}
+    assert result["config"]["metadata"] == {"new_meta": "value3"}


lib/collection/src/collection/collection_ops.rs
```diff
@@ -163,6 +163,20 @@ impl Collection {
...
+            crate::config::CollectionConfigInternal,
+        > = self.collection_config.write().await;
+
+        if let Some(current_metadata) = collection_config_guard.metadata.as_mut() {
+            current_metadata.merge(&metadata);
+        } else {
+            collection_config_guard.metadata = Some(metadata);
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
generall: :+1:

lib/storage/src/content_manager/conversions.rs
```diff
@@ -191,6 +199,11 @@ impl TryFrom<api::grpc::qdrant::UpdateCollection> for CollectionMetaOperations {
                     .map(SparseVectorsConfig::try_from)
                     .transpose()?,
                 strict_mode_config: strict_mode_config.map(StrictModeConfig::from),
+                metadata: if metadata.is_empty() {
+                    None
+                } else {
+                    Some(json::proto_to_payloads(metadata)?)
+                },
```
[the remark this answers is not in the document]
generall: > This prevents a PATCH of { "metadata": {} } from ever clearing existing metadata.

That is correct, however it it is possible to clean specific metadata key by doing

```
{ "metadata": {"key": null} }
```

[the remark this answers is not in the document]
generall: we can't have ` { "metadata": {} } ` to clear whole object, as there are no difference between empty map and null in gPRC, so API would be inconsistent

tests/openapi/test_collection_metadata.py
```diff
@@ -0,0 +1,49 @@
...
+
+    response = request_with_validation(
+        api='/collections/{collection_name}',
+        method="GET",
+        path_params={'collection_name': collection_name},
+    )
+    assert response.ok
+    result = response.json()["result"]
+
+    assert result["config"]["metadata"] == {"new_meta": "value3", "new_meta_2": "value2"}
```
timvisee: As far as I can see, there is no way to unset keys from the metadata.

How about making the patch request replace the whole metadata object, not merging keys? That would allow unsetting keys.

I understand that isn't exactly great either, because it still doesn't allow us to unset the whole object. But I'd argue it's a bit better.

generall: there is `{"metadata": {"key": null}}`

timvisee: That doesn't work through our Python client with `None`. That's what I tried in the included test :smiley:

generall: > That doesn't work through our Python client with None. That's what I tried in the included test 😃

why not?

timvisee: :man_shrugging: Looks like it works now. I've added it to the test.

timvisee (approved):

PR #7126, opened 2025-08-25
Fix flaky save_on_disk test race condition

fabracht:
Replace sleep-based synchronization with Barrier synchronization in save_on_disk tests to eliminate race conditions causing intermittent failures.

Links to #7114

### DIFF
--- a/lib/common/common/src/save_on_disk.rs
+++ b/lib/common/common/src/save_on_disk.rs
@@ -182,14 +182,20 @@ impl<T> DerefMut for SaveOnDisk<T> {
 
 #[cfg(test)]
 mod tests {
-    use std::sync::Arc;
+    use std::sync::Barrier;
     use std::thread::sleep;
-    use std::time::Duration;
+    use std::time::{Duration, Instant};
     use std::{fs, thread};
 
     use tempfile::Builder;
 
     use super::SaveOnDisk;
+
+    const TEST_IMMEDIATE_TIMEOUT: Duration = Duration::from_millis(500);
+
+    const TEST_SYNC_DELAY: Duration = Duration::from_millis(50);
+
+    const TEST_UPDATE_INTERVAL: Duration = Duration::from_millis(100);
 
     #[test]
     fn saves_data() {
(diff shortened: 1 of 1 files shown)


timvisee:
It looks like a CI test is failing because of your changes:

https://github.com/qdrant/qdrant/actions/runs/17198740860/job/48824454328?pr=7126#step:10:1391

lib/common/common/src/save_on_disk.rs
```diff
@@ -85,21 +85,27 @@ impl<T: Serialize + for<'de> Deserialize<'de> + Clone> SaveOnDisk<T> {
     where
         F: Fn(&T) -> bool,
     {
-        let start = std::time::Instant::now();
-        while start.elapsed() < timeout {
+        let deadline = std::time::Instant::now()
+            .checked_add(timeout)
+            .unwrap_or_else(std::time::Instant::now);
```
timvisee: How about a [saturating add](https://doc.rust-lang.org/std/time/struct.Duration.html#method.saturating_add) here?

lib/common/common/src/save_on_disk.rs
```diff
@@ -225,37 +239,105 @@ mod tests {
...
+                counter.write(|c| *c += 3).unwrap();
+                sleep(TEST_UPDATE_INTERVAL);
+                counter.write(|c| *c += 7).unwrap();
+            });
 
-        assert!(counter.wait_for(|counter| *counter > 5, Duration::from_secs(2)));
-        handle.join().unwrap();
+            barrier.wait();
+            assert!(counter.wait_for(|c| *c > 5, TEST_IMMEDIATE_TIMEOUT));
+        });
```
timvisee, line 255: Is this one only necessary if spawning the thread takes a very long time?

lib/common/common/src/save_on_disk.rs
```diff
@@ -225,37 +239,105 @@ mod tests {
...
+    #[test]
+    fn test_wait_for_immediate_condition() {
+        let dir = Builder::new().prefix("test").tempdir().unwrap();
+        let counter_file = dir.path().join("counter");
+        let counter: SaveOnDisk<u32> = SaveOnDisk::new(counter_file, 10).unwrap();
+
+        let start = Instant::now();
+        assert!(counter.wait_for(|c| *c > 5, TEST_IMMEDIATE_TIMEOUT));
+        // Allow headroom for slow CI while still catching accidental sleeps
+        assert!(start.elapsed() <= TEST_IMMEDIATE_TIMEOUT - TEST_SYNC_DELAY);
```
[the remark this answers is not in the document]
timvisee: Actually, let's do `checked_sub` here and unwrap.

lib/common/common/src/save_on_disk.rs
fabracht: I don't think there's saturating add for instants.

lib/common/common/src/save_on_disk.rs
fabracht: Not sure I understand your comment. Are you asking about the barrier? No the barrier is there to assure both threads start the actual work when everything is setup and ready to go. It doesn't matter how long the thread takes. In fact, that's the issue, we don't know how long it takes to spawn the thread on every possible machine.

lib/common/common/src/save_on_disk.rs
timvisee: You're right.

lib/common/common/src/save_on_disk.rs
timvisee: Yes, so the barrier is purely to synchronize the work in both threads since we're not sure how much time the setup takes. Makes sense!

timvisee (approved):

timvisee:
Thanks for your contribution! :pray:

timvisee:
I'm afraid these changes actually introduced flakiness :smiling_face_with_tear: 

https://github.com/qdrant/qdrant/issues/7156

fabracht:
I didn't even notice this was merged.

PR #7128, opened 2025-08-25
fix dockerhub username in builds

KShivendu:
fix dockerhub username for CI

### DIFF
--- a/.github/actions/branch-build-and-push/action.yaml
+++ b/.github/actions/branch-build-and-push/action.yaml
@@ -49,11 +49,11 @@ runs:
         docker buildx create --use
 
         # Authenticate on registries
-        if [ "${{ inputs.push-to-ghcr }}" = "true" ]; then
-          docker login ghcr.io -u qdrant --password "${{ inputs.ghcr-password }}"
-        fi
         if [ "${{ inputs.push-to-dockerhub }}" = "true" ]; then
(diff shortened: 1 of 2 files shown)


.github/actions/branch-build-and-push/action.yaml
```diff
@@ -53,7 +53,7 @@ runs:
           docker login ghcr.io -u qdrant --password "${{ inputs.ghcr-password }}"
         fi
         if [ "${{ inputs.push-to-dockerhub }}" = "true" ]; then
-          docker login docker.io -u qdrant --password "${{ inputs.dockerhub-password }}"
```
timvisee, line 56: Let's also fix the following warning in this PR:

```
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
```

xzfc: Nit: if we are going full hardening, we'll should use `docker login ... <<< "${...}"` instead of `echo "${...}" | docker login ...`. This way, this value will be avoided in the logs:

```console
$ set -x
+ set -x

$ echo topsecret | md5sum -
+ md5sum -
+ echo topsecret
072692aa1a2d7c2eaece991897f5026d  -

$ md5sum - <<< topsecret
+ md5sum -
072692aa1a2d7c2eaece991897f5026d  -
```

(tho, realistically, these values will be filtered out from the logs by GitHub Actions)

PR #7130, opened 2025-08-25
Make points upsert ordering deterministic

agourlay:
This PR makes the points ordering on upsert deterministic by picking the order from the user.

This improves the debugging and testing experience while taking us one more step towards a more deterministic code base.

### DIFF
--- a/lib/collection/src/collection_manager/segments_updater.rs
+++ b/lib/collection/src/collection_manager/segments_updater.rs
@@ -538,8 +538,15 @@ pub(crate) fn upsert_points<'a, T>(
 where
     T: IntoIterator<Item = &'a PointStructPersisted>,
 {
-    let points_map: AHashMap<PointIdType, _> = points.into_iter().map(|p| (p.id, p)).collect();
-    let ids: Vec<PointIdType> = points_map.keys().copied().collect();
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection_manager/segments_updater.rs
```diff
@@ -538,8 +538,15 @@ pub(crate) fn upsert_points<'a, T>(
...
-    let ids: Vec<PointIdType> = points_map.keys().copied().collect();
+    // Conserve initial order of points
+    let points = points.into_iter();
+    let points_count = points.size_hint().0;
+    let mut ids = Vec::with_capacity(points_count);
+    let mut points_map: AHashMap<PointIdType, _> = AHashMap::with_capacity(points_count);
+    for point in points {
+        ids.push(point.id);
+        points_map.insert(point.id, point);
+    }
```
timvisee: We can use some iterator magic here:

```rust
// Conserve initial order of points
let (ids, points_map) = points
    .into_iter()
    .map(|p| (p.id, (p.id, p)))
    .collect::<(Vec<_>, AHashMap<_, _>)>();
```

It also does just one allocation for both based on the size hint.

PR #7004, opened 2025-08-08
Open and load full text payload indices in single stage

timvisee:
Same as <https://github.com/qdrant/qdrant/pull/6999>, but for the full text index.

### Tasks
 - [x] Merge <https://github.com/qdrant/qdrant/pull/7003>
 - [x] Rebase on `dev`
 - [x] Undraft this PR

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

PR #7133, opened 2025-08-25
Restore missed index loading

generall:
Proper loading was lost in https://github.com/qdrant/qdrant/pull/7004

Index persists all changes, but doesn't keep them loaded, so after applying this patch all previously created indexes should be fixed automatically on load

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/immutable_text_index.rs
+++ b/lib/segment/src/index/field_index/full_text_index/immutable_text_index.rs
@@ -2,8 +2,8 @@ use std::path::PathBuf;
 
 use common::types::PointOffsetType;
 
-use super::inverted_index::immutable_inverted_index::ImmutableInvertedIndex;
 use super::inverted_index::InvertedIndex;
+use super::inverted_index::immutable_inverted_index::ImmutableInvertedIndex;
 #[cfg(feature = "rocksdb")]
 use super::inverted_index::mutable_inverted_index::MutableInvertedIndex;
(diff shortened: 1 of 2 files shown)


lib/segment/src/index/field_index/full_text_index/immutable_text_index.rs
```diff
@@ -67,10 +66,8 @@ impl ImmutableFullTextIndex {
 
     /// Open and load immutable full text index from mmap storage
     pub fn open_mmap(index: MmapFullTextIndex) -> Self {
-        let inverted_index = match index.inverted_index.storage.postings {
-            MmapPostingsEnum::Ids(_) => ImmutableInvertedIndex::ids_empty(),
-            MmapPostingsEnum::WithPositions(_) => ImmutableInvertedIndex::positions_empty(),
-        };
+        let inverted_index = ImmutableInvertedIndex::from(&index.inverted_index);
+
```
coszio, line 70: 🤦

coszio (approved):
🥾 🐛

lib/segment/src/index/field_index/full_text_index/immutable_text_index.rs
timvisee: I missed this line here: https://github.com/qdrant/qdrant/pull/7004/files#diff-11f65a2be5444195ffec029c4e72c07a71135e9050d228d00cdb9044f3518d06L138

☹️

timvisee (approved):

PR #7145, opened 2025-08-26
Enable BQ in appendable segments

IvanPleshkov:
This PR introduces new feature `appendable_quantization` and implements this feature.


lib/collection/src/config.rs
```diff
@@ -529,8 +549,15 @@ impl CollectionParams {
...
-        let vector_data = self.to_base_vector_data().map_err(|err| {
+    /// Convert into unoptimized segment config
+    ///
+    /// It is the job of the segment optimizer to change this configuration with optimized settings
+    /// based on threshold configurations.
+    pub fn to_base_segment_config(
+        &self,
+        collection_quantization: &Option<QuantizationConfig>,
+    ) -> CollectionResult<SegmentConfig> {
+        let vector_data = self.to_base_vector_data(collection_quantization).map_err(|err| {
```
timvisee: Is it necessary to pass this in here? Can't we derive it from the `CollectionParams` itself, like we do with all the other config properties?

lib/collection/src/config.rs
```diff
@@ -467,7 +467,28 @@ impl CollectionParams {
     /// based on threshold configurations.
     pub fn to_base_vector_data(
         &self,
+        collection_quantization: &Option<QuantizationConfig>,
     ) -> CollectionResult<HashMap<VectorNameBuf, VectorDataConfig>> {
+        let quantization_fn = |quantization_config: &Option<QuantizationConfig>| {
+            common::flags::feature_flags()
+                .appendable_quantization
+                .then(|| {
```
timvisee: Nitpick: I vote to move this `common::flags::feature_flags()` and `then()` to the place where we call `quantization_fn` below. That'll make it a bit cleaner, imo.

timvisee: Applied this here: https://github.com/qdrant/qdrant/pull/7145/commits/b828c5581e0cc47c0bba9ee29eac87d18b7b74b8

lib/collection/src/config.rs
IvanPleshkov: Good idea, I gonna try

IvanPleshkov: `CollectionParams` itself does not have this information, it's presented on a higher level of aggregation, like `CollectionConfigInternal`, where we have:
```
pub struct CollectionConfigInternal {
    #[validate(nested)]
    pub params: CollectionParams,
    #[serde(default)]
    .....
}
```
It's not a wrong design of configs, quantization can be applied to a single named vector or to the whole collection (each named vector will have collection-level quantization if a named vector does not specify it).

We can try, for instance, to move this method to a `CollectionConfigInternal` where we have all required data. But we can't do it directly, `CollectionConfigInternal` is not everywhere. For instance, `SegmentOptimizer` does not have the whole collection config, only their parts:
```
pub trait SegmentOptimizer {
    /// Get basic segment config
    fn collection_params(&self) -> CollectionParams;

    /// Get quantization config
    fn quantization_config(&self) -> Option<QuantizationConfig>;

......
}
```
And then we construct temp appendable segment:
```
    /// Build temp segment
    fn temp_segment(&self, save_version: bool) -> CollectionResult<LockedSegment> {
        let collection_params = self.collection_params();
        let quantization_config = self.quantization_config();
        let config = SegmentConfig {
            vector_data: collection_params.to_base_vector_data(&quantization_config)?,
            sparse_vector_data: collection_params.to_sparse_vector_data()?,
            payload_storage_type: collection_params.payload_storage_type(),
        };
        Ok(LockedSegment::new(build_segment(
            self.segments_path(),
            &config,
            save_version,
        )?))
    }
```

It seems I cannot use `CollectionConfigInternal` in `SegmentOptimizer` because `CollectionConfigInternal` has a lot of additional fields which are unknown in optimizer like UUID.

IvanPleshkov: The short answer is: it seems, we cannot do it easily. If I'm wrong somewhere, please correct me

lib/collection/src/config.rs
```diff
@@ -467,7 +467,23 @@ impl CollectionParams {
...
+                    .then(|| quantization_config.clone())
+            } else if let Some(collection_quantization) = collection_quantization {
+                // Only if there is no `quantization_config` we may start using `collection_quantization` (to avoid mixing quantizations between segments)
+                collection_quantization
+                    .is_appendable()
+                    .then(|| collection_quantization.clone())
+            } else {
+                None
+            }
+        };
```
JojiiOfficial:
```suggestion
            let quantization_fn = |quantization_config: Option<&QuantizationConfig>| {
            quantization_config
            // Only if there is no `quantization_config` we may start using `collection_quantization` (to avoid mixing quantizations between segments)
                .or(collection_quantization.as_ref())
                .and_then(|i| i.is_appendable().then(|| i.clone()))
        };
```

lib/collection/src/config.rs
```diff
@@ -467,7 +467,23 @@ impl CollectionParams {
     /// based on threshold configurations.
     pub fn to_base_vector_data(
         &self,
+        collection_quantization: &Option<QuantizationConfig>,
```
JojiiOfficial:
```suggestion
        collection_quantization: Option<&QuantizationConfig>,
```

lib/collection/src/config.rs
```diff
@@ -479,8 +495,11 @@ impl CollectionParams {
                         distance: params.distance,
                         // Plain (disabled) index
                         index: Indexes::Plain {},
-                        // Disabled quantization
-                        quantization_config: None,
+                        // Quantizaton config in appendable segment if runtime feature flag is set
+                        quantization_config: common::flags::feature_flags()
+                            .appendable_quantization
+                            .then(|| quantization_fn(&params.quantization_config))
```
JojiiOfficial:
```suggestion
                            .then(|| quantization_fn(params.quantization_config.as_ref()))
```

lib/collection/src/config.rs
```diff
@@ -529,8 +548,15 @@ impl CollectionParams {
...
 
-    pub fn to_segment_config(&self) -> CollectionResult<SegmentConfig> {
-        let vector_data = self.to_base_vector_data().map_err(|err| {
+    /// Convert into unoptimized segment config
+    ///
+    /// It is the job of the segment optimizer to change this configuration with optimized settings
+    /// based on threshold configurations.
+    pub fn to_base_segment_config(
+        &self,
+        collection_quantization: &Option<QuantizationConfig>,
```
JojiiOfficial:
```suggestion
        collection_quantization: Option<&QuantizationConfig>,
```

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -419,7 +419,12 @@ impl LocalShard {
                 "Shard has no appendable segments, this should never happen. Creating new appendable segment now",
             );
             let segments_path = LocalShard::segments_path(shard_path);
-            let segment_config = collection_config.read().await.params.to_segment_config()?;
+            let segment_config = {
```
JojiiOfficial: I see the same code being used in other functions. Does it make sense to create a function for it to reduce code duplication?

lib/collection/src/update_handler.rs
```diff
@@ -464,6 +464,7 @@ impl UpdateHandler {
         segments: &LockedSegmentHolder,
         segments_path: &Path,
         collection_params: &CollectionParams,
+        collection_quantization: &Option<QuantizationConfig>,
```
JojiiOfficial:
```suggestion
        collection_quantization: Option<&QuantizationConfig>,
```

JojiiOfficial:
Small nit but I'd suggest to always prefer `Option<&T>` over `&Option<T>` in function signatures.

lib/collection/src/config.rs
timvisee: Thanks for your elaborate answer. I don't have a bright idea to improve the current implementation, so let's keep it as is. I understand that the config being owned at different levels of nesting makes it hard.

lib/collection/src/config.rs
timvisee: Good suggestion. I applied it with a `filter()` rather than `.and_then(.then())` here: https://github.com/qdrant/qdrant/pull/7145/commits/c8dd953ad6e0aaed49cab61a29fd83e55d3589be

JojiiOfficial: TIL there is a Option::filter(). Thanks!

docs/redoc/master/openapi.json
```diff
@@ -11746,6 +11746,11 @@
             "description": "Migrate RocksDB based payload indices into new format on start.\n\nRebuilds a new payload index from scratch.",
             "default": false,
             "type": "boolean"
+          },
+          "appendable_quantization": {
+            "description": "Use appendable quantization in appendable plain segments.",
+            "default": false,
+            "type": "boolean"
           }
```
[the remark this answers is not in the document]
IvanPleshkov: It's an autogenerated file

lib/collection/src/update_handler.rs
```diff
@@ -464,6 +464,7 @@ impl UpdateHandler {
         segments: &LockedSegmentHolder,
         segments_path: &Path,
         collection_params: &CollectionParams,
+        collection_quantization: Option<&QuantizationConfig>,
         thresholds_config: &OptimizerThresholds,
         payload_index_schema: Arc<SaveOnDisk<PayloadIndexSchema>>,
     ) -> OperationResult<()> {
```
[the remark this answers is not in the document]
IvanPleshkov: WDYM?

lib/collection/src/shards/local_shard/mod.rs
IvanPleshkov: Done, I added `CollectionConfigInternal::to_base_segment_config` to simplify this part and make it looking like before.

xzfc (approved):

PR #7157, opened 2025-08-27
Bump segment version on {delete,update,upsert}-by-filter requests

JojiiOfficial:
Currently we don't acknowledge update-requests by filters that don't match any points, in write ahead log.
This triggers WAL replay when restart Qdrant. If there were many of such no-op operations, or large filters, this
sometimes significantly slowed down startup needlessly.
This PR fixes this issue by bumping segment versions manually, if no point was matched.


lib/shard/src/proxy_segment/mod.rs
```diff
@@ -320,6 +320,13 @@ impl ProxySegment {
 
         Ok(())
     }
+
+    pub(super) fn bump_version(&mut self, op: SeqNumberType) {
+        match &self.wrapped_segment {
+            LockedSegment::Original(rw_lock) => rw_lock.write().bump_segment_version(op),
+            LockedSegment::Proxy(rw_lock) => rw_lock.write().bump_version(op),
+        }
+    }
```
timvisee: In a proxy segment, the wrapped segment should remain unchanged. I'd argue the same is true for the segment version.

Let's bump the the write segment instead.

lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -948,60 +948,6 @@ impl SegmentEntry for ProxySegment {
         self.write_segment.get().read().check_error()
     }
 
-    fn delete_filtered<'a>(
```
timvisee, line 945: Self-check: can we actually remove this?

lib/shard/src/segment_holder/mod.rs
```diff
@@ -295,6 +295,16 @@ impl SegmentHolder {
...
 
+    /// Bumps the version of a random appendable segment.
+    pub fn bump_version_of_random_appendable(&self, op_num: SeqNumberType) {
+        if let Some(segment) = self.random_appendable_segment() {
+            match segment {
+                LockedSegment::Original(rw_lock) => rw_lock.write().bump_segment_version(op_num),
+                LockedSegment::Proxy(rw_lock) => rw_lock.write().bump_version(op_num),
+            }
+        }
+    }
```
timvisee: I _think_ it may be better to pick a segment through `self.aloha_random_write()` and supply `self.appendable_segments_ids()`.

That way we're sure we're getting a segment with write access. It may prevent blocking for a long time if some long running write operation is going on on the random segment we've selected.

lib/shard/src/proxy_segment/segment_entry.rs
timvisee: Yes, instead we select point IDs and delete those points directly:

https://github.com/qdrant/qdrant/blob/5936926c885cdfc6446d158eac50c2befa508026/lib/collection/src/collection_manager/segments_updater.rs#L785

lib/shard/src/proxy_segment/mod.rs
JojiiOfficial: Fixed!

lib/shard/src/segment_holder/mod.rs
JojiiOfficial: Fixed!

timvisee:
Thanks! Happy to see a test for this as well.

Before I approve, I tagged @ffuugoo for a review as well. We must validate this doesn't have a negative effect on partial snapshots for R/W segregation.

generall:
> We must validate this doesn't have a negative effect on partial snapshots

it will likely have.

timvisee:
@JojiiOfficial could you do some tests on this one to assert the behavior alongside partial snapshots?

The concern is that all segments that get their version bumped are transferred (almost) in full, even though there is no data change. That may be expensive on (large) R/W deployments.

An alternative might be to keep a separate version number in the segment holder. These empty updates could simply bump this separate version number. Though I'd rather keep it as is - not adding an extra version number if not required.

ffuugoo:
Why WAL flush/truncate does not solve this? Do we only truncate up to common (min? max?) version among all segments?

**UPD:** I think I remember now, that we truncate to either min or max flushed segment version, so if you do 100 ops that did not hit any segment you get 100 untruncated ops in WAL?

timvisee:
> I think I remember now, that we truncate to either min or max flushed segment version, so if you do 100 ops that did not hit any segment you get 100 untruncated ops in WAL?

Correct.

So here we (currently) artificially bump a segment version if a -by-filter operation matched zero points, and then acknowledge it in the WAL, so that we prevent this problem.

Or well, prevent a problem with WAL replay and very expensive operations taking a very long time, even if they were already properly flushed.

ffuugoo:
Hm. Maybe we could add an in-memory counter specifically for filtered ops that did not match anything then... 😞

lib/shard/src/segment_holder/mod.rs
```diff
@@ -70,6 +71,11 @@ pub struct SegmentHolder {
 
     /// Holds the first uncorrected error happened with optimizer
     pub optimizer_errors: Option<String>,
+
+    /// A special segment version that is usually used to keep track of manually bumped segment versions.
+    /// An example for this are operations that don't modify any points but could be expensive to recover from during WAL recovery.
+    /// To acknowledge them in WAL, we overwrite the max_persisted value in `Self::flush_all` with the segment version stored here.
+    max_persisted_segment_version_overwrite: AtomicOptionU64,
```
timvisee: Ah, I didn't consider that we need mutable access to the holder (if not using atomics or others).

Using an atomic is fine, but I think that `AtomicOptionU64` may be a bit over engineered.

What about a regular AtomicU64 that is not optional, being 0 by default? Having 0 by default will not be an issue in the flush stage when taking the max version across segments.

Wdyt?

JojiiOfficial: I considered a type like `AtomicOptionU64` would be a nice type to reuse in other places in future, with more clear semantics than just using 0 as `None` value.
But if you prefer using 0 here instead, I can change it 👍🏻

lib/shard/src/segment_holder/mod.rs
```diff
@@ -295,6 +300,12 @@ impl SegmentHolder {
         self.non_appendable_segments.keys().copied().collect()
     }
 
+    /// Bumps the version of a random appendable segment.
+    pub fn bump_version_of_random_appendable(&self, op_num: SeqNumberType) {
+        self.max_persisted_segment_version_overwrite
+            .store(op_num, Ordering::Relaxed);
+    }
```
timvisee: Let's update the function name, since we don't bump segment versions anymore.

JojiiOfficial: Done!

timvisee (approved):
Before merging I'd also like to see a review by @generall or @ffuugoo on the final approach.

lib/shard/src/segment_holder/mod.rs
```diff
@@ -792,6 +805,13 @@ impl SegmentHolder {
         // Grab and keep to segment RwLock's until the end of this function
         let segments = self.segment_locks(lock_order.iter().cloned())?;
 
+        // We can never have zero segments
+        // Having zero segments could permanently corrupt the WAL by acknowledging u64::MAX
+        assert!(
+            !segments.is_empty(),
+            "must always have at least one segment",
+        );
+
```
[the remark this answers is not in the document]
timvisee: I think a panic is fine because we deem the situation impossible.

PR #7165, opened 2025-08-28
Remove mlock flag from chunked mmap vectors

timvisee:
Remove the `mlock` flag, because it is always `false`.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/segment/src/vector_storage/chunked_mmap_vectors.rs
+++ b/lib/segment/src/vector_storage/chunked_mmap_vectors.rs
@@ -36,8 +36,6 @@ struct ChunkedMmapConfig {
     chunk_size_vectors: usize,
     dim: usize,
     #[serde(default)]
-    mlock: Option<bool>,
-    #[serde(default)]
     populate: Option<bool>,
 }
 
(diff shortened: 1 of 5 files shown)


lib/segment/src/vector_storage/chunked_mmap_vectors.rs
```diff
@@ -36,8 +36,6 @@ struct ChunkedMmapConfig {
     chunk_size_vectors: usize,
     dim: usize,
     #[serde(default)]
-    mlock: Option<bool>,
```
IvanPleshkov, line 39: This config is serializable. Is it save to remove field here without breaking backward compatibility?

IvanPleshkov: I place changes reques status because of this topic

IvanPleshkov (changes requested):

lib/segment/src/vector_storage/chunked_mmap_vectors.rs
timvisee: Yes because it'll default to `None`.

It is possible that it'll disable `mlock` when downgrading, but we don't support it in our latest version anyway.

IvanPleshkov: Got it! Didn't pay attention to `#[serde(default)]`

IvanPleshkov (approved):

xzfc (approved):

PR #7170, opened 2025-08-28
Qdrant on Edge

ffuugoo:
Adds `edge` crate with a simplified `LocalShard` re-implementation, that could be used as "embeddable" (e.g., SQLite) version of Qdrant. Only implements update and search, but more stuff would be added in follow-up PRs.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/edge/Cargo.toml
+++ b/lib/edge/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "edge"
+version = "0.1.0"
+authors = ["Qdrant Team <info@qdrant.tech>"]
+license = "Apache-2.0"
+edition = "2024"
+
+
+[dependencies]
+common = { path = "../common/common" }
+segment = { path = "../segment", default-features = false }
+shard = { path = "../shard" }
+
+log = { workspace = true }
+parking_lot = { workspace = true }
+thiserror = { workspace = true }
+wal = { workspace = true }
+
+
+[dev-dependencies]
+anyhow = "1.0"
--- a/lib/edge/src/lib.rs
+++ b/lib/edge/src/lib.rs
@@ -2,7 +2,7 @@ use std::num::NonZero;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
-use std::{fmt, fs};
+use std::{cmp, fmt, fs};
 
 use common::counter::hardware_accumulator::HwMeasurementAcc;
 use common::counter::hardware_counter::HardwareCounterCell;
@@ -36,19 +36,39 @@ const SEGMENTS_PATH: &str = "segments";
 impl Shard {
     pub fn load(path: &Path, mut config: Option<SegmentConfig>) -> OperationResult<Self> {
         let wal_path = path.join(WAL_PATH);
+
+        if !wal_path.exists() {
+            fs::create_dir(&wal_path).map_err(|err| {
+                OperationError::service_error(format!(
+                    "failed to create WAL directory {}: {err}",
+                    wal_path.display(),
+                ))
+            })?;
+        }
+
         let wal: SerdeWal<CollectionUpdateOperations> =
             SerdeWal::new(&wal_path, default_wal_options()).map_err(|err| {
                 OperationError::service_error(format!(
                     "failed to open WAL {}: {err}",
-                    wal_path.display()
+                    wal_path.display(),
                 ))
             })?;
 
         let segments_path = path.join(SEGMENTS_PATH);
+
+        if !segments_path.exists() {
+            fs::create_dir(&segments_path).map_err(|err| {
+                OperationError::service_error(format!(
+                    "failed to create segmens directory {}: {err}",
+                    segments_path.display(),
+                ))
+            })?;
+        }
+
         let segments_dir = fs::read_dir(&segments_path).map_err(|err| {
             OperationError::service_error(format!(
                 "failed to read segments directory {}: {err}",
-                segments_path.display()
+                segments_path.display(),
             ))
         })?;
 
@@ -58,7 +78,7 @@ impl Shard {
             let entry = entry.map_err(|err| {
                 OperationError::service_error(format!(
                     "failed to read entry in segments directory {}: {err}",
-                    segments_path.display()
+                    segments_path.display(),
                 ))
             })?;
 
@@ -79,7 +99,7 @@ impl Shard {
             {
                 log::warn!(
                     "Skipping hidden segment directory {}",
-                    segment_path.display()
+                    segment_path.display(),
                 );
                 continue;
             }
@@ -87,7 +107,7 @@ impl Shard {
             let segment = load_segment(&segment_path, &AtomicBool::new(false)).map_err(|err| {
                 OperationError::service_error(format!(
                     "failed to load segment {}: {err}",
-                    segment_path.display()
+                    segment_path.display(),
                 ))
             })?;
 
@@ -119,7 +139,7 @@ impl Shard {
             segment.check_consistency_and_repair().map_err(|err| {
                 OperationError::service_error(format!(
                     "failed to repair segment {}: {err}",
-                    segment_path.display()
+                    segment_path.display(),
                 ))
             })?;
 
@@ -300,7 +320,7 @@ impl Shard {
             }
         }
 
-        let _ = points.drain(..offset);
+        let _ = points.drain(..cmp::min(points.len(), offset));
 
         Ok(points)
     }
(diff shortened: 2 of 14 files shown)


lib/edge/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "edge"
+version = "0.1.0"
+authors = [
+    "Andrey Vasnetsov <vasnetsov93@gmail.com>",
```
generall:
```suggestion
```

lib/segment/src/types.rs
```diff
@@ -3133,25 +3133,31 @@ impl From<bool> for WithPayload {
...
-                payload_selector: Some(x.clone()),
+                payload_selector: Some(selector),
             },
         }
     }
 }
 
+impl From<&WithPayloadInterface> for WithPayload {
+    fn from(interface: &WithPayloadInterface) -> Self {
+        interface.clone().into()
```
generall:
```suggestion
        WithPayload::from(interface.clone())
```

lib/edge/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+            QueryVector::Context(_) => (),
+        }
+
+        if let Some(score_threshold) = score_threshold {
+            // TODO: find+truncate instead of retain
+            points.retain(|point| distance.check_threshold(point.score, score_threshold));
+        }
+
+        let _ = points.drain(..points.len().saturating_sub(offset));
+
```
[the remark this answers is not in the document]
ffuugoo: lol, thanks

lib/edge/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+                && name.starts_with(".")
+            {
+                log::warn!(
+                    "Skipping hidden segment directory {}",
+                    segment_path.display()
+                );
+                continue;
+            }
+
+            let segment = load_segment(&segment_path, &AtomicBool::new(false)).map_err(|err| {
```
generall, line 107: `AtomicBool` might be better to expose to function args, as we might still want to handle SIGINT

lib/edge/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+                })?;
+
+                continue;
+            };
+
+            if let Some(config) = &config {
+                // TODO: Is simple equality check sufficient? 🤔
+                if config != segment.config() {
+                    return Err(OperationError::service_error(format!(
+                        "segment {} does not match expected segment config: expected {:?}, but received {:?}",
```
generall: Segments can have different configs within same shard, e.g. indexed and plain segments.

ffuugoo: Good point. I'll add some "base" config and/or a better compatibility check then.

lib/edge/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+const WAL_PATH: &str = "wal";
+const SEGMENTS_PATH: &str = "segments";
+
+impl Shard {
+    pub fn load(path: &Path, mut config: Option<SegmentConfig>) -> OperationResult<Self> {
+        let wal_path = path.join(WAL_PATH);
+        let wal: SerdeWal<CollectionUpdateOperations> =
+            SerdeWal::new(&wal_path, default_wal_options()).map_err(|err| {
```
generall, line 50: shouldn't it be persisted WAL options?

ffuugoo: Maybe. For now, I just copied default options that we have in Qdrant. But _I think_ it should be able to load existing WAL with any options. In the worst case, `segment_capacity` must match/be persisted, `segment_queue_len` and `retain_closed` should work regardless. Anyway, I plan to test this sometime later and adjust accordingly.

lib/edge/src/lib.rs
ffuugoo: Also good point, but I think I'll expose it later, for now it's just noise.

lib/edge/src/lib.rs
ffuugoo: Some questions:
- `storage_type` and `index` could be different between segments
- what about `quantization_config`, `multivector_config` and `datatype`? I suppose they all should be the same for all segments?
- and I think `sparse_vector_data` and `payload_storage_type` could also be different?

generall: `quantization_config` can be different.

lib/edge/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+log = { workspace = true }
+parking_lot = { workspace = true }
+thiserror = { workspace = true }
+wal = { workspace = true }
+
+
+[dev-dependencies]
+anyhow = "1.0"
+# clap = { version = "4.5", features = ["derive"] }
```
timvisee: Shall we just remove this one?

```suggestion
```

lib/edge/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+                    "Skipping non-directory segment entry {}",
+                    segment_path.display(),
+                );
+
+                continue;
+            }
+
+            if let Some(name) = segment_path.file_name()
+                && let Some(name) = name.to_str()
+                && name.starts_with(".")
```
timvisee, line 98: I didn't know, or realize, the right side of `&&` could use a variable from the left side 🤯

lib/edge/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+            }
+
+            QueryVector::RecommendBestScore(_) => (),
+            QueryVector::RecommendSumScores(_) => (),
+            QueryVector::Discovery(_) => (),
+            QueryVector::Context(_) => (),
+        }
+
+        if let Some(score_threshold) = score_threshold {
+            // TODO: find+truncate instead of retain
```
timvisee: Can we do this now?

Maybe with a debug assertion before it that asserts the list of points is sorted the way we expect.

lib/edge/src/lib.rs
ffuugoo: bye bye `is_some_and`/`is_none_or` 👋

lib/edge/src/lib.rs
ffuugoo: Two questions:
- `ScoredPoint::cmp` is sorted by `ScoredPoint::score` OR `ScoredPoint::order_value`. I'd guess, when using plain search (e.g., `Segment::search_batch`) `order_value` is always `None`?
- `Distance::check_threshold` is `score > threshold` or `score < threshold` depending on `Distance`, does this mean that points returned by `Segment::search_batch` and `BatchResultAggregator::into_topk` are sorted accordingly?

timvisee: > does this mean that points returned by Segment::search_batch and BatchResultAggregator::into_topk are sorted accordingly?

Yes, simply because the distance metric defines sorting order.

The debug assertion might help us prove that.

ffuugoo: Done.

lib/edge/src/lib.rs
ffuugoo: Done.

ffuugoo:
I've implemented `SegmentConfig::is_compatible` check, I think this was the last remaining comment on the PR. @generall @timvisee can you, please, do a final review pass?

lib/edge/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+            let below_threshold = points
+                .iter()
+                .enumerate()
+                .find(|(_, point)| distance.check_threshold(point.score, score_threshold));
+
+            if let Some((below_threshold_idx, _)) = below_threshold {
+                points.truncate(below_threshold_idx + 1);
+            }
+        }
```
ffuugoo, line 321: Nice find. Fixed. Good bot.

lib/segment/src/types.rs
```diff
@@ -1526,6 +1576,29 @@ impl VectorDataConfig {
...
+            storage_type: _,
+            index: _,
+            quantization_config: _,
+            multivector_config: _,
+            datatype: _,
+        } = self;
+
+        self.size == other.size
+            && self.distance == other.distance
+            && self.multivector_config == other.multivector_config // TODO: 🤔
```
timvisee: Maybe we want to implement is_compatible on the config struct itself, just to make this behavior a bit less hidden.

lib/segment/src/types.rs
```diff
@@ -1519,9 +1587,45 @@ impl VectorDataConfig {
...
+                MultiVectorConfig::is_compatible,
+            )
+    }
+}
+
+fn is_opt_compatible<T, F: Fn(T, T) -> bool>(this: Option<T>, other: Option<T>, check: F) -> bool {
+    match (this, other) {
+        (Some(this), Some(other)) => check(this, other),
+        _ => false,
+    }
```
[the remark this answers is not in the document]
ffuugoo: Good bot. Fixed.

PR #7166, opened 2025-08-28
Generic AccessPattern

xzfc:
This PR removes duplications of `get_something`/`get_something_sequential` by replacing them with a single generic method. Similar to #6750.

Before this PR:
```rust
fn get_dense(&self, key: PointOffsetType) -> &[T] {
    self.mmap_store
        .as_ref()
        .unwrap()
        .get_vector_opt(key)
        .unwrap_or_else(|| panic!("vector not found: {key}"))
}

fn get_dense_sequential(&self, key: PointOffsetType) -> &[T] {
    self.mmap_store
        .as_ref()
        .unwrap()
        .get_vector_opt_sequential(key)
        .unwrap_or_else(|| panic!("vector not found: {key}"))
}

let _ = vector_storage.get_dense(i);
let _ = vector_storage.get_dense_sequential(i);
```

After this PR:
```rust
fn get_dense<P: AccessPattern>(&self, key: PointOffsetType) -> &[T] {
    self.mmap_store
        .as_ref()
        .unwrap()
        .get_vector_opt::<P>(key)
        .unwrap_or_else(|| panic!("vector not found: {key}"))
}

let _ = vector_storage.get_dense::<Random>(i);
let _ = vector_storage.get_dense::<Sequential>(i);
```

PR #7171, opened 2025-08-28
CompressedWithVectors: include base/full vectors

xzfc:
Depends on #7166. (ignore the first commit during review)
PR #6982 introduced new on-disk format that includes HNSW graph links and (quantized) link vectors. This PR extends this format to include also base vectors (aka full vectors).

Updated description of the format:
> Compressed links with vectors.
>
> Similar to [`CompressionInfo::Compressed`], but includes vectors.
> - Each node on level 0 has a fixed-size "base" vector.
> - Each link is accompanied by a fixed-size "link" vector.
>
> ```text
> [BBBB#ccccccc_LLLLLL_][BBBB#ccccccc_LLLLLL_][BBBB#ccccccc_LLLLLL_]
> [neighbors for node 0][neighbors for node 1][neighbors for node 2]...
> ```
> Where:
> 1. `B` is a base vector (i.e. `Vec<u8>` of fixed size).
>    Only present on level 0, omitted on higher levels.
> 2. `#` is a varint-encoded length.
>    This value == number of links == number of link vectors.
> 3. `c` are compressed links (i.e. a compressed form of `Vec<u32>`).
> 4. `_` is a padding to make link vectors aligned.
> 5. `L` are encoded link vectors, one per link (i.e. `Vec<Vec<u8>>`).
> 6. `_` is a padding to make the next base vector aligned.
>    Only present on level 0, omitted on higher levels.


lib/segment/src/index/hnsw_index/graph_links/header.rs
```diff
@@ -42,10 +42,18 @@ pub(super) struct HeaderCompressedWithVectors {
...
 pub(super) const HEADER_VERSION_COMPRESSED: u64 = 0xFFFF_FFFF_FFFF_FF01;
 pub(super) const HEADER_VERSION_COMPRESSED_WITH_VECTORS: u64 = 0xFFFF_FFFF_FFFF_FF02;
+
+#[derive(FromBytes, Immutable, IntoBytes, KnownLayout)]
+#[repr(C)]
+/// Packed representation of [`std::alloc::Layout`].
+pub(super) struct PackedVectorLayout {
+    pub(super) size: LittleU64,
+    pub(super) alignment: u8,
+}
```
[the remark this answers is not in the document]
xzfc: Reject. I'd like to keep it simple: I do not expect alignment be larger than 16 (`u128`).

lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -68,34 +69,51 @@ pub enum GraphLinksFormatParam<'a> {
...
+
+    /// Get the layout of base and link vectors.
+    fn vector_layout(&self) -> OperationResult<VectorLayout>;
 }
 
-impl GraphLinksVectors for QuantizedVectors {
-    fn get_vector(&self, point_id: PointOffsetType) -> OperationResult<&[u8]> {
-        Ok(self.get_quantized_vector(point_id))
-    }
+pub struct VectorLayout {
```
IvanPleshkov: I propose to rename this struct because I find non-obvious that `VectorLayout`is not a parameter of vector, it's a pair of vector and quantization parameters. We can discuss proper name, I can propose at least `LinksVectorsLayout` to mark that this struct is related to the links data and not related to vector storage/user vector input etc

lib/segment/src/index/hnsw_index/graph_links/header.rs
```diff
@@ -42,10 +42,18 @@ pub(super) struct HeaderCompressedWithVectors {
...
+    pub(super) zero_padding: [u8; 3],
 }
 
 pub(super) const HEADER_VERSION_COMPRESSED: u64 = 0xFFFF_FFFF_FFFF_FF01;
 pub(super) const HEADER_VERSION_COMPRESSED_WITH_VECTORS: u64 = 0xFFFF_FFFF_FFFF_FF02;
+
+#[derive(FromBytes, Immutable, IntoBytes, KnownLayout)]
+#[repr(C)]
+/// Packed representation of [`std::alloc::Layout`].
+pub(super) struct PackedVectorLayout {
```
IvanPleshkov, line 60: The same as of `VectorLayout`: rename proposal because it's not obvious that `PackedVectorLayout` contains quantization params.

lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -68,34 +69,51 @@ pub enum GraphLinksFormatParam<'a> {
...
 }
 
-#[cfg(test)]
-impl GraphLinksVectors for Vec<Vec<u8>> {
-    fn get_vector(&self, point_id: PointOffsetType) -> OperationResult<&[u8]> {
-        Ok(&self[point_id as usize])
+impl<'a> GraphLinksVectors for StorageGraphLinksVectors<'a> {
+    fn get_base_vector(&self, point_id: PointOffsetType) -> OperationResult<&[u8]> {
+        self.vector_storage
+            .get_vector_bytes_opt::<Sequential>(point_id)
```
IvanPleshkov, line 103: I didnt review https://github.com/qdrant/qdrant/pull/7166.
Can you please explain why `Sequential` here instead of `Random`?

lib/segment/src/index/hnsw_index/graph_links/header.rs
```diff
@@ -42,10 +42,18 @@ pub(super) struct HeaderCompressedWithVectors {
     pub(super) offsets_parameters: bitpacking_ordered::Parameters,
     pub(super) m: LittleU64,
     pub(super) m0: LittleU64,
-    pub(super) vector_size_bytes: LittleU64,
-    pub(super) vector_alignment: u8,
-    pub(super) zero_padding: [u8; 4],
+    pub(super) base_vector_layout: PackedVectorLayout,
+    pub(super) link_vector_layout: PackedVectorLayout,
+    pub(super) zero_padding: [u8; 3],
```
IvanPleshkov: Is it possible to avoid this field and have zeroing automatically in the serializer? Not in this PR, in the near future as a part of a milestone. Or at least can you please make a unit test that checks that all headers should have a fixed size +comments

lib/segment/src/index/hnsw_index/graph_links/view.rs
```diff
@@ -164,8 +171,10 @@ impl GraphLinksView<'_> {
         let (header, data) = HeaderCompressedWithVectors::ref_from_prefix(data)
             .map_err(|_| error_unsufficent_size())?;
         debug_assert_eq!(header.version.get(), HEADER_VERSION_COMPRESSED_WITH_VECTORS);
-        let vector_alignment = header.vector_alignment;
-        if !vector_alignment.is_power_of_two() {
+
+        // Check alignment of link vectors. Alignment of base vectors is checked
+        // in `Layout::from_size_align` later.
+        if !header.link_vector_layout.alignment.is_power_of_two() {
```
IvanPleshkov: Very unimportant nit. Can we do a separate function `PackedVectorLayout::validate(&self) -> OperationResult<()>` and move `is_power_of_two` with a comment?

lib/segment/src/index/hnsw_index/graph_links/view.rs
```diff
@@ -76,26 +77,32 @@ pub(super) enum CompressionInfo<'a> {
...
+        ///    Only present on level 0, omitted on higher levels.
         neighbors: &'a [u8],
         offsets: bitpacking_ordered::Reader<'a>,
         hnsw_m: HnswM,
         bits_per_unsorted: u8,
+        base_vector_layout: Layout,
         /// `NonZero` to avoid handling unlikely corner cases.
-        vector_size_bytes: NonZero<usize>,
-        vector_alignment: u8,
+        link_vector_size: NonZero<usize>,
```
IvanPleshkov, line 104: Can we use for `link_vector_size` and `link_vector_alignment` the `Layout` type as for `base_vector_layout`? If yes, you can avoid `header.link_vector_layout.alignment.is_power_of_two()` check in `load_compressed_with_vectors` and use better validation `Layout::from_size_align` for both base and link vectors

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -33,7 +34,12 @@ pub fn serialize_graph_links<W: Write + Seek>(
         GraphLinksFormatParam::Compressed => None,
         GraphLinksFormatParam::CompressedWithVectors(v) => {
             let vector_layout = v.vector_layout()?;
-            if vector_layout.size() % vector_layout.align() != 0 {
+            if vector_layout.base.size() % vector_layout.base.align() != 0 {
+                return Err(OperationError::service_error(
+                    "Vector size must be a multiple of its alignment",
```
IvanPleshkov:
```suggestion
                    "Base vector size must be a multiple of its alignment",
```

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -82,9 +88,10 @@ pub fn serialize_graph_links<W: Write + Seek>(
     }
 
     // 4. Write neighbors padding (if applicable)
-    if let Some(vector_layout) = vector_layout {
+    if let Some(vector_layout) = vector_layout.as_ref() {
         let pos = writer.stream_position()? as usize;
-        writer.write_zeros(pos.next_multiple_of(vector_layout.align()) - pos)?;
+        let alignment = vector_layout.base.align().max(vector_layout.link.align());
```
IvanPleshkov: You did `max` function between base and link alignments in multiple places. I propose to add `VectorLayout::alignment` function which calculates `max` and has a comment why do we need to calculate `max`

lib/segment/src/vector_storage/vector_storage_base.rs
```diff
@@ -89,6 +90,12 @@ pub trait VectorStorage {
     /// Get the vector by the given key if it exists
     fn get_vector_opt<P: AccessPattern>(&self, key: PointOffsetType) -> Option<CowVector<'_>>;
 
+    /// Get the raw bytes of the vector by the given key if it exists
+    fn get_vector_bytes_opt<P: AccessPattern>(&self, key: PointOffsetType) -> Option<&[u8]>;
```
IvanPleshkov: Is it possible to move this function from this trait into `DenseVectorStorage`? And in `DenseVectorStorage` we can do an implementation of this func using `get_dense`.
To access it from `VectorStorageEnum` we can do that same as `try_multi_vector_config`.
My proposal is ugly but at least we don't touch all storages implementation and avoid an unreachable code.

lib/segment/src/index/hnsw_index/graph_links/header.rs
xzfc: It's for alignment (`#[repr(C, align(8))]`), and it's enforced by compiler and the `zerocopy` derive macro. E.g. try to replace `3` with other value: you'll get a compiler error.

lib/segment/src/index/hnsw_index/graph_links.rs
xzfc: `serialize_graph_links` writes base vectors in a sequential order. I'll add a comment.

lib/segment/src/index/hnsw_index/graph_links/header.rs
IvanPleshkov: Okay, it there any comment about it? Would you like to add it?

lib/segment/src/index/hnsw_index/graph_links.rs
IvanPleshkov: I don't think it's required here

lib/segment/src/index/hnsw_index/graph_links/view.rs
xzfc: It is expanded so `NonZero<usize>` can be used. `<[T]>::chunks_exact(chunk_size)` has an `assert!(chunk_size != 0)` inside. By using `NonZero` we'll get:
- One less branch in a hot loop.
- One less potential panic.

lib/segment/src/index/hnsw_index/graph_links/serializer.rs
xzfc: Aligning to both base and link layouts simplifies the `GraphLinksView` implementation. I'd prefer to keep this logic inlined to avoid ravioli code.

lib/segment/src/vector_storage/vector_storage_base.rs
xzfc: Thanks, it made things simpler.

lib/segment/src/index/hnsw_index/graph_links/view.rs
xzfc: Added `PackedVectorLayout::try_into_layout`. The validation happen in `Layout::from_size_align()`.

lib/segment/src/index/hnsw_index/graph_links/header.rs
xzfc: Added `// for 8-byte alignment`.

lib/segment/src/index/hnsw_index/graph_links.rs
xzfc: Renamed into `GraphLinksVectorsLayout` and added more comments to this part of code.

PR #7173, opened 2025-08-28
Populate multivector quantization offsets

IvanPleshkov:
In dev there is a bug which is hard to catch. When we call popuate for quantized multivectors, we forget to call populate for multivector offsets file, only for inner vectors.

This PR fixes it. Also, it wraps `ChunkedMmapStorage` in a separate struct to make `quantized_vectors.rs` in the segment more readable.


generall:
huh, that explains something

PR #7174, opened 2025-08-28
Quantization storage builder for chunked mmap

IvanPleshkov:
This PR implements quantization storage builder for `ChunkedMmap`.

It allows the use of `ChunkedMmap` for the `encoding` function in the quantization crate, which is required to build quantization from non-empty `VectorStorage`.

Because `ChunkedMmap::push` returns `Result`, this PR also changed `EncodedStorageBuilder` trait and adds `Result` to `push_vector_data` function to avoid unwarps.

### DIFF
--- a/lib/segment/src/vector_storage/quantized/quantized_chunked_mmap_storage.rs
+++ b/lib/segment/src/vector_storage/quantized/quantized_chunked_mmap_storage.rs
@@ -88,13 +88,13 @@ impl QuantizedChunkedMmapStorageBuilder {
 }
 
 impl quantization::EncodedStorageBuilder for QuantizedChunkedMmapStorageBuilder {
-    type Storage = ChunkedMmapVectors<u8>;
+    type Storage = QuantizedChunkedMmapStorage;
 
-    fn build(self) -> std::io::Result<ChunkedMmapVectors<u8>> {
+    fn build(self) -> std::io::Result<QuantizedChunkedMmapStorage> {
         self.data.flusher()().map_err(|e| {
             std::io::Error::other(format!("Failed to flush quantization storage: {e}"))
         })?;
-        Ok(self.data)
+        Ok(QuantizedChunkedMmapStorage { data: self.data })
     }
 
     fn push_vector_data(&mut self, other: &[u8]) -> std::io::Result<()> {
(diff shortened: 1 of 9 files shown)


lib/segment/src/vector_storage/quantized/quantized_ram_storage.rs
```diff
@@ -117,9 +117,10 @@ impl quantization::EncodedStorageBuilder for QuantizedRamStorageBuilder {
         })
     }
 
-    fn push_vector_data(&mut self, other: &[u8]) {
-        // Memory for ChunkedVectors are already pre-allocated,
-        // so we do not expect any errors here.
-        self.vectors.push(other).unwrap();
+    fn push_vector_data(&mut self, other: &[u8]) -> std::io::Result<()> {
```
timvisee, line 120: Does a different error type here make more sense, like OperationResult/OperationError?

lib/segment/src/vector_storage/quantized/quantized_chunked_mmap_storage.rs
```diff
@@ -57,3 +59,48 @@ impl quantization::EncodedStorage for QuantizedChunkedMmapStorage {
...
+}
+
+impl quantization::EncodedStorageBuilder for QuantizedChunkedMmapStorageBuilder {
+    type Storage = QuantizedChunkedMmapStorage;
+
+    fn build(self) -> std::io::Result<QuantizedChunkedMmapStorage> {
+        self.data.flusher()().map_err(|e| {
+            std::io::Error::other(format!("Failed to flush quantization storage: {e}"))
+        })?;
+        Ok(QuantizedChunkedMmapStorage { data: self.data })
```
timvisee:
```suggestion
        let Self {
            data,
            hw_counter: _,
        } = self;

        data.flusher()().map_err(|e| {
            std::io::Error::other(format!("Failed to flush quantization storage: {e}"))
        })?;

        Ok(QuantizedChunkedMmapStorage { data })
```

timvisee (approved):

lib/segment/src/vector_storage/quantized/quantized_ram_storage.rs
IvanPleshkov: `OperationError` is defined in the segment crate, I cannot use it in the quantization crate

PR #7176, opened 2025-08-29
Use locked segment by reference without cloning Arc

timvisee:
Prefer to use a locked segment by reference, without cloning an Arc.

We use this function in at least 189 different places (including tests). This way we prevent unnecessarily bumping an atomic counter twice with each interaction.

I don't expect to see a noticeable difference in end to end benchmarks, but doing less work at runtime is always good. This function may be considered hot, and we've seen performance problems on very large CPUs when bumping a lot of atomics.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/shard/src/segment_holder/mod.rs
+++ b/lib/shard/src/segment_holder/mod.rs
@@ -579,7 +579,7 @@ impl SegmentHolder {
     /// Try to acquire read lock over the given segment with increasing wait time.
     /// Should prevent deadlock in case if multiple threads tries to lock segments sequentially.
     fn aloha_lock_segment_read(
-        segment: &'_ Arc<RwLock<dyn SegmentEntry>>,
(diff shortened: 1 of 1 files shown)


lib/shard/src/segment_holder/mod.rs
```diff
@@ -947,7 +947,7 @@ impl SegmentHolder {
             // Get segment to snapshot
             let segment = match proxy_segment {
                 LockedSegment::Proxy(proxy_segment) => {
-                    proxy_segment.read().wrapped_segment.clone().get()
+                    proxy_segment.read().wrapped_segment.get_cloned()
```
timvisee, line 950: Effectively the only place (outside tests) where we actually need the owned `Arc` by cloning it.

agourlay (approved):

JojiiOfficial (approved):

PR #7177, opened 2025-08-29
do not anonymize peer id

generall:


generall:
> Any specific reason for this one?

1. Peer ids are already exposed in other structures, so we are not hiding anything additional here
2. Peer ids are random, exposing them doesn't let us de-anon anything
3. With this PR change we will be able to restore shards-to-peers distribution, which is needed for checking if cluster is evenly balanced

PR #7185, opened 2025-08-29
make sure that grpc::Vector can always use new format

generall:
Some internal APIs were still relying on the old format of `grpc::Vector`. 
This PR covers last missing parts of missing conversions, so that in the next version we can deprecate old format and always generate structures using a new one

### DIFF
--- a/lib/api/src/conversions/vectors.rs
+++ b/lib/api/src/conversions/vectors.rs
@@ -7,6 +7,7 @@ use sparse::common::sparse_vector::SparseVector;
 use tonic::Status;
 
 use crate::grpc::qdrant as grpc;
+use crate::grpc::vector::Vector;
 use crate::rest::schema as rest;
 
 fn convert_to_plain_multi_vector(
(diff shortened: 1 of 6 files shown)


coszio (approved):

lib/api/src/conversions/vectors.rs
```diff
@@ -411,8 +411,8 @@ impl TryFrom<grpc::VectorsOutput> for VectorStructInternal {
 
 impl From<VectorInternal> for grpc::Vector {
     fn from(vector: VectorInternal) -> Self {
-        // ToDo: before deprecating `data`, `indices`, and `vectors_count`, ensure
-        // ToDo: that `vector` field is generated here.
+        // ToDo(v1.18): before deprecating `data`, `indices`, and `vectors_count`, ensure
```
generall: this specific part can be updated earlier, as it is only for internal cluster operations

lib/api/src/conversions/vectors.rs
```diff
@@ -410,6 +411,8 @@ impl TryFrom<grpc::VectorsOutput> for VectorStructInternal {
 
 impl From<VectorInternal> for grpc::Vector {
     fn from(vector: VectorInternal) -> Self {
+        // ToDo(v1.18): before deprecating `data`, `indices`, and `vectors_count`, ensure
```
generall:
```suggestion
        // ToDo(v1.17): before deprecating `data`, `indices`, and `vectors_count`, ensure
```

PR #7188, opened 2025-08-31
slow requests log

generall:
This PR creates a special debug API for logging top slow requests.

To do that efficiently, and don't clog API with huge amount of requests and vector-level details, this PR introduces `Generalize` trait, which is supposed to remove excessive details from request representation, while preserving structure and important details.


lib/collection/src/operations/generalizer/count.rs
```diff
@@ -0,0 +1,12 @@
+use crate::operations::generalizer::Generalizer;
+use crate::operations::types::CountRequestInternal;
+
+impl Generalizer for CountRequestInternal {
+    fn remove_details(&self) -> Self {
+        let CountRequestInternal { filter, exact } = self;
+        Self {
+            filter: filter.clone(),
+            exact: *exact,
+        }
```
timvisee, line 10: How about just `self.clone()`?


```suggestion
        self.clone()
```

Is it just so that the compiler tells us to update this when we change the type? Do you think it's worth all the extra boilerplate, as I see this used quite often?

lib/collection/src/operations/generalizer/update_persisted.rs
```diff
@@ -0,0 +1,298 @@
...
+
+use crate::operations::generalizer::Generalizer;
+
+impl Generalizer for Payload {
+    fn remove_details(&self) -> Self {
+        let mut stripped_payload = Payload::default();
+        stripped_payload.0.insert(
+            "keys".to_string(),
+            Value::Array(self.keys().cloned().sorted().map(Value::String).collect()),
+        );
```
timvisee, line 23: Is sorting still required here? I believe we don't use this to hash anymore, do we?

lib/collection/src/operations/conversions.rs
```diff
@@ -1028,7 +1028,7 @@ impl<'a> From<CollectionCoreSearchRequest<'a>> for api::grpc::qdrant::CoreSearch
             limit: *limit as u64,
             with_vectors: with_vector.clone().map(|wv| wv.into()),
             with_payload: with_payload.clone().map(|wp| wp.into()),
-            params: params.map(|sp| sp.into()),
+            params: params.clone().map(|sp| sp.into()),
```
timvisee: With `Copy` on the params type we can remove the clone.

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -1198,7 +1198,7 @@ impl VectorIndex for HNSWIndex {
 
         let exact_params = if exact {
             params.map(|params| {
-                let mut params = *params;
+                let mut params = params.clone();
```
timvisee: Let's implement `Copy` on `SearchParams` to use `*params` like before, that is cheaper.

lib/segment/src/types.rs
```diff
@@ -2140,6 +2152,23 @@ pub enum AnyVariants {
...
+                }
+            }
+            AnyVariants::Integers(index_set) => {
+                for item in index_set.iter() {
+                    item.hash(state);
+                }
+            }
+        }
+    }
+}
```
timvisee: Sorting?

lib/segment/src/types.rs
```diff
@@ -2847,8 +2906,16 @@ pub struct HasIdCondition {
     pub has_id: MaybeArc<AHashSet<PointIdType>>,
 }
 
+impl Hash for HasIdCondition {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        for id in self.has_id.iter().sorted() {
+            id.hash(state);
+        }
+    }
```
timvisee: I'm afraid the sorting over various iterators may be expensive. In this specific case it means we allocate a temporary Vec on the heap just to sort. We've seen users with humongous lists here.

How about we just xor all item hashes? Then it is ordering agnostic. I believe it'll be good enough, and doesn't need the intermediate Vec.

Could be done like this:

```suggestion
        self.has_id
            .iter()
            .map(|id| {
                let mut hasher = DefaultHasher::new();
                id.hash(&mut hasher);
                hasher.finish()
            })
            .reduce(std::ops::BitXor::bitxor)
            .hash(state);
    }

```

cc @xzfc

lib/segment/src/types.rs
```diff
@@ -2947,6 +3014,49 @@ pub enum Condition {
     CustomIdChecker(Arc<dyn CustomIdCheckerCondition + Send + Sync + 'static>),
 }
 
+impl Hash for Condition {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        match self {
+            Condition::Field(field_condition) => {
+                0u8.hash(state);
```
timvisee: I suggest to replace these manual variant values with something like

```rust
std::mem::discriminant(self).hash(state);
```

before the match statement.

[Docs](https://doc.rust-lang.org/std/mem/fn.discriminant.html)

lib/segment/src/types.rs
```diff
@@ -2947,6 +3014,49 @@ pub enum Condition {
...
+                6u8.hash(state);
+                filter.hash(state);
+            }
+            Condition::CustomIdChecker(_) => {
+                7u8.hash(state);
+                // We cannot hash the inner function, so we just use a constant value.
+                // This means that two different CustomIdChecker conditions will have the same hash,
+                // but that's acceptable since we cannot do better, and only expected to be used
+                // for logging and profiling purposes.
+                0u8.hash(state);
```
timvisee:
```suggestion
```

lib/shard/src/operations/point_ops.rs
```diff
@@ -454,6 +455,36 @@ pub enum BatchVectorStructPersisted {
...
+                for vector in multidense {
+                    for v in vector {
+                        for element in v {
+                            state.write_u32(element.to_bits());
+                        }
+                    }
+                }
+            }
+            BatchVectorStructPersisted::Named(named) => {
+                let keys: Vec<_> = named.keys().sorted().collect();
```
timvisee: Another candidate for an items+xor hash?

lib/shard/src/operations/point_ops.rs
```diff
@@ -569,6 +600,32 @@ pub enum VectorStructPersisted {
...
+            }
+            VectorStructPersisted::MultiDense(multi_vec) => {
+                for vec in multi_vec {
+                    for v in vec {
+                        state.write_u32(v.to_bits());
+                    }
+                }
+            }
+            VectorStructPersisted::Named(map) => {
+                let keys: Vec<_> = map.keys().sorted().collect();
```
timvisee: Another candidate for an items+xor hash?

src/actix/api/profiler_api.rs
```diff
@@ -0,0 +1,43 @@
...
+struct LogParams {
+    limit: Option<usize>,
+}
+
+#[derive(Serialize, JsonSchema)]
+struct SlowRequestsResponse {
+    requests: Vec<LogEntry>,
+}
+
+const DEFAULT_SLOW_REQUESTS_LIMIT: usize = 10;
```
timvisee, line 22: Not 100?

lib/collection/src/profiling/interface.rs
```diff
@@ -0,0 +1,62 @@
...
+    duration: std::time::Duration,
+) {
+    if duration < MIN_SLOW_REQUEST_DURATION {
+        return;
+    }
+
+    if let Some(listener) = REQUESTS_COLLECTOR.get() {
+        let message =
+            RequestProfileMessage::new(Box::new(request), duration, collection_name.into());
+        listener.send_if_available(message);
```
timvisee, line 28: Warning if queue is full?

lib/segment/src/types.rs
xzfc: For xor, duplicates will cancel each out. I.e., `hash([1,2]) == hash([3,1,2,3])`. (not sure if it matters in our case tho)

A bit better alternative is to use addition instead of `bitxor`. An even bit better alternative is to combine xor, addition, and addition over squares. (to make harder to find collisions; again, not sure if that matters in our case; but it's trivial to implement)

```rust
fn hash_unordered<T: Hash, I: Iterator<Item = T>, H: Hasher>(hasher: &mut H, iter: I) {
    let mut sx = Wrapping(0u64);
    let mut s1 = Wrapping(0u64);
    let mut s2 = Wrapping(0u64);

    for item in iter {
        let mut hasher = DefaultHasher::new();
        item.hash(&mut hasher);
        let h = Wrapping(hasher.finish());

        sx ^= h;
        s1 += h;
        s2 += h * h;
    }

    sx.0.hash(hasher);
    s1.0.hash(hasher);
    s2.0.hash(hasher);
}
```

Other option is to use some crate that implements "multiset hash" or "muhash", e.g. https://crates.io/crates/multiset-hash.

src/actix/api/profiler_api.rs
generall: this is for API response

lib/collection/src/operations/generalizer/update_persisted.rs
generall: It is needed for deduplicating requests. This simplified payload will be hashed.

lib/collection/src/operations/conversions.rs
generall: I don't think `params` should be copyable. I would prefer to make explicit clone, as it might potentially get expensive

lib/segment/src/index/hnsw_index/hnsw.rs
generall: I though copy and clone have the same price, but clone is explicit

lib/segment/src/types.rs
generall: Please feel free to include this into PR instead of sorting

lib/segment/src/types.rs
generall: sounds good

lib/collection/src/operations/conversions.rs
timvisee: Fair enough. In this case the optimizer should still do a memcopy for us anyway.

lib/segment/src/index/hnsw_index/hnsw.rs
timvisee: Hey might, but it's not guaranteed.

Copy always does a (fast!) memcopy. Clone normally uses custom clone logic, to satisfy heap allocated data for example. If a type is Copy compatible, it might also do the memcopy but it is not guaranteed.

[Docs](https://doc.rust-lang.org/std/marker/trait.Copy.html#whats-the-difference-between-copy-and-clone)

timvisee: [This](https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy) suggests that

> Generally speaking, if your type can implement Copy, it should.

So I'll make it copy.

Since this type isn't exposed as part of a public API, we can always change it in the future.

lib/collection/src/update_handler.rs
```diff
@@ -789,6 +805,12 @@ impl UpdateHandler {
                         )
                     });
 
+                    let duration = start_time.elapsed();
+
+                    log_request_to_collector(&collection_name, duration, move || {
+                        loggable_operation
+                    });
+
```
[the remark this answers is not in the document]
timvisee: Both are equal

lib/collection/src/operations/generalizer/update_persisted.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+}
+
+impl Generalizer for VectorPersisted {
+    fn remove_details(&self) -> Self {
+        match self {
+            VectorPersisted::Dense(dense) => VectorPersisted::Dense(vec![dense.len() as f32]),
+            VectorPersisted::Sparse(sparse) => VectorPersisted::Sparse(
+                SparseVector::new(vec![sparse.len() as DimId], vec![0.0]).unwrap(),
+            ),
```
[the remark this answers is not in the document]
timvisee: Unfailable

lib/segment/src/types.rs
xzfc: Done: added `unordered_hash_unique`.

Notes:
- In this particular PR it's used only for hashmaps so it is guaranteed that there are no duplicates. (so I did not include addition/addition over squares)
- `.reduce` returns `Option<u64>`, not `u64`, making it a bit slower, so I've used `.fold` instead.

lib/segment/src/vector_storage/query/reco_query.rs
```diff
@@ -12,6 +16,34 @@ pub struct RecoQuery<T> {
...
+            positives,
+            negatives,
+        } = self;
+
+        let mut state = serializer.serialize_struct("RecoQuery", 2)?;
+        state.serialize_field("positives", positives)?;
+        state.serialize_field("negatives", negatives)?;
+        state.end()
+    }
+}
```
timvisee, line 34: I tried to just derive on the type. But that means we have to make generic `T` support serialize all throughout the codebase. That turns out to be much more work than this custom implementation.

So I agree to do it like this 👍

lib/segment/src/data_types/vectors.rs
```diff
@@ -296,7 +323,24 @@ impl<T> TypedMultiDenseVector<T> {
...
 
+impl Hash for MultiDenseVectorInternal {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let Self {
+            flattened_vectors,
+            dim,
+        } = self;
+        dim.hash(state);
+        for element in flattened_vectors {
+            element.to_bits().hash(state);
```
timvisee: Utilize OrderedFloat here too?

```suggestion
            OrderedFloat(*element).hash(state)
```

lib/collection/src/operations/loggable.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+impl Loggable for CollectionUpdateOperations {
+    fn to_log_value(&self) -> Value {
+        serde_json::to_value(self).unwrap_or_default()
+    }
+
+    fn request_name(&self) -> &'static str {
+        "points-update"
+    }
```
timvisee, line 24: Want more differentiation here?

I'd be less curious about a long running field index creation operation, versus a point payload set.

Of course we can always dive into the request body. Just curious what your idea was around this name.

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -280,13 +284,19 @@ impl ShardOperation for LocalShard {
...
+            .do_planned_query(
+                planned_query,
+                search_runtime_handle,
+                timeout,
+                hw_measurement_acc,
+            )
+            .await;
+
+        let elapsed = start_time.elapsed();
+        log_request_to_collector(&self.collection_name, elapsed, || requests.remove_details());
```
timvisee, line 297: Why do we only have this collector line in `query_batch`? And not in `facet`, `retrieve`, `count`... for example.

(I'm checking this in the meanwhile)

lib/collection/src/update_handler.rs
```diff
@@ -789,6 +805,12 @@ impl UpdateHandler {
                         )
                     });
 
+                    let duration = start_time.elapsed();
+
+                    log_request_to_collector(&collection_name, duration, move || {
+                        loggable_operation
+                    });
```
timvisee, line 812: We should only log here if we actually run `CollectionUpdater::update`. It may not happen if flushing WAL above failed.

lib/segment/src/types.rs
```diff
@@ -2502,6 +2573,15 @@ pub struct GeoRadius {
     pub radius: f64,
 }
 
+impl Hash for GeoRadius {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        let GeoRadius { center, radius } = self;
+        center.hash(state);
+        // Hash f64 by converting to bits
+        radius.to_bits().hash(state);
```
timvisee, line 2581:
```suggestion
        OrderedFloat(*radius).hash(state);
```

lib/collection/src/operations/loggable.rs
generall: The idea is to use this as a separator for multiple queues later. 

> long running field index creation operation

I would actually want to know this one, because it might be a reason for blocking other requests which user didn't expect

lib/collection/src/shards/local_shard/shard_ops.rs
generall: this is in todo for further PRs

lib/collection/src/operations/loggable.rs
timvisee: Yeah. I don't suggest to hide it. Only asking whether we want to define the request name in more detail.

I don't have a strong objection.

lib/collection/src/update_handler.rs
generall: fixed

PR #7193, opened 2025-09-01
Create and load for an appendable quantization

IvanPleshkov:
This PR implements load and create for appendable quantization.

The are 4 main changes here:
1. Quantization config `QuantizedVectorsConfig` was changed. A new field `storage type` was added to mark if quantization is appendable or not.
2. `QuantizedVectors::create` was changed. It takes a new argument `storage_type: QuantizedVectorsStorageType` which is a marker, if a quantization is appendable or not. If yes, this function created new appendable quantization based on `ChunkedMmapVectors`.
3. `QuantizedVectors::create` supports appendable case. If quantization is not present, create a new appendable one.
4. Segment builder does not skip the appendable segment in the `update_quantization` function.

Everything else is an implementation of these changes.

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -775,7 +775,7 @@ impl QuantizationConfig {
         self != other
     }
 
-    pub fn is_appendable(&self) -> bool {
+    pub fn supports_appendable(&self) -> bool {
         matches!(self, QuantizationConfig::Binary(_))
     }
 }
--- a/lib/segment/src/segment_constructor/segment_builder.rs
+++ b/lib/segment/src/segment_constructor/segment_builder.rs
@@ -718,7 +718,7 @@ impl SegmentBuilder {
 
             if let Some(quantization_config) = config.quantization_config(vector_name) {
                 // Don't build quantization for appendable vectors if quantization method does not support it
-                if is_appendable && !quantization_config.is_appendable() {
+                if is_appendable && !quantization_config.supports_appendable() {
                     continue;
                 }
 


lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
```diff
@@ -771,7 +780,10 @@ mod tests {
         let quantization_files = quantized_vectors.files();
 
         // test save-load
-        let quantized_vectors = QuantizedVectors::load(&storage, dir.path()).unwrap();
+        let quantized_vectors =
+            QuantizedVectors::load(&config, &storage, dir.path(), &false.into())
+                .unwrap()
+                .unwrap();
         assert_eq!(files, storage.files());
```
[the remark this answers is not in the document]
IvanPleshkov: Fixed in all places. Here I used `stopped` variable which is defined before this call

lib/segment/src/vector_storage/quantized/quantized_vectors.rs
```diff
@@ -691,13 +867,40 @@ impl QuantizedVectors {
...
+        let quantized_vectors = Self::create(
+            vector_storage,
+            quantization_config,
+            QuantizedVectorsStorageType::Mutable,
+            path,
+            1,
+            stopped,
+        )?;
+        Ok(Some(quantized_vectors))
     }
```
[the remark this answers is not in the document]
IvanPleshkov: Not here, all cases are covered and if appendable quantization is not supported, the error is returned

lib/segment/src/vector_storage/quantized/quantized_chunked_mmap_storage.rs
```diff
@@ -14,6 +14,21 @@ pub struct QuantizedChunkedMmapStorage {
...
+            AdviceSetting::Global
+        };
+        let data = ChunkedMmapVectors::<u8>::open(
+            path,
+            quantized_vector_size,
+            advice,
+            Some(in_ram), // populate
+        )?;
+        Ok(Self { data })
+    }
```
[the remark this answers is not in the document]
IvanPleshkov: `ChunkedMmapVectors` has zero protection. I don't find this check is required here

lib/segment/src/types.rs
```diff
@@ -774,6 +774,10 @@ impl QuantizationConfig {
     pub fn mismatch_requires_rebuild(&self, other: &Self) -> bool {
         self != other
     }
+
+    pub fn is_appendable(&self) -> bool {
```
timvisee: I renamed this one to `supports_appendable()`

lib/segment/src/vector_storage/quantized/quantized_vectors.rs
```diff
@@ -691,13 +867,40 @@ impl QuantizedVectors {
         Ok(quantized_vectors)
     }
 
-    pub fn config_exists(path: &Path) -> bool {
-        path.join(QUANTIZED_CONFIG_PATH).exists()
+    pub fn load(
```
timvisee, line 870: How about we call this `load_or_create`?

One of the main things I find very confusing in our current storage implementations is all the inconsistencies in what does or does not create new data if it doesn't exist.

lib/segment/src/vector_storage/quantized/quantized_vectors.rs
```diff
@@ -790,10 +999,16 @@ impl QuantizedVectors {
         scalar_config: &ScalarQuantizationConfig,
         multivector_config: &MultiVectorConfig,
     ) -> OperationResult<QuantizedVectorStorage> {
+        if !config.storage_type.is_immutable() {
+            return Err(OperationError::service_error(
+                "Mutable quantized multivector storage is not supported for Scalar Quantization",
```
timvisee, line 1004: Is quantization supported for multi vectors in general?

lib/segment/src/types.rs
IvanPleshkov: Thanks!

lib/segment/src/vector_storage/quantized/quantized_vectors.rs
IvanPleshkov: Yes, multivectors does not have any restrictions in quantization, multivectors are supported for SQ, PQ, BQ.
Here is a check that SQ/PQ does not support appendable quantization, appendable quantization is supported for BQ only

timvisee:
When setting [this](https://github.com/qdrant/qdrant/blob/7a21b2e4a61ea8ed6c7e64d28c603938fae76574/lib/segment/src/vector_storage/quantized/quantized_vectors.rs#L49) constant to true. I don't see any appendable quantized storage files appear on disk. Is that correct behavior?

```bash
bfb -n 1024 --indexing-threshold 0 --quantization binary
```

IvanPleshkov:
> When setting [this](https://github.com/qdrant/qdrant/blob/7a21b2e4a61ea8ed6c7e64d28c603938fae76574/lib/segment/src/vector_storage/quantized/quantized_vectors.rs#L49) constant to true. I don't see any appendable quantized storage files appear on disk. Is that correct behavior?
> 
> ```shell
> bfb -n 1024 --indexing-threshold 0 --quantization binary
> ```

If you wish to check manually, please check on https://github.com/qdrant/qdrant/pull/7145
7145 is rebased on this branch and contains all changes including feature flag, set all feature flags and your BFB test will work.

IvanPleshkov:
> When setting [this](https://github.com/qdrant/qdrant/blob/7a21b2e4a61ea8ed6c7e64d28c603938fae76574/lib/segment/src/vector_storage/quantized/quantized_vectors.rs#L49) constant to true. I don't see any appendable quantized storage files appear on disk. Is that correct behavior?
> 
> ```shell
> bfb -n 1024 --indexing-threshold 0 --quantization binary
> ```

Also, you can use docker image
https://github.com/qdrant/qdrant/actions/runs/17376598326

timvisee:
It sounds like the behavior I stated is expected, and the essential part to make that work will be in the next PR. :+1: 

Edit: yes, the current collection is still built without a quantization config in place, so this is expected.

PR #7194, opened 2025-09-01
Fix duplicate ids in upsert_points

agourlay:
Revert incorrect behavior introduced in https://github.com/qdrant/qdrant/pull/7130

There might be duplicated ids.


KShivendu (approved):

agourlay:
Future work.

We need a test capturing our current behavior with duplicated point ids to avoid such silent change in the future.

PR #7195, opened 2025-09-01
HNSW: use new graph format during search

xzfc:
~~Based on #7171, ignore first three commits during the review.~~

This PR:
1. Adds new method `GraphLayers::links_with_vectors` that uses both quantized and non-quantized vectors stored in the file format implemented in #7171.
2. Uses this new method in `HNSWIndex::search_with_graph`.

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -959,57 +959,53 @@ impl HNSWIndex {
             .unwrap_or_else(|| id_tracker.deleted_point_bitslice());
 
         let hw_counter = vector_query_context.hardware_counter();
-        let filter_context = filter.map(|f| payload_index.filter_context(f, &hw_counter));
+        let oversampled_top = get_oversampled_top(quantized_vectors.as_ref(), params, top);
 
-        // Try to use graph with vectors first.
-        let search_result = 'search_result: {
+        let search_with_vectors = || -> OperationResult<Option<Vec<ScoredPointOffset>>> {
             if !self.graph.has_vectors() || !is_quantized_search(quantized_vectors.as_ref(), params)
             {
-                break 'search_result None;
+                return Ok(None);
             }
             let Some(quantized_vectors) = quantized_vectors.as_ref() else {
-                break 'search_result None;
+                return Ok(None);
             };
 
-            let quantized_scorer_filtered = FilteredScorer::new(
+            // Quantized vectors are "link vectors"
+            let link_scorer_filtered = FilteredScorer::new(
                 vector.to_owned(),
                 &vector_storage,
                 Some(quantized_vectors),
-                filter_context
-                    .as_ref()
-                    .map(|x| BoxCow::Borrowed(x.as_ref())),
+                filter.map(|f| BoxCow::Owned(payload_index.filter_context(f, &hw_counter))),
                 deleted_points,
                 vector_query_context.hardware_counter(),
             )?;
-            let Some(quantized_scorer_filtered_bytes) = quantized_scorer_filtered.scorer_bytes()
-            else {
-                break 'search_result None;
+            let Some(link_scorer_filtered_bytes) = link_scorer_filtered.scorer_bytes() else {
+                return Ok(None);
             };
 
-            let full_scorer = new_raw_scorer(
+            // Full vectors are "base vectors"
+            let base_scorer = new_raw_scorer(
                 vector.to_owned(),
                 &vector_storage,
                 vector_query_context.hardware_counter(),
             )?;
-            let Some(full_scorer_bytes) = full_scorer.scorer_bytes() else {
-                break 'search_result None;
+            let Some(base_scorer_bytes) = base_scorer.scorer_bytes() else {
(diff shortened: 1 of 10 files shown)


lib/segment/src/index/hnsw_index/point_scorer.rs
```diff
@@ -301,6 +308,15 @@ pub enum BoxCow<'a, T: ?Sized> {
     Boxed(Box<T>),
 }
 
+impl<'a, T: ?Sized> BoxCow<'a, T> {
```
IvanPleshkov: WDYT about moving `BoxCow` to the common crate?

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -959,37 +959,88 @@ impl HNSWIndex {
...
 
-        let search_result = self.graph.search(
-            oversampled_top,
-            ef,
-            points_scorer,
-            custom_entry_points,
-            &is_stopped,
-        )?;
+        // Try to use graph with vectors first.
+        let search_result = 'search_result: {
```
IvanPleshkov: I propose to use a separate function for search with/without vectors and avoid `'search_result`, I find it as a kind of GO TO and a very controversial trick

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -959,37 +959,88 @@ impl HNSWIndex {
...
+                deleted_points,
+                vector_query_context.hardware_counter(),
+            )?;
+            let Some(quantized_scorer_filtered_bytes) = quantized_scorer_filtered.scorer_bytes()
+            else {
+                break 'search_result None;
+            };
 
-        Ok(res)
+            let full_scorer = new_raw_scorer(
```
IvanPleshkov: Nit. I propose to rename into `base_scorer` because you defined a name "base vector" for the vector in links

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -959,37 +959,88 @@ impl HNSWIndex {
...
+                vector.to_owned(),
+                &vector_storage,
+                vector_query_context.hardware_counter(),
+            )?;
+            let Some(full_scorer_bytes) = full_scorer.scorer_bytes() else {
+                break 'search_result None;
+            };
+
+            Some(self.graph.search_with_vectors(
```
IvanPleshkov: I don't see oversampling here; `top` is provided here directly, and the oversampling parameter is ignored.
I find it critical because oversampling is very important for BQ

IvanPleshkov: I mark PR as "request changes" because of this thread

lib/segment/src/index/hnsw_index/point_scorer.rs
xzfc: Done, along with `SimpleCow`.

lib/segment/src/index/hnsw_index/hnsw.rs
xzfc: Done.

lib/segment/src/index/hnsw_index/hnsw.rs
xzfc: Updated.

It behaves a bit different than oversampling for regular search. In case of regular search, oversampling does two things:
1. `ef = max(ef, oversampling_top)` (IOW, increase `ef` if it is smaller than `oversampling_top`)
2. Search for `oversampling_top` items, then re-score, sort and truncate to `top`.

For the new search implementation (which does implicit rescoring during the search) the second thing makes no sense, so only the first thing is implemented.

lib/segment/src/index/hnsw_index/hnsw.rs
xzfc: Replaced labeled block and breaks with closures.

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -958,38 +959,90 @@ impl HNSWIndex {
...
+                &vector_storage,
+                vector_query_context.hardware_counter(),
+            )?;
+            let Some(base_scorer_bytes) = base_scorer.scorer_bytes() else {
+                return Ok(None);
+            };
+
+            Ok(Some(self.graph.search_with_vectors(
+                top,
+                std::cmp::max(ef, oversampled_top),
```
IvanPleshkov, line 998: Proposal. As you mentioned, the search with the graph is different in the case of oversampling. WDYT about `ef * papams.quantization.oversampling` instead of `max`? It's a completely different behaviour, but it's more predictable because `oversampling=2` is guaranteed to improve acc for customers

xzfc: The proposed behavior is still surprising in sense that sometimes oversampling would act as a multiplier of `top` (hnsw with vecs), and sometimes as a multiplier of `ef` (regular hnsw). Especially since the new feature is planned to be enabled implicitly: there would not be user-facing `hnsw_with_vectors: bool` option, but rather it would be controlled by some heuristic.

PR #7196, opened 2025-09-01
Chain nested locks for proxy segments

agourlay:
Keep the chain of locks on the proxy segment for the integrity of the access.


timvisee:
Nice!

I pushed a commit on this branch to fully remove the `get_cloned` function I introduced in <https://github.com/qdrant/qdrant/pull/7176>.

PR #7203, opened 2025-09-02
Fix scroll lock ordering

agourlay:
Always take the scroll lock and segment locks in the same order. 

The update path takes the scroll lock in the opposite order as the read paths.

https://github.com/qdrant/qdrant/blob/dev/lib/collection/src/collection_manager/collection_updater.rs#L55-L70

Now it is always

1. scroll_lock
2. segments lock
3. segment lock

### DIFF
--- a/lib/collection/src/shards/local_shard/scroll.rs
+++ b/lib/collection/src/shards/local_shard/scroll.rs
@@ -153,8 +153,8 @@ impl LocalShard {
         let stopping_guard = StoppingGuard::new();
         let segments = self.segments.clone();
 
-        let (non_appendable, appendable) = segments.read().split_segments();
         let scroll_lock = self.scroll_read_lock.read().await;
+        let (non_appendable, appendable) = segments.read().split_segments();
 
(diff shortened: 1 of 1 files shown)


timvisee (approved):

agourlay:
For the record, I still get a deadlock on `segments` with this fix.

```rust
    #[allow(clippy::too_many_arguments)]
    pub async fn scroll_by_id(
        &self,
        offset: Option<ExtendedPointId>,
        limit: usize,
        with_payload_interface: &WithPayloadInterface,
        with_vector: &WithVector,
        filter: Option<&Filter>,
        search_runtime_handle: &Handle,
        timeout: Option<Duration>,
    ) -> CollectionResult<Vec<RecordInternal>> {
        let start = Instant::now();
        let timeout = timeout.unwrap_or(self.shared_storage_config.search_timeout);
        let stopping_guard = StoppingGuard::new();

        let scroll_lock = self.scroll_read_lock.read().await;
        let (non_appendable, appendable) = segments.read().split_segments(); <---- stuck here
```

PR #7074, opened 2025-08-18
[WIP] Continous snapshot bug

agourlay:
```
❯ RUST_LOG=info cargo nextest run --all continuous --nocapture
[2025-09-09T15:19:07Z INFO  collection::collection::snapshots] Creating collection snapshot test-0-2025-09-09-15-19-07.snapshot into "/tmp/test_snapshotsH4iD4u/test-0-2025-09-09-15-19-07.snapshot"
[2025-09-09T15:19:07Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-09T15:19:07Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-09T15:19:07Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-09T15:19:07Z WARN  collection::collection_manager::collection_updater] Update operation declined: No point with id 0 found
[2025-09-09T15:19:07Z WARN  collection::shards::replica_set::update] Failed to update shard test:0 on peer 0, error: No point with id 0 found

thread 'continuous_snapshot_test::test_continuous_snapshot' panicked at lib/collection/tests/integration/continuous_snapshot_test.rs:201:31:
points_task error: No point with id 0 found
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[2025-09-09T15:19:07Z ERROR collection::update_handler] Optimization error: Service internal error: IO Error: No such file or directory (os error 2)

thread 'tokio-runtime-worker' panicked at lib/collection/src/update_handler.rs:447:41:
Optimization error: Service internal error: IO Error: No such file or directory (os error 2)
[2025-09-09T15:19:07Z WARN  collection::shards::local_shard] Update workers failed with: Service internal error: task 29 was cancelled
[2025-09-09T15:19:07Z INFO  wal] segment creator shutting down
test continuous_snapshot_test::test_continuous_snapshot ... FAILED

failures:

failures:
    continuous_snapshot_test::test_continuous_snapshot

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 40 filtered out; finished in 1.27s
```

PR #7208, opened 2025-09-03
Unit test with proxy that propagates older delete to wrapped segment

timvisee:
Add a unit test to assert a proxy segment propagating an older delete to a wrapped segment with a newer point is sound. It proves that this situation is possible in specific circumstances.

This eliminates a faulty debug assertion:
https://github.com/qdrant/qdrant/blob/60e0c07b0ab4cbd1eec053542b9db85441b4c4ae/lib/shard/src/proxy_segment/mod.rs#L298-L304

The debug assertion assumed that the logical point data in a wrapped segment of a proxy cannot change. But that is not true. If the wrapped segment is also a proxy, it may share logical state with other segments through it's write segment.

It means the following is required to make this happen:
- there must be two layers of two proxies
- one of the outer proxies is unproxied half way
- a point is updated through the unproxied segment (now being the inner proxy)

We have the same debug assertion in the optimizer [here](https://github.com/qdrant/qdrant/blob/5235d8643552c3106e4e91b522a5b20d5b86b689/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs#L842-L848). But that one is sound, because the optimizer never proxies a proxy, and so we can't have this shared state in a wrapped segment. I've added an extra debug assertion to ensure we don't do that.

Other examples triggering this debug assertion:
- https://github.com/qdrant/qdrant/pull/7074

### All Submissions:

### DIFF
--- a/lib/shard/src/segment_holder/tests.rs
+++ b/lib/shard/src/segment_holder/tests.rs
@@ -677,6 +677,8 @@ fn test_double_proxies() {
 /// - there are at least two layers of two proxies
 /// - one of the outer proxies is unproxied half way
 /// - a new point version is upserted through the unproxied segment (now being the inner proxy)
+///
+/// See: <https://github.com/qdrant/qdrant/pull/7208>
 #[test]
 fn test_proxy_propagate_older_delete_to_wrapped() {
     let hw_counter = HardwareCounterCell::disposable();
--- a/lib/shard/src/segment_holder/tests.rs
+++ b/lib/shard/src/segment_holder/tests.rs
@@ -13,6 +13,7 @@ use tempfile::Builder;
 
 use super::*;
 use crate::fixtures::*;
+use crate::proxy_segment::ProxyDeletedPoint;
 
 #[test]
 fn test_add_and_swap() {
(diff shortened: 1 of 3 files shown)


lib/shard/src/segment_holder/tests.rs
```diff
@@ -665,3 +666,285 @@ fn test_double_proxies() {
...
+    assert_eq!(
+        outer_tmp_segment.get().read().point_version(pid),
+        Some(300),
+        "outer tmp segment has point version 300",
+    );
+    assert_eq!(
+        inner_tmp_segment.get().read().point_version(pid),
+        Some(400),
+        "inner tmp segment has point version 400 from latest upsert",
+    );
```
timvisee, line 876: Interesting quirk: outer proxies still see point version 300, because the shared outer tmp segment has point version 300. We assume a write segment always has the newest point version. But in this case the wrapped segment is newer as the inner tmp segment has point version 400.

It means this comment about the write segment always having newer point data is invalid:
https://github.com/qdrant/qdrant/blob/71a8d96372241bd76bab4d159550cc89f1bedd72/lib/shard/src/proxy_segment/segment_entry.rs#L34

I don't think the point version logic needs adjusting there though. From the segment holder perspective we're still able to see the latest point version. All our retrieval logic prefers the latest version.

PR #7212, opened 2025-09-04
Bump http 0.2.12

agourlay:
We are stuck on an old version of the `http` crate because of our old `tonic` stack.

We can at least update to the latest patch version here.

- https://github.com/hyperium/http/releases/tag/v0.2.10
- https://github.com/hyperium/http/releases/tag/v0.2.11
- https://github.com/hyperium/http/releases/tag/v0.2.12


lib/collection/Cargo.toml
timvisee, line 1: Thanks for these removals too!

timvisee (approved):

PR #7215, opened 2025-09-04
refactor matrix api to use query instead of deprecated search

generall:
`do_core_search_batch`, which was used in search matrix API is going to be deprecated and removed. This PR replaces matrix internals with query API and makes some small fixes along the way

### DIFF
--- a/lib/collection/src/grouping/types.rs
+++ b/lib/collection/src/grouping/types.rs
@@ -3,7 +3,6 @@ use segment::data_types::groups::GroupId;
 use segment::json_path::JsonPath;
 use segment::types::{PointIdType, ScoredPoint};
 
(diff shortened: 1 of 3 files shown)


lib/collection/src/grouping/types.rs
```diff
@@ -56,9 +55,6 @@ pub struct QueryGroupRequest {
 
     /// Limit of groups to return
     pub groups: usize,
-
-    /// Options for specifying how to use the group id to lookup points in another collection
-    pub with_lookup: Option<WithLookup>,
```
generall, line 61: we didn't need `with_lookup` here, it was not used and lookup was resolved before creating this structure

PR #7216, opened 2025-09-05
[WIP] Fix missing API validations

agourlay:
Found a few missing API validations:
- update operations in `update_batch`
- filter in `set_payload` and `delete_payload`
- conditions in `min_should`
- mmr in `query`
- filter in `facet_request`
- filter in `filter_selector`
- quantization_config and strict mode config in `collection_config` 
- inference types

PR #7219, opened 2025-09-05
Fix missing REST API validation for point batch update

agourlay:
The validation is completely missing on the point batch update operations.

The new test fails without the fix.

### DIFF
--- a/tests/openapi/test_batch_update_multivec.py
+++ b/tests/openapi/test_batch_update_multivec.py
@@ -33,6 +33,39 @@ def assert_points(collection_name, points, nonexisting_ids=None, with_vectors=Fa
         points, key=itemgetter('id')
     )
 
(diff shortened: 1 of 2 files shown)


tests/openapi/test_batch_update_multivec.py
```diff
@@ -33,6 +33,39 @@ def assert_points(collection_name, points, nonexisting_ids=None, with_vectors=Fa
...
+                            "vector": {
+                                "image": [1.0, 0.0, 9.0],
+                            },
+                            "payload": {},
+                        },
+                    ],
+                    "update_filter": {
+                        "must": [
+                            {
+                                "key": "empty" # empty field conditions are forbidden
```
agourlay, line 56: Using a malformed filter is an easy test for the missing validation call.

But keep in mind that the entire Update enum in depth is not validated.

PR #7220, opened 2025-09-08
Fix missing REST API validations misc.

agourlay:
Various missing API validations.

### DIFF
--- a/lib/storage/src/content_manager/collection_meta_ops.rs
+++ b/lib/storage/src/content_manager/collection_meta_ops.rs
@@ -245,6 +245,7 @@ pub struct UpdateCollection {
     /// Custom params for Optimizers.  If none - it is left unchanged.
     /// This operation is blocking, it will only proceed once all current optimizations are complete
     #[serde(alias = "optimizer_config")]
+    #[validate(nested)]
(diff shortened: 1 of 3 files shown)


lib/storage/src/content_manager/collection_meta_ops.rs
```diff
@@ -245,6 +245,7 @@ pub struct UpdateCollection {
     /// Custom params for Optimizers.  If none - it is left unchanged.
     /// This operation is blocking, it will only proceed once all current optimizations are complete
     #[serde(alias = "optimizer_config")]
+    #[validate(nested)]
     pub optimizers_config: Option<OptimizersConfigDiff>, // TODO: Allow updates for other configuration params as well
```
timvisee, line 249: We should likely add another validation rule on `OptimizersConfigDiff`.

Candidate:
- `deleted_threshold`: [0.0, 1.0]

agourlay: Good call, it is already present in gRPC :+1: 

I took the liberty then to backport `vacuum_min_vector_number` as well from gRPC.

timvisee (approved):

PR #7221, opened 2025-09-08
Fix missing gRPC api validations points selection

agourlay:
The filter within the `PointsSelector` in gRPC should be validated.


agourlay:
Why no CI? :thinking: 

Edit: Fixed with a rebase :shrug:

timvisee (approved):

PR #7222, opened 2025-09-08
[strict-mode] Add max number of payload index count

agourlay:
Naive implementation fetching the full collection info on each requests.


lib/segment/src/types.rs
```diff
@@ -1026,6 +1026,11 @@ pub struct StrictModeConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[validate(nested)]
     pub sparse_config: Option<StrictModeSparseConfig>,
+
+    /// Max number of payload index in a collection
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[validate(range(min = 1))]
```
JojiiOfficial: Why is 1 the minimum? Wouldn't be 0 also useful to fully prevent creating indices?

lib/api/src/grpc/proto/collections.proto
```diff
@@ -374,6 +374,7 @@ message StrictModeConfig {
   optional StrictModeMultivectorConfig multivector_config = 16;
   optional StrictModeSparseConfig sparse_config = 17;
   optional uint64 max_points_count = 18;
+  optional uint64 max_payload_index_count = 19;
```
timvisee:
```suggestion
  optional uint64 max_payload_index_count = 19; // Max number of payload index in a collection
```

lib/segment/src/types.rs
timvisee: I agree, a minimum of 0 makes sense.

tests/openapi/test_strictmode.py
```diff
@@ -2413,3 +2413,108 @@ def filtered_query():
 
     # finally allowed
     filtered_query().raise_for_status()
+
+def test_strict_mode_payload_index_count(collection_name):
+    set_strict_mode(collection_name, {
+        "enabled": True,
+        "max_payload_index_count": 1,
```
timvisee, line 2455: Let's also test with `0` once we allow it in validation

lib/api/build.rs
```diff
@@ -166,8 +166,15 @@ fn configure_validation(builder: Builder) -> Builder {
             ("UpdateCollectionClusterSetupRequest.operation", ""),
             ("StrictModeConfig.max_query_limit", "range(min = 1)"),
             ("StrictModeConfig.max_timeout", "range(min = 1)"),
+            ("StrictModeConfig.max_points_count", "range(min = 1)"),
             ("StrictModeConfig.read_rate_limit_per_minute", "range(min = 1)"),
             ("StrictModeConfig.write_rate_limit_per_minute", "range(min = 1)"),
```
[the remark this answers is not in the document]
agourlay: good catch :face_exhaling:

PR #7233, opened 2025-09-09
Limit concurrent calls to `load_segment`

xzfc:
# Problem

The `load_segment()` function is executed in parallel per segment.

https://github.com/qdrant/qdrant/blob/8235a76fde322cb7cc2c8fb4030693a2aae8a6df/lib/collection/src/shards/local_shard/mod.rs#L335-L338

Inside `load_segment()`, the migration to from RocksDB is performed (gated by a feature flag for now).

https://github.com/qdrant/qdrant/blob/8235a76fde322cb7cc2c8fb4030693a2aae8a6df/lib/segment/src/segment_constructor/segment_constructor_base.rs#L751-L761

That means with 100 segments we'll get 100 parallel migrations once we enable these feature flags.

# Solution

This PR implements the the following limit:

```rust
/// Maximum number of segments to load concurrently when loading a collection.
pub const MAX_CONCURRENT_SEGMENT_LOADS: usize = 8;
```

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -21,6 +21,7 @@ use arc_swap::ArcSwap;
 use common::budget::ResourceBudget;
 use common::counter::hardware_accumulator::HwMeasurementAcc;
 use common::counter::hardware_counter::HardwareCounterCell;
+use common::defaults::MAX_CONCURRENT_SEGMENT_LOADS;
 use common::rate_limiting::RateLimiter;
 use common::save_on_disk::SaveOnDisk;
 use common::{panic, tar_ext};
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -356,38 +349,43 @@ impl LocalShard {
...
-                .as_ref()
-                .map(|sparse_vectors| {
-                    check_sparse_compatible_with_segment_config(
-                        sparse_vectors,
-                        &segment.config().sparse_vector_data,
-                        true,
-                    )
-                })
-                .unwrap_or(Ok(()))?;
+        while let Some(result) = load_segment_tasks.join_next().await {
```
timvisee: How about we use something like [`buffer_unordered`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffer_unordered) here?

I think that would be the easiest way to limit concurrency, without us having to manually implement it.

PR #7248, opened 2025-09-12
Fix non visible points during snapshots

agourlay:
This PR fixes an issue where points previously inserted/updated are not visible during a snapshot.

During a snapshot, all segments are proxified and share the same write segment for the sake of efficiency.

Each segment is snapshotted and unproxified immediately if possible, meaning the inner wrapped segment is promoted to the segment holder.

As a follow up, after all segments have been unproxified, the shared write segment is promoted to the segment holder if it contains points accumulated during the snapshot process.

An issue appears between the unproxifying of the last segment and the promotion of the shared write segment.

During a short period of time, the content of the write segment is not visible to readers because:
- it is not accessible via any proxy segments
- the write segment itself has not been promoted yet

This problem is visible in the provided test which displays a breakage of the read-my-write consistency during snapshot.

This PR proposes to fix the issue by making sure we keep at least one proxy segment alive to be promoted atomically with the shared write segment.

### DIFF
--- a/lib/shard/src/segment_holder/mod.rs
+++ b/lib/shard/src/segment_holder/mod.rs
@@ -969,7 +969,7 @@ impl SegmentHolder {
         log::trace!("Applying function on all proxied shard segments");
         let mut result = Ok(());
         let mut unproxied_segment_ids = Vec::with_capacity(proxies.len());
-        // Reverse to unproxify first non-appenable segments
+        // Reverse to unproxify first non-appendable segments
         proxies.reverse();
         for (segment_id, proxy_segment) in &proxies {
             // Get segment to snapshot
@@ -1003,8 +1003,8 @@ impl SegmentHolder {
             // Make sure to keep at least one proxy segment to maintain access to the points in the shared write segment.
             // The last proxy and the shared write segment will be promoted into the segment_holder atomically
             // by `Self::unproxy_all_segments` afterwards to maintain the read consistency.
-            let is_last_proxy = unproxied_segment_ids.len() >= proxies.len() - 1;
-            if !is_last_proxy {
+            let remaining = proxies.len() - unproxied_segment_ids.len();
+            if remaining > 1 {
                 match Self::try_unproxy_segment(segments_lock, *segment_id, proxy_segment.clone()) {
                     Ok(lock) => {
                         segments_lock = lock;


lib/shard/src/proxy_segment/mod.rs
```diff
@@ -55,7 +55,10 @@ impl ProxySegment {
                 let already_deleted = raw_segment_guard.get_deleted_points_bitvec();
                 Some(already_deleted)
             }
-            LockedSegment::Proxy(_) => None,
+            LockedSegment::Proxy(_) => {
+                log::debug!("Double proxy segment creation");
```
timvisee, line 59: Shall we demote this to `trace`?

lib/shard/src/segment_holder/mod.rs
```diff
@@ -969,6 +969,8 @@ impl SegmentHolder {
         log::trace!("Applying function on all proxied shard segments");
         let mut result = Ok(());
         let mut unproxied_segment_ids = Vec::with_capacity(proxies.len());
+        // Reverse to unproxify first non-appenable segments
+        proxies.reverse();
```
agourlay, line 973: I do not have a rock solid rational for this change, but it is necessary for the fix below to work properly

lib/shard/src/segment_holder/mod.rs
```diff
@@ -997,13 +999,18 @@ impl SegmentHolder {
...
-            // Unproxying now lets us release the segment earlier, prevent unnecessary writes to the temporary segment
-            match Self::try_unproxy_segment(segments_lock, *segment_id, proxy_segment.clone()) {
-                Ok(lock) => {
-                    segments_lock = lock;
-                    unproxied_segment_ids.push(*segment_id);
+            // Unproxying now lets us release the segment earlier, prevent unnecessary writes to the temporary segment.
+            // Make sure to keep at least one proxy segment to maintain access to the points in the shared write segment.
+            // The last proxy and the shared write segment will be promoted into the segment_holder atomically
+            // by `Self::unproxy_all_segments` afterwards to maintain the read consistency.
+            if unproxied_segment_ids.len() < proxies.len() - 1 {
```
timvisee: Nit, might clarify a bit more to readers:

```suggestion
            let is_last_proxy = unproxied_segment_ids.len() >= proxies.len() - 1;
            if !is_last_proxy {
```

lib/shard/src/proxy_segment/mod.rs
agourlay: I'd like to keep this way because double proxies are mysterious beasts.

Seeing it in the log is a clear sign that we are dealing with a complex situation like snapshot + optimizer.

lib/shard/src/segment_holder/mod.rs
agourlay: I like the idea :+1: 

Why `>=` over a strict equality though?

timvisee: > Why >= over a strict equality though?

Because I don't trust computers

PR #7252, opened 2025-09-15
Fix phrase match to not match with unknown tokens

agourlay:
original report on Discord https://discord.com/channels/907569970500743200/1415930786171064380

The phrase matcher is currently ignoring unknown tokens in the input query.

This behavior creates false positive during document matching.

The proposed solution is to not build `Document` containing unknown tokens.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/tests/mod.rs
+++ b/lib/segment/src/index/field_index/full_text_index/tests/mod.rs
@@ -287,7 +287,7 @@ fn test_phrase_matching() {
 
         // Test valid phrase up to a token that doesn't exist
         let query_with_unknown_token = index.parse_phrase_query("quick brown bird", &hw_counter);
-        // the phrase query is not valid because it contains an unknow token
+        // the phrase query is not valid because it contains an unknown token
         assert!(query_with_unknown_token.is_none());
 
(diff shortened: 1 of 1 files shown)


lib/segment/src/index/field_index/full_text_index/text_index.rs
```diff
@@ -286,7 +286,7 @@ impl FullTextIndex {
         phrase: &str,
         hw_counter: &HardwareCounterCell,
     ) -> Option<ParsedQuery> {
```
agourlay, line 288: interestingly the signature of the function was already setup to return an `Option` and the `rustdoc` even says

> If there are any unseen tokens, returns `None`

I guess this was an oversight :)

generall:
Does this PR fix behavior in provided snapshot?
I saw at least one document which actually contains all tokens. Is it because of multiple segments, so the token is known in one segment but not another?

agourlay:
> Does this PR fix behavior in provided snapshot?

Yes, it is a query time fix.

> I saw at least one document which actually contains all tokens. Is it because of multiple segments, so the token is known in one segment but not another?

Yes exactly, the second segment did not know the token so it started offering partial matches.

PR #7263, opened 2025-09-15
Fix missing flush, flush before sync

generall:
Also checked other places with sync_all, seems to be the only one missing.


lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -336,8 +336,12 @@ impl ImmutableIdTracker {
...
         // Write mappings to disk.
         let file = File::create(Self::mappings_file_path(path))?;
-        let mut writer = BufWriter::new(&file);
+        let mut writer = BufWriter::new(file);
         Self::store_mapping(&mappings, &mut writer)?;
+
+        // Explicitly fsync file contents to ensure durability
+        writer.flush()?;
+        let file = writer.into_inner().unwrap();
         file.sync_all()?;
```
timvisee, line 345: Key change is here: first flush, then `sync_all`.

PR #7267, opened 2025-09-17
Fix deadlock on Actix RT in Cluster Info

agourlay:
Follow up on https://github.com/qdrant/qdrant/pull/7265

When the Actix runtime runs on a single thread, it can be starved by blocking the running thread on a sync lock.

This happens when calling the cluster info API while:
- have long running streaming snapshot going out holding read lock
- a segment lock (not segment holder lock)
- a pending write to the same segment
- request cluster info that blocks actix runtime with blocking read lock

I was able to reproduce the issue by adapting the existing deadlock test by interleaving calls to the cluster info API.

```
Thread 8 (Thread 0x7b408ebff6c0 (LWP 313205) "actix-rt|system"):
#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
#1  0x000060f10badea19 in parking_lot_core::thread_parker::imp::ThreadParker::futex_wait () at /home/agourlay/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parking_lot_core-0.9.11/src/thread_parker/linux.rs:112
#2  0x000060f10bade84c in <parking_lot_core::thread_parker::imp::ThreadParker as parking_lot_core::thread_parker::ThreadParkerT>::park () at /home/agourlay/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parking_lot_core-0.9.11/src/thread_parker/linux.rs:66
#3  0x000060f10badaf2d in parking_lot_core::parking_lot::park::{{closure}} () at /home/agourlay/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parking_lot_core-0.9.11/src/parking_lot.rs:635
[...]

PR #7241, opened 2025-09-10
Fix deadlock during streaming shard snapshot

ffuugoo:
This PR:
- refactors telemetry methods on `LocalShard` to be `async` and use `spawn_blocking`, so that telemetry API won't block Actix runtime with a blocking read-lock call
- adds a test for this specific deadlock we've seen in chaos-testing 

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your core changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #7269, opened 2025-09-18
Always run Telemetry on a blocking thread

agourlay:
The collection telemetry acquires several synchronous locks under the hood.

e.g:
- https://github.com/qdrant/qdrant/blob/4012ea9e517ffa5ec9ae0a48486376e3b627c9ae/lib/collection/src/shards/shard_holder/mod.rs#L465
- https://github.com/qdrant/qdrant/blob/b01b36bb86fec22fc40c6b8973110c0cbe21cb13/lib/collection/src/shards/transfer/transfer_tasks_pool.rs#L83
- https://github.com/qdrant/qdrant/blob/a3ce7165900aa1dfc6a816a9bf4b43f9b8eba1c9/lib/collection/src/shards/replica_set/telemetry.rs#L31
- https://github.com/qdrant/qdrant/blob/4012ea9e517ffa5ec9ae0a48486376e3b627c9ae/lib/collection/src/shards/shard_holder/mod.rs#L492

For context, we did make an effort to encapsulate such usage in dedicated blocking tasks in the past:
- https://github.com/qdrant/qdrant/pull/7241
- https://github.com/qdrant/qdrant/pull/7267
- https://github.com/qdrant/qdrant/pull/7265

But the number of locks is difficult to track and the call sites not necessarily trivial to migrate.

This PR proposes to tackle the issue higher in the stack and run the whole collection telemetry on a dedicated thread.

Hopefully this will be enough to not starve the Actix RT when used in the `telemetry` and the `metrics` endpoints.

### DIFF
--- a/tests/consensus_tests/test_shard_snapshot_deadlock.py
+++ b/tests/consensus_tests/test_shard_snapshot_deadlock.py
@@ -3,8 +3,8 @@ import pathlib
 import requests
 import time
 
-from .assertions import *
-from .fixtures import *
+from .assertions import assert_http_ok
+from .fixtures import upsert_random_points, create_collection
 from .utils import *
 
 
@@ -80,3 +80,5 @@ def test_shard_snapshot_deadlock(tmp_path: pathlib.Path):
         # uncomment `sleep` and remove `raise` to investigate the deadlock with gdb
         #time.sleep(100_000) 
         raise
+    finally:
+        executor.shutdown(wait=False)


src/common/telemetry.rs
timvisee, line 1: The

```rust
&self.actix_telemetry_collector.lock(),
&self.tonic_telemetry_collector.lock(),
```

also peak my interest. Though I haven't seen these causing issues yet.

timvisee (approved):

src/common/telemetry.rs
agourlay: Yes, I will definitely look for those in potential future traces.

PR #7270, opened 2025-09-18
Add trigger reason to optimizer telemetry

JojiiOfficial:
This PR adds a new field, `trigger_reason`, to the optimizer logs in telemetry.
In a future PR, this data will be used to calculate statistics on how often optimizers are triggered and for what reasons, which will be exposed in our `/metrics` API.

```json
{
  "name": "indexing",
  "segment_ids": [13],
  "status": "done",
  "start_at": "2025-09-18T09:44:20.107724473Z",
  "end_at": "2025-09-18T09:44:32.481401239Z",
  "trigger_reason": "to_index"
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ default-run = "qdrant"
 workspace = true
 
 [features]
-default = ["rocksdb"]
+default = []
 service_debug = ["parking_lot/deadlock_detection"]
 tracing = [
     "api/tracing",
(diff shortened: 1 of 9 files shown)


generall (changes requested):
We already have `name` of the optimizer. I don't think introducing we entity for this is justified

JojiiOfficial:
We decided to do it a less complex way - closing.

PR #7272, opened 2025-09-18
parallel scroll

generall:
Current implementation of scroll works in 2 stages:

1. It reads point ids from all segments in parallel and merges them into final list of IDs
2. It retrieves payload & vectors from segments according to resolved list. 

The second stage, however, is done sequntially.

This PR changes the second stage to parallel read. In theory it should better utilize IOPS, especially in case of large limits (100+). And improve shard transfer performance.

WARN: 

The current version tries to read points in non-appendable -> appendable sequence. This PR removes it, as it looks like this pattern is not required:

1. It doesn't actually prevent what it was designed to prevent
2. We have another mechanism, which guarantees that there will be no updates during scroll requests - `scroll_lock`, as we rely on scroll operation for many internal processes

Main changes are in this PR: https://github.com/qdrant/qdrant/commit/d2910df472ec03b6c72f9f694d4412dc10cbca7e

### DIFF
--- a/lib/collection/src/collection_manager/segments_retriever.rs
+++ b/lib/collection/src/collection_manager/segments_retriever.rs
@@ -1,0 +1,185 @@
+use std::collections::BTreeSet;
+use std::collections::hash_map::Entry;
+use std::sync::atomic::AtomicBool;
+
+use ahash::AHashMap;
+use common::counter::hardware_accumulator::HwMeasurementAcc;
+use segment::common::operation_error::OperationError;
+use segment::data_types::named_vectors::NamedVectors;
+use segment::data_types::vectors::VectorStructInternal;
+use segment::types::{Filter, PointIdType, SeqNumberType, WithPayload, WithVector};
+use shard::segment_holder::LockedSegmentHolder;
+use tokio::runtime::Handle;
+
+use crate::common::stopping_guard::StoppingGuard;
+use crate::operations::types::{CollectionResult, RecordInternal};
+
+#[derive(Default)]
+pub struct SegmentsRetriever;
+
+impl SegmentsRetriever {
+    /// Retrieve records for the given points ids from the segments
+    /// - if payload is enabled, payload will be fetched
+    /// - if vector is enabled, vector will be fetched
+    ///
+    /// The points ids can contain duplicates, the records will be fetched only once
+    ///
+    /// If an id is not found in the segments, it won't be included in the output.
+    pub async fn retrieve(
+        segments: LockedSegmentHolder,
+        points: &[PointIdType],
+        with_payload: &WithPayload,
+        with_vector: &WithVector,
+        runtime_handle: &Handle,
+        hw_measurement_acc: HwMeasurementAcc,
+    ) -> CollectionResult<AHashMap<PointIdType, RecordInternal>> {
+        let stopping_guard = StoppingGuard::new();
+        runtime_handle
+            .spawn_blocking({
+                let segments = segments.clone();
+                let points = points.to_vec();
+                let with_payload = with_payload.clone();
+                let with_vector = with_vector.clone();
+                let is_stopped = stopping_guard.get_is_stopped();
+                // TODO create one Task per segment level retrieve
+                move || {
+                    Self::retrieve_blocking(
+                        segments,
+                        &points,
+                        &with_payload,
+                        &with_vector,
+                        &is_stopped,
+                        hw_measurement_acc,
+                    )
+                }
+            })
+            .await?
+    }
+
+    pub fn retrieve_blocking(
+        segments: LockedSegmentHolder,
+        points: &[PointIdType],
+        with_payload: &WithPayload,
+        with_vector: &WithVector,
+        is_stopped: &AtomicBool,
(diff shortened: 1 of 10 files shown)


lib/collection/src/collection_manager/segments_retriever.rs
```diff
@@ -0,0 +1,229 @@
...
+                            None,
+                            filter.as_ref(),
+                            &is_stopped,
+                            &hw_counter,
+                        )
+                    })
+                    .collect();
+                Ok(all_points)
+            })
+            .await?
```
timvisee, line 193: Any specific reason for not applying the parallelism here?

timvisee (approved):

lib/collection/src/collection_manager/segments_retriever.rs
```diff
@@ -0,0 +1,229 @@
...
+                                &points,
+                                &with_payload,
+                                &with_vector,
+                                &is_stopped,
+                                hw_counter,
+                            )
+                        }
+                    })
+                })
+                .collect()
```
JojiiOfficial, line 73: `FuturesUnordered` implements `FromIterator` which means we can collect directly into it and don't need the intermediate Vec. 

```suggestion
                .collect::<FuturesUnordered>()
```

lib/collection/src/collection_manager/segments_retriever.rs
```diff
@@ -0,0 +1,229 @@
...
+                match existing {
+                    Entry::Vacant(v) => {
+                        v.insert(record);
+                    }
+                    Entry::Occupied(mut occupied) => {
+                        // If we already have the latest point version, keep that and continue
+                        if occupied.get().version < record.version {
+                            occupied.insert(record);
+                        }
+                    }
```
JojiiOfficial, line 91: Suggestion to reduce the codes length. Totally fine to keep as it is.

```suggestion
                versioned_results.entry(point_id).and_modify(|occupied| {
                    if occupied.get().version < record {
                        *occupied = record;
                    }
                })
                .or_insert(record);

```

lib/collection/src/collection_manager/segments_retriever.rs
```diff
@@ -0,0 +1,229 @@
...
+        hw_counter: HardwareCounterCell,
+    ) -> CollectionResult<AHashMap<PointIdType, RecordInternalVersioned>> {
+        let mut point_records: AHashMap<PointIdType, RecordInternalVersioned> = Default::default();
+
+        let segment_arc = segment.get();
+        let read_segment = segment_arc.read();
+
+        let points = points
+            .iter()
+            .cloned()
```
JojiiOfficial: Makes it more clear that this is cheap
```suggestion
            .copied()
```

lib/collection/src/collection_manager/segments_retriever.rs
generall: there are none, but I don't yet know a good benchmark for this one. Propose to do separatelly

timvisee (changes requested):
Unfortunately I'm seeing suboptimal results with a simple bfb benchmark.

```bash
# setup
bfb -n 10000000 --segments 24 --shards 1

# scrolls
bfb -n 100000 --skip-setup --scroll
```

Before this PR:

```
 [00:00:14] [6,806.711/s] 100000/100000 (eta:0s)
--- Request timings ---
Min request time: 0.00014834
Avg request time: 0.000291846136679996
Median request time: 0.00026391
p95 request time: 0.00043275
p99 request time: 0.00048837
Max request time: 0.001393619
--- Server timings ---
Min server time: 0.00005385
Avg server time: 0.00010830880848999967
Median server time: 0.0001026
p95 server time: 0.000160139
p99 server time: 0.00018172
Max server time: 0.00061532
--- RPS ---
Min rps: 2676.0441112407584
Avg rps: 6877.770043344811
Median rps: 6797.213069898325
Max rps: 8314.010761269363
```

After this PR:

```
 [00:00:19] [5,140.954/s] 100000/100000 (eta:0s)
--- Request timings ---
Min request time: 0.00015164
Avg request time: 0.0003866557055800005
Median request time: 0.00038128
p95 request time: 0.000592659
p99 request time: 0.00064038
Max request time: 0.001312089
--- Server timings ---
Min server time: 0.00004492
Avg server time: 0.00012108577258000077
Median server time: 0.00012018
p95 server time: 0.00017441
p99 server time: 0.00019616
Max server time: 0.00056245
--- RPS ---
Min rps: 2858.924200027366
Avg rps: 4922.8886732117535
Median rps: 4886.3107039699335
Max rps: 5176.297752068113
```

JojiiOfficial (approved):

lib/collection/src/collection_manager/segments_retriever.rs
generall: this doesn't work, `record` has to be cloned.

timvisee: :point_up: tried exactly the same locally :smile: 

Let's keep the match because the record is not cheap to clone :+1:

generall (commented):
review fixes applied

generall:
Benchmarks are not conclusive, drafting for now

PR #7275, opened 2025-09-19
Add more Optimizer data in metrics

JojiiOfficial:
Adds 2 new metrics, `collections_optimizer_trigger_count` and `optimizer_total_processes_running`, to the metrics API. The first one measures the amount of triggers of our optimizers on per (local) collection level.
The second one measures the total amount of optimization processes running on all local-shards of the given node.

```
# HELP collections_optimizer_run_count number of optimization triggers per optimizer
# TYPE collections_optimizer_run_count counter
collections_optimizer_run_count{id="benchmark",optimizer="merge"} 1
collections_optimizer_run_count{id="benchmark",optimizer="index"} 4
```

and 

```
# HELP optimizer_total_processes_running number of optimization processes running in total
# TYPE optimizer_total_processes_running gauge
optimizer_total_processes_running 1
````

### DIFF
--- a/lib/collection/src/shards/local_shard/telemetry.rs
+++ b/lib/collection/src/shards/local_shard/telemetry.rs
@@ -70,7 +70,7 @@ impl LocalShard {
             optimizations: OptimizerTelemetry {
                 status,
                 optimizations,
-                log: (detail.level >= DetailsLevel::Level4)
+                log: (detail.level >= DetailsLevel::Level3)
                     .then(|| self.optimizers_log.lock().to_telemetry()),
             },
             async_scorer: Some(get_async_scorer()),
--- a/lib/collection/src/shards/local_shard/telemetry.rs
+++ b/lib/collection/src/shards/local_shard/telemetry.rs
@@ -46,10 +46,10 @@ impl LocalShard {
             })
             .fold(Default::default(), |total, stats| total + stats);
 
-        let optimizer_triggers = self
+        let optimizer_runs = self
             .optimizers
             .iter()
-            .map(|optimizer| *optimizer.get_telemetry_counter().triggers.lock())
+            .map(|optimizer| *optimizer.get_telemetry_counter().run_counters.lock())
             .fold(Default::default(), |total, stats| total + stats);
 
         let status = self.get_optimization_status().await;
@@ -79,7 +79,7 @@ impl LocalShard {
                 optimizations,
                 log: (detail.level >= DetailsLevel::Level4 || detail.optimizer_logs)
                     .then(|| self.optimizers_log.lock().to_telemetry()),
-                triggers: optimizer_triggers,
+                runs: optimizer_runs,
             },
             async_scorer: Some(get_async_scorer()),
         }
--- a/lib/collection/src/shards/local_shard/telemetry.rs
+++ b/lib/collection/src/shards/local_shard/telemetry.rs
@@ -40,9 +40,16 @@ impl LocalShard {
             .map(|optimizer| {
                 optimizer
                     .get_telemetry_counter()
+                    .durations_aggregator
                     .lock()
                     .get_statistics(detail)
             })
+            .fold(Default::default(), |total, stats| total + stats);
+
+        let optimizer_triggers = self
+            .optimizers
+            .iter()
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: lib/collection/src/shards/local_shard/telemetry.rs [hunk 1]
```diff
@@ -70,7 +70,7 @@ impl LocalShard {
             optimizations: OptimizerTelemetry {
                 status,
                 optimizations,
-                log: (detail.level >= DetailsLevel::Level4)
+                log: (detail.level >= DetailsLevel::Level3)
```
JojiiOfficial: In order to make the logs available in /metrics, I had to change the DetailsLevel from 4 to 3, since this is the level metrics are generated from: https://github.com/qdrant/qdrant/blob/dev/src/actix/api/service_api.rs#L82

Alternatively we could generate metrics from telemetry at Level4.

JojiiOfficial: With the new approach we don't need to touch this!

lib/collection/src/update_handler.rs
```diff
@@ -355,6 +355,9 @@ impl UpdateHandler {
                     &nonoptimal_segment_ids
                 );
 
+                // Optimizer has been triggered, so we need to increment the trigger-counter.
+                optimizer.increment_run_counter();
```
timvisee: Though it is triggered, it may not actually run now because of the CPU/IO budget code below. It is possible to hit this function a lot of times without actually doing any optimization work.

Maybe it's better to move this further down to where we actually spawn the optimization task.

src/common/metrics.rs
```diff
@@ -174,6 +176,68 @@ impl MetricsProvider for CollectionsTelemetry {
...
+                        value as f64,
+                        &[("id", &collection.id), ("optimizer", name)],
+                    ));
+                }
+            };
+
+            add_optimizer_metric("config-mismatch", optimizer_triggers.config_mismatch);
+            add_optimizer_metric("index", optimizer_triggers.index);
+            add_optimizer_metric("merge", optimizer_triggers.merge);
+            add_optimizer_metric("vacuum", optimizer_triggers.vacuum);
```
timvisee: Hit: let's destruct the struct here, so that we don't forget any new fields we might add in the future.

timvisee (changes requested):

lib/collection/src/update_handler.rs
JojiiOfficial: This value is populated as optimization "triggers" in API, so I added the increment here, even though we might not start executing the optimization right away.
I agree the naming here should be improved to make this more clear.

I'd suggest to rename the functions. Alternatively, if we want to count the executions of optimizers (instead of triggers), I'd suggest to rename it in the API together with moving this further down.

What do you think?

timvisee: It may happen that we 'trigger' multiple thousands of times.

Do you think it makes sense to trigger that? I don't see a clear use case for it to be honest.

I am more interested in the total number of started and finished optimizations. Or started and running.

JojiiOfficial: I see! I'll rename it and move the counter increment down the function 👍🏻

timvisee: cc @generall, thoughts?

generall: I am mostly interested in total number of actually running optimizations at the current moment.

generall (changes requested):
Could you please make a separate PR which only introduces `optimizer_total_processes_running`

There are a lot of code changes for `collections_optimizer_run_count` which I am not sure that it worth it

JojiiOfficial:
Done: https://github.com/qdrant/qdrant/pull/7316

timvisee:
> There are a lot of code changes for collections_optimizer_run_count which I am not sure that it worth it

@generall wouldn't it be an interesting metric to detect optimizer loops, or just very hot optimizer runs? While they aren't common, we have seen it happen once every so often.

We cannot detect those with just the current running count. We might see it through CPU/IO usage, but that isn't very direct.

We might simplify it by not partitioning per optimizer type.

JojiiOfficial:
Closing in favor of #7316

PR #7291, opened 2025-09-22
Implement logging of slow requests for scroll, facet, count, and retrieve

generall:


lib/collection/src/operations/loggable.rs
```diff
@@ -47,3 +50,73 @@ impl Loggable for Vec<ShardQueryRequest> {
         hasher.finish()
     }
 }
+
+pub struct ScrollRequestLoggable(pub Value);
```
timvisee: Why do we have a custom wrapping type here, not like the others? Is it just because we don't have a proper type for it?

Maybe it's worth to implement this - a type with all the shared parameters for the scroll variants. That way we also prevent forgetting to log/hash new parameters in the future.

lib/collection/src/operations/loggable.rs
```diff
@@ -47,3 +50,73 @@ impl Loggable for Vec<ShardQueryRequest> {
...
+        "facet"
+    }
+
+    fn request_hash(&self) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        self.request_name().hash(&mut hasher);
+        self.as_ref().hash(&mut hasher);
+        hasher.finish()
+    }
+}
```
timvisee: I vote to implement on `FacetParams` instead, as it allows wider usage.

We could do it together with an auto trait on `Arc<Loggable>` like this:


```suggestion

impl<L> Loggable for Arc<L>
where
    for<'a> &'a L: Loggable,
{
    #[inline]
    fn to_log_value(&self) -> Value {
        self.as_ref().to_log_value()
    }

    #[inline]
    fn request_name(&self) -> &'static str {
        self.as_ref().request_name()
    }

    #[inline]
    fn request_hash(&self) -> u64 {
        self.as_ref().request_hash()
    }
}

impl Loggable for &FacetParams {
    fn to_log_value(&self) -> Value {
        serde_json::to_value(self).unwrap_or_default()
    }

    fn request_name(&self) -> &'static str {
        "facet"
    }

    fn request_hash(&self) -> u64 {
        let mut hasher = DefaultHasher::new();
        self.request_name().hash(&mut hasher);
        self.hash(&mut hasher);
        hasher.finish()
    }
}
```

lib/collection/src/operations/loggable.rs
```diff
@@ -47,3 +50,73 @@ impl Loggable for Vec<ShardQueryRequest> {
...
+        "retrieve"
+    }
+
+    fn request_hash(&self) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        self.request_name().hash(&mut hasher);
+        self.as_ref().hash(&mut hasher);
+        hasher.finish()
+    }
+}
```
timvisee: If we apply the change above, we can simplify these too:


```suggestion
impl Loggable for &CountRequestInternal {
    fn to_log_value(&self) -> Value {
        serde_json::to_value(self).unwrap_or_default()
    }

    fn request_name(&self) -> &'static str {
        "count"
    }

    fn request_hash(&self) -> u64 {
        let mut hasher = DefaultHasher::new();
        self.request_name().hash(&mut hasher);
        self.hash(&mut hasher);
        hasher.finish()
    }
}

impl Loggable for &PointRequestInternal {
    fn to_log_value(&self) -> Value {
        serde_json::to_value(self).unwrap_or_default()
    }

    fn request_name(&self) -> &'static str {
        "retrieve"
    }

    fn request_hash(&self) -> u64 {
        let mut hasher = DefaultHasher::new();
        self.request_name().hash(&mut hasher);
        self.hash(&mut hasher);
        hasher.finish()
    }
}
```

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -313,13 +336,27 @@ impl ShardOperation for LocalShard {
...
+            self.approx_facet(
+                request.clone(),
+                search_runtime_handle,
+                timeout,
+                hw_measurement_acc,
+            )
+            .await?
         };
+        let elapsed = start_time.elapsed();
+        log_request_to_collector(&self.collection_name, elapsed, || request.clone());
```
timvisee: No need to clone here:

```suggestion
        log_request_to_collector(&self.collection_name, elapsed, || request);
```

lib/collection/src/operations/loggable.rs
generall: yes, the source function operates with individual fields. Agreed, that it is not that fancy

Huarong:
Do we need a config to enable slow this request feature?

timvisee:
> Do we need a config to enable slow this request feature?

No. This is enabled by default since [Qdrant 1.16.0](https://github.com/qdrant/qdrant/releases/tag/v1.16.0). See `GET /profiler/slow_requests`.

Huarong:
@timvisee Thanks the great work. Maybe we should update the [API documents](https://api.qdrant.tech/api-reference)

generall:
> Maybe we should update the [API documents](https://api.qdrant.tech/api-reference)

At this point we don't want to document this API, as it might change or even be removed.

PR #7292, opened 2025-09-23
log requests by type

generall:
- **store slow request log per request type**
- **filter by request name in API**

### DIFF
--- a/lib/collection/src/profiling/slow_requests_log.rs
+++ b/lib/collection/src/profiling/slow_requests_log.rs
@@ -17,7 +17,7 @@ pub struct LogEntry {
     #[serde(serialize_with = "duration_as_seconds")]
     duration: Duration,
     datetime: DateTime<Utc>,
-    request_name: String,
+    request_name: &'static str,
     approx_count: usize,
     request_body: serde_json::Value,
     /// Used for fast comparison and lookup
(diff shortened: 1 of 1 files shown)


lib/collection/src/profiling/slow_requests_log.rs
```diff
@@ -69,21 +70,28 @@ impl Ord for LogEntry {
 }
 
 pub struct SlowRequestsLog {
-    log_priority_queue: FixedLengthPriorityQueue<LogEntry>,
+    log_priority_queue: AHashMap<String, FixedLengthPriorityQueue<LogEntry>>,
```
timvisee: Aren't our request names a static string reference? Let's use that as hash map key.

lib/collection/src/profiling/slow_requests_log.rs
```diff
@@ -130,7 +137,22 @@ impl SlowRequestsLog {
...
+            // No entries yet, just insert
+            let entry = LogEntry::new(
+                collection_name.to_string(),
+                duration,
+                datetime,
+                request.request_name().to_string(),
+                request.to_log_value(),
+                content_hash,
+            );
+            return self.try_insert_dedup(entry);
```
timvisee: Why not use the entry API here?

lib/collection/src/profiling/slow_requests_log.rs
```diff
@@ -130,7 +137,22 @@ impl SlowRequestsLog {
...
+        };
```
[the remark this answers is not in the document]
generall: Number of requests is limited by our APIs, it can't be more that this amount of different requests.

PR #7293, opened 2025-09-23
Add CollectionInfo::warnings field

xzfc:
This PR adds `CollectionInfo::configuration_status` that contains a warning when collection configuration is not consistent.  Particularly it checks requirements for the `hnsw_config.copy_vectors` option (#7232): quantization is enabled; multivectors are disabled.

Also, this PR performs a refactoring change in `trait DiffConfig`: it swaps `DiffConfig::update` arguments (as I find `config.update(diff)` to be more natural than `diff.update(&config)`) and adds `DiffConfig::update_opt` helper. More changes in `DiffConfig` in a subsequent PR.

### DIFF
--- a/lib/api/src/grpc/proto/collections.proto
+++ b/lib/api/src/grpc/proto/collections.proto
@@ -162,9 +162,8 @@ message OptimizerStatus {
   string error = 2;
 }
 
-message ConfigurationStatus {
-  bool ok = 1;
-  string warning = 2;
+message CollectionWarning {
+  string message = 1;
 }
 
 message HnswConfigDiff {
@@ -577,7 +576,7 @@ message CollectionInfo {
   map<string, PayloadSchemaInfo> payload_schema = 8; // Collection data types
   optional uint64 points_count = 9; // Approximate number of points in the collection
   optional uint64 indexed_vectors_count = 10; // Approximate number of indexed vectors in the collection.
-  ConfigurationStatus configuration_status = 11; // status of configuration inconsistencies
+  repeated CollectionWarning warnings = 11; // Warnings related to the collection
 }
 
 message ChangeAliases {
--- a/lib/collection/src/config.rs
+++ b/lib/collection/src/config.rs
@@ -1,5 +1,4 @@
 use std::collections::{BTreeMap, HashMap, HashSet};
-use std::fmt::Write as _;
 use std::io::{Read, Write as _};
 use std::num::{NonZeroU32, NonZeroUsize};
 use std::path::Path;
@@ -23,8 +22,8 @@ use wal::WalOptions;
 
 use crate::operations::config_diff::{DiffConfig, QuantizationConfigDiff};
 use crate::operations::types::{
-    CollectionError, CollectionResult, ConfigurationStatus, SparseVectorParams,
-    SparseVectorsConfig, VectorParams, VectorParamsDiff, VectorsConfig, VectorsConfigDiff,
+    CollectionError, CollectionResult, CollectionWarning, SparseVectorParams, SparseVectorsConfig,
+    VectorParams, VectorParamsDiff, VectorsConfig, VectorsConfigDiff,
 };
 use crate::operations::validation;
 use crate::optimizers_builder::OptimizersConfig;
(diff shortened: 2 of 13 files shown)


lib/api/src/grpc/proto/collections.proto
```diff
@@ -162,6 +162,11 @@ message OptimizerStatus {
   string error = 2;
 }
 
+message ConfigurationStatus {
+  bool ok = 1;
+  string warning = 2;
+}
```
timvisee: Nit: how about we rename `warning` to `message`? I'd argue `ok` already defines whether we are ok/warning. And, make the message optional?

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -617,6 +617,7 @@ impl HNSWIndex {
             .unwrap_or_default()
             .then(|| {
                 // NOTE: the configuration is silently ignored if try_new fails.
+                // TODO: add comment
```
timvisee: What is this one about?

lib/collection/src/config.rs
```diff
@@ -276,6 +277,44 @@ impl CollectionConfigInternal {
...
+                        "The `hnsw_config.copy_vectors` option for vector '{vector_name}' \
+                         is not compatible with multivectors. This option will be ignored."
+                    )
+                    .unwrap();
+                }
+            }
+        }
+
+        if !messages.is_empty() {
+            messages.pop(); // trim the last newline
```
timvisee: Why not push each individual message in a `Vec`, and use `messages.join("\n")`. Popping the new line coming from the last write is a bit of a hack to me.

generall:
What about making `configuration_status` more general, so we could also reuse it for other warnings, if needed?

For example, calling it `warnings` and making it type `Option<Vec<String>>` to not limit ourselves with API definitions we won't be able to change later

lib/segment/src/index/hnsw_index/hnsw.rs
xzfc: Forgot to remove, sorry.

xzfc:
Updated. Now `CollectionInfo` includes `warnings: Vec<CollectionWarning>` field.
For now, `CollectionWarning` has only one field: `message: String`.

lib/collection/src/operations/types.rs
```diff
@@ -127,6 +127,16 @@ pub enum OptimizersStatus {
     Error(String),
 }
 
+#[derive(
+    Debug, Default, Serialize, JsonSchema, Anonymize, PartialEq, Eq, PartialOrd, Ord, Clone,
+)]
+#[serde(rename_all = "snake_case")]
+pub struct CollectionWarning {
+    /// Warning message
+    #[anonymize(false)]
```
generall: Message can contain names of the vectors, I would rather anonymise it.

lib/collection/src/operations/types.rs
```diff
@@ -219,6 +229,9 @@ pub struct CollectionInfo {
     pub status: CollectionStatus,
     /// Status of optimizers
     pub optimizer_status: OptimizersStatus,
+    /// Warnings related to the collection
+    #[serde(default)]
```
generall: let's skip serialization if there are no warning. Otherwise it might be unnecessary breaking change for old clients

xzfc: Why? Do clients break on unknown/new fields?

generall: > Do clients break on unknown/new fields?

**maybe**. Some old versions

xzfc: Added skip, tho I believe this approach is a fragile workaround.

---

- It seems that the python client `≤v1.7.3` will crash producing this error message:
  ```
    File "pydantic/main.py", line 347, in pydantic.main.BaseModel.__init__
  pydantic.error_wrappers.ValidationError: 4 validation errors for ParsingModel[InlineResponse2005] (for parse_as_type)
  obj -> result -> config -> optimizer_config -> max_optimization_threads
    none is not an allowed value (type=type_error.none.not_allowed)
  obj -> result -> config -> wal_config -> wal_retain_closed
    extra fields not permitted (type=value_error.extra)
  obj -> result -> config -> strict_mode_config
    extra fields not permitted (type=value_error.extra)
  obj -> result -> warnings
    extra fields not permitted (type=value_error.extra)
  ```
- Although it will still crash regardless of this particular new filed.
- Doesn't crash starting from `≥v1.8.0`. (properly ignores unknown fields)
- I didn't check other language clients.

PR #7294, opened 2025-09-23
DiffConfig: replace serde-based impl with explicit impl

xzfc:
This PR replaces serde-based `DiffConfig` implementation with explicit implementation. Depends on #7293.

# Issue

In `dev`, we have this type coercion that copies fields from `*ConfigDiff` structs into `*Config` structs:
```rust
/// Hacky way to update configuration structures with diff-updates.
/// Intended to only be used in non critical for speed places.
/// TODO: replace with proc macro
pub fn update_config<T: DeserializeOwned + Serialize, Y: DeserializeOwned + Serialize + Merge>(
    config: &T,
    update: Y,
) -> CollectionResult<T> {
    let mut config_values = serde_json::to_value(config)?;
    let diff_values = serde_json::to_value(&update)?;
    merge_level_0(&mut config_values, diff_values);
    let res = serde_json::from_value(config_values)?;
    Ok(res)
}
```

It's hacky indeed. It's error-prone: missing/wrong field names not caught by a compiler). And it's makes the code hard to explore: rust-analyzer "find usages" feature will newer help you find out that `HnswConfig::copy_vectors` can be populated through `HnswConfigDiff::copy_vectors`.

# Solution

This PR replaces this coercion-via-json with a few hand-rolled implementations of `impl DiffConfig` and `impl From`. Additionally, `DiffConfig` trait methods are infallible now (as there no JSON serialization/serialization errors possible).

# Discarded approaches

- Q: What about generating `*ConfigDiff` from `*Config` definition?
  A: I decided to keep them explicitly hand-rolled as these structures have some subtle differences, e.g., see the following diff. While it's possible encode these differences via proc macro and attributes, I find it less maintainable. And probably `macro_rules` give better LSP experience than proc macros.

  ```diff
  -pub struct OptimizersConfig {
  +pub struct OptimizersConfigDiff {
       …
       /// Max number of threads (jobs) for running optimizations per shard.
       /// Note: each optimization job will also use `max_indexing_threads` threads by itself for index building.
  -    /// If "auto" - have no limit and choose dynamically to saturate CPU.
  +    /// If null - have no limit and choose dynamically to saturate CPU.
       /// If 0 - no optimization threads, optimizations will be disabled.
       #[serde(default)]
  -    pub max_optimization_threads: Option<usize>,
  +    pub max_optimization_threads: Option<MaxOptimizationThreads>,
   }
  ```
- Previous version of this PR were using macro_rules.
  <details><summary>old description</summary>

  This PR replaces this coercion-via-json with a `impl_diff_config` that macro that produces the conversion boilerplate. Any missing field will result in a compile error; and "find usages" on `HnswConfig` fields will find this definition. Additionally, `DiffConfig` trait methods are infallible now (as there no JSON serialization/serialization errors possible). Also, the macro syntax mimics real Rust syntax, thus `rustfmt` will format it correctly.
  
  ```rust
  impl_diff_config!(
      {
          impl DiffConfig<HnswConfigDiff> for HnswConfig {}
          impl DiffConfig<HnswConfigDiff> for HnswConfigDiff {}
          impl From<HnswConfig> for HnswConfigDiff {}
      },
      common_fields(
          m,
          ef_construct,
          full_scan_threshold,
          max_indexing_threads,
          on_disk,
          payload_m,
          copy_vectors,
      ),
      config_only_fields()
  );
  ```
  
  </details>

### DIFF
--- a/lib/collection/src/operations/config_diff.rs
+++ b/lib/collection/src/operations/config_diff.rs
@@ -1,65 +1,37 @@
+//! Structures for partial update of collection params
+
 #![allow(deprecated)] // hack to remove warning for memmap_threshold deprecation below
 
 use std::num::NonZeroU32;
 
 use api::rest::MaxOptimizationThreads;
-use merge::Merge;
 use schemars::JsonSchema;
 use segment::types::{
     BinaryQuantization, HnswConfig, ProductQuantization, ScalarQuantization, StrictModeConfig,
 };
-use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
 use validator::{Validate, ValidationErrors};
 
 use crate::config::{CollectionParams, WalConfig};
-use crate::operations::types::CollectionResult;
 use crate::optimizers_builder::OptimizersConfig;
 
-// Structures for partial update of collection params
-// TODO: make auto-generated somehow...
-
-pub trait DiffConfig<Diff>: Clone + DeserializeOwned + Serialize
-where
-    Diff: Sized + Serialize + DeserializeOwned + Merge,
-{
+pub trait DiffConfig<Diff>: Clone {
     /// Update this config with field from `diff`
     ///
     /// The `diff` has higher priority, meaning that fields specified in
     /// the `diff` will always be in the returned object.
-    fn update(&self, diff: Diff) -> CollectionResult<Self> {
-        update_config(self, diff)
-    }
+    fn update(&self, diff: &Diff) -> Self;
 
-    fn update_opt(&self, diff: Option<Diff>) -> Self {
+    fn update_opt(&self, diff: Option<&Diff>) -> Self {
         match diff {
-            Some(diff) => self.update(diff).unwrap_or_else(|e| {
-                log::warn!("Failed to apply config diff: {e}");
-                self.clone()
-            }),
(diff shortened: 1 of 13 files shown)


xzfc:
@coderabbitai review

lib/collection/src/operations/config_diff.rs
```diff
@@ -1,65 +1,37 @@
+//! Structures for partial update of collection params
+
 #![allow(deprecated)] // hack to remove warning for memmap_threshold deprecation below
 
 use std::num::NonZeroU32;
 
 use api::rest::MaxOptimizationThreads;
-use merge::Merge;
```
monoid, line 6: It seems that `merge` may be removed completely at other places too (at least it passes tests).

generall:
I understand the benefit of having explicit approach with compile-time errors if we forget something.

But why can't we just write the code instead of macro? Please don't tell me that macros are simpler.

xzfc:
> But why can't we just write the code instead of macro?

Sure.

lib/collection/src/operations/config_diff.rs
xzfc: Thanks.

timvisee:
I liked the macro more :smiling_face_with_tear: 

But let's merge it like this!

PR #7296, opened 2025-09-23
Refactor scroll by arguments

generall:
separate internal scroll operation from the one used by user

Previously, we had a single `scroll_by` function on a shard level, which were used by both: user requests and internal scrolls.

This PR creates one additional entry point for internal scroll requests, and switches all internal functions to it. In this way we can guarantee that:

- we won't use rate limits for internal requests
- we can log slow requests for internal/external operations separately

### DIFF
--- a/lib/collection/src/shards/dummy_shard.rs
+++ b/lib/collection/src/shards/dummy_shard.rs
@@ -8,12 +8,19 @@ use common::tar_ext;
 use segment::data_types::facets::{FacetParams, FacetResponse};
 use segment::data_types::manifest::SnapshotManifest;
 use segment::index::field_index::CardinalityEstimation;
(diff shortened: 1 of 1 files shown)


lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -138,48 +147,68 @@ impl ShardOperation for LocalShard {
...
-                "with_vector": with_vector,
-                "filter": filter,
-                "order_by": order_by,
-            }))
-        });
+        log_request_to_collector(&self.collection_name, elapsed, || request);
         Ok(result)
     }
 
+    async fn local_scroll_by_id(
```
timvisee, line 188: To self check: can we bypass the rate limiter somehow as a user?

PR #7298, opened 2025-09-24
Fix corruption due to concurrent update during snapshot

agourlay:
This PR fixes the non-atomic copy on write by forcing the serialization of all updates.

During snapshot, a single write segment is shared by all proxy segments.

At the end of the segments snapshot process, deletes are propagated to the wrapped segments.

This happens concurrently with user update operations which are serialized through an update channel.

As demonstrated by the test `test_continuous_snapshot`, it is possible that the delete propagation happens right in between of the CoW for the `set_payload` operation.

```rust
 fn set_payload(
        &mut self,
        op_num: SeqNumberType,
        point_id: PointIdType,
        payload: &Payload,
        key: &Option<JsonPath>,
    ) -> OperationResult<bool> {
        self.move_if_exists(op_num, point_id, hw_counter)?;
        // NOT ATOMIC - write segment can be updated from another proxy segment
        self.write_segment
            .get()
            .write()
            .set_payload(op_num, point_id, payload, key, hw_counter)
    }
```

This PR introduces a new update lock at the level of the segment holder to ensure that the snapshot process does not interfere with on-going client updates.

### DIFF
--- a/lib/shard/src/segment_holder/mod.rs
+++ b/lib/shard/src/segment_holder/mod.rs
@@ -16,7 +16,7 @@ use common::iterator_ext::IteratorExt;
 use common::save_on_disk::SaveOnDisk;
 use common::tar_ext;
 use io::storage_version::StorageVersion;
-use parking_lot::{RwLock, RwLockReadGuard, RwLockUpgradableReadGuard, RwLockWriteGuard};
+use parking_lot::{Mutex, RwLock, RwLockReadGuard, RwLockUpgradableReadGuard, RwLockWriteGuard};
 use rand::seq::IndexedRandom;
 use segment::common::operation_error::{OperationError, OperationResult};
 use segment::data_types::manifest::SnapshotManifest;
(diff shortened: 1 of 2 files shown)


lib/shard/src/segment_holder/mod.rs
```diff
@@ -1261,6 +1270,7 @@ impl SegmentHolder {
                 LockedSegment::Proxy(proxy_segment) => Some((segment_id, proxy_segment)),
                 LockedSegment::Original(_) => None,
             }).for_each(|(proxy_id, proxy_segment)| {
+                let _update_guard = segments_lock.update_lock.lock();
```
timvisee, line 1273: Do we not use the same lock below because we assume we have exclusive access to it, because we have a write lock on the whole segment holder?

agourlay: In the `Batch 1` we have only a read lock on the segment holder, so there can be other update concurrently.

Here I am grabbing the update lock on the smallest scope possible.

Would you prefer it to be acquired once for all proxes?

timvisee: > Would you prefer it to be acquired once for all proxes?

No, I think the current approach is fine.

I agree that we should keep this lock as short as possible.

PR #7295, opened 2025-09-23
[WIP] Continuous snapshot data loss

agourlay:
```
RUST_LOG=debug cargo nextest run --all continuous --nocapture

────────────
 Nextest run ID 8c335d13-b250-42b7-9da4-32c74ff69005 with nextest profile: default
    Starting 1 test across 22 binaries (1350 tests skipped)
       START             collection::integration continuous_snapshot_test::test_continuous_snapshot

running 1 test
[2025-09-23T14:10:41Z INFO  wal] Wal { path: "/tmp/test_collectionf8Rfmt/0/wal", segment-count: 1, entries: [0, 0)  }: opened
[2025-09-23T14:10:41Z INFO  collection::collection::snapshots] Creating collection snapshot test-0-2025-09-23-14-10-41.snapshot into "/tmp/test_snapshotstnfDIe/test-0-2025-09-23-14-10-41.snapshot"
[2025-09-23T14:10:41Z INFO  collection::collection::snapshots] Creating collection snapshot test-0-2025-09-23-14-10-41.snapshot into "/tmp/test_snapshotstnfDIe/test-0-2025-09-23-14-10-41.snapshot"
[2025-09-23T14:10:41Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-23T14:10:41Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-23T14:10:41Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-23T14:10:41Z INFO  collection::collection::snapshots] Creating collection snapshot test-0-2025-09-23-14-10-41.snapshot into "/tmp/test_snapshotstnfDIe/test-0-2025-09-23-14-10-41.snapshot"
[2025-09-23T14:10:41Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-23T14:10:41Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-23T14:10:41Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-23T14:10:41Z INFO  collection::collection::snapshots] Creating collection snapshot test-0-2025-09-23-14-10-41.snapshot into "/tmp/test_snapshotstnfDIe/test-0-2025-09-23-14-10-41.snapshot"
[...]

PR #7300, opened 2025-09-24
Add debug assertion for deleted version on proxy

agourlay:
The continuous snapshot test failing in https://github.com/qdrant/qdrant/pull/7295 has shown that we are applying stale delete operation on write segments.

I discovered this while sprinkling the handling of delete version with debug assertions.

The fix proposed here sadly does not fix https://github.com/qdrant/qdrant/pull/7295 but improve correctness.

### DIFF
--- a/lib/shard/src/proxy_segment/segment_entry.rs
+++ b/lib/shard/src/proxy_segment/segment_entry.rs
@@ -237,16 +237,13 @@ impl SegmentEntry for ProxySegment {
         };
 
         self.set_deleted_offset(point_offset);
-        let mut write_segment_guard = self.write_segment.get().write();
-        let point_version_in_write_segment = write_segment_guard
-            .point_version(point_id)
-            .unwrap_or_default();
-        let mut was_deleted_in_writable = write_segment_guard.has_point(point_id);
-        // Make sure deletion operation is not stale
-        if op_num > point_version_in_write_segment {
-            was_deleted_in_writable =
-                write_segment_guard.delete_point(op_num, point_id, hw_counter)?;
-        }
+
+        let was_deleted_in_writable = self
+            .write_segment
+            .get()
+            .write()
+            .delete_point(op_num, point_id, hw_counter)?;
+
         Ok(was_deleted || was_deleted_in_writable)
     }
 


lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -198,46 +198,59 @@ impl SegmentEntry for ProxySegment {
...
-            .write_segment
-            .get()
-            .write()
-            .delete_point(op_num, point_id, hw_counter)?;
-
+        let mut write_segment_guard = self.write_segment.get().write();
+        let point_version_in_write_segment = write_segment_guard
+            .point_version(point_id)
+            .unwrap_or_default();
+        let mut was_deleted_in_writable = false;
```
agourlay: a `debug_assert` here failed showing that we apply the same `op_num` twice on the write segment

lib/shard/src/proxy_segment/mod.rs
```diff
@@ -194,13 +194,21 @@ impl ProxySegment {
...
                     operation_version: op_num,
                 },
             );
+            if let Some(prev) = prev {
+                debug_assert!(
+                    prev.operation_version < op_num,
+                    "Overriding deleted flag {:?} with older op_num:{}",
+                    prev,
+                    op_num
+                )
```
timvisee, line 208: Nit:

```suggestion
                debug_assert!(
                    prev.operation_version < op_num,
                    "Overriding deleted flag {prev:?} with older op_num:{op_num}",
                )
```

lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -198,46 +198,59 @@ impl SegmentEntry for ProxySegment {
...
+                        },
+                    );
+                    was_deleted = prev.is_none();
+                    if let Some(prev) = prev {
+                        debug_assert!(
+                            prev.operation_version < op_num,
+                            "Overriding deleted flag {:?} with older op_num:{}",
+                            prev,
+                            op_num
                         )
```
timvisee, line 213:
```suggestion
                        debug_assert!(
                            prev.operation_version < op_num,
                            "Overriding deleted flag {prev:?} with older op_num:{op_num}",
                        )
```

lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -198,46 +198,59 @@ impl SegmentEntry for ProxySegment {
...
```
timvisee, line 232:
```suggestion
                        debug_assert!(
                            prev.operation_version < op_num,
                            "Overriding deleted flag {prev:?} with older op_num:{op_num}",
                        )
```

lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -198,46 +198,59 @@ impl SegmentEntry for ProxySegment {
...
+        let mut write_segment_guard = self.write_segment.get().write();
+        let point_version_in_write_segment = write_segment_guard
+            .point_version(point_id)
+            .unwrap_or_default();
+        let mut was_deleted_in_writable = write_segment_guard.has_point(point_id);
+        // Make sure deletion operation is not stale
+        if op_num > point_version_in_write_segment {
+            was_deleted_in_writable =
+                write_segment_guard.delete_point(op_num, point_id, hw_counter)?;
+        }
```
timvisee: Yes, we should already ignore the operation deeper down here:

https://github.com/qdrant/qdrant/blob/febf72a47efc63442e026ecfa596812c5e5bba1e/lib/segment/src/segment/segment_ops.rs#L279-L288

agourlay: Thanks for double checking, I will remove this check and just propose the debug assertions :+1:

timvisee (approved):

PR #7301, opened 2025-09-24
Add replica metrics

JojiiOfficial:
Depends on #7275

Adds the following new metrics to the /metric api

```
# HELP active_total_replicas accumulated total number of active replicas across all shards
# TYPE active_total_replicas gauge
active_total_replicas 4

# HELP active_expected_replicas accumulated total number of expected replicas across all shards
# TYPE active_expected_replicas gauge
active_expected_replicas 4

# HELP active_total_replicas_max minimum number of replicas across shards
# TYPE active_total_replicas_max gauge
active_total_replicas_max 2

# HELP active_total_replicas_min maximum number of replicas across shards
# TYPE active_total_replicas_min gauge
active_total_replicas_min 2
```


timvisee:
> ```
> # HELP active_expected_replicas accumulated total number of expected replicas across all shards
> # TYPE active_expected_replicas gauge
> active_expected_replicas 4
> ```

Doesn't it make more sense to count this per collection?

JojiiOfficial:
I made this metric not per-collection to keep consistent with the other metrics. Additionally, we can compare this parameter against `active_total_replicas` more easily if we don't measure it on per-collection level.

src/common/metrics.rs
```diff
@@ -199,6 +208,35 @@ impl MetricsProvider for CollectionsTelemetry {
                 .filter(|i| i.status == TrackerStatus::Optimizing)
                 .count();
 
+            let min_max_active_replicas = collection
+                .shards
+                .iter()
+                .flatten()
+                // Ignore shards in resharding state.
+                .filter(|shard| !shard.replicate_states.values().any(|i| i.is_resharding()))
```
timvisee: I don't think this is good enough to filter in case of resharding.

In case of resharding up, this only works during the very first stage of resharding. The new shard will become active before the extra replicas are created. By that time we should still not consider the new shard because it can be safely dropped.

In case of resharding down all replicas in all but the last shard go into resharding state at some point.

I think it'd be better to look at the current resharding state directly (not replica state, but resharding state). Then, based on it's current state we can decide whether to ignore some shard or not.

src/common/metrics.rs
```diff
@@ -177,10 +178,18 @@ impl MetricsProvider for CollectionsTelemetry {
...
         let mut optimizer_counter_metrics = vec![];
 
+        // Amount of optimizations currently runnig on the local node.
         let mut total_optimizations_running = 0;
 
+        // Min/Max/Expected/Active replicas over all shards.
+        let mut total_min_active_replicas = usize::MAX;
+        let mut total_max_active_replicas = 0;
+        let mut total_expected_replicas = 0;
+        let mut total_active_replicas = 0;
```
generall: Sum of number of replicas doesn't make sense to me, I would rather include min/max for expected replication factor.

src/common/metrics.rs
```diff
@@ -199,6 +208,35 @@ impl MetricsProvider for CollectionsTelemetry {
                 .filter(|i| i.status == TrackerStatus::Optimizing)
                 .count();
 
+            let min_max_active_replicas = collection
```
generall, line 196: let's move replica counting into a separate function/file. This one becomes bulky

src/common/metrics.rs
```diff
@@ -232,6 +270,38 @@ impl MetricsProvider for CollectionsTelemetry {
...
+        if total_max_active_replicas != 0 {
+            metrics.push(metric_family(
+                "active_replicas_max",
+                "minimum number of active replicas across all shards",
+                MetricType::GAUGE,
+                vec![gauge(total_max_active_replicas as f64, &[])],
+            ));
+        }
+
+        if total_min_active_replicas != usize::MAX {
```
generall: cloud doesn't like when empty metric is not displayed. Let's report 0 in this case

generall (dismissed):

src/common/metrics.rs
timvisee: For some context, not reporting 0 results in gaps in the graphs.

src/common/metrics.rs
JojiiOfficial: I'd suggest to do this in a separate PR once the other ones are merged too. This way rebasing and merging with the other currently open PRs is significantly easier.

src/common/metrics.rs
timvisee: I understand that

```rust
        let mut total_expected_replicas = 0;
        let mut total_active_replicas = 0;
```

may not be useful.

But

> I would rather include min/max for expected replication factor.

do you also suggest to group `total_{min,max}_active_replicas` per configured replication factor?

timvisee: After a short discussion, we're happy with just the first two (`total_{min,max}_active_replicas`), which is a global min/max count.

src/common/metrics.rs
timvisee: Discussed in a call. Handled in <https://github.com/qdrant/qdrant/pull/7301/commits/7044a84f28e15d28760402e4c5d2f3560a80085d>

PR #7302, opened 2025-09-24
Add num points/vectors to metrics API

JojiiOfficial:
Depends on  #7301

Add per-collection point and vector approximations to /metrics API:

```
# HELP collection_points approximate amount of points per collection
# TYPE collection_points gauge
collection_points{id="benchmark"} 100000
collection_points{id="1234"} 100000

# HELP collection_vectors approximate amount of vectors per collection
# TYPE collection_vectors gauge
collection_vectors{id="benchmark"} 100000
collection_vectors{id="1234"} 100000
```

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -177,11 +177,16 @@ impl MetricsProvider for CollectionsTelemetry {
             vec![gauge(vector_count as f64, &[])],
         ));
 
+        // Optimizers
         let mut total_optimizations_running = 0;
 
         // Min/Max/Expected/Active replicas over all shards.
         let mut total_min_active_replicas = usize::MAX;
         let mut total_max_active_replicas = 0;
+
+        // Points & Vectors per collection
+        let mut vectors_per_collection = vec![];
+        let mut points_per_collection = vec![];
 
         for collection in self.collections.iter().flatten() {
             let collection = match collection {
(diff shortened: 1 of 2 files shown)


timvisee (approved):

src/common/metrics.rs
```diff
@@ -194,6 +209,20 @@ impl MetricsProvider for CollectionsTelemetry {
             MetricType::GAUGE,
             vec![gauge(total_optimizations_running as f64, &[])],
         ));
+
+        metrics.push(metric_family(
+            "collection_points",
```
timvisee, line 296: We might need to name it like this?

```suggestion
            "collection_points_total",
```

We can decide on that separately. I'd be fine with adjusting all metric names at the end in a single PR to follow convention. Then we can review together with some of our cloud people whom are more familiar with this.

timvisee (approved):

timvisee:
Does it maybe make sense to also partition the vector counts by their vector name?

It could give interesting insights, like a named vector having zero vectors.

src/common/metrics.rs
JojiiOfficial: According to the prometheus docs https://prometheus.io/docs/instrumenting/writing_exporters/
> _total is a convention for counters, you should use it if you’re using the COUNTER type.

`collection_points` is a gauge, not counter type so per naming convention, this should not have a `_total` suffix.
If we already use this suffix for gauges in other places, I'm fine with adding this suffix to make it consistent.

JojiiOfficial:
> Does it maybe make sense to also partition the vector counts by their vector name?
It could give interesting insights, like a named vector having zero vectors.

Good Idea! I'll tackle this in a separate PR

PR #7306, opened 2025-09-25
Use former scroll_lock to prevent parallel updates during snapshot

generall:
This PR includes refactoring of segment-holder related functions to unlock usage of async lock during snapshot.

This PR does the following:

- moves snapshot related functions into a dedicated file in both: `shard` and `collection` crates
- Moves some snapshot-related functions from `shard` into `collection` to unblock usage of tokio rwlock
- Changes fixture to be `ctg testing` so some tests can be also moved to `collection`
- Removes `update_lock` in segment holder
- Renames and propagates scroll lock into `snapshot_all_segments` function

### DIFF
--- a/lib/collection/src/shards/local_shard/scroll.rs
+++ b/lib/collection/src/shards/local_shard/scroll.rs
@@ -153,7 +153,7 @@ impl LocalShard {
         let stopping_guard = StoppingGuard::new();
         let segments = self.segments.clone();
 
-        let scroll_lock = self.update_operation_lock.read().await;
+        let update_operation_lock = self.update_operation_lock.read().await;
         let (non_appendable, appendable) = segments.read().split_segments();
 
         let read_filtered = |segment: LockedSegment, hw_counter: HardwareCounterCell| {
@@ -210,7 +210,7 @@ impl LocalShard {
         .await
         .map_err(|_: Elapsed| CollectionError::timeout(timeout.as_secs() as usize, "retrieve"))??;
 
-        drop(scroll_lock);
+        drop(update_operation_lock);
 
         let ordered_records = point_ids
             .iter()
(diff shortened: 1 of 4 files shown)


lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -250,7 +258,7 @@ impl LocalShard {
             total_optimized_points,
             disk_usage_watcher,
             read_rate_limiter,
-            scroll_read_lock,
+            update_operation_lock: scroll_read_lock,
```
generall, line 261: todo: rename

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -153,7 +153,7 @@ impl LocalShard {
         let stopping_guard = StoppingGuard::new();
         let segments = self.segments.clone();
 
-        let scroll_lock = self.scroll_read_lock.read().await;
```
generall, line 156: todo: rename

lib/collection/src/shards/local_shard/snapshot.rs
```diff
@@ -0,0 +1,352 @@
...
+                Err(lock) => segments_lock = lock,
+            }
+        }
+    }
+    proxies.retain(|(id, _)| !unproxied_segment_ids.contains(id));
+
+    // Unproxy all segments
+    // Always do this to prevent leaving proxy segments behind
+    log::trace!("Unproxying all shard segments after function is applied");
+    let _update_guard = update_lock.blocking_write();
```
timvisee, line 348: 💡  The important change (lock) is here, and above before `try_unproxy_segment`

timvisee:
:heavy_check_mark: Great idea to share the lock

PR #7307, opened 2025-09-25
Add unindexed vectors to telemetry and metrics API

JojiiOfficial:
Depends on #7302 

Adds information about unindexed vectors to the /telemetry and /metrics API

```
# HELP collection_unindexed_vectors approximate amount of vectors per collection that are not indexed
# TYPE collection_unindexed_vectors gauge
collection_unindexed_vectors{id="benchmark"} 0
```


lib/collection/src/shards/local_shard/telemetry.rs
```diff
@@ -59,8 +59,11 @@ impl LocalShard {
             vectors_size_bytes,
             payloads_size_bytes,
             num_points,
+            num_indexed_vectors,
         } = self.get_size_stats().await;
 
+        let num_unindexed_vectors = num_vectors.saturating_sub(num_indexed_vectors);
```
timvisee: Maybe it's better to explicitly count vectors on segments that don't have an index? 

I'm not sure if we want to count points in very small segments (below the indexing threshold) to be unindexed? If we don't, it'll be easier to count with the way I describe above.

Wdyt?

JojiiOfficial: Implemented!

src/common/metrics.rs
```diff
@@ -201,6 +206,37 @@ impl MetricsProvider for CollectionsTelemetry {
...
+                indexed_only_excluded.push(gauge(
+                    vector_size as f64,
+                    &[("id", &collection.id), ("vector", name)],
+                ))
+            }
+        }
+
+        if !indexed_only_excluded.is_empty() {
+            metrics.push(metric_family(
+                "collection_index_only_excluded",
```
timvisee: Following open metrics convention, shouldn't this be called something like?

`collection_index_only_excluded_vectors_total`

Where we have:
- 'vectors' - as a unit for the count
- 'total' - to emphasize this is an aggregated count

lib/collection/src/shards/local_shard/telemetry.rs
```diff
@@ -133,3 +150,61 @@ impl LocalShard {
         stats.unwrap_or_default()
     }
 }
+
+/// Returns vectors which will be excluded from requests with `index-only` enabled.
```
timvisee: Nit:

```suggestion
/// Returns number of vectors which will be excluded from requests with `indexed_only` enabled.
///
/// This effectively counts vectors in large unindexed segments.
```

src/actix/api/service_api.rs
```diff
@@ -80,9 +79,8 @@ async fn metrics(
         .prepare_data(
             &access,
             TelemetryDetail {
-                level: DetailsLevel::Level3,
+                level: DetailsLevel::Level4,
```
timvisee, line 82: This will make the call to `/metrics` more expensive, but I guess we don't really have a choice.

docs/redoc/master/openapi.json
```diff
@@ -12165,6 +12165,15 @@
           "async_scorer": {
             "type": "boolean",
             "nullable": true
+          },
+          "index_only_excluded_vectors": {
```
timvisee: In [other](https://api.qdrant.tech/api-reference/search/query-points#request.body.params.SearchParams.indexed_only) places we call it `indexed_only`, I vote to keep that convention:

```suggestion
          "indexed_only_excluded_vectors": {
```

src/common/metrics.rs
```diff
@@ -201,6 +206,37 @@ impl MetricsProvider for CollectionsTelemetry {
...
+                indexed_only_excluded.push(gauge(
+                    vector_size as f64,
+                    &[("id", &collection.id), ("vector", name)],
+                ))
+            }
+        }
+
+        if !indexed_only_excluded.is_empty() {
+            metrics.push(metric_family(
+                "collection_index_only_excluded_points",
```
```suggestion
                "collection_indexed_only_excluded_points",
```

lib/collection/src/shards/local_shard/telemetry.rs
```diff
@@ -133,3 +150,63 @@ impl LocalShard {
...
+
+                    let points = segment_guard.available_point_count();
+                    Some((vector_name, vector_storage_size.unwrap(), points))
+                })
+        })
+        .filter(|(_, vector_size_bytes, _)| {
+            // Filter out only large segments that do not support full-scan, as smaller segments can
+            // be searched quickly without using an index and are included in index-only searches.
+            *vector_size_bytes > search_optimized_threshold_bytes
+        })
```
timvisee, line 204: Thought: maybe we can add a function to segments itself that tells us whether to include it in `indexed_only` search. That would make the logic here simpler.

timvisee: Implemented in <https://github.com/qdrant/qdrant/pull/7463>

I suggest to merge it into this one if you like it, then we can merge this PR as a whole.

src/common/metrics.rs
timvisee: I took the liberty of making this change

lib/collection/src/shards/local_shard/telemetry.rs
timvisee: We well merge it later to not block our current PR queue.

PR #7310, opened 2025-09-25
Dead shards in /metrics

JojiiOfficial:
Depends on  #7307

Adds dead-shards metrics to /metrics API


lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -1337,6 +1337,22 @@ impl ReplicaState {
...
+    /// Returns `true` if the replica state is `Dead`.
+    pub fn is_dead(&self) -> bool {
+        match self {
+            ReplicaState::Dead => true,
+
+            ReplicaState::Partial
+            | ReplicaState::PartialSnapshot
+            | ReplicaState::Recovery
+            | ReplicaState::Active
+            | ReplicaState::Initializing
```
timvisee: I was thinking whether it makes sense to also include initializing shards in the count.

But I don't think so. Doing so might make the number of dead shards blink when creating a new collection, and I don't think that is desired.

generall: I would rather do `!status.is_active`

src/common/metrics.rs
```diff
@@ -259,6 +269,13 @@ impl MetricsProvider for CollectionsTelemetry {
             MetricType::GAUGE,
             vectors_per_collection,
         ));
+
+        metrics.push(metric_family(
+            "dead_shards_total",
+            "total amount of dead shards",
```
```suggestion
            "total amount of shard replicas in non-active state",
```

timvisee (approved):
See the review remark below, I think that description is a bit more on-point:

Other than that, all good!

PR #7312, opened 2025-09-25
Python bindings for Qdrant on Edge

ffuugoo:
This PR Implements partial Python bindings for Qdrant Edge. So far only basic upsertion and search are implemented, but bindings are boilerplate-heavy, so more types and methods would be covered in follow-up PRs.

```shell
cd lib/edge/python

python -m venv .venv
source .venv/bin/activate

pip install --user maturin
maturin develop --no-default-features

cd examples
python qdrant-edge.py
```

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/edge/python/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "edge-py"
+version = "0.1.0"
+authors = ["Qdrant Team <info@qdrant.tech>"]
+license = "Apache-2.0"
+edition = "2024"
```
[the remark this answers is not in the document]
timvisee: Grandpa, you're getting old! :older_man: 

https://blog.rust-lang.org/2025/02/20/Rust-1.85.0/

lib/edge/python/examples/qdrant-edge.py
timvisee, line 1: My only nit here is that this interface seems very different than what we currently offer in your Python client.

I'm curious if that is desired. Ideally we use something similar so that we can reuse parts of our documentation, and don't confuse users.

Thoughts?

lib/edge/python/src/config.rs
```diff
@@ -0,0 +1,241 @@
...
+impl PyDistance {
+    fn _variants(distance: Distance) {
+        match distance {
+            Distance::Cosine => (),
+            Distance::Euclid => (),
+            Distance::Dot => (),
+            Distance::Manhattan => (),
+        }
+    }
+}
```
timvisee, line 95: What is this for?

lib/edge/python/src/config.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub const IN_RAM_MMAP: Self = Self(PayloadStorageType::InRamMmap);
+}
+
+impl PyPayloadStorageType {
+    fn _variants(storage_type: PayloadStorageType) {
+        #[allow(unreachable_patterns)]
+        match storage_type {
+            PayloadStorageType::Mmap => (),
+            PayloadStorageType::InRamMmap => (),
+            _ => todo!(), // TODO: Ignore RocksDB storage types
```
timvisee, line 238: We can feature flag it and implement the bare minimum (probably still with `unimplemented!()`).

Then we'll fully remove it once dropping the RocksDB feature, and don't keep the `_ =>` branch here.

lib/edge/python/src/config.rs
ffuugoo: To get compilation error when new variant is added to Rust code.

timvisee: :ok_hand:

lib/edge/python/src/search.rs
```diff
@@ -0,0 +1,192 @@
...
+use pyo3::prelude::*;
+use segment::data_types::vectors::*;
+use segment::types::*;
+use shard::search::*;
+
+use super::*;
+
+#[pyclass(name = "SearchRequest")]
+#[derive(Clone, Debug, Into)]
+pub struct PySearchRequest(CoreSearchRequest);
```
generall, line 12: `CoreSearchRequest` is soon-to-be deprecated structure, in the core it is only used for old `search` method, which we want to continue gradually removing in 1.16.x

I would avoid making any more dependants on it

generall (approved):

PR #7308, opened 2025-09-25
Test continuous snapshot data loss

agourlay:
This test used to be broken with:

```
RUST_LOG=debug cargo nextest run -p collection test_continuous_snapshot --nocapture

...
[2025-09-25T09:37:08Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-25T09:37:08Z DEBUG segment::segment::snapshot] Taking snapshot of segment af3cee4d-a4a5-4474-a344-749a12aca0ba
[2025-09-25T09:37:08Z DEBUG segment::segment::snapshot] Taking snapshot of segment c496b4a2-e09e-42c9-97ba-f15e5f362061
[2025-09-25T09:37:08Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-25T09:37:08Z DEBUG segment::segment::snapshot] Taking snapshot of segment 5387da80-1800-4d41-ae7e-d81437068c1e
[2025-09-25T09:37:08Z DEBUG segment::segment::snapshot] Taking snapshot of segment c496b4a2-e09e-42c9-97ba-f15e5f362061
[2025-09-25T09:37:08Z DEBUG segment::segment::snapshot] Segment c496b4a2-e09e-42c9-97ba-f15e5f362061 is already snapshotted!
[2025-09-25T09:37:08Z INFO  shard::proxy_segment::snapshot_entry] Taking a snapshot of a proxy segment
[2025-09-25T09:37:08Z DEBUG segment::segment::snapshot] Taking snapshot of segment 7497eab1-38e7-42bf-86ab-db21a3fe4e12
[2025-09-25T09:37:08Z DEBUG segment::segment::snapshot] Taking snapshot of segment c496b4a2-e09e-42c9-97ba-f15e5f362061
[2025-09-25T09:37:08Z DEBUG segment::segment::snapshot] Segment c496b4a2-e09e-42c9-97ba-f15e5f362061 is already snapshotted!
[2025-09-25T09:37:08Z DEBUG collection::shards::replica_set::execute_read_operation] Read operation failed: Service internal error: No version for point 0
[2025-09-25T09:37:08Z DEBUG collection::update_handler] Optimizer 'merge' running on segments: [1692, 1693, 1694]
[2025-09-25T09:37:08Z DEBUG collection::collection_manager::optimizers::segment_optimizer] Available space: 27.23 GiB, needed for optimization: 1.33 MiB
[...]

PR #7314, opened 2025-09-25
[WIP] Mitigate concurrent deletion propagation

agourlay:
Explore lockless solution https://github.com/qdrant/qdrant/pull/7308

Mitigates TOCTU on points retrieval

### DIFF
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -419,9 +419,10 @@ impl SegmentsSearcher {
         segments
             .read()
             .read_points(points, is_stopped, |id, segment| {
-                let version = segment.point_version(id).ok_or_else(|| {
-                    OperationError::service_error(format!("No version for point {id}"))
-                })?;
+                let Some(version) = segment.point_version(id) else {
+                    // ignore point
+                    return Ok(false);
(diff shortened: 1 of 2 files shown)


lib/collection/src/collection_manager/segments_searcher.rs
```diff
@@ -419,9 +419,10 @@ impl SegmentsSearcher {
         segments
             .read()
             .read_points(points, is_stopped, |id, segment| {
-                let version = segment.point_version(id).ok_or_else(|| {
-                    OperationError::service_error(format!("No version for point {id}"))
-                })?;
+                let Some(version) = segment.point_version(id) else {
+                    // ignore point
+                    return Ok(false);
+                };
```
timvisee, line 425: What is the function of this one?

I'm not sure this is reliable enough to eliminate TOCTU. Even though we have a read lock on the segment here, it does not mean that its logical points cannot change. I'm thinking about proxies with a shared write segment here.

Or is it?

timvisee: Nevermind!

agourlay:
Closing in favor of a proper architecture change to avoid the problem completely https://github.com/qdrant/qdrant/pull/7345

PR #7316, opened 2025-09-26
Currently running optimizers in Metrics

JojiiOfficial:
Supersedes #7275

Implements the `optimizer_running_processes` metrics as discussed:

```
# HELP optimizer_running_processes number of optimization processes running in total
# TYPE optimizer_running_processes gauge
optimizer_running_processes 1
```

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -189,8 +189,8 @@ impl MetricsProvider for CollectionsTelemetry {
         }
 
         metrics.push(metric_family(
-            "optimizer_processes_running",
-            "number of optimization processes running in total",
+            "optimizer_running_processes",
+            "number of currently running optimization processes",
             MetricType::GAUGE,
             vec![gauge(total_optimizations_running as f64, &[])],
         ));


generall (approved):

lib/collection/src/telemetry.rs
```diff
@@ -51,6 +52,20 @@ impl CollectionTelemetry {
...
+    ///       Otherwise, this function will return 0, which may not be correct.
+    pub fn count_optimizers_running(&self) -> usize {
+        self.shards
+            .iter()
+            .flatten()
+            .filter_map(|replica_set| replica_set.local.as_ref())
+            .flat_map(|local_shard| local_shard.optimizations.log.iter().flatten())
+            .filter(|log| log.status == TrackerStatus::Optimizing)
+            .count()
+    }
```
timvisee, line 68: When reading the PR description I thought this could be done with counting with some atomic number and a guard.

But I think your approach is better, where we use state that we already have. 👍

timvisee (approved):

PR #7320, opened 2025-09-29
smaller update batch size

generall:
Noticed in benchmark, that update of 1k+ poitns at a sime could increase latency of the scroll operation to 2 seconds. 
Ususally this is unacceptable performance, so we have to make update batches smaller.

- **make update batch smaller to prevent long blocks**
- **limit vector update operations as well**
- **limit update chunk for point upsertion**

### DIFF
--- a/lib/shard/src/update.rs
+++ b/lib/shard/src/update.rs
@@ -166,7 +166,7 @@ pub fn process_field_index_operation(
     }
 }
 
-/// No not insert more than this number of points in a single update operation chunk
+/// Do not insert more than this number of points in a single update operation chunk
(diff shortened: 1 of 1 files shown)


lib/shard/src/update.rs
```diff
@@ -166,6 +166,11 @@ pub fn process_field_index_operation(
     }
 }
 
+/// No not insert more than this number of points in a single update operation chunk
```
timvisee:
```suggestion
/// Do not insert more than this number of points in a single update operation chunk
```

timvisee (commented):
Running some tests before I hit approve 👍

PR #7322, opened 2025-09-29
total segment proxy lock

generall:
- **Continuous snapshot data loss**
- **decouple segment-entry and snapshot-entry**
- **attempt to deep-lock proxy segment for consistency**

### DIFF
--- a/lib/collection/tests/integration/continuous_snapshot_test.rs
+++ b/lib/collection/tests/integration/continuous_snapshot_test.rs
@@ -125,8 +125,8 @@ async fn test_continuous_snapshot() {
                     )
                     .await?;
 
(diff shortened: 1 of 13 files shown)


timvisee:
After some experimenting, I was not able to get this branch into a nice state. It is doable with a lot of unsafe code, but I don't think we should take that approach.

Also, I'm not confident this on its own will solve the issues we've been seeing. To fix the problem, I think a read lock must cascade into all wrapped/write segments of proxies. That will be the only way to prevent writes into some inner write segment which may be shared with another proxy. I don't see it being implemented in that way in this PR.

timvisee:
I believe we dropped this idea in favor of other improvements. Closing this.

PR #7339, opened 2025-09-30
Improve snapshot transfer tests

agourlay:
This branch is the result of refurbishing the learning from hunting down an inconsistency in the shard transfer.

Some changes are I believe a net improvements and should be kept to make the next debugging session easier.

The changes are:
- enable trace logging during tests to gather more info by default
- check data consistency on count failure
- add a few more trace logging
- document trick to slow down qdrant nodes with CPU quotas
- remove sleep

### DIFF
--- a/tests/consensus_tests/test_shard_snapshot_transfer.py
+++ b/tests/consensus_tests/test_shard_snapshot_transfer.py
@@ -190,12 +190,12 @@ def test_shard_snapshot_transfer_throttled_updates(tmp_path: pathlib.Path):
     )
 
     # Insert some initial number of points
-    upsert_random_points(peer_api_uris[0], 10000)
+    upsert_random_points(peer_api_uris[0], 10_000)
 
     # Start pushing points to the cluster
     upload_process_1 = run_update_points_in_background(peer_api_uris[0], COLLECTION_NAME, init_offset=100, throttle=True)
-    upload_process_2 = run_update_points_in_background(peer_api_uris[1], COLLECTION_NAME, init_offset=10000, throttle=True)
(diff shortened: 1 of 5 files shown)


tests/consensus_tests/utils.py
```diff
@@ -94,7 +94,7 @@ def get_env(p2p_port: int, grpc_port: int, http_port: int) -> Dict[str, str]:
     env["QDRANT__CLUSTER__P2P__PORT"] = str(p2p_port)
     env["QDRANT__SERVICE__HTTP_PORT"] = str(http_port)
     env["QDRANT__SERVICE__GRPC_PORT"] = str(grpc_port)
-    env["QDRANT__LOG_LEVEL"] = "DEBUG,raft::raft=info"
+    env["QDRANT__LOG_LEVEL"] = "TRACE,raft::raft=info,actix_http=info,tonic=info,want=info,mio=info"
```
agourlay, line 97: I manually excluded things that were too noisy.

tests/consensus_tests/test_shard_snapshot_transfer.py
```diff
@@ -219,13 +219,20 @@ def test_shard_snapshot_transfer_throttled_updates(tmp_path: pathlib.Path):
         })
     assert_http_ok(r)
 
+    # Wait for start of shard transfer
+    wait_for_collection_shard_transfers_count(peer_api_uris[0], COLLECTION_NAME, 1)
```
timvisee, line 223: I have seen this being flaky historically, because transfers aren't always long enough.

I'll do a few tests locally with 10k points to see if I can make this fail.

timvisee: Couldn't get it to fail just yet. Happy to merge and adjust if it shows problems again.

Issue #7335, opened 2025-09-30
Flaky consensus test_shard_snapshot_transfer_throttled_updates


agourlay:
```
 =================================== FAILURES ===================================
________________ test_shard_snapshot_transfer_throttled_updates ________________

tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-1/test_shard_snapshot_transfer_t0')

    def test_shard_snapshot_transfer_throttled_updates(tmp_path: pathlib.Path):
        assert_project_root()
    
        # seed port to reuse the same port for the restarted nodes
        peer_api_uris, peer_dirs, bootstrap_uri = start_cluster(tmp_path, N_PEERS, 20000)
    
        create_collection(peer_api_uris[0], shard_number=N_SHARDS, replication_factor=N_REPLICA)
        wait_collection_exists_and_active_on_all_peers(
            collection_name=COLLECTION_NAME,
            peer_api_uris=peer_api_uris
        )
    
        # Insert some initial number of points
        upsert_random_points(peer_api_uris[0], 10000)
    
        # Start pushing points to the cluster
        upload_process_1 = run_update_points_in_background(peer_api_uris[0], COLLECTION_NAME, init_offset=100, throttle=True)
        upload_process_2 = run_update_points_in_background(peer_api_uris[1], COLLECTION_NAME, init_offset=10000, throttle=True)
        upload_process_3 = run_update_points_in_background(peer_api_uris[2], COLLECTION_NAME, init_offset=20000, throttle=True)
    
        transfer_collection_cluster_info = get_collection_cluster_info(peer_api_uris[0], COLLECTION_NAME)
        receiver_collection_cluster_info = get_collection_cluster_info(peer_api_uris[2], COLLECTION_NAME)
    
        from_peer_id = transfer_collection_cluster_info['peer_id']
        to_peer_id = receiver_collection_cluster_info['peer_id']
    
        shard_id = transfer_collection_cluster_info['local_shards'][0]['shard_id']
    
        # Transfer shard from one node to another
    
        # Move shard `shard_id` to peer `target_peer_id`
        r = requests.post(
            f"{peer_api_uris[0]}/collections/{COLLECTION_NAME}/cluster", json={
                "replicate_shard": {
                    "shard_id": shard_id,
                    "from_peer_id": from_peer_id,
                    "to_peer_id": to_peer_id,
                    "method": "snapshot",
                }
            })
        assert_http_ok(r)
    
        # Wait for end of shard transfer
[...]

PR #7341, opened 2025-10-02
Use first WAL index instead of last in shard snapshot transfer

generall:
instead of last WAL index, we need to use first one as there are no guarantees that last WAL entries will make it into snapshot.

Should fix: https://github.com/qdrant/qdrant/issues/7335

Potential downsides:

- We will post-transfer more operations, maybe with duplicates. This means that potentially WAL delta will be compromized for some interval.

### DIFF
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -414,7 +414,13 @@ impl Inner {
         wal_keep_from: Arc<AtomicU64>,
         progress: Arc<ParkingMutex<TransferTaskProgress>>,
     ) -> Self {
-        let start_from = wrapped_shard.wal.wal.lock().await.last_index() + 1;
+        // If nothing else is specified, we need to keep all WAL entries, which might be
+        // potentially not yet applied(persisted) with in the storage.
+        //
+        // We assume, that snapshot made from this point forward is guaranteed to have all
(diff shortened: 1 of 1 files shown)


timvisee:
I don't think this change is right.

I'd argue the queue proxy should handle logical data, not physical/durably stored data. Mainly because the queue proxy isn't specific to snapshots. That means that it should start queueing all new logical operations from the point it is created, even if some data before it was not persisted yet.

You're completely right that this could be problematic in case of snapshots, where we copy/transfer persisted data. But that's an implementation detail specific to snapshots. And so our snapshot (transfer) logic should take care of handling this correctly, not the queue proxy itself.

I think that using `first_index` when using this for snapshots specifically makes more sense.

So:
- keep [this](https://github.com/qdrant/qdrant/blob/f66dc31452e49300b802e5fa4ebcc3a6b68f2154/lib/collection/src/shards/queue_proxy_shard.rs#L417) at `last_index`
- change [this](https://github.com/qdrant/qdrant/blob/3e536347e12daf592716615147d2d34be4a9ee18/lib/collection/src/shards/transfer/snapshot.rs#L185-L188) to use `first_index`

Even with <https://github.com/qdrant/qdrant/pull/7342> merged, I still believe we should make this change.

---

PR: https://github.com/qdrant/qdrant/pull/7346

timvisee:
Closing because of <https://github.com/qdrant/qdrant/pull/7346#issuecomment-3371411058>

PR #7346, opened 2025-10-06
In snasphot transfer, queue WAL operations from last acknowledged

timvisee:
Improves <https://github.com/qdrant/qdrant/pull/7341> based on [this](https://github.com/qdrant/qdrant/pull/7341#issuecomment-3370505121) comment.

I'd argue that this is the better implementation of <https://github.com/qdrant/qdrant/pull/7341>. Please see my motivation [here](https://github.com/qdrant/qdrant/pull/7341#issuecomment-3370505121).

I also argue this is better than <https://github.com/qdrant/qdrant/pull/7342> because this is cheaper. Snapshots are already very expensive, flushing them makes this worse. We already have a background task that is supposed to take care of this.

### All Submissions:

### DIFF
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -414,13 +414,10 @@ impl Inner {
         wal_keep_from: Arc<AtomicU64>,
         progress: Arc<ParkingMutex<TransferTaskProgress>>,
     ) -> Self {
-        // If nothing else is specified, we need to keep all WAL entries, which might be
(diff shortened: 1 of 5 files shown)


timvisee:
After another pass, I agree that <https://github.com/qdrant/qdrant/pull/7342> is the better approach.

The force flush in each segment on snapshot should not be a concern.

Leaving the queue proxy as-is is fine. In case of snapshot transfer, we instantiate it before doing all flushes. All flushes are forced, and use a plunger to ensure parallel operations are also properly flushed.

Closing this in favor of <https://github.com/qdrant/qdrant/pull/7342>.

PR #7356, opened 2025-10-06
Introduce fallback shard key for read and write requests

KShivendu:
Replacement of https://github.com/qdrant/qdrant/pull/7067 and https://github.com/qdrant/qdrant/pull/7069 since we decided to define fallback shard key on request level. Specifying it on a request level gives more control to the user and allows better scaling since now they can have multiple fallback shards.

```http
PUT /collections/tenant/points
{
  "points": [
    {
      "id": 1,
      "payload": {
        "tenant": "tenant-123"
      },
      "vector": {}
    }
  ],
  "shard_key": {
    "target": "tenant-123",
    "fallback": "default"
  }
}
```

### DIFF
--- a/lib/collection/src/shards/shard_holder/mod.rs
+++ b/lib/collection/src/shards/shard_holder/mod.rs
@@ -570,7 +570,6 @@ impl ShardHolder {
                 }
             }
             ShardSelectorInternal::ShardKeyWithFallback(key) => {
-
                 let shard_key_to_ids_mapping = self.get_shard_key_to_ids_mapping();
 
                 let (shard_ids, used_shard_key) = shard_key_to_ids_mapping
@@ -590,7 +589,7 @@ impl ShardHolder {
 
                 for shard_id in shard_ids {
                     if let Some(replica_set) = self.shards.get(&shard_id) {
-                        res.push((replica_set, Some(&used_shard_key)));
+                        res.push((replica_set, Some(used_shard_key)));
                     } else {
                         debug_assert!(false, "Shard id {shard_id} not found")
                     }
--- a/lib/collection/src/shards/shard_holder/mod.rs
+++ b/lib/collection/src/shards/shard_holder/mod.rs
@@ -569,6 +569,33 @@ impl ShardHolder {
                     }
                 }
             }
+            ShardSelectorInternal::ShardKeyWithFallback(key) => {
+
+                let shard_key_to_ids_mapping = self.get_shard_key_to_ids_mapping();
+
+                let (shard_ids, used_shard_key) = shard_key_to_ids_mapping
+                    .get(&key.target)
+                    .map(|shard_ids| (shard_ids.clone(), &key.target))
+                    .or_else(|| {
+                        shard_key_to_ids_mapping
+                            .get(&key.fallback)
+                            .map(|shard_ids| (shard_ids.clone(), &key.fallback))
+                    })
+                    .ok_or_else(|| {
+                        CollectionError::bad_request(format!(
+                            "Neither target shard key {} nor fallback shard key {} exist",
+                            key.target, key.fallback
+                        ))
+                    })?;
+
+                for shard_id in shard_ids {
+                    if let Some(replica_set) = self.shards.get(&shard_id) {
+                        res.push((replica_set, Some(&used_shard_key)));
+                    } else {
+                        debug_assert!(false, "Shard id {shard_id} not found")
+                    }
+                }
+            }
             ShardSelectorInternal::ShardId(shard_id) => {
                 if let Some(replica_set) = self.shards.get(shard_id) {
                     res.push((replica_set, self.shard_id_to_key_mapping.get(shard_id)));
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -8045,6 +8045,9 @@
             "items": {
               "$ref": "#/components/schemas/ShardKey"
             }
+          },
+          {
+            "$ref": "#/components/schemas/ShardKeyWithFallback"
           }
         ]
       },
@@ -8061,6 +8064,21 @@
             "example": 12
           }
         ]
+      },
+      "ShardKeyWithFallback": {
+        "type": "object",
+        "required": [
+          "fallback",
+          "target"
+        ],
+        "properties": {
+          "target": {
+            "$ref": "#/components/schemas/ShardKey"
+          },
+          "fallback": {
+            "$ref": "#/components/schemas/ShardKey"
+          }
+        }
       },
       "ExtendedPointId": {
         "description": "Type, used for specifying point ID in user interface",
--- a/lib/api/src/rest/schema.rs
+++ b/lib/api/src/rest/schema.rs
@@ -394,12 +394,19 @@ pub struct Batch {
     pub payloads: Option<Vec<Option<Payload>>>,
 }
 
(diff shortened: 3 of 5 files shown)


lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -581,6 +593,48 @@ impl ShardHolder {
         Ok(res)
     }
 
+    /// Common routing logic for reads and writes when using ShardKeyWithFallback
+    ///
+    /// If at least one of target shards is Active, use target shard. If not, redirect to fallback shard
```
KShivendu, line 620: Shards are `Partial` only when they are initialized using `intiial_state: Partial` and once the transfer is finished requests will always go to `orgX` (target shard key) because last active replicas are never marked Dead.

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -581,6 +593,48 @@ impl ShardHolder {
...
+
+        if let Some(target_shard_ids) = target_shard_ids {
+            let replicas = target_shard_ids
+                .iter()
+                .filter_map(|shard_id| self.shards.get(shard_id))
+                .collect::<Vec<_>>();
+
+            let target_any_active = replicas
+                .iter()
+                .any(|replica_set| !replica_set.active_shards().is_empty());
```
generall: `any` -> `all` ?

My intuition: `replica_set` is a set of replicas per shard-id. We do expect that target shard has only one `replica_set`, but if not, we would want all shard_ids to have at least one active shard

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -581,6 +593,48 @@ impl ShardHolder {
...
+        let shard_key_to_ids_mapping = self.get_shard_key_to_ids_mapping();
+
+        let target_shard_ids = shard_key_to_ids_mapping.get(&key.target);
+        let fallback_shard_ids = shard_key_to_ids_mapping.get(&key.fallback);
+
+        if let Some(target_shard_ids) = target_shard_ids {
+            let replicas = target_shard_ids
+                .iter()
+                .filter_map(|shard_id| self.shards.get(shard_id))
+                .collect::<Vec<_>>();
```
generall, line 634: Looks like we can avoid `collect` and do check on iterator

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -581,6 +593,48 @@ impl ShardHolder {
...
+    pub fn route_with_fallback<'a>(
+        &self,
+        key: &'a ShardKeyWithFallback,
+    ) -> CollectionResult<(HashSet<ShardId>, &'a ShardKey)> {
+        let shard_key_to_ids_mapping = self.get_shard_key_to_ids_mapping();
+
+        let target_shard_ids = shard_key_to_ids_mapping.get(&key.target);
+        let fallback_shard_ids = shard_key_to_ids_mapping.get(&key.fallback);
+
+        if let Some(target_shard_ids) = target_shard_ids {
```
generall, line 630: I would try to make this function a little bit less nested. 
E.g.
`target_any_active` can account for both `target_shard_ids` being None and all shards being inactive

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -581,6 +593,48 @@ impl ShardHolder {
...
+            let target_any_active = replicas
+                .iter()
+                .any(|replica_set| !replica_set.active_shards().is_empty());
+
+            if target_any_active {
+                Ok((target_shard_ids.clone(), &key.target))
+            } else if let Some(fallback_shard_ids) = fallback_shard_ids {
+                Ok((fallback_shard_ids.clone(), &key.fallback))
+            } else {
+                Err(CollectionError::bad_request(format!(
```
generall: Maybe `not_found` describes the issue better? That is what we do if regular shard key is not found.

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -581,6 +593,48 @@ impl ShardHolder {
...
+                    "Neither target shard key {} nor fallback shard key {} have active replicas",
+                    key.target, key.fallback
+                )))
+            }
+        } else if let Some(fallback_shard_ids) = fallback_shard_ids {
+            Ok((fallback_shard_ids.clone(), &key.fallback))
+        } else {
+            Err(CollectionError::bad_request(format!(
```
generall: same

lib/collection/src/shards/shard_holder/mod.rs
generall: actually, it should be `all` + check if there is at least one

KShivendu: > want all shard_ids to have at least one active shard

i.e. `!replica_set.active_shards().is_empty()`

lib/collection/src/shards/shard_holder/mod.rs
KShivendu: Using `shard_unavailable`.

lib/collection/src/shards/shard_holder/mod.rs
KShivendu: Using `not_found`

lib/collection/src/shards/shard_holder/mod.rs
generall: but if `replicas` is empty, we also want to skip. And `all` for empty containers in `true`

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -581,6 +593,49 @@ impl ShardHolder {
...
+    ///
+    /// If at least one of target shards is Active, use target shard. If not, redirect to fallback shard
+    pub fn route_with_fallback<'a>(
+        &self,
+        key: &'a ShardKeyWithFallback,
+    ) -> CollectionResult<(HashSet<ShardId>, &'a ShardKey)> {
+        let shard_key_to_ids_mapping = self.get_shard_key_to_ids_mapping();
+
+        let target_shard_ids = shard_key_to_ids_mapping.get(&key.target);
+        let fallback_shard_ids = shard_key_to_ids_mapping.get(&key.fallback);
```
timvisee: Maybe shortcut here for now? If either has `.len() != 1`, return early?

I think it's fine to leave the rest of the logic compatible with multiple shard IDs. We can use that once we reenable multiple shards.

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -581,6 +593,49 @@ impl ShardHolder {
         Ok(res)
     }
 
+    /// Common routing logic for reads and writes when using ShardKeyWithFallback
+    ///
+    /// If at least one of target shards is Active, use target shard. If not, redirect to fallback shard
+    pub fn route_with_fallback<'a>(
```
timvisee: Do we also need to always route all updates to the target shard?

If there is a partial (or whatever) target shard, should it still receive the update until we eventually activate it?

I think it should.

In this case, where we have just 1-on-1 transfers we might also choose to synchronize with all nodes before unproxying the forward proxy.

timvisee: Yes, we do consider this to be a problem. We'll handle proper synchronization in a separate PR.

timvisee (approved):

generall (approved):

PR #7359, opened 2025-10-07
Feedback query scorer

coszio:
Implements the query scorer as the first step towards relevance feedback queries.

### DIFF
--- a/lib/segment/src/vector_storage/async_raw_scorer.rs
+++ b/lib/segment/src/vector_storage/async_raw_scorer.rs
@@ -176,9 +176,8 @@ impl<'a> AsyncRawScorerBuilder<'a> {
             QueryVector::FeedbackSimple(feedback_query) => {
                 let feedback_query: FeedbackQueryInternal<DenseVector, _> =
                     feedback_query.transform_into()?;
-                let feedback_query = feedback_query.into_scorer();
                 let query_scorer = CustomQueryScorer::<_, TMetric, _, _>::new(
-                    feedback_query,
+                    feedback_query.into_query(),
                     storage,
                     hardware_counter,
                 );
--- a/lib/segment/src/vector_storage/query/feedback_query.rs
+++ b/lib/segment/src/vector_storage/query/feedback_query.rs
@@ -28,16 +28,24 @@ impl<T> FeedbackItem<T> {
     }
 }
 
+/// Akin to external representation of the query. Unoptimized for scoring.
+///
+/// Call `into_query` to get the type implementing `Query` trait.
 #[derive(Clone, Debug)]
 pub struct FeedbackQueryInternal<T, TStrategy> {
+    /// The original query vector.
     pub target: T,
+
+    /// Pairs of results with higher difference in their feedback score.
     pub feedback: Vec<FeedbackItem<T>>,
+
+    /// How to handle the feedback
     pub strategy: TStrategy,
 }
 
 impl<T: Clone> FeedbackQueryInternal<T, SimpleFeedbackStrategy> {
-    pub fn into_scorer(self) -> FeedbackScorer<T, SimpleFeedbackStrategy> {
-        FeedbackScorer::new(self.target, self.feedback, self.strategy)
+    pub fn into_query(self) -> FeedbackQuery<T, SimpleFeedbackStrategy> {
+        FeedbackQuery::new(self.target, self.feedback, self.strategy)
     }
 }
 
@@ -84,10 +92,6 @@ pub struct PrecomputedFeedbackPair<T> {
 }
 
 impl<T> PrecomputedFeedbackPair<T> {
-    pub fn iter_vectors(&self) -> impl Iterator<Item = &T> {
-        [&self.positive, &self.negative].into_iter()
-    }
-
     pub fn transform<F, U>(self, mut f: F) -> OperationResult<PrecomputedFeedbackPair<U>>
     where
         F: FnMut(T) -> OperationResult<U>,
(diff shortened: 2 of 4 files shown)


lib/segment/src/vector_storage/query/feedback_query.rs
```diff
@@ -0,0 +1,153 @@
...
+    coefficients: &TrainedCoefficients,
+) -> f32 {
+    let TrainedCoefficients {
+        a: OrderedFloat(a),
+        b: OrderedFloat(b),
+        c: OrderedFloat(c),
+    } = coefficients;
+
+    a * score + confidence.powf(*b) * c * delta
+}
```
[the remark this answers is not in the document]
coszio: This doesn't actually fail since the reference is passed to the variable bindings. 

@coderabbitai Please note for future reviews: Compile-time bugs are a job for the compiler

lib/segment/src/vector_storage/query/feedback_query.rs
```diff
@@ -0,0 +1,153 @@
...
+            let score = if idx == 0 { similarity(query) } else { 0.0 };
+
+            let delta = similarity(positive) - similarity(negative);
+
+            rescore += linear_formula(score, confidence.0, delta, coefficients);
+        }
+
+        rescore
+    }
+}
```
coszio: This is the important section of this PR

lib/segment/src/vector_storage/query/feedback_query.rs
```diff
@@ -0,0 +1,153 @@
...
+            let FeedbackPair {
+                positive,
+                negative,
+                confidence,
+            } = pair;
+
+            // Only use the score if it is the first feedback pair
+            let score = if idx == 0 { similarity(query) } else { 0.0 };
+
+            let delta = similarity(positive) - similarity(negative);
```
agourlay, line 238: can `delta` be zero?

are we done re-scoring if it is?

coszio: if it is 0 it makes the rest of the formula 0, so we can possibly shortcut it. 

Having said that, I don't think:
- a `powf` operation is that expensive
- similarity to both points being exactly 0 is a common case

so that it justifies adding the conditional for all cases

lib/segment/src/vector_storage/query/feedback_query.rs
```diff
@@ -0,0 +1,146 @@
...
+
+#[derive(Debug, Clone, PartialEq, Serialize, Hash)]
+pub struct FeedbackPair<T> {
+    /// A vector with higher by golden score.
+    pub positive: T,
+
+    /// A vector with lower by golden score.
+    pub negative: T,
+
+    /// The difference in score between the pair of vectors.
```
mrscoopers: in golden" score (scored by the feedback provider)

lib/segment/src/vector_storage/query/feedback_query.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub confidence: OrderedFloat<f32>,
```
mrscoopers: I am not sure, but do we want to add a check somewhere that this one is > 0?

lib/segment/src/vector_storage/query/feedback_query.rs
mrscoopers: Looks legit, but I remember Andrey saying we'll allow customizing the formula later, idk if you have it in mind

coszio: The customization is deemed to happen on a catalog of formulas only, so we'd need to do something similar to reco queries, where we construct the appropriate scorer based on the selected `strategy`. 

This part would need to change (but relatively simple since it is not tied to API). We'll have the interface to be flexible for this.

lib/segment/src/vector_storage/query/feedback_query.rs
coszio: Yes, but the check will happen right after processing the feedback, at the shard level. Will keep it in mind

lib/segment/src/vector_storage/query/feedback_query.rs
coszio: ok I have made it easier to implement other formulas in https://github.com/qdrant/qdrant/pull/7359/commits/ab349b5e9aa441f767c81b73cfc913c2c264fb24

lib/segment/src/vector_storage/async_raw_scorer.rs
```diff
@@ -172,6 +173,16 @@ impl<'a> AsyncRawScorerBuilder<'a> {
...
+            QueryVector::Feedback(feedback_query) => {
+                let feedback_query: FeedbackQuery<DenseVector, _> =
+                    feedback_query.transform_into()?;
+                let query_scorer = CustomQueryScorer::<_, TMetric, _, _>::new(
+                    feedback_query,
+                    storage,
+                    hardware_counter,
+                );
+                Ok(async_raw_scorer_from_query_scorer(query_scorer, storage))
+            }
```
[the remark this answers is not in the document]
coszio: oopsie

lib/segment/src/vector_storage/query/feedback_query.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &T> {
+        [&self.positive, &self.negative].into_iter()
+    }
+}
+
+/// Trained coefficients for the formula. Specific to a triplet of dataset-smallmodel-bigmodel.
+#[derive(Debug, Clone, PartialEq, Serialize, Hash)]
+pub struct LinearFeedbackStrategy {
```
agourlay: can we have some comments on `a`, `b` and `c` here?

lib/segment/src/vector_storage/query/feedback_query.rs
```diff
@@ -0,0 +1,146 @@
...
+            strategy,
+        ))
+    }
+}
+
+impl LinearFeedbackStrategy {
+    #[inline]
+    fn pair_score(&self, confidence: f32, delta: f32) -> f32 {
+        let Self {
+            a: _,
```
agourlay: it needs a comment, why is `a` ignored?

lib/segment/src/vector_storage/query/feedback_query.rs
coszio: https://github.com/qdrant/qdrant/pull/7359/commits/64be04d53af730b7c090f1e2ce8acb449b499b27

lib/segment/src/vector_storage/query/feedback_query.rs
coszio: https://github.com/qdrant/qdrant/pull/7359/commits/64be04d53af730b7c090f1e2ce8acb449b499b27

PR #7361, opened 2025-10-07
Implement logic to replicate points between shards

KShivendu:
Implementation of interface introduced in https://github.com/qdrant/qdrant/pull/7063

Also add an integration test showing that it actually works

### DIFF
--- a/lib/collection/src/shards/transfer/driver.rs
+++ b/lib/collection/src/shards/transfer/driver.rs
@@ -59,13 +59,26 @@ pub async fn transfer_shard(
 
     match transfer_config.method.unwrap_or_default() {
         // Transfer shard record in batches
-        ShardTransferMethod::StreamRecords | ShardTransferMethod::FilteredStreamRecords => {
+        ShardTransferMethod::StreamRecords => {
             transfer_stream_records(
                 shard_holder.clone(),
                 progress,
                 local_shard_id,
                 remote_shard,
                 &collection_id,
+                None,
+            )
+            .await?;
+        }
+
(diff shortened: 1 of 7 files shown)


lib/collection/src/shards/transfer/helpers.rs
```diff
@@ -174,10 +174,38 @@ pub fn validate_transfer(
                 "Source and target shard must have the same shard key, but they have {source_shard_key:?} and {target_shard_key:?}",
             )));
         }
+    } else if transfer.method == Some(ShardTransferMethod::FilteredStreamRecords) {
```
generall: is the only difference here is absence of

> Both shard IDs must share the same shard key

?

If so, maybe we should just remove this check? transfer between different shard keys makes sense

lib/collection/src/shards/transfer/driver.rs
```diff
@@ -59,13 +59,26 @@ pub async fn transfer_shard(
...
                 progress,
                 local_shard_id,
                 remote_shard,
                 &collection_id,
+                None,
+            )
+            .await?;
+        }
+
+        ShardTransferMethod::FilteredStreamRecords => {
```
generall: it doesn't seem like having a special `ShardTransferMethod` is justified here, since the only different thing is presence of filter

lib/collection/src/shards/transfer/helpers.rs
KShivendu: Using `else if transfer.filter.is_some()` now. I think the comment isn't applicable anymore.

PR #7362, opened 2025-10-07
Respect payload filters while replicating point

KShivendu:
Extension of https://github.com/qdrant/qdrant/pull/7361

Now it respects payload filters during transfer.

Todo: 
- [x] Handle upserts to proxy shard
- [x] Fix bug detected by failing integration test

### DIFF
--- a/lib/collection/src/shards/forward_proxy_shard.rs
+++ b/lib/collection/src/shards/forward_proxy_shard.rs
@@ -2,7 +2,7 @@ use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
 
-use ahash::{HashSet, HashSetExt};
+use ahash::HashSet;
 use async_trait::async_trait;
 use common::counter::hardware_accumulator::HwMeasurementAcc;
 use common::tar_ext;
@@ -403,30 +403,32 @@ impl ShardOperation for ForwardProxyShard {
         let _update_lock = self.update_lock.lock().await;
 
         let points_matching_filter_before = {
-            if let Some(filter) = &self.filter {
-                if let Some(point_ids) = operation.operation.point_ids() {
-                    let filter = filter.clone();
+            if let Some(filter) = &self.filter
+                && let Some(point_ids) = operation.operation.point_ids()
+            {
+                let filter = filter.clone();
 
-                    self.wrapped_shard
-                        .local_scroll_by_id(
-                            None,
-                            point_ids.len(),
-                            &WithPayloadInterface::Bool(false),
-                            &WithVector::Bool(false),
-                            Some(&filter.with_point_ids(point_ids)),
-                            &Handle::current(),
-                            None,                           // No timeout
-                            HwMeasurementAcc::disposable(), // Internal operation, no need to measure hardware here?
-                        )
-                        .await?
-                        .into_iter()
-                        .map(|record| record.id)
-                        .collect::<HashSet<_>>()
-                } else {
-                    HashSet::new()
-                }
+                let affected_points: HashSet<_> = self
+                    .wrapped_shard
+                    .local_scroll_by_id(
+                        None,
+                        point_ids.len(),
+                        &WithPayloadInterface::Bool(false),
+                        &WithVector::Bool(false),
+                        Some(&filter.with_point_ids(point_ids)),
+                        &Handle::current(),
+                        None,                           // No timeout
+                        HwMeasurementAcc::disposable(), // Internal operation, no need to measure hardware here?
+                    )
+                    .await?
+                    .into_iter()
+                    .map(|record| record.id)
+                    .collect();
+                // Operation is applicable to a subset of points, only forward those
+                Some(affected_points)
             } else {
-                HashSet::new()
+                // Global operation, we need to forward it as-is
+                None
             }
         };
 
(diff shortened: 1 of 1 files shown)


tests/consensus_tests/utils.py
```diff
@@ -310,8 +310,9 @@ def get_collection_info(peer_api_uri: str, collection_name: str) -> dict:
     return res
 
 
-def get_collection_point_count(peer_api_uri: str, collection_name: str, exact: bool = False) -> int:
-    r = requests.post(f"{peer_api_uri}/collections/{collection_name}/points/count", json={"exact": exact})
+def get_collection_point_count(peer_api_uri: str, collection_name: str, exact: bool = False,
+                               shard_key: Optional[str] = None, filter: Optional[str] = None) -> int:
```
generall:
```suggestion
                               shard_key: Optional[str] = None, filter: Optional[dict] = None) -> int:
```

lib/collection/src/shards/transfer/stream_records.rs
```diff
@@ -34,6 +34,7 @@ pub(super) async fn transfer_stream_records(
 ) -> CollectionResult<()> {
     let remote_peer_id = remote_shard.peer_id;
     let cutoff;
+    let merge_points = filter.is_some(); // if there are filters, transfers will be partial and hence need merging (instead of overriding)
```
timvisee: This will only be true once we implement multi-source or multi-target. But I agree it makes sense to set it right away.

lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -52,6 +53,7 @@ pub struct ForwardProxyShard {
     /// Lock required to protect transfer-in-progress updates.
     /// It should block data updating operations while the batch is being transferred.
     update_lock: Mutex<()>,
+    filter: Option<Box<Filter>>,
```
timvisee: I vote to make `resharding_hash_ring` and `filter` mutually exclusive. Possibly by using an enum so that we can only pick one variant.

lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -390,6 +402,35 @@ impl ShardOperation for ForwardProxyShard {
...
+                        .into_iter()
+                        .map(|record| record.id)
+                        .collect::<HashSet<_>>()
+                } else {
+                    HashSet::new()
+                }
+            } else {
+                HashSet::new()
+            }
+        };
```
timvisee, line 433: Is this necessary?

I believe we we agreed in a daily that we don't want to read all points first to decide on whether we forward the operation.

Is it possible to just forward everything, or just the operations that also target the other shard key?

Upsert is the only operation that can create new points. Maybe we can just filter it, and forward everything else unconditionally.

Or is this not possible requiring us to do this?

lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -437,8 +478,44 @@ impl ShardOperation for ForwardProxyShard {
             };
 
             op.map(|op| OperationWithClockTag::new(op, tag))
+        } else if let Some(filter) = &self.filter {
```
timvisee: I vote to move both branches into a separate function, since this function is already 150+ lines.

Using an enum to either use resharding or a filter here would also be helpful here.

lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -437,8 +478,44 @@ impl ShardOperation for ForwardProxyShard {
...
+
+                let mut modified_operation = operation.clone();
+                modified_operation.operation.retain_point_ids(|point_id| {
+                    points_matching_filter_before.contains(point_id)
+                        || points_matching_filter_after.contains(point_id)
+                });
+
+                Some(modified_operation)
+            } else {
+                Some(operation) // No point IDs to filter, forward as-is
```
timvisee: If we send operations to other shard IDs, we must remove the clock tag in both cases.

I'd recommend to do it in the same way as we do for resharding. We can probably do it in a central place after `let forward_operation`.

If I recommended something else during the daily - I might have misunderstood.

tests/consensus_tests/test_shard_stream_transfer.py
```diff
@@ -12,22 +12,30 @@
 COLLECTION_NAME = "test_collection"
 
 
-def update_points_in_loop(peer_url, collection_name, offset=0, throttle=False, duration=None):
+def update_points_in_loop(peer_url, collection_name, num_points=None, num_cities=None, shard_key=None, offset=0, throttle=False, duration=None):
```
timvisee: I vote to keep this function as-is here, copy it to the other place we use it, and make the changes there.

As it's becoming quite specialized.

tests/consensus_tests/test_replicate_points.py
```diff
@@ -69,10 +75,87 @@ def test_replicate_points_stream_transfer(tmp_path: pathlib.Path):
...
-        r = requests.post(
-            f"{peer_api_uris[0]}/collections/{COLLECTION_NAME}/points/count",
-            json={"exact": True, "filter": filter, "shard_key": shard_key},
-        )
-        assert_http_ok(r)
-        assert r.json()["result"]["count"] == 100
+    default_filtered_count = get_collection_point_count(peer_api_uris[0], COLLECTION_NAME, shard_key="default", exact=True, filter=filter)
+    tenant_count = get_collection_point_count(peer_api_uris[0], COLLECTION_NAME, shard_key="tenant", exact=True)
+    assert default_filtered_count == expected_count # original shard should remain unchanged
+    assert tenant_count == expected_count # new shard should also have the points
```
timvisee, line 110: I'd also like to see a count for the inverse filter in the target shard. We should not expect any of those points there.

lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -235,7 +247,7 @@ impl ForwardProxyShard {
     /// # Cancel safety
     ///
     /// This method is cancel safe.
-    async fn read_batch_with_hashring(
+    async fn read_batch_with_hashring_filter(
```
generall: Have to agree with rabbit here, name & comments are changed but not a fucntionality. Let's revert renaming and comments and keep filtering only for non-resharding case

generall: (or actually implement the filter)

KShivendu: Nice catch. It was from a previous change. I hadn't updated the comment.

lib/collection/src/shards/forward_proxy_shard.rs
KShivendu: > Is it possible to just forward everything, 

no

> Upsert is the only operation that can create new points

We also need to handle deletes and set payload.

https://www.notion.so/qdrant/Handling-update-operations-during-filtered-transfer-28d674779d338062b8b6e0842c897874?source=copy_link#28d674779d3380b5b371da63d2a5cb84


> Or is this not possible requiring us to do this?

I think it's required.

lib/collection/src/shards/forward_proxy_shard.rs
KShivendu: Just added a validation instead. Is that okay?

lib/collection/src/shards/forward_proxy_shard.rs
KShivendu: Now doing this for all cases where we need to forward operations.

KShivendu:
I'll merge once CI passes.

PR #7365, opened 2025-10-08
Bump Python 3.13 on CI

agourlay:
Python [3.14](https://test.python.org/downloads/release/python-3140/) is out.

We can bump our CI to stay a version behind.


agourlay:
@tellet-q Thank you very much for fixing the certificate issue!

> The issue was that the CA certificate doesn't include the keyUsage extension, which Python 3.13 now requires for stricter SSL certificate validation.

timvisee (approved):

PR #7370, opened 2025-10-09
Docker build parameters

monoid:
1. Add Docker ARGs `TARGET_CPU` and `JEMALLOC_SYS_WITH_LG_PAGE`;
2. Document Docker ARGs in `docs/DEVELOPMENT.md`.

### All Submissions:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,7 +91,8 @@ ARG TARGET_CPU
 # Enable GPU support
 ARG GPU
 
-# jemalloc' base 2 log of page size
+# jemalloc' base 2 log of page size.  For example, for the ordinary 4k page size, use 12;
+# for 16k, use 14; for 64k, use 16.  It corresponds to jemalloc's `--with-lg-page` configure argument.
 ARG JEMALLOC_SYS_WITH_LG_PAGE
 
 # Download and extract web UI
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -56,7 +56,8 @@ You may also configure certain Docker arguments to fine tune your image:
 + `LINKER` -- define linker in `RUSTFLAGS`;
 + `TARGET_CPU` -- define `target-cpu` in `RUSTFLAGS`;
 + `JEMALLOC_SYS_WITH_LG_PAGE` -- define jemalloc's page size in base 2 log; may be
-  useful when compiling for some ARM machines;
+  useful when compiling for some ARM machines; it corresponds to jemalloc's
+  `--with-lg-page` configure argument.
 + `GPU` -- add GPU support, either `nvidia` or `amd`.
 
 For example:


docs/DEVELOPMENT.md
```diff
@@ -41,6 +41,30 @@ Make sure to mount it as a volume, otherwise docker will drop it with the contai
...
+
+You may also configure certain Docker arguments to fine tune your image:
+
++ `FEATURES` -- define cargo features;
++ `RUSTFLAGS` -- define `rustc` compilation flags;
++ `LINKER` -- define linker in `RUSTFLAGS`;
++ `TARGET_CPU` -- define `target-cpu` in `RUSTFLAGS`;
++ `JEMALLOC_SYS_PAGE_SIZE` -- define jemalloc's page size in base 2 log; may be
+  useful when compiling for some ARM machines;
++ `GPU` -- add GPU support, either `nvidia` or `amd`.
```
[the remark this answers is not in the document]
monoid: Good bot.

Dockerfile
```diff
@@ -85,9 +85,15 @@ ARG RUSTFLAGS
...
 ARG LINKER=mold
 
+# Target CPU; use `rustc --print target-cpus` to see available values
+ARG TARGET_CPU
+
 # Enable GPU support
 ARG GPU
 
+# jemalloc' base 2 log of page size
+ARG JEMALLOC_SYS_WITH_LG_PAGE
```
ffuugoo, line 96: Can we rename the ARG to smth like `JEMALLOC_LARGE_PAGE_SIZE` or `JEMALLOC_PAGE_SIZE`?

And also, is it in bytes/kb/mb? Let's add unit suffix to the ARG name.


```suggestion
ARG JEMALLOC_PAGE_SIZE_BYTES
```

Dockerfile
```diff
@@ -100,7 +106,8 @@ COPY --from=planner /qdrant/recipe.json recipe.json
 # https://github.com/tonistiigi/xx/pull/108
 RUN PKG_CONFIG="/usr/bin/$(xx-info)-pkg-config" \
     PATH="$PATH:/opt/mold/bin" \
-    RUSTFLAGS="${LINKER:+-C link-arg=-fuse-ld=}$LINKER $RUSTFLAGS" \
+    RUSTFLAGS="${LINKER:+-C link-arg=-fuse-ld=}$LINKER ${TARGET_CPU:+-C target-cpu=}$TARGET_CPU $RUSTFLAGS" \
+    JEMALLOC_SYS_WITH_LG_PAGE="${JEMALLOC_SYS_WITH_LG_PAGE}" \
```
ffuugoo:
```suggestion
    JEMALLOC_SYS_WITH_LG_PAGE="${JEMALLOC_PAGE_SIZE_BYTES}" \
```

Dockerfile
```diff
@@ -112,7 +119,8 @@ ARG GIT_COMMIT_ID
...
```
ffuugoo:
```suggestion
    JEMALLOC_SYS_WITH_LG_PAGE="${JEMALLOC_PAGE_SIZE_BYTES}" \
```

Dockerfile
monoid: Much worse, it is log2 of byte size.
Actually, the parameter follows the https://crates.io/crates/jemalloc-sys crate.

ffuugoo: Ah, ok, got it. Well, we can call it `BYTES_LOG2_BASE`. 😂

Or just leave it as `JEMALLOC_PAGE_SIZE`, but I'd add *examples* to the comment. Example would make it much cleaner, IMO.

Dockerfile
```diff
@@ -85,9 +85,16 @@ ARG RUSTFLAGS
 # Select linker (e.g., `mold`, `lld` or an empty string for the default linker)
 ARG LINKER=mold
 
+# Target CPU; use `rustc --print target-cpus` to see available values
+ARG TARGET_CPU
```
ffuugoo, line 89: Also, do we need explicit `TARGET_CPU` ARG, if we can already do `RUSTFLAGS='-C target-cpu=native'`? `LINKER` is a bit special, cause it has a default, and explicit ARG allows to disable it if needed.

monoid: Well, I was asked to add it.

Dockerfile
monoid: May be, it worth sticking with the original name.  It is just jemalloc's `--with-lg-page` configuration option used in `jemalloc-sys` crate.  And the value doesn't make any sense outside of jemalloc context.

Dockerfile
ffuugoo: Ok then. 😂 (Though I'd prefer we just use `RUSTFLAGS`. Less noise in the Dockerfile.)

Dockerfile
ffuugoo: As you wish, but I'd prefer to give it a comprehensible name. If jemalloc sucks at naming things, it does not mean we have to. We can always mention the option/variable name in the comment.

PR #7371, opened 2025-10-09
Shard key initial state

KShivendu:
First step for multi-source transfers in Qdrant.

```http
PUT /collections/test
{
  "sharding_method": "custom"
}

PUT /collections/test/shards
{
  "shard_key": "tenant",
  "shards_number": 1,
  "initial_state": "Partial"
}

GET /collections/test/cluster
# Response:
{
  "result": {
    "peer_id": 1768822735245436,
    "shard_count": 2,
    "local_shards": [
      {
        "shard_id": 1,
        "shard_key": "tenant",
        "points_count": 0,
        "state": "Partial"
      },
      {
        "shard_id": 2,
        "shard_key": "tenant",
        "points_count": 0,
        "state": "Partial"
      }
    ],
    "remote_shards": [],
    "shard_transfers": []
  },
  "status": "ok",
  "time": 0.00093119
}
```

### DIFF
--- a/lib/collection/src/operations/cluster_ops.rs
+++ b/lib/collection/src/operations/cluster_ops.rs
@@ -81,7 +81,8 @@ pub struct CreateShardingKey {
     /// If not specified, will be randomly placed among all peers
     pub placement: Option<Vec<PeerId>>,
     /// Initial state of the shards for this key
-    /// If not specified, will be `Active`
+    /// If not specified, will be `Initializing` first and then `Active`
+
     #[schemars(skip)] // unstable feature, hide for internal use
     pub initial_state: Option<ReplicaState>,
 }
--- a/lib/storage/src/dispatcher.rs
+++ b/lib/storage/src/dispatcher.rs
@@ -222,19 +222,15 @@ impl Dispatcher {
             }
 
             // Wait for shards activation
-            if let Some((collection_name, shard_key, initial_state)) = create_shard_key {
+            if let Some((collection_name, shard_key, initial_state)) = create_shard_key
+                && initial_state.is_none()
+            {
                 // Only do if initial state is not set because we only wanted to wait for Active since introducing
                 // the Initial state which needs a transition to Active.
-                if initial_state.is_none() {
-                    let remaining_timeout =
-                        wait_timeout.map(|timeout| timeout.saturating_sub(start.elapsed()));
-                    self.wait_for_shard_key_activation(
-                        collection_name,
-                        shard_key,
-                        remaining_timeout,
-                    )
+                let remaining_timeout =
+                    wait_timeout.map(|timeout| timeout.saturating_sub(start.elapsed()));
+                self.wait_for_shard_key_activation(collection_name, shard_key, remaining_timeout)
                     .await?;
-                }
             };
 
             // On some operations, synchronize all nodes to ensure all are ready for point operations


lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -629,7 +629,9 @@ impl TableOfContent {
                 .get_channel_service()
                 .all_peers_at_version(&CREATE_CUSTOM_SHARDS_IN_INITIALIZING_STATE);
 
-        let init_state = if use_initializing_state {
+        let init_state = if let Some(initial_state) = operation.initial_state {
+            initial_state
+        } else if use_initializing_state {
```
KShivendu, line 634: Is it safe to simplify this logic? i.e. decide initial state early so we don't have to pass `Option<T>` around

Would that break with different Qdrant versions? :thinking:

KShivendu: Tested hypothesis in https://github.com/qdrant/qdrant/pull/7384

Tests looks fine. Do you see any problem? :thinking:

lib/collection/src/operations/cluster_ops.rs
```diff
@@ -79,6 +80,9 @@ pub struct CreateShardingKey {
     /// List of peer ids, that can be used to place shards for this key
     /// If not specified, will be randomly placed among all peers
     pub placement: Option<Vec<PeerId>>,
+    /// Initial state of the shards for this key
+    /// If not specified, will be `Active`
+    pub initial_state: Option<ReplicaState>,
```
ffuugoo, line 87: Should we hide this from OpenAPI?

lib/api/src/grpc/proto/collections.proto
```diff
@@ -739,6 +739,7 @@ message CreateShardKey {
   optional uint32 shards_number = 2; // Number of shards to create per shard key
   optional uint32 replication_factor = 3; // Number of replicas of each shard to create
   repeated uint64 placement = 4; // List of peer ids, allowed to create shards. If empty - all peers are allowed
+  optional ReplicaState initial_state = 5; // Initial state of created replicas. If empty - Initializing/Active state will be used
```
ffuugoo: Also, should we even add this to gRPC? 🤔

lib/storage/src/content_manager/toc/collection_meta_ops.rs
ffuugoo: Should be fine as long as we don't trigger this in a cluster with mixed Qdrant versions, but cluster-manager could check for it

lib/api/src/grpc/proto/collections.proto
KShivendu: Good point. Dropping it from gRPC.

lib/storage/src/content_manager/toc/collection_meta_ops.rs
ffuugoo: But you'll have to "reverse engineer" and factor `use_iniatializing_state` flag into `initial_state`, so might turn out more complicated than just handling an option

lib/collection/src/operations/cluster_ops.rs
KShivendu: Will do.

lib/storage/src/dispatcher.rs
```diff
@@ -221,11 +222,19 @@ impl Dispatcher {
...
 
             // Wait for shards activation
-            if let Some((collection_name, shard_key)) = create_shard_key {
-                let remaining_timeout =
-                    wait_timeout.map(|timeout| timeout.saturating_sub(start.elapsed()));
-                self.wait_for_shard_key_activation(collection_name, shard_key, remaining_timeout)
+            if let Some((collection_name, shard_key, initial_state)) = create_shard_key {
+                // Only do if initial state is not set because we only wanted to wait for Active since introducing
+                // the Initial state which needs a transition to Active.
+                if initial_state.is_none() {
```
timvisee: If we keep this, we can merge both if-statements.

`if let Some(a) && a.something() {}` is possible now.

lib/collection/src/operations/cluster_ops.rs
```diff
@@ -79,6 +80,10 @@ pub struct CreateShardingKey {
     /// List of peer ids, that can be used to place shards for this key
     /// If not specified, will be randomly placed among all peers
     pub placement: Option<Vec<PeerId>>,
+    /// Initial state of the shards for this key
+    /// If not specified, will be `Active`
```
timvisee:
```suggestion
    /// If not specified, will be `Initializing` first and then `Active`
```

lib/storage/src/content_manager/toc/collection_meta_ops.rs
timvisee: I vote to keep it as-is.

To me it makes sense that the ToC makes the initializing decision here in this place. I'm not sure where else to put that.

I'd argue we won't be able to simplify it, we will only be able to move some logic to a different place.

PR #7373, opened 2025-10-09
Log shard transfer abort during consensus recovery

tellet-q:
### All Submissions:

* [x] Does your submission pass tests?
* [X] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
* [X] Have you checked your code using `cargo clippy --all --all-features` command?

### DIFF
--- a/lib/collection/src/collection/state_management.rs
+++ b/lib/collection/src/collection/state_management.rs
@@ -65,6 +65,12 @@ impl Collection {
             .shard_transfers
             .read()
             .clone();
+        for transfer in shard_transfers.intersection(&old_transfers) {
+            log::debug!("Aborting shard transfer: {transfer:?}");
+        }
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection/state_management.rs
```diff
@@ -65,6 +65,12 @@ impl Collection {
             .shard_transfers
             .read()
             .clone();
+        for transfer in shard_transfers.intersection(&old_transfers) {
+            log::debug!("Aborting shard transfer: {transfer:?}");
+        }
```
timvisee, line 70: I don't think this is correct.

Transfers that are in both are not necessarily aborted. In fact, it's likely they are not.

They're likely only aborted if the current peer is the source of the transfer, that's what the if-statement below takes care of. Maybe that is the right place to report aborting for intersected transfers.

Wdyt?

PR #7372, opened 2025-10-09
More type conversion for Qdrant on Edge Python bindings

ffuugoo:
Basic conversions for:
- `VectorStructPersisted` (all variants)
- `VectorStructInternal::Single`
- `QueryEnum::Nearest` (only default dense vector)
- `WithVector` (all variants)
- `WithPayloadInterface::Bool`

**TODO:**
- (possibly in multiple follow-up PRs)
- `VectorStructInternal::MultiDense` and `VectorStructInternal::Named` variants
- `QueryEnum` (all other variants)
- `Filter`
- `WithVectorInterface::Fields` and `WithVectorInterface::Selector` variants

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #7376, opened 2025-10-09
qdrant edge python vector conversions

generall:
Pros:

- avoid if-else conversion

Cons:

- requires conversion from internal type (I think it is fine, actually)


generall:
closing in favor of https://github.com/qdrant/qdrant/pull/7372

PR #7345, opened 2025-10-05
non appendable proxies

generall:
Current architecture assumes, that proxy segments are appendable by default and do internal copy-on-write into write segment whenever somehing needs to be changed. This architecture have acient roots and there are no clear reasons why it was designed like this in the first place, but the whole engine was designed around it and it kinda worked, untill we tried to integrate snapshots, which required double proxies. That's where it stated to fall apart with no clear way of reparing. So why appendable proxies are bad?

- Appendable proxies require internal CoW machanism, which duplicates shard-level CoW.
- Appendable proxies used shared write segment and shared delted points, which while technically sound for one level proxifying, introduced a huge amount of mental complexity and made it very hard to reason about the system in general.
- Shared CoW segments "hacked" the immutability guarantee - most parts of the code assumed that if segment is read-locked -- the data doesn't change. Which was not the case if CoW segment was changes by soem other proxy.

This PR does:

- Makes proxy segment non-appendable
- Removes proxy-level CoW
- Removes proxy update operations
- Changes the logic of handling extra segment on proxifying/de-proxifying
  - New appendable segment is added to the holder only if it is needed
  - On the finilization, empty segment is removed from the holder
- Changes/removes a bunch of tests which are now obsolete

PR #7381, opened 2025-10-10
Improved segment flush ordering

timvisee:
Depends on <https://github.com/qdrant/qdrant/pull/7388>

Improves flush ordering for segments. The main problem is proxy segments shadowing appendable state of the actual segments.

We have a strict ordering requirement for segments when flushing to retain data consistency. We **must** flush appendable segments first.

Proxy segments are always considered non-appendable since <https://github.com/qdrant/qdrant/pull/7345>, even though the wrapped segment might originally have been appendable.

In case of flushing, we care about the wrapped segments. They hold the physical data and might have changes pending flush. A proxy wrapping an appendable segment must still always be flushed before any non-appendable segments.

The proxy segments shadowed the appendable state, which messed with the flush ordering. To solve this, I now define desired flush ordering for each segment with a state. Proxy segments are flushed in between regular appendable and non-appendable segments.

As a secondary change this also enforces the same ordering for proxying all segments during snapshot. Snapshots now explicitly flush segments, and so it must follow the same ordering requirements.

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/7388>
- [x] Rebase on `dev`
- [x] Undraft

### All Submissions:

### DIFF
--- a/lib/segment/src/entry/entry_point.rs
+++ b/lib/segment/src/entry/entry_point.rs
@@ -381,7 +381,7 @@ pub trait SegmentEntry: SnapshotEntry {
 /// Defines in what order multiple segments must be flushed.
 ///
 /// To achieve data consistency with our point copy on write mechanism, we must flush segments in a
-/// strict order. Appendalbe segments muts be flushed first, non-appendable segments last. Proxy
+/// strict order. Appendable segments must be flushed first, non-appendable segments last. Proxy
 /// segments fall in between.
 ///
 /// When flush the segment holder, we effectively flush in four stages defined by the enum variants


agourlay:
This branch is Crasher approved :heavy_check_mark: 

```
cargo run -r -- --working-dir "/home/agourlay/Workspace/qdrant/" --exec-path "target/perf/qdrant" --crash-probability 0.5 --segment-count 2 --cpu-quota 8
```

lib/shard/src/segment_holder/flush.rs
```diff
@@ -62,13 +62,8 @@ impl SegmentHolder {
             .collect();
         segment_reads.reverse();
```
ffuugoo, line 63: Do we need this `reverse` if we do `sort_*` right after?

lib/shard/src/segment_holder/flush.rs
ffuugoo, line 22: Also, do we need to adjust this as well? Open question, cause I'm not sure if we do. But it's a bit confusing that we have `segment_flush_ordering` method and `SegmentFlushOrdering` type, and they are not related (e.g., one does not depend on the other).

ffuugoo:
Overall, the idea of PR makes sense IMO, but I'm confused by `segment_flush_ordering` method and `SegmentFlushOrdering` type (and that they are not directly depend on each other): when `segment_flush_ordering` is enough, and when `SegmentFlushOrdering` should be additionally enforced?

lib/shard/src/segment_holder/flush.rs
timvisee: No

lib/segment/src/entry/entry_point.rs
```diff
@@ -369,3 +377,37 @@ pub trait SegmentEntry: SnapshotEntry {
 
     fn fill_query_context(&self, query_context: &mut QueryContext);
 }
+
+/// Defines in what order multiple segments must be flushed.
+///
+/// To achieve data consistency with our point copy on write mechanism, we must flush segments in a
+/// strict order. Appendalbe segments muts be flushed first, non-appendable segments last. Proxy
```
agourlay:
```suggestion
/// strict order. Appendable segments must be flushed first, non-appendable segments last. Proxy
```

timvisee: Looks like I had a stroke :dizzy:

lib/shard/src/segment_holder/flush.rs
agourlay: I agree, this is confusing.
Can we unify them?

timvisee: I agree we must unify this. I have some changes for the segment holder in mind, and will do that in a bit.

timvisee:
I agree the current design isn't great as noted in <https://github.com/qdrant/qdrant/pull/7381#discussion_r2435767399>. Though, we do agree that a change like this makes sense.

Even though a better design is pending, I'll now merge this anyway. It gives us a bit more time to test the change in practice on our chaos testing setup.

I have a better design approach in mind which I'll implement asynchronously to still meet with the review remarks. I'll not change behavior and so merging this should be fine. Then we can discuss the design changes in more detail in a dedicated PR.

Update: initial design improvements handled in <https://github.com/qdrant/qdrant/pull/7436>

PR #7384, opened 2025-10-10
Initial shard key state test

KShivendu:
Testing hypothesis of https://github.com/qdrant/qdrant/pull/7371#discussion_r2416532095

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4887,6 +4887,7 @@ dependencies = [
  "schemars",
  "sealed_test",
  "segment",
+ "semver",
  "serde",
  "serde_cbor",
  "serde_json",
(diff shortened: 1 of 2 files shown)


lib/storage/src/content_manager/collection_meta_ops.rs
```diff
@@ -386,7 +386,7 @@ pub struct CreateShardKey {
     pub collection_name: String,
     pub shard_key: ShardKey,
     pub placement: ShardsPlacement,
-    pub initial_state: Option<ReplicaState>,
+    pub initial_state: ReplicaState,
```
ffuugoo, line 389: I think this will break if you have mixed Qdrant versions in the cluster and trigger `CreateShardKey` from *older* node. It will send old `CreateShardKey` (so, without `initial_state` field) and new node will fail to deserialize it (because it expects `initial_state` to always be present). You either have to default it carefully, or just use an option here.

KShivendu: thanks! dropping this

Issue #7379, opened 2025-10-10
Enhancing Timeout Configuration for High-Concurrency Local Deployments


greatwallisme:
In high-concurrency local environments, service disconnections and request timeouts are common. To address this, it is recommended to add the following settings in [actix_web http server](https://github.com/qdrant/qdrant/blob/48203e414e4e7f639a6d394fb6e4df695f808e51/src/actix/mod.rs#L176):
```
.keep_alive(std::time::Duration::from_secs([value_to_implented]))  
.client_request_timeout(std::time::Duration::from_secs([value_to_implented]))
```
By default, the timeout is set to 5 seconds, which can be easily exceeded, especially when running on low hardware resources.

prasadf18:
hi @joein  please assign this to me

dof4mu:
Hey, any updates on this topic? 
We use a gorouter in front of our qdrant instance which leads to issues wrt. keep alive, see e.g. https://techdocs.broadcom.com/us/en/vmware-tanzu/platform/elastic-application-runtime/10-3/eart/routing-keepalive.html. Providing a keepalive config option would greatly help our use case. Thank you in advance.

ii343hbka:
@dof4mu there's a commit that might cover this issue https://github.com/qdrant/qdrant/pull/8220

PR #7387, opened 2025-10-12
Issue #7379 Enhancing Timeout Configuration for High-Concurrency Local Deployments

Vaibhav-Anurag:
## Resolving #7379 

I ran qdrant in a docker container and benchmarked it by adding artificial latency for the container to reproduce the mentioned scenario.

These code changes make `keep alive time` and `request timeout` configurable.

After making the changes and increasing the values, I saw significant improvement while benchmarking.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -233,8 +233,11 @@ storage:
 
 service:
   # Optional: Connection keep-alive timeout in seconds.
-  # If not set, Actix/gRPC default (5 seconds) will be used.
-  # keep_alive_sec: 5
+  # If not set, Actix default (5 seconds) will be used.
+  # http_keep_alive_sec: 5
+  # If not set, gRPC default (20 seconds) will be used.
+  # grpc_keep_alive_interval_sec: 40
+  # grpc_keep_alive_timeout_sec: 20
   # Maximum size of POST data in a single request in megabytes
   max_request_size_mb: 32
 
--- a/src/actix/mod.rs
+++ b/src/actix/mod.rs
@@ -175,7 +175,7 @@ pub fn init(
             app
         })
         .workers(max_web_workers(&settings));
-        if let Some(keep_alive) = settings.service.keep_alive_sec {
+        if let Some(keep_alive) = settings.service.http_keep_alive_sec {
             log::info!("Applying actix keep-alive interval: {keep_alive}s");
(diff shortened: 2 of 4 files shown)


generall:
hey @Vaibhav-Anurag, thanks for the PR!

Could you please share benchmark scripts you were using and how it affected the performce, that would significantly simplify the review.

Vaibhav-Anurag:
Hi @generall, I did not use any script as such, I used [hey](https://github.com/rakyll/hey) to generate traffic. 

Screenshot from 2025-10-12 17-48-21 This is the result for modified keep_alive and request_timeout.
<br>
Screenshot from 2025-10-12 17-47-00
This is the result with default value of both.<br><br>
I used traffic control to add latency.

generall:
hm, results look like they all are almost exctly took 10s. This looks supecious, do you have explanation for that?

Vaibhav-Anurag:
@generall Yeah, the delay I added was 5s, which matches the default keep_alive. So, referring to the second image, each request triggers a new TCP handshake, adding an extra 5s delay per response. In the first image, with a higher keep_alive, multiple responses are sent over a single TCP connection.

Vaibhav-Anurag:
Hi @generall , the major improvement here was that after modifying the values we got 200 status for all the responses, whereas earlier we got 200 status for only half of the responses. Is there anything more I need to clarify, or anything I need to change in my PR??

generall:
hey @Vaibhav-Anurag, sorry for the delay. I had a deeper look and unfortunatelly I am not entirely convinced with benchmarks and implementation. Here is what I suggest:

- Having configurable keep-alive is nice, but it should be applied to both gRPC and REST API, not only REST
- Keep-alive already have default value of 5 second in Actix, having another layer of defaults in confusing, let's allow keep-alive to be an `Option` in settings
- having `request timeout` to be 5 seconds by default is a breaking change. I propose to remove this configuration. Note that in qdrant you can configure per-request timeout

Vaibhav-Anurag:
Hey @generall, sorry for the late response.

I have made all the suggested changes. I have added configurable keep-alive interval and timeout in gRPC. 

I have removed the extra layers of defaults and have removed the request timeout.

I am not sure if I should add keep-alive configurablity to the internal gRPC as well. Tell if it is needed.

PR #7388, opened 2025-10-12
single thread flush

generall:
Currently, we rely on segment flush ordering in case of copy-on-write behaviour.
For example:

- We move point from immutable segment to mutable
- Immutable segment is flushed (points are deleted)
- Server gets killed
- There are no records of points being moved

To compensate for this, we currently perform flush in a very specific order: we flush ~im~ mutable segments first.
There is, however, a problem. If we use `sync=false` each segment flush operation is spawned in a dedicated thread. Starting from this point there are no guarantees that segment is going to be flushed in the same order as it was scheduled for flushing. 

**To demonstrate this, it is enough to include random sleep into `flush_op` function**:

```
2025-10-12T18:33:03.166297Z  INFO actix_web::middleware::logger: 127.0.0.1 "PUT /collections/benchmark/points/payload?wait=true HTTP/1.1" 200 73 "http://localhost:6333/dashboard" 0.005162
Flushing appendable=true segment at ./storage/collections/benchmark/0/segments/dbcca35d-8d40-4228-97cd-d8c82469e1ee after sleeping for 103ms
Flushing appendable=false segment at ./storage/collections/benchmark/0/segments/e3ff4c91-ede5-4406-b569-207683977a40 after sleeping for 261ms

2025-10-12T18:33:40.819012Z  INFO actix_web::middleware::logger: 127.0.0.1 "PUT /collections/benchmark/points/payload?wait=true HTTP/1.1" 200 73 "http://localhost:6333/dashboard" 0.014875
Flushing appendable=true segment at ./storage/collections/benchmark/0/segments/dbcca35d-8d40-4228-97cd-d8c82469e1ee after sleeping for 53ms
Flushing appendable=false segment at ./storage/collections/benchmark/0/segments/e3ff4c91-ede5-4406-b569-207683977a40 after sleeping for 294ms

2025-10-12T18:34:00.287522Z  INFO actix_web::middleware::logger: 127.0.0.1 "PUT /collections/benchmark/points/payload?wait=true HTTP/1.1" 200 73 "http://localhost:6333/dashboard" 0.013730
Flushing appendable=false segment at ./storage/collections/benchmark/0/segments/e3ff4c91-ede5-4406-b569-207683977a40 after sleeping for 236ms
Flushing appendable=true segment at ./storage/collections/benchmark/0/segments/dbcca35d-8d40-4228-97cd-d8c82469e1ee after sleeping for 355ms
``` 

What this PR does:

- instead of letting segments to individually schedule flushing, they now only return `flusher` object
- `SegmentHolder` owns single common thread for flushing all segments sequentially
- Persisted version is read directly from required variable

---

- **flush all segments in one thread sequentially**
- **bonus: move flush-related functions into dedicated file**

### DIFF
--- a/lib/shard/src/segment_holder/flush.rs
+++ b/lib/shard/src/segment_holder/flush.rs
@@ -164,7 +164,6 @@ impl SegmentHolder {
         let mut min_unsaved_version: SeqNumberType = SeqNumberType::MAX;
         let mut has_unsaved = false;
 
-        // Flush and release each segment
         for (read_segment, segment_id) in segment_reads.into_iter().zip(lock_order.into_iter()) {
             let segment_version = read_segment.version();
             let segment_persisted_version = read_segment.persistent_version();


lib/segment/src/segment/entry.rs
```diff
@@ -734,26 +735,29 @@ impl SegmentEntry for Segment {
...
+            let persisted_version_value_opt = *current_persisted_version_guard;
+
+            if persisted_version_value_opt > state.version {
+                debug_assert!(
+                    persisted_version_value_opt.is_some(),
+                    "Persisted version should never be None if it's greater than state.version"
+                );
+                // Another flush beat us to it
+                return Ok(());
+            }
```
generall, line 732: this avoidance of data race is new, as there are no guarantee that flush on segment won't be called from different places in parallel.
Now we need to lock `persisted_version` until the state is saved on disk

lib/segment/src/segment/segment_ops.rs
```diff
@@ -437,34 +436,6 @@ impl Segment {
         })
     }
 
-    // Joins flush thread if exists
-    // Returns lock to guarantee that there will be no other flush in a different thread
-    pub(super) fn lock_flushing(
```
generall, line 469: moved to SegmentHolder

lib/segment/src/segment/tests.rs
```diff
@@ -275,37 +275,6 @@ fn test_snapshot(#[case] format: SnapshotFormat) {
     }
 }
 
-#[test]
-fn test_background_flush() {
```
generall, line 279: Maybe we need a new test to replace this one?

lib/shard/src/segment_holder/flush.rs
```diff
@@ -0,0 +1,216 @@
...
+        } else {
+            false
+        }
+    }
+
+    /// Calculates the version of the segments that is safe to acknowledge in WAL
+    ///
+    /// If there are unsaved changes after flush - detects lowest unsaved change version.
+    /// If all changes are saved - returns max version.
+    fn get_max_persisted_version(
```
generall, line 150: This function doesn't require flush to collect persisted versions anymore

lib/segment/src/segment/entry.rs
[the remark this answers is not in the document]
generall: nope. If versions are the same, we still want to proceed.

lib/shard/src/segment_holder/flush.rs
```diff
@@ -0,0 +1,216 @@
...
+            max_persisted_version = max(max_persisted_version, segment_persisted_version);
+
+            drop(read_segment);
+        }
+
+        if has_unsaved {
+            log::trace!(
+                "Some segments have unsaved changes, lowest unsaved version: {min_unsaved_version}"
+            );
+            Ok(min_unsaved_version)
```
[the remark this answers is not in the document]
generall: this logic wasn't changed, and it is sound. We want to assume that last version might have been not fully applied, so we always retry operation, if version matches.

timvisee:
In the PR comment:

> To compensate for this, we currectly perform flush in a very specific order: we flush immutable segments first.

No, we flush mutable first. It has the new CoW data, and so we always want to flush that first before any deletes in immutable.

I expect the code to be correct, but I'll pay extra attention there.

lib/segment/src/entry/entry_point.rs
```diff
@@ -262,11 +266,20 @@ pub trait SegmentEntry: SnapshotEntry {
...
-    /// Flushes current segment state into a persistent storage, if possible
-    /// if sync == true, block current thread while flushing
+    /// Returns a function, which when called, will flush all pending changes to disk.
+    /// If there are currently no changes to flush, returns None.
+    /// If `force` is true, will return a flusher even if there are no changes to flush.
     ///
-    /// Returns maximum version number which is guaranteed to be persisted.
-    fn flush(&self, sync: bool, force: bool) -> OperationResult<SeqNumberType>;
+    /// Returned function returns maximum version number which is guaranteed to be persisted.
+    fn flusher(&self, force: bool) -> Option<Flusher>;
```
timvisee, line 272: I really like that we have one boolean parameter now.

This currently does not explicitly block parallelism. I'm not sure how this behaves if we flush from two different places. If you haven't done so yet, this might need careful checking.

We do have <https://github.com/qdrant/qdrant/pull/7388#discussion_r2425676643>, but I'm not confident it is sound.

Note-to-self: check flush implementation in all storages

lib/segment/src/segment/entry.rs
```diff
@@ -39,6 +40,10 @@ impl SegmentEntry for Segment {
         self.version.unwrap_or(0)
     }
 
+    fn persistent_version(&self) -> SeqNumberType {
+        (*self.persisted_version.lock()).unwrap_or(0)
```
timvisee, line 44: How about we return `None` here? Might prevent silently promoting this artificial version into a real one.

lib/segment/src/segment/entry.rs
```diff
@@ -619,23 +624,19 @@ impl SegmentEntry for Segment {
...
 
-        let mut background_flush_lock = self.lock_flushing()?;
         match (self.version, current_persisted_version) {
             (None, _) => {
                 // Segment is empty, nothing to flush
-                return Ok(current_persisted_version.unwrap_or(0));
+                return None;
             }
             (Some(version), Some(persisted_version)) => {
                 if !force && version == persisted_version {
```
timvisee, line 619: Note-to-self: check if any of our internal batching implementations would allow to flush just half of the changes for a version. If it's possible to flush only the first half of a version, this shortcut may break recovery.

lib/segment/src/segment/entry.rs
```diff
@@ -734,26 +735,29 @@ impl SegmentEntry for Segment {
...
             })?;
+
+            let mut current_persisted_version_guard = persisted_version.lock();
+            let persisted_version_value_opt = *current_persisted_version_guard;
+
+            if persisted_version_value_opt > state.version {
+                debug_assert!(
+                    persisted_version_value_opt.is_some(),
+                    "Persisted version should never be None if it's greater than state.version"
+                );
```
timvisee, line 729:
```suggestion
```

This debug assertion is always true, so we can remove it.

The lowest possible values for `persisted_version_value_opt` and `state.version` are `Some(0) > None`, enforced by the if-statement above. And so `persisted_version_value_opt.is_some()` is always true.

lib/segment/src/segment/mod.rs
```diff
@@ -104,10 +102,6 @@ impl fmt::Debug for VectorData {
 
 impl Drop for Segment {
     fn drop(&mut self) {
-        if let Err(flushing_err) = self.lock_flushing() {
-            log::error!("Failed to flush segment during drop: {flushing_err}");
-        }
```
timvisee, line 113: Why not lock the persisted version here?

lib/segment/src/segment/entry.rs
timvisee: The implementation is not correct.

Currently it only prevents parallelism for saving the segment state file to disk. IIRC we don't even need a lock there because it is already using atomic save.

What I think is much more important is to disallow parallelism during the flush operations from all storages above. This lock currently doesn't fulfill that function.

I'm note sure whether it is fine to just lock the persisted version longer, as it potentially makes `persisted_version()` blocking for much longer too.

timvisee:
To fully fix segment ordering, even with proxies, I think we also need a change like [this](https://github.com/qdrant/qdrant/pull/7381). Though I suggest to keep the PRs separate.

lib/shard/src/segment_holder/flush.rs
```diff
@@ -0,0 +1,216 @@
...
+            segment_reads
+                .windows(2)
+                .all(|s| s[0].is_appendable() || !s[1].is_appendable()),
+            "Must flush appendable segments first",
+        );
+
+        if !sync && self.is_background_flushing() {
+            // There is already a background flush ongoing, return current max persisted version
+            return self.get_max_persisted_version(segment_reads, lock_order);
+        }
```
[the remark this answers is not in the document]
timvisee: This is fine.

We normally only start a flush if there are any changes. If any flush has changes, we only acknowledge the lowest version number in the WAL. It means that it'll replay the missed operations on restart.

An exception may be that we also start a flush if the force flag is provided. If the first forced flush has no changes, but the follow up non-force flush does (and we ignore it), we might acknowledge a version that is too recent. This needs further investigation.

agourlay:
FYI running Crasher on this branch locally shows no sign of missing points after 30 minutes of runtime :crossed_fingers:

timvisee:
<https://github.com/qdrant/qdrant/pull/7381> is now adjusted and rebased on top of this PR.

lib/segment/src/segment/entry.rs
generall: yeah, it is mostly for self-control

lib/segment/src/segment/entry.rs
generall: atomic save in this case doesn't guarantee ordering, as nothing prevents those atomic saves to be called in random order. 

Regarding safety of parallel flushes in the above components, I think the existing implementation in master doesn't prevent those as well - there are a few places where we do `sync` flush, and it ignores existance of another parallel backgroud flush.

generall:
depends on https://github.com/qdrant/qdrant/pull/7400 to fix tests

lib/segment/src/entry/entry_point.rs
timvisee: I manually checked all flushers. All are fine, with one exception and one note:

1. the mutable ID tracker doe snot support concurrent flushes, we cannot open the files on disk twice concurrently
2. the last persisted version in the segment holder is not sound when concurrent flushes are happening, we might acknowledge the wrong version

Since these two problems currently don't clearly show, I propose to handle them in a separate PR once all our current work is merged.

PR #7390, opened 2025-10-13
Introduce multi source transfer meta operation

KShivendu:
This PR introuces multi-source transfer interface/logic that only accepts single source (for now).

We decided not to extend the existing `TransferShard` with new enum variants, because that would require implementing those variants in unrelated transfer methods like snapshots, making the code harder to read and maintain.

This one basically stores `multi_source_transfers_state` in shard holder and refers to this state to trigger normal `TransferShard` operations

For now, I’m starting only implementing the start and finish stages to keep PRS small and reviewable.

It can be triggered via `replicate_points` action:
```http
POST /collections/demo/cluster
{
  "replicate_points": {
    "filter": {
      "must": { "key": "group_id", "match": { "value": "user_3" } }
    },
    "from_shard_key": "default",
    "to_shard_key": "user_3"
  }
}
```

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -550,27 +550,7 @@ impl Collection {
         on_finish: impl Future<Output = ()> + Send + 'static,
         on_failure: impl Future<Output = ()> + Send + 'static,
     ) -> CollectionResult<()> {
-        // Check if the transfer is already in progress
-        {
-            let shard_holder = self.shards_holder.read().await;
-            if shard_holder.multi_source_transfer_state.read().is_some() {
-                // todo: Allow more than one at a time?
-                return Err(CollectionError::bad_input(
-                    "Multi source transfer is already in progress",
-                ));
-            };
-        }
-
-        // Initialize the state in the shard holder
-        {
-            let shard_holder = self.shards_holder.write().await;
-            shard_holder.multi_source_transfer_state.write(|state| {
-                *state = Some(MultiSourceTransferState {
-                    transfer: transfer.clone(),
-                    completed_sources: vec![],
-                });
-            })?;
-        }
+        let transfer_clone = transfer.clone();
 
         let MultiSourceTransfer {
             from_peer_ids,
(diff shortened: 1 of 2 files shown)


timvisee:
> This one basically stores multi_source_transfers_state in shard holder and refers to this state to trigger normal TransferShard operations

Could you elaborate on how this works?

Does this indeed decouple multi transfers and regular transfers. Where multi transfers have separate state and orchestrate regular transfers? I assume we (can eventually) report the multi transfer state/metadata in collection info too.

I see that `POST /collections/demo/cluster { "replicate_points": {} }` abstracts these multi transfers away. I think that is fine. We could always expose some other interface to setup and control _custom multi transfers_.

If it works like the way I describe above, I wonder if we should call it 'multi transfers' (not multi source transfers). Maybe there will be use cases in the future where targeting multiple shards is reasonable. Don't quote me on it, thinking out loud here :smile:. Resharding down actually currently uses a mechanism like this, though it's controlled in a different way and managed by cluster manager.

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -53,18 +53,20 @@ use crate::shards::replica_set::ShardReplicaSet;
...
 const RESHARDING_STATE_FILE: &str = "resharding_state.json";
+const MULTI_SOURCE_TRANSFER_STATE_FILE: &str = "multi_source_transfer.json";
 pub const SHARD_KEY_MAPPING_FILE: &str = "shard_key_mapping.json";
 
 pub struct ShardHolder {
     shards: AHashMap<ShardId, ShardReplicaSet>,
     pub(crate) shard_transfers: SaveOnDisk<HashSet<ShardTransfer>>,
     pub(crate) shard_transfer_changes: broadcast::Sender<ShardTransferChange>,
     pub(crate) resharding_state: SaveOnDisk<Option<ReshardState>>,
+    pub(crate) multi_source_transfer_state: SaveOnDisk<Option<MultiSourceTransferState>>,
```
timvisee, line 69: Can we support at most one such multi transfer?

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -541,4 +542,91 @@ impl Collection {
...
+                shard_id: from_shard_id,
+                to_shard_id: Some(to_shard_id),
+                from: from_peer_id,
+                to: to_peer_id,
+                method: Some(ShardTransferMethod::StreamRecords),
+                sync: true,
+                filter,
+            },
+            consensus,
+            self.path.join("temp"),
```
timvisee, line 609: I don't think this is a valid temp directory.

lib/collection/src/shards/transfer/mod.rs
```diff
@@ -102,6 +102,22 @@ impl ShardTransfer {
     }
 }
 
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize)]
+pub struct MultiSourceTransferState {
+    pub transfer: MultiSourceTransfer,
+    pub completed_sources: Vec<(PeerId, ShardId)>,
```
timvisee, line 108: A potential issue may be that moving/replicating replicas may mess with this list of sources.

Maybe it makes sense to add some sort of UUID to replicas that is kept when a replica is moved.

It may be implicit because cluster manager does some rebalancing.

Resharding suffers from the same problem. Though active resharding currently prevents rebalancing.

PR #7400, opened 2025-10-13
fix segment repair on load

generall:
Consider following situation:

- We remove some point from the segment
- Flush operation begins, it flushes point ids (deletes them)
- Process gets killed

In this situation we end up in a configuration, where WAL contains delete operation by point-id, but no such point-id is registered anywhere. But the payload and payload indexes are still there, displaying wrong statistics.

We actually had a mechanism, which was supposed to clean-up leftovers of the deleted points.

Unfortunatelly, the existing mechanism was broken. 
This PR fixes the segment repair process. In order to do so effectively, it intoduces two hacks:

- Points versions are set to 0 on deletions. If recovery process sees the non-zero version with no point-is attaches, it assumes this point needs cleaning.
- LocalShards are initialized with empty WAL operation to bump first meaningful operration version to be >0. This is our long-living problem, which often causes troubles. In this case I didn't find a suitable workaround. (I did consider u64:MAX, but it will compromise the efficiency of compressed versions representation)

### DIFF
--- a/lib/segment/src/id_tracker/id_tracker_base.rs
+++ b/lib/segment/src/id_tracker/id_tracker_base.rs
@@ -176,7 +176,7 @@ pub trait IdTracker: fmt::Debug {
     ///
     /// Returns a list of internal ids, that have non-zero versions, but are missing in the id mapping.
     /// Those points should be removed from all other parts of the segment.
-    fn cleanup_versions(&mut self) -> OperationResult<Vec<PointOffsetType>> {
+    fn fix_inconsistencies(&mut self) -> OperationResult<Vec<PointOffsetType>> {
         // Points with mapping, but no version.
         // Can happen if insertion didn't complete.
         // We need to remove mapping and assume the point is going to be re-inserted by WAL.
@@ -472,13 +472,13 @@ impl IdTracker for IdTrackerEnum {
         }
     }
 
-    fn cleanup_versions(&mut self) -> OperationResult<Vec<PointOffsetType>> {
+    fn fix_inconsistencies(&mut self) -> OperationResult<Vec<PointOffsetType>> {
         match self {
-            IdTrackerEnum::MutableIdTracker(id_tracker) => id_tracker.cleanup_versions(),
-            IdTrackerEnum::ImmutableIdTracker(id_tracker) => id_tracker.cleanup_versions(),
-            IdTrackerEnum::InMemoryIdTracker(id_tracker) => id_tracker.cleanup_versions(),
+            IdTrackerEnum::MutableIdTracker(id_tracker) => id_tracker.fix_inconsistencies(),
+            IdTrackerEnum::ImmutableIdTracker(id_tracker) => id_tracker.fix_inconsistencies(),
+            IdTrackerEnum::InMemoryIdTracker(id_tracker) => id_tracker.fix_inconsistencies(),
             #[cfg(feature = "rocksdb")]
-            IdTrackerEnum::RocksDbIdTracker(id_tracker) => id_tracker.cleanup_versions(),
+            IdTrackerEnum::RocksDbIdTracker(id_tracker) => id_tracker.fix_inconsistencies(),
         }
     }
 
--- a/lib/segment/src/segment/segment_ops.rs
+++ b/lib/segment/src/segment/segment_ops.rs
@@ -496,7 +496,7 @@ impl Segment {
     /// Removes partially persisted points.
     pub fn check_consistency_and_repair(&mut self) -> OperationResult<()> {
         // Get rid of versionless points.
-        let ids_to_clean = self.cleanup_versions()?;
+        let ids_to_clean = self.fix_id_tracker_inconsistencies()?;
 
         // There are some leftovers to clean from segment.
         // After that we need to set internal version to 0, so that
@@ -653,9 +653,10 @@ impl Segment {
         self.id_tracker.borrow().total_point_count()
     }
 
+    /// Fixes inconsistencies in the ID tracker, if any.
     /// Returns list of IDs, which should be removed from segment
-    pub fn cleanup_versions(&mut self) -> OperationResult<Vec<PointOffsetType>> {
-        self.id_tracker.borrow_mut().cleanup_versions()
+    pub fn fix_id_tracker_inconsistencies(&mut self) -> OperationResult<Vec<PointOffsetType>> {
+        self.id_tracker.borrow_mut().fix_inconsistencies()
     }
 }
 
--- a/lib/segment/src/fixtures/payload_context_fixture.rs
+++ b/lib/segment/src/fixtures/payload_context_fixture.rs
@@ -184,7 +184,7 @@ impl IdTracker for FixtureIdTracker {
         Box::new(iter::empty())
     }
 
-    fn cleanup_versions(&mut self) -> OperationResult<Vec<PointOffsetType>> {
+    fn fix_inconsistencies(&mut self) -> OperationResult<Vec<PointOffsetType>> {
         // This structure does not support cleaning up orphan versions
         Ok(vec![])
     }


lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -603,7 +605,23 @@ impl LocalShard {
...
+    /// We need it mostly to force WAL to start with something besides zero as a first operation number.
+    pub async fn insert_fake_operation(&self) -> CollectionResult<()> {
+        let mut operation = OperationWithClockTag {
+            operation: CollectionUpdateOperations::PointOperation(PointOperations::UpsertPoints(
+                PointInsertOperationsInternal::from(vec![]),
+            )),
+            clock_tag: None,
+        };
+        self.wal.lock_and_write(&mut operation).await?;
+        Ok(())
```
generall, line 624: This is a hack I couldn't avoid.

lib/segment/src/segment/mod.rs
```diff
@@ -104,9 +104,11 @@ impl fmt::Debug for VectorData {
...
 impl Drop for Segment {
     fn drop(&mut self) {
-        if let Err(flushing_err) = self.lock_flushing() {
-            log::error!("Failed to flush segment during drop: {flushing_err}");
-        }
+        // This should trigger test failure, as instead of nice shutdown
+        // We discard flushing process
+        // if let Err(flushing_err) = self.lock_flushing() {
+        //     log::error!("Failed to flush segment during drop: {flushing_err}");
+        // }
```
generall: This is the cause of the test failure. It is completely removed in https://github.com/qdrant/qdrant/pull/7388 and only serves to hide problem

lib/segment/src/segment/segment_ops.rs
```diff
@@ -299,6 +299,33 @@ impl Segment {
         Ok(applied)
     }
 
+    pub fn delete_point_internal(
```
generall, line 302: Moved this into separate fucntion to have a single source of truth for clearing segment data on delete

lib/segment/src/segment/segment_ops.rs
```diff
@@ -469,49 +496,24 @@ impl Segment {
     /// Removes partially persisted points.
     pub fn check_consistency_and_repair(&mut self) -> OperationResult<()> {
         // Get rid of versionless points.
-        self.cleanup_versions()?;
+        let ids_to_clean = self.cleanup_versions()?;
 
-        // Get rid of unmapped points.
-        let mut internal_ids_to_delete = HashSet::new();
-        let id_tracker = self.id_tracker.borrow();
-        for internal_id in id_tracker.iter_ids() {
```
generall, line 477: this was broken, cause `iter_ids` skips points which are already partially deleted.

lib/segment/src/id_tracker/simple_id_tracker.rs
```diff
@@ -231,6 +231,15 @@ impl IdTracker for SimpleIdTracker {
...
     }
 
+    fn drop_internal(&mut self, internal_id: PointOffsetType) -> OperationResult<()> {
+        if let Some(external_id) = self.mappings.external_id(internal_id) {
+            self.mappings.drop(external_id);
+            self.delete_key(&external_id)?;
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
generall: Yes, simple ID tracker doesn't have the fix, cause it is going to be deprecated and removed anyway. It is also have differnt version storage structure, which makes it incompatible.

lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -22,6 +22,9 @@ use crate::types::{PointIdType, SeqNumberType};
 /// would otherwise be introduced by HNSW/ID tracker point sampling.
 const SEED: u64 = 0b1011000011011110001110010101001010001011001101001010010001111010;
 
+/// This version should be assigned to a point when it is deleted.
```
timvisee, line 25:
```suggestion
/// This version should be assigned to a point when it is deleted.
/// It does not mean a point with this version is always deleted.
```

lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -156,10 +163,54 @@ pub trait IdTracker: fmt::Debug {
...
+    ) -> Box<dyn Iterator<Item = (PointOffsetType, SeqNumberType)> + '_>;
+
     /// Finds inconsistencies between id mapping and versions storage.
     /// It might happen that point doesn't have version due to un-flushed WAL.
     /// This method makes those points usable again.
-    fn cleanup_versions(&mut self) -> OperationResult<()>;
+    ///
+    /// Returns a list of internal ids, that have non-zero versions, but are missing in the id mapping.
+    /// Those points should be removed from all other parts of the segment.
+    fn cleanup_versions(&mut self) -> OperationResult<Vec<PointOffsetType>> {
```
timvisee: Bit of a nitpick, but this function name suggests me it cleans stuff up for us - it doesn't.

Maybe it's reasonable to rename it to something like `orphan_versions()`

lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -156,10 +163,54 @@ pub trait IdTracker: fmt::Debug {
...
+        let mut to_return = Vec::new();
+
+        for (internal_id, version) in self.iter_internal_versions() {
+            if version != DELETED_POINT_VERSION && self.external_id(internal_id).is_none() {
+                to_return.push(internal_id);
+            }
+        }
+
+        for internal_id in self.iter_internal() {
+            if self.internal_version(internal_id).is_none() {
```
timvisee, line 197: We can't use `iter_internal_versions()` here because it gives us the inverse, only iterating points having a version, not points without a version. Right?

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -603,7 +605,23 @@ impl LocalShard {
...
+
+    /// This operation inserts an empty operation into WAL.
+    /// Operation does nothing, but takes a spot in WAL.
+    /// We need it mostly to force WAL to start with something besides zero as a first operation number.
+    pub async fn insert_fake_operation(&self) -> CollectionResult<()> {
+        let mut operation = OperationWithClockTag {
+            operation: CollectionUpdateOperations::PointOperation(PointOperations::UpsertPoints(
+                PointInsertOperationsInternal::from(vec![]),
+            )),
+            clock_tag: None,
```
timvisee, line 621: I think it's best to set this clock tag, to aid with WAL delta. Let me validate that on the side.

In WAL delta resolution, we use [version 0](https://github.com/qdrant/qdrant/blob/895318913afbad4a3a3d89d5a667369a66964469/lib/collection/src/wal_delta.rs#L185-L189) as a 'receive everything' trick. We rely on an operation with clock tick 0 to exist, or at least we hope it exists.

If an operation with tick 0 cannot be found, it will [fall back](https://github.com/qdrant/qdrant/blob/895318913afbad4a3a3d89d5a667369a66964469/lib/collection/src/wal_delta.rs#L239) to other transfer mechanisms such as stream records.

Having an operation with clock tick 0 is semi-critical. So I vote to set the clock tag here. Sending this operation over to other peers is not a problem.

Even when clock tick 0 is requested, the WAL delta transfer may still be much faster than stream records. This may be the case when a new peer recently joined, and we only request clock tick 0 for that peer.

timvisee: Let me rectify that. After closer inspection, I believe we're good with not setting the clock.

Clocks track their own versions, and it is not related to WAL operation versions. That means that the next operation will get clock tick 0 which we desire to have.

Having this no-op operation here acts exactly the same as if the WAL was still completely empty.

Conclusion: no change is necessary! :raised_hands:

lib/segment/src/id_tracker/simple_id_tracker.rs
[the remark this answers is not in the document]
timvisee: I agree this is fine.

We actively migrate away from this simple ID tracker since Qdrant 1.15.0. It'll therefore not be used in practice anymore.

If some old storage or snapshot versions still uses it, we use the structure to load it. But then it is immediately migrated to something else.

lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -156,10 +163,54 @@ pub trait IdTracker: fmt::Debug {
...
+        }
+        for external_id in to_remove {
+            self.drop(external_id)?;
+            #[cfg(debug_assertions)]
+            {
+                log::debug!("dropped version for point {external_id} without version");
+            }
+        }
+        Ok(to_return)
+    }
```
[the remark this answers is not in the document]
timvisee: Good call if the simple ID tracker was still actively used. But it is not, and we actively migrate away from it.

We can keep it as is for the same reasoning as elaborated upon here: https://github.com/qdrant/qdrant/pull/7400#discussion_r2428502848

lib/segment/src/segment/mod.rs
timvisee: Kept for RocksDB: https://github.com/qdrant/qdrant/pull/7400/commits/04599fd5adb93c6bb3f514600efdd2899c92ecc1

lib/segment/src/id_tracker/in_memory_id_tracker.rs
```diff
@@ -79,7 +79,20 @@ impl IdTracker for InMemoryIdTracker {
     }
 
     fn drop(&mut self, external_id: PointIdType) -> OperationResult<()> {
-        self.mappings.drop(external_id);
+        let internal_id = self.mappings.drop(external_id);
+        if let Some(internal_id) = internal_id {
+            self.set_internal_version(internal_id, DELETED_POINT_VERSION)?;
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
timvisee: Good call.

I flipped the ordering for this one in <https://github.com/qdrant/qdrant/pull/7400/commits/1451dcddc399f3d55b57be0790d3df3dda16299b>

cc @generall

PR #7402, opened 2025-10-14
add test for shard recovery after partial flush

generall:
Followup test for https://github.com/qdrant/qdrant/pull/7400

### DIFF
--- a/lib/collection/src/tests/wal_recovery_test.rs
+++ b/lib/collection/src/tests/wal_recovery_test.rs
@@ -197,9 +197,9 @@ async fn test_partial_flush_recovery() {
         .await
         .unwrap();
 
-    shard.full_flush();
-
     shard.stop_flush_worker().await;
+
+    shard.full_flush();
 
     let delete_point_op = delete_point_operation(4);
     shard


lib/collection/src/tests/wal_recovery_test.rs
```diff
@@ -139,3 +139,103 @@ async fn test_delete_from_indexed_payload() {
...
+            );
+        })
+        .unwrap();
+
+    shard
+        .update(index_op.into(), true, hw_acc.clone())
+        .await
+        .unwrap();
+
+    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
timvisee: Is this sleep necessary, we have wait=true above?

If it is because we want to flush, wouldn't it be better to:
- stop the flush worker before
- explicitly call flush now

generall: there are no explicit flush exposed in LocalShard, but I can try to make it as a helper function

timvisee: I think it's worth it, next to the partial flush. Sleeps tend to be flaky.

PR #7403, opened 2025-10-14
Refactor: merge `_search_on_level` into `search_on_level`

xzfc:
This PR merges `_search_on_level` into `search_on_level` and removes the dead code that used it.

---

The current `dev` has two related functions: `GraphLayersBase::search_on_level` and `GraphLayersBase::_search_on_level`.

The former is high-level, and the latter is a bit more flexible as it allows reusing the `SearchContext` and `VisitedListHandle` after calling it.

But the only place where these are reused is inside the following dead code:
https://github.com/qdrant/qdrant/blob/02de82674839178e0e058ab48581b60c0b3e9e92/lib/segment/src/index/hnsw_index/graph_layers_builder.rs#L523-L530
This code is dead because `existing_links` is always empty:
- `link_new_point`/`link_new_point_on_level` or `add_new_point`[^1] is called only once per point, so it shouldn't be filled by them before.
- It can't be filled by the backlinks mechanism because this point can't appear in search until marked as ready.
- Perhaps it was in use early on for the additional links feature, but as of now, the additional links are built in a separate scratch graph.

This is confirmed by codecov:
https://app.codecov.io/gh/qdrant/qdrant/blob/dev/lib%2Fsegment%2Fsrc%2Findex%2Fhnsw_index%2Fgraph_layers_builder.rs#L524

[^1]: `add_new_point` is from incremental HNSW

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_layers.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers.rs
@@ -188,27 +188,32 @@ pub trait GraphLayersWithVectors: GraphLayersBase {
         level: usize,
     ) -> (&[u8], impl Iterator<Item = (PointOffsetType, &[u8])> + '_);
 
-    /// Similar to [`GraphLayersBase::_search_on_level`].
-    #[allow(clippy::too_many_arguments)]
-    fn _search_on_level_with_vectors(
+    /// Similar to [`GraphLayersBase::search_on_level`].
+    fn search_on_level_with_vectors(
         &self,
-        links_searcher: &mut SearchContext,
-        base_searcher: &mut SearchContext,
+        level_entry: ScoredPointOffset,
         level: usize,
-        visited_list: &mut VisitedListHandle,
+        ef: usize,
         links_scorer: &FilteredBytesScorer,
         base_scorer: &dyn QueryScorerBytes,
         is_stopped: &AtomicBool,
-    ) -> CancellableResult<()> {
+    ) -> CancellableResult<FixedLengthPriorityQueue<ScoredPointOffset>> {
+        let mut visited_list = self.get_visited_list_from_pool();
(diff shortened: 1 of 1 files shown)


lib/segment/src/index/hnsw_index/graph_layers_builder.rs
```diff
@@ -509,28 +497,15 @@ impl GraphLayersBuilder {
...
         let level_m = self.hnsw_m.level_m(curr_level);
         let scorer = |a, b| points_scorer.score_internal(a, b);
 
         let selected_nearest = {
+            let iter = nearest.into_iter_sorted();
             let mut existing_links = self.links_layers[point_id as usize][curr_level].write();
-            {
-                let ready_list = self.ready_list.read();
-                for existing_link in existing_links.iter() {
-                    if !visited_list.check(existing_link) && ready_list[existing_link as usize] {
```
IvanPleshkov, line 524: Great that you proved that this code is unreachable. We also need to remove this logic from gpu implementation in the future

IvanPleshkov (approved):

lib/segment/src/index/hnsw_index/graph_layers_builder.rs
xzfc: I've tried to find where this logic is implemented for GPU, but can't find it. I don't see it there:

- https://github.com/qdrant/qdrant/blob/969cc82018dd722262b93d003175b3d69961a135/lib/segment/src/index/hnsw_index/gpu/shaders/run_insert_vector.comp
- https://github.com/qdrant/qdrant/blob/969cc82018dd722262b93d003175b3d69961a135/lib/segment/src/index/hnsw_index/gpu/shaders/search_context.comp

(not that I want to remove it right now, just curious)

PR #7405, opened 2025-10-14
make flush worker loop async

generall:
One less thread per shard.

### DIFF
--- a/lib/collection/src/update_handler.rs
+++ b/lib/collection/src/update_handler.rs
@@ -928,17 +928,16 @@ impl UpdateHandler {
         shard_path: PathBuf,
     ) {
         loop {
-            // Stop flush worker on signal or if sender was dropped
-            // Even if timer did not finish
-            let stop = tokio::select! {
-                _ = tokio::time::sleep(Duration::from_secs(flush_interval_sec)) => false,
-                _ = &mut stop_receiver => true,
+            tokio::select! {
+                biased;
+                // Stop flush worker on signal or if sender was dropped
+                _ = &mut stop_receiver => {
+                    log::debug!("Stopping flush worker for shard {}", shard_path.display());
+                    return;
+                },
+                // Flush at the configured flush interval
+                _ = tokio::time::sleep(Duration::from_secs(flush_interval_sec)) => {},
             };
-
-            if stop {
-                log::debug!("Stopping flush worker for shard {}", shard_path.display());
-                return;
-            }
 
             let segments_clone = segments.clone();
             let wal_clone = wal.clone();


lib/collection/src/update_handler.rs
```diff
@@ -854,89 +850,111 @@ impl UpdateHandler {
...
+
+            tokio::task::spawn_blocking(move || {
+                Self::flush_worker_internal(
+                    segments_clone,
+                    wal_clone,
+                    wal_keep_from_clone,
+                    clocks_clone,
+                    shard_path_clone,
+                )
+            });
```
generall, line 960: this is the main change, not time is async, but we spawn thread on wach timeout. (actually, thread keep-alive is higher than that, only one thread is spawned per multiple collections)

src/common/helpers.rs
```diff
@@ -35,8 +35,11 @@ pub fn create_search_runtime(max_search_threads: usize) -> io::Result<Runtime> {
 pub fn create_update_runtime(max_optimization_threads: usize) -> io::Result<Runtime> {
     let mut update_runtime_builder = runtime::Builder::new_multi_thread();
 
+    let num_cpus = common::cpu::get_num_cpus();
+
     update_runtime_builder
         .enable_time()
+        .worker_threads(num_cpus)
```
ffuugoo, line 42: [We don't really need to set this explicitly, because `tokio` creates thread-per-core by default](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.worker_threads)

generall: yes, but we want tokio to respect our `QDRANT_NUM_CPUS`, we need this.

ffuugoo: ah, didn't notice it's `common`. makes sense then

lib/collection/src/update_handler.rs
```diff
@@ -854,89 +850,115 @@ impl UpdateHandler {
...
+        flush_interval_sec: u64,
+        mut stop_receiver: oneshot::Receiver<()>,
+        shard_path: PathBuf,
+    ) {
+        loop {
+            // Stop flush worker on signal or if sender was dropped
+            // Even if timer did not finish
+            let stop = tokio::select! {
+                _ = tokio::time::sleep(Duration::from_secs(flush_interval_sec)) => false,
+                _ = &mut stop_receiver => true,
```
timvisee: Maybe we can bias this to the stop receiver? That way we will neve3r do a full flush iteration if we already stopped.

https://docs.rs/tokio/latest/tokio/macro.select.html#fairness

```suggestion
            let stop = tokio::select! {
                biased;
                _ = &mut stop_receiver => true,
                _ = tokio::time::sleep(Duration::from_secs(flush_interval_sec)) => false,
```

We can also just move the stop handling into the `select!` itself, similar to what we do in some `match`es:

```rust
// Stop flush worker on signal or if sender was dropped
// Even if timer did not finish
tokio::select! {
    biased;
    _ = &mut stop_receiver => {
        log::debug!("Stopping flush worker for shard {}", shard_path.display());
        return;
    },
    _ = tokio::time::sleep(Duration::from_secs(flush_interval_sec)) => {},
};
```

lib/collection/src/update_handler.rs
```diff
@@ -854,89 +850,111 @@ impl UpdateHandler {
...
+                Self::flush_worker_internal(
+                    segments_clone,
+                    wal_clone,
+                    wal_keep_from_clone,
+                    clocks_clone,
+                    shard_path_clone,
+                )
+            });
         }
     }
```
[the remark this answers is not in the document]
timvisee: No. We await the spawned future and so we wait for it to terminate. It can therefore not overlap.

timvisee (approved):

lib/collection/src/update_handler.rs
```diff
@@ -854,89 +850,111 @@ impl UpdateHandler {
...
-            if let Err(err) = wal_flush_res {
-                log::error!("{err}");
-                segments.write().report_optimizer_error(err);
-                continue;
-            }
+        if let Err(err) = wal.blocking_lock().ack(ack) {
+            log::warn!("Failed to acknowledge WAL version: {err}");
+            segments.write().report_optimizer_error(err);
+        }
+    }
```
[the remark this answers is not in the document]
timvisee: Nope

PR #7414, opened 2025-10-16
Add ACORN-1 search

xzfc:
This PR adds support for ACORN-1 as described in the [ACORN](https://arxiv.org/abs/2403.04871v1) paper.
This is a variation of HNSW search that "hops" over neighbors that do not pass the filter.
The index structure remains the same.

The new search is controlled/enabled via a new search parameter `acorn_limit_factor`, e.g.:
```bash
curl -X POST 'http://localhost:6333/collections/benchmark/points/search' -d '
    {
        "vector": […],
        "filter": {…},
        "top": 10,
        "params": {
            "acorn_limit_factor": 10.0
        }
    }'
```

When the new parameter is `null` or not present, the search behaves the same as before (without ACORN-1).

The design/name of a new parameter is subject to change later as we perform more experiments (we might add more knobs or change it to a boolean flag).

---

Also, since the number of "search on level" functions reached five, this PR adds a module doc comment describing their differences.

# Benchmarks

Benchmark results on random vectors:
- `dim=128`, `m=16`, `payload_m=0`, `ef_search=256`, `num_points=500_000`
- randomly generated vectors (from -1.0 to 1.0)
- randomly generated payloads (two fields, each has 5 possible values)
- search filters match both of the fields (i.e. cardinality is 1/25)

Without ACORN:
- Mean accuracy 0.006
- Mean time 7.5 ms (132.8 RPS)

With ACORN:
- Mean accuracy 0.631
- Mean time 30.2 ms (33.2 RPS)

### DIFF
--- a/docs/grpc/docs.md
+++ b/docs/grpc/docs.md
@@ -2446,10 +2446,12 @@ The JSON representation for `Value` is a JSON value.
 
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
-| enable | [bool](#bool) | optional | When enabled, the ACORN might be enabled for the HNSW search.
+| enable | [bool](#bool) | optional | If true, then ACORN may be used for the HNSW search based on filters selectivity.
 
 Improves search recall for searches with multiple low-selectivity payload filters, at cost of performance. |
-| max_selectivity | [double](#double) | optional | Maximum selectivity of the filter to enable ACORN. If filter selectivity estimated as higher than this value across all points, ACORN will not be used. Selectivity is estimated as: `estimated number of points satisfying the filter / total number of points`.
+| max_selectivity | [double](#double) | optional | Maximum selectivity of filters to enable ACORN.
+
+If estimated filters selectivity is higher than this value, ACORN will not be used. Selectivity is estimated as: `estimated number of points satisfying the filters / total number of points`.
 
 0.0 for never, 1.0 for always. Default is 0.4. |
 
--- a/lib/segment/src/index/hnsw_index/graph_layers.rs
+++ b/lib/segment/src/index/hnsw_index/graph_layers.rs
@@ -174,15 +174,15 @@ pub trait GraphLayersBase {
         let mut search_context = SearchContext::new(ef);
         search_context.process_candidate(level_entry);
 
-        // Limit is per neighbor.
-        let limit_hop1 = self.get_m(level);
+        // Limits are per every explored 1-hop or 2-hop neighbors, not total.
+        // This is necessary to avoid over-scoring when there are many
+        // additional graph links.
+        let hop1_limit = self.get_m(level);
+        let hop2_limit = self.get_m(level);
+        debug_assert_ne!(self.get_m(level), 0); // See `FilteredBytesScorer::score_points`
 
-        // How many 2-hop neighbors to collect per neighbor.
-        // This is necessary to avoid over-scoring in case of extra links per neighbor.
-        let limit_hop2 = self.get_m(level);
-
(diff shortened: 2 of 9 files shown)


docs/grpc/docs.md
```diff
@@ -4738,6 +4738,7 @@ Parameterized reciprocal rank fusion
 | exact | [bool](#bool) | optional | Search without approximation. If set to true, search may run long but with exact results. |
 | quantization | [QuantizationSearchParams](#qdrant-QuantizationSearchParams) | optional | If set to true, search will ignore quantized vector data |
 | indexed_only | [bool](#bool) | optional | If enabled, the engine will only perform search among indexed or small segments. Using this option prevents slow searches in case of delayed index, but does not guarantee that all uploaded vectors will be included in search results |
+| acorn_limit_factor | [double](#double) | optional |  |
```
timvisee: Even though it's subject to change, lets consistent naming for now

`acorn_limit_factor`

lib/segment/src/types.rs
```diff
@@ -534,6 +534,10 @@ pub struct SearchParams {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub hnsw_ef: Option<usize>,
 
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[validate(range(min = 1.0))]
+    pub acorn_limit_factor: Option<OrderedFloat<f64>>,
```
timvisee: Why not an integer? Or is it a factor over `m`

lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -103,6 +141,89 @@ pub trait GraphLayersBase {
...
+            // Collect 1-hop neighbors (direct neighbors)
+            self.for_each_link(candidate.idx, level, |link| {
+                if visited_list.check(link) || expanded_list.check_and_update_visited(link) {
+                    return;
+                }
+
+                if points_scorer.filters().check_vector(link) {
+                    if points_ids.len() < limit {
+                        points_ids.push(link);
+                    }
```
timvisee: Does a `try_for_each_link` make sense here as well? To shortcut when we've filled the points ID list?

lib/segment/src/index/hnsw_index/graph_layers.rs
```diff
@@ -103,6 +141,89 @@ pub trait GraphLayersBase {
...
+    ) -> CancellableResult<FixedLengthPriorityQueue<ScoredPointOffset>> {
+        let mut visited_list = self.get_visited_list_from_pool();
+        visited_list.check_and_update_visited(level_entry.idx);
+
+        let mut expanded_list = self.get_visited_list_from_pool();
+
+        let mut search_context = SearchContext::new(ef);
+        search_context.process_candidate(level_entry);
+
+        let limit = (self.get_m(level) as f64 * acorn_limit_factor).ceil() as usize;
```
timvisee: Shouldn't we only use this limit for the 2nd hop? Not for the initial search on each node?

lib/segment/src/types.rs
xzfc: It's a factor over `m`, just like `oversampling: Option<f64>`.

lib/segment/src/index/hnsw_index/graph_layers.rs
xzfc: Yes, your suggestion makes sense for consistency.
(either both `limit` and `try_for_each_link`; or neither)

- [ ] note to self

lib/segment/src/index/hnsw_index/graph_layers.rs
generall: +1

generall (approved):

PR #7415, opened 2025-10-17
Minor Gridstore improvements

timvisee:
Minor Gridstore improvements I found along the way, that don't solve any immediate problems.

Most notable changes:
- use `repr(C)` attributes for types we map to disk for a stable layout
- grow tracker file size by a power of two, instead of by an arbitrary number
- debug panic for getting out of bound point from Gridstore

### All Submissions:

### DIFF
--- a/lib/gridstore/src/tracker.rs
+++ b/lib/gridstore/src/tracker.rs
@@ -468,16 +468,20 @@ mod tests {
     }
 
     #[rstest]
-    #[case(10)]
-    #[case(100)]
-    #[case(1000)]
-    fn test_page_tracker_resize(#[case] initial_tracker_size: usize) {
+    #[case(10, 16)]
+    #[case(100, 128)]
+    #[case(1000, 1024)]
+    #[case(1024, 1024)]
+    fn test_page_tracker_resize(
+        #[case] desired_tracker_size: usize,
+        #[case] actual_tracker_size: usize,
+    ) {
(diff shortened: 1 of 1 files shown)


lib/gridstore/src/tracker.rs
```diff
@@ -222,18 +223,16 @@ impl Tracker {
...
+
+        // Grow tracker file if it isn't big enough
         if self.mmap.len() < end_offset {
-            // flush the current mmap
             self.mmap.flush().unwrap();
-            let missing_space = end_offset - self.mmap.len();
-            // reopen the file with a larger size
-            // account for missing size + extra to avoid resizing too often
-            let new_size = self.mmap.len() + missing_space + Self::DEFAULT_SIZE;
+            let new_size = end_offset.next_power_of_two();
```
timvisee, line 230: This now grows the tracker file by a power of two. It means the tracker file will have a size of: 1MB, 2MB, 4MB, 8MB...

Before the number was quite arbitrary, not being round numbers.

While it is not strictly required, I like the round numbers better.

ffuugoo (approved):

lib/gridstore/src/tracker.rs
ffuugoo: I like powers of two

timvisee: :zap:.:zap:

PR #7416, opened 2025-10-17
use mutex to ensure sequential segment flushed + await on drop

generall:
Fixes

```
2025-10-17T11:12:20.503980Z ERROR collection::update_handler: Failed to flush: Service runtime error: Failed to flush id_tracker mapping: Service runtime error: IO Error: failed to open file `./storage/collections/benchmark/0/segments/1a570d1b-07c5-4c40-8adf-d7a8ada7030d/mutable_id_tracker.mappings`: No such file or directory (os error 2)
```

introduced in https://github.com/qdrant/qdrant/pull/7388

### DIFF
--- a/lib/segment/src/segment/entry.rs
+++ b/lib/segment/src/segment/entry.rs
@@ -689,7 +689,17 @@ impl SegmentEntry for Segment {
         //
         //  400
 
+        let is_alive_flush_lock = self.is_alive_flush_lock.clone();
+
         let flush_op = move || {
+            // Keep the guard till the end of the flush to prevent concurrent flushes
+            let is_alive_flush_guard = is_alive_flush_lock.lock();
+
+            if !*is_alive_flush_guard {
+                // Segment is removed, skip flush
+                return Ok(());
+            }
+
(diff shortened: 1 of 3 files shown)


lib/segment/src/segment/entry.rs
```diff
@@ -689,7 +689,17 @@ impl SegmentEntry for Segment {
...
+        let is_alive_flush_lock = self.is_alive_flush_lock.clone();
+
         let flush_op = move || {
+            // Keep the guard till the end of the flush to prevent concurrent flushes
+            let is_alive_flush_guard = is_alive_flush_lock.lock();
+
+            if !*is_alive_flush_guard {
+                // Segment is removed, skip flush
+                return Ok(());
+            }
```
timvisee, line 701: I'd prefer a separate `Mutex<()>` and stopped flag, I think. It's a mechanism we use already, and then we can more easily propagate the flag throughout the rest of the flusher operation.

But it's not critical. Let's merge this and revisit it if it ever becomes important.

timvisee (approved):

PR #7420, opened 2025-10-18
feat(groups): add offset pagination to grouped queries (REST/gRPC/internal)

### LINKED ISSUE #7117: [Feature request] Pagination on group queries
Seems like [query-points-groups](https://api.qdrant.tech/api-reference/search/query-points-groups) doesn't support pagination with `offset` or something like `searchAfter`. This would be a nice feature to have!

khanSoliheen:
Introduce offset for grouped queries to enable forward pagination over groups.

Default offset to 0. Propagate through REST, gRPC, and internal execution.

Apply offset after establishing a deterministic global order of groups to prevent overlap/flakiness.
Ordering tuple: (score DESC, group_id ASC, best_point_id ASC[, shard_id ASC]).

Compute enough candidates to cover the requested window (offset + limit groups).
If the implementation gathers point candidates as an intermediate step, clarify: “internally we fetch up to (offset + limit) * group_size points before grouping.”

Extend tests and conversions to carry offset through REST/gRPC paths.

Fixes #7117.

API / Schema:

REST: add optional offset to QueryBaseGroupRequest / BaseGroupRequest (default 0).

gRPC: add offset to corresponding messages.

Backward-compatible: omitting offset keeps prior behavior.

### DIFF
--- a/lib/api/src/rest/schema.rs
+++ b/lib/api/src/rest/schema.rs
@@ -1062,7 +1062,7 @@ pub struct LookupLocation {
 }
 
 #[allow(clippy::unnecessary_wraps)]
-fn default_group_offset_u32() -> Option<u32> {
+fn default_group_offset_u64() -> Option<u64> {
     Some(0)
 }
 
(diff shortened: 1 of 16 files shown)


lib/api/src/rest/schema.rs
```diff
@@ -1088,9 +1088,9 @@ pub struct BaseGroupRequest {
     pub limit: u32,
 
     /// Offset of the group to start returning results from.
-    #[serde(default = "default_group_offset_u32")]
-    #[schemars(default = "default_group_offset_u32")]
-    pub offset: Option<u32>,
+    #[serde(default = "default_group_offset_u64")]
+    #[schemars(default = "default_group_offset_u64")]
+    pub offset: Option<u64>,
```
generall: AI doesn't always give a good advice. In this case `u32` is consistent with the type of `group_size` and didn't need to be changed into `u64`

oelhammouchi:
Any news on this?

PR #7427, opened 2025-10-20
Fix shard snapshots IO error, proxy first then flush

timvisee:
Fixes <https://github.com/qdrant/qdrant/pull/7404>

When taking a shard snapshot, we proxy all segments and flush them.

It is important that we proxy first, and then flush. Proxying ensures that the segments won't change anymore. After that we can flush the last changes that were still pending.

If we do it the other way around, new changes might land in the segments after flush but before we proxy them. The background flush task might trigger another flush during the snapshot process, which conflicts on file IO.

### All Submissions:


timvisee:
Status: @agourlay is running some test on this locally

PR #7434, opened 2025-10-21
rate limit slow request warning queue

generall:
I wasn't able to reproduce it even under 512 parallel search threads, but some users reported this message being spammy

### DIFF
--- a/lib/collection/src/profiling/slow_requests_collector.rs
+++ b/lib/collection/src/profiling/slow_requests_collector.rs
@@ -79,17 +79,14 @@ impl RequestsCollector {
                 .unwrap_or_else(|_| 0);
 
             // Atomically update if enough time has passed
-            let updated = LAST_SEND_WARN.fetch_update(
-                Ordering::Relaxed,
-                Ordering::Relaxed,
-                |prev| {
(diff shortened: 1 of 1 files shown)


lib/collection/src/profiling/slow_requests_collector.rs
```diff
@@ -45,6 +47,12 @@ pub struct RequestsCollector {
 const MAX_REQUESTS_LOGGED: usize = 32;
 const QUEUE_CAPACITY: usize = 64;
 
+/// Rate-limit interval for warning logs (seconds)
+const WARN_INTERVAL_SECS: u64 = 10;
+
+/// Last time a send warning was emitted (unix seconds)
+static LAST_SEND_WARN: AtomicU64 = AtomicU64::new(0);
```
timvisee, line 54: Why not `Instant::now()` and `last.elapse() > Duration::from_secs(10)`?

Just to skip a lock, as we prefer not to lock at this stage?

generall: `last.elapse()` would need to have Instannt type, which is not atomic

PR #7438, opened 2025-10-22
Add global metrics prefix

JojiiOfficial:
According to the [Prometheus naming convention](https://prometheus.io/docs/practices/naming/):
```
A metric name SHOULD have a (single-word) application prefix relevant
to the domain the metric belongs to.
```

This PR adds the `qdrant_` prefix to all metrics to align with this convention. This prefix can be configured in the config file.

<br>

**Note, this is potentially breaking and might need discussion before merging**

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::{env, io};
 
 use api::grpc::transport_channel_pool::{
@@ -9,7 +10,7 @@ use common::flags::FeatureFlags;
 use config::{Config, ConfigError, Environment, File, FileFormat, Source};
 use serde::Deserialize;
 use storage::types::StorageConfig;
-use validator::Validate;
+use validator::{Validate, ValidationError};
 
 use crate::common::debugger::DebuggerConfig;
 use crate::common::inference::config::InferenceConfig;
@@ -60,6 +61,7 @@ pub struct ServiceConfig {
 
     /// Global prefix for metrics.
     #[serde(default)]
+    #[validate(custom(function = validate_metrics_prefix))]
     pub metrics_prefix: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


config/config.yaml
```diff
@@ -297,6 +297,10 @@ service:
   #
   # Uncomment to enable.
   # hardware_reporting: true
+  #
+  # Uncomment to enable.
```
JojiiOfficial, line 301: We enable this for now. When we release this among the other new metrics, we can disable it again.

config/config.yaml
```diff
@@ -297,6 +297,10 @@ service:
...
+  # Prefix for the names of metrics in the /metrics API.
+  metrics_prefix: qdrant_
```
generall: should not be enabled by default

timvisee: Agreed. We can probably set it when we launch our "big new metrics revamp".

I see <https://github.com/qdrant/qdrant/pull/7438#discussion_r2451966753>. But I'd prefer to keep it consistent for now and this way we also don't forget to remove it just before the release.

src/settings.rs
```diff
@@ -57,6 +57,10 @@ pub struct ServiceConfig {
     /// Whether to enable reporting of measured hardware utilization in API responses.
     #[serde(default)]
     pub hardware_reporting: Option<bool>,
+
+    /// Global prefix for metrics.
+    #[serde(default)]
+    pub metrics_prefix: Option<String>,
```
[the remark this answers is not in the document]
timvisee: Yes, I think it'd be nice to add something like this. We can probably simplify it a bit:

Require one of:
- empty string
- a string with
  - all chars [`is_ascii_alphanumeric`](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_alphanumeric) or `'_'`

If this validation fails in configuration, Qdrant will only print a warning on startup and will still continue. I actually think that is desired behavior in this case. In the end the user is responsible for configuring it correctly.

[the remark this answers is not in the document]
JojiiOfficial: Implemented in ddf1181c8b91c1b1ec6d832f6af12389ecf62561

timvisee (approved):

PR #7441, opened 2025-10-22
Metrics vectors by name per collection

JojiiOfficial:
Depends on #7310

This PR adds a new field in telemetry and metrics API, providing information about the vector count of each individual vector name.
Initially suggested here https://github.com/qdrant/qdrant/pull/7302#issuecomment-3375560205

New metric:
```
# HELP collection_name_vectors amount of vectors grouped by vector name
# TYPE collection_name_vectors gauge
collection_name_vectors{collection="test2",vector=""} 100
collection_name_vectors{collection="test",vector=""} 2
collection_name_vectors{collection="test",vector="text"} 1
```

New telemetry field:
```json
"num_vectors_by_name": {"text":1 ,"" : 2}
```

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -171,8 +171,7 @@ impl MetricsProvider for CollectionsTelemetry {
         let mut total_min_active_replicas = usize::MAX;
         let mut total_max_active_replicas = 0;
 
-        // Points & Vectors per collection
-        let mut vectors_per_collection = vec![];
+        // Points per collection
         let mut points_per_collection = vec![];
 
         // Vectors excluded from index-only requests.
@@ -250,11 +249,6 @@ impl MetricsProvider for CollectionsTelemetry {
 
             points_per_collection.push(gauge(
                 collection.count_points() as f64,
-                &[("id", &collection.id)],
-            ));
-
-            vectors_per_collection.push(gauge(
-                collection.count_vectors() as f64,
                 &[("id", &collection.id)],
             ));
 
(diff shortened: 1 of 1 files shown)


src/common/metrics.rs
```diff
@@ -240,6 +249,15 @@ impl MetricsProvider for CollectionsTelemetry {
                 .count();
         }
 
+        if !vector_count_by_name.is_empty() {
+            metrics.push(metric_family(
+                "collection_name_vectors",
```
timvisee: I'd vote to rename this to:

```suggestion
                "collection_vectors_total",
```

And totally remove the existing `collections_vector_total`.

That means we remove the aggregation and require users to aggregate this. I think it's fine though.

timvisee (approved):
This looks fine, except for how we report it in metrics.

I suggest to replace the aggregation with the splitted entries: only show the vector count per named vector.

src/common/metrics.rs
JojiiOfficial: Should we also remove `collection_vectors` introduced by #7302 as well?

JojiiOfficial: Removed in 2a8c242e3b743ab06307178274c1b3cc954c3327

lib/collection/src/telemetry.rs
```diff
@@ -75,6 +78,29 @@ impl CollectionTelemetry {
...
+            .fold(
+                TinyMap::<VectorNameBuf, usize>::new(),
+                |mut acc, shard_vectors| {
+                    for (name, count) in shard_vectors {
+                        *acc.get_or_insert_default(name) += count;
+                    }
+                    acc
+                },
+            )
+    }
```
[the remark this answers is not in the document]
timvisee: :rabbit:, you are not a compiler

PR #7443, opened 2025-10-23
Draft: flush segments in two stages

timvisee:
_WIP_

Flush all segments in two stages:
1. inserts and updates
2. deletes

This is a naive implementation that allows us to test this idea. If it works out, we can polish it further.

We've seen a variety of issues where points did go missing. Almost all of them are related to us deleting them on disk too early. Segments have a strong dependency with each other in terms of flushing order to ensure data consistency.

The rough idea of this change is to always flush deletes as very last step. This will hopefully prevent all such issues. We rely on the WAL to repair all cases where we might have partially flushed.

### All Submissions:

* [ ] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
4. [ ] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
5. [ ] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #7446, opened 2025-10-23
Add deferred flusher to Gridstore

timvisee:
Add a deferred flusher to Gridstore, using which we flush asynchronously through a closure.

This makes the flushing behavior of Gridstore consistent with the other storage types we have.

Specifically this:
1. takes a copy of pending updates in the tracker
2. deferred, in a closure:
  a. write the copied updates to disk (ignoring any new updates)
  b. bump the list of pending updates to remove the now persisted entries

This is a building block I need for <https://github.com/qdrant/qdrant/pull/7443>.

As a bonus this removes all `Arc<RwLock<_>>` wrappers around Gridstore, making the world a nicer place :four_leaf_clover:. It now also respects `&self`/`&mut self`.

Crasher run: https://github.com/qdrant/crasher/actions/runs/18755857443 :green_circle: 

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/geo_index/mutable_geo_index.rs
+++ b/lib/segment/src/index/field_index/geo_index/mutable_geo_index.rs
@@ -1,6 +1,7 @@
 use std::cmp::max;
 use std::collections::BTreeMap;
 use std::path::PathBuf;
+#[cfg(feature = "rocksdb")]
 use std::sync::Arc;
 
 use ahash::AHashSet;
@@ -9,6 +10,7 @@ use common::types::PointOffsetType;
 use delegate::delegate;
 use gridstore::Gridstore;
 use gridstore::config::StorageOptions;
+#[cfg(feature = "rocksdb")]
 use parking_lot::RwLock;
 #[cfg(feature = "rocksdb")]
 use rocksdb::DB;
--- a/lib/segment/src/index/field_index/map_index/mutable_map_index.rs
+++ b/lib/segment/src/index/field_index/map_index/mutable_map_index.rs
@@ -2,12 +2,14 @@ use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::iter;
 use std::path::PathBuf;
+#[cfg(feature = "rocksdb")]
 use std::sync::Arc;
 
 use common::counter::hardware_counter::HardwareCounterCell;
 use common::types::PointOffsetType;
 use gridstore::config::StorageOptions;
 use gridstore::{Blob, Gridstore};
+#[cfg(feature = "rocksdb")]
 use parking_lot::RwLock;
 use roaring::RoaringBitmap;
 #[cfg(feature = "rocksdb")]
--- a/lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
+++ b/lib/segment/src/index/field_index/numeric_index/mutable_numeric_index.rs
@@ -2,12 +2,14 @@ use std::collections::BTreeSet;
 use std::ops::Bound;
 use std::ops::Bound::{Excluded, Unbounded};
 use std::path::PathBuf;
+#[cfg(feature = "rocksdb")]
 use std::sync::Arc;
 
 use common::counter::hardware_counter::HardwareCounterCell;
 use common::types::PointOffsetType;
 use gridstore::config::StorageOptions;
 use gridstore::{Blob, Gridstore};
+#[cfg(feature = "rocksdb")]
 use parking_lot::RwLock;
 #[cfg(feature = "rocksdb")]
 use rocksdb::DB;


lib/gridstore/src/gridstore.rs
```diff
@@ -556,33 +559,59 @@ impl<V> Gridstore<V> {
...
-                            self.config.block_size_bytes,
-                        );
-                    start as usize..end as usize
-                });
-                guard.mark_blocks_batch(page_id, local_ranges, false);
+        if deferred_err.is_none()
+            && let Err(err) = bitmask.read().flush()
+        {
+            deferred_err.replace(err);
+        }
```
timvisee: Here we still flush Gridstore pages immediately outside (before) of the deferred closure.

As far as I can see, moving page flushing into the closure means we must wrap all pages in `Arc<RwLock<_>>`, affecting all accesses. I'd like to prevent that.

Flushing before the closure here is still sound. Written blocks are not changed, until we've had a full flush cycle. This is guarded by the bitmask and tracker. If we flush too early and we crash (never invoking the closure), the blocks will just be reused on next start.

I plan to revisit this later once I've completed my other efforts.

lib/gridstore/src/gridstore.rs
```diff
@@ -556,33 +559,59 @@ impl<V> Gridstore<V> {
...
     /// Flush all mmaps and pending updates to disk
     pub fn flush(&self) -> std::result::Result<(), mmap_type::Error> {
-        let mut bitmask_guard = self.bitmask.upgradable_read();
-        bitmask_guard.flush()?;
+        self.flusher()()
+    }
+
+    /// Create flusher that durably persists all pending changes when invoked
+    pub fn flusher(&self) -> Flusher {
+        let pending_updates = self.tracker.read().pending_updates.clone();
```
timvisee, line 567: Key change: here we now take a snapshot of all pending changes. Only these are flushed. This is consistent with our other flushers.

lib/gridstore/src/tracker.rs
```diff
@@ -186,7 +250,14 @@ impl Tracker {
...
             }
-            old_pointers.extend(updates.into_outdated_pointers());
+            old_pointers.extend(updates.to_outdated_pointers());
+
+            // Bump the pending updates for this point offset, drop entry if fully persisted
+            if let Some(pending_updates) = self.pending_updates.get_mut(&point_offset)
+                && pending_updates.drain_persisted_and_drop(&updates)
+            {
+                self.pending_updates.remove(&point_offset);
+            }
```
timvisee, line 260: This function now takes the snapshot of pending changes.

Here we bump the latest pending changes, so that we drain all changes that we now persist.

Once this function is complete, the list of pending changes will only contain changes that still need to be persisted.

lib/gridstore/src/gridstore.rs
timvisee: Not relevant anymore. I've now also moved page flushing inside the closure. This makes everything a lot easier to understand.

The flush logic now looks very similar to what it was before - but now in a closure.

Done in https://github.com/qdrant/qdrant/pull/7446/commits/627d31395127ac89d785084a4bfae2a734f6c3df

timvisee:
This now has a bunch of tests added to ensure correct behavior. The tests are actually the majority of this PR.

lib/gridstore/src/gridstore.rs
```diff
@@ -556,33 +559,59 @@ impl<V> Gridstore<V> {
...
+                    let local_ranges = pointer_group.map(|pointer| {
+                        let start = pointer.block_offset;
+                        let end = pointer.block_offset
+                            + Self::blocks_for_value(pointer.length as usize, block_size_bytes);
+                        start as usize..end as usize
+                    });
+                    guard.mark_blocks_batch(page_id, local_ranges, false);
+                }
+            });
+            bitmask_guard.flush()?;
```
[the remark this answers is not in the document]
timvisee: It is fine. We don't hold multiple locks at the same there, so it cannot deadlock.

lib/gridstore/benches/flush_bench.rs
```diff
@@ -37,7 +37,7 @@ pub fn flush_bench(c: &mut Criterion) {
                     // Benchmark the flush operation after accumulating updates
                     let instant = Instant::now();
 
-                    storage.flush().unwrap();
+                    storage.flusher()().unwrap();
```
agourlay, line 40: Could you please run the `flush_bench` for a quick sanity check that the performance is similar to `dev` 🤞

timvisee: image

It is worse than I expected. It may just be because now there's extra work. Not sure yet if we have to be very concerned about it.

agourlay: I have done some basic profiling on the benchmark and I cannot find a clear root cause for the degradation.

The previous code was sub-millisecond fast already, maybe the additional locking makes things slower 🤷 

I think this is fine.

timvisee: 10 μs even for 100 updates. Likely measuring locks. 

Agreed 👍

coszio: `bustle_bench` does not show any noticeable degradation 👍

timvisee:
Let's postpone merging this till after the 1.16 release. Then we can test it on chaos testing for some time without the risk of affecting the release.

Issue #4235, opened 2024-05-14
Custom positive-negative strategy


pavelshutov-pr3vision:
Hello everyone!

I have one question: is it possible to simply implement my own strategy for positive and negative vectors (or modify an existing [best-score strategy](https://qdrant.tech/documentation/concepts/explore/#best-score-strategy)) for another specific situation?

For example, if I understood correctly:
- the best-score strategy has a strict condition: if an element has a maximum score with positive vectors lower than with negative ones, then the element is excluded from the results due to a low and negative score:
```rust
let score = if best_positive_score > best_negative_score {
    best_positive_score;
} else {
    -(best_negative_score * best_negative_score);
};
```

I'm looking for another way — instead of a hard restriction, make a softer one, for example:
```rust
let score = best_positive_score - best_negative_score;
```

How difficult is it to implement, maybe I can just rewrite the score function or something like that? Where can I start?

Thanks in advance. And I'm sorry if my question is too dumb

generall:
It feels like the [Average vector strategy](https://qdrant.tech/documentation/concepts/explore/#average-vector-strategy) should give you something closer to what you are asking. But if not and you really see a benefit in implementing this new strategy, it is, of course, possible to implement one.

PR #7447, opened 2025-10-23
[WIP/RFC] feat: add /collections/{name}/points/average_vector

CyberShadow:
Allows quickly retrieving the average vector from existing points, e.g. to enable custom recommendation search.

Example use case: for https://github.com/qdrant/qdrant/issues/4235, get the average of the positive and negative examples, and apply custom weights with the Rocchio algorithm.

Still needs work but what do you think, worth adding?

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/api/src/grpc/proto/points_service.proto
+++ b/lib/api/src/grpc/proto/points_service.proto
@@ -79,6 +79,10 @@ service Points {
   */
   rpc RecommendGroups (RecommendPointGroups) returns (RecommendGroupsResponse) {}
   /*
+  Compute the average vector from a list of point IDs or vectors
+  */
+  rpc AverageVector (AverageVectorPoints) returns (AverageVectorResponse) {}
+  /*
   Use context and a target to find the most similar points to the target, constrained by the context.
 
   When using only the context (without a target), a special search - called context search - is performed where
--- a/lib/api/src/grpc/proto/points.proto
+++ b/lib/api/src/grpc/proto/points.proto
@@ -467,6 +467,16 @@ message RecommendBatchPoints {
   optional uint64 timeout = 4; // If set, overrides global timeout setting for this request. Unit is seconds.
 }
 
+message AverageVectorPoints {
+  string collection_name = 1; // Name of the collection
+  repeated VectorExample examples = 2; // List of point IDs or vectors to average
+  optional string using = 3; // Define which vector to use, if not specified - use default vector
+  optional LookupLocation lookup_from = 4; // Name of the collection to use for points lookup, if not specified - use current collection
+  optional ReadConsistency read_consistency = 5; // Options for specifying read consistency guarantees
+  optional uint64 timeout = 6; // If set, overrides global timeout setting for this request. Unit is seconds.
+  optional ShardKeySelector shard_key_selector = 7; // Specify in which shards to look for the points, if not specified - look in all shards
+}
+
 message RecommendPointGroups {
   string collection_name = 1; // Name of the collection
   repeated PointId positive = 2; // Look for vectors closest to the vectors from these points
@@ -995,6 +1005,12 @@ message GetResponse {
 
 message RecommendResponse {
   repeated ScoredPoint result = 1;
+  double time = 2; // Time spent to process
+  optional Usage usage = 3;
+}
+
+message AverageVectorResponse {
+  Vector vector = 1; // The computed average vector
   double time = 2; // Time spent to process
   optional Usage usage = 3;
 }
--- a/lib/api/src/grpc/proto/points_internal_service.proto
+++ b/lib/api/src/grpc/proto/points_internal_service.proto
@@ -22,6 +22,7 @@ service PointsInternal {
   rpc Scroll (ScrollPointsInternal) returns (ScrollResponse) {}
   rpc Count (CountPointsInternal) returns (CountResponse) {}
   rpc Recommend (RecommendPointsInternal) returns (RecommendResponse) {}
+  rpc AverageVector (AverageVectorPointsInternal) returns (AverageVectorResponse) {}
   rpc Get (GetPointsInternal) returns (GetResponse) {}
   rpc QueryBatch (QueryBatchPointsInternal) returns (QueryBatchResponseInternal) {}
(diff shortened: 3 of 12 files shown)


generall:
hey @CyberShadow, thanks for your interest in our project!

> Allows quickly retrieving the average vector from existing points

could you please elaborate that is the difference of your proposed search to https://qdrant.tech/documentation/concepts/explore/#recommendation-api ?


> and apply custom weights with the Rocchio algorithm.

I didn't find custom weights in your proposal, but to my understanding the same effect could be achieved if you multiply vectors themselves by coefficients. Please correct me if I am wrong

lib/api/src/grpc/proto/points_service.proto
```diff
@@ -79,6 +79,10 @@ service Points {
   */
   rpc RecommendGroups (RecommendPointGroups) returns (RecommendGroupsResponse) {}
   /*
+  Compute the average vector from a list of point IDs or vectors
+  */
+  rpc AverageVector (AverageVectorPoints) returns (AverageVectorResponse) {}
```
generall, line 84: normally, we would prefer to implement new type of scoring strategy as a part of `query` endpoint, instead of introducing a new one

generall:
Also, since you mentioned Rocchio, you might be interested to take a loot at this https://github.com/qdrant/qdrant/milestone/38?closed=1

CyberShadow:
Hi @generall ,

> could you please elaborate that is the difference of your proposed search to https://qdrant.tech/documentation/concepts/explore/#recommendation-api ?

I see multiple use cases for getting the averages, a custom search being one. Another use case is getting the average vector in order to then quickly calculate the distance to any point without first having to find it in a search result.

The recommendation API is currently a little limited - the flow is as follows:

1. resolve inputs to vector
2. calculate average vector
3. calculate search target vector
4. search for points near target vector

Steps 2 and 3 in particular are not customizable, e.g. there is no way to specify a custom coefficient to the negative vector delta as per Rocchio.

> I didn't find custom weights in your proposal, but to my understanding the same effect could be achieved if you multiply vectors themselves by coefficients. Please correct me if I am wrong

I'm not aware of a way to specify a list of point IDs + coefficients, did I miss something in the API?

> normally, we would prefer to implement new type of scoring strategy as a part of `query` endpoint, instead of introducing a new one

Another way to approach this would be to add a parameter to the recommendation endpoint to return averages; then, callers could simply enable that option + limit results to 0 points to get just the averages. Does that sound better?

> Also, since you mentioned Rocchio, you might be interested to take a loot at this https://github.com/qdrant/qdrant/milestone/38?closed=1

Thanks, looks interesting! I don't have a use case for that type of request but it definitely looks like a useful addition in this space. I guess you could also specify `a = 0` and specify an arbitrary initial vector to get a more tunable recommendation search.

generall:
I think I misunderstood your suggestion. I see now, that you don't want to search anything, just compute averate vector of sotred points. Then indeed, it doesn't match the intent of query endpoint.

My followup question then would be:

- What is the advantage over retrieving points by IDs and averaging vectors on the client side? I assume avg operation is cheap, and only thing we actually gain is lower network traffic, is that right?

I also assume, that number of points to average expected to be low (p99 < 10 vectors), otherwise avg vector will likely be meaningless. If that is a case, does network traffic overhead really worth maintaining a dedicated API for this?

CyberShadow:
> What is the advantage over retrieving points by IDs and averaging vectors on the client side? I assume avg operation is cheap, and only thing we actually gain is lower network traffic, is that right?

Correct (plus encoding/decoding overhead). That was the first approach I tried - even on localhost for me this operation takes many minutes, whereas Qdrant takes about a second.

> I also assume, that number of points to average expected to be low (p99 < 10 vectors), otherwise avg vector will likely be meaningless. If that is a case, does network traffic overhead really worth maintaining a dedicated API for this?

I'm not sure why you say it would be meaningless? E.g. you can use it to quickly obtain distance as a metric of how close an arbitrary point (even without uploading it to Qdrant) is to the "class archetype", regardless of class population. This value is used in recommendation search regardless of class size as well which helps building custom client-side algorithms as needed.

generall:
> whereas Qdrant takes about a second.

How many points do you need to average? This sounds like a lot

CyberShadow:
Up to 100,000 (but multiplied by a few hundred classes). It actually might be less than a second!

generall:
Could you please also explain an interface a little bit more:

- if we are interested in getting average of the collection subset (aka  "class archetype"), why would we need to provide any additional vectors in the query itself? Isn't just a list is enough?

- If you need to average 100k points, wouldn't it be more convenient to provide a filter condition instead of just a huge list of IDs?

CyberShadow:
> * if we are interested in getting average of the collection subset (aka  "class archetype"), why would we need to provide any additional vectors in the query itself? Isn't just a list is enough?

Correct, here the list is the field called `examples` (I do need to pick a better name!), reusing the type from the recommend query - the expected use case is to supply point IDs (vectors are accepted but would be meaningless except unless combined with point IDs).

> * If you need to average 100k points, wouldn't it be more convenient to provide a filter condition instead of just a huge list of IDs?

Indeed, however this assumes that class labels are tracked in Qdrant payloads - in my case they are not. I would expect that if they are, requesting point IDs and then passing them back in would still be faster than downloading and averaging the vectors.

CyberShadow:
> If you need to average 100k points, wouldn't it be more convenient to provide a filter condition instead of just a huge list of IDs?

BTW, on this note, I am thinking that the question is roughly the same as Qdrant allowing a recommendation search straight from the results of a filter condition. Does that make sense?

generall:
> allowing a recommendation search straight from the results of a filter condition

Averaging more than a few vectors was never an intended usage of recommendation API, so we didn't do that. 
Also, in some conditions too broad filter can easity make averaging very expensive, especially if vectors are stored on disk and not cached in ram.

CyberShadow:
Makes sense.

Let me know which way you want take this (or I'm happy to continue using a local fork if this is not generally useful).

generall:
We will try to figure out maybe a more universal interface, which would cover your use-case as well. So far it looks like a special case of aggregation request

PR #7448, opened 2025-10-23
Do not merge threadpool experiment

generall:
An exaple of using different threading layout for segment search.

Current dev:

- Assigns search in each segment to independent thread

This PR:

- Assign search in all segments to a single thread

Intuition:

Maybe if segments are small, the overhead for moving tasks to different threads is greated than benefit of threading?

Does the intuition work? **No**

### DIFF
--- a/config/development.yaml
+++ b/config/development.yaml
@@ -21,9 +21,9 @@ cluster:
   resharding_enabled: true
 
 storage:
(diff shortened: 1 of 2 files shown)


generall:
We understood that the porblem is not a threadpool, but memory bandwidth

PR #7450, opened 2025-10-23
remove filter level jwt

generall:
Payload filter in JWT was deprecated in 1.15.0

This PR removes it

### DIFF
--- a/lib/storage/src/rbac/ops_checks.rs
+++ b/lib/storage/src/rbac/ops_checks.rs
@@ -363,7 +363,7 @@ mod tests_ops {
     ) {
         access
             .check_point_op("col", op)
-            .expect_err("Should be allowed");
+            .expect_err("should be forbidden");
     }
 
     /// Operation requires write + whole collection access.


lib/storage/src/rbac/mod.rs
```diff
@@ -38,18 +38,31 @@ pub struct CollectionAccess {
 
     /// Payload constraints.
     /// An object where each key is a JSON path, and each value is JSON value.
+    ///
+    /// Deprecation: this parameter is kept for preventing old keys to become valid after parameter removal.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[deprecated(since = "1.15.0")]
+    #[validate(custom(function = "validate_payload_empty"))]
```
timvisee, line 44: **Thanks!** Came back to check exactly this, happy to see you implemented it this way!

lib/storage/src/rbac/mod.rs
```diff
@@ -38,18 +38,31 @@ pub struct CollectionAccess {
...
     pub payload: Option<PayloadConstraint>,
```
timvisee: We can change it to be a [`Value`](https://docs.rs/serde_json/latest/serde_json/enum.Value.html) so that we can drop the `PayloadConstraint` type while still allowing us to deserialize these.


```diff
- pub payload: Option<PayloadConstraint>,
+ pub payload: Option<serde_json::Value>,
```

timvisee (commented):
Partial review:

lib/storage/src/rbac/mod.rs
timvisee: Implemented in https://github.com/qdrant/qdrant/pull/7450/commits/ecd6de1c34d50a02d0d3d659f3653d73a3453276

PR #7452, opened 2025-10-24
[#7446] Refactor draining pointer updates

coszio:
A bit of refactoring for #7446

### DIFF
--- a/lib/gridstore/src/tracker.rs
+++ b/lib/gridstore/src/tracker.rs
@@ -71,8 +71,8 @@ impl PointerUpdates {
         }
     }
 
-    /// Returns pointers that need to be freed, i.e. They have been written, and are no longer needed
-    fn to_outdated_pointers(&self) -> impl Iterator<Item = ValuePointer> {
+    /// Returns pointers that need to be freed, i.e. The locations they are pointing to are no longer needed.
+    fn outdated_pointers(&self) -> impl Iterator<Item = ValuePointer> {
         let take = if self.latest_is_set {
             // all but the latest one
             self.history.len().saturating_sub(1)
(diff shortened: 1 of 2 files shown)


lib/gridstore/src/tracker.rs
```diff
@@ -119,22 +118,19 @@ impl PointerUpdates {
...
+                // Latest was set before, but now it is not.
+                // We need to process this last pointer on the next flush
+                let start = persisted.history.len().saturating_sub(1);
+                self.history = self.history[start..].into()
+            }
+            _ => {
+                // in all other cases, we can get rid of persisted history
+                self.history = self.history[persisted.history.len()..].into()
+            }
+        };
```
timvisee, line 133: I really like this simplification, and taking the inverse (start.. instead of ..end).

I'm quite sure this won't be compatible with the two staged flusher because it breaks chronological order of histories. But I'm still happy to merge this change, as I can always adjust it later.

But: 

---

I believe this match statement assumes that `(false, true)` is not possible, right?

Because we cannot set a block again without fully flushing the unset first.

If that's true I vote to add a debug assertion here. And maybe even use the same logic as in the first branch to handle it if we magically ever hit it.

coszio: I analyzed statically. If it was not set before, and now it is set
```
persisted:     pending:
set 0          set 0
unset (0)      unset (0)
               set 1
```
we can just remove the persisted history, and it HAS to have a new pointer to correspond with the set operation

coszio:
Oops, tests fail

coszio:
all this is bound to change, and tests are failing right now, so will close this PR

PR #7455, opened 2025-10-24
Expose `AcornSearchParams` to edge Python bindings

xzfc:
Follow-up for #7414.
This PR adds a new field to Python bindings.


generall (approved):

ffuugoo:
> `#[derive(Copy, Clone, Debug, Into)]`

Thanks for putting derives in The Correct Order™️ 🥹

PR #7456, opened 2025-10-24
Misc edge setup fixes

xzfc:
- Add `.venv` to the project `.gitignore` as it is mentioned in the [Python Bindings readme](https://github.com/qdrant/qdrant/blob/c1c9eb9ddbf76fa38cee3cee639ddadc75ea5382/lib/edge/python/README.md). Seems Maturin don't like the name `venv` without leading dot.
- Remove `mkdir shard` from the readme. (in the [example](https://github.com/qdrant/qdrant/blob/c1c9eb9ddbf76fa38cee3cee639ddadc75ea5382/lib/edge/python/examples/qdrant-edge.py#L9), the `DATA_DIRECTORY` is `"data"`, not `"shard"`; and it created automatically)
- Add `/data` to the examples' `.gitignore`.
- Add `#!/usr/bin/env python3` and `chmod +x` for the example script.
- Let the example script write to the script's directory (`os.path.dirname(__file__)`), not the current directory, so it won't end up with a random non-ignored files when running outside its dir.
- Add `maturin` package to `shell.nix`


ffuugoo (approved):
That's actually much better. Thanks!

PR #7457, opened 2025-10-25
extra cloud params for web-ui customization

generall:
This PR adds some internal Settings (and associated ENV parameteres) to be propagated into web-ui from cloud.

For example:

```
QDRANT__CLOUD_PARAMS__BACKLINK='https://cloud.qdrant.io/accounts/db3db48e-678f-463a-bf87-52ddc33ee788/clusters/61768340-a723-4a60-ab07-bbf74275ee0d/overview'
```

Would allow web-ui to render navigation link back to qdrant cloud. 

Unfortunately, not all planned parameters are suitable for transferring over ENV, as ENV can only be changed on cluster restart (e.g. cluster name,  user name, available upgrade), so we would need to find other ways to get this information

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -11687,6 +11687,7 @@
       "AppBuildTelemetry": {
         "type": "object",
         "required": [
+          "cloud_params",
           "name",
           "startup",
           "version"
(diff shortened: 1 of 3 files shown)


timvisee (approved):
I assume Web UI will just fetch these from telemetry, and we don't have to make any other changes in this PR.

generall:
We decided to use mounted volume to propagate cloud params into UI explicitly

PR #7459, opened 2025-10-27
Move Query Context into shard crate

generall:
Query context is required for proper query execution, therefore we must have it in Shard instead of Collection.

There are problems:

- IDF Modifier parameter for sparse vector is not known on shard level, it should be. This PR moves it into segment

### DIFF
--- a/lib/shard/src/query/query_context.rs
+++ b/lib/shard/src/query/query_context.rs
@@ -1,7 +1,6 @@
 use common::counter::hardware_accumulator::HwMeasurementAcc;
 use segment::data_types::query_context::QueryContext;
 use segment::types::VectorName;
-use smallvec::SmallVec;
 
 use crate::common::stopping_guard::StoppingGuard;
 use crate::search::CoreSearchRequest;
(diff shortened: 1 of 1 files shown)


ffuugoo (approved):

lib/shard/src/query/query_context.rs
```diff
@@ -0,0 +1,63 @@
...
+        search_request
+            .query
+            .iterate_sparse(|vector_name, sparse_vector| {
+                if idf_vectors.contains(&vector_name) {
+                    query_context.init_idf(vector_name, &sparse_vector.indices);
+                }
+            })
+    }
+
+    query_context
```
ffuugoo: Though, why can't it be simplified to something like this?


```suggestion
    let mut query_context = QueryContext::new(search_optimized_threshold_kb, hw_measurement_acc)
        .with_is_stopped(is_stopped_guard.get_is_stopped());

    for search_request in batch_request {
        search_request
            .query
            .iterate_sparse(|vector_name, sparse_vector| {
                if check_idf_required(&vector_name) {
                    query_context.init_idf(vector_name, &sparse_vector.indices);
                }
            })
    }

    query_context
```

generall: ~in this case we will do `init_idf` for the same vector multiple times, that will corrupt statistics~

scratch that

PR #7463, opened 2025-10-28
Add function for `indexed_only` decision

timvisee:
Improves <https://github.com/qdrant/qdrant/pull/7307>

Add a `in_indexed_only_search()` function to segments and their indices. Allow us to use just this function to decide how `indexed_only` behaves.

This specifically improves the way we list excluded points for <https://github.com/qdrant/qdrant/pull/7307> by keeping the `indexed_only` decision within segments.

### All Submissions:

### DIFF
--- a/lib/collection/src/shards/telemetry.rs
+++ b/lib/collection/src/shards/telemetry.rs
@@ -65,6 +65,8 @@ pub struct LocalShardTelemetry {
     pub optimizations: OptimizerTelemetry,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub async_scorer: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub index_only_excluded_vectors: Option<HashMap<String, usize>>,
 }
 
 #[derive(Serialize, Clone, Debug, JsonSchema, Anonymize, Default)]
--- a/lib/collection/src/shards/telemetry.rs
+++ b/lib/collection/src/shards/telemetry.rs
@@ -66,7 +66,7 @@ pub struct LocalShardTelemetry {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub async_scorer: Option<bool>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub index_only_excluded_vectors: Option<HashMap<String, usize>>,
+    pub indexed_only_excluded_vectors: Option<HashMap<String, usize>>,
 }
 
 #[derive(Serialize, Clone, Debug, JsonSchema, Anonymize, Default)]
--- a/lib/segment/src/segment/entry.rs
+++ b/lib/segment/src/segment/entry.rs
@@ -462,6 +462,21 @@ impl SegmentEntry for Segment {
         Ok(size)
     }
 
+    fn in_indexed_only_search(
+        &self,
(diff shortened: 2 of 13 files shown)


lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -533,6 +533,45 @@ impl SegmentEntry for ProxySegment {
...
+        //     filter,
+        //     self.deleted_points.keys().copied(),
+        // );
+
+        // self.wrapped_segment.get().read().in_indexed_only_search(
+        //     vector_name,
+        //     search_optimized_threshold_kb,
+        //     Some(&wrapped_filter),
+        //     hw_counter,
+        // )
```
timvisee, line 572: The current behavior is actually the above. But it might be more correct to use this commented-out section instead.

I'd have to think about it for a bit.

If we do enable this, we must also adjust how we use the parameter in search.

lib/segment/src/segment/entry.rs
```diff
@@ -462,6 +462,21 @@ impl SegmentEntry for Segment {
...
 
+    fn in_indexed_only_search(
+        &self,
+        vector_name: &VectorName,
+        search_optimized_threshold_kb: usize,
+        filter: Option<&Filter>,
+        hw_counter: &HardwareCounterCell,
+    ) -> OperationResult<bool> {
+        check_vector_name(vector_name, &self.segment_config)?;
+        let include = self.vector_data[vector_name]
```
JojiiOfficial, line 473: I think a checked access here via `.get()` should be used instead to prevent panicking.
In my [original implementation](https://github.com/qdrant/qdrant/blob/2f25d7da6eaee68106db35d1c04e863a1b65dcf6/lib/collection/src/shards/local_shard/telemetry.rs#L181), it actually panicked in tests because one of the names in `segment.vector_names()` didn't have a `vector_data` entry.

To propagate the availability info, I'd suggest this function to return a `OperationResult<Option<bool>>`.

timvisee: The `check_vector_name(...)?;` before it should prevent it. But I agree it's probably good to be cautious, everywhere.

I'll apply this change to `dev` in a separate PR because we have more usages like this.

JojiiOfficial (approved):

lib/segment/src/segment/entry.rs
timvisee: https://github.com/qdrant/qdrant/pull/7469

I can bump this PR once that is merged.

timvisee:
We can postpone merging this until after the Qdrant 1.16 release. That gives us a bit more time to decide upon <https://github.com/qdrant/qdrant/pull/7463#pullrequestreview-3388291146>.

lib/collection/src/shards/telemetry.rs
coszio, line 69: This name seems complicated and depends on a specific query parameter. 
```suggestion
    pub unindexed_vectors: Option<HashMap<String, usize>>,
```

Do we even need to expose this?

Idea: do not expose this new metric, and instead, consider those unoptimized segments with less than `full_scan_threshold` as optimized, and include them in the `total_optimized_points` count.

timvisee:
This is stale, and I'm not sure it's worth the effort to revive it

PR #7467, opened 2025-10-28
Don't reuse links with vectors, optimize `OldIndexCandidate::evaluate`

xzfc:
# 1. Don't reuse links with vectors

This PR disables incremental HNSW for graphs in an `inline_storage` format. Reason: evaluating/reusing makes a lot of random reads, making it slower than building from scratch. E.g., on cohere dataset, 768 dims, 500k points: 25 seconds to build HNSW from scratch vs 130 seconds to evaluate/reuse links.

# 2. Tiny optimization: `links_empty`

Also, this PR optimizes `OldIndexCandidate::evaluate` a bit.

When merging two segments of the same size, both are evaluated, but at most one is used. Previously, `OldIndexCandidate::evaluate` were unpacking a single neighbor of a point to determine whether the graph contains this point[^1]. This were causing random reads even for unused indices.

In this PR, a new method is added (`links_empty(point_id, level) -> bool`). It checks `level_offsets`/`reindex` but doesn't read/unpack the links.

On low-dimensional datasets (e.g. `deep-image-96`) this change makes overall time of merging two segments of the same size ~5% faster when the index is on disk.

[^1]: since otherwise it's not possible to distinguish between "graph built without that particular point/vector" vs "this particular point/vector deleted after the graph was built".

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_links/view.rs
+++ b/lib/segment/src/index/hnsw_index/graph_links/view.rs
@@ -212,6 +212,8 @@ impl GraphLinksView<'_> {
         })
     }
 
+    /// Note: it is safe to use `idx + 1` on the result of this function,
+    /// because `level_offsets` always contains an additional element.
     #[inline]
     fn offset_idx(&self, point_id: PointOffsetType, level: usize) -> usize {
         if level == 0 {
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -1010,7 +1010,8 @@ impl HNSWIndex {
                 // ACORN is not implemented for graph with vectors yet (but possible)
                 SearchAlgorithm::Acorn => return Ok(None),
             }
-            if !self.graph.has_vectors() || !is_quantized_search(quantized_vectors.as_ref(), params)
+            if !self.graph.has_inline_vectors()
+                || !is_quantized_search(quantized_vectors.as_ref(), params)
             {
                 return Ok(None);
             }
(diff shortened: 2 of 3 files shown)


lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -1543,7 +1543,10 @@ impl<'a> OldIndexCandidate<'a> {
             || config.m0 != old_index.config.m0
             || config.ef_construct != old_index.config.ef_construct
             || new_quantization_config != old_quantization_config;
-        if no_main_graph || configuration_mismatch {
+        // If old graph has vectors, reusing it will cause a lot of random reads,
+        // making it slower than building from scratch.
+        let old_graph_is_with_vectors = old_index.graph.has_vectors();
+        if no_main_graph || configuration_mismatch || old_graph_is_with_vectors {
```
timvisee, line 1550: Nit: renaming `has_vectors()` to `has_inline_vectors()` is a little bit more intuitive to me.

I was confused because I'd argue all graphs have _vectors_.

Alternatively I'm fine with just this change:


```suggestion
        // If old graph has inline vectors, reusing it will cause a lot of random reads,
        // making it slower than building from scratch.
        let old_graph_is_with_vectors = old_index.graph.has_vectors();
        if no_main_graph || configuration_mismatch || old_graph_is_with_vectors {
```

timvisee (approved):

lib/segment/src/index/hnsw_index/graph_links/view.rs
```diff
@@ -212,13 +212,34 @@ impl GraphLinksView<'_> {
...
-        };
+        }
+    }
 
+    /// Returns `true` if [`Self::links`] would return an empty iterator.
+    pub(super) fn links_empty(&self, point_id: PointOffsetType, level: usize) -> bool {
+        let idx = self.offset_idx(point_id, level);
+        match self.compression {
+            CompressionInfo::Uncompressed { offsets, .. } => {
+                offsets[idx].get() == offsets[idx + 1].get()
```
generall, line 231: could you please comment why `offsets[idx + 1]` is not out of bound

IvanPleshkov (approved):

PR #7474, opened 2025-10-29
Add `ReadActive` replica state, use in some stream records transfers

timvisee:
Extend <https://github.com/qdrant/qdrant/pull/7361>

Add a new `ReadActive` replica state.

The functions as intermediate state between `Partial` and `Active`. We now use this at the end of stream records transfers that point to a different shard ID.

The new state functions as `Active` for readers, and `Partial` for writers.

We need this as extra synchronization step for transfers. It ensures that we switch readers before writers. Otherwise, a lagging (consensus) node may direct a read to stale data. When we switch to the new state in the stream records transfer, we use explicit synchronization and confirm that all nodes have applied the state switch before the transfer actually finalizes.

Follows the design discussed here: https://www.notion.so/qdrant/Multi-tenancy-transfer-synchronization-29a674779d33807ba196e653dbc086d0?source=copy_link

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -475,6 +475,15 @@ impl ShardReplicaSet {
             .collect()
     }
 
+    pub fn readable_shards(&self) -> Vec<PeerId> {
+        let replica_state = self.replica_state.read();
+        replica_state
+            .readable_peers() // This includes `ActiveRead`, `Active`, and `ReshardingScaleDown` replicas!
+            .into_iter()
+            .filter(|&peer_id| !self.is_locally_disabled(peer_id))
+            .collect()
+    }
+
     /// Wait for a local shard to be initialized.
     ///
     /// Uses a blocking thread internally.
@@ -1026,17 +1035,7 @@ impl ShardReplicaSet {
     fn peer_is_readable(&self, peer_id: PeerId) -> bool {
         let peer_state = self.peer_state(peer_id);
         let is_readable = match peer_state {
-            Some(ReplicaState::Active) => true,
-            Some(ReplicaState::ReshardingScaleDown) => true,
-            Some(ReplicaState::ActiveRead) => true,
-            // False from here on
-            Some(ReplicaState::Dead) => false,
-            Some(ReplicaState::Partial) => false,
-            Some(ReplicaState::Initializing) => false,
-            Some(ReplicaState::Listener) => false,
-            Some(ReplicaState::PartialSnapshot) => false,
-            Some(ReplicaState::Recovery) => false,
-            Some(ReplicaState::Resharding) => false,
+            Some(state) => state.is_readable(),
             None => false,
         };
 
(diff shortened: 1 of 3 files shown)


lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -1281,6 +1282,8 @@ pub enum ReplicaState {
     Resharding,
     // Points are being migrated to this shard as part of resharding down
     ReshardingScaleDown,
+    // Active for readers, Partial for writers
+    ReadActive,
```
timvisee: Nit: I might like the name `ActiveRead` a little bit better. As I consider it a sub variant of `Active`.

lib/collection/src/shards/transfer/stream_records.rs
```diff
@@ -24,18 +30,39 @@ pub(super) const TRANSFER_BATCH_SIZE: usize = 100;
...
     let cutoff;
+    let filter = transfer_config.filter;
     let merge_points = filter.is_some();
 
+    // Whether we need an intermediate replica state (ReadActive) during transfer to sync nodes
+    // We use this when transferring between different shard IDs to ensure data consistency, this
+    // way all readers can be switched to the new shard before any writers
+    let sync_intermediate_state = transfer_config
+        .to_shard_id
+        .is_some_and(|id| transfer_config.shard_id == id);
```
KShivendu:
```suggestion
        .is_some_and(|id| transfer_config.shard_id != id);
```

I think this is what you intended to have?

KShivendu: We could also consider `&& filter.is_some()` in this

KShivendu (commented):
The changes make sense. There was a small mistake that I've fixed. I checked that it sets the right state via wal_inspector. 

I'm gonna add a few integration tests now.

lib/collection/src/shards/transfer/stream_records.rs
timvisee: Good call on the shard ID!

> We could also consider && filter.is_some() in this

It is not necessary. We would need the same intermediate state if we'd transfer to another shard ID without a filter. So lets keep it as is. We don't do such transfers yet, but it's good to think ahead here and not make the condition too strict.

lib/collection/src/shards/transfer/stream_records.rs
```diff
@@ -24,18 +30,39 @@ pub(super) const TRANSFER_BATCH_SIZE: usize = 100;
...
+    let filter = transfer_config.filter;
     let merge_points = filter.is_some();
 
+    // Whether we need an intermediate replica state (ActiveRead) during transfer to sync nodes
+    // We use this when transferring between different shard IDs to ensure data consistency, this
+    // way all readers can be switched to the new shard before any writers
+    let sync_intermediate_state = transfer_config
+        .to_shard_id
+        .is_some_and(|id| transfer_config.shard_id != id);
+
```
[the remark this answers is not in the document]
generall: you are not a compiler. You don't check compilation error. Stop doing that

generall:
I propose to refactor https://github.com/qdrant/qdrant/blob/3ff0a3482128ef55f36b343a842f7ea3b415482d/lib/collection/src/shards/replica_set/execute_read_operation.rs#L147

with the idea to deconstrcut `peer_is_active`  [here](https://github.com/qdrant/qdrant/blob/3da6527c8e8731726fedce39f79edfb0cf10de0b/lib/collection/src/shards/replica_set/mod.rs#L1016) into `peer_readable` and `peer_is_active`.
Also avoid usage of `matches!` macro, as it doesn't enforce exhaustive match.

generall (approved):

src/common/health.rs
```diff
@@ -325,7 +325,7 @@ impl Task {
                     continue;
                 };
 
-                if state.is_active_or_listener_or_resharding() {
+                if state.is_healthy() {
                     continue;
                 }
```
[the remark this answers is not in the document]
generall: `ActiveRead` require shard transfer operation to finish. So we consider it as not-healthy.

PR #7475, opened 2025-10-29
Re-style .proto files

xzfc:
This PR improves the style of our `.proto` files. Commits in this PR:

## 1. doc comment style (automated)

Currenly, our `.proto` files use three different styles of comments: line comments on a separate line, multi-line block comments, and line comments at the end of a line. This snippet illustrates all three of them: https://github.com/qdrant/qdrant/blob/96a7a707595a92fc93890db72f4d1b2aaa0b051d/lib/api/src/grpc/proto/collections.proto#L292-L305

The first commit in this PR converts all comments into a single style which is proposed by the [Protobuf Language Guide](https://protobuf.dev/programming-guides/editions/):
> Prefer C/C++/Java line-end-style comments ‘//’ on the line before the .proto code element

<details><summary>Conversion is done by running this script.</summary>

```python
#!/usr/bin/env python3

import os
import re
import sys

RE_MULTILINE_COMMENT_START = re.compile(r"^( *)/\*\*?$")
RE_MULTILINE_COMMENT_END = re.compile(r"^ *\*/$")
RE_TRAILING_COMMENT = re.compile(r"^( *)([^ ].*;) +// (.*)$")

spaces = ""
in_comment = False
for fname in sys.argv[1:]:
    result = []
    with open(fname, "r") as file:
        for line in file:
            line = line.rstrip("\n")

            if m := RE_MULTILINE_COMMENT_START.match(line):
                assert not in_comment
                spaces = m[1]
                in_comment = True
            elif RE_MULTILINE_COMMENT_END.match(line):
                assert in_comment
                in_comment = False
            elif m := RE_TRAILING_COMMENT.match(line):
                assert not in_comment
                spaces, code, comment = m.groups()
                result.append(f"{spaces}// {comment}\n")
                result.append(f"{spaces}{code}\n")
            else:
                if in_comment:
                    line = line.lstrip()
                    result.append(f"{spaces}//{" " + line if line else ""}\n")
                else:
                    result.append(f"{line}\n")
    with open(fname + ".new", "w") as file:
        file.writelines(result)
    os.replace(fname + ".new", fname)
```

</details>

## 2. clang-format (automated)

The second commit runs `clang-format` to unify indenting/spaces. I believe that's the tool the company behind protobuf uses for their files.

```bash
clang-format \
	-style='{"BasedOnStyle": "Google", "ReflowComments": false}' \
	-i lib/api/src/grpc/proto/*.proto
```

While I like auto-reflow, it's too blunt for an automated mass-edit, so I disabled it and manually split some of the longer lines.

## 3. manually reflow some comments

See line above.

### DIFF
--- a/lib/api/src/grpc/qdrant.rs
+++ b/lib/api/src/grpc/qdrant.rs
@@ -1556,7 +1556,7 @@ pub struct PayloadSchemaInfo {
     /// Field index parameters
     #[prost(message, optional, tag = "2")]
     pub params: ::core::option::Option<PayloadIndexParams>,
-    /// Number of points indexed within this field indexed
+    /// Number of points indexed within this field
     #[prost(uint64, optional, tag = "3")]
     pub points: ::core::option::Option<u64>,
 }
@@ -2436,7 +2436,7 @@ pub enum ReplicaState {
     /// Deprecated: snapshot shard transfer is in progress.
     /// Updates should not be sent to (and are ignored by) the shard.
     PartialSnapshot = 5,
-    /// Shard is undergoing recovered by an external node.
+    /// Shard is undergoing recovery by an external node.
     /// Normally rejects updates, accepts updates if force is true.
     Recovery = 6,
     /// Points are being migrated to this shard as part of scale-up resharding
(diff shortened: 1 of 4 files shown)


lib/api/src/grpc/qdrant.rs
timvisee, line 1: I'm confident all of this is fine because only comments have changed in the generated `qdrant.rs` file. 👍

timvisee (approved):

timvisee:
We agreed to merge this after the 1.16 release. That way we prevent this messing with proto diff's when updating clients. We can merge this after the release.

PR #7479, opened 2025-10-30
Avoid panics on empty metric families

JojiiOfficial:
Qdrant panics if we create an empty `MetricFamily`.
I've run into this a few times since in some cases it requires a manual check that's easy to miss.
One example is #7478.

This PR improves the metrics collection logic to prevent such panics in the future.
It should help avoid most similar issues in future.


generall:
@JojiiOfficial please resolve the conflict

generall (approved):

PR #7482, opened 2025-10-30
In metrics, only show cumulative number of page faults

timvisee:
In metrics, only expose a cumulative number of minor and major page faults. Remove the metrics for page faults in joined children:

```bash
# HELP qdrant_procfs_minor_page_faults_total count of minor page faults which didn't cause a 
# TYPE qdrant_procfs_minor_page_faults_total counter
qdrant_procfs_minor_page_faults_total 180664
# HELP qdrant_procfs_major_page_faults_total count of disk accesses caused by a mmap page 
# TYPE qdrant_procfs_major_page_faults_total counter
qdrant_procfs_major_page_faults_total 0
```

In my opinion, we should **only** expose the above two. I see no point in reporting separate values for children. In fact, it is currently incorrectly implemented, more on that later.

In practice we're only interested in the cumulative number of page faults that happen during the life time of the Qdrant process, in the process itself and in all its children. If the number grows faster than we expect, the cluster may need attention.

The minor/major metric needs to be a sum of:

1. page faults in the current PID
2. plus, page faults in all current child processes
3. plus, historic page faults from all joined child processes

~Actually, Qdrant currently does not even spawn any child processes at all. It means that the child metrics will always be 0 at this time.~ (Might be wrong of this, but the fact remains that we currently don't report it correctly)

But, to make our reporting more future proof I still recommend to implement it properly. For point two we'd have to iterate over all child processes recursively and fetch its page fault stats (left a TODO). Point three is a different number and we've already implemented this.

### All Submissions:

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -1016,6 +1016,9 @@ mod procfs_metrics {
 
     #[test]
     fn test_child_processes() {
+        use nix::sys::wait::waitpid;
+        use nix::unistd::{ForkResult, fork};
+
         let my_pid = procfs::process::Process::myself().unwrap().pid;
 
         // We have no children now
@@ -1023,22 +1026,51 @@ mod procfs_metrics {
 
         // Spawn two child processes
         let mut child_1 = std::process::Command::new("sleep")
-            .arg("10s")
+            .arg("3s")
             .spawn()
             .unwrap();
         let mut child_2 = std::process::Command::new("sleep")
-            .arg("10s")
+            .arg("3s")
             .spawn()
             .unwrap();
         let child_1_pid = child_1.id() as Pid;
         let child_2_pid = child_2.id() as Pid;
 
+        // Recursively fork process to create a grandchild
+        let fork_1 = match unsafe { fork().unwrap() } {
+            ForkResult::Parent { child } => child,
+            ForkResult::Child => {
+                let fork_2 = match unsafe { fork().unwrap() } {
+                    ForkResult::Parent { child } => child,
+                    ForkResult::Child => {
+                        std::thread::sleep(std::time::Duration::from_secs(3));
+                        std::process::exit(0);
+                    }
+                };
+                waitpid(fork_2, None).unwrap();
+                std::process::exit(0);
+            }
+        };
+
+        // Give forks a second to spawn
+        std::thread::sleep(std::time::Duration::from_secs(1));
+
+        // We expect exactly 4 children (3 direct, 1 grandchild)
+        // We don't know the PID of the grandchild here, so we expect three children and one extra
+        let child_pids = child_processes_helper(my_pid).unwrap();
+        assert_eq!(child_pids.len(), 4);
         assert_eq!(
-            child_processes_helper(my_pid).unwrap(),
-            AHashSet::from([child_1_pid, child_2_pid]),
+            child_pids
+                .difference(&AHashSet::from([child_1_pid, child_2_pid, fork_1.into()]))
+                .count(),
+            1,
+            "expect exactly one extra child",
         );
 
-        // Cleanup
+        // Cleanup fork
+        waitpid(fork_1, None).unwrap();
+
+        // Cleanup processes
         child_1.kill().unwrap();
         child_2.kill().unwrap();
         let _ = child_1.wait();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5159,6 +5159,7 @@ dependencies = [
  "memory",
  "mockito",
  "murmur3",
+ "nix 0.29.0",
  "ordered-float 5.1.0",
  "parking_lot",
  "procfs",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ sealed_test = "1.1.0"
 
 tempfile = { workspace = true }
 rusty-hook = "^0.11.2"
+nix = { workspace = true, features = ["process"] }
 
 mockito = "1.7"
 


src/common/metrics.rs
```diff
@@ -888,35 +888,23 @@ impl MetricsProvider for ProcFsMetrics {
             prefix,
         ));
 
+        // TODO: include counts for alive child processes
```
timvisee: This still needs to be implemented.

src/common/metrics.rs
```diff
@@ -806,138 +806,229 @@ fn label_pair(name: &str, value: &str) -> LabelPair {
...
+    /// Returns the minor and major page faults of all children including grandchildren, etc.
+    pub fn faults_for_all_children(parent: Pid) -> Result<(u64, u64), ProcError> {
+        let children = child_processes_helper(parent)?;
+        let mut min = 0;
+        let mut maj = 0;
+        for child in children {
+            let stat = match Process::new(child).and_then(|i| i.stat()) {
+                Ok(proc) => proc,
+                // Ignore children that don't exist anymore.
+                Err(ProcError::NotFound(_)) => continue,
```
timvisee, line 950: Great safety measure!

src/common/metrics.rs
```diff
@@ -806,138 +806,229 @@ fn label_pair(name: &str, value: &str) -> LabelPair {
...
-        metrics.push(metric_family(
-            "process_open_mmaps",
-            "count of open mmaps",
-            MetricType::GAUGE,
-            vec![gauge(self.mmap_count as f64, &[])],
-            prefix,
-        ));
+            min += stat.minflt;
+            maj += stat.majflt;
+        }
```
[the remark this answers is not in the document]
timvisee: Yes, we need to do this.

src/common/metrics.rs
```diff
@@ -806,138 +806,229 @@ fn label_pair(name: &str, value: &str) -> LabelPair {
...
-        };
-        metrics.push(metric_family(
-            "process_max_fds",
-            "limit for open file descriptors",
-            MetricType::GAUGE,
-            vec![gauge(fds_limit as f64, &[])],
-            prefix,
-        ));
+        child_processes_recursive(pid, &mut seen, &mut out)?;
+        Ok(out.into_iter().collect())
```
timvisee: We can probably use a single hashset here. But it's not super important.

timvisee: Done in https://github.com/qdrant/qdrant/pull/7482/commits/faadcc315d716226dadd7cab9f933cdca20cdbf2

We can use the returned boolean from [`HashSet::insert(...)`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert) to determine if we've seen a PID or not.

src/common/metrics.rs
```diff
@@ -806,138 +806,245 @@ fn label_pair(name: &str, value: &str) -> LabelPair {
...
+                MetricType::GAUGE,
+                vec![gauge(self.open_fds as f64, &[])],
+                prefix,
+            ));
+
+            let fds_limit = match (self.max_fds_soft, self.max_fds_hard) {
+                (0, hard) => hard,               // soft unlimited, use hard
+                (soft, 0) => soft,               // hard unlimited, use soft
+                (soft, hard) => min(soft, hard), // both limited, use minimum
+            };
```
xzfc, line 907: Why convert `LimitValue::Unlimited` to `0` only to `match` these zeroes later?

I suggest to remove `fn format_limit` and store `max_fds_soft`/`max_fds_hard` as `LimitValue`, not as `u64`.

src/common/metrics.rs
```diff
@@ -806,138 +806,245 @@ fn label_pair(name: &str, value: &str) -> LabelPair {
...
 
-        metrics.push(metric_family(
-            "process_open_fds",
-            "count of currently open file descriptors",
-            MetricType::GAUGE,
-            vec![gauge(self.open_fds as f64, &[])],
-            prefix,
-        ));
+        // Exclude parent Pid from results
+        pids.remove(&pid);
```
xzfc: Nit: I think it's not required as we don't add parent pid.

src/common/metrics.rs
```diff
@@ -806,138 +806,245 @@ fn label_pair(name: &str, value: &str) -> LabelPair {
...
-        metrics.push(metric_family(
-            "process_minor_page_faults_total",
-            "count of minor page faults which didn't cause a disk access",
-            MetricType::COUNTER,
-            vec![counter(self.minor_page_faults as f64, &[])],
-            prefix,
-        ));
+    /// Recursively collects all children of a process, specified by `pid`.
+    fn child_processes_recursive(pid: Pid, pids: &mut AHashSet<Pid>) -> Result<(), ProcError> {
+        for child_pid in child_processes(pid)? {
```
xzfc: Very minor performance nit: `child_processes` allocates a vector that would be dropped right after iterating. Tho, I think it's far from being a bottleneck.

src/common/metrics.rs
```diff
@@ -806,138 +806,245 @@ fn label_pair(name: &str, value: &str) -> LabelPair {
...
+            Ok(proc) => proc,
+            // Ignore children that don't exist anymore.
+            Err(ProcError::NotFound(_)) => return Ok(AHashSet::default()),
+            Err(err) => return Err(err),
+        };
+
+        process
+            .tasks()?
+            .filter_map(|task| {
+                let children = task.and_then(|task| task.children());
```
xzfc: TIL we can get a list of children PIDs directly via `/proc/*/task/*/children`.

The classic approach was to read `/proc/*/stat` for every process and rebuild the process tree by looking at parent PIDs.

The `/children` approach requires kernel with `CONFIG_PROC_CHILDREN`. I think it should be enabled practically everywhere on servers, but not sure.

xzfc (approved):

src/common/metrics.rs
timvisee: Replaced it with a debug assertion

PR #7484, opened 2025-10-31
Show active thread count in metrics

timvisee:
```bash
# HELP qdrant_process_threads count of active threads
# TYPE qdrant_process_threads gauge
qdrant_process_threads 126
```

### All Submissions:

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -809,6 +809,7 @@ fn label_pair(name: &str, value: &str) -> LabelPair {
 /// Structure for holding /procfs metrics, that can be easily populated in metrics API.
 #[cfg(target_os = "linux")]
 struct ProcFsMetrics {
+    thread_count: usize,
     mmap_count: usize,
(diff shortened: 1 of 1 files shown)


src/common/metrics.rs
```diff
@@ -827,6 +828,7 @@ impl ProcFsMetrics {
 
         let current_process = Process::myself()?;
 
+        let thread_count = current_process.tasks()?.flatten().count();
```
JojiiOfficial, line 832: I'm not sure if this is correct. What if a thread spawns another one?

JojiiOfficial: Looking into the docs of `Process::tasks`, it looks about right though:

> Iterate over all the Tasks (aka Threads) in this process

timvisee: Yeah this works fine for threads spawned in other threads.

Though this doesn't consider sub processes. But we don't create these yet, and we might want to expose those differently anyway.

PR #7486, opened 2025-10-31
Fix clippy 🙄

ffuugoo:
Rust 1.91 was released yesterday, and we got a bunch of new useless lints in clippy, yay!

Almost everything was fixed automatically by `clippy --fix`, except for the two lints that I've explicitly commented on below.

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -1,4 +1,3 @@
-use std::cmp::min;
 use std::collections::HashMap;
 
 use api::rest::models::HardwareUsage;
@@ -808,7 +807,7 @@ fn label_pair(name: &str, value: &str) -> LabelPair {
 }
 
 /// Structure for holding /procfs metrics, that can be easily populated in metrics API.
-#[cfg_attr(not(target_os = "linux"), expect(dead_code))]
+#[cfg(target_os = "linux")]
 struct ProcFsMetrics {
     mmap_count: usize,
     open_fds: usize,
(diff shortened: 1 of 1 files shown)


lib/segment/src/index/hnsw_index/graph_links/serializer.rs
```diff
@@ -99,6 +99,9 @@ pub fn serialize_graph_links<W: Write + Seek>(
     let mut offset = 0; // elements for Plain, bytes for Compressed/CompressedWithVectors
     let mut offsets = Vec::with_capacity(total_offsets_len as usize);
     offsets.push(0);
+
+    #[expect(clippy::needless_range_loop)]
+    // this clippy lint is positively demented, can't wait till they remove it 🙄
```
ffuugoo, line 104: The most braindead diagnostic that I've seen for a while. I think it will be reworked or removed soon, but maybe we should just disable it globally.

src/common/metrics.rs
```diff
@@ -808,6 +808,7 @@ fn label_pair(name: &str, value: &str) -> LabelPair {
 }
 
 /// Structure for holding /procfs metrics, that can be easily populated in metrics API.
+#[cfg_attr(not(target_os = "linux"), expect(dead_code))]
```
ffuugoo: Not sure what else to do here, so I only `dead_code` it on non-Linux

JojiiOfficial: We could make it `pub` but I think this doesn't make it much better.

timvisee: We probably need to feature flag it so the struct only exists on Linux builds.

timvisee: Done in https://github.com/qdrant/qdrant/pull/7486/commits/5ba317b3f630690c084404d443c71f8ea7b7db9f

PR #7491, opened 2025-11-01
Fusion & MMR support in Edge

generall:
`MergePlan` struct uses a lot of indirect state, which creates problems once we try to re-purpose it for Shard search.

For example:

- `rescore_params = None` was serving double implicit purpose of indicating either no prefeth case or fusion case. Those cases, actualually, should be handled differently.

- `Mmr` rescore function is tamperred with `Nearest` request, which make all underlying processes thing it is a regular nearest request. Thus, losing information about the original request

### DIFF
--- a/lib/edge/python/examples/common.py
+++ b/lib/edge/python/examples/common.py
@@ -4,9 +4,9 @@ import uuid
 
 from qdrant_edge import *
 
-print("---- Load shard ----")
 
 def load_new_shard():
+    print("---- Load shard ----")
     DATA_DIRECTORY = os.path.join(os.path.dirname(__file__), "data")
 
     # Clear and recreate data directory
--- a/lib/edge/src/query.rs
+++ b/lib/edge/src/query.rs
@@ -55,8 +55,14 @@ impl Shard {
             scored_points_batch.push(scored_points)
         }
 
-        debug_assert_eq!(scored_points_batch.len(), 1);
-        let scored_points = scored_points_batch.pop().unwrap_or_default();
+        let [scored_points] = scored_points_batch
+            .try_into()
+            .map_err(|unconverted: Vec<_>| {
+                OperationError::service_error(format!(
+                    "unexpected scored points batch size: expected 1, received {}",
+                    unconverted.len(),
+                ))
+            })?;
 
         Ok(scored_points)
     }
@@ -82,15 +88,20 @@ impl Shard {
             hw_measurement_acc.clone(),
         )?;
 
-        let result = self
+        let [result] = self
             .fill_with_payload_or_vectors(
                 vec![results],
                 with_payload,
                 with_vector,
                 hw_measurement_acc,
             )?
-            .pop()
-            .unwrap_or_default();
+            .try_into()
+            .map_err(|unconverted: Vec<_>| {
+                OperationError::service_error(format!(
+                    "expected single result after filling payload/vectors, got {}",
+                    unconverted.len(),
+                ))
+            })?;
         Ok(result)
     }
 
@@ -150,6 +161,7 @@ impl Shard {
             let collection_result = if let Some(rescore_params) = collection_level {
                 self.rescore(shard_stage_result, rescore_params, hw_counter_acc)?
             } else {
+                // Only one shard result is expected at this point.
                 shard_stage_result.into_iter().next().unwrap_or_default()
             };
 
@@ -160,7 +172,14 @@ impl Shard {
             // It should be a query without prefetches.
             debug_assert_eq!(depth, 0);
             debug_assert_eq!(sources.len(), 1);
-            Ok(sources.pop().unwrap_or_default())
+            let [result] = sources.try_into().map_err(|unconverted: Vec<_>| {
+                OperationError::service_error(format!(
+                    "expected single source without rescore stages, got {}",
+                    unconverted.len(),
+                ))
+            })?;
+
+            Ok(result)
         }
     }
 
--- a/lib/shard/src/query/planned_query.rs
+++ b/lib/shard/src/query/planned_query.rs
@@ -237,7 +237,7 @@ impl PlannedQuery {
         let sources =
             recurse_prefetches(&mut self.searches, &mut self.scrolls, prefetches, &filter)?;
 
-        let rescore_params = match rescoring_query {
+        let rescore_stages = match rescoring_query {
             ScoringQuery::Mmr(mmr) => {
                 let MmrInternal {
                     vector,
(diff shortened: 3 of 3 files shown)


lib/edge/python/examples/common.py
```diff
@@ -0,0 +1,51 @@
+import os
+import shutil
+import uuid
+
+from qdrant_edge import *
+
+print("---- Load shard ----")
+
+def load_new_shard():
```
ffuugoo: I think `print` should be part of `load_new_shard`?

```suggestion
def load_new_shard():
    print("---- Load shard ----")

```

lib/edge/src/query.rs
```diff
@@ -55,7 +55,10 @@ impl Shard {
             scored_points_batch.push(scored_points)
         }
 
-        Ok(scored_points_batch)
+        debug_assert_eq!(scored_points_batch.len(), 1);
+        let scored_points = scored_points_batch.pop().unwrap_or_default();
```
ffuugoo:
```suggestion
        let [scored_points] = scored_points_batch.try_into().map_err(|_| {
            OperationError::service_error(format!(
                "unexpected scored points batch size: expected 1, received {}",
                scored_points_batch.len(),
            )
        })?;
```

lib/edge/src/query.rs
```diff
@@ -79,7 +82,16 @@ impl Shard {
...
-        self.fill_with_payload_or_vectors(results, with_payload, with_vector, hw_measurement_acc)
+        let result = self
+            .fill_with_payload_or_vectors(
+                vec![results],
+                with_payload,
+                with_vector,
+                hw_measurement_acc,
+            )?
+            .pop()
+            .unwrap_or_default();
```
ffuugoo: Ditto.

lib/edge/src/query.rs
```diff
@@ -110,30 +122,45 @@ impl Shard {
...
+            let shard_stage_result = if let Some(rescore_params) = shard_level {
+                vec![self.rescore(sources, rescore_params, hw_counter_acc.clone())?]
+            } else {
+                sources
+            };
+
+            let collection_result = if let Some(rescore_params) = collection_level {
+                self.rescore(shard_stage_result, rescore_params, hw_counter_acc)?
+            } else {
+                shard_stage_result.into_iter().next().unwrap_or_default()
```
ffuugoo, line 165: Do we only expect length 0 or 1? Then I'd prefer some sort of conversion that enforces it. **Or** a comment, that we only use first result, even when there are more than one available.

lib/edge/src/query.rs
```diff
@@ -110,30 +122,45 @@ impl Shard {
...
+            // In Edge, both shard-level and collection-level rescoring are handled the same way.
+            Ok(collection_result)
         } else {
             // The sources here are passed to the next layer without any extra processing.
-            // It is either a query without prefetches, or a fusion request and the intermediate results are passed to the next layer.
+            // It should be a query without prefetches.
             debug_assert_eq!(depth, 0);
-            Ok(sources)
+            debug_assert_eq!(sources.len(), 1);
+            Ok(sources.pop().unwrap_or_default())
```
ffuugoo: Ditto.

lib/shard/src/query/planned_query.rs
```diff
@@ -197,40 +238,60 @@ impl PlannedQuery {
             recurse_prefetches(&mut self.searches, &mut self.scrolls, prefetches, &filter)?;
 
         let rescore_params = match rescoring_query {
```
ffuugoo:
```suggestion
        let rescore_stages = match rescoring_query {
```

lib/shard/src/query/planned_query.rs
```diff
@@ -197,40 +238,60 @@ impl PlannedQuery {
...
+                    score_threshold: score_threshold.map(OrderedFloat),
+                    params,
+                }))
+            }
         };
 
         let merge_plan = MergePlan {
             sources,
-            rescore_params,
+            rescore_stages: rescore_params,
```
ffuugoo:
```suggestion
            rescore_stages,
```

lib/shard/src/query/planned_query.rs
```diff
@@ -197,40 +238,60 @@ impl PlannedQuery {
...
+            ScoringQuery::Mmr(mmr) => {
+                let MmrInternal {
+                    vector,
+                    using,
+                    lambda: _,
+                    candidates_limit,
+                } = &mmr;
+
+                // Although MMR gets computed at collection level, we select top candidates via a nearest rescoring first
+                let shard_level_rescore_params = RescoreParams {
```
ffuugoo: Just call this `shard_level`? Hard to get confused here, IMO.

```suggestion
                let shard_level = RescoreParams {
```

lib/shard/src/query/planned_query.rs
```diff
@@ -197,40 +238,60 @@ impl PlannedQuery {
...
                     ))),
-                    limit: candidates_limit,
+                    limit: *candidates_limit,
                     score_threshold: score_threshold.map(OrderedFloat),
                     params,
                 };
-                // Although MMR gets computed at collection level, we select top candidates via a nearest rescoring first
-                Some(rescore_params)
+
+                let collection_level_rescore_params = RescoreParams {
```
ffuugoo: Same.


```suggestion
                let collection_level = RescoreParams {
```

lib/shard/src/query/planned_query.rs
```diff
@@ -197,40 +238,60 @@ impl PlannedQuery {
...
+                    rescore: ScoringQuery::Mmr(mmr),
+                    limit,
+                    score_threshold: score_threshold.map(OrderedFloat),
+                    params,
+                };
+
+                Some(RescoreStages {
+                    shard_level: Some(shard_level_rescore_params),
+                    collection_level: Some(collection_level_rescore_params),
```
ffuugoo:
```suggestion
                    shard_level: Some(shard_level),
                    collection_level: Some(collection_level),
```

lib/shard/src/query/planned_query.rs
```diff
@@ -288,14 +349,19 @@ fn recurse_prefetches(
                 )
             })?;
 
+            // Even if this is a fusion request, it can only be executed at shard level here,
+            // because we can't forward the inner results to the collection level without
+            // materializing them first.
+            let rescore_stage = RescoreStages::shard_level(RescoreParams {
```
ffuugoo:
```suggestion
            let rescore_stages = RescoreStages::shard_level(RescoreParams {
```

lib/shard/src/query/planned_query.rs
```diff
@@ -288,14 +349,19 @@ fn recurse_prefetches(
...
+
             let merge_plan = MergePlan {
                 sources: inner_sources,
-                rescore_params: Some(RescoreParams {
-                    rescore,
-                    limit,
-                    score_threshold,
-                    params,
-                }),
+                rescore_stages: Some(rescore_stage),
```
ffuugoo:
```suggestion
                rescore_stages: Some(rescore_stages),
```

Issue #7494, opened 2025-11-04
Feature Request: Truncate API


myyrakle:
**Is your feature request related to a problem? Please describe.**
When implementing and testing features, it's often necessary to delete and recreate data. However, qdrant doesn't have this feature, so the quickest way is to delete the table and then recreate it. 
However, this is quite cumbersome. It would be nice to have a feature that could solve this problem.

**Describe the solution you'd like**
Add a "truncate" API to remove all data from a specific collection.

**Describe alternatives you've considered**
none

**Additional context**
none

generall:
Hey! Thanks for the proposal!

Have you considered `delete-by-filter` API with empty filter?
It is not really recommended to use in production, as record-by-record deletion might be slow on large scale, but for experimenting it should do the work.

myyrakle:
Oh! I didn't know there was an empty filter.
Sure, it might not be great for performance, but it might be useful for light use cases.

However, in my case, I often experiment with and delete millions of data units, so it would be nice to have an optimized function.

I understand if you think this is a trivial feature. If you disagree with its necessity, I'll close this issue.

timvisee:
I don't think we can implement something much more efficient than delete by filter. Note that it's already quite optimized when used with a payload index.

Another possibility may be to use custom sharding. You'd put points in a specific named shard, and you can drop such shard with a single API call. It will be quite similar to dropping the whole collection, but allows you to use some partitioning.
Docs: https://qdrant.tech/documentation/guides/distributed_deployment/#user-defined-sharding

Since I don't think we have a more optimized 'large delete by filter' or 'truncate' API call planned I'm closing this.

Thanks for the suggestion!

timvisee:
After closing I now notice you opened <https://github.com/qdrant/qdrant/pull/7496> in an attempt to tackle it. I'll still make sure to cover it and will share my opinion on it there.

PR #7496, opened 2025-11-04
Add Truncate API

myyrakle:
This is a feature proposal. related issue: #7494. 

I'm not sure if it aligns with the project's direction.
Documentation and test code additions are not yet included.
(Please ignore the changes to Cargo.toml. These were temporary changes made because rocksdb wouldn't compile.)

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -787,20 +787,18 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.69.4"
+version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a00dc851838a2120612785d195287475a3ac45514741da670b735818822129a0"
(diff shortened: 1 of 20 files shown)


timvisee:
Thank you for the submission!

I now get what you're trying to achieve. You do want to delete **all** data efficiently, by wiping segments. I missed that at first, sorry about that.

I like the idea of dropping segments, as that's definitely much faster than marking all points as delete. There are some additional things that must be considered though. For example, we always need to leave at least one mutable segment in place. And we must abort any ongoing optimizations. We're also making some internal changes in segment holder handling, which does affect how this should be implemented.

Also, I'd prefer to use the existing delete by filter API for this. If an empty filter is provided (which effectively means delete everything), we can use this separate branch to drop all segments. That way we don't to introduce a new API call.

Since I believe this change requires quite a bit of nuances, I'll take over the task as described above a bit later. Hope you don't mind. :smiley:

PR #7497, opened 2025-11-05
Metrics for snapshots

JojiiOfficial:
Depends on #7482 

Adds the following new snapshot metrics:

```bash
# HELP snapshot_creation_running amount of snapshot creations that are currently running
# TYPE snapshot_creation_running gauge
snapshot_creation_running{id="benchmark"} 1

# HELP snapshot_recovery_running amount of snapshot recovery operations currently running
# TYPE snapshot_recovery_running gauge
snapshot_recovery_running{id="benchmark"} 0

# HELP snapshot_created_total total amount of snapshots created
# TYPE snapshot_created_total counter
snapshot_created_total{id="benchmark"} 1
```

### Design decision
There are already a few shard level information about shard snapshots in the telemetry API. 
However they are only enabled in certain scenarios (if a snapshot manifest is provided) and only for streaming snapshots. 
We had to manually add new metrics anyways (like `snapshot_creations_total`) in order to expose all metrics discussed, so I decided to implement these consistently uncoupled from the existing telemetry data. This also ensures that these changes don't introduce any unexpected behavior or mis-measurements.

As always, I'm open for any suggestions on these decisions and happy about any feedback!

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -190,9 +190,9 @@ impl MetricsProvider for CollectionsTelemetry {
         let mut total_dead_replicas = 0;
 
         // Snapshot metrics
-        let mut snapshots_currently_running = vec![];
-        let mut snapshot_recovery_running = vec![];
-        let mut snapshots_total = vec![];
+        let mut snapshots_creation_running = vec![];
+        let mut snapshots_recovery_running = vec![];
+        let mut snapshots_created_total = vec![];
 
         let mut vector_count_by_name = vec![];
 
@@ -307,18 +307,18 @@ impl MetricsProvider for CollectionsTelemetry {
         for snapshot_telemetry in self.snapshots.iter().flatten() {
             let id = &snapshot_telemetry.id;
 
-            snapshot_recovery_running.push(gauge(
+            snapshots_recovery_running.push(gauge(
                 snapshot_telemetry
                     .running_snapshot_recovery
                     .unwrap_or_default() as f64,
                 &[("id", id)],
             ));
-            snapshots_currently_running.push(gauge(
+            snapshots_creation_running.push(gauge(
                 snapshot_telemetry.running_snapshots.unwrap_or_default() as f64,
                 &[("id", id)],
             ));
 
-            snapshots_total.push(counter(
+            snapshots_created_total.push(counter(
                 snapshot_telemetry
                     .total_snapshot_creations
                     .unwrap_or_default() as f64,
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection/telemetry.rs
```diff
@@ -0,0 +1,41 @@
+use common::types::{DetailsLevel, TelemetryDetail};
```
JojiiOfficial, line 1: This small refactor is not necessary anymore, but I decided to keep it anyways since it declusters the mod.rs file.
If you prefer it to be reverted, let me know!

lib/collection/src/shards/replica_set/partial_snapshot_meta.rs
```diff
@@ -88,44 +92,10 @@ impl PartialSnapshotMeta {
...
+        self.recovery_timestamp.store(timestamp, Ordering::Relaxed);
     }
 }
 
 fn recovery_in_progress() -> CollectionError {
     CollectionError::shard_unavailable("partial snapshot recovery is in progress")
 }
-
-#[derive(Debug, Default)]
-pub struct RequestTracker {
```
JojiiOfficial, line 101: I noticed a bit too late that there was already some sort of `ScopeTracker`. I decided to keep my implementation because it is located at a better suited place, has tests and is implemented a bit more generic.

lib/storage/src/content_manager/toc/telemetry.rs
```diff
@@ -1,23 +1,78 @@
...
 use common::types::TelemetryDetail;
+use dashmap::DashMap;
 
 use crate::content_manager::toc::TableOfContent;
 use crate::rbac::Access;
 
+/// Collects various telemetry handled by ToC.
+#[derive(Default)]
+pub(super) struct TocTelemetryCollector {
+    snapshots: DashMap<String, SnapshotTelemetryCollector>,
```
JojiiOfficial, line 17: I decided to let ToC handle these for 2 reasons: 
-  Creating new collections from snapshots can't be fully measured inside `Collection` because this type
   only exists for the second half of the recovery process (archive extracting, collection creation are skipped)
- To keep snapshot metrics consistent and in one place. (An alternative way would be to only measure snapshot recovery here or in a similar 'global' location).

Additionally we can extend this later for other metrics that must be handled 'globally'.

If you don't agree with 1. and measuring tar extracting, or collection creation isn't important, I can move this to another place like `Collection`!

src/common/snapshots.rs
```diff
@@ -34,6 +34,11 @@ pub async fn create_shard_snapshot(
         .check_collection_access(&collection_name, AccessRequirements::new().write().extras())?;
     let collection = toc.get_collection(&collection_pass).await?;
 
+    let _telemetry_scope_guard = toc
+        .snapshot_telemetry_collector(&collection_name)
+        .running_snapshots
+        .measure_scope();
```
timvisee, line 40: When we hit this, it might still take some time for the snapshot to actually be created due to locks deeper down. But I think it's fine to still count it from here on.

lib/storage/src/content_manager/toc/telemetry.rs
```diff
@@ -37,4 +92,38 @@ impl TableOfContent {
...
+        // Increment current running counter.
+        let running_snapshots_guard = self
+            .snapshot_telemetry_collector(collection_name)
+            .running_snapshots
+            .measure_scope();
+
+        // Increment total counter.
+        self.snapshot_telemetry_collector(collection_name)
+            .snapshots_total
+            .fetch_add(1, Ordering::Relaxed);
```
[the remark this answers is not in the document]
timvisee: I don't see this as significant problem

Issue #3321, opened 2024-01-04
Allow Qdrant to run on read-only FS


generall:
**Is your feature request related to a problem? Please describe.**

Currently qdrant requires write access to storage directory, even if there are no write operations happening in the system.
It would be interesting in some scenarios to have an ability to run qdrant from read-only FS. For example, from the mounted network FS, so that multiple qdrant instances can serve same data.

**Describe the solution you'd like**

Introduce a new CLI argument `--read-only` which will enable read only mode in qdrant.

For update APIs read-only instance should behave in the same way as for read-only API key - respond with `403 Forbidden` and explanation that the instance is in read-only mode.

Distributed deployment can't be read-only.

**Describe alternatives you've considered**

Automatic fallback into read-only if the file system doesn't allow writes, but it seems less flexible and error-prone.

**Additional context**

read-only mode assumes skipping WAL reading, so all changes should be applied and flushed to the real storage in advance

---

Note for contributors: Please consider this as tracking issue. If you think that it would be beneficial to split the task into multiple smaller PRs, please you are welcome to do so. Bounty will be rewarded for each PR independently

pqcfox:
Would love to help with this PR if possible?

rajdip-b:
Hey @pqcfox feel free to get started with this one. I might not be able to fi ish it at this point.

pqcfox:
Thanks! I'll add an attempt here and get started tomorrow.

/attempt #3321

pqcfox:
Just sat down and planned out how this could go, my current plan to start with `src/actix` is to:

- Add the `--read-only` flag, and store its value in the TOC
- Move `is_read_only` and `READ_ONLY_POST_PATTERNS` out of `api_key.rs` to somewhere common in `src/actix`
- Create a new file in `src/actix` with a `ReadOnlyMiddleware` struct with just a boolean member, whose `Service` impl uses a call to `is_read_only` and on failure sends out a 

@generall would this approach be acceptable to you for this PR? I'm hesitant to move something from `api_key.rs` just because the PR for that was so recent, but I figured having the API read only and FS read only mechanisms in the same place and both implemented as Actix middleware made the most sense.

Thanks!

generall:
> would this approach be acceptable to you for this PR?

Looks ok, but we should also take gRPC into account

pqcfox:
Sounds great, thanks so much--I'll kick off the PR today :)

pqcfox:
[...]

PR #7498, opened 2025-11-07
feat: add read-only mode and integration tests for it

zelosleone:
### All Submissions:

* [X] Contributions should target the `main` branch. Did you create your branch from `main`? (targeted main, not dev.)

### New Feature Submissions:

### Changes to Core Features:

/claim #3321

- no WAL-reading, we directly using write-mode flush to local storage for read-only mode to read only.
- made WAL option Type for us to skip it with none on read-only mode, safer option although it required some refactoring to main codebase and tests.
- there were clippy and fmt errors way before my change, so ran them too. didnt fix all the clippy errors before my change but cant do anything about cargo format.
- updated the existing integration test setup with new parameters and added a helper for us to load for read-only mode.
- added checks for read-only mode to fail fast if directory doesn't exist (no data), if filesystem doesnt support it, in cluster mode as well.
- added new 4 integration tests for read-only mode
- manually tested it as well via podman instance + locally on ubuntu.


lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -246,7 +246,13 @@ impl LocalShard {
             collection_config,
             shared_storage_config,
             payload_index_schema,
-            wal: RecoverableWal::new(locked_wal, clocks.newest_clocks, clocks.oldest_clocks),
+            wal: locked_wal.map(|w| {
+                RecoverableWal::new(
+                    w,
+                    clocks.newest_clocks.clone(),
```
generall: clocks can be deconstructed to avoid clone:

```
        let LocalShardClocks {
            newest_clocks,
            oldest_clocks,
        } = clocks;
```

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -572,8 +586,10 @@ impl LocalShard {
             segment_holder.add_new(segment);
         }
 
-        let wal: SerdeWal<OperationWithClockTag> =
-            SerdeWal::new(wal_path.to_str().unwrap(), (&config.wal_config).into())?;
+        let wal: Option<SerdeWal<OperationWithClockTag>> = Some(SerdeWal::new(
+            wal_path.to_str().unwrap(),
+            (&config.wal_config).into(),
```
generall: Please avoid `.into`, prefer `Type::from` instead

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -72,19 +72,25 @@ impl ShardOperation for LocalShard {
...
+                            status: UpdateStatus::ClockRejected,
+                            clock_tag: operation.clock_tag,
+                        });
+                    }
 
-                Err(err) => return Err(err.into()),
+                    Err(err) => return Err(err.into()),
+                }
+            } else {
+                return Err(CollectionError::service_error(
```
generall: maybe bad_request instead of service_error? Service error usually means that something unexpected happens on server-side.

lib/collection/src/shards/local_shard/snapshot.rs
```diff
@@ -108,9 +108,15 @@ impl LocalShard {
 
             if save_wal {
                 // snapshot all shard's WAL
-                Self::snapshot_wal(wal, &tar_c)
-            } else {
+                if let Some(wal) = wal {
+                    Self::snapshot_wal(wal, &tar_c)
+                } else {
+                    Ok(())
```
generall: I don't think just skipping WAL during the snapshot, even under read-only mode is a correct approach. The obtained snapshot will not be readable with regular deployment anymore. I think we should always copy existing WAL files, even though the WAL itself is not loaded. 

Also need to make sure `temp_path` is writable - if it is part of `./storage` - it is not

generall (changes requested):
Hey @zelosleone, thanks for the contribution, I think it is on the right track!

There are a few comments from my side above, and also could you please take a look at clippy warnings/errors, so that CI is green?

Apart from that, I noticed that your PR introduces a lot of places where WAL is checked for presence. Have you considered making WAL object itself read-only, so that it acks like an empty WAL you can't write to? I am not sure, but it seems like it could require less boilerplate checks and will resolve the edge-case in snapshot creation

lib/collection/src/wal_delta.rs
```diff
@@ -1979,4 +2115,129 @@ mod tests {
...
+        let (read_only_wal, _dir) = fixture_read_only_wal();
+        let wal_mode = WalMode::ReadOnly(read_only_wal);
+        let mut operation =
+            OperationWithClockTag::new(mock_operation(1), Some(ClockTag::new(1, 0, 0)));
+        let result = wal_mode.lock_and_write(&mut operation).await;
+        assert!(result.is_err());
+        assert!(matches!(
+            result.unwrap_err(),
+            shard::wal::WalError::InitWalError(_)
+        ));
```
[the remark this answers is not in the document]
zelosleone: quite interesting it picked this up btw, but already solved with the second commit. nice though

zelosleone:
Hey @generall thanks for the review and comments! I went with your suggestions. I also went ahead and fixed all the pre-existing clippy errors, which spawned over a lot of files btw. Now WAL handling in snapshots is:

- When WAL is available and `save_wal=true`: copies actual WAL files
- When WAL is available and `save_wal=false`: creates empty WAL compatible with data
- When in read-only mode and `save_wal=false`: creates empty WAL from disk WAL files
- When in read-only mode and `save_wal=true`: copies existing WAL files from disk

So now snapshots are always readable and ready. Plus, i made the WAL-object read-only with a polymorphic enum, this should be better and solves the edge case you mentioned.

generall:
I was trying to resurrect this PR, but there too many conflicts =(
Also, it should be pointing `dev` instead of `master`

zelosleone:
> I was trying to resurrect this PR, but there too many conflicts =( Also, it should be pointing `dev` instead of `master`

I will get it done! I assume everything besides the merge conflicts and target branch are alright right?

BOPOHA:
When do you think we could get this released?

zelosleone:
@generall all done!

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -109,14 +109,14 @@ impl UpdateWorkers {
         operation: CollectionUpdateOperations,
         op_num: SeqNumberType,
         wait: bool,
-        wal: LockedWal,
+        wal: Option<LockedWal>,
```
generall: I would prefer to not even create update workers if wal is empty, instead of checking it internally

generall:
When I am trying to load with read-only fs, this happens:

```
2025-12-13T21:28:35.135137Z ERROR qdrant::startup: Panic backtrace: 
   0: qdrant::startup::setup_panic_hook::{{closure}}
   1: std::panicking::panic_with_hook
   2: std::panicking::panic_handler::{{closure}}
   3: std::sys::backtrace::__rust_end_short_backtrace
   4: __rustc::rust_begin_unwind
   5: core::panicking::panic_fmt
   6: collection::shards::shard_holder::ShardHolder::load_shards::{{closure}}
   7: tokio::runtime::park::CachedParkThread::block_on
   8: tokio::runtime::context::runtime::enter_runtime
   9: tokio::runtime::runtime::Runtime::block_on
  10: storage::content_manager::toc::TableOfContent::new
  11: qdrant::main
  12: std::sys::backtrace::__rust_begin_short_backtrace
  13: std::rt::lang_start::{{closure}}
  14: std::rt::lang_start_internal
  15: main
  16: <unknown>
  17: __libc_start_main
  18: _start

2025-12-13T21:28:35.135156Z ERROR qdrant::startup: Panic occurred in file /qdrant/lib/collection/src/shards/replica_set/mod.rs at line 319: Failed to load local shard "./storage/collections/test/0": Service internal error: Failed to open mmap payload storage: failed to open file `./storage/collections/test/0/segments/4ea958d8-0b64-4312-9a53-0cd857e93535/payload_storage/tracker.dat`: Read-only file system (os error 30)
```

zelosleone:
@generall Sorry for the delay! There was christmas + new year + changing from main to dev introduced a set of new things i needed to add! All done though, including the mmap issue! Also tested with every single endpoint i can find (GET and POST) while running with read-only mode. All works!

zelosleone:
Since original issue was determined as not planned, i am closing this.

PR #7499, opened 2025-11-07
Metrics for LocalShard updates

JojiiOfficial:
Depends on #7497 

Adds the following new metric to the metrics API:

```
# HELP collection_update_queue_size update queue of a collection separated by points, vectors, index, payloads and other (such as control signals)
# TYPE collection_update_queue_size gauge
collection_update_queue_size{id="test",update_type="points"} 0
collection_update_queue_size{id="test",update_type="vectors"} 0
collection_update_queue_size{id="test",update_type="index"} 0
collection_update_queue_size{id="test",update_type="payloads"} 0
collection_update_queue_size{id="test",update_type="other"} 0

# HELP collection_update_queue_total currently running and enqueued shard updates
# TYPE collection_update_queue_total gauge
collection_update_queue_total{id="test"} 0
```

These measure the amount of currently running update operations on local shards, grouped by collections.


lib/collection/src/shards/update_tracker.rs
```diff
@@ -1,12 +1,13 @@
...
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::Ordering;
 
+use common::scope_tracker::{ScopeTracker, ScopeTrackerGuard};
 use tokio::sync::watch;
 
 #[derive(Clone, Debug)]
 pub struct UpdateTracker {
-    update_operations: Arc<AtomicUsize>,
+    update_operations: ScopeTracker,
```
JojiiOfficial, line 10: We can reuse the new `ScopeTracker` here too

src/common/metrics.rs
```diff
@@ -423,6 +430,16 @@ impl MetricsProvider for CollectionsTelemetry {
                 prefix,
             ));
         }
+
+        if !collection_updates.is_empty() {
+            metrics.push(metric_family(
+                "collection_running_updates",
```
```suggestion
                "collection_updates_running",
```

JojiiOfficial:
> I actually think we only need the latter. If we have any thing in the queue, it implies an update is currently being applied.

This would require that a running update operation is still considered as enqueued.
In my opinion this could lead to confusion because items in queues are usually waiting to get scheduled, and get removed from the queue when about to schedule. However if we are fine with the queue also containing currently running operation(s), I totally agree!

timvisee:
> > I actually think we only need the latter. If we have any thing in the queue, it implies an update is currently being applied.
> 
> This would require that a running update operation is still considered as enqueued. In my opinion this could lead to confusion because items in queues are usually waiting to get scheduled, and get removed from the queue when about to schedule. However if we are fine with the queue also containing currently running operation(s), I totally agree!

Maybe we can include the queue and call it `collection_update_queue_total` (which sums both applying updates + queue size).

Then if we ever want to report the number of actual updates being applied we can add an extra metrics.

lib/collection/src/common/update_queue.rs
```diff
@@ -0,0 +1,238 @@
...
+        let permit = self.sender.reserve().await?;
+        self.increment_counter(update_type);
+        let counter = self.counter.get_counter(update_type);
+        Ok(UpdateQueuePermit::new(permit, counter))
+    }
+
+    pub fn try_send(
+        &self,
+        value: T,
+        update_type: Option<UpdateType>,
```
timvisee, line 119: Not a huge fan of adding this parameter.

Can't we drive it from `T` here?

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -239,7 +240,7 @@ impl LocalShard {
         );
 
         let (update_sender, update_receiver) =
-            mpsc::channel(shared_storage_config.update_queue_size);
+            new_update_queue_channel(shared_storage_config.update_queue_size);
```
timvisee, line 243: I'd strongly prefer to keep the Tokio queue. But I guess that'll make it hard to track the operation type, right? Would it be easier if we remove operation types?

lib/collection/src/common/update_queue.rs
JojiiOfficial: Nope, it's always the same type (theoretically we could remove the generic type `T`).
Different operation types are encoded in enums, not types: https://github.com/qdrant/qdrant/blob/master/lib/collection/src/shards/local_shard/shard_ops.rs#L91

lib/collection/src/shards/local_shard/mod.rs
JojiiOfficial: I agree.
We probably can workaround this channel wrapper, even with counters for separate operation types. 
However, we'd need to have to manually count each place we use sending or receiving, and due to it's current implementation, passing the counters around would become quite ugly.
For consistency and cleaner code, I went with the wrapping approach but I'm fine to adjust or drop it if it's complexity doesn't justify it's benefits.

PR #7500, opened 2025-11-07
Limit the timeout parameter for all requests to 1h

JojiiOfficial:
- Limit timeout to 1h for both rest and grpc endpoints.
- use `Duration` instead of `u64` in strict mode for better consistency and cleaner code

### DIFF
--- a/src/actix/api/cluster_api.rs
+++ b/src/actix/api/cluster_api.rs
@@ -11,16 +11,40 @@ use storage::dispatcher::Dispatcher;
 use storage::rbac::AccessRequirements;
 use validator::Validate;
 
+use crate::actix::api::cluster_api::request_params::QueryParams;
 use crate::actix::auth::ActixAccess;
 use crate::actix::helpers;
 
-#[derive(Debug, Deserialize, Validate)]
-struct QueryParams {
-    #[serde(default)]
-    force: bool,
-    #[serde(default)]
-    #[validate(range(min = 1))]
-    timeout: Option<u64>,
+mod request_params {
+    use std::cmp;
+    use std::time::Duration;
+
+    use serde::Deserialize;
+    use validator::Validate;
+
+    use crate::actix::api::read_params::HOUR_IN_SECONDS;
+
+    #[derive(Debug, Deserialize, Validate)]
+    pub(super) struct QueryParams {
+        #[serde(default)]
+        force: bool,
+        #[serde(default)]
+        #[validate(range(min = 1))]
+        timeout: Option<u64>,
+    }
+
+    impl QueryParams {
+        /// Returns the passed timeout, limited to 1h.
+        pub fn timeout(&self) -> Option<Duration> {
+            self.timeout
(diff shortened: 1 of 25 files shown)


src/actix/api/cluster_api.rs
```diff
@@ -11,16 +11,40 @@ use storage::dispatcher::Dispatcher;
...
 use crate::actix::helpers;
 
-#[derive(Debug, Deserialize, Validate)]
-struct QueryParams {
-    #[serde(default)]
-    force: bool,
-    #[serde(default)]
-    #[validate(range(min = 1))]
-    timeout: Option<u64>,
+mod request_params {
```
JojiiOfficial, line 18: Moved `QueryParams` to a separate module to prevent direct access to the `timeout` attribute here, and in some other places.
This helps ensuring the timeout limit is properly applied.

agourlay:
Why secretly limiting the timeout to 1h instead of adding a max value for it on the API?

JojiiOfficial:
The reasons I went this way

- Backward compatibility (Throwing errors for values that were allowed before)
- Timeout of > 1h does not cause Qdrant to reach an illegal state that must be prevented
- The description for this feature sounded like a value limitation, rather than a hard API constraint (I could've understood it wrong though)

On a second thought though, I agree that being completely silent here might be confusing as well. Logging a warning could be a good solution in my opinion but I'm open to suggestions.

Any thoughts on this?

timvisee:
How about setting no upper bound? Or is there a clear case you'd like to solve with this?

I do like the use of `Duration` though :+1:

agourlay:
Thanks for the answer, I definitely sympathize with the goal of not having super long running computations :+1:  

> Backward compatibility (Throwing errors for values that were allowed before)

Do we have any data regarding the usage of timeout values higher than 1h?
It seems a bit arbitrary to cut at 1h.

> Timeout of > 1h does not cause Qdrant to reach an illegal state that must be prevented

I think a validation is fine if we do not want to support a given parameter value.
It easier to let users know via a client side error I believe.

> The description for this feature sounded like a value limitation, rather than a hard API constraint (I could've understood it wrong though)

I think I am a bit puzzled by the approach because we already have a `max_timeout` knob in strict mode for a max allowed timeout value.

I would prefer having default fallback timeout everywhere and let strict mode care about the limits of accepted values.

timvisee:
> Do we have any data regarding the usage of timeout values higher than 1h?

Some people has been using patterns like `?timeout=99999` just to allow _something large_.

> Thanks for the answer, I definitely sympathize with the goal of not having super long running computations 👍

I think cancelling a request internally when the user request is cancelled is a much better way to handle this. Then the user still has control after submitting the operation, and can still cancel extremely expensive operations.

One exception is wait time on locks. But I don't consider that a problem here. Either locks are obtained relatively quickly and then the operation still aborts. Or we hold a lock for a ling time (which is cheap!) and we abort the operation after.

In that case I'd rather have no upper timeout bound.

@agourlay is working on correct handling of request cancellation.

JojiiOfficial:
Closing as discussed

PR #7504, opened 2025-11-09
Don't return failure when writing to `Partial` replicas for tenant promotion

KShivendu:
If you have collection with something like this:

```json
{
    "result": {
        "peer_id": 283000563180723,
        "shard_count": 2,
        "local_shards": [
            {
                "shard_id": 1,
                "shard_key": "default",
                "points_count": 1000,
                "state": "Active"
            },
            {
                "shard_id": 2,
                "shard_key": "tenant1",
                "points_count": 0,
                "state": "Partial"
            }
        ],
        "remote_shards": [],
        "shard_transfers": []
    },
    "status": "ok",
    "time": 0.000541685
}
```

Using shard fallback for writes will return error response because it internally tries to write to both shard keys. This happens because `tenant1` has only `Partial` replicas (i.e. no `Active` replicas)

```http
PUT /collections/benchmark/points
{
  "points": [
    {
      "id": 1,
      "payload": {"org": "tenant1"},
      "vector": [0.1, 0.2, 0.3, 0.4]
    }
  ],
  "shard_key": {
    "target": "tenant1",
    "fallback": "default"
  }
}
```

```console
{'status': {'error': 'Service internal error: Failed to apply operation to at least one `Active` replica. Consistency of this update is not guaranteed. Please retry. '}, 'time': 0.191400375, 'usage': {'hardware': {'cpu': 0, 'payload_io_read': 0, 'payload_io_write': 0, 'payload_index_io_read': 0, 'payload_index_io_write': 0, 'vector_io_read': 0, 'vector_io_write': 0}, 'inference': None}}
```

The write succeeds anyways in this case but we return an error which can be confusing for the user.

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -1034,18 +1034,17 @@ impl ShardReplicaSet {
         is_active_or_resharding && !self.is_locally_disabled(peer_id)
     }
 
-    fn are_all_partial(&self) -> bool {
-        if self.replica_state.read().peers().is_empty() {
+    /// Is the current replica set used for tenant promotion
+    fn is_tenant_promotion(&self) -> bool {
+        // If there's only on peer and it's in Partial state, it's a promotion
+        // since we don't support promotion with multiple destination replicas
+        if self.shard_key.is_none() {
+            // No custom sharding means no promotion
             return false;
         }
 
-        // todo: Check if we have custom sharding mode enabled. It's a must
-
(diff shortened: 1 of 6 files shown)


generall:
closing in favor of https://github.com/qdrant/qdrant/pull/7474

timvisee:
> closing in favor of #7474

The key change is that we now don't send updates to the new shard in partial state from nodes directly.

Instead, we only forward updates through a proxy. Sending updates this way prevents the 'not enough active' problem.

PR #7507, opened 2025-11-10
Fix missing workspace lints

agourlay:
Fix missing workspace lints declaration in `edge/*` and `shard`.


ffuugoo:
Thanks for fixing our mess. 😌🙏

Issue #7336, opened 2025-09-30
Flaky test_consensus_compaction


agourlay:
```
=================================== FAILURES ===================================
__________________________ test_consensus_compaction ___________________________

tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-1/test_consensus_compaction0')

    def test_consensus_compaction(tmp_path: pathlib.Path):
        """
        Basic consensus compaction test.
        Does a few consensus operations on a cluster. Then adds a new peer and
        asserts it boots with correct consensus state.
        """
    
        assert_project_root()
    
        env = {
            # Aggressively compact consensus WAL
            "QDRANT__CLUSTER__CONSENSUS__COMPACT_WAL_ENTRIES": "1",
        }
    
        # Start cluster
        peer_api_uris, peer_dirs, bootstrap_uri = start_cluster(tmp_path, N_PEERS, port_seed=10000, extra_env=env)
    
        create_collection(peer_api_uris[0], shard_number=N_SHARDS, replication_factor=N_REPLICA)
        wait_collection_exists_and_active_on_all_peers(collection_name="test_collection", peer_api_uris=peer_api_uris)
    
        # Add cluster metadata
        put_metadata_key(peer_api_uris, 'my_metadata', 'My value!')
    
        # Repeatedly drop, re-create collection to accumulate Raft log entries
        for i in range(0, 5):
            print(f"recreating collection {i}")
            # Drop test_collection
            drop_collection(peer_api_uris[0], "test_collection", timeout=5)
            # Re-create test_collection
            create_collection(peer_api_uris[0], shard_number=N_SHARDS, replication_factor=N_REPLICA, timeout=3)
            res = requests.get(f"{peer_api_uris[1]}/collections")
            print(res.json())
    
        # Wait for collection to be ready on all peers
        for peer_uri in peer_api_uris:
            wait_for_all_replicas_active(peer_uri, "test_collection")
    
        # Add extra node
        # Due to aggressive consensus WAL compaction, the peer has to join by consensus snapshot
        peer_dirs.append(make_peer_folder(tmp_path, N_PEERS))
        new_url = start_peer(peer_dirs[-1], "peer_3_extra.log", bootstrap_uri, port=21000, extra_env=env)
        peer_api_uris.append(new_url)
        wait_all_peers_up([new_url])
    
        # Ensure cluster metadata is consistent on all peers
        # Failed before <https://github.com/qdrant/qdrant/pull/6014>
[...]

PR #7513, opened 2025-11-11
Fix flaky test consensus compaction

agourlay:
Another try at fixing https://github.com/qdrant/qdrant/issues/7336 because it broke some of my builds.

Looking at the logs, I can clearly see that the consensus snapshot is applied after we request the metadata key for the new node.

Here is a lazy screenshot of the unified logs.

term

I was never able to reproduce locally, but I assume my fix is legit :ok_hand:

### DIFF
--- a/tests/consensus_tests/test_consensus_compaction.py
+++ b/tests/consensus_tests/test_consensus_compaction.py
@@ -56,11 +56,14 @@ def test_consensus_compaction(tmp_path: pathlib.Path):
     peer_dirs.append(make_peer_folder(tmp_path, N_PEERS))
     new_url = start_peer(peer_dirs[-1], "peer_3_extra.log", bootstrap_uri, port=21000, extra_env=env)
     peer_api_uris.append(new_url)
+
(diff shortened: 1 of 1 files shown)


tests/consensus_tests/test_consensus_compaction.py
```diff
@@ -56,11 +56,14 @@ def test_consensus_compaction(tmp_path: pathlib.Path):
     peer_dirs.append(make_peer_folder(tmp_path, N_PEERS))
     new_url = start_peer(peer_dirs[-1], "peer_3_extra.log", bootstrap_uri, port=21000, extra_env=env)
     peer_api_uris.append(new_url)
+
+    # Wait for collection to be ready on the new peer
     wait_all_peers_up([new_url])
+    wait_collection_exists_and_active_on_all_peers(collection_name="test_collection", peer_api_uris=[new_url])
```
timvisee, line 62: This makes sense! If we query too early we might not have the data applied from the snapshot yet. IIRC we normally handle this through `/readyz` in cloud.

PR #5556, opened 2024-11-30
Reduce debug size

xzfc:
Generic methods and iterator chains tend to bloat the binary size of debug builds. This is particularly bad when they are used together (as it's multiplicative, not additive).
This PR reduces the binary size of debug builds by 275M by rewriting some iterator chains with plain loops and removing unnecessary closures.

Sizes calculated by `cargo build --bin qdrant && du -m target/debug/qdrant`:
| Commit                                      |  Size | Change |
| ------------------------------------------- | ----: | -----: |
| Original (ec7caae3)                         | 1660M |        |
| RawScorerImpl::peek_top_iter                | 1439M |  -221M |
| TypedMultiDenseVector::multi_vectors() uses | 1404M |   -35M |
| stop_condition -> stopped                   | 1385M |   -19M |

PR #7514, opened 2025-11-11
Batched scans for dense vectors

monoid:
Previously each element of batch request was processed separately.  This change handles them together in case of linear scans, exploiting memory caching to increase throughput by 2.8x.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/hnsw_index/point_scorer.rs
+++ b/lib/segment/src/index/hnsw_index/point_scorer.rs
@@ -459,6 +459,7 @@ impl<'a> BatchFilteredSearcher<'a> {
                 break;
             }
 
+            // Switching the loops improves batching performance, but slightly degrades single-query performance.
             for BatchSearch { raw_scorer, pq } in &mut self.scorer_batch {
                 raw_scorer.score_points(&chunk[..chunk_size], &mut scores_buffer[..chunk_size]);
 


generall:
@monoid please resolve merge conflicts

monoid:
> @monoid please resolve merge conflicts

Done.

lib/segment/src/index/hnsw_index/point_scorer.rs
IvanPleshkov, line 1: At the beginning of this file there is a doc comment with the full scorers architecture. I guess we need to update this doc if we add a new entity to this schema.

lib/segment/src/index/hnsw_index/point_scorer.rs
```diff
@@ -309,3 +310,158 @@ impl<'a> FilteredScorer<'a> {
...
+            },
+            top,
+        }
+    }
+
+    pub fn peek_top_all(
+        self,
+        is_stopped: &AtomicBool,
+    ) -> CancellableResult<Vec<Vec<ScoredPointOffset>>> {
+        let iter = self
```
IvanPleshkov, line 421: I see that you copied this logic from `FilteredScorer` (btw can we somehow avoid copying of logic and reuse it?). But it seems we have a bug here and we don't use `vec_deleted` and we can score the vector from existing point but without their vector (it's possible). Need a second eye from @timvisee who, as I remember, implemented a feature of points without vectors

IvanPleshkov: I guess we didn't meet this problem because of zeroing of non-existing vector. But it is still unnecessary scores and some corner cases.

IvanPleshkov: As a reference of correct usage of deleted flags you may find in `check_deleted_condition`. It's important logic and the order of checks matters and also we cover cases where deleted bit is missing - there were some bugs about it. I place a changes request because of this thread

lib/segment/src/index/hnsw_index/point_scorer.rs
```diff
@@ -309,3 +310,158 @@ impl<'a> FilteredScorer<'a> {
         Ok(pq.into_sorted_vec())
     }
 }
+
+struct BatchSearch<'a> {
+    raw_scorer: Box<dyn RawScorer + 'a>,
+    pq: FixedLengthPriorityQueue<ScoredPointOffset>,
```
IvanPleshkov, line 330: Do we actually need to keep `pq` as a field? Anyway, the only one usage of `pq` is `peek_top_iter` where we take ownership. We can create `FixedLengthPriorityQueue` inside `peek_top_iter`. After removing this field, only `RawScorer` is left and we can just remove this structure and have less entities in the project

lib/segment/src/index/hnsw_index/point_scorer.rs
```diff
@@ -309,3 +310,158 @@ impl<'a> FilteredScorer<'a> {
...
+}
+
+pub struct BatchFilteredSearcher<'a> {
```
IvanPleshkov, line 333: Proposal. Because this struct is a batched variant of `FilteredScorer`, I propose to name it `BatchedFilteredScorer`. Of course it's funny that this proposal produces a `BatchedFilteredScorer` without `score` method. As an alternative, we can rename `FilteredScorer`. The proposal is about names which show the equivalence between `FilteredScorer` and `BatchFilteredSearcher`.

IvanPleshkov: Not a proposal, just an idea to discuss: do we need `FilteredScorer` after this changes? I mean we can use only batched variant of filtered scorer and extend the logic in `FilteredScorer` methods. In this way we don't introduce any new `struct` in scorers

lib/segment/src/index/hnsw_index/point_scorer.rs
```diff
@@ -309,3 +310,158 @@ impl<'a> FilteredScorer<'a> {
...
+    ) -> CancellableResult<Vec<Vec<ScoredPointOffset>>> {
+        if self.top == 0 {
+            return Ok(vec![vec![]; self.scorer_batch.len()]);
+        }
+
+        // Reuse the same buffer for all chunks, to avoid reallocation
+        let mut chunk = [0; VECTOR_READ_BATCH_SIZE];
+        let mut scores_buffer = [0.0; VECTOR_READ_BATCH_SIZE];
+
+        loop {
```
IvanPleshkov, line 442: Is it possible to refactor this part and avoid infinity `loop`? I can propose something like this:
```
use itertools::Itertools;
let mut chunk = [0; VECTOR_READ_BATCH_SIZE];
let filter = |point_id: &PointOffsetType| { !self.filters.check_vector(*point_id) };
for c in &points.filter(filter).chunks(64) {
    check_process_stopped(is_stopped)?;
    for (i, point_id) in c.enumerate() {
        chunk[i] = point_id;
    }
}
```
Stopped flag is checked once per 64 filtering checks. I find it OK. and we can do something if we want to check a stop condition each time

lib/segment/src/index/hnsw_index/point_scorer.rs
IvanPleshkov: I got that you need a separate structure to avoid a second `SmallVec`. So I don't push this proposal, it's up to you, you can close this conversation. But I would like to have a comment if I'm right with my guess. And better naming to show that it's just a helper structure to avoid allocation. It wasn't obvious

monoid: Values stored together indeed reduce allocations, and also increase data locality.

lib/segment/src/index/hnsw_index/point_scorer.rs
monoid: The plan is to remove `FilteredScorer`'s methods after introducing `BatchFilteredSearcher` to sparse vectors.

agourlay:
> exploiting memory caching to increase throughput by 2.8x.

What is exactly the benchmark setup here? Micro benchmark or end to end?

lib/segment/src/index/hnsw_index/point_scorer.rs
IvanPleshkov: Please comment or mark it somehow in the code, it wasn't obvious at first look

monoid:
> > exploiting memory caching to increase throughput by 2.8x.
> 
> What is exactly the benchmark setup here? Micro benchmark or end to end?

It was a bfb benchmarks with 8-16 search threads and batch search.  But it was an edge-case: full linear scan.

lib/segment/src/index/hnsw_index/point_scorer.rs
xzfc: Historical note: it was iterator-based before #5556; changed to plain loops to reduce debug builds size. Tho, perhaps, such optimization is not that relevant nowadays since currently this method is not a part of `RawScorer`. (`RawScorer` impls are genericized like hell)

IvanPleshkov: Thanks for this note, didn't expect so huge cost of such change. Resolve this thread

lib/segment/src/index/hnsw_index/point_scorer.rs
monoid: Yep, I've added additional scheme.

lib/segment/src/index/hnsw_index/point_scorer.rs
monoid: Done.

lib/segment/src/index/hnsw_index/point_scorer.rs
IvanPleshkov: I found that everything is fine here. But not obvious. We actually call `check_deleted_condition` inside `peek_top_iter`, which is called in `peek_top_all`.
If you wish, you can mark in this line that we do an actual delete filtration later and there is just a superset of actual points. But it's not a strong proposal because I dont know how to describe is properly

lib/segment/src/index/plain_vector_index.rs
```diff
@@ -97,101 +97,64 @@ impl VectorIndex for PlainVectorIndex {
...
-                        Ok(res)
-                    })
-                    .collect()
+                batch_searcher.peek_top_iter(&mut filtered_ids_vec.iter().copied(), &is_stopped)?
             }
+            None => batch_searcher.peek_top_all(&is_stopped)?,
+        };
+
+        for (search_result, query_vector) in search_results.iter_mut().zip(query_vectors) {
+            *search_result = postprocess_search_result(
```
IvanPleshkov, line 146: I like that you reduced allocations count

monoid: Thanks.  BTW, I think about making it more explicit by `postprocess_search_result` having a `search_result: &mut ...`.

IvanPleshkov: I propose to do it as a separate PR after merging this changes

monoid:
For a more realistic example, `bfb -d 128 --keywords 10 --search-batch-size 50` gives 2x speedup.

lib/segment/src/index/hnsw_index/point_scorer.rs
```diff
@@ -37,6 +38,19 @@ use crate::vector_storage::{
 ///                                                              - Scoring logic
 ///                                                              - Complex queries
 /// ```
+///
+/// The `BatchFilteredSearcher` contains an array of `RawScorer`s, a common filter and certain parameters.
+///
+/// ```plaintext
+///  BatchFilterScorer      RawScorer
```
timvisee:
```suggestion
///  BatchFilteredSearcher  RawScorer
```

timvisee (approved):
Everything seems handled. Thanks! 🙌

lib/segment/src/index/hnsw_index/point_scorer.rs
monoid: Thanks, it is already in my private commmit :)

lib/segment/src/index/hnsw_index/point_scorer.rs
IvanPleshkov: Open thread back after https://github.com/qdrant/qdrant/commit/6918419614ceeb302aa800a270ecd079c6d34725
where `check_deleted` was removed from `peek_top_iter `

monoid:
> After this commit [6918419](https://github.com/qdrant/qdrant/commit/6918419614ceeb302aa800a270ecd079c6d34725) the topic about deleted flags [#7514 (comment)](https://github.com/qdrant/qdrant/pull/7514#discussion_r2533153646) is still actual, we dont use deleted flag correctly in `peek_top_all`. Mark back as changes request

This commit is incorrect, I'm fixing it.

~~However, in postprocessing deleted point is not actually used.  It is a required argument for `FilteredScorer`, but is used in `peek_*` methods which will be eventually removed.~~

PR #7516, opened 2025-11-11
Final metrics tweaks for Qdrant 1.16.0

timvisee:
Here I make and propose some final changes to our current metrics for the release of Qdrant 1.16.0.

Specifically this does:

1. re-adds `collections_vector_total` metric (for backwards compatibility)
2. namespaces new replica metrics with `collection_` (to best follow my interpretation of [recommendations](https://prometheus.io/docs/practices/naming/#metric-names)):
    - `active_replicas_{min,max}` -> `collection_active_replicas_{min,max}`
    - `dead_replicas` -> `collection_dead_replicas`
3. count `collection_running_optimizations` per collection, as I think that's much more interesting in practice

We keep the new `qdrant_` prefix disabled for all metrics to not break existing deployments. Users have control over the prefix if they ever want to enable it for their deployment.

Please see this PR as a request for comment. If we're happy with this final result, we can merge.

Here is a dump of all metrics: https://gist.github.com/timvisee/6a3e7968bcfc63b3cc1471788ca9ccbc

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -174,27 +174,29 @@ impl MetricsProvider for CollectionsTelemetry {
             prefix,
         ));
 
+        let num_collections = self.collections.as_ref().map_or(0, |c| c.len());
+
         // Optimizers
-        let mut total_optimizations_running = vec![];
+        let mut total_optimizations_running = Vec::with_capacity(num_collections);
 
         // Min/Max/Expected/Active replicas over all shards.
         let mut total_min_active_replicas = usize::MAX;
         let mut total_max_active_replicas = 0;
 
         // Points per collection
-        let mut points_per_collection = vec![];
+        let mut points_per_collection = Vec::with_capacity(num_collections);
 
         // Vectors excluded from index-only requests.
-        let mut indexed_only_excluded = vec![];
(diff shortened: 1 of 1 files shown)


src/common/metrics.rs
```diff
@@ -353,28 +369,30 @@ impl MetricsProvider for CollectionsTelemetry {
...
         ));
 
-        metrics.push(metric_family(
-            "optimizer_running_processes",
-            "number of currently running optimization processes",
-            MetricType::GAUGE,
-            vec![gauge(total_optimizations_running as f64, &[])],
-            prefix,
-        ));
+        if !total_optimizations_running.is_empty() {
```
JojiiOfficial, line 388: We don't need these empty checks anymore but we can clean them up later

timvisee: I believe we need to merge <https://github.com/qdrant/qdrant/pull/7479> for that.

I didn't do that yet because I figured it wasn't very important for the upcoming release.

JojiiOfficial: Good catch, I thought this was already merged 👍🏻

PR #7517, opened 2025-11-11
Fix hash ring initialization for custom sharding, prevent resharding panic

timvisee:
We currently don't initialize hash rings correctly when custom sharding is used.

Specifically, when custom sharding is used, we also initialize a hash ring for the `None` key. I'd argue this is not correct.

This PR changes the shard holder loading behavior to not create the `None` hash ring anymore. We now strictly define what kind of sharding is used, and derive the default rings from there.

We had a production cluster panicking on resharding because no shard key was provided. Loading the hash rings as described in this PR would have prevented the panic with a graceful error. Applying this fix on a cluster that is already in a crash loop due to this problem will resolve the issue.

This PR adds some tests to ensure we check the shard count precondition before resharding. It also ensures we don't crash like we did before.

Tried with:

```
PUT collections/test
{
  "vectors": {
    "distance": "Cosine",
    "size": 1
  },
  "sharding_method": "custom"
}

POST collections/test/cluster
{
  "create_sharding_key": {
    "shard_key": "test",
    "shards_number": 2
  }
}

POST collections/test/cluster
{
  "start_resharding": {
    "direction": "down",
    "shard_key": null
  }
}
```
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/shard_holder/mod.rs
+++ b/lib/collection/src/shards/shard_holder/mod.rs
@@ -271,7 +271,8 @@ impl ShardHolder {
                     (self.sharding_method, &shard_key),
                     (ShardingMethod::Auto, None) | (ShardingMethod::Custom, Some(_)),
                 ),
-                "auto sharding cannot have shard key, custom sharding must have shard key",
+                "auto sharding cannot have shard key, custom sharding must have shard key ({:?}, {shard_key:?})",
+                self.sharding_method,
             );
             rings
                 .entry(shard_key)


lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -77,7 +78,10 @@ impl ShardHolder {
         }
     }
 
-    pub fn new(collection_path: &Path) -> CollectionResult<Self> {
+    pub fn new(
+        collection_path: &Path,
+        collection_config: &CollectionConfigInternal,
```
ffuugoo: I'd say just pass sharding method, not config (less headache in the future due to redundant dependency on collection config), but we can fix this later.

timvisee: I agree that'd be better.

I'll patch that now and merge the PR after.

agourlay:
Do we have time for adding tests right now?

timvisee:
> Do we have time for adding tests right now?

Yes!

timvisee:
@agourlay I've added some end to end tests to ensure this cannot happen again.

Most critically it asserts that:
1. that we don't crash on the customers edge case, like we did before
2. that we check preconditions for shard count before resharding

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -246,10 +255,24 @@ impl ShardHolder {
...
+        // Add shards and shard keys
         let ids_to_key = self.get_shard_id_to_key_mapping();
         for shard_id in self.shards.keys() {
             let shard_key = ids_to_key.get(shard_id).cloned();
+            debug_assert!(
+                matches!(
+                    (self.sharding_method, &shard_key),
+                    (ShardingMethod::Auto, None) | (ShardingMethod::Custom, Some(_)),
+                ),
+                "auto sharding cannot have shard key, custom sharding must have shard key",
```
ffuugoo:
```suggestion
                "auto sharding cannot have shard key, custom sharding must have shard key ({:?}, {shard_key:?})",
                self.sharding_method,
```

agourlay (approved):
Thanks a lot for adding those great tests 👏

PR #7520, opened 2025-11-12
P-Square one pass quantile estimation method

IvanPleshkov:
This PR implements P square alg for approximate quantiles.
https://www.cse.wustl.edu/~jain/papers/ftp/psqr.pdf

# Problem
In SQ we calculate min-max values to make an integer approximation of float. Also, we have `quantile` parameter to skip very large/small values.

# Quantile definition
If, for instance, the `quantile` is equal to 0.95, we need to calculate min-max without 2.5% of the largest and 2.5% of the smallest values.

In dev, we take a random sample from dataset, sort all values in this sample, remove largest-smallest and calculate min-max.

But for improvements of SQ we cannot use this method because we want to calculate this min-max with quantile for each dimension. It means, that we should take `DIM` times more sample (too much for RAM). Or we have an inaccurate quantile.

# Solution
There are methods to calculate approximate quantile using one-pass through the data. One of such alg is P-square. It takes quantile (for instance 2%), take an iterator of values and returns the approximate`x` with condition: 2% of values are smaller than `x`. The same for quantile 98% - 2% are larger than `x`.

# Another methods
There are modern algs like TDigits or Greenwald–Khanna ect. They allow us to ask the quantile value after data observation, but they require more memory (like hundreds of markers for each estimation). Because for impoved SQ we want to calculate quantiles for each dimension in vector, I find it expensive. Moreover, the quantile value we know before observation and P-square is a good here. TDigits should work better on heavy tails.

# P-Square extension
Because P-Square can have less accuracy on heavy tails, this PR implements a simple extension. By the article, alg requires 5 control markers for the estimation. This PR allows to use any odd number of markers to improve accuracy in cost of performance.

# Tests
This PR contains tests with different distributions including 2-Student with heavy tails and infinite variance.

### DIFF
--- a/lib/quantization/src/p_square.rs
+++ b/lib/quantization/src/p_square.rs
@@ -38,23 +38,22 @@ impl<const N: usize> P2Quantile<N> {
 
     /// Push one observation.
     pub fn push(&mut self, x: f64) {
+        let Ok(x) = NotNan::new(x) else {
+            return;
+        };
         if !x.is_finite() {
             return;
         }
 
         match self {
             P2Quantile::Linear(linear) => {
-                let Ok(x) = NotNan::new(x) else {
-                    return;
-                };
-
                 // in linear case just collect observations until we have N of them
                 linear.observations.push(x);
                 if linear.observations.len() == N {
                     *self = P2Quantile::Impl(P2QuantileImpl::new_from_linear(linear));
                 }
             }
-            P2Quantile::Impl(p2) => p2.push(x),
+            P2Quantile::Impl(p2) => p2.push(*x),
         }
     }
 


lib/quantization/src/p_square.rs
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    fn update_desired_position(&mut self, n: usize) {
+        self.n_desired = 1.0 + self.target_propability * (n as f64 - 1.0);
+    }
+}
+
+pub struct P2QuantileLinear<const N: usize> {
+    quantile: f64,
+    observations: smallvec::SmallVec<[f64; N]>,
```
monoid: Should be `arrayvec` (never allocates, less branches).

lib/quantization/src/p_square.rs
```diff
@@ -0,0 +1,471 @@
...
+    count: usize,
+    markers: [Marker; N],
+}
+
+impl<const N: usize> P2QuantileImpl<N> {
+    fn new_from_linear(linear: &P2QuantileLinear<N>) -> Self {
+        assert_eq!(linear.observations.len(), N);
+
+        let mut buf = linear.observations.clone();
+        buf.sort_unstable_by(|a, b| a.partial_cmp(b).unwrap());
```
monoid: Here and below it relies on `!x.is_finite()` in another type. I suggest expressing it at type level with `ordered_float::NotNan`.

It seems that only `P2QuantileLinear` needs it.

lib/quantization/src/p_square.rs
IvanPleshkov: Done, I also moved the linear case into a separate method with a slice. I use it in tests to check that the approximate result is close to the linear one.

lib/quantization/src/p_square.rs
IvanPleshkov: Thanks! I added `NotNan` and removed unnecessary `unwrap`s. I kept `is_finite` check for non-linear case

xzfc:
EDIT: the following probably not that important as per https://github.com/qdrant/qdrant/pull/7520#issuecomment-3545033496

---

If we really want to constrain memory usage:

1. Do we need it to operate on `f64` rather than `f32`? Our vector data is `f32`.
Making `Marker` to use `f32` would cut memory usage by half; tho I'm not sure whether would it affect quality of estimations.

2. `count` can be `u32` instead of `usize`. Our segments doesn't grow that much. Saves 4 bytes.

3. Instead of `enum P2Quantile`, we can have just `struct P2QuantileImpl`. The `count` decides whether it's linear or P². Saves $4…8$ bytes.

4. We don't need to store `P2QuantileLinear::quantile` and `Marker::target_probability` inside each estimator.
   We can use a single precalculated value per vector (so store it once, not for each dimension).
   Tho, it would slightly complicate the API: we'd need to pass these precalculated values during each `push()`.
   Saves $N \cdot 4…8$ bytes.
   (I'm assuming that `quantile` is expected to be the same across all vector dimensions)

5. I think `Marker::n_position` can be `u8`.
   (would require replacing array-of-structs with struct-of-arrays)
   Saves $3…7$ bytes.

All of the above would cut memory from $240$ to $68$ bytes per estimator. ($N=7$)

xzfc:
> There are modern algs like TDigits or Greenwald–Khanna ect. They allow us to ask the quantile value after data observation, but they require more memory (like hundreds of markers for each estimation). Because for impoved SQ we want to calculate quantiles for each dimension in vector, I find it expensive.

Maybe a naive question, sorry for that:
I assume we need this estimator only for mutable segments?
1 KiB per dimension would mean 2 MiB of data (for 2K vectors); or the size of 256 vectors.
Can't we afford the size of 256 vectors per mutable segment?

lib/quantization/src/p_square.rs
```diff
@@ -0,0 +1,482 @@
...
+        if N < 5 {
+            return Err(EncodingError::EncodingError(
+                "P2Quantile requires at least 5 markers".to_string(),
+            ));
+        }
+        if N.is_multiple_of(2) {
+            return Err(EncodingError::EncodingError(
+                "P2Quantile requires an odd number of markers".to_string(),
+            ));
+        }
```
xzfc: Nit: can be a compile-time error rather than runtime check:

```suggestion
        const {
            assert!(N >= 5, "P2Quantile requires at least 5 markers");
            assert!(N % 2 == 1, "P2Quantile requires an odd number of markers");
        };
```

lib/quantization/src/p_square.rs
```diff
@@ -0,0 +1,482 @@
...
+
+/// Extended version of P-square one-quantile estimator by Jain & Chlamtac (1985).
+/// https://www.cse.wustl.edu/~jain/papers/ftp/psqr.pdf
+/// By default, P-square uses 5 markers to estimate a single quantile.
+/// This implementation is extended to support an arbitrary odd number of markers N >= 5
+///
+/// Usage:
+/// let mut p2 = P2Quantile::<7>::new(0.99).unwrap();
+/// for x in data { p2.push(x).unwrap(); }
+/// let q_hat = p2.estimate();
```
xzfc, line 15: Rustdoc markup nit:
1. See https://rust-lang.github.io/rust-clippy/master/index.html#too_long_first_doc_paragraph
2. Put code blocks in ` ``` ` so it won't be treated as markdown.

```suggestion
/// Extended version of P-square one-quantile estimator by Jain & Chlamtac (1985).
///
/// <https://www.cse.wustl.edu/~jain/papers/ftp/psqr.pdf>
/// By default, P-square uses 5 markers to estimate a single quantile.
/// This implementation is extended to support an arbitrary odd number of markers N >= 5
///
/// Usage:
/// ```ignore
/// let mut p2 = P2Quantile::<7>::new(0.99).unwrap();
/// for x in data { p2.push(x).unwrap(); }
/// let q_hat = p2.estimate();
/// ```
```

IvanPleshkov: Thanks! Despite we dont use rustdoc as I know, I agree to follow rustdoc markup

lib/quantization/src/p_square.rs
IvanPleshkov: Not a nit, I find it important, thanks!

IvanPleshkov:
> > There are modern algs like TDigits or Greenwald–Khanna ect. They allow us to ask the quantile value after data observation, but they require more memory (like hundreds of markers for each estimation). Because for impoved SQ we want to calculate quantiles for each dimension in vector, I find it expensive.
> 
> Maybe a naive question, sorry for that: I assume we need this estimator only for mutable segments? 1 KiB per dimension would mean 2 MiB of data (for 2K vectors); or the size of 256 vectors. Can't we afford the size of 256 vectors per mutable segment?

Nothing to be sorry about. We need this estimator once per segment building as a temporary step. We do estimation, build SQ and destroy the estimator. In comparison with TDigits I guess you are right, and I was too dramatic. I vibecoded for testing TDigits implementation and found that a better quality than this PR I have on >1K markers. For 2K vectors it's less than 10MB. Despite P2 requires much less, we can afford TDigits requirements.

But still I prefer P-square. It's simpler to impelement; an extended version can fight heavy tails (I don't expect heavy tails on real or random data, it's a corner case). We don't need advantage of TDigits: ask any quantile after observation. We can just keep 2 P-Square for min and for max.

There is TDigits on crates,
https://github.com/MnO2/t-digest
But it's not maintained and doesn't have stars, I don't want to include this psckage into production environment

lib/quantization/src/p_square.rs
```diff
@@ -0,0 +1,479 @@
...
+            quantile: q,
+            observations: Default::default(),
+        }))
+    }
+
+    /// Push one observation.
+    pub fn push(&mut self, x: f64) {
+        if !x.is_finite() {
+            return;
+        }
```
monoid, line 46:
```suggestion
        let Ok(x) = NotNan::new(x) else {
            return;
        };
```

And for the `P2Quantile::Impl` branch just use `*x`.  Express the information in types!

lib/quantization/src/p_square.rs
monoid: I'm sorry, you forgot to remove it here.

lib/quantization/src/p_square.rs
```diff
@@ -0,0 +1,479 @@
...
+    observations.sort_unstable();
+
+    let k = quantile * (observations.len() as f64 - 1.0);
+    let lo = k.floor() as usize;
+    let hi = k.ceil() as usize;
+    if lo == hi {
+        observations[lo].into_inner()
+    } else {
+        let frac = k - lo as f64;
+        observations[lo].into_inner() + frac * (observations[hi] - observations[lo]).into_inner()
```
monoid: IMHO, this is cleaner:
```suggestion
        *observations[lo] + frac * (*observations[hi] - *observations[lo])
```
And another `NotNan::into_inner` occurrences.

lib/quantization/Cargo.toml
```diff
@@ -31,11 +31,14 @@ io = { path = "../common/io" }
 strum = { workspace = true }
 bytemuck = { workspace = true }
 parking_lot = { workspace = true }
+ordered-float = { workspace = true }
+arrayvec = "0.7.6"
```
monoid, line 35: Do we have a policy for workspace deps?

lib/quantization/src/p_square.rs
IvanPleshkov: `NotNan` allows storing infinity. That's why I didn't remove `is_finite` which also filters infinity and negative infinity

monoid: I see, sorry.

lib/quantization/src/p_square.rs
IvanPleshkov: Thanks! Done

lib/quantization/Cargo.toml
IvanPleshkov: I cannot find it in notions. In case of arrayvec there is only one usage in quantization. But I moved to workspace `rand_distr` crate because it also is used in segment

monoid (approved):

lib/quantization/src/p_square.rs
```diff
@@ -0,0 +1,479 @@
...
+    pub fn push(&mut self, x: f64) {
+        if !x.is_finite() {
+            return;
+        }
+
+        match self {
+            P2Quantile::Linear(linear) => {
+                let Ok(x) = NotNan::new(x) else {
+                    return;
+                };
```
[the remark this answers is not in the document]
IvanPleshkov: in this case I find unwrap worse than double-check for nan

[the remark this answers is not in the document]
monoid: Perhaps, getting a `NotNan` and then checking it is finite would be better.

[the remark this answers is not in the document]
IvanPleshkov: I changed the order of the checks

xzfc (approved):

PR #7523, opened 2025-11-12
Clarify multivector inline storage warning

agourlay:
Multivector supports quantization which creates confusing error messages.

e.g. 

```
CollectionWarning {
    message: "The `hnsw_config.inline_storage` option for vector 'multi-dense-vector-float16-2' is not compatible with multivectors. This option will be ignored.",
},
CollectionWarning {
    message: "The `hnsw_config.inline_storage` option for vector 'multi-dense-vector-on-disk-1' requires quantization to be enabled. This option will be ignored.",
},
```

By checking first the multivector aspect, we can have an early exit and never pretend that quantization is required.

### DIFF
--- a/tests/openapi/test_vector_config.py
+++ b/tests/openapi/test_vector_config.py
@@ -244,7 +244,7 @@ def test_disable_indexing(on_disk_vectors):
                 "hnsw_config": {"inline_storage": True},
                 "multivector_config": {"comparator": "max_sim"},
             },
-            2,
+            1,
         ),
     ],
 )


lib/collection/src/config.rs
```diff
@@ -289,19 +289,19 @@ impl CollectionConfigInternal {
...
                     warnings.push(CollectionWarning {
                         message: format!(
                             "The `hnsw_config.inline_storage` option for vector '{vector_name}' \
-                             requires quantization to be enabled. This option will be ignored."
+                             is not compatible with multivectors. This option will be ignored."
                         ),
                     });
                 }
-                if vector_config.multivector_config.is_some() {
+                if !vector_quantization {
```
xzfc: Did you intend to use `else if` here?

agourlay: Just moved the block around.

I like the `else if` idea 👍

xzfc: I thought by "have an early exit" you mean something like this.

agourlay: I misread the code, I hallucinated a return after push.

So the `else if` is what I actually wanted, good catch :+1:

PR #7524, opened 2025-11-12
Fix missing version debug log

agourlay:
This `DEBUG` log appears often during crash testing, decided to finally fix it!


timvisee (approved):

lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -202,7 +202,7 @@ pub trait IdTracker: fmt::Debug {
             self.drop(external_id)?;
             #[cfg(debug_assertions)]
             {
-                log::debug!("dropped version for point {external_id} without version");
+                log::debug!("dropped mapping for point {external_id} without version");
             }
```
timvisee: We can do this now:

```suggestion
            log::debug!("dropped mapping for point {external_id} without version");
```

PR #7528, opened 2025-11-13
allow initial_state in API

generall:


joein:
https://github.com/qdrant/qdrant-client/pull/1109

PR #7530, opened 2025-11-13
Abort pending blocking search tasks when we drop/invalidate the search

timvisee:
Critical fix for large clusters with huge search load.

Searches spawn blocking search tasks on a dedicated thread pool. The pool size is limited, and so these search tasks may be queued. If there is a humongous amount of incoming searches, this queue may get infinitely long.

When a search is invalidated (timed out, or completed through fan out), we drop the async task. The idea is that we abort all work related to this search to release resources.

It turns out that these spawned tasks will always remain queued and will be run to completion, even if the owning future was already dropped and aborted. Instead, we must explicitly cancel these pending spawned blocking tasks.

Luckily tokio provides us the [`AbortOnDropHandle`](https://docs.rs/tokio-util/latest/tokio_util/task/struct.AbortOnDropHandle.html) utility, which is what I've implemented in this PR.

On a huge cluster that accumulate a massive amount of pending searches, it was possible to keep segments busy for more than an hour. Even though each search itself might only be a few seconds of work. This can break optimizations if they cannot release old segments for over an hour. This PR will help prevent this issue from happening by cancelling all invalidated searches early. More specifically, it helps prevent this error from happening:

```
Service internal error: Removing proxy segment which is still in use
```

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -251,6 +251,7 @@ impl SegmentsSearcher {
                     // all searches for futures that are already dropped. Not using this allows
                     // users to create a humongous queue of search tasks, even though the searches
                     // are already invalidated.
+                    // See: <https://github.com/qdrant/qdrant/pull/7530>
                     let search = AbortOnDropHandle::new(search);
 
                     (segment, search)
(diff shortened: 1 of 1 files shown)


agourlay (approved):
I trust that you have tested this manually 👍

timvisee:
> I trust that you have tested this manually 👍

Correct.

We can add a test for it, but I'd consider that obsolete. Because then we'd be testing tokio fundamentals. [`spawn_`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) functions are intended to keep the task even if the caller is dropped. It was simply an oversight from our side when implementing this.

PR #7531, opened 2025-11-13
more stop flags

generall:
`iter_filtered_points` under some special conditions could consume significant resources and didn't respect stopflag in all cases.
Now it does.

I went ahead with another implementation of stopable iterator, because it is easier and apparently also slightly faster (i did   benchmark).

I tried to remove newly introduced double checking where I spotted it.

Also we want to check stopflag right after starting search, cause it could have become outdated while waiting in runtime queue.

### DIFF
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -523,8 +523,8 @@ pub trait SegmentOptimizer {
         let desired_cpus = permit.num_io as usize;
         let indexing_permit = resource_budget
             .replace_with(permit, desired_cpus, 0, stopped)
-            .map_err(|_| CollectionError::Cancelled {
-                description: "optimization cancelled while waiting for budget".to_string(),
+            .map_err(|_| {
+                CollectionError::cancelled("optimization cancelled while waiting for budget")
             })?;
 
         let mut rng = rand::rng();
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -534,9 +534,9 @@ fn search_in_segment(
     segment_query_context: &SegmentQueryContext,
 ) -> CollectionResult<(Vec<Vec<ScoredPoint>>, Vec<bool>)> {
     if segment_query_context.is_stopped() {
-        return Err(CollectionError::Cancelled {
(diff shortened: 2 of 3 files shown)


lib/segment/src/index/struct_payload_index.rs
generall, line 1: main changes in this file

lib/common/common/src/iterator_ext/stoppable_iter.rs
```diff
@@ -0,0 +1,28 @@
...
+    I: Iterator,
+{
+    type Item = I::Item;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // AI says, that if atomic is not updated frequently,
+        // it is cheap enough to check it on every iteration.
+        if self.is_stopped.load(Ordering::Relaxed) {
+            return None;
+        }
```
timvisee, line 25: So here's a stupid idea:

I'm curious if we could just ignore atomics here and concurrently set the boolean.

It'd be cheaper by not using atomic operations, at the cost of potentially 'recognizing' the changed flag a bit later due to local caches.

lib/common/common/src/iterator_ext/stoppable_iter.rs
```diff
@@ -0,0 +1,28 @@
...
+impl<'a, I> Iterator for StoppableIter<'a, I>
+where
+    I: Iterator,
+{
+    type Item = I::Item;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // AI says, that if atomic is not updated frequently,
+        // it is cheap enough to check it on every iteration.
+        if self.is_stopped.load(Ordering::Relaxed) {
```
timvisee, line 23: I also checked the [`unlikely`](https://doc.rust-lang.org/std/hint/fn.unlikely.html) and `cold_path` intrinsic just for fun, but they aren't stable yet :(

timvisee (approved):
🙌

lib/common/common/src/iterator_ext/stoppable_iter.rs
timvisee: Turns out the non-atomic variant is exactly the same for reads on x86_64 and ARM:

https://rust.godbolt.org/z/oe4f17xEo

Thanks @xzfc !

agourlay:
Thanks for the benchmarks supporting the change :+1:

PR #7533, opened 2025-11-14
Audit all spawn blocking calls, prematurely abort them

timvisee:
Similar to <https://github.com/qdrant/qdrant/pull/7530>, but for all other spawned blocking tasks.

Here I have audited all usages of spawning a blocking task. It turns out that we can prematurely cancel almost all of them. Some are in a similar vein as <https://github.com/qdrant/qdrant/pull/7530>, being for retrieve, scroll or facet operations.

Some of these tasks already use a cancellation token inside of them. But I still argue also prematurely aborting the task is better because it is free.

Other cases include aborting blocking tasks for fetching state, waiting on state, loading data.

I'd recommend to review this PR on a per-commit basis. I've separated each logical change into a dedicated commit. Each needs careful review.

I consider these changes critical as they might have significant effect on search performance under high load.

<https://github.com/qdrant/qdrant/pull/7531> takes care of better stop flag handling inside blocking tasks.

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/snapshot.rs
+++ b/lib/collection/src/shards/local_shard/snapshot.rs
@@ -15,6 +15,7 @@ use shard::locked_segment::LockedSegment;
 use shard::payload_index_schema::PayloadIndexSchema;
 use shard::segment_holder::{LockedSegmentHolder, SegmentHolder};
 use tokio::sync::oneshot;
+use tokio_util::task::AbortOnDropHandle;
 use wal::{Wal, WalOptions};
 
 use crate::operations::types::{CollectionError, CollectionResult};
@@ -92,7 +93,7 @@ impl LocalShard {
         let tar_c = tar.clone();
         let update_lock = self.update_operation_lock.clone();
 
-        tokio::task::spawn_blocking(move || {
+        let handle = tokio::task::spawn_blocking(move || {
             // Do not change segments while snapshotting
             snapshot_all_segments(
                 segments.clone(),
@@ -112,8 +113,8 @@ impl LocalShard {
             } else {
                 Self::snapshot_empty_wal(wal, &temp_path, &tar_c)
             }
-        })
-        .await??;
+        });
+        AbortOnDropHandle::new(handle).await??;
 
         LocalShardClocks::archive_data(&self.path, tar).await?;
 


lib/common/cancel/src/blocking.rs
```diff
@@ -45,7 +50,8 @@ where
         move || task(cancel)
     };
 
-    let output = future::cancel_on_token(cancel, tokio::task::spawn_blocking(task)).await??;
+    let handle = tokio::task::spawn_blocking(task);
+    let output = future::cancel_and_abort_on_token(cancel, handle).await?;
```
ffuugoo, line 54: I think we can simplify this. If token is triggered, `cancel_on_token` will simply exit early, which will drop `AbortOnDropHandle`, which will abort the task.

```suggestion
    let output = future::cancel_on_token(cancel, AbortOnDropHandle::new(tokio::task::spawn_blocking(task))).await??;
```

lib/common/cancel/src/future.rs
```diff
@@ -42,3 +44,28 @@ where
...
+    tokio::select! {
+        biased;
+        _ = cancel.cancelled() => {
+            // Prematurely abort blocking task if not started yet
+            abort_handle.abort();
+            Err(Error::Cancelled)
+        },
+        output = handle => Ok(output.map_err(Error::Join)?),
+    }
+}
```
ffuugoo, line 71: We can remove this then.

```suggestion
```

lib/common/cancel/src/future.rs
```diff
@@ -1,5 +1,7 @@
 use std::future::Future;
 
+use tokio::task::JoinHandle;
+
```
ffuugoo, line 4: And this.

```suggestion
```

src/common/telemetry.rs
```diff
@@ -79,14 +80,13 @@ impl TelemetryCollector {
...
-            runtime_handle
-                .spawn_blocking(move || {
-                    // Re-enter the async runtime in this blocking thread
-                    tokio::runtime::Handle::current().block_on(async move {
-                        CollectionsTelemetry::collect(detail, &access_collection, &toc).await
-                    })
+            let handle = runtime_handle.spawn_blocking(move || {
+                // Re-enter the async runtime in this blocking thread
+                tokio::runtime::Handle::current().block_on(async move {
+                    CollectionsTelemetry::collect(detail, &access_collection, &toc).await
```
ffuugoo, line 86: `AbortOnDrop` is fine here, but... why we do this in the first place? Blocking calls inside `CollectionsTelemetry::collect`?

```rust
runtime_handle.spawn(async move {
    CollectionsTelemetry::collect(detail, &access_collection, &toc).await
})
.await
```

lib/collection/src/common/stoppable_task.rs
```diff
@@ -3,11 +3,16 @@ use std::sync::atomic::{AtomicBool, Ordering};
...
 
 type PanicPayload = Box<dyn Any + Send + 'static>;
 
+/// A task that can be asked to stop
+///
+/// If this future is dropped the blocking task may be aborted prematurely if it has not started
+/// yet.
 pub struct StoppableTaskHandle<T> {
-    pub join_handle: JoinHandle<Option<T>>,
+    pub join_handle: AbortOnDropHandle<Option<T>>,
```
ffuugoo, line 15: > Prematurely abort stoppable task on drop, all are safe to abort early

I'll trust you on this one, but mark with a comment to double-check later.

lib/collection/src/shards/local_shard/snapshot.rs
```diff
@@ -112,8 +113,8 @@ impl LocalShard {
             } else {
                 Self::snapshot_empty_wal(wal, &temp_path, &tar_c)
             }
-        })
-        .await??;
+        });
+        AbortOnDropHandle::new(handle).await??;
```
agourlay, line 117: Not sure about this one 🤔 

I thought the snapshot should NOT be terminated even if the user drops the connection.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -486,7 +487,7 @@ impl Collection {
                     )
             });
 
-            match shard_transfer_requested.await {
+            match AbortOnDropHandle::new(shard_transfer_requested).await {
```
ffuugoo, line 490: It _should_ be fine, but marked with comment to double-check later. 😬

src/common/telemetry.rs
ffuugoo: @agourlay was this you?

lib/collection/src/shards/local_shard/snapshot.rs
timvisee: I agree it would be better to also terminate the snapshot operation half way. But I suggest to do that in a separate PR.

This PR was only intended to abort pending blocking tasks before they start to prematurely drain a humongous queue of pending blocking tasks. 

This is a blocking task, which we cannot abort/cancel without implementing a cancellation token in there.

A critical part in this snapshot logic is the proxying/unproxying of segments. All segments must be unproxied gracefully to recover the original segment holder state, even if cancelled.

src/common/telemetry.rs
agourlay: Here you go https://github.com/qdrant/qdrant/pull/7269

ffuugoo: Disregard this then. 🫡

lib/common/cancel/src/blocking.rs
timvisee: I don't think it behaves the same.

In my version it will not prematurely abort the task if the future (`cancel_on_token`) is dropped. In your version it does.

My thought was that we do not want that behavior, and that we should only abort/cancel when the token is explicitly triggered.

lib/collection/src/shards/local_shard/snapshot.rs
ffuugoo: I _think_ there should be one (or two??) `spawn` calls up the callstack from here, so it _should_ be fine to cancel this task, but I can double-check. 😬

lib/collection/src/collection/shard_transfer.rs
timvisee: This only waits on replica state changes. It doesn't change any state itself.

I therefore am sure this is safe.

Before this PR the check would still run at some future point in time, but we throw away the result of that check.

agourlay:
I am not sure about some of those changes TBH.

If the idea is to drain a humongous queue of pending blocking tasks, then we only need to apply the `AbortOnDropHandle` transformation to high frequency operations.

That is, covering the read requests should be enough, I feel like applying the pattern everywhere might be an over reaction and introduce some subtle cancellation changes.

timvisee:
> If the idea is to drain a humongous queue of pending blocking tasks, then we only need to apply the `AbortOnDropHandle` transformation to high frequency operations.

I see no point in distinguishing between the two as this change is practically free - except for a few extra lines of code.

I'd argue its the other way around. In many places we 'just' use `spawn_blocking` to run some blocking task, while we'd actually like to have the cancellation behavior. And so I don't see a point in throwing away this behavior for simple tasks. We never seem to have considered it properly.

lib/common/cancel/src/blocking.rs
ffuugoo: Ah, yeah, you're right. Disregard this then.

lib/collection/src/shards/local_shard/snapshot.rs
timvisee: I don't see a spawn on all call stacks.

For example, this path does not spawn: https://github.com/qdrant/qdrant/blob/7994816a1e383ffd7b1d840a843cdabcd967081d/src/actix/api/snapshot_api.rs#L378

ffuugoo: `helpers::time_or_accept` does spawn

timvisee:
For future readers: here is documentation on [`spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html), and [`abort`](https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#method.abort) which states it only prevents a task from running if not started yet, not aborting it mid-way.

PR #7538, opened 2025-11-14
avoid docstring comments for enum fields to make openapi nicer

generall:


joein:
https://github.com/qdrant/qdrant-client/pull/1114

joein:
In order to support `ReplicatePoints` in python client this PR is enough
However, if we want to add support of the other methods of `distributed API` in a patch release, we'd need to clean up a couple of more enums like `ClusterStatus` and `ConsensusThreadStatus`

PR #7556, opened 2025-11-18
Adjust log warning about trying to read lock a segment

timvisee:
The log message is wrong. It only tries to lock a single segment, not all of them.

I'd be nice to add a bit more context about the exact segment, but I don't see a good way of doing this without passing extra context into the function.

### All Submissions:

### DIFF
--- a/lib/shard/src/segment_holder/mod.rs
+++ b/lib/shard/src/segment_holder/mod.rs
@@ -607,7 +607,7 @@ impl SegmentHolder {
             interval = interval.saturating_mul(2);
             if interval.as_secs() >= 10 {
                 log::warn!(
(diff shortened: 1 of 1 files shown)


lib/shard/src/segment_holder/mod.rs
```diff
@@ -607,7 +607,7 @@ impl SegmentHolder {
             interval = interval.saturating_mul(2);
             if interval.as_secs() >= 10 {
                 log::warn!(
-                    "Trying to read-lock all collection segments is taking a long time. This could be a deadlock and may block new updates.",
+                    "Trying to read-lock a segments is taking a long time. This could be a deadlock and may block new updates.",
```
ffuugoo:
```suggestion
                    "Trying to read-lock a segment is taking a long time. This could be a deadlock and may block new updates.",
```

PR #7557, opened 2025-11-18
Missing validations for shard transfers

agourlay:
Adding missing validations for `RestartTransfer` for `gRPC` and `REST` APIs.

### DIFF
--- a/tests/consensus_tests/auth_tests/test_jwt_access.py
+++ b/tests/consensus_tests/auth_tests/test_jwt_access.py
@@ -1125,7 +1125,7 @@ def test_restart_transfer_operation():
             "restart_transfer": {
                 "shard_id": SHARD_ID,
                 "from_peer_id": PEER_ID,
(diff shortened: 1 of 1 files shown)


lib/api/src/grpc/validate.rs
```diff
@@ -178,13 +178,27 @@ impl Validate for grpc::DeleteShardKey {
...
+            self.shard_id,
+            self.to_shard_id,
+        )
     }
 }
 
 impl Validate for grpc::ReplicatePoints {
     fn validate(&self) -> Result<(), ValidationErrors> {
-        Ok(())
+        if self.from_shard_key != self.to_shard_key {
```
agourlay, line 192: opportunistic change copying the validation logic from REST

timvisee: Good change. This doesn't make sense between the same shard keys, because the points to replicate would already be there.

PR #7558, opened 2025-11-18
Await for consensus before unproxifying

generall:
We observed the `Unwrapping proxy shard` log message shortly followed by consensus error in production.

There are several problems with that:

- Current logic doesn't prevent `Unwrapping proxy shard` in case if the consensus is significantly delayed on receiver. It was checking for sahrd to be local, but didn't check if the relevant transfer status was received.
- There were no guarantee, that no other consensus operations should have been executed after `Unwrapping proxy shard`.
- There is no clear way to reproduce original consensus error, but having `Unwrapping proxy shard` in the out-of-order operation is suspecious enough for this PR

What this PR does:

- instead of unproxifying and then checking for consensus status, we first await that the receiver accepts consensus operation with the exact shard transfer we are waiting for.
- Only after awaiting for consensus we do checks for local shards, including the check for `Unwrapping proxy shard`
- `Unwrapping proxy shard` is promoted to an error, as we should not see it anymore, assuming the logic is not compromised somewhere else.

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -413,9 +413,9 @@ impl Collection {
         let collection_path = self.path.clone();
 
         async move {
-            let shards_holder = shards_holder.read_owned().await;
+            let shards_holder_guard = shards_holder.clone().read_owned().await;
 
-            let Some(replica_set) = shards_holder.get_shard(shard_id) else {
+            let Some(replica_set) = shards_holder_guard.get_shard(shard_id) else {
                 return Err(CollectionError::service_error(format!(
                     "Shard {shard_id} doesn't exist, repartition is not supported yet"
                 )));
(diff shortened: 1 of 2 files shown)


lib/collection/src/collection/shard_transfer.rs
```diff
@@ -505,7 +477,50 @@ impl Collection {
...
+
+            let shards_holder_guard = shards_holder.read_owned().await;
+
+            let Some(replica_set) = shards_holder_guard.get_shard(shard_id) else {
+                return Err(CollectionError::service_error(format!(
+                    "Shard {shard_id} doesn't exist, repartition is not supported yet"
+                )));
+            };
+
+            if replica_set.is_proxy().await {
```
timvisee, line 494: I like that we use `is_proxy()` now, which is much more clear.

agourlay (approved):

PR #7566, opened 2025-11-19
Storage compat. tests all supported versions

agourlay:
To avoid breaking changes between patch versions, we need to exhaustively test all compatible versions.

This requires downloading more archives and modifying the test runners to support a larger workload (thanks to @tellet-q).

~~All tests that share the same group name are guaranteed to run in the same worker process. This way we download each archive only once.~~

> Each test runs independently, so it downloads and then removes the files, any other strategy will lead to disk saturation. The grouping ensures that even we run pytest with several parallel workers these particular tests will run on the same worker sequentially.

# Future work

Handle properly the archive for `v1.15.0` which currently creates issues due to file system access control.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -261,9 +261,9 @@ jobs:
       - name: Verify Docker image
         run: docker images | grep qdrant
       - name: Run e2e tests
-        run: poetry --project tests run pytest -n 2 tests/e2e_tests -v --tb=short --durations=5
+        run: poetry --project tests run pytest -n 4 --dist=loadgroup tests/e2e_tests -v --tb=short --durations=5
         shell: bash
-        timeout-minutes: 5
+        timeout-minutes: 10
       - name: Upload test logs on failure
         uses: actions/upload-artifact@v5
         if: failure()
--- a/tests/e2e_tests/test_data_compatibility.py
+++ b/tests/e2e_tests/test_data_compatibility.py
@@ -10,7 +10,11 @@ from e2e_tests.utils import extract_archive
 
 
 class TestStorageCompatibility:
-    """Test storage and snapshot compatibility with defined previous Qdrant versions."""
+    """Test storage and snapshot compatibility with defined previous Qdrant versions.
+
+    These tests are grouped to run sequentially on the same worker using xdist_group marker.
+    This prevents parallel downloads of large archives and disk saturation.
+    """
 
     VERSIONS = [
         "v1.16.0",
@@ -19,7 +23,7 @@ class TestStorageCompatibility:
         "v1.15.3",
         "v1.15.2",
         "v1.15.1",
-        #"v1.15.0", # TODO when restoring Qdrant adds a missing payload_storage folder owned by `root` which messes up the cleanup
+        # "v1.15.0", # TODO when restoring Qdrant adds a missing payload_storage folder owned by `root` which messes up the cleanup
     ]
 
     EXPECTED_COLLECTIONS = [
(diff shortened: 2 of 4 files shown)


tests/e2e_tests/test_data_compatibility.py
```diff
@@ -12,8 +12,15 @@
...
-    PREV_PATCH_VERSION = "v1.16.0"
-    PREV_MINOR_VERSION = "v1.15.5"
+    VERSIONS = [
+        "v1.16.0",
+        "v1.15.5",
+        "v1.15.4",
+        "v1.15.3",
+        "v1.15.2",
+        "v1.15.1",
+        #"v1.15.0", # TODO when restoring Qdrant adds a missing payload_storage folder owned by `root` which messes up the cleanup
```
agourlay:
```
Using existing Docker image qdrant/qdrant:e2e-tests
Downloading compatibility data for v1.15.0...
Extracted /home/agourlay/Workspace/qdrant/tests/e2e_tests/test_storage_compatibility[v1.15.0]/storage/compatibility_v1.15.0_52880638.tar to /home/agourlay/Workspace/qdrant/tests/e2e_tests/test_storage_compatibility[v1.15.0]/storage
Cleaned up archive: /home/agourlay/Workspace/qdrant/tests/e2e_tests/test_storage_compatibility[v1.15.0]/storage/compatibility_v1.15.0_52880638.tar
Extracted /home/agourlay/Workspace/qdrant/tests/e2e_tests/test_storage_compatibility[v1.15.0]/storage/storage.tar.bz2 to /home/agourlay/Workspace/qdrant/tests/e2e_tests/test_storage_compatibility[v1.15.0]/storage
Cleaned up archive: /home/agourlay/Workspace/qdrant/tests/e2e_tests/test_storage_compatibility[v1.15.0]/storage/storage.tar.bz2
Container already removed. OK.
F
==================================================
TEST FAILED - DUMPING CONTAINER LOGS
==================================================
==================================================
Warning: Failed to remove temp storage directory /home/agourlay/Workspace/qdrant/tests/e2e_tests/test_storage_compatibility[v1.15.0]: [Errno 13] Permission denied: '/home/agourlay/Workspace/qdrant/tests/e2e_tests/test_storage_compatibility[v1.15.0]/storage/storage/collections/test_collection_product_x32/0/segments/930ece4a-cdb0-4fb5-86bf-5dfde0ab0024/payload_storage'
```

agourlay: I failed to run the Docker container with the current user via the Python API here.

.github/workflows/integration-tests.yml
```diff
@@ -261,7 +261,7 @@ jobs:
       - name: Verify Docker image
         run: docker images | grep qdrant
       - name: Run e2e tests
-        run: poetry --project tests run pytest -n 4 tests/e2e_tests -v --tb=short --durations=5
+        run: poetry --project tests run pytest -n 2 tests/e2e_tests -v --tb=short --durations=5
```
agourlay: 4 in parallel is filling the disk too fast

tellet-q: fixed in https://github.com/qdrant/qdrant/pull/7568

tests/e2e_tests/test_data_compatibility.py
tellet-q: Fixed in https://github.com/qdrant/qdrant/pull/7568

tests/e2e_tests/test_data_compatibility.py
```diff
@@ -32,7 +43,15 @@ class TestStorageCompatibility:
 
     @staticmethod
     def _download_compatibility_data(version: str, storage_test_dir: Path):
-        """Download compatibility data for a specific version."""
+        """Download compatibility data for a specific version.
```
agourlay, line 46: @tellet-q this function seems to always download the archive without looking if it is already present on disk.

I thought the test grouping was enabling to reuse the downloaded archive.
How does it work exactly?

tellet-q: Each test runs independently, so it downloads and then removes the files, any other strategy will lead to disk saturation. The grouping ensures that even we run pytest with several parallel workers these particular tests will run on the same worker sequentially.

PR #7569, opened 2025-11-20
Fix use the update runtime for update operations

agourlay:
The following update operations are running on the calling Tokio runtime:
- update_all_local
- update_from_peer
- update_from_client

For the REST API that would be the `actix-rt` and for gRPC that would be the `general` runtime.

We actually have a dedicated runtime for update operations that we should be using instead.

https://github.com/qdrant/qdrant/blob/master/src/common/helpers.rs#L26-L45

This PR changes those 3 call sites to use it which should result in better stability in the other runtimes.

Sorry for the reformating noise.

# Future work

There are more `tokio::task::spawn` usages to audit to check their correctness.

### DIFF
--- a/src/common/helpers.rs
+++ b/src/common/helpers.rs
@@ -30,6 +30,7 @@ pub fn create_update_runtime(max_optimization_threads: usize) -> io::Result<Runt
 
     update_runtime_builder
         .enable_time()
+        .enable_io()
         .worker_threads(num_cpus)
         .thread_name_fn(move || {
             static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);


agourlay:
@timvisee I had to enable the IO component on the update runtime to pass the consensus tests.
https://github.com/qdrant/qdrant/pull/7569/commits/6470b8ef77592dd84de2a43eea18aee7bf562182

src/common/helpers.rs
```diff
@@ -30,6 +30,7 @@ pub fn create_update_runtime(max_optimization_threads: usize) -> io::Result<Runt
 
     update_runtime_builder
         .enable_time()
+        .enable_io()
```
timvisee, line 33: Let's just do [`enable_all()`](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.enable_all) on all three runtimes. Wdyt?

PR #7571, opened 2025-11-20
rename common -> qdrant_common to avoid name with other python packages

generall:
Related to https://github.com/qdrant/qdrant-client/issues/1119

### DIFF
--- a/lib/api/src/grpc/proto/collections.proto
+++ b/lib/api/src/grpc/proto/collections.proto
@@ -4,7 +4,7 @@ package qdrant;
 option csharp_namespace = "Qdrant.Client.Grpc";
 
 import "json_with_int.proto";
-import "common.proto";
+import "qdrant_common.proto";
 
 enum Datatype {
   Default = 0;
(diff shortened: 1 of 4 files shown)


generall:
@Anush008 could you please check that it doesn't break java client?

xzfc:
So such file name turns to be quite _common_, amirite?

---

We have `package qdrant;`, so the idiomatic approach would be to put all files in a `qdrant/` subdirectory, and import it like this:
```proto
import "qdrant/common.proto";
import "qdrant/collections.proto";
```

See how Google organizes it: https://github.com/googleapis/googleapis/blob/ded7ed1e4cce7c165c56a417572cebea9bc1d82c/google/bigtable/v2/bigtable.proto
Also see https://buf.build/docs/reference/protobuf-files-and-packages/#file-paths

generall:
> so the idiomatic approach would be to put all files in a qdrant/ subdirectory, and import it like this:

I am afraid it will unpredictable break some clients.

PR #7573, opened 2025-11-20
Replica should have PartialSnapshot state while recovering

KShivendu:
Ensure that we use `PartialSnapshot` replica state during recovery so read traffic can be routed accordingly.


tests/consensus_tests/test_partial_snapshot.py
```diff
@@ -110,6 +110,21 @@ def test_partial_snapshot_recovery_lock(tmp_path: pathlib.Path, wait: bool):
...
+
+    res = recover_partial_snapshot_from(read_peer, write_peer, wait = False)
+    # Immediately check replica state after starting recovery
+    read_telemetry = get_telemetry_collections(read_peer)
+
+    read_replica_states = set(read_telemetry[0]['shards'][0]['replicate_states'].values())
+    read_replica_partial_snapshot_state = read_telemetry[0]['shards'][0]['partial_snapshot']
+
+    assert read_replica_partial_snapshot_state['is_recovering']
+    assert read_replica_states == {"PartialSnapshot"}
```
timvisee, line 126: `PartialSnapshot` is deprecated, I'd expect `Recovery` instead

PR #7577, opened 2025-11-21
Fix WAL handling on consensus snapshot

timvisee:
Fix how we handle WAL clearing when we apply a consensus snapshot.

The main problem this tries to solve is the following: in [this](https://github.com/qdrant/qdrant/blob/1b6e52554a394c0db54854b73b07fcac073b8930/lib/storage/src/content_manager/consensus_manager.rs#L557-L563) function, we clear the WAL before we persist the new consensus state from the snapshot. If we crash in between, we only clear the WAL while keeping old consensus state. This will cause a crash loop on startup.

To fix the problem we flip the order. We now clear the WAL after persisting the new consensus state. On startup we truncate any extra entries from the WAL if we failed to clear it.

Needs some more testing before we apply this change.

This needs fixes in WAL crate here: <https://github.com/qdrant/wal/pull/99>

Applying <https://github.com/qdrant/wal/pull/99> also fixes WAL CRC issues that pop up like this log line:

```log
2025-11-21T15:23:22.205425Z  WARN wal::segment: CRC mismatch at offset 40: 414236364 != 838619149
```

### Testing

Testing this change is not easy, and I don't see how I can write a nice automated test for it. I tested this PR including <https://github.com/qdrant/wal/pull/99>.

What I did is this:

1. set `QDRANT__CLUSTER__CONSENSUS__COMPACT_WAL_ENTRIES=5` to enable aggressive consensus WAL compaction
2. start cluster of 3 nodes
3. kill last node
4. send `DELETE /cluster/peer/1` at least 5 times to trigger WAL compaction
5.
    a. without this PR: add panic in between these two statements to simulate crash at the right time: https://github.com/qdrant/qdrant/blob/1b6e52554a394c0db54854b73b07fcac073b8930/lib/storage/src/content_manager/consensus_manager.rs#L557-L563
    b. or with this PR: add panic in between these two statements to simulate crash at the right time: https://github.com/qdrant/qdrant/blob/0f274c54a174db554847308e23b5f2271d48d653/lib/storage/src/content_manager/consensus_manager.rs#L581-L593
6. (recompile and) start 3rd node again
7. node now applies consensus snapshot and crashes due to panic inserted in point 5
8. remove panic added in point 5
9. (recompile and) start 3rd node again
10.
    a. before this PR: crash loop!
    b. or after this PR: start and continue fine, and notice a log line like this:
       ```
       2025-11-21T15:31:06.595041Z  WARN storage::content_manager::consensus_manager: Consensus WAL has 34 unapplied entries, truncating from index 0 onwards
       ```

### Tasks
- [x] Review and merge <https://github.com/qdrant/wal/pull/99>
- [x] Bump `wal` dependency: <https://github.com/qdrant/qdrant/pull/7587>
- [x] Validate we don't miss any consensus operations this way, we must re-receive what we truncate

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/consensus/consensus_wal.rs
+++ b/lib/storage/src/content_manager/consensus/consensus_wal.rs
@@ -52,16 +52,6 @@ impl ConsensusOpWal {
         Ok(())
     }
 
-    /// Truncate WAL from the given index
-    ///
-    /// Truncate the WAL from the given WAL index (not Raft index!). The index is inclusive,
-    /// meaning the given index is also removed.
-    pub fn truncate(&mut self, from_index: u64) -> Result<(), StorageError> {
-        self.wal.truncate(from_index)?;
-        self.wal.flush_open_segment()?;
-        Ok(())
-    }
-
     pub fn entry(&self, raft_index: u64) -> raft::Result<RaftEntry> {
         // Raft entries are expected to have index starting from 1
         if raft_index < 1 {
--- a/lib/storage/src/content_manager/consensus_manager.rs
+++ b/lib/storage/src/content_manager/consensus_manager.rs
@@ -823,10 +823,6 @@ impl<C: CollectionContainer> ConsensusManager<C> {
         self.wal.lock().clear()
     }
 
-    pub fn truncate(&self, from_index: u64) -> Result<(), StorageError> {
-        self.wal.lock().truncate(from_index)
-    }
-
     pub fn compact_wal(&self, min_entries_to_compact: u64) -> Result<bool, StorageError> {
         if min_entries_to_compact == 0 {
             return Ok(false);


lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -111,10 +111,34 @@ impl<C: CollectionContainer> ConsensusManager<C> {
...
+
+            let extra_entries = last_wal_commit.saturating_sub(last_wal_commit_applied);
+            if extra_entries > 0 {
+                log::warn!(
+                    "Consensus WAL has {extra_entries} unapplied entries, truncating from index {last_wal_commit_applied} onwards"
+                );
+                wal.truncate(last_wal_commit_applied)
+                    .expect("Failed to truncate WAL on startup");
+            }
+        }
```
ffuugoo: I'd say it's wrong to rely on `hard_state.commit`. We need to rely on `latest_snapshot_metadata.index`.

ffuugoo: `latest_snapshot_metadata` is only updated during snapshot. `commit` is updated during normal consensus operation. If we, e.g., have a WAL flushing issue in the future (like, unrelated/without consensus snapshots) and we lost last entries from WAL, we would happily truncate WAL and move on, which is wrong.

ffuugoo: We need to check last WAL against `latest_snapshot_metadata.index` and truncate WAL if there's a "hole" between WAL and `latest_snapshot_metadata`.

E.g., last entry in the WAL is `index: 100`, we applied snapshot with `commit: 200` and crashed before clean. We check `if wal.last_index() < latest_snapshot_metadata.index { wal.clear() }`.

timvisee: I agree, the current logic is not correct.

We had a call to discuss how best to change this.

It now works like this:

https://github.com/qdrant/qdrant/blob/7a1edfc97e0694efb44f447a89201a7c75d7d6ed/lib/storage/src/content_manager/consensus_manager.rs#L117-L134

Meaning: if Raft index and snapshot index are the same our last operation was a snapshot. Then, if our WAL still has old entries we fully clear it. That matches up with the WAL clear behavior in the happy path.

lib/storage/src/content_manager/consensus/consensus_wal.rs
```diff
@@ -46,7 +46,19 @@ impl ConsensusOpWal {
...
+        Ok(())
+    }
+
+    /// Truncate WAL from the given index
+    ///
+    /// Truncate the WAL from the given WAL index (not Raft index!). The index is inclusive,
+    /// meaning the given index is also removed.
+    pub fn truncate(&mut self, from_index: u64) -> Result<(), StorageError> {
+        self.wal.truncate(from_index)?;
+        self.wal.flush_open_segment()?;
```
ffuugoo: Can we remove this?

lib/storage/src/content_manager/consensus_manager.rs
```diff
@@ -796,6 +823,10 @@ impl<C: CollectionContainer> ConsensusManager<C> {
         self.wal.lock().clear()
     }
 
+    pub fn truncate(&self, from_index: u64) -> Result<(), StorageError> {
+        self.wal.lock().truncate(from_index)
+    }
```
ffuugoo: And this. I don't think we use these anymore.

lib/storage/src/content_manager/consensus/persistent.rs
```diff
@@ -99,6 +99,13 @@ impl Persistent {
         *peer_metadata_by_id.write() = metadata_by_id;
         *cluster_metadata = new_cluster_metadata;
 
+        // Last Raft commit and last snapshot index must be equal and persisted in one operation
+        // Our `ConsensusManager::new` function relies on this for reconciling WAL clears
+        debug_assert_eq!(
+            state.hard_state.commit, latest_snapshot_meta.index,
+            "applied Raft commit and last snapshot index must be equal",
+        );
```
ffuugoo, line 107: Good idea to put an assert here. 🫡🤝

lib/storage/src/content_manager/consensus/consensus_wal.rs
timvisee: I'll remove them. Thanks

PR #7579, opened 2025-11-21
Add timeout to cancel metrics and telemetry calls

agourlay:
This PR adds a new timeout support for the telemetry endpoint and the metrics endpoint.

The goal is to prevent potentially long running computations to conflict with other workloads.

This new timeout is now always present internally with a default value of 60 seconds.
It can be changed to any positive value via the API to handle extreme cases.

The telemetry/metrics workload cancellation works at different level to juggle with the sync. operations.

1. an async `tokio::timeout` on the top level future
2. a propagation into blocking tasks via a cancellation token on drop
3. an upper bound on the maximum time to wait to acquire the underlying sync. locks

Before this PR, the telemetry/metrics API were not fallible. 
They would always return successfully at the cost of returning partial results.

Both APIs are now fallible to be able to bubble up timeouts internally.

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -252,7 +252,8 @@
             "required": false,
             "schema": {
               "type": "integer",
-              "minimum": 1
+              "minimum": 1,
+              "default": 60
             }
           }
         ],
@@ -341,7 +342,8 @@
             "required": false,
             "schema": {
               "type": "integer",
-              "minimum": 1
+              "minimum": 1,
+              "default": 60
             }
           }
         ],
--- a/openapi/openapi-service.ytt.yaml
+++ b/openapi/openapi-service.ytt.yaml
@@ -46,6 +46,7 @@ paths:
           schema:
             type: integer
             minimum: 1
+            default: 60
       responses: #@ response(reference("TelemetryData"))
 
   /metrics:
@@ -69,6 +70,7 @@ paths:
           schema:
             type: integer
             minimum: 1
+            default: 60
       responses:
         "200":
           description: Metrics data in Prometheus format


src/actix/api/service_api.rs
```diff
@@ -23,15 +26,23 @@ use crate::common::telemetry::TelemetryCollector;
 use crate::settings::ServiceConfig;
 use crate::tracing;
 
-#[derive(Deserialize, Serialize, JsonSchema)]
+#[derive(Deserialize, Serialize, JsonSchema, Validate)]
 pub struct TelemetryParam {
     pub anonymize: Option<bool>,
     pub details_level: Option<usize>,
+    #[validate(range(min = 1, max = 10))]
```
timvisee: Let's set the maximum at something more unreasonable, like 1 hour.

I've seen valid cases where fetching telemetry took 12 seconds. It was on a (very!) busy cluster with thousands of collections, shards and segments.

lib/collection/src/shards/replica_set/telemetry.rs
```diff
@@ -1,23 +1,34 @@
...
 use crate::shards::replica_set::ShardReplicaSet;
 use crate::shards::telemetry::{PartialSnapshotTelemetry, ReplicaSetTelemetry};
 
 impl ShardReplicaSet {
-    pub(crate) async fn get_telemetry_data(&self, detail: TelemetryDetail) -> ReplicaSetTelemetry {
+    pub(crate) async fn get_telemetry_data(
+        &self,
+        detail: TelemetryDetail,
+        timeout: Duration,
+        is_stopped_guard: &StoppingGuard,
```
timvisee, line 17: I don't think it's correct to use a stopped guard here.

Since this is an async function, I'd argue the right way of aborting it is to drop it. I only expect to see an actual stop guard/structure where we switch to sync code. We have some utilities in lib/common/cancel to help with this, specifically to bridge async/sync cancellation.

I'm fine with adjusting this behavior in a separate PR.

lib/storage/src/content_manager/toc/telemetry.rs
```diff
@@ -41,12 +44,21 @@ impl TableOfContent {
...
-    ) -> TocTelemetryData {
-        let mut collection_telemetry = Vec::new();
+        timeout: Duration,
+        is_stopped_guard: &StoppingGuard,
+    ) -> CollectionResult<TocTelemetryData> {
         let all_collections = self.all_collections_access(access).await;
+        let mut collection_telemetry = Vec::new();
         for collection_pass in &all_collections {
+            if is_stopped_guard.is_stopped() {
+                break;
```
timvisee, line 54: Shouldn't we return with an empty state instead? Otherwise we still collect snapshot statistics.

timvisee (commented):
Nice! I like that this handles both cancellation flavors - aborting the request and a timeout.

I think we should adjust how we bridge cancellation between sync/async code, but I'd be happy with settling on that in a separate PR.

lib/storage/src/content_manager/toc/telemetry.rs
agourlay: I went with this solution because it is not trivial to implement Default for `TocTelemetryData`.

There are a handful of inner nested types which would require custom implementation because they use `NonZero` field for instance.

I decided to avoid the noise and just break the loop at the extra cost of performing the snapshot statistics.

timvisee: Got it! Those statistics are supposed to be cheap anyway.

timvisee (approved):

lib/collection/src/shards/replica_set/telemetry.rs
agourlay: I think you are correct :+1: 

I will fix this in a follow up PR to not hold the release.

src/common/telemetry.rs
```diff
@@ -21,6 +25,8 @@ use crate::common::telemetry_ops::requests_telemetry::{
 };
 use crate::settings::Settings;
 
+const DEFAULT_TELEMETRY_TIMEOUT: Duration = Duration::from_secs(60);
```
xzfc, line 28: I'm late to the party, but here's my nit:

Add cross-reference comments so we won't forget update both places if we change this constant.

E.g.:
```rust
// Keep in sync with openapi/openapi-service.ytt.yaml
const DEFAULT_TELEMETRY_TIMEOUT: Duration = …;
```

openapi/openapi-service.ytt.yaml:
```yaml
  /metrics:
    get:
      …
      parameters:
        …
        - name: timeout
          in: query
          description: "Timeout for this request"
          required: false
          schema:
            type: integer
            minimum: 1
            default: 60 # Keep in sync with DEFAULT_TELEMETRY_TIMEOUT
```
(worded in a way that `rg DEFAULT_TELEMETRY_TIMEOUT` yields both places)

Perhaps the same for the `minimum: 1`/`#[validate(range(min = 1))]`.

PR #7588, opened 2025-11-24
Timeout on point retrieve when acquiring segment holder

agourlay:
This PR adds a timeout on `retrieve_blocking` to make sure it does not wait indefinitely while trying to acquire the segment holder lock.

A unit test has been added showing that the timeout kicks if a writer holds onto the lock for a long period of time.

The function `retrieve_blocking` is used mostly on the read patch when searching/querying or scrolling points.

I decided to hard-code a few timeouts for convenience:
- in tests
- in the edge crate

TBH I do not like the approach for the edge crate, happy to hear better suggestions.

### DIFF
--- a/lib/edge/src/lib.rs
+++ b/lib/edge/src/lib.rs
@@ -183,4 +183,5 @@ fn default_wal_options() -> WalOptions {
     }
 }
 
-pub const DEFAULT_EDGE_TIMEOUT: Duration = Duration::from_hours(1);
+// Default timeout of 1h used as a placeholder in Edge
+pub(crate) const DEFAULT_EDGE_TIMEOUT: Duration = Duration::from_secs(3600);


lib/collection/src/collection_manager/segments_searcher.rs
```diff
@@ -891,13 +894,33 @@ mod tests {
...
             &true.into(),
+            Duration::from_secs(1),
             &AtomicBool::new(false),
             HwMeasurementAcc::new(),
         )
         .unwrap();
         assert_eq!(records.len(), 3);
     }
 
+    #[test]
```
agourlay, line 904: test 👇

lib/segment/src/common/operation_error.rs
```diff
@@ -48,6 +48,8 @@ pub enum OperationError {
     OutOfMemory { description: String, free: u64 },
     #[error("Operation cancelled: {description}")]
     Cancelled { description: String },
+    #[error("Timeout error: {description}")]
+    Timeout { description: String },
```
agourlay, line 52: a new variant was required to propagate the error

lib/collection/src/collection_manager/fixtures.rs
```diff
@@ -23,6 +24,8 @@ use crate::config::CollectionParams;
 use crate::operations::types::VectorsConfig;
 use crate::operations::vector_params_builder::VectorParamsBuilder;
 
+pub const TEST_TIMEOUT: Duration = Duration::from_hours(1);
```
timvisee: Wow!

Since this is in tests, shall we do a much lower timeout (like 10 seconds?) so we catch any operations that are stuck for some time.

lib/edge/src/query.rs
```diff
@@ -403,6 +404,7 @@ impl Shard {
             &point_ids,
             &WithPayload::from(with_payload),
             &with_vector,
+            Duration::from_hours(1),
```
timvisee: Move this to a constant (as well as the other usages in edge)?

lib/shard/src/retrieve/retrieve_blocking.rs
```diff
@@ -24,60 +26,65 @@ pub fn retrieve_blocking(
...
-        .read_points(points, is_stopped, |id, segment| {
-            let version = segment.point_version(id).ok_or_else(|| {
-                OperationError::service_error(format!("No version for point {id}"))
-            })?;
+    let Some(segments_guard) = segments.try_read_for(timeout) else {
+        return Err(OperationError::timeout(
+            timeout.as_secs() as usize,
+            "retrieve points",
+        ));
+    };
```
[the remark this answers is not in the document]
agourlay: good point I will take care of all instances of that problem in a different PR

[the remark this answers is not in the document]
agourlay: Yeah sure go ahead and do that.

Please list precisely in the issue all the instances of this in the code base to make my life easier.

Also assign the issue to myself so I don't forget.

Do not disappoint me.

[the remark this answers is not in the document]
agourlay: Good rabbit! :clap:

PR #7598, opened 2025-11-24
fix/inference timeout error

dancixx:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/common/inference/service.rs
+++ b/src/common/inference/service.rs
@@ -76,6 +76,11 @@ pub struct InferenceService {
 
 static INFERENCE_SERVICE: RwLock<Option<Arc<InferenceService>>> = RwLock::new(None);
 
+/// We assume that the inference provider will handle timeouts itself, if
+/// not provided by the user or configured. But we need ensurance, that we don't
+/// wait forever for a response.
+static DEFAULT_INFERENCE_TIMEOUT_SECS: u64 = 10 * 60; // 10 minutes
+
 impl InferenceService {
     pub fn new(config: Option<InferenceConfig>) -> Self {
         let config = config.unwrap_or_default();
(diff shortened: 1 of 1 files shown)


src/common/inference/service.rs
```diff
@@ -79,11 +79,23 @@ static INFERENCE_SERVICE: RwLock<Option<Arc<InferenceService>>> = RwLock::new(No
...
+            token: _,
+        } = &config;
+
+        let client_builder = Client::builder();
+
+        let client_builder = if let Some(timeout_secs) = timeout {
+            client_builder.timeout(Duration::from_secs(*timeout_secs))
+        } else {
+            client_builder
+        };
```
timvisee: Disclaimer: I only skimmed the PR.

Am I seeing correctly that it is possible for us to create a client that has no timeout at all? If so, I strongly vote against that. We should always provide _some_ timeout.

Relying in a timeout in the system we talk to (inference proxy) is not good enough.

timvisee: In a short discussion we agreed to default to a timeout of 10 minutes. That should be more than enough for regular requests, and makes sure we always have at least _some_ timeout to clean up requests that get stuck.

src/common/inference/service.rs
```diff
@@ -181,13 +198,21 @@ impl InferenceService {
...
         };
 
-        let response = self.client.post(url).json(&request).send().await;
+        let request = self.client.post(url);
+
+        let request = if let Some(timeout) = timeout {
+            request.timeout(timeout)
+        } else {
+            request
+        };
```
timvisee, line 213: We must also set default timeout here

generall: this request uses pre-configured service, it alredy have timeout by default

timvisee: I see, missed that. Thanks!

PR #7599, opened 2025-11-24
[cluster telemetry] internal service types

coszio:
As a first step towards aggregating telemetry from all peers, we need to be able to communicate telemetry with other peers. This can only be done through the internal service.

This PR adds internal service types for mapping current telemetry into internal service. For now, we include only info about the peer, resharding, and shard transfers.

### DIFF
--- a/lib/api/src/grpc/proto/telemetry_internal.proto
+++ b/lib/api/src/grpc/proto/telemetry_internal.proto
@@ -27,7 +27,7 @@ message CollectionTelemetry {
     reserved 4; // shards
     repeated ShardTransferTelemetry transfers = 5; // Shard transfers in progress
     repeated ReshardingTelemetry resharding = 6; // Resharding(s) in progress
-    map<uint32, ShardCleanTelemetry> shard_clean_tasks = 7; // Tasks that clean points after completing a resharding sequence
+    map<uint32, ShardCleanStatusTelemetry> shard_clean_tasks = 7; // Tasks that clean points after completing a resharding sequence
 }
 
 message ShardTransferTelemetry {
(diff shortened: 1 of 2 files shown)


lib/api/src/grpc/proto/telemetry_internal.proto
```diff
@@ -0,0 +1,111 @@
...
+
+message CollectionTelemetry {
+    string id = 1; // Name of the collection
+    reserved 2; // init_time_ms
+    reserved 3; // config
+    reserved 4; // shards
+    repeated ShardTransferInfo transfers = 5; // Shard transfers in progress
+    repeated ReshardingInfo resharding = 6; // Resharding(s) in progress
+    map<uint32, ShardCleanTelemetry> shard_clean_tasks = 7; // Tasks that clean points after completing a resharding sequence
+}
```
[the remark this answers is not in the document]
coszio: It did compile, but it was using the definition from `collections.proto`. I indeed meant to use the `*Telemetry` definitions. Good bot.

lib/api/src/grpc/proto/telemetry_internal.proto
```diff
@@ -0,0 +1,140 @@
...
+
+message GetPeerTelemetryResponse {
+    reserved 1; // id
+    reserved 2; // app
+    map<string, CollectionTelemetry> collections = 3; // Mapping from collection name to its telemetry
+    ClusterTelemetry cluster = 4; // Telemetry about the cluster and peers
+    reserved 5; // requests
+    reserved 6; // memory
+    reserved 7; // hardware
+}
```
timvisee, line 21: So this basically follows the same structure as our existing telemetry, but it is very limited.

Is there a strong argument for doing it this way, versus implementing a whole new kind of telemetry with it's own fields?

coszio: Yes, and it is made to replicate it closely. 

We already have the information we need from existing telemetry, so we can reuse the collectors we already have for internal service.

PR #7601, opened 2025-11-25
Document storage compatibility test for v1.15.0

agourlay:
The archive for `v1.15.0` cannot be restored because it fails with a debug assertion.

```
ERROR qdrant::startup: Panic occurred in file lib/common/common/src/mmap_hashmap.rs at line 474: Error reading u128 from mmap: The conversion failed because the address of the source is not a multiple of the alignment of the destination type.
```

The original bug has been fixed in later version https://github.com/qdrant/qdrant/pull/6916

### DIFF
--- a/tests/e2e_tests/test_data_compatibility.py
+++ b/tests/e2e_tests/test_data_compatibility.py
@@ -23,7 +23,7 @@ class TestStorageCompatibility:
         "v1.15.3",
         "v1.15.2",
         "v1.15.1",
-        "v1.15.0",
+        # "v1.15.0", the archive triggers debug_assertion for the UUID payload index https://github.com/qdrant/qdrant/pull/6916
     ]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -263,7 +263,7 @@ jobs:
       - name: Run e2e tests
         run: poetry --project tests run pytest -n 4 --dist=loadgroup tests/e2e_tests -v --tb=short --durations=5
         shell: bash
-        timeout-minutes: 10
+        timeout-minutes: 12
```
agourlay: opportunistic bump as I have seen a handful of timeouts on CI

tellet-q: Fixed in https://github.com/qdrant/qdrant/pull/7611

PR #7602, opened 2025-11-25
Scalar quantization encoding parameter

IvanPleshkov:
This PR is a preparation step before new SQ methods. 
SQ config has a `type` parameter
https://github.com/qdrant/qdrant/blob/master/lib/segment/src/types.rs#L745

This parameter was introduced with SQ and has only one variant `int8`
https://github.com/qdrant/qdrant/blob/master/lib/segment/src/types.rs#L735

Because in dev we don't have any other SQ encoding methods, `type` was ignored. But now we want to introduce new SQ encodings. And this PR just propagates `type` parameter into SQ implementation

### DIFF
--- a/lib/quantization/src/encoded_vectors_u8.rs
+++ b/lib/quantization/src/encoded_vectors_u8.rs
@@ -21,7 +21,7 @@ pub const ALIGNMENT: usize = 16;
 
 #[derive(Clone, PartialEq, Debug)]
 pub enum ScalarQuantizationMethod {
-    Uint8,
+    Int8,
     // Future methods can be added here
 }
 
@@ -61,7 +61,7 @@ impl<TStorage: EncodedStorage> EncodedVectorsU8<TStorage> {
         meta_path: Option<&Path>,
         stopped: &AtomicBool,
     ) -> Result<Self, EncodingError> {
-        assert_eq!(method, ScalarQuantizationMethod::Uint8);
+        assert_eq!(method, ScalarQuantizationMethod::Int8);
         let actual_dim = Self::get_actual_dim(vector_parameters);
 
         if count == 0 {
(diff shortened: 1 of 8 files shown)


lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -19,6 +19,13 @@ use crate::quantile::{find_min_max_from_iter, find_quantile_interval};
 
 pub const ALIGNMENT: usize = 16;
 
+#[derive(Clone, Default, PartialEq, Debug)]
```
monoid: It seems `Default` is not really used now.  But if you have a good reason to keep it, it is OK.  I think it can be dangerous.

lib/segment/src/vector_storage/quantized/quantized_vectors.rs
```diff
@@ -1808,6 +1815,12 @@ impl QuantizedVectors {
         }
     }
 
+    fn convert_scalar_encoding(encoding: ScalarType) -> ScalarQuantizationMethod {
+        match encoding {
+            ScalarType::Int8 => ScalarQuantizationMethod::Uint8,
```
monoid: Is (will be) it always the case?

IvanPleshkov: In the next steps, I gonna add new variants because the goal is to introduce new SQ types

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: Removed it, if needed - will add later

lib/segment/src/vector_storage/quantized/quantized_vectors.rs
monoid: I mean if the function only gets a single argument, it will always return the same value if the argument is the same.

monoid (approved):

PR #7607, opened 2025-11-25
Improve timeout logic for Telemetry

agourlay:
Follow up on https://github.com/qdrant/qdrant/pull/7579

- Remove unnecessary stop guard https://github.com/qdrant/qdrant/pull/7579#discussion_r2559460381
- Add new timeout on `get_optimization_status` to handle case with lower telemetry details
- Add test for different level of telemetry details
- Fix wrong usage of the `StoppingGuard` where it got moved into the task (https://github.com/qdrant/qdrant/pull/7607/commits/e063af11f8a5662232974c316a4a4ce736bf1f0a)

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -787,13 +787,17 @@ impl Collection {
         &self,
         timeout: Duration,
     ) -> CollectionResult<CollectionsAggregatedTelemetry> {
+        let start = std::time::Instant::now();
         let shards_holder = self.shards_holder.read().await;
 
         let mut shard_optimization_statuses = Vec::new();
         let mut vectors = 0;
 
         for shard in shards_holder.all_shards() {
-            let shard_optimization_status = match shard.get_optimization_status(timeout).await {
+            let shard_optimization_status = match shard
+                .get_optimization_status(timeout.saturating_sub(start.elapsed()))
+                .await
+            {
                 None => OptimizersStatus::Ok,
                 Some(status) => status?,
             };


lib/collection/src/shards/local_shard/telemetry.rs
```diff
@@ -22,15 +22,14 @@ impl LocalShard {
...
     ) -> CollectionResult<LocalShardTelemetry> {
+        let start = std::time::Instant::now();
         let segments = self.segments.clone();
-
         let segments_data = if detail.level < DetailsLevel::Level4 {
             Ok((vec![], HashMap::default()))
         } else {
             let locked_collection_config = self.collection_config.clone();
-            let is_stopped_guard = is_stopped_guard.clone();
+            let is_stopped_guard = StoppingGuard::new();
```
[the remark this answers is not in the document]
agourlay: the is_stopped_guard is created outside of the blocking task and will kicked in when the outer future is dropped to shortcut the loop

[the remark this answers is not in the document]
agourlay: No, I think you're absolutely right :)

The guard should not be moved into the closure otherwise it can't be dropped from outside.

lib/storage/src/content_manager/toc/telemetry.rs
```diff
@@ -45,20 +44,16 @@ impl TableOfContent {
         detail: TelemetryDetail,
         access: &Access,
         timeout: Duration,
-        is_stopped_guard: &StoppingGuard,
+        is_stopped: &AtomicBool,
```
timvisee: This is an async function so we can drop it instead.

lib/collection/src/collection/mod.rs
```diff
@@ -783,17 +783,20 @@ impl Collection {
...
 
         let mut shard_optimization_statuses = Vec::new();
         let mut vectors = 0;
 
         for shard in shards_holder.all_shards() {
-            let shard_optimization_status = shard
-                .get_optimization_status()
-                .await
-                .unwrap_or(OptimizersStatus::Ok);
+            let shard_optimization_status = match shard.get_optimization_status(timeout).await {
```
timvisee: Because this loops, maybe timeout should be `timeout - start.elapsed()`

lib/collection/src/tests/shard_telemetry.rs
```diff
@@ -0,0 +1,71 @@
...
+            .await
+            .unwrap();
+        assert_eq!(telemetry.num_points, Some(0));
+
+        // test that it timeouts if the segment_holder lock can't be acquired
+        let write_segment_holder_guard = shard.segments().write();
+        let telemetry = shard
+            .get_telemetry_data(details, Duration::from_millis(10))
+            .await;
+        assert!(matches!(telemetry, Err(CollectionError::Timeout { .. })));
```
timvisee, line 62: Nice 👌

Issue #7610, opened 2025-11-25
Qdrant startup panic after upgrade to 1.16.1 from 1.16.0 in docker


avvertix:
Starting version 1.16.1 over 1.16.0 via Docker results in a startup panic. The data is stored in a mounted volume residing on Windows, while the container uses the official Qdrant Docker image.

## Current Behavior

Crash at startup

## Steps to Reproduce

1. Start a container using v1.16.0 with a mounted volume as data folder
2. Create a collection
3. Add some data
4. Stop the instance
5. Change the docker image version to v1.16.1
6. Start the instance

Docker compose file

```yaml
services: 
    qdrant:
        image: qdrant/qdrant:v1.16.1
        ports:
        - 6333:6333
        - 6334:6334
        expose:
            - 6333
            - 6334
        volumes:
            - ./qdrant_data:/qdrant/storage
```

Collection `config.json` as found in storage

```json
{"params":{"vectors":{"fast-bge-small-en-v1.5":{"size":384,"distance":"Cosine"}},"shard_number":1,"replication_factor":1,"write_consistency_factor":1,"on_disk_payload":true},"hnsw_config":{"m":16,"ef_construct":100,"full_scan_threshold":10000,"max_indexing_threads":0,"on_disk":false},"optimizer_config":{"deleted_threshold":0.2,"vacuum_min_vector_number":1000,"default_segment_number":0,"max_segment_size":null,"memmap_threshold":null,"indexing_threshold":10000,"flush_interval_sec":5,"max_optimization_threads":null},"wal_config":{"wal_capacity_mb":32,"wal_segments_ahead":0,"wal_retain_closed":1},"quantization_config":null,"uuid":null}
```

`payload_index.json`

```json
{"schema":{"document":{"type":"text","tokenizer":"word","min_token_len":2,"max_token_len":15,"lowercase":true}}}
```

`shard_key_mapping.json`

```json
[]
```

`version.info`

```
0.4.2
```

## Expected Behavior

Qdrant starts

## Context (Environment)

Here the startup log

```
           _                 _    
  __ _  __| |_ __ __ _ _ __ | |_  
 / _` |/ _` | '__/ _` | '_ \| __|
| (_| | (_| | | | (_| | | | | |_
 \__, |\__,_|_|  \__,_|_| |_|\__|
    |_|

Version: 1.16.1, build: e293ab1f
Access web UI at http://localhost:6333/dashboard

[...]

Issue #6676, opened 2025-04-25
Vectors set to zero on docker container removal - Macos


rayenebech:
I am building my Qdrant service on macos using docker image as follows: 
```bash
qdrant:
    image: qdrant/qdrant
    restart: unless-stopped
    ports:
      - "6335:6333"
    networks:
      - app-network
    volumes:
      - ./backend/data-source/qdrant-storage:/qdrant/storage
 ```

And I am using LlamaIndex Client to create the index

```
class QdrantIndexer:
    def __init__(self, collection_name=None):
        """Initialize the Qdrant vector database."""
        Settings.embed_model = FastEmbedEmbedding(model_name = os.getenv("EMBEDDER"), cache_dir= os.path.join(PARENT_DIR, "model_checkpoints"))
        self.qdrant_client = QdrantClient(os.getenv("QDRANT_HOST", "http://qdrant:6335"))
        self.collection_name = os.getenv("QDRANT_COLLECTION_NAME", "legal") if collection_name is None else collection_name
        self.vector_store = QdrantVectorStore(
            client=self.qdrant_client,
            collection_name=self.collection_name,
        )
        self.storage_context = StorageContext.from_defaults(vector_store=self.vector_store)

    def create_index(self, chunks: List[Document]):
        """Store processed chunks in Qdrant as vector embeddings."""
        self.index = VectorStoreIndex.from_documents(documents=chunks, storage_context=self.storage_context)
```

When I index any file, I can see that the vector embeddings are correctly computed and saved to the collection. However, If I delete the container and build a new one using the same storage, then all the vector values are set to a zero (A vector of zero values).

This happens only on Macos Silicon. On Ubuntu it works normally and vector are restored without issues.
Using `Python3.10` and `llama-index-vector-stores-qdrant==0.4.3`

**Comment by Catolampus** [comment]
I encountered the same problem. Loading batches using qdrant inside docker and bind mount (Windows) leads to zeroing of previously loaded correct vectors.

**Comment by bechhray** [comment]
As suggested by @dwgebler , Naming the volumes this way solved the issue for me:

```
services:
  qdrant:
    image: qdrant/qdrant:v1.14.0
    container_name: qdrant_server
    restart: unless-stopped
    ports:
      - "6333:6333"
    networks:
      - app-network
    volumes:
      - qdrant-storage:/qdrant/storage 
      
volumes:
  qdrant-storage:
    name: qdrant-storage

```

Did you try it this way @Catolampus ?

[...]

PR #7614, opened 2025-11-26
Add runtime flag to disable opening multiple mmaps on the same file

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/7610>, <https://github.com/qdrant/qdrant/issues/6676>

Add the `QDRANT_NO_MULTI_MMAP` environment variable to disable multi-mmap at runtime. It disables opening multiple memory maps to the same piece of data on disk in all of Qdrant.

This pattern of opening multiple memory maps is broken in some environment, corrupting storage on disk. That's why you can now disable it on demand.

An example of such environment is Docker on Windows with a storage mount to the Windows host. This flag effectively brings us better compatibility.

Usage example:

```bash
# run without multiple mmaps for better compatibility
QDRANT_NO_MULTI_MMAP=1 ./qdrant
```

Once memory maps are first used a warning pops up if this feature is disabled:

```bash
# create collection with mmaps
$ bfb -n 100

# Qdrant logs now show:
2025-11-26T10:46:22.885080Z  WARN memory::mmap_ops: QDRANT_NO_MULTI_MMAP is set, you may see reduced performance
```

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/common/memory/src/mmap_ops.rs
+++ b/lib/common/memory/src/mmap_ops.rs
@@ -1,5 +1,6 @@
 use std::mem::{align_of, size_of};
 use std::path::Path;
+use std::sync::LazyLock;
 use std::{io, mem, ptr};
 
 use fs_err as fs;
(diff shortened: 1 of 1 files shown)


timvisee:
> Is it possible to add this as part of settings? I figure it can be more cleanly deployed if this setting is part of the config file.

In this PR I didn't because I pitched this as hidden setting.

Though I have no objection in including it in `config.yaml`, albeit a bit tricky to implement since I'm using a static variable.

Are you okay with changing this in a separate PR?

timvisee:
To clarify, I definitely do **not** recommend to deploy with this setting. That's why I kept it hidden in the first place.

It shows that the IO behavior on Docker/WSL is broken. There may definitely be more dragons we're not aware of yet, breaking production workloads in unexpected ways.

These changes were only intended to not immediately break naive deployments on this environment.

coszio (approved):
@timvisee Thanks for elaborating

Issue #4002, opened 2024-04-10
Add request List shard_keys


ava57r:
Hello.
v1.7 has user-defined-sharding
API has 2 endpoints

- "Create shard key"
- "Delete shard key"

but it haven't endpoint

"List shard keys"

Could You discuss about adding this endpoint?

generall:
partially, this functionality is available in `GET /collection/{name}/cluster`. Those are no unique, though

I would agree that it might be a good idea to have this API

ava57r:
I am using `GET /collection/{name}/cluster` now.

PR #7615, opened 2025-11-26
Add list shard keys API

amiremohamadi:
related to #4002

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -16007,15 +16007,14 @@
       },
       "ShardKeysResponse": {
         "type": "object",
-        "required": [
-          "shard_keys"
-        ],
         "properties": {
           "shard_keys": {
+            "description": "The existing shard keys. Only available when sharding method is `custom`",
             "type": "array",
             "items": {
               "$ref": "#/components/schemas/ShardKeyDescription"
-            }
+            },
+            "nullable": true
           }
         }
       },
--- a/lib/api/src/grpc/conversions.rs
+++ b/lib/api/src/grpc/conversions.rs
@@ -134,6 +134,7 @@ impl From<(Instant, ShardKeysResponse)> for ListShardKeysResponse {
         let ShardKeysResponse { shard_keys } = response;
         let shard_keys = shard_keys
             .into_iter()
+            .flatten()
             .map(|key_desc| {
                 let key = Some(convert_shard_key_to_grpc(key_desc.key));
                 ShardKeyDescription { key }
--- a/lib/api/src/rest/models.rs
+++ b/lib/api/src/rest/models.rs
@@ -171,5 +171,7 @@ pub struct ShardKeyDescription {
 #[derive(Debug, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub struct ShardKeysResponse {
-    pub shard_keys: Vec<ShardKeyDescription>,
+    /// The existing shard keys. Only available when sharding method is `custom`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub shard_keys: Option<Vec<ShardKeyDescription>>,
 }
--- a/src/common/collections.rs
+++ b/src/common/collections.rs
@@ -100,11 +100,16 @@ pub async fn do_get_collection_shard_keys(
     let collection = toc.get_collection(&collection_pass).await?;
 
     let state = collection.state().await;
-    let shard_keys = state
-        .shards_key_mapping
-        .iter_shard_keys()
-        .map(|k| ShardKeyDescription { key: k.clone() })
-        .collect();
+    let shard_keys = match state.config.params.sharding_method.unwrap_or_default() {
+        ShardingMethod::Auto => None,
+        ShardingMethod::Custom => Some(
+            state
+                .shards_key_mapping
+                .iter_shard_keys()
+                .map(|k| ShardKeyDescription { key: k.clone() })
+                .collect(),
+        ),
+    };
 
     Ok(ShardKeysResponse { shard_keys })
 }


amiremohamadi:
quick question! do these changes  require `openapi` tests as well? we didn't have separate tests for shards, so i assume i should create a new file for it? (E.g. `tests/openapi/test_shard_key.py`)

cc: @coszio

timvisee:
> quick question! do these changes require `openapi` tests as well? we didn't have separate tests for shards, so i assume i should create a new file for it? (E.g. `tests/openapi/test_shard_key.py`)

Yes. Please feel free to create a new test file there.

src/tonic/api/collections_api.rs
```diff
@@ -247,6 +248,30 @@ impl Collections for CollectionsService {
...
+
+        // Nothing to verify here.
+        let pass = new_unchecked_verification_pass();
+
+        let shard_keys: CollectionShardKeys = do_get_collection_shard_keys(
+            self.dispatcher.toc(&access, &pass),
+            access,
+            request.into_inner().collection_name.as_str(),
+        )
+        .await?;
```
timvisee: Shard keys are only used when custom sharding is enabled.

I wonder if it makes sense to just return an error if custom sharding is not enabled.

lib/collection/src/operations/conversions.rs
```diff
@@ -1435,6 +1435,24 @@ impl From<ReshardingDirection> for api::grpc::qdrant::ReshardingDirection {
...
+                    ShardKey::Keyword(k) => {
+                        api::grpc::qdrant::shard_key::Key::Keyword(k.to_string())
+                    }
+                    ShardKey::Number(n) => api::grpc::qdrant::shard_key::Key::Number(n),
+                };
+                api::grpc::qdrant::ShardKey { key: Some(key) }
+            })
+            .collect()
+    }
+}
```
coszio: nit: There should be a direct conversion from `segment::types::ShardKey` to `api::grpc::qdrant::ShardKey`, and then use it here.

lib/collection/src/shards/shard_holder/shard_mapping.rs
```diff
@@ -66,6 +66,10 @@ impl ShardKeyMapping {
             .flat_map(|shard_ids| shard_ids.iter().copied())
     }
 
+    pub fn iter_shard_keys(&self) -> impl Iterator<Item = &ShardKey> {
```
coszio, line 70: Function name is clear, but just for consistency
```suggestion
    /// Iterate over all shard keys
    pub fn iter_shard_keys(&self) -> impl Iterator<Item = &ShardKey> {
```

src/actix/api/shards_api.rs
```diff
@@ -1,19 +1,35 @@
...
 
-// ToDo: introduce API for listing shard keys
+#[get("/collections/{name}/shards")]
+async fn list_shard_keys(
+    dispatcher: web::Data<Dispatcher>,
+    collection: Path<CollectionPath>,
+    ActixAccess(access): ActixAccess,
+) -> impl Responder {
+    // No request to verify
+    let pass = new_unchecked_verification_pass();
```
coszio: This should verify that the `access` is at least `read` globally or for this collection

src/actix/api/shards_api.rs
```diff
@@ -1,19 +1,35 @@
...
+use crate::common::collections::{do_get_collection_shard_keys, do_update_collection_cluster};
 
-// ToDo: introduce API for listing shard keys
+#[get("/collections/{name}/shards")]
+async fn list_shard_keys(
+    dispatcher: web::Data<Dispatcher>,
+    collection: Path<CollectionPath>,
+    ActixAccess(access): ActixAccess,
+) -> impl Responder {
+    // No request to verify
```
coszio: I confused this with `CollectionPass` on this review 😆. A bit clearer:
```suggestion
    // No strict-mode checks to verify
```

timvisee:
Having quickly discussed this internally. Lets use a dedicated type for each item in the list we return. That allows us to extend it in the future.

I expect this to be similar to the collections listing (`GET /collections`). It doesn't return a list of strings, but it returns a list of `CollectionDescription`s, which we might extend in the future.

In code:
- https://github.com/qdrant/qdrant/blob/2a29432983b86a3eec231703a945ca19c09aec86/lib/api/src/rest/models.rs#L157-L162
- https://github.com/qdrant/qdrant/blob/2a29432983b86a3eec231703a945ca19c09aec86/lib/api/src/rest/models.rs#L132-L136
- ```json
  {
    "result": {
      "collections": [
        {
          "name": "test1"
        },
        {
          "name": "test2"
        }
      ]
    },
    "status": "ok",
    "time": 0.00003392
  }
  ```

So, for the shard key listing I'd expect to use a wrapping struct like this:

```rust
#[derive(Debug, Serialize, JsonSchema)]
pub struct ShardKeyDescription {
    pub shard_key: ShardKey,
}
```

src/tonic/api/collections_api.rs
coszio: I don't think we should error.

The endpoint is `/collections/{collection_name}/shards`, and right now the response we want is

```
{
  "shard_keys": [
     { "key": "my_shard_key" }
  ]
}
```

Since semantics refer to shards in general, this same endpoint can be extended to add another field called `"shards"`. Something like

```
{
  "shard_keys": [
     { "key": "my_shard_key", "shard_ids": [1] }
  ],
  "shards": [
    { "id": 1 }
  ]
}
```

Or just "shards" in the case of `ShardingMethod::Auto`
```
{
  "shards": [
    { "id": 1 }
  ]
}
```

amiremohamadi: I hadn’t considered errors in my last changes, as @coszio point makes sense to me. I’ll defer to you guys’ call.

generall (approved):

timvisee:
We've just released [Qdrant 1.16.2](https://github.com/qdrant/qdrant/releases/tag/v1.16.2) which includes these changes.

Note that we only update our API spec and clients in 1.17.0, but the implemented call is usable now.

amiremohamadi:
@timvisee i've submitted a PR to the client
https://github.com/qdrant/rust-client/pull/254

would appreciate a review when you have time!

timvisee:
We've released [Qdrant 1.17.0](https://github.com/qdrant/qdrant/releases/tag/v1.17.0) which includes this feature.

Documentation: https://qdrant.tech/documentation/guides/distributed_deployment/#user-defined-sharding

PR #7616, opened 2025-11-26
Add a ClusterOperation to execute a sleep on the node

tellet-q:
For builds with `staging` feature enabled allow an API call to POST /collections/{name}/cluster with SlowDownNode operation. It performs a simple sleep on the node.

```bash
  # Sleep for 1 second (default)
  curl -X POST "http://localhost:6333/collections/{collection_name}/cluster" \
    -H "Content-Type: application/json" \
    -d '{"test_slow_down": {}}'

  # Sleep for 5 seconds
  curl -X POST "http://localhost:6333/collections/{collection_name}/cluster" \
    -H "Content-Type: application/json" \
    -d '{"test_slow_down": {"duration": 5000}}'
```

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/operations/cluster_ops.rs
+++ b/lib/collection/src/operations/cluster_ops.rs
@@ -361,26 +361,4 @@ pub struct FinishResharding {}
 pub struct AbortResharding {}
 
 #[cfg(feature = "staging")]
-fn default_test_slow_down_duration_secs() -> f64 {
-    1.0
-}
-
-#[cfg(feature = "staging")]
-#[derive(Copy, Clone, Debug, Deserialize, Serialize, JsonSchema, Validate)]
-pub struct TestSlowDownOperation {
-    #[validate(nested)]
-    pub test_slow_down: TestSlowDown,
-}
-
-#[cfg(feature = "staging")]
-#[derive(Copy, Clone, Debug, Deserialize, Serialize, JsonSchema, Validate)]
-pub struct TestSlowDown {
-    /// Target peer ID to execute the sleep on.
-    /// If not specified, the operation will be executed on all peers.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub peer_id: Option<PeerId>,
-    /// Duration of the sleep in seconds (default: 1.0, max: 300.0).
-    #[serde(default = "default_test_slow_down_duration_secs")]
-    #[validate(range(min = 0.0, max = 300.0))]
-    pub duration_secs: f64,
-}
+pub use super::staging::{TestSlowDown, TestSlowDownOperation};
--- a/lib/collection/src/operations/mod.rs
+++ b/lib/collection/src/operations/mod.rs
@@ -11,6 +11,8 @@ pub mod shard_selector_internal;
 pub mod shared_storage_config;
 pub mod snapshot_ops;
 pub mod snapshot_storage_ops;
+#[cfg(feature = "staging")]
+pub mod staging;
 pub mod types;
 pub mod universal_query;
 pub mod validation;
--- a/lib/collection/src/operations/staging.rs
+++ b/lib/collection/src/operations/staging.rs
@@ -1,0 +1,31 @@
+//! Staging-only operations for testing and debugging purposes.
+//!
+//! This module contains operations that are only available when the `staging` feature is enabled.
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use validator::Validate;
+
+use crate::shards::shard::PeerId;
+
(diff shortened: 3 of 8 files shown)


lib/collection/src/operations/cluster_ops.rs
```diff
@@ -47,6 +47,10 @@ pub enum ClusterOperations {
     AbortResharding(AbortReshardingOperation),
     /// Trigger replication of points between two shards
     ReplicatePoints(ReplicatePointsOperation),
+
+    /// Introduce artificial delay to a node
+    #[cfg(feature = "staging")]
+    SlowDownNode(SlowDownNodeOperation),
```
timvisee: Nice!

Nit: I suggest to prefix these with 'test', so that they are clearly distinguishable from regular operations even in places where we call them. So it might become `TestSlowDown` and `test_slow_down`.

Wdyt?

lib/collection/src/operations/cluster_ops.rs
```diff
@@ -353,3 +359,24 @@ pub struct FinishResharding {}
...
+    pub slow_down_node: SlowDownNode,
+}
+
+#[cfg(feature = "staging")]
+#[derive(Copy, Clone, Debug, Deserialize, Serialize, JsonSchema, Validate)]
+pub struct SlowDownNode {
+    /// Duration of the sleep in milliseconds (default: 1000, max: 300000).
+    #[serde(default = "default_slow_down_duration_ms")]
+    #[validate(range(max = 300_000))]
+    pub duration_ms: u64,
```
timvisee: Another nit: I vote to keep seconds as base unit. We could do a float for sub second.

lib/collection/src/operations/cluster_ops.rs
```diff
@@ -353,3 +359,26 @@ pub struct FinishResharding {}
...
+
+#[cfg(feature = "staging")]
+#[derive(Copy, Clone, Debug, Deserialize, Serialize, JsonSchema, Validate)]
+pub struct TestSlowDown {
+    /// Target peer ID to execute the sleep on.
+    pub peer_id: PeerId,
+    /// Duration of the sleep in seconds (default: 1.0, max: 300.0).
+    #[serde(default = "default_test_slow_down_duration_secs")]
+    #[validate(range(max = 300.0))]
+    pub duration_secs: f64,
```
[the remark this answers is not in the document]
tellet-q: ok

generall:
Functionally, it looks good.
But WDYT about making it more separate from the other code?
I propose this:

- Large blocks of code can be moved into a function into a separate file, so it won't requires imports with `#[cfg(...` and won't bloat the actual production code
- Maybe do not make internal functions conditional? cc @timvisee

timvisee:
> Functionally, it looks good. But WDYT about making it more separate from the other code? I propose this:
> 
>     * Large blocks of code can be moved into a function into a separate file, so it won't requires imports with `#[cfg(...` and won't bloat the actual production code
> 
>     * Maybe do not make internal functions conditional? cc @timvisee

I like the idea.

The exception is the enum variant, and usage of that variant in the `match` block.

Your idea is that everything related to this that is currently in a match branch, should be in a dedicated function in a separate file. Right?

generall:
> Your idea is that everything related to this that is currently in a match branch, should be in a dedicated function in a separate file. Right?

yes

timvisee:
Another thing I noticed is this:

```http
POST /collections/{collection_name}/cluster
{"test_slow_down": {}}
```

The operation itself does not rely on any collection. But to make the API call, we are required to have a collection.

I don't suggest to change it now, as I find creating a dedicated API for this a bit extreme. But it might be worth to do in the future if we collect more of these calls not related to a collection.

lib/collection/src/operations/cluster_ops.rs
```diff
@@ -353,3 +359,28 @@ pub struct FinishResharding {}
...
+#[derive(Copy, Clone, Debug, Deserialize, Serialize, JsonSchema, Validate)]
+pub struct TestSlowDown {
+    /// Target peer ID to execute the sleep on.
+    /// If not specified, the operation will be executed on all peers.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub peer_id: Option<PeerId>,
+    /// Duration of the sleep in seconds (default: 1.0, max: 300.0).
+    #[serde(default = "default_test_slow_down_duration_secs")]
+    #[validate(range(min = 0.0, max = 300.0))]
+    pub duration_secs: f64,
```
timvisee: Imo it's fine to remove `_secs` as seconds are a base SI unit.

We also commonly use `timeout`, not `timeout_secs`.

PR #7618, opened 2025-11-26
Automatically opt-out of multi-mmap support by testing support

timvisee:
Depends on <https://github.com/qdrant/qdrant/pull/7614>

Fixes <https://github.com/qdrant/qdrant/issues/7610>, <https://github.com/qdrant/qdrant/issues/6676>

Automatically opt-out of using multiple memory maps on the same file if the runtime support check for it fails at startup. This way users don't have to specify `QDRANT_NO_MULTI_MMAP` and it should work out of the box.

We already had this check in place to log a warning on environments that suffer from this problem.

I am definitely not proud of how the result of the check ends up in `MULTI_MMAP_IS_SUPPORTED`. Please let me know if you have a better suggestion for it.

When starting on an unsupported environment, users will see this warning in the logs:

```
2025-11-26T14:48:21.920579Z  WARN memory::mmap_ops: Not using multi-mmap due to limited support, you may see reduced performance
```

### Tasks
- [x] Merge <https://github.com/qdrant/qdrant/pull/7614>
- [x] Rebase on dev
- [x] Undraft this PR

### All Submissions:

### New Feature Submissions:

### DIFF
--- a/lib/common/memory/src/mmap_ops.rs
+++ b/lib/common/memory/src/mmap_ops.rs
@@ -1,6 +1,6 @@
 use std::mem::{align_of, size_of};
 use std::path::Path;
-use std::sync::LazyLock;
+use std::sync::{LazyLock, OnceLock};
 use std::{io, mem, ptr};
 
 use fs_err as fs;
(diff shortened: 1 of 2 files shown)


timvisee:
I did some manual testing with Docker/WSL on Windows. Good news, these changes work!

### Without this patch

On Windows, I start a Qdrant container like this with a bind mount:

```bash
docker run --rm -it -p 6333:6333 -p 6334:6334 -v /c/Users/timvisee/qdrant/storage:/qdrant/storage:z qdrant/qdrant
```

Then I upsert points:

```bash
docker run -it --rm --net=host qdrant/bfb:dev ./bfb --skip-wait-index --indexing-threshold 0
```

After I restart the container, vector data is gone an reports all zeroes:

Screenshot From 2025-12-01 17-23-05

### With this patch

When using this patch, we now see a warning printed in the logs (4th line) telling us multi-mmap is disabled on this environment:

Screenshot From 2025-12-01 17-14-54

Then, after upserting again, vector data survives any number of container restarts!

Screenshot From 2025-12-01 17-24-04

lib/common/memory/src/mmap_ops.rs
```diff
@@ -17,13 +17,40 @@ pub const TEMP_FILE_EXTENSION: &str = "tmp";
...
+        log::warn!(
+            "Not using multi-mmap because QDRANT_NO_MULTI_MMAP is set, you may see reduced performance"
+        );
     }
+
     supported
 });
 
+/// If multi-mmap support is checked at Qdrant startup, the result is stored in this cell.
+pub static MULTI_MMAP_SUPPORT_CHECK_RESULT: OnceLock<bool> = OnceLock::new();
```
coszio, line 52: Feel free to disregard this stylistic comment.

This intermediate global variable should be possible to avoid by wrapping the above logic in a function, and use it to set `MULTI_MMAP_IS_SUPPORTED` directly after doing the check in `main.rs`

lib/common/memory/src/mmap_ops.rs
```diff
@@ -17,13 +17,40 @@ pub const TEMP_FILE_EXTENSION: &str = "tmp";
...
+            // We hit this branch if MULTI_MMAP_SUPPORT_CHECK_RESULT was never explicitly set
+            // In tests this can be safely ignored because we don't start through main.rs there
+            log::warn!(
+                "MULTI_MMAP_SUPPORT_CHECK_RESULT should be initialized before accessing MULTI_MMAP_IS_SUPPORTED"
+            );
+        }
+    }
+
+    // Opt-out if environment variable is set
+    if supported && std::env::var_os("QDRANT_NO_MULTI_MMAP").is_some_and(|val| !val.is_empty()) {
```
xzfc, line 41: Can this be a config option rather than an env var?

timvisee: We can, though it'd be a bit harder to propagate.

Since I deem this an experimental feature (which we don't know enough about yet), I vote to not put it in settings and keep it as undocumented variable.

Once we know more about this, both in terms of behavior and uses cases, we can still promote it to proper setting.

What do you think?

I'll merge this one now. But if you'd still prefer a setting, we can always do that in a follow-up PR.

lib/common/memory/src/mmap_ops.rs
timvisee: I don't fully get what you mean. Mind to open a PR on top on dev as proposal?

PR #7620, opened 2025-11-26
Fix search in empty HNSW segments

monoid:
For an empty HNSW segment, sampling `top` becomes 0, and `BatchFilteredSearcher` doesn't handle it properly, panicing.

### All Submissions:

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -1132,10 +1132,6 @@ impl HNSWIndex {
         params: Option<&SearchParams>,
         vector_query_context: &VectorQueryContext,
     ) -> OperationResult<Vec<Vec<ScoredPointOffset>>> {
-        if top == 0 {
-            return Ok(vec![vec![]; query_vectors.len()]);
-        }
-
         let id_tracker = self.id_tracker.borrow();
         let vector_storage = self.vector_storage.borrow();
         let quantized_vectors = self.quantized_vectors.borrow();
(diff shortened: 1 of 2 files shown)


timvisee:
Am I right that this only resolves a panic in our recent release if top becomes 0?

Out of curiosity; does that ever happen in practice, because we validate `limit >= 1` ([ref](https://api.qdrant.tech/api-reference/search/query-points#request.body.limit)).

monoid:
> Am I right that this only resolves a panic in our recent release if top becomes 0?
> 
> Out of curiosity; does that ever happen in practice, because we validate `limit >= 1` ([ref](https://api.qdrant.tech/api-reference/search/query-points#request.body.limit)).

Yes, it resolves panic if `top` becomes 0.

I observed it when I was experimented with parallel read and write.  Somehow an empty(!) HNSW segment got created, and [here](https://github.com/qdrant/qdrant/blob/e293ab1f08d3eb8fb8d9c3380c1de71b7a76570c/lib/collection/src/collection_manager/segments_searcher.rs#L644-L649) `top` becomes 0.  I presume it indexed a segment where all the points got deleted during the write.  Unfortunately, I cannot confirm it because the relevant piece of log was evicted by stacktraces.

Upd: but I think it should be a rare event.

timvisee:
It's good to protect against this anyway. Thank you for including tests!

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -1132,6 +1132,10 @@ impl HNSWIndex {
         params: Option<&SearchParams>,
         vector_query_context: &VectorQueryContext,
     ) -> OperationResult<Vec<Vec<ScoredPointOffset>>> {
+        if top == 0 {
```
IvanPleshkov: Do we also need this check for the indexed pass?

monoid: I will have a look.  It also worth adding an extra test.

IvanPleshkov: Thanks for the test, I like your decision about moving the check to the upper call stack

PR #7621, opened 2025-11-26
Fix BufferedDynamicFlags flushing with concurrent drop/deletion

agourlay:
Fixes flushing error happening when deleting a payload index concurrently with flushing.

```
2025-11-26T10:31:55.833087Z DEBUG collection::update_handler: Optimizer 'indexing' running on segments: [18, 19]
2025-11-26T10:31:55.834739Z DEBUG collection::collection_manager::optimizers::segment_optimizer: Available space: 801.99 GiB, needed for optimization: 110.93 MiB
2025-11-26T10:31:56.052782Z DEBUG segment::index::hnsw_index::hnsw: building HNSW for 10693 vectors with 8 CPUs
2025-11-26T10:32:00.428258Z  INFO actix_web::middleware::logger: 127.0.0.1 "GET /collections/benchmark HTTP/1.1" 200 462 "http://localhost:6333/dashboard" "Mozilla/5.0 (X11; Linux x86_64; rv:145.0) Gecko/20100101 Firefox/145.0" 0.009905
2025-11-26T10:32:04.295185Z DEBUG storage::content_manager::toc::collection_meta_ops: Drop payload index DropPayloadIndex { collection_name: "benchmark", field_name: JsonPath { first_key: "a", rest: [] } }
2025-11-26T10:32:06.021544Z  INFO actix_web::middleware::logger: 127.0.0.1 "DELETE /collections/benchmark/index/a?wait=true HTTP/1.1" 200 78 "http://localhost:6333/dashboard" "Mozilla/5.0 (X11; Linux x86_64; rv:145.0) Gecko/20100101 Firefox/145.0" 1.731689
2025-11-26T10:32:10.919258Z ERROR collection::update_handler: Failed to flush: Service runtime error: Failed to flush payload_index: Service runtime error: Failed to flush payload_index: Service runtime error: IO Error: failed to open file `./storage/collections/benchmark/0/segments/4fc17df6-0864-417e-a56b-fbe82bc7706f/payload_index/pabo24ma3euslum266ppk8us-a-null/has_values/flags_a.dat`: No such file or directory (os error 2)
```

This PR makes sure that drop and flushing are exclusive and uses a weak reference on storage to not trip over a deleted folder.

This was tested manually with `bfb` and adding/removing a keyword payload index.

### DIFF
--- a/lib/segment/src/index/field_index/null_index/mutable_null_index.rs
+++ b/lib/segment/src/index/field_index/null_index/mutable_null_index.rs
@@ -220,8 +220,11 @@ impl PayloadFieldIndex for MutableNullIndex {
     }
 
     fn cleanup(self) -> OperationResult<()> {
-        if self.base_dir.is_dir() {
-            fs::remove_dir_all(&self.base_dir)?;
+        let base_dir = self.base_dir.clone();
+        // drop mmap handles before deleting files
+        drop(self);
+        if base_dir.is_dir() {
+            fs::remove_dir_all(&base_dir)?;
         }
         Ok(())
     }


timvisee:
Yeah, it makes sense to me that we see it on this particular storage/flusher.

See, I believe this problem only appears when we try to reopen the file. In almost all storage types we don't and keep a file handle. When we remove the file, the file stays accessible through then handle as long as we keep it and we can still write to it. The kernel silently cleans it up once all handles are closed.

This storage does indeed try to reopen the file when the number of flags grows, breaking the pattern. We reopen through [`set_len`](https://github.com/qdrant/qdrant/blob/0a1a3d620b954c429cc0a3afaf697d7604f33871/lib/segment/src/common/flags/buffered_dynamic_flags.rs#L66), which reopens [here](https://github.com/qdrant/qdrant/blob/a0d62330c768c217743d5f6095d241a5ee058cec/lib/segment/src/common/flags/dynamic_mmap_flags.rs#L175).

So our other payload indices suffer from a similar problem. We just don't see an error report for it, as all late writes are silently voided. I believe we should audit all our storage types for similar problems.

Just thinking out loud here.

lib/segment/src/common/flags/buffered_dynamic_flags.rs
```diff
@@ -77,6 +97,13 @@ impl BufferedDynamicFlags {
     }
 }
 
+impl Drop for BufferedDynamicFlags {
+    fn drop(&mut self) {
+        // Wait for all background flush operations to finish
```
timvisee, line 102:
```suggestion
        // Wait for all background flush operations to finish, and cancel future flushes
```

lib/segment/src/common/flags/buffered_dynamic_flags.rs
```diff
@@ -56,8 +60,24 @@ impl BufferedDynamicFlags {
             (updates, required_len)
         };
 
-        let flags_arc = self.storage.clone();
+        // Weak reference to detect when the storage has been deleted
+        let flags_arc = Arc::downgrade(&self.storage);
```
coszio, line 64: I did similarly for #7624, this is an improvement so that we don't _start_ flushing a dropped component 👍

lib/segment/src/common/flags/buffered_dynamic_flags.rs
```diff
@@ -19,15 +19,19 @@ pub(crate) struct BufferedDynamicFlags {
 
     /// Pending changes to the storage flags.
     buffer: Arc<RwLock<AHashMap<PointOffsetType, bool>>>,
+
+    /// Lock to prevent concurrent flush and drop
+    pub is_alive_flush_lock: Arc<Mutex<bool>>,
```
coszio, line 24: After some consideration, I am not sure this will actually prevent trying to interact with an inexistent file/directory. In [`PayloadIndex::drop_index`](https://github.com/qdrant/qdrant/blob/80646ebba2bda62aad12461077cb33c11c910a4d/lib/segment/src/index/struct_payload_index.rs#L900) we actually remove the directory BEFORE dropping the index. 

This lock only makes the dropping wait for the end of the flush, but the files may already be deleted externally.

Please correct me if I missed something 🤔

coszio (changes requested):
See https://github.com/qdrant/qdrant/pull/7621/files#r2566368975

lib/segment/src/common/flags/buffered_dynamic_flags.rs
timvisee: Good call. That is a problem!

We must only delete files once we unload the payload index structure.

I am quite sure it's important to only delete after also releasing all files handles. Because Windows doesn't allow us to delete files which we still have open. Though then I'm also a bit puzzled on why we haven't seen issues about this. Maybe it needs a specific setup, like explicitly deleting a payload index which may not be very common.

timvisee: On second thought:

image

I do remember quite a few reports where Windows users complained about 'permission denied' errors. It may very well be caused by this.

agourlay: Thanks I did push a fix to order the drop properly https://github.com/qdrant/qdrant/pull/7621/commits/34d031dde1d3e161d06c9a982c94422a4c5655c1

> This lock only makes the dropping wait for the end of the flush, but the files may already be deleted externally.

My guess it that the original patch worked for me locally because the flush would happen after the drop anyway.

There is not a lot of time to squeeze the flusher between the FS remove and the implicit drop.

lib/segment/src/common/flags/buffered_dynamic_flags.rs
```diff
@@ -56,8 +60,24 @@ impl BufferedDynamicFlags {
...
-        let flags_arc = self.storage.clone();
+        // Weak reference to detect when the storage has been deleted
+        let flags_arc = Arc::downgrade(&self.storage);
+        let is_alive_flush_lock = self.is_alive_flush_lock.clone();
+
         Box::new(move || {
+            // Keep the guard till the end of the flush to prevent concurrent drop/flushes
+            let is_alive_flush_guard = is_alive_flush_lock.lock();
+
+            if !*is_alive_flush_guard {
```
coszio, line 71: We should be able to avoid this check by also using a weak reference to the lock. Then it would also simplify the mutex to just hold a unit struct `()`

coszio (approved):

lib/segment/src/common/flags/buffered_dynamic_flags.rs
agourlay: FYI started an audit here https://github.com/qdrant/qdrant/pull/7626

PR #7626, opened 2025-11-27
Fix payload indexes drop/flush sequence

agourlay:
Follow up on https://github.com/qdrant/qdrant/pull/7621

I performed an audit to make sure that other types of payload index also:
- drop their mmap handles **before** removing the files on disk
- do not drop during flush
- do not have concurrent flush

The `MutableBoolIndex` uses `BufferedDynamicFlags` which flushes properly since https://github.com/qdrant/qdrant/pull/7621

All the others indexes modified are built on top of `MmapBitSliceBufferedUpdateWrapper` which this PR fixes.

I also implemented the following previous suggestions:
- ~[simplify locking for drop with weak reference](https://github.com/qdrant/qdrant/pull/7621#discussion_r2568345639)~
- [better comment](https://github.com/qdrant/qdrant/pull/7621#discussion_r2565653578)

### DIFF
--- a/lib/segment/src/common/flags/buffered_dynamic_flags.rs
+++ b/lib/segment/src/common/flags/buffered_dynamic_flags.rs
@@ -21,17 +21,17 @@ pub(crate) struct BufferedDynamicFlags {
     buffer: Arc<RwLock<AHashMap<PointOffsetType, bool>>>,
 
     /// Lock to prevent concurrent flush and drop
-    flushing_lock: Arc<Mutex<()>>,
+    is_alive_flush_lock: Arc<Mutex<bool>>,
 }
 
 impl BufferedDynamicFlags {
     pub fn new(mmap_flags: DynamicMmapFlags) -> Self {
         let buffer = Arc::new(RwLock::new(AHashMap::new()));
-        let flushing_lock = Arc::new(Mutex::new(()));
+        let is_alive_flush_lock = Arc::new(Mutex::new(true));
         Self {
             storage: Arc::new(Mutex::new(mmap_flags)),
             buffer,
-            flushing_lock,
+            is_alive_flush_lock,
         }
     }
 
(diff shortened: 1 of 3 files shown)


lib/segment/src/index/field_index/geo_index/mmap_geo_index.rs
```diff
@@ -342,7 +342,9 @@ impl MmapGeoMapIndex {
 
     pub fn wipe(self) -> OperationResult<()> {
         let files = self.files();
-        let Self { path, .. } = self;
```
agourlay, line 345: the deconstruction is semantically equivalent but feels a bit hacky compared to an explicit drop

lib/segment/src/common/flags/buffered_dynamic_flags.rs
```diff
@@ -91,16 +88,15 @@ impl BufferedDynamicFlags {
...
         })
     }
 }
 
 impl Drop for BufferedDynamicFlags {
     fn drop(&mut self) {
-        // Wait for all background flush operations to finish
-        *self.is_alive_flush_lock.lock() = false;
+        // Wait for all background flush operations to finish, and cancel future flushes
+        _ = self.flushing_lock.lock();
```
timvisee: I don't think this works.

When we obtain this lock, the `Arc`s are still alive. And so a flusher thread is still able to upgrade, lock and flush after this drop function if the timing is right.

agourlay: Thanks for catching this, I will revert all usages for having a lock of a bool to signal properly that the drop occurred

agourlay: Applied revert https://github.com/qdrant/qdrant/pull/7626/commits/b1003fba95d5e571490c82d1008ee1866646a6e5

lib/segment/src/common/mmap_bitslice_buffered_update_wrapper.rs
```diff
@@ -82,3 +100,10 @@ impl MmapBitSliceBufferedUpdateWrapper {
         })
     }
 }
+
+impl Drop for MmapBitSliceBufferedUpdateWrapper {
+    fn drop(&mut self) {
+        // Wait for all background flush operations to finish, and cancel future flushes
+        _ = self.flushing_lock.lock();
+    }
```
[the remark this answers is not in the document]
timvisee: It does not defeat the purpose. We use this mechanism as a barrier, to ensure we have exclusive access when we pass it.

lib/segment/src/common/mmap_bitslice_buffered_update_wrapper.rs
```diff
@@ -68,10 +72,28 @@ impl MmapBitSliceBufferedUpdateWrapper {
...
+            }
+
+            let (Some(bitslice), Some(pending_updates_arc)) =
+                (bitslice.upgrade(), pending_updates_arc.upgrade())
+            else {
+                log::debug!(
+                    "Aborted flushing on a dropped MmapBitSliceBufferedUpdateWrapper instance"
+                );
+                return Ok(());
+            };
```
timvisee, line 95: I like that we still use weak refs.

It allows us to more reliably drop the Arc'ed structures together with the structure itself. Rather than having to wait on the flushers to 'recognize' the stop too.

PR #7627, opened 2025-11-27
Fix wiping of gridstore

coszio:
Related to #7624, #7621 

This PR patches 2 more things:
1. Makes sure we don't drop files while flushing, by waiting on `bitmask` guard to be released.
2. Swaps usage of `clear()` which would recreate the storage, where `wipe()` was supposed to be used

### DIFF
--- a/lib/gridstore/src/gridstore.rs
+++ b/lib/gridstore/src/gridstore.rs
@@ -41,6 +41,8 @@ pub struct Gridstore<V> {
     /// Bitmask to represent which "blocks" of data in the pages are used and which are free.
     ///
     /// 0 is free, 1 is used.
+    ///
+    /// Additionally, this is also used as a barrier to wait for a flush to finish when [`wipe`](Self::wipe)'ing.
     bitmask: Arc<RwLock<Bitmask>>,
     /// Path of the directory where the storage files are stored
     base_path: PathBuf,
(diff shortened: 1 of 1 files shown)


lib/gridstore/src/gridstore.rs
```diff
@@ -481,10 +481,18 @@ impl<V: Blob> Gridstore<V> {
...
     /// not allow further usage. Use [`clear`] instead to clear and reuse the storage.
     pub fn wipe(self) -> Result<()> {
-        // clear pages
-        self.pages.write().clear();
+        let base_path = self.base_path.clone();
+
+        // make sure we are not deleting files mid-flush
+        {
+            let _ = self.bitmask.write();
+        }
```
timvisee: Maybe a bit cleaner?

```suggestion
        // Barrier to wait for any ongoing flush to finish
        drop(self.bitmask.write());
```

Could you add a comment to the bitmask field in the struct to explain we also use it as barrier for dropping.

lib/segment/src/index/field_index/map_index/mutable_map_index.rs
```diff
@@ -301,7 +301,7 @@ where
         match self.storage {
             #[cfg(feature = "rocksdb")]
             Storage::RocksDb(db_wrapper) => db_wrapper.remove_column_family(),
-            Storage::Gridstore(mut store) => store.clear().map_err(|err| {
+            Storage::Gridstore(store) => store.wipe().map_err(|err| {
```
timvisee, line 304: Good catch!

PR #7629, opened 2025-11-27
Make `snapshot_manifest` methods `async`

ffuugoo:
So that they don't block async runtime. 💁‍♀️

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/snapshots.rs
+++ b/lib/collection/src/collection/snapshots.rs
@@ -326,7 +326,7 @@ impl Collection {
 
         // `ShardHolder::restore_shard_snapshot` is *not* cancel safe, so we spawn it onto runtime,
         // so that it won't be cancelled if current future is dropped
-        let restore = tokio::spawn(async move {
+        let restore = self.update_runtime.spawn(async move {
             shard_holder
                 .restore_shard_snapshot(
                     &snapshot_path,


ffuugoo:
Small caveat: I spawn `snapshot_manifest` on `search` runtime (cause it's a read-only op on segments) and `restore_shard_snapshot` on `update` runtime (cause, well, it's defo not "search"). Given the issues we've seen, I'd say maybe possibly this might have a small potential for a dead/live lock. Maybe.

But I'm also not sure if it would be any better to spawn both tasks on the same runtime.

KShivendu:
> But I'm also not sure if it would be any better to spawn both tasks on the same runtime

I think it should be better. Like deadlocks happens when one thread is waiting on other to free up stuff. But if it's all in the same runtime/thread, deadlock shouldn't happen, right?

timvisee:
> > But I'm also not sure if it would be any better to spawn both tasks on the same runtime
> 
> I think it should be better. Like deadlocks happens when one thread is waiting on other to free up stuff. But if it's all in the same runtime/thread, deadlock shouldn't happen, right?

In this case there's a spawn and a blocking spawn. Even within the same runtime they would already be isolated in different 'pools'.

PR #7630, opened 2025-11-27
Standardize delete payload index terminology

agourlay:
We have too many terms for deleting which causes confusion.

I propose to standardize the usage of `wipe` for deleting a payload index.

### DIFF
--- a/lib/segment/src/index/struct_payload_index.rs
+++ b/lib/segment/src/index/struct_payload_index.rs
@@ -272,11 +272,11 @@ impl StructPayloadIndex {
             }
             self.config.skip_rocksdb.replace(true);
 
-            // Clean-up all existing indices
+            // Wipe all existing indices
             for index in indexes.drain(..) {
-                index.cleanup().map_err(|err| {
+                index.wipe().map_err(|err| {
                     OperationError::service_error(format!(
-                        "Failed to clean up payload index for field `{field}` before rebuild: {err}"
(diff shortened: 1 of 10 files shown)


lib/segment/src/index/struct_payload_index.rs
```diff
@@ -272,11 +272,11 @@ impl StructPayloadIndex {
...
             self.config.skip_rocksdb.replace(true);
 
-            // Clean-up all existing indices
+            // Wipe all existing indices
             for index in indexes.drain(..) {
-                index.cleanup().map_err(|err| {
+                index.wipe().map_err(|err| {
                     OperationError::service_error(format!(
-                        "Failed to clean up payload index for field `{field}` before rebuild: {err}"
+                        "Failed to delete existing payload index for field `{field}` before rebuild: {err}"
```
coszio, line 279: 🤔

agourlay: I wanted to use a term that our users would for sure understand.
Want me to use wipe there?

timvisee: What demographic?

```suggestion
                        "Failed to yeet payload index for field `{field}` before rebuild: {err}"
```

coszio: Just pointing out the non-homogenous change, I think wipe would sound worse 👍

PR #7631, opened 2025-11-27
[cluster telemetry] Internal service conversions

coszio:
Depends on #7599 

Implements conversions between internal telemetry service and existing types.

### Tasks
- [x] grpc -> existing
- [x] existing -> grpc

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -11670,7 +11670,7 @@
             "minimum": 0
           },
           "to_shard_id": {
-            "description": "Target shard ID if different than source shard ID\n\nUsed exclusively with `ReshardStreamRecords` transfer method.",
+            "description": "Target shard ID if different than source shard ID\n\nUsed exclusively with `ReshardingStreamRecords` transfer method.",
             "type": "integer",
             "format": "uint32",
             "minimum": 0,
--- a/lib/api/src/grpc/proto/telemetry_internal.proto
+++ b/lib/api/src/grpc/proto/telemetry_internal.proto
@@ -112,10 +112,10 @@ message ShardCleanStatusTelemetry {
 
 message ClusterTelemetry {
   ClusterStatusTelemetry status = 1;
-  reserved 2;                      // config
-  map<uint64, PeerInfo> peers = 3; // peers
-  reserved 4;                      // peer_metadata
-  reserved 5;                      // metadata
+  reserved 2; // config
+  map<uint64, PeerInfo> peers = 3;
+  reserved 4; // peer_metadata
+  reserved 5; // metadata
 }
 
 message ClusterStatusTelemetry {
--- a/lib/storage/src/content_manager/conversions.rs
+++ b/lib/storage/src/content_manager/conversions.rs
@@ -273,7 +273,7 @@ impl TryFrom<grpc::AliasOperations> for AliasOperations {
             Some(grpc::alias_operations::Action::CreateAlias(create)) => Ok(create.into()),
             Some(grpc::alias_operations::Action::DeleteAlias(delete)) => Ok(delete.into()),
             Some(grpc::alias_operations::Action::RenameAlias(rename)) => Ok(rename.into()),
-            _ => Err(Status::invalid_argument("Malformed AliasOperation type")),
+            None => Err(Status::invalid_argument("Malformed AliasOperation type")),
         }
     }
 }
--- a/lib/api/src/grpc/qdrant.rs
+++ b/lib/api/src/grpc/qdrant.rs
@@ -12602,7 +12602,6 @@ pub mod shard_clean_status_telemetry {
 pub struct ClusterTelemetry {
     #[prost(message, optional, tag = "1")]
     pub status: ::core::option::Option<ClusterStatusTelemetry>,
-    /// peers
     #[prost(map = "uint64, message", tag = "3")]
     pub peers: ::std::collections::HashMap<u64, PeerInfo>,
 }


lib/api/src/grpc/proto/telemetry_internal.proto
```diff
@@ -11,6 +11,11 @@ message GetPeerTelemetryRequest {
 }
 
 message GetPeerTelemetryResponse {
+    PeerTelemetry result = 1;
+    double time = 2;
+}
```
timvisee, line 18: Any particular reason for using nesting into `result` here? Just so that it matches with the REST API?

lib/collection/src/telemetry.rs
```diff
@@ -180,3 +182,269 @@ impl From<CollectionConfigInternal> for CollectionConfigTelemetry {
...
+                method: Some(ShardTransferMethod::from(
+                    grpc::ShardTransferMethod::try_from(value.method).map_err(|err| {
+                        Status::invalid_argument(format!("cannot decode ShardTransferMethod {err}"))
+                    })?,
+                )),
+                comment: if value.comment.is_empty() {
+                    None
+                } else {
+                    Some(value.comment)
+                },
```
timvisee: Alternative. I'll let you choose

```suggestion
                comment: Some(value.comment).filter(|c| !c.is_empty()),
```

src/common/telemetry_ops/conversions.rs
```diff
@@ -0,0 +1,126 @@
...
+            .map(|(peer_id, grpc::PeerInfo { uri })| (peer_id, PeerInfo { uri }))
+            .collect::<HashMap<_, _>>();
+
+        Ok(ClusterTelemetry {
+            enabled: true, // Not provided in gRPC, default to true if we have telemetry
+            status: status.map(ClusterStatusTelemetry::try_from).transpose()?,
+            config: None, // Not provided in gRPC
+            peers: (!peers.is_empty()).then_some(peers),
+            peer_metadata: None, // Not provided in gRPC
+            metadata: None,      // Not provided in gRPC
```
timvisee, line 99: I don't like these `None`s here so much. It may not be obvious in code that these will always be empty for some code paths.

At the same time I don't have any better suggestion.

I don't like duplicating all types another time just to remove those fields. And having `reserved` in the gRPC variants makes a lot of sense.

src/common/telemetry.rs
```diff
@@ -134,3 +145,78 @@ impl TelemetryCollector {
...
+                Ok(CollectionTelemetryEnum::Full(Box::new(
+                    CollectionTelemetry::try_from(collection)?,
+                )))
+            })
+            .try_collect::<_, Vec<_>, Status>()?;
+
+        let cluster = cluster.map(ClusterTelemetry::try_from).transpose()?;
+
+        Ok(TelemetryData {
+            id: "".to_string(),
```
timvisee, line 171: Don't we need a collection name here?

lib/api/src/grpc/proto/telemetry_internal.proto
coszio: Yes, but also matches other grpc endpoints

src/common/telemetry.rs
coszio: No no, this is the telemetry collector ID, which we don't care about in internal service

src/common/telemetry_ops/conversions.rs
coszio: I think this conversion + reserved fields is the best we can do. Unless there is another suggestion to evaluate

lib/collection/src/telemetry.rs
```diff
@@ -180,3 +182,265 @@ impl From<CollectionConfigInternal> for CollectionConfigTelemetry {
...
+
+    impl From<ShardTransferInfo> for grpc::ShardTransferTelemetry {
+        fn from(value: ShardTransferInfo) -> Self {
+            grpc::ShardTransferTelemetry {
+                shard_id: value.shard_id,
+                to_shard_id: value.to_shard_id,
+                from: value.from,
+                to: value.to,
+                sync: value.sync,
+                method: grpc::ShardTransferMethod::from(value.method.unwrap_or_default()) as i32,
```
timvisee: I'm not confident this is valid.

We don't use `ShardTransferMethod::default()` in all places, and actually prefer to use the user configured default instead (in configuration).

Any strict reason for requiring this, and doing it this way?

docs/redoc/master/openapi.json
```diff
@@ -11692,7 +11691,14 @@
...
           "app": {
-            "$ref": "#/components/schemas/AppBuildTelemetry"
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/AppBuildTelemetry"
+              },
+              {
+                "nullable": true
+              }
+            ]
```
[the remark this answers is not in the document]
coszio: This is handled by `schemars`, and autogenerated clients can use it just fine

lib/collection/src/telemetry.rs
coszio: That's true, I believe the best way to do it would be to leave it as optional, and rely on telemetry collector infra to populate it correctly.

lib/storage/src/content_manager/conversions.rs
```diff
@@ -261,31 +264,25 @@ impl From<api::grpc::qdrant::RenameAlias> for AliasOperations {
...
-                Ok(delete.into())
-            }
-            Some(api::grpc::qdrant::alias_operations::Action::RenameAlias(rename)) => {
-                Ok(rename.into())
-            }
+            Some(grpc::alias_operations::Action::CreateAlias(create)) => Ok(create.into()),
+            Some(grpc::alias_operations::Action::DeleteAlias(delete)) => Ok(delete.into()),
+            Some(grpc::alias_operations::Action::RenameAlias(rename)) => Ok(rename.into()),
             _ => Err(Status::invalid_argument("Malformed AliasOperation type")),
         }
```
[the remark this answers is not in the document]
coszio: Not part of this PR, but ok 👍

PR #7636, opened 2025-11-28
Hotfix to prevent deadlock when restoring snapshot on single CPU

timvisee:
Temporary runtime tweak for <https://github.com/qdrant/qdrant/pull/7629> because it currently allows to easily trigger a deadlock.

A proper fix is more involved and will take more time to implement. Once implemented, this PR will be reverted.

The proper fix consists of restructuring the on drop handler for local shards. We'll have to gracefully stop shard workers through an async function, not through the drop handler.

### All Submissions:

### DIFF
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -80,6 +80,8 @@ pub struct Collection {
     // Lock is acquired for read on update operation and can be acquired for write externally,
     // which will block all update operations until the lock is released.
     updates_lock: Arc<RwLock<()>>,
+    // General runtime handle.
+    general_runtime: Handle,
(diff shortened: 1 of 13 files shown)


ffuugoo (approved):

timvisee:
@generall is already working on a better fix, so we might want to merge that instead

I'll leave this open for the time being so we have this available if we need to roll a quick patch release.

timvisee:
Superseded by <https://github.com/qdrant/qdrant/pull/7640>

timvisee:
Fix released as part of [Qdrant 1.16.2](https://github.com/qdrant/qdrant/releases/tag/v1.16.2).

PR #7637, opened 2025-11-28
Make `ShardTransfer.method` mandatory

coszio:
Partially, a sanity refactor. This PR makes `method` field consistent with the mental model, where it should always have a value once it has been accepted. 

Original motivation was because this was forcing the internal telemetry service proto to set the `method` as optional, when in all cases it should already have a value.

This PR moves the decision point for selecting a method to the caller (`MoveShard`, `ReplicateShard`, `request_shard_transfer`) by using the configured default from collection.

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -11403,6 +11403,7 @@
         "type": "object",
         "required": [
           "from",
+          "method",
           "shard_id",
           "sync",
           "to"
@@ -11437,14 +11438,7 @@
             "type": "boolean"
           },
           "method": {
-            "anyOf": [
-              {
-                "$ref": "#/components/schemas/ShardTransferMethod"
-              },
-              {
-                "nullable": true
-              }
-            ]
+            "$ref": "#/components/schemas/ShardTransferMethod"
           },
           "comment": {
             "description": "A human-readable report of the transfer progress. Available only on the source peer.",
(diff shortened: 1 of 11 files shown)


lib/collection/src/shards/transfer/mod.rs
```diff
@@ -43,9 +43,9 @@ pub struct ShardTransfer {
     /// If this flag is true, this is a replication related transfer of shard from 1 peer to another
     /// Shard on original peer will not be deleted in this case
     pub sync: bool,
-    /// Method to transfer shard with. `None` to choose automatically.
+    /// Method to transfer shard with.
     #[serde(default)]
-    pub method: Option<ShardTransferMethod>,
```
coszio, line 48: Please note.

By the point this is written to disk, it should always have a value. If this assumption is correct, then there should be no problem here.

However, if by some reason it didn't (maybe some earlier version did not have this), `ShardTransferMethod::default()` will be set automatically here, overriding the collection's default method set in the config.

generall:
I would prefer to keep Option. In my experience, it is always way easier to deal with optional fields if we need to migrate something to somewhere.

docs/redoc/master/openapi.json
```diff
@@ -11437,14 +11438,7 @@
...
-            "anyOf": [
-              {
-                "$ref": "#/components/schemas/ShardTransferMethod"
-              },
-              {
-                "nullable": true
-              }
-            ]
+            "$ref": "#/components/schemas/ShardTransferMethod"
           },
```
timvisee, line 11442: Hint for the others, this makes it mandatory in `ShardTransferInfo`. Not in the user request.

generall: still, I would prefer to keep it optional

coszio:
:ok_honey:

PR #7638, opened 2025-11-28
Fix Gridstore has_pointer to account for delete pending operation

agourlay:
This PR fixes a bug in Gridstore's `has_pointer` operation.

The bug manifests itself by an incorrect result from the `put_value` operation.

The contract for the returned boolean is

```rust
    /// Put a value in the storage.
    ///
    /// Returns true if the value existed previously and was updated, false if it was newly inserted.
    pub fn put_value(
        &mut self,
        point_offset: PointOffset,
        value: &V,
        hw_counter: HwMetricRefCounter,
    ) -> Result<bool> {
```

The value returned was incorrect in case of a delete operation present in the pending operation.

Luckily for us, no call site for `put_value` seems to rely on this returned information :relieved:

### DIFF
--- a/lib/gridstore/src/gridstore.rs
+++ b/lib/gridstore/src/gridstore.rs
@@ -878,12 +878,14 @@ mod tests {
         Put(PointOffset, Payload),
         Delete(PointOffset),
         Update(PointOffset, Payload),
+        Get(PointOffset),
+        Flush,
     }
 
     impl Operation {
         fn random(rng: &mut impl Rng, max_point_offset: u32) -> Self {
             let point_offset = rng.random_range(0..=max_point_offset);
-            let operation = rng.random_range(0..3);
+            let operation = rng.random_range(0..4);
             match operation {
                 0 => {
                     let size_factor = rng.random_range(1..10);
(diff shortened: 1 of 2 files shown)


lib/gridstore/src/tracker.rs
```diff
@@ -361,7 +361,7 @@ impl Tracker {
     }
 
     pub fn has_pointer(&self, point_offset: PointOffset) -> bool {
-        self.pending_updates.contains_key(&point_offset) || self.get(point_offset).is_some()
+        self.get(point_offset).is_some()
```
agourlay, line 364: `pending_updates.contains_key` is not correct, the pending operation might be a delete

the `get` method implements this check properly internally

lib/gridstore/src/gridstore.rs
```diff
@@ -878,12 +878,14 @@ mod tests {
         Put(PointOffset, Payload),
         Delete(PointOffset),
         Update(PointOffset, Payload),
+        Get(PointOffset),
```
agourlay, line 881: discovered the bug while improving our randomized test :)

lib/gridstore/src/tracker.rs
timvisee: Thanks!

The key thing is that `self.get(point_offset)` always represents the latest data. And so `pending_updates` must be ignored here as it's purely for persisting storage separately.

PR #7640, opened 2025-11-29
Explicit gracefull stop for LocalShard

generall:
Motivation:

Having long running `stop_gracefully` function inside Drop is bad on it's own, but even worse if this long drop needs to call for async runtime operations.
In our case it created a situation, where Drop can not finish because it wants to spawn another async task on a runtime, which is has already used for drop itself. This results in a Deadlock, in those cases when runtime has only one available thread.

This PR refactors Drop in a way, that it only makes very small and syncronous operations, while relying on external explicit shutdown, which should be executed on a real async runtime.

- **sync function to ask to stop workers**
- **graceful shard stop on ShardReplicaSet changes**
- **make stop consuming**
- **implement graceful stop for replica set**
- **graceful shard stop in start_resharding_unchecked**
- **stop_gracefully fn on shard_holder**
- **collection graceful stop on delete**
- **explain when it is ok to shutdown**

Recommed to review by commits

### DIFF
--- a/lib/collection/src/collection/collection_ops.rs
+++ b/lib/collection/src/collection/collection_ops.rs
@@ -280,12 +280,20 @@ impl Collection {
     ///
     /// Partially blocking. Stopping existing optimizers is blocking. Starting new optimizers is
     /// not blocking.
+    ///
+    /// ## Cancel safety
+    ///
+    /// This function is cancel safe, and will always run to completion.
     pub async fn recreate_optimizers_blocking(&self) -> CollectionResult<()> {
-        let shard_holder = self.shards_holder.read().await;
-        let updates = shard_holder
-            .all_shards()
-            .map(|replica_set| replica_set.on_optimizer_config_update());
-        future::try_join_all(updates).await?;
+        let shards_holder = self.shards_holder.clone();
+        tokio::task::spawn(async move {
+            let shard_holder = shards_holder.read().await;
+            let updates = shard_holder
+                .all_shards()
+                .map(|replica_set| replica_set.on_optimizer_config_update());
+            future::try_join_all(updates).await
+        })
+        .await??;
         Ok(())
     }
 
--- a/lib/storage/src/dispatcher.rs
+++ b/lib/storage/src/dispatcher.rs
@@ -65,7 +65,14 @@ impl Dispatcher {
     }
 
     /// If `wait_timeout` is not supplied - then default duration will be used.
+    ///
     /// This function needs to be called from a runtime with timers enabled.
+    ///
+    /// ## Cancel safety
+    ///
+    /// This function is cancel safe.
+    ///
+    /// On deployments without consensus - a submitted operation is always run to completion.
     pub async fn submit_collection_meta_op(
         &self,
         operation: CollectionMetaOperations,
@@ -247,7 +254,9 @@ impl Dispatcher {
 
             Ok(res)
         } else {
-            self.toc.perform_collection_meta_op(operation).await
+            let toc = self.toc.clone();
+            tokio::task::spawn(async move { toc.perform_collection_meta_op(operation).await })
+                .await?
         }
     }
 


lib/collection/src/shards/local_shard/drop.rs
```diff
@@ -0,0 +1,34 @@
...
+        // However.
+        // It is not possible to explicitly catch all cases of LocalShard being dropped.
+        // For example, then the service is shutdown by interruption, we might not have a
+        // deep enough control over the shutdown sequence.
+        // So we have to assume, that it is fine to not await for explicit shutdown in some cases.
+        // But we still want to call explicit shutdown on all removes and internal operations.
+        if !already_stopped {
+            log::debug!("Dropping LocalShard while it is not already stopped");
+        }
+    }
```
timvisee, line 63: The scoped thread here does not make sense. Because we join it, it'll always block the current thread for the lifetime of the spawned thread. The idea that spawning a thread here will unblock the current (drop) runtime is not correct. Spawning the extra thread is therefore obsolete. In this case, we can just run `self.blocking_ask_workers_to_stop()` directly on our current thread.

Also, this still has the potential to deadlock if we lock `update_handler` through other means. So this only shifts the problem.

For example, we will deadlock again if:
- we run on 1 CPU
- we yield the thread at one of these await points: https://github.com/qdrant/qdrant/blob/28112fdffa7e393d76c9d38e6a843ae2e91e42bf/lib/collection/src/shards/local_shard/updaters.rs#L64-L67
- we run `on_optimizer_config_update` and implicit `drop` on same runtime

I'd argue assuming that won't happen, or trying our best to not run these on the same runtime is not good enough.

Instead I think the only right way to solve this is to:
- still externally stop it gracefully first (like in this PR)
- make this drop handler not block (nor join) at all
- use an atomic bool or other lock free structure to probe status, print an error if not stopped
- if not stopped, stop it asynchronously (stop msg through channel?)

I've yet to validate whether this is easily doable.

lib/storage/src/content_manager/toc/collection_meta_ops.rs
```diff
@@ -216,8 +217,7 @@ impl TableOfContent {
                     state.key(),
                 );
             }
-
-            drop(removed);
+            removed.stop_gracefully().await;
```
timvisee, line 223: I'd like to still explicitly drop here, so that we have control at what **exact** time the structure is dropped.

```suggestion
            removed.stop_gracefully().await;
            drop(removed);
```

Shouldn't change runtime behavior in this case though.

lib/collection/src/tests/fix_payload_indices.rs
```diff
@@ -65,7 +65,7 @@ async fn test_fix_payload_indices() {
 
     std::thread::sleep(std::time::Duration::from_secs(1));
 
-    drop(shard);
+    shard.stop_gracefully().await;
```
timvisee, line 68:
```suggestion
    shard.stop_gracefully().await;
    drop(shard);
```

lib/collection/src/tests/wal_recovery_test.rs
```diff
@@ -82,7 +82,7 @@ async fn test_delete_from_indexed_payload() {
         .unwrap()
         .points;
 
-    drop(shard);
+    shard.stop_gracefully().await;
```
timvisee, line 85:
```suggestion
    shard.stop_gracefully().await;
    drop(shard);
```

lib/collection/src/tests/wal_recovery_test.rs
```diff
@@ -109,7 +109,7 @@ async fn test_delete_from_indexed_payload() {
         .await
         .unwrap();
 
-    drop(shard);
+    shard.stop_gracefully().await;
```
timvisee, line 112:
```suggestion
    shard.stop_gracefully().await;
    drop(shard);
```

lib/collection/src/tests/wal_recovery_test.rs
```diff
@@ -210,7 +210,7 @@ async fn test_partial_flush_recovery() {
     // This only flushed id-tracker-mapping, but not the storage change
     shard.partial_flush();
 
-    drop(shard);
+    shard.stop_gracefully().await;
```
timvisee, line 213:
```suggestion
    shard.stop_gracefully().await;
    drop(shard);
```

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -110,7 +110,7 @@ impl ShardHolder {
     }
 
     // TODO: do not leave broken intermediate state if this fails midway?
-    pub fn start_resharding_unchecked(
+    pub async fn start_resharding_unchecked(
```
timvisee, line 116: This function now is not cancel safe.

It doesn't look like we can reach this in a cancellable way, but let's at least annotate it.

Maybe it'd actually best to always spawn inside [`perform_collection_meta_op`](https://github.com/qdrant/qdrant/blob/fe437272493ec55a9107799ed27061c2c3a764bf/lib/storage/src/content_manager/toc/collection_meta_ops.rs#L35) to prevent any kind of cancellation causing issues with ToC/consensus operations.

We do this when consensus is enabled [here](https://github.com/qdrant/qdrant/blob/fe437272493ec55a9107799ed27061c2c3a764bf/lib/storage/src/content_manager/toc/collection_meta_ops.rs#L27-L33). But we don't seem to always do it for [this](https://github.com/qdrant/qdrant/blob/7b11a3782ea6ef2ad1bbd8230b9220c97ea710b6/lib/storage/src/dispatcher.rs#L250).

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -200,13 +213,18 @@ impl ShardHolder {
         Ok(())
     }
 
-    pub fn add_shard(
+    pub async fn add_shard(
```
timvisee, line 219: This function is now not cancel safe.

lib/collection/src/shards/local_shard/updaters.rs
```diff
@@ -0,0 +1,85 @@
...
+        // This can block longer, if the channel is full
+        // If channel is closed, assume it is already stopped
+        if let Err(err) = self.update_sender.load().blocking_send(UpdateSignal::Stop) {
+            log::trace!("Error sending update signal to update handler: {err}");
+        }
+
+        false
+    }
+
+    pub async fn on_optimizer_config_update(&self) -> CollectionResult<()> {
```
timvisee, line 34: This is **not** cancel safe. We must annotate that, and ensure we don't cancel it in any way.

lib/collection/src/shards/replica_set/mod.rs
timvisee, line 607: We must gracefully stop before calling this `clear`, as this `clear` touches files.

If we then fail to build the new shard, we fall back to starting the necessary workers again.

lib/collection/src/tests/wal_recovery_test.rs
generall: `stop_gracefully` is consuming shard

lib/collection/src/tests/wal_recovery_test.rs
generall: `stop_gracefully` is consuming shard

lib/collection/src/tests/wal_recovery_test.rs
generall: `stop_gracefully` is consuming shard

lib/collection/src/tests/fix_payload_indices.rs
generall: `stop_gracefully` is consuming shard

lib/storage/src/content_manager/toc/collection_meta_ops.rs
generall: `stop_gracefully` is consuming collection

lib/collection/src/shards/local_shard/updaters.rs
generall: this is not a new function, it is simply moved to another file

timvisee: I added a spawn upstream to ensure we don't cancel this function half way (will push shortly).

lib/collection/src/shards/local_shard/drop.rs
timvisee: While we didn't come up with a perfect solution for this, it is much better now.

Under normal circumstances - when `stop_gracefully` is correctly called - drop does not block anymore and shortcuts early.

In cases where we forgot the shard workers will be cancelled asynchronously.

This means that all places that depend on releasing everything need to call ``stop_gracefully().await`. This has been implemented across the code base.

timvisee (approved):

PR #7642, opened 2025-11-30
refactor update workers

generall:
- **refactor: move flush workers into a dedicated file**
- **move optimizer worker into a dedicated file**
- **move update worker into a dedicated file**


generall:
@copilot please check that moved functions do not contain any changes in the logic. This PR is intended for refactoring only

PR #7643, opened 2025-11-30
Prevent unoptimized updates

generall:
## Motivation

Large amount of updates, especially in bulk/spiky load pattern can temporary create a large amount of unindexed vectors.
Those unindexed vectors may create higher latencies of the search requests, even though the application might not need just-in-time updates.

In order to prevent those latencies, qdrant already have `indexed_only` parameter in the query. So it is covered from the search point of view, there is, however, a side-effect that updated but not yet indexed points might disappear from the search results. This PR addresses this problem.

To prevent disappearence of the updated points from the search results, we can throttle the update operation to match the indexation speed. In other words we can delay the updates, if index is not able to keep up. In this configuration we will always have all points available in search, but the update operations migth take more time to be applied.

## Implementation

~Itroduce a new parameter inside `strict_mode` (?) configuration, which would introduce an additional threshold:~

Introduce a new parameter inside optimizers config, which would enable or disable unoptimized updates.
Threshold is the same as indexing threshold (it seems there are no benefit in starting optimization earlier, as we would need to merge segments anyway)

```rust
/// If this option is set, service will try to prevent creation of large unoptimized segments.
/// When enabled, updates may be delayed if there are unoptimized segments larger than indexing threshold.
/// Updates will be resumed when optimization is completed and segments are optimized below the threshold.
/// Using this option may lead to increased delay between submitting an update and its application.
/// Default is disabled.
#[serde(default)]
pub prevent_unoptimized: Option<bool>,
```

# ToDo:

- [x] make actual ~strict-mode~ optimizers parameter
- [ ] ~Validate that indexing threshold is larger that strcict mode parameter (otherwise we might deadlock updates)~
- [x] Validate that we only trottle if actual optimizations are running

# To consider

- In case of long-running optimization on a big segment we might not have a budget to index small segments. How to handle that?

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -7587,7 +7587,7 @@
             "nullable": true
           },
           "prevent_unoptimized": {
-            "description": "If this option is set, service will try to prevent creation of large unoptimized segments. When enabled, updates may be delayed if there are unoptimized segments larger than indexing threshold. Updates will be resumed when optimization is completed and segments are optimized below the threshold. Using this option may lead to increased delay between submitting an update and its application. Default is disabled.",
+            "description": "If this option is set, service will try to prevent creation of large unoptimized segments. When enabled, updates may be blocked at request level if there are unoptimized segments larger than indexing threshold. Updates will be resumed when optimization is completed and segments are optimized below the threshold. Using this option may lead to increased delay between submitting an update and its application. Default is disabled.",
             "default": null,
             "type": "boolean",
             "nullable": true
@@ -10220,7 +10220,7 @@
             ]
           },
           "prevent_unoptimized": {
-            "description": "If this option is set, service will try to prevent creation of large unoptimized segments. When enabled, updates may be delayed if there are unoptimized segments larger than indexing threshold. Updates will be resumed when optimization is completed and segments are optimized below the threshold. Using this option may lead to increased delay between submitting an update and its application. Default is disabled.",
+            "description": "If this option is set, service will try to prevent creation of large unoptimized segments. When enabled, updates may be blocked at request level if there are unoptimized segments larger than indexing threshold. Updates will be resumed when optimization is completed and segments are optimized below the threshold. Using this option may lead to increased delay between submitting an update and its application. Default is disabled.",
             "default": null,
             "type": "boolean",
             "nullable": true
--- a/lib/api/src/grpc/qdrant.rs
+++ b/lib/api/src/grpc/qdrant.rs
@@ -856,7 +856,7 @@ pub struct OptimizersConfigDiff {
     #[prost(message, optional, tag = "9")]
     pub max_optimization_threads: ::core::option::Option<MaxOptimizationThreads>,
     /// If this option is set, service will try to prevent creation of large unoptimized segments.
-    /// When enabled, updates may be delayed if there are unoptimized segments larger than indexing threshold.
+    /// When enabled, updates may be blocked at request level if there are unoptimized segments larger than indexing threshold.
     /// Updates will be resumed when optimization is completed and segments are optimized below the threshold.
     /// Using this option may lead to increased delay between submitting an update and its application.
     /// Default is disabled.
--- a/lib/collection/src/operations/config_diff.rs
+++ b/lib/collection/src/operations/config_diff.rs
@@ -160,7 +160,7 @@ pub struct OptimizersConfigDiff {
     pub max_optimization_threads: Option<MaxOptimizationThreads>,
 
     /// If this option is set, service will try to prevent creation of large unoptimized segments.
-    /// When enabled, updates may be delayed if there are unoptimized segments larger than indexing threshold.
+    /// When enabled, updates may be blocked at request level if there are unoptimized segments larger than indexing threshold.
     /// Updates will be resumed when optimization is completed and segments are optimized below the threshold.
     /// Using this option may lead to increased delay between submitting an update and its application.
     /// Default is disabled.
--- a/lib/collection/src/optimizers_builder.rs
+++ b/lib/collection/src/optimizers_builder.rs
@@ -84,7 +84,7 @@ pub struct OptimizersConfig {
     pub max_optimization_threads: Option<usize>,
 
     /// If this option is set, service will try to prevent creation of large unoptimized segments.
-    /// When enabled, updates may be delayed if there are unoptimized segments larger than indexing threshold.
+    /// When enabled, updates may be blocked at request level if there are unoptimized segments larger than indexing threshold.
     /// Updates will be resumed when optimization is completed and segments are optimized below the threshold.
     /// Using this option may lead to increased delay between submitting an update and its application.
     /// Default is disabled.


lib/collection/src/update_workers/update_worker.rs
```diff
@@ -103,6 +123,55 @@ impl UpdateWorkers {
...
+
+            // If unoptimized segments are too large, the only way it can be fixed is optimization
+            // So we wait the notification of optimization completion to re-check the sizes
+            if let Err(err) = optimization_finished_receiver.changed().await {
+                // this can be if optimization is cancelled, we don't need to wait anymore
+                log::debug!("Optimization thread terminated with an error: {err}");
+                return Ok(());
+            }
+        }
+    }
```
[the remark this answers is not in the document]
agourlay: took care of it in https://github.com/qdrant/qdrant/pull/7643/commits/1ab48ac61f53c7b870b06c706ca0a4bfef30cd43

agourlay:
Local experiments with bfb:

```
cargo run -r -- -n 1M -d 256 --skip-create
```

| prevent_unoptimized | Upload Time |Wait  Indexing Time (bfb) |
|--------------------|------------|---------------|
| false (default)     | 7 seconds  | 10 seconds    |
| true                | 2m 30s     | 3 seconds     |

The slowdown seems a bit excessive in that case.
Needs further investigation.

generall:
it seems we can only update in batches of 10k between optimizations

lib/api/src/grpc/proto/collections.proto
```diff
@@ -306,6 +306,13 @@ message OptimizersConfigDiff {
   // If "auto" - have no limit and choose dynamically to saturate CPU.
   // If 0 - no optimization threads, optimizations will be disabled.
   optional MaxOptimizationThreads max_optimization_threads = 9;
+
+  // If this option is set, service will try to prevent creation of large unoptimized segments.
+  // When enabled, updates may be delayed if there are unoptimized segments larger than indexing threshold.
```
timvisee:
```suggestion
  // When enabled, updates may be blocked at request level if there are unoptimized segments larger than indexing threshold.
```

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -103,6 +126,62 @@ impl UpdateWorkers {
...
+            let can_proceed = tokio::task::spawn_blocking(move || {
+                let segments = locked_segments.read();
+                let largest_unoptimized_segment_size_opt =
+                    get_largest_unindexed_segment_vector_size(&segments);
+
+                let largest_unoptimized_segment_size =
+                    largest_unoptimized_segment_size_opt.unwrap_or(0);
+
+                // True, if we can proceed with updates
+                largest_unoptimized_segment_size <= optimization_threshold
```
timvisee, line 159: Isn't there an edge case where our current update will make it go over the threshold - in which case we have linking again?

Case:
- segment is 9kb
- update adds 2gb
- segment is 11kb and disappears in `indexed_only=true` search

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -103,6 +126,62 @@ impl UpdateWorkers {
...
+                let optimizations_guard = optimization_handles.lock().await;
+                if optimizations_guard.iter().all(|h| h.is_finished()) {
+                    return Ok(());
+                }
+            }
+
+            // If unoptimized segments are too large, the only way it can be fixed is optimization
+            // So we wait the notification of optimization completion to re-check the sizes
+            log::debug!("waiting for optimization to allow updates");
+            if let Err(err) = optimization_finished_receiver.changed().await {
```
timvisee, line 179: This has a data race!

Because we use `send()`, `changed()` does not trigger if the signal it sent between the above `can_proceed` and this `changed()` barrier.

There's to ways to solve it:
1. use [`send_modify()`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.send_modify) to also trigger this barrier if it was sent before
2. move the `changed()` call before the `can_proceed` check and only await here

I believe the second solution is a bit better, potentially skipping some extra loops.

Relevant docs:
- https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.send
- https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.send_modify
- https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.changed
- https://docs.rs/tokio/latest/tokio/sync/watch/index.html#change-notifications

lib/collection/src/update_workers/optimization_worker.rs
```diff
@@ -235,6 +238,10 @@ impl UpdateWorkers {
             optimizer_resource_budget,
             segments.clone(),
             move || {
+                // Notify other components that optimization is finished
+                // We do not care if there are no receivers or if they are lagging behind
+                let _ = optimization_finished_sender.send(());
```
timvisee, line 244: Potential fix for [this](https://github.com/qdrant/qdrant/pull/7643#discussion_r2686375216) using [`send_modify()`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.send_modify):

```suggestion
                optimization_finished_sender.send_modify(());
```

timvisee (changes requested):

lib/collection/src/update_workers/update_worker.rs
agourlay: I am unsure about the data race, the `changed` doc says.

```
If the current value in the channel has not yet been marked seen when this method is called, the method marks that value seen and returns immediately. 
```

My understanding is that the `watcher` still has the value marked has unread which we will get when calling `changed` afterwards.

What am I missing here?

timvisee: I think you're right. The working is a little bit confusing. Both `send` and `send_modify` state "notifying all receivers." which refers to a change notification marking values as unseen.

Thank you for watching along with me! :smiley:

lib/collection/src/update_workers/optimization_worker.rs
timvisee: Not necessary anymore

https://github.com/qdrant/qdrant/pull/7643#discussion_r2686718380

timvisee (commented):
(removing 'changes required' marker, though one remark is still pending)

lib/collection/src/update_workers/update_worker.rs
generall: yes, the same problem with mass updates, we can't really estimate volume before execution

timvisee: Naive suggestion: how about a boolean marker to _still_ include it in `indexed_only=true` search, even if slightly over the threshold?

E.g. set the flag for updates like this, and unset it for regular updates. Then multiply the full scan threshold by two if the flag is set for this segment. Or just include it unconditionally.

lib/collection/src/update_workers/update_worker.rs
IvanPleshkov: I have another concern about it. What if `send` fails? Regarding to doc:
https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.send
`when send fails, the value isn’t made available for future receivers (but returned with the [SendError](https://docs.rs/tokio/latest/tokio/sync/watch/error/struct.SendError.html))`
If I understood it correctly, in case of `Err` in `send`, the deadlock appears.

Possible solution: check if `send` returned an error and close the channel in case of an error to trigger `changed()`.

IvanPleshkov: I'm asking about `Err` case because in the optimizer callback `Err` cases are covered. It means that errors while `send` must be covered too.
https://github.com/qdrant/qdrant/blob/master/lib/collection/src/update_workers/optimization_worker.rs#L242

timvisee: It should never fail, until we tear down.

We don't need to wait on `changed()` for the send to succeed. It's fine to just hold the receiver object in memory. We [keep](https://github.com/qdrant/qdrant/blob/a6de38c4a4b90eee0c4e6f07d9bc9ff928e8309a/lib/collection/src/update_workers/update_worker.rs#L35) it for the duration of the update worker.

lib/shard/src/segment_holder/mod.rs
```diff
@@ -117,6 +117,12 @@ impl SegmentHolder {
         })
     }
 
+    /// Iterate over all appendable segments with their IDs
+    /// Do not include non-appendable segments.
+    pub fn iter_appendable(&self) -> impl Iterator<Item = (&SegmentId, &LockedSegment)> {
```
IvanPleshkov: This `fn` is unused

timvisee: :exploding_head: I'm surprised the compiler doesn't tell us this

agourlay:
After fixing the [missing conversion](https://github.com/qdrant/qdrant/pull/7643/commits/9b7933f25c233adb0e99f58a1be65add7f07e3ed) on the threshold, we get a better throughput.

previous https://github.com/qdrant/qdrant/pull/7643#issuecomment-3739587601

| prevent_unoptimized | Upload Time | Wait Indexing Time (bfb) |
|--------------------|------------|---------------|
| false (default)     | 7 seconds  | 10 seconds    |
| true                | 1m 35s     | 3 seconds     |

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -103,6 +126,62 @@ impl UpdateWorkers {
...
+                // this can be if optimization is cancelled, we don't need to wait anymore
+                log::debug!("Optimization thread terminated with an error: {err}");
+                return Ok(());
+            }
+        }
+    }
```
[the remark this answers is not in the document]
agourlay: thank you very much :rabbit: 
https://github.com/qdrant/qdrant/pull/7643/commits/9b7933f25c233adb0e99f58a1be65add7f07e3ed

timvisee (approved):
Looks good.

We are aware the blinking problem still exists in this PR, but agreed to work on that in a separate PR to not stall this one further.

PR #7665, opened 2025-12-01
activate `fs-err` features for nicer error messages

coszio:
In the new version of `fs-err` we can opt-in to more descriptive error messages for **dev builds**. See #7662

### DIFF
--- a/lib/collection/Cargo.toml
+++ b/lib/collection/Cargo.toml
@@ -20,6 +20,7 @@ staging = []
 
 [dev-dependencies]
 criterion = { workspace = true }
+fs-err = { workspace = true, features = ["debug"] }
 proptest = { workspace = true }
 rstest = { workspace = true }
 tar = { workspace = true }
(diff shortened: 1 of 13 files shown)


xzfc:
Relevant PR with examples: https://www.github.com/andrewhickman/fs-err/pull/81

lib/collection/Cargo.toml
```diff
@@ -20,6 +20,7 @@ staging = []
 
 [dev-dependencies]
 criterion = { workspace = true }
+fs-err = { workspace = true, features = ["debug"] }
```
monoid, line 23: Can we just add it to the root `Cargo.toml`?

coszio: no, because there is no `[workspace.dev-dependencies]`. 

An alternative is to use an alias in workspace dependencies (`fs-err-dev = { version = ..., package = "fs-err", features = ["debug"] }`, not sure if it actually works), but it still needs to be added in dev-dependencies for each subcrate.

PR #7669, opened 2025-12-02
SQ with metadata as enum

IvanPleshkov:
This PR is a refector of SQ, where we want to add new encodings.

New changes:

# Metadata is a enum
Before:
```
#[derive(Serialize, Deserialize)]
struct Metadata {
//....
}
```

After:
```
#[derive(Serialize, Deserialize)]
#[serde(untagged)]
enum Metadata {
    Int8(MetadataInt8),
}

#[derive(Serialize, Deserialize)]
struct MetadataInt8 {
//....
}
```

This change introduces an easy way to add a new encoding method.

Everything else is a change to fix the compilation. Or minor refactors like removing logic duplication

### DIFF
--- a/lib/quantization/src/encoded_vectors_u8.rs
+++ b/lib/quantization/src/encoded_vectors_u8.rs
@@ -18,6 +18,8 @@ use crate::encoded_vectors::{
 use crate::quantile::{find_min_max_from_iter, find_quantile_interval};
 
 pub const ALIGNMENT: usize = 16;
+// Each encoded vector stores an additional f32 at the beginning. Define it's size here.
+const ADDITIONAL_CONSTANT_SIZE: usize = std::mem::size_of::<f32>();
 
 #[derive(Clone, PartialEq, Debug)]
 pub enum ScalarQuantizationMethod {
@@ -37,12 +39,95 @@ pub struct EncodedQueryU8 {
 }
 
 #[derive(Serialize, Deserialize)]
-struct Metadata {
+#[serde(untagged)]
+enum Metadata {
+    Int8(MetadataInt8),
+}
+
+impl Metadata {
+    pub fn vector_parameters(&self) -> &VectorParameters {
+        match self {
+            Metadata::Int8(meta) => &meta.vector_parameters,
+        }
+    }
+
+    pub fn actual_dim(&self) -> usize {
+        match self {
+            Metadata::Int8(meta) => meta.actual_dim,
+        }
+    }
+
+    pub fn postprocess_score(&self, score: f32, query_offset: f32, vector_offset: f32) -> f32 {
+        match self {
+            Metadata::Int8(metadata) => {
+                metadata.postprocess_score(score, query_offset, vector_offset)
+            }
+        }
+    }
+
+    pub fn postprocess_internal_score(
+        &self,
+        score: f32,
+        query_offset: f32,
+        vector_offset: f32,
+    ) -> f32 {
+        match self {
+            Metadata::Int8(metadata) => {
+                metadata.postprocess_internal_score(score, query_offset, vector_offset)
+            }
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+struct MetadataInt8 {
     actual_dim: usize,
     alpha: f32,
     offset: f32,
     multiplier: f32,
     vector_parameters: VectorParameters,
+}
+
+impl MetadataInt8 {
+    #[inline]
+    pub fn encode_value(&self, value: f32) -> u8 {
+        let i = (value - self.offset) / self.alpha;
+        i.clamp(0.0, 127.0).round() as u8
+    }
+
+    #[inline]
+    fn postprocess_score(&self, score: f32, query_offset: f32, vector_offset: f32) -> f32 {
+        self.multiplier * score + query_offset + vector_offset
+    }
+
+    #[inline]
+    fn postprocess_internal_score(
+        &self,
+        score: f32,
+        vector_1_offset: f32,
+        vector_2_offset: f32,
+    ) -> f32 {
+        let query_offset = vector_1_offset - self.get_shift();
+        self.postprocess_score(score, query_offset, vector_2_offset)
+    }
+
+    fn get_shift(&self) -> f32 {
+        // Dotprod after shifting produces a number which is not related to vector and query
+        // (x - a)(y - a) = xy - ax - ay + a^2
+        // this a^2 is returned here
+        // L2 is handled the same way as Dot here
+        let shift = match self.vector_parameters.distance_type {
+            DistanceType::Dot | DistanceType::L2 => {
+                self.actual_dim as f32 * self.offset * self.offset
+            }
+            DistanceType::L1 => 0.0,
+        };
+        if self.vector_parameters.invert {
+            -shift
+        } else {
+            shift
+        }
+    }
 }
 
 impl<TStorage: EncodedStorage> EncodedVectorsU8<TStorage> {
@@ -65,13 +150,13 @@ impl<TStorage: EncodedStorage> EncodedVectorsU8<TStorage> {
         let actual_dim = Self::get_actual_dim(vector_parameters);
 
         if count == 0 {
-            let metadata = Metadata {
+            let metadata = Metadata::Int8(MetadataInt8 {
                 actual_dim,
                 alpha: 0.0,
                 offset: 0.0,
                 multiplier: 0.0,
                 vector_parameters: vector_parameters.clone(),
-            };
+            });
             if let Some(meta_path) = meta_path {
                 meta_path
                     .parent()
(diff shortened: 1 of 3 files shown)


lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -225,70 +260,183 @@ impl<TStorage: EncodedStorage> EncodedVectorsU8<TStorage> {
...
-                }
-                DistanceType::L1 => {
-                    impl_score_l1_neon(q_ptr, v_ptr, self.metadata.actual_dim as u32)
+                let (vector_offset, v_ptr) = Self::parse_vec_data(bytes);
+                let q_ptr = query.encoded_query.as_ptr();
+
+                unsafe {
+                    let score = match metadata.vector_parameters.distance_type {
+                        DistanceType::Dot | DistanceType::L2 => {
+                            impl_score_dot_neon(q_ptr, v_ptr, metadata.actual_dim as u32)
```
monoid: I think unsafe's scope should be limited to particular call (here and elsewhere).

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -347,71 +495,73 @@ impl<TStorage: EncodedStorage> EncodedVectorsU8<TStorage> {
...
                 query.push(encoded);
             }
         }
-        let offset = match self.metadata.vector_parameters.distance_type {
+        let offset = match metadata.vector_parameters.distance_type {
             DistanceType::Dot => {
-                query.iter().map(|&x| f32::from(x)).sum::<f32>()
-                    * self.metadata.alpha
-                    * self.metadata.offset
+                query.iter().map(|&x| f32::from(x)).sum::<f32>() * metadata.alpha * metadata.offset
```
monoid: It worth creating a variable for the iter/sum expression.

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -433,109 +583,65 @@ impl<TStorage: EncodedStorage> EncodedVectors for EncodedVectorsU8<TStorage> {
...
-        self.metadata.multiplier * score as f32 + offset
+        self.score_point_simple_internal(i, j)
     }
 
     fn quantized_vector_size(&self) -> usize {
         // actual_dim rounds up vector_dimension to the next multiple of ALIGNMENT
         // also add scaling factor to the tally
-        self.metadata.actual_dim + std::mem::size_of::<f32>()
+        match &self.metadata {
+            Metadata::Int8(_) => self.metadata.actual_dim() + std::mem::size_of::<f32>(),
```
monoid: It worth defining a named constant for `std::mem::size_of::<f32>()`: `METADATA_HEADER_BYTES` or so.

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -433,109 +583,65 @@ impl<TStorage: EncodedStorage> EncodedVectors for EncodedVectorsU8<TStorage> {
...
-        };
+        match &self.metadata {
+            Metadata::Int8(metadata) => {
+                let offset_difference = match metadata.vector_parameters.distance_type {
+                    DistanceType::Dot => {
+                        metadata.actual_dim as f32 * metadata.offset * metadata.offset
+                    }
+                    DistanceType::L1 => 0.0,
+                    DistanceType::L2 => {
+                        metadata.actual_dim as f32 * metadata.offset * metadata.offset
```
monoid: Same as `DistanceType::Dot`, worth merging.

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -31,20 +31,55 @@ pub struct EncodedVectorsU8<TStorage: EncodedStorage> {
...
     actual_dim: usize,
     alpha: f32,
     offset: f32,
     multiplier: f32,
     vector_parameters: VectorParameters,
 }
 
+impl MetadataInt8 {
+    pub fn get_internal_score_diff(&self) -> f32 {
+        let diff = self.actual_dim as f32 * self.offset * self.offset;
```
monoid: It's hard to understand why offset is squared here.

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -225,70 +260,183 @@ impl<TStorage: EncodedStorage> EncodedVectorsU8<TStorage> {
     }
 
     pub fn score_point_simple(&self, query: &EncodedQueryU8, bytes: &[u8]) -> f32 {
-        let (vector_offset, v_ptr) = Self::parse_vec_data(bytes);
-        let q_ptr = query.encoded_query.as_ptr();
+        match query {
+            EncodedQueryU8::Int8(query) => {
+                let Metadata::Int8(metadata) = &self.metadata;
+
+                let (vector_offset, v_ptr) = Self::parse_vec_data(bytes);
```
monoid, line 324: "Offset" sounds like something memory address related, not a float value.  It is out of scope of this PR, but at least it worth discussing.

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -31,20 +31,55 @@ pub struct EncodedVectorsU8<TStorage: EncodedStorage> {
...
+impl MetadataInt8 {
+    pub fn get_internal_score_diff(&self) -> f32 {
+        let diff = self.actual_dim as f32 * self.offset * self.offset;
+        if self.vector_parameters.invert {
+            -diff
+        } else {
+            diff
+        }
+    }
+}
```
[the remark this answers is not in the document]
IvanPleshkov: fixed, thanks

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -347,64 +519,50 @@ impl<TStorage: EncodedStorage> EncodedVectorsU8<TStorage> {
...
                 query.push(encoded);
             }
         }
-        let offset = match self.metadata.vector_parameters.distance_type {
+        let offset = match metadata.vector_parameters.distance_type {
             DistanceType::Dot => {
-                query.iter().map(|&x| f32::from(x)).sum::<f32>()
-                    * self.metadata.alpha
-                    * self.metadata.offset
+                let query_elemets_sum = query.iter().map(|&x| f32::from(x)).sum::<f32>();
```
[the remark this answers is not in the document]
IvanPleshkov: fixed

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: Fixed, thanks!

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: I added a contant in method using constant `ADDITIONAL_CONSTANT_SIZE`, metadata name is used for the quantization config, not for an additional data of a quantized vector

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: Fixed

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: Added a comment and moved all squares into a single method: `MetadataInt8::get_shift`

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: Regarding to article with calculus explanation
https://qdrant.tech/articles/scalar-quantization/?q=scalar+quantization
I propose to keep it as in an article: offset

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -37,14 +37,89 @@ pub struct EncodedQueryU8 {
...
     alpha: f32,
     offset: f32,
     multiplier: f32,
     vector_parameters: VectorParameters,
 }
 
+impl MetadataInt8 {
+    pub fn encode_value(&self, value: f32) -> u8 {
+        let i = (value - self.offset) / self.alpha;
+        i.clamp(0.0, 127.0) as u8
```
monoid: Which rounding mode does `f64 as u8` use?

monoid: Also, worth making `#[inline]`.

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -37,14 +37,89 @@ pub struct EncodedQueryU8 {
...
+    }
+
+    fn postprocess_score(&self, score: f32, query_offset: f32, vector_offset: f32) -> f32 {
+        self.multiplier * score + query_offset + vector_offset
+    }
+
+    fn postprocess_internal_score(&self, score: f32, query_offset: f32, vector_offset: f32) -> f32 {
+        // We store shift in vector constant, in this case we added it twice, so we need to subtract it once
+        self.multiplier * score + query_offset + vector_offset - self.get_shift()
```
monoid: Can we implement it as `self.postprocess_score(score, query_offset, vector_offset) - self.get_shift()`?

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -225,70 +303,164 @@ impl<TStorage: EncodedStorage> EncodedVectorsU8<TStorage> {
...
+            Metadata::Int8(metadata) => {
+                let (vector_offset, v_ptr) = Self::parse_vec_data(bytes);
+                let q_ptr = query.encoded_query.as_ptr();
 
-        let score = match self.metadata.vector_parameters.distance_type {
-            DistanceType::Dot | DistanceType::L2 => {
-                impl_score_dot(q_ptr, v_ptr, self.metadata.actual_dim)
+                let score = match metadata.vector_parameters.distance_type {
+                    DistanceType::Dot | DistanceType::L2 => {
+                        impl_score_dot(q_ptr, v_ptr, metadata.actual_dim)
```
monoid, line 329: `impl_score_dot` and `impl_score_l1` must be `unsafe` functions even if they are not public.  But it may be a subject of another PR.

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -433,109 +599,52 @@ impl<TStorage: EncodedStorage> EncodedVectors for EncodedVectorsU8<TStorage> {
...
-
-        self.metadata.multiplier * score as f32 + offset
+        self.score_point_simple_internal(i, j)
     }
 
     fn quantized_vector_size(&self) -> usize {
         // actual_dim rounds up vector_dimension to the next multiple of ALIGNMENT
         // also add scaling factor to the tally
-        self.metadata.actual_dim + std::mem::size_of::<f32>()
+        const ADDITIONAL_CONSTANT_SIZE: usize = std::mem::size_of::<f32>();
```
monoid: I think the constant may be useful in more than one place in the file.

IvanPleshkov: Done, moved and added a comment

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: You are right, and we can do this in this PR where we move `impl_score_dot` into a separate file
https://github.com/qdrant/qdrant/pull/7763

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: Good idea! Done

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: Added inlining, added `round` to make it more explicit, thanks!

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -114,6 +199,25 @@ impl<TStorage: EncodedStorage> EncodedVectorsU8<TStorage> {
             (alpha, offset)
         };
 
+        let multiplier = match vector_parameters.distance_type {
+            DistanceType::Dot => alpha * alpha,
+            DistanceType::L1 => alpha,
+            DistanceType::L2 => -2.0 * alpha * alpha,
```
monoid, line 211: I think `-` here worth a comment.

lib/quantization/src/encoded_vectors_u8.rs
```diff
@@ -131,31 +235,31 @@ impl<TStorage: EncodedStorage> EncodedVectorsU8<TStorage> {
...
                 }
             };
             let vector_offset = if vector_parameters.invert {
                 -vector_offset
             } else {
                 vector_offset
             };
+            // apply `a^2` shift
+            let vector_offset = metadata.get_shift() + vector_offset;
             encoded_vector[0..std::mem::size_of::<f32>()]
```
monoid: Is it `ADDITIONAL_CONSTANT_SIZE`?  There are still some occurrences of `size_of`.

monoid (approved):

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: done

lib/quantization/src/encoded_vectors_u8.rs
IvanPleshkov: done

PR #7670, opened 2025-12-02
Metrics cluster working state and last update

JojiiOfficial:
Depends on #7479

Adds the following two new metric families to the metrics API:

```
# HELP cluster_last_update_delta time since last update
# TYPE cluster_last_update_delta gauge
cluster_last_update_delta 1.223929901

# HELP cluster_working_state working state of the cluster
# TYPE cluster_working_state gauge
cluster_working_state{state="working"} 1
cluster_working_state{state="stopped"} 0
```

`cluster_working_state` always has two metrics, one with `state=working` and one `state=stopped`.
The active state exports the value `1`.

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -528,8 +528,8 @@ impl MetricsProvider for ClusterStatusTelemetry {
                 metrics.push_metric(metric_family(
                     "cluster_last_update_timestamp_seconds",
                     "unix timestamp of last update",
-                    MetricType::GAUGE,
-                    vec![gauge(timestamp as f64, &[])],
+                    MetricType::COUNTER,
+                    vec![counter(timestamp as f64, &[])],
                     prefix,
                 ));
 


timvisee:
> cluster_working_state always has two metrics, one with state=working and one state=stopped.
The active state exports the value 1.
If the working state is equal to "stopped", an optional error is provided with an empty string indicating no error.
To make the metrics consistent, an empty value for error is always provided, similarly to a non active state having the value 0.

I'm not sure if this is a conventional way of doing things? Maybe it's better to only report the current one?

> `cluster_last_update_delta`

I suggest to rename this to `cluster_last_update_seconds`, or `cluster_last_update_delta_seconds` or `cluster_last_update_age_seconds`

https://prometheus.io/docs/practices/naming/

src/common/metrics.rs
```diff
@@ -349,7 +351,10 @@ impl MetricsProvider for CollectionsTelemetry {
         let vector_count = vector_count_by_name
             .iter()
             .map(|m| m.get_gauge().get_value())
-            .sum();
+            .sum::<f64>()
+            // The sum of an empty f64 iterator returns `-0`. Since a negative
+            // number of vectors is impossible, taking the absolute value is always safe.
+            .abs();
```
timvisee, line 356: I ported this to <https://github.com/qdrant/qdrant/pull/7678> so we can quickly merge it and include it in the upcoming patch release.

JojiiOfficial:
> I'm not sure if this is a conventional way of doing things? Maybe it's better to only report the current one?

It probably is not a conventional way. It seemed to be more suitable for plotting to me. 


Checking with ChatGPT, it seems to confirm my approach:

```
✅ 1. Is it conventional to represent states like this?

Yes — for multi-state enumerations.

Example (good):
    my_service_state{state="running"} 1
    my_service_state{state="stopped"} 0
    my_service_state{state="error"} 0

⚠️ 2. What about one time series that changes its state label value over time?

    This is **not recommended**.

    Prometheus treats each unique label set as a separate time series.
    If the value of the state label changes (e.g. xy → yz → qq), you are:

    - creating a new time series every time the state changes
    - causing churn, which is bad for performance
    - making long-term trend queries messy
    - losing the ability to track how long each state existed (unless you manually reconstruct it)

Prometheus best practices:
👉 Label values should not have high cardinality or change frequently.
```

An alternative could be to include the state in the metrics name:
```
cluster_working_state_working{} 1
cluster_working_state_stopped{error=""} 0
```

I know ChatGPT could be wrong here so please let me know if you want me to look deeper into the conventions here!

Edit: On a second thought, maybe we shouldn't include the error itself for the same reason. I assume the error, if exists, can be read somewhere else too.

timvisee:
> Edit: On a second thought, maybe we shouldn't include the error itself for the same reason. I assume the error, if exists, can be read somewhere else too.

Yeah, that sounds reasonable!

I'm fine with that ChatGPT explanation.

Let's exclude the error message, but just denote the succesful and error states. That would be enough to detect 'an error'. And then a user is responsible for looking into the cluster to see the actual error.

src/common/metrics.rs
```diff
@@ -515,6 +517,52 @@ impl MetricsProvider for ClusterStatusTelemetry {
                 prefix,
             ));
         }
+
+        // Initialize all states so that every state has a zeroed metric by default.
+        let mut state_working = 0.0;
+        let mut state_stopped = 0.0;
+        let mut error = None::<String>;
```
timvisee: Let's remove the `error` strings.

JojiiOfficial: Done in https://github.com/qdrant/qdrant/pull/7670/commits/6e1b82117de53d486ae70dd69e9c8684983a2ffc

src/common/metrics.rs
```diff
@@ -515,6 +517,43 @@ impl MetricsProvider for ClusterStatusTelemetry {
...
+                let delta = Utc::now() - last_update;
+                let delta_secs = delta.as_seconds_f64();
+
+                metrics.push_metric(metric_family(
+                    "cluster_last_update_seconds",
+                    "time since last update",
+                    MetricType::GAUGE,
+                    vec![gauge(delta_secs, &[])],
+                    prefix,
+                ));
```
timvisee, line 534: According to a quick ChatGPT search, Prometheus prefers a timestamp, not an age in seconds.

So I guess it's best to change it to be a Unix timestamp: `cluster_last_update_timestamp_seconds`

JojiiOfficial: Done in https://github.com/qdrant/qdrant/pull/7670/commits/b905e4ffbe8203cbbe7c259d84cddae12e22d187

timvisee: Maybe a counter is more reasonable? It only goes up.

JojiiOfficial: Good point! Changed in https://github.com/qdrant/qdrant/pull/7670/commits/ad788f64db97c3362cd89933d5f5a899c330933d

PR #7673, opened 2025-12-02
Fix shard clean up on snapshot restore without manifest

timvisee:
Also gracefully stop local shard if we:
- recover shard from snapshot
- don't provide snapshot manifest

Should fix crasher run:
- https://github.com/qdrant/crasher/actions/runs/19861366275

### All Submissions:

### DIFF
--- a/lib/collection/src/shards/replica_set/snapshots.rs
+++ b/lib/collection/src/shards/replica_set/snapshots.rs
@@ -235,7 +235,7 @@ impl ShardReplicaSet {
                 shard.stop_gracefully().await;
 
                 None
-            },
+            }
             None if snapshot_manifest.is_empty() => None,
 
             Some(shard) => {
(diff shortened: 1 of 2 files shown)


lib/collection/src/shards/replica_set/snapshots.rs
```diff
@@ -230,7 +230,13 @@ impl ShardReplicaSet {
         }
 
         let local_manifest = match local.take() {
-            _ if snapshot_manifest.is_empty() => None,
+            Some(shard) if snapshot_manifest.is_empty() => {
+                // Shard is no longer needed and can be dropped
+                shard.stop_gracefully().await;
+
+                None
+            }
```
timvisee, line 238: Here's a branch where we also dropped `shard` (`LocalShard`). So we should gracefully stop.

PR #7678, opened 2025-12-03
Fix vector count in metrics showing minus zero

timvisee:
Picked from <https://github.com/qdrant/qdrant/pull/7670> to allow including it in the patch release

Prevent showing `-0.0` in metrics if we have no vectors.

Documentation: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.sum

### All Submissions:

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -349,7 +349,10 @@ impl MetricsProvider for CollectionsTelemetry {
         let vector_count = vector_count_by_name
             .iter()
             .map(|m| m.get_gauge().get_value())
-            .sum();
+            .sum::<f64>()
+            // The sum of an empty f64 iterator returns `-0`. Since a negative
+            // number of vectors is impossible, taking the absolute value is always safe.
+            .abs();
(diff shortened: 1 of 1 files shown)


agourlay (approved):
I was super surprised by the behavior but I did get that -0 on the Rust playground 🫨

timvisee:
> I was super surprised by the behavior but I did get that -0 on the Rust playground 🫨

There's a ""good"" reason for it.

`sum()` uses fold and needs some default value to start summing.

It [picked](https://github.com/rust-lang/rust/blob/672388edbee9e93c35e5fdf7dac818a6612a5103/library/core/src/iter/traits/accum.rs#L162) `-0.0` for floats to keep the sign bit if summing `-0.0` items:
- `0.0` + `0.0` = `0.0`
- `-0.0` + `0.0` = `0.0`
- `0.0` + `-0.0` = `0.0`
- `-0.0` + `-0.0` = `-0.0` (we only keep sign bit here)

PR #7681, opened 2025-12-03
Cleanup a few clones

agourlay:
Removing a few unnecessary clones :broom:

### DIFF
--- a/lib/collection/src/collection/clean.rs
+++ b/lib/collection/src/collection/clean.rs
@@ -206,7 +206,7 @@ impl ShardCleanTask {
 
         ShardCleanTask {
             handle: task,
-            status: receiver.clone(),
+            status: receiver,
             cancel: cancel.drop_guard(),
         }
     }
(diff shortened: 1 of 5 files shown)


JojiiOfficial (approved):

xzfc:
BTW, clippy has a lint [clippy::redundant_clone](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone) for that; and perhaps other clone-related lints.

agourlay:
@xzfc thanks, I actually found those instances exactly like that while playing with the `nursery` level in Clippy!
I should have mentioned it in the description :saluting_face:

JojiiOfficial:
Does it make senses to enable this (and similar) lints, or is this what you were initially working on?

agourlay:
@JojiiOfficial the lints in the [nursery](https://doc.rust-lang.org/stable/clippy/lints.html#nursery) can have false positives AFAIK so I would not add it to CI yet.

xzfc:
> Does it make senses to enable this (and similar) lints

IMO, `clippy::redundant_clone`, is too annoying to enable it globally. E.g., it wants this change:

```diff
     let acc_f32_freq_3 = do_search(&index_f32.index, freq_query2.clone());
     let acc_f16_freq_3 = do_search(&index_f16.index, freq_query2.clone());
-    let acc_u8_freq_3 = do_search(&index_u8.index, freq_query2.clone());
+    let acc_u8_freq_3 = do_search(&index_u8.index, freq_query2);
```

While it's not a false positive, I feel the initial version is a bit easier to reason/maintain. And probably the compiler optimizes it away with a high chance.

PR #7685, opened 2025-12-03
Improve telemetry timeout for shard size_stats

agourlay:
Follow up on https://github.com/qdrant/qdrant/pull/7607

This handles the `size_stats` information with a timeout.

### DIFF
--- a/src/actix/api/service_api.rs
+++ b/src/actix/api/service_api.rs
@@ -56,11 +56,11 @@ fn telemetry(
             level: details_level,
             histograms: false,
         };
-        let telemetry_collector = telemetry_collector.lock().await;
         let telemetry_data = telemetry_collector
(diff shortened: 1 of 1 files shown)


src/actix/api/service_api.rs
```diff
@@ -60,6 +60,7 @@ fn telemetry(
         let telemetry_data = telemetry_collector
             .prepare_data(&access, detail, params.timeout())
             .await?;
+        drop(telemetry_collector);
```
timvisee: Maybe just?

```suggestion
let telemetry_data = telemetry_collector
    .lock()
    .await
    .prepare_data(&access, detail, params.timeout())
    .await?;

```

(note that this suggestion diff looks broken 🤷‍♂️ )

PR #7687, opened 2025-12-04
Add staging endpoint for testing point operation delays

tellet-q:
Add a new POST /collections/{name}/points/staging endpoint (behind `staging` feature flag) that introduces an artificial delay in the point update pipeline. Useful for testing timeout handling, race conditions, and distributed system behavior.

**Notes**:
1. The operation is currently executed on local shard and is not forwarded to the remote ones. A new grpc call is required to forward to the remote shards and will be done in a separate PR.

**Usage**:
`curl -X POST "http://localhost:6333/collections/test/points/staging" -H "Content-Type: application/json" -d '{"duration": 5.0}'
`

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/shard/src/operations/staging.rs
+++ b/lib/shard/src/operations/staging.rs
@@ -15,6 +15,13 @@ pub struct TestDelayOperation {
 }
 
 impl TestDelayOperation {
+    /// Create a new TestDelayOperation with the given duration in seconds.
+    pub fn new(duration_secs: f64) -> Self {
+        Self {
+            duration: OrderedFloat(duration_secs),
+        }
+    }
+
     /// Execute the delay operation (blocking).
     pub fn execute(&self) {
         let duration_secs = self.duration.into_inner();
--- a/lib/storage/src/rbac/ops_checks.rs
+++ b/lib/storage/src/rbac/ops_checks.rs
@@ -727,12 +727,9 @@ mod tests_ops {
             }
             #[cfg(feature = "staging")]
             PointOperationsDiscriminants::TestDelay => {
-                use ordered_float::OrderedFloat;
-                use shard::operations::point_ops::TestDelayOperation;
+                use shard::operations::staging::TestDelayOperation;
                 let op = CollectionUpdateOperations::PointOperation(PointOperations::TestDelay(
-                    TestDelayOperation {
-                        duration: OrderedFloat(1.0),
-                    },
+                    TestDelayOperation::new(1.0),
                 ));
                 assert_requires_whole_write_access(&op);
             }
--- a/src/actix/api/update_api.rs
+++ b/src/actix/api/update_api.rs
@@ -454,11 +454,11 @@ async fn staging_test_delay(
     dispatcher: web::Data<Dispatcher>,
     collection: Path<CollectionPath>,
     operation: Json<TestDelayRequest>,
+    params: Query<UpdateParams>,
     ActixAccess(access): ActixAccess,
 ) -> impl Responder {
     use collection::operations::point_ops::PointOperations;
     use collection::operations::verification::new_unchecked_verification_pass;
-    use ordered_float::OrderedFloat;
     use shard::operations::CollectionUpdateOperations;
     use shard::operations::staging::TestDelayOperation;
 
@@ -466,9 +466,7 @@ async fn staging_test_delay(
     let operation = operation.into_inner();
     let collection_name = collection.into_inner().name;
 
-    let point_operation = PointOperations::TestDelay(TestDelayOperation {
-        duration: OrderedFloat(operation.duration),
-    });
+    let point_operation = PointOperations::TestDelay(TestDelayOperation::new(operation.duration));
 
     let collection_operation = CollectionUpdateOperations::PointOperation(point_operation);
 
@@ -481,10 +479,7 @@ async fn staging_test_delay(
         &collection_name,
         collection_operation,
         InternalUpdateParams::default(),
-        UpdateParams {
-            wait: true,
-            ordering: Default::default(),
-        },
+        params.into_inner(),
         None, // shard_key
         access,
         HwMeasurementAcc::disposable(),


tellet-q:
Example log: 

Screenshot from 2025-12-04 10-52-46

lib/shard/src/operations/staging.rs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+impl TestDelayUpsertPointsOperation {
+    /// Execute the delay operation (blocking).
+    pub fn execute(&self) {
+        let duration_secs = self.duration.into_inner();
+        log::debug!("TestDelayUpsertPoints: sleeping for {duration_secs}s");
+        std::thread::sleep(Duration::from_secs_f64(duration_secs));
+        log::debug!("TestDelayUpsertPoints: finished sleeping");
+    }
```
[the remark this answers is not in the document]
timvisee: Fair concern. Fine with Rabbit's suggestion.

@tellet-q

lib/collection/src/operations/point_ops.rs
```diff
@@ -79,6 +79,10 @@ impl SplitByShard for PointOperations {
                 #[cfg(not(debug_assertions))]
                 OperationToShard::by_shard(vec![])
             }
+            #[cfg(feature = "staging")]
+            test_delay @ PointOperations::TestDelayUpsertPoints(_) => {
+                OperationToShard::to_all(test_delay)
+            }
```
timvisee: Yeah, so here we _could_ do shard routing later if we like

lib/collection/src/shards/remote_shard.rs
```diff
@@ -301,6 +301,11 @@ impl RemoteShard {
                         )?;
                         Update::Sync(request)
                     }
+                    #[cfg(feature = "staging")]
+                    PointOperations::TestDelayUpsertPoints(_) => {
+                        // Staging test delay operations should not be forwarded to remote shards
+                        continue;
+                    }
```
timvisee, line 308: Good call.

Otherwise it's likely to receive the operation twice on the other end. Not an issue for idempotent operations, but it is an issue for this delay.

lib/collection/src/operations/generalizer/update_persisted.rs
```diff
@@ -64,6 +64,10 @@ impl Generalizer for PointOperations {
             PointOperations::SyncPoints(sync_operation) => {
                 PointOperations::SyncPoints(sync_operation.remove_details())
             }
+            #[cfg(feature = "staging")]
+            PointOperations::TestDelayUpsertPoints(op) => {
```
timvisee: This isn't strictly an upsert operation. I vote to change it from `TestDelayUpsertPoints` to `TestDelay`.

I think it'd be even better to move it one level up, into [`CollectionUpdateOperations`](https://github.com/qdrant/qdrant/blob/094aa5c19418cdad352196fb0f132ec8c64a2bce/lib/shard/src/operations/mod.rs#L15-L23). We can do that in a separate PR.

lib/collection/src/shards/remote_shard.rs
```diff
@@ -571,6 +576,22 @@ impl RemoteShard {
                     .await?
                     .into_inner()
                 }
+                #[cfg(feature = "staging")]
+                PointOperations::TestDelayUpsertPoints(op) => {
+                    // TODO: Add gRPC support to forward staging operations to remote shards
+                    // For now, staging test delay operations only execute on local shards
+                    log::debug!(
+                        "TestDelayUpsertPoints: skipping remote shard {} (duration: {}s)",
+                        shard_id.unwrap_or(0),
```
timvisee:
```suggestion
                        self.id,
```

lib/shard/src/operations/point_ops.rs
```diff
@@ -105,8 +105,14 @@ pub enum PointOperations {
...
     /// Points Sync
     SyncPoints(PointSyncOperation),
+    /// Introduce artificial delay for testing purposes
+    #[cfg(feature = "staging")]
+    TestDelayUpsertPoints(super::staging::TestDelayUpsertPointsOperation),
 }
 
+#[cfg(feature = "staging")]
+pub use super::staging::TestDelayUpsertPointsOperation;
+
```
timvisee: This import is obsolete. We already directly target the type in the above enum variant.

```suggestion
```

lib/storage/src/rbac/ops_checks.rs
```diff
@@ -725,6 +725,10 @@ mod tests_ops {
                 ));
                 assert_requires_whole_write_access(&op);
             }
+            #[cfg(feature = "staging")]
+            PointOperationsDiscriminants::TestDelayUpsertPoints => {
+                // Staging operation - requires write access
+            }
```
timvisee, line 735: I guess we still need to actually implement the check here?

lib/shard/src/operations/staging.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub duration: OrderedFloat<f64>,
+}
+
+impl TestDelayUpsertPointsOperation {
+    /// Execute the delay operation (blocking).
+    pub fn execute(&self) {
+        let duration_secs = self.duration.into_inner();
+        log::debug!("TestDelayUpsertPoints: sleeping for {duration_secs}s");
+        std::thread::sleep(Duration::from_secs_f64(duration_secs));
+        log::debug!("TestDelayUpsertPoints: finished sleeping");
```
timvisee: The `Duration` type itself it quite good at formatting time, I suggest to use that.

Demo: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=029f33c3a44399848df807365dab9d3c

```suggestion
        let duration = Duration::from_secs_f64(self.duration.into_inner().max(0.0));
        log::debug!("TestDelayUpsertPoints: sleeping for {duration:?}");
        std::thread::sleep(duration);
        log::debug!("TestDelayUpsertPoints: finished sleeping");
```

This also applies Rabbit's fix.

src/actix/api/update_api.rs
```diff
@@ -430,6 +430,70 @@ async fn delete_field_index(
...
+}
+
+/// Staging endpoint that introduces an artificial delay for testing purposes.
+/// Only available when the `staging` feature is enabled.
+#[cfg(feature = "staging")]
+#[post("/collections/{name}/points/staging")]
+async fn staging_test_delay(
+    dispatcher: web::Data<Dispatcher>,
+    collection: Path<CollectionPath>,
+    operation: Json<TestDelayRequest>,
```
timvisee, line 456: I'd be better to accept a PointOperation type here. Or a dedicated type for staging stuff once we've added a variant in `CollectionUpdateOperations`.

It'll be a bit more verbose in usage, but it also means all new operations we add are immediately usable without implementing a new API.

lib/collection/src/operations/generalizer/update_persisted.rs
tellet-q: Renamed to TestDelay.

> I think it'd be even better to move it one level up, into [CollectionUpdateOperations](https://github.com/qdrant/qdrant/blob/094aa5c19418cdad352196fb0f132ec8c64a2bce/lib/shard/src/operations/mod.rs#L15-L23). We can do that in a separate PR.

I agree that this should go into separate PR.

lib/storage/src/rbac/ops_checks.rs
tellet-q: Added a check

src/actix/api/update_api.rs
tellet-q: Let's do it in that separate PR you suggested above.

src/actix/api/update_api.rs
```diff
@@ -430,6 +430,70 @@ async fn delete_field_index(
...
+    let pass = new_unchecked_verification_pass();
+    let toc = dispatcher.toc(&access, &pass);
+
+    let result = crate::common::update::update(
+        toc,
+        &collection_name,
+        collection_operation,
+        InternalUpdateParams::default(),
+        UpdateParams {
+            wait: true,
```
timvisee: I think it's reasonable to grab the `wait` param from API parameters.

It is useful to just block updates, and not the whole test.

Let's use the `UpdateParams` struct and also include ordering: https://github.com/qdrant/qdrant/blob/ec6cce8291be0e135e4c21dbfd9292690c1f0b24/src/actix/api/update_api.rs#L37

PR #7701, opened 2025-12-05
Metrics incoming and outgoing shard transfers

JojiiOfficial:
Depends on #7670

Adds the following two new metrics to the metrics API:

```
# HELP collection_shard_transfer_incoming incoming shard transfers currently running
# TYPE collection_shard_transfer_incoming gauge
collection_shard_transfer_incoming{id="test"} 0

# HELP collection_shard_transfer_outgoing outgoing shard transfers currently running
# TYPE collection_shard_transfer_outgoing gauge
collection_shard_transfer_outgoing{id="test"} 0
```

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -346,7 +346,8 @@ impl CollectionsTelemetry {
                 for transfer in collection.transfers.iter().flatten() {
                     if transfer.to == this_peer_id {
                         incoming_transfers += 1;
-                    } else if transfer.from == this_peer_id {
+                    }
+                    if transfer.from == this_peer_id {
                         outgoing_transfers += 1;
                     }
                 }


src/common/metrics.rs
```diff
@@ -180,8 +184,13 @@ impl MetricsProvider for AppFeaturesTelemetry {
     }
 }
 
-impl MetricsProvider for CollectionsTelemetry {
-    fn add_metrics(&self, metrics: &mut MetricsData, prefix: Option<&str>) {
+impl CollectionsTelemetry {
```
JojiiOfficial, line 187: Since we need access to the peer id of the current instance, I had to change this from a trait to a type function. We don't use this trait for it's trait benefits, so this approach seemed best to me.
If there is a different preferred approach, please let me know.

src/common/metrics.rs
```diff
@@ -323,6 +336,30 @@ impl MetricsProvider for CollectionsTelemetry {
...
+            let mut incoming_transfers = 0;
+            let mut outgoing_transfers = 0;
+
+            if let Some(this_peer_id) = peer_id {
+                for transfer in collection.transfers.iter().flatten() {
+                    if transfer.to == this_peer_id {
+                        incoming_transfers += 1;
+                    } else if transfer.from == this_peer_id {
+                        outgoing_transfers += 1;
+                    }
```
timvisee, line 352: There are some edge cases where the source and target is the same.

So in that case, let's count both.


```suggestion
                    if transfer.to == this_peer_id {
                        incoming_transfers += 1;
                    }
                    if transfer.from == this_peer_id {
                        outgoing_transfers += 1;
                    }
```

src/common/metrics.rs
```diff
@@ -448,6 +485,22 @@ impl MetricsProvider for CollectionsTelemetry {
             snapshots_created_total,
             prefix,
         ));
+
+        metrics.push_metric(metric_family(
+            "collection_shard_transfer_incoming",
```
timvisee, line 491: Maybe `collection_incoming_shard_transfers` is better. I have no idea, and have a hard time deciding 😃

JojiiOfficial: Yeah I also considered both and couldn't decide. I finally went with this name because it preserves the namespace. I think this is more important than the counting unit, which can be deferred from the help text.
But I don't have a strong opinion here, just my reasoning.

timvisee: Good reasoning. Let's keep your version.

timvisee (approved):

PR #7703, opened 2025-12-06
peer-id change on snapshot recovery

generall:
I was not able to find any problems, which this change would fix. But current code definitelly does not look right.


ffuugoo (approved):

timvisee (approved):

KShivendu:
I think this is now covered in https://github.com/qdrant/qdrant/pull/7706 which has been merged. So we can close this.

PR #7706, opened 2025-12-06
refactor replica set

generall:
- **peer-id change on snapshot recovery**
- **move peer state management into a dedicated file**

### DIFF
--- a/lib/collection/src/shards/replica_set/replica_set_state.rs
+++ b/lib/collection/src/shards/replica_set/replica_set_state.rs
@@ -1,0 +1,245 @@
+use std::collections::HashMap;
+
+use schemars::JsonSchema;
+use segment::common::anonymize::Anonymize;
+use serde::{Deserialize, Serialize};
+
+use crate::shards::shard::PeerId;
+
+/// Represents a replica set state
+#[derive(Debug, Deserialize, Serialize, Default, PartialEq, Eq, Clone)]
+pub struct ReplicaSetState {
+    pub is_local: bool,
+    pub this_peer_id: PeerId,
+    peers: HashMap<PeerId, ReplicaState>,
+}
+
+impl ReplicaSetState {
+    pub fn get_peer_state(&self, peer_id: PeerId) -> Option<ReplicaState> {
+        self.peers.get(&peer_id).copied()
+    }
+
+    pub fn set_peer_state(&mut self, peer_id: PeerId, state: ReplicaState) {
+        self.peers.insert(peer_id, state);
+    }
+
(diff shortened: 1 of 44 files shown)


lib/collection/src/shards/replica_set/replica_set_state.rs
```diff
@@ -0,0 +1,245 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::shards::shard::PeerId;
+
+/// Represents a replica set state
+#[derive(Debug, Deserialize, Serialize, Default, PartialEq, Eq, Clone)]
+pub struct ReplicaSetState {
+    pub is_local: bool,
+    pub this_peer_id: PeerId,
+    peers: HashMap<PeerId, ReplicaState>,
```
generall, line 14: this is now private field. So this PR also changes access to be done through `peers()` getter.

lib/collection/src/shards/replica_set/replica_set_state.rs
```diff
@@ -0,0 +1,245 @@
...
+        self.peers.insert(peer_id, state);
+    }
+
+    pub fn remove_peer_state(&mut self, peer_id: PeerId) -> Option<ReplicaState> {
+        self.peers.remove(&peer_id)
+    }
+
+    pub fn peers(&self) -> &HashMap<PeerId, ReplicaState> {
+        &self.peers
+    }
```
generall, line 32: This getter doesn't clone anymore

lib/collection/src/shards/replica_set/replica_set_state.rs
```diff
@@ -0,0 +1,245 @@
...
+        self.peers
+            .remove(&old_peer_id)
+            .and_then(|replica_state| self.peers.insert(new_peer_id, replica_state));
+    }
+
+    /// Remove all remote peers from the replica set state and activate local peer.
+    pub fn force_local_active(&mut self) {
+        self.peers.clear();
+        self.peers.insert(self.this_peer_id, ReplicaState::Active);
+    }
```
generall, line 86: Those are new functions, both only used in snapshot recovery porcess

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -468,14 +469,11 @@ impl ShardReplicaSet {
...
-                                                ReplicaState::Active
-                                                    | ReplicaState::Resharding
-                                                    | ReplicaState::ReshardingScaleDown
-                                            )
-                                        );
+                                        let is_active = state
+                                            .peers()
+                                            .get(peer_id)
+                                            .map(|state| state.can_be_source_of_truth())
+                                            .unwrap_or(false);
```
generall, line 476: This is changed to avoid partial match

PR #7714, opened 2025-12-08
Metrics more shard transfer stats

JojiiOfficial:
Depends on #7701

This PR adds more metrics for shard transfers into the metrics API:

```
# HELP collection_shard_transfer_failed number of failed shard transfers
# TYPE collection_shard_transfer_failed gauge
collection_shard_transfer_failed{id="test"} 0

# HELP collection_shard_transfer_transferred_points number of points already transferred
# TYPE collection_shard_transfer_transferred_points gauge
collection_shard_transfer_transferred_points{id="test"} 0

# HELP collection_shard_transfer_transferred_total_points total number of points to transfer
# TYPE collection_shard_transfer_transferred_total_points gauge
collection_shard_transfer_transferred_total_points{id="test"} 0
```

There are also some very small refactors done for the `TransferTaskStatus` type, which help avoiding duplicated code or locking more than necessary.

### DIFF
--- a/src/common/metrics.rs
+++ b/src/common/metrics.rs
@@ -228,7 +228,6 @@ impl CollectionsTelemetry {
         let mut shard_transfers_out = Vec::with_capacity(num_collections);
         let mut shard_transfers_failed = Vec::with_capacity(num_collections);
         let mut shard_transfers_points_transferred = Vec::with_capacity(num_collections);
-        let mut shard_transfers_points_total = Vec::with_capacity(num_collections);
 
         for collection in self.collections.iter().flatten() {
             let collection = match collection {
@@ -346,7 +345,6 @@ impl CollectionsTelemetry {
             let mut outgoing_transfers = 0;
             let mut failed_transfers = 0;
             let mut points_transferred = 0;
-            let mut points_total = 0;
 
             if let Some(this_peer_id) = peer_id {
                 for transfer in collection.transfers.iter().flatten() {
@@ -371,7 +369,6 @@ impl CollectionsTelemetry {
                     }
 
                     points_transferred += status.map(|i| i.points_transferred).unwrap_or_default();
-                    points_total += status.map(|i| i.points_total).unwrap_or_default();
                 }
             }
 
@@ -389,8 +386,6 @@ impl CollectionsTelemetry {
             ));
             shard_transfers_points_transferred
                 .push(gauge(points_transferred as f64, &[("id", &collection.id)]));
-            shard_transfers_points_total
-                .push(gauge(points_total as f64, &[("id", &collection.id)]));
         }
 
         for snapshot_telemetry in self.snapshots.iter().flatten() {
@@ -546,14 +541,6 @@ impl CollectionsTelemetry {
             "number of points already transferred",
             MetricType::GAUGE,
             shard_transfers_points_transferred,
-            prefix,
-        ));
-
-        metrics.push_metric(metric_family(
-            "collection_shard_transfer_transferred_total_points",
-            "total number of points to transfer",
-            MetricType::GAUGE,
-            shard_transfers_points_total,
             prefix,
         ));
     }


lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -523,14 +524,25 @@ impl ShardHolder {
...
             let status = tasks_pool.get_task_status(&shard_transfer.key());
+
+            let failed = status.map(|i| i.result == TaskResult::Failed);
+            let running = status.map(|i| i.result == TaskResult::Running);
+            let shard_transfer_status = ShardTransferStatus {
+                running,
+                failed,
+                points_transferred: status.map(|i| i.points_transferred),
+                points_total: status.map(|i| i.points_total),
+            };
```
xzfc: Why structure with optional fields rather than optional structure with mandatory fields?

lib/collection/src/operations/types.rs
```diff
@@ -328,6 +328,27 @@ pub struct ShardTransferInfo {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[anonymize(false)]
     pub comment: Option<String>,
+
+    /// Internal status of the shard transfer, do not expose.
+    #[serde(skip)]
+    #[schemars(skip)]
+    #[anonymize(false)]
+    pub status: ShardTransferStatus,
+}
```
xzfc: Nits:
- reword the doc to make it clear that `comment` and `status` contain the same information.
- `#[serde(skip)]` implies `#[schemars(skip)]`, so the latter could be removed.
- Put into option (see comment below).

```suggestion
    /// Same as [`Self::comment`], but in a machine-readable form.
    /// For internal use only, do not expose.
    #[serde(skip)]
    #[anonymize(false)]
    pub status: Option<ShardTransferStatus>,
}
```

lib/collection/src/operations/types.rs
```diff
@@ -328,6 +328,27 @@ pub struct ShardTransferInfo {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[anonymize(false)]
     pub comment: Option<String>,
+
+    /// Internal status of the shard transfer, do not expose.
+    #[serde(skip)]
+    #[schemars(skip)]
```
xzfc: Nits:
- reword the doc to make it clear that `comment` and `status` contain the same information.
- `#[serde(skip)]` implies `#[schemars(skip)]`, so the latter could be removed.

```suggestion
    /// Same as [`Self::comment`], but in a machine-readable form.
    /// For internal use only, do not expose.
    #[serde(skip)]
```

lib/collection/src/operations/types.rs
```diff
@@ -328,6 +328,27 @@ pub struct ShardTransferInfo {
...
+
+    #[anonymize(false)]
+    pub failed: Option<bool>,
+
+    #[anonymize(false)]
+    pub points_transferred: Option<usize>,
+
+    #[anonymize(false)]
+    pub points_total: Option<usize>,
 }
```
xzfc: Why not just reuse `TransferTaskStatus`? And I think optional struct is makes more sense than struct with optional fields.

```suggestion
    pub status: Option<ShardTransferStatus>,
}
```

lib/collection/src/shards/transfer/transfer_tasks_pool.rs
```diff
@@ -60,6 +86,10 @@ impl TransferTaskProgress {
         self.points_total = total;
         self.eta.set_progress(transferred);
     }
+
+    pub fn total(&self) -> usize {
+        max(self.points_total, self.points_transferred)
+    }
```
xzfc: Nit: we don't need the new method if we put cap on the field itself:

```suggestion
        self.points_total = max(total, transferred);
        self.eta.set_progress(transferred);
    }
```

lib/collection/src/shards/transfer/transfer_tasks_pool.rs
```diff
@@ -27,16 +29,40 @@ pub struct TransferTaskProgress {
...
+
+impl TransferTaskStatus {
+    pub fn comment(&self) -> String {
+        let mut comment = format!(
+            "Transferring records ({}/{}), started {}s ago, ETA: ",
+            self.points_transferred,
+            self.points_total,
+            chrono::Utc::now()
+                .signed_duration_since(self.started_at)
+                .num_seconds(),
```
xzfc: Nit: add `elapsed` field to `TransferTaskStatus` to keep this method pure.

Otherwise this structure would be weird: elapsed time grows as time pass, but ETA is frozen in time when this structure is created in `get_task_status()`

lib/collection/src/operations/types.rs
JojiiOfficial: Addressed in [8ad0c0f](https://github.com/qdrant/qdrant/pull/7714/commits/8ad0c0f073351ec874d04318998e990a0a04e6a0)

lib/collection/src/operations/types.rs
JojiiOfficial: Addressed in [8ad0c0f](https://github.com/qdrant/qdrant/pull/7714/commits/8ad0c0f073351ec874d04318998e990a0a04e6a0)

lib/collection/src/shards/transfer/transfer_tasks_pool.rs
JojiiOfficial: Indeed, missed that part. Addressed in [8ad0c0f](https://github.com/qdrant/qdrant/pull/7714/commits/8ad0c0f073351ec874d04318998e990a0a04e6a0)

lib/collection/src/shards/transfer/transfer_tasks_pool.rs
JojiiOfficial: Addressed in [8ad0c0f](https://github.com/qdrant/qdrant/pull/7714/commits/8ad0c0f073351ec874d04318998e990a0a04e6a0)

JojiiOfficial:
> Do we support multiple simultaneous shard transfers? If so, these two gauges would jump up and down as shard transfers got started/completed/cancelled. IMO, the following pair would make more sense:
> 
>     * `counter` total points transferred, included completed. Only grows.
> 
>     * `gauge` remaining amount of points to transfer. Jumps up when the transfer is initiated, then slowly goes down to 0.

I implemented both as gauges because it allows for easier (total) percentage calculation in prometheus. If we have a monotonic increasing total value, it becomes harder to get the currents overall progress. However I agree with your approach depending on how we actually want to use these metrics.

CC @generall

timvisee:
> > Do we support multiple simultaneous shard transfers? If so, these two gauges would jump up and down as shard transfers got started/completed/cancelled. IMO, the following pair would make more sense:
> > ```
> > * `counter` total points transferred, included completed. Only grows.
> > 
> > * `gauge` remaining amount of points to transfer. Jumps up when the transfer is initiated, then slowly goes down to 0.
> > ```
> 
> I implemented both as gauges because it allows for easier (total) percentage calculation in prometheus. If we have a monotonic increasing total value, it becomes harder to get the currents overall progress. However I agree with your approach depending on how we actually want to use these metrics.
> 
> CC @generall

Shard transfers are unique for the following key:

`key { collection_name, shard_id, from_peer_id, to_peer_id }`

It means that for each key, we can have at most 1 transfer.

Would it make sense to include these 4 values as metric attributes so that we report each transfer separately?

I don't see how else we could report progress in a nice way without jumping up and down.

generall:
I don't think we should use telemetry to track transfer progress. Keep in mind that each transfer will be treated as separate chart

xzfc:
> Keep in mind that each transfer will be treated as separate chart

Per-transfer stats still can be useful to display in Grafana dashboards as table with progressbars. See examples here:
https://play.grafana.org/d/U_bZIMRMk/table-panel-showcase?orgId=1&from=now-6h&to=now&timezone=utc
https://habrastorage.org/r/w1560/getpro/habr/upload_files/42e/7fa/aa9/42e7faaa91bd430b9a515150986e2c32.png

Yet, if we decide to add it, I think should be in addition to [what I proposed](https://github.com/qdrant/qdrant/pull/7714#pullrequestreview-3565360136), not instead of it.

timvisee:
As agreed in a daily:
- we keep: failed count, transferred points count
- we remove: total points to transfer (including pending)

We do not report individual transfers [per key](https://github.com/qdrant/qdrant/pull/7714#issuecomment-3654923836).

timvisee (approved):

JojiiOfficial:
> The count reverts to zero once a transfer completes. We should keep the transferred count value until restart.

Can we address that in a separate PR, or after merging #7734 into this PR to avoid (more) merge conflicts?

We could disable this one specific metric for now and enable it after that has been addressed, to keep it consistent.

PR #7715, opened 2025-12-08
Log snapshot download duration

KShivendu:
Logging file size, duration, and download rate makes the system more transparent.  Doing this for full, partial, and s3 snapshot downloads

### DIFF
--- a/lib/storage/src/content_manager/snapshots/download.rs
+++ b/lib/storage/src/content_manager/snapshots/download.rs
@@ -4,6 +4,7 @@ use std::path::Path;
 use common::tempfile_ext::MaybeTempPath;
 use fs_err::tokio as tokio_fs;
 use futures::StreamExt;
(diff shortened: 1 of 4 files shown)


lib/storage/src/content_manager/snapshots/download.rs
```diff
@@ -51,14 +52,30 @@ async fn download_file(
     }
 
     let mut stream = response.bytes_stream();
+    let mut total_bytes_downloaded = 0u64;
```
timvisee, line 56: How about we just do `file.len()` at the end?

KShivendu: Will have to do `file.metadata().await?.len()` which can be unreliable due to flushing and will be a sys call. It's fine like this imo. It will measure some response bytes too, which is okay.

PR #7732, opened 2025-12-09
Fix to run all tests on CI

agourlay:
Reintroduce workspace flag lost in https://github.com/qdrant/qdrant/pull/7552/files#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dL45 in order to run all unit tests.

### DIFF
--- a/.github/workflows/rust-gpu.yml
+++ b/.github/workflows/rust-gpu.yml
@@ -32,10 +32,10 @@ jobs:
         sudo apt-get install clang mesa-vulkan-drivers libvulkan1 vulkan-tools vulkan-validationlayers
         /usr/bin/vulkaninfo --summary
     - name: Build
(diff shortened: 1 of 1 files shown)


lib/segment/src/payload_storage/payload_storage_enum.rs
```diff
@@ -487,7 +487,7 @@ mod tests {
         // loose assertion because value not stable across different OS
         let storage_size = storage.get_storage_size_bytes().unwrap();
         assert!(
-            storage_size > 2000 && storage_size < 2500,
+            storage_size > 2000 && storage_size < 2600,
```
agourlay, line 490: https://github.com/qdrant/qdrant/actions/runs/20057319635/job/57525497656#step:10:1449

PR #7733, opened 2025-12-09
Use p square to find ranges

IvanPleshkov:
Continue of https://github.com/qdrant/qdrant/pull/7520

This PR applies P-Square alg and provides an interface for min-max estimation for vector data per coordinate:
```
fn find_interval_per_coordinate<'a>(
    vector_data: impl Iterator<Item = impl AsRef<[f32]> + 'a> + Clone,
    dim: usize,
    count: usize,
    quantile: f32,
    stopped: &AtomicBool,
) -> Result<Vec<(f32, f32)>, EncodingError>
```

The implementation contains test, bench and it's multithreaded.

It also refactores old method `find_quantile_interval`. It reuses random sample taking and added `stopper`.

### Performance

This PR provides a criterion benchmark. In case of OpenAPI embeddings with 4 thread I got a result of 100ms locally.
```
p_square_vectors/p_square_vectors
                        time:   [115.84 ms 116.05 ms 116.35 ms]
```

Because we take only 5K random vectors to do a min-max estimation, this benchmark represents a real-world scenario.

### DIFF
--- a/lib/quantization/src/quantile.rs
+++ b/lib/quantization/src/quantile.rs
@@ -130,10 +130,6 @@ fn find_interval_per_coordinate_p2<'a>(
     stopped: &AtomicBool,
 ) -> Result<Vec<(f32, f32)>, EncodingError> {
     let selected_vectors = take_random_vectors(vector_data, count, SAMPLE_SIZE, stopped)?;
-    let selected_vectors = selected_vectors
-        .iter()
-        .map(|v| v.as_ref())
-        .collect::<Vec<&[f32]>>();
 
     let pool = rayon::ThreadPoolBuilder::new()
         .thread_name(|idx| format!("p-square-{idx}"))
@@ -163,6 +159,7 @@ fn find_interval_per_coordinate_p2<'a>(
                         return Err(EncodingError::Stopped);
                     }
 
+                    let vector = vector.as_ref();
                     let value = f64::from(vector[d]);
                     min.push(value);
                     max.push(value);


lib/quantization/src/quantile.rs
```diff
@@ -61,11 +64,140 @@ pub(crate) fn find_quantile_interval<'a>(
...
+
+        let vector = vector.as_ref();
+        debug_assert_eq!(
+            vector.len(),
+            dim,
+            "Vector length does not match the expected dimension"
+        );
+
+        for (i, min_max) in result.iter_mut().enumerate() {
+            if vector[i] < min_max.0 {
```
monoid: Perhaps, using `std::cmp::min` here would be clearer.  And `zip` instead of indexing the `vector`.

monoid: Also, the initial value of `f32::MIN` makes this condition true only for minus infinity.

lib/quantization/src/quantile.rs
```diff
@@ -61,11 +64,140 @@ pub(crate) fn find_quantile_interval<'a>(
...
+                min_max.0 = vector[i];
+            }
+            if vector[i] > min_max.1 {
```
monoid: Ditto.

lib/quantization/src/quantile.rs
```diff
@@ -61,11 +64,140 @@ pub(crate) fn find_quantile_interval<'a>(
...
+    let permutor = Permutor::new(count as u64);
+    let mut selected_vectors: Vec<usize> = permutor.map(|i| i as usize).take(slice_size).collect();
+    selected_vectors.sort_unstable();
+
+    (
+        slice_size,
+        vector_data
+            .enumerate()
+            .filter_map(move |(vector_index, vector)| {
+                if selected_vectors.binary_search(&vector_index).is_ok() {
```
monoid: It could be implemented as linear operation because both `selected_vectors`'s values and `vector_index` is increasing.

Upd: actually it was before.

lib/quantization/src/quantile.rs
```diff
@@ -61,11 +64,188 @@ pub(crate) fn find_quantile_interval<'a>(
...
+        for (min, max) in per_coordinate.into_iter() {
+            assert!(
+                (min - (1.0 - quantile).abs() < acc),
+                "Min value is out of expected range"
+            );
+            assert!(
+                (max - quantile).abs() < acc,
+                "Max value is out of expected range"
+            );
+        }
```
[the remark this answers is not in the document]
IvanPleshkov: fixed

lib/quantization/src/quantile.rs
IvanPleshkov: Thanks! Fixed!

lib/quantization/src/quantile.rs
IvanPleshkov: Fixitto )

lib/quantization/src/quantile.rs
IvanPleshkov: Fixed!

lib/quantization/src/quantile.rs
```diff
@@ -61,11 +57,192 @@ pub(crate) fn find_quantile_interval<'a>(
...
+fn find_min_max_interval_per_coordinate<'a>(
+    vector_data: impl Iterator<Item = impl AsRef<[f32]> + 'a> + Clone,
+    dim: usize,
+    count: usize,
+    stopped: &AtomicBool,
+) -> Result<Vec<(f32, f32)>, EncodingError> {
+    let mut result = Vec::with_capacity(dim);
+    for _ in 0..dim {
+        result.push((f32::MAX, f32::MIN));
+    }
```
monoid:
```suggestion
    let result = vec![(f32::MAX, f32::MIN); dim];
```

lib/quantization/src/quantile.rs
```diff
@@ -61,11 +57,192 @@ pub(crate) fn find_quantile_interval<'a>(
...
+    count: usize,
+    sample_size: usize,
+    stopped: &AtomicBool,
+) -> Result<Vec<f32>, EncodingError> {
+    let slice_size = std::cmp::min(count, sample_size);
+    let permutor = Permutor::new(count as u64);
+    let mut selected_vectors: Vec<usize> = permutor.map(|i| i as usize).take(slice_size).collect();
+    selected_vectors.sort_unstable();
+
+    let mut data_slice = Vec::with_capacity(slice_size * dim);
```
monoid: I wonder if we can return a filtered iterator over the original data instead of copying to a buffer.

IvanPleshkov: I cannot use rayon in this case to parallel P2 alg because I cannot share the iterator between threads and keep performing P2 in separatre thread for each coordinate. I find parallelization more important because P2 computational expensive. Moreover, old quantile estimation does this copy because of sorting.
Data copying is not critical here, max 5K points means max `5K * 1636 * 4 = 32.7MB` which I find not critical for the temp buffer

monoid: What does profiling tell us?

lib/quantization/src/quantile.rs
IvanPleshkov: done

lib/quantization/src/quantile.rs
monoid: Can you return an array of references?

IvanPleshkov: I can return a vector of references, and it looks better than a flattened buffer

IvanPleshkov: Done

lib/quantization/src/quantile.rs
```diff
@@ -23,36 +30,27 @@ pub(crate) fn find_quantile_interval<'a>(
...
-            if selected_index == slice_size {
-                break;
-            }
-        }
+    let selected_vectors = take_random_vectors(vector_data, count, SAMPLE_SIZE / dim, stopped)?;
+    let selected_vectors_count = selected_vectors.len();
+    let mut data_slice: Vec<f32> = Vec::with_capacity(selected_vectors_count * dim);
+    for vector in selected_vectors.iter() {
+        data_slice.extend_from_slice(vector.as_ref());
     }
```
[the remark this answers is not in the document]
IvanPleshkov: resolved, division by dim is removed

lib/quantization/src/quantile.rs
```diff
@@ -61,11 +63,199 @@ pub(crate) fn find_quantile_interval<'a>(
...
+    count: usize,
+    stopped: &AtomicBool,
+) -> Result<Vec<(f32, f32)>, EncodingError> {
+    let mut result = vec![(f32::MAX, f32::MIN); dim];
+
+    let selected_vectors = take_random_vectors(vector_data, count, SAMPLE_SIZE, stopped)?;
+    let selected_vectors = selected_vectors
+        .iter()
+        .map(|v| v.as_ref())
+        .collect::<Vec<&[f32]>>();
```
monoid:
```suggestion
```

lib/quantization/src/quantile.rs
```diff
@@ -61,11 +63,199 @@ pub(crate) fn find_quantile_interval<'a>(
...
+    count: usize,
+    quantile: f32,
+    num_threads: usize,
+    stopped: &AtomicBool,
+) -> Result<Vec<(f32, f32)>, EncodingError> {
+    let selected_vectors = take_random_vectors(vector_data, count, SAMPLE_SIZE, stopped)?;
+    let selected_vectors = selected_vectors
+        .iter()
+        .map(|v| v.as_ref())
+        .collect::<Vec<&[f32]>>();
```
monoid: It could be eliminated to, but requires adding `Send + Sync` to impls extensively.

lib/quantization/src/quantile.rs
IvanPleshkov: Done

lib/quantization/src/quantile.rs
IvanPleshkov: Done, it's possible to use `Send + Sync`, placed this impls also into the segment crate where there is an actual usage of quantization to make sure, that `Send + Sync` does not break anything

monoid (approved):

PR #7735, opened 2025-12-09
Expose segment UUIDs in telemetry response

KShivendu:
Segment telemetry now includes segment ID (as UUID) which makes it easy to correlate with segment manifest (for partial snapshot) and understand the node.

```json
"info": {
    "uuid": "bc63eb4b-ce4e-4836-9993-ea62fa0e8e8e", // new
    "segment_type": "plain",
    "num_vectors": 17536,
    ...
}
```

Also, shard optimization logs previously used segment_ids as `Vec<u64>` which was confusing since it was internal ID (in memory) while on the disk you could only see the UUID based file names. It made things hard to debug. So fixed it too. 

```jsonc
{
    "log": [
        {
            "name": "merge",
            "segment_uuids": ["bc63eb4b-ce4e-4836-9993-ea62fa0e8e8e", "e200ce33-6397-4b1f-87dd-b4bc85d43116"],
           "segment_ids": [0, 1],
            "status": "optimizing",
            "start_at": "2025-12-09T12:29:49.458357380Z",
            "end_at": null
        },
        ...
    ]
}
```

### DIFF
--- a/lib/segment/src/segment/snapshot.rs
+++ b/lib/segment/src/segment/snapshot.rs
@@ -116,26 +116,6 @@ impl SnapshotEntry for Segment {
 }
 
 impl Segment {
-    pub fn segment_id(&self) -> OperationResult<&str> {
-        let id = self
-            .segment_path
-            .file_stem()
-            .and_then(|segment_dir| segment_dir.to_str())
-            .ok_or_else(|| {
-                OperationError::service_error(format!(
-                    "failed to extract segment ID from segment path {}",
-                    self.segment_path.display(),
-                ))
-            })?;
-
-        debug_assert!(
-            Uuid::try_parse(id).is_ok(),
-            "segment ID {id} is not a valid UUID",
-        );
-
-        Ok(id)
-    }
-
     fn _get_segment_manifest(&self) -> OperationResult<SegmentManifest> {
         let segment_id = self.segment_id()?;
         let segment_version = self.version();
@@ -210,7 +190,7 @@ impl Segment {
         }
 
         Ok(SegmentManifest {
-            segment_id: segment_id.to_string(),
+            segment_id,
             segment_version,
             file_versions,
         })


lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -576,47 +576,11 @@ impl SegmentEntry for ProxySegment {
...
-            num_points: self.available_point_count(),
-            num_deleted_vectors: wrapped_info.num_deleted_vectors
-                + deleted_points_count * vector_name_count,
-            vectors_size_bytes: wrapped_info.vectors_size_bytes, //  + write_info.vectors_size_bytes,
-            payloads_size_bytes: wrapped_info.payloads_size_bytes,
-            ram_usage_bytes: wrapped_info.ram_usage_bytes,
-            disk_usage_bytes: wrapped_info.disk_usage_bytes,
-            is_appendable: false,
-            index_schema: wrapped_info.index_schema,
-            vector_data,
```
KShivendu, line 641: This implementation seemed redundant and this `.info()` and `.size_info()` function are called only in tests.

timvisee:
Can we do this on shard level instead? That makes more sense to me.

KShivendu:
> Can we do this on shard level instead? That makes more sense to me.

It will be difficult to co-relate which segment ID has which config/info. It's better to keep them together.

Local shard has this:
```jsonc
"segments": [ {"info": <struct-described-above>, "config": {...}, ...}, ... ] 
```

Also, shard optimization logs are like this:

```jsonc
{
    "log": [
        {
            "name": "indexing",
            "segment_ids": [46, 47], // ---> this is even more confusing since it doesn't align with filesystem dir name
            "status": "optimizing",
            "start_at": "2025-12-09T12:29:49.458357380Z",
            "end_at": null
        },
        ...
    ]
}
```

So I'm gonna replace or extend it with proper UUID segment ID.

lib/segment/src/segment/entry.rs
```diff
@@ -512,6 +512,7 @@ impl SegmentEntry for Segment {
     }
 
     fn size_info(&self) -> SegmentInfo {
+        let segment_id = self.segment_id().unwrap_or_else(|_| "").to_string();
```
timvisee: I'd rather pick `None` instead

lib/segment/src/segment/snapshot.rs
```diff
@@ -122,7 +122,7 @@ impl Segment {
     }
 
     fn get_segment_manifest(&self) -> OperationResult<SegmentManifest> {
-        let segment_id = self.segment_id()?;
+        let segment_id = self.segment_id()?; //
```
timvisee:
```suggestion
        let segment_id = self.segment_id()?;
```

lib/shard/src/proxy_segment/segment_entry.rs
timvisee: That might actually be a problem :thinking: 

IIRC a telemetry request on segments that are proxied _should_ report their proxied state. And currently we just bypass it [here](https://github.com/qdrant/qdrant/blob/e01babffab1dbfa1dc512d616ef8378c5cd9c66d/lib/shard/src/proxy_segment/segment_entry.rs#L756-L758).

timvisee: I cannot come up with a good use case for either approach, to be honest. So let's keep it as is :man_shrugging:

JojiiOfficial: Aren't we using it e.g. [here](https://github.com/qdrant/qdrant/blob/dev/lib/collection/src/shards/local_shard/telemetry.rs#L153) or am I missing something?

timvisee: We do! Find references is broken :dizzy_face:

KShivendu: Brought it back.

timvisee:
Could you bump the OpenAPI spec?

lib/segment/src/segment/entry.rs
```diff
@@ -507,11 +508,32 @@ impl SegmentEntry for Segment {
...
+                ))
+            })?;
+
+        debug_assert!(
+            Uuid::try_parse(id).is_ok(),
+            "segment ID {id} is not a valid UUID",
+        );
+
+        Ok(id.to_string())
+    }
```
timvisee: I don't really like this is fallible. Also, this would now try to parse the path on every UUID request.

I suggest to modify [`Segment`](https://github.com/qdrant/qdrant/blob/037c27200f4a0c40d8e0e4f987a33a9905ca3bd9/lib/segment/src/segment/mod.rs#L63) and add a `uuid` field in there. We can populate when loading or building a segment. Then we always have a UUID.

xzfc:
Merging #7783 introduced merge conflicts in this branch, so I rebased it.

lib/shard/src/segment_holder/mod.rs
```diff
@@ -96,6 +96,24 @@ impl SegmentHolder {
...
+                    .or_else(|| self.non_appendable_segments.get(id))
+                    .map(|segment| segment.get().read().segment_uuid())
+                    .or_else(|| {
+                        // Shouldn't happen
+                        log::warn!("Failed to get UUID for segment with internal id {id}");
+                        None
+                    })
+            })
+            .collect::<Vec<_>>()
+    }
```
xzfc, line 134: Nit: I'd rather inline this function into the caller, e.g.:

```rust
let nonoptimal_segment_uuids = {
    let segments = segments.read();
    nonoptimal_segment_ids
        .iter()
        .filter_map(|id| segments.get(*id))
        .map(|segment| segment.get().read().segment_uuid())
        .collect::<Vec<_>>()
};
```

Note: with a small chance it is possible that `segments.get()` will return `None` because the segment can removed while we're not holding the lock.

lib/segment/src/types.rs
```diff
@@ -451,6 +451,7 @@ pub struct VectorDataInfo {
 #[derive(Debug, Serialize, JsonSchema, Anonymize, Clone, PartialEq, Eq)]
 #[serde(rename_all = "snake_case")]
 pub struct SegmentInfo {
+    pub segment_uuid: String,
```
xzfc: I suggest to drop the `segment_` prefix.

```suggestion
    pub uuid: String,
```

Perhaps in other places too.

lib/shard/src/segment_holder/mod.rs
KShivendu: neat. thanks!

lib/segment/src/types.rs
KShivendu: done

lib/collection/src/collection_manager/optimizers/mod.rs
```diff
@@ -99,6 +98,8 @@ pub struct Tracker {
     pub name: String,
     /// Segment IDs being optimized
     pub segment_ids: Vec<SegmentId>,
+    /// Segment UUIDs being optimized (same as segment_ids but filesystem aligned)
+    pub segment_uuids: Vec<String>,
```
coszio, line 104: I can see how having these two versions of IDs could trip me up in the future. Maybe a more detailed comment would help.

```suggestion
   /// Internal segment IDs being optimized. 
   /// These are local and in-memory, meaning that they can refer to different segments after a service restart.
   pub segment_ids: Vec<SegmentId>,
    /// Segment UUIDs being optimized. 
    /// Refers to same segments as in `segment_ids`, but trackable across restarts, and reflect their directory name
    pub segment_uuids: Vec<String>,
```

lib/collection/src/update_workers/optimization_worker.rs
```diff
@@ -287,10 +287,19 @@ impl UpdateWorkers {
...
+                        .filter_map(|id| segments.get(*id))
+                        .map(|segment| segment.get().read().segment_uuid())
+                        .collect::<Vec<_>>()
+                };
+
                 log::debug!(
                     "Optimizer '{}' running on segments: {:?}",
                     optimizer.name(),
-                    &nonoptimal_segment_ids
+                    &nonoptimal_segment_uuids
```
coszio: This is going to make it harder to read in logs 😢

xzfc:
> Question for the group: Since we don't persist internal IDs anywhere, would it be fair to refactor to only use UUID as ids?

I'm for it.

Also, we can force UUIDs to proper `Uuid` rather than `String`. (if, for some highly unlikely reason, some segment has a non-UUID path, Qdrant could rename it on load and assign a new UUID name)

timvisee:
Yes, we can now switch to dedicated UUIDs for segments.

This was not possible before, because proxy segments would hold multiple sub segments. But that has changed recently.

@KShivendu could you rebase this one to resolve the conflict?

> Also, we can force UUIDs to proper Uuid rather than String. (if, for some highly unlikely reason, some segment has a non-UUID path, Qdrant could rename it on load and assign a new UUID name)

It is indeed possible for segments to have non-UUID paths. I think your suggestion to just rename and assign a new UUID is reasonable. The most common case in which it doesn't have a valid UUID path is on deletion where we pre/suffix the segment name. But in that case we won't try to load it anyway.

PR #7734, opened 2025-12-09
Refactor metrics.rs

JojiiOfficial:
Depends on #7714

This PR contains a refactors of `metrics.rs` which is now split into multiple smaller files. Additionally the large metric function from collection metrics is much smaller now.
No logic changes have been made.

PR #7739, opened 2025-12-10
Metrics: Total time spent on optimizations

JojiiOfficial:
Depends on #7734 

Adds the following new metric to the metrics API

```
# HELP collection_optimizations_time_seconds total (cpu) time spent on optimizations in seconds
# TYPE collection_optimizations_time_seconds counter
collection_optimizations_time_seconds{id="test"} 11
```

This accumulates the time spent on optimizations. If multiple optimizations are run at the same time, this metric accounts for both and therefore **represents the CPU time** and not the time where at least one optimization was running.
I found this approach to be more granular as performance metric, and it requires a less complex implementation.

### DIFF
--- a/lib/collection/src/shards/local_shard/updaters.rs
+++ b/lib/collection/src/shards/local_shard/updaters.rs
@@ -1,6 +1,6 @@
-use std::mem::take;
 use std::sync::Arc;
 
+use ahash::HashMap;
 use tokio::sync::mpsc;
 
 use crate::operations::types::CollectionResult;
@@ -44,7 +44,14 @@ impl LocalShard {
         update_handler.stop_flush_worker();
 
         update_handler.wait_workers_stops().await?;
-        let old_optimizer = take(&mut update_handler.optimizers);
+
+        // Collect the old optimizers' telemetry collectors back keep their values.
+        let previous_telemetry_counters = update_handler
+            .optimizers
+            .iter()
+            .map(|opt| (opt.optimizer_type(), opt.get_telemetry_counter().clone()))
+            .collect::<HashMap<_, _>>();
+
         let new_optimizers = build_optimizers(
             &self.path,
             &config.params,
@@ -52,7 +59,7 @@ impl LocalShard {
             &config.hnsw_config,
             &self.shared_storage_config.hnsw_global_config,
             &config.quantization_config,
-            Some(old_optimizer),
+            Some(previous_telemetry_counters),
         );
         update_handler.optimizers = new_optimizers;
         update_handler.flush_interval_sec = config.optimizer_config.flush_interval_sec;
--- a/lib/collection/src/telemetry.rs
+++ b/lib/collection/src/telemetry.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::time::Duration;
 
 use itertools::Itertools;
 use schemars::JsonSchema;
@@ -17,8 +18,6 @@ use crate::optimizers_builder::OptimizersConfig;
 use crate::shards::replica_set::replica_set_state::ReplicaState;
 use crate::shards::shard::ShardId;
 use crate::shards::telemetry::{LocalShardTelemetry, ReplicaSetTelemetry};
-
-const MICROSECONDS_PER_SECOND: usize = 1_000_000;
 
 #[derive(Serialize, Clone, Debug, JsonSchema, Anonymize)]
 pub struct CollectionTelemetry {
@@ -180,16 +179,17 @@ impl CollectionTelemetry {
             .count()
     }
 
-    pub fn optimization_time_spent_seconds(&self) -> usize {
+    pub fn optimization_time_spent(&self) -> Duration {
         self.local_shard_iter()
             .map(|i| {
-                i.optimizations
-                    .optimizations
-                    .total_duration_micros
-                    .unwrap_or(0) as usize
-                    / MICROSECONDS_PER_SECOND
+                Duration::from_micros(
+                    i.optimizations
+                        .optimizations
+                        .total_duration_micros
+                        .unwrap_or_default(),
+                )
             })
-            .sum::<usize>()
+            .sum()
     }
 }
 
--- a/src/common/metrics/collection_metrics.rs
+++ b/src/common/metrics/collection_metrics.rs
@@ -89,7 +89,7 @@ impl CollectionsTelemetry {
             total_dead_replicas += collection.dead_replicas();
 
             total_optimizers_time_spent.push(counter(
-                collection.optimization_time_spent_seconds() as f64,
+                collection.optimization_time_spent().as_secs_f64(),
                 &[("id", &collection.id)],
             ));
 
--- a/lib/collection/src/optimizers_builder.rs
+++ b/lib/collection/src/optimizers_builder.rs
@@ -1,8 +1,9 @@
-use std::collections::HashMap;
 use std::path::Path;
 use std::sync::Arc;
 
+use ahash::HashMap;
 use fs_err as fs;
+use parking_lot::Mutex;
 use schemars::JsonSchema;
 use segment::common::anonymize::Anonymize;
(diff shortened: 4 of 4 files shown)


xzfc:
Optimizers already store similar data, can't we fix/reuse it rather adding yet another counter on top?

---

Each optimizer has `OperationDurationsAggregator`, e.g. https://github.com/qdrant/qdrant/blob/d2834de0b51be23a7b22b023e424cbb9456d0e75/lib/collection/src/collection_manager/optimizers/vacuum_optimizer.rs#L37

`OperationDurationsAggregator` has `total_duration_micros` field:
https://github.com/qdrant/qdrant/blob/d2834de0b51be23a7b22b023e424cbb9456d0e75/lib/segment/src/common/operation_time_statistics.rs#L42-L45

One problem is that optimizers start with empty `telemetry_durations_aggregator`: 
https://github.com/qdrant/qdrant/blob/d2834de0b51be23a7b22b023e424cbb9456d0e75/lib/collection/src/collection_manager/optimizers/vacuum_optimizer.rs#L63
Optimizers recreated on collection config update, meaning that we lose this statistics on collection config update. But we can update optimizers to take existing `OperationDurationsAggregator` instead of creating a new one.

Thoughts?

src/common/metrics/collection_metrics.rs
```diff
@@ -168,6 +174,14 @@ impl CollectionsTelemetry {
             prefix,
         ));
 
+        metrics.push_metric(metric_family(
+            "collection_optimizations_time_seconds",
+            "total (cpu) time spent on optimizations in seconds",
```
xzfc: It's not a proper CPU time. E.g. building HNSW for 10 seconds in 8 threads results in 10 seconds in wall time and 80 seconds in CPU time. And sleeping for 1 minute takes 0 seconds of CPU time. (optimizations might sleep for a significant time when waiting for a permit)

Perhaps reword the description?

Or if we need the real CPU time, I think it's doable, at least on Linux.

lib/collection/src/telemetry.rs
```diff
@@ -179,6 +177,12 @@ impl CollectionTelemetry {
             .filter(|i| i.replicate_states.values().any(|state| !state.is_active()))
             .count()
     }
+
+    pub fn optimization_time_spent(&self) -> usize {
```
xzfc: Style nit: either include time units in the name, or use proper type.
Also, please use `u64` instead of `usize` here and in other places.

```rust
pub fn optimization_time_spent_seconds(&self) -> u64 {
// OR
pub fn optimization_time_spent(&self) -> Duration {
// OR
pub fn optimization_time_spent(&self) -> TimeDelta {
```

docs/redoc/master/openapi.json
```diff
@@ -12912,6 +12912,12 @@
               "$ref": "#/components/schemas/TrackerTelemetry"
             },
             "nullable": true
+          },
+          "time_spent": {
+            "type": "integer",
+            "format": "uint",
+            "minimum": 0,
+            "nullable": true
```
xzfc: Should be named `time_spent_sec` as we already have `flush_interval_sec`, `bootstrap_timeout_sec`, `duration_sec` in the API.

src/common/metrics/collection_metrics.rs
JojiiOfficial: Yeah you're right, this phrasing is not completely correct. I tried to make it clear that multiple running optimizations are accumulated and result in a faster increment that 1 per second. I don't think we need real CPU time, so I'll adjust this comment 👍🏻

docs/redoc/master/openapi.json
JojiiOfficial: Not needed anymore:  [685e9a8](https://github.com/qdrant/qdrant/pull/7739/commits/685e9a8fdebebdadb35be0476e0dfd3933f75d31)

src/common/metrics/collection_metrics.rs
JojiiOfficial: Done: [685e9a8](https://github.com/qdrant/qdrant/pull/7739/commits/685e9a8fdebebdadb35be0476e0dfd3933f75d31)

timvisee:
> This accumulates the time spent on optimizations. If multiple optimizations are run at the same time, this metric accounts for both and therefore represents the CPU time and not the time where at least one optimization was running.
I found this approach to be more granular as performance metric, and it requires a less complex implementation.

Do I get it right? It is wall clock time for each individual optimization. That means that if two are running, the counter might go up twice as fast.

lib/collection/src/collection_manager/optimizers/config_mismatch_optimizer.rs
```diff
@@ -207,8 +210,8 @@ impl ConfigMismatchOptimizer {
 }
 
 impl SegmentOptimizer for ConfigMismatchOptimizer {
-    fn name(&self) -> &str {
-        "config mismatch"
+    fn optimizer_type(&self) -> OptimizerType {
+        OptimizerType::ConfigMismatch
     }
```
timvisee, line 215: @xzfc you have any opinion on this?

Would we still have 'types' like this if we implement your future optimization (planning) changes?

lib/collection/src/shards/local_shard/updaters.rs
```diff
@@ -43,13 +44,15 @@ impl LocalShard {
...
+        let old_optimizer = take(&mut update_handler.optimizers);
         let new_optimizers = build_optimizers(
             &self.path,
             &config.params,
             &config.optimizer_config,
             &config.hnsw_config,
             &self.shared_storage_config.hnsw_global_config,
             &config.quantization_config,
+            Some(old_optimizer),
         );
```
timvisee, line 63: Can we change this to only take out and pass along the statistics (telemetry counters)?

lib/collection/src/telemetry.rs
```diff
@@ -179,6 +179,18 @@ impl CollectionTelemetry {
...
+            .map(|i| {
+                i.optimizations
+                    .optimizations
+                    .total_duration_micros
+                    .unwrap_or(0) as usize
+                    / MICROSECONDS_PER_SECOND
+            })
+            .sum::<usize>()
+    }
 }
```
timvisee: I'd prefer to use `Duration` here.

Or does the use of atomics in aggregators make this hard?

Either way. Maybe we can report in metrics with a bit more detail than seconds. It uses the f64 for metrics anyway. If we cannot use the duration type, maybe we can use microseconds.

Thoughts?

lib/collection/src/collection_manager/optimizers/config_mismatch_optimizer.rs
xzfc: I'm fine with this.

lib/collection/src/telemetry.rs
JojiiOfficial: This would require changing [this](https://github.com/qdrant/qdrant/blob/master/lib/segment/src/common/operation_time_statistics.rs#L45) to `Duration`, which is used in several other places too.
If that's what we want, I'd suggest to tackle it in a separate PR. For now I'll make it return micros 👍🏻

JojiiOfficial: Done in https://github.com/qdrant/qdrant/pull/7739/commits/04755f08585cc4979b3c60bc1f6f43e029f309b7

lib/collection/src/shards/local_shard/updaters.rs
JojiiOfficial: Done in https://github.com/qdrant/qdrant/pull/7739/commits/04755f08585cc4979b3c60bc1f6f43e029f309b7

JojiiOfficial:
My initial approach was real-time, however if we want to reuse the `OperationDurationsAggregator`, I'm afraid that it might require quite a lot changes in that type. If real-time feature is important enough to justify my additional separate counter, reverting to my initial approach might probably be the cleanest way.

PR #7740, opened 2025-12-10
Optimize gridstore flush

agourlay:
No need to upgrade the lock and flush the bitmask if there are no pending updates.

### DIFF
--- a/lib/gridstore/src/gridstore.rs
+++ b/lib/gridstore/src/gridstore.rs
@@ -614,23 +614,25 @@ impl<V> Gridstore<V> {
             }
 
             let old_pointers = tracker.write().write_pending_and_flush(pending_updates)?;
-            if !old_pointers.is_empty() {
-                // Update all free blocks in the bitmask
-                bitmask_guard.with_upgraded(|guard| {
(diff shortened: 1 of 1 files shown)


lib/gridstore/src/gridstore.rs
```diff
@@ -614,23 +614,23 @@ impl<V> Gridstore<V> {
...
-                        let end = pointer.block_offset
-                            + Self::blocks_for_value(pointer.length as usize, block_size_bytes);
-                        start as usize..end as usize
-                    });
-                    guard.mark_blocks_batch(page_id, local_ranges, false);
-                }
-            });
-            bitmask_guard.flush()?;
-
+            if !old_pointers.is_empty() {
```
coszio: Style nit: This is a good place to do an early return instead of nesting

coszio (approved):
🧠

lib/gridstore/src/gridstore.rs
agourlay: Good point!

timvisee (approved):

timvisee (approved):

PR #7741, opened 2025-12-10
Fix Gridstore flushing, correctly implement drain persist

timvisee:
Fix a critical Gridstore bug that breaks flushing with alternating puts and deletes.

Gridstore uses deferred flushing. It means that we copy the list of pending updates when the flusher is created. Once flushing is invoked and updates are persisted, we remove these from the current list of updates (which might already have new entries). This 'removing' we call 'drain persisted'. It is critical to persist every change only once.

This drain function was broken and failed on some edge cases. For the problematic edge case we found in testing, I've added a new test.

The problem only appears when deferring a flush (for some time). If flushing immediately, the pending and persisted set of updates are equal and everything works as expected.

Bear in mind that I use the concept of a 'set' and 'unset' list in this PR. I plan to open a separate PR to apply this concept everywhere as it's much more easy to understand. Since it's only partially applied in this PR it might look like boilerplate.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/gridstore/src/tracker.rs
+++ b/lib/gridstore/src/tracker.rs
@@ -709,8 +709,11 @@ mod tests {
         }
     }
 
+    /// Test pointer drain edge case that was previously broken.
+    ///
+    /// See: <https://github.com/qdrant/qdrant/pull/7741>
     #[test]
-    fn test_value_pointer_drain_set_unset_set() {
+    fn test_value_pointer_drain_bug_7741() {
         // current:
         // - latest: true
         // - history: [block_offset:1, block_offset:2]
@@ -745,6 +748,9 @@ mod tests {
             expected.set(ValuePointer::new(1, 2, 1));
             expected
         };
-        assert_eq!(updates, expected, "must have correct remaining updates");
+        assert_eq!(
+            updates, expected,
+            "must have one pending update to set block offset 2",
+        );
     }
 }


lib/gridstore/src/tracker.rs
```diff
@@ -649,4 +708,43 @@ mod tests {
             assert!(updates.drain_persisted_and_drop(&persisted));
         }
     }
+
+    #[test]
+    fn test_value_pointer_drain_set_unset_set() {
```
timvisee: New test is here. This failed before fixing `drain_persisted_and_drop`.

lib/gridstore/src/tracker.rs
```diff
@@ -124,20 +124,69 @@ impl PointerUpdates {
...
 
-        // Remove all persisted pointers from history
-        self.history
-            .retain(|pointer| !persisted_pointers.contains(pointer));
+        // In current history, range of unset entries
+        let unset_range = if self.latest_is_set {
+            0..self.history.len().saturating_sub(1)
+        } else {
+            0..self.history.len()
+        };
```
timvisee, line 164: I'll significantly improve this in a future PR where I'll use a separate set/unset list.

lib/gridstore/src/tracker.rs
```diff
@@ -124,20 +124,69 @@ impl PointerUpdates {
...
+        self.history.is_empty()
+    }
+
+    /// Remove persisted entries from current so that we don't persist them again
+    fn update_from_persisted(&mut self, persisted: &Self) {
+        debug_assert_eq!(
+            self.history.iter().copied().collect::<AHashSet<_>>().len(),
+            self.history.len(),
+            "self must not have duplicate pointers in history",
+        );
```
timvisee, line 138: We also do the same debug assertion in the caller of this function. But I vote to keep it for now, until I more properly refactor it later. It does not affect release builds.

lib/gridstore/src/tracker.rs
```diff
@@ -124,20 +124,69 @@ impl PointerUpdates {
...
+
+        let (persist_set, persist_unset) = persisted.to_set_unset();
+
+        // Update current set if necessary
+        match (self.latest(), persist_set) {
+            // Current and persisted set are equal, pop set from current
+            (Some(last), Some(set)) if last == set => {
+                self.history.pop();
+                self.latest_is_set = false;
+            }
```
coszio, line 148: Isn't this the same as checking if the entire `PointerUpdates` is the same? In such case we can return early here, since the rest of the code will remove the remaining history.

timvisee: I also thought so.

But in this function I implemented the complete thing on purpose, because I don't want to make any assumptions anymore. Don't want to spend another week on it in the future. :upside_down_face:

PR #7743, opened 2025-12-10
Simplify `is_alive_flush_lock`

coszio:
The places where we use a `is_alive_flush_lock` is to achieve two things:
- ensure completion of flush
- prevent flushing a dropped instance

These two things can be already covered by a `Weak<Mutex<()>>`, so there is no need to have an additional boolean inside.

### DIFF
--- a/lib/segment/src/common/flags/buffered_dynamic_flags.rs
+++ b/lib/segment/src/common/flags/buffered_dynamic_flags.rs
@@ -21,17 +21,17 @@ pub(crate) struct BufferedDynamicFlags {
     buffer: Arc<RwLock<AHashMap<PointOffsetType, bool>>>,
 
     /// Lock to prevent concurrent flush and drop
-    is_alive_flush_lock: Arc<Mutex<bool>>,
+    flush_lock: Arc<Mutex<()>>,
 }
 
 impl BufferedDynamicFlags {
     pub fn new(mmap_flags: DynamicMmapFlags) -> Self {
         let buffer = Arc::new(RwLock::new(AHashMap::new()));
-        let is_alive_flush_lock = Arc::new(Mutex::new(true));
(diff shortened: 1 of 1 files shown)


lib/gridstore/src/gridstore.rs
```diff
@@ -585,27 +584,20 @@ impl<V> Gridstore<V> {
     pub fn flusher(&self) -> Flusher {
         let pending_updates = self.tracker.read().pending_updates.clone();
 
-        let pages = Arc::downgrade(&self.pages);
-        let tracker = Arc::downgrade(&self.tracker);
-        let bitmask = Arc::downgrade(&self.bitmask);
+        let pages = Arc::clone(&self.pages);
+        let tracker = Arc::clone(&self.tracker);
+        let bitmask = Arc::clone(&self.bitmask);
```
coszio, line 589: I also took the liberty of simplifying the extra `downgrade` usages, since it is fine to rely on this lock's `Weak` reference. If a value is inside a reference counter, it is already on the heap, so the struct can be dropped just fine. See this playground gist:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d5533e8c054da399ebfb18754c9dba1d

coszio:
Thanks @timvisee for bringing up an important detail.

There is an edge case that this new design does not protect against. 

Consider the following sequence:

| flusher | drop |
| -- | -- |
| upgrade to arc |  |
|  | obtain lock | 
|  | release lock | 
| obtain lock (when it shouldn't have) | finalize drop |
| continue flushing | |

With this PR's changes, once the weak reference has been upgraded, there is no protection against this scenario, even if it is unlikely to happen exactly like this.

Additionally, keeping weak references to structures that keep files open is better design because it allows them to be dropped as soon as the structure is dropped

PR #7745, opened 2025-12-11
[gridstore] Fix draining bug by using opnums

coszio:
Alternative to #7741 

- Move `PointerUpdates` to its own file, along with tests.
- Introduce opnums for each pointer update, to make sure we remove the exact history that was persisted.
  - Disclaimer: this implementation resets the opnum count once all updates to a single ID have been persisted. This means that it might not behave well if we take two flusher copies of pending changes, and execute them out of order. But this should be already guaranteed by flushing sequence (callers).
- Make `drain_persisted` to take by value, so it drops directly within the function; at the expense of removing an reinserting updates.

### DIFF
--- a/lib/gridstore/src/lib.rs
+++ b/lib/gridstore/src/lib.rs
@@ -4,6 +4,7 @@ pub mod config;
 pub mod fixtures;
 mod gridstore;
 mod page;
+mod pointer_updates;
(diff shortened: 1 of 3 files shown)


coszio:
This is already refactored differently by #7749

PR #7748, opened 2025-12-11
Fix Clippy v1.92

agourlay:
[Rust 1.92.0](https://releases.rs/docs/1.92.0/)
[Clippy](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-192)

Merge when the new version lands on CI in a few hours.

### DIFF
--- a/lib/segment/src/index/field_index/histogram.rs
+++ b/lib/segment/src/index/field_index/histogram.rs
@@ -23,6 +23,7 @@ pub struct Counts {
     pub right: usize,
 }
 
+#[allow(clippy::derive_ord_xor_partial_ord)]
 #[derive(PartialEq, PartialOrd, Debug, Clone, Serialize, Deserialize)]
 #[repr(C)]
 pub struct Point<T> {
@@ -38,7 +39,6 @@ impl<T> Point<T> {
 
 impl<T: PartialEq> Eq for Point<T> {}
 
-#[allow(clippy::derive_ord_xor_partial_ord)]
 impl<T: PartialOrd + Copy> Ord for Point<T> {
     fn cmp(&self, other: &Point<T>) -> std::cmp::Ordering {
         (self.val, self.idx)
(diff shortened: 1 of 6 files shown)


lib/segment/src/spaces/metric_uint/sse2/cosine.rs
```diff
@@ -7,8 +7,6 @@ use crate::spaces::simple_sse::hsum128_ps_sse;
 #[allow(clippy::missing_safety_doc)]
 pub unsafe fn sse_cosine_similarity_bytes(v1: &[u8], v2: &[u8]) -> f32 {
     debug_assert!(v1.len() == v2.len());
-    debug_assert!(is_x86_feature_detected!("sse"));
-    debug_assert!(is_x86_feature_detected!("sse2"));
```
agourlay, line 11:
```
error: this assertion has a constant value
  --> lib/segment/src/spaces/metric_uint/sse2/manhattan.rs:11:5
   |
11 |     debug_assert!(is_x86_feature_detected!("sse2"));
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: consider moving this into a const block: `const { assert!(..) }`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/beta/index.html#assertions_on_constants
```

we do not have those debug assertions in other metrics implementations so I just removed those

lib/segment/src/index/field_index/histogram.rs
```diff
@@ -23,6 +23,7 @@ pub struct Counts {
     pub right: usize,
 }
 
+#[allow(clippy::derive_ord_xor_partial_ord)]
```
agourlay, line 26: Clippy just wanted to have it on the struct instead

lib/segment/src/spaces/metric_uint/sse2/cosine.rs
timvisee: I argue this is a bad lint, and it shouldn't consider this to be a constant.

But I have no objection to removing it.

PR #7750, opened 2025-12-11
feat/shard-ops-timeout

dancixx:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/forward_proxy_shard.rs
+++ b/lib/collection/src/shards/forward_proxy_shard.rs
@@ -504,7 +504,7 @@ impl ShardOperation for ForwardProxyShard {
         // Strip the clock tag from the operation, because clock tags are incompatible between different shards.
         if self.shard_id != self.remote_shard.id {
             operation.clock_tag = None;
-        };
+        }
 
         let remote_result = self
             .remote_shard


lib/api/src/grpc/proto/points.proto
```diff
@@ -127,6 +127,7 @@ message UpsertPoints {
   optional ShardKeySelector shard_key_selector = 5; // Option for custom sharding to specify used shard keys
   optional Filter update_filter = 6; // If specified, only points that match this filter will be updated, others will be inserted
   optional uint64 timeout = 7; // Timeout for the request in seconds
+  optional uint64 wait_timeout = 8; // Time limit for waiting for the changes to be applied
```
generall: please avoid this extra parameter

lib/api/src/grpc/proto/points.proto
```diff
@@ -908,6 +918,7 @@ enum UpdateStatus {
   Acknowledged = 1; // Update is received, but not processed yet
   Completed = 2; // Update is applied and ready for search
   ClockRejected = 3; // Internal: update is rejected due to an outdated clock
+  WaitTimeout = 4; // Timeout of awaited operations
```
generall: ~let's not expose it~

lib/collection/src/collection/point_ops.rs
```diff
@@ -29,6 +29,7 @@ impl Collection {
         &self,
         operation: CollectionUpdateOperations,
         wait: bool,
+        timeout: Option<Duration>,
```
generall: Do we always call in inside consensus? If yes, there should be no timeout possible

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -93,19 +95,37 @@ impl ShardOperation for LocalShard {
                 operation: operation.operation,
                 sender: callback_sender,
                 wait,
+                timeout,
```
generall: Worker can't timeout update operation, so it is better if it doesn't even know about it

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -93,19 +95,37 @@ impl ShardOperation for LocalShard {
...
         };
 
         if let Some(receiver) = callback_receiver {
-            let _res = receiver.await??;
-            Ok(UpdateResult {
-                operation_id: Some(operation_id),
-                status: UpdateStatus::Completed,
-                clock_tag: operation.clock_tag,
-            })
+            let fut_timeout = Box::pin(async {
```
generall: let's use `tokio::time::timeout` for this, should be easier to read

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -93,19 +95,37 @@ impl ShardOperation for LocalShard {
...
+                    let _ = res??;
+                    Ok(UpdateResult {
+                        operation_id: Some(operation_id),
+                        status: UpdateStatus::Completed,
+                        clock_tag: operation.clock_tag,
+                    })
+                }
+                Either::Right(((), _wait_future)) => Ok(UpdateResult {
+                    operation_id: Some(operation_id),
+                    status: UpdateStatus::WaitTimeout,
```
generall: let's just return `Acknowledged`

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -428,6 +460,7 @@ impl ShardReplicaSet {
                 UpdateStatus::Completed => true,
                 UpdateStatus::Acknowledged => false,
                 UpdateStatus::ClockRejected => false,
+                UpdateStatus::WaitTimeout => false,
```
generall, line 463: `WaitTimeout` status should dominate the response, but still should not cause dead replicas

lib/api/src/grpc/proto/points.proto
generall: disregard this

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -35,6 +35,7 @@ impl UpdateWorkers {
                     operation,
                     sender,
                     wait,
+                    timeout: _,
```
generall: let's remove this

xzfc:
Merging #7475 introduced merge conflicts in this branch, so I rebased it.

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -678,11 +711,61 @@ impl ShardReplicaSet {
...
+
+    /// Merges successful update results from a replica set.
+    ///
+    /// If there are WaitTimeout results, the merged result will be WaitTimeout.
+    /// If there are Completed results, the merged result will be Completed.
+    /// Otherwise, the merged result will be Acknowledged.
+    fn merge_successful_update_results<F>(
+        this_peer_id: PeerId,
+        successes: &[(PeerId, UpdateResult)],
+        is_source_of_truth: F,
```
generall: check for source of truth is kind of expensive, is requires acquiring several read locks. And it also kind of unnecessary either. 

I propose to take response from largest peer_id, so that the response is consistent regardless of LB routing

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -678,11 +711,61 @@ impl ShardReplicaSet {
...
+        let mut result = successes
+            .iter()
+            .max_by_key(|(peer_id, _)| (is_source_of_truth(*peer_id), *peer_id == this_peer_id))
+            .map(|(_, res)| *res)
+            .expect("successes is not empty");
+
+        // Overwrite with the aggregated status across all successful replies
+        result.status = status;
+
+        // Propagate the freshest clock we observed among successful replies.
```
generall: I think previous implementation didn't care about it, right? I propose to keep old behaviour just in case

src/common/inference/service.rs
```diff
@@ -201,15 +202,27 @@ impl InferenceService {
...
+        let mut request_body = InferenceRequest {
             inputs: inference_inputs,
             inference: Some(inference_type),
             token,
         };
 
         let request = self.client.post(url);
-
         let request = if let Some(timeout) = timeout {
+            // Apply timeout from params as fallback
```
generall: Is this necessary? A bit hard to read this part and I am not sure what it is intended to do exactly

generall (changes requested):
Update response should also be resolved here https://github.com/qdrant/qdrant/blob/fa94a1874715399de366fedcda641100f6517c9a/lib/collection/src/collection/point_ops.rs#L250

docs/redoc/master/openapi.json
```diff
@@ -9471,7 +9471,8 @@
         "type": "string",
         "enum": [
           "acknowledged",
-          "completed"
+          "completed",
+          "wait_timeout"
```
timvisee, line 9565: Did you check and confirm this (and the change in gRPC) is compatible with all our existing clients?

I'm afraid they might fail to successfully parse a response status like this.

lib/collection/src/operations/types.rs
```diff
@@ -382,11 +382,33 @@ pub struct RemoteShardInfo {
 pub enum UpdateStatus {
     Acknowledged,
     Completed,
+    WaitTimeout,
```
timvisee, line 400: We should update the documentation/comment of this enum to include `WaitTimeout`

src/common/update.rs
```diff
@@ -38,26 +39,31 @@ pub struct UpdateParams {
     #[serde(default)]
     pub ordering: WriteOrdering,
     #[serde(default)]
-    pub timeout: Option<Duration>,
+    pub timeout: Option<NonZeroU64>,
```
timvisee: In my opinion we should use `Duration` everywhere possible. Not plain numbers.

src/schema_generator.rs
timvisee, line 1: Please revert this permission change. It does not need to be executable.

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -43,10 +43,12 @@ impl ShardOperation for LocalShard {
         &self,
         mut operation: OperationWithClockTag,
         wait: bool,
+        timeout: Option<Duration>,
         hw_measurement_acc: HwMeasurementAcc,
     ) -> CollectionResult<UpdateResult> {
         // `LocalShard::update` only has a single cancel safe `await`, WAL operations are blocking,
         // and update is applied by a separate task, so, surprisingly, this method is cancel safe. :D
+        let start = Instant::now();
```
timvisee: Using:

```suggestion
        let deadline = Instant::now() + timeout;
```

and 

```rust
tokio::time::timeout_at(deadline, receiver)
```

may be more reasonable to get rid of the manual `time_left` calculation.

I'll let you decide.

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -100,12 +102,29 @@ impl ShardOperation for LocalShard {
         };
 
         if let Some(receiver) = callback_receiver {
```
timvisee: I'd like to replace this whole if-else-block with a match:

```rust
match (callback_receiver, timeout) {
    // Wait indefinitely
    (Some(receiver), None) => {
        let _ = receiver.await??;
        Ok(UpdateResult {
            operation_id: Some(operation_id),
            status: UpdateStatus::Completed,
            clock_tag: operation.clock_tag,
        })
    }

    // Wait with timeout
    (Some(receiver), Some(t)) => {
        let time_left = t.saturating_sub(start.elapsed());
        match tokio::time::timeout(time_left, receiver).await {
            Ok(res) => {
                res??;
                Ok(UpdateResult {
                    operation_id: Some(operation_id),
                    status: UpdateStatus::Completed,
                    clock_tag: operation.clock_tag,
                })
            }
            Err(_) => Ok(UpdateResult {
                operation_id: Some(operation_id),
                status: UpdateStatus::WaitTimeout,
                clock_tag: operation.clock_tag,
            }),
        }
    }

    // Don't wait at all
    (None, _) => {
        Ok(UpdateResult {
            operation_id: Some(operation_id),
            status: UpdateStatus::Acknowledged,
            clock_tag: operation.clock_tag,
        })
    }
}
```

lib/collection/src/collection/point_ops.rs
```diff
@@ -241,8 +247,44 @@ impl Collection {
...
+                });
+            }
+
+            let max_operation_id = results.into_iter().map(|r| r.operation_id).max().unwrap(); // We checked that results is not empty above
+
+            Ok(UpdateResult {
+                operation_id: max_operation_id,
+                status,
+                clock_tag: None, // clock_tag is not used in the user response
+            })
```
timvisee, line 287: This changes behavior. Is there any specific reason for doing that?

I'd rather return the last result instead like we did before.

lib/collection/src/collection/point_ops.rs
```diff
@@ -241,8 +247,44 @@ impl Collection {
...
+
+            let results: Vec<_> = results.into_iter().flatten().collect();
+            // Aggregate status: WaitTimeout > .. > ClockRejected
+            let status = results
+                .iter()
+                .map(|res| res.status)
+                .max_by_key(|s| s.priority())
+                .unwrap_or(UpdateStatus::Acknowledged);
+
+            if !is_user_timeout && status.is_timeout() {
```
timvisee: I don't like this.

The priorities create a hidden footgun in combination with the timeout check below it. If we change priorities and/or add a new variant this might break.

I strongly prefer this instead:

```rust
if !is_user_timeout && results.iter().any(|res| res.status.is_timeout()) {
```

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -142,11 +165,17 @@ impl ShardReplicaSet {
...
+                wait,
+                timeout,
+                update_only_existing,
+                hw_measurement_acc,
+            )
+            .await
         } else {
             // Forward the update to the designated leader
-            self.forward_update(leader_peer, operation, wait, ordering, hw_measurement_acc)
+            self.forward_update(leader_peer, operation, wait, None, ordering, hw_measurement_acc)
```
timvisee: Why do we omit the timeout here?

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -100,12 +102,29 @@ impl ShardOperation for LocalShard {
...
+            let success = UpdateResult {
                 operation_id: Some(operation_id),
                 status: UpdateStatus::Completed,
                 clock_tag: operation.clock_tag,
-            })
+            };
+
+            if let Some(t) = timeout {
+                let time_left = t.saturating_sub(start.elapsed());
+                return match tokio::time::timeout(time_left, receiver).await {
```
timvisee: I'd like to see a unit test for this behavior. We may utilize the staging 'delay' operation for that. I'm fine with gating the test with `#[cfg(feature = "staging")]`.

lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -408,6 +410,7 @@ impl ShardOperation for ForwardProxyShard {
         &self,
         operation: OperationWithClockTag,
         _wait: bool,
+        _timeout: Option<Duration>,
```
timvisee: I believe we need to use the timeout here too. The shard behaves as a regular shard, with extra functionality to also propagate the changes somewhere else.

src/common/update.rs
```diff
@@ -1011,11 +1017,7 @@ pub async fn update(
         clock_tag,
     } = internal_params;
 
-    let UpdateParams {
-        wait,
-        ordering,
-        timeout: _,
-    } = params;
+    let UpdateParams { wait, ordering, .. } = params;
```
timvisee: Let's keep this here:

```suggestion
    let UpdateParams {
        wait,
        ordering,
        timeout: _,
    } = params;
```

src/common/update.rs
dancixx: @timvisee I followed the `ReadParams`, if you don't like, I'll switch to `Duration`.

```rust
#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Deserialize, JsonSchema, Validate)]
pub struct ReadParams {
    #[serde(default, deserialize_with = "deserialize_read_consistency")]
    #[validate(nested)]
    pub consistency: Option<ReadConsistency>,
    /// If set, overrides global timeout for this request. Unit is seconds.
    pub timeout: Option<NonZeroU64>,
}
```

lib/collection/src/shards/local_shard/shard_ops.rs
dancixx: @timvisee yeah sure.

docs/redoc/master/openapi.json
generall: Clients will not receive new `wait_timeout` status unless they explicitly set timeout (which is only available in new version)

lib/collection/src/collection/point_ops.rs
generall: You mean `clock_tag` or something else?

Previous behavior was to return first responded result, which is essentially random. By introducing more consistent results we improve behavior.

src/common/update.rs
timvisee: In that case, let's not bother and keep it as is. If you already made the change, that's also good.

lib/collection/src/collection/point_ops.rs
timvisee: No, previously we returned the last result (with pop).

src/common/update.rs
```diff
@@ -39,7 +38,7 @@ pub struct UpdateParams {
     #[serde(default)]
     pub ordering: WriteOrdering,
     #[serde(default)]
-    pub timeout: Option<NonZeroU64>,
+    pub timeout: Option<Duration>,
```
timvisee, line 43: We might need something like this (untested) from [`serde_with`](https://docs.rs/serde_with/):

```suggestion
    #[serde_as(as = "DurationSeconds<String>")]
    pub timeout: Option<Duration>,
```

lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -505,7 +508,7 @@ impl ShardOperation for ForwardProxyShard {
 
         let remote_result = self
             .remote_shard
-            .update(operation, false, hw_measurement_acc)
+            .update(operation, false, None, hw_measurement_acc)
```
timvisee, line 511: Is it correct we don't use a timeout here?

timvisee: Yes, this is reasonable because `wait=false`.

timvisee:
I'd like to do a proper (manual) test before I approve this PR.

I do see the new `wait_timeout` status when running this branch, which is good.

However, on `dev` I still fail to reproduce the case where we mark replicas as dead.

generall:
> However, on dev I still fail to reproduce the case where we mark replicas as dead.

were you able to trigger timeout for update operation?

PR #7751, opened 2025-12-12
Introduce `IsAliveLock`

coszio:
The pattern of having an `Arc<Mutex<bool>>` for protecting flusher closures from being executed when the structure is dropped, is used in various places in the project.

This PR abstracts this into a new `IsAliveLock`, which has a custom `Drop` implementation (so we don't need custom ones on the callers), and abstracts away the boolean by just providing a `lock_if_alive` function.

### DIFF
--- a/lib/common/common/src/is_alive_lock.rs
+++ b/lib/common/common/src/is_alive_lock.rs
@@ -1,7 +1,6 @@
 use std::sync::{Arc, Weak};
 
-use parking_lot::{ArcMutexGuard, Mutex, MutexGuard, RawMutex};
-use self_cell::self_cell;
+use parking_lot::{ArcMutexGuard, Mutex, RawMutex};
 
 /// Structure which ensures that the lock is alive at the time of locking,
 /// and will prevent dropping while guarded.


lib/common/common/src/is_alive_lock.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// Get a handle for this lock.
+    pub fn handle(&self) -> IsAliveHandle {
+        IsAliveHandle {
+            inner: self.inner.clone(),
+        }
+    }
+
+    /// Waits for lock and poisons without dropping.
+    /// Lock will no longer be usable after this.
+    pub fn poison(&self) {
```
agourlay: I am a bit confused by the term `poison`.

For me it refers to handling panics in threads holding locks so I feel the term is ambiguous here.

coszio: renamed to `mark_dead`

lib/segment/src/segment/mod.rs
```diff
@@ -70,9 +71,7 @@ pub struct Segment {
     /// Locked structure on which we hold the lock during flush to prevent concurrent flushes
     pub persisted_version: Arc<Mutex<Option<SeqNumberType>>>,
     /// Lock to prevent concurrent flushes and used for waiting for ongoing flushes to finish.
-    /// Default value is true - means segment is alive and flushes can be performed.
-    /// If value is false - means segment was dropped and no further flush is required.
-    pub is_alive_flush_lock: Arc<Mutex<bool>>,
+    pub is_alive_flush_lock: IsAliveLock,
     /// Path of the storage root
```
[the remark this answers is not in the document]
coszio: Segment is not public

lib/common/common/src/is_alive_lock.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// Get a handle for this lock.
+    pub fn handle(&self) -> IsAliveHandle {
+        IsAliveHandle {
+            inner: self.inner.clone(),
+        }
+    }
+
+    /// Waits for lock and marks as dead without dropping.
+    /// Lock will no longer be usable after this.
+    pub fn mark_dead(&self) {
```
timvisee: I vote to rename this to `block_and_mark_dead()` to emphasize this may block for (a long?) time.

lib/common/common/src/is_alive_lock.rs
```diff
@@ -0,0 +1,141 @@
...
+/// Handle for `IsAliveLock` which can access the lock at a later time.
+///
+/// This is a separate structure so it does not change the boolean on drop.
+pub struct IsAliveHandle {
+    inner: Arc<Mutex<bool>>,
+}
+
+impl IsAliveHandle {
+    /// Get a guard of this lock if the parent hasn't been dropped
+    pub fn lock_if_alive(&self) -> Option<IsAliveGuard<'_>> {
```
timvisee:
```suggestion
    /// Get a guard of this lock if the parent hasn't been dropped
    #[must_use = "Guard must be held for lifetime of operation, abort if None is returned"]
    pub fn lock_if_alive(&self) -> Option<IsAliveGuard<'_>> {
```

lib/common/common/src/is_alive_lock.rs
```diff
@@ -0,0 +1,141 @@
...
+        self.mark_dead();
+    }
+}
+
+/// Handle for `IsAliveLock` which can access the lock at a later time.
+///
+/// This is a separate structure so it does not change the boolean on drop.
+pub struct IsAliveHandle {
+    inner: Arc<Mutex<bool>>,
+}
```
timvisee, line 56: Now that we have a dedicated structure, it is a bit more comfortable to use a weak reference here. While it doesn't have any significant effect, shall we just use it anyway?

lib/gridstore/src/gridstore.rs
```diff
@@ -598,16 +597,14 @@ impl<V> Gridstore<V> {
...
 
-        let is_alive_flush_lock = self.is_alive_flush_lock.clone();
+        let is_alive_flush_lock = self.is_alive_flush_lock.handle();
 
         Box::new(move || {
             // Keep the guard till the end of the flush to prevent concurrent flushes
-            let is_alive_flush_guard = is_alive_flush_lock.lock();
-
-            if !*is_alive_flush_guard {
+            let Some(_is_alive_flush_guard) = is_alive_flush_lock.lock_if_alive() else {
```
timvisee: I vote to explicitly `drop(is_alive_flush_guard)` at the end of the function. I'm afraid we'll miss this important bit when refactoring in the future. And incorrectly handling this lock will not show immediate problems.

Same for other places we use it.

lib/common/common/src/is_alive_lock.rs
```diff
@@ -0,0 +1,141 @@
...
+        Self::new()
+    }
+}
+
+impl Drop for IsAliveLock {
+    fn drop(&mut self) {
+        // prevent dangling handles from accessing the lock
+        self.mark_dead();
+    }
+}
```
timvisee, line 49: I'm afraid this may introduce a huge problem! A problem that breaks flushing _once again_. If so, it would have existed before and it not introduced in this PR. Though I do suggest to fix it in here.

Specifically, I'm afraid about graceful shutdown, undefined cancellation/abort order and WAL acknowledgement.

We have a flush worker in each local shard to flush all segments. It writes the latest acknowledged version into the WAL, so that we only replay newer versions on restart.

That logic you can find here: https://github.com/qdrant/qdrant/blob/6fe46d0d6097ba772f2cb654f8e02b99f6db2181/lib/collection/src/update_workers/flush_workers.rs#L65-L97

I'm afraid that on regular (graceful) shutdown we might:
1. shortcut flushing in some structures (due to is alive lock)
2. segment holder will still assume successful flush
3. update worker will still acknowledge newest version even though it isn't fully persisted

I'm not exactly sure this **can** happen. The problem is that we currently don't have strict control over the order of dropping/cancellation/aborting of flush tasks and futures. I don't suggest to make it well ordered as that'll be a huge effort.

My guess is that the only right way of solving this is to not shortcut flushers with `Ok(())`, but with `Err(Cancelled)`. That way we can explicitly handle cancellation down the line.

lib/common/common/src/is_alive_lock.rs
```diff
@@ -0,0 +1,141 @@
...
+impl Default for IsAliveLock {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Drop for IsAliveLock {
+    fn drop(&mut self) {
+        // prevent dangling handles from accessing the lock
+        self.mark_dead();
```
timvisee: Also we must not forget that implicitly blocking on a drop here may be a potential footgun deadlocking Tokio runtimes with a low runtime. Currently the only thing that protects us against it is that the lock holders (flusher tasks) don't use threads in the same pool.

We can specialize this structure to flushers, and debug assert that all handles never call in [any runtime context](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.try_current).

timvisee (changes requested):

lib/segment/src/common/mmap_bitslice_buffered_update_wrapper.rs
```diff
@@ -74,16 +74,14 @@ impl MmapBitSliceBufferedUpdateWrapper {
...
         Box::new(move || {
             // Keep the guard till the end of the flush to prevent concurrent drop/flushes
-            let is_alive_flush_guard = is_alive_flush_lock.lock();
-
-            if !*is_alive_flush_guard {
+            let Some(_is_alive_flush_guard) = is_alive_flush_lock.lock_if_alive() else {
                 // Already dropped, skip flush
                 return Ok(());
-            }
+            };
```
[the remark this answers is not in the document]
timvisee: Thank you for echoing my concerns

lib/common/common/src/is_alive_lock.rs
coszio: This PR only refactored the current logic into a self-contained struct. I believe the changes you are proposing belong to a separate PR.

Moreover, changing from Ok(()) to Err(_) is determined by how `.lock_if_alive` is used, so it depends on the caller side.

timvisee: I'd be fine with handling it in a separate PR.

lib/common/common/src/is_alive_lock.rs
coszio: In the end it required a non-trivial change by adding a self-referencing struct to be able to return the upgraded arc along with the guard.

I still pushed it because a Weak is more fitting in this case

lib/common/common/src/is_alive_lock.rs
```diff
@@ -52,21 +53,48 @@ impl Drop for IsAliveLock {
...
 }
 
 impl IsAliveHandle {
     /// Get a guard of this lock if the parent hasn't been dropped
     #[must_use = "Guard must be held for lifetime of operation, abort if None is returned"]
-    pub fn lock_if_alive(&self) -> Option<IsAliveGuard<'_>> {
-        let guard = self.inner.lock();
-        guard.then_some(IsAliveGuard(guard))
+    pub fn lock_if_alive(&self) -> Option<IsAliveGuard> {
+        if let Some(mutex) = self.inner.upgrade() {
```
timvisee: Nit, lets remove nesting:

```rust
let Some(mutex) = self.inner.upgrade() else { return None; };
```

lib/common/common/src/is_alive_lock.rs
coszio: thx for the rename

lib/common/common/src/is_alive_lock.rs
```diff
@@ -52,21 +53,48 @@ impl Drop for IsAliveLock {
...
+// Private self_cell implementation
+self_cell!(
+    struct IsAliveGuardCell {
+        owner: Arc<Mutex<bool>>,
+
+        #[covariant]
+        dependent: IsAliveGuardInner,
+    }
+);
+struct IsAliveGuardInner<'a>(#[expect(dead_code)] MutexGuard<'a, bool>);
```
timvisee: I didn't realize it'd become so ugly just because we have to hold the Arc alive.

Can't we use the ArcMutexGuard type here?

https://docs.rs/parking_lot/latest/parking_lot/struct.ArcMutexGuard.html ([via](https://docs.rs/parking_lot/latest/parking_lot/type.Mutex.html#method.lock_arc))

I believe it only needs the Arc when obtaining the lock, after which we can drop the upgraded reference.

If possible, I strongly prefer that over this self cell hack.

lib/common/common/src/is_alive_lock.rs
coszio: Even better with `?` usage 😄 
https://github.com/qdrant/qdrant/pull/7751/commits/f3b1c4d50fc3e61c28f4838523631cce4dfbd3c5

lib/common/common/src/is_alive_lock.rs
coszio: I didn't know about this! It makes it SOOO much better, thanks

PR #7752, opened 2025-12-12
OptimizationPlanner

xzfc:
This PR refactors optimizers to let them write the full optimization plan before performing any optimization. In subsequent PR this plan will be used to estimate amount of points to be indexed.

Before this PR:
- `SegmentOptimizer::check_condition()` returns a single batch of segments to optimize. (aka the worst batch)
- `UpdateWorkers::launch_optimization()` calls `check_condition` in two nested loops and schedules optimization immediately.

In this PR:
- `SegmentOptimizer::check_condition()` is replaced with `SegmentOptimizer::plan_optimizations()`. It returns all possible batches of segments to optimize, not just the first one.
- `UpdateWorkers::launch_optimization()` works in two stages: first make plan, then launch optimizations.

In pseudo-code, before this PR:
```python
OPTIMIZERS = [MergeOptimizer, IndexingOptimizer, VacuumOptimizer, ConfigMismatchOptimizer]

def UpdateWorkers::launch_optimization(all_segments):
    scheduled_segment_ids = set()
    for optimizer in OPTIMIZERS:
        while True:
            nonoptimal_segment_ids := optimizer.check_condition(
                all_segments,
                excluded_ids=scheduled_segment_ids,
            )
            if not nonoptimal_segment_ids:
                break # continue to next optimizer
            scheduled_segment_ids.update(nonoptimal_segment_ids)
            if available_permits > 0:
                launch_optimizations_in_background(segment_ids=nonoptimal_segment_ids)
            else:
                return
```

In this PR:
```python
def UpdateWorkers::launch_optimization(all_segments):
    # Plan
    planner = OptimizationPlanner(all_segments)
    for optimizer in OPTIMIZERS:
        optimizer.plan_optimizations(planner)

    # Launch
    for batch in planner.segments_to_optimize:
        if available_permits > 0:
            launch_optimizations_in_background(segment_ids=batch)
        else:
            break
```

# Commits

- First 7 commits are preparation refactoring: minor fixes, renames and code moves. No changes in logic.

- > OptimizationPlanner

  This is the main part.

- > MergeOptimizer: allow merging two segments, update doc

  This tweaks `MergeOptimizer` to be more efficient.
  The goal of `MergeOptimizer` is not changed: it should guerantee that optimizing will decrease overall segment count.
  Old approach: merge at least three segments into one.
  New approach: merge at least three segments OR merge at least two batches (merge 4 segments into 2 segments).

  This would help us predict more than one step ahead; and also should trigger parallel optimizations earlier.

### DIFF
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -101,7 +101,18 @@ pub trait SegmentOptimizer {
             let mut planner =
                 OptimizationPlanner::new(i + 1, remaining.iter().map(|(&id, &seg)| (id, seg)));
             self.plan_optimizations(&mut planner);
-            assert_eq!(planner.into_scheduled_for_test(), &result[i + 1..]);
+            let actual = planner.into_scheduled_for_test();
+            let expected = &result[i + 1..];
+            if self.name() == "merge"
+                && actual.is_empty()
+                && expected.len() == 1
+                && expected[0].len() == 2
+            {
+                // Special case for MergeOptimizer:
+                // `[A B] [C D]` is allowed, but `[C D]` is not. See its doc.
+                continue;
+            }
+            assert_eq!(actual, expected);
         }
 
         result
--- a/lib/collection/src/collection_manager/optimizers/merge_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/merge_optimizer.rs
@@ -14,11 +14,34 @@ use crate::config::CollectionParams;
 
 const BYTES_IN_KB: usize = 1024;
 
-/// Optimizer that tries to reduce number of segments until it fits configured value.
+/// Optimizer that tries to reduce number of segments until it fits configured
+/// value.
 ///
-/// It merges 3 smallest segments into a single large segment.
-/// Merging 3 segments instead of 2 guarantees that after the optimization the number of segments
-/// will be less than before.
+/// ```text
+/// Suppose we have a set of mergeable segments, sorted by size.
+/// `A` is smallest, `M` is largest.
+///
+///     A B C D E F G H I J K L M
+///
+/// MergeOptimizer greedily arranges them in batches up to the size threshold.
+///
+///     [A B C D] [E F G] [H I J] K L M
+///     └───X───┘ └──Y──┘ └──Z──┘
+///
+/// After merging these batches, our segments would look like this:
+///
+///     ∅ X Y Z K L M
+///
+/// `∅` is the newly created appendable segment that Qdrant could potentially
+/// create because MergeOptimizer merged the last appendable segment.
+///
+/// To guarantee that the number of segments will be reduced after the merge,
+/// either merge a batch of at least 3 segments, or merge at least two batches.
+///
+/// - bad:   [A B]        →  ∅ X    (segment count is the same)
+/// - good:  [A B C]      →  ∅ X    (one segment less)
+/// - good:  [A B] [C D]  →  ∅ X Y  (one segment less)
+/// ```
 pub struct MergeOptimizer {
     default_segments_number: usize,
     thresholds_config: OptimizerThresholds,
@@ -109,6 +132,7 @@ impl SegmentOptimizer for MergeOptimizer {
             .max_segment_size_kb
             .saturating_mul(BYTES_IN_KB);
 
+        let mut first_batch = None;
         let mut taken_candidates = 0;
         let mut last_candidate =
             (planner.expected_segments_number() + 2).saturating_sub(self.default_segments_number);
@@ -121,13 +145,22 @@ impl SegmentOptimizer for MergeOptimizer {
                 })
                 .collect_vec();
 
-            // We need at least 3 segments because in this case we can guarantee
-            // that total segments number will be less.
-            if batch.len() < 3 {
+            if batch.len() < 2 {
                 return;
             }
+            let is_first_batch = taken_candidates == 0;
             taken_candidates += batch.len();
             last_candidate += 1;
+            if is_first_batch && batch.len() < 3 {
+                // First batch has length 2. To guarantee that the number of
+                // segments will be reduced, we need another batch.
+                // So, hold the first batch until we find the second one.
+                first_batch = Some(batch);
+                continue;
+            }
+            if let Some(first_batch) = first_batch.take() {
+                planner.plan(first_batch);
+            }
             planner.plan(batch);
         }
     }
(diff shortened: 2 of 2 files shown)


lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
```diff
@@ -269,12 +153,72 @@ impl SegmentOptimizer for IndexingOptimizer {
...
+        // Select the largest unindexed segment
+        while let Some((selected_segment_id, selected_segment_size)) = unindexed.pop_back() {
+            if !planner.remaining().contains_key(&selected_segment_id) {
+                continue;
+            }
+
+            // If the number of segments if equal or bigger than the default_segments_number
+            // We want to make sure that we at least do not increase number of segments after optimization, thus we take more than one segment to optimize
+
+            if planner.expected_segments_number() < self.default_segments_number {
```
IvanPleshkov, line 190: Previously, the count of segments was calculated as
```
let number_of_segments = segments_read_guard.len();
```
It seems `segments_read_guard.len()` also counts proxy segments.

Hovewer, `planner.expected_segments_number()` has an implementation:
```
/// The expected resulting number of segments after the optimization plan is
/// executed, and all currently running optimizations are finished.
pub fn expected_segments_number(&self) -> usize {
    self.remaining.len() + self.scheduled.len() + self.current
}
```
Where remainings is a list of segments that are not included in the planned optimization, scheduled is the resulting amount of segments when all optimizations are done, and current is the number of currently running optimizations.

Proxy segments are filtered and Planner does not have proxy segments in the list of remaining (see `fn launch_optimization` in `UpdateWorkers`):
```
let mut planner = OptimizationPlanner::new(
    segments.running_optimizations.count(),
    segments.iter_original(), // It's a filter by non-proxy segments
);
```

I cannot understand if the logic is the same, and while it's unclear, I block the merge

IvanPleshkov (changes requested):

lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
```diff
@@ -58,181 +58,65 @@ impl IndexingOptimizer {
...
+
+        for (vector_name, vector_config) in self.collection_params.vectors.params_iter() {
+            if let Some(vector_data) = segment_config.vector_data.get(vector_name) {
+                let is_indexed = vector_data.index.is_indexed();
+                let is_on_disk = vector_data.storage_type.is_on_disk();
+                let storage_size_bytes = segment
+                    .available_vectors_size_in_bytes(vector_name)
                     .unwrap_or_default();
 
-                if read_segment.segment_type() == SegmentType::Special {
```
IvanPleshkov, line 76: Why did you remove all `SegmentType::Special` checks?

IvanPleshkov: After discussion with @xzfc I realized, that it's a check of the proxy segment. It's not required anymore because `OptimizationPlanner::remaining` already contains non-proxy segments only

xzfc: These checks filter out proxy segments.

I've moved this filtering logic up by one level:
https://github.com/qdrant/qdrant/blob/09b62250c55e857a49aa12ae5e0692a9ca667cd1/lib/collection/src/update_workers/optimization_worker.rs#L271-L274
`segments.iter_original()` returns only non-proxy segments.

Note that we have two possible ways to check for proxy segment: `SegmentType::Special` or `LockedSegment::Proxy(_)`. Both checks should yield the same result.

lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
```diff
@@ -269,12 +153,72 @@ impl SegmentOptimizer for IndexingOptimizer {
...
+                && segment_id != selected_segment_id
+                && selected_segment_size + size < max_segment_size_bytes
+            {
+                indexed.pop_front();
+                planner.plan(vec![selected_segment_id, segment_id]);
+                continue;
+            }
+
+            planner.plan(vec![selected_segment_id]);
+        }
```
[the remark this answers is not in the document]
xzfc: I doubt we will ever have 9000 petabyte collections.

lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
xzfc: The outcome of new `MergeOptimizer` and `IndexingOptimizer` is not the same as before this PR.

The new approach takes currently running optimizations into account, so it should produce better plans. E.g. `MergeOptimizer` can decide to merge more segment at once if `OptimizationPlanner` knows that there would be `N` new segments after ongoing optimizations are completed.

We discussed the new logic with @generall and he agreed to give the new approach a go.

generall:
@IvanPleshkov are you happy with changes?

lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
IvanPleshkov: Clear! Thanks

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -950,3 +958,82 @@ pub trait SegmentOptimizer {
...
+
+    /// The expected resulting number of segments after the optimization plan is
+    /// executed, and all currently running optimizations are finished.
+    pub fn expected_segments_number(&self) -> usize {
+        self.remaining.len() + self.scheduled.len() + self.current
+    }
+
+    /// Schedule this batch of segments to be optimized/merged into new segment.
+    pub fn plan(&mut self, segments: Vec<SegmentId>) {
+        if segments.is_empty() {
```
IvanPleshkov: Nit: `debug_assert!(!segments.is_empty());`

lib/collection/src/collection_manager/optimizers/config_mismatch_optimizer.rs
```diff
@@ -72,137 +71,102 @@ impl ConfigMismatchOptimizer {
...
-                                            .and_then(|vector_params| {
-                                                vector_params.hnsw_config.as_ref()
-                                            }),
-                                    );
-                                    if effective_hnsw.mismatch_requires_rebuild(&target_hnsw) {
-                                        return true;
-                                    }
-                                }
-                            }
+    fn config_mismatch(&self, read_segment: &dyn SegmentEntry) -> bool {
```
IvanPleshkov: Nit: rename pls `read_segment` to `segment`. Previously it was a name of read guard, here there is no locking in funciton signature and we can use just `segment` name

lib/common/common/src/process_counter.rs
```diff
@@ -0,0 +1,70 @@
...
+/// counter -= 1; // on process end
+/// ```
+/// but decrements on drop, making it impossible to forget decreasing the counter.
+/// ```ignore
+/// let mut counter = ProcessCounter::default();
+/// let handle = counter.inc(); // on process start
+/// drop(handle); // on process end
+/// ```
+#[derive(Debug, Default)] // No `Clone` as it will break the counting logic.
+pub struct ProcessCounter(Arc<()>);
```
IvanPleshkov, line 18: Can we move this whole file to the common crate?

lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
```diff
@@ -58,181 +58,65 @@ impl IndexingOptimizer {
...
-        segments
-            .iter()
-            // Excluded externally, might already be scheduled for optimization
-            .filter(|(idx, _)| !excluded_ids.contains(idx))
-            .filter_map(|(idx, segment)| {
-                let segment_entry = segment.get();
-                let read_segment = segment_entry.read();
-                let vector_size = read_segment
-                    .max_available_vectors_size_in_bytes()
+    fn require_optimization(&self, segment: &Segment) -> bool {
```
IvanPleshkov: Nit. Because it's a boolean check of condition, can we use `is_optimization_required` name? It's related also to `config_mismatch -> is_config_mismatched`

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -950,3 +958,82 @@ pub trait SegmentOptimizer {
...
+    ///
+    /// Each entry contains
+    /// - a batch of segments to be optimized/merged into a new segment,
+    /// - an optional optimizer so you can call [`SegmentOptimizer::optimize`]
+    ///   on it later.
+    scheduled: Vec<(Option<Arc<Optimizer>>, Vec<SegmentId>)>,
+
+    /// Amount of currently running optimizations. We'll assume that each of
+    /// them eventually produces one new segment.
+    current: usize,
```
IvanPleshkov: Nit. Because of the topic about the count of segments, I propose to rename this field to make it clearer that we use this value as an amount of segments afterthe  currently running optimizations are done.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -950,3 +958,82 @@ pub trait SegmentOptimizer {
...
+    }
+
+    /// The expected resulting number of segments after the optimization plan is
+    /// executed, and all currently running optimizations are finished.
+    pub fn expected_segments_number(&self) -> usize {
+        self.remaining.len() + self.scheduled.len() + self.current
+    }
+
+    /// Schedule this batch of segments to be optimized/merged into new segment.
+    pub fn plan(&mut self, segments: Vec<SegmentId>) {
```
IvanPleshkov, line 1029: Can we also check that `self.optimizer` is `Some`? Honestly, I don't like how the optimizer is provided here, but I cannot propose something better now; I understand the reason for such a decision.

lib/common/common/src/process_counter.rs
xzfc: It's already there.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
xzfc: 1. I think it would be inconvenient for tests. (as we have to wrap every optimizer into `Arc`).
2. In production, in one place this check is performed on a different level:
   https://github.com/qdrant/qdrant/blob/c7f5732723d370265a45db29f6eacd845ee2a533/lib/collection/src/update_workers/optimization_worker.rs#L282-L286
4. In production, in the other place it is not required:
   https://github.com/qdrant/qdrant/blob/c7f5732723d370265a45db29f6eacd845ee2a533/lib/collection/src/update_handler.rs#L299-L306

lib/collection/src/collection_manager/optimizers/indexing_optimizer.rs
xzfc: Renamed
`IndexingOptimizer::{require_optimizations => is_optimization_required}`
`SegmentOptimizer::{config_mismatch => has_config_mismatch}`

PR #7753, opened 2025-12-12
Cross reference telemetry timeout value

agourlay:
https://github.com/qdrant/qdrant/pull/7579#pullrequestreview-3553764431


xzfc:
Friday Trivia:

Google has a CI step that checks that similar constants are synchronized.
https://www.chromium.org/chromium-os/developer-library/guides/development/keep-files-in-sync/

(not proposing to use it here, but I find it cool)

PR #7755, opened 2025-12-12
SegmentSearcher times out on lock contention

agourlay:
This PR adds timeout on the acquisition of the locks for segment_holder and individual segments for search operations.

### DIFF
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -1,6 +1,6 @@
 use std::collections::BTreeSet;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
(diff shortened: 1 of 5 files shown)


lib/collection/src/collection_manager/segments_searcher.rs
```diff
@@ -665,6 +679,8 @@ fn execute_batch_search(
         segment_query_context,
     )?;
 
+    drop(read_segment);
```
agourlay, line 682: small optimization as the read segment is not necessary to process the further results

agourlay:
As discussed internally, I will follow up with a refactoring PR to use https://lskatz.github.io/fasten/parking_lot/struct.RwLock.html#method.try_read_until where possible to improve the tracking of the deadline during processing.

PR #7757, opened 2025-12-12
Allow filtering collections in `get_telemetry_data`

coszio:
Adds `only_collection` parameter to telemetry collection. 

This does not expose it yet in API, but will allow internal service telemetry to work nicely with JWT Access, by separating the need to only including some collections, while also including cluster info.

### DIFF
--- a/lib/storage/src/content_manager/toc/telemetry.rs
+++ b/lib/storage/src/content_manager/toc/telemetry.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::time::Duration;
 
+use ahash::HashSet;
 use collection::operations::types::CollectionResult;
 use collection::telemetry::{
     CollectionSnapshotTelemetry, CollectionTelemetry, CollectionsAggregatedTelemetry,
@@ -52,6 +53,7 @@ impl TableOfContent {
         &self,
         detail: TelemetryDetail,
         access: &Access,
+        only_collections: Option<HashSet<String>>,
         timeout: Duration,
         is_stopped: &AtomicBool,
     ) -> CollectionResult<TocTelemetryData> {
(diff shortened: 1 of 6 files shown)


lib/storage/src/content_manager/toc/telemetry.rs
```diff
@@ -43,6 +44,7 @@ impl TableOfContent {
         &self,
         detail: TelemetryDetail,
         access: &Access,
+        only_collections: Option<HashSet<String>>,
```
timvisee, line 56: How about we change it to `collection: Option<String>`?

Does it make sense to specify multiple for our use case?

coszio: For web-ui, yes. But to make it work nicely with jwt token, it is better to accept multiple ones. 

The changes are basically the same for both options, so why not choose multiple collections?

timvisee: Got it! Now I understand what you meant in the original comment.


Am I right that for the web UI this is internal only? It'll still do the same telemetry request. But internally we'll scope collections based on the JWT payload so that we don't get access errors on other collections in telemetry.

coszio: Idea is that web-ui can use `/cluster/telemetry?collections=my-collection`, but it should also work with some jwt token that specifies multiple collections.

In the first case, it will include cluster info. In the second case it won't, but it will still show the collection telemetry

timvisee: In this URL example, how would we specify multiple collections?

coszio: Since we can assume only a few things about collection names, we can set this query parameter up to accept just one collection name at a time.

Being able to handle multiple ones is still needed for collection-level access of JWT, though

PR #7758, opened 2025-12-12
[cluster telemetry] allow collection access with jwt

coszio:
Instead of requiring global access, this PR enables cluster telemetry to just show specific collections if jwt access is restricted by collection


coszio:
> During a live testing session we noticed the transfer info currently isn't included.

We missed setting `details_level=3` for it to be included. Once set, it does work as expected

PR #7759, opened 2025-12-13
use urlencode to prevent breaking collection name with % in it

generall:


lib/collection/src/shards/transfer/snapshot.rs
```diff
@@ -226,7 +228,7 @@ pub(super) async fn transfer_snapshot(
         snapshot_temp_paths.push(snapshot_checksum_temp_path);
 
         shard_download_url.set_path(&format!(
-            "/collections/{collection_id}/shards/{shard_id}/snapshots/{}",
+            "/collections/{encoded_collection_name}/shards/{shard_id}/snapshots/{}",
             &snapshot_description.name,
```
xzfc: Maybe do the same for `snapshot_description.name`?

PR #7764, opened 2025-12-15
Use ghcr.io/qdrant/qdrant over docker.pkg.github.com/qdrant/qdrant/qdrant

KShivendu:
Basically push releases tags to:
- `ghcr.io/qdrant/qdrant:<tag>` ([link](https://github.com/qdrant/qdrant/pkgs/container/qdrant)) instead of `docker.pkg.github.com/qdrant/qdrant/qdrant:<tag>` ([link](https://github.com/qdrant/qdrant/pkgs/container/qdrant%2Fqdrant))
- Also push `:latest` (not happening previously)

Note: It's a breaking change. It would be painful for anyone who relying on it (overhead dockerhub) for whatever reason.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -42,7 +42,7 @@ jobs:
 
         # Authenticate on registries
         echo "${{ secrets.DOCKERHUB_TOKEN }}" | docker login --username generall --password-stdin
-        echo "${{ secrets.GITHUB_TOKEN }}" | docker login https://docker.pkg.github.com -u qdrant --password-stdin
+        echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u qdrant --password-stdin
 
         # Build regular image for Docker Hub
         DOCKERHUB_TAG="qdrant/qdrant:${{ github.ref_name }}"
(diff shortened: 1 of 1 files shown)


timvisee (commented):
Looking at what we had before, don't we already push it to GitHub packages?

KShivendu:
> don't we already push it to GitHub packages?

Yes. But to `ghcr.io/qdrant/qdrant/qdrant` (notice the nesting)

I can continue doing it there but then it looks ugly in CM and it was always a mistake. That we can choose to fix now.

timvisee:
Ah I see now. No we cannot make this breaking change because we recommended people to use that repository for Docker images before.

KShivendu:
We can just push additional `ghcr.io/qdrant/qdrant/qdrant:latest` then :shrug: 

> looks ugly in CM

Just realised that we started testing with `:dev` so it barely matters now. We can continue using `qdrant/qdrant:latest` and `ghcr.io/qdrant/qdrant:dev` for tests. No in between.

KShivendu:
Or maybe just leave it. I don't want more ppl to have dependency on `ghcr.io/qdrant/qdrant/qdrant:latest` and we aren't gonna use it.

PR #7780, opened 2025-12-15
new `GridstoreError`

coszio:
`gridstore` crate does not have its own error, so it was using `String` everywhere.

This PR, in preparation for returning a "cancelled flush" error, implements a `GridstoreError` to allow it to handle errors more gracefully, without requiring to move `OperationError` to a common crate.

### DIFF
--- a/lib/gridstore/src/bitmask/gaps.rs
+++ b/lib/gridstore/src/bitmask/gaps.rs
@@ -9,7 +9,6 @@ use memory::mmap_type::MmapSlice;
 
 use super::{RegionId, StorageConfig};
 use crate::Result;
-use crate::error::GridstoreError;
 
 /// Gaps of contiguous zeros in a bitmask region.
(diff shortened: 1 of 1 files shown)


lib/gridstore/src/bitmask/gaps.rs
```diff
@@ -128,15 +129,12 @@ impl BitmaskGaps {
         })
     }
 
-    pub fn flush(&self) -> Result<(), mmap_type::Error> {
-        self.mmap_slice.flusher()()
+    pub fn flush(&self) -> Result<()> {
+        self.mmap_slice.flusher()().map_err(GridstoreError::from)
```
agourlay: why not relying on the automatic conversion with `?` ?

agourlay (approved):

timvisee:
Nice!

PR #7781, opened 2025-12-15
return an error when cancelling a flush

coszio:
Follow up to this discussion: https://github.com/qdrant/qdrant/pull/7751#discussion_r2618638794

This PR prevents a potential error, where, after cancelling some flusher execution, the rest of the operations should be cancelled as well. 

These changes prevent the segment to update the persisted version if **any** of the inner flusher operations gets cancelled. At the same time, the flush worker will not report it as an optimizer error, it just logs it at debug level.

### DIFF
--- a/lib/segment/src/common/flags/buffered_dynamic_flags.rs
+++ b/lib/segment/src/common/flags/buffered_dynamic_flags.rs
@@ -66,13 +66,9 @@ impl BufferedDynamicFlags {
         let is_alive_flush_lock = self.is_alive_flush_lock.handle();
 
         Box::new(move || {
-            let Some(is_alive_flush_guard) = is_alive_flush_lock.lock_if_alive() else {
-                return Err(OperationError::cancelled(
-                    "Aborted flushing on a dropped BufferedDynamicFlags instance",
-                ));
-            };
-
-            let Some(flags_arc) = flags_arc.upgrade() else {
+            let (Some(is_alive_flush_guard), Some(flags_arc)) =
+                (is_alive_flush_lock.lock_if_alive(), flags_arc.upgrade())
+            else {
                 return Err(OperationError::cancelled(
                     "Aborted flushing on a dropped BufferedDynamicFlags instance",
                 ));


agourlay:
FYI tried this branch on crasher and it does **not** fix the current issue.

Does not mean I am against the change though!

timvisee:
> FYI tried this branch on crasher and it does **not** fix the current issue.
> 
> Does not mean I am against the change though!

Thanks for checking! Was worth a shot.

lib/collection/src/update_workers/flush_workers.rs
```diff
@@ -66,8 +66,19 @@ impl UpdateWorkers {
         let confirmed_version = match confirmed_version {
             Ok(version) => version,
             Err(err) => {
-                log::error!("Failed to flush: {err}");
-                segments.write().report_optimizer_error(err);
+                // Since Self::flush_segments is flushing asynchronously, we can get the error
+                // from the previous flush cycle, not necessarily this one.
+                match err {
+                    OperationError::Cancelled { description: _ } => {
+                        log::debug!("Flush was cancelled: {err}");
```
timvisee, line 73: Great!

We should watch for this log line for some time after merger.

If we see it appear, we definitely have fixed a serious bug.

If we'd reach this before this patch with `Ok(())`, we'd definitely reach `wal.blocking_lock().ack(ack)` below incorrectly ack'ing WAL!

lib/segment/src/common/flags/buffered_dynamic_flags.rs
```diff
@@ -67,12 +67,15 @@ impl BufferedDynamicFlags {
...
+                ));
             };
 
             let Some(flags_arc) = flags_arc.upgrade() else {
-                log::debug!("skipping flushing on deleted storage");
-                return Ok(());
+                return Err(OperationError::cancelled(
+                    "Aborted flushing on a dropped BufferedDynamicFlags instance",
+                ));
             };
```
timvisee: Nit: we _can_ use a single if-statement here like this to not duplicate the message:

```rust
let (Some(is_alive_flush_guard), Some(flags_arc)) =
    (is_alive_flush_lock.lock_if_alive(), flags_arc.upgrade())
else {
    return Err(OperationError::cancelled(
        "Aborted flushing on a dropped BufferedDynamicFlags instance",
    ));
};
```

lib/collection/src/update_workers/flush_workers.rs
coszio: Not necessarily, if the segment was dropped, the max persisted version of that flush for that segment will not be part of the result. 

But who knows if there's some strange sequence that could make it problematic.

PR #7784, opened 2025-12-15
Expose planned optimizations in API

xzfc:
Depends on #7752.

This PR exposes numerical stats about pending optimizations through HTTP API.

```jsonc
// GET /collections/{collection_name}/optimizations
{
  "ongoing": [/*…*/],
  "completed": [/*…*/],
  // ↓ new
  "pending": {
    "optimizations": 48,
    "segments": 49,
    "points": 597284
  }
}
```

# Detailed representation (not in this PR)

Where these numbers come from? A debug representation could help answer this.

```jsonc
"pending": {
  "optimizations": 5,
  "segments": 8,
  "points": 58956,
  "debug": [
    "merge: [id=55 points=1400] [id=47 points=3576] [id=45 points=4460]",
    "merge: [id=50 points=5556] [id=54 points=6464]",
    "indexing: [id=53 points=12500]",
    "indexing: [id=52 points=12500]",
    "indexing: [id=49 points=12500]"
  ]
}
```

I'm not sure that such detailed info could be useful outside of testing/debugging `OptimizationPlanner`, so I'd keep it in [a separate commit](https://github.com/qdrant/qdrant/commit/report-pending-optimizers-api-debug).

### DIFF
--- a/lib/collection/src/collection/collection_ops.rs
+++ b/lib/collection/src/collection/collection_ops.rs
@@ -439,10 +439,12 @@ impl Collection {
             let Some(log) = replica_set.optimizers_log().await else {
                 continue;
             };
+
             let log = log.lock();
             let IndexingProgressViews { ongoing, completed } = log.progress_views();
             pending.merge(&log.pending);
             drop(log);
+
             all_ongoing.extend(ongoing);
             if let Some(all_completed) = all_completed.as_mut() {
                 all_completed.extend(completed);
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -317,7 +317,7 @@ pub struct PendingOptimizations {
     pub optimizations: usize,
     /// Number of unoptimized segments pending optimization.
     pub segments: usize,
-    /// Total number of points in unoptimized segments.
+    /// Total number of non-deleted points in unoptimized segments.
     pub points: usize,
 }
 
--- a/lib/collection/src/update_workers/optimization_worker.rs
+++ b/lib/collection/src/update_workers/optimization_worker.rs
@@ -10,6 +10,7 @@ use common::panic;
 use common::save_on_disk::SaveOnDisk;
 use parking_lot::Mutex;
 use segment::common::operation_error::{OperationError, OperationResult};
+use segment::entry::SegmentEntry;
 use segment::index::hnsw_index::num_rayon_threads;
 use segment::types::QuantizationConfig;
 use shard::locked_segment::LockedSegment;
@@ -286,7 +287,7 @@ impl UpdateWorkers {
                 pending_segments += segment_ids.len();
                 for &segment_id in segment_ids {
                     if let Some(LockedSegment::Original(segment)) = segments.get(segment_id) {
-                        pending_points += segment.read().total_point_count()
+                        pending_points += segment.read().available_point_count();
                     }
                 }
             }
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -16502,7 +16502,7 @@
             "minimum": 0
           },
           "points": {
-            "description": "Total number of points in unoptimized segments.",
+            "description": "Total number of non-deleted points in unoptimized segments.",
             "type": "integer",
             "format": "uint",
             "minimum": 0


lib/collection/src/collection/collection_ops.rs
```diff
@@ -432,13 +432,17 @@ impl Collection {
...
         let mut all_completed = completed_limit.map(|_| Vec::new());
+        let mut pending = PendingOptimizations::default();
 
         let shards_holder = self.shards_holder.read().await;
         for (_shard_id, replica_set) in shards_holder.get_shards() {
             let Some(log) = replica_set.optimizers_log().await else {
                 continue;
             };
-            let IndexingProgressViews { ongoing, completed } = log.lock().progress_views();
+            let log = log.lock();
```
IvanPleshkov, line 443: Nit: I find the scope for `log.lock()` easier to read than manual drop

lib/collection/src/update_workers/optimization_worker.rs
```diff
@@ -266,7 +267,10 @@ impl UpdateWorkers {
         let mut handles = vec![];
         let is_optimization_failed = Arc::new(AtomicBool::new(false));
 
-        let scheduled = {
+        let scheduled;
+        let mut pending_segments = 0;
+        let mut pending_points = 0;
+        {
```
IvanPleshkov, line 274: Nit. WDYT about returning `scheduled` from the scope?
```
let scheduled = {
    ...
}
```

lib/collection/src/update_workers/optimization_worker.rs
```diff
@@ -276,7 +280,22 @@ impl UpdateWorkers {
...
                 optimizer.plan_optimizations(&mut planner);
             }
-            planner.into_scheduled()
+            scheduled = planner.into_scheduled();
+
+            for (_, segment_ids) in &scheduled {
+                pending_segments += segment_ids.len();
+                for &segment_id in segment_ids {
+                    if let Some(LockedSegment::Original(segment)) = segments.get(segment_id) {
+                        pending_points += segment.read().total_point_count()
```
IvanPleshkov: Or `available_point_count`? If the pending optimization is a vacuum optimizer, the `pending_points` may be much larger than the result after optimization

lib/collection/src/operations/types.rs
```diff
@@ -304,6 +304,34 @@ pub struct OptimizationsResponse {
...
+    pub pending: PendingOptimizations,
+}
+
+/// Estimation of pending optimizations
+#[derive(Debug, Default, Clone, Serialize, JsonSchema)]
+pub struct PendingOptimizations {
+    /// Number of pending optimizations.
+    /// Each optimization will take one or more unoptimized segments and produce
+    /// one optimized segment.
+    pub optimizations: usize,
```
IvanPleshkov, line 317: Because we take value from this field from the `OptimizationPlanner`, we can use `Vec<ProgressTree>` here with a detailed plan instead of a number. WDYT?

lib/collection/src/update_workers/optimization_worker.rs
IvanPleshkov: Lock merge because of this thread. I find it important to resolve this question first.

lib/collection/src/collection_manager/optimizers/mod.rs
```diff
@@ -25,6 +26,7 @@ const KEEP_LAST_TRACKERS: usize = 16;
 #[derive(Default, Clone, Debug)]
 pub struct TrackerLog {
     descriptions: VecDeque<Tracker>,
+    pub pending: PendingOptimizations,
```
IvanPleshkov, line 29: My personal preference is to make everything public or private. But I guess we don't have such a guideline. So, I just ask to make `descriptions` public too if you find that access methods to `pending` are not required.

lib/collection/src/update_workers/optimization_worker.rs
IvanPleshkov: After personal discussion with @xzfc I realized, that it's not a trivial question.
If we keep as is, in case of vacuum optimization the amount of pending points is larger. But we can use this info to know how many points were deleted.
On another hand, with avaliable points the `pending_points` is stable. But in case of rewrite the whole dataset user will see in webUI that there are a lot of optimizations with `0` points. It's true but it's weird.

I let @xzfc to do this choice and unblock PR to merge

lib/collection/src/operations/types.rs
xzfc: See "Detailed representation" section in the PR description.

IvanPleshkov: Thanks! Misreaded it

lib/collection/src/collection_manager/optimizers/mod.rs
xzfc: I'd keep `descriptions` to be private because it holds some invariants (e.g. `KEEP_LAST_TRACKERS`).

lib/collection/src/update_workers/optimization_worker.rs
xzfc: Replaced with `available_point_count`, but I have no strong opinion on them.
Either way the goal is to show *some* numbers to the user; so they can make a rough estimation of how long it will take.

PR #7787, opened 2025-12-16
Add test for broken WAL delta after stream records abort

timvisee:
Fixed by <https://github.com/qdrant/qdrant/pull/7791>

Aborting a stream records (or other) transfer may break future WAL delta transfers.

This PR adds a test to reproduce the problematic behavior.

Specifically, any new updates coming in during a stream records (or other) transfer may bump the last seen clocks. This also happens on the node that is receiving the transfer. If we abort the stream records transfer, the last seen clocks may have jumped over a huge gap. A follow up WAL delta transfer will only transfer changes since the last seen clocks, missing a huge set of changes in that _jump_.

In practice, this sequence is problematic:
- start stream records transfer to recover replica
- send new update through all peers
- abort stream records transfer
- start WAL delta transfer
- ⚠️ WAL delta resolves very short or empty diff

Imagine the initial stream records transfer only covered 10% of the points before it got aborted. The WAL delta transfer after it will then miss the remaining 90% of point changes, corrupting the target replica.

Test:

```bash
cargo build --features staging
pytest tests/consensus_tests/test_shard_wal_delta_transfer.py -k test_abort_stream_records_breaks_wal_delta
```

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your core changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #7791, opened 2025-12-16
Fix broken WAL delta after stream records abort

timvisee:
Fixes <https://github.com/qdrant/qdrant/pull/7787>

I suggest to review and merge this into <https://github.com/qdrant/qdrant/pull/7787>, so that we can merge the fix and test as a whole into `dev`. See the mentioned PR for more details on the actual bug.

This PR fixes the problem by taking a 'snapshot' of last seen clocks when the replica goes in any non-active state. When doing WAL delta recovery on the replica, we derive the recovery point from the snapshot rather than the actual latest clocks.

When the replica becomes active again we're sure it's in good state. Then the clocks snapshot is cleared.

Relevant test:

```bash
pytest tests/consensus_tests/test_shard_wal_delta_transfer.py -k test_abort_stream_records_breaks_wal_delta
```

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -808,9 +808,8 @@ impl ShardReplicaSet {
             state.set_peers(replicas.clone());
         })?;
 
-        if replicas.contains_key(&self.this_peer_id()) {
-            self.on_local_state_updated(replicas.get(&self.this_peer_id()).copied().unwrap())
-                .await?;
+        if let Some(&state) = replicas.get(&self.this_peer_id()) {
+            self.on_local_state_updated(state).await?;
         }
 
         self.locally_disabled_peers.write().clear();


lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -15,7 +15,9 @@ use crate::shards::shard::PeerId;
 #[serde(from = "ClockMapHelper", into = "ClockMapHelper")]
 pub struct ClockMap {
     clocks: HashMap<Key, Clock>,
-    /// Whether this clock map has changed since the last time it was persisted.
+    /// Optional snapshot with earlier version of clocks
+    snapshot: Option<HashMap<Key, Clock>>,
+    /// Whether this clock map has changed since the last time it was persisted
     changed: bool,
 }
```
timvisee, line 22: Here I persist the optional snapshot together with regular clocks.

It took me a few iterations to finally land here. In my opinion this is the best approach because:
- we prevent adding new files
- we prevent conditional flushers
- we now flush actual and snapshot clocks atomically

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -126,15 +128,40 @@ impl ClockMap {
...
+    ///
     /// The recovery point contains every clock that is in this clock map. So, it represents all
     /// the clock ticks we have.
-    pub fn to_recovery_point(&self) -> RecoveryPoint {
+    pub fn to_recovery_point(&self, use_snapshot: bool) -> RecoveryPoint {
+        let clocks = if use_snapshot {
+            self.snapshot.as_ref().unwrap_or(&self.clocks)
+        } else {
+            &self.clocks
+        };
```
ffuugoo: I'd remove `use_snapshot` parameter and just always do `self.snapshot.as_ref().unwrap_or(&self.clocks)`.

lib/collection/src/shards/local_shard/clock_map.rs
```diff
@@ -126,15 +128,40 @@ impl ClockMap {
...
+            ClockMapSnapshot::TakeIfMissing => {
+                self.snapshot.replace(self.clocks.clone());
+                self.changed = true;
+            }
+            ClockMapSnapshot::Clear => {
+                self.snapshot.take();
+                self.changed = true;
+            }
+        }
+    }
```
ffuugoo: Tbh, I'd just make two separate methods: `take_snapshot` and `clear_snapshot`.

Makes two methods to propagate instead of one, but also removes `ClockMapSnapshot` fluff, so I think it comes up similar at the end, but more direct.

timvisee: > Makes two methods to propagate instead of one

:cry: 

Everything there is already so bloated

lib/collection/src/shards/local_shard/clock_map.rs
timvisee: We also call this on oldest clocks, and use it to determine if remote clocks are within our acceptable range. In that case we don't want to use the snapshot clocks.

For oldest clocks we currently never create a snapshot and so we should never use it. However, I'd rather be explicit here in case we do in the future.

Wdyt?

timvisee: Implemented in <https://github.com/qdrant/qdrant/pull/7791/commits/854fe281be94ac54d773cd4d5ab462a80372801d>

lib/collection/src/shards/local_shard/clock_map.rs
timvisee: Implemented in <https://github.com/qdrant/qdrant/pull/7791/commits/0c3510cddc0b7cdfa1f598082456530bdabbfaf1>

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -775,6 +808,11 @@ impl ShardReplicaSet {
             state.set_peers(replicas.clone());
         })?;
 
+        if replicas.contains_key(&self.this_peer_id()) {
```
agourlay: can we just do a replicas.get once instead of checking and then getting?

that would remove that ugly unwrap

timvisee: Good idea

timvisee: Updated in <https://github.com/qdrant/qdrant/pull/7791/commits/dddd87414403b61497d93bf6100e828d65440cdb>

PR #7792, opened 2025-12-17
Fix unrelated transfer cancellation

generall:
Turns out to be trivial fix. The only non-trivial part is that we actually need to check version.

I am also removing version check for 1.9, as direct upgrade from <1.9 -> 1.16.3 should not be possible for many different reasons.

### DIFF
--- a/lib/storage/src/content_manager/toc/collection_container.rs
+++ b/lib/storage/src/content_manager/toc/collection_container.rs
@@ -311,9 +311,8 @@ impl TableOfContent {
                 );
                 if let Err(send_error) = proposal_sender.send(operation) {
                     log::error!(
-                        "Can't send proposal to abort transfer of shard {} of collection {}. Error: {send_error}",
+                        "Can't send proposal to abort transfer of shard {} of collection {collection_name}. Error: {send_error}",
                         transfer.shard_id,
-                        collection_name,
                     );
                 }
             }
@@ -329,9 +328,8 @@ impl TableOfContent {
                     ConsensusOperations::finish_transfer(collection_name.clone(), transfer.clone());
                 if let Err(send_error) = proposal_sender.send(operation) {
                     log::error!(
-                        "Can't send proposal to complete transfer of shard {} of collection {}. Error: {send_error}",
+                        "Can't send proposal to complete transfer of shard {} of collection {collection_name}. Error: {send_error}",
                         transfer.shard_id,
-                        collection_name,
                     );
                 }
             }


lib/collection/src/collection/collection_ops.rs
```diff
@@ -247,19 +248,27 @@ impl Collection {
...
-                    self.abort_shard_transfer_and_resharding(transfer.key(), Some(&shard_holder))
-                        .await?;
-                }
+                .all_peers_at_version(&ABORT_TRANSFERS_ON_SHARD_DROP_FIX_FROM_VERSION);
+
+            // Collect shard transfers related to removed shard...
+            let transfers = if all_nodes_fixed_cancellation {
+                shard_holder.get_transfers(|transfer| {
+                    (transfer.from == peer_id || transfer.to == peer_id)
+                        && transfer.shard_id == shard_id
```
timvisee: We must also include `to_shard_id` here.

I'll patch this one up.

timvisee (changes requested):
Good call to keep this version aware. I agree that removing the 1.9.0 logic is fine.

We need to adjust the transfer selection logic a little bit more, see below:

timvisee (approved):
Asking a file review pass by @generall before merger

lib/collection/src/shards/transfer/mod.rs
```diff
@@ -66,6 +66,24 @@ impl ShardTransfer {
...
+    #[inline]
+    pub fn is_source(&self, peer_id: PeerId, shard_id: ShardId) -> bool {
+        self.from == peer_id && self.shard_id == shard_id
+    }
+
+    /// Checks whether this peer and shard ID pair is the target of this transfer
+    #[inline]
+    pub fn is_target(&self, peer_id: PeerId, shard_id: ShardId) -> bool {
+        self.to == peer_id && self.to_shard_id.unwrap_or(self.shard_id) == shard_id
+    }
```
ffuugoo, line 85: very nice 👨‍🍳🤌

lib/storage/src/content_manager/toc/collection_container.rs
```diff
@@ -311,10 +311,9 @@ impl TableOfContent {
                 );
                 if let Err(send_error) = proposal_sender.send(operation) {
                     log::error!(
-                        "Can't send proposal to abort transfer of shard {} of collection {}. Error: {}",
+                        "Can't send proposal to abort transfer of shard {} of collection {}. Error: {send_error}",
                         transfer.shard_id,
                         collection_name,
```
ffuugoo: Inline `collection_name` as well?

lib/storage/src/content_manager/toc/collection_container.rs
```diff
@@ -330,10 +329,9 @@ impl TableOfContent {
                     ConsensusOperations::finish_transfer(collection_name.clone(), transfer.clone());
                 if let Err(send_error) = proposal_sender.send(operation) {
                     log::error!(
-                        "Can't send proposal to complete transfer of shard {} of collection {}. Error: {}",
+                        "Can't send proposal to complete transfer of shard {} of collection {}. Error: {send_error}",
                         transfer.shard_id,
                         collection_name,
```
ffuugoo: Ditto.

PR #7795, opened 2025-12-17
Gridstore noisy log into assert

agourlay:
This log is an artifact from a previous debugging session which is in practice very noisy for crasher's runs.

```
2025-12-17T11:33:21.084481Z TRACE gridstore::tracker: removed pending update offset:451 prev:PointerUpdates { current: None, to_free: [] }
2025-12-17T11:33:21.084489Z TRACE gridstore::tracker: removed pending update offset:443 prev:PointerUpdates { current: None, to_free: [] }
2025-12-17T11:33:21.084498Z TRACE gridstore::tracker: removed pending update offset:440 prev:PointerUpdates { current: None, to_free: [] }
2025-12-17T11:33:21.084507Z TRACE gridstore::tracker: removed pending update offset:444 prev:PointerUpdates { current: None, to_free: [] }
2025-12-17T11:33:21.084515Z TRACE gridstore::tracker: removed pending update offset:446 prev:PointerUpdates { current: None, to_free: [] }
2025-12-17T11:33:21.084524Z TRACE gridstore::tracker: removed pending update offset:448 prev:PointerUpdates { current: None, to_free: [] }
``` 

The pointer updates are expected to be empty at this point, so let's remove the log and turn it into an assertion.

### DIFF
--- a/lib/gridstore/src/tracker.rs
+++ b/lib/gridstore/src/tracker.rs
@@ -281,7 +281,10 @@ impl Tracker {
                 if is_empty {
                     let prev = self.pending_updates.remove(&point_offset);
                     if let Some(prev) = prev {
-                        log::trace!("removed pending update offset:{point_offset} prev:{prev:?}");
+                        debug_assert!(
+                            prev.is_empty(),
+                            "remove pending element should be empty but got {prev:?}"
+                        );
                     }
                 }
             }


lib/gridstore/src/tracker.rs
```diff
@@ -281,7 +281,10 @@ impl Tracker {
                 if is_empty {
                     let prev = self.pending_updates.remove(&point_offset);
                     if let Some(prev) = prev {
-                        log::trace!("removed pending update offset:{point_offset} prev:{prev:?}");
+                        debug_assert!(
+                            prev.is_empty(),
+                            "remove pending element should be empty but got {prev:?}"
+                        );
                     }
```
timvisee, line 288: This may be a bit cleaner, removing the if-statement.

Whichever you prefer:

```rust
                    debug_assert!(
                        prev.as_ref().is_none_or(|p| p.is_empty()),
                        "remove pending element should be empty but got {prev:?}",
                    );
```rust

agourlay: TBH I do not like those combinators very much :see_no_evil:

PR #7798, opened 2025-12-17
More graceful handle of flush cancellation

coszio:
On #7781 we switched to a cancellation error when a flush would try to act on a dropped storage.

However, we only gracefully handled this cancellation within the flush worker, not everywhere else that the flush was being used, like when snapshotting.

While I don't have yet a good explanation why snapshotting was attempting to flush a dropped segment, we can do the graceful handling within the segment flusher, so it affects all callers equally.

The "graceful" part of this handling to avoid updating the segment persisted version if there was some incomplete flush.

### DIFF
--- a/lib/segment/src/segment/entry.rs
+++ b/lib/segment/src/segment/entry.rs
@@ -705,27 +705,33 @@ impl SegmentEntry for Segment {
 
             let flush_components = || {
                 // Flush mapping first to prevent having orphan internal ids.
-                id_tracker_mapping_flusher().map_err(|err| {
-                    OperationError::service_error(format!(
+                id_tracker_mapping_flusher().map_err(|err| match err {
+                    OperationError::Cancelled { .. } => err,
+                    _ => OperationError::service_error(format!(
                         "Failed to flush id_tracker mapping: {err}"
-                    ))
+                    )),
                 })?;
                 for vector_storage_flusher in vector_storage_flushers {
-                    vector_storage_flusher().map_err(|err| {
-                        OperationError::service_error(format!(
+                    vector_storage_flusher().map_err(|err| match err {
+                        OperationError::Cancelled { .. } => err,
+                        _ => OperationError::service_error(format!(
                             "Failed to flush vector_storage: {err}"
-                        ))
+                        )),
                     })?;
                 }
                 for quantization_flusher in quantization_flushers {
-                    quantization_flusher().map_err(|err| {
-                        OperationError::service_error(format!(
+                    quantization_flusher().map_err(|err| match err {
+                        OperationError::Cancelled { .. } => err,
+                        _ => OperationError::service_error(format!(
                             "Failed to flush quantized vectors: {err}"
-                        ))
+                        )),
                     })?;
                 }
-                payload_index_flusher().map_err(|err| {
-                    OperationError::service_error(format!("Failed to flush payload_index: {err}"))
+                payload_index_flusher().map_err(|err| match err {
+                    OperationError::Cancelled { .. } => err,
+                    _ => OperationError::service_error(format!(
+                        "Failed to flush payload_index: {err}"
+                    )),
                 })?;
                 // Id Tracker contains versions of points. We need to flush it after vector_storage and payload_index flush.
                 // This is because vector_storage and payload_index flush are not atomic.
@@ -733,10 +739,11 @@ impl SegmentEntry for Segment {
                 // If Id Tracker flush fails, we are also able to recover data from WAL
                 //  by simply overriding data in vector and payload storages.
                 // Once versions are saved - points are considered persisted.
-                id_tracker_versions_flusher().map_err(|err| {
-                    OperationError::service_error(format!(
+                id_tracker_versions_flusher().map_err(|err| match err {
+                    OperationError::Cancelled { .. } => err,
+                    _ => OperationError::service_error(format!(
                         "Failed to flush id_tracker versions: {err}"
-                    ))
+                    )),
                 })?;
 
                 Ok(())


lib/collection/src/update_workers/flush_workers.rs
```diff
@@ -68,17 +68,8 @@ impl UpdateWorkers {
...
-                        // Don't report as optimizer error here, since cancellation can happen due to
-                        // trying to flush an already dropped segment
-                    }
-                    _ => {
-                        log::error!("Failed to flush: {err}");
-                        segments.write().report_optimizer_error(err);
-                    }
-                }
+                log::error!("Failed to flush: {err}");
+                segments.write().report_optimizer_error(err);
```
coszio, line 72: This rolls back previous changes, as the cancellation is handled at segment level

lib/segment/src/segment/entry.rs
```diff
@@ -706,39 +706,66 @@ impl SegmentEntry for Segment {
         let flush_op = move || {
             let Some(is_alive_flush_guard) = is_alive_flush_lock.lock_if_alive() else {
                 // Segment is removed, skip flush
-                return Err(OperationError::cancelled(
-                    "Aborted flushing on a dropped segment",
-                ));
+                log::debug!("Segment was dropped, skip flush");
+                return Ok(());
```
coszio, line 703: Do not error here anymore, just not update persisted version

timvisee:
I'm afraid this still suffers from the same problem. Not at segment level, but at level of storage components inside a segment.

Imagine we only drop one of the payload indices in a segment, possibly because we just dropped the payload index config.

Then we still:
- return an error from segment flusher
- shortcut and don't flush the other pending components

I think using the `Ok(Flushed)` and `Ok(Cancelled)` style everywhere will help with that. We would be able to do the same with `Err(Cancelled)` if we explicitly handle it everywhere, but those cases may be more hidden.

timvisee:
Good (internal) note form @coszio:

This shouldn't cause inconsistencies. If we do half a flush, a next iteration of flushing will still write all changes.

coszio:
Update: PR #7781 was reverted on dev. So I have now rebased and reapplied it in this PR, so that we include cancellation error, but still only propagate it up to segment flusher.

lib/segment/src/segment/entry.rs
```diff
@@ -699,37 +699,66 @@ impl SegmentEntry for Segment {
...
+                        // Flush was cancelled, bypass
+                        OperationError::Cancelled { description } => {
+                            log::debug!("Segment flush cancelled: {description}");
+                            Ok(())
+                        }
+                        // Propagate other errors
+                        _ => Err(err),
+                    };
+                }
+            };
```
timvisee: Nit: I like this flattened match a bit better.

I'll let you decide.

```suggestion
            match flush_components() {
                // Only continue if all components flushed Ok
                Ok(()) => {}

                // Return early to avoid updating persisted version
                // Flush was cancelled, bypass
                Err(OperationError::Cancelled { description }) => {
                    log::debug!("Segment flush cancelled: {description}");
                    return Ok(());
                }

                // Propagate other errors
                Err(err) => return Err(err),
            }
```

lib/segment/src/segment/entry.rs
```diff
@@ -699,37 +699,66 @@ impl SegmentEntry for Segment {
...
-            //  by simply overriding data in vector and payload storages.
-            // Once versions are saved - points are considered persisted.
-            id_tracker_versions_flusher().map_err(|err| {
-                OperationError::service_error(format!("Failed to flush id_tracker versions: {err}"))
-            })?;
+
+                Ok(())
+            };
+
+            match flush_components() {
```
timvisee, line 752: We now clearly catch the cancellation error here and prevent bubbling it up to callers. I like it 👍

agourlay:
> While I don't have yet a good explanation why snapshotting was attempting to flush a dropped segment, 

Crasher had a missing synchronization bug where the snapshotting process was running concurrently with the collection deletion.

https://github.com/qdrant/crasher/commit/99272cafe4567fc9ce3736f774210be1d63e5170

timvisee:
> > While I don't have yet a good explanation why snapshotting was attempting to flush a dropped segment,
> 
> Crasher had a missing synchronization bug where the snapshotting process was running concurrently with the collection deletion.
> 
> [qdrant/crasher@99272ca](https://github.com/qdrant/crasher/commit/99272cafe4567fc9ce3736f774210be1d63e5170)

Also, in the previous cancellation implementation the snapshot logic would also catch and return flush failures for an earlier async flush task. This is not the case anymore, and we should therefore not see the same problem again.

lib/segment/src/segment/entry.rs
```diff
@@ -699,37 +699,63 @@ impl SegmentEntry for Segment {
...
-            })?;
-            for vector_storage_flusher in vector_storage_flushers {
-                vector_storage_flusher().map_err(|err| {
-                    OperationError::service_error(format!("Failed to flush vector_storage: {err}"))
+            let flush_components = || {
+                // Flush mapping first to prevent having orphan internal ids.
+                id_tracker_mapping_flusher().map_err(|err| {
+                    OperationError::service_error(format!(
+                        "Failed to flush id_tracker mapping: {err}"
+                    ))
```
timvisee: We wrap potential cancellation errors here as service error. This breaks cancellation handling further below!

timvisee (changes requested):
Did a quick test, and noticed the cancellation behavior is still broken.

lib/segment/src/segment/entry.rs
coszio: I solved this by not wrapping cancellation errors in https://github.com/qdrant/qdrant/pull/7798/commits/1f41e33b6efd0cf27e858a231bae2a6f5a8893f1. It does not look _too_ clean, but should be enough.

PR #7801, opened 2025-12-18
Clone pending updates in buffered storages

coszio:
Audits and implements better delayed flushing. 

This fixes the case where a flusher could lose its pending updates if the flusher closure fails or is never executed. 

Instead of taking the current pending updates, and leaving them initialized, we should clone them, and reconcile them after persisting them (often times asyncronously).

This PR applies this fix to `MutableIdTracker`, `MmapSliceBufferedUpdateWrapper`, and `MmapBitsliceBufferedUpdateWrapper`, which were the only storages missing this behavior in my audit.

### DIFF
--- a/lib/segment/src/common/mmap_slice_buffered_update_wrapper.rs
+++ b/lib/segment/src/common/mmap_slice_buffered_update_wrapper.rs
@@ -4,7 +4,7 @@ use ahash::AHashMap;
 use common::is_alive_lock::IsAliveLock;
 use common::types::PointOffsetType;
 use memory::mmap_type::MmapSlice;
-use parking_lot::{Mutex, MutexGuard, RwLock};
+use parking_lot::{Mutex, RwLock};
 
 use crate::common::Flusher;
 
@@ -86,7 +86,7 @@ where
             }
             mmap_slice_write.flusher()()?;
 
-            Self::reconcile_persisted_changes(pending_updates_arc.lock(), updates);
+            Self::reconcile_persisted_changes(&pending_updates_arc, updates);
 
             drop(is_alive_guard);
 
@@ -96,10 +96,10 @@ where
 
     /// Removes the persisted updates from the pending ones.
     fn reconcile_persisted_changes(
-        mut pending: MutexGuard<'_, AHashMap<PointOffsetType, T>>,
+        pending: &Mutex<AHashMap<PointOffsetType, T>>,
         persisted: AHashMap<PointOffsetType, T>,
     ) {
-        pending.retain(|point_offset, pending_value| {
+        pending.lock().retain(|point_offset, pending_value| {
             persisted
                 .get(point_offset)
                 .is_none_or(|persisted_value| pending_value != persisted_value)
--- a/lib/segment/src/id_tracker/mutable_id_tracker.rs
+++ b/lib/segment/src/id_tracker/mutable_id_tracker.rs
@@ -10,7 +10,7 @@ use common::types::PointOffsetType;
 use fs_err::File;
 use itertools::Itertools;
 use memory::fadvise::OneshotFile;
-use parking_lot::{Mutex, MutexGuard};
+use parking_lot::Mutex;
 use uuid::Uuid;
 
 use super::point_mappings::FileEndianess;
@@ -314,7 +314,7 @@ impl IdTracker for MutableIdTracker {
 
             store_mapping_changes(&mappings_path, &changes)?;
 
-            reconcile_persisted_mapping_changes(pending_mappings_arc.lock(), &changes);
+            reconcile_persisted_mapping_changes(&pending_mappings_arc, &changes);
 
             drop(is_alive_guard);
 
@@ -349,7 +349,7 @@ impl IdTracker for MutableIdTracker {
 
             store_version_changes(&versions_path, &changes)?;
 
-            reconcile_persisted_version_changes(pending_versions_arc.lock(), changes);
+            reconcile_persisted_version_changes(&pending_versions_arc, changes);
 
             drop(is_alive_guard);
 
@@ -821,10 +821,10 @@ where
 }
 
 fn reconcile_persisted_version_changes(
-    mut pending: MutexGuard<'_, BTreeMap<PointOffsetType, SeqNumberType>>,
+    pending: &Mutex<BTreeMap<PointOffsetType, SeqNumberType>>,
     changes: BTreeMap<PointOffsetType, SeqNumberType>,
 ) {
-    pending.retain(|point_offset, pending_version| {
+    pending.lock().retain(|point_offset, pending_version| {
         changes
             .get(point_offset)
             .is_none_or(|persisted_version| pending_version != persisted_version)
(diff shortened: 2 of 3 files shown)


lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -289,20 +293,34 @@ impl IdTracker for MutableIdTracker {
...
-        let pending_mappings = {
-            let mut pending_mappings = self.pending_mappings.lock();
-            let count = pending_mappings.len();
-            mem::replace(&mut *pending_mappings, Vec::with_capacity(count))
-        };
+        let changes = self.pending_mappings.lock().clone();
+
+        if changes.is_empty() {
+            return Box::new(|| Ok(()));
+        }
```
timvisee: Bit of a nit, but I'd like to check empty before clone (even though clone of empty should be cheap).

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -289,20 +293,34 @@ impl IdTracker for MutableIdTracker {
...
+                return Err(OperationError::cancelled(
+                    "Aborted flushing on a dropped MutableIdTracker instance",
+                ));
+            };
+
+            let pending_mappings_guard = pending_mappings_arc.lock();
+
+            store_mapping_changes(&mappings_path, &changes)?;
+
+            reconcile_persisted_mapping_changes(pending_mappings_guard, &changes);
```
timvisee: We should lock after storing.

Storing might take some time (on slow hardware). And we want to hold this lock for the shortest amount possible, since it affects all reads.

Same applies to version flusher below.

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -782,6 +823,42 @@ where
...
+        let Some(pending_version) = pending.remove(&point_offset) else {
+            continue;
+        };
+
+        // Reinsert if it has changed.
+        if pending_version != persisted_version {
+            pending.insert(point_offset, pending_version);
+        }
+    }
+}
```
timvisee, line 832: How about we use [`pending.retain(...)`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.retain) here?

We retain all in 'pending' that are not in 'changes'.

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -782,6 +823,42 @@ where
...
+            // This should not happen, since flushers are supposed to be requested->executed
+            // one at a time, but if it does, it should be fine.
+            //
+            // The only consequence is to persist some operations again.
+            break;
+        }
+        delete_up_to += 1;
+    }
+
+    *pending = pending.split_off(delete_up_to);
```
timvisee, line 851: We can probably use `retain` here too. As I remember it is very efficient on diffs like this. And it prevents us having to do this index magic.

Also, this indexed approach will break if we have concurrent flushers. We shouldn't have that, but you never know what happens in the future.

lib/segment/src/common/mmap_slice_buffered_update_wrapper.rs
```diff
@@ -31,33 +34,76 @@ where
...
+                is_alive_handle.lock_if_alive(),
+                pending_updates_weak.upgrade(),
+                slice.upgrade(),
+            ) else {
+                return Err(OperationError::cancelled(
+                    "MmapSliceBufferedUpdateWrapper was dropped before finishing flush",
+                ));
+            };
+
+            let pending_updates_guard = pending_updates_arc.lock();
```
timvisee: Also recommend to move this lock down to the reconciliation step.

lib/segment/src/common/mmap_slice_buffered_update_wrapper.rs
```diff
@@ -31,33 +34,76 @@ where
...
+            // remove persisted point offset
+            let Some(pending_value) = pending.remove(&point_offset) else {
+                continue;
+            };
+
+            // reinsert if it is not the persisted change
+            if pending_value != persisted_value {
+                pending.insert(point_offset, pending_value);
+            }
+        }
```
timvisee: Another retain candidate

lib/segment/src/common/mmap_slice_buffered_update_wrapper.rs
coszio: Initially I thought this was needed, but you are right.
Writing is already happening asyncronously, no need to lock while writing

lib/segment/src/id_tracker/mutable_id_tracker.rs
coszio: I am not sure about this one, since they are vecs representing a log.

If we have a sequence like:

```
changes: [1, 2, 3]
pending_updates: [1, 2, 3, 4, 5, 2]
```

and we do something like `.retain(|pending| !changes.contain(pending))`, we will end up with
```
pending_updates: [4, 5]
```

which will lose the latest mapping.

coszio:
I'm afraid this still gets missing points when running crasher locally. Let's see the effect on chaos... Merging

lib/segment/src/id_tracker/mutable_id_tracker.rs
timvisee: You're right. It is a log. We should keep the same ordering and retain breaks that.

The good thing is that these are idempotent. It means that if a flush didn't clear anything from the pending updates list, it's fine if we apply it again in the next flush iteration.

I do have something in mind that makes it more resilient against concurrent flushers though. I might implement that in the future.

My idea is that the beginning of the lists don't have to match.

- pending_updates: `[1, 2, 3]`
- flusher 1 created: `[1, 2, 3]`
- update comes in
- pending_updates: `[1, 2, 3, 4]`
- flusher 2 created: `[1, 2, 3, 4]`
- flusher 1 flushes: `[1, 2, 3]`
  pending_updates: `[4]`
- flusher 2 flushes: `[1, 2, 3, 4]`
  pending_updates: `[]` _(first three were missing)_

Or a slightly more complicated example:

- pending_updates: `[1, 2, 3]`
- flusher 1 created: `[1, 2, 3]`
- update comes in
- pending_updates: `[1, 2, 3, 4]`
- flusher 2 created: `[1, 2, 3, 4]`
- update comes in:
  pending_updates: `[1, 2, 3, 4, 5]`
- flusher 1 flushes: `[1, 2, 3]`
  pending_updates: `[4, 5]`
- flusher 2 flushes: `[1, 2, 3, 4]`
  pending_updates: `[5]` _(first three were missing)_

And maybe we have to enforce flushers are also applied in the same order because:

- pending_updates: `[1, 2, 3]`
- flusher 1 created: `[1, 2, 3]`
- update comes in
- pending_updates: `[1, 2, 3, 4]`
- flusher 2 created: `[1, 2, 3, 4]`
- flusher 2 flushes: `[1, 2, 3, 4]`
  pending_updates: `[]`
- flusher 1 flushes: `[1, 2, 3]` (⚠️ we now don't write 4 at the end!)
  pending_updates: `[]`

timvisee:
Noticed there were two more components that needed this treatment. I've implemented that in <https://github.com/qdrant/qdrant/pull/7805>.

PR #7811, opened 2025-12-19
Use RwLock in flushers, drop is alive lock early

timvisee:
Does two things to some of our flushers:
- Use `RwLock` instead of `Mutex` for pending changes, if we can have multiple concurrent readers
- `drop(is_alive_flush_guard)` before reconciliation - it is safe because we don't touch files there

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/gridstore/src/gridstore.rs
+++ b/lib/gridstore/src/gridstore.rs
@@ -622,7 +622,6 @@ impl<V> Gridstore<V> {
             bitmask_guard.flush()?;
 
             // Keep the guard till the end of the flush to prevent concurrent drop/flushes
-            // We don't touch files from here on and can drop the alive guard
             drop(is_alive_flush_guard);
 
             Ok(())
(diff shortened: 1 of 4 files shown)


lib/segment/src/common/rocksdb_buffered_delete_wrapper.rs
```diff
@@ -17,26 +17,17 @@ use crate::common::rocksdb_wrapper::{DatabaseColumnWrapper, LockedDatabaseColumn
...
 ///
 /// WARN: this structure is expected to be write-only.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct DatabaseColumnScheduledDeleteWrapper {
     db: DatabaseColumnWrapper,
-    deleted_pending_persistence: Arc<Mutex<AHashSet<Vec<u8>>>>,
-}
-
-impl Clone for DatabaseColumnScheduledDeleteWrapper {
```
coszio, line 26: 🧹

PR #7828, opened 2025-12-22
implement handelling for abandon Patrial shards after cancelled snapshot recovery

generall:
based in the test provided in https://github.com/qdrant/qdrant/pull/7762

Now, on the service restart, if node detects that it has `Partial` shards with no associated transfers,
it will either convert it into `Dead` or `Active` depending on the presence of other replicas.

### DIFF
--- a/lib/storage/src/content_manager/toc/mod.rs
+++ b/lib/storage/src/content_manager/toc/mod.rs
@@ -552,33 +552,24 @@ impl TableOfContent {
                     .iter()
                     .any(|(_peer_id, &replica_state)| replica_state.can_be_source_of_truth());
 
-                if has_other_source_of_truth {
-                    log::info!(
-                        "Cleaning up Partial replica state for collection {collection} shard {shard_id} on peer {this_peer_id}: marking as Dead",
-                        collection = collection.name(),
-                    );
-                    Self::send_set_replica_state_proposal_op(
-                        proposal_sender,
-                        collection.name().to_string(),
-                        this_peer_id,
-                        shard_id,
-                        ReplicaState::Dead,
-                        Some(ReplicaState::Partial),
-                    )?;
+                let target_state = if has_other_source_of_truth {
+                    ReplicaState::Dead
                 } else {
-                    log::info!(
-                        "Cleaning up Partial replica state for collection {collection} shard {shard_id} on peer {this_peer_id}: marking as Active",
-                        collection = collection.name(),
-                    );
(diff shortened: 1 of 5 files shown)


lib/storage/src/content_manager/toc/mod.rs
```diff
@@ -507,6 +507,86 @@ impl TableOfContent {
...
+            return Ok(());
+        };
+
+        for collection in collections.values() {
+            let state = collection.state().await;
+
+            let is_custom_sharding = match state.config.params.sharding_method.unwrap_or_default() {
+                ShardingMethod::Auto => false,
+                ShardingMethod::Custom => true,
+            };
```
timvisee, line 532: Nit: lets extract this into a function `ShardingMethod::is_custom_sharding()`

lib/storage/src/content_manager/toc/mod.rs
```diff
@@ -507,6 +507,86 @@ impl TableOfContent {
...
+                    .replicas
+                    .iter()
+                    .any(|(_peer_id, &replica_state)| replica_state.can_be_source_of_truth());
+
+                let target_state = if has_other_source_of_truth {
+                    Some(ReplicaState::Dead)
+                } else if is_custom_sharding {
+                    // This is a tiered multitenancy case, we didn't finish promotion,
+                    // but it is fine to leave Partial state here.
+                    None
```
timvisee, line 566: Even with tiered multitenancy we might leave replicas in bad state due to failed snapshots.

In my opinion just ignoring partial state with custom sharding is not good enough.

lib/storage/src/content_manager/toc/mod.rs
```diff
@@ -507,6 +507,86 @@ impl TableOfContent {
...
+
+                if has_inbound_transfers {
+                    // State resolution will be handled by the transfer completion or abortion.
+                    continue;
+                }
+
+                let has_other_source_of_truth = shard_info
+                    .replicas
+                    .iter()
+                    .any(|(_peer_id, &replica_state)| replica_state.can_be_source_of_truth());
```
timvisee, line 559: I'd argue this check is not valid because we don't wait for consensus to catch up before we call this function.

Ideally we wait for all pending operations to be applied, and only call this function after so that our view of 'active' replicas is most accurate.

Luckily consensus will reject state transitions that are disallowed based on the current state. But that also means we'll leave the replica in partial state.

PR #7830, opened 2025-12-24
Safe delete

xzfc:
This PR improves crash robustness a bit.

# Safe delete

In some places we use "rename then delete" approach to safely delete directories. But we still have some code paths that don't use this approach. This PR covers remaining places when we remove segments and collections.

Also it adds fsync on parent dir in places where another approach is used (touch/delete the sentinel file then delete directory). See `sync_parent_dir` comment.

Aside segment/collections, there are still some places that use direct `remove_dir_all` (WAL/indexes within a segment/etc). These are not fixed; I haven't checked them thoroughly and have no plan to fix them in this PR.

# Refactoring for clarity

This PR also brings small refactorings for clarity:

- Rename field `Segment::{current_path -> segment_path}`.
- Replace some `Path`/`PathBuf` with `str`/`String`.
- Replace `Result<Option<Segment>>` with `Result<LoadSegmentOutcome>`.
  Now the result of `load_segment` is more explicit:
  | before              | after                                     |
  |---------------------|-------------------------------------------|
  | `Ok(Some(segment))` | `Ok(LoadSegmentOutcome::Loaded(segment))` |
  | `Ok(None)`          | `Ok(LoadSegmentOutcome::Skipped)`         |
  | `Err(e)`            | `Err(e)`                                  |

### DIFF
--- a/lib/collection/src/shards/replica_set/snapshots.rs
+++ b/lib/collection/src/shards/replica_set/snapshots.rs
@@ -2,6 +2,7 @@ use std::collections::HashSet;
 use std::io;
 use std::path::Path;
 
+use ::io::safe_delete::{safe_delete_with_suffix, sync_parent_dir_async};
 use common::save_on_disk::SaveOnDisk;
 use common::tar_ext;
 use fs_err as fs;
@@ -217,6 +218,7 @@ impl ShardReplicaSet {
         let shard_flag = shard_initializing_flag_path(collection_path, self.shard_id);
         let flag_file = tokio_fs::File::create(&shard_flag).await?;
         flag_file.sync_all().await?;
+        sync_parent_dir_async(&shard_flag).await?;
 
         // Check `cancel` token one last time before starting non-cancellable section
         if cancel.is_cancelled() {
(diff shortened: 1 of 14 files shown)


lib/collection/src/shards/replica_set/snapshots.rs
```diff
@@ -293,14 +295,16 @@ impl ShardReplicaSet {
...
 
-                        tokio_fs::remove_dir_all(&segment_path)
-                            .await
-                            .map_err(|err| {
-                                CollectionError::service_error(format!(
-                                    "failed to remove outdated segment {}: {err}",
-                                    segment_path.display(),
-                                ))
-                            })?;
+                        tokio::task::spawn_blocking({
```
generall, line 298: would it be possible to avoid `spawn_blocking`? We had those problems with not enough threads recently, so if we can do it async, let's do it async

xzfc: WDYM?

`tokio::fs::delete_dir_all` is just `std::fs::delete_dir_all` inside `spawn_blocking`. (just like other `tokio::fs` wrappers)
In my code I did the same, just more verbose on the caller side.

generall: okay

generall (approved):

PR #7832, opened 2025-12-25
Add f16 NEON support for ARM Windows

### LINKED ISSUE #7718: Add f16 NEON support in ARM Windows
Fix f16 NEON compilation in ARM Windows for `segment` (`aarch64-pc-windows-msvc` target).

## Current Behavior
For now, f16 NEON acceleration is disabled for Windows ARM because of linking error between rust and C.

## Expected Behavior
It's needed to remove the disabling and fix the compilation for `aarch64-pc-windows-msvc`. Disabling was introduced while hotfixing this issue: https://github.com/qdrant/qdrant/issues/7686

## Possible Solution
Probably it's related to MCVS. It has different defines from GCC and Clang. But the first attempt didn't fix the problem:
https://github.com/qdrant/qdrant/pull/7691

majiayu000:
## Summary

This PR enables half-precision (f16) NEON SIMD acceleration on Windows ARM64 systems (`aarch64-pc-windows-msvc`).

Previously, f16 NEON was disabled on Windows due to linking errors between Rust and C code (see #7686). This was a hotfix that excluded Windows ARM from f16 NEON optimizations entirely.

### Changes

- **`lib/segment/src/spaces/metric_f16/mod.rs`**: Remove the `not(windows)` condition from the neon module cfg
- **`lib/segment/build.rs`**: Detect MSVC and use appropriate compiler flags (`/O2` instead of `-O3`, `-march`)
- **`lib/segment/src/spaces/metric_f16/cpp/neon.c`**: Add MSVC compatibility:
  - Use `arm64_neon.h` instead of `arm_neon.h` for MSVC
  - Add `QDRANT_FP16_NEON_ENABLED` macro for unified cross-platform feature detection
  - Provide `vabsh_f16` fallback implementation for MSVC compatibility

### Technical Details

The core issue was that MSVC does not define `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` (which GCC/Clang define), causing the C code to be empty and resulting in unresolved external symbols during linking.

The fix introduces a unified `QDRANT_FP16_NEON_ENABLED` macro that is set:
- On MSVC: Always enabled (MSVC ARM64 supports FP16 intrinsics)
- On GCC/Clang: Only when `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` is defined

## Test plan

- [x] Verified compilation on macOS ARM64 (Apple Silicon)
- [x] All f16 NEON tests pass (`cargo test -p segment --lib -- metric_f16`)
- [x] Full workspace compilation succeeds
- [ ] Needs testing on actual Windows ARM64 device

Fixes #7718

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/segment/build.rs
+++ b/lib/segment/build.rs
@@ -4,6 +4,9 @@ fn main() {
     let target_arch = env::var("CARGO_CFG_TARGET_ARCH")
         .expect("CARGO_CFG_TARGET_ARCH env-var is not defined or is not UTF-8");
 
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
+    let target_env = env::var("CARGO_CFG_TARGET_ENV").unwrap_or_default();
+
     // TODO: Is `CARGO_CFG_TARGET_FEATURE` *always* defined?
     //
     // Cargo docs says that "boolean configurations are present if they are set,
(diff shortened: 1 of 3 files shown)


nos1609:
@IvanPleshkov @JojiiOfficial submitted a PR closing the  
> Needs testing on actual Windows ARM64 device

PR #7850, opened 2026-01-02
CoW defines flush topology

generall:
Instead of implicit flush ordering, defined by segement type (which is flacky, as segment type might change randomly because of optimizations or snapshot process), we define explicit order based on used CoW operations.

This is an (hopefully easier) alternative for 2-stage flush.

### DIFF
--- a/lib/common/common/src/sort_utils.rs
+++ b/lib/common/common/src/sort_utils.rs
@@ -6,6 +6,12 @@ pub fn sort_permutation<T: Eq + Hash + Copy, V>(
     keys: &[T],
     sort: impl FnOnce(&[T]) -> Vec<T>,
 ) {
+    debug_assert_eq!(
+        values.len(),
+        keys.len(),
+        "values and keys must have the same length",
+    );
+
     if values.len() <= 1 {
         return;
     }
@@ -25,7 +31,7 @@ pub fn sort_permutation<T: Eq + Hash + Copy, V>(
     // Apply permutation in-place using cycle sort
 
     // Example:
-    // keys: [K1, k3, K0, K2]
+    // keys: [K1, K3, K0, K2]
     // sorted_keys: [K0, K1, K2, K3]
     // perm: [1, 3, 0, 2]
     for i in 0..perm.len() {


lib/shard/src/segment_holder/snapshot.rs
```diff
@@ -55,18 +55,7 @@ impl SegmentHolder {
...
-        // Re-sort segments for flush ordering, required to guarantee data consistency
-        // TODO: sort in a better place to not lock each segment
-        segment_ids.sort_by_cached_key(|segment_id| {
-            segments_lock
-                .get(*segment_id)
-                .unwrap()
-                .get()
-                .read()
-                .flush_ordering()
-        });
```
timvisee, line 67: We can indeed safely remove this.

We needed it before because we flushed each segment separately, which did need correct flush ordering. Now this logic is gone.

timvisee:
Test run in <https://github.com/qdrant/crasher/actions/runs/20657450147/job/59312956132>

lib/shard/src/segment_holder/mod.rs
```diff
@@ -691,7 +694,17 @@ impl SegmentHolder {
                 } else {
                     self.aloha_random_write(
                         &appendable_segments,
-                        |_appendable_idx, appendable_write_segment| {
+                        |appendable_idx, appendable_write_segment| {
+                            // If we are moving point from one segment to another,
+                            // we must guarantee, that data in new segment will be persister before
+                            // deleting point from old segment.
+                            // Do ensure that, we add flush
```
timvisee:
```suggestion
                            // Do ensure that, we add a flush dependency
```

lib/shard/src/segment_holder/flush.rs
```diff
@@ -49,21 +50,18 @@ impl SegmentHolder {
...
         //
         // Locking must happen in this order because `apply_points_to_appendable` can take two
         // write locks, also in this order. If we'd use different ordering we will eventually end
         // up with a deadlock.
         let mut segment_reads: Vec<_> = segments
             .iter()
-            .rev()
             .map(|segment| Self::aloha_lock_segment_read(segment))
             .collect();
-        segment_reads.reverse();
```
timvisee, line 63: I'm not confident we can just remove this double reverse.

See the comment above. Might be a deadlock footgun.

lib/shard/src/segment_holder/flush.rs
```diff
@@ -100,18 +109,38 @@ impl SegmentHolder {
...
+        let flush_topology = self.flush_dependency.lock().clone();
+        let mut iter = flush_topology.sort_elements(segment_ids);
+        let sorted_keys: Vec<_> = iter.by_ref().collect();
+
+        // Collect any remaining elements (circular dependencies)
+        let remaining = iter.into_unordered_vec();
+
+        debug_assert!(
+            remaining.is_empty(),
+            "circular dependencies detected in flush topology"
```
timvisee:
```suggestion
            "circular dependencies detected in flush topology: {remaining:?}",
```

lib/common/common/src/toposort.rs
```diff
@@ -0,0 +1,253 @@
...
+    fn test_missing_dependencies() {
+        let mut topo = TopoSort::new();
+        let elements = [3, 2, 1];
+        topo.add_dependency(3, 4, ()); // This one is extra and should be ignored
+        topo.add_dependency(2, 1, ());
+        topo.add_dependency(3, 2, ());
+        let result: Vec<_> = topo.sort_elements(&elements).collect();
+        // 4 is missing, so 3 should be treated as having no dependencies
+        assert_eq!(result, vec![1, 2, 3]);
+    }
```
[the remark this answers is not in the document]
agourlay: check again

lib/shard/src/segment_holder/flush.rs
generall: I think commet logic is preserved:

- We DO lock lock non-appendable first, then appendable.
- We sort fo flushing order after.

what am I missing?

lib/common/common/src/toposort.rs
```diff
@@ -0,0 +1,253 @@
...
+    /// This means `depends_on` must come before `element` in the sorted output.
+    ///
+    /// Additionally, stores a value associated with the dependency.
+    /// Overwrites any existing value for the same dependency.
+    pub fn add_dependency(&mut self, element: T, depends_on: T, value: V) {
+        self.dependencies
+            .entry(element)
+            .or_default()
+            .insert(depends_on, value);
+    }
```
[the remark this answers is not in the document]
timvisee: We always work with latest versions, which means that we can only insert the newest version. `max` therefore is not required.

timvisee:
Another crasher run just in case: https://github.com/qdrant/crasher/actions/runs/20820617456

PR #7852, opened 2026-01-03
restore snapshot in edge

generall:
- **Restore shard snapshot in Edge python bindings**

Main challenge of this PR is to make file operations universal in sync and async versions of the code.
To achieve this, I introduce two extra structures, which are intended to force both implementations to use all fields (files) mentioned in those strucutres


generall:
ToDo in follow-up PRs:

- remove support for rocksdb in partial snapshots
- Unify snapshot recovery in edge and server versions (use `SnapshotMergePlan`)
- make partial snapshots useful when modifying immutable segments (remove payload update in immutable segments)

PR #7854, opened 2026-01-04
remove rocksdb from creating snapshots path

generall:
Continuation for https://github.com/qdrant/qdrant/pull/7852

- remove support for rocksdb in partial snapshots. Partial snapshots will be GA definitely later than we completely remove rocksdb.
- Unify snapshot recovery in edge and server versions (use SnapshotMergePlan)

### DIFF
--- a/lib/segment/tests/integration/byte_storage_hnsw_test.rs
+++ b/lib/segment/tests/integration/byte_storage_hnsw_test.rs
@@ -106,12 +106,6 @@ fn test_byte_storage_hnsw(
             .vector_storage
             .borrow();
         let raw_storage: &VectorStorageEnum = &borrowed_storage;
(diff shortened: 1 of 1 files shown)


lib/segment/src/types.rs
```diff
@@ -1504,14 +1504,7 @@ pub enum VectorStorageType {
...
 impl Default for VectorStorageType {
     fn default() -> Self {
-        #[cfg(feature = "rocksdb")]
-        {
-            VectorStorageType::Memory
-        }
-        #[cfg(not(feature = "rocksdb"))]
-        {
-            VectorStorageType::InRamChunkedMmap
-        }
```
generall, line 1514: This was only used in tests. I don't think we need to test rocksdb anymore

PR #7856, opened 2026-01-05
Handle operations when having no shards with custom sharding

KShivendu:
When you have custom sharding but no shard keys in the collection, requests return error saying `Shard key not specified` which is wrong.

```http
# Request:
PUT /collections/demo/index
{
  "field_name": "group_id",
  "field_schema": {
    "type": "keyword",
    "is_tenant": true
  }
}

# Before:
{
  "error": "Wrong input: Shard key not specified"
}

# After:
{
  "result": {
    "status": "acknowledged"
  },
  "status": "ok",
  "time": 0.019617464
}
```

Also, this operation was already updating collection's `payload_index.json` but it isn't returned in the `GET /collections/benchmark` API's `payload_schema` field. So I've fixed that too. 

```http
# Request:
GET /collection/demo

# Before:
{
  "result": {
    // ...
    "payload_schema": {},
  "status": "ok",
  "time": 0.000223741
}

# After:
{
  "result": {
    // ...
    "payload_schema": {
      "group_id": {
        "data_type": "keyword",
        "params": {
          "type": "keyword",
          "is_tenant": true
        },
        "points": 0
      }
    }
  },
  "status": "ok",
  "time": 0.000223741
}
```

### DIFF
--- a/lib/collection/src/shards/shard_holder/mod.rs
+++ b/lib/collection/src/shards/shard_holder/mod.rs
@@ -154,8 +154,8 @@ impl ShardHolder {
         self.key_mapping.read().clone()
     }
 
-    pub fn get_sharding_method(&self) -> &ShardingMethod {
-        &self.sharding_method
+    pub fn get_sharding_method(&self) -> ShardingMethod {
+        self.sharding_method
     }
 
     /// Set the shard key mappings
--- a/lib/collection/src/collection/mod.rs
+++ b/lib/collection/src/collection/mod.rs
@@ -357,7 +357,7 @@ impl Collection {
     pub async fn get_sharding_method_and_keys(&self) -> (ShardingMethod, Vec<ShardKey>) {
         let shards_holder = self.shards_holder.read().await;
 
-        let sharding_method = shards_holder.get_sharding_method().clone();
+        let sharding_method = shards_holder.get_sharding_method();
         let shard_keys = shards_holder
             .get_shard_key_to_ids_mapping()
             .keys()


lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -74,7 +74,7 @@ pub struct ShardHolder {
     key_mapping: SaveOnDisk<ShardKeyMapping>,
     // Duplicates the information from `key_mapping` for faster access, does not use locking
     shard_id_to_key_mapping: AHashMap<ShardId, ShardKey>,
-    sharding_method: ShardingMethod,
+    pub(crate) sharding_method: ShardingMethod,
```
generall: please avoid making fields public and use getter instead

lib/storage/src/content_manager/toc/point_ops.rs
```diff
@@ -550,8 +551,15 @@ impl TableOfContent {
             }
 
             ShardSelectorInternal::All => {
-                let shard_keys = collection.get_shard_keys().await;
+                let (sharding_method, shard_keys) = collection.get_sharding_method_and_keys().await;
+
                 if shard_keys.is_empty() {
+                    if sharding_method == ShardingMethod::Custom {
```
generall: please prefer full match in cases like this

generall (changes requested):
Why do we want to reject index creation without shard present?

As far as I understand, we can create payload index if shard_key `s1` exist, and then create shard_key `s2` on creation will receive the same payload index automatically.

I would expect the same behavior without `s1`

KShivendu:
> I would expect the same behavior without s1

Yes it happens but payload index isn't shown in API response. Now that I think about it, we can just return acknowledge here and show payload index from collection if no shards are present. Did these changes.

generall (approved):

PR #7866, opened 2026-01-05
include optimization step into status

generall:
We need this to differenciate pending and skipped in UI


xzfc:
Alternative fix: #7873

PR #7872, opened 2026-01-06
Id tracker single write of pending updates

IvanPleshkov:
### Problem:

The mutable ID tracker collects all pending updates and stores them in the file in `flush`. 

But it is implemented as a separate `write` call for each mapping item. In the out-of-disk case, there is a bug where data may be written partially, and the mappings file will be corrupted.

### Solution:

This PR collects all pending updates into the temporary buffer first and then applies them as a single `write` call. Temporary buffer uses `try_set_capacity_exact` to catch OOM situations.

Please make attention while reviewing that file flushing was changed

### DIFF
--- a/lib/segment/src/id_tracker/mutable_id_tracker.rs
+++ b/lib/segment/src/id_tracker/mutable_id_tracker.rs
@@ -3,6 +3,7 @@ use std::io::{self, BufReader, BufWriter, Read, Seek, Write};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
+use ::io::counting_write::CountingWrite;
 use bitvec::prelude::{BitSlice, BitVec};
 use byteorder::{ReadBytesExt, WriteBytesExt};
 use common::is_alive_lock::IsAliveLock;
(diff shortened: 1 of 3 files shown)


lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -433,15 +476,11 @@ fn store_mapping_changes(
...
-    changes: &Vec<MappingChange>,
+    changes: &[MappingChange],
 ) -> OperationResult<()> {
     for &change in changes {
         write_entry(&mut writer, change)?;
     }
-
-    // Explicitly flush writer to catch IO errors
-    writer.flush()?;
-
```
generall, line 444: why is this removed?

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -395,27 +397,68 @@ fn versions_path(segment_path: &Path) -> PathBuf {
...
+        OperationError::service_error(format!(
+            "Failed to estimate ID tracker point mappings size ({}): {err}",
+            mappings_path.display(),
+        ))
+    })?;
+
+    // Create a buffer with exact required size.
+    let mut writer: Vec<u8> = Vec::default();
+    writer
+        .try_set_capacity_exact(counting_write.bytes_written() as usize)
```
generall, line 413: Is this atomic?

timvisee:
On second thought this approach has other problems.

Because now we might:
- grow mappings file by 100 bytes (fill with zeroes)
- write only 50 bytes of data
- crash

Which leaves us with a file having a lot of zeroes at the end. In that case we cannot properly determine what entries are valid.

This is superseded by the alternative in <https://github.com/qdrant/qdrant/pull/7877>

PR #7876, opened 2026-01-06
Faster gridstore model test with less clearing

coszio:
Improves `behave_like_hashmap` testing time by reducing the probability of clearing operations, which were taking a lot of time.

It also tries to mimic a more realistic workload by using different weights for each operation.

Locally, the 6 tests this spawns pass in about 10s now. It should be quicker on windows too now.


generall (approved):
Indeed faster now

timvisee:
Any good reason for clear taking so long? Is it just waiting on a big flush and file IO?

PR #7877, opened 2026-01-06
id tracker persisted mappings offset

IvanPleshkov:
This PR fixes the ID tracker flushing bug and continues PR https://github.com/qdrant/qdrant/pull/6173.

## Bug description

In the production, there is an error while ID tracker loading:
```
026-01-05T18:21:40.350395Z ERROR qdrant::startup: Panic occurred in file /qdrant/lib/collection/src/shards/replica_set/mod.rs at line 315: Failed to load local shard "./storage/collections/embeddings/0": Service internal error: Failed to load ID tracker mappings: Service runtime error: IO Error: Corrupted ID tracker mapping storage, got malformed mapping change byte 0x6D
```

## Possible repro steps

There are no confirmed repro steps yet, but there is a possible scenario of how it could happen:
1. Flushing is called.
2. ID Tracker tried to flush their pending mapping data but out-of-disk happened. ID tracked appends the pending changes into the file; the changes have been written partially(!!!).
3. Flush returned an error, WAL is not cleared
4. Something happened, and there is an additional disk space (the optimizer finishes the job, or another process cleaned something).
5. Wait until flush retry. From WAL, there are non-empty pending changes; however, because there is sufficient disk space, the pending changes were written successfully after the partially written data.
6. Tada! We got a malformed ID Tracker mapping file, which returns such an error after segment reloading.

In this scenario, file shrinking from here https://github.com/qdrant/qdrant/pull/6173 is not helping because the data is malformed, but we don't meet EOF.

## Solution

As a solution, this PR tracks the amount of successfully persisted bytes amount. In code, it's called `persisted_mappings_size`.

This `persisted_mappings_size` is equal to the file size at the beginning. It's increased after a successful flush, and the flush is starting appends at `persisted_mappings_size` position (using `file::Seek`).

## How it affects the described bug scenario

In the scenario of the possible repro steps, we don't update `persisted_mappings_size` if the flush failed. It means, at step 5, we won't append the unfinished file. We will override non-succesfull part in this file.

## Open questions
Do we need to shrink the file in the `persisted_mappings_size` is not equal to the file size while flushing?

### DIFF
--- a/lib/segment/src/id_tracker/mutable_id_tracker.rs
+++ b/lib/segment/src/id_tracker/mutable_id_tracker.rs
@@ -111,11 +111,14 @@ pub struct MutableIdTracker {
 
     is_alive_lock: IsAliveLock,
 
-    /// Size of correctly persisted mappings in bytes
+    /// Expected length of the mappings file in bytes
     ///
     /// We initialize this on load, and keep bumping it after reach successful flush. Pending
     /// changes are written to the file after this offset.
-    persisted_mappings_size: Arc<AtomicU64>,
+    ///
+    /// If we have more bytes on disk it probably indicates a partial flush. If we have less bytes
+    /// on disk we hit some kind of a bug.
+    mappings_expected_len: Arc<AtomicU64>,
 }
 
 impl MutableIdTracker {
@@ -142,7 +145,7 @@ impl MutableIdTracker {
             );
         }
 
-        let (mappings, persisted_mappings_size) = if has_mappings {
+        let (mappings, mappings_expected_len) = if has_mappings {
             load_mappings(&mappings_path).map_err(|err| {
                 OperationError::service_error(format!("Failed to load ID tracker mappings: {err}"))
             })?
@@ -181,7 +184,7 @@ impl MutableIdTracker {
             pending_versions: Default::default(),
             pending_mappings: Default::default(),
             is_alive_lock: IsAliveLock::new(),
-            persisted_mappings_size: Arc::new(AtomicU64::new(persisted_mappings_size)),
+            mappings_expected_len: Arc::new(AtomicU64::new(mappings_expected_len)),
         })
     }
 
@@ -312,7 +315,7 @@ impl IdTracker for MutableIdTracker {
 
         let is_alive_handle = self.is_alive_lock.handle();
         let pending_mappings_weak = Arc::downgrade(&self.pending_mappings);
-        let persisted_mappings_size = self.persisted_mappings_size.clone();
+        let mappings_expected_len = self.mappings_expected_len.clone();
 
         Box::new(move || {
             let (Some(is_alive_guard), Some(pending_mappings_arc)) = (
@@ -322,14 +325,13 @@ impl IdTracker for MutableIdTracker {
                 return Ok(());
             };
 
-            let stored = store_mapping_changes(&mappings_path, &changes, &persisted_mappings_size);
+            let stored = store_mapping_changes(&mappings_path, &changes, &mappings_expected_len);
 
             // If persisting mappings failed, try to truncate mappings file to what we had before
             // in an best effort to get rid of partially persisted mappings. We can safely ignore
             // truncate errors because load should properly handle partial entries as well.
             if let Err(err) = stored {
-                let expected_len =
-                    persisted_mappings_size.load(std::sync::atomic::Ordering::Relaxed);
+                let expected_len = mappings_expected_len.load(std::sync::atomic::Ordering::Relaxed);
                 let truncate_result = File::options()
                     .write(true)
                     .open(&mappings_path)


lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -398,12 +404,40 @@ fn versions_path(segment_path: &Path) -> PathBuf {
...
+        std::cmp::Ordering::Greater => {
+            // File is larger than persisted mappings size, seek to the end of persisted mappings
+            file.seek(io::SeekFrom::Start(file_start_appending))?;
+        }
+        std::cmp::Ordering::Less => {
+            return Err(OperationError::service_error(format!(
+                "Mutable ID tracker mappings file size is less than persisted mappings size, cannot append new mappings (file size: {file_len}, persisted mappings size: {file_start_appending})",
+            )));
+        }
+    }
```
IvanPleshkov: Thanks! Didnt know that append mode ignores seek. Fixed!

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -420,6 +454,10 @@ fn store_mapping_changes(
         OperationError::service_error(format!("Failed to fsync ID tracker point mappings: {err}"))
     })?;
 
+    // Update persisted mappings size.
+    let appended_file_size = file.metadata()?.len();
+    persisted_mappings_size.store(appended_file_size, std::sync::atomic::Ordering::Relaxed);
+
```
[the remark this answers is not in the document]
IvanPleshkov: Thanks! It's a non-obvious problem. Fixed!

timvisee:
I wonder if a better alternative would be to try and grow the file to the exact required size on flush. If we don't have enough space, we'll abort without touching the file.

We currently grow the file automatically by writing to it, which means it's more likely to abort outside of entry boundaries.

One edge case with my suggestion is that we might have a bunch of zeroes at the end of the file. If is possible if we crash half way during the flush while we've already grown the file. But, we can easily detect this as there's no valid `0x00` [entry](https://github.com/qdrant/qdrant/blob/ecf6554261f19080115c4631eee18ae90959bf09/lib/segment/src/id_tracker/mutable_id_tracker.rs#L45-L52) (did this on purpose while building this tracker yay!).

Speaking of which: a crash during flush is likely another scenario to reproduce the problem you described - writing a partial entry.

Thoughts?

IvanPleshkov:
> I wonder if a better alternative would be to try and grow the file to the exact required size on flush. If we don't have enough space, we'll abort without touching the file.
> 
> We currently grow the file automatically by writing to it, which means it's more likely to abort outside of entry boundaries.
> 
> One edge case with my suggestion is that we might have a bunch of zeroes at the end of the file. If is possible if we crash half way during the flush while we've already grown the file. But, we can easily detect this as there's no valid `0x00` [entry](https://github.com/qdrant/qdrant/blob/ecf6554261f19080115c4631eee18ae90959bf09/lib/segment/src/id_tracker/mutable_id_tracker.rs#L45-L52) (did this on purpose while building this tracker yay!).
> 
> Speaking of which: a crash during flush is likely another scenario to reproduce the problem you described - writing a partial entry.
> 
> Thoughts?

As we discussed with @agourlay, in general, we cannot check if the data in the file is correct. We can check the entry, but if the malformed data is the last bytes after the last entry, we cannot check it - the external id may be any possible value. In case of out-of-space it's fine, but I'm not sure if this proposal works fine in case of crash

IvanPleshkov:
> I wonder if a better alternative would be to try and grow the file to the exact required size on flush. If we don't have enough space, we'll abort without touching the file.
> 
> We currently grow the file automatically by writing to it, which means it's more likely to abort outside of entry boundaries.
> 
> One edge case with my suggestion is that we might have a bunch of zeroes at the end of the file. If is possible if we crash half way during the flush while we've already grown the file. But, we can easily detect this as there's no valid `0x00` [entry](https://github.com/qdrant/qdrant/blob/ecf6554261f19080115c4631eee18ae90959bf09/lib/segment/src/id_tracker/mutable_id_tracker.rs#L45-L52) (did this on purpose while building this tracker yay!).
> 
> Speaking of which: a crash during flush is likely another scenario to reproduce the problem you described - writing a partial entry.
> 
> Thoughts?

There is a draft PR with collecting all pending mappings into a single buffer and writing it in one shot. It contains the size estimation of the pending changes and is a partial implementation of your proposal, we can extend it to implement your idea
https://github.com/qdrant/qdrant/pull/7872

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -398,12 +404,40 @@ fn versions_path(segment_path: &Path) -> PathBuf {
...
+            ))
+        })?
+        .len();
+
+    // Seek to the end of the persisted part of the file.
+    // If previously call of `store_mapping_changes` failed,
+    // `persisted_mappings_size` may be less than actual file size.
+    let file_start_appending = persisted_mappings_size.load(std::sync::atomic::Ordering::Relaxed);
+    if file_start_appending <= file_len {
+        file.seek(io::SeekFrom::Start(file_start_appending))
```
timvisee: Idea: how about we truncate the file here instead (or do both). Then we also 'repair' the file for the next load if we don't finish this flush.

If we hit this case, we expect to always write the exact same sequence of bytes. And so it should not corrupt the file even _more_. But truncating may be a bit safer.

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -413,13 +447,29 @@ fn store_mapping_changes(
...
-    let file = writer.into_inner().unwrap();
+    let mut file = writer.into_inner().map_err(|err| {
+        OperationError::service_error(format!(
+            "Failed to flush ID tracker point mappings write buffer: {err}"
+        ))
+    })?;
+
+    // Get new persisted size as a position after writing all changes.
+    // Stream position is used here ho handle cases where the pending changes are shorter than non-persisted part.
+    let new_persisted_size = file.stream_position().map_err(|err| {
```
timvisee, line 487: TODO(timvisee): check potential off by one error.

lib/segment/src/id_tracker/mutable_id_tracker.rs
timvisee: Implemented in <https://github.com/qdrant/qdrant/pull/7877/commits/31752068fec5e80f78fa25f99a8db1024e523ee4>

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -398,12 +426,53 @@ fn versions_path(segment_path: &Path) -> PathBuf {
...
+                    ))
+                })?;
+        }
+        // File is smaller than expected, indicates a bug we cannot recover from
+        Ordering::Less => {
+            return Err(OperationError::service_error(format!(
+                "Mutable ID tracker mappings file size is less than persisted mappings size, cannot append new mappings (file size: {file_len}, persisted mappings size: {file_start_appending})",
+            )));
+        }
+    }
```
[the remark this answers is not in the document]
timvisee: This is indeed problematic, the behavior is described here: https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.append

This means we can:
- require the truncate to succeed
- don't seek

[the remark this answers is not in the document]
timvisee: Fixed in <https://github.com/qdrant/qdrant/pull/7877/commits/b28fe473f8493cd0fd85e4793cda750f706ebc0a>

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -312,7 +325,24 @@ impl IdTracker for MutableIdTracker {
                 return Ok(());
             };
 
-            store_mapping_changes(&mappings_path, &changes)?;
+            let stored = store_mapping_changes(&mappings_path, &changes, &mappings_expected_len);
+
+            // If persisting mappings failed, try to truncate mappings file to what we had before
+            // in an best effort to get rid of partially persisted mappings. We can safely ignore
+            // truncate errors because load should properly handle partial entries as well.
+            if let Err(err) = stored {
```
agourlay, line 333: it seems `store_mapping_changes` can throw a bunch of errors, I am not sure about matching on any error returned here.

Maybe we could at least check if the file len changed before trying to set it back?

timvisee: If the file wasn't touched, the following 'truncation' will be a no op.

Do you think it's still better to isolate the error paths?

agourlay: On a second though I think it is fine, wanted to have your opinion.
Isolating error path will be super verbose.

timvisee: Yeah.

There is extra work to re-open the file to allow truncation. But it only happens in the error path.

lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -413,13 +475,29 @@ fn store_mapping_changes(
...
-    let file = writer.into_inner().unwrap();
+    let mut file = writer.into_inner().map_err(|err| {
+        OperationError::service_error(format!(
+            "Failed to flush ID tracker point mappings write buffer: {err}"
+        ))
+    })?;
+
+    // Get new persisted size as a position after writing all changes.
+    // Stream position is used here to handle cases where the pending changes are shorter than non-persisted part.
+    let new_persisted_size = file.stream_position().map_err(|err| {
```
agourlay, line 487: Did you test this `stream_position` thingy?

Is it supposed to give you the current cursor position after all the writes?

timvisee: Yes.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=aca587b249281c64c5c478f42b890273

In this case it gives us the size of the file in bytes, or the position we write the first new byte to.

agourlay (approved):

IvanPleshkov:
I tested this PR locally using tmpfs (filesystem in RAM) with small disk size (8MB). Test code with repro steps. It shows the same error `Corrupted ID tracker mapping storage, got malformed mapping change byte 0x00`. This PR fixes this problem

```rust
    #[test]
    fn mutable_id_tracker_overflow() {
        // First, create a file of 1MB size to simulate later disk space cleaning.
        let tmp_file_path = Path::new("/mnt/tiny/tmp_file");
        if fs::exists(&tmp_file_path).unwrap() {
            fs::remove_file(&tmp_file_path).unwrap();
        }
        let mut tmp_file = fs::File::create(&tmp_file_path).unwrap();
        tmp_file.write(&vec![0; 1024 * 1024]).unwrap();
        tmp_file.flush().unwrap();
        tmp_file.sync_all().unwrap();
        drop(tmp_file);

        // Then, create ID tracker and fill it while possible.
        let path = Path::new("/mnt/tiny/id_tracker_overflow_test");
        if fs::exists(path).unwrap() {
            fs::remove_dir_all(path).unwrap();
        }
        std::fs::create_dir_all(path).unwrap();
        let mut id_tracker = make_mutable_tracker(&path);
        for i in 0.. {
            let internal_id = id_tracker.total_point_count() as PointOffsetType;
            let point_id = PointIdType::NumId(internal_id as u64);
            match id_tracker.set_link(point_id, internal_id) {
                Ok(_) => {}
                Err(err) => {
                    println!("Stopped at iteration {i} due to error: {err}");
                    break;
                }
            }

            // flush each 1000 points
            if i % 1000 == 0 {
                match id_tracker.mapping_flusher()() {
                    Ok(_) => {}
                    Err(err) => {
                        println!("Flushing stopped at iteration {i} due to error: {err}");
                        break;
                    }
                }
            }
        }

        // check that free space is not available
        let fs_stats = fs2::available_space(path).unwrap();
        println!("Available space before cleanup: {fs_stats} bytes");

        // then, add additional space by removing the temporary file.
        fs::remove_file(&tmp_file_path).unwrap();

        // flush file deletion to ensure the OS registers the freed space
        #[cfg(unix)]
        {
            use std::process::Command;
            // Force filesystem sync
            let _ = Command::new("sync").output();
        }
        std::thread::sleep(std::time::Duration::from_secs(3)); // wait for the system to register the freed space

        // check that free space is available
        let fs_stats = fs2::available_space(path).unwrap();
        println!("Available space after cleanup: {fs_stats} bytes");

        // Add some more points to verify that ID tracker can continue working.
        for i in 0..10 {
            let point_id = PointIdType::Uuid(Uuid::new_v4());
            let internal_id = id_tracker.total_point_count() as PointOffsetType;
            match id_tracker.set_link(point_id, internal_id) {
                Ok(_) => {}
                Err(err) => {
                    println!("Stopped at iteration {i} due to error: {err}");
                    break;
                }
            }
        }

        // Flush has to work at the end.
        id_tracker.mapping_flusher()().unwrap();

        // Reload ID tracker to verify data integrity.
        drop(id_tracker);

        // The bug appears here with some probability.
        MutableIdTracker::open(&path).unwrap();
    }

```

timvisee:
> Corrupted ID tracker mapping storage, got malformed mapping change byte 0x00

~Actually, this is critical.~

~This shows that we do get 0's in places where we don't expect it. It means the file was grown without actually writing bytes to it.~

Turns out that message was from before the fix. No pending concerns then. :+1:

PR #7883, opened 2026-01-08
Use syncfs

xzfc:
# Problem

Restoring snapshot that contains a lot of small files takes a lot of time.
This happen because we set [`tar::Archive::set_sync`](https://github.com/qdrant/tar-rs/blob/856dbd090eede1736604f23cfe99a104b5639734/src/archive.rs#L164-L170) (from our tar-rs fork) that calls `fsync` on every file in a sequence. Fsyncing each file individually is slow.

# Solution

Disable `set_sync` and use `syncfs` instead. `syncfs` syncs the whole filesystem at once, not just a particular single file.
A potential drawback is that background writes (outside of snapshot restore code) can slow `syncfs` down, but I think it's a worthy tradeoff.

Also, since `set_sync` was the last thing from our `tar-rs` fork, we can switch to the upstream `tar-rs`.

# Alternatives & synthetic benchmarks

I've considered the

1. `fsync` in sequence (baseline).
2. `sync_file_range` in sequence then `fsync` in sequence.
   - `sync_file_range` acts as a hint for the kernel to schedule flushing. It is non-blocking.
   - `fsync` acts as a barrier until the scheduled flush is done.
   - Idea taken from https://github.com/postgres/postgres/blob/31ddbb38eeff60ad5353768c7416fea3a0ecafce/src/common/file_utils.c. Also, see https://bonsaidb.io/blog/durable-writes/.
3. `syncfs` on a single file.
4. `io_uring`: call `fsync` through `io_uring` pipeline.

Results and code: https://gist.github.com/xzfc/b22591fa5a5a192a1574857b00d7cb8d#file-results-md

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1281,6 +1281,7 @@ dependencies = [
  "log",
  "memmap2",
  "memory",
+ "nix 0.30.1",
  "num-traits",
  "num_cpus",
  "ordered-float 5.1.0",
--- a/lib/collection/src/shards/shard_holder/mod.rs
+++ b/lib/collection/src/shards/shard_holder/mod.rs
@@ -1267,6 +1267,7 @@ impl ShardHolder {
                         this_peer_id,
                         is_distributed,
                     )?;
+                    common::fs::bulk_sync_dir(&snapshot_temp_dir)?;
 
                     Ok(())
                 },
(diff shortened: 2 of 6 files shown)


timvisee (approved):
Simple and effective, brilliant!

I like that we can now use upstream `tar-rs`.

PR #7885, opened 2026-01-08
Improve logs regarding cleanup on startup

agourlay:
Prints information about shards and segments being modified for investigation.

### DIFF
--- a/lib/segment/src/segment/segment_ops.rs
+++ b/lib/segment/src/segment/segment_ops.rs
@@ -469,7 +469,7 @@ impl Segment {
     /// Check consistency of the segment's data and repair it if possible.
     /// Removes partially persisted points.
     pub fn check_consistency_and_repair(&mut self) -> OperationResult<()> {
-        // Get rid of versionless points.
+        // Get rid of mappingless points.
         let ids_to_clean = self.fix_id_tracker_inconsistencies()?;
 
         // There are some leftovers to clean from segment.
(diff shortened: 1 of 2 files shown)


lib/segment/src/segment/segment_ops.rs
```diff
@@ -637,7 +641,7 @@ impl Segment {
     }
 
     /// Fixes inconsistencies in the ID tracker, if any.
-    /// Returns list of IDs, which should be removed from segment
+    /// Returns list of IDs without mappings which should be removed from segment
     pub fn fix_id_tracker_inconsistencies(&mut self) -> OperationResult<Vec<PointOffsetType>> {
         self.id_tracker.borrow_mut().fix_inconsistencies()
```
agourlay, line 646: here is the source of truth

https://github.com/qdrant/qdrant/blob/f937d0260ffd7705c6f34d9c25da1e27ebf5ddfd/lib/segment/src/id_tracker/id_tracker_base.rs#L172

lib/segment/src/segment/segment_ops.rs
```diff
@@ -479,7 +479,11 @@ impl Segment {
         // This is internal operation, no hw measurement needed
         let disposable_hw_counter = HardwareCounterCell::disposable();
         if !ids_to_clean.is_empty() {
-            log::debug!("cleaning up {} points without version", ids_to_clean.len());
+            log::debug!(
+                "Cleaning up {} points with version but no mapping in segment {:?}",
+                ids_to_clean.len(),
+                self.data_path()
```
timvisee, line 485:
```suggestion
                self.data_path(),
```

🤡

timvisee (approved):

PR #7893, opened 2026-01-10
disposable hygienics

generall:
no not overuse HwMeasurementAcc::disposable everywhere, for better code navigation

### DIFF
--- a/lib/shard/src/query/mmr/tests.rs
+++ b/lib/shard/src/query/mmr/tests.rs
@@ -216,14 +216,8 @@ fn test_mmr_points_without_required_vector() {
         candidates_limit: 100,
     };
 
-    let result = mmr_from_points_with_vector(
-        points,
-        mmr,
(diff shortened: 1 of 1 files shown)


lib/common/common/src/counter/hardware_accumulator.rs
```diff
@@ -119,6 +119,7 @@ impl HwMeasurementAcc {
     /// Create a disposable accumulator, which will not accumulate any values.
     /// WARNING: This is intended for specific internal use-cases only.
     /// DO NOT use it in tests or if you don't know what you're doing.
+    /// Prefer `new` to be used in tests.
```
timvisee, line 122: Do I understand correctly that you'd just like to separate call sites over different functions to make 'jump to usages' more useful?

generall: exactly

PR #7919, opened 2026-01-15
fix: improve datetime_range RFC3339 parse error

### LINKED ISSUE #3531: Better error response for wrong datetime format in REST filter

krapcys1-maker:
Fixes #3531 

### Problem
When `datetime_range` in REST/JSON filter contains an invalid datetime (e.g. missing timezone offset / malformed),
the API returned a generic serde error like:
`data did not match any variant ... Condition`.

### Solution
- Make datetime parsing errors explicit and user-friendly.
- The error now includes the invalid value, mentions RFC3339, and provides an example format.
- Prevent untagged enum deserialization from swallowing datetime parse errors.

Example error:
`'2014-01-01T00:00:00BAD' does not match accepted datetime format (RFC3339). Example: 2014-01-01T00:00:00Z`

### Tests
- `cargo test -p segment`
- [x] I checked there are no duplicate PRs for this issue.
- [x] I ran tests locally (`cargo test -p segment`).
 
/claim #3531

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -86,17 +86,12 @@ impl<'de> Deserialize<'de> for DateTimePayloadType {
     /// Returns a clear user-facing error when the format is invalid.
     /// Example accepted value: `2014-01-01T00:00:00Z`.
     ///
-    /// For binary formats (CBOR/MessagePack/WAL), delegates to chrono's deserializer
-    /// which handles timestamps and other non-string representations.
+    /// Binary formats (CBOR/MessagePack/WAL) also serialize as RFC3339 strings, so we reuse
+    /// the same parsing path everywhere.
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: Deserializer<'de>,
     {
-        if !deserializer.is_human_readable() {
-            return chrono::DateTime::<chrono::Utc>::deserialize(deserializer)
-                .map(DateTimePayloadType::from);
-        }
-
         let str_datetime: Cow<'de, str> = Cow::deserialize(deserializer)?;
 
         match DateTimePayloadType::from_str(str_datetime.as_ref()) {
@@ -4075,7 +4070,7 @@ mod tests {
     }
 
     /// Regression test: DateTimePayloadType binary serialization roundtrip.
-    /// Ensures the `!is_human_readable()` path in DateTimePayloadType::deserialize works.
+    /// Ensures DateTimePayloadType parses binary-encoded RFC3339 strings.
     #[test]
     fn test_datetime_payload_type_binary_roundtrip() {
         let original = DateTimePayloadType::from_str("2024-06-15T12:30:45Z").unwrap();


lib/segment/src/types.rs
```diff
@@ -2620,6 +2624,48 @@ impl Hash for RangeInterface {
...
+        // Fallback to existing untagged behavior
+        let parsed = serde_json::from_value::<RangeInterfaceUntagged>(value)
+            .map_err(serde::de::Error::custom)?;
+
+        Ok(match parsed {
+            RangeInterfaceUntagged::Float(r) => RangeInterface::Float(r),
+            RangeInterfaceUntagged::DateTime(r) => RangeInterface::DateTime(r),
+        })
+    }
+}
```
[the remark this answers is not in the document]
krapcys1-maker: Good catch. I’ll restrict the serde_json::Value intermediary to human-readable deserializers only.
For binary formats (CBOR/MessagePack/Bincode) we’ll fall back to the plain untagged deserialization path,
so datetime-string detection and the extra overhead apply only to REST/JSON-like inputs.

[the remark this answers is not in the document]
krapcys1-maker: ok

lib/segment/src/types.rs
```diff
@@ -3234,6 +3287,103 @@ pub enum Condition {
...
+            }
+
+            serde_json::from_value::<ConditionUntagged>(value)
+                .map(Condition::from)
+                .map_err(serde::de::Error::custom)
+        } else {
+            ConditionUntagged::deserialize(deserializer).map(Condition::from)
+        }
+    }
+}
```
timvisee, line 3357: Do you think there's any way to eliminate this part?

It seems to contain a lot of boiler plate. It's also easy to forget changes here when more conditions are added because the compiler won't tell us so.

krapcys1-maker: Good question. I agree this part is somewhat boilerplate-heavy and not ideal from a maintenance perspective.
Unfortunately this is mostly intentional. In this case we can’t rely on a purely #[serde(untagged)] enum, because untagged deserialization can silently “swallow” values that look like valid datetime ranges (RFC3339), which is exactly what this change is trying to prevent. The explicit contains_key routing is there to control the deserialization order and fail fast on ambiguous inputs.
You’re also right that the compiler won’t help if a new condition variant is added and this logic isn’t updated. I see this as a trade-off between:
explicit, predictable parsing with clear error behavior, and
less boilerplate but weaker guarantees around ambiguous inputs.
There may be ways to reduce repetition (e.g. a small helper or macro, or a more explicit tagged representation), but they would add additional complexity and still need to preserve the current ordering and error semantics. I wanted to keep this fix minimal and focused on correctness rather than introducing a larger refactor.
If you have a specific pattern in mind that keeps the same parsing guarantees, I’m happy to explore it.

krapcys1-maker:
I pushed a small follow-up commit to reduce the boilerplate in the custom Condition deserializer.
The explicit contains_key routing is now wrapped in a tiny macro, keeping the same ordering (FieldCondition first) so datetime parse errors remain visible, but making it easier to maintain.

lib/segment/src/types.rs
generall: I am less concerned with boilerplate, but not heaving a compiler-level check is a deal-breaker

lib/segment/src/types.rs
```diff
@@ -82,16 +82,19 @@ impl DateTimeWrapper {
...
-            None => Err(serde::de::Error::custom(format!(
-                "'{str_datetime}' is not in a supported date/time format, please use RFC 3339"
+        let str_datetime: Cow<'de, str> = Cow::deserialize(deserializer)?;
+
+        match DateTimePayloadType::from_str(str_datetime.as_ref()) {
+            Ok(datetime) => Ok(datetime),
+            Err(_) => Err(serde::de::Error::custom(format!(
+                "'{str_datetime}' does not match accepted datetime format (RFC3339). Example: 2014-01-01T00:00:00Z"
             ))),
         }
```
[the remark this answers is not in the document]
krapcys1-maker: Good catch — added is_human_readable() guard in DateTimePayloadType::deserialize.
For non-human-readable formats (CBOR/MessagePack/WAL) we now delegate to chrono::DateTime<Utc>::deserialize, keeping binary timestamp compatibility while preserving the RFC3339 string error for JSON.

lib/segment/src/types.rs
krapcys1-maker: I’ll drop the explicit routing list and keep only a single special-case for FieldCondition ("key"), so we preserve the datetime parse error (FieldCondition tried first) while all other variants are delegated to ConditionUntagged.
This way new Condition variants won’t require updating this deserializer logic.

krapcys1-maker: dropped the explicit routing list and kept only a single FieldCondition (“key”) special-case so RFC3339 datetime parse errors remain visible.
All other variants are now delegated to ConditionUntagged, so adding new Condition variants no longer requires updating this deserializer (compiler-level safety restored).
Also added an is_human_readable() guard for DateTimePayloadType to preserve binary (CBOR/MessagePack/WAL) compatibility.
cargo test -p segment and cargo clippy -p segment --lib pass locally.

timvisee:
@krapcys1-maker could you please resolve the outstanding linter issues?

See:

```bash
cargo clippy --workspace --all-targets -- -D warnings
```

krapcys1-maker:
Hey! I fixed the clippy uninlined_format_args warnings in lib/segment/src/types.rs and added regression tests around datetime deserialization (binary + JSON fallback).
CI should be clean now.
I also reverted the unrelated formatting-only changes so the PR stays focused.

timvisee:
Thank you.

The approach is still more complex than I'd like for this purpose, but I guess we cannot easily simplify it because of how serde works. So I'm fine with keeping it like this.

lib/segment/src/types.rs
```diff
@@ -82,16 +82,27 @@ impl DateTimeWrapper {
...
-        let str_datetime = <&str>::deserialize(deserializer)?;
-        let parse_result = DateTimePayloadType::from_str(str_datetime).ok();
-        match parse_result {
-            Some(datetime) => Ok(datetime),
-            None => Err(serde::de::Error::custom(format!(
-                "'{str_datetime}' is not in a supported date/time format, please use RFC 3339"
+        if !deserializer.is_human_readable() {
+            return chrono::DateTime::<chrono::Utc>::deserialize(deserializer)
+                .map(DateTimePayloadType::from);
+        }
```
xzfc: `chrono::DateTime::deserialize` always deserializes through string:
https://github.com/chronotope/chrono/blob/v0.4.43/src/datetime/serde.rs#L93-L100
So `is_human_readable` check is not required.

krapcys1-maker: I removed the `is_human_readable()` branch. `chrono::DateTime::deserialize` indeed goes through the RFC3339 string path anyway, so the extra check was unnecessary.

Now `DateTimePayloadType` always deserializes via the same RFC3339 parsing path (human-readable + binary), keeping the friendly RFC3339 error message.

Ran:
- cargo fmt
- cargo test -p segment --lib ✅
- cargo clippy -p segment --lib --tests ✅

timvisee:
Thanks a lot! Merging now. :smile:

krapcys1-maker:
Thank you :)

PR #7921, opened 2026-01-16
Optimize SegmentBuilder update with proper Vec sizing

agourlay:
The vector resizing shows up in profiler at around 1% CPU time for the update threads when running the optimizer. 

Pre-allocating the vector based on the maximum number of available points is a low hanging fruit.

### DIFF
--- a/lib/segment/src/segment_constructor/segment_builder.rs
+++ b/lib/segment/src/segment_constructor/segment_builder.rs
@@ -295,8 +295,12 @@ impl SegmentBuilder {
             return Err(OperationError::service_error("Too many segments to update"));
         }
 
-        let mut points_to_insert = Vec::new();
         let locked_id_trackers = segments.iter().map(|s| s.id_tracker.borrow()).collect_vec();
+        let max_point_count = locked_id_trackers
+            .iter()
+            .map(|id_tracker| id_tracker.available_point_count())
+            .max();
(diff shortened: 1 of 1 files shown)


lib/segment/src/segment_constructor/segment_builder.rs
```diff
@@ -295,8 +295,12 @@ impl SegmentBuilder {
             return Err(OperationError::service_error("Too many segments to update"));
         }
 
-        let mut points_to_insert = Vec::new();
         let locked_id_trackers = segments.iter().map(|s| s.id_tracker.borrow()).collect_vec();
+        let max_point_count = locked_id_trackers
+            .iter()
+            .map(|id_tracker| id_tracker.available_point_count())
+            .max();
```
agourlay, line 302: could be the `sum` if the segments are completely disjoint but I was unsure about allocating a very large vector when there is a lot of overlap 🤷

timvisee: I think this is reasonable.

It means we'd have at most $\log_{2}\left(x\right)$ re-allocations where x is the number of segments.

On the other hand, we don't expect the available point count in each segment to overlap because we delete duplicates in different stages.

agourlay: Going to go ahead and merge this one as it removed the hotspot, so it is strictly better.

We can revisit later a more precise sizing.

PR #7928, opened 2026-01-16
async scroll

generall:
This is an experimental PR aimed to test a hypothesis of slow reads during shard transfers being caused by sequential reads.

This RP changes approach of how we read vectors for scroll operation: instead of reading vectors one-by-one sequntially, it reads batch of vectors. Each batch is processed with `io_uring` if enabled. 

Concurrent reads of vectors significancly decrease time to transfer shards, especially when receiving side is not empty (see charts below)

### DIFF
--- a/lib/segment/src/segment/segment_ops.rs
+++ b/lib/segment/src/segment/segment_ops.rs
@@ -417,8 +417,8 @@ impl Segment {
             if total_vectors <= point_offset as usize {
                 debug_assert!(
                     false,
-                    "Vector storage is inconsistent, total_vector_count: {}, point_offset: {}, external_id: {:?}",
-                    total_vectors, point_offset, id_tracker.external_id(point_offset)
+                    "Vector storage is inconsistent, total_vector_count: {total_vectors}, point_offset: {point_offset}, external_id: {:?}",
+                    id_tracker.external_id(point_offset),
                 );
                 return false;
             }


generall:
left to right: async_io transfer, regular transfer, transfer with madvise=sequential

image

generall:
Transfer under higher memory preasure + UUID instead of sequential.

image

Left: sequential reads
Right: async IO

lib/segment/src/data_types/segment_record.rs
```diff
@@ -0,0 +1,20 @@
+use crate::data_types::vectors::VectorInternal;
+use crate::types::{Payload, PointIdType, VectorNameBuf};
+
+pub type NamedVectorsOwned = Vec<(VectorNameBuf, VectorInternal)>;
+
+pub struct SegmentRecord {
+    pub id: PointIdType,
+    pub vectors: Option<NamedVectorsOwned>,
+    pub payload: Option<Payload>,
+}
```
generall, line 10: inctead of exposing `vectors` and `payloads` functions separatelly, segment now expose `retrieve` fucntion, which returns this. 

Reason for this is that we can't have `impl ..` with `dyn SegmentEntry` so it is hard to expose callback function in this trait

lib/segment/src/entry/entry_point.rs
```diff
@@ -139,6 +140,20 @@ pub trait SegmentEntry: SnapshotEntry {
...
+    /// This function may return fewer records than requested, if some points are not found.
+    /// Order of returned records is not guaranteed to match order of requested point ids.
+    fn retrieve(
+        &self,
+        point_ids: &[PointIdType],
+        with_payload: &WithPayload,
+        with_vector: &WithVector,
+        hw_counter: &HardwareCounterCell,
+        is_stopped: &AtomicBool,
+    ) -> OperationResult<Vec<SegmentRecord>>;
```
generall, line 155: This function, actually, the only thing we need, `vectors` `all_vectors` and `payloads` are now only used in a few not-so-important places, which can be refactored out later

lib/segment/src/segment/search.rs
```diff
@@ -25,72 +27,81 @@ impl Segment {
...
-                    }
-                }
+                let point_id = id_tracker.external_id(point_offset);
+                // This can happen if point was modified between retrieving and post-processing
+                // But this function locks the segment, so it can't be modified during its execution
+                debug_assert!(
+                    point_id.is_some(),
+                    "Point with internal ID {point_offset} not found in id tracker"
+                );
+                point_id.map(|id| (id, scored_point_offset))
```
generall, line 44: Handling of unexpected situations changed a bit in this function, mostly because it is a bit harder to return error from batch of points instead of single one.

But I think it is fine, as we don't want to crash production if some edge-case point is broken.

lib/segment/src/segment/segment_ops.rs
```diff
@@ -384,54 +384,60 @@ impl Segment {
...
+        self.vectors_by_offsets(
+            vector_name,
+            std::iter::once(point_offset),
+            hw_counter,
+            |_, vector_internal| {
+                result = Some(vector_internal);
+            },
+        )?;
+        Ok(result)
+    }
```
generall, line 397: this is changed to increase test coverage of underlying `vectors_by_offsets`

lib/segment/src/segment/vectors.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use common::counter::hardware_counter::HardwareCounterCell;
+use common::iterator_ext::IteratorExt;
+
+use crate::common::operation_error::OperationResult;
+use crate::data_types::vectors::VectorInternal;
+use crate::segment::Segment;
+use crate::types::{PointIdType, VectorName};
+
+impl Segment {
```
generall, line 11: more vector-related function can be moved into this file later

lib/shard/src/operations/point_ops.rs
```diff
@@ -525,6 +526,54 @@ impl PointStructPersisted {
         }
         named_vectors
     }
+
+    pub fn is_equal_to(&self, segment_record: &SegmentRecord) -> bool {
```
generall, line 530: this function if used for deduplication during `sync` calls

lib/shard/src/retrieve/record_internal.rs
```diff
@@ -36,6 +47,23 @@ impl RecordInternal {
...
+    fn from(record: SegmentRecord) -> Self {
+        let SegmentRecord {
+            id,
+            payload,
+            vectors,
+        } = record;
+        Self {
+            id,
+            payload,
+            vector: vectors.map(VectorStructInternal::from),
```
generall, line 60: this is used to handle API expectation of points with no vector be displayed as `vector: {}`.

timvisee:
Nice. The change looks significant, assuming I parse the graphs correctly.

I did start an effort a few weeks back for concurrent reading+transferring. I think this approach is better. Though I might implement the same concurrentness on top of this in a separate PR. It'd eliminate waiting on roundtrip time.

lib/segment/src/segment/entry.rs
```diff
@@ -349,6 +363,87 @@ impl SegmentEntry for Segment {
...
 
+    fn retrieve(
+        &self,
+        point_ids: &[PointIdType],
+        with_payload: &WithPayload,
+        with_vector: &WithVector,
+        hw_counter: &HardwareCounterCell,
+        is_stopped: &AtomicBool,
+    ) -> OperationResult<Vec<SegmentRecord>> {
+        let mut records = AHashMap::with_capacity(point_ids.len());
```
timvisee, line 374: Why use a hashmap here?

Is it to work with the unordered nature of data?

lib/segment/src/vector_storage/dense/mmap_dense_vectors.rs
```diff
@@ -46,7 +46,7 @@ pub struct MmapDenseVectors<T: PrimitiveVectorElement> {
     _mmap_seq: Option<Arc<Mmap>>,
     /// Context for io_uring-base async IO
     #[cfg_attr(not(target_os = "linux"), allow(dead_code))]
-    uring_reader: Mutex<Option<UringReader<T>>>,
+    uring_reader: Option<Mutex<UringReader<T>>>,
```
timvisee, line 49: I like it is swapped around now

lib/segment/src/vector_storage/vector_storage_base.rs
```diff
@@ -984,6 +996,64 @@ impl VectorStorage for VectorStorageEnum {
         }
     }
 
+    fn read_vectors<P: AccessPattern>(
```
timvisee, line 999: What is the reason for using callbacks here? Any way to eliminate them?

lib/segment/src/segment/entry.rs
timvisee: I currently understand it like this:

the logic inside this retrieve function is ordered. But we fetch multiple points and process them through a callback. Adding them in a hash map is easier than other structures. The hash map breaks ordering.

lib/segment/src/vector_storage/vector_storage_base.rs
generall: I found code with callback easier than iterators. Iterators are fine with streaming approach, but as soon as we need batch - it practically means that we would need to collect on each step. Callbacks are also easier with lifetimes

timvisee:
```rust
for chunk in iter.chunks(10) { }
```

would give you easy to work with chunks from iterators.

> Callbacks are also easier with lifetimes

That is definitely true, especially with locks.

No strong objection, was just curious. Fine to merge like this.

lib/segment/src/segment/entry.rs
generall: Instead of fetching like this:

point 1 -> vector -> payload
point 2 -> vector -> payload

we now fetch

- all vectors
- all payloads

Fetch for vectors is not ordered, as it is async.
So we have to store incomming vectors somewhere, in this case - in hashmap

timvisee: I see the `vector_names > point_ids > retrieve` loop structure makes it hard to implement in a different way :+1:

PR #7929, opened 2026-01-18
implement a new configuration for read fan-out delay

generall:
There are cases, where we might want to smooth out tail latencies of read requests.
One of the typical exmples might be intecation happening in one replica but not another.

If we have an estimation of how long a certain request should take, we we can configure `read-fan-out-delay` to fire another request to another replica, to prevent one slow replica slowing down whole service.

Unlike `read-fan-out-factor` this configuration doesn't always require executing both requests, therefore it is more resource friendly at a cost of backed-in minimal delay.

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -7066,7 +7066,7 @@
             "nullable": true
           },
           "read_fan_out_delay_ms": {
-            "description": "Define number of milliseconds to wait before attempting to read from another replica. This setting can help to reduce latency spikes in case of occasional slow replicas. Default is 0, which means, that no additional request is sent.",
+            "description": "Define number of milliseconds to wait before attempting to read from another replica. This setting can help to reduce latency spikes in case of occasional slow replicas. Default is 0, which means delayed fan out request is disabled.",
             "type": "integer",
             "format": "uint64",
             "minimum": 0,
--- a/lib/collection/src/config.rs
+++ b/lib/collection/src/config.rs
@@ -120,7 +120,7 @@ pub struct CollectionParams {
     pub read_fan_out_factor: Option<u32>,
     /// Define number of milliseconds to wait before attempting to read from another replica.
     /// This setting can help to reduce latency spikes in case of occasional slow replicas.
-    /// Default is 0, which means, that no additional request is sent.
+    /// Default is 0, which means delayed fan out request is disabled.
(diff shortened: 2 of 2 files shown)


lib/collection/src/config.rs
```diff
@@ -118,6 +118,12 @@ pub struct CollectionParams {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[anonymize(false)]
     pub read_fan_out_factor: Option<u32>,
+    /// Define number of milliseconds to wait before attempting to read from another replica.
+    /// This setting can help to reduce latency spikes in case of occasional slow replicas.
+    /// Default is 0, which means, that no additional request is sent.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[anonymize(false)]
+    pub read_fan_out_delay_ms: Option<u64>,
```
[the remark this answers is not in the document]
generall: `None` is not something user can explicitly set for this API, so from user point of view 0 is a default value, unless it is explicitly changed to something else.

docs/redoc/master/openapi.json
```diff
@@ -7065,6 +7065,13 @@
             "minimum": 0,
             "nullable": true
           },
+          "read_fan_out_delay_ms": {
+            "description": "Define number of milliseconds to wait before attempting to read from another replica. This setting can help to reduce latency spikes in case of occasional slow replicas. Default is 0, which means, that no additional request is sent.",
+            "type": "integer",
+            "format": "uint64",
+            "minimum": 0,
+            "nullable": true
+          },
```
timvisee, line 7074: This means we cannot configure a delay of 0ms. Is that desired behavior?

How about using -1 as default to disable fan-out?

generall: 0ms fan-out doesn't make sense. It is essentially the same as just `read-fan-out-factor`

timvisee: The comment gave me the impression setting 0m would completely disable fan-out.

Looking at the code reveals that it doesn't. It only disables fan out based on delay, and the fan-out triggered on a busy shard still applies.

I suggest to update the comment to make it a bit more clear.

timvisee: I've updated it to the following

> Default is 0, which means delayed fan out request is disabled.

If you don't like it we can drop the commit :+1:

PR #7951, opened 2026-01-20
Update queue dont keep ops in ram

IvanPleshkov:
## Summary
This PR reduces the RAM usage of the pending operations in Update Worker.

## Description
Update worker has a channel with pending operations. Each pending has an `OperationData`, which contains the request data.

In case of bulk upload the channel may be huge and it consumes a lot of RAM.

This PR changes the logic of the Update Worker, and instead of storing the request in RAM, the Update Worker reads the request from the WAL instead.

## Performance
With buffering, there were no performance drops found. 

Using the vector db benchmark, the size of the update didnt reach the buffer size limit.
Using BFB, I got a full update queue using params `--threads=8 --num-vectors=1000000`. But still, the difference between runs is larger than the difference between branches.

Without buffering (by changing `DEFAULT_UPDATE_QUEUE_RAM_BUFFER` to zero), there is a performance drop. Because of deserialization, local measurements shown that deserialization might take 1-2ms of a large batched upsert. It's a fine price for unlimited update queue where ram usage has a higher priority than deserialization cost.

### DIFF
--- a/lib/collection/src/shards/local_shard/shard_ops.rs
+++ b/lib/collection/src/shards/local_shard/shard_ops.rs
@@ -72,7 +72,7 @@ impl ShardOperation for LocalShard {
         let operation_id = {
             let update_sender = self.update_sender.load();
             // Estimate pending operations count in the channel.
-            // `Sender::capacity` is returns available slots in the channel regarging tokio docs.
+            // `Sender::capacity` is returns available slots in the channel regarding tokio docs.
             // To calculate pending operations we need to subtract it from the max capacity,
             // which is the total capacity defined while creating the channel.
             let pending_operations_count = update_sender


lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -90,9 +94,14 @@ impl ShardOperation for LocalShard {
                 Err(err) => return Err(err.into()),
             };
 
+            // If there are too many pending operations, don't keep operation data in RAM.
+            // Instead, read operation data from the WAL when processing the operation.
+            let keep_operation_in_ram = pending_operations_count < DEFAULT_UPDATE_QUEUE_RAM_BUFFER;
+            let operation = keep_operation_in_ram.then_some(Box::new(operation.operation));
+
             channel_permit.send(UpdateSignal::Operation(OperationData {
```
timvisee, line 106: This struct is at least 152 bytes per operation, even for just a pointer to the WAL.

Just the `hw_measurements` field takes 120 bytes.

With that, we should probably box it too. An alternative would be an enum that satisfies all variants, but I cannot come up with a good one with these three fields.

I think it's worth a little bit of effort.

IvanPleshkov: Thanks for this measurement, I didnt attach it to the PR description.
Just a boxing wont help because we still need this field, box will move a memory without a profit. But I can try as a separate PR to refactor hw counter, where I can try to reduce `Arc` count here (by arc-ing the whole structure):
```
pub struct HwSharedDrain {
    pub(crate) cpu_counter: Arc<AtomicUsize>,
    pub(crate) payload_io_read_counter: Arc<AtomicUsize>,
    pub(crate) payload_io_write_counter: Arc<AtomicUsize>,
    pub(crate) payload_index_io_read_counter: Arc<AtomicUsize>,
    pub(crate) payload_index_io_write_counter: Arc<AtomicUsize>,
    pub(crate) vector_io_read_counter: Arc<AtomicUsize>,
    pub(crate) vector_io_write_counter: Arc<AtomicUsize>,
}
```
Is it fine for you?

timvisee: > Just a boxing wont help because we still need this field, box will move a memory without a profit. But I can try as a separate PR to refactor hw counter, where I can try to reduce Arc count here (by arc-ing the whole structure):

Good point. Even for values that we drop from memory will still have their HW aggregator attached. Thanks for the correction.

If this is only to report hardware statistics to wait=true responses. Maybe it'd be worth making wait=true and using the queue mutually exclusive. :thinking: 

> But I can try as a separate PR to refactor hw counter, where I can try to reduce Arc count here (by arc-ing the whole structure):

Not sure on that one. Let's focus on the current feature first.

lib/shard/src/wal.rs
```diff
@@ -79,6 +79,21 @@ impl<R: DeserializeOwned + Serialize> SerdeWal<R> {
         }
     }
 
+    pub fn read_single_record(&self, idx: u64) -> Result<Option<R>> {
```
agourlay, line 82: I'd love a basic unit test for this one just for the sake of coverage.

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -70,6 +71,9 @@ impl ShardOperation for LocalShard {
 
         let operation_id = {
             let update_sender = self.update_sender.load();
+            let pending_operations_count = update_sender
+                .max_capacity()
+                .saturating_sub(update_sender.capacity());
```
agourlay, line 80: Can you add a little comment that `capacity` is actually `remaining available slots`.
This is rather confusing from Tokio.

lib/collection/src/operations/shared_storage_config.rs
```diff
@@ -14,6 +14,10 @@ use crate::shards::transfer::ShardTransferMethod;
 const DEFAULT_SEARCH_TIMEOUT: Duration = Duration::from_secs(60);
 const DEFAULT_UPDATE_QUEUE_SIZE: usize = 100;
 const DEFAULT_UPDATE_QUEUE_SIZE_LISTENER: usize = 10_000;
+/// Maximum number of operations which are stored in RAM in update worker queue.
+/// If there are more pending operations, operation data
+/// will be read from WAL when processing the operation.
+pub const DEFAULT_UPDATE_QUEUE_RAM_BUFFER: usize = 50;
```
timvisee: Any strict reason for lowering this number from 100 or 200?

IvanPleshkov:
> I'm seeing a performance degradation which I don't expect.
> 
> Using `bfb -t8 -p8 -b1 --num-vectors=250000 --skip-wait-index --indexing-threshold 0`:
> 
> * on PR: 50 seconds upsert time (~4800/s)
> * on dev: 38 seconds upset time (~6500/s)

Checked locally. Maybe my hardware was not good enough to measure the drop, as I have around 29K/s on both branches.

I described in PR that deserialization has a huge price, and while reading from WAL cannot be avoided. Your measurement is more expected than mine.
As a solution, I propose increasing the buffer size up to the default update queue size. In this case, the behaviour won't be changed in comparison with dev. But we are ready to increase the update queue up to millions in the next PRs

timvisee:
> Checked locally. Maybe my hardware was not good enough to measure the drop, as I have around 29K/s on both branches.

Note that I tested on a debug build.

> I described in PR that deserialization has a huge price, and while reading from WAL cannot be avoided. Your measurement is more expected than mine.
> As a solution, I propose increasing the buffer size up to the default update queue size. In this case, the behaviour won't be changed in comparison with dev. But we are ready to increase the update queue up to millions in the next PRs

Are you suggesting that the 'lock, deserialize, unlock's on the WAL are causing this? Yeah I'll give setting the buffer size to unlimited a shot to see if it shows a different picture.

I'll also run two release builds to see how they behave.

lib/collection/src/shards/local_shard/shard_ops.rs
IvanPleshkov: Good point, I also was confused when I met `capacity`. Added a comment

lib/shard/src/wal.rs
IvanPleshkov: Added, the coverage is better now

lib/collection/src/operations/shared_storage_config.rs
IvanPleshkov: To make the changes measurable. And you did a measurement where you found an upload speed drop.
I replied to your message with bfb and reverted this constant back to `200`. Now this PR does not affect dev perf while the update queue is `100`.

IvanPleshkov:
> > Checked locally. Maybe my hardware was not good enough to measure the drop, as I have around 29K/s on both branches.
> 
> Note that I tested on a debug build.
> 
> > I described in PR that deserialization has a huge price, and while reading from WAL cannot be avoided. Your measurement is more expected than mine.
> > As a solution, I propose increasing the buffer size up to the default update queue size. In this case, the behaviour won't be changed in comparison with dev. But we are ready to increase the update queue up to millions in the next PRs
> 
> Are you suggesting that the 'lock, deserialize, unlock's on the WAL are causing this? Yeah I'll give setting the buffer size to unlimited a shot to see if it shows a different picture.
> 
> I'll also run two release builds to see how they behave.

To make sure that the reason is a deserialization, I measured possible drops (WAL blocking, IO, deserialization) using regular times and got milliseconds for a deserialization even in a release build. All other possible drop points were much smaller

timvisee:
> > Checked locally. Maybe my hardware was not good enough to measure the drop, as I have around 29K/s on both branches.
> 
> Note that I tested on a debug build.
> 
> > I described in PR that deserialization has a huge price, and while reading from WAL cannot be avoided. Your measurement is more expected than mine.
> > As a solution, I propose increasing the buffer size up to the default update queue size. In this case, the behaviour won't be changed in comparison with dev. But we are ready to increase the update queue up to millions in the next PRs
> 
> Are you suggesting that the 'lock, deserialize, unlock's on the WAL are causing this? Yeah I'll give setting the buffer size to unlimited a shot to see if it shows a different picture.
> 
> I'll also run two release builds to see how they behave.

@IvanPleshkov 

Using debug build and `bfb -t8 -p8 -b1 --num-vectors=250000 --skip-wait-index --indexing-threshold 0`:

- on PR(limit: 50): 50 seconds upsert time (~4800/s)
- on PR(limit: 200): 38 seconds upsert time (~6500/s)
- on PR(limit: 1m): 38 seconds upsert time (~6500/s)
- on dev: 38 seconds upset time (~6500/s)

Using perf build and `bfb -t8 -p8 -b1 --num-vectors=1000000 --skip-wait-index --indexing-threshold 0`:

- on pr(limit: 50): 33 seconds upsert time
- on pr(limit: 200): 33 seconds upsert time
- on pr(limit: 1m): 33 seconds upsert time
- on dev: 33 seconds upsert time

It shows the problem goes away in release builds, at least for this work load.

Any objection for reverting the queue size limit from 50 to 200, to at least make the debug build case better?

IvanPleshkov:
> > > Checked locally. Maybe my hardware was not good enough to measure the drop, as I have around 29K/s on both branches.
> > 
> > 
> > Note that I tested on a debug build.
> > > I described in PR that deserialization has a huge price, and while reading from WAL cannot be avoided. Your measurement is more expected than mine.
> > > As a solution, I propose increasing the buffer size up to the default update queue size. In this case, the behaviour won't be changed in comparison with dev. But we are ready to increase the update queue up to millions in the next PRs
> > 
> > 
> > Are you suggesting that the 'lock, deserialize, unlock's on the WAL are causing this? Yeah I'll give setting the buffer size to unlimited a shot to see if it shows a different picture.
> > I'll also run two release builds to see how they behave.
> 
> @IvanPleshkov
> 
> Using debug build and `bfb -t8 -p8 -b1 --num-vectors=250000 --skip-wait-index --indexing-threshold 0`:
> 
> * on PR(limit: 50): 50 seconds upsert time (~4800/s)
> * on PR(limit: 200): 38 seconds upsert time (~6500/s)
> * on PR(limit: 1m): 38 seconds upsert time (~6500/s)
> * on dev: 38 seconds upset time (~6500/s)
> 
> Using perf build and `bfb -t8 -p8 -b1 --num-vectors=1000000 --skip-wait-index --indexing-threshold 0`:
> 
> * on pr(limit: 50): 33 seconds upsert time
> * on pr(limit: 200): 33 seconds upsert time
> * on pr(limit: 1m): 33 seconds upsert time
> * on dev: 33 seconds upsert time
> 
> It shows the problem goes away in release builds, at least for this work load.
> 
> Any objection for reverting the queue size limit from 50 to 200, to at least make the debug build case better?

No objections, I have already pushed the buffer size limit changes (to 500, after 100 it's no matter how large it is)

PR #7952, opened 2026-01-20
Always Copy-on-Write when updating payload in immutable segments

ffuugoo:
This makes partial snapshots more efficient. Without this change, if we modify a single payload (key) in immutable segment, the whole payload storage (for this segment) have to be included into the snapshot.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/vector_storage/mod.rs
+++ b/lib/segment/src/vector_storage/mod.rs
@@ -1,26 +1,25 @@
 #[cfg(target_os = "linux")]
-pub mod async_raw_scorer;
-mod chunked_mmap_vectors;
-pub mod chunked_vectors;
-pub mod quantized;
-pub mod raw_scorer;
-mod vector_storage_base;
-
-#[cfg(test)]
-mod tests;
-
-#[cfg(target_os = "linux")]
 mod async_io;
 mod async_io_mock;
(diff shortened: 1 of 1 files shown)


agourlay:
What is the rational for this change exactly?

ffuugoo:
> What is the rational for this change exactly?

Partial snapshots don't work well enough if we modify payload in-place in immutable segments.

timvisee:
> What is the rational for this change exactly?

Make immutable segments properly immutable.

It has two benefits:
1. simpler R/W synchronization (partial snapshots) implementation, we now don't have to track changes in immutable segments
2. make immutable segments lock free (hopefully aids performance)

There is one exception that remains: deleted flags. But it's likely we can implement this in a lock free fashion with atomics.

Our guess is that impact on users is minimal to nothing. For existing users this would only have been beneficial if they'd update payload that also is not indexed. We believe that doesn't really happen often in practice.

generall:
@ffuugoo please resolve conflict

ffuugoo:
> @ffuugoo please resolve conflict

Who did this!? 😡

ffuugoo:
> @ffuugoo please resolve conflict

Rebased

timvisee (approved):

PR #7946, opened 2026-01-20
Update `GET collections/{}/optimizations`

xzfc:
This PR rehauls `/collections/{}/optimizations` endpoint.

- Now this endpoint contains the following fields:
  - `summary` (replaces old summary in `pending`)
  - `running` (was `ongoing`) - list
  - `queued` (new) - list, predicted by `OptimizerPlanner`
  - `completed` - list
  - `idle_segments` (new) - list of good segments that don't require optimizations
- Previously, `running`/`completed` lists were containing naked `ProgressTree`s.
  - In this PR, they are wrapped with additional meta-information.
  - `uuid`s are included to let the UI refresh currently selected progress tree.
  - The difference, illustrated:
    ```diff
     "running": [
    +  {
    +    "uuid": "b17f761d-be66-4cab-b931-3f631c746066",
    +    "segments": [
    +      {"uuid": "ea103d7a-f379-4b37-814c-9f8b6400c321", "points_count": 12345}
    +    ],
    +    "optimizer": "indexing",
    +    "status": "optimizing",
    +    "progress":
           { /* progress tree */ },
    +  }
     ]
    ```
- New query parameter `?with` to control whether to include new optional fields.
  - To include all fields, use `?with=queued,completed,idlle_segments`.
  - Why: these lists might be very chunky.
  - Why `running` is not optional: if we need just a single number of current optimizations, then we can add this number it to `GET /collections/{}` endpoint.
  - The old parameter `?completed=true` is replaced with `?with=completed`.
- Use-case for `OptimizationSegmentInfo` (`idle_segments[]`, `queued[].segments[]`, `running[].segments[]`):
  The UI might visualize the collection as a bunch of rectangles; with color/size based on the status/points count. (perhaps it might look like [the old defrag](https://defrag98.com/)) UUIDs can help keep ordering/transitions.

Example output:
```javascript
{
  "summary": {
    "queued_optimizations": 8,
    "queued_segments": 8,
    "queued_points": 50400,
    "idle_segments": 8
  },

  "running": [
    {
      "uuid": "9254c533-f78e-4b14-9935-593ef2bc8519",
      "optimizer": "indexing",
      "status": "optimizing",
      "segments": [
        {"uuid": "bfcd34c7-eebd-47b1-a30e-781ac2a28dc7", "points_count": 6300}
      ],
      "progress": { /* progress */ }
    },
    /* … */
  ],

  "queued": [
    {
      "optimizer": "indexing",
      "segments": [
[...]

PR #7958, opened 2026-01-21
Enforce segment UUIDs

xzfc:
Continuation of #7735

The PR #7735 added `Segment::uuid` field of type `String`.
This PR enforces `Segment::uuid` to be actual UUIDs. Basically:
```diff
 pub struct Segment {
-    pub uuid: String,
+    pub uuid: Uuid,
```

This is a preparation work for upcoming changes in the `…/optimizations` endpoint.

# Renaming non-UUID segments

In this PR, during load, segments with non-UUID names are assigned a random UUID and their directories are renamed.
The renaming is performed by the new function `normalize_segment_dir` which should be called together with `load_segment`.

Usage of `load_segment` before this PR
```rust
// (simplified code to illustrate the idea)

let segment = match load_segment(path, …)? {
    // Happy case: segment is loaded.
    LoadSegmentOutcome::Loaded(segment) => segment,

    // `load_segment` said that the segment is incomplete.
    LoadSegmentOutcome::Skipped => {
        // It is the responsibility of the caller to delete it.
        safe_delete_with_suffix(path)?;
        // It's not an error, so continue.
        continue;
    }
};
```

Usage of `normalize_segment_dir`/`load_segment` in this PR:
```rust
// (simplified code to illustrate the idea)

let segment = match normalize_segment_dir(path)? {
    // Happy case. It is possible that `normalize_segment_dir` renamed
    // the segment: the `updated_path` would contain the new name.
    // Otherwise, the `updated_path` is the same as `path`.
    Some((updated_path, uuid)) => load_segment(updated_path, uuid)?,

    // `normalize_segment_dir` deleted the incomplete segment.
    None => continue,
};
```

In tests or in the segment exporter, we don't need to call `normalize_segment_dir`.

# `Tracker::uuid`

The third commit in this PR adds the UUID of the new segment to the `optimizers::Tracker` and to the telemetry output.

```jsonc
// curl -s '127.1:6333/telemetry?details_level=9' | jq '.result.collections.collections[].shards[].local.optimizations.log[0]'
{
  "name": "indexing",
  "uuid": "bd23fe1d-c101-4360-9354-4b319d8d42bf", // ← Added in this PR
  "segment_ids": [ 5, 3 ],
  "segment_uuids": [
    "1f1dd8a6-d87b-416d-9171-e6d6e74869fb", // ← Added in #7735
    "268006f6-0400-40ab-b6d0-46ba5539764d"  // ←
  ],
  "status": "optimizing",
  "start_at": "2026-01-21T03:04:30.903823168Z",
  "end_at": null
}
```

Why: having an UUID in the tracker would be useful for the `…/optimizations` endpoint. (#7946)
The use-case is to let the UI refresh the currently selected progress tree.
(i.e., UI should re-fetch `/optimizations` and find the matching entry by its UUID)

### DIFF
--- a/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
+++ b/lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
@@ -409,21 +409,12 @@ pub trait SegmentOptimizer {
     /// Warn: this function might be _VERY_ CPU intensive,
     /// so it is necessary to avoid any locks inside this part of the code
     ///
-    /// # Arguments
-    ///
-    /// * `optimizing_segments` - Segments to optimize
-    /// * `proxy_deleted_points` - Holds a set of points, deleted while optimization was running
-    /// * `proxy_changed_indexes` - Holds a set of indexes changes, created or deleted while optimization was running
-    /// * `stopped` - flag to check if optimization was cancelled by external thread
-    ///
-    /// # Result
-    ///
-    /// Constructs optimized segment
+    /// Returns the newly constructed optimized segment.
     #[allow(clippy::too_many_arguments)]
     fn build_new_segment(
         &self,
-        uuid: Uuid,
-        optimizing_segments: &[LockedSegment],
+        input_segments: &[LockedSegment], // Segments to optimize/merge into one
+        output_segment_uuid: Uuid,        // The UUID of the resulting optimized segment
         proxies: &[LockedSegment],
         permit: ResourcePermit, // IO resources for copying data
         resource_budget: ResourceBudget,
@@ -431,7 +422,7 @@ pub trait SegmentOptimizer {
         hw_counter: &HardwareCounterCell,
         progress: ProgressTracker,
     ) -> CollectionResult<Segment> {
-        let mut segment_builder = self.optimized_segment_builder(optimizing_segments)?;
+        let mut segment_builder = self.optimized_segment_builder(input_segments)?;
 
         check_process_stopped(stopped)?;
 
@@ -439,7 +430,7 @@ pub trait SegmentOptimizer {
         let progress_populate_storages = progress.subtask("populate_vector_storages");
         let progress_wait_permit = progress.subtask("wait_cpu_permit");
 
-        let segments: Vec<_> = optimizing_segments
+        let segments: Vec<_> = input_segments
             .iter()
             .map(|i| match i {
                 LockedSegment::Original(o) => o.clone(),
@@ -551,7 +542,7 @@ pub trait SegmentOptimizer {
         let mut rng = rand::rng();
         let mut optimized_segment = segment_builder.build(
             self.segments_path(),
-            uuid,
+            output_segment_uuid,
             indexing_permit,
             stopped,
             &mut rng,
@@ -607,8 +598,8 @@ pub trait SegmentOptimizer {
         let budget = ResourceBudget::new(permit_cpu_count, permit_cpu_count);
         self.optimize(
             segments,
-            Uuid::new_v4(),
             ids,
+            Uuid::new_v4(),
             budget.try_acquire(0, permit_cpu_count).unwrap(),
             budget,
             &AtomicBool::new(false),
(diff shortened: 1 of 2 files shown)


lib/segment/src/segment/snapshot.rs
```diff
@@ -69,7 +69,7 @@ impl SnapshotEntry for Segment {
                         ))
                     })?;
 
-                let tar = tar.descend(Path::new(&segment_id))?;
+                let tar = tar.descend(Path::new(&segment_id.to_string()))?;
```
timvisee, line 72: To check: are we sure this formats the UUID in the same was as we format it on disk

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -704,42 +691,71 @@ fn create_segment_id_tracker(
...
+        .and_then(|fname| fname.to_str())
+        .ok_or_else(|| {
+            OperationError::service_error(format!(
+                "Failed to get segment folder name: {}",
+                path.display()
+            ))
+        })?;
+    match Uuid::try_parse(file_name) {
+        Ok(uuid) => Ok(Some((path.to_path_buf(), uuid))),
+        Err(_) => {
```
timvisee, line 738: Lets add a `log::warn!()` here about the rename.

We don't expect non-UUID names under normal circumstances. And in my opinion this rename is quite significant.

lib/segment/src/segment/snapshot.rs
timvisee: It's consistent.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -614,6 +640,7 @@ pub trait SegmentOptimizer {
     fn optimize(
         &self,
         segments: LockedSegmentHolder,
+        uuid: Uuid,
```
KShivendu: nitpick: mention purpose of uuid in the function docs.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -422,6 +422,7 @@ pub trait SegmentOptimizer {
     #[allow(clippy::too_many_arguments)]
     fn build_new_segment(
         &self,
+        uuid: Uuid,
```
KShivendu: nitpick: mention purpose of uuid in function docs

lib/shard/src/segment_holder/mod.rs
```diff
@@ -118,7 +119,7 @@ impl SegmentHolder {
     }
 
     /// Get filesystem aligned UUIDs instead of internal segment IDs for reporting in telemetry
-    pub fn segment_uuids(&self, segment_ids: &[SegmentId]) -> Vec<String> {
+    pub fn segment_uuids(&self, segment_ids: &[SegmentId]) -> Vec<Uuid> {
```
KShivendu, line 122: We can change the log warning to `Failed to get segment with internal id {id}` since UUID is now guaranteed.

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -704,42 +691,76 @@ fn create_segment_id_tracker(
...
+            let segment_uuid = Uuid::new_v4();
+            let new_path = path.with_file_name(segment_uuid.to_string());
+            log::warn!(
+                "Segmant name is not a valid UUID: {}. Renaming to {segment_uuid}",
```
timvisee: Typo spotted by coderabbit

```suggestion
                "Segment name is not a valid UUID: {}. Renaming to {segment_uuid}",
```

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
xzfc: Done in d5116c990b3613c52450333ebdccfc68de0e4784.

lib/shard/src/segment_holder/mod.rs
xzfc: Done in d5116c990b3613c52450333ebdccfc68de0e4784.

PR #7959, opened 2026-01-21
Allow partial snapshots even when no local shard exists

KShivendu:
We will be using this for for [this](https://www.notion.so/qdrant/Changing-RF-in-read-cluster-2e3674779d3380a48c96d5246fbe0dba)

```http
# Request:
POST /collections/benchmark/shards/0/snapshot/partial/recover_from?wait=false
{
  "peer_url": "http://127.0.0.1:6333"
}

# Response before:
{
  "error": "Bad request: local shard benchmark:0 does not exist on peer 205"
}

# Response after:
{
  "status": "accepted"
}
```

### DIFF
--- a/src/actix/api/snapshot_api.rs
+++ b/src/actix/api/snapshot_api.rs
@@ -742,7 +742,7 @@ async fn recover_partial_snapshot_from(
             // Empty snapshot manifest allows us to use partial snapshots even if local shard doesn't exist
             let snapshot_manifest = match collection.get_partial_snapshot_manifest(shard_id).await {
                 Ok(manifest) => manifest,
-                Err(CollectionError::NotFound { .. }) => SnapshotManifest::empty(),
+                Err(CollectionError::NotFound { .. }) => SnapshotManifest::default(),
                 Err(err) => return Err(StorageError::from(err))
             };
 


src/actix/api/snapshot_api.rs
```diff
@@ -738,7 +738,8 @@ async fn recover_partial_snapshot_from(
                 "{peer_url}/collections/{encoded_collection_name}/shards/{shard_id}/snapshot/partial/create"
             );
 
-            let snapshot_manifest = collection.get_partial_snapshot_manifest(shard_id).await?;
+            // Empty snapshot manifest allows us to use partial snapshots even if local shard doesn't exist
+            let snapshot_manifest = collection.get_partial_snapshot_manifest(shard_id).await.unwrap_or_else(|_| SnapshotManifest::empty());
```
ffuugoo: There might be different error types, it's incorrect to ignore _all_ errors and unwrap into empty manifest.

src/common/snapshots.rs
```diff
@@ -65,7 +65,9 @@ pub async fn stream_shard_snapshot(
         .running_snapshots
         .measure_scope();
 
-    if let Some(old_manifest) = &manifest {
+    if let Some(old_manifest) = &manifest
+        && !old_manifest.is_empty()
```
ffuugoo: Can you clarify why this is needed? What corner case does this handle?

KShivendu: Just a small optimization to not compare manifest if we know it's empty (which happens when read replica doesn't exist on the read cluster). This partial snapshot will create it

src/actix/api/snapshot_api.rs
KShivendu: Hmm okay. How about we return `CollectionError::NotFound` [here](https://github.com/qdrant/qdrant/blob/master/lib/collection/src/shards/replica_set/snapshots.rs#L460-L465) and return empty manifest only then?

I don't want to always return empty manifest from `get_partial_snapshot_manifest()` because it's used in other places where it's more suitable to throw error. 

Also comparing strings (`err.contains("does not exist on peer")` for error doesn't make sense

src/common/snapshots.rs
generall: I would keep old logic if it is only about optimization

src/actix/api/snapshot_api.rs
KShivendu: Hmm technically there are only two errors possible from `get_partial_snapshot_manifest()` - `CollectionError::NotFound` and `CollectionError::BadInput`. 

But fine. I'll just use `CollectionError::NotFound` in both, handle it, and raise other errors.

src/actix/api/snapshot_api.rs
```diff
@@ -738,7 +739,12 @@ async fn recover_partial_snapshot_from(
                 "{peer_url}/collections/{encoded_collection_name}/shards/{shard_id}/snapshot/partial/create"
             );
 
-            let snapshot_manifest = collection.get_partial_snapshot_manifest(shard_id).await?;
+            // Empty snapshot manifest allows us to use partial snapshots even if local shard doesn't exist
+            let snapshot_manifest = match collection.get_partial_snapshot_manifest(shard_id).await {
+                Ok(manifest) => manifest,
+                Err(CollectionError::NotFound { what: _ }) => SnapshotManifest::empty(),
+                Err(e) => Err(e)?
```
ffuugoo:
```suggestion
                Err(CollectionError::NotFound { .. }) => SnapshotManifest::empty(),
                Err(err) => return Err(err),
```

KShivendu: > return Err(err),

Actually we need to convert `CollectionError` into `StorageError`. So I added `?`

ffuugoo: Then `return Err(CollectionError::from(err))`. `Err(err)?` works, but it feels _wrong_. I don't think I ever seen it used like this in practice, everyone do explicit `return`.

KShivendu: Updated the PR. Thanks!

lib/collection/src/shards/replica_set/snapshots.rs
timvisee: `None` branches can be merged.



We _could_ move the Some branches into `local.take().map(...)` to prevent `None => None`.
```suggestion
```

lib/shard/src/snapshots/snapshot_manifest.rs
```diff
@@ -17,6 +17,13 @@ pub struct SnapshotManifest {
 }
 
 impl SnapshotManifest {
+    /// Create an empty snapshot manifest
+    pub fn empty() -> Self {
+        Self {
+            segments: HashMap::new(),
+        }
+    }
```
timvisee: This is equal to `SnapshotManifest::default()`. I suggest to only keep one of them, where `default()` would be more conventional.

lib/collection/src/shards/replica_set/snapshots.rs
KShivendu: Nice catch on the 1st suggestion :D 

For the 2nd, `local.take().map(...)` This will get ugly because we have `.await` and `return Err(...)`

timvisee (approved):

lib/collection/src/shards/replica_set/snapshots.rs
timvisee: Makes sense, I missed the await.

PR #7963, opened 2026-01-21
introduce update_mode parameter for upsert operation to control if we want to insert, update, or upsert

generall:
… want to insert, update, or upsert

### DIFF
--- a/tests/consensus_tests/test_resharding.py
+++ b/tests/consensus_tests/test_resharding.py
@@ -324,6 +324,8 @@ def bootstrap_resharding(
     # Bootstrap cluster
     peer_uris, peer_ids = bootstrap_cluster(tmp_path, collection, **kwargs)
 
+    wait_for_all_peers_versions(peer_uris)
+
     # Select target peer
     peer_id = None
 
--- a/tests/consensus_tests/utils.py
+++ b/tests/consensus_tests/utils.py
@@ -384,6 +384,19 @@ def all_nodes_cluster_info_consistent(peer_api_uris: [str], expected_leader: str
             return False
     return True
 
+def peers_have_version(peer_api_uris: [str]) -> bool:
+    peer_api_uri = peer_api_uris[0]
+    try:
+        r = requests.get(f"{peer_api_uri}/cluster/telemetry")
+        assert_http_ok(r)
+        peers = r.json()["result"]["cluster"]["peers"]
+        for peer_id, peer_status in peers.items():
+            if peer_status["details"].get("version") is None:
+                return False
+    except requests.exceptions.ConnectionError:
+        print(f"Could not contact peer {peer_api_uri} to fetch versions")
+        return False
+    return True
 
 def all_nodes_have_same_commit(peer_api_uris: [str]) -> bool:
     commits = []
@@ -606,6 +619,12 @@ def wait_for_uniform_cluster_status(peer_api_uris: [str], expected_leader: str, 
         print_clusters_info(peer_api_uris)
         raise e
 
+def wait_for_all_peers_versions(peer_api_uris: [str]):
+    try:
+        wait_for(peers_have_version, peer_api_uris)
+    except Exception as e:
+        print_clusters_info(peer_api_uris)
+        raise e
 
 def wait_for_same_commit(peer_api_uris: [str]):
     try:


lib/shard/src/operations/point_ops.rs
```diff
@@ -224,149 +237,6 @@ impl PointInsertOperationsInternal {
...
-        self,
-        update_filter: Option<Filter>,
-    ) -> Vec<CollectionUpdateOperations> {
-        let mut operations = Vec::new();
-
-        match self {
-            Self::PointsBatch(batch) => {
-                let mut update_vectors = UpdateVectorsOp {
-                    points: Vec::new(),
-                    update_filter: update_filter.clone(),
```
generall, line 238: seems like this function is no longer needed for good. cc @ffuugoo

lib/shard/src/segment_holder/mod.rs
```diff
@@ -765,6 +765,43 @@ impl SegmentHolder {
         Ok(read_points)
     }
 
+    /// Out of a list of point IDs, select only those that exist in at least one segment.
+    ///
+    /// Optimized for many segments and long lists of IDs:
+    /// - Removes found IDs from consideration for subsequent segments
+    /// - Early terminates when all IDs are found
+    /// - Pre-allocates result set
+    pub fn select_existing_points(&self, ids: Vec<PointIdType>) -> AHashSet<PointIdType> {
```
generall, line 774: this is slightly optimized function, assumed to be working better if we have 10+ segments

lib/shard/src/update.rs
```diff
@@ -284,14 +284,36 @@ pub fn conditional_upsert(
...
+            // Only update existing points that match the condition, don't insert new points
+            let points_to_exclude =
+                select_excluded_by_filter_ids(segments, point_ids.clone(), condition, hw_counter)?;
+            let existing_points = segments.select_existing_points(point_ids);
+            // Keep only points that exist AND are not excluded by the condition
+            points_op.retain_point_ids(|idx| {
+                existing_points.contains(idx) && !points_to_exclude.contains(idx)
+            });
+        }
+    }
```
generall, line 315: main change is here

docs/redoc/master/openapi.json
```diff
@@ -10858,6 +10858,17 @@
                 "nullable": true
               }
             ]
+          },
+          "update_mode": {
```
timvisee, line 10862: I cannot decide whether 'insert mode', 'update mode' or 'upsert mode' is better.

How about we just call it 'mode'?

lib/collection/src/shards/conversions.rs
```diff
@@ -84,6 +84,7 @@ pub fn internal_upsert_points(
             shard_key_selector: None,
             update_filter: None,
             timeout: wait_timeout,
+            update_mode: None, // Default mode (Upsert)
```
timvisee, line 87: Shouldn't we use the user setting here?

lib/shard/src/update.rs
```diff
@@ -284,14 +284,36 @@ pub fn conditional_upsert(
...
+        UpdateMode::InsertOnly => {
+            // Only insert new points, skip all existing points entirely
+            let existing_points = segments.select_existing_points(point_ids);
+            points_op.retain_point_ids(|idx| !existing_points.contains(idx));
+        }
+        UpdateMode::UpdateOnly => {
+            // Only update existing points that match the condition, don't insert new points
+            let points_to_exclude =
+                select_excluded_by_filter_ids(segments, point_ids.clone(), condition, hw_counter)?;
+            let existing_points = segments.select_existing_points(point_ids);
```
timvisee, line 309: Potential optimization, how about we:
- mutate `point_ids` here after the first call
- it may reduce point count for the next call, and may potentially even shortcut earlier

lib/shard/src/update.rs
timvisee: I like that this logic is now centralized here :ok_hand:

docs/redoc/master/openapi.json
ffuugoo: IMO, `update_mode` is fine.

lib/collection/src/shards/conversions.rs
generall: This oeprations converts from unconditional upsert. Conditional upsert is done in `internal_conditional_upsert_points`.

 I think overall we can start simplifying things in further releases, as it becomes redundant and mostly done for compatibility purposes.

lib/shard/src/update.rs
generall: can it be postponed for now? I am lookign into it, it is not so trivial

timvisee: Of course. I doubt it'll make a huge difference.

docs/redoc/master/openapi.json
timvisee: I understand it goes well with `update_filter` too. Let's keep it.

lib/shard/src/operations/point_ops.rs
ffuugoo: Is it possible for a _partially upgraded_ cluster to be in process of resharding? E.g., some nodes already use `update_mode`, but some nodes don't yet "understand" it?

timvisee: Good question.

Resharding should not progress (or be triggered) if nodes are down. I'll verify this on my end.

Also, resharding is currently aborted when a node restarts.

timvisee: Yes: https://github.com/qdrant/cluster-manager/blob/4e6c3c1ac499375ff9e797ff14b192fdbc98468c/src/cluster_manager/manager.rs#L165

ffuugoo: Yeah, but
1) is it possible to trigger node upgrade while resharding is in progress?
2) can we _guarantee_ that we will notice a quick node restart?
3) do we care?

ffuugoo: By itself, resharding won't be cancelled if you restart a node (e.g., if the node is not currently doing resharding transfer or smth). Only if cluster-manager observed node being unavailable.

Also you don't even need restarting nodes _during_ resharding. It's enough to upgrade one node right _before_ starting a resharding. And then we would have nodes with (silently) incompatible protocols.

I'm just not sure if there's already logic to prevent this in CM. If there is, then maybe we don't care.

docs/redoc/master/openapi.json
timvisee, line 10862: In validation: shall we make `update_mode: insert_only` and `update_filter` mutually exclusive?

The update filter is completely ignored in that case. It might help prevent accidental misuse.

ffuugoo:
Note that the [comment](https://github.com/qdrant/qdrant/pull/7963#discussion_r2720771286) about multi-versioned cluster still stands.

lib/shard/src/operations/point_ops.rs
generall: If I understand correctly, worst case it will casue transfer error and cancellation of resharding?
In this case we don't care.

docs/redoc/master/openapi.json
generall: I was thinking about it, and decided not to.

lib/shard/src/operations/point_ops.rs
ffuugoo: I think if we send operations with `UpdateOnly` to old node it will just ignore it? 🤔

generall: you are right, added version gate

PR #7969, opened 2026-01-22
fix: replica state peer id remap

1995chen:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

Fixes replica state remap on shard load: when the stored this_peer_id differs from the current node ID, the code now removes the old peer entry, re-attaches that local state under the new this_peer_id, and then updates rs.this_peer_id.
Prevents leaving the replica state under the old peer ID (previously shadowed the new ID), avoiding state loss/mismatch after node ID changes or data moves.

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -264,8 +264,8 @@ impl ShardReplicaSet {
         if replica_state.read().this_peer_id != this_peer_id {
             replica_state
                 .write(|rs| {
-                    let this_peer_id = rs.this_peer_id;
-                    let local_state = rs.remove_peer_state(this_peer_id);
+                    let old_peer_id = rs.this_peer_id;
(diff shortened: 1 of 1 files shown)


1995chen:
@timvisee Hello, I looked at the code implementation here and it doesn’t seem to be following the designer’s intent.

1995chen:
Fixes replica state remap on shard load: when the stored this_peer_id differs from the current node ID, the code now removes the old peer entry, re-attaches that local state under the new this_peer_id, and then updates rs.this_peer_id.
Prevents leaving the replica state under the old peer ID (previously shadowed the new ID), avoiding state loss/mismatch after node ID changes or data moves.

timvisee (commented):
I do believe the change makes sense, thanks!

But, do you have a clear reproducible example that shows this fixes the problem?

generall:
I thought I fixed it, but either it is something different or I forgot to push.

Thanks @1995chen !

PR #7971, opened 2026-01-22
in ram single mmap file

generall:
- **WIP: introduce new vector store type**
- **handling of InRamMmap**
- **fmt**
- **feature-flag**
- **fmt**

### DIFF
--- a/lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
+++ b/lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
@@ -18,8 +18,9 @@ use crate::data_types::vectors::{TypedMultiDenseVectorRef, VectorElementType, Ve
 use crate::types::{Distance, MultiVectorConfig, VectorStorageDatatype};
 use crate::vector_storage::chunked_mmap_vectors::ChunkedMmapVectors;
 use crate::vector_storage::chunked_vector_storage::{ChunkedVectorStorage, VectorOffsetType};
-use crate::vector_storage::dense::memmap_dense_vector_storage::{
-    open_memmap_vector_storage, open_memmap_vector_storage_byte, open_memmap_vector_storage_half,
+use crate::vector_storage::dense::appendable_dense_vector_storage::{
+    open_appendable_memmap_vector_storage_byte, open_appendable_memmap_vector_storage_full,
+    open_appendable_memmap_vector_storage_half,
 };
 use crate::vector_storage::{
     AccessPattern, MultiVectorStorage, Random, Sequential, VectorStorage, VectorStorageEnum,
@@ -308,15 +309,27 @@ pub fn open_appendable_memmap_vector_storage(
     populate: bool,
 ) -> OperationResult<VectorStorageEnum> {
     match storage_element_type {
-        VectorStorageDatatype::Float32 => {
-            open_memmap_vector_storage(vector_storage_path, size, distance, madvise, populate)
-        }
-        VectorStorageDatatype::Uint8 => {
-            open_memmap_vector_storage_byte(vector_storage_path, size, distance, madvise, populate)
-        }
-        VectorStorageDatatype::Float16 => {
-            open_memmap_vector_storage_half(vector_storage_path, size, distance, madvise, populate)
-        }
+        VectorStorageDatatype::Float32 => open_appendable_memmap_vector_storage_full(
+            vector_storage_path,
+            size,
+            distance,
+            madvise,
+            populate,
+        ),
+        VectorStorageDatatype::Uint8 => open_appendable_memmap_vector_storage_byte(
+            vector_storage_path,
+            size,
+            distance,
+            madvise,
+            populate,
+        ),
+        VectorStorageDatatype::Float16 => open_appendable_memmap_vector_storage_half(
+            vector_storage_path,
+            size,
+            distance,
+            madvise,
+            populate,
+        ),
     }
 }
 


lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -320,8 +320,16 @@ pub trait SegmentOptimizer {
...
+                    Some(false) => {
+                        if common::flags::feature_flags().single_file_mmap_vector_storage  {
+                            config.storage_type = VectorStorageType::InRamMmap;
+                        }
+                    } // on_disk=false wins, do nothing
+                    None => if threshold_is_on_disk { config.storage_type = VectorStorageType::Mmap } else {
+                        if common::flags::feature_flags().single_file_mmap_vector_storage  {
+                            config.storage_type = VectorStorageType::InRamMmap;
+                        }
+                    }, // Mmap threshold wins
```
ffuugoo:
```suggestion
                    None => if threshold_is_on_disk { 
                        config.storage_type = VectorStorageType::Mmap
                    } else if common::flags::feature_flags().single_file_mmap_vector_storage  {
                        config.storage_type = VectorStorageType::InRamMmap;
                    }, // Mmap threshold wins
```

lib/segment/src/vector_storage/dense/appendable_dense_vector_storage.rs
```diff
@@ -213,9 +213,11 @@ pub fn open_appendable_memmap_vector_storage(
     path: &Path,
     dim: usize,
     distance: Distance,
+    populate: bool,
 ) -> OperationResult<VectorStorageEnum> {
-    let storage =
-        open_appendable_memmap_vector_storage_impl::<VectorElementType>(path, dim, distance)?;
+    let storage = open_appendable_memmap_vector_storage_impl::<VectorElementType>(
+        path, dim, distance, populate,
+    )?;
```
ffuugoo, line 216: Explicit `VectorElementType` is not required here (it's not used in most other similar functions).

```suggestion
    let storage = open_appendable_memmap_vector_storage_impl(
        path, dim, distance, populate,
    )?;
```

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -146,61 +186,40 @@ pub(crate) fn open_vector_storage(
...
-                        vector_config.distance,
-                    ),
-                }
-            }
+            let populate = false;
+            open_mmap_vector_storage(vector_storage_path, vector_config, populate)
         }
         // Chunked mmap on disk, appendable
         VectorStorageType::ChunkedMmap => {
+            let populate = false;
```
ffuugoo: Can we, please, make this `const POPULATE: bool = false`?

I _always_ get confused by these `let something = constant_value`, because `do_action(something)` looks as if `something` is propagated from somewhere. `do_action(SOMETHING)` would make it much cleaner.

ffuugoo (changes requested):
Same everywhere, pls.

lib/common/common/src/flags.rs
```diff
@@ -80,6 +85,7 @@ impl Default for FeatureFlags {
             migrate_rocksdb_payload_storage: true,
             migrate_rocksdb_payload_indices: true,
             appendable_quantization: true,
+            single_file_mmap_vector_storage: true,
```
timvisee: This should still be false?


```suggestion
            single_file_mmap_vector_storage: false,
```

lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
```diff
@@ -265,6 +283,7 @@ impl<T: PrimitiveVectorElement> VectorStorage for MemmapDenseVectorStorage<T> {
             &self.deleted_path,
             dim,
             with_async_io,
+            false, // No need to populate - data was just written to page cache
```
timvisee: Maybe we can still do it for good measure? Or is it costly in some way?

lib/segment/src/segment_constructor/segment_constructor_base.rs
timvisee: I'd also be fine by just putting `false` as parameter.

ffuugoo:
I have a proposal how to make this sane, because right now it's completely unreadable.

```rust
match (vector_config.storage_type, vector_storage.multivector_config) {
    (VectorStorageType::Mmap, None) => open_mmap_vector_storage(..., true),
    (VectorStorageType::InRamMmap, None) => open_mmap_vector_storage(..., true),
    (VectorStorageType::Mmap, Some(multivector)) => open_appendable_memmap_multi_vector_storage(..., false),
    (VectorStorageType::InRamMmap, Some(multivector)) => open_appendable_memmap_multi_vector_storage(..., true),
    // etc...
}
```

lib/common/common/src/flags.rs
generall: yes, right. This is no agent mistake, this one is on me;)

lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
generall: no, it is not needed. This operation is about merging segments during optimization, we don't actually want it to populate cache, I will remove comment

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -316,31 +299,60 @@ impl<
...
+            open_memmap_vector_storage(vector_storage_path, size, distance, madvise, populate)
+        }
+        VectorStorageDatatype::Uint8 => {
+            open_memmap_vector_storage_byte(vector_storage_path, size, distance, madvise, populate)
+        }
+        VectorStorageDatatype::Float16 => {
+            open_memmap_vector_storage_half(vector_storage_path, size, distance, madvise, populate)
+        }
+    }
+}
```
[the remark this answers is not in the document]
generall: yeap, this was a bug

timvisee:
Since this is feature flagged it should be safe to merge. Doing so now.

PR #7976, opened 2026-01-23
Persist applied seq at regular interval

agourlay:
In the context of the new upcoming update queue, we need to have a clear view on which updates are applied but not yet flushed.

Concretely, the blue elements in the WAL diagram below.

image

Refer to design doc for more information https://www.notion.so/qdrant/Update-queue-2ee674779d3380b18636dd858ceb9099

To that end this PR adds a tracking of the latest applied seq on a local shard basis.

I have decided to use a JSON object format to give us the opportunity to extend if necessary in the future.

Saving this information on disk requires strong durability, therefore in order to amortize the cost of `fsync` we perform it only each 64 updates.

Preliminary benchmark results show that this does not impact the performance of updates.

This value is currently only written and not used for anything.

## Future work

Use the last `applied_seq` during restart.

### DIFF
--- a/lib/collection/src/update_workers/applied_seq.rs
+++ b/lib/collection/src/update_workers/applied_seq.rs
@@ -13,6 +13,7 @@ use crate::operations::types::CollectionResult;
 /// How often the `applied_seq` is persisted
 pub const APPLIED_SEQ_SAVE_INTERVAL: u64 = 64;
 
+/// Data structure, used for (de)serialization of the `applied_seq` file
 #[derive(Debug, Default, Clone, Serialize, Deserialize)]
 struct AppliedSeq {
     pub op_num: u64,


lib/collection/src/update_workers/update_worker.rs
```diff
@@ -93,6 +95,10 @@ impl UpdateWorkers {
                         Err(err) => Err(CollectionError::from(err)),
                     };
 
+                    if let Err(err) = applied_seq_handler.update(op_num) {
+                        log::error!("Can't update last applied_seq {err}")
```
IvanPleshkov, line 126: Do we need to use `feedback.send` to notify this error and stop this operation by calling `continue`?

lib/collection/src/update_workers/applied_seq.rs
```diff
@@ -0,0 +1,123 @@
...
+    #[cfg(test)]
+    fn persisted_op_num(&self) -> u64 {
+        read_json::<AppliedSeq>(&self.path).unwrap().op_num
+    }
+
+    pub fn load_or_init(shard_path: &Path) -> CollectionResult<Self> {
+        let path = shard_path.join(APPLIED_SEQ_FILE);
+        let file = SaveOnDisk::load_or_init_default(&path)?;
+
+        let op_num = AtomicU64::new(0);
```
IvanPleshkov: Why `op_num` is not read from `file`?

lib/collection/src/update_handler.rs
```diff
@@ -149,8 +153,9 @@ impl UpdateHandler {
         shard_path: PathBuf,
         scroll_read_lock: Arc<tokio::sync::RwLock<()>>,
         update_tracker: UpdateTracker,
-    ) -> UpdateHandler {
-        UpdateHandler {
+    ) -> CollectionResult<Self> {
+        let applied_seq_handler = AppliedSeqHandler::load_or_init(&shard_path)?;
```
timvisee: Would it make sense to handle load failures here?

If we cannot load from disk due to corrupted data, maybe we can force replay everything on restart. I'd be better than flat out crashing. Using that suggested `None` variant would probably be good here.

lib/collection/src/update_workers/applied_seq.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    #[cfg(test)]
+    fn persisted_op_num(&self) -> u64 {
+        read_json::<AppliedSeq>(&self.path).unwrap().op_num
+    }
+
+    pub fn load_or_init(shard_path: &Path) -> CollectionResult<Self> {
+        let path = shard_path.join(APPLIED_SEQ_FILE);
+        let file = SaveOnDisk::load_or_init_default(&path)?;
```
IvanPleshkov: Is default zero a good value for the op_num when the file is not existed?

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -93,6 +95,10 @@ impl UpdateWorkers {
                         Err(err) => Err(CollectionError::from(err)),
                     };
 
+                    if let Err(err) = applied_seq_handler.update(op_num) {
```
timvisee, line 125: Note to self: we also want to update on wait=false requests, because from a logical data perspective, we don't want to go back in time after restart.

lib/collection/src/update_workers/applied_seq.rs
IvanPleshkov: I find the last index of WAL a better default value because we want to use op_num as a red point index - operations which must be applied while replay

lib/collection/src/update_workers/applied_seq.rs
agourlay: :face_with_head_bandage: 
https://github.com/qdrant/qdrant/pull/7976/commits/6913237c787d146af5256fc291c7895c02add8cc

lib/collection/src/update_workers/update_worker.rs
agourlay: I think this error is not visible for the user.
Only the result of the operation matters.
Do you see things differently?

lib/collection/src/update_workers/applied_seq.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// Update the underlying file every `APPLIED_SEQ_SAVE_INTERVAL` updates.
+    /// Always update memory representation
+    pub fn update(&self, op_num: u64) -> CollectionResult<()> {
+        // update in-memory
+        self.op_num.store(op_num, Ordering::Relaxed);
+        let prev_count = self.update_count.fetch_add(1, Ordering::Relaxed);
+        // update on disk according to interval to amortize fsync
+        if prev_count + 1 >= APPLIED_SEQ_SAVE_INTERVAL {
+            self.save(op_num)?;
+            self.update_count.store(0, Ordering::Relaxed);
```
timvisee: I'd rather remove this line and call save each `prev_count % 64 == 0` instead.

That separate store could data race in a case where we'd call this concurrently. I understand that should never happen in the current implementation, but you never know how things change in the future. Because of it I'd like the above suggestion better with just a single atomic call.

agourlay: https://github.com/qdrant/qdrant/pull/7976/commits/8dfb2d223535bd0399634ccfc88fbf0e583e3e77

lib/collection/src/update_workers/applied_seq.rs
```diff
@@ -0,0 +1,147 @@
...
+            let timestamp = SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_secs();
+            debug_assert!(current.timestamp <= timestamp);
+            current.op_num = op_num;
+            current.timestamp = timestamp;
+        })?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
agourlay: removed

lib/collection/src/update_workers/update_worker.rs
IvanPleshkov: It's hard to say.
Let's image if op_num is not stored.
Then, in case of crash, just part of blue operations will be replayed. Everything else is still in WAL. In means for some time after http client starting, the data is not correct (because of red operation on diagram).
It's hard to say is it better than skipping operation if op_num is not applied.

I let you decide

agourlay:
I have pushed a significant change to handle the case of failing to load.
Please review again :pray:

lib/collection/src/update_workers/applied_seq.rs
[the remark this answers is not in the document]
agourlay: I pushed a significant change, please review the whole thing again!

lib/collection/src/update_workers/applied_seq.rs
```diff
@@ -0,0 +1,219 @@
...
+                    path,
+                    op_num,
+                    update_count,
+                }
+            }
+        }
+    }
+
+    fn save(&self, op_num: u64) -> CollectionResult<()> {
+        if let Some(file) = self.file.as_ref() {
```
IvanPleshkov, line 118: If file was not loaded, there is no chance to update this file. Even if we have a new actual op number. What happens if out-of-disk appears? Does it mean that in case of OOD we will replay the whole large update queue with yellow operations?

IvanPleshkov: I find this question important to block merge, while I didn't get a recovery mechanism

agourlay: Alright I will regenerate a fresh file in that case with the first entry from WAL

IvanPleshkov: Or not, if by design your approach works better. Right now, it's hard to understand for me if it's ok that corruped file (or no file because of OOD) should disable this feature

lib/collection/src/update_handler.rs
```diff
@@ -149,7 +153,9 @@ impl UpdateHandler {
         shard_path: PathBuf,
         scroll_read_lock: Arc<tokio::sync::RwLock<()>>,
         update_tracker: UpdateTracker,
-    ) -> UpdateHandler {
+    ) -> Self {
+        let wal_first_index = wal.lock().await.first_index();
```
agourlay: Needed to make the function async to lock the WAL because the `blocking_lock` caused

```
Panic occurred in file lib/collection/src/update_handler.rs at line 158: Cannot block the current thread from within a runtime. This happens because a function attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

lib/collection/src/update_workers/applied_seq.rs
```diff
@@ -0,0 +1,243 @@
...
+
+    #[cfg(test)]
+    fn persisted_op_num(&self) -> u64 {
+        read_json::<AppliedSeq>(&self.path).unwrap().op_num
+    }
+
+    /// Load or create the underlying applied seq file.
+    ///
+    /// `wal_first_index` is expected to be the next used WAL index.
+    pub fn load_or_init(shard_path: &Path, wal_first_index: u64) -> Self {
```
timvisee: Though this is a little complex, I like how we handle edge cases now 👍 

E.g. if disk is full, we don't crash. If the data is malformed, we try to recreate it replaying 100% of the WAL.

agourlay:
Looks like I broke the integration tests somehow :detective:

agourlay:
It seems my debug assertion is incorrect when performing transfers.

```
ERROR qdrant::startup: Panic occurred in file lib/collection/src/update_workers/applied_seq.rs at line 76: persisted applied_seq:65 cannot be larger than the last_applied_wal_index:0
```

lib/collection/src/update_handler.rs
IvanPleshkov: I expect this function is called after replay and wal truncating, isn't it?

lib/collection/src/update_workers/applied_seq.rs
```diff
@@ -0,0 +1,247 @@
...
```
timvisee: We must use the last index, not the first.

Adding this to prevent accidental merger.

timvisee:
This probably needs adjusting too: https://github.com/qdrant/qdrant/blob/9c5cec1b876343f9d9983444af275c8487b4ed72/lib/shard/src/files/mod.rs#L40-L48

lib/shard/src/snapshots/snapshot_utils.rs
```diff
@@ -100,13 +100,15 @@ impl SnapshotUtils {
             segments_path: from_segments_path,
             newest_clocks_path: from_newest_clocks_path,
             oldest_clocks_path: from_oldest_clocks_path,
+            applied_seq_path: _, // not part of the snapshot
```
timvisee: Shouldn't it be part of the snapshot?

In most cases the segment files on disk should reflect the latest applied state.

But if we have a large pending queue of operations in the WAL, I guess it'd be nice to include all of it in the snapshot. Then we'd need the applied seq file too to not force replay everything on recovery, so that it'll behave the same as the source shard.

lib/collection/src/update_workers/applied_seq.rs
```diff
@@ -0,0 +1,242 @@
...
+                    log::error!("Error while creating new applied_seq at {path:?} {err}");
+                    Self {
+                        file: None,
+                        path,
+                        op_num: AtomicU64::new(wal_last_index),
+                        update_count,
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
agourlay: If `fs::remove_file` succeeds, the recursion will hit a different branch as the file does not exist anymore.
It will simply error out on failing to create.

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -250,6 +265,10 @@ impl LocalShard {
             .unwrap_or_default()
             .then(|| config.optimizer_config.get_indexing_threshold_kb());
 
+        let wal_last_index = locked_wal.lock().await.last_index();
+        let applied_seq_handler =
+            Arc::new(AppliedSeqHandler::load_or_init(shard_path, wal_last_index));
+
```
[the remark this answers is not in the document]
agourlay: `applied_seq.json` is now propagated via the snapshot.

Can you elaborate regarding when this assertion can trigger exactly?

lib/collection/src/shards/local_shard/snapshot.rs
```diff
@@ -181,6 +184,22 @@ impl LocalShard {
...
+        tar: &tar_ext::BuilderExt,
+    ) -> CollectionResult<()> {
+        if applied_seq_path.exists() {
+            tar.blocking_append_file(applied_seq_path.as_path(), Path::new(APPLIED_SEQ_FILE))
+                .map_err(|err| {
+                    CollectionError::service_error(format!(
+                        "Error while archiving applied_seq: {err}"
+                    ))
+                })?;
+        }
```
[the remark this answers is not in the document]
agourlay: When do you think the `applied_seq` file can be deleted?

lib/collection/src/shards/local_shard/mod.rs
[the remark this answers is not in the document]
agourlay: > WAL truncation (after flush) can reduce last_index

WAL truncation does not decrease the WAL logical index.
It is used to physically delete the entries that precede a given index.

PR #7983, opened 2026-01-24
gridstore iter livelock

generall:
Currently, there is a livelock problem, which reproduces in the following situation:

- we have a stream of search requests hitting a collection
- On the same collection we run create_payload request, which is supposed to create an index for unrelated field

During index building search gets stuck.

Why it happens?

- For building index we iterate over payload storage, which is currently implemented via gridstore.
- We read payloads, and everything is good until suddenly it stucks. Why?
- Flush happes and tries to lock `tracker` inside gridstore. It is unternally mutable.
- Flush can't lock the `tracker` as we have an iterator going on, but it also stays in a queue and prevents incomming requests from reading gridstore. 
- So, everything is waiting on the iterator to finish.

This PR changes behaviour of the iterator, so it doesn't require locking the tracker whole time.
It also includes couple of cosmetic changes

### DIFF
--- a/lib/gridstore/src/gridstore.rs
+++ b/lib/gridstore/src/gridstore.rs
@@ -1,3 +1,4 @@
+use std::cmp::min;
 use std::io::BufReader;
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -103,7 +104,7 @@ impl<V: Blob> Gridstore<V> {
         self.pages.read().len() as PageId
     }
 
-    pub fn max_point_id(&self) -> PointOffset {
+    pub fn max_point_offset(&self) -> PointOffset {
         self.tracker.read().pointer_count()
     }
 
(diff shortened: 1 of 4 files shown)


lib/gridstore/src/gridstore.rs
```diff
@@ -532,28 +533,50 @@ impl<V: Blob> Gridstore<V> {
     where
         F: FnMut(PointOffset, V) -> std::result::Result<bool, E>,
     {
-        for (point_offset, pointer) in
-            self.tracker
-                .read()
-                .iter_pointers()
-                .filter_map(|(point_offset, opt_pointer)| {
-                    opt_pointer.map(|pointer| (point_offset, pointer))
-                })
```
generall, line 541: this locked tracker for the whole existance of the iterator

lib/gridstore/src/gridstore.rs
```diff
@@ -600,7 +623,8 @@ impl<V> Gridstore<V> {
                 page.flush()?;
             }
 
-            let old_pointers = tracker.write().write_pending_and_flush(pending_updates)?;
+            let old_pointers = tracker.write().write_pending(pending_updates);
```
generall, line 626: this was waiting for iterator and blocking everything

lib/gridstore/src/gridstore.rs
```diff
@@ -532,28 +533,50 @@ impl<V: Blob> Gridstore<V> {
...
-                return Ok(());
+            // Process buffer without holding the tracker lock
+            for &(point_offset, pointer) in &buffer {
+                let ValuePointer {
+                    page_id,
+                    block_offset,
+                    length,
+                } = pointer;
+
+                let raw = self.read_from_pages::<true>(page_id, block_offset, length);
```
timvisee, line 571: This looks sound 👍 

I was afraid it could data race where newly incoming data could be written to pages before we read here. But writes require a `&mut self`, while this holds a `&self`.

PR #7986, opened 2026-01-26
chore: Optimize the shard loading process during startup

1995chen:
### All Submissions:

We optimized the data-loading phase during Qdrant startup by changing the original serial shard-loading process to a concurrent one, and we updated MAX_CONCURRENT_SEGMENT_LOADS so that segments can be loaded faster on machines with more CPUs; the original hard-coded value of 8 was too slow for large datasets.

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -80,7 +80,7 @@ use crate::optimizers_builder::{OptimizersConfig, build_optimizers, clear_temp_s
 use crate::shards::CollectionId;
 use crate::shards::shard::ShardId;
 use crate::shards::shard_config::ShardConfig;
-use crate::update_handler::{Optimizer, UpdateHandler, UpdateSignal};
+use crate::update_handler::{OperationData, Optimizer, UpdateHandler, UpdateSignal};
 use crate::update_workers::applied_seq::AppliedSeqHandler;
 use crate::wal_delta::RecoverableWal;
 
@@ -682,7 +682,17 @@ impl LocalShard {
     pub async fn load_from_wal(&self, collection_id: CollectionId) -> CollectionResult<()> {
         let mut newest_clocks = self.wal.newest_clocks.lock().await;
         let wal = self.wal.wal.lock().await;
-        let bar = ProgressBar::new(wal.len(false));
+
+        let from = wal.first_index();
+        let last_wal_index = from + wal.len(false);
+        let to = self
+            .applied_seq_handler
+            .op_num_upper_bound()
+            .unwrap_or(last_wal_index);
+        let to = std::cmp::min(to, last_wal_index);
+        let wal_entries_to_replay = to - from;
+
+        let bar = ProgressBar::new(wal_entries_to_replay);
 
         let progress_style = ProgressStyle::default_bar()
             .template("{msg} [{elapsed_precise}] {wide_bar} {pos}/{len} (eta:{eta})")
@@ -690,10 +700,11 @@ impl LocalShard {
         bar.set_style(progress_style);
 
         log::debug!(
-            "Recovering shard {} starting reading WAL from {} up to {}",
+            "Recovering shard {} starting reading WAL from {} up to {} (last_applied_seq:{:?})",
             self.path.display(),
-            wal.first_index(),
-            wal.last_index(),
+            from,
+            to,
+            self.applied_seq_handler.op_num(),
         );
 
         bar.set_message(format!("Recovering collection {collection_id}"));
@@ -705,7 +716,7 @@ impl LocalShard {
             log::info!(
                 "Recovering shard {}: 0/{} (0%)",
                 self.path.display(),
-                wal.len(false),
+                wal_entries_to_replay,
             );
         }
 
@@ -722,7 +733,7 @@ impl LocalShard {
         // (`SerdeWal::read_all` may even start reading WAL from some already truncated
         // index *occasionally*), but the storage can handle it.
 
-        for (op_num, update) in wal.read_all(false) {
+        for (op_num, update) in wal.read_range(from..to) {
             if let Some(clock_tag) = update.clock_tag {
                 newest_clocks.advance_clock(clock_tag);
             }
@@ -767,8 +778,8 @@ impl LocalShard {
                 let progress = bar.position();
                 log::info!(
                     "{progress}/{} ({}%)",
-                    wal.len(false),
-                    (progress as f32 / wal.len(false) as f32 * 100.0) as usize,
+                    wal_entries_to_replay,
+                    (progress as f32 / wal_entries_to_replay as f32 * 100.0) as usize,
                 );
                 last_progress_report = Instant::now();
             }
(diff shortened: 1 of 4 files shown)


lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -871,41 +872,64 @@ impl ShardHolder {
...
-                this_peer_id,
-                update_runtime.clone(),
-                search_runtime.clone(),
-                optimizer_resource_budget.clone(),
-            )
-            .await;
+                // Validate that shard exists on disk
+                let shard_path = check_shard_path(collection_path, shard_id)
+                    .await
+                    .expect("Failed to check shard path");
```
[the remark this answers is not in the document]
1995chen: @generall Could you help me check whether the timing of this panic is appropriate?

timvisee:
Whether this is efficient highly depends on hardware and cluster configuration. We've usually seen disk throughput (not IOPS) being a bottleneck.

Because of it I can't just blindly trust the PR and we'd have to have some benchmarks for this.

Do you have any short of benchmarks on this one you can share?

1995chen:
> Whether this is efficient highly depends on hardware and cluster configuration. We've usually seen disk throughput (not IOPS) being a bottleneck.
> 
> Because of it I can't just blindly trust the PR and we'd have to have some benchmarks for this.
> 
> Do you have any short of benchmarks on this one you can share?

Of course, we're currently conducting tests in this area. If they prove effective, we'll share this data with you as soon as possible.

1995chen:
> Whether this is efficient highly depends on hardware and cluster configuration. We've usually seen disk throughput (not IOPS) being a bottleneck.
> 
> Because of it I can't just blindly trust the PR and we'd have to have some benchmarks for this.
> 
> Do you have any short of benchmarks on this one you can share?

We conducted tests on AWS EBS (16k IOPS, 1000 throughput). Before the changes, the startup time lasted from 2026-01-26 11:03:00 to 2026-01-26 12:16:00, totaling 70 minutes. 
image
image
As seen from the screenshots, the intermittent drops in read/write operations were due to reading the next shard. When we used the image built from this PR for startup, the startup time lasted from 2026-01-27 11:02:00 to 2026-01-27 11:48:00, totaling 46 minutes.
image
image
Previously, we were at a loss with this situation because during startup, the IOPS of 2.73k and throughput of 682.22 were far below our configured values. However, after this update, the situation has improved significantly—the IOPS reached 4,002, and the throughput hit 1,000, directly reaching the maximum configured value. I believe that if we were given a higher throughput configuration, the startup time could potentially be reduced to under 30 minutes.

1995chen:
our test collection config:
```shell
{
     "status": "green",
     "optimizer_status": "ok",
     "indexed_vectors_count": 261495088,
     "points_count": 261495088,
     "segments_count": 75,
     "config": {
          "params": {
               "vectors": {
                    "passage_vector": {
                         "size": 768,
                         "distance": "Dot",
                         "hnsw_config": {
                              "m": 48,
                              "ef_construct": 512,
                              "on_disk": false,
                              "payload_m": 48,
                              "inline_storage": true
                         },
                         "quantization_config": {
                              "binary": {
                                   "always_ram": true,
                                   "encoding": "two_bits"
                              }
                         },
                         "on_disk": true
                    }
               },
               "shard_number": 15,
               "sharding_method": "auto",
               "replication_factor": 1,
               "write_consistency_factor": 2,
               "on_disk_payload": true
          },
          "hnsw_config": {
               "m": 16,
               "ef_construct": 100,
               "full_scan_threshold": 10000,
               "max_indexing_threads": 0,
               "on_disk": false
          },
          "optimizer_config": {
               "deleted_threshold": 0.2,
               "vacuum_min_vector_number": 1000,
               "default_segment_number": 4,
               "max_segment_size": 16777216,
               "memmap_threshold": null,
               "indexing_threshold": 30000,
               "flush_interval_sec": 5,
               "max_optimization_threads": null,
               "prevent_unoptimized": null
          },
          "wal_config": {
               "wal_capacity_mb": 32,
               "wal_segments_ahead": 0,
               "wal_retain_closed": 1
          },
          "quantization_config": null
     }
}`

**Comment by 1995chen** [comment]
Also, I’m a bit lost here—could you guys clear this up for me?
https://github.com/qdrant/qdrant/pull/3745/changes#r2731093238

**Comment by 1995chen** [comment]
> Also, I’m a bit lost here—could you guys clear this up for me? https://github.com/qdrant/qdrant/pull/3745/changes#r2731093238

If this is wrong, just tell me what the expected behavior should be and I’ll fix it.

**Comment by timvisee** [comment]
> Also, I’m a bit lost here—could you guys clear this up for me? https://github.com/qdrant/qdrant/pull/3745/changes#r2731093238

What are you asking exactly?

File: lib/common/common/src/defaults.rs [hunk 2]
```diff
@@ -17,7 +17,16 @@ pub static APP_USER_AGENT: LazyLock<String> =
     LazyLock::new(|| format!("Qdrant/{QDRANT_VERSION_STRING}"));
 
 /// Maximum number of segments to load concurrently when loading a collection.
-pub const MAX_CONCURRENT_SEGMENT_LOADS: usize = 8;
+///
+/// Dynamic based on CPU size, with a minimum of 8.
+pub static MAX_CONCURRENT_SEGMENT_LOADS: LazyLock<usize> =
+    LazyLock::new(|| cpu::get_num_cpus().max(8));
```
timvisee: This is actually mostly about memory usage. Loading a lot of segments concurrently allows a lot of concurrent segment migrations, which might consume a lot of memory.

Having a machine with more CPU does not strictly have more memory.

Do you see similar performance if you keep this value at 8?

lib/common/common/src/defaults.rs
```diff
@@ -17,7 +17,16 @@ pub static APP_USER_AGENT: LazyLock<String> =
...
+
+/// Maximum number of shard to load concurrently when loading a collection.
+///
+/// Dynamic based on CPU size, with a minimum of 8.
+pub static MAX_CONCURRENT_SHARD_LOADS: LazyLock<usize> =
+    LazyLock::new(|| cpu::get_num_cpus().max(8));
```
timvisee: I think it'd make sense to move this to a configurable setting in `config.yaml`.

1995chen: Hi @timvisee, thank you for the valuable input. You're absolutely right that user control is the better option. I've completed the PR as suggested. Would you mind reviewing it again at your convenience?

1995chen:
> > Also, I’m a bit lost here—could you guys clear this up for me? https://github.com/qdrant/qdrant/pull/3745/changes#r2731093238
> 
> What are you asking exactly?

I understand this method groups requests by key. With the current implementation, if a user makes requests like:
```rust
run_batch_requests(&[('a', 1), ('b', 2), ('a', 3)])
```
they will be grouped as:
```rust
vec![
    ('a', vec![('a', 1)]),
    ('b', vec![('b', 2)]),
    ('a', vec![('a', 3)]),
]
```
This behavior works fine. However, I'd like to consult whether it would be better to modify it to:
```rust
vec![
    ('a', vec![('a', 1),('a', 3)]),
    ('b', vec![('b', 2)]),
]
```
Since the current ordering causes the grouping to be fragmented/interrupted for the same key, is this something worth optimizing?

1995chen:
@timvisee  @generall What remaining concerns do you have about this PR? Could you share your feedback with us?

1995chen (commented):
All review comments have been addressed.

PR #7999, opened 2026-01-26
do not block ToC on in-collection operations by arc-ing collections

generall:
This PR attempts to address the following observed behavoir:

- 1 collection exist in the cluster, this collection receives heavy traffic.
- We create another collection nearby
- Collection creation and traffic block each other.

Why it might happen?

When we perform any operation like search, upsert, snapshots ... we lock `TableOfContent` for read. 
When we want to create new collection, we need to lock `TableOfContent` for writes.
Lock for write works like this:

- it waits for all current operations to finish (this might be couple of seconds+)
- it doesn't allow any new operations to preceed
- queue of operations builds up
- result == longer latency, maybe even timeout

(unconfirmed, but it might be, that queue of operations might also cause exhaustion of the threads)

This PR removes requirement to hold `TableOfContent` lock for collection ops, making critical section significantly smaller.
As a result: write doesn't wait for queue => operations don't have time to queue up.

Downside: we have a bit less strict guarantees about collection ownership during deletion of the collection, so we need to spinlock there.

### DIFF
--- a/lib/storage/src/content_manager/toc/collection_meta_ops.rs
+++ b/lib/storage/src/content_manager/toc/collection_meta_ops.rs
@@ -212,7 +212,8 @@ impl TableOfContent {
         let collection_path = self.get_collection_path(collection_name);
         let safe_delete_path = self.storage_config.storage_path.join(".deleted");
 
-        if let Some(removed) = self.collections.write().await.remove(collection_name) {
+        let removed_opt = self.collections.write().await.remove(collection_name);
+        if let Some(removed) = removed_opt {
             if let Some(state) = removed.resharding_state().await
                 && let Err(err) = removed.abort_resharding(state.key(), true).await
             {


lib/storage/src/common/utils.rs
```diff
@@ -0,0 +1,23 @@
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+pub async fn try_unwrap_with_timeout_async<T>(
```
timvisee, line 4: We only spin lock on:
- collection deletion
- collection creation, while collection already existed (unlikely)

No concerns there 👍

lib/storage/src/content_manager/toc/create_collection.rs
```diff
@@ -243,13 +245,32 @@ impl TableOfContent {
...
+                write_collections.insert(collection_name.to_string(), Arc::new(collection));
             if let Some(existing_collection) = existing_collection {
                 debug_assert!(
                     false,
                     "Collection `{collection_name}` was not expected to exist"
                 );
+
                 existing_collection.stop_gracefully().await;
+
+                let removed_collection_res = try_unwrap_with_timeout_async(
```
timvisee, line 257: Note: concurrent collection creation calls are not a concern here. They are serialized by a 'creation' lock and consensus.

PR #8000, opened 2026-01-27
Drop wal from api

IvanPleshkov:
## Summary
This PR adds an experimental non-documented feature that allows to drop the unapplied records in the WAL.

It's helpful in cases where the update queue is stuck with tons of unapplied requests.

## Description
PR adds a new HTTP entry `POST /collection/collection_name/drop_wal`. This command removes all WAL entries, which are guaranteed to be non-applied. It performs the operation over the local shard. In distributed systems, this command should be called on each node.

The result is a simple json with an integer with the amount of dropped records in all shards of the collection. It's not a json because of its experimental feature without documentation.

It also has a `wait` flag for the cases, where WAL drop takes a lot of time and may be timeouted (example of a dropping performance is below).

## Security
Not sure if some strict mode checks are required. However, WAL dropping is allowed only for users with `manage` RBAC access.

## Implementation details
Providing the operation from endpoint up to shard is clear.

On the shard level, it's not enough to just clear the WAL. Because the update operations are also stored in the update worker channel, where each operation has a seq num already.

Before locking the WAL, the truncate operation clear first the update worker channel by the following steps:
1. This PR addes a update lock, which can lock sender of the update worker channel. Lock it first
2. This PR addes a skip updates flag, which on receiver skip update operation. Set this flag to true as a second step.
3. Next, send the plunger to the update worker to get the point, where all pendings in channel are skipped.
4. After plunger response, it's save to lock WAL (but still keep update lock)

After this steps, it's safe to call truncation in WAL, which is also trivial.

## Plunger usage safety
At first glance, plunger usage might cause a deadlock when the update worker has `Stop` operation. In this case, it's easy to expect that the plunger sender is not consumed and the receiver will wait forever.

But this easy test show that if `oneshot::Sender` is destroyed, the connected receiver is not blocked and returns `RecvError`:

```rust
#[tokio::test]
async fn test_update_signal_plugrer_drop() {
    let (sender, receiver) = channel::<UpdateSignal>(10);
        
    let (plunger_tx, plunger_rx) = oneshot::channel();
    let plunger = UpdateSignal::Plunger(plunger_tx);
    sender.send(plunger).await.unwrap();

    // Simulate stopping the update worker by dropping the receiver.
    let handler = tokio::spawn(async move {
        std::thread::sleep(std::time::Duration::from_millis(100));
        drop(receiver);    
    });

    // Check that the plunger finished even it is not explicitly plunged.
    let plunger_result = plunger_rx.await;
    println!("plunger result: {:?}", plunger_result);
    assert!(plunger_result.is_err());

    handler.await.unwrap();
}
```
This test is not included because it's a tokio test; there is no qdrant code usage here.

## Testing
This PR was tested manually using BFB. At this section, the manual test case is described with observable result.

Steps to test:
1. Increase the update queue limit by changing hardcoded const `DEFAULT_UPDATE_QUEUE_SIZE` up to 10M.
5. Run this patched qdrant in release mode
6. Run BFB with commands: `bfb --threads=8 --num-vectors=10000000 --indexing-threshold=0`
7. When BFB is close to the end of the upload operations, clean the collected WAL using this bash:
```
curl -X POST "http://localhost:6333/collections/benchmark/drop_wal?wait=true"
```

Example of the output
```
{"result":74923,"status":"ok","time":0.866964453}
```
8. Check in WebUI the amount of existing points. It should be less that 10M because points were removed from the WAL.
By example above, 75K records were removed (it's 7.5M points because the batch is equal to 100), and there are only 2.5M points in collection.
Also regarging to the measurement, `0.86s` was required to clear 7.5M points.

## Open problems
There is a problem with replicas. Because the applied `seq_num` is not synchronized between replicas, some points were applied in only some replicas (but at least in one). This PR does not provide a solution for it now

### DIFF
--- a/lib/collection/src/collection/state_management.rs
+++ b/lib/collection/src/collection/state_management.rs
@@ -274,11 +274,6 @@ impl Collection {
             })
             .await?;
 
-        let mut total_dropped_records = 0;
-        for result in results {
-            total_dropped_records += result?;
-        }
-
-        Ok(total_dropped_records)
+        results.into_iter().sum()
     }
 }
--- a/src/actix/api/service_api.rs
+++ b/src/actix/api/service_api.rs
@@ -229,12 +229,21 @@ async fn truncate_unapplied_wal(
     ActixAccess(access): ActixAccess,
 ) -> impl Responder {
     let future = async move {
-        let _ = access.check_global_access(AccessRequirements::new().manage())?;
+        let collection_pass = access
+            .check_global_access(AccessRequirements::new().manage())?
+            .issue_pass(&collection.name)
+            .into_static();
+
         let pass = new_unchecked_verification_pass();
-        dispatcher
+        let collection = dispatcher
             .toc(&access, &pass)
-            .truncate_unapplied_wal(&collection.name, access)
+            .get_collection(&collection_pass)
+            .await?;
+
+        collection
+            .truncate_unapplied_wal()
             .await
+            .map_err(StorageError::from)
     };
     helpers::time_or_accept(future, params.wait.unwrap_or(true)).await
 }
--- a/lib/storage/src/content_manager/toc/mod.rs
+++ b/lib/storage/src/content_manager/toc/mod.rs
@@ -702,24 +702,4 @@ impl TableOfContent {
     pub fn general_runtime_handle(&self) -> &Handle {
         self.general_runtime.handle()
     }
-
-    /// Truncate unapplied WAL records for a given collection.
-    /// Returns amount of removed records.
-    pub async fn truncate_unapplied_wal(
-        &self,
-        collection_name: &str,
-        access: Access,
-    ) -> Result<usize, StorageError> {
-        let requirements = AccessRequirements::new().manage();
-        let collection_pass = access
-            .check_collection_access(collection_name, requirements)?
-            .into_static();
-
-        let collection = self.get_collection(&collection_pass).await?;
-
-        collection
-            .truncate_unapplied_wal()
-            .await
-            .map_err(StorageError::from)
-    }
 }


src/actix/api/service_api.rs
```diff
@@ -212,6 +216,35 @@ async fn update_logger_config(
...
+#[post("/collections/{name}/drop_wal")]
+async fn drop_wal(
+    dispatcher: web::Data<Dispatcher>,
+    collection: Path<CollectionPath>,
+    params: Json<DropWalParams>,
+    ActixAccess(access): ActixAccess,
+) -> impl Responder {
+    // nothing to verify.
+    let future = async move {
+        let pass = new_unchecked_verification_pass();
```
generall, line 237: are you sure?

IvanPleshkov: It's a good question. There is a requirement for `manage` access in ToC. Here is a strict mode check. I'm actually not sure if we need some check for the strict mode.

timvisee: Let's enforce it here too, like:

https://github.com/qdrant/qdrant/blob/4c917ea155e6be8130481656dd7ef1da2b298b7c/src/actix/api/service_api.rs#L202

lib/shard/src/wal.rs
```diff
@@ -240,6 +240,12 @@ impl<R: DeserializeOwned + Serialize> SerdeWal<R> {
         let normal_retention = self.options.retain_closed.get();
         self.wal.set_retention(normal_retention);
     }
+
+    pub fn drop_from(&mut self, from_index: u64) -> Result<()> {
+        self.wal
+            .truncate(from_index)
+            .map_err(|err| WalError::TruncateWalError(format!("{err:?}")))
+    }
```
timvisee, line 249: Let's add a (debug) assertion here:

`from_index >= self.first_index()`

to prevent truncating entries that we already acknowledged. It won't prevent truncating items that have just been applied but it's better than nothing.

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -37,6 +38,7 @@ impl UpdateWorkers {
         optimization_handles: Arc<TokioMutex<Vec<StoppableTaskHandle<bool>>>>,
         mut optimization_finished_receiver: watch::Receiver<()>,
         applied_seq_handler: Arc<AppliedSeqHandler>,
+        skip_uploads: Arc<AtomicBool>,
```
timvisee: I suggest to rename this to `skip_updates`, or even `skip_operations`.

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -46,6 +48,10 @@ impl UpdateWorkers {
                     sender,
                     hw_measurements,
                 }) => {
+                    if skip_uploads.load(std::sync::atomic::Ordering::Relaxed) {
+                        continue;
```
timvisee: The idea is that we drop the entire operation, including the `sender`. And other places that wait on such sender should therefore error and return early. Am I correct?

If that is true, I vote to add a comment describing this.


```suggestion
                    if skip_uploads.load(std::sync::atomic::Ordering::Relaxed) {
                        // This will also drop `sender`, unblocking any operations waiting on it
                        continue;
```

lib/collection/src/shards/local_shard/wal_ops.rs
```diff
@@ -11,4 +16,72 @@ impl LocalShard {
     pub async fn set_normal_wal_retention(&self) {
         self.wal.set_normal_retention().await;
     }
+
+    /// Drop WAL from the last applied sequence number.
+    /// Returns amount of removed records.
+    pub async fn drop_wal(&self) -> CollectionResult<usize> {
+        // First, lock the WAL to prevent new writes.
+        let mut wal_lock = Mutex::lock_owned(self.wal.wal.clone()).await;
```
timvisee: I'm afraid this can deadlock if the update worker just started processing a new operation, and tries to lock the WAL to get the operation for a pointer here:

https://github.com/qdrant/qdrant/blob/4c917ea155e6be8130481656dd7ef1da2b298b7c/lib/collection/src/update_workers/update_worker.rs#L64

lib/collection/src/shards/local_shard/wal_ops.rs
```diff
@@ -11,4 +16,72 @@ impl LocalShard {
...
+
+        // Next, lock the update handler to prevent changes in update handler state.
+        let update_handler = self.update_handler.lock().await;
+
+        // If there are no `op_num` in the `applied_seq`, it's unsafe to drop WAL.
+        // Because there is no good save sequence number to drop from.
+        let applied_seq = update_handler.applied_seq();
```
timvisee: I'm afraid this might not match up with the queue/WAL. Because this number is only updated **after** application, and not at the moment we popped it from the channel.

Would it be possible to:
- set skip updates flag
- store `op_num` of first operation that is popped of the channel
- drain whole channel
- truncate from that `op_num` onwards

because then we can **ensure** that `op_num` never hit the update worker and everything before it is (eventually) applied.

lib/collection/src/shards/local_shard/wal_ops.rs
```diff
@@ -11,4 +16,72 @@ impl LocalShard {
...
+
+        // Update worker channel is now clean.
+        // Reset the flag and check the plunger result.
+        update_handler
+            .skip_updates_flag
+            .store(false, std::sync::atomic::Ordering::Relaxed);
+
+        // Now the plunger result can be checked.
+        // If something went wrong, WAL won't be dropped.
+        plunger_result?;
```
timvisee: Since we skip updates with the flag above, this will always error because we drop the sender don't we?

agourlay:
I think we need a minimum level of automated testing on this PR.

I recommend to add a basic openapi Python integration tests that:
- shows everything is wired properly
- check the access requirement
- perform some basic validations (e.g. write after truncation)

lib/collection/src/shards/local_shard/wal_ops.rs
```diff
@@ -11,4 +16,72 @@ impl LocalShard {
...
+            // `applied_seq_num` is persisted with `APPLIED_SEQ_SAVE_INTERVAL` step.
+            // So WAL can be dropped from `applied_seq_num + APPLIED_SEQ_SAVE_INTERVAL`.
+            // Add 1 also because the last applied record must stay in the WAL.
+            let safe_drop_seq_num = applied_seq_num + APPLIED_SEQ_SAVE_INTERVAL + 1;
+
+            if safe_drop_seq_num < last_wal_op_num {
+                wal_lock.drop_from(safe_drop_seq_num)?;
+                wal_lock.flush()?;
+                (last_wal_op_num - safe_drop_seq_num) as usize
+            } else {
```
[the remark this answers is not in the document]
IvanPleshkov: Please check if fixed

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -46,6 +48,10 @@ impl UpdateWorkers {
...
+                    }
```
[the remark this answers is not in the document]
IvanPleshkov: Please check if fixed

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -1351,6 +1351,27 @@ impl ShardReplicaSet {
         let local = self.local.read().await;
         local.as_ref().and_then(|shard| shard.optimizations())
     }
+
+    /// Drop WAL from the last applied sequence number for all shards in the collection.
+    /// Returns amount of removed records.
+    pub async fn drop_wal(&self) -> CollectionResult<usize> {
```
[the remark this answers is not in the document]
IvanPleshkov: Please check if fixed

lib/collection/src/shards/local_shard/wal_ops.rs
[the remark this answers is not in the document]
IvanPleshkov: Thanks! Both indices are inclusive, and that's why we need to add one.
Added a comment to the code about it.

src/actix/api/service_api.rs
```diff
@@ -212,6 +215,29 @@ async fn update_logger_config(
...
+    ActixAccess(access): ActixAccess,
+) -> impl Responder {
+    let future = async move {
+        let pass = new_unchecked_verification_pass();
+        dispatcher
+            .toc(&access, &pass)
+            .drop_wal(&collection.name, access)
+            .await
+    };
+    helpers::time_or_accept(future, params.wait.unwrap_or(true)).await
```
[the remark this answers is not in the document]
IvanPleshkov: Why do you think that there is a plain‑text requirement?

lib/collection/src/update_workers/update_worker.rs
IvanPleshkov: Send an explicit feedback: `Operation {op_num} skipped due to ongoing WAL truncation`.
It's better than the sender dropping.

lib/collection/src/update_workers/update_worker.rs
IvanPleshkov: Thanks! Renamed

lib/shard/src/wal.rs
IvanPleshkov: It's a good assertion, added it

IvanPleshkov:
> I think we need a minimum level of automated testing on this PR.
> 
> I recommend to add a basic openapi Python integration tests that:
> 
> * shows everything is wired properly
> * check the access requirement
> * perform some basic validations (e.g. write after truncation)

Added a test, which tries to fill the WAL, truncate it, and check if we can apply update operations after

src/actix/api/service_api.rs
IvanPleshkov: added

lib/collection/src/shards/local_shard/wal_ops.rs
IvanPleshkov: You are right about deadlock! While testing, I didn't catch this, but it's obviously incorrect.
I changed the approach and now truncate function clears the channel first, and lock the WAL only when update worker channel is empty.
Updated the PR description with detailed explanation how it works now

lib/collection/src/shards/local_shard/wal_ops.rs
IvanPleshkov: `update_handler.applied_seq()` returns here the `Arc` to the handler of the applied sequence, not an actual `seq_num`. The actual `seq_num` is taken later, when the update worker channel is already empty.

lib/collection/src/shards/local_shard/wal_ops.rs
IvanPleshkov: It's a plunger result.
The skip flag is applied only for update operations. Plunger, nop and stop signals are working like before.
So if there are no stop signals in the update channel, the plunger will be finished without errors.

src/actix/api/service_api.rs
timvisee: Can we grab the pass via `_.issue_pass(...)` instead and completely remove `new_unchecked_verification_pass`?

We'd have to name `_` above it of course.

lib/collection/src/collection/state_management.rs
```diff
@@ -252,4 +254,31 @@ impl Collection {
...
+                results
+            })
+            .await?;
+
+        let mut total_dropped_records = 0;
+        for result in results {
+            total_dropped_records += result?;
+        }
+
+        Ok(total_dropped_records)
```
timvisee: We can simplify this to:

```suggestion
        results.into_iter().sum()
```

We can sum results and catch errors: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=9532b1459e45aada1422685665898904

lib/collection/src/shards/local_shard/wal_ops.rs
```diff
@@ -11,4 +38,50 @@ impl LocalShard {
...
+        let mut wal_lock = Mutex::lock_owned(self.wal.wal.clone()).await;
+        let last_wal_op_num = wal_lock.last_index();
+
+        let applied_seq = update_handler.applied_seq();
+        let applied_seq_num = applied_seq.op_num().unwrap_or(last_wal_op_num);
+
+        // `applied_seq_num` is persisted with `APPLIED_SEQ_SAVE_INTERVAL` step.
+        // So WAL can be dropped from `applied_seq_num + APPLIED_SEQ_SAVE_INTERVAL`.
+        // Add 1 also because the last applied record must stay in the WAL.
+        let safe_drop_seq_num = applied_seq_num + APPLIED_SEQ_SAVE_INTERVAL + 1;
```
timvisee: This is problematic.

As far as I can see we:
- drain the full channel
- drain the same **or less** entries from the WAL

If that is true, we must submit the difference of those two into the channel again. We currently don't, which means we'll be missing operations that remain persisted.

I again strongly prefer not to rely on `applied_seq` here:
- the `applied_seq` is an extra abstraction that hides details
- we can grab the correct `op_num` from the channel/WAL itself
- we don't have to handle `APPLIED_SEQ_SAVE_INTERVAL`
- we eliminate a potential future foot-gun of the applied seq not being exactly in sync

Let me give a quick attempt to implement this idea. 👍

IvanPleshkov: You are totally right about this problem.
I modified a test to reproduce it.
Checked on your idea - it works! 🔥

timvisee: Resolved by merging <https://github.com/qdrant/qdrant/pull/8030> into this PR

IvanPleshkov: PR with another approach of starting seq number is merged https://github.com/qdrant/qdrant/pull/8030

src/actix/api/service_api.rs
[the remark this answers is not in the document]
IvanPleshkov: I updated PR description, description was wrong

lib/collection/src/shards/local_shard/wal_ops.rs
IvanPleshkov: I resolve this question as obsolete, there is another starting point provided by @timvisee

lib/collection/src/shards/local_shard/wal_ops.rs
IvanPleshkov: I resolve this question as obsolete, there is another starting point provided by @timvisee

lib/collection/src/collection/state_management.rs
IvanPleshkov: It looks really better, applied

src/actix/api/service_api.rs
IvanPleshkov: It didn't help to remove `new_unchecked_verification_pass`. For instance, `get_partial_snapshot_manifest` uses both `issue_pass` and `new_unchecked_verification_pass`.
But I find this suggestion very helpful. I removed a ToC method and use `issue_pass` to call truncation directly from the collection.

PR #8002, opened 2026-01-27
more reliable version check

generall:


timvisee:
I didn't test this locally, but the change makes sense.

I'm betting on CI showing us this works :+1:

PR #8007, opened 2026-01-28
faster segment holder locks

generall:
This PR showcases an alternative approach for dealing with a lock of SegmentsHolder.

Previously, we used to lock SegmentsHolder for the whole duration of the read and write operations, while locking for write during swap of optimized segments.
This approach could cause the following:

- when we want to lock for write, we need to wait for updates and searches to finish
- searches and updates keep piling up
- Once optimization swap finishes, all updates and searches are allowed to proceed

This can cause tail latencies spikes and maybe even timeouts.

What new approach does:

- on search: we lock SegmentHolder just to copy list of segments, holder it immidiatelly released
- on updates: we introduce an extra mutex to prevent parallel updates during optimizer swap
- on optimizer: instead of write, we use upgradable read + write, minimizing critical section.

As a result, searches should be minimally affected by optimized segment swap.
Downside: we need to do extra copy of segment arcs, but it seems to be an unmeasurable difference.

UPD:

During implementation, I faced unexpected deadlock scenario, which led to further investigation of how LockerSegmentHolder is used. I changed it from type alias to dedicated type, so code navidation now becomes a little bit easier. Additionally I had to inspect and change locking pattern in the snapshot fucntion.

Overall locking schema how looks like this:

Screenshot from 2026-01-31 22-30-43

ToDo:

this PR only introduces new mechanism to one search request. Further PRs should apply it to all suitable read methods.

### DIFF
--- a/lib/shard/src/segment_holder/read_points.rs
+++ b/lib/shard/src/segment_holder/read_points.rs
@@ -39,7 +39,7 @@ impl SegmentHolder {
         Ok(read_points)
     }
 
-    fn segments_for_retrieval(&self) -> impl Iterator<Item = LockedSegment> + '_ {
+    fn segments_for_retrieval(&self) -> impl Iterator<Item = LockedSegment> {
         // We must go over non-appendable segments first, then go over appendable segments after
         // Points may be moved from non-appendable to appendable, because we don't lock all
         // segments together read ordering is very important here!


lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -795,15 +798,15 @@ pub trait SegmentOptimizer {
         if let Some(cow_segment_id) = cow_segment_id_opt {
             // Temp segment might be taken into another parallel optimization
             // so it is not necessary exist by this time
-            write_segments_guard.remove_segment_if_not_needed(cow_segment_id)?;
+            writable_segment_holder.remove_segment_if_not_needed(cow_segment_id)?;
```
timvisee: This function could take some time, making the critical (write locked) section longer.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -926,15 +926,17 @@ pub trait SegmentOptimizer {
...
+        // so all write operations should use `upgradable_read` as well.
+        let segment_holder_write = segment_holder.upgradable_read();
 
         let proxy_index_changes = self.proxy_index_changes(proxies);
 
         // Apply index changes before point deletions
         // Point deletions bump the segment version, can cause index changes to be ignored
         for (field_name, change) in proxy_index_changes.iter_ordered() {
-            // Warn: change version might be lower than the segment version,
+            // Warn: change version might be lower than the segment versiowriten,
```
timvisee:
```suggestion
            // Warn: change version might be lower than the segment version,
```

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
generall: It was previously a write section already

timvisee: I know. I'm just curious if it's something we can improve upon.

I believe we can take it out of the holder holding the mutex, and then finally clean it up outside of the segment holder lock.

lib/collection/src/shards/local_shard/snapshot.rs
```diff
@@ -362,8 +356,12 @@ where
     // Unproxy all segments
     // Always do this to prevent leaving proxy segments behind
     log::trace!("Unproxying all shard segments after function is applied");
-    let _update_guard = update_lock.blocking_write();
```
generall, line 365: I think this was incorrectly used. Logic inside `try_unproxy_segment` and `unproxy_all_segments` was usign 2-step propagation of updates, which assumes segment could change during the first stage, but this lock should have prevented it already on the first step. 

New logic changes it into just a single step propagation with updates guard as an explicit parameter, so function can't be misused.

lib/shard/src/segment_holder/snapshot.rs
```diff
@@ -113,20 +113,18 @@ impl SegmentHolder {
...
     /// shard holder.
-    pub fn try_unproxy_segment(
-        segments_lock: RwLockUpgradableReadGuard<SegmentHolder>,
+    pub fn try_unproxy_segment<'a>(
+        segments_lock: RwLockUpgradableReadGuard<'a, SegmentHolder>,
         segment_id: SegmentId,
         proxy_segment: LockedSegment,
-    ) -> Result<RwLockUpgradableReadGuard<SegmentHolder>, RwLockUpgradableReadGuard<SegmentHolder>>
-    {
+        updates_guard: parking_lot::MutexGuard<'a, ()>, // Guarantee no updates are happening
```
generall: this lock should guarantee there are no parallel updates happening

lib/shard/src/segment_holder/snapshot.rs
```diff
@@ -149,20 +147,11 @@ impl SegmentHolder {
...
-                log::error!(
-                    "Propagating proxy segment {segment_id} changes to wrapped segment failed, ignoring: {err}",
-                );
-            }
-            proxy_segment.wrapped_segment.clone()
-        };
+        let wrapped_segment = proxy_segment.read().wrapped_segment.clone();
         write_segments.replace(segment_id, wrapped_segment).unwrap();
 
+        drop(updates_guard); // Release updates lock as soon as possible
```
generall, line 154: section where reads are locked is very minimal now. Updates can hold in queue

lib/shard/src/update.rs
```diff
@@ -146,17 +126,17 @@ pub fn process_payload_operation(
             }
         }
         PayloadOps::ClearPayload { ref points, .. } => {
-            clear_payload(&segments.read(), op_num, points, hw_counter)
+            clear_payload(segments, op_num, points, hw_counter)
         }
```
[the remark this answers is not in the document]
generall: this is outside of scope for this PR

lib/shard/src/segment_holder/read_points.rs
```diff
@@ -0,0 +1,96 @@
...
+                .collect();
+            check_stopped(is_stopped)?;
+            if !segment_point_ids.is_empty() {
+                read_points += f(&segment_point_ids, &read_segment)?;
+            }
+        }
+        Ok(read_points)
+    }
+
+    fn segments_for_retrieval(&self) -> impl Iterator<Item = LockedSegment> + '_ {
```
JojiiOfficial: Nit: not needed
```suggestion
    fn segments_for_retrieval(&self) -> impl Iterator<Item = LockedSegment> {
```

lib/shard/src/segment_holder/snapshot.rs
monoid: Making it a newtype could make it clearer.

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -926,10 +910,28 @@ pub trait SegmentOptimizer {
...
+        proxy_ids: &[SegmentId],
+        cow_segment_id_opt: Option<SegmentId>,
+        stopped: &AtomicBool,
+        hw_counter: &HardwareCounterCell,
+    ) -> CollectionResult<usize> {
+        // This block locks all write operations with colle ction. It should be fast.
+        let upgradable_segment_holder = segment_holder.upgradable_read();
+
+        // This mutex prevents update operations, which could create inconsistency during transition.
+        let update_guard = segment_holder.acquire_updates_lock();
```
monoid, line 932: The `update` takes calls the `segment_holder.acquire_updates_lock()` before the `segment_holder.read()` (calling it `segments` though).  It is a way for deadlock, isn't it?

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
```diff
@@ -926,10 +910,28 @@ pub trait SegmentOptimizer {
...
+        segment_holder: &LockedSegmentHolder,
+        locked_proxies: Vec<LockedSegment>,
+        mut optimized_segment: Segment,
+        already_remove_points: &DeletedPoints,
+        proxy_ids: &[SegmentId],
+        cow_segment_id_opt: Option<SegmentId>,
+        stopped: &AtomicBool,
+        hw_counter: &HardwareCounterCell,
+    ) -> CollectionResult<usize> {
+        // This block locks all write operations with colle ction. It should be fast.
```
monoid:
```suggestion
        // This block locks all write operations with collection. It should be fast.
```

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
generall: could you please elaborate a dealock sequence here?

lib/shard/src/segment_holder/snapshot.rs
generall: new type for guard? you mean like this `pub type = ...` ?

monoid: I mean a wrapper type: https://doc.rust-lang.org/rust-by-example/generics/new_types.html

`pub type` doesn't create a new type, it creates merely an alias.

generall: @copilot suggest a type wrapper for `updates_guard`

lib/collection/src/collection_manager/optimizers/segment_optimizer.rs
monoid: I'm wrong, sorry.

PR #8008, opened 2026-01-28
WAL replay honors applied_seq

agourlay:
```
INFO storage::content_manager::toc: Loading collection: benchmark
DEBUG collection::optimizers_builder: Removing temp_segments directory: "./storage/collections/benchmark/0/temp_segments"
DEBUG collection::shards::local_shard: Recovering shard ./storage/collections/benchmark/0 starting reading WAL from 267948 up to 274626 (last_applied_seq:Some(274561))
Recovering collection benchmark [00:00:00] █████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████ 6678/6678 (eta:0s)
INFO collection::shards::local_shard: Loading remaining 774673 WAL entries from:274626 into update queue
```

## How to test

Run Qdrant with `QDRANT__STORAGE__UPDATE_QUEUE_SIZE=1000000`

```http
PUT collections/benchmark
{
  "vectors": {
    "size": 256,
    "distance": "Dot"
  },
  "optimizers_config": {
    "prevent_unoptimized": true
  }
}
```

```bash
./bfb -n 20M -d 256 -b 1 --skip-create
```

Wait a bit so it results in:
- a large amount of pending updates in the WAL
- a large amount of pending updates in the update channel
- no timeouts or slowdown should be observed by `bfb`

Just restart to trigger the WAL replay which should be very fast by leveraging the last `applied_seq`.

### DIFF
--- a/lib/collection/src/tests/wal_recovery_test.rs
+++ b/lib/collection/src/tests/wal_recovery_test.rs
@@ -5,7 +5,6 @@ use common::counter::hardware_accumulator::HwMeasurementAcc;
 use common::save_on_disk::SaveOnDisk;
 use segment::data_types::vectors::VectorStructInternal;
 use segment::types::{PayloadFieldSchema, PayloadSchemaType, WithPayload, WithVector};
-use segment::types::{PayloadFieldSchema, PayloadSchemaType};
 use shard::operations::CollectionUpdateOperations;
 use shard::operations::point_ops::{
     PointInsertOperationsInternal, PointOperations, PointStructPersisted,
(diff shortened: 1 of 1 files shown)


IvanPleshkov:
Added a test for this feature.
The test performs uploads into the shard, drops it, reload and checks that the update queue is not empty.
There is a trick inside. Test changes persistent seq num manually if it's too big (which reproduces each time in test).

generall (approved):

IvanPleshkov (approved):

PR #8010, opened 2026-01-29
Update queue info into CollectionInfo

agourlay:
In order to increase the visibility in the update process, this PR adds information about the update queue into the collection info.

e.g

info-q

I did add a integration test for the queue size as it is brittle due to the time dependence. 

Feel free to bikeshed naming.

### DIFF
--- a/tests/openapi/test_update_queue.py
+++ b/tests/openapi/test_update_queue.py
@@ -2,6 +2,7 @@ import pytest
 import requests
 
 from openapi.helpers.helpers import (
+    skip_if_no_feature,
     get_api_string,
     request_with_validation,
 )
@@ -93,6 +94,8 @@ def test_queue_op_num(collection_name):
 
 
 def test_queue_length(collection_name):
+    skip_if_no_feature("staging")
+    
     queue_info = get_queue_info(collection_name)
 
     # empty collection
@@ -125,7 +128,7 @@ def test_queue_length(collection_name):
 
     queue_info = get_queue_info(collection_name)
 
-    # wait=true so ack. after application
+    # wait=false so updates are enqueued but not yet applied
     assert queue_info["length"] == 1
     assert queue_info["op_num"] == 0
 
@@ -147,6 +150,6 @@ def test_queue_length(collection_name):
     assert response.ok
 
     queue_info = get_queue_info(collection_name)
-    # wait=true so ack. after application
+    # wait=false so updates are enqueued but not yet applied
     assert queue_info["length"] == 2
     assert queue_info["op_num"] == 0
--- a/tests/openapi/helpers/helpers.py
+++ b/tests/openapi/helpers/helpers.py
@@ -1,6 +1,6 @@
+import pytest
 import json
 import time
-from logging import warning
 from typing import Any, Dict, List
 import jsonschema
 import requests
@@ -178,3 +178,20 @@ def distribution_based_score_fusion(responses: List[List[Any]], limit: int = 10)
 def qdrant_host_headers():
     headers = json.loads(QDRANT_HOST_HEADERS)
     return headers
+
+
+def check_feature_enabled(feature) -> bool:
+    response = request_with_validation(
+        api='/telemetry',
+        method="GET",
+        query_params={'details_level': 10},
+    )
+    assert response.ok
+    features = response.json()['result']['app']['features']
+    result = features[feature]
+    return result
+
+def skip_if_no_feature(feature):
+    feature_is_enabled = check_feature_enabled(feature)
+    if not feature_is_enabled:
+        pytest.skip(f"Skipping because the feature {feature} is disabled at runtime.")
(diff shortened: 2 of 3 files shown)


timvisee:
Nice!

What is `last_applied_seq` for? Is it so we can derive the application care from it, and possibly also compare it with point versions exposed in some reads?

agourlay:
> What is last_applied_seq for? Is it so we can derive the application care from it, and possibly also compare it with point versions exposed in some reads?

I see it mostly as a monitoring/debugging tool to track the progress on the update commit stream.
Could be compared across replicas etc.

Do you have maybe concern about it? :)

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -883,6 +883,15 @@ impl LocalShard {
...
     }
 
+    pub async fn local_update_queue_info(&self) -> UpdateQueueInfo {
+        let last_applied_seq = self.applied_seq_handler.op_num().map(|s| s as usize);
+        let length = self.update_queue_length();
+        UpdateQueueInfo {
+            length,
+            last_applied_seq,
+        }
+    }
```
xzfc: Nit: remove async, make compact.

```suggestion
    pub fn local_update_queue_info(&self) -> UpdateQueueInfo {
        UpdateQueueInfo {
            length: self.update_queue_length(),
            last_applied_seq: self.applied_seq_handler.op_num().map(|s| s as usize),
        }
    }
```

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -70,14 +70,8 @@ impl ShardOperation for LocalShard {
         }
 
         let operation_id = {
+            let pending_operations_count = self.update_queue_length();
             let update_sender = self.update_sender.load();
```
xzfc, line 74: Now `update_sender` loads twice here. Can't it lead to race conditions? (`update_sender` is overwritten in `on_optimizer_config_update`)

(personally, I doubt it tho)

lib/collection/src/shards/local_shard/mod.rs
agourlay: Good catch!

lib/collection/src/shards/local_shard/shard_ops.rs
agourlay: Correct, I did notice as well and decided that it should be fine for the sake of keeping the code easy to read.

IvanPleshkov: Link a doc for further readers
https://docs.rs/arc-swap/latest/arc_swap/struct.ArcSwapAny.html#method.load

lib/api/src/grpc/proto/collections.proto
```diff
@@ -748,6 +748,13 @@ message PayloadSchemaInfo {
   optional uint64 points = 3;
 }
 
+message UpdateQueueInfo {
+    // Number of elements in the queue
+    uint64 length = 1;
+    // last operation sequence number processed
+    optional uint64 last_applied_seq = 2;
```
IvanPleshkov: Nit. I propose to rename this field. Because it's not exactly the last applied seq. For instance, if the user performs updates and restart qdrant, the loaded value will be less than the real last applied seq num.
Not sure which name fits better. maybe `last_relaxed_applied_seq`.
Because I dont have something better, I wont block merge because of this thread

generall: I would ask to use field names with less than 3 words please

agourlay: I will rename the field to simply `op_num`

agourlay:
I'd like to merge this one very soon to leverage the new info fields to test the partial WAL restart :)

lib/collection/src/operations/types.rs
```diff
@@ -205,6 +213,8 @@ pub struct CollectionInfo {
     pub config: CollectionConfig,
     /// Types of stored payload
     pub payload_schema: HashMap<PayloadKeyType, PayloadIndexInfo>,
+    /// Update queue info
+    pub update_queue: Option<UpdateQueueInfo>,
```
timvisee, line 218: Is this needed?

```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub update_queue: Option<UpdateQueueInfo>,
```

tests/openapi/test_update_queue.py
```diff
@@ -0,0 +1,94 @@
...
+                    "payload": {}
+                }
+            ]}        
+    )
+    assert response.ok
+    
+    queue_info = get_queue_info(collection_name)
+    # wait=true so ack. after application
+    assert queue_info["length"] == 0
+    assert queue_info["op_num"] == 2
```
timvisee: Could we also test `length` here, or is that hard to achieve?

We've implemented `staging` operations which you can use to send an update that just blocks for a given time period. If you'd send an update with a 5 second delay, and 4 quick updates. I'd expect `length` to show 4 or 5.

Requiring the `staging` feature flag here might be a problem though.

lib/collection/src/operations/types.rs
agourlay: I don't think it is necessary per se but decided to add it just in case I am missing something.

PR #8011, opened 2026-01-29
Fix gridstore Option unsoundness

monoid:
A value of type `Option<ValuePointer`> was stored into a mmap-ed memory.  While `ValuePointer` is a `repr(C)` POD type, `Option<ValuePointer>`'s layout and bit patterns is not specified.

The commit contains a quick and dirty fix adding an artificial `_discriminant: NonZeroU8` field creating a possibility for a niche `Option` optimization.

A similar unsoundness is using a tuple for storing values.  The tuple was replace with a `repr(C)` tuple.

Also, some `transumute_*` functions were safe by mistake; they were made unsafe.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/gridstore/src/tracker.rs
+++ b/lib/gridstore/src/tracker.rs
@@ -19,11 +19,8 @@ pub type PageId = u32;
 
 const TRACKER_MEM_ADVICE: Advice = Advice::Random;
 
-const OPTIONAL_NONE: u32 = 0;
-const OPTIONAL_SOME: u32 = 1;
-
-/// A type similar to `std::option::Option`, but with stable layout. It is intended to be compatible with older
-/// gridstore files, but it is well-defined, unlike `std::option::Option`.
+/// A type similar to [`std::option::Option`], but with stable layout. It is intended to be compatible with older
+/// gridstore files, but it is well-defined, unlike [`std::option::Option`].
 ///
 /// Please note that it uses 32-bit tag and is intended to be used for `ValuePointer` without padding bytes.
 ///
@@ -45,10 +42,13 @@ impl<T: FromZeros> From<Option<T>> for Optional<T> {
 }
 
 impl<T: FromZeros> Optional<T> {
+    const OPTIONAL_NONE: u32 = 0;
+    const OPTIONAL_SOME: u32 = 1;
+
     /// None value is all zeroes.
     pub fn none() -> Self {
         Self {
-            discriminant: OPTIONAL_NONE,
+            discriminant: Self::OPTIONAL_NONE,
             value: FromZeros::new_zeroed(),
         }
     }
@@ -56,13 +56,13 @@ impl<T: FromZeros> Optional<T> {
     /// Some is 1 for the discriminant, and value is stored as is.
     pub const fn some(value: T) -> Self {
         Self {
-            discriminant: OPTIONAL_SOME,
+            discriminant: Self::OPTIONAL_SOME,
             value,
         }
     }
 
     pub fn is_some(&self) -> Option<&T> {
-        if self.discriminant == OPTIONAL_NONE {
+        if self.discriminant == Self::OPTIONAL_NONE {
             None
         } else {
             Some(&self.value)


lib/common/memory/src/mmap_ops.rs
```diff
@@ -118,7 +118,11 @@ pub fn open_write_mmap(path: &Path, advice: AdviceSetting, populate: bool) -> io
 
     Ok(mmap)
 }
-pub fn transmute_from_u8<T>(v: &[u8]) -> &T {
+
+/// # Safety
+///
+/// `data` must have correct alignment and size for `T` and contain correct bit patterns for the type `T`.
+pub unsafe fn transmute_from_u8<T>(v: &[u8]) -> &T {
```
xzfc, line 126: I'd mark these `transumte_*` functions as deprecated. The new code should use `zerocopy` or `bytemuck` instead.

See also: https://github.com/qdrant/qdrant/pull/5297/changes/BASE..be387be795977cfe99fabe77694ba31c66226f32

monoid: Makes sense!

But it seems neither `zerocopy` nor `bytemuck` have methods for transforming `&[T]` to `&[u8]` (their slice methods work elementwise).

xzfc: AFAIR both can transmute slices.

Bytemuck has separate methods (`cast_slice`), and `zerocopy` uses the same method both for slices and non-slices.

monoid: Indeed, bytemuck requires `input.len() * size_of::<A>() == output.len() * size_of::<B>()`.

monoid: Unfortunately, deprecating them makes CI to fail, and `#![allow(deprecate)]` makes the deprecation useless.

xzfc: > `#![allow(deprecate)]` makes the deprecation useless

Mark usage sites only (since you already touch all of them in this PR).

```diff
 // Safey: T implements zerocopy::IntoBytes.
+#[expect(deprecated, reason = "old code")]
 let raw_bites = unsafe { mmap_ops::transmute_to_u8_slice(vector.as_ref()) };
```

That way we are less likely to add new uses by accident.

xzfc: Why I don't like these functions: they are footguns.

Try to run this in miri (Tools -> Miri in the playground) or valgrind (build it locally). Both will scream as reading unitialized/padding bytes is UB.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=62af298cf9ddb95bc5f0923c406b3c87

Or this compiles without errors/warnings but segfaults. Spot the error.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cad2674dd843817262e79b6f46f8804a

`zerocopy` and `bytemuck` would throw a compile error in these cases.

lib/gridstore/src/tracker.rs
```diff
@@ -20,6 +22,10 @@ const TRACKER_MEM_ADVICE: Advice = Advice::Random;
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 #[repr(C)]
 pub struct ValuePointer {
+    /// This is an artificial field to make `Option<ValuePointer>` to have a stable ABI.
+    pub _discriminant: NonZeroU8,
+    pub _padding: [u8; 3],
```
xzfc: Note: this change made `ValuePointer` a bit larger (12 -> 16 bytes). I wonder, can it affect performance/runtime memory usage of Gridstore?

Next-token-predictor said not much, tho.

<details><summary>prompt used</summary>

@lib/gridstore/src/tracker.rs

We added `_discriminant` and `_padding` fields to `ValuePointer` to "legalize" layout of `Option<ValuePointer>`.
On-disk representation of `Option<ValuePointer>` didn't changed, so from correctness point of view it's good.

But the question from the performance point of view.
Now `ValuePointer` a bit larger (12->16 bytes). Can it affect the performance and RAM usage of Gridstore?
Check whether it is used in hot loops or large arrays of data.

</details>

---

A bit cleaner alternative option is to "reimplement" option enum like this instead:
```rust
struct ValuePointerOption {
    is_some: u32, // 0 or 1
    value_pointer: ValuePointer,
}
```
Tho the current version of this PR is still good, I think.

monoid: That's should be OK because `ValuePointer` is used mostly either by reference or as `Option<ValuePointer>` that has the same size and layout.

But it worth benchmarking just to make sure.

Yes, we can implement custom `Option`, but it would be more intrusive (will require changing function signatures).

generall: >  this change made ValuePointer a bit larger

why it doesn't affect backward compatibility?

xzfc: > why it doesn't affect backward compatibility?

We do not store `ValuePointer`. We store `Option<ValuePointer>`, and it's remain 16 as before.
There is a test in this PR that checks compatibility. (`test_option_value_pointer_layout`)

monoid: I think such a hack should be better documented.

timvisee: I do like @xzfc's suggestion more, if we can use it. It seems like there's less implicit compiler magic involved.

The boolean `is_some` state could be an u32 enum.

lib/sparse/src/index/inverted_index/inverted_index_compressed_mmap.rs
```diff
@@ -221,24 +229,37 @@ impl<W: Weight> InvertedIndexCompressedMmap<W> {
...
         Some(CompressedPostingListView::new(
-            self.slice_part(header.ids_start, header.ids_len),
-            self.slice_part(
-                header.ids_start + u64::from(header.ids_len),
-                header.chunks_count,
-            ),
-            transmute_from_u8_to_slice(
-                &self.mmap[remainders_start as usize..remainders_end as usize],
-            ),
+            // TODO Safety
```
timvisee, line 232: Nit: shall we just wrap the 3 consequent calls in one unsafe block?

monoid: They are function arguments.  We can wrap the whole expression with the unsafe, but I don't like the safe part being here too.

monoid: Unsafe must sore your eyes and motivate you to use a safe approach!

lib/gridstore/src/tracker.rs
monoid: Well, it seems that we also collect bare `ValuePointer`s into vectors, and it may affect the performance and the memory consumption.

I thinks more conservative approach should be used: `u8` with padding (it would work even if Rust used whole `u32` for the discriminant, but only on little-endian arches).

xzfc: I vote to use `u32` to avoid padding bytes in the structure, so it would match [`bytemuck::Pod`](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html) and [`zerocopy::FromBytes`](https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html#warning-padding-bytes) requirements.

monoid: I believe we will get it with explicit padding bytes too.

lib/common/memory/src/mmap_ops.rs
monoid: > Both will scream as reading unitialized/padding bytes is UB.

BTW, I think the mmaped memory must to be `[MaybeUninit<u8>]`.

lib/gridstore/src/tracker.rs
```diff
@@ -17,7 +19,58 @@ pub type PageId = u32;
...
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+const OPTIONAL_NONE: u32 = 0;
+const OPTIONAL_SOME: u32 = 1;
+
+/// A type similar to `std::option::Option`, but with stable layout. It is intended to be compatible with older
+/// gridstore files, but it is well-defined, unlike `std::option::Option`.
+///
+/// Please note that it uses 32-bit tag and is intended to be used for `ValuePointer` without padding bytes.
+///
```
xzfc: Nit: use this syntax so the rust-analyzer will make these links clickable.

```rust
/// A type similar to [`Option`], …
```

lib/gridstore/src/tracker.rs
```diff
@@ -17,7 +19,58 @@ pub type PageId = u32;
...
+/// If `T` is a POD type, then `Optional<T>` is a POD type.
+#[derive(Copy, Clone, zerocopy::FromBytes)]
+#[repr(C)]
+struct Optional<T> {
+    discriminant: u32,
+    value: T,
+}
```
xzfc, line 33: I think the generic will make it harder to switch from `transmute_*` functions to `zerocopy`/`bytemuck`.
Reason: some `zerocopy` derive macros want to know the alignment of `T`, but they don't have access to the type information. Haven't checked for `bytemuck` tho.

Don't want to add to the pile of this PR, so perhaps we can keep it as is and postpone changes for the later. <sup><sub>(never)</sub></sup>

monoid: The type is mod-level, not even crate level, changing it is easy.

lib/gridstore/src/tracker.rs
```diff
@@ -17,7 +19,58 @@ pub type PageId = u32;
 
 const TRACKER_MEM_ADVICE: Advice = Advice::Random;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+const OPTIONAL_NONE: u32 = 0;
+const OPTIONAL_SOME: u32 = 1;
```
xzfc: Style nit: we can put these inside `Optional` impl:

```rust
impl<T: FromZeros> Optional<T> {
    const OPTIONAL_NONE: u32 = 0;
    const OPTIONAL_SOME: u32 = 1;

    /// None value is all zeroes.
    pub fn none() -> Self {
        Self {
            discriminant: Self::OPTIONAL_NONE,
```

PR #8012, opened 2026-01-29
[trififo] seqlock with interior mutability

coszio:
Implements a seqlock which has the ability to have interior access to the protected resource. 

It also documents why it is not safe to use it if the protected resource can change allocations.

### DIFF
--- a/lib/trififo/src/seqlock.rs
+++ b/lib/trififo/src/seqlock.rs
@@ -128,14 +128,14 @@ impl<T: SeqLockSafe> SeqLock<T> {
 /// - busy-wait if a write is taking place.
 /// - or repeat the read if a write took place in between the start and end
 ///   of the read.
-pub struct SeqLockReader<T> {
+pub struct SeqLockReader<T: SeqLockSafe> {
     lock: Arc<SeqLock<T>>,
 }
 
-unsafe impl<T: Send> Send for SeqLockReader<T> {}
-unsafe impl<T: Sync> Sync for SeqLockReader<T> {}
+unsafe impl<T: Send + SeqLockSafe> Send for SeqLockReader<T> {}
+unsafe impl<T: Sync + SeqLockSafe> Sync for SeqLockReader<T> {}
 
-impl<T> Clone for SeqLockReader<T> {
+impl<T: SeqLockSafe> Clone for SeqLockReader<T> {
     fn clone(&self) -> Self {
         SeqLockReader {
             lock: Arc::clone(&self.lock),
@@ -284,30 +284,35 @@ mod tests {
     }
 
     fn assert_correct_send_sync() {
-        use std::cell::Cell;
         use std::rc::Rc;
         use std::sync::Mutex;
 
         use static_assertions::{assert_impl_all, assert_not_impl_any};
 
+        // ==============
+        // SeqLockWriter
+        // ==============
+
         // Writer is Send but not Sync
-        assert_impl_all!(SeqLockWriter<u32>: Send);
-        assert_not_impl_any!(SeqLockWriter<u32>: Sync);
-        // it works only if we wrap it in a sync abstraction.
-        assert_impl_all!(Mutex<SeqLockWriter<u32>>: Sync);
-        // if the wrapped type is not Send, Send doesn't work.
-        assert_not_impl_any!(SeqLockWriter<Rc<u32>>: Send);
-
-        // Reader is Send AND Sync
-        assert_impl_all!(SeqLockReader<u32>: Sync, Send);
-
-        // but only if the wrapped type is also Sync/Send
         //
-        // Cell is Send, but not Sync
-        assert_impl_all!(SeqLockReader<Cell<u32>>: Send);
-        assert_not_impl_any!(SeqLockReader<Cell<u32>>: Sync);
-        // Rc is neither
-        assert_not_impl_any!(SeqLockReader<Rc<u32>>: Sync, Send);
-        assert_not_impl_any!(SeqLockReader<Arc<Cell<u32>>>: Sync, Send);
+        // The way we ensure SeqLockWriter can't write concurrently
+        // is by making it `!Sync`
+        assert_not_impl_any!(SeqLockWriter<usize>: Sync);
+
+        // However, it can be shared between threads if it is wrapped
+        // in a `Sync` abstraction.
+        assert_impl_all!(SeqLockWriter<usize>: Send);
+        assert_impl_all!(Mutex<SeqLockWriter<usize>>: Send, Sync);
+
+        // if the wrapped type is not Send, Send doesn't work.
+        assert_not_impl_any!(SeqLockWriter<Rc<usize>>: Send);
+
+        // ==============
+        // SeqLockReader
+        // ==============
+
+        // SeqLockReader is Send AND Sync, and T needs to be SeqLockSafe
+        assert_impl_all!(usize: SeqLockSafe);
+        assert_impl_all!(SeqLockReader<usize>: Sync, Send);
     }
 }


lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,261 @@
...
+        loop {
+            let seq1 = self.seq.load(Ordering::Acquire);
+
+            // if odd, it is locked, wait until unlocked.
+            if seq1 & 1 == 1 {
+                std::hint::spin_loop();
+                continue;
+            }
+
+            let result = callback(unsafe { &*self.inner.get() });
```
xzfc, line 113: (about rust safety conventions)

This is extremely fragile.

Public methods `read` and `write` should be marked as `unsafe`. Or, alternatively, the callbacks should accept `*T`/`*mut T` instead of `&T`/`&mut T` to signify that these are not innocent references.

Either way, all callers should have `unsafe {}` block somewhere to be able to use these functions.

xzfc, line 113: (about correctness)

I'm not sure that this work well on platforms with weak memory ordering (e.g. non-Apple ARM).

Related: https://www.github.com/Amanieu/seqlock/issues/9. Maintainer of the upstream `seqlock` described this approach as "The code actually works, but it's a bit subtle." and "It's a bit of a gray area in the language, but it works in practice". But our implementation is more complicated (arbitrary callback vs just a `read_volatile`) and I'm not experienced with memory ordering to say whether it's safe.

lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// let (_reader, writer) = SeqLock::new_reader_writer(Pair { a: 0, b: 0 });
+    /// let shared = Arc::new(writer);
+    /// let shared_clone = shared.clone();
+    ///
+    /// // Moving `shared_clone` into a new thread requires `Arc<SeqLockWriter<_>>` to be Send,
+    /// // which in turn requires `SeqLockWriter<_>` to be Sync. This should fail to compile.
+    /// thread::spawn(move || {
+    ///     shared_clone.write(|p| { p.a = 1; p.b = 1; });
+    /// }).join().unwrap();
+    /// ```
```
xzfc: This test doesn't run.
- Doctests are not collected for items under `#[cfg(test)]`/`#[test]`.
- On CI we use `nextest` which doesn't support doctests.
- And even if we fix it, we can't be sure that it fails for the right reason. (see the bot comment above)

I suggest not to use doctests and use `static_assertions::assert_not_impl_any!` instead to prove it's not `Sync`/`Send`.

Or, at least, make them far shorter and follow https://rust-unofficial.github.io/too-many-lists/sixth-send-sync.html, and manually verify that they fail for the right reason by running `cargo test -p trififo --doc -- --no-capture`.

lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,261 @@
...
+        (reader, writer)
+    }
+
+    fn read<U>(&self, callback: impl Fn(&T) -> U) -> U {
+        loop {
+            let seq1 = self.seq.load(Ordering::Acquire);
+
+            // if odd, it is locked, wait until unlocked.
+            if seq1 & 1 == 1 {
+                std::hint::spin_loop();
```
xzfc, line 109: Why spin loop, and not a mutex?
For short waits, parking lot mutex spins for a few iteration before parking the thread.
https://docs.rs/parking_lot/latest/src/parking_lot/raw_mutex.rs.html#228-232

coszio: I mean, the whole point of this seqlock is to avoid using a mutex entirely. I just tried the mutex and it impacts benchmarks negatively

lib/trififo/src/seqlock.rs
coszio: I think we can avoid marking `read()` as `unsafe`, as long as we mark `write()` as `unsafe`. Accessing the value should be safe as long as the writer does not change the protected resource in a way that may cause readers to panic.

coszio: what about fencing before and after?

lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,261 @@
...
+    fn write(&self, callback: impl FnOnce(&mut T)) {
+        let seq = self.seq.load(Ordering::Acquire);
+        self.seq.store(seq + 1, Ordering::Release);
+
+        fence(Ordering::Release);
+
+        callback(unsafe { &mut *self.inner.get() });
+
+        self.seq.store(seq + 2, Ordering::Relaxed);
+    }
```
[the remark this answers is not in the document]
coszio: Thanks, I have corrected this in https://github.com/qdrant/qdrant/pull/8012/commits/cf7d39ddd51ef01ad0ef8f581703c347987870f3

Can you give it another look at the `read()` counterpart? is it possible that the read gets reordered in a way that it happens in a different order?

lib/trififo/src/seqlock.rs
xzfc: My concern is `read`'s callback can still experience torn reads, so `unsafe` can be used to draw attention to these places too. It can cause problems even if it's not related to allocations.

lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,312 @@
...
+/// The [`SeqLock`] struct itself is not Send, nor Sync, so it is kind of useless on its own.
+///
+/// To allow multiple readers, and ensure a single writer, a `new_reader_writer` method is provided.
+///
+/// # SAFETY
+/// While this lock ensures that a read is consistent, it does not protect against
+/// writers creating use-after-free errors. This is because readers can see the
+/// shared resource even as it is being modified. It is the user responsibility to make
+/// sure the underlying resource does not change allocations, and that the worst that
+/// can happen is a garbage/torn read.
```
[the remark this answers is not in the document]
coszio: I have improved the comment in https://github.com/qdrant/qdrant/pull/8012/commits/824dbaa49553a5eac91ed4289bcf84130915a94b

The point of this structure is that it is fine if readers get garbage data, as long as the UB introduced by reading and modifying at the same time does not cause a panic.

lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,302 @@
...
+///     std::thread::spawn(move || {
+///         unsafe {
+///             writer.write(|value| {
+///                 *value += 10;
+///             })
+///         };
+///     });
+/// }
+/// ```
+pub struct SeqLock<T> {
```
timvisee, line 53: We can add a new unsafe marker trait that is empty. We can use it to enforce all `T`s have to opt-in to using it. Maybe that would make us developers more aware of what types can be used with a structure like this. And it prevents us from dumping random types in here in the future if we don't read our own documentation well enough.

Like this:

```rust
unsafe impl MyTypeIsSafe for usize {}
```

coszio: This makes it much clearer, I like it. Implemented in https://github.com/qdrant/qdrant/pull/8012/commits/a26a78a1cb919ffc09bfb5c33036f795bf84fe77

lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,315 @@
...
+/// However, it is fine to let readers return garbage/torn reads, as it is the seqlock responsibility
+/// to ensure this garbage is never considered valid; in such case the read is repeated.
+///
+/// The type must only expose functions/fields that satisfy these requirements. Anything that takes
+/// `&self` is considered a read, and `&mut self` is considered a write.
+pub unsafe trait SeqLockSafe {}
+
+// Blanket implementation for all types that are Send and Sync.
+// It is trivial to know the type is also SeqLockSafe if it is already Sync.
+unsafe impl<T: Send + Sync> SeqLockSafe for T {}
```
timvisee: Are you sure about this blanked impl? Can't we still have issues with `Sync` types?

lib/trififo/src/seqlock.rs
timvisee: With the addition of

```rust
pub unsafe trait SeqLockSafe {}
```

doesn't this problem go away?

With the trait we define it's fine for the type to experience a torn read. We will not return torn data.

lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,315 @@
...
+            let seq2 = self.seq.load(Ordering::Relaxed);
+
+            // only return if seq did not change.
+            if seq1 == seq2 {
+                return result;
+            }
+        }
+    }
+
+    fn write(&self, callback: impl FnOnce(&mut T)) {
```
timvisee: This function has no protection against concurrent writes.

In my opinion we should either:
- make it `&mut` (which we don't like)
- panic if the seq we read is odd

coszio: There is protection in place, please correct me if it is not enough:
1. `SeqLock` is private to the module
2. The only constructor creates a `SeqLockReader` and a `SeqLockWriter`
3. The writer is not `Sync`, so even if there can be multiple active references, it is not possible that they are on different threads. Even if you put it behind an `Arc`, you won't be able to send it to other threads due to the blanket implementation of `Send` requiring `T` to be `Send + Sync`

lib/trififo/src/seqlock.rs
coszio: Wait, good point, actually it is not sound to assume this. There are many types which implement `Send + Sync` that are not safe to be read during mutation, for example `Vec` which can cause dangling pointers mid-reallocation.

lib/trififo/src/seqlock.rs
coszio: Thinking about this made me realize we need to make `SeqLockWriter::write` take `&mut self`. Otherwise it is possible to capture the seqlock within the callback and write again, which would break everything.

timvisee: Minor confusion from my side we cleared up during a call: readers have a separate type. And we can always use `&mut self` on the writer side without locking anything.

lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,313 @@
...
```
timvisee: We still have `&self` here and want to change that into `&mut self` am I right?

lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,313 @@
...
+    fn clone(&self) -> Self {
+        SeqLockReader {
+            lock: Arc::clone(&self.lock),
+        }
+    }
+}
+
+impl<T: SeqLockSafe> SeqLockReader<T> {
+    pub fn read<U, F: Fn(&T) -> U>(&self, callback: F) -> U {
+        self.lock.read(callback)
```
xzfc, line 148: (meet recap)
inline/move `SeqLock::read()` → `SeqLockReader::read()`, and the same for `write()`

lib/trififo/src/seqlock.rs
```diff
@@ -0,0 +1,313 @@
...
+        assert_impl_all!(SeqLockReader<u32>: Sync, Send);
+
+        // but only if the wrapped type is also Sync/Send
+        //
+        // Cell is Send, but not Sync
+        assert_impl_all!(SeqLockReader<Cell<u32>>: Send);
+        assert_not_impl_any!(SeqLockReader<Cell<u32>>: Sync);
+        // Rc is neither
+        assert_not_impl_any!(SeqLockReader<Rc<u32>>: Sync, Send);
+        assert_not_impl_any!(SeqLockReader<Arc<Cell<u32>>>: Sync, Send);
```
xzfc: (meet recap)
Add a high level comment explaining these.

---

Context:

For an external viewer it looks like you check a particular set of properties, but it not explained *why* did you choose *this particular set*, and not any other.

Imagine: if in N months, I'll edit this module, and my edits will make this test to fail. I won't be sure whether I need to fix my edits, or the test (and how).

lib/trififo/src/seqlock.rs
xzfc: The public method is `&mut self`.
The internal method have to be `&self` (because it's in `Arc<>` that shared between N readers and 1 writer).

And probably, we'll get rid of internal methods anyway.
(https://github.com/qdrant/qdrant/pull/8012/changes#r2770204420)

lib/trififo/src/seqlock.rs
coszio: After thinking about it, read implementation can be done by both reader/writer, but write implementation should only be in the writer

PR #8018, opened 2026-01-30
out of disk error variant

IvanPleshkov:
To implement operation retry in the OOD case, we need to first detect if the error is an OOD.

This PR adds an `OutOfDisk` error variant for the `CollectionError` and `Storage` error.

### DIFF
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -1279,7 +1279,9 @@ impl From<JsonError> for CollectionError {
 
 impl From<std::io::Error> for CollectionError {
     fn from(err: std::io::Error) -> Self {
-        if err.kind() == std::io::ErrorKind::StorageFull {
+        if err.kind() == std::io::ErrorKind::StorageFull
+            || err.kind() == std::io::ErrorKind::FileTooLarge
+        {
             CollectionError::OutOfDisk {
                 description: format!("File IO error: {err}"),
             }
@@ -1432,7 +1434,9 @@ impl From<cancel::Error> for CollectionError {
 
 impl From<tempfile::PathPersistError> for CollectionError {
     fn from(err: tempfile::PathPersistError) -> Self {
-        if err.error.kind() == std::io::ErrorKind::StorageFull {
+        if err.error.kind() == std::io::ErrorKind::StorageFull
+            || err.error.kind() == std::io::ErrorKind::FileTooLarge
+        {
             Self::out_of_disk(format!(
                 "failed to persist temporary file path {}: {}",
                 err.path.display(),
--- a/lib/collection/src/operations/types.rs
+++ b/lib/collection/src/operations/types.rs
@@ -962,6 +962,8 @@ pub enum CollectionError {
     ForwardProxyError { peer_id: PeerId, error: Box<Self> },
     #[error("Out of memory, free: {free}, {description}")]
     OutOfMemory { description: String, free: u64 },
+    #[error("Out of disk space: {description}")]
+    OutOfDisk { description: String },
     #[error("Timeout error: {description}")]
     Timeout { description: String },
     #[error("Precondition failed: {description}")]
@@ -1100,6 +1102,12 @@ impl CollectionError {
         }
     }
 
+    pub fn out_of_disk(description: impl Into<String>) -> Self {
+        Self::OutOfDisk {
+            description: description.into(),
+        }
+    }
+
     /// Returns true if the error is transient and the operation can be retried.
     /// Returns false if the error is not transient and the operation should fail on all replicas.
     pub fn is_transient(&self) -> bool {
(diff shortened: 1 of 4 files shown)


lib/storage/src/content_manager/errors.rs
```diff
@@ -161,6 +161,10 @@ impl StorageError {
                 description: overriding_description,
                 backtrace: None,
             },
+            CollectionError::OutOfDisk { .. } => StorageError::ServiceError {
+                description: overriding_description,
+                backtrace: None,
+            },
```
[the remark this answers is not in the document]
IvanPleshkov: This decision was made because out-of-memory is also converted to a service error. Maybe in the next changes, both errors will keep their semantics

agourlay:
Were you able to trigger it manually?

FYI we do have an end to end test for out of disk https://github.com/qdrant/qdrant/blob/f937d0260ffd7705c6f34d9c25da1e27ebf5ddfd/tests/e2e_tests/test_low_disk.py#L11

lib/collection/src/operations/types.rs
```diff
@@ -962,6 +962,8 @@ pub enum CollectionError {
     ForwardProxyError { peer_id: PeerId, error: Box<Self> },
     #[error("Out of memory, free: {free}, {description}")]
     OutOfMemory { description: String, free: u64 },
+    #[error("Out of disk space: {description}")]
+    OutOfDisk { description: String },
```
timvisee, line 966: I like adding this one. It makes a lot of sense to have it.

lib/collection/src/operations/types.rs
```diff
@@ -1265,9 +1279,15 @@ impl From<JsonError> for CollectionError {
 
 impl From<std::io::Error> for CollectionError {
     fn from(err: std::io::Error) -> Self {
-        CollectionError::ServiceError {
-            error: format!("File IO error: {err}"),
-            backtrace: Some(Backtrace::force_capture().to_string()),
+        if err.kind() == std::io::ErrorKind::StorageFull {
```
timvisee: Shall we also include [`FileTooLarge`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.FileTooLarge) here?

> File larger than allowed or supported.
> 
> This might arise from a hard limit of the underlying filesystem or file access API, or from an administratively imposed resource limitation. Simple disk full, and out of quota, have their own errors.

lib/collection/src/operations/types.rs
```diff
@@ -1412,11 +1432,19 @@ impl From<cancel::Error> for CollectionError {
 
 impl From<tempfile::PathPersistError> for CollectionError {
     fn from(err: tempfile::PathPersistError) -> Self {
-        Self::service_error(format!(
-            "failed to persist temporary file path {}: {}",
-            err.path.display(),
-            err.error,
-        ))
+        if err.error.kind() == std::io::ErrorKind::StorageFull {
```
timvisee: Idem ditto

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -64,7 +64,7 @@ impl ShardOperation for LocalShard {
             .await?
             .unwrap_or(false)
         {
-            return Err(CollectionError::service_error(
+            return Err(CollectionError::out_of_disk(
                 "No space left on device: WAL buffer size exceeds available disk space".to_string(),
```
timvisee:
```suggestion
                "No space left on device: WAL buffer size exceeds available disk space",
```

lib/segment/src/common/operation_error.rs
```diff
@@ -190,6 +198,9 @@ impl From<IoError> for OperationError {
                     free: free_memory,
                 }
             }
+            ErrorKind::StorageFull => OperationError::OutOfDisk {
```
timvisee, line 201: Idem ditto

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -64,8 +64,8 @@ impl ShardOperation for LocalShard {
             .await?
             .unwrap_or(false)
         {
-            return Err(CollectionError::service_error(
-                "No space left on device: WAL buffer size exceeds available disk space".to_string(),
+            return Err(CollectionError::out_of_disk(
+                "No space left on device: WAL buffer size exceeds available disk space",
```
xzfc, line 68: **suggestion (wording):** take into account that `OutOfDisk` error message prepends error text on its own.

So this message would repeat mention disk/device space three times, which is highly redundant:
```
Out of disk space: No space left on device: WAL buffer size exceeds available disk space
```

Instead, I suggest to make these messages short, e.g.:
```rust
CollectionError::out_of_disk("writing WAL buffer")
```

Which would give us this:
```
Out of disk space: writing WAL buffer
```

lib/collection/src/operations/types.rs
xzfc: It's for `EFBIG` errno which is not related to the amount of free space. I propose to not to include it.

PR #8021, opened 2026-01-30
chore: Use the qdrant/murmur3 fork

Anush008:
# Description

Point the `murmur3` dependency at our fork to ensure long-term build availability if upstream disappears.

EDIT: Because, If https://github.com/stusmall/murmur3 is deleted, our builds will break.

### All Submissions:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,7 +104,7 @@ prometheus = { version = "0.14.0", default-features = false }
 validator = { workspace = true }
 jsonwebtoken = { version = "10.0", features = ["rust_crypto"] }
 
-murmur3 = { git = "https://github.com/stusmall/murmur3", rev = "2c39087" }
+murmur3 = { git = "https://github.com/qdrant/murmur3", rev = "2c39087" }
 
 tempfile = { workspace = true }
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -104,7 +104,7 @@ prometheus = { version = "0.14.0", default-features = false }
 validator = { workspace = true }
 jsonwebtoken = { version = "10.0", features = ["rust_crypto"] }
 
-murmur3 = { git = "https://github.com/stusmall/murmur3", rev = "2c39087" }
+murmur3 = { git = "https://github.com/qdrant/murmur3", rev = "2c39087" }
```
xzfc, line 107: The only usage is here: https://github.com/qdrant/qdrant/blob/f937d0260ffd7705c6f34d9c25da1e27ebf5ddfd/src/common/inference/bm25.rs#L97

Perhaps add a note like this?
```toml
# until upstream releases v0.6.0 with `murmur3_32_of_slice`
murmur3 = { git = … }
```

anush008: We can cut a release in our fork.

timvisee: But we cannot publish to crates.io in a sane way. I'd prefer to wait some time for upstream to do a proper release.

timvisee, line 107: What is the benefit of this change?

I don't think this is needed, since we already specify a commit hash - though we can change it to a full hash.

anush008: We only need to push a git tag.

anush008: > What is the benefit of this change?

If https://github.com/stusmall/murmur3 is deleted, our builds will break. So we use our fork instead.

PR #8030, opened 2026-02-02
Drop wal from api: don't use applied seq number

timvisee:
Rework WAL drop in <https://github.com/qdrant/qdrant/pull/8000>. Don't rely on the applied seq number. But truncate from the first pending `op_num` in the channel instead.

### All Submissions:

### DIFF
--- a/lib/collection/src/shards/local_shard/wal_ops.rs
+++ b/lib/collection/src/shards/local_shard/wal_ops.rs
@@ -67,6 +67,13 @@ impl LocalShard {
             return Ok(0);
         };
 
+        // A corner case: truncate was called twice in a row without new updates in between.
+        // In this case, `truncate_from_op_num` is equal to last_wal_op_num + 1`.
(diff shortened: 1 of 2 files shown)


lib/collection/src/update_handler.rs
```diff
@@ -56,6 +56,9 @@ pub enum UpdateSignal {
     Nop,
     /// Ensures that previous updates are applied
     Plunger(oneshot::Sender<()>),
+    /// Ensures that previous updates are applied and returns the first `op_num`
+    /// that was skipped while `skip_updates` was set, if any.
+    SkipUpdatesPlunger(oneshot::Sender<Option<SeqNumberType>>),
```
timvisee: I don't like this extra variant, but it was a quick hack to put it together.

timvisee: We decided to remove it and repurpose the existing plunger.

https://github.com/qdrant/qdrant/pull/8030/commits/6531004040da5def6ce18e23e47bad6b8b63bd33

PR #8032, opened 2026-02-02
[trififo] ringbuffer with direct indexing

coszio:
Builds on top of #8012.

Adds the base container to be used by S3FIFO algorithm. It allows:
- direct indexing, just accessing an offset.
- simple reinsertion, by changing a single pointer.
- overwriting of last value, constrained to that single cell.

Uses atomic integers for `len` and `write_position` so that readers can correctly bounds-check concurrently with a writer.

### DIFF
--- a/lib/trififo/src/ringbuffer.rs
+++ b/lib/trififo/src/ringbuffer.rs
@@ -1,0 +1,228 @@
+use std::mem::MaybeUninit;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+/// Fixed-size ringbuffer with support for absolute indexing into its contents.
+///
+/// There is explicitly no way to remove elements before they are overwritten by
+/// filling the capacity. This can be changed, but other methods like `is_in_range` would
+/// need to be reviewed carefully.
+pub struct RingBuffer<T> {
+    /// Container for elements
+    buffer: Box<[MaybeUninit<T>]>,
+
+    /// Maximum number of elements in the buffer
+    capacity: usize,
+
+    /// Current write position (0 to capacity-1, wraps around)
+    write_pos: AtomicUsize,
+
+    /// Number of elements currently stored
+    len: AtomicUsize,
+}
+
+impl<T> RingBuffer<T> {
+    pub fn new(capacity: usize) -> Self {
+        let buffer: Box<[MaybeUninit<T>]> = (0..capacity).map(|_| MaybeUninit::uninit()).collect();
+
+        Self {
+            buffer,
+            capacity,
+            write_pos: AtomicUsize::new(0),
+            len: AtomicUsize::new(0),
+        }
+    }
+
+    /// Returns the number of elements currently stored in the buffer.
+    #[cfg(test)]
+    pub fn len(&self) -> usize {
+        self.len.load(Ordering::Acquire)
+    }
+
+    /// Returns true if the buffer is full.
+    #[inline]
+    pub fn is_full(&self) -> bool {
+        self.len.load(Ordering::Acquire) == self.capacity
+    }
+
+    /// Returns the current write position.
+    #[inline]
+    pub fn write_position(&self) -> usize {
+        self.write_pos.load(Ordering::Acquire)
+    }
+
+    /// Returns the oldest valid position offset in the buffer.
+    ///
+    /// Returns None if the buffer is empty.
+    pub fn read_position(&self) -> Option<usize> {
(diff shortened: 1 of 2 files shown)


lib/trififo/src/ringbuffer.rs
```diff
@@ -0,0 +1,231 @@
+use std::mem::MaybeUninit;
+use std::num::NonZeroUsize;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+/// Fixed-size ringbuffer with support for absolute indexing into its contents.
+///
+/// There is explicitly no way to remove elements before they are overwritten by
+/// filling the capacity. This can be changed, but other methods like `is_in_range` would
+/// need to be reviewed carefully.
```
xzfc, line 14: Perhaps add a comment justifying/listing differences of this implementation compared to the `ringbuf` (or other popular) crate.
I think it's about the `get_absolute` method?

coszio: Yep, none of the libraries allow to use direct indexing easily. Only `ringbuffer` can, but you cannot construct an offset yourself, since it handles wrap-arounds with an epoch number.

lib/trififo/src/ringbuffer.rs
```diff
@@ -0,0 +1,236 @@
...
+            *slot = item;
+        }
+
+        // Update write position with Release ordering to ensure the write
+        // is visible to readers before they see the new position
+        let new_write_pos = (offset + 1) % self.capacity;
+        self.write_pos.store(new_write_pos, Ordering::Release);
+
+        // Update length
+        let current_len = self.len.load(Ordering::Relaxed);
```
xzfc: Performance nit: we don't need to call `self.len.load()` twice in this method.

lib/trififo/src/ringbuffer.rs
```diff
@@ -0,0 +1,236 @@
...
+    #[inline]
+    pub fn get_absolute(&self, position: usize) -> Option<&T> {
+        if !self.is_in_range(position) {
+            return None;
+        }
+
+        // Safety: We just checked the value is within bounds
+        let value = &self.buffer[position];
+        let value = unsafe { value.assume_init_ref() };
+        Some(value)
```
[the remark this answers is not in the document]
coszio: It mentions SeqLock safety because in this crate, we intend to use it wrapped in a SeqLock, which allows readers to coexist with writers

lib/trififo/src/ringbuffer.rs
```diff
@@ -0,0 +1,236 @@
...
+            self.buffer[offset].write(item);
+        } else {
+            // SAFETY: It is within initialized bounds
+            let slot = unsafe { self.buffer[offset].assume_init_mut() };
+            *slot = item;
+        }
+
+        // Update write position with Release ordering to ensure the write
+        // is visible to readers before they see the new position
+        let new_write_pos = (offset + 1) % self.capacity;
```
xzfc: Performance: avoid division (`%`), e.g.:

```rust
let new_write_pos = if offset >= … { 0 } else { offset + 1 };
```

xzfc: And the same for other place that uses `%`.

lib/trififo/src/ringbuffer.rs
```diff
@@ -0,0 +1,236 @@
...
+
+impl<T> Drop for RingBuffer<T> {
+    fn drop(&mut self) {
+        let len = *self.len.get_mut();
+
+        for pos in 0..len {
+            unsafe {
+                std::ptr::drop_in_place(self.buffer[pos].as_mut_ptr());
+            }
+        }
```
xzfc: Style nit: use `assume_init_drop` (the same as `drop_in_place`):

```suggestion
        for slot in &mut self.buffer[..len] {
            unsafe { slot.assume_init_drop() };
        }
```

lib/trififo/src/ringbuffer.rs
coszio: Addressed in https://github.com/qdrant/qdrant/pull/8032/commits/f734ac59eb0cf93debb3015f9fe6be41ea36419d

lib/trififo/src/ringbuffer.rs
coszio: Addressed in https://github.com/qdrant/qdrant/pull/8032/commits/372547fb6bb9030e8795b59b4449e23c05e6fff7

PR #8033, opened 2026-02-02
Callback-based `DenseVectorStorage::for_each_in_dense_batch`

xzfc:
# What

This PR replaces APIs like these:

```rust
// accepts a slice of `MaybeUninit`s and return the same slice but
// initialized with references to the vector storage
fn get_batch<'a>(
    &'a self,
    keys: &[VectorOffsetType],
    vectors: &'a mut [MaybeUninit<&'a [T]>],
) -> &'a [&'a [T]];
```

with APIs like these:
```rust
// calls `f` on each vector from the storage
fn for_each_in_batch(
    &self,
    keys: &[impl VectorOffset], // accepts both `u32` and `usize` offsets
    f: impl FnMut(usize, &[T]),
);
```

# Why

It's a first step to implement local SSD cache for vector storages.
Letting storages accept callbacks would make lifetime juggling easier.

# Details

The drawback: because of generics used, the debug binary size grows from 580.9 MiB to 583.4 MiB.

Only batch methods are updated as these used in the hot path of search.
I'm not sure if non-batch methods (`DenseVectorStorage::get_dense`) are performance-critical.
Later I'll try to make them return `Cow<…>` instead of `&[T]` and see what happens.
Particularly, non-batch methods are used in `GraphLayersBuilder`/`GraphLayersHealer`.

# Performance

The bfb search performace is the same.

Also, I had to keep `MaybeUninit`/`maybe_uninit_fill_from` magic (#6303) because this code:

```rust
// Not in this PR: 15% slower on `bfb --dim 96`
for key in keys {
    f(self.get_vector(key));
}
```

is slower than this code (perhaps because of the cache locality):
```rust
// In this PR: same performance as in `dev` branch
let mut vectors_buffer = [MaybeUninit::uninit(); VECTOR_READ_BATCH_SIZE];
let vectors = maybe_uninit_fill_from(
    &mut vectors_buffer,
    keys.iter().map(|&key| self.get_vector(key)),
)
.0;
for vec in vectors {
    f(vec);
}
```

### DIFF
--- a/lib/segment/src/vector_storage/dense/mmap_dense_vectors.rs
+++ b/lib/segment/src/vector_storage/dense/mmap_dense_vectors.rs
@@ -170,19 +170,23 @@ impl<T: PrimitiveVectorElement> MmapDenseVectors<T> {
             .map(|offset| self.raw_vector_offset::<P>(offset))
     }
 
-    pub fn get_vectors<'a>(
-        &'a self,
-        keys: &[PointOffsetType],
-        vectors: &'a mut [MaybeUninit<&'a [T]>],
-    ) -> &'a [&'a [T]] {
-        debug_assert_eq!(keys.len(), vectors.len());
+    pub fn for_each_in_batch<F: FnMut(usize, &[T])>(&self, keys: &[PointOffsetType], mut f: F) {
         debug_assert!(keys.len() <= VECTOR_READ_BATCH_SIZE);
-        if is_read_with_prefetch_efficient(keys) {
+
+        // The `f` is most likely a scorer function.
+        // Fetching all vectors first then scoring them is more cache friendly
+        // then fetching and scoring in a single loop.
+        let mut vectors_buffer = [MaybeUninit::uninit(); VECTOR_READ_BATCH_SIZE];
+        let vectors = if is_read_with_prefetch_efficient(keys) {
             let iter = keys.iter().map(|key| self.get_vector::<Sequential>(*key));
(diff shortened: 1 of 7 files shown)


lib/segment/src/vector_storage/dense/mmap_dense_vectors.rs
```diff
@@ -170,19 +170,21 @@ impl<T: PrimitiveVectorElement> MmapDenseVectors<T> {
...
-        vectors: &'a mut [MaybeUninit<&'a [T]>],
-    ) -> &'a [&'a [T]] {
-        debug_assert_eq!(keys.len(), vectors.len());
-        debug_assert!(keys.len() <= VECTOR_READ_BATCH_SIZE);
-        if is_read_with_prefetch_efficient_points(keys) {
+    pub fn for_each_in_batch<F: FnMut(usize, &[T])>(&self, keys: &[PointOffsetType], mut f: F) {
+        // The `f` is most likely a scorer function.
+        // Fetching all vectors first then scoring them is more cache friendly
+        // then fetching and scoring in a single loop.
+        let mut vectors_buffer = [MaybeUninit::uninit(); VECTOR_READ_BATCH_SIZE];
```
coszio, line 179: If the buffer size is fixed, we should still at least `debug_assert!` that we don't have more keys than the buffer size. 

Alternatively, the size could be part of the function's const generics, but it would need more drill-down of the same constant everywhere.

coszio (approved):

PR #8044, opened 2026-02-03
Fix rescoring with inline storage + quantized search

StanByriukov02:
I did a little digging into your code and found a way to improve the data flow and speed it up, as well as improve your power consumption.

In this setup the graph-with-vectors path (`search_with_vectors`) can return results without going through the rescoring post-processing stage. That breaks the rescoring invariant and causes the quantized HNSW integration test to fail.

Fix:
- If rescoring is enabled, skip `search_with_vectors`.
- Select candidates using the quantized scorer and apply rescoring only to the oversampled top-k in `postprocess_search_result`.

Test coverage:
- Run the quantized HNSW integration test with `inline_storage=true` so this behavior stays covered.

Test plan:
- cargo test -p segment --test integration hnsw_quantized_search_cosine_test -- --nocapture

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -995,6 +995,11 @@ impl HNSWIndex {
         }
     }
 
+    /// Perform HNSW graph search for a single query vector.
+    ///
+    /// This method chooses between the "graph-with-vectors" traversal path and the regular
+    /// traversal path. Important invariant: when quantization rescoring is enabled, results must
+    /// go through `postprocess_search_result` so rescoring is applied to the oversampled top-k.
     #[allow(clippy::too_many_arguments)]
     fn search_with_graph(
         &self,
--- a/lib/segment/tests/integration/hnsw_quantized_search_test.rs
+++ b/lib/segment/tests/integration/hnsw_quantized_search_test.rs
@@ -43,6 +43,12 @@ pub fn sames_count(a: &[Vec<ScoredPointOffset>], b: &[Vec<ScoredPointOffset>]) -
         .count()
 }
 
+/// Integration test helper for quantized HNSW search.
+///
+/// Covers three invariants:
+/// - quantized search stays reasonably close to exact search (recall/overlap),
+/// - oversampling doesn't make results worse,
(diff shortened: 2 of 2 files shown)


generall:
@StanByriukov02 thanks for the PR! Could you please rebase it to `dev` branch?

StanByriukov02:
Done :)

xzfc:
> Fix:
> If rescoring is enabled, skip search_with_vectors.

It should be the other way around. If we use `search_with_vectors` then we don't need rescoring step, even if it is enabled via parameters.


Reasoning mentioned in [1.16 release notes](https://qdrant.tech/blog/qdrant-1.16.x/#inline-storage---disk-efficient-vector-search):
> Moreover, an original non-quantized vector is also embedded into the same graph node. The original vector is used to perform an implicit rescoring during the search, eliminating the separate rescore step which is usually performed after the search.


Implicit rescore is here:
https://github.com/qdrant/qdrant/blob/f937d0260ffd7705c6f34d9c25da1e27ebf5ddfd/lib/segment/src/index/hnsw_index/graph_layers.rs#L374-L377

xzfc:
As for why test fails without this fix if we enable `inline_storage` in the test. The test wrecks the original vector storage and assumes/expects that the search function will read these wrecked vectors during rescoring.
https://github.com/qdrant/qdrant/blob/f937d0260ffd7705c6f34d9c25da1e27ebf5ddfd/lib/segment/tests/integration/hnsw_quantized_search_test.rs#L177-L188

Note that the `inline_storage`/`graph_with_vectors` path don't use neither original nor quantized vector storages, thus this test fails.

I'd say it's work as intended, so closing it.

StanByriukov02:
Thanks for the detailed explanation and the release-notes pointer — you’re right. I incorrectly assumed rescoring must go through postprocess when rescore is enabled. With inline_storage/graph_with_vectors, the base (non-quantized) vector embedded in the graph node is used for implicit rescoring during traversal, so a separate rescore stage is not required.\n\nAlso agreed on the test point: the integration test ‘wrecks’ the vector storage and expects rescoring to read those wrecked vectors; that assumption doesn’t hold for inline_storage since it doesn’t use the vector storages. My change to force inline_storage in that test therefore wasn’t a valid correctness signal.\n\nAppreciate the clarification.

PR #8050, opened 2026-02-03
Remove `UpdateSignal::Stop`

IvanPleshkov:
## Summary
This PR removes `UpdateSignal::Stop` and changes WAL truncation mechanism.

## Description
When implementing the update queue feature, we discovered a problem with the current worker stop mechanism. If the update worker has many pending operations in the channel, sending UpdateSignal::Stop means we have to wait until all operations are processed before the worker actually stops - the stop signal sits at the end of the queue behind all pending operations.

As the solution, update worker does not have a stop signal. Instead, channel and stop signal are checked using `tokio::select`. Also, update worker returns the channel with all pending operations.

## Wal truncation
Because new stopping mechanism allows to stop update worker immediately and get all pending operations, this PR also refactors WAL truncation, where it's not necessary anymore to make overcomplicated update worker.

## Open problems
Can we change the update channel size via config update? If yes, there is a problem in the config optimizer update, where we cannot fill the new update worker if the config shrinks the update queue size

### DIFF
--- a/lib/collection/src/shards/local_shard/updaters.rs
+++ b/lib/collection/src/shards/local_shard/updaters.rs
@@ -54,7 +54,10 @@ impl LocalShard {
                 // It's not expected to happen in normal operation,
                 // Because it means that there were no update workers running.
                 // Just in case, we create a new channel to avoid panics in update handler.
-                debug_assert!(false, "Update receiver was None during optimizer config update");
+                debug_assert!(
+                    false,
+                    "Update receiver was None during optimizer config update"
+                );
                 let (update_sender, update_receiver) =
                     mpsc::channel(self.shared_storage_config.update_queue_size);
                 let _old_sender = self.update_sender.swap(Arc::new(update_sender));


lib/collection/src/update_workers/update_worker.rs
```diff
@@ -38,34 +38,28 @@ impl UpdateWorkers {
...
+                biased; // biased to check cancellation first
+                _ = cancel.cancelled() => {
+                    break receiver;
+                }
+                signal = receiver.recv() => signal,
+            };
+
+            let Some(signal) = signal else {
+                break receiver;
+            };
```
timvisee: Nit: I'd prefer it like this to contain the breakpoints within one block:

```suggestion
                signal = receiver.recv() => match signal {
                    Some(signal) => signal,
                    None => break receiver,
                }
            };
```

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -426,14 +426,28 @@ impl ShardOperation for LocalShard {
 
     /// Finishes ongoing update tasks
     async fn stop_gracefully(mut self) {
-        if let Err(err) = self.update_sender.load().send(UpdateSignal::Stop).await {
-            log::warn!("Error sending stop signal to update handler: {err}");
+        {
+            // Send stop signals to workers
+            let mut update_handler = self.update_handler.lock().await;
+            update_handler.stop_flush_worker();
+            update_handler.stop_update_worker();
```
timvisee, line 433: If wonder if we should also add `update_handler.notify_optimization_handles_to_stop();` here?

lib/collection/src/shards/local_shard/updaters.rs
```diff
@@ -37,12 +42,25 @@ impl LocalShard {
...
+
+        // Forward pending operations from old receiver to new channel
+        if let Some(mut old_receiver) = pending_receiver {
+            let sender = self.update_sender.load();
+            while let Ok(signal) = old_receiver.try_recv() {
+                // Forward pending operations to new channel
+                // Use try_send to avoid blocking - if channel is full, operations are dropped
+                let _ = sender.try_send(signal);
+            }
+        }
```
timvisee: Why do we create a new channel here? Can't we repurpose the existing receiver since we now get it back?

lib/collection/src/shards/local_shard/shard_ops.rs
IvanPleshkov: It's a good idea. And I want to do it too.
It's non-trivial logic, where update worker stopping sends stop to optimizers.
But this PR does not touch optimizers
I propose to refactor it as a next step and keep this PR easier to review

lib/collection/src/shards/local_shard/updaters.rs
IvanPleshkov: Not sure how to behave properly.
If we repurpose the existing one - we cannot change the update queue size, only after restart.
If we change the update queue size, we may lost operations if a new size is smaller.
What I can propose. Create a new channel only when there is a new update queue size and old update queue fits the new size

IvanPleshkov: Because update queue size is a node-level config and cannot be changed in runtime, I changed this logic and reuse old receiver.
In case of a non-existing receiver, I propose to create a new channel instead of returning an error.

timvisee: This cannot be changed at runtime. We agreed to handle the case of 'lower setting than current required queue size' in a separate PR.

generall (approved):

Issue #7975, opened 2026-01-23
Cold-data loading optimization


1995chen:
Hi, after importing 200 M vectors into 15 shards the cold-start is painfully slow. We turned on binary quantization and keep the HNSW index in RAM, while the raw vectors stay on disk. Whenever we migrate to new hosts and restart the cluster, the startup takes up to one hour. My first guess is that the code reads the shard index files sequentially: it first loads the quantized index and only then the HNSW index. With this data size the serial load becomes a bottleneck; in an emergency the long recovery time can break our SLA.
I’d like to prefetch all required files concurrently before the service starts, so later index access will hit the OS cache. 
Users can set a concurrency parameter to match the maximum IOPS their disks can deliver. It’s admittedly ugly, but it’s the quickest way to buy us time. In the long run we still need to make the index-loading code itself parallel.
Below are the startup logs:
2026-01-23T02:38:44.426125Z  INFO storage::content_manager::consensus::persistent: Loading raft state from ./storage/raft_state.json
2026-01-23T02:38:44.435406Z  INFO storage::content_manager::toc: Loading collection: t_scalar
2026-01-23T03:24:14.428535Z  INFO collection::shards::local_shard: Recovering shard ./storage/collections/t_scalar/1: 0/1 (0%)
2026-01-23T03:24:14.487988Z  INFO collection::shards::local_shard: Recovered collection t_scalar: 1/1 (100%)
2026-01-23T03:26:30.083950Z  INFO collection::shards::local_shard: Recovering shard ./storage/collections/t_scalar/3: 0/1 (0%)
2026-01-23T03:26:30.190590Z  INFO collection::shards::local_shard: Recovered collection t_scalar: 1/1 (100%)
2026-01-23T03:28:48.091480Z  INFO collection::shards::local_shard: Recovering shard ./storage/collections/t_scalar/5: 0/1 (0%)
2026-01-23T03:28:48.201508Z  INFO collection::shards::local_shard: Recovered collection t_scalar: 1/1 (100%)
2026-01-23T03:30:50.456139Z  INFO collection::shards::local_shard: Recovering shard ./storage/collections/t_scalar/7: 0/1 (0%)
2026-01-23T03:30:50.568487Z  INFO collection::shards::local_shard: Recovered collection t_scalar: 1/1 (100%)

We’re running on AWS r7g.8xlarge (32 vCPU, 256 GiB) with gp3 volumes configured at 16 000 IOPS.

generall:
[...]

PR #8053, opened 2026-02-04
feat: Optimize the collection loading process during startup

1995chen:
Fixes <https://github.com/qdrant/qdrant/issues/7975>

We have added some configurations to support concurrency settings for collection, shard, and segment loading, while maintaining the original default values, enabling Qdrant to achieve faster loading speeds on AWS EBS network volumes.

### DIFF
--- a/lib/storage/tests/integration/alias_tests.rs
+++ b/lib/storage/tests/integration/alias_tests.rs
@@ -57,11 +57,7 @@ fn test_alias_operation() {
             incoming_shard_transfers_limit: Some(1),
             outgoing_shard_transfers_limit: Some(1),
             async_scorer: None,
-            load_concurrency: LoadConcurrencyConfig {
-                max_concurrent_collection_loads: Some(1),
-                max_concurrent_shard_loads: Some(1),
-                max_concurrent_segment_loads: Some(8),
-            },
+            load_concurrency: LoadConcurrencyConfig::default(),
         },
         hnsw_index: Default::default(),
         hnsw_global_config: Default::default(),


lib/storage/src/types.rs
```diff
@@ -53,6 +54,9 @@ pub struct PerformanceConfig {
     pub outgoing_shard_transfers_limit: Option<usize>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub async_scorer: Option<bool>,
+
```
timvisee:
```suggestion
```

src/main.rs
```diff
@@ -177,7 +178,8 @@ fn main() -> anyhow::Result<()> {
             .async_scorer
             .unwrap_or_default(),
     );
-
+    // Set max concurrent for collection load
+    set_concurrent_loads_config(&settings.storage.performance.concurrent_loads);
```
timvisee: I'm not a fan of setting globals for this.

I'm checking whether we can easily adjust it to prevent this.

lib/common/common/src/concurrent_loads.rs
```diff
@@ -0,0 +1,85 @@
...
+        None => default,
+        Some(0) => {
+            log::warn!(
+                "Received 0 for max concurrent {name} loads config, using default {default}"
+            );
+            default
+        }
+        Some(v) => v,
+    }
+}
```
timvisee: We can use `NonZeroUsize` instead to prevent `0` as input during deserialization

timvisee:
I tested the following setup locally:
- 1 peer
- 20 collections
- 12 shards each
- 32 segments each
- local NVMe disk

The default case (1/1/8) loads in 75 seconds.

With a (4/4/8) configuration loading took longer, taking 105 seconds.

```bash
QDRANT__STORAGE__PERFORMANCE__MAX_CONCURRENT_COLLECTION_LOADS=4 QDRANT__STORAGE__PERFORMANCE__MAX_CONCURRENT_SHARD_LOAD=4 QDRANT__STORAGE__PERFORMANCE__MAX_CONCURRENT_SEGMENT_LOAD=8 ./target/perf/qdrant
```

My best guess is that the high amount of context switching makes load performance worse on a setup like this. I can imagine it being effective on a networked disk with higher latency though. And so I don't have an objection to merge this.

1995chen:
> I tested the following setup locally:
> 
> * 1 peer
> * 20 collections
> * 12 shards each
> * 32 segments each
> * local NVMe disk
> 
> The default case (1/1/8) loads in 75 seconds.
> 
> With a (4/4/8) configuration loading took longer, taking 105 seconds.
> 
> ```shell
> QDRANT__STORAGE__PERFORMANCE__MAX_CONCURRENT_COLLECTION_LOADS=4 QDRANT__STORAGE__PERFORMANCE__MAX_CONCURRENT_SHARD_LOAD=4 QDRANT__STORAGE__PERFORMANCE__MAX_CONCURRENT_SEGMENT_LOAD=8 ./target/perf/qdrant
> ```
> 
> My best guess is that the high amount of context switching makes load performance worse on a setup like this. I can imagine it being effective on a networked disk with higher latency though. And so I don't have an objection to merge this.

Thank you so much, @timvisee  and @generall !
Our dataset is still quite large, so we're not able to move to SSD storage yet. This PR has made a really big difference — our service startup time and especially failure recovery have improved noticeably. It's been a huge help for us in production.
Really appreciate your work on this — thank you both again! 🙏

timvisee:
We've released [Qdrant 1.17.0](https://github.com/qdrant/qdrant/releases/tag/v1.17.0) which includes this improvement.

Thank you for implementing this one.

PR #8056, opened 2026-02-04
Don't lock `SegmentHolder` for the entire duration of read operations

JojiiOfficial:
Continuation of #8024

### Setup
Parallel updates + searches using the following commands:
```bash
# Upsert
bfb  -n 10M -d 512 --shards 3 --replication-factor 2 --on-disk-vectors true \
     --keywords 5000 --hnsw-m 0 --hnsw-payload-m 16 --tenants true -b 10 --timeout 60 --rps 100

# Search
bfb -n 230k -d 512 --skip-setup --search --keywords 5000 --rps 200 --measure-slow-requests true
```
For hardware, a 2 core 8GB cloud cluster with 1 node was used.

For the searches, a slightly modified version of BFB was used, mostly to measure slow requests in detail.
The changes can be found here: https://github.com/qdrant/bfb/pull/97

### Results
The results of 2-3 runs on each Qdrant@1.16.3 and this PR averaged.

PR
```
--- QPS (averaged) ---
Min qps:	192.970664440867
Avg qps:	199.982515579988
Median qps:	199.999541695647
Max qps:	201.976920003441

--- Slow requests (averaged) ---
Total: 976
Min slow request	time	0.100097543
Avg slow request	time	0.221581551442721
Median slow request	time	0.1821884285
p95 slow request	time	0.512773123
p99 slow request	time	0.578455576
Max slow request	time	0.6067236445


```

Qdrant@1.16.3
```
--- QPS (averaged) ---
Min qps:	194.190540742037
Avg qps:	199.985118634619
Median qps:	199.999534991853
Max qps:	204.55869924178

--- Slow requests (averaged) ---
Total: 1463
Min slow	request	time	0.100111962
Avg slow	request	time	0.314250290654919
Median slow	request	time	0.276687310333333
p95 slow	request	time	0.650480541
p99 slow	request	time	0.728994755
Max slow	request	time	0.792417519666667
```

Furthermore:
- On Dev, the upsert command almost always timeouted after ~11 minutes.
- On this PR, the timeout was around ~13-15 minutes.

These results show that this PR improves on slow requests during search

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -1317,7 +1317,11 @@ impl ShardReplicaSet {
                     let mut total_payload_size = 0;
                     let mut total_points = 0;
 
-                    for (_, segment) in local.segments.read().iter() {
+                    // Collect the segments first so we don't have the segment holder locked for the entire duration of the loop.
+                    let segments: Vec<_> =
+                        local.segments.read().iter().map(|i| i.1).cloned().collect();
+
+                    for segment in segments {
                         let size_info = segment.get().read().size_info();
                         total_vector_size += size_info.vectors_size_bytes;
                         total_payload_size += size_info.payloads_size_bytes;


lib/shard/src/segment_holder/snapshot.rs
```diff
@@ -21,6 +21,8 @@ impl SegmentHolder {
     pub fn snapshot_manifest(&self) -> OperationResult<SnapshotManifest> {
         let mut manifest = SnapshotManifest::default();
 
+        // TODO(Jonas): Maybe make this function take a Vec of locked segments and remove the `&self` reference
```
generall: I don't think we care too much if snapshot manifest got delayed 100ms

lib/collection/src/shards/replica_set/mod.rs
timvisee: We need the same treatment here. This still read locks the segment holder for the duration of the for loop.

Calling `size_info()` should be relatively cheap, but there's another read lock on the segment itself.

Two problems:
1. on long writes the inner read lock might hang for some time
2. the sync read lock blocks the async runtime!

In my opinion we should clone the segments. Then use `spawn_blocking` and run the for loop with read locks in blocking context.

JojiiOfficial: Good catch! I fixed it in https://github.com/qdrant/qdrant/pull/8056/commits/34cf12f61c715c3a5bb738e5f4282803b282f73f


> In my opinion we should clone the segments. Then use spawn_blocking and run the for loop with read locks in blocking context.

I'll tackle this in a separate PR 👍🏻

timvisee: https://github.com/qdrant/qdrant/pull/8069

PR #8059, opened 2026-02-04
For snapshots directly restored, use default temp path in storage volume

timvisee:
With <https://github.com/qdrant/qdrant/pull/8025> we now stream unpacking on snapshot recovery.

This changes the default temporary directory for those actions from the snapshot directory to the storage directory.

In practice this means the snapshot is stored on the storage directory directly. This is significant for environments having a separate storage and snapshot volume, like our own cloud. Before a snapshot had to be moved across volumes to be promoted, which is very expensive. Now promotion is a simple directory rename making it super fast.

A minimal change to make the above behavior happen. We might do a more significant refactoring of how we configure and handle temporary files in the future.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/storage/src/content_manager/snapshots/recover.rs
+++ b/lib/storage/src/content_manager/snapshots/recover.rs
@@ -122,7 +122,8 @@ async fn _do_recover_from_snapshot(
     } = download_snapshot(
         client,
         location,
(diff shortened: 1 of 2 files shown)


tellet-q (approved):
lgtm

PR #8060, opened 2026-02-04
Update queue status in metrics and telemetry

IvanPleshkov:
This PR continues https://github.com/qdrant/qdrant/pull/8010
And adds update queue status to the telemetry and metrics.

```
# HELP collection_update_queue_length number of pending operations in update queues per collection
# TYPE collection_update_queue_length gauge
collection_update_queue_length{id="benchmark"} 123
```

### DIFF
--- a/docs/redoc/master/openapi.json
+++ b/docs/redoc/master/openapi.json
@@ -12603,6 +12603,17 @@
               "minimum": 0
             },
             "nullable": true
+          },
+          "update_queue": {
+            "description": "Update queue status",
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/UpdateQueueInfo"
+              },
+              {
+                "nullable": true
+              }
+            ]
           }
         }
       },


src/common/metrics.rs
```diff
@@ -504,6 +519,14 @@ impl CollectionsTelemetry {
...
             prefix,
         ));
+
+        metrics.push_metric(metric_family(
+            "collection_update_queue_length",
+            "number of pending operations in update queues per collection",
+            MetricType::GAUGE,
+            update_queue_length,
+            prefix,
+        ));
```
timvisee, line 529: This renders the queue size per collection.

Would it also be interesting to show it per shard ID or shard key? Since shards can have wildly different queue sizes it may be useful, especially in the case where user defined sharding is used.

I do vote for keeping it at collection level like this for now. And then we can extend it in the future if there is any desire. That way we don't overload metrics.

timvisee, line 529: I did also expect to expose the last `op_num` here. But I understand we cannot really do that at collection level. 👍

timvisee (approved):
🧪🏁

src/common/metrics.rs
IvanPleshkov: Agree, for now keep as is, will see later if we want to extend metrics

PR #8063, opened 2026-02-05
Weighted rrf

generall:
Based on https://www.notion.so/qdrant/Weighted-RRF-255674779d3380b19f29da1de6ad81d6

```
{
  "prefetch": [
    {
      "query": 1,
      "filter": {
        "must": {
          "key": "a",
          "match": {
            "value": "keyword_1"
          }
        }
      }
    },
    {
      "query": 1,
      "filter": {
        "must": {
          "key": "a",
          "match": {
            "value": "keyword_2"
          }
        }
      }
    }
  ],
  "query": {
    "rrf": {
      "weights": [2, 1]
    }
  },
  "with_payload": true
}
```

response:
<details>

```
{
  "result": {
    "points": [
      {
        "id": 7874,
        "version": 81,
        "score": 0.6666667,
        "payload": {
          "a": "keyword_1"
        }
      },
      {
        "id": 885,
        "version": 11,
        "score": 0.5,
        "payload": {
          "a": "keyword_2"
        }
      },
      {
        "id": 2068,
        "version": 23,
        "score": 0.5,
        "payload": {
          "a": "keyword_1"
        }
      },
      {
        "id": 3508,
        "version": 38,
        "score": 0.4,
        "payload": {
          "a": "keyword_1"
        }
      },
      {
        "id": 5333,
        "version": 56,
        "score": 0.33333334,
        "payload": {
          "a": "keyword_1"
        }
      },
      {
        "id": 3519,
        "version": 38,
        "score": 0.33333334,
        "payload": {
          "a": "keyword_2"
        }
      },
      {
        "id": 6841,
        "version": 71,
        "score": 0.2857143,
        "payload": {
          "a": "keyword_1"
        }
      },
      {
        "id": 6784,
        "version": 70,
        "score": 0.25,
        "payload": {
          "a": "keyword_1"
        }
      },
      {
        "id": 334,
        "version": 6,
        "score": 0.25,
        "payload": {
          "a": "keyword_2"
        }
      },
      {
        "id": 3470,
        "version": 37,
        "score": 0.22222222,
        "payload": {
          "a": "keyword_1"
        }
      }
    ]
  },
  "status": "ok",
  "time": 0.010133347
}
```

</details>

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -186,7 +186,6 @@ impl StableHash for ExtendedPointId {
 }
 
 impl ExtendedPointId {
-
     #[cfg(any(test, feature = "testing"))]
     pub fn as_u64(&self) -> u64 {
         match self {


lib/segment/src/common/reciprocal_rank_fusion.rs
```diff
@@ -11,27 +11,54 @@ use crate::types::{ExtendedPointId, ScoredPoint};
...
+/// - Position 3 in source with weight 3 has score: 1.0 / (3 * (1/3) + k) = 1.0 / (1 + k)
+///
+/// This means a 3:1 weight ratio is equivalent to "for each 3 results of first prefetch,
+/// have one result of second".
+fn position_score(position: usize, k: usize, weight: f32) -> f32 {
+    // Avoid division by zero - if weight is 0, treat as negligible contribution
+    if weight <= 0.0 {
+        return 0.0;
+    }
+    1.0 / (position as f32 * (1.0 / weight) + k as f32)
```
coszio: Initially I was going to suggest to put the weight in the numerator of the original function. Like $\frac{x}{p + k}$, but after analyzing it more carefully, the semantic of ratios holds better when applied the way it is done here 👍 .

In $\frac{1}{\frac{p}{w} + k}$, setting 3:1 indeed means "The 3rd result of the first set is as good as the 1st result of the second set".

I would only suggest to do the division directly:

```suggestion
    1.0 / (position as f32 / weight) + k as f32)
```

lib/api/src/rest/schema.rs
```diff
@@ -537,6 +537,12 @@ pub struct Rrf {
     #[validate(range(min = 1))]
     #[serde(default)]
     pub k: Option<usize>,
+
+    /// Weights for each prefetch source. Higher weight gives more influence on the final ranking.
+    /// If not specified, all prefetches are weighted equally.
+    /// The number of weights should match the number of prefetches.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub weights: Option<Vec<f32>>,
```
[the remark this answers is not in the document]
generall: validated in MergePlan constructor

lib/collection/src/operations/universal_query/shard_query.rs
```diff
@@ -23,7 +23,7 @@ pub fn query_result_order(
                 }
             }
             ScoringQuery::Fusion(fusion) => match fusion {
-                FusionInternal::RrfK(_) | FusionInternal::Dbsf => Some(Order::LargeBetter),
+                FusionInternal::Rrf { .. } | FusionInternal::Dbsf => Some(Order::LargeBetter),
```
coszio: We should destructure for letting the compiler complain in the future

lib/segment/src/common/reciprocal_rank_fusion.rs
generall: I believe you ment

```
1.0 / (position as f32 / weight + k as f32)
```

lib/collection/src/operations/universal_query/shard_query.rs
generall: not sure about this specific case. Usually the problem happens when we use one parameter, but not others. In this case we don't care about any parameters what so ever. 
But I guess I have to follow my own requirements.

generall:
cc @abdonpijpelink

lib/segment/src/common/reciprocal_rank_fusion.rs
```diff
@@ -4,34 +4,77 @@
...
+///
+/// This means a 3:1 weight ratio is equivalent to "for each 3 results of first prefetch,
+/// have one result of second".
+fn position_score(position: usize, k: usize, weight: f32) -> f32 {
+    // Avoid division by zero - if weight is 0, treat as negligible contribution
+    if weight <= 0.0 {
+        return 0.0;
+    }
+
+    1.0 / ((position + 1) as f32 / weight + k as f32 - 1.0)
```
generall, line 38: this formula doesn't exactly match document, but have 2 nice properties:

- if weight = 1.0, reduces to regular rrf as we had before
- if weight != 1.0, score for position=0  _is_ affected by weight

coszio: I wish we initially started ranks at 1 so that we don't need to do this for backward compatibility. Original function works nicely if ranks start at 1. The only inconvenience is that `k` gets slightly affected, compared to the original function.

Maybe docs can use the simpler formula, and stating that ranks start at 1.

The alternative would be to have the 2 different formula implementations

PR #8064, opened 2026-02-05
[trififo]: `cache_comparison_io` benchmark

xzfc:
A new benchmark which somehow reflects the real intended use of our cache controller. Assumes we will have one big dense file on the cold storage and one smaller dense file on the hot storage. Compares three implementations.

2026-02-05-cache-bench drawio

The most important part is `with_vector` implementations. It roughly highlights what we need from cache primitives.

Command to run:
```
systemd-run --user --scope -p MemoryMax=512M -- \
cargo bench -p trififo --bench cache_comparison_io -- \
--threads 32 --controller qc /path/to/hot/storage.bin /path/to/cold/storage.bin
```

(first line to simulate "there are less RAM than the file size")

There are some missing things in this benchmark, but I want to provide something runnable right now. Missing things:
- [ ] "dumb_mmap" controller without `O_DIRECT`. (to treat it as a baseline)
- [ ] I'll need to run it on some cloud machine.
- [ ] Maybe use `Vec` instead of `VecDeque` for tracking unused ids.
  - [ ] Or more sophisticated structure like bitmap/interval tree/etc.
- [ ] Some way to prevent evicting items between `get_value_or_guard`/`get` and `read_from_local_ssd`.
- [ ] More kinds of distributions. (not only zipf)
  - [ ] Zipf with shiffled ids.
  - [ ] Linear scans.
  - [ ] Others.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7960,11 +7960,16 @@ name = "trififo"
 version = "0.0.0"
 dependencies = [
  "ahash",
+ "anyhow",
  "cap",
+ "clap",
  "criterion",
  "foyer",
+ "fs-err",
  "hashbrown 0.16.1",
  "itertools 0.14.0",
+ "memmap2",
+ "nix 0.31.1",
  "parking_lot",
  "quick_cache",
  "rand 0.9.2",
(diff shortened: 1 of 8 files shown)


xzfc:
Superseded by #8109.

PR #8065, opened 2026-02-05
refactor/opt

dancixx:
The goal is to separate the optimization process so it’s independent of the collection crate, allowing it to be used on Edge. The requirement is to keep the orchestration part in the collection crate, while moving the execution into the shard crate. @xzfc 

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/shard/src/optimizers/vacuum_optimizer.rs
+++ b/lib/shard/src/optimizers/vacuum_optimizer.rs
@@ -76,7 +76,7 @@ impl VacuumOptimizer {
     /// We are only interested in indexed vectors, as they are the ones affected by soft-deletes.
     ///
     /// This finds the maximum deletion ratio for a named vector. The ratio is based on the number
-    /// of deleted vectors versus the number of indexed vector.s
+    /// of deleted vectors versus the number of indexed vectors.
     ///
     /// Returns `None` if littered ratio did not reach vacuum thresholds for no named vectors.
     fn littered_vectors_index_ratio(&self, segment: &Segment) -> Option<f64> {


config/config.yaml
```diff
@@ -210,29 +210,29 @@ storage:
...
 
-      # Whether exact search is allowed or not.
-      #search_allow_exact: null
+    # Whether exact search is allowed or not.
+    #search_allow_exact: null
 
-      # Max oversampling value allowed in search.
-      #search_max_oversampling: null
+    # Max oversampling value allowed in search.
+    #search_max_oversampling: null
```
timvisee: I think we should revert this config change

dancixx: Yeah sorry, fixed.

xzfc:
Could you please explain "what for" and "why this way" in the PR description?

I suppose it's to make some of the moved items to be available in `edge`, but I'd appreciate more details.

dancixx:
@xzfc, the pr description is updated. If you need more details, please ping me.

xzfc:
@dancixx I'm trying to understand full picture and future steps.

1. I suppose the plan is to make it like the following. Am I correct?
   - The `collection` (aka non-edge qdrant) will decide:
      - (§A) which segments to optimize based on `config_mismach`/`indexing`/`merge`/`vacuum` logic,
      - and which `SegmentConfig` to use based on `CollectionConfig` (or other non-edge configuration).
   
   - The `edge` will decide:
      - (§A) which segments to optimize based on its own logic (so, nothing in common with `config_mismach`/`indexing`/`merge`/`vacuum`),
      - and which `SegmentConfig` to use also based on its own logic.

2. (the same question about (§A) but from the other perspective)
   I see that `OptimizationStrategy` is now in the `shard`.
   But all four `OptimizationStrategy` implementations are still in the `collection` crate.
   Do we want to move these four impls to the `shard`? Or we want to add a new edge-specific `OptimizationStrategy` implementation instead?

dancixx:
@xzfc That’s a good point. I’ve moved the PR back to draft. I’m on PTO this week, but I’ll clarify everything once I’m back.

lib/collection/src/optimizers_builder.rs
```diff
@@ -168,6 +173,57 @@ pub fn clear_temp_segments(shard_path: &Path) {
     }
 }
 
+pub fn build_segment_optimizer_config(
+    collection_params: &CollectionParams,
+    hnsw_config: &HnswConfig,
+    quantization_config: &Option<QuantizationConfig>,
+) -> CollectionResult<SegmentOptimizerConfig> {
```
generall: I think we can avoid `CollectionResult`, as this function might be in-failable

PR #8069, opened 2026-02-06
Use spawn_blocking for segment locking

JojiiOfficial:
Follow up of #8056

Uses spawn_blocking for the potentially blocking `segment.lock()` operation which was discussed here https://github.com/qdrant/qdrant/pull/8056#discussion_r2773075953

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -1327,18 +1327,22 @@ impl ShardReplicaSet {
             return Ok(None);
         };
 
-        let mut total_vector_size = 0;
-        let mut total_payload_size = 0;
-        let mut total_points = 0;
+        let handle = spawn_blocking(move || {
+            let mut total_vector_size = 0;
+            let mut total_payload_size = 0;
+            let mut total_points = 0;
 
-        for segment in segments {
-            let size_info =
-                AbortOnDropHandle::new(spawn_blocking(move || segment.get().read().size_info()))
-                    .await?;
-            total_vector_size += size_info.vectors_size_bytes;
-            total_payload_size += size_info.payloads_size_bytes;
-            total_points += size_info.num_points;
-        }
+            for segment in segments {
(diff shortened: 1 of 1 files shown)


lib/collection/src/common/collection_size_stats.rs
```diff
@@ -33,18 +35,20 @@ impl CollectionSizeStatsCache {
...
-            let updated = update_fn().await?;
+            let Some(updated) = update_fn().await? else {
+                return Ok(None);
+            };
             self.update(updated);
         }
 
         // Give caller access to cached (inner) values which are always updated if required
-        self.stats.as_ref()
+        Ok(self.stats.as_ref())
```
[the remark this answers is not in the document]
JojiiOfficial: No, if the update function returned `None` before, the `?` in the line `let updated = update_fn().await?;` would've caused the early return. Therefore the new code is semantically equivalent

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -1306,40 +1307,42 @@ impl ShardReplicaSet {
...
-                | Shard::QueueProxy(_)
-                | Shard::Dummy(_) => None,
-            })
-            .unwrap_or_default()
+        let mut total_vector_size = 0;
+        let mut total_payload_size = 0;
+        let mut total_points = 0;
+
+        for segment in segments {
+            let size_info = spawn_blocking(move || segment.get().read().size_info()).await?;
```
timvisee: We should wrap this spawn in [`AbortOnDropHandle`](https://docs.rs/tokio-util/latest/tokio_util/task/struct.AbortOnDropHandle.html)

It's unlikely to have an effect in the majority of cases, but it could be useful if we've overloaded the thread pool. Adding it is basically free.

timvisee: I suggest to run the whole for loop in the blocking context. That ways we have less thread juggling if the number of segments is very high.

PR #8071, opened 2026-02-07
audit logging

generall:
This PR introduces optional Audit Logging functionality.

This functionality, if enabeld, logs all authentication events (with exception for healthcheck, metrics and internal ops) into a dedicated directory with reach metadata.

Config example:

```
audit:
   enabled: false
   dir: ./storage/audit
   rotation: daily
   max_log_files: 7
   trust_forwarded_headers: false
```

<details>
<summary>Log examples</summary>

```
{"timestamp":"2026-02-07T11:38:47.774671900Z","method":"cluster_status","auth_type":"ApiKey","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:38:47.775804751Z","method":"telemetry_cluster","auth_type":"ApiKey","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:38:47.775878592Z","method":"telemetry_requests","auth_type":"ApiKey","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:38:47.782785083Z","method":"get_collection_cluster","auth_type":"ApiKey","remote":"127.0.0.1","collection":"mj","result":"ok"}
{"timestamp":"2026-02-07T11:38:48.850584097Z","method":"list_collections","auth_type":"ApiKey","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:38:48.851509857Z","method":"list_collections","auth_type":"ApiKey","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:39:06.058730125Z","method":"list_collections","auth_type":"Jwt","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:39:06.059294963Z","method":"telemetry_cluster","auth_type":"Jwt","remote":"127.0.0.1","result":"denied","error":"Forbidden: Global access is required"}
{"timestamp":"2026-02-07T11:39:06.059325539Z","method":"telemetry_requests","auth_type":"Jwt","remote":"127.0.0.1","result":"denied","error":"Forbidden: Global access is required"}
{"timestamp":"2026-02-07T11:39:06.131026875Z","method":"list_collections","auth_type":"Jwt","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:39:06.132768220Z","method":"telemetry_cluster","auth_type":"Jwt","remote":"127.0.0.1","result":"denied","error":"Forbidden: Global access is required"}
{"timestamp":"2026-02-07T11:39:06.132855791Z","method":"telemetry_requests","auth_type":"Jwt","remote":"127.0.0.1","result":"denied","error":"Forbidden: Global access is required"}
{"timestamp":"2026-02-07T11:39:06.201205875Z","method":"list_collections","auth_type":"Jwt","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:39:06.202292016Z","method":"list_collections","auth_type":"Jwt","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:39:06.204635985Z","method":"list_aliases","auth_type":"Jwt","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:39:06.205305160Z","method":"list_aliases","auth_type":"Jwt","remote":"127.0.0.1","result":"ok"}
{"timestamp":"2026-02-07T11:39:06.207033235Z","method":"get_collection","auth_type":"Jwt","remote":"127.0.0.1","collection":"mj","result":"ok"}
{"timestamp":"2026-02-07T11:39:06.209684136Z","method":"get_collection","auth_type":"Jwt","remote":"127.0.0.1","collection":"mj","result":"ok"}
{"timestamp":"2026-02-07T11:39:07.630188759Z","method":"get_collection","auth_type":"Jwt","remote":"127.0.0.1","collection":"mj","result":"ok"}
{"timestamp":"2026-02-07T11:39:07.630418614Z","method":"strict_mode_check","auth_type":"Jwt","remote":"127.0.0.1","collection":"mj","result":"ok"}
{"timestamp":"2026-02-07T11:39:07.630498184Z","method":"scroll","auth_type":"Jwt","remote":"127.0.0.1","collection":"mj","result":"ok"}
{"timestamp":"2026-02-07T11:39:10.864667330Z","method":"strict_mode_check","auth_type":"Jwt","remote":"127.0.0.1","collection":"mj","result":"ok"}
{"timestamp":"2026-02-07T11:39:10.864786859Z","method":"delete_points","auth_type":"Jwt","remote":"127.0.0.1","collection":"mj","result":"denied","error":"Forbidden: Write access to collection mj is required"}
```

</details>

Initial implementation guides:

```
* `Claims` should have a new optional field `subject`, which can contain any string, provided by the client.

* Currently we propagate only the `Access` wrapped into `ActixAccess` or `extract_access` in tonic.
We would need to additionally propagate `subject` and the `remote` - the IP adress from the request.
New struct would be needed for that, call it `Auth`, and replace usage of `Access` with `Auth` in the codebase.

* In addition to `remote` and `subject`, `Auth` should also contain an `auth_type` enum, which can be `Jwt`, `ApiKey`, `None`.
Depending on how we authenticated the request.

* `Access` have methods like `check_global_access`, `check_collection_access`. 
Now `Auth` object should wrap those methods, but add logging layer to them.
In addition to regular function arguments, it should require also method name, as a `&str`. 

* Logging layer in `Auth` should be congigurable to global audit-log logger. It can be empty, then no logging will be done.

* Audit logger should be possible to configure to log into specific directory with log rotation and retention policies. Logs should be in JSON format.

* Initialization of global audit logger should be done in `main.rs` from the same place everything is configured - `Settings` structure. 
```
---

Refactoring of the https://github.com/qdrant/qdrant/pull/8028

```
Refactor for the use of `X-Forwarded-For` header.

Recently we introduced an extra config parameter called `trust_forwarded_headers`.
This parameter allows usage of `X-Forwarded-For` for logging source of the access.

But we don't actually intend this parameter for usign in regular logs.
It's original intent was to be used in audit logs, as we want to log the real source of the access, not the proxy IP address.

We would need to refactor `trust_forwarded_headers` in the following way:

- move the configuration parameter into the `AuditConfig`
- use the parameter only in the audit logger, use it to decide what to use in the `remote` field of the `Auth` struct.
- Move the logic of parsing `X-Forwarded-For` header into a separate files for both tonic and actix for better readability. Avoid deep parsing of the header, just put it as is into the `remote` field, as it can contain multiple IP addresses, and we don't want to mess with that. Just log it as is, and let the log analysis tools handle it.
- Restore the default behaviour for regular loggers, so the won't depend on `trust_forwarded_headers`

```

### DIFF
--- a/lib/storage/src/content_manager/collection_verification.rs
+++ b/lib/storage/src/content_manager/collection_verification.rs
@@ -27,11 +27,12 @@ where
     // Check access here first since strict-mode gets checked before `access`.
     // If we simply bypassed here, requests to a collection a user doesn't has access to could leak
     // information, like existence, strict mode config, payload indices, ...
-    let collection_pass = auth.check_collection_access(
-        collection_name,
-        AccessRequirements::new(),
-        "strict_mode_check",
-    )?;
+    //
+    // Strict mode have an unlogged access, as actual
+    // logging happens in the operations after strict mode check
+    let collection_pass = auth
+        .unlogged_access() // expected for strict mode check
+        .check_collection_access(collection_name, AccessRequirements::new())?;
     let collection = toc.get_collection(&collection_pass).await?;
     if let Some(strict_mode_config) = &collection.strict_mode_config().await
         && strict_mode_config.enabled.unwrap_or_default()


lib/storage/src/audit.rs
```diff
@@ -0,0 +1,185 @@
...
+    pub remote: Option<String>,
+    /// Collection name, if the check was collection‑scoped.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub collection: Option<String>,
+    /// `"ok"` when the access check passed, `"denied"` otherwise.
+    pub result: &'static str,
+    /// Error message when the access check failed.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub error: Option<String>,
+}
```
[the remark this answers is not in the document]
generall: Enabling audit logs are up the user + we do not expect service to be exposed to the public with that.

src/tonic/logging.rs
```diff
@@ -3,90 +3,24 @@ use std::task::{Context, Poll};
...
-        .get(&X_FORWARDED_FOR)
-        .and_then(|v| v.to_str().ok())
-        .and_then(parse_x_forwarded_for)
-}
-
-/// Parses the standard Forwarded header (RFC 7239) and extracts the client IP.
-/// Format: Forwarded: for=192.0.2.43;proto=https
-/// IPv6 format: Forwarded: for="[2001:db8::1]:4711"
-/// Note: RFC 7239 specifies parameter names are case-insensitive.
-fn parse_forwarded_header(header: &str) -> Option<String> {
```
timvisee, line 58: While I agree that this is complicated logic, I'm not too happy about removing all of it. In my opinion logging real IPs in access logs is still valuable.

lib/storage/src/audit.rs
```diff
@@ -0,0 +1,206 @@
...
+        // atomic message to the non-blocking writer (avoids interleaved
+        // partial writes from concurrent callers).
+        let mut buf = match serde_json::to_vec(event) {
+            Ok(buf) => buf,
+            Err(err) => {
+                log::error!("Failed to serialize audit log entry: {err}");
+                return;
+            }
+        };
+        buf.push(b'\n');
```
timvisee, line 146: This is not cross platform

src/actix/forwarded.rs
```diff
@@ -0,0 +1,28 @@
...
+
+/// Same as [`forwarded_for`] but accepts an [`HttpRequest`] (used in the
+/// extractor path where no `ServiceRequest` is available).
+pub fn forwarded_for_http(req: &HttpRequest) -> Option<String> {
+    forwarded_for_headers(req.headers())
+}
+
+fn forwarded_for_headers(headers: &actix_web::http::header::HeaderMap) -> Option<String> {
+    headers
+        .get(X_FORWARDED_FOR)
```
timvisee, line 32: In my opinion this is wrong, because it does not follow HTTP [standard](https://datatracker.ietf.org/doc/html/rfc7239) ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Forwarded)).

Ideally we read `Forwarded` first, and then fall back to `X-Forwarded-For`.

We do seem to handle it properly (with fallback) for gRPC.

lib/storage/src/audit.rs
timvisee: This would be good enough:

```suggestion
        #[cfg(not(windows))]
        buf.push(b'\n');
        #[cfg(windows)]
        buf.extend_from_slice(b"\r\n");
```

timvisee:
I do have a few concerns about this PR.

With a simple bfb upload **or** search load I can grow the audit file by multiple megabytes per second. I easily reach 1GB in a short amount of time, and have generated multiple millions of entries. I feel like this needs to me much better, possibly by logging less and/or by changing the lot format.

With the above in mind, we might also want to consider compressing rotated files. The current format would definitely benefit from that. Using `gzip` already reduces 117MB into 4.9MB.

The audit logging infrastructure doesn't seem very robust. For example, if I delete the audit file, Qdrant happily continues without any warnings or errors. Audit entries are written into the void and are permanently lost.

In the current implementation I do still see a lot of strict mode checks, I believe we want to get rid of them.

- Upsert with bfb:
  ```json
  {"timestamp":"2026-02-09T14:33:08.370406428Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:33:08.370406448Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:33:08.370438058Z","method":"upsert_points","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:33:08.370440228Z","method":"upsert_points","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:33:08.371099895Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:33:08.371107395Z","method":"upsert_points","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:33:08.371303954Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:33:08.371312434Z","method":"upsert_points","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:33:08.371424004Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  ```
- Query with bfb:
  ```json
  {"timestamp":"2026-02-09T14:34:12.690573046Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.690577576Z","method":"query_batch","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.690991134Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.690998224Z","method":"query_batch","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.691137854Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.691143074Z","method":"query_batch","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.691575662Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.691581072Z","method":"query_batch","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.691653822Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.691659562Z","method":"query_batch","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.692167740Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.692172990Z","method":"query_batch","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.692269470Z","method":"strict_mode_check","auth_type":"None","collection":"benchmark","result":"ok"}
  {"timestamp":"2026-02-09T14:34:12.692275459Z","method":"query_batch","auth_type":"None","collection":"benchmark","result":"ok"}
  ```

lib/storage/src/audit.rs
generall: do we want it to be cross-platform? Even if audit logs are created on windows, linux users should be able to read it.

src/tonic/logging.rs
generall: but we didn't bother to do it before the whole discussion about audit logging

src/actix/forwarded.rs
generall: :+1:

Issue #7489, opened 2025-10-31
Detach snapshot creation from ToC/collections, run in the background


timvisee:
A long running collection snapshot may block operations such as deleting a collection.

#### How to reproduce

1. create a large collection (1m points)
2. create a small collection (1 point)
3. create a snapshot of the large collection with `wait=false`
4. delete the small collection
5. the delete operation hangs (and times out) with the snapshot happening the background

Alternatively this may be reproduced without a large collection, by adding a sleep [here](https://github.com/qdrant/qdrant/blob/b8ac6741d457501b2e14e17e23f7f4ae0dfa2ce2/lib/storage/src/content_manager/toc/snapshots.rs#L67).

Related test: https://github.com/qdrant/qdrant/pull/7510

#### Problem

The problem is around this `RwLock` around our collections:
- https://github.com/qdrant/qdrant/blob/9aea6e1b95e4aa2c6594bdbd50c1387d5e867ea4/lib/storage/src/content_manager/toc/mod.rs#L59

When we create a snapshot we hold a read lock on it during the whole snapshot process:
- https://github.com/qdrant/qdrant/blob/b8ac6741d457501b2e14e17e23f7f4ae0dfa2ce2/lib/storage/src/content_manager/toc/snapshots.rs#L64-L70

When deleting a collection we require a write lock which is blocked:
- https://github.com/qdrant/qdrant/blob/b8ac6741d457501b2e14e17e23f7f4ae0dfa2ce2/lib/storage/src/content_manager/toc/collection_meta_ops.rs#L197

#### Expected result

I'd argue that it should process the collection delete quickly. I think it's fine if deleting the collection we're snapshotting hangs until the snapshot is done. But deleting a different collection should complete quickly.

This is quite significant in fact, because it currently fully blocks the consensus thread. Causing consensus instability. Because of this I classified the issue as bug.

#### Potential solution

Taking a snaphsot of a large collection can take a huge amount of time. It is quite exceptional compared to other operations.

We should investigate whether it is possible to detach the snapshot process from this lock, and let it complete in the background somehow. I'm sure it must be doable since we use `Arc`s a few levels deeper too. We can use a message channel to signal for completion. If this would prevent holding a read lock on collections for a long time, the problem would be resolved.

Alternatively we might double lock collections (the container and individual collections) `RwLock<Vec<Arc<RwLock<Collection>>>>`. But it might have other side effects.

amiremohamadi:
[...]

PR #8072, opened 2026-02-08
Avoid shards holder locks on snapshot

generall:
Fixes: <https://github.com/qdrant/qdrant/issues/7489>
Unblocks: <https://github.com/qdrant/qdrant/pull/7510>

[AI-free]

- **refactor shard.create_snapshot to own everything necessary in the future for snapshotting**
- **do not lock shards_holder and replica_set during snapshot**
- **separate locks and snapshot creation all the way down**

### DIFF
--- a/lib/collection/src/collection/snapshots.rs
+++ b/lib/collection/src/collection/snapshots.rs
@@ -93,30 +93,29 @@ impl Collection {
                         global_temp_dir.display(),
                     ))
                 })?;
-            let shards_holder = self.shards_holder.read().await;
 
             let mut futures = Vec::new();
-            // Create snapshot of each shard
-            for (shard_id, replica_set) in shards_holder.get_shards() {
-                let shard_snapshot_path = shard_path(Path::new(""), shard_id);
+            {
+                let shards_holder = self.shards_holder.read().await;
 
-                // If node is listener, we can save whatever currently is in the storage
-                let save_wal = self.shared_storage_config.node_type != NodeType::Listener;
-                let future = replica_set
-                    .create_snapshot(
-                        snapshot_temp_temp_dir.path(),
-                        tar.descend(&shard_snapshot_path)?,
-                        SnapshotFormat::Regular,
-                        None,
-                        save_wal,
-                    )
-                    .await?;
-                futures.push(future);
+                // Create snapshot of each shard
+                for (shard_id, replica_set) in shards_holder.get_shards() {
+                    let shard_snapshot_path = shard_path(Path::new(""), shard_id);
+
+                    // If node is listener, we can save whatever currently is in the storage
+                    let save_wal = self.shared_storage_config.node_type != NodeType::Listener;
+                    let future = replica_set
+                        .create_snapshot(
+                            snapshot_temp_temp_dir.path(),
+                            tar.descend(&shard_snapshot_path)?,
+                            SnapshotFormat::Regular,
+                            None,
+                            save_wal,
+                        )
+                        .await?;
+                    futures.push(future);
+                }
             }
-
-            // Do not lock shards_holder for the whole duration of snapshot creation,
-            // because it may take a long time, and we don't want to block other operations on the collection.
-            drop(shards_holder);
 
             for future in futures {
                 future.await.map_err(|err| {
--- a/lib/collection/src/shards/local_shard/snapshot.rs
+++ b/lib/collection/src/shards/local_shard/snapshot.rs
@@ -83,12 +83,11 @@ impl LocalShard {
         };
 
         let future = async move {
-            let tar_internal = tar_c.clone();
-
             if let Some(plunger_notify) = plunger_notify {
                 plunger_notify.await?;
             }
 
+            let tar_internal = tar_c.clone();
             let handle = tokio::task::spawn_blocking(move || {
                 // Do not change segments while snapshotting
                 snapshot_all_segments(


lib/collection/src/shards/replica_set/snapshots.rs
```diff
@@ -19,36 +19,42 @@ use crate::shards::shard_config::ShardConfig;
...
         temp_path: &Path,
-        tar: &tar_ext::BuilderExt,
+        tar: tar_ext::BuilderExt,
         format: SnapshotFormat,
         manifest: Option<SnapshotManifest>,
         save_wal: bool,
-    ) -> CollectionResult<()> {
-        let local_read = self.local.read().await;
-
+    ) -> impl Future<Output = CollectionResult<()>> + use<> {
```
generall: this is the main change

lib/collection/src/shards/shard_holder/mod.rs
```diff
@@ -1128,9 +1128,9 @@ impl ShardHolder {
         collection_name: &str,
         shard_id: ShardId,
         temp_dir: &Path,
-    ) -> CollectionResult<SnapshotDescription> {
+    ) -> CollectionResult<impl Future<Output = CollectionResult<SnapshotDescription>> + use<>> {
```
generall, line 1131: this is the second main change

lib/collection/src/shards/local_shard/snapshot.rs
```diff
@@ -46,66 +46,81 @@ impl LocalShard {
...
-    pub async fn create_snapshot(
+    pub async fn get_snapshot_creator(
         &self,
         temp_path: &Path,
         tar: &tar_ext::BuilderExt,
         format: SnapshotFormat,
         manifest: Option<SnapshotManifest>,
         save_wal: bool,
-    ) -> CollectionResult<()> {
+    ) -> CollectionResult<impl Future<Output = CollectionResult<()>> + use<>> {
```
generall, line 56: and this as well

lib/collection/src/shards/local_shard/snapshot.rs
```diff
@@ -46,66 +46,81 @@ impl LocalShard {
     }
 
     /// Create snapshot for local shard into `target_path`
-    pub async fn create_snapshot(
+    pub async fn get_snapshot_creator(
```
timvisee, line 49: Nit: convention would be to keep `create_snapshot` as name. The fact that we return a future tells us we need to await on it to complete the process.

generall: we are actually returitng future which returns future

timvisee: Ah, I missed that detail. All good 👍

ffuugoo: I would:

- keep `create_snapshot`
  - it's not that uncommon for a function to return a `Future`, esp if we need to "unbind" it from `self`
- *document* that it returns a `Future`
  - because, IMO, `get_snapshot_creator` is *not* a better name
  - I mean, @timvisee missed this detail just now
- **add `#[must_use]` attribute**

This gives you sane name and also ensures you can't easily forget to `await` on returned future.

lib/collection/src/shards/local_shard/snapshot.rs
ffuugoo, line 56: Would `+ 'static` work here? IMO, it's less weird/niche than `+ use<>`, I'd prefer to use `'static` if it works.

```suggestion
    ) -> CollectionResult<impl Future<Output = CollectionResult<()>> + 'static> {
```

generall: I would argue that `use<>` is more precise compare to `'static`. We declare that future doesn't capture any lifetimes.

ffuugoo: Maybe. Though, I think in this case both are equivalent. But also, why do we care? If we accidentally capture non-static lifetime or smth, we would get a compilation error anyway. IMO, rule of thumb is "use strictly necessary, least restrictive bound".

And also `'static` is better recognized than `use<>`. I'd prefer `'static` as long as it works.

lib/collection/src/shards/local_shard/snapshot.rs
ffuugoo: @generall 👆

lib/collection/src/collection/snapshots.rs
```diff
@@ -94,24 +94,34 @@ impl Collection {
...
-                    .map_err(|err| {
-                        CollectionError::service_error(format!("failed to create snapshot: {err}"))
-                    })?;
+                    .await?;
+                futures.push(future);
+            }
+
+            // Do not lock shards_holder for the whole duration of snapshot creation,
+            // because it may take a long time, and we don't want to block other operations on the collection.
+            drop(shards_holder);
```
ffuugoo: I'd restructure to use `{}` block instead of explicit `drop`. We had bugs due to explicit drops before.

```rust
let mut futures = Vec::new();

{
  // ...
}
```

lib/collection/src/shards/local_shard/snapshot.rs
```diff
@@ -46,66 +46,81 @@ impl LocalShard {
...
+            let tar_internal = tar_c.clone();
+
+            if let Some(plunger_notify) = plunger_notify {
+                plunger_notify.await?;
             }
-        });
-        AbortOnDropHandle::new(handle).await??;
 
-        LocalShardClocks::archive_data(&self.path, tar).await?;
+            let handle = tokio::task::spawn_blocking(move || {
```
ffuugoo: I'd move `tar_internal` closer to `tokio::task::spawn_blocking`. I was confused why do we need to `tar_c.clone()` for a second.

```suggestion
            if let Some(plunger_notify) = plunger_notify {
                plunger_notify.await?;
            }

            let tar_internal = tar_c.clone();
            let handle = tokio::task::spawn_blocking(move || {
```

lib/collection/src/shards/replica_set/snapshots.rs
```diff
@@ -22,33 +22,50 @@ impl ShardReplicaSet {
...
+            Some(
+                local
+                    .get_snapshot_creator(&temp_path, &tar, format, manifest, save_wal)
+                    .await?,
+            )
+        } else {
+            None
+        };
+
+        drop(local_read);
```
ffuugoo, line 52: 🙅‍♀️ no explicit drops 🚫

I'd also just call it `create_snapshot` instead of `maybe_local_snapshot_future`

```suggestion
        let maybe_local_snapshot_future = if let Some(local) = &*self.local.read().await {
            let create_snapshot = local
                .get_snapshot_creator(&temp_path, &tar, format, manifest, save_wal)
                .await?;
                
            Some(create_snapshot)
        } else {
            None
        };
```

lib/collection/src/shards/dummy_shard.rs
```diff
@@ -47,8 +47,8 @@ impl DummyShard {
         _format: SnapshotFormat,
         _manifest: Option<SnapshotManifest>,
         _save_wal: bool,
-    ) -> CollectionResult<()> {
-        self.dummy()
+    ) -> CollectionError {
+        self.dummy_error()
     }
```
ffuugoo, line 52: Why?

lib/collection/src/shards/replica_set/snapshots.rs
generall: this one is actually not necessary, and it will work the same way without drop. But I prefer to keep the drop to make it explicit.

lib/collection/src/shards/dummy_shard.rs
generall: simpler code

lib/collection/src/shards/replica_set/snapshots.rs
ffuugoo: `if let Some(_) = local.read() { ... }` is still simpler. 🤷‍♀️

lib/collection/src/shards/dummy_shard.rs
ffuugoo: Ah, yeah, I see. Then, maybe, remove `create_snapshot` at all and just do `dummy.dummy_error()` in that match? This does not make any sense at all.

lib/collection/src/shards/shard.rs
```diff
@@ -126,41 +127,43 @@ impl Shard {
...
+                    .get_snapshot_creator(temp_path, tar, format, manifest, save_wal)
+                    .await?,
+            )),
             Shard::Dummy(dummy_shard) => {
-                dummy_shard
+                return Err(dummy_shard
                     .create_snapshot(temp_path, tar, format, manifest, save_wal)
-                    .await
+                    .await);
             }
```
ffuugoo, line 163: Just do `return Err(dummy_shard.dummy_error())` and remove `DummyShard::create_snapshot` method entirely?

PR #8093, opened 2026-02-09
Dont lock WAL while serialization

IvanPleshkov:
This PR is a try to fix an empty update queue issue, where the upate queue is empty under pressure on some configurations.

Local tests didnt show differences but it seems this changes are nice to merge anyway.

The core idea is making WAL lock more efficient. Currently, we lock WAL for IO and for serialization.

Actually, it's not needed to lock the WAL for expensive serialization; this PR fixes it by introducing the serializable record entity,

### DIFF
--- a/lib/shard/src/wal.rs
+++ b/lib/shard/src/wal.rs
@@ -64,11 +64,9 @@ impl<R: DeserializeOwned + Serialize> WalRawRecord<R> {
         let record: R = serde_cbor::from_slice(record)
             .or_else(|_err| rmp_serde::from_slice(record))
             .map_err(|err| {
-                let error = format!(
+                WalError::WriteWalError(format!(
                     "Can't deserialize entry, probably corrupted WAL or version mismatch: {err:?}"
-                );
-                log::error!("{error}");
-                WalError::WriteWalError(error)
+                ))
             })?;
         Ok(record)
     }


lib/shard/src/wal.rs
```diff
@@ -31,6 +31,54 @@ const FIRST_INDEX_FILE: &str = "first-index";
 /// (this is used to extend recoverable history and allow WAL shard transfers)
 const INCREASED_RETENTION_FACTOR: usize = 10;
 
+pub struct WalSerializedRecord<R> {
```
timvisee: Nit: I'd like `WalRawRecord` a bit better

lib/shard/src/wal.rs
```diff
@@ -31,6 +31,54 @@ const FIRST_INDEX_FILE: &str = "first-index";
...
+    record: Vec<u8>,
+    _phantom: PhantomData<R>,
+}
+
+impl<R: DeserializeOwned + Serialize> WalSerializedRecord<R> {
+    pub fn new(record: &R) -> Result<Self> {
+        // ToDo: Replace back to faster rmp, once this https://github.com/serde-rs/serde/issues/2055 solved
+        let record = serde_cbor::to_vec(record).map_err(|err| {
```
timvisee, line 42: This vec seems to over allocate (by powers of two). But we only keep these in memory for a short period, don't we?

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -73,25 +73,27 @@ impl UpdateWorkers {
                     } else {
                         let record = wal.lock().await.read_single_record(op_num);
                         match record {
-                            Ok(Some(op)) => op.operation,
-                            Ok(None) => {
+                            Some(serialized_op) => match serialized_op.deserialize() {
+                                Ok(op) => op.operation,
+                                Err(err) => {
```
timvisee, line 105: Should we also log some error in stdout if we fail to deserialize?

timvisee:
It makes using the WAL slightly more difficult, but I understand the benefit. Not locking the WAL during (de)serialization can be significant with very large operations.

lib/shard/src/wal.rs
```diff
@@ -80,18 +126,14 @@ impl<R: DeserializeOwned + Serialize> SerdeWal<R> {
         }
     }
 
-    pub fn read_single_record(&self, idx: u64) -> Result<Option<R>> {
+    pub fn read_single_record(&self, idx: u64) -> Option<WalSerializedRecord<R>> {
```
agourlay: nit: I would prefer `read_raw_record`

as https://github.com/qdrant/qdrant/pull/8093/changes#r2792096551 is getting my vote

lib/collection/src/update_workers/update_worker.rs
timvisee: It's not necessary. I now notice it only returns `None` if the record is not found. Not when it fails to deserialize.

timvisee:
@IvanPleshkov there still seems to be a merge conflict.

PR #8095, opened 2026-02-10
HTTPS snapshot links

generall:
Currently, on the snapshot shard transfer we assumed that REST api have the same protocol as internal grpc, which is wrong.

This PR makes the protocol setting explicit.

### DIFF
--- a/lib/collection/src/shards/channel_service.rs
+++ b/lib/collection/src/shards/channel_service.rs
@@ -222,19 +222,16 @@ impl ChannelService {
                     "Cannot determine REST address, cannot specify port on address {url} for peer ID {this_peer_id}",
                 ))
             })?;
-        if self.rest_tls_enabled {
-            url.set_scheme("https").map_err(|()| {
-                CollectionError::service_error(format!(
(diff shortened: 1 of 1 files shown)


lib/collection/src/shards/channel_service.rs
```diff
@@ -217,6 +222,20 @@ impl ChannelService {
...
+                    "Cannot determine REST address, cannot set HTTPS scheme on address {url} for peer ID {this_peer_id}",
+                ))
+            })?;
+        } else {
+            url.set_scheme("http").map_err(|()| {
+                CollectionError::service_error(format!(
+                    "Cannot determine REST address, cannot set HTTP scheme on address {url} for peer ID {this_peer_id}",
+                ))
+            })?;
+        }
```
timvisee: I'd like this a bit more:

```suggestion
        let scheme = if self.rest_tls_enabled {
            "https"
        } else {
            "http"
        };
        url.set_scheme(scheme).map_err(|()| {
            CollectionError::service_error(format!(
                "Cannot determine REST address, cannot set {scheme} scheme on address {url} for peer ID {this_peer_id}",
            ))
        })?;
```

timvisee (approved):

PR #8097, opened 2026-02-10
fix: handle `score_threshold` in Formula queries

coszio:
As pointed out by @joein, we weren't handling `score_threshold` parameter in formula queries. This PR fixes it


generall:
@copilot  please generate integration test for this

PR #8099, opened 2026-02-10
Ignore cacert for internal requests if not configured

generall:


src/common/http_client.rs
```diff
@@ -87,6 +87,12 @@ fn https_client(
                     );
                 }
             }
+        } else if !verify_https_client_certificate {
+            // If ca_cert is not provided, and we are not verifying client certificate,
+            // there is no way to verify https connection.
```
timvisee, line 92: Doesn't it use the system provided CA bundle? Or do we omit it in our Docker images for example?

It should _just work_ without `add_root_certificate` assuming no custom certs are used.

generall: It is potentially possible, but if we are connecting to another Qdrant instance, we have to use our own certificate or don't use at all.

Also root certificates require explicit configuration

src/common/http_client.rs
```diff
@@ -87,6 +87,14 @@ fn https_client(
...
+            //
+            // So we have to disable certificate verification in order to be able to connect to the server.
+            builder = builder
+                .danger_accept_invalid_certs(true)
+                .danger_accept_invalid_hostnames(true);
```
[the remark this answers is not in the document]
generall: We don't and can't use well-known certificates to connect between nodes in cluster

src/common/http_client.rs
timvisee: Yeah. I missed that this branch still requires `verify_https_client_certificate: false` to be set. So it's all good!

PR #8104, opened 2026-02-11
Ensure WAL and shard clocks consistency when creating shard snapshot

ffuugoo:
Hold WAL lock while archiving shard clocks, so that WAL and clocks are always consistent.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/actix/api/debug_api.rs
+++ b/src/actix/api/debug_api.rs
@@ -1,14 +1,7 @@
 use actix_web::{Responder, get, patch, web};
-use collection::operations::verification;
-use collection::shards::shard::ShardId;
-use segment::types::SeqNumberType;
-use serde::{Deserialize, Serialize};
-use shard::operations::OperationWithClockTag;
-use storage::dispatcher::Dispatcher;
 use storage::rbac::AccessRequirements;
 
 use crate::actix::auth::ActixAuth;
-use crate::actix::helpers;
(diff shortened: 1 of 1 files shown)


src/actix/api/debug_api.rs
```diff
@@ -29,8 +36,94 @@ async fn update_debugger_config(
...
+    entries: u64,
+}
+
+impl Default for GetShardWalQuery {
+    fn default() -> Self {
+        Self { entries: 10 }
+    }
+}
+
+#[get("/collections/{collection}/shards/{shard}/wal")]
```
timvisee: We can only enable these APIs when having the `staging` feature enabled.

We currently bypass all authentication and so we cannot blindly expose this.

generall: I don't think it bypasses authentication. there is `check_collection_access`

generall: if you are looking at

```
let pass = verification::new_unchecked_verification_pass();
```

It is strict mode check, and yeah, we probably need a more explicit name for it

ffuugoo: Ah yes, sorry, I think maybe we should also add `manage` to the auth requirements? Or put under staging flag, I'm Ok with either.

ffuugoo: Done.

PR #8105, opened 2026-02-11
Limit number of Clocks

generall:
Clocks are used to track partial ordering of updates across whole cluster. Number of clocks limit the total update parallelism of individual peer, but this limitation seems to acceptable in order to prevent clogging state with too many clocks

### DIFF
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -222,26 +222,13 @@ impl ShardReplicaSet {
         self.replica_state.read().peers().keys().max().cloned()
     }
 
-    pub async fn get_clock_with_timeout(
-        &self,
-        timeout: Option<Duration>,
-    ) -> CollectionResult<ClockGuard> {
-        let start_time = std::time::Instant::now();
+    pub async fn get_clock(&self) -> ClockGuard {
         loop {
-            if let Some(clock) = self.clock_set.lock().await.get_clock() {
-                break Ok(clock);
+            match self.clock_set.lock().await.get_clock() {
+                Some(clock) => return clock,
+                // Prevent blocking async runtime with spinlock
+                None => yield_now().await,
             }
-            if let Some(timeout) = timeout
-                && start_time.elapsed() >= timeout
-            {
-                break Err(CollectionError::timeout(
-                    timeout,
-                    format!("Failed to acquire clock for update operation within {timeout:?}"),
-                ));
-            }
-
-            // Prevent blocking async runtime with spinlock
-            yield_now().await;
         }
     }
 
@@ -264,7 +251,15 @@ impl ShardReplicaSet {
         // - lock `remotes`, `local`, `clock` (in specified order!) on the *first* iteration of the loop
         // - then release and lock `remotes` and `local` *only* for all next iterations
         // - but keep initial `clock` for the whole duration of `update`
-        let mut clock = self.get_clock_with_timeout(timeout).await?;
+        let clock_timeout = timeout.unwrap_or(Duration::MAX);
+        let mut clock = tokio::time::timeout(clock_timeout, self.get_clock())
+            .await
+            .map_err(|_| {
+                CollectionError::timeout(
+                    clock_timeout,
+                    format!("Failed to acquire clock for update operation within {timeout:?}"),
+                )
+            })?;
 
         for attempt in 1..=UPDATE_MAX_CLOCK_REJECTED_RETRIES {
             let is_non_zero_tick = clock.current_tick().is_some();


lib/collection/src/shards/replica_set/update.rs
```diff
@@ -219,6 +220,26 @@ impl ShardReplicaSet {
...
+            if let Some(timeout) = timeout
+                && start_time.elapsed() >= timeout
+            {
+                break Err(CollectionError::timeout(
+                    timeout,
+                    format!("Failed to acquire clock for update operation within {timeout:?}"),
+                ));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
generall: Doesn't `self.clock_set.lock().await` yield?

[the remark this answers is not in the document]
ffuugoo: You can use [`yield_now().await`](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html) to explicitly yield between iterations.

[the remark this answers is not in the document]
ffuugoo: Also might be easier to structure like:
- `async fn get_clock(&self) -> ClockGuard` that just does loop on `get_clock`
- wrap it into `tokio::time::timeout` at the place of usage
	- `let mut clock = tokio::time::timeout(self.get_clock(), timeout).await?;`

generall:
> async fn get_clock(&self) -> ClockGuard

there are about 100 usage in sync tests, it would be a huge refactoring

ffuugoo:
> > async fn get_clock(&self) -> ClockGuard
> 
> there are about 100 usage in sync tests, it would be a huge refactoring

~~I mean, you literally changed it from `ClockGuard` to `Option<ClockGuard>` and added `unwrap` everywhere. Just revert. 🤷‍♀️~~

Nevermind the above comment. I'll explain shortly.

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -220,6 +222,27 @@ impl ShardReplicaSet {
...
+                && start_time.elapsed() >= timeout
+            {
+                break Err(CollectionError::timeout(
+                    timeout,
+                    format!("Failed to acquire clock for update operation within {timeout:?}"),
+                ));
+            }
+            yield_now().await;
+        }
+    }
```
ffuugoo: What I meant is, instead of `get_clock_with_timeout` do:

```suggestion
    pub async fn get_clock(&self) -> ClockGuard {
        loop {
            match self.clock_set.lock().await.get_clock() {
                Some(clock) => return clock,
                None => yield_now().await,
            }
        }
    }
```

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -239,7 +262,7 @@ impl ShardReplicaSet {
         // - lock `remotes`, `local`, `clock` (in specified order!) on the *first* iteration of the loop
         // - then release and lock `remotes` and `local` *only* for all next iterations
         // - but keep initial `clock` for the whole duration of `update`
-        let mut clock = self.clock_set.lock().await.get_clock();
+        let mut clock = self.get_clock_with_timeout(timeout).await?;
```
ffuugoo: And then here, you could do something like this to add a timeout:

```suggestion
        let mut clock = tokio::time::timeout(timeout.unwrap_or(Durtaion::MAX), self.get_clock()).await?;
```

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -238,6 +239,7 @@ impl ShardReplicaSet {
                     format!("Failed to acquire clock for update operation within {timeout:?}"),
                 ));
             }
+            yield_now().await;
```
timvisee: I'd even add an explicit comment since blocking this async thread is much worse than spin locking on a regular system thread.

```suggestion

            // Prevent blocking async runtime with spinlock
            yield_now().await;
```

lib/collection/src/shards/replica_set/update.rs
timvisee: I agree, it makes sense to do it this way. It's a future and it's cancel safe.

timvisee: https://github.com/qdrant/qdrant/pull/8105/commits/e5893bce1ad3666c05932ecacad0a2242bf83089

lib/collection/src/shards/replica_set/update.rs
timvisee: https://github.com/qdrant/qdrant/pull/8105/commits/e5893bce1ad3666c05932ecacad0a2242bf83089

lib/collection/src/wal_delta.rs
```diff
@@ -1454,7 +1454,13 @@ mod tests {
...
 
             // Release some kept clocks
-            kept_clocks.retain(|(keep_for, _)| *keep_for > 1);
+            kept_clocks.retain_mut(|(keep_for, _)| {
+                if *keep_for == 0 {
+                    return false;
+                }
+                *keep_for -= 1;
+                true
+            });
```
timvisee, line 1463: This artificial release mechanism in the test was broken, as it did not decrease the counter.

I've also swapped the number of rounds we keep locks for in the above two loops. Now the smaller loop keeps clocks for longer, while the bigger loop keeps locks for shorter. This way we stay below the specified maximum of 256 clocks and prevent the test panicking.

I believe it's a reasonable way to make the test happy.

lib/collection/src/shards/replica_set/update.rs
```diff
@@ -220,6 +222,16 @@ impl ShardReplicaSet {
         self.replica_state.read().peers().keys().max().cloned()
     }
 
+    pub async fn get_clock(&self) -> ClockGuard {
+        loop {
+            match self.clock_set.lock().await.get_clock() {
+                Some(clock) => return clock,
+                // Prevent blocking async runtime with spinlock
+                None => yield_now().await,
```
monoid, line 230: We may have a condvar that we can await on.

timvisee: There is many clocks. We should not wait with a condvar on all of them.

Having a separate atomic or notification channel is possible, but it makes the structure more complicated.

We don't expect to hit this branch under normal circumstances. Only when an excessive amount of parallelism is used, which is a problem on its own.

PR #8108, opened 2026-02-11
Fix update queue length aggregation

agourlay:
The queue length reported in the collection info needs to be aggregated across shards.

The PR fixes it by summing the lengths.

Interestingly, switching to a deconstruction approach highlighted a missing aggregation for the `warnings` field.

### DIFF
--- a/lib/collection/src/collection/collection_ops.rs
+++ b/lib/collection/src/collection/collection_ops.rs
@@ -334,19 +334,33 @@ impl Collection {
         };
 
         while let Some(response) = requests.try_next().await? {
-            info.status = cmp::max(info.status, response.status);
-            info.optimizer_status = cmp::max(info.optimizer_status, response.optimizer_status);
+            let CollectionInfo {
+                status,
+                optimizer_status,
+                warnings,
+                indexed_vectors_count,
+                points_count,
+                segments_count,
+                config: _,
+                payload_schema,
+                update_queue,
+            } = response;
+            info.status = cmp::max(info.status, status);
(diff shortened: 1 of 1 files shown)


lib/collection/src/collection/collection_ops.rs
```diff
@@ -334,19 +334,33 @@ impl Collection {
...
-            info.points_count = info
-                .points_count
-                .zip(response.points_count)
-                .map(|(a, b)| a + b);
-            info.segments_count += response.segments_count;
-
-            for (key, response_schema) in response.payload_schema {
+            info.points_count = info.points_count.zip(points_count).map(|(a, b)| a + b);
+            info.segments_count += segments_count;
+            info.warnings.extend(warnings);
```
agourlay, line 356: `warnings` were not aggregated 🙃

lib/collection/src/collection/collection_ops.rs
```diff
@@ -334,19 +334,33 @@ impl Collection {
...
+            if let Some(queue) = &mut info.update_queue {
+                // TODO remove field `op_num`, no sane way to aggregate across shards
```
timvisee, line 358: We can remove this TODO now can we?

Or does this comment mean we should remove it form the queue structure itself.

agourlay: Yes, I still need to remove the field itself in the API and restructure things.
Planned for the next PR.

timvisee: Handled in <https://github.com/qdrant/qdrant/pull/8127>

PR #8109, opened 2026-02-11
[ssd cache] Implement caching

coszio:
Supercedes/Builds on top of #8079 

## WIP
This is a cooperative PR from @xzfc and me which implements the first POC of local ssd caching. 

More details to come.


coszio:
Not committing to this implementation for now

Issue #8106, opened 2026-02-11
Merge Python Dependabot PR commits into a single PR


agourlay:
Merge the following Python Dependabot PR commits into a single PR targeting the dev branch.

- https://github.com/qdrant/qdrant/pull/8089
- https://github.com/qdrant/qdrant/pull/8090
- https://github.com/qdrant/qdrant/pull/8091
- https://github.com/qdrant/qdrant/pull/8092

agourlay:
https://github.com/qdrant/qdrant/pull/8119

PR #8119, opened 2026-02-12
Update Python dependencies: urllib3, protobuf, werkzeug, pyasn1

agourlay:
Publishing Copilot's [work](https://github.com/qdrant/qdrant/pull/8117) for https://github.com/qdrant/qdrant/issues/8106

It seems the agent can't target `dev` on its own.


xzfc (approved):

agourlay:
> I think it would be simpler to run uv --project tests lock --upgrade rather than ask an AI to merge four automated pull-requests.

I agree, wanted to give Copilot a try but that was not a good task for it :+1:

PR #8120, opened 2026-02-13
distributed optimization info

generall:
This PR makes collection optimizations API functionality to query the whole cluster.

Task prompted to AI:

```
# Distributed optimization info

Currently `/collections/{name}/optimizations` API reads optimizations only from the local node.
In cluster mode with remote shards it meands that API response will depend on the node which receives the request.
It is not very convenient in cases, when API results are visualized in UI.


We need to make this API collect optimizations from all shards in the cluster, including remote ones.

## The plan:

- refactor `pub async fn optimizations` inside `lib/collection/src/collection/collection_ops.rs`:
    - currently it generates `OptimizationsResponse` based on individual shards. We need to change that, so each shard generates `OptimizationsResponse` and collection level function merges them togheter.


- On the shard level in `lib/collection/src/shards/replica_set/mod.rs`, there are
    - `pub async fn optimizers_log(&self) -> Option<Arc<ParkingMutex<TrackerLog>>>`
    - `pub async fn optimizations(&self) -> Option<LocalShardOptimizations>`

we need the following changes there:

- make it into one function that returns `OptimizationsResponse`.
- work with loced sctrucutres like `Option<Arc<ParkingMutex<TrackerLog>>>` should not be exposed in the public interface of the replica set.


- `pub async fn optimizations` should be implemented for `RemoteShard` as well, so we can get optimizations from remote shards.
  - In order to do that, we would require:
    - Create a new internal grpc API in `lib/api/src/grpc/proto/collections_internal_service.proto`, that will be used to get optimizations from remote shards.
    - the API should allow to retrieve optimizations for a specific shard.
    - Follow the example of other internal API to implement it
```

Followup fixes:

```
- OptimizationsResponse.merge method could benefit from deconstruction of the objects, to make it impossile to forget to use any fields.
- Same for `options: OptimizationsRequestOptions` in `optimizations` method
- `pub async fn optimizations(` method in the ReplicaSet doesn't return error, only `OptimizationsResponse`. But there should be a possibility of an error when talking to remote shards. Error is incorrectly ignored, it should be propagated all the way to the top level API.
- `// Collect from remote shards` loop and `local_optimizations.await` in the replica set should use try_join_all instead of iterating one-by-one
- in the `merge` of `OptimizationsResponse`, check for `Some` in optional fields discards values from `other` if the current object has `None` in that field. That might lead to loss of data.
```

### DIFF
--- a/lib/collection/src/shards/replica_set/mod.rs
+++ b/lib/collection/src/shards/replica_set/mod.rs
@@ -1155,6 +1155,16 @@ impl ShardReplicaSet {
         is_readable && !self.is_locally_disabled(peer_id)
     }
 
+    fn peer_is_updatable(&self, peer_id: PeerId) -> bool {
+        let peer_state = self.peer_state(peer_id);
+        let is_updatable = match peer_state {
+            Some(state) => state.is_updatable(),
+            None => false,
+        };
+
+        is_updatable && !self.is_locally_disabled(peer_id)
+    }
+
     /// Check if this shard is active.
     /// By active, we mean, that at least one replica have `is_active` state.
     /// It is possible, that some replicas are not active, if they are created in a `Partial` state.
@@ -1391,6 +1401,7 @@ impl ShardReplicaSet {
             let remotes = self.remotes.read().await;
             let futures: Vec<_> = remotes
                 .iter()
+                .filter(|remote| self.peer_is_updatable(remote.peer_id))
                 .map(|remote| remote.optimizations(options))
                 .collect();
             futures::future::try_join_all(futures).await
@@ -1443,7 +1454,12 @@ impl ShardReplicaSet {
         let mut idle_segments_count = 0;
 
         let local = self.local.read().await;
-        if let Some(shard) = local.as_ref() {
+
+        let is_updatable = self.peer_is_updatable(self.this_peer_id());
+
+        if let Some(shard) = local.as_ref()
+            && is_updatable
+        {
             if let Some(log) = shard.optimizers_log() {
                 for tracker in log.lock().iter() {
                     if tracker.state.lock().status.is_running() {


lib/api/src/grpc/proto/collections_internal_service.proto
```diff
@@ -79,3 +82,23 @@ message UpdateShardCutoffPointRequest {
...
+  // Whether to include idle segments
+  bool with_idle_segments = 5;
+}
+
+message GetShardOptimizationsResponse {
+  // JSON-serialized OptimizationsResponse
+  bytes optimizations_json = 1;
+  // Time spent to process
+  double time = 2;
+}
```
generall, line 104: About the idea to use JSON field inside gRPC to transfer.

From one point of view, it is not how we usually utilize our gRPC APIs. For example 
`telemetry_internal` is strongly typed. Likely, the json overlay introduce some (small) overhead, compare to the native gRPC serialization.

On the other hand:

- `telemetry_internal.proto` is VERY verbose, it requires thinking about reserved fields for future use.
- Our APIs are internal, therefore we can serialize and deserialize with exactly same structure.
- Avoid huge boilerplate of converters.
- Potentially easier backward compatibility.

I am leaning towards the JSON field option.

lib/collection/src/collection/collection_ops.rs
```diff
@@ -449,67 +449,53 @@ impl Collection {
...
             summary: OptimizationsSummary {
-                queued_optimizations,
-                queued_segments,
-                queued_points,
-                idle_segments: idle_segments_count,
+                queued_optimizations: 0,
+                queued_segments: 0,
+                queued_points: 0,
+                idle_segments: 0,
             },
```
xzfc: **nipick (simplify):** why initializing it to `None` only to `unwrap_or_else` it later?

Suggestion to drop `Option`.

```diff
-let mut merged: Option<OptimizationsResponse> = None;
+let mut merged = OptimizationsResponse { /* some defaults */ };

-let mut response = merged.unwrap_or_else(|| OptimizationsResponse {
-    /* some defaults */
-});
```

lib/api/src/grpc/proto/collections_internal_service.proto
xzfc: At this point, why not to call REST API instead?

This way we don't even need to query per shard. (instead, for the whole collection)

lib/collection/src/collection/mod.rs
```diff
@@ -561,6 +564,26 @@ impl Collection {
...
+        options: OptimizationsRequestOptions,
+    ) -> CollectionResult<OptimizationsResponse> {
+        let shard_holder_read = self.shards_holder.read().await;
+
+        let shard = shard_holder_read.get_shard(shard_id);
+        let Some(replica_set) = shard else {
+            return Err(CollectionError::NotFound {
+                what: format!("Shard {shard_id}"),
+            });
+        };
```
xzfc: **nitpick (simplify):** use `ok_or_else`, `CollectionError::not_found()`, and `?`.

xzfc:
It throws an error if there are dead shards. Should we instead ignore dead peers in the cluster?

```
$ curl 127.1:6333/collections/benchmark/optimizations  
{"status":{"error":"Service internal error: Tonic status error: status: Unavailable, message: \"Failed to connect to http://127.0.0.1:6375/, error: transport error\", details: [], metadata: MetadataMap { headers: {} }"},"time":0.435550089}
```

---

Also, would it make sense to add `?local_only=true` query parameter? (or `?include_remote=true`)

lib/api/src/grpc/proto/collections_internal_service.proto
generall: because we don't know how the rest interface is exposed. We only do it for snapshot transfers and only because it is not possible to use grpc for file streaming. It also requires explicit url generation on the source side and usage of the public certificate

generall:
> Should we instead ignore dead peers in the cluster?

I think you are right, ignoring non-healthy shards is good idea.

> , would it make sense to add ?local_only=true query parameter? (or ?include_remote=true)

I don't think I see a use-case for it right now, maybe later

coszio:
> This merges distributed metrics into the existing /collections/{name}/optimizations. Doesn't it make more sense to use a dedicated endpoint for it? Maybe just /cluster/telemetry?

My thoughts as well, although in cluster/telemetry we don't want the level of detail of this API, probably just indexing status and progress for each shard

generall:
I would prefer not to create a universal APIs which report everything unless it is needed.
`/telemtery` is needed for cluster manager, so it doesn't make a dozen calls all the time

PR #8121, opened 2026-02-13
Do not create a dedicated runtime for Actix

agourlay:
We spawn many threads on startup, an idle Qdrant has 84 threads on my 24 cores machine.

This PR proposes to not create a dedicated Tokio runtime for actix but simply use the general runtime like for gRPC.

We still create the same amount of workers with `.workers(max_web_workers(&settings));` but the additional runtime is gone.

Disclaimer: I see no difference in performance but also no difference in thread count :man_shrugging:

### DIFF
--- a/src/actix/mod.rs
+++ b/src/actix/mod.rs
@@ -22,6 +22,7 @@ use collection::operations::validation;
 use collection::operations::verification::new_unchecked_verification_pass;
 use storage::dispatcher::Dispatcher;
 use storage::rbac::{Access, Auth};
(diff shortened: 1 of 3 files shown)


src/actix/mod.rs
timvisee, line 174: This is actually the place where we start actix workers.

https://docs.rs/actix-web/latest/actix_web/struct.HttpServer.html#method.workers

agourlay:
I can't prove that this PR is beneficial in all honesty.

I will close it for the time being.

PR #8128, opened 2026-02-13
Fix `too_many_internal_resets` error

ffuugoo:
Patch `tonic` and `hyper` crates to expose `max_local_error_reset_streams`, and *disable* it when creating internal gRPC connections.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3089,8 +3089,7 @@ checksum = "135b12329e5e3ce057a9f972339ea52bc954fe1e9358ef27f95e89716fbc5424"
 [[package]]
 name = "hyper"
 version = "0.14.26"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ab302d72a6f11a3b910431ff93aae7e773078c769f0a3ef15fb9ec692ed147d4"
+source = "git+https://github.com/qdrant/hyper?branch=v0.14.26-qdrant#923b57d062119b6d223beff7f34f8955f9c3dded"
 dependencies = [
  "bytes",
  "futures-channel",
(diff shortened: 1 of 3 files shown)


timvisee:
Some more context for the future :mechanical_arm: :robot: :mechanical_leg: if we hit this again:

We hit a security feature here that helps us clean up borked connections. Since this is for internal cluster communication and we manage both sides of the connection we can disable this security feature, which is what this PR takes care of.

We mainly hit this because we drop connections before waiting on the result. We count an error when we drop a connection before we receive it's response, and eventually hit a limit. This happens a lot in fanning out reads which races with the local replica. Forcefully dropping these connections once we get a result is much easier and fits better with our implementation than waiting for all responses and dropping connections gracefully.

agourlay:
How can this fix be validated? :)

ffuugoo:
> How can this fix be validated? :)

```bash
bfb \
    --uri $QDRANT_HOST \
    -n 10M \
    -d 512 \
    --skip-setup \
    --search \
    --keywords 5000 \
    --rps 300
```

@generall used this collection setup (not sure if critical, though, search is what repro the bug)

```bash
bfb \
    --uri $QDRANT_HOST \
    -n 10M \
    -d 512 \
    --shards 9 \
    --replication-factor 2 \
    --on-disk-vectors true \
    --keywords 5000 \
    --hnsw-m 0 \
    --hnsw-payload-m 16 \
    --tenants true \
    -b 10 \
    --timeout 60 \
    --rps 100
```

generall:
important detail is fan-out-factor

ffuugoo:
> important detail is fan-out-factor

Repro with default setup, without any explicit fan-out factor setting on my machine.

Behavior with and without fix is different:
- without the fix `too_many_internal_resets`
  - note "internal"
  - that's what we observed recently and what we expected from this test
- with the fix there's still error, but it's `too_many_resets`
  - no "internal"
  - it's abort-reset-streams error, which is different
  - I'd consider this as indication that "fix works", because "internal" error goes away as expected

generall:
It seems to fix problem on my repro setup

generall (approved):

PR #8129, opened 2026-02-13
Remove resharding version check, don't require 1.16.4-dev or up

timvisee:
Closes <https://github.com/qdrant/qdrant/pull/8123>

Remove the resharding transfer version check implemented in <https://github.com/qdrant/qdrant/pull/7963> that required all peers to be 1.16.4-dev or up.

On reconsideration I believe the check is unnecessary. It has also been causing more problem than it being beneficial.

A few reasons on why I think it isn't useful:
- nothing in (resharding) transfers has changed to deem this version check, it's actually hash ring routing that changed instead
- cluster manager does not trigger resharding transfers if there is a version mismatch, and so we should never dissatisfy the check in practice
- it potentially breaks resharding transfers in the future if a new peer joins which hasn't submitted their version to consensus jet

Removing this check also means <https://github.com/qdrant/qdrant/pull/8016> wasn't strictly necessary. Though I do vote to keep that as I'd consider it a better way to check versions anyway since more components rely on consensus distributed versions.

### All Submissions:

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -1,12 +1,11 @@
 use std::future::Future;
 use std::path::PathBuf;
-use std::sync::{Arc, LazyLock};
+use std::sync::Arc;
 use std::time::Duration;
 
 use common::defaults;
 use fs_err::tokio as tokio_fs;
 use parking_lot::Mutex;
-use semver::Version;
 use tokio_util::task::AbortOnDropHandle;
 
 use super::Collection;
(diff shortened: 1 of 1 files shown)


KShivendu:
> nothing in (resharding) transfers has changed to deem this version check, it's actually hash ring routing that changed instead

There's another difference. UpdateOnly skips inserts, Upsert doesn't. This can cause data consistency bugs.

> cluster manager does not trigger resharding transfers if there is a version mismatch, and so we should never dissatisfy the check in practice

Yeah but it only checks for version being same. Some resharding transfers might happen when version is below `v1.16.4-dev` and some might happen when it's above it.

timvisee:
> There's another difference. UpdateOnly skips inserts, Upsert doesn't. This can cause data consistency bugs.

That is not part of transfers, is it?

> Yeah but it only checks for version being same. Some resharding transfers might happen when version is below v1.16.4-dev and some might happen when it's above it.

That is not an issue. Both versions handle internal updates (through the hash ring) in different ways, but that mechanism is local and the net result is the same.

PR #8131, opened 2026-02-14
prevent proxy locking while snapshotting

generall:
This PR rolls back changes of https://github.com/qdrant/qdrant/pull/7196

### Why it is problematic?

- Proxy segment was holding a read-lock for the duration of the snapshot, which can be very long.
- Any update operation had to lock proxy segment for write, was stuck for the duration of the snapshot.
- with write operations, all other operation were stuck as well

### Why it should be ok not to hold the lock?

Proxy segments are design explicitly for this situation - allow read + write operations, while underlying segment is accessed for read for a long time. This is how optimization functions.
Requiring it to be locked for the same duration as wrapped segment essentially kills the idea of proxy segment.

---

~Extra changes of a new updates lock for the duration of proxifying didn't show any improvement, but I think it might be good to have anyway - it prevents update operation from reading/writing directly into wrapped segments after they are already proxified.~

:point_up: this doesn't work because of deadlock on parallel snapshots

### DIFF
--- a/lib/collection/src/shards/local_shard/snapshot.rs
+++ b/lib/collection/src/shards/local_shard/snapshot.rs
@@ -316,9 +316,6 @@ pub fn proxy_all_segments_and_apply<F>(
 where
     F: FnMut(&RwLock<dyn NonAppendableSegmentEntry>) -> OperationResult<()>,
 {
-    // Prevents update operations to accidentally lock on wrapped segments.
-    let update_lock = segments.acquire_updates_lock();
-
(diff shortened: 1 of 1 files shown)


timvisee (approved):
That's a subtle change! Looks good. I see we now only hold a lock on the segment to clone the wrapped segment for a very short period of time.

I did not re-test the deadlock locally, and am assuming you've done that already.

generall:
> I did not re-test the deadlock locally, and am assuming you've done that already.

yes, it is enough to just send update requests to block everything

PR #8132, opened 2026-02-14
Expose shard snapshot API

generall:
[AI]

```
generate API description in openapi/openapi-snapshots.ytt.yaml for #[get("/collections/{collection}/shards/{shard}/snapshot")]
```


generall:
> Or is it a requirement to work with edge?

it is required for edge. Also, we already expose shard snapsot API, but just not the streaming one.

PR #8133, opened 2026-02-15
fix: use c_char for Vulkan pointers to support ARM64

### LINKED ISSUE #8096: GPU feature fails to compile on ARM64 (aarch64) - c_char type mismatch
## Description

Building Qdrant with GPU support (`--build-arg GPU=nvidia`) fails on ARM64 (aarch64) systems due to a `c_char` type mismatch in the `gpu` crate.

## Environment

- **Platform**: NVIDIA DGX Spark (Grace Blackwell, ARM64/aarch64)
- **OS**: Ubuntu 24.04 ARM64
- **Qdrant version**: latest master (cloned 2026-02-10)
- **Build command**: `docker build -t qdrant-arm --build-arg GPU=nvidia .`

## Error

```
error[E0277]: a value of type `Vec<*const i8>` cannot be built from an iterator over elements of type `*const u8`
  --> lib/gpu/src/...
     |
     |  .map(|raw_name| raw_name.as_ptr())
     |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Iterator::Item` changed to `*const u8` here

error[E0308]: mismatched types
  --> /usr/local/cargo/registry/src/.../ash-0.38.0+1.3.281/src/vk/definitions.rs:1368:12
     |
1368 |     pub fn enabled_extension_names(mut self, enabled_extension_names: &'a [*const c_char]) -> Self {
     |            ^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `gpu` (lib) due to 6 previous errors
```

## Root Cause

On ARM64, `c_char` is `u8`, while on x86_64 it is `i8`. The `ash` Vulkan bindings expect `*const i8` (`*const c_char` on x86_64), but on ARM64 the iterator produces `*const u8` values, causing type mismatches.

This means the GPU feature currently only compiles on x86_64 platforms.

## Expected Behavior

The GPU feature should compile on ARM64 (aarch64) platforms, especially given NVIDIA's ARM-based GPU platforms like DGX Spark (Grace Blackwell GB10).

## Suggested Fix

Use explicit `c_char` casts in the GPU crate where `CStr::as_ptr()` results are collected, e.g.:

```rust
.map(|raw_name| raw_name.as_ptr() as *const std::ffi::c_char)
```

aviu16:
## Summary
Fixes #8096

This PR fixes compilation failures when building Qdrant with GPU support on ARM64 (aarch64) architecture.

## The Problem
On ARM64, `c_char` is defined as `u8`, while on x86_64 it is `i8`. The code was using hardcoded `*const i8` types for Vulkan string pointers, which caused type mismatches when compiling on ARM64:

```
error[E0308]: mismatched types
expected `*const i8`, found `*const u8`
```

Vulkan bindings expect `*const c_char` pointers, which resolves to the correct platform-specific type.

## The Solution
Replace all instances of `Vec<*const i8>` with `Vec<*const std::ffi::c_char>` in:
- `lib/gpu/src/instance.rs`: `extension_names_raw` (line 144)
- `lib/gpu/src/instance.rs`: `layers_raw` (line 157)
- `lib/gpu/src/device.rs`: `extension_names_raw` (line 228)

This ensures the code compiles correctly on both x86_64 (where `c_char = i8`) and ARM64 (where `c_char = u8`).

## Testing
This fix follows the standard Rust FFI pattern for C string pointers and is consistent with the Vulkan ash crate's API which expects `*const c_char`.

### DIFF
--- a/lib/gpu/src/device.rs
+++ b/lib/gpu/src/device.rs
@@ -225,7 +225,7 @@ impl Device {
             vk_physical_device.vk_physical_device,
             &extensions_cstr,
         )?;
-        let extension_names_raw: Vec<*const i8> = extensions_cstr
+        let extension_names_raw: Vec<*const std::ffi::c_char> = extensions_cstr
             .iter()
             .map(|raw_name| raw_name.as_ptr())
             .collect();
--- a/lib/gpu/src/instance.rs
+++ b/lib/gpu/src/instance.rs
@@ -141,7 +141,7 @@ impl Instance {
             .iter()
             .filter_map(|s| CString::new(s.clone().into_bytes()).ok())
             .collect();
-        let extension_names_raw: Vec<*const i8> = extensions_cstr
+        let extension_names_raw: Vec<*const std::ffi::c_char> = extensions_cstr
             .iter()
             .map(|raw_name| raw_name.as_ptr())
             .collect();
@@ -154,7 +154,7 @@ impl Instance {
             .iter()
             .filter_map(|s| CString::new(s.clone().into_bytes()).ok())
             .collect();
-        let layers_raw: Vec<*const i8> = layers_cstr
+        let layers_raw: Vec<*const std::ffi::c_char> = layers_cstr
             .iter()
             .map(|raw_name| raw_name.as_ptr())
             .collect();


aviu16:
closing this for now, gonna revisit later

PR #8139, opened 2026-02-15
Improve port logging in case port 0 requested

dgel:
When port 0 is requested a random free port is assigned, but log statements will show qdrant is listening on port 0. This MR fixes the log statements to show the assigned port instead

### All Submissions:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5481,6 +5481,7 @@ dependencies = [
  "tikv-jemalloc-ctl",
  "tikv-jemallocator",
  "tokio",
+ "tokio-stream",
  "tokio-util",
  "tonic 0.11.0",
  "tonic-reflection",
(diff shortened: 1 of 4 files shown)


xzfc:
What's your use-case on running it on auto-assigned port?

dgel:
I want to run potentially multiple instances on the same server for testing purposes, and since I already assign other services random ports for the same purpose, this would have made it easier to do so.

However, it appears CodeRabbit is correct that the current implementation would turn off TLS, and there also appears no way to run the tonic code as before and retrieve the assigned port afterwards. The only workaround seems be to use TlsIncoming, but that requires manually configuring Rustls and 1. I do not feel confident I would do so correctly and 2. it seems too much work for such a small improvement, so I'm inclined to just close the MR

PR #8140, opened 2026-02-15
Lock free clock set

IvanPleshkov:
After this PR https://github.com/qdrant/qdrant/pull/8105 it's possible to make `ClockSet` lock-free using atomic len.
This PR makes `ClockSet` lock free and removes tokio mutex.


generall (changes requested):
Are there any evidences that mutex is a problem here?

IvanPleshkov:
Close this PR because I don't have evidence of a measured performance boost

Issue #3322, opened 2024-01-04
Per-collection metrics for Prometheus


generall:
**Is your feature request related to a problem? Please describe.**

Currently, all metrics in `/metrics` are global, meaning that it’s impossible to see differences per collection.

In addition to that, all our metrics should have per-collection granularity to allow better aggregation in Prometheus, including:

- point/vector counts
- REST/gRPC requests

**Describe the solution you'd like**

Example:
```
grpc_responses_min_duration_seconds{endpoint="/qdrant.Points/Upsert"} 0.000046
grpc_responses_min_duration_seconds{endpoint="/qdrant.Points/Upsert",collection="my-collection"} 0.000049
grpc_responses_min_duration_seconds{endpoint="/qdrant.Points/Upsert",collection="my-collection-2"} 0.000046
```

**Describe alternatives you've considered**

Create dedicated endpoint for each collection `/collections/my-collecton/metrics`
but feedback from DevOps on this idea was negative.

**Additional context**

It might be beneficial to allow users to disable per-collection output. It is especially relevant if there are a lot of collections and metric response could become huge. But this is a nice-to-have requirement.

---

Note for contributors: Please consider this as tracking issue. If you think that it would be beneficial to split the task into multiple smaller PRs, please you are welcome to do so. Bounty will be rewarded for each PR independently

xzfc:
Currently, the `/metrics` collection is strongly coupled with the `/telemetry` collection. The `/metrics` endpoint is just a filtered-out telemetry metrics. (filtering happens in `MetricsProvider::add_metrics` trait method).
Does the `/telemetry` endpoint need to be updated as well?

I see multiple options to implement this.
1. Decouple metrics and telemetry collection. This would be more performant since we can filter out unwanted metrics right away instead of collecting all of the possible data and then filtering it out on each request.
2. Extend telemetry collection logic and also store collection name. However, to keep REST API stable we need to aggregate it back during the `GET /telemetry` request.

generall:
I the original design, the idea was that /metrics is just a different format for /telemetry. So I think extending telemetry is fine.
One thing to be careful about - anonymization of the collection names in the telemetry. I think anonymized stats should be aggregated across all collections always

gabrieldemian:
/attempt #3322

What metrics does make sense to add for single collections?

guhitb:
[...]

PR #8141, opened 2026-02-16
Add per-collection REST request metrics for Prometheus

omair445:
Hey! This is a first step towards per-collection metrics for Prometheus (ref #3322).

## What this does

Adds a `collection` label to REST request duration/count metrics so you can break down request performance per collection in your Prometheus dashboards.

The collection name gets extracted from the URL path (`/collections/{name}/...`) in the actix telemetry middleware, so it works for all whitelisted REST endpoints without needing changes to individual handlers.

### Metrics output example

Global metrics (unchanged, no breakage):
```
rest_responses_total{method="POST",endpoint="/collections/{name}/points/search",status="200"} 150
```

New per-collection metrics:
```
rest_collection_responses_total{method="POST",endpoint="/collections/{name}/points/search",status="200",collection="my-collection"} 100
rest_collection_responses_total{method="POST",endpoint="/collections/{name}/points/search",status="200",collection="my-other-collection"} 50
```

Same goes for `_avg_duration_seconds`, `_min_duration_seconds`, `_max_duration_seconds`, and the duration histogram.

## Design decisions

- **Separate prefix** (`rest_collection_`) instead of adding collection label to existing metrics — avoids breaking existing Grafana dashboards and keeps cardinality manageable for users who don't need per-collection granularity.
- **Collection names excluded from anonymized telemetry** to avoid leaking user data.
- Only REST for now — gRPC would need a different approach since collection info is in the protobuf body rather than the URL. Happy to follow up with that in a separate PR if this direction looks good.

I kept the scope small on purpose per the issue note about splitting into multiple PRs. Point/vector count metrics already have per-collection labels (`collection_points`, `collection_vectors`), so this focuses on the request metrics side.

### DIFF
--- a/src/common/telemetry_ops/requests_telemetry.rs
+++ b/src/common/telemetry_ops/requests_telemetry.rs
@@ -17,6 +17,12 @@ pub type HttpStatusCode = u16;
 /// gRPC status code type alias used for gRPC telemetry tracking.
 pub type GrpcStatusCode = i32;
 
+/// Type alias for per-endpoint response statistics keyed by status code.
+type StatusAggregators = HashMap<HttpStatusCode, Arc<Mutex<OperationDurationsAggregator>>>;
+
+/// Type alias for per-collection, per-endpoint response statistics.
+type CollectionMethodAggregators = HashMap<String, HashMap<String, StatusAggregators>>;
+
 /// Aggregated telemetry data for REST API responses.
 ///
 /// Tracks per-endpoint and per-collection request duration statistics,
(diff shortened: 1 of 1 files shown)


omair445:
hey, retargeted to dev as requested!

omair445:
Hey everyone, I am very very sorry I got stuck in work-related stuff and haven't been able to follow up on this. I am happy if anyone else can take care of this from here or if the maintainers want to push the final bits. Apologies again for the delay!

dancixx:
Closing this in favor of this PR: https://github.com/qdrant/qdrant/pull/8214

PR #8142, opened 2026-02-16
Enable lint detecting ignored unit patterns

agourlay:
> Matching with () explicitly instead of _ outlines the fact that the pattern contains no data. Also it would detect a type change that _ would ignore.

https://rust-lang.github.io/rust-clippy/master/index.html?search=ignored_unit_patterns#ignored_unit_patterns

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -192,6 +192,7 @@ used_underscore_binding = "warn"
 match_wildcard_for_single_variants = "warn"
 needless_pass_by_ref_mut = "warn"
 unused_async = "warn"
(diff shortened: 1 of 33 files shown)


timvisee:
I don't think this will fix any bugs for us. But I have no objection to change this :+1:

agourlay:
> I don't think this will fix any bugs for us.

I agree, it is not a game changer.
Just got annoyed at checking what's inside those `Ok(_)` :)

agourlay:
too much churn to get this green on CI

PR #5095, opened 2024-09-17
Properly preserve full error chain, when `handle_committed_entries` returns an error

ffuugoo:
Continuation of #5066.

As it turned out, we made some "lossy" error conversions in `consensus` module, and so at the point where we were logging consensus error, the error did not contain the cause of the error.

This PR fixes error conversions, so that we preserve error cause.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [ ] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [ ] Have you checked your code using `cargo clippy --all --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #8156, opened 2026-02-17
Fix lossy error propagation in consensus

### LINKED ISSUE #7191: Uninformative TLS related error messages
For the past couple of days, I've been trying to deploy a multi-node Qdrant application with TLS-encrypted cluster communications. After a deep investigation, I figured out that the problem was with my certificates; specifically, the "X509v3 Extended Key Usage" field was misconfigured. However, when I tried to run the Qdrant application on my replicas, I got the following logs:

```
2025-09-01T08:26:04.034340Z  INFO storage::content_manager::consensus::persistent: Initializing new raft state at ./storage/raft_state.json    
2025-09-01T08:26:04.057002Z ERROR qdrant::startup: Panic backtrace: 
   0: qdrant::startup::setup_panic_hook::{{closure}}
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys::backtrace::__rust_end_short_backtrace
   4: __rustc::rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::result::unwrap_failed
   7: core::result::Result<T,E>::expect
   8: qdrant::main
   9: std::sys::backtrace::__rust_begin_short_backtrace
  10: main
  11: <unknown>
  12: __libc_start_main
  13: _start
    
2025-09-01T08:26:04.057022Z ERROR qdrant::startup: Panic occurred in file src/main.rs at line 402: Can't initialize consensus: Failed to initialize Consensus for new Raft state: Failed to create timeout channel: transport error  
```

This is completely uninformative, and it's very difficult to identify the source of the problem. It would be much more helpful if the error included more information about what is going wrong.

leohenon:
Switched error wrapping in consensus.rs from `map_err + anyhow!` to `.context()` to keep the full error chain instead of just transport error.

Picks up from #5095

Fixes #7191

### All Submissions:

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -226,7 +226,7 @@ impl Consensus {
                 &runtime,
                 leader_established_in_ms,
             )
-            .map_err(|err| anyhow!("Failed to initialize Consensus for new Raft state: {err}"))?;
+            .context("Failed to initialize Consensus for new Raft state")?;
         } else {
             runtime
                 .block_on(Self::recover(
@@ -236,9 +236,7 @@ impl Consensus {
                     &config,
                     tls_config.clone(),
                 ))
-                .map_err(|err| {
-                    anyhow!("Failed to recover Consensus from existing Raft state: {err}")
-                })?;
+                .context("Failed to recover Consensus from existing Raft state")?;
 
             if bootstrap_peer.is_some() || uri.is_some() {
                 log::debug!("Local raft state found - bootstrap and uri cli arguments were ignored")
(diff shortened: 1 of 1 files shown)


leohenon:
> Thanks!
> 
> Could you share what the end result looks like? How does using `.context()` here change how we render the error in practice?

Sure!

`map_err(|err| anyhow!("...{err}"))` calls `Display` which just gives you `transport error`. The actual tls details are lost.

`.context()` preserves the original error. `.expect()` uses `Debug`, so you get the full chain:

Before:
```
Failed to create timeout channel: transport error
```

After:
```
Failed to create timeout channel
Caused by:
    0: transport error
    1: connection error
    2: invalid peer certificate
```

Same fix as #5095, that covered only two calls.

PR #8158, opened 2026-02-17
Optimize MmapPointToValues from_iter

agourlay:
Save one points iterator traversal.


JojiiOfficial (approved):

timvisee:
Other than saving a traversal, I see it also removes an iterator clone

PR #8159, opened 2026-02-17
Fix flaky test_cancel_optimization

### LINKED ISSUE #7794: Flaky test `tests::test_cancel_optimization`
### System error

```text
actual_optimization_duration = 183 ms

thread 'tests::test_cancel_optimization' (1772) panicked at lib\collection\src\tests\mod.rs:214:9:
  left: 0
 right: 1
```

### Context

Date: 17.02.2026 12:39

[Flaky failure run](https://github.com/qdrant/qdrant/actions/runs/22097813954/job/process-results)

[Commit](https://github.com/qdrant/qdrant/tree/067ac709a914c6dd3b78e918a50a06b099b930cc)

OS: windows-latest

Branch: refs/pull/8163/merge

#8163

leohenon:
Relax log count assertion from `assert_eq!` to `<=`. 
Early cancellations are intentionally excluded from log (`optimization_worker.rs:376`).

Fixes #7794

### All Submissions:

### DIFF
--- a/lib/collection/src/tests/mod.rs
+++ b/lib/collection/src/tests/mod.rs
@@ -211,7 +211,7 @@ async fn test_cancel_optimization() {
         };
 
         let log = optimizers_log.lock().to_telemetry();
-        assert_eq!(log.len(), expected_optimization_count);
+        assert!(log.len() <= expected_optimization_count);
         for status in log {
             assert_eq!(status.name, "indexing");
             assert!(matches!(status.status, TrackerStatus::Cancelled(_)));


timvisee:
I'm not confident yet this change is correct.

Are cancellations even relevant here? Don't they follow up with a new optimization to still satisfy that count assertion?

leohenon:
> I'm not confident yet this change is correct.
> 
> Are cancellations even relevant here? Don't they follow up with a new optimization to still satisfy that count assertion?

No retry here. The test calls `launch_optimization` directly. Then `stop()` is called on all of them. If a task gets cancelled before reaching `on_successful_start`, it never gets added to the log. 

I considered polling for the log to be populated, but went with the smaller fix since the name + status assertions still cover every entry that makes it into the log. The race only drops entries, never produces false ones.

PR #8163, opened 2026-02-17
Optimize process search results

agourlay:
The `retrieve` function at the segment level is using internally a HashMap which is turned into a Vector as return type.

However the most important caller rebuilds a HashMap from it right away.

This PR exposes the HashMap from the segment level and removes a large amount of unnecessary memory allocation.

Those back and forth conversions accounts for 3% of CPU time on the `search-*` threads.


timvisee:
Great find. 3% of CPU time is insane.

PR #8166, opened 2026-02-17
Timeout for snapshot stream

generall:
[Ai]

```
In the `download_and_unpack_tar` function we download stream of bytes and unpack a tar archive, but it might happen that stream is stuck and downloading does not progress. Is it possible to              
  implement timeout for stream in a way, that if we don't read any bytes from stream in 60 seconds, we timeout the download function
```

### DIFF
--- a/lib/storage/src/content_manager/snapshots/download_tar.rs
+++ b/lib/storage/src/content_manager/snapshots/download_tar.rs
@@ -90,10 +90,7 @@ impl<R> CancellableReader<R> {
 impl<R: Read> Read for CancellableReader<R> {
     fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
         if self.cancel.is_cancelled() {
-            return Err(std::io::Error::new(
-                std::io::ErrorKind::Interrupted,
-                "download cancelled",
-            ));
+            return Err(std::io::Error::other("download cancelled"));
         }
         self.inner.read(buf)
     }
@@ -224,6 +221,11 @@ pub async fn download_and_unpack_tar(
 
 #[cfg(test)]
 mod tests {
+    use std::sync::Arc;
+    use std::sync::atomic::AtomicBool;
+
+    use futures::StreamExt;
+
     use super::*;
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


lib/storage/src/content_manager/snapshots/download_tar.rs
```diff
@@ -1,15 +1,81 @@
...
+    ) -> Poll<std::io::Result<()>> {
+        // Safe to get &mut self because TimeoutReader is Unpin
+        // (all fields are Unpin: Pin<Box<_>> is Unpin, Duration is Unpin)
+        let this = &mut *self;
+        let filled_before = buf.filled().len();
+
+        // First, try to read from the inner reader
+        match this.inner.as_mut().poll_read(cx, buf) {
+            Poll::Ready(Ok(())) => {
+                eprintln!("buf.filled().len() = {:#?}", buf.filled().len());
```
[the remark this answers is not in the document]
generall: damn, right

lib/storage/src/content_manager/snapshots/download_tar.rs
```diff
@@ -25,10 +90,7 @@ impl<R> CancellableReader<R> {
 impl<R: Read> Read for CancellableReader<R> {
     fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
         if self.cancel.is_cancelled() {
-            return Err(std::io::Error::new(
-                std::io::ErrorKind::Interrupted,
-                "download cancelled",
-            ));
+            return Err(std::io::Error::other("download cancelled"));
```
generall, line 93: This change is the most important in the PR.

Turns out, the `Interrupted` is a very poor choice, as it is interpreted as transient error and encouraged to be retried. From docs:

>    This operation was interrupted.
>     Interrupted operations can typically be retried.


Which is the opposite of cancelled. This caused Tar Unpack process to stuck in infinite loop, even after cancellation token was properly triggered.

timvisee: > as it is interpreted as transient error and encouraged to be retried.

😆

PR #8167, opened 2026-02-18
Gridstore deferred updates

IvanPleshkov:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/gridstore/src/gridstore.rs
+++ b/lib/gridstore/src/gridstore.rs
@@ -1,8 +1,10 @@
 use std::cmp::min;
 use std::io::BufReader;
-use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


timvisee:
I believe <https://github.com/qdrant/qdrant/pull/8169> is a much simpler change solving the same problem - holding a lock for a long time on the bitmask.

Gridstore already has deferred updates implemented. You're right that we might write new values into pages immediately, but we still only flush mapping changes in a deferred manner. Values can only be read through such mappings. It means that values that might be written before the flush, are only used after their mappings have been persisted as well as part of the flusher.

Based on the above I therefore don't think this adds much value, and I'm afraid the extra logic makes Gridstore over complicated.

Issue #8094, opened 2026-02-10
[Performance/Liveness] Critical write path serialization: write_ordering_lock held across network I/O


YZL0v3ZZ:
## Summary

During a concurrency audit of the `dev` branch, I identified a significant performance bottleneck and liveness hazard in the `update_with_consistency` method within the `ShardReplicaSet`. The `write_ordering_lock` (a Tokio `Mutex`) is held across network I/O operations when updating remote shards, causing severe **head-of-line blocking** and throughput collapse under high concurrency.

When `WriteOrdering::Strong` or `WriteOrdering::Medium` is used, the lock remains held during the entire `self.update()` call, which includes network communication with remote replicas. This serializes all concurrent write requests on the leader node, reducing throughput to `1 / network_latency`.

## Location

* **File:** `lib/collection/src/shards/replica_set/update.rs`
* **Branch:** `dev` 
* **Method:** `update_with_consistency`
* **Lines:** 134-197 (lock acquisition at lines 163-166, held until line 176)

## Code Analysis

The problematic code flow:

```rust
// lib/collection/src/shards/replica_set/update.rs:160-176

// 1. Lock is acquired here based on write ordering
let _write_ordering_lock = match ordering {
    WriteOrdering::Strong | WriteOrdering::Medium => {
        Some(self.write_ordering_lock.lock().await)
    }
    WriteOrdering::Weak => None,
};

// 2. self.update() is awaited while the lock is still held
self.update(
    operation,
    wait,
    timeout,
    update_only_existing,
    hw_measurement_acc,
)
.await
```

The call chain that performs network I/O while holding the lock:

1. `update_with_consistency()` (line 160-176) - **holds lock**
2. → `update()` (line 168-176) - **lock still held**
3. → `update_impl()` (line 247) - **lock still held**
4. → Network I/O via `buffer_unordered()` or `FuturesUnordered` (lines 374-383) - **lock still held**

```rust
// lib/collection/src/shards/replica_set/update.rs:374-383
let all_res: Vec<Result<_, _>> = match self.shared_storage_config.update_concurrency {
    Some(concurrency) => {
        futures::stream::iter(update_futures)
            .buffer_unordered(concurrency.get())  // Network I/O happens here
            .collect()
            .await
    }
    None => FuturesUnordered::from_iter(update_futures).collect().await,
};
```

### 1. **Throughput Collapse**
- If a remote shard has 500ms latency, the lock is held for 500ms
[...]

PR #8174, opened 2026-02-19
[replica_set] Reduce medium write ordering lock contention

krapcys1-maker:
## Summary

This draft narrows write_ordering_lock usage in ShardReplicaSet::update_with_consistency:

- keep lock serialization for WriteOrdering::Strong
- stop locking WriteOrdering::Medium with the same global lock

The goal is to reduce head-of-line blocking for medium-ordered writes while preserving existing strong-order behavior.

## Why

Issue #8094 reports that write_ordering_lock is held across self.update(...).await, which includes remote I/O. That can serialize writes too aggressively.

This patch takes a conservative first step by removing lock contention only from Medium, while keeping Strong semantics unchanged.

## Changes

- lib/collection/src/shards/replica_set/update.rs
  - WriteOrdering::Strong => Some(self.write_ordering_lock.lock().await)
  - WriteOrdering::Weak | WriteOrdering::Medium => None
- lib/collection/src/shards/replica_set/mod.rs
  - updated write_ordering_lock field comment to reflect strong-only serialization

## Validation

- cargo test -p collection replica_set::update::tests --lib
- cargo test -p collection point_ops --lib
- cargo test -p collection

All passed locally.

## Scope and follow-up

- This is intentionally minimal and focused.
- If maintainers want stronger ordering guarantees for Medium, we can follow up with dedicated ordering tests and/or a separate lane/queue design.

Related: #8094

### DIFF
--- a/lib/collection/src/shards/replica_set/update.rs
+++ b/lib/collection/src/shards/replica_set/update.rs
@@ -163,7 +163,7 @@ impl ShardReplicaSet {
             // Lock updates only for strong ordering.
             let _write_ordering_lock = match ordering {
                 WriteOrdering::Strong => Some(self.write_ordering_lock.lock().await),
-                // Medium writes already tolerate short-term inconsistencies after leader changes.
+                // Medium may tolerate temporary ordering divergence on non-leader replicas.
                 // Avoid serializing all medium writes behind remote I/O.
                 WriteOrdering::Weak | WriteOrdering::Medium => None,
             };


krapcys1-maker:
Addressed the CodeRabbit clarity nit in commit 88cd8c1cf: updated the inline comment to explicitly mention temporary medium-order divergence on non-leader replicas and kept the direct motivation about avoiding remote-I/O serialization.

PR #8175, opened 2026-02-19
Optimize HNSW plain-search filtered allocation

agourlay:
This PR propagates the filtered point iterator properly avoiding unnecessary allocations.

Issue found through profiling where gigantic `collect()` calls shows up before scoring.

The HNSW plain search uses `query_points` to fetch the matching points from the payload index.
This allocates the full filtered result set in memory which is then used for scoring.

However the underlying `search_plain_batched` requires just an Iterator.

I was not able to refactor `query_points` to return an Iterator due to liftetime issues.
Therefore so I simply inlined this tiny body for this call site.


lib/segment/src/index/struct_payload_index.rs
agourlay, line 967: the bad `collect()`

PR #8177, opened 2026-02-19
[gpu] Use c_char pointers for Vulkan extension/layer names (ARM64)

krapcys1-maker:
## Summary

Fix ARM64 (arch64) GPU build type mismatch by using c_char pointers for Vulkan extension/layer names instead of hardcoded i8 pointers.

Related issue: #8096

## Problem

In lib/gpu, raw Vulkan name pointer vectors were typed as Vec<*const i8>. On ARM64, c_char differs, which causes pointer type mismatch during GPU build.

## Changes

- lib/gpu/src/instance.rs
  - extension_names_raw: Vec<*const c_char>
  - layers_raw: Vec<*const c_char>
- lib/gpu/src/device.rs
  - extension_names_raw: Vec<*const c_char>
- Added imports:
  - use std::ffi::{CString, c_char};

No behavior changes, only platform-correct pointer typing.

## Validation

- cargo check -p gpu
- cargo test -p gpu --lib
- cargo fmt --all -- --check

All pass locally on current environment.

## Notes

A similar community PR existed earlier and was closed by the author. This PR reintroduces the same focused fix for dev branch in a merge-ready form.

### DIFF
--- a/lib/gpu/src/device.rs
+++ b/lib/gpu/src/device.rs
@@ -1,4 +1,4 @@
-use std::ffi::CString;
+use std::ffi::{CString, c_char};
 use std::sync::Arc;
 
 use ash::vk;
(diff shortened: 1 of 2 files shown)


krapcys1-maker:
Additional validation: ran cargo check -p gpu --target aarch64-unknown-linux-gnu locally after this patch; it compiles successfully. (No ARM runtime smoke test was performed in this environment.)

PR #8179, opened 2026-02-19
[snapshot] Avoid leaving replicas in recovery state when shard data is absent

### LINKED ISSUE #7851: Restore of cluster snapshots creates unnecessary replicas in Partial state
When trying to restore snapshots taken from a three node cluster with three shards and two replicas, it appears that the restored shards can have three replicas instead of two. It doesn't appear to be consistent with every collection on every run and on every shard, sometimes some shards are restored as expected, but others aren't.
This leaves a dirty state with Partial replicas.

## Steps to Reproduce
1. Take a snapshot of a cluster (in my case version 1.13.4)
2. Try to restore the snapshots on a separate qdrant cluster (tried on versions 1.13.4 up to 1.16.3)

## Expected Behavior
I expect a consistent shard state

krapcys1-maker:
## Summary
- preserve pre-recovery local replica states before snapshot restore
- when a shard has no local data in the uploaded snapshot, revert that shard from Partial/ManualRecovery back to its previous state
- keep existing behavior for shards that are actually recovered from snapshot data

## Why
During collection snapshot recovery in distributed mode, all local shards are moved to recovery state first. If a given shard has no local data in this particular node snapshot, recovery skips it, and the replica could remain stuck in Partial (or ManualRecovery). This can leave unnecessary non-active replicas after sequential cluster restore.

## Validation
- cargo check -p storage --locked
- cargo test -p storage --lib --locked

Fixes #7851

### DIFF
--- a/lib/storage/src/content_manager/snapshots/recover.rs
+++ b/lib/storage/src/content_manager/snapshots/recover.rs
@@ -301,34 +301,26 @@ async fn _do_recover_from_snapshot(
             // This peer may have been switched into recovery state before restore. If the snapshot
             // does not contain local data for this shard, revert to previous state so the replica
             // is not left stuck in `Partial`/`ManualRecovery`.
-            if let Some(previous_state) = local_states_before_recovery.get(shard_id).copied() {
-                let current_state = collection
-                    .state()
-                    .await
-                    .shards
-                    .get(shard_id)
-                    .and_then(|info| info.replicas.get(&this_peer_id))
-                    .copied();
-
-                if current_state == Some(recovery_state) {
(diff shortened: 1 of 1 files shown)


krapcys1-maker:
Addressed the CodeRabbit TOCTOU finding in 
ecover.rs.

Change in follow-up commit 8c4c4d4e1:
- removed the runtime collection.state().await read in the !recovered branch,
- switched guard to previous_state != recovery_state,
- kept rom_state = Some(recovery_state) in state transitions so conditional apply remains enforced by the state machine.

This avoids suppressing the revert when deactivation proposals are still in-flight.

PR #8188, opened 2026-02-20
Gridstore non-blocking flush

IvanPleshkov:
This PR fixes search spikes while intensive upsert-search on low-IO configuration.
Spike appears when gridstore flush blocks writing by `RwLock::read`. Writing blocks searches - spike appears because of waiting of flush.

The solution is non-blocking flush, where flush does not block writes.
To achieve it, this PR changes `MmapMut` into `MmapSlice<u8>` which can be created from `MmapMut` and has `Flusher` support.

### DIFF
--- a/lib/gridstore/src/gridstore.rs
+++ b/lib/gridstore/src/gridstore.rs
@@ -3,6 +3,7 @@ use std::io::BufReader;
 use std::path::PathBuf;
 use std::sync::Arc;
 
+use ahash::AHashMap;
 use common::counter::hardware_counter::HardwareCounterCell;
 use common::counter::referenced_counter::HwMetricRefCounter;
 use common::fs::atomic_save_json;
@@ -19,7 +20,7 @@ use crate::blob::Blob;
 use crate::config::{Compression, StorageConfig, StorageOptions};
 use crate::error::GridstoreError;
 use crate::page::Page;
-use crate::tracker::{BlockOffset, PageId, PointOffset, Tracker, ValuePointer};
+use crate::tracker::{BlockOffset, PageId, PointOffset, PointerUpdates, Tracker, ValuePointer};
 
 const CONFIG_FILENAME: &str = "config.json";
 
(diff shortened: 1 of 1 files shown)


lib/gridstore/src/page.rs
```diff
@@ -18,7 +18,7 @@ pub(crate) struct Page {
     /// Main data mmap for read/write
     ///
     /// Best suited for random reads.
-    mmap: MmapMut,
+    mmap: MmapSlice<u8>,
```
timvisee, line 21: Why swap the type here? Is it just so that we have a premade `.flusher()`?

lib/gridstore/src/gridstore.rs
```diff
@@ -617,36 +617,44 @@ impl<V> Gridstore<V> {
...
 
-            let old_pointers = tracker.write().write_pending(pending_updates);
-            tracker.read().flush()?;
+            let (old_pointers, tracker_flusher) = {
+                let mut guard = tracker.write();
+                let old_pointers = guard.write_pending(pending_updates);
+                let flusher = guard.flusher();
+                (old_pointers, flusher)
+            };
+            tracker_flusher()?;
```
timvisee: In my opinion it'd be nicer to abstract all of this away into `write_pending` and maybe rename it to `write_pending_flusher(pending)` which then returns the old pointers and the flush closure to us.

No strong opinion though. Feel free to keep it as is.

lib/gridstore/src/page.rs
generall: I think this is fine, we are going to change this to universal file interface anyway

timvisee:
Thanks! I like this a lot better

https://github.com/qdrant/qdrant/pull/8188/commits/f5d46789a6add6e3d48386c17708557299ca89b8

lib/gridstore/src/page.rs
IvanPleshkov: Because `MmapSlice` has a ready-to-use `flusher()`. `MmapMut` is an external crate type without flusher.

lib/gridstore/src/gridstore.rs
IvanPleshkov: I keep this func as is but I did a refactor as we discussed before. Flusher is a list of subparts flush now

PR #8210, opened 2026-02-23
use `UniversalRead` in `ImmutableDenseVectors`

coszio:
Builds on top of #8196 

Refactors `MmapDenseVectors` to use a generic `UniversalRead` storage. By default still using mmap, but through `MmapUniversal` type. 

By doing this, we are forced to rename it to something more appropriate, I propose `ImmutableDenseVectors`.

## Future work
The current storage interface is made to work with slices everywhere, but `UniversalRead` works with `Cow`. What we probably want is a cow-like enum, which can handle either a slice (`&[T]`) or a boxed slice for the owned variant (`Box<[T]>`), which is a bit more lightweight than a `Vec<T>`.

### DIFF
--- a/lib/segment/src/vector_storage/async_raw_scorer.rs
+++ b/lib/segment/src/vector_storage/async_raw_scorer.rs
@@ -11,8 +11,8 @@ use crate::data_types::vectors::{DenseVector, QueryVector, VectorElementType, Ve
 use crate::spaces::metric::Metric;
 use crate::spaces::simple::{CosineMetric, DotProductMetric, EuclidMetric, ManhattanMetric};
 use crate::types::Distance;
+use crate::vector_storage::dense::immutable_dense_vectors::ImmutableDenseVectors;
 use crate::vector_storage::dense::memmap_dense_vector_storage::MemmapDenseVectorStorage;
-use crate::vector_storage::dense::mmap_dense_vectors::MmapDenseVectors;
 use crate::vector_storage::query::NaiveFeedbackQuery;
 use crate::vector_storage::query_scorer::QueryScorer;
 use crate::vector_storage::query_scorer::metric_query_scorer::MetricQueryScorer;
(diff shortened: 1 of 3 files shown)


lib/segment/src/vector_storage/dense/immutable_dense_vectors.rs
```diff
@@ -54,35 +48,32 @@ pub struct MmapDenseVectors<T: PrimitiveVectorElement> {
...
 }
 
-impl<T: PrimitiveVectorElement> MmapDenseVectors<T> {
+impl<T: PrimitiveVectorElement, S: UniversalRead<u8>> ImmutableDenseVectors<T, S> {
     pub fn open(
         vectors_path: &Path,
         deleted_path: &Path,
         dim: usize,
         with_async_io: bool,
-        madvise: AdviceSetting,
```
coszio, line 63: It also removes this propagated setting, which was actually set to `AdviseSetting::Global` everywhere

lib/segment/src/vector_storage/dense/memmap_dense_vector_storage.rs
```diff
@@ -36,15 +36,15 @@ const DELETED_PATH: &str = "deleted.dat";
 pub struct MemmapDenseVectorStorage<T: PrimitiveVectorElement> {
     vectors_path: PathBuf,
     deleted_path: PathBuf,
-    mmap_store: Option<MmapDenseVectors<T>>,
+    vectors: Option<ImmutableDenseVectors<T, MmapUniversal<u8>>>,
     distance: Distance,
 }
```
ffuugoo, line 41: Should we rename this to something like `ImmutableDenseVectorStorage` as well?

coszio: For now, the name still makes sense to keep, since `vectors` field is specifically mmap-backed

ffuugoo: Ah, ok, I see now

PR #8211, opened 2026-02-24
Remove rocksdb from gpu tests

IvanPleshkov:
This PR removes RocksDB dependency from GPU tests (it was required for an obsolete vector storage creation).
Also, this test increases the acc error for the final HNSW test to make it more stable on llvmpipe

### DIFF
--- a/lib/segment/tests/integration/gpu_hnsw_test.rs
+++ b/lib/segment/tests/integration/gpu_hnsw_test.rs
@@ -59,7 +59,7 @@ fn test_gpu_filterable_hnsw() {
         .try_init();
 
     let stopped = AtomicBool::new(false);
-    let max_failures = 5;
+    let max_failures = 7;
     let dim = 8;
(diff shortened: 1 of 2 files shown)


lib/segment/tests/integration/gpu_hnsw_test.rs
```diff
@@ -59,7 +59,7 @@ fn test_gpu_filterable_hnsw() {
         .try_init();
 
     let stopped = AtomicBool::new(false);
-    let max_failures = 5;
+    let max_failures = 7;
```
timvisee, line 62: Why do we increase this one?

IvanPleshkov: I mentioned it in the PR description. I got flakiness while testing on llvmpipe (emulator). We know that the emulator is not stable, and I find OK such accuracy decrease fine

PR #8212, opened 2026-02-24
universal ui rollback to bytes

generall:


lib/common/common/src/universal_io/elements_io.rs
```diff
@@ -0,0 +1,72 @@
...
+
+pub type ElementsOffset = u64;
+
+#[derive(Copy, Clone, Debug)]
+pub struct ElementsRange {
+    pub start: ElementsOffset,
+    pub length: u64,
+}
+
+pub trait UniversalElementRead<T: Copy + 'static>: crate::universal_io::UniversalRead {
```
coszio, line 14: We should add `zerocopy::FromBytes` requirement, so that we ensure that the byte pattern constitutes a valid `T`

lib/common/common/src/mmap/ops.rs
```diff
@@ -173,6 +173,34 @@ pub unsafe fn transmute_from_u8_to_slice<T>(data: &[u8]) -> &[T] {
...
+    );
+
+    let len = data.len() / size_of::<T>();
+    let ptr = data.as_ptr().cast::<T>();
+    let capacity = data.capacity() / size_of::<T>();
+
+    // We need to forget the original vector to avoid double free
+    std::mem::forget(data);
+
+    unsafe { Vec::from_raw_parts(ptr as *mut T, len, capacity) }
```
coszio, line 201: We also need to make sure `data.capacity()` is an exact multiple of `size_of::<T>()`, otherwise we might not deallocate the extra capacity properly.

```suggestion
    // Make sure capacity is an exact multiple of size of T
    data.shrink_to_fit();
    
    // Avoid dropping the original vector
    let data = ManuallyDrop::new(data);
    
    let len = data.len() / size_of::<T>();
    let ptr = data.as_ptr().cast::<T>();
    let capacity = data.capacity() / size_of::<T>();
    
    debug_assert_eq!(capacity, len);
    
    unsafe { Vec::from_raw_parts(ptr as *mut T, len, capacity) }
```

lib/common/common/src/universal_io/elements_io.rs
```diff
@@ -0,0 +1,72 @@
...
+            start: byte_offset,
+            length: byte_length,
+        };
+
+        let byte_data = self.read::<SEQUENTIAL>(byte_range)?;
+
+        #[expect(deprecated, reason = "fuck off")]
+        let res = match byte_data {
+            Cow::Borrowed(bytes) => {
+                Cow::Borrowed(unsafe { mmap::transmute_from_u8_to_slice(bytes) })
```
coszio, line 33: If we add `FromBytes` requirement, we can use `ref_from_bytes`

```suggestion
                let slice = <[T]>::ref_from_bytes(bytes)
                    .expect("byte slice is not a valid &[T] (alignment or size mismatch)");
                Cow::Borrowed(slice)
```

lib/common/common/src/universal_io/elements_io.rs
generall: why?

lib/common/common/src/mmap/ops.rs
```diff
@@ -173,6 +173,34 @@ pub unsafe fn transmute_from_u8_to_slice<T>(data: &[u8]) -> &[T] {
...
+        align_of::<T>(),
+        data.as_ptr().align_offset(align_of::<T>()),
+    );
+
+    let len = data.len() / size_of::<T>();
+    let ptr = data.as_ptr().cast::<T>();
+    let capacity = data.capacity() / size_of::<T>();
+
+    // We need to forget the original vector to avoid double free
+    std::mem::forget(data);
```
ffuugoo, line 199: [`Vec::into_raw_parts`] instead of `forget`?

```suggestion
    let (ptr, len, capacity) = data.into_raw_parts();
    
    let ptr = ptr.cast::<T>();
    let len = len / size_of::<T>();
    let capacity = capacity / size_of::<T>();
```

[`Vec::into_raw_parts`]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_raw_parts

lib/common/common/src/universal_io/elements_io.rs
coszio: It ensures we don't use any type here, only those which can be safely transmuted

PR #8215, opened 2026-02-24
Fix unused `self` lint in macos

coszio:
After porting WAL repo to here, I am seeing a new lint in macos. Let's just `expect` it.

### DIFF
--- a/lib/wal/src/mmap_view_sync.rs
+++ b/lib/wal/src/mmap_view_sync.rs
@@ -19,7 +19,7 @@ pub struct MmapViewSync {
 }
 
 impl MmapViewSync {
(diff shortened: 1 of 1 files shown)


lib/wal/src/mmap_view_sync.rs
```diff
@@ -19,6 +19,7 @@ pub struct MmapViewSync {
 }
 
 impl MmapViewSync {
+    #[cfg_attr(target_os = "macos", expect(clippy::unused_self))]
```
timvisee: Let's use `not(target_os = "linux")` to also include the other variants.

```suggestion
    #[cfg_attr(not(target_os = "linux"), expect(clippy::unused_self))]
```

PR #8219, opened 2026-02-24
fix(collection): stabilize snapshot checksum race

### LINKED ISSUE #5133: Flaky test `tests::snapshot_test::test_snapshot_collection_normal`
### System error

```text
thread 'tests::snapshot_test::test_snapshot_collection_normal' panicked at lib/collection/src/tests/snapshot_test.rs:100:5:
  left: 0
 right: 64
```

### Context

Date: 23.09.2024 20:12

[Flaky failure run](https://github.com/qdrant/qdrant/actions/runs/11001275574/job/process-results)

[Commit](https://github.com/qdrant/qdrant/tree/f9d72a0922483d528595545ce6f0e7805a3783ed)

OS: ubuntu-latest

Branch: refs/heads/dev

#

krapcys1-maker:
## Summary
Fix flaky `tests::snapshot_test::test_snapshot_collection_normal` by removing a race in snapshot checksum handling.

## Root cause
`SnapshotStorageLocalFS::store_file` wrote checksum bytes and immediately called `get_snapshot_description`, which reads the checksum file. The writer handle was still open, so on CI this occasionally read an empty checksum (`len = 0`) and triggered the flaky assertion expecting `64`.

## Fix
- call `file.sync_all().await?` after writing checksum
- explicitly `drop(file)` before reading snapshot description

This guarantees checksum contents are persisted and visible before `read_to_string`.

## Validation
- `cargo +nightly fmt --all`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test -p collection`
- `cargo test -p collection tests::snapshot_test::test_snapshot_collection_normal` (40 repetitions)

Fixes #5133

### DIFF
--- a/lib/collection/src/common/snapshots_manager.rs
+++ b/lib/collection/src/common/snapshots_manager.rs
@@ -258,6 +258,11 @@ impl SnapshotStorageLocalFS {
         Ok(snapshots)
     }
 
+    /// Stores a snapshot in local storage and writes its checksum first.
+    ///
+    /// The checksum file is explicitly flushed and closed before moving the
+    /// snapshot into place, so checksum readers do not observe a partially
+    /// written checksum file.
     async fn store_file(
         &self,
         source_path: &Path,
@@ -288,6 +293,8 @@ impl SnapshotStorageLocalFS {
         let checksum_file = TempPath::from_path(&checksum_path);
         let mut file = tokio_fs::File::create(checksum_path.as_path()).await?;
         file.write_all(checksum.as_bytes()).await?;
+        file.flush().await?;
+        drop(file);
 
         move_file(&source_path, &target_path_tmp).await?;
         target_path_tmp.persist(target_path).map_err(|e| e.error)?;


lib/collection/src/common/snapshots_manager.rs
```diff
@@ -288,6 +293,8 @@ impl SnapshotStorageLocalFS {
         let checksum_file = TempPath::from_path(&checksum_path);
         let mut file = tokio_fs::File::create(checksum_path.as_path()).await?;
         file.write_all(checksum.as_bytes()).await?;
+        file.sync_all().await?;
```
timvisee: Shouldn't we use `flush` here instead?


```suggestion
        file.flush().await?;
```

I believe we care that any pending bytes reach their destination. We don't necessarily care they are durably stored. I believe other readers still see our changes even if it's still in cache and not durably persisted yet.

Please correct me if I'm wrong - we had flakiness after all.

krapcys1-maker: Great point, thanks.

I’ve switched this to `file.flush().await?` and kept explicit `drop(file)` before proceeding, so pending bytes are flushed and the handle is closed before readers can observe the checksum file.
We don’t require durability guarantees here — only visibility and correct read ordering.

I also reran the snapshot regression path:
- `cargo test -p collection tests::snapshot_test::test_snapshot_collection_normal -- --nocapture`
- 40 repeated runs of the same test

All passed.

PR #8221, opened 2026-02-24
use `Cow<'_, [T]>` as return type in dense vector storages

coszio:
Implements the `Owned` variant left as TODO in #8210, and propagates the usage of `Cow<'_, [T]>` everywhere in the dense vector storages.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_links.rs
+++ b/lib/segment/src/index/hnsw_index/graph_links.rs
@@ -77,7 +77,7 @@ pub trait GraphLinksVectors {
 
     /// Link vectors will be included for each link per point.
     /// The layout of each vector must correspond to [`VectorLayout::link`].
-    fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<Cow<'_, [u8]>>;
+    fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<&[u8]>;
 
     /// Get the layout of base and link vectors.
     fn vectors_layout(&self) -> GraphLinksVectorsLayout;
@@ -128,10 +128,8 @@ impl<'a> GraphLinksVectors for StorageGraphLinksVectors<'a> {
     }
 
     /// Note: unlike base vectors, link vectors are written in a random order.
-    fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<Cow<'_, [u8]>> {
-        Ok(Cow::Borrowed(
-            self.quantized_vectors.get_quantized_vector(point_id),
-        ))
+    fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<&[u8]> {
+        Ok(self.quantized_vectors.get_quantized_vector(point_id))
     }
 
     fn vectors_layout(&self) -> GraphLinksVectorsLayout {
@@ -387,8 +385,8 @@ mod tests {
             Ok(Cow::Borrowed(&self.base_vectors[point_id as usize]))
         }
 
-        fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<Cow<'_, [u8]>> {
-            Ok(Cow::Borrowed(&self.link_vectors[point_id as usize]))
+        fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<&[u8]> {
+            Ok(&self.link_vectors[point_id as usize])
         }
 
         fn vectors_layout(&self) -> GraphLinksVectorsLayout {
@@ -436,7 +434,7 @@ mod tests {
                     assert!(base_vector.is_empty());
                 }
                 iter.map(|(link, bytes)| {
-                    assert_eq!(bytes, vectors.get_link_vector(link).unwrap().as_ref());
+                    assert_eq!(bytes, vectors.get_link_vector(link).unwrap());
                     link
                 })
                 .collect()
--- a/lib/segment/src/index/hnsw_index/graph_links/serializer.rs
+++ b/lib/segment/src/index/hnsw_index/graph_links/serializer.rs
@@ -156,7 +156,7 @@ pub fn serialize_graph_links<W: Write + Seek>(
                         if vector.len() != vectors_layout.link.size() {
                             return Err(OperationError::service_error("Vector size mismatch"));
                         }
-                        writer.write_all(&vector)?;
+                        writer.write_all(vector)?;
                         offset += vector.len();
                     }
 


generall:
What's the reason for this? There is a slight overhead for Cow usage in some hot loop scenarios

ffuugoo:
> What's the reason for this? There is a slight overhead for Cow usage in some hot loop scenarios

Well, `UniversalRead::read` returns `Cow`. So, if you have something like this, you _have to_ propagate `Cow` everywhere.


```rust
fn outer() {
    let data = inner();
}

fn inner() -> &[u8] {
    innermost()
}

fn innermost() -> &[u8] {
    let reader: UniversalMmap<u8> = ...;
    let data: Cow<'_, [u8]> = reader.read(...);

    // can't return slice here, because `data` is local and can't be borrowed beyond `innermost` body,
    // have to return `Cow` everywhere
    &data
}
```

@coszio did this in #8210, but once we implement `IoUringFile`, we can't rely on `Cow::Borrowed` anymore.

```rust
    match reader.read(...) {
        Cow::Borrowed(slice) => slice,
        Cow::Owned(_) => unreachable!(),
    }
```

lib/common/common/src/maybe_uninit.rs
```diff
@@ -31,3 +32,39 @@ pub fn maybe_uninit_fill_from<I: IntoIterator>(
...
+
+/// Wrapper around [`maybe_uninit_fill_from`] that doesn't leak on drop.
+#[inline(always)]
+pub fn maybe_uninit_fill_from_with_drop<I: IntoIterator>(
+    this: &mut [MaybeUninit<I::Item>],
+    it: I,
+) -> InitUninit<'_, I::Item> {
+    let (initted, remainder) = maybe_uninit_fill_from(this, it);
+    (RefDropper(initted), remainder)
+}
```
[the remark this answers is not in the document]
coszio: I have removed the const assertion, and backported the `<[MaybeUninit<T>]>::assume_init_mut` too, so that dropping does happen, as it is done in the [original implementation of `fill_from`](https://github.com/rust-lang/rust/pull/121280/changes#diff-5e72af2fe825c2180f437d54adbdbb19bb40a9cc82bc6c53c2d993d72b08ce3fR1291)

lib/common/common/src/maybe_uninit.rs
```diff
@@ -26,8 +21,23 @@ pub fn maybe_uninit_fill_from<I: IntoIterator>(
 
     // SAFETY: Valid elements have just been written into `init`, so that portion
     // of `this` is initialized.
-    (
-        unsafe { transmute::<&mut [MaybeUninit<I::Item>], &mut [I::Item]>(initted) },
-        remainder,
-    )
+    (unsafe { assume_init_mut(initted) }, remainder)
+}
```
ffuugoo, line 25: Not your fault, but this is terribly written. "initted"? 😹

```rust
pub fn maybe_uninit_fill_from<I: IntoIterator>(
    slice: &mut [MaybeUninit<I::Item>],
    iter: I,
) -> (&mut [I::Item], &mut [MaybeUninit<I::Item>]) {
    let mut initialized_len = 0;
    
    for (element, value) in slice.iter_mut().zip(iter) {
        element.write(value);
        initialized_len += 1;
    }

    // SAFETY: initialized_len <= slice.len()
    let (initialized, uninitialized) = unsafe { slice.split_at_mut_unchecked(initialized_len) };

    // SAFETY: Valid elements have just been written into `initialized`, so that portion
    // of `slice` is initialized.
    (unsafe { assume_init_mut(initialized) }, uninitialized)
}
```

lib/common/common/src/maybe_uninit.rs
```diff
@@ -26,8 +21,23 @@ pub fn maybe_uninit_fill_from<I: IntoIterator>(
...
+/// Calling this when the content is not yet fully initialized causes undefined
+/// behavior: it is up to the caller to guarantee that every `MaybeUninit<T>` in the
+/// slice really is in an initialized state. For instance, `.assume_init_mut()` cannot
+/// be used to initialize a `MaybeUninit` slice.
+#[inline(always)]
+pub const unsafe fn assume_init_mut<T>(this: &mut [MaybeUninit<T>]) -> &mut [T] {
+    // SAFETY: similar to safety notes for `slice_get_ref`, but we have a
+    // mutable reference which is also guaranteed to be valid for writes.
+    unsafe { &mut *(this as *mut [MaybeUninit<T>] as *mut [T]) }
 }
```
ffuugoo, line 43: This is not so stellar either. (Sorry...)

```suggestion
#[inline(always)]
pub const unsafe fn assume_init_mut<T>(slice: &mut [MaybeUninit<T>]) -> &mut [T] {
    // SAFETY: similar to safety notes for `slice_get_ref`, but we have a
    // mutable reference which is also guaranteed to be valid for writes.
    unsafe { slice::from_raw_parts_mut(slice.as_mut_ptr().cast(), slice.len()) }
}
```

(And `use std::slice` on top of the file.)

lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -127,8 +128,10 @@ impl<'a> GraphLinksVectors for StorageGraphLinksVectors<'a> {
     }
 
     /// Note: unlike base vectors, link vectors are written in a random order.
-    fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<&[u8]> {
-        Ok(self.quantized_vectors.get_quantized_vector(point_id))
+    fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<Cow<'_, [u8]>> {
+        Ok(Cow::Borrowed(
+            self.quantized_vectors.get_quantized_vector(point_id),
+        ))
```
ffuugoo: I'd leave this `&[u8]` until `get_quantized_vector` converted to `Cow`.

lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -72,11 +73,11 @@ pub enum GraphLinksFormatParam<'a> {
 pub trait GraphLinksVectors {
     /// Base vectors will be included once per point on level 0.
     /// The layout of each vector must correspond to [`VectorLayout::base`].
-    fn get_base_vector(&self, point_id: PointOffsetType) -> OperationResult<&[u8]>;
+    fn get_base_vector(&self, point_id: PointOffsetType) -> OperationResult<Cow<'_, [u8]>>;
 
     /// Link vectors will be included for each link per point.
     /// The layout of each vector must correspond to [`VectorLayout::link`].
-    fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<&[u8]>;
+    fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<Cow<'_, [u8]>>;
```
ffuugoo: Ditto.

lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -380,12 +383,12 @@ mod tests {
...
-            Ok(&self.base_vectors[point_id as usize])
+        fn get_base_vector(&self, point_id: PointOffsetType) -> OperationResult<Cow<'_, [u8]>> {
+            Ok(Cow::Borrowed(&self.base_vectors[point_id as usize]))
         }
 
-        fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<&[u8]> {
-            Ok(&self.link_vectors[point_id as usize])
+        fn get_link_vector(&self, point_id: PointOffsetType) -> OperationResult<Cow<'_, [u8]>> {
+            Ok(Cow::Borrowed(&self.link_vectors[point_id as usize]))
         }
```
ffuugoo: Ditto.

lib/common/common/src/maybe_uninit.rs
coszio: This was 95% copypasted from the unstable version of this function

https://doc.rust-lang.org/src/core/mem/maybe_uninit.rs.html#1513

lib/common/common/src/maybe_uninit.rs
coszio: This is exactly how it is named in `std::mem`, don't quote me 😄 

 https://doc.rust-lang.org/src/core/mem/maybe_uninit.rs.html#1384

ffuugoo: yeah yeah, I figured. still very meh 🤷‍♀️

lib/common/common/src/maybe_uninit.rs
ffuugoo: we are better than ~~rust std lib~~ this

PR #8222, opened 2026-02-24
Deduplicate points after optimization

generall:
This PR is part of deferred updates problem:

Once we optimize segment with deferred updates, we need to make sure that points for which we skipped deletion are properly removed.
This can be achieved by simply checking points versions and removing those points which have lower version.

ToDo:

- [x] actually identify which points should be addressed. This should be done from the removed proxy segments
- [x] Update logic of detecting point needs to marked as deleted (done in https://github.com/qdrant/qdrant/pull/8279)

### DIFF
--- a/lib/collection/src/tests/deferred_points_dedup.rs
+++ b/lib/collection/src/tests/deferred_points_dedup.rs
@@ -233,6 +233,17 @@ async fn test_deferred_points_dedup_after_optimization() {
         .await
         .unwrap();
 
+    // Check that there are deferred points in the shard, otherwise the test scenario is not valid.
+    let has_deferred = shard
+        .segments()
+        .read()
+        .iter()
+        .any(|segment| !segment.1.get().read().deferred_point_ids().is_empty());
+    assert!(
+        has_deferred,
+        "Expected to have deferred points after updates, but found none"
+    );
+
     // Wait for optimization triggered by the overwrites
     wait_optimization(&shard, timeout).await;
 
@@ -242,6 +253,17 @@ async fn test_deferred_points_dedup_after_optimization() {
         .update(upsert_op(random_points()), true, None, hw_acc.clone())
         .await
         .unwrap();
+
+    // Check that there are deferred points in the shard, otherwise the test scenario is not valid.
+    let has_deferred = shard
+        .segments()
+        .read()
+        .iter()
+        .any(|segment| !segment.1.get().read().deferred_point_ids().is_empty());
+    assert!(
+        has_deferred,
+        "Expected to have deferred points after updates, but found none"
+    );
 
     // Wait for final optimization
     wait_optimization(&shard, timeout).await;
--- a/lib/segment/src/segment/entry.rs
+++ b/lib/segment/src/segment/entry.rs
@@ -885,7 +885,7 @@ impl NonAppendableSegmentEntry for Segment {
         let id_tracker = self.id_tracker.borrow();
         id_tracker
             .iter_internal()
-            .filter(move |&internal_id| internal_id >= deferred_from)
+            .skip_while(|&internal_id| internal_id < deferred_from)
             .filter_map(|internal_id| id_tracker.external_id(internal_id))
             .collect()
     }
--- a/lib/shard/src/optimize.rs
+++ b/lib/shard/src/optimize.rs
@@ -493,6 +493,12 @@ fn finish_optimization(
     let deferred_points: Vec<PointIdType> = deferred_points_set.into_iter().collect();
 
     if !deferred_points.is_empty() {
+        // Deferred points in proxy segment may become visible for optimized segment (in most cases).
+        // It's time to deduplicate them and remove older versions from optimized segment,
+        // where they were visible while deferred status.
+        // There are a situations, when deferred point is still deferred after optimization,
+        // so `deduplicate_points` also cover this case and delete only older versions of the point,
+        // which are still visible for optimized segment.
         read_segment_holder.deduplicate_points(&deferred_points, hw_counter)?;
     }
 


lib/shard/src/optimize.rs
```diff
@@ -475,7 +477,15 @@ fn finish_optimization(
         writable_segment_holder.remove_segment_if_not_needed(cow_segment_id)?;
     }
 
-    drop(writable_segment_holder);
+    let read_segment_holder = RwLockWriteGuard::downgrade(writable_segment_holder);
+    // Can read, but can't yet write updates.
+
+    // ToDo: collect deferred points from proxies.
+    let deferred_points: Vec<PointIdType> = vec![];
```
timvisee: Not just from proxies, but also the points that are in this optimized segment itself. And we need to deduplicate those, correct?

lib/shard/src/optimize.rs
```diff
@@ -475,7 +477,15 @@ fn finish_optimization(
...
+
+    read_segment_holder.deduplicate_points(&deferred_points, hw_counter)?;
```
timvisee: Imo we should also detach segment Arcs from lock here and flag the deletes without holding the segment holder. I see no issue with that assuming we look at point versions.

lib/shard/src/optimize.rs
generall: Yes, I mean the variable is called `proxies`, but we need to unwrap them

lib/shard/src/optimize.rs
generall: I am not sure I understood what you mean

timvisee: In short: we don't have to hold a read lock on the segment holder while deleting old versions for 'deferred points'.

generall: but we need to make sure we clean up all segments. I would prefer to still hold the read lock tbh.

lib/collection/src/tests/deferred_points_dedup.rs
```diff
@@ -0,0 +1,257 @@
...
+        metadata: None,
+    };
+
+    let payload_index_schema_dir = Builder::new()
+        .prefix("qdrant-test-schema")
+        .tempdir()
+        .unwrap();
+    let payload_index_schema_file = payload_index_schema_dir.path().join("payload.json");
+    let payload_index_schema: Arc<SaveOnDisk<PayloadIndexSchema>> =
+        Arc::new(SaveOnDisk::load_or_init_default(payload_index_schema_file).unwrap());
```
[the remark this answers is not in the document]
IvanPleshkov: fixed

lib/segment/src/segment/entry.rs
```diff
@@ -877,6 +877,18 @@ impl NonAppendableSegmentEntry for Segment {
...
     }
+
+    fn deferred_point_ids(&self) -> Vec<PointIdType> {
+        let Some(deferred_from) = self.deferred_internal_id else {
+            return vec![];
+        };
+        let id_tracker = self.id_tracker.borrow();
+        id_tracker
+            .iter_internal()
+            .filter(move |&internal_id| internal_id >= deferred_from)
```
timvisee: The iterator is ordered by internal ID, so we can do:

```suggestion
            .skip_while(move |&internal_id| internal_id < deferred_from)
```

lib/shard/src/segment_holder/mod.rs
```diff
@@ -530,36 +531,53 @@ impl SegmentHolder {
...
+    /// It scans all segments for presence of the points, detects points with the highest version,
+    /// and removes all other versions of the points from all segments.
+    pub fn deduplicate_points(
+        &self,
+        points: &[PointIdType],
+        hw_counter: &HardwareCounterCell,
+    ) -> OperationResult<()> {
+        let (_to_keep, to_delete) = self.find_points_to_update_and_delete(points);
+        self.delete_points_from_segments(to_delete, hw_counter)
     }
```
timvisee, line 581: After indexing, we should only delete older point versions.

I see this abuses the existing deduplication logic to make that happen. Because the deferred points we just indexed are now visible, those will remain. Also, other deferred points will stick because we don't touch them.

So it works as intended. Just stating it here because I was a bit confused about it first.

lib/segment/src/segment/entry.rs
agourlay: Nice :+1:

lib/collection/src/tests/deferred_points_dedup.rs
```diff
@@ -0,0 +1,253 @@
...
+    // Wait for optimization triggered by the overwrites
+    wait_optimization(&shard, timeout).await;
+
+    // Step 3: Overwrite again — this creates another round of CoW on top of the
+    // previous deferred points, stressing the deduplication logic further.
+    shard
+        .update(upsert_op(random_points()), true, None, hw_acc.clone())
+        .await
+        .unwrap();
+
```
agourlay, line 256: Can we actually assert that there are indeed deferred points before running the optimization.

Without it it is not clear if the optimization actually did the dedup :)

lib/shard/src/segment_holder/mod.rs
IvanPleshkov: I agree it's confusing. Added a loong comment in optimizer, where we use `deduplicate_points`

lib/segment/src/segment/entry.rs
IvanPleshkov: I like this suggestion, done

lib/collection/src/tests/deferred_points_dedup.rs
IvanPleshkov: Nice, check that we DO deferred points deduplication.
Done

lib/shard/src/optimize.rs
```diff
@@ -480,7 +483,26 @@ fn finish_optimization(
...
+    let deferred_points: Vec<PointIdType> = deferred_points_set.into_iter().collect();
+
+    if !deferred_points.is_empty() {
+        // Deferred points in proxy segment may become visible for optimized segment (in most cases).
+        // It's time to deduplicate them and remove older versions from optimized segment,
+        // where they were visible while deferred status.
+        // There are a situations, when deferred point is still deferred after optimization,
+        // so `deduplicate_points` also cover this case and delete only older versions of the point,
+        // which are still visible for optimized segment.
+        read_segment_holder.deduplicate_points(&deferred_points, hw_counter)?;
```
timvisee, line 502: Hot take: on startup we need to use the find duplicates function to locate points with duplicates first.

In this specific case we already know what points are deferred and what version they have. It's more likely to have duplicates.

I wonder if sending a delete operation on all those point IDs (version aware) to all segments is faster, versus trying to resolve duplicates first using that find duplicates function.

Let's not stall the review. But we can experiment with this idea in a separate PR (after patch release) if it sounds reasonable.

IvanPleshkov: Mentioned this idea in the design doc in the new section with issues after release

PR #8223, opened 2026-02-24
Universal IO: gridstore pages

generall:
- **[manunal] Gridstore page use universal IO**
- **fmt**

### DIFF
--- a/lib/common/common/src/universal_io/mmap.rs
+++ b/lib/common/common/src/universal_io/mmap.rs
@@ -6,7 +6,8 @@ use crate::mmap::{
     open_write_mmap,
 };
 use crate::universal_io::{
-    ByteOffset, BytesRange, Flusher, OpenOptions, UniversalIoError, UniversalRead, UniversalWrite,
+    ByteOffset, BytesRange, Flusher, OpenOptions, Result, UniversalIoError, UniversalRead,
+    UniversalWrite,
 };
 
 #[derive(Debug)]
@@ -26,7 +27,7 @@ impl<T> UniversalRead<T> for MmapUniversal<T>
 where
     T: Copy + 'static,
 {
-    fn open(path: impl AsRef<Path>, options: OpenOptions) -> crate::universal_io::Result<Self>
+    fn open(path: impl AsRef<Path>, options: OpenOptions) -> Result<Self>
     where
         Self: Sized,
     {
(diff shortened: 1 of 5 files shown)


lib/common/common/src/universal_io/mmap.rs
```diff
@@ -102,6 +102,10 @@ where
         Ok(())
     }
 
+    fn len(&self) -> crate::universal_io::Result<u64> {
+        Ok(self.mmap.len() as u64)
+    }
```
ffuugoo: Nit: Can we `use super::Result` and just do `Result` instead of `crate::universal_io::Result` in this file?

lib/gridstore/src/gridstore.rs
```diff
@@ -266,20 +268,23 @@ impl<V: Blob> Gridstore<V> {
...
+    ) -> Result<Option<V>> {
+        let Some(ValuePointer {
             page_id,
             block_offset,
             length,
-        } = self.get_pointer(point_offset)?;
+        }) = self.get_pointer(point_offset)
+        else {
+            return Ok(None);
+        };
```
ffuugoo:
```suggestion
        let Some(pointer) = self.get_pointer(point_offset) else {
            return Ok(None);
        };

        let ValuePointer {
            page_id,
            block_offset,
            length,
        } = pointer;
```

lib/gridstore/src/gridstore.rs
```diff
@@ -458,17 +464,20 @@ impl<V: Blob> Gridstore<V> {
...
+        let Some(ValuePointer {
             page_id,
             block_offset,
             length,
-        } = self.tracker.write().unset(point_offset)?;
-        let raw = self.read_from_pages::<false>(page_id, block_offset, length);
+        }) = self.tracker.write().unset(point_offset)
+        else {
+            return Ok(None);
+        };
```
ffuugoo:
```suggestion
        let Some(pointer) = self.tracker.write().unset(point_offset) else {
            return Ok(None);
        };

        let ValuePointer {
            page_id,
            block_offset,
            length,
        } = pointer;
```

lib/gridstore/src/gridstore.rs
```diff
@@ -530,7 +539,7 @@ impl<V: Blob> Gridstore<V> {
         &self,
         mut callback: F,
         hw_counter: HwMetricRefCounter,
-    ) -> std::result::Result<(), E>
+    ) -> std::result::Result<(), Either<E, GridstoreError>>
```
ffuugoo: Would `where E: From<GridstoreError>` work here? This way we can get rid of `Either`, `?` would work on `GridstoreError`.

generall: Unfortunatelly no, this error is propagated from crates and there are not guaranteed to be compatible with Gridstore error in any way

ffuugoo: I mean, it's `callback`, so we can _require_ it to be compatible. Or the other way round, make it `where GridstoreError: From<E>` and then if you modify code that require new error subtype in `GridstoreError`, you just add it.

ffuugoo: [`where E: From<GridstoreError>` works fine](https://github.com/qdrant/qdrant/pull/8231)

**UPD:** Didn't notice you already implemented that in #8230 lol

PR #8045, opened 2026-02-03
feat/edge-facet-api

dancixx:
### 1. Start Qdrant in Docker
```bash
docker run -d --name qdrant-test -p 6333:6333 -p 6334:6334 qdrant/qdrant:latest
```
### 2. Create a collection
```bash
curl -X PUT 'http://localhost:6333/collections/test_facet' \
  -H 'Content-Type: application/json' \
  -d '{"vectors": {"size": 4, "distance": "Dot"}}'
```
### 3. Insert points with payload
```bash
curl -X PUT 'http://localhost:6333/collections/test_facet/points?wait=true' \
  -H 'Content-Type: application/json' \
  -d '{
    "points": [
      {"id": 1, "vector": [0.1, 0.2, 0.3, 0.4], "payload": {"color": "red", "city": "Berlin"}},
      {"id": 2, "vector": [0.2, 0.3, 0.4, 0.5], "payload": {"color": "blue", "city": "Paris"}},
      {"id": 3, "vector": [0.3, 0.4, 0.5, 0.6], "payload": {"color": "green", "city": "London"}},
      {"id": 4, "vector": [0.4, 0.5, 0.6, 0.7], "payload": {"color": "red", "city": "London"}},
      {"id": 5, "vector": [0.5, 0.6, 0.7, 0.8], "payload": {"color": "blue", "city": "Berlin"}},
      {"id": 6, "vector": [0.6, 0.7, 0.8, 0.9], "payload": {"color": "red", "city": "Paris"}},
      {"id": 7, "vector": [0.7, 0.8, 0.9, 1.0], "payload": {"color": "green", "city": "Berlin"}},
      {"id": 8, "vector": [0.8, 0.9, 1.0, 1.1], "payload": {"color": "yellow", "city": "Tokyo"}},
      {"id": 9, "vector": [0.9, 1.0, 1.1, 1.2], "payload": {"color": "blue", "city": "Seoul"}},
      {"id": 10, "vector": [1.0, 1.1, 1.2, 1.3], "payload": {"color": "red", "city": "Tokyo"}}
    ]
  }'
```
### 4. Create payload indexes (required for faceting)
```bash
curl -X PUT 'http://localhost:6333/collections/test_facet/index' \
  -H 'Content-Type: application/json' \
  -d '{"field_name": "color", "field_schema": "keyword"}'

curl -X PUT 'http://localhost:6333/collections/test_facet/index' \
  -H 'Content-Type: application/json' \
[...]

PR #8224, opened 2026-02-24
feat/edge segment opt

dancixx:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/edge/src/optimize.rs
+++ b/lib/edge/src/optimize.rs
@@ -26,7 +26,7 @@ impl EdgeShard {
     ///
     /// This is synchronous and does not spawn background optimization workers.
     pub fn optimize(&self) -> OperationResult<bool> {
-        let optimizers = self.build_blocking_optimizers()?;
+        let optimizers = self.build_blocking_optimizers();
         let stopped = AtomicBool::new(false);
         let mut optimized_any = false;
 
@@ -77,7 +77,7 @@ impl EdgeShard {
         }
     }
 
-    fn build_blocking_optimizers(&self) -> OperationResult<Vec<Arc<Optimizer>>> {
+    fn build_blocking_optimizers(&self) -> Vec<Arc<Optimizer>> {
         let segments_path = self.path.join(SEGMENTS_PATH);
         let temp_segments_path = self.path.join(TEMP_SEGMENTS_PATH);
 
@@ -88,7 +88,7 @@ impl EdgeShard {
         let threshold_config = Self::default_optimizer_thresholds(hnsw_config);
         let default_segments_number = default_segment_number();
 
-        Ok(vec![
+        vec![
             Arc::new(MergeOptimizer::new(
                 default_segments_number,
                 threshold_config,
@@ -125,7 +125,7 @@ impl EdgeShard {
                 hnsw_config,
                 hnsw_global_config,
             )),
-        ])
+        ]
     }
 
     fn default_optimizer_thresholds(hnsw_config: HnswConfig) -> OptimizerThresholds {


lib/edge/src/optimize.rs
```diff
@@ -0,0 +1,969 @@
...
+
+        result
+    }
+
+    fn build_blocking_optimizers(&self) -> OperationResult<Vec<Arc<Optimizer>>> {
+        let segments_path = self.path.join(SEGMENTS_PATH);
+        let temp_segments_path = self.path.join(EDGE_OPTIMIZER_TEMP_PATH);
+        Self::reset_temp_segments_dir(&temp_segments_path)?;
+
+        let hnsw_config = HnswConfig::default();
```
generall, line 84: This configuration needs to be either inferred from existing segments, or provided by the user in arguments. Only of neither present we can fallback to default.

lib/edge/src/optimize.rs
```diff
@@ -0,0 +1,969 @@
...
+                temp_segments_path.display(),
+            ))
+        })?;
+
+        Ok(())
+    }
+}
+
+/// Target segment count for the merge optimizer.
+/// Mirrors `OptimizerConfig::get_number_segments` from `lib/collection/src/optimizers_builder.rs`.
```
generall: let's move function inside `shard` crate

lib/edge/src/optimize.rs
```diff
@@ -0,0 +1,969 @@
...
+            memmap_threshold_kb: usize::MAX,
+            indexing_threshold_kb: DEFAULT_INDEXING_THRESHOLD_KB,
+            max_segment_size_kb: indexing_threads.saturating_mul(DEFAULT_MAX_SEGMENT_PER_CPU_KB),
+            deferred_points_threshold_bytes: None,
+        }
+    }
+
+    fn build_segment_optimizer_config(
+        &self,
+        default_hnsw_config: HnswConfig,
```
generall: just `hnsw_config`?

lib/edge/src/optimize.rs
```diff
@@ -0,0 +1,969 @@
...
+        let indexing_threads = num_rayon_threads(hnsw_config.max_indexing_threads);
+        OptimizerThresholds {
+            memmap_threshold_kb: usize::MAX,
+            indexing_threshold_kb: DEFAULT_INDEXING_THRESHOLD_KB,
+            max_segment_size_kb: indexing_threads.saturating_mul(DEFAULT_MAX_SEGMENT_PER_CPU_KB),
+            deferred_points_threshold_bytes: None,
+        }
+    }
+
+    fn build_segment_optimizer_config(
```
generall: let's try to make unified logic with collection and put it into `shard`. If there are some types used from collection crate, just implement adapters/converters in `shard`

lib/edge/src/optimize.rs
```diff
@@ -0,0 +1,939 @@
...
+
+use crate::{EdgeShard, SEGMENTS_PATH};
+
+impl EdgeShard {
+    /// Run shard optimizers in-process and blocking until no more optimization plans are produced.
+    ///
+    /// This is synchronous and does not spawn background optimization workers.
+    ///
+    /// NOTE: This method is not safe to call concurrently — it is only
+    /// intended to be called once during [`EdgeShard::load()`].
```
generall: This seems to be outdated, also I think it might be actually safe to run this function concurrently. And if it is not, I propose to enforce it by having `&mut self` as a reference.

lib/edge/src/optimize.rs
```diff
@@ -0,0 +1,939 @@
...
+        }
+    }
+
+    /// Infer HNSW config from the first indexed vector in the segment config.
+    /// Falls back to default if no HNSW-indexed vector exists (e.g. all Plain).
+    fn infer_hnsw_config(&self) -> HnswConfig {
+        self.config
+            .vector_data
+            .values()
+            .find_map(|vdc| match &vdc.index {
```
generall: This is not correct. This function tries to find hnsw config for any vector, and applies it to all vectors. It should preserve `name -> config` mapping instead.

lib/edge/python/src/lib.rs
```diff
@@ -111,6 +111,11 @@ impl PyEdgeShard {
         Ok(())
     }
 
+    pub fn optimize(&self) -> Result<bool> {
```
generall, line 114: It is ok for this PR, but more I think about it, more I understand that we would need to allow index configuration here.

dancixx: seperate pr: https://github.com/qdrant/qdrant/pull/8304

lib/edge/src/update.rs
```diff
@@ -16,6 +16,7 @@ impl EdgeShard {
 
         let operation_id = wal.write(&record).map_err(service_error)?;
         let hw_counter = HardwareCounterCell::disposable();
+        let _update_guard = self.segments.acquire_updates_lock();
```
generall, line 19: good catch!

lib/edge/src/lib.rs
```diff
@@ -1,12 +1,13 @@
...
+pub mod count;
+pub mod facet;
+pub mod info;
+pub mod optimize;
+pub mod query;
+pub mod retrieve;
+pub mod scroll;
+pub mod search;
+pub mod snapshots;
+pub mod update;
```
xzfc, line 10: FYI: we don't want these to be public in the `qdrant-edge` crate.

dancixx: @xzfc thanks. i will revert it in a seperate pr.

lib/edge/python/examples/optimize.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python3
+"""
+Test edge optimize by downloading a snapshot from a running Qdrant instance.
+
+Requires a running Qdrant on localhost:6333 with a 'bench_dense_384' collection.
```
xzfc, line 5: How to get that collection?

I'm trying to set up automated tests all edge examples: https://github.com/qdrant/qdrant/actions/runs/22778507330/job/66077936830?pr=831

We already have some examples that need data from Qdrant:

1. - prepare step
     - Runs on CI.
     - Based steps you provided #8045
     - Source: https://github.com/qdrant/qdrant/blob/61e046f80e67ea6787686af2f0eb3fe8398a6253/lib/edge/publish/examples/prepare_facet_snapshot.sh
   - run step
     - Source: https://github.com/qdrant/qdrant/blob/61e046f80e67ea6787686af2f0eb3fe8398a6253/lib/edge/publish/examples/src/bin/facet_test.rs

2. - prepare step
     - Runs manually, then manually uploaded to google storage (already done)
     - Source: https://github.com/qdrant/qdrant/blob/61e046f80e67ea6787686af2f0eb3fe8398a6253/lib/edge/python/examples/download_partial.py
   - run step
     - Downloads from Google storage.
     - Source: https://github.com/qdrant/qdrant/blob/61e046f80e67ea6787686af2f0eb3fe8398a6253/lib/edge/python/examples/restore-snapshot.py

So, we need some automation for the prepare step for this snippet too.
(if it's trivial/fast, then maybe this example should do it automatically using qdrant_client)

PR #8228, opened 2026-02-25
Indexing threshold in segment config

IvanPleshkov:
This PR addes indexing threshold parameter into the internal segment config and does an integration where we decide where we need this new parameter.

### DIFF
--- a/lib/shard/src/segment_holder/tests.rs
+++ b/lib/shard/src/segment_holder/tests.rs
@@ -570,8 +570,13 @@ fn test_double_proxies() {
         .unwrap();
 
     let (outer_proxies, outer_tmp_segment, outer_segments_lock) =
-        SegmentHolder::proxy_all_segments(inner_segments_lock, segments_dir.path(), &segment_config, schema)
-            .unwrap();
+        SegmentHolder::proxy_all_segments(
+            inner_segments_lock,
+            segments_dir.path(),
+            &segment_config,
+            schema,
+        )
+        .unwrap();
 
     let mut has_point = false;
     for (_proxy_id, proxy) in &outer_proxies {


lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -607,6 +607,7 @@ impl LocalShard {
                 vector_data: vector_params.clone(),
                 sparse_vector_data: sparse_vector_params.clone(),
                 payload_storage_type: config.params.payload_storage_type(),
+                indexing_threshold_kb: config.optimizer_config.get_deferred_segment_threshold_kb(),
```
JojiiOfficial, line 610: We set `indexing_threshold_kb` here for **all segments** if `prevent_unoptimized` is `true`.
However, we don't check if the segment is actually appendable. Therefore I'd suggest to add a check here whether the segment is appendable or not.

Even if the config option doesn't have a direct influence on non-appendable segments, I'd still suggest to not store this parameter for consistency.

lib/collection/src/config.rs
```diff
@@ -621,6 +622,7 @@ impl CollectionParams {
             vector_data,
             sparse_vector_data,
             payload_storage_type,
+            indexing_threshold_kb: optimizers_config.get_deferred_segment_threshold_kb(),
```
JojiiOfficial, line 625: Same here. We use it in two places,
1. https://github.com/qdrant/qdrant/blob/master/lib/collection/src/shards/local_shard/snapshot.rs#L68
2. https://github.com/qdrant/qdrant/blob/master/lib/collection/src/shards/local_shard/mod.rs#L487

The second location is fine because we always use it to create an appendable segment anyways. But for snapshots we _might_ not want it as it uses this config to create temporary appendable segments for proxy shards.

lib/segment/src/types.rs
```diff
@@ -1377,6 +1377,9 @@ pub struct SegmentConfig {
     pub sparse_vector_data: HashMap<VectorNameBuf, SparseVectorDataConfig>,
     /// Defines payload storage type
     pub payload_storage_type: PayloadStorageType,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub indexing_threshold_kb: Option<usize>,
```
JojiiOfficial, line 1382: I'd vote to change this name to `deferring_threshold_kb` or similar.
We set this option only if deferred mode is enabled and renaming to something with `deferred` makes the semantic more clear. Otherwise, I'd suggest to always set this value for consistency.

lib/collection/src/update_workers/optimization_worker.rs
```diff
@@ -450,6 +450,9 @@ impl UpdateWorkers {
                 .into_iter()
                 .filter_map(|segment_id| segments_read.get(segment_id))
                 .all(|segment| {
+                    // TODO: if we have deferred segment, we might want to continue growing it's size.
```
JojiiOfficial: I opened https://github.com/qdrant/qdrant/pull/8237 to address this

lib/collection/src/shards/local_shard/mod.rs
IvanPleshkov: This is an appendable segment because of how we construct `vector_params`. We use plain version of it:
```
let vector_params = config
    .params
    .to_base_vector_data(config.quantization_config.as_ref());
```

lib/collection/src/config.rs
IvanPleshkov: For snapshot we might not want, true. Let me think about it

IvanPleshkov:
Closed because of https://github.com/qdrant/qdrant/pull/8246

PR #8232, opened 2026-02-25
use `CowMultiVector` as return type in multivec storages

coszio:
Similar to #8221, but for multi-dense storages.

This one comes with a slight (and more explicit) overhead when dereferencing, since the borrowed type is an owned struct with a lifetime, it needs to copy the `dim` and the slice's fat pointer when doing `as_ref`.

There is an alternative, that comes with a higher refactoring cost: To just return the flattened vec, and read the `dim` field at the usage sites. It would require lots of extra changes though, so I decided not to do it and use this cheap-copying approach instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2880,6 +2880,7 @@ version = "2.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6ea2d84b969582b4b1864a92dc5d27cd2b77b622a8d79306834f1be5ba20d84b"
 dependencies = [
+ "bytemuck",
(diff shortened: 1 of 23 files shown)


ffuugoo (approved):
Seems good, but can we close this PR and include these changes into #8233 instead? It's hard to review on its own, because without context from #8233 it's hard to see if changes in this PR make sense or not. 😬

coszio:
> can we close this PR and include these changes into https://github.com/qdrant/qdrant/pull/8233 instead?

Sure, I am fine with that if you prefer a larger, self-contained PR.

PR #8242, opened 2026-02-26
fix(startup): avoid double panic in telemetry error reporting

### LINKED ISSUE #7831: Qdrant 1.16.3 crashes on startup with “failed to set up alternative stack guard page: Cannot allocate memory (os error 12)” after starting HTTP server
## Current Behavior
Qdrant starts normally and successfully:
- Recovers a large number of collections and shards
- Binds REST (`6333`) and gRPC (`6334`) endpoints
- Starts Actix with 13 workers

However, a few seconds after startup, the process panics and aborts with a core dump. The crash occurs after the server reports that it is listening for requests.

Error observed:
```yaml
2025-12-24T22:28:31.083448Z  INFO qdrant: Distributed mode disabled
2025-12-24T22:28:31.084330Z  INFO qdrant: Telemetry reporting enabled, id: b6633f4d-c87c-4c74-b8fe-3422dd55fbfa
2025-12-24T22:28:31.177826Z  INFO qdrant::tonic: Qdrant gRPC listening on 6334
2025-12-24T22:28:31.177860Z  INFO qdrant::tonic: TLS disabled for gRPC API
2025-12-24T22:28:31.180815Z  INFO qdrant::actix: TLS disabled for REST API
2025-12-24T22:28:31.181563Z  INFO qdrant::actix: Qdrant HTTP listening on 6333
2025-12-24T22:28:31.181577Z  INFO actix_server::builder: starting 13 workers
2025-12-24T22:28:31.181586Z  INFO actix_server::server: Actix runtime found; starting in Actix runtime
2025-12-24T22:28:31.181594Z  INFO actix_server::server: starting service: "actix-web-service-0.0.0.0:6333", workers: 13, listening on: 0.0.0.0:6333
2025-12-24T22:28:35.020403Z ERROR qdrant::startup: Panic backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
panicked at src/common/error_reporting.rs:20:14:

thread panicked while processing panic. aborting.
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>

2025-12-24T22:28:35.020429Z ERROR qdrant::startup: Panic occurred in file library/std/src/sys/pal/unix/stack_overflow.rs at line 231: failed to set up alternative stack guard page: Cannot allocate memory (os error 12)
./entrypoint.sh: line 25:     7 Aborted                 (core dumped) ./qdrant $@
           _                 _
  __ _  __| |_ __ __ _ _ __ | |_
 / _` |/ _` | '__/ _` | '_ \| __|
| (_| | (_| | | | (_| | | | | |_
 \__, |\__,_|_|  \__,_|_| |_|\__|
    |_|

Version: 1.16.3, build: bd49f45a
Access web UI at http://localhost:6333/dashboard
```

The container restarts and the same crash repeats on every startup.

## Steps to Reproduce
1. Run Qdrant `v1.16.3` (build `bd49f45a`) using the default `entrypoint.sh` (Docker/containerized setup).
2. Use an existing `./storage` directory containing a large number of collections and shards.
3. Start Qdrant normally with:
   - Distributed mode disabled
   - REST API on port `6333`
   - gRPC API on port `6334`
4. Observe that Qdrant:
   - Recovers all shards successfully
   - Starts Actix and begins listening on HTTP/gRPC
   - Crashes shortly after with an out-of-memory related panic.

## Expected Behavior
Qdrant should either:
- Start successfully and remain running after shard recovery and server initialization, **or**
- Fail gracefully with a clear, actionable error message indicating insufficient memory.

It should not panic and abort after partially starting the service.

## Possible Solution
- Handle low-memory conditions more gracefully instead of panicking.
- Add clearer diagnostics during startup (available memory, memory limits, worker count).
- Make Actix worker count configurable or adaptive to available memory.
- Document minimum memory requirements for startup with many collections/shards.

## Context (Environment)
- **Qdrant Version:** `1.16.3`
- **Build:** `bd49f45a`
- **Mode:** Distributed mode disabled
- **APIs:** REST (`6333`), gRPC (`6334`)
- **Startup Logs:** Shard recovery completes successfully for many collections before crash.
- **Impact:** Qdrant cannot stay running, causing a restart loop and preventing application usage.

## Detailed Description
Qdrant consistently crashes during startup in a memory-constrained environment. The panic occurs *after* shard recovery and *after* the HTTP and gRPC servers report that they are listening.

The error originates from Rust’s stack overflow handling when attempting to allocate an alternative stack guard page, which fails due to insufficient memory. A secondary panic occurs while handling the initial panic, resulting in an abort and core dump.

This behavior makes Qdrant unstable when restarting with a large number of existing collections.

## Possible Implementation
- Perform a preflight check against container or cgroup memory limits at startup.
- Reduce default Actix worker count when memory is limited.
- Convert this panic path into a structured startup error with:
  - Available memory
  - Memory limits
  - Worker count
  - Number of collections detected
  - Suggested remediation steps (increase memory, reduce workers, prune collections)

## Docker Compose Configuration
The following `docker-compose.yml` is used to reproduce the issue:

```yaml
services:
  qdrant:
    image: qdrant/qdrant:latest
    container_name: qdrant_container
    restart: unless-stopped

    ports:
      - "6333:6333"
      - "6334:6334"

    environment:
      QDRANT__SERVICE__HTTP_PORT: 6333
      QDRANT__SERVICE__GRPC_PORT: 6334

      QDRANT__STORAGE__SNAPSHOTS_PATH: /qdrant/snapshots
      QDRANT__STORAGE__ON_DISK: "true"

      QDRANT__CLUSTER__ENABLED: "false"
      QDRANT__LOG_LEVEL: INFO
      QDRANT__TELEMETRY__DISABLE: "true"

      QDRANT__OPTIMIZERS__DEFAULT_SEGMENT_NUMBER: 3
      QDRANT__OPTIMIZERS__MEMMAP_THRESHOLD_KB: 3000000

    volumes:
      - ./qdrant_data:/qdrant/storage
      - ./qdrant_snapshots:/qdrant/snapshots

    healthcheck:
      test: ["CMD", "wget", "-qO-", "http://localhost:6333/healthz"]
      interval: 30s
      timeout: 5s
      retries: 3
      start_period: 10s

    ulimits:
      nofile:
        soft: 65535
        hard: 65535
```

krapcys1-maker:
Fixes #7831

## Summary

This PR hardens startup panic telemetry so low-memory panic paths do not trigger a secondary panic in the error reporter.

## What changed

- Removed panic-prone unwraps from ErrorReporter::report and replaced them with guarded error handling/logging.
- Added a dedicated payload builder helper and unit tests for with/without-backtrace serialization.
- Kept behavior backward-compatible: telemetry reporting is best-effort and silently degrades on reporter/client failures.

## Why

Issue #7831 shows abort loops with 'thread panicked while processing panic' after an initial OOM-related panic. The previous report path could panic itself (client build/JSON serialization unwrap), causing a double panic. This change makes the reporting path non-panicking to preserve a single primary failure signal.

## Tests

- `cargo test -p qdrant error_reporting -- --nocapture`
- `cargo +nightly fmt --all`
- `cargo clippy -p qdrant --all-targets -- -D warnings`

### DIFF
--- a/src/common/error_reporting.rs
+++ b/src/common/error_reporting.rs
@@ -1,6 +1,7 @@
 use std::time::Duration;
 
 use common::defaults::APP_USER_AGENT;
+use serde_json::json;
 
 pub struct ErrorReporter;
 
@@ -13,24 +14,63 @@ impl ErrorReporter {
         }
     }
 
-    pub fn report(error: &str, reporting_id: &str, backtrace: Option<&str>) {
-        let client = reqwest::blocking::Client::builder()
-            .user_agent(APP_USER_AGENT.as_str())
-            .build()
-            .unwrap();
-
-        let report = serde_json::json!({
+    /// Build serialized JSON payload for telemetry error reporting.
+    fn build_report_payload(error: &str, reporting_id: &str, backtrace: Option<&str>) -> String {
+        let report = json!({
             "id": reporting_id,
             "error": error,
-            "backtrace": backtrace.unwrap_or(""),
+            "backtrace": backtrace,
         });
+        report.to_string()
+    }
 
-        let data = serde_json::to_string(&report).unwrap();
-        let _resp = client
+    pub fn report(error: &str, reporting_id: &str, backtrace: Option<&str>) {
+        let client = match reqwest::blocking::Client::builder()
+            .user_agent(APP_USER_AGENT.as_str())
+            .build()
+        {
+            Ok(client) => client,
+            Err(err) => {
+                log::warn!("Failed to build telemetry reporter client: {err}");
+                return;
+            }
+        };
+
+        let data = Self::build_report_payload(error, reporting_id, backtrace);
+
+        if let Err(err) = client
             .post(Self::get_url())
             .body(data)
             .header("Content-Type", "application/json")
             .timeout(Duration::from_secs(1))
-            .send();
+            .send()
+        {
+            log::debug!("Telemetry panic report was not sent: {err}");
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::ErrorReporter;
+
+    #[test]
+    /// Ensure panic report payload contains the expected fields.
+    fn test_build_report_payload_with_backtrace() {
+        let payload = ErrorReporter::build_report_payload("panic", "node-1", Some("bt-line"));
+
+        assert!(payload.contains("\"id\":\"node-1\""));
+        assert!(payload.contains("\"error\":\"panic\""));
+        assert!(payload.contains("\"backtrace\":\"bt-line\""));
+    }
+
+    #[test]
+    /// Missing backtrace must serialize as null.
+    fn test_build_report_payload_without_backtrace() {
+        let payload = ErrorReporter::build_report_payload("panic", "node-2", None);
+
+        assert!(payload.contains("\"id\":\"node-2\""));
+        assert!(payload.contains("\"error\":\"panic\""));
+        assert!(payload.contains("\"backtrace\":null"));
     }
 }


src/common/error_reporting.rs
```diff
@@ -13,24 +14,75 @@ impl ErrorReporter {
...
+    /// Build serialized JSON payload for telemetry error reporting.
+    fn build_report_payload(
+        error: &str,
+        reporting_id: &str,
+        backtrace: Option<&str>,
+    ) -> Result<String, serde_json::Error> {
+        let report = json!({
             "id": reporting_id,
             "error": error,
             "backtrace": backtrace.unwrap_or(""),
```
timvisee: How about this, where backtrace would become `null` if not provided?

```suggestion
            "backtrace": backtrace,
```

src/common/error_reporting.rs
```diff
@@ -13,24 +14,75 @@ impl ErrorReporter {
...
+
+        let data = match Self::build_report_payload(error, reporting_id, backtrace) {
+            Ok(payload) => payload,
+            Err(err) => {
+                log::warn!("Failed to serialize telemetry panic payload: {err}");
+                return;
+            }
+        };
 
-        let data = serde_json::to_string(&report).unwrap();
```
timvisee, line 28: Is this unwrap really a concern? Can we ever trigger a panic here?

src/common/error_reporting.rs
krapcys1-maker: Good point — `unwrap_or("")` itself is not panic-prone, but I agree that using `"backtrace": backtrace` (so it becomes `null` when missing) is semantically cleaner. I’ll switch to that.

src/common/error_reporting.rs
krapcys1-maker: For this specific payload (`serde_json::Value` built from string/option fields), `serde_json::to_string` is effectively infallible in practice.

The actual double-panic risk we wanted to remove is the client initialization path:
`reqwest::blocking::Client::builder().build().unwrap()` inside the panic hook.

So this change mainly improves failure-path safety (panic hook execution), not the normal path.

krapcys1-maker: I simplified this path further:

- `backtrace` is now serialized as `Option` (`null` when absent), no `unwrap_or("")`.
- payload serialization now uses `report.to_string()` (no `serde_json::to_string(...).unwrap()`).

So the panic hook path has no remaining unwrap/panic sources from JSON/backtrace handling.

PR #8246, opened 2026-02-27
Deferred threshold integration

IvanPleshkov:
## Summary
This PR introduces a deferred threshold for segments with a deferred internal id.

Part of the deferred points feature, described here:
https://www.notion.so/qdrant/Deferred-points-in-appendable-segments-311674779d33808a9e97f9449267b2f6

## Changes list

This PR adds two new fields into `Segment` struct:
```rust
deferred_points_threshold_bytes: Option<NonZeroUsize>,
deferred_internal_id: Option<PointOffsetType>,
```

`deferred_points_threshold_bytes` is presented in each segment, including indexed ones; it's just a propagation data from the collection - the value is `Some` if `prevent_unoptimized = true` and `indexing_threshold > 0`.

Segment owns the logic of deferred points and decided himself if he has deferred points inside.
On the segment level, we disable the deferred points feature in case when:
- Segment is non-appendable
- There are no dense vector storages

The segment decides that he has deferred points, he sets `Some` for `deferred_internal_id`, which describes the start internal ID to filter in read operations (including `deferred_internal_id`).

PR has a function `update_deferred_internal_id()`. It computes the deferred ID from vector storage layout sizes, taking the minimum across all non-multivector data. Called on segment creation/load and after each point insertion (put_all_vectors, upsert_vector, insert).

`has_deferred_points()` — quick check used by the indexing optimizer to trigger optimization on segments that have deferred points, ensuring they get indexed promptly.

Threshold derived from config — get_deferred_points_threshold_bytes() on OptimizersConfig returns Some(indexing_threshold_kb * 1024) when prevent_unoptimized == true, None otherwise.

All callers updated — every call site for build_segment, load_segment, SegmentBuilder::build, proxy_all_segments, and snapshot_all_segments now passes the threshold (mostly None in tests/edge).

## Future work
For simpler review, multivectors and sparse vectors will be applied in the next changes of this milestone (it's marked in the design document plan).

### DIFF
--- a/lib/segment/src/segment/mod.rs
+++ b/lib/segment/src/segment/mod.rs
@@ -95,11 +95,13 @@ pub struct Segment {
     pub error_status: Option<SegmentFailedState>,
     #[cfg(feature = "rocksdb")]
     pub database: Option<Arc<parking_lot::RwLock<DB>>>,
-    /// If set, shows that all points with internal id >= deferred_internal_id
-    /// are not visible for read operations, but are already present in the segment.
+    /// Deferred-points threshold in bytes propagated from optimizer config.
+    /// If `None`, deferred-points behavior is disabled for this segment.
     pub(crate) deferred_points_threshold_bytes: Option<NonZeroUsize>,
-    // Cached deferred internal id,
-    // which is used for quick checks of point visibility without accessing vector storage.
+    /// Cached deferred internal ID for fast visibility checks.
+    /// Points with internal id >= this value are hidden from reads.
+    /// It's `None` if deferred_points_threshold_bytes is `None` or if there are no deferred points.
+    /// Also `None` for non-appendable segments, as they don't accept new points and thus don't have deferred points.
     pub(crate) deferred_internal_id: Option<PointOffsetType>,
 }
 


lib/segment/src/segment/segment_ops.rs
```diff
@@ -652,6 +652,42 @@ impl Segment {
...
+                        .borrow()
+                        .get_deferred_id(deferred_threshold)
+                })
+                .min();
+            if let Some(deferred_id) = deferred_id
+                && (deferred_id as usize) < self.total_point_count()
+            {
+                self.deferred_internal_id = Some(deferred_id);
+            }
+        }
```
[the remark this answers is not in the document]
IvanPleshkov: fixed, check please

lib/segment/src/vector_storage/vector_storage_base.rs
```diff
@@ -207,6 +208,17 @@ pub trait DenseVectorStorage<T: PrimitiveVectorElement>: VectorStorage {
...
+    /// Returns the point offset at which cumulative storage size reaches the threshold.
+    /// For dense storage, each vector has constant size, so simple division suffices.
+    fn get_deferred_id(&self, deferred_threshold: usize) -> Option<PointOffsetType> {
+        let size_per_vector = self.vector_dim() * std::mem::size_of::<T>();
+        if size_per_vector.is_zero() {
+            return Some(self.total_vector_count() as PointOffsetType);
+        }
+        let deferred_id = deferred_threshold / size_per_vector;
+        Some(deferred_id.min(self.total_vector_count()) as PointOffsetType)
+    }
```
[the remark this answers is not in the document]
IvanPleshkov: removed

lib/segment/src/vector_storage/vector_storage_base.rs
```diff
@@ -238,6 +250,41 @@ pub trait MultiVectorStorage<T: PrimitiveVectorElement>: VectorStorage {
...
+            let cumulative_bytes =
+                self.inner_vectors_count_up_to(mid as PointOffsetType) * bytes_per_inner_vector;
+            if cumulative_bytes >= deferred_threshold {
+                hi = mid;
+            } else {
+                lo = mid + 1;
+            }
+        }
+        Some(lo as PointOffsetType)
+    }
```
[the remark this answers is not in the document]
IvanPleshkov: removed

lib/segment/src/segment/segment_ops.rs
```diff
@@ -652,6 +653,42 @@ impl Segment {
...
+            return;
+        }
+
+        if let Some(deferred_threshold) = self.deferred_threshold {
+            // Corner case for zero threshold
+            if deferred_threshold.is_zero() {
+                return;
+            }
+
+            let is_deferred_reached = self.vector_data.values().any(|vector_data| {
```
IvanPleshkov: Please take additional attention here.
The main problem here is `size_of_available_vectors_in_bytes`, which depends on deletes.
I find it OK because we set `deferred_internal_id` once, and if the user deletes the point, we have an additional check in the index optimizer

lib/collection/src/optimizers_builder.rs
```diff
@@ -158,6 +160,11 @@ impl OptimizersConfig {
             num_indexing_threads.saturating_mul(DEFAULT_MAX_SEGMENT_PER_CPU_KB)
         }
     }
+
+    pub fn get_deferred_threshold(&self) -> Option<usize> {
```
JojiiOfficial: Suggestion

```suggestion
    pub fn get_deferred_threshold_bytes(&self) -> Option<usize> {
```

lib/segment/src/segment/segment_ops.rs
```diff
@@ -652,6 +653,42 @@ impl Segment {
...
         self.id_tracker.borrow_mut().fix_inconsistencies()
     }
+
+    pub fn has_deferred_points(&self) -> bool {
+        if let Some(deferred_internal_id) = self.deferred_internal_id {
+            (deferred_internal_id as usize) < self.total_point_count()
+        } else {
+            false
+        }
+    }
```
JojiiOfficial:
```suggestion
        self.deferred_internal_id.is_some_and(|deferred_internal_id| (deferred_internal_id as usize) < self.total_point_count())
    }
```

lib/collection/src/optimizers_builder.rs
IvanPleshkov: renamed everywhere

lib/segment/src/segment/segment_ops.rs
IvanPleshkov: We decided during the pair review session check for `Some` only

lib/shard/src/optimizers/indexing_optimizer.rs
```diff
@@ -83,7 +83,9 @@ impl IndexingOptimizer {
                     is_big_for_mmap && !is_on_disk
                 };
 
-                if optimize_for_index || optimize_for_mmap {
+                let has_deferred_points = segment.has_deferred_points();
```
IvanPleshkov: Attention point. Here, we update the index optimizer and do optimization if there are any deferred points

IvanPleshkov: everybody is agreed that we need this check in the indexing optimizer

lib/segment/src/segment/segment_ops.rs
IvanPleshkov: outdated, we use the total size

lib/segment/src/segment/segment_ops.rs
```diff
@@ -652,6 +656,50 @@ impl Segment {
...
+                        .get_vector_layout()
+                        .ok()
+                        .map(|layout| layout.size())
+                        .and_then(NonZeroUsize::new)?;
+                    let deferred_internal_id = deferred_points_threshold_bytes
+                        .get()
+                        .div_ceil(vector_size.get());
+                    if deferred_internal_id < storage.total_vector_count() {
+                        Some(deferred_internal_id as PointOffsetType)
+                    } else {
```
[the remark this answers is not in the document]
IvanPleshkov: total_vector_count cannot be larger than PointOffsetType::MAX, this conversion is safe

PR #8256, opened 2026-03-01
MultiFile Universal UI in gridstore pages

generall:
This PR takes multi-file universal UI and applies it to gridstore Pages.

The most critical part of the PR is reading values from Pages and the support of the case, where value can be splitted between multiple pages. 
Previously we had to do multiple read operations, but with multi-io we can read from all pages in one batch request.

This is not the full porential of multi-page reads, but further improvements belong to further PRs


generall:
This PR is waiting for the alternative interface suggestion from @ffuugoo

ffuugoo:
> This PR is waiting for the alternative interface suggestion from @ffuugoo

#8263

lib/gridstore/src/pages.rs
```diff
@@ -0,0 +1,137 @@
...
+                self.page_access_range(page_id, block_offset, length, block_size_bytes)?;
+
+            range_offset.push(length_so_far as usize);
+            length_so_far += element_range.length;
+
+            read_ranges.push((SourceId(page_id as usize), element_range));
+
+            block_offset = 0;
+            length = unread_bytes as u32;
+        }
```
xzfc, line 131: Should be `break` somewhere.

lib/common/common/src/maybe_uninit.rs
```diff
@@ -41,3 +41,18 @@ pub const unsafe fn assume_init_mut<T>(this: &mut [MaybeUninit<T>]) -> &mut [T]
...
+/// vec really is in an initialized state. For instance, `.assume_init_vec()` cannot
+/// be used to initialize a `MaybeUninit` vec.
+#[inline(always)]
+pub unsafe fn assume_init_vec<T>(this: Vec<MaybeUninit<T>>) -> Vec<T> {
+    // SAFETY: caller must guarantee every element was initialized.
+    unsafe {
+        let (ptr, len, cap) = this.into_raw_parts();
+        Vec::from_raw_parts(ptr.cast::<T>(), len, cap)
+    }
+}
```
xzfc, line 58: LGTM.

PR #8259, opened 2026-03-02
Introduce Segment API point_is_deferred

agourlay:
New segment API to check whether a point is deferred.

The interesting part is the addition of a test showing that our threshold detection on update is correct.

### DIFF
--- a/lib/segment/src/segment/tests.rs
+++ b/lib/segment/src/segment/tests.rs
@@ -32,6 +32,7 @@ fn init_logger() {
 
 #[test]
 fn test_search_batch_equivalence_single() {
+    init_logger();
     let dir = Builder::new().prefix("segment_dir").tempdir().unwrap();
     let dim = 4;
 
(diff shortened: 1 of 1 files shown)


lib/segment/src/segment/tests.rs
```diff
@@ -717,3 +724,88 @@ fn test_handle_point_version() {
...
+    // Initially, no deferred points (empty segment)
+    assert!(!segment.has_deferred_points());
+    for i in 0..=20 {
+        assert!(!segment.point_is_deferred((i as u64).into()));
+    }
+
+    // Insert points with small vectors to reach the threshold
+    // Each vector is 4 f32 = 16 bytes
+    // We need to insert enough vectors to exceed 200 bytes
+    // Points 1-12 will be non-deferred (internal_id 0-11), points 13+ will be deferred
```
timvisee: Is this description off by one?

Below we check `1..=13` and `14..=20`.

agourlay: Yeah, I fixed the test but not the comment :grimacing:

PR #8261, opened 2026-03-02
Explicit `point_mappings` guard in `IdTracker`

monoid:
The `IdTracker` methods do not return iterators anymore. Instead, you should take explicitely the `PointMappingsRefEnum` guard and use its methods.  It prepares for moving the `point_mappings` under a lock to allow parallel searches and deletions.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/id_tracker/id_tracker_base.rs
+++ b/lib/segment/src/id_tracker/id_tracker_base.rs
@@ -301,7 +301,7 @@ self_cell! {
     ///
     /// Used by `NonAppendableSegmentEntry::iter_points()` to return an iterator
     /// over external point IDs without creating the iterator inside the unsafe block.
-    pub(crate) struct PointMappingsGuard<'a> {
+    pub struct PointMappingsGuard<'a> {
         owner: AtomicRef<'a, IdTrackerEnum>,
 
         #[covariant]


lib/common/common/src/guards.rs
```diff
@@ -0,0 +1,56 @@
...
+/// Owning guard holder.
+pub struct OwningGuard<'ctx, Outer, Inner>
+where
+    Outer: 'ctx,
+{
+    // Fields are dropped in the order of their definition.  It is important to drop the `inner_value` first.
+    inner_value: Inner,
+    _outer_guard: Outer,
+    phantom: PhantomData<&'ctx ()>,
+}
```
xzfc: I think this struct can be replaced by `self_cell`. We already have a few usages of it.

lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -213,6 +183,90 @@ pub trait IdTracker: fmt::Debug {
...
+impl<'a> PointMappingsRefEnum<'a> {
+    /// Iterate over all external IDs.
+    ///
+    /// Excludes soft deleted points.
+    pub fn iter_external(self) -> Box<dyn Iterator<Item = PointIdType> + 'a> {
+        match self {
+            PointMappingsRefEnum::Plain(m) => m.iter_external(),
+            PointMappingsRefEnum::Compressed(m) => m.iter_external(),
+        }
+    }
```
xzfc, line 204: **Performance**: use enum dispatch instead of `Box<dyn Iterator>`.

`either::Either` implements iterator boilerplate for two-variant enums. So, it'll look like this:

```rust
pub fn iter_external(self) -> impl Iterator<Item = PointIdType> {
    match self {
        PointMappingsRefEnum::Plain(m) => Either::Left(m.iter_external()),
        PointMappingsRefEnum::Compressed(m) => Either::Right(m.iter_external()),
    }
}
```

If we decide to add more variants, it can be chained, e.g.:
`Either<Either<A, B>, C>`

generall:
Please add explicit description and motivation of what is changed in this PR and why

lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -232,10 +233,10 @@ impl NonAppendableSegmentEntry for ProxySegment {
...
 
     /// Not implemented for proxy
-    fn iter_points(&self) -> Box<dyn Iterator<Item = PointIdType> + '_> {
-        // iter_points is not available for Proxy implementation
+    fn get_points(&self) -> PointMappingsGuard<'_> {
+        // get_points is not available for Proxy implementation
         // Due to internal locks it is almost impossible to return iterator with proper owning, lifetimes, e.t.c.
-        unimplemented!("call to iter_points is not implemented for Proxy segment")
+        unimplemented!("call to get_points is not implemented for Proxy segment")
     }
```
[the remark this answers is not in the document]
timvisee: Because this panic existed before I don't have immediate concerns

lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -213,6 +183,90 @@ pub trait IdTracker: fmt::Debug {
     }
 }
 
+/// Enum holding a reference to point mappings from an ID tracker.
+///
+/// Provides iteration methods over external/internal IDs without requiring
+/// the `IdTracker` trait to return boxed iterators.
```
timvisee, line 188: While we don't return a boxed iterator from the trait, now we do still return it from `pub fn iter_external(self)` (and others) below

lib/segment/src/id_tracker/id_tracker_base.rs
monoid: I've tried to implement it.  There are cases where there's 3 variants returned from `PointMappings`, but then it goes down to `IdTrackerEnum` which has 4 cases, so there's deeply nested `Either`. I wonder if it is still faster than dyn dispatch.

coszio: I'd say enum dispatch would still be preferred because it compiles with better optimizations, even for like inlining `next()` calls. If we keep using `impl Iterator` as return type, the nesting should not be a problem for ergonomics.

There's also [`auto_enums`](https://crates.io/crates/auto_enums) crate which makes the +2 variants case more ergonomic to write

lib/segment/src/segment/entry.rs
```diff
@@ -266,12 +266,8 @@ impl NonAppendableSegmentEntry for Segment {
         Ok(records)
     }
 
-    fn iter_points(&self) -> Box<dyn Iterator<Item = PointIdType>> {
-        // Sorry for that, but I didn't find any way easier.
-        // If you try simply return iterator - it won't work because AtomicRef should exist
-        // If you try to make callback instead - you won't be able to create <dyn SegmentEntry>
-        // Attempt to create return borrowed value along with iterator failed because of insane lifetimes
-        unsafe { self.id_tracker.as_ptr().as_ref().unwrap().iter_external() }
+    fn get_points(&self) -> PointMappingsGuard<'_> {
```
coszio: Probably best to refer to mappings, rather than points themselves
```suggestion
    fn get_point_mappings(&self) -> PointMappingsGuard<'_> {
```

lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -257,6 +183,138 @@ pub trait IdTracker: fmt::Debug {
...
+        owner: AtomicRef<'a, IdTrackerEnum>,
+
+        #[covariant]
+        dependent: PointMappingsRefEnum,
+    }
+}
+impl<'a> PointMappingsGuard<'a> {
+    pub fn iter(&self) -> Box<dyn Iterator<Item = PointIdType> + '_> {
+        self.borrow_dependent().iter_external()
+    }
```
coszio: If we ever need to add `iter_internal`, we don't want to rename this one
```suggestion
    pub fn iter_external(&self) -> Box<dyn Iterator<Item = PointIdType> + '_> {
        self.borrow_dependent().iter_external()
    }
```

lib/segment/src/id_tracker/id_tracker_base.rs
monoid: Well, it can be simplified if implementations refer to point_mappings directly. While there are several implementations of `IdTracker`, they all use only two point mappings implementations.  But I suggest doing it in another PR.

lib/segment/src/id_tracker/id_tracker_base.rs
monoid: I think it is not very likely because internal IDs is an implementation detail.

Actually, with nested self_cell I can try to leave `iter_points()` as it was, a `SegmentEntry` method.

lib/segment/src/segment/entry.rs
monoid: I've just returned the original `iter_points()`.

PR #8262, opened 2026-03-02
Interior mutability payload indices

monoid:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/hnsw_index/hnsw.rs
+++ b/lib/segment/src/index/hnsw_index/hnsw.rs
@@ -90,7 +90,7 @@ pub struct HNSWIndex {
     id_tracker: Arc<AtomicRefCell<IdTrackerEnum>>,
     vector_storage: Arc<AtomicRefCell<VectorStorageEnum>>,
     quantized_vectors: Arc<AtomicRefCell<Option<QuantizedVectors>>>,
-    payload_index: Arc<AtomicRefCell<StructPayloadIndex>>,
+    payload_index: Arc<parking_lot::RwLock<StructPayloadIndex>>,
     config: HnswGraphConfig,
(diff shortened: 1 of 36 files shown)


monoid:
Converted to draft for a while.

lib/segment/src/index/hnsw_index/hnsw.rs
```diff
@@ -90,7 +90,7 @@ pub struct HNSWIndex {
     id_tracker: Arc<AtomicRefCell<IdTrackerEnum>>,
     vector_storage: Arc<AtomicRefCell<VectorStorageEnum>>,
     quantized_vectors: Arc<AtomicRefCell<Option<QuantizedVectors>>>,
-    payload_index: Arc<AtomicRefCell<StructPayloadIndex>>,
+    payload_index: Arc<parking_lot::RwLock<StructPayloadIndex>>,
```
timvisee, line 93: Do we now use an `RwLock` here to ensure we properly serialize read/writes? Because we now don't have `&mut self` anymore on the segment level to guarantee this for us? And we don't want to panic or error by any other means here because of it.

lib/segment/src/segment/entry.rs
```diff
@@ -41,7 +41,7 @@ use crate::vector_storage::VectorStorage;
 /// any kind of proxy or wrapping.
 impl NonAppendableSegmentEntry for Segment {
     fn version(&self) -> SeqNumberType {
-        self.version.unwrap_or(0)
+        self.version.lock().unwrap_or(0)
```
timvisee, line 44: Does it make sense to change this version to an atomic? Or is it difficult because of the `None` variant?

PR #8263, opened 2026-03-02
Simpler `ReadMulti`/`WriteMulti` interface

ffuugoo:
### All Submissions:

* [ ] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --workspace --all-features` command?

### Changes to Core Features:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #8274, opened 2026-03-02
Reduce traits: MultiUniversalRead -> StorageRead

generall:
Alternative to https://github.com/qdrant/qdrant/pull/8263

Idea:

- Avoid using multiple traits, but instead of generalizing single reader, implement `SingleFile` adapter

Renames

- `MultiUniversalRead` -> `StorageRead`
- `MultiUniversalWrite` -> `StorageWrite`
- `UniversalRead` / `UniversalWrite` -> removed, replaced by `SingleFile<S>`


generall:
too much code

PR #8279, opened 2026-03-03
Adjust points selection for deferred points update

agourlay:
This PR handles deferred points in `find_points_to_update_and_delete`.

Previously, the deduplication logic did not distinguish between deferred and non-deferred points which could cause possible data loss/inconsistencies later.

The new rules for handling a given point are:

 1. Always update the latest version regardless of deferred status
 2. Never delete deferred points — optimizer handles them
 3. Delete older non-deferred copies only if the latest version is not deferred
 4. Keep older non-deferred copies when the latest is deferred 

## Changes

  - Move `point_is_deferred` to `NonAppendableSegmentEntry` trait and implement for `ProxySegment`
  - Refactor `find_points_to_update_and_delete` with new aforementioned rules.

## Tests

Mostly AI generated.

  - `test_find_points_to_update_and_delete`: baseline without deferred points
  - `test_find_points_to_update_and_delete_with_deferred`: deferred loses to newer non-deferred
  - `test_find_points_to_update_and_delete_with_deferred_only`: deferred-only point with no non-deferred copy
  - `test_find_points_to_update_and_delete_with_deferred_winning`: deferred wins when strictly newer than non-deferred

## Future work

Handle copy on write on update operation (https://github.com/qdrant/qdrant/pull/8290)


lib/shard/src/segment_holder/mod.rs
```diff
@@ -363,63 +351,113 @@ impl SegmentHolder {
...
-                                    .push(segment_point);
+                    Entry::Occupied(mut entry) => {
+                        let (existing_version, ref mut existing_segments) = *entry.get_mut();
+                        match point_version.cmp(&existing_version) {
+                            std::cmp::Ordering::Greater => {
+                                entry.insert((point_version, smallvec![segment_id]));
+                            }
+                            std::cmp::Ordering::Equal => {
+                                existing_segments.push(segment_id);
+                            }
```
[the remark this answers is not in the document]
agourlay: it compiles

lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -702,6 +702,15 @@ impl NonAppendableSegmentEntry for ProxySegment {
...
             .fill_query_context(query_context)
     }
+
+    fn point_is_deferred(&self, point_id: PointIdType) -> bool {
+        !self.deleted_points.contains_key(&point_id)
+            && self
+                .wrapped_segment
+                .get()
+                .read()
+                .point_is_deferred(point_id)
```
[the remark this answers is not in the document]
agourlay: good catch :+1:

lib/shard/src/segment_holder/mod.rs
```diff
@@ -363,63 +351,113 @@ impl SegmentHolder {
...
-            let segment_points = Self::segment_points(ids, segment_lock.deref());
-            for segment_point in segment_points {
-                let Some(point_version) = segment_lock.point_version(segment_point) else {
+            for &point_id in ids {
+                // point_version returns Some for both normal and deferred points,
+                // but None for soft-deleted or non-existent points
+                let Some(point_version) = segment_lock.point_version(point_id) else {
                     continue;
                 };
+                let is_deferred = segment_lock.point_is_deferred(point_id);
```
IvanPleshkov: It seems we check the deferred status of the point before actual insertion.
Because the deferred point id is updated after insertion, here we have false negative results.
As a fix proposal, we can upgrade `point_is_deferred` implementation, and in case of `deferred_point_id = None` check the threshold and reaching vector storage total size to reach this threshold - the knowledge of the point status before upsertion.
This thread is a reason to requests change in the PR status

lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -702,6 +702,23 @@ impl NonAppendableSegmentEntry for ProxySegment {
...
+        // Mirror point_version: only consider deletion effective when
+        // wrapped_version <= delete.local_version
+        if let Some(delete) = self.deleted_points.get(&point_id) {
+            let dominated = wrapped
+                .point_version(point_id)
+                .is_none_or(|v| v <= delete.local_version);
+            if dominated {
+                return false;
+            }
+        }
```
timvisee: IIRC `delete.local_version` can never be lower than the wrapped point version, making this extra check obsolete.

lib/shard/src/segment_holder/mod.rs
agourlay: Can you come up with a scenario where there would be false negative results?

This function is responsible for picking which points to update and which points to delete.
It is indeed being executed **before** applying the operation.

If the operation is a new insert. There are no points to update and no points to delete.
Therefore the current implementation is fine.

For the points being moved because of CoW they will be handed afterwards in https://github.com/qdrant/qdrant/pull/8290

If the operation is an update it gets more interesting.
The deferred information regarding existing points is valid when the function starts.

Can a point from a batch become deferred mid-batch later while applying the operation?
I do not think it is possible given that the amount of storage is reserved for dense named vector.

Outside of multivectors & sparse vectors which are dynamically size.
Is this what you are thinking about?

lib/shard/src/proxy_segment/segment_entry.rs
agourlay: We perform the same check in the `point_version` for a proxy segment.

https://github.com/qdrant/qdrant/blob/master/lib/shard/src/proxy_segment/segment_entry.rs#L42-L58

Should I just rely on `self.deleted_points.contains_key(&point_id) `?

lib/shard/src/segment_holder/mod.rs
```diff
@@ -363,63 +351,113 @@ impl SegmentHolder {
...
+        // latest_non_deferred: point_id -> (version, segment_ids)
+        // latest_deferred: point_id -> (version, segment_ids)
+        let mut latest_non_deferred: AHashMap<
+            PointIdType,
+            (SeqNumberType, SmallVec<[SegmentId; 1]>),
+        > = AHashMap::with_capacity(ids.len());
+        let mut latest_deferred: AHashMap<PointIdType, (SeqNumberType, SmallVec<[SegmentId; 1]>)> =
+            AHashMap::new();
+        // Track all occurrences so we can mark non-winners for deletion
+        let mut all_occurrences: Vec<(PointIdType, SeqNumberType, bool, SegmentId)> = Vec::new();
```
timvisee: Why like this, and not a deletion list right away like we did before?

I believe we only have to collect (older) visible points, like we did before. We should not delete deferred points.

lib/shard/src/segment_holder/mod.rs
```diff
@@ -363,63 +351,113 @@ impl SegmentHolder {
...
         for (segment_id, segment) in self.iter() {
             let segment_arc = segment.get();
             let segment_lock = segment_arc.read();
-            let segment_points = Self::segment_points(ids, segment_lock.deref());
-            for segment_point in segment_points {
-                let Some(point_version) = segment_lock.point_version(segment_point) else {
+            for &point_id in ids {
+                // point_version returns Some for both normal and deferred points,
+                // but None for soft-deleted or non-existent points
+                let Some(point_version) = segment_lock.point_version(point_id) else {
```
timvisee: I'm not sure if `point_version` versus `has_point` is a sound replacement. With had inconsistency problems between them before, where `has_point` was more reliable.

lib/shard/src/segment_holder/mod.rs
```diff
@@ -363,63 +351,113 @@ impl SegmentHolder {
...
-                            for old_segment_id in old_segment_ids {
-                                to_delete
-                                    .entry(old_segment_id)
-                                    .or_default()
-                                    .push(segment_point);
+                    Entry::Occupied(mut entry) => {
+                        let (existing_version, ref mut existing_segments) = *entry.get_mut();
+                        match point_version.cmp(&existing_version) {
+                            std::cmp::Ordering::Greater => {
+                                entry.insert((point_version, smallvec![segment_id]));
```
timvisee: Can we update the existing value? It'd get rid of an allocation+deallocation.

lib/shard/src/segment_holder/mod.rs
```diff
@@ -363,63 +351,113 @@ impl SegmentHolder {
...
-        // Group points to update by segments
-        let segment_count = self.len();
-        let mut to_update = AHashMap::with_capacity(min(segment_count, latest_points.len()));
-        let default_capacity = ids.len() / max(segment_count / 2, 1);
-        for (point_id, (_point_version, segment_ids)) in latest_points {
-            for segment_id in segment_ids {
+        // Pass 2: Decide what to update and delete
+        //
+        // Rules:
+        // - Latest non-deferred → always to_update
```
timvisee: Is this correct!?

I figured we cannot update the latest non-deferred if we have any deferred point in place. Otherwise, why would we have the deferred point?

And, if we only update the latest or latest deferred point we can greatly simplify the above and make it more like we had before. We don't consider the deferred state and just update the latest version we know about. Then we can keep a single map of points to update.

lib/shard/src/segment_holder/mod.rs
```diff
@@ -363,63 +351,113 @@ impl SegmentHolder {
...
+        // - Latest deferred → to_update only if its version > latest non-deferred version
```
timvisee: Shouldn't this be impossible?

If we promote a deferred point through the optimizer, we must also drop all older point versions along with it. That way we prevent having deferred points older than what is currently visible.

lib/shard/src/segment_holder/mod.rs
```diff
@@ -363,63 +351,113 @@ impl SegmentHolder {
...
+        // - All other copies → to_delete
+        let segment_count = self.len().max(1);
+        let default_vec_capacity = ids.len() / segment_count;
+        let mut to_update: AHashMap<SegmentId, Vec<PointIdType>> =
+            AHashMap::with_capacity(segment_count);
+        let mut to_delete: AHashMap<SegmentId, Vec<PointIdType>> =
+            AHashMap::with_capacity(segment_count);
+
+        for (point_id, point_version, is_deferred, segment_id) in all_occurrences {
```
timvisee: This extra loop over all occurrences makes resolving point placement more expensive. I wonder if it's needed.

lib/shard/src/segment_holder/mod.rs
agourlay: Alright, I can reinstate a pre-filter with `has_point` 👍

timvisee: If we validate both are consistent and maybe add a test for it I'd be fine with keeping this.

lib/shard/src/segment_holder/mod.rs
agourlay: The multivector case will be resolved later ✔️

agourlay:
I believe I applied all comments, including:
- not deleting old deferred version
- performing a single pass
- staying close to the original implementation

lib/shard/src/segment_holder/mod.rs
```diff
@@ -376,50 +391,74 @@ impl SegmentHolder {
...
+                            std::cmp::Ordering::Less => {
+                                // Older non-deferred: only delete if latest has a
+                                // non-deferred copy (don't discard the last
+                                // non-deferred copy in favor of a deferred latest)
+                                if !is_deferred && *latest_has_non_deferred {
+                                    to_delete
+                                        .entry(segment_id)
+                                        .or_insert_with(|| Vec::with_capacity(default_vec_capacity))
+                                        .push(segment_point);
+                                }
```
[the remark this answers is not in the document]
agourlay: Come on I just changed to a single pass solution.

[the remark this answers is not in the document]
agourlay: check this https://github.com/qdrant/qdrant/pull/8279/commits/d34db754f02ed68c0bcdb5c7f5c266f2d2c5a695

JojiiOfficial (approved):

lib/shard/src/proxy_segment/segment_entry.rs
timvisee: I see. Since it's already there let's keep it.

I'll experiment with some debug assertion locally if we can ever trigger the branch.

PR #8283, opened 2026-03-03
Filter deferred points in read operations Part 1 (Search, Scroll, Count)

JojiiOfficial:
Implements deferred point filtering for the following read operations:
- Search
- Count
- Scroll

It also partially implements filtering in other read operations, such as facet, since some operations sometimes use the same functions. Therefore I suggest to wait with merging until the remaining read operations are implemented, which will be done in a separate PR to not make this PR too big. 

This PR is best reviewed on per-commit level. However, the unit tests might be best to review at the end.

### DIFF
--- a/lib/segment/src/segment/tests.rs
+++ b/lib/segment/src/segment/tests.rs
@@ -920,9 +920,9 @@ fn test_dense_deferred_points() {
 fn test_dense_deferred_point_segment_combinations() {
     init_logger();
 
-    for dim in (2..40).step_by(3) {
-        for n_deferred in 0..15 {
-            for n_vectors in 1..15 {
+    for dim in [2, 17, 50] {
+        for n_deferred in [0, 1, 5, 14] {
+            for n_vectors in [1, 5, 14] {
                 let dir = Builder::new().prefix("segment_dir").tempdir().unwrap();
                 create_deferred_segment(&dir, dim, n_vectors, n_deferred);
             }


agourlay:
What about adding helper methods on the `IdTracker` instead of applying the same filters around the codebase?

Something like:

```rust
  trait IdTracker {

      fn iter_internal_visible(&self, deferred_from: Option<PointOffsetType>)
          -> Box<dyn Iterator<Item = PointOffsetType> + '_>
      {
          match deferred_from {
              None => self.iter_internal(),
              Some(d) => Box::new(self.iter_internal().take_while(move |&id| id < d)),
          }
      }

      fn iter_from_visible(&self, external_id: Option<PointIdType>, deferred_from: Option<PointOffsetType>)
          -> Box<dyn Iterator<Item = (PointIdType, PointOffsetType)> + '_>
      {
          match deferred_from {
              None => self.iter_from(external_id),
              Some(d) => Box::new(self.iter_from(external_id).filter(move |(_, iid)| *iid < d)),
          }
      }
  }
  ```

File: lib/segment/src/index/sparse_index/sparse_vector_index.rs [hunk 1]
```diff
@@ -353,7 +353,8 @@ impl<TInvertedIndex: InvertedIndex> SparseVectorIndex<TInvertedIndex> {
         let ids = match prefiltered_points {
             Some(filtered_points) => filtered_points.iter(),
             None => {
-                let filtered_points = payload_index.query_points(filter, &hw_counter, &is_stopped);
+                let filtered_points =
+                    payload_index.query_points(filter, &hw_counter, &is_stopped, None);
```
agourlay, line 357: why are we not applying the deferred_id filter here?

the sparse vectors are always indexed even on an appendable segment

JojiiOfficial: ~If I remember correctly, we decided to not filter sparse vectors since they are always indexed and there is no need to hide them.~

I'll tackle sparse vectors in a separate PR 👍🏻

lib/segment/src/segment/segment_ops.rs
```diff
@@ -671,6 +694,18 @@ impl Segment {
         false
     }
 
+    /// Returns `true` if the given point is a deferred point (and therefore must be excluded from read operations).
+    pub fn is_point_deferred_internal(&self, internal_id: PointOffsetType) -> bool {
+        Self::is_internal_id_deferred(internal_id, self.deferred_internal_id)
+    }
+
+    pub fn is_internal_id_deferred(
```
agourlay: I think it is time to rebase and use the new official function for this :)

lib/segment/src/segment/segment_ops.rs
```diff
@@ -657,6 +657,29 @@ impl Segment {
...
 
+    /// Returns the (estimated) amount of deferred points.
+    ///
+    /// This value is an estimation because it does not account for deferred points
+    /// that have been deleted before becoming visible.
+    pub fn deferred_point_count_estimated(&self) -> usize {
+        match self.deferred_internal_id {
+            Some(internal_id) => {
+                let id_tracker = self.id_tracker.borrow();
+                let max_id = id_tracker.total_point_count() as u32;
```
agourlay: why the `u32` casting?

```rust
 let total = id_tracker.total_point_count();
 total.saturating_sub(internal_id as usize)
```

lib/segment/src/segment/segment_ops.rs
JojiiOfficial: I don't see a new function for this. We take internal ID as parameter here, where the new functions from https://github.com/qdrant/qdrant/pull/8279 take the external ID.

lib/segment/src/segment/segment_ops.rs
JojiiOfficial: You're right, we can do it with less casting via `usize` too. My approach is more close to the actual types, since point IDs are u32 (or `PointOffsetType`). I don't have a strong opinion, so I'll change it to `usize` 👍🏻

lib/segment/src/index/sparse_index/sparse_vector_index.rs
JojiiOfficial: https://github.com/qdrant/qdrant/pull/8299

lib/segment/src/segment/segment_ops.rs
IvanPleshkov: We can remove a duplicate method later because this PR will be merged earlier than https://github.com/qdrant/qdrant/pull/8279

lib/segment/src/data_types/query_context.rs
```diff
@@ -141,7 +141,11 @@ impl<'a> SegmentQueryContext<'a> {
         self.query_context.available_point_count()
     }
 
-    pub fn get_vector_context(&self, vector_name: &VectorName) -> VectorQueryContext<'_> {
+    pub fn get_vector_context(
+        &self,
+        vector_name: &VectorName,
+        deferred_internal_id: Option<PointOffsetType>,
```
IvanPleshkov, line 147: I propose to remove ' deferred_internal_id ' here and move it to the higher level as a new `VectorQueryContext::deferred_internal_id` field. The motivation of the suggestion is to do like another help params: deleted points, hardware_counter ect

lib/segment/src/id_tracker/id_tracker_base.rs
```diff
@@ -101,6 +101,50 @@ pub trait IdTracker: fmt::Debug {
     /// Excludes soft deleted points.
     fn iter_random(&self) -> Box<dyn Iterator<Item = (PointIdType, PointOffsetType)> + '_>;
 
+    /// Iterate over all internal IDs. Optionally filter all deferred points.
+    fn iter_internal_visible(
```
IvanPleshkov, line 105: What if we move this filtering inside `iter_internal`? We will store a deferred id also as a field in the ID tracker by this proposal.
It's a question to discuss. I see pros and cons about it

IvanPleshkov: Motivation of such proposal:
There are new methods such as `iter_random_visible` and `iter_from_visible`. It would be nice if we can skip creation of this methods and use existing ones

lib/segment/src/index/hnsw_index/point_scorer.rs
```diff
@@ -349,15 +350,22 @@ impl<'a> BatchFilteredSearcher<'a> {
     pub fn peek_top_all(
         self,
         is_stopped: &AtomicBool,
+        deferred_internal_id: Option<PointOffsetType>,
```
IvanPleshkov, line 352: Can we provide this argument during scorer construction? Currently, we have `peek_top_all` and `peek_top_iter` with different usage - one requires parameters, the second one requires external filtering. I find it important to keep consistency between these methods

lib/segment/src/segment/segment_ops.rs
```diff
@@ -657,6 +657,29 @@ impl Segment {
...
+            None => 0,
+        }
+    }
+
+    /// Returns the amount of points that are not deferred.
+    pub fn non_deferred_point_count_estimated(&self) -> usize {
+        self.id_tracker
+            .borrow()
+            .available_point_count()
+            .saturating_sub(self.deferred_point_count_estimated())
```
IvanPleshkov, line 676: Inside `deferred_point_count_estimated` you define deferred points amount as `internal_id` value. Why do we perform more complicated estimation here instead of just taking `internal_id`?

lib/segment/src/id_tracker/id_tracker_base.rs
JojiiOfficial: To be honest I don't think that adding this field to IdTracker makes it much cleaner. Also in my opinion, the caller should decide when to filter and when not, to prevent filtering at places where we must not.

What exactly is the issue with adding those helper functions?

lib/segment/src/index/hnsw_index/point_scorer.rs
JojiiOfficial: I agree with your concerns about consistency. We already provide `deferred_internal_id` in the constructor of `BatchFilteredSearcher` but we can't apply it in `peek_top_iter` as the passed iterator can be of any order.

As an alternative approach, I'd suggest to pass it to both functions instead and remove it from the constructor/struct itself. Let me know which approach you prefer!

lib/segment/src/id_tracker/id_tracker_base.rs
IvanPleshkov: My concern is that we don't use `iter_random` anymore, a lot of glue and public api of ID tracker, which is unused.

My proposal about keeping deferred_point_id inside ID tracker is about keeping the public API of ID tracker unchanged (except for get-set methods). But it's a proposal, maybe it's enough just provide deferred id as an argument to old functions

lib/segment/src/index/hnsw_index/point_scorer.rs
IvanPleshkov: I preferred to have deferred id inside the scorer.
My arguments about such an approach.
Deferred ID is a special kind of filter. We keep all the other filters inside the scorer (like the payload filter, point deleted, and vector deleted filter).
Maybe it's enough to use it in BatchFilteredSearcher only and modify the filter from BatchFilteredSearcher into this call

IvanPleshkov: Let me ask if we can keep deferred id in BatchFilteredSearcher and modify the filter inside BatchFilteredSearcher to filter deferred points if they are presented

agourlay (approved):

agourlay:
please rebase :)

lib/segment/src/segment/tests.rs
```diff
@@ -843,7 +911,217 @@ fn test_dense_deferred_points() {
...
     assert_eq!(
         segment.deferred_internal_id,
-        Some(DEFERRED_POINTS_ID),
+        Some(13),
         "Deferred internal ID should still be `DEFERRED_POINTS_ID` after reopening"
     );
 }
+
+#[test]
+fn test_dense_deferred_point_segment_combinations() {
```
agourlay, line 920: this test is not playing well with the Windows CI :s

https://github.com/qdrant/qdrant/actions/runs/22906263470/job/66466398406?pr=8283#step:10:1538

```
TRY 1 SLOW [>420.000s] (─────────) segment segment::tests::test_dense_deferred_point_segment_combinations
```

JojiiOfficial: https://github.com/qdrant/qdrant/pull/8283/commits/ea8415a14bf280690ec8c26c2a69e075a6ae273e

PR #8284, opened 2026-03-03
Remove `shard` dependency on `api`

xzfc:
# Problem statement

Right now, `qdrant-edge` requires `protoc` to be installed to build it, which doesn't make sense.

# FAQ

- **Q:** Why it `qdrant-edge` needs `protoc`?
  **A:** Because `prost-wkt-types` (a dependency of `api`) requires it.

- **Q:** Why `qdrant-edge` depends on `api`?
  **A:** It shouldn't, but right now the dependency path is `qdrant-edge` → `shard` → `api`.

- **Q:** Why `qdrant-client` doesn't need `protoc` but `qdrant`/`api` does?
  **A:** Because `qdrant-client` [uses `::prost_types::Timestamp`](https://github.com/qdrant/rust-client/blob/a507f30f5ce734d056457402454e7bf6d7039ac2/src/qdrant.rs#L290), and `qdrant`/`api` [uses `::prost_wkt_types::Timestamp`](https://github.com/qdrant/qdrant/blob/09f87abbd8215031609988f3b0e220b739db2d9a/lib/api/src/grpc/qdrant.rs#L350).
  And `prost_wkt_types` (unlike `prost_types`) requires `protoc`.

- **Q:** Does `prost_wkt_types` really needs `protoc`?
  **A:** I didn't investigated yet, but I guess it's an implementation detail that could be changed if we want to. Put probably we don't.

- **Q:** Why we even need `prost_wkt_types` instead of `prost_types`?
  **A:** My fault! See https://github.com/qdrant/qdrant/pull/3395#discussion_r1466615680
  Particularly, because `prost_wkt_types` (unlike `prost_types`) is compatible with `serde` and `validator`. We don't need that in `qdrant-client`, so `prost_types` there is fine.
  
- **Q:** Why don't we just put a `#[cfg_attr()]` in generated—
  **A**: Enough.

# Approach

We don't need `qdrant-edge` to depend on `api` at all.
So, this PR makes `api` to be an optional dependency of `shard` and hides `api`-related items under `#[cfg(feature = "api")]`.
These items are mostly type conversions (`impl From<…> for …`).
Some `shard` structures are also hidden, but these are not seems to be used in `qdrant-edge` or `qdrant-edge-py`.

But there are still a few non-trivial/controversial places remain:
1. `ScrollRequestInternal::order_by` is `api::rest::OrderByInterface`.
2. `PointIdsList::shard_key` is `api::rest::ShardKeySelector`.

# Quick Validation

With `api`:
```
cargo check [--all-targets]
cargo check -p shard [--all-targets]
```

Without `api`:
```
cargo check -p shard --no-default-features [--all-targets]
cargo check -p edge [--all-targets]
cargo check -p qdrant-edge-py [--all-targets]
```

### DIFF
--- a/.github/workflows/edge-py-release.yml
+++ b/.github/workflows/edge-py-release.yml
@@ -39,9 +39,6 @@ jobs:
           - target: aarch64
             manylinux: manylinux_2_17
 
-    env:
-      PROTOC_VERSION: '33.1'
-
     steps:
       - name: Checkout Qdrant
         uses: actions/checkout@v6
@@ -54,21 +51,6 @@ jobs:
           working-directory: lib/edge/python
           target: ${{ matrix.target }}
           manylinux: ${{ matrix.manylinux }}
-          before-script-linux: |
-            if command -v unzip; then
-                :
-            elif command -v apt-get; then
-                apt-get update -y && apt-get install -y unzip
-            elif command -v yum; then
-                yum install -y unzip
-            elif command -v dnf; then
-                dnf install -y unzip
-            elif command -v apk; then
-                apk add --no-cache unzip
-            fi
-
-            curl -f -L https://github.com/protocolbuffers/protobuf/releases/download/v${{ env.PROTOC_VERSION }}/protoc-${{ env.PROTOC_VERSION }}-linux-x86_64.zip -o protoc.zip
-            unzip protoc.zip -d /usr
 
       - name: Upload Python wheel
         uses: actions/upload-artifact@v7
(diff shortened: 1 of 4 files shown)


lib/shard/src/scroll.rs
xzfc: Controversial change `#1`: `ScrollRequestInternal::order_by` is `api::rest::OrderByInterface`, so I've reimplemented `OrderByInterface` in `shard`.

lib/shard/src/operations/point_ops.rs
```diff
@@ -47,14 +42,16 @@ pub enum UpdateMode {
 #[serde(rename_all = "snake_case")]
 pub struct PointIdsList {
     pub points: Vec<PointIdType>,
+    #[cfg(feature = "api")]
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub shard_key: Option<ShardKeySelector>,
+    pub shard_key: Option<api::rest::ShardKeySelector>,
 }
```
xzfc, line 48: Controversial change `#2`:  `PointIdsList::shard_key` is `api::rest::ShardKeySelector`, so I put this field under `#[cfg(feature = "api")]`.

PR #8286, opened 2026-03-03
universal file ops

generall:
Gridstore not only reads from files, but also requires some extra File-system level operations.
We want to make those opeations independent of actual IO layer (aka should work for s3 as well)

### DIFF
--- a/lib/common/common/src/universal_io/file_ops.rs
+++ b/lib/common/common/src/universal_io/file_ops.rs
@@ -1,0 +1,13 @@
+use std::path::{Path, PathBuf};
+
+pub trait UniversalReadFileOps {
+    /// List files in the storage with the given prefix.
+    /// The prefix is used to filter files, e.g. by directory or filename pattern.
+    ///
+    /// Example: `./gridstore/page_`
+    /// should return
+    /// - `./gridstore/page_1.dat`
+    /// - `./gridstore/page_2.dat`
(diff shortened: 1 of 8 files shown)


lib/common/common/src/universal_io/file_ops.rs
```diff
@@ -0,0 +1,13 @@
...
+pub trait UniversalReadFileOps {
+    /// List files in the storage with the given prefix.
+    /// The prefix is used to filter files, e.g. by directory or filename pattern.
+    ///
+    /// Example: `./gridstore/page_`
+    /// should return
+    /// - `./gridstore/page_1.dat`
+    /// - `./gridstore/page_2.dat`
+    /// - `./gridstore/page_3.dat`
+    fn list_files(prefix_path: &Path) -> crate::universal_io::Result<Vec<PathBuf>>;
```
coszio, line 12: In [`fusio`](https://docs.rs/fusio/latest/fusio/path/index.html) crate, they have a custom `Path` implementation that it is friendlier with url-encoded stuff. Do we want something like that in the future?

coszio (approved):

lib/common/common/src/universal_io/file_ops.rs
generall: I don't think we will have full urls in here, the idea is that from "user" point of view is stays regular filesystem

PR #8292, opened 2026-03-04
`IoUringFile` based on universal I/O interface

ffuugoo:
Initial implementation of `IoUringFile` for universal I/O.

Few caveats:
- implements `UniversalRead<u8>`/`UniversalWrite<u8>` instead of `UR<T>`/`UW<T>`
  - internally, it's always `u8`, so I started with this for simplicity, I'll tweak it into `T` in a follow-up PR
- also no `read_multi`/`write_multi`, yet
  - follow-up PR, it's a small refactoring

**TODO:**
- [ ] ~~handle partial reads/writes 😭~~
	- _#8300_
- [x] cleaner error handling 🫳
- [ ] tests 🧪

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/common/common/src/universal_io/io_uring.rs
+++ b/lib/common/common/src/universal_io/io_uring.rs
@@ -196,7 +196,7 @@ impl<'a> IoUringRuntime<'a> {
             self.io_uring
                 .submission()
                 .push(&entry)
-                .map_err(|err| io::Error::other(err))?;
+                .map_err(io::Error::other)?;
         }
 
         Ok(())


lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -0,0 +1,244 @@
...
+                            .read(id as _, self.fd(), range)
+                            .expect("ID is unique");
+
+                        if let Err(_) = unsafe { sqe.push(&entry) } {
+                            break;
+                        }
+
+                        let _ = ranges.next();
+                    }
+                }
```
ffuugoo: I'm open to suggestions if you have bright ideas how to make this part less ugly.

Note that `push` is fallible, so ideally we need to
- push as many as we can
- if `push` fails, keep `entry` until next iteration of outer cycle
- we can't break up `IoUringRuntime`, because we need `Drop`

ffuugoo: Nevermind.

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -0,0 +1,317 @@
...
+
+            for result in self.completed() {
+                match result {
+                    Ok(_) => (),
+                    Err(err) => log::debug!("{err}"),
+                }
+            }
+        }
+    }
+}
```
ffuugoo, line 270: So, the interesting part, that is different from our existing `IoUringReader`, is this.

We want to ensure, that if we, e.g., throw an error in the middle of `IoUringFile::read_multi`, we complete or cancel all pending operations. Because if we don't, and we just propagate the error and return from `read_multi`, then we would deallocate our `BufferStore`, but kernel will still hold pointers into these buffers and write into deallocated memory.

And it does not matter that much, if we put `BufferStore` inside TLS next to `IoUring`, because we would have the same problem: if we exit the thread, TLS would be dropped, but kernel might still have pointers into deallocated `BufferStore`. 🤷‍♀️

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -0,0 +1,318 @@
...
+                    .offset(offset)
+                    .build()
+                    .user_data(0);
+
+                Some(entry)
+            });
+
+            rt.submit_and_wait(1).expect("write operation submitted");
+            rt.completed().next().expect("write operation completed")?;
+            Ok(())
```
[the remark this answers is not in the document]
ffuugoo: Good point. Good bot.

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -0,0 +1,317 @@
...
+        local_file_ops::local_list_files(prefix_path)
+    }
+
+    fn exists(path: &Path) -> crate::universal_io::Result<bool> {
+        fs::exists(path).map_err(UniversalIoError::from)
+    }
+}
+
+impl UniversalRead<u8> for IoUringFile {
+    fn open(path: impl AsRef<Path>, _options: OpenOptions) -> Result<Self>
```
generall, line 42: Proposal for handling of unsupported IO_URING:

- on open, check that io_uring is either already created or actually create it. 
- if creation errors out, return error in the open result

Assume user will choose different storage backend if this happens

coszio: We need a custom error for this as well

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -0,0 +1,317 @@
...
+            in_progress: 0,
+        }
+    }
+
+    pub fn enqueue(&mut self, mut entries: impl FnMut(&mut BufferStore) -> Option<squeue::Entry>) {
+        let mut sqe = self.io_uring.submission();
+
+        if sqe.is_full() {
+            return;
+        }
```
coszio: If I am reading this right, this means it just won't submit anything if submission queue is full?

I think this function should return number of submitted entries, or something that helps `read`/`write` methods retry or wait for the appropriate number of entries

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -0,0 +1,317 @@
...
+    pub fn read(&mut self, id: u64, fd: Fd, range: ElementsRange) -> Option<squeue::Entry> {
+        let ElementsRange {
+            start: offset,
+            length,
+        } = range;
+
+        let length = u32::try_from(length).expect("range length is within u32");
+
+        let buffer = self.create(id, length as _)?;
+        let entry = opcode::Read::new(fd, buffer.as_mut_ptr(), length as _)
```
coszio: nit: `length` is already `u32`

lib/common/common/src/universal_io/io_uring.rs
ffuugoo: kinda. there's `submit_and_wait` method, and when you call it, kernel takes entries out of the queue.

so the workflow is 1) push into queue 2) submit 3) check completed ops 4) repeat

lib/common/common/src/universal_io/io_uring.rs
ffuugoo: yep yep, I'll do a small update shortly with a fix

lib/common/common/src/universal_io/io_uring.rs
ffuugoo: Tweaked this just now, because I realized we should track how many requests we submit, to not overflow completion queue.

lib/common/common/src/universal_io/io_uring.rs
ffuugoo: Fixed.

lib/common/common/src/universal_io/io_uring.rs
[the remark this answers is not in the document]
ffuugoo: Handled in follow-up PR.

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -0,0 +1,317 @@
...
+                let entry = buffers.read(0, self.fd(), range).expect("ID is unique");
+                Some(entry)
+            });
+
+            rt.submit_and_wait(1).expect("read operation submitted");
+
+            let (_, buffer) = rt.completed().next().expect("read operation completed")?;
+            let buffer = buffer.expect("associated buffer returned");
+            Ok(Cow::from(buffer))
+        })
```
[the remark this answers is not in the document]
ffuugoo: Handled in follow-up PR.

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -0,0 +1,318 @@
...
+use std::sync::Arc;
+
+use ::io_uring::types::Fd;
+use ::io_uring::{IoUring, opcode, squeue};
+
+use super::*;
+
+thread_local! {
+    static IO_URING: RefCell<IoUring> = RefCell::new(IoUring::new(IO_URING_QUEUE_LENGTH).expect("io_uring initialized"));
+}
```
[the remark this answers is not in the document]
ffuugoo: Fixed.

PR #8293, opened 2026-03-05
Return error instead of panicking for corrupted aliases file on startup

### LINKED ISSUE #5004: Panic seen if aliases file is non-valid JSON
## Current Behavior
When the aliases file (default location `/qdrant/storage/aliases/data.json`) is non-valid JSON, Qdrant panics when it tries to parse it.

## Steps to Reproduce
1. Run Qdrant in Docker (`docker run -d --name qdrant --pull=always qdrant/qdrant`)
2. Simulate empty (or corrupted) aliases file (`/qdrant/storage/aliases/data.json`): `docker exec qdrant bash -c "> /qdrant/storage/aliases/data.json"`
3. Restart container: `docker restart qdrant`)

Logs:

```
           _                 _    
  __ _  __| |_ __ __ _ _ __ | |_  
 / _` |/ _` | '__/ _` | '_ \| __| 
| (_| | (_| | | | (_| | | | | |_  
 \__, |\__,_|_|  \__,_|_| |_|\__| 
    |_|                           

Version: 1.11.3, build: 9fa86106
Access web UI at http://localhost:6333/dashboard

2024-09-02T14:32:30.302323Z  INFO storage::content_manager::consensus::persistent: Loading raft state from ./storage/raft_state.json    
2024-09-02T14:32:30.307701Z ERROR qdrant::startup: Panic backtrace: 
   0: std::backtrace::Backtrace::create
   1: qdrant::startup::setup_panic_hook::{{closure}}
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::begin_panic_handler::{{closure}}
   4: std::sys_common::backtrace::__rust_end_short_backtrace
   5: rust_begin_unwind
   6: core::panicking::panic_fmt
   7: core::result::unwrap_failed
   8: storage::content_manager::toc::TableOfContent::new
   9: qdrant::main
  10: std::sys_common::backtrace::__rust_begin_short_backtrace
  11: main
  12: <unknown>
  13: __libc_start_main
  14: _start
    
2024-09-02T14:32:30.307720Z ERROR qdrant::startup: Panic occurred in file lib/storage/src/content_manager/toc/mod.rs at line 162: Can't open database by the provided config: ServiceError { description: "EOF while parsing a value at line 1 column 0", backtrace: Some("   0: std::backtrace::Backtrace::create\n   1: storage::content_manager::errors::StorageError::service_error\n   2: <storage::content_manager::errors::StorageError as core::convert::From<io::file_operations::Error>>::from\n   3: storage::content_manager::alias_mapping::AliasPersistence::open\n   4: storage::content_manager::toc::TableOfContent::new\n   5: qdrant::main\n   6: std::sys_common::backtrace::__rust_begin_short_backtrace\n   7: main\n   8: <unknown>\n   9: __libc_start_main\n  10: _start\n") }    

```

## Expected Behavior
Avoid a panic, either provide a good error and exit or log the error, delete the invalid file and start.

## Possible Solution
Depends on what the intended action for the user should be. At least avoid the panic.

I read that we only backup aliases when using full storage snapshot, which are only suitable for single node deployments. So if something goes wrong there, I don't see how we can exit and instruct the user to do something different? Possibly recreate the full storage snapshot?

If there are no concerns, I would propose to log as a warning, delete the file and start. But I might be lacking context around aliases.

## Context (Environment)
Yes this is a rare situation and manually triggered but still, avoiding the panic and creating a proper error seems realistic here. Feel free to close if not important/disagree.

leohenon:
Replace the startup panic/backtrace with a clean error message when the aliases file contains invalid JSON, by having `TableOfContent::new` return `Result` like `Persistent::load_or_init`.

Closes #5004

### DIFF
--- a/lib/storage/src/content_manager/toc/mod.rs
+++ b/lib/storage/src/content_manager/toc/mod.rs
@@ -46,7 +46,7 @@ use crate::content_manager::alias_mapping::AliasPersistence;
 use crate::content_manager::collection_meta_ops::CreateCollectionOperation;
 use crate::content_manager::collections_ops::{Checker, Collections};
 use crate::content_manager::consensus::operation_sender::OperationSender;
-use crate::content_manager::errors::StorageError;
+use crate::content_manager::errors::{StorageError, StorageResult};
 use crate::content_manager::shard_distribution::ShardDistributionProposal;
 use crate::content_manager::toc::telemetry::TocTelemetryCollector;
 use crate::rbac::{Access, AccessRequirements, CollectionMultipass, CollectionPass};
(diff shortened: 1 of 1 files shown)


xzfc:
Hi. Thanks for digging it.

I believe this should be solved on a different level. I.e., `common::fs::ops::read_json` (and other similar functions) should return affected file path in their errors; so all callers will be covered, not just this particular usage of `aliases.json`.

I've pushed an update to your branch implementing this.

leohenon:
Good call, much better to handle it at that level. Thanks for pushing the changes!

PR #8298, opened 2026-03-05
Refactor `PointToValues` to use `UniversalRead`

coszio:
Since the reads in new `universal_io` return `Cow`, we need to change how we deal with each set of values for a particular point. Previously, it was "deserialized" iteratively, but this PR changes it to be done at once

## What I tried
Naively, the change would be to just hold the `Cow<[u8]>` of the read bytes in an iterator, and hand out references to it, so that callers can decide what to do with the reference. 

```rust
struct ValuesIter<'a, T: MmapValue> {
    bytes: Cow<'a, [u8]>,
    count: usize,
}

impl<'a, T> Iterator for ValuesIter<'a, T> {
    type Item: &'a T;
    // ... rest of the code
}
```
But the trait definition of `Iterator` does not allow to define the `Item` as a reference to an owned section of the slice. It doesn't work because it would mean that in the case of owned bytes (`Vec<u8>`), would need to live as long as the mmap itself, which is not true.

The idea of handing out references to things owned by the iterator is not wrong though. It _can_ work, if only the `Iterator` trait could have GAT (`type Item<'a>`) such that the lifetime can be tied to the `next()` function, not to the iterator itself.

There are [crates][1] that solve this, but it is not super elegant and drifts from well known patterns. It would also make all upstream code aware of this pattern, so it wouldn't be self-contained.

## What this PR does

To solve this more ergonomically, I have:
- modified the `MmapValue` trait to convert the bytes of ALL the values of a particular point at once, and collect them into an associated type container.
- All zero-copyable types now use bytemuck to cast into a slice/vec, and share a common implementation.
- The odd case is `str`, which is implemented separately

The other changes are necessary propagation of these changes, which includes:
- more propagation of `OperationResult`
- `FacetValueRef` now takes copy types by value, and only `Keyword` variant is `Cow`
- ambiguity was introduced between `MapIndexKey::Owned` and `ToOwned::Owned`, so we need to specify `<N as MapIndexKey>::Owned` everywhere.
- derived `bytemuck::Pod` for `GeoPoint` and included `"bytemuck"` feature for `OrderedFloat`

[1]: https://docs.rs/crate/lender/latest

### DIFF
--- a/lib/segment/src/index/field_index/mmap_point_to_values.rs
+++ b/lib/segment/src/index/field_index/mmap_point_to_values.rs
@@ -215,11 +215,11 @@ where
         Ok(false)
     }
 
-    pub fn values_iter<'a>(
-        &'a self,
+    pub fn values_iter(
+        &self,
         point_id: PointOffsetType,
         hw_counter: ConditionedCounter, // TODO: make it by reference
-    ) -> OperationResult<Option<impl Iterator<Item = Cow<'a, T>>>> {
+    ) -> OperationResult<Option<impl Iterator<Item = Cow<'_, T>>>> {
         let hw_cell = hw_counter.payload_index_io_read_counter();
 
         // first, get range of values for point
--- a/lib/segment/src/data_types/facets.rs
+++ b/lib/segment/src/data_types/facets.rs
@@ -40,45 +40,45 @@ pub enum FacetValueRef<'a> {
 impl FacetValueRef<'_> {
     pub fn to_owned(&self) -> FacetValue {
         match self {
-            FacetValueRef::Keyword(s) => FacetValue::Keyword((*s).to_string()),
-            FacetValueRef::Int(i) => FacetValue::Int(*i),
-            FacetValueRef::Uuid(uuid) => FacetValue::Uuid(*uuid),
-            FacetValueRef::Bool(b) => FacetValue::Bool(*b),
+            FacetValueRef::Keyword(str) => FacetValue::Keyword(str.to_string()),
+            &FacetValueRef::Int(int) => FacetValue::Int(int),
+            &FacetValueRef::Uuid(uuid) => FacetValue::Uuid(uuid),
+            &FacetValueRef::Bool(bool) => FacetValue::Bool(bool),
         }
     }
 }
 
 impl<'a> From<Cow<'a, str>> for FacetValueRef<'a> {
-    fn from(s: Cow<'a, str>) -> Self {
-        FacetValueRef::Keyword(s)
+    fn from(str: Cow<'a, str>) -> Self {
+        FacetValueRef::Keyword(str)
     }
 }
 
 impl<'a> From<&'a str> for FacetValueRef<'a> {
-    fn from(s: &'a str) -> Self {
-        FacetValueRef::Keyword(Cow::Borrowed(s))
+    fn from(str: &'a str) -> Self {
+        FacetValueRef::Keyword(Cow::Borrowed(str))
     }
 }
 
-impl<'a> From<Cow<'a, IntPayloadType>> for FacetValueRef<'a> {
-    fn from(i: Cow<'a, IntPayloadType>) -> Self {
-        FacetValueRef::Int(*i)
+impl<'a> From<Cow<'_, IntPayloadType>> for FacetValueRef<'a> {
+    fn from(int: Cow<'_, IntPayloadType>) -> Self {
+        FacetValueRef::Int(int.into_owned())
     }
 }
 
-impl<'a> From<&'a IntPayloadType> for FacetValueRef<'a> {
-    fn from(i: &'a IntPayloadType) -> Self {
-        FacetValueRef::Int(*i)
+impl<'a> From<&IntPayloadType> for FacetValueRef<'a> {
+    fn from(int: &IntPayloadType) -> Self {
+        FacetValueRef::Int(*int)
     }
 }
-impl<'a> From<Cow<'a, UuidIntType>> for FacetValueRef<'a> {
-    fn from(uuid: Cow<'a, UuidIntType>) -> Self {
-        FacetValueRef::Uuid(*uuid)
+impl<'a> From<Cow<'_, UuidIntType>> for FacetValueRef<'a> {
+    fn from(uuid: Cow<'_, UuidIntType>) -> Self {
+        FacetValueRef::Uuid(uuid.into_owned())
     }
 }
 
-impl<'a> From<&'a UuidIntType> for FacetValueRef<'a> {
-    fn from(uuid: &'a UuidIntType) -> Self {
+impl<'a> From<&'_ UuidIntType> for FacetValueRef<'a> {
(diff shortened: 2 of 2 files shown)


lib/segment/src/index/field_index/mmap_point_to_values.rs
```diff
@@ -1,153 +1,94 @@
...
+                        .ok()?
+                        .to_owned();
+                    start += Self::mmapped_size(&string);
+                    holder.push(Cow::Owned(string));
+                }
+                holder
+            }
+        };
+        Some(cow)
     }
```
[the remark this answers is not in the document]
coszio: Ohp, good point!

lib/segment/src/index/field_index/numeric_index/mod.rs
```diff
@@ -286,7 +286,10 @@ where
         match self {
             NumericIndexInner::Mutable(index) => index.check_values_any(idx, check_fn),
             NumericIndexInner::Immutable(index) => index.check_values_any(idx, check_fn),
-            NumericIndexInner::Mmap(index) => index.check_values_any(idx, check_fn, hw_counter),
+            // FIXME: don't silently ignore error, change output of this function and propagate
+            NumericIndexInner::Mmap(index) => index
+                .check_values_any(idx, check_fn, hw_counter)
+                .unwrap_or(false),
```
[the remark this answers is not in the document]
coszio: Already acknowledged it :)

lib/segment/src/index/field_index/mmap_point_to_values.rs
```diff
@@ -1,153 +1,95 @@
...
-            .and_then(|bytes| value.lat.write_to_prefix(bytes).ok())
-    }
-
-    fn from_referenced<'a>(value: &'a Self::Referenced<'_>) -> &'a Self {
-        value
+    fn read_from_mmap(bytes: Cow<'_, [u8]>, _count: usize) -> Option<Self::CowValues<'_>> {
+        let cow = match bytes {
+            Cow::Borrowed(slice) => Cow::Borrowed(bytemuck::cast_slice(slice)),
+            Cow::Owned(vec) => Cow::Owned(bytemuck::cast_vec(vec)),
+        };
```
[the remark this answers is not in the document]
coszio: Thanks, good eye!

lib/segment/src/index/field_index/mmap_point_to_values.rs
```diff
@@ -253,26 +194,37 @@ impl<T: MmapValue + ?Sized> MmapPointToValues<T> {
...
         mmap.flush()?;
-        Ok(Self {
-            file_name,
-            mmap: mmap.make_read_only()?,
-            header,
-            phantom: std::marker::PhantomData,
-        })
+        drop(mmap);
+
+        Self::open(&file_name, true)
```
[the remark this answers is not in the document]
coszio: Thank you

lib/segment/src/index/field_index/mmap_point_to_values.rs
```diff
@@ -290,66 +242,62 @@ impl<T: MmapValue + ?Sized> MmapPointToValues<T> {
...
 
-        // finally, return iterator
-        Some(
-            std::iter::successors(read_value(range), move |range| read_value(range.1))
-                .map(|(value, _)| value),
-        )
+        let bytes = self.store.read::<false>(bytes_range)?;
+        let count = self.get_values_count(point_id)?.unwrap_or(0);
+
+        let cow_values = T::read_from_mmap(bytes, count)
```
generall: proposal for next PR: for hot path in `check_values_any` we can use callback, other uses seems to be less hot, so collect is fine

coszio:
Update: Thanks to a discussion with @ffuugoo, https://github.com/qdrant/qdrant/pull/8298/changes/cc1fdbfbe776c611a57cc649323dda6401d442d6 now makes it so that we don't need to pre-collect, and still hand out even `&str` references without reallocating

lib/segment/src/index/field_index/mmap_point_to_values.rs
```diff
@@ -290,66 +201,57 @@ impl<T: MmapValue + ?Sized> MmapPointToValues<T> {
...
+        Ok(false)
     }
 
-    pub fn get_values<'a>(
+    pub fn values_iter<'a>(
         &'a self,
         point_id: PointOffsetType,
-    ) -> Option<impl Iterator<Item = T::Referenced<'a>> + 'a> {
+        hw_counter: ConditionedCounter, // TODO: make it by reference
+    ) -> OperationResult<Option<impl Iterator<Item = Cow<'a, T>>>> {
```
ffuugoo: This could be:

```suggestion
    pub fn values_iter(
        &self,
        point_id: PointOffsetType,
        hw_counter: ConditionedCounter, // TODO: make it by reference
    ) -> OperationResult<Option<impl Iterator<Item = Cow<'_, T>>>> {
```

Not saying it has to, but it could. Personally, I don't like lifetimes, when they are not strictly required, and for me `'_` is _more obvious_ indicator of "it borrows `&self`", than explicit `'a`.

lib/segment/src/index/field_index/mmap_point_to_values.rs
```diff
@@ -290,66 +201,57 @@ impl<T: MmapValue + ?Sized> MmapPointToValues<T> {
...
+                start: range.start,
+                length: range.end - range.start,
+            };
+            // Measure IO overhead of `self.get_bytes_range()` and the length of the values
+            hw_cell.incr_delta(MMAP_PTV_ACCESS_OVERHEAD + range.length as usize);
+
+            range
+        }) else {
+            return Ok(None);
         };
```
ffuugoo: `value_iter` returns `Option<impl Iterator>`. Would it make sense to return _an empty iterator_ instead of `None`? Do we explicitly rely on this `Option` anywhere?

ffuugoo:
> ambiguity was introduced between MapIndexKey::Owned and ToOwned::Owned, so we need to specify <N as MapIndexKey>::Owned everywhere

Maybe rename `MapIndexKey::Owned` into `OwnedKey` and `MapIndexKey::to_owned` into `to_owned_key`?

lib/segment/src/index/field_index/mmap_point_to_values.rs
ffuugoo: Fixed [`d63fc51` (this PR)](https://github.com/qdrant/qdrant/pull/8298/commits/d63fc51954092a78607f22eee3585ffa798538d7)

coszio:
> Maybe rename MapIndexKey::Owned into OwnedKey and MapIndexKey::to_owned into to_owned_key?

I can handle renames in the following PR

lib/segment/src/index/field_index/mmap_point_to_values.rs
coszio: Well, it could always be an `ExactSizeIterator` but it makes handling `None` values easier with a `?`

PR #8299, opened 2026-03-05
Filter deferred points sparse search

JojiiOfficial:
Depends on #8283 

Implements efficient filtering of deferred points in sparse vector search.

### DIFF
--- a/lib/sparse/src/index/search_context.rs
+++ b/lib/sparse/src/index/search_context.rs
@@ -208,10 +208,10 @@ impl<'a, 'b, T: PostingListIter> SearchContext<'a, 'b, T> {
         // Select appropriate last-id by also considering deferred points.
         let max_point = self
             .deferred_internal_id
-            .map_or(PointOffsetType::MAX, |defered_id| {
+            .map_or(PointOffsetType::MAX, |deferred_id| {
                 // `for_each_till_id` expects the last ID to be included.
                 // Therefore we have to subtract 1.
-                defered_id.saturating_sub(1)
+                deferred_id.saturating_sub(1)
             });
 
         posting


lib/sparse/src/index/search_context.rs
```diff
@@ -197,29 +204,47 @@ impl<'a, 'b, T: PostingListIter> SearchContext<'a, 'b, T> {
...
         let posting = &mut self.postings_iterators[0];
-        posting.posting_list_iterator.for_each_till_id(
-            PointOffsetType::MAX,
-            &mut (),
-            |_, id, weight| {
+
+        // Select appropriate last-id by also considering deferred points.
+        let max_point = self
+            .deferred_internal_id
+            .map_or(PointOffsetType::MAX, |defered_id| {
```
agourlay: `deferred_id` :)

JojiiOfficial: 8bfe14a5d498c2305e5d8f6e123607f0435d26a4

lib/sparse/src/index/search_context.rs
```diff
@@ -197,29 +204,47 @@ impl<'a, 'b, T: PostingListIter> SearchContext<'a, 'b, T> {
...
-            &mut (),
-            |_, id, weight| {
+
+        // Select appropriate last-id by also considering deferred points.
+        let max_point = self
+            .deferred_internal_id
+            .map_or(PointOffsetType::MAX, |deferred_id| {
+                // `for_each_till_id` expects the last ID to be included.
+                // Therefore we have to subtract 1.
+                deferred_id.saturating_sub(1)
```
IvanPleshkov, line 214: If deferred id is equal to `Some(0)`. We're gonna include the first point in the search. I propose to return if the deferred id is zero. In practice, we don't expect such a situation, but we can do it after release for additional safety.

lib/sparse/src/index/search_context.rs
```diff
@@ -34,6 +34,7 @@ pub struct SearchContext<'a, 'b, T: PostingListIter = PostingListIterator<'a>> {
     pooled: PooledScoresHandle<'b>,         // handle to pooled scores
     use_pruning: bool,
     hardware_counter: &'a HardwareCounterCell,
+    deferred_internal_id: Option<PointOffsetType>,
```
IvanPleshkov, line 37: Maybe we can remove `SearchContext::deferred_internal_id` and patch `max_record_id` in the constructor. As a result, we may skip all other changes in `SearchContext`.
For now, it's not critical to fix, let's investigate this thread after release

JojiiOfficial: To me "max_id" always sounds as if this point itself is included. I couldn't find another term that empathizes the exclusion so I decided to name it 'deferred_internal_id'.

IvanPleshkov: We can discuss it after the release. As the variant to handle this case, search context creation returns `None` and the search fallbacks to return nothing

PR #8300, opened 2026-03-05
Improvements for `IoUringFile`

ffuugoo:
Follow-up to #8292.

- handle read and write requests more explicitly
- assert against partial reads/writes

__[TODO]:__
- [x] implement `read_multi` and `write_multi`
  - [ ] unify/deduplicate `*_batch` and `*_multi` (and maybe `read_*` and `write_*`) impls?
- [x] implement `populate` and `clear_ram_cache`
- [x] implement `UniversalRead`/`UniversalWrite` for `T`
- [x] handle errors without panic
- [x] check if `io_uring` available/supported during `IoUringFile::open`
- [ ] use `Vec`/`VecDeque` in `IoUringState` instead of `HashMap`

[TODO]: https://github.com/qdrant/qdrant/pull/8292#pullrequestreview-3896276596

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1450,6 +1450,7 @@ dependencies = [
 name = "common"
 version = "0.0.0"
 dependencies = [
+ "ahash",
  "atomicwrites",
  "bincode 1.3.3",
  "bitvec",
--- a/lib/common/common/Cargo.toml
+++ b/lib/common/common/Cargo.toml
@@ -16,6 +16,7 @@ workspace = true
 testing = []
 
 [dependencies]
+ahash = { workspace = true }
 atomicwrites = { workspace = true }
 bincode = { workspace = true }
 bitvec = { workspace = true }


lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -58,15 +58,14 @@ impl UniversalRead<u8> for IoUringFile {
 
     fn read<const SEQUENTIAL: bool>(&self, range: ElementsRange) -> Result<Cow<'_, [u8]>> {
         IO_URING.with_borrow_mut(|io_uring| {
-            let mut rt = IoUringRuntime::new(io_uring);
+            let mut rt = Runtime::new(io_uring);
```
generall: nit: I like full name better, it is easier navigation. Including for the agents

ffuugoo: Fixed.

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -56,118 +83,213 @@ impl UniversalRead<u8> for IoUringFile {
...
+        let items_len = byte_len / size_of::<T>() as u64;
+        Ok(items_len)
     }
 
     fn populate(&self) -> Result<()> {
+        let mut file = self.file.as_ref();
+        let mut buffer = vec![0u8; 1024 * 1024];
+
+        while file.read(&mut buffer)? > 0 {}
         Ok(())
```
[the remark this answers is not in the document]
ffuugoo: Good point.

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -38,11 +51,25 @@ impl UniversalReadFileOps for IoUringFile {
...
+
+            if probe.is_supported(opcode::Read::CODE) && probe.is_supported(opcode::Write::CODE) {
+                Ok(())
+            } else {
+                Err(io::Error::other(
+                    "io_uring does not support required operations",
+                ))
+            }
+        })??;
+
```
coszio: Why do we need to check this when opening every file? Naively I would only check this at uring initialization time

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -56,118 +83,213 @@ impl UniversalRead<u8> for IoUringFile {
...
-    fn read<const SEQUENTIAL: bool>(&self, range: ElementsRange) -> Result<Cow<'_, [u8]>> {
-        IO_URING.with_borrow_mut(|io_uring| {
+    fn read<const SEQUENTIAL: bool>(&self, range: ElementsRange) -> Result<Cow<'_, [T]>> {
+        with_uring(|io_uring| {
             let mut rt = IoUringRuntime::new(io_uring);
 
-            let entry = rt.buffers.read(0, self.fd(), range)?;
+            let (byte_offset, byte_length) = Self::element_range_to_bytes::<T>(range);
+            let entry = rt.state.read(0, self.fd(), byte_offset, byte_length)?;
             rt.enqueue_single(entry)?;
```
coszio, line 83: I am not 100% sure returning the `PushError` (when submission queue was full) back to the caller is the best way to go here. Since the `sq` is shared across all files, if disk is slow and there's many threads submitting, I can definitely see this happening.

IMO it should have some retry mechanism internally.

(Same for write methods)

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -56,118 +83,213 @@ impl UniversalRead<u8> for IoUringFile {
...
                         return Ok(None);
                     };
 
-                    let entry = buffers.read(id as _, self.fd(), range)?;
+                    let (byte_offset, byte_length) = Self::element_range_to_bytes::<T>(range);
+                    let entry = state.read(id as _, self.fd(), byte_offset, byte_length)?;
+                    Ok(Some(entry))
+                })?;
+
+                rt.submit_and_wait(1)?;
```
coszio, line 110: If for each read we always submit and wait for 1, even in batched case, we are making it synchronous, aren't we?

lib/common/common/src/universal_io/io_uring.rs
ffuugoo: Good point, I guess. I'll tweak this.

lib/common/common/src/universal_io/io_uring.rs
ffuugoo: This can never happen. io_uring is per-thread. The thread is blocked while operations are executed. There can never be two threads using the same io_uring instance. And queue depth is managed, so that we never overrun it.

lib/common/common/src/universal_io/io_uring.rs
ffuugoo: No, because all submitted entries run in parallel. So we submit _as many as we can_, and then wait for completion of _at least_ one (but more than one might have completed if we're lucky), and then we submit as many as we can again, so that there's always 16 ops running in parallel.

ffuugoo: Or, more precisely, it's not strictly guaranteed that _all_ (or any) ops running in parallel, kernel is free to execute them as it sees fit. But in practice they do, and the best we can do is submit as many as practical, so kernel _can_ run them in parallel.

ffuugoo: Ah, I get where confusion comes from. `submit_and_wait` submits _all_ operations, but waits for _at least_ 1 to complete.

lib/common/common/src/universal_io/io_uring.rs
ffuugoo: Fixed in `fix(io_uring)` PR that we merged into this one yesterday.

PR #8302, opened 2026-03-05
Rename to `StoredPointToValues`

coszio:
This is basically an LSP-generated PR that renames:

- `MmapPointToValues` to `StoredPointToValues`
- `MmapValue` -> `StoredValue`, and its methods

It is done as a separate PR to make the review of #8298 easier


timvisee (approved):
Thanks for keeping the PR scoped to the rename

PR #8310, opened 2026-03-06
Disable deferred point filtering in resharding

JojiiOfficial:
This PR adds a flag to internal scroll functions, which allows us to disable filtering deferred points in resharding.

### DIFF
--- a/lib/collection/src/tests/deferred_points_tests.rs
+++ b/lib/collection/src/tests/deferred_points_tests.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use common::budget::ResourceBudget;
 use common::counter::hardware_accumulator::HwMeasurementAcc;
 use common::save_on_disk::SaveOnDisk;
+use common::types::DeferredBehavior;
 use segment::data_types::vectors::VectorStructInternal;
 use segment::types::{WithPayload, WithVector};
 use shard::operations::CollectionUpdateOperations;
@@ -83,6 +84,7 @@ async fn retrieve_point(shard: &LocalShard, point_id: u64) -> bool {
             &current_runtime,
             None,
             HwMeasurementAcc::new(),
+            DeferredBehavior::Filter,
         )
         .await
         .unwrap();


lib/collection/src/shards/proxy_shard.rs
```diff
@@ -223,6 +223,8 @@ impl ShardOperation for ProxyShard {
                             &runtime_handle,
                             hw_measurement_acc.clone(),
                             None, // no timeout on update path
+                            // Ignoring deferred points here since they could be part of the update operation.
```
JojiiOfficial: I suggest to ignore deferred points here as well for the commented reason. 
Please let me know if this is not preferred!

lib/collection/src/shards/replica_set/mod.rs
```diff
@@ -1071,6 +1071,7 @@ impl ShardReplicaSet {
                     &self.search_runtime,
                     None,
                     hw_measurement_acc.clone(),
+                    true,
```
JojiiOfficial: This  function is only used in resharding. More precisely, when aborting. Therefore it could make sense to include deferred points when cleaning up partial transfers.

lib/collection/src/shards/replica_set/read_ops.rs
```diff
@@ -81,6 +81,7 @@ impl ShardReplicaSet {
                             &search_runtime,
                             timeout,
                             hw_acc,
+                            false, // TODO(jonas): Maybe use `true` here?
```
JojiiOfficial: This function is only used in [ShardCleanTask::clean_task()](https://github.com/qdrant/qdrant/blob/249b9ff0dc5b5780f2bf7663fe464d7440cbc32e/lib/collection/src/collection/clean.rs#L286) and it seems that this cleanup function should also consider deferred points.
Thoughts?

lib/collection/src/shards/replica_set/mod.rs
timvisee: It should include deferred points.

The operation must be able to get rid of all point IDs that don't belong in the shard, including deferred points.

lib/collection/src/shards/replica_set/read_ops.rs
timvisee: Let's add a parameter. The clean task should also (scroll and) delete deferred points.

lib/collection/src/shards/shard_holder/resharding.rs
```diff
@@ -339,7 +340,13 @@ impl ShardHolder {
                 let filter = self.hash_ring_filter(id).expect("hash ring filter");
                 let filter = Filter::new_must_not(Condition::new_custom(Arc::new(filter)));
                 shard
-                    .delete_local_points(filter, HwMeasurementAcc::disposable(), true) // Internal operation, no performance tracking needed
+                    .delete_local_points(
+                        filter,
+                        // Internal operation, no performance tracking needed
+                        HwMeasurementAcc::disposable(),
+                        true,
+                        OverwriteDeferredFiltering::IncludeAll, // Including also deferred points when aborting resharding.
```
timvisee: In my opinion the comment(s) are obsolete. Because the context and variant describes the same.

```suggestion
                        OverwriteDeferredFiltering::IncludeAll,
```

But feel free to keep it. No strong opinion.

lib/shard/src/update.rs
```diff
@@ -890,7 +893,14 @@ fn points_by_filter(
...
     let is_stopped = AtomicBool::new(false);
     segments.for_each_segment(|s| {
-        let points = s.read_filtered(None, None, Some(filter), &is_stopped, hw_counter);
+        let points = s.read_filtered(
+            None,
+            None,
+            Some(filter),
+            &is_stopped,
+            hw_counter,
+            OverwriteDeferredFiltering::IncludeAll,
```
timvisee:
```suggestion
            // Read operation used for updates, so we must handle all points
            OverwriteDeferredFiltering::IncludeAll,
```

lib/common/common/src/types.rs
```diff
@@ -92,3 +92,50 @@ impl From<usize> for DetailsLevel {
         }
     }
 }
+
+/// Overwrite the filtering of deferred points.
+/// Can be used in places where we sometimes must access all points, including deferred ones.
+#[derive(Clone, Copy)]
+pub enum OverwriteDeferredFiltering {
```
timvisee: Nit: to me this name is quite verbose. I also find the concept of 'overwrite' a bit confusing.

Simpler names may be: `DeferredBehavior`, `DeferredVisibility`, `WithDeferred` or just `Deferred`.

And have `With/Without`, `IncludeDeferred/ExcludeDeferred`, `Include/Exclude`. Or are  those still ambiguous?

lib/common/common/src/types.rs
```diff
@@ -92,3 +92,50 @@ impl From<usize> for DetailsLevel {
...
+    // No overwrite. Points will not be included in
+    // the result if they are deferred.
```
timvisee:
```suggestion
    /// Deferred points are not affected nor visible.
```

lib/common/common/src/types.rs
```diff
@@ -92,3 +92,50 @@ impl From<usize> for DetailsLevel {
...
+    None,
+
+    /// Overwrite the filtering of deferred points and include *all* points in the response, even deferred ones.
```
timvisee:
```suggestion
    /// Deferred points are affected and visible.
```

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -315,6 +316,7 @@ impl LocalShard {
                 search_runtime_handle,
                 timeout,
                 hw_measurement_acc,
+                DeferredBehavior::IncludeAll, // Points are already filtered in `point_ids`.
```
timvisee: This isn't sound.

Even though we've collected point IDs, including deferred points may still gives us a newer version of the point. That is not desired.

I'd recommend to simply use the same behavior here.

Do you have any specific reason for ignoring the user provided behavior here?

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -459,6 +461,7 @@ impl LocalShard {
                 search_runtime_handle,
                 timeout,
                 hw_measurement_acc,
+                DeferredBehavior::IncludeAll, // Points are already filtered in `selected_points`.
```
timvisee: Ditto. Change to user provided.

lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -210,6 +210,7 @@ impl LocalShard {
                 search_runtime_handle,
                 timeout,
                 hw_measurement_acc,
+                DeferredBehavior::IncludeAll, // Points are already filtered in `point_ids`.
```
timvisee: Ditto. Change to user provided.

lib/edge/src/scroll.rs
```diff
@@ -171,6 +171,7 @@ impl EdgeShard {
             DEFAULT_EDGE_TIMEOUT,
             &AtomicBool::new(false),
             hw_measurement_acc,
+            DeferredBehavior::IncludeAll, // Points are already filtered in `point_ids`.
```
timvisee: Let's use same as above here.

lib/edge/src/scroll.rs
```diff
@@ -226,6 +227,7 @@ impl EdgeShard {
...
```
timvisee: Let's use same as above here.

lib/edge/src/scroll.rs
```diff
@@ -330,6 +332,7 @@ impl EdgeShard {
             DEFAULT_EDGE_TIMEOUT,
             &AtomicBool::new(false),
             hw_measurement_acc,
+            DeferredBehavior::IncludeAll, // Points are already filtered in `random_point_ids`.
```
timvisee: Let's use same as above here.

lib/segment/src/segment/search.rs
```diff
@@ -51,6 +53,7 @@ impl Segment {
             with_vector,
             hw_counter,
             is_stopped,
+            DeferredBehavior::IncludeAll, // We already filter in read operations.
```
timvisee: I don't think this is correct, and it should use the same behavior as the caller. Similar to the other comments.

lib/collection/src/shards/remote_shard.rs
```diff
@@ -1280,6 +1282,7 @@ impl ShardOperation for RemoteShard {
         _search_runtime_handle: &Handle,
         timeout: Option<Duration>,
         hw_measurement_acc: HwMeasurementAcc,
+        _deferred_behavior: DeferredBehavior,
```
timvisee, line 1287: I see the deferred behavior is simply ignored here. I'm not sure it's good enough, and it's a potential foot gun. At the caller, we might not know what kind of shard we're working with.

Just stating it here, because I don't see an obvious solution right away. I have to think about it for a second.

In case of count in transfers, maybe we've to get the local shard explicitly and call count directly on it so we don't have the expose deferred logic in the shared trait.

lib/segment/src/segment/search.rs
timvisee: I believe this must use the same behavior as the caller. Otherwise we might return deferred points in read operations where we respect deferred points.

lib/collection/src/shards/local_shard/scroll.rs
JojiiOfficial: > Do you have any specific reason for ignoring the user provided behavior here?

I was trying to avoid the [additional allocation](https://github.com/qdrant/qdrant/blob/ab2bf0baeb375c51a08fce41879394e2bda9739b/lib/segment/src/segment/entry.rs#L185) that comes with filtering in `retrieve()`. I didn't think of the special case where a single segment has a point deferred and non-deferred.

lib/edge/src/query.rs
```diff
@@ -426,6 +427,7 @@ impl EdgeShard {
             DEFAULT_EDGE_TIMEOUT,
             &AtomicBool::new(false),
             hw_measurement_acc,
+            DeferredBehavior::Filter,
```
JojiiOfficial, line 430: Also changed the behavior to `Filter` here, since it's always used in already filtered context

lib/collection/src/shards/local_shard/query.rs
```diff
@@ -133,6 +136,7 @@ impl LocalShard {
                 &self.search_runtime,
                 timeout,
                 hw_measurement_acc,
+                DeferredBehavior::Filter,
```
JojiiOfficial, line 139: Same here

lib/segment/src/segment/search.rs
```diff
@@ -21,6 +21,8 @@ use crate::types::{ScoredPoint, WithPayload, WithVector};
 
 impl Segment {
     /// Converts raw ScoredPointOffset search result into ScoredPoint result
+    ///
+    /// Doesn't filter deferred points.
```
timvisee, line 25: I believe this comment must be updated

lib/collection/src/shards/remote_shard.rs
JojiiOfficial: I added a TODO: https://github.com/qdrant/qdrant/pull/8310/commits/3a0ae797ccba498c30529d9c77821ace0c019edf
I'll handle it later in a separate PR as discussed.

PR #8311, opened 2026-03-06
Feat/storage api over universal io

dancixx:
## Storage Read API over UniversalIO — Summary

This PR introduces a new **public** gRPC service (`StorageRead`) that exposes the `UniversalRead<u8>` trait over the network. Currently operates at the **node level** — reads are served from the local node's storage only.

> **Note:** Internal forwarding (routing reads to the correct peer based on shard ownership) is not yet implemented. Callers must currently target the correct node directly. Forwarding support is planned as a follow-up.

### Endpoints

| RPC | Maps to | Description |
|-----|---------|-------------|
| `FileExists` | `UniversalReadFileOps::exists` | Check if a file exists |
| `ListFiles` | `UniversalReadFileOps::list_files` | List files by prefix, returns paths relative to collection dir |
| `FileLength` | `UniversalRead::open` + `len()` | Get file size |
| `ReadBytes` | `UniversalRead::read` | Read a single byte range |
| `ReadBytesStream` | `UniversalRead::read` (chunked) | Stream large reads in ~1 MB chunks |
| `ReadWhole` | `UniversalRead::read_whole` | Read entire file |
| `ReadBatch` | `UniversalRead::read_batch` | Multiple ranges from one file |
| `ReadMulti` | `UniversalRead::read_multi` | Ranges across multiple files (deduplicates opens) |

### Key design decisions

- **Public API, node-level scope** — registered on the public gRPC server with reflection enabled; reads are served from the local node's storage
- **Generic over `S: UniversalRead<u8>`** — defaults to `IoUringFile` on Linux, `MmapUniversal<u8>` elsewhere; testable with other backends
- **Path security** — `resolve_path` rejects non-`Normal` path components (prevents `..` traversal) and canonicalizes existing paths to verify they remain under the collection directory
- **All I/O runs in `spawn_blocking`** — mmap reads can trigger page faults (blocking disk I/O), so every read operation is dispatched to the blocking thread pool
- **Stream backpressure** — `ReadBytesStream` opens the file once, then reads each chunk in a separate `spawn_blocking` call so backpressure doesn't pin a worker thread
- **Length clamping** — `ReadBytesStream` clamps the requested length to the actual file size so it stops at EOF instead of producing an `OutOfBounds` error
- **Error propagation** — `file_exists` propagates real errors (permission denied, I/O errors) instead of silently treating them as "not found"

### Test coverage

13 tests covering all endpoints including edge cases:
- Path traversal rejection
- Not-found error codes
- Out-of-bounds error codes  
- Stream chunking correctness
- Zero-length stream (early return, no file open)
- Length clamping when requested range exceeds file size

### Future work

- **Internal forwarding**: Route requests to the correct peer based on shard ID extracted from the file path, using `ChannelService` + `StorageReadClient` (same pattern as `RemoteShard`)

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/tests/consensus_tests/auth_tests/test_jwt_access.py
+++ b/tests/consensus_tests/auth_tests/test_jwt_access.py
@@ -1985,7 +1985,7 @@ def test_storage_read_read_bytes():
         grpc_request={
             "collection_name": COLL_NAME,
             "path": STORAGE_READ_TEST_PATH,
-            "offset": 0,
+            "byteOffset": 0,
             "length": 1,
         },
     )
@@ -1997,7 +1997,7 @@ def test_storage_read_read_bytes_stream():
         grpc_request={
             "collection_name": COLL_NAME,
             "path": STORAGE_READ_TEST_PATH,
-            "offset": 0,
+            "byteOffset": 0,
             "length": 1,
         },
     )
@@ -2016,7 +2016,7 @@ def test_storage_read_read_batch():
         grpc_request={
             "collection_name": COLL_NAME,
             "path": STORAGE_READ_TEST_PATH,
-            "ranges": [{"offset": 0, "length": 1}],
+            "ranges": [{"byteOffset": 0, "length": 1}],
         },
     )
 
@@ -2026,6 +2026,6 @@ def test_storage_read_read_multi():
         "storage_read_read_multi",
         grpc_request={
             "collection_name": COLL_NAME,
-            "reads": [{"path": STORAGE_READ_TEST_PATH, "offset": 0, "length": 1}],
+            "reads": [{"path": STORAGE_READ_TEST_PATH, "byteOffset": 0, "length": 1}],
         },
     )


src/tonic/api/storage_read_api.rs
```diff
@@ -0,0 +1,1155 @@
...
+
+/// Chunk size for streaming reads (~1 MB).
+const STREAM_CHUNK_SIZE: u64 = 1024 * 1024;
+
+pub struct StorageReadService<S: UniversalRead<u8> + Send + Sync + 'static = MmapUniversal<u8>> {
+    dispatcher: Arc<Dispatcher>,
+    _marker: PhantomData<S>,
+}
+
+impl<S: UniversalRead<u8> + Send + Sync + 'static> StorageReadService<S> {
```
generall: This file is too big, I would prefer to split it into: helpers + impl StorageReadService, impl StorageRead, tests

lib/api/src/grpc/proto/storage_read_service.proto
```diff
@@ -0,0 +1,141 @@
...
+// Maps to common::mmap::Advice — controls mmap access pattern hint.
+// If not set in StorageOpenOptions, the global default is used.
+enum MmapAdvice {
+  MmapAdviceNormal = 0;
+  MmapAdviceRandom = 1;
+  MmapAdviceSequential = 2;
+}
+
+// Maps to common::universal_io::OpenOptions — controls how a file is opened.
+message StorageOpenOptions {
```
generall: I don't think we should expose this in API, hard-coding default values should be enough

src/tonic/api/storage_read_api/helpers.rs
```diff
@@ -0,0 +1,228 @@
...
+        Self {
+            dispatcher,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Return the root directory that contains all collections.
+    fn collections_base_path(&self, auth: &storage::rbac::Auth) -> PathBuf {
+        let pass = new_unchecked_verification_pass();
+        let toc = self.dispatcher.toc(auth, &pass);
```
generall: this should verify we have at least read access to collection

PR #8315, opened 2026-03-06
Filter deferred points: retrieve

JojiiOfficial:
Depends on https://github.com/qdrant/qdrant/pull/8299

Implements filtering of deferred points from retrieve operations.

Todo:
- [x] Implementation
- [x] Tests

### DIFF
--- a/lib/segment/src/segment/tests.rs
+++ b/lib/segment/src/segment/tests.rs
@@ -966,6 +966,8 @@ fn test_dense_deferred_point_segment_combinations() {
     }
 }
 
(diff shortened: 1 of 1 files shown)


timvisee:
Imo filtering deferred should be optional as described here: <https://github.com/qdrant/qdrant/pull/8310#pullrequestreview-3921405942>

But we can probably handle it in that PR.

JojiiOfficial:
> But we can probably handle it in that PR.

It's implemented in that PR now: 9c3e18135a92c93a02c7b71b4681aa93df1da448

PR #8317, opened 2026-03-06
feat: add collection health recommendations endpoint

Shiven0504:
## Problem

Qdrant exposes collection info like segment count, vector count, and config — but gives no guidance on whether the current setup is optimal. Users troubleshooting slow searches or high memory usage have to manually check each config value and guess what to change.

## What this PR adds

A new `GET /collections/{name}/health` endpoint that analyzes the collection's current state and returns actionable tuning suggestions.

### New types in `types.rs`
- `RecommendationSeverity` — `info`, `warning`, `critical`
- `RecommendationCategory` — `indexing`, `segmentation`, `memory`, `hnsw_config`, `wal_config`, `payload_index`
- `HealthRecommendation` — a single recommendation with `severity`, `category`, `message`, and `suggestion`
- `CollectionHealthReport` — the response containing a list of recommendations and a `HealthSummary`

### New module `health_check.rs`
Contains the analysis logic with 6 rules:
1. **Low indexed vector ratio** — flags when less than 80% (warning) or 50% (critical) of vectors are indexed
2. **High segment count** — flags fragmentation at 20+ (warning) or 50+ (critical) segments
3. **No quantization on large collections** — suggests enabling quantization when 100K+ points have none
4. **Low `ef_construct`** — flags when large collections use a low `ef_construct` value
5. **Small WAL capacity** — flags when `wal_capacity_mb` is under 32 for large collections
6. **Missing payload indexes** — suggests creating indexes when 10K+ points have no payload indexes

### Wiring
- `do_get_collection_health()` added in `src/common/collections.rs` — fetches `CollectionInfo` and runs the analysis
- `get_collection_health` handler added in `src/actix/api/collections_api.rs` — registered as a service

### Example response
```json
{
  "collection_name": "my_collection",
  "recommendations": [
    {
      "severity": "warning",
      "category": "indexing",
      "message": "Only 45% of vectors are indexed (4500/10000).",
      "suggestion": "Wait for the optimizer to finish indexing, or lower indexing_threshold in optimizer config."
    }
  ],
  "summary": { "critical": 0, "warning": 1, "info": 0 }
}
```

## Files changed
| File | What changed |
|------|-------------|
| `lib/collection/src/operations/types.rs` | Added recommendation structs |
| `lib/collection/src/operations/health_check.rs` | New — analysis logic + 8 unit tests |
| `lib/collection/src/operations/mod.rs` | Registered `health_check` module |
| `src/common/collections.rs` | Added `do_get_collection_health()` |
| `src/actix/api/collections_api.rs` | Added endpoint + service registration |

## Test plan
- [ ] Unit tests in `health_check.rs` cover all 6 rules and summary counting
- [ ] Integration test: create a collection, insert points, call `GET /collections/{name}/health`
- [ ] Verify empty collection returns no recommendations
- [ ] Verify response schema matches `CollectionHealthReport`

### DIFF
--- a/lib/collection/src/operations/health_check.rs
+++ b/lib/collection/src/operations/health_check.rs
@@ -227,14 +227,15 @@ fn check_payload_indexes(info: &CollectionInfo, recs: &mut Vec<HealthRecommendat
 #[cfg(test)]
 mod tests {
     use std::collections::HashMap;
-    use std::num::NonZeroU32;
+    use std::num::{NonZeroU32, NonZeroU64};
 
-    use segment::types::HnswConfig;
+    use segment::types::{Distance, HnswConfig};
 
     use super::*;
     use crate::config::{CollectionParams, WalConfig, default_on_disk_payload};
     use crate::operations::types::{
-        CollectionConfig, CollectionInfo, CollectionStatus, OptimizersStatus, VectorsConfig,
+        CollectionConfig, CollectionInfo, CollectionStatus, OptimizersStatus, VectorParams,
+        VectorsConfig,
     };
 
     fn default_optimizers_config() -> crate::optimizers_builder::OptimizersConfig {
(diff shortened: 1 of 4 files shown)


generall:
Neither of the rules you suggested are universal, and will most likely give false-positive warnings. I don't agree that's useful.

Shiven0504:
Hey, I understand the concern about false positives, I have reverted those threshold and per-vector ef_construct changes. They were suggestions from CodeRabbitAI and I agree that they would add noise more than value.

The new commit just fixes two things:                                                                                                    
  - Added missing doc comments to pass the docstring coverage check (was failing at 70%) 
  - Fixed the test helper since it was using an empty vector config so the indexed-ratio test was silently passing without actually testing anything

PR #8318, opened 2026-03-06
Test edge examples on CI

xzfc:
This PR enables edge example testing on CI, adds `Justfile` to simplify edge development, and updates Python edge readme.

Review per commit. Commits without ❗ are minor.

The updated publish workflows are tested here (see CI checks): https://github.com/qdrant/qdrant/commits/ci-edge-examples-dontmerge; also see https://github.com/qdrant/qdrant/pull/8318#issuecomment-4035368652.

---

- **lib/edge/publish/cargo: improve target directory discovery**
  Don't hardcode workspace root, ask cargo instead.

- **Fix warning in qdrant-edge amalgamation**
  `cargo check` in `lib/edge/publish` produces the following warning:

      warning: type `PointerUpdates` is more private than the item `gridstore::tracker::Tracker::<S>::write_pending`

- **lib/edge/publish/examples: fix compiler warnings**

- **refactor: lib/edge/publish/examples: put DATA_DIRECTORY into lib.rs**
  To match python examples that have the same constant in `common.py`.

- **Use `lib/edge/data` dir for both Python and Rust examples**
  Before this commit, `lib/edge/publish/data` and `lib/edge/python/data`
  were used. I'd like Python and Rust to use the same prepared resources
  in the future.

- **❗Add lib/edge/Justfile with recipes to build/run examples**
  I constantly keep forgetting how to compile and run edge stuff.
  Also, this would be used in CI in subsequent commits.

- **❗doc: lib/edge/README.md: Combine Rust and Python Edge readmes**
  Also now we can rename `creates-io-readme.md` to `README.md` like
  reasonable people.

- **❗doc: Add qdrant-edge-py README.md**
  Since the previous commit removed qdrant-edge-py README.md, we need to
  add something in its place instead.
  The new README is adapted from Rust edge README.
  And the previous README wasn't good anyway, see \[1], it mentions maturin
  which is irrelevant for users.

  \[1]: https://pypi.org/project/qdrant-edge-py/0.5.0/

- **GHA: Add "Setup Qdrant" composite action**
  We need running Qdrant instance to test edge examples on CI. We can't
  use docker for this because we want to run tests on Windows and macOS.
  OTOH, this action downloads binaries for all platforms provided in
  https://github.com/qdrant/qdrant/releases/tag/v1.17.0

- **❗GHA: Use single worflow for python and rust edge examples**
  Merge workflows to reuse build cache. Also, modernize it by using `uv`,
  and recently introduced `setup-qdrant` and Justfile. Also, make it run
  on three platforms (only on merges).

- **❗GHA: edge-{py,rust}-release: also run examples before publishing**
  Use similar approach as in the previous commit. But note that this time
  we build/test also on ARM Linux and Windows.

### DIFF
--- a/.github/workflows/edge-py-release.yml
+++ b/.github/workflows/edge-py-release.yml
@@ -141,6 +141,60 @@ jobs:
           path: lib/edge/python/dist
 
 
+  edge-py-test:
+    name: Test Python wheels [${{ matrix.os }}, ${{ matrix.python }}]
+
+    runs-on: ${{ matrix.os }}
+    needs:
+      - edge-py-linux
+      - edge-py-macos
+      - edge-py-windows
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-24.04-arm
+          - ubuntu-latest
+          - windows-latest
+        python:
+          - '3.10'
+          - 'pypy3.11'
+
+    steps:
+      - name: Checkout Qdrant
+        uses: actions/checkout@v6
+
+      - uses: extractions/setup-just@v3
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+
+      - name: Install Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Download wheels
+        uses: actions/download-artifact@v8
+        with:
+          pattern: qdrant-edge-*
+          merge-multiple: true
+          path: lib/edge/python/dist
+
+      - name: Setup Qdrant
+        uses: ./.github/actions/setup-qdrant
+
+      - name: Install and test wheel
+        working-directory: lib/edge/python
+        run: |
+          uv venv --allow-existing --python ${{ matrix.python }}
+          uv pip install --no-index --find-links dist/ qdrant-edge-py
+          uv pip install requests
+          just py-examples
+
+
   merge-artifacts:
     name: Merge release artifacts
 
@@ -149,6 +203,7 @@ jobs:
       - edge-py-linux
       - edge-py-macos
       - edge-py-windows
+      - edge-py-test
 
     if: inputs.publish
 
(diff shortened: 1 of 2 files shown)


xzfc:
Curious finding: this example sometimes fails on pypi3.11, but not on python3.10:
https://github.com/qdrant/qdrant/actions/runs/22929684202/job/66550289791

```
Running example: examples/load-existing.py
  ---- Load shard ----
  ([Record(id=1, vector=[6, 9, 4, 2], payload={"null": None, "str": "string", "uint": 42, "int": -69, "float": 4.2, "bool": True, "obj": {"null": None, "str": "string", "uint": 42, "int": -69, "float": 4.2, "bool": True, "obj": {}, "arr": []}, "arr": [None, "string", 42, -69, 4.2, True, {}, []]}, order_value=None), Record(id="819d7aa4-00f0-4d25-9c87-ff9080f89c2c", vector=[1, 6, 4, 2], payload={"hello": "world", "price": 999.99}, order_value=None), Record(id="e9408f2b-b917-4af1-ab75-d97ac6b2c047", vector=[6, 9, 3, -2], payload={"hello": "world", "price": 199.99}, order_value=None)], None)
  Traceback (most recent call last):
    File "/home/runner/work/qdrant/qdrant/lib/edge/python/examples/load-existing.py", line 57, in <module>
      edge = EdgeShard(TMP_DIR, None)
             ^^^^^^^^^^^^^^^^^^^^^^^^
  Exception: Service runtime error: failed to open WAL /home/runner/work/qdrant/qdrant/lib/edge/data/tmp/wal: Can't init WAL: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
```

Perhaps GC/`__del__` race condition:
https://github.com/qdrant/qdrant/blob/f6b57fc464aa67a000d1b4633b618e5a9a9d4784/lib/edge/python/examples/load-existing.py#L56-L57

Will investigate later, maybe outside of this PR.

PR #8339, opened 2026-03-09
Use `UniversalIo` for `MmapBitSlice`

coszio:
Since bitslice uses primitive integers as elements which store the bits, we cannot just do `MmapUniversal<u1>`, so we need to re-map bit index to the integer element (and position within it).

For writes, two write methods are included:
- `set_bits_batch`: coalesces updates to a single element into a single write
- `write_bitslice`: used during creation when the source is another bitslice.

For reads, you can try reading a single bit (reads u64 element), or request the entire bitslice if needed

## Future work
- [ ] Finish propagating Results

### DIFF
--- a/lib/segment/src/common/stored_bitslice.rs
+++ b/lib/segment/src/common/stored_bitslice.rs
@@ -501,7 +501,7 @@ mod tests {
                     .flat_map(|el: u64| (el * 64..el * 64 + 64).map(|i| (i, true))),
             )
             .unwrap();
-        assert_bits(&storage, |i| matches!(i / 64, 0 | 3 | 7 | 15));
+        assert_bits(&storage, |i| matches!(i / 64, 0 | 3 | 7 | 111));
 
         // Out of bounds
         assert!(
@@ -514,7 +514,7 @@ mod tests {
         storage
             .set_ascending_bits_batch(std::iter::empty())
             .unwrap();
-        assert_bits(&storage, |i| matches!(i / 64, 0 | 3 | 7 | 15));
+        assert_bits(&storage, |i| matches!(i / 64, 0 | 3 | 7 | 111));
     }
 
     #[test]


lib/segment/src/common/stored_bitslice.rs
```diff
@@ -0,0 +1,455 @@
...
+                length: 1,
+            })?[0];
+            let old_element = element;
+            let new_element = &mut element;
+
+            let bitslice = BitSlice::from_element_mut(new_element);
+
+            // Do all modifications to this element
+            for (bit_idx, value) in chunk {
+                let bit_within_element = bit_idx as u8 & <u64 as BitRegister>::MASK;
```
generall: `bit_idx as u8` is this a way to compute global_offset % container_size = offset within container?

generall: claude thinks it is fine, but I would propose to add a comment with equivalent human-readable expression

lib/segment/src/common/stored_bitslice.rs
```diff
@@ -0,0 +1,455 @@
...
+impl<S: UniversalWrite<u64>> StoredBitSlice<S> {
+    /// Set multiple individual bits in a batch.
+    ///
+    /// Each `(bit_index, value)` pair sets a single bit. If contiguous, bits within the same
+    /// `u64` element are coalesced into a single read-modify-write.
+    pub fn set_bits_batch(&mut self, updates: impl IntoIterator<Item = (u64, bool)>) -> Result<()> {
+        // group by bits on the same element
+        let chunks = updates
+            .into_iter()
+            .chunk_by(|(bit_index, _)| bit_index >> <u64 as BitRegister>::INDX);
```
generall: also comment with human-readable equivalent here please

lib/segment/src/common/stored_bitslice.rs
```diff
@@ -0,0 +1,455 @@
...
+
+            // Do all modifications to this element
+            for (bit_idx, value) in chunk {
+                let bit_within_element = bit_idx as u8 & <u64 as BitRegister>::MASK;
+                bitslice.set(bit_within_element as usize, value);
+            }
+
+            // Write if it changed
+            if old_element != *new_element {
+                self.storage.write(element_idx, &[*new_element])?;
```
generall: running a lot of small writes in a loop is potentially expensive. Since we already are doing batch, let's do the batch write as well

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -246,14 +246,15 @@ impl ImmutableIdTracker {
...
+            Self::deleted_file_path(segment_path),
+            OpenOptions {
+                populate: Some(true),
+                ..OpenOptions::default()
+            },
         )?;
-        let deleted_mmap = MmapBitSlice::try_from(deleted_raw, 0)?;
-        let deleted_bitvec = deleted_mmap.to_bitvec();
-        let deleted_wrapper = MmapBitSliceBufferedUpdateWrapper::new(deleted_mmap);
+        let deleted_bitvec: BitVec = deleted_storage.read_all()?.iter().by_vals().collect();
```
generall: Would it be possible to avoid per-bit iterator here? As I understood it should be enough for `deleted_storage` to use usize instead of u64

lib/segment/src/index/field_index/geo_index/mmap_geo_index.rs
```diff
@@ -171,21 +171,24 @@ impl MmapGeoMapIndex {
         }
 
         {
-            let deleted_flags_count = dynamic_index.point_to_values.len();
-            let deleted_file = create_and_ensure_length(
+            let _ = create_and_ensure_length(
```
generall, line 174: logic like this seems to be duplicated in 4 places, would it be possible to unify?

lib/segment/src/common/stored_bitslice.rs
coszio: Yes, this is how it is done within the `bitvec` crate ([source](https://docs.rs/bitvec/latest/src/bitvec/index.rs.html#268))

lib/segment/src/id_tracker/immutable_id_tracker.rs
coszio: It's dependent on `CompressedPointMappings`. I have changed it so that, if `usize` is `u64`, it will be a cheap operation, by using `extend_from_bitslice` instead.

lib/segment/src/index/field_index/geo_index/mmap_geo_index.rs
coszio: It is possible, but I didn't want to get ahead and add the create function without having the necessary ops in `UniversalIoFileOps`

lib/segment/src/common/stored_bitslice.rs
```diff
@@ -143,71 +143,63 @@ impl<S: UniversalRead<BitStore>> StoredBitSlice<S> {
...
     /// Set multiple individual bits in a batch.
     ///
-    /// Each `(bit_index, value)` pair sets a single bit. Bits within the same
-    /// `u64` element are coalesced into a single read-modify-write, and
-    /// consecutive modified elements are grouped into contiguous runs
-    /// written via a single `write_batch` call.
+    /// Each `(bit_index, value)` pair sets a single bit. Updates are grouped
+    /// by 4KB page so that each page is read and written at most once per
+    /// consecutive run. Writes happen immediately per page so that repeated
+    /// visits to the same page observe prior modifications ("last write wins").
```
generall: I don't think users of the universal IO should optimize for page size.
This looks to me as an abstraction leak and in this current scenario, this migtht be especially wrong, consider the following:

- we need to update a bunch of random bits in bitmask, stored on S3.
- S3 implementation can't do implace updates, it can only re-upload the whole file. 

So the best approach from storage point of view - send all updates at once into universal-io implementation and let it figure out changes on its own

generall: Disregard this, figured out that writes can actually happen on flush, so S3 change will be file

coszio (commented):
Sorry for the huge changeset, but this PR grew in size during development. 

It now includes:
- generic storage-compatible bitslice
- batched updates, coalesced by edits to consecutive elements
- changing to BitSlice/BitVec with `<u64, Lsb0>` elements everywhere

lib/segment/src/vector_storage/dense/mmap_dense_vectors.rs
```diff
@@ -0,0 +1,325 @@
...
+use crate::vector_storage::query_scorer::is_read_with_prefetch_efficient;
+use crate::vector_storage::{AccessPattern, Random, Sequential};
+
+const HEADER_SIZE: usize = 4;
+const VECTORS_HEADER: &[u8; HEADER_SIZE] = b"data";
+const DELETED_HEADER: &[u8; HEADER_SIZE] = b"drop";
+
+/// Mem-mapped file for dense vectors
+#[derive(Debug)]
+pub struct MmapDenseVectors<T: PrimitiveVectorElement> {
```
generall: do we need this file in this PR?

PR #8341, opened 2026-03-09
Restrict snaphsot recovery to snapshots directory, hide file contents

generall:
Only allow `file://...` snapshots to be recoverred from snapshot directly, for security reasons.

### DIFF
--- a/tests/consensus_tests/auth_tests/test_jwt_access.py
+++ b/tests/consensus_tests/auth_tests/test_jwt_access.py
@@ -1,6 +1,6 @@
 import json
-import tempfile
 from inspect import isfunction
+from pathlib import Path
 from typing import Callable, List, Optional, Tuple, Union
 
 import grpc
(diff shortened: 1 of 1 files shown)


lib/storage/src/content_manager/snapshots/download.rs
```diff
@@ -51,13 +51,42 @@ async fn _download_snapshot(
...
+    })?;
+
+    let canonical_allowed = fs_err::canonicalize(allowed_dir).map_err(|err| {
+        StorageError::service_error(format!(
+            "Failed to resolve snapshots directory {allowed_dir:?}: {err}"
+        ))
+    })?;
+
+    if !canonical_path.starts_with(&canonical_allowed) {
+        return Err(StorageError::bad_request(format!(
```
timvisee: How about we change this to forbidden?


```suggestion
        return Err(StorageError::forbidden(format!(
```

tests/openapi/test_snapshot.py
```diff
@@ -422,3 +422,27 @@ def test_collection_snapshot_security(collection_name):
...
+    response = request_with_validation(
+        api='/collections/{collection_name}/snapshots/recover',
+        method="PUT",
+        path_params={'collection_name': "somethingthatdoesnotexist"},
+        body={
+            "location": "file:///etc/passwd",
+        }
+    )
+    assert response.status_code == 400
+    assert response.json()["status"]["error"].startswith("Bad request: Snapshot file path must be inside the snapshots directory")
```
timvisee: If we change the error response to forbidden, we need to change this to:

```suggestion
    assert response.status_code == 403
    assert response.json()["status"]["error"].startswith("Forbidden: Snapshot file path must be inside the snapshots directory")
```

lib/common/common/src/tar_unpack.rs
```diff
@@ -23,14 +23,32 @@ pub fn tar_unpack_reader<R: io::Read>(reader: R, dst: &Path) -> Result<R, io::Er
...
-    for entry in archive.entries()? {
-        let mut entry = entry?;
+    for entry in archive.entries().map_err(|err| {
+        io::Error::new(
+            err.kind(),
+            // Must hide error in release builds to not leak contents of potentially sensitive files
+            #[cfg(not(debug_assertions))]
+            format!("Malformed tar archive, unable to read entries"),
+            #[cfg(debug_assertions)]
+            format!("Malformed tar archive, unable to read entries: {err}"),
```
timvisee, line 33: A core problem was that we expose the tar error message, which could leak file contents.

This now hides the tar message in release builds. It is kept in debug builds as it's still useful for development.

Alternatively we can hide the message for both build types. Any opinions?

cc @generall

PR #8342, opened 2026-03-10
log cpu utilization

generall:
We want to know CPU utilization during slow requests:

- if utilization is high - CPU is a bottleneck and we are likely facing expensive request
- If utilization is low - it is either IO or locks (though, there are no locks inside segments, where we measure)

If IO is high, we can allow more threads for performance improvements

---

AI query:

```
find places where we use search runtime in qdrant. I want to measure if tasks on this runtime are CPU bounded or IO bounded. For this I want to record thread CPU utilization when we spawn tasks on this runtime.
CPU usage should be recorded into slow_requests log
```

### DIFF
--- a/lib/collection/src/collection_manager/segments_searcher.rs
+++ b/lib/collection/src/collection_manager/segments_searcher.rs
@@ -446,20 +446,28 @@ impl SegmentsSearcher {
             };
 
             let hw_counter = hw_measurement_acc.get_counter_cell();
-            let all_points: BTreeSet<_> = segments
-                .into_iter()
-                .flat_map(|segment| {
-                    segment.get().read().read_filtered(
-                        None,
-                        None,
-                        filter.as_ref(),
-                        &is_stopped,
-                        &hw_counter,
-                        deferred_behavior,
-                    )
-                })
-                .collect();
-            Ok(all_points)
+
+            let work = || {
+                let all_points: BTreeSet<_> = segments
+                    .into_iter()
+                    .flat_map(|segment| {
+                        segment.get().read().read_filtered(
+                            None,
+                            None,
+                            filter.as_ref(),
+                            &is_stopped,
+                            &hw_counter,
+                            deferred_behavior,
+                        )
+                    })
+                    .collect();
+                Ok(all_points)
+            };
+
+            match hw_counter.cpu_utilization() {
+                Some(cpu_util) => cpu_util.measure(work),
+                None => work(),
+            }
         });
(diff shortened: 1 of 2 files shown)


lib/common/common/src/cpu_utilization.rs
```diff
@@ -0,0 +1,139 @@
...
+impl CpuUtilization {
+    pub fn new() -> Self {
+        Self {
+            inner: Arc::new(CpuUtilizationInner::default()),
+        }
+    }
+
+    /// Wrap a closure, measuring wall time and thread CPU time.
+    /// Accumulates into the shared counters.
+    pub fn measure<F, R>(&self, f: F) -> R
```
timvisee, line 30: Do we expect any measurable impact from this?

generall: I don't think there will be anything expensive, but I haven't tested

dancixx:
@generall 

If the goal is to account for CPU utilization across all `spawn_blocking` work in this change, there are still a few gaps where blocking tasks run without `cpu_utilization.measure(...)`, so their CPU time will not be reflected in the reported ratio.

The missing cases are:

- `SegmentsSearcher::retrieve()` in `segments_searcher.rs#L399`
- `SegmentsSearcher::read_filtered()` in `segments_searcher.rs#L433`
- `SegmentsSearcher::rescore_with_formula()` in `segments_searcher.rs#L493`
- `LocalShard::unique_values()` in the exact facet path, in `facet.rs#L171`
- UpdateWorkers WAL record loading in `update_worker.rs#L84`
- `UpdateWorkers::wait_for_deferred_points_ready()` in `update_worker.rs#L203`

In addition, the main update path in `update_worker.rs#L121` is only partially covered: the blocking task is spawned there, but only the inner `CollectionUpdater::update(...)` section is wrapped in `measure(...)` at `update_worker.rs#L262`.

generall:
> If the goal is to account for CPU utilization across all spawn_blocking work in this change

I don't think this is a true goal.

What I want is - to cover all sections of query execution, which can be either compute-heavy or IO-heavy.

From this point of view - `wait_for_deferred_points_ready` is never a compute heavy operation.
(or `segment_guard.has_deferred_points()` is actually always compute heavy, but it is also very tiny)

I don't think I wan to log those parts of the query, as they would rather introduce noise then interesting signal.

I agree that measurement is meaningful for `read_filtered`, `rescore_with_formula`, `unique_values`, but not others.

I also don't think we should include any other parts of the `update_worker`

PR #8348, opened 2026-03-10
Use `array_windows` when statically sized `windows` were used

timvisee:
[`array_windows`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.array_windows) is now stabilized. Let's use it.

### All Submissions:

### DIFF
--- a/lib/segment/src/index/hnsw_index/gpu/batched_points.rs
+++ b/lib/segment/src/index/hnsw_index/gpu/batched_points.rs
@@ -155,7 +155,7 @@ impl BatchedPoints {
             batch_index += 1;
         }
 
-        for [a, b] in batches.array_windows(2) {
+        for [a, b] in batches.array_windows() {
             assert_eq!(a.end, b.start);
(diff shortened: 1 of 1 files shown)


lib/segment/src/index/hnsw_index/gpu/batched_points.rs
```diff
@@ -155,10 +155,8 @@ impl BatchedPoints {
             batch_index += 1;
         }
 
-        for batch_pair in batches.windows(2) {
-            if batch_pair.len() == 2 {
```
timvisee, line 159: Removed this if-statement because it's always true. The `windows` function already guaranteed a size of 2 before.

agourlay:
I think we need to bump the `rust-version` in the cargo.toml file to require 1.94 in that case.

timvisee:
> I think we need to bump the `rust-version` in the cargo.toml file to require 1.94 in that case.

Done

PR #8350, opened 2026-03-10
strictly read only mmap

generall:
```
For MmapUniversal implementation of UniversalRead I want to have completely read-only version, which doesn't open mutable memory map at all. At the same time I still need to have MmapUniversal            
  implementation with both read and writes, and I don't want to duplicate code. 
```


coszio:
Here is my suggestion: #8351

PR #8351, opened 2026-03-10
review: Alternative for #8350

coszio:
I did this as a review for #8350 in a separate PR since it was a substantial change.

Prompt:
> /simplify There are two very similar implementations, only varying in how they open the main mmap. Check if we can get rid of `MmapReadAccess`, or reuse it for opening based on the underlying type.

Then I just did a few nit edits, like renaming to `MmapAccess`.


generall:
To fix the test, we need to rebase on top of https://github.com/qdrant/qdrant/pull/8372 once merged

PR #8354, opened 2026-03-10
Deletions with deferred points bug

IvanPleshkov:
## Summary
Fix `delete_points` (by ID) failing to remove old non-deferred copies of points when the latest version is deferred
Add regression tests for delete-by-ID and delete-by-filter with deferred points

## Bug
When a point exists as both an old non-deferred copy in an optimized segment (visible to searches) and a newer deferred copy in an appendable segment (hidden until optimization), `delete_points` only deleted the deferred copy. The old non-deferred copy survived, leaving the point still visible to users after deletion.

Root cause: `delete_points` went through `apply_points` → `find_points_to_update_and_delete`, which intentionally keeps older non-deferred copies when the latest version is deferred-only (`latest_has_non_deferred == false`). This rule is correct for upsert/payload operations (keeps the point visible while deferred indexing completes), but wrong for deletes where the user expects the point to disappear entirely.

Reproduction: Insert points, wait for optimization, overwrite all points (creates deferred copies), then delete all by ID — 196/200 points survive.

Notably, delete_points_by_filter was not affected because it bypasses find_points_to_update_and_delete and iterates all segments directly.

## Fix
Changed `delete_points` to iterate all segments directly and call `delete_point` on each — the same pattern already used by `delete_points_by_filter`. No changes to `find_points_to_update_and_delete` or `apply_points` (their behavior remains correct for non-delete operations).

### DIFF
--- a/lib/shard/src/update.rs
+++ b/lib/shard/src/update.rs
@@ -338,7 +338,11 @@ fn upsert_with_payload(
 /// Max amount of points to delete in a batched deletion iteration
 const DELETION_BATCH_SIZE: usize = 512;
 
-/// Tries to delete points from all segments, returns number of actually deleted points
+/// Tries to delete points from all segments, returns number of actually deleted points.
+///
+/// Iterates all segments directly (rather than going through `apply_points`) to ensure
+/// that ALL copies of a point are deleted, including old non-deferred copies in optimized
+/// segments when the latest version is deferred in an appendable segment.
 pub fn delete_points(
     segments: &SegmentHolder,
     op_num: SeqNumberType,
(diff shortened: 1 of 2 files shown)


lib/shard/src/update.rs
```diff
@@ -348,12 +352,19 @@ pub fn delete_points(
...
+                if write_segment.delete_point(op_num, id, hw_counter)? {
+                    total_deleted_points += 1;
+                }
+            }
+        }
+    }
 
-        total_deleted_points += deleted_points;
+    if total_deleted_points == 0 {
+        segments.bump_max_segment_version_overwrite(op_num);
```
agourlay, line 367: This is new behavior change compared to the old code.

Why is it necessary?

IvanPleshkov: I stole this part from the deleted filtered implementation, where we have this check and guarantee version bump, even if nothing is deleted.
Because I have no idea if we need such a guarantee (I guess yes, but it's really wasnt present in dev), I decided to keep it here with potential deletion while review (better than skipping and missing it).

agourlay: I think it is required for the WAL replay.

Without this bump the operation would be replayed if it is the last one.

IvanPleshkov: The comment from the filtered deletion case:
```
// In case we didn't hit any points, we suggest this op_num to the segment-holder to make WAL acknowledge this operation.
// If we don't do this, startup might take up a lot of time in some scenarios because of recovering these no-op operations.
```
Reading this comment, I think we need such a version bump for regular deletion even if it was not added before

PR #8358, opened 2026-03-11
Default shard transfer with snapshot when prevent_unoptimized

agourlay:
Using stream record for shard with deferred points is not ideal as those deferred points are unlikely to be deferred on the target node.

For this reason we change the default transfer method for collection using `prevent_unindexed` to snapshot.

However the user can still pick stream records explicitly.

### DIFF
--- a/lib/collection/src/shards/transfer/driver.rs
+++ b/lib/collection/src/shards/transfer/driver.rs
@@ -168,7 +168,7 @@ pub async fn transfer_shard_fallback_default(
         .restart_shard_transfer_confirm_and_retry(&transfer_config, collection_id, fallback_method)
         .await?;
 
-    Ok(false)
+    Ok(true)
 }
 
 /// Return local shard back from the forward proxy


lib/collection/src/collection/shard_transfer.rs
```diff
@@ -49,10 +49,22 @@ impl Collection {
...
+                .unwrap_or(false);
+
+            let method = if prevent_unoptimized {
+                log::info!("Using snapshot transfer method because prevent_unoptimized is enabled");
+                ShardTransferMethod::Snapshot
+            } else {
+                self.shared_storage_config
+                    .default_shard_transfer_method
+                    .unwrap_or_default()
+            };
```
timvisee: Nit: I'd like a dedicated function such as `default_shard_transfer_method(prevent_unoptimized)` so that we clearly isolate the heuristic.

JojiiOfficial:
Do we also need to change it in other places, like [here](https://github.com/qdrant/qdrant/blob/b0fec5cf5b4b31f3f2b317cf0502b844b4f715b6/src/common/collections.rs#L426)?

agourlay:
Thanks for catching those usage, I will fix those later tonight!

agourlay:
CI should be green now, please review again :pray:

lib/collection/src/shards/transfer/driver.rs
```diff
@@ -164,10 +165,10 @@ pub async fn transfer_shard_fallback_default(
     // Propose to restart transfer with a different method
     transfer_config.method.replace(fallback_method);
     consensus
-        .restart_shard_transfer_confirm_and_retry(&transfer_config, collection_id)
+        .restart_shard_transfer_confirm_and_retry(&transfer_config, collection_id, fallback_method)
         .await?;
 
-    Ok(false)
+    Ok(true)
```
timvisee, line 171: Hah.

lib/collection/src/collection/shard_transfer.rs
```diff
@@ -35,6 +35,22 @@ impl Collection {
             .check_transfer_exists(transfer_key)
     }
 
+    pub async fn default_shard_transfer_method(&self) -> ShardTransferMethod {
+        let prevent_unoptimized = self
+            .effective_optimizers_config()
+            .await
+            .map(|config| config.prevent_unoptimized.unwrap_or(false))
+            .unwrap_or(false);
```
JojiiOfficial, line 43: Minor suggestion but also totally fine to keep as is
```suggestion
            .await
            .ok()
            .and_then(|i| i.prevent_unoptimized)
            .unwrap_or(false);
```

lib/collection/src/shards/transfer/driver.rs
agourlay: this one might make a few production clusters happy :crossed_fingers:

PR #8364, opened 2026-03-11
Ignore test_delete_by_filter_with_deferred_points

IvanPleshkov:
After merging test for deleted points in https://github.com/qdrant/qdrant/pull/8354 and after merging deferred points filtering, test `test_delete_by_filter_with_deferred_points` is broken because deleted points by filter are skipped because of deferred status.

The fix will appear here, where we add policy about deferred points: https://github.com/qdrant/qdrant/pull/8310

This PR just ignores test to make dev CI green


JojiiOfficial:
For reference, [this PR](https://github.com/qdrant/qdrant/pull/8310) already contains the fix: https://github.com/qdrant/qdrant/blob/4d85823b845faf0174fdd309b809e592410138fe/lib/shard/src/update.rs#L396

IvanPleshkov:
Close because https://github.com/qdrant/qdrant/pull/8364#:~:text=For%20reference%2C-,this%20PR,-already%20contains%20the has a fix already

PR #8366, opened 2026-03-11
Correct calculation of deferred point counts

JojiiOfficial:
Since we now exclude some points from read operations, we have to adjust the determination of point-counts in some places. Additionally we'd like to differentiate between visible and deferred points in telemetry / SegmentInfo.
This PR implements those changes.

<br>

Specifically this PR implements/changes the following:
  1. Correct calculation of available, non-deferred points
  2. Exclude deferred points from point count in some places where they are treated as hidden
  3. Adjust point counts in ProxySegment (including deletions from proxy segments)
  4. Revert `SegmentInfo` changes from previous PRs.
  5. Changes e2e tests to use telemetry instead of `CollectionInfo` to make them pass after change in 4.

### DIFF
--- a/lib/segment/src/segment/entry.rs
+++ b/lib/segment/src/segment/entry.rs
@@ -77,7 +77,7 @@ impl NonAppendableSegmentEntry for Segment {
             .get(vector_name)
             .ok_or_else(|| OperationError::vector_name_not_exists(vector_name))?;
         let vector_query_context =
-            query_context.get_vector_context(vector_name, self.deferred_internal_id);
+            query_context.get_vector_context(vector_name, self.deferred_internal_id());
         let internal_results = vector_data.vector_index.borrow().search(
             query_vectors,
             filter,
@@ -523,7 +523,7 @@ impl NonAppendableSegmentEntry for Segment {
             num_indexed_vectors,
             num_points: self.available_point_count(),
             num_deferred_points: Some(self.deferred_point_count()),
-            num_deleted_deferred_points: Some(self.deferred_deleted_count.unwrap_or_default()),
+            num_deleted_deferred_points: Some(self.deferred_deleted_count().unwrap_or_default()),
             num_deleted_vectors: self.deleted_point_count(),
             vectors_size_bytes,  // Considers vector storage, but not indices
             payloads_size_bytes, // Considers payload storage, but not indices
@@ -532,7 +532,7 @@ impl NonAppendableSegmentEntry for Segment {
             is_appendable: self.appendable_flag,
             index_schema: HashMap::new(),
             vector_data: vector_data_info,
-            deferred_internal_id: self.deferred_internal_id,
+            deferred_internal_id: self.deferred_internal_id(),
         }
     }
 
@@ -918,7 +918,7 @@ impl NonAppendableSegmentEntry for Segment {
     }
 
     fn point_is_deferred(&self, point_id: PointIdType) -> bool {
-        if let Some(deferred_from) = self.deferred_internal_id
+        if let Some(deferred_from) = self.deferred_internal_id()
             && let Some(internal_id) = self.id_tracker.borrow().internal_id(point_id)
         {
             return self.is_appendable() && internal_id >= deferred_from;
@@ -927,9 +927,13 @@ impl NonAppendableSegmentEntry for Segment {
     }
 
     fn deferred_point_ids(&self) -> Vec<PointIdType> {
-        let Some(deferred_from) = self.deferred_internal_id_if_any() else {
+        let Some(deferred_from) = self.deferred_internal_id() else {
             return vec![];
         };
+        if self.deferred_point_count() == 0 {
+            return vec![];
+        }
+
         let id_tracker = self.id_tracker.borrow();
         id_tracker
             .iter_internal()
@@ -946,9 +950,8 @@ impl NonAppendableSegmentEntry for Segment {
     }
 
     fn has_deferred_points(&self) -> bool {
-        self.deferred_internal_id.is_some_and(|deferred_from| {
-            self.is_appendable() && self.total_point_count() > deferred_from as usize
-        })
+        self.deferred_internal_id()
+            .is_some_and(|deferred_from| self.total_point_count() > deferred_from as usize)
     }
 }
 
--- a/lib/segment/src/segment/mod.rs
+++ b/lib/segment/src/segment/mod.rs
@@ -94,12 +94,18 @@ pub struct Segment {
     pub error_status: Option<SegmentFailedState>,
     #[cfg(feature = "rocksdb")]
     pub database: Option<Arc<parking_lot::RwLock<DB>>>,
+    pub(crate) deferred_point_status: Option<DeferredPointStatus>,
+}
+
+#[derive(Debug)]
+pub struct DeferredPointStatus {
     /// Points with internal id >= this value are hidden from reads.
     /// Available for appendable segments only.
-    pub(crate) deferred_internal_id: Option<PointOffsetType>,
+    pub(crate) deferred_internal_id: PointOffsetType,
+
     /// Amount of deleted deferred points. Must kept track of properly to be able
     /// to calculate the amount of available deferred and visible points.
-    pub(crate) deferred_deleted_count: Option<usize>,
+    pub(crate) deferred_deleted_count: usize,
 }
 
 pub struct VectorData {
--- a/lib/segment/src/segment/segment_ops.rs
+++ b/lib/segment/src/segment/segment_ops.rs
@@ -315,17 +315,15 @@ impl Segment {
 
         id_tracker.drop_internal(internal_id)?;
 
+        let deferred_point_status = self.deferred_point_status.as_mut();
+
         // Increase counter for deleted points.
-        if let Some(deferred_point) = self.deferred_internal_id
-            && internal_id >= deferred_point
+        if let Some(deferred_point_status) = deferred_point_status
+            && internal_id >= deferred_point_status.deferred_internal_id
             // Don't count the deletion of the same point twice
             && !is_point_already_deleted
         {
-            debug_assert!(
-                self.deferred_deleted_count.is_some(),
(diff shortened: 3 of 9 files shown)


lib/collection/src/shards/local_shard/scroll.rs
```diff
@@ -370,7 +370,7 @@ impl LocalShard {
                 let read_segment = get_segment.read();
 
                 (
-                    read_segment.available_point_count(),
+                    read_segment.available_point_count_without_deferred_estimated(),
```
JojiiOfficial: This change (and some other places) might need some further discussion because the point-count now is estimated.

For context, we currently can't reliably calculate the amount of deferred points. This is due to deleted points not being counted separately and we therefore can't distinguish between deleted deferred and visible points.

lib/edge/src/scroll.rs
```diff
@@ -261,7 +261,7 @@ impl EdgeShard {
                 let segment = segment.get();
                 let segment = segment.read();
 
-                let point_count = segment.available_point_count();
+                let point_count = segment.available_point_count_without_deferred_estimated();
```
JojiiOfficial: Same here

lib/segment/src/segment/entry.rs
```diff
@@ -936,6 +939,13 @@ impl NonAppendableSegmentEntry for Segment {
             .filter_map(|internal_id| id_tracker.external_id(internal_id))
             .collect()
     }
+
+    fn available_point_count_without_deferred_estimated(&self) -> usize {
+        self.id_tracker
+            .borrow()
+            .available_point_count()
+            .saturating_sub(self.deferred_point_count_estimated())
+    }
```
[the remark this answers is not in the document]
JojiiOfficial: Yeah this is the concern with this PR. To deal with that, I'm working on a proper implementation which keeps track of the _exact_ values for each visible and deferred points: https://github.com/qdrant/qdrant/pull/8386/changes

lib/segment/src/segment/entry.rs
```diff
@@ -523,7 +521,9 @@ impl NonAppendableSegmentEntry for Segment {
             segment_type: self.segment_type,
             num_vectors,
             num_indexed_vectors,
-            num_points: self.non_deferred_point_count_estimated(),
+            num_points: self.available_point_count(),
```
JojiiOfficial, line 524: Reverting this change as discussed.
To get the amount of available points without deferred, one can subtract `num_deferred_points` from `num_points`. We can also add a field for that later if desired.

lib/edge/src/scroll.rs
JojiiOfficial: Edit: Not an estimation anymore with latest changes

lib/collection/src/shards/local_shard/scroll.rs
JojiiOfficial: Edit: Not an estimation anymore with latest changes

tests/openapi/test_deferred_points.py
```diff
@@ -55,6 +55,36 @@ def get_collection_info():
     return response.json()['result']
 
 
+def get_point_count_excluding_deferred():
```
JojiiOfficial, line 58: Since this PR reverts previous changes to `SegmentInfo`, we have to use the new fields from telemetry for now.

lib/segment/src/segment/mod.rs
```diff
@@ -97,6 +97,9 @@ pub struct Segment {
     /// Points with internal id >= this value are hidden from reads.
     /// Available for appendable segments only.
     pub(crate) deferred_internal_id: Option<PointOffsetType>,
+    /// Amount of deleted deferred points. Must kept track of properly to be able
+    /// to calculate the amount of available deferred and visible points.
+    pub(crate) deferred_deleted_count: Option<usize>,
```
JojiiOfficial: This is the most important change in this PR.

We increase a Segment-level counter for each deletion of a _deferred_ point. This way we can correctly calculate how much of the deleted points are deferred. This then allows us to correctly calculate the amount of non-deleted deferred and visible (non-deferred) points.

For correctness, we iterate over all deferred points once at segment load and count how many deferred points are deleted. With the result of this value we initialize the segments `deferred_deleted_count`.

lib/collection/src/update_workers/update_worker.rs
```diff
@@ -205,7 +205,7 @@ impl UpdateWorkers {
                     let segments = locked_segments.read();
                     segments.iter().any(|(_, segment)| {
                         let segment_guard = segment.get().read();
-                        segment_guard.deferred_points_count() > 0
+                        segment_guard.has_deferred_points()
                     })
```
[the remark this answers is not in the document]
JojiiOfficial: We always optimize a segment if it has deferred points (even if they're all deleted).
Additionally, we do this wait after an update operation which inserts new (deferred) points.
This means if we reach this check, we always have non-deleted deferred points (namely the ones we just upserted).

docs/redoc/master/openapi.json
```diff
@@ -12749,6 +12749,8 @@
           "disk_usage_bytes",
           "index_schema",
           "is_appendable",
+          "num_deferred_points",
+          "num_deleted_deferred_points",
```
generall: please make those parameters optional. Otherwise 1.18 client will fail to read 1.17 server output

lib/segment/src/segment/segment_ops.rs
```diff
@@ -653,27 +670,37 @@ impl Segment {
...
+    pub(crate) fn calculate_deleted_deferred_point_count(&self) -> usize {
+        let Some(deferred_from) = self.deferred_internal_id else {
+            return 0;
+        };
+
+        let id_tracker = self.id_tracker.borrow();
+
+        (deferred_from..id_tracker.total_point_count() as u32)
+            .filter(|&internal_id| id_tracker.external_id(internal_id).is_none())
+            .count()
```
generall:
```suggestion
        let id_tracker = self.id_tracker.borrow();
        let total_points = id_tracker.total_point_count();
        id_tracker.deleted_point_bitslice()[deferred_from as usize..total_points].count_ones()
```

agourlay:
Please rebase.

I think I am ready to approve :)

lib/segment/src/segment/segment_ops.rs
```diff
@@ -309,7 +309,24 @@ impl Segment {
             .borrow_mut()
             .clear_payload(internal_id, hw_counter)?;
 
-        self.id_tracker.borrow_mut().drop_internal(internal_id)?;
+        let mut id_tracker = self.id_tracker.borrow_mut();
+
+        let is_point_already_deleted = id_tracker.is_deleted_point(internal_id);
+
+        id_tracker.drop_internal(internal_id)?;
```
IvanPleshkov, line 316: We may upgrade `drop_internal` and return if the ID was deleted. Not for now, maybe in the future

lib/segment/src/segment/mod.rs
IvanPleshkov: Because `Some` state is shared between `deferred_internal_id` and `deferred_deleted_count` (when one is some, a second one should be also some), I propose to create a struct like
```
pub struct DeferredPointStatus {
deferred_internal_id: Option<PointOffsetType>,
deferred_deleted_count: Option<usize>,
}
```

And use in segment:
```
deferred_points: Option<DeferredPointStatus >
```

We reduce the number of unnecessary checks/asserts in the PR

lib/segment/src/segment/segment_ops.rs
```diff
@@ -653,27 +670,40 @@ impl Segment {
...
+
+        if total_points < deferred_from as usize {
+            return 0;
+        }
+
+        id_tracker.deleted_point_bitslice()[deferred_from as usize..total_points].count_ones()
+    }
+
+    /// Returns the `deferred_internal_id` of the segment if there is at least one non-deleted deferred point in the segment.
+    pub(super) fn deferred_internal_id_if_any(&self) -> Option<PointOffsetType> {
```
IvanPleshkov: We have `has_deferred` method. We need to think do we need this check in `has_deferred` (it's a condition for optimizers and back pressure). I have no idea for now, but I think to let `has_deferred` unchanged.
Anyway I propose to remove this `deferred_internal_id_if_any`, we use it in one place only now and functions name looks complicated

lib/segment/src/segment/entry.rs
```diff
@@ -937,6 +937,19 @@ impl NonAppendableSegmentEntry for Segment {
...
+    fn available_point_count_without_deferred(&self) -> usize {
+        self.id_tracker
+            .borrow()
+            .available_point_count()
+            .saturating_sub(self.deferred_point_count())
+    }
+
+    fn has_deferred_points(&self) -> bool {
+        self.deferred_internal_id.is_some_and(|deferred_from| {
+            self.is_appendable() && self.total_point_count() > deferred_from as usize
```
IvanPleshkov: `self.is_appendable()` is not required anymore. For non-appendable segments, we don't set `Some` for deferred_internal_id.

lib/shard/src/proxy_segment/mod.rs
```diff
@@ -32,6 +32,7 @@ pub struct ProxySegment {
     /// May contain points which are not in wrapped_segment,
     /// because the set is shared among all proxy segments
     deleted_points: DeletedPoints,
+    deleted_deferred_count: usize,
```
IvanPleshkov, line 35: As I understood, we update `deleted_deferred_count` in the same way as the counter of deferred deletions in the regular segment.
Why do we need to have `deleted_deferred_count` here instead of getting the deleted deferred points count from the wrapped segment?

JojiiOfficial: Because ProxySegment has its own deleted-point counter allowing us to delete points while having a non-mutable reference to the wrapped segment.
Therefore we need to count here as well

IvanPleshkov: You are right, my mistake

PR #8367, opened 2026-03-11
fix: roll back .snapshot when checksum keep() fails in LocalFS store_file

RAJESHREDDY0508:
## Problem

`SnapshotStorageLocalFS::store_file` had a partial-failure window between
the two persistence steps:

1. `target_path_tmp.persist(target_path)` — atomically renames `.tmp → .snapshot`
2. `checksum_file.keep()` — promotes the `.snapshot.checksum` TempPath to permanent

If step 2 failed (e.g. disk full, I/O error), the `.snapshot` file was
already written to its final location with no TempPath guard, but the
`.snapshot.checksum` file was deleted by its TempPath destructor on function
exit.  This left a `.snapshot` on disk permanently with no matching checksum
— an inconsistent state that could not be recovered automatically.

On resource-constrained nodes this also leaked disk space, because the
orphaned snapshot was invisible to the normal cleanup path.

## Fix

In `SnapshotStorageLocalFS::store_file`
(`lib/collection/src/common/snapshots_manager.rs`):

After `persist()` succeeds, call `checksum_file.keep()` explicitly and, on
failure, delete the already-persisted `.snapshot` before returning the error.
This enforces the invariant: **both files are present, or neither is.**

```rust
// before
target_path_tmp.persist(target_path).map_err(|e| e.error)?;
checksum_file.keep()?;   // ← was implicit / missing

// after
target_path_tmp.persist(target_path).map_err(|e| e.error)?;

if let Err(err) = checksum_file.keep() {
    let _ = tokio_fs::remove_file(target_path).await;  // rollback snapshot
    return Err(err.into());
}


anaslimem:
Hey, the rollback inside store_file is a useful guard, but this alone is insufficient to fix the bug:

1. **The caller-side `temp_paths` are still dropped without `.keep()`**  on the happy path, files are still silently deleted when the `Vec<TempPath>` drops at end of scope.
2. **No cross-component rollback**  if `keep()` on the snapshot succeeds but fails on the checksum, the caller needs a keep-with-rollback loop; the internal fix doesn't cover this.
3. **The post-persist re-wrap** of the already-moved snapshot into a new `TempPath` in ShardHolder is also not addressed.

RAJESHREDDY0508:
Thanks for the detailed review!

You're right that the rollback added in `store_file` only protects the local persistence window and doesn't fully address the caller-side `TempPath` lifecycle or the cross-component rollback behavior.

I'm currently reviewing the caller paths and the `ShardHolder` logic you mentioned and update the PR to ensure:

• `TempPath::keep()` is explicitly handled on the happy path
• rollback logic covers partial `keep()` failures across snapshot and checksum
• the post-persist re-wrapping in `ShardHolder` does not risk accidental deletion

I'll push an updated patch shortly. Appreciate the guidance!

test-output.txt
timvisee, line 1: Can we drop this file?

anaslimem:
I have noticed that the PR is for master branch, you need to change it to dev

RAJESHREDDY0508:
Maintainers requested that the PR target the dev branch.

I’ve recreated the PR based on dev here: https://github.com/qdrant/qdrant/pull/8379

Closing this one to avoid confusion.

PR #8368, opened 2026-03-11
impl `UniversalRead` for local disk cache

coszio:
Takes #8109, but removes the WIP integration. Instead it implements `UniversalRead` for it.

### TODO
- [x] Make `CachedSlice` impl `UniversalRead`
- [x] Fix alignment errors
- [x] Make sure it doesn't blow up when the main cache file is saturated

## Future work
- [ ] Populate impl -> #8462
- [ ] Clear RAM cache impl
- [ ] Block pinning (to not evict while reference is active)
- [ ] Be able to return cache miss read before writing to main cache file.
- [ ] Use generic backend for slow disk reads -> #8467 
- [ ] Use a more efficient structure for tracking unused blocks


lib/common/common/src/universal_io/disk_cache/cached_slice.rs
```diff
@@ -0,0 +1,215 @@
...
+                Cow::Borrowed(bytes) => Cow::Borrowed(bytemuck::cast_slice(bytes)),
+                Cow::Owned(vec_u8) => {
+                    let mut vec_t = vec![T::zeroed(); vec_u8.len() / t_size];
+                    bytemuck::cast_slice_mut::<T, u8>(&mut vec_t).copy_from_slice(&vec_u8);
+                    Cow::Owned(vec_t)
+                }
+            };
+
+            return Ok(cow);
+        }
```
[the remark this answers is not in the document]
coszio: Sure, it is possible, but for the use-case, should be very rare. Two reasons:

1. cache disk should fit most of the dataset, so eviction policy is only enforced if there's frequent cache misses.

2. by the time the just-read block is to be evicted, at least `capacity` misses have been taken place. So the problem only occurs if we hold the reference as part of some structure, which isn't the case.


Will document as future work, though

lib/common/common/src/universal_io/disk_cache/controller.rs
```diff
@@ -0,0 +1,297 @@
...
+                        .as_mut_ptr()
+                        .add(offset)
+                        .copy_from(buf.as_ptr(), BLOCK_SIZE);
+                }
+
+                // 3. Commit.
+                // ----------
+
+                // FIXME: unwrap panics when `key` deleted while guard is still alive.
+                guard.insert(allocated_offset).unwrap();
```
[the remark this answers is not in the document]
coszio: There's no key deletions available in current interface

lib/common/common/src/universal_io/disk_cache/tests.rs
```diff
@@ -0,0 +1,259 @@
...
+        CacheController::new(&dir.path().join("cache.bin"), BLOCK_SIZE as u64 * 5).unwrap();
+
+    create_test_file(&dir.path().join("cold.0"), "cold.0", 10);
+
+    let fd = CachedSlice::<u8>::open(&cacher, &dir.path().join("cold.0")).unwrap();
+
+    eprintln!(
+        "data0: {}",
+        fancy_decode(&fd.get_range(BLOCK_SIZE * 3..BLOCK_SIZE * 3).unwrap())
+    );
```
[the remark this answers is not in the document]
coszio: It was intentional

lib/common/common/src/disk_cache/controller.rs
```diff
@@ -0,0 +1,327 @@
...
+use parking_lot::{Condvar, Mutex, RwLock};
+use quick_cache::UnitWeighter;
+use quick_cache::sync::GuardResult;
+
+use super::{BLOCK_SIZE, BlockId, BlockOffset, BlockRequest, FileId};
+use crate::fs::clear_disk_cache;
+
+const UNUSED_BLOCKS_MARGIN: u64 = 16;
+
+/// Caching layer stub.
```
generall: update comment?

lib/common/common/src/universal_io/disk_cache/controller.rs
```diff
@@ -0,0 +1,327 @@
...
+/// Once we implement it, it would be used to read static files from slow
+/// network-attached disks while caching recently accessed file pages on fast
+/// local SSD.
+///
+/// But for now the current implementation is a stub that provides rough
+/// public interface, but none of desired performance characteristics.
+#[derive(Debug)]
+pub struct CacheController {
+    /// Mapping from the assigned file id, to its file descriptor
+    files: RwLock<HashMap<FileId, fs::File>>,
```
generall, line 28: ToDo: map to universal IO reference?

lib/common/common/src/universal_io/disk_cache/controller.rs
```diff
@@ -0,0 +1,286 @@
...
+        opts.custom_flags(nix::libc::O_DIRECT);
+        let f = opts.open(path)?;
+        #[cfg(target_os = "macos")]
+        unsafe {
+            // TODO: add this to `nix` crate
+            use std::os::fd::AsRawFd;
+            nix::libc::fcntl(f.as_raw_fd(), nix::libc::F_NOCACHE, 1);
+        }
+
+        let len = f.metadata()?.len() as usize;
```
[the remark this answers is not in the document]
generall: we don't care about 32-bit systems

lib/common/common/src/disk_cache/controller.rs
```diff
@@ -0,0 +1,327 @@
...
+///
+/// But for now the current implementation is a stub that provides rough
+/// public interface, but none of desired performance characteristics.
+#[derive(Debug)]
+pub struct CacheController {
+    /// Mapping from the assigned file id, to its file descriptor
+    files: RwLock<HashMap<FileId, fs::File>>,
+
+    /// Used to assign file ids on new files.
+    file_id_counter: Mutex<FileId>,
```
generall: atomic?

lib/common/common/src/universal_io/disk_cache/controller.rs
```diff
@@ -0,0 +1,327 @@
...
+
+        match self.cache.get_value_or_guard(&key, None) {
+            // Cache hit.
+            GuardResult::Value(block_offset) => {
+                // Read from hot mmap.
+                let range = block_offset.bytes() + range.start..block_offset.bytes() + range.end;
+
+                // TODO: Pin the block_id, track the lifetime of the borrow,
+                // then release the block_id when the borrow ends
+                let slice = unsafe {
```
generall, line 156: consider callback to avoid data-race problem?

lib/common/common/src/universal_io/disk_cached.rs
```diff
@@ -0,0 +1,93 @@
...
+        Ok(self.get_range(range)?)
+    }
+
+    fn read_batch<const SEQUENTIAL: bool>(
+        &self,
+        ranges: impl IntoIterator<Item = crate::universal_io::ElementsRange>,
+        mut callback: impl FnMut(usize, &[T]) -> crate::universal_io::Result<()>,
+    ) -> crate::universal_io::Result<()> {
+        for (i, range) in ranges.into_iter().enumerate() {
+            let data = self.read::<SEQUENTIAL>(range)?;
```
generall: we want `self.get_range_batch` for this case, otherwise cold latency will be terrible

lib/common/common/src/universal_io/disk_cache/controller.rs
coszio: Using a callback only makes sense if we have this callback usage all the way into interface, but we will lose the ability to return a slice when possible (e.g. for non-batched interface).

lib/common/common/src/universal_io/disk_cached.rs
coszio: I propose to do this once we add the support for io-uring for slow disk

lib/common/common/src/universal_io/disk_cache/cached_slice.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// Get a Cow reference to a range of elements within the file.
+    ///
+    /// The `range` is in **elements of T**, not bytes. For `T = u8` this is
+    /// equivalent to a byte range.
+    ///
+    /// If the range is contained in a single block, it will return a borrowed
+    /// reference into the mmap. Otherwise, it will allocate a `Vec<T>` and copy
+    /// block data into it. Allocating as `Vec<T>` (rather than `Vec<u8>`)
+    /// guarantees correct alignment for any `T`.
+    pub fn get_range(&self, range: Range<usize>) -> io::Result<Cow<'_, [T]>> {
```
xzfc, line 50: I think later we'd need to replace `Range<usize>` with `ReadRange`. Maybe after #8493.

lib/common/common/src/lib.rs
```diff
@@ -11,6 +11,8 @@ pub mod cpu_utilization;
 pub mod defaults;
 pub mod delta_pack;
 pub mod disk;
+#[cfg(not(target_os = "windows"))]
+pub mod disk_cache;
```
xzfc: I suggest to move `common::disk_cache` into `common::universal_io::disk_cache`.

Like this: aa8afe49264039b1ab0b1f1a9285a66b9a7a0063 (not pushing this commit for now to avoid merge conflicts in #8462).

coszio: Good suggestion, will move it now, since I need to resolve conflicts anyway

lib/common/common/src/universal_io/disk_cache/cached_slice.rs
coszio: Yes, and will probably need a rename to `DiskCached` or smth

PR #8373, opened 2026-03-12
Fix transfers may lock shard holder for too long

timvisee:
Shard transfers may hold a shard holder lock for a very long time, which can cascade into blocking many operations and eventually freezing the whole cluster.

Some side effects we've seen in real clusters include unresponsive nodes, broken consensus and downtime.

Since we've implemented `prevent_unoptimized` and the update queue this is a lot more likely to happen.

For example, the last batch of a stream records transfer may block here for a very long time because the last batch sends with `wait=true`, which holds a shard holder read lock for all that time: https://github.com/qdrant/qdrant/blob/f91e2179d521978a1bec657a4f0767f80cd293d1/lib/collection/src/shards/transfer/stream_records.rs#L136-L163

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/forward_proxy_shard.rs
+++ b/lib/collection/src/shards/forward_proxy_shard.rs
@@ -20,7 +20,7 @@ use shard::scroll::ScrollRequestInternal;
 use shard::search::CoreSearchRequestBatch;
 use shard::snapshots::snapshot_manifest::SnapshotManifest;
 use tokio::runtime::Handle;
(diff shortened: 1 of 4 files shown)


timvisee:
Fix opened in <https://github.com/qdrant/qdrant/pull/8374>

timvisee:
@agourlay @generall Same as <https://github.com/qdrant/qdrant/pull/8374>, but including the test.

PR #8374, opened 2026-03-12
Don't lock shard holder for a long time in stream records transfer

timvisee:
Fix <https://github.com/qdrant/qdrant/pull/8373>

Splits transferring batches in stream records transfers into two. This ensures the shard holder read lock is only held for a very short time.

Stages:
1. read - read the batch (quick, and holds the lock)
2. write - send batch to remote and wait (takes long, but does not hold the lock)

The changes are a little bit invasive, but I don't see a nicer way of handling this problem.

### All Submissions:

### New Feature Submissions:

5. [x] Have you checked your code using `cargo clippy --workspace --all-features` command?


lib/collection/src/shards/transfer/stream_records.rs
```diff
@@ -134,19 +134,32 @@ pub(super) async fn transfer_stream_records(
...
         };
 
-        let result = replica_set
-            .transfer_batch(offset, TRANSFER_BATCH_SIZE, None, merge_points)
-            .await?;
+        // Send batch to remote shard without holding the shard holder lock.
+        // This is important because sending can take a very long time (especially the last
+        // batch which waits for the remote to fully process). Holding the shard holder lock
+        // during this time would block all other operations on the collection.
+        let result = prepared_batch.send(&remote_shard).await?;
```
timvisee, line 167: Critical section is here.

First we get a `prepared_batch` which is quick.

Then we transfer the batch and wait for application. This can take a very long time. And now we don't hold the shard holder lock anymore.

PR #8376, opened 2026-03-12
Use dedup() instead of unique()

JojiiOfficial:
Since `iter` is ordered here, we can use [dedup()](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.dedup) instead of [unique()](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique), which does unnecessary allocations.
dedup() prevents these, resulting in better performance and less allocations.

### DIFF
--- a/lib/segment/src/segment/facet.rs
+++ b/lib/segment/src/segment/facet.rs
@@ -85,6 +85,16 @@ impl Segment {
                     .iter_values_map(hw_counter)
                     .stop_if(is_stopped)
                     .filter_map(|(value, iter)| {
+                        #[cfg(debug_assertions)]
+                        let iter = {
+                            let mut prev_id = None;
+                            iter.inspect(move |&id| {
+                                let previous = prev_id.get_or_insert(id);
+                                debug_assert!(*previous <= id, "Sorted iter assertion broken");
+                                *previous = id;
+                            })
+                        };
+
                         let count = iter
                             .dedup()
                             .take_while(|&point_id| {


lib/segment/src/segment/facet.rs
```diff
@@ -86,7 +86,7 @@ impl Segment {
                     .stop_if(is_stopped)
                     .filter_map(|(value, iter)| {
                         let count = iter
-                            .unique()
+                            .dedup()
```
timvisee, line 99: Does it make sense to assert it is sorted in debug builds?

```suggestion
                        #[cfg(debug_assertions)]
                        let mut point_ids = AHashSet::new();
                        let count = iter
                            .dedup()
                            .inspect(|point_id| {
                                debug_assert!(
                                    point_ids.insert(*point_id),
                                    "Facet index iterator must be sorted",
                                );
                            })

```

Not beautiful, but it works.

JojiiOfficial: What do you think about this approach: https://github.com/qdrant/qdrant/pull/8376/changes/68d69ae4a42a2065fa6f31b1dd742a39f46e1c2c?

lib/segment/src/segment/facet.rs
```diff
@@ -85,8 +85,17 @@ impl Segment {
...
                     .stop_if(is_stopped)
                     .filter_map(|(value, iter)| {
+                        #[cfg(debug_assertions)]
+                        let iter = {
+                            let mut prev_id = None;
+                            iter.inspect(move |&id| {
+                                let previous = prev_id.get_or_insert(id);
+                                debug_assert!(*previous <= id, "Sorted iter assertion broken");
+                            })
+                        };
```
timvisee, line 96: Better. Thanks!

PR #8378, opened 2026-03-12
chore: `DiscoveryQuery` → `DiscoverQuery`

xzfc:
I discover~~y~~ed that we incorrectly call it `DiscoveryQuery` (it's `discover`/`DiscoverQuery` in the REST API, python client and python edge). This PR removes `y` in these and related places.

Not changed: `DiscoveryQuery` in GRPC API, and in places where the word "discovery" is correctly used as noun.
<details><summary><code>rg -i discovery  # not changed</code></summary>

```
README.md
104:### Explore Similar Image Search - Food Discovery 🍕
106:There's more to discovery than text search, especially when it comes to food. People often choose meals based on appearance rather than descriptions and ingredients. Let Qdrant help your users find their next delicious meal using visual search, even if they don't know the dish's name. [Check it out!](https://qdrant.to/food-discovery)

docs/redoc/master/openapi.json
15177:            "description": "Pairs of { positive, negative } examples to constrain the search.\n\nWhen using only the context (without a target), a special search - called context search - is performed where pairs of points are used to generate a loss that guides the search towards the zone where most positive examples overlap. This means that the score minimizes the scenario of finding a point closer to a negative than to a positive part of a pair.\n\nSince the score of a context relates to loss, the maximum score a point can get is 0.0, and it becomes normal that many points can have a score of 0.0.\n\nFor discovery search (when including a target), the context part of the score for each pair is calculated +1 if the point is closer to a positive than to a negative part of a pair, and -1 otherwise.",

docs/roadmap/README.md
37:* 🔭 Search and Discovery

docs/roadmap/roadmap-2023.md
48:  * [x] Discovery search - constrain the space in which the search is performed

docs/roadmap/roadmap-2024.md
68:  * [x] Improvements in Discovery API to support more use-cases

lib/api/build.rs
386:            ("DiscoveryQuery.target", ""),
387:            ("DiscoveryQuery.context", ""),

lib/api/src/grpc/conversions.rs
2878:impl From<segment_query::DiscoverQuery<segment_vectors::VectorInternal>> for raw_query::Discovery {
2891:impl TryFrom<raw_query::Discovery>
2895:    fn try_from(value: raw_query::Discovery) -> Result<Self, Self::Error> {
2896:        let raw_query::Discovery { target, context } = value;

lib/api/src/grpc/proto/points_internal_service.proto
185:message DiscoveryQuery {
202:    DiscoveryQuery discover = 3;
274:  message Discovery {
302:    Discovery discover = 3;

lib/api/src/grpc/qdrant.rs
10577:pub struct DiscoveryQuery {
10616:        Discover(super::DiscoveryQuery),
10778:    pub struct Discovery {
10825:        Discover(Discovery),

lib/collection/src/lib.rs
6:pub mod discovery;

lib/collection/src/operations/types.rs
804:    /// For discovery search (when including a target), the context part of the score for each pair

lib/edge/python/qdrant_edge.pyi
2067:    """Query for discovery using a target and context pairs."""

lib/segment/src/index/hnsw_index/hnsw.rs
1302:        const DISCOVERY_ENTRY_POINT_COUNT: usize = 10;
1308:                DISCOVERY_ENTRY_POINT_COUNT,

lib/segment/src/vector_storage/query/context_query.rs
34:    /// In the first stage of discovery search, the objective is to get the best entry point

lib/shard/src/query/conversions.rs
439:                query: Some(grpc::query_enum::Query::Discover(grpc::DiscoveryQuery {
578:                Variant::Discover(grpc::raw_query::Discovery::from(named.query))

lib/storage/src/content_manager/toc/point_ops.rs
16:use collection::{discovery, recommendations};
275:        discovery::discover(
308:        discovery::discover_batch(
```

</discovery>


generall:
It seems, that it only affects internal gRPC, therefore we shouldn't have any problems renaming it as well.
On another another hand, those internal APIs should be removed after we drop all deprecated search endpoints, so I wouldn't care too much.

PR #8379, opened 2026-03-12
Fix snapshot temp path leak dev

RAJESHREDDY0508:
Fixes snapshot file leaks and partial persistence introduced by unguarded
TempPath drops across three call sites.

### Changes

**`collection/snapshots.rs`**
Call `keep()` on `snapshot_temp_arc_file` after a successful `store_file`.
Previously the `NamedTempFile` destructor silently deleted the already-moved archive.

**`transfer/snapshot.rs`**
Replace `Vec<TempPath>` (implicit drop = delete) with an explicit `Vec<PathBuf>` cleanup loop:
- Cleanup runs on both success and failure
- `recover_result` captured before cleanup so errors are not overridden
- `NotFound` suppressed so double-delete is silent
- No `TempPath::from_path` re-wrap of already-persisted paths

**`shard_holder/mod.rs`**
Clarifying comment on existing `keep()` guard.

### Tests
Four unit tests added in `transfer/snapshot.rs`: happy path, failed recovery, double-delete, and streaming no-op.

### Known gap
If `get_shard_snapshot_path` fails after `create_shard_snapshot` has already written permanent files, those files are not cleaned up (pre-existing, marked `TODO`). Left as a follow-up.


lib/collection/src/common/snapshots_manager.rs
```diff
@@ -299,7 +299,14 @@ impl SnapshotStorageLocalFS {
...
         target_path_tmp.persist(target_path).map_err(|e| e.error)?;
 
-        checksum_file.keep()?;
+        // If keeping the checksum file as permanent fails, the .snapshot file is
+        // already at its permanent location with no TempPath guard.  Roll it back
+        // so we never leave a .snapshot on disk without a matching .snapshot.checksum.
+        if let Err(err) = checksum_file.keep() {
+            let _ = tokio_fs::remove_file(target_path).await;
+            return Err(err.into());
+        }
```
timvisee, line 308: I don't understand this one. Why would keep fail here? Could you elaborate?

PR #8381, opened 2026-03-12
Edge flat api

xzfc:
Compare: [Badge](https://docs.rs/qdrant-edge/0.0.0/qdrant_edge/) [Badge](https://xzfc.github.io/qdrant/doc/qdrant_edge/)

This PR rehauls `qdrant-edge` package public API. Previously it was exporting `segment` and `shard` crates as is, exposing a deeply nested tree of irrelevant items. In this PR, almost everything is exposed as a flat namespace. (see Fig. 1)

Details:
- Fig. 1 The new structure of qdrant-edge packageAlmost all our internal structures still exported as-is (but some with renames like `CountRequestInternal` → `CountRequest`). Probably they expose some internal methods that shouldn't be in the public API, but we decided to keep them as is for now, and provide convenient wrappers later.
- Points and vectors have convenient wrappers that mirror Rust qdrant-client API. (see `point.rs` and `vectors.rs`).
- Unrelated items are hidden.
- We still export `qdrant_edge::internal::{SnapshotManifest, clear_data, move_data}` because these are used in the [restore-snapshot.rs](https://github.com/qdrant/qdrant/blob/dev/lib/edge/publish/examples/src/bin/restore-snapshot.rs#L101-L110) example. Later we should remove this usage and remove these exports.
- Public dependencies are exposed as `qdrant_edge::external::{ordered_float, serde_json, uuid}`, so users don't need to add them to their `Cargo.toml` to be able to call `qdrant-edge` methods.

### DIFF
--- a/lib/edge/Cargo.toml
+++ b/lib/edge/Cargo.toml
@@ -26,6 +26,7 @@ ordered-float = { workspace = true }
 parking_lot = { workspace = true }
 rand = { workspace = true }
 serde = { workspace = true }
+serde_json = { workspace = true }
 uuid = { workspace = true }
 
 [dev-dependencies]
--- a/lib/edge/publish/amalgamate.py
+++ b/lib/edge/publish/amalgamate.py
@@ -127,6 +127,7 @@ def main() -> None:
             """
             #![allow(unexpected_cfgs)]
             #![allow(dead_code, unused_imports)]
+            // #![warn(unnameable_types)] // TODO: re-enable when cleaning up the API
             pub use edge::*;
             """
         ).lstrip()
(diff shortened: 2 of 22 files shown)


generall:
> Public dependencies are exposed as qdrant_edge::external::{ordered_float, serde_json, uuid}, so users don't need to add them to their Cargo.toml to be able to call qdrant-edge methods.

But does it mean, that if I still want to use UUID from direct dependency, it will still work, right?

xzfc:
Yes.

timvisee:
I see the Rust crate exposes a huge flat list of structures. Is the idea to roughly match this? https://docs.rs/qdrant-client/latest/qdrant_client/qdrant/index.html

generall (approved):

xzfc:
> Is the idea to roughly match this? docs.rs/qdrant-client/latest/qdrant_client/qdrant/index.html

Almost. As first priority, to roughly match `qdrant-edge-py`; as second priority to roughly match Python/Rust clients.

PR #8394, opened 2026-03-13
Add internal wait enum, don't let forward proxy wait on optimizers

timvisee:
Since merging <https://github.com/qdrant/qdrant/pull/8239> `wait=true` may wait for a full optimization cycle if deferred points are used.

A forward proxy shard, used during transfers, currently forces `wait=true` on update operations. A side effect of this is that some user operations may be blocked for a very long time even if `wait=false` was provided. We consider this to be a bug.

This PR adds a `WaitUntil` enum allowing a third wait variant in between the existing wait states. Rather than acknowledging right away or waiting on optimization, this falls in between. It just waits until the operations is applied in segments, similar to how `wait=true` functioned in previous versions. It fixes blocking user operations for a very long time.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/shards/forward_proxy_shard.rs
+++ b/lib/collection/src/shards/forward_proxy_shard.rs
@@ -488,7 +488,7 @@ impl ShardOperation for ForwardProxyShard {
         // Shard update is within a write lock scope, because we need a way to block the shard updates
         // during the transfer restart and finalization.
 
-        // We always have to wait for the operation to be written to segments, cause after we
+        // We always have to wait for the operation to be written to segments, because after we
         // release the lock, the transfer needs to have access to the latest version of points.
         // Note that we wait on `Segment` and not on `Visible`, because with the latter we'd also
         // wait on deferred points to be fully optimized which is not necessary for transfers.
--- a/lib/collection/src/shards/shard_trait.rs
+++ b/lib/collection/src/shards/shard_trait.rs
@@ -18,7 +18,7 @@ use crate::operations::universal_query::shard_query::{ShardQueryRequest, ShardQu
 
 /// Controls how an update operation waits for completion.
 ///
-/// Internal enum usually deferred from wait true/false as specified by a user request.
+/// Internal enum derived from `wait=true/false` as specified by a user request.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum WaitUntil {
     /// Wait until the operation is written in WAL.


lib/collection/src/shards/shard_trait.rs
```diff
@@ -16,12 +16,60 @@ use crate::operations::OperationWithClockTag;
...
+    /// Wait until the operation is written in a segment.
+    ///
+    /// This does not mean the operation will be visible. If deferred points are enabled it may
+    /// take some time for the change to appear.
+    Segment,
+    /// Wait until the operation is visible in search results.
+    ///
+    /// Corresponds with wait=true, acknowledging only when the change is fully applied and visible.
+    Visible,
+}
```
timvisee, line 37: I settled on:
- `WaitUntil::Wal` - same as wait=false, only wait until written in WAL
- `WaitUntil::Segment` - wait until written in segment, but not on optimizers
- `WaitUntil::Visible` - wait until point is visible (written in segment and optimized in case of deferred point)

lib/collection/src/update_handler.rs
```diff
@@ -41,6 +41,9 @@ pub struct OperationData {
     pub operation: Option<Box<CollectionUpdateOperations>>,
     /// Callback notification channel
     pub sender: Option<oneshot::Sender<CollectionResult<usize>>>,
+    /// Whether to wait for deferred points to be optimized before sending feedback.
+    /// Only relevant when `sender` is `Some`.
+    pub wait_for_deferred: bool,
```
timvisee, line 46: This makes the `UpdateSignal` 8 bytes larger, from 64 to 72 bytes. We store up to a million of these in memory with the update queue. So we might want to look into avoiding this.

lib/collection/src/tests/deferred_points_tests.rs
```diff
@@ -156,7 +158,9 @@ async fn test_deferred_points_wait_false() {
     // Push all points with wait=false — returns immediately without waiting for application
     for i in 1..=NUM_POINTS {
         let op = make_upsert_op(i);
-        let result = shard.update(op.into(), false, None, hw_acc.clone()).await;
+        let result = shard
+            .update(op.into(), WaitUntil::Wal, None, hw_acc.clone())
+            .await;
```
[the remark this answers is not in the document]
IvanPleshkov: I find it better to keep `Wal` because the chance of flakiness is less (we have less time until optimizations are done in case of WaitUntil::Segment)

lib/collection/src/shards/forward_proxy_shard.rs
```diff
@@ -488,11 +488,18 @@ impl ShardOperation for ForwardProxyShard {
...
         let mut result = self
             .wrapped_shard
-            .update(operation.clone(), true, timeout, hw_measurement_acc.clone())
+            .update(
+                operation.clone(),
+                WaitUntil::Segment,
+                timeout,
+                hw_measurement_acc.clone(),
+            )
             .await?;
```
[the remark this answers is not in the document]
IvanPleshkov: We're gonna do a manual test for now. These changes are a blocker performing high-load tests

lib/collection/src/update_handler.rs
generall: I think this is fine, since we normally shouldn't fill the queue that aggressively anymore

PR #8397, opened 2026-03-13
Concurrent gridstore value deletion

monoid:
Allow gridstore value deletion to be concurrent with value retrieval.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/gridstore/src/gridstore/mod.rs
+++ b/lib/gridstore/src/gridstore/mod.rs
@@ -304,7 +304,7 @@ impl<V: Blob> Gridstore<V> {
     ///
     /// Returns None if the point_offset, page, or value was not found.
     /// Returns the deleted value otherwise.
-    pub fn delete_value(&mut self, point_offset: PointOffset) -> Result<Option<V>> {
+    pub fn delete_value(&self, point_offset: PointOffset) -> Result<Option<V>> {
         let Some(pointer) = self.tracker.write().unset(point_offset)? else {
             return Ok(None);
         };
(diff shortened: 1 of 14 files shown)


lib/segment/Cargo.toml
```diff
@@ -43,6 +43,7 @@ pprof = { workspace = true }
 
 [dependencies]
 bytemuck = { workspace = true }
+dashmap = { workspace = true }
```
coszio, line 46: Dashmap is essentially a sharded HashMap, each with its own RwLock. Did you consider `papaya`? It is quite performant and lock-free

lib/gridstore/src/gridstore/mod.rs
```diff
@@ -304,7 +304,7 @@ impl<V: Blob> Gridstore<V> {
     ///
     /// Returns None if the point_offset, page, or value was not found.
     /// Returns the deleted value otherwise.
-    pub fn delete_value(&mut self, point_offset: PointOffset) -> Result<Option<V>> {
+    pub fn delete_value(&self, point_offset: PointOffset) -> Result<Option<V>> {
```
coszio, line 307: I didn't consider that this could be done, but I believe it is sound. We are just localizing the write lock to the tracker component

lib/segment/Cargo.toml
monoid: Well, it used in the [in-memory mock implementation and rocksdb-related](https://github.com/qdrant/qdrant/blob/master/lib/segment/src/payload_storage/payload_storage_enum.rs#L22-L27) code.  But I will have a look at `papaya`, thank you.

coszio (approved):
Assuming the choice of concurrent hashmap is made, LGTM

agourlay:
I took the liberty to cancel the test run, it seems to not finish

slow

monoid:
The PR doesn't seems to be needed.

Issue #8349, opened 2026-03-10
Fail test_streaming_snapshot_consensus_freeze.py


agourlay:
https://github.com/qdrant/qdrant/actions/runs/22897236917/job/66434346964

```
=================================== FAILURES ===================================
___________ test_streaming_snapshot_freezes_collection_via_consensus ___________

tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-1/test_streaming_snapshot_freeze0')

    def test_streaming_snapshot_freezes_collection_via_consensus(tmp_path: pathlib.Path):
        """Streaming snapshot READ guards + consensus WRITE = frozen collection."""
        assert_project_root()
    
        # Setup cluster
        peer_uris, _, _ = start_cluster(tmp_path, 3)
        create_collection(peer_uris[0])
        wait_for(check_collection_green, peer_uris[0], COLLECTION, wait_for_timeout=30)
        insert_data(peer_uris[0], NUM_POINTS)
        time.sleep(2)
    
        shard_peer, shard_id, peer_id = find_shard_peer(peer_uris)
        assert shard_peer, "No active shard found"
        print(f"Shard {shard_id} on peer {peer_id} at {shard_peer}")
    
        # Baseline search
        timedout, dur, resp = timed_request(
            requests.post, f"{shard_peer}/collections/{COLLECTION}/points/search",
            timeout=10, json={"vector": [0.1] * DIM, "limit": 5},
        )
        assert not timedout and resp.ok, f"Baseline search failed: timedout={timedout}"
        print(f"Baseline search: {dur:.3f}s")
    
        # Start slow downloads (hold READ guards) + trigger consensus write
        stop_event = threading.Event()
>       dl_results, dl_threads = start_slow_downloads(shard_peer, shard_id, stop_event)
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/consensus_tests/test_streaming_snapshot_consensus_freeze.py:187: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

peer_uri = 'http://127.0.0.1:52411/', shard_id = 0
stop_event = <threading.Event at 0x7ff7b054e570: unset>, n = 2

    def start_slow_downloads(peer_uri, shard_id, stop_event, n=2):
[...]

PR #8400, opened 2026-03-14
fix: race in start_slow_downloads causes streaming assert to fail (#8349)

Aristotengalunzeza100:
The test fails with:

    AssertionError: No downloads streaming: [{'phase': 'streaming'}, {'phase': 'streaming'}]
    assert 0 > 0

`slow_download` sets `result["phase"] = "streaming"` as soon as the HTTP
headers come back, then enters `iter_content()`. `result["bytes"]` is only
written when the first chunk lands. The wait loop in `start_slow_downloads`
broke out on `phase == "streaming"`, so the assert below it could run before
any chunk had been received — hence `bytes` missing from the dict entirely.

Fix: loop until `bytes > 0` as well (or a terminal phase). Also bumps the
per-download deadline to 30 s; with 50 k × dim-128 vectors the snapshot
generator needs a moment before the first byte goes out on a loaded runner.

Reproduces reliably with:
    pytest tests/consensus_tests/test_streaming_snapshot_consensus_freeze.py -s -v

### DIFF
--- a/tests/consensus_tests/test_streaming_snapshot_consensus_freeze.py
+++ b/tests/consensus_tests/test_streaming_snapshot_consensus_freeze.py
@@ -117,7 +117,7 @@ def trigger_write_lock(peer_uri, result):
 
 
 def start_slow_downloads(peer_uri, shard_id, stop_event, n=2):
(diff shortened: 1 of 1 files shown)


agourlay:
all contributions must target the `dev` branch :)

PR #8405, opened 2026-03-15
refactor mmap hashmap

generall:
- [manual] split mmap_hashmap into multilpe files
- [AI] first iteration
- [AI] impleent custor for each key


xzfc (approved):
Reviewed with @generall.

PR #8410, opened 2026-03-16
Fix update worker hanging in deferred wait loop on config change

agourlay:
## Problem

  When `prevent_unoptimized=true` and `max_optimization_threads=0`, any `wait=true` write enters `wait_for_deferred_points_ready`, which loops sending `Nop` to the optimizer and waiting for `optimization_finished`. 

Since the optimizer has `max_handles=0`, it immediately signals "finished" without doing any work, creating a tight busy-spin that never terminates.

When the user then changes `max_optimization_threads` to `"auto"` via a config update, `run_workers` spawns new workers with the correct config — but the **old** update worker is never cancelled. It stays stuck in the deferred wait loop, talking to the old optimizer. 

The old `JoinHandle` is merely detached on drop, not stopped.

Since `wait_for_deferred_points_ready` doesn't check any cancellation signal, the old worker spins forever, blocking the HTTP response for the in-flight `wait=true` write.

  ## Fix

  **Cancel old update worker on restart** (`update_handler.rs`)

  `run_workers` now calls `self.update_worker_cancel.cancel()` before creating a new token. This signals the old worker to stop.

  **Check cancellation in deferred wait** (`update_worker.rs`)

  `wait_for_deferred_points_ready` now takes a `CancellationToken` and uses `tokio::select!` to check it alongside `optimization_finished_receiver.changed()`.

On cancellation it returns a `Cancelled` error. The caller converts this to `UpdateStatus::WaitTimeout` (HTTP 200 with `"status": "wait_timeout"`) — the update is applied, only the deferred visibility wait was interrupted.

  **Remove spurious `optimization_finished` signals** (`optimization_worker.rs`)

  When the optimizer cannot launch any work (thread limit reached or no optimization planned), it no longer sends `optimization_finished`. This eliminates the busy-spin — without the signal, the deferred wait loop blocks quietly on `changed()` until either a real optimization completes or the cancellation token fires.

  ## Test

  The consensus test `test_shard_snapshot_transfer_deferred` now exercises the full scenario:

  - `wait=true` with `max_optimization_threads=0` → client times out (server blocks in deferred wait)
  - Config change to `max_optimization_threads="auto"` → old worker is cancelled
  - Retry `wait=true` → may get `wait_timeout` status (old worker cancelled) until the new worker handles it with `completed` status

### DIFF
--- a/lib/collection/src/shards/local_shard/shard_ops.rs
+++ b/lib/collection/src/shards/local_shard/shard_ops.rs
@@ -33,6 +33,7 @@ use crate::profiling::interface::log_request_to_collector;
 use crate::shards::local_shard::LocalShard;
 use crate::shards::shard_trait::{ShardOperation, WaitUntil};
 use crate::update_handler::{OperationData, UpdateSignal};
+use crate::update_workers::internal_update_result::InternalUpdateResult;
 
 #[async_trait]
 impl ShardOperation for LocalShard {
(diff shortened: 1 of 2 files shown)


lib/collection/src/update_workers/update_worker.rs
```diff
@@ -144,24 +145,41 @@ impl UpdateWorkers {
...
                         )
                         .await;
-                        if let Err(err) = wait_result {
-                            send_feedback(sender, Err(err), op_num);
-                            continue;
+                        // Waiting for deferred points should never error out as it would mark the shard as dead
+                        if let Err(wait_result) = wait_result {
+                            log::warn!("Failed to await for optimizers: {wait_result}");
+                            UpdateStatus::WaitTimeout // ToDo: Consider special status
+                        } else {
```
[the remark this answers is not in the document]
agourlay: this is done on purpose, waiting for deferred point should not return an error to the user

lib/collection/src/shards/local_shard/shard_ops.rs
```diff
@@ -125,18 +126,25 @@ impl ShardOperation for LocalShard {
...
+                            "Operation ID from WAL should match the one received from update worker"
+                        );
+                        Ok(UpdateResult {
+                            operation_id: Some(op_num),
+                            status,
+                            clock_tag: operation.clock_tag,
+                        })
+                    }
+                    Err(elapsed) => {
+                        let _elapsed: Elapsed = elapsed;
```
JojiiOfficial, line 141: Any reason to keep the `let _elapsed: Elapsed = elapsed;`?
```suggestion
                    Err(_elapsed) => {
```

JojiiOfficial (approved):

lib/collection/src/shards/local_shard/shard_ops.rs
agourlay: I want to enforce the error type that is being ignored.

In case we refactor this and the error type becomes `CollectionResult` which should not be ignored.

PR #8414, opened 2026-03-16
Add deferred point count to UpdateQueueInfo

JojiiOfficial:
Depends on #8366

Adds the amount of deferred points to the `UpdateQueueInfo` to make it public via `CollectionInfo`.

### DIFF
--- a/lib/collection/src/shards/local_shard/mod.rs
+++ b/lib/collection/src/shards/local_shard/mod.rs
@@ -1024,7 +1024,7 @@ impl LocalShard {
                 .await;
 
             if let Err(err) = &deferred_point_count {
-                log::warn!("Failed to get deferred ponit counts: {err:?}");
+                log::warn!("Failed to get deferred point counts: {err:?}");
             }
 
             deferred_point_count.ok()


lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -1080,7 +1119,7 @@ impl LocalShard {
 
         let (status, optimizer_status) = self.local_shard_status().await;
 
-        let update_queue = self.local_update_queue_info();
+        let update_queue = self.local_update_queue_info().await;
```
agourlay, line 1141: It seems `local_update_queue_info` is also re-iterating over all segments.

Can't we merge the `deferred_point_count` accumulation into the previous segments traversal?

Was it done this way for re-usability?

JojiiOfficial: > Was it done this way for re-usability?

Exactly. We also use this function for telemetry [here](https://github.com/qdrant/qdrant/blob/4bd27d384f8e5d61acc0e12f4f75b49903fd2aa4/lib/collection/src/shards/local_shard/telemetry.rs#L116). So it's duplicate code vs iterating twice. To have a single source of truth, I went with the iterating twice approach but I don't have a strong preference here.

JojiiOfficial: On a second thought, we could add a `segments: Option<&[LockedSegment]>` parameter to `local_shard_status()` and don't lock if `Some(..)`. It's a bit hacky but his way we can reuse this function and keep code-duplication low. Wdyt?

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -970,10 +970,56 @@ impl LocalShard {
...
+            f(&segments)
+        });
+
+        AbortOnDropHandle::new(handle)
+    }
+
+    pub async fn local_update_queue_info(&self) -> ShardUpdateQueueInfo {
+        let deferred_point_count = self
+            .do_with_segments(|segments| {
+                segments
```
agourlay: in that case, we actually just need to iterate over the appendable segments

lib/collection/src/shards/local_shard/mod.rs
agourlay: I agree sounds a bit hacky.

lib/collection/src/shards/local_shard/mod.rs
JojiiOfficial: https://github.com/qdrant/qdrant/pull/8414/commits/13d93a933d34558e454fa338417082c4c1ad015d

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -970,10 +970,66 @@ impl LocalShard {
...
+            };
+
+            f(&segments)
+        });
+
+        AbortOnDropHandle::new(handle)
+    }
+
+    pub async fn local_update_queue_info(&self) -> ShardUpdateQueueInfo {
+        let deferred_point_count = self
```
agourlay: why not computing this only if `prevent_unoptimized` is on?

JojiiOfficial: Good point. Done in https://github.com/qdrant/qdrant/pull/8414/commits/96c65eec22742203374a9a86cb3304d21646309e

lib/collection/src/shards/local_shard/mod.rs
```diff
@@ -971,10 +971,71 @@ impl LocalShard {
...
+            let deferred_point_count = self
+                .do_with_segments(
+                    |segments| {
+                        segments
+                            .iter()
+                            .map(|segment| segment.get().read().deferred_point_count())
+                            .sum::<usize>()
+                    },
+                    true,
+                )
```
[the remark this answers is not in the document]
JojiiOfficial: Only appendable segments have deferred points. That's by design and will never change

PR #8434, opened 2026-03-17
Further split segment traits

monoid:
Make Y-shaped segment trait hierarchy: introduce a `SearchSegmentEntry` with common functionality, and `NonAppendableSegmentEntry` for non-appendable segments, and `SegmentEntry` for appendable ones.

It would allow to to have `NonAppendableSegmentEntry` and `SegmentEntry` methods have different mutability, and have different segment type to have different implementation.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/entry/entry_point.rs
+++ b/lib/segment/src/entry/entry_point.rs
@@ -293,77 +293,7 @@ pub trait StorageSegmentEntry: ReadSegmentEntry {
 /// Assume all operations are idempotent - which means that no matter how many times an operation
 /// is executed - the storage state will be the same.
 pub trait NonAppendableSegmentEntry: StorageSegmentEntry {
-    fn delete_point_concurrent(
-        &mut self,
-        op_num: SeqNumberType,
-        point_id: PointIdType,
-        hw_counter: &HardwareCounterCell,
-    ) -> OperationResult<bool>;
-}
-
-/// Define mutable operations which can be performed with Segment or Segment-like entity.
-///
-/// Assume all operations are idempotent - which means that no matter how many times an operation
-/// is executed - the storage state will be the same.
-///
-/// This is not a superset of `NonAppendableSegmentEntry` as its operations (will) differ in mutability.
-pub trait SegmentEntry: StorageSegmentEntry {
     fn delete_point(
-        &mut self,
-        op_num: SeqNumberType,
-        point_id: PointIdType,
-        hw_counter: &HardwareCounterCell,
-    ) -> OperationResult<bool>;
-
-    fn upsert_point(
-        &mut self,
-        op_num: SeqNumberType,
-        point_id: PointIdType,
-        vectors: NamedVectors,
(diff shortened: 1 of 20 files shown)


lib/segment/src/entry/entry_point.rs
```diff
@@ -27,11 +27,11 @@ use crate::types::{
     VectorName, VectorNameBuf, WithPayload, WithVector,
 };
 
-/// Define all operations which can be performed with non-appendable Segment or Segment-like entity.
+/// Define all operations on segment that do not require mutable access.
 ///
 /// Assume all operations are idempotent - which means that no matter how many times an operation
 /// is executed - the storage state will be the same.
-pub trait NonAppendableSegmentEntry: SnapshotEntry {
+pub trait SearchSegmentEntry: SnapshotEntry {
```
generall: `flush`, `drop_data`, delete & create payload index should not be part of read-only segment.

generall: Also, how about renaming `Search` -> `Read`? Cause it is not only search happening here

lib/shard/src/locked_segment.rs
```diff
@@ -57,7 +57,7 @@ impl LockedSegment {
         }
     }
 
-    pub fn get_non_appendable(&self) -> &RwLock<dyn NonAppendableSegmentEntry> {
+    pub fn get_non_appendable(&self) -> &RwLock<dyn SearchSegmentEntry> {
```
generall: get_non_appendable -> get_read?

lib/shard/src/proxy_segment/segment_entry.rs
```diff
@@ -818,8 +774,63 @@ impl SegmentEntry for ProxySegment {
...
+                    }
+                }
+                None
+            }
+        };
+
+        self.set_deleted_offset(point_offset);
+
+        Ok(was_deleted)
     }
```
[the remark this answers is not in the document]
generall: this PR only moves function, it doesn't change it

lib/segment/src/entry/entry_point.rs
monoid: I think `flush` and `drop` make sense for the base trait as it is common functionality.  For the indices, I suggest to keep it as is, because later I will implement it differently (perhaps, I will do a separate method for index storage as I do for point mappings).

I can do refine the hierarchy in even more detail: `Base` (`flush`, `drop`) -> `Read` (search, etc) -> other.

monoid: Or `SegmentEntry: ReadEntry, StorageEntry`.

generall (changes requested):
newly introduced `*_mut` functions duplicate a lot of code. They are not strictly necessary for this PR, please remove

generall:
@monoid there are conflicts again =(

monoid:
> @monoid there are conflicts again =(

Rebased.

PR #8435, opened 2026-03-17
Don't insert deferred points into sparse index

JojiiOfficial:
Depends on #8415 

This PR reverts changes to sparse vector search from https://github.com/qdrant/qdrant/pull/8299 and instead of filtering, we now completely ignoring inserting deferred points into the sparse index. This reduces code complexity and potentially improves on search performance.

## Design decision

To prevent passing `deferred_internal_id` in the `VectorIndex` trait, the `SparseVectorIndex` struct now holds a copy of that value.
The disadvantage of this approach is having multiple source of truths and having to adjust the existing unit tests.
However, it greatly improves on the necessary changes and overall code style.

## Potential issues
An inconsistency issue that must be considered here is the scenario of a user changing `prevent_unoptimized` and then restarting Qdrant. Since we don't persist the parameter, we will show all previously deferred points _except_ sparse vectors, since they haven't been inserted into the index in the beginning. This means search operations could become inconsistent depending on which vector was used during search.
The optimizer will deal with this inconsistency and create proper sparse indices after optimization. With other words, this inconsistency scenario will only be present for a limited amount of time.

## Note
The changes in `lib/sparse/src/index/search_context.rs` have been checked out to the latest commit before #8299. Therefore this file can be easily reviewed by file comparison.

### DIFF
--- a/lib/segment/src/index/sparse_index/sparse_vector_index.rs
+++ b/lib/segment/src/index/sparse_index/sparse_vector_index.rs
@@ -575,7 +575,7 @@ impl<TInvertedIndex: InvertedIndex> VectorIndex for SparseVectorIndex<TInvertedI
         let mut results = Vec::with_capacity(vectors.len());
         let mut prefiltered_points = None;
 
-        assert_eq!(
+        debug_assert_eq!(
             self.deferred_internal_id,
             query_context.deferred_internal_id(),
             "SparseIndex and VectorQueryContext deferred_internal_id consistency violated."


lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -681,19 +683,26 @@ impl<TInvertedIndex: InvertedIndex> VectorIndex for SparseVectorIndex<TInvertedI
...
             old_vector.map(SparseVector::try_from).transpose()?;
 
-        // do not upsert empty vectors into the index
-        if !vector.is_empty() {
+        let point_is_deferred = self
+            .deferred_internal_id
+            .is_some_and(|deferred| id >= deferred);
+
+        // do not upsert empty or deferred vectors into the index
+        if !vector.is_empty() && !point_is_deferred {
```
JojiiOfficial: The main change of this PR, ignoring deferred points during insertion.

IvanPleshkov:
`VectorQueryContext` contains deferred id. I propose to add to the sparse search an assert to check that `VectorQueryContext::deferred_internal_id` is equal to the value in `SparseVectorIndex`.

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -681,19 +690,26 @@ impl<TInvertedIndex: InvertedIndex> VectorIndex for SparseVectorIndex<TInvertedI
...
```
IvanPleshkov: I propose to return because of deferred earlier before the conversion into `SparseVector`, we perform a couple of unnecessary allocations

JojiiOfficial: Done https://github.com/qdrant/qdrant/pull/8435/commits/2482aa99b1114cb09738c5f83cc67bb01c33de55

IvanPleshkov: Hmmm... Let's discuss if it's safe to just return before upserting into the index. I'm not deep in the context but I think we are good just to call the return statement before upsertion

lib/segment/src/segment_constructor/segment_constructor_base.rs
```diff
@@ -689,9 +692,7 @@ fn create_segment(
         deferred_point_status: None,
     };
 
-    if let Some(deferred_internal_id) = deferred_internal_id
-        && segment.is_appendable()
-    {
+    if let Some(deferred_internal_id) = deferred_internal_id {
```
IvanPleshkov, line 697: I like that you refactored the `is_appendable` check.
Maybe as a next change after release, but I think we can go deeper.
We can move `calculate_deleted_deferred_point_count` into if tracker trait (as a default fn implementation) with deferred internal id as an argument. It will allow us to remove `deferred_point_status: None` while segment struct construction, and we can remove this ugly part where we modify just the created segment

lib/segment/src/segment_constructor/segment_builder.rs
```diff
@@ -686,6 +688,9 @@ impl SegmentBuilder {
                 if sparse_vector_config.index.index_type.is_on_disk() {
                     index.clear_cache()?;
                 }
+
+                // Ensure we don't use the sparse vector index in future because it's missing proper setup of deferred points.
+                drop(index);
```
IvanPleshkov, line 693: I like this `drop` call to mark in the code that we don't use the created index with the tmp deferred internal ID param.

lib/segment/src/index/sparse_index/sparse_vector_index.rs
```diff
@@ -572,6 +574,13 @@ impl<TInvertedIndex: InvertedIndex> VectorIndex for SparseVectorIndex<TInvertedI
     ) -> OperationResult<Vec<Vec<ScoredPointOffset>>> {
         let mut results = Vec::with_capacity(vectors.len());
         let mut prefiltered_points = None;
+
+        assert_eq!(
```
agourlay: IMHO this should be a debug_assert + a warning log

blowing up the search thread is too radical

agourlay:
I am really on the fence regarding the duplication of the source of truth.

> To prevent passing deferred_internal_id in the VectorIndex trait, the SparseVectorIndex struct now holds a copy of that value.

How bad is it to change the trait to pass the deferred_id there?

JojiiOfficial:
> How bad is it to change the trait to pass the deferred_id there?

It would be required to change in these 4 places:
https://github.com/qdrant/qdrant/blob/bd34d52578399a368f203ffe029780a37d817136/lib/segment/src/index/vector_index_base.rs#L196
https://github.com/qdrant/qdrant/blob/6ddb81d1b6eeed235e29a4e545a6dd06b4258031/lib/segment/src/index/plain_vector_index.rs#L84
https://github.com/qdrant/qdrant/blob/6ddb81d1b6eeed235e29a4e545a6dd06b4258031/lib/segment/src/index/hnsw_index/hnsw.rs#L1405
https://github.com/qdrant/qdrant/blob/49cc0eb050b3ab828e4398e64b76eff9ab5651fc/lib/segment/src/index/sparse_index/sparse_vector_index.rs#L567

I think it is a potential footgun to make it a parameter in the trait signature and then only apply it in one of the implementations. So if we'd add `deferred_id` to the trait, we should also account for it in all of them and not silently ignore the value in 3/4 cases.

lib/segment/src/index/sparse_index/sparse_vector_index.rs
JojiiOfficial: Done: https://github.com/qdrant/qdrant/pull/8435/commits/d3d61eb8f20e481552d5a4e4c893c504e78737ba

PR #8443, opened 2026-03-18
Use `GeoHash` instead of `&GeoHash`

xzfc:
First commit in the stack #8433.


xzfc:
> I don't understand why we should care about it

My private branch that integrates UIO into `geo_index` depends on it. I extracted this change for easier review because it's self-contained.

PR #8444, opened 2026-03-18
Propagate OperationResult [1/4]: payload_blocks

xzfc:
Second commit in the stack #8433.

### DIFF
--- a/lib/segment/src/index/struct_payload_index.rs
+++ b/lib/segment/src/index/struct_payload_index.rs
@@ -1007,7 +1007,7 @@ impl PayloadIndex for StructPayloadIndex {
         &self,
         field: PayloadKeyTypeRef,
         threshold: usize,
-    ) -> Box<dyn Iterator<Item = PayloadBlockCondition> + '_> {
+    ) -> Box<dyn Iterator<Item = OperationResult<PayloadBlockCondition>> + '_> {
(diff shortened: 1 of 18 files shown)


lib/segment/src/index/struct_payload_index.rs
```diff
@@ -1007,14 +1008,16 @@ impl PayloadIndex for StructPayloadIndex {
...
             Some(indexes) => {
                 let field_clone = field.to_owned();
-                Box::new(indexes.iter().flat_map(move |field_index| {
-                    field_index.payload_blocks(threshold, field_clone.clone())
-                }))
+                let blocks: Vec<_> = indexes
+                    .iter()
+                    .map(|field_index| field_index.payload_blocks(threshold, field_clone.clone()))
+                    .process_results(|iter| iter.collect_vec())?;
+                Ok(Box::new(blocks.into_iter().flatten()))
```
generall: this is questionable. I would prefer to avoid collect

xzfc: I though it's acceptable to collect these because:
- It's only used during HNSW building, and I took an assumption that the total number of payload blocks shouldn't be that high.
- Many payload indices do something like `.collect_vec().into_iter()` anyway internally.

But fixed anyway (now it's an iterator of results).

PR #8455, opened 2026-03-19
Fix condition deserialization issue

generall:
```
I see Can't deserialize entry, probably corrupted WAL or version mismatch: WriteWalError("Can't deserialize entry, probably corrupted WAL   
  or version mismatch: ErrorImpl { code: Message(\"data did not match any variant of untagged enum CollectionUpdateOperations\"), offset: 0 }")
```

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -3428,14 +3428,13 @@ impl<'de> serde::Deserialize<'de> for Condition {
 
         // Special case: FieldCondition first to surface datetime parse errors.
         // Untagged enum would swallow these errors with generic message.
-        if let serde_value::Value::Map(ref map) = value {
-            let key_key = serde_value::Value::String("key".to_string());
-            if map.contains_key(&key_key) {
-                return value
-                    .deserialize_into::<FieldCondition>()
-                    .map(Condition::Field)
-                    .map_err(serde::de::Error::custom);
-            }
+        if let serde_value::Value::Map(obj) = &value
+            && obj.contains_key(&serde_value::Value::String("key".into()))
+        {
+            return value
+                .deserialize_into()
+                .map(Condition::Field)
+                .map_err(serde::de::Error::custom);
         }
 
         // All other variants handled by ConditionUntagged (compiler-safe)
--- a/lib/shard/src/operations/mod.rs
+++ b/lib/shard/src/operations/mod.rs
@@ -444,4 +444,31 @@ mod tests {
 
         assert_eq!(operation, deserialized);
     }
+
+    #[test]
+    fn test_wal_roundtrip_delete_by_filter_with_has_id_uuids() {
+        use crate::wal::WalRawRecord;
+
+        let uuids: Vec<PointIdType> = vec![ExtendedPointId::Uuid(
+            uuid::Uuid::parse_str("6ba7b810-9dad-11d1-80b4-00c04fd430c8").unwrap(),
+        )];
+
+        let filter = Filter {
+            should: None,
+            min_should: None,
+            must: None,
+            must_not: Some(vec![Condition::HasId(HasIdCondition::from(
+                uuids.into_iter().collect::<ahash::AHashSet<_>>(),
+            ))]),
+        };
+
+        let operation = CollectionUpdateOperations::PointOperation(
+            PointOperations::DeletePointsByFilter(filter),
+        );
+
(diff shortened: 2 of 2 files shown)


ffuugoo (approved):

lib/segment/src/types.rs
```diff
@@ -3421,26 +3419,30 @@ impl<'de> serde::Deserialize<'de> for Condition {
...
-        } else {
-            ConditionUntagged::deserialize(deserializer).map(Condition::from)
         }
+
+        // All other variants handled by ConditionUntagged (compiler-safe)
+        value
+            .deserialize_into::<ConditionUntagged>()
+            .map(Condition::from)
+            .map_err(serde::de::Error::custom)
     }
```
ffuugoo, line 3445:
```suggestion
            let value = serde_value::Value::deserialize(deserializer)?;

            // Special case: FieldCondition first to surface datetime parse errors.
            // Untagged enum would swallow these errors with generic message.
            if let serde_value::Value::Map(obj) = &value
                && obj.contains_key(&serde_value::Value::String("key".into()))
            {
                return value
                    .deserialize_into()
                    .map(Condition::Field)
                    .map_err(serde::de::Error::custom);
            }

            // All other variants handled by ConditionUntagged (compiler-safe)
            value
                .deserialize_into::<ConditionUntagged>()
                .map(Condition::from)
                .map_err(serde::de::Error::custom)
```

lib/segment/src/types.rs
```diff
@@ -5271,6 +5273,37 @@ mod tests {
...
+            ))]),
+            must_not: Some(vec![Condition::HasId(HasIdCondition {
+                has_id: [ExtendedPointId::Uuid(uuid)].into_iter().collect(),
+            })]),
+        };
+
+        let cbor_bytes = serde_cbor::to_vec(&filter).unwrap();
+        let deserialized: Filter = serde_cbor::from_slice(&cbor_bytes).unwrap();
+        assert_eq!(filter, deserialized);
+    }
```
ffuugoo, line 5305: Test copy-pasted two times?

lib/shard/src/operations/mod.rs
```diff
@@ -419,4 +419,29 @@ mod tests {
...
+
+        let operation = CollectionUpdateOperations::PointOperation(
+            PointOperations::DeletePointsByFilter(filter),
+        );
+
+        let cbor_bytes = serde_cbor::to_vec(&operation).unwrap();
+        let deserialized: CollectionUpdateOperations = serde_cbor::from_slice(&cbor_bytes).unwrap();
+
+        assert_eq!(operation, deserialized);
+    }
```
ffuugoo, line 446: Test copy-pasted two times?

lib/segment/src/types.rs
ffuugoo: A bit cleaner code (but exactly the same logic)

ffuugoo:
Minimal repro test:

```rust
#[test]
fn test_condition_serde_cbor_transcode() {
    let uuids: Vec<PointIdType> = vec![
        ExtendedPointId::Uuid(
            uuid::Uuid::parse_str("550e8400-e29b-41d4-a716-446655440000").unwrap(),
        ),
        ExtendedPointId::Uuid(
            uuid::Uuid::parse_str("6ba7b810-9dad-11d1-80b4-00c04fd430c8").unwrap(),
        ),
    ];

    let condition = Condition::HasId(HasIdCondition::from(
        uuids.into_iter().collect::<ahash::AHashSet<_>>(),
    ));

    // `#[serde(untagged)]` changes `is_human_readable` hint of the deserializer and caused
    // deserialization failure in `Condition` deserialize impl
    #[derive(Debug, PartialEq, serde::Deserialize, serde::Serialize)]
    #[serde(untagged)]
    enum Untagged<T> {
        Untagged(T),
    }

    let condition = Untagged::Untagged(condition);

    let cbor = serde_cbor::to_vec(&condition).expect("value succesfully serialized");
    let transcoded = serde_cbor::from_slice(&cbor).expect("value succesfully deserialized");

    assert_eq!(condition, transcoded);
}
```

lib/segment/src/types.rs
ffuugoo: Ah, they slightly different

lib/shard/src/operations/mod.rs
ffuugoo: Tests are slightly different

PR #8458, opened 2026-03-19
UIO: add flag to preserve callback order in `read_batch`

xzfc:
I plan to implement an iterator interface on top of `UniversalRead`.
Particularly, I want to implement the [`FallibleIterator::try_fold`](https://docs.rs/fallible-iterator/latest/fallible_iterator/trait.FallibleIterator.html#method.try_fold) method.

To do so, I need to adjust `read_batch` to let it call callbacks items in order.
This PR adds an additional generic parameter `YieldOrder` to force strict ordering.

For `io_uring` implementation, the strict ordering is achieved by using the new ringbuf-like structure that holds items that completed out of sequence. The same structure can be reused later for S3 or other async implementations.

Also, this PR adds a generic error parameter `E` to `read_batch` to let it fit into the `try_fold` interface. (and I think this approach is more idiomatic anyway)

### DIFF
--- a/lib/common/common/src/universal_io/io_uring.rs
+++ b/lib/common/common/src/universal_io/io_uring.rs
@@ -13,16 +13,16 @@ use ahash::AHashMap;
 use fs_err as fs;
 
 use super::*;
-use crate::generic_consts::AccessPattern;
+use crate::generic_consts::{AccessPattern, YieldOrder};
 
 thread_local! {
     static IO_URING: io::Result<RefCell<IoUring>> = init_io_uring().map(RefCell::new);
 }
 
-const IO_URING_QUEUE_LENGTH: u32 = 16;
+const IO_URING_QUEUE_LENGTH: usize = 16;
 
 fn init_io_uring() -> io::Result<IoUring> {
-    let io_uring = IoUring::new(IO_URING_QUEUE_LENGTH)
+    let io_uring = IoUring::new(IO_URING_QUEUE_LENGTH as _)
         .map_err(|err| io_error_context(err, "failed to setup io_uring"))?;
 
     let mut probe = Probe::new();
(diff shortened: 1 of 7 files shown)


ffuugoo:
Two questions:
- Can it be implemented as wrapper-adaptor over `UniversalRead` instead?
- For your use-case for `try_fold`, do you already _need_ items to be in-order? Can you fold out-of-order items, or it won't work for your algo?

lib/common/common/src/universal_io/io_uring.rs
```diff
@@ -95,13 +95,14 @@ impl<T: bytemuck::Pod + 'static> UniversalRead<T> for IoUringFile {
         })?
     }
 
-    fn read_batch<P: AccessPattern>(
+    fn read_batch<P: AccessPattern, O: YieldOrder, E: From<UniversalIoError>>(
```
ffuugoo, line 98: no offense, but this is so ugly 🥲

xzfc:
> Can it be implemented as wrapper-adaptor over `UniversalRead` instead?

From interface perspective, yes. (if we explicitly require `UniversalRead` to not reorder callbacks more than `N` elements apart)
But from performance perspective, it will probably will do unnecessary work for `mmap`. Tho... I'll try later, maybe I can convince the compiler to optimize it away. 

> For your use-case for `try_fold`, do you already need items to be in-order? Can you fold out-of-order items, or it won't work for your algo?

We have a buch of code that uses iterators over mmaped slices. Sometimes the order is expected (e.g. [here](https://github.com/qdrant/qdrant/blob/8733d9f516ae4f52c2ac32c8f43925428fab50e9/lib/segment/src/index/field_index/geo_index/mmap_geo_index.rs#L420) and [here](https://github.com/qdrant/qdrant/blob/8733d9f516ae4f52c2ac32c8f43925428fab50e9/lib/segment/src/index/field_index/geo_index/immutable_geo_index.rs#L153); `points_map` are sorted). I think we can rewrite these parts of code to not rely on strictly ordered iterators. But the approach "provide ordered iterators and don't touch old well-tested algorithms, at least for now" is less intrusive.

lib/common/common/src/universal_io/io_uring.rs
xzfc: "We have `comptime` args at home"
`comptime` args at home:

generall:
> We have a buch of code that uses iterators over mmaped slices.

TBH, I would prefer to rewrite this code to be more callback/unorder friendly.
It is very likely that we would need to rewrite it anyway, for better batch-read compatibility.

I agree with @ffuugoo here, that `read_batch` becomes too complicated

ffuugoo:
> I'll try later, maybe I can convince the compiler to optimize it away.

We could specialize for `MmapUniversal` though?

Something like

```rust
enum OrderedReadSimple {
  Mmap(MmapUniversal),
  Uring(IoUringFile)
}
```

or

```rust
enum OrderedReadGeneric<R> {
  Mmap(MmapUniversal),
  Generic(R),
}
```

Either use `From<R>` + downcast inside to specialize for mmap or use explicit `From` impls (we only have a few `UniversalRead` impls anyway)

PR #8461, opened 2026-03-19
More precise shard transfer for deferred points

agourlay:
## Summary

 Improve the default shard transfer method selection for collections with `prevent_unoptimized` enabled.

 Instead of always forcing snapshot transfer when the config flag is set, check whether deferred points actually exist in the segments.

  ## Problem

  Previously, any collection with `prevent_unoptimized=true` would unconditionally use snapshot transfer, even if all segments were fully optimized and had no deferred points.

 This prevented using `stream_records` transfer in cases where it would be safe and potentially more efficient.

  ## Fix

  Add `has_deferred_points()` throughout the shard hierarchy (`LocalShard` → `Shard` → `ShardReplicaSet` → `ShardHolder`) to check the actual segment state. The default transfer method selection now:

  1. Checks `prevent_unoptimized` config (cheap gate — deferred points can only exist with this enabled)
  2. Scans appendable segments for actual deferred points (only if the config gate passes)
  3. Forces snapshot transfer only if deferred points are found
  4. Falls back to the configured default transfer method otherwise

### DIFF
--- a/lib/collection/src/collection/shard_transfer.rs
+++ b/lib/collection/src/collection/shard_transfer.rs
@@ -41,14 +41,18 @@ impl Collection {
             .await
             .map(|config| config.prevent_unoptimized.unwrap_or(false))
             .unwrap_or(false);
+        // Only scan shards for deferred points if prevent_unoptimized is enabled,
+        // since deferred points can only exist with that configuration.
         if prevent_unoptimized {
-            log::info!("Using snapshot transfer method because prevent_unoptimized is enabled");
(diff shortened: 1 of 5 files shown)


generall:
> Instead of always forcing snapshot transfer when the config flag is set, check whether deferred points actually exist in the segments.

But deferred points might appear after the transfer is started. 
Taking into account that we want to switch to snapshot transfer by default, I would prefer easier option to always prefer `wal_delta -> snapshot`

agourlay:
closing because of the TOCTOU race

PR #8462, opened 2026-03-19
impl populate for `CachedSlice`

coszio:
On top of #8368 

Handle `populate()` function for `CachedSlice` by reading minimal data for each of the blocks it spans.

### DIFF
--- a/lib/common/common/src/universal_io/disk_cache/cached_slice.rs
+++ b/lib/common/common/src/universal_io/disk_cache/cached_slice.rs
@@ -107,7 +107,7 @@ impl<T: bytemuck::Pod> CachedSlice<T> {
                 key: BlockId {
                     file_id: self.file_id,
                     offset: BlockOffset(
-                        u32::try_from(block_idx).expect("file too large for block cache (>70 TiB)"),
+                        u32::try_from(block_idx).expect("file too large disk cache"),
                     ),
                 },
                 // Request a single byte — enough to trigger caching the whole block.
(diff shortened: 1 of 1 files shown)


lib/common/common/src/disk_cache/controller.rs
```diff
@@ -220,6 +220,12 @@ pub(super) enum CacheRead<'a, O> {
     Miss(O),
 }
 
+impl<O> CacheRead<'_, O> {
+    pub fn is_miss(&self) -> bool {
+        matches!(self, CacheRead::Miss(_))
+    }
+}
```
xzfc: I think we don't this method.

lib/common/common/src/universal_io/disk_cache/cached_slice.rs
```diff
@@ -93,6 +93,36 @@ impl<T: bytemuck::Pod> CachedSlice<T> {
...
+            return Ok(());
+        }
+
+        let num_blocks = self.len_bytes.div_ceil(BLOCK_SIZE);
+        for block_idx in 0..num_blocks {
+            let req = BlockRequest {
+                key: BlockId {
+                    file_id: self.file_id,
+                    offset: BlockOffset(
+                        u32::try_from(block_idx).expect("file too large for block cache (>70 TiB)"),
```
xzfc: Nit: don't mention 70 TiB as it depends on `BLOCK_SIZE` value.

PR #8464, opened 2026-03-20
Chunked points deduplication

IvanPleshkov:
It seems that without chunks, deduplication locks the segment holder for too long time


generall:
> deduplication locks the segment holder for too long time


In the fix segment holder is locked for the same time, isn't it? Maybe the problem with segments themselves?

IvanPleshkov:
> > deduplication locks the segment holder for too long time
> 
> In the fix segment holder is locked for the same time, isn't it? Maybe the problem with segments themselves?

Even if this PR does not change tests, it is still nice to have in the release because we cannot actually predict how long the segment holder may be locked

PR #8466, opened 2026-03-20
Allow shorter reads in IoUring

coszio:
For using io-uring for disk-cache, sometimes the aligned reads mean that we will go past EOF. 

`O_DIRECT` requires reads to be aligned to some page/block size.

Imagine having a file with a slice of 128b items. When having block-size 4096, the file can have length of 4864, so we necessarily have to issue a read for the last bytes as 4096..8012, even when we will use only 4096..4864

This PR allows this to happen, and truncates the returned vector to account for it.

### DIFF
--- a/lib/common/common/src/universal_io/disk_cache/mod.rs
+++ b/lib/common/common/src/universal_io/disk_cache/mod.rs
@@ -78,6 +78,7 @@ impl<T: bytemuck::Pod> UniversalRead<T> for CachedSlice<T> {
             disk_parallel: _,
             populate: _,
             advice: _,
+            prevent_caching: _, // This is cached in disk, backed by a mmap
         } = options;
 
         debug_assert!(!writeable);


ffuugoo (commented):
There's an issue with this approach, because with this change we can't distinguish between EOF vs spurious short read.

When you request to read 10 bytes, it does not guarantee that all 10 bytes would be read. It can read _less_ bytes, even if it did not yet reach EOF. To ensure that you reached EOF, you need to make _another_ read (at current offset), and if that read returned 0, _that_ should mean you reached EOF.

Image

So, to not have to deal with this _yet_, I've just put `assert_eq` to ensure all reads match requested length.

coszio:
As partially discussed, in https://github.com/qdrant/qdrant/pull/8466/commits/9d755dcf6fa1a3cc06079f54fcc3f6e2d50aacf4 I have added the `prevent_caching` option to `OpenOptions`. 

Within the io-uring implementation, it is used to dynamically choose to allow short reads.

PR #8467, opened 2026-03-20
Disk cache with `UniversalRead` for slow disk

coszio:
Builds on top of #8466 

 - adds a `prevent_caching` parameter for `OpenOptions`, so that, if possible for the implementation, it skips caching the reads.
   - this lets `CachedFile` be opened with a generic backend for reading from the slow disk. Particularly, it chooses `IoUringFile` by default.
- add `read_batch` implementation all the way into fetching from slow disk.


coszio:
Not committing to this implementation for now

PR #8477, opened 2026-03-22
account for deferred points in proxies

generall:
Previous PR incorrectly assumed that deferred points can only exist in appendable segment.
But ProxySegment is non-appendable and it can contain appendable segment with deferred points.

### DIFF
--- a/lib/shard/src/optimize.rs
+++ b/lib/shard/src/optimize.rs
@@ -459,7 +459,10 @@ fn finish_optimization(
         // case.
         debug_assert!(
             versions.operation_version >= optimized_segment.point_version(point_id).unwrap_or(0),
-            "proxied point deletes should have newer version than point in segment",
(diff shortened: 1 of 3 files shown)


lib/shard/src/optimize.rs
```diff
@@ -459,7 +459,10 @@ fn finish_optimization(
         // case.
         debug_assert!(
             versions.operation_version >= optimized_segment.point_version(point_id).unwrap_or(0),
-            "proxied point deletes should have newer version than point in segment",
+            "proxied point deletes should have newer version than point in segment {} < {:?}, point id: {}",
+            versions.operation_version,
+            optimized_segment.point_version(point_id),
+            point_id,
```
generall, line 465: this not stricitly related to this PR, but I found the original issue while debugging this assertion

PR #8483, opened 2026-03-23
Make certain fields of ID trackers atomic

monoid:
To allow concurrent deletion, make deletion bitmasks and version vector to be atomics.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/common/common/src/bitvec.rs
+++ b/lib/common/common/src/bitvec.rs
@@ -1,10 +1,12 @@
+use std::sync::atomic::AtomicU64;
+
 use bitvec::order::{BitOrder, Lsb0};
 use bitvec::store::BitStore;
 
 use crate::types::PointOffsetType;
 
-pub type BitVec = bitvec::vec::BitVec<u64, Lsb0>;
-pub type BitSlice = bitvec::slice::BitSlice<u64, Lsb0>;
+pub type BitVec = bitvec::vec::BitVec<AtomicU64, Lsb0>;
+pub type BitSlice = bitvec::slice::BitSlice<AtomicU64, Lsb0>;
 
 /// Set deleted state in given bitvec.
 ///
--- a/lib/common/common/src/bitvec.rs
+++ b/lib/common/common/src/bitvec.rs
@@ -1,12 +1,10 @@
-use std::sync::atomic::AtomicU64;
-
 use bitvec::order::{BitOrder, Lsb0};
 use bitvec::store::BitStore;
 
 use crate::types::PointOffsetType;
 
-pub type BitVec = bitvec::vec::BitVec<AtomicU64, Lsb0>;
-pub type BitSlice = bitvec::slice::BitSlice<AtomicU64, Lsb0>;
+pub type BitVec = bitvec::vec::BitVec<u64, Lsb0>;
+pub type BitSlice = bitvec::slice::BitSlice<u64, Lsb0>;
 
 /// Set deleted state in given bitvec.
 ///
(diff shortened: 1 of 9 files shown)


lib/common/common/src/bitvec.rs
```diff
@@ -1,10 +1,12 @@
...
+
 use bitvec::order::{BitOrder, Lsb0};
 use bitvec::store::BitStore;
 
 use crate::types::PointOffsetType;
 
-pub type BitVec = bitvec::vec::BitVec<u64, Lsb0>;
-pub type BitSlice = bitvec::slice::BitSlice<u64, Lsb0>;
+pub type BitVec = bitvec::vec::BitVec<AtomicU64, Lsb0>;
+pub type BitSlice = bitvec::slice::BitSlice<AtomicU64, Lsb0>;
```
coszio: Do we really need to change this everywhere? I just checked how it interacts with atomics and it is basically always `Relaxed` ordering.

This means there is no guarantee that changes to the bitslice are not necessarily going to take effect on other threads at any given time. 

We would essentially be depending on how fast hardware propagates the write. Technically this is fine, but may introduce flakiness in tests.

monoid: Good point.  Anyway, we will need a custom atomic bitvec for deleted points because `bitvec` has only `set_aliased` for concurrent operations, but for correct read-only indices we will need `replace_aliased` which is absent.

monoid: Anyway, right now all the mutations are under `&mut` ref, so the synchronization is guaranteed by some external lock.

generall: > This means there is no guarantee that changes to the bitslice are not necessarily going to take effect on other threads at any given time.

Is related ordering about visibility in other threads? I thought it is more of an ordering within function, which is done by compiler

monoid: Yes, atomic ordering affects both ordering of execution done by compiler (and CPU as well) and changes' visibility (cache coherence).  Unfortunately, there's no ordering "synchronize only this cell".  Perhaps, it can be emulated with `fetch_or(0, Relaxed)` and needs extra benchmarking, but it affects only ARM, as on x64 all the orderings are same except `SeqCst`.

coszio: > Anyway, right now all the mutations are under &mut ref, so the synchronization is guaranteed by some external lock.

But the point of this PR is so that we can drop the external lock, isn't it?

monoid: Yes, eventually.  But if we pursue our own atomic bitmask implementation, this particular change would not be needed.

lib/segment/src/id_tracker/immutable_id_tracker.rs
```diff
@@ -334,8 +335,12 @@ impl ImmutableIdTracker {
...
         };
 
-        internal_to_version_wrapper[..internal_to_version.len()]
-            .copy_from_slice(internal_to_version);
+        for (dst, src) in internal_to_version_wrapper[..internal_to_version.len()]
+            .iter_mut()
+            .zip(internal_to_version)
+        {
+            *dst = src.load(AtomicOrdering::Acquire)
+        }
```
timvisee, line 343: Is the for loop and atomic load strictly necessary in a `new` function?

monoid: Yes, atomic load is the only safe way to get value from the atomics that come from another IdTracker.

It can be replaced here with Relaxed, but I don't think measuring the difference is possible.

coszio:
Too many conflicts now, we can revisit later

PR #8484, opened 2026-03-23
Avoid cast_vec of bytemuck

generall:
`bytemuck::cast_vec` is not safe. We need to avoid to use it. We also need to avoid extra generic in HNSW graph builder, as we only care about bytes.

### DIFF
--- a/lib/segment/src/index/hnsw_index/graph_links.rs
+++ b/lib/segment/src/index/hnsw_index/graph_links.rs
@@ -72,9 +72,15 @@ pub enum GraphLinksFormatParam<'a> {
 
 /// This trait lets the [`serialize_graph_links`] to access vector values.
 pub trait GraphLinksVectors {
+    /// Call `f` with the raw bytes of the base vector for `point_id`.
+    ///
     /// Base vectors will be included once per point on level 0.
     /// The layout of each vector must correspond to [`VectorLayout::base`].
-    fn get_base_vector(&self, point_id: PointOffsetType) -> OperationResult<Cow<'_, [u8]>>;
+    fn for_base_vector(
+        &self,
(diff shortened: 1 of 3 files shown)


lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -427,10 +441,12 @@ mod tests {
...
-                    assert_eq!(
-                        base_vector,
-                        vectors.get_base_vector(point_id).unwrap().as_ref()
-                    );
+                    vectors
+                        .for_base_vector(point_id, &mut |bytes| {
+                            assert_eq!(base_vector, bytes);
+                            Ok(())
+                        })
+                        .unwrap();
```
timvisee, line 449: Do we want to debug assert?


```suggestion
                    #[cfg(debug_assertions)]
                    vectors
                        .for_base_vector(point_id, &mut |bytes| {
                            assert_eq!(base_vector, bytes);
                            Ok(())
                        })
                        .unwrap();
```

lib/segment/src/index/hnsw_index/graph_links.rs
```diff
@@ -118,13 +124,17 @@ impl<'a> StorageGraphLinksVectors<'a> {
 impl<'a> GraphLinksVectors for StorageGraphLinksVectors<'a> {
     /// Note: uses [`Sequential`] because [`serializer::serialize_graph_links`]
     /// traverses base vectors in a sequential order.
-    fn get_base_vector(&self, point_id: PointOffsetType) -> OperationResult<Cow<'_, [u8]>> {
+    fn for_base_vector(
```
coszio, line 127: rename to `with_base_vector`?

PR #8485, opened 2026-03-23
Deferred updates by filter

IvanPleshkov:
This PR extends https://github.com/qdrant/qdrant/pull/8389 and applies such an idea to all updates by filter, not only deletion by filter

### DIFF
--- a/lib/shard/src/update.rs
+++ b/lib/shard/src/update.rs
@@ -459,16 +459,11 @@ pub fn delete_points_by_filter(
             let present: Vec<_> = all_matched_points
                 .iter()
                 .copied()
-                .filter(|point_id| segment.has_point(*point_id))
+                .filter(|point_id| {
+                    segment.has_point(*point_id) && !points_to_keep.contains(point_id)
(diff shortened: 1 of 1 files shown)


lib/shard/src/update.rs
```diff
@@ -400,40 +446,21 @@ pub fn delete_points_by_filter(
...
-                    && segment.point_is_deferred(*point_id)
-                {
-                    points_to_keep.insert(*point_id);
-                }
-            }
-            // Store per-segment points for expansion below.
+            let present: Vec<_> = all_matched_points
+                .iter()
+                .copied()
+                .filter(|point_id| segment.has_point(*point_id))
```
JojiiOfficial: Can we maybe already filter out points_to_keep here, instead of using `.retain()` below?

JojiiOfficial: https://github.com/qdrant/qdrant/pull/8485/commits/78d693d656b6e27e2767e23b906da019bbcd3baa

PR #8493, opened 2026-03-23
Universal `MmapFile` that can read any `T`

ffuugoo:
Implement `MmapFile` that does not depend on generic parameter `T` and implement `UniversalRead`/`UniversalWrite` for "any" `T` (e.g., single `MmapFile` can read `u8`/`f32`/`u64`/etc).

I've provided `MmapUniversal<T>` re-implementation based on `MmapFile`, thought I plan to completely replace it with `MmapFile` (either in this, or in a follow-up PR).

I've also added:
- `writeable` flag to `OpenOptions` for *runtime* read/write control (e.g., similar to `fs::OpenOptions::write`)
- and *generic* `ReadOnly` wrapper, that enforces I/O being read-only
  - runtime assert that `writeable == false`
  - and statically/at compile-time it only implements `UniversalRead`, but not `UniversalWrite`

My reasoning for this is:
- it's just easier to implement, less helper-traits/generics
- if we can specify read-only *type* in code (e.g., use `MmapUniversalRo`), we can as well specify `writeable: false`
  - it's potentially more error-prone, yes...
  - but we can enforce it with `ReadOnly` wrapper
- `ReadOnly` wrapper *is* an additional generic
  - but it's absolutely trivial
  - and *generic* for *all* current and future `UniversalRead` implementations

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/common/common/src/universal_io/disk_cache/mod.rs
+++ b/lib/common/common/src/universal_io/disk_cache/mod.rs
@@ -73,11 +73,14 @@ impl<T: bytemuck::Pod> UniversalRead<T> for CachedSlice<T> {
 
         // Disk-cache is backed by a single file
         let OpenOptions {
+            writeable,
             need_sequential: _,
             disk_parallel: _,
             populate: _,
             advice: _,
         } = options;
+
+        debug_assert!(!writeable);
 
         Ok(CachedSlice::open(controller, path.as_ref())?)
     }


lib/common/common/src/universal_io/mmap.rs
```diff
@@ -1,257 +1,254 @@
...
+/// Helper type to force `bytemuck::NoUninit` requirement
+#[derive(Copy, Clone, bytemuck::TransparentWrapper)]
+#[repr(transparent)]
+struct ForceNoUninit<T>(T);
+unsafe impl<T: Copy + 'static> bytemuck::NoUninit for ForceNoUninit<T> {}
+
+/// Helper type to force `bytemuck::ForceAnyBitPattern` requirement
+#[derive(Copy, Clone, bytemuck::TransparentWrapper)]
+#[repr(transparent)]
+struct ForceAnyBitPattern<T>(T);
```
generall: please explain this magic

lib/common/common/src/universal_io/file_ops.rs
```diff
@@ -1,5 +1,7 @@
 use std::path::{Path, PathBuf};
 
+use super::*;
```
generall, line 3: glob imports are considered antipattern

lib/common/common/src/universal_io/mmap.rs
```diff
@@ -1,257 +1,254 @@
...
+
+        let bytes = cast_slice(items);
+        mmap.copy_from_slice(bytes);
+
+        Ok(())
+    }
+}
+
+/// [`bytemuck::cast_slice`] without [`bytemuck::Pod`] requirement 😬
+fn cast_slice<T, U>(slice: &[T]) -> &[U]
```
xzfc: I suggest to not to pretend that it's a safe method:

```suggestion
unsafe fn cast_slice<T, U>(slice: &[T]) -> &[U]
```

As it can trigger UB:
```rust
let bools: &[bool] = cast_slice::<u8, bool>(&[2u8]);
println!("{:?} {:?}", bools[0], !bools[0]); // true true

let chr: char = cast_slice::<u32, char>(&[0xFFFFFFFFu32])[0];
println!("{chr:?}"); // failed printing to stdout: Bad address (os error 14)
```

Or use our `transmute_from_u8_to_slice` which doesn't pretend either and does the same thing.

lib/common/common/src/universal_io/mmap.rs
ffuugoo: `bytemuck::cast_*` have specific requirements for `FromT` and `IntoT` types.

`AnyBitPattern` requires that for `IntoT` any byte garbage that you might be casting, is a valid value. E.g., `ptr::NotNull<T>` is *not* `AnyBitPattern`, because the inner pointer can never be `0`/`NULL`, so you _can't_ cast arbitrary `[u8]` (say, `[0x00; 8]`) into `NotNull`.

Strictly speaking, this is incorrect/unsound, but I have the following logic here:

Before, we had `MmapType`/`MmapSlice` types, that would take memmap `&[u8]` and cast it into `&[T]`. So, *basically*, we are doing the same as `bytemuck::cast_slice::<T>(&mmap)`, but we only require `T: Copy + 'static`, without any additional "correctness" traits.

In `MmapFile`, I did not want to re-use our "handwritten" transmute functions, and wanted to rely on `bytemuck`, because I trust them to be more correct than whatever we might implement by hand. (E.g., `bytemuck::cast_vec` was strict with alignment, and we would probably miss that requirement if using simple `transmute` or trying to reimplement it ourselves.)

Not all types that we are reading/writing through `MmapUniversal` derive required `bytemuck` traits (and also, all trait bounds use `Copy + 'static`, so I'd have to update them all as well).

It worked before with just `Copy + 'static` (even though strictly speaking requirements *should* have been the same as `bytemuck` requires), so I assume our types *probably* satisfy these requirements, we just have to derive the traits and update the trait bounds, so *for now* I simply "forced" `NoUninit` and `AnyBitPattern` with transparent-wrappers.

ffuugoo: What I *plan* to do next, is update trait bounds to `bytemuck::Pod` and derive required traits everywhere (or where possible), but it might be best to be done in a follow-up PR, because it will touch a lot of small places with `Copy + 'static` -> `bytemuck::Pod` and such.

lib/common/common/src/universal_io/mmap.rs
ffuugoo: Fair on `unsafe`, but I do trust `bytemuck` to potentially be more correct than whatever we came up with. `bytemuck::cast_vec` was definitely stricter that we would assume.

lib/common/common/src/universal_io/mmap.rs
generall: `transmute_from_u8_to_slice` does have an allignment check

lib/common/common/src/universal_io/mmap.rs
xzfc: Ah, I see this plan https://github.com/qdrant/qdrant/pull/8493#discussion_r2988702532. Marking as resolved.

lib/common/common/src/universal_io/mmap.rs
ffuugoo: Yes, but alignment check for `bytemuck::cast_vec` was stricter than we would assume. So are we sure we made _all_ required checks _correctly_ in `from_u8_to_slice`? I'd better rely on `bytemuck` for such things.

Also, it literally says "deprecated, use `bytemuck` or `zerocopy`". 🤷‍♀️

ffuugoo: Anyway, I mark `cast_slice` as `unsafe` and then try to change it from `Copy` to `bytemuck::Pod` to satisfy requirements.

ffuugoo: Fixed.

lib/common/common/src/universal_io/mmap.rs
ffuugoo: Fixed.

lib/segment/src/vector_storage/quantized/quantized_multivector_storage.rs
```diff
@@ -18,7 +18,18 @@ use crate::types::{MultiVectorComparator, MultiVectorConfig};
...
+    Clone,
+    Debug,
+    Default,
+    PartialEq,
+    Deserialize,
+    Serialize,
+    bytemuck::Pod,
+    bytemuck::Zeroable,
+)]
+#[repr(C)]
```
ffuugoo, line 32: We did not have `#[repr(C)]` before, so I'll check if it breaks anything in a second.

ffuugoo: Test:

```rust
    use std::slice;

    let offset = MultivectorOffset {
        start: 0xdead,
        count: 0xbeef,
    };

    let fields: &[PointOffsetType] = unsafe {
        slice::from_raw_parts(
            &offset as *const _ as *const _,
            size_of_val(&offset) / size_of::<PointOffsetType>(),
        )
    };

    print!("[");
    for field in fields {
        print!(" 0x{field:04x}");
    }
    println!(" ]");
```

Prints `0xdead 0xbeef` with and without `#[repr(C)]`, so I guess we're safe

PR #8501, opened 2026-03-25
[UIO, gridstore] generic `Bitmask` storage

coszio:
Refactors gridstore's `Bitmask` to use a generic backend for interacting with storage. Particularly, it does it for `BitmaskGaps` slice, and for the main bitslice itself. 

In doing so, I also moved the `stored_bitslice.rs` from `segment` into `common`


lib/gridstore/src/bitmask/gaps.rs
```diff
@@ -75,121 +74,148 @@ impl RegionGaps {
...
+            advice: Some(AdviceSetting::Advice(Advice::Normal)),
+        };
+        self.slice_store = S::open(&self.path, options)?;
 
-        debug_assert_eq!(self.mmap_slice[prev_len..].len(), iter.len());
+        debug_assert_eq!(self.len()? - prev_len, data.len());
 
-        self.mmap_slice[prev_len..].fill_with(|| iter.next().unwrap());
+        let byte_offset = (prev_len * size_of::<RegionGaps>()) as u64;
+        self.slice_store.write(byte_offset, &data)?;
```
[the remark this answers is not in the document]
coszio: This is just a refactor, but checking if the bitslice and the gaps structures are consistent is a good idea for a follow-up PR

lib/gridstore/src/bitmask/gaps.rs
```diff
@@ -75,121 +74,151 @@ impl RegionGaps {
...
+        Ok(self.slice_store.read_whole()?)
     }
 
     /// Find a gap in the bitmask that is large enough to fit `num_blocks` blocks.
     /// Returns the range of regions where the gap is.
-    pub fn find_fitting_gap(&self, num_blocks: u32) -> Option<Range<RegionId>> {
-        if self.mmap_slice.len() == 1 {
-            return if self.get(0).unwrap().max as usize >= num_blocks as usize {
+    pub fn find_fitting_gap(&self, num_blocks: u32) -> Result<Option<Range<RegionId>>> {
+        let slice = self.as_slice()?;
```
coszio, line 214: Note: 

This reads the entire thing every time, which is potentially expensive if it is not cached in memory. Otherwise, this structure is expected to be quite small, so it should always fit in memory

lib/segment/src/index/field_index/geo_index/mmap_geo_index.rs
```diff
@@ -240,7 +241,7 @@ impl MmapGeoMapIndex {
         };
         let point_to_values = StoredPointToValues::open(path, true)?;
 
-        let deleted = MmapBitSlice::open(
+        let deleted = <StoredBitSlice<MmapFile>>::open(
```
timvisee: Is it formatted with extra brackets for a specific reason?

```suggestion
        let deleted = StoredBitSlice<MmapFile>::open(
```

lib/gridstore/src/bitmask/gaps.rs
timvisee: What is the alternative, some kind of iterative/batching approach?

If my calculations are correct:
- 1 GB stored = 6KB read here
- 1 TB stored = 6MB read here

timvisee (approved):

lib/gridstore/src/bitmask/gaps.rs
```diff
@@ -74,121 +74,151 @@ impl RegionGaps {
...
-        &mut self.mmap_slice[idx]
+    pub fn set(&mut self, idx: usize, value: RegionGaps) -> Result<()> {
+        let byte_offset = (idx * size_of::<RegionGaps>()) as u64;
+        self.slice_store.write(byte_offset, &[value])?;
+        Ok(())
     }
 
-    pub fn as_slice(&self) -> &[RegionGaps] {
-        &self.mmap_slice
+    pub fn as_slice(&self) -> Result<Cow<'_, [RegionGaps]>> {
```
xzfc, line 207: Naming nit: I'd rename this method since `as_*` implies a cheap conversion.

https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

I'd rename it into `read_all` or something else.

lib/gridstore/src/bitmask/gaps.rs
xzfc: (reply to https://github.com/qdrant/qdrant/pull/8501#discussion_r3101389521)

So, ~16 KiB for a segment of 11M points. (assuming each point has a payload of 512 bytes)

I think a viable alternative is to keep the whole `BitmaskGaps` in the RAM.

lib/gridstore/src/bitmask/mod.rs
```diff
@@ -413,8 +427,9 @@ impl Bitmask {
 
             let gaps = Self::calculate_gaps(bitslice, self.config.region_size_blocks);
 
-            *self.regions_gaps.get_mut(region_id) = gaps;
+            self.regions_gaps.set(region_id, gaps)?;
```
xzfc, line 430: Performance: we can use `UniversalWrite::write_batch` here instead of calling `UniversalWrite::write` in a loop.

---

Or (maybe it's complicated/outside of scope of this PR), this can be merged with `set_ascending_bits_batch` so a single call `write_multi` can be used in `mark_blocks_batch` to write both `bitmask.dat` and `gaps.dat`.

lib/gridstore/src/bitmask/gaps.rs
coszio: That's probably better, I'll add it to the backlog

lib/segment/src/index/field_index/geo_index/mmap_geo_index.rs
coszio: syntax is not correct without them, but anyway, it has been switched to type alias

lib/gridstore/src/bitmask/gaps.rs
coszio: https://github.com/qdrant/qdrant/pull/8501/commits/28d46fcb50d075e556a47832222717cc270eacf9

lib/gridstore/src/bitmask/mod.rs
coszio: Added to #8630

PR #8502, opened 2026-03-25
Test resharding up/down with deferred points

agourlay:
Add integration test verifying that deferred points survive resharding (both up and down).

Deferred points are invisible to reads until the segment is optimized. 

The test creates deferred points by inserting with optimizers disabled, then runs the full resharding lifecycle with the optimizer enabled concurrently (required for wait=true on migration batches).

 After resharding completes and optimization resolves the deferred points, all original points must be present.

  - Down (3→2 shards): the removed shard's data is permanently deleted, so any deferred points not transferred are irrecoverably lost. Verifies collection-level count and that each surviving shard grew from
  received migrations.
  - Up (3→4 shards): verifies the new shard received the correct subset of points from source shards via shard-local counts with hash ring filtering.

### DIFF
--- a/tests/consensus_tests/test_resharding_deferred.py
+++ b/tests/consensus_tests/test_resharding_deferred.py
@@ -294,18 +294,20 @@ def test_resharding_transfer_deferred_points(tmp_path: pathlib.Path, direction: 
         assert_http_ok(resp)
         time.sleep(1)
 
-        # Clean up source shards: delete points that were migrated to the new
-        # shard and no longer belong on the source under the new hash ring.
+    # Complete resharding: commit hash rings and finish.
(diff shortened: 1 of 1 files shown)


tests/consensus_tests/test_resharding_deferred.py
```diff
@@ -0,0 +1,386 @@
...
+                }
+            },
+        )
+        assert_http_ok(resp)
+        time.sleep(1)
+
+    # Complete resharding: commit hash rings and finish.
+    resp = commit_read_hashring(peer_api_uris[0])
+    assert_http_ok(resp)
+
```
timvisee: Resharding up should also do a delete stage in between here.

PR #8507, opened 2026-03-26
Migrate geo_index to UIO

xzfc:


lib/common/common/src/universal_io/read.rs
```diff
@@ -37,10 +37,40 @@ pub trait UniversalRead<T: Copy + 'static>: UniversalReadFileOps {
...
-    ) -> impl Iterator<Item = Result<(Meta, Cow<'_, [T]>)>> {
-        ranges
+    ) -> Result<impl Iterator<Item = Result<(Meta, Cow<'_, [T]>)>>> {
+        Ok(ranges
             .into_iter()
-            .map(move |(meta, range)| self.read::<P>(range).map(|data| (meta, data)))
+            .map(move |(meta, range)| self.read::<P>(range).map(|data| (meta, data))))
+    }
+
+    fn read_iter_autochunks(
```
coszio: It was not clear by just the name that the input ranges can contain many elements.
```suggestion
    /// Iterates over the provided ranges, reading a few elements at a time, 
    /// but returning in a per-element basis.
    /// 
    /// The provided ranges can span many elements, they will be auto-chunked 
    /// by this function to ensure each read is not too big. The argument is an 
    /// iterator just so that it is possible to read from non-contiguous regions in 
    /// the same output iterator.
    fn read_iter_autobatched(
```

lib/segment/src/index/field_index/geo_index/immutable_geo_index.rs
```diff
@@ -47,43 +49,46 @@ pub struct ImmutableGeoMapIndex {
 }
 
 enum Storage {
-    Mmap(Box<MmapGeoMapIndex>),
+    Mmap(Box<MmapGeoMapIndex<MmapFile>>),
```
coszio: We can rename the structure to `StoredGeoMapIndex` since it is now generic

lib/segment/src/index/field_index/geo_index/mmap_geo_index.rs
```diff
@@ -403,33 +433,116 @@ impl MmapGeoMapIndex {
...
+        // Step 2: Read entries in chunks starting from `start` until we stumble
+        // into a chunk with an `end` entry.
+        //
+        //     start                                 end
+        // ─────┬┴──────┬───────┬───────┬───────┬─────┴─┬───────┬───────┬───────
+        // …<<<<│.......│.......│.......│.......│.....>>│>>>>>>>│>>>>>>>│>>>>>>…
+        // ─────┴── 0 ──┴── 1 ──┴── 2 ──┴── 3 ──┴── 4 ──┴── 5 ──┴── 6 ──┴── 7 ──
+        //
+        // The problem is that UniversalRead iterator gives us chunks in an
+        // arbitrary order, and we don't know where `end` is.
```
coszio, line 472: > UniversalRead iterator gives us chunks in an arbitrary order

The only case when this happens is when using io-uring, right? I think I have an alternative way of fixing this. Will make a PR about it.

generall: @coszio please don't spend too much time on it, unless you have really good idea.

coszio: This is the suggestion: https://github.com/qdrant/qdrant/pull/8667

PR #8508, opened 2026-03-26
Add TypedStorage wrapper

xzfc:
This PR introduces `TypedStorage` wrapper.

This wrapper lets you specify the type `T` only once in the storage type, and not on every callsite:
```rust
// before
let file = <IoUringFile as UniversalRead<u64>>::open(&path, opts)?;
let value = <IoUringFile as UniversalRead<u64>>::read::<Sequential>(&file, range)?;
let len = <IoUringFile as UniversalRead<u64>>::len(&file)?;

// after
let file = TypedStorage::<IoUringFile, u64>::open(&path, opts)?;
let value = file.read::<Sequential>(range)?;
let len = file.len()?;
```

Usually, it's not required as the Rust compiler is able to deduce the type `T` based on `where` clauses. But it helps when there are more than one type is used within the same structure, e.g.:
```rust
struct Storage<S> {
    counts_per_hash: TypedStorage<S, Counts>,
    points_map: TypedStorage<S, PointKeyValue>,
    points_map_ids: TypedStorage<S, PointOffsetType>,
    point_to_values: StoredPointToValues<GeoPoint, S>,
    deleted: MmapBitSliceBufferedUpdateWrapper,
}
```

Unlike `MmapUniversal<T>` that removed in #8493/#8505, this wrapper can work with any UIO implementation.

### DIFF
--- a/lib/common/common/src/universal_io/wrappers/typed.rs
+++ b/lib/common/common/src/universal_io/wrappers/typed.rs
@@ -10,7 +10,13 @@ use super::super::{
 };
 use crate::generic_consts::AccessPattern;
 
-/// A wrapper around [`UniversalRead`] / [`UniversalWrite`] that binds `T` to a specific type.
+/// A wrapper around [`UniversalRead`]/[`UniversalWrite`] that binds `T` to a
+/// specific type.
+///
+/// This wrapper is not needed for code with a single universal io trait bound,
+/// (e.g. `where S: UniversalRead<f32>`), but it helps the compiler to
+/// distinguish when more than one bound is used, e.g.
+/// `where S: UniversalRead<f32> + UniversalRead<PointOffsetType> + …`.
 #[derive(Debug, TransparentWrapper)]
 #[repr(transparent)]
 #[transparent(S)]


ffuugoo:
> `<IoUringFile as UniversalRead<u64>>`

Propaganda piece.

```rust
let bytes: Cow<[u8]> = file.read::<Sequential>()?;
file.read(ranges, |_, floats: &[f32]| {
    Ok(())
})
```

ffuugoo:
Another counter-point, the only place where you _need_ multiple `UniversalRead<T>` (on the same type `S`), is, well, where you'll be reading one-or-the-other _interchangeably_.

Like, you'd literally

```rust
let header: HeaderT = file.read(HEADER_RANGE)?;

file.read_batch(items_ranges(header), |_, items: &[ItemT]| {
    // do something
})?;
```

So it _kinda_ does not make sense to put *owned* `UniversalRead` into `TypedStorage`. Because if you put it into `TypedStorage<T1>`, you'd need to _take it out_ and put into `TypedStorage<T2>` if you need to read `T2`.

It only makes sense if `TypedStorage<T>` takes _reference_ to `UniversalRead`.

ffuugoo:
Ok, I see reasonable use-case for this now. Fair. 👍

```rust
struct ManyFiles<S: UniversalRead<Headers> + UniversalRead<Items>> {
    headers: TypedStorage<S, Header>,
    items: TypedStorage<S, Items>,
}
```

(Sorry, I can't read.)

But `<S as UniversalRead<T>>` is still a propaganda piece. 🙄

PR #8512, opened 2026-03-26
Deferred points id is owned by ID tracker

IvanPleshkov:
This PR moves and ownership from `Segment` into `PointsMapping`. The problem with the previous approach is duplication, where we have to store deferred point id in `SparseVectorIndex`.

This PR does not change logic ownership - we still check whether point is deferred on segment level, this change is the next step of refactoring. Here, it's not included to reduce review complexity.

### DIFF
--- a/lib/segment/src/segment/tests.rs
+++ b/lib/segment/src/segment/tests.rs
@@ -30,7 +30,6 @@ use crate::entry::entry_point::{
 };
 use crate::entry::{SnapshotEntry as _, StorageSegmentEntry as _};
 use crate::id_tracker::IdTracker;
-use crate::id_tracker::id_tracker_base::calculate_deleted_deferred_count;
 use crate::index::sparse_index::sparse_index_config::{SparseIndexConfig, SparseIndexType};
 use crate::json_path::JsonPath;
 use crate::segment_constructor::simple_segment_constructor::{
@@ -1471,13 +1470,7 @@ fn test_deleted_deferred_point_count() {
                 n_deferred.checked_sub(deleted_count).unwrap()
             );
             assert_eq!(
-                {
-                    let id_tracker = segment.id_tracker.borrow();
-                    calculate_deleted_deferred_count(
-                        &*id_tracker,
-                        id_tracker.deferred_internal_id().unwrap(),
-                    )
-                },
+                segment.id_tracker.borrow().deferred_deleted_count(),
                 deleted_count,
             );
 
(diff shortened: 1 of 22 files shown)


lib/segment/src/segment/tests.rs
```diff
@@ -1423,22 +1419,17 @@ fn assert_deferred_points_excluded<F, R, T>(
...
-                    drop(segment);
-                    dir.close().unwrap();
-                    continue;
-                }
-
-                let dir = Builder::new().prefix("segment_dir_2").tempdir().unwrap();
-                segment = create_deferred_segment(&dir, 5, N_POINTS + n_deferred, 0);
-            } else {
-                segment.deferred_point_status = None;
+            // Create a segment with all points visible (none deferred) and search again.
```
JojiiOfficial, line 1422: What exactly is the motivation for this change? 
The `create_deferred_segmen` function is quite slow, especially on CI. Therefore we might try to manually disable `deferred_point_status` if possible.

IvanPleshkov: The motivation here is removing deferred point status setter. We apply deferred id once during construction and don't change it again. It's the current behaviour, and only this test tries to change the deferred status in runtime. To make the test green, I have two segments: with and without deferred points, because we previously changed deferred id here just to disable it for a single request

JojiiOfficial: > It's the current behaviour, and only this test tries to change the deferred status in runtime

Yes because setting the `deferred_point_status` to `None` results in the same segment than recreating it. Recreating takes quite a lot of time so we did this ugly hack. Do you suggest to ignore CI time here and just rebuild?

IvanPleshkov: If it affects CI time, I can instead return the setter back, but with `#[cfg(feature = "testing")]` to allow it only for tests

JojiiOfficial: Yes It's just about CI 👍🏻 
I also like the `feature="testing"` flag!

timvisee (commented):
Great idea. Since deferred points involves full points and _affects_ point IDs this is a better place.

JojiiOfficial (changes requested):
We actually can now use the value from `self` and don't need the second parameter. Therefore I suggest to remove `deferred_internal_id`.

https://github.com/qdrant/qdrant/blob/5c74b173b5c0375844f47ebf63fcc9fc1dadf252/lib/segment/src/id_tracker/id_tracker_base.rs#L268

agourlay:
superseded by https://github.com/qdrant/qdrant/pull/9062

PR #8518, opened 2026-03-26
[UIO] migrate `SliceBufferedUpdateWrapper`

coszio:
Moves `MmapSliceBufferedUpdateWrapper` into `universal_io::wrapper` module. It does not directly implement the universal traits, as it is only meant to be used as buffered write-only storage.

For updates, to make batching (allegedly) more efficient, it coalesces consecutive writes into a single one, and then batches all of them together.

It also changes the callsites (it was only used in `ImmutableIdTracker`) to use this new type.


lib/common/common/src/universal_io/wrappers/buffered_update.rs
```diff
@@ -0,0 +1,141 @@
...
+/// flushed manually.
+/// This expects the underlying storage not to grow in size.
+///
+/// WARN: this structure is expected to be write-only.
+#[derive(Debug)]
+pub struct SliceBufferedUpdateWrapper<S: UniversalWrite<T>, T: Copy>
+where
+    T: 'static,
+{
+    slice: Arc<RwLock<TypedStorage<S, T>>>,
```
xzfc: Nit: no need for `TypedStorage` here:
```suggestion
    slice: Arc<RwLock<S>>,
```

Reason: `TypedStorage` solves this particular problem. Suppose you have multiple bounds, like `S: UniversalWrite<T1> + UniversalWrite<T2> + UniversalWrite<T3> + …`. Having multiple bounds at once makes calls ambiguous (so type annotations `T1`/`T2`/`T3`/… are required to decide which one to use). The `TypedStorage` resolves that ambiguity by "hiding" all bounds except only one (e.g. `S: UniversalWrite<T2>`).

But `SliceBufferedUpdateWrapper` has only one trait bound `S: UniversalWrite<T>`, so there are no ambiguity to resolve by `TypedStorage`. It doesn't hurt, but it doesn't help either.

lib/common/common/src/universal_io/wrappers/buffered_update.rs
```diff
@@ -0,0 +1,141 @@
...
+            self.len
+        );
+        self.pending_updates.lock().insert(index, value);
+    }
+}
+
+impl<S: UniversalWrite<T> + Send + Sync + 'static, T: Copy> SliceBufferedUpdateWrapper<S, T>
+where
+    T: 'static + Sync + Send + Clone + PartialEq,
+{
```
xzfc: Code style nit: keep all constraints together, e.g.:

```rust
impl<S, T> SliceBufferedUpdateWrapper<S, T>
where
    S: UniversalWrite<T> + Send + Sync + 'static,
    T: Copy + Clone + Sync + Send + PartialEq + 'static,
{
```

or


```rust
impl<
    S: UniversalWrite<T> + Send + Sync + 'static,
    T: Copy + Clone + Sync + Send + PartialEq + 'static,
> SliceBufferedUpdateWrapper<S, T> {
```

lib/common/common/src/universal_io/wrappers/buffered_update.rs
```diff
@@ -0,0 +1,141 @@
...
+
+                    (byte_start, values)
+                })
+                .collect_vec();
+            // Unfortunately, we need to collect again so that the values are taken as a slice for write_batch
+            let offset_data_refs: Vec<(u64, &[T])> = offset_data
+                .iter()
+                .map(|(offset, values)| (*offset, values.as_slice()))
+                .collect();
+            slice_guard.write_batch(offset_data_refs)?;
```
xzfc: Performance: use $$2$$ allocations instead of $$3 + N_{chunks}$$ allocations:
```suggestion
            let mut items: Vec<(PointOffsetType, T)> =
                updates.iter().map(|(k, v)| (*k, *v)).collect();
            items.sort_by_key(|(index, _)| *index);

            let all_values: Vec<T> = items.iter().map(|(_, value)| *value).collect();
            let mut remaining_values: &[T] = &all_values[..];

            let it = items.chunk_by(|(a, _), (b, _)| *a + 1 == *b).map(|chunk| {
                let first: PointOffsetType = chunk.first().unwrap().0;
                let byte_start = u64::from(first) * size_of::<T>() as u64;
                let chunk_values = remaining_values.split_off(..chunk.len()).unwrap();
                (byte_start, chunk_values)
            });
            slice_guard.write_batch(it)?;
```

lib/common/common/src/universal_io/wrappers/buffered_update.rs
coszio: Yes, it is not strictly necessary, but it helps for making the `open` operations a lil bit more ergonomic, that's the entire reason why I used it.

lib/common/common/src/universal_io/wrappers/buffered_update.rs
coszio: Thanks, that's so much better

lib/common/common/src/universal_io/wrappers/buffered_update.rs
xzfc: I've pushed cee0d71bb477347205874ef28e9cfd6277183bb0 to show what I mean.

xzfc (approved):

lib/common/common/src/universal_io/wrappers/buffered_update.rs
```diff
@@ -66,25 +66,40 @@ where
...
+            // Coalesce contiguous updates into the same write
+            let mut items: Vec<(PointOffsetType, T)> =
+                updates.iter().map(|(k, v)| (*k, *v)).collect();
+            items.sort_by_key(|(index, _)| *index);
+
+            let all_values: Vec<T> = items.iter().map(|(_, value)| *value).collect();
+
+            // Batch writes
+            let mut remaining_values: &[T] = &all_values[..];
+            let it = items.chunk_by(|(a, _), (b, _)| *a + 1 == *b).map(|chunk| {
```
timvisee, line 92: Woah. This `chunk_by` trick to glue consecutive items together is smart 🧠

timvisee (approved):

lib/common/common/src/universal_io/wrappers/buffered_update.rs
coszio: It was @xzfc suggestion, actually

lib/common/common/src/universal_io/wrappers/buffered_update.rs
coszio: Ah, nice, thanks!

PR #8525, opened 2026-03-27
Atomic bitvec for deleted flags  

monoid:
Using own `AtomicBitVec` allows to control atomic ordering at bit load and store operations.

The `AtomicBitVec` was generated by Claude Code with the following prompt:

> I need a custom bitvector type with underlying atomics (using `AtomicU64`) for deletion bitmask in @lib/segment/src/id_tracker .
> 
> It doesn't need all the functionality of the `bitvec` that we use now, but it certaintly needs:
> + `resize(&mut self, size: usize, filler: bool)`
> + `get_with_ordering(&self, idx: usize, ordering: std::atomic::Ordering) -> Option<bool>`
> + `get(&self, idx: usize) -> Option<bool>` (ordering is Aquire)
> + `replace_concurrent_with_ordering(&self, idx: usize, value: bool, ordering: Ordering) -> bool`
> + `replace_concurrent(&self, idx: usize, value: bool) -> bool` (returns previous value, Ordering::AcqRel)
> + iteration over bits, with ordering and with some default ordering
> + service methods like `len(&self)`, `is_empty(&self)`, etc.
> + methods to get reference to the underlying `&[AtomicU64]
> + additional atomic field for popcount. It seems that the counter may
  occasionally underflow, so I suggest using an `AtomicI64`, replacing occasional
  short-lived negative values with 0.
> + the number of bits in the bit vec shouldn't be more than 2^{32}, so `AtomicI64` is enough for the popcount,
>  and for the lengths `usize` on 64-bit platform is enough.
> + factory method for creating the bitvec from `&[u64]`
> + the bit ordering must be `Lsb0`, or you may parametrise the type with `bitvec::order::BitOrder`
> 
> Please, implement it in the `atomic_bitmask` module in the
> @lib/common/common/src/ . Write a testsuite in the same file, including
> concurrent test for replacing (see below).
> 
> I suggest implementing the concurrent test for `replace_concurrent` like this:
> starting with a bitmask set to all false, two (or more) threads replace all the
> bits to true in a unique random order. If returned value is false, increment a
> common counter. If it is true, continue. This counter should be equal to the
> number of bits.
> 
> I would like you to implementing other tests too.
> 
> Then use the new type in the id_tracker.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


lib/segment/src/id_tracker/point_mappings.rs
```diff
@@ -133,7 +134,7 @@ impl PointMappings {
             .unique()
             .take(max_internal)
             .filter_map(move |i| {
-                if self.deleted[i] {
+                if self.deleted.get(i) {
                     None
                 } else {
                     Some((self.internal_to_external[i], i as PointOffsetType))
```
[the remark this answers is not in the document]
monoid: It looks suspicious, but it is consistent with the original behavior.

coszio:
Considering the api surface requested, I think we can get away with just a wrapper type over BitSlice/BitVec of AtomicU64. Internally, everything is `Relaxed` ordering, but we could potentially ensure right ordering by fencing, such as:
```rust
// when storing
fence(Ordering::Release);
bitslice.set(i, true);

// when loading
let is_set = bitslice.get(i);
fence(Ordering::Acquire)
```

Notice the order of release and aquire fences is counter-intuitive if you compare it with a lock, but it's how it's supposed to be, see https://doc.rust-lang.org/std/sync/atomic/fn.fence.html#fence---fence

I think with this, we won't have to do the indexing operations ourselves, and will simplify implementation.

wdyt about this approach?

monoid:
> wdyt about this approach?

Well, it would greatly simplify the implementation, but the barriers are essentially undistriminative, so that's too much synchronization on every operation.  But it worth trying and benchmarking.

monoid:
> wdyt about this approach?

There's problem, though: `bitvec` doesn't provide `replace_concurrent()`, only `set_concurrent()`.  Without this method, it may be difficult to provide idempotency.

lib/segment/src/id_tracker/compressed/compressed_point_mappings.rs
```diff
@@ -41,13 +42,19 @@ pub struct CompressedPointMappings {
...
+        deleted_bv: BitVec,
         internal_to_external: CompressedInternalToExternal,
         external_to_internal: CompressedExternalToInternal,
     ) -> Self {
         // Resize deleted to have the same number of elements as internal_to_external
         // Not all structures we may source this from enforce the same size
-        deleted.resize(internal_to_external.len(), false);
+        let target_len = internal_to_external.len();
+        let raw = deleted_bv.as_raw_slice();
+        let mut deleted = AtomicBitVec::from_slice(raw);
```
JojiiOfficial, line 53: Why are we not using the `to_owned()` function from `AtomicBitSlice` here?

JojiiOfficial (changes requested):

lib/common/common/src/atomic_bitvec.rs
```diff
@@ -0,0 +1,996 @@
...
+/// negative values to `0`.
+// TODO bit ordering parameter like `BitVec` does.
+pub struct AtomicBitVec {
+    data: Vec<AtomicU64>,
+    /// Logical number of bits (may be < `data.len() * BITS_PER_WORD`).
+    len: usize,
+    /// Approximate count of set bits; may transiently be negative.
+    popcount: AtomicI64,
+}
+
```
timvisee, line 325: Do we reimplement `BitVec` here so that we have more control, and allow taking an atomic bitslice for example?

monoid: There are two reasons for that:

1. controlled atomic ordering with saner defaults.
2. `replace_concurrent()` method, the `BitVec` has only `set_aliased()` (aliased = concurrent here).  Knowing the previous value atomically is essential for idempotency.

`bitvec::slice::BitSlice` has the same limitations as the `BitVec` (actually, `BitVec` implements `Deref<Target=BitSlice>`, so most of its methods are from `BitSlice`, and it is its limitations).

Actually, while `bitvec` is stable, it is an abandonware.

timvisee (approved):

coszio:
Too many conflicts now, we can revisit later.

PR #8526, opened 2026-03-27
Use size-based batching for queue proxy WAL transfer

agourlay:
## Summary

 - Replaced the fixed `BATCH_SIZE = 10` operation count limit in queue proxy WAL transfer with size-based batching using a 16 MiB byte budget (`MAX_BATCH_BYTES`) and a 10,000 operation cap (`MAX_BATCH_OPS`)
 - Added `read_with_size()` and `read_range_with_size()` methods to `SerdeWal` that return the serialized byte size of each WAL entry alongside the deserialized record
 - Added unit tests for the new WAL read methods

  ## Motivation

 The queue proxy shard flushes buffered WAL operations to the remote during snapshot-based shard transfers.
 The previous fixed batch size of 10 operations was set without considering that each WAL operation can vary wildly in size; a single upsert operation can itself contain a large batch of high-dimensional vectors, while a delete is tiny.

With a batch size of 10, each batch paid the full gRPC round-trip cost regardless of payload size. Under sustained write pressure, this could cause the transfer to fall behind and never complete, as new writes accumulated faster than they could be drained.

Size-based batching amortizes network overhead by filling each batch up to a byte budget, naturally adapting to the operation mix. The operation count cap bounds memory usage and WAL lock duration for workloads with many small operations.

  ## Test plan

  - [x] Existing unit tests for `SerdeWal::read()` and `read_range()` still pass (behavior preserved via wrapper)
  - [x] New unit tests for `read_with_size()`: correctness, offset reading, empty WAL, equivalence with `read()`
  - [ ] Run existing shard transfer integration tests (`test_shard_snapshot_transfer` and related)
  - [ ] Performance tests on cluster setup

### DIFF
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -42,7 +42,7 @@ use crate::shards::telemetry::LocalShardTelemetry;
 /// Maximum total serialized byte size of a single transfer batch.
 /// Each WAL operation can vary widely in size (a delete vs an upsert of many high-dimensional
 /// vectors), so we use a byte budget rather than a fixed operation count.
-const MAX_BATCH_BYTES: usize = 16 * 1024 * 1024; // 16 MiB
+const MAX_BATCH_BYTES: usize = 64 * 1024 * 1024; // 64 MiB
 
 /// Maximum number of operations in a single transfer batch.
 /// Caps memory usage and WAL lock duration when operations are small.
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -42,7 +42,7 @@ use crate::shards::telemetry::LocalShardTelemetry;
 /// Maximum total serialized byte size of a single transfer batch.
 /// Each WAL operation can vary widely in size (a delete vs an upsert of many high-dimensional
 /// vectors), so we use a byte budget rather than a fixed operation count.
-const MAX_BATCH_BYTES: usize = 64 * 1024 * 1024; // 64 MiB
+const MAX_BATCH_BYTES: usize = 32 * 1024 * 1024; // 32 MiB
 
 /// Maximum number of operations in a single transfer batch.
 /// Caps memory usage and WAL lock duration when operations are small.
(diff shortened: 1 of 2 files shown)


generall:
If the premise is correct, the problem of shard transfer not being able to catch up is more easily reproducable with collection of small vectors than large vectors. It can be benchmarked

timvisee:
> Replaced the fixed BATCH_SIZE = 10 operation count limit in queue proxy WAL transfer with size-based batching using a 16 MiB byte budget (MAX_BATCH_BYTES) and a 10,000 operation cap (MAX_BATCH_OPS)

Maybe we can go bigger. 16 MiB is not a lot for a roundtrip on fast datacenter interconnects. Reading 16 MiB into a memory buffer also isn't a lot. :man_shrugging:

lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -555,25 +561,38 @@ impl Inner {
...
                 })?;
+
+                // Always include at least one operation per batch
+                if !batch.is_empty()
+                    && (batch_bytes + size > MAX_BATCH_BYTES || batch.len() >= MAX_BATCH_OPS)
+                {
+                    break;
+                }
+
+                batch_bytes += size;
```
timvisee: Nit. We can move `batch_bytes += size` before the if-statement and just check `batch_bytes > MAX_BATCH_BYTES`

lib/shard/src/wal.rs
```diff
@@ -134,26 +134,45 @@ impl<R: DeserializeOwned + Serialize> SerdeWal<R> {
...
                 WalError::ReadWalError(format!("Can't read entry {idx} from WAL"))
             })?;
 
+            let size = record_bin.len();
             let record: R = WalRawRecord::deserialize_from(&record_bin)?;
 
-            Ok((idx, record))
+            Ok((idx, size, record))
         })
     }
```
timvisee, line 177: How about we just always use this function to reduce the code footprint?

Calculating the size is cheap because it's a length we already know. In cases we don't use it the compiler will probably get rid of it automatically.

timvisee (approved):
Thanks!

I planned to run some benchmarks for finding the ideal batch size. But it turned out to be a huge task as there's multiple deployment styles, distributions, workloads and of course batch sizes to cover. And so I decided to not waste a lot of time on it.

With a naive test of "just" making a snapshot transfer with fixed load (2000 upserts per sec), this new implementation definitely is faster. The 'sync queued updates' stage took about half the time, being 35s before and now 17 seconds. I understand this isn't a very scientific test, but I'd argue it's good enough to merge.

I do still notice the transfer 'hanging' at the very end. It makes sense because the last batch still uses 'wait_until=wal'. If problems remain in production we may need another change to throttle uploads in this case.

generall (approved):

timvisee (approved):

PR #8529, opened 2026-03-27
Drop RocksDB

xzfc:
This PR removes RocksDB support from the codebase.

Most of this PR is done automatically using `ast-grep` to remove all items under `#[cfg(feature = "rocksdb")]` cargo feature flag. (see "Add rocksdb dropper" commit for the script that did the job).

## Next steps

There are still some work remaining for follow-up PRs:

- Broken tests/benchmarks need to be either fixed or removed.
- Some of the migration logic is no-op now. Again, these should be either removed, or updated to throw some appropriate error.
- Other code cleanup.

These items are marked with `TODO(rocksdb)` / `FIXME(rocksdb)` comments.


lib/segment/src/id_tracker/mutable_id_tracker.rs
```diff
@@ -1440,94 +1436,6 @@ pub(super) mod tests {
...
-
-        // Persist and reload mutable tracker and test again
-        mutable_id_tracker.mapping_flusher()().unwrap();
-        mutable_id_tracker.versions_flusher()().unwrap();
-        drop(mutable_id_tracker);
-        let mutable_id_tracker = MutableIdTracker::open(segment_dir.path()).unwrap();
-
-        check_trackers(&simple_id_tracker, &mutable_id_tracker);
-    }
-
```
timvisee, line 1530: I vote to keep this, but do a congruence test between the mutable (Gridstore) and volatile/in-memory ID trackers.

src/common/telemetry_ops/app_telemetry.rs
```diff
@@ -87,7 +87,7 @@ impl AppBuildTelemetry {
                 service_debug_feature: cfg!(feature = "service_debug"),
                 recovery_mode: settings.storage.recovery_mode.is_some(),
                 gpu: cfg!(feature = "gpu"),
-                rocksdb: cfg!(feature = "rocksdb"),
+                rocksdb: false,
```
timvisee, line 90: We can drop this one:

```suggestion
```

src/common/telemetry_ops/app_telemetry.rs
timvisee, line 31:
```suggestion
```

lib/segment/src/types.rs
timvisee, line 1729: We can drop this single variant enum if you ask me. I believe we have a similar situation in other storages.

It'd probalby best to do that in a separate PR though, to not bloat this one even more.

lib/segment/src/segment/segment_ops.rs
```diff
@@ -760,24 +760,6 @@ fn restore_snapshot_in_place(snapshot_path: &Path) -> OperationResult<()> {
...
-            crate::rocksdb_backup::restore(&db_backup_path, segment_path)?;
-            fs::remove_dir_all(&db_backup_path)?;
-        }
-
-        let payload_index_db_backup = segment_path.join(PAYLOAD_DB_BACKUP_PATH);
-        if payload_index_db_backup.is_dir() {
-            StructPayloadIndex::restore_database_snapshot(&payload_index_db_backup, segment_path)?;
-            fs::remove_dir_all(&payload_index_db_backup)?;
-        }
-    }
```
timvisee, line 779: Should we print a warning here if files like this are still found in the snapshot?

If we don't report anything, the files will be silently lost.

src/common/telemetry_ops/app_telemetry.rs
xzfc: It's a non-optional field in the API, and removing it breaks python client:

```console
>>> c = qdrant_client.QdrantClient()
>>> c.http.service_api.telemetry(details_level=1).result.app
…
qdrant_client.http.exceptions.ResponseHandlingException: 1 validation error for ParsingModel[InlineResponse2002] (for parse_as_type)
obj.result.app.features.rocksdb
  Field required [type=missing, input_value={'debug': True, 'service_...False, 'staging': False}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
```

timvisee: Interesting! For the sake of not breaking stuff we can keep it.

lib/segment/src/segment/segment_ops.rs
xzfc: Done in 61e3b51bf5be5906849fc26846fd904b9d9adefc.

generall (approved):

PR #8541, opened 2026-03-30
fix: include boundary points in GeoBoundingBox check

### LINKED ISSUE #1188: GeoBoundingBox boundary point is matched in in-memory backend but excluded by Rust server
## Current Behavior
`GeoBoundingBox` behaves inconsistently between the in-memory backend and the Rust server backend.
Using the same Python client, the same inserted point, and the same bounding box filter:
- in-memory backend returns `[1]`
- Rust server returns `[]`

Reproduction output:
--- Qdrant GeoBoundingBox Consistency Test ---
[Memory Emulator] Actual IDs returned: [1]
[Rust Server (localhost)] Actual IDs returned: []

## Steps to Reproduce

```python
from qdrant_client import QdrantClient
from qdrant_client.models import (
    PointStruct, Filter, FieldCondition, GeoBoundingBox, GeoPoint, VectorParams, Distance
)

def run_test(client_type, client):
    collection = "geo_boundary_test"
    
    if client.collection_exists(collection):
        client.delete_collection(collection)
        
    client.create_collection(
        collection_name=collection, 
        vectors_config=VectorParams(size=2, distance=Distance.DOT)
    )

    client.upsert(
        collection_name=collection,
        points=[
            PointStruct(id=1, vector=[0.1, 0.1], payload={"location_geo": {"lat": 90.0, "lon": 180.0}})
        ]
    )

    bbox = GeoBoundingBox(
        top_left=GeoPoint(lon=158.75, lat=90.00),
        bottom_right=GeoPoint(lon=180.00, lat=69.40)
    )
    
    geo_filter = Filter(must=[FieldCondition(key="location_geo", geo_bounding_box=bbox)])
    results = client.scroll(collection_name=collection, scroll_filter=geo_filter)[0]
    
    print(f"[{client_type}] Actual IDs returned: {[p.id for p in results]}")

if __name__ == "__main__":
    print("--- Qdrant GeoBoundingBox Consistency Test ---")
    
    client_memory = QdrantClient(":memory:")
    run_test("Memory Emulator", client_memory)
    
    try:
        client_server = QdrantClient(host="localhost", port=6333)
        run_test("Rust Server (localhost)", client_server)
    except Exception as e:
        print(f"[Rust Server (localhost)] Connection failed. Please ensure Qdrant is running on port 6333.")
```

## Expected Behavior
Both backends should behave the same.
If a point on the bounding-box boundary is considered inside the box, both should return [1].
If boundary points are intentionally excluded, both should return [].
The current inconsistency between in-memory and server backends is unexpected.

## Possible Solution
Align boundary handling for GeoBoundingBox between the in-memory backend and the Rust server backend.

## Context (Environment)
I found this while comparing Qdrant filter behavior between local testing and the real server backend.
This inconsistency makes geo filter behavior difficult to reason about and test reliably.

## Detailed Description
This looks like a boundary-handling issue for geo bounding boxes, specifically for points located exactly on the box edge/corner.
The same script produces different results depending on the backend:
QdrantClient(":memory:") matches the point
QdrantClient(host="localhost", port=6333) does not
Because the client-side filter construction is identical in both cases, this appears to be a backend inconsistency.

## Possible Implementation
Review how GeoBoundingBox handles points exactly on the boundary in the Rust server implementation and align it with the in-memory implementation.

hkc5:
Fixes qdrant/qdrant-client#1188

## Problem

GeoBoundingBox behaves inconsistently between the in-memory backend (Python) and the Rust server backend. Points exactly on the boundary of a GeoBoundingBox are:
- **Included** by the Python in-memory backend
- **Excluded** by the Rust server backend

## Root Cause

The Rust implementation of "GeoBoundingBox::check_point" uses strict inequalities (< and >), while the Python implementation uses inclusive inequalities (<= and >=).

### Python implementation (in-memory):
```python
longitude_condition = condition.top_left.lon <= lon <= condition.bottom_right.lon
latitude_condition = condition.top_left.lat >= lat >= condition.bottom_right.lat
```

### Old Rust implementation (server):
```rust
self.top_left.lon < point.lon && point.lon < self.bottom_right.lon
self.bottom_right.lat < point.lat && point.lat < self.top_left.lat
```

## Fix

Changed strict inequalities to inclusive inequalities in the Rust implementation to match the Python behavior:

```rust
self.top_left.lon <= point.lon && point.lon <= self.bottom_right.lon
self.bottom_right.lat <= point.lat && point.lat <= self.top_left.lat
```

This ensures consistent behavior between both backends - points exactly on the boundary are now included in the match.

### DIFF
--- a/lib/segment/src/types.rs
+++ b/lib/segment/src/types.rs
@@ -2878,12 +2878,12 @@ impl GeoBoundingBox {
     pub fn check_point(&self, point: &GeoPoint) -> bool {
         let longitude_check = if self.top_left.lon > self.bottom_right.lon {
             // Handle antimeridian crossing
-            point.lon > self.top_left.lon || point.lon < self.bottom_right.lon
+            point.lon >= self.top_left.lon || point.lon <= self.bottom_right.lon
         } else {
-            self.top_left.lon < point.lon && point.lon < self.bottom_right.lon
+            self.top_left.lon <= point.lon && point.lon <= self.bottom_right.lon
         };
 
-        let latitude_check = self.bottom_right.lat < point.lat && point.lat < self.top_left.lat;
+        let latitude_check = self.bottom_right.lat <= point.lat && point.lat <= self.top_left.lat;
 
         longitude_check && latitude_check
     }


hkc5:
Closing to reduce noise. Will reopen if needed.

PR #8564, opened 2026-03-31
fix: prevent panic when snapshot path contains non-UTF-8 characters

### LINKED ISSUE #8563: fix: prevent panic when snapshot path contains non-UTF-8 characters
fix: prevent panic when snapshot path contains non-UTF-8 characters

The code at `src/snapshots.rs:117` uses `.to_str().unwrap()` which can panic if the file path contains non-UTF-8 characters.

## Problem

When recovering snapshots, the code constructs a path string using:
```rust
snapshot_temp_path.join(snapshot_file).to_str().unwrap()
```

This will `panic!` if the joined path is not valid UTF-8. While rare on Unix systems (which typically use UTF-8 paths), this can occur on Windows systems and with certain locale/encoding configurations.

## Solution

Replace the unsafe `unwrap()` with proper error handling, consistent with how the codebase handles path conversions elsewhere (e.g., `actix/api/snapshot_api.rs:101`, `tonic/auth.rs:59` which use `.and_then(…)` or `.ok()`).

The mapping construction should gracefully handle non-UTF-8 paths by either:
1. Skipping the invalid entry with a warning
2. Using `format!("{:?}", path)` as a fallback which works for any path

## Impact

Low risk - defensive fix for an edge case that could cause process termination.

## Related Files

- `src/snapshots.rs:117`

mango766:
## Problem
The code at `src/snapshots.rs:117` uses `.to_str().unwrap()` which can panic if the file path contains non-UTF-8 characters.

When recovering snapshots, this `unwrap()` will cause the process to terminate immediately if the joined path is not valid UTF-8. While rare on Unix systems (which typically use UTF-8 paths), this can occur on Windows systems and with certain locale/encoding configurations.

## Solution
Replace the unsafe `unwrap()` with `.display()`, which gracefully handles any path encoding.

```diff
- snapshot_temp_path.join(snapshot_file).to_str().unwrap(),
+ snapshot_temp_path.join(snapshot_file).display(),
```

This matches the pattern used elsewhere in the codebase:
- `actix/api/snapshot_api.rs:101`
- `tonic/auth.rs:59`
- `tonic/forwarded.rs:26`

## Changes
- 1 file modified (`src/snapshots.rs`)
- 1 line changed
- No breaking changes
- Defensive fix for an edge case

Fixes #8563

### DIFF
--- a/src/snapshots.rs
+++ b/src/snapshots.rs
@@ -114,7 +114,7 @@ pub fn recover_full_snapshot(
         .map(|(collection_name, snapshot_file)| {
             format!(
                 "{}:{collection_name}",
-                snapshot_temp_path.join(snapshot_file).to_str().unwrap(),
+                snapshot_temp_path.join(snapshot_file).display(),
             )
         })
         .collect();


generall (approved):

timvisee (changes requested):
Thank you for the fix!

Though, could you replace target this PR to the `dev` branch instead? It may be easiest to recreate the PR.

It is described here: https://github.com/qdrant/qdrant/blob/master/docs/CONTRIBUTING.md#we-use-github-flow-so-all-code-changes-happen-through-pull-requests

mango766:
Thanks for the review\! I'll recreate this PR targeting the `dev` branch as mentioned in CONTRIBUTING.md. Sorry for missing that.

timvisee:
Superseded by https://github.com/qdrant/qdrant/pull/8577

PR #8565, opened 2026-03-31
Use `StorageError` funcs

nibblerenush:
Replaced the use of direct `StorageError` creation with special functions for creating `StorageError`


generall (approved):
Less token usage, great!

PR #8587, opened 2026-04-01
Use `OperationError` funcs

nibblerenush:
Replaced the use of direct `OperationError` creation with special functions for creating `OperationError`

### DIFF
--- a/lib/api/src/conversions/vectors.rs
+++ b/lib/api/src/conversions/vectors.rs
@@ -17,13 +17,11 @@ fn convert_to_plain_multi_vector(
 ) -> Result<MultiDenseVector, OperationError> {
     let dim = data.len() / vectors_count;
     if dim * vectors_count != data.len() {
-        return Err(OperationError::ValidationError {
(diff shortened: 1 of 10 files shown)


lib/api/src/conversions/vectors.rs
```diff
@@ -17,13 +17,11 @@ fn convert_to_plain_multi_vector(
...
-            description: format!(
-                "Data length is not divisible by vectors count. Data length: {}, vectors count: {}",
-                data.len(),
-                vectors_count
-            ),
-        });
+        return Err(OperationError::validation_error(format!(
+            "Data length is not divisible by vectors count. Data length: {}, vectors count: {}",
+            data.len(),
+            vectors_count
```
timvisee: We can inline the variable:

```suggestion
            "Data length is not divisible by vectors count. Data length: {}, vectors count: {vectors_count}",
            data.len(),
```

timvisee (approved):
Thanks! 😄

Issue #8580, opened 2026-04-01
Flaky `test_wait_timeout.py::test_wait_timeout_twice`


agourlay:
https://github.com/qdrant/qdrant/actions/runs/23839131814/job/69490019106

```
=================================== FAILURES ===================================
___________________________ test_wait_timeout_twice ____________________________

collection_name = 'test_wait_timeout'

    def test_wait_timeout_twice(collection_name):
        skip_if_no_feature("staging")
        sleep, op = run_parallel(
            {
                "api": "/collections/{collection_name}/debug",
                "method": "POST",
                "path_params": {"collection_name": collection_name},
                "query_params": {"wait": "true", "timeout": 5},
                "body": {"delay": {"duration_sec": 10.0}},
            },
            {
                "api": "/collections/{collection_name}/points",
                "method": "PUT",
                "path_params": {"collection_name": collection_name},
                "query_params": {"wait": "true", "timeout": 1},
                "body": {
                    "points": [
                        {
                            "id": 1,
                            "vector": {
                                "bm25": {
                                    "text": "Lorem ipsum, dolor sit amet.",
                                    "model": "qdrant/bm25",
                                }
                            },
                        }
                    ]
                },
            },
        )
        assert sleep.ok and sleep.json()["result"]["status"] == "wait_timeout"
>       assert op.ok and op.json()["result"]["status"] == "wait_timeout"
E       AssertionError: assert (True and 'completed' == 'wait_timeout'
E        +  where True = <Response [200]>.ok
E         
E         - wait_timeout
E         + completed)

tests/openapi/test_wait_timeout.py:155: AssertionError
```

timvisee:
Fixed in <https://github.com/qdrant/qdrant/pull/8591>

PR #8591, opened 2026-04-01
fix: timeout tests

dancixx:
fix of this nondeterministic issue: https://github.com/qdrant/qdrant/issues/8580


timvisee (approved):
I see we now trigger the barrier after we've initiated the first request, rather than before it

PR #8594, opened 2026-04-02
Immutable storage for mmap numeric index

monoid:
Do not use a deleted mmap-ed storage, instead, reconstruct a bitmask from the index and segment-level deleted mask.  It effectively makes the mmap numeric index storage immutable, and immutable numeric index storage too as it delegates to mmap numeric index.

It would reduce IO at Qdrant serverless instances.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/numeric_index/mmap_numeric_index.rs
+++ b/lib/segment/src/index/field_index/numeric_index/mmap_numeric_index.rs
@@ -10,7 +10,7 @@ use common::fs::{atomic_save_json, clear_disk_cache, read_json};
 use common::mmap;
 use common::mmap::{AdviceSetting, MmapSlice, create_and_ensure_length};
 use common::types::PointOffsetType;
-use common::universal_io::{MmapFile, UniversalRead};
+use common::universal_io::{MmapFile, OpenOptions, UniversalRead};
 use fs_err as fs;
 use memmap2::MmapMut;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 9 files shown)


lib/segment/src/index/field_index/numeric_index/mmap_numeric_index.rs
```diff
@@ -183,11 +162,21 @@ impl<T: Encodable + Numericable + Default + StoredValue> MmapNumericIndex<T> {
...
 
+        let mut deleted = BitVec::repeat(false, point_to_values.len());
+        for idx in 0..point_to_values.len() {
+            let deleted_payload = point_to_values.get_values_count(idx as _)?.unwrap_or(0) == 0;
+            let deleted_point = id_tracker.is_deleted_point(idx as PointOffsetType);
+            if deleted_payload || deleted_point {
+                deleted.set(idx, true);
+            }
+        }
+        let deleted_count = deleted.count_ones();
```
timvisee: What is the cost of this on each startup and/or shard load? Is it measurable on very large deployments?

monoid: That's reading 16*num_of_points bytes sequentially (line 167).  For 4G segment of 256-element f32 vectors, it is 64M of data.  With a bit flag file, it can be reduced to the original 512K of data.

monoid: I've restored the bitmask.

timvisee: We still walk over all points here

lib/segment/src/index/query_optimization/rescore_formula/value_retriever.rs
```diff
@@ -345,10 +346,13 @@ mod tests {
             PayloadStorageEnum::InMemoryPayloadStorage(InMemoryPayloadStorage::default()),
         )));
         let hw_counter = HardwareCounterCell::new();
+        // No deletions in this test — sized to comfortably exceed the
+        // stored deletion bitslice for the few points added below.
+        let deleted_points = BitVec::repeat(false, 64);
```
timvisee, line 351: Can we also test with a bitset that is too small? I don't see it here, at least not in the diff.

timvisee (approved):

PR #8599, opened 2026-04-03
IoUringState: replace HashMap with flat slice

xzfc:
# Replace HashMap with flat slice

```diff
 pub struct IoUringState<T> {
-    requests: ahash::HashMap<
-        RequestId,
-        IoUringRequest<T>,
-    >,
+    requests: Vec<Option<
+       (RequestId, IoUringRequest<T>)
+   >>,
 }
```
(in this PR it's more sophisticated than `Vec<Option<…>>`, but the idea is the same)

Currently, we use `HashMap` to store requests. Performance-wise that means we need to hash `RequestId` on each `io_uring` request/response, and also probably do allocations.
But since `io_uring` has fixed size queue, we can improve the performance by storing requests in a flat fixed-size array instead of `HashMap`:
```
io_uring/8bytes/read_batch_small_sequential
                        time:   [1.2550 µs 1.2566 µs 1.2583 µs]
                        change: [−7.8385% −7.6725% −7.5137%] (p = 0.00 < 0.05)
                        Performance has improved.
io_uring/1KiB/read_batch_small_sequential
                        time:   [1.8207 µs 1.8251 µs 1.8293 µs]
                        change: [−6.0646% −5.7986% −5.5121%] (p = 0.00 < 0.05)
                        Performance has improved.
```

# ~~Make `RequestId` to be a generic parameter~~

(this change removed from this PR; it will go into #8601)

<details>

```diff
-pub struct IoUringState<T> {
+pub struct IoUringState<T, RequestId> {
```

Previously, we were passing `RequestId` into `io_uring`'s `user_data`. In this PR we pass slot number (aka index within `IoUringState::request`).
Thus, now we don't need to check `RequestId` for uniqueness. (collisions won't break the logic)
Thus, now we can make `RequestId` to be generic parameter and let callers decide which `RequestId` to use:
- If order is important, `RequestId = usize`.
- If order is not important, `RequestId = ()`.
- Later, callers can use domain-specific `RequestId`, e.g.:
  ```rust
  // not in this PR
  enum UniversalHashmapRequestId {
      BucketOffset(usize),
      EntryHeader(usize),
      Value(usize),
  }
  ```

</details>

### DIFF
--- a/lib/common/common/src/universal_io/io_uring/runtime.rs
+++ b/lib/common/common/src/universal_io/io_uring/runtime.rs
@@ -1,10 +1,9 @@
-use std::collections::hash_map;
 use std::io;
 use std::mem::MaybeUninit;
 
 use ::io_uring::types::Fd;
 use ::io_uring::{opcode, squeue};
-use ahash::HashMapExt as _;
+use slab::Slab;
 
 use super::*;
 use crate::maybe_uninit;
@@ -17,9 +16,11 @@ pub struct IoUringRuntime<'data, T> {
 
 impl<'data, T> IoUringRuntime<'data, T> {
     pub fn new() -> Result<Self> {
+        let mut io_uring = pool::get_io_uring()?;
+        let capacity = io_uring.submission().capacity();
         let rt = Self {
-            io_uring: pool::get_io_uring()?,
-            state: IoUringState::new(),
+            io_uring,
+            state: IoUringState::with_capacity(capacity),
             in_progress: 0,
         };
 
@@ -100,20 +101,20 @@ impl<'data, T> IoUringRuntime<'data, T> {
         self.io_uring.completion().map(|entry| {
             self.in_progress -= 1;
 
-            let id = entry.user_data();
+            let slot = entry.user_data() as usize;
             let result = entry.result();
 
             if result < 0 {
-                self.state.abort(id);
+                self.state.abort(slot);
 
                 return Err(io_error_context(
                     io::Error::from_raw_os_error(-result),
-                    format!("io_uring operation {id} failed"),
+                    format!("io_uring operation in slot {slot} failed"),
                 ));
             }
 
             let length = result as _;
-            let resp = self.state.finalize(id, length)?;
+            let (id, resp) = self.state.finalize(slot, length)?;
             Ok((id, resp))
         })
     }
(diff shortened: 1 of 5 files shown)


generall:
There are a lot of unsafe and panics, so why not `Vec<Option<…>>`?

lib/common/common/src/universal_io/io_uring/runtime.rs
```diff
@@ -139,14 +142,24 @@ impl<'data, T> Drop for IoUringRuntime<'data, T> {
 }
 
 #[derive(Debug)]
-pub struct IoUringState<'data, T> {
-    requests: ahash::HashMap<RequestId, IoUringRequest<'data, T>>,
+pub struct IoUringState<'data, T, RequestId> {
+    requests: UninitSlice<(RequestId, IoUringRequest<'data, T>)>,
+    /// Safety invariant: all bits above `requests.len()` must be zero,
+    /// so we can skip bounds checks.
+    occupied: BitArray<[u64; 1]>,
```
generall: Let's move this logic and container into a separate file, so unsafe doesn’t leak into IoUring

lib/common/common/src/universal_io/io_uring/runtime.rs
```diff
@@ -139,14 +142,24 @@ impl<'data, T> Drop for IoUringRuntime<'data, T> {
 }
 
 #[derive(Debug)]
-pub struct IoUringState<'data, T> {
-    requests: ahash::HashMap<RequestId, IoUringRequest<'data, T>>,
+pub struct IoUringState<'data, T, RequestId> {
+    requests: UninitSlice<(RequestId, IoUringRequest<'data, T>)>,
```
ffuugoo: I'd just use `Vec`. I don't see what we gain by using boxed slice. inb4: saving `usize` worth of stack space, but come on...

ffuugoo (commented):
Neat idea, though I'm not a huge fan of this overall, because of added complexity. I'd prefer to keep io_uring code "as simple as reasonably possible". 😬

I see that generic `RequestId` _might_ be useful in the future, but not yet convinced it _will_ be. 

For the fixed-sized `State`, I dunno. I'd say if we can "freeze" the capacity of `HashMap` (so there's no realloc), it's fine to just use a `HashMap`.

lib/common/common/src/universal_io/io_uring/runtime.rs
xzfc: It's more about communication/intent, rather than about performance.

`Box<[T]>` means "not supposed to be resized", and `Vec<T>` means "maybe we will resize it, maybe not".

---

Tho, I plan to move this code away. I need experimenting, but it'll be either of these:
- I'll use `slab`, `stable-vec`, `slotmap`, or whatever, as this idea is not new.
- Or move this data structure into separate module like `common::slotmap` to keep these unsafes away from `IoUringState`.
- Or just use `Vec<Option<…>>`.

xzfc:
> Neat idea, though I'm not a huge fan of this overall, because of added complexity. I'd prefer to keep io_uring code "as simple as reasonably possible". 😬

See next PR #8601. Generic `RequestId` removed far more complexity than added. Particularly because now we don't use `RequestId` as hashmap key and don't care about collisions.

xzfc:
Update: replaced my unsafe thingy with `slab::Slab`.
This is not a new dependency as we already (indirectly) depend on it.
It's used in the tokio uring driver for the exactly the same purpose: https://github.com/tokio-rs/tokio/blob/tokio-1.51.0/tokio/src/runtime/io/driver/uring.rs#L17

ffuugoo (approved):
Very nice, thanks!

PR #8601, opened 2026-04-03
UniversalIO: generic RequestId

xzfc:
Follow up for #8599.

Makes `RequestId` to be a generic parameter of `UniversalRead` methods.
Lets callers to decide what to put into `RequestId`.

The second commit is an example when it can be convenient.

Also, this change makes `IoUringReadIter`/`IoUringReadMultiIter` separation unnecessary.
It gives performance improvement for callers that do not need `RequestId`:

```
io_uring/8bytes/read_batch_small_sequential
                        time:   [1.1978 µs 1.1996 µs 1.2016 µs]
                        change: [−5.2882% −5.1465% −5.0014%] (p = 0.00 < 0.05)
                        Performance has improved.
io_uring/1KiB/read_batch_small_sequential
                        time:   [1.7451 µs 1.7492 µs 1.7528 µs]
                        change: [−4.2483% −3.9277% −3.5943%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/lib/gridstore/src/pages.rs
+++ b/lib/gridstore/src/pages.rs
@@ -150,7 +150,12 @@ impl<S: UniversalRead<u8>> Pages<S> {
 
         let (read_ranges, buffer_offsets) = Self::get_page_value_ranges(pointer, config);
 
-        S::read_multi::<P>(self.pages.as_slice(), read_ranges, |idx, _, slice| {
+        let reads = read_ranges
+            .into_iter()
+            .enumerate()
+            .map(|(idx, (page_idx, range))| (idx, &self.pages[page_idx], range));
+
+        S::read_multi::<P, _>(reads, |idx, slice| {
             let offset = buffer_offsets[idx];
             raw_value[offset..offset + slice.len()].write_copy_of_slice(slice);
             Ok(())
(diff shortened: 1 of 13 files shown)


lib/gridstore/src/pages.rs
```diff
@@ -80,22 +76,16 @@ impl<S: UniversalRead<u8>> Pages<S> {
...
-    /// section can be read into or written from the correct position.
+    /// Computes the page ranges for reading or writing a value described by the given
+    /// [`ValuePointer`].
     ///
-    /// Returns a tuple of:
-    /// - `SmallVec<[(FileIndex, ReadRange); 2]>` — the per-page file ranges to access.
-    /// - `SmallVec<[usize; 2]>` — the offset within the value buffer that each page range
-    ///   corresponds to.
+    /// A value may span across two consecutive pages if it starts near the end of a page.
+    /// Returns `(buffer_offset, page_index, range)` entries that together cover the full
```
generall: what exactly `buffer_offset, page_index, range` are?

generall (approved):

ffuugoo (commented):
Needs rebase, but LGTM.

I'd rename `RequestId` into more generic `Meta` (user supplies arbitrary "metadata" with request, not strictly an "id"), but we can do this in a follow-up.

lib/gridstore/src/pages.rs
xzfc: Updated comment: https://github.com/qdrant/qdrant/pull/8601/changes/99dacf3540b00e852f2caa3bd1f9f400d99c84a4

ffuugoo (approved):

PR #8606, opened 2026-04-06
deep memory reporting

generall:
This PR exposes per-collection per-component memory usage:

```
GET /collections/<name>/memory
```

Each component reports following numbers about memory usage:
```
    /// Total bytes stored on disk (file sizes).
    pub disk_bytes: u64,
    /// Non-evictable heap RAM (in-memory data structures not backed by mmap).
    pub ram_bytes: u64,
    /// Evictable RAM: resident file pages from mmap (OS page cache).
    pub cached_bytes: u64,
    /// Bytes that should ideally be cached for best performance.
    /// Sum of file sizes for `Cached` intent files (mmap-accessed data).
    pub expected_cache_bytes: u64,
```

Disk cache is estimated using `mincore`. It opens all files of the component as a mmap and performs syscall to evaluate cache size.

It is planned for the future to re-use already opened mmaps from the pool, so we won't need to spend time opening and closing mmap if it is already opened.

`ram_bytes` is estimated with custom function for each component. It is the least accurate measurement in this API, as we can't take into account many factors: 

- data structures implemented in external packages (e.g. roaring bitmaps)
- allocator overhead
- minor components

This PR provides a manual test, which estimates error by comparing jemalloc reading before and after deleting components. Current results are:

```
====================================================================================================
Component                                   Estimated   Actual Freed        Error    Error %
----------------------------------------------------------------------------------------------------
payload_index/bool_field                    224.16 KB     -489.78 KB    713.94 KB     318.5%
payload_index/float_field                    23.08 MB       25.67 MB      2.59 MB      11.2%
payload_index/int_field                      64.97 MB       66.97 MB      2.01 MB       3.1%
payload_index/text_field                     12.59 MB       12.96 MB    383.46 KB       3.0%
payload_index/geo_field                     216.34 MB      236.04 MB     19.70 MB       9.1%
payload_index/keyword_field                  22.00 MB       23.95 MB      1.94 MB       8.8%
payload_index/datetime_field                 23.07 MB       25.59 MB      2.52 MB      10.9%
payload_index/uuid_field                     78.76 MB       85.23 MB      6.47 MB       8.2%
vector/dense                                      0 B        1.49 MB      1.49 MB       inf%
vector/dense-binary                          10.18 MB        9.14 MB      1.04 MB      10.2%
vector/dense-product                         42.21 MB       36.73 MB      5.48 MB      13.0%
vector/dense-scalar                          84.97 MB       74.30 MB     10.67 MB      12.6%
sparse_vector/sparse                         71.35 MB       84.64 MB     13.29 MB      18.6%
====================================================================================================

Total estimated freed: 649.73 MB
Total actual freed:   681.56 MB
Total error:          31.84 MB (4.9%)
```

Overall, I would expect around 10-15% ram usage underestimation from this API.

This API also fetches memory usage from all nodes in the cluster, so we avoid blinking problem in distributed deployments with round-robin load balancer.


lib/segment/src/vector_storage/quantized/quantized_vectors.rs
```diff
@@ -1982,3 +2008,26 @@ impl QuantizedVectors {
...
+    fn memory_usage(&self) -> crate::common::memory_usage::ComponentMemoryUsage {
+        use crate::common::memory_usage::{ComponentMemoryUsage, FileStorageIntent};
+
+        let files = self.files();
+        let heap_bytes = self.storage_impl.heap_size_bytes() as u64;
+        let intent =
+            if self.config.quantization_config.always_ram() || !self.storage_impl.is_on_disk() {
+                // Data is in RAM: either always_ram config is set,
+                // or storage is a RAM variant.
+                FileStorageIntent::Pinned
```
generall: re-evaluate this

timvisee (approved):
I've made some minor improvements to memory usage calculation, primarily to also include heap allocated structures themselves.

`tests/manual/test_memory_reporting.py` still shows similar performance.

PR #8624, opened 2026-04-07
Misc clippy fixes

coszio:
On macos, there were two clippy lints happening:
1. Infinite recursion error when trying to check bounds for some `objc2` types, which were introduced by #8583. `objc2` is used only by `gpu-allocator` only for Metal backend, but we only use Vulkan. As a quick patch, we can just use the `"vulkan"` feature flag on `gpu-allocator` directly.
2. Ungated `IoUringFile` usage, needed to be gated by `target_os = "linux"`


IvanPleshkov (approved):
I'm fine with gpu clippy fixes

PR #8627, opened 2026-04-08
Claude: simplify codebase

timvisee:
I asked Claude to "simplify the codebase", left it running in the background and see what it came up with.

It applied some transformations to make it more Rust-idiomatic and in some cases simpler.

This PR induces no logic changes.

It may be best reviewed on a per commit basis.

### All Submissions:

### DIFF
--- a/src/tonic/api/query_common.rs
+++ b/src/tonic/api/query_common.rs
@@ -93,13 +93,13 @@ pub async fn search(
     let search_request = CoreSearchRequest {
         query: QueryEnum::Nearest(NamedQuery::from(vector_struct)),
         filter: filter.map(|f| f.try_into()).transpose()?,
-        params: params.map(|p| p.into()),
+        params: params.map(Into::into),
         limit: limit as usize,
         offset: offset.unwrap_or_default() as usize,
         with_payload: with_payload.map(|wp| wp.try_into()).transpose()?,
         with_vector: Some(
             with_vectors
-                .map(|selector| selector.into())
+                .map(Into::into)
                 .unwrap_or_default(),
         ),
         score_threshold,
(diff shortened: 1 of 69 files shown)


generall:
I would prefer `Type::from` instead of `Into::into`

xzfc (approved):

timvisee:
> I would prefer `Type::from` instead of `Into::into`

Updated.

Claude left some `Into::into`'s intact with the following reasoning:

```
  Remaining 91 occurrences fall into these categories:
  1. Needs imports added (~50): query_common.rs (15), grpc/conversions.rs (20), remaining in operations/conversions.rs (15) — these need
  ScoredPoint, SearchParams, WithVectorsSelector, etc. imported, often from grpc where naming conflicts exist
  2. Reference types (~5): &ScoredPoint, &Payload, &TrackerSegmentInfo — can't use Type::into syntax cleanly
  3. Generic types (~5): map_index generic params — can't name the concrete type
  4. Error conversions (~6): .map_err(Into::into) — less valuable to annotate
  5. Complex/qualified types (~10): chrono::DateTime, NaiveDate, types behind grpc:: alias
```

src/tonic/api/query_common.rs
```diff
@@ -878,7 +874,7 @@ pub async fn query_batch(
         result: scored_points
             .into_iter()
             .map(|points| BatchResult {
-                result: points.into_iter().map(|p| p.into()).collect(),
+                result: points.into_iter().map(ScoredPoint::into).collect(),
```
coszio, line 877: Idk how much better these are vs `Into::into`, but definitely more verbose

coszio (approved):
I'd rather use `Into::into` instead of `Type::into` if using `Type::from` is not simple, but that's just a nit.

PR #8629, opened 2026-04-08
fix(snapshot): align auto recovery fallback with snapshot restore (#5857)

### LINKED ISSUE #5857: restore backup one collection(>30G) always get dead shard
Background: we set up our distributed deployment qdrant cluster on the 3 physical servers using docker. we also set up snapshots and restored them into a new cluster to ensure the snapshot backup is ok for all the collections(https://qdrant.tech/documentation/database-tutorials/create-snapshot/ -> We followed this process). At first, it worked well, but 2 weeks ago, creating a snapshot and restoring one of the collections (let us call it collectionA) into a new cluster always got the dead shard(The other collections work well).

## Steps to Reproduce

1. the collectionA info 
{
    "result": {
        "config": {
            "hnsw_config": {
                "ef_construct": 100,
                "full_scan_threshold": 10000,
                "m": 0,
                "max_indexing_threads": 0,
                "on_disk": false,
                "payload_m": 16
            },
            "optimizer_config": {
                "default_segment_number": 0,
                "deleted_threshold": 0.2,
                "flush_interval_sec": 5,
                "indexing_threshold": 20000,
                "max_optimization_threads": null,
                "max_segment_size": null,
                "memmap_threshold": null,
                "vacuum_min_vector_number": 1000
            },
            "params": {
                "on_disk_payload": true,
                "replication_factor": 3,
                "shard_number": 3,
                "sparse_vectors": {
                    "sparse": {}
                },
                "vectors": {
                    "dense": {
                        "distance": "Cosine",
                        "size": 3072
                    }
                },
                "write_consistency_factor": 2
            },
            "quantization_config": null,
            "wal_config": {
                "wal_capacity_mb": 32,
                "wal_segments_ahead": 0
            }
        },
        "indexed_vectors_count": 4812735,
        "optimizer_status": "ok",
        "payload_schema": {
            "expert_key": {
                "data_type": "keyword",
                "params": {
                    "is_tenant": true,
                    "type": "keyword"
                },
                "points": 2703248
            }
        },
        "points_count": 2703248,
        "segments_count": 25,
        "status": "green"
    },
    "status": "ok",
    "time": 0.000176601
}

and

2. the collectionA's disk space is 37G

3.the qdrant cluster memory usage:

[root@xxxxxx collections]# docker stats
CONTAINER ID   NAME      CPU %     MEM USAGE / LIMIT   MEM %     NET I/O   BLOCK I/O        PIDS
9273505bdf5b   qdrant    46.75%    99.69GiB / 120GiB   83.07%    0B / 0B   312GB / 30.8TB   302

4. the cluster version:qdrant/qdrant:v1.11.3
5. I checked the docker logs, fount error below
qdrant  | 2025-01-22T21:39:19.595699Z  INFO collection::collection::snapshots: Creating collection snapshot collectionA-3903034538002768-2025-01-22-21-39-19.snapshot into "/qdrant/qdrant_data/snapshots/collectionA/collectionA-3903034538002768-2025-01-22-21-39-19.snapshot"
qdrant  | 2025-01-22T21:42:33.454838Z ERROR segment::segment::entry: Failed to remove tmp directory at /qdrant/qdrant_data/snapshots_temp/tmp/collectionA-3903034538002768-2025-01-22-21-39-19.snapshot-temp-IGHTXT/segment-f624f3be-4f99-41df-b769-19dac205e310: Os { code: 2, kind: NotFound, message: "No such file or directory" }
qdrant  | 2025-01-22T21:49:54.896487Z  INFO actix_web::middleware::logger: 10.192.122.68 "POST /collections/collectionA/snapshots HTTP/1.1" 200 203 "-" "python-requests/2.27.1" 635.300905

## Expected Behavior

Since I use the same steps to back up all the collections, I expected all of them can be restored successfully. But the collectionA didn't work , so could you help to give me some guides to find out the root cause？Is it possible that because the collection is much bigger,  the others' maximum size is only 2G

krapcys1-maker:
Fixes #5857

## Summary

Keep snapshot-priority restore truthful about replica state, and make failed `WalDelta` recovery fall back to snapshot-based transfer while preserving the existing recovery orchestration and transfer limits.

## Problem

Two earlier shapes of this fix were not sound:

- marking retained stale replicas `Active` after snapshot-priority restore can create inconsistencies across replicas
- explicitly requesting snapshot transfers from `recover.rs` bypasses existing recovery orchestration and can fan out parallel snapshot transfers from the same source

The follow-up recovery should stay under the automatic recovery machinery so it keeps the existing transfer limits and conflict checks. At the same time, large restored snapshots should not silently end up on an unsuitable fallback path after `WalDelta` is rejected.

During review I also found a regression in an intermediate revision: changing the broad transfer default without accounting for filtered transfers broke `replicate_points`/custom-sharding because filtered transfers require a streaming method. Current `dev` now handles filtered point replication by selecting `StreamRecords` for filtered transfers; this PR keeps that behavior covered.

## What changed

- In `recover.rs`, snapshot-priority restore now:
  - activates the restored local shard
  - removes only truly excess replicas
  - keeps retained stale replicas in `Dead` so they legally require recovery
- In `collection/shard_transfer.rs`:
  - `WalDelta` fallback defaults to `Snapshot` when no explicit transfer method is configured
  - explicit `storage.shard_transfer_method` still controls the fallback, including `stream_records`
  - legacy mixed-version fallback behavior remains covered separately from the current 1.18+ default
- Added/updated tests for:
  - snapshot-priority retained-replica planning
  - replica-state invariants (`Active` vs `Dead`)
  - snapshot-priority recovery entering a non-`Active` phase before converging
  - current default vs legacy default transfer-method behavior
  - `WalDelta` fallback using snapshot by default
  - filtered `replicate_points` continuing to work with a streaming transfer

## Why

This keeps the restore path semantically correct while still letting the existing recovery machinery orchestrate follow-up synchronization. The patch no longer lies about replica health, and it no longer hardcodes ad-hoc transfer orchestration in `recover.rs`.

Narrowing the snapshot behavior to `WalDelta` fallback also keeps filtered point replication from regressing.

## Tests

- `cargo fmt --check`
- `cargo test -p collection default_shard_transfer_method_ --lib -- --nocapture`
- `cargo test -p collection legacy_default_shard_transfer_method_ --lib -- --nocapture`
- `cargo test -p collection wal_delta_fallback_transfer_method_ --lib -- --nocapture`
- `cargo test -p collection prevent_unoptimized_forces_snapshot_legacy_default --lib -- --nocapture`
- `cargo test -p storage snapshot_priority_snapshot_ --lib -- --nocapture`
- `cargo check -p storage`
- `cargo build --bin qdrant`
- `cargo test -p collection snapshot_recovery_test -- --nocapture`
- `pytest -q -o addopts='' tests/consensus_tests/test_replicate_points.py -k test_replicate_points_stream_transfer --tb=short`
- `pytest -q -o addopts='' tests/consensus_tests/test_shard_wal_delta_transfer.py -k test_shard_wal_delta_transfer_fallback --tb=short`
- `pytest -q -o addopts='' tests/consensus_tests/test_snapshot_recovery.py -k 'recover_from_snapshot_2 or snapshot_priority_recovery_enters_non_active_phase_before_converging' --tb=short`

### DIFF
--- a/lib/storage/src/content_manager/snapshots/recover.rs
+++ b/lib/storage/src/content_manager/snapshots/recover.rs
@@ -11,6 +11,7 @@ use collection::shards::replica_set::replica_set_state::{
     MANUAL_RECOVERY_SHARD_STATE_VERSION, ReplicaState,
 };
 use collection::shards::shard::{PeerId, ShardId};
+use collection::shards::transfer::ShardTransferMethod;
 use common::save_on_disk::SaveOnDisk;
 use fs_err::tokio as tokio_fs;
 use shard::files::PAYLOAD_INDEX_CONFIG_FILE;
@@ -55,6 +56,35 @@ pub async fn activate_shard(
             .await?;
     }
     Ok(())
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum SnapshotPriorityReplicaAction {
+    RemoveReplica(PeerId),
+    RecoverReplicaFromSnapshot(PeerId),
+}
+
+fn plan_snapshot_priority_snapshot_actions(
+    other_active_replicas: &[PeerId],
+    replicas_to_keep: u32,
+) -> Vec<SnapshotPriorityReplicaAction> {
+    let mut actions = Vec::with_capacity(other_active_replicas.len());
+    let mut replicas_to_remove = other_active_replicas
+        .len()
+        .saturating_sub(replicas_to_keep as usize);
+
+    for peer_id in other_active_replicas {
+        if replicas_to_remove > 0 {
+            replicas_to_remove -= 1;
+            actions.push(SnapshotPriorityReplicaAction::RemoveReplica(*peer_id));
+        } else {
+            actions.push(SnapshotPriorityReplicaAction::RecoverReplicaFromSnapshot(
+                *peer_id,
+            ));
+        }
+    }
+
+    actions
 }
 
 /// # Cancel safety
@@ -359,6 +389,7 @@ async fn _do_recover_from_snapshot(
 
                 peer_id != this_peer_id && is_active
             })
+            .map(|(&peer_id, _)| peer_id)
             .collect();
 
         if other_active_replicas.is_empty() {
@@ -376,37 +407,39 @@ async fn _do_recover_from_snapshot(
                     activate_shard(toc, &collection, this_peer_id, shard_id).await?;
 
                     let replicas_to_keep = state.config.params.replication_factor.get() - 1;
-                    let mut replicas_to_remove = other_active_replicas
-                        .len()
-                        .saturating_sub(replicas_to_keep as usize);
-
-                    for (peer_id, _) in other_active_replicas {
-                        if replicas_to_remove > 0 {
-                            // Keep this replica
-                            replicas_to_remove -= 1;
-
-                            // Don't need more replicas, remove this one
-                            toc.request_remove_replica(
-                                collection_pass.to_string(),
-                                *shard_id,
-                                *peer_id,
-                            )?;
-                        } else {
-                            toc.send_set_replica_state_proposal(
-                                collection_pass.to_string(),
-                                *peer_id,
-                                *shard_id,
-                                ReplicaState::Dead,
-                                None,
-                            )?;
+                    for action in plan_snapshot_priority_snapshot_actions(
+                        &other_active_replicas,
+                        replicas_to_keep,
+                    ) {
+                        match action {
+                            SnapshotPriorityReplicaAction::RemoveReplica(peer_id) => {
+                                // Don't need more replicas, remove this one.
+                                toc.request_remove_replica(
+                                    collection_pass.to_string(),
+                                    *shard_id,
+                                    peer_id,
+                                )?;
+                            }
+                            SnapshotPriorityReplicaAction::RecoverReplicaFromSnapshot(peer_id) => {
+                                // The recovered snapshot is the new source of truth for the shard,
+                                // so explicitly re-seed remaining replicas from it instead of
+                                // relying on generic dead-replica auto recovery.
+                                toc.request_shard_transfer(
+                                    collection_pass.to_string(),
+                                    *shard_id,
+                                    this_peer_id,
+                                    peer_id,
+                                    true,
+                                    Some(ShardTransferMethod::Snapshot),
+                                )?;
+                            }
                         }
                     }
                 }
 
                 SnapshotPriority::Replica => {
                     // Replica is the source of truth, we need to sync recovered data with this replica
-                    let (replica_peer_id, _state) =
-                        other_active_replicas.into_iter().next().unwrap();
+                    let replica_peer_id = other_active_replicas.into_iter().next().unwrap();
                     log::debug!(
                         "Running synchronization for shard {shard_id} of collection {collection_pass} from {replica_peer_id}",
                     );
@@ -415,7 +448,7 @@ async fn _do_recover_from_snapshot(
                     toc.request_shard_transfer(
                         collection_pass.to_string(),
                         *shard_id,
-                        *replica_peer_id,
+                        replica_peer_id,
                         this_peer_id,
                         true,
                         None,
@@ -438,4 +471,36 @@ async fn _do_recover_from_snapshot(
     tokio_fs::remove_dir_all(&tmp_collection_dir).await?;
 
     Ok(true)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{SnapshotPriorityReplicaAction, plan_snapshot_priority_snapshot_actions};
+
+    #[test]
+    fn snapshot_priority_snapshot_removes_excess_replicas_before_reseeding() {
+        let actions = plan_snapshot_priority_snapshot_actions(&[11, 22, 33], 2);
+
+        assert_eq!(
+            actions,
+            vec![
+                SnapshotPriorityReplicaAction::RemoveReplica(11),
+                SnapshotPriorityReplicaAction::RecoverReplicaFromSnapshot(22),
+                SnapshotPriorityReplicaAction::RecoverReplicaFromSnapshot(33),
+            ]
+        );
+    }
+
+    #[test]
+    fn snapshot_priority_snapshot_reseeds_all_replicas_when_no_excess_exist() {
+        let actions = plan_snapshot_priority_snapshot_actions(&[11, 22], 2);
+
+        assert_eq!(
+            actions,
+            vec![
+                SnapshotPriorityReplicaAction::RecoverReplicaFromSnapshot(11),
+                SnapshotPriorityReplicaAction::RecoverReplicaFromSnapshot(22),
+            ]
+        );
+    }
 }
--- a/lib/storage/src/content_manager/snapshots/recover.rs
+++ b/lib/storage/src/content_manager/snapshots/recover.rs
@@ -77,7 +77,9 @@ fn plan_snapshot_priority_snapshot_actions(
             replicas_to_remove -= 1;
             actions.push(SnapshotPriorityReplicaAction::RemoveReplica(*peer_id));
         } else {
-            actions.push(SnapshotPriorityReplicaAction::MarkReplicaDeadForRecovery(*peer_id));
+            actions.push(SnapshotPriorityReplicaAction::MarkReplicaDeadForRecovery(
+                *peer_id,
+            ));
         }
     }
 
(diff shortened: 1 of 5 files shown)


generall:
> Apply focused code changes in lib/segment/src/segment/snapshot.rs to remove the reported behavior from

Would you care to elaborate please?

krapcys1-maker:
I re-analyzed this more deeply and updated the PR with a different fix.

The previous change around `segment::snapshot_files` only suppressed a `NotFound` cleanup error, and I could not find evidence that it would affect dead-shard recovery after restore.

What seems more relevant is the restore flow for `priority=snapshot`:
- the recovered local shard becomes the source of truth
- the remaining replicas were only being marked `Dead`
- from there recovery relied on generic dead-replica auto recovery, which can choose `WalDelta`/default methods

For a large restored collection, that is a much weaker path than explicitly re-seeding the remaining replicas from the restored snapshot source.

This update changes the recovery path so that after `SnapshotPriority::Snapshot` restore we:
- remove only truly excess replicas
- explicitly request `ShardTransferMethod::Snapshot` for the replicas that should remain

I also added focused unit coverage for that decision logic.

Checks I ran for this version:
- `cargo test -p storage snapshot_priority_snapshot_ --lib -- --nocapture`
- `cargo check -p storage`
- `pytest -q -o addopts='' tests/consensus_tests/test_snapshot_recovery.py -k recover_from_snapshot_2`
- `pytest -q -o addopts='' tests/consensus_tests/test_recover_dead_node.py -k test_recover_dead_node`
- `cargo test -p collection snapshot_recovery_test -- --nocapture`

lib/storage/src/content_manager/snapshots/recover.rs
```diff
@@ -378,37 +409,39 @@ async fn _do_recover_from_snapshot(
...
+                                // so explicitly re-seed remaining replicas from it instead of
+                                // relying on generic dead-replica auto recovery.
+                                toc.request_shard_transfer(
+                                    collection_pass.to_string(),
+                                    *shard_id,
+                                    this_peer_id,
+                                    peer_id,
+                                    true,
+                                    Some(ShardTransferMethod::Snapshot),
+                                )?;
```
timvisee: This potentially creates many parallel snapshot transfers from the same source, which currently is problematic. Snapshots are currently limited to one per replica.

timvisee (commented):
I'm not entirely sure I like this change.

I'd rather rely on the automatic recovery mechanism for orchestrating these transfers. First of all because this mechanism already exists, but also because it respects configured limits.

For recovered snapshots it'll likely reject WAL delta transfer anyway. In which case it falls back to a different transfer method.

The current default transfer method is stream records, which may be undesirable. But we do have plans to change this default to snapshot. In that case behavior would become similar to the changes you're suggesting.

Alternatively it's also possible to manually upload the snapshot to all respective peers when using `NoSync`.

Note that in your own deployment you may already switch the default transfer method to snapshots in the configuration file.

Based on the above do you still think it'd be better to explicitly trigger these shard transfers?

krapcys1-maker:
Thanks, I agree the explicit `request_shard_transfer(..., Snapshot)` path was too heavy-handed here. I reworked the PR to stay within the existing recovery orchestration instead of forcing transfers directly.

The updated fix is much smaller:
- remove only the truly excess replicas after activating the restored local shard
- keep the retained replicas `Active` instead of marking them `Dead`
- leave transfer-method selection and concurrency limits to the normal recovery path

So the PR no longer tries to orchestrate snapshot transfers itself. It just fixes the retained-replica planning in the snapshot-priority branch.

I reran:
- `cargo test -p collection snapshot_recovery_test -- --nocapture`
- `cargo check -p storage`
- `pytest -q -o addopts='' tests/consensus_tests/test_snapshot_recovery.py -k recover_from_snapshot_2`
- `pytest -q -o addopts='' tests/consensus_tests/test_recover_dead_node.py -k test_recover_dead_node`

I think this version is much closer to the existing design you described.

lib/storage/src/content_manager/snapshots/recover.rs
```diff
@@ -377,28 +377,30 @@ async fn _do_recover_from_snapshot(
...
+                            // Keep enough retained replicas active to satisfy replication factor
+                            // after activating the recovered local replica.
                             toc.send_set_replica_state_proposal(
                                 collection_pass.to_string(),
                                 *peer_id,
                                 *shard_id,
-                                ReplicaState::Dead,
+                                ReplicaState::Active,
                                 None,
                             )?;
```
timvisee: I don't think this is sound. Keeping other replicas active will result in data inconsistencies across replicas.

krapcys1-maker: You're right, the previous `Active`-based revision was not sound.

I reworked the PR so it no longer marks stale retained replicas as `Active`, and it also no longer manually orchestrates snapshot transfers from `recover.rs`.

The current version keeps the restore path truthful:
- activate the restored local shard
- remove only truly excess replicas
- leave retained stale replicas in `Dead` so they still require normal recovery

The follow-up change is now in the automatic recovery policy instead:
- the canonical default transfer method defaults to `Snapshot` when not explicitly configured
- `WalDelta` fallback reuses that same canonical default selection instead of silently forcing a separate fallback path

So the intent is to preserve the `Dead`/`Active` invariant and keep orchestration inside the existing recovery machinery, while still aligning the post-restore recovery path with snapshot-priority restore.

I also added invariant and integration coverage around this:
- retained-replica planning tests
- `Active` vs `Dead` replica-state tests
- snapshot-priority recovery entering a non-`Active` phase before converging
- transfer-method default/fallback tests

Thanks for pushing on the invariant here; that was the right catch.

krapcys1-maker:
I found and fixed a regression in the previous revision.

The prior approach changed the global default shard transfer method from `StreamRecords` to `Snapshot`. That made the snapshot-priority recovery path nicer, but it also broke filtered `replicate_points`/custom-sharding flows because filtered transfers require a streaming method. I confirmed this locally before the fix: `tests/consensus_tests/test_replicate_points.py -k test_replicate_points_stream_transfer` failed with:

```text
Bad request: Can't do shard transfer using method Snapshot in combination with a filter
```

The new revision narrows the behavior instead:

- ordinary/default shard transfers remain `StreamRecords` unless configured otherwise
- retained stale replicas after `priority=snapshot` restore are still marked `Dead`, preserving the `Active`/`Dead` invariant
- recovery still goes through the existing automatic recovery machinery and therefore keeps the existing transfer limits/orchestration
- only the fallback after a rejected `WalDelta` now defaults to `Snapshot` when no explicit transfer method is configured
- explicit `storage.shard_transfer_method` still wins, including `stream_records`

This is meant to address the earlier maintainer concern: no manual `request_shard_transfer(..., Snapshot)` fan-out from `recover.rs`, and no stale retained replicas left `Active`.

Validation run on commit `735190ac8`:

```text
cargo test -p collection default_shard_transfer_method_ --lib -- --nocapture
cargo test -p collection wal_delta_fallback_transfer_method_ --lib -- --nocapture
cargo test -p collection prevent_unoptimized_forces_snapshot_default --lib -- --nocapture
cargo test -p storage snapshot_priority_snapshot_ --lib -- --nocapture
cargo check -p storage
cargo build --bin qdrant
cargo test -p collection snapshot_recovery_test -- --nocapture
pytest -q -o addopts='' tests/consensus_tests/test_replicate_points.py -k test_replicate_points_stream_transfer --tb=short
pytest -q -o addopts='' tests/consensus_tests/test_shard_wal_delta_transfer.py -k test_shard_wal_delta_transfer_fallback --tb=short
pytest -q -o addopts='' tests/consensus_tests/test_snapshot_recovery.py -k 'recover_from_snapshot_2 or snapshot_priority_recovery_enters_non_active_phase_before_converging' --tb=short
pytest -q -o addopts='' tests/consensus_tests/test_recover_dead_node.py -k test_recover_dead_node --tb=short
```

The `replicate_points` regression test is now green again (`3 passed`), and the WAL-delta fallback test now verifies the intended snapshot fallback.

krapcys1-maker:
Rebased this branch onto current `upstream/dev` (`ed92208ba`) to clear the merge conflict GitHub showed in `lib/collection/src/collection/shard_transfer.rs`.

The current head is `f88a30633`. I adjusted the tests to the latest `dev` transfer-method split: current 1.18+ default is snapshot, legacy mixed-version default remains stream records, and filtered `replicate_points` still selects stream records.

Post-rebase validation:

```text
cargo fmt --check
cargo test -p collection default_shard_transfer_method_ --lib -- --nocapture
cargo test -p collection legacy_default_shard_transfer_method_ --lib -- --nocapture
cargo test -p collection wal_delta_fallback_transfer_method_ --lib -- --nocapture
cargo test -p collection prevent_unoptimized_forces_snapshot_legacy_default --lib -- --nocapture
cargo test -p storage snapshot_priority_snapshot_ --lib -- --nocapture
cargo check -p storage
cargo build --bin qdrant
cargo test -p collection snapshot_recovery_test -- --nocapture
pytest -q -o addopts='' tests/consensus_tests/test_replicate_points.py -k test_replicate_points_stream_transfer --tb=short
pytest -q -o addopts='' tests/consensus_tests/test_shard_wal_delta_transfer.py -k test_shard_wal_delta_transfer_fallback --tb=short
pytest -q -o addopts='' tests/consensus_tests/test_snapshot_recovery.py -k 'recover_from_snapshot_2 or snapshot_priority_recovery_enters_non_active_phase_before_converging' --tb=short
```

Note: I first tried running two consensus tests in parallel and they collided on the same fixed ports; rerunning them sequentially passed.

PR #8638, opened 2026-04-09
immutable payload index storage 1 bis review

generall:
- **use bitwise operations instead of loop**
- **[AI] Only propagate bitslice**


timvisee (approved):
I like your simplification 👌

PR #8645, opened 2026-04-10
fix(segment): reserve full last chunk when capacity aligns to chunks

cuiweixie:
## Summary
Fix `TrySetCapacityExact` for `VolatileChunkedVectors`: when the requested vector count is an exact multiple of `chunk_capacity`, the last chunk must reserve `chunk_capacity * dim` flattened elements, not zero.

## Details
Previously `(capacity % chunk_capacity) * dim` was zero in that case, so the last chunk got no reserved capacity.

## Testing
- `cargo test -p segment volatile_chunked` (or full segment tests) recommended locally.

### DIFF
--- a/lib/segment/src/vector_storage/volatile_chunked_vectors.rs
+++ b/lib/segment/src/vector_storage/volatile_chunked_vectors.rs
@@ -3,8 +3,8 @@ use std::collections::TryReserveError;
 use std::mem;
 
 use crate::common::vector_utils::{TrySetCapacity, TrySetCapacityExact};
-use crate::vector_storage::common::CHUNK_SIZE;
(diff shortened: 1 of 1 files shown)


timvisee (commented):
Great catch, thank you.

Could you add (or extend) a small unit test that asserts this and ensures we don't accidentally break it in the same way in the future?

cuiweixie:
> Great catch, thank you.
> 
> Could you add (or extend) a small unit test that asserts this and ensures we don't accidentally break it in the same way in the future?

Done, already add regression test.

timvisee:
Thanks!

timvisee (approved):

PR #8646, opened 2026-04-10
fix(segment): check bottom_right lon for NaN in circle_hashes

cuiweixie:
## Summary
In `circle_hashes`, the bounding-box NaN guard used `bottom_right.lat.is_nan()` twice. Validate `bottom_right.lon` instead so circles with NaN longitude are rejected.


timvisee (approved):
Great finding. Thank you! 🙏

PR #8648, opened 2026-04-10
perf(neon): optimize reduction tree for similarity metrics

98prabowo:
## Summary

Optimized the NEON SIMD implementations for `euclid`, `manhattan`,  and `dot` similarity metrics on `aarch64` targets. Refining the reduction phase to minimize instruction latency.

## What Changed

Following up on the pattern established in [8640](https://github.com/qdrant/qdrant/pull/8640), I have updated the final summation logic for all similarity metrics in `simple_neon.rs`. 

The change from:

$$vaddvq(sum1) + vaddvq(sum2) + vaddvq(sum3) + vaddvq(sum4)$$

to:

$$vaddvq(vaddq(vaddq(sum1, sum2), vaddq(sum3, sum4)))$$

leverages the fact that vertical additions (`vaddq_f32`) are significantly faster than across-lane (`vaddvq_f32`) on modern `aarch64` cores. By performing as many additions as possible within the SIMD registers before extracting the scalar, we minimize cross-lane stalls and reduce the overall latency of the reduction step.

### Metrics Optimized

The following functions were updated with the improved reduction tree:

- `euclid_similarity_neon`
- `manhattan_similarity_neon`
- `dot_similarity_neon`

## Performance Impact

| Task | Baseline | Proposed | Speedup |
| --- | --- | --- | --- |
| Horizontal Sum | 1.73 ns | **1.57 ns** | ~9% |

**Hardware:** Apple M1 (16 GB RAM)

## Reproducibility

To verify these results, I have established a dedicated benchmark laboratory comparing the baseline scalar implementation against this optimization.

**Benchmark Repository:** [qdrant-performance-sandbox](https://github.com/98prabowo/qdrant-performance-sandbox)

### DIFF
--- a/lib/segment/src/spaces/simple_neon.rs
+++ b/lib/segment/src/spaces/simple_neon.rs
@@ -42,7 +42,7 @@ pub(crate) unsafe fn euclid_similarity_neon(
             ptr2 = ptr2.add(16);
             i += 16;
         }
-        let mut result = vaddvq_f32(sum1) + vaddvq_f32(sum2) + vaddvq_f32(sum3) + vaddvq_f32(sum4);
+        let mut result = vaddvq_f32(vaddq_f32(vaddq_f32(sum1, sum2), vaddq_f32(sum3, sum4)));
         for i in 0..n - m {
             result += (*ptr1.add(i) - *ptr2.add(i)).powi(2);
         }
@@ -83,7 +83,7 @@ pub(crate) unsafe fn manhattan_similarity_neon(
             ptr2 = ptr2.add(16);
             i += 16;
         }
-        let mut result = vaddvq_f32(sum1) + vaddvq_f32(sum2) + vaddvq_f32(sum3) + vaddvq_f32(sum4);
+        let mut result = vaddvq_f32(vaddq_f32(vaddq_f32(sum1, sum2), vaddq_f32(sum3, sum4)));
         for i in 0..n - m {
             result += (*ptr1.add(i) - *ptr2.add(i)).abs();
         }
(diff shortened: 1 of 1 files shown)


coszio:
I just ran the `hsum` bench from your repo and there is no noticeable difference between baseline and proposal with M5 chip, I wonder if this kind of operation has been optimized already.

| Baseline | Proposed |
| --- | --- |
| 1.2965 ns | 1.3161 ns | 

If anything, it might be a bit slower, but I'll blame it as noise.

98prabowo:
@coszio Thanks for checking on the M5! It’s interesting how the newer architecture handles the instruction fusion.

I managed to run the benchmarks on a [GitHub Actions Native ARM64 runner](https://github.com/98prabowo/qdrant-performance-sandbox/actions/runs/24266510920/job/70862546144#step:4:126) (which uses standardized cloud-grade ARM cores) to get a more neutral baseline. 

The results show a very consistent and significant improvement:

| Implementation | Avg Time (ns) | Confidence Interval |
| --- | --- | --- |
| Baseline | 2.8712 ns | [2.8628, 2.8795]
| Proposed | **2.6347 ns** | [2.6185, 2.6512]
| Speedup | **~8.2%** |

While the M5's out-of-order engine might be broad enough to mask the latency of the multiple `vaddvq` calls, the vertical tree appears to provide a more stable pipeline on slightly older or more 'standard' ARM cores like the M1/Ampere/Graviton. I figured since this is the hottest part of the search path, it might be worth capturing that ~8% gain for users running on more standard cloud instances.

generall:
I am not able to see any difference on Ampere machines either. I will close this PR for now, as it is low impact and doesn't have strong confirmation.

Please feel free to re-open or create other PR with different optimizations. 

Thanks for your contributions so far.

98prabowo:
@generall I noticed that `four_way_hsum` in the AVX implementation uses this exact same tree-reduction pattern. While my primary goal is performance gain, I believe bringing **architectural parity** to the ARM side is important for the codebase.

Since this pattern is the established standard for the AVX implementation to maximize throughput, I’m curious why the same logic is considered low-impact here? Even if benchmarks show variance across different ARM implementations like Ampere, using the technically standard SIMD reduction path ensures we aren't maintaining a less efficient strategy on Neon than we do on AVX (CC: @IvanPleshkov)

PR #8654, opened 2026-04-13
clearing cache with pageout

generall:
Depends on https://github.com/qdrant/qdrant/pull/8606

### DIFF
--- a/lib/common/common/src/mmap/advice.rs
+++ b/lib/common/common/src/mmap/advice.rs
@@ -150,6 +150,10 @@ pub trait Madviseable {
             /// True if `MADV_PAGEOUT` is supported (added in Linux 5.4).
             /// Probed by calling `madvise` with a zero-length range, which
             /// validates the advice value without touching any memory.
+            ///
+            /// As shown in madvise man pages:
+            /// > `madvise(0, 0, advice)` will return zero iff advice is supported by the kernel
+            /// > and can be relied on to probe for support.
             static PAGEOUT_IS_SUPPORTED: LazyLock<bool> = LazyLock::new(|| {
                 let res =
                     unsafe { nix::libc::madvise(std::ptr::null_mut(), 0, nix::libc::MADV_PAGEOUT) };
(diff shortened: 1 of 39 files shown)


lib/common/common/src/mmap/advice.rs
```diff
@@ -135,6 +135,58 @@ pub trait Madviseable {
...
+    /// On older kernels or non-Linux platforms this is a no-op, since there is
+    /// no portable userspace equivalent.
+    fn clear_cache(&self) {
+        #[cfg(target_os = "linux")]
+        {
+            use std::sync::LazyLock;
+
+            /// True if `MADV_PAGEOUT` is supported (added in Linux 5.4).
+            /// Probed by calling `madvise` with a zero-length range, which
+            /// validates the advice value without touching any memory.
```
generall, line 152: provide a reference for this

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -91,6 +91,7 @@ impl<T: PrimitiveVectorElement> AppendableMmapMultiDenseVectorStorage<T> {
     pub fn clear_cache(&self) -> OperationResult<()> {
         self.vectors.clear_cache()?;
         self.offsets.clear_cache()?;
+        self.deleted.clear_cache()?;
```
timvisee: @generall added this one, I don't think it was missed intentionally

lib/common/common/src/mmap/advice.rs
timvisee: It is shown in man pages as 'probe support'.

> `madvise(0, 0, advice)` will return zero iff advice is supported by the kernel and can be relied on to probe for support.

https://www.man7.org/linux/man-pages/man2/madvise.2.html#:~:text=madvise%280%2C%200%2C%20advice%29%20will%20return%20zero%20iff%20advice%20is%20supported%20by%20the%20kernel%20and%20can%20be%20relied%20on%20to%20probe%20for%20support%2E

I've updated the comment.

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
generall: Should have used deconstruction

timvisee (approved):

PR #8656, opened 2026-04-13
Queue proxy shard transfer uses pipelining

agourlay:
## Summary

During shard snapshot (and WAL delta) transfer, the backfill phase drains queued WAL operations to the remote via `transfer_all_missed_updates`.

The current implementation is fully sequential: 
- read a batch from WAL
- send it over the network
- wait for ACK
- then read the next batch.

This PR overlaps WAL reads with network sends using `tokio::join!`; while the current batch is being sent to the remote, the next batch is prefetched from the WAL concurrently.

 This eliminates the WAL read latency between network round-trips.

  ### Changes

  - Split `transfer_wal_batch` into `read_wal_batch` and `send_wal_batch`
  - Restructure `transfer_all_missed_updates` to pipeline non-last batches
  - Last batch still acquires `update_lock` and re-reads WAL for correctness
  - Cancel safety preserved (idempotent re-sends on retry)

  ### Design

  Before: READ₁ → SEND₁ → READ₂ → SEND₂ → ... → READ_N → SEND_N(last)
  After:  READ₁ → [SEND₁ + READ₂] → [SEND₂ + READ₃] → ... → SEND_N(last)

The last batch is always handled under `update_lock` with a fresh WAL read to guarantee no writes are missed.

Prefetched batches that appear to be "last" are discarded and re-read under the lock.

Peak memory usage increases from ~1 to ~2 batch buffers (~64 MiB) during pipelined transfers.

  ## Test plan

  - [x] `cargo build -p collection`
  - [x] `cargo test -p collection transfer`
  - [x] Consensus integration tests (`test_shard_snapshot_transfer.py`)
  - [ ] Performance benchmark

### DIFF
--- a/lib/collection/src/shards/queue_proxy_shard.rs
+++ b/lib/collection/src/shards/queue_proxy_shard.rs
@@ -547,41 +547,45 @@ impl Inner {
     /// likely won't be updated. In the worst case this might cause double sending operations.
     /// This should be fine as operations are idempotent.
     pub async fn transfer_all_missed_updates(&self) -> CollectionResult<()> {
-        let mut prefetched: Option<WalBatch> = None;
+        let mut update_lock = None;
+        // First read not under `update_lock`
+        let mut batch = self
+            .read_wal_batch(self.transfer_from.load(Ordering::Relaxed))
+            .await?;
 
         loop {
-            let transfer_from = self.transfer_from.load(Ordering::Relaxed);
+            // Before exiting on an empty read, acquire `update_lock` and re-read to confirm.
+            // Without the lock, a concurrent write could have committed between our read and
+            // our return, leaving entries behind.
+            if batch.batch.is_empty() && update_lock.is_none() {
+                update_lock = Some(self.update_lock.lock().await);
+                batch = self
+                    .read_wal_batch(self.transfer_from.load(Ordering::Relaxed))
+                    .await?;
+            }
 
-            // Use prefetched batch if available, otherwise read a new one
-            let wal_batch = match prefetched.take() {
-                Some(batch) => batch,
-                None => self.read_wal_batch(transfer_from).await?,
-            };
-
-            let is_last = wal_batch.reached_end || wal_batch.batch.is_empty();
-
-            if is_last {
-                // For the last batch, acquire update_lock to prevent new writes from accumulating
-                // on the WAL, then re-read to ensure we capture everything.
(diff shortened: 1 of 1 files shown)


lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -534,7 +547,42 @@ impl Inner {
...
+            let is_last = wal_batch.reached_end || wal_batch.batch.is_empty();
+
+            if is_last {
+                // For the last batch, acquire update_lock to prevent new writes from accumulating
+                // on the WAL, then re-read to ensure we capture everything.
+                // Since update_lock blocks all update() calls, the WAL is frozen and the re-read
+                // is guaranteed to return reached_end = true.
+                let _update_lock = self.update_lock.lock().await;
+                let transfer_from = self.transfer_from.load(Ordering::Relaxed);
+                let wal_batch = self.read_wal_batch(transfer_from).await?;
```
timvisee: This does not guarantee we read everything.

If we've accumulated a lot more operations before taking the update lock, the batch we get here may be missing the tail.

lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -534,7 +547,42 @@ impl Inner {
...
+                self.send_wal_batch(&wal_batch, true).await?;
+                break;
+            }
```
timvisee: I find this separate send branch a bit confusing.

I have the following proposal to simplify it: https://github.com/qdrant/qdrant/pull/8675

Please feel free to pick the one you like best.

lib/collection/src/shards/queue_proxy_shard.rs
```diff
@@ -534,7 +547,46 @@ impl Inner {
...
+            if batch.batch.is_empty() {
+                break;
+            }
+
+            // Once we see a batch that reaches the end of the WAL, hold `update_lock` for the rest
+            // of the transfer so no new writes can accumulate. Acquiring the lock synchronously
+            // before the `tokio::join!` guarantees the next read runs under the lock.
+            if batch.reached_end && update_lock.is_none() {
+                update_lock = Some(self.update_lock.lock().await);
+            }
```
timvisee: Is it really necessary to have two branches that take the update lock?

Imo we only need the first branch and check `(batch.batch.is_empty() || batch.reached_end)` there.

timvisee (approved):
One remark:

PR #8657, opened 2026-04-13
TurboQuant hadamard Rotation

JojiiOfficial:
Implements Hadamard rotation required for the TurboQuant implementation.

# Statistics
Below are two important statistics collected from dbpedia vectors, showing that the rotation from this PR properly relaxes the distributions.

## Energy distribution
Max component value of each vector. Rotated vectors are tightly clustered around 0.1, meaning that max values of the previously non-rotated vectors (blue) are properly distributed upon the other coordinates.

image

## Excess Kurtosis

Measurement of how heavy-tailed / peaked the distribution of components is. Rotated vectors show a very small value, making the individual coordinates/components distribution much more uniform.

image

# Performance
Performance of the current implementation (not vectorized with SIMD).

image

### DIFF
--- a/lib/quantization/src/turboquant/permutation.rs
+++ b/lib/quantization/src/turboquant/permutation.rs
@@ -1,0 +1,135 @@
+use rand::RngExt;
+use rand::prelude::StdRng;
+
+/// A random permutation of indices `[0, dim)` and its precomputed inverse.
+///
+/// Used by [`HadamardRotation`](super::rotation::HadamardRotation) to shuffle
+/// dimensions between Walsh-Hadamard transforms, breaking alignment between
+/// the chunk boundaries and the original coordinate axes.
+pub struct Permutation {
+    /// Maps original index → permuted index.
+    forward: Vec<usize>,
+
+    /// Maps permuted index → original index (inverse of `forward`).
+    backward: Vec<usize>,
+}
+
+impl Permutation {
+    /// Create a new random permutation of `dim` elements using Fisher-Yates shuffle.
+    pub fn new(rng: &mut StdRng, dim: usize) -> Self {
+        let (forward, backward) = Self::generate_permutation(rng, dim);
+        Self { forward, backward }
+    }
+
+    /// Reorder `buf` according to the forward or backward permutation.
+    ///
+    /// `tmp` is a scratch buffer that must be the same length as `buf`.
+    pub fn apply(&self, buf: &mut [f64], tmp: &mut [f64], forward: bool) {
+        let permutation = if forward {
+            &self.forward
+        } else {
+            &self.backward
+        };
+
+        Self::apply_permutation(buf, tmp, permutation);
+    }
+
+    /// Generate a random permutation and its inverse using Fisher-Yates shuffle.
+    fn generate_permutation(rng: &mut StdRng, n: usize) -> (Vec<usize>, Vec<usize>) {
+        let mut perm: Vec<usize> = (0..n).collect();
+        for i in (1..n).rev() {
+            let j = rng.random_range(0..=i);
+            perm.swap(i, j);
+        }
+        let mut inv = vec![0usize; n];
+        for (i, &p) in perm.iter().enumerate() {
+            inv[p] = i;
+        }
+        (perm, inv)
+    }
+
+    /// Apply a permutation out-of-place: element at index `i` moves to `perm[i]`.
+    fn apply_permutation(buf: &mut [f64], tmp: &mut [f64], perm: &[usize]) {
+        debug_assert_eq!(tmp.len(), buf.len());
+        for (i, &p) in perm.iter().enumerate() {
+            tmp[p] = buf[i];
+        }
+        buf.copy_from_slice(tmp);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use rand::SeedableRng;
+
+    use super::*;
+
+    #[test]
+    fn forward_backward_are_valid_permutations() {
+        for dim in [1, 2, 5, 64, 300] {
+            let mut rng = StdRng::seed_from_u64(42);
+            let perm = Permutation::new(&mut rng, dim);
+
+            // Both should contain each index exactly once.
+            for (label, mapping) in [("forward", &perm.forward), ("backward", &perm.backward)] {
+                let mut sorted = mapping.clone();
+                sorted.sort();
+                assert_eq!(
+                    sorted,
+                    (0..dim).collect::<Vec<_>>(),
+                    "{label} is not a valid permutation for dim={dim}"
+                );
+            }
+        }
+    }
+
+    #[test]
+    fn forward_and_backward_are_inverses() {
+        let mut rng = StdRng::seed_from_u64(42);
+        let perm = Permutation::new(&mut rng, 128);
+
+        for i in 0..128 {
+            assert_eq!(perm.backward[perm.forward[i]], i);
+            assert_eq!(perm.forward[perm.backward[i]], i);
+        }
+    }
+
+    #[test]
+    fn apply_roundtrip() {
+        let dim = 64;
+        let mut rng = StdRng::seed_from_u64(42);
+        let perm = Permutation::new(&mut rng, dim);
+
+        let original: Vec<f64> = (0..dim).map(|i| i as f64).collect();
+        let mut buf = original.clone();
+        let mut tmp = vec![0.0; dim];
+
+        // Forward then backward should recover the original.
+        perm.apply(&mut buf, &mut tmp, true);
+        assert_ne!(buf, original, "forward permutation should shuffle");
+        perm.apply(&mut buf, &mut tmp, false);
+        assert_eq!(buf, original, "roundtrip should recover original");
+    }
+
+    #[test]
+    fn deterministic_with_same_seed() {
+        let dim = 100;
+        let mut rng1 = StdRng::seed_from_u64(42);
+        let mut rng2 = StdRng::seed_from_u64(42);
+
+        let p1 = Permutation::new(&mut rng1, dim);
+        let p2 = Permutation::new(&mut rng2, dim);
+
+        assert_eq!(p1.forward, p2.forward);
+        assert_eq!(p1.backward, p2.backward);
+    }
+
+    #[test]
+    fn dim_one_is_identity() {
+        let mut rng = StdRng::seed_from_u64(42);
+        let perm = Permutation::new(&mut rng, 1);
+
+        assert_eq!(perm.forward, vec![0]);
+        assert_eq!(perm.backward, vec![0]);
+    }
+}
--- a/lib/quantization/src/turboquant/permutation.rs
+++ b/lib/quantization/src/turboquant/permutation.rs
@@ -21,20 +21,24 @@ impl ReversibleLcg {
     fn new(seed: u64) -> Self {
         Self { state: seed }
     }
+}
 
-    /// Advance the generator and return the new state.
+impl Iterator for ReversibleLcg {
+    type Item = u64;
+
     #[inline]
-    fn next(&mut self) -> u64 {
+    fn next(&mut self) -> Option<Self::Item> {
         self.state = self.state.wrapping_mul(Self::A).wrapping_add(Self::C);
-        self.state
+        Some(self.state)
     }
+}
 
-    /// Return the current state and step the generator backward.
+impl DoubleEndedIterator for ReversibleLcg {
     #[inline]
-    fn prev(&mut self) -> u64 {
+    fn next_back(&mut self) -> Option<Self::Item> {
         let val = self.state;
         self.state = self.state.wrapping_sub(Self::C).wrapping_mul(Self::A_INV);
-        val
+        Some(val)
     }
 }
 
@@ -70,9 +74,9 @@ impl Permutation {
     /// Apply the forward permutation in-place (Fisher-Yates replay).
     pub fn permute(&self, arr: &mut [f64]) {
         debug_assert_eq!(arr.len(), self.count);
-        let mut rng = ReversibleLcg::new(self.seed);
-        for i in (1..self.count).rev() {
-            let j = Self::bounded_rand(rng.next(), i as u64 + 1) as usize;
+        let rng = ReversibleLcg::new(self.seed);
+        for (i, rand) in ((1..self.count).rev()).zip(rng) {
+            let j = Self::bounded_rand(rand, i as u64 + 1) as usize;
             arr.swap(i, j);
         }
     }
@@ -80,11 +84,9 @@ impl Permutation {
     /// Apply the inverse permutation in-place (reversed Fisher-Yates).
     pub fn unpermute(&self, arr: &mut [f64]) {
         debug_assert_eq!(arr.len(), self.count);
-        let mut rng = ReversibleLcg {
-            state: self.end_state,
-        };
-        for i in 1..self.count {
-            let j = Self::bounded_rand(rng.prev(), i as u64 + 1) as usize;
+        let rng = ReversibleLcg::new(self.end_state);
+        for (i, rand) in (1..self.count).zip(rng.rev()) {
+            let j = Self::bounded_rand(rand, i as u64 + 1) as usize;
             arr.swap(i, j);
         }
     }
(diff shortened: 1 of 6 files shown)


IvanPleshkov:
Permutations can be implemented without tmp buffers and persistent map-remap arrays.
You did a Fisher–Yates shuffle already. It makes permutations in-place, so for the forward permutation, we don't need to have a map buffer because that's how the Fisher–Yates shuffle is designed.
For backward permutation, the question is more interesting.
It seems we can restore the permutation when we just reverse the `for` loop in the alg. But we need to restore randomly generated values. It's possible. The trivial way of such a kind of generator is `random(i) = hash(seed + i)`, such an approach can predict *"which value will be generated after N `gen` calls?"*. But I propose a linear congruential generator, which is more performant.

the ai generated code snippet based on my proposals:
```
struct ReversibleLcg {
    state: u64,
}

impl ReversibleLcg {
    const A: u64 = 6_364_136_223_846_793_005;
    const C: u64 = 1_442_695_040_888_963_407;
    const A_INV: u64 = 13_877_824_140_714_322_085;

    fn new(seed: u64) -> Self {
        Self { state: seed }
    }

    fn next(&mut self) -> u64 {
        self.state = self.state.wrapping_mul(Self::A).wrapping_add(Self::C);
        self.state
    }

    fn prev(&mut self) -> u64 {
        let val = self.state;
        self.state = self.state.wrapping_sub(Self::C).wrapping_mul(Self::A_INV);
        val
    }
}

struct Permutor {
    seed: u64,
    count: usize,
    end_state: u64,
}

impl Permutor {
    fn new(seed: u64, count: usize) -> Self {
        let mut rng = ReversibleLcg::new(seed);
        for _ in 1..count {
            rng.next();
        }
        Self { seed, count, end_state: rng.state }
    }

    fn permute<T>(&self, arr: &mut [T]) {
        assert_eq!(arr.len(), self.count);
        let mut rng = ReversibleLcg::new(self.seed);
        for i in (1..self.count).rev() {
            let j = (rng.next() % (i as u64 + 1)) as usize;
            arr.swap(i, j);
        }
    }

    fn unpermute<T>(&self, arr: &mut [T]) {
        assert_eq!(arr.len(), self.count);
        let mut rng = ReversibleLcg { state: self.end_state };
        for i in 1..self.count {
            let j = (rng.prev() % (i as u64 + 1)) as usize;
            arr.swap(i, j);
        }
    }
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_lcg_reversibility() {
        let mut rng = ReversibleLcg::new(12345);
        let values: Vec<u64> = (0..10).map(|_| rng.next()).collect();

        for v in values.iter().rev() {
            assert_eq!(rng.prev(), *v);
        }
    }

    #[test]
    fn test_permute_unpermute() {
        let original: Vec<u32> = (0..100).collect();
        let permutor = Permutor::new(42, 100);

        let mut arr = original.clone();
        permutor.permute(&mut arr);
        assert_ne!(arr, original);

        permutor.unpermute(&mut arr);
        assert_eq!(arr, original);
    }

    #[test]
    fn test_edge_cases() {
        let mut arr = vec![42];
        let p = Permutor::new(0, 1);
        p.permute(&mut arr);
        assert_eq!(arr, vec![42]);
        p.unpermute(&mut arr);
        assert_eq!(arr, vec![42]);

        let original = vec![1, 2];
        let p = Permutor::new(99, 2);
        let mut arr = original.clone();
        p.permute(&mut arr);
        p.unpermute(&mut arr);
        assert_eq!(arr, original);
    }
}
```

lib/quantization/src/turboquant/rotation.rs
```diff
@@ -0,0 +1,326 @@
...
+/// ```text
+///  dim  | chunks
+/// ------|--------
+///   128 | [128]
+///   300 | [256, 32, 8, 4]
+///   700 | [512, 128, 32, 16, 8, 4]
+///  1536 | [1024, 512]
+///  4096 | [4096]
+/// ```
+pub fn compute_chunk_sizes(dim: usize) -> Vec<usize> {
```
IvanPleshkov: The `usize` as a binary number has already this decomposition. I propose to extract bits:
```
pub fn compute_chunk_sizes(dim: usize) -> Vec<usize> {
    debug_assert!(dim > 0);
    let mut sizes = Vec::with_capacity(dim.count_ones() as usize);
    let mut bits = dim;
    while bits != 0 {
        let highest = 1 << (usize::BITS - 1 - bits.leading_zeros());
        sizes.push(highest);
        bits ^= highest; // clear that bit
    }
    sizes
}
```

IvanPleshkov (changes requested):

lib/quantization/src/turboquant/rotation.rs
```diff
@@ -0,0 +1,308 @@
...
+    pub fn new(dim: usize) -> Self {
+        let chunk_sizes = compute_chunk_sizes(dim);
+
+        let chunk_norms: Vec<f64> = chunk_sizes
+            .iter()
+            .map(|&s| 1.0 / (s as f64).sqrt())
+            .collect();
+
+        let permutations: [_; N_PERMUTATIONS] =
+            std::array::from_fn(|index| Permutation::new(index as u64, dim));
```
IvanPleshkov: lets apply better seed values then 0, 1, 2

lib/quantization/src/turboquant/rotation.rs
```diff
@@ -0,0 +1,308 @@
...
+            dim,
+            chunk_sizes,
+            chunk_norms,
+        }
+    }
+
+    pub fn apply(&self, x: &[f32]) -> Vec<f32> {
+        debug_assert_eq!(x.len(), self.dim);
+
+        let mut buf: Vec<f64> = x.iter().map(|&v| f64::from(v)).collect();
```
IvanPleshkov: Let's provide tmp buffer as an argument, we can reduce allocations count because the dimension is known, and we can reuse this buffer while encoding

lib/quantization/src/turboquant/rotation.rs
```diff
@@ -0,0 +1,308 @@
...
+
+        Self {
+            permutations,
+            dim,
+            chunk_sizes,
+            chunk_norms,
+        }
+    }
+
+    pub fn apply(&self, x: &[f32]) -> Vec<f32> {
```
IvanPleshkov: let's take `&mut [f32]` and apply the result for that. We can completely avoid allocations

lib/quantization/src/turboquant/rotation.rs
```diff
@@ -0,0 +1,308 @@
...
+    while bits != 0 {
+        let highest = 1 << bits.ilog2();
+        sizes.push(highest);
+        bits ^= highest;
+    }
+    sizes
+}
+
+/// Generate a random sign vector (±1.0).
+fn generate_signs(rng: &mut StdRng, n: usize) -> Vec<f64> {
```
IvanPleshkov: Obsolete, we don't need this anymore

lib/quantization/src/turboquant/permutation.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// Modular multiplicative inverse of A mod 2^64: A * A_INV ≡ 1 (mod 2^64).
+    const A_INV: u64 = 13_877_824_140_714_322_085;
+
+    fn new(seed: u64) -> Self {
+        Self { state: seed }
+    }
+
+    /// Advance the generator and return the new state.
+    #[inline]
+    fn next(&mut self) -> u64 {
```
IvanPleshkov: Maybe we can implement `Iterator` trait for `next` and reverse iterator for `prev`. We can make loops in a permutation more rusty with `Iterator::zip`

lib/quantization/src/turboquant/permutation.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /// Regression test: with bare `% bound` on raw LCG state, the lowest bit
+    /// strictly alternates (A and C are both odd), making the i=1 Fisher-Yates
+    /// step deterministic on seed parity. This halved the reachable permutation
+    /// space — e.g. only 12/24 permutations for count=4.
+    ///
+    /// The fix uses the upper 32 bits (`>> 32`) which have much longer periods.
+    #[test]
+    fn all_small_permutations_reachable() {
```
IvanPleshkov, line 209: How long does this test run?

JojiiOfficial: 0.021s on CI and 0.01s on my machine

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: This would mean that we have to do the computations with `f32` instead of `f64`. We'd need to test first whether this difference in accuracy doesn't have a bad impact on the rotation result.

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: Done in f131383029ef7c9cde09359cbe633e790049feac

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: Done in f131383029ef7c9cde09359cbe633e790049feac

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: Removed in f131383029ef7c9cde09359cbe633e790049feac

lib/quantization/src/turboquant/permutation.rs
JojiiOfficial: Done in f131383029ef7c9cde09359cbe633e790049feac

lib/quantization/src/turboquant/rotation.rs
IvanPleshkov: Hmmm, still want to avoid allocations because we can, the only question is a good design of the function. Let's move this problem to the encoder. Rotation takes `f64` and has no idea about `f32`:
```
pub fn apply(&self, x: &mut [f64]) { .. }
```
No allocations inside, it seems we can fire it in place

lib/quantization/src/turboquant/permutation.rs
```diff
@@ -0,0 +1,244 @@
...
+        }
+    }
+
+    /// Map a 64-bit LCG output to `[0, bound)` using the upper 32 bits.
+    ///
+    /// The low bits of an LCG have short periods (bit k has period 2^k),
+    /// so bare `% bound` produces degenerate permutations. Using the
+    /// upper half avoids this.
+    fn bounded_rand(val: u64, bound: u64) -> u64 {
+        (val >> 32) % bound
```
IvanPleshkov, line 100: I like this approach and doc comment!

lib/quantization/src/turboquant/rotation.rs
```diff
@@ -0,0 +1,305 @@
+#![allow(dead_code)]
+
+use crate::turboquant::permutation::Permutation;
+
+const N_PERMUTATIONS: usize = 3;
+
+/// Random seeds for Permutation. The values were picked arbitrarily, but they must not be changed.
+const PERMUTATION_SEEDS: [u64; 3] = [654605292835415893, 8636605637963351413, 1775280196666917949];
```
IvanPleshkov, line 11: Maybe more aggressive:
DO NOT CHANGE THESE CONSTANTS!!! THEY ARE PERSISTENT!!!!

Not like that (or?), but we can say that changing these values breaks backward compatibility implicitly

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: But then we just defer the allocation to each individual vector before calling this method. I think sharing a buffer (as implemented now), would be better than re-allocating each vector.

Also WHT is _significanntly_ slower than the allocation, making its influence negligible in comparison.

lib/quantization/src/turboquant/rotation.rs
```diff
@@ -0,0 +1,305 @@
...
+        for permutation in self.permutations.iter().rev() {
+            permutation.unpermute(buf);
+            self.wht_normalized_chunks(buf);
+        }
+
+        buf.iter().map(|&v| v as f32).collect()
+    }
+
+    /// Apply WHT + normalization to variable-size chunks.
+    fn wht_normalized_chunks(&self, buf: &mut [f64]) {
```
IvanPleshkov, line 76: I guess it would be nice to mention in comment that if we apply this operation twice - we get an original vector. This detail can help future researchers of our code

lib/quantization/src/turboquant/rotation.rs
```diff
@@ -0,0 +1,305 @@
...
+            offset += size;
+        }
+        debug_assert_eq!(offset, buf.len());
+    }
+}
+
+/// In-place unnormalized Walsh-Hadamard Transform in f64.
+/// Normalization is applied externally in `apply`/`apply_inverse`.
+/// Input length must be a power of 2.
+pub fn in_place_walsh_hadamard_transform(x: &mut [f64]) {
```
IvanPleshkov, line 96: Do we need pub?

lib/quantization/src/turboquant/rotation.rs
```diff
@@ -0,0 +1,305 @@
...
+/// ```text
+///  dim  | chunks
+/// ------|--------
+///   128 | [128]
+///   300 | [256, 32, 8, 4]
+///   700 | [512, 128, 32, 16, 8, 4]
+///  1536 | [1024, 512]
+///  4096 | [4096]
+/// ```
+pub fn compute_chunk_sizes(dim: usize) -> Vec<usize> {
```
IvanPleshkov: Do we need pub?

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: Yes for benches

lib/quantization/src/turboquant/rotation.rs
IvanPleshkov: Yes, we defer the allocation and make this PR the lovely one, let it be a problem for future Jonas. At least, on a higher level, we have more possibilities to do something

IvanPleshkov: current implementation still has an allocation - returned result. and we can avoid it

lib/quantization/src/turboquant/rotation.rs
IvanPleshkov: Do we need it anymore? I mean, I find OK if we have a benchmark for the final rotation api only. Because it's hard to imagine that somebody else will bench `in_place_walsh_hadamard_transform` as an implementation detail.
WDYT? If you are strong in your position that we need this benchmark, I'm OK

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: Nope, made it private in https://github.com/qdrant/qdrant/pull/8657/commits/f3727f69ba9952c30c3e42b05139a70c7bd5fc3e

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: https://github.com/qdrant/qdrant/pull/8657/commits/f3727f69ba9952c30c3e42b05139a70c7bd5fc3e

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: https://github.com/qdrant/qdrant/pull/8657/commits/f3727f69ba9952c30c3e42b05139a70c7bd5fc3e

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: It could be helpful to compare scalar version vs SIMD in future. I don't have a strong opinion on this though so we can drop it if you prefer

IvanPleshkov: Okay, lets be positive and expect that we will do simd for that =)

lib/quantization/src/turboquant/rotation.rs
JojiiOfficial: Made in-place in https://github.com/qdrant/qdrant/pull/8657/commits/2facafd00cd0adf9710e55e87b73907c0efe4f4f

IvanPleshkov (approved):
Love it!

PR #8667, opened 2026-04-13
[IoUringReadIter] Buffer out-of-order items

coszio:
**Disclaimer**: I know I know. We want to reduce the number of allocations as much as possible. This PR introduces a bit of overhead by using a hashmap as buffer, but I don't think it is that bad because:
- Shares the same hashmap for the iterator, so no new allocations unless we grow it.
- Uring already allocates a vector, in this case we are moving the pointer to the hashmap (I think?).
- Even though we are introducing hashing, the hashed type is `usize`, so should be quick, and IO costs a lot more anyway.

Returning things out-of-order makes handling the iterator a lot more complex ([example](https://github.com/qdrant/qdrant/pull/8507/changes#diff-1eb15cd42d6e7ef826b617ff2534202007c861edbe53969478b6828d5bd970bcR444-R532)), and makes users of UniversalRead iterator having to handle the special case of IoUring for every case.

For iterators, we almost always have to wait until the next sequential item, even if we are capable of processing others in the meantime. So I think there is no big benefit of complicating the callsites everywhere, when we can contain it to the read iterator itself.

This PR ensures items are in order by:
- always draining the completion queue.
- checking if the next sequential item is there.
- saturating the submission queue with more items.

### DIFF
--- a/lib/common/common/src/universal_io/io_uring/read_iter.rs
+++ b/lib/common/common/src/universal_io/io_uring/read_iter.rs
@@ -2,12 +2,16 @@ use std::iter;
 use std::marker::PhantomData;
 
 use ::io_uring::types::Fd;
+use ahash::AHashMap;
 
 use super::*;
 
 pub struct IoUringReadIter<T: 'static, Meta, I: Iterator> {
     ranges: iter::Peekable<I>,
-    runtime: IoUringRuntime<'static, T, Meta>,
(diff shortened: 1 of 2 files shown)


coszio:
After discussion with the team, if we were to use this, we would prefer to make it a wrapper, rather than changing the original implementation.

PR #8676, opened 2026-04-14
feat/internal-grpc-auth

dancixx:
### Cross-node JWT scenarios

  | Scenario | Expected | Status |
  |----------|----------|--------|
  | Client → jwt(A) → Node(A,B) | ✅ Accept | ✅ `api_key` match |
  | Client → jwt(A) → Node(A only) | ✅ Accept | ✅ `api_key` match |
  | Client → jwt(B) → Node(A,B) | ✅ Accept | ✅ `alt_api_key` match |
  | Client → jwt(B) → Node(A only, no B) | ❌ Reject | ✅ Correctly rejected |
  | Node(A,B) → internal → Node(A only) | ✅ Accept | ✅ `PoolInterceptor` sends primary `api_key` (A), not B |

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/src/tonic/api/collections_internal_api.rs
+++ b/src/tonic/api/collections_internal_api.rs
@@ -268,9 +268,10 @@ impl CollectionsInternal for CollectionsInternalService {
 
     async fn get_shard_memory_report(
         &self,
-        request: Request<GetShardMemoryReportRequest>,
+        mut request: Request<GetShardMemoryReportRequest>,
     ) -> Result<Response<GetShardMemoryReportResponse>, Status> {
         validate_and_log(request.get_ref());
+        let auth = extract_auth(&mut request);
 
         let timing = Instant::now();
         let GetShardMemoryReportRequest {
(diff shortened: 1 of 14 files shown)


src/tonic/api/collections_internal_api.rs
```diff
@@ -10,24 +10,20 @@ use api::grpc::qdrant::{
...
-            AccessRequirements::new(),
-            "internal_collection_access",
-        )
-        .map_err(Status::from)
+use crate::tonic::auth::extract_auth;
+
+fn access_pass<'a>(auth: &'a Auth, collection_name: &'a str) -> Result<CollectionPass<'a>, Status> {
+    auth.check_collection_access(
+        collection_name,
+        AccessRequirements::new(),
```
generall: since we now actually check access, it should check full permission here.

generall (changes requested):
This change seems to enable new behaviour unconditionally (as soon as regular API key in configued).

Since we are doing rolling version upgrade, it can be that some nodes are migrated to use API key in internal requests, while others are not.

This will render cluster dead for the period of migration, as nodes won't be able to reach out to each other.

For proper migration we need following:

- send API key in requests unconditionally
- check API in internal API only if special setting option is enabled

generall (approved):

PR #8678, opened 2026-04-14
TQ variant for `QuantizedVectorStorage`

IvanPleshkov:
This PR is a part of TQ integration.

It adds a new TQ variant for `QuantizedVectorStorage`.

Also, it adds a tmp GPU fallback. In case of TQ, we fallback to `f16` floats.

Open questions:
- do we need to enable appendable TQ for each bits count?
- do we need to rescore by default?

### DIFF
--- a/lib/segment/src/vector_storage/quantized/quantized_vectors.rs
+++ b/lib/segment/src/vector_storage/quantized/quantized_vectors.rs
@@ -10,6 +10,7 @@ use common::generic_consts::{Random, Sequential};
 use common::types::PointOffsetType;
 use quantization::encoded_vectors_binary::EncodedVectorsBin;
 use quantization::encoded_vectors_u8::ScalarQuantizationMethod;
+use quantization::turboquant::EncodedVectorsTQ;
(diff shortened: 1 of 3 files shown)


JojiiOfficial:
This PR also contains the changes from https://github.com/qdrant/qdrant/pull/8677
Could you maybe rebase it?

lib/segment/src/vector_storage/quantized/quantized_vectors.rs
```diff
@@ -136,6 +151,9 @@ pub enum QuantizedVectorStorage {
     BinaryRam(EncodedVectorsBin<u128, QuantizedRamStorage>),
     BinaryMmap(EncodedVectorsBin<u128, QuantizedMmapStorage>),
     BinaryChunkedMmap(EncodedVectorsBin<u128, QuantizedChunkedMmapStorage>),
+    TQRam(EncodedVectorsTQ<QuantizedRamStorage>),
```
IvanPleshkov, line 154: This is the key change

JojiiOfficial (approved):

PR #8681, opened 2026-04-15
simplify uio geo index

generall:
This PR proposes an alternative to returning actual lazy iterators from payload indexes.

Reasons:

- code is very complex. Auto-batching/chunking logic on top of CoWs e.t.c. makes it really hard to read or write UIO migrations for payload indexes.
- Lazy iterators froces `Item=Result<...>`, which might be a performance concern, which affects all types of indexes, even in-memory ones
- Pre-load is supposed to be better for latency. Lazy iterators are by definition doing sequential IO, and it requires a lot of extra processing to make this sequential IO efficient. With full prefetching - efficient  batched IO is easier.

### DIFF
--- a/lib/common/common/src/iterator_ext/ordering_iterator.rs
+++ b/lib/common/common/src/iterator_ext/ordering_iterator.rs
@@ -1,0 +1,263 @@
+use std::collections::BTreeMap;
+use std::convert::Infallible;
+
+/// Trait that abstracts over the item type of [`OrderingIterator`], allowing it
+/// to accept both plain `(usize, T)` items and `Result<(usize, T), E>` items
+/// from a single implementation.
+///
+/// Errors produced by the upstream iterator bypass the ordering buffer and are
+/// forwarded immediately in the order they were observed.
+pub trait OrderedItem: Sized {
+    type Value;
+    type Error;
+
+    fn split(self) -> Result<(usize, Self::Value), Self::Error>;
+    fn from_value(idx: usize, value: Self::Value) -> Self;
+    fn from_error(error: Self::Error) -> Self;
+}
+
+impl<T> OrderedItem for (usize, T) {
+    type Value = T;
+    type Error = Infallible;
+
+    fn split(self) -> Result<(usize, T), Infallible> {
+        Ok(self)
+    }
+
+    fn from_value(idx: usize, value: T) -> Self {
+        (idx, value)
+    }
+
+    fn from_error(error: Infallible) -> Self {
+        match error {}
+    }
+}
+
+impl<T, E> OrderedItem for Result<(usize, T), E> {
+    type Value = T;
+    type Error = E;
+
+    fn split(self) -> Result<(usize, T), E> {
+        self
+    }
+
+    fn from_value(idx: usize, value: T) -> Self {
+        Ok((idx, value))
+    }
+
+    fn from_error(error: E) -> Self {
+        Err(error)
+    }
+}
+
+/// An iterator adapter that takes an iterator of `(idx: usize, element: T)`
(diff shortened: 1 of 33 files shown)


lib/common/common/src/iterator_ext/ordering_iterator.rs
```diff
@@ -0,0 +1,263 @@
...
+                self.next_id += 1;
+                return Some(I::Item::from_value(idx, value));
+            }
+
+            let Some(item) = self.iter.next() else {
+                // Upstream iterator is exhausted: drain the buffer in ascending
+                // order, skipping over any gaps between ids.
+                let (idx, value) = self.buffer.pop_first()?;
+                self.next_id = idx + 1;
+                return Some(I::Item::from_value(idx, value));
```
coszio, line 116: I propose to enumerate the upstream ourselves. It has two benefits:

1. This codepath should never run. If the upstream has already been exhausted, it means the buffer has every remaining item, and the first part of the loop should catch it. 

2. If there is at least one gap, it means the buffer will grow with every remaining item from upstream.

We'd lose the ability to test easily though

lib/segment/src/index/field_index/geo_index/mod.rs
```diff
@@ -1709,4 +1691,112 @@ mod tests {
             }
         }
     }
+
+    /// Cross-check that [`StoredGeoMapIndex::all_points`] produces the same
+    /// point set as the mutable and immutable index implementations for a
+    /// variety of prefix configurations (single region, disjoint regions,
+    /// overlapping ancestor/descendant prefixes).
+    #[test]
+    fn test_all_points_congruence() {
```
coszio, line 1701: We should also check for point deletion in this test, see https://github.com/qdrant/qdrant/pull/8681/changes#r3083332573

coszio (approved):
LGTM, but I would like @xzfc to have a look at `all_points` for good measure

lib/common/common/src/iterator_ext/ordering_iterator.rs
coszio: I was wrong. The ordering idx is needed because it is provided before being used by uring

lib/segment/src/index/field_index/geo_index/mmap_geo_index.rs
```diff
@@ -433,57 +435,56 @@ impl<S: StoredGeoMapIndexStorage> StoredGeoMapIndex<S> {
...
-        // │<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<│.........│>>>>>>>>>>>>>>>>>>>>>>│
-        // │ entries < geohash_prefix      │ result  │ entries past the end │
-        // └───────────────────────────────┴─────────┴──────────────────────┘
+        geo_hashes.sort_unstable();
+        geo_hashes.dedup();
+        // Drop any prefix that is already subsumed by an earlier (shorter)
+        // one. After this pass the remaining prefixes are pairwise
+        // incomparable — no prefix is a prefix of another — so at most one of
+        // them can match any given entry.
+        geo_hashes.dedup_by(|later, earlier| later.starts_with(*earlier));
```
timvisee: The `dedup_by` also deduplicates equal entries, making the first dedup redundant:


```suggestion
        // Drop any prefix that is already subsumed by or equal to an earlier (shorter)
        // one. After this pass the remaining prefixes are pairwise
        // incomparable — no prefix is a prefix of another — so at most one of
        // them can match any given entry.
        geo_hashes.dedup_by(|later, earlier| later.starts_with(*earlier));
```

xzfc (approved):

timvisee (approved):
Great simplification.

PR #8692, opened 2026-04-16
TurboQuant basic Quantization + Encoding

JojiiOfficial:
Implements basic TurboQuant quantization (Cosine only).

### DIFF
--- a/lib/quantization/src/turboquant/lloyd_max.rs
+++ b/lib/quantization/src/turboquant/lloyd_max.rs
@@ -1,5 +1,6 @@
 #![allow(dead_code)]
 
+/// 1-bit (2 centroids): Lloyd-Max for N(0,1)
 const CENTROIDS_1BIT: [f32; 2] = [-0.797_884_6, 0.797_884_6];
 const CENTROIDS_1BIT_BOUNDARIES: [f32; 1] = calculate_boundaries(CENTROIDS_1BIT);
 
@@ -41,6 +42,13 @@ pub fn get_centroids(bits: u8) -> &'static [f32] {
     }
 }
 
+/// Return the centroid decision boundaries for a given bit-width (1/2/4).
+///
+/// Boundaries are the midpoints between consecutive centroids; a value maps to
+/// centroid `i` if it falls between boundary `i-1` and boundary `i`.
+///
+/// # Panics
+/// Panics if `bits` is not in [1, 2, 4].
 #[inline]
 pub fn get_centroid_boundaries(bits: u8) -> &'static [f32] {
     match bits {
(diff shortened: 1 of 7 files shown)


lib/quantization/src/turboquant/quantization.rs
```diff
@@ -0,0 +1,504 @@
+use crate::DistanceType;
+use crate::turboquant::rotation::HadamardRotation;
+use crate::turboquant::{Metadata, TQBits, TQMode};
+
+/// Quantize vectors using TurboQuant.
+pub struct TurboQuantizer {
```
IvanPleshkov, line 6: Why do we need `TurboQuantizer` as a separate structure? Can we implement these methods in the `EncodedVectorsTQ` and not introduce a new entity?

lib/quantization/src/turboquant/lloyd_max.rs
```diff
@@ -1,25 +1,60 @@
...
+
+/// Return the centroid decision boundaries for a given bit-width (1/2/4).
+///
+/// Boundaries are the midpoints between consecutive centroids; a value maps to
+/// centroid `i` if it falls between boundary `i-1` and boundary `i`.
+///
+/// # Panics
+/// Panics if `bits` is not in [1, 2, 4].
+#[inline]
+pub fn get_centroid_boundaries(bits: u8) -> &'static [f32] {
```
IvanPleshkov: You call get_centroid_boundaries inside TQBits structure, converting it into an integer. You can do match statement directly from enum variants and avoid `unreachable!` macro

IvanPleshkov: The same for get_centroids

lib/quantization/src/turboquant/quantization.rs
JojiiOfficial: I think it's cleaner to have it separated. `EncodedVectorsTQ` already does a lot (storage, metadata, loading, scoring..).
A separate structure that handles the quantization itself is much easier to test and benchmark e2e if it's a separate struct.

lib/quantization/src/turboquant/lloyd_max.rs
JojiiOfficial: Changed in https://github.com/qdrant/qdrant/pull/8692/commits/1fd1fa4b1e88c8fe719088bf38988b7c83b20dc2

lib/quantization/src/turboquant/quantization.rs
IvanPleshkov: THB dont like this approach because of the one-shot new entity. But I don't have strong arguments, it's more about personal preferences, resolve

IvanPleshkov (approved):

PR #8694, opened 2026-04-16
Apply new clippy lints for rust 1.95.0

JojiiOfficial:


lib/storage/src/content_manager/snapshots/recover.rs
```diff
@@ -258,17 +258,16 @@ async fn _do_recover_from_snapshot(
...
+            Some(state) if state != &recovery_state => {
+                toc.send_set_replica_state_proposal(
+                    collection_pass.to_string(),
+                    this_peer_id,
+                    *shard_id,
+                    recovery_state,
+                    None,
+                )?;
             }
+            Some(_) => {}
```
xzfc, line 270: Style nit: use if-let chain:

```rust
if let Some(state) = local_shard_state
    && state != &recovery_state
{
    toc.send_set_send_set_replica_state_proposal(
        …
    )?;
}
```

(and the same in `lib/segment/src/index/query_estimator.rs`)

xzfc (approved):

JojiiOfficial:
Closing in favor of #8695

PR #8697, opened 2026-04-16
Clear shard data before recovery

timvisee:
Pending tests:

- [ ] What happens if you send a read/write request to a recovering shard now?
- [ ] What happens with the shard on restart, does it keep it's recovering state?

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/collection/src/collection/snapshots.rs
+++ b/lib/collection/src/collection/snapshots.rs
@@ -266,6 +266,7 @@ impl Collection {
                 &self.path,
                 shard_id,
                 cancel,
+                None,
             )
             .await?;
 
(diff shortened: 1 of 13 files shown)


timvisee:
One problem with this approach is that we now hold a lock for much longer.

Previously we would download the snapshot. Then take a lock, move files and load it.

Now we also have to hold the lock for the duration of the snapshot download, which can take a long time. We cannot download it first, because it conflicts with not wanting to use the extra storage (headroom) for it.

timvisee:
Superseded by <https://github.com/qdrant/qdrant/pull/8782>

PR #8700, opened 2026-04-16
Fix flaky test_deferred_points_dedup_after_optimization

leohenon:
Addresses #8504

`wait_optimization` only checked for `ShardStatus::Green`, which can be true briefly before proxy segments are fully removed, causing the test to access an unimplemented proxy method.


timvisee (approved):
Thanks! 😄

PR #8685, opened 2026-04-15
Remove `AsyncRawScorer`

ffuugoo:
This PR refactors `ImmutableDenseVectors` and `DenseVectorStorageImpl`, so that default `RawScorerImpl` correctly uses io_uring, and we can get rid of `AsyncRawScorerImpl`.

I've tested all changes against `scorer_mmap`/`vector_search` benchmark, no noticeable impact on performance as far as I can tell.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [x] Have you formatted your code locally using `cargo +nightly fmt --all` command prior to submission?
3. [x] Have you checked your code using `cargo clippy --workspace --all-features` command?

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #8702, opened 2026-04-16
Refactor `Multi*QueryScorer` for (future) io_uring support

ffuugoo:
Similar to #8685, this PR adds a code path, that correctly (i.e., efficient for io_uring) utilizes universal I/O to `Multi*QueryScorer`s/`MultiVectorStorage`/`ChunkedVectors`.

**TODO:**
- [ ] benchmarks

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


generall (approved):

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -142,6 +159,33 @@ impl<T: PrimitiveVectorElement> MultiVectorStorage<T> for AppendableMmapMultiDen
...
+    where
+        F: FnMut(usize, TypedMultiDenseVectorRef<'_, T>),
+    {
+        // Collect multi-vector offsets
+        let mut point_indexes = Vec::with_capacity(keys.len());
+        let mut offsets = Vec::with_capacity(keys.len());
+
+        for (point_index, offset) in self.offsets.iter(keys) {
+            let &[offset] = offset.as_ref() else {
+                unreachable!();
```
xzfc, line 167: Readability nit: I'd like to see a comment why this is unreachable. E.g.,
```rust
                // <PointOffsetType as VectorOffset>::multi_vector_count is
                // always 1, so we expect to get exactly one value.
```

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
```diff
@@ -42,6 +43,22 @@ pub struct MultivectorMmapOffset {
     capacity: u32,
 }
 
+impl fmt::Display for MultivectorMmapOffset {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}x{}", self.offset, self.count)
+    }
+}
```
xzfc: I think the notation like `11721x5` is confusing.

But it seems it's not used anywhere, so I propose to drop `Display` at all:

```diff
-pub trait VectorOffset: Copy + fmt::Display + fmt::Debug {
+pub trait VectorOffset: Copy + fmt::Debug {

-impl fmt::Display for MultivectorMmapOffset {
-    …
-}
```

It was used only in `panic!("Vector {key} not found")` which is removed in this PR.

lib/segment/src/vector_storage/query_scorer/multi_metric_query_scorer.rs
```diff
@@ -111,24 +109,24 @@ impl<
...
-            .vector_storage
-            .get_batch_multi(ids, &mut vectors[..ids.len()]);
-
-        let total_read = vectors.iter().map(|v| v.as_ref().vectors_count()).sum();
+        let vectors_read = self.hardware_counter.vector_io_read();
+        self.vector_storage
+            .for_each_in_batch_multi(ids, |idx, vector| {
+                vectors_read.incr_delta(vector.vectors_count());
+                scores[idx] = self.score_ref(vector);
+            });
```
xzfc, line 120: Quick benchmark shows that this change doesn't affect the performance which is good.

(not related to this PR →) Tho I'm still wonder about similar changes for non-multi-vector usages of the same pattern (`[MaybeUninit; VECTOR_READ_BATCH_SIZE]`).

xzfc (approved):

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
ffuugoo: Done

lib/segment/src/vector_storage/multi_dense/appendable_mmap_multi_dense_vector_storage.rs
ffuugoo: Good catch. Done.

PR #8706, opened 2026-04-17
Sort HashMap keys for deterministic iteration

agourlay:
Sort `HashMap` / `AHashMap` keys at five call sites where randomized iteration order was leaking into HTTP responses, consensus operations, and ordered side-effects. 

Each fix collects into a `Vec` and calls  `sort_unstable()` before the existing usage.

## Changes

  - **`persistent.rs`** — `get_cluster_metadata_keys()` is served directly over `GET /cluster/metadata/keys`; response order was random.
  - **`collection_container.rs`** — during consensus-snapshot reconciliation, `delete_collection(...)` was invoked in random order over locally-existing collections.
  - **`shard_holder/mod.rs`** — resharding state-apply dropped extra shards in random order; the intermediate `HashSet` was never used for membership, replaced with a sorted `Vec<ShardId>`.
  - **`single_to_cluster.rs`** — replica peers per shard were pushed into the `placement` vector of a `CreateShardKey` consensus op in random order.
  - **`state_management.rs`** — `apply_shard_info` passed `shard_info.replicas.keys()` to `create_replica_set(...)` in random order.

## Future work

Further determinism candidates were found but I want to keep this PR small and not controversial.

### DIFF
--- a/lib/storage/src/content_manager/consensus/persistent.rs
+++ b/lib/storage/src/content_manager/consensus/persistent.rs
@@ -262,7 +262,9 @@ impl Persistent {
     }
 
     pub fn get_cluster_metadata_keys(&self) -> Vec<String> {
-        self.cluster_metadata.keys().cloned().collect()
+        let mut keys: Vec<String> = self.cluster_metadata.keys().cloned().collect();
+        keys.sort_unstable();
+        keys
     }
 
     pub fn get_cluster_metadata_key(&self, key: &str) -> serde_json::Value {
--- a/lib/collection/src/collection/state_management.rs
+++ b/lib/collection/src/collection/state_management.rs
@@ -226,7 +226,8 @@ impl Collection {
                         .await?;
                 }
                 None => {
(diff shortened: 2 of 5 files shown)


lib/storage/src/content_manager/consensus/persistent.rs
```diff
@@ -262,7 +262,9 @@ impl Persistent {
     }
 
     pub fn get_cluster_metadata_keys(&self) -> Vec<String> {
-        self.cluster_metadata.keys().cloned().collect()
+        let mut keys: Vec<String> = self.cluster_metadata.keys().cloned().collect();
+        keys.sort_unstable();
+        keys
```
timvisee, line 267: This one does not affect consensus behavior, but only user responses, doesn't it?

timvisee (approved):

lib/storage/src/content_manager/consensus/persistent.rs
agourlay: AFAIK we are not relying on the API internally.

So yes only users response here

Issue #8278, opened 2026-03-03
About Add fuzzy Match Support for Text Fields in Payload Filter


TY0909:
## Overview

This issue proposes adding **fuzzy match** support for text fields within the payload filter section. Fuzzy matching allows queries to tolerate minor spelling variations, typos, and OCR errors by leveraging the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) metric to find terms that are "close enough" to the query term.

This is a purely additive change — all existing query types and behaviors remain unaffected.

---

## Motivation

Currently, payload filter text matching requires an exact token match. This is too strict in many real-world scenarios:

- User-submitted search queries may contain typos (e.g., `"recieve"` instead of `"receive"`).
- Data ingested from OCR pipelines may have character-level noise.
- Multilingual content may have transliteration inconsistencies.

Fuzzy matching addresses all of these cases by expanding a query term into a set of candidate terms within a configurable edit distance.

---

### 1. New Query Type Variants

We will extend `TextQueryType` with new fuzzy variants. These mirror the existing variants but apply fuzzy term expansion before matching:

| New Variant     | Fuzzy counterpart of |
|-----------------|----------------------|
| `FuzzyText`     | `Text` (all tokens must match) |
| `FuzzyTextAny`  | `TextAny` (any token must match) |
| `FuzzyPhrase`   | `Phrase` (ordered proximity match) |

### 2. Extending the Query Dispatch Branch

The core change is to add new arms to the existing `match query_type` block inside the `FieldIndex::FullTextIndex` handler. This keeps the change fully isolated:

```rust
FieldIndex::FullTextIndex(full_text_index) => {
    let query_opt = match query_type {
        // Existing branches — unchanged
        TextQueryType::Phrase    => full_text_index.parse_phrase_query(&text, &hw_counter),
        TextQueryType::Text      => full_text_index.parse_text_query(&text, &hw_counter),
        TextQueryType::TextAny   => full_text_index.parse_text_any_query(&text, &hw_counter),

        // New fuzzy branches
        TextQueryType::FuzzyText    => full_text_index.parse_fuzzy_text_query(&text, &fuzzy_params, &hw_counter),
        TextQueryType::FuzzyTextAny => full_text_index.parse_fuzzy_text_any_query(&text, &fuzzy_params, &hw_counter),
[...]

Issue #8670, opened 2026-04-14
[Turbo Quant] tracking issue


IvanPleshkov:
This is a tracking issue for Turbo Quant implementation.

Design doc: https://www.notion.so/qdrant/TurboQuant-design-doc-334674779d33803ab94bca91863cf066
Reference TQ integration: https://github.com/qdrant/qdrant/pull/8544
Reference implementations:
https://github.com/qdrant/qdrant/pull/8567
https://github.com/qdrant/qdrant/pull/8568

- [x] Hadaramd Rotations with permutations https://github.com/qdrant/qdrant/pull/8657
- [x] Precomputed codebooks with Lloyd Max alg https://github.com/qdrant/qdrant/pull/8672
- [ ] Basic TQ implementation
  - [x] Empty TQ struct with `EncodedVectors` implementation https://github.com/qdrant/qdrant/pull/8677
  - [ ] Single vector encoding (cosine only) https://github.com/qdrant/qdrant/pull/8692
  - [ ] Metrics, scores
    - [ ] Cosine scoring with f32 query
    - [ ] Cosine internal scoring
    - [ ] Add dot product for query scoring
    - [ ] Add dot product for internal scoring
    - [ ] Add L2 for query scoring
    - [ ] Add L2 for internal scoring
    - [ ] Add L1 placeholder with complete decoding
  - [ ] TQ Plus
    - [ ] Add renormalization support
    - [ ] Collect rotated vectors stats
    - [ ] Add precalculated values and modify scoring
  - [ ] 1.5bit quantization
- [ ] Integration:
  - [ ] REST + Grpc API https://github.com/qdrant/qdrant/pull/8686
  - [ ] TQ in segment:
    - [x] Add TQ variant into `QuantizedVectorStorage` https://github.com/qdrant/qdrant/pull/8678
    - [ ] Integrate TQ encode https://github.com/qdrant/qdrant/pull/8701
    - [ ] Integrate TQ appendable https://github.com/qdrant/qdrant/pull/8701
    - [ ] Integrate TQ load https://github.com/qdrant/qdrant/pull/8701
  - [ ] Edge python bindings https://github.com/qdrant/qdrant/pull/8705
  - [x] GPU fallback https://github.com/qdrant/qdrant/pull/8678
- [ ] SIMD support
  - [ ] Hadamard SIMD
  - [ ] 4bit SIMD
    - [ ] Precomputed scalar quantized codebooks
    - [ ] `avx` implementation
[...]

PR #8707, opened 2026-04-17
Feature: Lexical Fuzzy Filter for `FullTextIndex` and `StrDist` Formula Scoring

VainJoker:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

---

## Summary

Related to #8278

This PR adds fuzzy matching support for text payload filters backed by `FullTextIndex`.

The new `match.fuzzy` condition allows text queries to tolerate typos, OCR noise, and minor lexical variation by expanding query tokens to nearby vocabulary terms under a bounded Levenshtein distance.

Alongside that main capability, this PR also adds `match.wildcard` for deterministic token-pattern expansion using `*` and `?`.

In addition, this PR adds a `str_dist` formula expression so that typo-tolerant filtering can be paired with payload-aware reranking.

This change is strictly additive. Existing exact `text`, `text_any`, and `phrase` filters continue to parse and execute through the same exact-match paths as before. Lexical expansion is only activated when the caller explicitly uses `match.fuzzy` or `match.wildcard` on a text field whose index was created with fuzzy support enabled.

## Motivation
Today, payload text filtering is exact at the token level. That is often too strict for real-world inputs:

- user-entered queries may contain typos, such as `recieve` instead of `receive`
- OCR pipelines often introduce single-character substitutions, insertions, or deletions
- transliteration and spelling variation can cause near-matches to be missed

This PR addresses that gap primarily through fuzzy matching.

Fuzzy is the main retrieval feature here because the target problem is approximate lexical matching: recover useful candidates even when the query token is misspelled, noisy, or slightly transformed.

Wildcard is included as a secondary lexical operator for cases where the caller already knows part of the token shape and wants explicit pattern matching, such as `inter*`, `*scope`, or `s?aphire`.

### 1. Create a text payload index with fuzzy support

Fuzzy filtering works on text fields indexed with `fuzzy_matching: true`. Wildcard uses the same lexical expansion structures. If phrase-level fuzzy search is needed, `phrase_matching: true` must also be enabled.

Wildcard is term-level rather than phrase-level in the current implementation: the pattern is expanded against vocabulary terms and the resulting matches are evaluated with union semantics.

```http
PUT /collections/documents/index?wait=true
Content-Type: application/json

{
	"field_name": "title",
	"field_schema": {
		"type": "text",
		"tokenizer": "word",
		"lowercase": true,
		"phrase_matching": true,
		"fuzzy_matching": true
	}
}
```

### 2. Send a fuzzy filter query

`match.fuzzy` accepts an array of fuzzy clauses. Each clause can be one of:

- `{"text": ...}` for all-token semantics
- `{"text_any": ...}` for OR semantics
- `{"phrase": ...}` for ordered phrase semantics

Example:

```http
POST /collections/documents/points/query
Content-Type: application/json

{
	"query": [0.12, -0.44, 0.98, 0.07],
	"filter": {
		"must": [
			{
				"key": "title",
				"match": {
					"fuzzy": [
						{
							"text": "quikly saphire",
							"params": {
								"max_edits": 1,
								"prefix_length": 1,
								"max_expansions": 30
							}
						},
						{
							"text": "falcon",
							"params": {
								"max_edits": 2,
								"prefix_length": 1,
								"max_expansions": 30
							}
						}
					]
				}
			}
		]
	},
	"limit": 10,
	"with_payload": true
}
```

Important notes:

- `params.max_edits` must be greater than `0` to enable actual fuzzy expansion. If `params` is omitted, the current default is `max_edits = 0`, which behaves like exact matching.
- Multiple entries inside one `match.fuzzy` array must all be the same clause family: all `text`, all `text_any`, or all `phrase`.
- Wildcard patterns operate on single indexed terms rather than on multi-token text fragments.

### 2.1 What multiple fuzzy clauses do

`match.fuzzy` accepts multiple clauses so the caller can compose a larger logical unit from several homogeneous fuzzy fragments.

Behavior by clause family:

- multiple `text` clauses are concatenated into one larger AND-style token sequence
- multiple `text_any` clauses are flattened into one larger OR-style token set
- multiple `phrase` clauses are concatenated in order into one larger phrase query

Examples:
At execution time, wildcard remains a binary lexical filter. It expands matching vocabulary terms and checks membership, but does not preserve phrase ordering or emit a graded similarity score, most databases typically go with this option.
```json
{
	"match": {
		"fuzzy": [
			{ "text": "quickly sapphire", "params": { "max_edits": 1 } },
			{ "text": "falcon", "params": { "max_edits": 2 } }
		]
	}
}
```

This behaves like one `text` query whose expanded token groups must all match.

```json
{
	"match": {
		"fuzzy": [
			{ "text_any": "quickly slowly", "params": { "max_edits": 1 } },
			{ "text_any": "falcon rabbit", "params": { "max_edits": 2 } }
		]
	}
}
```

This behaves like one `text_any` query over the union of all expanded tokens.

```json
{
	"match": {
		"fuzzy": [
			{ "phrase": "quickly sapphire", "params": { "max_edits": 1 } },
			{ "phrase": "falcon", "params": { "max_edits": 2 } }
		]
	}
}
```

This behaves like one ordered phrase query equivalent to `"quickly sapphire falcon"` after expansion.

Mixed clause families are not combined. For example, combining one `text` clause with one `phrase` clause in the same `match.fuzzy` array does not produce a parsed query, so the condition evaluates to an empty result.

### 2.2 Wildcard as a Complementary Capability

Wildcard matching is not a replacement for fuzzy matching; rather, it serves as a complementary enhancement. It enables callers who know part of a token’s structure to impose precise pattern constraints.

- `*` and `?` retain their special semantics during query parsing, rather than being treated as ordinary characters
- expansion is driven by explicit pattern matching over the vocabulary, rather than edit distance
- wildcard applies to single vocabulary terms, not to multi-token phrases
- wildcard can express structural constraints such as prefixes, suffixes, and masked positions, which cannot be efficiently or accurately modeled by fuzzy matching
- wildcard expansion currently produces union-style term matching rather than phrase-aware matching

As a result, wildcard is exposed as an independent lexical operator. It enhances the query capabilities alongside fuzzy matching, with each mode serving a distinct purpose and remaining orthogonal.

### 3. Use `str_dist` to rerank filtered candidates

The universal query API can run a formula over a prefetched candidate set. This makes it possible to first recall candidates with vector search plus fuzzy filtering, and then rerank them by string similarity on a payload field.

```http
POST /collections/documents/points/query
Content-Type: application/json

{
	"prefetch": {
		"limit": 100,
		"filter": {
			"must": [
				{
					"key": "title",
					"match": {
						"fuzzy": [
							{
								"text": "quikly saphire falcon",
								"params": {
									"max_edits": 1,
									"prefix_length": 1,
									"max_expansions": 30
								}
							}
						]
					}
				}
			]
		}
	},
	"query": {
		"formula": {
			"str_dist": {
				"field": "title",
				"query": "quikly saphire falcon",
				"func": "jaro_winkler"
			}
		}
	},
	"limit": 10,
	"with_payload": true
}
```

## Architecture & Implementation

High-level flow:

- parse the filter as `Match::Fuzzy` or `Match::Wildcard`
- dispatch into the corresponding lexical expansion path
- expand fuzzy tokens or wildcard patterns through the vocabulary index
- map expanded terms back to internal token IDs and build the parsed full-text query
- evaluate the query with the original AND / OR / phrase semantics preserved
- optionally rerank the filtered candidate set with `str_dist`

### 2.1 Lexical Expansion Pipeline

Fuzzy is the primary lexical expansion path in this PR. Wildcard follows the same broad execution shape, but remains a separate auxiliary mode.

Fuzzy path:

`Filter input -> Match::Fuzzy -> tokenization -> token expansion via FuzzyIndex -> token ID mapping -> normal set/phrase evaluation`

Wildcard path:

`Filter input -> Match::Wildcard -> pattern normalization -> wildcard expansion via FuzzyIndex -> token ID mapping -> normal set/phrase evaluation`

The key difference is the expansion rule:

- wildcard expands from an explicit pattern over vocabulary terms
- fuzzy expands from a normalized token under bounded edit distance

An additional semantic difference is that wildcard currently remains term-level after expansion: matched vocabulary terms are merged into a single union token set.

Semantics are preserved strictly:

- `text` keeps AND semantics: every query token must resolve to a non-empty expansion set
- `text_any` keeps OR semantics: all expansions are merged into one union token set
- `phrase` keeps ordering semantics: expanded token groups are matched in order against the stored token sequence

The parser intentionally rejects ambiguous or degraded cases instead of silently weakening the query:

- empty `match.fuzzy` arrays resolve to no parsed query and therefore no matches
- mixing `text`, `text_any`, and `phrase` in one `MatchFuzzy` resolves to no parsed query and therefore no matches
- if the fuzzy index is unavailable, parsing fails
- if a `text` or `phrase` token expands to an empty set, parsing fails
- if the input produces no query tokens, parsing fails

Another important guardrail is that very short terms do not participate in fuzzy expansion. Tokens whose character length is less than or equal to the tokenizer's minimum token length fall back to exact token lookup. This prevents extremely short tokens from exploding candidate sets.

When multiple fuzzy clauses are supplied, the parser first verifies that they all belong to the same family, then merges their expanded token groups before building the final parsed query:

The final internal query shape depends on whether expansion stayed exact or actually produced fuzzy groups:

- `text + text + ...` becomes `AllTokens` when every merged group resolves to a single exact token, otherwise it becomes `FuzzyAllTokens`
- `text_any` + `text_any` + ... always merges into one flat union of tokens; it remains AnyTokens in the general case, switches to FuzzyAnyTokens when fuzzy text_any matching is present.
- `phrase + phrase + ...` becomes `Phrase` when every position resolves to exactly one token, otherwise it becomes `FuzzyPhrase`

### 2.2 Why wildcard needs its own path

Wildcard is not just “another fuzzy mode”. It needs its own parser and expansion logic because the query carries explicit metacharacter semantics.

- `*` and `?` must survive query parsing rather than being treated as ordinary token characters
- wildcard expansion is driven by pattern matching over vocabulary terms, not by computing edit distance from one source token
- wildcard can express anchored intent such as prefixes, suffixes, and masked positions that fuzzy would either overmatch or fail to model cleanly

This is why wildcard is exposed as a separate lexical operator rather than being folded into `match.fuzzy`. The two modes are intentionally distinct and are not combined within a single query clause.

At execution time, wildcard remains a binary lexical filter. It expands matching vocabulary terms and checks membership, but does not preserve phrase ordering or emit a graded similarity score.

### 2.3 Index Backends (`FuzzyIndex` trait)
To support different storage modes and collection sizes, the fuzzy expansion layer is implemented behind a common `FuzzyIndex` trait:

Backend selection follows the same storage-mode decision as the underlying full-text index, so fuzzy support does not introduce a separate deployment model. Appendable text indexes use `MutableFuzzyIndex`; non-appendable indexes loaded into RAM use `ImmutableFuzzyIndex`; and fully on-disk mmap text indexes use `MmapFuzzyIndex`. In other words, the fuzzy vocabulary lives alongside the existing full-text index representation and mirrors its RAM-vs-disk behavior.

1. `MutableFuzzyIndex`
	 Backed by `BTreeSet<String>`. It seeks directly to the required prefix range and applies a cheap character-length lower bound before computing `strsim::levenshtein`. This is a good fit for mutable in-memory segments.

2. `ImmutableFuzzyIndex`
	 Backed by `fst::Set`. It traverses the finite-state transducer with a Levenshtein automaton and computes the exact distance only for matched terms. This is suited for immutable segments.

3. `MmapFuzzyIndex`
	 Persists the FST to disk and reopens it through memory mapping (`MmapFst`). This enables large fuzzy vocabularies without keeping the whole structure in heap memory.

For the FST-based backends, this PR also introduces `PrefixLevenshtein`, which enforces the exact prefix portion before entering the fuzzy automaton state space. `prefix_length` is interpreted in characters rather than bytes, so Unicode prefixes behave correctly.

### 2.4 `StrDist` Formula Query for Scoring

Fuzzy filtering improves recall, but it does not by itself provide payload-aware typo ranking. To address that, this PR adds `StrDistExpression` to the formula engine.

Syntax:

```json
{
	"str_dist": {
		"field": "title",
		"query": "quikly saphire falcon",
		"func": "jaro_winkler"
	}
}
```

Supported functions:

- `levenshtein`
- `jaro_winkler`

Current scoring behavior:

- `levenshtein` is implemented with `strsim::normalized_levenshtein`
- `jaro_winkler` is implemented with `strsim::jaro_winkler`
- both currently return a similarity-like score in `[0, 1]`
- higher is better for both functions in the current implementation

`str_dist` is the scoring layer in this design. Fuzzy filtering itself remains boolean: it expands candidate terms for recall, but it does not feed edit-distance weights into the final point score.

This makes `str_dist` a natural reranking layer on top of fuzzy retrieval.

## Compatibility

- public API changes are additive
- existing exact text filters continue to behave as before
- fuzzy is the main new lexical expansion mode, with wildcard available as a separate pattern mode
- no unsafe code is introduced
- wildcard and fuzzy behavior are opt-in at index creation time via `fuzzy_matching: true`

### Benchmark Results

Command used:

```bash
cargo bench -p segment --bench fuzzy_text_filter
```

Benchmark setup from the current harness:

- `50,000` documents
- `12` tokens per document
- vocabulary size `20,000`
- tokenizer: `word`
- lowercase normalization enabled
- measured backends in this benchmark run: `Mutable` and `Immutable`

High-level conclusion from the measured results:

- tokenization and posting traversal are cheap compared with fuzzy expansion; the main cost of fuzzy filtering is candidate-term expansion
- the immutable FST-backed implementation is consistently much faster than the mutable implementation in the measured single-token and multi-token cases
- increasing `max_edits` from `1` to `2` noticeably increases cost, so `max_edits=1` should be the practical default for most workloads
- in the measured corpus, fuzzy filtering remains within a reasonable latency range when bounded by the existing safety guards, but multi-token fuzzy queries are predictably more expensive than single-token queries

### E2E Spot Checks

In addition to the micro-benchmarks, we ran a few end-to-end spot checks using more realistic query shapes. These are not strict benchmarks, just directional measurements across the full request path.

We looked at three representative patterns:

exact lexical baseline: exact text filters on two indexed fields, with payload fetch
fuzzy rerank workload: fuzzy phrase prefilter on one field, followed by str_dist reranking on a title-like field
combined workload: wildcard on one field + fuzzy phrase on another, then the same rerank step

Exact rerank workload:
```json
{
  "prefetch": {
    "filter": {
      "must": [
        {
          "key": "text_field_b",
          "match": {
                "phrase": "query phrase",
          }
        }
      ]
    },
    "limit": 1600
  },
  "limit": 160,
  "query": {
    "formula": {
      "str_dist": {
        "field": "title_field",
        "query": "reference query text",
        "func": "jaro_winkler"
      }
    }
  },
  "with_payload": ["text_field_a", "text_field_b", "title_field", "doc_id"]
}
```

Measured result for this request shape:

```text
rps / p90 / p95 / p99
655.3 / 84 / 89 / 100
```

Fuzzy rerank workload:

```json
{
  "prefetch": {
    "filter": {
      "must": [
        {
          "key": "text_field_b",
          "match": {
            "fuzzy": [
              {
                "phrase": "query phrase",
                "params": {
                  "max_edits": 1,
                  "prefix_length": 2,
                  "max_expansions": 30
                }
              }
            ]
          }
        }
      ]
    },
    "limit": 1600
  },
  "limit": 160,
  "query": {
    "formula": {
      "str_dist": {
        "field": "title_field",
        "query": "reference query text",
        "func": "jaro_winkler"
      }
    }
  },
  "with_payload": ["text_field_a", "text_field_b", "title_field", "doc_id"]
}
```

Measured result for this request shape:

```text
rps / p90 / p95 / p99
392.8 / 120 / 120 / 130
```

Combined wildcard + fuzzy + rerank workload:

```json
{
  "prefetch": {
    "filter": {
      "must": [
        {
          "key": "text_field_a",
          "match": {
            "wildcard": "prefix*"
          }
        },
        {
          "key": "text_field_b",
          "match": {
            "fuzzy": [
              {
                "phrase": "query phrase",
                "params": {
                  "max_edits": 2,
                  "prefix_length": 0,
                  "max_expansions": 50
                }
              }
            ]
          }
        }
      ]
    },
    "limit": 1600
  },
  "limit": 160,
  "query": {
    "formula": {
      "str_dist": {
        "field": "title_field",
        "query": "reference query text",
        "func": "jaro_winkler"
      }
    }
  },
  "with_payload": ["text_field_a", "text_field_b", "title_field", "doc_id"]
}
```

Measured result for this request shape:

```text
rps / p90 / p95 / p99
127.9 / 410 / 420 / 450
```

Takeaways:
Exact filtering is still cheap — no expansion, no rerank.
Fuzzy processing, when applied within reasonable limits, introduces some overhead, but it generally remains well-controlled. Any noticeable increases in latency typically occur only in query patterns that combine multiple expansion-heavy features, reflecting the inherent complexity of those queries rather than unexpected behavior.
Overall, the behavior matches what we’d expect: higher recall configurations cost more, but the cost is coming from explicit knobs (max_edits, max_expansions, prefetch size), not from anything unbounded or surprising in the implementation.


generall:
Hey @VainJoker, thanks for your submission, I see you put a lot of work into your PR.
I recognize the need for this feature and would like to eventually be merged and released, however it would be quite challenging review process with such a big scope of changes.

This large PRs are usually taking us about a full-week to review, meaning that an engineed should continouesly spend several days on the PR review. And engineers have other tasks to do in parallel, so in practice the process will take even longer.

As you can see, your PR is already having conflicts, and longer it stays open - more conflicts it will have. For this reason, I would like to ask you to try to split your contribution into multiple independent parts, so we could merge them independently.

I also propose to start with:

- API design issue - a discussion of how final API should look like
- and tracking issue (see example https://github.com/qdrant/qdrant/issues/8670) - a place where all components and steps are tracked

Ideal size for individual PR could be around 500 lines (of course number if approximate, as some changes are pure mechanical while others are much more complicated). Each PR should be possible to merge independetly, so it doesn't expose unfinished functionality to the user, and doesn't introduce problems to release process.

Thank you again for your contribution, let's make it into release together.

generall (changes requested):
Too big to review in one step

VainJoker:
Hi @generall, thank you for your detailed feedback. 
I completely understand the challenges with reviewing such a large set of changes and the potential for conflicts over time. I appreciate your guidance and will start splitting the PR promptly, ensuring that each smaller PR can be merged independently without affecting unfinished functionality or the release process.

VainJoker:
Hi @generall 
As requested, I’ve created a tracing issue #8725  and submitted the first PR #8727

PR #8713, opened 2026-04-17
tests: stop uploader cleanly before killing peer in WAL delta tests

generall:
```python
upload_process_N.kill()   # SIGKILL the Python uploader
processes.pop().kill()    # SIGKILL the qdrant peer
```

is not reliable. If we kill peer while it is processing request - there are no guarantees about consistency and there are no upload process to report problem to.

So in some rare cases we can have inconsistent shard even after recovery process.

In this case recovery fallsback to streaming and there are no guarantees about which replica have update and which doesn't

Fix in test: do not kill between update operations

### DIFF
--- a/tests/consensus_tests/test_shard_wal_delta_transfer.py
+++ b/tests/consensus_tests/test_shard_wal_delta_transfer.py
@@ -10,11 +10,13 @@ from .utils import *
 COLLECTION_NAME = "test_collection"
 
 
-def update_points_in_loop(peer_url, collection_name, offset=0, throttle=False, duration=None):
+def update_points_in_loop(peer_url, collection_name, offset=0, throttle=False, duration=None, stop_event=None):
     start = time.time()
     limit = 3
 
     while True:
+        if stop_event is not None and stop_event.is_set():
+            break
         upsert_random_points(peer_url, limit, collection_name, offset=offset)
         offset += limit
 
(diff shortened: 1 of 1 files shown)


timvisee (commented):
This essentially adds a stop flag for uploads, good idea.

However, I'm not confident I fully understand the problem yet.

Killing the upload connection itself should not cause inconsistencies. We use Tokio's `spawn` [here](https://github.com/qdrant/qdrant/blob/acfb6503b1c2769349d2b668f3bcd06bf29fe9b3/lib/collection/src/collection/point_ops.rs#L156-L214) to make it cancel safe and prevent that.

In this case we're also killing a peer that is processing the update. That is of course critical and can result in partial application.

But then, why doesn't recover work when the peer is started again?

> In this case recovery fallback to streaming and there are no guarantees about which replica have update and which doesn't

Isn't the peer that we kill marked as dead in consensus because applying the update failed, and the rest of the peers are sound?

Or, do we completely stop update application on first failure (by killed peer) after which other replicas might also be missing the update?

timvisee:
We discussed this in call.

The critical problem here is that we kill the node we're also sending the update to. This node is responsible for distributing the update to all other peers, but if we kill it in the middle it might fail to distribute.

Normally the client is notified about this because the update will result in an error response. But here we explicitly ignore those because we kill the update sender.

This is a very specific edge case and the implemented fix here makes sense. All is still sound. :ok_hand:

timvisee (approved):

PR #8714, opened 2026-04-18
Low Memory mode

generall:
# Low memory mode

## Motivation

It is a frequent situation in production, when customer just keep pushing more data regardless of the machine capacity.
At some point capacity is exhausted and machine cashes. In worst case, machine goes into crash loop, and there are no
nice way to recover it from this situation, as we can't even change config as API are not available.

We need a way to recovery from this situation. 

## Proposal

Special configuration option `low_memory_mode` is added to the config. 

Should have 3 options:

- disabled (default) - no special handling, all collection modules are loaded as usual
- `no-resident`  
  When it is set, loading of all collection modules should not force anything to RAM if possible:
  * Quantization should be loaded as if `always_ram=false` and vectors are on disk
  * Payload indexes should be loaded as if `on_disk=true`

- `no-populate` - same as `no-resident`, but also no population of RAM from disk should be done. This affects loading of orginal vectors, HNSW index, payload storage

## Implementation details

* Make sure that all components that support loading into RAM or disk have compatible format on disk, so they can be loaded in both modes without any issues.
* Decide how to propagate parameter, either use global variable, or propagate it through function parameters. It would depend on how deep we need to propagate.
* Implement parameter check and handling in all relevant components

---

Testing scenario:

- load snapshot with all payloads and vector quantizations
- check out memory reporitng API with and without option enabled


lib/segment/src/vector_storage/quantized/quantized_vectors.rs
```diff
@@ -1860,6 +1861,12 @@ impl QuantizedVectors {
     }
 
     fn is_ram(always_ram: Option<bool>, on_disk_vector_storage: bool) -> bool {
+        // Low-memory mode forces the mmap (on-disk) backend regardless of the
+        // persisted `always_ram` flag. The on-disk byte layout is identical,
+        // so flipping back later still works without rebuild.
+        if low_memory_mode().prefer_disk() {
+            return false;
+        }
         !on_disk_vector_storage || always_ram == Some(true)
```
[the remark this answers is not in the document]
generall: It is used in ceration path, but onlu for the choise of structure and loading, on-disk format is sama and config is persisted independently. Validate that

config/config.yaml
```diff
@@ -42,6 +42,19 @@ storage:
...
+  # when a node crash-loops on out-of-memory.
+  #
+  # Options:
+  # - disabled (default): load segments as persisted.
+  # - no_resident: downgrade components to their on-disk variants where
+  #   possible — quantization loads as if always_ram=false, payload field
+  #   indexes as if on_disk=true, payload storage as mmap (not populated).
+  # - no_populate: same as no_resident, plus skip mmap prefault on load for
+  #   vectors, HNSW graph and payload storage.
+  #low_memory_mode: disabled
```
timvisee, line 56: If we only intend to use this in recovery scenarios, the middle option (`no_resident`) may be redundant.

To me it makes sense to either enable this mode or not. When it's enabled we do our very best effort to use the lowest amount of memory, which includes disabling population.

Any objection for removing the middle variant? Or do you have a clear motivation for this middle case?

lib/common/common/src/universal_io/disk_cache/cached_slice.rs
```diff
@@ -101,6 +101,12 @@ impl<T: bytemuck::Pod> CachedSlice<T> {
             return Ok(());
         }
 
+        // Low-memory mode `no_populate` suppresses mmap prefault globally.
+        // Pages will be faulted in on demand when queries touch them.
+        if crate::low_memory::low_memory_mode().skip_populate() {
+            return Ok(());
+        }
+
```
timvisee, line 109: It's my understanding the low memory mode should also suppress population of universal IO disk cache.

@xzfc could you also confirm this from your side?

lib/common/common/src/low_memory.rs
```diff
@@ -0,0 +1,105 @@
+use std::sync::OnceLock;
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+/// Global low-memory mode, initialized once at startup.
+static LOW_MEMORY_MODE: OnceLock<LowMemoryMode> = OnceLock::new();
```
timvisee, line 7: A pure atomic might be good enough for this case. It'd be cheaper, and we can represent the enum type as numeric value.

config/config.yaml
timvisee: Discussed this. Using `no_resident` as default may actually be more reasonable. It still prevents OOM crashes, but brings us in a more healthy for actual requests.

So I decided to keep the modes as-is. :+1:

lib/common/common/src/low_memory.rs
timvisee: Optimizing this would be overdoing it. It's still very fast.

lib/common/common/src/low_memory.rs
```diff
@@ -0,0 +1,105 @@
...
+    !is_on_disk && !low_memory_mode().skip_populate()
+}
+
+/// Initializes the global low-memory mode. Must only be called once at
+/// startup; subsequent calls are ignored with a warning.
+pub fn init_low_memory_mode(mode: LowMemoryMode) {
+    if LOW_MEMORY_MODE.set(mode).is_err() {
+        log::warn!("Low memory mode already initialized!");
+    }
+}
```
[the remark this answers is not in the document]
timvisee: Duplicate init should not happen. The warning will let us know there's an implementation problem.

lib/common/common/src/universal_io/disk_cache/cached_slice.rs
generall: I would say yes, because otherwise we can crash because of local disk cash is full

timvisee (approved):
Tested locally, works as expected 👌 

In fact, it clearly shows how slow loading into memory is for some of our storage components. In my test loading into memory takes 9 seconds, while starting with `no_resident` makes it startup in 0.5 seconds. I'm using a local NVMe disk.

PR #8715, opened 2026-04-18
low ram strict mode

generall:
# Prevent updates on high memory usage

## Motivation

Qdrant is usually running on a machine with limited resources, and as one of the fail-safe mechanisms, we need to prevent updates
when memory usage is higher than a certain threshold.

## Proposal

Implement strict mode configuration, what would validate memory usage before performing update operations.

Use jemalloc stats to get the current memory usage, it is OK to cache memory usage to check only every N seconds, to prevent impact on performance, in case if memory usage check is expensive.

It is important to not restrict deleting operations, as they are needed to free up memory, and if we restrict them, it can lead to a deadlock situation where no operations can be performed to reduce memory usage.

Memory checking should not affect internal operations, such as transfers and internal API calls, as it might provoke an inconsistent state and trigger shard transfers, we don't want that.

## Implementation details

Update operations with or without wait=true should report proper error with information about memory usage and threshold, so users can understand why the operation was rejected.

If strict mode is not enabled, there should be no change in behavior and no overhead for memory checking.


lib/storage/src/content_manager/collection_verification.rs
```diff
@@ -43,12 +43,24 @@ where
...
         for request in requests {
+            any_consumes_memory |= request.consumes_memory();
             request
                 .check_strict_mode(&collection, strict_mode_config)
                 .await?;
         }
 
+        if any_consumes_memory {
+            check_resident_memory(strict_mode_config, ::common::memory_usage::resident_bytes)?;
+        }
```
timvisee, line 62: Why don't we move this check into `check_strict_mode` per request above?

lib/common/common/src/memory_usage.rs
```diff
@@ -0,0 +1,46 @@
...
+/// Returns current process resident memory in bytes, served from a cached
+/// value that refreshes at most once per [`CACHE_TTL`].
+pub fn resident_bytes() -> Option<usize> {
+    let reader = *READER.get()?;
+
+    let mut guard = CACHE.lock().ok()?;
+    let now = Instant::now();
+
+    if let Some((cached_at, value)) = *guard
+        && now.duration_since(cached_at) < CACHE_TTL
```
timvisee, line 38:
```suggestion
        && now.duration_since(cached_at) <= CACHE_TTL
```

timvisee (approved):

lib/storage/src/content_manager/collection_verification.rs
timvisee: The batch level check is still a dedicated function and indeed a better place.

agourlay (approved):

PR #8731, opened 2026-04-20
A few improvements for parallel runs

tellet-q:
- Fix auth test port collisions under pytest-xdist --dist=loadfile: derive PORT_SEED from PYTEST_XDIST_WORKER in auth_tests/utils.py, remove now-inert xdist_group("auth") markers.
- Add -v to the CI pytest invocation so xdist tags per-test output with [gwN], making interleaved output and failure sites attributable to a worker.
- Add global timeout = 1800 (pytest-timeout) so a hung test fails with a traceback at 30 min instead of being masked by the 60-min job timeout.
- Switch LLVM_PROFILE_FILE from %m to %p so parallel COVERAGE=1 runs don't risk partial-merge corruption when peers are SIGKILLed.
- Kill leaked peer processes in every_test fixture

Example run: https://github.com/qdrant/qdrant/actions/runs/24710655731

### DIFF
--- a/tests/consensus_tests/utils.py
+++ b/tests/consensus_tests/utils.py
@@ -122,10 +122,8 @@ def get_qdrant_exec() -> str:
     return str(qdrant_exec)
 
 def get_llvm_profile_file() -> str:
-    # %m: keep merging results from each test into the same file
-    # If you have multiple tests running in parallel, you can use -%p OR -%{thread_count}m to have different files
-    # Not using -%p since each test will generate a new file
-    llvm_profile_file = PROJECT_ROOT / "target" / "llvm-cov-target" / "qdrant-consensus-tests-%m.profraw"
+    # %p (per-PID) avoids %m's partial-merge corruption when peers are SIGKILLed under -n auto.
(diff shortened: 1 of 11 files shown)


tests/consensus_tests/utils.py
```diff
@@ -18,6 +20,14 @@
 RETRY_INTERVAL_SEC = 0.2
 PROJECT_ROOT = Path(__file__).parent.parent.parent
 
+
+# Kernel SIGKILLs the child if the xdist worker dies abnormally, preventing orphaned peers.
+def _set_pdeathsig_sigkill():
+    if sys.platform != "linux":
+        return
+    PR_SET_PDEATHSIG = 1
+    ctypes.CDLL("libc.so.6", use_errno=True).prctl(PR_SET_PDEATHSIG, signal.SIGKILL, 0, 0, 0)
```
generall: what's this?

tellet-q: a SIGKILL, affects local runs mostly. Ensures that the spinned up qdrant processes are properly killed if tests are stopped mid-run

tellet-q: I reverted it btw

PR #8733, opened 2026-04-20
fix(segment): restore DenseVectorStorageImpl::is_on_disk semantics

agourlay:
PR #8606 changed `DenseVectorStorageImpl::is_on_disk()` from hard-coded `true` to `!self.populated`.

The method is load-bearing beyond reporting: `quantized_vectors.rs::is_ram()` reads it to decide whether to materialize the quantized blob as `QuantizedRamStorage` (heap) or `QuantizedMmapStorage` (disk) for every quantized variant (scalar, pq, binary, turbo).

For user configs with `on_disk: false` + quantization (no `always_ram`), the flip caused the quantized storage variant to move from disk to RAM, which in turn triggered segment rebuilds and transient 4x storage spikes during crasher-CI runs, hitting ENOSPC within 12-16 min vs. a stable 60 min on the parent commit.

Restore `is_on_disk() -> true` and expose `populated()` as a separate accessor. Memory reporter now reads `populated()` to pick `FileStorageIntent::Cached` vs. `OnDisk` -- the distinction the original PR wanted to surface -- without changing quantization placement.

Bisect: 4/4 green on aa3af834, 4/4 red on 5a899b74 (#8606).

### DIFF
--- a/lib/segment/src/vector_storage/dense/dense_vector_storage.rs
+++ b/lib/segment/src/vector_storage/dense/dense_vector_storage.rs
@@ -59,6 +59,15 @@ where
         }
     }
 
+    /// Whether this storage was pre-populated into the OS page cache at load
+    /// time (via `populate=true`). Distinct from [`VectorStorage::is_on_disk`]:
+    /// both populated and non-populated mmaps live on disk, but a populated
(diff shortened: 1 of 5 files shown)


agourlay:
does not fix crasher

Issue #8723, opened 2026-04-20
is_empty may miss cleared payload fields after payload index rebuild


litt1e-c:
## Current Behavior
`is_empty(status)` may intermittently miss a point after a payload field was cleared and the payload index was deleted and recreated.

I reduced this to a small reproducer with 3 points:

- point 1: `status = null`
- point 2: `status` missing
- point 3: `status = "live"`

Initial behavior is correct:

- baseline: `[1, 2]`
- after `clear_payload` on point 1: `[1, 2]`

However, after deleting and recreating the payload index on `status`, the result sometimes changes to:

- `[2]`

even though payload readback still shows:

- point 1: `{}`
- point 2: `{}`
- point 3: `{"status": "live"}`

So after rebuild, point 1 may be treated differently from point 2, even though both appear to have the same final payload state (`status` missing).

This is intermittent rather than deterministic. In my minimized script, I observed the issue in 6/20 runs.

## Steps to Reproduce
1. Create a collection with a keyword payload index on field `status`.
2. Insert 3 points:
   - `id=1`, payload `{"status": null}`
   - `id=2`, payload `{}`
   - `id=3`, payload `{"status": "live"}`
3. Verify that `is_empty(status)` returns `[1, 2]`.
4. Run `clear_payload` on point `1`.
5. Verify that `is_empty(status)` still returns `[1, 2]`.
6. Delete the payload index for `status`.
7. Recreate the payload index for `status`.
8. Run `is_empty(status)` again.

One observed failing run:

- baseline: `[1, 2]`
- after_clear: `[1, 2]`
- payloads_after_clear: `{1: {}, 2: {}, 3: {'status': 'live'}}`
- after_rebuild: `[2]`
- payloads_after_rebuild: `{1: {}, 2: {}, 3: {'status': 'live'}}`

A minimized Python reproducer is included below:

from qdrant_client import QdrantClient
from qdrant_client.http.models import (
    Distance,
    Filter,
    IsEmptyCondition,
    KeywordIndexParams,
    KeywordIndexType,
    PayloadField,
    PointStruct,
    VectorParams,
)

client = QdrantClient(host="127.0.0.1", port=6333)

name = "repro_is_empty_clear_rebuild"

try:
    client.delete_collection(name)
except Exception:
    pass

client.create_collection(
    collection_name=name,
    vectors_config=VectorParams(size=2, distance=Distance.DOT),
)

client.create_payload_index(
    collection_name=name,
    field_name="status",
[...]

PR #8734, opened 2026-04-20
Fix IsEmpty condition on null rebuilt index

timvisee:
Fixes <https://github.com/qdrant/qdrant/issues/8723>

AI summary:

Root cause (issue #8723): During build_field_indexes on payload-index rebuild, the null index's internal mmap flags track len from explicit `set()` calls. Points missing from payload storage (e.g., after `clear_payload`) never got an `add_point` call, so their IDs fell outside the flags' range — `iter_falses()` skipped them. Behavior was intermittent because it depended on where cleared points' internal IDs landed relative to the remaining points' IDs.

Fix: Thread `total_point_count` from the ID tracker into `MutableNullIndex::builder` → `open_or_create`, and pre-expand the `has_values` and `is_null` mmap flags to that length. This guarantees `iter_falses()` reports every segment point — cleared points included.

Files changed:
- `lib/segment/src/index/field_index/null_index/mutable_null_index.rs` — builder accepts `total_point_count`; `open_or_create` expands flags.
- `lib/segment/src/index/field_index/index_selector.rs` — null_builder threads the count.
- `lib/segment/src/index/struct_payload_index.rs` — build_field_indexes fetches the count from `id_tracker`.
- `tests/openapi/test_filter_is_empty.py` — regression test replicating the issue's reproducer.

Unit tests (`cargo test -p segment --lib -- null_index payload_index`) and the full-workspace cargo check `--all-targets` both pass.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/lib/segment/src/index/field_index/index_selector.rs
+++ b/lib/segment/src/index/field_index/index_selector.rs
@@ -375,10 +375,15 @@ impl IndexSelector<'_> {
         }
     }
 
-    pub fn null_builder(dir: &Path, field: &JsonPath) -> OperationResult<FieldIndexBuilder> {
+    pub fn null_builder(
+        dir: &Path,
+        field: &JsonPath,
+        total_point_count: usize,
+    ) -> OperationResult<FieldIndexBuilder> {
         // null index is always on disk and appendable
         Ok(FieldIndexBuilder::NullIndex(MutableNullIndex::builder(
             &null_dir(dir, field),
+            total_point_count,
         )?))
     }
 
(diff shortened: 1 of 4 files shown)


agourlay (approved):

JojiiOfficial (approved):

coszio (commented):
While this is correct, I think we can avoid growing the mmap if it is possible. I made this alternative PR as a suggestion -> #8736, which leverages most of this PR, except for the `set_len` part

coszio (approved):

PR #8736, opened 2026-04-20
[#8734] Alternative fix

coszio:
While reviewing I noticed that, while the #8734 fix works nicely, it can be possible to avoid growing the mmap for untracked points, if we know the total amount of points in the segment in advance.

This PR is just a suggestion as an alternative, which keeps the semantics of "any id that is not tracked by a bitslice should return `false`", and avoids growing the mmap just for `false` flags.


timvisee (approved):
Good suggestion. Thanks for the ready made PR!

PR #8749, opened 2026-04-20
TQ SIMD

IvanPleshkov:
## Summary

Adds SIMD-optimised scoring paths for every bit-width the `TurboQuant` encoder supports, and wires them into the quantiser so that both symmetric (vector-vs-vector) and asymmetric (query-vs-vector) scoring now run through the fastest kernel the host CPU offers. Follow-up to #8753 (TQ scoring) and #8783 (pre-SIMD refactor).

## What's new

Three sibling modules under `lib/quantization/src/turboquant/simd/` — `query1bit`, `query2bit`, `query4bit` — each exposing the same public surface:

- **`Query{N}bitSimd::new(query: &[f32])`** — rotation-applied query precomputed into the SIMD-friendly storage format (quantised integer halves, block-interleaved bit-planes for 1-bit). `dotprod(vector)` then scores against a PQ-encoded data vector.
- **`score_{N}bit_internal(a, b)`** — dot product between two already-encoded PQ vectors (symmetric path).

Both entry points dispatch at runtime to the best backend available:

| Bits | x86_64                                  | aarch64                |
|------|-----------------------------------------|------------------------|
|  1   | AVX-512 VPOPCNTDQ, AVX2, SSE4.1+SSSE3   | NEON                   |
|  2   | AVX-512 VNNI, AVX2, SSE4.1+SSSE3        | NEON + SDOT, NEON      |
|  4   | AVX-512 VNNI, AVX2, SSE4.1+SSSE3        | NEON + SDOT, NEON      |

A scalar reference kernel lives next to every SIMD path, serves as the fallback on other targets, and is the bit-exact parity baseline for the per-arch tests.

### Encoding strategy

- **1-bit**: codebook is `[-c, +c]`, so scoring collapses to `c² · (n_match − n_mismatch)` — i.e. `XOR + popcount`. The per-arch machinery is just picking the fastest popcount primitive (hardware `vpopcntq`, `pshufb`-nibble lookup, `vcntq_u8`). Query-side quantisation is `BITS`-bit (default 8) bit-plane-transposed and block-interleaved so the inner loop is a tight sequence of 16-byte `AND + popcount`s.
- **2/4-bit**: use arch-specific codebook representations to extract maximum precision out of each ISA's `i8×i8` / `u8×i8` dot primitive — full `i8 ∈ [−127,127]` on aarch64 (SDOT / `vmull_s8`), unsigned `u8 ∈ [0,255]` on x86_64 (`maddubs`, `VPDPBUSD`) with a `+128` shift absorbed into a per-query `bias_correction`. Query halves split as `q_signed = K · high + low` (K=256 on aarch64, K=128 on x86_64 to keep `maddubs` pair sums inside i16).

All codebooks derive from `CENTROIDS_{1,2,4}BIT` in `lloyd_max.rs`; `test_codebook_matches_lloyd_max` in each module catches drift.

### Tail handling

Every encoder is matryoshka-friendly: dims that don't fill a full SIMD chunk are handled either by a zero-padded scratch buffer fed through the same SIMD kernel (4-bit tail chunk, 1-bit block) or by a scalar fold-in helper (2-bit). Corner-case dims (`dim ∈ {16, 20, 28, 48, 60, 268, 1028, 2044, ...}`) are covered by parametrised parity tests.

### Integration into `TurboQuantizer`

- `score_symmetric` now dispatches to `score_{N}bit_internal` per `self.bits` (was going through f64 `unpack_vector + dot_impl`).
- `EncodedQueryTQData` replaces the old `Native(Precomputed)` variant with three SIMD variants — `Bits1(Query1bitSimd)`, `Bits2(Query2bitSimd)`, `Bits4(Query4bitSimd)`.
- `precompute_query` applies the Hadamard rotation, casts to f32, and hands off to the bit-width's SIMD encoder.
- `score_precomputed` splits the encoded vector into extras + data bytes and calls `Query{N}bitSimd::dotprod` directly — no f64 centroid materialisation.
- New `encoding.rs::split_vector` helper factors the extras-split out of `unpack_vector` so both the scalar and SIMD paths agree on the layout.

## Test plan

- [x] Per-arch SIMD parity: every `score_{N}bit_internal_{neon,sdot,sse,avx2,avx512*}` and `Query{N}bitSimd::dotprod_raw_*` is asserted bit-exactly against the scalar reference across matryoshka tail sizes.
- [x] Codebook consistency: `test_codebook_matches_lloyd_max` in each module re-derives `CODEBOOK_{I8,U8}` from `CENTROIDS_{N}BIT` and checks constants haven't drifted.
- [x] Precision: `test_dotprod_matches_float` / `test_score_*_matches_centroid_product` / `test_query_dotprod_matches_reference` verify the integer round-trip stays within bit-width-appropriate tolerance of the exact centroid dot.
- [x] Overflow safety: `test_score_neon_overflow_safety_64k` / `test_score_*_overflow_safety_64k` run 64 KiB of worst-case input through every 1-bit backend and compare to the u64-accumulator reference.
- [x] End-to-end dispatch: new `score_precomputed_dispatches_all_bit_widths` exercises `precompute_query → score_precomputed` across all three bit widths + Dot/Cosine (previously only `Bits4` was hit this way).
- [x] All existing TurboQuant precision tests (`score_approximates_true_similarity`, `score_self_similarity`, `score_antipodal_is_negative`, `rank_preservation`, `score_linearity_dot`, `higher_bits_reduce_error`) pass unchanged — tolerance bounds hold through the new SIMD path.
- [x] `codespell --toml tools/codespell.toml` clean (added `inout` to ignore-list — it's a Rust inline-asm operand keyword, not a typo).
- [x] `benches/turbo_simd.rs` builds on both architectures; covers each SIMD backend individually plus scalar reference and runtime-dispatch entry points.


lib/quantization/src/turboquant/simd/query4bit/x64.rs
```diff
@@ -0,0 +1,628 @@
...
+
+            let mut acc_low = _mm256_castsi256_si128(acc);
+            let mut acc_high = _mm256_extracti128_si256(acc, 1);
+
+            // Tail: one extra SSE chunk on a zero-padded scratch — matches
+            // the SSE variant's post-loop kernel.
+            if let Some(buf) = self.tail_chunk_scratch(vector) {
+                let nibble_mask_128 = _mm_set1_epi8(0x0F);
+                let ones_128 = _mm_set1_epi16(1);
+                let codebook_128 = _mm_loadu_si128(CODEBOOK_U8.as_ptr().cast::<__m128i>());
```
JojiiOfficial: Are we loading the code book again to free the registers during the main loop? Or is there another reason?

lib/quantization/src/turboquant/simd/query4bit/x64.rs
```diff
@@ -0,0 +1,628 @@
...
+
+/// SSE4.1 + SSSE3 implementation of [`super::score_4bit_internal`].
+///
+/// # Safety
+/// CPU must support `ssse3` and `sse4.1`.
+#[target_feature(enable = "sse4.1,ssse3")]
+pub unsafe fn score_4bit_internal_sse(a: &[u8], b: &[u8]) -> f32 {
+    use core::arch::x86_64::*;
+
+    assert_eq!(
```
JojiiOfficial, line 271: Do we maybe want to have assertions for the other simd functions too?

lib/quantization/src/turboquant/simd/query2bit/arm.rs
```diff
@@ -0,0 +1,404 @@
...
+        let mut rng = StdRng::seed_from_u64(7);
+        for &dim in PARITY_DIMS {
+            let (_, vec_a) = random_inputs(&mut rng, dim);
+            let (_, vec_b) = random_inputs(&mut rng, dim);
+            let scalar = score_2bit_internal_scalar(&vec_a, &vec_b);
+
+            let neon = unsafe { score_2bit_internal_neon(&vec_a, &vec_b) };
+            assert_eq!(scalar, neon, "score neon parity fail at dim {dim}");
+
+            if std::arch::is_aarch64_feature_detected!("dotprod") && dim.is_multiple_of(32) {
```
JojiiOfficial: Why multiple of 32?

lib/quantization/src/turboquant/simd/query4bit/x64.rs
IvanPleshkov: Fixed.
Interesting moment. For some implementations, such a change drops the criterion bench performance. I used AI to try different combinations and applied such changes if pefr is not affected

lib/quantization/src/turboquant/simd/query2bit/x64.rs
```diff
@@ -0,0 +1,622 @@
...
+                assert_eq!(scalar, q.dotprod_raw_avx2(&vector));
+            }
+            if std::is_x86_feature_detected!("avx512f")
+                && std::is_x86_feature_detected!("avx512bw")
+                && std::is_x86_feature_detected!("avx512vnni")
+            {
+                assert_eq!(scalar, q.dotprod_raw_avx512_vnni(&vector));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
IvanPleshkov: This is incorrect. Query2bitSimd::new does not apply any Hadamard rotation — it just quantizes the f32 query directly (see lib/quantization/src/turboquant/simd/query2bit/mod.rs, no hadamard/fwht/rotate anywhere in the module). The Walsh–Hadamard transform lives in the TurboQuant encoder and runs before a query reaches Query2bitSimd::new; the test deliberately bypasses it.

With vec![1.0; 65536] on x86 the encoding is:

q_abs_max = 1.0 → q_scale = QUERY_ABS_MAX / 1.0 = 8127
every dim hits the clamp: q_signed = 8127 (i.e. QUERY_ABS_MAX exactly)
decomposes to low = 63, high = 63 on every lane — query_data is filled with max-magnitude values, not "near-empty"
That is precisely what stresses the saturation paths:

maddubs / VPDPBUSD pair sum with c_raw = 3 (max-magnitude centroid, unsigned ≈ 255): 2 · 63 · 255 = 32130, right against the i16 ceiling of 32767.
i32 accumulator over 65 536 dims: ≈ 8127 · 127 · 65536 ≈ 6.76·10¹⁰, ~31× past i32::MAX — exactly the overflow this test is meant to catch.
So the test is doing what its name says; no change needed.

lib/quantization/src/turboquant/simd/query2bit/arm.rs
IvanPleshkov: Good catch - that gate was stale. Fixed

lib/quantization/src/turboquant/simd/query4bit/x64.rs
IvanPleshkov: Added, thanks!

JojiiOfficial (approved):

PR #8721, opened 2026-04-19
uio mmap posting

generall:
This is a Universal IO implementation of MmapPostings - a storage of posting lists for text index.

PR have the following stages:

1.  Refactoring of `MmapPostings` - it moves and splits structures into dedicated files, so it is easier to work with. There is also a detailed diagram of persisted format now.
2. Adjustment of `MmapPostings` methods. To make it more compatible with Universal IO, some of the methods of `MmapPostings` are updates to resemble batch operations better. Some non-needed iterators are replaces with pre-collection approach.
3. In parallel, `UniversalPostings` is implemented. It have similar methods as `MmapPostings`. It is done to gradully verify that we can migrate to universal IO, and not all at once.

As a last step, we replace `MmapPostings` with `UniversalPostings`, which should be not so hard, as we were adjusting the intefaces along the way

PR #8750, opened 2026-04-20
Evaluate on-disk posting lists lazily

coszio:
#8721 Migrated the full-text index posting lists to use UIO, but, in doing so, it introduced eager evaluation for getting all matching IDs. 

We have previously discussed that it **is** desired to do prefetching of the data that will be used for query. For this PR, I asked AI to propose a change to make evaluation lazy, so that there is no significant overhead for each request if the data is already cached in memory.

### DIFF
--- a/lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/lazy_iter.rs
+++ b/lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/lazy_iter.rs
@@ -1,0 +1,41 @@
+use common::types::PointOffsetType;
+use self_cell::self_cell;
+
+use crate::index::field_index::full_text_index::inverted_index::TokenId;
+use crate::index::field_index::full_text_index::inverted_index::mmap_inverted_index::raw_posting_list::RawPostingList;
+
+// Alias is required by `self_cell`: the dependent's lifetime parameter must be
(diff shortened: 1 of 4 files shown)


lib/segment/src/index/field_index/full_text_index/inverted_index/mmap_inverted_index/lazy_iter.rs
```diff
@@ -0,0 +1,41 @@
...
+    ///
+    /// Construction happens up-front (I/O is done in `fetch_*`), but the
+    /// actual query evaluation — posting-list traversal, intersection, phrase
+    /// matching — is driven on-demand from `next()`.
+    pub struct LazyPostingsIter<'a> {
+        owner: Vec<(TokenId, RawPostingList<'a>)>,
+        #[covariant]
+        dependent: BoxedPointOffsetIter,
+    }
+}
```
generall, line 25: Image

generall:
tbh, I am not convinced this is better. We will collect this lazy iterator anyway, but with lazyness we kinda loose branch predition. I would prefer to only merge this, if there are explicit benchmarks proving perfromance gain. 

Also, I don't think building those benchmarks are a priority at the moment

coszio:
I am ok revisiting this afterwards when we have time to benchmark

PR #8753, opened 2026-04-21
TQ Scoring

JojiiOfficial:
Implements scoring (internal + query) for cosine and dot distances.

Not included:
- 1.5 bit
- SIMD
- L1
- L2

This PR also implements all required functions of the `EncodedVectors` trait, meaning that TQ can now be used E2E from the API.

### DIFF
--- a/lib/segment/src/vector_storage/quantized/quantized_vectors.rs
+++ b/lib/segment/src/vector_storage/quantized/quantized_vectors.rs
@@ -813,16 +813,24 @@ impl QuantizedVectors {
             )?,
             QuantizationConfig::Turbo(TurboQuantization {
                 turbo: turbo_config,
-            }) => Self::create_turbo(
-                vectors,
-                &vector_parameters,
-                count,
-                turbo_config,
-                storage_type,
-                path,
-                on_disk_vector_storage,
-                stopped,
-            )?,
+            }) => {
+                let mut vector_parameters = vector_parameters;
+                // TODO(turbo): Remove this manual hack after releasing TQ. For more details see the TODO inside [`Self::construct_vector_parameters`]
+                if distance == Distance::Cosine {
+                    vector_parameters.distance_type = quantization::DistanceType::Cosine;
+                }
+
+                Self::create_turbo(
+                    vectors,
+                    &vector_parameters,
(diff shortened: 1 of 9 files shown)


lib/segment/src/vector_storage/quantized/quantized_vectors.rs
```diff
@@ -813,16 +813,24 @@ impl QuantizedVectors {
...
-                turbo_config,
-                storage_type,
-                path,
-                on_disk_vector_storage,
-                stopped,
-            )?,
+            }) => {
+                let mut vector_parameters = vector_parameters;
+                // TODO(turbo): Remove this manual hack after releasing TQ. For more details see the TODO inside [`Self::construct_vector_parameters`]
+                if distance == Distance::Cosine {
```
JojiiOfficial: We need to differentiate between Cosine and Dot distance in TQ. Therefore `Cosine` was added as new enum kind in the quantization crate's `DistanceType` enum.
However, this enum is persisted for all quantization storages (not only TQ). 
To allow downgrading, we're skipping this newly added kind in currently existing quantization storages and only add it for turboquant, which is a new feature anyway.

In a later release, we can remove this hack and allow persisting the correct Distance for other quantiaztion types too.
For non-TQ quantiaztion types it doesn't matter which type it is because dot and cosine haven't been differentiated before.

IvanPleshkov (approved):
List of changes:
- Use Vec::split_at in unpack_vector.
- unpack_vector always returns TqVectorExtras, no Option.
- unpack_extras_from check for extras size, no just empty vector.
- pack takes centroid indices, unpack returns centroid positions - no symmetry. Changed to floats, pack checks boundaries inside.
- change cosine metric setting for TQ
Another review remarks will be fixed while L2/L1 metrics implementation

lib/segment/src/vector_storage/quantized/quantized_vectors.rs
IvanPleshkov: I guess the approach in this commit is fine in combination with your changes
https://github.com/qdrant/qdrant/pull/8753/commits/6333d536ecc4427f881679821f2b6378d529d610

PR #8756, opened 2026-04-21
We dont actually need .worker_threads in search runtime

generall:


src/common/helpers.rs
```diff
@@ -12,7 +12,7 @@ use crate::settings::{Settings, TlsConfig};
 pub fn create_search_runtime(max_search_threads: usize) -> io::Result<Runtime> {
     let num_threads = common::defaults::search_thread_count(max_search_threads);
     runtime::Builder::new_multi_thread()
-        .worker_threads(num_threads)
+        .worker_threads(2.min(num_threads)) // 2 because of paranoia margin
```
timvisee, line 15: In my opinion we should set this to zero, but it appears that Tokio doesn't allow that and panics.

I therefore vote to fully replace this Tokio runtime with a threadpool that is sync-only. That way we prevent accidental misuse and eliminate potential foot guns. We can do that in a separate PR.

timvisee (approved):

src/common/helpers.rs
generall: I have some things to try with semafors, I am not sure we want something like Rayon here, as we don't want to pre-allocate all threads.

timvisee: No, a proper dynamic thread pool that fits our use case :+1:

